PR #2042, opened 2017-12-20
fix IsMultiplicativeElementWithOneTable documentation

### LINKED ISSUE #1237: IsMultiplicativeElementWithZero false when it should be true
MultiplicativeElementWithZero means that an element can be multiplied and zeroed; Zero(x) has the property that Zero(x)=Zero(y) and Zero(x)*y=Zero(x) for all x,y.

From this, I would expect
`IsMultiplicativeElementWithZero(1)=true`; but it's `false`.

I ask this because I implemented Nillity for multiplicative objects with zero: "is there n>0 such that x^n=0". The correct filter seemed to be IsMultiplicativeElementWithZero.

fingolfin:
It referred to  MultiplicativeZero, but that is only applicable to
magmas, not elements; for the latter, one should use
MultiplicativeZeroOp. The example a bit below also correctly does that.

Resolves #1237

### DIFF
--- a/lib/arith.gd
+++ b/lib/arith.gd
@@ -380,7 +380,7 @@ DeclareSynonym( "IsMultiplicativeElementWithOneTable",
 ##  <Description>
 ##  This is the category of elements in a family which can be the operands of 
 ##  <C>*</C> (multiplication) and the operation 
-##  <Ref Attr="MultiplicativeZero"/>.
+##  <Ref Attr="MultiplicativeZeroOp"/>.
 ##  <Example><![CDATA[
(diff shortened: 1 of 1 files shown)


laurentbartholdi:
All OK for me.

fingolfin:
@laurentbartholdi any particular reason you closed this?

james-d-mitchell:
What does "merge `MultiplicativeZeroOp` into `MultiplicativeZeroOp`" mean?

james-d-mitchell:
Also this is all ok with me.

fingolfin:
@james-d-mitchell sorry, the second "Op" was erroneous, I fixed my commented

PR #2045, opened 2017-12-20
Make T_EMPTY statements interruptible again

### LINKED ISSUE #2044: Problem with Ctrl-C interruption
### Observed behaviour
Running the following code in GAP master branch
```
gap> while true do od;
```
I cannot interrupt the loop with Ctrl-C (but I can stop gap with double Ctrl-C).

### Expected behaviour
The loop should be interruptable, as it was in all older versions of GAP which I tried (at least up to 4.8.6).

fingolfin:
Fixes #2044: interrupting `while true do od;` with ctrl-c is now possible again

### DIFF
--- a/src/code.h
+++ b/src/code.h
@@ -131,18 +131,23 @@ enum STAT_TNUM {
 
         T_PROCCALL_OPTS,
 
-        // The statement types between FIRST_NON_INTERRUPT_STAT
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
This looks good, thanks

fingolfin:
Now changed to target stable-4.9

frankluebeck:
Looks good, thanks.

PR #2048, opened 2017-12-21
Remove InstallIsomorphismMaintenanceFunction and some HPC-GAP syncing

fingolfin:
`InstallIsomorphismMaintenanceFunction` was added by @ThomasBreuer to support a package called `gpisotyp`. I have never heard about or seen this package. Nothing else seems to use it, and it is undocumented.

This PR removes it, and some related code, and also reduced diffs between the plain GAP and HPC-GAP versions of `lib/coll.{gi,gd}`.

It would be good to hear from @ThomasBreuer if he still needs `InstallIsomorphismMaintenanceFunction` after all?

### DIFF
--- a/hpcgap/lib/coll.gd
+++ b/hpcgap/lib/coll.gd
@@ -354,18 +354,24 @@ end );
 ##  <#/GAPDoc>
 ##
 BIND_GLOBAL( "DeclareSynonym", function( name, value )
-    BIND_GLOBAL( name, value );
+    if ISBOUND_GLOBAL(name) and IS_IDENTICAL_OBJ(VALUE_GLOBAL(name), value) then
+        if not REREADING then
+            INFO_DEBUG( 1, "multiple declarations for synonym `", name, "'\n" );
+        fi;
+    else
+        BIND_GLOBAL( name, value );
+    fi;
 end );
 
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
As far as I am concerned, there is no problem with the removal of `InstallIsomorphismMaintenanceFunction`.

hpcgap/lib/coll.gd
```diff
@@ -230,48 +230,47 @@ BIND_GLOBAL( "CategoryCollections", function ( elms_filter )
...
-        # Find the super category among the known collections categories.
-        super := IsCollection;
-        flags := WITH_IMPS_FLAGS( FLAGS_FILTER( elms_filter ) );
+    # Check whether in the meantime another thread defined the collections category
+    if LENGTH(CATEGORIES_COLLECTIONS) > len then
         for pair in CATEGORIES_COLLECTIONS do
-            if IS_SUBSET_FLAGS( flags, FLAGS_FILTER( pair[1] ) ) then
-                super := super and pair[2];
+            if IsIdenticalObj( pair[1], elms_filter ) then
+                return pair[2];
```
fingolfin: To reviewers: the changes made to `CategoryCollections` really only are whitespace changes -- the indentation was adjusted so that the GAP and HPC-GAP version have fewer diffs.

PR #2050, opened 2017-12-21
Allow Int(""), reject null bytes, clarify docs (for stable-4.9)

### LINKED ISSUE #2047: Int( "" ) does not work anymore?
Up to GAP 4.8.8, `Int( "" )` returns `0`.
In the master branch, one gets `fail`.
This change breaks some of my code.
Is there a good reason for this change?

fingolfin:
Strings which contain a null byte are now rejected by `Int`, instead of ignoring them and any characters after them.

Moreover, `Int("")` and `Int("-")` again returns 0 (as was the case in all previous GAP versions). I allowed both because I realized that also `Rat` accepts empty strings (i.e. `Rat("/1")` and `Rat("-/1")` both return `0`, and this way, we can write a concise uniform definition: we expect an optional sign character, followed by a sequence of decimal digits -- and that sequence can of course also be empty.

Also clarify the documentation of `Int`, `Rat` and `IntHexString` for strings: The primary reference entry for `Int` now states what happens in case of invalid input. It also now links to the secondary entry "Int for strings".

The documentation for `Int`, `Rat` and `IntHexString` for strings was previously jumbled together. Split this into three separate entries, and be more precise in documenting which inputs are accepted, and what happens for invalid inputs.

Fixes #2047

### DIFF
--- a/doc/ref/string.xml
+++ b/doc/ref/string.xml
@@ -458,7 +458,7 @@ see&nbsp;<Ref Sect="String Streams"/>.
 <Func Name="HexStringInt" Arg='int'/>
 
 <Description>
-returns a string which represents the integer <A>int</A> with hexa-decimal
+returns a string which represents the integer <A>int</A> with hexadecimal
 digits (using <C>A</C> to <C>F</C> as digits <C>10</C> to <C>15</C>).
 The inverse translation can be achieved with <Ref Func="IntHexString"/>.
 </Description>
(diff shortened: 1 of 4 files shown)


fingolfin:
I fixed the test failure, by forbidding `Int("--")` and `Int("--1")`. This then also matches the documentation updates in this PR.

alex-konovalov:
@fingolfin thanks, I am happy with this PR now. I'm not a big fan of convention that `Int("")` is zero, but on the other hand removing it breaks backwards compatibility. 

I was thinking where this may be useful - perhaps when reading a comma-separated list of integers with holes which have to be replaced by zeroes (OTOH, in another setting one may want to convert `"1,,,3"` to `[1,,,3]` instead of `[1,0,0,3]`).

I would leave this PR open for a bit more, aiming to merge it for GAP 4.9.0, but giving others more time to look at it.

PR #2052, opened 2017-12-21
Test/TestDirectory: default to `showProgress:=false` if stdout isn't a tty (for stable-4.9)

### LINKED ISSUE #2049: Changed default behaviour of Test
The documented default for the `showProgress` component
of the optional argument record for `Test` was `false` up to GAP 4.8.8.
In the master branch, the documented default is `"some"`.
This means that every personal test setup will have to be changed
in order to set the value explicitly to `false`.
Was this change really intended?

fingolfin:
Thus if you pipe the output of Test or TestDirectory into a file, no progress output is generated.

Note that only the default setting is changed; you can still manually override the `showProgress` setting.

Should resolve #2049 -- curious to hear what @ThomasBreuer thinks.

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -91,6 +91,9 @@ InstallGlobalFunction(RunTests, function(arg)
   # don't enter break loop in case of error during test
   tests := arg[1];
   opts := rec( breakOnError := false, showProgress := "some" );
+  if not IS_OUTPUT_TTY() then
+    opts.showProgress := false;
+  fi;
   if Length(arg) > 1 and IsRecord(arg[2]) then
     for f in RecNames(arg[2]) do
       opts.(f) := arg[2].(f);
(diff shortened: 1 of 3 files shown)


ThomasBreuer:
This looks fine.
Thank you for this change.

Once `IS_OUTPUT_TTY` becomes available in a released GAP version,
I will use it for example in some places in the Browse package.

lib/test.gi
```diff
@@ -535,7 +542,8 @@ end);
 ##  <Item>If <K>true</K>, stop as soon as any <Ref Func="Test" /> fails (defaults to <K>false</K>).
 ##  </Item>
 ##  <Mark><C>showProgress</C></Mark>
-##  <Item>Print information about how tests are progressing (defaults to <K>"some"</K>).
+##  <Item>Print information about how tests are progressing (defaults to <K>"some"</K>
```
alex-konovalov: This is the the place to replace <K> by <C> as @ThomasBreuer mentioned.

fingolfin:
Fixed the GAPDoc issue, thanks for spotting it!

alex-konovalov:
@fingolfin thanks - happy to merge this after next CI tests round.

PR #2056, opened 2018-01-01
Fix PositionProperty with from < 1

james-d-mitchell:
This PR fixes a bug in `PositionProperty(list, func, 0)` which previously only started trying to find an item in `list` for which `func` returns `true` from `list[2]` onwards. This contradicts both the documentation and is not consistent with how `Position` behaves.

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -1525,8 +1525,8 @@ InstallMethod( PositionProperty,
     function( list, func, from )
     local i;
 
-    if from < 1 then
-      from:= 1;
+    if from < 0 then
+      from:= 0;
     fi;
     for i in [ from+1 .. Length( list ) ] do
       if IsBound( list[i] ) then
@@ -1557,8 +1557,8 @@ InstallMethod( PositionProperty,
     function( list, func, from )
     local i;
 
-    if from < 1 then
-      from:= 1;
+    if from < 0 then
+      from:= 0;
     fi;
     for i in [ from+1 .. Length( list ) ] do
       if func( list[i] ) then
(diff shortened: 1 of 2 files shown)


fingolfin:
I think you mean `PositionProperty` everywhere? (Including the commit message, which thus should be adjusted) ?

fingolfin:
This is something that should definitely be mentioned in release notes, as it could easily affect the behavior of some existing code (although I'd hope anybody who actually used it would have noticed the bug, too, and report it; since we saw no report, hopefully just nobody used this feature?)

lib/list.gi
```diff
@@ -1557,8 +1557,8 @@ InstallMethod( PositionProperty,
     function( list, func, from )
     local i;
 
-    if from < 1 then
-      from:= 1;
+    if from < 0 then
+      from:= 0;
```
fingolfin: The exact same bug is in the code 30 lines before, please fix it there, too.

fingolfin:
For cream with a cherry on top, perhaps also add a test case? :-)

james-d-mitchell:
Done, with cream and a cherry on top.

james-d-mitchell:
Thanks @fingolfin I don't really have an opinion about whether it should go in 4.9.x or 4.10.x. Probably earlier is better since it's a bug fix, but maybe there are some reasons not to include it in 4.9.x?

fingolfin:
While this is technically a bug fix, there is a slight chance that some code relies on the existing broken behaviour. As such, putting this into GAP 4.9 would be a small risk. Personally, I'd be OK with that, but it's a decision that should be made consciously.

In order to put this into stable-4.9, though, this PR should be retargeted and then updated (so that it based on that branch, not `master`) -- ideally in that order, because retargeting a PR does not trigger a Travis rebuild. Could you do that, please, @james-d-mitchell ?

ChrisJefferson:
I feel this should be in 4.9, I suspect it is more likely someone has written incorrect code without realising, than they are relying on the behaviour.

fingolfin:
I changed the base to stable-4.9, but now @james-d-mitchell needs to rebase

PR #2035, opened 2017-12-19
FIX/ENHANCE: Small fixes and corrections, staged for 4.9.1

hulpke:
Includes #2025

Fix for #2055. #2104, #2131 (as far as LowIndex is concerned)

Multiple documenttion/testfile/formatting fixes.

PR #2057, opened 2018-01-01
Ongoing improvements that do not need to make 4.9

hulpke:
- Better behavior if FpGroups are used for serious calculations (since some users seem to love word expressions).
- Replace remaining `FactInt` calls.
- Improvent to heuristics for small matrix groups.
- Improved performance to GQuotients if there are many maps and more than two generators.

### DIFF
--- a/lib/ghomfp.gi
+++ b/lib/ghomfp.gi
@@ -953,6 +953,7 @@ local H, pres,map,mapi,opt;
   SetIsBijective(map,true);
   SetInverseGeneralMapping(map,mapi);
   SetInverseGeneralMapping(mapi,map);
+  ProcessEpimorphismToNewFpGroup(map);
 
   return map;
 end );
@@ -1226,6 +1227,46 @@ local F,str;
   fi;
   return 
     GroupHomomorphismByImagesNC(F,G,GeneratorsOfGroup(F),GeneratorsOfGroup(G));
+end);
+
+InstallGlobalFunction(ProcessEpimorphismToNewFpGroup,
+function(hom)
+local s,r,fam,fas,fpf,mapi;
+  if not (HasIsSurjective(hom) and IsSurjective(hom)) then
+    Info(InfoWarning,1,"fp eipimorphism is created in strange way, bail out");
+    return; # hom might be ill defined.
+  fi;
+  s:=Source(hom);
+  r:=Range(hom);
+  mapi:=MappingGeneratorsImages(hom);
+  if mapi[2]<>GeneratorsOfGroup(r) then
+    return; # the method does not apply here. One could try to deal with the
+    #extra generators separately, but that is too much work for what is
+    #intended as a minor hint.
+  fi;
+  s:=SubgroupNC(s,mapi[1]);
+  fam:=FamilyObj(One(r));
+  fas:=FamilyObj(One(s));
+  if IsPermCollection(s) or IsMatrixCollection(s) 
+     or IsPcGroup(s) or CanEasilyCompareElements(s) then
+    # in the long run this should be the inverse of the source restricted
+    # mapping (or the corestricted inverse) but that does not work well with
+    # current homomorphism code, thus build new map.
+    #fpf:=InverseGeneralMapping(hom);
+    fpf:=GroupHomomorphismByImagesNC(r,s,mapi[2],mapi[1]);
+  elif IsFpGroup(s) and HasFPFaithHom(fas) then
+    #fpf:=InverseGeneralMapping(hom)*FPFaithHom(fas);
+    fpf:=GroupHomomorphismByImagesNC(r,s,mapi[2],List(mapi[1],x->Image(FPFaithHom(fas),x)));
+  else
+    fpf:=fail;
+  fi;
+  if fpf<>fail then
+    SetEpimorphismFromFreeGroup(ImagesSource(fpf),fpf);
+    SetFPFaithHom(fam,fpf);
+    SetFPFaithHom(r,fpf);
+    if IsPermGroup(s) then SetIsomorphismPermGroup(r,fpf);fi;
+    if IsPcGroup(s) then SetIsomorphismPcGroup(r,fpf);fi;
+  fi;
 end);
 
 #############################################################################
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -3723,7 +3723,6 @@ local fgens,grels,max,gens,t,Attempt;
   gens:=Concatenation([t,
     #pseudorandom element - try if it works
     PseudoRandom(G:radius:=Random(2,3))],
-    #Product([1..Random([2,3])],i->Random(gens)^Random([1,-1]))],
     Filtered(gens,j->UnderlyingElement(j)<>UnderlyingElement(t)));
 
   # recursive search (via smaller and smaller partitions) for a finite index
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -4360,6 +4360,43 @@ local A,B,Q,gens,int,i,fam;
   return rec(subgroups:=B,inclusions:=int.inclusions);
 end);
 
+# test whether abelian invariants can be mapped
+InstallGlobalFunction(CanMapFiniteAbelianInvariants,function(from,to)
+local pf,pt,fp,tp,p,i,f;
+  # first get primes and then run for each prime
+  pf:=Union(List(from,Factors));
+  pt:=Union(List(to,Factors));
+  if not IsSubset(pf,pt) then
+    return false;
+  fi;
+  for p in pf do
+    fp:=[];
+    for i in from do
+      f:=Filtered(Factors(i),x->x=p);
+      if Length(f)>0 then
+        Add(fp,Product(f));
+      fi;
+    od;
+    tp:=[];
+    for i in to do
+      f:=Filtered(Factors(i),x->x=p);
+      if Length(f)>0 then
+        Add(tp,Product(f));
+      fi;
+    od;
+    #Print(fp,tp,"\n");
+    if Length(fp)<Length(tp) then return false;fi;
+    Sort(fp);Sort(tp);
+    fp:=Reversed(fp);
+    tp:=Reversed(tp);
+    if ForAny([1..Length(tp)],i->fp[i]<tp[i]) then
+      return false;
+    fi;
+  od;
+  return true;
+end);
+
+
 #############################################################################
 ##
 #F  GQuotients(<F>,<G>)  . . . . . epimorphisms from F onto G up to conjugacy
@@ -4377,6 +4414,9 @@ local Fgens,	# generators of F
       val,	# its value
       i,j,	# loop
       ma,
+      dp,emb1,emb2, # direct product
+      sameKernel,
+      A,bigG,Gmap,opt,
       h;	# epis
 
   Fgens:=GeneratorsOfGroup(F);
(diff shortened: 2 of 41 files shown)


lib/grpfp.gi
```diff
@@ -3720,7 +3720,8 @@ local fgens,grels,max,gens,t,Attempt;
   t:=MostFrequentGeneratorFpGroup(G);
   gens:=Concatenation([t,
     #pseudorandom element - try if it works
-    Product([1..Random([2,3])],i->Random(gens)^Random([1,-1]))],
+    PseudoRandom(G:radius:=Random(2,3))],
+    #Product([1..Random([2,3])],i->Random(gens)^Random([1,-1]))],
```
fingolfin: This change is introduced by the commit adding `ProcessEpimorphismToNewFpGroup` -- an accident?

lib/grpfp.gi
```diff
@@ -4935,7 +4936,17 @@ local iso,k,id,f;
...
+  #return function(w)
+  #  if not w in FreeGroupOfFpGroup(Source(iso)) then Error("flasch");fi;
+  #  w:=ElementOfFpGroup(fam,w);
+  #  Print("wa=",w,"\n");
+  #  w:=ImageElm(iso,w);
+  #  Print("wb=",w,"\n");
+  #  w:=Factorization(k,w);
+  #  Print("wc=",w,"\n");
+  #  return UnderlyingElement(w);
+  #end;
```
fingolfin: Also this change is introduced by the commit adding `ProcessEpimorphismToNewFpGroup` -- another accident?

tst/testinstall/semigrp.tst
```diff
@@ -477,7 +477,7 @@ true
 # test the methods for Random
 gap> S := FreeSemigroup(3);;
 gap> Random(S);
-s3*s1
+s1*s2
```
fingolfin: This test change perhaps was made necessary by those other accidental changes?

lib/ghomfp.gi
```diff
@@ -1229,6 +1230,44 @@ local F,str;
...
+  if mapi[2]<>GeneratorsOfGroup(r) then
+    return; # the method does not apply here. One could try to deal with the
+    #extra generators separately, but that is too much work for what is
+    #intended as a minor hint.
+    Error("fp generators?");
+  fi;
+  s:=SubgroupNC(s,mapi[1]);
+  fam:=FamilyObj(One(r));
+  fas:=FamilyObj(One(s));
+  if IsPermCollection(s) or IsMatrixCollection(s)
```
fingolfin: Any particular reason you are not using `IsPermGroup(s) or IsMatrixGroup(s)` here? There should be no functional difference, but IMHO it conveys the intention of the code more naturally.

lib/ghomfp.gi
```diff
@@ -1229,6 +1230,44 @@ local F,str;
...
+    return; # hom might be ill defined.
+  fi;
+  s:=Source(hom);
+  r:=Range(hom);
+  mapi:=MappingGeneratorsImages(hom);
+  if mapi[2]<>GeneratorsOfGroup(r) then
+    return; # the method does not apply here. One could try to deal with the
+    #extra generators separately, but that is too much work for what is
+    #intended as a minor hint.
+    Error("fp generators?");
```
fingolfin: Remove that `Error`?

lib/ghomfp.gi
```diff
@@ -1229,6 +1230,44 @@ local F,str;
...
+  fi;
+  s:=SubgroupNC(s,mapi[1]);
+  fam:=FamilyObj(One(r));
+  fas:=FamilyObj(One(s));
+  if IsPermCollection(s) or IsMatrixCollection(s) 
+     or IsPcGroup(s) or CanEasilyCompareElements(s) then
+    #fpf:=InverseGeneralMapping(hom);
+    fpf:=GroupHomomorphismByImagesNC(r,s,mapi[2],mapi[1]);
```
fingolfin: Why not `InverseGeneralMapping`? Naively, that would seem more attractive: easier to understand the intent of the code that way, and of course the inverse is cached. Are there perhaps some performance concerns in certain cases? If so, it would be nice if there was a comment explaining what they are (also, perhaps we can resolve them?)

lib/ghomfp.gi
```diff
@@ -1229,6 +1230,44 @@ local F,str;
     GroupHomomorphismByImagesNC(F,G,GeneratorsOfGroup(F),GeneratorsOfGroup(G));
 end);
 
+InstallGlobalFunction(ProcessEpimorphismToNewFpGroup,
+function(hom)
+local s,r,fam,fas,fpf,mapi;
+  if not (HasIsSurjective(hom) and IsSurjective(hom)) then
+    Info(InfoWarning,1,"ill created epimorphismfp");
+    return; # hom might be ill defined.
```
fingolfin: Why is this only a warning? Shouldn't it be an error?

lib/ghomfp.gi
```diff
@@ -1229,6 +1230,44 @@ local F,str;
     GroupHomomorphismByImagesNC(F,G,GeneratorsOfGroup(F),GeneratorsOfGroup(G));
 end);
 
+InstallGlobalFunction(ProcessEpimorphismToNewFpGroup,
+function(hom)
+local s,r,fam,fas,fpf,mapi;
+  if not (HasIsSurjective(hom) and IsSurjective(hom)) then
+    Info(InfoWarning,1,"ill created epimorphismfp");
```
fingolfin: "ill created" ? Perhaps "ill defined"?

lib/grpfp.gi
hulpke: I spotted this while going over code that uses homomorphisms with fp groups. It seemed to small a change to deserve its own commit. (The `PseudoRandom` is cleaner than the `Random(gens)^Random([-1,1]) hack.)

lib/grpfp.gi
hulpke: That is (commented out) debugging code in case an error arises later. I'll look before merging whether it is worth keeping.

tst/testinstall/semigrp.tst
hulpke: I don't think its the accidental changes but that an fp group created earlier from a finite group now immediately knew a faithful rep. and did not have to construct it.

In any case, tests for explicit results of `Random` should be expected to be highly unstable.

lib/ghomfp.gi
hulpke: That depends on whether the function is considered internal (what I believe) or user.

My understanding was that all cases in which I call the function at this point have a homomorphism that will not trigger this warning, but there is no need to make it stop the calculation (there is a minor, probably nonexistent, efficiency pay.) Basically this is a warning that should never be triggered by the existing code, but could be triggered in the future if other code creating homomorphisms changes.

lib/ghomfp.gi
hulpke: The reason for the category test was that I was thinking in the context of `CanEasily...`. It could as well be the `Group` test.

lib/ghomfp.gi
hulpke: What it actually should be is the co-restricted inverse (or the inverse of the source-restriction. This is iffy now, so I explicitly make a new map to guarantee that there is no larger gropup somewhere hidden.

lib/ghomfp.gi
fingolfin: Makes sense, thanks.

lib/ghomfp.gi
fingolfin: I see. That would be perfect for a comment; anybody looking at that code again in the future might wonder the same

lib/grpfp.gi
fingolfin: OK, so then, remove the old code, and mention it in the commit message?

lib/grpfp.gi
fingolfin: OK, fair enough

tst/testinstall/semigrp.tst
fingolfin: Sure.

fingolfin:
@hulpke This has a conflict now, which prevents it from merging.

Meta-question: So when is this "ready for merge"? I reviewed and approved this PR at some point, but since then it underwent major changes, making my "approval" worthless.

All in all, I find this approach of amassing changes in a single PR and then merging it all in one go problematic: it makes the PR a moving target for review; and it makes it unclear when to merge it. It also causes extra work when creating release notes, as a lot of (often unrelated) changes are grouped into a single PR.

E.g. the `FactorsInt` changes could have been merged a few weeks ago, and would have been tested since then by everybody working with the GAP master branch. If they cause any regressions, we might have already noticed and resolve them. As long as that change is stuck in this unmerged PR, this won't happen. At the same time, that single commit touches tons of files, cluttering up the diff for this PR, making a global review harder (one can still review commit-by-commit, though).

I am not completely sure why you work this way. I assume part is convenience of having all your fixes immediately at hand. That of course explains why one would want to keep those fixes on a single "work branch" -- but doing that (something I also sometimes do) does not make it necessary to put all commits in a single PR (and vice versa).

Instead, what I do, is to put separate changes/fixes into separate branches, and then into separate PRs. While locally, I may have an `integration` branch, where all my various bugfix branches are merged together. The only time this gets annoying is when one has to update a fix branch, and then has to update the `integration` branch. But there are tools to automate this. E.g. I use `git reintegrate` from https://github.com/felipec/git-reintegrate

hulpke:
@fingolfin 
I have rebased and the conflict is gone.

As for your meta-question, I've come from the old setup of CVS and it took a while to be comfortable with git in a way the did not cause issues for others, nor imposed a higher administrative cost on me.

What I settled is that I have my own repository copy that updates from GAP (master and current stable branch), and I have a work branch for myself (that gets updated from time to time to master.

To move things back into the repository, in the past I had just moved them into the (equivalent of master). This was considered as inappropriate and also missed the automatic tests. I thus have two branches (additions and fixes) into which I put my changes (through pull requests) back to master, respectively stable. (That is, I do the fix in my work branch and then cherry pick into (e.g.) additions.)

What I like about this setup is that it deals efficiently with changes that are not my current research interest, but which I have done as the person who caused the problem in the first place or knows the code best, often  in response to a bug or a request (such as the small dimension matrix groups, or the issues with automorphism groups of fp groups).

I want to be able to take these such quickly and put them into a processing queue that lets me forget about the issue and go quickly back to my own work, while ensuring that the changes eventually go back to the production system. Making a separate branch and PR for each such change takes longer and also forces me to monitor multiple PRs to see that they eventually have gone in.

This setup predates newer changes (such as the now mandated reviews) and assumes that a merge will happen within a few days, cleaning the queue. (It still is not clear to me who should merge -- at times nobody did so, and when I finally merged I was told this was wrong.)

This setup predates recent changes (such as the mandated reviews, or requests to not merge into release-ready branches. I understand why this is happening, but for issues that I feel obliged to do, I am looking rather for a "Babyklappe" than for opportunity (or obligation) to discuss a minor fix in a global context or alternative options.

I'm perfectly happy to try another model that does not require more attention or time. If there is a guarantee that someone will take care of stale PR's I can put subject specific changes into different PRs.  What I don't want to do is to have to go back a few weeks later and see whether changes I proposed were in fact incorporated or might still be hanging or have been dropped.

(This all affect such small fixes. I am prefectly happy to work through specific branches if it is substantial new material.)

fingolfin:
Now to reply to your last comment, in brief: I am happy to promise to handle small, self-contained PRs quickly! (And I am hopeful that @alex-konovalov, @ChrisJefferson, and @markuspf will do so, too, once they are not on strike anymore). If I don't then usually because I missed the notification (on some days, there are just to many, and something slips through -- I usually try to scan through open PRs from time to time to catch anything I missed, but for various reasons this failed for this PR and also #2035.

Anyway: I totally get what you are doing with your "additions" and "fixes" branches, from the developers perspective

The problem I have, as a reviewer, is dealing with big PRs, which mix lots of things together, and also get upgraded over time with new major changes, i.e., which are a shifting target. And this PR has been that, from my perspective. There are multiple changes in it I'd have merged weeks ago if they had been in their own PR, but there were new things being added, and it was not clear to me when the PR was supposed to "ready for review and merge".

To work with PRs, the ideal PRs (from the perspective of a reviewer, not necessarily from that of a submitter ;-) ) are those which are focused on one subject and don't change once submitted (other than to address review comments and perhaps bugs found by the submitter). Or alternatively PRs which contains multiple disparate minor changes (bug fixes and so on), but not too many, and each of them not too big.

So if you are willing to try smaller, more focused PRs, that'd be great. Perhaps we could also work out something with your current "additions" and "fixes" branches, though the best I can think of is somebody else doing it for you (i.e. you make a big PR, somebody else cherry-picks parts into separat PRs, those get reviewed and merged, you rebase, rince and repeat). But then somebody has to have the time and energy for that, adding a new bottleneck...

I kind of wish we could discuss this in person, because I find it rather difficult to write about this clearly and w/o writing far too much which nobody has time to read anyway ;-).

Anyway, I think *this* PR is good to go, which is why I approved it prior to writing this :-).

PR #2058, opened 2018-01-01
Fix inefficiency when dealing with certain algebra modules

fingolfin:
The issue was originally reported by Rudolf Tange on the GAP support mailing list. The fix is virtually identical to a fix for a similar issue, see 7aec45fa176be87c32afd53f3cbf88f7ee5c890f


fingolfin:
Note that this PR is against `stable-4.9`, and the fix should IMHO go into 4.9.0 or 4.9.1.

PR #2059, opened 2018-01-02
Fix travis failures (semigroups 3.x requires GCC >= 5), some other travis tweaks

fingolfin:


fingolfin:
I pushed this directly to stable-4.9 now.

PR #2060, opened 2018-01-03
"FullMatrixLieAlgebra(GF(2),0)"  issue

rtange:
Max Horn suggested that I put the suggestion to implement
FullMatrixLieAlgebra(F,0) here.
The reason for this is that gl(V) also exists when V is the zero space.
For example, the zero module V of a Lie algebra L may occasionally
show up in your computations and maybe you prefer to look at
representations as Lie algebra homs L->gl(V).
Max Horn suggested that, as part of the "MatrixObj"-project
(this includes distinguishing empty matrices over different fields, and even,
should we want to, a 0xn from a nx0 matrix) , someone
might want to implement it or improve its error messages.


fingolfin:
@rtange Thanks, however, you created a "pull request" for adding new code, instead of filing an issue. You can do the later via  
   https://github.com/gap-system/gap/issues/new

rtange:
Sorry,
I have now created the issue
gap-system/gap   "FullMatrixLieAlgebra(GF(2),0)" issue
but unfortunately I couldn’t delete the “pull request”.

Do you know how to do that?

alex-konovalov:
For the record, the issue has been created in #2062. Thanks @rtange !

PR #2061, opened 2018-01-03
More optimizations for integer functions

fingolfin:
This PR speeds up `GcdInt`, `PValuation` and `RootInt`; see the respective commit message for some timings.

UPDATE: also improved `SmallestRootInt` and `IsPrimePowerInt`

### DIFF
--- a/lib/integer.gi
+++ b/lib/integer.gi
@@ -1098,13 +1098,11 @@ InstallGlobalFunction(RootInt,function ( arg )
     else Error("usage: `Root( <n> )' or `Root( <n>, <k> )'");
     fi;
 
-    # check the arguments and handle trivial cases
-    if  k <= 0                  then Error("<k> must be positive");
-    elif k = 1                  then return n;
-    elif n < 0 and k mod 2 = 0  then Error("<n> must be positive");
-    elif n < 0 and k mod 2 = 1  then return -RootInt( -n, k );
-    elif n = 0                  then return 0;
-    elif n <= k                 then return 1;
+    # ask the kernel to compute the root; this can only fail for
+    # huge values of k and enormously huge values of n
+    r := ROOT_INT(n, k);
+    if r <> fail then
+        return r;
     fi;
 
     # r is the first approximation, s the second, we need: root <= s < r
--- a/hpcgap/lib/integer.gi
+++ b/hpcgap/lib/integer.gi
@@ -1111,13 +1111,11 @@ InstallGlobalFunction(RootInt,function ( arg )
     else Error("usage: `Root( <n> )' or `Root( <n>, <k> )'");
     fi;
 
-    # check the arguments and handle trivial cases
-    if  k <= 0                  then Error("<k> must be positive");
-    elif k = 1                  then return n;
-    elif n < 0 and k mod 2 = 0  then Error("<n> must be positive");
-    elif n < 0 and k mod 2 = 1  then return -RootInt( -n, k );
-    elif n = 0                  then return 0;
-    elif n <= k                 then return 1;
+    # ask the kernel to compute the root; this can only fail for
+    # huge values of k and enormously huge values of n
+    r := ROOT_INT(n, k);
+    if r <> fail then
+        return r;
     fi;
 
     # r is the first approximation, s the second, we need: root <= s < r
--- a/hpcgap/lib/primality.gi
+++ b/hpcgap/lib/primality.gi
@@ -578,8 +578,8 @@ function(N)
   # try straightforward method first
   if cheap=true and FactIntPartial=true then
     to_check:=Concatenation(
-      List(Set(PartialFactorization(N-1,7)),p->["F",p]),
-      List(Set(PartialFactorization(N+1,7)),p->["L",p]));
+      List(Set(PartialFactorization(N-1,7 : cheap)),p->["F",p]),
+      List(Set(PartialFactorization(N+1,7 : cheap)),p->["L",p]));
     if [] <> PrimalityProof_VerifyStructure(N,to_check)
     then return to_check;
     else Info(InfoPrimeInt,1,"Straightforward Fermat-Lucas primality proof failed on ",N);
--- a/lib/primality.gi
+++ b/lib/primality.gi
@@ -578,8 +578,8 @@ function(N)
   # try straightforward method first
   if cheap=true and FactIntPartial=true then
     to_check:=Concatenation(
-      List(Set(PartialFactorization(N-1,7)),p->["F",p]),
-      List(Set(PartialFactorization(N+1,7)),p->["L",p]));
+      List(Set(PartialFactorization(N-1,7 : cheap)),p->["F",p]),
+      List(Set(PartialFactorization(N+1,7 : cheap)),p->["L",p]));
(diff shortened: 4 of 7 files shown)


frankluebeck:
This looks essentially good. But it breaks:
```
IsPrimeInt(100000000000000000000000000000000000000000000000151);
NextPrimeInt(10^50);
```

fingolfin:
@frankluebeck could you please elaborate what is broken? I get the same output on this branch, master, and various older GAP versions, namely:
```
gap> IsPrimeInt(100000000000000000000000000000000000000000000000151);
true
gap> NextPrimeInt(10^50);
100000000000000000000000000000000000000000000000151
```

UPDATE: ah OK, if I load `factint` (which I do not do by default), I get an error. That's because `factint` abuses `RootInt to take roots of rationals.
```
gap> IsPrimeInt(100000000000000000000000000000000000000000000000151);
Error, Root: <n> must be an integer (not a rational) in
  r := ROOT_INT( n, k ); at /Users/mhorn/Projekte/GAP/gap.spielwiese/lib/integer.gi:1263 called from
RootInt( 2 ^ (LogInt( n, 10 ) - 50), 4 ) at /Users/mhorn/Projekte/GAP/gap.spielwiese/pkg/factint-1.5.4/lib/ecm.gi:323 called from
...
brk> n;
1/32
brk> k;
4
```

With the old code, we get this weird result (for unsupported inputs, of course, so there is nothing wrong with it per se)
```
gap> RootInt(1/32,4);
1
```

Ideally, `factint` should be fixed not to do that. Maybe @Stefan-Kohl can comment? Is this computation an accident, or is there really need for taking roots of rationals? 

In the meantime, I could add a workaround for backwards compatibility, I'll just have to figure out what it ought to do.

(Finally, there is an undocumented operation `Root`, but so far its only installed method calls `RootInt`)

fingolfin:
This patch for `FactInt` fixes the problem:
```
diff --git a/lib/ecm.gi b/lib/ecm.gi
index 80854ba..f41cd5b 100644
--- a/lib/ecm.gi
+++ b/lib/ecm.gi
@@ -319,8 +319,8 @@ function ( arg )
   if ArgCorrect then
     n      := GetArg(1,"DoesNotExist",fail);
     Curves := GetArg(2,"ECMCurves",
-                     n -> Maximum(4,3 * RootInt(2^(LogInt(n,10) - 24),8)
-                                      + RootInt(2^(LogInt(n,10) - 50),4)));
+                     n -> Maximum(4,3 * RootInt(Int(2^(LogInt(n,10) - 24)),8)
+                                      + RootInt(Int(2^(LogInt(n,10) - 50)),4)));
     if   IsInt(Curves)
     then NumberOfCurves := Curves; Curves := n -> NumberOfCurves; fi;
     if Curves(n) <= 0 then return [[],[n]]; fi;
```

There is a reason I don't load `factint` by default: primality testing seems to be much slower with it in cases relevant to me. E.g. the example given by Frank takes 22 milliseconds with me *without• FactInt; with FactInt, it's ~5 seconds.

fingolfin:
I made a pull request for FactInt to correct the issue there, see https://github.com/gap-packages/FactInt/pull/6

Moreover, I added a new commit to this PR which tweaks the primality prover, which (mostly) avoids the slowdown with FactInt loaded for me, and by coincidence (?) also the problem with RootInt.

Stefan-Kohl:
I silently assumed that RootInt does something sensible if the first argument is a rational. If that is not the case, the argument needs to be converted to an integer first, like in your fix. Thanks for pointing this out.

fingolfin:
@Stefan-Kohl `RootInt` never did anything sensible if the first argument was not an integer. Indeed, it even prints out an error message for non-integers -- but there was a bug in it, which caused it to silently accept any rational between 0 and 1 -- they were then treated as 1, i.e. `RootInt(1/32,4)` gave 1, which I wouldn't call "sensible" (0 would be)

@ChrisJefferson I am not sure I understand -- for what do you want additional tests?

BTW, this PR also has a test failure, which I don't understand right now: It seems some `Random` call now returns something different than before this PR. I am not yet sure why, though my suspicion is that factoring integers may invoke `Random` at some point, leading to the (harmless) test regression. I'll investigate next time I touch this PR (i.e. likely after @ChrisJefferson clarified and I then found some time to address his remarks)

lib/integer.gi
```diff
@@ -1215,13 +1258,11 @@ InstallGlobalFunction(RootInt,function ( arg )
     else Error("usage: `Root( <n> )' or `Root( <n>, <k> )'");
     fi;
 
-    # check the arguments and handle trivial cases
-    if  k <= 0                  then Error("<k> must be positive");
-    elif k = 1                  then return n;
-    elif n < 0 and k mod 2 = 0  then Error("<n> must be positive");
-    elif n < 0 and k mod 2 = 1  then return -RootInt( -n, k );
-    elif n = 0                  then return 0;
-    elif n <= k                 then return 1;
```
fingolfin: This is the problem with the old code which caused it to accept some non-integer inputs: suppose `n <= k` holds (and none of the prior checks either handled the input, or indirectly caused an assertion, e.g. `k mod 2` will error out if `k` is a rational with even denominator). Then it returns 1. This is only sensible if also `1 <= n` holds, which is of course automatic if `n` is an integer (due to the preceding checks), but fails for rationals (such as e.g. `n=1/32`).

Since this "support" for rational inputs clearly was accidental (it's not documented; it's wrong for certain inputs; most rational inputs are not supported, e.g. `RootInt(4+1/2, 2);` always gave an error), I am strongly tempted to proceed with this PR as-is (targeting GAP 4.10, not 4.9), at least once a new `FactInt` version with the fix has been released.

Stefan-Kohl:
In terms of FactInt, 1 was just as sensible as 0 -- and even 2 or 3 would hardly do more harm than causing a little slow-down on average ... . On the other hand, something like e.g. 100 would cause pretty poor performance in many cases.

ChrisJefferson:
For tests, was just thinking if testing some more numbers which don't fit in a small integer, which have more factors, also big numbers with small mods, small numbers with big mods. Nothing too serious.

fingolfin:
@ChrisJefferson I have no extended the tests, which revealed a bug in the new `ROOT_INT(n,k)` if `k` was a non-immediate negative number, or if `k` is not immediate and `n=0`. So, good suggestion ;-). Found no other issues.
Also added a commit to fix the regression in the test for `Random`.

It remains to be decided whether I should add a workaround to `RootInt` to accept formally invalid calls made by `FactInt`, which used to work; or if we instead wait for a new FactInt release which fixes this. Maybe also @alex-konovalov has a thought on that?

alex-konovalov:
@fingolfin new FactInt release within a week, maybe - would that be OK then?

fingolfin:
Rebased this, now that PR #2087 has been merged. Only three commits remain in here. Still waiting for new `FactInt` release, on which @alex-konovalov is currently working

hulpke:
@fingolfin  #2087 is so far only in my work branch. Shall I put it into master?

alex-konovalov:
Added to release notes for GAP 4.10 at https://github.com/gap-system/gap/wiki/Release-notes-for-GAP-4.10 (one common entry for this PR and #2086).

PR #2039, opened 2017-12-19
Check list length in GrowPlist

ChrisJefferson:
This ensures GrowPlist checks the size we are growing the list to is a valid list size. This stops us then overflowing in ResizeBag later.

I hope this fixes #1479, but I had trouble reproducing the error originally, so it is hard for me to check.

This also adds two explicit functions to check if a C signed or unsigned integer can be turned into an immediate integer. These functions can be used more widely, but I leave that for another PR (or at least until this one is checked to see if people like them).

Issue #1479, opened 2017-07-06
Crashes from Todd Coxeter


frankluebeck:
### Observed behaviour
Calling the following code twice in the same GAP session crashes GAP with a segmentation fault (leave the break loop after the first time):
```
f := FreeGroup( 3 );;
g := f/[f.1^2,f.2^2,f.3^2,(f.1*f.2)^3,(f.1*f.3)^3,(f.2*f.3)^3];
Size(g);
```
This occurs in various versions of GAP compiled in 32-bit mode (including current master).

### Expected behaviour
The code should run into a break loop every time.

### Remark
Maybe something is not cleaned up correctly in the kernel part of the Todd-Coxeter enumeration when a calculation is interrupted?

fingolfin:
I cannot reproduce this, neither in master nor in stable.

How long do you wait (roughly) till you abort `Size(g)` ? Then, what do you do the second time -- just run `Size(g);` again, or really redo the whole computation (i.e. also redefine `f` and `g`) ?
Perhaps you could add a full log of your session (including the backtrace in the first break loop; knowing where it happened might allow inserting an `if ... Error` combo to trigger the break at the right point.

dimpase:
On 32-bit Linux I get 
```
 ┌───────┐   GAP 4.8.6, 12-Nov-2016, build of 2017-07-02 08:49:34 (BST)
 │  GAP  │   http://www.gap-system.org
 └───────┘   Architecture: i686-pc-linux-gnu-gcc-default32
 Libs used:  gmp, readline
 Loading the library and packages ...
 Packages:   GAPDoc 1.5.1
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap> f := FreeGroup( 3 );;
gap> g := f/[f.1^2,f.2^2,f.3^2,(f.1*f.2)^3,(f.1*f.3)^3,(f.2*f.3)^3];
<fp group on the generators [ f1, f2, f3 ]>
gap> Size(g);
#I  Coset table calculation failed -- trying with bigger table limit
#I  Coset table calculation failed -- trying with bigger table limit
#I  Coset table calculation failed -- trying with bigger table limit
#I  Coset table calculation failed -- trying with bigger table limit
Error, reached the pre-set memory limit
(change it with the -o command line option) in
  prev[2 * limit] := 2 * limit - 1
 ; at /home/dima/sage/dev/sage/local/gap/latest/lib/grpfp.gi:1201 called from 
[...]

PR #2064, opened 2018-01-04
Restrict capacity of plists to 2^28 resp. 2^60 (revision of #2039)

fingolfin:
This is a rebased version of PR #2039, with the fixup commit squashed, and a new size restriction assert added to `NEW_PLIST`. As such, it addresses parts of issue #1479 (but does not fully resolve it; at least one overflow bug in GASMAN needs to be fixed, as well a bug in one of the GNU libc `memmove` implementations).

I also provided a new, extensive commit message for the second commit, and since that now makes the majority of the work in that commit, took the liberty of changing authorship of that commit to myself. Here is the new commit message:

By adding bounds checks to GrowPlist and NEW_PLIST, we restrict the capacity
of plain lists to the maximum value of an immediate integer, i.e. 2^28 on a
32 bit system and 2^60 on a 64 bit system. The check in GrowPlist triggers
an error, as it can be triggered by the user. The check in NEW_PLIST is a
run-time assertion, and should never be triggered by user actions, only by
buggy kernel code.

This restriction fixes overflows and other problems, which can lead to
crashes, corrupt data or nonsense computations.

It poses no actual limitation in practice, for the following reasons:

First off, many other places already effectively limited the length of a
plist they can interact with to the maximum value of an immediate integer.
E.g. such limitations exist for sublist access via l{poss}, EmptyPlist(),
ASS_PLIST_DEFAULT, and more.

Secondly, with this change,  the effective size (in bytes) of a plist of
this maximal length would be 2^30 resp. 2^63. The latter certainly poses no
actual limitation. The former corresponds to 1 GB. Conceivably, GAP could
support slightly larger bags on a 32bit system (the GASMAN heap can grow up
to 3GB if the host system supports it). However, there is not much you can
do with GAP if most of its heap is filled with a single gigantic plist, so
this restriction seems acceptable.

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -431,9 +431,8 @@ Obj GMP_REDUCE( Obj gmp )
     return gmp;
   }
   if ( SIZE_INT(gmp) == 1) {
-    if ( ( VAL_LIMB0(gmp) < ((1L<<NR_SMALL_INT_BITS)) ) ||
-         ( IS_INTNEG(gmp) &&
-           ( VAL_LIMB0(gmp) == (1L<<NR_SMALL_INT_BITS) ) ) ) {
+    if ( ( VAL_LIMB0(gmp) <= INT_INTOBJ_MAX ) ||
+         ( IS_INTNEG(gmp) && VAL_LIMB0(gmp) == -INT_INTOBJ_MIN ) ) {
       if ( IS_INTNEG(gmp) ) {
         return INTOBJ_INT( -(Int)VAL_LIMB0(gmp) );
       }
@@ -470,9 +469,8 @@ int IS_NORMALIZED_AND_REDUCED( Obj op, const char *func, int line )
     Pr("WARNING: non-normalized gmp value (%s:%d)\n",(Int)func,line);
   }
   if ( SIZE_INT(op) == 1) {
-    if ( ( VAL_LIMB0(op) < ((1L<<NR_SMALL_INT_BITS)) ) ||
-         ( IS_INTNEG(op) &&
-           ( VAL_LIMB0(op) == (1L<<NR_SMALL_INT_BITS) ) ) ) {
+    if ( ( VAL_LIMB0(op) <= INT_INTOBJ_MAX ) ||
+         ( IS_INTNEG(op) && VAL_LIMB0(op) == -INT_INTOBJ_MIN ) ) {
       if ( IS_INTNEG(op) ) {
         Pr("WARNING: non-reduced negative gmp value (%s:%d)\n",(Int)func,line);
         return 0;
@@ -499,7 +497,7 @@ Obj ObjInt_Int( Int i )
 {
   Obj gmp;
 
-  if ( (-(1L<<NR_SMALL_INT_BITS) <= i) && (i < 1L<<NR_SMALL_INT_BITS )) {
+  if (INT_INTOBJ_MIN <= i && i <= INT_INTOBJ_MAX) {
     return INTOBJ_INT(i);
   }
   else if (i < 0 ) {
@@ -516,9 +514,7 @@ Obj ObjInt_Int( Int i )
 Obj ObjInt_UInt( UInt i )
 {
   Obj gmp;
-  UInt bound = 1UL << NR_SMALL_INT_BITS;
-
-  if (i < bound) {
+  if (i <= INT_INTOBJ_MAX) {
     return INTOBJ_INT(i);
   }
   else {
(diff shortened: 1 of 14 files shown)


src/plist.c
```diff
@@ -67,12 +67,18 @@ Int             GrowPlist (
     UInt                plen;           /* new physical length             */
     UInt                good;           /* good new physical length        */
 
+    if (!UINT_FITS_IN_INTOBJ(need)) {
+        ErrorMayQuit("GrowPlist: List size too large", 0, 0);
+    }
+
     /* find out how large the plain list should become                     */
     good = 5 * (SIZE_OBJ(list)/sizeof(Obj)-1) / 4 + 4;
```
frankluebeck: I would check the overflow here:

```
    if (! UINT_FITS_IN_INTOBJ(good))
        good = (((UInt)1) << NR_SMALL_INT_BITS) - 1;
```

src/plist.c
```diff
@@ -67,12 +67,18 @@ Int             GrowPlist (
...
 
     /* but maybe we need more                                              */
-    if ( need < good ) { plen = good; }
-    else               { plen = need; }
+    if (need < good && UINT_FITS_IN_INTOBJ(good))
```
frankluebeck: Then this can be simplified to 
```
    if (need < good)
```
In your version adding and appending to lists of length > 4/5 * maximal length becomes very inefficient.

fingolfin: Excellent point, thanks, I'll update this PR.

frankluebeck:
I always thought that the length of plists is restricted to 2^28-1 (resp. 2^60-1). I agree that for the reasons given above this is not a serious restriction. I guess a lot of code does actually assume that the length of a plist fits into an immediate integer. Therefore I consider this pull request as a sensible bug fix.

Maybe this restriction should be mentioned somewhere in the manual?

A possible test on 32 bit systems could be
```
gap> l:=0*[1..QuoInt(2^28*6,7)];;
gap> Append(l,0+[1..100]);
gap> m:=MemoryUsage(l);;
gap> m>2^30+16;
false
```

(Or use `SIZE_OBJ` after it is fixed to work with large objects.)

Side remark:
> (the GASMAN heap can grow up to 3GB if the host system supports it)

On my 64 bit systems I can even call 32-bit gap with `gap -m 3.9g`.

fingolfin:
I'll look into addressing all of that. I'll also replace `SIZE_OBJ` by an alias for `SHALLOW_SIZE` (which already is correct and doesn't crash for immediate values as input)

fingolfin:
I have now addressed most of the points @frankluebeck brought up, I hope, plus some more:
* check in `GrowPlist` adjusted as suggested by Frank;
* fixed `SIZE_OBJ` to return sane values for large objects, and to not crash when called on an immediate object it (and while at it, merged it with `SHALLOW_OBJ`, which now is an obsolete synonym)
* replaced `INT_FITS_IN_INTOBJ` and `UINT_FITS_IN_INTOBJ` (which I found ungainly) by constants for the min/max allowed values; I also made a lot of code use them; I find the resulting checks easier to understand;
* 

I also looked into adding the length restriction to the manual, but couldn't find a good place. We already mention a similar restriction for ranges (were the two ends must be immediate integers). But I couldn't find any section dedicated to plain lists / "internally represented lists", although they are mentioned on and off. Perhaps such a section should be added? Or perhaps it's there and I just missed it?

I also noticed that section 21.2 "Basic Operations for Lists" starts out saying "The basic operations for lists are ..." and then provides a set of points which, curiously enough, does not cover *creation* of lists. I found no place in the whole manual where this is described (I focused on checking chapter 21, as well as chapter 4, "The Programming Language")

src/integer.c
```diff
@@ -516,9 +514,7 @@ Obj ObjInt_Int( Int i )
 Obj ObjInt_UInt( UInt i )
 {
   Obj gmp;
-  UInt bound = 1UL << NR_SMALL_INT_BITS;
-
-  if (i < bound) {
+  if (i <= INT_INTOBJ_MAX) {
```
ChrisJefferson: This is relying on how C handles signed / unsigned comparisons. I can't remember his they come out, and would have a small preference for assigning INT_INTOBJ_MAX to a UInt, just in case future people worry.

fingolfin: I briefly thought about adding an `UINT_INTOBJ_MAX`, but the C standard is pretty clear on this, and always has been. In a (simplified) nutshell, a signed and unsigned value are compared by casting the signed value to unsigned, which is always what we want for these bounds checks.

In short, `INT_INTOBJ_MAX` and `UINT_INTOBJ_MAX` would always do exactly the same thing, *except* when accidentally using `UINT_INTOBJ_MAX` to bound check a signed quantity... As such, `UINT_INTOBJ_MAX` could be misused by accident. I don't see a way to misuse `INT_INTOBJ_MAX` by accident.

alex-konovalov:
@fingolfin I will omit this from release notes: as #2039 sats, it ensures that GrowPlist checks the size we are growing the list to is a valid list size, and stops us then overflowing in ResizeBag later. This is an internal functionality.

PR #2068, opened 2018-01-04
Various minor kernel tweaks

### LINKED ISSUE #2065: Remove ReadSmall, RereadSmall, etc.?
The library still declares these:
```gap
BIND_GLOBAL("ReadSmall",ReadAndCheckFunc("small"));
BIND_GLOBAL("ReadTrans",ReadAndCheckFunc("trans"));
BIND_GLOBAL("ReadPrim",ReadAndCheckFunc("prim"));
```
and these
```gap
BIND_GLOBAL("RereadSmall",RereadAndCheckFunc("small"));
BIND_GLOBAL("RereadPrim",RereadAndCheckFunc("prim"));
BIND_GLOBAL("RereadTrans",RereadAndCheckFunc("trans"));
```

Shouldn't they be removed?

fingolfin:
This is a grab bag of various minor changes to the kernel.
  
UPDATE: I decided to modify this PR to only contain kernel changes; to this end, I removed all library related commits, and moved a few commits from PR #2067 which were not related to GASMAN to this PR. I hope it's sufficiently easy to review now.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -2063,6 +2063,9 @@ static Int InitKernel (
     InfoBags[ T_FUNCTION ].name = "function";
     InitMarkFuncBags( T_FUNCTION , MarkAllSubBags );
 
+    /* Allocate functions in the public region */
+    MakeBagTypePublic(T_FUNCTION);
+
     /* install the type functions                                          */
     ImportGVarFromLibrary( "TYPE_FUNCTION",  &TYPE_FUNCTION  );
     ImportGVarFromLibrary( "TYPE_OPERATION", &TYPE_OPERATION );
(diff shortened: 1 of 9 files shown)


ChrisJefferson:
I'm happy with this, but as a general comment I'd prefer to replace all memcpys with memmoves, I've never found any significant speed increase from memcpy, and it just seems simpler to me, rather than figuring out if memcpy is useful.

I'm happy for you to use memcpy, but I warn you I'll probably always memmove :)

fingolfin:
The main motivation to replace memcpy by memmove in this PR is not a concern for speed, but rather due to the fact that `memmove` is buggy in glibc. Now, granted, using the buggy `memmove` shouldn't be worse than using `memcpy` on overlapping memory. But with the changes in this PR, it becomes a tad easier to identify all places where we use `memmove`, and then use that information to decide how we want to deal with the `memmove` bug in glibc.

PR #2069, opened 2018-01-05
Add immediate Dimension method for trivial modules

fingolfin:
Also adjust a LeftModuleByGenerators method to set IsTrivial suitably.


fingolfin:
Update: this now also removes the diffs between the GAP and HPC-GAP versions of `lib/vspcrow.gi`

ssiccha:
@fingolfin looks fine to me

hulpke:
From the curmudgeon who always complains about immediate methods:

Is there a crucial performance bottleneck that requires such an immediate method?
Could it be achieved alternatively to simply supplement the `SetIsTrivial' by `SetDimension'?

fingolfin:
@hulpke it's good you are keeping us/me on our toes, so I appreciate it :-)

Short answer: I do not know of any crucial performance bottleneck requiring this patch. Though now I wonder if I could construct one... but it'd be artificial, so I won't try.

OTOH, I also know of no code which is  impacted *negatively* by this, and I think it does noticeable improve the user experience if, upon encountering a trivial algebra or module, the user is immediately told that this is an algebra/module "of dimension 0" by the printing.

I am kind of a fan of these immediate implications, for me they are one of the major distinct features of GAP. If there are concerns for adding them due to possible negative speed impact, I'd rather look into making them faster, than giving up on using them.

As to adding explicit `SetDimension` calls: A quick checks shows that are about a dozen potentially relevant `SetIsTrivial` calls in the code (though of course code may also indirectly deduce `IsTrivial`). So this alternative is certainly doable; though to me, if feels like playing whack-a-mole -- new code may very well forget to add this in again; and it's a kind of code duplication. And of course any package using `SetIsTrivial` then would have to be adapted similarly (but luckily, it seems no packages does this, at least not for modules or algebras, as far as I could tell on a quick check performed right now.  BTW, we also have an immediate method going in the other direction: if the dimension is known, we set `IsTrivial`).

My main motivation for this was that I got rather inconsistent printing in tests all over the place, because in some code paths, trivial modules knew their dimensions, in others they did not. I also found the printing much more helpful -- of course that could also be addressed by checking for `IsTrivial` in the printing functions, but then one looses other benefits that may be derived from knowing the `Dimension`.

Anyway, in the end, I am not married to this PR ;-) and we can go differently about it. Would be interesting to hear what other think, too.

Whatever we do, adding `SetIsTrivial` to that one LeftModuleByGenerators method is, I think, an uncontroversial improvement.

hulpke:
@fingolfin First I sympathize with the issue of tests. This is a pain if not fixed.
In '98 I would have addressed this issue by installing a setter method for `IsTrivial' for vector spaces that also sets the dimension to 0. (One could do similar for trivial groups with 5 generators, etc.)

My concern with immediate methods (and this is based on the behavior 20 years ago) is that the system can easily cope with an individual immediate method, but if there are several thousands it slows down (whenever a type changes it needs to be checked if any immediate method now applies), in particular if immediate methods trigger new immediate methods.

I know you're just installing one, but I have seen multiple times excited new users who (with the best intentions) want to use immediate methods to make GAP do automated reasoning -- that is implement a large number of theoretical implications (such as: If a finite group is abelian, all character degrees are one) as immediate methods. If we use exactly those methods in the library it is hard to argue why "our" methods are good and theirs are not so.

frankluebeck:
I would avoid immediate methods as much as possible, even seemingly cheap ones, for the reasons explained by @hulpke.

So, in this case I would change the proposed immediate method for `Dimension` to a normal method installation. (Is the next applicable method really so expensive that it is worth to add a special method?) If the main point here is to get a nicer view of trivial modules, just add a method for `ViewObj` for trivial modules, this is completely uncritical performancewise.

ssiccha:
Is there no datastructure that gives you only those immediate methods that have a chance of being applicable?

fingolfin:
I am withdrawing this, at least for the time being.

PR #2070, opened 2018-01-05
Test: allow setting rewriteToFile to `true`

fingolfin:
Setting rewriteToFile to `true` rewrites the input file, if given by a
filename; if the input is a stream, nothing happens. This mimics what
we do in `TestDirectory`.

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -234,7 +234,8 @@ end;
 ##  <Item>If this is bound to a string it is considered as a file name
 ##  and that file is written with the same input and comment lines as
 ##  <Arg>fname</Arg> but the output substituted by the newly generated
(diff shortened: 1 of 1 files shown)


fingolfin:
Note: I based this PR on `master`, and set the 4.10 milestone, as this is strictly speaking a new few feature, and thus should not be added to the stable branch now. But OTOH it's a trivial change which fits with other changes we made and that *are* in `stable-4.9`... so a point could be made for merging this PR there.

I am 100% fine either way, just wanted to mention this for completeness.

ChrisJefferson:
I think this should be put in 4.9, and I'd be happy to merge it if you point it there.

fingolfin:
Now rebased

PR #2073, opened 2018-01-09
More HPC-GAP syncing

fingolfin:
Amongst other things, this unifies the code in `lib/random.g`. But I am tempted to throw away `R_X` and `R_N` altogether in GAP 4.10. They were undocumented and obsolete since at least 2010; no active package seems to use them. Removing them at worst could break some local code people have, but it'd break it in a clear way (i.e. no subtile incorrect behavior creeping in), and affected code should be straightforward to fix.

What do people thing, e.g. @frankluebeck and @stevelinton ?


lib/stbcbckt.gi
```diff
@@ -2643,8 +2643,8 @@ local Pr, div, B, rbase, data, N;
     N := ShallowCopy( G );
   fi;
   # remove cached information
-  Unbind(G!.suborbits);
-  Unbind(E!.suborbits);
+  G!.suborbits:=[];
+  E!.suborbits:=[];
```
fingolfin: Note that this is correct because of line 466: if `!.suborbits` is missing, then it is immediately set to an empty list. So we may just as well set it to an empty list right away, instead of unbinding (as becomes clear when one inspects all three places accessing it -- all three are visible in the diff of this PR.

lib/ratfun1.gi
```diff
@@ -42,9 +42,9 @@ local f,typ,lc;
     typ := rfam!.threeLaurentPolynomialTypes[1];
   fi;
   
-  # slightly better to do this after the Length has been determined 
+  # slightly better to do this after the Length has been determined
   if IsFFECollection(coeff) and IS_PLIST_REP(coeff) then
-    ConvertToVectorRep(coeff);
+    coeff := ImmutableVector(DefaultRing(coeff), coeff);
```
fingolfin: Warning: in `lib/ratfun1.g`, there is actually one other diff, which I ignored: in line 55f, this:
```gap
   f := rec(IndeterminateNumberOfUnivariateRationalFunction:=inum,
           CoefficientsOfLaurentPolynomial:=Immutable([coeff,val]));
```
is replaced by
```gap
   f := rec(IndeterminateNumberOfUnivariateRationalFunction:=inum,
           CoefficientsOfLaurentPolynomial:=MakeReadOnlyObj(CopyRegion([coeff,val])));
```
However, I could not find any example where this made a difference. Given that `[coeff,val]` is a plain list, and `coeff` usually is either a plain list, or a compressed vector, and `val` is a ring element, I kind of believe that using `Immutable` here should always be fine -- otherwise, we'd get problems in many other places. But then, somebody originally made this change, perhaps to fix some specific problem?

PR #2078, opened 2018-01-10
Refactor and fix `DeclareAttribute`

markuspf:
This is a result of trying to fix a bug in `DeclareAttribute`, namely if a user did the following

```
gap> DeclareAttribute("PlaceHolder", IsGroup);
gap> DeclareAttribute("PlaceHolder", IsMultiplicativeElement);
```

Then the second instance of `PlaceHolder` would not be stored in the global array `ATTRIBUTES`.

In the process this PR also fixes a bug with argument parsing in `DeclareAttribute` and `NewAttribute`, where if one gave `"mutable"` as third argument, the argument for rank would be ignored.

We now allow passing `true` or `false` as well as the string "mutable" as mutability flag.

### DIFF
--- a/hpcgap/lib/oper.g
+++ b/hpcgap/lib/oper.g
@@ -1104,7 +1104,7 @@ end );
 
 #############################################################################
 ##
-#F  NewAttribute( <name>, <filter>[, "mutable"][, <rank>] ) . . new attribute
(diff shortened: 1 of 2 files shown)


markuspf:
Currently there is neither updated documentation nor are there tests.

markuspf:
I discovered that, to fix this issue properly, I will have to do more things, so I will submit smaller PRs that do the refactoring and the fixes I believe are correct separately.

PR #2080, opened 2018-01-12
Refactor declare attribute

markuspf:
This is the first part of what used to be #2078.

It rearranges the argument parsing code for `DeclareAttribute` and `NewAttribute`, and addresses the issue that the rank argument would be ignored if `"mutable"` was given.

It would be good if @stevelinton could comment whether this might have been intentional (not being able to rank a tester for a mutable attribute).

### DIFF
--- a/hpcgap/lib/oper.g
+++ b/hpcgap/lib/oper.g
@@ -1179,160 +1179,149 @@ BIND_GLOBAL( "OPER_SetupAttribute", function(getter, flags, mutflag, filter, ran
           return;
       end);
 
+# construct getter, setter and tester
+BIND_GLOBAL( "NewAttribute", function ( name, filter, args... )
+    local  flags, mutflag, getter, rank;
 
-BIND_GLOBAL( "NewAttribute", function ( arg )
-    local   name, filter, flags, mutflag, getter, rank;
+    rank := 1;
+    mutflag := LEN_LIST(args) >= 1 and (args[1] = "mutable" or args[1] = true);
+    if LEN_LIST(args) = 1 and IS_INT(args[1]) then
+        rank := args[1];
+    elif LEN_LIST(args) = 2 and IS_INT(args[2]) then
+        rank := args[2];
+    elif LEN_LIST(args) > 2 the// n
+//         Error("Usage: NewAttribute( <name>, <filter>, [<mutable>, <rank>] )");
+//     fi;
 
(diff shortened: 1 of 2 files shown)


hpcgap/lib/oper.g
```diff
@@ -1179,37 +1183,34 @@ BIND_GLOBAL( "OPER_SetupAttribute", function(getter, flags, mutflag, filter, ran
...
-    name   := arg[1];
-    filter := arg[2];
+    mutflag := LEN_LIST(args) >= 1 and (args[1] = "mutable" or args[1] = true);
+    if LEN_LIST(args) = 1 and IS_INT(args[1]) then
+        rank := args[1];
+    elif LEN_LIST(args) = 2 and IS_INT(args[2]) then
+        rank := args[2];
+    else
+        rank := 1;
+    fi;
```
fingolfin: Perhaps we should also reject invalid arguments? E.g.:
```gap
  mutflag:=false;
  rank:= 1;
  if LEN_LIST(args) = 1 and args[1] in [ true, false ] then
    mutflag := args[1];
  elif LEN_LIST(args) = 1 and IS_INT(args[1]) then
    rank := args[1];
  elif LEN_LIST(args) = 2 and args[1] in [true, false] and IS_INT(args[2]) then
    mutflag := args[1];
    rank := args[2];
  else
    Error("Usage: NewAttribute(...)");
  fi;
```

hpcgap/lib/oper.g
```diff
@@ -1179,37 +1183,34 @@ BIND_GLOBAL( "OPER_SetupAttribute", function(getter, flags, mutflag, filter, ran
...
-    if LEN_LIST(arg) = 3 and IS_INT(arg[3]) then
-        rank := arg[3];
-    else
-        rank := 1;
-    fi;
     STORE_OPER_FLAGS(getter, [ flags ]);
 
     atomic FILTER_REGION do
-    OPER_SetupAttribute(getter, flags, mutflag, filter, rank, name);
+        OPER_SetupAttribute(getter, flags, mutflag, filter, rank, name);
```
fingolfin: I deliberately removed indents like that to reduce spurious diffs without whitespace suppression. Granted, in this particular case it matters little, but for consistency, I'd like to keep it as it was.

markuspf:
If I just end up copy-and-pasting code you write into my PRs, I suggest you just write

PR #1707, opened 2017-09-12
Store declaration filters of operations in record

frankluebeck:
This substitutes a list OPERATIONS containing operations in odd positions
and corresponding lists of flag lists from declarations in the following
positions by:
- a list OPERATIONS, containing only the operations themselves and
- a record OPER_FLAGS containing the lists of flag lists.

More precisely, a pair [oper,flaglists] is stored in component
HANDLE_OBJ(oper) of OPER_FLAGS.
We need a backpointer to oper here because HANDLE_OBJ(oper) will
change after reloading a workspace. With the backpointer we can
clean this up in a post-restore function.

On my computer the startup time of GAP is reduced from about 1460msec
to 1260msec.

PR #2081, opened 2018-01-12
Fix bug in INSTALL_METHOD

markuspf:
GET_OPER_FLAGS returns `fail` not `false` if the operation is not known.

### DIFF
--- a/hpcgap/lib/oper1.g
+++ b/hpcgap/lib/oper1.g
@@ -462,7 +462,7 @@ BIND_GLOBAL( "INSTALL_METHOD",
 
       # find the operation
       req := GET_OPER_FLAGS(opr);
-      if req = false  then
+      if req = fail  then
         Error( "unknown operation ", NAME_FUNC(opr) );
       fi;
 
--- a/hpcgap/src/c_oper1.c
+++ b/hpcgap/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include <src/compiled.h>
-#define FILE_CRC  "81317498"
+#define FILE_CRC  "102816013"
 
 /* global variables used in handlers */
 static GVar G_REREADING;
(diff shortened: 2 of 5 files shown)


fingolfin:
Shouldn't this go to stable-4.9?

hpcgap/lib/oper1.g
```diff
@@ -462,7 +462,7 @@ BIND_GLOBAL( "INSTALL_METHOD",
 
       # find the operation
       req := GET_OPER_FLAGS(opr);
-      if req = false  then
+      if req = fail  then
```
fingolfin: Good catch, this slipped with b1a0e86e43e11bab71d2aef89fb54529f2fe4f24 from PR #1707 by @frankluebeck and I failed to notice it during review :-(

frankluebeck:
Thanks for catching this. Of course, this is only detected when `InstallMethod` is called with invalid arguments.  Would it be useful to add a test like this:
```
gap> InstallMethod(Integers, [IsInt], function(n) return n; end);
Error, <opr> is not an operation
gap> InstallMethod(Size, [IsInt], function(n) return n; end);
Error, required filters [ "IsInt", "IsRat", "IsCyc", "IsExtAElement", 
  "IsNearAdditiveElement", "IsNearAdditiveElementWithZero", 
  "IsNearAdditiveElementWithInverse", "IsAdditiveElement", "IsExtLElement", 
  "IsExtRElement", "IsMultiplicativeElement", "IsMultiplicativeElementWithOne"
    , "IsMultiplicativeElementWithInverse", "IsZDFRE", "IsAssociativeElement",
  "IsAdditivelyCommutativeElement", "IsCommutativeElement", "IsCyclotomic" ]
for 1st argument do not match a declaration of Size
```

fingolfin:
This PR updates `oper1.g` which is compiled. Thus `src/c_oper1.c` and `hpcap/src/c_oper1.c` must be updated, too (by running `make docomp` twice in a GAP and a HPC-GAP build).

markuspf:
Cleaned up, test added. 

Slightly unrelated note: there are more tests coming for the whole `Declare`, `New`, `Install` machinery for operations/methods.

I have no idea why it rearranges `c_type1.c` in `make docomp` for HPC-GAP.

PR #2082, opened 2018-01-12
Allow additional data to be stored for operations, and add documentation for OPERATIONS and OPER_FLAGS

markuspf:
This PR first renames `OPER_FLAGS` to `OPER_DATA`. This is because I would like to store additional information about operations at a later stage (such as mutability for attributes), and `OPER_DATA` seems a more fitting name.

It then introduces an optional argument to `STORE_OPER_DATA` which is stored alongside the flags and the location info.

I debated multiple slightly different options for this storage, as currently of course the optional argument is always the empty list, and hence only adds storage overhead.

I tried storing all information as a list of records (this makes startup times quite slow), but storing just the additional information (location plus optional information) as a list of records, not binding the component for the optional argument might work better than what there is now. Of course providing an API to these structures makes it easy to hide these details from the user (being the GAP systems programmer in this case).

This PR also makes an attempt to update the documentation for `OPERATIONS` and `OPER_DATA` as well as the functions that access these lists.

### DIFF
--- a/hpcgap/lib/oper.g
+++ b/hpcgap/lib/oper.g
@@ -155,18 +155,18 @@ BIND_GLOBAL( "NUMBERS_PROPERTY_GETTERS", [] );
 ##
 OPERATIONS_REGION := ShareSpecialObj("OPERATIONS_REGION");  # FIXME: remove
 BIND_GLOBAL( "OPERATIONS", MakeStrictWriteOnceAtomic([]) );
-BIND_GLOBAL( "OPER_FLAGS", MakeStrictWriteOnceAtomic(rec()) );
-BIND_GLOBAL( "STORE_OPER_FLAGS",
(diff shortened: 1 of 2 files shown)


fingolfin:
> This is because I would like to store additional information about operations at a later stage (such as mutability for attributes)
For that, wouldn't be it more efficient and much easier to add a kernel function `IS_MUTABLE_ATTRIBUTE` like this:
```C
Obj FuncIS_MUTABLE_ATTRIBUTE(Obj self, Obj oper)
{
    if (IS_OPERATION(oper) && HDLR_FUNC(oper,1) == DoMutableAttribute)
        return True;
    return False;
}
```
Perhaps I am missing something? Also, are there other use cases you envision?

fingolfin:
Well, OK, it's slightly more complicated, one also has to check for `DoVerboseMutableAttribute` to deal with tracing, but that's a minor complication

PR #2031, opened 2017-12-18
Cleanup module interface of io.c, move Input/Output parts of GAPState into IO module state

fingolfin:
This PR adds various accessor functions for stuff in io.c. It also unifies the handling of the "stack of input/output files" between GAP and HPC-GAP.

The end goal here is to move all `Input*` and `Output*` members of `GAPState` into a module state struct specific to `io.c`. With this PR, we are pretty close; indeed, only `STATE(Input)` is still referenced by a file other than `io.c`, namely by `code.c`, for `SetupGapname`. There are various ways to deal with that; the simplest (and so maybe most pragmatic) would be to expose the `gapnameid` field via an accessor function. But I am dissatisfied with this whole `gapnameid` thing all in all, so I'd like to think a bit more about it (e.g. we could replace with `RNams`, but I am worried about the performance impact on profiling, which is already rather slow).

Resolves #1612

PR #2088, opened 2018-01-14
Move Input/Output parts of GAPState into IO module state

fingolfin:
This contains the parts of PR #2031 which rely on the "RNam" hack that was deemed problematic in that PR. Hence this should not (yet) be merged before that is resolved.

One way to do that would be simply keep the `gapnameid` as it is right now, but add another global accessor function to query it. Like `GetFilenameIdOfCurrentInput()` or so.

OTOH, computing the filenameid from a filename string is not *that* expensive, and only has to be done once for each function which gets coded, of which there are only a few 10,000 when loading GAP. So, it would be perfectly reasonable to have recompute it each time in `SetupGapname()`. That would allow completely decoupling the io module from the code mapping filenames to ids.

That code then in turn could either be implemented as it is right now (with a linear search), plus some improvements to get the HPC-GAP version in sync with the GAP version (that could be achieved via some simple locking, I think; as we load the library in a single thread, that shouldn't be problematic at all).

Or we could use a simply hash table. In fact, we already have two (very similar) hash tables mapping strings to something: in `src/gvars.c` and in `src/records.c`. I already thought about factoring that out before; if that was done, then the filename cache could be a third place it is used.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -98,23 +98,25 @@ static inline void PopLoopNesting( void ) {
 
 static inline UInt SetupGapname(void)
 {
-    if (STATE(Input)->gapnameid == 0) {
+    UInt gapnameid = GetInputFilenameID();
+    if (gapnameid == 0) {
         Obj filename = MakeImmString(GetInputFilename());
 #ifdef HPCGAP
         // TODO/FIXME: adjust this code to work more like the corresponding
         // code below for GAP?!?
-        STATE(Input)->gapnameid = AddAList(FilenameCache, filename);
+        gapnameid = AddAList(FilenameCache, filename);
 #else
         Obj pos = POS_LIST(FilenameCache, filename, INTOBJ_INT(1));
         if (pos == Fail) {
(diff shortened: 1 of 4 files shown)


fingolfin:
I have no changed this to use two new helpers `GetInputFilenameID` and `SetInputFilenameID`, i.e. `gapnameid` is kept for now. I'd still like to either get rid of it, or change how we deal with it in the long run (e.g. by introducing a more generic string-hashmap type on the kernel level, which takes the common code between records and gvars, and turns it into a thread safe hashmap with string keys that we can then use to implement the filename cache efficiently, without having to store fileids inside `TypInputFile`).

*But* I don't know when or if this will ever happen, and I think modularizing the IO state is useful enough by itself, and I don't want to risk this PR bitrotting. So I'd rather merge this now, and get back to `gapnameid` another day.

markuspf:
I'd like this to come in soonish, so I can adapt my IO PR (for libgap, jupyter and other "goodness") and finally open it.

fingolfin:
Hopefully @ChrisJefferson can review and merge it ASAP then :-)

PR #2091, opened 2018-01-15
Refactor declare attribute

markuspf:
This PR rearranges the argument parsing code for DeclareAttribute and NewAttribute, and addresses the issue that the rank argument would be ignored if "mutable" was given.

It would be good if @stevelinton could comment whether this might have been intentional (not being able to rank a tester for a mutable attribute).

Most of `oper.g` (`DeclareProperty`, `DeclareConstructor`,...) could do with a similar treatment.

(This PR includes some adjustments made responding to comments on #2080 by @fingolfin)

### DIFF
--- a/hpcgap/lib/oper.g
+++ b/hpcgap/lib/oper.g
@@ -1095,7 +1095,7 @@ end );
 
 #############################################################################
 ##
-#F  NewAttribute( <name>, <filter>[, "mutable"][, <rank>] ) . . new attribute
+#F  NewAttribute( <name>, <filter>[, <mutable>][, <rank>] ) . . new attribute
 ##
 ##  <#GAPDoc Label="NewAttribute">
 ##  <ManSection>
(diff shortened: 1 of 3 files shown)


stevelinton:
I can't think of any reason for this being intentional. I can quite imagine wanting to rank up something like HasMutableStabChain. It makes less sense for something like HasKnownSylowSubgroups (or whatever it's called) because the filter doesn't tell you WHICH Sylow subgroups it has.

markuspf:
I am also getting the urge to factor the argument checking code into a separate function for `DeclareAttribute` and `NewAttribute`, since it is doing the same checks, and it was already a pain to keep the four copies in sync.

hpcgap/lib/oper.g
```diff
@@ -1179,33 +1186,35 @@ BIND_GLOBAL( "OPER_SetupAttribute", function(getter, flags, mutflag, filter, ran
...
+    rank := 1;
+    mutflag := false;
+    if LEN_LIST(args) = 1 and args[1] in [ "mutable", true ] then
+        mutflag := true;
+    elif LEN_LIST(args) = 1 and IS_INT(args[1]) then
+        rank := args[1];
+    elif LEN_LIST(args) = 2 and args[1] in [ "mutable", true ] and IS_INT(args[2]) then
+        mutflag := true;
+        rank := args[2];
+    elif LEN_LIST(args) > 2 then
```
ChrisJefferson: I think the logic is wrong here.. I think line 1202 should just be "else", otherwise if you get one or two arguments that don't match one of the earlier if/elifs, we silently accept them and do nothing?

hpcgap/lib/oper.g
```diff
@@ -1237,102 +1246,95 @@ end );
...
```
ChrisJefferson: Same problem as mentioned above

lib/oper.g
```diff
@@ -1148,33 +1155,35 @@ BIND_GLOBAL( "OPER_SetupAttribute", function(getter, flags, mutflag, filter, ran
...
+    rank := 1;
+    mutflag := false;
+    if LEN_LIST(args) = 1 and args[1] in [ "mutable", true ] then
+        mutflag := true;
+    elif LEN_LIST(args) = 1 and IS_INT(args[1]) then
+        rank := args[1];
+    elif LEN_LIST(args) = 2 and args[1] in [ "mutable", true ] and IS_INT(args[2]) then
+        mutflag := true;
+        rank := args[2];
+    elif LEN_LIST(args) > 2 then
```
ChrisJefferson: same issue as above

lib/oper.g
```diff
@@ -1204,100 +1213,93 @@ end );
...
```
ChrisJefferson: same issue as above

markuspf:
Argument checking code rewritten again.

I also added tests to convince myself I cover all cases.

alex-konovalov:
Added to release notes for GAP 4.10 at https://github.com/gap-system/gap/wiki/Release-notes-for-GAP-4.10

PR #2092, opened 2018-01-17
Alternate GC hooking into Julia GC.

rbehrends:
This is the GAP side aspects of the work for making the Julia GC work with GAP.

The modified version of Julia with conservative stack scanning lives in the `rb/gc-extensions` branch at https://github.com/rbehrends/julia and is tracked via [this pull request](https://github.com/oscar-system/julia/pull/1).

To do list for the integration:

### Checklist

- [x] Write generic alternate GC implementation.
- [x] Modify build system to allow for enabling it from configure.
- [x] Support conservative stackscanning in Julia.
- [x] Make that GC suitable for inclusion in Julia.
  Subgoals:
    - [x] Refactor API.
    - [x] Separate out code into separate files, with only hooks remaining.
    - [x] Rewrite treaps for pointers to live outside `bigval_t` struct.
    - [x] `bigval_t` allocation hooks.
    - [x] Allow hooking custom root scanners.
    - [x] Support (conservative) tracing of task stacks.
    - [x] Public interface for custom mark/sweep functions.
    - [x] Export necessary functionality from Julia's `src/gc.{h,c}`.
    - [x] Option: support a `jl_value_from_interior_ptr(void *p)` directly?
    - [ ] Performance regression tests: ensure Julia is not affected.
    - [ ] Functionality regression tests: ensure Julia changes don't break
          the API.
- [ ] Fix remaining bugs and add missing functionality in GC/Julia interop.
    Subgoals:
    - [x] Non-hackish mark routines to support tracing GAP objects.
    - [x] Under rare circumstances, small blocks aren't identified properly
          through interior pointers.
    - [x] Look into write barrier implementation (jl_gc_wb() vs.
          jl_gc_wb_back()).
    - [x] Add support for GAP finalizers (used by various packages)
    - [x] Implement `jl_task_root_scanner_hook`
    - [ ] Add HPC-GAP support (e.g. move some globals into GAPState; possibly get rid of `JuliaTLS`; ...)
- [x] Complete hooking the Julia GC into GAP.

Building the system:

Build Julia from the `rb/gc-extensions` branch as usual, say in directory `/julia/root/dir`. Then build GAP from this PR (resp. the `alt-gc` branch) as follows (don't forget the final `usr` in the path passed to `--with-julia`):
```
./configure --with-gc=julia --with-julia=/julia/root/dir/usr
make
```

Now you should be able to start GAP. Be warned, garbage collections are currently suppressed, and if one happens anyway, it'll crash. This is now being worked on.


fingolfin:
Thanks @rbehrends ! Would you please consider rebasing this PR, to get rid of those merges? That'd make it a tad easier to look at this on a per-commit level.

rbehrends:
I can do that. There's a certain irony to the fact that I had used merges to add structure to the changes, but I agree that as GitHub does not deal with merges well, that may have inadvertently resulted in the opposite effect.

fingolfin:
To be honest, I don't see how the two merges in your PR add structure in a meaningful way... Anyway, so far this PR is pretty small, I think we are fine rebasing it.

fingolfin:
Rebased and squashed version is here: https://github.com/fingolfin/gap/tree/mh/alt-gc

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+  if (!p2)
+    p2 = treap_find(bigvals, p);
+  if (p2) {
+    jl_gc_queue_root(p2);
+  }
+}
+
+void CHANGED_BAG(Bag bag) {
+  // This is the simple Julia write barrier; jl_gc_wb() is more accurate
+  // but requires a pointer to the originating back, too, so we go with
```
fingolfin: "back" -> "bag"? Also, what does "originating" mean here? After looking at `jl_gc_wb` vs `jl_gc_wb_back`, my understanding is that the former expects a pointer to the bag that was modified (which is `bag` here), and another to the `child` bag to which a reference was inserted. So, shouldn't the comment say e.g. "... but requires a pointer to the child bag which was inserted into `bag`"?

To do that, we'd have to add a variant of `CHANGED_BAG` with two arguments, and start using this; this sounds like a bigger change. Probably shouldn't worry about this for now...

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+          InfoBags[old_type].nrLive   -= 1;
+          InfoBags[new_type].nrLive   += 1;
+          InfoBags[old_type].nrAll    -= 1;
+          InfoBags[new_type].nrAll    += 1;
+          InfoBags[old_type].sizeLive -= size;
+          InfoBags[new_type].sizeLive += size;
+          InfoBags[old_type].sizeAll  -= size;
+          InfoBags[new_type].sizeAll  += size;
+    }
+#endif
```
fingolfin: Note to self: factor out the above statistics code, too

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+
+    header->type = new_type;
+}
+
+static inline Bag AllocateMasterPointer() {
+  // HOOK: Allocate memory for the master pointer.
+  // Master pointers require one word of memory.
+  void *result = (void *) jl_gc_alloc(JuliaTLS,
+    sizeof(void *), datatype_mptr);
+  return result;
```
fingolfin: I was wondering whether it would make more sense for Boehm and Julia GC to store at lest the type and flags (but possibly also the size) inside the master point instead? This way, querying the type of an object requires on indirection less.

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+     * and the header size is a multiple of twice the word size of the
+     * architecture, then the master pointer will actually point past
+     * the allocated area. Because this would result in the object
+     * being freed prematurely, we will allocate at least one extra
+     * byte so that the master pointer actually points to within an
+     * allocated memory area.
+     */
+    if (size == 0)
+      alloc_size++;
+    /* While we use the Boehm GC without the "all interior pointers"
```
fingolfin: This comment seems to be Boehm GC  specific...

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+#ifdef TREMBLE_HEAP
+    CollectBags(0,0);
+#endif
+
+    BagHeader * header = BAG_HEADER(bag);
+
+    /* get type and old size of the bag                                    */
+    type     = header->type;
+    flags    = header->flags;
+    old_size = header->size;
```
fingolfin: Does Julia GC offer a function similar to `GC_size`? If no, perhaps one could/should add one? (Small allocations seem to come from pools, where the allocation size should thus be known; big allocations seem to store the size in struct `bigval_t` -- but of course I only scratched the surface of that code, so what I just said may very well be just nonsense)

src/julia_gc.c
```diff
@@ -0,0 +1,641 @@
...
+        }
+    }
+    /* return success                                                      */
+    return 1;
+}
+
+
+/*****************************************************************************
+** The following functions are not required respectively supported, so empty
+** implementations are provided
```
fingolfin: That's just no true ... :-)

src/julia_gc.c
```diff
@@ -456,6 +456,7 @@ UInt CollectBags (
     UInt                full )
 {
     // HOOK: perform a full collection
+    jl_gc_collect(1);
```
fingolfin: Why not `jl_gc_collect(full);`?

fingolfin:
FYI, we just updated this PR branch with a force push, so if anybody here was/is using that, you may need to take care when updating that branch. News: it was rebased, history was squashed, `src/alt_gc.c` removed, and `clang-format` run on the code. More actual work will follow.

src/julia_gc.c
```diff
@@ -0,0 +1,755 @@
...
+            alloc_size++;
+        header = AllocateBagMemory(type, alloc_size);
+
+        header->type = type;
+        header->flags = flags;
+        header->size = new_size;
+
+        // copy data and update the masterpointer
+        src = PTR_BAG(bag);
+        memcpy(DATA(header), src, old_size < new_size ? old_size : new_size);
```
fingolfin: even simpler would be to copy *everything, including the `BagHeader`, and then only update `header->size`

src/julia_gc.c
```diff
@@ -0,0 +1,755 @@
...
+    header->type = new_type;
+}
+
+static inline Bag AllocateMasterPointer(void)
+{
+    // HOOK: Allocate memory for the master pointer.
+    // Master pointers require one word of memory.
+    void * result =
+        (void *)jl_gc_alloc(JuliaTLS, sizeof(void *), datatype_mptr);
+    memset(result, 0, sizeof(void *));
```
fingolfin: This memset call is not necessary here. And at this point, perhaps we should just merge this function into `NewBag?

src/julia_gc.c
```diff
@@ -0,0 +1,755 @@
...
+    GAP_ASSERT(jl_is_datatype(datatype_bag));
+    GAP_ASSERT(jl_is_datatype(datatype_largebag));
+    GapStackBottom = stack_bottom;
+    GapStackAlign = stack_align;
+}
+
+UInt CollectBags(UInt size, UInt full)
+{
+    // HOOK: perform a full collection
+    jl_gc_collect(1);
```
fingolfin: Why not `jl_gc_collect(full);`?

src/julia_gc.c
```diff
@@ -0,0 +1,755 @@
...
+    JuliaTLS = jl_extend_get_ptls_states();
+    // jl_gc_enable(0); /// DEBUGGING
+    max_pool_obj_size = jl_extend_gc_max_pool_obj_size();
+    Module = jl_new_module(jl_symbol("ForeignGAP"));
+    datatype_mptr = jl_new_foreign_type(jl_symbol("Bag"), Module, jl_any_type,
+                                        JMarkMPtr, NULL, 1, 0);
+    datatype_bag = jl_new_foreign_type(jl_symbol("BagInner"), Module,
+                                       jl_any_type, JMarkBag, NULL, 1, 0);
+    datatype_largebag = jl_new_foreign_type(
+        jl_symbol("BagInner"), Module, jl_any_type, JMarkBag, NULL, 1, 1);
```
fingolfin: Hmm, so two datatypes both use `BagInner` as their name. Perhaps this second one should be `BagLargeInner`?

Or perhaps it should be `MasterPointer` (or `MPtr`), `Bag`, `LargeBag`, matching the C variable names?

src/boehm_gc.c
```diff
@@ -561,17 +561,10 @@ UInt ResizeBag(Bag bag, UInt new_size)
...
+        memcpy(DATA(header), src, old_size < new_size ? old_size : new_size);
         SET_PTR_BAG(bag, DATA(header));
-
-        if (DATA(header) != src) {
-            memcpy(DATA(header), src,
-                   old_size < new_size ? old_size : new_size);
-        }
-        else if (new_size < old_size) {
-            memset(DATA(header) + new_size, 0, old_size - new_size);
-        }
```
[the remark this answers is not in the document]
fingolfin: Indeed, see PR #2455

fingolfin:
I recently wondered why we need the treap code, and it seems we don't: https://github.com/rbehrends/gap/pull/9 removes it, and seems to work just fine (at least in the GAP test suite). If you know of reasons why having it is beneficial, let's discuss it on that PR.

fingolfin:
I just squashed this PR down into one commit. It'd be great if we could get it into GAP 4.10, though first we should be fairly sure that the gcext API for Julia stays stable. The plan is to submit that to the Julia folks next week, then wait for review comments and revision requests. Hopefully, this will be settled well before September (we plan to branch `stable-4.10` on September 1st, 2018). But we'll see.

src/julia_gc.c
```diff
@@ -0,0 +1,824 @@
...
+    void * result;
+    if (size <= max_pool_obj_size) {
+        result = (void *)jl_gc_alloc_typed(JuliaTLS, size, datatype_bag);
+    }
+    else {
+        result = (void *)jl_gc_alloc_typed(JuliaTLS, size, datatype_largebag);
+    }
+    memset(result, 0, size);
+    if (TabFreeFuncBags[type])
+        jl_gc_set_needs_foreign_finalizer((jl_value_t *)result);
```
fingolfin: I just realized that this is not quite enough: we also need this in `RetypeBag`. And then we may also want to be able to undo this, i.e. add a second argument `int needed` to `jl_gc_set_needs_foreign_finalizer`, and replace the above by
```C
  jl_gc_set_needs_foreign_finalizer((jl_value_t *)result), TabFreeFuncBags[type] != 0);
```
and add the same to `RetypeBag`. Thoughts?

fingolfin: Actually, I take that back: I don't think Julia allows you to remove a finalizer. But there also is no need for that: Changing an object from one that does need a finalizer to one that does not should be very rare (if it happens at all), and we can just change `JFinalizer` to not do anything in that case.

fingolfin:
During a video conference call today, we seemed to agree that we this could be merged right away (and apparently, doing so would be helpful for @frankluebeck and @ThomasBreuer).

I am fine with this, although I'd like to rebase and squash this one last time before that, i.e.: I would like to perform the merge. However, in the meantime, perhaps those people who want this in can at least formally "approve" this PR; and all others could at least briefly have a look at it, to make sure there is nothing immediately upsetting in this PR.

fingolfin:
BTW, right now, the Julia GC is *not* supported together with HPC-GAP. But this could be changed -- it just wasn't a priority for now. I think it would be useful to work towards that, but first, we should wait for the Julia changes to land there; then adapt our code here to match; and only afterwards, start looking more into HPC-GAP integration.

src/weakptr.c
```diff
@@ -796,13 +882,19 @@ void SaveWPObj( Obj wpobj )
...
   Obj *ptr;
   ptr = ADDR_OBJ(wpobj)+1;
+#endif
   len =   LoadUInt();
   STORE_LEN_WPOBJ(wpobj, len);
   for (i = 1; i <= len; i++)
     {
+#ifdef USE_JULIA_GC
+      SET_ELM_WPOBJ(wpobj, i, LoadSubObj());
+#else
```
ChrisJefferson: Very minor suggestion, I think we could just use the Julia codepath here -- it would be valid for GASMAN (I'm 99% sure). It might be slightly less efficient, but would save code seperation.

ChrisJefferson:
This is not a suggested change for this PR.

Reading the weak pointer file, the main difference between the Julia and GASMAN implementation seems to be GASMAN has to handle DEAD_WEAK_BAGS.

I think if we made GASMAN's ELM_WPOBJ return 0 if a bag for DEAD_WEAK_BAGs (and almost all, if not all, code in weakptr.c doesn't care about the difference between 0 and a DEAD_WEAK_BAG), then I think the Julia weakpointer code (which uses cfunction calls rather than reading through bags with raw C pointers) could be used for both gasman and JuliaGC removing the code differences and cleaning things up.

src/weakptr.c
fingolfin: Fine by me, will change it.

frankluebeck:
I have merged this branch into the master branch some days ago. Together with @rbehrends corresponding branch of julia this seemed to work without problems on the GAP side. 

Merging this into the master branch now would simplify the task to install a combination of GAP/Julia that can be used for further development. 

I cannot see any problem for people who want to use GAP but not the new feature provided by this pull request.

src/weakptr.c
```diff
@@ -868,6 +960,9 @@ static Int InitKernel (
     InitMarkFuncBags ( T_WPOBJ +COPYING, MarkWeakPointerObj   );
```
markuspf: Completely unrelated stupid question (which does not need to be addressed for merging): why is it possible to have `USE_GASMAN` and `USE_THREADSAFE_COPYING`?

src/julia_gc.c
```diff
@@ -0,0 +1,849 @@
...
+#include "julia.h"
+#include "julia_gcext.h"
+
+#define MARK_CACHE_BITS 16
+#define MARK_CACHE_SIZE (1 << MARK_CACHE_BITS)
+
+#define MARK_HASH(x) (FibHash((x), MARK_CACHE_BITS))
+
+// #define STAT_MARK_CACHE
+
```
markuspf: Will this become a compile-time option?

fingolfin: Good question. It's also not clear to me how useful that cache really is. Which reminds me: I asked @rbehrends before to add more comments to this file, I'll prod him again about it once this is merged, and/or add some more comments myself.

fingolfin: Actually, I also just discovered that `MarkCache` doesn't actually seem to do anything! I.e. we never use it to prevent any computation. @rbehrends did we loose some code there, or what?

Issue #2087, opened 2018-01-13
FactorsInt in library


hulpke:
This is prompted by a forum question which claimed that `AbelianInvariants`can be held back by using `FactorsInt` instead of `Factors`.

[BTW: I have not found where `AbelianInvariants` calls `FactorsInt`, only calls to `Factors`, but that is not a disproof.]

There are about 2 dozen places in the library in which `FactorsInt` is called. This being a global function, it cannot be overridden by a package.

While there might be borderline cases that really need the Rho factorization, it seems to me that the main difference is a minor speedup in that `FactorsInt` does not need to construct the default ring and redispatch (as `Factors(n)` would), but that could be resolved by another method installation for condition `IsInt`.

Is this really time critical? Is there any reason to not replace the `FactorsInt` calls in the library (short of the factorization routine) with `Factors`? (I can do so)

fingolfin:
The library method in lib/grp.gi:631 does so:
```gap
InstallMethod( AbelianInvariants,
    "generic method for groups",
    [ IsGroup ],
    function ( G )
...
    # the parent of this will be G
    cmm := DerivedSubgroup(G);
    for p  in Set( FactorsInt( Size( G ) ) )  do
...
            if r <> 1  then
                Add( ranks, Length(FactorsInt(r)) );
            fi;
...
```
Many other functions also do, e.g. this one for `Exponent`:
```gap
InstallMethod( Exponent,
    "generic method for finite groups",
    [ IsGroup and IsFinite ],
function(G)
  local exp, primes, p;
  exp := 1;
  primes := Set(FactorsInt(Size(G)));
  for p in primes do
    exp := exp * Exponent(SylowSubgroup(G, p));
  od;
  return exp;
end);
```

fingolfin:
Also note that there used to be (and perhaps still are) places  in our that called `Factors` or `FactorsInt` and shouldn't have called either -- e.g. to test whether a number is a primer power, it is *not* necessary to compute a full factorization, and `IsPrimePowerInt` resp. `SmallestRootInt` are *much* faster for this (and get even faster with PR #2061 resp. #2086).

hulpke:
@fingolfin 
Can I interpret your remarks as in favor of such a change? Unless someone objects I'll do it in a few days.

fingolfin:
Yes, sorry, I am in favor

frankluebeck:
[...]

PR #2094, opened 2018-01-18
Avoid some integer factoring; use PrimeDivisors

fingolfin:
Also remove the commented out `YndexSymmetricGroup` (it contained a call to `FactorsInt`), and replace `Length(Filtered(...))` by `Number(...)` (motivation: such code was close to code calling `FactorsInt`).

This is motivated by PR #2087. Note that `PrimeDivisors` currently calls `FactorsInt`, but now this is in a single place.

### DIFF
--- a/benchmark/transgrp/examine4
+++ b/benchmark/transgrp/examine4
@@ -2902,7 +2902,7 @@ local b,w,a,fno,cnt,normalsyl,tf,nt,tno,ntno,i,j,k,com,hom,cs,ker,nf;
   fno:=PreImage(sys.blockhom,fno); # factor normalizer
 
   # are we effectively working modulo a prime?
-  normalsyl:=Filtered(Set(Factors(Size(fno))),p->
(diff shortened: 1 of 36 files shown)


alex-konovalov:
@frankluebeck said in #2087:
> A call of Factors(Integers, x) instead of just Factors(x) can give a significant speedup if this is called frequently with small x.

Should this PR use `Factors(Integers, x)` instead of `Factors(x)` where possible then?

fingolfin:
This PR does not use `Factors`, so: no.

stevelinton:
@alex-konovalov @fingolfin 

> Should this PR use Factors(Integers, x) instead of Factors(x) where possible then?

> This PR does not use Factors, so: no.

Aside from this, admittedly pertinent,  point though, the answer is yes. If you know the ring you want to factor over, it is always better to provide it. Apart from performance concerns, GAP will sometimes not default to the ring you expect.

PR #2096, opened 2018-01-19
Fix `IsFilter(SetIsZero)` and fix a HPC-GAP access violation in InstallTrueMethodNewFilter

fingolfin:


lib/filter.g
```diff
@@ -326,7 +326,7 @@ end );
 ##
 BIND_GLOBAL( "IsFilter",
     x -> IS_IDENTICAL_OBJ(x, IS_OBJECT)
-         or ( IS_OPERATION( x )
+         or ( IS_OPERATION( x ) and NARG_FUNC( x ) = 1
```
markuspf: Could you add a comment that there are filters (namely the and-filters) that are not contained in `FILTERS`? Its just something I stumbled over for the third time today, (and I have looked at this code in depth multiple times.)

fingolfin: Will do

fingolfin: Updated. I plan some more merging and refactoring here (in particular, `IS_ELEMENTARY_FILTER` could use the same method as `IdOfFilter` to avoid a search), but I figured it's better to do that in another PR.

PR #2098, opened 2018-01-21
ObjectifyWithAttributes: return the new object

### LINKED ISSUE #2083: No return value of ObjectifyWithAttributes?
`Objectify` returns the object it has created.
I find this very convenient.
Is there any reason why `ObjectifyWithAttributes` does not return the object in question?

fingolfin:
Resolves #2083 

This change could in theory go into stable-4.9.

### DIFF
--- a/hpcgap/lib/type1.g
+++ b/hpcgap/lib/type1.g
@@ -901,6 +901,7 @@ BIND_GLOBAL( "ObjectifyWithAttributes", function (arg)
         fi;
         Setter(extra[i])(obj,extra[i+1]);
     od;
+    return obj;
 end );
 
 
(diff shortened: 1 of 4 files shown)


alex-konovalov:
@fingolfin Agree for this to go to stable-4.9

fingolfin:
Ah, sorry, forgot to adjust the PR title: there used to be additional tweaks, but they caused some test failures I do not understand, so I ripped them out, figuring that it's better to have a minimal fix for this anyway

fingolfin:
This PR is already based (on the git side) on stable-4.9, so if people agree it should go into stable-4.9, just retarget the PR via the `Edit` button.

fingolfin:
Also cherry-picked this into `stable-4.9`

PR #2101, opened 2018-01-22
HPC-GAP: fully sync lib/type1.g and lib/oper1.g (plus some gac tweaks)

fingolfin:
This fully syncs `lib/type1.g`, by (a) using `if IsHPCGAP` liberally, and (b) tolerating that a few `atomic` statements are copied to the GAP library. The plan is that we'll use those `atomic` statements in a small handful of core GAP library files, and always in a way were for plain GAP, they can just be dropped. With PR #2100, those `atomic` also have zero performance impact on plain GAP.

To make the sync more elegant, I added code to the GAP compiler to optimize `not true` to `false` and vice versa, so that `if not IsHPCGAP then ... `can be optimized away. I also changed the compiler to emit nothing for `T_EMPTY` statements, purely for cosmetic reasons (as we now emit a few empty statements, for `if` statements that were omitted in the coded output, due to their condition being always `true` resp. always `false`.

If this is merged, I also have a similar PR for `lib/oper1.g` in the pipeline.


lib/type1.g
```diff
@@ -83,37 +83,61 @@ InstallAttributeFunction(
...
+    local   lock, type, pair, family;
 
     # Look whether the category of the desired family can be improved
     # using the categories defined by 'CategoryFamily'.
     imp_filter := WITH_IMPS_FLAGS( AND_FLAGS( imp_filter, req_filter ) );
     type := Subtype( typeOfFamilies, IsAttributeStoringRep );
+    if IsHPCGAP then
+        lock := READ_LOCK(CATEGORIES_FAMILY);
+    fi;
     for pair in CATEGORIES_FAMILY do
```
ChrisJefferson: Perhaps mention in a comment here that you are manually calling READ_LOCK because the compiler doesn't support atomic? (Assuming my guess is right. If not, add comment for why we aren't using atomic here :) )

fingolfin: That's precisely why, *but* this has been in those `*1.g` for a long time, i.e. it is not something added by this PR. While I think documenting this is the right thing, I really don't want to redo this PR once again (and wait for tests etc.) just for adding a comment that somebody else should have added a couple years ago...

lib/type1.g
```diff
@@ -294,9 +333,16 @@ BIND_GLOBAL( "NEW_TYPE", function ( typeOfTypes, family, flags, data, parent )
...
     family!.nTYPES := family!.nTYPES + 1;
+    if IsHPCGAP then
+        MakeReadOnlySingleObj(type);
+        UNLOCK(lock);
+    fi;
 
     # return the type
     return type;
+    if not IsHPCGAP then
+        return lock; # HACK to silence C compiler warnings on generated code
```
fingolfin: I'd like to draw attention to this nasty HACK I added to silence warnings by the C compiler about unused variables in the C code generated by `gac`. If somebody has a better idea...

ChrisJefferson: We could make the compiler output `(void)local;` for all local variables, which is (in my experience) the "standard" way of forcing a variable to look used, but that would be a bigger change.

fingolfin: Of course adding support for `atomic` to `gac` would allow us to change this code back to using `atomic`, which would neatly avoid the need for this HACK.

ChrisJefferson: This would be easy(ish) to do, if we required a compiler with the __cleanup__ extension, which includes gcc and clang. HPC-GAP doesn't (I imagine) work on anything but a recent gcc or clang anyway.

ChrisJefferson: Or, even more shocking, make the compiler output C++ (perhaps just for HPC-GAP at first)

fingolfin: I am very reluctant to make us rely strongly on a non-standard compiler extension, and don't like the idea of locking us in to GCC and clang (sure, that looks fine now, but in the future...?) Using C++ is more attractive (for reasons discussed before), though so far nobody has invested time in making that possible (it requires some tedious work).

But anyway: All in all, correctly implementing support for `atomic` in gac using plain C is not *that* hard. I am sure I could do it in a day or two -- but I lack motivation (i.e. the gain is minimal, and HPC-GAP currently doens't work very well for me anyway).

fingolfin: Anyway, discussion of atomic support in gac should go to issue #2108

fingolfin: Anyway, to get back to the point: I could indeed emit `(void)l_unused_local`, it's a tiny change, and doesn't do much in the `src/c_*.c` files. If people are OK with it, I'll augment this PR accordingly.

fingolfin:
Now that #2103 has been merged, I have rebased this PR; modified `gac` to suppress warnings about unused local variables; added another commit which also syncs `lib/oper1.g`; and added comments which indicate why we do manual lock handling in `lib/oper1.g` and `lib/type1.g`, as suggested by @ChrisJefferson

src/c_oper1.c
```diff
@@ -3726,6 +3791,7 @@ static Obj  HdlrFunc1 (
       fi;
       narg := LEN_LIST( flags );
       methods := METHODS_OPERATION( opr, narg );
+      ;
```
fingolfin: These `;` lines come from `T_EMPTY` statements, which in turn are leftovers of `if IsHPCGAP then ... fi;` statements which got optimized away.

We could suppress this in various ways: For starters, we could change `T_EMPTY` statements to not print anything.
Cleaner would of course not to emit them in the first place, but doing this right is a bit tricky, and way beyond the scope of this PR.

ChrisJefferson:
I'm happy with this, but it's conflicting again...

fingolfin:
Rebased

PR #2102, opened 2018-01-22
Refactor src/hpc/traverse.c

fingolfin:
One motivation was to remove the need for putting knowledge about how weak pointer objects work in there, which poses a (minor) complication for PR #2092 

It also correct some quirks in the traverse code (e.g. the traversal for *immutable* lists of cyclotomics and FFEs was not being set before, leading to suboptimal performance), and opens it up for use by kernel extensions.

This code might at some point become interesting for GAP, too. In particular, I'd be curious to test performance for copying objects with it vs. with our current copying code. Of course I expect the current copying code to be faster, but by how much? If it is not too much, I'd investigate switching GAP to this alternate copying code, simply because it is much simpler (and would allow us to get rid of the `COPYING` tnum trick).

### DIFF
--- a/src/hpc/traverse.h
+++ b/src/hpc/traverse.h
@@ -38,4 +38,17 @@ Obj CopyTraversed(Obj traversed);
 int PreMakeImmutableCheck(Obj obj);
 
 
+/****************************************************************************
+**
+*F * * * * * * * * * * * * * initialize package * * * * * * * * * * * * * * *
+*/
+
+
+/****************************************************************************
+**
+*F  InitInfoTraverse() . . . . . . . . . . . . . . .  table of init functions
+*/
+StructInitInfo * InitInfoTraverse( void );
+
+
 #endif // GAP_TRAVERSE_H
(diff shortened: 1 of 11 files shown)


src/hpc/traverse.h
```diff
@@ -8,12 +8,35 @@
...
+typedef enum {
+    TRAVERSE_NONE,
+    TRAVERSE_BY_FUNCTION,
+    TRAVERSE_ALL,
+    TRAVERSE_ALL_BUT_FIRST,
+} TraversalMethodEnum;
+
+//
+// TODO
+//
```
markuspf: What is it that needs to be done?

fingolfin: Ah, I wanted to add a comment there explaining what the function does. I'll remove this here, and add that comment in a future PR (I already have a bunch of them which I wrote while trying to understand the traverse code, and fixing some of it)

fingolfin:
I took care of the `TODO` comment. As to tests: well, I made sure that we at least test the existing `weakptr.tst` (it supposedly is flaky?!? resp. depends on "compiler behavior", but so far it seemed fine for me... if it starts causing trouble at some point, we can deal with it then).

And yeah, further improvements are planned for a future PR.

PR #2113, opened 2018-01-24
correction of `IsMonomial`, mainly for reducible characters

ThomasBreuer:
correction of `IsMonomial`, mainly for reducible characters

- change the `TestMonomialQuick` and `TestMonomial` methods
  for characters such that also reducible characters are handled correctly

- restrict the `IsPrimitiveCharacter` method that relies on
  quasi-primitivity to irreducible characters

- define `IsSubnormallyMonomial` only for irreducible characters

- add tests

### DIFF
--- a/lib/ctblmono.gi
+++ b/lib/ctblmono.gi
@@ -146,7 +146,7 @@ InstallMethod( Delta,
 ##
 #M  IsBergerCondition( <chi> )  . . . . . . . . . . . . . . . for a character
 ##
-InstallOtherMethod( IsBergerCondition,
+InstallMethod( IsBergerCondition,
     "for a class function",
     [ IsClassFunction ],
     function( chi )
@@ -292,6 +292,8 @@ InstallMethod( IsBergerCondition,
 ##
 #F  TestHomogeneous( <chi>, <N> )
 ##
+##  This works also for reducible <chi>.
+##
 InstallGlobalFunction( TestHomogeneous, function( chi, N )
 
     local t,        # character table of `G'
@@ -312,6 +314,11 @@ InstallGlobalFunction( TestHomogeneous, function( chi, N )
     else
       cl:= ClassPositionsOfNormalSubgroup( UnderlyingCharacterTable( chi ),
                                            N );
+    fi;
+
+    if Length( cl ) = 1 then
+      return rec( isHomogeneous := true,
+                  comment       := "restriction to trivial subgroup" );
     fi;
 
     t:= UnderlyingCharacterTable( chi );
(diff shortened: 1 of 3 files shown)


lib/ctblmono.gi
```diff
@@ -357,6 +364,10 @@ end );
 ##
 #M  TestQuasiPrimitive( <chi> ) . . . . . . . . . . . . . . . for a character
 ##
+##  This works also for reducible <chi>.
+##  Note that a representation affording <chi> maps the centre of <chi>
+##  to scalar matrices.
```
fingolfin: What is the relevance of this comment (on the centre) at this point? To me, a comment here is for somebody who wants to call the function without caring what it does... Wouldn't this factoid about the centre be more helpful below, perhaps near the "proof" comment?

lib/ctblmono.gi
```diff
@@ -1013,6 +1015,8 @@ InstallMethod( IsMonomialNumber,
 ##
 #M  TestMonomialQuick( <chi> )  . . . . . . . . . . . . . . . for a character
 ##
+##  We may assume that <chi> is an irreducible character.
```
fingolfin: "We may assume"? This makes it sound as if it was optional, or as if there was some underlying reason... Remove "may"?, resp. say "Assumes that..." ?

Or even clarify: "This function assumes that <chi> is an irreducible character, but does not verify this. Other inputs may thus leading to nonsensical output." ?

lib/ctblmono.gi
```diff
@@ -1330,38 +1344,38 @@ InstallMethod( TestMonomial,
 
     Info( InfoMonomial, 1, "TestMonomial called" );
 
-    # Start wirth elementary tests for monomiality.
-    test:= TestMonomialQuick( chi );
+    # Start with elementary tests for monomiality.
+    if   IsIrreducibleCharacter( chi ) then
+      test:= TestMonomialQuick( chi );
+    elif HasIsMonomialCharacter( chi ) then
```
fingolfin: Should that `HasIsMonomialCharacter` perhaps be first, given that it'll be the quickest test by far?

lib/ctblmono.gi
ThomasBreuer: No.
Somebody who wants to call the function should look at the documentation
in the corresponding `.gd` file.
Comments in the implementation part are intended for somebody who wants to look at the code.

lib/ctblmono.gi
ThomasBreuer: O.k.

lib/ctblmono.gi
ThomasBreuer: O.k.

fingolfin:
Something went wrong there -- you now have two commits with identical (?) content and a merge commit in this PR. Perhaps try to `git rebase master` (if that fails with an error, you can always `git rebase --abort`)?

fingolfin:
@ThomasBreuer Unfortunately, this PR causes multiple test failures, already in `testinstall`, e.g. `tst/testinstall/ctblmono.tst` now fails. You can reproduce this locally by reading the file `tst/testinstall.g` from the GAP root directory into GAP.

fingolfin:
The tests pass now, but this PR contains duplicates of all kinds of commits on master. You'll have to `git rebase master` once again -- and make sure your `master` really is in sync with the `master` branch of `gap-system/gap`...

PR #2115, opened 2018-01-25
Change more Random() methods to accept random sources (and some related tweaks)

fingolfin:


fingolfin:
Agreed. So, I actually wanted to (a) get the full existing test suite running (and as we see, it already found failures ;-) and (b) was hoping that I could then use codecov to figure out which functions are already tested / which are not, and then provide tests for the missing ones (and see in codecov the results ;-).

fingolfin:
I have now added tests for all changed places (and found and resolved a bug in my changes ;-). Sadly, codecov is broken for us and thus does not currently post PR comments (already told them about it, apparently, some commits in our history confuse codecov ?!? they are looking into it).

Anyway, you can see the coverage for this PR at <https://codecov.io/gh/gap-system/gap/pull/2115/diff>  (but right now the PR is of course rebuilding, so the coverage report will be updated afterwards again ;-)

fingolfin:
Codecov PR comments don't work, but the coverage for this PR is 100%, as you can see on https://codecov.io/gh/gap-system/gap/pull/2115

alex-konovalov:
@fingolfin is this for release notes for GAP 4.10? If so, what about a short phrase in the description, and maybe a better title (without "tweaks")?

PR #2119, opened 2018-01-25
Adding loading packages to testpackages Travis CI test

alex-konovalov:
This pull requests extends `testpackages` test in our current Travis CI setup. 

The presently used setup only _builds_ packages (with some hardcoded exceptions, like e.g. pargap, linboxing, etc). This way, a test fails in case of any packages with failed builds. It does not actually attempt to _load_ GAP packages, neither to _run their standard tests_.

This has certain limitations. It helps to detect changes in GAP which have an impact on packages build stage, but it does not help to notice changes which may impact package loading (for example, if a package uses a function which has been renamed or deprecated in the GAP library).

Improving this by performing standard tests for packages is not currently a viable option here, because it is time consuming. Moreover, only 75 packages tested with stable-4.9 branch currently specify a short standard test in their `PackageInfo.g` file. On a positive side, those packages are tested in nightly/weekly Jenkins tests in St Andrews, and also on Travis using the nightly build Docker container for the stable-4.9 branch [here](https://travis-ci.org/gap-system/gap-docker-pkg-tests-stable-4.9) and [here](https://travis-ci.org/gap-system/gap-docker-pkg-tests-stable-4.9-staging) (although testing them prior to merging PRs would be more convenient).

What is however doable on Travis with only a small increase of testing runtime is loading **each** GAP package in a separate GAP session in order to check that each package is loadable. This would exercise not only packages having standard tests, but actually **each** GAP package (except ParGAP and XGAP that are loaded in a different way). 

This is implemented in this pull request. The new part of the `testpackages` Travis CI test first will load GAP with only packages needed for GAP, and will save a workspace to speed up loading GAP in each new GAP session, one per package. The test fails if there will be an error during loading one of packages, or if the list of packages that can not be loaded is longer than expected (current exceptions are: carat, itc, linboxing, pargap, PolymakeInterface and xgap).

This pull request is a subset of extended test of package loading that is implemented in [this PR](https://github.com/gap-system/gap-docker-stable-4.9-testsuite/pull/1) to `gap-docker-stable-4.9-testsuite`. Using Docker container, I want to run four different scenarious: loading each package individually; loading each package individually with OnlyNeeded option; loading all packages; and loading all packages with reversed option. However, I don't think that this is essential for Travis tests of the main repository: in `gap-docker-stable-4.9-testsuite` we want to check that a set of packages fits well together, so we test _interactions between packages_, while here we test that changes in the core system do not break packages - and if they do, then the first scenario should cover most of the cases.

**Remark:** Travis CI test currently fails because it can not load Digraphs in a 64-bit build. So far I haven't managed to observe this problem elsewhere: Digraphs is loadable in Jenkins tests and in Travis CI tests that use Docker.

### DIFF
--- a/etc/ci.sh
+++ b/etc/ci.sh
@@ -82,6 +82,56 @@ do
         exit 1
     else
         echo "All packages were built succesfully"
+
+        #
+        # Now that we built packages, we try to load them.
+        # For this, we need to skip a few additional packages which cannot
+        # be loaded in the test environment.
+        #
+        # skip xgap as it can only be loaded in a GAP session started via
+        # special helper script, *and* it requires X Window
+        rm -rf xgap*
+        # also skip itc because it requires xgap
+        rm -rf itc*
+
+        if [[ x"$ABI" != "x32" ]]
+        then
+          # HACK: disable some packages in 64bit mode for now. Otherwise they
+          # builds, but loading fails with an "undefined symbol: atexit" error
+          rm -rf digraphs*
+          rm -rf float*
+        fi
+
+        cd ..
+        # Load GAP (without packages) and save workspace to speed up test.
+        # Also, save names of all packages into a file to be able to iterate over them
+        $GAP -b <<GAPInput
+        SaveWorkspace("testpackagesload.wsp");
+        PrintTo("packagenames", JoinStringsWithSeparator( SortedList(RecNames( GAPInfo.PackagesInfo )),"\n") );
+        QUIT_GAP(0);
+GAPInput
+        for pkg in $(cat packagenames)
+        do
+            $GAP -q -L testpackagesload.wsp <<GAPInput
+            Print("-----------------------------------------------------\n");
+            Print("Loading $pkg ... \n");
+            if LoadPackage("$pkg",false) = true then
+              Print("PASS: $pkg\n\n");
+            else
+              Print("FAIL: $pkg\n\n");
+              AppendTo("fail.log", "Loading failed : ", "$pkg", "\n");
+            fi;
+GAPInput
+
+        done
+
+        if [[ -f fail.log ]]
+        then
+            echo "Some packages failed to load:"
+            cat fail.log
+            exit 1
+        fi
+
     fi
 
     # TODO: actually run package tests


alex-konovalov:
@james-d-mitchell the problem with Digraphs mentioned above is 
```
#W dlopen() error: /home/travis/build/gap-system/gap/pkg/digraphs-0.11.0/bin/x\
86_64-pc-linux-gnu-default64/digraphs.so: undefined symbol: atexit
Error, module '/home/travis/build/gap-system/gap/pkg/digraphs-0.11.0/bin/x86_64-pc-li\
nux-gnu-default64/digraphs.so' not found in
  if not LOAD_DYN( arg[1], false ) then
    Error( "no support for dynamic loading" );
fi; at /home/travis/build/gap-system/gap/lib/files.gd:583 
```

etc/ci.sh
```diff
@@ -82,6 +82,41 @@ do
         exit 1
     else
         echo "All packages were built succesfully"
+
+        # also skip itc because it requires xgap package
```
fingolfin: I was briefly confused by that "also". Perhaps:
```
#
# Now that we built packages, we try to load them.
# For this, we need to skip a few additional packages which cannot be loaded in the
# test environment.
#
# skip xgap as it can only be loaded in a GAP session started via  special helper script, *and* it requires X Window
rm -rf xgap*
# also skip itc because it requires xgap
rm -rf itc*
```

etc/ci.sh
```diff
@@ -82,6 +82,41 @@ do
...
+        rm -rf itc*
+        # also skip xgap because it should be loaded in other way
+        rm -rf xgap*
+
+        cd ..
+        # Load GAP (without packages) and save workspace to speed up test
+        # save names of all packages into a file to be able to iterate over them
```
fingolfin: I read these two lines as one sentence and was confused. Perhaps end the first line with a period `.`, and start the second line as `Also, save names...` ?

alex-konovalov:
@fingolfin updated comments, thanks. Now would be good to figure out what to do with digraphs to fix the build...

fingolfin:
I have no immediate idea about the `digraphs` issue, other than to suspect that perhaps the older GCC / G++ version (4.8.4) used on Travis might be part of the problem, and/or the fact that googling for the error message reveals tons of hits involving C++ code in shared libraries (as is the case here), which I suspect is related. But I cannot reproduce the issue, so I can't really help w/o investing tons of efforts, which frankly, I don't want to... so we'll have to hope @james-d-mitchell @wilfwilson etc. will have some grand insights...

In the meantime, I recommend to just skip `digraphs` in this PR (just like xgap and itc) and add it back once this problem has been resolved.

alex-konovalov:
yes... I have updated this to drop Digraphs from failing build. So after CI tests, we may merge this, and then I will submit a new PR which will add Digraphs - so I will be able to rebase it periodically and check if there is any luck.

alex-konovalov:
Removed digraphs - now same happens to the `float` package:
```
gap> #W dlopen() error: /home/travis/build/gap-system/gap/pkg/float-0.9.0/bin/x86_6\
4-pc-linux-gnu-default64/float.so: undefined symbol: atexit
Error, module '/home/travis/build/gap-system/gap/pkg/float-0.9.0/bin/x86_64-pc-linux-\
gnu-default64/float.so' not found in
  if not LOAD_DYN( arg[1], false ) then
    Error( "no support for dynamic loading" );
```

alex-konovalov:
Do not merge this - I will resubmit it to master, and will improve the output a bit.

alex-konovalov:
I've updated the output and resubmitted to master - should be ready for merge.

PR #2122, opened 2018-01-29
Get rid of Imm and MakeImm (before GAP 4.9!)

fingolfin:
These were introduced as convenience aliases for Immutable and
MakeImmutable. In practice, they are not that convenient, but instead
just add more complexity.

Luckily, they were never in a released GAP version, so let's remove them
before GAP 4.9 is released. I.e. assuming we all agree on this, we'll merge this PR and then cherry-pick it into the stable-4.9 branch.

As far as I know, no packages uses `Imm` or `MakeImm` (I just checked this in a package archive I downloaded last week). Also, in the GAP commit history, only a handful commits, all by @ChrisJefferson , ever used it. So I think this is fairly safe.

### DIFF
--- a/lib/meatauto.gi
+++ b/lib/meatauto.gi
@@ -1681,8 +1681,8 @@ end;
 
 SMTX.BasisModuleEndomorphisms:=SMTX_BasisModuleEndomorphisms;
 
-SMTX.SetBasisEndomorphismsRadical:=SMTX.Setter(MakeImm("basisEndoRad"));
-SMTX.BasisEndomorphismsRadical:=SMTX.Getter(MakeImm("basisEndoRad"));
+SMTX.SetBasisEndomorphismsRadical:=SMTX.Setter("basisEndoRad");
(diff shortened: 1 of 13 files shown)


lib/meatauto.gi
```diff
@@ -1681,8 +1681,8 @@ end;
 
 SMTX.BasisModuleEndomorphisms:=SMTX_BasisModuleEndomorphisms;
 
-SMTX.SetBasisEndomorphismsRadical:=SMTX.Setter(MakeImm("basisEndoRad"));
```
ChrisJefferson: Elsewhere you dropped the `MakeImmutable`, as you moved them in method.

ChrisJefferson:
I approve. I hoped this would save typing when we lost backtick, but I don't think it's worth it.

lib/meatauto.gi
fingolfin: Oops, that happens when you create different parts of a patch at different times. Resolved.

alex-konovalov:
Now cherry-picked on stable-4.9.

PR #2123, opened 2018-01-29
kernel: simplify ModInt/RemInt/QuoInt; some other cleanup

fingolfin:


src/integer.c
```diff
@@ -1745,12 +1745,10 @@ Obj ModInt ( Obj opL, Obj opR )
     k = INT_INTOBJ(opR);
     
     /* compute the remainder, make sure we divide only positive numbers    */
-    if (      0 <= i && 0 <= k )  i =       (  i %  k );
```
ChrisJefferson: The whole point of this code was (I assume) that C doesn't define the behaviour of / and % completely regards to negative numbers. Please at least put in the PR why you believe that isn't the case, and perhaps make sure we test all cases of mod / div (we might already do that).

fingolfin: We are now using C99, and C99 *does* specify the behaviour of `/` and `%`. I can of course add another comment stating that here.

I don't know what you mean with "make sure we test all cases of mod / div".

ChrisJefferson: Adding that would be nice. By testing all cases of `mod / div`, I meant "make sure we have a test for a mod b and a/b with each of these 4 cases (so say 5/2, -5/2, -5/-2, 5/-2), just in case someone has a stupid compiler (but if it's in C99, that seems unlikely)

fingolfin: Yes, we already extensively test all these cases. It's section 6.5.5 of the C99 standard, which specifies that integer division "truncates towards zero", and that `(a/b)*b + a%b == a` holds (which implies that `%` has the same sign as the dividend; which is what GAP wants for RemInt, where I already added a C99 comment to explain that.

I'll push an updated PR soon.

fingolfin: heh, actually, I just realized out of three places, your comment is on the one were I did not (and do not plan to) add a comment regarding C99, because here it actually does not matter: there are only two possible values for `i % k`, and here we always check for a negative one, and make it positive if necessary, thus the code here in `ModInt` is always correct, even in nonconforming pre-C99 compilers :-).

But I did add a comment for `QuoInt`, and expanded the existing one in `RemInt`.

Issue #2121, opened 2018-01-29
Fix and improve support for empty compressed vectors


fingolfin:
(This came up during the discussion of issue #2117)

Currently, `ConvertToVectorRep` and `ImmutableVector` refuse to turn an empty list (= "empty vector") into an empty compressed vector. On the long run, we want that, because one of the motivations for MatrixObj objects is precisely to cleanly support empty vectors, and 0xn / mx0 matrices.

This is ignored right now:
```
gap> v:=[]; ConvertToVectorRep(v, 2); IsGF2VectorRep(v);
[  ]
true
false
```

Now, you can cheat around this, by either directly calling `CONV_GF2VEC` resp. `CONV_VEC8BIT`, or by taking a sublist:
```
gap> v:=[Z(2)];; ConvertToVectorRep(v, 2);; v:=v{[]};
<a GF2 vector of length 0>
```

But it turns out that some of the kernel code is not prepared to deal with empty vectors (and matrices?), e.g.:

```
gap> NUMBER_GF2VEC(v);
Segmentation fault: 11
```

It would be nice to resolve this:
* [ ] audit the kernel code in `vec8bit.c` and `vecgf2.c` to make sure it works with empty compressed vectors and matrices
* [ ] in fact, add test code to check that systematically
* [ ] once done, enable `ConvertToVectorRep` and `ImmutableVector` (and any other MatrixObj constructors for compressed vectors and matrices) to produce empty vectors and matrices as appropriate.

stevelinton:
There is a bit of a conceptual problem with empty vectors (and matrices) coming from the "a vector is a just a dense homogeneous list of ring elements"  ideas that were baked into GAP 4 in its very early stages.  The problem shows up in many places, for instance in whether the length zero vector over GF(3) and the one over GF(5) are equal, and whether either of them is equal to an empty plain list.

fingolfin:
I'd really prefer if empty vectors over GF(3) and GF(5) were not equal to each other -- and for that matter, not equal to the empty plain list.

Amongst various other things, this would also resolve the problem of `[] in GF(5)^0` returning `false`, even though `[] in Elements(GF(5)^0)` returns `true`.

Considering how bad, if at all, empty vectors are supported overall, I would hope there aren't really problems with that. That said, I have no investigated this closely at all, so perhaps it is completely infeasible for other reasons?

[...]

PR #2125, opened 2018-01-29
WIP: Fixes for vector spaces with DimensionOfVectors(V) = 0 (DO NOT MERGE)

### LINKED ISSUE #2117: Regression: Irr runs into an error
The following happens currently in the master branch:

```
gap> g:= DirectProduct( AlternatingGroup(5), SymmetricGroup(3) );;
gap> Irr( g );;
Error, no 1st choice method found for `[]' on 2 arguments
 at /.../gap-master/gap/lib/methsel2.g:250 called from
[...]
```

`git bisect` says that the first bad commit is b6a4985b6cc1ab6dfaa185ec63653b1f9babcaba.

fingolfin:
[This is motivated by #2117 resp #2121; in particular, it fixes #2117)
Given a vector space V with DimensionOfVectors(V) = 0, e.g. V:=GF(5)^0, we currently
have Zero(V) = []. This commit makes the following snippets work:

* Coefficients(Basis(V), [])
* Position(Enumerator(V), [])
* [] in V

Especially the last point requires a rather nasty hack: By definition, "[a] collection in GAP consists of elements in the same family"; and any vector space `V` in GAP is such a collection, yet its only element, the empty list `[]`, is not in the correct family. This already is strictly speaking a bug, and forcing `\in` to work regardless could equally be described as a "fix" or as "making things even worse".

So all in all I am conflicted about merging this... But what is the alternative? For vectors over small fields, we could realize #2121 and make empty compressed vectors safe to use, but that doesn't solve e.g. the case of `Rationals^0`.

### DIFF
--- a/hpcgap/lib/basis.gi
+++ b/hpcgap/lib/basis.gi
@@ -1403,6 +1403,17 @@ InstallMethod( Coefficients,
     fi;
     end );
 
+InstallOtherMethod( Coefficients,
+    "for empty basis and empty list",
+    [ IsBasis and IsEmpty, IsEmpty and IsList ], SUM_FLAGS,
+    function( B, v )
+    if v = Zero( UnderlyingLeftModule( B ) ) then
+      return [];
+    else
+      return fail;
+    fi;
+    end );
+
 InstallMethod( LinearCombination,
     "for empty basis and empty list",
     [ IsBasis and IsEmpty, IsList and IsEmpty ], SUM_FLAGS,
--- a/lib/basis.gi
+++ b/lib/basis.gi
@@ -1398,6 +1398,17 @@ InstallMethod( Coefficients,
     fi;
     end );
 
+InstallOtherMethod( Coefficients,
+    "for empty basis and empty list",
+    [ IsBasis and IsEmpty, IsEmpty and IsList ], SUM_FLAGS,
+    function( B, v )
+    if v = Zero( UnderlyingLeftModule( B ) ) then
+      return [];
+    else
+      return fail;
+    fi;
+    end );
+
 InstallMethod( LinearCombination,
     "for empty basis and empty list",
     [ IsBasis and IsEmpty, IsList and IsEmpty ], SUM_FLAGS,
--- a/lib/modulrow.gi
+++ b/lib/modulrow.gi
@@ -209,6 +209,15 @@ InstallMethod( \in,
            and IsSubset( LeftActingDomain( M ), v );
     end );
 
+# TODO: HACK: properly document this
+InstallMethod( \in,
+    "for empty list and vector space with length 0 vector",
+    [ IsEmpty and IsList, IsFreeLeftModule and IsFullRowModule ],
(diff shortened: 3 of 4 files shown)


ThomasBreuer:
I agree with the statement by @fingolfin that it is not clear
whether changing the behaviour of empty lists in the context of domains
means to improve or to worsen the system.

Let us look at some more examples.

- Once `[]` is established as an element of each vector space `GF(q)^0`,
  one may expect that `GF(2)^0 = GF(3)^0` yields `true`,
  since two domains are equal if they have the same sets of elements.
  (Note that no `[]` object appears in this comparison question.)

- Once one has accepted the above equality,
  one may expect also that `[]` is contained in 
  `Intersection( GF(2)^0, GF(3)^0 )`.
  And if this intersection shall be represented as a domain 
  (which is recommended in general)
  then which characteristic shall it have?

- Two compressed lists of length zero in different characteristics 
  should be equal, by the general rule that two lists are equal
  if and only if their entries are equal at all positions.
  (Currently `=` returns `false` in this situation.)
  Thus the problem is really one about empty lists and not just about 
  `[]` objects.

One can change the behaviour of empty lists in particular situations,
in order to use (or misuse?) it in some code,
but one cannot get a consistent behaviour simply because the underlying 
rules are not consistent.

I think the only clean way out of this problem would be
to forbid domains containing empty lists.
Adding more support for empty lists as elements of domains 
will uncover inconsistencies in other places.

In practice,
I would suggest to avoid situations like the one from which the discussion arose
(that of extended vectors formed from length zero vectors, #2117).
A special treatment of the zero dimensional case looks perhaps less elegant
but would be safe.

fingolfin:
Regarding empty compressed vectors comparing equal: The "ideal" fix for that would be to make compressed vectors and matrices *not* lie in `IsList`. But I am afraid this might easily open new up new problems... Still, it's something I'd love to try for the GAP 4.10 cycle. 

But yeah, for now, we need to deal with the specific problem at hand.

BTW, the thing I don't like about adding special cases for zero dimensions everywhere is not so much that it's less elegant, but rather that it's a game of whack-a-mole: We'll never know that we didn't miss a spot, and it's quite likely people will write new code which lacks special cases for the zero-dimensional case.

stevelinton:
I think the right way to do this is just to retire the existing compressed vectors entirely in favour of somethign based on cvec and/or meataxe64 which have never been designed to lie in IsList. Cleaning up all the crud in the existing compressed vectors which is only there because of compatibility with lists is probably more work than writing something better.

However, I think that means switching all the linear algebra and matrix groups to work via MatrixObj, which is a huge task.

fingolfin:
Just to clarify: I also don't like the changes in this PR very much. I mostly put it out here to have a better foundation for discussion. 

I'll close this now, though, as I think it's clear we want to go in another direction. I added some more issues (#2148 and #2149) regarding empty vectors and possible semantics for comparing MatrixObj and VectorObj.

PR #2127, opened 2018-01-30
buildsys: only check once for supported compiler warnings

fingolfin:
This could be cherry-picked to stable-4.9 (the issue fixed by this PR is mostly harmless, only makes the configure check slightly longer; the fix should be perfectly safe)


alex-konovalov:
Just to confirm that this is cherry-picked onto stable-4.9 (I suggest to leave such a comment, otherwise it is completely non-obvious, and requires verification).

PR #2128, opened 2018-01-30
Memmove fix

ChrisJefferson:
This is an attempt to fix the bug in 32-bit memove, identified by Max Horn.

I have checked and up to date clang and gcc seem unable to convert SyMemmove to memmove.

I decided to just always use my function for 32-bit GAP, as it seems only about 25% slower than the built-in, and I have never seen memmove occur as a significant part of any GAP profile.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -115,6 +115,7 @@
 
 #include <src/gaputils.h>
 #include <src/io.h>
+#include <src/sysfiles.h>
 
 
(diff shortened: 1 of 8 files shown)


ChrisJefferson:
pushed wrong branch

ChrisJefferson:
Now correct branch.

The code is slower, I was trying to write an optimised version, have now decided to go for something extremely simple.

PR #2129, opened 2018-01-30
lib: only load queue.g when absolutely needed

### LINKED ISSUE #2066: Remove `lib/queue.g` again?
When merging HPC-GAP in, we added `lib/queue.g` which is loaded relatively early on by the library.

But nothing uses it, and its code has been imported (in modified form) into the datastructures packages.

So, can we remove it again?

fingolfin:
In particular, restrict it to HPC-GAP.

Resolves #2066 

Once merged, this should be cherry-picked into stable-4.9


alex-konovalov:
Now cherry picked onto stable-4.9

Issue #1944, opened 2017-11-22
Write some text introducing HPC-GAP and warning about its status (e.g. as README.hpcgap.md)


fingolfin:
Since we are going to bundle HPC-GAP, people will discover it and wonder about it.

One really bad thing that could happen is that people find it, try it, and are disappointed, and then will never consider it again. Or people may be interested, but can't figure it out.

So I think we should have a `README.hpcgap.md` which briefly says what HPC-GAP is, what its status is, how to build it, and where to learn more. Perhaps also a suggestion where to write if one wants to learn more.

alex-konovalov:
if https://github.com/gap-system/gap/wiki/Building-HPC-GAP is up to date, that could provide half of this file...

alex-konovalov:
Closed by #2130

PR #2130, opened 2018-01-30
Minimalistic HPC-GAP README file

alex-konovalov:
This addresses some suggestions from #1944, and needs to be placed in GAP 4.9.0

### DIFF
--- a/README.hpcgap.md
+++ b/README.hpcgap.md
@@ -1,0 +1,23 @@
+# HPC-GAP
+
+GAP includes experimental code to support multithreaded programming in GAP,
+dubbed HPC-GAP (where HPC stands for "high performance computing"). GAP and
+HPC-GAP codebases diverged during the project, and we are currently working
+on unifying the codebases and incorporating the HPC-GAP code back into the
+mainstream GAP versions.
+
+This is work in progress, and HPC-GAP as it is included with GAP right now
+still suffers from various limitations and problems, which we are actively
+working on to resolve. However, including it with GAP (disabled by default)
+considerably simplifies development of HPC-GAP. It also means that you can
+very easily get a (rough!) sneak peak of HPC-GAP. It comes together with the
+new manual book called "HPC-GAP Reference Manual" and located in the `doc/hpc`
+directory.
+
+Users interested in experimenting with shared memory parallel programming in
+GAP can build HPC-GAP by following the instructions from
+https://github.com/gap-system/gap/wiki/Building-HPC-GAP. While it is possible
+to build HPC-GAP from a release version of GAP you downloaded from the GAP
+website, due to the ongoing development of HPC-GAP, we recommend that you
+instead build HPC-GAP from the latest development version available in the
+GAP repository at GitHub, i.e. https://github.com/gap-system/gap.


README.hpcgap.md
```diff
@@ -0,0 +1,14 @@
+# HPC-GAP
+
+The beta release of GAP 4.9 includes experimental code to support multithreaded
```
fingolfin: "the beta release"? But the final release also will include it, no?

Also, should we really mention a version here?

So perhaps just say: "GAP includes experimental code..."

README.hpcgap.md
```diff
@@ -0,0 +1,14 @@
...
+programming in GAP. GAP and HPC-GAP codebases diverged during the project, and
```
fingolfin: This text seems to assume the reader is already familiar with the term HPC-GAP. Perhaps:

> ... programming in GAP, dubbed HPC-GAP (where HPC is short for `high performance computing`).

README.hpcgap.md
```diff
@@ -0,0 +1,14 @@
...
+we are currently working on unifying the codebases and incorporating the HPC-GAP
+code back into the mainstream GAP versions.
+
+This is work in progress, but including it into GAP 4.9 release will allow to
+conduct further code refactoring more efficiently, and will expose some new
+code to make visible new names and keywords.
```
fingolfin: This is our reason, but I am not sure it is helpful to tell end users. Perhaps better to say something more like this:
> This is work in progress, and HPC-GAP as it is included with GAP right now still suffers from various limitations and problems, which we are actively working on to resolve. However, including it with GAP (disabled by default) considerably simplifies development of HPC-GAP. It also means that you can very easily get a (rough!) sneak peak of HPC-GAP.

Well.. or something in this line... I don't claim what I wrote is great, sorry :/

README.hpcgap.md
```diff
@@ -0,0 +1,14 @@
...
+we are currently working on unifying the codebases and incorporating the HPC-GAP
+code back into the mainstream GAP versions.
+
+This is work in progress, but including it into GAP 4.9 release will allow to
+conduct further code refactoring more efficiently, and will expose some new
+code to make visible new names and keywords. 
+
+Users interested in experimenting with shared memory parallel programming
+in GAP are advised to check out the GAP repository from GitHub and follow
+instructions from https://github.com/gap-system/gap/wiki/Building-HPC-GAP.
```
fingolfin: Perhaps:
> Users interested in experimenting with shared memory parallel programming in GAP can build HPC-GAP by following the instructions from https://github.com/gap-system/gap/wiki/Building-HPC-GAP. While you can build HPC-GAp from a release version of GAP you downloaded from the GAP website, due to the ongoing development of HPC-GAP, we recommend that you instead build HPC-GAP from the latest development version available in the GAP repository at GitHub, i.e. https://github.com/gap-system/gap

fingolfin: Also, somebody should double check those build instructions are up-to-date (perhaps @ssiccha would be willing to help).

alex-konovalov:
@fingolfin comments make sense, thanks - updated the PR with new text.

We need to keep in mind updating https://github.com/gap-system/gap/wiki/Building-HPC-GAP referenced in the submitted file - if @ssiccha could help, it will be much appreciated indeed.

alex-konovalov:
Actually "While you can build HPC-GAP from a release version of GAP you downloaded from the GAP website" is true, but you will have to satisfy dependencies and download Ward which is not bundled with GAP 4.9.0. But the instructions at https://github.com/gap-system/gap/wiki/Building-HPC-GAP mention Ward, so I assume that is ok to phrase like that.

ssiccha:
@alex-konovalov then I would rephrase that as 
> While it is possible to build HPC-GAP from a release version of GAP...

alex-konovalov:
Thanks @ssiccha @fingolfin - took all suggestions into account.

alex-konovalov:
@fingolfin @ssiccha should this also document the `if IsBound(HPCGAP)` trick?

ssiccha:
@alex-konovalov I don't think that the `README.hpcgap.md` would be the right place for that. Maybe one could put that into the manual at some point?

fingolfin:
I agree with @ssiccha (and that explanation should then also cover `if IsHPCGAP`, and when one should use which (basically, the former if one cares about GAP 4.8 compatibility, else the latter)

ssiccha:
I agree with @stevelinton. Does it make sense to add a link to an online version of the HPC-GAP manual?

alex-konovalov:
Good idea. It will be available online when we will make a public release of GAP 4.9. I will add a text saying that the documentation is in `/doc/hpc`.

alex-konovalov:
Added "It comes together with the
new manual book called "HPC-GAP Reference Manual" and located in the `doc/hpc`
directory."

ssiccha:
Hmm, I think the manual is not in a good state at the moment. Maybe say something like "the manual will be available in version 4.9.1"?

README.hpcgap.md
```diff
@@ -0,0 +1,23 @@
...
+HPC-GAP codebases diverged during the project, and we are currently working
+on unifying the codebases and incorporating the HPC-GAP code back into the
+mainstream GAP versions.
+
+This is work in progress, and HPC-GAP as it is included with GAP right now
+still suffers from various limitations and problems, which we are actively
+working on to resolve. However, including it with GAP (disabled by default)
+considerably simplifies development of HPC-GAP. It also means that you can
+very easily get a (rough!) sneak peak of HPC-GAP. It comes together with the
+new manual book called "HPC-GAP Reference Manual" and located in the `doc/hpc`
```
ssiccha: and _is_ located

alex-konovalov: hmm... I think either "which is called... and is located" or "called ... and located", but not just "is".

alex-konovalov:
@ssiccha I don't want to modify release wrapping scripts to drip the manual and add it later again. Is the state of it acceptable for *beta* release ? Why not to create an issue and lost specific things to be done for the public release?

ssiccha:
@alex-konovalov Ok, I'll have a look. There are a few things which don't fit quite right IIRC.

alex-konovalov:
Ok, but I am afraid fixing them, with CI tests and review round will fall behind the desired timescale. It's beta release, and this should either have been flagged for 4.9.0 to work earlier, or should not be hurried up.

alex-konovalov:
Cherry-picked onto stable-4.9

PR #2135, opened 2018-01-31
Fix typos in documentation of ContainingConjugates and ContainedConjugates

markuspf:
Really minor fix, but I stumbled over it too many times over the last couple of days.

Should be cherry-picked to `stable-4.9`.


markuspf:
In a current GAP it looks like this (because `\ge` is not valid latex) 
```
G ge A,B
```
the switch to the back is because I find it nicer that way around. If @hulpke objects I'll change it to the other way around, but stil fix the typo.

alex-konovalov:
Cherry-picked onto stable-4.9

hulpke:
Yes, the \ge should have been \geq. If the other way round pleases anyone more, I have no objection.

PR #2136, opened 2018-01-31
add shortname to IsomorphismTypeInfoFiniteSimpleGroup value

ThomasBreuer:
The record returned by `IsomorphismTypeInfoFiniteSimpleGroup`
contains a `name` component that is not suitable for
programmatic access.

The new component `shortname` fits to the names that are used in
`CTblLib` and `AtlasRep`.
It will simplify some of my example sessions.
In particular I will be able to get rid of some calls to
`StructureDescription`; I think it is not appropriate to advertise
that function by applying it to simple groups.

    TB

### DIFF
--- a/doc/tut/group.xml
+++ b/doc/tut/group.xml
@@ -202,7 +202,8 @@ gap> IsSimple( f ); IsomorphismTypeInfoFiniteSimpleGroup( f );
 true
 rec( 
   name := "A(1,7) = L(2,7) ~ B(1,7) = O(3,7) ~ C(1,7) = S(2,7) ~ 2A(1,\
-7) = U(2,7) ~ A(2,2) = L(3,2)", parameter := [ 2, 7 ], series := "L" )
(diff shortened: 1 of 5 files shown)


hulpke:
Let me just note that there also is now `DataAboutSimpleGroup` which includes the `IsomorphismType` record, but also easier parseable information about the group. It is used currently in subgroup lattice and automorphism group calculations for almost simple groups, and has a component tomName which seems to be the same as `shortname`.

ThomasBreuer:
@hulpke: Interesting, thanks for the hint.
`DataAboutSimpleGroup` seems to be undocumented.
I think `IsomorphismTypeInfoFiniteSimpleGroup` is a more suitable place for what I call `shortname`,
this function is applicable also to character tables and positive integers (documented).

alex-konovalov:
Thanks - should this be mentioned in release notes then?

ThomasBreuer:
@alex-konovalov:
I am not sure about the release notes.
On the one hand, I hope that this small change does not belong to the "most important" ones.
On the other hand, where would someone find a remark about the new record component?

PR #2146, opened 2018-02-01
HPC-GAP: sync lib/profile.g and lib/system.g

fingolfin:


fingolfin:
Actually, I did intend the `atomic` blocks in other similar diffs, must have forgotten about it here. Will do it now.

PR #2151, opened 2018-02-02
Memmove fix

ChrisJefferson:
This is a basic attempt to fix the problem of broken memmove on 32-bit glibc.

This is purposefully designed to be as simple as possible. We could obviously do various cleverer things than this, but I find the bug hard to test, so I didn't want to do anything particularly complicated.


src/sysfiles.h
```diff
@@ -612,6 +613,9 @@ extern Obj SyReadStringFid(Int fid);
 extern Obj SyReadStringFile(Int fid);
 extern Obj SyReadStringFileGeneric(Int fid);
 
+// Internal implementation of memmove, to avoid issues with glibc
+void *SyMemmove(void *dst, const void* src, UInt size);
```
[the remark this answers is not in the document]
ChrisJefferson: True, or even (which I normally disaprove) just doing `#define SyMemmove memmove`

src/sysfiles.c
```diff
@@ -3598,6 +3598,56 @@ Obj SyReadStringFid(Int fid) {
...
+#ifdef SYS_IS_64_BIT
+void * SyMemmove(void * dst, const void * src, UInt size)
+{
+    return memmove(dst, src, size);
+}
+#else
+// The memmove in glibc on 32-bit SSE2 systems, contained in
+// __memmove_sse2_unaligned, is buggy in at least versions
+// 2.21 - 2.25 when crossing the 2GB boundary, so GAP must
+// include it's own simple memmove implementation.
```
fingolfin: "it's" -> "its"

Perhaps also link to the glibc bug report. Oh, and they just release 2.26 which also has the bug.

src/sysfiles.c
```diff
@@ -3598,6 +3598,56 @@ Obj SyReadStringFid(Int fid) {
...
+            *(d - 3) = *(s - 3);
+            d -= 4;
+            s -= 4;
+            size -= 4;
+        }
+        // Finish
+        while (size > 0) {
+            *d-- = *s--;
+            size--;
+        }
```
fingolfin: I don't understand how this code can be correct: normally, any `memmove` implementation needs two code paths: one which moves data "forward", another which moves its "backwards". One deals with `memmove(ptr+1, ptr, size)` one with `memmove(ptr-1, ptr, size)`.

Looking at it, I think this code here always copies backwards, which would mean `memmove(ptr-1, ptr, size)` produces garbage. Am I missing something?

fingolfin: Once again, we could really benefit from test setup for pure C code, so that one can write unit tests for functions like this.

ChrisJefferson:
Now fixed up (hopefully)

src/sysfiles.c
```diff
@@ -3589,6 +3589,67 @@ Obj SyReadStringFid(Int fid) {
...
+        memcpy(dst, src, size);
+    }
+    else if (d > s) {
+        d = d + (size - 1);
+        s = s + (size - 1);
+
+        // This is about 4x slower than glibc, but
+        // is simple and also complicated enough that
+        // gcc or clang seem unable to "optimise" it back
+        // into a call to memmove.
```
fingolfin: Under which condition did you test this? E.g. did you also try it with `-O3`?

ChrisJefferson: I tried with `-O3` and `-flto`. I checked using `objdump -t` that the resulting GAP executable doesn't have any references to `memmove`.

PR #2155, opened 2018-02-04
Increase coverage for partial perm kernel code; remove unused (& buggy?) FuncPrintPPerm2/4

fingolfin:


markuspf:
I can't see anything wrong with the patch, maybe @james-d-mitchell can comment on the unused function before I (or he) merge(s).

james-d-mitchell:
Looks good to me, I have no memory of why there is a print function in C and in GAP, so am happy if it is removed especially if it contains some bugs.

fingolfin:
Thanks James! Re bugs: I tried to call it directly to see what the output looks like, and it got stuck in an endless loop printing stuff. Didn't look for the cause, though, as removing it seemed the easiest way to debug it ;-)

fingolfin:
@james-d-mitchell One remark: We usually  use "rebase" instead of "merge" these days (also from the GitHub UI)

james-d-mitchell:
@fingolfin ergh, right you told me that the last time I did this too, sorry!

PR #2158, opened 2018-02-05
Document LowIndexSubgroups, add LowLayerSubgroups, some fixes (partial rebase of #2035)

fingolfin:
This is a rebased version of some changes in PR #2035. For an explanation, see there. My hope is that we can merge this ASAP into `master`, then immediately after cherry-pick it into master. 

The changes in this branch should be cherry-picked into stable-4.9.

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -4495,27 +4495,26 @@ DeclareGlobalFunction("GroupEnumeratorByClosure");
 ##
 ##  <#GAPDoc Label="LowIndexSubgroups">
 ##  <ManSection>
-##  <Oper Name="LowIndexSubgroups"
-##   Arg='G, index'/>
 ##  <Oper Name="LowIndexSubgroups" Arg='G, index'/>
 ##
 ##  <Description>
-##  These operations computes representatives of the conjugacy classes of
-##  subgroups of the group <A>G</A> that 
-##  index less than or equal to
-##  <A>index</A>.
+##  The operation <Ref Oper="LowIndexSubgroups"/> computes representatives of
+##  the conjugacy classes of subgroups of the group <A>G</A> that index less
+##  than or equal to <A>index</A>.
 ##  <P/>
 ##  For finitely presented groups this operation simply defaults to
-##  <C>LowIndexSubgroupsFpGroup</C>. In other cases, it uses repeated
+##  <Ref Oper="LowIndexSubgroupsFpGroup"/>. In other cases, it uses repeated
 ##  calculation of maximal subgroups.
 ##  <P/>
+##  The function <Ref Func="LowLayerSubgroups"/> works similar but does not
+##  bound the index, but instead considers up to <A>layer</A>-th maximal
+##  subgroups.
 ##  <Example><![CDATA[
 ##  gap> g:=TransitiveGroup(18,950);;
 ##  gap> l:=LowIndexSubgroups(g,20);;Collected(List(l,x->Index(g,x)));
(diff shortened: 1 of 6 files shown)


lib/grp.gd
```diff
@@ -4495,12 +4495,12 @@ DeclareGlobalFunction("GroupEnumeratorByClosure");
...
 ##  <#GAPDoc Label="LowIndexSubgroups">
 ##  <ManSection>
-##  <Oper Name="LowIndexSubgroups"
-##   Arg='G, index'/>
 ##  <Oper Name="LowIndexSubgroups" Arg='G, index'/>
+##  <Oper Name="LowLayerSubgroups" Arg='G, layer'/>
 ##
 ##  <Description>
-##  These operations computes representatives of the conjugacy classes of
+##  The operation <C>LowIndexSubgroups</C> computes representatives of the
```
alex-konovalov: Actually GAPDoc recommends to use `<Ref>` inside `<ManSection>` even if it refers to the object documented in this ManSection: see example here:

https://github.com/frankluebeck/GAPDoc/blob/6108fcbca047c8b6484c512c3ba5c4bdd5bb8b7c/example/example.xml#L109-L118

GAPDoc will not expand the reference in this case. IIRC this is useful because if you copy and paste the text elsewhere, `<Ref>` will be there already, and also in case the manual layout will display `<Ref>` and `<C>` elements differently. Perhaps there are actually deeper reasons, which @frankluebeck may tell.

frankluebeck:
>  in case the manual layout will display <Ref> and <C> elements differently.

Yes, an output format could decide different layout for referenced functions/operations/... and for general code.

fingolfin:
So I can adjust the GAPDoc comments of course, no problem, just should warn that @hulpke will then have to deal with it when rebasing his PR on master -- though it shouldn't be a big issue, one of his commits will cause a conflict, but can then be skipped (via `git rebase --skip`).

lib/grp.gd
```diff
@@ -4509,13 +4509,15 @@ DeclareGlobalFunction("GroupEnumeratorByClosure");
 ##  <C>LowIndexSubgroupsFpGroup</C>. In other cases, it uses repeated
 ##  calculation of maximal subgroups.
 ##  <P/>
+##  The operation <C>LowLayerSubgroups</C> works similar but does not bound
```
fingolfin: While looking into tweaking the GAPDoc, I realized that `LowLayerSubgroups` actually is a global function, not an operation. Moreover, it takes additional (optional) arguments, and already has an  documentation comment in `grplatt.gd` (which also is in the manual). 

Perhaps the plan was/is to turn `LowLayerSubgroups` into an operation? In any case, the change as it is in this PR right now does not seem correct... @hulpke ?

lib/grp.gd
```diff
@@ -4495,27 +4495,27 @@ DeclareGlobalFunction("GroupEnumeratorByClosure");
 ##
 ##  <#GAPDoc Label="LowIndexSubgroups">
 ##  <ManSection>
-##  <Oper Name="LowIndexSubgroups"
-##   Arg='G, index'/>
 ##  <Oper Name="LowIndexSubgroups" Arg='G, index'/>
+##  <Oper Name="LowLayerSubgroups" Arg='G, layer'/>
```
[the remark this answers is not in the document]
hulpke: Yes, this probably should just be replaced by a reference to <C>LowLayerSubgroup</A> (or the REF-syntax...). Do you want me to send you a replacement draft?

fingolfin:
@hulpke I now updated the ref to `LowLayerSubgroups` in the documentation comment. Does it look OK to you?

hulpke:
@fingolfin Yes, its fine (and I think I know enough git to deal with the superfluous bits in the other PR as well once this is merged.)

PR #2160, opened 2018-02-05
Catch overflows in Gasman

ChrisJefferson:
The main purpose of this patch is to catch a couple of pointer overflows in gasman. I also add some tests to make future problems easier to catch.

I suspect there may be other overflows, particularly is the memory space gets very close to 4GB, but this fixes some problems I know about, and my prefered solution to fixing other problems would be instead to limit the size of the memory space, so (for example) the GAP workspace stops at least 1GB from the end of the memory space, and we limit largest object to 1GB.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -2089,7 +2089,7 @@ again:
     /* information after the check phase                                   */
     if ( MsgsFuncBags )
       (*MsgsFuncBags)( FullBags, 5,
-                       SpaceBetweenPointers(EndBags, stopBags)/(1024/sizeof(Bag)));
+                       (EndBags - stopBags)/(1024/sizeof(Bag)));
     if ( MsgsFuncBags )
         (*MsgsFuncBags)( FullBags, 6,
                          SizeWorkspace/(1024/sizeof(Bag)));
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -300,6 +300,7 @@ void CHANGED_BAG(Bag bag) {
 
 static inline UInt SpaceBetweenPointers(Bag * a, Bag * b)
 {
+    GAP_ASSERT(b <= a);
     UInt res = (((UInt)((UInt)(a) - (UInt)(b))) / sizeof(Bag));
     return res;
 }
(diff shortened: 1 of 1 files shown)


src/gasman.c
```diff
@@ -1325,7 +1339,8 @@ UInt ResizeBag (
         /* check that enough storage for the new bag is available          */
         if ( SizeAllocationArea <  WORDS_BAG(sizeof(BagHeader)+new_size)
               && CollectBags( new_size, 0 ) == 0 ) {
-            return 0;
+            fputs("gap: cannot extend the workspace any more!!!!!!\n",stderr);
+            SyExit( 1 );
```
fingolfin: Perhaps use `SyAbortBags` (not terribly important, just sayin')

src/gasman.c
```diff
@@ -1943,6 +1960,11 @@ UInt CollectBags (
 
     /* * * * * * * * * * * * * * * check phase * * * * * * * * * * * * * * */
 
+    // Check if this allocation would even fit into memory
+    if(((size_t)-1) - (size_t)(sizeof(BagHeader)+size) < (size_t)AllocBags) {
```
fingolfin: missing space after `if`. Also `((size_t)-1)`  should be `SIZE_MAX` (from ISO C).

But perhaps the more idiomatic way would be to add a check after setting `stopBags`:
```C
// check for an overflow
if (stopBags < AllocBags)
  SyAbortBags("gap: cannot extend the workspace any more!!");
```

(I assume you want to abort, not `return 0`?

src/gasman.c
```diff
@@ -1295,9 +1308,10 @@ UInt ResizeBag (
     else if (CONST_PTR_BAG(bag) + WORDS_BAG(old_size) == AllocBags) {
         CLEAR_CANARY();
         // check that enough storage for the new bag is available
-        if (EndBags < CONST_PTR_BAG(bag) + WORDS_BAG(new_size)
+        if (SpaceBetweenPointers(EndBags, CONST_PTR_BAG(bag)) < WORDS_BAG(new_size)
               && CollectBags( new_size-old_size, 0 ) == 0 ) {
-            return 0;
+            fputs("gap: cannot extend the workspace any more!!!!!\n",stderr);
```
fingolfin: So it seems the number of exclamation marks is supposed to help distinguish these errors?

ChrisJefferson: The seems to be the existing pattern, I just carried it on :)

src/gasman.c
ChrisJefferson: Ah, I didn't notice that function.

src/gasman.c
ChrisJefferson: Pointer overflow isn't defined. gcc will (for example) optimise `i < i + 1` away to true for pointer `i`

ChrisJefferson:
Hopefully all issues fixed up

fingolfin:
Do we want to backport this (and the `SyMemmove` PR) to stable-4.9 ?

PR #2162, opened 2018-02-06
Move AlgExtSquareHensel to dev/deleted-code/

fingolfin:
`AlgExtSquareHensel` was added by @hulpke in 2003, and then disabled by him in a commit 2014-05-05, with a comment stating:
> the second algorithm seem to have problems -- temp. disable

In that very same commit, the function `DecomPoly` was added, but nothing uses it.

This PR moves both of these functions (totalling over 1000 line of code) to `dev/deleted-code/`, where they can be reactivate from if somebody (e.g. @hulpke ) has use for them in the future.


fingolfin:
While I *personally* don't see the usefulness of 1, I am happy to accept that it would be useful for you, and thus gladly will do it. We already use the `dev/` and `etc/` dirs to collect things not distributed. How about `dev/deleted-code/` (hm, there is already a file `dev/DELETED`, and also `lib/DELETED`).

You write 
> [...] DecomPoly (which is the work routine for subfield finding [...]
which makes it sound as if this function was in use, which confuses me, as this function quite clearly never is called (you can grep for it). What am I missing?

Somewhat offtopic, but: personally, I'd think that having better interfaces to number theory systems like pari (via `Alnuth` or the new `PariInterface` @markuspf is working on) or Nemo and Hecke (e.g. via @sebasguts's `JuliaInterface` and the OSCAR project in general) are perhaps better ways to get access to state of the art number theory code.

fingolfin:
@hulpke I added the removed code back under the `dev/deleted-code` dir. OK now?

fingolfin:
@hulpke since you seem to be looking at PRs and issues right now, could you take another look here, too?

hulpke:
@fingolfin Sorry for the long delay -- still figuring out how to get through the huge amount of github mail.

Anyhow.: Thanks for keeping the hensel code in the repository.

I have added documentation (and examples) to `DecomPoly`, thus removing the reason for deleting the code. (I will approve in a moment but the commit about removing `DecomPoly` will need to be taken out when rebasing.)

Questions in earlier comments:

> [...] DecomPoly (which is the work routine for subfield finding [...]
> which makes it sound as if this function was in use

Sorry. not in GAP, but if you want to find subfields in characteristic zero, this is the routine that would be needed. It is similar to some of the number theoretic functions (like `Sigma`) which we provide but never use.

>  I'd think that having better interfaces to number theory systems like pari (via Alnuth or the new PariInterface @markuspf is working on) or Nemo and Hecke (e.g. via @sebasguts's JuliaInterface and the OSCAR project in general) are perhaps better ways to get access to state of the art number theory code.

I coldn't agree more once such functionality is available, but there is the question on getting such interfaces working for the general user (say, who has a managed Windows system.) Our own number theoretic routines will be fallback until such interfaces exist.
(In the course of GAP 4 over the last 20+ years we have far too often delayed work waiting for a perfect third-party solution which the never really materialized. This is a reason behind many of our holes in functionality.)

fingolfin:
I have now updated this PR to retain `DecomPoly`

PR #2165, opened 2018-02-06
Deal with Random changing attributes of object

ChrisJefferson:
This changes the tester to add one initial call to `Random` when testing random generation. This deals with the problem where the first call to `Random` adds attributes, which changes which `Random` method gets called. Therefore calling `Random` twice, even if we reset `GlobalMersenneTwister`, produces different values.

The other alternative (why I marked this discussion) is that we could say this is a bug, and different overloads must provide the same sequence of random values.

### DIFF
--- a/tst/testrandom.g
+++ b/tst/testrandom.g
@@ -6,6 +6,10 @@ randomTest := function(collection, method, checkin...)
         checkmethod := checkin[1];
     fi;
 
+    # We do a single call first, to deal with calling Random causing extra attributes
+    # of 'collection' to be set, changing the dispatch
+    method(collection);
+
     # Firstly, we will generate a base list
     Init(GlobalMersenneTwister, 6);
     test1 := List([1..1000], x -> method(collection));


fingolfin:
Could you give a concrete example where this takes effect?

ChrisJefferson:
Commit d51d4588a0283f in #2057

fingolfin:
@ChrisJefferson I don't understand, I see no`Random` in that commit?

fingolfin:
Ahhh, I assume f48dc217d8a7dee25abab63ffdd98813bfa7fc40 is meant -- I commented there on a changed `Random` output, which I did not understand. Now I get it.

ChrisJefferson:
I'm sure I do mean d51d458. 
While that commit doesn't exactly mention Random, it seems to effect it.

That can be observed by:

```
gap> g := CyclicGroup(IsMatrixGroup, GF(3), 3);
Group([ [ [ 0*Z(3), Z(3)^0, 0*Z(3) ], [ 0*Z(3), 0*Z(3), Z(3)^0 ], [ Z(3)^0, 0*Z(3), 0*Z(3) ] ] ])
gap> Read("tst/testrandom.g");
gap>  randomTest(g, Random);
```

Which will fail on that commit.

We can fix the problem by asking for `NiceMonomorphism(g)` before calling `randomTest`, or alternatively calling `Random(g)` at least once (which itself creates a NiceMonomorphism).

stevelinton:
Logically the "call this a bug" option would seem to mean that we want Random, with the same random seed, to produce the same results on equal objects, the fact that the two objects in this case are the same one with different known attributes is kind of incidental.

fingolfin:
I agree with @stevelinton -- imposing any kind of consistency rule here will be extremely painful to implement, for a very small gain. As it is, even without this rule, all computations are still 100% deterministic, but of course there are cases were it is tricky to repeat a computation unless one can repeat the input 100% precisely. That's annoying, but still relatively uncommon and exotic, and not enough reason to limit implementors by imposing a strict rule here.

Of course in specific cases, we can and should try to make separate `Random` methods compatible, but even in the one that triggered this PR, I don't see how, short of removing the useful optimization of computing random elements through a nice monomorphism (or something similar), if available...

BTW, I am going to @-mention @hulpke now, just in case he has something to add?

ChrisJefferson:
I'm happy to keep adjusting this test in future, where approriate. I certainly wouldn't want to enshrine into GAP any promises about repeatability.

At the moment it isn't even true that if you run the same GAP twice with the same inputs, you get the same outputs (as some algorithms run based on time).

PR #2168, opened 2018-02-07
Minor cleanup and refactoring

fingolfin:


james-d-mitchell:
The bits of this in `trans.c`, `pperm.c` and `permutat.c` look very sensible to me.

ChrisJefferson:
My only minor comment is if you fixed a bug in  b316348, then there should probably be a test to show the error is fixed :)

fingolfin:
There is no user visible bug; this is merely an optimization zhat never triggers, but the generic code ends up returning the right result anyway. So perhaps instead of "bug" Inshpuld write "invalid logic"? Seems a bit nitpicky, though....

PR #2170, opened 2018-02-07
Add sort by master pointer

ChrisJefferson:
This adds a new sort function which sorts plists by master pointer, and then makes use of this new function in pperm.c and trans.c

### DIFF
--- a/src/listfunc.c
+++ b/src/listfunc.c
@@ -630,6 +630,24 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
 
 #include <src/sortbase.h>
 
+// This is a variant of SortDensePlist, which sorts plists by
+// Obj pointer. It works on non-dense plists, and can be
+// used to efficiently sort lists of small integers.
+
+#define SORT_FUNC_NAME SortPlistByRawObj
+#define SORT_FUNC_ARGS Obj list
+#define SORT_ARGS list
+#define SORT_CREATE_LOCAL(name) Obj name;
+#define SORT_LEN_LIST() LEN_PLIST(list)
+#define SORT_ASS_LIST_TO_LOCAL(t, i) t = ELM_PLIST(list, i)
+#define SORT_ASS_LOCAL_TO_LIST(i, j) SET_ELM_PLIST(list, i, j);
+#define SORT_COMP(v, w) ((v) < (w))
+#define SORT_FILTER_CHECKS() \
+    RESET_FILT_LIST(list, FN_IS_NSORT); \
+    RESET_FILT_LIST(list, FN_IS_SSORT);
+
+#include <src/sortbase.h>
+
 /****************************************************************************
 **
 *F  SORT_LISTComp(<list>,<func>)  . . . . . . . . . . . . . . . . sort a list
--- a/src/pperm.c
+++ b/src/pperm.c
@@ -19,6 +19,7 @@
 #include <src/integer.h>
 #include <src/intfuncs.h>
 #include <src/io.h>
+#include <src/listfunc.h>
 #include <src/lists.h>
 #include <src/opers.h>
 #include <src/permutat.h>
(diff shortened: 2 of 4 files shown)


src/listfunc.c
```diff
@@ -630,6 +630,24 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
...
+
+#define SORT_FUNC_NAME SortPlistByMasterPointer
+#define SORT_FUNC_ARGS Obj list
+#define SORT_ARGS list
+#define SORT_CREATE_LOCAL(name) Obj name;
+#define SORT_LEN_LIST() LEN_PLIST(list)
+#define SORT_ASS_LIST_TO_LOCAL(t, i) t = ELM_PLIST(list, i)
+#define SORT_ASS_LOCAL_TO_LIST(i, j)                                         \
+    SET_ELM_PLIST(list, i, j);                                               \
+    CHANGED_BAG(list);
```
fingolfin: Do you really need to call `CHANGED_BAG` here? AFAICT, no operations which could trigger GC happen. And even if, the bags contained in the list don't change, they just are shuffled around, right?

src/listfunc.c
```diff
@@ -630,6 +630,24 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
...
+#define SORT_COMP(v, w) ((v) < (w))
+#define SORT_FILTER_CHECKS() RESET_FILT_LIST(list, FN_IS_NSORT);
```
fingolfin: Shouldn't this also reset `FN_IS_SSORT`?

Alternatively, leave it out completely, and document that it is up to the caller to adjust the TNUM.

src/listfunc.c
```diff
@@ -630,6 +630,24 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
 
 #include <src/sortbase.h>
 
+// This is a variant of SortDensePlist, which sorts plists by
+// master pointer. It works on non-dense plists, and can be
+// used to efficiently sort lists of small integers
+
+#define SORT_FUNC_NAME SortPlistByMasterPointer
```
fingolfin: I am not happy with that name, for two reasons: It does not make it clear at all that one can use it to sort immediates; and "master pointers" really are an implementation detail of GASMAN. Sure, we currently use "fake" master pointers with Boehm GC, too, but to me that's mostly to make it easier to use Boehm GC as a drop-in replacement.

Perhaps `RawSortPlist` or `SortPlistByRawObj` would be less loaded? (Though certainly not great either -- as usual, naming is one of the hardest problems in CS ;-)

src/pperm.c
```diff
@@ -288,28 +288,10 @@ UInt RANK_PPERM4(Obj f)
 
 static Obj SORT_PLIST_CYC(Obj res)
 {
-    Obj  tmp;
-    UInt h, i, k, len;
-
-    len = LEN_PLIST(res);
-    if (len == 0)
+    if(LEN_PLIST(res) == 0)
```
fingolfin: Add space after `if`

src/pperm.c
```diff
@@ -288,28 +288,10 @@ UInt RANK_PPERM4(Obj f)
 
 static Obj SORT_PLIST_CYC(Obj res)
```
fingolfin: Not your fault, but this function is really named incorrectly. It should be called `SORT_PLIST_INTOBJ`. I can take care of that in a later PR (but of course if you are willing to do it here, you are more than welcome ;-)

fingolfin: Alternatively, just fold it into the calling code (but again, I can take care of that in a later PR)

src/trans.c
```diff
@@ -1172,7 +1140,7 @@ Obj FuncIMAGE_SET_TRANS(Obj self, Obj f)
     Obj out = FuncUNSORTED_IMAGE_SET_TRANS(self, f);
 
     if (!IS_SSORT_LIST(out)) {
-        SORT_PLIST_CYC(out);
+        SortPlistByMasterPointer(out);
         RetypeBag(out, T_PLIST_CYC_SSORT + IMMUTABLE);
         CHANGED_BAG(out);
```
fingolfin: Reminder to self: get rid of this useless `CHANGED_BAG` (and other similar ones below) in a future PR (or feel free to remove it now; but no need to bloat this PR)

src/trans.c
```diff
@@ -1243,7 +1211,7 @@ Obj FuncIMAGE_SET_TRANS_INT(Obj self, Obj f, Obj n)
         }
         SHRINK_PLIST(new, (Int)rank);
         SET_LEN_PLIST(new, (Int)rank);
-        SORT_PLIST_CYC(new);
+        SortPlistByMasterPointer(new);
         RetypeBag(new, T_PLIST_CYC_SSORT);
```
fingolfin: Another reminder to self: This line is buggy, it is missing a `+ IMMUTABLE`. For that matter, one could simply create the bag with the right TNUM, and avoid the `RetypeBag` call here (assuming `SortPlistByMasterPointer` would not call RESET_FILTER)

stevelinton:
It seems to me there are two use cases for this function: sorting lists of immediate integers -- that works fine, and this is always the right way to do it; sorting lists of arbitrary objects so that we can recognise when an object is in the list twice (actually the same object, not just up to equality) or use binary search to see if an object is present (same caveat). This is going to be quite confusing if people do not understand the difference between equality and identity. For instance Z(3) and Z(9)^4 are not identical, although they are equal.  Also the order will not repeat reliably between sessions.

I wonder if presenting it as two separate functions for the two use cases, with the non-immediate integer one suitably plastered with warning signs, is the best way to do it.

fingolfin:
Going into the same way as @stevelinton, for the use cases in here and that I have in mind, I'd actually prefer if this was called `SORT_PLIST_INTOBJ` or `SortPlistOfImmediateInts` or something, as that makes it easier to understand (and trust) the code calling it...
No need for it to verify that the plist actually meets the criterion, though: the documentation of the function can just say that it expects this kind of input, and if somebody use it on something else: tough luck

ChrisJefferson:
Just one comment to @stevelinton : I am not currently making this function visible to users. I suspect (we could benchmark) the speedup over just sorting, if all you want to do is sort small integers, isn't worth it.

Now, providing a sort by raw obj ptr to users might be separately useful.

PR #2174, opened 2018-02-09
kernel: overhaul OnSets & OnTuples for perms, pperms, trans

fingolfin:
The code for these functions used LEN_LIST but elsewhere was written in such
a way that it was only corrected it the input list were non-empty plain (and
possibly having other properties). Luckily, the only callers for the six
involved functions already enforced that.

Documented this requirement, and then changed the code to explicitly use
plist accessor functions. Also use SortPlistByRawObj resp. SortDensePlist
where it (now) makes sense.

### DIFF
--- a/src/permutat.c
+++ b/src/permutat.c
@@ -46,6 +46,7 @@
 #include <src/gapstate.h>
 #include <src/integer.h>
 #include <src/io.h>
+#include <src/listfunc.h>
 #include <src/lists.h>
 #include <src/opers.h>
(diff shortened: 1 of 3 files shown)


fingolfin:
Updated to also change `FuncOnPosIntSetsPartialPerm` and `FuncOnPosIntSetsTrans` in a similar fashion and to tweak `REMOVE_DUPS_PLIST_CYC` (now called `REMOVE_DUPS_PLIST_INTOBJ`).

PR #2175, opened 2018-02-10
Warning about missing TestFile in ValidatePackageInfo

olexandr-konovalov:
This PR will add a warning like this:
```
gap>  ValidatePackageInfo("pkg/Convex/PackageInfo.g");
#W  ******************************************************
#W  There is no 'TestFile' component. It is recommended to
#W  specify a short test (to run for no more than several minutes)
#W  which may be used to check that a package works as expected.
#W  For deposited packages, these tests are run regularly
#W  as a part of the standard GAP test suite.
#W  See '?TestPackage', '?Tests files for a GAP package'
#W  and also '?TestDirectory' for more information
#W  ******************************************************
true
```

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -2270,6 +2270,14 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     TestOption( record, "BannerString", IsString, "a string" );
     TestOption( record, "TestFile", IsFilename,
                 "a string denoting a relative path to a readable file" );
+    if not IsBound( record.TestFile ) then
+      Print(
+        "#W  ************************************************************\n",
+        "#W  No 'TestFile' is given but is highly recommended. Package\n",
+        "#W  tests are run for deposited packages as part of the standard\n",
+        "#W  GAP test suite. See '?TestPackage' for more information.\n",
+        "#W  ************************************************************\n" );
+    fi;
     TestOption( record, "PreloadFile", IsFilename,
                 "a string denoting a relative path to a readable file" );
     TestOption( record, "Keywords", IsStringList, "a list of strings" );
--- a/tst/testinstall/package.tst
+++ b/tst/testinstall/package.tst
@@ -141,6 +141,11 @@ ps:// or ftp://
  https:// or ftp://
 #E  component `PackageDoc' must be bound to a record or a list of records
 #E  component `AvailabilityTest' must be bound to a function
+#W  ************************************************************
+#W  No 'TestFile' is given but is highly recommended. Package
+#W  tests are run for deposited packages as part of the standard
+#W  GAP test suite. See '?TestPackage' for more information.
+#W  ************************************************************
 false
 gap> info := rec(
 >     PackageName := "pkg",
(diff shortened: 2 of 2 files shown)


ChrisJefferson:
One comment on this, I don't think it should be in 4.9.1. If package authors have already updated their packages for 4.9 (I have), it's unfair to start making more requirements for 4.9 in my opinion.

This is fine for master for me, although the message does seem very long, nothing else in `ValidatePackageInfo` has such a long message?

ChrisJefferson:
How about this:

```
#W  ******************************************************
#W  No 'TestFile' given. Package tests are run for deposited
#W  packages as part of the standard GAP test suite.
#W  See '?TestPackage' for more information
#W  ******************************************************
```

On my GAP, `?Tests files for a GAP package` just produces an error?

frankluebeck:
So far, the `TestFile` entry is clearly optional.  Therefore I think this warning is not appropriate.

alex-konovalov:
I want to gradually move to making `TestFile` mandatory (for newly deposited packages, that must be already the case). 

As the first step towards this, I want to "advertise" this functionality more often, and make the warning text enough informative so that it contains all details necessary to start. 

I think that a simple warning "no test file" will more likely be ignored, since it does not explain where to find the documentation and examples.

alex-konovalov:
I've updated the message rephrasing a bit earlier @ChrisJefferson's text 
```
#W  ******************************************************
#W  No 'TestFile' is given but is highly recommended. Package
#W  tests are run for deposited packages as part of the standard
#W  GAP test suite. See '?TestPackage' for more information.
#W  ******************************************************
```

lib/package.gi
```diff
@@ -2259,6 +2259,13 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     TestOption( record, "BannerString", IsString, "a string" );
     TestOption( record, "TestFile", IsFilename,
                 "a string denoting a relative path to a readable file" );
+    if not IsBound( record.TestFile ) then
+          Print( "#W  ******************************************************\n",
+                 "#W  No 'TestFile' is given but is highly recommended. Package\n",
```
fingolfin: Shouldn't the message be wrapped to be not wider than row of asterisks ?

alex-konovalov: I've extended the row of asterisks instead.

alex-konovalov:
Oh wait, I need to fix `tst/testinstall/package.tst`...

fingolfin:
If this is really meant to go into 4.9.2, as indicated by the milestone, then the backport label should be set.

lib/package.gi
```diff
@@ -2270,6 +2270,14 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     TestOption( record, "BannerString", IsString, "a string" );
     TestOption( record, "TestFile", IsFilename,
                 "a string denoting a relative path to a readable file" );
+    if not IsBound( record.TestFile ) then
```
stevelinton: It would be more consistent with the style of the rest of the function to define something like `TestRecommended` 
which could uniformly provide feedback on aspects of a `PackageInfo.g`file which are recommended, but not mandatory. Optional arguments (or an info class) could be used to control whether this feedback was printed.

alex-konovalov: `TestRecommended` is a good idea. For example, one could detect if the package is hosted on GitHub and in this case recommend to add fields for source code repository and issue tracker. I will revisit this.

fingolfin:
This won't be in GAP 4.10, so I changed the milestone to 4.11. It needs to be rebased to resolve conflicts. Also, @stevelinton has requested changes. @alex-konovalov do you plan to revise this?

alex-konovalov:
@alex-konovalov yes, I plan to revise this and add `TestRecommended` (and maybe another similar check to test for obsolete components). I am happy for this to be moved to 4.11.

fingolfin:
No activity for 6 years. But also less need, basically all packages in the distro have this now.

PR #2180, opened 2018-02-14
Remove two package specific tests from testextra

fingolfin:
One belongs into `primgrp`, see https://github.com/gap-packages/primgrp/pull/14. The other belongs into `grpconst`, where I already added it (in an internal repo), hopefully a release of grpconst with this and other improvements will be made in early march.


fingolfin:
Why would we put this into GAP 4.9.x, though?

alex-konovalov:
@fingolfin because they will appear in packages in GAP 4.9.1, isn't it?

PR #1883, opened 2017-11-10
Start using bool,true,false from C99's stdbool.h

fingolfin:
I think it makes code more readable to explicitly use type `bool` and values `true`/`false` instead of generic `int` resp. `Int`, and `1`/`0` as return values. This PR is a first step towards this.

It can certainly wait till until GAP 4.9 has been branched. It should also wait till after superfail is removed, as that will conflict with this PR (in a trivial, easy to resolve way, so I am happy to do it here).

The main reason I am putting this out now is that I thought about this on multiple occasions (most recently when looking at `CHAR_PUSHBACK_EMPTY` in PR #1882), and I wanted to get it out here for discussion, to see what others think about that change.

If we merge it, there are tons of other places that could use `bool` (and which we could gradually and over time adapt to it). If we reject it, then I don't need to waste more time thinking about this possibility :-).

PR #1571, opened 2017-08-07
Reduce immutable/mutable diffs (UPDATE: now ready for merge)

fingolfin:
UPDATE: I believe this could be merged now. Also, I submitted issue #2325 for the general (im)mutability plan.

This PR builds extends PR #1570, and again is motivated by PR #1563: It unifies some code to reduce difference between mutable and immutable cases. The number of word occurrences of `IMMUTABLE`, as counted by `git grep -w IMMUTABLE src | wc -l` goes slightly down from 1328 to 565. But once one runs the script below, it goes down to 309, i.e. a quarter of what we started with.

This now makes it much easier to decide how to deal with the rest. Some immediate observations on those:
* [x] adding a `NEW_IMM_PLIST` function (possibly with a different name) would take care of quite a few of those occurences (and this function could immediately set the new `OBJ_FLAG_IMMUTABLE`)
* [ ] a lot of code deals with creating a new object with mutability identical to another object; could have a function for that, too
* [ ] but in a few cases, the mutability depends on a more complex logic... also, there are a bunch of `RetypeBag` calls to change immutability... it might be useful to have a faster variant of `MakeImmutable`, which is only useful for builtin lists and record, and directly modifies the TNUM (now) resp. object flag (with PR #1563)
* [ ] for HPC-GAP, `MakeBagTypePublic` cannot be used to automatically migrate immutable list and record bags to the public region. No big deal, though (and I actually like that we can thus remove some logic from `RetypeBag`)
* [ ] `IsMutableObjFuncs` likely can and should be removed once we get rid of the `IMMUTABLE` bit in the TNUM

Here is the script I mentioned above:

```gap
files:=[];
for dir in [ "src", "src/hpc" ] do
    dir:=Directory(dir);
    fs:=DirectoryContents(dir);
    fs:=Filtered(fs, f->EndsWith(f,".c"));
    Append(files, List(fs, f -> Filename(dir,f)));
od;


for f in files do
    Print("Processing ", f, "\n");
    data:=StringFile(f);
    lines:=SplitString(data,"\n");
    i := 2;
    while i <= Length(lines) do
        if PositionSublist(lines[i], "IMMUTABLE") = fail then
            i := i + 1;
            continue;
        fi;
        str := ReplacedString(lines[i], "+IMMUTABLE", "          ");
        if str = lines[i-1] then
            Remove(lines, i);
        fi;
        i := i + 1;
    od;
    out:=JoinStringsWithSeparator(lines, "\n");
    Add(out,'\n');
[...]

PR #2182, opened 2018-02-15
kernel: add NEW_PLIST_IMM and NEW_PLIST_MUT helpers

fingolfin:
This PR adds two `NEW_PLIST` wrapper functions: `NEW_PLIST_IMM` has the same signature as `NEW_PLIST`, but ensures that the resulting plist is immutable, currently by adjusting the TNUM (i.e. OR-ing it with `IMMUTABLE`), in the future it might instead set or clear a suitable object flag.

The second helper, `NEW_PLIST_MUT`, takes an additional boolean (0 or 1) as its first argument `mut`, which indicates whether the resulting plist should be mutable (`mut=1`) or immutable (`mut=0`). This is used in code where a new plist is created whose mutability depends on some external factors, e.g. it should match that of another object.

(Oh yeah: This PR contains some code from PR #1571)

One obvious bad thing about this PR is that the naming of the two helpers is somewhat in conflict: Based on the names, one could rightfully expect the second helper `NEW_PLIST_MUT` to be just like the first, except that it always forces a mutable plist, but it isn't. I don't expect us to merge the PR with this conflict, though; but there are several ways to resolve them, and I though I'd solicit some feedback before wasting time on implementing one of them:

1. We could rename the second helper to e.g. `NEW_PLIST_WITH_MUTABILITY` -- but I'd kind of prefer to keep a short name.
2. We could get rid of the first helper, and replace  `NEW_PLIST_IMM(tnum, len)` invocations by `NEW_PLIST_MUT(0, tnum, len)`. Does this have drawbacks? (perhaps that one might not immediately be able to guess that the second code is creating an immutable plist; also, one could accidentally permute the arguments, a possible source of errors -- and one of the reasons why I really would have liked to introduce a `BOOL` type, see #1883, but it seems nobody either likes or cares about this approach).
3. Rename the first helper, to... ?
4. Others?

I am kind of tempted to just go with approach 2, it is trivial to implement with a quick `sed` invocation. The main reason I did not yet do it is that reverting it again *would* be work, so I'd rather first get feedback.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1528,7 +1528,7 @@ Obj FuncLIST_BLIST (
     n = SizeBlist(blist);
 
     /* make the sublist (we now know its size exactly)                    */
-    sub = NEW_PLIST( IS_MUTABLE_OBJ(list) ? T_PLIST : T_PLIST+IMMUTABLE, n );
+    sub = NEW_PLIST_WITH_MUTABILITY( IS_MUTABLE_OBJ(list), T_PLIST, n );
     SET_LEN_PLIST( sub, n );
 
     /* loop over the boolean list and stuff elements into <sub>            */
--- a/src/calls.c
+++ b/src/calls.c
@@ -1548,7 +1548,7 @@ Obj FuncCALL_FUNC_LIST_WRAP (
 
     if (retval == 0)
     {
-        retlist = NEW_PLIST(T_PLIST_EMPTY + IMMUTABLE, 0);
+        retlist = NEW_PLIST_IMM(T_PLIST_EMPTY, 0);
     }
     else
     {
--- a/src/code.c
+++ b/src/code.c
@@ -3375,7 +3375,7 @@ static Int InitLibrary (
 #ifdef HPCGAP
     cache = NewAtomicList(T_ALIST, 1);
 #else
-    cache = NEW_PLIST(T_PLIST+IMMUTABLE, 1000L);
+    cache = NEW_PLIST_IMM(T_PLIST, 1000L);
     SET_LEN_PLIST(cache,0);
 #endif
     EAGER_FLOAT_LITERAL_CACHE = cache;
--- a/src/finfield.c
+++ b/src/finfield.c
@@ -1781,7 +1781,7 @@ Obj INT_FF (
     if ( LEN_PLIST(IntFF) < ff || ELM_PLIST(IntFF,ff) == 0 ) {
         q = SIZE_FF( ff );
         p = CHAR_FF( ff );
-        conv = NEW_PLIST( T_PLIST+IMMUTABLE, p-1 );
+        conv = NEW_PLIST_IMM( T_PLIST, p-1 );
         succ = SUCC_FF( ff );
         SET_LEN_PLIST( conv, p-1 );
         z = 1;
--- a/src/gap.c
+++ b/src/gap.c
@@ -1061,7 +1061,7 @@ Obj FuncCALL_WITH_CATCH( Obj self, Obj func, volatile Obj args )
     currStat = STATE(CurrStat);
     recursionDepth = GetRecursionDepth();
     tilde = STATE(Tilde);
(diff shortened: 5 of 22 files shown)


ChrisJefferson:
I do prefer the IMM version, as making immutable lists is fairly common and nice to be able to do simply.

ChrisJefferson:
New_plist_maybe_imm ?

markuspf:
I don't really like the `NEW_PLIST_MUT` approach, with the first parameter being mutability, because as you say it's not very readable. One could address this by using `BOOL`, but then again what is `true` in that position supposed to mean?

I think I prefer the explicit-but-long-name variant `NEW_PLIST_WITH_MUTABILITY`

fingolfin:
@markuspf wrote:
> I don't really like the NEW_PLIST_MUT approach, with the first parameter being mutability [...]
Could you turn this into something more constructive? First off: Is this a general objection to having a function which creates an object and sets its mutability in one go? Or just to this particular variant, so e.g.the position of the argument?

If this is just about the position of the argument, then my reply would be: well, putting it first made it by far easiest to convert the code; of course we could refactor, but I see no major gain, this feel like bike shedding.

If it is in general about having an argument for setting mutability: OK, so what alternative(s) would you prefer? The only one I can think of is to add a second like, like
```gap
  list =  NEW_PLIST( T_PLIST, len );
  SET_PLIST_MUTABILITY( list, mutable );
```
which of course would work, I guess (esp. when/if we move mutability from the TNUM into an OBJ_FLAG), but is slightly less efficient than the current code in this PR (as the new object is briefly mutable, and changing it to immutable requires touching  the tnum field, and also the region of the object in HPC-GAP.

So speaking from a purely pragmatic POV, having `NEW_PLIST_MUT` (or `NEW_PLIST_WITH_MUTABILITY`) seem both easier to do (I mean... I already did it), and also potentially more efficient. The only arguments against I understood so far seem to be purely subjective ("don't like it"), which is impossible to argue against :-(.

So right now, I see these options, but am open to concrete, constructive suggestions for furhter:
1. we merge this as-is
2. rename `NEW_PLIST_MUT` to `NEW_PLIST_WITH_MUTABILITY`, then merge
3. discard this PR, and resolve those `IMMUTABLE` flags in a different way.

I can certainly live with 3, too, but it'd be a bit sad, because I think this PR is a nice step-stone towards an (IM)MUTABLE object flag, in the sense that the main goal of this PR is to reduce the number of places that would have to be touched if / when we w replace the `tnum | IMMUTABLE` approach.

fingolfin:
BTW, small stat: I did, `git grep -w IMMUTABLE src | wc -l`, and got:
* on master: 568 hits
* in this PR: 481 hits (i.e. 87 less)
* on master, after running the script from PR #1571: 324
* in this PR, after running the script from PR #1571: 237
* with PR #1571 (which contains this PR) plus running the script:  223

Granted, there are several dozen more lines containing `IMMUTABLE` which can be simply be deleted if we remove `IMMUTABLE`, but still, this PR is getting rid of roughly 1/5 to 1/4 of all the `IMMUTABLE` occurrences. And of course nothing stops us from further refactoring later (and replacing `NEW_PLIST_MUT` resp. `NEW_PLIST_WITH_MUTABILITY` with something better then)

ChrisJefferson:
I think (@markuspf can disagaree) that we are all happy with the current PR, except for the naming of NEW_PLIST_MUT, which just looks too symmetric with NEW_PLIST_IMM (which I do like the name of)

markuspf:
Yes, I agree with what @ChrisJefferson says: I am happy with the PR, *except* for the naming of `NEW_PLIST_MUT` (in conjunction with having `NEW_PLIST_IMM` *and* the "obscure" parameter).

The most constructive I came up with was renaming `NEW_PLIST_MUT` to `NEW_PLIST_WITH_MUTABILITY` (i.e. following one of your less preferred suggestions).

fingolfin:
Rebased and changed to use `NEW_PLIST_WITH_MUTABILITY`

PR #2185, opened 2018-02-15
Add IsCentral methods where second argument is an element

### LINKED ISSUE #1795: Missing methods for IsCentral
I've got an example from a user:

```
gap> M:=DihedralGroup(16);
<pc group of size 16 with 4 generators>
gap> e:=Elements(M)[4];
f3
gap> IsCentral(M,e);
Error, no 1st choice method found for `IsCentral' on 2 arguments called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
called from read-eval loop at line 23 of *stdin*
you can 'return;' to continue
brk>
```
Indeed, the manual says that in `IsCentral( M, obj )` the 2nd argument must either be an element or a magma, but there seem no methods installed for this case at all:
```
brk> ShowMethods();
#I  Searching Method for IsCentral with 2 arguments:
#I  Total: 9 entries
#I  Method 1: ``IsCentral: for two associative FLMLORs-with-one'', value: 86
#I  Method 2: ``IsCentral: for two associative FLMLORs'', value: 76
#I  Method 3: ``IsCentral: for two associative rings-with-one'', value: 70
#I  Method 4: ``IsCentral: generic method for two groups'', value: 70
#I  Method 5: ``IsCentral: for two FLMLORs'', value: 68
#I  Method 6: ``IsCentral: for two associative rings'', value: 60
#I  Method 7: ``IsCentral: for two magmas-with-inverses'', value: 34
#I  Method 8: ``IsCentral: for two magmas-with-one'', value: 28
#I  Method 9: ``IsCentral: for two magmas'', value: 22
```

I've advised to use `e in Centre(M)` instead, but I think that a missing method should be added as well. Moreover, `Centre` is defined as an attribute for magmas.

fingolfin:
The documentation for `IsCentral` already claims these methods exist, but they never did. Add them in a systematic fashion.

Resolves #1795

### DIFF
--- a/lib/algebra.gi
+++ b/lib/algebra.gi
@@ -3280,6 +3280,28 @@ InstallMethod( IsCentral,
     IsCentralFromGenerators( GeneratorsOfAlgebraWithOne,
                              GeneratorsOfAlgebraWithOne ) );
 
+#############################################################################
+##
+#O  IsCentral( <A>, <x> )  . . . . . . . .  test if <x> is centralized by <A>
+##
+InstallMethod( IsCentral,
+    "for an FLMLOR and an element",
+    IsCollsElms,
+    [ IsFLMLOR, IsObject ],
+    IsCentralElementFromGenerators( GeneratorsOfLeftModule ) );
+
+InstallMethod( IsCentral,
+    "for an associative FLMLOR and an element",
+    IsCollsElms,
+    [ IsFLMLOR and IsAssociative, IsObject ],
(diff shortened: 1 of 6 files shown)


markuspf:
We should probably add some tests for them as well.

fingolfin:
Yeah but I was too tired to add those now, too. I'll revisit this PR later (but in the meantime, didn't want to loose these changes in my stack)

fingolfin:
I added some tests. Let's see how Travis and codecov like them.

PR #2186, opened 2018-02-15
Slightly improved EvalString

markuspf:
This version of EvalString is mostly backwards compatible, but does not use any hacks
with temporary variables, and allows for slightly better error catching.

### DIFF
--- a/lib/string.gi
+++ b/lib/string.gi
@@ -642,23 +642,21 @@ InstallGlobalFunction("RemoveCharacters", REMOVE_CHARACTERS);
 ##
 #F  EvalString( <expr> ) . . . . . . . . . . . . evaluate a string expression
 ##
-_EVALSTRINGTMP := 0;
-InstallGlobalFunction("EvalString", function( s )
-  local a, f, res;
-  a := "_EVALSTRINGTMP:=";
-  Append(a, s);
-  Add(a, ';');
-  Unbind(_EVALSTRINGTMP);
-  f := InputTextString(a);
-  Read(f);
-  if not IsBound(_EVALSTRINGTMP) then
-    Error("Could not evaluate string.\n");
-  fi;
-  res := _EVALSTRINGTMP;
-  Unbind(_EVALSTRINGTMP);
-  return res;
+InstallGlobalFunction("EvalString",
+function(string)
+    local instream, instring, res;
+    instring := Concatenation(string, ";");
+    instream := InputTextString(instring);
+
+    res := READ_ALL_COMMANDS(instream, false);
+    if Length(res) > 0 then
+       if res[1][1] = false then
+           Error("Could not evaluate string.\n"); 
+       elif IsBound(res[1][2]) then
+           return res[1][2];
+       fi;
+    fi;
 end);
-Unbind(_EVALSTRINGTMP);
 
 #############################################################################
 ##
--- a/tst/testinstall/evalstring.tst
+++ b/tst/testinstall/evalstring.tst
@@ -1,0 +1,24 @@
+gap> START_TEST("evalstring.tst");
+
+#
+gap> EvalString("");
+
+#
+gap> f := function() local x; return x; end;;
+gap> EvalString("f();");
+Error, Variable: 'x' must have an assigned value
+Error, Could not evaluate string.
+
+
+#
+gap> EvalString("x := 15;");
+15
+
+#
+gap> EvalString("Print(\"Hello, world\\n\");");
+Hello, world
+gap> EvalString("15 + 1; Print(\"Hello, world\");");
+Hello, world16
+
+#
+gap> STOP_TEST( "evalstring.tst", 1);


lib/string.gi
```diff
@@ -642,23 +642,23 @@ InstallGlobalFunction("RemoveCharacters", REMOVE_CHARACTERS);
...
+InstallGlobalFunction("EvalString",
+function(string)
+    local instream, outstream, instring, outstring, res;
+    instring := Concatenation(string, ";");
+    outstring := "";
+    instream := InputTextString(string);
+    outstream := OutputTextString(outstring, true);
+
+    SET_OUTPUT(outstream, true);
+    res := READ_ALL_COMMANDS(instream, false);
```
fingolfin: Can't you keep calling `Read(instream)` here? That might get rid of at least some of the test regressions.

markuspf: `Read()` does not return the value of an expression (which is the whole reason why this weird `_EVALRESULT` hack was there, I presume.

 I plan to replace the `READ_ALL_COMMANDS` by a `READ_ONE_COMMAND` from a stream soon (some work in progress is here: https://github.com/markuspf/gap/tree/inout-rewrite).

(Disclaimer: This is still part of libgap work, once the inout bit is done, we have incorporated every `libgap` bit and there'll be a PR to libgap in `master` proper)

ChrisJefferson:
Unfortunately, this breaks failure in EvalString. Current master:

```
gap> f := function() local x; return x; end;             
function(  ) ... end                                     
gap> EvalString("f()");                                  
Error, Variable: 'x' must have an assigned value in      
  return x; at *stdin*:1 called from                     
<function "f">( <arguments> )                            
 called from read-eval loop at stream:1                  
brk>                                                     
gap>                                                     
gap>                                                     
```

With this PR once I leave the break loop, I can't see anything GAP is printing, as SET_PREVIOUS_OUTPUT is never called.

markuspf:
Lets just delete things that I thought were a good idea originally until it works again ;)

frankluebeck:
Hm, the proposed new function has a different functionality, and not what is documented for `EvalString`.
It is more a "ReadString".  What is the new documentation you have in mind for `EvalString`?

ChrisJefferson:
@frankluebeck : I think the behaviour is the same. What do you think has changed?

There is some weird things, like the fact it runs all the commands in the string and returns the value of the first one, but that is how `EvalString` has always worked.

tst/testinstall/evalstring.tst
```diff
@@ -0,0 +1,24 @@
...
+
+#
+gap> f := function() local x; return x; end;;
+gap> EvalString("f();");
+Error, Variable: 'x' must have an assigned value
+Error, Could not evaluate string.
+
+
+#
+gap> EvalString("x := 15;");
```
fingolfin: This actually fails with the old `EvalString`. Perhaps this change in behavior is intentional, but I did no see any mention of it before. Same for the next example -- the old `EvalString` expected the first "statement" in the string to be an expression, as can be guessed from the name `expr` given to its argument string.

fingolfin: (`expr` is used in the documentation, at least, though apparently not in the implementation)

frankluebeck:
@ChrisJefferson All the examples in the test file are strange and not what this function was made for:
```
EvalString("x := 15;");
EvalString("Print(\"Hello, world\");15+1;");
EvalString("Print(\"Hello, world\\n\");");
```
The third should in my opinion run into an error. These are not expression one would evaluate.
The function is meant (and I think documented) to be used like in
```
EvalString("15+1");
EvalString("f(15)-12/5");
EvalString("ConjugacyClasses(Group((2,3,4)))");
```
I agree that there was no proper checking of the input before and strange things may happen if you call the function with a string that is not an expression that can be evaluated (and I didn't know how a good check could be done).

lib/string.gi
```diff
@@ -642,23 +642,21 @@ InstallGlobalFunction("RemoveCharacters", REMOVE_CHARACTERS);
...
-  res := _EVALSTRINGTMP;
-  Unbind(_EVALSTRINGTMP);
-  return res;
+InstallGlobalFunction("EvalString",
+function(string)
+    local instream, instring, res;
+    instring := Concatenation(string, ";");
+    instream := InputTextString(instring);
+
+    res := READ_ALL_COMMANDS(instream, false);
```
fingolfin: An alternative implementation, which would not break compatibility, could perhaps be written by using `ReadAsFunction`: just prepend `return ` to the string to be evaluated; then use `ReadAsFunction`; an error there indicates a syntax error resp. a non-expression. Finally, execute the function. Only the first part of the string will be executed then (this would still be a difference compared to EvalString, but IMHO one that is an improvement, and does not contradict the existing documentation)

markuspf:
Closing this as its changing behaviour of existing code (and there's not much point in spending more time on bikeshedding in my opinion).

PR #2188, opened 2018-02-16
Tweak tst/testinstall/coll.tst to work with semigroups pkg

fingolfin:
An immediate method for IsTrivial provided by semigroups causes Magma(1) to
learn prematurely that it is trivial. Avoid this by using a magma (or, in the
end, monoid) generated by a singular matrix. Also, use Magma(0) instead of
Magma(1) in one other test.

On the long run, we might want to replace all uses of Magma() in this test
by custom ("mock") collection objects, as there is always a chance that a
(perfectly valid) new method in some package throws some of them off. But for
now, just using Magma() to create our tests objects is very convenient.

### DIFF
--- a/tst/testinstall/coll.tst
+++ b/tst/testinstall/coll.tst
@@ -45,14 +45,14 @@ gap> List(props, p -> p(M0));
 
 # ... immediate methods for a collection which knows its size,
 # applied to trivial (= size 1) collection
-gap> M1:=Magma(1);;
-gap> ForAll(props, prop -> not Tester(prop)(M1));
+gap> N1:=Magma([[[1,0],[0,0]]]);;
+gap> ForAll(props, prop -> not Tester(prop)(N1));
 true
-gap> Size(M1);
+gap> Size(N1);
 1
-gap> ForAll(props, prop -> Tester(prop)(M1));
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
It'd be great if this could be merged soon, currently all our travis builds are failing because of this.

alex-konovalov:
@fingolfin and @james-d-mitchell after this has been merged, in the master branch with all packages loaded I have just one diff instead of five:
```
########> Diff in /Volumes/hudson-fs/workspace/GAP-master-test/GAPCOPTS/32buil\
d/GAPTARGET/install/label/fruitloop/GAP-master-snapshot/tst/testinstall/coll.t\
st:71
# Input is:
List(props, p -> p(Magma(FamilyObj([1]), [])));
# Expected output:
[ true, false, true, true ]
# But found:
Error, List Element: <list>[1] must have an assigned value
########
```
Unless someone will get there faster than me, I will try look tomorrow and post a new issue, if needed.

fingolfin:
That issue was already reported simultaneously to this PR.

james-d-mitchell:
This is fixed by 

https://github.com/gap-packages/Semigroups/pull/460

and was already reported to Semigroups as:

https://github.com/gap-packages/Semigroups/pull/460

PR #2189, opened 2018-02-16
tests: also extract and test tutorial examples on Travis

fingolfin:


alex-konovalov:
@fingolfin advance warning: tutorial examples are adjusted to run without diffs with *default packages*. It could require some efforts to make them more robust, since for the tutorial we do want to *show* some outputs. In case you will see any diffs, please do not change tutorial examples - ensure that you load the default set of packages to the test instead.

PR #2196, opened 2018-02-19
always use memcpy to enforce alignment in getblock8

dimpase:
hopefully the compiler is clever enough to optimise this away

### DIFF
--- a/src/intfuncs.c
+++ b/src/intfuncs.c
@@ -310,7 +310,9 @@ static inline uint64_t rotl64 ( uint64_t x, int8_t r )
 
 FORCE_INLINE uint64_t getblock8 ( const uint64_t * p, int i )
 {
-  return p[i];
+  uint64_t val;
+  memcpy(&val, p + i, sizeof(uint64_t));
+  return val;
 }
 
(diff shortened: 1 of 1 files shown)


dimpase:
Should I amend the commit to include the paragraph suggested above?
Perhaps it's easier that whoever merges this adds this into the commit message.

fingolfin:
This comment really should be in the code, not the commit message (where it is invisible when reading the code)

dimpase:
> This comment really should be in the code
OK, done.

src/intfuncs.c
```diff
@@ -307,10 +307,18 @@ static inline uint64_t rotl64 ( uint64_t x, int8_t r )
 //-----------------------------------------------------------------------------
 // Block read - if your platform needs to do endian-swapping or can only
 // handle aligned reads, do the conversion here
+//
+// The pointer p may not be aligned, which means that directly writing to it can
```
ChrisJefferson: very picky, it's "reading it", not "writing to it"

dimpase: OK, fixed.

PR #2197, opened 2018-02-19
Add FileId to function calls in profiling

ChrisJefferson:
This backwards-compatible change makes it easier to track where functions are called from in profiles, which will be used in the next version of profiling.

Really, we shouldn't be outputting `File` at all here, I mostly moved to always using `FileId`, but I leave `File` here to not break old versions of the profiling package.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -17,6 +17,7 @@
 #include <src/gap.h>
 #include <src/hookintrprtr.h>
 #include <src/io.h>
+#include <src/lists.h>
 #include <src/plist.h>
 #include <src/stringobj.h>
 #include <src/vars.h>
@@ -149,6 +150,27 @@ struct ProfileState
 /* We keep this seperate as it is exported for use in other files */
 UInt profileState_Active;
 
+static inline void outputFilenameIdIfRequired(UInt id)
+{
+    if (id == 0) {
+        return;
+    }
+    if (LEN_PLIST(OutputtedFilenameList) < id ||
+        ELM_PLIST(OutputtedFilenameList, id) != True) {
+        AssPlist(OutputtedFilenameList, id, True);
+        fprintf(profileState.Stream,
+                "{\"Type\":\"S\",\"File\":\"%s\",\"FileId\":%d}\n",
+                CSTR_STRING(ELM_LIST(FilenameCache, id)), (int)id);
+    }
+}
+
+// This function checks gets the filenameId of the current function.
+static inline UInt getFilenameIdOfCurrentFunction(void)
+{
+    Obj func = CURR_FUNC();
+    Obj body = BODY_FUNC(func);
+    return GET_GAPNAMEID_BODY(body);
+}
 
 
 void HookedLineOutput(Obj func, char type)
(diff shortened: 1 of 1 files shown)


src/profile.c
```diff
@@ -246,11 +251,8 @@ static void fcloseMaybeCompressed(struct ProfileState* ps)
 ** as approriate, and then pass through to the true function
 */
 
-// This function checks if we have ever printed out the id of stat
-static inline UInt getFilenameId(Stat stat)
+static inline UInt getFilenameIdOfBody(Obj body)
```
fingolfin: The comment right before this function now fits there even less than it did before... move it?

fingolfin: Also, I find it really surprising that a function with this name has a major side-effect (it prints some JSON to `profileState.Stream` under certain conditions). It would be nice if there was at least a comment in there which addressed this. I'd find it even better if this printing did *not* happen in there, but instead was moved to the callers, resp. to a separate function `outputFileIdIfNecessary` or so.

Of course this problem predates this PR, so I am no expecting it to be resolved in this PR, I can also take care of it myself after this PR is merged.

src/profile.c
```diff
@@ -175,9 +177,12 @@ void HookedLineOutput(Obj func, char type)
               (int)profileState.lastNotOutputted.fileID);
     }
 
-    fprintf(profileState.Stream,
-            "{\"Type\":\"%c\",\"Fun\":\"%s\",\"Line\":%d,\"EndLine\":%d,\"File\":\"%s\"}\n",
-            type, name_c, (int)startline, (int)endline, filename_c);
+    // We output 'File' here for backwards compatibility
```
fingolfin: "backwards compatibility" with what? Older versions of the `profiling` package that need / expect this?

Perhaps also explain what the replacement is? Perhaps like this?
> Output  information on the function into the JSON stream. Note that the 'File' member is only printed for backwards compatibility with older versions of the profiling package; newer versions use the 'FileId' member instead.

Bonus points for saying which versions of profiling use 'FileId' (and do they still accept older files with 'File' but not 'FileId'?). Of course this may not yet be possible, as presumably no such version of profiling has yet been released, but perhaps you can simply decide on the planned version now and put it in anyway?

fingolfin:
Will the new profiling still be able to parse old JSON files, though? Or does one have to use an older profiling version to deal with those?

ChrisJefferson:
The new version of profiling will parse the old files, and the old version will parse the new files (that is why I added a redundant extra field, rather than replacing File with FileId).

If you use a new version of profiling with a new GAP, you will get links from a function back to the places where it was called. Otherwise, you won't.

src/profile.c
ChrisJefferson: Can do.

src/profile.c
ChrisJefferson: I'm happy to rename, but I quite like the printing happens in here, as it means you can't forget to print out the id before using it.

ChrisJefferson:
Hopefully fixed issues with code

fingolfin:
Backported in a092343ddc7375d7a6c022d8f933784092599812

Issue #1829, opened 2017-10-30
`ViewString` hits recursion depth trap for recursive lists


markuspf:
```
gap> ll := [];; ll[1] := ll;;
gap> ViewString(ll);
Error, recursion depth trap (5000) in
  str := ViewString( list[i] ); at ./lib/list.gi:354 called from 
ViewString( list[i] ) at ./lib/list.gi:354 called from
ViewString( list[i] ) at ./lib/list.gi:354 called from
ViewString( list[i] ) at ./lib/list.gi:354 called from
ViewString( list[i] ) at ./lib/list.gi:354 called from
ViewString( list[i] ) at ./lib/list.gi:354 called from
...  at *stdin*:2
you may 'return;'
brk> 
```

stevelinton:
This could be fixed easily enough by moving these methods into the kernel and duplicating what already exists for ViewObj etc. 

It could probably be fixed clunkily by keeping a list of already visited subobjects or their HANDLE_OBJ values and checking it with IsIdentical or by comparing handles in GAP code cf MEMORY_USAGE.

It would be a bit cleaner with ObjHash -- is that in non-HPC GAP yet?

PR #2200, opened 2018-02-23
Add a simple Format function

ChrisJefferson:
This PR adds a simple method of formatting strings, strongly inspired by Python's "format" function. It is probably easiest to demonstrate by example:

```
gap> Format("I have {} cats and {} dogs", [3,2]);
"I have 3 cats and 2 dogs"
gap> Format("I have {2} cats and {1} dogs", [3,2]);
"I have 2 cats and 3 dogs"
gap> Format("I have {apple} cats and {pear} dogs", rec(apple:=3, pear:=2);
"I have 3 cats and 2 dogs"
```

There is also a `FormatTo` variant, which prints to a stream.

I've been using this for a while, as I find building these types of simple format strings is quite painful in GAP. There is some other stuff from Python (padding, formatting of floating points), I haven't implemented yet, and might, but I thought first I'd check if this function seemed generally interesting to people.


lib/string.gi
```diff
@@ -1258,9 +1258,86 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
     Append(s, units[shift+1]);
     return s;
 end);
-
-        
     
+InstallGlobalFunction(FormatTo, function(stream, s, args)
+    local pos, len, outstr, nextbrace, endbrace, id, argcounter;
```
markuspf: Is there a (good) reason why you don't use `args...`?

There is the option to pass a record, but that can be handled even in the presence of `args...`

lib/string.gi
```diff
@@ -1258,9 +1258,86 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
     Append(s, units[shift+1]);
     return s;
 end);
-
-        
     
+InstallGlobalFunction(FormatTo, function(stream, s, args)
```
markuspf: :+1:

lib/string.gi
```diff
@@ -1258,9 +1258,86 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+                AppendTo(stream, String(args[argcounter]));
+                argcounter := argcounter + 1;
+            elif Int(id) <> fail then
+                if not IsList(args) then
+                    ErrorNoReturn("data argument of Format must be a list when using {<integer>}");
+                fi;
+                if Int(id) < 1 or Int(id) > Length(args) then
+                    ErrorNoReturn("out of bounds in Format -- asked for {",Int(id),"} when there are only ",Length(args), " arguments");
+                fi;
+                AppendTo(stream, String(args[Int(id)]));
```
markuspf: `String` is problematic because not many objects support `String` and just return `<object>`, also one could argue that `String` builds up a buffer, whereas `PrintTo` or `AppendTo` let the object's implementor decide what to do.

why not just `args[Int(id)]`? 

There is of course the question whether you want to support `String`, `Display` and `View` as format strings...

markuspf:
I think we want this, but it might need some additional thought...

lib/string.gi
markuspf: See also https://github.com/gap-system/gap/issues/1829

ChrisJefferson: I'm happy to pick which we want (string, display or view), but if you just pass in strings then of course they'll be printeed fine (I could check `IsString` to avoid the extra call), but it seems a pain to make people call `String` if they just want to pass in say a permutation.

stevelinton: I'm not sure about DisplayString (they generally don't compose very well) but one way or another it could be useful to be able to include either the ViewString or the String of an object, depending on what you want the formatted string for. Could be an option encoded somehow in the format-string, or a second function FormatView (or some better name).

stevelinton: For consistency with names elsewhere the function should ideally be a noun describing the result, rather than a verb which suggests it modifies its argument. We have `FormattedString` but it's marked obsolete, so could possibly be resused.

lib/string.gi
ChrisJefferson: Good question. I started off with a list, but there isn't really a good reason for that. I'll switch to args...

lib/string.gi
fingolfin: Re String/Display/View: In a `printf` style function, I'd handle that with different placeholders in the format string, say:  `Format("%{display} vs. %{view} vs. %{string}", myObj, myObj, myObj);` to get a comparison of the three strings.

Looking at https://pyformat.info, I guess the "pythonic" way would be to use a "value conversion" suffix, like  `Format("{0!d} vs. {0!v}. vs {0!s}", myObj)`. If we default to using `String`, then there could also be something to force "raw" output, i.e. no conversion. Or conversely, we could default to *not* using `String` (as @markuspf suggested), but still allow people to enable it via a value conversion suffix.

frankluebeck:
I have never missed such a function but can see that  people find this useful. I'm only a bit concerned about its proposed name. I had a vague plan to propose "Format" as a name for some improved version of `String` which gets a second argument like '"Text"`, '*HTML"`, `"LaTeX"` (and maybe further arguments). Some version of this exists in Jean Michel's version of GAP3.

Would it be ok for you to change the name? Say, `ComplementedString`?

ChrisJefferson:
I am happy to change the name. I started with `Format` just because that is what Python uses, and I based my inspiration on that.

fingolfin:
While I don't mind `PrintFormatted`, I find the argument @frankluebeck raises as to why `Format` is problematic somewhat strange -- essentially it seems a case of "calling dibs" on a name?!? I.e. you seem to say that there is a possibility you might (or not) one day submit code which may perhaps want to use the name `Format` for another function?

That's not an argument I'd expect. Perhaps what is meant here is that "Jean Michel's version of GAP3" has a function `Format`, and the desire here is to have a similar function with the same name in GAP 4, which I'd find somewhat more compelling. Albeit not very much.

Moreover, the function in this PR seems to me (quite subjectively, of course) more useful to a wider audience than a function which produces HTML or LaTeX output (I can't really judge the latter, though, as I don't really know what it's supposed to be). So I'd rather have this here under the name `Format` than some function to produce HTML/LaTeX whatever which I likely would never use.

I am not so happy with the name `PrintFormatted`, because it suggests to me that output will be sent to stdout, not into a string. That is something I find a bit weird in the current PR, too, though; i.e. that `Formatted` returns a string, while `FormattedTo` prints into a stream. I'd find `FormattedString` (new name for `Format`), `PrintFormatted` (does not exist in this PR right now) and `PrintFormattedTo` (new name for `FormatTo`) more logical then. (This also matches an earlier suggestion by @stevelinton).

fingolfin:
Of course, as I just realize, *those* names kind of clash with GAPDoc's `PrintFormattedString` *sigh*

ChrisJefferson:
I've now gone for StringFormatted, PrintFormatted and PrintToFormatted (which are, hopefully obviously, designed to mirror String, Print and PrintTo, with a formatting option).

lib/string.gd
```diff
@@ -860,6 +860,73 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  <Func Name="PrintToFormatted" Arg='stream, string, data...'/>
+##
+##  <Description>
+##  These functions format an input string.
+##  <Ref Func="StringFormatted"/> returns the formatted string,
+##  <Ref Func="PrintFormatted"/> prints the formatted string and
+##  <Ref Func="PrintToFormatted"/> appends the formatted string to <A>stream</A>,
+##  which can be either an output stream or a filename.
+##  <P/>
+##  <A>data</A> is a list of values to printed, using the formatting rules below:
```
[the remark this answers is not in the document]
fingolfin: Also, I think this can be confusing: the user might think that the second (resp third, in case of `PrintToFormatted`) argument should be a list, containing the values. Perhaps like this:
> All arguments after `<A>string</A>` form the list `<A>data</A>` of values to be printed, ...

lib/string.gd
```diff
@@ -860,6 +860,73 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  The contents of <C>{ }</C> form a small language. The format is <C>{id!format}</C>,
+##  where both <C>id</C> and <C>format</C> are optional. If the <C>!</C> is ommitted, the bracket
+##  is treated as <C>{id}</C> with no <C>format</C>.
+##  <P/>
+##  <C>id</C> is interpreted as follows:
+##  <List>
+##  <Mark>An integer <C>i</C></Mark> <Item>
+##    Take the <C>i</C>th element of <A>data</A>.
+##  </Item>
+##  <Mark>No variable</Mark> <Item>
```
fingolfin: What does "No variable" mean here? Perhaps "Not given" or, better, "No id given" ?

lib/string.gd
```diff
@@ -860,6 +860,73 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##    Take the <C>j</C>th element of <A>data</A>, where <C>j</C> is the
+##    number of occurrences of <C>{}</C> in the string, including this one,
+##    up to this point.
```
fingolfin: Doesn't this also count `{!format}`?

lib/string.gd
```diff
@@ -860,6 +860,73 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  </Item>
+##  <Mark>A string <C>str</C></Mark> <Item>
+##    Take the <C>str</C> member of the first element of <A>data</A>, which 
+##    must be a record.
```
fingolfin: I like to have "definitions" before using them, in this case: saying that the first entry must be a record before makign use of it (by taking about its "str member"). How about this:
> If this is used, then the first element of data must be a record `r`; in this case, the value `r.(str)` is taken.

ChrisJefferson:
This is now (I think) fixed up.

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
 end);
-
-        
     
+InstallGlobalFunction(PrintToFormatted, function(stream, s, args...)
+    local pos, len, outstr, nextbrace, endbrace,
+          argcounter, var,
+          splitBracket, toprint;
+
+    splitBracket := function(string, startpos, endpos)
```
fingolfin: Perhaps add a brief comment saying what `splitBracket` does? It seems to parse a replacement field starting and ending at the given position in the string, is that correct? So perhaps `parseReplacementField` would be a better name?

Also note that the `string` argument could be replace, and instead `s` be accessed, as it is available in this closure.

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+      local posbang, type;
+      posbang := Position(s, '!', startpos-1);
+#      Print(startpos, ":",endpos, ":",posbang,"\n");
+      if posbang = fail or posbang > endpos then
+        posbang := endpos + 1;
+      fi;
+      type := string{[posbang + 1 .. endpos]};
```
fingolfin: In the documentation, you call this the "format", and now the "type". I'd recommend to use uniform language, to make it easier to match documentation and code.

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+      if type = "" then
```
fingolfin: Perhaps insert comment: "If no format/type is given, default to 's' as described in the documentation" ?

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+      fi;
+      return rec(id := string{[startpos..posbang-1]}, type := type);
+    end;
+      
+    argcounter := 1;
+    len := Length(s);
+    pos := 0;
+
+    if not (IsOutputStream(stream) or IsString(stream)) or not IsString(s) then
+        ErrorNoReturn("Usage: PrintToFormatted(<stream>, <string>, <args>...)");
```
fingolfin: Replace `args` by `data` to match documentation

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
     Append(s, units[shift+1]);
     return s;
 end);
-
-        
     
+InstallGlobalFunction(PrintToFormatted, function(stream, s, args...)
```
fingolfin: Replace `args` by `data` to match documentation? (Or vice versa, adjust the documentation)

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+        if endbrace = fail then
+            ErrorNoReturn("Invalid format string, no matching '}' at position ", nextbrace);
+        fi;
+
+        toprint := splitBracket(s, nextbrace+1,endbrace-1);
+
+
+        if toprint.id = "" then
+            if Length(args) < argcounter then
+                ErrorNoReturn("out of bounds in StringFormatted -- used ",argcounter," {} when there are only ",Length(args), " arguments");
```
fingolfin: change "{}" in this error message to "replacement fields without id"

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+        if toprint.type = "s" then
+          if not IsString(var) then
+            var := String(var);
+          fi;
+          AppendTo(stream, var);
+        elif toprint.type = "v" then
+          AppendTo(stream, ViewString(var));
+        elif toprint.type = "d" then
+          AppendTo(stream, DisplayString(var));
+        else ErrorNoReturn("Invalid formatting option: '", toprint.type, "'");
```
fingolfin: And a third name: after "format" and "type" this is now called "formatting option".

tst/testinstall/format.tst
```diff
@@ -0,0 +1,56 @@
...
+gap> r1 := SymmetricGroup([3..5]);;
+gap> r2 := AlternatingGroup([1,3,5]);;
+gap> r3 := AlternatingGroup([11,12,13]);;
+gap> StringFormatted("{1!s} {1!v} {1!d}", r1);
+"SymmetricGroup( [ 3 .. 5 ] ) Sym( [ 3 .. 5 ] ) <object>\n"
+gap> StringFormatted("{!s} {!v} {!d}", r1, r2, r3);
+"SymmetricGroup( [ 3 .. 5 ] ) Alt( [ 1, 3 .. 5 ] ) <object>\n"
+gap> StringFormatted("{a!s} {b!v} {c!d}", rec(a := r1, b := r2, c := r3));
+"SymmetricGroup( [ 3 .. 5 ] ) Alt( [ 1, 3 .. 5 ] ) <object>\n"
+gap> StringFormatted("{a!x}", rec(a := r1));
```
fingolfin: I find it difficult to read this test file, and to see what's there -- so if I wanted to add more tests, I'd struggle to decide where to do so.

Perhaps consider structuring this test file a bit by inserting some empty lines to separate tests into "logical" groups? (Well, empty lines followed by a comment, to suppress whitespace diffs in the test; like this:
```
gap> StringFormatted("{1}", rec());
"rec(  )"

# test format options
gap> r1 := SymmetricGroup([3..5]);;
gap> r2 := AlternatingGroup([1,3,5]);;
```

lib/string.gi
```diff
@@ -1041,7 +1041,7 @@ InstallGlobalFunction(PrintCSV,function(arg)
 end);
 
 
-# Format commands
+# StringFormatted commands
```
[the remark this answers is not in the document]
ChrisJefferson: Woops, that's just me search+replacing Format -> StringFormatted, and not being careful about what I hit!

lib/string.gi
```diff
@@ -1258,9 +1258,121 @@ InstallGlobalFunction(StringOfMemoryAmount, function(m)
...
+            if endbrace + 1 <= len and s[endbrace + 1] = '}' then
+                AppendTo(stream, s{[pos+1..endbrace]});
+                pos := endbrace + 1;
+                endbrace := Position(s, '}', pos);
+            else
+                ErrorNoReturn("Mismatched '}' at position ",endbrace);
+            fi;
+        od;
+
+        if nextbrace = fail then
```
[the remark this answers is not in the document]
ChrisJefferson: I wasn't planning on being this clever, I obviously just didn't think about some cases well enough, then the tests passed.

lib/string.gd
```diff
@@ -860,6 +860,81 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  <Mark>An integer <C>i</C></Mark> <Item>
+##    Take the <C>i</C>th element of <A>data</A>.
+##  </Item>
+##  <Mark>A string <C>str</C></Mark> <Item>
+##    If this is used, the first element of <A>data</A> must be a record <C>r</C>.
+##    In this case, the value <C>r.(str)</C> is taken.
+##  </Item>
+##  <Mark>No id given</Mark> <Item>
+##    Take the <C>j</C>th element of <A>data</A>, where <C>j</C> is the
+##    number of occurrences of <C>{}</C> in the string with no id given,
```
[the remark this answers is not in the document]
ChrisJefferson: Oh, I mis-interpreted what python did. I think I'll switch to this, as it seems simpler.

lib/string.gd
```diff
@@ -860,6 +860,85 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##    Take the <C>j</C>th element of <A>data</A>, where <C>j</C> is the
+##    number of replacement fields with no id in the format string so far.
+##    If any replacement field has no id, then all replacement fields must
+##    have no id.
+##  </Item>
+##  </List>
+##
+##  A single brace can be outputted by doubling, so <C>{{</C> in the format string
+##  produces <C>{</C> and <C>}}</C> produces <C>}</C>.
+##  </P>
```
fingolfin: syntax error, change `</P>` to `<P/>`

lib/string.gd
```diff
@@ -860,6 +860,85 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  <#GAPDoc Label="StringFormatted">
+##  <ManSection>
+##  <Func Name="StringFormatted" Arg='string, data...'/>
+##  <Func Name="PrintFormatted" Arg='string, data...'/>
+##  <Func Name="PrintToFormatted" Arg='stream, string, data...'/>
+##
+##  <Description>
+##  These functions perform a string formatting operation.
+##  They accept a format string, which can contain replacement fields
+##  which are delimated by braces {}.
```
fingolfin: delimated -> delimited

lib/string.gd
```diff
@@ -860,6 +860,85 @@ BindGlobal("BHINT", MakeImmutable("\>\<"));
...
+##  <P/>
+##  There are three formatting functions, which differ only in how they
+##  output the formatted string.
+##  <Ref Func="StringFormatted"/> returns the formatted string,
+##  <Ref Func="PrintFormatted"/> prints the formatted string and
+##  <Ref Func="PrintToFormatted"/> appends the formatted string to <A>stream</A>,
+##  which can be either an output stream or a filename.
+##  <P/>
+##  The arguments after <A>string</A> form a list <A>data</A> of the values to be
+##  inserted into <A>string</A>, using the following formatting rules:
```
fingolfin: Nitpick: the values are not inserted into `string` (which indeed can be immutable). Perhaps 
>  ... form a list `data` of values used to substitute the replacement fields in `string`, according to the following formatting rules

PR #2202, opened 2018-02-23
hpcgap: get rid of backquote alias for MakeImmutable

fingolfin:
The backquote / backtick syntax always was controversial, and with the whole unification project, we already got rid of most of its uses. This PR removes the last few remaining uses, and the implementation.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1737,19 +1737,6 @@ void ReadLiteral (
         ReadRecExpr( follow );
         break;
 
-#ifdef HPCGAP
-    /* `Literal                                                            */
-    case S_BACKQUOTE:
-        Match( S_BACKQUOTE, "`", follow );
-        TRY_READ {
-          IntrRefGVar(GVarName("MakeImmutable"));
-          IntrFuncCallBegin();
-        }
-        ReadAtom( follow, 'r' );
-        TRY_READ { IntrFuncCallEnd(1, 0, 1); }
-        break;
-#endif
-
     /* <Function>                                                          */
     case S_FUNCTION:
     case S_ATOMIC:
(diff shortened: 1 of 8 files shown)


src/read.c
```diff
@@ -1737,19 +1737,6 @@ void ReadLiteral (
...
-        Match( S_BACKQUOTE, "`", follow );
-        TRY_READ {
-          IntrRefGVar(GVarName("MakeImmutable"));
-          IntrFuncCallBegin();
-        }
-        ReadAtom( follow, 'r' );
-        TRY_READ { IntrFuncCallEnd(1, 0, 1); }
-        break;
-#endif
-
```
ChrisJefferson: My only thought, is there enough hpc-gap code out there that we should put a SyntaxError here telling people that ` should be rewritten as MakeImmutable?

ChrisJefferson:
Two things, as I mention in my review I wonder if we should make a SyntaxError just to let people know what's going on if they load old HPC-GAP demos.

Also, we should cherry-pick this into 4.9 after merging I think, to stop people using backtick in 4.9 releases.

fingolfin:
I guess a cherry pick into stable-4.9 is a good idea.

As to the `SyntaxError` -- hrm, that's more work that later also need to be undone, i.e. even more work. Do we really need this? As it is, there are not that many people using HPC-GAP, most of which are in close contact with us anyway. And I am pretty hopeful that they would either guess the reason, or else contact us via GAP support or the issue tracker for help, once they start to get syntax errors caused by backticks, like this one:
```
gap> `"123";
Syntax error: expression expected
`"123";
^
```

We'd also put this change in the CHANGES list for GAP 4.9.1, then the few people who don't take regularly to us can find out about the backtick there, too.

fingolfin:
Backported to stable-4.9 in commit c5b56a8c8a70f128cf54bafcfd1a57713fcaaee6

PR #2203, opened 2018-02-23
Refactor plist types handling

fingolfin:
Right now, the types of homogeneous plists are computed via the library function TYPE_LIST_HOM, which hard codes information about plist TNUMs but in a way that is not easy to find if one does not know about it. In particular, it encodes assumptions about the order of various TNUMs relative to each other; and also the assumption that (im)mutability can be determined from the parity of the TNUM. Since some of us are working towards getting rid of the latter, it seems prudent to get rid of this assumption from the library, which this PR does.

In addition, a HPC-GAP check for writeability to `TYPES_LIST_FAM` is removed. I assume that once made sense, but `TYPES_LIST_FAM` now always is a strict write-once atomic list, and we take care of potential races in the new `TypePlistHomHelper`.

Finally, an unrelated change reintroduces `UNCLONEABLE_TNUMS` -- now that we export the TNUMs as constants to GAP, IMHO the intent of the code is easier to understand with the use of `UNCLONEABLE_TNUMS`.

### DIFF
--- a/hpcgap/lib/variable.g
+++ b/hpcgap/lib/variable.g
@@ -172,6 +172,8 @@ if IsHPCGAP then
 BIND_GLOBAL( "FLUSHABLE_VALUE_REGION", NewSpecialRegion("FLUSHABLE_VALUE_REGION"));
 fi;
 
+BIND_GLOBAL( "UNCLONEABLE_TNUMS", MakeImmutable([T_INT,T_FFE,T_BOOL]) );
+
 BIND_GLOBAL( "InstallValue", function ( gvar, value )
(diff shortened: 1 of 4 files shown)


frankluebeck:
I have no opinion on the HPC related aspects.

But the cleanup in `lib/list.g`/`src/plist.c` looks nice and useful.

fingolfin:
@frankluebeck OK, good to hear. BTW, you can click on "Files changed", then "Review changes" to approve a PR (meaning you had a look at it and think merging it is sensible), or also "Request changes" (if you want to veto the merge until certain conditions are met, which you'd explain in the "Review summary".

PR #2204, opened 2018-02-23
Random list

ChrisJefferson:
The aim of this PR is to let Random for a collection take a random source. This required adding a random source also to RandomList, and do a little rearranging.

### DIFF
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -264,12 +264,10 @@ InstallGlobalFunction( RandomList, function(args...)
 
   return list[Random(source, 1, Length(list))];
 end );
-InstallMethod( Random, "for a (finite) collection",
-    [ IsCollection and IsFinite ],
-    C -> RandomList( Enumerator( C ) ) );
+
 
 RedispatchOnCondition(Random,true,[IsCollection],[IsFinite],0);
-
+RedispatchOnCondition(Random,true,[IsRandomSource,IsCollection],[IsFinite],0);
 
 #############################################################################
 ##
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -246,8 +246,23 @@ InstallMethod( RepresentativeSmallest,
 ##  an enumerator of <C> and selects a random element of this list using the
 ##  function `RandomList', which is a pseudo random number generator.
 ##
-InstallGlobalFunction( RandomList, function(list)
-  return list[Random(GlobalMersenneTwister, 1, Length(list))];
+
+# RandomList is not an operation to avoid the (often expensive) cost of
+# dispatch for lists
+InstallGlobalFunction( RandomList, function(args...)
+  local len, source, list;
+  len := Length(args);
+  if len = 1 then
+    source := GlobalMersenneTwister;
+    list := args[1];
+  elif len = 2 then
+    source := args[1];
+    list := args[2];
+  else
+    Error(" Usage: RandomList([rs], list))");
+  fi;
+
+  return list[Random(source, 1, Length(list))];
 end );
 InstallMethod( Random, "for a (finite) collection",
     [ IsCollection and IsFinite ],
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -311,3 +311,9 @@ end );
     InstallGlobalFunction("InstallMethodWithRandomSource", func(InstallMethod));
     InstallGlobalFunction("InstallOtherMethodWithRandomSource", func(InstallOtherMethod));
 end)();
+
+# This method must rank below Random(SomeRandomSource, IsList)
+# for any random source SomeRandomSource, to avoid an infinite loop.
+InstallMethodWithRandomSource( Random, "for a random source and a (finite) collection",
+    [ IsRandomSource, IsCollection and IsFinite ], -8,
+    {rs, C} -> RandomList(rs, Enumerator( C ) ) );
--- a/tst/testrandom.g
+++ b/tst/testrandom.g
@@ -1,35 +1,58 @@
-randomTest := function(collection, method, checkin...)
-    local test1, test2, test3, test4, test5, test6, localgen, checkmethod;
-    if Length(checkin) = 0 then
-        checkmethod := \in;
-    else
-        checkmethod := checkin[1];
-    fi;
+# Perform a variety of tests on Random Sources and functions which create
+# random objects.
+#
+# This function is used for a variety is different tests:
+#
+# Test that 'Random(C)' and 'Random(GlobalMersenneTwister, C)' produce 
+# the same answer.
+#
+# Test that 'Random(rs,C)' only uses 'rs', and no other source of random
+#
+# Test Random and RandomList
+#
+# Where there is a global instance of a random source
+# (GlobalMersenneTwister and GlobalRandomSource), they produce the same
+# sequence of answers as a new instance of the same random source.
+
+# filter: The type of random source we are testing.
+# global_rs: A pre-existing object of type 'filter'.
+# randfunc(rs, C): A two argument function which creates random elements of C using
+#              'rs' as the source.
+# global_randfunc(C): A one argument function which is equivalent to 
+#             {x} -> rand(global_rs, x). This lets us check 'Random(C)' and
+#             'Random(GlobalMersenneTwister,C)' produce the same answer when testing
+#             GlobalMersenneTwister. For other random sources, this can just
+#             be set to {x} -> rand(global_rs,x).
+# collection: The object (usually a collection) to find random members of.
+# checkin(e, C): returns if e is in C (usually checkin is '\in').
+
+randomTestInner := function(filter, global_rs, global_randfunc, randfunc, collection, checkin)
+    local test1, test2, test3, test4, test5, test6, local_rs;
 
     # We do a single call first, to deal with calling Random causing extra attributes
     # of 'collection' to be set, changing the dispatch
-    method(collection);
+    randfunc(collection);
 
     # Firstly, we will generate a base list
-    Init(GlobalMersenneTwister, 6);
-    test1 := List([1..1000], x -> method(collection));
-    # test2 should = test1
-    Init(GlobalMersenneTwister, 6);
-    test2 := List([1..1000], x -> method(collection));
+    Init(global_rs, 6);
+    test1 := List([1..1000], x -> global_randfunc(collection));
+    # test2 should equal test1
+    Init(global_rs, 6);
+    test2 := List([1..1000], x -> global_randfunc(collection));
     # test3 should also = test1
-    Init(GlobalMersenneTwister, 6);
-    test3 := List([1..1000], x -> method(GlobalMersenneTwister, collection));
+    Init(global_rs, 6);
+    test3 := List([1..1000], x -> randfunc(global_rs, collection));
     # test4 should be different (as it came from a different seed)
-    Init(GlobalMersenneTwister, 8);
-    test4 := List([1..1000], x -> method(collection));
+    Init(global_rs, 8);
+    test4 := List([1..1000], x -> global_randfunc(collection));
     # test5 should be the same as test4, as it is made from seed 8
     # test6 should be the same as test1. Also, it checks that making test5
     # did not touch the global source at all.
-    Init(GlobalMersenneTwister, 8);
-    localgen :=  RandomSource(IsMersenneTwister, 6);
-    test5 := List([1..1000], x -> method(localgen, collection));
-    test6 := List([1..1000], x -> method(collection));
-    if ForAny(Concatenation(test1, test2, test3, test4, test5, test6), x -> not (checkmethod(x, collection)) ) then
+    Init(global_rs, 8);
+    local_rs :=  RandomSource(filter, 6);
+    test5 := List([1..1000], x -> randfunc(local_rs, collection));
+    test6 := List([1..1000], x -> global_randfunc(collection));
+    if ForAny(Concatenation(test1, test2, test3, test4, test5, test6), x -> not (checkin(x, collection)) ) then
         Print("Random member outside collection: ", collection,"\n");
     fi;
     if test1 <> test2 then
@@ -45,49 +68,81 @@ randomTest := function(collection, method, checkin...)
         Print("Alt gen broken: ", collection, "\n");
     fi;
     if test4 <> test6 then
-        Print("Random with a passed in seed affected the global generator: ", collection, "\n");
+        Print("Random with a passed in seed affected the global source: ", collection, "\n");
     fi;
 end;;
 
+
 # A special test for collections of size 1
-randomTestForSizeOneCollection := function(collection, method)
-    local i, val, localgen, intlist1, intlist2;
-    if Size(collection) <> 1 then
-        Print("randomTestForSizeOneCollection is only for collections of size 1");
-        return;
-    fi;
+# Here we can't check different seeds produce different answers
+# We do check that the random source is not used, for efficency.
+randomTestForSizeOneCollectionInner := function(filter, global_rs, global_randfunc, randfunc, collection, checkin)
+    local i, val, local_rs, intlist1, intlist2;
 
     val := Representative(collection);
 
-    Init(GlobalMersenneTwister, 6);
-    intlist1 := List([1..1000], x -> Random([1..10]));
+    Init(global_rs, 6);
+    intlist1 := List([1..10], x -> global_randfunc([1..1000]));
 
-    for i in [1..1000] do
-        if method(collection) <> val then
+    for i in [1..100] do
+        if global_randfunc(collection) <> val then
             Print("Random returned something outside collection :", collection, ":", val);
         fi;
     od;
 
-    for i in [1..1000] do
-        if method(GlobalMersenneTwister, collection) <> val then
+    for i in [1..100] do
+        if randfunc(global_rs, collection) <> val then
             Print("Random returned something outside collection :", collection, ":", val);
         fi;
     od;
 
-    localgen := RandomSource(IsMersenneTwister, 6);
+    local_rs := RandomSource(filter, 6);
 
-    Init(GlobalMersenneTwister, 6);
-    for i in [1..1000] do
-        if method(localgen, collection) <> val then
+    Init(global_rs, 6);
+    for i in [1..100] do
+        if randfunc(local_rs, collection) <> val then
             Print("Random returned something outside collection :", collection, ":", val);
         fi;
     od;
 
-    # The previous loop should not have affected GlobalMersenneTwister,
+    # The previous loop should not have affected global_rs,
     # so this should be the same as intlist1
-    intlist2 := List([1..1000], x -> Random([1..10]));
+    intlist2 := List([1..10], x -> global_randfunc([1..1000]));
 
     if intlist1 <> intlist2 then
         Print("Random read from local gen affected global gen: ", collection);
     fi;
 end;;
+
+
+randomTest := function(collection, randfunc, checkin...)
+    local sizeone, randchecker;
+    if Length(checkin) = 0 then
+        checkin := \in;
+    else
+        checkin := checkin[1];
+    fi;
+
+    # Make a best attempt to find if the collection is size 1.
+    # There are implementations of random for objects which do not support
+    # Size or IsTrivial, e.g. PadicExtensionNumberFamily
+    if IsList(collection) then
+        sizeone := (Size(collection) = 1);
+    elif IsCollection(collection) then
+        sizeone := IsTrivial(collection);
+    else
+        sizeone := false;
+    fi;
+
+    if sizeone then
+        randchecker := randomTestForSizeOneCollectionInner;
+    else
+        randchecker := randomTestInner;
+    fi;
+
+    randchecker(IsMersenneTwister,
+        GlobalMersenneTwister, x -> randfunc(x), randfunc, collection, checkin);
+    randchecker(IsGAPRandomSource,
+        GlobalRandomSource, x -> randfunc(GlobalRandomSource, x), randfunc,
+        collection, checkin);
+end;
--- a/hpcgap/lib/coll.gd
+++ b/hpcgap/lib/coll.gd
@@ -1661,7 +1661,7 @@ DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 ##
 ##  <#GAPDoc Label="RandomList">
 ##  <ManSection>
-##  <Func Name="RandomList" Arg='list'/>
+##  <Func Name="RandomList" Arg='[rs,] list'/>
 ##
 ##  <Description>
 ##  <Index>random seed</Index>
@@ -1669,7 +1669,9 @@ DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 ##  <Ref Func="RandomList"/> returns a (pseudo-)random element with equal
 ##  distribution.
 ##  <P/>
-##  This function uses the <Ref Var="GlobalMersenneTwister"/> to produce the
+##  The random source <A>rs</A> is used to choose a random number.
+##  If <A>rs</A> is absent,
+##  this function uses the <Ref Var="GlobalMersenneTwister"/> to produce the
 ##  random elements (a source of high quality random numbers).
 ##  <P/>
 ##  </Description>
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -1631,7 +1631,7 @@ DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 ##
 ##  <#GAPDoc Label="RandomList">
 ##  <ManSection>
-##  <Func Name="RandomList" Arg='list'/>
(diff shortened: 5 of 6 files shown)


ChrisJefferson:
This in case anyone wonders, I kept `RandomList` as a function rather than making it a Method because I wanted to keep it fast and simple (and to be able to define it in the same place, early).

fingolfin:
If `RandomList` was an operation (I assume you meant that?), that would indeed be a major speed penalty, as method dispatch overhead for list arguments is the bit that's super-expensive. Might be worth adding a comment to `RandomList` indicating that it must be kept as a function (just like I added comments for `List` etc.:
```
    # handle built-in lists directly, to avoid method dispatch overhead
```

lib/random.gd
```diff
@@ -172,6 +172,15 @@ DeclareOperation( "Random", [IsRandomSource, IsInt, IsInt] );
...
 end)();
 
+
+InstallMethodWithRandomSource( Random,
+    "for a random source and a (finite) collection",
+    [ IsRandomSource, IsCollection and IsFinite ],
+    {rs,C} -> RandomList(rs, Enumerator( C ) ) );
+
+RedispatchOnCondition(Random,true,[IsCollection],[IsFinite],0);
+RedispatchOnCondition(Random,true,[IsRandomSource, IsCollection],[IsFinite],0);
```
fingolfin: Hm, I am not so happy about moving these collection specific functions out of `coll.gi`. Here is a possible alternative, which allows using `InstallMethodWithRandomSource` inside `coll.gi`:
```patch
diff --git a/lib/coll.gi b/lib/coll.gi
index 264a63de0..3f63919b2 100644
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -246,11 +246,6 @@ InstallMethod( RepresentativeSmallest,
 ##  an enumerator of <C> and selects a random element of this list using the
 ##  function `RandomList', which is a pseudo random number generator.
 ##
-if IsHPCGAP then
-    MakeThreadLocal( "GlobalMersenneTwister" );
-else
-    DeclareGlobalVariable( "GlobalMersenneTwister" );
-fi;
 InstallGlobalFunction( RandomList, function(list)
   return list[Random(GlobalMersenneTwister, 1, Length(list))];
 end );
diff --git a/lib/random.gd b/lib/random.gd
index c582a930c..763c31482 100644
--- a/lib/random.gd
+++ b/lib/random.gd
@@ -39,7 +39,7 @@
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-BindGlobal( "RandomSourcesFamily", NewFamily( "RandomSourcesFamily" ) );    
+BIND_GLOBAL( "RandomSourcesFamily", NewFamily( "RandomSourcesFamily" ) );
 DeclareCategory( "IsRandomSource", IsComponentObjectRep );
 
@@ -105,6 +105,14 @@ DeclareOperation( "Random", [IsRandomSource, IsInt, IsInt] );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
+
+if IsHPCGAP then
+    MakeThreadLocal( "GlobalMersenneTwister" );
+else
+    DeclareGlobalVariable( "GlobalMersenneTwister" );
+fi;
+
+
 (function()
     local func;
     func := function(installType)
diff --git a/lib/read1.g b/lib/read1.g
index b8d21f9c8..a360f1c4b 100644
--- a/lib/read1.g
+++ b/lib/read1.g
@@ -36,6 +36,8 @@ ReadLib( "set.gd"      );
 
 ReadLib( "record.gd"   );
 
+ReadLib( "random.gd"   );
+
 ReadLib( "cache.gi"    );
 ReadLib( "coll.gi"     );
 
@@ -70,8 +72,6 @@ ReadLib( "info.gi"     );
 ReadLib( "assert.gi"   );
 ReadLib( "global.gi"   );
 
-ReadLib( "random.gd"   );
-
 ReadLib( "options.gd"  );
 ReadLib( "options.gi"  );
 
```

To me, it anyway nicer to declare `GlobalMersenneTwister` inside `random.gd` instead of `coll.gi` -- this matches `GlobalRandomSource`. The patch above is not quite what I'd do, though: I'd rather move the definition of `GlobalMersenneTwister` where it belongs, i.e. next to the def of `GlobalRandomSource` (where we actually already have commented out definitions for `GlobalMersenneTwister`!). But this means the def for `InstallMethodWithRandomSource` has to be moved downwards, which is simple, but leads to a huge diff).

I can also make a PR with these changes now...

ChrisJefferson:
Assuming the tests pass, I'm not happy with this. I use `-SUM_FLAGS` to make sure the random collection overload is put below everything else.

lib/coll.gi
```diff
@@ -246,15 +246,35 @@ InstallMethod( RepresentativeSmallest,
...
+  return list[Random(source, 1, Length(list))];
 end );
-InstallMethod( Random, "for a (finite) collection",
-    [ IsCollection and IsFinite ],
+
+# Use -SUM_FLAGS, as this should always be the last choice method.
+# In particular, we want Random(SomeRandomSource, IsList) to come higher
+# for lists, to avoid an infinite loop.
+InstallMethodWithRandomSource( Random, "for a (finite) collection",
+    [ IsCollection and IsFinite ], -SUM_FLAGS,
```
frankluebeck: I would avoid extra ranking of methods whenever possible. And the `-SUM_FLAGS` does not seem needed here. E.g., in current master branch:

```
gap> DeclareCategory("IsDice", IsComponentObjectRep and IsDomain);
gap> dice := Objectify(NewType(NewFamily(""), IsDice), rec());
<object>
gap> InstallMethod(Enumerator, [IsDice], d-> [1..6]);
gap> 
gap> InstallMethod(Random, ["IsRandomSource", "IsCollection and IsFinite"],
> function(rs, coll) return Random(Enumerator(coll)); end);
gap> Random(dice);
2
```

ChrisJefferson: Firstly, there is a small bug here, sorry.

This shows the issue I'm trying to fix. Given the following code, then `Random(GlobalRandomSource,[1..10])` enters an infinite loop, because this newly installed method ranks more highly than the `InstallMethod(Random, [IsGAPRandomSource, IsList],...` in random.gi.

```
InstallMethod( Random, "for a source and a (finite) collection",
    [ IsRandomSource, IsCollection and IsFinite ],
    {rs,C} -> RandomList(rs, Enumerator( C ) ) );


MakeReadWriteGlobal("RandomList");
RandomList := function(args...)
  local len, source, list;
  len := Length(args);
  if len = 1 then
    source := GlobalMersenneTwister;
    list := args[1];
  elif len = 2 then
    source := args[1];
    list := args[2];
  else
    Error(" Usage: RandomList([rs], list))");
  fi;

  return list[Random(source, 1, Length(list))];
end;
```

ChrisJefferson:
I am copying this here so it doesn't get lost:

```
InstallMethod( Random, "for a source and a (finite) collection",
    [ IsRandomSource, IsCollection and IsFinite ],
    {rs,C} -> RandomList(rs, Enumerator( C ) ) );


MakeReadWriteGlobal("RandomList");
RandomList := function(args...)
  local len, source, list;
  len := Length(args);
  if len = 1 then
    source := GlobalMersenneTwister;
    list := args[1];
  elif len = 2 then
    source := args[1];
    list := args[2];
  else
    Error(" Usage: RandomList([rs], list))");
  fi;

  return list[Random(source, 1, Length(list))];
end;
```

lib/coll.gi
```diff
@@ -246,15 +246,28 @@ InstallMethod( RepresentativeSmallest,
 ##  an enumerator of <C> and selects a random element of this list using the
 ##  function `RandomList', which is a pseudo random number generator.
 ##
-InstallGlobalFunction( RandomList, function(list)
-  return list[Random(GlobalMersenneTwister, 1, Length(list))];
+
+# RandomList is not a Method to avoid the (often expensive) cost of
```
fingolfin: Method -> method  (only change this if you end up having to update this PR anyway; no need to trigger a full rebuild over such minor thing ;-)/

fingolfin: Also, strictly speaking, I guess this should be "is not an operation", though?

lib/random.gi
```diff
@@ -311,3 +311,10 @@ end );
     InstallGlobalFunction("InstallMethodWithRandomSource", func(InstallMethod));
     InstallGlobalFunction("InstallOtherMethodWithRandomSource", func(InstallOtherMethod));
 end)();
+
+# Use -SUM_FLAGS, as this should always be the last choice method.
+# In particular, we want Random(SomeRandomSource, IsList) to come higher
```
fingolfin: "to come higher" -> "to be ranked higher" ?

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
-randomTest := function(collection, method, checkin...)
-    local test1, test2, test3, test4, test5, test6, localgen, checkmethod;
-    if Length(checkin) = 0 then
-        checkmethod := \in;
-    else
-        checkmethod := checkin[1];
-    fi;
+# Perform a variety of tests on Random generators.
+# filter: The type of random source we are testing
+# globalobj: The name of the global instance of this iterator
```
fingolfin: "this iterator" ? Huh?

Also, why is this named `globalobj`, and in how far does the object passed in have to be "global" ? If I create such an instance on the fly, and also create gmethod on the fly, using this, it would be fine, too?

How about: 
> globalrs: an instance of a random source in <filter> (typically a global instance thereof, like <GlobalMersenneTwister>, to be used by gmethod.

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
-randomTest := function(collection, method, checkin...)
-    local test1, test2, test3, test4, test5, test6, localgen, checkmethod;
-    if Length(checkin) = 0 then
-        checkmethod := \in;
-    else
-        checkmethod := checkin[1];
-    fi;
+# Perform a variety of tests on Random generators.
```
fingolfin: I still don't understand "Random generators" here! First I thought you meant "random source", but that doesn't make sense. My next guess was "methods for the Random method", but that also doesn't seem to make sense, as the test are not restricted to `Random`.

Right now I think you might mean: "Perform a variety of tests with a given function <method> that is expected to produce/generate random output, and which should allow passing in a random source as optional first argument."

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
...
+# filter: The type of random source we are testing
+# globalobj: The name of the global instance of this iterator
+# gmethod: A single argument method which takes elements of the collection
+#          using globalobj (This exists to let us test single-argument Random)
+# method: Two argument random method, which takes a random source and collection
```
fingolfin: Why the name `gmethod`? Also, why "method" -- these two can be any function, can't they? And we frequently pass in operations, not methods...

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
...
+# collection: The collection we want to look for random methods of
+# checkmethod: a way of checking if a value is a member of collection (usually \in)
+
+randomTestInner := function(filter, globalobj, gmethod, method, collection, checkmethod)
+    local test1, test2, test3, test4, test5, test6, localgen;
 
     # We do a single call first, to deal with calling Random causing extra attributes
     # of 'collection' to be set, changing the dispatch
     method(collection);
```
fingolfin: Should this perhaps be `gmethod`? Otherwise, the documentation comment is wrong, as here `method` is called with a single argument.

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
...
+# filter: The type of random source we are testing
+# globalobj: The name of the global instance of this iterator
+# gmethod: A single argument method which takes elements of the collection
+#          using globalobj (This exists to let us test single-argument Random)
+# method: Two argument random method, which takes a random source and collection
+# collection: The collection we want to look for random methods of
+# checkmethod: a way of checking if a value is a member of collection (usually \in)
+
+randomTestInner := function(filter, globalobj, gmethod, method, collection, checkmethod)
+    local test1, test2, test3, test4, test5, test6, localgen;
```
fingolfin: So `localgen` is the sibling to `globalobj`, and also a randomsource... perhaps call them `rs` and `gobal_rs` then?

tst/testrandom.g
```diff
@@ -49,8 +53,22 @@ randomTest := function(collection, method, checkin...)
...
+randomTest := function(collection, method, checkin...)
+    local checkmethod;
+    if Length(checkin) = 0 then
+        checkmethod := \in;
+    else
+        checkmethod := checkin[1];
+    fi;
+
+    randomTestInner(IsMersenneTwister, GlobalMersenneTwister, x -> method(x), method, collection, checkmethod);
+    randomTestInner(IsGAPRandomSource, GlobalRandomSource, x -> method(GlobalRandomSource, x), method, collection, checkmethod);
```
fingolfin: Why do we once `x -> method(x)` and once `x -> method(GlobalRandomSource, x)`? Aha, I assume there is an implicit assumption is that if no RS is specified, then `GlobalMersenneTwister` is used? The manual only says this (emphasis mine):
> *Most* methods for `Random` in the GAP library use the `GlobalMersenneTwister` as random source.

I am happy to enforce this here in the testing code, but I think this should be stated explicitly, at least in a comment right here, e.g. like this:
> If no random source is specified, functions producing randomness are expected to default to using <GlobalMersenneTwister>

tst/testrandom.g
```diff
@@ -49,8 +53,22 @@ randomTest := function(collection, method, checkin...)
...
+end;
+
 # A special test for collections of size 1
-randomTestForSizeOneCollection := function(collection, method)
+# Here we can't check different seeds produce different answers
+# We do check that the random source is not used, for efficency.
```
fingolfin: While that seems sensible, is this documented anywhere?

tst/testrandom.g
```diff
@@ -59,35 +77,47 @@ randomTestForSizeOneCollection := function(collection, method)
...
     # so this should be the same as intlist1
-    intlist2 := List([1..1000], x -> Random([1..10]));
+    intlist2 := List([1..10], x -> gmethod([1..1000]));
 
     if intlist1 <> intlist2 then
         Print("Random read from local gen affected global gen: ", collection);
     fi;
 end;;
+
+randomTestForSizeOneCollection := function(collection, method, checkin...)
```
fingolfin: BTW, I wonder: is there a strong reason to have `randomTestForSizeOneCollection` be a separate function? Couldn't `randomTest` choose between calling `randomTestInner` and `randomTestForSizeOneCollectionInner` based on the collection size? (Even better, it could use `IsTrivial` on the collection, which means "size = 1" but may work even on collections which don't know their precise size)

tst/testrandom.g
ChrisJefferson: This is definately wrong

tst/testrandom.g
ChrisJefferson: No, it seemed to be true in practice, so I decided to test it. I'm not positive I'd want to enforce it, but if it was ever false I'd certainly want to investigate why, as I imagine we'd be doing pointless work.

tst/testrandom.g
ChrisJefferson: Yes, we should do that.

tst/testrandom.g
ChrisJefferson: I think we should possibly strengthen the documentation, to "All methods", I believe the only method which uses anything else is the undocumented RANDOM_LIST.

tst/testrandom.g
ChrisJefferson: method is just a bad choice of name, I tend to use method to refer to "any function", when it has a special meaning in GAP.

tst/testrandom.g
ChrisJefferson: I keep using random generator, when I mean source. Will generally clean up.

ChrisJefferson:
I believe this is now cleaned up and all issues fixed.

frankluebeck:
@ChrisJefferson  Thanks for the example. This is not a specific problem of `Random`, but occurs for all operations when you want to install a method that should be used for finite collections which are not lists.

Still, an inflationary `-SUM_FLAGS` should be avoided, a `-8` or so also would do the job.

But I would remove the `-SUM_FLAGS` here completely and instead push up the method for `[IsGlobalRandomSource, IsList]` by 100, that is what is done for the other cases.

fingolfin:
@frankluebeck could you elaborate *why* you would avoid ˋ-SUM_FLAGSˋ, resp. why the alternatives are "better"? (Are they? Why?) I.e., other than "people in the past did it that way". Right now this just seems like an equivalent alternative, not worth making @ChrisJefferson jump through hoops. Some factial justification would be very helpful, and motivating.

What I don't like about an hardcoded +100 for the Random method for IsList is that I am concerned that it might shadow some other method somebody might install for special kinds of lists. E.g. I might imstall one for ranges, now I need to know I have to give it a +100.

ChrisJefferson:
The problem with pushing up Random(IsGlobalRandomSource,IsList) is that I don't want it to get too high, else it might get above Random(IsRandomSource,IsX) for some X which implements IsList (I haven't checked everything we implement Random for). Also, we would have to do the same with every current and future added random source (and any random sources anyone has already implemented in their own code).

Of course the other option is my other plan, which was to delete Random(IsGlobalRandomSource,IsList) and replace it with Random(IsGlobalRandomSource,IsInt,IsInt), which removes the issue (and also requires changes of course).

PS sorry there is several random changes at the same time. When I started I did not expect this to get so complicated.

tst/testrandom.g
```diff
@@ -1,35 +1,58 @@
...
+# random objects.
+#
+# This function is used for a variety is different tests:
+#
+# Test that 'Random(C)' and 'Random(GlobalMersenneTwister, C)' produce 
+# the same answer.
+#
+# Test that 'Random(rs,C)' only uses 'rs', and no other source of random
+#
+# Test Random, PseudoRandom and RandomList
```
fingolfin: Actually, ˋPseudoRandomˋ does not yet allow a random source, and in fact for some methods (e.g. for groups) it may not be possible to reset the state of a random source and then get the same set of pseudo random elements, because there is additional mixing state in the group

tst/testrandom.g
```diff
@@ -1,35 +1,58 @@
...
+#
+# Where there is a global instance of a random generator
+# (GlobalMersenneTwister and GlobalRandomSource), they produce the same
+# sequence of answers as an new instance of the same random source.
```
fingolfin: An new -> a new

tst/testrandom.g
```diff
@@ -1,35 +1,58 @@
...
+
+# filter: The type of random source we are testing.
+# global_rs: A pre-existing object of type 'filter'.
+# randfunc(rs, C): A two argument function which creates random elements of C using
+#              'rs' as the source.
+# global_randfunc(C): A one argument function which is equivalent to 
+#             {x} -> rand(global_rs, x). This lets us check 'Random(C)' and
+#             'Random(GlobalMersenneTwister,C)' produce the same answer when testing
+#             GlobalMersenneTwister. For other random generators, this can just
```
fingolfin: Generator -> source (and perhaps grep for other occurrences?)

tst/testrandom.g
```diff
@@ -1,35 +1,58 @@
...
     # of 'collection' to be set, changing the dispatch
-    method(collection);
+    randfunc(collection);
 
     # Firstly, we will generate a base list
-    Init(GlobalMersenneTwister, 6);
-    test1 := List([1..1000], x -> method(collection));
+    Init(global_rs, 6);
+    test1 := List([1..1000], x -> global_randfunc(collection));
     # test2 should = test1
```
fingolfin: This confused me at first, maybe spell out "equal"?

tst/testrandom.g
```diff
@@ -49,45 +72,79 @@ randomTest := function(collection, method, checkin...)
...
+        if randfunc(local_rs, collection) <> val then
             Print("Random returned something outside collection :", collection, ":", val);
         fi;
     od;
 
-    # The previous loop should not have affected GlobalMersenneTwister,
+    # The previous loop should not have affected global_rs,
     # so this should be the same as intlist1
-    intlist2 := List([1..1000], x -> Random([1..10]));
+    intlist2 := List([1..10], x -> global_randfunc([1..1000]));
```
fingolfin: Why not just query the state before and after, and compare it? Make an immutable copy to make sure it's not been modified?

tst/testrandom.g
```diff
@@ -49,45 +72,79 @@ randomTest := function(collection, method, checkin...)
...
+          GlobalRandomSource, x -> randfunc(GlobalRandomSource, x), randfunc,
+          collection, checkin);
+    else
+        randomTestInner(IsMersenneTwister,
+          GlobalMersenneTwister, x -> randfunc(x), randfunc, collection, checkin);
+        randomTestInner(IsGAPRandomSource,
+          GlobalRandomSource, x -> randfunc(GlobalRandomSource, x), randfunc,
+          collection, checkin);
+    fi;
+end;
```
fingolfin: Add a trailing newline?

tst/testrandom.g
```diff
@@ -49,45 +72,79 @@ randomTest := function(collection, method, checkin...)
...
+    fi;
+
+    if sizeone then
+        randomTestForSizeOneCollectionInner(IsMersenneTwister, 
+          GlobalMersenneTwister, x -> randfunc(x), randfunc, collection, checkin);
+        randomTestForSizeOneCollectionInner(IsGAPRandomSource, 
+          GlobalRandomSource, x -> randfunc(GlobalRandomSource, x), randfunc,
+          collection, checkin);
+    else
+        randomTestInner(IsMersenneTwister,
```
fingolfin: You could avoid some code duplication by adding a local var which you set to either ˋrandomTestInnerˋ or the orher func.

tst/testrandom.g
ChrisJefferson: I quite like testing the actual sequence. However, it might be nice to also test the states are equal as well!

frankluebeck:
Ad why to avoid SUM_FLAGS, actually, why to avoid any extra ranking in method installation:
In early times of GAP4 there were many methods installed with SUM_FLAGS, then 2*SUM_FLAGS,
3*SUM_FLAGS+4 and so on because several people thought that their method is definitely the "best ever".  Many further methods where installed with other rank shifts. By the time we found and agreed that method ranking is easier to understand and maintain if the filters are ranked such that rank shifts in method installations can be avoided. And +/-SUM_FLAGS should only be used if a method is definitely the most/least significant one. More generally I would try to avoid rank shifts for methods and if this seems difficult, then use a mild shift. 

Now to the specific case: this method is not one of the last resort type (like methods to print certain error messages or to avoid "no method found" errors) and so not a case to use -SUM_FLAGS.

The case of installing a method for `IsFinite and IsCollection` which should not be called for `IsList` (which unfortunately implies `IsCollection`) is one where I don't see how to do it without a rank shift. Checking the filter ranks  a shift of `-3` would be enough. If you want to leave place to maybe put another method in between, then `-8` would be good.

Now to the methods for `IsList` with shift  `+100`: I don't know why `+100` was chosen and have not claimed that this is a better solution. But I have suggested that all cases of random sources should be dealt with in the same way. Maintaining the methods for `Random` does not become easier (and that is the goal of this pull request, isn't it?) when the base methods, say for `IsGAPRandomSource` and `IsMersenneTwister`, have completely different ranks.

If anyone thinks that I'm only writing comments to make poor "@ChrisJefferson jump through hoops" feel free to ignore them.

ChrisJefferson:
@frankluebeck : I don't mind discussing this for a while, this is a sensitive, very old and very well-used bit of code in GAP!

Stepping back for a moment, the short-term aim of this PR is to allow calling `Random(RS,C)` for any random source `RS` and collection `C` and `RandomList(RS,L)` for any random source `RS` and list `L`.

My final long term aim is to make it easy for users to add new implementations of `Random` for new collections which work with any random source, and also new implementations of `Random` for new random sources which work with any collection. I'd like it (as far as possible) to not require us to be careful about adjusting priorities when those things get done, particularly because they might be done in packages.

ChrisJefferson:
i believe all issues in this PR are now fixed.

lib/random.gi
```diff
@@ -311,3 +311,9 @@ end );
     InstallGlobalFunction("InstallMethodWithRandomSource", func(InstallMethod));
     InstallGlobalFunction("InstallOtherMethodWithRandomSource", func(InstallOtherMethod));
 end)();
+
+# This method must rank below Random(SomeRandomSource, IsList)
+# for any random source SomeRandomSource, to avoid an infinite loop.
+InstallMethodWithRandomSource( Random, "for a random source and a (finite) collection",
+    [ IsRandomSource, IsCollection and IsFinite ], -8,
```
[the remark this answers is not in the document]
ChrisJefferson: The `IsList and IsFinite` would fix the generators provided in the library, but still cause an infinite loop in any other random generator (like the one in IO).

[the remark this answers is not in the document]
ChrisJefferson: Doesn't the comment already explain? We need to rank below Random(SomeRandomSource, IsList) for any SomeRandomSource.

PR #2214, opened 2018-02-27
Rename GET_CHAR to GET_NEXT_CHAR and add a return value to it; some more scanner+io refactoring

fingolfin:
This PR contains some tweaks to the io and scanner code:
* change `GET_CHAR` (now: `GET_NEXT_CHAR`) to also return the new char value
* it renames `GET_CHAR` and `PEEK_CHAR` to `GET_NEXT_CHAR` and `PEEK_NEXT_CHAR`; at least for me, this was very helpful to remember what these functions really do, resp. how they relate to each other. An alternative might be to call them `READ_NEXT_CHAR_AND_ADVANCE` and `READ_NEXT_CHAR` or so... but that feels a bit clumsy. Perhaps somebody has another idea?
* use this to avoid accessing `STATE(In)` in the scanner; the number of lines referencing `STATE(In)` decreases from 108 down to 9.
* add `PEEK_CURR_CHAR` helper, which further reduces the places in `scanner.c` using `STATE(In)` to two, both of which *write* to it:
```
$ git grep -n -F "STATE(In)" src/scanner.c
src/scanner.c:961:    { STATE(In)--;
src/scanner.c:1054:  case '\377': STATE(Symbol) = S_EOF;                        *STATE(In) = '\0';  break;
```

Somewhat unrelated from this, I also cleaned up some code in `GetLine2`, and slightly refactored `CharHexDigit`.

@markuspf does this interfere with unpublished IO work of yours?

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -215,6 +215,10 @@ Char PEEK_NEXT_CHAR(void)
     return *IO()->RealIn;
 }
 
+Char PEEK_CURR_CHAR(void)
+{
+    return *STATE(In);
+}
 
 const Char * GetInputFilename(void)
 {
--- a/src/io.c
+++ b/src/io.c
@@ -188,7 +188,7 @@ static inline Int IS_CHAR_PUSHBACK_EMPTY(void)
     return STATE(In) != &IO()->Pushback;
 }
 
-void GET_NEXT_CHAR(void)
+Char GET_NEXT_CHAR(void)
 {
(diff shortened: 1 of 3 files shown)


markuspf:
I don't think it interferes with what I am doing (I am refactoring handling of `STATE(Input)`, `STATE(Output)`. A pull request is imminent. I am going to merge this PR.

PR #2215, opened 2018-02-27
Move line continuation handling to io.c

### LINKED ISSUE #2201: kernel idea: deal with line continuations in `io.c` (in `GetLine` ?), not in the scanner.
Right now, all kinds of code in `scanner.c` deals with line continuations, i.e., a backslash followed by a newline, as in this example:
```
gap> x:="foo\
> bar";
"foobar"
```
This is quite repetitive. There is code for this in at least `GetIdent`, `GetCleanedChar`, `GetStr`, `GetChar`, and in several places in `GetSymbol`:
```C
...
  case '!':   STATE(Symbol) = S_ILLEGAL;                     GET_CHAR();
    if ( *STATE(In) == '\\' ) { GET_CHAR();
      if ( *STATE(In) == '\n' ) { GET_CHAR(); } }
...
  case ':':   STATE(Symbol) = S_COLON;                       GET_CHAR();
    if ( *STATE(In) == '\\' ) {
      GET_CHAR();
      if ( *STATE(In) == '\n' )
        { GET_CHAR(); }
    }
    if ( *STATE(In) == '=' ) { STATE(Symbol) = S_ASSIGN;  GET_CHAR(); break; }
    break;
...
  case '=':   STATE(Symbol) = S_EQ;                          GET_CHAR();  break;
  case '<':   STATE(Symbol) = S_LT;                          GET_CHAR();
    if ( *STATE(In) == '\\' ) { GET_CHAR();
      if ( *STATE(In) == '\n' ) { GET_CHAR(); } }
    if ( *STATE(In) == '=' ) { STATE(Symbol) = S_LE;      GET_CHAR();  break; }
    if ( *STATE(In) == '>' ) { STATE(Symbol) = S_NE;      GET_CHAR();  break; }
    break;
  case '>':   STATE(Symbol) = S_GT;                          GET_CHAR();
    if ( *STATE(In) == '\\' ) { GET_CHAR();
      if ( *STATE(In) == '\n' ) { GET_CHAR(); } }
    if ( *STATE(In) == '=' ) { STATE(Symbol) = S_GE;      GET_CHAR();  break; }
    break;

  case '+':   STATE(Symbol) = S_PLUS;                        GET_CHAR();  break;
  case '-':   STATE(Symbol) = S_MINUS;                       GET_CHAR();
    if ( *STATE(In) == '\\' ) { GET_CHAR();
      if ( *STATE(In) == '\n' ) { GET_CHAR(); } }
    if ( *STATE(In) == '>' ) { STATE(Symbol)=S_MAPTO;     GET_CHAR();  break; }
    break;
...
```

This also means that sometimes you are allowed to use a line continuation; sometimes you can use one, but don't have to; and sometimes don't need to use one at all.

E.g. inside a (non-triple quote) string, you can't have newlines, so if you want to break it, you need a line continuation:
```
gap> x:="foo\
> bar";
"foobar"
gap> x:="foo
Syntax error: String must not include <newline>
x:="foo
      ^
> ;
```
You can use one to break `:=` or `<=`  etc. "in the middle:
```
gap> x :\
> =1;
1
gap> x :
Syntax error: ; expected
x :
  ^
```
You can use line continuations to break identifiers and keywords:
```
gap> x m\
> od 5;
1
```

But you cannot use them to e.g. split double or triple dots:
```
gap> [1.\
> .4];
Error, Record Element: <rec> must be a record (not a macfloat)
not in any function at *stdin*:36
you can replace <rec> via 'return <rec>;'
brk>
gap> {x...}->x;
function( x... ) ... end
gap> {x..\
Syntax error: Three dots required for variadic argument list
{x..\
   ^
> ;
Syntax error: -> expected
;
^
gap>
```

Whenever one adds some new syntax, one has to be super careful about thinking how to deal with line continuations. All this complexity would go away if the IO code handled with them, getting rid of line continuations before the scanner sees them.

There is one drawback, though: one has to be very careful about not breaking `SyntaxError` and `SyntaxWarning` output, and about using correct line number references in e.g. coded functions. But since the current line number comes from the IO module anyway (via `GetInputLineNumber()`) I am hopeful this can be achieved without too much hassle (one just needs to be aware of it, and add suitable test cases to verify this does not break).

fingolfin:
Resolve #2201.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -220,6 +220,14 @@ Char PEEK_CURR_CHAR(void)
     return *STATE(In);
 }
 
+void IGNORE_REST_OF_LINE(void)
+{
+    Char c = *STATE(In);
+    while (c != '\n' && c != '\r' && c != '\377')
+        c = GET_NEXT_CHAR();
+}
+
+
 const Char * GetInputFilename(void)
 {
     GAP_ASSERT(IO()->Input);
--- a/src/io.c
+++ b/src/io.c
@@ -195,14 +195,57 @@ Char GET_NEXT_CHAR(void)
     }
     else
         STATE(In)++;
+
+    // handle line continuation, i.e., backslash followed by new line; and
+    // also the case when we run out of buffered data
+    while (*STATE(In) == '\\' || *STATE(In) == 0) {
+
+        // if we run out of data, get more, and try again
+        if (*STATE(In) == 0) {
+            GetLine();
+            continue;
+        }
+
+        // we have seen a backslash; so check now if it starts a
+        // line continuation, i.e., whether it is followed by a line terminator
+        if (STATE(In)[1] == '\n') {
+            // LF is the line terminator used in Unix and its relatives
+            STATE(In) += 2;
+        }
+        else if (STATE(In)[1] == '\r' && STATE(In)[2] == '\n') {
+            // CR+LF is the line terminator used by Windows
+            STATE(In) += 3;
+        }
+        else {
+            // if we see a backlash without a line terminator after it, stop
+            break;
+        }
+    }
+
+    return *STATE(In);
+}
+
+// GET_NEXT_CHAR_NO_LC is like GET_NEXT_CHAR, but does not handle
+// line continuations. This is used when skipping to the end of the
+// current line, when handling comment lines.
+static Char GET_NEXT_CHAR_NO_LC(void)
+{
+    if (STATE(In) == &IO()->Pushback) {
+        STATE(In) = IO()->RealIn;
+    }
+    else
+        STATE(In)++;
+
     if (!*STATE(In))
         GetLine();
+
     return *STATE(In);
 }
 
 Char PEEK_NEXT_CHAR(void)
 {
     assert(IS_CHAR_PUSHBACK_EMPTY());
+
     // store the current character
     IO()->Pushback = *STATE(In);
 
@@ -220,11 +263,11 @@ Char PEEK_CURR_CHAR(void)
     return *STATE(In);
 }
 
-void IGNORE_REST_OF_LINE(void)
+void SKIP_TO_END_OF_LINE(void)
 {
     Char c = *STATE(In);
     while (c != '\n' && c != '\r' && c != '\377')
-        c = GET_NEXT_CHAR();
+        c = GET_NEXT_CHAR_NO_LC();
 }
 
 
--- a/src/io.c
+++ b/src/io.c
@@ -1067,26 +1067,30 @@ static Int GetLine2 (
             bptr++;
 
         /* copy piece of input->sline into buffer and adjust counters */
-        UInt count = input->spos;
-        Char *ptr = (Char *)CHARS_STRING(input->sline) + count;
-        UInt len = GET_LEN_STRING(input->sline);
-        Char *bend = buffer + length - 2;
-        while (bptr < bend && count < len && *ptr != '\n' && *ptr != '\r') {
-            *bptr++ = *ptr++;
-            count++;
-        }
-        /* we also copy an end of line if there is one */
-        if (*ptr == '\n' || *ptr == '\r') {
-            *bptr++ = *ptr++;
-            count++;
+        Char *ptr = CSTR_STRING(input->sline) + input->spos;
+        const Char * const end = CSTR_STRING(input->sline) + GET_LEN_STRING(input->sline);
+        const Char * const bend = buffer + length - 2;
+        while (bptr < bend && ptr < end) {
+            Char c = *ptr++;
+
+            // ignore CR, so that a Window CR+LF line terminator looks
+            // to us the same as a Unix LF line terminator
+            if (c == '\r')
+                continue;
+
+            *bptr++ = c;
+
+            // check for line end
+            if (c == '\n')
+                break;
         }
         *bptr = '\0';
-        input->spos = count;
+        input->spos = ptr - (Char *)CHARS_STRING(input->sline);
 
         /* if input->stream is a string stream, we have to adjust the
            position counter in the stream object as well */
         if (input->isstringstream) {
-            ADDR_OBJ(input->stream)[1] = INTOBJ_INT(count);
+            ADDR_OBJ(input->stream)[1] = INTOBJ_INT(input->spos);
         }
     }
     else {
--- a/src/io.h
+++ b/src/io.h
@@ -25,8 +25,9 @@ extern Char GET_NEXT_CHAR(void);
 extern Char PEEK_NEXT_CHAR(void);
 extern Char PEEK_CURR_CHAR(void);
 
-// skip the rest of the current line
-extern void IGNORE_REST_OF_LINE(void);
+// skip the rest of the current line, ignoring line continuations
+// (used to handle comments)
+extern void SKIP_TO_END_OF_LINE(void);
 
 /****************************************************************************
 **
--- a/src/io.h
+++ b/src/io.h
@@ -25,6 +25,8 @@ extern Char GET_NEXT_CHAR(void);
 extern Char PEEK_NEXT_CHAR(void);
 extern Char PEEK_CURR_CHAR(void);
 
+// skip the rest of the current line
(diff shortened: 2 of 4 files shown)


fingolfin:
A test was failing, which did this:
```
gap> EvalString("1234\\\r\n567");
1234567
```
The problem came from a difference in behavior between (kernel) string streams and file streams in `io.c`. For the former, both CR and LF were treated as line terminators; for the latter, it depends on the OS, so Unix, only LF is a line terminator. When reading a CRLF from a file, we thus see `\\\r\n\0` in the line buffer, and can correctly treat this as a single ending in a line continuation. But when reading from the string, the buffer would end with `\\\r\0`; to correctly deal with this, one needs to call `GetLine` again to get the missing `\n`. The old line continuation code *sometimes* did that, sometimes not. 

To fix this, I changed the string stream code to ignore all `\r` and only accept `\n` as line terminator. This is a bit of a hack, and if a file uses isolated CRs, it could lead to some changed error behavior, but otherwise I think it's fine.

This still is not quite perfect, as regular file streams could in border cases also produce a line buffer ending in `\\\r\0`, namely when we see a very long (~32k) line which ends in a CRLF, where the CR (plus a trailing zero) just fits into the buffer, but the LF does not. That's not great, but OTOH it only causes a problem if this super long line ends in a line continuation, too. And the bad result then is an unexpected error, i.e. no silent data loss, so I am OK with it. We need to look closer at lines overflowing the buffer at some point anyway.

Another fix is that I change the prompt when we do an explicit line continuation (which is much easier now that we handle them in a single place). As a result, this example (in master)
```
gap> 123\
gap> 456;
123456
gap> "123\
gap> "456";
Syntax error: ; expected
"456";
   ^
```
becomes this:
```
gap> 123\
> 456;
123456
gap> "123\
> 456";
"123456"
```

fingolfin:
BTW, I am aware that this PR is missing tests, I'll work on that.

fingolfin:
I have now changed this, as discussed on issue #2201, to *not* handle line continuations inside of comments, which thus made it possible to drop the changes to `lib/stbcckt.gi` again. This means that the current behavior of GAP is not changed. I figure this is the best way to go: If we really want to change how we deal with line continuations, fine, we can do it; but let's not do it as part of this PR, but rather "intentionally", in another PR.

In any case, this PR here still lacks explicit test cases; I'll work on them when I am back from vacation. Also, it contains a ton of refactoring to the scanner code which was enabled by the (re)moved line continuation code. I could put that into a separate PR, to simplify reviewing, too.

fingolfin:
I have now added tests. Indeed, now the first commit adds some tests for line continuations. Then, as mentioned in #2201, a new helper function `IGNORE_REST_OF_LINE` is added to `io.c`, which the next commit uses to move line continuation handling to `io.c`; in the same commit, the new test file is adjusted: first off, the behaviour of triple quotes with line continuations changes; but as discussed in #2201, this is deemed to be an improvement (or even a fix?). 
Also, a few more tests are added, which showcase uses of line continuation which were not supported before, such as breaking keywords and multi character operators such as `:=` along a line continuation (this is of course *not* something I'd recommend people should do, but it opens the possibility of simplifying line wrapping rules for width constrained GAP output, e.g. in tests...)

The rest then is a ton of refactoring and simplification in the scanner, which is enabled by the code simplification thanks to the moving of the line continuation code.

Another improvement is that now the line continuation code always changesthe GAP prompt to the line continuation prompt `>`. I think this will fix some bugs and perhaps allow further simplifications...

Thing is, all these extra commits likely make reviewing this harder than it is. I'll look into moving them into one or multiple additional PRs, and reducing this PR here to the bare minimum, but that'll have to wait for another day.

In the meantime, feel free to already review parts or all of this PR.

src/io.c
```diff
@@ -195,14 +195,47 @@ Char GET_NEXT_CHAR(void)
...
         STATE(In)++;
+
+    // handle line continuation, i.e., backslash followed by new line or CRLF;
+    // and also the end of a line in general
+    while (*STATE(In) == '\\' || *STATE(In) == 0) {
+        if (!*STATE(In))
+            GetLine();
+        else if (STATE(In)[1] == '\n') {
+            STATE(In) += 2;
+            /* print only a partial prompt from now on                             */
```
fingolfin: Note to self: change this to a `//` comment

src/io.c
```diff
@@ -195,14 +195,47 @@ Char GET_NEXT_CHAR(void)
...
     else
         STATE(In)++;
+
+    // handle line continuation, i.e., backslash followed by new line or CRLF;
+    // and also the end of a line in general
+    while (*STATE(In) == '\\' || *STATE(In) == 0) {
+        if (!*STATE(In))
+            GetLine();
+        else if (STATE(In)[1] == '\n') {
+            STATE(In) += 2;
```
fingolfin: Maybe add a comment here which indicates that a line continuation was detected?

Also: the term "line continuation" should be documented somewhere (resp. be replaced by a different, documented name, if somebody has a good suggestion?)

src/io.c
```diff
@@ -195,14 +195,47 @@ Char GET_NEXT_CHAR(void)
...
+            /* print only a partial prompt from now on                             */
+            if ( !SyQuiet )
+              STATE(Prompt) = "> ";
+            else
+              STATE(Prompt) = "";
+        }
+        else if (STATE(In)[1] == '\r')
+            STATE(In) += (STATE(In)[2] == '\n') ? 3 : 2;
```
fingolfin: Mention here that this is about handling Windows/DOS style CR+LF?!?

src/io.c
```diff
@@ -195,14 +195,47 @@ Char GET_NEXT_CHAR(void)
...
+        else
+            break;
+    }
+
+    return *STATE(In);
+}
+
+static Char GET_NEXT_CHAR2(void)
```
fingolfin: Add a comment which explains that this internal function is like `GET_NEXT_CHAR`, except that it does not care about line coninuations. Also, come up with a better name, perhaps???

Alternatively, merge this code into `IGNORE_REST_OF_LINE`?

src/io.c
```diff
@@ -220,6 +253,14 @@ Char PEEK_CURR_CHAR(void)
     return *STATE(In);
 }
 
+void IGNORE_REST_OF_LINE(void)
```
fingolfin: Any suggestioins for a better name? `SKIP_TO_END_OF_LINE`? `SKIP_LINE`? ...?

fingolfin:
I have now reduced this PR to its bare minimum. Further refactoring and fixes will come in separate PRs.

src/io.c
```diff
@@ -195,14 +195,44 @@ Char GET_NEXT_CHAR(void)
     }
     else
         STATE(In)++;
+
+    // handle line continuation, i.e., backslash followed by new line or CRLF;
+    // and also the end of a line in general
+    while (*STATE(In) == '\\' || *STATE(In) == 0) {
```
ChrisJefferson: Out of interest, why do we handle null characters like backslashes?

src/io.c
ChrisJefferson: SKIP_TO_END_OF_LINE sounds good to me

src/io.c
fingolfin: We don't, we just have to keep reading for both: for null chars, we have to read more data via `GetLine()`, then start over; and for backslash, we have to see if next comes a lineend, and if so, skip the backslash and the lineend. And in theory, either of these can happen after the other, hence the loop.

fingolfin: Does that make sense? Would a better comment help? Suggestions welcome!

ChrisJefferson: Reading this again, I didn't realise that the second part of 201 and 202 are the same case. Maybe change 202 to `if(*STATE(In) == 0)`, which would make it more obvious that is matches the second condition in 201?

fingolfin: will do, and will try to write some comments, too

fingolfin: Done

src/io.c
fingolfin: Renamed to `SKIP_TO_END_OF_LINE`

PR #2217, opened 2018-02-27
Modernise random number generation

ChrisJefferson:
This cleanups up some random number generation.

The main change here is that random number generators only implement `Random(RandomSource, IsInt, isInt)`. This helps with overloading resolution.

I also add tests for the different generators contained in GAP.

### DIFF
--- a/lib/random.gd
+++ b/lib/random.gd
@@ -107,7 +107,14 @@ DeclareOperation( "Random", [IsRandomSource, IsInt, IsInt] );
 ##  data which are needed. Optionally, it should allow one to specify a 
 ##  <A>seed</A> for the initial state, as explained for <Ref Oper="Reset"/>.
 ##  <P/>
-##  Most methods for <Ref Oper="Random" Label="for a list or collection"/> 
+##  Random sources should also implement <Ref Oper="Random"
+##  Label="for random source and two integers"/> for a random source and two integers.
+##  All other implementations
+##  of <Ref Oper="Random" BookName="ref"/> call other overloads of
+##  <Ref Oper="Random" BookName="ref"/>, or
+##  <Ref Oper="Random" Label="for random source and two integers"/> for a random source
+##  and two integers.
+##  <P/>
 ##  in the &GAP; library use the 
 ##  <Ref Var="GlobalMersenneTwister"/> as random source. It can be reset 
 ##  into a known state as in the following example.
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -54,9 +54,6 @@ InstallMethod(Random, [IsRandomSource, IsInt, IsInt], function(rs, a, b)
   fi;
 end);
 
-InstallMethod(Random, [IsRandomSource, IsList], function(rs, list)
-  return list[Random(rs, 1, Length(list))];
-end);
 
 # A print method.
 InstallMethod(PrintObj, [IsRandomSource], function(rs)
@@ -231,13 +228,6 @@ else
 InstallValue(GlobalMersenneTwister, RandomSource(IsMersenneTwister, "1"));
 fi;
 
-# default random method for lists and pairs of integers using the Mersenne
-# twister
-InstallMethod( Random, "for an internal list",
-    [ IsList and IsInternalRep ], 100, function(l) 
-  return l[Random(GlobalMersenneTwister, 1, Length(l))]; 
-end );
-
 InstallMethod( Random,
     "for two integers",
     IsIdenticalObj,
@@ -318,3 +308,8 @@ InstallMethodWithRandomSource( Random, "for a random source and a (finite) colle
     [ IsRandomSource, IsCollection and IsFinite ],
     {} -> -RankFilter(IsCollection and IsFinite),
     {rs, C} -> RandomList(rs, Enumerator( C ) ) );
+
+InstallMethodWithRandomSource( Random,
+    "for a random source and a list",
+    [ IsRandomSource, IsList ],
+    {rs, C} -> C[Random(rs,1,Length(C))]);
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -29,9 +29,11 @@ InstallMethod(Reset, [IsRandomSource], function(rs)
   return Reset(rs, 1);
 end);
 
-# Generic fallback methods, such that it is sufficient to install Random for 
-# lists or for pairs of integers.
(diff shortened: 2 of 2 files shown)


lib/random.gd
```diff
@@ -303,7 +310,7 @@ DeclareOperation( "RandomSource", [IsOperation, IsObject] );
             CallFuncList(installType, args);
 
             # Install random, wrapping random source argument
-            argscopy := List(args);
+            argscopy := ShallowCopy(args);
```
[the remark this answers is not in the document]
fingolfin: I made PR #2218 to demonstrate what I mean. But please don't let that PR stop you: I do not expect you to rebase your work on that, I'll be happy to adapt that PR once this PR is merged.

lib/random.gi
```diff
@@ -29,9 +29,11 @@ InstallMethod(Reset, [IsRandomSource], function(rs)
...
 end);
 
-# Generic fallback methods, such that it is sufficient to install Random for 
-# lists or for pairs of integers.
-InstallMethod(Random, [IsRandomSource, IsInt, IsInt], function(rs, a, b)
+DeclareGlobalFunction("RANDOM_LARGE_RANGE");
+
+# Generic method to build an arbitary random number generator from a generator
+# which can generate numbers at least in the range [0..2^10].
+InstallGlobalFunction("RANDOM_LARGE_RANGE", function(gen, a, b)
```
fingolfin: Why 2^10 and not e.g. 2^28? How about making the exponent "10" an argument to the function, for optimal performance?
(I am also OK with leaving this as is, though)

lib/random.gi
```diff
@@ -98,11 +96,11 @@ InstallMethod(Reset, [IsGlobalRandomSource, IsObject], function(rs, seed)
...
 
-InstallMethod(Random, [IsGlobalRandomSource, IsList], function(rs, l)
-  if Length(l) < 2^28 then
-    return RANDOM_LIST(l);
+InstallMethod(Random, [IsGlobalRandomSource, IsInt, IsInt], function(rs, l, u)
+  if u - l + 1 < 2^28 then
+    return RANDOM_LIST([l..u]);
   else
-    return l[Random(rs, 1, Length(l))];
+    return RANDOM_LARGE_RANGE(x -> RANDOM_LIST([0..x]), l, u);
```
fingolfin: If it wasn't for the fact that the `io` package already implements a custom random source, I'd suggest to switch the primitive Random to a function `Random(rs, max)` which returns values from the interval `[0..max]` (or `[1..max]` if it was more convenient). Anyway, changing this and syncing it with `io` would be a lot of hassle, so let's leave it at this.

I do wonder what happens if l > u, though... we'll take a random element from an empty list then?

fingolfin: (again: these are old issues, I don't expect them to be addressed by this PR)

tst/testrandom.g
```diff
@@ -1,34 +1,38 @@
-randomTest := function(collection, method, checkin...)
-    local test1, test2, test3, test4, test5, test6, localgen, checkmethod;
-    if Length(checkin) = 0 then
-        checkmethod := \in;
-    else
-        checkmethod := checkin[1];
-    fi;
+# Perform a variety of tests on Random generators.
```
fingolfin: generators -> sources? methods? Not sure what is meant; perhaps a slightly longer description is needed?

lib/random.gi
ChrisJefferson: This just came from pulling out an old function fairly unchanged. When I did this I had hoped to not change the random numbers produced (that turned out to be impossible for other reasons).

lib/random.gd
```diff
@@ -107,7 +107,14 @@ DeclareOperation( "Random", [IsRandomSource, IsInt, IsInt] );
 ##  data which are needed. Optionally, it should allow one to specify a 
 ##  <A>seed</A> for the initial state, as explained for <Ref Oper="Reset"/>.
 ##  <P/>
-##  Most methods for <Ref Oper="Random" Label="for a list or collection"/> 
+##  Random sources should also implement <Ref Oper="Random"
+##  Label="for random source and two integers"/> for a random source and two integers.
```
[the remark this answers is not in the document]
frankluebeck: I support the "nitpick".

lib/random.gi
```diff
@@ -145,17 +143,22 @@ InstallMethod(Reset, [IsGAPRandomSource, IsObject], function(rs, seed)
   return old;
 end);
 
-InstallMethod(Random, [IsGAPRandomSource, IsList], function(rs, list)
-  local rx, rn;
-  if Length(list) < 2^28 then
+InstallMethod(Random, [IsGAPRandomSource, IsInt, IsInt], function(rs, l, u)
+  local gen; # rx, rn;
```
frankluebeck: Please, add a check of the arguments:
```
  if u < l then
    return fail;
  fi;
```

frankluebeck: Making this comment I notice that lines 213 and 215 also need a fix (it should be `d < 1` and `d = 1`, respectively.

lib/random.gi
frankluebeck: The interface with the two integer arguments is used in many places. I would not change that at this stage.

frankluebeck:
I'm still not sure what problem is solved by this pull request. But otherwise it looks fine to me to reduce everything to the two integer versions of methods for specific types of random sources.

lib/random.gd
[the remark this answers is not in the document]
fingolfin: Since @ChrisJefferson merged #2218 in the meantime, I guess this PR needs to be rebased now after all.

lib/random.gi
fingolfin: Agreed.

lib/random.gi
fingolfin: It would indeed be nice if such checks could be added, and perhaps also tests that verify them; if not in this PR, then in some later on.

ChrisJefferson:
Just to say, I am going to wait until #2204 is merged, and then rebase this patch on top of that.

wilfwilson:
@ChrisJefferson What's the status of this? Note that #2204 was merged 🙂

ChrisJefferson:
I'm going to chuck this, because it's not clear to me the churn it would cause in random number generation is worth the minor internal cleanups it brings.

Issue #1705, opened 2017-09-10
Chapter on relations has no single example at all


alex-konovalov:
My attention was turned to this by [this question](https://math.stackexchange.com/questions/2359463/poset-test-and-hasse-diagram-in-gap). 

Some examples of this functionality may be found in the test file https://github.com/gap-system/gap/blob/master/tst/testinstall/relation.tst and in my 10+ old teaching materials at https://www.gap-system.org/ukrgap/Examples/relation.htm.

grouptheoryenthusiast:
Hi,

I had a question about style/format for adding examples to the documentation chapter.

Is it allowed for an example to demonstrate several different functions.

For instance could there be an example which defines a relation, show it is reflexive, symmetric, transitive, then that it is an equivalence relation and finally finds the equivalence classes,  or is it that the point of the examples is to  demonstrate only the function in whose description it has been embedded. If this is allowed, where would be the correct place for such an example?

Also, as the entries in the chapter are numbered, can one use a function with a later number in an earlier example?

Thank you

James

AnnaKDS:
@ssiccha it seems like somebody forgot to close this issue

PR #2219, opened 2018-02-28
Adding examples to the relations chapter

grouptheoryenthusiast:
Responding to issue #1705 I have added some examples into the relations chapter.

In particular, have added examples into the gapdoc sections within `relation.gd`.

### DIFF
--- a/lib/relation.gd
+++ b/lib/relation.gd
@@ -95,6 +95,10 @@ DeclareSynonym("IsBinaryRelation",IsEndoGeneralMapping);
 ##  to <A>list</A><C>[1]</C>, <M>2</M> to <A>list</A><C>[2]</C> and so on.
 ##  The first version checks whether the list supplied is valid.
 ##  The the <C>NC</C> version skips this check.
+##  <Example><![CDATA[
+##  gap> R:=BinaryRelationOnPoints([[1,2],[2],[3]]);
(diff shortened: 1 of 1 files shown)


lib/relation.gd
```diff
@@ -156,6 +168,10 @@ DeclareGlobalFunction("IdentityBinaryRelation");
 ##  consisting of the tuples collection <A>elms</A>.
 ##  This construction is similar to <Ref Func="GeneralMappingByElements"/>
 ##  where the source and range are the same set.
+##  <Example><![CDATA[
+##  gap> r:=BinaryRelationByElements(Domain([1..3]),[Tuple([1,2]),Tuple([1,3])]);
+##  <general mapping: <object> -> <object> >
+##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
```
fingolfin: How about also adding an example for `EmptyBinaryRelation`, e.g.
```
gap> EmptyBinaryRelation(3) = BinaryRelationOnPoints([ [], [], [] ]);
true
```

grouptheoryenthusiast: Hi @fingolfin , thank you for the suggestion. I have now added this and also added Is Symmetric/Reflexive/Transitive BinaryRelation based on your comment above. 
I've pushed these changes to this branch (as well as fixing a missing `;`).

PR #2225, opened 2018-03-01
lib: improve LocationFunc for kernel funcs

fingolfin:
Before:

    gap> LocationFunc(APPEND_LIST_INTR);
    ""

After:

    gap> LocationFunc(APPEND_LIST_INTR);
    "src/listfunc.c:APPEND_LIST_INTR"

### DIFF
--- a/lib/function.g
+++ b/lib/function.g
@@ -344,13 +344,23 @@ BIND_GLOBAL( "LocationFunc", function(x)
     if not(IS_FUNCTION(x)) then
         return "";
     fi;
-    nam := FILENAME_FUNC(x);
-    line := STARTLINE_FUNC(x);
     ret := "";
-    if nam <> fail and line <> fail then
+    nam := FILENAME_FUNC(x);
+    if nam = fail then
+        return ret;
+    fi;
+    line := STARTLINE_FUNC(x);
+    if line <> fail then
         APPEND_LIST(ret, nam);
         APPEND_LIST(ret, ":");
         APPEND_LIST(ret, STRING_INT(line));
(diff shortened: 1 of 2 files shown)


frankluebeck:
I wasn't aware of the function `LocationFunc`. You may want to look at the code of `PageSource`, try
```
gap> PageSource(PageSource);
```
This could already handle many kernel functions, like
```
gap> PageSource(APPEND_LIST_INTR);
```
(it is also showing the location in the terminal). `PageSource` can also show the locations of declarations of operations. Maybe you want to add that to `LocationFunc` as well.
```
gap> PageSource(Size);
gap> PageSource(Size, 2);
```

fingolfin:
I was aware of `PageSource`, but it's not useful for me. I use my own function which opens the relevant file in an editor at the right position.

But this all seems irrelevant to this PR: `LocationFunc` is mainly useful to me to produce error messages, or for other kinds of debug output (e.g. to give a list of locations of methods and where they are defined; indeed, that's what it is used for in `ApplicableMethodTypes`.

fingolfin:
I am also aware of the fact that `PageSource` knows how to locate operations (I was reviewer of the PR adding it, after all ;-), and even thought briefly about adding it to `LocationFunc `, but since it can return multiple results, it didn't strike me as that useful for `LocationFunc`, at least without bigger changes to its API. Since this goes way beyond the little itch I want to fix with this PR, I have no interest in pursuing it at this time for `LocationFunc`.

fingolfin:
Just added a test, let's see if it passes CI.

PR #2237, opened 2018-03-04
Remove DEBUG_LOADING, InfoRead1, and InfoRead2

markuspf:
They are a leftover of a legacy debugging mechanism. `InfoRead1` and `InfoRead2` are moved into `obsoletes.gd`.

Unfortunately the following packages still refer to `InfoRead1` or `InfoRead2`:

- [ ] atlasrep (uses `InfoRead1` for info output, should be replaced by `Info()`)
- [ ] gbnp (only saves and restores state in tests)
- [x] guava (only saves and restores state) done in https://github.com/gap-packages/guava/pull/33 
- [x] hecke (spurious misuse of `InfoRead1`): TODO https://github.com/gap-packages/hecke/issues/5
- [x] xgap: TODO https://github.com/gap-packages/xgap/issues/7

These packages will break with this PR if GAP is started without obsoletes.

### DIFF
--- a/lib/files.gi
+++ b/lib/files.gi
@@ -202,23 +202,23 @@ InstallMethod( Read,
     "string",
     [ IsString ],
 function ( name )
-    local   readIndent,  found;
+    local readIndent;
 
-    name := UserHomeExpand(name);
-
-    readIndent := SHALLOW_COPY_OBJ( READ_INDENT );
-    APPEND_LIST_INTR( READ_INDENT, "  " );
     if GAPInfo.CommandLineOptions.D then
-    	Print( "#I", READ_INDENT, "Read( \"", name, "\" )\n" );
+        readIndent := SHALLOW_COPY_OBJ( READ_INDENT );
+        APPEND_LIST_INTR( READ_INDENT, "  " );
+        Print( "#I", READ_INDENT, "Read( \"", name, "\" )\n" );
     fi;
-    found := (IsReadableFile(name)=true) and READ(name);
-    READ_INDENT := readIndent;
-    if GAPInfo.CommandLineOptions.D and
-       found and READ_INDENT = ""  then
-        Print( "#I  Read( \"", name, "\" ) done\n" );
-    fi;
-    if not found  then
+
+    if not READ(UserHomeExpand(name)) then
         Error( "file \"", name, "\" must exist and be readable" );
+    fi;
+
+    if GAPInfo.CommandLineOptions.D then
+        READ_INDENT := readIndent;
+        if READ_INDENT = "" then
+            Print( "#I  Read( \"", name, "\" ) done\n" );
+        fi;
     fi;
 end );
 
--- a/lib/files.gi
+++ b/lib/files.gi
@@ -208,11 +208,14 @@ function ( name )
 
     readIndent := SHALLOW_COPY_OBJ( READ_INDENT );
     APPEND_LIST_INTR( READ_INDENT, "  " );
-    InfoRead1( "#I", READ_INDENT, "Read( \"", name, "\" )\n" );
+    if GAPInfo.CommandLineOptions.D then
+    	Print( "#I", READ_INDENT, "Read( \"", name, "\" )\n" );
+    fi;
     found := (IsReadableFile(name)=true) and READ(name);
     READ_INDENT := readIndent;
-    if found and READ_INDENT = ""  then
-        InfoRead1( "#I  Read( \"", name, "\" ) done\n" );
+    if GAPInfo.CommandLineOptions.D and
+       found and READ_INDENT = ""  then
+        Print( "#I  Read( \"", name, "\" ) done\n" );
     fi;
     if not found  then
         Error( "file \"", name, "\" must exist and be readable" );
--- a/lib/init.g
+++ b/lib/init.g
@@ -142,15 +142,6 @@ end;
 
 #############################################################################
 ##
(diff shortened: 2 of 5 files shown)


fingolfin:
hecke, guava and xgap are on GitHub, so we can send pull requests. For `atlasrep`, we need to talk to @ThomasBreuer -- it seems indent logical to use `Info` there, but then an info class is needed. Perhaps we should define an `InfoRead` class after all, and set it based on the initial value of `GAPInfo.CommandLineOptions.D` (we still couldn't use it in the early library initialization; also, it should probably be reset in a `PostRestore` method)

This leaves `gbnp`, which "only" uses this in `test/testall.g` -- but note: This is *not* the `TestFile` in `PackageInfo.g` -- that's `test/test-all.tst`. Nothing seems to reference `test/testall.g`, so I think that file can just be removed; and we don't need to worry about breaking `gbnp` with this change (though we should still notify its maintainers about this)

markuspf:
Re-added `InfoRead1` and `InfoRead2` to `obsoletes.gd` to prevent packages from breaking should we decide to merge this PR.

lib/init.g
```diff
@@ -259,7 +241,9 @@ CallAndInstallPostRestore( function()
 end);
 
 ReadOrComplete := function(name)
-    InfoRead1( "#I  reading ", name, "\n" );
+    if GAPInfo.CommandLineOptions.D then
+        Print( "#I  reading ", name, "\n" );
+    fi;
```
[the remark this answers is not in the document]
markuspf: I tried introducing an `InfoClass` for this purpose, but as you say the infrastructure is not initialised when we'd want to use it.

One could (independently of this PR) try moving the info infrastructure as early as possible in startup.

lib/files.gi
```diff
@@ -202,21 +202,28 @@ InstallMethod( Read,
...
-    fi;
-    if not found  then
+    if IsReadableFile(name) then
+        if GAPInfo.CommandLineOptions.D then
+            readIndent := SHALLOW_COPY_OBJ( READ_INDENT );
+            APPEND_LIST_INTR( READ_INDENT, "  " );
+            Print( "#I", READ_INDENT, "Read( \"", name, "\" )\n" );
+        fi;
+
+        READ(name);
```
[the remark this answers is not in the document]
markuspf: I now return the return value of `READ`. I also re-added the `UserHomeExpand` which got lost in the rewrite.

Note that we do not have to test whether the file is readable in `Read` as `FuncREAD` does this for us (and returns `False` if the file doesn't exist).

ThomasBreuer:
Concerning the use of `InfoRead1` in the `AtlasRep` package,
I will replace this by a mechanism of its own inside the package.
From this point of view, there is no need to introduce a new `Info` class.

lib/init.g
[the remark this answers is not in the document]
fingolfin: I think it is fine to leave this with a `Print` statement, but perhaps add a comment like
> We cannot use `Info` yet, because info.gi has not yet been read

lib/files.gd
```diff
@@ -10,6 +10,11 @@
 ##
 
 
+DeclareInfoClass("InfoRead");
```
fingolfin: Perhaps add a brief comment here, indicating what this is about?

lib/files.gd
```diff
@@ -10,6 +10,11 @@
...
+if GAPInfo.CommandLineOptions.D then
+    SetInfoLevel(InfoRead, 5);
```
fingolfin: Why the value 5? and not 1, or 3, or ...? Perhaps add a comment with the reason?

fingolfin: Alternatively, get rid of the Info class again, as it seems atlasrep won't need/use it anyway. (But I am also fine with keeping it, just saying we have both options)

fingolfin: BTW, I now wonder if we serialize info class levels when we save workspaces? If we do, then there might be a difference in behavior between checking for `GAPInfo.CommandLineOptions.D`, and checking the `InfoRead` level (i.e. when using `-D`, then saving a workspace, then loading it later on w/o `-D`;  and also vice-versa).

fingolfin:
I anybody working on PRs for the affected packages, resp. emailing the GBNP maintainer?

markuspf:
I will email the GBNP people. I think it's best to then remove `InfoRead` as its not used anymore.

fingolfin:
@markuspf did you already do that? One of the authors, Arjeh Cohen, is retired; the other is Jan-Willem Knopper, and I am not sure how interested he is in maintaining GBNP he is. Perhaps we can convince them to let us maintain GBNP for them, and move it to GitHub etc. I know both authors, as well as GBNP, and could also talk to them.

fingolfin:
Just for reference, I emailed Arjeh and Jan Willem, and JW replied and promised to work on an update.

markuspf:
Rebased, waiting for AppVeyor to get its act together.

I agree this should then be merged.

lib/obsolete.gd
```diff
@@ -682,6 +681,19 @@ DeclareObsoleteSynonym( "RecFields", "RecNames" );
...
 
+
+#############################################################################
+##
+#V  InfoRead? 
+##
+##  InfoRead used to be used to print when a file is read using `Read()`
+##  
+##  These variables are still referred to by atlasrep, gbnp, guava, hecke
+##  and xgap.
```
alex-konovalov: @markuspf please update this comment - only atlasrep and gbnp remain (hecke and xgap were done today by @fingolfin).

PR #2241, opened 2018-03-04
Add HPCGAP and ADDGUARDS2 to sysinfo.gap

markuspf:
This facilitates using it in package builds as prototyped in
the datastructures build system.


markuspf:
An alternative approach to running `ward` directly on package source files would be to allow `gac` to run ward (maybe if given a command line argument to do so).

PR #2242, opened 2018-03-05
Implement CyclesFromList, and fix potential crash in CYCLE_TRANS_INT

markusbaumeister:
Add a convenient functionality to generate a cycle from a list of
positive integers.

### DIFF
--- a/lib/permutat.g
+++ b/lib/permutat.g
@@ -481,6 +481,64 @@ end );
 
 #############################################################################
 ##
+#F  CycleFromList( <list> )  . . . . . . . . . . . cycle defined from a list 
+##
+##  <#GAPDoc Label="CycleFromList">
+##  <ManSection>
+##  <Func Name="CycleFromList" Arg='list'/>
+##
+##  <Description>
+##  For the given dense, duplicate-free list of positive integers
+##  <M>[a_1, a_2, ..., a_n]</M>
+##  return the <M>n</M>-cycle <M>(a_1,a_2,...,a_n)</M>. For the empty list
+##  the trivial permutation <M>()</M> is returned.
+##  <P/>
+##  If the given <A>list</A> contains duplicates or holes, return <K>fail</K>.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> CycleFromList( [1,2,3,4] );
+##  (1,2,3,4)
+##  gap> CycleFromList( [3,2,6,4,5] );
+##  (2,6,4,5,3)
+##  gap> CycleFromList( [2,3,2] );
+##  fail
+##  gap> CycleFromList( [1,,3] );
+##  fail
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+BIND_GLOBAL( "CycleFromList", function( list )
+    local max, images, set, i;
+
+    # Trivial case
+    if Length(list) = 0 then
+        return ();
+    fi;
+
+    if ForAny( list, i -> not IsPosInt(i) ) then
+        Error("CycleFromList: List must only contain positive integers.");
+    fi;
+
+    set := Set(list);
+    if Length(set) <> Length(list) then
+        # we found duplicates (or list was not dense)
+        return fail;
+    fi;
+    max := Maximum( set );
+    images := [1..max];
+    for i in [1..Length(list)-1] do
+        images[ list[i] ] := list[i+1];
+    od;
+    images[ list[Length(list)] ] := list[1];
+
+    return PermList(images);
+end );
+
+
+#############################################################################
+##
 #O  RestrictedPerm(<perm>,<list>)  restriction of a perm. to an invariant set
 #O  RestrictedPermNC(<perm>,<list>)  restriction of a perm. to an invariant set
 ##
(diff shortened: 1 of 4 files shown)


lib/permutat.g
```diff
@@ -479,6 +479,63 @@ BIND_GLOBAL( "ListPerm", function( arg )
...
+BIND_GLOBAL( "CycleFromList", function( list )
+    local max, images, set, i;
+
+    # Trivial case
+    if Length(list) = 0 then
+        return ();
+    fi;
+
+    set := Set(list);
+    if not IsDenseList(list) then
```
ChrisJefferson: I don't think you need this check, as the next check will fail if the list is not sense (as the set will be smaller).

ChrisJefferson: Also, maybe check Minimum is > 0

lib/permutat.g
```diff
@@ -479,6 +479,64 @@ BIND_GLOBAL( "ListPerm", function( arg )
...
+##  <ManSection>
+##  <Func Name="CycleFromList" Arg='list'/>
+##
+##  <Description>
+##  For the given dense, duplicate-free list of positive integers
+##  <M>[a_1, a_2, ..., a_n]</M>
+##  return the <M>n</M>-cycle <M>(a_1,a_2,...,a_n)</M>. For the empty list
+##  the trivial permutation <M>()</M> is returned.
+##  <P/>
+##  If the given <A>list</A> contains duplicates or holes, return <K>fail</K>.
```
fingolfin: And if there is a something that is not a positive integer in the list? Right now, probably various errors, depending on whether its a negative integer (gives proper error), a string or some such (will run into a method not found at some point), a rational (same, but likely in a different) spot, ...

lib/permutat.g
```diff
@@ -479,6 +479,64 @@ BIND_GLOBAL( "ListPerm", function( arg )
...
+
+    set := Set(list);
+    if Length(set) <> Length(list) then
+        # we found duplicates (or list was not dense)
+        return fail;
+    fi;
+    min := Minimum( set );
+    if min <= 0 then
+        Error("CycleFromList: List must contain only positive integers.");
+    fi;
```
fingolfin: Instead of this, perhaps do `if not ForAll(list, IsPosInt) then Error(...); fi;` even before computing `set`?

lib/permutat.g
```diff
@@ -479,6 +479,64 @@ BIND_GLOBAL( "ListPerm", function( arg )
...
+    max := Maximum( set );
+    images := [1..max];
+    for i in [1..Length(list)-1] do
+        images[ list[i] ] := list[i+1];
+    od;
+    images[ list[Length(list)] ] := list[1];
+
+    return PermList(images);
```
fingolfin: While manual test cases are provided, and exercise the regular parts of the function, not all failure cases are tested, e.g. `CycleFromList([0])`, `CycleFromList([1/2])`, `CycleFromList([true])` to name a few. Perhaps add some (though perhaps to a .tst file, not the manual example)?

I'll not block the PR over this, but it is needed to maximize the test coverage.

fingolfin: There is also no test for the empty list. All this can be seen on <https://codecov.io/gh/gap-system/gap/pull/2242/diff?src=pr&el=tree#diff-bGliL3Blcm11dGF0Lmc=> (link comes from the codecov report in the PR)

markusbaumeister: I don't know much about the test directory structure of GAP. Where exactly should I put the tests?

fingolfin:
I suspect the test failure is due to a bug in `CYCLE_TRANS_INT`, and your PR is just by luck triggering the bug. Try if this patch (added to a new commit helps)
```patch
diff --git a/src/trans.c b/src/trans.c
index deb807172..f0ea6c1ea 100644
--- a/src/trans.c
+++ b/src/trans.c
@@ -3102,6 +3102,7 @@ Obj FuncCYCLE_TRANS_INT(Obj self, Obj f, Obj pt)
         i = cpt;
         do {
             AssPlist(out, ++len, INTOBJ_INT(i + 1));
+            ptf2 = CONST_ADDR_TRANS2(f);
             i = ptf2[i];
         } while (i != cpt);
     }
@@ -3116,6 +3117,7 @@ Obj FuncCYCLE_TRANS_INT(Obj self, Obj f, Obj pt)
         i = cpt;
         do {
             AssPlist(out, ++len, INTOBJ_INT(i + 1));
+            ptf4 = CONST_ADDR_TRANS4(f);
             i = ptf4[i];
         } while (i != cpt);
     }
```

lib/permutat.g
fingolfin: Somewhere into `tst/testinstall/`. In this case, I'd suggest to create `tst/testinstall/opers/CyclesFromList.tst`.  Look at the other files in that directory to get an idea what to put in, in case you are not familiar with .tst files. And feel free to ask more questions.

tst/testinstall/opers/CyclesFromList.tst
```diff
@@ -0,0 +1,33 @@
+gap> START_TEST("CyclesFromList")
```
fingolfin: This is missing a semicolon, hence the test fails.

You can test the .tst file locally using `Test()`, just do
```
Test("tst/testinstall/opers/CyclesFromList.tst");
```
(adjust the path depending on from where you run GAP)

PR #2224, opened 2018-03-01
Improve ShowImpliedFilters

fingolfin:
This improves `ShowImpliedFilters` to really show all automatic implications (which it often did not do before); and for the extra implication ("May imply with:"), only show the *additional* implications. Besides the example in the diff of this PR, here are some more.

Before this PR:
```
gap> ShowImpliedFilters(IsFFEMatrixGroup);


May imply with:
+IsFinitelyGeneratedGroup
+HasIsFinitelyGeneratedGroup
   IsListOrCollection
   IsCollection
   IsFinite
   HasIsFinite
   CanComputeSize
   IsDuplicateFree
   HasIsDuplicateFree
   IsExtAElement
   CategoryCollections(IsExtAElement)
   CategoryCollections(CategoryCollections(IsExtAElement))
   CategoryCollections(CategoryCollections(CategoryCollections(IsNearAdditiveElement)))
   CategoryCollections(CategoryCollections(CategoryCollections(IsNearAdditiveElementWithZero)))
   CategoryCollections(CategoryCollections(CategoryCollections(IsNearAdditiveElementWithInverse)))
   CategoryCollections(CategoryCollections(CategoryCollections(IsAdditiveElement)))
   IsExtLElement
   CategoryCollections(IsExtLElement)
   CategoryCollections(CategoryCollections(IsExtLElement))
   IsExtRElement
   CategoryCollections(IsExtRElement)
   CategoryCollections(CategoryCollections(IsExtRElement))
   CategoryCollections(IsMultiplicativeElement)
   CategoryCollections(CategoryCollections(CategoryCollections(IsMultiplicativeElement)))
   CategoryCollections(IsMultiplicativeElementWithOne)
   CategoryCollections(CategoryCollections(CategoryCollections(IsMultiplicativeElementWithOne)))
   CategoryCollections(IsMultiplicativeElementWithInverse)
   CategoryCollections(CategoryCollections(CategoryCollections(IsMultiplicativeElementWithInverse)))
   CategoryCollections(IsFiniteOrderElement)
   CategoryCollections(CategoryCollections(CategoryCollections(IsFFE)))
   IsGeneralizedDomain
   IsMagma
   IsMagmaWithOne
   IsMagmaWithInversesIfNonzero
   IsMagmaWithInverses
   IsAssociative
   HasIsAssociative
   HasMultiplicativeNeutralElement
   IsGeneratorsOfSemigroup
   HasIsGeneratorsOfSemigroup
   IsSimpleSemigroup
[...]

PR #2243, opened 2018-03-05
Omit tester filters from `ShowImpliedFilters` output

ThomasBreuer:
This is a follow-up to #2224.  As discussed there,
those property tester filters are omitted from the output
for which the property itself is already shown.

### DIFF
--- a/lib/methwhy.g
+++ b/lib/methwhy.g
@@ -273,7 +273,7 @@ end);
 ##  <Description>
 ##  Displays information about the filters that may be
 ##  implied by <A>filter</A>. They are given by their names.
-##  <C>ShowImpliedFilters</C> first displays the names of all filters
+##  <Ref Func="ShowImpliedFilters"/> first displays the names of all filters
 ##  that are unconditionally implied by <A>filter</A>. It then displays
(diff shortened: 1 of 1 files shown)


lib/methwhy.g
```diff
@@ -322,20 +319,24 @@ local flags, implied, f, extra_implications, implication, name, diff_reqs, diff_
     implied := SUB_FLAGS(implied, flags);
   fi;
 
+  reduced:= trues -> Filtered( trues,
+      i -> not ( INFO_FILTERS[i] in FNUM_TPRS
+                 and FLAG1_FILTER( FILTERS[i] ) in trues ) );
```
fingolfin: For HPC-GAP, this requires `atomic readonly FILTER_REGION do ... od` around the access to `INFO_FILTERS ` and `FILTERS`, I am afraid. So (modulo typos):
```gap
reduced := functon(trues)
  atomic readonly FILTER_REGION do
    return not ( INFO_FILTERS[i] in FNUM_TPRS
                and FLAG1_FILTER( FILTERS[i] ) in trues ) );
  od;
end;
```

PR #2246, opened 2018-03-06
Introduce -N switch to not use hidden implications

frankluebeck:
If GAP is normally called then this patch only causes
 - that the kernel list HIDDEN_IMPS is visible from GAP level
 - that two debugging functions are available
     - WriteFilterRanks(filename);
     - WriteMethodOverview(filename);

Furthermore a command line switch -N is introduced. When gap is called with
this -N then:
  - no hidden implications are installed anymore
  - RankFilter only uses proper implications
  - InstallMethod uses only proper implication during its check if
    the installation fits to one declaration. When it does not find
    an appropriate declaration then it does not run into an error but only
    prints a line of useful information. (This allows to systematically
    either add explicit filter implications or change the filters in some
    method installations until these messages disappear.)

This patch should be safe to be merged into master. Actual changes in
filter implications will be provided in separate patches.


frankluebeck:
This could also be interesting in the context of pull request #2222.

lib/oper1.g
```diff
@@ -515,12 +519,27 @@ BIND_GLOBAL( "INSTALL_METHOD",
         # If the requirements do not match any of the declarations
         # then something is wrong or `InstallOtherMethod' should be used.
         if notmatch=0 then
-          Error("the number of arguments does not match a declaration of ",
-                NAME_FUNC(opr) );
+          if not GAPInfo.CommandLineOptions.N then          
+            Error("the number of arguments does not match a declaration of ",
+                  NAME_FUNC(opr) );
+          else
+            Print("IM: the number of arguments does not ",
```
fingolfin: Why "IM: "?

ChrisJefferson:
So I had a look at this, it looks fine, the only problem is the same changes need making to HPC-GAP, else (at the moment) this breaks HPC-GAP from starting.

fingolfin:
I am also somewhat unhappy about using up a single-letter option for this rather obscure debugging tool. How about instead naming it, `--disable-hidden-implications` or `--no-hidden-imps` or so?

frankluebeck:
@ChrisJefferson: I think it is just not needed that this (hopefully short term) temporary switch works with HPCGAP.

@fingolfin: I chose -N because it was already there (but not used for anything). Since the switch is really only meant for temporary debugging I would prefer to not put more effort in introducing a new switch in all necessary places.

lib/oper1.g
fingolfin: Perhaps "IM" is meant to be a shorthand for `InstallMethod`?

If one sees this message as a user, it's a bit unclear what it implies. Perhaps this would be better? "Warning, the number of ..."

lib/operdebug.g
```diff
@@ -0,0 +1,69 @@
...
+    fi;
+    return Concatenation(":",String(n),"\c");
+  end;
+  f := function()
+    local nam, m, s, op, i, j;
+    for op in OPERATIONS do
+      nam := NameFunction(op);
+      for i in [0..6] do
+        m := METHODS_OPERATION(op, i);
+        s := 4+i;
```
fingolfin: This needs to be changed to `s := BASE_SIZE_METHODS_OPER_ENTRY + i;`

lib/operdebug.g
```diff
@@ -0,0 +1,69 @@
...
+        for j in [1..Length(m)/s] do
+          Print(nam,";\c",i,";",
+                name(m[j*s-2]),
+                sline(m[j*s-2]),";\c",
```
fingolfin: This writes out the location of the definition of the *function*, if available. It does not yet take advantage of the location of the *method installation*, which we now also track (see e.g. the code of `ApplicableMethodTypes`)

lib/operdebug.g
fingolfin: I also suggest rebasing this PR on master and verifying these functions still work. Ideally, there'd be a test case for them, too, just to make sure we don't break them horribly. Of course one may argue they are just debugging tools; but also debugging tools are more useful if in order to use them you don't first have to track down what caused a regression in them some time in the past ;-).

src/opers.c
```diff
@@ -4394,6 +4394,7 @@ static Int InitLibrary (
     SET_LEN_PLIST(HIDDEN_IMPS, 0);
     WITH_HIDDEN_IMPS_FLAGS_CACHE = NEW_PLIST(T_PLIST, HIDDEN_IMPS_CACHE_LENGTH * 2);
     SET_LEN_PLIST(WITH_HIDDEN_IMPS_FLAGS_CACHE, HIDDEN_IMPS_CACHE_LENGTH * 2);
+    AssGVar(GVarName("HIDDEN_IMPS"), HIDDEN_IMPS);
```
fingolfin: Why is this needed?

lib/filter.g
```diff
@@ -165,7 +165,9 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
         fi;
       fi;
     od;
-    InstallHiddenTrueMethod( tofilt, from );
+    if not GAPInfo.CommandLineOptions.N then
+      InstallHiddenTrueMethod( tofilt, from );
+    fi;
```
fingolfin: Wouldn't it be much easier to put the check for the command line option into `InstallHiddenTrueMethod`? Then one can even keep calling `WITH_HIDDEN_IMPS_FLAGS` below, as we'd be guaranteed that there are not hidden implications beyond the regular implications, and almost all changes on the library side could be removed.

src/opers.c
frankluebeck: To be able to see the content of `HIDDEN_IMPS` from GAP level.

lib/filter.g
frankluebeck: The `if GAPInfo.CommandLineOptions.N` are introduced such that it is easy to find the places where the behaviour is changed by the `-N` switch.

alex-konovalov:
@frankluebeck any reasons why this is not documented in [Chapter 3](http://www.gap-system.org/Manuals/doc/ref/chap3.html) ? 

Perhaps one should check whether there are more cases like this...

lib/system.g
```diff
@@ -92,7 +92,7 @@ BIND_GLOBAL( "GAPInfo", rec(
       rec( short:= "B", default := "",    arg := "<name>", help := [ "current architecture"] ),
       rec( short:= "D", default := false, help := ["enable/disable debugging the loading of files"] ),
       rec( short:= "M", default := false, help := ["disable/enable loading of compiled modules"] ),
-      rec( short:= "N", default := false, help := ["unused, for backward compatibility only"] ),
+      rec( short:= "N", default := false, help := ["do not use hidden implications"] ),
```
alex-konovalov: @frankluebeck The comment above says "These options must be kept in sync with those in system.c" but that file is not even modified by this PR. Am I missing something?

frankluebeck:
@alex-konovalov: As discussed above, this -N switch is a temporary change, just for debugging. This should not be documented. Once the "hidden implications" are no longer used we can revert the change in 'system.g' (and simplify the few places where the -N switch is used).

PR #2247, opened 2018-03-06
Add explicit implications

frankluebeck:
WORK IN PROGRESS

This is a start to add explicit implications to the GAP library that can
substitute the undocumented "hidden implications" which GAP uses so far for
the computation of filter ranks and when checking if installed methods
correspond to appropriate declarations.

Here, we only add implications which are explicitly mentioned in the
documentation.

It can be debugged (see branch fl_nohiddenamps) that the changes here have
almost no impact on the relative ranks of methods. Using the -N switch
from the mentioned branch one can call gap -A -N without any messages from
InstallMethod. But there is still some work to do because many filters have
a much lower rank without hidden implications and a few rank changes of
methods may need to be fixed.


frankluebeck:
The motivation of this patch is similar to that of #2222. There may be some intersection (but hopefully no conflict). 

Using the debugging switch -N provided by #2246 and this patch one can call `gap -A -N` without complaining messages from `InstallMethod`. But there is still something to do to get all filter ranks to similar values as before without hidden implications.

fingolfin:
This has been rendered obsolete by the merge of PR #2222

PR #2250, opened 2018-03-07
Add tests for interpreter edge cases and deep thought

fingolfin:


fingolfin:
Seems codecov comments are broken again... But here is the direct link: <https://codecov.io/gh/gap-system/gap/pull/2250>. Coverage goes up from 70.38% to 70.56%.

tst/testinstall/interpreter.tst
```diff
@@ -0,0 +1,127 @@
...
+function() quit; end;
+               ^
+
+#
+# return is not allowed in interpreter
+#
+gap> return;
+'return' must not be used in file read-eval loop
+gap> return 1;
+'return' must not be used in file read-eval loop
```
fingolfin: This message differs from what on gets when entering the command interactively, which is this:
```
gap> return;
'return' cannot be used in this read-eval-print loop
```
And `READ_GAP_ROOT()` prints yet another message:
```C
                if ( type & (STATUS_RETURN_VAL | STATUS_RETURN_VOID) ) {
                    Pr( "'return' must not be used in file", 0L, 0L );
                }
```
I think we should unify this. Also, we print similar message in 5 locations (or 6, depending on how you count the on in `Shell()` in `gap.c`), which smells like a code refactoring opportunity. For starters, couldn't `READ_GAP_ROOT` call `READ_INNER()` instead of implementing yet another REPL?

fingolfin: To answer my last question, I guess the reason is that `READ_GAP_ROOT` tries very hard to read the whole file, trying to skip over errors, while `READ_INNER()` aborts on the first error. Still, I think there is a chance to unify the two, e.g. by adding another arg to `READ_INNER` (or a third value for the existing arg) to indicate that it should ignore all errors.

PR #2255, opened 2018-03-08
Remove special treatment of 'carat' package (backport to 4.9!)

fingolfin:
Let's see how that works out...

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -146,46 +146,6 @@ echo_run() {
   "$@"
 }
 
-build_carat() {
-(
-# TODO: FIX Carat
-# Installation of Carat produces a lot of data, maybe you want to leave
-# this out until a user complains.
-# It is not possible to move around compiled binaries because these have the
-# path to some data files burned in.
-zcat carat-2.1b1.tgz | tar pxf -
-# If the symbolic link does not exist then create it.
-# Does not check if bin exists already with some content or as a file, etc.
-# Thus this code is unstable, but should be corrected by package author
-# and not by this script.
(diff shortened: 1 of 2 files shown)


fingolfin:
This seems to work. However, log output still is quite a bit, with about 2400 lins (from line ~2600 till ~5000). This results in the Travis log being truncated (Travis imposes a total limit of 10,000 lines, though the raw log has them all). See https://travis-ci.org/gap-system/gap/jobs/350740345 for the build log.

However, the truncated part only contains info about the uploading of coverage data, so for now, I guess it would be OK. Still, 25% of the whole log is a lot. I think this could be improved substantially by somehow convincing carat to disable all warnings (via the `-w` compiler flag), as those make up a considerable part of the build output. Note that I am not suggesting that carat should do this in general, just that we could try to do this for the Travis builds of carat.
I guess we might be able to achieve this via custom `CFLAGS` for carat.

fingolfin:
Update to PR tries this.

fingolfin:
Actually, I just realized I got the logs wrong. The carat build already ends around line 3340, which is fine!

So, I think merging this PR in its original form is fine. And then a few other packages (e.g. cohomolo should reduce the number of warnings they produce...)

fingolfin:
This should be backported / cherry picked to stable-4.8

fingolfin:
Several other package produces lots of warnings and output, too, and we could perhaps save the 1500 lines there with suitable fixes...

fingolfin:
Cherry picked to stable-4.9 as 57a138f4c3676aaedb8a1c52525f09434c2de772

Issue #2257, opened 2018-03-10
GAP to C compilation fails for tilde expressions and call options using expression


fingolfin:
The following code snippets cause `gac` to produce C code which does not compiler, because it contains lines starting with `PROBLEM`; this comes from a check in `FreeTemp()` in `src/compiler.c`. Perhaps @stevelinton or @ChrisJefferson have some insights on this?

For the call options using expressions, this snippet triggers it twice (in different parts of the compiler):
```gap
vararg_fun := function(args...)
    return Length(args);
end;
Print(vararg_fun(:("myopt")), "\n");
Print(vararg_fun(:("myopt"):="value"), "\n");
```
For reference, this works fine:
```gap
Print(vararg_fun(:myopt), "\n");
Print(vararg_fun(:myopt:="value"), "\n");
```

For the tilde expressions, this snippet triggers it three times (in different parts of the compiler):
```gap
# list tilde expression
x := [~];
Display(x);

# record tilde expression
x := rec( next := ~);
Display(x);

# tilde expression
x := [ [ 1, 2 ], ~[ 1 ] ];
Display(x);
```

ChrisJefferson:
So I certainly have no idea about this. I'm happy for someone else to look, but I'm tempted to say "won't fix", unless it turns out we greatly need it to work, or we want to push the compiler for use by more people/code.

Issue #2256, opened 2018-03-10
Uninterruptible loop


fingolfin:
This code snippet cannot be interrupted with ctrl-c:
```gap
repeat continue; until false;
```
This is present in master but also in GAP 4.8 and older. The problem also affects `for` loops and `while` loops.

One way to fix this would be to move `T_CONTINUE` out of the `FIRST_NON_INTERRUPT_STAT .. LAST_NON_INTERRUPT_STAT` range. Another would be to leave it in there, but add additional interrupt checks to all loop constructs. If we do that, we could then also move `T_EMPTY` into the range of non-interruptible stats.

ChrisJefferson:
Why would we want t_empty to be non-interuptable? I would think we'd want as much to be interupttable as reasonable?

fingolfin:
Well, `T_EMPTY` is a special case of `T_SEQ_STAT`, `T_SEQ_STAT2`, `T_SEQ_STAT3`, ... (i.e., it i `T_SEQ_STAT0`, if you will). My personal explanation for why we do not interrupt these (I could not find any documentation, not even in a commit message, explaining it), as well as other compound statements like loops and `if/than/else` statements is: because that leads to poor user experience in the break loop. To illustrate what I mean, first consider the example, where I interrupt a simple loop with a non-empty body:
```
gap> x:=0;; repeat x:=0; until false;
^CError, user interrupt in
  x := 0; at *stdin*:2 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:2
you can 'return;'
brk>
```
As you can see, in the break loop, we try to print the statement that was interrupted. Which makes a lot of sense for single statements like `x:=0`. In contrast, if we interrupt a `T_SEQ_STAT`, it wouldn't be to helpful to print all the statements in it. So instead, we do not interrupt it, and instead rely on the sub-statements to be interrupted. For the same reason, loops like `T_REPEAT` are not 

Now consider what happens for a `T_EMPTY` statement:
```
gap> x:=0;; repeat until false;
^CError, user interrupt in
  ; at *stdin*:2 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:2
you can 'return;'
brk>
```
It prints out the empty statement, which is not very helpful. If instead the loop code handled the interrupt (and I don't mean making `T_REPEAT` etc. interruptible, I mean making it explicitly handle the interrupt in a "clever" way),  then it could print something more intelligent.

[...]

PR #2258, opened 2018-03-10
Extend code coverage further, mostly for gac

fingolfin:
This PR adds many new tests for the GAP compiler, which should increase code coverage, and also revealed several bugs, see issues #2256 and #2257.

This PR also removes `FillInVersion()` from the kernel. No distributed package (nor any other I know of) still uses this, with one exception: ParGAP, which however is unmaintained and already suffers from other compilation issues. However, I actually did remove use of `FillInVersion` from https://github.com/gap-packages/pargap, but no release has been made from there since many other things remain broken.


fingolfin:
Perhaps instead of putting this all in a single file, there should be multiple files, each corresponding to one of the tests? Seems mostly like a matter of taste, though, unless somebody can think of good reasons for one or the other.

PR #2259, opened 2018-03-10
Make `break`, `continue`, and `return` interruptible

### LINKED ISSUE #2256: Uninterruptible loop

fingolfin:
Fixes #2256 

We may want to backport this to stable-4.9

I don't see a good way to add a test for this.

### DIFF
--- a/src/code.h
+++ b/src/code.h
@@ -137,13 +137,13 @@ enum STAT_TNUM {
         T_EMPTY,
 
         // The statement types between FIRST_NON_INTERRUPT_STAT and
-        // LAST_NON_INTERRUPT_STAT will not be interrupted (which may
-        // happen for two reasons: the user interrupted, e.g. via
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.9 as ea1d381a4c9769dac5115dbd6748f906b71477e5

PR #2260, opened 2018-03-11
Fix TriangulizedMat([])

fingolfin:
This PR ensures that `TriangulizedMat([])` returns an empty list / matrix, instead of triggering an error.

This in turn fixes a regression in the modisom package in GAP 4.9 (compared to GAP 4.8 and before), see https://github.com/gap-packages/modisom/issues/2

This change should be cherry-picked to stable-4.9

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -2949,6 +2949,9 @@ local m;
   return m;
 end);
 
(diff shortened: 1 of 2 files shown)


fingolfin:
Backported to stable-4.9 via 189635d2709a63b0ff082b046f19b4d3b394e3e9

PR #2263, opened 2018-03-12
Test start GAP only after packages have been downloaded

fingolfin:
Also, test a package banner

### DIFF
--- a/tst/testinstall/package.tst
+++ b/tst/testinstall/package.tst
@@ -44,6 +44,81 @@ gap> for entry in Set( Concatenation( Concatenation( [ sml, equ ] ) ) ) do
 > od;
 gap> ReadPackage("packagename");
 Error, packagename is not a filename in the form 'package/filepath'
+
+#
+# Test the default package banner
+#
+gap> oldTermEncoding := GAPInfo.TermEncoding;;
+gap> GAPInfo.TermEncoding := "ISO-8859-1";; # HACK
+
+#
+gap> Display(DefaultPackageBannerString(rec()));
+-----------------------------------------------------------------------------
+
+-----------------------------------------------------------------------------
+
+
+#
+gap> pkginfo := rec(
+>         PackageName := "TestPkg",
+>         Version := "1.0",
+>         PackageWWWHome := "https://www.gap-system.org",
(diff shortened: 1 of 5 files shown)


tst/testinstall/package.tst
```diff
@@ -44,6 +44,81 @@ gap> for entry in Set( Concatenation( Concatenation( [ sml, equ ] ) ) ) do
 > od;
 gap> ReadPackage("packagename");
 Error, packagename is not a filename in the form 'package/filepath'
+
+#
+# Test the default package banner
+#
+gap> oldTermEncoding := GAPInfo.TermEncoding;;
+gap> GAPInfo.TermEncoding := "ISO-8859-1";; # HACK
```
ssiccha: Why is this hack needed here?
I see that then the ` sep:= ListWithIdenticalEntries( len, '-' );` is called. My editor says that this file is utf-8 encoded anyways. Do the `Test*` functions not like unicode characters?

fingolfin: Sorry, did not see this comment before. The hack is needed because the unicode character used as a dash takes up three bytes in UTF-8 encoding. But most of GAP is not yet unicode aware, and thus measure the screen width of a string naively as its length in bytes. The result is that when one restricts the output line length to e.g. 80 (as the tests do), then the dash line produced if the terminal encoding is set to UTF-8 looks really strange, like so:
```
gap> DefaultPackageBannerString(rec());
"─────────────────────────────────────────────────────────────────────────────�\
�─────────────────────────────────────────────────────────────────────────────�\
�─────────────────────────────────────────────────────────────────────────────\n\
\n─────────────────────────────────────────────────────────────────────────────�\
��─────────────────────────────────────────────────────────────────────────────�\
��─────────────────────────────────────────────────────────────────────────────\
\n"
```
What happens here is that GAP decides to wrap an "overlong" output line; but (a) this is done way to early (after output 30 bytes, the user sees 10 characters, but GAP thinks its 30), and (b) it breaks an UTF-8 character in the middle, leading to output with broken UTF-8 encoding.

Instead of trying to work around all that, it was easier to just hack the terminal encoding temporarily.

PR #2266, opened 2018-03-17
Include ./autogen.sh

saraedum:
There is no `configure` if you clone this repository.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -130,16 +130,18 @@ additional information about compilation)
 
 Change to the directory `gap-4.X.Y` (which you just created by unpacking).
 To get started quickly you may simply build GAP with default settings
-by issuing the two commands
+by issuing the commands
 
+    ./autogen.sh
     ./configure
     make
 
 (note that on BSD systems you have to call `gmake` instead of `make`).
 
-Both will produce a lot of text output. You should end up with a shell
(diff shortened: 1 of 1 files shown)


fingolfin:
First off, thanks for your contribution!

These instructions are meant to be usable for people who get a release version of GAP, and which hence most likely should not run autogen.sh (they may not have the required tools). 

I.e. I don't think we should merge this PR as is. Instead, we'll need some text which explains when to run autogen.sh and when not. Like "if you got this from git, run autogen.sh; if you got this from a stable GAP release, do not run it".

Alternatively, we could use a modified README for the release, but that'd require a complication for the release scripts....

saraedum:
@fingolfin Sure, I am used to tars already including the configure. But the ones in the "Releases" here at github do not seem to include this (tried with the 4.9.0 one). Is the 300MB+ download available at https://www.gap-system.org/Releases/index.html a source release?

saraedum:
Oh, I had not seen the header. So the ones at gap-system.org do actually require `./configure; make`.

saraedum:
I just realized that your `README.md` does contain the information necessary to build from a git clone.

fingolfin:
Yeah, it does. So, I think this PR can be closed. However, I am now looking into improving the interaction of README.md and INSTALL.md a bit.

Issue #2138, opened 2018-02-01
Allow meta-packages for distribution


sebasguts:
Some projects, for example CAP and homalg, consist of multiple packages. Currently these packages are distributed independently.

It would be nice if one could produce meta-packages, containing all packages belonging to project. Benefits would be the following:
* Coherent versions among the packages in a project
* Easier distribution of projects
* More structure in the `pkg` dir, as all packages of a project would be in a subfolder.

alex-konovalov:
To collect some aspects of this to think about:
- how such projects will be distributed and picked up for redistribution with GAP (individually, or as a single archive) 
- which new fields in `PackageInfo.g` files will be needed to reflect this
- how projects will be presented on the GAP website

alex-konovalov:
Instead of commenting on #2267 which I have no time to review now, I will post a quick note here. 

My concern at the moment is that any changes which will make possible to build packages and to load packages located in subdirectories should also work with the existing package redistribution mechanism, or allow its adaptation.

If a meta-package would have PackageInfo.g file and will have packages in subdirectories, then the current system for checking for package update would not require any modification at all. I will simply add new package for the redistribution, and will remove from the redistribution all packages that will now be part of this meta-package.

Scripts that produce auto-generated pages with package overviews for the GAP website would require some adjustments, but that does not seem to be an obstacle.

alex-konovalov:
(comes from #2267).

As I said above, one of the questions is **which new field(s) in PackageInfo.g files will be needed to reflect this**. For example, there could be an entry in `PackageInfo.g` which lists names of subpackages forming the project, for example

    PackageComponents:=["ExamplesForHomalg", "GaussForHomalg", "GradedModules",...],

Then package loading will traverse subdirectories ("ExamplesForHomalg-2017.09.02", "GaussForHomalg-2018.02.05" etc.) , but will assume that they contain a packages that has to be loaded if and only if they have `PackageInfo.g` file for a package whose name matches the one in `PackageComponents`. This will thus allow to reject `gh-pages` directory (since it will have `PackageInfo.g` which will provide a name of a meta-package, but not the one from `PackageComponents`).

alex-konovalov:
We need also to identify things which have to be updated following the proposed change:
* `bin/BuildPackages.sh` if meta package has a component that requires compilation (e.g. Gauss which is a part of homalg)

alex-konovalov:
To summarise, the following aspects should be decided for meta-packages:
- Loading in GAP
[...]

PR #2267, opened 2018-03-19
Add support for meta packages resp. packages with subpackages

sebasguts:
even if a PackageInfo.g is found already. This would allow
meta packages, i.e., folders containing more than one package,
to be distributed with GAP.

Currently, this is WIP belonging to #2138. It currently breaks the loading of GBNP, which could be easily fixed by renaming `version/PackageInfo.g` in the package into something like `version/PackageInfo_template`.

### DIFF
--- a/lib/package.gd
+++ b/lib/package.gd
@@ -162,6 +162,21 @@ DeclareGlobalFunction( "SetPackageInfo" );
 
 #############################################################################
 ##
+#F  AddPackageInfos( files, pkgdir, ignore, metapackagerecord... )
+##
+##  Function that adds package info records from the PackageInfo.g files
+##  from the list <files> to the GAPInfo.PackagesInfo list. <files> is a list
+##  of pairs, each pair consisting of the absolute path of a PackageInfo.g
+##  file, and the path of its containing directory, relative to <pkgdir>.
+##  <ignore> contains a list of packages to ignore.
+##  If <metapackagerecord> is set, it contains the package record of a
+##  metapackage, and all <files> refer to subpackages of that metapackage.
+##
+DeclareGlobalFunction( "AddPackageInfos" );
+
+
+#############################################################################
+##
 #F  InitializePackagesInfoRecords()
 ##
 ##  <ManSection>
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -2278,14 +2278,6 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     fi;
     return true;
     end;
-    
-    TestOption( record, "HasSubpackage", IsBool, "a bool" );
-
-    hasSubpackage := IsBound( record.HasSubpackage ) and record.HasSubpackage;
-
-    TestOption( record, "SubpackageOf", IsString, "a string" );
-
-    isSubpackage := IsBound( record.SubpackageOf );
 
     TestMandat( record, "PackageName",
         x -> IsString(x) and 0 < Length(x),
@@ -2298,6 +2290,13 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
         x -> IsString(x) and Length(x) = 10 and x{ [3,6] } = "//"
                  and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ),
         "a string of the form `dd/mm/yyyy'" );
+
+    TestOption( record, "HasSubpackage", IsBool, "a bool" );
+    hasSubpackage := IsBound( record.HasSubpackage ) and record.HasSubpackage;
+
+    TestOption( record, "SubpackageOf", IsString, "a string" );
+    isSubpackage := IsBound( record.SubpackageOf );
+
     if not isSubpackage then
       TestOption( record, "License",
           x -> IsString(x) and 0 < Length(x),
@@ -2306,13 +2305,13 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
       TestMandat( record, "ArchiveFormats", IsString, "a string" );
       TestOption( record, "TextFiles", IsStringList, "a list of strings" );
       TestOption( record, "BinaryFiles", IsStringList, "a list of strings" );
-      TestOption( record, "TextBinaryFilesPatterns", 
-          x -> IsStringList(x) and 
+      TestOption( record, "TextBinaryFilesPatterns",
+          x -> IsStringList(x) and
               ForAll( x, i -> Length(i) > 0 ) and
-              ForAll( x, i -> i[1] in ['T','B'] ),  
+              ForAll( x, i -> i[1] in ['T','B'] ),
           "a list of strings, each started with 'T' or 'B'" );
-      if Number( [ IsBound(record.TextFiles), 
-                  IsBound(record.BinaryFiles), 
+      if Number( [ IsBound(record.TextFiles),
+                  IsBound(record.BinaryFiles),
                   IsBound(record.TextBinaryFilesPatterns) ],
                 a -> a=true ) > 1 then
         Print("#W  only one of TextFiles, BinaryFiles or TextBinaryFilesPatterns\n");
@@ -2386,32 +2385,33 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     if hasSubpackage then
       # for metapackages, documentation is optional, hence so is PackageDoc;
       # but if it exists, we should of course still validate it
-      docTester := {record, name, type, typename} -> IsBound( record.(name) ) and TestOption(record, name, type, typename);
+      docTester := TestOption;
     else
       docTester := TestMandat;
     fi;
-    if docTester( record, "PackageDoc",
+    docTester( record, "PackageDoc",
           x -> IsRecord( x ) or IsRecordList( x ),
-          "a record or a list of records" ) then
+          "a record or a list of records" );
+    if IsBound( record.PackageDoc ) then
       if IsRecord( record.PackageDoc ) then
         list:= [ record.PackageDoc ];
       else
         list:= record.PackageDoc;
       fi;
       for subrec in list do
-        docTester( subrec, "BookName", IsString, "a string" );
+        TestMandat( subrec, "BookName", IsString, "a string" );
         if IsBound(subrec.Archive) then
           Print("#W  PackageDoc.Archive is withdrawn, use PackageDoc.ArchiveURLSubset instead\n");
         fi;
-        docTester( subrec, "ArchiveURLSubset", IsFilenameList,
+        TestMandat( subrec, "ArchiveURLSubset", IsFilenameList,
             "a list of strings denoting relative paths to readable files or directories" );
-        docTester( subrec, "HTMLStart", IsFilename,
+        TestMandat( subrec, "HTMLStart", IsFilename,
                     "a string denoting a relative path to a readable file" );
-        docTester( subrec, "PDFFile", IsFilename,
+        TestMandat( subrec, "PDFFile", IsFilename,
                     "a string denoting a relative path to a readable file" );
-        docTester( subrec, "SixFile", IsFilename,
+        TestMandat( subrec, "SixFile", IsFilename,
                     "a string denoting a relative path to a readable file" );
-        docTester( subrec, "LongTitle", IsString, "a string" );
+        TestMandat( subrec, "LongTitle", IsString, "a string" );
       od;
     fi;
     if     TestOption( record, "Dependencies", IsRecord, "a record" )
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -161,59 +161,124 @@ InstallGlobalFunction( SetPackageInfo, function( record )
 
 #############################################################################
 ##
-#F  FindPackageInfosInSubdirectories( pkgdir, name )
+#F  SubdirectoriesOfPath( path )
 ##
-##  Finds all PackageInfos in subdirectories of directory name in
-##  directory pkgdir, return a list of their paths.
+##  Returns all subdirectories of <path> as absolute paths.
 ##
-BindGlobal( "FindPackageInfosInSubdirectories", function( pkgdir, name )
-    local pkgpath, file, files, subdir;
-    pkgpath:= Filename( [ pkgdir ], name );
-    # This can be 'fail' if 'name' is a void link.
-    if pkgpath = fail then
-      return [];
+BindGlobal( "SubdirectoriesOfPath", function( path )
+  local name, file, subs;
+  subs := [];
+  path := Directory(path);
+  # iterate over directory contents in alphabetical order, to get identical
+  # behavior across all platforms
+  for name in Set(DirectoryContents(path)) do
+    if name in [ ".", ".." ] then 
+      continue;
     fi;
-
-    if not IsDirectoryPath( pkgpath ) then
-      return [];
+    file := Filename(path, name);
+    if IsDirectoryPath(file) then
+      Add(subs, file);
     fi;
-    if name in [ ".", ".." ] then
-      return [];
-    fi;
-
-    file:= Filename( [ pkgdir ],
-                      Concatenation( name, "/PackageInfo.g" ) );
-    if file = fail then
-      files := [];
-      # Perhaps some subdirectories contain `PackageInfo.g' files.
-      for subdir in Set( DirectoryContents( pkgpath ) ) do
-        if not subdir in [ ".", ".." ] then
-          pkgpath:= Filename( [ pkgdir ],
-                              Concatenation( name, "/", subdir ) );
-          if pkgpath <> fail and IsDirectoryPath( pkgpath )
-                              and not subdir in [ ".", ".." ] then
-            file:= Filename( [ pkgdir ],
-                Concatenation( name, "/", subdir, "/PackageInfo.g" ) );
-            if file <> fail then
-              Add( files,
-                    [ file, Concatenation( name, "/", subdir ) ] );
-            fi;
-          fi;
-        fi;
-      od;
-    else
-      files:= [ [ file, name ] ];
-    fi;
-    return files;
+  od;
+  return subs;
 end );
 
 
 #############################################################################
 ##
-#F  AddPackageInfo( files )
+#F  FindPackageInfosInSubdirectories( dir )
 ##
-BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
-    local file, record, pkgname, version;
+##  This function checks if there is a `PackageInfo.g` file in <dir> or,
+##  if there is no `PackageInfo.g` in <dir>, in any subdirectory of <dir>.
+##  The function returns a list of pairs, each containing the absolute path
+##  to the found PackageInfo.g, and the directory the PackageInfo.g file
+##  is contained in, relative to <dir>.
+##
+BindGlobal( "FindPackageInfosInSubdirectories", function( dir )
+    local name, file, files, subdir, subname;
+
+    ## Check if dir actually exists
+    if not IsDirectoryPath( dir ) then
+      return [];
+    fi;
+
+    name := SplitString( dir, "/" );
+    name := name[ Length( name ) ];
+
+    # trick: by passing the directory inside a list, `Filename` checks whether
+    # the file exists, and returns fail if not
+    file := Filename( [ Directory( dir ) ], "PackageInfo.g" );
+
+    if file <> fail then
+      return [ [ file, name ] ];
+    fi;
+
+    files := [];
+    # Perhaps some subdirectories contain `PackageInfo.g' files.
+    for subdir in SubdirectoriesOfPath( dir ) do
+      file := Filename( [ Directory( subdir ) ], "PackageInfo.g" );
+      if file <> fail then
+        subname := SplitString( subdir, "/" );
+        subname := subname[ Length( subname ) ];
+        Add( files,
+              [ file, Concatenation( name, "/", subname ) ] );
+      fi;
+    od;
+
+    return files;
+end );
+
+#############################################################################
+##
+#F  EnrichMetapackageRecord( subpackagerecord, metapackagerecord )
+##
+##  Enriches the PackageInfo of the metapackage with information from the
+##  subpackage. In particular, it makes the metapackage depend on the given
+##  subpackage, in the precise version contained in the metapackage. This is
+##  important to allow different versions of a metapackage to be installed and
+##  used in parallel.
+
+BindGlobal( "EnrichMetapackageRecord", function( subpackagerecord, metapackagerecord )
+    local version, name;
+    if not IsBound( metapackagerecord.Dependencies.NeededOtherPackages ) then
+        metapackagerecord.Dependencies.NeededOtherPackages := [];
+    fi;
+
+    name := subpackagerecord.PackageName;
+    version := Concatenation( "= ", subpackagerecord.Version );
+
+    Add( metapackagerecord.Dependencies.NeededOtherPackages, [ name, version ] );
+
+end );
+
+#############################################################################
+##
+#F  EnrichSubpackageRecord( subpackagerecord, metapackagerecord )
+##
+##  Enriches the PackageInfo of the subpackage with information from the
+##  metapackage. In particular, the SubpackageOf field in the PackageInfo is
+##  not set, it will be set to the metapackage name.
+
+BindGlobal( "EnrichSubpackageRecord", function( subpackagerecord, metapackagerecord )
+    if not IsBound( subpackagerecord.SubpackageOf ) then
+        subpackagerecord.SubpackageOf := metapackagerecord.PackageName;
+    fi;
+end );
+
+#############################################################################
+##
+#F  AddPackageInfos( files, pkgdir, ignore, metapackagerecord... )
+##
+##
+InstallGlobalFunction( AddPackageInfos, function( files, pkgdir, ignore, metapackagerecord... )
+    local file, record, pkgname, version, isSubpackage, name, dir;
+    if Length( metapackagerecord ) = 1 then
+      isSubpackage := true;
+      metapackagerecord := metapackagerecord[ 1 ];
+    else
+      isSubpackage := false;
+    fi;
+
     for file in files do
       # Read the `PackageInfo.g' file.
       Unbind( GAPInfo.PackageInfoCurrent );
@@ -224,6 +289,13 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
         pkgname:= LowercaseString( record.PackageName );
         NormalizeWhitespace( pkgname );
         version:= record.Version;
+
+        # If we have this version already then leave it out.
+        if ForAny( GAPInfo.PackagesInfo,
+                    r -> r.PackageName = record.PackageName
+                    and r.Version = version ) then
+          continue;
+        fi;
 
         # Check whether GAP wants to reset loadability.
         if     IsBound( GAPInfo.PackagesRestrictions.( pkgname ) )
@@ -241,6 +313,20 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
           elif IsRecord( record.PackageDoc ) then
             record.PackageDoc:= [ record.PackageDoc ];
           fi;
+
+          if IsBound( record.HasSubpackage ) and record.HasSubpackage = true then
+            dir := record.InstallationPath;
+            for name in SubdirectoriesOfPath( dir ) do
+              files := FindPackageInfosInSubdirectories( name );
+              AddPackageInfos( files, Directory( dir ), ignore, record );
+            od;
+          fi;
+
+          if isSubpackage then
+            EnrichMetapackageRecord( record, metapackagerecord );
+            EnrichSubpackageRecord( record, metapackagerecord );
+          fi;
+
           if IsHPCGAP then
             # FIXME: we make the package info record immutable, to
             # allow access from multiple threads; but that in turn
@@ -313,11 +399,10 @@ InstallGlobalFunction( InitializePackagesInfoRecords, function( arg )
       fi;
 
       # Loop over subdirectories of this package directory.
-      for name in Set( DirectoryContents( Filename( pkgdir, "" ) ) ) do
+      for name in SubdirectoriesOfPath( Filename( pkgdir, "" ) ) do
 
           ## Get all package dirs
-          files := FindPackageInfosInSubdirectories( pkgdir, name );
-
+          files := FindPackageInfosInSubdirectories( name );
           AddPackageInfos( files, pkgdir, ignore );
 
       od;
@@ -807,17 +892,19 @@ InstallGlobalFunction( PackageAvailabilityInfo,
       fi;
 
       # Locate the `init.g' file of the package.
-      if Filename( [ Directory( inforec.InstallationPath ) ], "init.g" )
-           = fail  then
-        LogPackageLoadingMessage( PACKAGE_WARNING,
-            Concatenation( "PackageAvailabilityInfo: cannot locate `",
-              inforec.InstallationPath,
-              "/init.g', please check the installation" ),
-            inforec.PackageName );
-        if not checkall then
-          continue;
+      if not IsBound( inforec.HasSubpackage ) or inforec.HasSubpackage = false then
+        if Filename( [ Directory( inforec.InstallationPath ) ], "init.g" )
+            = fail  then
+          LogPackageLoadingMessage( PACKAGE_WARNING,
+              Concatenation( "PackageAvailabilityInfo: cannot locate `",
+                inforec.InstallationPath,
+                "/init.g', please check the installation" ),
+              inforec.PackageName );
+          if not checkall then
+            continue;
+          fi;
+          skip:= true;
         fi;
-        skip:= true;
       fi;
 
       record_local:= StructuralCopy( record );
@@ -1657,16 +1744,18 @@ InstallGlobalFunction( LoadPackage, function( arg )
         # Notify the documentation (for the available version).
         LoadPackageDocumentation( info );
 
-        # Read the `init.g' files.
-        LogPackageLoadingMessage( PACKAGE_DEBUG,
-            "start reading file 'init.g'",
-            info.PackageName );
-        GAPInfo.PackageCurrent:= info;
-        ReadPackage( pkgname, "init.g" );
-        Unbind( GAPInfo.PackageCurrent );
-        LogPackageLoadingMessage( PACKAGE_DEBUG,
-            "finish reading file 'init.g'",
-            info.PackageName );
+        # Read the `init.g' files if package is not a metapackage.
+        if not IsBound( info.HasSubpackage ) or info.HasSubpackage = false then
+          LogPackageLoadingMessage( PACKAGE_DEBUG,
+              "start reading file 'init.g'",
+              info.PackageName );
+          GAPInfo.PackageCurrent:= info;
+          ReadPackage( pkgname, "init.g" );
+          Unbind( GAPInfo.PackageCurrent );
+          LogPackageLoadingMessage( PACKAGE_DEBUG,
+              "finish reading file 'init.g'",
+              info.PackageName );
+        fi;
 
         filename:= Filename( [ Directory( info.InstallationPath ) ],
                              "read.g" );
@@ -2130,7 +2219,7 @@ InstallGlobalFunction( DeclareAutoreadableVariables,
 InstallGlobalFunction( ValidatePackageInfo, function( info )
     local record, pkgdir, i, IsStringList, IsRecordList, IsProperBool, IsURL,
           IsFilename, IsFilenameList, result, TestOption, TestMandat, subrec,
-          list;
+          list, hasSubpackage, isSubpackage, docTester;
 
     if IsString( info ) then
       if IsReadableFile( info ) then
@@ -2189,6 +2278,14 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     fi;
     return true;
     end;
+    
+    TestOption( record, "HasSubpackage", IsBool, "a bool" );
+
+    hasSubpackage := IsBound( record.HasSubpackage ) and record.HasSubpackage;
+
+    TestOption( record, "SubpackageOf", IsString, "a string" );
+
+    isSubpackage := IsBound( record.SubpackageOf );
 
     TestMandat( record, "PackageName",
         x -> IsString(x) and 0 < Length(x),
@@ -2201,24 +2298,26 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
         x -> IsString(x) and Length(x) = 10 and x{ [3,6] } = "//"
                  and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ),
         "a string of the form `dd/mm/yyyy'" );
-    TestOption( record, "License",
-        x -> IsString(x) and 0 < Length(x),
-        "a nonempty string containing an SPDX ID" );
-    TestMandat( record, "ArchiveURL", IsURL, "a string started with http://, https:// or ftp://" );
-    TestMandat( record, "ArchiveFormats", IsString, "a string" );
-    TestOption( record, "TextFiles", IsStringList, "a list of strings" );
-    TestOption( record, "BinaryFiles", IsStringList, "a list of strings" );
-    TestOption( record, "TextBinaryFilesPatterns", 
-        x -> IsStringList(x) and 
-             ForAll( x, i -> Length(i) > 0 ) and
-             ForAll( x, i -> i[1] in ['T','B'] ),  
-        "a list of strings, each started with 'T' or 'B'" );
-    if Number( [ IsBound(record.TextFiles), 
-                 IsBound(record.BinaryFiles), 
-                 IsBound(record.TextBinaryFilesPatterns) ],
-               a -> a=true ) > 1 then
-      Print("#W  only one of TextFiles, BinaryFiles or TextBinaryFilesPatterns\n");
-      Print("#W  components must be bound\n");
+    if not isSubpackage then
+      TestOption( record, "License",
+          x -> IsString(x) and 0 < Length(x),
+          "a nonempty string containing an SPDX ID" );
+      TestMandat( record, "ArchiveURL", IsURL, "a string started with http://, https:// or ftp://" );
+      TestMandat( record, "ArchiveFormats", IsString, "a string" );
+      TestOption( record, "TextFiles", IsStringList, "a list of strings" );
+      TestOption( record, "BinaryFiles", IsStringList, "a list of strings" );
+      TestOption( record, "TextBinaryFilesPatterns", 
+          x -> IsStringList(x) and 
+              ForAll( x, i -> Length(i) > 0 ) and
+              ForAll( x, i -> i[1] in ['T','B'] ),  
+          "a list of strings, each started with 'T' or 'B'" );
+      if Number( [ IsBound(record.TextFiles), 
+                  IsBound(record.BinaryFiles), 
+                  IsBound(record.TextBinaryFilesPatterns) ],
+                a -> a=true ) > 1 then
+        Print("#W  only one of TextFiles, BinaryFiles or TextBinaryFilesPatterns\n");
+        Print("#W  components must be bound\n");
+      fi;
     fi;
     if     TestOption( record, "Persons", IsRecordList, "a list of records" )
        and IsBound( record.Persons ) then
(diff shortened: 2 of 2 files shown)


ChrisJefferson:
I like looking in subdirectories, but not if they contain a PackageInfo.g.

In particular, all packages which use the standard 'gh-pages' config have `gh-pages/PackageInfo.g` inside their directories.

sebasguts:
True. What would you suggest instead? Check if there is also an `init.g` and `read.g`.

ChrisJefferson:
That seems like a seperate issue, which we could do generally.

I don't really understand why you really want to put packages inside packages, as opposed to having a normal directory, which contains packages in subdirectories (and could also still contain a readme, and such things, of course, just not a PackageInfo.g)

sebasguts:
Well, those meta-packages should have a versioning, and some information about Authors, etc. Best thing would be a `PackageInfo.g`, or maybe a new `MetapackageInfo.g`?

ChrisJefferson:
I'm not sure what GAP knowledge GAP would have about these metapackages. This sounds like it would need planning. What should a metapackage do?

If you just want a package that loads everything in your metapackage, you can still just make that a normal package, and put it into your directory of packages.

Remember that (without significant other changes) if two metapackages contain the same package, GAP will just load the one of highest version (I don't know how GAP breaks ties.. I'd guess either first or last one found?)

sebasguts:
> This sounds like it would need planning. What should a metapackage do?
I think so. So, here is the backstory, along with the stuff I want to do:

Homalg and CAP both consists of multiple packages. Currently, those packages are in one git repo, but need to be shipped separately to GAP. This causes issues:
* Incompatible versions
* Complicated release process
* No consistent versioning (Technically, for computations with CAP to be reproducible, one would have to specify 4 package versions, and for homalg aroung 10).

I think it would be very desirable to be able to ship those projects as one tarball to GAP, and I think one of the most sensible ways (which keeps the modularity of the projects) would be to have some kind of meta-package.

For CAP, this would be called CAP_project. In my opinion, the meta package should have the following features:
* a version
* authors
* a released tarball
* and it should be placed in a folder in the GAP package directory.

For myself, a proper PackageInfo would not be necessary, just some file that holds some of the information, e.g.,
* Version
* Release-tarball URL
* Authors
* Issue tracker URL
* Maybe a list of packages in the meta-package, probably also with versions.

I am not sure if this would have to be in a PackageInfo, this was suggested by @alex-konovalov to be conform with the current pick up and release process.
@alex-konovalov could you comment on this?

alex-konovalov:
@sebasguts will look. Put myself as reviewer to keep it in todo list.

fingolfin:
@sebasguts you give a list after "the meta package should have the following features"; then you give a different list right after, which I found a bit confusing... Perhaps the first list is for required fields, the second  for "optional" ones?!?

Anyway, II think that at least from a technical point of view, it would be easiest to get this going if we reuse the existing `PackageInfo.g` system. To avoid the problems with that, I'd add a new optional `IsMetaPackage` field into the PackageInfo record (with default value `false`, obviously). Then, traverse into subdirectories of a package *only* if `IsMetaPackage` is set to true. This means that it can only be done later in the function, when we already are reading the `PackageInfo.g` files. But no problem, we can simply put the code iterating over subdirs into a subfunction and then call that in more than one place.

I would also consider adding a `IsSubpackage` entry into the bundled packageinfo files... or perhaps better, a `ParentPackage := NAME` entry? Only question is whether to ask the package authors to do that (easy to implement; but error prone, as it can easily be forgotten, or be wrong in other ways). Or do it automatically -- makes the loading code in GAP yet again more complicated, but wouldn't be that hard... 
Thinking about it, I am tempted to strongly suggest the latter course, and up it by one: iterate over each subpackage, and insert a `ParentPackage := parent_pkg_info_rec` entry, so that it is truly linked to the *precise* parent package; and conversely, add to the meta package a `ChildPackages := [ child1_info_rec, child2_info_rec, ...]` list. Then, we could use that when loading a meta package to get the *precise* info files. This would prevent any kind of version mismatch, at least when the user does a `LoadPackage("CAP_project")`; if they load individual sub packages, things can still go wrong, but IMHO that's OK.

Anyway, going with `PackageInfo.g` for the meta package means that GAP will "see" the packages distributed this way right away, with relatively few changes needed. The main remaining problem then is to determine how the distribution tooling will interact with that, and how/where to change it. Some thoughts:

* for fetching package updates, we'd only want the meta package listed in the [currentPackageInfoURLList](https://github.com/gap-system/gap-distribution/blob/master/DistributionUpdate/PackageUpdate/currentPackageInfoURLList). This shouldn't be hard to ensure, might even happen automatic without any change (as the existing tooling probably does not descend into the tarball looking for further PackageInfo.g files)

* package list on the GAP website: Do we want an entry for all subpackages there, or just for the meta package? My guess is (which might be totally wrong) that w/o further changes, only the meta package would be shown there. So that code at least may need to be taught to iterate over subpackages. But (again without actually having looked at the code yet), I'd hope that this is a relatively easy matter.

* is anything else potentially affected? @alex-konovalov probably knows this best by far, so I hope he'll chime in when he finds some time.

In closing, I really think that many fields in the meta package can make sense; e.g. the people/person fields can be used to specify a central maintainer team, like "the homalg team" with a support email address; there could be a meta package documentation, which gives an overview of all packages it bundles, with links to their documentation; etc.; of course we can forbid certain fields for it, but I'd not do that unless we find a specific reason to do so. We could (should?), however, consider making more currently required fields optional for meta packages (which would be easy to implement with the `IsMetapackage` boolean field)

ChrisJefferson:
One thing to consider with metapackages -- how to implement `BuildPackages.sh`. We will want to look for packages inside metapackages, but not inside non-metapackages. It would make things considerably more complicated if we had to use GAP to decide where to look for packages to build, and fragile (but possible) if we tried looking for `IsMetaPackage := true` using (say) grep, as we might not match the string exactly (although if we used a multiline grep, we'd probably be able to get almost any valid GAP).

fingolfin:
I am not so worried about that. `BuildPackages.sh` is a hacked up crutch in the end whatever we do (no offense to anybody who invested effort in it, but that's how I view it). 

So, from a purely pragmatic point of view, I don't think a grep for `IsMetaPackage := true` (with some regex for the whitespace) would be a problem at all. Sure, some meta packager could do something crazy to obfuscate it, or a non-meta package could include it say hidden away in a comment) just to annoy us -- but then, so what? It'll be noticed immediately by package tests, and we'll just reject such package updates.

Of course this is not a good long-term solution, but neither is `BuildPackages.sh` to start with. Long term, I want a toll which I can just point at an URL and say  `gap-pkg-manager install https://PACKAGE.homepage`. And similar for updating a package, or just rebuilding it.

BTW, using GAP to parse those `PackageInfo.g` files from `BuildPackages.sh` also doesn't sound to bad to me -- I do this in the release tools, after all, and so far nobody complained much? If you want to avoid the overhead of starting GAP N times when building N packages, no problem: once the list of package is clear, we can start GAP once, scan all PackageInfo.g files, and determine which are meta packages; then recursively scan the subpackages etc.
We might want to do that on the road to the for the future better tool anyway, in case we plan to include a way to specify build instructions inside `PackageInfo.g`.

Or perhaps those build instructions will go into a new file, say `gap-build.json`. But if we do that, then this file might as well indicate whether this is a meta packages, too (or it could just list subpackages that need compilation, if any, to save us one step).

alex-konovalov:
As I said in https://github.com/gap-system/gap/issues/2138 : 

> My concern at the moment is that any changes which will make possible to build packages and to load packages located in subdirectories should also work with the existing package redistribution mechanism, or allow its adaptation.
> If a meta-package would have PackageInfo.g file and will have packages in subdirectories, then the current system for checking for package update would not require any modification at all. I will simply add new package for the redistribution, and will remove from the redistribution all packages that will now be part of this meta-package.
> Scripts that produce auto-generated pages with package overviews for the GAP website would require some adjustments, but that does not seem to be an obstacle.

I think we should continue discussion at https://github.com/gap-system/gap/issues/2138 before revising this PR. One of the questions is **which new field(s) in PackageInfo.g files will be needed to reflect this**. For example, there could be an entry in `PackageInfo.g` which lists names of subpackages forming the project, for example

    PackageComponents:=["ExamplesForHomalg", "GaussForHomalg", "GradedModules",...],

Then package loading will traverse subdirectories ("ExamplesForHomalg-2017.09.02", "GaussForHomalg-2018.02.05" etc.) , but will assume that they contain a packages that has to be loaded if and only if they have `PackageInfo.g` file for a package whose name matches the one in `PackageComponents`. This will thus allow to reject `gh-pages` directory (since it will have `PackageInfo.g` which will provide a name of a meta-package, but not the one from `PackageComponents`).

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+      return [];
+    fi;
+
+    file:= Filename( [ pkgdir ],
+                      Concatenation( name, "/PackageInfo.g" ) );
+    if file = fail then
+      files := [];
+      # Perhaps some subdirectories contain `PackageInfo.g' files.
+      for subdir in Set( DirectoryContents( pkgpath ) ) do
+        if not subdir in [ ".", ".." ] then
```
fingolfin: Optional: I'd prefer `if subdir in [ ".", ".." ] then continue; fi;` (don't mind if it is on one line or three, the important bit is: I don't have to parse quite as deeply nested code in my brain)

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+          pkgpath:= Filename( [ pkgdir ],
+                              Concatenation( name, "/", subdir ) );
+          if pkgpath <> fail and IsDirectoryPath( pkgpath )
+                              and not subdir in [ ".", ".." ] then
```
fingolfin: You already tested subdir above, so this should not be necessary here, no?

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+            file:= Filename( [ pkgdir ],
+                Concatenation( name, "/", subdir, "/PackageInfo.g" ) );
```
fingolfin: Hmm... in light of this, why is it even necessary to create and check `pkgpath`? If `pkgpath` does not refer to a dir, then surely `file` here also will be set to `fail`, no?

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+
+BindGlobal( "EnrichMetapackageRecord", function( subpackagerecord, metapackagerecord )
+    local version, name;
+    if not IsBound( metapackagerecord.Dependencies.NeededOtherPackages ) then
+        metapackagerecord.Dependencies.NeededOtherPackages := [];
+    fi;
+
+    name := subpackagerecord.PackageName;
+    version := subpackagerecord.Version;
+    version := Concatenation( "= ", version );
```
fingolfin: A comment here would be nice, say:
> Make the metapackage depend on its subpackage, in the precise version contained in the meta package. This is important to allow different versions of a meta package to be installed and used in parallel."

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+end );
+
+#############################################################################
+##
+#F  AddPackageInfo( files )
+##
+InstallGlobalFunction( AddPackageInfos, function( files, pkgdir, ignore, metapackagerecord... )
+    local file, record, pkgname, version, is_metapackage, name;
+    if Length( metapackagerecord ) = 1 then
+      is_metapackage := true;
```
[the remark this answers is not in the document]
fingolfin: Also, in terms of the discussion of whether "metapackage" is a good name, perhaps it'd be better to call this `hasSubpackages`?

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+        record:= GAPInfo.PackageInfoCurrent;
+        Unbind( GAPInfo.PackageInfoCurrent );
+        pkgname:= LowercaseString( record.PackageName );
+        NormalizeWhitespace( pkgname );
+        version:= record.Version;
+
+        # If we have this version already then leave it out.
+        if ForAll( GAPInfo.PackagesInfo,
+                    r ->    r.PackageName <> record.PackageName
+                        or r.Version <> version ) then
```
fingolfin: Optional: I know this is old code, but: once again, I'd prefer if it looked like this (and IMHO, that'd also match the comment before it better):
```gap
if ForAny( GAPInfo.PackagesInfo, r -> [r.PackageName, r.Verson] = [record.PackageName, verson] ) then
  continue;
fi;

File: lib/package.gi [hunk 7]
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+                " (user preference PackagesToIgnore)" ), "GAP" );
+          else
+            record.InstallationPath:= Filename( [ pkgdir ], file[2] );
+            if not IsBound( record.PackageDoc ) then
+              record.PackageDoc:= [];
+            elif IsRecord( record.PackageDoc ) then
+              record.PackageDoc:= [ record.PackageDoc ];
+            fi;
+
+            if IsBound( record.IsMetaPackage ) and record.IsMetaPackage = true then
```
fingolfin: Didn't we want to change this to `HasSubpackages`?

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+              for name in Set( DirectoryContents( record.InstallationPath ) ) do
+                files := FindPackageInfosInSubdirectories( Directory( record.InstallationPath ), name );
```
fingolfin: Optional: extract `dir := Directory( record.InstallationPath );` here and in the next line?

lib/package.gi
```diff
@@ -2096,7 +2185,7 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     else
       Error( "<info> must be either a record or a filename" );
     fi;
-
+
```
fingolfin: Gratuitous whitespace change

lib/package.gi
```diff
@@ -2127,6 +2216,22 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
     return true;
     end;
+    
+    TestOption( record, "IsMetaPackage", IsBool, "a bool" );
+
+    if IsBound( record.IsMetaPackage ) then
+      is_metapackage := record.IsMetaPackage;
+    else
+      is_metapackage := false;
+    fi;
```
fingolfin: optional: `is_metapackage := IsBound( record.IsMetaPackage )  and IsBound( record.IsMetaPackage );` (resp. `hasSubpackages := ...`

lib/package.gi
```diff
@@ -2127,6 +2216,22 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+
+    TestOption( record, "SubpackageOf", IsString, "a string" );
+
+    if IsBound( record.SubpackageOf ) then
+      is_subpackage := true;
+    else
+      is_subpackage := false;
+    fi;
```
fingolfin: optonal: `isSubpackage := IsBound( record.SubpackageOf )`

lib/package.gi
```diff
@@ -2068,7 +2157,7 @@ InstallGlobalFunction( DeclareAutoreadableVariables,
 InstallGlobalFunction( ValidatePackageInfo, function( info )
     local record, pkgdir, i, IsStringList, IsRecordList, IsProperBool, IsURL,
           IsFilename, IsFilenameList, result, TestOption, TestMandat, subrec,
-          list;
+          list, is_metapackage, is_subpackage;
```
fingolfin: Please use camelCase.

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+                Concatenation( name, "/", subdir, "/PackageInfo.g" ) );
+            if file <> fail then
+              Add( files,
+                    [ file, Concatenation( name, "/", subdir ) ] );
+            fi;
+          fi;
+        fi;
+      od;
+    else
+      files:= [ [ file, name ] ];
```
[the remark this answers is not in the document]
sebasguts: All right. This logic was there before, and I didn't change it. Will nevertheless change it now.

lib/package.gi
sebasguts: No, `subdir` is just one element of the `DirectoryContents`, not really a subdir.

lib/package.gi
```diff
@@ -159,6 +159,159 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+##  directory pkgdir, return a list of their paths.
+##
+BindGlobal( "FindPackageInfosInSubdirectories", function( pkgdir, name )
+    local pkgpath, file, files, subdir;
+    
+    if name in [ ".", ".." ] then
+      return [];
+    fi;
+
+    pkgpath:= Filename( [ pkgdir ], name );
```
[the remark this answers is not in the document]
sebasguts: `Filename(pkgdir, name)` does not check for the existence of a file, but just returns the string. `Filename( [ pkgdir ], name );` does check, so this transition is not possible in all places.

lib/package.gd
```diff
@@ -159,6 +159,16 @@ DeclareGlobalFunction( "RECORDS_FILE" );
...
 
+#############################################################################
+##
+#F  AddPackageInfo( files, pkgdir, ignore, metapackages... )
+##
+##  Adds package info records from a file list to the global list
+##  of info records.
+##
+DeclareGlobalFunction( "AddPackageInfos" );
+
```
alex-konovalov: The comment says `AddPackageInfo` without `s`. Also, it's not clear to me what is a "file list". It it a list of files? Or a file containing a list? Or something else? What's the use case for this function?

lib/package.gi
```diff
@@ -159,6 +159,166 @@ InstallGlobalFunction( SetPackageInfo, function( record )
     end );
 
 
+#############################################################################
+##
+#F  FindPackageInfosInSubdirectories( pkgdir, name )
+##
+##  Finds all PackageInfos in subdirectories of directory name in
+##  directory pkgdir, return a list of their paths.
```
alex-konovalov: Nitpick: find & return or finds & returns

lib/package.gi
```diff
@@ -159,6 +159,166 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+      fi;
+    od;
+
+    return files;
+end );
+
+#############################################################################
+##
+#F  EnrichMetapackageRecord( subpackagerecord, metapackagerecord )
+##
```
alex-konovalov: A comment what it does?

lib/package.gi
```diff
@@ -159,6 +159,166 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+    version := Concatenation( "= ", version );
+
+    Add( metapackagerecord.Dependencies.NeededOtherPackages, [ name, version ] );
+
+end );
+
+#############################################################################
+##
+#F  EnrichSubpackageRecord( subpackagerecord, metapackagerecord )
+##
```
alex-konovalov: Also a comment?

lib/package.gi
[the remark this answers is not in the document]
fingolfin: Indeed, my mistake

lib/package.gi
fingolfin: I don't understand... Sure, it can be anything, e.g. a regular file. And if that happens, `.../subir/PackageInfo.g` of course is not a valid file, so `Filename` will return `fail`. What am I missing?

sebasguts: Ah, I understand. No, we can remove this declaration.

fingolfin:
@sebasguts I think the biggest change I am missing is the addition of `tst/mockmetapkg`

alex-konovalov:
I'm still observing this problem:
```
* Metapackage - new version 0.1 discovered!!!
  ============================================
  Getting new archives from 
  https://github.com/sebasguts/MetaPackage/releases/download/MetaPackage-0.1/M\
etaPackage-0.1[ ".tar.gz" ]
  unpacking MetaPackage-0.1.tar.gz
#E  component `PackageDoc' must be bound to a record or a list of records
  ERROR (Metapackage): validation of the info file not successful! SKIPPING!!!
```

lib/package.gi
```diff
@@ -2212,7 +2325,11 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     TestOption( record, "IssueTrackerURL", IsURL, "a string started with http://, https:// or ftp://" );
     TestOption( record, "SupportEmail", IsString, "a string" );
     TestMandat( record, "AbstractHTML", IsString, "a string" );
-    TestMandat( record, "PackageWWWHome", IsURL, "a string started with http://, https:// or ftp://" );
+    if isSubpackage then
+      TestOption( record, "PackageWWWHome", IsURL, "a string started with http://, https:// or ftp://" );
+    else
+      TestMandat( record, "PackageWWWHome", IsURL, "a string started with http://, https:// or ftp://" );
+    fi;
     if TestMandat( record, "PackageDoc",
```
alex-konovalov: The error 
```
#E  component `PackageDoc' must be bound to a record or a list of records
  ERROR (Metapackage): validation of the info file not successful! SKIPPING!!!
```
is produced by `ValidatePackageInfo` here.

lib/package.gd
```diff
@@ -159,6 +159,17 @@ DeclareGlobalFunction( "RECORDS_FILE" );
 DeclareGlobalFunction( "SetPackageInfo" );
 
 
+#############################################################################
+##
+#F  AddPackageInfo( files, pkgdir, ignore, metapackages... )
```
fingolfin:
```suggestion
#F  AddPackageInfos( files, pkgdir, ignore, metapackages... )
```

lib/package.gi
```diff
@@ -159,6 +159,172 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+          record.InstallationPath:= Filename( [ pkgdir ], file[2] );
+          if not IsBound( record.PackageDoc ) then
+            record.PackageDoc:= [];
+          elif IsRecord( record.PackageDoc ) then
+            record.PackageDoc:= [ record.PackageDoc ];
+          fi;
+
+          if IsBound( record.HasSubpackage ) and record.HasSubpackage = true then
+            for name in Set( DirectoryContents( record.InstallationPath ) ) do
+              dir := Directory( record.InstallationPath );
```
fingolfin: Move this line out of the for loop

lib/package.gi
```diff
@@ -159,6 +159,172 @@ InstallGlobalFunction( SetPackageInfo, function( record )
     end );
 
 
+#############################################################################
+##
+#F  FindPackageInfosInSubdirectories( pkgdir, name )
+##
+##  Finds all PackageInfos in subdirectories of directory name in
+##  directory pkgdir, returns a list of their paths.
```
fingolfin: I find this description difficult to understand. How about this:
> This function tests if <pkgdir> has a subdirectory named <name>`; if so, it iterates over all subdirectories
> of <name>, and returns a list of those which contain a PackageInfo.g files.

But thinking some more about it, I think it would be even better to replace this function by one that only takes one arguments, `path`, and which returns all subdirs of path that contain a PackageInfo.g. That is much easier to understand and explain.

Then, add another function which returns all subdirs of a path, if any; this function has more users, after all, even in this PR (in fact, even in this very function :-) ). It could look like this:
```gap
SubdirectoriesOfPath := function(path)
  local name, file, subs;
  subs := [];
  path := Directory(path);
  for name in Set(DirectoryContents(path)) do
    if name in [ ".", ".." ] then continue; fi;
    file := Filename(path, name);
    if IsDirectoryPath(file) then Add(subs, file); fi;
  od;
  return subs;
end;
```

lib/package.gi
```diff
@@ -159,6 +159,172 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+      if subdir in [ ".", ".." ] then
+        continue;
+      fi;
+      file:= Filename( [ pkgdir ],
+          Concatenation( name, "/", subdir, "/PackageInfo.g" ) );
+      if file <> fail then
+        Add( files,
+              [ file, Concatenation( name, "/", subdir ) ] );
+      fi;
+    od;
```
fingolfin: This loop could now also be simplified, e.g. to something like this:
```gap
  files := Filtered(SubdirectoriesOfPath(pkgpath), dir -> fail <> Filename([Directory(dir)], "PackageInfo.g"));
```

lib/package.gi
```diff
@@ -159,6 +159,172 @@ InstallGlobalFunction( SetPackageInfo, function( record )
...
+
+    return files;
+end );
+
+#############################################################################
+##
+#F  EnrichMetapackageRecord( subpackagerecord, metapackagerecord )
+##
+##  Enriches the PackageInfo of the metapackage with information from the
+##  Subpackage:
```
fingolfin:
```suggestion
##  subpackage:
```

alex-konovalov:
Picked up!
```
* Metapackage - new version 0.1 discovered!!!
  ============================================
  Getting new archives from 
  https://github.com/sebasguts/MetaPackage/releases/download/MetaPackage-0.1/M\
etaPackage-0.1[ ".tar.gz" ]
  unpacking MetaPackage-0.1.tar.gz
  Validation of the info file successful!
WARNING (Metapackage): has status dev
which is is not one of accepted/submitted/deposited
  Package Metapackage 0.1 from 21/09/2018 has dependencies:
  * NeededOtherPackages [ [ "GAPDoc", ">= 1.5" ] ]
  * GAP >= 4.6
  finding text files  . . .

=====================text files==========================
MetaPackage/PackageInfo.g
MetaPackage/README.md
MetaPackage/makedoc.g
MetaPackage/ComponentTwo/PackageInfo.g
MetaPackage/ComponentTwo/README
MetaPackage/ComponentTwo/makedoc.g
MetaPackage/ComponentTwo/init.g
MetaPackage/ComponentTwo/read.g
MetaPackage/ComponentTwo/gap/ComponentTwo.gd
MetaPackage/ComponentTwo/gap/ComponentTwo.gi
MetaPackage/ComponentTwo/doc/chap0_mj.html
MetaPackage/ComponentTwo/doc/_AutoDocMainFile.xml
MetaPackage/ComponentTwo/doc/chap1_mj.html
MetaPackage/ComponentTwo/doc/title.xml
MetaPackage/ComponentTwo/doc/chapInd.html
MetaPackage/ComponentTwo/doc/chapInd.txt
MetaPackage/ComponentTwo/doc/nocolorprompt.css
MetaPackage/ComponentTwo/doc/lefttoc.css
MetaPackage/ComponentTwo/doc/ComponentTwo.tex
MetaPackage/ComponentTwo/doc/times.css
MetaPackage/ComponentTwo/doc/toggless.css
MetaPackage/ComponentTwo/doc/ComponentTwo.xml
MetaPackage/ComponentTwo/doc/_Chapter_ComponentTwo_automatic_generated_documentation.xml
MetaPackage/ComponentTwo/doc/chap0.html
MetaPackage/ComponentTwo/doc/chap1.txt
MetaPackage/ComponentTwo/doc/chap0.txt
MetaPackage/ComponentTwo/doc/manual.css
MetaPackage/ComponentTwo/doc/ragged.css
MetaPackage/ComponentTwo/doc/chap1.html
MetaPackage/ComponentTwo/doc/chapInd_mj.html
MetaPackage/ComponentTwo/doc/chooser.html
MetaPackage/ComponentTwo/tst/testall.g
MetaPackage/ComponentOne/PackageInfo.g
MetaPackage/ComponentOne/README
MetaPackage/ComponentOne/makedoc.g
MetaPackage/ComponentOne/init.g
MetaPackage/ComponentOne/read.g
MetaPackage/ComponentOne/gap/ComponentOne.gd
MetaPackage/ComponentOne/gap/ComponentOne.gi
MetaPackage/ComponentOne/doc/ComponentOne.xml
MetaPackage/ComponentOne/doc/chap0_mj.html
MetaPackage/ComponentOne/doc/_AutoDocMainFile.xml
MetaPackage/ComponentOne/doc/chap1_mj.html
MetaPackage/ComponentOne/doc/title.xml
MetaPackage/ComponentOne/doc/chapInd.html
MetaPackage/ComponentOne/doc/chapInd.txt
MetaPackage/ComponentOne/doc/nocolorprompt.css
MetaPackage/ComponentOne/doc/lefttoc.css
MetaPackage/ComponentOne/doc/ComponentOne.tex
MetaPackage/ComponentOne/doc/_Chapter_ComponentOne_automatic_generated_documentation.xml
MetaPackage/ComponentOne/doc/times.css
MetaPackage/ComponentOne/doc/toggless.css
MetaPackage/ComponentOne/doc/chap0.html
MetaPackage/ComponentOne/doc/chap1.txt
MetaPackage/ComponentOne/doc/chap0.txt
MetaPackage/ComponentOne/doc/manual.css
MetaPackage/ComponentOne/doc/ragged.css
MetaPackage/ComponentOne/doc/chap1.html
MetaPackage/ComponentOne/doc/chapInd_mj.html
MetaPackage/ComponentOne/doc/chooser.html
MetaPackage/ComponentOne/tst/testall.g
MetaPackage/tst/testall.g

=====================end of the list of text files=======
=====================binary files========================
MetaPackage/ComponentTwo/doc/ComponentTwo.blg
MetaPackage/ComponentTwo/doc/manual.six
MetaPackage/ComponentTwo/doc/ComponentTwo.log
MetaPackage/ComponentTwo/doc/ComponentTwo.aux
MetaPackage/ComponentTwo/doc/manual.js
MetaPackage/ComponentTwo/doc/manual.lab
MetaPackage/ComponentTwo/doc/ComponentTwo.out
MetaPackage/ComponentTwo/doc/ComponentTwo.ind
MetaPackage/ComponentTwo/doc/rainbow.js
MetaPackage/ComponentTwo/doc/ComponentTwo.idx
MetaPackage/ComponentTwo/doc/ComponentTwo.toc
MetaPackage/ComponentTwo/doc/manual.pdf
MetaPackage/ComponentTwo/doc/ComponentTwo.ilg
MetaPackage/ComponentTwo/doc/ComponentTwo.pnr
MetaPackage/ComponentTwo/doc/ComponentTwo.bbl
MetaPackage/ComponentTwo/doc/toggless.js
MetaPackage/ComponentOne/doc/ComponentOne.pnr
MetaPackage/ComponentOne/doc/manual.six
MetaPackage/ComponentOne/doc/ComponentOne.ilg
MetaPackage/ComponentOne/doc/manual.js
MetaPackage/ComponentOne/doc/ComponentOne.bbl
MetaPackage/ComponentOne/doc/ComponentOne.out
MetaPackage/ComponentOne/doc/ComponentOne.idx
MetaPackage/ComponentOne/doc/ComponentOne.toc
MetaPackage/ComponentOne/doc/ComponentOne.ind
MetaPackage/ComponentOne/doc/manual.lab
MetaPackage/ComponentOne/doc/rainbow.js
MetaPackage/ComponentOne/doc/manual.pdf
MetaPackage/ComponentOne/doc/ComponentOne.aux
MetaPackage/ComponentOne/doc/ComponentOne.log
MetaPackage/ComponentOne/doc/ComponentOne.blg
MetaPackage/ComponentOne/doc/toggless.js

=====================end of the list of binary files=====
=====================ignored files=======================
MetaPackage/.gitignore

=====================end of the list of ignored files====
#   Error (Metapackage): could not get README file from
   https://sebasguts.github.io/MetaPackage/README
adding PackageInfo.g
adding listbinaryfiles.txt
adding listtextfiles.txt
adding pkg/MetaPackage/ComponentOne/PackageInfo.g
adding pkg/MetaPackage/ComponentOne/README
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.aux
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.bbl
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.blg
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.idx
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.ilg
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.ind
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.log
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.out
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.pnr
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.tex
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.toc
adding pkg/MetaPackage/ComponentOne/doc/ComponentOne.xml
adding pkg/MetaPackage/ComponentOne/doc/_AutoDocMainFile.xml
adding pkg/MetaPackage/ComponentOne/doc/_Chapter_ComponentOne_automatic_generated_documentation.xml
adding pkg/MetaPackage/ComponentOne/doc/chap0.html
adding pkg/MetaPackage/ComponentOne/doc/chap0.txt
adding pkg/MetaPackage/ComponentOne/doc/chap0_mj.html
adding pkg/MetaPackage/ComponentOne/doc/chap1.html
adding pkg/MetaPackage/ComponentOne/doc/chap1.txt
adding pkg/MetaPackage/ComponentOne/doc/chap1_mj.html
adding pkg/MetaPackage/ComponentOne/doc/chapInd.html
adding pkg/MetaPackage/ComponentOne/doc/chapInd.txt
adding pkg/MetaPackage/ComponentOne/doc/chapInd_mj.html
adding pkg/MetaPackage/ComponentOne/doc/chooser.html
adding pkg/MetaPackage/ComponentOne/doc/lefttoc.css
adding pkg/MetaPackage/ComponentOne/doc/manual.css
adding pkg/MetaPackage/ComponentOne/doc/manual.js
adding pkg/MetaPackage/ComponentOne/doc/manual.lab
adding pkg/MetaPackage/ComponentOne/doc/manual.pdf
adding pkg/MetaPackage/ComponentOne/doc/manual.six
adding pkg/MetaPackage/ComponentOne/doc/nocolorprompt.css
adding pkg/MetaPackage/ComponentOne/doc/ragged.css
adding pkg/MetaPackage/ComponentOne/doc/rainbow.js
adding pkg/MetaPackage/ComponentOne/doc/times.css
adding pkg/MetaPackage/ComponentOne/doc/title.xml
adding pkg/MetaPackage/ComponentOne/doc/toggless.css
adding pkg/MetaPackage/ComponentOne/doc/toggless.js
adding pkg/MetaPackage/ComponentOne/gap/ComponentOne.gd
adding pkg/MetaPackage/ComponentOne/gap/ComponentOne.gi
adding pkg/MetaPackage/ComponentOne/init.g
adding pkg/MetaPackage/ComponentOne/makedoc.g
adding pkg/MetaPackage/ComponentOne/read.g
adding pkg/MetaPackage/ComponentOne/tst/testall.g
adding pkg/MetaPackage/ComponentTwo/PackageInfo.g
adding pkg/MetaPackage/ComponentTwo/README
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.aux
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.bbl
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.blg
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.idx
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.ilg
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.ind
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.log
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.out
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.pnr
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.tex
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.toc
adding pkg/MetaPackage/ComponentTwo/doc/ComponentTwo.xml
adding pkg/MetaPackage/ComponentTwo/doc/_AutoDocMainFile.xml
adding pkg/MetaPackage/ComponentTwo/doc/_Chapter_ComponentTwo_automatic_generated_documentation.xml
adding pkg/MetaPackage/ComponentTwo/doc/chap0.html
adding pkg/MetaPackage/ComponentTwo/doc/chap0.txt
adding pkg/MetaPackage/ComponentTwo/doc/chap0_mj.html
adding pkg/MetaPackage/ComponentTwo/doc/chap1.html
adding pkg/MetaPackage/ComponentTwo/doc/chap1.txt
adding pkg/MetaPackage/ComponentTwo/doc/chap1_mj.html
adding pkg/MetaPackage/ComponentTwo/doc/chapInd.html
adding pkg/MetaPackage/ComponentTwo/doc/chapInd.txt
adding pkg/MetaPackage/ComponentTwo/doc/chapInd_mj.html
adding pkg/MetaPackage/ComponentTwo/doc/chooser.html
adding pkg/MetaPackage/ComponentTwo/doc/lefttoc.css
adding pkg/MetaPackage/ComponentTwo/doc/manual.css
adding pkg/MetaPackage/ComponentTwo/doc/manual.js
adding pkg/MetaPackage/ComponentTwo/doc/manual.lab
adding pkg/MetaPackage/ComponentTwo/doc/manual.pdf
adding pkg/MetaPackage/ComponentTwo/doc/manual.six
adding pkg/MetaPackage/ComponentTwo/doc/nocolorprompt.css
adding pkg/MetaPackage/ComponentTwo/doc/ragged.css
adding pkg/MetaPackage/ComponentTwo/doc/rainbow.js
adding pkg/MetaPackage/ComponentTwo/doc/times.css
adding pkg/MetaPackage/ComponentTwo/doc/title.xml
adding pkg/MetaPackage/ComponentTwo/doc/toggless.css
adding pkg/MetaPackage/ComponentTwo/doc/toggless.js
adding pkg/MetaPackage/ComponentTwo/gap/ComponentTwo.gd
adding pkg/MetaPackage/ComponentTwo/gap/ComponentTwo.gi
adding pkg/MetaPackage/ComponentTwo/init.g
adding pkg/MetaPackage/ComponentTwo/makedoc.g
adding pkg/MetaPackage/ComponentTwo/read.g
adding pkg/MetaPackage/ComponentTwo/tst/testall.g
adding pkg/MetaPackage/PackageInfo.g
adding pkg/MetaPackage/README.md
adding pkg/MetaPackage/makedoc.g
adding pkg/MetaPackage/tst/testall.g
  ============================================
```

sebasguts:
@alex-konovalov nice, will try to incorporate the comments on the code now.

lib/package.gi
sebasguts: The new loop is a bit different now.

sebasguts:
I think I adressed all comments

lib/package.gi
```diff
@@ -226,35 +289,50 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
...
+
+          if IsBound( record.HasSubpackage ) and record.HasSubpackage = true then
+            dir := record.InstallationPath;
+            for name in SubdirectoriesOfPath( dir ) do
+              files := FindPackageInfosInSubdirectories( name );
+              AddPackageInfos( files, Directory( dir ), ignore, record );
+            od;
           fi;
+
+          if isSubPackage then
```
fingolfin:
```suggestion
          if isSubpackage then
```
For consistency with `record.HasSubpackage` but also with Subgroup, Subspace, Subring, etc. elsewhere in GAP.

lib/package.gi
```diff
@@ -2282,29 +2375,38 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
-    if TestMandat( record, "PackageDoc",
-           x -> IsRecord( x ) or IsRecordList( x ),
-           "a record or a list of records" ) then
+    if isSubpackage then
+      TestOption( record, "PackageWWWHome", IsURL, "a string started with http://, https:// or ftp://" );
+    else
+      TestMandat( record, "PackageWWWHome", IsURL, "a string started with http://, https:// or ftp://" );
+    fi;
+    if hasSubpackage then
+      docTester := function( args... ) return IsBound( args[1].(args[2]) ) and CallFuncList( TestOption, args ); end;
```
fingolfin:
```suggestion
      # for metapackages, documentation is optional, hence so is PackageDoc;
      # but if it exists, we should of course still validate it
      docTester := {record, name, type, typename} -> IsBound( record.(name) ) and TestOption(record, name, type, typename);
```

lib/package.gi
```diff
@@ -2282,29 +2375,38 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+          x -> IsRecord( x ) or IsRecordList( x ),
+          "a record or a list of records" ) then
       if IsRecord( record.PackageDoc ) then
         list:= [ record.PackageDoc ];
       else
         list:= record.PackageDoc;
       fi;
       for subrec in list do
-        TestMandat( subrec, "BookName", IsString, "a string" );
+        docTester( subrec, "BookName", IsString, "a string" );
```
fingolfin: Why is `docTest` used here and elsewhere? If `record.PackageDoc` is present, then it should pass its full validation check, shouldn't it?

fingolfin:
The other thing missing in this PR of course is some basic documentation in the manual chapter on packages about metapackages and subpackages. @sebasguts and me wanted to get started on that in Halle, but that didn't pan out (just like me finishing the MatrixObj specification did not :/).

@sebasguts perhaps that's something we can work on in Siegen in the coming weeks?

alex-konovalov:
@sebasguts changes requested by @fingolfin - would be nice to address them quickly to allow this in GAP 4.11 - otherwise, we have to delay it until 4.12...

fingolfin:
@alex-konovalov So @sebasguts and me would like to get this resolved soon; besides my review comments, and the need for documentation, I think the only thing left is updating the website and/or distribution tools to handle meta packages. From what you wrote, it sounded as if that mostly is the case already anyway -- but I am not quite sure whether that's true? Also, it would probably be good to at least show links between meta packages and their subpackages; better would be to group packages together under their meta package... 

The problem for us is that we don't know how to go about implementing this, simply because we have no means to test any changes we make there. I guess we need to figure out how to setup a test environment where we can replicate the process of updating the package webpages. What's the best place to start with that -- should we start with reading https://github.com/gap-system/gap-distribution/blob/master/DistributionUpdate/README.md and then moving on from there? Any other pointers / warnings / suggestions you may have for us?

lib/package.gi
sebasguts: Well, because for subpackages, documentation is optional (see comment some lines above). So we use `docTester`, which is either set to `TestMandat` (in case it is not a metapackage) or so something just tests the entries if they are there.

sebasguts:
@fingolfin I adressed your comments, except those regarding `NameDirectory`.

fingolfin:
The current form of this PR breaks loading of the package:
```
Syntax warning: Unbound global variable in /home/travis/build/gap-system/gap/l\
ib/package.gi:328
          if isSubpackage then
             ^^^^^^^^^^^^
and packages ...
Error, Variable: 'isSubpackage' must have an assigned value in
  if isSubpackage then
    EnrichMetapackageRecord( record, metapackagerecord );
    EnrichSubpackageRecord( record, metapackagerecord );
fi; at /home/travis/build/gap-system/gap/lib/package.gi:331 called from
```

lib/package.gi
```diff
@@ -226,35 +294,50 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
...
+
+          if IsBound( record.HasSubpackage ) and record.HasSubpackage = true then
+            dir := record.InstallationPath;
+            for name in SubdirectoriesOfPath( dir ) do
+              files := FindPackageInfosInSubdirectories( name );
+              AddPackageInfos( files, Directory( dir ), ignore, record );
+            od;
           fi;
+
+          if isSubpackage then
```
fingolfin:
```suggestion
          if isSubPackage then
```

But actually, I think you should change it the other way around, and use `isSubpackage` everywhere instead of `isSubPackage`.

fingolfin: I pushed a commit to resolve this.

sebasguts:
@fingolfin I think your latest comments were now addressed by yourself?

lib/package.gi
fingolfin: That explains the first use of `docTester` in line 2392. But I'd expect a subpackage without manual to have no `PackageDoc` record at all. Conversely, if it has one, shouldn't it be complete?

fingolfin: I've now updated the PR to address this, too, as well as a few other minor things. I also squashed most of it in one commit; only the most recent changes I just made are stilling a separate commit, so that you can more easily review them. Assuming you are fine with them, we can squash them into the main commit.

fingolfin:
@sebasguts From my POV, the code is in principle ready for merging, and the main things still missing are:

1. documentation, at least some minimal
2. tests: we have `tst/mockpkg/`, perhaps we could add `tst/mockmetapkg/`?
4. feedback from @alex-konovalov as to what he expects us to do for integration with the website / package distribution

fingolfin:
Unfortunately, this PR is abandoned, and I don't think anyone has plans to revive it. I also don't think it is really necessary anymore; the new package distro "knows" that the various Homalg (resp. CAP) packages belong together, and tests them together whenever there are updates. This worked quite well in the past year.

PR #2268, opened 2018-03-19
Handle spaces in filenames of gziped filenames

### LINKED ISSUE #2264: gap -L fails if compressed file has a space in its filename
GAP is able to load workspaces even if the filename has a space in it.  But if the workspace is compressed, loading the workspace fails.

Since GAP Workspaces are huge, but compress well, this would be worthwhile to fix.

If you start with a compressed workspace file named Workspace.gz in the folder 'Test Session.gapsession/", then the following should work, but does not.

### Example
_The following should load GAP, but does not_
$ gap -L 'Test Session.gapsession/Workspace'
sh: Test: No such file or directory
Unexpected End of File in Load

_The following works fine_
$ gunzip Test\ Session.gapsession/Workspace.gz 
$ gap -L 'Test Session.gapsession/Workspace'

_Another workaround, using advanced bash features, is as follows_
$ gap -L <(gunzip < 'Test Session.gapsession/Workspace.gz')

### Suggestions for fix
It appears to me that the problem is in sysfiles.c, in the routine SyFopen.  (You can easily find the exact lines by searching for gunzip or for namegz.)  Where the routine calls gunzip, it should probably escape the filename with apostrophes or similar.
I'm not confident enough with the GAP code to suggest the right fix.

 ┌───────┐   GAP 4.8.10, 15-Jan-2018, build of 2018-02-19 15:55:06 (CET)
 └───────┘   Architecture: x86_64-apple-darwin16.7.0-gcc-default64

ChrisJefferson:
Wrap filenames we pass to gzip/gunzip via shell in single quotes,
so we handle spaces and some other characters which need escaping.

This does not handle all possible filenames which could cause a
problem, but hopefully covers most which will arise in practice.

Resolves #2264

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -240,10 +240,12 @@ static void fopenMaybeCompressed(char* name, struct ProfileState* ps)
 {
 #ifdef HAVE_POPEN
   char popen_buf[4096];
-  if(endsWithgz(name) && strlen(name) < 3000)
+  // Need space for "gzip < '", ".gz'" and terminating \0.
+  if(endsWithgz(name) && strlen(name) < sizeof(popen_buf) - 8 - 4 - 1)
   {
-    strcpy(popen_buf, "gzip > ");
-    strcat(popen_buf, name);
+    strxcpy(popen_buf, "gzip > '", sizeof(popen_buf));
+    strxcat(popen_buf, name, sizeof(popen_buf));
+    strxcat(popen_buf, "'", sizeof(popen_buf));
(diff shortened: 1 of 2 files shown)


fingolfin:
I took the liberty of adding "Resolves #2264" to your PR description, so that merging this PR will close the associated issue.

RussWoodroofe:
Would this be a good candidate to backport for gap 4.9.2?  It seems unlikely to do much harm.

PR #2269, opened 2018-03-19
set *errin* isTTY properly in package mode (-p)

RussWoodroofe:
When running in package mode (-p), as used by xgap, GAP 
sets *errin* to be the same as *stdin*.  It should at the same 
time also set the isTTY property of *errin* to be the same as 
that for *stdin*.

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -1935,9 +1935,9 @@ void InitSystem (
     if ( SyWindow ) {
       /*         SyLineEdit   = 1;
                  SyCTRD       = 1; */
-        syBuf[2].fp = fileno(stdin);  syBuf[2].echo = fileno(stdout);
+        syBuf[2].fp = syBuf[0].fp;  syBuf[2].echo = syBuf[0].echo;
         syBuf[2].isTTY = syBuf[0].isTTY;
-        syBuf[3].fp = fileno(stdout);
+        syBuf[3].fp = syBuf[1].fp;
         syWinPut( 0, "@p", "1." );
     }
    
--- a/src/system.c
+++ b/src/system.c
@@ -1936,6 +1936,7 @@ void InitSystem (
       /*         SyLineEdit   = 1;
                  SyCTRD       = 1; */
         syBuf[2].fp = fileno(stdin);  syBuf[2].echo = fileno(stdout);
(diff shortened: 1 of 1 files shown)


src/system.c
```diff
@@ -1936,6 +1936,7 @@ void InitSystem (
       /*         SyLineEdit   = 1;
                  SyCTRD       = 1; */
         syBuf[2].fp = fileno(stdin);  syBuf[2].echo = fileno(stdout);
+        syBuf[2].isTTY = syBuf[0].isTTY;
```
fingolfin: This code is not "obviously" right. If we really "know" that `syBuf[0]` refers to stdin/stdout, i'd rather copy all three values from it.

Or else, use `isatty` directly, i.e. `syBuf[2].isTTY = isatty(syBuf[2].fp);`.

RussWoodroofe: I agree with your comment.  To expand a little, the line added and preceding line should read either as:
         syBuf[2].fp = syBuf[0].fp;  syBuf[2].echo = syBuf[0].echo;
         syBuf[2].isTTY = syBuf[0].isTTY;
or else as:
         syBuf[2].fp = fileno(stdin);  syBuf[2].echo = fileno(stdout);
         syBuf[2].isTTY = (isatty(syBuf[2].fp) && ttyname(syBuf[2].fp) != NULL);
The former one seems a little more correct to me, but the latter is a bit more conservative.  In the former case, the next line in the code (setting syBuf[3].fp) should probably also be changed to a similar style.   I'm not sure if it should also set syBuf[3].echo or not.  The code at the top of the init routine is a little inconsistent in that respect.

Note that the ttyname check in the latter solution mirrors the check for isTTY earlier in the code.

I'll try and edit the proposed patch for the less conservative solution.

fingolfin: Either of the two solutions would be fine by me (assuming they both "work" for you, i.e. solve your original problem).

RussWoodroofe: I actually tested both (including sending an ACK in a break loop), and both seemed to work.  Neither seems likely to cause any new problems.

I edited the pull request to match the first solution.  (Let me know if I did it incorrectly.)

PR #2274, opened 2018-03-21
kernel: ensure uniform printing of macfloats nan, inf, -inf

### LINKED ISSUE #2193: float's inf is Inf on SPARC Solaris 11
### Observed behaviour
```
########> Diff in /datapool/dima/Sage/gap/tst/testinstall/float.tst:130
# Input is:
posinf := Float(infinity);
# Expected output:
inf
# But found:
Inf
```
---and similar messages in all tests involving `inf` in `testinstall/float.tst`

### Expected behaviour
tests should pass

```
*********   GAP 4.8.8-5085-gb6d4e8e-dirty of today
 *********   Architecture: sparc-sun-solaris2.11-default64
 Configuration:  gmp 6.1.2, KernelDebug
 Packages:   AClib 1.2, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.8, CRISP 1.4.4, Cryst 4.1.13, CrystCat 1.1.6, CTblLib 1.2.2, FactInt 1.6.1, FGA 1.3.1, 
             GAPDoc 1.6.1, IRREDSOL 1.4, LAGUNA 3.8.0, Polenta 1.3.8, Polycyclic 2.11, PrimGrp 3.3.0, RadiRoot 2.7, ResClasses 4.7.1, smallgrp 1.2, 
             Sophus 1.23, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.53
```
This is git master of 19.02.2018. Same errors observed on GAP 4.9 beta of 1st of Feb,
too. Built with gcc 5.4.0.

fingolfin:
On some platforms, NaN values might be printed as `nan`, on others as `NaN`;
on some, the sign of the Nan value may be printed, on other not so (and the
sign of a NaN is something which highly depends on the float implementation to
start with).

This made it difficult to write tests which rely on the way these values are
printed. Hence we now deal with these values explicitly and in a manner which
produces identical results across all platforms.

Fixes #2193

### DIFF
--- a/src/macfloat.c
+++ b/src/macfloat.c
@@ -50,18 +50,39 @@ Obj TypeMacfloat (
 }
 
 
+// helper function for printing a "decimal" representation of a macfloat
+// into a buffer.
+static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precision)
+{
+    // handle printing of NaN and infinities ourselves, to ensure
+    // they are printed uniformly across all platforms
+    if (isnan(val)) {
+        strcpy(buf, "nan");
+    }
+    else if (isinf(val)) {
+        if (val > 0)
+            strcpy(buf, "inf");
+        else
+            strcpy(buf, "-inf");
+    }
+    else {
+        snprintf(buf, bufsize, "%.*" PRINTFFORMAT, precision, val);
+    }
+}
+
+
 /****************************************************************************
 **
-*F  PrintMacfloat( <macfloat> ) . . . . . . . . . . . . . . . . print a macfloat value
+*F  PrintMacfloat( <macfloat> ) . . . . . . . . . . .  print a macfloat value
 **
 **  'PrintMacfloat' prints the macfloating value <macfloat>.
 */
-void PrintMacfloat (
-    Obj                 x )
+void PrintMacfloat(Obj x)
 {
-  Char buf[32];
-  snprintf(buf, sizeof(buf), "%.16" PRINTFFORMAT, (TOPRINTFFORMAT) VAL_MACFLOAT(x));
-  Pr("%s",(Int)buf, 0);
+    Char buf[32];
+    // TODO: should we use PRINTFDIGITS instead of 16?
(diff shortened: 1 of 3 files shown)


dimpase:
OK, this does give `inf` on GAP built with vanilla gcc 7.2 on SPARC Solaris 11, good.

Is there a way to test `NaN`'s from GAP?

fingolfin:
You mean a way to generate a NaN, like `0.0/0.0` ? Or the `IsNaN` property to test for a nan?

dimpase:
Right, this works on SPARC Solaris
```
gap> Float(0.0/0.0);
nan
```
Would you mind adding tests for this in `tst/testinstall/float.tst` ? 
There are only tests for `inf` there.

fingolfin:
@stevelinton so far, we only know of one system which does not print `nan`, namely SPARC Solari 11. I highly doubt that many people use GAP there (or that many people in mathematics still use such systems). And of those, even fewer will use floats inside GAP. And of those, even fewer will ever see a `nan`... My guess is that the number is 1, and consists precisely of @dimpase (thanks for catching this, BTW!!!) but of course I could be wrong ;-).

So, I'd rather not spend time on documenting this. Of course on the long run, it would be nice to document how nan and inf etc. are printed in general, but that's part of the bugger "floating points in GAP need to be better documented" issue and so IMHO beyond the scope of this PR...

dimpase:
> My guess is that the number is 1, and consists precisely of @dimpase (thanks for catching this, BTW!!!) but of course I could be wrong ;-).

It's bigger than 1, e.g. @jdemeyer can also be considered counted in :-) but not that we know  more.

I'd reiterate that printing `nan` in at least one test sounds like a reasonable thing, though.

fingolfin:
I added some tests and also made `SIGN_MACFLOAT` and `SIGNBIT_MACFLOAT` consistent

jdemeyer:
> I'd reiterate that printing nan in at least one test sounds like a reasonable thing, though.

I agree. At least it checks that GAP doesn't crash when encountering a `NaN` for example.

PR #2276, opened 2018-03-21
Not all finitely generated magmas are groups

### LINKED ISSUE #2252: MakeMagmaWithInversesByFiniteGenerators shouldn't set IsFinitelyGeneratedGroup
Consider this example:
```
gap> A:=[[1,2,3,4],[2,1,4,2],[3,4,1,3],[4,2,3,1]];
[ [ 1, 2, 3, 4 ], [ 2, 1, 4, 2 ], [ 3, 4, 1, 3 ], [ 4, 2, 3, 1 ] ]
gap> M:=MagmaWithInversesByMultiplicationTable(A);
<magma-with-inverses with 4 generators>
gap> IsFinitelyGeneratedGroup(M);
true
gap> IsGroup(M);
false
```
It is quite obvious that `M` is no group from the multiplication table (the 2nd and 3rd row are not permutations). Yet the magma "thinks" that it is in the filter `IsFinitelyGeneratedGroup`, which seems like a contradiction.

The problem is in `MakeMagmaWithInversesByFiniteGenerators`, which sets this filter unconditionally. This can be traced back to this commits (from the our "secret" internal repository):
```patch
commit ac50226cb8a39c978d43243a5546822fc2a2f2af
Date:   1999-04-29 09:41:11 +0000

    Changed the installation of `MagmaByGenerators' methods to the case of a
    finite generators list. This permits to set the `IsFinitelyGenerated' filter
    and thus gets rid of immediate method calls.

diff --git a/lib/magma.gi b/lib/magma.gi
index c5f731d3a..70550e70b 100644
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -15,7 +15,6 @@ Revision.magma_gi :=
 
 ##
-
 #M  PrintObj( <M> ) . . . . . . . . . . . . . . . . . . . . . . print a magma
 ##
 InstallMethod( PrintObj,
@@ -606,14 +605,16 @@ InstallOtherMethod( MagmaWithOneByGenerators,
 ##
 #M  MagmaWithInversesByGenerators( <gens> ) . . . . . . . .  for a collection
 ##
-MakeMagmaWithInversesByGenerators:=function(family,gens)
+MakeMagmaWithInversesByFiniteGenerators:=function(family,gens)
 local M;
   if not IsBound(family!.defaultMagmaWithInversesByGeneratorsType) then
     family!.defaultMagmaWithInversesByGeneratorsType :=
       NewType( FamilyObj( gens ),
 		IsMagmaWithInverses and IsAttributeStoringRep 
-		and HasGeneratorsOfMagmaWithInverses);
+		and HasGeneratorsOfMagmaWithInverses 
+		and IsFinitelyGeneratedGroup);
   fi;
+
   M:=rec();
   ObjectifyWithAttributes( M,family!.defaultMagmaWithInversesByGeneratorsType,
     GeneratorsOfMagmaWithInverses, AsList( gens ) );
@@ -621,9 +622,9 @@ local M;
 end;
 
 InstallMethod( MagmaWithInversesByGenerators, "for collection", true,
-    [ IsCollection ] , 0,
+    [ IsCollection and IsFinite] , 0,
 function( gens )
-  return MakeMagmaWithInversesByGenerators(FamilyObj(gens),gens);
+  return MakeMagmaWithInversesByFiniteGenerators(FamilyObj(gens),gens);
 end);
 
@@ -631,13 +632,13 @@ end);
 #M  MagmaWithInversesByGenerators( <Fam>, <gens> )  . . . for family and list
 ##
 InstallOtherMethod( MagmaWithInversesByGenerators, "for family and list",
-    true, [ IsFamily, IsList ], 0,
+    true, [ IsFamily, IsList and IsFinite], 0,
 function( family, gens )
 local M;
   if not ( IsEmpty(gens) or IsIdenticalObj( FamilyObj(gens), family ) ) then
     Error( "<family> and family of <gens> do not match" );
   fi;
-  return MakeMagmaWithInversesByGenerators(family,gens);
+  return MakeMagmaWithInversesByFiniteGenerators(family,gens);
 end );
 
 
```

fingolfin:
... hence we should not set IsFinitelyGeneratedGroup for them. Instead, only set IsFinitelyGeneratedGroup if the family indicates that its elements are associative.

Note that this all is only done to avoid a call to the default immediate method for IsFinitelyGeneratedGroup. I am not really sure whether this is worth the hassle... Perhaps it is when one creates thousands and thousands of groups in a tight loop?

Fixes #2252

### DIFF
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -652,13 +652,17 @@ InstallOtherMethod( MagmaWithOneByGenerators,
 #M  MagmaWithInversesByGenerators( <gens> ) . . . . . . . .  for a collection
 ##
 MakeMagmaWithInversesByFiniteGenerators:=function(family,gens)
-local M;
+local M, filter;
   if not IsBound(family!.defaultMagmaWithInversesByGeneratorsType) then
+    filter := IsMagmaWithInverses and IsAttributeStoringRep
+                and HasGeneratorsOfMagmaWithInverses;
+    # check if the family implies associativity, and hence that the
+    # final magma is a group
+    if IS_SUBSET_FLAGS(family!.IMP_FLAGS, FLAGS_FILTER(IsAssociativeElementCollection)) then
+       filter := filter and IsFinitelyGeneratedGroup;
+    fi;
     family!.defaultMagmaWithInversesByGeneratorsType :=
-      NewType( FamilyObj( gens ),
-                IsMagmaWithInverses and IsAttributeStoringRep 
-                and HasGeneratorsOfMagmaWithInverses 
-                and IsFinitelyGeneratedGroup);
+      NewType( FamilyObj( gens ), filter );
   fi;
 
   M:=rec();
--- a/tst/testbugfix/2018-03-21-MagmaWithInversesByGenerators.tst
+++ b/tst/testbugfix/2018-03-21-MagmaWithInversesByGenerators.tst
@@ -1,0 +1,12 @@
+# GAP used to set the IsFinitelyGeneratedGroup property for every finitely
+# generated magma-with-inverses, even those which are not groups.
+gap> A:=[[1,2,3,4],[2,1,4,2],[3,4,1,3],[4,2,3,1]];
+[ [ 1, 2, 3, 4 ], [ 2, 1, 4, 2 ], [ 3, 4, 1, 3 ], [ 4, 2, 3, 1 ] ]
+gap> M:=MagmaWithInversesByMultiplicationTable(A);
+<magma-with-inverses with 4 generators>
+gap> IsFinitelyGeneratedGroup(M);
+Error, no method found! For debugging hints type ?Recovery from NoMethodFound
+Error, no 2nd choice method found for `IsFinitelyGeneratedGroup' on 1 argument\
+s
+gap> IsGroup(M);
+false


fingolfin:
Sure, that would be simpler, but it would also be redundant, as by that time, the immediate method would have been triggered anyway. So, this would be equivalent to simply not setting `IsFinitelyGeneratedGroup` in the initial filter list. I'd be happy with that, too. Not sure if @hulpke is, though?

ThomasBreuer:
Just one remark about immediate methods:
According to the documentation, one cannot rely on the fact that this feature is switched on at all.

fingolfin:
Of course that's what it says -- and I consider this claim of the manual to be rather unrealistic, because I am pretty sure nobody ever tests GAP with `IGNORE_IMMEDIATE_METHODS` set to `true`.

But anyway, I understand this remark in the manual as saying that *correctness* of code should not rely on immediate methods -- which we wouldn't be here, would we? But not running immediate methods certainly can result in changes in the method selection, and in possibly selecting worse methods. This is what would happen for all kinds of objects and properties already now, e.g. `IsFinite`, `IsTrivial` etc. would not be set once e.g. the size of a group is known. Why would we worry more about `IsFinitelyGeneratedGroup` than any of these?

fingolfin:
Anyway, I'll prepare an alternative PR introducing `IsFinitelyGeneratedMagma`.

fingolfin:
@james-d-mitchell I'd be fine with this, I still plan to prepare that alternate PR using `IsFinitelyGeneratedMagma` resp. `IsFinitelyGenerated`, I can make a third one which does as  @ThomasBreuer suggests, and of course it'd be trivial to go from that to the radical variant you propose.

In any case, it would be very good if @hulpke could chime in, as this code is by him.

Also, to underline my conjecture that nobody tests with `IGNORE_IMMEDIATE_METHODS` set to true, I run the tests, and looked at the test failures, and immediately found some which show that already code violates the idea that code should not rely on immediate methods firing . Here is one example (which at least doesn't give a mathematically wrong result, it just runs into an unnecessary error):
```
gap> IGNORE_IMMEDIATE_METHODS;
false
gap>
gap> F := FreeGroup("r", "s");; r := F.1;; s := F.2;;
gap> G := F/[s^2, s*r*s*r];;
gap> H := DerivedSubgroup(G);
Group(<fp, no generators known>)
gap> IsSolvableGroup(G);
true
gap>
gap> IGNORE_IMMEDIATE_METHODS:=true;;
gap> F := FreeGroup("r", "s");; r := F.1;; s := F.2;;
gap> G := F/[s^2, s*r*s*r];;
gap> H := DerivedSubgroup(G);
Group(<fp, no generators known>)
gap> IsSolvableGroup(G);
Error, Derived subgroup has infinite index, cannot represent
...
```

There are lots of similar examples, this was just the first.

Of course an existing bug does not justify adding another one; but a rule that is already violated in many places perhaps should be re-evaluated?

PR #2277, opened 2018-03-22
Remove support for PackageInfo.g field `PreloadFile`

fingolfin:
This undocumented field is not used by any distributed package, and its usefulness seems dubious to me at best.

Unless somebody remembers / knows a reason why this field could be useful, I'd suggest to remove it (as done by this PR), as this is one less piece of completely untested code to worry about.


fingolfin:
I guess it depends on what you consider "documented". For me, something is only documented if it is mentioned in one of the GAP manuals. Otherwise, almost any library function would be "documented", as almost all of them have comments explaining what they do.

ThomasBreuer:
Concerning the question about the documented contents of a `PackageInfo.g` file,
I ask GAP `?PackageInfo.g` and get the section `Structure of a GAP Package`
which says that the `PackageInfo.g` file of the `Example` package serves as a prototype.
As far as I know, it was once decided that the supported components should be documented this way,
and I am not aware of other documentation for `PackageInfo.g` files.

stevelinton:
I don't object considering the Example package as documentation, I think that's what it's for. It might be helpful to clarify this somewhere.  The applications Thomas mentions seem potentially useful, although I'm not quite sure how using a PreLoadFile differs from just doing something at the start of `init.g` -- if we think it might be useful maybe it should be clarified. On the other hand, in 12 years no one has actually used this feature that we know of. 

If we keep it, then, for testing, we should add a PreLoadFile to the example package.

ChrisJefferson:
I would be happy to merge this if it was rebased + fixed up

fingolfin:
Rebased, tests are running

fingolfin:
@ChrisJefferson @alex-konovalov so, shall we merge this then?

PR #2280, opened 2018-03-22
Fix incorrect IsFinitelyGeneratedMonoid method

fingolfin:
Turns out we also have `IsFinitelyGeneratedMonoid` (which I did not realize), and we also have an invalid method for that (see PR #2275).


fingolfin:
Yes, I agree on the "ideal world" remark, and I hope we can get there -- I just thought that in the meantime, it'd be a good idea to fix the bugs immediately as I notice them

fingolfin:
There was a test which required infinitely generated free monoids to know have `IsFinitelyGeneratedMonoid(M) = false`, which caused a test failure; this should now be resolved.

PR #2283, opened 2018-03-22
Add minimal test for AutomGrpSR

fingolfin:


fingolfin:
This test is quite simplistic, of course, but at least it ensures `AutomGrpSR` is called at least once by the full test suite. More extensive tests would of course be nice, but this is better than the nothing we have right now.

PR #2286, opened 2018-03-22
Add some tests for direct and wreath products of matrices

fingolfin:


fingolfin:
This uncovered a crash in `COPY_GF2VEC` if the input is in 8bit rep (or any list which is neither a plist nor in GF2 rep). This is now fixed by this PR, too. Since `COPY_GF2VEC` is newly added by HPC-GAP, I don't think we need to mention this fix in the release notes.

src/vecgf2.c
```diff
@@ -1463,21 +1463,22 @@ Obj NewGF2Vec (
           SetTypeDatObj(res, TYPE_LIST_GF2VEC_IMM);
         return res;
     }
+
+    if (!IS_LIST(list)) {
+        ErrorMayQuit("COPY_GF2VEC: argument must be a list of compressed GF2 vectors", 0L, 0L);
```
ssiccha: Should this rather be `argument must be a list of GF2 elements`?
- it should be elements not vectors
- If I read the code correctly, it can also handle GF2 elements over some arbitrarily huge field. Or are elements GFq, q = 2^r _always_ compressed?

fingolfin: You are right about the first one (I blame this on copy&paste). I don't understand what you mean with the second one -- what is a "GF2 element over some arbitrarily huge field" ? There are exactly two elements of GF2, period (albeit possibly in different reps).

fingolfin:
Fixed the error message. Also changed two assert into `Error` calls -- it doesn't cost us anything to gracefully error out, instead of segfaulting (or, if asserts are disabled, silently returning garbage data).

src/vecgf2.c
ssiccha: Ah, I was confused by the term _compressed_ wrt different representations.

PR #2296, opened 2018-03-25
src/opers.c: tweak and cleanup code; add tests

fingolfin:


src/opers.c
```diff
@@ -357,64 +362,74 @@ Obj FuncSIZE_FLAGS (
...
+        flags1 = ErrorReturnObj( "<flags1> must be a flags list (not a %s)",
+            (Int)TNAM_OBJ(flags1), 0L,
+            "you can replace <flags1> via 'return <flags1>;'" );
+    }
+    while ( TNUM_OBJ(flags2) != T_FLAGS ) {
+        flags2 = ErrorReturnObj( "<flags2> must be a flags list (not a %s)",
+            (Int)TNAM_OBJ(flags2), 0L,
+            "you can replace <flags2> via 'return <flags2>;'" );
+    }
+    if ( flags1 == flags2 ) {
```
sebasguts: Why don't you do this check in the `EqFlags` method?

fingolfin: Because this is an internal API, where the caller is expected to already have checked the input. And when called via EQ / EqFuncs, the tnums are automatically verified, too.

sebasguts: Yeah, I think I put the comment in a slightly confusing place, and of course did not specify what I really meant, sorry.

What I actually meant was:
Why is the `if ( flags1 == flags2 )` check not in the `EqFlags` method?

fingolfin: Ahhh, sorry, I overlooked that line. Good question, I guess I forgot to move it, oops.

sebasguts: No worries :). I think I should have specified this more.

fingolfin:
Move the check for identical filters into `EqFlags`, and actually added a test for using `=` on equal but not identical filters (for identical filters, GAP has a hotpath that causes `=` to return immediately, so it is impossible to call `EqFlags` with identical arguments that way)

src/opers.c
```diff
@@ -642,53 +657,41 @@ Obj FuncAND_FLAGS (
...
-                SET_AND_CACHE_FLAGS( flags2, cache );
-                CHANGED_BAG(flags2);
+
+#       ifdef HPCGAP
+            if (!PreThreadCreation) {
+                locked = flags2;
+                HashLock(locked);
             }
+#       endif
+        cache  = AND_CACHE_FLAGS(flags2);
```
ChrisJefferson: Your comment above, and code here, don't seem to agree -- you say you get the cache of the smaller, but you get the cache of the bigger (I don't mine which gets fixed, as I don't think it matters).

fingolfin: Oops, indeed, will fix.

fingolfin: Fixed

PR #2297, opened 2018-03-25
WIP: tests: pass CFLAGS and LDFLAGS to gac

fingolfin:
... so that on Travis, we pass `-fprofile-arcs -ftest-coverage` to gac, and thus get coverage data for the compiled test code.

At least that's the theory, let's see if it works...

### DIFF
--- a/tst/test-compile/run_compiled_dynamic.sh
+++ b/tst/test-compile/run_compiled_dynamic.sh
@@ -20,7 +20,7 @@ rm -rf .libs "$gfile.comp"*
 
 "$gac" "$gfile" -d -C -o "$gfile.dynamic.c" 2>&1 >/dev/null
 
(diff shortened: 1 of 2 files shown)


fingolfin:
This is missing a suitable call to `gcov`, I think.

Anyway, the ultimate goal is to have coverage of e.g. https://codecov.io/gh/gap-system/gap/src/master/src/compiled.h go up - we compile code in `tst/test-compile` which e.g. uses `C_MAKE_INTEGER_BAG`, but right now this does not show up in the coverage results.

fingolfin:
I think now the problem is not a missing gcov call, but rather that we delete all intermediate files, and that they are in the /tmp dir to start with. Anyway, resolving this seems tricky, and I don't want to attempt it at this point.

PR #2299, opened 2018-03-25
Add configure option to select GC; introduce USE_GASMAN macro

### LINKED ISSUE #1470: Add configure flag for selecting the garbage collector
This might be an option `--with-gc=VALUE`, where `VALUE` currently could be `gasman` or `boehm` (and possibly also `none`). In the future, `julia` might be another possible value.

If HPC-GAP is disabled, then `gasman` would be default; if HPC-GAP is enabled, then `boehm`. But we'd want to allow `boehm` in regular GAP, too. As for `gasman`, I don't think it is thread safe, so for HPC-GAP we'd probably show an error that the two cannot be combined.

Vaguely related: issue #1272

fingolfin:
Also rename `BOEHM_GC` to `USE_BOEHM_GC`, for uniformity. Also change the code to use `USE_GASMAN` and `USE_BOEHM_GC` more consistently; e.g. don't assume that HPC-GAP uses Boehm and non-HPC-GAP uses GASMAN.

This is part of #2092 (I wrote this code for @rbehrends but since that work currently doesn't seem to go anywhere, I think it's best to merge this now, instead of letting it bitrot).

Resolves #1470

### DIFF
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -21,6 +21,9 @@ ABI_CFLAGS = @ABI_CFLAGS@
 HPCGAP = @HPCGAP@
 ADDGUARDS2 = @ADDGUARDS2@
 
+# garbage collector source files
+GC_SOURCES = @GC_SOURCES@
+
 # compatibility mode
 COMPAT_MODE = @COMPAT_MODE@
 GAPARCH = @GAPARCH@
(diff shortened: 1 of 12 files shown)


ChrisJefferson:
This looks reasonable. I haven't checked the resulting GAP (boehm in non-HPC GAP) works.

fingolfin:
Actually, right now this does *not* allow selecting Boehm GC in plain GAP ! It just lays the ground work for that -- I have not attempted to even compile GAP in that mode so far.

PR #2300, opened 2018-03-25
Add FileId to function calls in profiling

ChrisJefferson:
Cherry-pick #2197 onto 4.9. While not technically required, it will make life easier if the formats in 4.9 and master are kept close. This won't break using old version of the profile package.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -16,6 +16,8 @@
 #include <src/code.h>
 #include <src/gap.h>
 #include <src/hookintrprtr.h>
+#include <src/io.h>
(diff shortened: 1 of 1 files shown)


fingolfin:
Please make sure to set suitable labels and milestones for any PRs you create (and ideally issues, too). In this case, a milestone of 4.9.1 would be appropriate. Also, is this a change for the release notes? If not, please add a "not for release notes" label, to save us trouble later on. If it *is* for the release nots, adding a "bug" or "enhancement" label would be very helpful.

PR #2303, opened 2018-03-25
Restrict CallbackForAllBags to builds using GASMAN

fingolfin:
This function is not available in Boehm GC -- we did provide an
empty stub for it, but that's a dangerous illusion, so we are better
off removing it.


fingolfin:
Adjusted to now also avoid avoid COMPACT_TYPE_IDS and FLUSH_ALL_METHOD_CACHES  in HPC-GAP.

PR #2304, opened 2018-03-25
Travis: reset CFLAGS and LDFLAGS when building packages

### LINKED ISSUE #2248: Fix codecov report for io.c
Right now, <https://codecov.io/gh/gap-system/gap/src/master/src/io.c> is clearly invalid -- e.g. there are comment lines in there which are marked as code.

I suspect this is because we also compile the IO package, and it also has a file `src/io.c`. And the codecov uploader tries to grab any profiling/coverage data it can find, and upload that. My suspicion is that the two then get mixd up.

To fix this, we just need to stop codecov from seeing coverage data for the IO package. On way to ensure that would be to remove `-fprofile-arcs -ftest-coverage` from the CFLAGS for IO. Another would be to insert an `rm` command which deletes the relevant .gcov files. I suspect that inspecting a Travis log file, esp. the part where it uploads coverage data, might help to locate that gcov file.

fingolfin:
This prevents them from being compiled with coverage gathering; we do
this because otherwise gcov may confuse IO's src/io.c with GAP's.

Resolve #2248 (I hope...)

### DIFF
--- a/etc/ci-prepare.sh
+++ b/etc/ci-prepare.sh
@@ -95,7 +95,10 @@ then
   fi
 
   # Compile io and profiling packages
-  "$SRCDIR/bin/BuildPackages.sh" --strict --with-gaproot="$BUILDDIR" io* profiling*
+  # we deliberately reset CFLAGS and LDFLAGS to prevent them from being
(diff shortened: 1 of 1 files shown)


fingolfin:
It works :-)

Issue #2301, opened 2018-03-25
Crash in partial perm


ChrisJefferson:
With kernel debug turned on:

```
gap> PartialPerm([1,2,8],[3,4,1,2]);
gap: src/pperm.c:166: SET_IMG_PPERM: Assertion `DOM_PPERM(f) == ((void *)0) || LEN_PLIST(img) == LEN_PLIST(DOM_PPERM(f))' failed.
```

james-d-mitchell:
Weird, this shouldn't even be called. Fix coming.

fingolfin:
This was fixed by PR #2305

PR #2305, opened 2018-03-26
Fix sparse pperm

james-d-mitchell:
This PR resolves Issue #2301 by checking that the arguments to `PartialPerm` are of equal length, if there are more than two of them.

And increases the code coverage of the tests to ~97%, removes unnecessary randomness from the test file, fixes some bugs in the c, and GAP code, and reformats the GAP code.

### DIFF
--- a/lib/pperm.gi
+++ b/lib/pperm.gi
@@ -545,12 +545,13 @@ function(arg)
     fi;
   elif Length(arg)=2 then  
     if IsSSortedList(arg[1]) and ForAll(arg[1], IsPosInt) and
-     IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsPosInt) then 
+     IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsPosInt) and
+     Length(arg[1]) = Length(arg[2]) then 
       return SparsePartialPermNC(arg[1], arg[2]); 
     else
       ErrorNoReturn("usage: the 1st argument must be a set of positive integers ",
       "and the 2nd argument must be a duplicate-free list of positive ",
-      "integers");
+      "integers of equal length to the first");
     fi;
   fi; 
  
--- a/lib/pperm.gi
+++ b/lib/pperm.gi
@@ -11,8 +11,13 @@
 InstallMethod(IsGeneratorsOfMagmaWithInverses,
  "for a partial perm collection",
 [IsPartialPermCollection],
-coll-> ForAll(coll, x-> DomainOfPartialPerm(x)=DomainOfPartialPerm(coll[1]) and
-ImageSetOfPartialPerm(x)=DomainOfPartialPerm(coll[1])));
+function(coll)
+  return ForAll(coll, x -> DomainOfPartialPerm(x)
+                           = DomainOfPartialPerm(coll[1])
+                           and
+                           ImageSetOfPartialPerm(x)
+                           = DomainOfPartialPerm(coll[1]));
+end);
 
 # attributes
 
@@ -40,51 +45,53 @@ InstallMethod(NrComponentsOfPartialPerm, "for a partial perm",
 InstallMethod(ComponentsOfPartialPerm, "for a partial perm",
 [IsPartialPerm], COMPONENTS_PPERM);
 
-InstallMethod(IsIdempotent, "for a partial perm", 
+InstallMethod(IsIdempotent, "for a partial perm",
 [IsPartialPerm], IS_IDEM_PPERM);
 
-InstallMethod(IsOne, "for a partial perm", 
+InstallMethod(IsOne, "for a partial perm",
 [IsPartialPerm], IS_IDEM_PPERM);
 
-InstallMethod(FixedPointsOfPartialPerm, "for a partial perm", 
+InstallMethod(FixedPointsOfPartialPerm, "for a partial perm",
 [IsPartialPerm], FIXED_PTS_PPERM);
 
-InstallMethod(NrFixedPoints, "for a partial perm", 
+InstallMethod(NrFixedPoints, "for a partial perm",
 [IsPartialPerm], NR_FIXED_PTS_PPERM);
 
-InstallMethod(MovedPoints, "for a partial perm", 
+InstallMethod(MovedPoints, "for a partial perm",
 [IsPartialPerm], MOVED_PTS_PPERM);
 
-InstallMethod(NrMovedPoints, "for a partial perm", 
+InstallMethod(NrMovedPoints, "for a partial perm",
 [IsPartialPerm], NR_MOVED_PTS_PPERM);
 
-InstallMethod(LargestMovedPoint, "for a partial perm", 
+InstallMethod(LargestMovedPoint, "for a partial perm",
 [IsPartialPerm], LARGEST_MOVED_PT_PPERM);
 
 InstallMethod(LargestImageOfMovedPoint, "for a partial perm",
 [IsPartialPerm],
 function(f)
   local max, i;
-  
-  if IsOne(f) then 
+
+  if IsOne(f) then
     return 0;
   fi;
-  
-  max:=0;
-  for i in [SmallestMovedPoint(f)..LargestMovedPoint(f)] do 
-    if i^f>max then max:=i^f; fi;
+
+  max := 0;
+  for i in [SmallestMovedPoint(f) .. LargestMovedPoint(f)] do
+    if i ^ f > max then
+      max := i ^ f;
+    fi;
   od;
   return max;
 end);
 
-InstallMethod(SmallestMovedPoint, "for a partial perm", 
-[IsPartialPerm], 
+InstallMethod(SmallestMovedPoint, "for a partial perm",
+[IsPartialPerm],
 function(f)
   local m;
   m := SMALLEST_MOVED_PT_PPERM(f);
-  if m = fail then 
+  if m = fail then
     return infinity;
-  else 
+  else
     return m;
   fi;
 end);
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
While we are here (this could go into a different PR),

PartialPerm([1,2],[2^100,2^1000]) also doesn't work, do you also need to check for IsSmallIntRep?

james-d-mitchell:
@ChrisJefferson good idea, I'll adjust this PR for the sake of simplicity.

fingolfin:
@james-d-mitchell do you still plan to make that adjustment? Or should we merge this, and you'll submit another PR later?

james-d-mitchell:
I am making an adjustment, but it will take a little while.

fingolfin:
Great, thanks!

lib/pperm.gi
```diff
@@ -225,7 +226,8 @@ InstallMethod(AsPartialPerm, "for a transformation and list",
 [IsTransformation, IsList], 
 function(f, list)
   
-  if not IsSSortedList(list) or not ForAll(list, IsPosInt) then 
+  if not IsSSortedList(list)
+      or not ForAll(list, x -> IsSmallIntRep(x) and IsPosInt(x)) then
```
fingolfin: Off-topic: you can actually write `ForAll(list`, IsSmallIntRep and IsPosInt);` (whether that's a good idea or not is probably a matter of test; in any case, I am not saying you should change this, just pointing it out ;-)

src/pperm.c
```diff
@@ -1699,125 +1694,71 @@ Obj FuncNaturalLeqPartialPerm(Obj self, Obj f, Obj g)
...
 
     def = DEG_PPERM(f);
     deg = DEG_PPERM(g);
-    dej = MAX(def, deg);
-    if (dej < 65536) {
-        join = NEW_PPERM2(dej);
-        SET_CODEG_PPERM2(join, dej);
+
+    if (def > deg) {
+        return FuncJOIN_IDEM_PPERMS(self, g, f);
```
fingolfin: Nice way to simplify the following code a lot! As a minor tweak, you could also do this:
```C
  SWAP(Obj, f, g);
  SWAP(UInt, def, deg);
```

src/pperm.c
```diff
@@ -2539,7 +2471,7 @@ Obj FuncShortLexLeqPartialPerm(Obj self, Obj f, Obj g)
         if (DEG_PPERM4(g) == 0)
             return False;
         rankg = RANK_PPERM4(g);
-        domg = DOM_PPERM(f);
+        domg = DOM_PPERM(g);
```
fingolfin: Oops, good catch!

src/pperm.c
```diff
@@ -6322,8 +6214,8 @@ Obj OnSetsPPerm(Obj set, Obj f)
                     *ptres++ = tmp;
                 }
             }
-            else { /* this case cannot occur since I think POW is not defined
-                      */
+            else {
+                // This case cannot occur since POW is not defined
```
fingolfin: I don't understand this comment. Best I can come up with is that `POW` is a macro, and clearly defined as such... so clearly you mean something else, but what?

src/pperm.c
```diff
@@ -6465,7 +6355,8 @@ Obj FuncOnPosIntSetsPartialPerm(Obj self, Obj set, Obj f)
     }
 
     PLAIN_LIST(set);
-    res = NEW_PLIST_WITH_MUTABILITY(IS_MUTABLE_PLIST(set), T_PLIST_CYC_SSORT, LEN_LIST(set));
+    res = NEW_PLIST_WITH_MUTABILITY(IS_MUTABLE_PLIST(set), T_PLIST_CYC_SSORT,
+                                    LEN_LIST(set));
```
fingolfin: Doesn't really matter, but you could use `LEN_PLIST` here, since we just converted `set` to a plist in the line before.

src/pperm.c
james-d-mitchell: Nice, I didn't know about `SWAP`, I've updated the PR to use this instead.

src/pperm.c
james-d-mitchell: I couldn't get the code to enter one case, which lead me to spot this 👍

src/pperm.c
james-d-mitchell: This is an old comment, and I do not know what I meant either, my memory is that I couldn't get the code to enter this case. I will update this.

src/pperm.c
james-d-mitchell: Done.

src/pperm.c
james-d-mitchell: Actually, I remember what this means now, after putting the analogous code from `permutat.c` in here instead of the error. I have updated the comment.

james-d-mitchell:
@fingolfin I think I have fixed all the things you requested @fingolfin.

fingolfin:
Tests fail in 32bit builds

james-d-mitchell:
Yup, I’ll fix it in the morning

src/pperm.c
```diff
@@ -341,7 +336,7 @@ static Obj PreImagePPermInt(Obj pt, Obj f)
 * GAP functions for partial perms
 *****************************************************************************/
 
-
+// FIXME why is this a function???
 Obj FuncEmptyPartialPerm(Obj self)
```
fingolfin: Alas, whether there is a good justification or not, this function is documented in the GAP reference manual, so we cannot just remove it. Of course it is also documented as a "short hand for PartialPerm([]);." So one could replace this C function by a GAP function. And then, if you felt strongly about it, also mark it as obsolete.

Not sure that new C comment is going to help anybody, though ;-).

src/pperm.c
```diff
@@ -6322,9 +6212,18 @@ Obj OnSetsPPerm(Obj set, Obj f)
...
-                ErrorQuit("not yet implemented!", 0L, 0L);
+            else {
+                // This case does not work since PowIntPPerm2/4 only work for
+                // INTOBJ's (i.e. small integers). The analogous code in
+                // permutat.c uses the macro POW, which calls PowIntPerm2/4,
+                // which if called with a non-small positive integer returns
+                // that integer, since every permutation fixes every positive
+                // integer (small or not). The methods for PowIntPPerm2/4 give
+                // an error if the argument is a non-small integer, and so we
+                // cannot call POW here.
```
fingolfin: Actually, why does it do that? Why not also have a method for (positive) bigint+pperm which returns the bigint unchanged, mirroring the permutation code?

src/pperm.c
```diff
@@ -6322,9 +6212,18 @@ Obj OnSetsPPerm(Obj set, Obj f)
...
+                ErrorQuit("the first argument (a set) must consist of "
+                          "integers less than 2 ^ %d but contains %s",
+                          NR_SMALL_INT_BITS, (Int)TNAM_OBJ(*ptset));
```
fingolfin: The standard way in the kernel to phrase such an error would be this: `ErrorQuit("<set> must be a list of small integers", 0,0);`. This uses that this kernel function is called for `OnSets`, which is documented as `OnSets( set, g )`.

But of course the `OnSets` function for permutation does not contain such a message -- because it relies on `POW` producing an error, if needed. And that in turn makes it possible for the user resp. packages to install custom methods for `\^` -- which then in turn would automatically be useable via `OnSets`.

I am not saying that's necessarily a good idea, but it's a possibility there. We could of course make it stricter, but then it would be kinda nice if the pperm and perm code would behave similarly for this.

(That said, I'd accept either approach in this PR, we can work on unification in another PR)

src/pperm.c
```diff
@@ -6322,9 +6212,18 @@ Obj OnSetsPPerm(Obj set, Obj f)
                     *ptres++ = tmp;
                 }
             }
-            else { /* this case cannot occur since I think POW is not defined
-                      */
-                ErrorQuit("not yet implemented!", 0L, 0L);
+            else {
+                // This case does not work since PowIntPPerm2/4 only work for
+                // INTOBJ's (i.e. small integers). The analogous code in
```
fingolfin: We call them "small integers" or sometimes "immediate integers" elsewhere, and I'd stick to that terminology here, w/o brining it up again.

src/pperm.c
```diff
@@ -6322,9 +6212,18 @@ Obj OnSetsPPerm(Obj set, Obj f)
...
+                // permutat.c uses the macro POW, which calls PowIntPerm2/4,
+                // which if called with a non-small positive integer returns
+                // that integer, since every permutation fixes every positive
+                // integer (small or not). The methods for PowIntPPerm2/4 give
```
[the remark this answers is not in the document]
james-d-mitchell: Oops, will fix.

src/pperm.c
james-d-mitchell: I'll update this.

src/pperm.c
james-d-mitchell: Yes, we could do that, but for now this is not implemented, and so I opted to fix the comment and make the error message more meaningful.

src/pperm.c
james-d-mitchell: I agree it would be better if the pperm code worked in more or less the same way as the permutat code, but didn't want to make further changes in this PR.

src/pperm.c
james-d-mitchell: Ok.

src/pperm.c
james-d-mitchell: Also just to point out that partial perms unlike permutations do not fix all points where they are not defined, they are simply undefined on those points, so `PowIntPPerm2/4` should return `0` in those cases.

james-d-mitchell:
I've updated this according to your comments @fingolfin and the 32-bit tests should also be fixed by these changes.

james-d-mitchell:
Any reason to not to merge this? @fingolfin @ChrisJefferson

fingolfin:
Two reasons why @ChrisJefferson and me don't merge: we are both on vacation; and new changes keep being added to this PR, instead of new PRs, i.e. you added changes right after asking why this isn't being merged... So I won't merge this from here, as I can't review the new changes sensibly from here. But of course somebody else who is not on vacation or strike or busy with other things could step up. Otherwise, I'll get to it next week.

In the meantime, people shall feel free to review and merge all my open PRs, some of which have been open and waiting for weeks or even months.

james-d-mitchell:
Thanks @fingolfin, and apologies for disturbing your vacation, I didn't know. I added two commits today, but otherwise this is unchanged from when you previously approved it.  

I'm not sure I have the expertise to review all your open PR's...

james-d-mitchell:
I'll amend the second to last commit to avoid OOB access, and drop the last commit altogether since it's not really related.

PR #2307, opened 2018-03-26
Minor fix for HPC-GAP docs; fix and simplify DisplayCacheStats

fingolfin:


lib/profile.g
```diff
@@ -986,16 +991,11 @@ BIND_GLOBAL("DisplayCacheStats",function()
         "NEW_TYPE misses",
     ];
 
-    pos := [ 1, 2, 3, 4, 9, 5, 6, 7, 8, 10, 11, 12, 13, 14, 15 ];
```
fingolfin: Just to explain: this list `pos` was used to permute the output of `OPERS_CACHE_INFO`. Instead of that, I changed `OPERS_CACHE_INFO` to put the data into the list in the desired data from the start

PR #2309, opened 2018-03-27
Fix input validation in SHELL(); add tests for src/gap.c (mostly for input validation)

fingolfin:


ChrisJefferson:
I'm happy with these. It's possible some of them will have test conflicts on weird OSes (including Cygwin), but the easiest way to catch that will be to merge and see if anyone spots anything.

PR #2313, opened 2018-03-27
Avoid GC problem in Cyclotomic

ChrisJefferson:
DIFF can cause a GC, which might move ResultCyc

Found using 'memory canary'.

This should also be cherry-picked onto 4.9.

### DIFF
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -688,7 +688,9 @@ Obj             Cyclotomic (
             res[0] = cof;
         else {
             CHANGED_BAG( ResultCyc );
-            res[0] = DIFF( INTOBJ_INT(0), cof );
+            Obj negcof = DIFF( INTOBJ_INT(0), cof );
+            res = BASE_PTR_PLIST(ResultCyc);
+            res[0] = negcof;
         }
         n = 1;
     }


src/cyclotom.c
```diff
@@ -688,7 +688,9 @@ Obj             Cyclotomic (
             res[0] = cof;
         else {
             CHANGED_BAG( ResultCyc );
-            res[0] = DIFF( INTOBJ_INT(0), cof );
+            Obj negcof = DIFF( INTOBJ_INT(0), cof );
```
fingolfin: I am happy to merge this, though you might want to change this to `Obj negcof = AInvInt(cof);`.

ChrisJefferson: No I can't, because cof can be a rational :) (seperate note, some int related functions could produce more helpful errors when given non integres)

fingolfin: Ah, sorry for the red herring

fingolfin: (Actually, I have a déjà-vu ... didn't we discuss the exact same thing at the same spot before?)

fingolfin:
Backported to stable-4.9 as 01b3b13417718fe8bf2d568a10948e5db38bad72

PR #2314, opened 2018-03-27
Remove code that was commented out for a looong time

fingolfin:
These seem to be GAP3 leftovers. Perhaps Thomas or Steve have some insights on this?


fingolfin:
Regarding `Denominator` and `Numerator`: We currently have `NumeratorOfRationalFunction` and `NumeratorRat` (and similar for the denominator). I can only guess that there was a reason (but which?) to replace the single attributes by multiple ones? In any case, I wouldn't mind having these available under the shorter and uniform names. But this is beyond the scope of this PR. Also, the removed methods really are trivial, i.e., if somebody wants to add `Denominator` and `Numerator` attributes (or operations), the hard part is not rewriting those methods.

I'd thus like to merge this PR as is. If you think it's worthwhile, we could make an issue that suggests to add `Denominator` and `Numerator`?

ThomasBreuer:
The pull request can be merged as is.
I have created an issue (#2324) in order not to forget about `Denominator` and `Numerator`.

fingolfin:
Thanks, Thomas!

PR #2316, opened 2018-03-27
fix inconsistent comment format in src/*.h files

ThomasBreuer:
This makes it easier to deal programmatically with the comments.
The most relevant changes concern '*V' vs. '*F',
round and square brackets in '*V' and '*T' lines,
and the analogue of GAPDoc's <C> element.
Besides that, I have fixed some obvious typos.

### DIFF
--- a/src/blister.h
+++ b/src/blister.h
@@ -164,8 +164,9 @@ static inline const UInt * CONST_BLOCK_ELM_BLIST_PTR(Obj list, UInt pos)
 **
 *F  MASK_POS_BLIST( <pos> )  . . . .  bit mask for position of a Boolean list
 **
-**  MASK_POS_BLIST(<pos>) returns a UInt with   a single set bit in  position
-**  (pos-1) % BIPEB, useful for accessing the pos'th element of a blist
+**  'MASK_POS_BLIST(<pos>)' returns a UInt with a single set bit in  position
+**  '(<pos>-1) % BIPEB',
+**  useful for accessing the <pos>-th element of a blist.
 */
 static inline UInt MASK_POS_BLIST(UInt pos)
 {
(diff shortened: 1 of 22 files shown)


src/blister.h
```diff
@@ -164,8 +164,9 @@ static inline const UInt * CONST_BLOCK_ELM_BLIST_PTR(Obj list, UInt pos)
 **
 *F  MASK_POS_BLIST( <pos> )  . . . .  bit mask for position of a Boolean list
 **
-**  MASK_POS_BLIST(<pos>) returns a UInt with   a single set bit in  position
-**  (pos-1) % BIPEB, useful for accessing the pos'th element of a blist
+**  'MASK_POS_BLIST(<pos>)' returns a UInt with a single set bit in  position
+**  '(<pos>-1) % BIPEB',
+**  useful for accessing the <pos>-th element of a blist.
```
fingolfin: Why the line break after the comma? Same question for the two other places this occurs.

src/lists.h
```diff
@@ -308,8 +308,8 @@ extern Obj (*ElmListFuncs[LAST_REAL_TNUM+1]) ( Obj list, Int pos );
 **  is the responsibility  of the caller to  ensure that <pos>  is a positive
 **  integer.
 **
-**  The difference between ELM_LIST and ELMB_LIST is that ELMB_LIST accepts
-**  an object as the second argument
+**  The difference between 'ELM_LIST' and 'ELMB_LIST' is that 'ELMB_LIST'
+**  accepts an object as the second argument
```
fingolfin: Perhaps add a full stop?

src/objects.h
```diff
@@ -28,7 +28,7 @@
 
 /****************************************************************************
 **
-*T  Obj . . . . . . . . . . . . . . . . . . . . . . . . . . . type of objects
+*t  Obj . . . . . . . . . . . . . . . . . . . . . . . . . . . type of objects
```
fingolfin: Are the valid letters and their meaning documented somewhere?

src/blister.h
ThomasBreuer: Without line break, the line would be longer than 77 characters,
and if a line break must be introduced, the comma is the best place for it,
concerning readability.

src/objects.h
ThomasBreuer: No.
As far as I know, the whole markup in `src` files is not defined anywhere.
It would make sense to document this
if the current format or some extension of it would be kept for the future.
This is one of the questions which I had raised in a message to gap@gap-system.org,
concerning an official documentation for the GAP C API.

PR #2318, opened 2018-03-27
Add ability for FormatOutput to print GAP strings directly

### LINKED ISSUE #2317: FormatOutput and GAP strings
`FormatOutput` is a GAP kernel function which does low-level printing. It understands C strings for example, by writing `%s`. This function is used in many error-related functions. For example, in FuncOUTPUT_LOG_TO, streams.c

```
   ErrorReturnVoid( "OutputLogTo: cannot log to %s",
                         (Int)CSTR_STRING(filename), 0L,
                         "you can 'return;'" );
```

Or gvars.c, AssGVar:

```
ErrorReturnVoid(
            "Variable: '%s' is read only",
            (Int)NameGVar(gvar), 0L,
            "you can 'return;' after making it writable" );
```

Both of these pieces of code are invalid, because before `FormatOutput` is eventually called inside `ErrorReturnVoid`, an object is allocated, possibly causing a GC which will make the `CSTR_STRING` in the first case, and the `NameGVar` in the second case, invalid.

There are similar problems in a few other functions, and it's very easy to introduce similar bugs.

I can see two fixes for this:

1) Extend `FormatOutput` to understand GAP strings directly (we could require it only understand strings in `StringRep`. This would be quickest.

2) Add and make use of some slightly higher-level formatting function, which only uses GAP-level strings.

ChrisJefferson:
This adds '%g' and '%G' to GAP's FormatOutput, to allow print GAP strings directly. This replaces many places where we pass a C pointer to a GAP string around, in some cases leading to problems with strings getting GCed.

Resolves #2317

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1530,6 +1530,10 @@ void ResetOutputIndent(void)
 **          character string which is printed.
 **  '%S'    the corresponding argument is the address of  a  null  terminated
 **          character string which is printed with escapes.
+**  '%g'    the corresponding argument is the address of an Obj which points
+**          to a string in STRING_REP format which is printed in '%s' format
+**  '%G'    the corresponding argument is the address of an Obj which points
+**          to a string in STRING_REP format which is printed in '%S' format
 **  '%C'    the corresponding argument is the address of  a  null  terminated
 **          character string which is printed with C escapes.
 **  '%d'    the corresponding argument is a signed integer, which is printed.
@@ -1551,6 +1555,7 @@ static inline void FormatOutput(
     void *state, const Char *format, Int arg1, Int arg2 )
 {
   const Char *        p;
+  Obj                 arg1obj;
   Int                 prec,  n;
   Char                fill;
 
@@ -1612,8 +1617,19 @@ static inline void FormatOutput(
       arg1 = arg2;
     }
 
-    /* '%s' print a string                                         */
-    else if ( *p == 's' ) {
+    /* '%s' or '%g' print a string                               */
+    else if ( *p == 's' || *p == 'g') {
+
+      // If arg is a GAP obj, get out the contained sting, and
+      // set arg1obj so we can re-evaluate after any possible GC
+      // which occurs in put_a_char
+      if (*p == 'g') {
+        arg1obj = (Obj)arg1;
+        arg1 = (Int)CSTR_STRING(arg1obj);
+      }
+      else {
+        arg1obj = 0;
+      }
 
       /* compute how many characters this identifier requires    */
       for ( Char * q = (Char *)arg1; *q != '\0' && prec > 0; q++ ) {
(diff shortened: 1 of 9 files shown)


src/io.c
```diff
@@ -1613,8 +1617,13 @@ static inline void FormatOutput(
       arg1 = arg2;
     }
 
-    /* '%s' print a string                                         */
-    else if ( *p == 's' ) {
+    /* '%s' or '%g' print a string                               */
+    else if ( *p == 's' || *p == 'g') {
+
+      /* If a GAP obj, get out the contained string              */
+      if(*p == 'g') {
```
fingolfin: add space after `if`

src/io.c
```diff
@@ -1643,8 +1652,14 @@ static inline void FormatOutput(
       arg1 = arg2;
     }
 
-    /* '%S' print a string with the necessary escapes              */
-    else if ( *p == 'S' ) {
+    /* '%S' or '%G' print a string with the necessary escapes    */
+    else if ( *p == 'S' || *p == 'G' ) {
+
+      /* If a GAP obj, get out the contained string              */
+      if(*p == 'G') {
```
fingolfin: Add space after `if`

fingolfin:
I'd appreciate if you could add labels to all new PRs you make, though: at least indicate if it is a bug fix and/or enhancement, and use the "not for release notes" label if appropriate (it is here, I think?).

ChrisJefferson:
Now attached tags. While this touches lots of pieces of code, I think it should be backported to 4.9, in case anyone hits one of the crashes it fixes (which will be extremely rare, as GC problems always are).

ChrisJefferson:
steve has pointed out a possible problem, so dont merge :)

src/io.c
```diff
@@ -1613,8 +1617,13 @@ static inline void FormatOutput(
...
 
-    /* '%s' print a string                                         */
-    else if ( *p == 's' ) {
+    /* '%s' or '%g' print a string                               */
+    else if ( *p == 's' || *p == 'g') {
+
+      /* If a GAP obj, get out the contained string              */
+      if(*p == 'g') {
+          arg1 = (Int)CSTR_STRING((Obj)arg1);
+      }
```
fingolfin: Thinking about this some more, isn't this still unsafe, because `put_a_char` could still trigger a GC?

ChrisJefferson: Exactly, that's what @stevelinton said to me. Will come up with a fix.

fingolfin: At least that's what I encountered when working on printing for integers. I ended up converting huge strings first into a string object, then calling `PrintString1` on it, see `PrintInt` `in src/integer.c`

fingolfin:
I am not so sure about the backporting, at least a hypothetical rare crash does not necessarily justify one. To me, a backport is justified if it's a crippling behavior: e.g. a very frequent crash, or something which leads to mathematically incorrect results. That's something to address ASAP. Of course one may argue that a GC issue can also lead to corrupt memory but not a crash, and that this could then lead to mathematically incorrect results... But that's quite a stretch to make...

Anyway, I am not going to block a backport, I am just saying we should have really high bars for that, and we have to weigh the benefits against the potential regressions.

I still hope we can get the final 4.9 release out in April, and I *really* would like to have 4.10 out at the end of the year, too (ideally in October or so, i.e. half a year after 4.9)

fingolfin:
@ChrisJefferson I assume this resolves #2317 ? I added a corresponding "resolves #2317" to the PR description, so that when this gets merged, the issue is auto-closed, and doesn't stay open (as has happened frequently in the past). I recommend adding such markers to the description for future PRs, too!

src/stringobj.c
```diff
@@ -754,23 +754,7 @@ void PrintString (
...
-         off++, scanout++) {
-      PrStrBuf[scanout] = p[off];
-    }
-    PrStrBuf[scanout] = '\0';
-    Pr( "%s", (Int)PrStrBuf, 0L );
-    for (; off<len && CHARS_STRING(list)[off]==0; off++) {
-      Pr("%c", 0L, 0L);
-    }
-  }
+  Pr("%g", (Int)list, 0L);
```
fingolfin: Might be better to just replace all callers of `PrintString1`? But this can be done in a future PR, of course.

fingolfin: For the record, no distributed or hosted on GitHub package uses `PrintString1` directly

ChrisJefferson:
I agree with not backporting this, as you say, let's get 4.9 released and head towards 4.10.

PR #2320, opened 2018-03-27
Avoid GC problem in PowPerm2Int/PowPerm4Int

ChrisJefferson:
AInvInt can cause a garbage collection, which moves the memory that ptKnown points to, so calculate the negation first.


fingolfin:
Backported (with minor necessary adjustment) to stable-4.9 as 15ce49d37f45a1ba458067600657d6634bb7f477

PR #2321, opened 2018-03-28
trans: resolve GC bug Issue 2319

### LINKED ISSUE #2319: GC bug in MOVED_PTS_TRANS (trans.c:2652/2662)
In trans.c:2652, we have: `AssPlist(out, ++len, INTOBJ_INT(i + 1));`. This can cause a memory allocation, which breaks `ptf2` (mentioned in the previous line). The same bug occurs with `ptf4` a few lines later.

james-d-mitchell:
Resolves #2319

### DIFF
--- a/src/trans.c
+++ b/src/trans.c
@@ -2650,6 +2650,7 @@ Obj FuncMOVED_PTS_TRANS(Obj self, Obj f)
         for (i = 0; i < deg; i++) {
             if (ptf2[i] != i) {
                 AssPlist(out, ++len, INTOBJ_INT(i + 1));
+                ptf2 = CONST_ADDR_TRANS2(f);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


fingolfin:
@james-d-mitchell Thanks! One minor request: If possible, I'd appreciate if you could please put "Resolves #XY" in the description body instead of (or in addition to) the subject line next time, so that the issues automatically gets closed when the PR is merged. I did it for you (well, or rather, for us, I guess ;-) this time.

fingolfin:
@ChrisJefferson actually, I don't think we should mark these GC fixes as "not for release notes". Instead, I'll add a single entry a la "fixed lots of potential GC related crashes", referencing the PRs

james-d-mitchell:
@fingolfin Will do!

fingolfin:
Reminder to self re release notes: of course for fixes we backport, we should add them to the GAP 4.9 release notes, not 4.10

fingolfin:
Backported to stable-4.9 via ae5aa319cf7655337bed75dcb628689126cae870

PR #2323, opened 2018-03-28
Various minor changes and tweaks

fingolfin:
This PR is probably best reviewed commit-by-commit. It fixes comment typos, remove some duplicate comments and code, tweaks some error messages, etc.


fingolfin:
Urgh, package tests fail because `cvec` does not load: it references `extern unsigned long PolsFF[];`, so turning that `static const` breaks `cvec` -- silently, of course, because this is accessing an undocumented global in GAP. This is actually an extra motivation for me to make our global vars `static` whenever possible, to prevent abuse: if a package really needs access to a variable, then this should be officially requested and an "API" be made available (e.g. by documenting the variable in question, and adding it to a header, or by providing an accessor function).

Anyway, I'll revert this part of the change for now, and file an issue with cvec.

PR #2326, opened 2018-03-28
SimpleGroup: some fixes; better input validation; tests

fingolfin:
* adjust case of error messages
* catch various invalid parameters classical groups, instead of
  returning non-simple groups with IsSimpleGroup forcibly set
* fix error message for 2E(n,q) with n<>6 (referred to 3D with n<>4)
* add tests

### DIFF
--- a/grp/simple.gi
+++ b/grp/simple.gi
@@ -349,7 +349,7 @@ local brg,str,p,a,param,g,s,small,plus,sets;
 	SetIsSimpleGroup(g,true);
 	return g;
       else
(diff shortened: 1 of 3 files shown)


fingolfin:
hmm, so what do you call groups of order 1, 6 or 18 labelled by GAP as simple groups? :-). They lead to all kinds of interesting interactions subsequently.

hulpke:
They are not simple, of course. If the name corresponds to a simple group an isomorphic group is returned. Otherwise the result is undefined :-)

PR #2328, opened 2018-03-28
Fix QuaternionGroup for non-abelian number fields, tests

fingolfin:
* fix QuaternionGroupCons for matrix groups over non-abelian
  number fields
* remove unused TorusOfNaturalGL
* add more tests

I did not add a test for the bug fix, because the only way I know how to get a non-abelian number field in GAP is to use `AlgebraicExtension` and that doesn't work very well; I had to use all kinds of trickery to get one that triggered the (obvious) bug in the code, and then the resulting group is not very pleasant to work with. So, not test, but the fix is IMHO clearly correct.

### DIFF
--- a/grp/basicmat.gi
+++ b/grp/basicmat.gi
@@ -101,7 +101,7 @@ function( filter, fld, n )
     cyc := E( n / 2 );
     bas := CanonicalBasis( Field( fld, [ cyc ]  ) );
     one := 1;
-  elif 0 = n mod Characteristic( fld )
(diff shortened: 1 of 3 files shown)


fingolfin:
Re `TorusOfNaturalGL`: this is a local function, i.e., nobody was ever able to call this (just try entering `TorusOfNaturalGL` into any GAP session, it'll say "not defined"). So I doubt this. (And of course the code  is still in the repo).

PR #2330, opened 2018-03-29
Use --enable-large-config for Boehm GC

markuspf:
Without this setting Boehm GC runs out of heap sections very quickly for
any kind of orbit enumeration, rendering HPC-GAP effectively useless for this
purpose.


fingolfin:
@markuspf could you please add labels to future PRs? At least some basic ones: "not for release notes" in this case would be most helpful to those of us who try to produce usable release notes ;-). Of course here some more are relevant, but less important for work flow.

PR #2332, opened 2018-03-29
Fix various GC bugs

ChrisJefferson:
This fixes another little bunch of gc related bugs.

With this, test install passes memory-checking, except for a trans bug (in investigation somewhere else), and an xgap problem (which I haven't investigated, and could be because xgap is doing something which memory-checking doesn't like).

### DIFF
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -239,52 +239,68 @@ void            PrintCyc (
 {
     UInt                n;              /* order of the field              */
     UInt                len;            /* number of terms                 */
-    const Obj *         cfs;            /* pointer to the coefficients     */
-    const UInt4 *       exs;            /* pointer to the exponents        */
     UInt                i;              /* loop variable                   */
 
     n   = INT_INTOBJ( NOF_CYC(cyc) );
     len = SIZE_CYC(cyc);
     Pr("%>",0L,0L);
     for ( i = 1; i < len; i++ ) {
-        /* get pointers, they can change during Pr */
-        cfs = CONST_COEFS_CYC(cyc);
-        exs = CONST_EXPOS_CYC(cyc,len);
-        if (      cfs[i]==INTOBJ_INT(1)    && exs[i]==0 )
+        // Store value in local variable, as they can change during Pr
+        const Obj *   cfs = CONST_COEFS_CYC(cyc);
+        const UInt4 * exs = CONST_EXPOS_CYC(cyc, len);
+        Obj           cfsi = cfs[i];
+        UInt4         exsi = exs[i];
+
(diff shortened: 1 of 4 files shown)


src/cyclotom.c
```diff
@@ -239,52 +239,53 @@ void            PrintCyc (
...
-        else if ( LT(INTOBJ_INT(0),cfs[i]) && exs[i]==1 ) {
-            Pr("%>+",0L,0L); PrintObj(cfs[i]); Pr("%>*%<E(%d%<)",n,0L); }
-        else if ( LT(INTOBJ_INT(0),cfs[i])              && i==1 ) {
-            Pr("%>",0L,0L); PrintObj(cfs[i]);
-            Pr("%>*%<E(%d)%>^%2<%d",n,(Int)exs[i]); }
-        else if ( LT(INTOBJ_INT(0),cfs[i]) ) {
-            Pr("%>+",0L,0L); PrintObj(cfs[i]);
-            Pr("%>*%<E(%d)%>^%2<%d",n,(Int)exs[i]); }
-        else if ( cfs[i]==INTOBJ_INT(-1)   && exs[i]==0 )
+        else if ( cfsi==INTOBJ_INT(1)                       && i==1 )
```
fingolfin: Since this rewrites `PrintCyc` almost entirely anyway, perhaps `clang-format` it?

src/tietze.c
```diff
@@ -240,6 +240,7 @@ Obj FuncTzSortC (
         SET_LEN_PLIST( rels,  i );  SHRINK_PLIST( rels,  i );
         SET_LEN_PLIST( lens,  i );  SHRINK_PLIST( lens,  i );
         SET_LEN_PLIST( flags, i );  SHRINK_PLIST( flags, i );
+        ptTietze = ADDR_OBJ( tietze );
         ptTietze[TZ_NUMRELS] = INTOBJ_INT(i);
```
fingolfin: Note that `ptTietze` is a plist, so perhaps instead replace both lines by `SET_ELM_PLIST(ptTietze, TZ_NUMRELS, INTOBJ_INT(i));`?

ChrisJefferson: That's a good point. In general, there seems to be far too much storing of pointers in this file, for (I assume) efficiency. Keeping them in the odd tight loop is fine, but I'd prefer to be proactive (grab the pointer just before you need it) rather than reactive (try to regrab the pointer whenever you might break it, which is easy to get wrong).

src/cyclotom.c
ChrisJefferson: I will. I originally left it so it would be easy for you to spot that I'd done a search+replace

PR #2334, opened 2018-03-29
Split gap.c into three files

fingolfin:
The new `errors.c` contains  error & break loop related code; `modules.c` contains code to handle modules. Motivation for this is that `gap.c` has become the dumping grounding for any new additions -- people follow the rule "if you don't know where to put it, stick it into `gap.c`", and I find this makes it hard to navigate it, and find stuff.

Here's a breakdown on how it looks after this PR
```
wc -l src/gap.c src/modules.c src/error.c
    1856 src/gap.c
    1044 src/modules.c
     597 src/error.c
    3497 total
```

I'd like to continue this work later, e.g. move gasman specific stuff into `gasman.c`.


fingolfin:
To clarify: the first few commits which move around a lot of stuff are indeed hard to verify. But I really only moved code around, with the absolute minimum set of changes to fix compilation (i.e. mostly adding the right headers into the new files, and in 1-2 cases including the new headers from other files). Any other changes, such as for code formatting, or minor fixes I noticed while moving the code, are in the final set of commits.

BTW, one obvious thing to do next is also not yet in this PR: namely to check all files which include gap.h, to see if they could/should instead include modules.h resp. error.h, and then do that -- and afterwards, removing these two includes from gap.h

PR #2337, opened 2018-03-30
More fixes for memory issues

ChrisJefferson:
More memory changes found with --enable-memory-checking. In some cases we could look at removing functions which give C pointers to strings, but I didn't check what functions are in use by what packages.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -57,9 +57,14 @@ void SET_NAME_FUNC(Obj func, Obj name)
     FUNC(func)->name = name;
 }
 
+Obj NAMI_OBJ_FUNC(Obj func, Int i)
+{
+    return ELM_LIST(NAMS_FUNC(func),i);
+}
+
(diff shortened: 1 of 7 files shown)


src/calls.c
```diff
@@ -57,9 +57,14 @@ void SET_NAME_FUNC(Obj func, Obj name)
     FUNC(func)->name = name;
 }
 
+Obj NAMI_OBJ_FUNC(Obj func, Int i)
+{
+    return ELM_LIST(NAMS_FUNC(func),i);
+}
+
```
fingolfin: I'd prefer to not do this, let's instead change NAMI_FUNC, NAME_LVAR, etc to return a GAP string, to fix a ton of additional identical problems not yet covered by this PR.

I just sat down (during vacation, yes...) and had a quick look, it shouldn't be a problem, and no package calls these, either. Might even be able to do this for NAME_RNAM and NameGVar, too

PR #2338, opened 2018-03-30
Add IsFinitelyGeneratedMagma property for magmas

fingolfin:
UPDATE: you can ignore the original description of this PR below, and directly skip to [my comment down below](https://github.com/gap-system/gap/pull/2338#issuecomment-851678170) and/or read the commit message and the changes. I'll leave the old description in, even though it is outdated and contains tidbits that are no longer relevant.

OLD TEXT FOLLOWS:

Also make `IsFinitelyGeneratedGroup` equivalent to `IsFinitelyGenerated and IsGroup` via suitable implications, and similarly for `IsFinitelyGeneratedMonoid` and `IsFinitelyGenerated and IsMonoid`.

As a byproduct, since `IsGroup` implies `IsMonoid`, now also `IsFinitelyGeneratedGroup` implies `IsFinitelyGeneratedMonoid`

Note that we cannot just use synonyms here, i.e. this is no good:
```gap
DeclareSynonymAttr( "IsFinitelyGeneratedGroup", IsFinitelyGenerated );
```
First off, one cannot install methods for and-filters, and the `lpres` package installs a method for `IsFinitelyGeneratedGroup`. Secondly, one cannot set and-filters to false via a set, and there are packages out there doing this:
```
    SetIsFinitelyGeneratedGroup(G, false);
```

One thing that may be controversial is the name `IsFinitelyGenerated`. @ThomasBreuer pointed out on #2252  that `IsFinitelyGenerated` ...
> ... might lead to confusion in the situation of rings and ideals where it makes a difference whether a domain is finitely generated as an additive structure, as a multiplicative structure, as a ring, as an ideal

The alternative discussed there was to call it `IsFinitelyGeneratedMagma` (or even, as originally suggested, `IsFinitelyGeneratedMagmaWithInverses`). Now, if the consensus ends up preferring that, I can of course adjust this PR to use `IsFinitelyGeneratedMagma` or whatever instead. However, please consider this:

* We already have `IsCommutative` (=`IsAbelian`) and `IsAssociative` in place for magmas, even though they can be misleading, too. Now, of course a past mistake should not justify a new one; but I mention this because I wonder whether having these defined for magmas in general really caused problems anywhere? And if so, where? I looked in `lib/overload.g`, where we deal with things like `IsNilpotent`, `IsSolvable` etc., but there is just this comment:
```
##  Note that operations such as `IsCommutative' are not of this type,
##  since the result means the same for any multiplicative structure.
```
(actually, I do not understand this claim: one of the main contention points addressed by `lib/overload.g` are Lie algebras; but for those `IsCommutative` actually *does* have a different interpretation, at least in characteristic 2, where `[x,y]=[y,x]` always holds, but the Lie algebra is only commutative it `[x,y]=0` for all x,y)

* Personally, I am more concerned about the clash between multiplicative and additive "finite generation" (or variants beyond)... There are very many applications where one is concerned about finite generations of modules, for example;  `IsFinitelyGeneratedModule`. But we do ignore this for `IsCommucative`, too (resp. we have `IsAdditivelyCommutative` for the additive case; and have nothing like `IsAssociative` for additive magmas, which apparently are defined in GAP to be associative (?).

* Even if we call it `IsFinitelyGeneratedMagma`, doesn't the ambiguity remain anyway?  E.g. the principal ideal generated by `x` in the polynomial ring `Q[x]` is of course finitely generated as an ideal in the polynomial ring; but as a Q-module / vector space, it is not finitely generated (granted, this example now is about the *additive* structure, but I don't see why similar example couldn't possibly exist in the multiplicative setting, too?)


fingolfin:
I've just revised this PR. I agree now with @ThomasBreuer that this should *not* be named `IsFinitelyGenerated` -- because indeed one might want to add one day something like `IsFinitelyGeneratedRing`, and that would clash with `IsFinitelyGeneratedMagma` (e.g. `Integers` is finitely generated as a ring but not as a magma).

I've cleaned this PR now up; here's the new commit message:

> Add `IsFinitelyGeneratedMagma` property for magmas
> 
> ... and a set of related changes:
> 
> - Ensure that for groups, it is equivalent to `IsFinitelyGeneratedGroup`, and for monoids, it is equivalent to `IsFinitelyGeneratedMonooid`, in the sense that if one is set to true/false, then so is the other.
> 
> - Set `IsFinitelyGeneratedMagma` for Integers, PositiveIntegers, etc.
> 
> - Modify `Subfields` method for `IsField and IsFFECollection` to instead require `IsField and IsFinite and IsFFECollection` to ensure it has higher rank than the method `IsField and IsFinite` (this was the case before this PR, but anymore with it). Note that strictly speaking, this is also a correctness change: the (hypothetical) union of all finite fields of fixed characteristic `p` is a field and FFE collection but not finite (this kind of field is not currently representable in GAP, but in theory it could be added)

lib/magma.gd
```diff
@@ -562,6 +562,38 @@ InstallFactorMaintenance( IsCommutative,
...
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareProperty( "IsFinitelyGeneratedMagma", IsMagma );
+
+InstallFactorMaintenance( IsFinitelyGeneratedMagma,
+    IsMagma and IsFinitelyGeneratedMagma, IsObject, IsMagma );
+
+InstallTrueMethod( IsFinitelyGeneratedMagma, IsMagma and IsFinite );
+
```
ThomasBreuer, line 595: Why not also `InstallTrueMethod( IsMagma, IsFinitelyGeneratedMagma );`, as for `IsFinitelyGeneratedGroup` and `IsFinitelyGeneratedMonoid`?

ThomasBreuer (approved):
This looks fine.
I have just left two comments.

tst/testinstall/opers/IsFiniitelyGenerated.tst
```diff
@@ -0,0 +1,27 @@
+gap> START_TEST("IsFiniitelyGenerated.tst");
```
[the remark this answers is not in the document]
fingolfin: Oops. copy and pasted the filename, which was misspelled. Will fix.

lib/magma.gd
fingolfin: Thanks, added

PR #2340, opened 2018-03-30
Tweak ShowImpliedFilters some more

fingolfin:
Now for "May imply with", we also hide implications which already
follow from the other filter. For example,  before:
```
gap> ShowImpliedFilters(IsFinitelyGeneratedGroup);
...
+IsNilpotentGroup
   IsSupersolvableGroup
   IsSolvableGroup
   IsPolycyclicGroup
   IsNilpotentByFinite
```
After this change:
```
gap> ShowImpliedFilters(IsFinitelyGeneratedGroup);
...
+IsNilpotentGroup
   IsPolycyclicGroup
```


fingolfin:
Fixed

PR #2341, opened 2018-03-30
kernel: split sysmem.c/.h off from system.c/.h; other kernel cleanup

fingolfin:


ChrisJefferson:
This breaks `./configure --enable-debug --enable-memory-checking`

fingolfin:
@ChrisJefferson should be fixed now

PR #2342, opened 2018-03-30
Move unused code from grp/classic.gi to dev/deleted-code/

### LINKED ISSUE #2327: Undocumented and unused functions in grp/classic.gi: remove? document?
While looking for code that needs testing, I noticed in https://codecov.io/gh/gap-system/gap/src/master/grp/classic.gi that there are several undocumented functions which are never used anywhere, and in a few cases, seem to have functionality duplicated by functions that were added later. These are not called anywhere in the library, nor by any distributed package, nor by any package I am aware on here on GitHub

For each, I propose that we either remove it (possibly moving it to e.g. `dev/deleted-code/` or so instead); or else document it and add tests.

These are the functions:
* `EichlerTransformation`
* `WallForm` (used by `SpinorNorm`)
* `SpinorNorm`
* `WreathProductOfMatrixGroup` (see also`MatWreathProduct` in `lib/gprdmat.gi`, which is equally undocumented; added by @hulpke in 3f339849d20415d4999b835b455e805be3fb1650)
* `TensorWreathProductOfMatrixGroup`
* `CentralProductOfMatrixGroups`

fingolfin:
This code is never called by anything, nor is it documented. For some
of it (matrix group wreath products), newer code exists in the library.

Since the code may become relevant again for Alexander Hulpke one day,
let's not remove it right away, but instead move it to dev/deleted-code/

Resolves #2327

### DIFF
--- a/dev/deleted-code/grp-classic.g
+++ b/dev/deleted-code/grp-classic.g
@@ -1,0 +1,184 @@
+#
+# This code used to be in grp/classic.gi
+#
+
+#############################################################################
+##
+#F  EichlerTransformation( <g>, <u>, <x> )  . .  eichler trans of <u> and <x>
+##
+BindGlobal( "EichlerTransformation", function( g, u, x )
+    local   e,  b,  i;
+
+    # construct matrix of eichler transformation in <e>
+    e := [];
+
+    # loop over the standard vectors
+    for b  in One( g )  do
+        i := b
+             + (b*InvariantBilinearForm(g).matrix*x)*u
+             - (b*InvariantBilinearForm(g).matrix*u)*x
+             - (b*InvariantBilinearForm(g).matrix*u)
+	       *((x*InvariantQuadraticForm( g ) )*x)*u;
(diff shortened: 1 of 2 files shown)


fingolfin:
There is no "urgency" here at all. Also, this is less about removing unused code, than identifying it, and moving it to a place where it can either be retrieved again from for future use, or eventually removed.

As it is, there are thousands and thousands of lines of code in GAP which are unused, undocumented, never called, and haven't been touched for a decade or two. The problem I have with that is that e.g. it makes it harder to identify actually used code. A recent example is the Gcd code; or with this PR, somebody looking for matrix wreath products by grepping the GAP code base might well have ended up looking at the "wrong" (unused and old) code, instead of your new code (although that, too, is not yet really hooked up, but I gather you are working on that).

As to working on composition tree functionality: I'd be delighted to contribute to that, too, but don't have a plan for how to do that myself, and nobody else seems to have prepared a "plan", so there is that :-(.

PR #2344, opened 2018-04-06
Reduce number of functions and macros which return pointers into GAP objects

fingolfin:
More specifically, avoid returning `CSTR_STRING(foo)` for GAP strings `foo`, as this is very prone to introduce bugs triggered by a garbage collection in the wrong moment.

Some thoughts:
* the new `%H` added to `Pr()` actually replaced *all* instances of `%I`, so we could drop the old `%I` and rename `%H` to `%I`, thus reducing the final diff in this PR quite a bit.
* I checked carefully that none of the functions changed and/or renamed by this PR are used by any packages (with a tiny exception for ParGAP, which wraps NAME_LVAR, but only to aid debuggers like GDB/LLDB; this is trivial to fix, but I consider ParGAP unmaintained at this point and thus don't care about breaking it further).

Closes #2337

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -57,9 +57,9 @@ void SET_NAME_FUNC(Obj func, Obj name)
     FUNC(func)->name = name;
 }
 
-Char * NAMI_FUNC(Obj func, Int i)
+Obj NAMI_FUNC(Obj func, Int i)
 {
-    return CSTR_STRING(ELM_LIST(NAMS_FUNC(func),i));
+    return ELM_LIST(NAMS_FUNC(func),i);
 }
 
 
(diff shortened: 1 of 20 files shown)


ChrisJefferson:
Thanks, this is better than my changes were. I'll trust you checked all the packages for possible breakage! (Although tests will show us fairly soon after merging).

After this is merged and settled, I will do another full pass of memory checking, just to make sure nothing was missed.

fingolfin:
Yes, I checked for conflicts with packages rather extensively: in my worktree, I have both a "bootstrap" pkg dir (i.e. with all the currently distributed packages, resp. their latest versions ready for distribution), and also another pkg dir with clones of all packages in `gap-packages`, plus a few more hosted elsewhere. I then used ripgrep to grep for any function or macro whose return value I modified, to see if any package uses them. That yielded only one hit: pargap has a C function which calls `NAMI_FUNC`, but (a) it's in an `#if 0` block, and (b) a comment explains that it is meant to help debugging (presumably in gdb/lldb) -- that comment also expresses the wish that GAP itself should do that. Well, good news, we do now, so that part of pargap is obsolete. ;-)

PR #2345, opened 2018-04-07
Add some tests for associative words; remove some dead code; cleanup

fingolfin:


markusbaumeister:
Why should some code be removed that is currently unused? It looks like a method that could be useful in some algorithms.

fingolfin:
Unused code is untested code, and thus quite possibly buggy code. It is also undocumented; reverse engineering what inputs it accepts, with which constraints, and what its output is, may well require more work than writing this code again from scratch. And even if not: if someone needs this functionality, then with high probability they would not know this commented out code even exists, and thus would write it from scratch anyway. (People even do this for documented code, after all).

fingolfin:
Finally, this code has been there, commented out, since 2001-01-26. If nobody missed it in the past 17 years, what are the chances they'll miss it in the next 17 years?

Keeping this around is akin to keeping old junk around in the attic; who know, perhaps one day you might need it? For 95% of it, you'd better have thrown it away years ago...

fingolfin:
@markusbaumeister thanks for the review!

The C code changes really are only cosmetic; they rename two Cs function to match the names of the GAP functions they define, as we do for (almost) all other such functions. Shouldn't be user visible at all.

PR #2349, opened 2018-04-09
Add line continuation tests; move a kernel comment to right spot; make some scanner functions static

fingolfin:
This is a subset of PR #2215. I hope we can merge this quickly, then I can simplify that PR. Also, by merging this first, we have assurance that the line continuations tests really work before the change to line continuation processing.

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -597,10 +597,7 @@ void GetNumber ( UInt StartingStatus )
 
 /****************************************************************************
 **
-*F  GetEscapedChar() . . . . . . . . . . . . . . . . get an escaped character
-**
-**  'GetEscapedChar' reads an escape sequence from the current input file
-**  into the variable *dst.
+*F  GetOctalDigits()
 **
 */
 static inline Char GetOctalDigits( void )
@@ -639,6 +636,15 @@ static inline Char CharHexDigit( Char c )
     }
 }
 
+
+/****************************************************************************
+**
+*F  GetEscapedChar() . . . . . . . . . . . . . . . . get an escaped character
+**
+**  'GetEscapedChar' reads an escape sequence from the current input file
+**  into the variable *dst.
+**
+*/
 Char GetEscapedChar( void )
 {
   Char result = 0;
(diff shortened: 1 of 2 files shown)


src/scanner.c
```diff
@@ -359,7 +359,7 @@ static Char GetCleanedChar( UInt *wasEscaped ) {
 }
 
 
-void GetNumber ( UInt StartingStatus )
+static void GetNumber(const UInt StartingStatus)
 {
   Int                 i=0;
```
ChrisJefferson: This is a minor comment, but do we construct non-pointer arguments anywhere else? (Never noticed it). This is a possibly significant style change, so I wanted it to be intentional.

fingolfin: What do you mean with "construct"? Perhaps a typo and you meant "use const argument"? The only reason to add this is to help humans reason about the code: it is now clear that the starting status variable doesn't get changed later on.

I see no significant style change with that, though?!

ChrisJefferson: Sorry, autocorrect. I meant constraint argument. It's just we could probably make a significant chunk of the arguments to C functions construction, but I thought this was the first.

fingolfin: Autocorrect agsin? :-)

Anyway, I will remove the const here, and move it to the commmit it belongs to, then we can discuss more when it comes up again.

fingolfin:
@ChrisJefferson removed the `const` here for now.

BTW, I just checked, and at least const pointer (such as `const char *`) are already used frequently for arguments. But perhaps you meant to specifically highlight const non-pointer arguments? Those I already used for e.g. `EvalRef/AssignRef/UnbindRef/IsBoundRef` in `src/read.c`, though that indeed seems to about it. (There, I also used it to make sure I don't accidentally modify the argument, thus simplifying how one reasons about the code).

ChrisJefferson:
I don't mind us adding const to arguments (const char* is more common I agree, mainly because otherwise you can't pass in const pointers), but I wanted it to be a concious choice, not just a dribble coming in here and there.

PR #2352, opened 2018-04-11
Fix tests when 'ToolsForHomalg' is loaded

fingolfin:
`ToolsForHomalg ` by @sebasguts replaces `RunImmediateMethods` with a custom function, which throws of the test output. Fixed by using another function to test the printing of compiled functions.


sebasguts:
Hmm, should we might change ToolsForHomalg in a certain way, so that test output might not depend on the loading and behaviour of ToolsForHomalg?

fingolfin:
Nah -- I don't see how, and anyway, this fix already exists and is fine for us.

sebasguts:
Alright then :)

PR #2355, opened 2018-04-11
HPC-GAP: add some tests for channels, fix crash in InspectChannel, reduce code duplication

fingolfin:
This should further decrease coverage fluctuations.


hpcgap/tst/testinstall/hpc/channels.tst
```diff
@@ -0,0 +1,61 @@
+# Test("../../hpcgap/tst/testinstall/hpc/channels.tst");
```
fingolfin: Oops, that line should be removed (and perhaps instead a comment added which explains that the tests below so far are single-threaded only, and more tests should be added which tests channels across threads).

PR #2356, opened 2018-04-12
Refactor parts of the scanner

fingolfin:
Removing the line continuation logic from the scanner enables several ways to simplify the code.

Also, I reformatted `GetSymbol` to restore the manual whitespace alignment it used to have. I know some people prefer to just run `git clang-format`, but IMHO, this much more compact version of the code makes it much easier to see very quickly what's going on, so I think it's worth keeping it. In any case, the current formatting is messed up, so I think this is arguably an improvement either way, and `git clang-format` could still be run in the future if that's really what we want...

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -622,7 +622,7 @@ static inline Char CharHexDigit( Char c )
 static Char GetEscapedChar(void)
 {
   Char result = 0;
-  Char c = PEEK_CURR_CHAR();
+  Char c = GET_NEXT_CHAR();
 
   if ( c == 'n'  )       result = '\n';
   else if ( c == 't'  )  result = '\t';
@@ -695,7 +695,6 @@ static void GetStr(void)
 
     /* handle escape sequences                                         */
     if ( c == '\\' ) {
-      c = GET_NEXT_CHAR();
       STATE(Value)[i] = GetEscapedChar();
     }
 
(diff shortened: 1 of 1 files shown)


src/scanner.c
```diff
@@ -207,36 +207,37 @@ static void GetIdent(void)
...
-            else if ( c == 'r'  && i < SAFE_VALUE_SIZE-1 )  STATE(Value)[i] = '\r';
-            else if ( c == 'b'  && i < SAFE_VALUE_SIZE-1 )  STATE(Value)[i] = '\b';
-            else if ( i < SAFE_VALUE_SIZE-1 )  {
-                STATE(Value)[i] = c;
+            switch(c) {
+            case 'n': c = '\n'; break;
+            case 't': c = '\t'; break;
+            case 'r': c = '\r'; break;
+            case 'b': c = '\b'; break;
+            default:
```
ChrisJefferson: Just checking you are aware you've changed the semantisc here (now isQuoted=1 is true even if i < SAFE_VALUE_SIZE-1)

fingolfin: I suppose you mean `i >= SAFE_VALUE_SIZE-1`.  Yes, I am aware: while this change the semantics locally, it shouldn't matter globally, because if `i >= SAFE_VALUE_SIZE-1` then we trigger a syntax error anyway.

This reminds me: I should continue my PR for consistently enforcing a single length restriction for identifiers (IIRC, currently, some places use 1022, some 1023, some 1024 and some none).

PR #2360, opened 2018-04-13
Added Documentation for DecomPoly and NormalizerViaRadical

hulpke:
Apart from a minor syntax change, this PR installs DecomPoly and NormalizerViaRadical as proper functions and adds basic documentation.

This comes as a PR, since git told me I had to do a PR and could not just push the changes.

### DIFF
--- a/lib/algfld.gd
+++ b/lib/algfld.gd
@@ -188,3 +188,45 @@ DeclareGlobalFunction("AlgExtEmbeddedPol");
 
 DeclareGlobalFunction("AlgExtSquareHensel");
 
+#############################################################################
+##
+#F  IdealDecompositionsOfPolynomial( <f> [:"onlyone"] )  finds ideal decompositions of rational f
+##
+##  <#GAPDoc Label="IdealDecompositionsOfPolynomial">
+##  <ManSection>
+##  <Func Name="IdealDecompositionsOfPolynomial" Arg='pol'/>
+##
+##  <Description>
+##  Let <M>f</M> be a univariate, rational, irreducible, polynomial. A
+##  pair <M>g</M>,<M>h</M> of polynomials of degree strictly
+##  smaller than that of <M>f</M>, such that <M>f(x)|g(h(x))</M> is
+##  called an ideal decomposition. In the context of field
+##  extensions, if <M>\alpha</M> is a root of <M>f</M> in a suitable extension
+##  and <M>Q</M> the field of rational numbers. Such decompositions correspond
+##  to (proper) subfields <M>Q\lt Q(\beta)\lt Q(\alpha)</M>, where <M>g</M> is the
+##  minimal polynomial of <M>\beta</M>.
+##  This function determines such decompositions up to equality of the subfields
+##  <M>Q(\beta)</M>, thus determining subfields of a given algebraic extension.
+##  It returns a list of pairs <M>[g,h]</M> (and an empty list if no such
+##  decomposition exists). If the option <A>onlyone</A> is given it returns at
+##  most one such decomposition (and performs faster).
+##  <Example><![CDATA[
+##  gap> x:=X(Rationals,"x");;pol:=x^8-24*x^6+144*x^4-288*x^2+144;;
+##  gap> l:=IdealDecompositionsOfPolynomial(pol);
+##  [ [ x^2+72*x+144, x^6-20*x^4+60*x^2-36 ],
+##    [ x^2-48*x+144, x^6-21*x^4+84*x^2-48 ],
+##    [ x^2+288*x+17280, x^6-24*x^4+132*x^2-288 ],
+##    [ x^4-24*x^3+144*x^2-288*x+144, x^2 ] ]
+##  gap> List(l,x->Value(x[1],x[2])/pol);
+##  [ x^4-16*x^2-8, x^4-18*x^2+33, x^4-24*x^2+120, 1 ]
+##  gap> IdealDecompositionsOfPolynomial(pol:onlyone);
+##  [ [ x^2+72*x+144, x^6-20*x^4+60*x^2-36 ] ]
+##  ]]></Example>
+##  In this example the given polynomial is regular with Galois group
+##  <M>Q_8</M>, as expected we get four proper subfields.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction("IdealDecompositionsOfPolynomial");
+DeclareSynonym("DecomPoly",IdealDecompositionsOfPolynomial);
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -4530,5 +4530,34 @@ DeclareOperation( "LowIndexSubgroups",
 
 #############################################################################
 ##
+#F  NormalizerViaRadical(<G>,<S>)
+##
+##  <#GAPDoc Label="NormalizerViaRadical">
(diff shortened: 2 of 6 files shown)


lib/algfld.gd
```diff
@@ -188,3 +188,45 @@ DeclareGlobalFunction("AlgExtEmbeddedPol");
...
+##  <Func Name="DecomPoly" Arg='pol'/>
+##
+##  <Description>
+##  Let <M>f</M> be a univariate, rational, irreducible, polynomial. A
+##  pair <M>g</M>,<M>h</M> of polynomials of degree strictly
+##  smaller than that of <M>f</M>, such that <M>f(x)|g(h(x))</M> is
+##  called an ideal decomposition. In the context of field
+##  extensions, if <M>\alpha</M> is a root of <M>f</M> in a suitable extension
+##  and <M>Q</M> the field of rational numbers, such a decomposition corresponds
+##  to (proper) subfields <M>Q\lt Q(\beta)\lt Q(\alpha)</M>, where <M>g</M> is the
```
fingolfin: "such a decomposition" vs. "subfields" -- shouldn't either both be in plural, or both be in singular? 
I guess technically you can argue that of course Q and Q(\alpha) are also subfields of Q(\alpha), but I still find this mildly irritating to read... But really only mildly, I think I got the gist, thanks :-)

lib/algfld.gd
```diff
@@ -188,3 +188,45 @@ DeclareGlobalFunction("AlgExtEmbeddedPol");
...
+##  minimal polynomial of <M>\beta</M>.
+##  This function determines such decompositions up to equality of the subfields
+##  <M>Q(\beta)</M>, thus determining subfields of a given algebraic extension.
+##  It returns a list of pairs <M>[g,h]</M> (and an empty list if no such
+##  decomposition exists). If the option <A>onlyone</A> is given it returns at
+##  most one such decomposition (and performs faster).
+##  <Example><![CDATA[
+##  gap> x:=X(Rationals,"x");;pol:=x^8-24*x^6+144*x^4-288*x^2+144;;
```
fingolfin: Offtopic: the reference manual says this on `X`:
> X is simply a synonym for Indeterminate. However, we do not recommend to use this synonym which is supported only for the backwards compatibility.

Now, don't get me wrong, I am not suggesting you should change the examples; rather, I always use `X` myself, so I am wondering if this last sentence in the manual should just be deleted?

lib/algfld.gd
```diff
@@ -188,3 +188,45 @@ DeclareGlobalFunction("AlgExtEmbeddedPol");
 
 DeclareGlobalFunction("AlgExtSquareHensel");
 
+#############################################################################
+##
+#F  DecomPoly( <f> [:"onlyone"] )  finds ideal decompositions of rational f
```
fingolfin: Open question: should the name of this function stay `DecomPoly`? Wouldn't e.g. `DecompositionsOfPolynomial` more "GAPish"? I don't mind much either way, but once we document the name, we are more or less stuck with it, so I just want this to be a conscious decision.

lib/algfld.gd
```diff
@@ -188,3 +188,45 @@ DeclareGlobalFunction("AlgExtEmbeddedPol");
...
+##    [ x^2-48*x+144, x^6-21*x^4+84*x^2-48 ],
+##    [ x^2+288*x+17280, x^6-24*x^4+132*x^2-288 ],
+##    [ x^4-24*x^3+144*x^2-288*x+144, x^2 ] ]
+##  gap> List(l,x->Value(x[1],x[2])/pol);
+##  [ x^4-16*x^2-8, x^4-18*x^2+33, x^4-24*x^2+120, 1 ]
+##  gap> DecomPoly(pol:onlyone);
+##  [ [ x^2+72*x+144, x^6-20*x^4+60*x^2-36 ] ]
+##  ]]></Example>
+##  In this example the given polynomial is regular with Galois group
+##  <M>Q_8</M>, thus exposing the four proper subfields.
```
fingolfin: I don't understand the last part of this sentence. How does the fact that the Galois group is Q_8 "expose" the four proper subfields?

lib/grp.gd
```diff
@@ -4528,6 +4528,35 @@ DeclareGlobalFunction("GroupEnumeratorByClosure");
...
+##  provided only as a separate function, and not as method for the operation
+##  <C>Normalizer</C>, as it can often be slower than other built-in routines.
+##  In certain hard cases (non-solvable groups with nontrivial radical), however
+##  its performance is substantially superior.  The function thus provided as a
+##  non-automated tool for advanced users.
+##  <Example><![CDATA[
+##  gap> g:=TransitiveGroup(30,2030);;
+##  gap> s:=SylowSubgroup(g,5);;
+##  gap> Size(NormalizerViaRadical(g,s));
+##  28800
```
fingolfin: Sounds good to me.

lib/algfld.gd
hulpke: Yes. (At the time when this was written there was a push to make `X` obsolete as operation.)

lib/algfld.gd
hulpke: Q8 is the only group of order 8 with four proper subgroups.

lib/algfld.gd
hulpke: If we want to change it, then `IdealDecompositionsOfPolynomial` would be the longish name.

lib/algfld.gd
fingolfin: With that hint, I see what you are getting at, but I still don't see how that sentence possibly could convey it... How about this formulation instead:
> In this example the given polynomial is regular with Galois group <M>Q_8</M>, which has exactly four proper subgroups, and hence four proper subfields, which are exposed by the decompositions we obtained.

lib/grp.gd
fingolfin: That's OK, I think (though we might want to mention that the given example is not actually one where `NormalizerViaRadical` is faster?)

lib/algfld.gd
fingolfin: If we document it, I'd kind of prefer the longish name (a shorter alias can still be added). That said, I don't terribly mind if you greatly prefer the shorter name.

hulpke:
@ssiccha 
The worst case cost (regular Galois group 2^n) is exponential in the input, as the output is.

As for best complexity there is a paper by Klüners, van Hoeij and Belabas (I think) that presents a better method (that is harder to implement).

lib/grp.gd
hulpke: Added a line about example and inserted `is`.

lib/grp.gd
```diff
@@ -4543,14 +4543,17 @@ DeclareOperation( "LowIndexSubgroups",
...
+##  its performance is substantially superior.
+##  The function thus is provided as a
 ##  non-automated tool for advanced users.
 ##  <Example><![CDATA[
 ##  gap> g:=TransitiveGroup(30,2030);;
 ##  gap> s:=SylowSubgroup(g,5);;
 ##  gap> Size(NormalizerViaRadical(g,s));
 ##  28800
 ##  ]]></Example>
+##  Note that this example only demonstartes usage, but that in this case
```
fingolfin: Typo: demonstartes -> demonstrates

alex-konovalov:
This is not what the title says. This results in new `IdealDecompositionsOfPolynomial` being documented, but not `DecomPoly`.

PR #2363, opened 2018-04-14
kernel: remove precords code in ariths.c

fingolfin:
The functions `InstallOneObject`, `InstallSumObject`, etc. are meant to
switch between normal and verbose operations for arithmetic operations
on external objects (i.e. positional and component objects, as well as
datobjs, weakpointer objects and possibly those provided by kernel
extensions in packages).

But for some reason, they also did that for precords. My guess is that
this was for backwards compatibility for GAP3 code, but AFAIK we removed
most of that elsewhere, so this code in ariths.c doesn't seem to serve a
purpose anymore; hence we remove it.

Note that even if my understanding is flawed, the impact of this change
should be harmless: It only affects tracing of such methods, but not
their actual use.

Nevertheless, I'd be grateful if people who've been around back then (Steve, Frank, Thomas) could have a look and possibly comment. Thanks!

For context, I just dug out the commit adding these lines:
```
Author: Frank Celler
Date:   1997-05-30 11:32:30 +0000

    started to rewritte record to allow method selection
```

### DIFF
--- a/lib/record.gi
+++ b/lib/record.gi
@@ -135,18 +135,6 @@ InstallMethod( ViewObj,
     Print(" \<\<\<\<)");
 end);
 
-InstallMethod( \=,
-               "record = record",
-               IsIdenticalObj,
-               [ IsRecord, IsRecord ],
-               EQ_PREC );
-
-InstallMethod( \<,
-               "record < record",
-               IsIdenticalObj,
-               [ IsRecord, IsRecord ],
-               LT_PREC );
-
 
 # methods to catch error cases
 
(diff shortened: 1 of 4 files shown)


fingolfin:
Oops, typo in commit title (copied to PR title). Will fix once back at computer.

frankluebeck:
I agree with these changes. I also think that the installations in `ariths.c` were meant for some GAP3 compatibility, but this was never used.

With the installation of EqPrec and LtPRec in the kernel tables it should be no longer necessary to export `EQ_PREC` and `LT_PREC` to GAP level, I would just remove them completely.

fingolfin:
@frankluebeck I now removed `EQ_PREC` and `LT_PREC`, and also fixed the typo in the commit message. If you think this is good now, please consider also formally "approving" this PR, so that it can be merged, as described on https://help.github.com/articles/approving-a-pull-request-with-required-reviews/

PR #2364, opened 2018-04-14
hpcgap: add tests for serialization code; fix or flag bugs in it; also increase kernel ABI version

fingolfin:
Based on hpcgap/demo/serialize2.g

### DIFF
--- a/hpcgap/tst/testinstall/hpc/serialize.tst
+++ b/hpcgap/tst/testinstall/hpc/serialize.tst
@@ -1,0 +1,173 @@
+#
+# Test the HPC-GAP serialization code
+#
+
+#
+gap> CheckSerialization := function(x)
+>   local err, x2;
+>   x2 := DeserializeNativeString(SerializeToNativeString(x));
+>   if TNUM_OBJ(x) <> TNUM_OBJ(x2) or x2 <> x then
+>     err := "Serialization error: ";
+>     Append(err, String(x));
+>   fi;
+> end;;
+gap> CheckSerialization2 := function(x, f)
+>   local err;
+>   if not f(DeserializeNativeString(SerializeToNativeString(x))) then
+>     err := "Serialization error: ";
+>     Append(err, String(x));
+>   fi;
+> end;;
+
+#
+# integers (note that the serialization code switches behaviour
+# depending on bit size: 5 bits (~32), 14 bits (~8192), or more
+#
+gap> for i in [-1000..1000] do CheckSerialization(i); od;
+gap> for i in [8190, 8191, 8192, 8193, 2^30, 2^60, 2^100] do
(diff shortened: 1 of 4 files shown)


fingolfin:
I tried to add a test for serialization of objects, specifically vectors, and my first example did not work; I think this is due to a design flaw of the serialization code (it seems to assume that types are "unique", and even says so in a source comment, but this is simply not true: two compressed immutable GF(2) vectors may have non-identical types (which are "equal" in the sense that their bound positions are equal, but we don't have a `\=` method installed for types, so there's that).

I also wonder if anybody ever tested serialization with component or positional objects (compressed vectors and matrices are data objects).

fingolfin:
I also fixed the serialization of rationals, which was broken: it serialized *pointers* if the numerator/denominator didn't happen to be a small integer. D'oh.

ChrisJefferson:
Just as a sanity check -- is this serialisation code required for HPC-GAP, and are we sure we want it as opposed to extending / improving pickling in IO, which is already well tested.

fingolfin:
I don't think anything in HPC-GAP explicitly depends on it (yet), but OTOH I thought it was a kind of important part of implementing cross-process / cross-machine communication. But of course other approach for that exists; besides IO's pickling, also OpenMath and SCSCP could be listed here.

So we could just drop it, but I'd rather have that an explicit discussion.

Compared to IO's pickling, the two actually are conceptually quite similar from a high level (not surprising, given what they are about), but there are lots of difference in details. IO's pickling may be well tested, but it also has conceptual flaws -- e.g. if you implement a non-standard list type, without implementing explicitly pickling support for it, then IO will pickle into a plist, without a warning. Both require custom code to support serializing/pickling high level objects. Pickling has the strategic advantage that it already implements a lot of those, while for serialization, only compressed vectors and matrices are "supported", but as I mentioned above, this support only semi-works.

One major reservation about IO's pickling is the same I have about IO in general: It ties you to the POSIX API. Of course pickling could in principle be made independent of that, but fact is, nobody has done so, and I don't see anybody doing it anytime soon.

That said, it'd probably make more sense to improve IO's pickling than to keep this code. *HOWEVER* I don't claim to have the full picture here, So I think it would be good to hear @stevelinton and @rbehrends chime in before we just drop this code. In the meantime, I see no problem with merging my fixes, given that I already made them now...

rbehrends:
The main reason for the existence of the serialization code is the speed advantage it has over IO, especially for larger data structures. If IO could be made sufficiently fast, that would be a perfectly acceptable alternative. Right now, the serialization code is primarily used in HPC-GAP in conjunction with ZeroMQ message passing code ([example](https://github.com/gap-packages/ZeroMQInterface/blob/master/gap/znode.g)).

PR #2369, opened 2018-04-17
HPC-GAP: add some tests

fingolfin:
... with the hope that this will take care of the coverage fluctuation in `hpcgap/lib/hpc/stdtasks.g`


fingolfin:
Good point, got rid of the `Display`.

PR #2370, opened 2018-04-18
Move part of testinstall/ctblmoli.tst to teststandard

fingolfin:
This part of the test was relatively recently added by me there, but
turns out to be rather slow (on my system, it takes up 25% or so of the
total runtime of testinstall).


ChrisJefferson:
Thanks, I've been meaning to look at moving some things out of testinstall, to help speed up testing.

PR #2040, opened 2017-12-19
Raise error if eager float literal conversion fails (fixes #1105)

fingolfin:

PR #2371, opened 2018-04-18
WIP: refactor scanner and reader

fingolfin:
This is currently based on #2040, so that PR should be merged first.

Also, I have not yet fully rewritten the float and integer literal scanning (to avoid the `S_PARTIAL*` states); or rather, I had changes for that, but decided to start from scratch with this PR, in an attempt to make the change more incremental, and hence easier to review, at least when looking at it commit-by-commit.

However, I wanted to get this out now, to (a) see what the full test suite makes of it, and (b) to give people an early chance to spot issues and leave comments (though of course most likely nobody will have time to do that ;-).

One thing on which everybody can comment:The commit "scanner: remove GetCleanedChar" is a bit experimental, as it changes user visible behaviour; what do people think about this in terms of how "bad" it is? To illustrate, before the change, you might see this:
```
gap> 12.34\56;
12.34\56;
     ^
```
after:
```
gap> 12.34\56;
Syntax error: Badly Formed Number
12.34\56;
    ^
Error, Variable: '56' must have a value
```

To me, this is acceptable if this allow us to gain substantially simpler code, *but* I wonder if there are other reasons for the existence and use of `GetCleanedChar` in the float parser that I missed? Perhaps @stevelinton has some thought on this?

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -1956,6 +1956,7 @@ void CodeLongFloatExpr(Obj s)
         CodeEagerFloatExpr(s, mark);
     }
     else {
+        // FIXME: this is not GC safe
         CodeLazyFloatExpr(str, l);
     }
 }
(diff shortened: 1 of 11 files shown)


fingolfin:
This has been superseded by various other PRs, most recently PR #2467

Issue #2374, opened 2018-04-18
last2, last3 not available inside a break loop


cdwensley:
This happens in a fresh 4.9.0 :-
gap> 1;2;3;[last,last2,last3];
1
2
3
[ 3, 2, 1 ]
gap> Error("to enter a break loop");
Error, to enter a break loop called from
not in any function at *stdin*:2
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> 1;2;3;[last,last2,last3];      
1
2
3
[ 3, 3, 2 ]

When I queried this on Slack, Chris Jefferson replied: 
Short answer, no. Longer answer, the "bug" is we have a variable called `LastDepth`, which lets you pick how many of the last variables are changed. In this case we have set `LastDepth=1`, which means only last is getting updated. Why would we ever want to support that? 
The simple fix (you could test it) is to change the 4th argument to SHELL in error.g:240 from '1' to '3'. 

This change has been made; testinstall and teststandard pass; and a PR has been made.

alex-konovalov:
@cdwensley shall this be closed now?

cdwensley:
Yes - covered by PR #2375 and #2379.

PR #2375, opened 2018-04-18
make last2 and last3 available in a break loop

cdwensley:
This PR addresses issue #2374. 
I came upon this by accident - it is not a feature that I particularly wish to use. 
Only 'last' is currently available inside a break loop because LastDepth=1 is set there. 
The fix suggested by @ChrisJefferson is to change '1' to '3' in lib/error.g line 240. 
Alternatively we could add "(Note that last2 and last3 are not available inside a break loop.)" to the reference manual. 
The additional suggestion by @stevelinton, "Would it be nicer if the old values of last* were reset on return from the break loop?", has not been addressed.

### DIFF
--- a/lib/error.g
+++ b/lib/error.g
@@ -237,7 +237,7 @@ BIND_GLOBAL("ErrorInner",
         prompt := "brk> ";
     fi;
     if not justQuit then
-        res := SHELL(context,mayReturnVoid,mayReturnObj,1,false,prompt,false,"*errin*","*errout*",false);
+        res := SHELL(context,mayReturnVoid,mayReturnObj,3,false,prompt,false,"*errin*","*errout*",false);
     else
         res := fail;
     fi;


ChrisJefferson:
While there are many other changes we could make, this fixes an obvious bug (at the moment we change last, but not last2 and last3 in break loops, which is just silly)

fingolfin:
Seems OK to me, though of course as always, a test case would be nice. I.e. add a new file to `tst/test-error/`, say `last-access.g`. In it, repeat the inputs you used, like this:
```
1;2;3;[last,last2,last3];
Error("err");
1;2;3;[last,last2,last3];
```
Then, run the `regenerate_error_tests.sh` script in that dir. It should produce a file `last-access.g.out` which you also need to add to the repository. For me, it has this content (without your fix, so the final one will be different):
```
gap> 1;2;3;[last,last2,last3];
1
2
3
[ 3, 2, 1 ]
gap> Error("err");
Error, err called from
not in any function at *stdin*:3
you can 'return;' to continue
brk> 1;2;3;[last,last2,last3];
1
2
3
[ 3, 3, 2 ]
brk> QUIT;
```


That said, it's also fine to merge this as-is, and we'll add a test later.

cdwensley:
Made a test-error/last-access.g as described above, but ./regenerate_error_tests.sh messed up all the *.out files (and created *.out-e files) by putting brk> in place of gap> and brk-2> in place of brk>, e.g. last-access.g.out looks like this: 
brk> 1;2;3;[last,last2,last3];
1
2
3
[ 3, 2, 1 ]
brk> Error("err");
Error, err at GAPROOT/lib/package.gi:1836 called from
func(  ); at GAPROOT/lib/system.g:208 called from
 called from read-eval loop at *errin*:3
you can 'return;' to continue
brk_2> 4;5;6;[last,last2,last3];
4
5
6
[ 6, 5, 4 ]
brk_2> QUIT;
Not sure what to do about this!

fingolfin:
Don't worry about it.

PR #2380, opened 2018-04-20
kernel: make GetSymbol local to scanner.c, and rename it to NextSymbol

fingolfin:
By not exposing `GetSymbol` to the outside, refactoring becomes easier. Also, to me `GetFOO` sounds like an idempotent operation, and I always find it confusing when this is violated. This is why I earlier renamed `GET_CHAR` to `GET_NEXT_CHAR`, and now propose to rename `GetSymbol` to `NextSymbol`.

Subset of PR #2371

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -25,7 +25,7 @@
 #include <src/stringobj.h>
 
 
-static void GetSymbol(void);
+static void NextSymbol(void);
 
 
 /****************************************************************************
@@ -152,7 +152,7 @@ void Match (
 
     // if 'STATE(Symbol)' is the expected symbol match it away
     if ( symbol == STATE(Symbol) ) {
-        GetSymbol();
+        NextSymbol();
     }
 
     /* else generate an error message and skip to a symbol in <skipto>     */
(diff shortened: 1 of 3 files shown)


src/scanner.c
```diff
@@ -876,17 +879,17 @@ static void GetHelp(void)
...
 **
-*F  GetSymbol() . . . . . . . . . . . . . . . . .  get the next symbol, local
+*F  NextSymbol() . . . . . . . . . . . . . . . . . get the next symbol, local
 **
-**  'GetSymbol' reads  the  next symbol from   the  input,  storing it in the
+**  'NextSymbol' reads  the  next symbol from  the  input,  storing it in the
 **  variable 'Symbol'.  If 'Symbol' is  'S_IDENT', 'S_INT' or 'S_STRING'  the
-**  value of the symbol is stored in the variable 'STATE(Value)'.  'GetSymbol' first
+**  value of the symbol is stored in 'STATE(Value)'.  'NextSymbol' first
 **  skips all <space>, <tab> and <newline> characters and comments.
```
fingolfin: Note that this comment is also outdated, it doesn't mention floats. The reason I am not fixing that here is that changes in PR #2371 (resp. the other PRs spawned from it) will require more change here anyway, so I'd prefer to fix this at the end; this ensures that the various PRs derived from PR #2371 "commute", i.e. can be merged in arbitrary order. And anyway, these comments were wrong (or at least incomplete) for years and nobody complained, surely we can wait a bit longer ;-).

PR #2381, opened 2018-04-20
kernel: simplify scanning of large string literals

fingolfin:
This PR simplifies scanning of large string literals. As a side effect, we get rid of `STATE(ValueLen)` (this requires also tweaking the scanner code for `S_HELP`). To achieve this, we introduce `Obj STATE(ValueObj)` which is now used by the string scanning code to store its result; since the scanned data is turned into a string object later on anyway, there is no harm in this.

As an optimization, we use an intermediate (stack based) buffer in which we accumulate data, until the buffer is full resp. we are done scanning, when we transfer the data into the final string object. This way, we don't have to worry about garbage collection, and the compiler can emit better code.

Subset of PR #2371

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1159,39 +1159,6 @@ void ReadLongNumber(
 
 /****************************************************************************
 **
-*F  ReadString( <follow> )  . . . . . . . . . . . . . . read a (long) string
-**
-**  A string is  read by copying parts of `STATE(Value)'  (see scanner.c) given
-**  by `STATE(ValueLen)' into  a string GAP object. This is  repeated until the
-**  end of the string is reached.
-**
-*/
-void ReadString(
-      TypSymbolSet        follow )
-{
-     Obj  string;
-     UInt len;
-
-     C_NEW_STRING( string, STATE(ValueLen), STATE(Value) );
-     len = STATE(ValueLen);
-
-     while (STATE(Symbol) == S_PARTIALSTRING || STATE(Symbol) == S_PARTIALTRIPSTRING) {
-         Match(STATE(Symbol), "", follow);
-         GROW_STRING(string, len + STATE(ValueLen));
-         memcpy(CHARS_STRING(string) + len, STATE(Value),
-                                        STATE(ValueLen));
-         len += STATE(ValueLen);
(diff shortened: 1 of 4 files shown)


src/scanner.c
```diff
@@ -986,7 +996,7 @@ void GetSymbol ( void )
   case '^':         STATE(Symbol) = S_POW;           GET_NEXT_CHAR(); break;
 
   case '~':         STATE(Symbol) = S_TILDE;         GET_NEXT_CHAR(); break;
-  case '?':         STATE(Symbol) = S_HELP;          GetHelp(); break;
+  case '?':                                          GetHelp(); break;
```
ChrisJefferson: Why's this gone? I can't see any bug it could have been causing?

src/read.c
```diff
@@ -2566,10 +2566,8 @@ void ReadTryNext (
 
 void ReadHelp(TypSymbolSet follow)
 {
-    Obj topic;
-
-    C_NEW_STRING(topic, STATE(ValueLen), (void *)STATE(Value));
-    TRY_READ { IntrHelp(topic); };
+    TRY_READ { IntrHelp(STATE(ValueObj)); }
+    STATE(ValueObj) = 0;
```
ChrisJefferson: Why clean ValueObj?

src/scanner.c
fingolfin: It's not gone, I moved it into `GetHelp` for symmetry with the other `GetFOO` functions in the following lines.

src/read.c
fingolfin: Paranoia, I guess: The idea is that we already used this string in `IntrHelp`, and don't want to accidentally reuse it elsewhere (say, due to a big, where some function expects the scanner to set `ValueObj`, but this doesn't happen -- either due to a bug in the scanner, or because the calling code was n fact wrong to assume the scanner would set `ValueObj`).

 In any case, I have another change coming up in another PR where I reset `ValueObj` and `Value` at the start of `Match`, which would make this change partially obsolete (but not generally, I guess).

Anyway, if you are concerned about it now, I can get rid of it before merging.

Issue #2377, opened 2018-04-19
Regression in testextra/grpauto.tst (memory usage)


alex-konovalov:
The following is observed in `teststandard` tests on Jenkins:
```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-master-test/GAPCOPT\
S/64build/GAPTARGET/standard/label/kovacs/GAP-master-snapshot/tst/testextra/gr\
pauto.tst:105
# Input is:
IsomorphismGroups(G,PcGroupCode(CodePcGroup(G),Size(G)))=fail;
# Expected output:
false
# But found:
Error, reached the pre-set memory limit
(change it with the -o command line option)
########
```
In the master branch this appens on both macos and linux jenkins slaves. In the stable-4.9 branch - only on macos. I've tried to run just that example starting GAP with `-o 1g` option and so war was able to reproduce this problem only once. Apparently, there is some interaction between other tests and loaded packages, and balancing around the 1g limit with some randomness and garbage collections making this hard to reproduce.

This is the input:
```
LoadAllPackages();
G:=PcGroupCode( 741231213963541373679312045151639276850536621925972119311,11664);;
H:=PcGroupCode( 888658311993669104086576972570546890038187728096037768975,11664);;
IsomorphismGroups(G,H);
IsomorphismGroups(G,PcGroupCode(CodePcGroup(G),Size(G)))=fail;
```
and this is the memory usage that I observe under macos:
```
Load GAP                131           131 
Load all packages       ---           344
After 1st iso test      142           436
After 2nd iso test      548           1.01g
```
(the last 1.01g value is measured after a break loop happened and I've entered "return" to resume the calculation).

The problem can be reproduced easier if GAP is started with just slightly smaller memory limit `-o 1000m`. Then after `Error, reached the pre-set memory limit` I see this:

```
brk> Where(50);
ImagesRepresentative( map, gen ) at /Users/alexk/GITREPS/gap-stable/lib/mapphomo.gi:311 called from
func( C[i] ) at /Users/alexk/GITREPS/gap-stable/lib/coll.gi:746 called from
List( GeneratorsOfMagmaWithInverses( elms ), function ( gen )
      return ImagesRepresentative( map, gen );
  end ) at /Users/alexk/GITREPS/gap-stable/lib/mapphomo.gi:311 called from
ImagesSet( map, elm ) at /Users/alexk/GITREPS/gap-stable/lib/mapping.gi:185 called from
[...]

PR #2383, opened 2018-04-22
Speed improvements for automorphism groups

hulpke:
Improvements to element order, to `SmallGeneratingSet` for large degree permutation groups (that come up in the automorphisms groups code) and for isomorphism test in the case of solvable automorphism groups.

(This makes the issue in #2377 disappear)

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -941,7 +941,11 @@ local d,a,map,possibly,cG,cH,nG,nH,i,j,sel,u,v,asAutomorphism,K,L,conj,e1,e2,
     u:=ClosureGroup(i,K);
     v:=ClosureGroup(i,L);
     if u<>v then
-      gens:=SmallGeneratingSet(api);
+      if IsSolvableGroup(api) then
+        gens:=Pcgs(api);
+      else
+        gens:=SmallGeneratingSet(api);
+      fi;
       pre:=List(gens,x->PreImagesRepresentative(iso,x));
       map:=RepresentativeAction(SubgroupNC(a,pre),u,v,asAutomorphism);
       if map=fail then
--- a/lib/morpheus.gi
+++ b/lib/morpheus.gi
@@ -16,36 +16,111 @@
 ##
 MORPHEUSELMS := 50000;
 
+# this method calculates a chief series invariant under `hom` and calculates
+# orders of group elements in factors of this series under action of `hom`.
+# Every time an orbit length is found, `hom` is replaced by the appropriate
+# power. Initially small chief factors are preferred. In the end all
+# generators are used while stepping through the series descendingly, thus
+# ensuring the proper order is found.
 InstallMethod(Order,"for automorphisms",true,[IsGroupHomomorphism],0,
 function(hom)
-local map,phi,o,lo,i,start,img;
+local map,phi,o,lo,i,j,start,img,d,nat,ser,jord,first;
+  d:=Source(hom);
+  if Size(d)<=10000 then
+    ser:=[d,TrivialSubgroup(d)]; # no need to be clever if small
+  else
+    if HasAutomorphismGroup(d) then
+      if IsBound(d!.characteristicSeries) then
+        ser:=d!.characteristicSeries;
+      else
+        ser:=ChiefSeries(d); # could try to be more clever, introduce attribute
+        # `CharacteristicSeries`.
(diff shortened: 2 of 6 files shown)


fingolfin:
Typo in commit message: shuitable

lib/morpheus.gi
```diff
@@ -18,34 +18,101 @@ MORPHEUSELMS := 50000;
 
 InstallMethod(Order,"for automorphisms",true,[IsGroupHomomorphism],0,
 function(hom)
-local map,phi,o,lo,i,start,img;
+local map,phi,o,lo,i,j,start,img,d,nat,ser,jord,first;
+  d:=Source(hom);
```
fingolfin: A short comment here as to what this method does would be very helpful. Right now, I see the commit message, which heelps, but if I read that code in a year, I won't know about it... BTW the commit message says a "characteristic series" is computed, but in some cases it only checks if the series is invariant under `hom` (which of course is enough)

So, wow about this
> `#` To compute the order, use a suitable characteristic series 'ser' and work in the factor groups.

(And perhaps "characteristic series" should be "series invariant under 'hom'" ?

lib/morpheus.gi
```diff
@@ -18,34 +18,101 @@ MORPHEUSELMS := 50000;
...
+          if j<Length(ser) then
+            nat:=NaturalHomomorphismByNormalSubgroup(d,ser[j]);
+            start:=ImagesRepresentative(nat,map[1][i]);
+            img:=map[2][i];
+            while ImagesRepresentative(nat,img)<>start do
+              img:=ImagesRepresentative(phi,img);
+              lo:=lo+1;
+
+              # do the bijectivity test only if high local order, then it does not
+              # matter. IsBiojective is cached, sop second test is cheap.
```
fingolfin: typos: IsBiojective, sop

lib/morpheus.gi
```diff
@@ -18,34 +18,101 @@ MORPHEUSELMS := 50000;
...
+
+          else
+            start:=map[1][i];
+            img:=map[2][i];
+            while img<>start do
+              img:=ImagesRepresentative(phi,img);
+              lo:=lo+1;
+
+              # do the bijectivity test only if high local order, then it does not
+              # matter. IsBiojective is cached, sop second test is cheap.
```
fingolfin: again: typos

lib/autsr.gi
```diff
@@ -941,7 +941,11 @@ local d,a,map,possibly,cG,cH,nG,nH,i,j,sel,u,v,asAutomorphism,K,L,conj,e1,e2,
     u:=ClosureGroup(i,K);
```
fingolfin: Typo in second commit message: `AmallGeneratingSet`

alex-konovalov:
@hulpke will this work in stable-4.9 too? We need to fix #2377 to release GAP 4.9.1.

hulpke:
@alex-konovalov 
Yes, this should also work in 4.9

fingolfin:
@alex-konovalov Why do we need to backport this to 4.9? I am not convinced, and very reluctant to do so. See my comment on PR #2377 for details

alex-konovalov:
@fingolfin your reluctance definitely has some point. If this breaks manual examples, it may also break tests in packages. If we will decide to backport, then only after we will check which impact that may have on packages. If we do not backport, then we need to decide whether to release with #2377 unresolved or not.

lib/morpheus.gi
hulpke: is in now

lib/autsr.gi
hulpke: fixed

lib/morpheus.gi
```diff
@@ -16,36 +16,111 @@
 ##
 MORPHEUSELMS := 50000;
 
+# this method calculates a chief series invariant under `hom` and calculates
+# orders of group elements in factors of this series under action of `hom`.
+# Every time an orbit length is found, `hom` is replaced by the appropriate
+# power. Initiall small chief factors are preferred. In the end all
```
fingolfin: Initiall -> Initially ?

hulpke:
@fingolfin Thanks. I will wait until #2377 is in some way or another sorted out.

lib/morpheus.gi
hulpke: Corrected

Issue #2239, opened 2018-03-04
Idea: record location of `InstallMethod` call


fingolfin:
Right now, one can use e.g. `LocationFunc` or `PageSource` on methods installed for an operation, which is quite handy. But sometimes the same function is installed several times; or a kernel function is installed as a method (possibly also several times), and in these cases, the location of the function definition does not coincide with the location of the method installation.

Thus, it would be nice if we could store the location of the method installation, too. I assume it could be e.g. stored in the methods array, along with the other methods.

The tricky part might be to figure out the location itself. I assume this may require some work on the kernel to be possible. We'd somehow need to retrieve the location of the place `InstallMethod` was called... Perhaps some clever lvar walking can help?!? hum

markuspf:
I have some prototype code that does this (in multiple variations).

My first incarnation had a global variable that complemented the `OPERATIONS` array.

 I think an approach similar to `GET_OPER_FLAGS` (which stores the location of the declarations for each operation already) can work here, too.

fingolfin:
Well, I assume we'd have to add one more entry per method into each `METHODS_OPERATION(oper,n)` list, storing the location info there. Of course *that* in turn requires adjusting all places that deal with `METHODS_OPERATION` -- so at least `INSTALL_METHOD_FLAGS` and the kernel code in `GetMethodUncached`

PR #2384, opened 2018-04-22
When tracing immediates, print location of definition

fingolfin:
When tracing immediates, print location of definition

For now, we print the location of the *method*. Ideally, we should
instead track the location of the InstallImmdiateMethod call. See also
issue #2239.

Before:

    gap> TraceImmediateMethods( );
    gap> g:= Group( (1,2,3), (1,2) );;
    #I  immediate: Size
    #I  immediate: IsCyclic
    #I  immediate: IsCommutative
    #I  immediate: IsTrivial

After:

    gap> TraceImmediateMethods( );
    gap> g:= Group( (1,2,3), (1,2) );;
    #I RunImmediateMethods
    #I  immediate: Size at GAPROOT/lib/coll.gi:174
    #I  immediate: IsCyclic at GAPROOT/lib/grp.gi:34
    #I  immediate: IsCommutative at GAPROOT/lib/magma.gi:190
    #I  immediate: IsTrivial at GAPROOT/lib/magma.gi:124

### DIFF
--- a/hpcgap/src/c_oper1.c
+++ b/hpcgap/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include <src/compiled.h>
-#define FILE_CRC  "452946"
+#define FILE_CRC  "-3929700"
 
 /* global variables used in handlers */
 static GVar G_REREADING;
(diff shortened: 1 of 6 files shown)


fingolfin:
I actually decided to one up this, and fully resolve issue #2239, too, which is now in PR #2393

PR #2385, opened 2018-04-22
kernel: de-duplicate code reading `local` statements; make reader functions static

fingolfin:


src/read.c
```diff
@@ -1499,6 +1499,41 @@ static void ReadFuncExprBody(
 }
 
 
+/****************************************************************************
+**
```
ChrisJefferson: Being super picky, maybe mention nams already contains the list if functiin arguments, and either mention narg==LEN_LIST(nams), or drop it and recalculate it.
?

fingolfin: Actually, I didn't think about this at all. Good catch! I'll drop the `narg` argument.

fingolfin: Done.

fingolfin:
The test failure is rather strange:
```
########> Diff in /home/travis/build/gap-system/gap/tst/testbugfix/2012-07-13-t00248.tst:2
# Input is:
IsDocumentedWord( "d" );
# Expected output:
false
# But found:
Error, no 1st choice method found for `SplitString' on 3 arguments
```
So far, I failed to reproduce this locally, on macOS and Ubuntu. My only guess at this point is that somehow this PR relies on undefined behavior, but I can't figure out how. If @ChrisJefferson or @stevelinton or anybody else has an idea...

PR #2387, opened 2018-04-22
Reduce impact of immediate methods to calculations with groups

### LINKED ISSUE #2386: Time taken by Immediate Methods
This is motivated by #2377:

When running the following example from the test suite
```
G:=PcGroupCode( 741231213963541373679312045151639276850536621925972119311,11664);;
IsomorphismGroups(G,PcGroupCode(CodePcGroup(G),Size(G)))=fail;time;
```
on my machine (just minimal packages, but AutPGroup is loaded), in the current master the calculation takes 234 seconds on my Laptop. If I restart GAP the same way, but issue the command
```
    IGNORE_IMMEDIATE_METHODS:=true;
```
first, the calculation takes 200 seconds, that is close to 20% of the time is spent by immediate methods.

It is possible that this reflects bad programming of mine in the isomorphism test -- a lot of groups and vector spaces are created. However `TraceImmediateMethods` shows that time is spent on (e.g.) testing whether the group is cyclic and perfect because it has size 1, to check whether Size should be set to infinity, because IsFinite has been set to `false`, and so on. 

The immediate methods themselves are probably rather harmless, but they change the type of an object (sometimes multiple tiles) just after it has been created, and the changed type then need to be compared with existing types to avoid duplication of types.

hulpke:
This is an attempt to resolve #2386. It changes group generation that many immediate methods are not triggered any longer, and removes immediate methods that are taken care off by the creation process or that have small payoff.

See the commit text for detailled description.

Note that it will require #2429 which was requested to be taken out (and might not test cleanly without)

### DIFF
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -94,7 +94,7 @@ InstallMethod( PrintObj,
 ##
 InstallImmediateMethod( IsEmpty,
     IsCollection and HasSize, 0,
-    C -> Size( C ) = 0 );
+    C->Size( C ) = 0);
 
 InstallMethod( IsEmpty,
     "for a collection",
@@ -120,8 +120,8 @@ InstallMethod( IsTrivial,
     [ IsCollection ],
     C -> Size( C ) = 1 );
 
-InstallImmediateMethod( IsTrivial,
-    IsCollection and HasIsNonTrivial, 0,
+InstallMethod( IsTrivial,
+    [IsCollection and HasIsNonTrivial], 0,
     C -> not IsNonTrivial( C ) );
 
 
@@ -129,8 +129,8 @@ InstallImmediateMethod( IsTrivial,
 ##
 #M  IsNonTrivial( <C> ) . . . . . . . . .  test if a collection is nontrivial
 ##
-InstallImmediateMethod( IsNonTrivial,
-    IsCollection and HasIsTrivial, 0,
+InstallMethod( IsNonTrivial,
+    [IsCollection and HasIsTrivial], 0,
     C -> not IsTrivial( C ) );
 
 InstallMethod( IsNonTrivial,
@@ -169,8 +169,12 @@ InstallMethod( IsWholeFamily,
 ##
 #M  Size( <C> ) . . . . . . . . . . . . . . . . . . . .  size of a collection
 ##
-InstallImmediateMethod( Size,
-    IsCollection and HasIsFinite and IsAttributeStoringRep, 0,
+#  This used to be an immediate method. It was replaced by an ordinary
+#  method, as the immediate method would get called for every group that
+#  knows it is finite but does not know its size -- e.g.  permutation, pc.
+#  The benefit of this is minimal beyond showing off a feature.
+InstallMethod( Size,true, [IsCollection and HasIsFinite], 
+  100, # rank above object-specific methods
     function ( C )
     if IsFinite( C ) then
         TryNextMethod();
@@ -3048,6 +3052,36 @@ end);
 
 InstallMethod( CanComputeIsSubset,"default: no, unless identical",
   [IsObject,IsObject],IsIdenticalObj);
+
+# This setter method is installed to implement filter settings in response
+# to an objects size as part of setting the size. This used to be handled
+# instead by immediate methods, but in a situation as here it would trigger
+# multiple immediate methods, several of which could apply and each changing
+# the type of the object. Doing so can be costly and thus should be
+# avoided.
+InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
+  100, # override system setter
+function(obj,sz)
+local filt;
+  if HasSize(obj) and Size(obj)<>sz then
+    if AssertionLevel()>2 then
+      # Make this an ordinary error (not ErrorNoReturn as suggested) to
+      # preserve all debugging options -- even use `return` to investigate
+      # what would have happened before this methods was introduced.
+      Error("size of ",obj," already set to ",Size(obj),
+        ", cannot be changed to ",sz);
+    fi;
+    return;
+  fi;
+  if sz=0 then filt:=IsEmpty and IsNonTrivial; # IsNonTrivial hold
+  elif sz=1 then filt:=IsTrivial;
+  elif sz=infinity then filt:=IsNonTrivial and HasIsFinite;
+  else filt:=IsNonTrivial and IsFinite;
+  fi;
+  filt:=filt and HasSize;
+  obj!.Size:=sz;
+  SetFilterObj(obj,filt);
+end);
 
 #############################################################################
 ##
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -411,18 +411,17 @@ end);
 InstallOtherMethod(DoubleCoset,"with size",true,
   [IsGroup,IsObject,IsGroup,IsPosInt],0,
 function(U,g,V,sz)
-local d,fam;
+local d,fam,typ;
   fam:=FamilyObj(U);
-  if not IsBound(fam!.doubleCosetsDefaultSizeType) then
-    fam!.doubleCosetsDefaultSizeType:=NewType(fam,IsDoubleCosetDefaultRep
-	  and HasSize and HasIsFinite and IsFinite
+  typ:=NewType(fam,IsDoubleCosetDefaultRep
+	  and HasIsFinite and IsFinite
           and HasLeftActingGroup and HasRightActingGroup
 	  and HasRepresentative);
-  fi;
   d:=rec();
-  ObjectifyWithAttributes(d,fam!.doubleCosetsDefaultSizeType,
-    LeftActingGroup,U,RightActingGroup,V,Representative,g,
-    Size,sz);
+  ObjectifyWithAttributes(d,typ,
+    LeftActingGroup,U,RightActingGroup,V,Representative,g);
+  SetSize(d,sz); # Size has private setter which will cause problems with
+  # HasSize triggering an immediate method.
   return d;
 end);
 
@@ -574,21 +573,25 @@ end);
 InstallMethod(RightCoset,"use subgroup size",IsCollsElms,
   [IsGroup and HasSize,IsObject],0,
 function(U,g)
-local d,fam;
+local d,fam,typ;
   # noch tests...
 
   fam:=FamilyObj(U);
-  if not IsBound(fam!.rightCosetsDefaultSizeType) then
-    fam!.rightCosetsDefaultSizeType:=NewType(fam,IsRightCosetDefaultRep and
-          HasActingDomain and HasFunctionAction and HasRepresentative and
-	  HasSize and HasCanonicalRepresentativeDeterminatorOfExternalSet);
-  fi;
+  typ:=NewType(fam,IsRightCosetDefaultRep and
+          HasActingDomain and HasFunctionAction and HasRepresentative
+	  and HasCanonicalRepresentativeDeterminatorOfExternalSet);
 
   d:=rec();
-  ObjectifyWithAttributes(d,fam!.rightCosetsDefaultSizeType,
+  ObjectifyWithAttributes(d,typ,
     ActingDomain,U,FunctionAction,OnLeftInverse,Representative,g,
-    Size,Size(U),CanonicalRepresentativeDeterminatorOfExternalSet,
+    CanonicalRepresentativeDeterminatorOfExternalSet,
     RightCosetCanonicalRepresentativeDeterminator);
+  # We cannot set the size in the previous ObjectifyWithAttributes as there is
+  # a custom setter method (the one added in this commit). In such a case
+  # ObjectifyWith Attributes just does `Objectify` and calls all setters
+  # separately which is what we want to avoid here.
+  SetSize(d,Size(U)); 
+
   return d;
 end);
 
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -592,6 +592,7 @@ DeclareProperty( "IsPerfectGroup", IsGroup );
 InstallFactorMaintenance( IsPerfectGroup,
     IsGroup and IsPerfectGroup, IsObject, IsGroup );
 
+InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
 
 #############################################################################
 ##
@@ -782,6 +783,8 @@ InstallFactorMaintenance( IsSolvableGroup,
 ##  we need the direct implication from supersolvability to solvability.
 InstallTrueMethod( IsSolvableGroup, IsMonomialGroup );
 InstallTrueMethod( IsSolvableGroup, IsSupersolvableGroup );
+
+InstallTrueMethod( HasIsPerfectGroup, IsGroup and IsSolvableGroup and IsNonTrivial );
 
 
 #############################################################################
@@ -3565,6 +3568,11 @@ DeclareOperation( "GroupByGenerators",
 DeclareOperation( "GroupWithGenerators", [ IsCollection ] );
 DeclareOperation( "GroupWithGenerators",
     [ IsCollection, IsMultiplicativeElementWithInverse ] );
+
+#F  MakeGroupyType( <fam>, <filt>, <gens>, <isgroup> )
+# type creator function to incorporate basic deductions so immediate methods
+# are not needed
+DeclareGlobalFunction("MakeGroupyType");
 
 
 #############################################################################
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -30,7 +30,9 @@ InstallImmediateMethod( IsFinitelyGeneratedGroup,
 ##
 #M  IsCyclic( <G> ) . . . . . . . . . . . . . . . . test if a group is cyclic
 ##
-InstallImmediateMethod( IsCyclic, IsGroup and HasGeneratorsOfGroup, 0,
+#  This used to be an immediate method. It was replaced by an ordinary
+#  method since the flag is typically set when creating the group.
+InstallMethod( IsCyclic, true, [IsGroup and HasGeneratorsOfGroup], 0,
     function( G )
     if Length( GeneratorsOfGroup( G ) ) = 1 then
       return true;
@@ -43,10 +45,14 @@ InstallMethod( IsCyclic,
     "generic method for groups",
     [ IsGroup ],
     function ( G )
+    local a;
 
     # if <G> has a generator list of length 1 then <G> is cyclic
     if HasGeneratorsOfGroup( G ) and Length( GeneratorsOfGroup(G) ) = 1 then
-      SetMinimalGeneratingSet(G,GeneratorsOfGroup(G));
+      a:=GeneratorsOfGroup(G)[1];
+      if CanEasilyCompareElements(a) and not IsOne(a) then
+        SetMinimalGeneratingSet(G,GeneratorsOfGroup(G));
+      fi;
       return true;
 
     # if <G> is not commutative it is certainly not cyclic
@@ -433,11 +439,6 @@ InstallMethod( IsNilpotentGroup,
 ##
 #M  IsPerfectGroup( <G> ) . . . . . . . . . . . .  test if a group is perfect
 ##
-InstallImmediateMethod( IsPerfectGroup,
-    IsSolvableGroup and HasIsTrivial,
-    0,
-    IsTrivial );
-
 InstallImmediateMethod( IsPerfectGroup,
     IsGroup and HasIsAbelian and IsSimpleGroup,
     0,
@@ -4283,6 +4284,39 @@ local s,b;
   return s;
 end );
 
+#F  MakeGroupyType( <fam>, <filt>, <gens>, <id>, <isgroup> )
+# type creator function to incorporate basic deductions so immediate methods
+# are not needed. Parameters are family, filter to start with, generator
+# list, is it indeed a group (or only magma)?
+InstallGlobalFunction(MakeGroupyType,
+function(fam,filt,gens,id,isgroup)
+
+  if IsFinite(gens) then
+    if isgroup then
+      filt:=filt and IsFinitelyGeneratedGroup;
+    fi;
+
+    if Length(gens)>0 and CanEasilyCompareElements(gens) then
+      if id=false then
+        id:=One(gens[1]);
+      fi;
+      if id<>fail then # cannot do identity in magma
+        if ForAny(gens,x->x<>id) then
+          filt:=filt and HasIsTrivial and IsNonTrivial;
+          if isgroup and Length(gens)<=1 then # cyclic not for magmas
+            filt:=filt and IsCyclic;
+          fi;
+        else
+          filt:=filt and IsTrivial and HasIsNonTrivial;
+        fi;
+      fi;
+    elif isgroup and Length(gens)<=1 then # cyclic not for magmas
+      filt:=filt and IsCyclic;
+    fi;
+  fi;
+  return NewType(fam,filt);
+end);
+
 #############################################################################
 ##
 #M  GroupWithGenerators( <gens> ) . . . . . . . . group with given generators
@@ -4291,86 +4325,57 @@ end );
 InstallMethod( GroupWithGenerators,
     "generic method for collection",
     [ IsCollection ],
-    function( gens )
-    local G,fam,typ;
+function( gens )
+local G,typ;
 
-    fam:=FamilyObj(gens);
-    if IsFinite(gens) then
-      if not IsBound(fam!.defaultFinitelyGeneratedGroupType) then
-        fam!.defaultFinitelyGeneratedGroupType:=
-          NewType(fam,IsGroup and IsAttributeStoringRep
-                      and HasGeneratorsOfMagmaWithInverses
-                      and IsFinitelyGeneratedGroup);
-      fi;
-      typ:=fam!.defaultFinitelyGeneratedGroupType;
-    else
-      if not IsBound(fam!.defaultGroupType) then
-        fam!.defaultGroupType:=
-          NewType(fam,IsGroup and IsAttributeStoringRep
-                      and HasGeneratorsOfMagmaWithInverses);
-      fi;
-      typ:=fam!.defaultGroupType;
-    fi;
+  typ:=MakeGroupyType(FamilyObj(gens),
+          IsGroup and IsAttributeStoringRep 
+          and HasIsEmpty and HasGeneratorsOfMagmaWithInverses,
+          gens,false,true);
 
-    G:=rec();
-    ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
+  G:=rec();
+  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
 
-    return G;
-    end );
+  return G;
+end );
 
 InstallMethod( GroupWithGenerators,
     "generic method for collection and identity element",
     IsCollsElms, [ IsCollection, IsMultiplicativeElementWithInverse ],
-    function( gens, id )
-    local G,fam,typ;
+function( gens, id )
+local G,typ;
 
-    fam:=FamilyObj(gens);
-    if IsFinite(gens) then
-      if not IsBound(fam!.defaultFinitelyGeneratedGroupWithOneType) then
-        fam!.defaultFinitelyGeneratedGroupWithOneType:=
-          NewType(fam,IsGroup and IsAttributeStoringRep
-                      and HasGeneratorsOfMagmaWithInverses
-                      and IsFinitelyGeneratedGroup and HasOne);
-      fi;
-      typ:=fam!.defaultFinitelyGeneratedGroupWithOneType;
-    else
-      if not IsBound(fam!.defaultGroupWithOneType) then
-        fam!.defaultGroupWithOneType:=
-          NewType(fam,IsGroup and IsAttributeStoringRep
-                      and HasGeneratorsOfMagmaWithInverses and HasOne);
-      fi;
-      typ:=fam!.defaultGroupWithOneType;
-    fi;
+  typ:=MakeGroupyType(FamilyObj(gens),
+          IsGroup and IsAttributeStoringRep 
+            and HasIsEmpty and HasGeneratorsOfMagmaWithInverses and HasOne,
+            gens,id,true);
 
-    G:=rec();
-    ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
-                            One,id);
+  G:=rec();
+  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
+                          One,id);
 
-    return G;
+  return G;
 end );
 
-InstallMethod( GroupWithGenerators,
-    "method for empty list and element",
-    [ IsList and IsEmpty, IsMultiplicativeElementWithInverse ],
-    function( empty, id )
-    local G,fam,typ;
+InstallMethod( GroupWithGenerators,"method for empty list and element",
+  [ IsList and IsEmpty, IsMultiplicativeElementWithInverse ],
+  function( empty, id )
+local G,fam,typ;
 
-    fam:= CollectionsFamily( FamilyObj( id ) );
-    if not IsBound( fam!.defaultFinitelyGeneratedGroupWithOneType ) then
-      fam!.defaultFinitelyGeneratedGroupWithOneType:=
-        NewType( fam, IsGroup and IsAttributeStoringRep
-                      and HasGeneratorsOfMagmaWithInverses
-                      and IsFinitelyGeneratedGroup and HasOne );
-    fi;
-    typ:= fam!.defaultFinitelyGeneratedGroupWithOneType;
+  fam:= CollectionsFamily( FamilyObj( id ) );
 
-    G:= rec();
-    ObjectifyWithAttributes( G, typ,
-                             GeneratorsOfMagmaWithInverses, empty,
-                             One, id );
+  typ:=IsGroup and IsAttributeStoringRep
+        and HasGeneratorsOfMagmaWithInverses and HasOne and IsTrivial and
+        HasIsEmpty and HasIsNonTrivial;
+  typ:=NewType(fam,typ);
 
-    return G;
-    end );
+  G:= rec();
+  ObjectifyWithAttributes( G, typ,
+                            GeneratorsOfMagmaWithInverses, empty,
+                            One, id );
+
+  return G;
+end );
 
 
 #############################################################################
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -955,6 +955,11 @@ local iso,hom,u;
   if HasIsAbelian(G) and IsAbelian(G) then
     return TrivialSubgroup(G);
   elif Size(Image(hom))=infinity then
+    # test a special case -- one generator
+    if Length(GeneratorsOfGroup(G))=1 then
+      SetIsAbelian(G,true);
+      return TrivialSubgroup(G);
+    fi;
     Error("Derived subgroup has infinite index, cannot represent");
   elif Size(Range(hom))=1 then
     return G; # this is needed because the trivial quotient is represented
@@ -1539,7 +1544,7 @@ end );
 #M  FactorFreeGroupByRelators(<F>,<rels>) .  factor of free group by relators
 ##
 BindGlobal( "FactorFreeGroupByRelators", function( F, rels )
-    local G, fam, gens;
+    local G, fam, gens,typ;
 
     # Create a new family.
     fam := NewFamily( "FamilyElementsFpGroup", IsElementOfFpGroup );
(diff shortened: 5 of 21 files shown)


lib/coll.gi
```diff
@@ -169,14 +169,17 @@ InstallMethod( IsWholeFamily,
...
+# not know its size -- e.g. permutation, pc
+# the benefit OTOH is minimal beyond showing off a feature.
+#InstallImmediateMethod( Size,
+#    IsCollection and HasIsFinite and IsAttributeStoringRep, 0,
+#    function ( C )
+#    if IsFinite( C ) then
+#        TryNextMethod();
+#    fi;
+#    return infinity;
+#    end );
```
fingolfin: I agree this should not be an immediate method, but perhaps it should be retained as a regular method? Then with the `IsAttributeStoringRep` removed.

lib/grp.gi
```diff
@@ -30,14 +30,15 @@ InstallImmediateMethod( IsFinitelyGeneratedGroup,
...
-    end );
+# this is now set when creating the groups
+#InstallImmediateMethod( IsCyclic, IsGroup and HasGeneratorsOfGroup, 0,
+#    function( G )
+#    if Length( GeneratorsOfGroup( G ) ) = 1 then
+#      return true;
+#    else
+#      TryNextMethod();
+#    fi;
+#    end );
```
fingolfin: Fine by me.

Ooops, the next method after this one is incorrect: it does
```
if HasGeneratorsOfGroup( G ) and Length( GeneratorsOfGroup(G) ) = 1 then
  SetMinimalGeneratingSet(G,GeneratorsOfGroup(G));
```
which is incorrect if `GeneratorsOfGroup(G)` consists only of an identity element. This is of course currently hidden by the immediate method. Mental note to self: try to find a tests case triggering this (after the immediate is gone), and fix it. (But anybody else feel free to fix this first, I won't work on it today for sure ;-)

lib/grp.gi
```diff
@@ -4291,86 +4294,121 @@ end );
...
+    if Length(gens)>0 and CanEasilyCompareElements(gens) then
+      id:=One(gens[1]);
+      if ForAny(gens,x->x<>id) then
+        typ:=typ and HasIsTrivial and HasIsNonTrivial;
+        triv:=false;
+      else
+        typ:=typ and HasIsTrivial and HasIsNonTrivial;
+        triv:=true;
+        # basic consequence
+        typ:=typ and HasIsPerfectGroup and IsPerfectGroup;
```
fingolfin: I'd rather have a `InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );`, this will cover the above (with zero overhead).

lib/grp.gi
```diff
@@ -4291,86 +4294,121 @@ end );
...
-    return G;
-    end );
+  G:=rec();
+  if triv=fail then
+    ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
+                            IsEmpty,false);
+  else
+    ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
+                            IsEmpty,false,IsTrivial,triv,
+                            IsNonTrivial,not triv);
```
fingolfin: Hmm, why set `IsTrivial` resp. `IsNonTrivial` this way, instead of just anding them into `typ` like you do with e.g. `IsCyclic`? Does this have some functional difference?

If not, I'd just set this all in `typ`, then you can get rid of `triv` and simplify the code here and elsewhere accordingly.

lib/grp.gi
```diff
@@ -4291,86 +4294,121 @@ end );
...
+        # basic consequence
+        typ:=typ and HasIsPerfectGroup and IsPerfectGroup;
       fi;
-      typ:=fam!.defaultGroupType;
     fi;
 
-    G:=rec();
-    ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
+  fi;
+  typ:=NewType(fam,typ);
```
fingolfin: Just to make it explicit: the changes in this PR get rid of the cached values `fam!.defaultFinitelyGeneratedGroupType` and `fam!.defaultGroupType`. I assume that you verified that there is a net gain (probably because this initial type (almost?) always ended up being replaced by immediate methods anyway.

lib/grp.gi
```diff
@@ -4291,86 +4294,121 @@ end );
...
+  fam:=FamilyObj(gens);
+  triv:=fail; # might not find out
+
+  typ:=IsGroup and IsAttributeStoringRep 
+        and HasIsEmpty and HasGeneratorsOfMagmaWithInverses and HasOne;
+  if IsFinite(gens) then
+    typ:=typ and IsFinitelyGeneratedGroup;
+    if Length(gens)<=1 then
+      typ:=typ and HasIsCyclic and IsCyclic 
+          and HasIsCommutative and IsCommutative;
```
fingolfin: Note that this is equivalent to `typ := typ and IsCyclic`, as `IsCyclic` implies `IsCommutative`, and each implies its tester, and handling these implications comes at zero cost here.

lib/grp.gi
```diff
@@ -4291,86 +4294,121 @@ end );
...
 
-    G:= rec();
-    ObjectifyWithAttributes( G, typ,
-                             GeneratorsOfMagmaWithInverses, empty,
-                             One, id );
+  typ:=IsGroup and IsAttributeStoringRep 
+        and HasGeneratorsOfMagmaWithInverses and HasOne
+        and IsFinitelyGeneratedGroup and HasIsTrivial and HasIsNonTrivial
+        and HasIsFinite and HasIsCyclic
+        and HasIsPerfectGroup and HasIsEmpty;
```
fingolfin: Can't you just replace the above with `typ:= IsGroup and IsAttributeStoringRep and HasGeneratorsOfMagmaWithInverses and HasOne and IsTrivial` ? The `IsTrivial` implies (via zero cost implications) all the remaining flags.

lib/grppc.gi
```diff
@@ -272,7 +272,8 @@ InstallMethod( Pcgs,
 ##
 #M  GeneralizedPcgs( <G> )
 ##
-InstallImmediateMethod( GeneralizedPcgs, IsGroup and HasPcgs, 0, Pcgs );
+# is now set when creating
+#InstallImmediateMethod( GeneralizedPcgs, IsGroup and HasPcgs, 0, Pcgs );
```
fingolfin: Well, it is set when creating via `GroupWithGenerators`, but that's not the only way to create a group object! However, I agree this does not need to be an immediate method; esp. as nothing tests `HasGroupWithGenerators`. So let's just turn it into a regular method.

lib/magma.gi
```diff
@@ -184,16 +184,17 @@ InstallImmediateMethod( IsCommutative,
...
-    IsMagmaWithInverses and IsAssociative
-                        and HasGeneratorsOfMagmaWithInverses, 0,
-    function( M )
-    if Length( GeneratorsOfMagmaWithInverses( M ) ) = 1 then
-      return true;
-    else
-      TryNextMethod();
-    fi;
-    end );
+# this is now set when creating groups
```
fingolfin: This should say "when creating magmas". But even then it's not quite true: If a magma is created in a non-standard way, it may not be set. So, like with most of the other immediate methods you remove/commented out, I'd prefer to instead turn it into a regular method.

lib/magma.gi
```diff
@@ -652,17 +653,44 @@ InstallOtherMethod( MagmaWithOneByGenerators,
...
+
+  typ:=IsMagmaWithInverses and IsAttributeStoringRep 
+              and HasGeneratorsOfMagmaWithInverses
+              and HasIsEmpty;
+
+  if IsFinite(gens) then
+    # typ:=typ and IsFinitelyGeneratedGroup; don't know whether group
+    if Length(gens)<=1 then
+      typ:=typ and HasIsCyclic and IsCyclic 
+          and HasIsCommutative and IsCommutative;
```
fingolfin: This deduction is not correct for non-associative magmas.

fingolfin (changes requested):

lib/coll.gi
hulpke: If we do so, we need to rank this high, so that it overrides any other.

(Also removing this immediate method broke specific tests for this method will need to be adjusted.)

lib/grp.gi
hulpke: When I left the caching in I ran into problems -- a non-nilpotent group claimed to be nilpotent. The best I could come up (#2394 ) is that the cached type in fact gets changed later and thus cannot be cached. If I'm doing something wrong here I'd love to know (and then would add type caching again).

lib/grp.gi
hulpke: Ah -- I wasn't sure. Will fix.

lib/grp.gi
fingolfin: I am not sure how useful those caches would be, unless done very carefully: in the old code, there were already two caches, one for finitely generated groups, and one for those where we don't know whether they are f.g. or not. In order to preset `IsTrivial` and other flags, you'd have to either create more caches (i.e. have one cached type for trivial groups; one for non-trivial cyclic; one for non-cyclic f.g.; one for the generic case). Of course that number grows as you add special cases. Luckily, in this case at least the number of types does not seem to grow exponentially, as most of the checks differentiate based on the generating set along.

fingolfin: At the start of writing the previous comment, I wasn't sure, because I thought you'd need to add 2^n caches, for n 3 or 4, but now it seems only four may be needed, which is at least not totally crazy.

lib/grp.gi
fingolfin: Every property implies it tester; for the rest, you can check this e.g. as follows :
```
gap> ShowImpliedFilters(IsCyclic);


May imply with:
+IsMagmaWithInverses
+IsAssociative
   IsCommutative
   IsNilpotentGroup
   IsSupersolvableGroup
   IsSolvableGroup
   IsNilpotentByFinite

+IsOrdinaryTable
   IsCommutative
   IsNilpotentCharacterTable
   IsSolvableCharacterTable
   IsSupersolvableCharacterTable
```

Or, more technically:
```
gap> IsEquivFilters:={f1,f2}->IS_EQUAL_FLAGS(WITH_IMPS_FLAGS(FLAGS_FILTER(f1)), WITH_IMPS_FLAGS(FLAGS_FILTER(f2)));
function( f1, f2 ) ... end
gap> IsEquivFilters(IsGroup and IsCyclic, HasIsCyclic and IsCyclic and HasIsCommutative and IsCommutative and IsGroup);
true
```

lib/grp.gi
fingolfin: Another cause for the problems you have seen may be that you add e.g. `HasIsTrivial` to the flag lists of the type. I think that means that any object created with that type is *non-trivial*, no matter what; once `HasIsTrivial` is set, one cannot change the value of `IsTrivial` anymore (at least not without low-level bit fiddling); this is a feature, not a bug.

lib/grp.gi
hulpke: I'd love to do. But how do I set  a flag to `false' as part of a type?

IsTrivial and IsNonTrivial are twins that are set the opposite way and currently immediate methods are used to set either to false if the other is set.

lib/grp.gi
```diff
@@ -4291,86 +4300,145 @@ end );
...
+
+  if IsFinite(gens) then
+    typ:=typ and IsFinitelyGeneratedGroup;
+    typename:="defaultFinitelyGeneratedGroupType";
+
+    if Length(gens)>0 and CanEasilyCompareElements(gens) then
+      id:=One(gens[1]);
+      if ForAny(gens,x->x<>id) then
+        typ:=typ and HasIsTrivial and HasIsNonTrivial;
+        triv:=false;
```
fingolfin: Hmm, why does this still use `triv` instead of setting `typ:=typ and HasIsTrivial and IsNonTrivial`; similar below

lib/grp.gi
```diff
@@ -4291,86 +4300,145 @@ end );
...
+        if Length(gens)<=1 then
+          typ:=typ and IsCyclic;
+          typename:="defaultFinitelyGeneratedNontrivialCyclicGroupType";
+        else
+          typename:="defaultFinitelyGeneratedNontrivialGroupType";
+        fi;
+      else
+        typ:=typ and HasIsTrivial and HasIsNonTrivial;
+        triv:=true;
+        typename:="defaultFinitelyGeneratedTrivialGroupType";
```
fingolfin: The FG part of this name seems redundant: trivial already implies iy, no? Same for the cyclic types.

lib/grp.gi
hulpke: Thanks! Will do so.

lib/grp.gi
```diff
@@ -4291,86 +4300,145 @@ end );
...
+        fi;
+      else
+        typ:=typ and HasIsTrivial and HasIsNonTrivial;
+        triv:=true;
+        typename:="defaultFinitelyGeneratedTrivialGroupWithOneType";
       fi;
-      typ:=fam!.defaultGroupWithOneType;
+    elif Length(gens)<=1 then
+      typ:=typ and IsCyclic;
+      typename:="defaultFinitelyGeneratedCyclicGroupWithOneType";
```
fingolfin: This type computation is repeated several times across multiple GroupWithGenerators methods - can't it perhaps be factored out into a common helper function?

lib/grp.gi
```diff
@@ -433,10 +439,13 @@ InstallMethod( IsNilpotentGroup,
...
-    IsSolvableGroup and HasIsTrivial,
-    0,
-    IsTrivial );
+# Replace by TrueMethod (though this will not make nontrivial solvable
+# groups not perfect
+#InstallImmediateMethod( IsPerfectGroup,
+#    IsSolvableGroup and HasIsTrivial,
+#    0,
+#    IsTrivial );
+InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
```
fingolfin: InstallTrueMethod calls belong into .gd files

lib/coll.gi
```diff
@@ -169,8 +169,11 @@ InstallMethod( IsWholeFamily,
 ##
 #M  Size( <C> ) . . . . . . . . . . . . . . . . . . . .  size of a collection
 ##
-InstallImmediateMethod( Size,
-    IsCollection and HasIsFinite and IsAttributeStoringRep, 0,
+#  replace immediate method by ordinary, as immediate gets called for every
+#  group that knows it is finite but does not know its size -- e.g.
+#  permutation, pc.  The benefit of this is minimal beyond showing off a
+#  feature.
```
fingolfin: This comment (as well as several other that follow) only make full sense when read in the context of this commit resp. PR; but if somebody reads this in a year, they might have a hard time understanding what is meant here. "What is being replaced here???"

Not having this commit would IMHO be better; or start it with something like "the following method used to be immediate. However..." but I wonder whether this would be a useful comment. I think this would better fit into the commit message.

lib/coll.gi
```diff
@@ -169,8 +169,11 @@ InstallMethod( IsWholeFamily,
...
+InstallMethod( Size,true, [IsCollection and HasIsFinite], 0,
```
fingolfin: You can omit true and 0 here

lib/grp.gi
```diff
@@ -30,7 +30,9 @@ InstallImmediateMethod( IsFinitelyGeneratedGroup,
 ##
 #M  IsCyclic( <G> ) . . . . . . . . . . . . . . . . test if a group is cyclic
 ##
-InstallImmediateMethod( IsCyclic, IsGroup and HasGeneratorsOfGroup, 0,
+# this is now set when creating the groups, make ordinary instead of
```
fingolfin: See above

lib/grp.gi
```diff
@@ -4291,86 +4300,125 @@ end );
...
+        typ:=typ and HasIsTrivial and IsNonTrivial;
+        if Length(gens)<=1 then
+          typ:=typ and IsCyclic;
+          typename:="defaultFinitelyGeneratedNontrivialCyclicGroupType";
+        else
+          typename:="defaultFinitelyGeneratedNontrivialGroupType";
+        fi;
+      else
+        typ:=typ and IsTrivial and HasIsNonTrivial;
+        typename:="defaultFinitelyGeneratedTrivialGroupType";
```
fingolfin: The FG part of this name seems redundant: trivial already implies iy, no? Same for the cyclic types.

lib/grppcfp.gi
```diff
@@ -69,6 +69,7 @@ InstallGlobalFunction( IsomorphismFpGroupByPcgs, function( pcgs, str )
         od;
     od;
     H := F / rels;
+    SetIsFinite(H,true);
```
fingolfin: I find it rather awkward to add all this `SetIsFinite` calls before each `SetSize`; and it is easy to forget, too. I'd rather add this into a custome setter for `Size`. I can submit a PR doing that tomorrow. Such a setter could also set IsTrivial, IsEmpty, IsNontrivial in one go.

Also, I assume this is only beneficial because setting `IsFinite` does not trigger any/fewer immediate methods than setting `Size`, correct?

hulpke: Yes to both. The setter also is on my to-do list.

lib/coll.gi
hulpke: OK, but is it a problem if they're in? (Some methods actually might need a higher ranking).

lib/grp.gi
hulpke: The best I was able to come up with would have been preprocessor macros, but for a function there are too many small differences that would require input parameter and extra processing.  (This is an observation of my incompetence, not a statement that it cannot be done.)

lib/coll.gi
hulpke: I'll change the text. IF it it hidden in a commit message, however I think it is less likely to be seen and also loses the reference to the place in the source code.

lib/grppcfp.gi
hulpke: Also, the IsFinite might be overkill -- I was trying to get rid of calls to immediate methods.

lib/grp.gi
hulpke: Done.

lib/grp.gd
```diff
@@ -592,6 +592,7 @@ DeclareProperty( "IsPerfectGroup", IsGroup );
 InstallFactorMaintenance( IsPerfectGroup,
     IsGroup and IsPerfectGroup, IsObject, IsGroup );
 
+InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
```
fingolfin: In this case, you could also compensate for the now missing "other" direction, via
`InstallTrueMethod( HasIsPerfectGroup, IsGroup and IsSolvableGroup and IsNonTrivial );`

lib/grp.gi
```diff
@@ -433,10 +439,12 @@ InstallMethod( IsNilpotentGroup,
...
-InstallImmediateMethod( IsPerfectGroup,
-    IsSolvableGroup and HasIsTrivial,
-    0,
-    IsTrivial );
+#  This used to be an immediate method. It was replaced by a TrueMethod (in
+#  the gd file)
+#InstallImmediateMethod( IsPerfectGroup,
+#    IsSolvableGroup and HasIsTrivial,
+#    0,
+#    IsTrivial );
```
fingolfin: Let's just remove this completely -- see also my suggestion above for the other "missing" direction.

lib/grp.gi
```diff
@@ -4291,86 +4299,125 @@ end );
...
+
+  typ:=IsGroup and IsAttributeStoringRep 
+        and HasIsEmpty and HasGeneratorsOfMagmaWithInverses and HasOne;
+  typename:="defaultGroupWithOneType";
+
+  if IsFinite(gens) then
+    typ:=typ and IsFinitelyGeneratedGroup;
+    typename:="defaultFinitelyGeneratedGroupWithOneType";
+
+    if Length(gens)>0 and CanEasilyCompareElements(gens) then
```
fingolfin: Use `>=0` to deal with empty collections? Or even better, as this also works if `CanEasilyCompareElements(gens)` is false, start like this:
```gap
if Length(gens) = 0 then
 ...
elif CanEasilyCompareElements(gens) then
 ...
elif Length(gens) = 1 then
 ...
fi;
```

lib/grp.gi
```diff
@@ -4291,86 +4299,125 @@ end );
...
+      if ForAny(gens,x->x<>id) then
+        typ:=typ and HasIsTrivial and IsNonTrivial;
+        if Length(gens)<=1 then
+          typ:=typ and IsCyclic;
+          typename:="defaultNontrivialCyclicGroupWithOneType";
```
fingolfin: BTW, `NEW_TYPE` *also* performs caching of types for a given family. You can inspect `NEW_TYPE_CACHE_MISS` and `NEW_TYPE_CACHE_HIT` before/after creating a group to see it was effective. So.... that might in fact render the cache here moot. (While the explicit cache here may be a tiny bit faster, it also adds a lot of complexity...)

tst/testinstall/coll.tst
```diff
@@ -97,15 +97,16 @@ Error, cannot test whether <C> contains the family of its elements
...
 #
 
-# immediate method for collections knowing they are infinite
-gap> c2:=ConjugacyClass(F, F.1);;
-gap> HasSize(c2);
-false
-gap> SetIsFinite(c2, false);
-gap> HasSize(c2);
-true
-gap> Size(c2);
```
fingolfin: This test foremost was meant to test that `Size` method. Since it is still there, just not immediate, we could keep the test, and just remove the two checks for `HasSize`.

lib/grp.gd
hulpke: OK

lib/grp.gi
hulpke: ditto OK

tst/testinstall/coll.tst
hulpke: OK.

lib/grp.gi
hulpke: But there is a specialized method below for the case of `IsEmpty` for the collection `gens`. Wouldn't it override it?

lib/grp.gi
hulpke: I'll put this in a separate remark as a push will make it go outdated in code

hulpke:
> BTW, NEW_TYPE also performs caching of types for a given family. You can inspect  NEW_TYPE_CACHE_MISS and NEW_TYPE_CACHE_HIT before/after creating a group to see it was effective. So.... that might in fact render the cache here moot. (While the explicit cache here may be a tiny bit faster, it also adds a lot of complexity...)

lib/grp.gi
hulpke: (after all it did not, but should not be only commit remark)

lib/grp.gi
```diff
@@ -30,7 +30,9 @@ InstallImmediateMethod( IsFinitelyGeneratedGroup,
 ##
 #M  IsCyclic( <G> ) . . . . . . . . . . . . . . . . test if a group is cyclic
 ##
-InstallImmediateMethod( IsCyclic, IsGroup and HasGeneratorsOfGroup, 0,
+#  This used to be an immediate method. It was replaced by an ordinary
+#  method sine the flag is typically set when creating the group.
```
fingolfin: sine -> since

lib/grppcaut.gi
```diff
@@ -971,6 +975,7 @@ local f,act,xset,stb,hom,n,m,l,i;
     Add(stb,m);
     stb:=List(stb,x->x^act); # base change
     stb:=Group(stb);
+    SetIsFinite(stb,true);
```
fingolfin: Are all these `SetIsFinite` calls still necessary?

hulpke: No. SetSize (or SetIsGroupOfAutomorphismsFiniteGroup) should take care of it.

lib/coll.gi
```diff
@@ -3049,6 +3053,20 @@ end);
...
+# avoid immediate methods triggering multiple type changes once the Size of
+# an object is known.
+InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
+  100, # override system setter
+function(obj,sz)
+  if not HasSize(obj) then
+    SetIsEmpty(obj,sz=0);
+    SetIsTrivial(obj,sz=1);
+    SetIsNonTrivial(obj,sz<>1);
+    SetIsFinite(obj,sz<>infinity);
```
fingolfin: Doesn't this still change the type four times (less for sz=0 and sz=1 because of implications, but stilll).  Plus a fifth time for the size. So, how about something like this:
```gap
function(obj, size)
local filt;
  if HasSize(obj) and Size(obj) <> size then
    ErrorNoReturn("size of <obj>" already set to ", Size(obj), ", cannot change it to ", size);
  fi;
  if size = 0 then filt := IsEmpty;
  elif size = 1 then filt := IsTrivial;
  elif size = infinity then filt := IsNonTrivial and HasIsFinite;
  else filt := IsNonTrivial and IsFinite;
  fi;
  filt := filt and HasSize;
  obj!.Size := size;
  SetFilterObj(obj, filt);
end;
```

This assumes we already have or add (negative) implications between IsEmpty, IsTrivial and IsNonTrivial (the three are after all mutually exclusive).

(We may also wish to add `IsNonEmpty` and `IsInfinite`, but better not in this PR)

lib/grp.gi
fingolfin: That method is only for empty lists, but not for empty collections-that-aren't-lists. Though these are rather rare and special beast.

But as I said before, it would also be nice if all this logic would not have to be repeated in almost identical form several times, but rather was put into one (or at least very few) helper functions which can be called from multiple places.

hulpke: The deductions logic is now in a separate function.

lib/coll.gi
hulpke: This is done. However (the manual specifies that an `SetSize` if size is given has no effect) I have wrapped the error into a check that requires assertion level to be >=3.

lib/coll.gi
```diff
@@ -3052,6 +3053,31 @@ end);
...
+# avoid immediate methods triggering multiple type changes once the Size of
+# an object is known.
+InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
+  100, # override system setter
+function(obj,sz)
+local filt;
+  if HasSize(obj) and Size(obj)<>sz then
+    if AssertionLevel()>2 then
+      # Make this an ordinary error to enter break loop so that one can
+      # investigate call order for debugging
```
fingolfin: I don't understand this comment. What is a "non-ordinary error"? Perhaps you are referring to `Assert`? But that enters a break loop for me just fine, doesn't it?

lib/coll.gi
```diff
@@ -3052,6 +3053,31 @@ end);
 InstallMethod( CanComputeIsSubset,"default: no, unless identical",
   [IsObject,IsObject],IsIdenticalObj);
 
+# avoid immediate methods triggering multiple type changes once the Size of
+# an object is known.
```
fingolfin: This comment starts mid-thought. Perhaps: "Custom SetSize method which sets various properties implied by the given size. This avoids ..." ?

lib/csetgrp.gi
```diff
@@ -573,21 +573,20 @@ end);
...
+	  and HasCanonicalRepresentativeDeterminatorOfExternalSet);
 
   d:=rec();
-  ObjectifyWithAttributes(d,fam!.rightCosetsDefaultSizeType,
+  ObjectifyWithAttributes(d,typ,
     ActingDomain,U,FunctionAction,OnLeftInverse,Representative,g,
-    Size,Size(U),CanonicalRepresentativeDeterminatorOfExternalSet,
+    CanonicalRepresentativeDeterminatorOfExternalSet,
     RightCosetCanonicalRepresentativeDeterminator);
+  SetSize(d,Size(U)); # as own setter
```
fingolfin: What does "as own setter" mean?

Also, doesn't `ObjectifyWithAttributes` call custom setters anyway? It contains specific code to detect them. But perhaps you want to avoid that? In that case, I guess a comment explaining that (and why) would be indeed appropriate, but this comment isn't doing that.

lib/grp.gd
```diff
@@ -784,6 +784,8 @@ InstallFactorMaintenance( IsSolvableGroup,
 InstallTrueMethod( IsSolvableGroup, IsMonomialGroup );
 InstallTrueMethod( IsSolvableGroup, IsSupersolvableGroup );
 
+InstallTrueMethod( HasIsPerfectGroup, IsGroup and IsSolvableGroup and IsNonTrivial );
```
fingolfin: Why is this in the third commit, which is about changing "testinstall" according to its summary? Something seems to be off with how these commits are split.

lib/grp.gi
```diff
@@ -31,7 +31,7 @@ InstallImmediateMethod( IsFinitelyGeneratedGroup,
 #M  IsCyclic( <G> ) . . . . . . . . . . . . . . . . test if a group is cyclic
 ##
 #  This used to be an immediate method. It was replaced by an ordinary
-#  method sine the flag is typically set when creating the group.
+#  method since the flag is typically set when creating the group.
```
fingolfin: This fixes a typo from the first commit. It should be instead melded into the first commit...

fingolfin (commented):
Only some quick remarks, I have not yet looked at everything, need to prepare class material etc. now instead.

However, I am rather unhappy with the commit structure. At the very least, the commit messages need work. They reference work which is not visible anymore, which is highly confusing. They also contain typos (e.g. `Empty` instead of `IsEmpty`), and incomplete sentences, reducing their usefulness.

However, it also happens that one commit introduces a typo and then a later one fixes it (instead of just fixing the first commit); or that a commit contains changed not indicated by its summary line (which then turns out to not actually summarize what the commit is about).

I know it's tiresome to work so long on such a big patch, but it's also very tiresome if one has to debug these changes in a year down the road, and has to painstakingly reverse engineer what the changes were about, etc., so doing this right now will save us work down the road.

In this case, it might actually be best to first squash all changes together into one commit, or perhaps two (using `git reset` to uncommit all changes, followed by using `git add -p` judiciously to create fresh commits): one which turns immediates into implications; and one with the rest. I haven't had a chance to read through everything to say for sure, I'll try to do that during this week.

lib/coll.gi
hulpke: You had suggested `ErrorNoReturn` in an earlier remark on this function. For debugging (and understanding how code worked in the past despite this issue) it seemed better to not forbid `return;`.  (Yes, it could mislead a user, but this is really an error that should never come up for a user. (I'm happy to replace by `Assert`, if that is preferred here.

lib/grp.gd
hulpke: See remark below about making all a single commit.

lib/csetgrp.gi
hulpke: We cannot set the size in the previous `ObjectifyWithAttributes` as there is a custom setter method (the one just added here). In such a case `ObjectifyWith Attributes just does `Objectify` and calls all setters separately which is what we want to avoid here.

lib/grp.gi
hulpke: ditto

lib/coll.gi
hulpke: I did not see this as mid-thought, but I'm happy to change it.

hulpke:
@fingolfin 
I understand that all of us have other work to do. (I will refrain on offering unsolicited opinion about the enforced review process.)

Yes, my plan is to merge everything into a single commit (and then revise the comments) before merging. (Some of this has been done now.)
The reason for doing so (instead of multiple single-topic commits) is that subsets of the commits do not really make sense once one starts dealing with the immediate methods in one context.

So far I have done partial merges but left  some recent commits separately so it was clear what has changed.

lib/csetgrp.gi
```diff
@@ -574,21 +573,25 @@ end);
...
   d:=rec();
-  ObjectifyWithAttributes(d,fam!.rightCosetsDefaultSizeType,
+  ObjectifyWithAttributes(d,typ,
     ActingDomain,U,FunctionAction,OnLeftInverse,Representative,g,
-    Size,Size(U),CanonicalRepresentativeDeterminatorOfExternalSet,
+    CanonicalRepresentativeDeterminatorOfExternalSet,
     RightCosetCanonicalRepresentativeDeterminator);
+  # We cannot set the size in the previous ObjectifyWithAttributes as there is
+  # a custom setter method (the one added in this commit). In such a case
+  # ObjectifyWith Attributes just does `Objectify` and calls all setters
```
fingolfin: "ObjectifyWith Attributes" -> "ObjectifyWithAttributes" (one word)

lib/grp.gi
```diff
@@ -433,11 +439,6 @@ InstallMethod( IsNilpotentGroup,
 ##
 #M  IsPerfectGroup( <G> ) . . . . . . . . . . . .  test if a group is perfect
 ##
-InstallImmediateMethod( IsPerfectGroup,
-    IsSolvableGroup and HasIsTrivial,
-    0,
-    IsTrivial );
-
```
fingolfin: Why remove this, instead of turning this into a plain method? Wouldn't that still be useful?

fingolfin: (I don't think it's necessary to add a "this used to be an immediate method ..." comment)

lib/grp.gi
```diff
@@ -4283,6 +4284,39 @@ local s,b;
...
+        if ForAny(gens,x->x<>id) then
+          filt:=filt and HasIsTrivial and IsNonTrivial;
+          if isgroup and Length(gens)<=1 then # cyclic not for magmas
+            filt:=filt and IsCyclic;
+          fi;
+        else
+          filt:=filt and IsTrivial and HasIsNonTrivial;
+        fi;
+      fi;
+    elif isgroup and Length(gens)<=1 then # cyclic not for magmas
```
fingolfin: You could replace this `elsif` by `fi;  if ` and then get rid of the identical check a few lines above.

lib/grpfp.gi
```diff
@@ -955,6 +955,11 @@ local iso,hom,u;
   if HasIsAbelian(G) and IsAbelian(G) then
     return TrivialSubgroup(G);
   elif Size(Image(hom))=infinity then
+    # test a special case -- one generator
+    if Length(GeneratorsOfGroup(G))=1 then
+      SetIsAbelian(G,true);
+      return TrivialSubgroup(G);
+    fi;
```
fingolfin: So this seems to be a separate improvement for `DerivedSubgroup` on fp subgroups of rank 1... But why is it needed? Because we removed the immediate method for "size of generating set <= 1 implies abelian" ? 
This raises two questions:

1. Where is the place we create a group w/o taking care of that? can we do something about it (i.e. fix the issue at its root, not the symptom)
2. Do we really need to *remove* (or replace by normal ones) those immediate methods? After all, if e.g. `IsFinite` is set, then no immediate method for `IsFinite` is run. So leaving those immediate methods in should not cause a speed penalty anymore -- or does it? I should run some tests....

lib/grppc.gi
```diff
@@ -418,29 +405,31 @@ InstallOtherMethod( GroupWithGenerators,
...
-end );
+  pcgs:=DefiningPcgs(ElementsFamily(fam));
 
+  G:= rec();
+  ObjectifyWithAttributes( G, typ,
+                          GeneratorsOfMagmaWithInverses, empty,
+                          One, id,
+                          FamilyPcgs,pcgs,HomePcgs,pcgs,GeneralizedPcgs,pcgs);
+
+  #SetGroupOfPcgs (pcgs, G);
```
fingolfin: Why is this commented out? It used to be there. Is it not needed anymore? (Why?) If so, I'd just remove that line.

lib/grp.gi
```diff
@@ -4283,6 +4284,39 @@ local s,b;
...
+        id:=One(gens[1]);
+      fi;
+      if id<>fail then # cannot do identity in magma
+        if ForAny(gens,x->x<>id) then
+          filt:=filt and HasIsTrivial and IsNonTrivial;
+          if isgroup and Length(gens)<=1 then # cyclic not for magmas
+            filt:=filt and IsCyclic;
+          fi;
+        else
+          filt:=filt and IsTrivial and HasIsNonTrivial;
```
fingolfin: Since we now have implications IsTrivial => HasIsNonTrivial and IsNonTrivial => HasIsTrivial I'd kind of prefer if this line (and similar ones throughout this PR) were changed to e.g. `filt:=filt and IsTrivial` -- I find that much easier to grok; indeed, several times when reading through this PR did I misread a `HasIsTrivial` / `IsTrivial` / `HasIsNonTrivial` / `IsNonTrivial`

tst/testbugfix/00005.tst
```diff
@@ -15,4 +15,4 @@ gap> G:=GroupByGenerators( [ [ [ 0, -1, 0, 0 ], [ 1, 0, 0, 0 ],
 Group([ [ [ 0, -1, 0, 0 ], [ 1, 0, 0, 0 ], [ 0, 0, 0, -1 ], [ 0, 0, 1, 0 ] ] 
  ])
 gap> Centralizer(N,G);
-<matrix group of size infinity with 6 generators>
+<matrix group with 6 generators>
```
fingolfin: Interesting. Do you by chance know why we don't know that this centralize is infinite anymore?

fingolfin (commented):
This looks pretty good now. I have some minor comments. Several of them (e.g. the wish to change `IsTrivial and HasIsNonTrival to just `IsTrivial`) can be addressed in a separate PR (and also by somebody else, e.g. me), I just mentioned them because I tried to document everything that occurred to me.

Anyway, I'd like to run a few more tests with this tonight, and (barring more remarks from @hulpke) plan to merge it then or tomorrow.

fingolfin (approved):

PR #2388, opened 2018-04-23
Add zlib as a library

ChrisJefferson:
This is a prototype attempt at adding gzip support to GAP by using zlib. I am putting this up as a PR because I wanted any opinions on if the general design (storing a gzfile in a union) is a broadly sensible idea.

I haven't yet added tests, or run clang-format. Also, possibly once this is merged, I want to add more useful new functionality (like producing gzipped files, and some other small tweaks).

### DIFF
--- a/cnf/m4/m4_ax_check_zlib.m4
+++ b/cnf/m4/m4_ax_check_zlib.m4
@@ -1,0 +1,142 @@
+# ===========================================================================
+#      https://www.gnu.org/software/autoconf-archive/ax_check_zlib.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_CHECK_ZLIB([action-if-found], [action-if-not-found])
+#
+# DESCRIPTION
+#
+#   This macro searches for an installed zlib library. If nothing was
+#   specified when calling configure, it searches first in /usr/local and
+#   then in /usr, /opt/local and /sw. If the --with-zlib=DIR is specified,
+#   it will try to find it in DIR/include/zlib.h and DIR/lib/libz.a. If
+#   --without-zlib is specified, the library is not searched at all.
+#
+#   If either the header file (zlib.h) or the library (libz) is not found,
+#   shell commands 'action-if-not-found' is run. If 'action-if-not-found' is
+#   not specified, the configuration exits on error, asking for a valid zlib
+#   installation directory or --without-zlib.
+#
+#   If both header file and library are found, shell commands
+#   'action-if-found' is run. If 'action-if-found' is not specified, the
+#   default action appends '-I${ZLIB_HOME}/include' to CPFLAGS, appends
+#   '-L$ZLIB_HOME}/lib' to LDFLAGS, prepends '-lz' to LIBS, and calls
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -1,8 +1,8 @@
 dnl
 dnl Setup autoconf
 dnl
-AC_PREREQ([2.68])
-AC_INIT([GAP], [4.dev], [support@gap-system.org], [gap], [http://www.gap-system.org/])
+AC_PREREQ([2.69])
```
fingolfin: Why these changes?

cnf/m4/m4_ax_check_zlib.m4
```diff
@@ -0,0 +1,142 @@
...
+# Locate zlib, if wanted
+#
+if test -n "${zlib_places}"
+then
+	# check the user supplied or any other more or less 'standard' place:
+	#   Most UNIX systems      : /usr/local and /usr
+	#   MacPorts / Fink on OSX : /opt/local respectively /sw
+	for ZLIB_HOME in ${zlib_places} ; do
+	  if test -f "${ZLIB_HOME}/include/zlib.h"; then break; fi
+	  ZLIB_HOME=""
```
fingolfin: I really think that's a bad way to go about it. Hard coding a list of search paths for include files, and the trying to replicate the compiler logic is how bad autoconf files were written 20 years ago. And it breaks on systems which places includes into different subdirs.

I realize you just took this stock M4 file, which is perfectly understandable; but I think we really shouldn't use it. Instead, the existing GMP logic should in `configure.ac` resp. `Makefile.rules` should be duplicated. I can help with that, e.g. when we are together in St Andrews.

cnf/m4/m4_ax_check_zlib.m4
```diff
@@ -0,0 +1,142 @@
...
+zlib_places="/usr/local /usr /opt/local /sw"
+AC_ARG_WITH([zlib],
+[  --with-zlib=DIR         root directory path of zlib installation @<:@defaults to
+                          /usr/local or /usr if not found in /usr/local@:>@
+  --without-zlib          to disable zlib usage completely],
+[if test "$withval" != no ; then
+  AC_MSG_RESULT(yes)
+  if test -d "$withval"
+  then
+    zlib_places="$withval $zlib_places"
```
fingolfin: I think this should be `zlib_places="$withval` instead -- i.e. if the user specifies a location, *use that*, don't silently fallback to something else (e.g. if there was a typo).

ChrisJefferson:
I'm going to chuck this and start again, due to too many conflicts.

fingolfin:
I just tried it, and there was a single file with conflicts in the rebase, which were easy to resolve (for me, at least: I just took your version of the patch; then run `git add src/streams.c && git clang-format src/streams.c && git add src/streams.c` to resolve it). The result can be found under https://github.com/fingolfin/gap/tree/gzip in case you'd like to use this for restarting this PR.

ChrisJefferson:
There was a previous whitespace conflict, which I incorrectly rebased, so the patch was broken before this conflict, so I don't think it's worth trying to fix.

PR #2391, opened 2018-04-23
Remove hack that disabled 64-bit builds of digraphs and float

alex-konovalov:
PR #2119 added Travis test of loading packages, but disabled building Digraphs and Float because they can be compiled, but their loading failed with an "undefined symbol: atexit" error.

This PR removes that hack in order to check if the problem persists and act accordingly.

### DIFF
--- a/etc/ci.sh
+++ b/etc/ci.sh
@@ -97,14 +97,6 @@ do
         # also skip itc because it requires xgap
         rm -rf itc*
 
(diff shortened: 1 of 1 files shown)


alex-konovalov:
I don't know what helped to this, but now both Float and Digraphs are both buildable and loadable (CC @james-d-mitchell and @laurentbartholdi)

PR #2393, opened 2018-04-23
Track location of `InstallMethod` and `InstallImmediateMethod`

### LINKED ISSUE #2239: Idea: record location of `InstallMethod` call

fingolfin:
This PR makes the information displayed by `TraceMethods` and by `ApplicableMethod` much more useful in many case, which resolves #2239. For example, consider this method for `AsList`, installed in `coll.gi`:

    InstallMethod( AsList,
        "for collections that are constant time access lists",
        [ IsCollection and IsConstantTimeAccessList ],
        Immutable );

With GAP 4.8, no location information is shown at all:

    gap> TraceMethods(AsList); AsList("abc");;
    #I  AsList: for collections that are constant time access lists

With GAP 4.9 / master, the location where the method (i.e., the
underlying function) was defined is shown, which already does the job in
many cases, but not here:

    gap> TraceMethods(AsList); AsList("abc");;
    #I  AsList: for collections that are constant time access lists at src/objects.c:fail

With this PR, we get this (note that instead of GAPROOT, the actual path
is printed)

    gap> TraceMethods(AsList); AsList("abc");;
    #I  AsList: for collections that are constant time access lists at GAPROOT/lib/coll.gi:302

For we immediate methods, this is what we got before this PR:

    gap> TraceImmediateMethods( );
    gap> g:= Group( (1,2,3), (1,2) );;
    #I  immediate: Size
    #I  immediate: IsCyclic
    #I  immediate: IsCommutative
    #I  immediate: IsTrivial
    gap>
    gap> M0:=Magma(FamilyObj([1]), []);;
    gap> AsSSortedList(M0);
    #I  immediate: RepresentativeSmallest
    #I  immediate: AsList
    #I  immediate: EnumeratorSorted
    #I  immediate: GeneratorsOfDomain
    #I  immediate: Size
    #I  immediate: RepresentativeSmallest
    #I  immediate: IsFinite
    #I  immediate: IsTrivial
    #I  immediate: IsEmpty
    #I  immediate: IsNonTrivial
    #I  immediate: Size
    #I  immediate: RepresentativeSmallest
    [  ]

After (note how e.g. the entry for EnumeratorSorted points at
coll.gi:379, and not at the location where the method function
was defined):

    gap> TraceImmediateMethods( );
    gap> g:= Group( (1,2,3), (1,2) );;
    #I RunImmediateMethods
    #I  immediate: Size at GAPROOT/lib/coll.gi:179
    #I  immediate: IsCyclic at GAPROOT/lib/grp.gi:40
    #I  immediate: IsCommutative at GAPROOT/lib/magma.gi:196
    #I  immediate: IsTrivial at GAPROOT/lib/magma.gi:130
    gap>
    gap> M0:=Magma(FamilyObj([1]), []);;
    #I RunImmediateMethods
    #I RunImmediateMethods
    gap> AsSSortedList(M0);
    #I RunImmediateMethods
    #I  immediate: RepresentativeSmallest at GAPROOT/lib/coll.gi:226
    #I  immediate: AsList at GAPROOT/lib/domain.gi:206
    #I  immediate: EnumeratorSorted at GAPROOT/lib/coll.gi:379
    #I  immediate: GeneratorsOfDomain at GAPROOT/lib/domain.gi:181
    #I  immediate: Size at GAPROOT/lib/coll.gi:183
    #I  immediate: RepresentativeSmallest at GAPROOT/lib/coll.gi:215
    #I  immediate: IsFinite at GAPROOT/lib/coll.gi:148
    #I  immediate: IsTrivial at GAPROOT/lib/coll.gi:116
    #I  immediate: IsEmpty at GAPROOT/lib/coll.gi:97
    #I  immediate: IsNonTrivial at GAPROOT/lib/coll.gi:134
    #I RunImmediateMethods
    #I  immediate: Size at GAPROOT/lib/coll.gi:179
    #I RunImmediateMethods
    #I  immediate: RepresentativeSmallest at GAPROOT/lib/coll.gi:215
    [  ]


fingolfin:
BTW, while I found outputting `RunImmediateMethods` helpful for debugging, not everybody might agree, so I can remove this again. Or perhaps we should have different levels of tracing: off = level 0; level 1 = only print the invoked methods;  level 2: also print the `RunImmediateMethods` string, and perhaps in addition, print the value of `RUN_IMMEDIATE_METHODS_RUNS` (and print `RUN_IMMEDIATE_METHODS_CHECKS` next to each executed method).
We then might also want to use an `InfoLevel` (and perhaps `Info`, though I am less clear on that; in any case, one can use an `InfoLevel` without `Info` just fine)

fingolfin:
@ThomasBreuer this, however, breaks a function in `Browse` for browsing methods. To fix this, `Browse` needs to be changed to use `BASE_SIZE_METHODS_OPER_ENTRY` when it iterates over the methods (this will make it future proof against further additions). To keep it compatible with older GAP versions, you could simply bind it to the old value 4 if missing.

PR #2395, opened 2018-04-23
Make types returned by NEW_TYPE immutable

fingolfin:
We already do that in HPC-GAP.

However, I am slightly worried that this might break some package which might try to modify a type object. Perhaps @alex-konovalov can run this PR through the full set of package tests somehow (if there is a way to do that w/o too much work, that is!)

@stevelinton any thoughts about this?

### DIFF
--- a/hpcgap/src/c_type1.c
+++ b/hpcgap/src/c_type1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include "compiled.h"
-#define FILE_CRC  "-43167549"
+#define FILE_CRC  "-25327671"
 
 /* global variables used in handlers */
 static GVar G_NAME__FUNC;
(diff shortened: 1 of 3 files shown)


alex-konovalov:
@fingolfin I have a special parametrised Jenkins build for such purposes - all it requires is to specify the PR number as a parameter. I will check this PR - you can invite me to review it so it will be on my radar.

stevelinton:
I'm not convinced this change does anything much. MakeImmutable applied to a positional object (which a type is) just resets the IsMutableObj bit in the type (which is not set anyway) and calls PostMakeImmutable on the object. 

You can only change a type through ![] and I can see no library code that does that. Any package that modifies types "by brute force" in this way, probably deserves whatever it gets :-)

alex-konovalov:
This has some impact on packages, indeed - the log reports:
```
LoadAllPackages test failed: 3 configurations successful out of 4
Additionally, 1 packages crashed during loading
```
I will look for more details in the log later.

alex-konovalov:
@fingolfin I will run tests again with the newer version of this PR

stevelinton:
I remain unsure why this change will do anything at all. If it does cause changes, can we try and understand why  before we merge anything.

fingolfin:
I fully agree and added the DO NOT MERGE LABEL.

alex-konovalov:
I've triggered a new Jenkins test this morning to check this PR with the latest selection of packages. Will post updates from the test logs later today.

PR #2398, opened 2018-04-24
Fix bug in PositionStream, add tests for reading files

ChrisJefferson:
This fixes a bug in `PositionStream`, and also adds a big general test for reading of files, and positions in streams.


src/streams.c
```diff
@@ -1582,17 +1582,31 @@ Obj FuncPOSITION_FILE (
...
             (Int)TNAM_OBJ(fid), 0L,
             "you can replace <fid> via 'return <fid>;'" );
     }
-    
-    ret = SyFtell( INT_INTOBJ(fid) );
-    return ret == -1 ? Fail : INTOBJ_INT(ret);
+
+    Int ifid = INT_INTOBJ(fid);
+
+
```
fingolfin: In case you end up making other changes: perhaps remove second empty line? or even both empty lines, and then for symmetry change `ret = ...` to `Int ret = ...`, and remove the declaration of `ret` above?

tst/teststandard/files/files.tst
```diff
@@ -0,0 +1,104 @@
+gap> START_TEST("files.tst");
+gap> testRead := function(testName, openFunc, lines, hasPosition)
```
fingolfin: so... `testName` is only used for printing in error messages, presumably to identify the failing tests case? `openFunc` opens a new stream, `lines` contains the expected content of that stream, and `hasPosition` indicates whether `PositionStream` is supported. Correct?

Perhaps add this into a comment, and/or rename `lines` to `expectedLines` or `expected`; and `hasPosition` to `supportsPositionStream`?

fingolfin:
Clarification for release notes: the bug in `PositionStream` was that it did not take buffering into account, and thus could return incorrect values.

PR #2399, opened 2018-04-24
kernel: reduce code duplication in function dispatchers

fingolfin:
... by putting the duplicate code into `static ALWAYS_INLINE` functions and then relying on the compiler to optimize the code. Which it does, in my tests with clang and gcc.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -332,7 +332,8 @@ Obj DoWrap6args (
 /* Pull this out to avoid repetition, since it gets a little more complex in 
    the presence of partially variadic functions */
 
-Obj NargError( Obj func, Int actual) {
+Obj NargError(Obj func, Int actual)
+{
   Int narg = NARG_FUNC(func);
 
   if (narg >= 0) {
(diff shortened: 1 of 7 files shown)


src/calls.c
```diff
@@ -550,267 +568,88 @@ Obj DoProf0args (
...
-    timeCurr = SyTime() - TimeDone - timeCurr;
-    SET_TIME_WOUT_PROF( prof, TIME_WOUT_PROF(prof) + timeCurr );
-    TimeDone += timeCurr;
-    storCurr = SizeAllBags - StorDone - storCurr;
-    SET_STOR_WOUT_PROF( prof, STOR_WOUT_PROF(prof) + storCurr );
-    StorDone += storCurr;
-
-    /* return the result from the function                                 */
-    return result;
+    return DoProfNNNargs(self, 3, arg1, arg2, arg3, arg4, 0, 0);
```
ChrisJefferson: `self, 4`? Also, this passed tests so we might want some boring tests just to make sure all these different length functions work properly.

fingolfin: Ouch, bad case of copy & paste. Actually, the tests (some which were rather recently added) showed me already another bug I introduced while doing this.

Unfortunately, adding sensible tests for function profiling is rather annoying.

fingolfin: Now fixed. In my defense, at least now it was very easy to spot this; we had similar bugs in very similar code for a long time, which were only resolved last year, and they were much harder to spot, as things were spread out over pages and pages of code.

That's of course no excuse for wrong code, but it shows that reducing the duplication makes it easier to spot bugs :-)

PR #2403, opened 2018-04-25
lib: define POS_{FAMILY,FLAGS}_TYPE constants

fingolfin:
UPDATE: now also moves the POS_*_TYPE constants into the kernel, as their values are also used there. Moreover, this allows the GAP compiler to produce optimized code when these constants are used.

### DIFF
--- a/hpcgap/src/c_type1.c
+++ b/hpcgap/src/c_type1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include <src/compiled.h>
-#define FILE_CRC  "-95946497"
+#define FILE_CRC  "132228056"
 
 /* global variables used in handlers */
 static GVar G_NAME__FUNC;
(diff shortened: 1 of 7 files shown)


fingolfin:
Indeed, it is. I checked, though, and even in 1997, they were not there. My only guess is that because they were so often accessed, using 1 and 2 directly was an "important" (?) optimization, to avoid the overhead of an extra gvar lookup whenever accessing the family or flags of a type (I wonder how much that actually costs...). But now that we have proper "constants", this doesn't matter, and so it's a clear win to use named constants, as the code gets easier to read.

hpcgap/src/c_type1.c
```diff
@@ -1083,17 +1087,23 @@ static Obj  HdlrFunc11 (
...
   l_cached = t_1;
   
-  /* if IS_EQUAL_FLAGS( flags, cached![2] ) then */
+  /* if IS_EQUAL_FLAGS( flags, cached![POS_FLAGS_TYPE] ) then */
   t_3 = GF_IS__EQUAL__FLAGS;
-  C_ELM_POSOBJ_NLE( t_4, l_cached, 2 );
+  t_5 = GC_POS__FLAGS__TYPE;
+  CHECK_BOUND( t_5, "POS_FLAGS_TYPE" )
+  CHECK_INT_SMALL_POS( t_5 )
+  C_ELM_POSOBJ_NLE( t_4, l_cached, INT_INTOBJ(t_5) );
```
fingolfin: Hmmpf, looking at this, the compiler does not pick up these constants and generates suboptimal code as a result. The reason is that gac does not load the GAP library. To fix this, I'll have to move the constants to the kernel after all *sigh*. Will do that later.

PR #2407, opened 2018-04-25
Remove from grpauto.tst a test which exceeds memory limits

alex-konovalov:
This will allow to avoid test failure reported in #2377 in
stable-4.9 branch. The fix in the master branch will be done
differently. This test is duplicated in benchmarks, so we will
monitor its status there.

### DIFF
--- a/tst/testextra/grpauto.tst
+++ b/tst/testextra/grpauto.tst
@@ -93,18 +93,6 @@ gap> G:=Group((1,3,8,21,37,43,36,35)(2,6,15,30,41,29,19,4)(5,7,18,34,46,
 gap> Size(AutomorphismGroup(G)); 
 2880000
 
-#
-# hard-iso4
-#
-gap> G:=PcGroupCode(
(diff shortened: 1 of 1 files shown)


hulpke:
Just curiosity: How did we decide on the 1G limit? These larger group theoretic calculations are testing a lot of different (and involved) functions on the way.

alex-konovalov:
@hulpke `-o 1g` was used in `make test...` targets for quite long time - I can't find a link to some discussion about possibly raising limits. IIRC last time we still decided to keep `-o 1g` which is lower than default. My argument in favour of 1g is that it helps to track memory regressions - like in this case, which would remain unnoticed in default settings.

Today, if you want to add a genuine test which demands more memory, I suggest to put it to `benchmark`, which allow to use different (also standardised) memory limits (this is the current setup, open for suggestions and improvements). Benchmarks are run weekly on Jenkins CI in St Andrews.

PR #2408, opened 2018-04-26
kernel: use more accurate marking

fingolfin:
This PR how we mark-for-garbage-collection in various bag types with GASMAN. This is motivated by @rbehrends' work on integrating Julia GC, see PR #2092 

Specifically, when marking bags during garbage collection, we were very liberal in what we passed on to `MarkBag` (directly or indirectly). This is no big deal with GASMAN, as GASMAN can easily decide whether a given pointer is a master pointer or not (it has to be in a certain range), and ignore all other values passed to it. For other GCs, this is not quite as easy, as master pointers typically are not allocated from a fixed pool (though one could think about doing that... with the provision that one may have to maintain multiple such pools...)

This PR resolves most of these issues:

* we abused `T_BODY` bags in `code.c` to maintain two stacks of statements resp. expressions, which are not bags. But in a `T_BODY`, the first three slots are expected to be bag refs. This wasn't the case for these stacks. So now we use a `T_DATOBJ` for them (and make sure to leave its type slot intact)
* we marked all slots in plists via `MarkAllSubBags`; but the first slot contains the length of the list; fixed via new marking function `MarkAllButFirstSubBags`
* we marked all slots in precords and component objects via `MarkAllSubBags`; but only slot 0 (for component objects), and the odd slots starting with slot 3, contain bag refs. Fixed with a custom marking function `MarkPRecSubBags`
* we marked all slots in `T_FUNCTION` objects via `MarkAllSubBags`; but we store pointers to C functions in the first 8 slots, and also stored raw `UInt` values in two other slots. The latter were replaced by immediate integer objects; and a custom marking function `MarkFunctionSubBags` was added, which skips the first 8 slots, and only marks the rest.
* I also added a debug `#define DEBUG_GASMAN_MARKING` which enables code that tests if `MarkBag` is called on something that is neither a valid bag ref, a valid immediate object (integer or FFE), nor a null pointer. This can be used to track down more instances of the problem type described here, which is much easier than trying to do so while also hooking up a new GC at the same time ;-).
* For LVars and HVars, there is an `Expr` in slot 1 which should not marked. I moved it from slot 1 into slot 0, and then made it use `MarkAllButFirstSubBags`. This also required tweaking the LVars pool code, which I did (it is now also documented and more readable).

### DIFF
--- a/src/boehm_gc.c
+++ b/src/boehm_gc.c
@@ -625,6 +625,10 @@ void MarkAllSubBags(Bag bag)
 {
 }
 
+void MarkAllButFirstSubBags(Bag bag)
+{
+}
+
 void MarkArrayOfBags(const Bag array[], UInt count)
 {
 }
--- a/src/calls.c
+++ b/src/calls.c
@@ -957,6 +957,7 @@ Obj NewFunctionT (
     SET_NAME_FUNC(func, ConvImmString(name));
     SET_NARG_FUNC(func, narg);
     SET_NAMS_FUNC(func, nams);
+    SET_NLOC_FUNC(func, 0);
     if (nams) MakeBagPublic(nams);
     CHANGED_BAG(func);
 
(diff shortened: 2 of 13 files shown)


ChrisJefferson:
Looks good to me. Just in case anyone has the same thought as me, I wondered about only checking up to `LEN_PLIST` or `LEN_PREC`, rather than the end of the bag, but that doesn't work as we often put values into lists/records before we update their length (and I guess the problems will be even worse in parallel!)

PR #2409, opened 2018-04-26
Some Input/Output cleanup

ChrisJefferson:
These two commits are some cleanups I wrote while adding zlib support. I decided to seperate them out as to make review easier.

The first basically simplifies how we handle `.echo` and `.fp`, to consistently always write to `.echo`. In practice this rarely make a difference because except for descriptors 0 and 2, for any descriptor we can write to `.echo` and `.fp` are always the same value.

The second makes it clearer when we are checking if a descriptor is currently in use.

Both of these patches will make the later gzip support much easier to read and smaller, because now I only have to change writing, and checking if streams are in use, in one place.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -736,6 +736,17 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
     i = append ? OpenAppend( CSTR_STRING(filename) )
                : OpenOutput( CSTR_STRING(filename) );
     if ( ! i ) {
+        if (strcmp(CSTR_STRING(filename), "*errout*")) {
+            // When trying to print an error opening *errout* failed,
+            // We exit GAP after trying to print an error.
+            // First try printing an error to stderr
+            int ret = fputs("gap: panic, could not open *errout*!\n", stderr);
+            // If that failed, try printing to stdout
+            if(ret == EOF) {
+                fputs("gap: panic, could not open *errout*!\n", stdout);
+            }
+            SyExit(1);
+        }
         ErrorQuit( "%s: cannot open '%g' for output",
                    (Int)funcname, (Int)filename );
         return 0;
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -87,12 +87,13 @@ typedef void sig_handler_t ( int );
 
 
 /* utility to check return value of 'write'  */
-ssize_t writeandcheck(int fd, const char *buf, size_t count) {
+ssize_t echoandcheck(int fid, const char *buf, size_t count) {
   int ret;
-  ret = write(fd, buf, count);
+
+  ret = write(syBuf[fid].echo, buf, count);
   if (ret < 0)
     ErrorQuit("Cannot write to file descriptor %d, see 'LastSystemError();'\n",
-               fd, 0L);
+               fid, 0L);
   return ret;
 }
 
@@ -588,7 +589,6 @@ void syWinPut (
     const Char *        cmd,
     const Char *        str )
 {
-    Int                 fd;             /* file descriptor                 */
     Char                tmp [130];      /* temporary buffer                */
     const Char *        s;              /* pointer into the string         */
     Char *              t;              /* pointer into the temporary      */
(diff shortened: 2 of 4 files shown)


src/sysfiles.c
```diff
@@ -800,7 +816,7 @@ Int SyFopen (
     else if ( strcmp( name, "*errin*" ) == 0 ) {
         if ( strcmp( mode, "r" ) != 0 )
           return -1;
-        else if ( syBuf[2].fp == -1 )
+        else if ( !SyBufInUse( 2 ) )
```
fingolfin: off-topic, but one other thing we may want to do (I thought about this before) is to add an enum for the special values 0, 1, 2, 3 (stdin/out, errin/out). Say `GAP_STDIN` or `ID_STDIN` etc. ?

ChrisJefferson:
This is now passing tests (actually, it was fine before, but it was hitting that memory corruption bug)

src/sysfiles.c
```diff
@@ -597,12 +597,8 @@ void syWinPut (
     if ( ! SyWindow || 4 <= fid )
         return;
 
-    /* get the file descriptor                                             */
-    if ( fid == 0 || fid == 2 )  fd = syBuf[fid].echo;
-    else                         fd = syBuf[fid].fp;
-
```
fingolfin: Is there a particular benefit in renaming `writeandcheck` to `echoandcheck`, and changing its `fd` parameter to `fid`, as opposed to just replacing the above few lines by `fd = syBuf[fid].echo`? That does the same, doesn't it?

src/sysfiles.c
```diff
@@ -753,6 +749,22 @@ SYS_SY_BUF syBuf [256];
...
+
+// There is no explicit method to mark syBufs as used,
+// they are marked as used when created.
+
+// Check if a member of syBuf is in use
+static Int SyBufInUse(Int i)
+{
+    GAP_ASSERT(i >= 0 && i < 256);
+    return syBuf[i].fp != -1;
+}
```
fingolfin: (Hmm, I tried to comment earlier with my phone, but it seems those comments didn't make it?)
It's a bit odd that marking a buffer unused is public, but checking whether it is in use is private. BTW, `src/streams.c` also could in theory use `SyBufInUse` when it does `if (syBuf[ifid].bufno >= 0) {`. So perhaps it should be public, too?

Alternatively, the code using `SyMarkBufUnused` in `src/system.c` could be moved into a new `InitSysBuf` function in `sysfiles.c` which is then call from `system.c`

src/streams.c
```diff
@@ -736,6 +736,10 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
     i = append ? OpenAppend( CSTR_STRING(filename) )
                : OpenOutput( CSTR_STRING(filename) );
     if ( ! i ) {
+        if (strcmp(CSTR_STRING(filename), "*errout*")) {
+            // When trying to print an error opening *errout* failed
+            SyExit(1);
```
fingolfin: Good on, I actually run into this problem recently. An error message before exiting would be helpful, though, e.g.
```C
  fputs("gap: panic, could not open *errout*!\n", stderr);
  SyExit(1);
```

ChrisJefferson: I will add that. I don't think it will actually output anything (as errout is usually stderr, so if we can't print to errout, we probably can't print to stderr either). However, it doesn't hurt to try!

src/sysfiles.c
ChrisJefferson: I don't mind make syBufInUse public, I just didn't need it anywhere outside this file, so made it static.

`syBuf[ifid].bufno >= 0` isn't actually checking if `ifid` is in use -- `SyFtell` earlier in the function already called SyBufInUse and returned -1 if it failed. It's checking if the `syBuf` has a `syBuffer` attached to it, which is an entirely different problem (yes, the SyBuf / SyBuffer naming might be unfortunate).

src/sysfiles.c
ChrisJefferson: It's not that useful here, but it will be useful for my later patch, as it means that places where we have raw file handles are much smaller (and all those places need to be changed when we add gzip support).

PR #2410, opened 2018-04-26
Fix order of sed arguments for bsd/macOS sed compatibility

xzcvczx:
Change the order of sed arguments in Makefile.rules to make it compatible with bsd/macOS sed.


xzcvczx:
i am not installing to system, just inside the gap folder, and yeah i see it doesn't work

fingolfin:
OK. However, I don't think this change is actually correct: I am on Mac OS X, and just checked in the man page: `-E` comes before `-e`. And indeed, it certainly must not come between `-e` and its command argument...

The real problem here is the user of `-i`, which is a non-POSIX extension, which differs significantly between GNU sed and BSD sed. To work around that, one either has to detect which `sed` it is, and then use the right arguments (but then there are yet other `sed` variants which may not support  either form of `-i`!). Or, better, avoid using `-i` altogether. In this case, that could be either done by piping the original files through sed, and the output into the install location; or else by using temporary files. We could also create our own wrapper script, I guess...

Though ultimately, this ugly use of `sed` will become redundant anyway, once we change back from `#include <src/FOO.h>` to `#include "FOO.h"`. This was actually on my build system TODO / issue list, but I don't see an issue here now, so I guess it was lost in the migration from my repo to here? Anyway, I'll add one now.

fingolfin:
See issue #2411

Issue #2411, opened 2018-04-26
Change back from `#include <src/FOO.h>` to `#include "FOO.h"`


fingolfin:
The use of `<>` instead of `""` was to allow HPC-GAP to override any `*.h` file with a custom version. But now `hpcgap/src/` is reduced to two `.c` files, so this part is not necessary anymore.

This leaves the `src/` prefix. Part of the reason this was in was to allow packages with kernel extensions to build with HPC-GAP, which means they need to be able to `#include` files from GAP's `src/hpc/` subdir; but the standard `ac_find_gap.m4` used in most packages did not support that. But I think that reason is invalid, because if packages are supposed to work with HPC-GAP, then those packages must be adjusted anyway... The only two packages with kernel extension we rely on heavily (for CI on Travis) are IO and profiling. Both are either already using the right include path, or else I am confident we can't make fixed released within a short period of time.

All other packages need to be adapted for HPC-GAP anyway, the build system is only a minor part of that.

(This was on my original buildsystem TODO / issue list, and I thought we had an issue for it here, but I can't find it now, so I must have forgotten when migrating away from the issue tracker of my GAP fork.... ahhh, I see now: https://github.com/fingolfin/gap/issues/89 and then a failed attempt to implement this, with discussion:  #1443 )

fingolfin:
This was done by PR #2412

PR #1443, opened 2017-06-21
Change #include <src/foo.h> to #include "foo.h"

fingolfin:
The main reason we used the former since the HPC-GAP merge was to allow us to override GAP headers with modified headers for HPC-GAP during compilation. But now the kernel header files all have been merged, so this is no longer necessary.

We also change <gen/config.h> back to "config.h", and adjust the build system suitably.

Finally, we rename our util.h to utils.h, to avoid a conflict with the BSD system header <util.h>.

PR #2412, opened 2018-04-26
Change #include <src/foo.h> to #include "foo.h"

fingolfin:
This essentially brings us back to how things were in GAP 4.8. Ideally, I would have wanted to get this already reverted in GAP 4.9, but alas, that didn't work out.

A major motivation for this change is that it is needed (or at least very, very useful) for future work on implementing `make install` -- if the GAP headers are to be installed into, say `PREFIX/include/gap/`, then they must not include the `src/` bit in the path. And the `<>` instead of `""` only was ever done to support a build system hack for HPC-GAP; the need for that is gone by now, though.

The main (and only) argument *against* this PR I can think of is that it may conflict with some other in-flight PRs. But IMHO that's not a big deal, as those PRs are easy to adjust. And there are not that many open PRs adjusting the kernel which are not by me, *and* which also modify  some `#include`. (Of course it'd be easier to count them if people used the `kernel` label consistently... )

Some technical notes: This is a revival of PR #1443 ; for details, see also issue #2411.

Unlike that issue, I have split the changes into two commits, for easier reviewing. The second commit is generated in a purely mechanical fashion, with the changes having been generated by these commands (using BSD sed syntax; for GNU sed, replace `-i ''` by just `-i`):
```sh
sed -i '' -E -e 's;#include <(src|gen)/([^"]+)>;#include "\2";' src/*.* src/hpc/*.* hpcgap/src/*.* tst/test-compile/*.c
```

This means the first commit won't actually built, which is mildly annoying when bisecting, but that's what `git bisect skip` was made for. I think being able to cleanly review the changes and regenerate the PR is more important. (Also, we could "squash" this PR instead of rebasing it, if we truly worry about this).


fingolfin:
This requires a change to the profiling package, see https://github.com/gap-packages/profiling/pull/51

PR #2413, opened 2018-04-26
kernel: track LoopNesting in the reader, not coder

fingolfin:
These constitute syntax errors, so it makes sense to report then before
coding. This also yields uniform error handling, including when using
break/continue outside of coding (which is always illegal).

As an added bonus, we get rid of LoopStack and LoopStackCount.

I also think that the new code is much easier to understand, verify and debug, but I may be biased ;-).

(The end game here is that I want to move all `SyntaxError` calls to either `read.c` or `scanner.c`; right now we also have some in `code.c` and `intrprt.c`. This PR gets rid of those in the coder. I have more coming up which get rid of some of those in the interpreter.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -28,7 +28,6 @@
 #include <src/read.h>
 #include <src/records.h>
 #include <src/saveload.h>
-#include <src/scanner.h>
 #include <src/stringobj.h>
 #include <src/vars.h>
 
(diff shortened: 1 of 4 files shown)


fingolfin:
@stevelinton Thanks! And I think that's all one can expect from a reviewer (just like with a paper referee). If this PR introduces a regression after all, of course the author is to "blame", not the reviewers. And of course I'll work on resolving any such regressions.

Anyway, the tests did pass, but now it needs to be rebased due to some conflicts in `read.c`, we'll see afterwards if the tests still pass...

PR #2416, opened 2018-04-26
kernel: remove S_PARTIALFLOAT1, mark float parsing hack

fingolfin:
Parsing float literals that start with a dot, like `.123`, is
implemented using a hack, where the reader tells the scanner that it
should look for a number expression (normally, the scanner never changes
the state of the reader, except for parsing long integers and float
literals).

We change the way this is done to make it stand out more. Further
benefits from the change are that we can get rid of S_PARTIALFLOAT1,
simplifying the intricate state machine for parsing. Moreover,
STATE(Symbol) now is read-only in the reader.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -980,11 +980,6 @@ void ReadLongNumber(
            /*      Match(S_PARTIALINT, "integer", follow);*/
            break;
 
-         case S_PARTIALFLOAT1:
-           assert(0);
-           Pr("Parsing error, this should never happen", 0L, 0L);
-           SyExit(2);
-
          case S_PARTIALFLOAT2:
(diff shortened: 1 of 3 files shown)


fingolfin:
@stevelinton Thanks for the review, and sorry for the confusion; since this is also the commit message, I'd like to fix it. However, I am not quite sure what you mean, could you quote the specific part you think is wrong, and ideally also a suggested improvement?

fingolfin:
Ohh, I think I got it: "normally, the scanner never changes the state of the reader" should be reversed (though I guess technically, both statements can be regarded as either right, or wrong, depending on how one interprets "changes the state" -- perhaps best to just drop this sentence).

PR #2421, opened 2018-04-30
Fix iterating over an empty cartesian product

### LINKED ISSUE #2420: IteratorOfCartesianProduct and empty lists
`IteratorOfCartesianProduct` gives the following error if one of the arguments is empty:
```
	Error, Each arguments must be a finite collection called from
	IteratorOfCartesianProduct2( arg ...
```
I think this is a bug. For example:
```
	IteratorOfCartesianProduct([1,2],[]);
```
gives this error, while 
```
	Cartesian([1,2],[]);
```
correctly returns `[]`.
Calling
```
	IteratorOfCartesianProduct([[1,2],[]]);
```
also gives an error.

I have tried this in GAP 4.8.4 and GAP 4.8.10.

fingolfin:
Fixes #2420

### DIFF
--- a/lib/combinat.gi
+++ b/lib/combinat.gi
@@ -805,8 +805,11 @@ BindGlobal( "ShallowCopy_Cartesian",
 BindGlobal( "IteratorOfCartesianProduct2",
     function( listsets )
     local s, n, x;
-    if not ForAll( listsets, IsCollection ) and ForAll( listsets, IsFinite ) then
-      Error( "Each arguments must be a finite collection" );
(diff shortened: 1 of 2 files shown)


fingolfin:
Perhaps this would be a good candidate for backporting to GAP 4.9.2 - the fix is obvious and self-contained, with very little chance of causing regressions.

fingolfin:
Backported in 463f45f1ed92c3b9263d5a3a325b2b731921c684

PR #2422, opened 2018-04-30
Improve tracking of InstallMethod and DeclareOperation

fingolfin:
So far, we used `INPUT_LINENUMBER()` to record the location of an `InstallMethod` or `DeclareOperation` call. But in case such a call takes multiple lines, that points to the *last* line, while we would prefer if it pointed to the *first* line. To achieve that, we add a new global variable `READEVALCOMMAND_LINENUMBER` which is updated at the start of `ReadEvalCommand`.

One may debate whether it might be better if `READEVALCOMMAND_LINENUMBER` was a global function which returns the desired result. If somebody has a good argument for that, I'll be happy to adapt the patch. The only reason I used a global variable is that it was the quickest way to implement this.

### DIFF
--- a/hpcgap/src/c_oper1.c
+++ b/hpcgap/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include <src/compiled.h>
-#define FILE_CRC  "-121710233"
+#define FILE_CRC  "-126023111"
 
 /* global variables used in handlers */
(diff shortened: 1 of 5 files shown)


ssiccha:
I don't have strong feelings about `READEVALCOMMAND_LINENUMBER` being a gvar or a function.

LGTM

fingolfin:
@ThomasBreuer good point; I changed it to now store a triple `[ filename, firstline, lastline ]`.

PR #2425, opened 2018-04-30
Fix list error messages for too large indices; add some more tests

fingolfin:
We called INT_INTOBJ on indices without verifying that given value actually *is* an immediate integer.

Also add more content to `tst/testinstall/kernel/lists.tst` which tests the above issue, and more

### DIFF
--- a/src/lists.c
+++ b/src/lists.c
@@ -824,11 +824,7 @@ Obj ElmsListCheck (
     Obj                 list,
     Obj                 poss )
 {
-    if ( ! IS_POSS_LIST(poss) ) {
-        ErrorQuit(
-      "List Elements: <positions> must be a dense list of positive integers",
-            0L, 0L );
-    }
+    CheckIsPossList("List Elements", poss);
     return ELMS_LIST( list, poss );
 }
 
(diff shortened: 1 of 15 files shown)


src/lists.c
```diff
@@ -1760,20 +1756,8 @@ void            AssListLevel (
...
-        }
-        if ( LEN_LIST(lists) != LEN_LIST(objs) ) {
-            objs = ErrorReturnObj(
-         "List Assignment: <objs> must have the same length as <lists> (%d)",
-                LEN_LIST(lists), 0L,
-                "you can replace <objs> via 'return <objs>;'" );
-        }
-    }
+    CheckIsDenseList("List Assignment", "objs", objs);
+    CheckSameLength("List Assignment", "objs", "lists", objs, lists);
```
fingolfin: Just to be explicit about it: Here we loose some flexibility, in that previously we allowed the user to `return <obj>` a new list of objects, and now we error out. I thought about restoring this here (and in a few other similar places), but it just didn't seem worth the effort. If somebody disagrees, let me know.

PR #2427, opened 2018-04-30
Improve and test support for custom functions

fingolfin:
* NamesLocalVariablesFunction actually is a kernel operation, mark it as such
* fix a crash caused by infinite recursion when trying to display a function
  which is not a `T_FUNCTION`
* add tests for installing custom methods for NameFunction,
  NamesLocalVariablesFunction, and NumberArgumentsFunction


fingolfin:
@ThomasBreuer good remarks, thanks, hopefully all addressed now (also the test failure in HPC-GAP)

Issue #50, opened 2015-03-17
io for hpcgap


mohamed-barakat:
* I downloaded io-hpc-4.2.tar.gz from http://hpc.gap-system.org/hpcgap-alpha/
* ./configure worked
* make failed:

```
$ make
/bin/sh ./libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I./src  -I/Users/mo/software/gap/hpcgap/gap -I/Users/mo/software/gap/hpcgap/gap/bin/x86_64-apple-darwin14.1.0-gcc-hpc -DCONFIG_H -isystem /sw/include  -g -O2 -MT io_la-io.lo -MD -MP -MF .deps/io_la-io.Tpo -c -o io_la-io.lo `test -f 'src/io.c' || echo './'`src/io.c
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./src -I/Users/mo/software/gap/hpcgap/gap -I/Users/mo/software/gap/hpcgap/gap/bin/x86_64-apple-darwin14.1.0-gcc-hpc -DCONFIG_H -isystem /sw/include -g -O2 -MT io_la-io.lo -MD -MP -MF .deps/io_la-io.Tpo -c src/io.c  -fno-common -DPIC -o .libs/io_la-io.o
In file included from src/io.c:17:
In file included from /Users/mo/software/gap/hpcgap/gap/src/compiled.h:19:
In file included from /Users/mo/software/gap/hpcgap/gap/src/gasman.h:50:
/Users/mo/software/gap/hpcgap/gap/src/atomic.h:5:10: fatal error: 'atomic_ops.h' file not found
#include <atomic_ops.h>
         ^
1 error generated.
make[1]: *** [io_la-io.lo] Error 1
make: *** [all-recursive] Error 1
```

ChrisJefferson:
For now, configure with:

``` ./configure CFLAGS="`cat ../../build/cflags`" ```

alex-konovalov:
Yes, there is already a note on this at https://github.com/gap-system/gap/wiki/Building-HPC-GAP

mohamed-barakat:
Thanks :+1:

PR #2428, opened 2018-05-01
Mark perm cycles as executed

ChrisJefferson:
The profiler normally detects an expression is evaluated when we
call EVAL_EXPR, but we never do this for perm cycles, which leads
to them being marked as read but never executed.

This fixes issue #50 from the profiling package.


fingolfin:
@ChrisJefferson can you please add suitable labels before merging this?

ChrisJefferson:
OK. I don't think this needs adding to the relase notes, and I also see no need to backport it to 4.9, as it isn't fixing any significant bug.

Issue #2400, opened 2018-04-25
Magma/Group confusion


hulpke:
This is something that came up in the context of #2387, but really is an earlier problem that has nothing to do with these changes:

Start the master branch (from scratch and with -A to minimize complications) and add the following function (which makes some basic deductions from an objects order):

    InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
      100, # override system setter
    function(obj,sz)
      SetIsEmpty(obj,sz=0);
      SetIsTrivial(obj,sz=1);
      SetIsNonTrivial(obj,sz<>1);
      SetIsFinite(obj,sz<>infinity);
      TryNextMethod(); # to enforce size setting
    end);

Now call (this is tested in one of the test files...)

    AsGroup([(1,2)]);

and get an error message:

Error, Value property is already set the other way

The reason is that the calculation defines the empty submagma of the magma defined by [(1,2)].
Properties and implication make this submagma a group and thus it contains the trivial permutation, so is not empty (and has Size 1)  but GAP has set before that it is empty. The new setter method just makes this visible.

What is at issue here? Code for Magmas? The implication that they could be groups? Or is it the test for AsGroup?

An easy fix would be to take the test out, but there might be some bug being hidden in the current master branch already.

Puzzled...

hulpke:
The same happens in the tutorial example

    gap> mats:= [ [ [ 0*Z(2), Z(2)^0 ],
    >               [ Z(2)^0, 0*Z(2) ] ],
    >             [ [ Z(2)^0, 0*Z(2) ],
    >               [ 0*Z(2), Z(2)^0 ] ] ];; 
    gap> Size( AdditiveMagma( mats ) );

stevelinton:
Got it, I think.
```
gap> x := MagmaByGenerators([(1,2)]);
<group with 1 generator>
gap> s := Submagma(x,[]);
<group with 0 generators>
gap>    
```
`s` here is wrong, it should be the empty set and not a group. I'm not sure where this implication is coming from, but I would guess that there an InstallSubsetMaintenance call somewhere that forgot about this case.

fingolfin:
This sounds related to https://github.com/gap-system/gap/issues/2240

ThomasBreuer:
Technically, the wrong type comes from the implication
```
IsFiniteOrderElementCollection and IsMagma => IsMagmaWithInverses
```
in `lib/magma.gd`, because any collection of elements in `PermutationsFamily`
[...]

PR #2429, opened 2018-05-01
Partial fix and workaround for #2400

hulpke:
As requested made this a PR of its own. It avoids creating magmas from group elements that have the wrong order.

Tests for this are already in the tst files (therefore no extra test provided here) but will be triggered only once the new method for SetSize in #2387 has been merged.

### DIFF
--- a/lib/addmagma.gi
+++ b/lib/addmagma.gi
@@ -175,7 +175,7 @@ InstallGlobalFunction( SubadditiveMagmaNC, function( M, gens )
     if IsEmpty( gens ) then
       K:= NewType( FamilyObj(M),
                        IsAdditiveMagma
-                   and IsTrivial
+                   and IsEmpty
                    and IsAttributeStoringRep );
       S:= Objectify( K, rec() );
       SetGeneratorsOfAdditiveMagma( S, [] );
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -1209,6 +1209,9 @@ InstallMethod( AsMagma,
     D := AsSSortedList( D );
     L := ShallowCopy( D );
     M := Submagma( MagmaByGenerators( D ), [] );
+    if IsGroup(M) then
+      M:=Submagma(MagmaByGenerators(D),[One(ElementsFamily(FamilyObj(D)))]);
+    fi;
     SubtractSet( L, AsSSortedList( M ) );
     while not IsEmpty(L)  do
         M := ClosureMagmaDefault( M, L[1] );
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -1209,6 +1209,11 @@ InstallMethod( AsMagma,
     D := AsSSortedList( D );
     L := ShallowCopy( D );
     M := Submagma( MagmaByGenerators( D ), [] );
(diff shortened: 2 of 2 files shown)


lib/magma.gi
```diff
@@ -1209,6 +1209,9 @@ InstallMethod( AsMagma,
     D := AsSSortedList( D );
     L := ShallowCopy( D );
     M := Submagma( MagmaByGenerators( D ), [] );
+    if IsGroup(M) then
+      M:=Submagma(MagmaByGenerators(D),[One(ElementsFamily(FamilyObj(D)))]);
+    fi;
```
fingolfin: If I were to look at this code in a year, I'd be completely confused as to what it's about. Even if I spend the extra effort to use `git annotate` to find the commit message and look at it, I don't think I'd be much wiser: "... replacing the empty submagma with the one-submagma, if it turns out a group" -- Huh? How can the submagma turn out to be a group, if it is empty? Isn't it a bug that it's marked as a group in the first place? Or is the bug that it's marked as a group"?

In fact, reading up on #2400 again, I am not convinced this change is correct. Rather, it looks like a workaround for a deeper issue. As such, I don't feel comfortable merging it, until somebody explains why it is correct after all (and ideally that explanation then should be part of a comment before this `if/fi` block).

lib/addmagma.gi
```diff
@@ -175,7 +175,7 @@ InstallGlobalFunction( SubadditiveMagmaNC, function( M, gens )
     if IsEmpty( gens ) then
       K:= NewType( FamilyObj(M),
                        IsAdditiveMagma
-                   and IsTrivial
+                   and IsEmpty
```
fingolfin: I agree with this change.

lib/magma.gi
hulpke: Yes, it is a workaround (and I've added a comment to this effect). The problem is that this bug arises in test suites (in examples of considering groups as magmas) and has the potential of holding up other commits, while it is completely unclear to me how one would fix this.

fingolfin:
A proper fix for the second issue would be this (as described by @ThomasBreuer on issue #2400):
```diff
diff --git a/lib/magma.gd b/lib/magma.gd
index d939bcd62..6577c0921 100644
--- a/lib/magma.gd
+++ b/lib/magma.gd
@@ -130,7 +130,7 @@ DeclareCategory( "IsMagmaWithInverses",
     and IsMultiplicativeElementWithInverseCollection );

 InstallTrueMethod( IsMagmaWithInverses,
-    IsFiniteOrderElementCollection and IsMagma );
+    IsFiniteOrderElementCollection and IsMagmaWithOne );

 #############################################################################
 ##
```
In theory, that would loose some features, but in practice, this does not seem to matter, at least for the testsuite.

We could also add a comment saying e.g.
> # TODO: replace 'IsMagmaWithOne' with 'IsMagma and IsNonEmpty' if it ever becomes available.

I also have a local patch which adds `IsNonEmpty` and `IsInfinite`, with suitable implications (and in the future, also the negative implications, although in practice they seem less important/useful). But I think for now, the above patch should be fine, and be much better than a workaround, IMHO -- assuming it *does* solve the problem you are seeing?

PR #2435, opened 2018-05-03
kernel: speed up IS_SUBSET_FLAGS

fingolfin:
I investigated `IS_SUBSET_FLAGS` because callgrind shows that it is #1 in terms of instructions during GAP startup (followed by `GET_NEXT_CHAR`, `FuncWITH_HIDDEN_IMPS_FLAGS`, `strcmp` and `NextSymbol`):
```
3,617,597,065  PROGRAM TOTALS
360,902,963  src/opers.c:IS_SUBSET_FLAGS.isra.45 [/home/horn/Projekte/GAP/gap.github/gap]
299,271,263  src/io.c:GET_NEXT_CHAR [/home/horn/Projekte/GAP/gap.github/gap]
194,587,622  src/opers.c:FuncWITH_HIDDEN_IMPS_FLAGS [/home/horn/Projekte/GAP/gap.github/gap]
112,844,582  /build/glibc-Cl5G7W/glibc-2.23/string/../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:__strcmp_sse2_unaligned [/lib/x86_64-linux-gnu/libc-2.23.so]
 99,232,409  src/scanner.c:NextSymbol [/home/horn/Projekte/GAP/gap.github/gap]
...
```
Moreover, cachegrind indicated a huge amount of L1 read misspredictions. With this PR, it actually still has a lot, but overall timings improve.

Accessing TRUES_FLAGS requires a pointer dereference, and often leads to
a cache miss, eating up all advantage to be gained from using it; so
drop it.

To verify, I run

    echo "QUIT;" > quit.g
    perf stat -r 10 -d ./gap -A -q quit.g

on a Linux test machine, which measures the startup performance of GAP
10 times (I also verified that there was an improvement on macOS, but
less rigorously).

Before this change:

    Performance counter stats for './gap.master -A -q quit.g' (10 runs):

     760.858383      task-clock (msec)         #    0.999 CPUs utilized            ( +-  1.22% )
              7      context-switches          #    0.009 K/sec                    ( +- 12.04% )
              0      cpu-migrations            #    0.000 K/sec
            355      page-faults               #    0.467 K/sec                    ( +-  0.09% )
     2834967239      cycles                    #    3.726 GHz                      ( +-  0.38% )  (39.57%)
     1384871452      stalled-cycles-frontend   #   48.85% frontend cycles idle     ( +-  0.26% )  (39.68%)
      835157974      stalled-cycles-backend    #   29.46% backend  cycles idle     ( +-  0.26% )  (40.12%)
     3588976596      instructions              #    1.27  insns per cycle
                                               #    0.39  stalled cycles per insn  ( +-  0.42% )  (50.64%)
      834750192      branches                  # 1097.116 M/sec                    ( +-  0.32% )  (50.91%)
       15136518      branch-misses             #    1.81% of all branches          ( +-  0.57% )  (51.05%)
     1003981324      L1-dcache-loads           # 1319.538 M/sec                    ( +-  0.38% )  (49.19%)
       75716887      L1-dcache-load-misses     #    7.54% of all L1-dcache hits    ( +-  0.40% )  (20.19%)
       60158560      LLC-loads                 #   79.067 M/sec                    ( +-  1.44% )  (19.90%)
        1873602      LLC-load-misses           #    3.11% of all LL-cache hits     ( +- 10.08% )  (29.63%)

    0.761377984 seconds time elapsed                                          ( +-  1.22% )

After this change:

    Performance counter stats for './gap -A -q quit.g' (10 runs):

     718.481321      task-clock (msec)         #    0.999 CPUs utilized            ( +-  1.24% )
              8      context-switches          #    0.011 K/sec                    ( +- 11.44% )
              0      cpu-migrations            #    0.000 K/sec
            355      page-faults               #    0.494 K/sec                    ( +-  0.10% )
     2670255520      cycles                    #    3.717 GHz                      ( +-  0.11% )  (39.23%)
     1273776668      stalled-cycles-frontend   #   47.70% frontend cycles idle     ( +-  0.19% )  (39.36%)
      773319911      stalled-cycles-backend    #   28.96% backend  cycles idle     ( +-  0.45% )  (40.15%)
     3532240401      instructions              #    1.32  insns per cycle
                                               #    0.36  stalled cycles per insn  ( +-  0.10% )  (50.85%)
      809874472      branches                  # 1127.203 M/sec                    ( +-  0.18% )  (51.10%)
       12333874      branch-misses             #    1.52% of all branches          ( +-  0.29% )  (51.18%)
      960149151      L1-dcache-loads           # 1336.359 M/sec                    ( +-  0.14% )  (49.34%)
       65644690      L1-dcache-load-misses     #    6.84% of all L1-dcache hits    ( +-  0.73% )  (20.23%)
       50626980      LLC-loads                 #   70.464 M/sec                    ( +-  1.37% )  (19.81%)
        3477462      LLC-load-misses           #    6.87% of all LL-cache hits     ( +-  1.50% )  (29.43%)

    0.719034384 seconds time elapsed                                          ( +-  1.24% )

So we go from ~0.76 seconds to ~0.72 seconds startup time on average on this machine.

I also checked the testinstall testsuite, but less thoroughly, using

    perf stat -r 5 -d ./gap -A -q tst/testinstall.g

Before: ~56.408 (+- 0.15%) seconds time elapsed
After: ~56.280 (+- 0.15%) seconds time elapsed

Here it provides not much of an advantage; but it also doesn't cause a slow down.

I suspect we could speed flags up quite a bit more by making use of the fact that they are usually very sparse. E.g. I looked at the flag lists used for implications, and while they were up to ~2200 entries long or beyond, they had at most 90 bits set to true; but on average only 7.82, median 3.
```
gap> printStats:=function(vals)
>     Print("min ", Minimum(vals), " - avg ", Int(100*Average(vals))/100.,
>           " - med ", Median(vals), " - max ", Maximum(vals), "\n");
> end;;
gap> filters:=Concatenation(Flat(IMPLICATIONS_SIMPLE),Flat(IMPLICATIONS_COMPOSED));;
gap> printStats(List(filters, SIZE_FLAGS));
min 1 - avg 7.82 - med 3 - max 90
gap> printStats(List(filters, LEN_FLAGS));
min 1 - avg 1024.97 - med 1017 - max 2210
gap> printStats(List(filters, SIZE_OBJ));
min 40 - avg 164.15 - med 160 - max 312
```
Also note how the average flag object takes 160 bytes, so multiple flags together will quickly exceed the L1 cache.  Thus, perhaps with a more clever encoding (even a simple RLE scheme), we could gain a lot more, simply by virtue of it leading to much smaller flags objects which better fit into the cache. 

BTW, I also tried reversing the order we loop over memory in `IS_SUBSET_FLAGS`, thinking that it might help the CPU in predicting the memory access better, but this actually caused a slowdown. I have not yet investigate why, but I think part of the reason is that the flags we compare often are identical in the first few blocks.

### DIFF
--- a/lib/profile.g
+++ b/lib/profile.g
@@ -981,8 +981,6 @@ BIND_GLOBAL("DisplayCacheStats",function()
         "Operation TryNextMethod",
         "Operation cache misses",
         "IS_SUBSET_FLAGS calls",
-        "IS_SUBSET_FLAGS less trues",
-        "IS_SUBSET_FLAGS few trues",
         "WITH_HIDDEN_IMPS hits",
         "WITH_HIDDEN_IMPS misses",
         "WITH_IMPS hits",
--- a/src/opers.c
+++ b/src/opers.c
@@ -437,8 +437,6 @@ Obj FuncIS_EQUAL_FLAGS (
 
 #ifdef COUNT_OPERS
 static Int IsSubsetFlagsCalls;
-static Int IsSubsetFlagsCalls1;
-static Int IsSubsetFlagsCalls2;
 #endif
 
 /****************************************************************************
@@ -452,49 +450,10 @@ static Int IS_SUBSET_FLAGS(Obj flags1, Obj flags2)
     UInt * ptr1;
     UInt * ptr2;
     Int    i;
-    Obj    trues;
 
-/* do the real work                                                    */
 #ifdef COUNT_OPERS
     IsSubsetFlagsCalls++;
 #endif
-
-    /* first check the trues                                               */
-    trues = TRUES_FLAGS(flags2);
-    if (trues != 0) {
-        len2 = LEN_PLIST(trues);
-        if (TRUES_FLAGS(flags1) != 0) {
-            if (LEN_PLIST(TRUES_FLAGS(flags1)) < len2) {
-#ifdef COUNT_OPERS
-                IsSubsetFlagsCalls1++;
-#endif
-                return 0;
-            }
-        }
-        if (len2 == 0) {
-            return 1;
-        }
-
-        /* If flags2 has only a "few" set bits then the best way is to
-           simply check if those bits are set in flags1. The optimal
-           value of "few" depends on compilers, hardware and the
-           length of flags1. Experiments in 2017 suggest that it is
-           somewhere between 10 and 20 for current setups. */
-        if (len2 < 16) {
-#ifdef COUNT_OPERS
-            IsSubsetFlagsCalls2++;
-#endif
-            if (LEN_FLAGS(flags1) < INT_INTOBJ(ELM_PLIST(trues, len2))) {
-                return 0;
-            }
-            for (i = len2; 0 < i; i--) {
-                if (!C_ELM_FLAGS(flags1, INT_INTOBJ(ELM_PLIST(trues, i)))) {
-                    return 0;
-                }
-            }
-            return 1;
-        }
-    }
 
     /* compare the bit lists                                               */
     len1 = NRB_FLAGS(flags1);
@@ -3808,8 +3767,8 @@ Obj FuncOPERS_CACHE_INFO (
     Obj                 list;
     Int                 i;
 
-    list = NEW_PLIST_IMM(T_PLIST, 15);
-    SET_LEN_PLIST(list, 15);
+    list = NEW_PLIST_IMM(T_PLIST, 13);
+    SET_LEN_PLIST(list, 13);
 #ifdef COUNT_OPERS
     SET_ELM_PLIST(list, 1, INTOBJ_INT(AndFlagsCacheHit));
     SET_ELM_PLIST(list, 2, INTOBJ_INT(AndFlagsCacheMiss));
@@ -3818,12 +3777,10 @@ Obj FuncOPERS_CACHE_INFO (
     SET_ELM_PLIST(list, 5, INTOBJ_INT(OperationNext));
     SET_ELM_PLIST(list, 6, INTOBJ_INT(OperationMiss));
     SET_ELM_PLIST(list, 7, INTOBJ_INT(IsSubsetFlagsCalls));
-    SET_ELM_PLIST(list, 8, INTOBJ_INT(IsSubsetFlagsCalls1));
-    SET_ELM_PLIST(list, 9, INTOBJ_INT(IsSubsetFlagsCalls2));
-    SET_ELM_PLIST(list, 10, INTOBJ_INT(WITH_HIDDEN_IMPS_HIT));
-    SET_ELM_PLIST(list, 11, INTOBJ_INT(WITH_HIDDEN_IMPS_MISS));
-    SET_ELM_PLIST(list, 12, INTOBJ_INT(WITH_IMPS_FLAGS_HIT));
-    SET_ELM_PLIST(list, 13, INTOBJ_INT(WITH_IMPS_FLAGS_MISS));
+    SET_ELM_PLIST(list, 8, INTOBJ_INT(WITH_HIDDEN_IMPS_HIT));
+    SET_ELM_PLIST(list, 9, INTOBJ_INT(WITH_HIDDEN_IMPS_MISS));
+    SET_ELM_PLIST(list, 10, INTOBJ_INT(WITH_IMPS_FLAGS_HIT));
+    SET_ELM_PLIST(list, 11, INTOBJ_INT(WITH_IMPS_FLAGS_MISS));
     
     /* Now we need to convert the 3d matrix of cache hit counts (by
        precedence, location found and number of arguments) into a three
@@ -3846,7 +3803,7 @@ Obj FuncOPERS_CACHE_INFO (
                     INTOBJ_INT(CacheHitStatistics[i - 1][j - 1][k]));
         }
     }
-    SET_ELM_PLIST(list, 14, tensor);
+    SET_ELM_PLIST(list, 12, tensor);
     CHANGED_BAG(list);
 
     /* and similarly the 2D matrix of cache miss information (by
@@ -3862,10 +3819,10 @@ Obj FuncOPERS_CACHE_INFO (
             SET_ELM_PLIST(vec, k + 1,
                           INTOBJ_INT(CacheMissStatistics[j - 1][k]));
     }
-    SET_ELM_PLIST(list, 15, mat);
+    SET_ELM_PLIST(list, 13, mat);
     CHANGED_BAG(list);
 #else
-    for (i = 1; i <= 15; i++)
+    for (i = 1; i <= 13; i++)
         SET_ELM_PLIST(list, i, INTOBJ_INT(0));
 #endif
     return list;
@@ -3886,8 +3843,6 @@ Obj FuncCLEAR_CACHE_INFO (
     OperationHit = 0;
     OperationMiss = 0;
     IsSubsetFlagsCalls = 0;
-    IsSubsetFlagsCalls1 = 0;
-    IsSubsetFlagsCalls2 = 0;
     OperationNext = 0;
     WITH_HIDDEN_IMPS_HIT = 0;
     WITH_HIDDEN_IMPS_MISS = 0;
--- a/tst/testinstall/kernel/opers.tst
+++ b/tst/testinstall/kernel/opers.tst
@@ -272,7 +272,7 @@ Error, <obj> must be a component object
 #
 gap> CLEAR_CACHE_INFO();
 gap> OPERS_CACHE_INFO();
-[ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
+[ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
 
 #
 # method tracing


ChrisJefferson:
I seem to remember I previously flipped the order, for exactly the reason you describe :)

I didn't realise the first bit of code isn't actually required.

ChrisJefferson:
I dug up some code i had about this function. It is called
11,959,225 times during startup. Of those, 97% of the time we detect they are not subsets when we compare the last pair of blocks, with this PR. Therefore if you try adjusting, I'd still do that check first. Of the remaing 3%, about half pass and half fail.

Intersting thing I just noticed, if I add an assert just after 'if(len1 < len2)' with 'assert(ptr2[len2-1] != 0)', it passes -- do we promise not to put empty blocks on the end of BLOCKS_FLAGS? if so, the whole first branch can be reduced to 'return 0'. I suspect this won't greatly increase the speed as we are exiting almost straight away anyway.

fingolfin:
@ChrisJefferson We do not make any such promise, and it doesn't hold:
```
gap> f:=SUB_FLAGS(f,f);
<flag list>
gap> LEN_FLAGS(f);
112
gap> TRUES_FLAGS(f);
[  ]
```
But of course we could change the code to do so -- given that creating filters is done far less often than testing them, this might pay off... And actually, I think `SUB_FLAGS` is the only function where this can happen, and changing it to resize the flags bag to minimal size shouldn't be hard (and can be done quite efficiently).

fingolfin:
@stevelinton we do *NOT* guarantee "equal iff identical" for flag lists, and indeed, this is frequently violated.

Storing a list of 16 bit flag numbers is one of various possible alternate data formats, yes. For the worst case with 90 true flags, that would amount to 180 bytes (plus header), but of course the smaller ones are indeed much, much better.

If we always store the flags as short as possible, we can also get rid of the `LEN_FLAGS` field (also, on the GAP level, there is only one call to it in `lib/oper1.g:605` which can and should be changed to use `TRUES_FLAGS` instead -- after which we could get rid of both `FuncLEN_FLAGS` and `FuncELM_FLAGS`). We could then also get rid of the `TRUES_FLAGS` field, and instead return a fresh plist each time `TRUES_FLAGS` is called -- the only two performance critical places calling it are in `RunImmediateMethods`; but the first one of those is making  a shallow copy of it anyway, the second appends it to another list -- so we could remove `FuncTRUES_FLAGS` and replace it by a `FuncAPPEND_TRUES_FLAGS`.


Of course there are other possible formats: e.g. various variations of run length encoding, such as: store runs as signed bytes: for each byte `n`, if `n>=0` then this indicates `n+1` zero bits; otherwise, `-n` one bits. Since very often two adjacent bits are set (for properties), this can sometimes save a bit; but given the extreme sparsity of our flags in practice, I am not sure the extra complication would be worth it.

Just out of curiosity, I experimented a bit to confirm this; while I don't think this line of research is going to be fruitful, here's my test code snippet anyway, which takes a trues list `t` and "compresses" it into ranges:
```
gap> t:=TRUES_FLAGS(IMPLICATIONS_COMPOSED[1][1]);
[ 15, 35, 56, 69, 70, 88, 89, 90, 99, 102, 106, 110, 114, 140, 141, 235 ]
gap> r:=[];; i:=1;; while i <= Length(t) do n:=t[i]; m:=n; while i<Length(t) and m+1=t[i+1] do i:=i+1;m:=m+1; od; Add(r, [n..m]); i:=i+1; od; r;
[ [ 15 ], [ 35 ], [ 56 ], [ 69, 70 ], [ 88 .. 90 ], [ 99 ], [ 102 ], [ 106 ], [ 110 ], [ 114 ], [ 140, 141 ], [ 235 ] ]
gap> Length(t); Length(r);
16
12
```

frankluebeck:
I tried the patch on some machines. It is difficult to judge if there is a speedup (one is mainly measuring the strategy of the system to change the speed of CPU cores). But there is certainly no obvious slow down and therefore the simpler code is welcome. 

For a possible improvement I would also try to store arrays of 16-bit unsigned shorts. But how much memory is needed by all current bit lists in a typical GAP session? There is probably not much to save. And when `IS_SUBSET_FLAGS` returns `false` in most cases, this is probably often detected very early in the bit list comparison. So, it is not immediately clear, if a change will actually lead to a performance gain that justifies the efforts.

fingolfin:
Saving memory is not a goal here, I agree -- this is about making flag lists small enough so that they have a higher chance of being in the L1 cache. Alas, this may still not work out in any meaningful way, due to other effects (such as the GASMAN not grouping objects in a cache friendly way). 
So yeah, if somebody wants to experiment and try changing the storage format of flag lists, it's completely unclear if this will help at all. Only way to find out is to try it. 


Just for kicks, here is the start of my cachegrind results. First *without* this PR:
```
--------------------------------------------------------------------------------
I1 cache:         32768 B, 64 B, 8-way associative
D1 cache:         32768 B, 64 B, 8-way associative
LL cache:         10485760 B, 64 B, 20-way associative
Command:          ./gap.master -A -q quit.g
Data file:        cachegrind.out.24909.master
Events recorded:  Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Events shown:     Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Event sort order: Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Thresholds:       0.1 100 100 100 100 100 100 100 100
Include dirs:
User annotated:
Auto-annotation:  off

--------------------------------------------------------------------------------
           Ir       I1mr   ILmr            Dr       D1mr      DLmr          Dw      D1mw      DLmw
--------------------------------------------------------------------------------
3,613,035,706 13,923,961 21,309 1,021,732,154 61,000,899 2,517,007 445,845,927 7,213,518 5,019,542  PROGRAM TOTALS

--------------------------------------------------------------------------------
         Ir    I1mr ILmr         Dr       D1mr    DLmr         Dw      D1mw      DLmw  file:function
--------------------------------------------------------------------------------
359,809,432  51,348   18 86,666,182 26,764,264  33,193  3,539,855    13,166         0  /home/horn/Projekte/GAP/gap.github/src/opers.c:IS_SUBSET_FLAGS.isra.45
299,266,590  29,237   12 81,688,620        466       1 54,209,710     4,944        14  /home/horn/Projekte/GAP/gap.github/src/io.c:GET_NEXT_CHAR
193,997,545  50,778   27 89,672,817 14,609,529   5,752 13,396,187     8,721         0  /home/horn/Projekte/GAP/gap.github/src/opers.c:FuncWITH_HIDDEN_IMPS_FLAGS
112,910,815  18,172   31 26,425,083    215,831   4,094          0         0         0  /build/glibc-Cl5G7W/glibc-2.23/string/../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:__strcmp_sse2_unaligned
 97,023,472 141,857   56 18,139,140     46,577      34 19,105,455    45,735       160  /home/horn/Projekte/GAP/gap.github/src/scanner.c:NextSymbol
 88,098,006  29,190   19 27,275,276     30,514     129 18,091,395     2,526         0  /home/horn/Projekte/GAP/gap.github/src/funcs.c:EvalFunccall2args
 84,571,556  37,961   14 24,831,510      8,148      33  5,110,374     4,248         0  /home/horn/Projekte/GAP/gap.github/src/opers.c:FuncTRUES_FLAGS
 80,075,894  19,504    6  8,119,065          6       0  7,554,062        24         0  /home/horn/Projekte/GAP/gap.github/src/io.c:SKIP_TO_END_OF_LINE
```

Now *with* it:
```
--------------------------------------------------------------------------------
I1 cache:         32768 B, 64 B, 8-way associative
D1 cache:         32768 B, 64 B, 8-way associative
LL cache:         10485760 B, 64 B, 20-way associative
Command:          ./gap -A -q quit.g
Data file:        cachegrind.out.24856.with-opt
Events recorded:  Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Events shown:     Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Event sort order: Ir I1mr ILmr Dr D1mr DLmr Dw D1mw DLmw
Thresholds:       0.1 100 100 100 100 100 100 100 100
Include dirs:
User annotated:
Auto-annotation:  off

--------------------------------------------------------------------------------
           Ir       I1mr   ILmr          Dr       D1mr      DLmr          Dw      D1mw      DLmw
--------------------------------------------------------------------------------
3,554,055,636 13,880,048 21,283 970,812,364 53,505,613 2,514,018 442,308,267 7,200,922 5,020,506  PROGRAM TOTALS

--------------------------------------------------------------------------------
         Ir    I1mr ILmr         Dr       D1mr    DLmr         Dw      D1mw      DLmw  file:function
--------------------------------------------------------------------------------
332,012,960  32,379    9 70,778,359 24,135,781  26,788          0         0         0  /home/horn/Projekte/GAP/gap.github/src/opers.c:IS_SUBSET_FLAGS.isra.45
299,266,590  39,698   12 81,688,620        463       0 54,209,710     4,642        10  /home/horn/Projekte/GAP/gap.github/src/io.c:GET_NEXT_CHAR
182,108,623  65,521   35 64,521,795 14,697,037   5,779 13,384,790     8,177         1  /home/horn/Projekte/GAP/gap.github/src/opers.c:FuncWITH_HIDDEN_IMPS_FLAGS
113,324,084  20,724   29 26,502,953    215,514   4,011          0         0         0  /build/glibc-Cl5G7W/glibc-2.23/string/../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:__strcmp_sse2_unaligned
 97,023,472 139,745   56 18,139,140     43,933      32 19,105,455    46,595       162  /home/horn/Projekte/GAP/gap.github/src/scanner.c:NextSymbol
 88,098,006  25,919   19 27,275,276     31,635     134 18,091,395     2,179         0  /home/horn/Projekte/GAP/gap.github/src/funcs.c:EvalFunccall2args
 84,718,501  38,553   14 24,874,322      9,657      34  5,119,088     4,330         0  /home/horn/Projekte/GAP/gap.github/src/opers.c:FuncTRUES_FLAGS
 80,075,894  24,664    6  8,119,065          6       0  7,554,062        29         0  /home/horn/Projekte/GAP/gap.github/src/io.c:SKIP_TO_END_OF_LINE
```

So the main benefit of this PR is this: without it, we execute 359 million instruction reads (Ir)
during `IS_SUBSET_FLAGS`, with it just 332 million; and the number of data reads (Dr) goes down from 86 to 70 millions.
Yet the L1 cache misses shrink much less, from 26.8 down 24.1 million. The number of L3 misses
goes from 33193 down to 26788.

To put this a bit into perspective, starting `./gap -A` and then entering `DisplayCacheStats();` I get this (with the patch):
```
AND_FLAGS cache hits:                 7370
AND_FLAGS cache miss:                39136
AND_FLAGS cache losses:              24106
Operation L1 cache hits:             22788
Operation TryNextMethod:                 0
Operation cache misses:                367
IS_SUBSET_FLAGS calls:            14113016
WITH_HIDDEN_IMPS hits:               29767
WITH_HIDDEN_IMPS misses:              2968
WITH_IMPS hits:                        277
WITH_IMPS misses:                     1654
NEW_TYPE hits:                         180
NEW_TYPE misses:                       115
```
So 14 million calls to IS_SUBSET_FLAGS; so each averages 5 reads, and 1.7 L1 cache misses.

Interestingly we also save a bunch 20 million data reads in `FuncWITH_HIDDEN_IMPS_FLAGS`,
wich I don't understand.

I think for further improvements, one may have to think about ways to avoid `IS_SUBSET_FLAGS`
calls altogether.

And looking the cache misses in `FuncWITH_HIDDEN_IMPS_FLAGS`, I bet this is because we store the implications as lists of lists, which is very bad for locality. It'd be much better if we either stored them as a single list, with entries alternating between requirements and implication; or else, store it as two lists, with requirements and implications running "in parallel" in them.

fingolfin:
Actually, that last bit is nonsense: the *hidden* implications are already stored as an alternating list `HIDDEN_IMPS`, I confused this with the GAP level lists `IMPLICATIONS_COMPOSED` and `IMPLICATIONS_SIMPLE`.

stevelinton:
@fingolfin You are right about flags lists not being equal iff identical. However, I have still found essentially no code that modifies them in place. 

I'm working on experimental code to shift to a sorted list of UInt2s for the flags. It doesn't seem to be very hard, and I suspect it has the potential to speed up and simplify the code in a number of ways, not just cache misses.

fingolfin:
@stevelinton great. perhaps we should merge this and PR #2437 first, though?

Another opportunity for optimization is the call to `FuncAND_FLAGS` in `FuncWITH_HIDDEN_IMPS_FLAGS`. This is one place where we'd be better off to modify a flags list in place, until we are done. This would avoid creating 27398 flags objects during startup, assuming I counted right.

stevelinton:
@fingolfin Agreed to merge this when you are ready. 

I had spotted `WITH_HIDDEN_IMPS_FLAGS` as well. Should be possible to do it in my version by accumulating the list of all filters to be set, then sorting and eliminating duplicates.

PR #2437, opened 2018-05-03
Remove user accessible LEN_FLAGS and ELM_FLAGS functions

fingolfin:
See PR #2435

### DIFF
--- a/hpcgap/src/c_oper1.c
+++ b/hpcgap/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include <src/compiled.h>
-#define FILE_CRC  "-126023111"
+#define FILE_CRC  "-110560148"
 
 /* global variables used in handlers */
 static GVar G_REREADING;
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
If this is removed, is there another way of getting the list of flags in a FLAGS object (I've always used ELM_FLAGS before).

fingolfin:
@ChrisJefferson sure: `TRUES_FLAGS` gives you exactly that.

ChrisJefferson:
I've just noticed ELM_FLAGS doesn't do length checking, but it doesn't matter obviously, as it is being removed :)

ChrisJefferson:
I am tempted to leave for a little while, in case someone really wants it. It would be easy to add a GAP level version, if anyone does have code using it.

stevelinton:
LEN_FLAGS is meaningless after the kernel changes I am experimenting with, so withdrawing it at GAP level makes sense. It might still be useful to have a function that tests if a flags list includes a specific flag (which is actually ELM_FLAGS) without extracting the entire TRUES_FLAGS list, but it might need a better name.

fingolfin:
Well, `ELM_FLAGS` actually might be useful for a minor optimization to the handling of `IMPLICATIONS_SIMPLE`, but the time critical part of that is handled in the kernel anyway, so no real need for `ELM_FLAGS` on the library level from that point of view... That said, I am also fine with leaving in `ELM_FLAGS` and just removing `LEN_FLAGS`. (Or, if somebody, e.g. @frankluebeck or @ThomasBreuer , is really unhappy about removing them, we can also drop the second commit in this commit and just keep the first)

frankluebeck:
I'm happy with removing `LEN_FLAGS` and `ELM_FLAGS` from GAP level. Of course, if @stevelinton  or @ChrisJefferson may want to use `ELM_FLAGS` instead of `TRUES_FLAGS` it also would not hurt to keep that function.

PR #2438, opened 2018-05-03
Convert assert calls to GAP_ASSERT

ChrisJefferson:
This changes most of the asserts in GAP into GAP_ASSERTS.

This is the kind of PR which could end up causing a whole bunch of conflicts, so I don't mind waiting until we seem to be at a "quiet period", even waiting until near the 4.10 branch.

I left the `assert(0)` calls.

This does reduce the executable, and startup time, by about 1% for me (there is in practice a couple of asserts which hit often during startup).

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -76,12 +76,12 @@ Obj FilenameCache;
 /* TL: UInt OffsBodyCount = 0; */
 
 static inline void PushOffsBody( void ) {
-  assert(STATE(OffsBodyCount) <= MAX_FUNC_EXPR_NESTING-1);
+  GAP_ASSERT(STATE(OffsBodyCount) <= MAX_FUNC_EXPR_NESTING-1);
   STATE(OffsBodyStack)[STATE(OffsBodyCount)++] = STATE(OffsBody);
 }
 
 static inline void PopOffsBody( void ) {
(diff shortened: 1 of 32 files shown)


src/code.c
```diff
@@ -800,6 +804,7 @@ void CodeFuncExprEnd(UInt nr)
 
     /* get the function expression                                         */
     fexp = CURR_FUNC();
+    GAP_ASSERT(!STATE(LoopNesting));
```
fingolfin: Where is this coming from?

src/hpc/misc.c
```diff
@@ -75,7 +75,7 @@ static void MergeSortRecurse(char * data,
                              int (*lessThan)(const void * a, const void *))
 {
     UInt nleft, nright;
-    /* assert(count > 1); */
+    /* GAP_ASSERT(count > 1); */
```
fingolfin: Shouldn't we just remove this? Or else activate it?

ChrisJefferson:
I'm actually going to close this, and make it an issue. Already while writing it, I'm having conflict issues (which is where that first random thing came from, it was from a previous assert I had merged). I don't want to conflict a whole bunch of open PRs.

PR #2439, opened 2018-05-03
Tweak flag lists

fingolfin:
@stevelinton I made these changes before I learned you were looking into changing the flags format. I hope they won't conflict (too much) with your changes; ideally not at all, but we'll see.

Perhaps you can just base your work on this branch, otherwise, we'll figure out how to merge later on, no worries :-).


fingolfin:
The test failures in here will go once #2437 is merged.

src/opers.c
```diff
@@ -2180,6 +2148,8 @@ static ALWAYS_INLINE Obj DoOperationNArgs(Obj  oper,
             CacheMissStatistics[(prec >= CACHE_SIZE) ? CACHE_SIZE : prec]
                                [n]++;
         }
+        if (prec > 0)
```
stevelinton: If this fixes a bug should it have its own ticket and a regression test?

stevelinton: Looks good. Just FYI in my WIP change (C_)ELM_FLAGS  is automatically safe, but that's easy to deal with. LEN_FLAGS also goes.

stevelinton: The second of there is really a general comment. I put it in the wrong place.

fingolfin: True, `OperationNext` fix should go to a separate PR. Though I am not completely sure how to "regression test" it, as it only is used when `COUNT_OPERS` is `#defined`... I guess we'd have to set up another GAP build which enables it? But there are also several other #defines we never test right now, such as `CHECK_NDENSE_BAGS`, `DEBUG_MASTERPOINTERS`, `DEBUG_GASMAN_MARKING`, `DEBUG_RATIONALS`

PR #2441, opened 2018-05-04
Fix broken assert

ChrisJefferson:
This just fixes a typo in an assert.

I would backport it to 4.9 just because people who build GAP are likely to see it (I updated to gcc 7.3, which is when I started seeing it).


fingolfin:
Backported in 450c01c29b2b50ec568886aa90f6dd78f784ce23

PR #2443, opened 2018-05-05
Add Normal Backward and Forward History

xzcvczx:
Renamed existing search functions to DirHistorySearch
Added functions for DirHistory

Set Ctrl-p/n for DirHistory
changed keys to Ctrl-R/T for DirHistorySearch

bound up/down to Ctrl-R/T to avoid changing behavior

while Ctrl-R/T are not anything special on macOS (that i know of) i am not sure about other platforms. The Search functions are unbound by default in readline so i was unsure of the best keys to put them on.

### DIFF
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -504,8 +504,60 @@ end;
 GAPInfo.CommandLineEditFunctions.Functions.0 := 
                        GAPInfo.CommandLineEditFunctions.Functions.AddHistory;
 
-##  C-p: previous line starting like current before point
+# For those more used to the default readline history behavior
+##  C-p: previous history
 GAPInfo.CommandLineEditFunctions.Functions.BackwardHistory := function(l)
+  local hist, n;
+  if UserPreference("HistoryMaxLines") <= 0 then
+    return [];
+  fi;
+
+  hist := GAPInfo.History.Lines;
+  n := GAPInfo.History.Pos;
+
+  if n > 1 then
+    n := n - 1;
+    GAPInfo.History.Pos := n;
+    GAPInfo.History.Last := n;
+
+    return [1,Length(l[3]) + 1, hist[n], Length(hist[n])+1];
+  else
+    return [100];
+  fi;
+end;
+# bind to C-p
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Could you just explain a little bit what this is changing, from a user perspective?

xzcvczx:
by default in readline ctrl-p/n are set to just go back and forward through the history (regardless of where the cursor is or what has already been typed), GAP seems to use the functions history-search-forwards/backwards instead, this change would retain this behavior by default but allow the user to set the readline init line to return up/down to ctrl-p/n if they prefer what readline in other applications does by default.

fingolfin:
Hi, thank you for contribution, and the following explanation.

I'd like to point out that this kind of description really should be in the commit message. Ideally, there would also some comments in the code that point out what these functions do, and how one might use them. Otherwise I am afraid nobody will ever discover and use them, other than their author (and then we don't need this PR, as you could also have the changes in your `gap.ini`). And unused (as reported by code coverage) and undocumented code is likely to be removed these days.

frankluebeck:
A few comments:  
- I'm not in favour of changing the default behaviour of Ctrl-p and Ctrl-n in GAP (which is essentially the behaviour GAP had already 30 years ago).
- GAP is not using the history part of the readline library.
- I find the mentioned default behaviour of other programs which use readline (like bash) annoying.
- The choice of the Ctrl-R key in your patch does not work because Ctrl-R is later bound to a different function. 
- You can get the behaviour you want with unpatched GAP just by putting your cursor in the front of the new row.

To summarize: It is intentional that users can configure and overwrite key bindings to their like. In this case I suggest that you put the functions you want for Ctrl-p and Ctrl-n in your gaprc file. But I would not change this in the distributed GAP.

xzcvczx:
i am sorry i didn't realise there were any further questions

fingolfin:
Indeed, the questions / requests were only indirect: I pointed out various concerns avout this PR, with no reaction. Similar for PR #2410. I'll make sure to formulate these as explicit questions / requests next time, sorry for the miscommunication :-(

PR #1015, opened 2016-12-13
Quit evaluator if first symbol had syntax error 

markuspf:
Please make sure that this pull request:

- [x] is submitted to the correct branch (the stable branch is only for bugfixes)
- [x] contains an accurate description of changes for the release notes below
- [x] provides new tests or relies on existing ones
- [x] correctly refers to other issues and related pull requests

### Tick all what applies to this pull request

- [ ] Adds new features
- [ ] Improves and extends functionality
- [x] Fixes bugs that could lead to crashes
- [ ] Fixes bugs that could lead to incorrect results
- [ ] Fixes bugs that could lead to break loops

### Write below the description of changes (for the release notes)

This PR addresses #995.

If the parser encountered a syntax error in the first symbol it read, it did not abandon the expression it was trying to parse early enough. This lead to inconsistent state with the prompt `gap>` being printed, even though the interpreter is still waiting for the expression to finish parsing.

Issue #995, opened 2016-12-07
Parser bug on character constants


fingolfin:
I was looking at our old private tracker (we really should go through all issues there, and migrate any sensible issues to this tracker).

One of them was by @markuspf (see http://tracker.gap-system.org/issues/470) and reported the following buggy behavior: Enter a single quote, then press return twice and you get this:
```
gap> '
Syntax error: Newline not allowed in character literal
'
^
gap>
Syntax error: Missing single quote in character constant
```
So it look as if GAP did never finish parsing that character constant, despite already showing a syntax error (that first syntax error was not present in Markus' original report, so it has been added since then).

Apparently Markus even managed to get it to crash.

markuspf:
I thought we had fixed it back then. I will check this again, maybe we didn't fix it properly or merges between HPC-GAP and GAP re-broke it.

markuspf:
A similar problem occurs here:
```
gap> s := "abc
Syntax error: String must not include <newline>
s := "abc
        ^
> 
> ";
```

markuspf:
There are more instances of when we don't abandon the parse early when a syntax error occurs, like this here:
```
gap> f := function(abc);
> local x
Syntax error: end expected
local x
```

I will have a closer look at this this week.

ChrisJefferson:
@markuspf : There are many, many places where we don't abandon parsing when a syntax error occurs, that is I think "standard behaviour".

If we wanted to fix that, we should introduce some kind of general method of cleaning that up -- some kind of "empty out the input stream" (that needs planning better before we start it)

fingolfin:
This has been mostly resolved, except for that last instance @markuspf reported in December 2016.

@markuspf whatever happened to this?

alex-konovalov:
@fingolfin said
> I was looking at our old private tracker (we really should go through all issues there, and migrate any sensible issues to this tracker).

+10. To prevent flooding the tracker with mass issue migration, would be good if those who had created them would go through them themselves and either close or migrate them...

fingolfin:
@alex-konovalov note that the comment you are quoting is from Dec 7, 2016 ;-)

alex-konovalov:
@fingolfin figured that out 5 minutes ago ;-)

fingolfin:
Personally, I consider this fixed, now that PR #2445 is merged. People may still dislike aspects of the current behaviour, but I am not aware of anything actively buggy or misleading. If somebody is, please open a new issue.

PR #2445, opened 2018-05-06
Flush input line after early syntax error; some minor reader/scanner tweaks

fingolfin:
This fixes a mild regression introduce by PR #1015, which in turn attempted to fix parts of issue #995. With GAP 4.8 and also with this fix:

    gap> 1.2\3;
    1.2\3;
       ^

In GAP 4.9:

    gap> 1.2\3;
    1.2\3;
       ^
    3

Also added some tests for the other issues in #995, tweak two syntax error messages, and reorder some core for clarity.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -511,7 +511,7 @@ static LHSRef ReadSelector(TypSymbolSet follow, UInt level)
             ref.narg++;
         }
         if (ref.narg > 2) {
-          SyntaxError("[] only supports 1 or 2 indices");
+          SyntaxError("'[]' only supports 1 or 2 indices");
         }
         Match(S_RBRACK, "]", follow);
         ref.type = R_ELM_LIST;
@@ -2362,7 +2362,7 @@ static void ReadAtomic (
       nexprs ++;
 #ifdef HPCGAP
       if (nexprs > MAX_ATOMIC_OBJS) {
-        SyntaxError("atomic statement can have at most 256 objects to lock");
+        SyntaxError("'atomic' statement can have at most 256 objects to lock");
         return;
       }
 #endif
--- a/src/read.c
+++ b/src/read.c
@@ -2713,8 +2713,9 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
     /* get the first symbol from the input                                 */
     Match( STATE(Symbol), "", 0UL );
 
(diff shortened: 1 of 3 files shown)


fingolfin:
In the first iteration of this PR, I also had this change in:
```diff
diff --git a/src/read.c b/src/read.c
index dc237a8a3..b667573e9 100644
--- a/src/read.c
+++ b/src/read.c
@@ -2781,7 +2782,7 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
     }

     /* every statement must be terminated by a semicolon                  */
-    if (!IS_IN(STATE(Symbol), S_SEMICOLON) && STATE(Symbol) != S_HELP) {
+    if (!IS_IN(STATE(Symbol), S_SEMICOLON) && STATE(Symbol) != S_HELP && !STATE(NrError)) {
         SyntaxError( "; expected");
     }
```

The idea was / is that this should kill the second `SyntaxError` in e.g. this example (where I pressed ctrl-d on the continuation prompt):
```
gap> s:="
Syntax error: String must not include <newline>
s:="
   ^
>
Syntax error: ; expected
^
```

The reason this shows up is that after the scanner determined in `GetStr` that there was a syntax error,  control resumes in `ReadLiteral`, here:
```C
    case S_STRING:
        GAP_ASSERT(STATE(ValueObj) != 0);
        TRY_READ { IntrStringExpr(STATE(ValueObj)); }
        Match(S_STRING, "", follow);    //  <<--- we got an "incomplete" string and continue here...
        STATE(ValueObj) = 0;
        break;
```
But since we reached the end of the line, when `Match` calls `NextSymbol`, that in turn will call `GET_NEXT_CHAR`, and we end up in `GetLine`, which prompts the user for more input. Since we are technically in a new line of input now, the logic which suppresses syntax errors after the first in each line does not kick in, and we see the second syntax error (also visible in the `read.tst` change in this PR).

The above diff suppressed that, but it is a gross hack. A proper solution would be to bail out instead of getting more data, if we saw a syntax error before. BUT while that seems natural for user input (to me, at least, w/o having though tooo much about it), it might be somewhat risky for reading files, where GAP so far has been rather resilient, skipping over local errors and doing a best effort to read on afterwards...

Thoughts?

ChrisJefferson:
I have seen people rely on GAP reading a file with errors, and still successfully reading parts of the files which don't have errors, so I'd be worried about breaking that (even though.. it's really silly to have a file with errors :)

PR #2446, opened 2018-05-06
kernel: remove GetCleanedChar from scanner

fingolfin:
This is part of PR #2371. It simplifies the `GetNumber` code in the scanner a bit, which helps with further refactoring. I suggest viewing the changes with "Hide whitespace changes" enabled under "Diff settings". 

This causes one user visible changes: Before the PR, we have this on `master`:
```
gap> 12.34\56;
12.34\56;
     ^
56
```
Note the 56 at the end: This is because `GetCleanedChar` gobbles up the backslash; `GetNumber` bails out; and as we resume scanning, the scanner sees `56` in `STATE(In)`. With this commit, the backslash is *not* eaten, and so we get this, as the scanner sees `\56`:
```
gap> 12.34\56;
Syntax error: Badly Formed Number
12.34\56;
    ^
Error, Variable: '56' must have a value
```

Note that both are regressions in GAP 4.9 compared to GAP 4.8. There, we get this:
```
gap> 12.34\56;
Syntax error: Badly Formed Number
12.34\56;
    ^
```
This regressions is fixed by #2445, and with it, I think this patch has no user visible effect (or if it has, then I am confident it still is "safe").

I think `GetCleanedChar` was introduced by @stevelinton back then for two reasons: one, to handle line continuations, but that need was removed recently. Two, for handling identifiers -- but this, too, was recently moved out, we now pass control to `GetIdent` in that case (which also ensure identical processing). But perhaps I am missing something else.... Steve?

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -341,35 +341,11 @@ static void GetIdent(Int i)
 **  exponent digit.
 **
 */
-static Char GetCleanedChar(UInt * wasEscaped)
-{
-    Char c = GET_NEXT_CHAR();
-    *wasEscaped = 0;
-    if (c == '\\') {
-        c = GET_NEXT_CHAR();
-        *wasEscaped = 1;
-        switch (c) {
-        case 'n':  return '\n';
-        case 't':  return '\t';
-        case 'r':  return '\r';
-        case 'b':  return '\b';
-        case '>':  return '\01';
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
I'm happy with this, and I'm not too concerned about exactly what is produced when you put in a weird slash like that :)

PR #2448, opened 2018-05-08
kernel: turn lvars macros into functions

fingolfin:


src/vars.h
```diff
@@ -171,11 +192,12 @@ void FreeLVarsBag(Bag bag);
 *F  MakeHighVars( <bag> ) . . turn all frames on the stack into high vars
 */
 
-static inline void MakeHighVars( Bag bag ) {
-  while (bag && TNUM_OBJ(bag) == T_LVARS) {
-    RetypeBag(bag, T_HVARS);
-    bag = PARENT_LVARS(bag);
-  }
```
ChrisJefferson: Nothing seems to have happened here?

src/vars.c
```diff
@@ -2533,15 +2534,15 @@ void VarsAfterCollectBags ( void )
...
 {
-  UInt len,i;
-  const Obj *ptr;
-  SaveSubObj(FUNC_LVARS(lvars));
-  SaveUInt(STAT_LVARS(lvars));
-  SaveSubObj(PARENT_LVARS(lvars));
-  len = (SIZE_OBJ(lvars) - (2*sizeof(Obj)+sizeof(UInt)))/sizeof(Obj);
-  ptr = CONST_ADDR_OBJ(lvars)+3;
-  for (i = 0; i < len; i++)
-    SaveSubObj(*ptr++);
```
ChrisJefferson: What's actually happened here? Nothing?

src/vars.h
```diff
@@ -278,8 +301,15 @@ static inline void SwitchToOldLVarsAndFree(Obj old, const char *file, int line)
...
-#define ASS_LVAR(lvar,val)      (STATE(PtrLVars)[(lvar)+2] = (val))
-#define ASS_LVAR_WITH_CONTEXT(context,lvar,val)      (ADDR_OBJ(context)[(lvar)+2] = (val))
+static inline void ASS_LVAR(UInt lvar, Obj val)
+{
+    STATE(PtrLVars)[lvar + 2] = val;
+}
+
+static inline void ASS_LVAR_WITH_CONTEXT(Obj context, UInt lvar, Obj val)
+{
+    ADDR_OBJ(context)[lvar + 2] = val;
```
ssiccha: Could this in principle also be a `CONST_ADDR_OBJ` like in `OBJ_LVAR_WITH_CONTEXT` below?

src/vars.h
```diff
@@ -145,7 +166,7 @@ static inline void SetBrkCallTo( Expr expr, const char * file, int line ) {
             (int)expr, (int)STATE(CurrLVars), file, line);
   }
 #endif
-  STAT_LVARS_PTR(STATE(PtrLVars)) = expr;
+  ((LVarsHeader *)STATE(PtrLVars))->stat = expr;
```
ssiccha: Would it make sense to have an eg `SET_STAT_LVARS_PTR` to use instead of this line?

src/vars.h
```diff
@@ -298,8 +335,15 @@ static inline void SwitchToOldLVarsAndFree(Obj old, const char *file, int line)
...
-#define NAME_LVAR(lvar)         NAMI_FUNC( CURR_FUNC(), lvar )
-#define NAME_LVAR_WITH_CONTEXT(context,lvar)         NAMI_FUNC( FUNC_LVARS(context), lvar )
+static inline Obj NAME_LVAR(UInt lvar)
+{
+    return NAMI_FUNC(CURR_FUNC(), lvar);
+}
+
+static inline Obj NAME_LVAR_WITH_CONTEXT(Obj context, UInt lvar)
+{
+    return NAMI_FUNC(FUNC_LVARS(context), lvar);
```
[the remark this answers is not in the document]
fingolfin: Not really, as `NAMI_FUNC` already takes care of that (ultimately via `CONST_FUNC`).

ChrisJefferson:
I'm happy with all the technical changes here, but I don't really like the two functions that were whitespace reformatted. At the least those two should be in a seperate commit (so it's clear in history that commit doesn't do anything), but I'd prefer to not reformat them at all.

fingolfin:
@ChrisJefferson I already removed those...

fingolfin:
I just added a second commit here which changes `compiled.h` to use `ASS_LVAR` / `OBJ_LVAR` and their hvar counterparts.

src/vars.h
ssiccha: Ah, I thought it would also make it easier to grep for. But  one can also just grep for `STATE(PtrLVars))->stat`

PR #2449, opened 2018-05-08
Fix crash when returning from "method not found" break loop

### LINKED ISSUE #2447: GAP crashes when returning from 'method no found' errors.
Example:

```
gap> f := function( a ) return a + a; end;

function( a ) ... end
gap> f( () );
Error, no 1st choice method found for `+' on 2 arguments at /mnt/c/Users/caj/files/reps/gap/gap/lib/methsel2.g:250 called from

a + a at *stdin*:2 called from
<function "f">( <arguments> )
you can 'quit;' to quit to outer loop, or you can 'return;' to continue
brk> return;
<segfault>
```

The problem seems to be around opers.c:2178, where we call HandleMethodNotFound, and don't check the return value isn't 0. In fact, I think (from my brief usage), it's impossible to actually return a function from CallHandleMethodNotFound, so this should instead just error out, rather than try to loop (but that should be checked).

Git blame suggests @stevelinton was last around this code, but it could be a combination of various function related changes. It did work in 4.8.

fingolfin:
Fixes #2447 

This should be backported to stable-4.9. I am not sure how far the 4.9.1 release process is -- if there is still time, we might even want to squeeze it in there. If not, then it should go into GAP 4.9.2, and we should consider documenting this as a "known regression" in GAP 4.9.1.

### DIFF
--- a/src/opers.c
+++ b/src/opers.c
@@ -2115,10 +2115,6 @@ static ALWAYS_INLINE Obj DoOperationNArgs(Obj  oper,
                 CacheMethod(oper, n, prec, ids, method);
         }
 
-        if (!method) {
-            ErrorQuit("no method returned", 0L, 0L);
-        }
-
         /* If there was no method found, then pass the information needed
            for the error reporting. This function rarely returns */
         if (method == Fail) {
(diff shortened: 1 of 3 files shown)


fingolfin:
Backported to stable-4.9 via b5c0c786a2527d5d486c02619b1f3b6a3ac5cdb6

PR #2451, opened 2018-05-09
More tests; fix DisplayInformationPerfectGroups(1)

fingolfin:
This is primarily meant to increase coverage; but as usual, also an (unimportant) bug was uncovered by it.


ssiccha:
In `grp/imf.gi`it looks like it's easy to add coverage for lines 77, 81, 104 and also line 462.

fingolfin:
Yup, updated the PR now to hopefully catch those lines, too (I did not bother to try to get everything on the first round, as there were big chunks of uncovered code -- it's easier to just add some test, then let codecov run and see what's left, and then sniper attack those)

ssiccha:
Ah, I see. :)

PR #2457, opened 2018-05-11
WIP: Reformat Flags Lists

stevelinton:
As discussed elsewhere, this is an experimental PR which stores flags lists as lists of 16 bit indices of the set bits, rather than as a long bit vector. In this version, flags lists actually behave as kernel lists, equivalent to their TRUES_FLAGS so TRUES_FLAG is just x->x. The hope is that this is more cache-friendly and causes less garbage collection. At present, my quick experiments suggest that it slows down testinstall by a few percent, but reduces time spent in garbage collection.

### DIFF
--- a/lib/oper1.g
+++ b/lib/oper1.g
@@ -35,6 +35,7 @@ BIND_GLOBAL( "RunImmediateMethods", function ( obj, flags )
             meth,
             res,        # result of an immediate method
             loc,
+            flag,
             newflags;   # newly  found filters
 
     # Avoid recursive calls from inside a setter,
@@ -47,7 +48,10 @@ BIND_GLOBAL( "RunImmediateMethods", function ( obj, flags )
     if IS_SUBSET_FLAGS( IMM_FLAGS, flags ) then return; fi;
     flags := SUB_FLAGS( flags, IMM_FLAGS );
 
-    flagspos := SHALLOW_COPY_OBJ(TRUES_FLAGS(flags));
+    flagspos := [];
+    for flag in flags do
+        ADD_LIST(flagspos,flag);
+    od;    
     tried    := [];
     type     := TYPE_OBJ( obj );
     flags    := type![2];
@@ -111,9 +115,9 @@ BIND_GLOBAL( "RunImmediateMethods", function ( obj, flags )
 
                           newflags := SUB_FLAGS( type![2], IMM_FLAGS );
                           newflags := SUB_FLAGS( newflags, flags );
-                          APPEND_LIST_INTR( flagspos,
-                                            TRUES_FLAGS( newflags ) );
-
+                          for flag in newflags do
+                              ADD_LIST(flagspos, flag);
+                          od;
                           flags := type![2];
 
                         fi;
--- a/src/lists.c
+++ b/src/lists.c
@@ -2610,6 +2610,7 @@ static Int InitKernel (
         }
     }
 
+    SetupFlagsListsAsLists();
     /* return success                                                      */
     return 0;
 }
--- a/src/opers.c
+++ b/src/opers.c
@@ -126,6 +126,32 @@ void LoadFlags(
 
 /****************************************************************************
 **
+*F  C_ELM_FLAGS( <list>, <pos> )  . . . . . . . . . . . element of a flags list
+**
+**
+**  returns 1 if the flags list <list> contains the filter <pos>, else 0
+*/
+UInt C_ELM_FLAGS(Obj list, UInt pos) {
+    UInt len = INT_INTOBJ(SIZE_FLAGS(list));
+    Int hi = len -1;
+    Int lo = 0;
+    while (hi >= lo) {
+        Int mid = (hi + lo)/2;
+        UInt2 x = TRUE_FLAGS(list, mid);
+        if (x == pos)
+            return 1;
+        if (x < pos) {
+            lo = mid+1;
+        } else {
+            hi = mid -1;
+        }
+    }
+    return 0;
+}
+
+
+/****************************************************************************
+**
 *F * * * * * * * * * * * * *  GAP flags functions * * * * * * * * * * * * * *
 */
 
@@ -388,7 +414,7 @@ Obj FuncSUB_FLAGS (
 
     len1 = INT_INTOBJ(SIZE_FLAGS(flags1));
     len2 = INT_INTOBJ(SIZE_FLAGS(flags2));
-    NEW_FLAGS(flags, len1);
+    flags = NEW_FLAGS(len1);
     i = 0;
     j = 0;
     k = 0;
@@ -464,9 +490,9 @@ Obj FuncAND_FLAGS (
 
     if (flags1 == flags2)
         return flags1;
-    if (LEN_FLAGS(flags2) == 0)
+    if (SIZE_FLAGS(flags2) == INTOBJ_INT(0))
         return flags1;
-    if (LEN_FLAGS(flags1) == 0)
+    if (SIZE_FLAGS(flags1) == INTOBJ_INT(0))
         return flags2;
 
     // check the cache
@@ -526,7 +552,7 @@ Obj FuncAND_FLAGS (
     /* do the real work                                                    */
     len1 = INT_INTOBJ(SIZE_FLAGS(flags1));
     len2 = INT_INTOBJ(SIZE_FLAGS(flags2));
-    NEW_FLAGS(flags, len1+len2);
+    flags = NEW_FLAGS(len1+len2);
     i = 0;
     j = 0;
     k = 0;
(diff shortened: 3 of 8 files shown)


src/opers.c
```diff
@@ -394,43 +365,43 @@ static Int IsSubsetFlagsCalls;
...
-                return 0;
+    len1 = SIZE_FLAGS(flags1);
+    len2 = SIZE_FLAGS(flags2);
+    if (len2 > len1)
+        return 0;
+    if (len2== 0)
+        return 1;
+    UInt i = 0;
+    UInt j = 0;
+    UInt2 x,y;
```
fingolfin: add space after `,` (and/or use `git clang-format` on this PR)

src/opers.c
```diff
@@ -394,43 +365,43 @@ static Int IsSubsetFlagsCalls;
...
+    while (i < len1 && j < len2) {
+        x = trues1[i];
+        y = trues2[j];
+        if (x == y) {
+            i++;
+            j++;
+        } else if (y < x) {
+            return 0;
+        } else {
+            i++;
```
fingolfin: So this keeps re-reading the same `trues2[j]` value -- probably fine, as it'll be in the L1 cache. And perhaps
the compiler will super optimize it anyway? But I still wonder how it compares to this:
```C
while (i < len1 && j < len2) {
  y = trues2[j++];
  do {
    x = trues1[i++];
  while (y > x);
  if (y < x)
    return 0;
}
```

src/opers.h
```diff
@@ -221,17 +222,35 @@ static inline void SET_ENABLED_ATTR(Obj oper, Int x)
 /****************************************************************************
 **
 *F * * * * * * * * * * * * internal flags functions * * * * * * * * * * * * *
+**
+** Attempting change April 2018. Flags will be stored as 
+** AND_CACHE -- Obj
+** Size (UInt)  -- could possibly be a UInt2 instead
+** Hash (Uint)
+** Set positions (array of UInt2).
```
fingolfin: What does `Set positions` mean? `Set` as in the verb? Or the mathematical object? also, how are they stored? In ascending order?

src/opers.h
```diff
@@ -221,17 +222,35 @@ static inline void SET_ENABLED_ATTR(Obj oper, Int x)
...
 */
 
+static inline UInt IS_FLAGS(Obj f) {
```
fingolfin: The `{` should be on its own line (use `clang-format`?)

src/opers.h
```diff
@@ -242,98 +261,82 @@ static inline Obj NEW_FLAGS(UInt len)
...
-*F  MASK_POS_FLAGS( <pos> ) . . .  . .  bit mask for position of a flags list
-**
-**  'MASK_POS_FLAGS(<pos>)' returns a UInt with a single set  bit in position
-**  '(<pos>-1) % BIPEB',
-**  useful for accessing the <pos>-th element of a 'FLAGS' list.
-**
-**  Note that 'MASK_POS_FLAGS'  is a macro, so  do not call it with arguments
-**  that have side effects.
+*F  SET_TRUE_FLAGS( <flags>, <ix>, <filt> )  the <ix>th set position in flags
+**                               caller is responsible for <ix> being in range
```
fingolfin: What does that mean? I.e. what is "in range" in the new model? I assume it means that it doesn't leave the bag -- OK, so let's add a `GAP_ASSERT` to check that anyway?

src/opers.h
```diff
@@ -221,17 +222,35 @@ static inline void SET_ENABLED_ATTR(Obj oper, Int x)
 /****************************************************************************
 **
 *F * * * * * * * * * * * * internal flags functions * * * * * * * * * * * * *
+**
+** Attempting change April 2018. Flags will be stored as
```
fingolfin: We could introduce a `struct FlagsHeader` for this, just like we did for other types by now. That's self-documenting and can be used to make some of the accessor functions (well, all but the one for the TRUES_FLAGS) more readable.

lib/oper1.g
```diff
@@ -47,7 +48,10 @@ BIND_GLOBAL( "RunImmediateMethods", function ( obj, flags )
     if IS_SUBSET_FLAGS( IMM_FLAGS, flags ) then return; fi;
     flags := SUB_FLAGS( flags, IMM_FLAGS );
 
-    flagspos := SHALLOW_COPY_OBJ(TRUES_FLAGS(flags));
+    flagspos := [];
+    for flag in flags do
+        ADD_LIST(flagspos,flag);
+    od;
```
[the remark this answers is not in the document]
stevelinton: There's no skipping, but it would obviously be better to put that loop into the kernel. APPEND_TRUES_FLAG is the nice answer, I think.

src/lists.c
```diff
@@ -2610,6 +2610,7 @@ static Int InitKernel (
         }
     }
 
+    SetupFlagsListsAsLists();
```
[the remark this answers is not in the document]
stevelinton: Adding the API turned out to be a bit more trouble than I expected. I was trying to avoid storing TRUES_FLAG as a PLIST (waste of memory) or computing it on the fly each time (waste of CPU), so I thought "why not simply let the flags list (which is simply a packed version of its TRUES_FLAGS now) just be the list".  It's not a lot of code in the end, but it's a lot  of trivial functions and table entries.

src/opers.c
```diff
@@ -146,135 +170,73 @@ void LoadFlags(
 **  the lower addressed half-word is the less significant
 **
 */
-#define HASH_FLAGS_SIZE (Int4)67108879L
+#define fnvp 1099511628211ULL
+#define fnvob 14695981039346656037ULL
```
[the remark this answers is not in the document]
stevelinton: It's called the FNV hash. There will be some cost on a 32 bit system, but it's still only about 4 multiplications per byte hashed.  I'm not sure what happens if you are doing a 32 bit compile on a 64 bit capacble processor. The compiler mgiht be able to generate better code.

We can swap it out. Any usable non-cryptographic hash is fine. I like this one because it's simple.

src/opers.c
```diff
@@ -146,135 +170,73 @@ void LoadFlags(
...
-        hash = (hash + (*ptr % HASH_FLAGS_SIZE) * x) % HASH_FLAGS_SIZE;
-        x    = ((8*sizeof(UInt4)-1) * x) % HASH_FLAGS_SIZE;
-        ptr++;
+
+    UInt8 h = fnvob;
+    UInt len = SIZE_FLAGS(flags);
+    for (UInt i = 0; i < len; i++) {
+        UInt2 pos = TRUE_FLAGS(flags, i);
+        h = (h*fnvp) ^ (pos & 0xFF);
+        h = (h*fnvp) ^ (pos >>8);
```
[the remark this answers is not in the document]
stevelinton: No probably not. I think I just didn't find the kernel interface for that.

src/opers.c
ChrisJefferson: While this can wait, I'd try running this loop backwards -- flag lists often disagree at the end more than the beginning.

src/lists.c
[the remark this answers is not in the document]
fingolfin: So, what is gained by all these complications? I mean, for most uses, the overhead of `TRUES_FLAG` (be it computational or memory wise) is irrelevant (e.g. when viewing flags). The only really crucial places using it is `RunImmediates`, which ironically is the one place in this PR where the overhead is not really avoided ;-) . Well, perhaps `RankFilter` also is performance critically, but it could easily be reimplemented in C, too.

So unless there is some clear win in having this list interface for flags, I'd rather remove it again, and keep this PR small(er)...

And I am really curious if implementing and using `APPEND_TRUES_FLAG` helps at all... and whether perhaps we should reimplement `RunImmediates` in C anyway...

BTW, did you perform any kind of profiling before/after to see where the bottlenecks are resp. what shifted?

src/opers.c
stevelinton: That form might be better, but, as you say rereading from L1 doesn't seem to be much trouble. I hadn't seen the two loop version.

src/opers.h
stevelinton: Will imptove comment

fingolfin:
BTW, thanks for working on this, I appreciate it!

src/lists.c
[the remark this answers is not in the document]
stevelinton: I haven't done any profiling yet. Ran out of time. If you're set up to get cachegrind output for startup, that would save me the trouble of creating it.

Reworking RunImmediates and WITH_IMPS_FLAGS to be sensible implementations given this data structure was next on the agenda.

APPEND_TRUES_FLAG does seem like a potentially better answer, though, in which case much of this PR can go.  

It would be quite nice to get TRUES_FLAG off the critical path so that we don't have to cache the result, though.

stevelinton:
Superceded by #2462

PR #2462, opened 2018-05-14
WIP: Reformat Flags Lists 2

stevelinton:
This is a heavily reworked version of #2457 which avoids all the complication of making flags bags be kernel lists by adding a few more GAP callable functions. Also does some code tidying.

The changes to opers.c and opers.h were sufficiently wide-ranging that I have clang-formatted the entire files.

### DIFF
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -539,6 +539,7 @@ BIND_GLOBAL( "InstallSubsetMaintenance",
           triple,         # loop over `SUBSET_MAINTAINED_INFO[2]'
           req,
           flag,
+          flags,
           filt1,
           filt2,
           i;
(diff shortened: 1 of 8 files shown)


stevelinton:
So I've been profiling with cachegrind and other tools. The bottom line seems to be that this change definitely reduces memory allocation and L1 data cache misses, but nevertheless slows down the system significantly, which I suspect is the fault of the unpredictable branches in IS_SUBSET_FLAGS (and to a lesser extent other functions). Depressing.

ChrisJefferson:
I wondered if the problem was deferencing the pointer in TRUES_FLAGS, but if I change the function to use the raw pointers, it doesn't help.

stevelinton:
I've poked around with cachegrind some more, and looked at the generated code for the critical routines, and I'm pretty sure the problem is simply the time taken to do IS_SUBSET_FLAGS this way, including all the branches, versus  just doing bit operations 64 at a time. 

The calls to IS_SUBSET_FLAGS mostly come from WITH_IMPS_FLAGS and WITH_HIDDEN_IMPS_FLAGS, so one approach would be to rewrite those to build and use some much more efficient data structure to represent the whole set of implications (since it changes very rarely, at least after startup). I haven't worked through it thoroughly, but I suspect you could compile some kind of FSA that read the sequence of filters in a flags list and triggered whenever there was an implication to apply. However then I realised this is really not that important, except possibly as an interesting data structures & algorithms exercise, so I stopped.

PR #2465, opened 2018-05-15
gap.sh should respect the GAP_DIR environment var

RussWoodroofe:
In previous versions of GAP, if the GAP_DIR environment variable was set, then GAP used this variable to determine the location of GAP.  (Otherwise, it uses sensible defaults.)  This patch returns to this behavior.

### DIFF
--- a/cnf/compat/gap.sh.in
+++ b/cnf/compat/gap.sh.in
@@ -1,9 +1,9 @@
 #!/bin/sh
 
 GAP_EXE=$GAP_DIR
-if [ "x$GAP_DIR" = "x" ];  then
-GAP_DIR=$(cd "@abs_top_srcdir@" && pwd)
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Out of interest, have you tried just running the GAP executable directly? The `-l` option is now optional, as GAP will find it's location, and the location of the library/packages, from the location of the GAP executable.

RussWoodroofe:
Apologies for taking a day or two to get back.  I made the requested formatting updates.  Backporting would indeed be sensibleshould I have done something different when submitting the pull request to indicate this?

ChrisJefferson:
@RussWoodroofe : We have a tag for backporting, but discussing it in the PR is fine!

fingolfin:
Backported to stable-4.9 via 20dd8f19a077ef036fdee15d6683395b9ec5be9a

PR #2466, opened 2018-05-15
More debugging

ChrisJefferson:
This adds some more debugging checks, by enabling some pre-existing debugging checks in GAP. This is part of me searching down a bug I haven't found yet, but I think it's generally useful.

In case anyone wonders "why not also enable debug flag X?", most flags have rotted and require some fixing. I'm working through them, but thought I'd do this set as they aren't too complicated to fix.


fingolfin:
Now conflicts in the headers, sorry for that. (Also, please add labels)

ChrisJefferson:
After tracking down a couple of places which failed the "NonDensePlist" test, I have instead deleted it, to avoid future confusion. Also pulled removed all uses of <stdio.h> by putting it in system.h

PR #2472, opened 2018-05-16
More tests

fingolfin:


fingolfin:
Indeed, they were! Thanks for carefully checking this, I now removed the spurious `clasmax.grp` (what happened is this: in the past, I submitted a PR with `tst/testinstall/grp/clasmax.grp` used (incorrectly) as filename for a `.tst` file; I fixed this in that PR before it was merged, but did that on a computer which is not my main work computer.
Then, back on my laptop, the old version of that branch was still around, and also was not cleaned up by my `git obsolete` tool (which deleted merged branches) -- after all, it indeed hadn't been merged (a fixed version was merged).
Today, I went through some of my old branches, and polished them up to become PRs, and I stumbled again about this old (bad) branch, not realizing that it had of course already been merged in fixed form.

So, that's how it happened. I dropped the "duplicate commit", and instead added a commit which fixes some typos ;-)

tst/testinstall/ffe.tst
```diff
@@ -118,6 +130,62 @@ gap> Subfields( GF(81) );
...
+true
+gap> ForAll(odds, x -> Z(2) * x = Z(2));
+true
+gap> ForAll(evens, x -> x * Z(2) = 0*Z(2));
+true
+gap> ForAll(evens, x -> Z(2) * x = 0*Z(2));
+true
+
+#
+gap> LargeGaloisField(4,1);
```
stevelinton: Is LargeGaloisField intended to "public". As far as I can see it's called from GaloisField when needed and it's not documented.  It looks to me like it's part of the implementation of GaloisField, and should be tested as such, in case someone wants to change that implementation.

fingolfin: `LargeGaloisField` is not "public", but I also don't think this matter: we test many non-public functions in our .tst files, and I think this is well and indeed required in many cases. Of course if the behavior of `LargeGaloisField` were to be changed, then now one needs to adjust the tests -- but I don't think that's a major hurdle (and very unlikely to happen anyway).

Many people would argue that having (unit) tests for virtually every function in your program, including all "internal" ones, is a good thing, because despite being undocumented to the public, they still have an *internal* API, and we are very much interested in catching regressions or other bugs in them.

stevelinton: OK. Fair enough.

PR #2473, opened 2018-05-16
io: only increment line number for LF, not for CR

fingolfin:
This avoids counting lines terminated by CRLF twice, fixing a very obscure bug.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1228,8 +1228,7 @@ Char GetLine ( void )
     }
 
     /* bump the line number                                                */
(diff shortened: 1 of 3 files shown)


stevelinton:
Does this cause problems on MacOS 9? Even if it does, do we care?

fingolfin:
I have no idea, but we definitely do not care, as we removed all code for supporting classic Mac OS (so 9 and older) many years ago, and this bit would be the list of our concerns if were to support a system with CR line ends again...

PR #2474, opened 2018-05-17
Improve Is{Constant,Univariate}RationalFunction, add tests

fingolfin:
* Change IsConstantRationalFunction and IsUnivariateRationalFunction to
  return false for objects which are not even rational functions
* Improve UNIVARTEST_RATFUN to deal with some more "trivial" cases before
  giving up
* Add convenience helpers for multiplying rational numbers by rational
  functions over rings which don't contain the rational numbers,
  complementing the existing helpers doing this for addition
* Add tests

### DIFF
--- a/hpcgap/lib/ratfun.gi
+++ b/hpcgap/lib/ratfun.gi
@@ -249,6 +249,26 @@ function(o)
   fi;
 end);
 
+InstallOtherMethod(IsConstantRationalFunction,"fallback for non-ratfun",true,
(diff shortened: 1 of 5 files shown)


fingolfin:
I just extended this some more, adding helpers for multiplying rational numbers with rational functions over e.g. finite fields (we already have such helpers for addition).

ssiccha:
@fingolfin The helper functions were not added to the HPC-GAP folder. :)

PR #2475, opened 2018-05-17
Teach BuildPackages.sh how to build simpcomp

fingolfin:
... by working around missing executable bits, and moving the bistellar executable into a place where GAP can find it

I would suggest backporting this to stable-4.9. Even if they release a version with fixed, this patch is still useful, as it prevents the script from passing `--with-gaproot` to the simpcomp `configure`, which does not support it.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -214,6 +214,10 @@ build_one_package() {
     ;;
 
     simpcomp*)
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to GAP 4.9 via 28592e83947c7aae7fa40b68f13e1fbdecb46f6a

PR #2477, opened 2018-05-17
Handle lists being modified by use of ~

ChrisJefferson:
Adjust how lists are built when ~ is used, so we do not crash
even if the list is modified as it is being constructed.

This also makes the behaviour of GAP consistent between bytecode
and immediate code.

### DIFF
--- a/src/exprs.c
+++ b/src/exprs.c
@@ -920,8 +920,8 @@ Obj             EvalPermExpr (
 **  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
 **  list expression.
 */
-Obj             ListExpr1 ( Expr expr );
-void            ListExpr2 ( Obj list, Expr expr );
+Obj             ListExpr1(Expr expr, Int tildeInUse);
+void            ListExpr2(Obj list, Expr expr, Int tildeInUse);
 Obj             RecExpr1 ( Expr expr );
 void            RecExpr2 ( Obj rec, Expr expr );
 
@@ -931,8 +931,8 @@ Obj             EvalListExpr (
     Obj                 list;         /* list value, result                */
 
     /* evaluate the list expression                                         */
-    list = ListExpr1( expr );
-    ListExpr2( list, expr );
+    list = ListExpr1(expr, 0);
+    ListExpr2(list, expr, 0);
 
     /* return the result                                                   */
     return list;
@@ -964,13 +964,13 @@ Obj             EvalListTildeExpr (
     tilde = STATE( Tilde );
 
     /* create the list value                                               */
-    list = ListExpr1( expr );
+    list = ListExpr1(expr, 1);
 
     /* assign the list to '~'                                              */
     STATE(Tilde) = list;
 
     /* evaluate the subexpressions into the list value                     */
-    ListExpr2( list, expr );
+    ListExpr2(list, expr, 1);
 
     /* restore old value of '~'                                            */
     STATE(Tilde) = tilde;
@@ -982,8 +982,8 @@ Obj             EvalListTildeExpr (
 
 /****************************************************************************
 **
-*F  ListExpr1(<expr>) . . . . . . . . . . . make a list for a list expression
-*F  ListExpr2(<list>,<expr>)  . . .  enter the sublists for a list expression
+*F  ListExpr1(<expr>,<tildeInUse>) . . . .  make a list for a list expression
+*F  ListExpr2(<list>,<expr>,<tildeInUse>)  enter the sublists for a list expr
 **
 **  'ListExpr1' and 'ListExpr2'  together evaluate the list expression <expr>
 **  into the list <list>.
@@ -998,9 +998,15 @@ Obj             EvalListTildeExpr (
 **  This two step allocation  is necessary, because  list expressions such as
 **  '[ [1], ~[1] ]'  requires that the value of  one subexpression is entered
 **  into the list value before the next subexpression is evaluated.
+**
+**  'tildeInUse' is 1 when this list is part of the value of ~.
+**
+**  When 'tildeInUse' is 1, use a slower code path which ensures the list is
+**  in a valid state after each element is added and handles the elements
+**  and TNUM of the list changing whenever a new element is constructed.
+**
 */
-Obj ListExpr1 (
-    Expr                expr )
+ALWAYS_INLINE Obj ListExpr1(Expr expr, Int tildeInUse)
 {
     Obj                 list;           /* list value, result              */
     Int                 len;            /* logical length of the list      */
@@ -1015,15 +1021,14 @@ Obj ListExpr1 (
     else {
         list = NEW_PLIST( T_PLIST, len );
     }
-    SET_LEN_PLIST( list, len );
+
+    SET_LEN_PLIST(list, tildeInUse ? 0 : len);
 
     /* return the list                                                     */
     return list;
 }
 
-void ListExpr2 (
-    Obj                 list,
-    Expr                expr )
+ALWAYS_INLINE void ListExpr2(Obj list, Expr expr, Int tildeInUse)
 {
     Obj                 sub;            /* value of a subexpression        */
     Int                 len;            /* logical length of the list      */
@@ -1053,18 +1058,27 @@ void ListExpr2 (
           {
             if (posshole == 1)
               {
-                SET_FILT_LIST(list, FN_IS_NDENSE);
+                if (!tildeInUse) {
+                    SET_FILT_LIST(list, FN_IS_NDENSE);
+                }
                 posshole = 2;
               }
             sub = EVAL_EXPR( ADDR_EXPR(expr)[i-1] );
-            SET_ELM_PLIST( list, i, sub );
+            if (tildeInUse) {
+                ASS_LIST(list, i, sub);
+            }
+            else {
+                SET_ELM_PLIST(list, i, sub);
+            }
             CHANGED_BAG( list );
           }
 
     }
-    if (!posshole)
-      SET_FILT_LIST(list, FN_IS_DENSE);
 
+    // If tildeInUse = 1, elements of 'list' may have been removed
+    if (!posshole && !tildeInUse) {
+        SET_FILT_LIST(list, FN_IS_DENSE);
+    }
 }
 
 
--- a/tst/testinstall/tilde.tst
+++ b/tst/testinstall/tilde.tst
@@ -85,4 +85,15 @@ true
 gap> i := InputTextString( "local a; a := 10; return rec(a := a*10, b := ~);" );;
 gap> r := rec(a := ~, b := ReadAsFunction(i)(), c := ~.b.a);
 rec( a := ~, b := rec( a := 100, b := ~.b ), c := 100 )
+gap> [Length(~),Length(~),Length(~)];
+[ 0, 1, 2 ]
+gap> (function() return [Length(~),Length(~),Length(~)]; end)();
+[ 0, 1, 2 ]
+gap> rem := function(l) Remove(l); return 1; end;;
+gap> [2,rem(~),rem(~),rem(~)];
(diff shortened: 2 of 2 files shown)


src/exprs.c
```diff
@@ -919,9 +919,18 @@ Obj             EvalPermExpr (
 **
 **  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
 **  list expression.
+**
+**  The 'safe' argument to ListExpr1 and ListExpr2 handle code which abuses ~
+**  access and change the partially built list.
```
fingolfin: handle -> handles ? though I don't see how an argument can "handle" anything by itself?

src/exprs.c
```diff
@@ -919,9 +919,18 @@ Obj             EvalPermExpr (
...
+**
+**  When 'safe' is 1, we use a slower code path which ensures the list is
+**  in a valid state after each element is added, and not crash if the
```
fingolfin: missing words... perhaps "... is added, and that we don't crash ..."

src/exprs.c
```diff
@@ -999,8 +1008,9 @@ Obj             EvalListTildeExpr (
 **  '[ [1], ~[1] ]'  requires that the value of  one subexpression is entered
 **  into the list value before the next subexpression is evaluated.
 */
-Obj ListExpr1 (
-    Expr                expr )
+ALWAYS_INLINE Obj ListExpr1 (
```
fingolfin: Why `ALWAYS_INLINE`?

src/exprs.c
```diff
@@ -1015,15 +1025,21 @@ Obj ListExpr1 (
     else {
         list = NEW_PLIST( T_PLIST, len );
     }
-    SET_LEN_PLIST( list, len );
+    if(safe) {
```
fingolfin: add space after `if`

src/exprs.c
```diff
@@ -1053,17 +1069,25 @@ void ListExpr2 (
           {
             if (posshole == 1)
               {
-                SET_FILT_LIST(list, FN_IS_NDENSE);
+                if(!safe || TNUM_OBJ(list) == T_PLIST) {
```
fingolfin: add space after `if` (or just `git clang-format`)

tst/testinstall/tilde.tst
```diff
@@ -85,4 +85,15 @@ true
 gap> i := InputTextString( "local a; a := 10; return rec(a := a*10, b := ~);" );;
 gap> r := rec(a := ~, b := ReadAsFunction(i)(), c := ~.b.a);
 rec( a := ~, b := rec( a := 100, b := ~.b ), c := 100 )
+gap> [Length(~),Length(~),Length(~)];
+[ 0, 1, 2 ]
+gap> (function() return [Length(~),Length(~),Length(~)]; end)();
+[ 0, 1, 2 ]
+gap> rem := function(l) Remove(l); return 1; end;;
+gap> [2,rem(~),rem(~),rem(~)];
```
fingolfin: Wow, that's pretty evil :-)

ssiccha: I agree

src/exprs.c
ChrisJefferson: Will fix all other comments.

With this one, ALWAYS_INLINE because I want the 'safe' compiled out when safe=0, so the fast route (without tilde) is as fast as it used to be.

tst/testinstall/tilde.tst
ChrisJefferson: I don't expect anyone to ever write this, but I feel there shouldn't be ways of making the interpreter crash (at least, without calling ALL_CAPS functions).

fingolfin: Yeah, sure, it's good to add that test. People do evil things, we shouldn't reward them for it with a crash ;-)

src/exprs.c
fingolfin: ok

ChrisJefferson:
Hopefully changes made, added to release notes.

src/exprs.c
```diff
@@ -919,9 +919,19 @@ Obj             EvalPermExpr (
 **
 **  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
 **  list expression.
+**
+**  The 'safe' argument to ListExpr1 and ListExpr2 should be set to 1 when
+**  the list expression may be modified when its elements are being
```
fingolfin: "when its elements are being constructed" -> "while its ..." ? (It took me a moment to parse the current phrasing, but then I did understand it, so it's probably fine as it is -- I just try to be super sensitive and try to simulate how I might read a paragraph if I knew less about the kernel; but of course that's imperfect. Perhaps e.g. @ssiccha would be willing to chime and say whether the above text seems clear to him -- hopefull it is, but  if not, it'd be useful to know which bits are not.

src/exprs.c
```diff
@@ -919,9 +919,19 @@ Obj             EvalPermExpr (
 **
 **  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
 **  list expression.
+**
+**  The 'safe' argument to ListExpr1 and ListExpr2 should be set to 1 when
```
fingolfin: Shouldn't this text be part of the documentation comment of `ListExpr1` and `ListExpr2` instead?

I understand that there are prototypes for those two functions below, but considering the rest of their documentation is elsewhere... Perhaps we could move them up, before this function, to avoid using prototypes? Then they could also be made static.

src/exprs.c
```diff
@@ -1053,18 +1065,25 @@ void ListExpr2 (
           {
             if (posshole == 1)
               {
-                SET_FILT_LIST(list, FN_IS_NDENSE);
+                if (!safe || TNUM_OBJ(list) == T_PLIST) {
```
fingolfin: Hmm, is that safe? Couldn't the tnum be, say `T_PLIST_DENSE` at this point? That would suggest using `if (!safe || IS_PLIST(list))` - or perhaps even `IS_LIST`?

But why is this necessary at all? `SET_FILT_LIST` uses a tnum dispatch table -- if the user did something really bad here, wouldn't that just trigger an error via that dispatch table?

src/exprs.c
```diff
@@ -919,9 +919,19 @@ Obj             EvalPermExpr (
...
 **  list expression.
+**
+**  The 'safe' argument to ListExpr1 and ListExpr2 should be set to 1 when
+**  the list expression may be modified when its elements are being
+**  constructed. This can occur when an element of the list refers to ~.
+**
+**  When 'safe' is 1, we use a slower code path which ensures the list is
+**  in a valid state after each element is added, and we don't crash if the
+**  size or TNUM of the list is changed during the construction of elements
+**  in the list.
```
fingolfin: Perhaps that should say "we don't crash if the size or type of the list changed, or even if gets swapped for a non-list object" ? After all, the user could use SWAP_OBJ to replace it by, say, a bigint.

OTOH, perhaps I am just overthinking that; and people using SWAP_OBJ deserve what they get ;-).

src/exprs.c
```diff
@@ -919,9 +919,19 @@ Obj             EvalPermExpr (
 **
 **  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
 **  list expression.
+**
+**  The 'safe' argument to ListExpr1 and ListExpr2 should be set to 1 when
+**  the list expression may be modified when its elements are being
+**  constructed. This can occur when an element of the list refers to ~.
+**
+**  When 'safe' is 1, we use a slower code path which ensures the list is
```
fingolfin: Hmm, when I was reading the code below, I constantly had to refer back to this comment, because I could not remember whether `safe` meant: either "we are safe; it is safe to assume that the list won't be messed with", or perhaps "play it safe; use safe code which works even if the list gets messed with".

Since these are opposite, that's a bit of a problem, IMHO. Is it just me? Perhaps we can find a better name. Say `protectAgainstListModification` or `safelyGrowList`, `safelyExtendList`; or perhaps reverse the value, and say `assumeListIsPlist` or ...? Sorry, no great idea.

src/exprs.c
ChrisJefferson: You are right, let me rethink this.

src/exprs.c
ChrisJefferson: I could just call it `tildeInUse`, seeing as that's all we use it for :)

fingolfin: Yes, that actually sounds very good to me!

src/exprs.c
ssiccha: @fingolfin @ChrisJefferson Apart from knowing what exactly an expression is (`system.h` says its a `UInt`, but that's all I found out quickly), the formulation is clear to me.

ssiccha: > "when its elements are being constructed" -> "while its ..." 
"while its" also sounds better to me

ChrisJefferson:
Now changed/improved

src/exprs.c
```diff
@@ -998,9 +998,15 @@ Obj             EvalListTildeExpr (
...
+**  'tildeInUse' is 1 when this list is part of the value of ~.
+**
+**  When 'tildeInUse' is 1, use a slower code path which ensures the list is
+**  in a valid state after each element is added and handles the elements
+**  and TNUM of the list changing whenever a new element is constructed.
+**
 */
-Obj ListExpr1 (
-    Expr                expr )
+ALWAYS_INLINE Obj ListExpr1(Expr expr, Int safe)
```
fingolfin: The documentation now says `tildeInUse`, great -- but the code still uses `safe` :-)

src/exprs.c
```diff
@@ -998,9 +998,15 @@ Obj             EvalListTildeExpr (
 **  This two step allocation  is necessary, because  list expressions such as
 **  '[ [1], ~[1] ]'  requires that the value of  one subexpression is entered
 **  into the list value before the next subexpression is evaluated.
+**
+**  'tildeInUse' is 1 when this list is part of the value of ~.
```
fingolfin: OK. Could you please adjust lines 984/985? You could copy&past this over them
```C
*F  ListExpr1(<expr>,<tildeInUse>) . . . .  make a list for a list expression
*F  ListExpr2(<list>,<expr>,<tildeInUse>) . . enter sublists for a list expr.
```

src/exprs.c
```diff
@@ -1015,15 +1021,18 @@ Obj ListExpr1 (
     else {
         list = NEW_PLIST( T_PLIST, len );
     }
-    SET_LEN_PLIST( list, len );
+    if (safe) {
+        SET_LEN_PLIST(list, 0);
+    }
+    else {
+        SET_LEN_PLIST(list, len);
+    }
```
fingolfin: Could also use `SET_LEN_PLIST(list, tildeInUse ? 0 : len);` for compactness -- but of course it's also perfectly fine as it is :-).

src/exprs.c
```diff
@@ -1053,18 +1062,25 @@ void ListExpr2 (
...
+            }
             CHANGED_BAG( list );
           }
 
     }
-    if (!posshole)
-      SET_FILT_LIST(list, FN_IS_DENSE);
-
+    if (!posshole && !safe) {
+        SET_FILT_LIST(list, FN_IS_DENSE);
```
fingolfin: Why do we need the `&& !safe` (resp. `&& !tildeInUse`) here? Isn't the list fully constructed at this point?

Ahh, because we might have used unbind while constructing it, through the use of `~`? Is that the reason?
Perhaps a small comment could be added to point that out (or, if I am wrong, then point out the actual reason)?

src/exprs.c
ChrisJefferson: bla!

src/exprs.c
ChrisJefferson: I'm just assuming that SET_FILT_LIST is always an optimisation, and as you say, who knows what state the list is in, or if it's even a list.

PR #2478, opened 2018-05-18
Some more tests for stream objects

fingolfin:


fingolfin:
@ChrisJefferson so in this PR, I moved some stream tests from teststandard to testinstall -- main motivation is so that they also get run in HPC-GAP. However, you said those tests were slow on Windows, hence why you did not put this into testinstall before. But at least AppVeyor's runtime does not seem to be affected negatively...

fingolfin:
Alternatively to moving those tests, we could also change the single HPC-GAP build on Travis to use `testtravis` instead of `testinstall`.

PR #2484, opened 2018-05-22
Fix some HPC-GAP bugs and switch HPC-GAP build to use testtravis ...

fingolfin:
... instead of testinstall. While this makes it slower, this also means we get slightly higher code coverage on the HPC-GAP code.

Also, it uncovered several bugs in HPC-GAP. Two of them are fixed in here, others were fixed in other PRs.


fingolfin:
There are some test failing here, which I think are worth investigating.

fingolfin:
Nice, this adds 0.62% coverage and we go over 75% total coverage :-).

PR #2488, opened 2018-05-24
FIX: (Maximal) subgroups computation.

hulpke:
The routine for `MaximalSubgroupClassReps allows options to calculate
maximal subgroups only up to specified limits, or if it is not too
difficult. This causes problems is such a partial list is stored as
attribute. Thus separate into two attributes: One to calculate the
guaranteed full list, one to calculate a potentially partial list subject to
restrictions. Also make sure that limiting options do not get accidentally
inherited.
Finally allow a soft fallback to old intermediate subgroup routines, if
maximal subgroups are not available. (This can go away once proper maximal
subgroups code is available.)

This fixes a bug reported (Subject:`Intermediate subgroup-Gap 4.9.1`) on the support email list.

Included is also a fix for the main bug in #2586

Also added test files.

In the best of all worlds this should be backported to 4.9

### DIFF
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -209,7 +209,7 @@ end);
 ##  the operation of G on the Right Cosets of U.
 ##
 InstallGlobalFunction( IntermediateGroup, function(G,U)
-local o,b,img,G1,c,m,hardlimit,gens,t,k,intersize;
+local o,b,img,G1,c,m,mt,hardlimit,gens,t,k,intersize;
 
   if U=G then
     return fail;
@@ -221,34 +221,38 @@ local o,b,img,G1,c,m,hardlimit,gens,t,k,intersize;
     return fail; # avoid infinite recursion
   fi;
 
-  # use maximals
-  m:=MaximalSubgroupClassReps(G:cheap,intersize:=intersize);
+  # use maximals, use `Try` as we call with limiting options
+  IsNaturalAlternatingGroup(G);
+  IsNaturalSymmetricGroup(G);
+  m:=TryMaximalSubgroupClassReps(G:cheap,intersize:=intersize,nolattice);
+  if m<>fail and Length(m)>0 then
 
-  m:=Filtered(m,x->Size(x) mod Size(U)=0 and Size(x)>Size(U));
-  SortBy(m,x->Size(G)/Size(x));
-  
-  gens:=SmallGeneratingSet(U);
-  for c in m do
-    if Index(G,c)<50000 then
-      t:=RightTransversal(G,c:noascendingchain); # conjugates
-      for k in t do
-        if ForAll(gens,x->k*x/k in c) then
-	  Info(InfoCoset,2,"Found Size ",Size(c),"\n");
-          # U is contained in c^k
-          return c^k;
+    m:=Filtered(m,x->Size(x) mod Size(U)=0 and Size(x)>Size(U));
+    SortBy(m,x->Size(G)/Size(x));
+    
+    gens:=SmallGeneratingSet(U);
+    for c in m do
+      if Index(G,c)<50000 then
+        t:=RightTransversal(G,c:noascendingchain); # conjugates
+        for k in t do
+          if ForAll(gens,x->k*x/k in c) then
+            Info(InfoCoset,2,"Found Size ",Size(c),"\n");
+            # U is contained in c^k
+            return c^k;
+          fi;
+        od;
+      else
+        t:=DoConjugateInto(G,c,U,true:intersize:=intersize,onlyone:=true);
+        if t<>fail and t<>[] then 
+          Info(InfoCoset,2,"Found Size ",Size(c),"\n");
+          return c^(Inverse(t));
         fi;
-      od;
-    else
-      t:=DoConjugateInto(G,c,U,true:intersize:=intersize,onlyone:=true);
-      if t<>fail and t<>[] then 
-	Info(InfoCoset,2,"Found Size ",Size(c),"\n");
-        return c^(Inverse(t));
       fi;
-    fi;
-  od;
+    od;
 
-  Info(InfoCoset,2,"Found no intermediate subgroup ",Size(G)," ",Size(U));
-  return fail;
+    Info(InfoCoset,2,"Found no intermediate subgroup ",Size(G)," ",Size(U));
+    return fail;
+  fi;
 
   # old code -- obsolete
 
@@ -824,7 +828,7 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
 	fi;
       end;
 
-      for i in MaximalSubgroupClassReps(G:cheap) do
+      for i in TryMaximalSubgroupClassReps(G:cheap) do
 	if Index(G,i)<maxidx(c) and Index(G,i)<badlimit then
 	  p:=Intersection(a,i);
 	  if Index(a,p)<uplimit then
@@ -846,7 +850,7 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
     
     if maxidx(c)>10*actlimit then
 
-      r:=ShallowCopy(MaximalSubgroupClassReps(a:cheap));
+      r:=ShallowCopy(TryMaximalSubgroupClassReps(a:cheap));
       r:=Filtered(r,x->Index(a,x)<uplimit);
 
       Sort(r,function(a,b) return Size(a)<Size(b);end);
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -1158,8 +1158,9 @@ DeclareAttribute( "ConjugacyClasses", IsGroup );
 ##  <Ref Func="MaximalSubgroupClassReps"/>.
 ##  <Example><![CDATA[
 ##  gap> ConjugacyClassesMaximalSubgroups(g);
-##  [ AlternatingGroup( [ 1 .. 4 ] )^G, Group( [ (1,2,3), (1,2) ] )^G, 
-##    Group( [ (1,2), (3,4), (1,3)(2,4) ] )^G ]
+##  [ Group( [ (2,4,3), (1,4)(2,3), (1,3)(2,4) ] )^G,
+##    Group( [ (3,4), (1,4)(2,3), (1,3)(2,4) ] )^G,
+##    Group( [ (3,4), (2,4,3) ] )^G ]
 ##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
@@ -1205,8 +1206,8 @@ DeclareAttribute( "MaximalSubgroups", IsGroup );
 ##  of <A>G</A>.
 ##  <Example><![CDATA[
 ##  gap> MaximalSubgroupClassReps(g);
-##  [ Alt( [ 1 .. 4 ] ), Group([ (1,2,3), (1,2) ]), Group([ (1,2), (3,4),
-##      (1,3)(2,4) ]) ]
+##  [ Group([ (2,4,3), (1,4)(2,3), (1,3)(2,4) ]), Group([ (3,4), (1,4)
+##    (2,3), (1,3)(2,4) ]), Group([ (3,4), (2,4,3) ]) ]
 ##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
@@ -1214,7 +1215,11 @@ DeclareAttribute( "MaximalSubgroups", IsGroup );
 ##
 DeclareAttribute("MaximalSubgroupClassReps",IsGroup);
 
+# utility attribute: Allow use with limiting options, so could hold `fail'.
+DeclareAttribute("TryMaximalSubgroupClassReps",IsGroup,"mutable");
+
 # utility function in maximal subgroups code
+DeclareGlobalFunction("TryMaxSubgroupTainter");
 DeclareGlobalFunction("DoMaxesTF");
 DeclareGlobalFunction("MaxesAlmostSimple");
 
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -1357,6 +1357,53 @@ end);
 ##
 #M  MaximalSubgroups( <G> )
 ##
+InstallMethod(MaximalSubgroupClassReps,"default, catch dangerous options",
+  true,[IsGroup],0,
+function(G)
+local H,a,m,i,l;
+  # easy case, go without options
+  if not HasTryMaximalSubgroupClassReps(G) then
+    return TryMaximalSubgroupClassReps(G:
+           # as if options were unset
+           cheap:=fail,intersize:=fail,inmax:=fail,nolattice:=fail);
+  fi;
+
+  # hard case -- `Try` is stored
+  if not IsBound(G!.maxsubtrytaint) or G!.maxsubtrytaint=false then
+    # stored and untainted, just go on
+    return TryMaximalSubgroupClassReps(G); 
+  fi;
+
+  # compute anew for new group to avoid taint
+  H:=Group(GeneratorsOfGroup(G));
+  for i in [Size,IsNaturalAlternatingGroup,IsNaturalSymmetricGroup] do
+    if Tester(i)(G) then Setter(i)(H,i(G));fi;
+  od;
+  m:=TryMaximalSubgroupClassReps(H:
+          cheap:=false,intersize:=false,inmax:=false,nolattice:=false);
+  l:=[];
+  for i in m do
+    a:=SubgroupNC(G,GeneratorsOfGroup(i));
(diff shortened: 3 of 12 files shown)


lib/grp.gd
```diff
@@ -1202,7 +1202,11 @@ DeclareAttribute( "MaximalSubgroups", IsGroup );
 ##
 DeclareAttribute("MaximalSubgroupClassReps",IsGroup);
 
+# utility attribute: Allow use with limiting options, so could hold `fail'.
+DeclareAttribute("TryMaximalSubgroupClassReps",IsGroup);
```
fingolfin: What I don't understand about this is why `TryMaximalSubgroupClassReps` even is an attribute, and not just an operation, as its value may depend on options, and there is no way for which options the set attribute value was computed, is there?

(Perhaps also a mutable attribute would be a choice, so I don't quite see why/how)?

Mind you, I am not saying this is wrong, just that I don't understand the rationale, and would appreciate some illuminating words. Thanks!

hulpke: @fingolfin 
I'm still trying to get the tests run through, so it might be worth to wait with reviewing until the checks show green.

The reason for making Try... an attribute is that a calculation might (in fact that happens when calculating double cosets) several times ask for cheap maximal subgroups of the same group.

hulpke: @fingolfin OK, it now has run through and I have rebased away patches. It affects a comparatively large number of files, but basically because the use of options in the interface was badly designed. As far as the functionality is affected it is actually a quite local patch at the interface of Maximal subgroups and intermediate subgroups.

hulpke:
@fingolfin 
OK, it now has run through and I have rebased away patches. It affects a comparatively large number of files, but basically because the use of options in the interface was badly designed. As far as the functionality is affected it is actually a quite local patch at the interface of Maximal subgroups and intermediate subgroups.

[Moved comment out of `outdated' bit.

tst/testbugfix/2018-05-24-IntermediateSubgroups.tst
```diff
@@ -0,0 +1,16 @@
+# test for MaximalSubgroupClassReps with options (reported through
+# observation by S.Alavi with IntermediateGroup
+
+gap> g:= AtlasGroup( "U4(4)" );;  Size( g );
+1018368000
+gap> h:= Image( IsomorphismPermGroup( PSL(2,16) ) );;
+gap> l:= IsomorphicSubgroups( g, h );;
```
fingolfin: This computation can take minutes when I try it here. Since only one of those groups is needed, can't we just explicitly describe it? Or, just switch to a smaller example, like this (I verified that it fails in `master`, too):
```
gap> g:=SU(IsPermGroup,3,4);
Perm_SU(3,4)
gap> h:=PSL(IsPermGroup, 2, 4);;
gap> l:= IsomorphicSubgroups( g, h );; Length(l);
1
gap> s1:= Image( l[1] );;  Size( s1 );
60
gap> n1:= Normalizer( g, s1 );;  Size( n1 );
300
gap> int:=IntermediateGroup(g,s1);;
gap> IsGroup(int);
true
```

That way, the bugfix test suite does not get slowed down, nor does it require the atlasrep package.

lib/grp.gi
```diff
@@ -1342,6 +1342,50 @@ end);
 ##
 #M  MaximalSubgroups( <G> )
 ##
+InstallMethod(MaximalSubgroupClassReps,"default, catch dangerous options",
```
fingolfin: Should there perhaps also be a `TryMaximalSubgroupClassReps` method in the filter `HasMaximalSubgroupClassReps`, like this:
```gap
InstallMethod(TryMaximalSubgroupClassReps,
    [IsGroup and HasMaximalSubgroupClassReps],
    MaximalSubgroupClassReps);
```
Possibly with a high rank adjustment.

lib/grp.gi
```diff
@@ -1342,6 +1342,50 @@ end);
...
+  l:=[];
+  for i in m do
+    a:=SubgroupNC(G,GeneratorsOfGroup(i));
+    if HasSize(i) then SetSize(a,Size(i));fi;
+    Add(l,a);
+  od;
+
+  # now we know list is untained, store 
+  Unbind(G!.maxsubtrytaint);
+  G!.TryMaximalSubgroupClassReps:=l;
```
fingolfin: So you bypass the setter function. Any particular reason for that? It also means that `Has TryMaximalSubgroupClassReps` won't be set.

lib/grp.gi
```diff
@@ -1342,6 +1342,50 @@ end);
...
+  return l;
+
+end);
+
+InstallGlobalFunction(TryMaxSubgroupTainter,function(G)
```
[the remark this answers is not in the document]
fingolfin: Things get even simpler if you also add an operation `TryMaximalSubgroupClassRepsOp`, and move most current `MaximalSubgroupClassReps` resp. `TryMaximalSubgroupClassReps` methods to that operation. Then, both `MaximalSubgroupClassReps` resp. `TryMaximalSubgroupClassReps` can dispatch to that operation. This would also remove the need to create a temporary "copy" of the group, and later transfer subgroups back to the original group.

And the main (only?) `TryMaximalSubgroupClassReps` method can look roughly like this:
```
function(G)
  local maxs;
  if HasMaximalSubgroupClassReps(G) then return MaximalSubgroupClassReps(G); fi;
  maxs := TryMaximalSubgroupClassRepsOp(G);
  if NO-BAD-OPTIONS-SET then
    SetMaximalSubgroupClassReps(G, maxs);
  fi;
  return maxs`
end;
```
That's still slightly brittle, because some `TryMaximalSubgroupClassRepsOp` might accept more options in the future, and if we forget to update the above method, we have a problem... But at least we won't be off worse than we are now.

(BTW, I'd actually turn those options into explicit arguments to `TryMaximalSubgroupClassRepsOp`, but that's probably a matter of taste)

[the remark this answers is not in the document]
fingolfin: PS: of course one could also try to come up with a better name... perhaps the second attribute should be `SomeMaximalSubgroupClassReps`, and the operation then could be `TryMaximalSubgroupClassReps`, or `ComputeSomeMaximalSubgroupClassReps`, or ... but that's of course in the end irrelevant, the functionality is what counts.

lib/csetgrp.gi
```diff
@@ -221,34 +221,38 @@ local o,b,img,G1,c,m,hardlimit,gens,t,k,intersize;
...
-    if Index(G,c)<50000 then
-      t:=RightTransversal(G,c:noascendingchain); # conjugates
-      for k in t do
-        if ForAll(gens,x->k*x/k in c) then
-	  Info(InfoCoset,2,"Found Size ",Size(c),"\n");
-          # U is contained in c^k
-          return c^k;
+  # use maximals, use `Try` as we call with limiting options
+  IsNaturalAlternatingGroup(G);
+  IsNaturalSymmetricGroup(G);
```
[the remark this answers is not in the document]
hulpke: Yes. But this optimization is checked for in manual and test examples.

lib/grp.gi
hulpke: In this situation the attribute is already set with a list that is not guaranteed to be complete. We also have just computed a complete list and we store it in place. Can I use a setter to change the value?

fingolfin: For mutable attributes: yes!
```
gap> G:=SymmetricGroup(5);
Sym( [ 1 .. 5 ] )
gap> SetIsFoo(G, "foo");
gap> IsFoo(G);
"foo"
gap> SetIsFoo(G, "nar");
gap> IsFoo(G);
"nar"
```

lib/grp.gi
```diff
@@ -1342,6 +1342,50 @@ end);
...
+  # hard case -- `Try` is stored
+  if not IsBound(G!.maxsubtrytaint) or G!.maxsubtrytaint=false then
+    # stored and untainted, just go on
+    return TryMaximalSubgroupClassReps(G); 
+  fi;
+
+  # compute anew for new group to avoid taint
+  H:=Group(GeneratorsOfGroup(G));
+  for i in [Size,IsNaturalAlternatingGroup,IsNaturalSymmetricGroup] do
+    if Tester(i)(G) then Setter(i)(H,i(G));fi;
```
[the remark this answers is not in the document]
hulpke: They are currently the only two for which special methods for maximal subgroups are installed.

tst/testbugfix/2018-05-24-IntermediateSubgroups.tst
hulpke: Yes, the `IsomorphicSubgroups` can go and has done so. The issue of the bug was that it was a (comparatively) small index subgroup in a large group, so large indeed that it is unlikely to be added to the table of marks library anytime soon.

My personal inclination is at this time to have the original bug, but if someone wants to change it afterwards for speed reasons I have no objection.

lib/grp.gi
hulpke: Will certainly not harm. Added.

lib/csetgrp.gi
```diff
@@ -221,34 +221,38 @@ local o,b,img,G1,c,m,hardlimit,gens,t,k,intersize;
...
+  m:=TryMaximalSubgroupClassReps(G:cheap,intersize:=intersize,nolattice);
+  if m<>fail and Length(m)>0 then
```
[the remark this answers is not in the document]
hulpke: Yes.

hulpke:
As far as I'm concerned this fix is complete. If someone wants to make additions to it after merging (or propose a different fix) feel free to do so. I am happy to help with integrating it into 4.9, but not so much in discussing possible other code structures or function names.

I have included the suggestions that to me made sense as part of a fix. I have not done code restructuring or manual additions. Let me explain why:

This is a fix that should be back ported to 4.9. For this it is unfortunate that it requires a new attribute, but still, the changes ought to be as small as possible. This is the reason for not doing structural changes.

As for undocumented options, these serve two purposes:
- Stop an infinite recursion (because in some cases it is helpful to have maximal subgroups of smaller groups)
- Work around the real issue, namely GAP's lack of maximal subgroup information for classical (or even Lie-Type) groups. This means that a maximal subgroup calculation might need to fall back on expensive subgroup lattice computations, this needs to be stopped if maximal subgroups are only used to get some proper subgroups cheap and quickly.

In the end, I hope we will have (as a port of the Magma functionality that I am working on) this lacking functionality and at that point both the `Try...` attribute (which is there to store down incomplete information) as well as most of the options can go.
If we document either, they might get used by other code, and then are hard to get rid off again.

I am also happy to look, for future releases, at changes for the attribute to clean up the ``quick'' versus ``complete'' situation without abusing attributes.
This however will likely introduce some syntax incompatibilities and thus is not appropriate for a bugfix.

tst/testbugfix/2018-05-24-IntermediateSubgroups.tst
```diff
@@ -0,0 +1,24 @@
+# test for MaximalSubgroupClassReps with options (reported through
+# observation by S.Alavi with IntermediateGroup
+# More complicated to construct w/o AtlasSubgroup to ensure the 325 points action, as it is too
+# slow otherwise. Also construct the smaller subgroup s1 directly.
+# Finally do not slow down with assertions that don't need testing here
+
+gap> SetAssertionLevel(0);;
```
fingolfin: If you turn off assertions here, this will also turn them off in subsequent tests -- until some test uses `START_TEST/STOP_TEST`, which not all .tst files do right now.
So please either manually save/restore the assertion level, or else (probably simpler) add `START_TEST` and `STOP_TEST` calls.

Alternatively, somebody else can add this after this PR gets merged.

fingolfin (approved):
While I still am not really happy with `TryMaxSubgroupTainter` (because I think it's very easy to produce incorrect code with this approach; while the alternative I suggested should not require *more* changes, in fact, I think it needs fewer), I also appreciate that reworking this requires time, effort and nerves, and this fix is definitely *much* better than no fix! We can indeed (if we want to) improve on it later.

I am slightly unsure about backporting this to GAP 4.9, given the magnitude of the change. But if others are fine with it... @alex-konovalov could this still make GAP 4.9.2 (I am not quite sure about the release schedule for that).

BTW, note that our plan is that GAP 4.10 will branch around September 1st, and 4.10.0 be released around October 1st (whether this will be already public, or again "just" a beta as before, with the first public release 4.10.1. on 1st of November, still needs to be discussed.

For record, the `cryst` package also installs two methods for `MaximalSubgroupClassReps` which are checking various options (though I have not looked further into it, i.e. I don't know if these options affect the result). So I guess once this PR is merged, we should contact @gaehler to discuss with him whether cryst needs an update, too?

tst/testbugfix/2018-06-28-grplatt.tst
```diff
@@ -0,0 +1,7 @@
+# #2586
```
fingolfin: If we ever switch away from GitHub, this reference will be completely inscrutable. Could you perhaps change it to e.g. this?
```
# Verify that ConjugacyClassesSubgroups returns classes with representatives that actually are
# contained in the group (bug #2586 on GitHub)
```
Or if you want to stay brief (as the test actually is quite self-explanatory), perhaps at least
```
# verify fix for bug #2586 on GitHub
```

lib/grplatt.gi
```diff
@@ -2819,11 +2819,10 @@ InstallGlobalFunction("SubgroupsTrivialFitting",function(G)
 	ValueOption(NO_PRECOMPUTED_DATA_OPTION)<>true then
       Info(InfoPerformance,2,"Using Table of Marks Library");
       go:=ImagesSource(tom[1]);
-      tom:=tom[2];
       Info(InfoLattice,1, "Fetching subgroups of simple ",
-	  Identifier(tom)," from table of marks");
-      len:=LengthsTom(tom);
-      sub:=List([1..Length(len)],x->RepresentativeTom(tom,x));
+	  Identifier(tom[2])," from table of marks");
+      len:=LengthsTom(tom[2]);
```
fingolfin: This looks plausible to me, thanks.

Quick remark on the commit message: It contains the string "partially fixes #2586" -- GitHub sees that, but only the "fixes #2586" bit, it doesn't care about the "partially". As a result, if this commit gets merged, GitHub will close issue #2586. 

One workaround is to rephrase this to e.g. `fixes issue #2586` (I think). Well, or we wait for you to also fix the second bug, then closing the issue is fine, after all ;-)

tst/testbugfix/2018-06-28-grplatt.tst
```diff
@@ -0,0 +1,7 @@
...
+
+gap> g:=TransitiveGroup(6,14);;
+gap> cl:=ConjugacyClassesSubgroups(g);;
+gap> ForAll(cl,x->IsSubset(g,Representative(x)));
+true
+
```
fingolfin: This test file actually fails the test due to this extra new line at the end. Removing it should be enough to fix the test.

markuspf (approved):
Subject to the tests passing this looks good to me.

hulpke:
The two fixes now are in two separate, non-interleaved commits.

markuspf:
Is the intention for this to be merged once the tests pass, then?

hulpke:
@alex-konovalov 
As for back porting, this is of course your decision.

Just in case this is relevant information, this first fix (for MaximalSubgroups) resolves a bug reported by S. Alavi from Iran on 5/24 (subject Intermediate subgroup-Gap 4.9.1).
That report BTW never got an official reply.

fingolfin:
Re: the bug report by S. Alavi not having been replied to: that's of course unfortunate. But it's the usual problem when people start replying to the report, but not to the reporter, and then nobody remembers to reply with at least an "we are looking into it".
So, to make up for that, once this is merged, "somebody" (who?) should reply to S. Alavi after all, and confirm the issue, and indicate the upcoming bug fix.

fingolfin:
As to backporting: I think only the first commit can and should be backported -- the second fixes a regression that does not affect stable-4.9, after all (and modifies code that's not even on that branch, so it wouldn't even apply).

fingolfin:
I'll reply to S. Alavi

hulpke:
@fingolfin 
Thank you very much for replying to Alavi.

gaehler:
I still do not understand what the impact of this change on the cryst package is. According to the manual, attribute values computed from a call with more than one argument are never stored. Is this still true? If so, no changes to cryst are necessary. Otherwise, I do have a problem...

For affine crystallographic groups, MaximalSubgroupClassReps must always be called with two arguments, because the total number of classes is generally not finite. This is checked, and an error is returned otherwise. I hope this will just continue to work, without using TryMaximalSubgroupClassReps.

gaehler:
I have to correct myself. A method for one argument is actually missing, and the default one won't work unless the group is finite. Maybe I should just add a method that catches the finite case, and returns  an error with explanation otherwise.

PR #2494, opened 2018-05-27
GASMAN: make some vars static; cleanup InitBags interface & docs

fingolfin:
The main change here is the removal of the `stack_func` argument from `InitBags`, which was never used. The only scenario in which it might be relevant is if some future system architecture we want to support is not supported by our current stack scanner. But I think we'd then simply put the new code into `gasman.c`; we wouldn't set a custom `stack_func`.

Note that this does not touch the `ExtraMarkFuncBags` which @markuspf added for libgap, it stays in place.

### DIFF
--- a/src/boehm_gc.c
+++ b/src/boehm_gc.c
@@ -314,7 +314,6 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
 }
 
 void InitBags(UInt              initial_size,
-              TNumStackFuncBags stack_func,
               Bag *             stack_bottom,
               UInt              stack_align)
 {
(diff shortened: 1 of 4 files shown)


fingolfin:
Hopefully it won't break anything for libgap -- as I said, `ExtraMarkFuncBags` is left intact, and AFAIK the rest I touched is not used by libgap. Of course it's a bit difficult for me to tell, as there is no PR for libgap nor do I know where I can find a "current" branch for it to check *shrug*.

(I did, however, recently update a version of libgap which @sebasguts gave to me to work with the current master branch; some other changes I made there some time ago caused trouble for it, but this was easy to resolve (and IMHO the result is better than what we had before).

Anyway, I hope @markuspf can review this PR then soon, and let us know if there are conflicts or not.

PR #2496, opened 2018-05-28
Finite groups of order not divisible by 4 are solvable

fingolfin:
This is a simple extension of the odd order theorem. See comment
inside the patch.


lib/grp.gi
```diff
@@ -550,10 +550,22 @@ InstallMethod( IsAlmostSimpleGroup,
...
+##  involution h; then \phi(h) decomposes into a product of m disjoint
+##  transpositions, hence sign(\phi(h)) = -1. permutation. Hence the kernel N
+##  of the composition x \mapsto sign(\phi(x)) is a normal subgroup of G of
+##  index 2, hence |N| = m.
+##
+##  By the odd order theorem, N is solvable, and so is G. Thus the order of
+##  any non-solvable finite group is a multiple of 4.
 InstallImmediateMethod( IsSolvableGroup, IsGroup and HasSize, 10,
     function( G )
     G:= Size( G );
```
stevelinton: I appreciate that this is not code added in the patch, but personally I really don't like
this reuse of G. Adding a new local variable would add one line of source and 8 bytes of stack frame.

fingolfin: Agreed.

fingolfin: and done

PR #2498, opened 2018-05-29
Allow a help expression to return a value

markuspf:
We introduced special syntax to access help in 60892e60, but didn't allow
a help expression to return a result.

This broke how help works in the JupyterKernel, this commits merely repairs
the previous behaviour.


fingolfin:
Backported to stable-4.9 in af8d800783e2a488072ef69eb76122d6b7cf6a84

PR #2499, opened 2018-05-30
Add more wordrep tests, and fix two bugs founds in the process

fingolfin:


fingolfin:
One thing to note is that `8Bits_Quotient`, `16Bits_Quotient` and `32Bits_Quotient` actually are never used. We might either install them as methods for `/`, or else remove them.

PR #448, opened 2016-01-13
Don't try and load dynamic modules in place of GAP files.

stevelinton:
Remove the feature where dynamically loaded compiled files woudl be automatically loaded 
in place of GAP files in the library and packages.  See Issue #60.

It's not used and it's complexity we don't need.

PR #2507, opened 2018-06-03
kernel: rewrite SyFindOrLinkGapRootFile, update comments

fingolfin:
Rewriting `SyFindOrLinkGapRootFile` extends simplification work begun by @stevelinton in PR #448. I also update relevant documentation comments to actually reflect current reality. Specifically, we (well, @stevelinton) removed the ability to load GAC compiled replacements for GAP library files *dynamically* -- we now only look for statically linked ones, like `src/c_oper1.c`. However, the documentation for `SyFindOrLinkGapRootFile` and various other places still referred to the dynamic versions. There also was a return code for that.

The first commit in this PR updates all these comments.

The second commit then completely rewrites `SyFindOrLinkGapRootFile` for clarity.

The third and final commit simplifies `READ_GAP_ROOT` by taking advantage of the fact that the return value for dynamlically linked files is now unused. It also corrects and expands a comment.

UPDATE: this PR also used to modify the `StructInitInfo` typedef, but that was moved to PR #2541

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -425,7 +425,7 @@ Int READ_GAP_ROOT ( const Char * filename )
             Pr( "#I  READ_GAP_ROOT: loading '%s' as GAP file\n",
                 (Int)filename, 0L );
         }
-        if ( OpenInput(result.pathname) ) {
+        if ( OpenInput(result.path) ) {
             while ( 1 ) {
                 ClearError();
                 Obj evalResult;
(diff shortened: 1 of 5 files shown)


fingolfin:
BTW, I guess the `StructInitInfo` change could be put into a separate PR, if somebody prefers that (I mostly put it in here as all the commits in here were created on the same branch geared towards refactoring our "module status init/destroy API").

And the remaining changes could be squashed into a single commit. But I figured it's easier to do that squashing upon request, than restoring the current step-by-step form ;-).

PR #2508, opened 2018-06-04
Warning update

ChrisJefferson:
This updates the list of warnings we use to the latest version, keeping our modifications.

I also disable -Wswitch-enum, as it doesn't let you use 'default' in a switch.

EDIT: There is one warning i needed to fix. This commit first had a big giant diff, now I have a much smaller one. While this does make this function slightly worse (i think), the warning is generally useful so I'd like to keep it.

### DIFF
--- a/cnf/m4/ax_compiler_flags_cflags.m4
+++ b/cnf/m4/ax_compiler_flags_cflags.m4
@@ -19,13 +19,14 @@
 # LICENSE
 #
 #   Copyright (c) 2014, 2015 Philip Withnall <philip@tecnocode.co.uk>
+#   Copyright (c) 2017, 2018 Reini Urban <rurban@cpan.org>
 #
 #   Copying and distribution of this file, with or without modification, are
 #   permitted in any medium without royalty provided the copyright notice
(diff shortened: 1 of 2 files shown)


cnf/m4/ax_compiler_flags_cflags.m4
```diff
@@ -69,12 +77,9 @@ AC_DEFUN([AX_COMPILER_FLAGS_CFLAGS],[
             -Wall dnl
             -Wextra dnl
             -Wundef dnl
-            -Wnested-externs dnl
```
fingolfin: First thought this got remove, but apparently it was just moved to a C specific section (though we always use a C compiler anyway, don't we?!?)

src/scanner.c
```diff
@@ -461,7 +463,7 @@ static UInt GetNumber(Int readDecimalPoint)
     }
 
     // If the next thing is the start of the exponential notation, read it now.
-    UInt seenExp = 0;
+
```
fingolfin: Why is that change necessary?

fingolfin: Ahh, nevermind, it's explained in the commit message.

PR #2509, opened 2018-06-04
Gzip support, take 2

ChrisJefferson:
This is a cleaned up attempt at adding gzip support to GAP.

I recommend reading this commit by commit. The changes are:

First two commits: PR #2508, to disable a warning in the new code

3rd commit: Add zlib to configure (note, this should / will be improved).

4th/5th commit: This abstracts out most uses of `read` and `write` to use a `SyRead` and `SyWrite` function. At this point this seems fairly useless, but it will make adding gzip support a much smaller change.

6th commit: Actually add gzip

7th commit: Testing


ChrisJefferson:
The 'DO NOT MERGE' is because I'm not happy with how I find zlib (that is, the user must have it installed, or else GAP won't build).

fingolfin:
Hmm, tests are failing (and issue labels are missing). Will wait with reviewing until this is resolved.

fingolfin:
Travis now fails, but AppVeyor still fails, because of missing zlib headers.

This of course would be fixed if zlib was bundled.

ChrisJefferson:
Some changes:

1) I have forced the built-in zlib to always use the static build. This is because zlib isn't libtool-aware, so while GAP builds and links, when GAP is loaded the extern/build/zlib directory isn't in the library path and GAP either goes off and uses /lib/zlib1.so (which could be bad), or just fails to load.

2) At the moment tests are failing because some packages (for example cvec) aren't picking up the zlib header location. However, the most recent round of configure changes seems to have fixed this (it works on cvec 2.6.1, released yesterday), so things should look much better in a couple of days.

configure.ac
```diff
@@ -349,6 +391,64 @@ dnl remove this eventually.
...
+  ])
+
+# Use bundled zlib if requested
+AS_IF([test x$BUILD_ZLIB = xyes],[
+  AS_IF([test x$CYGWIN = xyes],[
+    AC_MSG_ERROR([Builtin zlib not supported on cygwin. Install zlib-devel package])
+  ])
+  BUILD_ZLIB=yes
+  ZLIB_CPPFLAGS='-I${abs_builddir}/extern/install/zlib/include'
+  ZLIB_LDFLAGS='-L${abs_builddir}/extern/install/zlib/lib -l:libz.a'
```
fingolfin: `-l:libz.a` is not portable (in fact, I could find no documentation on it at all). But it's also not needed here: just change this to:
```
ZLIB_LDFLAGS='${abs_builddir}/extern/install/zlib/lib/libz.a'
```

configure.ac
```diff
@@ -349,6 +391,64 @@ dnl remove this eventually.
...
+          AC_MSG_NOTICE([No usable zlib found, switching to included zlib])
+        ],
+        [AC_MSG_ERROR([zlib not found at prefix $with_zlib])]
+        )
+    ])
+  ])
+
+# Use bundled zlib if requested
+AS_IF([test x$BUILD_ZLIB = xyes],[
+  AS_IF([test x$CYGWIN = xyes],[
```
fingolfin: It seems a bit overkill to move the a huge section of `configure.ac` just to check for cygwin here, esp. as that only requires matching `$host_os` against `*cygwin*`...

fingolfin:
This still has a "DO NOT MERGE" label, and I am not sure what it's status is, so I'll refrain from a full review.

Just one more comment: GitHub's UI shows that commits ordered by date, which here does not fit their actual order. Perhaps consider doing a `git rebase --ignore-date master` next time you update it?

configure.ac
ChrisJefferson: How do I do substring matching in configure scripts? Google isn't turning up anything useful.

ChrisJefferson: I eventually managed to track down you are supposed to use AC_CASE of course....

configure.ac
ChrisJefferson: changed

configure.ac
ChrisJefferson: Fixed

ChrisJefferson:
Now updated (hopefully)

src/sysfiles.c
```diff
@@ -84,15 +84,27 @@ typedef void sig_handler_t ( int );
 #include <mach-o/dyld.h>
 #endif
 
+#include <zlib.h>
+
 
 /* utility to check return value of 'write'  */
 ssize_t echoandcheck(int fid, const char *buf, size_t count) {
```
fingolfin: (I am not asking you to actually change this, but: wouldn't this have been one of the few cases were one really could have fixed the formatting of the whole function, given that you touch all of it except the first two and last two lines?)

src/sysfiles.c
```diff
@@ -865,26 +878,26 @@ Int SyFopen (
...
                ) ) {
-        syBuf[fid].pipe = 1;
-        syBuf[fid].fp = fileno(syBuf[fid].pipehandle);
-        syBuf[fid].ateof = 0;
-        syBuf[fid].crlast = 0;
-        syBuf[fid].bufno = -1;
-        syBuf[fid].isTTY = 0;
+       syBuf[fid].type = pipe_socket;
+       syBuf[fid].fp = fileno(syBuf[fid].pipehandle);
+       syBuf[fid].bufno = -1;
```
fingolfin: Weirdly, the indentation in this case is wrong by one space (but again, I am not really asking you to change this)

src/sysfiles.c
```diff
@@ -84,15 +84,27 @@ typedef void sig_handler_t ( int );
...
+
 
 /* utility to check return value of 'write'  */
 ssize_t echoandcheck(int fid, const char *buf, size_t count) {
   int ret;
+  if (syBuf[fid].type == gzip_socket) {
+      ret = gzwrite(syBuf[fid].gzfp, buf, count);
+      if (ret < 0)
+          ErrorQuit(
+              "Cannot write to compressed file, see 'LastSystemError();'\n",
```
fingolfin: Language nitpick: If I saw that message, I would think "ah, writing to compressed files is not supported"; but I think what really is meant here is that writing to a compressed file failed. Perhaps "Could not write" or "Failed to write" would be clearer (or maybe that's me as a non-native speaker just not getting it)?

(Again, not necessarily asking you to change this, just wondering)

src/sysfiles.c
```diff
@@ -84,15 +84,27 @@ typedef void sig_handler_t ( int );
...
+              0L, 0L);
+  }
+  else {
+      ret = write(syBuf[fid].echo, buf, count);
+      if (ret < 0)
+          ErrorQuit("Cannot write to file descriptor %d, see "
```
fingolfin: Same as above (but of course this message is old and predates this PR)

src/sysfiles.h
```diff
@@ -139,23 +139,36 @@ extern const Char * SyWinCmd (
...
-  int         fp;          /* file descriptor for this file. This is used
-                              for input */
-  int         echo;        /* file descriptor for the echo. This is used
-                              for output. Is set to 0 if read-only file */
-  UInt        pipe;        /* file is really a pipe */
-  FILE       *pipehandle;  /* for pipes we need to remember the file handle */
-  UInt        ateof;       /* set to 1 by any read operation that hits eof
+    /* gzfp is used if type == gzip_socket
+    ** Stored as a void* to avoid including zlib.h */
+    void* gzfp;
```
fingolfin: Formatting: space before `*` (but we can also fix this afterwards -- indeed, with your PR merged, I think we are close to making `syBuf` private to `sysfiles.h`, then we could move this out of the header and into the .c file, and take care of formatting in this struct overall)

ChrisJefferson: Yes, we could try moving this into sysfiles.h later. I did at some point run clang-format on this part, but it totally changed the whole struct and I wanted it to be clear what I'd added/changed.

ChrisJefferson:
@fingolfin , while many of reformatting comments are reasonable, I'd prefer to merge this and then make cleanups later, partly because it will be a pain to figure out which changes have to go in which of the existing commits.

fingolfin:
I'd merge this now but there was a build failure, seemingly due to a network problem. I just restarted that job, if it passes, we can merge.

PR #2511, opened 2018-06-04
HPC-GAP: explain about use of HPC-GAP code in kernel and library; and restrict some of that code

fingolfin:
This is partially inspired by a discussion with @ThomasBreuer some time ago on the mailing list, on the use of HPC-GAP specific code inside `lib/` (though I did not have that exchange at hand when writing this, so I may have forgotten some stuff we discussed back then).

I hope that `README.hpcgap.md` now is a bit more helpful. Thomas, I'd appreciate if you could take a look at let me know what you think. Also, suggestions for further things that should be explained are welcome.

At the same time, I restricted some HPC-GAP kernel code, forcing it to be used inside `#ifdef HPCGAP`, to make sure `README.hpcgap.md` actually says correct things ;-).


fingolfin:
@ThomasBreuer Thanks for the corrections. Note that you can also leave line comments in a pull request, which IMHO is quite convenient for this kind of feedback; see https://help.github.com/articles/commenting-on-a-pull-request/

PR #2512, opened 2018-06-04
kernel: make module state initialization order consistent

fingolfin:
This PR ensures that the order in which the various module initialization functions are called is consistent between the main thread, and other threads.

Specifically, with this PR, we (still) always call `InitKernel` first (and only once per module, on the main thread); then `InitLibrary` (also only once and on the main thread); and then `InitModuleState` (once per thread). 

In contrast, previously we called `InitLibrary` *after* `InitModuleState` on the main thread; but on all other threads, of course `InitModuleState` was by necessity called after the single `InitLibrary` call during startup.

OLD DESCRIPTION FOLLOWS:

This changes the way we handle module state initialization: `InitModuleState` is now a member of `StructInitInfo`, just like `InitKernel`, `InitLibrary` etc. -- and so are the size of the module state (if any) etc.. As a result, `RegisterModuleState` can be removed, and module state handling is now similar to all other aspects of the module "API".

This is work in progress; the commits are not yet cleaned up, some comments still need to be adapted, moved and written etc. I mainly submit it here already now to (a) get some test feedback from Travis, and (b) to get some early feedback on this: no point in wasting time on cleaning this up if it has no chance of ever being merged anyway...

Contains PR #2507

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1774,7 +1774,6 @@ void InitializeGap (
 #ifdef HPCGAP
     InitMainThread();
 #endif
-    ModulesInitModuleState();
 
     InitGlobalBag(&POST_RESTORE, "gap.c: POST_RESTORE");
     InitFopyGVar( "POST_RESTORE", &POST_RESTORE);
@@ -1813,9 +1812,9 @@ void InitializeGap (
 
     /* if we are restoring, load the workspace and call the post restore   */
     if ( SyRestoring ) {
+        ModulesInitModuleState();
         LoadWorkspace(SyRestoring);
         SyRestoring = NULL;
-
 
         /* Call POST_RESTORE which is a GAP function that now takes control, 
            calls the post restore functions and then runs a GAP session */
@@ -1836,6 +1835,8 @@ void InitializeGap (
 
         SyInitializing = 1;
         ModulesInitLibrary();
+        ModulesInitModuleState();
+
     }
 
     /* check initialisation                                                */
--- a/src/objcftl.c
+++ b/src/objcftl.c
@@ -368,18 +368,6 @@ static Int InitKernel (
 
 /****************************************************************************
 **
-*F  PostRestore( <module> ) . . . . . . . . . . . . . after restore workspace
-*/
-static Int PostRestore (
-    StructInitInfo *    module )
-{
-    /* return success                                                      */
-    return 0;
-}
-
-
-/****************************************************************************
-**
 *F  InitLibrary( <module> ) . . . . . . .  initialise library data structures
 */
 static Int InitLibrary (
@@ -445,7 +433,6 @@ static StructInitInfo module = {
     .name = "objcftl",
     .initKernel = InitKernel,
     .initLibrary = InitLibrary,
-    .postRestore = PostRestore,
 
     .moduleStateSize = sizeof(struct CFTLModuleState),
     .moduleStateOffsetPtr = &CFTLStateOffset,
--- a/src/stats.c
+++ b/src/stats.c
@@ -1582,19 +1582,6 @@ void            PrintReturnVoid (
 */
 
 
-
-/****************************************************************************
-**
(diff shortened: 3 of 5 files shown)


src/gapstate.c
```diff
@@ -7,72 +7,4 @@
...
-    Int offset;
-    ModuleConstructor constructor;
-    ModuleDestructor destructor;
-} StateDescriptor;
-
-static StateDescriptor StateDescriptors[STATE_MAX_HANDLERS];
-static Int StateDescriptorCount = 0;
-static Int StateNextFreeOffset = 0; // Start of next free memory area (as offset into GAPState.StateSlots)
-
 GAPState MainGAPState;
```
fingolfin: This is the only non-preprocessor non-comment line remaining in `gapstate.c` -- perhaps we should remove the file, and move this line somewhere else (e.g. `gap.c`?)

src/gapstate.c
```diff
@@ -7,72 +7,6 @@
...
-    ModuleConstructor constructor;
-    ModuleDestructor destructor;
-} StateDescriptor;
-
-static StateDescriptor StateDescriptors[STATE_MAX_HANDLERS];
-static Int StateDescriptorCount = 0;
-static Int StateNextFreeOffset = 0; // Start of next free memory area (as offset into GAPState.StateSlots)
-
+#ifndef HPCGAP
 GAPState MainGAPState;
```
fingolfin: This is the only non-preprocessor non-comment line remaining in `gapstate.c` -- perhaps we should remove the file, and move this line somewhere else (e.g. `gap.c`?)

markuspf:
This looks good to me, and I wanted to put the module state into the init info struct from the beginning.

The reason that I didn't was that I didn't understand why Reimer made the module state live separately from these structs for HPC-GAP; so essentially if we still care about HPC-GAP, and these patches dont break it, I am happy. (Also if we don't care about HPC-GAP, then there's no objection at all).

I also agree with removing the `gapstate.c` file as it doesn't contain any useful code anymore.

fingolfin:
@markuspf Thanks for the feedback. This code was written with HPC-GAP in mind, and seems to work there just fine. I can only speculate why @rbehrends wrote the original code the way he did; my guess would be that he simply was not aware of the `StructInitInfo` code and its role, and/or that it was less of a fit back then (we refactored it quite a bit back then). In any case, I strongly doubt that there was a deep fundamental reason.

src/gap.c
```diff
@@ -1772,7 +1772,11 @@ void InitializeGap (
 #ifdef HPCGAP
     InitMainThread();
 #endif
-    InitGAPState(ActiveGAPState());
+    ModulesInitModuleState();
+    // FIXME: calling ModulesInitModuleState before InitLibrary has a weird effect:
+    // it means that the first time the `InitModuleState` function of a module is
+    // called is *before* its `InitLibrary` function has been run; but all other
+    // invocations, in other threads, of course are run *after* `InitLibrary`.
```
fingolfin: Note: The issue described in the above comment is the main blocker for this PR, at least from my point of view. To resolve it, we may have to audit all `InitModuleState` functions, resp. all `InitLibrary functions of related modules.

However, it may very well be that some modules A, B are such that B's `InitLibrary` requires that A's `InitLibrary` *and* `InitModuleState` have already run. This suggest another, relatively simple fix: simply call `InitModuleState` right after each `InitLibrary`, from inside `ModulesInitLibrary`. I haven't tested this yet, so it may also run into problems; my point is more that there are more avenues to investigate.

fingolfin:
Workspaces are currently broken in this PR

fingolfin:
This could be reviewed and merged now.

fingolfin:
@ChrisJefferson @markuspf this old PR is finally ready to merge; but I am afraid it might be lost in the "graveyard" of old PRs nobody really looks at anymore... ;-)

PR #2514, opened 2018-06-05
kernel: add missing IntrReturning check to IntrNot

fingolfin:
Using `IntrReturning` or `IntrIgnoring` is a possible way to use the parser without executing code. But that requires that we really always check both in every `Intr*` function.


fingolfin:
It's not so surprising: it only matters if one uses `return` in interpreted mode, i.e., only in a break loop. Perhaps like this: `return; not true;`

PR #2515, opened 2018-06-05
More scanner refactoring

fingolfin:
Motivation: this (IMHO) simplifies the code parsing strings, i.e., makes it more readable. It also encapsulates the handling of the value buffer, meaning that if we ever want to change the buffer size, doing so is trivial (on does not have to adjust checks in multiple places, nor use strange checks like `if (i + 2 >= sizeof(buf))`

### DIFF
--- a/src/scanner.h
+++ b/src/scanner.h
@@ -207,7 +207,6 @@ typedef UInt            TypSymbolSet;
 **  'GetIdent' truncates an identifier after that many characters.
 */
 
-#define         SAFE_VALUE_SIZE 1024
 
 /****************************************************************************
 **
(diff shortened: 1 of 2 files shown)


fingolfin:
This should be ready for review now. Sorry for the confusion with earlier buggy versions.

src/scanner.h
```diff
@@ -216,7 +216,6 @@ typedef UInt            TypSymbolSet;
 */
```
ssiccha: `We only fill 'Value' up to SAFE_VALUE_SIZE normally.` above should then also be changed to `We only fill 'Value' up to MAX_VALUE_LEN normally.`

fingolfin: Thanks for looking very carefully, I appreciate it! But actually, that `SAFE_VALUE_SIZE` is not there anymore, as it has been removed in master already. But I'll rebase this PR to "prove" it ;-).

ssiccha: Haha. OK :)

PR #2516, opened 2018-06-05
Add zlib to extern (DO NOT MERGE)

fingolfin:
This adds zlib 1.2.11 to the GAP git repository, as well as build system modifications to make use of a system installed zlib, with a fallback to ours. I expect that PR #2509 will be rebased on this. As such, this PR likely should not be merged independently. But we can discuss this tomorrow.


ChrisJefferson:
This is subsumed by #2509, so I think can be closed.

PR #2519, opened 2018-06-06
Change Test() to reject unexpected or invalid inputs

### LINKED ISSUE #2506: Print an alert when trying to test a non-test file
While working on a GAP package I tried to test a file called `testparallel.g` which itself already calls `TestDirectory()`. As a GAP beginner I didn't know that this file should be called by writing `Read("testparallel.g")` and wouldn't have found out if I wasn't explicitely told about it.
Maybe we could make this part easier for beginners by printing a warning when a non-test file is called by `Test()`?
Me and @ssiccha where thinking about something like ```The tested file is not a test file and should preferably be called with Read().```

Would it be easy to distinguish between test files and non-test files? What do you think about this suggestion?

fingolfin:
This is a somewhat experimental change to fix #2506 -- let's see if the GAP test suite passes with it.

Assuming this works for the test suite, I wonder if there are any reasonable scenarios where the new `Error` call is triggered and where this is undesirable. If anybody knows such an example, please bring it up. In that case, we could add an option to turn off that error; or restrict the cases in which it is triggered.

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -63,6 +63,7 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments)
       Add(outp[Length(outp)], '\n');
       i := i+1;
     else
+      Error("Invalid data in test file");
       i := i+1;
     fi;
   od;


ChrisJefferson:
I suggest we merge this, and then see if any packages fail

PR #2520, opened 2018-06-06
Overhaul tracking of current statement

### LINKED ISSUE #616: "no method found" points to the wrong line of code
When I run

``` gap
f := function() 
  local l;
  l := 0 * [1..6];
  l[[1..3]] := 1;
end;
f();
```

I get the error msg

``` gap
Error, no 1st choice method found for `[]:=' on 3 arguments called from
0 * [ 1 .. 6 ] at stream:3 called from
<function "f">( <arguments> )
 called from read-eval loop at line 5 of stream
...
```

although `l[[1..3]] := 1;` is the offending part of code.
This does not happen with my version of GAP 4.7.9.

### LINKED ISSUE #2656: Error message <corrupted statement> when using `[i,j]` with invalid indices
In GAP 4.9.1 and in the current master branch I get:
```
gap> l := [[1]];
[ [ 1 ] ]
gap> f := {} -> l[2,1];
function(  ) ... end
gap> f();
Error, List Element: <list>[2] must have an assigned value in
  return m[i][j]; at /home/sergio/projects/gap-master/lib/matrix.gi:30 called from 
<corrupted statement>  called from
<function "f">( <arguments> )
brk> 
```

Is it worth the effort to let GAP show the line from where the offending `l[2,1]` was called?

fingolfin:
This PR unifies the code we use to track the current statement and expression. It resolves various bugs where backtraces indicated the wrong statement as the current one, ans also gets rid of some cases where "<corrupt statement>" was printed.

This needs more work on the long run (in future PRs), in particular for the case where there is an error in the condition expression of an `if` statement or a loop: then what should we print? Just the expression in which the error occurred? But that could be a simple integer (in `if 1 then ... fi`), which is not helpful (and in this particular case, would even lead to a "corrupt statement" message for technical reasons).

Or perhaps the whole statement should be printed? But that could span dozens of lines. Slightly better is to print, say, a while loop as `while COND do ... od;`, i.e. replace the body by `...`. But (a) the condition still could be multiple lines long, and (b) this does not work very well for an `if/elif/elif/.../else/fi` with multiple branches. 

Or "just the line on which the error occurred" -- but that then depends on the on-disk representation of the data, which opens its own can of worms, so I'd rather avoid it.

For now, I think my favorite would be to just print `while COND do` (no `...` or `od;`), resp. `if COND then` or `elif COND then` etc. But we lack the facilities for that -- esp. for the `elif` we currently don't even have place where to store on which of the various "branches" of a "multi-condition if statement" we are.

So, this needs work, but not today... I plan to turn the above into an issue later.

Resolves #616
Fixes #2656

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -219,7 +219,6 @@ Obj FuncCALL_WITH_CATCH(Obj self, Obj func, volatile Obj args)
     volatile Obj       currLVars;
     volatile Obj       tilde;
     volatile Int       recursionDepth;
-    volatile Stat      currStat;
 
     if (!IS_FUNC(func))
         ErrorMayQuit(
@@ -238,7 +237,9 @@ Obj FuncCALL_WITH_CATCH(Obj self, Obj func, volatile Obj args)
     memcpy((void *)&readJmpError, (void *)&STATE(ReadJmpError),
            sizeof(syJmp_buf));
     currLVars = STATE(CurrLVars);
-    currStat = STATE(CurrStat);
+#ifdef GAP_KERNEL_DEBUG
+    volatile Stat currStat = BRK_CALL_TO();
+#endif
     recursionDepth = GetRecursionDepth();
     tilde = STATE(Tilde);
     res = NEW_PLIST_IMM(T_PLIST_DENSE, 2);
@@ -253,7 +254,7 @@ Obj FuncCALL_WITH_CATCH(Obj self, Obj func, volatile Obj args)
         CHANGED_BAG(res);
         STATE(ThrownObject) = 0;
         SWITCH_TO_OLD_LVARS(currLVars);
-        STATE(CurrStat) = currStat;
+        GAP_ASSERT(currStat == BRK_CALL_TO());
         SetRecursionDepth(recursionDepth);
         STATE(Tilde) = tilde;
 #ifdef HPCGAP
@@ -380,7 +381,7 @@ Obj CallErrorInner(const Char * msg,
     l = NEW_PLIST_IMM(T_PLIST_HOM, 1);
     SET_ELM_PLIST(l, 1, EarlyMsg);
     SET_LEN_PLIST(l, 1);
-    SET_BRK_CALL_TO(STATE(CurrStat));
+
     // Signal functions about entering and leaving break loop
     for (i = 0; i < ARRAY_SIZE(signalBreakFuncList) && signalBreakFuncList[i];
          ++i)
--- a/src/stats.c
+++ b/src/stats.c
@@ -209,7 +209,6 @@ UInt            ExecIf (
     Stat                body;           /* body                            */
 
     /* if the condition evaluates to 'true', execute the if-branch body    */
-    SET_BRK_CURR_STAT( stat );
     cond = READ_STAT(stat, 0);
     if ( EVAL_BOOL_EXPR( cond ) != False ) {
 
@@ -230,7 +229,6 @@ UInt            ExecIfElse (
     Stat                body;           /* body                            */
 
     /* if the condition evaluates to 'true', execute the if-branch body    */
-    SET_BRK_CURR_STAT( stat );
     cond = READ_STAT(stat, 0);
     if ( EVAL_BOOL_EXPR( cond ) != False ) {
 
@@ -260,7 +258,6 @@ UInt            ExecIfElif (
     for ( i = 1; i <= nr; i++ ) {
 
         /* if the condition evaluates to 'true', execute the branch body   */
-        SET_BRK_CURR_STAT( stat );
         cond = READ_STAT(stat, 2 * (i - 1));
         if ( EVAL_BOOL_EXPR( cond ) != False ) {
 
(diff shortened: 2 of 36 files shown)


tst/test-error/debug-var.g.out
```diff
@@ -109,7 +109,7 @@ brk_2> IsBound(y);
 true
 brk_2> unbound_higher;
 Error, Variable: <debug-variable-64-4> must have a value in
-  <corrupted statement>  called from 
+  Error( "foobar" ); at *stdin*:20 called from
```
fingolfin: This is a clear improvement, I'd say

tst/test-error/method-not-found.g.out
```diff
@@ -2,7 +2,7 @@ gap> # test returning from a 'method not found' error
 gap> f:=a->a+a;; f(());
 Error, no method found! For debugging hints type ?Recovery from NoMethodFound
 Error, no 1st choice method found for `+' on 2 arguments at GAPROOT/lib/methsel2.g:250 called from
-a + a at *stdin*:3 called from
+return a + a; at *stdin*:3 called from
```
fingolfin: Hm.... Is this new behaviour OK? Or do we want the old back? Opinions, anybody?

tst/test-error/debug-var.g.out
ChrisJefferson: Yup!

fingolfin:
There is still much more that should be done on this subject, but I think instead of letting this rot, we should just merge it. To this end, I cleaned up the PR and rebased it again.

src/error.c
```diff
@@ -253,7 +254,7 @@ Obj FuncCALL_WITH_CATCH(Obj self, Obj func, volatile Obj args)
         CHANGED_BAG(res);
         STATE(ThrownObject) = 0;
         SWITCH_TO_OLD_LVARS(currLVars);
-        STATE(CurrStat) = currStat;
+        GAP_ASSERT(currStat == BRK_CALL_TO());
```
fingolfin: This is a paranoia check, to verify I did not break anything fundamental in the refactoring. It could/should likely be removed eventually.

ChrisJefferson: I personally don't think you can have too many asserting paranoia checks :)

src/stats.c
```diff
@@ -689,8 +672,12 @@ static ALWAYS_INLINE UInt ExecWhileHelper(Stat stat, UInt nr)
     body2 = (nr >= 2) ? READ_STAT(stat, 2) : 0;
     body3 = (nr >= 3) ? READ_STAT(stat, 3) : 0;
 
+    // HACK: point current "statement" at the condition expression, so
+    // that if there is an error in it, it resp. its location is printed,
+    // not the whole loop statement.
+    SET_BRK_CALL_TO(cond);
```
fingolfin: We might want to add a few more of these "hacks" for now. The purpose of these hacks is as follows: Without them, you might see:
```
Error, Rational operations: <divisor> must not be zero in
  if true = 1 / 0 then
    return 1;
fi; at *stdin*:9 called from
```
with the hacks, you get:
```
Error, Rational operations: <divisor> must not be zero in
  1 / 0 at *stdin*:9 called from
```
I.e. it's the difference between printing the condition expression of a loop or `if` clause, vs. printing the whole loop resp. `if` statement. Now, a better approach probably would be to either print the actual line(s) containing the expression (as e.g. GDB would do it); or else add a way to print just the "head" of a loop or `if` (resp.: print the loop, but with the body replaced by, say `...`). The former can be difficult because a single expression can of course span many lines, including comments; also, the source file may be modified, leading to output that's not right (just as can happen with gdb); if we wanted to avoid that, we'd have to store the whole file in memory.

The second approach seems much simpler to me, and shouldn't be too hard to implement... (famous last words)

fingolfin: BTW, these HACKS are also limited in so far as they break down if `cond` is an "immediate" expression, i.e., an `INTEXPR` or a `REFLVAR`

I'll look into implementing the second approach now

tst/testspecial/backtrace.g.out
```diff
@@ -0,0 +1,167 @@
...
+brk> Where();
+<function "f">( <arguments> )
+ called from read-eval loop at *errin*:1
+brk> quit;
+gap> 
+gap> 
+gap> f:=function() if 1 < 0 then return 1; elif 1 then return 2; fi; return 3; end;;
+gap> f();
+Error, <expr> must be 'true' or 'false' (not a integer) in
+  1 < 0 at *stdin*:18 called from
```
fingolfin: This is wrong, and right now I am not quite sure why it is...

But what *is* the right thing to print? In a previous comment, I suggested that for loops and if clauses, one could just hide the "bodies", and so e.g. print `while 1 do ... od;` instead of the full loop. But for `if/elif/fi` clauses with multiple conditions, this still doesn't seem quite good enough: ideally, we only want to print the `if` resp. `elif` with the expression that was executed. But there is no separate statement for that -- we are "in the middle" of a single statement here! I don't see a simple solution for that right now, so I opted to drop both approaches I tried from this PR right now, to be revisited in a future PR. For now, we are at least not doing worse than we were before.

fingolfin:
I have rebased this now again, to resolve conflicts. Would be nice if this could be either merged, or rejected, so that I can move on. My guess is that nobody is looking at it because it is has fallen of the first page of PRs. But if @ChrisJefferson or @stevelinton (who wrote a lot of the affected code) could just have a look at it...

Issue #2513, opened 2018-06-05
Method order does not change when filter ranks change, leading to subtle conflict between nq and lpres


fingolfin:
We recently observed a test failure with the `nq` tests after `LoadAllPackages()` had run. Here is the cause:

First off, the `lpres` package installs another method for `NilpotentQuotient` but at a slightly lower priority (adjusting the rank by -1):
```gap
InstallOtherMethod( NilpotentQuotient,
  "for an FpGroup using the LPRES-package", true,
  [ IsFpGroup, IsPosInt ], -1, # give priority to NQ package
  function( G, c )
  return( NilpotentQuotient( Range( IsomorphismLpGroup( G ) ), c ) );
  end);
```
If you just load nq and lpres, there is no problem.

But if you first load nq; then load e.g. semigroups, then load lpres, suddenly the lpres method for `NilpotentQuotient` will be preferred over nq's -- because semigroups installs implications for `IsFpGroup`, which increases the rank of `IsFpGroup`, which increases the rank of  the above `NilpotentQuotient` at the moment it is installed...

So a quick workaround would be to change the -1 above to a lower value. In plain GAP, the rank of `IsFpGroup` right now seems to be 38. After `LoadAllPackages` it has rank 111; so we'd need at least something like -80 or so to make it "safe", but of course if enough implications for `IsFpGroup` is installed, any fixed rank adjustment could become too small.

IMHO in this particular case, the proper fix is to just delete that method from lpres - I see no point in it anyway, as no normal user would ever be able to call it that way (at least once the issue reported here has been resolved). Somebody who wants to compare the lpres nq implementation with that in the nq package could / should just use `NilpotentQuotient( Range( IsomorphismLpGroup( G ) ), c )`.  @laurentbartholdi any objections to that?

But this still leaves the underlying issue here, which is is that the rank of filters can change, but we never adjust the rank of methods based on that, once they have been installed, nor their order.

One fix would be to reorder all methods of all operations whenever a new implication is installed, or some variation thereof. Of course this may expose more issues in method installation which used a hard coded rank adjustment. We'd have to adjust them one by one.

stevelinton:
Running over all the methods installed (after starting master with standard packages) there are something like 240 places where the relative rank of methods changes when you call `WITH_HIDDEN_IMPS_FLAGS` on all the requirements.  In most cases it looks like the methods which switch apply to disjoint sets of objects, for instance permutation group methods now rank slightly higher than matrix group methods.

[...]

PR #2521, opened 2018-06-06
Reorder methods after new implications are added

stevelinton:
This is the start of an attempt to address #2513. At this stage it provides a function `RECALCULATE_ALL_METHOD_RANKS()` which does what it says. It is not called automatically. It still has a debugging printout when method ordering changes. It only seems to break one thing in `testinstall`


lib/oper1.g
```diff
@@ -144,21 +144,26 @@ fi;
...
         # TODO: once the GAP compiler supports 'atomic', use that
         # to replace the explicit locking and unlocking here.
         lk := WRITE_LOCK(METHODS_OPERATION_REGION);
     fi;
     # add the number of filters required for each argument
+    if IS_FUNCTION(rank) then
+        ranknow := rank();
+    else 
+        ranknow := rank;
+    fi;
```
fingolfin: Perhaps rename rank to baserank, and ranknow to rank, in order to reduce the diffs?

stevelinton:
So the problem (or at least the first one) turns out to be for ` IsomorphismPermGroup` where the method for permutation groups overtakes the method for finite nilpotent groups. I'm not sure why exactly this causes the computation to fail, and that is almost certainly a bug that needs fixing.

stevelinton:
This is now working, as far as I can see, although the diagnostic printout needs to be removed before it is merged.  The method rankings are recomputing to respect current implications: 

- when the library and its needed/recomended packages are loaded and before reading files from the command-line
- after loading a package (except that LOAD_ALL_PACKAGES only does it once at the end)
- any other time when the implications caches are being refreshed

This takes about 60ms (on my laptop) in a standard GAP session if there is nothing to do (I had to expand a couple of caches to get it down to that) but about 1.5second with all packages (more or less) loaded. 

Documentation and possibly tests are also still due.

fingolfin:
Tests fail: first off, `make docomp` needs to be run in HPC-GAP, too (it has its own set of `c_*.c` files). Secondly, `testspecial` is broken, because those tests "see" all those debug outputs.

lib/package.gi
```diff
@@ -1626,12 +1635,14 @@ InstallGlobalFunction( LoadPackage, function( arg )
...
-    	List( RecNames( GAPInfo.PackagesInfo ), LoadPackage );
-    fi;	
+    InstallGlobalFunction( LoadAllPackages, function()
+        SuspendMethodReordering();        
+        if ValueOption( "reversed" ) = true then
+            List( Reversed( RecNames( GAPInfo.PackagesInfo ) ), LoadPackage );
+        else
+            List( RecNames( GAPInfo.PackagesInfo ), LoadPackage );
+        fi;	
+        ResumeMethodReordering();
```
fingolfin: This change of indentation increases the diff needlessly... It seems the only actual changes was that a call to `SuspendMethodReordering` was inserted?

fingolfin: Oh and also ResumeMethodReordering

lib/oper.g
```diff
@@ -1922,6 +1917,96 @@ BIND_GLOBAL( "FLUSH_ALL_METHOD_CACHES", function()
...
+                fi;
+                for j in [1..n] do
+                    req := meths[base+1+j];
+                    rank := rank+RankFilter(WITH_IMPS_FLAGS(req));
+                od;
+                if rank = meths[base+n+3] then
+                    continue;
+                fi;
+                meths[base+n+3] := rank;
+                if i = 1 or rank <= meths[base-3] then
```
fingolfin: Why -3? Wait, is that trying to access the previous method's entry? But that could break if `BASE_SIZE_METHODS_OPER_ENTRY` changes again, no? It's also not immediately clear which "field" `base-3` refers to. While this makes it clear: `meths[base+n+3 - BASE_SIZE_METHODS_OPER_ENTRY]` (I'd still ad a comment to explain what's going on, though?)

doc/ref/methsel.xml
```diff
@@ -252,6 +252,7 @@ one can install also <E>immediate methods</E>.
 <Heading>Logical Implications</Heading>
 
 <#Include Label="InstallTrueMethod">
+<#Include Label="MethodReordering">
```
fingolfin: Do we really want to make this publicly documented? Why?

lib/filter.g
```diff
@@ -171,6 +171,60 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
 end );
 
 
+#############################################################################
+##
+#F SuspendMethodReordering()
+#F ResumeMethodReordering()
+#F ResetMethodReordering()
```
fingolfin: For consistency with other files, there should be two spaces after each `#F`, and also `()` should be `( )` (but we'll survive w/o this changed, too ;-)

lib/filter.g
```diff
@@ -171,6 +171,60 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
...
+##
+##  <#GAPDoc Label="MethodReordering">
+##  <ManSection>
+##  <Func Name="SuspendMethodReordering" Arg=""/>
+##  <Func Name="ResumeMethodReordering" Arg=""/>
+##  <Func Name="ResetMethodReordering" Arg=""/>
+##
+##  <Description> These functions control whether the method reordering process
```
fingolfin: Virtually every other GAPDoc comment in the library has `<Description>` and `</Description>` on their own lines. Just sayin'

lib/filter.g
```diff
@@ -171,6 +171,60 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
...
+##  described in <Ref Func="InstallTrueMethod"/> is invoked or not. Since this 
+##  process can be comparatively time-consuming, it is usually suspended when
+##  a lot of implications are due to be installed, for instance when loading
+##  the library, or a package. This is done by called <C>SuspendMethodReordering()</C>
```
fingolfin: called -> calling? Also, the recommend usage is that *every* reference to a function should use `<Ref Func=...>`, even if it is inside the body of the documentation of the function itself.

lib/package.gi
stevelinton: Thanks, will fix.

lib/oper.g
stevelinton: Will do

lib/filter.g
```diff
@@ -171,6 +171,60 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
...
+##  once the installations are done, <C>ResumeMethodReordering()</C> should be called. 
+##  These pairs of calls can be nested. When the outermost pair is complete, method
+##  reordering takes place and is enabled in <C>InstallTrueMethod</C> thereafter.
+##
+##  <C>ResetMethodReordering()</C> effectively exits all nested suspensions, resuming
+##  reordering immediately. This function is mainly provided for error recovery and 
+##  similar purposes.</Description>
```
fingolfin: Indeed, what if there is an error duing package loading? That happens frequently for me if I try `LoadAllPackages()` (because I have tons of non-distributed packages and development versions, with all kinds of conflicts). For options, we have the `OnQuit` hook. Perhaps a call to `ResetMethodReordering` should be added to `OnQuit`, too? (Both in the GAP and in the HPC-GAP version).

lib/filter.g
```diff
@@ -209,18 +263,35 @@ end );
...
-##  <Ref Func="IsCommutative"/>.
+##  <Ref Func="IsCommutative"/>.<P/>
+##
+##  Adding logical implications can change the rank of filters 
+##  (see <Ref Func="RankFilter"/>) and consequently the rank, and so choice of
+##  methods for operations (see <Ref Sect="Applicable Methods and Method Selection"/>).
+##  By default <C>InstallTrueMethod</C> adjusts the method selection data structures 
+##  to take care of this, but this process can be time-consuming, so functions
+##  <Ref Func="SuspendMethodReordering"/> and <Ref Func="ResumeMethodReordering"/>
+##  are provided to allow control of this process.
```
fingolfin: I think it's good to document this somewhere. But do we really want to document it for the public? I.e. perhaps move this out of the GAPDoc comment; or put it inside an XML comment?

lib/filter.g
```diff
@@ -209,18 +263,35 @@ end );
...
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
+
+
```
fingolfin: Why insert these three empty lines?

lib/oper.g
```diff
@@ -1922,6 +1917,96 @@ BIND_GLOBAL( "FLUSH_ALL_METHOD_CACHES", function()
 end);
 
 fi;
+#############################################################################
```
fingolfin: Insert an empty line before this comment starts?

src/opers.c
```diff
@@ -1946,6 +1946,9 @@ static ALWAYS_INLINE Obj GetMethodUncached(
         // entry n+3 is the rank
         // entry n+4 is the info text
         // entry n+5 is, if set, the location where the method was installed
+        // entry n+6 is, if set, the relative rank that was supplied when
+        //                       the method was installed, either as a small integer or a function
+        //                       of no arguments
```
fingolfin: That's a somewhat odd wrapping (looking at the differing line lengths); also, could perhaps align at "if set"?

And perhaps should say: "a function of no arguments, which returns a small integer to be used as rank"?

This use of such a function as "base rank" should ideally be documented somewhere, too, no? And I guess we'd also have to go through the library to search for places where this should be used...

tst/testinstall/method-reordering.tst
```diff
@@ -0,0 +1,37 @@
...
+true
+gap> ResumeMethodReordering();
+gap> Unbind(CheckReorder);
+gap> STOP_TEST("method-reordering.tst");
+
+ 
+
+
+             
+
```
fingolfin: You may need to remove these trailing empty lines, at least if/once this is rebased on master (as we added tests for "malformed" test files -- not sure if they accept trailing empty lines... perhaps they do, after all)

stevelinton:
@fingolfin thanks for the review. I'll pick up the small things after I've fixed constructors and actually supported the function for the rank adjustment in InstallMethod (which I've just realised I need to do).

If we're going to do this, I do think it should be documented publicly (although it doesn't need a huge song and dance) because there is a small but real risk of two things:

1. Package code or future library development installs implications which change the ranking of existing library methods  causing surprises
2. Some package (homalg -- I'm looking at you) turns out to install lots of implications at "run time" after package loading and suddenly gets a lot slower. 

In either case I think it's better to be able to point people at documentation.

stevelinton:
Subject to the tests, I think this is now complete.

fingolfin:
@stevelinton well, tests fails... also, if you think this is good for review, I suggest removing the `WIP` from the PR title.

stevelinton:
The tests relate to issue #2594 .

fingolfin:
@stevelinton I fixed #2594 in PR #2604, so if approve and merge the latter, it should be possible to resume with this PR

stevelinton:
The only test now failing is testpackages and that seems to be failing to loaf float, which loads OK for me on my laptop in this branch. Anyone know what's happening.

alex-konovalov:
We seen this in float before, I am sure it's not related to this PR:
```
Loading float ... 
#W dlopen() error: /home/travis/build/gap-system/gap/pkg/float-0.9.1/bin/x86_6\
4-pc-linux-gnu-default64/float.so: undefined symbol: atexit
Error, module '/home/travis/build/gap-system/gap/pkg/float-0.9.1/bin/x86_64-pc-linux-\
gnu-default64/float.so' not found in
  if not LOAD_DYN( arg[1], false ) then
    Error( "no support for dynamic loading" );
fi; at /home/travis/build/gap-system/gap/lib/files.gd:583 called from 
```
I've restated it so see if this will stay - someone suggested that this may depend on underlying Travis CI setups, and does not happen each time.

stevelinton:
I'm completely mystified by the test failure. float won't load because "atexit" is not defined, but (a) it's not actually mentioned anywhere in float that I can see and (b) it's a standard library function. I also don't see how the changes in this PR can possibly be related.

fingolfin:
This is a problem with float, orvrsther, the binutils/linker version installed on Travis. It affected many PRs and also master, hence we disabled float in the tests now. If you rebase on latest master, this should go away

fingolfin:
@stevelinton any plans to rebase this (and perhaps squash some of the comments)?

fingolfin:
@stevelinton ping?

fingolfin:
Since I'd like to make some more changes in lib/oper.g, it would be good to finalize this. In order to make progress, I will push a cleaned up and rebased version of this PR as a fresh PR (if @stevelinton agrees, I can also later update this PR here instead, but for now, I don't want to modify his work w/o permission)

fingolfin:
See PR #2773

PR #2523, opened 2018-06-06
Miscellaneous kernel + doc changes

fingolfin:
This is a grab bag of changes, best to review this commit-by-commit

### DIFF
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -248,16 +248,10 @@ static inline void FinishAndCallFakeFuncExpr(void)
 */
 void IntrBegin ( Obj frame )
 {
-    Obj                 intrState;      /* old interpreter state           */
-
     /* remember old interpreter state                                      */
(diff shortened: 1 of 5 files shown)


src/intrprtr.c
```diff
@@ -322,8 +314,7 @@ ExecStatus IntrEnd (
     }
 
     // switch back to the old state
-    STATE(StackObj)  = ELM_PLIST(STATE(IntrState), 2);
-    STATE(IntrState) = ELM_PLIST(STATE(IntrState), 1);
+    STATE(StackObj) = PopPlist(STATE(IntrState));
```
fingolfin: Note that `PopPlist` not just removes the element from the list; it also overwrites the slot in the plist that used to store the pop'ed element by zero, which ensures that it can be garbage collected right away (instead of leaving a "fake" reference to it inside the plist).

PR #2524, opened 2018-06-07
Add warning a file given to Test contains no tests

ChrisJefferson:
This adds a simple warning when a test file is run which doesn't contain any tests. This is most common because someone has misformatted their test file.

Also add some very simple tests for Test. None of these tests check for failures, because failures contain the full path of the failing test, which will change depending on where GAP is installed.


markuspf:
Gna. Sorry I thought the tests had passed just now. I hope it didn't break :/

PR #2527, opened 2018-06-07
Added IsLIBGAP constant

sebasguts:
to detect wether GAP is run as a library. 

This PR contains #2526.

PR #2528, opened 2018-06-07
Added IsLIBGAP constant

sebasguts:
This constant can be set to true in the initialize_libgap function (future work)
to stop GAP from starting an interactive session

This replaces #2527

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -927,6 +927,17 @@ BindGlobal ("ShowSystemInformation", function ()
     ShowPackageInformation();
 end );
 
+
+#############################################################################
+##
+## Initialize the IsLIBGAP variable (if not done before). If this variable
+## is false, an interactive session will be started.
+## Otherwise no interactive session is started.
+##
+if not IsBound( IsLIBGAP ) then
+  BIND_CONSTANT( "IsLIBGAP", false );
+fi;
+
 #############################################################################
 ##
 ##  Finally, deal with the lists of global variables.
@@ -1016,7 +1027,9 @@ InstallAndCallPostRestore( function()
     od;
 end );
 
-if IsHPCGAP and THREAD_UI() then
+if IsLIBGAP then
+  # GAP is used as a library, do not start an interactive session
+elif IsHPCGAP and THREAD_UI() then
   ReadLib("hpc/consoleui.g");
   MULTI_SESSION();
 else
--- a/tst/testinstall/IsLIBGAP.tst
+++ b/tst/testinstall/IsLIBGAP.tst
@@ -1,0 +1,12 @@
+#############################################################################
+##
+#W  IsLIBGAP.tst                  GAP library               Sebastian Gutsche
+##
+##
+#Y  Copyright (C)  2018,  GAP Group
+##
+gap> START_TEST("IsLIBGAP.tst");
+gap> IsBound( IsLIBGAP );
+true
+gap> STOP_TEST( "IsLIBGAP.tst", 1);
+


lib/init.g
```diff
@@ -1019,8 +1019,11 @@ end );
 if IsHPCGAP and THREAD_UI() then
   ReadLib("hpc/consoleui.g");
   MULTI_SESSION();
-else
+elif not IsBound( IsLIBGAP ) or IsLIBGAP = false then
+  BIND_GLOBAL( "IsLIBGAP", false );
   SESSION();
+else
+  BIND_GLOBAL( "IsLIBGAP", true );
```
fingolfin: Let's use `BIND_CONSTANT` instead

lib/init.g
```diff
@@ -1019,8 +1019,11 @@ end );
 if IsHPCGAP and THREAD_UI() then
   ReadLib("hpc/consoleui.g");
   MULTI_SESSION();
-else
+elif not IsBound( IsLIBGAP ) or IsLIBGAP = false then
+  BIND_CONSTANT( "IsLIBGAP", false );
```
fingolfin: Test fail, partially because of your .tst file... e.g. in HPC-GAP, we will *not* define `IsLIBGAP`. Better is to do this *before* line 1019:
```gap
  if not IsBound(IsLIBGAP) then
    BIND_CONSTANT( "IsLIBGAP", false );
  fi;
```
and then change this check here back to `elif not IsLIBGAP`, and drop the `else` part.

tst/testinstall/IsLIBGAP.tst
```diff
@@ -0,0 +1,12 @@
+#############################################################################
+##
+#W  IsLIBGAP.tst                  GAP library               Sebastian Gutsche
+##
+##
+#Y  Copyright (C)  2018,  GAP Group
+##
+gap> START_TEST("IsLIBGAP.tst");
+gap> IsBound( IsLIBGAP );
+true
```
fingolfin: You may want to run `./gap tst/testinstall.g` before submitting an updated version of this PR, to make sure it really works.

sebasguts:
I updated the PR.

I now initialize the `IsLIBGAP` variable before the systems gvar list is created. Otherwise the test will not work (and I think `IsLIBGAP` should be a system gvar).

lib/init.g
```diff
@@ -927,6 +927,17 @@ BindGlobal ("ShowSystemInformation", function ()
     ShowPackageInformation();
 end );
 
+
+#############################################################################
+##
+## Initialize the IsLIBGAP variable (if not done before). If this variable is false, an interactive
```
fingolfin: Please line-wrap this (I think at 78 chars... anyway, at the same width as the `####...` line above.

lib/init.g
```diff
@@ -936,7 +947,6 @@ end );
 #T as system variables!
 ##
 
-
```
fingolfin: Please don't remove this line.

lib/init.g
```diff
@@ -927,6 +927,18 @@ BindGlobal ("ShowSystemInformation", function ()
     ShowPackageInformation();
 end );
 
+
+#############################################################################
+##
+## Initialize the IsLIBGAP variable (if not done before). If this variable
+## is false, an interactive session will be started.
+## Otherwise no interactive session is started.
+## This should be done before the system GVAR list is created.
```
fingolfin: Change "should" -> "must" Or rather, just drop that line?

Anyway, let's not worry about this and merge as-is; I only mention it in case we need to update this again, say due to a build issue.

lib/init.g
```diff
@@ -1018,8 +1030,11 @@ end );
 
 if IsHPCGAP and THREAD_UI() then
   ReadLib("hpc/consoleui.g");
-  MULTI_SESSION();
-else
+  if IsLIBGAP = false then
+    MULTI_SESSION();
+  fi;
+elif IsLIBGAP = false then
+  BIND_CONSTANT( "IsLIBGAP", false );
```
fingolfin: Actually, this call to `BIND_CONSTANT` is redundant, isn't it? After all, you just checked that value.

lib/init.g
```diff
@@ -1018,8 +1030,11 @@ end );
 
 if IsHPCGAP and THREAD_UI() then
   ReadLib("hpc/consoleui.g");
-  MULTI_SESSION();
-else
+  if IsLIBGAP = false then
```
fingolfin: Hmm... I was confused by the new logic for a moment (sorry for missing it in the previous version)... How about this
```gap
if IsLIBGAP then
   # GAP is used as a library, so don't start a session
elif IsHPCGAP and THREAD_UI() then
  MULTI_SESSION();
else
  SESSION();
fi;
```

sebasguts: I am not sure actually. If we ever decide to have a HPC-GAP libGAP, the way it is now we could leave this as it is.

lib/init.g
sebasguts: It was supposed to make `IsLIBGAP` constant, which does not work in this case anyway, so I will remove it.

lib/init.g
sebasguts: On the otherhand, nevermind. I will change it to do have this logic.

lib/init.g
sebasguts: "Should" is the right word here. It works if you do it afterwards (however that should be done), but it will not be marked as a system gvar.

fingolfin: My point is more like this: There are countless variables which "should" or "must" be system vars, and none of them has a comment like this. It feels odd to point this out for just this variable. While I understand that this was a stumbling block during development of this PR, I don't quite see what it adds -- OK, hypothetically it might stop somebody from "moving" it around, but then, why would somebody do that? OTOH, if I just want to understand what it is, then reading this adds cognitive burden.

That said, I am not going to get stuck on that -- I merely point it out.

sebasguts:
I removed the questionable sentence :).

PR #1815, opened 2017-10-26
Introduce GAP-Level StandardInput, StandardOutput, StandardError Objects

markuspf:
This introduces the `InputStream` `StandardInput` and the `OutputStream`s
`StandardOutput` and `StandardError` at the GAP level, and attaches
 * InputTextFile("*stdin*") to StandardInput,
 * OutputTextFile("*stdout*", true) to StandardOutput
 * OutputTextFile("*errout*", true) to StandardError

This small change already enables refactoring `error.g` to `PrintTo`
the `StandardError` stream, which can be replaced by an
`OutputTextString` (or a different output stream) to capture error messages.

One obvious use case of this is to capture output for alternative
REPL implementations such as Jupyter.

PR #2529, opened 2018-06-07
Make error output use the same stream consistently

ssiccha:
- PRINT_CURRENT_STATEMENT now takes a stream as additional input.
- Adjust its tests accordingly.
- Make ErrorInner, Where, and WHERE consistently pass "*errout*"
to print functions

### DIFF
--- a/lib/error.g
+++ b/lib/error.g
@@ -50,17 +50,17 @@ BIND_GLOBAL("WHERE", function( context, depth, outercontext)
     bottom := GetBottomLVars();
     lastcontext := outercontext;
     while depth > 0  and context <> bottom do
-        PRINT_CURRENT_STATEMENT(context);
-        Print(" called from\n");
+        PRINT_CURRENT_STATEMENT("*errout*", context);
+        PrintTo("*errout*", " called from\n");
         lastcontext := context;
         context := ParentLVars(context);
         depth := depth-1;
     od;
     if depth = 0 then 
-        Print("...  ");
+        PrintTo("*errout*", "...  ");
     else
         f := ContentsLVars(lastcontext).func;
-        Print("<function \"",NAME_FUNC(f)
+        PrintTo("*errout*", "<function \"",NAME_FUNC(f)
               ,"\">( <arguments> )\n called from read-eval loop ");
     fi;
 end);
@@ -75,11 +75,11 @@ BIND_GLOBAL("Where", function(arg)
     fi;
     
     if ErrorLVars = fail or ErrorLVars = GetBottomLVars() then
-        Print("not in any function ");
+        PrintTo("*errout*", "not in any function ");
     else
         WHERE(ParentLVars(ErrorLVars),depth, ErrorLVars);
     fi;
-    Print("at ",INPUT_FILENAME(),":",INPUT_LINENUMBER(),"\n");
+    PrintTo("*errout*", "at ",INPUT_FILENAME(),":",INPUT_LINENUMBER(),"\n");
 end);
 
 OnBreak := Where;
@@ -177,7 +177,7 @@ BIND_GLOBAL("ErrorInner",
         
     earlyMessage := arg[2];
     if Length(arg) <> 2 then
-        PrintTo("*errout*","ErrorInner: new format takes exactly two arguments\n");
+        PrintTo("*errout*", "ErrorInner: new format takes exactly two arguments\n");
         LEAVE_ALL_NAMESPACES();
         JUMP_TO_CATCH(1);
     fi;
@@ -186,35 +186,35 @@ BIND_GLOBAL("ErrorInner",
     ERROR_COUNT := ERROR_COUNT+1;
     errorLVars := ErrorLVars;
     ErrorLVars := context;
+    # BreakOnError is defined by the `-T` command line flag in init.g
     if QUITTING or not BreakOnError then
-        PrintTo("*errout*","Error, ");
+        PrintTo("*errout*", "Error, ");
         for x in earlyMessage do
-            PrintTo("*errout*",x);
+            PrintTo("*errout*", x);
         od;
-        PrintTo("*errout*","\n");
+        PrintTo("*errout*", "\n");
         ErrorLevel := ErrorLevel-1;
         ErrorLVars := errorLVars;
         if ErrorLevel = 0 then LEAVE_ALL_NAMESPACES(); fi;
         JUMP_TO_CATCH(0);
     fi;
-    PrintTo("*errout*","Error, ");
+    PrintTo("*errout*", "Error, ");
     for x in earlyMessage do
-        PrintTo("*errout*",x);
+        PrintTo("*errout*", x);
     od;
     if printThisStatement then 
         if context <> GetBottomLVars() then
-            PrintTo("*errout*"," in\n  \c");
-            PRINT_CURRENT_STATEMENT(context);
-            Print("\c");
-            PrintTo("*errout*"," called from \n");
+            PrintTo("*errout*", " in\n  ");
+            PRINT_CURRENT_STATEMENT("*errout*", context);
+            PrintTo("*errout*", " called from \n");
         else
-            PrintTo("*errout*","\c\n");
+            PrintTo("*errout*", "\n");
         fi;
     else
         location := CURRENT_STATEMENT_LOCATION(context);
         if location <> fail then          PrintTo("*errout*", " at ", location[1], ":", location[2]);
         fi;
-        PrintTo("*errout*"," called from\c\n");
+        PrintTo("*errout*", " called from\n");
     fi;
 
     if SHOULD_QUIT_ON_BREAK() then
@@ -225,7 +225,7 @@ BIND_GLOBAL("ErrorInner",
         OnBreak();
     fi;
     if IsString(lateMessage) then
-        PrintTo("*errout*",lateMessage,"\n");
+        PrintTo("*errout*", lateMessage,"\n");
     elif lateMessage then
         if IsBound(OnBreakMessage) and IsFunction(OnBreakMessage) then
             OnBreakMessage();
--- a/src/error.c
+++ b/src/error.c
@@ -152,22 +152,38 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
     return retlist;
 }
 
-Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj context)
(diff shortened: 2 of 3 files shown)


ssiccha:
I just realized that Where and WHERE also should take a stream ad their first argument. But OnBreak is bound to Where. I'll also have to adjust calls to OnBreak then.

ssiccha:
OnBreak is documented to have only one argument though

lib/error.g
```diff
@@ -204,7 +204,7 @@ BIND_GLOBAL("ErrorInner",
     if printThisStatement then 
         if context <> GetBottomLVars() then
             PrintTo("*errout*"," in\n  \c");
-            PRINT_CURRENT_STATEMENT(context);
+            PRINT_CURRENT_STATEMENT("*errout*", context);
             Print("\c");
```
sebasguts: Shouldn't this also be printed to `*errout*`?

ssiccha: Ofc. Thanks :)

ssiccha:
And Where is also documented. I'll let them take a stream ad an optional first argument then.

fingolfin:
I actually don't think it's that important or useful for `Where` to have a stream argument. Do you really need that for jupyter?

lib/error.g
```diff
@@ -204,8 +204,8 @@ BIND_GLOBAL("ErrorInner",
     if printThisStatement then 
         if context <> GetBottomLVars() then
             PrintTo("*errout*"," in\n  \c");
-            PRINT_CURRENT_STATEMENT(context);
-            Print("\c");
+            PRINT_CURRENT_STATEMENT("*errout*", context);
+            PrintFto("*errout*", "\c");
```
fingolfin: Typo: `PrintFto`

src/error.c
```diff
@@ -152,8 +152,16 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
 }
 
-Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj context)
+Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj stream, Obj context)
 {
+    /* HACK: we want to redirect output */
+    int openedOutput = 1;
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        Pr("Can't open output stream\n", 0L, 0L);
```
fingolfin: Can't -> Cannot

ssiccha:
@fingolfin I will introduce a global variable that tells Where, WHERE, and ErrorInner where to send all error output.
The jupyter kernel can then set that variable instead of e.g. rebinding Error

src/error.c
```diff
@@ -152,8 +152,16 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
+Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj stream, Obj context)
 {
+    /* HACK: we want to redirect output */
+    int openedOutput = 1;
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        Pr("Can not open output stream\n", 0L, 0L);
+        openedOutput = 0;
+    }
+
```
markuspf: So when you did not manage to open the output, you still print an error message to the current output and set a flag? 

Did you try just raising an error using `ErrorMayQuit`?

lib/error.g
```diff
@@ -187,34 +193,34 @@ BIND_GLOBAL("ErrorInner",
...
-        PrintTo("*errout*",x);
+        PrintTo(GAP_ERROR_STREAM, x);
     od;
     if printThisStatement then 
         if context <> GetBottomLVars() then
-            PrintTo("*errout*"," in\n  \c");
-            PRINT_CURRENT_STATEMENT(context);
-            Print("\c");
-            PrintTo("*errout*"," called from \n");
+            PrintTo(GAP_ERROR_STREAM, " in\n  \c");
```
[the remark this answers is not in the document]
ssiccha: I wanted to refactor all of ErrorInner once this PR is merged. I will remove "all" of them then.

src/error.c
ssiccha: I thought calling other Error functions may lead to an infinite loop. `ErrorMayQuit` itself also calls `ErrorInner`.

I will change it to print to `*errout*` - or quit if that also fails - to prevent that GAP could silently write to e.g. a file when encountering errors.

ssiccha: Is there a better option than quitting if I also can't open *errout*?

ssiccha: Ah, I can tell ErrorMayQuit to not go into a break loop. I'll try that.

ChrisJefferson: If you can't open errout, I would just exit. For example, the C level function PRINT_OR_APPEND_TO, which is used for printing to streams, if it tries to print to errout and fails, just exits GAP.

ssiccha: gah, the two arguments to `ErrorMayQuit` are forwarded to `Pr`. I thought they were arguments that toggle the behaviour. :/

ssiccha: Ah, thanks for the hint.

src/error.c
```diff
@@ -152,8 +152,25 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
+    /* HACK: we want to redirect output */
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        if (OpenOutput("*errout*")) {
+            Pr("PRINT_CURRENT_STATEMENT: can not open error stream\n", 0, 0 );
+        } else {
+            int ret = fputs("gap: panic, can not open *errout*!\n", stderr);
+            // If that failed, try printing to stdout
+            if(ret == EOF) {
```
fingolfin: Please run `clang-format` resp. `git clang-format` on this PR.

ssiccha:
As long as #1815 is not merged, `GAP_ERROR_STREAM` can be used to feed weird streams to `OpenOutputStream`. E.g. if the variable is set to a closed stream GAP segfaults (maybe tries to write to it, or maybe OpenOutputStream itself crashes).

ssiccha:
I just realized that one travis test-suite is failing..

ssiccha:
I fixed the bug. I forgot to do `CloseOutputStream` when returning early.

src/error.c
```diff
@@ -152,22 +152,46 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
     return retlist;
 }
 
-Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj context)
+
+Obj FuncPRINT_CURRENT_STATEMENT(Obj self, Obj stream, Obj context)
 {
-    if (context == STATE(BottomLVars))
-        return 0;
```
fingolfin: Why not leave this check here at the start, *before* trying to open the output stream?

ssiccha: That's a lot better. I'll adjust it.

ssiccha:
I also got rid of all flush characters in ErrorInner. These are unnecessary because now we always print to the same stream.

src/error.c
```diff
@@ -152,22 +152,44 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
+    /* HACK: we want to redirect output */
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        if (OpenOutput("*errout*")) {
+            Pr("PRINT_CURRENT_STATEMENT: can not open error stream\n", 0, 0);
+        }
+        else {
+            int ret = fputs("gap: panic, can not open *errout*!\n", stderr);
+            /* If that failed, try printing to stdout */
```
fingolfin: Hmm, why do this?  Is there a specific problem case you have in mind there? Why would printing to stderr fail? And if it fails,  would printing to stdout still work? And, are we really that desperate to print this message?

ssiccha: I didn't have any specific situation in mind. I thought a fallback option for when *errout* can't be openend would make sense. As to why first try stderr and then stdout: I saw similar code in `PRINT_OR_APPEND_TO` in `streams.c` and thought "better be safe than sorry".
Should I change this to just print to stderr and exit directly?

ssiccha: I changed this to be consistent with basically any other "GAP panic".

src/error.c
```diff
@@ -152,22 +152,39 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
+    /* HACK: we want to redirect output */
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        if (OpenOutput("*errout*")) {
+            Pr("PRINT_CURRENT_STATEMENT: can not open error stream\n", 0, 0);
+        }
+        else {
+            fputs("gap: panic, can not open *errout*!\n", stderr);
+            SyExit(1);
```
fingolfin: These two lines could be changed to use the new `Panic` helper function. But that's not very important.

src/error.c
```diff
@@ -152,22 +152,39 @@ Obj FuncCURRENT_STATEMENT_LOCATION(Obj self, Obj context)
...
 {
     if (context == STATE(BottomLVars))
         return 0;
 
+    /* HACK: we want to redirect output */
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    if ((IsStringConv(stream) && !OpenOutput(CSTR_STRING(stream))) ||
+        (!IS_STRING(stream) && !OpenOutputStream(stream))) {
+        if (OpenOutput("*errout*")) {
+            Pr("PRINT_CURRENT_STATEMENT: can not open error stream\n", 0, 0);
```
fingolfin: Hmm, maybe "could not" or "failed to" instead of "can not? Same in the message a few lines below?

src/error.c
ssiccha: Done.

src/error.c
ssiccha: Sounds good.

PR #2531, opened 2018-06-08
Test Error improvements

ChrisJefferson:
This makes two changes to #2519 .

Firstly, we produce a better error message, so we know where the problem is.

Secondly, add to the definition of Test files the option of having blank lines at the start of tests. This wasn't technically legal before, but it was used in various tests and was silently accepted.


lib/test.gi
```diff
@@ -22,9 +22,10 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments)
   while i <= Length(lines) do
     if i = 1 and  Length(lines[1]) > 0 and lines[1][1] = '#' then
       if ignorecomments = true then
-        # ignore comment lines at beginning of file
-        while i <= Length(lines) and Length(lines[i]) > 0 and
-              lines[i][1] = '#' do
+        # ignore comment lines and empty lines at beginning of file
+        while i <= Length(lines) and
+            ((Length(lines[i]) > 0 and lines[i][1] = '#') or lines[i] = "")
```
fingolfin: Simpler expresdion: `Length(lines[i]) = 0 or lines[i][1] = '#'`

ChrisJefferson: Fixed (also the failing test)

lib/test.gi
```diff
@@ -20,19 +20,18 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments)
   ign := [];
   i := 1;
   while i <= Length(lines) do
-    if i = 1 and  Length(lines[1]) > 0 and lines[1][1] = '#' then
+    if i = 1 and (lines[1] = "" or lines[1][1] = '#') then
```
fingolfin: Note that this line creates a temporary empty string object each time it is executed, while the old version with `Length` does not create any temporaries.

ChrisJefferson: Usually I'd assume function calls cost, but I agree that Length on strings is cheap. I don't really care about one string, but I have made the change so I can get this merged.

PR #2532, opened 2018-06-08
Refactor READ_ALL_COMMANDS

sebasguts:
EDIT: I decided to take the ReadTo functions out of this PR for now.

I changed `READ_ALL_COMMANDS` to take a function callback that is executed after each statement is evalutated. Maybe we can discuss/review/merge this first, before we write the ReadTo-Functions.

THE FOLLOWING IS OBSOLETE:

~~The two functions take a string of GAP statements, execute the statements, and return a list of results.~~
~~The PR consists of two parts:~~

~~* First part exposes the functions `OpenOutputStream` and `CloseStream` to the GAP level, and furthermore adds an output stream argument to `READ_ALL_COMMANDS`~~
~~* Second part adds the two functions, and tests for them~~

### DIFF
--- a/lib/streams.gi
+++ b/lib/streams.gi
@@ -234,6 +234,14 @@ function( stream )
     CloseStream(stream);
 end );
 
+BindGlobal("LIBGAP_EvalString",
+function(string)
+   local instream, obj;
+   instream := InputTextString(string);
+   obj := READ_ALL_COMMANDS(instream, 0, false);
+   return obj;
+end);
+
 
 #############################################################################
 ##
--- a/src/streams.c
+++ b/src/streams.c
@@ -15,6 +15,7 @@
 #include "streams.h"
 
 #include "bool.h"
+#include "calls.h"
 #include "error.h"
 #include "funcs.h"
 #include "gap.h"
@@ -23,6 +24,7 @@
 #include "lists.h"
 #include "modules.h"
 #include "io.h"
+#include "opers.h"
 #include "plist.h"
 #include "precord.h"
 #include "read.h"
@@ -79,69 +81,113 @@ static Int READ_COMMAND(Obj *evalResult)
     return 1;
 }
 
-
-/*
- * This function reads all code from the stream and returns either `fail` if the stream
- * cannot be opened, or executes all statements in the stream and returns a list of lists,
- * each entry of which reflects the result of the execution of one statement.
- *
- * The results are returned as lists with at most three bound positions.
- *
- * If the first entry is `true`, then the second entry is bound to the result of
- * the statement if there was one, and unbound otherwise, and the third entry
- * reflects whether the statement ended in a dual semicolon,
- * if the first entry is `false', an error occurred during the execution of a statement,
- * and no other positions of the list are bound.
- *
- * This function is currently used in interactive tools such as the GAP Jupyter kernel to
- * execute cells and is likely to be replaced by a function that can read a single command
- * from a stream without losing the rest of its content.
- */
-Obj FuncREAD_ALL_COMMANDS( Obj self, Obj stream, Obj echo )
+/****************************************************************************
+**
+*F  FuncREAD_ALL_COMMANDS( <self>, <instream>, <echo>, <outputFunc> )
+**
+**  FuncREAD_ALL_COMMANDS reads all code from <instream> and returns either
+**  fail if the stream cannot be opened, or executes all statements in the
+**  stream and returns a list of lists, each entry of which reflects the
+**  result of the execution of one statement.
+**
+**  The results are returned as lists of length three if <outputFunc> is
+**  false, or five, if <outputFunc> is a function.
+**
+**  If the first entry is true, then the second entry is bound to the result
+**  of the statement if there was one, and unbound otherwise, and the third
+**  entry reflects whether the statement ended in a dual semicolon. If the
+**  first entry is false, an error occurred during the execution of a
+**  statement. In this case, if <outputFunc> is false, only the first entry
+**  is set. If <outputFunc> is a function, only the first and the fifth entry
+**  are set. If <outputFunc> is a function, the fourth entry of the list is
+**  <outputFunc> applied to the result, if there was one, and if the
+**  statement did not end in a dual semicolon. Furthermore, if <outputFunc>
+**  is not false, the fifth entry contains a string of all output printed
+**  while the corresponding statement was executed. If <outputFunc> is false,
+**  all intermediate output is discarded.
+**
+**  This function is currently used in interactive tools such as the GAP
+**  Jupyter kernel to execute cells and is likely to be replaced by a
+**  function that can read a single command from a stream without losing the
+**  rest of its content.
+*/
+Obj FuncREAD_ALL_COMMANDS(Obj self, Obj instream, Obj echo, Obj outputFunc)
 {
     ExecStatus status;
-    Int resultCount, resultCapacity;
-    UInt dualSemicolon;
-    Obj result, resultList;
-    Obj evalResult;
+    UInt       dualSemicolon;
+    Obj        result, resultList;
+    Obj        evalResult;
+    Obj        outstream = 0;
+    Obj        outstreamString = 0;
 
-    /* try to open the stream */
-    if (!OpenInputStream(stream, echo == True)) {
+    /* try to open the streams */
+    if (!OpenInputStream(instream, echo == True)) {
         return Fail;
     }
 
-    resultCount = 0;
-    resultCapacity = 16;
-    resultList = NEW_PLIST( T_PLIST, resultCapacity );
-    SET_LEN_PLIST(resultList, resultCount);
+
+    if (outputFunc != False) {
+        outstreamString = NEW_STRING(0);
+        outstream = DoOperation2Args(ValGVar(GVarName("OutputTextString")),
+                                     outstreamString, True);
+    }
+    if (outstream && !OpenOutputStream(outstream)) {
+        CloseInput();
+        return Fail;
+    }
+
+    resultList = NEW_PLIST(T_PLIST, 16);
 
     do {
         ClearError();
-        status = ReadEvalCommand(STATE(BottomLVars), &evalResult, &dualSemicolon);
 
-        if(!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))) {
-            resultCount++;
-            if (resultCount > resultCapacity) {
-                resultCapacity += 16;
-                GROW_PLIST(resultList, resultCapacity);
+        if (outstream) {
+            // Clean in case Callback function had output
+            SET_LEN_STRING(outstreamString, 0);
+        }
+
+        status =
+            ReadEvalCommand(STATE(BottomLVars), &evalResult, &dualSemicolon);
+
+        if (!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))) {
+            if (outputFunc != False) {
+                Obj copy = CopyToStringRep(outstreamString);
+                SET_LEN_STRING(outstreamString, 0);
+                result = NEW_PLIST(T_PLIST, 5);
+                SET_LEN_PLIST(result, 5);
+                SET_ELM_PLIST(result, 5, copy);
+                CHANGED_BAG(result);
             }
-            result = NEW_PLIST( T_PLIST, 3 );
-            SET_LEN_PLIST(result, 1);
-            SET_ELM_PLIST(result, 1, False);
-            SET_ELM_PLIST(resultList, resultCount, result );
-            SET_LEN_PLIST(resultList, resultCount );
-
-            if(!(status & STATUS_ERROR)) {
+            else {
+                result = NEW_PLIST(T_PLIST, 3);
                 SET_LEN_PLIST(result, 3);
+            }
+            SET_ELM_PLIST(result, 1, False);
+            PushPlist(resultList, result);
+
+            if (!(status & STATUS_ERROR)) {
                 SET_ELM_PLIST(result, 1, True);
                 SET_ELM_PLIST(result, 3, dualSemicolon ? True : False);
 
                 if (evalResult) {
                     SET_ELM_PLIST(result, 2, evalResult);
+                    CHANGED_BAG(result);
+                }
+
+                if (evalResult && outputFunc != False && !dualSemicolon) {
+                    Obj tmp = CALL_1ARGS(outputFunc, evalResult);
+                    SET_ELM_PLIST(result, 4, tmp);
+                    CHANGED_BAG(result);
                 }
             }
+            else {
+                SET_LEN_PLIST(result, 1);
+            }
         }
-    } while(!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT)));
+    } while (!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT)));
+
+    if (outstream)
+        CloseOutput();
     CloseInput();
     ClearError();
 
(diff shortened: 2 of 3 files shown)


lib/streams.gi
```diff
@@ -244,6 +243,104 @@ InstallOtherMethod( ReadAsFunction,
...
+#F  ReadToString()  . . . . . . . . . . . Reads commands and returns a string
+##
+##  <#GAPDoc Label="ReadToString">
+##  <ManSection>
+##  <Func Name="ReadToString" Arg='stream[,output_stream]'/>
+##
+##  <Description>
+##    The same as <Ref Func="ReadToList"/>, except that it executes
+##    <Ref Oper="ViewObj"/> on every result and returns the resulting
+##    string.
```
fingolfin: If I hadn't been in a discussion on this last week, I don't think I'd understand this. For starters, in which order are the commands executed and ViewObj called? Interleaved, as the REPL would do? Or are first all commands executed, then all results printed? Looking at the code, it seems like the latter; note that this can lead to different results compared to the former (e.g. if printing the objects causes additional computations, such as computing the size, which then affects subsequent commands).

Also, what is "the resulting string? After all, `ViewObj` returns nothing. So perhaps this:

> This invokes ReadToList, and then calls ViewObj on every result, but with the output redirected to a string, which is then returned.

lib/streams.gi
```diff
@@ -244,6 +243,104 @@ InstallOtherMethod( ReadAsFunction,
...
+    else
+        output_stream := false;
+    fi;
+
+    return_list := ReadToList(stream,output_stream);
+
+    return_string := "";
+    return_stream := OutputTextString(return_string,true);
+
+    OPEN_OUTPUT_STREAM(return_stream);
```
fingolfin: I think the name `OPEN_OUTPUT_STREAM` is rather misleading or at least confusing -- after all, that stream already *was* opened. Perhaps `REDIRECT_OUTPUT_TO_STREAM`?

That said, I am somewhat unhappy about adding this back, after all, I just *removed* this functionality. I'd strongly prefer to hide this kernel implementation details from GAP code as much as possible. 

Moreover, I wonder if you really want this code to behave differently from the REPL, where execution and viewing are interleaved. A solution for both concerns would be to teach `READ_ALL_COMMANDS` to optionally invoke `ViewObj`, with output temporarily redirected (e.g. by adding yet another stream argument to `READ_ALL_COMMANDS`)

src/io.c
```diff
@@ -1007,6 +1007,16 @@ UInt OpenOutputStream (
     return 1;
 }
 
+Obj FuncOPEN_OUTPUT_STREAM (
+    Obj                      self,
+    Obj                      stream )
+{
+    if(OpenOutputStream(stream)){
```
fingolfin: Please use `git clang-format` on any newly added kernel code.

tst/testinstall/ReadTo.tst
```diff
@@ -0,0 +1,64 @@
+#############################################################################
+##
+#W  info.tst
```
fingolfin: This is not `info.tst`

tst/testinstall/ReadTo.tst
```diff
@@ -0,0 +1,64 @@
...
+##
+##
+#Y  Copyright (C)  2018
```
fingolfin: That line makes no sense.

src/io.c
markuspf: This function kinda re-introduces "SET_OUTPUT" that @fingolfin removed a couple of days ago. I am not sure this is a good idea?

lib/streams.gi
markuspf: Maybe we should stop mucking about with `READ_ALL_COMMANDS` and fix input streams to allow reading single commands from an input stream?

sebasguts: For now, I will do as @fingolfin suggested and implement an output stream argument to READ_ALL_COMMANDS, I think. Lets see how this turns out.

fingolfin: @markuspf sure, if you have a patch to do that, it'd be much better! But in the meantime, I think it's fairly easy to adjust `READ_ALL_COMMANDS` accordingly (or add a new kernel function doing this), so if the goal is to have something reasonable working quickly, I'd do that first, and switch to the "proper" solution once available.

That said, of course I use none of these things, so you guys can decide between yourselves how you want it :-)

src/streams.c
```diff
@@ -81,35 +82,56 @@ static Int READ_COMMAND(Obj *evalResult)
...
+Obj FuncREAD_ALL_COMMANDS(Obj self, Obj instream, Obj echo, Obj printToOutput)
 {
     ExecStatus status;
     Int resultCount, resultCapacity;
     UInt dualSemicolon;
     Obj result, resultList;
     Obj evalResult;
+    Obj        outstream;
+    Obj        outstreamString;
+    Obj        outstreamStringCopy;
```
markuspf: I think you should clang-format the whole thing to make formatting consistent.

sebasguts: Ähm. This was not done by me, but I think by clang-format (did not check its results :/).

src/streams.c
```diff
@@ -81,32 +82,57 @@ static Int READ_COMMAND(Obj *evalResult)
...
+    /* try to open the streams */
+    if (!OpenInputStream(instream, echo == True)) {
+        return Fail;
+    }
+
+
+    if (printToOutput == True) {
+        outstreamString = NEW_STRING(0);
+        outstream = DoOperation2Args(ValGVar(GVarName("OutputTextString")),
+                                     outstreamString, True);
```
fingolfin: Hmm, is there a reason to create the stream here? Why not just let the caller pass in a stream object, and also accept e.g. `fail` to indicate the old behaviour? That would simplify the logic here quite a bit. But perhaps I am missing something?

fingolfin: Oh I see, you need to create a custom string stream of course, and caputre its output after each single operation. Never mind then.

src/streams.c
```diff
@@ -125,23 +151,41 @@ Obj FuncREAD_ALL_COMMANDS( Obj self, Obj stream, Obj echo )
...
+                SET_LEN_PLIST(result, 5);
+                SET_ELM_PLIST(result, 5, outstreamStringCopy);
+            }
+            else {
+                result = NEW_PLIST(T_PLIST, 3);
+                SET_LEN_PLIST(result, 3);
+            }
             SET_ELM_PLIST(result, 1, False);
             SET_ELM_PLIST(resultList, resultCount, result );
             SET_LEN_PLIST(resultList, resultCount );
```
fingolfin: This code is missing a call to `CHANGED_BAG(resultList)` (this is an existing bug in it, not caused by this PR).

Even better would be to stop mucking around with low level calls like SET_ELM_PLIST and GROW_PLIST, and instead use PushPlist or AssPlist

src/streams.c
```diff
@@ -125,23 +151,41 @@ Obj FuncREAD_ALL_COMMANDS( Obj self, Obj stream, Obj echo )
...
                 GROW_PLIST(resultList, resultCapacity);
             }
-            result = NEW_PLIST( T_PLIST, 3 );
-            SET_LEN_PLIST(result, 1);
+            if (printToOutput == True) {
+                outstreamStringCopy = CopyToStringRep(outstreamString);
+                SET_LEN_STRING(outstreamString, 0);
+                result = NEW_PLIST(T_PLIST, 5);
+                SET_LEN_PLIST(result, 5);
+                SET_ELM_PLIST(result, 5, outstreamStringCopy);
```
fingolfin: Missing after this line: `CHANGED_BAG(result)`

src/streams.c
```diff
@@ -125,23 +151,41 @@ Obj FuncREAD_ALL_COMMANDS( Obj self, Obj stream, Obj echo )
...
 
                 if (evalResult) {
                     SET_ELM_PLIST(result, 2, evalResult);
                 }
+
+                if (evalResult && printToOutput == True && !dualSemicolon) {
+                    ViewObjHandler(evalResult);
+                    outstreamStringCopy = CopyToStringRep(outstreamString);
+                    SET_LEN_STRING(outstreamString, 0);
+                    SET_ELM_PLIST(result, 4, outstreamStringCopy);
```
fingolfin: Missing after this line: CHANGED_BAG(result)

src/streams.c
```diff
@@ -81,32 +82,57 @@ static Int READ_COMMAND(Obj *evalResult)
...
+
+
+    if (printToOutput == True) {
+        outstreamString = NEW_STRING(0);
+        outstream = DoOperation2Args(ValGVar(GVarName("OutputTextString")),
+                                     outstreamString, True);
+    }
+    else {
+        outstreamString = False;
+        outstream = False;
```
fingolfin: IMHO it would be more natural to init both of these to null, not `False`

src/streams.c
```diff
@@ -125,23 +151,41 @@ Obj FuncREAD_ALL_COMMANDS( Obj self, Obj stream, Obj echo )
...
             SET_ELM_PLIST(resultList, resultCount, result );
             SET_LEN_PLIST(resultList, resultCount );
 
             if(!(status & STATUS_ERROR)) {
-                SET_LEN_PLIST(result, 3);
                 SET_ELM_PLIST(result, 1, True);
                 SET_ELM_PLIST(result, 3, dualSemicolon ? True : False);
 
                 if (evalResult) {
                     SET_ELM_PLIST(result, 2, evalResult);
```
fingolfin: Missing after this line: CHANGED_BAG(result)

src/streams.c
```diff
@@ -2136,7 +2180,7 @@ static StructGVarFunc GVarFuncs[] = {
 
     GVAR_FUNC(READ, 1, "filename"),
     GVAR_FUNC(READ_NORECOVERY, 1, "filename"),
-    GVAR_FUNC(READ_ALL_COMMANDS, 2, "stream, echo"),
+    GVAR_FUNC(READ_ALL_COMMANDS, 3, "instream, echo, print_to_output"),
```
fingolfin: Please use camelCase not snake_case

tst/testinstall/read.tst
```diff
@@ -83,14 +83,19 @@ gap> FileString( Filename(dir, "test.g.gz"), "\037\213\b\b0,\362W\000\ctest.g\00
 32
 gap> StringFile( Filename(dir, "test.g") ) = "1+1;\n" or ARCH_IS_WINDOWS(); # works only when Cygwin installed with gzip
 true
-gap> READ_ALL_COMMANDS(InputTextString(""), false);
+gap> READ_ALL_COMMANDS(InputTextString(""), 0, false);
```
fingolfin: That 0 should be a boolean, no?

fingolfin: @sebasguts ping?

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+    }
+    if (outstream != False && !OpenOutputStream(outstream)) {
+        CloseInput();
         return Fail;
     }
 
     resultCount = 0;
     resultCapacity = 16;
-    resultList = NEW_PLIST( T_PLIST, resultCapacity );
+    resultList = NEW_PLIST(T_PLIST, resultCapacity);
```
fingolfin: Optional: I would just get rid of both `resultCount ` and esp. `resultCapacity`l the latter does nothing useful, as GROW_PLIST already grows list "intelligently", by increasing its size by at least 25%. So, just remove `resultCapacity`, hardcode `16` as second argument for `NEW_PLIST` above... then [continued below]

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+
+        if (!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))) {
             resultCount++;
             if (resultCount > resultCapacity) {
                 resultCapacity += 16;
                 GROW_PLIST(resultList, resultCapacity);
             }
-            result = NEW_PLIST( T_PLIST, 3 );
-            SET_LEN_PLIST(result, 1);
+            if (outputFunc != False) {
```
fingolfin: ... you can get rid of the if containing `GROW_PLIST` completely ...

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+            }
+            else {
+                result = NEW_PLIST(T_PLIST, 3);
+                SET_LEN_PLIST(result, 3);
+            }
             SET_ELM_PLIST(result, 1, False);
-            SET_ELM_PLIST(resultList, resultCount, result );
-            SET_LEN_PLIST(resultList, resultCount );
+            SET_ELM_PLIST(resultList, resultCount, result);
+            CHANGED_BAG(result_list);
```
fingolfin: There is no `result_list` -- edit typo?

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+            SET_LEN_PLIST(resultList, resultCount);
```
fingolfin: ... and then replace the preceding three lines by `PushPlist(resultList, result);` (which grows the list, adds the new entry, adjusts the list length and even calls `CHANGED_BAG` for you.

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+ * false, or five, if outputFunc is a function.
  *
- * If the first entry is `true`, then the second entry is bound to the result of
- * the statement if there was one, and unbound otherwise, and the third entry
- * reflects whether the statement ended in a dual semicolon,
- * if the first entry is `false', an error occurred during the execution of a statement,
- * and no other positions of the list are bound.
+ * If the first entry is `true`, then the second entry is bound to the result
+ * of the statement if there was one, and unbound otherwise, and the third
+ * entry reflects whether the statement ended in a dual semicolon, if the
```
fingolfin: Change `semicolon, if` to `semicolon. If`

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
+ * first entry is `false', an error occurred during the execution of a
+ * statement, and no other positions of the list are bound. If outputFunc
```
fingolfin: > ... and no other positions of the list are bound
This is (a) factually wrong (if `outputFunc` is given, then position 5 is always set), and (b) in the cases where it is not, it leads to a semi-broken plist being returned, one which has length 3, but indeed only the first position is set. In the latter case, I'd suggest to fix the length of the list; in the former case, the documentation or the code need to be adapted.

src/streams.c
```diff
@@ -81,67 +83,121 @@ static Int READ_COMMAND(Obj *evalResult)
...
  *
- * If the first entry is `true`, then the second entry is bound to the result of
- * the statement if there was one, and unbound otherwise, and the third entry
- * reflects whether the statement ended in a dual semicolon,
- * if the first entry is `false', an error occurred during the execution of a statement,
- * and no other positions of the list are bound.
+ * If the first entry is `true`, then the second entry is bound to the result
+ * of the statement if there was one, and unbound otherwise, and the third
+ * entry reflects whether the statement ended in a dual semicolon, if the
+ * first entry is `false', an error occurred during the execution of a
```
fingolfin: Elsewhere you use MarkDown syntax, like
```
`false`
```
but here TeX-style"
```
`false'
```

But let's not worry too much about it: This documentation comment completely violates the "standard" we use everywhere else in the kernel, so I'd be tempted to rewrite it in a follow up commit anyway. The content of this comment is much more important!

lib/streams.gi
```diff
@@ -234,6 +234,31 @@ function( stream )
     CloseStream(stream);
 end );
 
+BindGlobal("EvaluateString",
```
fingolfin: Not a change request per se, just rambling: I am somewhat concerned about having an undocumented function `EvaluateString` next to the documented `EvalString`, as I am pretty sure people will discover the undocumented one quickly from tab completion, and there will be confusion about this.

For `LIBGAP_EvalString`, the name makes it clear what it is for. Not so for, which has zero documentation, and also is not mentioned in the commit message nor the description of this PR... So what goes?

src/streams.c
```diff
@@ -79,69 +81,125 @@ static Int READ_COMMAND(Obj *evalResult)
...
+    }
+    if (outstream != False && !OpenOutputStream(outstream)) {
+        CloseInput();
         return Fail;
     }
 
     resultCount = 0;
-    resultCapacity = 16;
-    resultList = NEW_PLIST( T_PLIST, resultCapacity );
+    resultList = NEW_PLIST(T_PLIST, 16);
```
fingolfin: I took the liberty of fixing the formatting of your comment for `FuncREAD_ALL_COMMANDS`.

But now I see that you dropped most if not all of the other changes I had added previously to your PR (via pull request that you merged). Is there a reason for that?

ChrisJefferson:
I've lost track of this PR, @sebasguts , do you think you have dealt with all of @fingolfin 's comments?

PR #2533, opened 2018-06-08
Don't use fast method for composing with identity mapping if source & range don't match

stevelinton:
The code that was there could produce a composition whose range didn't match the second argument, which can never be right.

### DIFF
--- a/tst/testbugfix/2018-06-11-mapping.tst
+++ b/tst/testbugfix/2018-06-11-mapping.tst
@@ -1,0 +1,13 @@
+# There was a bug where composition of an identity mapping and another
+# mapping x where the Source of the identity mapping contained the
+# ImagesSource of x, but not its whole range, simply returned x,
+# resulting in a composition whose Range was strictly bigger than that
+# of its first argument, and causing problems in IsomorphismPermGroup
+#
+gap> g := SymmetricGroup(7);;
+gap> phi := GroupGeneralMappingByImages(g,g,[(1,2,3,4,5),(1,2)],[(1,2,3,4,6),(1,2)]);
+[ (1,2,3,4,5), (1,2) ] -> [ (1,2,3,4,6), (1,2) ]
+gap> psi := IdentityMapping(SymmetricGroup(6));
+IdentityMapping( Sym( [ 1 .. 6 ] ) )
+gap> Range(CompositionMapping(psi,phi));
+Sym( [ 1 .. 6 ] )
--- a/lib/mapprep.gi
+++ b/lib/mapprep.gi
@@ -1387,9 +1387,7 @@ InstallMethod( CompositionMapping2,
   [ IsGeneralMapping and IsOne, IsGeneralMapping ],
   SUM_FLAGS + 1,  # should be higher than the rank for a zero mapping
 function( id, map )
-  if    not IsSubset(Source(id),Range(map))
-    and not IsSubset(Source(id),ImagesSource(map))
-  then
+  if not IsSubset(Source(id),Range(map)) then
     # if the identity is defined on something smaller, we need to take a
     # true `CompositionMapping'.
     TryNextMethod();


stevelinton:
Needs "Squash and merge" when the time comes.

stevelinton:
Probably also needs a test that works in master. I discovered this with #2521. Adding "Do Not Merge" for now.

tst/testbugfix/2018-06-11-mapping.tst
```diff
@@ -0,0 +1,13 @@
+#
+# There was a bug where composition of an identify mapping and another mapping x where the Source of the identity mapping contained the
```
fingolfin: identify -> identity

Also, could you please wrap your lines after 80 or so characters?

tst/testbugfix/2018-06-11-mapping.tst
```diff
@@ -0,0 +1,13 @@
...
+# ImagesSource of x, but not its whole range, simply returned x, resulting in a compoition whose Range was strictly bigger than that
+# of its first argument, and causign problems in IsomorphismPermGroup
```
fingolfin: causign -> causing

tst/testbugfix/2018-06-11-mapping.tst
```diff
@@ -0,0 +1,13 @@
+#
+# There was a bug where composition of an identify mapping and another mapping x where the Source of the identity mapping contained the
+# ImagesSource of x, but not its whole range, simply returned x, resulting in a compoition whose Range was strictly bigger than that
```
fingolfin: compoition -> composition

tst/testbugfix/2018-06-11-mapping.tst
```diff
@@ -0,0 +1,13 @@
...
+# of its first argument, and causign problems in IsomorphismPermGroup
+#
+#
+gap> g := SymmetricGroup(7);;
+gap> phi := GroupGeneralMappingByImages(g,g,[(1,2,3,4,5),(1,2)],[(1,2,3,4,6),(1,2)]);
+[ (1,2,3,4,5), (1,2) ] -> [ (1,2,3,4,6), (1,2) ]
+gap> psi := IdentityMapping(SymmetricGroup(6));
+IdentityMapping( Sym( [ 1 .. 6 ] ) )
+gap> Range(CompositionMapping(psi,phi));
+Sym( [ 1 .. 6 ] )
```
fingolfin: Hmm, phi has Sym(7) as its range (but not as its image!), so on a purely formal level, it shouldn't compose with psi like that, should it?

But OK, GAP doesn't seem to specify what it allows for composition, so I'll assume that only the image is relevant for the legality of the composition, not the range. I guess being stricter about this might be annoying in some situations, while the added "type safety" it provides may not be worth it. (Also, changing existing behavior like that of course is risky)

stevelinton: @fingolfin -- the manual is rather silent on what is allowed for the sources and ranges in a composition, whether there is an implicit restriction, and so on. I'll flag an issue to document and test some set of rules, but I don't want it to get tangled in this, which seems to be a safe bugfix.

fingolfin: Thanks, that's fine by me. So I'll just wait for the typos in the comments above to be fixed.

stevelinton:
@hulpke The dual code (which is just a few lines further up the same file) doesn't have the same problem, which is in and of itself, curious. The "next method" in general, produces an object in `IsCompositionMappingRep`. One could, I suppose have (in at least one of the two orders) a method that delegates to `RestrictedMapping` (is there an obvious reason why there isn't a `RangeRestrictedMapping` which would handle the other case? I don't know if it would make any real difference. We really need to sort out the more general questions in #2540 first.

PR #2535, opened 2018-06-08
Add tests for ReadCSV and PrintCSV

alex-konovalov:
This should increase coverage of `lib/string.gi`.


alex-konovalov:
@hulpke IIRC the tested functionality was added by you - could you review this PR, please? In particular, `RCSVSplitString` is not documented, and still is not exercised by the new tests. What's it purpose (I could add a comment to the source code), and how to construct an example to use it?

hulpke:
@alex-konovalov 
RCSVSplitString is a variant of `SplitString` but deals with quotes for string entries (removing them) at the same time. It is used when reading in spreadsheets while obeying the column order. If you add

    gap> s:=ReadCSV(fname,true);
    [ rec( field1 := "f", field2 := "f1", field3 := "f 2" ),
      rec( field1 := 2, field2 := 1, field3 := 3 ), rec( field2 := 4, field3 := 6 ),
      rec( field2 := 7 ), rec( field1 := 11, field3 := 12 ),
      rec( field1 := "yy", field2 := "x", field3 := "zzz" ) ]

to the test, it will be exercised as well.

PR #2536, opened 2018-06-10
kernel: don't include debug.h everywhere; and more

fingolfin:
UPDATE: the GMP related changes cause NormalizInterface to fail. I'll revisit them later.


src/scanner.c
```diff
@@ -37,11 +37,12 @@ static UInt NextSymbol(void);
 */
 static void SyntaxErrorOrWarning(const Char * msg, UInt error)
 {
-    // open error output
-    OpenOutput("*errout*");
-
     // do not print a message if we found one already on the current line
     if (STATE(NrErrLine) == 0) {
+
```
ChrisJefferson: Why do this?

Also, other people are (as I understand it) working on streams/error things at the moment. I worry this might conflict with them, particularly if it isn't actually doing anything useful?

fingolfin: To quote the commit message: "kernel: avoid unnecessarily opening output stream"

There are three error related open PRs: #2529, #2530, #2531. None of them modifies this file. Based on the discussion last week, I am also not aware of plans by @markuspf or @ssiccha to change this function, hopefully they'll correct me on that if necessary. I *am* aware of more planned changes related to error handling, namely by me, see PR #2520, but there is ni conflict there either...

PR #2537, opened 2018-06-10
Further improve test coverage for ReadCSV

alex-konovalov:
I was going to edit #2535 to add one more test as @hulpke suggested in #2535, but it was merged in the meantime by @fingolfin, so I am doing this in a separate PR.

### DIFF
--- a/tst/testinstall/strings.tst
+++ b/tst/testinstall/strings.tst
@@ -262,6 +262,15 @@ gap> fname := Filename(tmpdir, "output.csv");;
 gap> PrintCSV( fname, s );
 gap> s = ReadCSV( fname );
 true
+gap> s[2].emil:="\"Alas, poor Yorick\", the call went";;
+gap> s[3].empty:="";;
+gap> PrintCSV(fname,s);
+gap> ReadCSV(fname,true);
+[ rec( field1 := "emil", field2 := "empty", field3 := "f", field4 := "f1", 
+      field5 := "f 2" ), rec( field3 := 2, field4 := 1, field5 := 3 ), 
+  rec( field1 := "\"Alas, poor Yorick\", the call went", field4 := 4, 
+      field5 := 6 ), rec( field4 := 7 ), rec( field3 := 11, field5 := 12 ), 
+  rec( field3 := "yy", field4 := "x", field5 := "zzz" ) ]
 
 # Long Strings
 # All these strings are 2000 or 3000 characters long, to make sure we fill the string buffer at least once or twice


tst/testinstall/strings.tst
```diff
@@ -262,6 +262,12 @@ gap> fname := Filename(tmpdir, "output.csv");;
 gap> PrintCSV( fname, s );
 gap> s = ReadCSV( fname );
 true
+gap> s := ReadCSV( fname, true );
```
fingolfin: This new test actually does not use RCSVSplitString. To do that, the input must contain quotes " and use commas as sepetators. The quotes then are used to mask fields containing commas.

hulpke: @fingolfin 
You're right. We need to force commas in strings, and maybe lets add some gratuitous quotes. But it does not need to be the mask.

What about

    gap> s[2].emil:="\"Alas, poor Yorick\", the call went";;
    gap> PrintCSV(fnam,s);
    gap> ReadCSV(fnam,true);
    [ rec( field1 := "emil", field2 := "f", field3 := "f1", field4 := "f 2" ),
      rec( field2 := 2, field3 := 1, field4 := 3 ),
      rec( field1 := "\"Alas, poor Yorick\", the call went", field3 := 4, field4 := 6 ),
      rec( field3 := 7 ),  rec( field2 := 11, field4 := 12 ),
      rec( field2 := "yy", field3 := "x", field4 := "zzz" ) ]

alex-konovalov:
Pushed "poor Yorick" example - let's see...

alex-konovalov:
Ok, now https://codecov.io/gh/gap-system/gap/pull/2537/changes shows 6.3% increase in `string.gi` coverage, thanks @hulpke for the tip!

tst/testinstall/strings.tst
```diff
@@ -262,6 +262,14 @@ gap> fname := Filename(tmpdir, "output.csv");;
...
 gap> s = ReadCSV( fname );
 true
+gap> s[2].emil:="\"Alas, poor Yorick\", the call went";;
+gap> PrintCSV(fname,s);
+gap> ReadCSV(fname,true);
+[ rec( field1 := "emil", field2 := "f", field3 := "f1", field4 := "f 2" ), 
+  rec( field2 := 2, field3 := 1, field4 := 3 ), 
+  rec( field1 := "\"Alas, poor Yorick\", the call went", field3 := 4, 
+      field4 := 6 ), rec( field3 := 7 ), rec( field2 := 11, field4 := 12 ), 
+  rec( field2 := "yy", field3 := "x", field4 := "zzz" ) ]
```
fingolfin: Can you please also add an "empty" field? I.e. containing an empty string.

alex-konovalov: Done

PR #2539, opened 2018-06-11
kernel: remove redundant SET_LEN_PLIST(l,0) calls

fingolfin:
This used to be part of PR #2523, but to ease reviewing, I now split it into a separate PR.

These calls to `SET_LEN_PLIST(l,0)` are redundant, because the result of `NEW_PLIST` is guaranteed to already have its length set to 0.


sebasguts:
Looks good to me.

PR #2542, opened 2018-06-11
Add DEBUG_MASTERPOINTERS to --enable-debug.

ChrisJefferson:
This PR adds some existing gasman checks to --enable-debug.

In the process, I found most of the checks were failing. The failures in header->reserved were harmless (we weren't setting it to 0 in some place). The bad issue could, in some circumstances, cause a very hard to reproduce crash (I believe, unless there is some other bug hiding). Either way, both issues should get cleaned up.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -914,6 +914,10 @@ Bag NextBagRestoring( UInt type, UInt flags, UInt size )
   header->flags = flags;
   header->size = size;
   header->link = NextMptrRestoring;
+#if SIZEOF_VOID_P == 4
+  header->reserved = 0;
+#endif
+
   NextMptrRestoring++;
 
   if ((Bag *)NextMptrRestoring >= MptrEndBags)
@@ -1267,6 +1271,9 @@ Bag NewBag (
     header->type = type;
     header->flags = 0;
     header->size = size;
+#if SIZEOF_VOID_P == 4
+    header->reserved = 0;
+#endif
 
     /* enter link word                                                     */
     header->link = bag;
@@ -1411,6 +1418,9 @@ UInt ResizeBag (
     UInt type     = header->type;
     UInt flags    = header->flags;
     UInt old_size = header->size;
+#if SIZEOF_VOID_P == 4
+    GAP_ASSERT(header->reserved == 0);
+#endif
 
 #ifdef COUNT_BAGS
     /* update the statistics                                               */
@@ -1446,6 +1456,9 @@ UInt ResizeBag (
         }
 
         header->size = new_size;
+#if SIZEOF_VOID_P == 4
+        GAP_ASSERT(header->reserved == 0);
+#endif
     }
 
     // if the last bag is enlarged ...
@@ -1474,6 +1487,9 @@ UInt ResizeBag (
         ADD_CANARY();
 
         header->size = new_size;
+#if SIZEOF_VOID_P == 4
+        GAP_ASSERT(header->reserved == 0);
+#endif
     }
 
     // if the bag is enlarged ...
@@ -1493,6 +1509,9 @@ UInt ResizeBag (
         header->type = 255;
         header->flags = 0;
         header->size = sizeof(BagHeader) + (WORDS_BAG(old_size) - 1) * sizeof(Bag);
+#if SIZEOF_VOID_P == 4
+        GAP_ASSERT(header->reserved == 0);
+#endif
 
         /* allocate the storage for the bag                                */
         BagHeader * newHeader = (BagHeader *)AllocBags;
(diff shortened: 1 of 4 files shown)


src/gasman.c
```diff
@@ -915,6 +915,10 @@ Bag NextBagRestoring( UInt type, UInt flags, UInt size )
   header->flags = flags;
   header->size = size;
   header->link = NextMptrRestoring;
+#if SIZEOF_VOID_P == 4
```
stevelinton: Are we obsoleting SYS_IS_64_BIT in favour of checking SIZEOF_VOID_P directly? No objection to doing so, but we
should probably do it globally if so

ChrisJefferson: I was continuing to use SIZEOF_VOID_P, which was already being used to check this reserved value.

ChrisJefferson:
Woops, I added two extra commits to this PR. However, I hope they are uncontroversial small fixes to crashes from holding pointers over GCs.

src/gasman.c
fingolfin: I guess it is my "fault" that `gasman.h` uses `SIZEOF_VOID_P` and not `SYS_IS_64_BIT`; but in my defense (if that's necessary :-), that was because the "old" code, before I introduced `BagHeader`, also checked `SIZEOF_VOID_P` to set `USE_NEWSHAPE` -- perhaps it was added before `SYS_IS_64_BIT` existed?

Anyway, I have a slightly preference to use `SYS_IS_64_BIT`, *but* I am fine with this part of the PR either way; we can change `SIZEOF_VOID_P` checks to use `SYS_IS_64_BIT` later on if we want to -- or also not do it, I think the code is fine either way (if we ever get 128 bit systems, we'll have to go through it all anyway ;-).

src/gasman.c
```diff
@@ -2304,6 +2339,10 @@ void CheckMasterPointers( void )
         if (!IS_BAG_BODY(*ptr))
             SyAbortBags("Bad master pointer detected");
 
+        if(GET_MARK_BITS(LINK_BAG(bag))) {
```
fingolfin: Code formatting

src/trans.c
```diff
@@ -3151,10 +3151,11 @@ Obj FuncCYCLES_TRANS(Obj self, Obj f)
         return out;
     }
 
-    seen = ResizeInitTmpTrans(deg);
     out = NEW_PLIST(T_PLIST, 0);
     nr = 0;
 
+    seen = ResizeInitTmpTrans(deg);
+
```
fingolfin: Argh, damn, we went through that code so often, looking for exactly these kinds of problems, and still missed this one ... :-(

src/gasman.c
```diff
@@ -1955,7 +1955,16 @@ UInt CollectBags (
         // extract the head from the linked list
         first = MarkedBags;
         MarkedBags = LINK_BAG(first);
-        LINK_BAG(first) = MARKED_ALIVE(first);
+        // gasman is inconsistent if a bag starting at YoungBags is
```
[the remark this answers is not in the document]
fingolfin: Ah, maybe what is meant here is that in some places, Gasman will treat a bag with `PTR_BAG(bag) == YoungBags` as young, and in other as old? Shouldn't we then just fix it one way or another? (Maybe that's too big a change for this PR, I didn't look at the code, just wondering)

fingolfin:
Also, please set appropriate labels.

src/gasman.c
```diff
@@ -1955,7 +1955,16 @@ UInt CollectBags (
...
+        // a young bag. However at present such bags cannot exist.
+        // Add a check to ensure this property never changes accidentally.
```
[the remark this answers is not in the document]
ChrisJefferson: When we construct a bag, we always allocate a header just before the bag data, so the first bag after a GC is a little further along. This could break if anyone started rearranging bag internal structure.

This is a weird bit of code. I realised we were inconsistent with how we handled a bag at YoungBags, worried about it, then realised it can't actually happen. This check is mainly here so if anyone makes bags which can be at YoungBags, they'll trigger this and think about it :)

src/gasman.c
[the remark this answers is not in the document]
ChrisJefferson: My problem was you can't actually really check you are right, as that condition is never true :)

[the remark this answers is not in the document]
ChrisJefferson: I rephrased this comment, and also moved it to a seperate check.

Basically, I spend a while worrying about PTR_BAG(bag) == YoungBags, then decided I'm sure it can't ever happen. However, just for sanity's sake I would like to add a check, just to make sure my reasoning was right (if this assert ever triggers, we'll have to find out why, and then check bags starting at YoungBags carefully).

fingolfin:
Oops, missed on `SyAbortBags` -- the fix for the fix should now work, though.

PR #2543, opened 2018-06-12
Start to refactor module state initialization

fingolfin:
This is the first part of PR #2512; the changes in here should be fairly obvious (at least if read commit-by-commit), and have no harmful effect.

Another semi-required part for PR #2512 is PR #2507, simply because it touches some overlapping code; but we can also postpone PR #2507, though then I'll have to rewrite parts of all involved PRs later on.

After this PR is merged, and the fate of PR #2507 is decided, I can rebase PR #2512, and it'll only contain the "difficult" bits of the module state changes (namely, those affecting in which order we call InitLibrary vs. InitModuleState

### DIFF
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -2277,10 +2277,13 @@ static StructInitInfo module = {
     .name = "cyclotom",
     .initKernel = InitKernel,
     .initLibrary = InitLibrary,
+
+    .moduleStateSize = sizeof(struct CycModuleState),
+    .moduleStateOffsetPtr = &CycStateOffset,
+    .initModuleState = InitModuleState,
 };
 
 StructInitInfo * InitInfoCyc ( void )
 {
-    CycStateOffset = RegisterModuleState(sizeof(struct CycModuleState), InitModuleState, 0);
     return &module;
 }
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -2255,11 +2255,14 @@ static Int InitLibrary (
 }
 
 
-static void InitModuleState(ModuleStateOffset offset)
+static Int InitModuleState(void)
 {
     ResultCyc = 0;
     LastECyc = 0;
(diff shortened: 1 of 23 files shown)


src/cyclotom.c
```diff
@@ -143,7 +143,7 @@ static inline void SET_NOF_CYC(Obj cyc, Obj val)
 #define XXX_CYC(cyc,len)        (EXPOS_CYC(cyc,len)[0])
 
 
-static ModuleStateOffset CycStateOffset = -1;
+static UInt CycStateOffset;
```
fingolfin: I should point out that changing these offsets to UInt looses the extra sanity check of initializing them to -1, which we used to detect modules that forgot to init the offset. But:
1. we could just get that back by setting to `~(UInt)0` or just any value larger than the maximal allowed offset (using some `MAX_UINT_FOOBAR` value from the `limits.h` or so);
2. we could add 1 to all offsets making 0 invalid (and subtracting 1 again when using the offset);
3. we could also just not care (as this PR does), thinking that now that we register the module state and initialize the offsets in a uniform way, it's far less likely that this gets "forgotten", and anyway, we don't worry about this for countless other things in the kernel...

ChrisJefferson: Stupid question, why not make it an Int, then you can keep -1? These offsets are never going to be need an unsigned type.

This might be a place where we disagree, I always want to make things an Int unless they have to be unsigned, whereas I know some people like UInt for things which should be unsigned.

ChrisJefferson:
I am happy with this, I made a comment but it doesn't need changing.

fingolfin:
@ChrisJefferson I dropped the commit which removed `ModuleStateOffset`, so in particular, the offset stays as an `Int`, as you preferred.  Hopefully the tests will pass once again and this can be merged.

If you could also look at https://github.com/gap-system/gap/pull/2507 that'd be awesome.

PR #2551, opened 2018-06-14
New command line option --alwaystrace

ssiccha:
Also introduces a new global variable `AlwaysPrintTracebackOnError`.

Funtionality and documentation yet to be implemented.

### DIFF
--- a/lib/error.g
+++ b/lib/error.g
@@ -185,7 +185,7 @@ BIND_GLOBAL("ErrorInner",
         LEAVE_ALL_NAMESPACES();
         JUMP_TO_CATCH(1);
     fi;
-        
+
     # Local functions that print the user feedback.
     printEarlyMessage := function(stream)
         PrintTo(stream, "Error, ");
@@ -222,12 +222,21 @@ BIND_GLOBAL("ErrorInner",
     if QUITTING or not BreakOnError then
         # If we skip the break loop, the standard behaviour is to print only
         # the earlyMessage. If SilentNonInteractiveErrors is true we do not
-        # print any messages.
-        # This is used by HPC-GAP to e.g. suppress error messages in worker
+        # print any messages. If AlwaysPrintTracebackOnError is true we also
+        # call OnBreak(), which by default prints the traceback.
+        # SilentNonInteractiveErrors superseeds AlwaysPrintTracebackOnError.
+        # It is used by HPC-GAP to e.g. suppress error messages in worker
         # threads.
         if not SilentNonInteractiveErrors then
             printEarlyMessage("*errout*");
-            PrintTo("*errout*", "\n");
+            if AlwaysPrintTracebackOnError then
+                printEarlyTraceback("*errout*");
+                if IsBound(OnBreak) and IsFunction(OnBreak) then
+                    OnBreak();
+                fi;
+            else
+                PrintTo("*errout*", "\n");
+            fi;
         fi;
         if IsHPCGAP then
             # In HPC-GAP we want to access error messages encountered in
@@ -236,6 +245,17 @@ BIND_GLOBAL("ErrorInner",
             LastErrorMessage := "";
             lastErrorStream := OutputTextString(LastErrorMessage, true);
             printEarlyMessage(lastErrorStream);
+            if AlwaysPrintTracebackOnError then
+                printEarlyTraceback(lastErrorStream);
+                # FIXME: Also make HPCGAP work with OnBreak().
+                # If AlwaysPrintTracebackOnError is true, the output of
+                # OnBreak() should also be put into LastErrorMessage.
+                # To do this there needs to be a way to put its output
+                # into lastErrorStream.
+                # OnBreak() is documented to not take any arguments.
+                # One could work around that if there were e.g. a GAP error
+                # stream which all error functions print to.
+            fi;
             CloseStream(lastErrorStream);
             MakeImmutable(LastErrorMessage);
         fi;
@@ -249,6 +269,14 @@ BIND_GLOBAL("ErrorInner",
     printEarlyTraceback("*errout*");
 
     if SHOULD_QUIT_ON_BREAK() then
+        # Again, the default is to not print the rest of the traceback.
+        # If AlwaysPrintTracebackOnError is true we do so anyways.
+        if
+            AlwaysPrintTracebackOnError
+            and IsBound(OnBreak) and IsFunction(OnBreak)
+        then
+            OnBreak();
+        fi;
         FORCE_QUIT_GAP(1);
     fi;
 
--- a/lib/error.g
+++ b/lib/error.g
@@ -117,7 +117,7 @@ BIND_GLOBAL("ErrorInner",
             printEarlyMessage, printEarlyTraceback, lastErrorStream,
             shellOut, shellIn;
 
-	context := arg[1].context;
+    context := arg[1].context;
     if not IsLVarsBag(context) then
         PrintTo("*errout*", "ErrorInner:   option context must be a local variables bag\n");
         LEAVE_ALL_NAMESPACES();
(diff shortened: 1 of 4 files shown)


ssiccha:
I improved the naming (I hope). The new option is now called `--alwaystrace`. `--nobreakloop` now is the long version of `-T`.

ChrisJefferson:
Out of interest, do you want to merge this then implement the functionality (which would be a little unusual), or carry on working on the function in this PR?

ssiccha:
I want to finish this over the next few days. :)

ssiccha:
I added a new commit which implements the new behaviour. I'll have to think about whether this is correct tomorrow again.

Still missing: documentation of the new global variable `AlwaysPrintTracebackOnError`

ssiccha:
Let's see whether tests pass on travis. :)

lib/error.g
```diff
@@ -189,21 +189,50 @@ BIND_GLOBAL("ErrorInner",
...
-            PrintTo("*errout*", "Error, ");
-            for x in earlyMessage do
-                PrintTo("*errout*", x);
-            od;
-            PrintTo("*errout*", "\n");
+    # The variables BreakOnError and QUITTING determine whether we
+    # start an interactive break loop or whether we use JUMP_TO_CATCH
+    # to exit this function.
+    # BreakOnError is defined by the `-T` command line flag in init.g.
+    doBreakLoop := BreakOnError and not QUITTING;
```
markuspf: `BreakOnError` is initialised depending on whether (and how many times?) `-T` is given on the command line, the user can (unfortunately) set it as well.

lib/error.g
```diff
@@ -189,21 +189,50 @@ BIND_GLOBAL("ErrorInner",
...
+    # If we do not start a break loop, the variables SilentErrors and
+    # AlwaysPrintTracebackOnError determine whether any messages or
+    # respectively tracebacks are printed.
```
markuspf: How?

lib/error.g
```diff
@@ -189,21 +189,50 @@ BIND_GLOBAL("ErrorInner",
...
+    fi;
+    # Do we want to skip the break loop?
+    if not doBreakLoop then
         if IsHPCGAP then
             LastErrorMessage := "";
             lastErrorStream := OutputTextString(LastErrorMessage, true);
             for x in earlyMessage do
                 PrintTo(lastErrorStream, x);
             od;
+            # FIXME: Also make HPCGAP work with AlwaysPrintTracebackOnError
```
markuspf: What needs to be done to achieve this? I see some code added for HPC-GAP, is this comment obsolete?

lib/init.g
```diff
@@ -248,10 +250,15 @@ CallAndInstallPostRestore( function()
 
     if IsHPCGAP then
       BindThreadLocal( "BreakOnError", not GAPInfo.CommandLineOptions.T );
```
markuspf: Not really relevant to this PR: We should comment here why these variables are thread local (I suppose we don't want a break look in a worker thread; I have had students trying to debug task-based code, which was a big pain becuase of no break loops or error messages.

lib/error.g
fingolfin: I think it is quite fortunate that the user can do this, it has been crucial for some code I wrote, esp. for testing code.

lib/error.g
```diff
@@ -212,33 +241,11 @@ BIND_GLOBAL("ErrorInner",
...
-        PrintTo("*errout*", " called from\n");
-    fi;
-
     if SHOULD_QUIT_ON_BREAK() then
         FORCE_QUIT_GAP(1);
     fi;
 
-    if IsBound(OnBreak) and IsFunction(OnBreak) then
-        OnBreak();
-    fi;
```
fingolfin: Huh, so `OnBreak` is now called *before*` the test for `SHOULD_QUIT_ON_BREAK`? That's quite a change in semantics, one which I am quite concerned about. I'll need to look at your whole PR with a bit more time to understand what is going on.
(It might help if the refactoring did not happen at the same time as the changes in behavior -- note that to me, "refactoring" means a change which does *not* change behavior (or at most changes in a very minimalistic way, which then however should be documented in the commit message)

fingolfin:
In the commit message, you write:
> This commit also refactors ErrorInner a bit. Otherwise the code duplication would have been unbearable.

I can't judge if this is really code refactoring or not, as it is mixed with actual code changes, some of which match what you describe, but some I don't see reflected in your descriptions, so I wonder if they are intentational. If the refactoring was done in a separate commit, this would be far easier to evaluate.

As to avoiding code duplication: Another technique in this case that might work is to put the code being duplicatated into an inner function (which thus has access to all local vars), and simply call that from all the places that need to call it. That's a change which is fairly easy to review, too.

lib/error.g
ssiccha: Good point. I'll change it to `initialize`.

lib/error.g
ssiccha: I'll expand on that. :+1:

lib/error.g
ssiccha: This isn't obsolete. IMO if `AlwaysPrintTracebackOnError` is true, the traceback should also be put into `LastErrorMessage`. To do this there needs to be a way to put the output of `Where()` into `lastErrorStream`.

I'll add a comment to the source.

lib/error.g
ssiccha: You're totally right. I'll make separate commits out of that. Note to myself: never mix refactoring and behaviour changes in a single commit.

I realize that I forgot to check for `SHOULD_QUIT_ON_BREAK()` in the if-clause that controls whether `OnBreak()` is called. That change in semantics was not on purpose.

lib/error.g
markuspf: The unfortunate bit is that the user can set this to, say, a function which makes GAP crash when an error is hit.

lib/init.g
ssiccha: Did that too.

lib/error.g
```diff
@@ -249,6 +269,14 @@ BIND_GLOBAL("ErrorInner",
     printEarlyTraceback("*errout*", context, printThisStatement);
 
     if SHOULD_QUIT_ON_BREAK() then
+        # Again, the default is to not print the rest of the traceback.
+        # If AlwaysPrintTracebackOnError is true we do so anyways.
+        if
+            AlwaysPrintTracebackOnError
```
fingolfin: This is a weird line break, please merge the two preceding lines into one.

lib/error.g
```diff
@@ -184,26 +185,77 @@ BIND_GLOBAL("ErrorInner",
         LEAVE_ALL_NAMESPACES();
         JUMP_TO_CATCH(1);
     fi;
-        
+
+    # Local functions that print the user feedback.
+    printEarlyMessage := function(stream, earlyMessage)
```
fingolfin: You don't need to pass `earlyMessage` as an argument, it is available automatically in this closure. I.e. just drop the second argument here and in calls, no other change, and things will work just fine.

lib/init.g
```diff
@@ -247,12 +249,22 @@ CallAndInstallPostRestore( function()
...
+      );
+      BindThreadLocal( "SilentNonInteractiveErrors", false );
+      # We store the error messages on a per thread basis to be able to
+      # e.g. access them independently from the main thread.
       BindThreadLocal( "LastErrorMessage", "" );
     else
       ASS_GVAR( "BreakOnError", not GAPInfo.CommandLineOptions.T );
-      ASS_GVAR( "SilentErrors", false );
+      ASS_GVAR( "AlwaysPrintTracebackOnError", GAPInfo.CommandLineOptions.alwaystrace );
+      ASS_GVAR( "SilentNonInteractiveErrors", false );
```
fingolfin: Note a change request, just a remark: if the renaming of `SilentErrors` to `SilentNonInteractiveErrors` were in a separate commit, that'd make things even easier to review. But it's fine to keep it now as it is.

lib/error.g
```diff
@@ -185,25 +186,56 @@ BIND_GLOBAL("ErrorInner",
...
         if IsHPCGAP then
+            # In HPC-GAP we want to access error messages encountered in
+            # tasks via TaskError. To this end we store the error message
+            # in the thread local variable LastErrorMessage.
             LastErrorMessage := "";
             lastErrorStream := OutputTextString(LastErrorMessage, true);
-            for x in earlyMessage do
-                PrintTo(lastErrorStream, x);
-            od;
+            printEarlyMessage(lastErrorStream, earlyMessage);
```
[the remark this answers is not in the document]
fingolfin: I dug a bit, and as far as I can tell, this only affects the `TaskError()` function (and of course the `LastErrorMessage` global). I'd still prefer if the redundant arguments were dropped, as they are rather "un-GAPish", but if this is merged with it, I can also fix it myself *shrug*

ssiccha:
I've rebased the commits. I squashed `Remove redundant arguments to inner functions` but left the reformatting in its own commit.

PR #2552, opened 2018-06-14
Fix typo in checking if a filename is *stderr*

ChrisJefferson:
This fixes a typo in some code I previously wrote, and adds tests to make sure it works now.


ChrisJefferson:
woops, didn't check if it effected 4.9 (it doesn't)

PR #2560, opened 2018-06-18
kernel: fix crash in GASMAN("global")

fingolfin:


fingolfin:
I marked this as "not for release notes", as this is in an undocumented function. But if anybody disagrees, I don't mind adding this to the release note (the crash is a regression in GAP 4.9)

PR #2564, opened 2018-06-20
kernel: merge gapstate.c into modules.c; allow kernel extensions to have module state

### LINKED ISSUE #2013: ModuleState is not properly initialized in kernel extensions
Right now, module constructors are only run once, during `InitializeGap`. Thus if a module is loaded late on, as part of a kernel extension in a package, its module state constructor is *not* run. Contrast this to its initKernel and initLibrary functions, were great care is taken to also run them for module loaded like this.

In single threaded GAP, a package can easily work around this, as you can simply not use a module constructor and instead put everything into your `InitLibrary` function. But for HPC-GAP, this means that in all already running states, a packages module state would not be initialized.

Of course no packages uses module state yet, simply because no package kernel extension takes advantage of HPC-GAP yet (or even bothers to try to be compatible). But in order to not prevent them from doing so, we'd have to solve this issue.

The obvious fix is to add the relevant calls to the constructor (if any) to `READ_GAP_ROOT`, `FuncLOAD_DYN` and `FuncLOAD_STAT`. (And while we are it, it seems these functions do not  bother to check whether `info->initLibrary != 0`, which they should).

Right now, that would mean extending the interface exposed by `gapstate.{c,h}` to include a way to run the constructor of a single module, say by using its name.

I think the more elegant solution, which IMHO also fits in better with the traditional module interface, is to add `initModuleState` and `deinitModuleState` members to struct `StructInitInfo`, instead of passing them to `RegisterModuleState`. This way, it's trivial to find and call the right module state methods when loading a kernel extension.

While I am at it, I think `doc/dev/kernel.xml` then should be updated to mention when exactly `initModuleState` is called, relative to `initKernel`/`initLibrary`, and what to move into it vs. `initLibrary`. (Maybe that should also be in a comment for struct

fingolfin:
This contains the main missing part of #2512, all except for the reordering of InitModuleState vs. InitLibrary, which is more trick than I thought (I'll discuss this elsewhere).

Resolves #2013


src/modules.c
```diff
@@ -949,6 +968,47 @@ void ModulesCheckInit(void)
...
+        if (info->initModuleState) {
+            if (SyDebugLoading) {
+                FPUTS_TO_STDERR("#I  InitModuleState(");
+                FPUTS_TO_STDERR(info->name);
+                FPUTS_TO_STDERR(")\n");
+            }
+            Int ret = info->initModuleState();
+            if (ret) {
+                FPUTS_TO_STDERR("#I  InitModuleState(");
+                FPUTS_TO_STDERR(info->name);
```
ChrisJefferson: If this happens, how bad is it? I feel like this message isn't very informative to the user if it ever appeared, and I would if we should SyExit if it does, as something REALLY horrible has happened?

ChrisJefferson:
I just have one comment (which could be done later) -- I'm generally in favour of SyExiting when something that appears REALLY awful seems to have happened, rather than trying to carry on in a broken state.

src/modules.c
fingolfin: I agree that calling `Panic` here (which also exits) would probably be better, *but* if we do that, I'd rather do it simultaneously for the code calling `initLibrary`, `initKernel`, etc., upon which this code is modelled. So if you don't mind, I'd rather merge this first, then follow up with a PR changing this to `Panic`

PR #2574, opened 2018-06-26
Change SyntaxError to underline complete last token

fingolfin:
This is an experimental patch, which changes the highlighting to syntax errors: instead of just pointing at the position we were when the syntax error occurred (which usually is at the end or just after the end of the problematic token), we now "underline" the whole token.

Example: Previously one got
```
gap> 1.x1;
Syntax error: Badly formed number in stream:1
1.x1;
  ^
```
With this PR
```
gap> 1.x1;
Syntax error: Badly formed number in stream:1
1.x1;
^^^
```
which I find slightly more helpful (esp. for longer expressions). For more examples, look at the changes to the test files.

Thoughts? Do people like or hate this? Or are neutral?

Note that `SymbolStartPos` and `SymbolStartLine` will be useful to implement something else we discussed in St Andrews: using the scanner+reader to tokenize and parse GAP code, w/o executing or coding it; instead, to only (programmatically) determine if it syntactically valid, and to extract the tokenization (e.g. for syntax highlighting in command line prompts). For that, we want to record start and end of tokens (we might want to do this in terms of an absolute seek position for a stream, though, but that's a separate concern), amongst other things.


alex-konovalov:
@fingolfin what are those Travis CI failures however?

fingolfin:
@alex-konovalov should be fixed now. We'll see for sure after the tests cycled.

frankluebeck:
I agree that `SymbolStartPos` and `SymbolStartLine` will be useful.

Concerning the `^` marks my  mild preference is for the old behaviour. In most cases the token is short and easy to detect from the trailing `^` mark, and for long tokens I find the huge number of `^` more confusing than helpful.

It would be nice if the code of the token could be underlined or colored instead of having the `^` marks.

fingolfin:
Yeah, coloring tokens is on my wish list, too (along with coloring tokens on the REPL, and when printing functions inside GAP)

ChrisJefferson:
I quite like this, and also like the code related to tracking tokens. The `?` output is very weird, but that' just because how we handle and parse `?` is very weird :)

ChrisJefferson:
I decided to merge this as I think there was a mild preference for merging, and also the debugging output gives us a useful test for checking if we continue tracking tokens correctly.

PR #2575, opened 2018-06-26
HPC-GAP: fix and improve object traversal code some more

fingolfin:
While the future of this code is not quite clear, these are some no-brainer fixes and improvements, so I don't think there is any harm in merging them.


ChrisJefferson:
Are we testing any of this?

fingolfin:
The diff coverage is 100%, i.e., all of the modified code is being exercised by the test suite. Object traversal is e.g. used for MakeImmutable and copying of objects in HPC-GAP.

PR #2578, opened 2018-06-26
Fix calling AtExit methods after loading a workspace

### LINKED ISSUE #2558: History does not work anymore if work space was loaded
### Observed behaviour

Saving the GAP history does not work anymore if a workspace was loaded during startup. So there is a difference between the behavior of

`gap` and `gap -L xyz`

Without `-L xyz` the history is saved and with `-L xyz` the history is not saved. In both cases GAPInfo.History starts with the content of the history file and ends with the commands of the current session.

### Expected behaviour

The history should be saved regardless of a workspace being loaded or not.

```
 ┌───────┐   GAP 4.9.1 of 05-May-2018
 └───────┘   Architecture: x86_64-apple-darwin17.5.0-default64
 Configuration:  gmp 6.1.2, readline
 Packages:   AClib 1.3, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.9, Browse 1.8.7, CRISP 1.4.4, Cryst 4.1.17, CrystCat 1.1.8, CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0,
             GAPDoc 1.6.1, IO 4.5.1, IRREDSOL 1.4, LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.13.1, PrimGrp 3.3.1, RadiRoot 2.8, ResClasses 4.7.1, SmallGrp 1.3,
             Sophus 1.24, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.54
```

ChrisJefferson:
This patch just ensures we call "CLEAN_UP_PROGRAM" after restoring a workspace, mainly so the AtExit handlers are run.

While fixing this I noticed a different (small) bug, which is when restoring a workspace some handlers end up getting installed twice. The easiest fix for this (in my opinion) is to just skip the second time the handler is installed.

Fixes #2558

### DIFF
--- a/lib/oper.g
+++ b/lib/oper.g
@@ -1616,6 +1616,7 @@ end );
 #F  InstallAtExit( <func> ) . . . . . . . . . . function to call when exiting
 ##
 BIND_GLOBAL( "InstallAtExit", function( func )
+    local f;
     if not IS_FUNCTION(func)  then
         Error( "<func> must be a function" );
     fi;
@@ -1624,6 +1625,13 @@ BIND_GLOBAL( "InstallAtExit", function( func )
             Error( "<func> must accept zero arguments" );
         fi;
     fi;
+    # Return if function has already been installed
+    # Use this long form to support both List and AtomicList
+    for f in GAPInfo.AtExitFuncs do
+        if f = func then
+            return;
+        fi;
+    od;
     ADD_LIST( GAPInfo.AtExitFuncs, func );
 end );
 
--- a/lib/session.g
+++ b/lib/session.g
@@ -71,6 +71,7 @@ BindGlobal("POST_RESTORE", function()
         fi;
     od;
     SESSION();
+    PROGRAM_CLEAN_UP();
 end);
 
 if IsHPCGAP then


ChrisJefferson:
Fixes #2558. This is unfortunately hard to test automatically, so I'd appresiate someone sanity checking it works. Basically, you should find saving history, and clean up of TemporaryDirectory, didn't work before this patch and does after this patch, if GAP is loaded with `-L`.

ChrisJefferson:
Fixes #2558. This is unfortunately hard to test automatically, so I'd appresiate someone sanity checking it works. Basically, you should find saving history didn't work before this patch and does after this patch, if GAP is started from a workspace.

fingolfin:
@ChrisJefferson is this perhaps related to issue #2558? If so, then the description of this PR should contain something like `Fixes #2558`, no?

fingolfin:
@mohamed-barakat please test if this solves your problem

alex-konovalov:
I've assigned this to 4.9.2 milestone - in a hope that @mohamed-barakat may confirm it quickly. Otherwise may be delayed till 4.9.3.

mohamed-barakat:
I am happy, thanks.

PR #2579, opened 2018-06-27
Add special case for comparing list with an infinite domain

### LINKED ISSUE #2503: Broken `Integers = [0];` in resclasses
Until 3252d1e3e8ce2f55, the code `Integers=[0]` worked, due to a method in resclasses. Somehow it's priority has fallen. The applicable methods are now:

```
function( D, C ) ... end
gap> ApplicableMethod(\=, [Integers, [0]], 1, 2);
#I  Searching Method for = with 2 arguments:
#I  Total: 187 entries
#I  Method
168: ``=: for a domain and a list'' at /mnt/c/Users/caj/files/reps/gap/gap/lib/domain.gi:
35 , value: 10
#I  Skipped:
#I  Method
172
 : ``=: for an infinite domain and a list of elements (ResClasses)'' at /mnt/c/Users/caj/files\ /reps/gap/gap/pkg/resclasses-4.7.1/lib/resclass.gi:1281 , value: 9
function( D, l ) ... end
```

Whereas they used to be:

```
#I  Searching Method for = with 2 arguments:
#I  Total: 187 entries
#I  Method
159
 : ``=: for an infinite domain and a list of elements (ResClasses)'' at /mnt/c/Users/caj/files/\ reps/gap/gap-clean/pkg/resclasses-4.7.1/lib/resclass.gi:1281 , value: 12
#I  Skipped:
#I  Method
171: ``=: for a domain and a list'' at /mnt/c/Users/caj/files/reps/gap/gap-clean/lib/domain.gi:
35 , value: 10
function( D, C ) ... end
```

It might be reasonable to add `HasIsFinite(D) and not(IsFinite(D))` to the `=` for a domain and a list in GAP, which would also solve this problem.

alex-konovalov:
Closes #2503.

### DIFF
--- a/lib/domain.gi
+++ b/lib/domain.gi
@@ -22,6 +22,9 @@ InstallMethod( \=,
     IsIdenticalObj,
     [ IsCollection and IsList, IsDomain ],
     function( C, D )
+    if HasIsFinite(C) and HasIsFinite(D) and (IsFinite(C) <> IsFinite(D)) then
+      return false;
+    fi;
     return IsSSortedList( C ) and AsSSortedList( D ) = C;
     end );
 
@@ -37,6 +40,9 @@ InstallMethod( \=,
     IsIdenticalObj,
     [ IsDomain, IsCollection and IsList ],
     function( D, C )
+    if HasIsFinite(C) and HasIsFinite(D) and (IsFinite(C) <> IsFinite(D)) then
+      return false;
+    fi;
     return IsSSortedList( C ) and AsSSortedList( D ) = C;
     end );
 
--- a/tst/testinstall/list.tst
+++ b/tst/testinstall/list.tst
@@ -19,6 +19,12 @@ false
 gap> [] = l;
 false
 
+# equality for a list and an infinite domain
+gap> CF(4)=[E(4)];
+false
+gap> [E(4)]=CF(4);
+false
+
 #
 # assignment / extraction, also with selectors
 #


lib/domain.gi
```diff
@@ -22,7 +22,11 @@ InstallMethod( \=,
     IsIdenticalObj,
     [ IsCollection and IsList, IsDomain ],
     function( C, D )
-    return IsSSortedList( C ) and AsSSortedList( D ) = C;
+    if HasIsFinite( D ) and not IsFinite( D ) then
+      return false;
```
fingolfin: Actually... Why is this a correct? Collections can be infinite, and so can lists, no (at least I just couldn't find anything in the documentaton which excludes the latter, and the filter `IsList` does not imply `IsFinite`). Am I missing something?

markuspf: Indeed:

```
gap> rats := Enumerator(Rationals);
<enumerator of Rationals>
gap> IsList(rats)
true
gap> IsFinite(rats);
false
```

fingolfin: So the test really needs to be more like this:
```gap
   if HasIsFinite(C) and HasIsFinite(D) and (IsFinite(C) <> IsFinite(D)) then
    return false;
  fi;
```
(I'd also not use the `fi` to keep the diff minimal)

alex-konovalov: @fingolfin @markuspf pushed updated PR.

lib/domain.gi
```diff
@@ -37,6 +40,9 @@ InstallMethod( \=,
     IsIdenticalObj,
     [ IsDomain, IsCollection and IsList ],
     function( D, C )
+    if HasIsFinite(C) and HasIsFinite(D) and (IsFinite(C) <> IsFinite(D)) then
+      return false;
+    fi;
```
fingolfin: Just some ramblings:

An alternative to this change would be this:
```gap
InstallMethod( \=, [HasIsFinite, HasIsFinite], 10,
function(x, y)
  if IsFinite(x) <> IsFinite(y) then return false; fi;
  TryNextMethod();
end);
```

Thinking some more about this: since this is a fallback method *and* we are going to enumerate `C` anyway, we could drop the two `HasIsFinite` checks, and just ask for `IsFinite` -- I am pretty sure that if `IsFinite(C)` fails to terminate, then so will `IsSSortedList(C)` (at least for any "sane" object).

PR #2580, opened 2018-06-27
kernel: don't include gmp.h from integer.h

fingolfin:
Instead only include it from .c files which really need it.

This way, kernel extensions do not need to be able to include gmp.h, which in
turn helps avoid complications in the compilation process, such as a mismatch
between the gmp.h used by GAP, and the one used by a kernel extension.


ChrisJefferson:
This is great. In the long term this will allow kernel extensions which don't care about GMP to not worry about finding the header, and handling it correctly in C++, but that should obviously wait until at least 4.10. I checked and this doesn't seem to break any packages.

PR #2584, opened 2018-06-27
HPC-GAP: fix varadic atomic functions

fingolfin:
Also reduce some code duplication.

### DIFF
--- a/src/funcs.c
+++ b/src/funcs.c
@@ -89,8 +89,7 @@ void SetRecursionDepth(Int depth)
 static Obj PushOptions;
 static Obj PopOptions;
 
-UInt ExecProccallOpts(
-    Stat                call )
+static UInt ExecProccallOpts(Stat call)
 {
   Obj opts;
   
@@ -204,8 +203,7 @@ static ALWAYS_INLINE Obj EvalOrExecCall(Int ignoreResult, UInt nr, Stat call)
 }
 
 
-
-UInt ExecProccall0args(Stat call)
+static UInt ExecProccall0args(Stat call)
 {
     EvalOrExecCall(1, 0, call);
 
@@ -213,7 +211,7 @@ UInt ExecProccall0args(Stat call)
     return 0;
 }
 
-UInt ExecProccall1args(Stat call)
+static UInt ExecProccall1args(Stat call)
 {
     EvalOrExecCall(1, 1, call);
 
@@ -221,7 +219,7 @@ UInt ExecProccall1args(Stat call)
     return 0;
 }
 
-UInt ExecProccall2args(Stat call)
+static UInt ExecProccall2args(Stat call)
 {
     EvalOrExecCall(1, 2, call);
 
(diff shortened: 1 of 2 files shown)


fingolfin:
I just rewrote this PR from scratch, after I realized that most of this code duplication in HPC-GAP is pointless: it all was there to avoid one check for `LCKS_FUNC(func)` in case of a function which does *not* use locks (which is of course still the vast majority in HPC-GAP). But this check of a value inside a struct which at this point will be in the cache anyway is virtually free, while duplicating all those function handlers takes up valuable space inside the cache, reducing the chance for other things to get cached.

Benchmarks reveal no performance change either way.

src/funcs.c
```diff
@@ -496,17 +485,16 @@ static void LockFuncArgs(Obj func, Int narg, const Obj * args)
...
+
 #ifdef HPCGAP
-    if (isAtomic)
+    REMEMBER_LOCKSTACK();
+    if (LCKS_FUNC(func))
         LockFuncArgs(func, narg, arg);
-#else
-    (void)isAtomic; // unused
 #endif
-    CHECK_RECURSION_BEFORE
```
fingolfin: So to summarize the changes here: in the above, we moved `CHECK_RECURSION_BEFORE` to come first; then we call `REMEMBER_LOCKSTACK` (but now only for HPC-GAP; in non-HPC-GAP, it was a no-op anyway). And finally we check `LCKS_FUNC(func)` to see if `LockFuncArgs` should be called.

src/funcs.c
```diff
@@ -528,81 +518,50 @@ static ALWAYS_INLINE Obj DoExecFunc(Obj func, Int narg, const Obj *arg, Int isAt
...
 
-Obj             DoExecFuncXargs (
-    Obj                 func,
-    Obj                 args )
+static Obj DoExecFuncXargs(Obj func, Obj args)
 {
     Bag                 oldLvars;       /* old values bag                  */
-    REMEMBER_LOCKSTACK();
     UInt                len;            /* number of arguments             */
     UInt                i;              /* loop variable                   */
```
fingolfin: In `DoExecFuncXargs`, the diff is a bit hard to read, because `git` resp. GitHub "merges" it with the (now removed) `DoExecFuncXargsL`. The realy diffs here are: `REMEMBER_LOCKSTACK` was moved move down; and `#ifdef HPCGAP ... LockFuncArgs` has been inserted.

src/funcs.c
```diff
@@ -791,9 +645,10 @@ Obj DoPartialUnWrapFunc(Obj func, Obj args)
     ASS_LVAR(named+1, args);
 
     /* execute the statement sequence                                      */
-    REMEMBER_LOCKSTACK();
```
fingolfin: The `REMEMBER_LOCKSTACK` above previously did nothing, as this function (`DoPartialUnWrapFunc`) never was called in HPC-GAP; this is the bug fixed by this PR. Since we now call `DoPartialUnWrapFunc` from HPC-GAP, we must  make sure to call `REMEMBER_LOCKSTACK` *before* we take any locks. Also, we call `CHECK_RECURSION_BEFORE` as the very first thing, before any other argument handling.

PR #2585, opened 2018-06-28
Rewrite EvalString

fingolfin:
Do not use a temporary variable to store results, and be more strict about
what we accept (to match the documentation).

This is an alternative to #2186. On the one hand, unlike that PR, this PR tries to stick to what is documented about `EvalString`. On the other hand, this PR makes `EvalString` a bit stricter, so that means one has to worry that it could break packages which rely on the current undocumented behaviour. So all in all, I am not sure it's worth it -- then again, it *is* IMHO quite a bit cleaner.

Closes #2186.

### DIFF
--- a/lib/string.gi
+++ b/lib/string.gi
@@ -642,23 +642,19 @@ InstallGlobalFunction("RemoveCharacters", REMOVE_CHARACTERS);
 ##
 #F  EvalString( <expr> ) . . . . . . . . . . . . evaluate a string expression
 ##
-_EVALSTRINGTMP := 0;
 InstallGlobalFunction("EvalString", function( s )
   local a, f, res;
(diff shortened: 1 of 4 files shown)


ChrisJefferson:
While I'm sure we've already spent too long on this, I prefer #2186. I probably have a greater preference for waiting until something like `ReadAsList`, or one of it's friends, are merged, so we can just evaluate a single expression and check it has a value (which we can then return). This will let us produce more easily a bunch of errors, like checking if the expression has no value, or checking if there was multiple expressions (I don't really like just silently ignoring other expressions, even though as you say the documentation says not to do that).

We could add some kind of check for `;` in the string other than at the end and Error in that case, which would make me happy (so at least users would not get an explicit Error not silent changes), but that seems like a lot of work :)

Issue #2434, opened 2018-05-02
GAP 4.10 should not require PrimgGrp/SmallGrp/TransGrp packages


alex-konovalov:
The libraries of small, primitive and transitive groups which previously were an integral part of &GAP; were split into three separate packages:
- [PrimgGrp](http://gap-packages.github.io/primgrp/)
- [SmallGrp](https://gap-packages.github.io/smallgrp/)
- [TransGrp](http://www.math.colostate.edu/~hulpke/transgrp/)

For backwards compatibility, these are required packages in GAP 4.9 (i.e., GAP will not start without them). We plan to change this for GAP 4.10, once all packages which currently implicitly rely on these new packages had time to add explicit dependencies on them. 

This issue under the milestone 4.10 is created to track this:
- list all packages that need to be updated
- notify package authors
- track progress

fingolfin:
Started work on this via #2589, but I am pretty sure tons of packages need changes...

E.g. lots of packages use `smallgrp` in their test files. This can be resolved by adding an explicit `LoadPackage("smallgrp")` to their `tst/testall.g` file. 

Just to get a first rough approximation of which packages might be affected, I did some very simple grepping to get a first approximation of the issue (note that will be both false positives and false negatives with this method):

Packages which seem to use `TransitiveGroup`:
- [ ] ctbllib -- patch sent to the maintainer by email (there is no public git repository)
- [ ] hap -- not investigated; it needs (almost?) all of its optional dependencies, so it is excluded from the `--bare` test runs
- [x] loops -- resolved in version 3.4.5 (TransGrp and SmallGrp are now dependencies)
- [x] radiroot -- fixed, TransGrp is now a dependency; merged, not yet released
- [x] SCSCP -- no issue
- [ ] simpcomp -- see https://github.com/simpcomp-team/simpcomp/pull/68
- [x] transgrp -- no issue

Packages which seem to use `PrimitiveGroup`:
- [ ] Browse -- not investigated
- [ ] ctbllib -- patch sent to the maintainer by email
- [x] digraphs -- `PrimitiveGroup` only appears in comments
- [x] genss -- fixed in git, the affected test is disabled; not yet released -- see https://github.com/gap-packages/genss/pull/20
- [x] grape -- resolved in version 4.9.3 -- see https://github.com/gap-packages/grape/pull/51
- [x] primgrp -- no issue
- [ ] recog -- not investigated
- [x] SCSCP -- no issue
[...]

PR #2589, opened 2018-06-28
Remove primgrp, smallgrp, transgrp from required packages

fingolfin:
This starts work on issue #2434. Let's see what breaks, if anything.

Though we may not see any breakage without dedicated tests, at least in packages...

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -1745,9 +1745,24 @@ DeclareUserPreference( rec(
 For backwards compatibility, the default lists most of packages \
 that were autoloaded in GAP 4.4 (add or remove packages as you like)."
     ],
-  default:= [ "autpgrp", "alnuth", "crisp", "ctbllib", "factint", "fga", 
-              "irredsol", "laguna", "polenta", "polycyclic", "resclasses", 
-              "sophus", "tomlib" ],
+  default:= [
+    "autpgrp",
+    "alnuth",
+    "crisp",
+    "ctbllib",
+    "factint",
+    "fga",
+    "irredsol",
+    "laguna",
+    "polenta",
+    "polycyclic",
+    "primgrp",
+    "resclasses",
+    "smallgrp",
+    "sophus",
+    "tomlib",
+    "transgrp",
+    ],
   values:= function() return RecNames( GAPInfo.PackagesInfo ); end,
   multi:= true,
   ) );
--- a/lib/system.g
+++ b/lib/system.g
@@ -30,9 +30,6 @@ BIND_GLOBAL( "GAPInfo", rec(
     Dependencies := MakeImmutable(rec(
       NeededOtherPackages := [
         [ "gapdoc", ">= 1.2" ],
-        [ "primgrp", ">= 3.1.0" ],
-        [ "smallgrp", ">= 1.0" ],
-        [ "transgrp", ">= 1.0" ],
       ],
     )),
 # There is no SuggestedOtherPackages here because the default value of


alex-konovalov:
@fingolfin thanks - just to let you know I've seen this and will run extended tests of PR using my parametrised Jenkins build - I'm specifying the number of the PR, and then id does the same as in 
 the nightly build for master or stable, but for this PR.

The changes are obviously correct, it's the impact of them that we need to estimate and report to authors of affected packages.

alex-konovalov:
Observed problems are of different kinds:

* Tests calling functions from these packages, e.g.

    List(AllSmallGroups(60), g -> Size(FrattiniSubgroup(g)));

2. Tests calling library functions relying on these packages, e.g.

```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-pull-request-quickt\
est/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/install/label/kovacs/GAP-pull-reques\
t-snapshot/tst/testinstall/opers/MaximalNormalSubgroups.tst:32
# Input is:
SortedList(List(MaximalNormalSubgroups(D2), StructureDescription));
# Expected output:
[ "C18", "D18", "D18" ]
# But found:
Error, the Small Groups identification is required but not installed
########
```

3. Side effects e.g.:

```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-pull-request-quickt\
est/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/install/label/kovacs/GAP-pull-reques\
t-snapshot/tst/testinstall/ctblmono.tst:117
# Input is:
TestMonomial( chi );
# Expected output:
rec( comment := "degree is index of Hall subgroup", isMonomial := true )
# But found:
rec( comment := "codegree is prime power", isMonomial := true )
########
```

4. Warnings and break loops when packages are loaded

5. Manual examples 

6. An of course, default tests of GAP packages specified in their `PackageInfo.g` files.

I think one of the first modifications of this PR would be to add these three packages to the set of packages loaded by default: 

https://github.com/gap-system/gap/blob/42789e61c61440f24663e08731b5351ef1ae2b6b/lib/package.gi#L1751

alex-konovalov:
@fingolfin ping?

fingolfin:
Updated as requested

alex-konovalov:
The number of diffs is still overwhelming. I would prefer to deal with this one by one. Either there should be three separate PRs, or you can try now to remove only PrimGrp and see what happens. The majority of diffs are coming from SmallGrp, and we need to discuss how to deal with that.

fingolfin:
To be honest, I am not sure it's worth the bother, so I don't intend to do anything about this PR. It really was meant more as a reminder; for now I think there are far more important issues to think about.

alex-konovalov:
Fine, so the summary is that many problems show up, they are classified as below, and we need to discuss how to deal with them and in which priority.

1. Tests calling functions from these packages
2. Tests calling library functions relying on these packages
3. Side effects (diffs when the test does not call a function from a package explicitly - less obvious)
4. Warnings and break loops when packages are loaded
5. Manual examples (now they run without diffs with default packages, but with errors when those three packages are missing)
6. Default tests of GAP packages specified in their PackageInfo.g files.

The complexity for untangling this for each package may be different.

fingolfin:
This PR has no chance to be merged anytime soon, if ever, and at the same time is trivial, so I don't see a point in keeping this open.

PR #2590, opened 2018-06-28
kernel: make reader state module local

fingolfin:


fingolfin:
@stevelinton not as far as I could tell. Here is what I tried: I read this file into GAP
```gap
f:=function() local g, x;
g:=function() if Factorial(2) = 3 then x:=[1,2,3,~,4,[5,6,~],7,8,[~]]; else x:=rec(a:=1,b:=2,c:=~.a); fi; return x; end;
... 16833 copies of the previous line ...
end;
```
and then run `GASMAN("collect"); Read("reader-bench.g"); time;` in GAP to see how long it takes 
* master without this PR: ~117 msec
* master with this PR: ~117 msec
* version 4.8.10: ~153 msec
* stable-4.9: ~400 msec (I have no idea what's going on there)

Anyway, despite this, I still updated the PR: Now, if HPC-GAP is disabled, the ReaderState is *not* put into the GAPState, but rather is put into a global variable, meaning that performance should be equal or *better* than before this PR. Not sure if people like this idea -- if yes, we could apply it to other modules, if not, I am also OK with removing it again (as it doesn't really seem to affect performance either way in practice). Perhaps @markuspf has an opinion on this?

stevelinton:
If there's no performance hit then I'd mildly suggest making the GAP and HPCGAP  versions as similar as possible, but I don't mind

fingolfin:
I am absolutely fine with ditching the HPC-GAP/non-HPC-GAP variations again (and kind of would prefer it myself, too), but I thought it'd be a good idea to at least mention the possibility (and showing the code beats words, I think ;-).

So I'll leave it a bit in, in case @ChrisJefferson or @markuspf would like to have a look, and also to let the tests cycle; I can the either change it, or leave it as is, say on Monday.

ChrisJefferson:
I'm a big fan of reducing changes, particularly when they make no measurable difference. Also my experience is the effects of adding/removing global variables is very unpredictable and changes by OS, so it's best to just write the simplest code / whatever you like best.

fingolfin:
@ChrisJefferson I am not quite sure how to interpret your comment. Are you suggesting any changes to this PR? Which changes? Esp. the bit about simplicity... this code is about as simple with the HPC-GAP specific changes as without them, IMHO. Though without them, there are fewer `#ifdef HPCGAP`, so perhaps you mean that with "simpler"? I.e. like @stevelinton, you suggest to return to the old form, where HPC-GAP and GAP were treated equally?

ChrisJefferson:
Sorry, I wrote too much.

I like the patch. You should write whatever code is best and simplest, not worrying about a possible tiny overhead.

fingolfin:
OK then. I unified the GAP and HPC-GAP versions again, and fixed some GCC-on-linux compiler warnings (I used `volatile` on pointers incorrectly). Should be ready to merge now.

PR #2592, opened 2018-06-29
Fix MemoizePosIntFunction for errorHandler not returning a value

markuspf:
Before this patch, if the optional errorHandler did not return a value,
we ran into a followup error that the error handler needs to return a value.

This PR changes the behaviour of the function returned by
`MemoizePosIntFunction` to just return if no value was returned by the
errorHandler.

### DIFF
--- a/lib/cache.gi
+++ b/lib/cache.gi
@@ -37,7 +37,7 @@ function(func, extra...)
     fi;
 
     original := AtomicList(options.defaults);
-    
+
     boundvals := MakeWriteOnceAtomic(AtomicList(original));
 
     if options.flush then
@@ -51,7 +51,11 @@ function(func, extra...)
     return function(val)
         local v, boundcpy;
         if not IsPosInt(val) then
-            return options.errorHandler(val);
+            v := CallFuncListWrap(options.errorHandler, [val]);
+            if v <> [] then
+                return v[1];
+            fi;
+            return;
         fi;
         # Make a copy of the reference to boundvals, in case the cache
         # is flushed, which will causes boundvals to be bound to a new list.
--- a/tst/testinstall/memoize.tst
+++ b/tst/testinstall/memoize.tst
@@ -46,6 +46,15 @@ gap> f3(2);
 gap> f3(6);
 Check:6
 36
+gap> f4 := MemoizePosIntFunction(func,
+> rec(defaults := [10,,20], errorHandler := function(x) Print("Woops\n"); end));;
+gap> f4(1);
+10
+gap> f4(2);
+Check:2
+4
+gap> f4("Hello, world");
+Woops
 
 # HPCGAP currently disables flushing caches
 gap> FlushCaches();


ChrisJefferson:
In general, i tend to prefer these kinds of functions to try to act like the functions they are wrapping, so if the error handler doesnt return a value, just do that.

(I realise that raises the question of if we shoukd support the function being memoized not returning a value. I think we should, but that is a bigger seperate discussion i suspect).

fingolfin:
@ChrisJefferson But what is the motivation for this preference? As stated, it's just a preference for its own sake. At least superficially, it only seems to complicate the calling code, but for no clear (to me gain); also, allowing the `errorHandler` to not return anything then needs to be documented; and it also means that now the function which is being memoized can sometimes return a value, and sometimes not -- and I thought we agreed in the past that this is in general highly undesirable. So why would we explicitly encourage it now?

Usual disclaimer: I am not saying there isn't a good reason for all this, just that I don't see it yet, and so am asking for an explanation.

BTW, I also kinda regret the name `errorHandler`, as this function is not just about "handling errors", but can also be about "dealing with inputs which are not positive integers -- in the latter case, it returns something, in the former case, it should either not return at all (only exit to the REPL), *or* it should return a value. To use kernel terminology: It should either use `ErrorReturnObj`, or else`ErrorQuit` resp. `ErrorMayQuit`, but *not* `ErrorReturnVoid`.

fingolfin:
(Re: `errorHandler`, just to be clear:  I approved that name back then, perhaps even suggested it (I forgot), so this is not to be read as blame, or at most as blame for me. I still wonder if we could change it to something better; but since GAP 4.9 was released with it, even if we rename it, we'd have to support the old name for some time at least)

ChrisJefferson:
*shrug*, while I might not like functions which sometimes return, and sometimes don't, I thought if it was easy to support them we may as well do. I would point out we don't have to change the documentation, the documentation already lines up with this patch (or at least, that is how I would read obviously, at minimum it's unclear what happens if the ErrorHandler doesn't return a value).

```
  errorHandler (defaults to Error (6.6-1))
        A  function  to  be called when an input which is not a positive integer is passed to the cache. If this function returns a value,
        that value is returned by the cache.
```

fingolfin:
Yes, I realize this is in line with what the documentation says. But I still don't understand with why what the documentations says right now is sensible or useful.

Point is: not everything that *can* be supported also *should* be supported. Sometimes it's better to disallow something, to make it harder to accidentally misuse it.

Now, maybe this is a case where I am wrong, and the `errorHandler` (which really should be called `invalidInputHandler` should be allowed to return "void"). But I find "we can do it, so why not" not very compelling.

Surely @markuspf came upon this while writing some concrete code. That code could serve as an example which can resolve this dispute, no?

markuspf:
I really don't care what GAP does. If people think this is not the correct way, just close this PR.

fingolfin:
@markuspf I am sad to hear that you don't care. But that leaves me confused: you certainly must have had a reason to open this PR, no? Perhaps you used `MemoizePosIntFunction` and run into this problem, or something like that? It would be interesting to learn what that was, to better understand the issue at hand.

But if you are indeed not interested in explaining, then yes, it's best to close this.

Issue #1800, opened 2017-10-23
Wrong syntax warning about undefined global variable.


sebasguts:
In 4.6, 4.7, 4.8, and master, the code
```gap
test_string := [ "a", "b", "c" ];
for i in [ 1 .. 10 ] do
 for k in [ 1 .. 10 ] do
   string := "aaaabbbb";
   xx := List( test_string, j -> PositionSublist( string, j ) );
 od;
od;
```
gives the syntax warning
```
Syntax warning: unbound global variable
    xx := List( test_string, j -> PositionSublist( string, j ) );
```
which is wrong. However,
```gap
test_string := [ "a", "b", "c" ];
for i in [ 1 .. 10 ] do
  for k in [ 1 .. 10 ] do
    string := "aaaabbbb";
    List( test_string, j -> PositionSublist( string, j ) );
  od;
od;
```
does not produce a warning. The above example is minimal in the sense that the warning does not appear with a single for loop, but only if there are at least two nested for loops.

Copy of banner:
```
 ┌───────┐   GAP 4.8.8-4401-g5f25772 of today
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-gcc-default64
 Configuration:  gmp 6.1.2, readline
```

fingolfin:
Actually, I am not so sure the warning is "wrong"; maybe it is undesirable, but GAP cannot easily tell the state of `string` at that point. For me, the bug is that the two snippets of code behave differently. I do not care that much whether they both show a warning, or both don't.

So, here is what (I think) is going wrong:

We only issue `SyntaxWarning("Unbound global variable");` in one place, namely inside function `ReadCallVarAss` from `read.c`. There, we have this check:
```C
    if ( type == 'g'                // Reading a global variable
      && mode != 'i'                // Not inside 'IsBound'
      && LEN_PLIST(STATE(StackNams)) != 0   // Inside a function
      && var != STATE(CurrLHSGVar)  // Not LHS of assignment
      && ValGVar(var) == 0          // Not an existing global var
      && ExprGVar(var) == 0         // Or an auto var
      && ! STATE(IntrIgnoring)      // Not currently ignoring parsed code
      && ! GlobalComesFromEnclosingForLoop(var) // Not loop variable
      && (GAPInfo == 0 || !IS_REC(GAPInfo)
          || !ISB_REC(GAPInfo,WarnOnUnboundGlobalsRNam) // Warning enabled
[...]

PR #2595, opened 2018-06-29
Fix missing syntax warning for using undefined gvar

fingolfin:
We normally issue a warning when code is parsed that uses a variable
identifier which has not yet been defined, which GAP then resolves as the
name of a gvar.

However, we suppress this warning if it is for the assignment to a global
variable.

But in some rare cases, this suppressed too much: if the very next line after
such an assignment wasn't an assignment itself, and accessed the gvar, we did
not print a warning.

This addresses part of issue #1800, though it does not implement the feature "requested" by @sebasguts there.

It also shows up one "problem" (?) in transgrp:
```
Syntax warning: Unbound global variable in GAPROOT/pkg/transgrp/lib/trans.grp:240
	  TRANSAVAILABLE[tradeg]:=true;
	                ^
```
Note that `TRANSAVAILABLE` actually gets defined before the offending line; but since it is wrapped in an atomic (and hence coding is enabled), the warning gets triggered. Simple fix would be to init `TRANSAVAILABLE` outside of that atomic; or even get rid of the atomic (it's not clear to me what it is for). I'll try to look into this and prepare a patch for transgrp. In the meantime, we should of course not merge this.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -806,11 +806,13 @@ static void ReadCallVarAss(TypSymbolSet follow, Char mode)
         }
         else {
             Match( S_ASSIGN, ":=", follow );
+            UInt currLHSGVar = ReaderState()->CurrLHSGVar;
             if ( LEN_PLIST(STATE(StackNams)) == 0 || !STATE(IntrCoding) ) {
                 ReaderState()->CurrLHSGVar = (ref.type == R_GVAR ? ref.var : 0);
             }
             ReadExpr( follow, 'r' );
             AssignRef(ref);
+            ReaderState()->CurrLHSGVar = currLHSGVar;
         }
     }
 
--- a/tst/testbugfix/2018-06-29-CurrLHSGVar.tst
+++ b/tst/testbugfix/2018-06-29-CurrLHSGVar.tst
@@ -1,0 +1,19 @@
+# This always produced a warning"
+gap> Unbind(string);
+gap> for i in [ 1 .. 10 ] do
+>     string := "aaaabbbb";
+>     xxx := List( [], x -> string );
+> od;
+Syntax warning: Unbound global variable in stream:3
+    xxx := List( [], x -> string );
+                                 ^
+
+# ... but this did not; now it does
+gap> Unbind(string);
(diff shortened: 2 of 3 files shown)


ChrisJefferson:
Good catch.

markuspf:
This needs a rebase.

fingolfin:
The relevant changes are in transgrp 2.0.3. So once that is in the distribution, I can update this PR to change GAP to require transgrp >= 2.0.3, after which we could in principle merge it.

hulpke:
@alex-konovalov 
2.0.4 has been in place since Friday.

alex-konovalov:
@hulpke ah, thanks for letting know. Check for package updates is scheduled for 9am daily from Monday till Friday, so I had no chance to find it out yet.

fingolfin:
Let me know when that new transpkg is picked up, then I can rebase this PR to trigger a rebuild and fresh tests.

That said, from my POV, there really is no need to rush this PR into 4.10; it could easily wait till 4.11.

alex-konovalov:
@fingolfin 
> Let me know when that new transpkg is picked up, then I can rebase this PR to trigger a rebuild and fresh tests.

You want to ask me to let you know when it is picked up *and* appears in packages-master.tar.gz archive. The former already happened, the latter - not yet. I will let you know. The chance of getting it for 4.10 is 80%.

alex-konovalov:
@fingolfin new TransGrp now approved and appears in packages-master.tar.gz and packages-stable-4.9.tar.gz archives. I will restart travis test to see whether it will now pass.

alex-konovalov:
@fingolfin please rebase, otherwise tests fail due to package build errors, addressed recently.

fingolfin:
Done

fingolfin:
GitHub did not receive the CI completion event, so I rerun one of the Travis jobs, and now GitHub noticed that all tests passed.

fingolfin:
I am going to rebase this once more before merge, just to make sure no weird conflict with master has crept in (I really wish we had something like <https://bors.tech> to take care of that; that one unfortunately does not support our rebase work flow, although they plan to add it eventually).

PR #2600, opened 2018-07-02
MemoizePosIntFunction: errorHandler -> invalidInputHandler

fingolfin:
Also clarify the documentation, and add a test case for when the invalidInputHandler returns without a value

This is an alternative to PR #2592.


fingolfin:
I understand the reluctance to change a document name. However, in this case, the old name, and any code using it, will keep working as-is, no need for your or anybody else to modify the code.

PR #2601, opened 2018-07-02
Fix crash in MakeImmutable(rec(x:=~));

### LINKED ISSUE #1912: `MakeImmutableObj` is broken for records
The following code crashes HPC-GAP

```
gap> x := rec();; x.x := x;;
gap> MakeImmutable(x);
```

fingolfin:
Also adds some comments and cleans up the code (also for plists)

Fixes #1912

### DIFF
--- a/src/plist.c
+++ b/src/plist.c
@@ -2643,7 +2643,6 @@ void MakeImmutablePlistInHom(Obj list)
         Obj elm = ELM_PLIST(list, i);
         if (elm != 0) {
             MakeImmutable(elm);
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
Nice catch, and nice fix (I was worried we would have to store some kind of stack of visited objects).

While this does introduce a tiny window of problems in HPC-GAP, I think it will be OK for now. Longer term we could change to another TNUM (not sure which would be good, maybe need a new one), just to stop people touching the object while it is being changed.

PR #2602, opened 2018-07-02
Debug improvements

ChrisJefferson:
This PR:

a) Fixes MEMORY_CANARY and adds a configure option for it.

b) Adds some awful, but at least existing, documentation for kernel debugging (further improvements / extensions welcome).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,24 @@ AC_ARG_ENABLE([memory-checking],
 AC_MSG_CHECKING([whether to enable memory checking])
 AC_MSG_RESULT([$enable_memory_checking])
 
+AC_ARG_ENABLE([valgrind],
+    [AS_HELP_STRING([--enable-valgrind], [enable valgrind extensions to GASMAN])],
+    [AC_DEFINE([MEMORY_CANARY], [1], [define if building with valgrind extensions])],
+    [enable_valgrind=no]
+    )
+AC_MSG_CHECKING([whether to enable valgrind extensions to GASMAN])
+AC_MSG_RESULT([$enable_valgrind])
+
+if test "x$enable_valgrind" != "xno" -a "x$enable_memory_checking" != "xno"; then
+    AC_MSG_ERROR([--enable-valgrind and --enable-memory-checking cannot be used at the same time])
+fi
+
+if test "x$with_gc" != "xgasman"; then
+    if test "x$enable_valgrind" != "xno" -o "x$enable_memory_checking" != "xno"; then
+        AC_MSG_ERROR([--enable-valgrind and --enable-memory-checking are only compatible with GASMAN])
+    fi
+fi
+
 dnl
 dnl User setting: Enable -Werror (off by default)
 dnl
--- a/doc/dev/lib.xml
+++ b/doc/dev/lib.xml
@@ -12,6 +12,69 @@ kernel among the developers.
 
 <Chapter Label="Chap-Lib">
 <Heading>Maintaining the &GAP; kernel and library</Heading>
+
+<Section Label="Sect-debugging">
+<Heading>Debugging the &GAP; Kernel</Heading>
+
+The &GAP; kernel supports a variety of options which can be enabled by running
+<C>configure</C> which provide various checks which can be useful when writing
+and debugging &GAP; kernel code.
+<P/>
+
+The first option, <C>--enable-debug</C>, is well supported can can be used
+whenever the &GAP; kernel, or kernel modules, are being developed (although
+developers should still check their code works without it).
+<P/>
+<C>--enable-debug</C> enables assertions throughout the kernel. These
+assertions are implemented using the <C>GAP_ASSERT</C> macro. These
+assertions are disabled when &GAP; is compiled without <C>--enable-debug</C>.
+When <C>--enable-debug</C> is enabled, <C>KernelDebug</C> will be shown in
+the line starting <C>Configuration:</C> when &GAP; is started.
+<P/>
+
+The next two options are more experimental, and are not designed to be constantly
+enabled. They are particularly designed to track down bugs relating to memory
+corruption relating to GASMAN (&GAP;'s memory manager). They cannot both be enabled
+at the same time.
+
+These options assume you are familiar somewhat familiar with the internals of GASMAN
+(see gasman.c for more information). In particular, GASMAN represents memory using
+an object of type Bag. The contents of these Bags can be moved during garbage collection.
+
+<P/>
+<C>--enable-memory-checking</C> makes &GAP; check for C pointers to the content of Bags
+being used after a new Bag has been created or a Bag is resized. GASMAN
+moves Bags during garbage collection, which can happen whenever a new Bag is created
+or a Bag is resized. However, as it is very unlikely that any single allocation
+will cause a garbage collection, these bugs trigger very rarely. Also the problems
+caused by these bugs are, if the C pointer is written to, a random other Bag,
+somewhere in &GAP;, is changed.
+<P/>
+
+After configuring, the memory checking must still be turned on. This can be done
+either by passing <C>--enableMemCheck</C> to &GAP;'s command line, or calling <C>GASMAN_MEM_CHECK(1)</C>.
+Note that enabling these tests makes &GAP; VERY, VERY slow. It can (depending on the
+machine and operating system) take &GAP; over a day to start, and load all standard
+packages. The recommending way to use this option is to start &GAP;, and then load
+small test files to try to isolate the problem. <C>MemCheck</C> will be shown in
+the line starting <C>Configuration:</C> when &GAP; is started.
+<P/>
+
+Known bugs: &GAP; will crash if <C>IO_fork</C> from the IO package is called while
+memory checking is enabled.
+<P/>
+
+<C>--enable-valgrind</C> makes changes to GASMAN so it is compatible with the
+<URL Text="valgrind">http://valgrind.org</URL> memory checking program.
+Without this, Valgrind will report many incorrect warnings relating to GASMAN
+and no useful warnings.
+<P/>
+
+At present, this <C>--enable-valgrind</C> only checks for invalid writes to the
+last bag which was created. Also, this option does not do anything unless &GAP;
+is run through Valgrind (for example by running <C>valgrind gap</C>.
+</Section>
+
 <Index>modules</Index>
 
 Modules, PROPOSALs, ...
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -271,40 +271,6 @@ Bag *                   AllocBags;
 UInt                    AllocSizeBags;
 Bag *                   EndBags;
 
-#if defined(MEMORY_CANARY)
-
-#include <valgrind/valgrind.h>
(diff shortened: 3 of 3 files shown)


doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
 
 <Chapter Label="Chap-Lib">
 <Heading>Maintaining the &GAP; kernel and library</Heading>
+
+<Section Label="Sect-debugging">
+<Heading>Debugging the GAP Kernel</Heading>
+
+The GAP kernel supports a variety of options which enabled when running <C>configure</C>,
```
alex-konovalov: `GAP` should be `&GAP;` throughout because as an entity it is displayed in a different font (as a `<Package` element).

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+After configuring, the memory checking must still be turned on. This can be done
+either by passing <C>--</C> to GAP's command line, or calling <C>GASMAN_MEM_CHECK(1)</C>.
+Note that enabling these tests makes GAP VERY, VERY slow. It can (depending on the
+machine and operating system) take GAP over a day to start, and load all standard
+packages. The recommending way to use this option is to start GAP, and then load
+small test files to try to isolate the problem. <C>MemCheck</C> will be shown in
+the Configuration line when GAP is started.
+<P/>
+
+Known bugs: GAP will crash if IO_fork is called while memory checking is enabled.
```
alex-konovalov: `IO_fork` should be a cross-reference to the IO package manual.

src/gasman.c
```diff
@@ -484,6 +450,39 @@ static inline UInt IS_BAG_BODY(void * ptr)
...
+  VALGRIND_MAKE_MEM_NOACCESS(AllocBags, canary_size());
+}
+void CLEAR_CANARY() {
+  VALGRIND_MAKE_MEM_DEFINED(AllocBags, canary_size());
+}
+#define CANARY_DISABLE_VALGRIND()  VALGRIND_DISABLE_ERROR_REPORTING
+#define CANARY_ENABLE_VALGRIND() VALGRIND_ENABLE_ERROR_REPORTING
+
+void CHANGED_BAG(Bag bag) {
+    CANARY_DISABLE_VALGRIND();
```
fingolfin: Optional: if you move this around, you could also run `git clang-format` on your commit

configure.ac
```diff
@@ -246,6 +246,18 @@ AC_ARG_ENABLE([memory-checking],
 AC_MSG_CHECKING([whether to enable memory checking])
 AC_MSG_RESULT([$enable_memory_checking])
 
+AC_ARG_ENABLE([valgrind],
+    [AS_HELP_STRING([--enable-valgrind], [add valgrind extensions to GASMAN])],
+    [AC_DEFINE([MEMORY_CANARY], [1], [define if build with valgrind extensions])],
+    [enable_valgrind=no]
+    )
+AC_MSG_CHECKING([whether to add valgrind extensions to GASMAN])
```
fingolfin: Optional: Instead of "add" here and above, I'd use "enable", to match the name of the option. Also, to me "add" has the wrong meaning here, but that's probably just me being a non-native speaker.

configure.ac
```diff
@@ -246,6 +246,18 @@ AC_ARG_ENABLE([memory-checking],
 AC_MSG_CHECKING([whether to enable memory checking])
 AC_MSG_RESULT([$enable_memory_checking])
 
+AC_ARG_ENABLE([valgrind],
+    [AS_HELP_STRING([--enable-valgrind], [add valgrind extensions to GASMAN])],
+    [AC_DEFINE([MEMORY_CANARY], [1], [define if build with valgrind extensions])],
```
fingolfin: optional: build -> built? Though everywhere else, we actually use "building"

configure.ac
```diff
@@ -246,6 +246,18 @@ AC_ARG_ENABLE([memory-checking],
...
+    [enable_valgrind=no]
+    )
+AC_MSG_CHECKING([whether to add valgrind extensions to GASMAN])
+AC_MSG_RESULT([$enable_valgrind])
+
+AS_IF([test "x$enable_valgrind" != "xno" -a "x$enable_memory_checking" != "xno"],
+    AC_MSG_ERROR([--enable-valgrind and --enable-memory-checking cannot be used at the same time])
+    )
+
```
fingolfin: Optional: If either is enabled, we could also check whether `$with_gc = "gasman"` holds, and if not, print an error.

doc/dev/lib.xml
fingolfin: Required: This sentence seems to be missing a word, like "which *can* be enabled *by* running" ? Or maybe reorder the sentence for readability (and use the appropriate XML for filenames):

> The GAP kernel supports a variety of compile-time options to help writing and debugging GAP kernel code.
> They can be enabled by passing one or more of the following options to the <F>configure</F> script.

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+being used after a new Bag has been created or a Bag is resized. GASMAN
+moves Bags during garbage collection, which can happen whenever a new Bag is created
+or a Bag is resized. However, as it is very unlikely that any single allocation
+will cause a garbage collection, these bugs trigger very rarely. Also the problems
+caused by these bugs are, if the C pointer is written to, a random other Bag,
+somewhere in GAP, is changed.
+<P/>
+
+After configuring, the memory checking must still be turned on. This can be done
+either by passing <C>--</C> to GAP's command line, or calling <C>GASMAN_MEM_CHECK(1)</C>.
```
fingolfin: Required: Shouldn't that `--` be `--enableMemCheck`?

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+
+The first option, <C>--enable-debug</C>, is well supported can can be used
+whenever the GAP kernel, or kernel modules, are being developed (although
+developers should still check their code works without it).
+<P/>
+<C>--enable-debug</C> enables assertions throughout the kernel. These
+assertions are implemented using the <C>GAP_ASSERT</C> macro. These
+assertions are disabled when GAP is compiled without <C>--enable-debug</C>.
+When <C>--enable-debug</C> is enabled, <C>KernelDebug</C> will be shown in
+the Configuration line when GAP is started.
```
fingolfin: Optional: Why is `Configuration` in upper case? If it is meant to "quote" the output of GAP, I'd mark it as such, e.g.
> the line starting with <C>Configuration:</C> when &GAP; is started.
Else, I'd use lower case. Same below.

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+<P/>
+
+The next two options are more experimental, and are not designed to be constantly
+enabled. They are particularly designed to track down bugs relating to memory
+corruption relating to GASMAN (GAP's memory manager). They cannot both be enabled
+at the same time.
+<P/>
+<C>--enable-memory-checking</C> makes GAP check for C pointers to the content of Bags
```
fingolfin: I'd not capitalize "Bag" here, though one might want to insert an explanation of what a bag is. Then again, it makes little sense to use this option if one does not know what a GAP "bag" is. Perhaps Insert in the previous line something like this:
> The following explanations assume you are somewhat familiar with the internals of GASMAN (see e.g. <F>src/gasman.c</F> for more information).

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+<P/>
+
+<C>--enable-valgrind</C> makes changes to GASMAN so it is compatible with the
+valgrind memory checking program. Without this, valgrind will report many 
+incorrect warnings relating to GASMAN, and no useful warnings.
+<P/>
+
+At present, this <C>--enable-valgrind</C> only checks for invalid writes to the
+last bag which was created. Also, this option does not do anything unless GAP
+is run through Valgrind (for example by running <C>valgrind ./gap</C>.
```
fingolfin: Why is Valgrind capitalized here but not above?

doc/dev/lib.xml
```diff
@@ -12,6 +12,61 @@ kernel among the developers.
...
+packages. The recommending way to use this option is to start GAP, and then load
+small test files to try to isolate the problem. <C>MemCheck</C> will be shown in
+the Configuration line when GAP is started.
+<P/>
+
+Known bugs: GAP will crash if IO_fork is called while memory checking is enabled.
+<P/>
+
+<C>--enable-valgrind</C> makes changes to GASMAN so it is compatible with the
+valgrind memory checking program. Without this, valgrind will report many
```
fingolfin: Optional: How about adding a URL here?
```
<URL Text="valgrind">http://valgrind.org</URL> memory checking program. Without this [...]
```

ChrisJefferson:
All improvements seem good, so I have made all of them. The documentation should still be significantly improved, but at least something exists now.

The only change I didn't make was cross-linking to IO_fork because for some reason I was getting errors I didn't understand, and making to the dev manual is quite annoying as it takes so long to run `make doc`.

doc/dev/lib.xml
```diff
@@ -12,6 +12,68 @@ kernel among the developers.
...
+the line starting <C>Configuration:</C> when &GAP; is started.
+<P/>
+
+The next two options are more experimental, and are not designed to be constantly
+enabled. They are particularly designed to track down bugs relating to memory
+corruption relating to GASMAN (&GAP;'s memory manager). They cannot both be enabled
+at the same time.
+
+These options assume you are familiar somewhat familiar with the internals of GASMAN
+(see gasman.c for more information). In particular, GASMAN represents memory using
```
fingolfin: strictly speaking, this should be `<F>gasman.c</F>`. But can also be changed later.

doc/dev/lib.xml
```diff
@@ -12,6 +12,68 @@ kernel among the developers.
...
+After configuring, the memory checking must still be turned on. This can be done
+either by passing <C>--enableMemCheck</C> to &GAP;'s command line, or calling <C>GASMAN_MEM_CHECK(1)</C>.
+Note that enabling these tests makes &GAP; VERY, VERY slow. It can (depending on the
+machine and operating system) take &GAP; over a day to start, and load all standard
+packages. The recommending way to use this option is to start &GAP;, and then load
+small test files to try to isolate the problem. <C>MemCheck</C> will be shown in
+the line starting <C>Configuration:</C> when &GAP; is started.
+<P/>
+
+Known bugs: &GAP; will crash if IO_fork is called while memory checking is enabled.
```
fingolfin: If this is not a ref for technical reasons, then shouldn't it at least be `<C>IO_fork</C>`? And perhaps add "from the IO package"?

fingolfin:
As these are low-level kernel changes, I don't think this needs to be documented in the release notes... But I am not 100% certain (it depends on how exactly we define what should and what shouldn't go into release notes.... and at whom we target them: end-users only? Or also kernel devs? Thoughts?

alex-konovalov:
Since this adds a new section "Debugging the GAP Kernel" to the manual, I would add to release notes some text conveniently linking it to that section.

Issue #1637, opened 2017-08-28
`InstallValue` is broken for (at least) Types


markuspf:
Currently GAP handles forward declarations and installations of values for global variables as follows:

 * A call to `DeclareGlobalVariable` registers a global variable name and installs a dummy value `ToBeDefined` for it
 * A later call to `InstallValue` copies the new value over the dummy object using `CLONE_OBJ`

Now the following is possible: 
```
gap> cache := [];
gap> makeval := function()
          local v;
          v := rec( canary := "hello" );
          Add(cache, v);
          return v;
end;;
gap> DeclareGlobalVariable("Cheese");
gap> InstallValue(Cheese, makeval());
gap> Cheese;
rec( canary := "hello" )
gap> cache;
[ rec( canary := "hello" ) ]
gap> Cheese.canary := "world";
"world"
gap> cache;
[ rec( canary := "hello" ) ]
```

This is roughly how the `NEW_TYPE` cache works, which means at the very least we must not install values for types this way.

I couldn't come up with a better solution for the `DeclareGlobalVariable`. `InstallValue` yet, and we might have to forbid `InstallValue` (or rather CLONE_OBJ) for all kinds of objects.

Maybe @stevelinton @fingolfin or @ChrisJefferson  have a bright idea.

ChrisJefferson:
So, what do we think things should look like?

I guess `InstallValue` should take a string containing the name of the variable, and then `InstallValue` would just look internally like `Cheese := makeval()`?

ChrisJefferson:
Stepping back a minute, what would we like `InstallValue` to do? Would we like it to be more like `Cheese := makeval()`?

stevelinton:
What we want is, I think, impossible. We want to find every reference to the "ToBeDefined" placeholder anywhere in the workspace and replace it by a reference to the exact object we are trying to install as its value. 

Since that's too hard, we change the object which is the ToBeDefined placeholder into a clone of the object we are trying to install. Mostly that's OK, but there is a problem when the identity of the value we are installing matters, which it does for a few things like types (because we maintain the invariant that equal types are identical.

fingolfin:
So here is a (nasty) way how we could perhaps resolve this: We add a new TNUM `T_PLACEHOLDER` (better name TBD). When we `DeclareGlobalVariable`, then we create a new instance of that type, and use it as placeholder object (instead of `"ToBeDefined"`). In it, we store the name resp. gvar id of the declared global var.

Now we go on in one of several ways:
[...]

PR #2604, opened 2018-07-03
HPC-GAP: flush method cache if the list of methods changes

### LINKED ISSUE #2594: SET_METHODS_OPERATION does not flush the methods cache.
### Observed behaviour

tests in #2521 fail in HPCGAP and I can't see how to rewrite the critical function to make them pass

### Expected behaviour

There should be a way to flush the methods cache (of an operation and number of arguments)
when needed.

```
 *********   HPC-GAP 4.10dev-862-g3c8b617 of today
 *********   Architecture: x86_64-apple-darwin17.6.0-hpcgap64
             Maximum concurrent threads: 4
 Configuration:  gmp 6.0.0

#W <<< This is an alpha release.      >>>
#W <<< Do not use for important work. >>>
```

It seems (looking at INSTALL_METHOD_FLAGS in oper1.g for instance, that `SET_METHODS_OPERATION(oper,n,MakeReadOnlySingleObj(meths));`                  
(after copying the old methods list into `meths` and editing it, should be the correct way to modify a methods list in HPCGAP, replacing editing it in-place and calling `CHANGED_METHODS_OPERATION(oper,n);` but this doesn't work. Nothing in `SET_METHODS_OPERATION` flushes the method cache.

fingolfin:
Fixes #2594

### DIFF
--- a/src/opers.c
+++ b/src/opers.c
@@ -1796,6 +1796,9 @@ static inline Obj CacheOper(Obj oper, UInt i)
 
     if (cache == 0) {
         len = (i < 7 ? CACHE_SIZE * (i + 2) : CACHE_SIZE * (1 + 2));
+#ifdef HPCGAP
+        len++; // reserve one slot for pointer to methods list
+#endif
         cache = NEW_PLIST(T_PLIST, len);
         SET_LEN_PLIST(cache, len);
 #ifdef HPCGAP
@@ -1831,7 +1834,10 @@ static ALWAYS_INLINE Obj GetMethodCached(Obj  cacheBag,
     UInt  i;
     const UInt cacheEntrySize = n + 2;
 
-    cache = 1 + ADDR_OBJ(cacheBag);
+    cache = BASE_PTR_PLIST(cacheBag);
+#ifdef HPCGAP
+    cache++; // skip over the pointer to the methods list
+#endif
 
     /* Up to CACHE_SIZE methods might be in the cache */
     if (prec < CACHE_SIZE) {
@@ -1886,7 +1892,10 @@ CacheMethod(Obj cacheBag, UInt n, Int prec, Obj * ids, Obj method)
     /* We insert this method at position <prec> and move
        the older methods down */
     UInt  cacheEntrySize = n + 2;
-    Obj * cache = 1 + prec * cacheEntrySize + ADDR_OBJ(cacheBag);
+    Obj * cache = BASE_PTR_PLIST(cacheBag) + prec * cacheEntrySize;
+#ifdef HPCGAP
+    cache++; // skip over the pointer to the methods list
+#endif
     SyMemmove(cache + cacheEntrySize, cache,
             sizeof(Obj) * (CACHE_SIZE - prec - 1) * cacheEntrySize);
     cache[0] = method;
@@ -2087,6 +2096,15 @@ static ALWAYS_INLINE Obj DoOperationNArgs(Obj  oper,
 
     Obj cacheBag = CacheOper(oper, n);
     Obj methods = METHS_OPER(oper, n);
+
+#ifdef HPCGAP
+    // reset the method cache if necessary
+    if (ELM_PLIST(cacheBag, 1) != methods) {
+        Obj * cache = BASE_PTR_PLIST(cacheBag);
+        cache[0] = methods;
+        memset(cache + 1, 0, SIZE_OBJ(cacheBag)-2*sizeof(Obj));
+    }
+#endif
 
     /* outer loop deals with TryNextMethod */
     prec = -1;
--- a/src/opers.c
+++ b/src/opers.c
@@ -1810,18 +1810,6 @@ static inline Obj CacheOper(Obj oper, UInt i)
     return cache;
 }
 
-
-#ifdef HPCGAP
-
-#define GET_METHOD_CACHE(oper, i)                                            \
-    (STATE(MethodCacheItems)[INT_INTOBJ(CACHE_OPER(oper, i))])
-
-#else
-
-#define GET_METHOD_CACHE(oper, i) CACHE_OPER(oper, i)
-
-#endif
-
 #ifdef COUNT_OPERS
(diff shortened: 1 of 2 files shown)


fingolfin:
I was confused by a test failure in `tst/testinstall/memoize.tst`; that file states "HPCGAP currently disables flushing caches" (I think @ChrisJefferson added that).

Well, turns out HPC-GAP does not so much "disable" `FlushCaches`, but rather just breaks it, due to method caching... So this PR fixes `FlushCaches` in HPC-GAP as a side effect. Nice.

ChrisJefferson:
I seem to remember we didn't want to enable flushcaches in hpcgap due to worries about how it would interact with threads. I think some cached are thread local, have you looked to see what flushcaches in hpcgap is now doing? I guess flushing the thread?

Sorry, going from memory and away from computer.

fingolfin:
@ChrisJefferson it is possible that somebody wanted to disable flushing of caches (perhaps @stevelinton or @rbehrends know or remember). But nothing actually explicitly "disabled" this -- they were not flushed due to a bug. If there is reason to disable cache flushing in HPC-GAP, then the way to do that is to insert code into `FlushCaches`, say inside of `if IsHPCGAP then`, and issue an error or warning message, instead of flushing.

fingolfin:
Looking at the available flush methods, the one for `MemoizePosIntFunction` of course goes to great lengths to be thread safe.  The only two other such methods in the library are installed by each call to `InstallFlushableValue` resp. `InstallFlushableValueFromFunction`. Nothing uses `InstallFlushableValueFromFunction` (except for our test suite). In general, I would try to get rid of both of them, and also `InstallValue`, because they rely on `CLONE_OBJ`. *That* is indeed a serious problem in multi threaded code -- but at the same time, I strongly suspect that most (if not all) code using `InstallFlushableValue` does not actually rely on that behavior. Actually, to the contrary, I think some code could *break* due to this, if `FlushCaches` was called at a bad moment -- so, yeah, perhaps that's a reason to disable cache flushing -- but I think a much better fix wold be to change `InstallFlushableValue` to not use `CLONE_OBJ`, which would solve two problems at once. See also issue #1637, where I just added some remarks on this, too.

fingolfin:
To explain why this change fixes `FlushCaches`, consider this: `FlushCaches` is a simple operation; to make use of it, one does something like this: `InstallMethod( FlushCaches, "return method", [], function() DoSomeFlushing();  TryNextMethod(); end );`. Note that `TryNextMethod` at the end, which is crucial, it chains the separate methods together. The last method being called then is the first  `InstallMethod( FlushCaches, "return method", [], function() end );` which has no `TryNextMethod` call in it.

All methods one install here have the same rank. For methods of the same rank, GAP orders the more recently installed before the previously installed ones. This means that the above `"return method"`, which is the very first method installed for `FlushCaches`, always gets called last.

... except that in HPC-GAP, we never flushed the method caches for `FlushCaches`. So the first time somebody called `FlushCaches`, we recorded some method as being applicable. Then on any subsequent call, we look at the method cache, and end up using that very same method. Never mind that perhaps new methods were installed in the meantime, which rank higher. We keep using the outdated cache.

Since this PR fixes flushing of method caches, it also fixes `FlushCaches`. Here is a much simple test for this. In GAP:
```
gap> DeclareOperation("Foobar", []);
gap> InstallMethod(Foobar, [], {} -> 1);
gap> InstallMethod(Foobar, [], {} -> 2);
gap> Foobar();
2
gap> InstallMethod(Foobar, [], {} -> 3);
gap> Foobar();
3
```
The same in HPC-GAP without this PR:
```
gap> DeclareOperation("Foobar", []);
gap> InstallMethod(Foobar, [], {} -> 1);
gap> InstallMethod(Foobar, [], {} -> 2);
gap> Foobar();
2
gap> InstallMethod(Foobar, [], {} -> 3);
gap> Foobar();
2   # ooops, new method is not used
```

I suspect many more things are silently broken due to this. In fact, I am surprised that there are not some catastrophic breakages caused by this -- probably because for most operations, we install all methods before ever calling them. But I suspect some issues would show up as soon as one uses packages which install new methods for old operations; my guess is that few people us HPC-GAP, and even less with packages that do this, which is why nobody noticed so far?!

src/opers.c
```diff
@@ -1796,6 +1796,9 @@ static inline Obj CacheOper(Obj oper, UInt i)
 
     if (cache == 0) {
         len = (i < 7 ? CACHE_SIZE * (i + 2) : CACHE_SIZE * (1 + 2));
```
stevelinton: What is going on here. I'm finding it hard to see why this line does anything.

src/opers.c
```diff
@@ -1796,6 +1796,9 @@ static inline Obj CacheOper(Obj oper, UInt i)
...
+#ifdef HPCGAP
```
stevelinton: Is it worth the complexity to save one word in non-HPC GAP. Why not just leave this word unused in regular GAP and save on code complexity. There might be a case for reorganising the method cache to  make more efficient use of the CPU cache, but that's a different problem

src/opers.c
```diff
@@ -3540,7 +3547,7 @@ Obj FuncCHANGED_METHODS_OPERATION (
     n = INT_INTOBJ( narg );
     cacheBag = CacheOper( oper, (UInt) n );
     cache = ADDR_OBJ( cacheBag );
-    for ( i = 0;  i < SIZE_OBJ(cacheBag) / sizeof(Obj);  i++ ) {
+    for ( i = 1;  i < SIZE_OBJ(cacheBag) / sizeof(Obj);  i++ ) {
```
stevelinton: This should probably be a memset.

src/opers.c
stevelinton: Oh, my bad, I misread the `1` in the else part as another `i`. This line is fine

stevelinton:
I've  merged this. The outstanding suggestions I have are cosmetic and I'd like to see if I can also merge the method reordering PR which is a bit more impactful.

PR #2605, opened 2018-07-03
kernel: reduce code duplication in FuncHASH_FLAGS

fingolfin:


src/opers.c
```diff
@@ -172,23 +172,25 @@ Obj FuncHASH_FLAGS (
...
+    // amounts to the same code, only the value of NRB_FLAGS has to be
+    // adjusted
+    len = NRB_FLAGS(flags) * (sizeof(UInt) / sizeof(UInt4));
     ptr = (UInt4 *)BLOCKS_FLAGS(flags);
     hash = 0;
     x    = 1;
     for ( i = len; i >= 1; i-- ) {
         hash = (hash + (*ptr % HASH_FLAGS_SIZE) * x) % HASH_FLAGS_SIZE;
-        x    = ((8*sizeof(UInt4)-1) * x) % HASH_FLAGS_SIZE;
+        x    = (31 * x) % HASH_FLAGS_SIZE;
```
fingolfin: I made the constant 31 explicit, as I find this much more readable; it also becomes clear that the factor is a prime, which makes it more plausible to call this a hash function.

PR #2609, opened 2018-07-03
kernel: add MAX_FUNC_LVARS, MAX_FUNC_EXPR_NESTING

### LINKED ISSUE #2206: kernel idea: make MAX_FUNC_EXPR_NESTING global, use it in read.c, and lower it fro 1024 to e.g. 32
Right now, there are at least two places which limit how deep functions (resp. closures) can be nested in each other:
* in `code.c`, we `#define MAX_FUNC_EXPR_NESTING 1024` and use that as a hard limit on how deep one can nest functions. Going beyond this limit triggers an assertion (resp. silently corrupts memory, if assertions are turned off).
* in `read.c`, specifically in `ReadCallVarAss`, we limit the nesting to 65536 on 32bit platforms and not at all (resp. to 2^48) on 64bit platforms:
```C
#ifndef SYS_IS_64_BIT
        // we store nest and indx together in var; if we are on a 32bit
        // system, then this limits both to 16 bits, which we enforce here
        if (nest >= 65536) {
            Pr("Warning: abandoning search for %s at 65536th higher frame\n",
               (Int)STATE(Value), 0L);
            break;
        }
#endif
```

These two limits should be aligned, say by making `MAX_FUNC_EXPR_NESTING` global. For technical reasons, it should stay a power of 2.

Furthermore,, I don't think I ever saw a nesting deeper than 4 or so (perhaps a bit more with the use of nested `x -> foo` lambda functions). Realistically, going much deeper becomes rather painful, and I doubt it matters. So we might also want to consider reducing the nesting limit to, say, 32. This frees up valuable bits which can be used to store variable reference more compactly (I am thinking about unifying the var identifiers for gvars, lvars, hvars, dvars into a single 32 / 64 bit word)... Anyway, right now there is no pressing reason to reduce the limit, but it will be easier if the code was refactored to use `MAX_FUNC_EXPR_NESTING` uniformly, which to me is another reason why this refactoring would be useful.

Lastly, I think it's not good that `code.c` crashes if the nesting is too deep -- it really should produce an error. To test a hypothetical fix for this, I'd temporarily reduce `MAX_FUNC_EXPR_NESTING` to a much smaller value, making it easier to trigger the problem with test code... Alternatively, one can write a function which generates a string containing code for a set of functions nested to a certain fixed level, in order to generate tests cases, even for the 1024 level function.

So in summary:
* make `MAX_FUNC_EXPR_NESTING` global and change `read.c` to use it,
* change `code.c` to give an error instead of crashing when `MAX_FUNC_EXPR_NESTING` is exceeded,
* add a test case for the latter,
* perhaps reduce `MAX_FUNC_EXPR_NESTING` to a smaller value

fingolfin:
... instead of hardcoding these values in various places (which in fact was also done inconsistently before)

Also simplify a check in `EvalRefHVar`

Resolves #2206

### DIFF
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -2697,7 +2697,7 @@ void            IntrAssDVar (
 
     if ( STATE(IntrCoding) > 0 ) {
         ErrorQuit( "Variable: <debug-variable-%d-%d> cannot be used here",
-                   dvar >> 10, dvar & 0x3FF );
+                   dvar >> MAX_FUNC_LVARS_BITS, dvar & MAX_FUNC_LVARS_MASK );
     }
 
 
@@ -2728,7 +2728,7 @@ void            IntrUnbDVar (
 
     if ( STATE(IntrCoding) > 0 ) {
         ErrorQuit( "Variable: <debug-variable-%d-%d> cannot be used here",
-                   dvar >> 10, dvar & 0x3FF );
+                   dvar >> MAX_FUNC_LVARS_BITS, dvar & MAX_FUNC_LVARS_MASK );
     }
 
     /* assign the right hand side                                          */
@@ -2761,7 +2761,7 @@ void            IntrRefDVar (
 
     if ( STATE(IntrCoding) > 0 ) {
         ErrorQuit( "Variable: <debug-variable-%d-%d> cannot be used here",
-                   dvar >> 10, dvar & 0x3FF );
+                   dvar >> MAX_FUNC_LVARS_BITS, dvar & MAX_FUNC_LVARS_MASK );
     }
 
     /* get and check the value                                             */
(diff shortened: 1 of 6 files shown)


ChrisJefferson:
While this looks fine, it seems to be failing testing (but is that just a Travis problem? Needs investigating)

src/intrprtr.c
```diff
@@ -2697,7 +2697,7 @@ void            IntrAssDVar (
 
     if ( STATE(IntrCoding) > 0 ) {
         ErrorQuit( "Variable: <debug-variable-%d-%d> cannot be used here",
-                   dvar >> 10, dvar & 0x3FF );
+                   dvar >> MAX_FUNC_LVARS_BITS, dvar & MAX_FUNC_LVARS_MASK );
```
fingolfin: Note that we replace 10 by 16 here (also in the mask). But that's actually correct: The limit used to be 1024 at some point, and was raised to 65536 later, but these four ErrorQuit calls were not adapted.

fingolfin:
@ChrisJefferson there was a stupid typo in the definition of `MAX_FUNC_LVARS_MASK` (it was being set to `MAX_FUNC_LVARS_BITS-1` instead of `MAX_FUNC_LVARS-1`...). Should be fixed now. We'll see once the tests cycled.

PR #2613, opened 2018-07-04
Fixed the generators in OmegaZero(2m+1,2^k)

### LINKED ISSUE #2576: No suitable invariant bilinear/quadratic form pairs exist for `Omega(0,d,q)` with d odd, q even
I recently decided to try and add the `InvariantBilinearForm` and `InvariantQuadraticForm` attributes for `Omega(e, d, q)`.

It turns out that for `e=0` and `q` even, this is not possible?! Consider for example `Omega(0,3,4);` -- an exhaustive search over `GF(4)^[3,3]` reveals that the only non-trivial bilinear form preserved by this group in GAP is (up to scalars) given by the Gram matrix:
```
 . . .
 . 1 .
 . . .
```
Actually, this can be verified  for Omega(0,3,q) for any even q>=4 with a simple hand calculation, by inspecting the two generators of the group.

But this matrix cannot be written as `Q + Q^tr` for some matrix `Q` over `GF(4)`, as such a matrix always has only zeros on the diagonal. But this is required by the GAP manual entry for `InvariantQuadraticForm`.

I could fix this by using a different representation of the group (e.g. the one obtained via `DerivedSubgroup(GO(e,d,q))`), but that might break compatibility in unforeseen ways? See also issue #500.

Perhaps @ThomasBreuer @frankluebeck @hulpke have some insights on this?

ThomasBreuer:
For odd d and even q,
the generators for the matrix groups Omega(0,d,q)
from the Rylands/Taylor paper were not correct;
one would have to transpose the matrices in order to get
a group that respects a form as required.

Instead of transposing the generators, we delegate to `SO`,
which has the advantage to reduce the possible confusion
about the choice of different forms for related orthogonal groups.

Resolves #2576

### DIFF
--- a/grp/classic.gi
+++ b/grp/classic.gi
@@ -1491,24 +1491,30 @@ end);
 #F  OmegaZero( <d>, <q> ) . . . . . . . . . . . . . . . . \Omega^0_{<d>}(<q>)
 ##
 BindGlobal( "OmegaZero", function( d, q )
-    local f, o, m, mo, n, i, x1, x2, x, g, xi, h, s, q2, q2i;
+    local f, o, m, mo, n, i, x, g, xi, h, s, q2, q2i;
 
     # <d> must be odd
     if d mod 2 = 0 then
       Error( "<d> must be odd" );
     elif d < 3 then
       Error( "<d> must be at least 3" );
+    elif q mod 2 = 0 then
+      # For even q, the generators claimed in [RylandsTalor98] are wrong:
+      # For (d,q) = (5,2), the matrices generate only S4(2)' not S4(2).
+      # In the other cases, the matrices would have to be transposed
+      # in order to respect a form as required;
+      # thus they describe a group of the right isomorphism type
+      # but not an orthogonal group.
+      # We return the isomorphic group SO(d,q) in these cases;
+      # note that this is the definition of Omega(d,q) for odd d and even q
+      # in the ATLAS of Finite Groups [CCN85, p. xi].
+      return SO( d, q );
     fi;
     f:= GF(q);
     o:= One( f );
     m:= ( d-1 ) / 2;
 
-    if d = 5 and q = 2 then
-      # The matrices given in [RylandsTalor98] generate only A6 not S6.
-      # So we take the isomorphic group SO( 5, 2 ) instead.
-      return SO( 5, 2 );
-
-    elif 3 < d then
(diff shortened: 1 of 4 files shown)


tst/testinstall/grp/classic-G.tst
```diff
@@ -131,7 +131,7 @@ Error, <subfield> must be a prime or a finite field
 
 #
 gap> Omega(3,2);
-Omega(0,3,2)
+GO(0,3,2)
```
fingolfin: Please also update `tst/testinstall/grp/classic-forms.tst` to verify the forms in these cases, i.e.,
```patch
--- a/tst/testinstall/grp/classic-forms.tst
+++ b/tst/testinstall/grp/classic-forms.tst
@@ -126,16 +126,9 @@ gap> for d in [3,5,7] do
 > od;
 gap> ForAll(grps, CheckGeneratorsSpecial);
 true
-
-# FIXME: forms are not implemented for odd d, even q
-# gap> ForAll(grps, CheckBilinearForm);
-# true
-# gap> ForAll(grps, CheckQuadraticForm);
-# true
-#
-gap> ForAll(Filtered(grps, g -> Characteristic(g)<>2), CheckBilinearForm);
+gap> ForAll(grps, CheckBilinearForm);
 true
-gap> ForAll(Filtered(grps, g -> Characteristic(g)<>2), CheckQuadraticForm);
+gap> ForAll(grps, CheckQuadraticForm);
 true
 gap> ForAll(grps, CheckSize);
 true
```

grp/classic.gi
```diff
@@ -1525,22 +1531,11 @@ BindGlobal( "OmegaZero", function( d, q )
...
 
-      if q mod 2 = 0 then
-        # $x = x_{\epsilon_1 - \epsilon_m}(1) x_{-\alpha_1}(1)$
-        x1:= IdentityMat( d, f );
-        x1[1][m]:= o;
-        x1[ m+2 ][d]:= o;
-        x2:= IdentityMat( d, f );
-        x2[ m+1 ][m]:= o;
-        x2[ m+2 ][m]:= o;
-        x:= x1 * x2;
```
fingolfin: Optional: now x1 and x2 are not used anymore in this function, and could be removed from the `local` list above.

fingolfin:
LGTM!

PR #2615, opened 2018-07-04
kernel: remove unused {8,16,32}Bits_Quotient

### LINKED ISSUE #2504: Unused kernel function 8Bits_Quotient, 16Bits_Quotient and 32Bits_Quotient -- use or remove?
While working on PR #2499 I discovered that the kernel functions `8Bits_Quotient`, `16Bits_Quotient` and `32Bits_Quotient` are never used, though in theory we could install them as methods for computing quotients of associative words in syllable representation.

Pros: they might be faster than using inverson + multplication
Con: Since they were never used, they are mostly untested, and might not work correctly.

Thoughts? Pinging esp. @ThomasBreuer and @hulpke as their names appear on `lib/wordrep.gi`

fingolfin:
This is an alternative to #2616.

Resolves #2504. Closes #2616 .

### DIFF
--- a/src/objfgelm.c
+++ b/src/objfgelm.c
@@ -947,90 +947,6 @@ Obj Func8Bits_Product (
 
 /****************************************************************************
 **
-*F  Func8Bits_Quotient( <self>, <l>, <r> )
-*/
-Obj Func8Bits_Quotient (
-    Obj         self,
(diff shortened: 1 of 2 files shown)


fingolfin:
@hulpke that's what the other PR does ...

PR #2616, opened 2018-07-04
Install {8,16,32}Bits_Quotient kernel funcs as methods

### LINKED ISSUE #2504: Unused kernel function 8Bits_Quotient, 16Bits_Quotient and 32Bits_Quotient -- use or remove?

fingolfin:
This is an alternative to #2615.

Note that some benchmarks indicate that these kernel functions do provide a noticeable speed up: (here `x32` and `u32` are defined in `tst/testinstall/wordrep.tst`):
```
gap> GASMAN("collect"); for i in [1..1000000] do x:=x32/u32; od; time;
1542
gap> GASMAN("collect"); for i in [1..1000000] do x:=x32/u32; od; time;
1475
gap> GASMAN("collect"); for i in [1..1000000] do x:=32Bits_Quotient(x32,u32); od; time;
91
gap> GASMAN("collect"); for i in [1..1000000] do x:=32Bits_Quotient(x32,u32); od; time;
87
```
But I wonder if there is a reason they were not used?

Resolves #2504. Closes #2615.

### DIFF
--- a/lib/wordrep.gi
+++ b/lib/wordrep.gi
@@ -468,6 +468,12 @@ InstallMethod( \*,
     [ Is8BitsAssocWord, Is8BitsAssocWord ], 0,
     8Bits_Product );
 
+InstallMethod( \/,
+    "for two 8 bits assoc. words",
+    IsIdenticalObj,
+    [ Is8BitsAssocWord, Is8BitsAssocWord and IsMultiplicativeElementWithInverse ], 0,
+    8Bits_Quotient );
+
 InstallMethod( OneOp,
     "for an 8 bits assoc. word-with-one",
     true,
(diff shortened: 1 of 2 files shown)


fingolfin:
@hulpke This is the alternative PR which tests `32Bit_Quotient` and shows that it provides a performance boost.

lib/wordrep.gi
```diff
@@ -468,6 +468,12 @@ InstallMethod( \*,
     [ Is8BitsAssocWord, Is8BitsAssocWord ], 0,
     8Bits_Product );
 
+InstallOtherMethod( \/,
+    "for two 8 bits assoc. words",
+    IsIdenticalObj,
+    [ Is8BitsAssocWord, Is8BitsAssocWord ], 0,
+    8Bits_Quotient );
```
ThomasBreuer: This method would be applicable also to elements of, for example, `FreeMonoid( IsSyllableWordsFamily, 2 )`,
with the effect that corrupted objects arise.
If `IsMultiplicativeElementWithInverse` is required for the second argument then this cannot happen,
and one can use `InstallMethod` instead of `InstallOtherMethod`.

fingolfin:
@ThomasBreuer Thank you! I just updated the PR based on your comments

markuspf:
I think we should keep the functions and install them as methods, i.e. I prefer this PR over #2615

PR #2617, opened 2018-07-04
Fix HighestWeightModule for Lie algebras in certain cases

### LINKED ISSUE #2491: HighestWeightModule function in Lie Algebras package
The HighestWeightModule(-) is giving an unexpected error message for certain weights:

### Observed behaviour

First, creating a Lie algebra works fine:
```
gap> L := SimpleLieAlgebra("G",2,Rationals);
<Lie algebra of dimension 14 over Rationals>
```
Then the following command gives the subsequent error message:
```
gap> W := HighestWeightModule(L,[4,1]);
Error, no 1st choice method found for `[]' on 2 arguments
 at /proc/cygdrive/C/gap-4.9.1/lib/methsel2.g:250 called from
o[pos] at /proc/cygdrive/C/gap-4.9.1/lib/lierep.gi:3285 called from
lexord( novar, y, x ) at /proc/cygdrive/C/gap-4.9.1/lib/lierep.gi:3569 called from
Sort( sortmn[j], function ( x, y )
      return lexord( novar, y, x );
  end ); at /proc/cygdrive/C/gap-4.9.1/lib/lierep.gi:3569 called from
<function "unknown">( <arguments> )
you can 'return;' to continue
brk>
```
Replacing the weight [4,1] with other "large" weights, gives a similar error message.

### Expected behaviour

For a smaller weight, one gets a result such as the following:
```
gap> V := HighestWeightModule(L,[2,2]);
<729-dimensional left-module over <Lie algebra of dimension 14 over Rationals>>
```
It is possible that there is some sort of sort of memory issue  as the dimension of the module gets larger, but one would expect a more clear memory error then.

The same issue results in both a 32-bit and 64-bit installation on a Windows laptop:
```
 GAP 4.9.1 of 05-May-2018
 └───────┘   Architecture: i686-pc-cygwin-default32
 Configuration:  gmp 6.0.0, readline
 Packages:   AClib 1.3, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.9, Browse 1.8.7, CRISP 1.4.4, Cryst 4.1.17,
             CrystCat 1.1.8, CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.1, IO 4.5.1, IRREDSOL 1.4,
             LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.13.1, PrimGrp 3.3.1, RadiRoot 2.8, ResClasses 4.7.1,
             SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.54
```
```
 GAP 4.9.1 of 05-May-2018
 └───────┘   Architecture: x86_64-unknown-cygwin-default64
 Configuration:  gmp 6.1.0, readline
 Packages:   AClib 1.3, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.9, Browse 1.8.7, CRISP 1.4.4, Cryst 4.1.17,
             CrystCat 1.1.8, CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.1, IO 4.5.1, IRREDSOL 1.4,
             LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.13.1, PrimGrp 3.3.1, RadiRoot 2.8, ResClasses 4.7.1,
             SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.54
```
A colleagues has also tried the code on a Macbook, with the same result.

fingolfin:
This is a regression in GAP 4.9 compared to GAP 4.8. As far as I can tell, it
is caused by the new sort functions; for some reason I don't quite
understand, the new sort ends up comparing an element to itself. But the
custom `lexord` sorting function used by `HighestWeightModule` was not
prepared to deal with two equal (much less identical) inputs.

Fixes #2491 

I suggest to backport this to stable-4.9 -- it can wait for GAP 4.9.3, though, if adding it now is inconvenient.

### DIFF
--- a/lib/lierep.gi
+++ b/lib/lierep.gi
@@ -3282,6 +3282,9 @@ InstallMethod( HighestWeightModule,
         
         o:= n2-n1;
         pos:= PositionProperty( o, x -> x <> 0 );
+        if pos = fail then
+            return false;
+        fi;
         return o[pos] < 0;
     end;
 
@@ -3552,7 +3555,7 @@ InstallMethod( HighestWeightModule,
         # be a factor of `b' if we have `a=b'. So reduction within a
         # weight component is the same as linear algebra. We use the 
         # mutable bases in `sps' to perform the linear algebra.
-                                 
+    
         sps:= [ ];
         sortmn:= [ ];                         
         for j in [1..Length(w)] do
--- a/tst/testbugfix/2018-07-04-HighestWeightModule.tst
+++ b/tst/testbugfix/2018-07-04-HighestWeightModule.tst
@@ -1,0 +1,5 @@
+gap> L := SimpleLieAlgebra("G",2,Rationals);
+<Lie algebra of dimension 14 over Rationals>
+gap> W := HighestWeightModule(L,[4,1]);
+<924-dimensional left-module over <Lie algebra of dimension 
+14 over Rationals>>


fingolfin:
@ChrisJefferson I really wonder why the sorting function ends up comparing an element to itself (the input is free of duplicates; the comparison function really receives the same (identical) object twice as input.

fingolfin:
Note that PR #2621 also fixes this issue, but I really want this one in, too, as arguably the comparator is wrong otherwise (we could have to equal but non-identical entries in a list being sorted, after all... OK, maybe not in the code being modified here, but...)

fingolfin:
It worked before because the comparator never was called with two inputs that happened to be equal. But this changed when we changed our sorting algo.

PR #2621, opened 2018-07-04
Stop Sort from comparing identical objects

### LINKED ISSUE #2618: `Sort` sometimes compares an object to itself
The following code triggers the assertion, i.e., `Sort` ends up comparing an object to itself. While this is not incorrect as such, it seems like a missed micro optimization. Also, it's just generally suspicious, so I'd love to understand this better:
```gap
l:=[
  [ 1, 1, 2, 1, 3, 1, 4, 2, 5, 1 ], [ 1, 1, 2, 1, 3, 2, 5, 2 ],
  [ 1, 3, 2, 1, 3, 1, 4, 1, 6, 1 ], [ 1, 1, 2, 1, 4, 4 ],
  [ 1, 2, 3, 2, 4, 1, 6, 1 ], [ 1, 1, 3, 2, 4, 3 ],
  [ 1, 1, 3, 2, 5, 1, 6, 1 ], [ 1, 1, 3, 3, 4, 1, 5, 1 ],
  [ 1, 1, 4, 1, 6, 2 ], [ 1, 2, 2, 1, 3, 1, 5, 1, 6, 1 ],
  [ 1, 2, 2, 1, 3, 2, 4, 1, 5, 1 ], [ 1, 2, 2, 1, 4, 2, 6, 1 ],
  [ 1, 2, 3, 1, 6, 2 ], [ 1, 1, 3, 1, 4, 2, 6, 1 ],
  [ 1, 1, 2, 1, 4, 1, 5, 1, 6, 1 ], [ 1, 3, 2, 1, 6, 2 ],
  [ 2, 1, 3, 1, 4, 1, 5, 2 ], [ 2, 1, 4, 3, 5, 1 ], [ 2, 1, 5, 2, 6, 1 ],
  [ 3, 1, 4, 1, 5, 1, 6, 1 ], [ 3, 1, 4, 4 ], [ 3, 2, 4, 2, 5, 1 ],
  [ 3, 3, 5, 2 ], [ 4, 3, 6, 1 ], [ 5, 1, 6, 2 ]
];
Sort(l, function(x,y) Assert(0, not IsIdenticalObj(x,y)); return x<y; end);
```

This example extracted from issue #2491; to fix that, PR #2617 contains  a change to the comperator which fixes the problem there.

ChrisJefferson:
Some comparators do not handle comparing identical objects, which
occurs when we compare values with the pivot during quicksort.

This fixes #2618.

The fundamental problem is that we compare values with the pivot when doing quicksort partitioning.

Just in case anyone thinks "Wait, this is stupid. Can't I just move the quicksort pivot out of the way, maybe to the start of the array?", I warn you that that is move difficult than it sounds -- by moving the pivot to the start of the array (and then later to the pivot place), you then break that the "middle value from first, last, middle" is picking (reasonably) random elements of the array, which makes quicksort hit it's worst cases much more often.

This can be fixed (you have never pivot using elements of the array where you have put special things)- but then you need to be careful with array length, and this tends to make the pdqsort algorithm less efficient, as bits of the array get suffled. I did this long ago in g++ and messed it up a couple of times, so don't really want to go through it again if it's not absolutely required :)

In short, if you want to try to do this better, be prepared to carefully benchmark a bunch of cases (array nearly sorted and array nearly reversed are the main two).


src/sortbase.h
```diff
@@ -65,6 +65,20 @@
...
 #define JOIN2(x, y) x##y
 
+
+// Compare a and b, first checking if a and b are equal objects,
+// and if so returning false straight away. We do this because
+// some comparators do not work when comparing equal objects.
+// This can occur either because the original list had identical
+// objects, or when comparing against the pivot in quicksort.
+#define SORT_COMP_CHECK_EQ PREFIXNAME(SortCompCheckEqObj)
+Int SORT_COMP_CHECK_EQ(SORT_FUNC_ARGS, Obj a, Obj b) {
```
fingolfin: Should this perhaps be marked static and/or inline?

Issue #639, opened 2016-02-25
Obsoleting gapmacro.tex


olexandr-konovalov:
I compiled a list of packages still not using GAPDoc. 

Among 130 packages that will appear in GAP 4.8.2, there are 33 packages which do not have GAPDoc-based manuals:
- [ ] [ace](https://github.com/gap-packages/ace) - see https://github.com/gap-packages/ace/issues/1
- [ ] [aclib](https://github.com/gap-packages/aclib)
- [ ] [alnuth](https://github.com/gap-packages/alnuth) - see https://github.com/gap-packages/alnuth/issues/5
- [ ] [automgrp](https://github.com/gap-packages/automgrp)
- [ ] [autpgrp](https://github.com/gap-packages/autpgrp)
- [ ] [CaratInterface](https://github.com/gap-packages/CaratInterface)
- [ ] [cohomolo](https://github.com/gap-packages/cohomolo) - see https://github.com/gap-packages/cohomolo/issues/1
- [ ] [crisp](https://github.com/bh11/crisp)
- [ ] [cryst](https://github.com/gap-packages/cryst)
- [ ] [crystcat](https://github.com/gap-packages/crystcat)
- [ ] [cubefree](https://github.com/gap-packages/cubefree)
- [ ] [design](https://github.com/gap-packages/design/)
- [ ] [fga](https://github.com/chsievers/fga)
- [ ] [format](https://github.com/gap-packages/format)
- [x] [fplsa](https://github.com/gap-packages/FPLSA) - done by @fingolfin in https://github.com/gap-packages/FPLSA/issues/2
- [ ] [fwtree](https://github.com/gap-packages/fwtree)
- [ ] [grape](https://github.com/gap-packages/grape/)
- [ ] [grpconst](https://github.com/gap-packages/grpconst)
- [ ] [irredsol](https://github.com/bh11/irredsol)
- [ ] [itc](https://github.com/gap-packages/itc) - see https://github.com/gap-packages/itc/issues/1
- [x] [kbmag](https://github.com/gap-packages/kbmag) - done by @cdwensley in https://github.com/gap-packages/kbmag/issues/1
- [ ] [liepring](https://github.com/gap-packages/liepring)
- [ ] [matgrp](https://github.com/hulpke/matgrp)
- [ ] [modisom](https://github.com/gap-packages/modisom) - see https://github.com/gap-packages/modisom/issues/9
[...]

PR #2622, opened 2018-07-04
Add gapmacro-based mockpkg manual to test helpt2t.gi

alex-konovalov:
This PR comes from the discussion in #639. It adds a simple gapmacro-based manual to the mock package in `tst/fakepkg` to be able to test the code from `helpt2t.g{d,i}` which converts documentation written in `gapmacro.tex` format into text for the "screen" online help viewer. 

That file is more than 1000 lines long, and at the moment has zero test coverage, so there is a risk of overlooking an accidental change that will break this functionality.

Testing and maintaining `helpt2t.g{d,i}` is consistent with stating that `gapmacro.tex` is not supported since GAP 4.5. We are not testing an ability to *build* new manuals in `gapmacro.tex` format - instead, we are testing backwards compatibility, i.e. the ability to *use* `gapmacro.tex`-based manuals for existing packages which did not switch to GAPDoc. 

Do not merge this yet - I will see after code coverage reports if more text should be added to the manual to exercise more special cases from  `helpt2t.gi`. 

P.S. for the test itself, one does not have to have TeX and run the `make_doc` script. This can be run by the test maintainer, who will then commit produced files to the repository.

P.P.S. This PR also renames `fakepkg` to `mockpkg`.


tst/fakepkg/doc/make_doc
```diff
@@ -0,0 +1,16 @@
...
+
+echo "TeXing documentation"
+# TeX the manual 
+tex manual
+# TeX the manual again to incorporate the ToC ... and build the index
+tex manual; ../../../doc/manualindex manual
+# Finally TeX the manual again to get cross-references right
+tex manual
+# Create PDF version
+pdftex manual; pdftex manual
```
fingolfin: Is that PDF version used for anything? I'd much rather *not* add that manual.pdf to the repository, as it is a binary blob.

tst/fakepkg/doc/make_doc
```diff
@@ -0,0 +1,16 @@
...
+
+#########################################################################
+##
+#E
```
fingolfin: Drop these final comment lines?

tst/teststandard/helptools.tst
```diff
@@ -16,4 +16,27 @@ gap> HELP_SEARCH_ALTERNATIVES("setismapping");
 [ "hasismapping", "ismapping", "setismapping" ]
 gap> HELP_SEARCH_ALTERNATIVES("ismapping");
 [ "ismapping" ]
+
+# Testing the code from `lib/helpt2t.g{d,i}` which converts TeX source code
+# in `gapmacro.tex` format into text for the "screen" online help viewer.
+gap> HELP_ADD_BOOK("fakepkg","fakepkg",DirectoriesLibrary("tst/fakepkg/doc")[1]);
+gap> l:=HELP_PRINT_SECTION_TEXT(HELP_BOOK_INFO("fakepkg"),1,1).lines;;
```
fingolfin: This fails in HPC-GAP due to a missing region lock

tst/fakepkg/doc/make_doc
alex-konovalov: It's mentioned in PackageInfo.g and its used for its validation. But it's fine to not to commit the PDF produced by `make_doc` (will describe that in some README), since it is not needed for _this_ test.

fingolfin: I would recommend this then: remove the `pdftex` calls here, and replace `manual.pdf` by a plain text file saying something like
> This is a placeholder, to enable validation of this mock package, while avoiding the need to add a binary file to the repository.

fingolfin: Or, if you really want to keep these two lines (but why?), then add the `manual.pdf` to a `.gitignore` file, to prevent people from adding it accidentally in the future.

Note that you can have a local `.gitignore` file inside `tst/fakepkg/` or even `tst/fakepkg/doc/`

alex-konovalov: I like the idea of plain text `manual.pdf` then. I don't really need to produce pdf - I can judge whether the manual is produced from the dvi file.

alex-konovalov:
Oops - the `Test` function trips over
```
This is a command to document a function:

> FunctionName( <arg1>, <arg2> [: <options>] )                                                                 F
```
because it interprets this as a partial prompt. I think I will have to prepend each printed line by some symbol to avoid it being interpreted as a GAP input.

alex-konovalov:
I also want to rename "fakepkg" to "mockpkg" - maybe on a separate PR later, or on a separate commit in this PR.

alex-konovalov:
https://codecov.io/gh/gap-system/gap/pull/2622/changes already shows 55% coverage of `helpt2t.gi` - will add more text now to exercise it even more. The attempt to fix HPC-GAP tests did not work.

alex-konovalov:
Hope 2nd attempt will fix the test for HPC-GAP. Read-only access is not enough, the test passes with full access.
```
gap> HELP_ADD_BOOK("fakepkg","fakepkg",DirectoriesLibrary("tst/fakepkg/doc")[1]);
gap> l:=HELP_PRINT_SECTION_TEXT(HELP_BOOK_INFO("fakepkg"),1,1).lines;;
Error, Attempt to read object 4577071536 of type list (string) without having read ac\
cess in
  if w - 7 < Length( l ) then
    l := Concatenation( l{[ 1 .. w - 7 ]}, "..." );
fi; at /Users/alexk/GITREPS/gap/hpcgap/lib/helpbase.gi:427 called from 
FILLED_LINE( info.sections[chapter][section], info.chapters[chapter], '_' 
 ) at /Users/alexk/GITREPS/gap/lib/helpt2t.gi:237 called from
<function "HELP_PRINT_SECTION_TEXT">( <arguments> )
 called from read-eval loop at stream:2
brk>
```

tst/fakepkg/README.md
```diff
@@ -1,3 +1,20 @@
...
+related to GAP packages, for example to validate `PackageInfo.g`
+files.
+
+It also has a manual written in `gapmacro.tex` format. This
+format is obsolete, being superseded by GAPDoc; however, it
+is still used by some packages, and hence we would like to
+ensure that we are still able to render their documentation
+using the "screen" online help viewer. This is why we use 
+this manual to test for the code from `lib/helpt2t.g{d,i}`
+which converts `gapmacro.tex`-based manuals into plain text.
```
fingolfin: BTW, we could *also* add a GAPDoc manual to the mock package -- a package can have more than one help book, after all. (This is not a request for this PR, just wanted to point this out)

alex-konovalov:
This looks OK now - I only want now to rename `fakepkg` to `mockpkg` in another commit.

alex-konovalov:
Now renamed, grep for `fake` in `tst` directory returns nothing. Should be ready.

PR #2623, opened 2018-07-05
Travis: treat compiler warnings as errors

fingolfin:
I think it's very useful to strive for zero compiler warnings, at least on "supported" configurations, and so far, we are doing a pretty good job on this. But sometimes warnings slip through; e.g. on Linux systems, gcc issues warnings if some variables in a function using `longjmp` should be made `volatile`. But since I normally only compile on macOS, I tend to not see them; so for me, it's helpful if Travis reports these warnings.

Also add some comments, and make `--enable-debug --enable-Werror` the default configure flags, which are turned off by two jobs, instead of the other way around, as it was before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ env:
   global:
     - CFLAGS="--coverage -O2"
     - LDFLAGS="--coverage"
+    # default config flags: enable debug asserts, and treat compiler warnings as errors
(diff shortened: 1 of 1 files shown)


fingolfin:
Yeah, well, if people put in hacks to disable warnings, instead of properly understanding and resolving them, that's bad either way. But no way stopping them from doing that, other than trying to catch this in code review.

PR #2625, opened 2018-07-05
Rename tst/test-error to tst/testspecial; and split the `testspecial` travis test

fingolfin:
`test-error` now contains tests which do not deal with errors. I just suggested to @alex-konovalov to add another test there which is not related to errors, namely a test of the help system. 

Since the Travis "test suite" which executed test-error and test-compile together was named `testspecial`, I renamed `test-error` to `testspecial`, and then split the Travis test suite into two: one which just does `testspecial`, and one which does just `test-compile`. Both suites are still executed by the same Travis job, so nothing should really change, except that it is now hopefully a little bit clearer what is what and how things fit together. In particular, Travis "test suites" now corresponds 1-1 to subdirs of `tst/` again/


.travis.yml
```diff
@@ -99,7 +99,7 @@ matrix:
     - env: TEST_SUITES=testbugfix CONFIGFLAGS="--enable-debug"
 
     # test error reporting and compiling (quickest job in this test suite)
-    - env: TEST_SUITES=testspecial CONFIGFLAGS="--enable-debug"
+    - env: TEST_SUITES="testspecial test-compile" CONFIGFLAGS="--enable-debug"
```
alex-konovalov: Hmm, am I missing something? The description says "split the Travis test suite into two: one which just does testspecial, and one which does just test-compile" but it runs them in a single job.

fingolfin:
It splits the test suite, but not the job.

.travis.yml
fingolfin: test suite != test job. The test suite was split into two; but both are still run by the same job (this is also explained in the commit message).

fingolfin:
Rebased to resolve a conflict in `.travis.yml`

alex-konovalov:
LGTM. I presume that can be done via "rebase and merge"?

PR #2626, opened 2018-07-05
refactor access to function bodies

fingolfin:
Add READ_{EXPR,STAT}, WRITE_{EXPR,STAT}, CONST_ADDR_{EXPR,STAT} helpers to improve over code readability, and to make code which only reads from the body immediately distinguishable from code which writes to it.

With this change, it becomes trivial to see that code.c is the only place we ever write into function bodies.

This enables future refactoring of the coder; e.g. we can stop using `PtrBody` in it, which is a useful optimization for the code executor, but less so for generating code.

@markuspf I am not sure if this has any potential overlap with your syntax tree work; since I don't want to sabotage that, please let me know if this is the case, and we should put this PR here on hold.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -1579,7 +1579,7 @@ void CodeCharExpr (
 
     /* allocate the character expression                                   */
     litr = NewExpr( T_CHAR_EXPR, sizeof(UChar) );
-    ((UChar*)ADDR_EXPR(litr))[0] = chr;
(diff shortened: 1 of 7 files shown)


fingolfin:
I just checked, and this does not conflict with PR #2628 at all (though once this is merged, the syntax tree code should be adjusted to use `READ_STAT` and `READ_EXPR`, too (which is easy, a matter of running a regex search&replace)

PR #2627, opened 2018-07-05
kernel: cleanup and split system.h

fingolfin:
Right now, system.h has two roles: on the one hand, it serves as a general purpose header which basically is included from *everywhere* in order to get some common system includes; some central typedefs (like for `UInt`, `Int4` etc.), etc.

OTOH, it also serves its original purpose, which is to expose some operating system APIs in a semi-portable fashion to the rest of GAP.

And then it also contains variables for storing command line option setting!

I'd like to resolve this eventually. This PR splits out the CLI options into a separate header, and removes some other stuff.

A future PR then can split the other two functions: My suggestion is to add a new header, say `common.h` (better names welcome) which becomes the one included by every other GAP file, and which defines types, includes some system headers etc.; and leave the "OS interface" stuff in `system.h`. And some typedef might also be moved into yet other places (be it existing or new header files)

Any feedback on this plan is welcome (including better suggestions for `common.h`).


src/system.c
```diff
@@ -1118,19 +1119,23 @@ void InitSystem (
...
-#else
+  #else
     SyAllocPool = 1536L*1024*1024;   /* Note this is in bytes! */
-#endif
+  #endif
 #endif
     SyStorOverrun = 0;
     SyStorKill = 0;
-    SyStorMin = SY_STOR_MIN;         /* see system.h */
+#if SIZEOF_VOID_P == 8
```
ChrisJefferson: Is there supposed to be some code here?

ChrisJefferson:
This is fine, except for one small pointless (I think) piece of code

src/system.c
fingolfin: Ouch, that originally contained code to set `SyStorMin`, before I realized that it makes more sense to move that next to where we set `SyStorMax`.  Thanks for checking closely, a fixed version will be up in a few seconds.

PR #1811, opened 2017-10-26
Pragmas

markuspf:
This PR introduces "pragmas" into the GAP language using the following syntax

```
f := function(x)
   local y;
   #@ y : Int
end;
```

Pragmas are currently stored as strings in the statement sequence (which means they can currently only usefully appear in function bodies, after `local`, its a minor change to be able to add them before `local` as well).

The idea of pragmas is that we get a (even backwards compatible) way to annotate code with documentation, typing hints, or other things (like code for @frankluebeck's new Demo code).

Together with the syntax-tree module this could get us closer to not having to scrape GAP code manually for packages like `AutoDoc`, but currently we could introduce a function `GetPragmas(func)` to get the pragmas that are in the function object as a list of strings.

We possibly want a mechanism to have namespaces for pragmas, so that packages can have their own names without clashing.

This PR serves the purpose to discuss 
 * Whether pragmas are considered useful, or whether we should come up with different approaches for the usecases we can come up with.
 * Whether there are objections

PR #2628, opened 2018-07-05
Syntax Trees

markuspf:
*Note* the incomplete code for compiling syntax trees has been removed, and will be added (completed) in a later PR.

This PR adds a function to convert a GAP function into a GAP-level syntax tree, it also contains some (untested, possibly still incomplete) code to compile such a syntax tree into a "cleaned up" syntax tree, and to pretty print functions.

Once this is merged the next step is to add a function that compiles a syntax tree back into GAP's internal binary representation. I have prototype code for such a function but it doesn't work reliably yet.

Note that this code predates quite a few changes made to the internal representation of GAP code. This PR contains a test that compiles all globally findable functions into a syntax tree, but this test passing of course only means that the syntax tree code does not crash: As soon as I have the coding function working, we can test rountrips (and compare the binary representations of functions).

TODO:

 - [x] Check for completeness (Floats are not supported at the moment)
 - [x] Check that code for `LVars`,  `HVars` is still valid after @fingolfin's changes 
 - [x] Possibly other things.

### DIFF
--- a/lib/syntaxtree.gi
+++ b/lib/syntaxtree.gi
@@ -1,0 +1,10 @@
+# syntaxtree.gi
+
+InstallGlobalFunction( SyntaxTree,
+function(func)
+    return Objectify( SyntaxTreeType, rec( tree := SYNTAX_TREE(func) ) );
+end);
+
+InstallMethod( ViewString, "for a syntax tree"
+               , [ IsSyntaxTree ]
+               , t -> "<syntax tree>" );
--- a/src/syntaxtree.c
+++ b/src/syntaxtree.c
@@ -1,0 +1,810 @@
+//
+//  syntaxtree.c
+//
+#include "syntaxtree.h"
+
+#include "bool.h"
+#include "calls.h"
+#include "code.h"
+#include "error.h"
+#include "exprs.h"
+#include "gvars.h"
+#include "integer.h"
+#include "lists.h"
+#include "modules.h"
+#include "opers.h"
+#include "plist.h"
+#include "precord.h"
+#include "records.h"
+#include "stats.h"
+#include "stringobj.h"
+#include "system.h"
+#include "vars.h"
+
+#include <stdarg.h>
+#include <ctype.h>
+
+typedef Obj (*CompileFuncT)(Obj node, Expr expr);
+typedef Obj (*CompileArgT)(Expr expr);
+
+typedef struct {
+    const Char * argname;
+    CompileArgT  argcomp;
+} ArgT;
+
+typedef struct {
+    UInt         tnum;
+    CompileFuncT compile;
+    const Char * name;
+    UInt         arity;
+    ArgT         args[4]; /* The maximum compiler arity is currently 4 */
+} CompilerT;
+
+// We put compilers for statements and expressions into the same static array,
+// assuming that the set of their respective ids are disjoint.
+static const CompilerT Compilers[];
+#define COMPILER_ARITY(...) (sizeof((ArgT[]){ __VA_ARGS__ }) / sizeof(ArgT))
+#define COMPILER(tnum, compiler, ...)                                        \
+    [tnum] = {                                                               \
+        tnum, compiler, #tnum, COMPILER_ARITY(__VA_ARGS__), { __VA_ARGS__ }  \
+    }
+
+#define COMPILER_(tnum, ...)                                                 \
+    COMPILER(tnum, SyntaxTreeDefaultCompiler, __VA_ARGS__)
+
+#define ARG(name, func)                                                      \
+    {                                                                        \
+        name, func                                                           \
+    }
+#define ARG_(name) ARG(name, SyntaxTreeCompiler)
+
+static Obj SyntaxTreeFunc(Obj result, Obj func);
+
+static inline Obj NewSyntaxTreeNode(const char * type)
+{
+    Obj result;
+    Obj typestr;
+
+    typestr = MakeImmString(type);
+    result = NEW_PREC(2);
+    AssPRec(result, RNamName("type"), typestr);
+
+    return result;
+}
+
+static Obj SyntaxTreeCompiler(Expr expr)
+{
+    Obj       result;
+    UInt      tnum;
+    CompilerT comp;
+
+    // TODO: GAP_ASSERT
+    tnum = TNUM_EXPR(expr);
+    comp = Compilers[tnum];
+
+    result = NewSyntaxTreeNode(comp.name);
+
+    comp.compile(result, expr);
+
+    return result;
+}
+
+static Obj SyntaxTreeArgcompInt(UInt i)
+{
+    return INTOBJ_INT(i);
+}
+
+static Obj SyntaxTreeRefLVar(Obj result, Expr expr)
+{
+    return INTOBJ_INT(LVAR_REFLVAR(expr));
+}
+
+static Obj SyntaxTreeDefaultCompiler(Obj result, Expr expr)
+{
+    int       i;
+    UInt      tnum;
+    CompilerT comp;
+
+    // TODO: GAP_ASSERT tnum range
+    tnum = TNUM_EXPR(expr);
+    comp = Compilers[tnum];
+
+    for (i = 0; i < comp.arity; i++) {
+        AssPRec(result, RNamName(comp.args[i].argname),
+                comp.args[i].argcomp(READ_EXPR(expr, i)));
+    }
+    return result;
+}
+
+static Obj SyntaxTreeFunccall(Obj result, Expr expr)
+{
+    Obj  func;
+    Obj  args, argi;
+    UInt narg, i;
+
+    func = SyntaxTreeCompiler(FUNC_CALL(expr));
+    AssPRec(result, RNamName("funcref"), func);
+
+    /* compile the argument expressions */
+    narg = NARG_SIZE_CALL(SIZE_EXPR(expr));
+    args = NEW_PLIST(T_PLIST, narg);
+    SET_LEN_PLIST(args, narg);
+
+    for (i = 1; i <= narg; i++) {
+        argi = SyntaxTreeCompiler(ARGI_CALL(expr, i));
+        SET_ELM_PLIST(args, i, argi);
+        CHANGED_BAG(args);
+    }
+    AssPRec(result, RNamName("args"), args);
+    return result;
+}
+
+static Obj SyntaxTreeFuncExpr(Obj result, Expr expr)
+{
+    Obj fexs;
+    Obj fexp;
+
+    fexs = FEXS_FUNC(CURR_FUNC());
+    fexp = ELM_PLIST(fexs, READ_EXPR(expr, 0));
+
+    SyntaxTreeFunc(result, fexp);
+
+    return result;
+}
+
+static Obj SyntaxTreeIntExpr(Obj result, Expr expr)
+{
+    Obj  value;
+    UInt size;
+
+    if (IS_INTEXPR(expr))
+        value = OBJ_INTEXPR(expr);
+    else {
+        size = SIZE_EXPR(expr) / sizeof(UInt) - 1;
+        value = MakeObjInt(CONST_ADDR_EXPR(expr) + 1, size);
+    }
+
+    AssPRec(result, RNamName("value"), value);
+
+    return result;
+}
+
+static Obj SyntaxTreeCharExpr(Obj result, Expr expr)
+{
+    AssPRec(result, RNamName("value"), ObjsChar[(UChar)READ_EXPR(expr, 0)]);
+    return result;
+}
+
+static Obj SyntaxTreePermExpr(Obj result, Expr expr)
+{
+    Obj  cycles;
+    Obj  cycle;
+    Obj  val;
+    Expr cycleexpr;
+    Int  csize, n;
+    Int  i, j;
+
+    /* determine number of cycles */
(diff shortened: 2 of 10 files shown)


fingolfin:
Huh, what LVars/HVars changes did I make that could affect this? I am not aware of having changed anything related to that, other than inside the reader (which does not affect the internal representation). Am I forgetting something?

markuspf:
I think I was thinking of the interpreter code, which is of course not relevant for this PR.

I am getting a bit irritated by the ward failure now...

fingolfin:
Now that ward has been updated, I restarted the HPC-GAP Travis job for this PR, let's see if it passes now.

ChrisJefferson:
I know in the past you mentioned translating all functions (was that global functions?). While the result wouldn't be useful, perhaps add a test which tries to translate all functions, just so we know the code at least doesn't crash / get broken to where it crashes?

src/syntaxtree.c
```diff
@@ -0,0 +1,907 @@
...
+              ARG_("comobj"), ARG("name", SyntaxTreeRNam)),
+    COMPILER_(T_ISB_COMOBJ_EXPR,
+              ARG_("comobj"), ARG_("expression")),
+};
+
+Obj FuncSYNTAX_TREE(Obj self, Obj func)
+{
+    Obj result;
+
+    if (!IS_FUNC(func)) {
```
fingolfin: You can add `|| !IsKernelFunction(func)` to also catch non-GAP-functions.

src/syntaxtree.c
```diff
@@ -0,0 +1,907 @@
...
+/*
+
+Obj SyntaxTreeRefHVar(Obj result, Expr expr)
+{
+    AssPRec(result, RNamName("hvar"), SyntaxTreeHVar(expr));
+    return result;
+}
+*/
+
+Obj SyntaxTreeRefGVar(Obj result, Expr expr)
```
fingolfin: This function is unused

lib/syntaxtree.gi
```diff
@@ -0,0 +1,831 @@
...
+    compiler.T_QUO := compile_record;
+    compiler.T_MOD := compile_record;
+    compiler.T_POW := compile_record;
+
+    # These come from literals
+    # TODO: Maybe make the component names uniform
+    compiler.T_TRUE_EXPR := expr -> expr;
+    compiler.T_FALSE_EXPR := expr -> expr;
+    compiler.T_INTEXPR := expr -> expr;
+    compiler.T_INT_EXPR := expr -> expr;
```
fingolfin: Off-topic, but: we probably should rename `T_INTEXPR` and `T_INT_EXPR`: one is for small/immediate integers, the other for large/big integers. Right now, they are super easy to confuse, and no way to tell which is which.

More off topic: we refer to "large integers" in some places, but it seems the rest of the world calls them "big integers" or "bigints", so perhaps we should also switch to that terminology.

lib/syntaxtree.gi
```diff
@@ -0,0 +1,831 @@
...
+    compiler.T_ASS2_LIST := compile_record;
+    compiler.T_ASSX_LIST := compile_record;
+
+    return Objectify(GAPCompilerType, rec( name := "CleanupCompiler"
+                                         , compiler := compiler) );
+end);
+
+# Pretty printer based on Wadler's Prettier Printer.
+# It is hugely inefficient I guess, but fun.
+# Also this only works on a "cleaned" syntax tree (probably flag this somehow)
```
fingolfin: It'd be really nice to have an example for this in a .tst or as a GAPDoc example, or even in a plain source comment here :-).

src/syntaxtree.c
```diff
@@ -0,0 +1,907 @@
...
+        else {
+            /* TODO: Probably put the number in */
+            C_NEW_STRING_DYN(str, "localvar");
+        }
+        SET_ELM_PLIST(locnams, i, str);
+        CHANGED_BAG(locnams);
+    }
+
+    /* switch to this function (so that 'ADDR_STAT' and 'ADDR_EXPR' work) */
+    SWITCH_TO_NEW_LVARS(func, narg, nloc, oldFrame);
```
fingolfin: My suggestion (and one on which I'd be happy to work on, if people agree with the general idea) would be to *not* use `ADDR_STAT` nor `ADDR_EXPR` here. Instead, pass the `T_BODY` object as an additional argument `body` to each syntax tree callback; then replace `ADDR_STAT(stat)[i]` by `ADDR_OBJ(body)[stat+i]` -- and of course one can use helper functions or macros to make this nicer, and e.g. replace `ADDR_STAT(stat)[i]` by, say `ReadStat(body, stat, i)` with
```C
static inline Stat ReadStat(Obj body, Stat stat, UInt i)
{
   return ADDR_OBJ(body)[stat+i];
}
```

Some more work is needed (e.g. for `ARGI_INFO`), so it's not a quick change, but if one introduces the above `ReadStat` helper, it could easily be done incrementally. 

Added bonus: an error during syntax tree compilation does not risk leaving `STATE(PtrBody)` in a bad state.

markuspf:
I rebased and cut down the amount of code added (removing incomplete example compilers and other fancy technology).

I did add floats.

Once this is merged I will first make a PR that adds a function that takes a syntax tree that is produced by `SYNTAX_TREE` and turns it back into a callable GAP function.

I have some example code that shows how I imagine writing "compilers" for GAP code, for example a "Pretty Printer" which I will complete and submit as a PR.
We could do fairly simple constant folding, certain amounts of inlining (possibly with the help of #1811), and other fun stuff with this technology.

I'd like to incrementally work on this now, for example one can also consider moving from that slightly horrible macro trap to C++ templates (though I have not carefully thought about this).

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+
+    return result;
+}
+
+Obj SyntaxTreeCharExpr(Obj result, Expr expr)
+{
+    Obj value;
+
+    value = NewBag(T_CHAR, 1);
+    *(UChar*)ADDR_OBJ(value) = (UChar)ADDR_EXPR(expr)[0];
```
fingolfin: Those two lines are wrong, I am afraid -- T_CHAR instances are singletons, so you should use `ObjsChar[i]` instead; see also the code of `EvalCharExpr`.

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+    locnams = NEW_PLIST(T_PLIST, nloc);
+    SET_LEN_PLIST(locnams, nloc);
+    AssPRec(result, RNamName("locnams"), locnams);
+    for (i = 1; i <= nloc; i++) {
+        if (NAMI_FUNC(func, narg + i) != 0) {
+            str = CopyToStringRep(NAMI_FUNC(func, narg + i));
+        }
+        else {
+            /* TODO: Probably put the number in */
+            C_NEW_STRING_DYN(str, "localvar");
```
fingolfin:
```suggestion
            str = MakeImmString("localvar");
```

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+    Obj  subrec;
+    Expr tmp;
+    Int  i, len;
+
+    len = SIZE_EXPR(expr) / (2 * sizeof(Expr));
+    list = NEW_PLIST(T_PLIST, len);
+    SET_LEN_PLIST(list, len);
+
+    for (i = 1; i <= len; i++) {
+        // Check whether this can happen and *why*
```
fingolfin:
```suggestion
        // TODO: check whether this can happen and *why*
```

fingolfin: I am not quite sure what this comment refers to, though. Perhaps the `tmp == 0` case? Then perhaps it should be moved into that, to make it clearer?

But then, looking at the coder, and also at `RecExpr2`, I am pretty sure `tmp == 0` can never happen, so I'd replaced this by `GAP_ASSERT(tmp != 0);` and be done with it.

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+    }
+
+    return INTOBJ_INT(lvar);
+}
+
+Obj SyntaxTreeHVar(Expr expr)
+{
+    HVar hvar;
+
+    hvar = (HVar)(expr);
```
fingolfin: Huh -- shouldn't this rather be `hvar = *ADDR_EXPR(expr);`?

EDIT: Oh, I get it now: this function is only used as an entry for an `argcomp` field, which already performed the ADDR_EXPR for us.

I think this would be much clearer if the function names reflected it. E.g. `SyntaxTreeArgCompHVar`, or just  `static ArgCompHVar` or so.

And similarly, the "compiler" functions could have a separate name. That would make a lot of things clearer to me.

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+    return INTOBJ_INT(hvar);
+}
+
+Obj SyntaxTreeGVar(Expr expr)
+{
+  return NameGVar((GVar)(expr));
```
fingolfin: And here I'd also expect `*ADDR_EXPR(expr)`. Yet apparently your code works -- what am I missing?

EDIT: ah, see above: this is an `argcomp` entry

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+}
+
+Obj SyntaxTreeRefLVar(Obj result, Expr expr)
+{
+    AssPRec(result, RNamName("lvar"), SyntaxTreeLVar(expr));
```
fingolfin: Why is `SyntaxTreeLVar` a separate function that is only called here, and why are gvars and hvars not put into records?

src/syntaxtree.c
fingolfin: Still is unused

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+    const Char * argname;
+    CompileArgT  argcomp;
+} ArgT;
+
+typedef struct {
+    UInt         tnum;
+    CompileFuncT compile;
+    const Char * name;
+    UInt         arity;
+    ArgT         args[6]; /* This is not really a restriction */
```
fingolfin: I don't quite understand this comment. Do you mean to express that you actually allow instances to have more than 6 entries here? But then why not use a flexible array member, i.e. `ArgT args[];`, then this is self-explanatory and unambiguous.

src/syntaxtree.c
```diff
@@ -0,0 +1,898 @@
...
+
+static Int InitKernel(StructInitInfo * module)
+{
+    /* init filters and functions */
+    InitHdlrFuncsFromTable(GVarFuncs);
+
+    return 0;
+}
+
+static Int PostRestore(StructInitInfo * module)
```
fingolfin: Does nothing, remove?

src/syntaxtree.c
markuspf: The storage size of args has to be known at compile time if I want the compilers to be statically setup (otherwise I get `error: initialization of flexible array member in a nested context`).

The arity of compilers has a statically known upper bound though (it's currently 4 if I am not mistaken).

I admit that the comment should be clearer though.

src/syntaxtree.h
```diff
@@ -0,0 +1,13 @@
+//
+//  syntaxtree.h
+//
+
+
+#ifndef GAP_SYNTAXTREE_H
+#define GAP_SYNTAXTREE_H
+
+#include "objects.h"
```
fingolfin:
```suggestion
#include "system.h"
```

src/syntaxtree.c
```diff
@@ -0,0 +1,850 @@
...
+              ARG_("comobj"), ARG_("expression")),
+};
+
+Obj FuncSYNTAX_TREE(Obj self, Obj func)
+{
+    Obj result;
+
+    if (!IS_FUNC(func) || IsKernelFunction(func) || IS_OPERATION(func)) {
+        ErrorQuit("SYNTAX_TREE: <function> must be a plain GAP function",
+                  (Int)TNAM_OBJ(func), 0L);
```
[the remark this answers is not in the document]
fingolfin: This change would also require an update to `tst/testinstall/syntaxtree.tst` afterwards.

src/syntaxtree.c
```diff
@@ -0,0 +1,850 @@
...
+
+static Obj SyntaxTreeRefLVar(Obj result, Expr expr)
+{
+    LVar lvar;
+
+    if (IS_REFLVAR(expr)) {
+        lvar = LVAR_REFLVAR(expr);
+    } else {
+        lvar = (LVar)(READ_EXPR(expr,0));
+    }
```
fingolfin: OK, I see you have folded `SyntaxTreeLVar` in here now. But I am still confused as to why you handle `T_REFLVAR` and `T_ISB_LVAR` via `SyntaxTreeRefLVar`, but then also have
```C
    COMPILER_(T_ASS_LVAR,
               ARG("lvar", SyntaxTreeIntObjInt), ARG_("rhs")),
     COMPILER_(T_UNB_LVAR, ARG("lvar", SyntaxTreeIntObjInt)),
```
I would expect all three to be handled uniformly, and the above code put into a `SyntaxTreeArgcompLVar` helper (which would probably be identical to `SyntaxTreeIntObjInt`)

fingolfin: Such code would also break the executor, as in `vars.c`, there is no such check. This seems to be a "bug" in the compiler code (which was written later)

markuspf: @fingolfin figured out that this code used to handle references to `LVar`s before they were changed in 8d2ca938c2f4102208b72e21a87d723f9d8446e0

Issue #2620, opened 2018-07-04
Incorrect delegation String -> DisplayString


alex-konovalov:
https://www.gap-system.org/Manuals/doc/ref/chap6.html#X8074A8387C9DB9A8 sets delegations for output operations. A method to output an object may only delegate to another operation which appears further right in the following list: Display, ViewObj, PrintObj, DisplayString, ViewString, PrintString, String. 

`CheckOutputDelegations` in `tst/testutil.g` checks this (not a part of an automated test though). It now works in GAP 4.9.1, where it detects this incorrect delegation:
```
---------------------------------------------------------------
String: for a function, with whitespace reduced
 delegates to [ "DisplayString" ]
Filename: /usr/local/Cellar/gap/4.9.1/libexec/lib/function.gi, line : 75
function ( fun )
    local s, str;
    s := ShallowCopy( DisplayString( fun ) );
    Remove( s );
    NormalizeWhitespace( s );
    return MakeImmutable( s );
end
---------------------------------------------------------------
```
This delegation is also present in the master branch:
```
Detected incorrect delegations for String
---------------------------------------------------------------
String: for a (finite) list
 delegates to [ "DisplayString" ]
Filename: /Users/alexk/GITREPS/gap/lib/function.gi, line : 75
function ( fun )
    local s, str;
    s := ShallowCopy( DisplayString( fun ) );
    Remove( s );
    NormalizeWhitespace( s );
    return MakeImmutable( s );
end
---------------------------------------------------------------
```
Also it says `String: for a (finite) list` instead of `String: for a function, with whitespace reduced` - why?

fingolfin:
The incorrect delegation was introduced in PR #1591 resp. commit c78bb4c4d89a8c17bb931675df6d39a87b9e32d3 by @stevelinton 

As to the incorrect info string printed by `CheckOutputDelegations` on master vs. stable-4.9: that may be another regression in `CheckOutputDelegations`, i.e., a separate issue should be filed (but I'll look into it right now)

PR #2632, opened 2018-07-06
Add MethodsOperation

fingolfin:
This is a high-level wrapper for METHODS_OPERATION; it has some overhead, as
it creates a fresh list of records for each method when called; but it's
output has the advantage of being easier to process, leading to much easier
to understand code; and at the same time, it's future proof against
modifications to the way we store methods internally.

This would then also be useful for e.g. the browse package.

Also added a test for ApplicableMethod.

This PR also fixes the bad output of `CheckOutputDelegations` in master described in issue #2620.

Open question: should this be documented for users or not?

### DIFF
--- a/lib/oper.g
+++ b/lib/oper.g
@@ -1931,6 +1931,36 @@ end);
 
 fi;
 
+if BASE_SIZE_METHODS_OPER_ENTRY <> 5 then
+    Error("MethodsOperation must be updated for new BASE_SIZE_METHODS_OPER_ENTRY");
+fi;
+
+# TODO: document this?!
+BIND_GLOBAL("MethodsOperation", function(oper, nargs)
+    local meths, len, result, i, m;
+
+    meths := METHODS_OPERATION(oper, nargs);
+    if meths = fail then
+        return fail;
+    fi;
+    len := BASE_SIZE_METHODS_OPER_ENTRY + nargs;
+    result := [];
+    for i in [0, len .. LENGTH(meths) - len] do
+        m := rec(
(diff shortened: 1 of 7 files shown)


fingolfin:
Having this being used in Browse was one of the things I was hoping for. Even if we don't "officially" document this, it should be a much a more stable interface to accessing method operations.

Anyway, the first version was buggy and only worked for 1-arg methods *ouch*. I fixed this now, and added more dedicated test cases to check 0 to 6 arguments. Let's see if the tests pass now.

lib/oper.g
```diff
@@ -1931,6 +1931,36 @@ end);
...
+    meths := METHODS_OPERATION(oper, nargs);
+    if meths = fail then
+        return fail;
+    fi;
+    len := BASE_SIZE_METHODS_OPER_ENTRY + nargs;
+    result := [];
+    for i in [0, len .. LENGTH(meths) - len] do
+        m := rec(
+            famPred := meths[i + 1],
+            argFilt := meths{[i + 2 .. i + nargs + 1]},
```
fingolfin: Anybody want to nitpick the record names chosen here? E.g. `argFilt` could also be `filters` or `argumentFilters`. In the documentation for `InstallMethod`, it is called `args-filts` (which isn't a valid identifier, though).  And `famPred` is `famp`, `rank` is `val`, and `func` is `method`.

Otherwise, I'd say we can merge this, no?

ThomasBreuer: Yes, I think this should be merged.

PR #2635, opened 2018-07-10
Fix zlib on windows

ChrisJefferson:
This fixes builtin zlib on windows. It does make building a bit less neat, as the win32 Makefile for zlib will only work inside the zlib source directory.


ChrisJefferson:
I have added a commit to check that appveyor works both with and without gmp-devel and zlib-devel, but once it passes I'd like to remove it before commit, as running that test forever more seems unnessasary.

ChrisJefferson:
woops, branch on wrong repository. sorry.

PR #2636, opened 2018-07-10
Win cyg fix

ChrisJefferson:
This fixes builtin zlib on windows. It does make building a bit less neat, as the win32 Makefile for zlib will only work inside the zlib source directory.

I have added a commit to check that appveyor works both with and without gmp-devel and zlib-devel, but once it passes I'd like to remove it before commit, as running that test forever more seems unnecessary.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,15 +2,22 @@
 clone_depth: 5                  # clone entire repository history if not defined
 
 environment:
-  CFLAGS: "--coverage"
-  LDFLAGS: "--coverage"
   TEST_SUITES: testinstall
+
+  # To test building GAP both using the bundled libraries (zlib and GMP), as
+  # well as using the versions distributed with cygwin, we do the former in
+  # the 32 bit build and the latter in the 64 bit build. But building zlib
+  # does not see to work if '--coverage' is used, so we only use that flag in
+  # the 64 bit build.
   matrix:
     - CYG_ARCH: x86
       CYG_ROOT: C:\cygwin
       ABI: 32
     - CYG_ARCH: x86_64
       CYG_ROOT: C:\cygwin64
+      PACKAGES: "-P libgmp-devel,zlib-devel"
+      CFLAGS: "--coverage"
+      LDFLAGS: "--coverage"
 # FIXME: HPC-GAP builds on AppVeyor are disabled for now, as they experience
 # weird errors in building ward (which I cannot replicate on a Windows VM).
 #    - CYG_ARCH: x86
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
The failures are being caused by #2637 (I'm not sure why that has only appeared here for the first time. Possibly because I am testing building with the internal gmp?)

.appveyor.yml
```diff
@@ -2,15 +2,24 @@
...
       CYG_ROOT: C:\cygwin
       ABI: 32
     - CYG_ARCH: x86_64
       CYG_ROOT: C:\cygwin64
+    - CYG_ARCH: x86
+      CYG_ROOT: C:\cygwin
+      ABI: 32
+      PACKAGES: "-P libgmp-devel,zlib-devel"
+      CFLAGS: "--coverage"
+      LDFLAGS: "--coverage"
```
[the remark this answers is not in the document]
ChrisJefferson: This might now be out of date. Adding --coverage to the internal zlib build seems to break it, so I only add --coverage when we use external packages

cnf/build-cygwin-zlib.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cd extern/zlib &&
+make -f win32/Makefile.gcc \
+		SHAREDLIB=cygz.dll IMPLIB=libz.dll.a \
+		SHARED_MODE=1 \
+		prefix= \
+		BINARY_PATH=../install/zlib/bin \
+		INCLUDE_PATH=../install/zlib/include \
+		LIBRARY_PATH=../install/zlib/lib \
+        all install
```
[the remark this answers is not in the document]
ChrisJefferson: I did consider just trying to grab all the .o files out of zlib. Something like miniz would be good too. To be honest, my main issue is I understand libtool even less than I understand automake/autoconf... :)

.appveyor.yml
```diff
@@ -2,15 +2,17 @@
...
+  # Test both 32-bit and 64-bit, and with and without built-in libraries
   matrix:
     - CYG_ARCH: x86
       CYG_ROOT: C:\cygwin
       ABI: 32
     - CYG_ARCH: x86_64
       CYG_ROOT: C:\cygwin64
+      PACKAGES: "-P libgmp-devel,zlib-devel"
+      CFLAGS: "--coverage"
+      LDFLAGS: "--coverage"
```
fingolfin: So... why is --coverage not used for 32bit cygwin builds anymore?

ChrisJefferson: To increase testing, 64 bit uses external libraries, 32 bit uses internal libraries. zlib doesn't seem to build when we set cflags=--coverage.

ChrisJefferson: It looked like there was no 32bit specific cygwin code we need to make sure we cover.

fingolfin: This really should be documented via comments, else this becomes unmaintainable down the road

fingolfin: I updated the comment above; feel free to squash that added commit into your other commits, or edit the message further if you are not happy with it.

ChrisJefferson:
Comment merged. Thanks. Should probably let the tests cycle, just in case I somehow broke appveyor (although I can't imagine how)

alex-konovalov:
For the record, that broke Windows release candidate tests, which I fixed later by including `/usr/bin/cygz.dll` and `/usr/x86_64-w64-mingw32/sys-root/mingw/bin/zlib1.dll` into the `-win.zip` archive.

PR #2637, opened 2018-07-10
Extend buffer size for outputting floats to 1K

ChrisJefferson:
The gcc warning 'format-truncation' warns that this buffer must be
at least 310 bytes on 32-bit cygwin. As this function is not
called recursively, just extend the buffer to 1K.

### DIFF
--- a/src/macfloat.c
+++ b/src/macfloat.c
@@ -78,7 +78,7 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
 */
 void PrintMacfloat(Obj x)
 {
-    Char buf[32];
+    Char buf[1024];
     // TODO: should we use PRINTFDIGITS instead of 16?
     PrintMacfloatToBuf(buf, sizeof(buf), VAL_MACFLOAT(x), 16);
     Pr("%s", (Int)buf, 0);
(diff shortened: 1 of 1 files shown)


src/macfloat.c
```diff
@@ -78,7 +78,7 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
 */
 void PrintMacfloat(Obj x)
 {
-    Char buf[32];
+    Char buf[1024];
```
fingolfin: This change is actually not needed, and no warning is triggered here - at least not for me.

src/macfloat.c
```diff
@@ -433,7 +433,7 @@ Obj FuncINTFLOOR_MACFLOAT( Obj self, Obj obj )
 
 Obj FuncSTRING_DIGITS_MACFLOAT( Obj self, Obj gapprec, Obj f)
 {
-  Char buf[50];
+  Char buf[1024];
   Obj str;
   int prec = INT_INTOBJ(gapprec);
   if (prec > 40) /* too much anyways, and would risk buffer overrun */
```
fingolfin: Actually, while looking at this, I realized that this code could be called with a negative precision... Anyway, the problem here is that some GCC versions are not clever enough to detect that prec <= 40 holds after this, and hence issue that spurious warning, based on the maximal possible precision. This warning is not triggered for the first call to `PrintMacfloatToBuf`, because there the precision is specified as a hard coded constant (`16`).

src/macfloat.c
ChrisJefferson: You are right, but I decided to be on the safe side I may as well make this bigger too, in case some other compiler ends up producing the same warning, and for consistency.

fingolfin: Hrm, that sounds more like voodoo to me than a valid justification... Alas, I don't really care sufficiently to argue this point ;-).

PR #2640, opened 2018-07-12
Merge some MatrixObj work into master

fingolfin:
Last year, we made some progress on MatrixObj. Sadly, we got stalled again. But I think even the partial work we did would already be useful.

I took the MatrixObj2 branch, and looked through it, and cleaned it up a lot. I also dropped (for now, this is only temporary) a few changes that I think are potentially more controversial (such as @ssiccha MultRowVector to MultVector change). Assuming this PR gets merged, I would then prepare a new branch and PR (or possibly multiple) with all the remaining work, which then can either merge, too, or continue to work on.

Some TODOs (?) before merge:
* [x] the tests in `tst/test-matobj` can be run via `tst/test-matobj.g`, but this is not currently hooked up to Travis. Is there any reason why we don't just move these tests into `testinstall` and/or (for very slow parts) into `teststandard`?
* [ ] writing a summary of the changes in here for the GAP 4.10 release notes
* [ ] running package tests against this PR, to see if it breaks any of them (maybe @alex-konovalov can help with that?)


frankluebeck:
I support to proceed as suggested here. Most changes will not influence existing GAP code and it would be good to start over from a master branch that contains most of the changes from the MatrixObj branch.

Looking through the diffs I did not notice anything problematic. Moving the files from `tst/test-matobj` to
`tst/testinstall` seems ok.

Currently, some tests of `TraceMat` fail. This could be fixed by a method in `matobjplist.gi`:
```
InstallMethod( \[\], "for a plist matrix and two positive integers",
  [ IsPlistMatrixRep, IsPosInt, IsPosInt ],
  function( m, r, c )
    return m![ROWSPOS][r][c];
  end );
```
Or should we also merge the more efficient kernel code that provides this functionality?

fingolfin:
The failing tests are a very recent regression (the previous build was fine), and I'll fix it. I also already moved the tests into `testinstall`, and added a few more.

fingolfin:
Actually, the TraceMat one is already fixed by adding methods for `[i,j]` resp. `[i,j]:=`; both for `IsPlistMatrixRep`, and also generic ones for `IsMatrixObj` which delegate to `MatElm` and `SetMatElm`.

The current test failure is something new, I'll investigate

fingolfin:
There were changes in `hpcgap/lib/vec8bit.gi` to align it closer to the code in `lib/vec8bit.gi`, making the actual differences (ConvertTo*Rep vs. Copy*Rep) stand out more. Anyway, while it is on the old MatrixObj2 branch, it doesn't really fit into this PR, so I dropped it.

markuspf:
I enthusiastically support merging this; it's useful code that shouldn't lie around somewhere bitrotting.

I haven't really worked on this code, so I don't have in-depth knowledge of what is added/changed.

I remember that `MatrixObj`s have been part of GAP for a long time already (and are used in packages), but I see that now they are documented and tested, so we should make extra-sure that (if we put this into 4.10), we're ready to continue supporting this code.

alex-konovalov:
@fingolfin hear this - doing tests now.

fingolfin:
@alex-konovalov any news on this?

alex-konovalov:
I have checked test logs from July 18th and I don't see diffs in package tests which are different from diffs that I've seen in their tests run with the GAP master branch. Since that was almost a month ago, may be worth to rebase this PR and run another round of tests to be sure.

fingolfin:
Thanks. Rebased now, let's see how the tests fare.

fingolfin:
@alex-konovalov could you please run another round of extended tests, so that we can ideally merge this ASAP? Thanks!

alex-konovalov:
@fingolfin I've triggered extended tests yesterday, so I've just checked the outcome. I will be happy to merge this now.

PR #2643, opened 2018-07-13
DO NOT MERGE: unused generic character tables for double covers of alternating/symmetric groups

fingolfin:
This PR should *not* be merged; I only made it to highlight the code in question.

@ThomasBreuer added the code which this PR removes, in a commit with this meta data:
```
Author: Thomas Breuer <sam@math.rwth-aachen.de>
Date:   2003-11-19 09:41:32 +0000

    added Felix Noeske's generic character tables of double covers of
    alternating and symmetric groups

    (Yes, it is too late for such additions, it was my fault not to
    do this earlier.  But it is a nice feature, isn't it?)
        TB
```

However, this code is not called anywhere in GAP as far as I could tell, nor does our test suite exercise it, as can be seen on https://codecov.io/gh/gap-system/gap/src/master/lib/ctblsymm.gi

Nor is it documented for users. And I could not find access to it in any package, but I did not search very hard, so it's possible that I missed some -- I assume @ThomasBreuer will know best.

So, I wonder, what should we do with it? Some possibilities:

1. Document some of the functions for the public?
2. Hook it up to some existing APIs, so that it gets used (w/o documenting its functions)
3. Move it to some package that is more appropriate (if there is any?)
4. Actually remove it (that seems rather undesirable; although, unless some package currently uses this, for 99.99% of all GAP users this won't make a difference, as they'll never know this code actually exists :-/).

Thoughts?


fingolfin:
It turns out that these are used in the `ctbllib` package (and I somehow overlooked them there the first time around).  But this is in `ctbllib/data/ctgeneri.tbl`:
```
LIBTABLE.ctgeneri.("DoubleCoverSymmetric"):=
    ShallowCopy( CharTableDoubleCoverSymmetric );
LIBTABLE.ctgeneri.("DoubleCoverAlternating"):=
    ShallowCopy( CharTableDoubleCoverAlternating );
```

Now I wonder: Wouldn't it make more sense to move those tables to the `ctbllib` package then?

fingolfin:
There is no "reject" button, but a "request for changes" with a message conveying that one is fundamentally opposed to the change IMHO does the same job just fine.

PR #2647, opened 2018-07-13
Remove dead code in grppcaut.gi

fingolfin:


ChrisJefferson:
i agree this code isn't used in any package, or referenced anywhere on the internet.

PR #2649, opened 2018-07-13
Remove dead & undocumented code from grppcext.{gi,gd}

fingolfin:
This code is not called from anywhere (other by itself), is not documented, and not used by any package.


hulpke:
I've removed myself as reviewer as this is not my code, I suspect it is early versions of material that migrated into grpconst?

PR #2651, opened 2018-07-15
Fix functions reading off end of bags

ChrisJefferson:
This is a selection of small fixes which stop functions reading off the end of bags.

These are not examples of where GAP is reading off the end of the Bag just so whole words can be read, but actively reading into other memory.

I don't think any of these could actually cause a bug (other than if, using some memory manager, GAP crashed because it read into disallowed memory), so they can't really be tested.


ChrisJefferson:
I decided I should write a test for LeftShift, and in the process found a bug in PlistVector when shifting by more than the length of the vector. I added that fix here, but it could be moved elsewhere.

PR #2652, opened 2018-07-15
Add NewWordSizedBag and ResizeWordSizedBag

ChrisJefferson:
This adds two new functions, NewWordSizedBag and ResizeWordSizedBag. These functions act the same as NewBag/ResizeBag, except they always round the allocation size up to the nearest word (sizeof(UInt)).

The code in vec8bit.c and vecgf2.c assume this behaviour, so let's make it explicit instead of implict.

### DIFF
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -432,6 +432,17 @@ extern  Bag             NewBag (
             UInt                size );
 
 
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
It is my belief (although there may still be something I've missed) that these are the only two pieces of code in the kernel which need to use `WordSized` bags.

PR #2654, opened 2018-07-18
Make Sortex stable

ChrisJefferson:
This patch makes sortex stable. As it is now stable and efficient, make SortingPerm use sortex.


ChrisJefferson:
I agree with using `ShallowCopy`. Also I have now improved the documentation to describe what we actually do! (briefly, but at least correctly)

doc/ref/lists.xml
```diff
@@ -1468,9 +1468,15 @@ copy using <Ref Func="Immutable"/>.
...
-length at most a few thousands, and has better worst-case behaviour.
+GAP implements three different families of sorting algorithms. The
+default algorithm is pattern-defeating quicksort, a variant of quicksort
+which performs better on partially sorted lists and has good worst-case
+behaviour. The functions which begin <C>Stable</C> are stable (equal
+elements keep the same relative order in the sorted list) and use
+merge sort. Finally, the functions which begin <C>Shell</C> use the shell
+sort which was GAP's default search algorithm before 4.9.
+
+<Ref Oper="Sortex"/> and <Ref Attr="SortingPerm"/> are also stable.
```
ThomasBreuer: I would move the paragraph to the beginning of the manual section on `Sorting Lists`.
In particular for someone reading the manual (and not the source),
it is not clear that this paragraph belongs to all subsections and not just to the last one.

ThomasBreuer: Great.
Thanks for this improvement.

lib/list.gi
```diff
@@ -2434,37 +2434,12 @@ end );
...
-    # return the permutation mapping <list> onto the sorted list
-    return PermList( perm )^(-1);
+        copy := ShallowCopy(list);
+        return Sortex(copy);
     end );
 
-InstallMethod( SortingPerm,
-    "for a dense and sorted list",
-    [ IsDenseList and IsSortedList ], SUM_FLAGS,
-    list -> () );
```
fingolfin: @ChrisJefferson any particular reason this method was removed? I see no mention of this change in the PR description nor the commit message.

ChrisJefferson: It was part of making SortingPerm be exactly the same as SortEx.

fingolfin: I don't see why that requires removing this optimization, though? Now if the input is sorted, we make a copy, sort that, and compute the identity permutation from that...

ChrisJefferson: It could probably have been left in, I wouldn't have a problem with it being re-added but I wasn't sure it was worth it, for a fairly unusual case.

ThomasBreuer: I see this method as an illustration of the idea to avoid working by using known information.
If the presence of such *methods for fairly unusual cases* does not slow down
the computations in the usual cases
then these methods are beneficial.
If this is not the case then it might be interesting to discuss the situation in detail,
perhaps in some tutorial.

PR #2657, opened 2018-07-23
Immutable entries of mutable attributes of char. tables

ThomasBreuer:
Character tables use several mutable attributes whose values are
extendible lists:
`ComputedClassFusions`, `ComputedIndicators`, `ComputedPowerMaps`,
`ComputedPrimeBlockss`.

Up to now, the entries of these lists were mutable.
 @frankluebeck suggested to make these entries immutable.
This change makes sense.

*Note*:
The change may have side-effects w.r.t. users' GAP code.
For example, the object returned by
```
0 * ComputedPowerMaps( CharacterTable( "A5" ) )[2]`
```
had been a *mutable* list before the changes,
and will be an *immutable* list after the changes.

In order to achieve the immutability,
the following changes were necessary.

- Extended `SupportedCharacterTableInfo` and `DeclareAttributeSuppCT`
  such that the mutability of the attribute in question is recorded;
  there seems to be no way (and no need) to access this information
  from the attribute itself,

- changed `DeclarePropertySuppCT`, here mutability makes no sense,

- let `ConvertToCharacterTableNC` make the entries of mutable attributes
  immutable.

Other changes in the context of mutability of attributes for character tables:

- Removed the `DeclareAttributeSuppCT` call for `InfoText`;
  meanwhile this attribute is declared for arbitrary GAP objects,
  redeclaring it for special objects is logically wrong,

- call `MakeImmutable` for (entries of) attribute values before the
  conversion of a record to a character table object,
  in order to avoid the creation of a copy in the standard process of
  setting attribute values,

- documented some optional components of an argument for
  `ContainedPossibleCharacters`.

### DIFF
--- a/lib/ctbl.gd
+++ b/lib/ctbl.gd
@@ -357,20 +357,18 @@ BindGlobal( "NearlyCharacterTablesFamily",
 ##  <Ref Var="SupportedCharacterTableInfo"/> is a list that contains
 ##  at position <M>3i-2</M> an attribute getter function,
 ##  at position <M>3i-1</M> the name of this attribute,
-##  and at position <M>3i</M> a list containing one or two of the
-##  strings <C>"class"</C>, <C>"character"</C>,
+##  and at position <M>3i</M> a list containing a subset of
+##  <C>[ "character", "class", "mutable" ]</C>,
 ##  depending on whether the attribute value relies on the ordering of
-##  classes or characters.
-##  This allows one to set exactly the components with these names in the
-##  record that is later converted to the new table,
-##  in order to use the values as attribute values.
-##  So the record components that shall <E>not</E> be regarded as attribute
-##  values can be ignored.
-##  Also other attributes of the old table are ignored.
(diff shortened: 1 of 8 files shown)


markuspf:
@frankluebeck @ThomasBreuer the test failures were due to the package `float` which we have disabled right now, as debugging this issue is not on our list of priorities.

fingolfin:
@frankluebeck you "requested changes", which blocks merging of this PR. But in what you write, I don't see any changes you "request" -- only a suggestion for some changes.

Could you either drop the "request for changes" (using the "Dismiss review" link on the PR page), or else clarify what the requested changes are?

ThomasBreuer:
I have now added some tests of (im)mutability, as suggested by @frankluebeck.
This was a useful hint.

ThomasBreuer:
@fingolfin  I have added an item to the release notes.

fingolfin:
Thank you!

PR #2663, opened 2018-07-26
Remove popen

ChrisJefferson:
This removes code which checks for POPEN to decide if we support gzipped files, as now we always do. This is required to support gzipped files when compiling gap to javascript, and is good general cleanup.

There is some gzip / popen code left in profiling, as there is a subtle issue there (we start compressing so early GAP hasn't even set up streams yet).


markuspf:
can your rebase this so it doesn't include the travis disable commit?

ChrisJefferson:
Weird, I assumed that would disappear by itself. Now removed.

PR #2666, opened 2018-07-28
FIX: Add IsBiCoset

### LINKED ISSUE #2555: Cosets can be multiplied even if the subgroup is not normal
In GAP it is possible to multiply (and invert) cosets. For example, take `G := S_3` and `U := <(1,2)>` and consider the cosets.
```
gap> G := SymmetricGroup(3);;
gap> U := Group( (1,2) );;
gap> cos := RightCosets(G,U);
[ RightCoset(Group([ (1,2) ]),()), RightCoset(Group([ (1,2) ]),(1,3)), RightCoset(Group([ (1,2) ]),(1,3,2)) ]
```
GAP allows the multiplication of cosets, for example `U(1,2) * U(1,2)` returns `U`.
```
gap> cos[2]*cos[2];
RightCoset(Group([ (1,2) ]),())
```
But in fact, `U(1,3)U(1,3)` equals `{(),(1,2),(2,3),(1,2,3)}` (not a coset).

This behaviour is not documented in the section about cosets.

hulpke:
and allow coset multiplication only for bicosets (as otherwise its not well-defined.

This fixes #2555

There is no test for the condition of #2555, as the illegal coset multiplication now runs into a `no method found' error.

### DIFF
--- a/lib/csetgrp.gd
+++ b/lib/csetgrp.gd
@@ -322,6 +322,24 @@ DeclareCategory("IsRightCoset", IsDomain and IsExternalOrbit and
 
 #############################################################################
 ##
+#P  IsBiCoset( <C> )
+##
+##  <#GAPDoc Label="IsBiCoset">
+##  <ManSection>
+##  <Prop Name="IsBiCoset" Arg='C'/>
+##
+##  <Description>
+##  A (right) cose is considered a <E>bicoset</E> if its set of elements simultaneously forms a left
+##  coset for the same subgroup. This is the case, for example, if the coset representative normalizes
+##  the subgroup.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareProperty( "IsBiCoset", IsRightCoset );
+
+#############################################################################
+##
 #O  RightCoset( <U>, <g> )
 ##
 ##  <#GAPDoc Label="RightCoset">
@@ -353,6 +371,8 @@ DeclareCategory("IsRightCoset", IsDomain and IsExternalOrbit and
 ##  6
 ##  gap> AsList(c);
 ##  [ (2,3,4), (1,4,2), (1,3,4,2), (1,3)(2,4), (2,4), (1,4,2,3) ]
+##  gap> IsBiCoset(c);
+##  false
 ##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -629,6 +629,14 @@ function(d)
   Print(ViewString(d));
 end);
 
+InstallMethod(IsBiCoset,"test property",true,[IsRightCoset],0,
+function(c)
+local s,r;
+  s:=ActingDomain(c);
+  r:=Representative(c);
+  return ForAll(GeneratorsOfGroup(s),x->r*x/r in s);
+end);
+
 InstallMethodWithRandomSource(Random,
   "for a random source and a RightCoset",
   [IsRandomSource, IsRightCoset],0,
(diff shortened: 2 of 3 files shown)


lib/csetgrp.gd
```diff
@@ -320,6 +320,24 @@ DeclareAttribute( "RepresentativesContainedRightCosets", IsDoubleCoset );
...
+#############################################################################
+##
+#P  IsBiCoset( <C> )
+##
+##  <#GAPDoc Label="IsBiCoset">
+##  <ManSection>
+##  <Prop Name="IsBiCoset" Arg='C'/>
+##
+##  <Description>
+##  A (right) cose is considered a <E>bicoset</E> if its set of elements simultaneously forms a left
```
markuspf: Typo: coset

markuspf:
Just as a remark (not strictly required to merge but desirable) we can test this behaviour:

```
gap> G := SymmetricGroup(3);;
gap> U := Group( (1,2) );;
gap> cos := RightCosets(G,U);;
gap> cos[2]*cos[2];
Error, no 2nd choice method found for `*' on 2 arguments

#
```

lib/csetgrp.gd
```diff
@@ -320,6 +320,24 @@ DeclareAttribute( "RepresentativesContainedRightCosets", IsDoubleCoset );
...
+##
+#P  IsBiCoset( <C> )
+##
+##  <#GAPDoc Label="IsBiCoset">
+##  <ManSection>
+##  <Prop Name="IsBiCoset" Arg='C'/>
+##
+##  <Description>
+##  A (right) coset is considered a <E>bicoset</E> if its set of elements simultaneously forms a left
+##  coset for the same subgroup. This is the case, for example, if the coset representative normalizes
```
fingolfin: Why "for example"? This is an equivalence, isn't it?

lib/csetgrp.gi
```diff
@@ -629,6 +629,14 @@ function(d)
   Print(ViewString(d));
 end);
 
+InstallMethod(IsBiCoset,"test property",true,[IsRightCoset],0,
+function(c)
+local s,r;
+  s:=ActingDomain(c);
+  r:=Representative(c);
+  return ForAll(GeneratorsOfGroup(s),x->r*x/r in s);
```
fingolfin: Perhaps test for `x^r in s` instead, as that will be a bit more efficient in some cases (e.g. for permutations)?

lib/csetgrp.gi
```diff
@@ -656,10 +664,17 @@ end);
...
 function(a,b)
-  if ActingDomain(a)<>ActingDomain(b) then
-    Error("no multiplication defined for cosets of different subgroups");
+local c;
+  if ActingDomain(a)<>ActingDomain(b) then TryNextMethod();fi;
+  if not IsBiCoset(a) then # product does not require b to be bicoset
+    TryNextMethod();
+  fi; 
+  c:=RightCoset(ActingDomain(a), Representative(a) * Representative(b) );
+  if IsBiCoset(b) then
```
fingolfin: Perhaps first check HasIsBicoset to avoid a computation?

lib/csetgrp.gi
```diff
@@ -668,10 +683,12 @@ function(a)
 local s,r;
   s:=ActingDomain(a);
   r:=Representative(a);
-  if ForAny(GeneratorsOfGroup(s),x->not x^r in s) then
+  if not IsBiCoset(a) then
     Error("Inversion only works for cosets of normal subgroups");
```
fingolfin: That error message is wrong (though of course it was before, too)

fingolfin:
@hulpke to clarify (esp. in light of what was discussed during yesterdays video conference call; you may have read the notes): all my comments are about minor issues. If you want, you can ignore them, and just merge this, though personally I think it would make sense to at least add a test as suggested by @markuspf. I can address any remaining points with a PR of my own afterwards, too (or alternatively, I can also push fixes to this PR, if you are fine with that)

lib/csetgrp.gi
hulpke: Yes, if this ever could be time-critical this has the potential to be faster. I decided on the inverse-conjugation as it translated directly to the condition as I had written it doen on paper.

hulpke:
Ignore the close/open. Clumsy fingers when carrying the laptop with the web page open...

markuspf:
Just to clarify: Do we want to merge this now?

hulpke:
@markuspf I wasn't planning to add more, so as far as I'm concerned this is ready to merge or to discard.

PR #2667, opened 2018-07-30
Add C++ support to kernel; use it to reduce code duplication in permutat.c, objfgelm.c and others

fingolfin:
This is an incomplete proof of concept (POC) to show how using C++ code could be used to simplify some kernel code. For now, my main goal is to reduce code duplication, which is why I started with `permutat.c`. This modest PR already removes about a thousand lines of code.

It was mentioned that not everybody who works (or wants to work) on the kernel is fluent in C++. Thus we should be careful about which features we use, and avoid overly complicated or intransparent / "magical" code, and add sufficient comments. Consider this PR as a first opportunity to get an idea of what the result of such a conversion to C++ code might look like, and please point out anything you don't understand, dislike (or like), etc.

Some things which still need to be done (assuming this PR gets a green light):
* [x] Ensure this PR passes the GAP test suite.
* [x] Require C++ compiler in `configure`, and abort (with a helpful message) if none is found.
* [x] Update the README and other documentation to mention that a C++ compilers is required
* [ ] Optional: teach `gac` to compile `.cc` files
* [x] Decide which C++ standard to allow (and enable, via compiler flags): for now, I restricted myself to C++98; but we could use C++11, which was added in GCC 4.8.1, which is e.g. available in Ubunutu 14.04 LTS (and thus also on Travis)
* [ ] document which subset of C++ we allow for now? E.g. we likely want to avoid the C++ runtime (so no exceptions, no standard C++ library, no RTTI, no global constructors, ...?). I guess in the end, we can just forbid some things, explicitly allow some others, and for the rest, rely on PR reviews to determine what we want and what not
* [ ] ...

Other places that could benefit from a similar use of C++ templates:
* more functions in `permutat.cc` could actually be treated, the remaining ones each simply require a bit more work than the ones I already converted, so I didn't want to spend that effort for a proof of concept
* `pperm.c`: very similar to `permutat.c`
* `trans.c`: very similar to `permutat.c`
* `objccoll-impl.h` and `objscoll-impl.h` implement "poor man's template's" by abusing the C preprocessor, and are included multiple times by `objscoll.c` for that
* similarly, `sortbase.h` is included multiple times by `listfunc.c`
* `objfgelm.c`: contains 8, 16 and 32 bit variants of many functions, e.g.: `Func8Bits_ExponentSums3`, `Func16Bits_ExponentSums3`, `Func32Bits_ExponentSums3`
* `objpcgel.c`: contains 8, 16 and 32 bit variants of many functions


src/permutat.cc
```diff
@@ -67,6 +69,36 @@
...
+{
+    return (T *)(ADDR_OBJ(perm) + 1);
+}
+
+template<typename T> static inline const T * CONST_ADDR_PERM(Obj perm)
+{
+    return (const T *)(CONST_ADDR_OBJ(perm) + 1);
+}
+
+template<typename TL, typename TR> struct ResultType { typedef UInt4 type; };
```
ChrisJefferson: Very low-level comment, might want to give this another name before merging because I imagine we'll have other similar types. However, the same thing can be used for perms, trans and pperms.. not sure of a good name.

src/permutat.cc
```diff
@@ -67,6 +69,36 @@
 #include "sysfiles.h"
 #include "trans.h"
 
+}
+
+template<typename T> static inline UInt SIZEBAG_PERM(UInt deg)
+{
+    return sizeof(Obj) + deg * sizeof(T);
+}
+
```
ChrisJefferson: Might be nice to have some kind of `CHECK_PERM_TYPE<T>(Obj)` function, that could be used with GAP_ASSERT to check if `T` is the right kind of type to access `Obj` Something like (sorry, not checked with compiler)

```
template<typename T> static inline CHECK_PERM_TYPE<T>(Obj) { return false; }
template<> static inline CHECK_PERM_TYPE<UInt2>(Obj o) { return TNUM_OBJ(o) == T_PERM2; }
template<> static inline CHECK_PERM_TYPE<UInt4>(Obj o) { return TNUM_OBJ(o) == T_PERM4; }
```

ssiccha:
Looks nice. I like the idea. :+1:

sebasguts:
This PR looks really nice, and I really like the way it it cleans up the code.

Two opinions here:
* I think we should go with C++11. It is available on every recent Linux Distro, and has a lot of features one would really like to use, and for which there is no reason to forbid them:
  + `auto` and `decltype`: Specially useful when working with templates
  + lambdas
* Allowing stuff: I think we should at first only allow a certain subset. Forbidding some stuff seems to have the caveat that C++ is so large that you easily forget something. But do we even want to try to not use the C++ runtime? Many things might be useful, specially exceptions and STL. Plus, several packages already load it anyway.

stevelinton:
I like the simplification of permutat.c a lot, and I'd be happy to see use of C++ templates of this level of sophistication. Essentially it's just a better C pre-processor. I'd want to move very slowly with things like the C++ runtime or use of the STL for two reasons:

1. because of possible interactions with gasman/libGAP/the existing GAP error and exception handling and so on. 
2. because I'd like to still be able to read, understand and modify the code. My knowledge of C++ is rudimentary at best and I'm pretty sure I'm not the only developer in that position.

frankluebeck:
I completely agree with the previous comment by @stevelinton.

frankluebeck:
With respect to  the content: Wouldn't it be easy and sensible to introduce a type `T_PERM1` with the simplified code in `permutat.cc`? I guess "most" permutations in "most" GAP sessions would fit in a `T_PERM1`?

fingolfin:
@frankluebeck  Adding a `T_PERM1` indeed should be a relatively easy thing to do; this is a nice bonus effect of the proposed change. However, first the rest of the code in `permutat.cc` should be turned into templates

And I acknowledge the reservations about adding too much C++. While I personally am completely comfortable with using C++, I agree we should be careful about using it too much. Thus any PR that would add more C++ usage should be carefully reviewed for that (and I personally think we should strive to make sure that people with concerns about C++ usage, like Frank and Steve, get a chance to comment on whether they are comfortable with it).

That said, there are a few things beyond basic templates which I think we could benefit from; but let's discuss those if and once we get there (and then ideally again based on a concrete PR to showcase what is intended).

For now, the HPC-GAP problem with `ward` remains...

ChrisJefferson:
Note ward is no more broken with this PR than it was before (that is, you can still build HPCGAP, and there are no guards). Of course, merging this gets us even further from a functioning ward.

I would personally be interested in trying to merge this sooner rather than later, rather than letting it sit for a long time.

I would suggest to start, we fix on C++98, and just templated functions, as you have done here. Extra features can be considered when we have a good quality use of them (like you have already got here). My preference would be to have a reasonably high bar for adding new C++ features, rather than writing things in a C++ style just because it's the more "C++" way of doing something.

fingolfin:
The test suite passes now. As you can see, I also "completed" conversion to C++ of `permutat.c` and also `objpcgel.c`.  So yeah, I am also interested in not letting this bitrot.

I also updated `configure`, `README.md` and `INSTALL.md` to state that a C++ compiler is required; of course the release notes should then also mention this; and probably the website, too.

Regarding which C++ standard to use: Let me point out that we recently got a bug report by somebody still using GCC 4.4 (first released in 2009, almost exactly 9 years ago). Obviously this has not support for C++11, only at most a subset (see <https://gcc.gnu.org/gcc-4.4/cxx0x_status.html> for details). I think this should inform the decision. While personally, I wouldn't find it unreasonable to require people to use GCC 4.8 (released March 2015, i.e. 5+ years ago) and enable `C++11`, there also seems no strong need for that right now. Hence, for the time being, I'd suggest to stick to C++98, and possibly even activate the `-std=c++98` compiler flag, to ensure one does not accidentally use newer features. Finally, it would also be possible to use a subset of C++11 already supported in GCC 4.4; e.g. `auto` is available in GCC 4.4 (but not in GCC 4.3, first released March 2008). But then we need to be extra careful about what we use and what not...

fingolfin:
BTW one thing that is annoying right now is that the renaming of `.c` files to `.cc` files triggers a minor glitch in the dependency tracking -- it'll error our, complaining that it cannot find e.g. `src/permutat.c`. However, this can be resolve by deleting the `permutat.d` file, or by just doing a `make clean && make`.

james-d-mitchell:
Not sure if my opinion counts for anything, but I'm in favour of this change. It could even go further there is a fair amount of overlap between `src/trans.c`, `src/pperm.c` and `src/permutat.c`, and these could be combined to some extent.

fingolfin:
@james-d-mitchell absolutely, some of that code could and probably should be "merged" back together eventually. But let's first get this PR merged...

I guess to move forward, it would be prudent to notify the gap@ and support@ lists to announce the plan to start requiring a C++ compiler, and then wait a little. Anybody volunteering to do that? Also, any other MUST-DOs before a merge?

fingolfin:
BTW, I encourage anybody in favor of this PR to actually "approve" it via the GitHub web interface. Right now, only @stevelinton did so. Also, if you have serious concerns, of course feel free to "require changes" and explain your concerns.

ChrisJefferson:
One minor comment (on a phone). If we decide to hardwire a version, we should use gnu++98 (or whatever version) rather than c++98. The only difference is "allow gnu extensions", and on some OSes, you can't even include the standard C headers with c++98 or c++11 (we hit this problem in semigroups previously).

fingolfin:
TODO: figure out why codecov shows no coverage for the .cc files

alex-konovalov:
This does not compile on macOS with `gcc -v` reporting
```
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
This is what happens:
```
=== DONE building gmp ===
  C       src/blister.c => obj/src/blister.lo
  C       src/bool.c => obj/src/bool.lo
  C       src/c_oper1.c => obj/src/c_oper1.lo
  C       src/c_type1.c => obj/src/c_type1.lo
  C       src/calls.c => obj/src/calls.lo
  C       src/code.c => obj/src/code.lo
  C       src/compiler.c => obj/src/compiler.lo
  C       src/compstat.c => obj/src/compstat.lo
  C       src/costab.c => obj/src/costab.lo
  C       src/cyclotom.c => obj/src/cyclotom.lo
  C       src/debug.c => obj/src/debug.lo
  C       src/dt.c => obj/src/dt.lo
  C       src/dteval.c => obj/src/dteval.lo
  C       src/error.c => obj/src/error.lo
  C       src/exprs.c => obj/src/exprs.lo
  C       src/ffdata.c => obj/src/ffdata.lo
  C       src/finfield.c => obj/src/finfield.lo
  C       src/funcs.c => obj/src/funcs.lo
  C       src/gap.c => obj/src/gap.lo
  SED     src/gap_version.c.in => gen/gap_version.c
  C       gen/gap_version.c => obj/gen/gap_version.lo
  C       src/gvars.c => obj/src/gvars.lo
  C       src/hookintrprtr.c => obj/src/hookintrprtr.lo
  C       src/integer.c => obj/src/integer.lo
  C       src/intfuncs.c => obj/src/intfuncs.lo
  C       src/intrprtr.c => obj/src/intrprtr.lo
  C       src/io.c => obj/src/io.lo
  C       src/iostream.c => obj/src/iostream.lo
  C       src/listfunc.c => obj/src/listfunc.lo
  C       src/listoper.c => obj/src/listoper.lo
  C       src/lists.c => obj/src/lists.lo
  C       src/macfloat.c => obj/src/macfloat.lo
  C       src/modules.c => obj/src/modules.lo
  C       src/modules-builtin.c => obj/src/modules-builtin.lo
  C       src/objcftl.c => obj/src/objcftl.lo
  C       src/objects.c => obj/src/objects.lo
  C       src/objfgelm.c => obj/src/objfgelm.lo
   * new C++ compiler flags
  CXX     src/objpcgel.cc => obj/src/objpcgel.lo
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
make: *** [obj/src/objpcgel.lo] Error 1
Build step 'Execute shell' marked build as failure
```

alex-konovalov:
`make V=1` ends with 
```
    * new C++ compiler flags
/bin/sh ./libtool --mode=compile c++ -MQ "obj/src/objpcgel.lo" -MMD -MP -MF "obj/src/.deps/objpcgel.d"  -g -O2 -fno-strict-aliasing -Wall -Wextra -Wundef -Wwrite-strings -Wpointer-arith -Wredundant-decls -Wno-unused-parameter -Wmissing-field-initializers -Wformat=2 -Wformat-nonliteral -Wformat-security -Wno-sign-compare -Wstrict-aliasing -Wshadow -Winline -Wpacked -Wmissing-format-attribute -Winit-self -Wmissing-include-dirs -Warray-bounds -Wreturn-type -Wno-inline -Wdouble-promotion  -I./gen -I./src -DHAVE_CONFIG_H -I/Volumes/hudson-fs/workspace/GAP-pull-request-quicktest/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/install/label/fruitloop/GAP-pull-request-snapshot/extern/install/gmp/include       -c src/objpcgel.cc -o obj/src/objpcgel.lo
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
make: *** [obj/src/objpcgel.lo] Error 1
```

alex-konovalov:
Also, I can't get libtool version there (similarly to the case reported in https://apple.stackexchange.com/questions/239830/change-libtool-used-in-command-line). I think both macOS machines where I can reproduce this problem have libtool that comes with macOS.

fingolfin:
To get the libtool version, just do `./libtool --version` inside the GAP directory

james-d-mitchell:
In case it helps, the `libtool` shipped with recent versions of Mac OSX (at least in 10.13.*) seems to be badly broken, I had to install it again using `brew`. The reported versions of both the standard `libtool` and `glibtool` installed using `brew` are:

~~~~
glibtool (GNU libtool) 2.4.6
Written by Gordon Matzigkeit, 1996

Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
~~~~

fingolfin:
I do not think Apple ever shipped any version of GNU libtool. But they always has a completely unrelated executable called `libtool`.

So, I still wonder which version @alexk is using, which my instructions would reveal.

alex-konovalov:
@fingolfin you've just pinged another Alex :)

Also, that's what your instructions reveal:

```
$ ./libtool --version
libtool (GNU libtool) 2.4.6
Written by Gordon Matzigkeit, 1996

Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

fingolfin:
@markuspf actually, I just looked, and it seems that we already do have some basic support for compiling .cc/.cpp files in `gac` -- in `master`! But I don't know if it (still) works. Give it a try?

@alex-konovalov so that's the latest `libtool`. Very odd, as I have a pretty similar setup; besides, I can't think of *any* reason this error would appear here. Is this is in a clean build dir (e.g. fresh checkout, or after `make distclean`)

fingolfin:
@alex-konovalov what does `c++ --version` output (it may differ from `gcc --version`)

fingolfin:
@alex-konovalov you could also try applying this patch, and see if that resolves the issue:
```patch
diff --git a/Makefile.rules b/Makefile.rules
index c9cb249d3..b5d9d79fa 100644
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -361,12 +361,12 @@ DEPFLAGS = -MQ "$@" -MMD -MP -MF $(DEPFILE)
 # Build rule for C source files
 obj/%.lo: %.c cnf/GAP-CFLAGS cnf/GAP-CPPFLAGS libtool
 	@$(MKDIR_P) $(@D)/$(DEPDIRNAME)
-	$(QUIET_CC)$(LIBTOOL) --mode=compile $(CC) $(DEPFLAGS) $(GAP_CFLAGS) $(WARN_CFLAGS) $(GAP_CPPFLAGS) -c $< -o $@
+	$(QUIET_CC)$(LIBTOOL) --mode=compile --tag CC $(CC) $(DEPFLAGS) $(GAP_CFLAGS) $(WARN_CFLAGS) $(GAP_CPPFLAGS) -c $< -o $@
 
 # Build rule for C++ source files
 obj/%.lo: %.cc cnf/GAP-CXXFLAGS cnf/GAP-CPPFLAGS libtool
 	@$(MKDIR_P) $(@D)/$(DEPDIRNAME)
-	$(QUIET_CXX)$(LIBTOOL) --mode=compile $(CXX) $(DEPFLAGS) $(GAP_CXXFLAGS) $(WARN_CXXFLAGS) $(GAP_CPPFLAGS) -c $< -o $@
+	$(QUIET_CXX)$(LIBTOOL) --mode=compile --tag CXX $(CXX) $(DEPFLAGS) $(GAP_CXXFLAGS) $(WARN_CXXFLAGS) $(GAP_CPPFLAGS) -c $< -o $@
 
 
 ########################################################################
```

alex-konovalov:
@fingolfin 
```
$ c++ --version
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
and
```
$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
fruitloop:~ hudson$ 
```
I will look at patch later.

fingolfin:
That's the exact same compiler version I am using. That made me wonder if something else in e.g. the PATH might affect things. So I rerun the test over here, but with a restricted PATH:
```
PATH=/usr/bin:/bin:/usr/sbin:/sbin ./configure
PATH=/usr/bin:/bin:/usr/sbin:/sbin make -j8
```
And now I can reproduce the issue. Still not quite what causes it, but my patch indeed fixes it. Will update this PR ASAP.

ChrisJefferson:
Very stupid possible fix: Rename the files, for example to permutation.cc (instead of permutat.cc).

I actually just tried this and tried then changing all places in the build system where this would matter, but it turned out I couldn't find them all, I'm not sure which bit I missing.

fingolfin:
@ChrisJefferson there is exactly one spot in the build system to you need to modify in order to rename a source file: `Makefile.rules`, when setting the `SOURCES` variable (in contrast to the old build system, where one also had to remember to regenerate other parts of the build system to match).

And yeah, of course we could just change the stem of the filename, as a work around. Though this may not be as clear with other files as in this case.

But I guess it wouldn't hurt to choose a new name for `objpcgel.cc`, too ;-) (that's the other file converted to C++ by this PR).

alex-konovalov:
@fingolfin confirm that problems on macOS resolved. Let me just check what happens on Windows...

alex-konovalov:
And on Windows I still can't get it to work when started from a bat.file. It starts on the machine with Cygwin, but does not work when transferred onto a Cygwin-free machine. No error messages, and `cygcheck` and `ldd` do not report any missing dll.

ChrisJefferson:
@alex-konovalov : Are you grabbing cygstdc++-6.dll ?

alex-konovalov:
@ChrisJefferson yes. here is the full list:

/bin/cyggcc_s-1.dll /bin/cyggmp-10.dll /bin/cygncurses-10.dll /bin/cygncursesw-10.dll /bin/cygpanel-10.dll /bin/cygpanelw-10.dll /bin/cygpopt-0.dll /bin/cygreadline7.dll /bin/cygstart.exe /bin/cygstdc++-6.dll /bin/cygwin1.dll /bin/libW11.dll /bin/mintty.exe /usr/bin/cygz.dll /usr/bin/rxvt.exe /usr/bin/regtool.exe /usr/x86_64-w64-mingw32/sys-root/mingw/bin/zlib1.dll 

All placed in bin/i686-pc-cygwin-default32

alex-konovalov:
1. Maybe - it goes a bit further before disappearing then in the cases of missing dlls.

2. All latest builds (this morning on Jenkins) are runnable so far.

ChrisJefferson:
Are you sure they run on another computer?

I'm finding the master branch, if I do something like: `gap tst\testinstall.g > out.txt`, works fine, out.txt has the output I expect.

However, if I try to run interactively in a cmd shell, then GAP doesn't work. I haven't tried running GAP in alternative shells.

I'm not going to have any time myself to look at this until at least Tuesday. Some bisecting would be nice, but will take ages as GAP takes so long to build on cygwin, and you have to then run the GAP on another machine :(

alex-konovalov:
@ChrisJefferson: GAP master branch works well on Cygwin-free machine: https://gap-ci.cs.st-andrews.ac.uk/view/GAP-nightly/job/GAP-master-windows-test/217/. The build of the PR where t does not work is https://gap-ci.cs.st-andrews.ac.uk/job/GAP-pull-request-windows-test/14/console. I've compared setup for both builds, don't see any differences (all paths are adjusted correctly).

alex-konovalov:
@ChrisJefferson it does 
screen shot 2018-08-31 at 13 28 01

alex-konovalov:
Also, with `gapcmd.bat` 
screen shot 2018-08-31 at 13 29 33

ChrisJefferson:
Weird. Thanks. I'll try to figure out why I can't even get master to work on my VM.

alex-konovalov:
I will also have another go at comparing build setup for the Jenkins job to check pull requests (4 jobs a https://gap-ci.cs.st-andrews.ac.uk/view/Pull%20requests/)

ChrisJefferson:
Just in case anyone else has a go at this while I'm away over the weekend.

I can make GAP run by doing:

`mintty ./ash.exe -c ./gap.exe` (copy `ash.exe` into GAP's directory)

I don't know why running through `ash` fixes things, but I found some bad google pages which send running programs through a shell sometimes gets stdin/stdout set up directly.

I don't know why we now need this, when we didn't before.

alex-konovalov:
I have tried another PT and still have the same effect. I don't think the problem is in this PR - likely some trivial error in Jenkins job configuration, but I don't have time to look just now. Will try by the end of the week - it's still very desirable to get this tested on Windows properly.

ChrisJefferson:
@dimpase : I would say at the moment we don't want people using `auto`, so it doesn't matter if we ban it by staying on C++98. Similarly, I wouldn't want people using variadic templates or lambda functions.

As a matter of practical experience (and I'm happy for other people to have input), both semigroups and ferret tried setting the standard to (in our cases) `-std=c++11`, and on several machines the code wouldn't compile as the standard OS headers wouldn't compile unless we used `-std=gnu++11`. The differences between `gnu++` and `c++` are quite small, and clang supports them very well, because they know in practice most code until recently was written for gcc/g++ which defaults to the `gnu` standards by default.

fingolfin:
I'd really like to get this merged, esp. as ongoing word on HPC-GAP / unward in PR #2845 is likely to touch `src/objpcgel.c` which this PR here converts to C++. There are a few more files where I think it'd be best to convert them to C++ first before trying to CONSTify code in them, as that is less work after the code duplication has been reduced.

My understanding is that the only blocker for this is that @alex-konovalov is not able to build resp. run this under Cygwin, while @ChrisJefferson can. To quote @alex-konovalov from 23 days ago:
> I don't think the problem is in this PR - likely some trivial error in Jenkins job configuration, but I don't have time to look just now. Will try by the end of the week
Clearly this did not work out (I am not blaming anybody  here, I know how it is), but it'd still be great if we could make progress on this soon.

In the meantime, I'll look into the annoyance that causes one to have to do a `make clean` (resp. remove the offending `.deps/*.d` files) when switching a checkout across a commit that renamed a .c file to .cc (or vice versa), which currently results in errors like this:
```
make: *** No rule to make target 'src/objpcgel.c', needed by 'obj/src/objpcgel.lo'.  Stop.
```
I might just take up @ChrisJefferson suggestion and change the stem name of the affected files, though I'll first think some more about a "clean" solution.

ChrisJefferson:
I'm going to go for the merge once tests pass, then investigate the cygwin issue if it isn't cleared up.

fingolfin:
@ChrisJefferson tests passed now

alex-konovalov:
The windows issue described above is not happening in the master branch. I suspect there is a bug somewhere in paths in scripts from Jenkins jobs to build and tests PRs on Windows.

However, there are new Windows issues which seem to be cause by this PR. I've reported them in #2890.

Finally, when this PR was tested on Windows, the directory for the binaries had name `bin/i686-pc-cygwin-default32-kv5`, but in the master branch the name is `bin/i686-pc-cygwin-default32-kv5`. I had to adjust Windows Jenkins script to add `-kv5`.

PR #2669, opened 2018-07-30
Adjust comment in DoSnAnGiantTest

ssiccha:
Size for permutation groups is not an immediate method anymore.
The function is still correct though.

### DIFF
--- a/lib/gpprmsya.gi
+++ b/lib/gpprmsya.gi
@@ -563,7 +563,7 @@ local dom, n, mine, root, d, k, b, m, l,lh;
   fi;
   # so the group is primitive
 
-  if n>10 then # otherwise the size is immediate
+  if n>10 then
     # test whether its socle could be A_l^m, acting on k-tuples in product
     # action. 
     mine:=Minimum(List(Collected(Factors(n)),i->i[2]));


lib/gpprmsya.gi
```diff
@@ -563,7 +563,7 @@ local dom, n, mine, root, d, k, b, m, l,lh;
   fi;
   # so the group is primitive
 
-  if n>10 then # otherwise the size is immediate
+  if n>10 then
```
fingolfin: I don't think this refers to immediate methods. Rather, the point is that for degree <= 10, a direct computation is so fast that it is essentially "immediate", and there is no point in using more clever code (in fact, the more clever code may be slower).

I don't think that the size of a non-cyclic permutation group was ever computed via an immediate methods (and I don't see how it could have been done either).

ssiccha: Ah, I assumed this meant that the size got computed (in GAP 4.8.8):
```
gap> G := Group((1,2,3), (1,2,3,4,5,6,7));
#I  immediate: Size
#I  immediate: IsCyclic
#I  immediate: IsCommutative
#I  immediate: IsTrivial
Group([ (1,2,3), (1,2,3,4,5,6,7) ])
```
But `HasSize` returns false.

PR #2670, opened 2018-07-30
lib: remove unused CheckAuto()

fingolfin:
This is neither used by the library nor by any packages

### DIFF
--- a/lib/grppcaut.gi
+++ b/lib/grppcaut.gi
@@ -9,31 +9,6 @@
 
 #############################################################################
 ##
-#F CheckAuto( auto )
-##
-CheckAuto := function( auto )
-    local new,mapi;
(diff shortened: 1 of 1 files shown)


stevelinton:
Looks like a debugging tools that somebody needed once.  Unlike many such, it's fairly clear what it does, and it doesn't look like it's suffered bitrot (yet). On the other hand anyone who's likely to be able to use it effectively can probably rewrite it in two minutes. Is it better to document and retain such things (maybe in a separate directory), or to delete them?

fingolfin:
Well, that function was originally added by Bettina 1997-07-07; it was updated later by e.g. @hulpke 

Instead of removing this, one could of course also add test cases using it, and or `Assert` statements to relevant functions. Perhaps @hulpke has a preference on this?

PR #2671, opened 2018-07-31
kernel: use zlib to write profile data

fingolfin:
This also gets rid of the last use of `popen`, so remove the check for it from `configure.ac`.

This completes the work begun in PR #2663.


ChrisJefferson:
It would be good to do some checking of performance. Also now the profiling has settled a little, it might be a good idea to try outputting a compressed binary format, instead of JSON, which would reduce the amount of data outputted.

(Alternatively, we could be even better and output less data, but that's another, bigger, change).

PR #2676, opened 2018-08-01
Mock-up for hpc task functionality

emmakatherina:
This commit adds mock-ups for ScheduleTask and WaitTask similar to the mock-ups
for RunTask and TaskResult that already exist. This makes it easier to execute
code written for hpcgap in gap.

Please review and merge!

### DIFF
--- a/tst/testinstall/hpc/tasks.tst
+++ b/tst/testinstall/hpc/tasks.tst
@@ -1,11 +1,11 @@
 #############################################################################
 ##
-#W  task.tst                 GAP tests                   Alexander Konovalov
+#W  tasks.tst                 GAP tests                   Alexander Konovalov
 ##
 ##
 #Y  Copyright (C)  2012
 ##
-gap> START_TEST("task.tst");
+gap> START_TEST("tasks.tst");
 gap> CallAsTask := function(arg)
 > return TaskResult( RunTask( CallFuncList, arg[1], arg{[2..Length(arg)]} ) );
 > end;;
@@ -73,4 +73,15 @@ rec( B := [  ], basis := [  ], mue := [  ] )
 gap> CallAsTask(LLLReducedBasis,[ [ 0, 0 ], [ 0, 0 ] ], "linearcomb" );
 rec( B := [  ], basis := [  ], mue := [  ], 
   relations := [ [ 1, 0 ], [ 0, 1 ] ], transformation := [  ] )
-gap> STOP_TEST( "task.tst", 1 );
+gap> task := RunTask(GF, 7);;
+gap> TaskResult(task);
+GF(7)
+gap> task2 := ScheduleTask(task, x -> x^2, 4);;
+gap> WaitTask(task, task2);
+gap> task3 := ScheduleTask([task, task2], x -> x^2, 5);;
+gap> TaskResult(task3);
+25
+gap> task4 := ScheduleTask([], GcdInt, 6, 10);;
+gap> TaskResult(task4);
+2
+gap> STOP_TEST( "tasks.tst", 1 );
--- a/lib/hpc/tasks.g
+++ b/lib/hpc/tasks.g
@@ -23,3 +23,11 @@ end;
 TaskResult := function(task)
     return task.taskresult;
 end;
+
+ScheduleTask := function(cond, func, args...)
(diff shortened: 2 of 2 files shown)


ChrisJefferson:
Hi, thanks for your patch!

While the basic patch is right, the tests are unfortunately failed. ScheduleTask takes as it's first argument usually a single task, or list of tasks (although this is very poorly documented!)

Perhaps change the middle part of your test to:

```
gap> task2 := ScheduleTask(task, x -> x^2, 4);;
gap> WaitTask(task, task2);
gap> task3 := ScheduleTask([task, task2], x -> x^2, 5);;
gap> TaskResult(task3);
```

(or something else would be fine too!)

markuspf:
Just to clarify what @ChrisJefferson says a bit

Looking at the code in HPC-GAP reveals (https://github.com/gap-system/gap/blob/a833a5a2d269e8a228b9f479348ce954a6b1c4e9/hpcgap/lib/hpc/stdtasks.g#L397) that the arguments for `ScheduleTask(cond, func, args...)` are 

  * a condition, or a list `cond` of conditions
  * a function `func`
  * the parameters `args...` for the task function 

This is documented in the HPC-GAP manual (https://github.com/gap-system/gap/blob/a833a5a2d269e8a228b9f479348ce954a6b1c4e9/doc/hpc/tasks.xml#L106), conditions can be more complex than just tasks (which is documented in section 1.5 of the HPC-GAP manual).

ChrisJefferson:
The waiting for tasks bit is ok, because in the Mock up we execute all tasks before RunTask returns (as that seems like the only sensible option to me).

markuspf:
The current `ScheduleTask` mock up does not touch `cond`. Code like this:

```
gap> t1 := DelayTask(...);
gap> t2 := ScheduleTask(t1, ...);
gap> RunTask(t1);
```

Won't work (not that `DelayTask` is stubbed yet, one could just run the tasks in TaskResult). In general this stubbing is turning up quite icky questions about semantics: if you use 2 algorithms to achieve some result and one of them could for some reasons actually not terminate, but the other one will this code will work with proper HPC-GAP, but not with GAP:

```
gap> t1 := DelayTask(...); t2 := DelayTask(...);
gap> t3 := ScheduleTask([t1, t2], ...);
gap> RunTask(t1);
```

(I am sure we would have discussed this before; how much of this stubbing should we actually support, then?)

ChrisJefferson:
My preference is to implement whatever people want, in the simplest possible format (just to make people's life easier).

As you say, DelayTask raises more issues, and would require changes to the ScheduleTask mock. However, we can do that later, if anyone needs it.

emmakatherina:
If I understand it correctly than the error mentioned above by @markuspf can't be avoided anyway if we don't want to implement a system that executes both tasks at the same time in a sequential way through alternating between the calculation of the two tasks (e.g. until one of the tasks finishes). Therefore the error persists anyway and I don't see a reason not to implement the mock-ups.

tst/testinstall/hpc/tasks.tst
```diff
@@ -1,11 +1,11 @@
 #############################################################################
 ##
-#W  task.tst                 GAP tests                   Alexander Konovalov
+#W  tasks.tst                 GAP tests                   Alexander Konovalov
 ##
 ##
 #Y  Copyright (C)  2012
 ##
-gap> START_TEST("task.tst");
+gap> START_TEST("tasks.tst");
```
fingolfin: Any particular reason for renaming the file (which increases the diffs in this PR but otherwise seems to serve no purpose)?

emmakatherina:
I will update this PR in the next days, sorry about the delay!

tst/testinstall/hpc/tasks.tst
emmakatherina: It describes the content of the file better and is more consistent with the naming of the other file in the directory. But I don't insist on the changes...

fingolfin:
Looks OK to me now, although in an ideal world, I'd squash all changes down into a single commit ... @emmakatherina would you be comfortable doing that? But don't worry if your git-foo is not that strong, then we'll just do without it.

emmakatherina:
I will try it in a minute!

Issue #2677, opened 2018-08-01
Should `\<` and `\>` be balanced?


ThomasBreuer:
The markers `\<` and `\>` can be used to indicate preferences for line breaks.
I would expect that the two markers should be balanced inside individual
`ViewObj`, `ViewString` etc. methods.
Is this the case?

Background  of the question:
Recently I stumbled over an unbalanced example, in `lib/csetgrp.gi`:
```
InstallMethod(ViewString,"DoubleCoset",true,[IsDoubleCoset],0,
function(d)
  return(STRINGIFY("DoubleCoset(\<",
                   ViewString(LeftActingGroup(d)),",\>",
                   ViewString(Representative(d)),",\>",
                   ViewString(RightActingGroup(d)),")"));
end);
```
A heuristic filter for finding such cases is the following.
```
gap> unbalanced:= function( func )
>       local str;
>       str:= String( func );
>       return Length( PositionsSublist( str, "\\<" ) ) <>
>              Length( PositionsSublist( str, "\\>" ) );
>    end;
gap> meth:= MethodsOperation( ViewString, 1 );;
gap> cand:= Filtered( meth, r -> unbalanced( r.func ) );;
gap> Length( cand );
13
```
(Of course, `unbalanced` may find candidates that return balanced strings,
and methods that return unbalanced strings may be overlooked by `unbalanced`,
for example due to `if` statements in the code of the method in question.)

fingolfin:
This is a bug. It throws of the printing logic, and also causes a spurious newline to be added when viewing a double coset:
```
gap> G:=SymmetricGroup(2);
Sym( [ 1 .. 2 ] )
gap> s:=DoubleCoset(G, (1,2,3), G);
DoubleCoset(Sym( [ 1 .. 2 ] ),(1,2,3),Sym( [ 1 .. 2 ] ))

gap> RightCoset(G, (1,2,3));
RightCoset(Sym( [ 1 .. 2 ] ),(1,2,3))
```

fingolfin:
Twelve of the matches reported go back to the same two functions in the library, namely `_ViewStringForSemigroups` and `_ViewStringForSemigroupsGroups`. In one case, this is a spurious match, in the other an actual bug. I will make a PR fixing both.

[...]

PR #2680, opened 2018-08-02
Balance \> and \< in semigroups ViewString methods

fingolfin:
See issue #2677

I don't think this needs to be in the release notes.


james-d-mitchell:
@fingolfin sorry I only saw this now (was on holiday).

Issue #2683, opened 2018-08-06
Unexpected slowness when computing `NormalSubgroups(SymmetricGroup(n))` for some n


markuspf:
@mtorpey and I just happened upon the following unexpected behaviour:

```
gap> for i in [1..1000] do S := SymmetricGroup(40); NormalSubgroups(S);  Print("hi\n"); od;
```

Now sometimes GAP computes the normal subgroups of `Sym(40)` quite quickly (1-2 seconds), sometimes it takes a minute or (on @mtorpey's computer) seems to get stuck (in partition backtrack).

Maybe @hulpke can shine some light on where randomisation (pointing fingers randomly at reasons why this should even happen) is involved in this computation.

Of course for symmetric and alternating groups we can just install methods that give us the normal subgroups (@mtorpey is working on this) in basically no time at all, but there might be an underlying issue that makes this computation sometimes slower than it needs to be, and I'd like to know what it is.

markuspf:
Update: The code seems to never get stuck, it just has very large variations in runtime: from `1.2` seconds all the way to one hour and 20 minutes letting it run over night.

hulpke:
What happens is that the code of a centralizer of a random element is computed. In permutation groups probably a subgroup centralizer could be faster.

In any case, this should be faster, better etc. with proper recognition.

hulpke:
Looking at the example, the problem seems to be in Core(G,C) when C is a transitive subgroup of G. If G is the full symmetric group there is no quick reduction, so this takes a bit of time. I have added a silly heuristic (prefer intransitive centralizers) that avoids this bad case.

Issue #260, opened 2015-09-26
Improve the test setup


fingolfin:
This was prompted by some questions on PR #248. I think our test setup can be improved. There are some things with it that bother me; and some ideas for improvements; and I hope others have more ideas.
### Speed

My understanding so far was that `make testinstall` is meant to run fast. It is not clear what that should mean.... `testinstall.g` claims this: "The test requires about 750MB of memory and runs about one minute on an Intel Core 2 Duo / 2.53 GHz machine."

For me, `testinstall` runs in slightly more than two minutes on my 2012er MBP, which has a 2.3 GHz Intel Core i7 -- so this is a much faster machine than in the comment I just quoted, yet it takes longer than the promised one minute.

I think staying below one minute on some fixed reference machine is a good idea. The faster these tests are, the easier it is to convince devs to regularly run them. Let's make them faster then!

For starters, `make testinstall` (attempts to) run the test twice; the second time with "LoadAllPackages()". Can we please get rid of that? For me it always errors out anyway, and the chance for that increases if you have various in-development packages floating around.

I realize that this _might_ be a useful test for Jenkins, but then there should be a second targets `testinstall_all_packages` or so.

Going beyond that, I'd actually prefer if the tests run in, say, under 10 seconds, but that might be unrealistic. Still, I wonder whether e.g. `trans.tst` and `grppcnrm.tst` (the two slowest tests on my system) really "need" to do everything they do? Maybe they could be split into a "quick" version in "testinsall", and a "long" version in "teststandard"?
### Memory

We run all tests with `-m 100m`. Why? I never would run my GAP with so little memory, it just slows things down. Why don't we just use whatever default gap.sh runs with? Surely that default should be "good" for everybody on average, and since it is what most users will end up with by default, we should use it, no? If we are concerned about comparing timings (which we shouldn't), or about reproducibility, we could simply let the tests print the active settings at the start, no?
### Names of the test suites

Perhaps rename `testinstall` to `testquick` and `teststandard` to `testall`? Or even `test-quick` and `test-all`, which I find much more readable?

alex-konovalov:
Thank you for initiating this discussions. Just some details on where the current setup comes from.

The claim about 1 minute runtime was made a while ago, and after that some more tests were added (and even more things changed after that tests were rearranged). So it's not a speed regression. This of course should be updated. 750 MB memory comes from the `-o` option used in the Makefile - it is less than default, but we want it to be conservative to catch for regressions in memory usage.

[...]

PR #2684, opened 2018-08-06
Add NormalSubgroups methods for symmetric and alternating permutation groups

mtorpey:
The normal subgroups of a symmetric group are well-known and simply described: the trivial group, the Klein 4-group (if n=4), the alternating group, and the symmetric group itself.  An alternating group's normal subgroups are even simpler: the same but without the symmetric group.  The `NormalSubgroups` attribute in GAP should reflect this knowledge by having special methods for permutation groups that are known to be symmetric or alternating.

At the moment, some laborious work is done which means that calling, for example, `NormalSubgroups(SymmetricGroup(40))`, takes several seconds, while larger examples run for a very long time.  This PR introduces simple methods that return the appropriate subgroups.

Note that this only applies to *permutation* groups, since it is easy to describe subgroups based on their moved points.  Note also that this is **not** a fix for Issue #2683, a bug that probably applies to non-symmetric groups as well.

### DIFF
--- a/lib/gpprmsya.gi
+++ b/lib/gpprmsya.gi
@@ -2447,6 +2447,30 @@ InstallMethod( RadicalGroup, "symmetric", true,
 InstallMethod( RadicalGroup, "alternating", true,
     [ IsNaturalAlternatingGroup and IsFinite], 0,RadicalSymmAlt);
 
+InstallMethod(NormalSubgroups,
+"for a symmetric group",
+[IsSymmetricGroup],
+RankFilter(IsPermGroup),
+function(S)
+  if SymmetricDegree(S) <= 4 then
+    # S is soluble, so this includes the trivial group (and Klein 4)
+    return DerivedSeriesOfGroup(S);
+  fi;
+  # DerivedSubgroup is the alternating group
+  return [S, DerivedSubgroup(S), TrivialSubgroup(S)];
+end);
+
+InstallMethod(NormalSubgroups,
+"for an alternating group",
+[IsAlternatingGroup],
+RankFilter(IsPermGroup),
+function(A)
+  if AlternatingDegree(A) <= 4 then
+    # S is soluble, so this includes the trivial group (and Klein 4)
+    return DerivedSeriesOfGroup(A);
+  fi;
+  return [A, TrivialSubgroup(A)];
+end);
 
 #############################################################################
 ##
--- a/tst/testinstall/opers/NormalSubgroups.tst
+++ b/tst/testinstall/opers/NormalSubgroups.tst
@@ -1,0 +1,113 @@
+gap> START_TEST("NormalSubgroups.tst");
+
+# Natural symmetric groups
+gap> NormalSubgroups(SymmetricGroup(0)) = [Group(())];
+true
+gap> NormalSubgroups(SymmetricGroup(1)) = [Group(())];
+true
+gap> Set(NormalSubgroups(SymmetricGroup(2))) =
+> Set([Group(()), SymmetricGroup(2)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup(3))) =
+> Set([Group(()), AlternatingGroup(3), SymmetricGroup(3)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup(4))) =
+> Set([Group(()), Group((1,2)(3,4), (1,3)(2,4)),
+>  AlternatingGroup(4), SymmetricGroup(4)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup(5))) =
+> Set([Group(()), AlternatingGroup(5), SymmetricGroup(5)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup(6))) =
+> Set([Group(()), AlternatingGroup(6), SymmetricGroup(6)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup(15))) =
+> Set([Group(()), AlternatingGroup(15), SymmetricGroup(15)]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup([3,7,4,11,70]))) =
+> Set([Group(()), AlternatingGroup([3, 4, 7, 11, 70]),
+>      SymmetricGroup([3, 4, 7, 11, 70])]);
+true
+gap> Set(NormalSubgroups(SymmetricGroup([7,4,13,21]))) =
+> Set([Group(()), Group((4,7)(13,21), (4,13)(7,21)),
+>      AlternatingGroup([4, 7, 13, 21]), SymmetricGroup([4, 7, 13, 21])]);
+true
+gap> NormalSubgroups(SymmetricGroup([42])) = [Group(())];
+true
+gap> NormalSubgroups(SymmetricGroup([])) = [Group(())];
+true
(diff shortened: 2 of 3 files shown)


lib/gpprmsya.gi
```diff
@@ -2447,6 +2447,42 @@ InstallMethod( RadicalGroup, "symmetric", true,
...
     [ IsNaturalAlternatingGroup and IsFinite], 0,RadicalSymmAlt);
 
+InstallMethod(NormalSubgroups,
+"for a perm group isomorphic to the symmetric group",
+[IsPermGroup and IsSymmetricGroup],
+function(G)
+  local pts, list;
+  pts := MovedPoints(G);
+  list := [Group(())];  # Trivial group
+  if Length(pts) = 4 then  # Klein 4-group
```
fingolfin: This check is only valid if the symmetric group is in its "natural" form, as tested by `IsNaturalSymmetricGroup`. But  `Sym(4)` also has e.g. a permutation representation acting on 24 points (i.e. the regular module):
```
gap> G:=Image(RegularActionHomomorphism(SymmetricGroup(4)));;
gap> IsSymmetricGroup(G);
true
gap> IsNaturalSymmetricGroup(G);
false
gap> NrMovedPoints(G);
24
```

lib/gpprmsya.gi
```diff
@@ -2447,6 +2447,42 @@ InstallMethod( RadicalGroup, "symmetric", true,
...
+  fi;
+  if Length(pts) > 1 then
+    Add(list, G);  # Symmetric group
+  fi;
+  return list;
+end);
+
+InstallMethod(NormalSubgroups,
+"for a perm group isomorphic to the alternating group",
+[IsPermGroup and IsAlternatingGroup],
```
fingolfin: I guess something similar could also be installed for `IsSimpleGroup`... Not saying you should, just thinking out loud. The case with few moved points are mostly trivial to compute anyway.

lib/gpprmsya.gi
```diff
@@ -2447,6 +2447,42 @@ InstallMethod( RadicalGroup, "symmetric", true,
...
+function(G)
+  local pts, list;
+  pts := MovedPoints(G);
+  list := [Group(())];
+  if Length(pts) = 4 then
```
fingolfin: Again, this test only is sufficient if you restrict to `IsNaturalAlternatingGroup`. Or switch the condition of this method to `IsSimpleGroup`.

fingolfin:
BTW, It might be better to instead modify `NormalSubgroupsCalc` to teach it about these cases. Probably @hulpke has something to say about that, too... My personal impression here is that this is something where group recognition would help for the full "proper" solution, but @hulpke will know far better.

mtorpey:
@ChrisJefferson, you're right.  Amended.

mtorpey:
In response to @fingolfin's comments, I've gone for `IsNaturalSymmetric/AlternatingGroup` as the filter since, as he points out, this algorithm only applies to groups that are the symmetric group on their set of moved points.  I've also made the tests less display-focused, which should allow them to pass regardless of which packages are loaded.

I don't know anything about `NormalSubgroupsCalc`, so I'll let others consider that. :)

markuspf:
If a group knows `IsSymmetric`, using `DerivedSubgroup` as the one proper normal subgroup would suffice (indeed to compute `IsSymmetric`, GAP computes the derived subgroup of `G`),

randomised recognition I can only find in `recog`.

Adding an additional case to `NormalSubgroupCalc` seems to be a rather complicated fix to this fairly simple problem.

fingolfin:
@markuspf I am not sure I understand what you are trying to say with the first paragraph of your last comment. Note that Sym(4) has another normal subgroup besides it derived subgroup.

markuspf:
@fingolfin, I was ignoring "trivial" cases.

Let's wait for @hulpke to comment.

fingolfin:
Observation: one could simply install `DerivedSeries` as a method in all these cases (does not require "natural" either)

fingolfin:
(Well, almost: the trivial group needs to be added in the nonsolvable cases)

markuspf:
Ok, now I see why one would want to integrate this into the `NormalSubgroupsCalc`: It'd work for *any* permutation group, even if it doesn't know its natural symmetric (or with improvements isomorphic to a symmetric group).

I think generalising the proposed code to at least work for any group that knows its symmetric is a fairly good thing to do (as @hulpke says this code wouldn't interfere with more sophisticated code that works for examples that don't even know that they're symmetric)

Furthermore, the test for `IsSymmetric` seems fairly cheap, essentially computing the derived subgroup and excluding some pesky counterexamples in the case of `n=6`, but now I am probably veering towards what the code in `NormalSubgroupsCalc` would do anyway.

Incidentally, @hulpke, is here a reference/publication that describes how `NormalSubgroupsCalc` works? That function looks quite daunting at first glance.

mtorpey:
I've updated this to use `DerivedSeries` and `DerivedSubgroup`, and therefore to work even for non-natural symmetric and alternating groups.  For now I've left everything in special methods, since I don't know a lot about `NormalSubgroupsCalc`.

In order to have this method selected over the method for `IsPermGroup` (which redirects to `NormalSubgroupsCalc`), I've added `RankFilter(IsPermGroup)` to the methods' priority.  Does this seem reasonable?

@hulpke: Perhaps I went a little overboard writing tests.  I'll rein it in a bit in future! :)

tst/testinstall/opers/NormalSubgroups.tst
```diff
@@ -0,0 +1,113 @@
...
+gap> IsNaturalSymmetricGroup(G);
+true
+gap> Set(NormalSubgroups(G)) =
+> Set([Group(()), Group((3,5)(7,11), (3,7)(5,11)),
+>      AlternatingGroup([3, 5, 7, 11]), SymmetricGroup([3, 5, 7, 11])]);
+true
+
+# Non-natural symmetric groups
+gap> S4 := Group((1,2,3,4), (1,2));;
+gap> hom := ActionHomomorphism(g, Arrangements([1..4], 4), OnTuples);;
```
fingolfin: change `g` to `S4` here (else the test fails when run on its own).

mtorpey: Thank you!  It's always the little things...

fingolfin:
I actually think that the tests added here are very good, and wish we had similarly thorough tests for many more operands. In particular, they cover the cases where the set we act on is *not* `[1..n]` (which is often forgotten), deals with edge cases (here: small n) etc.. However, the whole test runs in 250ms on my laptop, which is indeed a bit long (I mean, it's not much, but if you do the same thing for 1000 functions, that turns into 4 minutes). Simple solution: drop the two tests for degree 100 (or replace them by, say, degree 15), then it goes down to 20 milliseconds. Sure, those cases were what motivated you originally, but if we wanted to check for performance regressions, we really need to create a dedicated test setup for that first (see also issue #260).

Anyway: tests are failing because the changes here affect the example in the manual, in `lib/grp.gd:1870`.

hulpke:
@markuspf 
The algorithm in `NormalSubgroupsCalc` is described in  my 1998 ISSAC paper (
http://www.math.colostate.edu/~hulpke/paper/normalsub.pdf )

PR #2686, opened 2018-08-08
Improve bicoset code

fingolfin:
- add `bicoset` to the reference manual index
- clarify documentation for `IsBiCoset` ("for example" -> "if and only if")
- improve `IsBiCoset` by replacing `r*x/r` with `x^r`
- improve multiplication of right cosets to only transfer knowledge about whether this is a bicoset if this was already known for the right operand, instead of unconditionally computing it; but also transfer if it is not (a bicoset times a non-bicoset is not a bicoset)
- replace `TryNextMethod` in multiplication method for right cosets with helpful error message, just like we do in the inversion method
- improve error message when inverting a right coset which is not a bicoset
- add a test that ensures we indeed produce an error when multiplying or inverting right cosets for which these operations are not defined

This addresses all my remarks on PR #2666 plus some more.

### DIFF
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -634,7 +634,7 @@ function(c)
 local s,r;
   s:=ActingDomain(c);
   r:=Representative(c);
-  return ForAll(GeneratorsOfGroup(s),x->r*x/r in s);
+  return ForAll(GeneratorsOfGroup(s),x->x^r in s);
 end);
 
 InstallMethodWithRandomSource(Random,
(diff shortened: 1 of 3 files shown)


lib/csetgrp.gi
```diff
@@ -667,11 +667,11 @@ function(a,b)
 local c;
   if ActingDomain(a)<>ActingDomain(b) then TryNextMethod();fi;
   if not IsBiCoset(a) then # product does not require b to be bicoset
-    TryNextMethod();
+    ErrorNoReturn("right cosets can only be multiplied if the left operand is a bicoset");
```
hulpke: Making this an error will hard block any future attempt to define multiplication of ordinary cosets (e.g. as multiplication of sets) in a more general way through other methods. Is this deliberate?

fingolfin: This was quite deliberate, i.e., I thought quite a while about whether to change this into an error; or whether to instead change the `Inverse` method to use `TryNextMethod()` instead of `Error` -- in any case, I think they should match.

However, in the end I think a clear helpful error message is far better than a "method not found", just to support a hypothetical future extension which may never come. Besides, we actually don't stop anybody from adding one: they just have to rank their method higher than ours (and make sure to stay compatibly / invoke `TryNextMethod()` if needed). Finally, we can of course always change the error into something else if it ever becomes necessary.

PR #2688, opened 2018-08-09
Updates to the Julia GC integration

rbehrends:
This brings the implementation in line with the most recent API changes for the Julia GC extensions and cleans up some code.

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -564,6 +564,11 @@ BindGlobal( "ShowKernelInformation", function()
     fi;
     Add( config, str );
   fi;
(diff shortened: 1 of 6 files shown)


fingolfin:
Since those Julia changes are not yet quite stable, let's wait with merging this for now.

For the status of the Julia side of things, see here: https://github.com/JuliaLang/julia/pull/28368

fingolfin:
Memo to @rbehrends and myself: we should also modify `init.g` to print something like `with Julia GC` or even `with Julia 1.0` or so (and then export from the kernel `JULIA_VERSION`, just like `GMP_VERSION` is exported).

PR #2691, opened 2018-08-10
kernel: fix GAP_STATIC_ASSERT fallback code

fingolfin:
If `_Static_assert` resp. `static_assert` are not available, we fall back to a
trick involving a typedef that's invalid if the assertion fails.

In order to avoid clashes with repeating (identical) typedefs, we tried to
insert the line number (via the `__LINE__` macro), but that did not actually
work due to the way the C preprocessor works. To fix this, we need to use two
helper macros which ensure `__LINE__` get evaluated before being concatenated
to another string.

I am not sure whether it's worth to mention this in the release notes; but then we *did* get a report involving this problem, so perhaps it is... In any case, I've not yet added any "release notes" labels.

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -36,8 +36,10 @@
 #else
 // If the compiler does not support _Static_assert resp. static_assert,
 // fall back to a hack (the error message is a bit ugly in that case).
+#define _intern_CONCAT_(X, Y)  X ## Y
+#define _intern_CONCAT(X, Y)  _intern_CONCAT_(X,Y)
 #define GAP_STATIC_ASSERT(cond, msg) \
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Good fix. I should have spotted this originally, as I've had exactly this kind of stupid preprocessor issue before.

alex-konovalov:
@ChrisJefferson we are getting further reports about this problem and advise to apply the patch to GAP 4.9.2. Should this be backported into stable-4.9 branch then?

fingolfin:
If there is going to be a 4.9.3 release, then we should definitely backport this.

PR #2693, opened 2018-08-11
kernel: simplify bool.c and add test cases

fingolfin:
Since `Undefined` should never leak out of low level code, there is no
point in adding print support for it. Also simplify EqBool and LtBool.

This should get code coverage to almost 100% (only exception is the `PrintBool` case for "very strange boolean value"

### DIFF
--- a/src/bool.c
+++ b/src/bool.c
@@ -98,9 +98,6 @@ void PrintBool (
     else if ( bool == Fail ) {
         Pr( "fail", 0L, 0L );
     }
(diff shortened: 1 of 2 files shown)


fingolfin:
How exactly would we turn that into an error, though? I guess one idea would be to make some functions in bool.c stricter about checking for "weird" boolean values? E.g. the print function could actually trigger an error instead of printing `<<very strange boolean value>>`. Not sure it's worth the effort, though?

PR #2694, opened 2018-08-11
kernel: remove some unused blist/plist code

fingolfin:
* remove PLEN_SIZE_BLIST (unused)
* remove IS_BLIST_REP_WITH_COPYING (use IS_BLIST_REP instead)
* change IS_PLIST_OR_POSOBJ to not support COPYING flag

For the latter two, it is unclear to me why they ever dealt with COPYING
objects: as far as I can tell, such objects are never accessed with
anything but raw (CONST_)ADDR_OBJ() calls.

@ChrisJefferson You added those, perhaps you remember the reason?


fingolfin:
As for the plist code, I figured it out: the problem there never was `COPYING`, it was `TESTING` -- but back then, the two were indistinguishable. When `KTNumPlist` was rewritten to use an object flag to implement `TESTING`, instead of re-using th COPYING tnums, the need for the check in `IS_PLIST_OR_POSOBJ` went away.

But I can't figure out why the check ever was added for blsts. I went back to 658aa8d91d9e50efafb2c80a96416ef0d98c96f9 and edited `IS_BLIST_REP_WITH_COPYING` there to not check the `COPYING` part of the tnum, and the test suite did run just fine.

ChrisJefferson:
It's quite possible I copied the PLIST stuff across to BLISTs, assuming the implementation would be the same. If it works without and no tests are failing (and, if we don't want to allow COPYING in future), it's fine to disable in my opinion.

PR #2696, opened 2018-08-11
kernel: unify forwarding pointer in list copying code

fingolfin:
This gets rid of some 2-element plists used to store forwarding pointers, which were still used by the blist, pos obj and wp obj code. Also changes some `ADDR_OBJ` to `CONST_ADDR_OBJ`

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -236,20 +236,18 @@ Obj CopyBlist (
     Int                 mut )
 {
     Obj copy;
-    Obj tmp;
 
     if (!IS_MUTABLE_OBJ(list)) {
         return list;
     }
 
     copy = DoCopyBlist(list, mut);
+
     /* leave a forwarding pointer */
-    tmp = NEW_PLIST( T_PLIST, 2 );
-    SET_LEN_PLIST( tmp, 2 );
-    SET_ELM_PLIST( tmp, 1, CONST_ADDR_OBJ(list)[0] );
-    SET_ELM_PLIST( tmp, 2, copy );
-    ADDR_OBJ(list)[0] = tmp;
+    ADDR_OBJ(list)[0] = copy;
     CHANGED_BAG(list);
+
+    /* now it is copied                                                    */
     RetypeBag( list, TNUM_OBJ(list) + COPYING );
     return copy;
 }
@@ -271,7 +269,7 @@ Obj ShallowCopyBlist ( Obj list)
 */
 Obj CopyBlistCopy(Obj list, Int mut)
 {
-    return ELM_PLIST(CONST_ADDR_OBJ(list)[0], 2);
+    return CONST_ADDR_OBJ(list)[0];
 }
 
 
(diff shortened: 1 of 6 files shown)


fingolfin:
This PR actually went a bit to far by also changing the forwarding pointer for positional/component/data objects, which luckily was revealed by a test suite failure. I now added comments that explain why for these types, the forward pointer must be indirect after all.

src/weakptr.c
```diff
@@ -823,8 +819,7 @@ void CleanObjWPObjCopy (
     Obj                 elm;            /* subobject                       */
 
     /* remove the forwarding pointer                                       */
-    ADDR_OBJ(obj)[0] = ELM_PLIST( CONST_ADDR_OBJ(obj)[0], 1 );
-    CHANGED_BAG(obj);
```
fingolfin: Note that the `CHANGED_BAG` here is dropped intentionally: the value restored to `obj` is a raw integer (the "stored length" of the wp obj, and never a bag ref.

src/blister.c
```diff
@@ -236,20 +236,18 @@ Obj CopyBlist (
...
 
     copy = DoCopyBlist(list, mut);
+
     /* leave a forwarding pointer */
-    tmp = NEW_PLIST( T_PLIST, 2 );
-    SET_LEN_PLIST( tmp, 2 );
-    SET_ELM_PLIST( tmp, 1, CONST_ADDR_OBJ(list)[0] );
-    SET_ELM_PLIST( tmp, 2, copy );
-    ADDR_OBJ(list)[0] = tmp;
+    ADDR_OBJ(list)[0] = copy;
```
ChrisJefferson: Can you just give me a brief idea what's going on here? I'm not really clear what the code was doing before, but assume you understand.

ChrisJefferson: I mean, we might just have been making some weird list for no obvious reason?

fingolfin: When we copy an object A, we must make sure that if we copy it multiple times, then we return the same "copied object" B each time (think of copying the list `[A,A]`, the result should be `[B,B]`, not `[B,C]`). To implement that, we make the copy B, then keep a forwarding pointer to the copy inside A; if asked to copy A again later during the same copy operation, we just return that forwarding pointer to B. Finally, when copying is done, restore the value overwritten by the forwarding pointer.

In this case (actually, in all?), the value overwritten is the first word / first slot of the blist. What the old code did (and the code for positional and component and data object still does) is to create a plist of length 2, which stores the old length value, plus the pointer to the copy B; then inside of A, we store a pointer to B.

However, this is not really necessary here: to restore the length of A, we can simply read it from the copy B, which after all has the same length! Thus, the code change avoid creating this temporary plist, and one indirection.

We can't do this for pos/comp/dat objs, as there the overwritten value is the type, and the type of a copy can differ (e.g. if the copy changes the mutability).

ChrisJefferson: thanks

PR #2697, opened 2018-08-13
Tweak READ_ALL_COMMANDS (yet again)

markuspf:
A (hopefully final, this has become a festering wound now), tweak to `READ_ALL_COMMANDS` to accommodate for some oddities.
The following things are changed:

 * Rename the `outputFunc` parameter to `resultCallback`
 * Introduce an extra parameter `capture` which requests the function
   to capture all output, including output that might be generated
   by the `resultCallback`
 * Make output capture independent of whether a `resultCallback` was given
 * Always return a list of length 5
 * Make the documentation a bit simpler
 * Flush the output stream before we copy the result out, otherwise we might miss some output

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -87,40 +87,43 @@ static Int READ_COMMAND(Obj *evalResult)
 *F  FuncREAD_ALL_COMMANDS( <self>, <instream>, <echo>, <capture>, <outputFunc> )
 **
 **  FuncREAD_ALL_COMMANDS attempts to execute all statements read from the
-**  stream <instream> and returns fail if the stream cannot be opened, or a list
-**  of lists, each entry of which reflects the result of the execution of one
-**  statement.
+**  stream <instream>. It returns 'fail' if the stream cannot be opened,
+**  otherwise a list of lists, each entry of which reflects the result of the
+**  execution of one statement.
 **
-**  If the parameter <echo> is True, then the statements are echoed to the
+**  If the parameter <echo> is 'true', then the statements are echoed to the
 **  current output.
 **
-**  If the parameter <capture> is True, then any output occurring during
-**  execution of a statement, including the output of <outputFunc>, is captured
-**  into a string.
+**  If the parameter <capture> is 'true', then any output occurring during
+**  execution of a statement, including the output of <outputFunc>, is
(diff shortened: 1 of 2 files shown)


src/streams.c
```diff
@@ -84,39 +84,51 @@ static Int READ_COMMAND(Obj *evalResult)
...
-**  is set. If <outputFunc> is a function, only the first and the fifth entry
-**  are set. If <outputFunc> is a function, the fourth entry of the list is
-**  <outputFunc> applied to the result, if there was one, and if the
-**  statement did not end in a dual semicolon. Furthermore, if <outputFunc>
-**  is not false, the fifth entry contains a string of all output printed
-**  while the corresponding statement was executed. If <outputFunc> is false,
-**  all intermediate output is discarded.
+**  of the statement if there was one, and unbound otherwise.
+**
+**  The third entry is true if the statement ended in a dual semicolon.
```
fingolfin: Maybe add "... and false otherwise"?

src/streams.c
```diff
@@ -84,39 +84,51 @@ static Int READ_COMMAND(Obj *evalResult)
...
+**
+**  The fourth entry contains the return value of <resultCallback> if applicable.
+**
+**  The fifth entry contains the captured output as a string, if <capture> is True.
```
fingolfin: So the lists are *not* always of length 5. Or rather, they are, but entry 5 may be unbound, which technically is an invalid plist. I guess for low-level code one may just ignore this. But it'd be easy to fix, too, both in this documentation, and in the code: In the docs, change 
> The results are returned as lists of length five
to e.g.
> The results are returned as lists of length at most five
In the code, simply remove `SET_LEN_PLIST(result, 5);` and change all `SET_ELM_PLIST` to `AssPlist`.

I took the liberty of adding a commit which implements all suggested changes (and some more for the docs). Feel free to either drop it again, or to squash it into your previous commit(s) (IMHO this whole PR could be squashed down into a single commit)

fingolfin:
I took the liberty of adding a commit which implements all suggested changes (and some more for the docs). Feel free to either drop it again, or to squash it into your previous commit(s) (IMHO this whole PR could be squashed down into a single commit)

PR #2701, opened 2018-08-14
Avoid calling `SaveCStr` with NULL

markuspf:
Don't crash after warning that the saved workspace will be corrupt.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -1700,21 +1700,19 @@ Int IsKernelFunction(Obj func)
 
 #ifdef USE_GASMAN
 
-static void SaveHandler( ObjFunc hdlr )
+static void SaveHandler(ObjFunc hdlr)
 {
-  const Char * cookie;
-  if (hdlr == (ObjFunc)0)
-    SaveCStr("");
-  else
-    {
-      cookie = CookieOfHandler(hdlr);
-      if (!cookie)
-	{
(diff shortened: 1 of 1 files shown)


markuspf:
Arguably the saved workspace will still be useless, so maybe we shold just assert?

markuspf:
Thinking about this again: when one's trying to save some results this way, it's probably better not to just crash GAP just because we can't store a function object.

src/calls.c
```diff
@@ -1700,20 +1700,19 @@ Int IsKernelFunction(Obj func)
...
+    const Char * cookie;
+    if (hdlr == (ObjFunc)0)
+        SaveCStr("");
+    else {
+        cookie = CookieOfHandler(hdlr);
+        if (!cookie) {
+            Pr("No cookie for Handler -- workspace will be corrupt\n", 0, 0);
+            SaveCStr("");
+        }
+        else
```
fingolfin: So essentially all this does is insert this `else` here, hidden among a complete reformatting of the function? I think @ChrisJefferson explicitly stated that he wanted to refuse reviewing PRs with gratuitous reformatting in them... Though personally I prefer the formatted version *shrug*

ChrisJefferson: Just quick drive-by comment, as I was reading my email. While *I* don't want to review PRs with big code formatting, as it annoys me to carefully check was has changed, I certainly have no problem if other people want to make them and review them :)

Issue #1917, opened 2017-11-16
Add a unit test framework for C code, to test e.g. ObjInt_UInt, ObjInt_Int8, ObjInt_Int


fingolfin:
It would be good to have a C unit test framework where we can test dedicated kernel functions without having to come up in a "natural" way. Right now, the best one can seems to be to write C functions `FuncTEST_SOME_CODE`, export them to GAP, and then call them from there.

A general framework could also do that, but with a single function, which then is responsible for executing the whole test suite(s).
Or perhaps some other way would do -- whatever, in the end, it just would be nice to have tests for things like `ObjInt_Int8`, which used to have bugs for edge cases in the past; so it would be really good to be able to add systematic tests for such relatively self-contained functions.

See also PR #1909 for some more relevant C functions.

ChrisJefferson:
I previously thought about doing this with a special package, although that would still require exporting the functions to the GAP level.

I started with a simple package: https://github.com/gap-packages/kerneltest . We could extend and improve this.

fingolfin:
`ObjInt_UInt` are now tested via regular .tst files, with the help of some helper kernel functions.

For libgap, we now have started to create C based tests; this essentially allows doing what this issue describes.

Issue #430, opened 2015-12-26
spurious CR at start of output in quiet mode


eduardkareletc:
Running gap as:
$ echo 'Display("test");; quit;;' |gap -q |od -N 10 -t a
0000000   cr   t   e   s   t  nl 

This shows an initial CR character in the output. In my understanding, the -q flag should have suppressed any output before the output generated by the program  (t e s t nl)". In further output from a gap program each line is terminated by NL, that's what I found inspecting the output of my program that lead to me running the test above to show this bug.

My gap version info is:
 ┌───────┐   GAP, Version 4.7.7 of 13-Feb-2015 (free software, GPL)
 │  GAP                │   http://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin13.4.0-gcc-default64

ChrisJefferson:
I don't seem able to reproduce in 4.7.8, or the current git version. I don't have 4.7.7, could you try 4.7.8 (although I don't think it will fix it...)

```
~/r/g/r/gap4r7p8 $ echo 'Display("test");; quit;;' | bin/gap.sh -q | od -N 10 -t a
0000000    t   e   s   t  nl
0000005
```

alex-konovalov:
This is triggered by some package, because this does not happen with -A option:
```
Mac-0c4de9c1b942:gap4r7p9 alexk$ echo 'Display("test");; quit;;' | bin/gap.sh -q | od -N 10 -t a
0000000   cr   #   I  sp  sp   Y   o   u  sp   m                        
0000012
Mac-0c4de9c1b942:gap4r7p9 alexk$ echo 'Display("test");; quit;;' | bin/gap.sh -q -A -r | od -N 10 -t a
0000000    t   e   s   t  nl                                            
0000005
Mac-0c4de9c1b942:gap4r7p9 alexk$ echo 'Print("test");; quit;;' | bin/gap.sh -q -A -r | od -N 10 -t a
0000000    t   e   s   t                                                
0000004
Mac-0c4de9c1b942:gap4r7p9 alexk$ echo 'Print("test");; quit;;' | bin/gap.sh -q | od -N 10 -t a
0000000   cr   #   I  sp  sp   Y   o   u  sp   m                        
0000012
```

alex-konovalov:
This is triggered by some package, because this does not happen with -A option:
```
$ echo 'Display("test");; quit;;' | bin/gap.sh -q -r| od -N 10 -t a
[...]

PR #2702, opened 2018-08-15
Add GAP kernel API (aka libgap)

markuspf:
This pull request is the next logical step in providing GAP as a dynamically loadable library (known by some as LibGAP).

I squashed a whole history of commits developed by @sebasguts, @fingolfin, and me into a single commit because it's not that big anymore. If there's certain bits you'd like committed separately, speak now or forever hold your peace.

It would be good if @sebasguts (for @oscar-system) and @nthiery (for SageMath) as the main "clients" could have a look to see whether this is going towards what they are hoping for.

For the future there needs to be more documentation, and more thorough testing, and a more comprehensive API.

For a future pull-request I am thinking about making a command-line-switch `--disable-shell` which disables the repl (basically removing the need for `IsLIBGAP`, at least right now, and I can't think of another usecase for this particular constant.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -16,9 +16,6 @@
 all: gap$(EXEEXT) gac
 .PHONY: all
 
-libgap: libgap.la
-.PHONY: libgap
-
 # Backwards compatibility: add "default" target as alias for "all"
 default: all
 .PHONY: default
@@ -64,6 +61,9 @@ SOURCES += src/intfuncs.c
 SOURCES += src/intrprtr.c
 SOURCES += src/io.c
 SOURCES += src/iostream.c
+ifeq ($(HPCGAP),no)             # we don't support a kernel API in HPC-GAP atm
+SOURCES += src/libgap-api.c
+endif
 SOURCES += src/listfunc.c
 SOURCES += src/listoper.c
 SOURCES += src/lists.c
@@ -120,7 +120,6 @@ ifeq ($(HPCGAP),yes)
   SOURCES += src/hpc/tls.c
   SOURCES += src/hpc/traverse.c
 endif
-
 
 ########################################################################
 # Preprocessor flags
@@ -389,12 +388,15 @@ gap$(EXEEXT): libgap.la cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-OBJS
 
 else
 
+all: libgap.so
+libgap.so: symlinks libgap.la
+	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) -shared $(OBJS) $(GAP_LIBS) -o $@
+
 # Linking rule and dependencies for the main gap executable
 gap$(EXEEXT): $(OBJS) cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-OBJS
 	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) $(OBJS) $(GAP_LIBS) -o $@
 
 endif
-
 
 ########################################################################
 # The "docomp" target regenerates the various src/c_*.c files, and
@@ -458,6 +460,7 @@ clean:
 	rm -rf obj
 	rm -f gap$(EXEEXT) gac ffgen
 	rm -f libgap.la
+	rm -f libgap.so
 	rm -f gen/gap_version.c
 	rm -f doc/wsp.g
 	rm -f cnf/GAP-{CFLAGS,CPPFLAGS,LDFLAGS,LIBS,OBJS}
@@ -985,6 +988,12 @@ testbugfix: all
 	( echo 'SetUserPreference("UseColorsInTerminal",false); LoadAllPackages(); \
           ReadGapRoot( "tst/testbugfix.g" );' | $(TESTGAP) | \
             tee `date -u +dev/log/testbugfix2_%Y-%m-%d-%H-%M` )
+
+testlibgap: libgap.la obj/tst/testlibgap/basic.lo 
+	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) obj/tst/testlibgap/basic.lo libgap.la -o test-libgap
+	./test-libgap -A -l $(top_srcdir) -m 32m -q -T --nointeract > basic.out
+	diff $(top_srcdir)/tst/testlibgap/basic.expect basic.out
+.PHONY: testlibgap
 
 coverage:
 	gcov -o . $(SOURCES)
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -1,0 +1,62 @@
+// LibGAP API - API for using GAP as shared library.
+
+#include "libgap-api.h"
+
+#include "bool.h"
+#include "opers.h"
+#include "calls.h"
+#include "gapstate.h"
+#include "gvars.h"
+#include "lists.h"
+#include "streams.h"
+#include "stringobj.h"
+
+//
+// Setup and initialisation
+//
+void GAP_Initialize(int          argc,
+                    char **      argv,
+                    char **      env,
+                    CallbackFunc markBagsCallback,
+                    CallbackFunc errorCallback)
+{
+    InitializeGap(&argc, argv, env);
+    SetExtraMarkFuncBags(markBagsCallback);
+    STATE(JumpToCatchCallback) = errorCallback;
+}
+
+
+// Combines GVarName and ValGVar. For a given string, it returns the value
+// of the gvar with name <name>, or NULL if the global variable is not
+// defined.
+Obj GAP_ValueGlobalVariable(const char * name)
+{
+    UInt gvar = GVarName(name);
+    // TODO: GVarName should never return 0?
+    if (gvar != 0) {
+        return ValGVar(gvar);
+    }
+    else {
+        return NULL;
+    }
+}
+
+//
+// Evaluate a string of GAP commands
+//
+// To see an example of how to use this function
+// see tst/testlibgap/basic.c
+//
+Obj GAP_EvalString(const char * cmd)
+{
+    Obj instream;
+    Obj res;
+    Obj viewObjFunc, streamFunc;
+
+    streamFunc = GAP_ValueGlobalVariable("InputTextString");
+    viewObjFunc = GAP_ValueGlobalVariable("ViewObj");
+
+    instream = DoOperation1Args(streamFunc, MakeString(cmd));
+    res = READ_ALL_COMMANDS(instream, False, True, viewObjFunc);
+    return res;
+}
--- a/tst/testlibgap/basic.c
+++ b/tst/testlibgap/basic.c
@@ -1,0 +1,77 @@
+/*
+ * Small program to test libgap linkability and basic working
+ */
+#include <stdio.h>
+#include <unistd.h>
+#include <compiled.h>
+#include <libgap-api.h>
+extern char ** environ;
+
+UInt GAP_List_Length(Obj list)
+{
+    return LEN_LIST(list);
+}
+
+Obj GAP_List_AtPosition(Obj list, Int pos)
+{
+    return ELM_LIST(list, pos);
+}
+
+UInt GAP_String_Length(Obj string)
+{
+    return GET_LEN_STRING(string);
+}
+
+Int GAP_String_GetCString(Obj string, Char * buffer, UInt n)
+{
+    UInt len;
+
+    if (IS_STRING(string)) {
+        if (!IS_STRING_REP(string))
+            string = CopyToStringRep(string);
+        len = GET_LEN_STRING(string) + 1;
+        if (len >= n)
+            len = n - 1;
+        // Have to use mempcy because GAP strings can contain
+        // \0.
+        memcpy(buffer, CSTR_STRING(string), len);
+        if (len == n - 1)
+            buffer[n] = '\0';
+        return 1;
+    }
+    return 0;
+}
+
+void test_eval(const char * cmd)
+{
+    Obj  res, ires;
+    Int  rc, i;
+    Char buffer[4096];
+    printf("gap> %s\n", cmd);
+    res = GAP_EvalString(cmd);
+    rc = GAP_List_Length(res);
+    for (i = 1; i <= rc; i++) {
+        ires = GAP_List_AtPosition(res, i);
+        if (GAP_List_AtPosition(ires, 1) == True) {
+            GAP_String_GetCString(GAP_List_AtPosition(ires, 5), buffer,
+                                  sizeof(buffer));
+            printf("%s\n", buffer);
+        }
+    }
+}
+int main(int argc, char ** argv)
+{
+    printf("# Initializing GAP...\n");
+    GAP_Initialize(argc, argv, environ, 0L, 0L);
+    CollectBags(0, 1);    // full GC
+    test_eval("1+2+3;");
+    test_eval("g:=FreeGroup(2);");
+    test_eval("a:=g.1;");
+    test_eval("b:=g.2;");
+    test_eval("lis:=[a^2, a^2, b*a];");
+    test_eval("h:=g/lis;");
+    test_eval("c:=h.1;");
+    test_eval("Set([1..1000000], i->Order(c));");
+    printf("# done\n");
+    return 0;
+}
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ doc/gapmacrodoc.idx
 /hpcgap/ward
 
 /builds/
+
+/libgap.la
+/.libs
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,6 +108,9 @@ matrix:
     # test error reporting and compiling (quickest job in this test suite)
     - env: TEST_SUITES="testspecial test-compile"
 
+    # test libgap
+    - env: TEST_SUITES="testlibgap"
+
 script:
   - gcov --version
   - bash etc/ci-prepare.sh && bash etc/ci.sh
--- a/etc/ci.sh
+++ b/etc/ci.sh
@@ -177,6 +177,10 @@ GAPInput
 
     ;;
 
+  testlibgap)
+    make testlibgap
+    ;;
+
   *)
     if [[ ! -f  $SRCDIR/tst/${TEST_SUITE}.g ]]
     then
--- a/lib/streams.gi
+++ b/lib/streams.gi
@@ -234,14 +234,6 @@ function( stream )
     CloseStream(stream);
 end );
 
-BindGlobal("LIBGAP_EvalString",
-function(string)
-   local instream, obj;
-   instream := InputTextString(string);
-   obj := READ_ALL_COMMANDS(instream, 0, false);
-   return obj;
-end);
-
 
 #############################################################################
 ##
--- a/src/libgap-api.h
+++ b/src/libgap-api.h
@@ -1,0 +1,21 @@
+// LibGAP API - API for using GAP as shared library.
+
+#ifndef LIBGAP_API_H
+#define LIBGAP_API_H
+
+#include "gap.h"
+
+typedef void (*CallbackFunc)(void);
+
+// Initialisation and finalization
+
+void GAP_Initialize(int          argc,
+                    char **      argv,
+                    char **      env,
+                    CallbackFunc markBagsCallback,
+                    CallbackFunc errorCallback);
+
+Obj GAP_ValueGlobalVariable(const char * name);
+Obj GAP_EvalString(const char * cmd);
+
+#endif
--- a/tst/testlibgap/basic.expect
+++ b/tst/testlibgap/basic.expect
@@ -1,0 +1,18 @@
+# Initializing GAP...
+gap> 1+2+3;
+6
+gap> g:=FreeGroup(2);
+<free group on the generators [ f1, f2 ]>
+gap> a:=g.1;
+f1
+gap> b:=g.2;
+f2
+gap> lis:=[a^2, a^2, b*a];
+[ f1^2, f1^2, f2*f1 ]
+gap> h:=g/lis;
+<fp group on the generators [ f1, f2 ]>
+gap> c:=h.1;
+f1
+gap> Set([1..1000000], i->Order(c));
+[ 2 ]
+# done


tst/testlibgap/0001.c
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+g:=FreeGroup(2);
+a:=g.1;
+b:=g.2;
+lis:=[a^2, a^2, b*a];
+h:=g/lis;
+c:=h.1;
+Set([1..300000], i->Order(c));
+ */
```
fingolfin: Is this supposed to be the same code as in the test above, which then is supposed to be comparable via copy&paste into a GAP session?

Then I think this is a bad idea, as the two will quickly run out of sync. Of course it's not my problem for now, so I'll not overly complain, but why not do this instead: Put this code into a separate `.g` file, then use fopen + fgets + fclose (or open + gets + close) to process that file line bby line from within this .c file.
Or, if you prefer to avoid dealing with opening files, just read from stdin line-by-line, and change the test script to pipe the `.g` file into the test executable.

tst/testlibgap/0001.c
```diff
@@ -0,0 +1,57 @@
...
+
+void test_eval(const char * cmd)
+{
+    Obj res, ires;
+    Int rc, i;
+
+    printf("Input:\n%s", cmd);
+    res = GAP_EvalString(cmd);
+    rc = GAP_LenPList(res);
+    for(i=1;i<=rc;i++) {
```
fingolfin: clang-format?

tst/testlibgap/0001.c
```diff
@@ -0,0 +1,57 @@
...
+        ires = GAP_ElmPList(res, i);
+        if( GAP_ElmPList(ires, 1) == True )
+            printf("Output:\n%s\n", CSTR_STRING(GAP_ElmPList(ires, 5)));
+    }
```
fingolfin: Would it perhaps be helpful to print the output in the same format as a `.tst` file?
(Fine by me to leave it as is, just wondering)

Makefile.rules
```diff
@@ -995,6 +1002,14 @@ testbugfix: all
           ReadGapRoot( "tst/testbugfix.g" );' | $(TESTGAP) | \
             tee `date -u +dev/log/testbugfix2_%Y-%m-%d-%H-%M` )
 
+obj/libgap-test-0001.lo: $(top_srcdir)/tst/testlibgap/0001.c
+	$(QUIET_CC)$(COMPILE) $(DEPFLAGS) $(GAP_CFLAGS) $(CC_EXTRA_FLAGS) $(GAP_CPPFLAGS) -c $< -o $@
```
fingolfin: What is `CC_EXTRA_FLAGS`? Is this an accidental left-over, or meant to introduce C flags just for `testlibgap`?

Other than that, with PR #2704 this build rule could go, and the general *.c build rule would take of compiling this C file

Makefile.rules
```diff
@@ -995,6 +1002,14 @@ testbugfix: all
...
+
+testlibgap: libgap.la obj/libgap-test-0001.lo
```
fingolfin: `testlibgap` needs to be marked as `.PHONY`.

Makefile.rules
```diff
@@ -995,6 +1002,14 @@ testbugfix: all
           ReadGapRoot( "tst/testbugfix.g" );' | $(TESTGAP) | \
             tee `date -u +dev/log/testbugfix2_%Y-%m-%d-%H-%M` )
 
+obj/libgap-test-0001.lo: $(top_srcdir)/tst/testlibgap/0001.c
+	$(QUIET_CC)$(COMPILE) $(DEPFLAGS) $(GAP_CFLAGS) $(CC_EXTRA_FLAGS) $(GAP_CPPFLAGS) -c $< -o $@
+
+testlibgap: libgap.la obj/libgap-test-0001.lo 
+	$(QUIET_LINK)$(LINK) obj/libgap-test-0001.lo libgap.la -o test-libgap
```
fingolfin: I would recommend to always use `$<`, `$@` etc. instead of duplicating the names of inputs and outputs, for the usual reasons why one should avoid code duplication. So here that could become:
```
testlibgap: libgap.la obj/libgap-test-0001.lo
	$(QUIET_LINK)$(LINK) $+ -o test-libgap
	...
```

Additional thought: all other files related to this test are named `0001.FOO`, would you mind if the executable then was named `0001` (or `0001.exe` or `0001.prog` or whatever)? That would then suggest a clear why how to generalize this to more tests (i.e. just provide a list of stem names for tests; then for each such name "FOO", compile `FOO.c` into `FOO` (or `FOO.prog` or whatever), run it with input file `FOO.in` (if present), send output to `FOO.out` and diff it against `FOO.expect`.

And most importantly, it would enable us to resolve issue #1917, i.e., start providing kernel based tests for some features which are otherwise difficult to test (such as `ObjInt_Int` and its various siblings).

I can make these changes later, of course, simply want to hear what others think of the general idea.

src/gasman.h
```diff
@@ -1050,4 +1050,10 @@ void *AllocateMemoryBlock(UInt size);
 Int enableMemCheck(Char ** argv, void * dummy);
 #endif
 
+/*
+ * If not 0 this function will be called in
+ * CollectBags to allow users of libgap to mark bags
+ */
+extern TNumExtraMarkFuncBags ExtraMarkFuncBags;
```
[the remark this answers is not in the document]
markuspf: See #2706

src/libgap-api.c
```diff
@@ -0,0 +1,127 @@
...
+                    CallbackFunc errorCallback)
+{
+    InitializeGap(&argc, argv, env);
+    ExtraMarkFuncBags = markBagsCallback;
+    STATE(JumpToCatchCallback) = errorCallback;
+}
+
+void GAP_Finalize(void)
+{
+    FinishBags();
```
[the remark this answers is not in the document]
markuspf: The intention is to finish the GAP "session", and be able to `GAP_Initialize` again, having a fresh GAP session. 

Of course better yet, in future, if we have all state encapsulated, we'd be able to run multiple GAPs in one process.

That said, if noone sees a use for this, we can just leave it out for now.

src/libgap-api.c
```diff
@@ -0,0 +1,127 @@
...
+
+// Setup and initialisation
+
+void GAP_Initialize(int          argc,
+                    char **      argv,
+                    char **      env,
+                    CallbackFunc markBagsCallback,
+                    CallbackFunc errorCallback)
+{
+    InitializeGap(&argc, argv, env);
```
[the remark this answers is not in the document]
markuspf: That would be my preferred goal. Figuring out what the correct API is is the current challenge.

src/libgap-api.c
```diff
@@ -0,0 +1,127 @@
+// LibGAP API - API for using GAP as shared library.
+
+#include "libgap-api.h"
```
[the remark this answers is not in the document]
markuspf: There used to be a files `libgap_internal.h`, `libgap.h`, `libgap.c`; Also I'd probably want to rename the file at some point anyway, so I didn't really bother coming up with a particularly good name.

Of course `gap.h` would be the best choice...

src/libgap-api.c
```diff
@@ -0,0 +1,127 @@
...
+#include "bool.h"
+#include "calls.h"
+#include "gapstate.h"
+#include "gvars.h"
+#include "intobj.h"
+#include "opers.h"
+#include "plist.h"
+#include "stringobj.h"
+
+Obj FuncREAD_ALL_COMMANDS(Obj self, Obj instream, Obj echo, Obj capture, Obj resultCallback);
```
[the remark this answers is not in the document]
markuspf: See #2707

src/libgap-api.c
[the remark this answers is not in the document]
sebasguts: I do not think we would need this in OSCAR for now, so I think it could be removed.

src/libgap-api.c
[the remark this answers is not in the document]
sebasguts: Exactly.

There are two things to concern: How stable do we want the API to be? Having separate function allows us to have a different version for the API than for GAP itself. This is imho a pro for separate functions.

I think using the kernel functions directly would really force us to prefix them all. I am not against this, but it might break a lot of code.

Makefile.rules
markuspf: This is, I believe, an accidental leftover from the history of trying to get libgap into `master`

tst/testlibgap/0001.c
markuspf: In principle the test only needs to execute one string, wherever it comes from. If we can get the kernel tests to work (#1917).

This is not a particularly good test, but the initial testing example that I either copied from SageMath or wrote in France 18 months ago...

Real tests would surely want to make sure that the (to be defined) API works as expected.

markuspf:
Should we get rid of the "--enable-libgap" configure switch? It currently (as far as I can tell) serves no useful purpose other than to confuse people (me); a dynamically linkable object is created either way; we could just link the `libgap-api.o` to the dynamic object (but not the executable) for now.

src/libgap-api.c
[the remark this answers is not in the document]
markuspf: I'd quite like to be able to finalize GAP for my racket bindings, otherwise using libgap in drracket is a bit of a pain, as it crashes on reload. I might be able to address this issue by unloading the .so file, but have to check whether that's possible.

Either way, its for me to implement if noone else wants it.

markuspf:
To add some more detail: the `GAP_Initialize` function could set the constant `IsLIBGAP`, or have a parameter that starts a shell.

(I am not sure about the usefulness of the `IsLIBGAP` constant right now as I still can't see a usecase; GAP code should not behave differently just because its run "as a library", even less so if we start implementing alternative repls)

src/libgap-api.c
```diff
@@ -0,0 +1,134 @@
...
+
+void GAP_Initialize(int          argc,
+                    char **      argv,
+                    char **      env,
+                    CallbackFunc markBagsCallback,
+                    CallbackFunc errorCallback)
+{
+    InitializeGap(&argc, argv, env);
+    SetExtraMarkFuncBags(markBagsCallback);
+    STATE(JumpToCatchCallback) = errorCallback;
```
markuspf: So since we have `SetExtraMarkFuncBags`, I suppose we want `SetErrorCallback`?

fingolfin: I would think that's preferable, yes (overall, I'd prefer to get rid of as many global (non-static) variables as possible.

src/libgap-api.c
```diff
@@ -0,0 +1,136 @@
...
+}
+
+void GAP_AssPList(Obj list, Int pos, Obj elem)
+{
+    AssPlist(list, pos, elem);
+}
+
+void GAP_SetElmPList(Obj list, Int pos, Obj elem)
+{
+    SET_ELM_PLIST(list, pos, elem);
```
fingolfin: I wrote the following before, but due to updates to this PR, it is now hidden; but I still would like a reply by @sebasguts on this:

Honest open question: I wonder whether there is an actual reason (e.g. extreme performance needs?) for why this "API" exposes potentially dangerous low-level functions like `SET_ELM_PLIST`, or even `AssPlist` -- wouldn't it be more useful and safer to expose `ASS_LIST`, `ELM0_LIST`, `LEN_LIST`, etc.?

I just had a look at the sage code, and also at https://hackmd.io/emNi76svSWCh1fBeLKqPdA# (which @nthiery and me created together). Some notes:
* `LEN_PLIST` is used once in SageMath -- incorrectly, though, as it is also called on ranges, blists, strings -> should really use `LEN_LIST` instead
* as far as I can tell, no other GAP function with the string `plist` (upper or lower case) is used by SageMath, although a few more are imported (but then not used)

Perhaps @sebasguts "needs" some of them? Is there perhaps a code repository somewhere to look at?

fingolfin: Also, @markuspf replied:
> Finding what the correct API is is, in my opinon, the name of the game for the immediate future of "LibGAP".

I have a clear opinion on this: such an API should overall aim to provide high-level APIs at first; and only add lower-level APIs if there is a strong justification that make the higher level APIs unsuitable (and even then, I'd first investigate making the higher level APIs better). In this case, I think it's a mistake to export all those plist accessors, when we already have generic list accessors. The only reason to use the plist ones is raw performance, but a sizeable portion of that is sacrificed by these non-inlined wrappers anyway. There may be other reasons to export plist-only APIs, but these then should be conscious decisions (and ideally explained in comments).

fingolfin: That said, if you guys really, really want to expose these plist APIs (even though e.g. `GAP_SetElmPList` is impossible to use correctly via the provided API GAP_SetElmPList, since `CHANGED_BAG` is not exposed), fine, go ahead -- but I think it runs totally against the idea of creating a long-term stable API, as naturally lower level interfaces are much more likely to change than higher level ones.

tst/testlibgap/basic.c
```diff
@@ -0,0 +1,46 @@
...
+        if( GAP_ElmPList(ires, 1) == True )
+            printf("%s\n", CSTR_STRING(GAP_ElmPList(ires, 5)));
+    }
+}
+
+int main(int argc, char **argv)
+{
+    printf("Initializing GAP...\n");
+    GAP_Initialize(argc, argv, environ, 0L, 0L);
+    printf("done\n");
```
fingolfin: Optional: Maybe change to
```C
printf("# Initializing GAP...\n");
GAP_Initialize(argc, argv, environ, 0L, 0L);
printf("# done\n");
```
so that the output file becomes a valid .tst file?

fingolfin:
The spurious ^M comes from the Browse package, see issue #430

fingolfin:
(So you can avoid it if you e.g. start gap with the `-A` option)

src/libgap-api.c
fingolfin: One more note regarding the symbols used by SageMath: a lot of the low level ones come from the function `gap_eval`. My suggestion here would be to try and replace most of it by using `READ_ALL_COMMANDS`, or a similar new function. To get somewhere quickly, I'd not worry about making the ideal generic function we all want in the long run, but just something that does precisely what SageMath needs; we then may have to touch this again later down the road, but I think that'll be much easier once we on the GAP side have clear insights on what you need. Perhaps @nthiery has thoughts on this, too?

markuspf:
Just to note: I did not intend to hide the comments (I even copied them into a textfile on my machine to not forget about the points raised), but I am fiddling on this PR to move it closer to mergeability (or being obsoleted by smaller PRs, whichever comes first)

src/libgap-api.c
markuspf: This is precisely what `GAP_EvalString` does (or, well, is supposed to do!)

markuspf: @fingolfin said

> Also, @markuspf replied:
> > Finding what the correct API is is, in my opinon, the name of the game for the immediate future of "LibGAP".

> I have a clear opinion on this: such an API should overall aim to provide high-level APIs at first;  and only add lower-level APIs if there is a strong justification that make the higher level APIs unsuitable (and even then, I'd first investigate making the higher level APIs better). In this case, I think it's a mistake to export all those plist accessors, when we already have generic list accessors. The only reason to use the plist ones is raw performance, but a sizeable portion of that is sacrificed by these non-inlined wrappers anyway. There may be other reasons to export plist-only APIs, but these then should be conscious decisions (and ideally explained in comments).

I completely agree. The API that is in `libgap-api.c` right now is what I stole from @sebasguts' code, the only "high level" function I added is `GAP_EvalString`.

I will update this PR with a basic API which I think is useful (informed by the hackmd document you hinted at, and an effort of mine of tying GAP into [racket](https://racket-lang.org) and [idris](https://idris-lang.org). Needless to say that feedback from serious users of the API (i.e. SageMath and Oscar) would be incredibly useful.

stevelinton:
Generally this looks good to me. 

On the "plist" question. I agree with the general sentiment of not exposing `ELM_PLIST`, `SET_ELM_PLIST`, etc. Way back when Martin Schoenert suggested that kernel modules that worked with lists they were given should go through the generic list interface in almost all cases. The PLIST Interface was primarily for functions that needed to create a new list.

In this context, the use case would be a libGAP client that needed to put some objects in a list to pass to some GAP function. In that context it is natural and safe to use 
```
l  = NEW_PLIST( type, len);
for (i = 1; i <= len; i++) {
  SET_ELM_PLIST(l, i, thing);
  CHANGED_BAG(l);  # unless you know for sure that thing is kept alive
}
SET_LEN_PLIST(l);
```

I don't know what the performance penalty fro going through `ASS_LIST` instead would be.

I wonder if it might be appropriate to add an API function that simply does this whole job -- takes a C array (or C++ vector) of GAP objects and makes them into a GAP list. Less critically, one could do something similar for records.

fingolfin:
While I agree that a conversion function that takes a C array of GAP refs and turns it into a plist, I would suggest to only add APIs they are explicitly requested.

ChrisJefferson:
I also would want to think about a function that made a plist from an array, because I would worry about people mallocing that array, which would be bad (as the GC wouldn't see it while it was in malloced memory).

ChrisJefferson:
With regards list functions, I would prefer to only share this like AssList, and maybe wrap new plist to not require the last option which requires a tnum.

My preferred design would be to have some way of marking functions which can cause crashes with UNSAFE, or some similar marker, and encourage people to mostly use functions which check their arguments. However some external users might think that wouldn't be fast enough (I think the they would be wrong)

src/libgap-api.c
sebasguts: Sorry for being late to the show. I would be perfectly happy with the lowercase-functions, actually. I am not very picky here, we would need to do the input checks done by the lowercase versions for OSCAR anyway. I doubt there is a speed difference when doing it from julia or from GAP. Code to look at can be found here: https://github.com/oscar-system/LibGAP.jl/blob/master/src/libgap.jl

Furthermore, I also think the way @markuspf did the new version is also good!

src/libgap-api.c
fingolfin: So, I misunderstood about this: this is *only* meant to be called by `GAP_Initialize`, so there actually is no need for a setter for `STATE(JumpToCatchCallback)`, at least for now.

I still think `SetExtraMarkFuncBags` was "correct", because its presence means we can avoid exposing local state of GASMAN.

In other words, it's fine by me if the above line stays as it is. Sorry for the confusion.

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+//
+// Booleans
+//
+bool GAP_IsBoolean(Obj val)
+{
+    return TNUM_OBJ(val) == T_BOOL;
+}
+
+// What about fail :/
+Obj GAP_MakeBoolean(bool val)
```
fingolfin: So are we now using `bool`? A previous PR of mine suggesting that was shot down.

I am actually confused as to why this code compiles -- I thought to access `bool` in C99 code, one has to `#include <stdbool.h>` first? Are we sure this code is not relying on a GCC extension then (which might make `bool` unconditionally available in C code)?

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+    GAP_ASSERT(res != NULL);
+    if (val == True)
+        *res = 1;
+    else if (val == False)
+        *res = 0;
+    else if (val == Fail)
+        *res = 2;
+    else
+        *res = -1;
+}
```
fingolfin: This function has a return type but returns nothing, which will trigger compiler warnings.

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+    return False;
+}
+
+inline Obj GAP_Fail()
+{
+    return Fail;
+}
+
+
+// TODO: depending on 32/64bit? Int?
```
fingolfin: I don't understand this comment.

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+bool GAP_IsImmediateInteger(Obj val)
+{
+    return IS_INTOBJ(val);
+}
+
+Obj GAP_MakeImmediateInteger(int64_t val)
+{
+    INTOBJ_INT(val);
```
fingolfin: This is missing a return; but worth, it can be easily misused, namely if the input does not fit into an immediate object. Also, why hardcode 64 bit API here? Why not use `Int val` as argument?

But overall, I question the usefulness of this function. I think it would be much better to expose these functions suitably:
```
extern Obj ObjInt_Int(Int i);
extern Obj ObjInt_UInt(UInt i);
extern Obj ObjInt_Int8(Int8 i);
extern Obj ObjInt_UInt8(UInt8 i);

extern Int Int_ObjInt(Obj i);    
extern UInt UInt_ObjInt(Obj i);    
extern Int8 Int8_ObjInt(Obj i);    
extern UInt8 UInt8_ObjInt(Obj i);    
```

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+}
+
+int GAP_GetImmediateInteger(Obj val, uint64_t * res)
+{
+    GAP_ASSERT(res);
+    res = INT_INTOBJ(val);
```
fingolfin: Again: return type, but nothing is returned

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// What about Immediate FFE elements?
```
fingolfin: What about them? Is this a TODO for possible further API extensions?

If it was e, I'd add these (and other things) strictly on a as-needed/as-requested base, always with the question "is there perhaps a more generic API we should export instead" in the back of your head.

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+    // Or test and error?
+    return GET_LEN_STRING(string);
+}
+
+size_t GAP_LengthCString(
+    Obj string)    // Document which length we mean (including/excluding \0)!
+{
+    // assert this is a string?
+    // Or test and error?
+    return GET_LEN_STRING(string) + 1;
```
fingolfin: I find this function extremely confusing. From the name, I'd expect it be "strlen", but it isn't. I can't think of a useful application either.

So I'd just remove it, instead of trying to figure out how to do it right

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+Obj GAP_MakeString(const char * str)
+{
+    return MakeString(str);
+}
+
+size_t GAP_LengthString(
+    Obj string)    // Document which length we mean (including/excluding \0)!
+{
+    // assert this is a string?
+    // Or test and error?
```
fingolfin: `GET_LEN_STRING` already does `GAP_ASSERT(IS_STRING_REP(list));`

But `GAP_IsString` is checking for `IS_STRING`, not `IS_STRING_REP`, so callers wouldn't be able to combine `GAP_IsString` and `GAP_LengthString` reliably without a crash.

Several fixes are possible; e.g. `GAP_LengthString` could call `ConvString` or `IsStringConv` suitably.

Overall, one thing is clear: this API needs a clear decision how to handle errors: Via return code (with the actual return value in a return argument)? Via assertions? Via an error callback? Via setjmp/longjmp? C++ exceptions? ... etc

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+    return GET_LEN_STRING(string) + 1;
+}
+
+// Copying is necessary because at least GASMAN might otherwise move the
+// pointer to
+// the C-string away
+int GAP_GetString(Obj val, char * buf, size_t n)
+{
+    size_t len = MIN(max, GET_LEN_STRING(val));
+    memcpy(buf, CSTR_STRING(val), len);
```
fingolfin: So this ignore `n`, and once again has a return type but no return value.

configure.ac
```diff
@@ -230,6 +230,18 @@ AS_CASE([$with_gc],
...
+dnl User setting: dynamic library mode (off by default)
+dnl
+AC_ARG_ENABLE([libgap],
+    [AS_HELP_STRING([--enable-libgap], [enable building of dynamic library])],
+    [AC_DEFINE([BUILD_LIBGAP], [1], [define if building libgap])],
+    [enable_libgap=no]
+)
+AC_MSG_CHECKING([whether to enable dynamic library mode])
+AC_MSG_RESULT([$enable_libgap])
+AC_SUBST([BUILD_LIBGAP], [$enable_libgap])
```
fingolfin: So... what about your idea to just drop `--enable-libgap`, and setting `IsLIBGAP` in `GAP_Initialize` instead? It seemed sensible to me. @sebasguts any concerns?

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+#include "calls.h"
+#include "gapstate.h"
+#include "gvars.h"
+#include "intobj.h"
+#include "opers.h"
+#include "plist.h"
+#include "stringobj.h"
+
+Obj FuncREAD_ALL_COMMANDS(
+    Obj self, Obj instream, Obj echo, Obj capture, Obj resultCallback);
```
fingolfin: This can now be removed

src/libgap-api.c
```diff
@@ -0,0 +1,337 @@
...
+{
+    Obj instream;
+    Obj res;
+    Obj viewObjFunc, streamFunc;
+
+    streamFunc = GAP_ValueGlobalVariable("InputTextString");
+    viewObjFunc = GAP_ValueGlobalVariable("ViewObj");
+
+    instream = DoOperation1Args(streamFunc, MakeString(cmd));
+    res = FuncREAD_ALL_COMMANDS(0L, instream, False, True, viewObjFunc);
```
fingolfin: Use `READ_ALL_COMMANDS` instead

markuspf:
What would be that "first half" to be committed? Something like this?

 * Make a PR that allows disabling the REPL
 * Makefile updates (removing --enable-libgap)
 * `libgap.{h,c}` that only contain `GAP_Initialize` and `GAP_EvalString`? None of the rest is useful at this stage in my opinion.

Also, while I am not overly fussed with names, I have in the back of my head some opposition to calling this baby `libgap`. Can we (this is not "markus and max we", but "whoever wants to speak, we") please agree (on some level) on a name and then move on? I think we have now suitably nailed down what we're trying to achieve.

My (current) preferred name is GAP (or just API), and calling the includes either `api.c` and `api.h`, or `gap-api.c` and `gap-api.h`.

markuspf:
@fingolfin you might notice that I indeed put "WIP" in the title to signify that this is work in progress, i.e. experimentation, but visible to interested parties so they can, if they want to, comment.

I also don't believe that this code, as is, compiles at all.

I am aware of the problems wrt `bool`, missing return values, and inconsistent design; that's because I am experimenting with writing the code at the moment (also to actually have something to discuss in a developer call, or on the mailing list, or a PR, for that matter).

fingolfin:
Sorry, I missed the title change

markuspf:
Updated to only contain the "first half" of the original PR; To make even `testlibgap` run, I added some functions for dealing with lists.
Of course, I can also "hide" this in  `tst/testlibgap/basic.c` file by calling GAP kernel functions directly there to prevent bikeshedding right now.

I'd like to move towards merging this PR now, which means: please comment about things that need to be changed now to get this merged, and I'll open a new PR with a proposed API for more advanced work.

markuspf:
Note that we need #2723 for this now to stop GAP from starting an invisible shell.

nthiery:
>    Well, I have to admit that I am losing patience with this as well. I am
>    not (productively) using this API, and none of the "stakeholders" have
>    chimed in either; yet "we" (mainly you and me) have sunk quite a bit of
>    time dancing around this tree.

Yeah, I know; I am sorry the timing does not work right. It's
frustrating for me as well not to be able to jump in the discussion at
a time where you are available and investing a lot of efforts. I'd
rather be playing with the new GAP API than taking care of reports as
I have been doing in the last two weeks. Gosh, one more week to go;
and then teaching starts ...

Overall my impression is that, on Sage's side, we could adapt quite
easily whatever the API is. From the discussions we had you know the
kind of feature we need. So proceed to wherever feels best from GAP's
perspective.

Makefile.rules
```diff
@@ -389,13 +388,16 @@ gap$(EXEEXT): libgap.la cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-OBJS
 
 else
 
+all: libgap.so
+libgap.so: symlinks libgap.la
+	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) -shared $(OBJS) $(GAP_LIBS) -o $@
```
fingolfin: Disclaimer: the following is not meant to block this PR as such, just tries to clarify some background.

When I just looked at the above lines, I wondered again about the relation between libgap.so and libgap.la. There is now a dependency here -- but there seems no strict technical reason.

Moreover, on Mac OS X, the suffix .so is wrong, it should be .dylib. And of course on Windows one would expect gap.dll or so. But on Windows, we technically *already* build GAP into a DLL, as we need that, in order to be able to support kernel extensions at all.

However, all this has an "easy" solution: libgap.la *already builds a shared library*. That's what libtool .la files are for: they abstract over static/shared libraries in a portable way. Building `libgap.la` creates `.libs/libgap.dylib` for me on Mac OS X. Specifically:
```
$ ls .libs/
libgap.0.dylib  libgap.dylib  libgap.la  libgap.lai
```
So it already takes care of things like file extensions.

We abuse this for the Cygwin builds:
```
bin/$(GAPARCH)/gap.dll: symlinks libgap.la
	cp .libs/cyggap-0.dll $@  # FIXME: HACK to support kernel extensions
```

But that' a bad HACK, which I only added to make things work quickly. In reality, one should not directly access files in `.libs`, but rather use the `libtool` script to use these files, e.g. to link against them, or to install them.

Indeed, installing a `.la` may cause libtool to relink it (because the rpaths and other settings that one needs to use the library may differ between when using it inside a build directory during development; and when using it as a system wide installed library).

This all means that this `libgap.so` target should be temporary at best. What we should do instead, though, depends a bit on how `libgap.so` is supposed to be used: are people right now linking against it by specifying `-L/path/to/GAPROOT -lgap`? If so, and if they are using GNU libtool as well, they are better of specifying `/path/to/GAPROOT/libgap.la`. If so, but they are not using GNU libtool, it becomes a bit tricky.
Of course, if instead one requires people to do a `make install` from within GAP (which is not fully implemented, in particular the libgap part...),  then all that goes away... 

All in all, the best long term solution probably is to finally implement `make install`, or at least enough of it to make libgap usable by clients. I always planned to work on that eventually (but demand wasn't exactly high), and would be happy to collaborate on it, too.

nthiery:
>    What we should do instead, though, depends a bit on how libgap.so is
>    supposed to be used: are people right now linking against it by
>    specifying -L/path/to/GAPROOT -lgap? If so, and if they are using GNU
>    libtool as well, they are better of specifying
>    /path/to/GAPROOT/libgap.la. If so, but they are not using GNU libtool,
>    it becomes a bit tricky.
>    ...
>    All in all, the best long term solution probably is to finally
>    implement make install, or at least enough of it to make libgap usable
>    by clients. I always planned to work on that eventually (but demand
>    wasn't exactly high), and would be happy to collaborate on it, too.

From Sage's (and packagers I assume) perspective: make install would
be great. But in the meantime, since we are using autotools in many
other places anyway, it should be fine to use libtool. I would just
need to double check our build rule since Cython is involved too, and
I don't remember whether Cython does the linking too.

Makefile.rules
markuspf: The way I use `libgap` at the moment is by 

```
(define-ffi-definer define-gap
  (ffi-lib "/home/makx/git/gap/master/.libs/libgap.so"))
```

in Racket (well knowing that this is of course at best a hack, too). If libgap.so were installed, I'd probably be able to just use `ffi-lib "gap". 

I do not require header files, but I have to give `GAP_Initialize` the path to the GAP library. I believe this is mostly how SageMath uses libgap as well.

The reason the `.so` target is there is because I apparently still did not understand libtool. I did remember that the `.la` file is all we need, and that the `.so` file just appeared for me in `.libs`. 

Happy to learn how to do the `make install` so that enough people are happy (but for that we need people to *use* it).

markuspf:
Squashed and rebased on master. If tests pass I'll merge this.

PR #2706, opened 2018-08-15
Introduce SetExtraMarkFuncBags

markuspf:
Requested for LibGAP #2702

### DIFF
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -1003,6 +1003,12 @@ extern  void            InitCollectFuncBags (
             TNumCollectFuncBags after_func );
 #endif
 
+// ExtraMarkFuncBags, if not NULL, is called during garbage collection
+// This is used for integrating GAP (possibly linked as a shared library) with
+// other code bases which use their own form of garbage collection. For
+// example, with Python (for SageMath).
+typedef void (*TNumExtraMarkFuncBags)(void);
+extern void SetExtraMarkFuncBags(TNumExtraMarkFuncBags func);
 
 /****************************************************************************
 **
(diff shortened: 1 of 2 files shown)


src/gasman.h
```diff
@@ -1003,6 +1003,12 @@ extern  void            InitCollectFuncBags (
             TNumCollectFuncBags after_func );
 #endif
 
+// ExtraMarkFuncBags, if not NULL, is called during garbage collection
+// This is used for integrating GAP (possibly linked as a shared library) with
+// other code bases which use their own form of garbage collection. For
+// example, with Python (for SageMath) or Julia.
```
fingolfin: Not really for Julia, as we are using the Julia GC there.

markuspf: It'd be somewhat helpful if I knew which bits of garbage collection I should care about; at the moment I run with "if it works and tests pass in the master branch, I fire and forget", should I start worrying about boehm and julia gcs?

And if so, I need ways to test them.

fingolfin: I don't see how my remarks imply that you should/need to worry about Julia's GC and/or Boehm GC... All I am saying is that for the GAP-Julia interface being created as part of the OSCAR project, `SetExtraMarkFuncBags` won't be used, because there we don't actually use GASMAN; hence "or Julia" should be dropped from that comment.

fingolfin: (To clarify: I don't mind if it stays in, or is removed on a later date; feel free to merge right away; I am merely pointing out that it is not based on an actual application that I am aware of).

PR #1685, opened 2017-09-07
MatrixObj: MultRowVector

ssiccha:
- Adjust some test files.
- Get rid of IsCheckingVector for IsPlistVectorRep and same for matrices
- Add some GAPDoc documentation
- Implementation and Tests for MultRowVector

PR #2709, opened 2018-08-15
MatrixObj: Rename MultRowVector to MultVector, and add doc, code, tests

fingolfin:
With PR #2640 merged, I am submitting the remaining MatrixObj changes as separate PRs.

This contains @ssiccha work on `MultRowVector` resp. `MultVector` from PR #1685. I think these change are fine, but perhaps we want to squash them into a single commit (and/or split them differently across commits than right now). Also, the changes are a bit more extensive, and I felt it was best to give people a chance to review them separately once again (back then, @frankluebeck briefly commented on the PR, but I think I was the only one to extensively review it.

@ssiccha feel free to approve, even though these are your own changes ;-). In fact, that means you are uniquely qualified to review this now ;-).

### DIFF
--- a/doc/ref/obsolete.xml
+++ b/doc/ref/obsolete.xml
@@ -174,6 +174,10 @@ Here are some further name changes.
   <Item><C>MutableNullMat</C></Item>
   <Item><Ref Func="NullMat"/></Item>
 </Row>
+<Row>
+  <Item><C>MultRowVector</C></Item>
+  <Item><Ref Oper="MultVectorLeft"/></Item>
+</Row>
 </Table>
 
 
@@ -185,6 +189,11 @@ Instead of <C>PositionFirstComponent(list,obj)</C>, you may use
 <C>PositionProperty(list,x->x[1]=obj)</C> as a replacement, depending on
 your specific use case.
 
+<Index Key="MultRowVector"><C>MultRowVector</C></Index>
+The five argument version of the operation <C>MultRowVector</C> has been
+deprecated in GAP 4.10 since it was unused and only available for coefficient
+lists.
+Note that <C>MultRowVector</C> was also renamed to <C>MultVectorLeft</C>.
 
 <ManSection>
 <InfoClass Name="InfoObsolete"/>
--- a/src/listoper.c
+++ b/src/listoper.c
@@ -1315,7 +1315,7 @@ Obj FuncINV_MATRIX_IMMUTABLE( Obj self, Obj mat)
 /* We need these to redispatch when the user has supplied a replacement value. */
 
 static Obj AddRowVectorOp;   /* BH changed to static */
-static Obj MultRowVectorLeftOp; /* BH changed to static */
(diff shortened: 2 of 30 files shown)


doc/ref/obsolete.xml
```diff
@@ -185,6 +189,11 @@ Instead of <C>PositionFirstComponent(list,obj)</C>, you may use
 <C>PositionProperty(list,x->x[1]=obj)</C> as a replacement, depending on
 your specific use case.
 
+<Index Key="MultRowVector"><C>MultRowVector</C></Index>
+The five argument version of the operation <C>MultRowVector</C> has been
+deprecated in GAP 4.8 since it was unused and only available for coefficient
```
ThomasBreuer: GAP 4.8 should be replaced by GAP 4.10?

src/listoper.c
```diff
@@ -1571,33 +1571,50 @@ Obj FuncADD_ROW_VECTOR_2_FAST ( Obj self,
 
 /****************************************************************************
 **
-*F  FuncMULT_ROW_VECTOR_2( <self>, <list>, <mult> )
+*F  FuncMULT_VECTOR_LEFT_RIGHT_2( <self>, <list>, <mult> )
```
ThomasBreuer: The last argument is missing.

doc/ref/obsolete.xml
fingolfin: Indeed

fingolfin:
@ssiccha Sergio, could you please indicate whether you think it's reasonable if I squash this into a single commit?

ssiccha:
@fingolfin 62344d4 should definitely be squashed into bde1e35. I would leave 1e466c6 alone since it touches a lot more files than bde1e35 and 62344d4.

fingolfin:
Squashed that one commit, rebased, and addressed @ThomasBreuer other remark.

hulpke:
Will the identifier `MultRowVector` be kept as a synonym for a number of releases?

fingolfin:
We never remove an identifier that has been made obsolete before *at least* all distributed packages stopped using it. Beyond that, we might remove it -- but if you have special need to keep it around, of course that'd be considered, too; and e.g. a comment noting this requirement could be added next to `DeclareObsoleteSynonym( "MultRowVector", "MultVector" );`. And in addition, of course you'd be able to veto the PR removing the synonym (but of course it's possible to miss such a PR; so I'd suggest we add a comment declaring the need, so that before removing it, you'd be consulted explicitly).

fingolfin:
@ThomasBreuer is this PR OK now, from your perspective? If so, please approve (or at least drop the "request for changes"). Thanks!

fingolfin:
At this point, I think it might be best to not rush this into GAP 4.10 at the last minute, but rather delay it to GAP 4.11. Setting milestone accordingly. Feel free to override me if you think it's important to get this in *now*, though.

PR #2710, opened 2018-08-16
kernel: encapsulate syBuf and syBuffers

fingolfin:
* remove leftovers of pipe support (pipehandle, pipe_socket)
* make SyBufInUse internal
* make SyMarkBufUnused internal and rename to SyBufMarkUnused
* rename SyFileno to SyBufFileno
* add InitSysFiles()
* add SyBufIsTTY()
* add SyReadWithBuffer()
* add SyRedirectStderrToStdOut()
* add SyBufSetEOF()  (this one is a bit of a hack; the EOF handling really should be done inside `sysfiles.c`; but I wanted to keep the change as minimal as possible for now, to avoid regressions)
* make syBuf and syBuffers private

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -1929,21 +1929,19 @@ Obj FuncFD_OF_FILE(Obj self,Obj fid)
 #ifdef HPCGAP
 Obj FuncRAW_MODE_FILE(Obj self, Obj fid, Obj onoff)
 {
-  Int fd;
-  int fdi;
-  while (fid == (Obj) 0 || !(IS_INTOBJ(fid)))
-  {
-    fid = ErrorReturnObj(
-           "<fid> must be a small integer (not a %s)",
-           (Int)TNAM_OBJ(fid),0L,
-           "you can replace <fid> via 'return <fid>;'" );
-  }
-  fd = INT_INTOBJ(fid);
-  fdi = SyBufFileno(fd);
-  if (onoff == False || onoff == Fail)
-    return syStopraw(fdi), False;
-  else
-    return syStartraw(fdi) ? True : False;
+    while (!IS_INTOBJ(fid)) {
+        fid = ErrorReturnObj("<fid> must be a small integer (not a %s)",
+                             (Int)TNAM_OBJ(fid), 0L,
+                             "you can replace <fid> via 'return <fid>;'");
+    }
+
+    Int fd = INT_INTOBJ(fid);
+    if (onoff == False || onoff == Fail) {
+        syStopraw(fd);
+        return False;
+    }
+    else
+        return syStartraw(fd) ? True : False;
 }
 #endif
 
--- a/src/streams.c
+++ b/src/streams.c
@@ -1613,12 +1613,6 @@ Obj FuncPOSITION_FILE (
         return Fail;
     }
 
-    // Need to account for characters in buffer
-    if (syBuf[ifid].bufno >= 0) {
-        UInt bufno = syBuf[ifid].bufno;
-        ret -= syBuffers[bufno].buflen - syBuffers[bufno].bufstart;
-    }
-
     return INTOBJ_INT(ret);
 }
 
@@ -1746,31 +1740,13 @@ Obj FuncREAD_ALL_FILE (
     len = 0;
     lstr = 0;
 
-
-    if (syBuf[ifid].bufno >= 0) {
-        UInt bufno = syBuf[ifid].bufno;
-
-        /* first drain the buffer */
-        lstr = syBuffers[bufno].buflen - syBuffers[bufno].bufstart;
-        if (ilim != -1) {
-            if (lstr > ilim)
-                lstr = ilim;
-            ilim -= lstr;
-        }
-        GROW_STRING(str, lstr);
-        memcpy(CHARS_STRING(str),
-               syBuffers[bufno].buf + syBuffers[bufno].bufstart, lstr);
-        len = lstr;
-        SET_LEN_STRING(str, len);
-        syBuffers[bufno].bufstart += lstr;
-    }
 #ifdef SYS_IS_CYGWIN32
  getmore:
 #endif
     while (ilim == -1 || len < ilim ) {
       if ( len > 0 && !HasAvailableBytes(ifid))
           break;
-      if (syBuf[ifid].isTTY) {
+      if (SyBufIsTTY(ifid)) {
           if (ilim == -1) {
               Pr("#W Warning -- reading to  end of input tty will never "
                  "end\n",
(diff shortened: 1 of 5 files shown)


src/streams.c
```diff
@@ -1607,12 +1607,6 @@ Obj FuncPOSITION_FILE (
         return Fail;
     }
 
-    // Need to account for characters in buffer
-    if (syBuf[ifid].bufno >= 0) {
-        UInt bufno = syBuf[ifid].bufno;
-        ret -= syBuffers[bufno].buflen - syBuffers[bufno].bufstart;
-    }
-
```
fingolfin: FYI: The above code was moved into `SyFtell`

src/streams.c
```diff
@@ -1740,31 +1734,13 @@ Obj FuncREAD_ALL_FILE (
...
-                lstr = ilim;
-            ilim -= lstr;
-        }
-        GROW_STRING(str, lstr);
-        memcpy(CHARS_STRING(str),
-               syBuffers[bufno].buf + syBuffers[bufno].bufstart, lstr);
-        len = lstr;
-        SET_LEN_STRING(str, len);
-        syBuffers[bufno].bufstart += lstr;
-    }
```
fingolfin: FYI: the above code was moved into `SyReadWithBuffer`

src/streams.c
```diff
@@ -1848,11 +1824,6 @@ Obj FuncSEEK_POSITION_FILE (
             "you can replace <pos> via 'return <pos>;'" );
     }
     
-    if (syBuf[INT_INTOBJ(fid)].bufno >= 0)
-    {
-            syBuffers[syBuf[INT_INTOBJ(fid)].bufno].buflen = 0;
-            syBuffers[syBuf[INT_INTOBJ(fid)].bufno].bufstart = 0;
-    }
```
fingolfin: FYI: the above code was moved into `SyFseek`

src/streams.c
```diff
@@ -1945,12 +1916,7 @@ Obj FuncFD_OF_FILE(Obj self,Obj fid)
   }
 
   Int fd = INT_INTOBJ(fid);
-
-  // gzipped files don't have a fd
-  if (syBuf[fd].type == gzip_socket)
-      return INTOBJ_INT(-1);
-
-  Int fdi = syBuf[fd].fp;
+  Int fdi = SyBufFileno(fd);
```
fingolfin: FYI: `SyBufFileno` was taught to deal with type `gzip_socket`

src/sysfiles.c
```diff
@@ -711,41 +774,67 @@ const Char * SyWinCmd (
...
 
 // Check if a member of syBuf is in use
-Int SyBufInUse(Int i)
+static Int SyBufInUse(Int i)
 {
     GAP_ASSERT(i >= 0 && i < 256);
     return syBuf[i].type != unused_socket;
 }
 
+void SyRedirectStderrToStdOut(void)
```
fingolfin: Hmm, `Stderr` vs. `StdOut`... should unify the capitalization... Also, strictly speaking errin/errout are mapped to stdin/stdout, so the name is not quite correct; if anybody has a better idea, let me know!

src/streams.c
```diff
@@ -1783,11 +1759,11 @@ Obj FuncREAD_ALL_FILE (
           do {
               csize =
                   (ilim == -1 || (ilim - len) > 20000) ? 20000 : ilim - len;
-              lstr = SyRead(ifid, buf, csize);
+              lstr = SyReadWithBuffer(ifid, buf, csize);
           } while (lstr == -1 && errno == EAGAIN);
       }
       if (lstr <= 0) {
-          syBuf[ifid].ateof = 1;
+          SyBufSetEOF(ifid);
```
fingolfin: `SyBufSetEOF` is a bit of a hack. It would be better if the code in `sysfiles` took care of setting `ateof`. However, that would have involved modifying not just the new `SyReadWithBuffer ` (which would be OK, I guess), but also `SyFgetsSemiBlock`. I'd prefer to worry about that in a future PR, as this starts to touch some really complicated (if not "messy") code that we want to eventually untangle anyway (e.g. the readline integration)

src/sysfiles.c
```diff
@@ -711,41 +774,67 @@ const Char * SyWinCmd (
...
-SYS_SY_BUFFER syBuffers [ 32];
-
 // Mark a member of syBuf as unused
-void SyMarkBufUnused(Int i)
+static void SyBufMarkUnused(Int i)
 {
     GAP_ASSERT(i >= 0 && i < 256);
     memset(&(syBuf[i]), 0, sizeof(syBuf[i]));
     syBuf[i].type = unused_socket;
+    syBuf[i].fp = -1;
```
ChrisJefferson: This is not a requirement, but I wonder if we should just do a memset here(ISH), to blank the buffer back to initial settings (before setting unused_socket, and fp=-1)

src/sysfiles.c
```diff
@@ -3701,6 +3823,65 @@ static StructGVarFunc GVarFuncs [] = {
...
 
+// This function is called by 'InitSystem', before the usual module
+// initialization.
+void InitSysFiles(void)
+{
+    memset(syBuffers, 0, sizeof(syBuf));
+
+    memset(syBuf, 0, sizeof(syBuf));
+    for (Int i = 0; i < ARRAY_SIZE(syBuf); i++)
+        syBuf[i].type = unused_socket;
```
ChrisJefferson: If we need fp=-1 when we set unused, do we need it here too?

src/sysfiles.c
fingolfin: There already is an memset - perhaps I misunderstand?

src/sysfiles.c
fingolfin: I am not sure we "need" `fp=-1`, but some code seemed to check for it w/o checking whether the `suBuf` entry was in use. But I agree that we should at least be consistent. To that end, it might be best to just call `SyBufMarkUnused` in this loop - it'll be slightly less efficient, but that's totally irrelevant here, and is totally worth the automatic consistency in behaviour.

fingolfin: I just had a closer look, and that lead me to discover several places in our code which confuses OS file descriptors with an index into `syBuf` :-/. I'll submit an update to this PR later.

src/sysfiles.c
ChrisJefferson: You don't misunderstand, I just somehow missed it. Woops.

src/sysfiles.c
ChrisJefferson: I think defensive programming is a good idea (setting fp=-1), as we don't want code accidentally accessing 0!

Also, I did try to find places where we mishandle OS file descriptors, I hope I didn't introduce any new ones, sorry if I did.

fingolfin:
I hope this is ready for merging now. There certainly are more improvements possible, but better to not cram too much into one PR. At the very least, now that syBuf and syBuffers is completely local to a single file, I feel much more confident about refactoring it.

(In this way, this PR is yet another small step towards untangling the readline integration).

PR #2712, opened 2018-08-20
kernel: reduce ALWAYS_INLINE to inline in debug mode

fingolfin:
... else debugging functions using ALWAYS_INLINE is virtually impossible, even in debug mode.


ChrisJefferson:
I'm happy for this to be merged once the error is fixed (which looks like some silly interaction of inline functions and warning messages)

PR #2713, opened 2018-08-20
kernel: fix `for x in iter do` optimization

fingolfin:
This optimization is supposed to bypass method selection overhead for standard
iterators. It was never used, because all our standard iterators are component
objects, but only looked for precords (legacy from GAP 3??).

Also do `IS_PREC_OR_COMOBJ(list)` before `CALL_1ARGS(STD_ITER, list)`.

There are already plenty of tests which exercise the newly enabled code, so I did not add new tests. I also did not try to run any (micro-)benchmarks to find out if this optimization is worth it. OTOH, there is a case to be made that it is unsafe: the method returned by method lookup could change from iteration to iteration, so technically, this optimization is incorrect. But then, I can't imagine any serious code (i.e. not written specifically to triggers this) ever would. 

Still, one may argue that this code should simply be removed, instead of being fixed. I am open to discussing this.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -429,7 +429,7 @@ static ALWAYS_INLINE UInt ExecForHelper(Stat stat, UInt nr)
         /* get the iterator                                                */
         list = CALL_1ARGS( ITERATOR, list );
 
-        if ( CALL_1ARGS( STD_ITER, list ) == True && IS_PREC(list) ) {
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
I think this is OK, as if anyone did write such stupid code, we would only run the original functions, not crash/segfault.

PR #2714, opened 2018-08-20
Introduce SetJumpToCatchCallback

markuspf:
as requested for #2702

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -139,4 +139,10 @@ static inline void * StateSlotsAtOffset(ModuleStateOffset offset)
 #define MODULE_STATE(module) \
     (*(module ## ModuleState *)StateSlotsAtOffset(module ## StateOffset))
 
+typedef void (*FuncJumpToCatchCallback)(void);
+static inline void SetJumpToCatchCallback(FuncJumpToCatchCallback func)
+{
+    STATE(JumpToCatchCallback) = func;
+}
+
 #endif    // GAP_GAPSTATE_H


src/gapstate.h
```diff
@@ -139,4 +139,10 @@ static inline void * StateSlotsAtOffset(ModuleStateOffset offset)
 #define MODULE_STATE(module) \
     (*(module ## ModuleState *)StateSlotsAtOffset(module ## StateOffset))
 
+typedef void (*FuncJumpToCatchCallback)(void);
+static inline void SetJumpToCatchCallback(FuncJumpToCatchCallback func)
+{
+    STATE(JumpToCatchCallback) = func;
+}
```
fingolfin: While this is marginally better than asking clients of libgap to set `JumpToCatchCallback` directly, it still makes the precise layout and implementation of GAPState a part of the ABI, which thus is highly unstable. It also forces those clients to include `gapstate.h`, instead of just `libgap.h`. While I guess that's fine for all current applications, a cleaner approach would be to have just a function decl for `SetJumpToCatchCallback` "somewhere" (e.g. in `libgap.h`), and then hide its implementation in a C file.

I guess the only bit useful about this PR then is that client code using libgap can be rewritten to use `SetJumpToCatchCallback` now, and just needs to be recompiled when/if `SetJumpToCatchCallback` gets replaced with a better version (but by then, this client code already had to include `gapstate.h`, which then one would have to remember to undo)...

markuspf: To be honest, I was never imagining "client" code to use `SetJumpToCallback`, I saw it as a bit of GAP internal encapsulation.

At the moment `JumpToCatchCallback` is set precisely once, namely when calling `GAP_Initialize`, and hence the "client" will not see this function.

I have not seen a usecase for setting the callback in the middle of GAP runtime (if it comes along, we can add `GAP_SetErrorCallback` to the API). The things that I think the client should see, I am adding into `libgap-api.h` (which does not mean that the current file is *complete*, it just means if I put something there I think it should be visible).

In that sense this whole PR is unnecessary for "libgap" (as, in the same sense, is the callback setter for marking bags in GASMAN).

fingolfin: OK, so if this is only set by `GAP_Initialize`, then I think we could also just live w/o this PR at all, and directly set `JumpToCatchCallback` there. Either is fine by me.

fingolfin:
I think this is no longer needed, per the discussion above resp. on #2702, so we can close it?

markuspf:
agreed.

PR #2718, opened 2018-08-20
Improve valgrind support

ChrisJefferson:
This patch greatly improves --enable-valgrind. This puts padding before and after every Bag, so we can detect invalid reads or writes off the beginning and end of every Bag

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -468,24 +468,55 @@ static inline UInt IS_BAG_BODY(void * ptr)
 
 #include <valgrind/valgrind.h>
 #include <valgrind/memcheck.h>
-static Int canary_size(void)
+
+// tell valgrind that the masterpointer, bag contents and bag header of Bag
+// should all be accessible
+static void CANARY_ALLOW_ACCESS_BAG(Bag bag)
 {
-    Int bufsize = (Int)EndBags - (Int)AllocBags;
-    return bufsize < 4096 ? bufsize : 4096;
+    VALGRIND_MAKE_MEM_DEFINED(bag, sizeof(Bag));
+    char * ptr = (char *)PTR_BAG(bag);
+    Int    bagLength = SIZE_BAG(bag);
+    VALGRIND_MAKE_MEM_DEFINED(ptr, bagLength);
+
+    BagHeader * header = BAG_HEADER(bag);
+    VALGRIND_MAKE_MEM_DEFINED(
+        header, sizeof(*header) - sizeof(header->memory_canary_padding));
 }
 
-static void ADD_CANARY(void)
+// Reverse CANARY_ALL_ACCESS_BAG, making the masterpointer, bag contents and
+// bag header all inaccessible
+static void CANARY_FORBID_ACCESS_BAG(Bag bag)
 {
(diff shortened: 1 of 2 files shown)


src/gasman.c
```diff
@@ -405,6 +417,18 @@ Bag                     ChangedBags;
...
+/*  These are used to overwrite masterpointers which may still be
+linked from weak pointer objects but whose bag bodies have been
+collected.  Two values are used so that old masterpointers of this
+kind can be reclaimed after a full garbage collection. The values must
+not look like valid pointers, and should be congruent to 1 mod sizeof(Bag),
+to ensure that IS_WEAK_DEAD_BAG works correctly.
+*/
+
+Bag * NewWeakDeadBagMarker = (Bag *)(1000*sizeof(Bag) + 1L);
+Bag * OldWeakDeadBagMarker = (Bag *)(1001*sizeof(Bag) + 1L);
```
fingolfin: Why is this moved around?

src/gasman.c
```diff
@@ -456,24 +480,52 @@ static inline UInt IS_BAG_BODY(void * ptr)
...
 #include <valgrind/memcheck.h>
-static Int canary_size(void)
+
+// tell valgrind that the masterpointer, bag contents and bag header of Bag should all be accessible
+static void CANARY_ALLOW_ACCESS_BAG(Bag bag)
 {
-    Int bufsize = (Int)EndBags - (Int)AllocBags;
-    return bufsize < 4096 ? bufsize : 4096;
+    VALGRIND_MAKE_MEM_DEFINED(bag, sizeof(Bag));
+    char* ptr = (char*)PTR_BAG(bag);
```
fingolfin: clang-format the new code?

src/gasman.c
```diff
@@ -2043,7 +2094,7 @@ UInt CollectBags (
             /* free the identifier                                         */
             *(Bag*)(header->link) = FreeMptrBags;
             FreeMptrBags = header->link;
-
+
```
fingolfin: ?

src/gasman.c
ChrisJefferson: I need to refer to it in an earlier place in the file

ChrisJefferson: But turns out I didn't end up using that. Taken this part of the PR out.

src/gasman.c
ChrisJefferson: All formatted

src/gasman.c
ChrisJefferson: fixed

src/gasman.c
```diff
@@ -202,11 +202,23 @@ enum {
 };
 
 
-static inline UInt WORDS_BAG(UInt size)
+// BAG_SLACK is used to define a block of empty space at the end of each
+// bag, which can then be marked as "not accessable" in the memory checker
```
fingolfin: Typo: accessable -> accessible?

src/gasman.c
fingolfin: Still is in the diff? But anyway, not important

fingolfin:
There is now a merge commit by @markuspf in here. Perhaps @ChrisJefferson can rebase the PR once again to get rid of that?

Issue #1089, opened 2017-01-18
Tighten up rules on loading workspaces


ChrisJefferson:
At the moment GAP checks the kernel version when loading workspaces (so for example 4.4.11 won't load a 4.4.10 workspace). I would like to suggest the following tightening of the rules:

* Reject if `HAVE_LIBREADLINE` is different
* Reject if `BUILD_VERSION` is different (that is, the git version the kernel was build from)

Only the last one I imagine might be controversial, but I think we should do it, as a few times I have found it can produce hard-to-debug errors. We could instead in this last case just produce a warning if people really want to be able to try it.

EDIT: I wondered about rejecting if `GMP_VERSION` is different, but I don't think that should effect workspaces, and would break workspaces when the OS updates the GMP version.

The coding for this is easy, but I wanted to get thoughts before starting.

Would fix #1079

stevelinton:
The GMP version would only matter if gmp changed it's mpn binary data format, which seems pretty unlikely.

fingolfin:
This was fixed by PR #2720 
@ChrisJefferson the "fixes" keyword must appear for before each issue number in order for the issue to be auto-closed when the PR gets merged.

PR #2720, opened 2018-08-21
Encode if we have readline in workspaces

### LINKED ISSUE #1079: Segfault loading workspace wo readline support
I compiled GAP 4.8.6 wo readline support, saved a workspace, then recompiled GAP 4.8.6 with readline support. Starting GAP then yields a segfault. For a full run see below, the error is basically
```GAP
Error, Record: '<rec>.CommandLineEditFunctions' must have an assigned value in
  SHELL( GetBottomLVars(  ), false, false, 3, true, prompt, function (  )
...
```

This could in principle lead to problems if users would want to share or send each other GAP-workspaces. ( Does anybody ever do this? )

### Observed behaviour
Note that the banner says  `Libs used:  gmp` although this GAP version actually uses readline. (Hence the error I guess)
```GAP
 ┌───────┐   GAP 4.8.6, 12-Nov-2016, build of 2017-01-12 17:56:36 (CET)
 │  GAP  │   http://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-gcc-default64
 Libs used:  gmp
 Loaded workspace: /home/sergio/.gap/emptyWorkspace
 Components: trans 1.0, prim 2.1, small* 1.0, id* 1.0
 Packages:   AClib 1.2, Alnuth 3.0.0, AtlasRep 1.5.1, AutPGrp 1.6, CRISP 1.4.4, Cryst 4.1.12, CrystCat 1.1.6, CTblLib 1.2.2, FactInt 1.5.3, FGA 1.3.1, 
             GAPDoc 1.5.1, IRREDSOL 1.3.1, LAGUNA 3.7.0, Polenta 1.3.7, Polycyclic 2.11, RadiRoot 2.7, ResClasses 4.5.0, Sophus 1.23, SpinSym 1.5, 
             TomLib 1.2.6, Utils 0.43
gap> Error, Record: '<rec>.CommandLineEditFunctions' must have an assigned value in
  SHELL( GetBottomLVars(  ), false, false, 3, true, prompt, function (  )                                                                                     
      if IsBound( OnGAPPromptHook) and IsFunction( OnGAPPromptHook )  then                                                                                    
          OnGAPPromptHook(  );                                                                                                                                
      else                                                                                                                                                    
          return;                                                                                                                                             
      fi;                                                                                                                                                     
      return;                                                                                                                                                 
  end, "*stdin*", "*stdout*", true ); at /opt/gap/gap4r8/lib/session.g:49 called from                                                                         
SESSION(  ); at /opt/gap/gap4r8/lib/session.g:70 called from                                                                                                  
<function "POST_RESTORE">( <arguments> )                                                                                                                      
 called from read-eval loop at line 1 of *stdin*                                                                                                              
brk> Error, Record: '<rec>.CommandLineEditFunctions' must have an assigned value in                                                                           
  res := SHELL( context, mayReturnVoid, mayReturnObj, 1, false, prompt, false, "*errin*", "*errout*", false ); at /opt/gap/gap4r8/lib/error.g:235 called from 
SHELL( GetBottomLVars(  ), false, false, 3, true, prompt, function (  )                                                                                       
      if IsBound( OnGAPPromptHook) and IsFunction( OnGAPPromptHook )  then                                                                                    
          OnGAPPromptHook(  );                                                                                                                                
      else                                                                                                                                                    
          return;                                                                                                                                             
      fi;                                                                                                                                                     
      return;                                                                                                                                                 
  end, "*stdin*", "*stdout*", true ); at /opt/gap/gap4r8/lib/session.g:49 called from                                                                         
SESSION(  ); at /opt/gap/gap4r8/lib/session.g:70 called from                                                                                                  
<function "POST_RESTORE">( <arguments> )                                                                                                                      
 called from read-eval loop at line 1 of *errin*                                                                                                              
```

ChrisJefferson:
Encode if readline is enabled in workspaces. Fixes #1079 and #1089

Also, will (hopefully) make it easier in future to add info to workspaces in a backwards compatible way (at least, backwards compatible that the errors will be reasonable when the workspace doesn't load).

### DIFF
--- a/src/saveload.c
+++ b/src/saveload.c
@@ -528,13 +528,24 @@ static void RemoveSaveIndex( Bag bag)
   LINK_BAG(bag) = bag;
 }
 
+
+static Char * GetKernelDescription(void)
+{
+    static Char SyKernelDescription[256];
+    strcpy(SyKernelDescription, SyKernelVersion);
+    if (SyUseReadline) {
+        strcat(SyKernelDescription, " with readline");
+    }
+    return SyKernelDescription;
+}
+
 static void WriteSaveHeader( void )
 {
   UInt i;
   UInt globalcount = 0;
   
   SaveCStr("GAP workspace");
-  SaveCStr(SyKernelVersion);
+  SaveCStr(GetKernelDescription());
 
 #ifdef SYS_IS_64_BIT             
   SaveCStr("64 bit");
@@ -661,10 +672,11 @@ void LoadWorkspace( Char * fname )
   if (strcmp (buf, "GAP workspace") == 0) {
 
      LoadCStr(buf,256);
-     if (strcmp (buf, SyKernelVersion) != 0) {
-        Pr("This workspace is not compatible with GAP kernel (%s, present: %s).\n", 
-           (long)buf, (long)SyKernelVersion);
-        SyExit(1);
+     if (strcmp(buf, GetKernelDescription()) != 0) {
+         Pr("This workspace is not compatible with GAP kernel (%s, present: "
+            "%s).\n",
+            (long)buf, (long)GetKernelDescription());
+         SyExit(1);
      }
 
      LoadCStr(buf,256);


src/saveload.c
```diff
@@ -528,13 +528,26 @@ static void RemoveSaveIndex( Bag bag)
   LINK_BAG(bag) = bag;
 }
 
+
+
+static Char * GetKernelDescription( void )
+{
+    static Char SyKernelDescription[256];
+    strcpy(SyKernelDescription, SyKernelVersion);
+    if(SyUseReadline)
```
fingolfin: clang-format? add issue labels?

Perhaps we can setup a bot to point this out automatically, it's a bit tiresome. Alternatively, we can also agree to not bother about clang-format anymore, and just let everybody format whichever way they feel like?

ChrisJefferson: Id be happy with a bot. Maybe i can write a git script to run on my own machine that clangformats bevore each commit. (I actually do use formst on save on other projects i work on, but that reformats the whole file, so is no good for gap).

ChrisJefferson: Now reformatted

PR #2723, opened 2018-08-21
Introduce command line options `--norepl` and `--nointeract`, remove `IsLIBGAP` constant

markuspf:
This is useful for "libgap" as introduced in #2702 and possibly using GAP in "batch mode" as one can leave out the `QUIT_GAP(0);` in GAP scripts.

This makes the `IsLIBGAP` constant obsolete, so we remove it as well.


markuspf:
Could be two PRs but the two commits are trivial enough I think.

ChrisJefferson:
Seems sensible, it's a thing I've wanted before actually, when running GAP from a script.

Super picky, should it be --norepl (or something like that) , just because you can still drop into the break loop and do things in there, which feels a bit like the shell, or should it automatically also enable --nobreakloop (do you want to always pair noshell with nobreakloop for your purposes?)

ChrisJefferson:
Note, I'm happy for the option, and it's name, to be exactly what it is now, just wanted to see if it had been thought about :)

markuspf:
I considered calling it `--norepl`, but the REPL is called `Shell` in GAP as far as I know;

You're right about the breakloop (when I tried it out here just now I ended up in the breakloop); For "libgap" purposes one would never want a break loop as far as I am concerned.

Not sure whether one wants `--nointeractive` or `--batch` that just disables both?

markuspf:
@ChrisJefferson I introduced `--norepl` and `--nointeract`; `--norepl` just disables the repl, but still gives you a break loop, whereas `--nointeract` disables all interaction (so it sets `--norepl` and --nobreakloop`).

markuspf:
If anyone (maybe in particular @stevelinton @fingolfin @ChrisJefferson @sebasguts) has any opinions on the naming bikeshed (on these command line options), I am all ears.

ChrisJefferson:
Those are fine names for me

lib/system.g
```diff
@@ -111,7 +111,11 @@ BIND_GLOBAL( "GAPInfo", rec(
            help := [ "Run ProfileLineByLine(<filename>) with recordMem := true on GAP start"] ),
       rec( long := "cover", default := "", arg := "<file>",
            help := [ "Run CoverageLineByLine(<filename>) on GAP start"] ),
-      rec( long := "enableMemCheck", default := false)
+      rec( long := "enableMemCheck", default := false),
+      rec( long := "norepl", default := false,
+           help := [ "Disable the GAP REPL" ] ),
```
fingolfin: There are people who don't know what a REPL is. There is also no help under `?REPL`. Of course one can google, but I'd prefer if we did not force people to do that. Can we perhaps expand this, e.g. by writing "Disable the GAP read-eval-print loop (REPL)" or so?

markuspf:
@fingolfin updated the help text, and rebased on current master.

I mentioned in the discussion before that "REPL" might not be as well-known a phrase as one might think, but also agreed with @ChrisJefferson that `--noshell` might be misunderstood to mean no interaction at all.

PR #2726, opened 2018-08-22
Validate PackageInfo.g when loading packages

fingolfin:
This helps package authors notice earlier if a package update they made failed.

One could also be more aggressive and validate every package record as we load it; but I have far too many packages which are not validating sitting around (e.g. undeposited ones from gap-packages)


alex-konovalov:
Nice - from the packages redistributed with GAP, there is only one that triggers a warning:
```
Loading crime ... 
#E  component `Date' must be bound to a string of the form `dd/mm/yyyy'
#E Validation of package crime from /home/travis/build/gap-system/gap/pkg/crime failed 
```
because it's date is `01/6/2011` and there were no new releases ever since.

fingolfin:
Rebased.

PR #2729, opened 2018-08-22
Rename QuaternionGroup to DicyclicGroup, Document IsDihedralGroup and IsQuaternionGroup

### LINKED ISSUE #2725: Quaternion groups and Dicyclic groups, bug? confusion?
On the current GAP master branch (and probably past GAPs) the following happens, which is at least surprising:

```
gap> Q := QuaternionGroup(12);;
gap> IsQuaternionGroup(Q);
false
```

Now this could be because of different definitions of `QuaternionGroup` being used in the constructor vs the method for `IsQuaternionGroup`; it looks like the documentation uses `DicyclicGroup` and `QuaternionGroup` interchangeably (and `grp/basic.gd` declares `QuaternionGroup` and `DicyclicGroup` as synonyms).

 "My definion" of `QuaternionGroup` would be "Dicyclic group whose order is a power of 2" (though I don't have a reference to back that up right now, other than wikipedia).

This view is supported by the following (except that there seems to be a minimum size of 8 as well):

```
gap> Qs := List(4*[1..32], QuaternionGroup);;
gap> List(Qs, x->[Size(x), IsQuaternionGroup(x)]);
[ [ 4, false ], [ 8, true ], [ 12, false ], [ 16, true ], [ 20, false ], [ 24, false ], 
  [ 28, false ], [ 32, true ], [ 36, false ], [ 40, false ], [ 44, false ], 
  [ 48, false ], [ 52, false ], [ 56, false ], [ 60, false ], [ 64, true ], 
  [ 68, false ], [ 72, false ], [ 76, false ], [ 80, false ], [ 84, false ], 
  [ 88, false ], [ 92, false ], [ 96, false ], [ 100, false ], [ 104, false ], 
  [ 108, false ], [ 112, false ], [ 116, false ], [ 120, false ], [ 124, false ], 
  [ 128, true ] ]
```

Maybe @hulpke has an opinion on this?

markuspf:
This PR renames `QuaternionGroup` to `DicyclicGroup` and `QuaternionGroupCons` to `DicyclicGroupCons`. 

This is because these functions/constructors allow creation of groups that are not strictly speaking (generalised) quaternion groups: Their only requirement is that the given size of the group is a multiple of four, but the common definition of a generalised quaternion group is that its size is a power of 2 greater than 8. 

The polycyclic package installs a method for `QuaternionGroupCons` which should probably be changed to `DicyclicGroupCons` as well, as it happily produces Dicyclic groups. I'll submit a PR for polycyclic, but have left the declaration of `QuaternionGroupCons` in here for the time being to not break code.

Additionally we enable existing documentation for `IsDihedralGroup` and `IsQuaternionGroup`, methods to recognise whether a given group is a dihedral group or a (generalised) quaternion group: Issue #2725 arose because one could create a group using `QuaternionGroup` for which `IsQuaternionGroup` was `false`. 

Closes #2725

### DIFF
--- a/grp/basic.gd
+++ b/grp/basic.gd
@@ -384,23 +384,81 @@ DeclareSynonym("QuaternionGroupCons", DicyclicGroupCons);
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-BindGlobal( "DicyclicGroup", function ( arg )
+BindGlobal( "GRPLIB_DicyclicParameterCheck",
+function(args, quaternion)
+    local res, size;
 
-  if Length(arg) = 1  then
-    return DicyclicGroupCons( IsPcGroup, arg[1] );
-  elif IsOperation(arg[1]) then
-    if Length(arg) = 2  then
-      return DicyclicGroupCons( arg[1], arg[2] );
-    elif Length(arg) = 3  then
-      # some filters require extra arguments, e.g. IsMatrixGroup + field
-      return DicyclicGroupCons( arg[1], arg[2], arg[3] );
+    res := [];
+
+    if Length(args) = 1 then
+        res[1] := IsPcGroup;
+        res[2] := args[1];
+        size := res[2];
+    elif Length(args) = 2 then
+        res[1] := args[1];
+        res[2] := args[2];
+        size := res[2];
+    elif Length(args) = 3 then
+        res[1] := args[1];
+        res[2] := args[2];
+        res[3] := args[3];
+        size := res[3];
+
+        if not IsField(res[2]) then
+            ErrorNoReturn("usage: <field> must be a field");
+        fi;
+    else
+        ErrorNoReturn("usage: DicyclicGroup( [<filter>, [<field>, ] ] <size> )");
     fi;
-  fi;
-  Error( "usage: DicyclicGroup( [<filter>, [<field>, ] ]<size> )" );
 
-end );
+    if not IsFilter(res[1]) then
+        ErrorNoReturn("usage: <filter> must be a filter");
+    fi;
 
-DeclareSynonym( "QuaternionGroup", DicyclicGroup );
+    if not (IsPosInt(size) and (size mod 4 = 0)) then
+        ErrorNoReturn("usage: <size> must be a positive integer divisible by 4");
+    fi;
+
+    if not IsPrimePowerInt(size) or size < 8 then
+        if quaternion = "error" then
+            ErrorNoReturn("usage: <size> must be a power of 2 and at least 8");
+        elif quaternion = "warn" then
+            Info(InfoWarning, 1, "Warning: QuaternionGroup called with <size> ", size,
+                                 " which is less than 8, or not a power of 2.");
+        fi;
+    fi;
+
+    return res;
+end);
+
+BindGlobal( "DicyclicGroup",
+function(args...)
+    local res;
+
+    res := GRPLIB_DicyclicParameterCheck(args, "");
+
+    return CallFuncList(DicyclicGroupCons, res);
+end);
+
+BindGlobal( "QuaternionGroup",
+function(args...)
+    local res;
+
+    res := GRPLIB_DicyclicParameterCheck(args, "warn");
+
+    return CallFuncList(DicyclicGroupCons, res);
+end);
+
+BindGlobal( "GeneralisedQuaternionGroup",
+function(args...)
+    local res, grp;
+
+    res := GRPLIB_DicyclicParameterCheck(args, "error");
+    grp := CallFuncList(DicyclicGroupCons, res);
+    SetIsQuaternionGroup(grp, true);
+
+    return grp;
+end);
 
 
 #############################################################################
--- a/grp/basic.gd
+++ b/grp/basic.gd
@@ -343,31 +343,31 @@ end );
 
 #############################################################################
 ##
-#O  QuaternionGroupCons( <filter>, <n> )
+#O  DicyclicGroupCons( <filter>, <n> )
 ##
 ##  <ManSection>
-##  <Oper Name="QuaternionGroupCons" Arg='filter, n'/>
+##  <Oper Name="DicyclicGroupCons" Arg='filter, n'/>
 ##
 ##  <Description>
 ##  </Description>
 ##  </ManSection>
 ##
-DeclareConstructor( "QuaternionGroupCons", [ IsGroup, IsInt ] );
+DeclareConstructor( "DicyclicGroupCons", [ IsGroup, IsInt ] );
+DeclareSynonym("QuaternionGroupCons", DicyclicGroupCons);
 
 
 #############################################################################
 ##
-#F  QuaternionGroup( [<filt>, ]<n> )  . . . . . . . quaternion group of order <n>
+#F  DicyclicGroup( [<filt>, [<field>, ] ]<n> )  Dicyclic group of order <n>
 ##
-##  <#GAPDoc Label="QuaternionGroup">
+##  <#GAPDoc Label="DicyclicGroup">
 ##  <ManSection>
-##  <Func Name="QuaternionGroup" Arg='[filt, ]n'/>
-##  <Func Name="DicyclicGroup" Arg='[filt, ]n'/>
+##  <Func Name="DicyclicGroup" Arg='[filt, [field, ]]n'/>
+##  <Func Name="QuaternionGroup" Arg='[filt, [field, ]]n'/>
 ##
 ##  <Description>
-##  constructs the generalized quaternion group (or dicyclic group) of size
-##  <A>n</A> in the category given by the filter <A>filt</A>.  Here, <A>n</A>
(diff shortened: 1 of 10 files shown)


markuspf:
This is still far from perfect; we don't have `IsDycyclicGroup` yet, and the recognition is in `lib/grpnames.g{d,i}` and construction is in `grp/`; I don't know whether we want to move anything anywhere anytime soon though (and I don't want this PR to baloon into a huge project).

grp/basic.gd
```diff
@@ -343,64 +343,122 @@ end );
...
-##  <Func Name="DicyclicGroup" Arg='[filt, ]n'/>
+##  <Func Name="DicyclicGroup" Arg='[filt, [field, ]]n'/>
+##  <Func Name="QuaternionGroup" Arg='[filt, [field, ]]n'/>
 ##
 ##  <Description>
-##  constructs the generalized quaternion group (or dicyclic group) of size
-##  <A>n</A> in the category given by the filter <A>filt</A>.  Here, <A>n</A>
-##  is a multiple of 4.
+##  constructs the dicyclic group of size <A>n</A> in the category given by the
+##  filter <A>filt</A>. Here, <A>n</A> is a multiple of 4.
```
fingolfin: This text seems to refer to implicitly refer to `DicyclicGroup` only, but is of course listed directly after `QuaternionGroup`. I'd rephrase to e.g. "DicyclicGroup constucts ... " and then perhaps insert one sentence that mentions `QuaternionGroup` does the same but is restricted to inputs which are powers of 2, and that is provided for backwards compatibility.

lib/grpnames.gd
```diff
@@ -455,6 +457,7 @@ InstallTrueMethod( IsGroup, IsDihedralGroup );
 #P  IsQuaternionGroup( <G> )
 #A  QuaternionGenerators( <G> )
 ##
+##  <#GAPDoc Label="IsQuaternionGroup">
 ##  <ManSection>
 ##  <Prop Name="IsQuaternionGroup" Arg="G"/>
 ##  <Attr Name="QuaternionGenerators" Arg="G"/>
```
fingolfin: This change and the one a bit above changes four functions from undocumented to documented  -- that's important for release notes, but not mentioned in the PR title or description (and hence easy to miss when creating release notes). Please mention this change in the PR title or description.

But in addition, I am a bit concerned about this: documenting `QuaternionGenerators` will cause people want to use it. But AFAICT we don't provide any methods for computing this attribute -- it is only set as a side effect of `IsQuaternionGroup`. Same for `DihedralGenerators`.

That said, I am not actively opposed to this change, but I am wary about the seemingly carefree approach to making it (I am sure you actually *did* spend time thinking carefully about it, it just isn't that clear from what is visibly documented in the system).

lib/grpnames.gd
```diff
@@ -459,21 +460,30 @@ InstallTrueMethod( IsGroup, IsDihedralGroup );
...
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareProperty( "IsQuaternionGroup", IsGroup );
-DeclareAttribute( "QuaternionGenerators", IsGroup );
+DeclareProperty( "IsGeneralisedQuaternionGroup", IsGroup );
+DeclareAttribute( "GeneralisedQuaternionGenerators", IsGroup );
+# Backwards compatibility
+DeclareSynonym( "IsQuaternionGroup", IsGeneralisedQuaternionGroup );
+DeclareSynonym( "QuaternionGenerators", GeneralisedQuaternionGenerators );
```
fingolfin: Both probably should use `DeclareSynonymAttr`, so that `HasFOO` works.

lib/grpnames.gi
```diff
@@ -949,59 +949,82 @@ InstallMethod( DecompositionTypesOfGroup,
...
 
     # choose t in G\Zn and check dihedral structure
     repeat t := Random(G); until not t in Zn;
     if not (Order(t) = 2 and ForAll(GeneratorsOfGroup(Zn),s->t*s*t*s=s^0))
-    then return false; fi;
+    then return fail; fi;
 
     # choose generator s of Zn
     repeat s := Random(Zn); until Order(s) = n;
     SetDihedralGenerators(G,[t,s]);
```
fingolfin: Remove `SetDihedralGenerators` call here?

grp/basic.gd
```diff
@@ -343,64 +343,125 @@ end );
...
+    if not (IsPosInt(size) and (size mod 4 = 0)) then
+        ErrorNoReturn("usage: <size> must be a positive integer divisible by 4");
+    fi;
+
+    if not IsPrimePowerInt(size) or size < 8 then
+        if quaternion = "error" then
+            ErrorNoReturn("usage: <size> must be a power of 2 and at least 8");
+        elif quaternion = "warn" then
+            Info(InfoWarning, 1, "Warning: QuaternionGroup called with <size> ", size,
+                                 " which is less than 8, or not a power of 2.");
```
fingolfin: Optional: we could be less mysterious and just tell the user explicitly what the problem is (also, I'd remove the trailing period, which is unusual in info messages):
```gap
elif quaternion = "warn" then
  if not IsPrimePowerInt(size) then
     Info(InfoWarning, 1, "Warning: QuaternionGroup called with <size> ", size, " which is not a power of 2");
  else
     Info(InfoWarning, 1, "Warning: QuaternionGroup called with <size> ", size, " which is less than 8");
  fi;
```

lib/grpnames.gd
```diff
@@ -455,21 +458,32 @@ InstallTrueMethod( IsGroup, IsDihedralGroup );
...
-##    the attribute <C>QuaternionGenerators</C> to [<A>t</A>,<A>s</A>],
-##    where <A>t</A> and <A>s</A> are two elements such that <A>G</A> =
-##    <M><A>t, s | s^(2^k) = 1, t^2 = s^(2^k-1), s^t = s^-1</A></M>.
+##    <Ref Prop="IsGeneralisedQuaternionGroup"/> indicates whether the group
+##    <A>G</A> is a generalized quaternion group of size <M>N = 2^(k+1)</M>,
+##    <M>k >= 2</M>.
+##    If it is, methods may set the attribute <Ref Attr="GeneralisedQuaternionGenerators" />
+##    to [<A>t</A>,<A>s</A>], where <A>t</A> and <A>s</A> are two elements such that <A>G</A> =
+##    <M>\langle t, s | s^{(2^k)} = 1, t^2 = s^{(2^k-1)}, s^t = s^{-1} \rangle</M>.
+##    <Ref Prop="IsQuaternionGroup"/> and <Ref Attr="QuaternionGenerators" /> are
```
fingolfin: Optional: perhaps start this sentence with the "The synonyms ..." ?

lib/grpnames.gi
```diff
@@ -949,59 +949,81 @@ InstallMethod( DecompositionTypesOfGroup,
...
+
+InstallMethod( DihedralGenerators,
+               "for a group",
+               [ IsGroup and IsFinite ],
+function(G)
+    local gens;
+
+    gens := DoComputeDihedralGenerators(G);
+    if gens = fail then
+        ErrorNoReturn("G is not a dihedral group");
```
fingolfin: Optional: insert before the error: `SetIsDihedralGroup(G, false);`. Or insert before the `if` the following: ` SetIsDihedralGroup(G, gens <> fail);` (and then remove the other call to `SetIsDihedralGroup` below). The same can of course be done in the quaternion case.

lib/grpnames.gi
```diff
@@ -949,59 +949,81 @@ InstallMethod( DecompositionTypesOfGroup,
...
+end;
+
+InstallMethod( IsDihedralGroup,
+               "for a group",
+               true,
+               [ IsGroup and IsFinite ], 0,
+function(G)
+    local gens;
+
+    if not HasDihedralGenerators(G) then
```
fingolfin: Options:  `InstallTrueMethod(IsDihedralGroup, HasDihedralGenerators);` in the `.gd` file, and then remove this `if` (the method can only be called if the dihedral generators are not (yet) known; though even if my claim was wrong, clearly it wouldn't be wrong to recompute them, only inefficient)

Of course the same could be done in the quaternion case.

grp/basic.gd
```diff
@@ -343,64 +343,122 @@ end );
...
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-BindGlobal( "QuaternionGroup", function ( arg )
+BindGlobal( "GRPLIB_DicyclicParameterCheck",
+function(args, quaternion)
+    local size;
+
+    if Length(args) = 1 then
+        res := [ IsPcGroup, args[1] ];
```
fingolfin: This now errors out, because the `local res` is missing. Either add it back, or simply replace `res` by `args` (args is not used anymore later).

The other test failure seems to be due to the changed warning. Perhaps this tidbit is helpful (it's been for me, at least):
```gap
TestDirectory("tst/testinstall", rec(rewriteToFile:=true));
```

PR #2730, opened 2018-08-22
kernel: remove lvars tracing code (TRACEFRAMES, STEVES_TRACING) and SET_CURR_LVARS

fingolfin:
The tracing code was IMO not that useful (and I was working on that code recently); and if somebody needs it, could be added back. In the meantime, it complicates th code

That said, others might disagree, and then I don't mind dropping that part of the patch.

More important to me is to replace SET_CURR_LVARS by SWITCH_TO_OLD_LVARS


fingolfin:
@stevelinton I am removing a feature named after you, OK with that? ;-)

stevelinton:
@fingolfin entirely. It was a long time ago, and I think I committed because at the time that was the only way to conveniently access it on multiple machines (or something like that).

PR #2732, opened 2018-08-22
Add kernel helper IS_FILTER, other tweaks

fingolfin:
In particular, before this PR, we get this:
```
gap> Center and IsAssociative;
<Filter "(Centre and IsAssociative)">
gap> FirstOp and IsAssociative;
Error, <flags1> must be a flags list (not a boolean or fail)
```
Note that Center is an attribute, not a filter.

With this PR:
```
gap> Center and IsAssociative;
Error, <expr> must be 'true' or 'false' or a filter (not a function)
gap> FirstOp and IsAssociative;
Error, <expr> must be 'true' or 'false' or a filter (not a function)
```

In addition, argument validation for `and` differed slightly between interpreter and executor, which this PR also fixes.

I am on the fence whether this should be considered relevant for the release notes. On the one hand, probably nobody ever noticed these quirks. On the other, this is a change in what kind of inputs GAP accepts...

### DIFF
--- a/src/opers.h
+++ b/src/opers.h
@@ -47,10 +47,20 @@ typedef struct {
     // cache of an operation
     Obj cache[8];
 
-    // 1 if the operation is an attribute and storing is enabled (default) else 0
-    Obj enabled;
+    // small integer encoding a set of bit flags with information about the
+    // operation, see OperExtras below
+    //
+    // note: this is encoded as an integer object, and not just stored
+    // directly as C bitfield, to avoid the need for a custom marking function
+    // which does not call 'MarkBag' on this field (while that would be safe
+    // to do with GASMAN, it may not be in alternate GC implementations)
+    Obj extra;
 } OperBag;
 
+enum OperExtras {
+    OPER_IS_ATTR_STORING = (1 << 0),
+    OPER_IS_FILTER       = (1 << 1),
+};
 
 /****************************************************************************
 **
@@ -63,9 +73,9 @@ extern Obj TRY_NEXT_METHOD;
 **
 *F  IS_OPERATION( <obj> ) . . . . . . . . . . check if object is an operation
 */
-static inline Int IS_OPERATION(Obj func)
+static inline Int IS_OPERATION(Obj obj)
 {
-    return TNUM_OBJ(func) == T_FUNCTION && SIZE_OBJ(func) == sizeof(OperBag);
+    return TNUM_OBJ(obj) == T_FUNCTION && SIZE_OBJ(obj) == sizeof(OperBag);
 }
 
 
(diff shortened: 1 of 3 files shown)


src/opers.h
```diff
@@ -214,7 +215,33 @@ static inline Int ENABLED_ATTR(Obj oper)
 */
 static inline void SET_ENABLED_ATTR(Obj oper, Int x)
 {
-    OPER(oper)->enabled = INTOBJ_INT(x);
+    Obj val = CONST_OPER(oper)->enabled;
+    Int v = val ? INT_INTOBJ(val) : 0;
+    OPER(oper)->enabled = INTOBJ_INT(v | 1);
```
fingolfin: So, this abuses `enabled` as a bit array. Since it now stores two bits (not just one), perhaps we should rename it. But I couldn't think of a good name -- my only idea, `flags`, already is used by another field in struct `OperBag`...

fingolfin:
Actually I screwed up `SET_ENABLED_ATTR` in here. I'll fix that now and also rename `enabled`.

markuspf:
Not a fan of the name `extras`, but with the lack of a better name, I won't stand in the way of this being merged.

fingolfin:
Updated. Turns out that tests were failing due to an actual bug revealed by this PR: `CanEasilyDetermineCanonicalRepresentativeExternalSet` was declared as an attribute, but should have been a property. It is used for a method installation, but only in an and-filter, like so:
```gap
InstallMethod( \=, "xorbs with canonicalRepresentativeDeterminator",
  IsIdenticalObj,
  [IsExternalOrbit and CanEasilyDetermineCanonicalRepresentativeExternalSet,
   IsExternalOrbit and CanEasilyDetermineCanonicalRepresentativeExternalSet],
  0,
...
```
Which GAP silently accepted; the resulting filter then was `IsExternalOrbit and HasCanEasilyDetermineCanonicalRepresentativeExternalSet`, i.e., *not* at all what was intended...

I already found similar problems in homalg, see <https://github.com/homalg-project/homalg_project/pull/205>; and also in xmodalg, see <https://github.com/gap-packages/xmodalg/pull/25>. I am looking for more right now.

@alex-konovalov perhaps you could run package tests against this PR before we merge it? It might reveal additional similar bugs in further packages.

fingolfin:
Since this breaks various packages, there are two options:

1. We wait with merging this until all affected packages have a patch released.
2. I update this PR to only print a (big and fat) warning) for the abuse, but otherwise accept it as before; then once all packages are fixed, we turn that back into an error.

Option 2 is a bit more work, but it has the advantage that it reduces the likelihood of new instances of this broken behavior being introduced. OTOH, I am not sure how big the likelihood for that is...

ChrisJefferson:
If 2 isn't too much work, it quite appeals, because it also means this doesn't get lost.

alex-konovalov:
This is the full list of packages where problems occur (some of they may be cause by dependencies): - homalgtocas, io_forhomalg, linearalgebraforcap, localizeringforhomalg, matricesforhomalg, modulepresentationsforcap, modules, qpa, ringsforhomalg, sco, simpcomp, toricvarieties, xmodalg. Perhaps simpcomp is not yet looked at.

fingolfin:
I rebased this, and for now dropped the commit which tightened the rules for AND-ing filters. Hence this PR could be merged now.

I'll put the removed commit into a new PR and/or will make a PR which just produces a warning for "bad" and-filters.

fingolfin:
@alex-konovalov as far as I could tell, all those packages fail due to one of homalg, qpa or xmodalg. The latter two already merged the fixes, so it's mainly waiting for @mohamed-barakat or @sebasguts to merge the homalg fix and release that.

Issue #2724, opened 2018-08-21
Bug in mapping from an iterated semidirect product?


cdwensley:
I'm reluctant to claim that there is a bug in what must be long-established code.  Perhaps someone can persuade me that there is an error in my example? 
I have a mapping h between permutation groups which I expect to be a homomorphism, but GAP claims that it is not single valued.  On repeating the whole construction in pc-groups, the corresponding mapping H _is_ a homomorphism. 
The groups involved are the quaternion group q8; 2 normal cyclic subgroups of order 4, c4a & c4b, and its normal cyclic subgroup of order 2, c2c.  These groups all act on each other by conjugation. 
The semidirect products formed are c4a:q8, c2c:c4b (a direct product) and G=(c2c:c4b):(c4a:q8). 
Then h : G -> c4a:q8 is the mapping which fails to be single valued, but the corresponding pc-group mapping H : PCG -> C4A:Q8 is a homomorphism. 
The attached example file q8.txt (GitHub objected to q8.g) is rather lengthy.  The first part constructs the permutation groups; the second part the pc-groups; and the third part contains some isomorphisms between these.  The mapping G -> PCG also fails to be an isomorphism. 
My tentative conclusion is that there is a problem in the construction of G: I've tried to trace this but baulk at delving into the StabChain coding. 

### Copy and paste GAP banner (to tell us about your setup)
 ┌───────┐   GAP 4.9.2 of 04-Jul-2018
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin17.4.0-default64
 Configuration:  gmp 6.0.0
 Loading the library and packages ...
 Packages:   GAPDoc 1.6.1.dev, PrimGrp 3.3.1, SmallGrp 1.3, TransGrp 2.0.2
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
(Also the development version, GAP 4.8.8-5807-g786fe9e.)

[q8.txt](https://github.com/gap-system/gap/files/2307377/q8.txt)

    ##  expt/q8.txt  
    ##  part 1 : construct permutation groups and homomorphisms: 
    p := (1,2,3,4)(5,8,7,6); 
    q := (1,5,3,7)(2,6,4,8);
    q8 := Group( p, q ); 
    SetName( q8, "q8" );
    a := p;  b := q;  c := p^2; 
    c4a := Subgroup( q8, [a] ); 
    SetName( c4a, "c4a" ); 
[...]

PR #2733, opened 2018-08-23
Minor performance improvements, code cleanup and very local fixes

hulpke:
Most of this is changes that are entirely "under the hood" for a user and improve performance in particular (often too large to allow for specific test examples) cases I observed, or that clean up ugly code. It is not worth mentioning it specifically in a release announcement.

The utility function `MaxesAlmostSimple` is made an operation to allow for future packages to overload (and thus improve the maximal subgroups code).

The only user-visible changes are:

- clarifications in how to interpret the output of `SubgroupsSolvableGroup`.
- A (prior forgotten) warning about a calculation using a classification from the primitive groups library.
- a (very local) fix for #2724 .

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -861,6 +861,10 @@ InstallTrueMethod( IsPolycyclicGroup,
 ##
 DeclareAttribute( "AbelianInvariants", IsGroup );
 
+# minimal number of generators for abelianization. Used internally to check
+# whether it is worth to attempt to reduce generator number
+DeclareAttribute( "AbelianRank", IsGroup );
+
 #############################################################################
 ##
 #A  IsInfiniteAbelianizationGroup( <G> )
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -1868,8 +1868,8 @@ DeclareAttribute( "MinimalNormalSubgroups", IsGroup );
 ##  returns a list of all normal subgroups of <A>G</A>.
 ##  <Example><![CDATA[
 ##  gap> g:=SymmetricGroup(4);;NormalSubgroups(g);
-##  [ Sym( [ 1 .. 4 ] ), Alt( [ 1 .. 4 ] ), Group([ (1,4)(2,3), (1,2) 
-##    (3,4) ]), Group(()) ]
+##  [ Sym( [ 1 .. 4 ] ), Alt( [ 1 .. 4 ] ), Group([ (1,4)(2,3), (1,3) 
+##    (2,4) ]), Group(()) ]
 ##  ]]></Example>
 ##  <P/>
 ##  The algorithm for the computation of normal subgroups is described in
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -710,6 +710,9 @@ InstallTrueMethod( IsSimpleGroup, IsSporadicSimpleGroup );
 DeclareProperty( "IsAlmostSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup, IsAlmostSimpleGroup );
 
+# a simple group is almost simple
+InstallTrueMethod( IsAlmostSimpleGroup, IsSimpleGroup );
+
 
 #############################################################################
 ##
@@ -1221,7 +1224,9 @@ DeclareAttribute("TryMaximalSubgroupClassReps",IsGroup,"mutable");
 # utility function in maximal subgroups code
 DeclareGlobalFunction("TryMaxSubgroupTainter");
 DeclareGlobalFunction("DoMaxesTF");
-DeclareGlobalFunction("MaxesAlmostSimple");
+
+# make this an operation to allow for overloading and TryNextMethod();
+DeclareOperation("MaxesAlmostSimple",[IsGroup]);
 
 #############################################################################
 ##
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -701,6 +701,18 @@ InstallMethod( AbelianInvariants,
     return inv;
     end );
 
+InstallMethod( AbelianRank ,"generic method for groups", [ IsGroup ],0,
+function(G)
+local a,r;
+  a:=AbelianInvariants(G);
+  r:=Number(a,IsZero);
+  a:=Filtered(a,x->not IsZero(x));
+  if Length(a)=0 then return r; fi;
+  a:=List(Set(a,SmallestRootInt),p->Number(a,x->x mod p=0));
+  return r+Maximum(a);
+end);
+
+
 #############################################################################
 ##
 #M  IsInfiniteAbelianizationGroup( <G> ) 
--- a/lib/grpffmat.gi
+++ b/lib/grpffmat.gi
@@ -125,13 +125,25 @@ end );
 ##
 #M  NiceMonomorphism( <ffe-mat-grp> )
 ##
+MakeThreadLocal("FULLGLNICOCACHE"); # avoid recreating same homom. repeatedly
+FULLGLNICOCACHE:=[];
 InstallGlobalFunction( NicomorphismFFMatGroupOnFullSpace, function( grp )
     local   field,  dim,  V,  xset,  nice;
     
     field := FieldOfMatrixGroup( grp );
     dim   := DimensionOfMatrixGroup( grp );
+
+    #check cache
+    V:=Size(field);
+    nice:=First(FULLGLNICOCACHE,x->x[1]=V and x[2]=dim);
+    if nice<>fail then return nice[3];fi;
+
+    if not (HasIsNaturalGL(grp) and IsNaturalGL(grp)) then
+      grp:=GL(dim,field); # enforce map on full GL
+    fi;
     V     := field ^ dim;
     xset := ExternalSet( grp, V );
+
 
     # STILL: reverse the base to get point sorting compatible with lexicographic
     # vector arrangement
@@ -144,6 +156,16 @@ InstallGlobalFunction( NicomorphismFFMatGroupOnFullSpace, function( grp )
     SetFilterObj(nice,IsNiceMonomorphism);
     # because we act on the full space we are canonical.
     SetIsCanonicalNiceMonomorphism(nice,true);
+    if Size(V)>10^5 then 
+      # store only one big one and have it get thrown out quickly
+      FULLGLNICOCACHE[1]:=[Size(field),dim,nice];
+    else
+      if Length(FULLGLNICOCACHE)>4 then
+        FULLGLNICOCACHE:=FULLGLNICOCACHE{[2..5]};
+      fi;
+      Add(FULLGLNICOCACHE,[Size(field),dim,nice]);
+    fi;
+
     return nice;
 end );
 
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1519,7 +1519,7 @@ end );
 ##
 #M  Socle( <G> )  . . . . . . . . . . .  socle of primitive permutation group
 ##
-InstallMethod( Socle,"for permgrp", true, [ IsPermGroup ], 0,
+InstallMethod( Socle,"test primitive", true, [ IsPermGroup ], 0,
     function( G )
     local   Omega,  deg,  shortcut,  coll,  d,  m,  c,  ds,  L,  z,  ord,
             p,  i;
@@ -1539,25 +1539,27 @@ InstallMethod( Socle,"for permgrp", true, [ IsPermGroup ], 0,
     deg := Length( Omega );
     if deg >= 6103515625  then
         TryNextMethod();
-    elif deg < 12960000  then
-        shortcut := true;
-        if deg >= 3125  then
-            coll := Collected( Factors(Integers, deg ) );
-            d := Gcd( List( coll, c -> c[ 2 ] ) );
-            if d mod 5 = 0  then
-                m := 1;
-                for c  in coll  do
-                    m := m * c[ 1 ] ^ ( c[ 2 ] / d );
-                od;
-                if m >= 5  then
-                    shortcut := false;
-                fi;
-            fi;
-        fi;
-        if shortcut  then
-            ds := DerivedSeriesOfGroup( G );
-            return ds[ Length( ds ) ];
-        fi;
+    # the normal closure actually seems to be faster than derived series, so
+    # this is not really a shortcut
+    #elif deg < 12960000  then
+    #    shortcut := true;
+    #    if deg >= 3125  then
+    #        coll := Collected( Factors( deg ) );
+    #        d := Gcd( List( coll, c -> c[ 2 ] ) );
+    #        if d mod 5 = 0  then
+    #            m := 1;
+    #            for c  in coll  do
+    #                m := m * c[ 1 ] ^ ( c[ 2 ] / d );
+    #            od;
+    #            if m >= 5  then
+    #                shortcut := false;
+    #            fi;
+    #        fi;
+    #    fi;
+    #    if shortcut  then
+    #        ds := DerivedSeriesOfGroup( G );
+    #        return ds[ Length( ds ) ];
+    #    fi;
     fi;
     
     coll := Collected( Factors(Integers, Size( G ) ) );
@@ -1577,7 +1579,10 @@ InstallMethod( Socle,"for permgrp", true, [ IsPermGroup ], 0,
         until ord mod p = 0;
         z := z ^ ( ord / p );
     until Index( G, Centralizer( G, z ) ) mod p <> 0;
-    L := NormalClosure( G, SubgroupNC( G, [ z ] ) );
+
+    # immediately add conjugate as this will be needed anyhow and seems to
+    # speed up the closure process
+    L := NormalClosure( G, SubgroupNC( G, [ z,z^Random(G) ] ) );
     if deg >= 78125  then
         ds := DerivedSeriesOfGroup( L );
         L := ds[ Length( ds ) ];
--- a/lib/maxsub.gi
+++ b/lib/maxsub.gi
@@ -536,42 +536,16 @@ local hom,embs,s,k,agens,ad,i,j,perm,dia,ggens,e,tgens,d,m,reco,emba,outs,id;
   return m;
 end);
 
-InstallGlobalFunction(MaxesAlmostSimple,function(G)
-local id,m,epi,cnt,h;
+InstallMethod(MaxesAlmostSimple,"fallback to lattice",true,[IsGroup],0,
+function(G)
+  if ValueOption("cheap")=true then return [];fi;
+  Info(InfoLattice,1,"MaxesAlmostSimple: Fallback to lattice");
+  return MaxesByLattice(G);
+end);
 
-  # is a permutation degree too big?
-  if IsPermGroup(G) then
-    if NrMovedPoints(G)>
-        SufficientlySmallDegreeSimpleGroupOrder(Size(PerfectResiduum(G))) then
-      h:=G;
-      for cnt in [1..5] do
-	epi:=SmallerDegreePermutationRepresentation(h);
-	if NrMovedPoints(Range(epi))<NrMovedPoints(h) then
-	  m:=MaxesAlmostSimple(Image(epi,G));
-	  m:=List(m,x->PreImage(epi,x));
-	  return m;
-	fi;
-	# new group to avoid storing the map
-	h:=Group(GeneratorsOfGroup(G));
-	SetSize(h,Size(G));
-      od;
-    fi;
-
-
-    # Are just finding out that a group is symmetric or alternating?
-    # if so, try to use method that uses data library
-    if 
-    #not (HasIsNaturalSymmetricGroup(G) or HasIsNaturalAlternatingGroup(G)) and
-        (IsNaturalSymmetricGroup(G) or IsNaturalAlternatingGroup(G)) then
-      Info(InfoLattice,1,"MaxesAlmostSimple: Use S_n/A_n");
-      m:=MaximalSubgroupsSymmAlt(G,false);
-      if m<>fail then
-	return m;
-      fi;
-    fi;
-  fi;
-
-  # is the degree bad?
+InstallMethod(MaxesAlmostSimple,"table of marks and linear",true,[IsGroup],0,
+function(G)
+local m,id,epi;
 
   # does the table of marks have it?
   m:=TomDataMaxesAlmostSimple(G);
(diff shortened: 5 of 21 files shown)


lib/grp.gi
```diff
@@ -701,6 +701,18 @@ InstallMethod( AbelianInvariants,
...
     end );
 
+InstallMethod( AbelianRank ,"generic method for groups", [ IsGroup ],0,
+function(G)
+local a,r;
+  a:=AbelianInvariants(G);
+  r:=Number(a,IsZero);
+  a:=Filtered(a,x->not IsZero(x));
+  if Length(a)=0 then return r; fi;
+  a:=List(Set(Factors(Product(a))),p->Number(a,x->x mod p=0));
```
fingolfin: Multiplying prime powers together only to factor them again is wasteful. This could be avoided e.g. as follows:
```
a:=List(Set(a, SmallestRootInt), p->Number(a,x->x mod p=0));
```

lib/matrix.gd
```diff
@@ -1677,6 +1677,8 @@ DeclareGlobalFunction( "RandomMat" );
 ##  returns a new random mutable <A>m</A><M>\times</M><A>m</A> matrix with integer
 ##  entries that is invertible over the integers.
 ##  Optionally, a random source <A>rs</A> can be supplied.
+##  If the option <A>domain</A> is given, random seection is made from <A>domain</A>, otherwise
```
fingolfin: typo `seection`

hulpke: Fixed.

lib/grp.gi
hulpke: If this is considered nicer, I'll change it.
In practice this is a purely academic argument as we are talking about factorization into known primes, i.e. trial division.

fingolfin: Any review remark I make of course just constitutes my personal opinion, and you are of course free to ignore it.

Anyway, all in all, over the past months I got a strong feeling that my remarks on your PRs are perceived more as nuisance than useful, and as "wasting your time". Therefore, in order to not waste *both* our time, I will refrain from reviewing your PRs for the time being, and leave it to others to take care of that.

hulpke:
Dear @fingolfin 

> Any review remark I make of course just constitutes my personal opinion, and you are of course free to ignore it.

Well, since proposed changes cannot be merged in without an approving review, any review that comments but does not approve is implicitly stating that these are issues without which the proposed contribution should be rejected.

> Anyway, all in all, over the past months I got a strong feeling that my remarks on your PRs are perceived more as nuisance than useful, and as "wasting your time".

I know that I do not react well to stylistic suggestions of changes. I'm sorry.
In fact, for changes like these (which are cleanup and bug fixes) I am not soliciting for reviews but for the fact that I am blocked from merging without an approving review.

> Therefore, in order to not waste both our time, I will refrain from reviewing your PRs for the time being, and leave it to others to take care of that.

Then, why do we have the enforced review for every contribution? I fear that we have technical features of github dictate the way how we collaborate.

fingolfin:
> > Any review remark I make of course just constitutes my personal opinion, and you are of course free to ignore it.
> 
> Well, since proposed changes cannot be merged in without an approving review, any review that comments but does not approve is implicitly stating that these are issues without which the proposed contribution should be rejected.

I disagree with this interpretation of what a "review that comments but does not approve" means. It would ultimately imply that one must not comment on a PR, ever, unless to either approve it or reject it. So, if I spot a typo, I shouldn't point it out, unless I am willing and able to review the whole PR? That makes no sense to me. Esp. bigger PRs are hard to review in one go, so a partial review sometimes is the best I can afford with limited time and energy.

In any case, this interpretation does not match my intentions at all. Indeed, if I reject a PR in its current form, I use the "request changes" option. If I think a PR could be merged in its current form (even if there are some changes I'd prefer to see), I'll approve it. If I neither "approve" nor "request changes", that can mean a lot of things: that I did not have time to review it all; that I do not feel informed enough to express an opinion either way; that I am conflicted about it; or something completely else. And based on past discussions with other GAP developers, many (most?) seem to have a rather similar view on this. But they can and should speak for themselves, of course.

In the case of this particular PR, it's really that I did not have the energy to slog through all of it. There are tons of different, unrelated changes mingled up together in it. IMHO, if instead of this, it was split into several smaller commits, somebody likely would have reviewed most or even all of them already. As it is, I'd have to either review all of them (for which I don't have the energy, and quite frankly, also don't see why I should spend the effort to accommodate you, when I'd tell anybody else to split up the PR into multiple). Or I could either blindly rubber stamp it (but I don't like doing that); or request that you break it up (but based on past experience, I don't expect you to react favorably to this, so I didn't bother to do so). Nor do I want to reject it, as I have no actual concerns about it (by virtue of not having reviewed it).

Thus, I am instead leaving it for somebody else to review.


> 
> > Anyway, all in all, over the past months I got a strong feeling that my remarks on your PRs are perceived more as nuisance than useful, and as "wasting your time".
> 
> I know that I do not react well to stylistic suggestions of changes. I'm sorry.
> In fact, for changes like these (which are cleanup and bug fixes) I am not soliciting for reviews but for the fact that I am blocked from merging without an approving review.

Well, here we have a fundamental disagreement then. In my view, every change should be reviewed by at least one other party. From that view point, it does not matter if you are soliciting for reviews: any changes that goes in is reviewed, period. To avoid a misunderstanding, let me emphasize that I expressed that this is simply *my personal view* (although I believe that several other GAP developers share it), and that it's not me being a dictator blocking your PR out of a whim. You can raise this on the mailing list if you like, perhaps the GAP team can agree on a different approach.


> 
> > Therefore, in order to not waste both our time, I will refrain from reviewing your PRs for the time being, and leave it to others to take care of that.
> 
> Then, why do we have the enforced review for every contribution? I fear that we have technical features of github dictate the way how we collaborate.

This is not a github dictate; it's a setting we chose to activate. And which we can of course deactivate again, should we (as a collective) decide to do so. Feel free to bring this up on one of the mailing lists for discussion, too!

Here's my view on it, though:

We enforce reviews of every contribution to ensure a certain minimal quality level for them. E.g. that they are understandable, sufficiently explained/documented, accompanied by sensible tests, etc. To be frank, I usually have (mild!) concerns about that with your PRs (e.g. lack of tests -- and I am not talking about tests for performance improvements, where you explain that the tests you have take hours or days -- that's of course perfectly reasonable!; I am talking about tests that verify that e.g. a bug fix does actually fix a bug; or tests that verify that a given change does not break existing functionality). That said, so far I usually was happy to give you far more leeway than I'd give almost any other contributor with my reviews, because I of course know you are a master of your domain, and you make very valuable contributions.

However, I learned that you do not appreciate the feedback I give on that work, and even complained to others about it (indirectly at least). Very well. I could write a lot about that, but in the end, I'll just accept it for the time being, and move on, and, as I wrote, not waste both our time with it then.

But it's not my intention to block your PRs with that! Rather, I expect others to review them. There is no magical law that says that I am the only one who can review your PRs.
If I may suggest, you can do what I often do if I don't get a review on a PR quickly: I specifically request reviews from other people (Chris, Markus, Steve, Frank, Thomas, ...), via the GitHub UI for that -- I am hopeful that they will either review your PR, or explain why they won't.

That said, we do indeed have a shortage of people who actively review PRs. That is a problem. And perhaps it means we should stop requiring reviews again (although personally I'd consider it a shame). Once again, this is of course something that can be discussed, ideally on the mailing list.

alex-konovalov:
Perhaps would be good to discuss on a mailing list and formulate a policy and explain the reasoning behind some decisions (like above). I would also consider stopping requiring reviews to be a shame. Reviewing is hard, and it takes time, but it's useful when others look at the code, not only because they may have suggestions, but also because they will learn something by looking at it. And we do manage to merge PRs - for a long time, we keep the number of PRs below 50, and we have 1831 closed PR by now! 

OTOH, the larger is the change, the harder is to review it. @hulpke I looked at your PR when it appeared, and seen that the earliest commit sits there since May. It's good that your changes are kept in individual commits, but wouldn't be more efficient to go further and keep unrelated changes in separate PRs, and submit them as soon as they are ready? In particular, we plan to branch off stable-4.10 on September 1st, and seems unlikely that this PR will go there (cherry-picking later will require extra work), while having some of these commits merged earlier would help to that...

hulpke:
@alex-konovalov 

I will try to write to the mailing list in the next days. I find the current setup deeply discouraging to contributions.

> OTOH, the larger is the change, the harder is to review it.

You can click on the individual commits (in the same time required as if they were several PR) and see the (often pathetically trivial) content.

> wouldn't be more efficient to go further and keep unrelated changes in separate PRs, and submit them as soon as they are ready

If the main part of GAP development is the reviewing and wrapping process, then yes.  Maybe even if these are deliberate changes that happen as main work.
In my work these changes came up ``on the side'' when doing other work, which I did not want to interrupt more for doing a PR every time. Frankly, the alternative is not to do a PR, but to invite others to cherry pick such changes off my development branch.
This PR is the result of an attempt to get my Working branch back closer to the master branch. I will aim to move this into smaller units, but I cannot envision to have something like this become 15 separe PRs.
If there are significant changes (as #2741) I am doing these already as individual PR.

> In particular, we plan to branch off stable-4.10 on September 1st, and seems unlikely that this PR will go there

There once was a policy that releases would aim to correct all known bugs. They would certainly include all bugfixes that had been made available. It is disheartening to see this go.

lib/grp.gd
```diff
@@ -858,6 +861,10 @@ InstallTrueMethod( IsPolycyclicGroup,
 ##
 DeclareAttribute( "AbelianInvariants", IsGroup );
 
+# minimal number of generators for abelianization. Used internally to check
+# whether it is worth to attempt to reduce generator number
+DeclareAttribute( "AbelianRank", IsGroup );
```
alex-konovalov: [Question] would it be useful to document this attribute?

lib/maxsub.gi
```diff
@@ -595,10 +569,41 @@ local id,m,epi,cnt,h;
     fi;
 
   fi;
+  TryNextMethod();
+end);
 
-  if ValueOption("cheap")=true then return [];fi;
-  Info(InfoLattice,1,"MaxesAlmostSimple: Fallback to lattice");
-  return MaxesByLattice(G);
+InstallMethod(MaxesAlmostSimple,"permutation group",true,[IsPermGroup],0,
```
alex-konovalov: [Question] Do we have any guidance regarding writing the free text comment? Is "permutation group" ok here, of "for permutation group" would be better?

lib/matrix.gi
```diff
@@ -3651,8 +3651,13 @@ end );
 #F  RandomUnimodularMat( [rs ,] <m> )  . . . . . . . random unimodular matrix
 ##
 InstallGlobalFunction( RandomUnimodularMat, function ( arg )
-    local  rs, m, mat, c, i, j, k, l, a, b, v, w, gcd;
+local  rs, m, mat, c, i, j, k, l, a, b, v, w, gcd,dom;
```
alex-konovalov: [Minor] Sorry for the nitpicking, but why there are spaces between all commas except the last one?

lib/matrix.gd
```diff
@@ -1677,6 +1677,8 @@ DeclareGlobalFunction( "RandomMat" );
 ##  returns a new random mutable <A>m</A><M>\times</M><A>m</A> matrix with integer
 ##  entries that is invertible over the integers.
 ##  Optionally, a random source <A>rs</A> can be supplied.
+##  If the option <A>domain</A> is given, random selection is made from <A>domain</A>, otherwise
+##  from <A>Integers</A>
 ##  <Example><![CDATA[
```
alex-konovalov: [Suggestion] I think it will be useful to have an example demonstrating how this option is used, and also indicate that in the `Func` element above. This will also make the usage of `<A>` less confusing.

lib/permutat.g
```diff
@@ -780,6 +780,7 @@ InstallMethod( Order,
     [ IsPerm ],
     ORDER_PERM );
 
+
 #############################################################################
```
alex-konovalov: [Minor] Accidental empty line pollutes changes history for this file.

lib/grpperm.gi
```diff
@@ -1539,25 +1539,27 @@ InstallMethod( Socle,"for permgrp", true, [ IsPermGroup ], 0,
...
-                if m >= 5  then
-                    shortcut := false;
-                fi;
-            fi;
-        fi;
-        if shortcut  then
-            ds := DerivedSeriesOfGroup( G );
-            return ds[ Length( ds ) ];
-        fi;
+    # the normal closure actually seems to be faster than derived series, so
```
alex-konovalov: [Maybe] Remove the code instead of commenting it out, and describe in the comment that formerly there was some code here which was removed because of some reasons.

lib/grpffmat.gi
```diff
@@ -125,11 +125,18 @@ end );
 ##
 #M  NiceMonomorphism( <ffe-mat-grp> )
 ##
+FULLGLNICOCACHE:=[];
```
alex-konovalov: [Question] This will certainly require attention in HPC-GAP. Is this code covered by test? Codecov is not accessible right now, so I can't check.

hulpke: AH, yes. This is triggered as soon as a permutation representation for GL (or a subgroup on the whole space) is calculated, that will be part of existing tests.

alex-konovalov:
It looks good to me - I suggest to look at least at the comment marked [Suggestion] because it will help to make documentation more clear. Also, before merging you need to do rebase and instead of two separate commits dealing with comments they should be incorporated into commits to which those changes belong. With `rebase -i` that should not be a problem.

alex-konovalov:
@hulpke wrote
> There once was a policy that releases would aim to correct all known bugs. They would certainly include all bugfixes that had been made available. It is disheartening to see this go. 

Of course we *aim* to correct all known bugs! But we do not want neverending release cycles, when we are waiting for some bugfixes to appear in the release branch, and thus delaying giving to the users access to all other bugfixes and new features that are already there. On the other hand, we tend to release often, so this PR has all chances to appear in GAP 4.11, and perhaps parts of it will be cheery-picked onto stable-4.10 after all...

lib/grp.gd
hulpke: My concern was that this could lead to confusion as `AbelianInvariants` will (in general) return a longer list by separating different primes, while this attribute gives the minimal generator number of $G/G'$.

lib/matrix.gd
hulpke: I added an example. This will require two iterations as the manual check will discover the appropriate random values.

alex-konovalov: First, there is a typo `domail`. Second, what about

```
<Func Name="RandomUnimodularMat" Arg='[rs ,] m [:domain:=d]'/>
...

##  If the option <C>domain</C> is given, random selection is made from the domain
##  <A>d</A> (allowing for larger coefficients), otherwise the default domain will be <Ref Var="Integers"/>.
```

lib/grpperm.gi
hulpke: I don't think I have run or analyzed enough examples to state categorically that the old (now commented out) code is clearly useless. Someone (well, myself) might stumble on this place again when looking at an inefficiency, and it might turn out that under certain circumstances the now commented-out code is in fact better. If it is immediately removed, it is hard to recover it from git.

lib/grpffmat.gi
alex-konovalov: Thanks for comment - then in HPC-GAP that could be a thread local variable, so each thread can assign a distinct value to it: http://www.gap-system.org/Manuals/doc/hpc/chap2.html#X7D93681D7B5E8DCD

lib/matrix.gd
alex-konovalov: Good example, which uncovers my misunderstanding. I thought that a domain should be a domain in a GAP sense (especially since `Integers` is a domain). But that's not necessary: 
```
gap> IsDomain([-100..100]);
false
```
Maybe a different word/option name to avoid confusion? I guess `d` should be a collection for the code to work?

hulpke: The option is now called `random` for lack of a better word. (It is the set from which random numbers are chosen, matrix coeffs can be larger.

alex-konovalov:
I also forgot to check commit messages (perhaps we do need a template for PRs with tickboxes for reviewers?)

```
CLEANUP: Make `MaxesAlmostSimple` an operation so packages can overload

new methods. Also have simple group imply almost simple group.
```

and 
```
CLEANUP: Use new attribute `AbelianRank` to clean up gen. nr. arguments

in multiple places.
```

would look better and match recommendations from [here](https://chris.beams.io/posts/git-commit/) if they could be rewritten to at least something like

```
CLEANUP: Make `MaxesAlmostSimple` an operation 

This will allow packages to install new methods for it.
Also have simple group imply almost simple group.
```

and 
```
CLEANUP: Introduce new attribute `AbelianRank`

This was used to clean up gen. nr. arguments in multiple places.
```
(also "gen. nr. arguments" could now be expanded).

alex-konovalov:
@hulpke just one more idea - we can avoid the necessity of choosing a name for the option for the domain in `RandomUnimodularMat` by making use of variadic functions ... and, oh, it is actually already variadic, and it has uses one argument `arg`. Then it's not hard to analyse the length of `arg` and in case it is 2, detect whether this is the case of `rs,m` or `m,range`. 

Since options stack is global, and there is no way to check when the option is misspelt, using variadic functions is IMHO more robust.

hulpke:
@alex-konovalov 
Yes, I had thought about the `arg` option, but three reasons spoke against it:
- It gets more complicated with optional arguments at front and back
- Once this becomes a complete syntax option, the code should deal with passing arbitrary rings for which `unimodular` makes sense -- padics, polynomial rings etc.
- Since I called the function from within another (private) function, use of an option allowed to pass it through the outer function without affecting its syntax.

Thinking of the last option, maybe it actually should be an option for the `Random` method for integers. I'll remove this commit and think about it.

alex-konovalov:
@hulpke I tested this PR in Jenkins to run extended tests with all packages loaded etc. All went well except that there is one diff when all packages are loaded - this should be fixed before merging:

```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-pull-request-quickt\
est/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/standard/label/kovacs/GAP-pull-reque\
st-snapshot/tst/testinstall/morpheus.tst:38
# Input is:
Size(AutomorphismGroup(g))/Size(g);
# Expected output:
24
# But found:
Error, usage: Image(<map>), Image(<map>,<elm>), Image(<map>,<coll>)
########
```

hulpke:
@alex-konovalov 
I have difficulty reproducing this bug, even with packages loaded. Is there any further information about it that could be given (can you email me the full tracelog)? Are these the packages as in the current `pkg` archive on the web pages, or newer versions?

alex-konovalov:
@hulpke this is with current package-master.tar.gz linked from https://github.com/gap-system/gap-distribution. There is no backtrace for this error in the test log, and in a fresh checkout of your branch with the said package archive I also can't reproduce it. Noticed that you have updated this PR - maybe that's why?

Also, while `morpheus.tst` is from `testinstall`, it did run well in `make testinstall` - the problem only showed up in Jenkings when `make teststandard` was used. Might be some interaction with other tests...

hulpke:
@alex-konovalov 
I similarly had a weird error in the package test on travis. The last update of the PR was simply a rebase to force a renewed test. If neither of us can reproduce the problem now, I suspect that it was an issue only with the previous branch and not a real error.

fingolfin:
I just tried to reproduce this weird error in `morpheus.tst` in various constellations (with all packages loaded; with and without teststandard, etc.), to no avail. I agree that it looks like it is not directly related to this PR, esp. if it is gone now on Jenkins, too (is it?).

That said, it's still somewhat worrying. I have a hard time imagining any reason that would cause such a test failure "randomly". If this failure happened on that Mac OS X test server in St Andrews, I'd suspect a HW fault, but on kovacs...?

alex-konovalov:
I am convinced that this is a genuine error. 

@hulpke if the error in the package test on travis was about missing symbol "atexit", please ignore it - that was due to some Travis irregularities. We have now disabled building packages that trigger the problem (cf. #2768), and @fingolfin will hopefully fix this properly in #2759.

I have rerun it today in Jenkins, and it happened again - both on kovacs and macOS nodes, and both in 32- and 64-bit builds. I am now running it manually on kovacs, trying to catch some more details.

alex-konovalov:
Update: reproduced interactively in the workspace of the failed Jenkins build:
1. gap.sh -r -A
2. LoadAllPackages();
3. Read("tst/teststandard.g");

I have modified `tst/teststandard.g` to not to quit from GAP after the test is finished. I hoped that after the full test, `Test("tst/testinstall/morpheus.tst")` will fail and I will be able to investigate the break loop by trying the example from the test - but it worked!

Also, I tried to exclude tests from `tst/testextra` from `tst/teststandard.g` but the the test passed. So my conjecture is that this is triggered by loading all packaged and running one of the tests from `tst/testextra`.

hulpke:
@alex-konovalov 
Thank you. I can reproduce it this way and will investigate.

hulpke:
@alex-konovalov 
OK, I found the (after all pathetically stupid) bug. No idea why only loading packages triggers it, but it is fixed now and there is a (much easier) test. I'll check that tests succeed now and then will be rebase once more on master.

alex-konovalov:
Thanks @hulpke - I will rerun the tests now.

alex-konovalov:
@hulpke now teststandard passes@. Next, do you plan to rebase after I've run the tests, and use the two last commits to fix some other commits in this branch?

hulpke:
@alex-konovalov 
Yes, the two fixes (well, one fix and a test file) are now rebased put under caching the GL homomorphisms.

alex-konovalov:
For the record, I've compared performance in this PR and in nightly builds yesterday.

1. teststandard

Linux:
* master: total   2914925 ms (799008 ms GC) and 490GB allocated
* this PR: total   2774530 ms (782115 ms GC) and 506GB allocated

macOS:
* master: total   5174297 ms (1372988 ms GC) and 446GB allocated
* this PR: total   5288641 ms (1438456 ms GC) and 467GB allocated

2. testinstall

Linux:
* master: total     86346 ms (43542 ms GC) and 10.5GB allocated
* this PR: total     91369 ms (45655 ms GC) and 10.4GB allocated

macOS:
* master: total    174664 ms (82735 ms GC) and 10.5GB allocated 
* this PR: total    168886 ms (80093 ms GC) and 10.4GB allocated

alex-konovalov:
Thanks @hulpke !

PR #2738, opened 2018-08-24
kernel: turn ariths.h macros into functions

fingolfin:
Now one can use SUM, PROD, EQ etc. from the debugger

### DIFF
--- a/src/pperm.c
+++ b/src/pperm.c
@@ -2322,15 +2322,16 @@ Obj FuncAS_PERM_PPERM(Obj self, Obj f)
     UInt2 *ptf2, *ptp2;
     UInt4 *ptf4, *ptp4;
     UInt   deg, i, j, rank;
-    Obj    p, dom;
+    Obj    p, dom, img;
 
+    img = FuncIMAGE_SET_PPERM(self, f);
+    dom = DOM_PPERM(f);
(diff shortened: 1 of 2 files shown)


src/pperm.c
```diff
@@ -2322,15 +2322,16 @@ Obj FuncAS_PERM_PPERM(Obj self, Obj f)
...
+    Obj    p, dom, img;
 
+    img = FuncIMAGE_SET_PPERM(self, f);
+    dom = DOM_PPERM(f);
+    if (!EQ(img, dom)) {
+        return Fail;
+    }
     if (TNUM_OBJ(f) == T_PPERM2) {
-        if (!EQ(FuncIMAGE_SET_PPERM(self, f), DOM_PPERM(f))) {
-            return Fail;
```
fingolfin: Note: the old EQ macro evaluated both operands up to 4 (!) times, and strictly in the order they appear as arguments. With EQ a function, the evaluation order can and does change; but DOM_PPERM return NULL if INIT_PPERM2/4 is not called before, which then leads to a crash inside EQ.

fingolfin:
Thanks. A large chunk of the changes was done via regex search & replace, in order to reduce the risk of introducing mistakes,  ut of course it is always good to double check!

PR #2740, opened 2018-08-24
Replace macros with functions in records.h

ChrisJefferson:
As in title. There didn't turn out to be any interesting asserts to apply here, as these functions accept any type (and TNUM_OBJ checks if the input is a valid Obj).

### DIFF
--- a/src/records.h
+++ b/src/records.h
@@ -63,16 +63,13 @@ extern  UInt            RNamObj (
 **
 *F  IS_REC(<obj>) . . . . . . . . . . . . . . . . . . . is an object a record
 *V  IsRecFuncs[<type>]  . . . . . . . . . . . . . . . . table of record tests
-**
-**  'IS_REC' returns a nonzero value if the object <obj> is a  record  and  0
-**  otherwise.
-**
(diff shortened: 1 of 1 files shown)


src/records.h
```diff
@@ -63,16 +63,13 @@ extern  UInt            RNamObj (
...
-**  Note that 'IS_REC' is a macro, so do not call  it  with  arguments  that
-**  have side effects.
 */
-#define IS_REC(obj)     ((*IsRecFuncs[ TNUM_OBJ(obj) ])( obj ))
+extern Int (*IsRecFuncs[LAST_REAL_TNUM + 1])(Obj obj);
 
-extern  Int             (*IsRecFuncs[LAST_REAL_TNUM+1]) ( Obj obj );
+static inline Int IS_REC(Obj obj)
+{
+    return (*IsRecFuncs[TNUM_OBJ(obj)])(obj);
```
fingolfin: One assert that one *could* add here is `GAP_ASSERT(TNUM(obj) <= LAST_REAL_TNUM)`; same for my PR #2738.

Not saying you should, but since you remarked that there are no "interesting" asserts to be added... ;-) (these asserts would e.g. catch if one was accessing an object which is in the middle of being copied).

PR #2741, opened 2018-08-24
ENHANCE: Improvements to double coset calculations for hard cases

hulpke:
Allow for larger up-steps, and improve up-step fusion speed.
Avoid rechecks of transversal steps.

The example I have at hand (H\G/H for G=Fi23, H=S12) runs about two days and thus is not suitable for automatic testing. 

This is unlikely to be urgent, as few users will do such calculations.

### DIFF
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -851,12 +851,14 @@ local live,orbs,orbset,done,nr,p,o,os,orbitextender,bahn,i,j,enum,dict,map,pam;
               if IsBound(map[j]) and map[j]>i then map[j]:=map[j]-1; fi;
             od;
 
-            p:=[i..Length(orbs)-1];
-            orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
-            orbset{p}:=orbset{p+1};Unbind(orbset[Length(orbset)]);
-            done{p}:=done{p+1};Unbind(done[Length(done)]);
-            bahn{p}:=bahn{p+1};Unbind(bahn[Length(bahn)]);
-            pam{p}:=pam{p+1};Unbind(pam[Length(pam)]);
+            # Remove entry i, i.e.
+            #p:=[i..Length(orbs)-1];
+            #orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
+            Remove(orbs,i);
+            Remove(orbset,i);
+            Remove(done,i);
+            Remove(bahn,i);
+            Remove(pam,i);
           else
             done[i]:=-p;
           fi;
@@ -1029,6 +1031,12 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
 
     fi;
 
+  elif ValueOption("sisyphus")=true then
+    # purely to allow for tests of up-step mechanism in smaller examples.
+    # This is creating unneccessary extra work and thus should never be used
+    # in practice, but will force some code to be run through.
+    c:=Concatenation([TrivialSubgroup(G)],c);
+    cano:=true;
   fi;
 
   r:=[One(G)];
--- a/lib/csetgrp.gi
+++ b/lib/csetgrp.gi
@@ -143,7 +143,19 @@ local cla,clb,i,j,k,imgs,bd,r,rep,b2,ex2,split,dc,
     return found;
   end;
 
-  gens:=MorFindGeneratingSystem(b,MorMaxFusClasses(MorRatClasses(b)));
+  # don't try the `MorGen...` search for more than two generators if
+  # generator number seems OK
+  if Length(SmallGeneratingSet(b))=AbelianRank(b) and
+    Length(SmallGeneratingSet(b))>2 then
+    gens:=SmallGeneratingSet(b);
+  elif IsPermGroup(b) and Size(b)<RootInt(NrMovedPoints(b)^3,2) then
+    r:=SmallerDegreePermutationRepresentation(b);
+    k:=Image(r,b);
+    gens:=MorFindGeneratingSystem(k,MorMaxFusClasses(MorRatClasses(k)));
+    gens:=List(gens,x->PreImagesRepresentative(r,x));
+  else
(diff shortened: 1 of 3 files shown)


lib/csetgrp.gi
```diff
@@ -749,6 +762,126 @@ local l;
...
+              if IsBound(map[j]) and map[j]=i then map[j]:=p; fi;
+            od;
+
+            # delete entry i, move higher ones one up
+            for j in [1..Length(map)] do
+              if IsBound(map[j]) and map[j]>i then map[j]:=map[j]-1; fi;
+            od;
+
+            p:=[i..Length(orbs)-1];
+            orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
```
fingolfin: Am I misunderstanding something, or is this just a complicated and less efficient way to say `Remove(orbs, i)` ... ? (It's less efficient because instead of just resulting in a `memmove`, it first creates a new plist with the values at the end and then copies them back; so beside copying twice as much data, it also creates a temporary object which adds pressure to the memory manager.

hulpke: Yes, it is. (I did not know `Remove`. I'll change it (but as `orbs` is short this will hardly make any impact on time or memory use).

hulpke:
@fingolfin 
I have added an option that will force an (otherwise pointless) recalculation that will however trigger the new code, as well as a test. Should codecov at some point look at this PR, it should notice this.

Personally, I think, this does not add significantly about correctness of the code beyond my claim that I used it to run through an otherwise impossible example. But since its part of the automatic tests it would spot problems that would come up if a function used by the new code changed.

> Just like I'd not want to put a theorem into a paper without a proof,

I think this analog fails: It is not a single-author paper, but a coauthor who claims that  by a general principle from a difficult area a certain fact can be stated without giving an explicit proof; respectively that they have proven the statement in the past but did not write it down.

Yes, in the best of the worlds there would be accompanying detailed documentation. Under real world constraints -- and this is something our development model has recognized in the past -- I find myself with insufficient time to do so. (Or, to phrase it more pointedly: If the requirement for GAP contributions would have been on the level as you aspire to, there would have been far less functionality, in particular as it comes for making calculations run better, and I would not have a tenured faculty position.)

As for the code itself, I would argue (though I recognize that others might disagree) that what it does -- fusing subgroup orbits given by representatives by calculating parts of the orbits until orbits overlap -- is not more cryptic than  other changes that have been done and for which I have to trust the respective authors (e.g. the configuration setup, changes for C++ and for HPCGAP) and for which only one person might know the details.

I agree that in the long run `DCFuseSubgroupOrbits` should be a more general function with a cleaner interface. But if I wait for this the code will build up more interdependencies and I would end up again with a large PR as has been deemed undesirable.

alex-konovalov:
@hulpke thanks - with a duration of 2-3 minutes, actually fine to put in into `teststandard` or `testextra`. The former will be tested on Travis with this PR, the latter - only on Jenkins.

lib/csetgrp.gi
```diff
@@ -749,6 +762,131 @@ local l;
...
+              if IsBound(map[j]) and map[j]>i then map[j]:=map[j]-1; fi;
+            od;
+
+            #p:=[i..Length(orbs)-1];
+            #orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
+            #orbset{p}:=orbset{p+1};Unbind(orbset[Length(orbset)]);
+            #done{p}:=done{p+1};Unbind(done[Length(done)]);
+            #bahn{p}:=bahn{p+1};Unbind(bahn[Length(bahn)]);
+            #pam{p}:=pam{p+1};Unbind(pam[Length(pam)]);
+            Remove(orbs,i);
```
alex-konovalov: If `Remove` works, the commented out code above can be removed.

hulpke:
> @hulpke thanks - with a duration of 2-3 minutes, actually fine to put in into `teststandard` or `testextra`. The former will be tested on Travis with this PR, the latter - only on Jenkins.

But its not 2-3 but rather 10 minutes. Do you still want it?

alex-konovalov:
Misread that - assumed it was 10 minutes before speeding up. In this case, please out it into `tst/testextra` - it's tolerable there.

tst/testextra/grplatt.tst
```diff
@@ -193,6 +193,13 @@ gap> Length(ConjugacyClassesSubgroups(SymmetricGroup(7):NoPrecomputedData));
 #I  Using (despite option) data library of perfect groups, as the perfect
 #I  subgroups otherwise cannot be obtained!
 96
+gap> g:=SimpleGroup("Co3");;
```
alex-konovalov: Please do not hesitate to add a new file e.g. `testextra/doublecosets.tst` with this test. Otherwise, the name of the test file and the comment above suggest that this is a test for subgroup lattice routines, and the test does not seem to belong here.  If you think that it does, then at lease prepend it with a couple of lines of comments telling what it tests. This will make easier figuring out the purpose of this test in the future.

hulpke:
@alex-konovalov My understanding is that all requests of your hold have been satisfied. Have I overlooked anything?

lib/csetgrp.gi
```diff
@@ -749,6 +762,137 @@ local l;
...
+            #p:=[i..Length(orbs)-1];
+            #orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
+=======
+            #p:=[i..Length(orbs)-1];
+            #orbs{p}:=orbs{p+1};Unbind(orbs[Length(orbs)]);
+            #orbset{p}:=orbset{p+1};Unbind(orbset[Length(orbset)]);
+            #done{p}:=done{p+1};Unbind(done[Length(done)]);
+            #bahn{p}:=bahn{p+1};Unbind(bahn[Length(bahn)]);
+            #pam{p}:=pam{p+1};Unbind(pam[Length(pam)]);
+>>>>>>> 57ac0bd2e... Added `sisyphus` option to force extra work for testing
```
fingolfin: You have commited merge conflict markers here, which is why this PR fails in all tests

hulpke: Thank you -- Something must have gone wrong in the last rebase. Is (I hope) fixed now.

PR #2748, opened 2018-08-27
kernel: restrict T_ATOMIC to HPC-GAP

fingolfin:
We still parse (and ignore...) `atomic` expressions in regular GAP just fine. In fact, this PR is a result of looking at code coverage, as the executor for T_ATOMIC was dead code in non-HPC mode.

### DIFF
--- a/src/code.h
+++ b/src/code.h
@@ -176,7 +176,9 @@ enum STAT_TNUM {
             T_REPEAT2,
             T_REPEAT3,
 
+#ifdef HPCGAP
             T_ATOMIC,
+#endif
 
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Good catch, I have a feeling (I might be remembering, isn't important) I made parsing atomic a no-op to make atomic statements zero cost in GAP, and should have cleaned this up at the same time.

fingolfin:
Yes, that's roughly how it went down. But IIRC, when you did that, the atomic code was *much* more convoluted, making this much harder to spot. And we didn't have helpful coverage data.

PR #2750, opened 2018-08-27
kernel: catch some invalid uses of '~'; add tests

fingolfin:
Also, print a (more) helpful error message.

Before:

    gap> ~;
    Error, '~' does not have a value here
    gap> x -> ~;
    function( x ) ... end
    gap> x -> (1,~);
    function( x ) ... end
    gap> (1,~);
    Error, '~' does not have a value here
    not in any function at *stdin*:4

After:

    gap> ~;
    Syntax error: '~' not allowed here
    ~;
    ^
    gap> x -> ~;
    Syntax error: '~' not allowed here
    x -> ~;
         ^
    gap> x -> (1,~);
    Syntax error: '~' not allowed here
    x -> (1,~);
            ^
    gap> (1,~);
    Syntax error: '~' not allowed here
    (1,~);
       ^

### DIFF
--- a/src/compiler.c
+++ b/src/compiler.c
@@ -2502,7 +2502,7 @@ CVar CompListTildeExpr (
 
     /* remember the old value of '~'                                       */
     tilde = CVAR_TEMP( NewTemp( "tilde" ) );
-    Emit( "%c = STATE( Tilde );\n", tilde );
+    Emit( "%c = STATE(Tilde);\n", tilde );
 
     /* create the list value                                               */
     list = CompListExpr1( expr );
@@ -2718,19 +2718,19 @@ CVar CompRecTildeExpr (
 
     /* remember the old value of '~'                                       */
     tilde = CVAR_TEMP( NewTemp( "tilde" ) );
-    Emit( "%c = STATE( Tilde );\n", tilde );
+    Emit( "%c = STATE(Tilde);\n", tilde );
 
     /* create the record value                                             */
     rec = CompRecExpr1( expr );
 
     /* assign the record value to the variable '~'                         */
-    Emit( "STATE( Tilde ) = %c;\n", rec );
+    Emit( "STATE(Tilde) = %c;\n", rec );
 
     /* evaluate the subexpressions into the record value                   */
     CompRecExpr2( rec, expr );
 
     /* restore the old value of '~'                                        */
-    Emit( "STATE( Tilde ) = %c;\n", tilde );
+    Emit( "STATE(Tilde) = %c;\n", tilde );
     if ( IS_TEMP_CVAR( tilde ) )  FreeTemp( TEMP_CVAR( tilde ) );
 
     /* return the record value                                             */
--- a/src/error.c
+++ b/src/error.c
@@ -255,14 +255,12 @@ Obj FuncCALL_WITH_CATCH(Obj self, Obj func, volatile Obj args)
         SWITCH_TO_OLD_LVARS(currLVars);
         STATE(CurrStat) = currStat;
         SetRecursionDepth(recursionDepth);
-#ifdef HPCGAP
         STATE(Tilde) = tilde;
+#ifdef HPCGAP
         PopRegionLocks(lockSP);
         TLS(currentRegion) = savedRegion;
         if (TLS(CurrentHashLock))
             HashUnlock(TLS(CurrentHashLock));
-#else
-        STATE(Tilde) = tilde;
 #endif
     }
     else {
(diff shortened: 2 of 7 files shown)


fingolfin:
The more I think about it, the less unsure am I whether we should do it... This is related to the somewhat unclear definition of what `~` is *supposed* to be, as opposed to what the implementation right now does.

Perhaps allowing `f := x -> ~;` always was the intent, so that one can use it in `l := [ f(1) ];`. Also, what's the point in forbidding it, if one still create the same function via `f:=[x->~][1];`... 

The tilde operator is really useful in some cases, but also under documented (actually... is there *any* documentation for it? where?); and it very tightly interlaces the reader, interpreter, coder/executor and compiler. I tried to isolate `STATE(Tilde)` but it's hopeless with the current state of thing. The only way around that I can think of would be to forbid `~` inside a function from referencing a list/record outside the function. I created a quick patch to implement that, and this actually raised no errors, except in the Browse package, in `app/sudoku.g` around line 500:
```gap
# global record to hold most data and functions of this section
BindGlobal("Sudoku",  rec(
# we count fields of board rowwise from 1..81; first collect relevant
# indices in row, column and subsquare for each entry
rows := List([0..8], c-> c*9+[1..9]),
cols := List([1..9], i-> i+[0,9..72]),
squs := List([1,4,7,28,31,34,55,58,61], i-> i-1+[1,2,3,10,11,12,19,20,21]),
rcsind := List([1..81], i-> [QuoInt(i-1,9)+1, (i-1) mod 9 + 1,
          First([1..9], j-> i in ~.squs[j])]),
inds := List(~.rcsind, ind -> Set(Concatenation(~.rows[ind[1]],
              ~.cols[ind[2]], ~.squs[ind[3]])))
              ));
```

But this can be rewritten quite easily to avoid that. E.g.:
```gap
BindGlobal("Sudoku",  rec(
# we count fields of board rowwise from 1..81; first collect relevant
# indices in row, column and subsquare for each entry
rows := List([0..8], c-> c*9+[1..9]),
cols := List([1..9], i-> i+[0,9..72]),
squs := List([1,4,7,28,31,34,55,58,61], i-> i-1+[1,2,3,10,11,12,19,20,21]),
));
Sudoku.rcsind := List([1..81], i-> [QuoInt(i-1,9)+1, (i-1) mod 9 + 1,
          First([1..9], j-> i in Sudoku.squs[j])]);
Sudoku.inds := List(Sudoku.rcsind, ind -> Set(Concatenation(Sudoku.rows[ind[1]],
              Sudoku.cols[ind[2]], Sudoku.squs[ind[3]])));
```
or in many other ways. At worst, one can convert any code using ~ to code which doesn't via a trick:  change `CODE` to  `(tilde->CODE)(~)` and then replace ~ inside of CODE by `tilde`. This yields:
```gap
BindGlobal("Sudoku",  rec(
# we count fields of board rowwise from 1..81; first collect relevant
# indices in row, column and subsquare for each entry
rows := List([0..8], c-> c*9+[1..9]),
cols := List([1..9], i-> i+[0,9..72]),
squs := List([1,4,7,28,31,34,55,58,61], i-> i-1+[1,2,3,10,11,12,19,20,21]),
rcsind := (r->List([1..81], i-> [QuoInt(i-1,9)+1, (i-1) mod 9 + 1,
          First([1..9], j-> i in r.squs[j])]))(~),
inds := (r->List(r.rcsind, ind -> Set(Concatenation(r.rows[ind[1]],
              r.cols[ind[2]], r.squs[ind[3]]))))(~)
              ));
```

The only other example of code using this "feature" is in the lpres package:
```gap
    return rec(group := TrivialSubgroup(Q),
               quo := GroupHomomorphismByFunction(G,~.group,x->One(~.group)),
               set := X,
               top := Q,
               wreath := WreathProduct(~.group,Q),
               epi := GroupHomomorphismByImages(~.wreath,~.group,GeneratorsOfGroup(~.wreath),List(GeneratorsOfGroup(~.wreath),x->One(Q))));
```

It's clear how to change this code away from using `~`, too.

ChrisJefferson:
I'm happy with these fixes.

My feeling on tilde in general is that it is (as you say) horribly underspecified and underdocumented. However, I don't think it's worth the work to either (a) specify it better and handle every horrible corner case, or (b) remove it. As you showed, both in packages (and I've seen in code in the wild) some weird uses, which I'd prefer not to break unless we have to.

If at some point it arises some new feature / optimisation we want to add is blocked by Tilde, then I'm happy to re-evaluate the pain to make substansal changes, but for now I'd just leave it as a minor feature, applying (like this PR) minor fixes and improves where necessary, and trying not to break it.

PR #2751, opened 2018-08-27
kernel: move PrintPrecedence to module state

fingolfin:
This global was previously shared by all threads. That could in theory lead to
a bug, if two threads tried to print expressions with binary operations
simultaneously (say, into a string stream).

### DIFF
--- a/src/exprs.c
+++ b/src/exprs.c
@@ -1445,6 +1445,8 @@ void            PrintUnknownExpr (
 }
 
 
+struct ExprsState {
+
 /****************************************************************************
 **
 *V  PrintPrecedence  . . . . . . . . . . . . . . . . current precedence level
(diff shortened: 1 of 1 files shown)


src/exprs.c
```diff
@@ -1458,8 +1460,18 @@ void            PrintUnknownExpr (
...
 **  doesn't know that a function call adds automatically parenthesis.
 */
-UInt            PrintPrecedence;
+UInt PrintPrecedence;
+
+};
+
+static ModuleStateOffset ExprsStateOffset = -1;
 
+extern inline struct ExprsState * ExprsState(void)
```
markuspf: extern inline with function definition?

fingolfin: Yes. The extern ensures that the compiler emits an actual instance of the function; that's useful so that one can invoke it from a debugger.

markuspf: I see, thanks for the clarification.

PR #2753, opened 2018-08-28
Miscellaneous kernel changes

fingolfin:
This is a grab bag of small changes, for which it felt like a waste to open separate PRs.


fingolfin:
Ah OK, indeed, their status twitter account reports issues with background jobs which sounds related. They are working on processing the backlog.

Issue #2737, opened 2018-08-23
Merging #2640 broke Semigroups


markuspf:
Merging of #2640 broke Semigroups

```
gap> LoadPackage("semigroups");;
gap> Monoid(Matrix(GF(4), [[Z(4)]]));
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 2nd choice method found for `OneMutable' on 1 arguments at /home/makx/git/gap/master/lib/methsel2.g:250 called from
OneOp( elm ) at /home/makx/git/gap/master/lib/arith.gi:267 called from
One( gens ) at /home/makx/git/gap/master/pkg/semigroups/gap/semigroups/semigrp.gi:165 called from
<compiled statement>  called from
<compiled statement>  called from
SetGeneratorsOfMagmaWithOne( M, gens ); at /home/makx/git/gap/master/lib/monoid.gi:99 called from
...  at *stdin*:2
type 'quit;' to quit to outer loop
brk> 
```

(Semigroups has its own implementation of matrices, and now the `Matrix` methods from the GAP library take precedence over the semigroups ones, reworking semigroups is a bit of an undertaking).

alex-konovalov:
That's unfortunate. The latest release of Semigroups did not compile with master branch since mid-June (see https://github.com/gap-packages/Semigroups/issues/497) and while testing this PR there was no way to spot this. The fix for the compilation problem was merged in Semigroups repository on June 18th (https://github.com/gap-packages/Semigroups/pull/499) but no release followed, so who knew?

First, I am now setting up another Travis test which will use nightly build Docker container for master branch with packages-master.tar.gz archive to test development versions of packages. This will be limited to packages from gap-packages organisation as an additional benefit of hosting a package repository there.

Second, dependently on how complicated reworking the Semigroups is as @james-d-mitchell thinks, we can:
- revert #2640, branch off stable-4.10, and then merge #2640 into master again to appear in GAP 4.11, assuming that Semigroups will become compatible with it by GAP 4.11 (in ~6 month?), or
- continue as it is, assuming that Semigroups will be fixed in September. @james-d-mitchell what do you think?

alex-konovalov:
[...]

PR #2754, opened 2018-08-28
Downrank methods for `Matrix` and `BaseDomain` introduced in #2640 to unbreak `Semigroups` on `master`

markuspf:
This addresses #2737 in that `SemigroupsTestStandard()` passes now.

An alternative solution would just rank down the appropriate methods for `Matrix` and `BaseDomain` until a fix is integrated into Semigroups.

The fix for `Semigroups` will likely involve taking most custom matrix code out, which was the long-term plan all along according to @james-d-mitchell. For that it would obviously be good if the `MatrixObj` code was suitably stable (at least by the time releases happen).

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -179,6 +179,9 @@ InstallMethod( Matrix,
 #
 InstallMethod( Matrix,
   [IsSemiring, IsList, IsInt],
+  # FIXME: Remove this downranking, it was introduced to prevent
+  #        Semigroups from breaking ahead of the 4.10 release
+  -SUM_FLAGS,
   function( basedomain, list, nrCols )
     local rep;
     rep := DefaultMatrixRepForBaseDomain(basedomain);
@@ -187,6 +190,9 @@ InstallMethod( Matrix,
 
 InstallMethod( Matrix,
   [IsSemiring, IsList],
+  # FIXME: Remove this downranking, it was introduced to prevent
+  #        Semigroups from breaking ahead of the 4.10 release
+  -SUM_FLAGS,
   function( basedomain, list )
     local rep;
     if Length(list) = 0 then Error("list must be not empty"); fi;
@@ -196,6 +202,9 @@ InstallMethod( Matrix,
 
 InstallMethod( Matrix,
   [IsSemiring, IsMatrixObj],
+  # FIXME: Remove this downranking, it was introduced to prevent
+  #        Semigroups from breaking ahead of the 4.10 release
+  -SUM_FLAGS,
   function( basedomain, mat )
     # TODO: can we do better? encourage MatrixObj implementors to overload this?
     return NewMatrix( DefaultMatrixRepForBaseDomain(basedomain), basedomain, NrCols(mat), Unpack(mat) );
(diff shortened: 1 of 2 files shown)


markuspf:
Mhm. of course this also requires tests to be adapted. Maybe an optional down-ranking of the methods in question is a better approach?

markuspf:
I agree, let my try that...

ChrisJefferson:
One possible concern, the fact the methods still exist means if anyone starts using the MatrixObj code, their stuff will get "broken" if they load semigroups?

markuspf:
I am tempted to say "don't do that then" right now ;)

If they write new code and want matrix objects in a particular filter, they can still make them. The `Matrix` functions are convenience constructors, anyway.

james-d-mitchell:
I'm still tempted to just rename the methods in Semigroups, and then in time to remove them altogether. I will try to look at this this afternoon.

james-d-mitchell:
I approve of this, as a temporary fix, provided that all of the Semigroups package tests run (pending), and the GAP tests run when Semigroups is loaded.

What I'd like to understand better is what a "proper" fix would look like. As far as I can see, the only proper fix for this is where Semigroups uses matrix objects (as it currently does) is for MatrixObjects should be a proper object (as per the comments I made in Issue #2640). Another option is for Semigroups to stop using matrix objects altogether. I'd like to know which of these options I should plan for.

markuspf:
This breaks some tests of `MatrixObj` if `Semigroups` is loaded. Two options: 

 1) Comment out these tests
 2) Ignore

I am all for 2, `MatrixObj`s are undocumented code.

markuspf:
For completeness, with `Semigroups` loaded I get the following `teststandard` test failing:

```
########> Diff in /home/mp397/git/gap/tst/testinstall/MatrixObj/TraceMat.tst:6
# Input is:
m2 := Matrix(Integers,l);
# Expected output:
<2x2-matrix over Integers>
# But found:
Matrix(IsIntegerMatrix, [[1, 2], [3, 4]])
########
########> Diff in /home/mp397/git/gap/tst/testinstall/MatrixObj/TraceMat.tst:8
# Input is:
m3 := Matrix(GF(7),l*One(GF(7)));
# Expected output:
[ [ Z(7)^0, Z(7)^2 ], [ Z(7), Z(7)^4 ] ]
# But found:
Matrix(GF(7), [[Z(7)^0, Z(7)^2], [Z(7), Z(7)^4]])
########
########> Diff in /home/mp397/git/gap/tst/testinstall/MatrixObj/TraceMat.tst:12
# Input is:
TraceMat( m2 );
# Expected output:
5
# But found:
Error, no 1st choice method found for `TraceMat' on 1 arguments
########
########> Diff in /home/mp397/git/gap/tst/testinstall/MatrixObj/TraceMat.tst:14
# Input is:
TraceMat( m3 );
# Expected output:
Z(7)^5
# But found:
Error, no 1st choice method found for `TraceMat' on 1 arguments
```

i.e. two of them is are printing differences, and two are the non-existence of a `TraceMat` method.

PR #2755, opened 2018-08-28
kernel: unify 'and' for filters in interpreter and executor

fingolfin:
Also start using IS_FILTER. Note that previously, this input was silently accepted:

    gap> Center and IsAssociative;
    <Filter "<<and-filter>>">

This was effectively treated equivalently to `HasCenter and IsAssociative;`.
We now could reject such code. But unfortunately a few packages accidentally
relied on this bug. We thus only print a warning for now, and will turn this
into a proper error only once all affect packages released a fixed version.

This continues the work begun in PR #2732


tst/testinstall/boolean.tst
```diff
@@ -64,17 +74,64 @@ Error, <expr> must be 'true' or 'false' (not a integer)
...
+gap> function() return IsAssociative and Center; end();
+Warning, operation 'Centre' used in AND-filter is not a filter in stream:1
+<Filter "(IsAssociative and Centre)">
+gap> function() return IsAssociative and FirstOp; end();
+Error, <expr> must be a filter (not a function)
+gap> function() return true and IsAssociative; end();
+Error, <expr> must be 'true' or 'false' (not a function)
+gap> function() return Center and IsAssociative; end();
+Warning, operation 'Centre' used in AND-filter is not a filter in stream:1
+<Filter "(Centre and IsAssociative)">
```
fingolfin: Here the warning can be seen in effect

ChrisJefferson:
I wonder if we should not merge this until after 4.10 branches, to give packages a full release cycle to fix bugs, rather than have warnings in 4.10.1? Or is it important those bugs get fixed ASAP?

fingolfin:
No, it's not important at all. I just suggested it as a possibility in PR #2732, and @ChrisJefferson wrote:
> If 2 isn't too much work, it quite appeals, because it also means this doesn't get lost.

;-). But I am also fine with skipping this PR completely, and just merging PR #2756 (squashed down) once all affected packages made a fixed release.

fingolfin:
Hmm, just realized that there should be a third place that needs to be adjusted: the compiler. Gotta do that, including a test

PR #2759, opened 2018-08-28
Switch Travis to use Ubuntu 16.04 "xenial", and other CI tweaks

fingolfin:
This should allow us to build semigroups and digraphs again (alas, it seems digraphs still runs into these weird atexit troubles...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ env:
   global:
     - CFLAGS="--coverage -O2"
     - LDFLAGS="--coverage"
-    # default config flags: enable debug asserts, and treat compiler warnings as errors
-    - CONFIGFLAGS="--enable-debug --enable-Werror"
+    # default config flags: enable debug asserts
+    - CONFIGFLAGS="--enable-debug"
 
 addons:
   apt_packages:
@@ -22,7 +22,7 @@ matrix:
     # the same in 32 bit mode, and with debugging turned off, to (a) make it
     # a lot faster, and (b) to make sure nobody accidentally put some vital
     # computing into code that is only run when debugging is on.
-    - env: TEST_SUITES="docomp testtravis" ABI=32 CONFIGFLAGS="--enable-Werror"
+    - env: TEST_SUITES="docomp testtravis" ABI=32 CONFIGFLAGS=""
       addons:
         apt_packages:
           - gcc-multilib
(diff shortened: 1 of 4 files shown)


markuspf:
Not sure whether it makes sense; for `meataxe64` I am now also using `binutils-2.26`: (https://github.com/gap-packages/meataxe64/blob/master/.travis.yml)

I spent *quite* a bit of time trying to debug a linker problem with `meataxe64` which boiled down to "outdated binutils".

markuspf:
sorry, I merged #2702 which caused a conflict with this PR...

alex-konovalov:
gcc at least 5.0 would be very useful. The list of hacks to disable packages might now be extended with Digraphs (https://github.com/gap-system/gap/pull/2768). Hope with this PR all those hacks will be gone. But test of this PR currently fail to build packages:
```
checking whether GMP C++ library is available... configure: error: GMP C++ library not found. Make sure it was compiled with --enable-cxx
```

alex-konovalov:
@fingolfin did you manage to get at least the gcc-5 of this PR working? I'd really love to have it because otherwise Semigroups can not be compiled on Travis, which severely reduces ability to identify potential problems.

fingolfin:
This seems to be working now, at least with reduced tests. I've now cleaned up the PR, and in particular re-enabled all tests, so let's see if it still passes.

.travis.yml
```diff
@@ -1,14 +1,13 @@
...
+# Force Ubuntu 16.04 "Xenial" to get newer GCC, binutils etc.
+dist: xenial
 language: c
 
 env:
   global:
     - CFLAGS="--coverage -O2"
     - LDFLAGS="--coverage"
     # default config flags: enable debug asserts, and treat compiler warnings as errors
-    - CONFIGFLAGS="--enable-debug --enable-Werror"
```
alex-konovalov: If we do not treat compiler warnings as errors (but why?) then the comment should be updated too. But I'd prefer to treat them as errors for the core GAP system, but not for packages.

fingolfin: If you look at the commit messages, they explain that we actually keep using --enable-Werror, it simply is redundant to set this here again

fingolfin: but yeah, the comment should be updated

.travis.yml
```diff
@@ -58,7 +61,7 @@ matrix:
           - libcurl4-openssl-dev:i386 # for curlInterface
           - libboost-dev:i386       # for NormalizInterface
           - libmpfr-dev:i386        # for float
-          - libmpfi-dev:i386        # for float
+          - libmpfi0-dev:i386       # for float
```
markuspf:
```suggestion
          - libmpfi0-dev:i386        # for float
```

PR #2762, opened 2018-08-30
Add missing { } around macro

ChrisJefferson:
Hopefully obvious -- the TRY_IF_NO_ERROR is a macro, and when expanded doesn't bind "correctly". The { } just enforce what we always wanted, and are generally good practice around macros anyway.


ChrisJefferson:
thanks for the labels :)

PR #2766, opened 2018-08-31
Fix several bugs (including crashes) 

### LINKED ISSUE #2765: Crash in coder/executor with `foo!{[1,2,3]}` syntax
Consider this:
```
gap> f := function(l) l!{[1,3]} := [42, 23]; end;;
gap> Display(f);
Error, Panic: cannot print statement of type '55' in
  <corrupted statement>  called from
...
gap> f([]);
Panic: tried to execute a statement of unknown type '55'
```

With some more effort, it can even lead to a crash:
```
gap> function()
>   local l;
>   l:=[1,2,3];
>   Display(l);
>   l![1] := 42;
>   Display(l![1]);
>   Display(IsBound(l![1]));
>   Unbind(l![1]);
>   Display(l);
>   l!{[1,3]} := [42, 23];
>   Display(l);
>   Display(l!{[3,1]});
> end();
[ 1, 2, 3 ]
42
true
[ , 2, 3 ]
Panic: tried to execute a statement of unknown type '55'
[ , 2, 3 ]
Panic: tried to evaluate an expression of unknown type '189'
Segmentation fault: 11
```

What's going on here? Well, that's because support for list slices via the `list{indices}` syntax was *not* fully extended to its "bang-variant", `list!{indices}`, which is for accessing slices of positional objects.

In the interpreter, we thus just print an error: "sorry, not supported". But when coding, the coder happily codes this. But printing and executing was never implemented for these kinds of statements, hence the weird errors and ultimately the crash. Specifically, at least these six statement/expression types are affected:
```
T_ASSS_POSOBJ
T_ASS_POSOBJ_LEV
T_ASSS_POSOBJ_LEV
T_ELMS_POSOBJ
T_ELM_POSOBJ_LEV
T_ELMS_POSOBJ_LEV
```

Now, we could implement printing and coding for them, in a minimalistic fashion. But I wonder if it is worth it -- do we *really* want to support `!{...}`? As far as I can tell, it's not documented. So I would rather prefer to remove it completely -- most of the existing support code ends up printing "sorry, not yet implemented" anyway.

Or perhaps somebody can think of a reason to keep it?

fingolfin:
1. remove incomplete support for !{...} syntax, fixes crashes; fixes #2765 
2. reject invalid permutations in code and compiler, instead of producing permutation in a broken and thus potentially dangerous state. Example of the old, buggy behaviour:
```
gap> f:=x->(1,2,3,2);; pi:=f(0);
(1,2)
gap> 3^pi;
2
gap> ;
gap> MovedPoints(pi);
[ 1, 2, 3 ]
```
3. fix crash (at least with assertions on) when coding `IsBound(dvar)` (probably the most obscure of the three bugs)
4. add more tests for the interpreter and coder (creating these tests ultimately revealed the three bugs above)
5. remove some dead code

### DIFF
--- a/doc/ref/language.xml
+++ b/doc/ref/language.xml
@@ -1653,8 +1653,7 @@ It is an error to use this statement other than inside a loop.
 <P/>
 <Log><![CDATA[
 gap> break;
-Error, A break statement can only appear inside a loop
-not in any function
+Syntax error: 'break' statement not enclosed in a loop
 ]]></Log>
 
 </Section>
@@ -1688,8 +1687,7 @@ It is an error to use this statement other than inside a loop.
 <P/>
 <Log><![CDATA[
 gap> continue;
-Error, A continue statement can only appear inside a loop
-not in any function
+Syntax error: 'continue' statement not enclosed in a loop
 ]]></Log>
 
 </Section>
--- a/src/code.c
+++ b/src/code.c
@@ -2316,13 +2316,13 @@ void CodeAssList ( Int narg )
 {
     Stat                ass;            /* assignment, result              */
 
+    GAP_ASSERT(narg == 1 || narg == 2);
+
     /* allocate the assignment                                             */
     if (narg == 1)
       ass = NewStat( T_ASS_LIST, 3 * sizeof(Stat) );
-    else if (narg == 2)
+    else /* if (narg == 2) */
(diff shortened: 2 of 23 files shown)


alex-konovalov:
@fingolfin Travis currently fails - this should be fixed by #2768, let's get that PR a fast-track.

PR #2772, opened 2018-09-04
Profile interpreted code

ChrisJefferson:
I noticed after a recent PR of Max's, it is now easy to add profiling of interpreted code.

This PR does that. I have benchmarked carefully and can't measure any more overhead -- this isn't surprising as the interpreter runs very little code (any function or loop is "coded", and then run as bytecode).

I export KERNEL_API_VERSION, because in the profiling package and debugging package I need to know if this PR happened. In profiling it just breaks testing, in debugging it breaks the package but I will release an updated version.

It might be the case we get some strange coverage out, I'm partly going to see what this PR produces, and compare with previous GAP.

### DIFF
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -26,6 +26,7 @@
 #include "funcs.h"
 #include "gapstate.h"
 #include "gvars.h"
+#include "hookintrprtr.h"
 #include "integer.h"
 #include "io.h"
 #include "lists.h"
@@ -82,8 +83,32 @@
 */
 /* TL: UInt IntrCoding; */
 
+// INTERPRETER_PROFILE_HOOK deals with profiling of immediately executed
+// code.
+// If STATE(IntrCoding) is true, profiling is handled by the AST
+// generation and execution. Otherwise, we always mark the line as
+// read, and mark as executed if STATE(IntrReturning) and STATE(IntrIgnoring)
+// are both false.
+#define INTERPRETER_PROFILE_HOOK()                                           \
+    if (!STATE(IntrCoding)) {                                                \
+        InterpreterHook(GetInputFilenameID(), STATE(InterpreterStartLine),   \
+                        STATE(IntrReturning) || STATE(IntrIgnoring));        \
+    }                                                                        \
+    STATE(InterpreterStartLine) = 0;
 
-#define SKIP_IF_RETURNING() if ( STATE(IntrReturning) > 0 ) { return; }
+
+// Put the profiling hook into SKIP_IF_RETURNING, as this is run in
+// (nearly) every part of the interpreter, avoid lots of extra code.
+#define SKIP_IF_RETURNING()                                                  \
+    INTERPRETER_PROFILE_HOOK();                                              \
+    SKIP_IF_RETURNING_NO_PROFILE_HOOK();
+
+// Need to
+#define SKIP_IF_RETURNING_NO_PROFILE_HOOK()                                  \
+    if (STATE(IntrReturning) > 0) {                                          \
+        return;                                                              \
+    }
+
 #define SKIP_IF_IGNORING()  if ( STATE(IntrIgnoring)  > 0 ) { return; }
 
 
@@ -377,7 +402,7 @@ void            IntrFuncCallEnd (
     UInt                i;              /* loop variable                   */
 
     /* ignore or code                                                      */
-    SKIP_IF_RETURNING();
+    SKIP_IF_RETURNING_NO_PROFILE_HOOK();
     SKIP_IF_IGNORING();
     if ( STATE(IntrCoding)    > 0 ) {
       CodeFuncCallEnd( funccall, options, nr );
@@ -608,6 +633,9 @@ Int            IntrIfEndBody (
     UInt                nr )
 {
     UInt                i;              /* loop variable                   */
+
+    /* explicitly check interpreter hooks, as not using SKIP_IF_RETURNING  */
+    INTERPRETER_PROFILE_HOOK();
 
     /* ignore or code                                                      */
     if ( STATE(IntrReturning) > 0 ) { return 0; }
--- a/src/io.c
+++ b/src/io.c
@@ -54,14 +54,21 @@
 **  terminated by the character '\0'.  Because 'line' holds  only part of the
 **  line for very long lines the last character need not be a <newline>.
 **
+**  'stream' is non-zero if the input points to a stream.
+**
+**  'sline' contains the next line from the stream as GAP string.
+**
+**  The following variables are used to store the state of the interpreter
+**  and stream when another input file is opened:
+**
 **  'ptr' points to the current character within that line.  This is not used
 **  for the current input file, where 'In' points to the  current  character.
 **
 **  'number' is the number of the current line, is used in error messages.
 **
-**  'stream' is none zero if the input points to a stream.
+**  'interpreterStartLine' is the number of the line where the fragment of
+**  code currently being interpreted started. This is used for profiling
 **
-**  'sline' contains the next line from the stream as GAP string.
 **
 */
 typedef struct {
@@ -72,6 +79,7 @@ typedef struct {
     Char   line[32768];
     Char * ptr;
     UInt   symbol;
+    Int    interpreterStartLine;
     Int    number;
     Obj    stream;
     UInt   isstringstream;
@@ -537,6 +545,7 @@ UInt OpenInputStream(Obj stream, UInt echo)
         GAP_ASSERT(IS_CHAR_PUSHBACK_EMPTY());
         IO()->Input->ptr = STATE(In);
         IO()->Input->symbol = STATE(Symbol);
+        IO()->Input->interpreterStartLine = STATE(InterpreterStartLine);
     }
 
     /* enter the file identifier and the file name                         */
@@ -561,6 +570,7 @@ UInt OpenInputStream(Obj stream, UInt echo)
     STATE(In) = IO()->Input->line;
     STATE(In)[0] = STATE(In)[1] = '\0';
     STATE(Symbol) = S_ILLEGAL;
+    STATE(InterpreterStartLine) = 0;
     IO()->Input->number = 1;
 
     /* indicate success                                                    */
@@ -602,6 +612,7 @@ UInt CloseInput ( void )
     IO()->Input = IO()->InputStack[sp - 1];
     STATE(In) = IO()->Input->ptr;
     STATE(Symbol) = IO()->Input->symbol;
+    STATE(InterpreterStartLine) = IO()->Input->interpreterStartLine;
 
     /* indicate success                                                    */
     return 1;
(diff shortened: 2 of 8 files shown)


src/io.c
```diff
@@ -59,6 +59,9 @@
 **
 **  'number' is the number of the current line, is used in error messages.
 **
+**  'numberInterpreter' is the number of the line where the fragment of code
+**  currently being interpreted started, is used for profiling
```
fingolfin: change the comma to a semicolon or period? I was at first slightly confused by reading it like that.

fingolfin: Also, the corresponding GAPState member is named `InterpreterStartLine` -- why not name the two variables similarly? I find `InterpreterStartLine` much clearer than `numberInterpreter`.

src/profile.c
```diff
@@ -365,88 +365,113 @@ static inline Int8 getTicks(void)
...
+    CheckLeaveFunctionsAfterLongjmp();
 
-  // Explicitly skip these two cases, as they are often specially handled
-  // and also aren't really interesting statements (something else will
-  // be executed whenever they are).
-  if (TNUM_STAT(stat) == T_TRUE_EXPR || TNUM_STAT(stat) == T_FALSE_EXPR) {
-    return;
-  }
+    // Catch the case we arrive here and profiling is already disabled
+    if (!profileState_Active) {
```
fingolfin: Off-topic: `profileState_Active` is not used outside this file anymore, as far as I can tell, so it could be moved back into `struct ProfileState`, couldn't it? (I am not suggesting you should do anything about this in this PR, just wondering while I am reading through this code again).
Also, there is no real reason why `fopenMaybeCompressed` and `fcloseMaybeCompressed` take `struct ProfileState* ps` as argument, instead of just accessing the `profileState` global directly, is there?

src/profile.c
```diff
@@ -365,88 +365,113 @@ static inline Int8 getTicks(void)
...
-  }
+    nameid = getFilenameIdOfCurrentFunction();
+    outputFilenameIdIfRequired(nameid);
 
-  nameid = getFilenameIdOfCurrentFunction();
-  outputFilenameIdIfRequired(nameid);
+    // Statement not attached to a file
+    if (nameid == 0) {
+        return;
+    }
```
fingolfin: Wouldn't it make more sense to check `nameid==0` before calling `outputFilenameIdIfRequired`?

Ah, OK, I see it was like that before (and `outputFilenameIdIfRequired` does nothing if nameid is zero). Ah well,

fingolfin:
Ahh, I see, thanks for the explanation.

I agree that this should wait for GAP 4.11

Dually to this PR, what we are still lacking is a way to find unused .g / .gd / .gi files which have zero coverage because they are never read.

fingolfin:
BTW, while browsing Codecov PR pages is still broken, one can inspect the *commit* at codecov to at least be able to browse through files and get their coverage. I.e.: 
 <https://codecov.io/gh/gap-system/gap/tree/d06176adeefdc7210ce6e59e6182e625a945e6e0>

Then one can use good old eyeballs to visually "diff" this against master.

(Note that https://codecov.io/gh/gap-system/gap also shows outdated information -- I sent a bugreport to Codecov about that earlier. Apparently, they don't show coverage the HEAD commit of master there, but rather they use the commit date to pick the commit/report to show... And as it happens, the last two PRs we "merged" via rebasing had older commit dates than the one before...)

wilfwilson:
Hi @ChrisJefferson! This is cool. With the Semigroups and Digraphs packages, we use the profiling stuff in Travis to check the code coverage of our packages (we make sure that the coverage never falls below a certain level). When I test with this PR, a lot of the files have decreased coverage, to the point that Travis fails.

This seems to be solely due to more missed lines, including lines containing a single else, as you mention above, but also broken lines that include things like:

screen shot 2018-09-05 at 17 37 54

screen shot 2018-09-05 at 17 57 24

screen shot 2018-09-05 at 17 58 12

A temporary solution would be to reduce our required coverage for now, but it's not the nicest solution. 

I can give you more specific details if you'd like.

fingolfin:
@ChrisJefferson FYI, `IntrIfEndBody` does not use `SKIP_IF_RETURNING()` because that function has  a non-void return value. So a `INTERPRETER_PROFILE_HOOK();` may need to be inserted there. That's the only place like this, though.

src/scanner.c
```diff
@@ -195,6 +195,10 @@ void Match (
 {
     Char                errmsg [256];
 
+    if (STATE(InterpreterStartLine) == 0) {
+        STATE(InterpreterStartLine) = STATE(SymbolStartLine);
```
fingolfin: BTW, `InterpreterStartLine` might be incorrect if there is a recursive interpreter call (right at its start, resp. right after it ended). So perhaps `InterpreterStartLine` should be stored and restored along with Symbol and other interpreter state. Unfortunately, that whole business of storing interpreter state is something I have not yet had time to clean up and refactor (though it is on my implicit TODO); as a result, `Symbol` is (re)stored via `IO()->Input->symbol` in `io.c`; while `StackObj` is (re)stored via `PushPlist(STATE(IntrState), STATE(StackObj));` in `intrprtr.c`; and other bits and pieces are (re)stored via various code in `read.c`.

Anyway, that's hypothetical, I don't know whether this really manifests as a problem.

ChrisJefferson: I am saving/restoring in io.c, in the same place as Symbol is. I started thinking about trying to neaten this up, then got scared away, congratulations on your progress in cleaning up this area, it's very hairy :)

src/io.c
ChrisJefferson: I named it numberInterpreter to line up with number (the previous variable), but I agree that's a silly name for a variable, and I shouldn't be carrying on it's usage :)

src/profile.c
ChrisJefferson: The taking as an argument was beginning progress towards allowing multiple profilestates, so one could profile multiple threads in HPC-GAP, but it was never remotely finished. Should have a cleanup in this area, I agree.

ChrisJefferson:
Hi @wilfwilson , I believe I've cleared up those random elses, could you try again and see if things are improved? Thanks for the test case.

There is now only one problem I know about, which is (I believe) unfixable, but I will mention. Consider the following snippet, from algfp.gi:


line | code
-- | --
381 | InstallMethod( RelatorsOfFpAlgebra,
382 | "for a full f.p. algebra",
383 | true,
384 | [ IsSubalgebraFpAlgebra and IsFullFpAlgebra ], 0,
385 | A -> ElementsFamily( FamilyObj( A ) )!.relators );

In current master, line 385 is marked as "not executed", and everything else is not marked at all.

With this PR, all lines are marked as executed. That is because the closing bracket of line 385 is considered to "execute" the InstallMethod call, which makes line 385 marked as executed. Note that if that closing bracket is moved to a new line, then the function on line 385 is correctly marked as not executed.

Of course a future token-based profiler could handle these issues of multiple statements on a line.

wilfwilson:
Hi @ChrisJefferson thanks for your changes. I haven't looked at everything in super close detail, but where I have looked I haven't seen any weird behaviour, and moreover, this does not seem to decrease the Semigroups package code coverage in the way that the previous version of this PR did. So: I'm happy for this to be merged whenever.

fingolfin:
Actually, that last caveat, despite making perfect sense, also means that I now have serious reservations about this PR: There are *tons* of 1-line method installations like this in the kernel, and now we'd not be able anymore to tell if they get coverage or not...

But I am not quite willing to give up hope for that yet... I mean, sure, there is code like this, where we'll never be able to resolve this unless we stored profiling data token-based:
```gap
InstallMethod(NestingDepthM, [IsGF2MatrixRep], m->2);
```
But in the example we quote, couldn't we hack `IntrFuncCallEnd` to *not* emit coverage information? Granted, that might produce some lines which should get coverage but don't, but personally, I'd prefer that over lines which incorrectly are shown as covered. (I realize that this already happens, e.g. when people use multiple statements per line, but that's no reason to make it worse).

As to tokenization: I am not sure whether any profiling visualization tools support displaying data on a finer than per-line level. It sure would be cool if they did, though.

src/intrprtr.c
```diff
@@ -82,8 +83,21 @@
 */
 /* TL: UInt IntrCoding; */
 
+// INTERPRETER_PROFILE_HOOK makes sure code hooking the interpreter sees
+// code which is not being coded.
+#define INTERPRETER_PROFILE_HOOK()                                           \
+    if(!STATE(IntrCoding)) {                                                 \
```
fingolfin: nitpick: missing space after `if`

src/intrprtr.c
```diff
@@ -82,8 +83,21 @@
...
+      InterpreterHook(GetInputFilenameID(), STATE(InterpreterStartLine),     \
+                      STATE(IntrReturning) ||   STATE(IntrIgnoring));        \
```
fingolfin: nitpick: excessive whitespace after `||`

src/io.c
```diff
@@ -54,14 +54,21 @@
 **  terminated by the character '\0'.  Because 'line' holds  only part of the
 **  line for very long lines the last character need not be a <newline>.
 **
+**  'stream' is none zero if the input points to a stream.
```
fingolfin: none zero -> "nonzero" or "non-zero" ?

ChrisJefferson:
So, with your suggestion it's easy to fix:

```
InstallMethod(NestingDepthM, [IsGF2MatrixRep], 
m->2);
```

Which is a fairly common pattern. The problem if it's all on one line is of course you parse and "execute" lots, include a list construction, grabbing a couple of identifiers, that kind of thing.

src/modules.h
```diff
@@ -40,7 +40,7 @@
 **
 */
 
-#define GAP_KERNEL_MAJOR_VERSION 3
+#define GAP_KERNEL_MAJOR_VERSION 4
```
fingolfin: This needs to be incremented again, as the T_COPYING PR also increment this.

src/intrprtr.c
```diff
@@ -82,8 +83,32 @@
 */
 /* TL: UInt IntrCoding; */
 
+// INTERPRETER_PROFILE_HOOK deals with profiling of immediately executed
+// code.
+// If STATE(IntrCoding) is true, profiling is handled by the bytecode
+// generation and interpreter, so skip. Otherwise, we mark the line as
```
fingolfin: It is misleading to talk about "bytecode" here, GAP doesn't generate bytecode but rather an (A)ST.

ChrisJefferson: changed

src/io.c
ChrisJefferson: fixed

ChrisJefferson:
I think this is now clean. I am in two minds about the whole "sneak something extra into `SKIP_IF_IGNORING`, I could go through and add an explicit `PROFILE_HOOK()` (or better name) to every occurrence of `SKIP_IF_IGNORING`, which wouldn't be much work but would make a 122-line patch.

PR #2773, opened 2018-09-04
Reorder methods after new implications are added (rebased PR #2521)

### LINKED ISSUE #2513: Method order does not change when filter ranks change, leading to subtle conflict between nq and lpres

fingolfin:
This is rebased and cleaned up version of #2521; I removed many whitespace-only changes, reformatted a few things and otherwise squashed most changes into a single commit.

Closes #2521 
Fixes #2513


lib/filter.g
```diff
@@ -184,6 +184,64 @@ BIND_GLOBAL( "InstallTrueMethodNewFilter", function ( tofilt, from )
...
+##  <Description>
+##  These functions control whether the method reordering process described
+##  in <Ref Func="InstallTrueMethod"/> is invoked or not. Since this process
+##  can be comparatively time-consuming, it is usually suspended when a lot
+##  of implications are due to be installed, for instance when loading the
+##  library, or a package. This is done by calling
+##  <C>SuspendMethodReordering()</C> once the installations are done,
+##  <C>ResumeMethodReordering()</C> should be called. These pairs of calls
+##  can be nested. When the outermost pair is complete, method reordering
+##  takes place and is enabled in <C>InstallTrueMethod</C> thereafter.
```
ThomasBreuer: Better write `<Ref Func="InstallTrueMethod"/>`.

lib/oper1.g
```diff
@@ -290,9 +298,11 @@ end );
 ##  if supplied <A>info</A> should be a short but informative string
 ##  that describes for what situation the method is installed,
 ##  <A>famp</A> should be a function to be applied to the families
-##  of the arguments,
-##  and <A>val</A> should be an integer that measures the priority
-##  of the method.
+##  of the arguments.
+##  a<A>val</A> should be an integer that measures the priority
```
ThomasBreuer: typo

lib/filter.g
fingolfin: Agreed; and the same for `SuspendMethodReordering` etc. -- all changed now.

lib/oper1.g
```diff
@@ -290,9 +298,11 @@ end );
...
 ##  that describes for what situation the method is installed,
 ##  <A>famp</A> should be a function to be applied to the families
-##  of the arguments,
-##  and <A>val</A> should be an integer that measures the priority
-##  of the method.
+##  of the arguments.
+##  <A>val</A> should be an integer that measures the priority
+##  of the method, or a function of no arguments which should return such an
+##  integer and will be called each time method order is being recalculated,
+##  (see <Ref Func="InstallTrueMethod"/>).
```
ThomasBreuer: either remove the brackets or the coma

fingolfin:
No problem, thanks for carefully reading! I removed the comma.

In addition, I added a commit which converts a bunch of `InstallMethod` calls in the library, which computed some `RankFilter` values in order to compute a single rank adjustment value, to instead use a rank adjustment *function*, which (thanks to the first commit) means the rank can be
correctly recomputed as necessary.

A similar thing likely should be done in pacakges...

Now it would be really good if @stevelinton could comment how he'd like to proceed...

stevelinton:
I'm happy with this. The older PR kept getting test failures which, I think, were actually not caused by the PR but by test infrastructure glitches. Otherwise I'd have merged it.

stevelinton:
This has passed all tests and been extensively reviewed. Time to merge.

PR #2774, opened 2018-09-04
Add etc/cov.sh local code coverage script

mtorpey:
I've written a short bash script that quickly and easily runs code coverage on a package, using GAP's `--cover` option and the profiling package.  I wrote it for my own use in the `MathInTheMiddle` package, but I realised it should work in any package containing a `tst/testall.g` file, so I thought I'd pop it in a PR in case it might be useful for anyone else.

Does this look like the sort of thing that might be useful?

### DIFF
--- a/etc/cov.sh
+++ b/etc/cov.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+#
+# cov.sh - Test code coverage in a package and show the result using HTML
+#
+# This script should be called from the root directory of a GAP package.  It
+# runs the file 'tst/testall.g' and puts an HTML code coverage report in the
+# tmp/ directory, reporting where the output can be found.
+#
+#   -b option: open output in browser when done.
+#
+
+set -e
+
+# Name the directory
+COVDIR=$PWD/tmp
+
+# Get a GAP command
+if [ "x$GAP" = x ] ; then
+    GAP="gap"
+fi
+command -v $GAP >/dev/null 2>&1 ||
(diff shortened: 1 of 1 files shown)


fingolfin:
First off, this indeed sounds useful. But, open question: is putting it into GAP the right place? We don't put e.g. the release tools into GAP either.

markuspf:
We thought about this today, too; Another canonical place could be the `profiling` package, arguably it needs the profiling package anyway. We could also describe some nice development cycle there (including running tests and coverage with low overhead, and mentioning that the coverage shown by this is the same as coverage shown in `codecov` (anything else would be a bug).

markuspf:
I would not want to put this into `PackageMaker`, because this would usually mean some copy of this script would end up in every newly created package, and then bitrot there.

A similar problem I presume for `Example`.

If we want to put it somewhere which isn't GAP's git repository (it *is* part of the testing framework which is part of GAP, but as mentioned above, it needs the `profiling` package to work). By this logic I still think the profiling package might be the best place to put it.

alex-konovalov:
I see your point. @ChrisJefferson ?

ChrisJefferson:
I would be happy for this to be in profiling. Then, it would probably be better if it was a GAP function, inside the profiling package, something like `ProfilePackage`, or `ProfileFile` (or both!)

mtorpey:
I've spoken with @ChrisJefferson for a bit about how this would work, and I'm now rewriting this to go into the profiling package.  The consensus seems to be to put nothing in the library, so I'll close this PR.  Thanks for the feedback! :)

PR #2778, opened 2018-09-05
hpcgap kernel: add enums for lock modes/status/qualifiers

fingolfin:
Various magic constants (0, 1, 2) were used to indicate the status of locks;
to indicate the desired mode for a lock; and to indicate the presence of
readonly/readwrite keywords.

The use of these magic constants made it rather difficult to correctly
understand some of the code. By introducing bespoke names for them, the code
flow is much easier to understand.

This commit also fixes a possible OOB access to the LCKS_FUNC of an atomic
function: we only added entries to it for arguments with a qualifier; but
code later on assumes that LCKS_FUNC has entries for every argument.

Also removed some bogus code in FuncLOCK FuncTRYLOCK which apparently was
meant to accept odd/even integer values instead of true/false (which is
undocumented); but that code used `&& 1` instead of `& 1`, so always produced
readwrite mode.

### DIFF
--- a/doc/hpc/regions.xml
+++ b/doc/hpc/regions.xml
@@ -746,6 +746,7 @@ It is generally safe to use if all threads but the current one are paused.
     <Subsection Label="The atomic statement.">
       <Heading>The <C>atomic</C> statement.</Heading>
 
+<Index Key="atomic" Subkey="no value"><K>atomic</K></Index>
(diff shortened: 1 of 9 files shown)


fingolfin:
Yes, this also was a PITA for me, which is precisely why I created this PR. Now changing the values of course is trivial, but also somewhat unnecessary, as we now have those enums.

PR #2779, opened 2018-09-06
fix and document `CharacterValueWreathSymmetric`

ThomasBreuer:
The undocumented function `CharValueWreathSymmetric` was not correct.
It was not used anywhere in the library,
but perhaps somebody else has used it.

I have fixed the bug and documented the function
(including a cross-reference from `PartitionTuples`).
For that, I have renamed it to `CharacterValueWreathSymmetric`,
and added the old name as an obsolete synonym.

It would be good to add a cross-reference to the function `BetaSet`.
This function belongs to the GAP library and *is* documented,
but the documentation belongs to the package `hecke`.
I think the right solution for this problem is to ask @dtraytel
to remove the documentation of `BetaSet` from his package;
afterwards we can then add documentation to the GAP Reference Manual.

(Besides that, I have corrected some misleading comments about the use
of obsolete variable names.)

### DIFF
--- a/lib/obsolete.gd
+++ b/lib/obsolete.gd
@@ -265,6 +265,19 @@ DeclareObsoleteSynonymAttr( "NormedVectors", "NormedRowVectors" );
 
 #############################################################################
 ##
+#F  CharValueWreathSymmetric( <tbl>, <n>, <beta>, <pi> )  . .
+#F                                        . . . .  character value in G wr Sn
+##
+##  This function was never documented but had been available for decades.
+##  Its functionality became documented under the more suitable name
+##  'CharacterValueWreathSymmetric' in GAP 4.10.
+##
+DeclareObsoleteSynonym( "CharValueWreathSymmetric",
+    "CharacterValueWreathSymmetric" );
+
+
+#############################################################################
+##
 #O  FormattedString( <obj>, <nr> )  . . formatted string repres. of an object
 ##
 ##  This variable name was never documented and is obsolete.
--- a/doc/manualbib.xml
+++ b/doc/manualbib.xml
@@ -2878,6 +2878,25 @@
   <other type="fjournal">Journal of Algebra</other>
 </article></entry>
 
+<entry id="JK81"><book>
+  <author>
+    <name><first>Gordon</first><last>James</last></name>
+    <name><first>Adalbert</first><last>Kerber</last></name>
+  </author>
+  <title>The representation theory of the symmetric group</title>
+  <publisher>Addison-Wesley Publishing Co., Reading, Mass.</publisher>
+  <year>1981</year>
+  <volume>16</volume>
+  <series>Encyclopedia of Mathematics and its Applications</series>
+  <note>With a foreword by P. M. Cohn,
+              With an introduction by Gilbert de B. Robinson</note>
+  <isbn>0-201-13515-9</isbn>
+  <mrnumber>644144</mrnumber>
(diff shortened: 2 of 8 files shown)


fingolfin:
Open question: Why should we wait for an update to the `hecke` package before, with adding `BetaSet` documentation? At worst, it'd be documented in two places for a time. I can't really see an issue with that, though?

lib/obsolete.gd
```diff
@@ -171,12 +171,18 @@ BindGlobal( "DeclarePackageAutoDocumentation", Ignore );
 # 03/2018: not used in any redistributed package
 
 DeclareObsoleteSynonym( "ReadPkg", "ReadPackage" );
-# 03/2018: still used in automgrp, ctbllib, fplsa, fwtree, Hap (HapCocyclic),
+# 03/2018: still used in automgrp, fplsa, fwtree, Hap (HapCocyclic),
 # itc, modisom, pargap, quagroup
+# (Note that 'grep' finds 'ReadPkg' also in 'ctbllib/init.g',
+# but that line of code is reached only if the file is read in GAP 3.
+# This occurrence of 'ReadPkg' will remain forever.)
```
fingolfin: Coincidentally, I've been preparing a PR which updates obsolete.gd, and which also addresses this bit about ctbllib (but also much more).

Could I convinced you to drop this here? Then we don't have two conflicting PRs.

fingolfin:
Regarding hecke, perhaps submitting an issue reminding about this to https://github.com/gap-packages/hecke/issues would be best; there are a few other smallish changes in that repository, perhaps @dtraytel could be convinced to either make a 1.4.1 or 1.5 release from there himself, or allow us ("The GAP Team") to take on co-maintainership of his package, as so many other people have done; then we can take care of the fix and the release ourselves.

dtraytel:
The latter would be definitely my preferred option. As I wrote in an earlier email to Alex Konovalov:

> I would like to hand over the maintenance of Hecke to someone who is more competent in today's GAP (or the topics of Hecke). I think Alex is in the best position to try to find such a volunteer.

What do I have to do to give "The GAP Team" co-maintainership?

fingolfin:
@dtraytel Great! Since it is already in github/gap-packages, you essentially have to do nothing, other than give us permission (which you did). We can take care of the rest

dtraytel:
Perfect! :)

fingolfin:
@ThomasBreuer this has conflicts now, because it still modifies `lib/obsolete.gd` (it adds a change and then reverts it in a separate commit). To fix that, you need to squash your second commit into the first. E.g.: run `git rebase -i master`, and then change the `pick` in the second line to `s`, then save&exit (and when prompted to edit the commit message, perhaps remove the references to "obsoletes", which are now, well, obsolete ;-).

Thanks!

ThomasBreuer:
@fingolfin 
Thank you for the instructions.
However, I do not understand why the variant with two commits is technically different from the squashed variant.
Logically, I found the setup with two commits better.
(If a fix is needed for such situations then would this perhaps be a fix in git?)

fingolfin:
I don't understand what you mean with this being "logically better": right now, there are two commits; the second only exists to undo something the previous commit does. That adds pointless working when one has to review past commits (e.g. as part of a `git bisect`).

In any case, I literally cannot merge this unless you rebase it.

fingolfin:
BTW, an update version of Hecke has been released, with the `BetaSet` documentation removed

Issue #1981, opened 2017-12-04
Add `--quit-on-syntax-error` option?


fingolfin:
For our CI tests, it would be helpful if we could reliably and globally detect syntax errors -- these often are not detected by tests, yet can contain crucial hints about issues.

To this end, having an options which turns every kernel call to `SyntaxError` (and possibly also `SyntaxWarning`) into a hard error with non-zero exit code would be useful.

markuspf:
Max Horn writes:

> For our CI tests, it would be helpful if we could reliably and globally detect syntax errors -- these often are not detected by tests, yet can contain crucial hints about issues.
>
> To this end, having an options which turns every kernel call to `SyntaxError` (and possibly also `SyntaxWarning`) into a hard error with non-zero exit code would be useful.


I would like to be able to catch SyntaxError and SyntaxWarning for
Jupyter (which will probably necessitate some "callback" mechanism?

ChrisJefferson:
Seperate from @markuspf wanting a callback, we could extend/rename `--quit-on-break` to also cover syntax errors.

alex-konovalov:
There is `--quitonbreak` which does this, but no `--quitonwarning`. The latter would also be useful, e.g. for #2782.

PR #2782, opened 2018-09-06
Travis: start GAP with disabled obsoletes in one of the tests

alex-konovalov:
This is an experiment to see how much Travis CI tests will fail. To start with, GAP is called with `-O` everywhere. Most likely, this PR will evolve into disabling loading obsoletes only in some tests - we do want to test obsolete parts of the code too, since they are loaded by default.

### DIFF
--- a/etc/ci.sh
+++ b/etc/ci.sh
@@ -131,7 +131,11 @@ GAPInput
     ;;
 
   testmanuals)
-    $GAP $SRCDIR/tst/extractmanuals.g
+    # Start GAP with -O option to disable obsoletes. The test
+    # will fail if there will be an error message, but warnings
+    # should be checked manually in the test log. Since some
+    # packages may still use obsoletes, we use -A option.
+    $GAP -O -A $SRCDIR/tst/extractmanuals.g
 
     $GAP <<GAPInput
         SetUserPreference("ReproducibleBehaviour", true);


alex-konovalov:
`gap -O` fails to load in `testinstall` because  of TomLib:
```
Error, Variable: 'Revision' must have a value
not in any function at /Users/alexk/GITREPS/gap/pkg/tomlib/gap/tmadmin.tmd:12
Error, was not in any namespace at /Users/alexk/GITREPS/gap/lib/variable.g:291 called from
LEAVE_NAMESPACE(  ); at /Users/alexk/GITREPS/gap/lib/package.gi:1234 called from
<function "ReadPackage">( <arguments> )
 called from read-eval loop at /Users/alexk/GITREPS/gap/pkg/tomlib/init.g:11
you can 'return;' to continue
brk> 
```
With `gap -A -O` GAP loads and `tst/testinstall.g` runs. Maybe at least `-O` in such test will be useful to prevent obsoletes reappearing in the GAP library.

alex-konovalov:
Rebased - there are three tests that pass, maybe use `-O` only for those tests - this will prevent obsoletes from reappearing in the library.

fingolfin:
@alex-konovalov I wonder what the results are now that there was a new tomlib release without `Revision`?

alex-konovalov:
There are some obsoletes still used by some packages loaded by default. This prevents from using `-O` more widely. To make at least some use of it, I suggest to use `-O` in at least one test where GAP is started to extract examples from GAP manuals for future testing.

This will fail the `testmanuals` test in case of an error. If however, using obsolete will only trigger a warning, the test will not fail. It will be useful to have `--quitonwarning` option in addition to `--quitonbreak`, see #1981.

alex-konovalov:
Also, I don't quite like the idea of hiding test for obsoletes into the test for manual examples. If the test fails because of obsoletes, it will not be immediately obvious for the submitter what's the reason. OTOH, having a separate test ONLY to build and start GAP is an overhead. Maybe rename Travis test instead? I don't have a preference at the moment.

alex-konovalov:
One could of course also start GAP with `-O` when testing manual examples. 

This will be useful since it will ensure that obsolete names do not reappear not only in the GAP library, but also in the manual examples. 

However, in GAP 4.10.2 installation there are four warnings caused by this: 
```
Syntax warning: Unbound global variable in /usr/local/Cellar/gap/4.10.2/libexe\
c/pkg/atlasrep/gap/access.gi:2568
        firstarg:= USER_HOME_EXPAND( firstarg );
                                   ^
Syntax warning: Unbound global variable in /usr/local/Cellar/gap/4.10.2/libexe\
c/pkg/Browse/app/filetree.g:21
    files:= STRING_LIST_DIR( USER_HOME_EXPAND( startfile ) );
                                             ^
Syntax warning: Unbound global variable in /usr/local/Cellar/gap/4.10.2/libexe\
c/pkg/ctbllib/gap4/ctadmin.tbi:114
      name:= USER_HOME_EXPAND( name );
                             ^
Syntax warning: Unbound global variable in /usr/local/Cellar/gap/4.10.2/libexe\
c/pkg/ctbllib/gap4/ctdbattr.g:2484
      NrMovedPoints, [ 2 .. TRANSDEGREES ],
                                         ^
```
So we first need to get new releases of these packages not using obsoletes, provided by @ThomasBreuer and @frankluebeck

alex-konovalov:
Perhaps we can merge this. Removed "do not merge", rebased, and edited the PR title.

etc/ci.sh
```diff
@@ -131,7 +131,7 @@ GAPInput
     ;;
 
   testmanuals)
-    $GAP $SRCDIR/tst/extractmanuals.g
+    $GAP -O -A $SRCDIR/tst/extractmanuals.g
```
wilfwilson: Could you maybe add a comment above this line, saying that the purpose of the `-O` is to test that GAP loads with obsoletes disabled, and `-A` is used is because some auto-loaded packages still use obsoletes.

That way it is clear why these options are being used here, and it's clear that it's not arbitrary/accidental. It is useful for someone wondering when/whether `-A` can be removed.

alex-konovalov: Done, good suggestion @wilfwilson !

PR #2785, opened 2018-09-06
Add more simple tests

ChrisJefferson:
Some simple tests, in particular making sure we hit all lines where we InstallMethod in a single line, as they may (shortly) not be profiled correctly.

### DIFF
--- a/tst/testinstall/kernel/object.tst
+++ b/tst/testinstall/kernel/object.tst
@@ -1,0 +1,41 @@
+#
+# Tests for objectify
+#
+gap> START_TEST("object.tst");
+
+# test some standard object types
+gap> r := Objectify(TYPE_KERNEL_OBJECT, rec());
+<kernel object>
+gap> KnownAttributesOfObject(r);
+[  ]
+gap> KnownAttributesOfObject((1,2));
+[  ]
+gap> SortedList(KnownPropertiesOfObject((1,2)));
+[ "CanEasilyCompareElements", "CanEasilySortElements" ]
+gap> KnownAttributesOfObject([]);
+[ "LENGTH" ]
+gap> SortedList(KnownPropertiesOfObject([]));
+[ "IS_SSORT_LIST", "IsDuplicateFree", "IsEmpty", "IsFinite", "IsNonTrivial", 
+  "IsSmallList", "IsSortedList", "IsTrivial" ]
+gap> SortedList(KnownTruePropertiesOfObject([]));
+[ "IsDuplicateFree", "IsEmpty", "IsFinite", "IsNonTrivial", "IsSSortedList", 
+  "IsSmallList", "IsSortedList" ]
+gap> KnownAttributesOfObject([3,2]);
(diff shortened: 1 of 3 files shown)


markuspf:
It';s clearly comparing against https://codecov.io/gh/gap-system/gap/branch/master/commits?from=2018-08-30%2000%3A00%3A00&to=2018-09-02%2000%3A00%3A00 not the "real" coverage of `master`.

tst/testinstall/kernel/object.tst
```diff
@@ -0,0 +1,62 @@
...
+[ "LENGTH" ]
+gap> KnownPropertiesOfObject([3,2]);
+[ "IsFinite", "IsSmallList" ]
+gap> KnownTruePropertiesOfObject([3,2]);
+[ "IsFinite", "IsSmallList" ]
+gap> KnownAttributesOfObject(Group((1,2,3)));
+[ "GeneratorsOfMagmaWithInverses", "MultiplicativeNeutralElement" ]
+
+# Don't want to hard wire this list as it is too prone to change
+gap> p := KnownPropertiesOfObject(Group((1,2,3)));
```
fingolfin: Well... that also means this test is prone to fail, which could become quite annoying over time.

At the very least, I'd change this to print the list after sorting, and perhaps with one string per line; that way, diffs are smaller and easier to understand. I.e.:
```
gap> p := KnownPropertiesOfObject(Group((1,2,3)));;
gap> Perform(Set(p), Display);
CanEasilyCompareElements
CanEasilySortElements
IsAssociative
IsCommutative
IsCompletelyRegularSemigroup
IsCompletelySimpleSemigroup
IsCyclic
IsDuplicateFree
IsEmpty
IsFinite
IsFinitelyGeneratedGroup
IsFreeAbelian
IsGeneratorsOfMagmaWithInverses
IsGeneratorsOfSemigroup
IsGroupAsSemigroup
IsInfiniteAbelianizationGroup
IsInverseSemigroup
IsMonoidAsSemigroup
IsMonomialGroup
IsNilpotentByFinite
IsNilpotentGroup
IsNonTrivial
IsOrthodoxSemigroup
IsPerfectGroup
IsPolycyclicGroup
IsRegularSemigroup
IsSimpleSemigroup
IsSolvableGroup
IsSubsetLocallyFiniteGroup
IsSupersolvableGroup
IsTorsionFree
IsTrivial
KnowsHowToDecompose
```

But I also worry that this test will be affected heavily by whether packages are loaded or now... Do we *really* need to print this list?

ChrisJefferson: Sorry, I meant to double semi-colon both of those, but forgot and didn't check my patch carefully before submitting. The plan is the next two tests check some properties (and true properties) which I trust will hang around and not change.

ChrisJefferson:
I have tweaked the tests to (hopefully) be much more resiliant to changes. Sorry, I did a last minute rewriteToFile, and didn't carefully check the output, which was very bad.

PR #2792, opened 2018-09-07
More kernel tests (and some minor kernel cleanup)

fingolfin:


fingolfin:
Sure, I can add such a comment. I am also happy to (in addition) rename them, if you have any good suggestions for a better prefix?  Perhaps `_TEST` or `INTERNAL_TEST` or... ?

PR #2797, opened 2018-09-10
Shorten testinstall/domain.tst to accommodate the Semigroups package

wilfwilson:
The last check in `tst/testinstall.tst` (testing the equality of an infinite semigroup and an infinite ideal) breaks when Semigroups 3.0.17 is loaded, which is currently the latest released version. See https://github.com/gap-packages/Semigroups/issues/508 for more discussion of this problem.

This PR comments out this line. Therefore we can branch `stable-4.10` at such a point that Semigroups does not break the GAP tests, and vice versa.

This test should be reinstated, or adjusted, at a point when it is compatible with the Semigroups package.

### DIFF
--- a/tst/testinstall/domain.tst
+++ b/tst/testinstall/domain.tst
@@ -18,8 +18,10 @@ gap> M = I;
 false
 gap> I = J;
 false
-gap> M = J;
-Error, no method found for comparing two infinite domains
+
+# TODO: Reinstate a version that is compatible with the Semigroups package
+#gap> M = J;
+#Error, no method found for comparing two infinite domains
 
 #
 gap> STOP_TEST("domain.tst");


wilfwilson:
Semigroups https://github.com/gap-packages/Semigroups/pull/513 has been merged, so the problem with the Semigroups package will be resolved when its next version is released. Note that the next released version of Semigroups is (currently) also required if GAP's `tst/MatrixObj/TraceMat.tst` is to pass with the Semigroups package loaded (due to issues with `TraceMat`).

alex-konovalov:
@wilfwilson do you know how long it is till the next Semigroups release will be ready?

wilfwilson:
I do not know, @james-d-mitchell is in charge of releases. Moreover, see https://github.com/gap-packages/Semigroups/issues/512; I imagine that a release will not be made until that issue is resolved (and, if agreed, the change implemented).

fingolfin:
Testes passed, but somehow the notification from Travis to GitHub about this didn't get through (and I already restarted one of the Travis jobs twice now in order to fix it, which usually does the trick -- but this time to no avail). So I'll just "force merge" it, (ab)using my admin powers...

PR #2798, opened 2018-09-11
Clarify MemoizePosIntFunction documentation, add a test

fingolfin:
Also clarify the documentation, and add a test case for when the
invalidInputHandler returns without a value

This contains part of the (now closed) PR #2600.

### DIFF
--- a/lib/cache.gd
+++ b/lib/cache.gd
@@ -17,18 +17,19 @@
 ##  <Func Name="MemoizePosIntFunction" Arg='function [,options]'/>
 ##
 ##  <Description>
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Just a general comment, as we have branched stable-4.10, this PR (and all other PRs) aren't going to get merged into 4.10 unless a label is added. While this PR is good, I don't think it needs to be in 4.10.

fingolfin:
I fully agree. I'd only consider PRs which fix issues for backporting; right now, I think only PR #2779 would qualify, but that one is not mergeable right now, @ThomasBreuer would need to rebase it first.

PR #2799, opened 2018-09-11
kernel: refactor and cleanup parts of stringobj.{c,h}

fingolfin:


src/stringobj.h
```diff
@@ -336,15 +338,16 @@ void MakeImmutableString(Obj str);
 
 static inline Obj MakeString(const Char * cstr)
```
stevelinton: What's the rationale for this change? The original was shorter and not obviously less efficient.

fingolfin: Your comment is on line 348/350, which actually was not changed by this PR; so I am not quite sure which change you are talking about?

ChrisJefferson: I would guess the C_NEW_STRING below? Where I agree with the change because I'd like to eliminate C_NEW_STRING, as it's a bit too "magic" (it couldn't be implemented as a C function as it effectively invisibly takes a pointer to it's first argument)

fingolfin: Ahhh, so the code below! My bad, I should have realized that.

Indeed, it is as @ChrisJefferson said: I am trying to get rid of `C_NEW_STRING`

PR #2802, opened 2018-09-11
lib/algsc.gi: merge HPC-GAP variant; add tests; fix an error message

fingolfin:


fingolfin:
I am also not 100% certain I did not mess it up, or that it wasn't broken before -- but at least it *seems* plausibel the way it is now, and we can only find out by testing this (and it wasn't tested that thoroughly before in HPC-GAP; anyway, with ward mostly non-functional, and us not really performing multi threaded tests right now, we'll have to wait for bug reports to find out anyway :/)

fingolfin:
Indeed, that test is not right -- but then why did it pass on Travis?!? Will investigate

PR #2803, opened 2018-09-12
Tweak some error messages

fingolfin:
This resolves the observation by @ThomasBreuer on PR #2802, where a test did not match a change to a certain error message.

The problem was that the same error message code was used in two other places, and my test did not actually test the one I changed. Now, the test is triggering one of the (in total, together with PR #2802) three changed error messages. I have not yet tried to write tests that trigger the other two.


fingolfin:
I think removing the code duplication would indeed be the better alternative on the long run; but after a quick look, it seems not completely trivial to me to unify the two methods, and I'd rather not spend time on this while risking to break things..

PR #2806, opened 2018-09-13
kernel: more tests for ariths.c; fix typo Ainv -> AInv

fingolfin:


tst/testinstall/kernel/ariths.tst
```diff
@@ -7,6 +7,38 @@ gap> START_TEST("kernel/ariths.tst");
...
+0
+gap> PROD(1,1);
+1
+gap> QUO(1,1);
+1
+gap> MOD(1,1);
+0
+
+#
+# suppress paths in trace output
```
stevelinton: Should this better be done generally for all tests? Reporting all  the errors "at stream:1" is clearly not useful.

fingolfin: Perhaps...

Actually, the `stream:1` is a good hint that something was defined on a test input, vs. defined in the library, so it's not entierely useless. Come to think about it, now I wonder if we could perhaps assign a "name" to streams, and then change the code to print the stream's name (if set) instead of just `stream`; then we could set the name to the name of the .tst file...

But those paths into the GAP library also have been an issue in other tests in the past, so there is indeed some merit to suppressing them. Or perhaps modify them on the fly: if the path is prefixed by the GAP root path, then substitute that prefix by `GAPROOT`; similar to how we do for `testspecial` with a script...

Issue #2809, opened 2018-09-13
Diffs when running testbugfix.g without certain packages available


wilfwilson:
I was considering running `tst/testbugfix.g` in the Semigroups CI  setup, as an extra check that Semigroups doesn't break things in GAP. But I can't be bothered installing all the necessary packages that some of the tests use. However, if I don't do so, I get diffs:

### Observed behaviour

For example (GAP master):

```
testing: /home/travis/gap/tst/testbugfix/2012-06-18-t00328.tst
########> Diff in /home/travis/gap/tst/testbugfix/2012-06-18-t00328.tst:2
# Input is:
if LoadPackage("anupq",false) <> fail then
for i in [1..192] do Q:=Pq( FreeGroup(2) : Prime:=3, ClassBound:=1 ); od; fi;
# Expected output:
# But found:
#I  anupq package is not available. Check that the name is correct
#I  and it is present in one of the GAP root directories (see '??RootPaths')
########
```

My understanding of the intention behind the line `if LoadPackage("anupq",false) <> fail then` is that is the package is installed, then run the following test, and if it is not installed, then don't do anything at all (in particular, no warnings). Instead, it prints some info statements.

### Expected behaviour

GAP should not print such info statements when running tests.

alex-konovalov:
I don't see this diff, because I run it with anupq being present in the system . The warning that you see above is displayed when one tries to call `LoadPackage` with the package name that is not known to GAP at all. Compare:
```
gap> LoadPackage("linboxing");
#I  linboxing: The compiled linboxing kernel module is not present.
#I  linboxing package is not available. To see further details, enter
#I  SetInfoLevel(InfoPackageLoading,4); and try to load the package again.
fail
gap> LoadPackage("aaa");
#I  aaa package is not available. Check that the name is correct
#I  and it is present in one of the GAP root directories (see '??RootPaths')
fail
brk> IsPackageMarkedForLoading("aaa","");
false
```
If you want to adapt bugfix test to run in your setup, please submit a PR to add to `testbugfix/2012-06-18-t00328.tst` an extra check for `IsPackageMarkedForLoading("anupq","")` before attempting to load it.

wilfwilson:
Thanks @alex-konovalov, to solve my problem I would have to do this against every `testbugfix` file that requires a package, and there are quite a few. But I am willing to do this. (Even though why do such tests exist in GAP? Surely they should exist in the appropriate packages?).

[...]

PR #1633, opened 2017-08-25
Allow local variables in test files

ChrisJefferson:
This is an experiment in allowing local variables to be declared in test files.

The current (horrible) notation is to write:

`#@local <list of variables to be local>` as the first line of the test. If this line is present, any other introduced variables will produce a warning, as if we were in a function.

There are two reasons to do this:

1) You don't pollute the global name space, local names do not exist at the end of the test.

2) (More importantly) you can run the test in two threads simultaniously, as the two local contexts will not interfere.

This will fail, to demonstrate the new warnings being produced.

UPDATEUPDATE:

This PR also now features `#@if, #@else, #@endif`, which provides an easy way to skip parts of a given test, or give different returns for a given test. This is now used to make the 32-bit, 64-bit and unix tests only run on the correct platforms.

PR #2810, opened 2018-09-13
Add calls to TestPackageAvailability in bugfix tests that load packages

wilfwilson:
Background: the purpose of this PR is to resolve the initial problem that I encountered in https://github.com/gap-system/gap/issues/2809. Basically: I want to run `tst/testbugfix.g` in the Semigroups CI setup, as an extra check that Semigroups doesn't break things in GAP. But I can't be bothered installing the necessary packages that some of the tests use.

However, if I don't install these packages, then I get diffs because of the way that `LoadPackage` produces `Info` statements when it returns `fail`. This causes the tests to fail. See https://github.com/gap-system/gap/issues/2809 for more discussion about this topic. I think that the behaviour of `LoadPackage` with respect to `Info` statements is actually a bug, as I discuss on issue https://github.com/gap-system/gap/issues/2809. This PR doesn't resolve that 'bug', but I will attempt to fix that bug in a separate PR if someone agrees with me that there is a bug.

This is an example of a diff that arises when you don't have a package installed:
```
testing: /home/travis/gap/tst/testbugfix/2012-06-18-t00328.tst
########> Diff in /home/travis/gap/tst/testbugfix/2012-06-18-t00328.tst:2
# Input is:
if LoadPackage("anupq",false) <> fail then
for i in [1..192] do Q:=Pq( FreeGroup(2) : Prime:=3, ClassBound:=1 ); od; fi;
# Expected output:
# But found:
#I  anupq package is not available. Check that the name is correct
########
```
This PR avoids these problems by replacing all occurrences `if LoadPackage("blah") <> fail then` in `testbugfix` test files with `if TestPackageAvailability("blah") <> fail and LoadPackage("blah") <> fail then`, because `TestPackageAvailability` prints no unexpected output when it returns `fail`. Therefore the test files 'pass' cleanly when the relevant package is not able to be loaded.

**There is one exception:** `tst/testbugfix/2005-12-08-t00323.tst`. Please look at the relevant commit. This is not formulated with an if statement, and just blindly loads the package. The test involves printing a table to the screen. Therefore, for the test file to pass without the package loaded, I add the `if` statement as before, and when the condition is not met, I manually print the expected output to screen. It feels weird to do so, but it seems to be the only way to ensure that `testbugfix.g` still passes when you don't have additional packages installed.

The obvious problem with my changes is that, if one of these packages stops loading for some reason, then this won't be captured by the `testbugfix` test files. If this happens and we don't notice that packages are failing to load, then we won't notice a potential regression of the relevant tests.

I'd like people's opinions on whether it is okay to do what I have done. I won't be offended by your comments: I made this PR at the suggestion of @alex-konovalov so it's not my idea anyway; I don't mind finding a different solution to my problem if others think that's best. Perhaps it would be better to move the `bugfix` files that require a package into a new collection? Perhaps we should ensure that such tests are also present in the relevant packages? Perhaps I should just get over it and install the relevant packages in my Travis setup...?

### DIFF
--- a/tst/testbugfix/2012-06-18-t00327.tst
+++ b/tst/testbugfix/2012-06-18-t00327.tst
@@ -1,3 +1,6 @@
 # 2012/06/18 (FL)
-gap> if LoadPackage("cvec",false) <> fail then mat := [[Z(2)]]; 
-> ConvertToMatrixRep(mat,2); cmat := CMat(mat); cmat := cmat^1000; fi;
+gap> if TestPackageAvailability("cvec") <> fail and
+>       LoadPackage("cvec",false) <> fail then
+>   mat := [[Z(2)]]; 
+>   ConvertToMatrixRep(mat,2); cmat := CMat(mat); cmat := cmat^1000;
+> fi;
--- a/tst/testbugfix/00317.tst
+++ b/tst/testbugfix/00317.tst
@@ -1,4 +1,5 @@
 ##  bug 2 for fix 6
-gap> if LoadPackage("tomlib", false) <> fail then
+gap> if TestPackageAvailability("tomlib") <> fail and
+>       LoadPackage("tomlib", false) <> fail then
 >      DerivedSubgroupsTom( TableOfMarks( "A10" ) );
 >    fi;
--- a/tst/testbugfix/00319.tst
+++ b/tst/testbugfix/00319.tst
@@ -1,5 +1,6 @@
 ##  Bug 18 for fix 4
-gap> if LoadPackage("ctbllib", false) <> fail then
+gap> if TestPackageAvailability("ctbllib") <> fail and
+>       LoadPackage("ctbllib", false) <> fail then
 >      if Irr( CharacterTable( "WeylD", 4 ) )[1] <>
 >           [ 3, -1, 3, -1, 1, -1, 3, -1, -1, 0, 0, -1, 1 ] then
 >        Print( "problem with Irr( CharacterTable( \"WeylD\", 4 ) )[1]\n" );
--- a/tst/testbugfix/2005-05-03-t00324.tst
+++ b/tst/testbugfix/2005-05-03-t00324.tst
@@ -1,5 +1,6 @@
 # 2005/05/03 (BH)
-gap> if LoadPackage("crisp", false) <> fail then
+gap> if TestPackageAvailability("crisp") <> fail and
+>       LoadPackage("crisp", false) <> fail then
 >      F:=FreeGroup("a","b","c");;
 >      a:=F.1;;b:=F.2;;c:=F.3;;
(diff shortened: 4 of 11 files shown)


fingolfin:
See also PR #1633 and the `#@skipif` extension to the `.tst` format it implements -- that could be applied to these tests, too, I think (but that doesn't mean we shouldn't merge this PR here as-is, just thought I'd bring it up for completeness.)

tst/testbugfix/2012-06-18-t00327.tst
```diff
@@ -1,3 +1,6 @@
 # 2012/06/18 (FL)
-gap> if LoadPackage("cvec",false) <> fail then mat := [[Z(2)]]; 
-> ConvertToMatrixRep(mat,2); cmat := CMat(mat); cmat := cmat^1000; fi;
+gap> if TestPackageAvailability("cvec") <> fail and
+>       LoadPackage("cvec",false) <> fail then
+>   mat := [[Z(2)]]; 
+>   ConvertToMatrixRep(mat,2); cmat := CMat(mat); cmat := cmat^1000;
+> fi;
```
fingolfin: This goes back to this commit (only available in my "secret" git history):
```diff
commit 614132d823d43e6f036e6352a54015b0889e77a7
Author: Frank Lübeck <frank.luebeck@math.rwth-aachen.de>
Date:   2012-06-18 12:43:23 +0000

    Avoid ConvertToMatrixRep and use Matrix. Now POW_MAT_INT also works
    for matrices in IsCMatRep.
    FL

diff --git a/lib/matrix.gi b/lib/matrix.gi
index 7ebde84e4..3db395796 100644
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -4168,7 +4168,7 @@ BindGlobal("POW_MAT_INT", function(mat, n)
         until r <> true;
       fi;
     od;
-    ConvertToMatrixRep(t);
+    t := Matrix(t, m);
     return t;
   end;
   # compared to standard method, we avoid some zero or identity matrices
```

wilfwilson:
Well-spotted @fingolfin. I will sort this out tomorrow at GAP Days

wilfwilson:
@fingolfin How does the backporting work? I've rebased on master and added the label `backport-to-4.10`; if/when this PR is merged into `master`, will this then be cherry-picked into `stable-4.10`? Or do I then open another PR against `stable-4.10`?

alex-konovalov:
@wilfwilson 
> when this PR is merged into master, will this then be cherry-picked into stable-4.10
yes, exactly - by one of the people who have appropriate access and who will then change the label from `backport-to-4.10` to `backport-to-4.10-DONE`. No need for you to open a second PR.

wilfwilson:
Thanks @alex-konovalov

fingolfin:
Backported to stable-4.10 in da66cbca414a57f24db80dd83190e2928bb4fd0a

PR #2813, opened 2018-09-15
kernel: make Panic() varadic and output where it is called from

fingolfin:


src/gasman.c
```diff
@@ -900,11 +900,7 @@ Bag * GlobalByCookie(
 {
   UInt i,top,bottom,middle;
   Int res;
-  if (cookie == 0L)
-    {
-      Pr("Panic -- 0L cookie passed to GlobalByCookie\n",0L,0L);
-      SyExit(2);
-    }
+  GAP_ASSERT(cookie != 0);
```
ChrisJefferson: This seems to change behaviour and allow 0 cookies in release mode -- why not `if(cookie == 0) { Panic(...) }`, to more closely match the old code?

fingolfin: 100% agreed -- I am confused to see this, not sure what I was thinking.

fingolfin: Fixed now

PR #2814, opened 2018-09-15
Misc changes: fix minor issue in gac output; add test for TraceImmediateMethods; tweak reader

fingolfin:


src/read.c
```diff
@@ -2137,16 +2137,17 @@ static void ReadAtomic (
...
+    }
 
     /* 'atomic' <QualifiedExpression> {',' <QualifiedExpression> } 'do'    */
     TRY_IF_NO_ERROR { IntrAtomicBegin(); }
 
     ReadQualifiedExpr( S_DO|S_OD|follow, 'r' );
     nexprs = 1;
     while (STATE(Symbol) == S_COMMA) {
-      Match( S_COMMA, "comma", follow | S_DO | S_OD );
+      Match( S_COMMA, ",", follow | S_DO | S_OD );
```
fingolfin: Note that we use `","` in all other places that match for `S_COMMA`; this change ensures that we generate consistent error messages.

PR #2816, opened 2018-09-16
Fix and improve CheckOutputDelegations

fingolfin:
Fix and improve CheckOutputDelegations

`CheckOutputDelegations` was unusable because the return value of
`MethodsOperation` slightly changed. In addition to fixing this, this PR also
makes `CheckOutputDelegations` use the `location` member of the records
returned by `MethodsOperation`, which results in more helpful output in some
cases.

Finally, some false positives and false negatives are avoided now: the code
works by printing the function, then scanning the output for matches to
"forbidden" strings, like the word "Display". It already check if the
forbidden string only occurred as a prefix, so "DisplayFOOBAR" would not match.
With this commit, it also checks if the forbidden string occurs in the middle
of another word, so "FOOBARDisplay" also won't match; this gets rid of some
false positives.

On the other hand, we now check *all* matches of a given forbidden string,
not just the first, until we either find a proper match, or have check them
all; this prevents some false negatives.

Regarding the improved location tracking, we used to show this:
```
Detected incorrect delegations for PrintObj
---------------------------------------------------------------
PrintObj: for a blue striped manifold
 delegates to [ "ViewObj" ]
Filename: fail, line : fail
<Operation "ViewObj">
```

Now, we show this:
```
Detected incorrect delegations for PrintObj
---------------------------------------------------------------
PrintObj: for a blue striped manifold
 delegates to [ "ViewObj" ]
Filename: GAPROOT/pkg/SOMEPACKAGE/gap/SOMEFILE.gi:123
<Operation "ViewObj">
```

### DIFF
--- a/tst/testutil.g
+++ b/tst/testutil.g
@@ -436,21 +436,20 @@ end);
 ##
 BindGlobal( "CheckOutputDelegations",
 function()
-local rules, name, f, str, ots, met, pos, nargs, r, i,
+local rules, name, f, str, ots, met, pos, nargs, r, i, tmp,
       report, line, m, n, illegal_delegations, checklist;
 
 rules := [ "Display", "ViewObj", "PrintObj", "DisplayString",
            "ViewString", "PrintString", "String" ];
 
-for name in rules do
-
-  pos:=Position( rules, name );
+for pos in [1..Length(rules)] do
+  name := rules[pos];
   report:=[];
 
   for nargs in [1..2] do
     f:=MethodsOperation( ValueGlobal(name), nargs );
     for m in f do
-      met := m.method;
+      met := m.func;
       str := "";
       ots := OutputTextString(str,true);;
       PrintTo( ots, met );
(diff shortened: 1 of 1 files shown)


fingolfin:
I suggest we backport this to stable-4.10: There, `CheckOutputDelegations` is also broken. And this change is in an optional file, used only for testing, and `CheckOutputDelegations` is never called directly, so I don't see harm coming from it; but backporting this fix has the advantage that package developers can use `CheckOutputDelegations` in stable 4.10 releases.

Alternatively, I could also just push the single line `m.method -> m.func` fix to `stable-4.10`.

alex-konovalov:
We need to insert this into some test, to ensure that we run it!

fingolfin:
Agreed; originally, I just wanted to move the whole function into a new file `tst/testinstall/CheckOutputDelegations .tst`, so that it is always tested. But that would break with `LoadAllPackage`. So we first need package to fix their delegation issues; then we can add a proper test.
Reports generated so far:
* https://bitbucket.org/jdebeule/fining/issues/16
* https://github.com/gap-packages/Semigroups/issues/518
* https://github.com/gap-packages/rcwa/issues/7

fingolfin:
In addition to running it as parts of tests, we might also want to move it to the library, and document it in the part of the GAP library that documents these delegation rules, as a mans for people to check that.

fingolfin:
Thanks @wilfwilson ; also Fininig already fixed it, and a release for it is planned for this week. That leaves RCWA

wilfwilson:
And @james-d-mitchell has now released Semigroups v3.0.19

alex-konovalov:
> In addition to running it as parts of tests, we might also want to move it to the library, and document it in the part of the GAP library that documents these delegation rules, as a mans for people to check that.

Support this @fingolfin - will you extend this PR then?

fingolfin:
I don't really want to spend any more time on this now. I'd rather post an issue reminding about this, and worry about it some other time, when I am not already busy with 100 other things

Issue #2818, opened 2018-09-16
Test failures due to method re-ordering


fingolfin:
[ ping to @alex-konovalov  @stevelinton @frankluebeck @ThomasBreuer @ChrisJefferson @markuspf ]

Having merged PR #2773, which updates the ordering of messages as new implications are added (fixing various issues, see issue #2513), we now are seeing a lot of test failures after `LoadAllPackages()`.

Some of them are fixed by PR #2815, but by far not all. The ones I looked at so far all follow the same pattern: two methods A, B in the GAP library where A used to be ranked above B reverse order when all packages are loaded, due to changes in the relative ranking of their involved filters.

In some cases, there already was a static rank adjustment, i.e., by a *fixed* number, applied in the library; but that change becomes insufficient if one filter increases massively when all packages are loaded, the other are not.

Example 1:
```
gap> Enumerator( Integers mod 9 );
<enumerator of (Integers mod 9)>
gap> LoadAllPackages();
...
gap> Enumerator( Integers mod 9 );
[ ZmodnZObj( 0, 9 ), ZmodnZObj( 1, 9 ), ZmodnZObj( 2, 9 ), ZmodnZObj( 3, 9 ),
  ZmodnZObj( 4, 9 ), ZmodnZObj( 5, 9 ), ZmodnZObj( 6, 9 ), ZmodnZObj( 7, 9 ),
  ZmodnZObj( 8, 9 ) ]
```

Explanation for example 1: Before loading all packages, this `ViewObj` method is used from `coll.gi:528`:
```gap
InstallMethod( ViewObj,
    "for an enumerator that perhaps has its own `ViewObj' function",
    [ IsEnumeratorByFunctions ], 20,
```
Note the manual rank increase by 20. After loading all packages, this method is used instead:
```gap
InstallMethod( ViewObj,
    "for finite lists",
    [ IsList and IsFinite ],
function( list )
```
Yet in a fresh GAP session, started with `gap -A`, we observe this:
```
gap> RankFilter(IsEnumeratorByFunctions);
7
gap> RankFilter(IsList and IsFinite);
4
gap> LoadAllPackages();
...
gap> RankFilter(IsEnumeratorByFunctions);
7
gap> RankFilter(IsList and IsFinite);
139
```

Oops. So the meager static rank adjustment by 20 is not enough. We'd have to use at least 133 instead. But of course that won't work, unless one adds more packages.

[...]

PR #2819, opened 2018-09-17
Add SUM_FLAGS to rank of IsomorphismPermGroup for IsPermGroup

wilfwilson:
(This PR arose from https://github.com/gap-system/gap/issues/2818).

As discussed by @fingolfin in https://github.com/gap-system/gap/issues/2818#issuecomment-421863573, different GAP library methods for `IsomorphismPermGroup` for a perm group apply, depending on which/how many packages are loaded. It seems sensible that `IsomorphismPermGroup` should always return the `IdentityMapping` for a perm group. Adding `SUM_FLAGS` should achieve this.

Note that when/if this PR is merged, the tests in the development version of Semigroups will be able to run without errors in both the `master` and `stable-4.10` branches, given https://github.com/gap-packages/Semigroups/pull/524.

### DIFF
--- a/lib/ghomperm.gi
+++ b/lib/ghomperm.gi
@@ -1928,7 +1928,8 @@ end);
 InstallMethod( IsomorphismPermGroup,
     "perm groups",
     true,
-    [ IsPermGroup ], 0,
+    [ IsPermGroup ],
+    SUM_FLAGS,
     IdentityMapping );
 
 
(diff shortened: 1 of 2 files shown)


wilfwilson:
Thanks @fingolfin – at least according to our tests, the `master` and `stable` branches of Semigroups now run fine against both the `master` and `stable` branches of GAP.

PR #2820, opened 2018-09-18
Remove lib/demo.g

ssiccha:
The functions in these file are unused, undocumented and a proper
demo functionality is now available via the browse package.


ssiccha:
I think the sensible thing to do would then be to document this function (and make sure that the documentation also points to the more sophisticated demo feature in browse). And yes, ideally there shouldn't be two similar versions of the same file. I've opened issue gap-packages/scscp#24 on SCSCP. :)

PR #2830, opened 2018-09-19
Fix GreensXClasses for a Green's class

wilfwilson:
Previously, there was undocumented and broken support for computing the `GreensXClasses` (where `X` is in `{H, L, R}` contained in a particular type of Green's class. This arose from the following issue on the Semigroups package issue tracker: https://github.com/gap-packages/Semigroups/issues/413

I have now made a simple fix, and documented this behaviour (it doesn't have brilliant performance, but it is better than an error).

### DIFF
--- a/lib/semirel.gd
+++ b/lib/semirel.gd
@@ -200,23 +200,33 @@ DeclareSynonymAttr("AssociatedSemigroup", ParentAttr);
 
 #############################################################################
 ##
-#A  GreensRClasses(<semigroup>)
-#A  GreensLClasses(<semigroup>)
-#A  GreensJClasses(<semigroup>)
-#A  GreensDClasses(<semigroup>)
-#A  GreensHClasses(<semigroup>)
+#A  GreensRClasses(<S>)
+#A  GreensLClasses(<S>)
+#A  GreensHClasses(<S>)
+#A  GreensJClasses(<S>)
+#A  GreensDClasses(<S>)
 ##
 ##  <#GAPDoc Label="GreensRClasses">
 ##  <ManSection>
-##  <Attr Name="GreensRClasses" Arg='semigroup'/>
-##  <Attr Name="GreensLClasses" Arg='semigroup'/>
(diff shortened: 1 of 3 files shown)


lib/semirel.gd
```diff
@@ -200,23 +200,33 @@ DeclareSynonymAttr("AssociatedSemigroup", ParentAttr);
...
-##  <Attr Name="GreensRClasses" Arg='semigroup'/>
-##  <Attr Name="GreensLClasses" Arg='semigroup'/>
-##  <Attr Name="GreensJClasses" Arg='semigroup'/>
-##  <Attr Name="GreensDClasses" Arg='semigroup'/>
-##  <Attr Name="GreensHClasses" Arg='semigroup'/>
+##  <Attr Name="GreensRClasses" Arg='S'/>
+##  <Attr Name="GreensLClasses" Arg='S'/>
+##  <Attr Name="GreensHClasses" Arg='S'/>
+##  <Attr Name="GreensJClasses" Arg='S'/>
+##  <Attr Name="GreensDClasses" Arg='S'/>
```
james-d-mitchell: I'm not sure the rationale behind using \" and \' here?

lib/semirel.gi
```diff
@@ -720,13 +720,13 @@ function(semi)
 end);
 
 InstallOtherMethod(GreensHClasses, "for a Green's Class", true, [IsGreensDClass], 0,
```
james-d-mitchell: It might be better to `DeclareAttribute("GreensHClasses", IsGreensDClass);` and drop the `Other` here, but maybe there is some reason I don't know for doing it this way.

Also I believe the `true` and the `0` are redundant, and could be removed.

wilfwilson:
Thank you for reviewing @james-d-mitchell

wilfwilson:
@fingolfin: I am happy with it as-is. @james-d-mitchell revoked his write access to GAP, and I never had write access, so it'll have to be merged by someone else.

Issue #2832, opened 2018-09-19
loading workspaces doesn't work with `--nointeract` option


dimpase:
Please use the following template to submit an issue
(you may delete lines which are not used). Thank You!

### Observed behaviour

`gap --nointeract -L foo.ws`
still gives a prompt

### Expected behaviour

It should not. More precisely, an executable linking libgap should be able to 
load a saved workspace and keep going after this. This also does not work ATM.

This is with `GAP 4.10dev-1163-g94ead70-dirty of today`

dimpase:
Note that #2833 presently is broken due to the current issue, at least according to @fingolfin and @markuspf

ChrisJefferson:
I suspect the `SESSION` at session.g:73 needs to be wrapped in `if not GAPInfo.CommandLineOptions.norepl then` like init.g:1045.

dimpase:
A fix that work for libgap, as well, won't touch `*.g` files, IMHO.

ChrisJefferson:
I'm not clear what the problem is for libgap, and how libgap loads workspaces and then runs GAP. If you don't want GAP to run the repl, you have to tell it not to run the repl.

dimpase:
It's apparently at the level of `argv` parsing for (lib)gap; if I run the corresponding test from #2833 without `-q --nointeract` I do see that otherwise everything seems to work:
```
$ ./test-libgapwsload   -A -l . -m 32m -T   -L /tmp/libgap.ws
# Initializing GAP...
 ┌───────┐   GAP 4.10dev-1164-geb4850f-dirty of today
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64
 Configuration:  gmp 6.1.2, readline
 Loaded workspace: /tmp/libgap.ws
 Packages:   GAPDoc 1.6.1, PrimGrp 3.3.1, SmallGrp 1.3, TransGrp 2.0.4
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap> h;
<fp group on the generators [ f1, f2 ]>
gap> 
```

dimpase:
#2840 works for me with the test from #2833.

PR #2833, opened 2018-09-19
testing for libgap saving/loading workspaces

dimpase:
Currently loading does not work due to GAP issue #2832


dimpase:
With #2840 the workspace loading test works - I'll polish it a bit more.

dimpase:
github doesn't want to update this pull request for some reason...

dimpase:
OK, so now it's ready for reviewing. I've also split the only previously present in tst/testlibgap
test, basic.c, by @markuspf ,  into two files, to be able to reuse what's there in what I added.

dimpase:
Certainly, copy-pasting make targets is far from ideal. However, the workspace test is somewhat special, as we want to execute two applications in the correct order.

Perhaps we might also want to test that libgap-generated workspace loads in a "normal" GAP, and the other way around?

One way or another, the makefile rules probably would need to be adapted to become crossplatforms, taking filename suffixes into account.

markuspf:
Have a look at the `tst/test-compile` directory: It has some scripts that run the tests; I imagine something like this is a viable approach to testing libgap without polluting `Makefile.rules`.

dimpase:
Building would still need to be done by make. A part of the issue is that we'd like a generic rule to make an executable FOO linked against libgap and common.lo, something like:
```
        $(LINK) $(GAP_LDFLAGS) obj/tst/testlibgap/$(FOO).lo  obj/tst/testlibgap/common.lo  libgap.la -o test-libgap$(FOO)
```
However I'd rather leave it for an expert in GAP makefiles (@fingolfin ?) to implement.

Once this is there, writing a script invoking make and running tests won't be a problem.

dimpase:
I will try putting `-L ...` into args to the call to `GAP_Initialize()` - this will allow to have just one make target for all the tests affected by this PR.

fingolfin:
I already added such a build rule to this PR on friday, while you were sitting next to me. It is still there, and in use.

dimpase:
You add the build rule, yes. I am talking about extending it to run tests,
as well.

Makefile.rules
```diff
@@ -984,26 +984,25 @@ testbugfix: all
           ReadGapRoot( "tst/testbugfix.g" );' | $(TESTGAP) | \
             tee `date -u +dev/log/testbugfix2_%Y-%m-%d-%H-%M` )
 
+# run a test in tst/testlibgap
 tst/testlibgap/%: obj/tst/testlibgap/%.lo obj/tst/testlibgap/common.lo libgap.la
-	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) $^ -o $@
+	$(LINK) $(GAP_LDFLAGS) $^ -o $@
```
fingolfin: Why is the QUIET_LINK dropped?

Makefile.rules
```diff
@@ -984,26 +984,25 @@ testbugfix: all
...
+	$@ -A -l $(top_srcdir) -m 32m -q -T --nointeract > $@.out
+	diff $(top_srcdir)/$@.expect $@.out
```
fingolfin: I disagree with this part of the change: IMHO, the build rules for the test executables should be separate from the (PHONY) rules that run the tests.

Indeed, right now these targets are *not* marked as .PHONY, so I expect (but did not try) that running `make testlibgap` twice will not actually run any tests the second time around.

One way to do this should be to remove the above two lines here again; and instead have a for loop in `testlibgap` which iterates over `$^` (resp. `tst/testlibgap/basic tst/testlibgap/wscreate tst/testlibgap/wsload`) and for each executes the above code. That way, you also automatically get the order right.

Makefile.rules
```diff
@@ -984,26 +984,25 @@ testbugfix: all
...
+
+# rm -f $@.out
```
fingolfin: Remove this line?

tst/testlibgap/wsload.c
```diff
@@ -4,7 +4,14 @@
...
 int main(int argc, char ** argv)
 {
-    GAP_Initialize(argc, argv, environ, 0L, 0L);
+    char *args[50];
+    char lpar[3] = "-L";
+    char wsname[16] = "/tmp/libgap.ws"; /* the name must match the one used in wscreate.c */
+    memcpy(args, argv, argc*sizeof(char*));
+    args[argc] = lpar;
+    args[argc+1] = wsname;
+    args[argc+2] = NULL;
```
fingolfin: I think (but have no  verified) that you could replace the above hackery by just this:
```C
  SyRestoring = wsname;
```

tst/testlibgap/wsload.c
```diff
@@ -4,7 +4,14 @@
 #include "common.h"
 int main(int argc, char ** argv)
 {
-    GAP_Initialize(argc, argv, environ, 0L, 0L);
+    char *args[50];
+    char lpar[3] = "-L";
+    char wsname[16] = "/tmp/libgap.ws"; /* the name must match the one used in wscreate.c */
```
fingolfin: Note that `/tmp` might not exist; it might also be a security issue to put a .ws there, in case the test is run on a shared machine. Finally, this will break if these tests are run by multiple jobs in parallel. However, none of these are really blockers for this PR, IMHO, but I wanted to point them out anyway.

tst/testlibgap/common.h
```diff
@@ -0,0 +1,11 @@
+#include <stdio.h>
+#include <unistd.h>
+#include <compiled.h>
```
fingolfin: I'd move the `compiled.h` include to `common.c`; after all, you don't want to expose all GAP innards to all test files, but rather want to focus on testing libgap-api.h, no?

tst/testlibgap/wscreate.c
```diff
@@ -0,0 +1,18 @@
+/*
+ * Small program to test libgap ability to save workspaces.
```
fingolfin: While I don't mind adding such a test, I wonder how useful it is in practice; I imagine that for most users of libgap, saving/restoring workspaces will not be usable, because any connections between GAP and GAP objects on the one hand, and client code (such as SageMath) and its "objects" will be lost.

dimpase: It is often useful in practice to load a workspace, as it speeds up starting GAP quite a bit.
This is the primary reason for this stuff by be in Sage's libGAP.

Besides, this test uncovered a real bug in libgap  :-)

Makefile.rules
dimpase: oops. Just forgot to put it back...

Makefile.rules
dimpase: ok

tst/testlibgap/common.h
dimpase: well, there are functions in `compiled.h` used outside of `common.c`, e.g. `CollectBags`. Thus doing this move gives
```
tst/testlibgap/basic.c:9:5: warning: implicit declaration of function ‘CollectBags’ [-Wimplicit-function-declaration]
     CollectBags(0, 1);    // full GC
     ^~~~~~~~~~~
...
```
I'd leave this as it is.

tst/testlibgap/wsload.c
dimpase: I added a comment in the source TODO this properly at some point.
(see the last commit)

Makefile.rules
dimpase: added back

Makefile.rules
dimpase: fixed in the latest commit

dimpase:
@fingolfin : hopefully this addresses all your concerns; (I'd squash the commits then).

Makefile.rules
```diff
@@ -990,10 +990,23 @@ testbugfix: all
...
+tst/testlibgap/%: obj/tst/testlibgap/%.lo obj/tst/testlibgap/common.lo libgap.la
+	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) $^ -o $@
+
+clean: clean-testlibgap
+clean-testlibgap:
+	rm -f ${LIBGAPTESTS}
+
+# run all the tests in tst/testlibgap
+testlibgap: ${LIBGAPTESTS}
+	$(foreach v,$^,$(v) -A -l $(top_srcdir) -m 32m -q -T --nointeract >$(v).out; diff $(top_srcdir)/$(v).expect $(v).out;)
```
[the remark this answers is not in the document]
dimpase: it's make's [foreach](https://www.gnu.org/software/make/manual/html_node/Foreach-Function.html). I can probably format this line better, split it, if you like...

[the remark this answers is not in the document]
dimpase: minimum workspace size has been set in @markuspf 's original line in `Makefile.rules` to run the only available libgap's test.

tst/testlibgap/wsload.c
fingolfin: Did you try this?

dimpase: for starters, this does not compile, and I wonder whether that `SyRestoring` should be exported.

fingolfin: Fair enough

Makefile.rules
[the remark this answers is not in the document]
fingolfin: I think reformatting this to cover multiple lines would improve the readability. I'd also drop the `-m 32m` -- if there was a reason for it, then it can be added back, but that time with a comment explaining it.

Other than that, this PR seems ready to me. Thank you for your persistence!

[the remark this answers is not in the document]
dimpase: Thanks, fixed  in the latest commit.

fingolfin:
LGTM; perhaps do some squashing, though?

dimpase:
rebasing all this properly is a bit beyond my git skills, with all these interleaving commits that creeped in while pulling from master, sorry. I get  weird-looking errors, like `Could not apply 11a347740b8ba89352c41d75f62a4ac177c56f6...`

fingolfin:
Hu, that's odd; just doing a `git rebase master` worked fine for me, followed by using `git rebase -i` to squash it.

dimpase:
OK, I've managed to squeeze it to 4 commits.

PR #2834, opened 2018-09-19
Improve manual section about Info classes

wilfwilson:
I've tweaked the documentation about `Info` classes. Mostly these are language adjustments, to make it clearer in my opinion, and to specify that an info level can be any non-negative integer. More seriously, the documentation previously said that `InfoWarning` is the only class to start with a non-zero info level; but there are several more, and so I have updated the documentation to match the behaviour of GAP.

Note that this addresses some of my confusion mentioned in the discussion on #2809.

### DIFF
--- a/doc/ref/debug.xml
+++ b/doc/ref/debug.xml
@@ -94,18 +94,34 @@ afterwards.
 <Section Label="Info Functions">
 <Heading>Info Functions</Heading>
 
-The <Ref Func="Info"/> mechanism permits operations to display intermediate results or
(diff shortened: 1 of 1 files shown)


wilfwilson:
@alex-konovalov I implemented some of your changes: I added an index entry, and I mentioned the lack of documentation and the existence of package info classes.

alex-konovalov:
Looks good, thanks @wilfwilson. I think it's worth to backport to GAP 4.10 (as those info classes that are explicitly mentioned there are all present in GAP 4.10).

PR #2835, opened 2018-09-19
Travis: run testinstall with all packages after testpackages

fingolfin:


fingolfin:
I already rebase 2 days ago

fingolfin:
The build log from 2 days ago had lots of errors I cannot reproduce locally. So I have rebased this again now; and I'll also re-run local tests, with a pristine `pkg` obtained via `make boostrap-pkg-full` (to make sure I see the exact same package versions as are used on Travis; I currently use a package distribution tarball from a few days ago, with a few very minor fixes applied; while I don't see how those could mask the errors reported on Travis, let's try to exclude all and any variation)

ChrisJefferson:
This is currently passing. Would it pass if I merged it right now?

alex-konovalov:
@ChrisJefferson no - It should not pass because of the remaining issues after the method reordering (see #2818). This PR passes now only because it has 3 commits instead of one. As far as I understand, this is WiP, and before merge one of the commits should be removed, and another should be moved to a different PR. I've added the "don't merge" label to make it clear.

alex-konovalov:
Two tests currently fail because of the broken XML:
```
XML Parse Error: Line 8273 Character 1
Original file: /home/travis/build/gap-system/gap/doc/ref/../../lib/test.gi, li\
ne number 280.
-----------
</List>
^
-----------
wrong end tag, expecting "</Description>" (starts line 8171)
!!! Type 'Show();' to watch the input string in pager - starting with
    line containing error !!!
```

alex-konovalov:
@fingolfin The problem with building manuals is not present in the master branch. Please rebase.

fingolfin:
I've rebased this now, and it there is on reported test failure before the tests hang for more than 10 minutes in `tst/testinstall/opers/MaximalNormalSubgroups.tst`, on or after line 32. The one test failure that is reported before this is in `tst/testinstall/opers/IsSolvableGroup.tst:15`:
```
# Input is:
IsSolvable(DihedralGroup(IsFpGroup,24));
# Expected output:
true
# But found:
Error, Record Element: '<rec>.absolutelyIrreducible' must have an assigned value
```

Looking closer at the hang, it might actually be related to the above test failure.

alex-konovalov:
@fingolfin can we try again this?

fingolfin:
Rebased

fingolfin:
The 64 bit version went fine through. For the ABI=32 job, all tests also passed. But for some reasons I cannot fathom right now, it failed in the coverage gathering step, see <https://api.travis-ci.org/v3/job/485322346/log.txt> resp. here:
```
> > > gap> gap> Scanning for coverage data...
gap> gap> > > > > >   coverage/testinstall-loadall.coverage
gap> > Merging 1 coverage files...
> Error, Unable to open file coverage/testinstall-loadall.coverage in
  READ_PROFILE_FROM_STREAM( UserHomeExpand( filename ), 0 
 ) at /home/travis/build/gap-system/gap/pkg/profiling/gap/profiling.gi:12 called from 
ReadLineByLineProfile( f 
 ) at /home/travis/build/gap-system/gap/pkg/profiling/gap/profiling.gi:120 called from
<function "MergeLineByLineProfiles">( <arguments> )
 called from read-eval loop at *stdin*:16
```
Perhaps @ChrisJefferson has an idea?

For now, I've added a change to show the content of the `coverage` dir just before the gathering step (I'd leave that change in anyway, we've had problems at the point before).

fingolfin:
So I added that `ls -l`, and am seeing this:
```
+COVDIR=coverage
+ls -l coverage
total 2234212
-rw-rw-r-- 1 travis travis 2287828466 Jan 28 17:31 testinstall-loadall.coverage
+bin/gap.sh --quitonbreak --alwaystrace -q -a 500M -m 500M -q
```
So it seems we got a coverage file with > 2 GB data in it. My guess would be that the 32 bit version of GAP and profiling just are not able to handle that.

Hence one option would be to simply apply this change only to the 64 bit version, but not the 32 bit version. That still serves our purposes, I think, and side steps the issue at hand. Thoughts? @ChrisJefferson @alex-konovalov ?

ChrisJefferson:
I think that's best. I'm quite suprised that coverage file loads in a reasonable time even in 64-bit GAP, but if it's working that's fine. Also, coverage files don't normally get that large, because they should only include each line once...

alex-konovalov:
Happy with 64-bit testing only. Shall we remove "do not merge" label?

PR #2840, opened 2018-09-19
Don't start a session when loading workspace and --nointeract is given

### LINKED ISSUE #2832: loading workspaces doesn't work with `--nointeract` option

markuspf:
Fixes #2832

### DIFF
--- a/lib/session.g
+++ b/lib/session.g
@@ -70,7 +70,9 @@ BindGlobal("POST_RESTORE", function()
             f();
         fi;
     od;
-    SESSION();
+    if not GAPInfo.CommandLineOptions.norepl then
+        SESSION();
+    fi;
     PROGRAM_CLEAN_UP();
 end);
 


embray:
Somehow this still appears to be broken, though I don't fully understand why.  First off, `--norepl` / `--nointeract` are not parsed by the kernel in `system.c`  during `InitSystem`.  That shouldn't be a problem though since `system.g` installs the post-restore function that re-scans the command-line arguments, and includes handling for `--nointeract` and `--norepl`.  That *should* be getting run before `SESSION()`, so we should have `CommandLineOptions.norepl = true` at this point.  But somehow that appears to be failing...

alex-konovalov:
@embray this went to master, while 4.10 release made from stable-4.10 branch. That branch was created off master more than a month ago. This is how GAP major releases are made.

embray:
I understand that, though I wonder why this wasn't backported to the 4.10-stable branch.

ChrisJefferson:
The boring answer is it wasn't fixing an (in our minds) serious issue, as this couldn't lead to incorrect reults and we tend to only merge minimum changes back to stable. Also, someone has to choose to mark the PR as to be merged back.

embray:
Sure, it happens.  Should I open a new issue or PR to backport this fix?

alex-konovalov:
@embray no, I can cherry-pick it once we agree that it could be backported (it seems for me that it could)

ChrisJefferson:
If you label this as should be backported to 4.10, then it should happen at some point (someone usually makes a pass over the tag every so often (I've actually just done it for you)

embray:
Thanks--I don't think I can apply labels.

PR #2843, opened 2018-09-19
kernel: use stat, lstat, mkdtemp, mkstemp unconditionally

fingolfin:
Remove configure checks for these functions, and use them unconditionally (in
fact, we already did for stat). They are part of POSIX and have been present
in Linux, BSD, Mac OS X and cygwin for a long time.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -3453,8 +3453,6 @@ Int SyRmdir ( const Char * name )
 **  for a real directory, 'C' for a character device, 'B' for a block
 **  device 'P' for a FIFO (named pipe) and 'S' for a socket.
 */
-#ifdef HAVE_LSTAT
-
 Obj SyIsDir ( const Char * name )
 {
   Int res;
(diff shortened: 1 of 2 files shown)


src/sysfiles.c
```diff
@@ -3483,15 +3481,6 @@ Obj SyIsDir ( const Char * name )
   else return ObjsChar['?'];
 }
 
-#else
-
-Obj SyIsDir ( const Char * name )
-{
-  return ObjsChar['?'];
```
fingolfin: I'd rather get an error on platforms that don't have `lstat`, so that an alternative can be implemented, than silently degrading this functionality.

src/sysfiles.c
```diff
@@ -3540,41 +3527,6 @@ Char *SyTmpname ( void )
...
-    count++;
-    if ( base == 0 ) {
-        base = tmpnam( (char*)0 );
-    }
-    if ( base == 0 ) {
-        SySetErrorNo();
-        return 0;
-    }
-    if (count == 0) { /* one time in 2^32 */
-        strcat( base, "x" ); /* FIXME: BUG, we must not modify the pointer returned by tmpnam */
```
fingolfin: Note the FIXME here; getting rid of this was the original motivation for this PR.

PR #2845, opened 2018-09-20
WIP: New HPC-GAP guard checking without using ward

### LINKED ISSUE #2565: HPC-GAP: access to plist `IntFF` in src/finfield.c is not thread safe
The kernel function `INT_FF` accesses and modifies the plist `IntFF` without taking any care about race conditions. This is in contrast to e.g. `SuccFF`, defined and used in the same kernel module, for which we use `ATOMIC_ELM_PLIST` (instead of `ELM_PLIST`), as well as `ATOMIC_SET_ELM_PLIST_ONCE` etc.

rbehrends:
This is the start of the work on getting rid of Ward as a dependency.

To try this code out, first install `unward` from https://github.com/rbehrends/unward; we assume that it is stored in a directory called `$UNWARD`. To build it, do:

```sh
cd $UNWARD
./configure
make
```

After that, go to the GAP directory and build that:
   
```
./configure --enable-hpcgap
$UNWARD/bin/unward --inplace src
make
```

Fixes #2565

### DIFF
--- a/etc/ci-prepare.sh
+++ b/etc/ci-prepare.sh
@@ -23,7 +23,7 @@ BUILDDIR=$PWD
 
 if [[ $HPCGAP = yes ]]
 then
-  CONFIGFLAGS="--enable-hpcgap $CONFIGFLAGS"
+  CONFIGFLAGS="--enable-hpcgap --enable-guards $CONFIGFLAGS"
 fi
 
 
@@ -40,8 +40,22 @@ then
 fi
 
 
-# configure and make GAP
+# configure
 time "$SRCDIR/configure" --enable-Werror $CONFIGFLAGS
+
+if [[ $HPCGAP = yes ]]
+then
+  git clone https://github.com/rbehrends/unward
+  cd unward
+  ./configure CC=gcc CXX=g++ CFLAGS=-O2 CXXFLAGS=-O2
+  make
+  cd ..
+  unward/bin/unward --inplace src
+  # commit the result to prevent the docomp test from triggering
+  git commit -m "Unward" src
+fi
+
+# build GAP
 time make V=1 -j4
 
 # Use alternative downloader which retries on failure and uses the Travis cache
--- a/src/boehm_gc.c
+++ b/src/boehm_gc.c
@@ -334,6 +334,9 @@ void RetypeBagIntern(Bag bag, UInt new_type)
     if (old_type == new_type)
         return;
 
+#if defined(HPCGAP) && defined(USE_HPC_GUARDS)
+    WriteGuard(bag);
+#endif
     /* change the size-type word                                           */
     header->type = new_type;
     {
@@ -456,6 +459,9 @@ UInt ResizeBag(Bag bag, UInt new_size)
     CollectBags(0, 0);
 #endif
 
+#if defined(HPCGAP) && defined(USE_HPC_GUARDS)
+    WriteGuard(bag);
+#endif
     BagHeader * header = BAG_HEADER(bag);
 
     /* get type and old size of the bag                                    */
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -311,6 +311,8 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr)
 **  calling 'CHANGED_BAG(old)' in the above example (see "CHANGED_BAG").
 */
 #ifdef HPCGAP
+extern int ExtendedGuardCheck(Bag) PURE_FUNC;
+
 EXPORT_INLINE int ReadCheck(Bag bag)
 {
     Region *region;
@@ -321,18 +323,18 @@ EXPORT_INLINE int ReadCheck(Bag bag)
         return 1;
     if (region->readers[TLS(threadID)])
         return 1;
-    return 0;
+    return ExtendedGuardCheck(bag);
 }
 
 EXPORT_INLINE int WriteCheck(Bag bag)
 {
-    Region *region;
+    Region * region;
     region = REGION(bag);
-    return !region || region->owner == GetTLS();
+    return !region || region->owner == GetTLS() || ExtendedGuardCheck(bag);
 }
 
-extern void HandleReadGuardError(Bag);
-extern void HandleWriteGuardError(Bag);
+extern void HandleReadGuardError(Bag) NORETURN;
+extern void HandleWriteGuardError(Bag) NORETURN;
 #endif // HPCGAP
 
 EXPORT_INLINE Bag *PTR_BAG(Bag bag)
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -151,6 +151,46 @@ EXPORT_INLINE UInt TNUM_BAG(Bag bag)
     return CONST_BAG_HEADER(bag)->type;
 }
 
+/****************************************************************************
+**
+*F  ReadCheck(<bag>) . . . . . . . . . . . . . is the bag readable in HPCGAP?
+*F  WriteCheck(<bag>)  . . . . . . . . . . . . is the bag writable in HPCGAP?
+*F  HandleReadGuardError(<bag>)  . . . . . . . . . . signal read access error
+*F  HandleWriteGuardError(<bag>) . . . . . . . . .  signal write access error
+**
+**  These funtion handle access checks in HPCGAP, i.e. whether a bag can
+**  safely be read or modified without causing race conditions. These are
+**  called in functions such as PTR_BAG() and CONST_PTR_BAG(), which should
+**  implicitly or explicitly guard all object accesses. In order to access
+**  objects without triggering checks, alternative functions UNSAFE_PTR_BAG()
+**  and UNSAFE_CONST_PTR_BAG() are provided.
+*/
+#ifdef HPCGAP
+extern int ExtendedGuardCheck(Bag) PURE_FUNC;
+
+EXPORT_INLINE int ReadCheck(Bag bag)
+{
+    Region *region;
+    region = REGION(bag);
+    if (!region)
+        return 1;
+    if (region->owner == GetTLS())
+        return 1;
+    if (region->readers[TLS(threadID)])
+        return 1;
+    return ExtendedGuardCheck(bag);
+}
+
+EXPORT_INLINE int WriteCheck(Bag bag)
+{
+    Region * region;
+    region = REGION(bag);
+    return !region || region->owner == GetTLS() || ExtendedGuardCheck(bag);
+}
+
+extern void HandleReadGuardError(Bag) NORETURN;
+extern void HandleWriteGuardError(Bag) NORETURN;
+#endif // HPCGAP
 
 /****************************************************************************
 **
@@ -310,33 +350,6 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr)
 **  the application  must inform {\Gasman}  that it  has changed  the bag, by
 **  calling 'CHANGED_BAG(old)' in the above example (see "CHANGED_BAG").
 */
-#ifdef HPCGAP
-extern int ExtendedGuardCheck(Bag) PURE_FUNC;
-
-EXPORT_INLINE int ReadCheck(Bag bag)
-{
-    Region *region;
-    region = REGION(bag);
-    if (!region)
-        return 1;
-    if (region->owner == GetTLS())
-        return 1;
-    if (region->readers[TLS(threadID)])
-        return 1;
-    return ExtendedGuardCheck(bag);
-}
-
-EXPORT_INLINE int WriteCheck(Bag bag)
-{
-    Region * region;
-    region = REGION(bag);
-    return !region || region->owner == GetTLS() || ExtendedGuardCheck(bag);
-}
-
-extern void HandleReadGuardError(Bag) NORETURN;
-extern void HandleWriteGuardError(Bag) NORETURN;
-#endif // HPCGAP
-
 EXPORT_INLINE Bag *PTR_BAG(Bag bag)
 {
     GAP_ASSERT(bag != 0);
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -38,6 +38,11 @@
 
 #include "common.h"
 
+#ifdef HPCGAP
+#include "hpc/region.h"
+#include "hpc/tls.h"
+#endif
+
 
 /****************************************************************************
 **
@@ -305,17 +310,70 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr)
 **  the application  must inform {\Gasman}  that it  has changed  the bag, by
 **  calling 'CHANGED_BAG(old)' in the above example (see "CHANGED_BAG").
 */
+#ifdef HPCGAP
+EXPORT_INLINE int ReadCheck(Bag bag)
+{
+    Region *region;
+    region = REGION(bag);
+    if (!region)
+        return 1;
+    if (region->owner == GetTLS())
+        return 1;
+    if (region->readers[TLS(threadID)])
+        return 1;
+    return 0;
+}
+
+EXPORT_INLINE int WriteCheck(Bag bag)
+{
+    Region *region;
+    region = REGION(bag);
+    return !region || region->owner == GetTLS();
+}
+
+extern void HandleReadGuardError(Bag);
+extern void HandleWriteGuardError(Bag);
+#endif // HPCGAP
+
 EXPORT_INLINE Bag *PTR_BAG(Bag bag)
+{
+    GAP_ASSERT(bag != 0);
+#ifdef USE_HPC_GUARDS
+    if (!WriteCheck(bag))
+      HandleWriteGuardError(bag);
+#endif
+    return *(Bag**)bag;
+}
+
+#ifdef USE_HPC_GUARDS
+EXPORT_INLINE Bag *UNSAFE_PTR_BAG(Bag bag)
 {
     GAP_ASSERT(bag != 0);
     return *(Bag**)bag;
 }
+#else
+#define UNSAFE_PTR_BAG PTR_BAG
+#endif
 
 EXPORT_INLINE const Bag *CONST_PTR_BAG(Bag bag)
 {
     GAP_ASSERT(bag != 0);
+#ifdef USE_HPC_GUARDS
+    if (!ReadCheck(bag))
+      HandleReadGuardError(bag);
+#endif
     return *(const Bag * const *)bag;
 }
+
+#ifdef USE_HPC_GUARDS
+EXPORT_INLINE const Bag *UNSAFE_CONST_PTR_BAG(Bag bag)
+{
+    GAP_ASSERT(bag != 0);
+    return *(const Bag * const *)bag;
+}
+#else
+#define UNSAFE_CONST_PTR_BAG CONST_PTR_BAG
+#endif
 
 EXPORT_INLINE void SET_PTR_BAG(Bag bag, Bag *val)
 {
--- a/src/hpc/guards.h
+++ b/src/hpc/guards.h
@@ -11,6 +11,7 @@
 #ifndef GAP_HPC_GUARD_H
 #define GAP_HPC_GUARD_H
 
+#include "gasman.h"
 #include "hpc/region.h"
 #include "hpc/tls.h"
 
@@ -18,101 +19,64 @@
 #error This header is only meant to be used with HPC-GAP
 #endif
 
-#ifdef VERBOSE_GUARDS
-void WriteGuardError(Bag bag,
-    const char *file, unsigned line, const char *func, const char *expr);
-void ReadGuardError(Bag bag,
-    const char *file, unsigned line, const char *func, const char *expr);
-#else
-void WriteGuardError(Bag bag);
-void ReadGuardError(Bag bag);
-#endif
-
-#ifdef VERBOSE_GUARDS
-#define WriteGuard(bag) WriteGuardFull(bag, __FILE__, __LINE__, __FUNCTION__, #bag)
-static ALWAYS_INLINE Bag WriteGuardFull(Bag bag,
-  const char *file, unsigned line, const char *func, const char *expr)
-#else
 static ALWAYS_INLINE Bag WriteGuard(Bag bag)
-#endif
 {
-  Region *region;
-  if (!IS_BAG_REF(bag))
+    if (!WriteCheck(bag))
+        HandleWriteGuardError(bag);
     return bag;
-  region = REGION(bag);
-  if (region && region->owner != GetTLS() && region->alt_owner != GetTLS())
-    WriteGuardError(bag
-#ifdef VERBOSE_GUARDS
-    , file, line, func, expr
-#endif
-    );
-  return bag;
 }
+
 
 EXPORT_INLINE Bag ImpliedWriteGuard(Bag bag)
 {
-  return bag;
+    return bag;
 }
 
 EXPORT_INLINE int CheckWriteAccess(Bag bag)
 {
-  Region *region;
-  if (!IS_BAG_REF(bag))
-    return 1;
-  region = REGION(bag);
-  return !(region && region->owner != GetTLS() && region->alt_owner != GetTLS())
-    || TLS(DisableGuards) >= 2;
+    Region * region;
+    if (!IS_BAG_REF(bag))
+        return 1;
+    region = REGION(bag);
+    return !(region && region->owner != GetTLS() &&
+             region->alt_owner != GetTLS()) ||
+           TLS(DisableGuards) >= 2;
 }
 
 EXPORT_INLINE int CheckExclusiveWriteAccess(Bag bag)
 {
-  Region *region;
-  if (!IS_BAG_REF(bag))
-    return 1;
-  region = REGION(bag);
-  if (!region)
-    return 0;
-  return region->owner == GetTLS() || region->alt_owner == GetTLS()
-    || TLS(DisableGuards) >= 2;
+    Region * region;
+    if (!IS_BAG_REF(bag))
+        return 1;
+    region = REGION(bag);
+    if (!region)
+        return 0;
+    return region->owner == GetTLS() || region->alt_owner == GetTLS() ||
+           TLS(DisableGuards) >= 2;
 }
 
-#ifdef VERBOSE_GUARDS
-#define ReadGuard(bag) ReadGuardFull(bag, __FILE__, __LINE__, __FUNCTION__, #bag)
-static ALWAYS_INLINE Bag ReadGuardFull(Bag bag,
-  const char *file, unsigned line, const char *func, const char *expr)
-#else
 static ALWAYS_INLINE Bag ReadGuard(Bag bag)
-#endif
 {
-  Region *region;
-  if (!IS_BAG_REF(bag))
+    if (!ReadCheck(bag))
+        HandleReadGuardError(bag);
     return bag;
-  region = REGION(bag);
-  if (region && region->owner != GetTLS() &&
-      !region->readers[TLS(threadID)] && region->alt_owner != GetTLS())
-    ReadGuardError(bag
-#ifdef VERBOSE_GUARDS
-    , file, line, func, expr
-#endif
-    );
-  return bag;
 }
-
 
 static ALWAYS_INLINE Bag ImpliedReadGuard(Bag bag)
 {
-  return bag;
+    return bag;
 }
 
 static ALWAYS_INLINE int CheckReadAccess(Bag bag)
 {
-  Region *region;
-  if (!IS_BAG_REF(bag))
-    return 1;
-  region = REGION(bag);
-  return !(region && region->owner != GetTLS() &&
-    !region->readers[TLS(threadID)] && region->alt_owner != GetTLS())
-    || TLS(DisableGuards) >= 2;
+    Region * region;
+    if (!IS_BAG_REF(bag))
+        return 1;
+    region = REGION(bag);
+    return !(region && region->owner != GetTLS() &&
+             !region->readers[TLS(threadID)] &&
+             region->alt_owner != GetTLS()) ||
+           TLS(DisableGuards) >= 2;
 }
 
-#endif // GAP_HPC_GUARD_H
+#endif    // GAP_HPC_GUARD_H
--- a/src/hpc/guards.h
+++ b/src/hpc/guards.h
@@ -27,11 +27,6 @@ static ALWAYS_INLINE Bag WriteGuard(Bag bag)
 }
 
 
-EXPORT_INLINE Bag ImpliedWriteGuard(Bag bag)
-{
-    return bag;
-}
-
 EXPORT_INLINE int CheckWriteAccess(Bag bag)
 {
     Region * region;
@@ -59,11 +54,6 @@ static ALWAYS_INLINE Bag ReadGuard(Bag bag)
 {
     if (!ReadCheck(bag))
         HandleReadGuardError(bag);
-    return bag;
-}
-
-static ALWAYS_INLINE Bag ImpliedReadGuard(Bag bag)
-{
     return bag;
 }
 
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -1311,11 +1311,20 @@ Region * CurrentRegion(void)
     return TLS(currentRegion);
 }
 
+// Kernel debugging information for guards.
+//
+// When compiled with -DDEBUG_GUARDS, the GAP variable GUARD_ERROR_STACK
+// will contain the C stack after an error, allowing us to located where
+// it occurred without having to fire up a debugger first.
+//
+// The variables NumReadErrors and NumWriteErrors track the number of
+// failed guard checks. These are used in conjunction with the low-level
+// function DISABLE_GUARDS() in order to track how many guard failures
+// occur in a given section of code.
+
 #ifdef DEBUG_GUARDS
 extern GVarDescriptor GUARD_ERROR_STACK;
 #endif
-
-// These are temporary debugging functions.
 
 static Int NumReadErrors = 0;
 static Int NumWriteErrors = 0;
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -1363,15 +1363,20 @@ void SetGuardErrorStack(void)
 }
 #endif
 
-void HandleReadGuardError(Bag bag)
+int ExtendedGuardCheck(Bag bag)
 {
-    NumReadErrors++;
     // We shift some of the rarer checks here to
     // avoid overloading ReadCheck().
     if (REGION(bag)->alt_owner == GetTLS())
-        return;
+        return 1;
     if (TLS(DisableGuards))
-        return;
+        return 1;
+    return 0;
+}
+
+void HandleReadGuardError(Bag bag)
+{
+    NumReadErrors++;
     SetGVar(&LastInaccessibleGVar, bag);
 #ifdef DEBUG_GUARDS
     SetGuardErrorStack();
@@ -1386,10 +1391,6 @@ void HandleWriteGuardError(Bag bag)
     NumWriteErrors++;
     // We shift some of the rarer checks here to
     // avoid overloading ReadCheck().
-    if (REGION(bag)->alt_owner == GetTLS())
-        return;
-    if (TLS(DisableGuards))
-        return;
     SetGVar(&LastInaccessibleGVar, bag);
 #ifdef DEBUG_GUARDS
     SetGuardErrorStack();
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -1313,9 +1313,10 @@ Region * CurrentRegion(void)
 
 // Kernel debugging information for guards.
 //
-// When compiled with -DDEBUG_GUARDS, the GAP variable GUARD_ERROR_STACK
-// will contain the C stack after an error, allowing us to located where
-// it occurred without having to fire up a debugger first.
+// When DEBUG_GUARDS is #defined, the GAP variable GUARD_ERROR_STACK
+// is set to a GAP plain list describing the C stack after an error,
+// allowing us to located where it occurred without having to fire up a
+// debugger first.
 //
 // The variables NumReadErrors and NumWriteErrors track the number of
 // failed guard checks. These are used in conjunction with the low-level
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -1311,73 +1311,82 @@ Region * CurrentRegion(void)
     return TLS(currentRegion);
 }
 
-#ifdef VERBOSE_GUARDS
+#ifdef DEBUG_GUARDS
+extern GVarDescriptor GUARD_ERROR_STACK;
+#endif
 
-static void PrintGuardError(char *       buffer,
-                            char *       mode,
-                            Obj          obj,
-                            const char * file,
-                            unsigned     line,
-                            const char * func,
-                            const char * expr)
-{
-    sprintf(buffer, "No %s access to object %llu of type %s\n"
-                    "in %s, line %u, function %s(), accessing %s",
-            mode, (unsigned long long)(UInt)obj, TNAM_OBJ(obj), file, line,
-            func, expr);
-}
-void WriteGuardError(Obj          o,
-                     const char * file,
-                     unsigned     line,
-                     const char * func,
-                     const char * expr)
-{
-    char * buffer = alloca(strlen(file) + strlen(func) + strlen(expr) + 200);
-    ImpliedReadGuard(o);
-    if (TLS(DisableGuards))
-        return;
-    SetGVar(&LastInaccessibleGVar, o);
-    PrintGuardError(buffer, "write", o, file, line, func, expr);
-    ErrorMayQuit("%s", (UInt)buffer, 0);
-}
+// These are temporary debugging functions.
 
-void ReadGuardError(Obj          o,
-                    const char * file,
-                    unsigned     line,
-                    const char * func,
-                    const char * expr)
-{
-    char * buffer = alloca(strlen(file) + strlen(func) + strlen(expr) + 200);
-    ImpliedReadGuard(o);
-    if (TLS(DisableGuards))
-        return;
-    SetGVar(&LastInaccessibleGVar, o);
-    PrintGuardError(buffer, "read", o, file, line, func, expr);
-    ErrorMayQuit("%s", (UInt)buffer, 0);
-}
+static Int NumReadErrors = 0;
+static Int NumWriteErrors = 0;
 
-#else
-void WriteGuardError(Obj o)
-{
-    ImpliedReadGuard(o);
-    if (TLS(DisableGuards))
-        return;
-    SetGVar(&LastInaccessibleGVar, o);
-    ErrorMayQuit(
-        "Attempt to write object %i of type %s without having write access",
-        (Int)o, (Int)TNAM_OBJ(o));
-}
+#ifdef DEBUG_GUARDS
 
-void ReadGuardError(Obj o)
+#include <execinfo.h>
+
+#define BACKTRACE_DEPTH 128
+
+// Record the backtrace in a global GAP variable.
+
+void SetGuardErrorStack(void)
 {
-    ImpliedReadGuard(o);
-    if (TLS(DisableGuards))
-        return;
-    SetGVar(&LastInaccessibleGVar, o);
-    ErrorMayQuit(
-        "Attempt to read object %i of type %s without having read access",
-        (Int)o, (Int)TNAM_OBJ(o));
+    void *  callstack[BACKTRACE_DEPTH];
+    int     depth = backtrace(callstack, BACKTRACE_DEPTH);
+    char ** calls = backtrace_symbols(callstack, depth);
+    Obj     stack = NEW_PLIST_IMM(T_PLIST, depth);
+    SET_LEN_PLIST(stack, depth - 1);
+    for (int i = 1; i < depth; i++) {
+        char * p = calls[i] + 1;
+        char * q = p;
+        while (*p) {
+            if (p[-1] == ' ' && p[0] == ' ')
+                p++;
+            else
+                *q++ = *p++;
+        }
+        *q++ = 0;
+        SET_ELM_PLIST(stack, i, MakeImmString(calls[i]));
+        CHANGED_BAG(stack);
+    }
+    free(calls);
+    SetGVar(&GUARD_ERROR_STACK, stack);
 }
 #endif
+
+void HandleReadGuardError(Bag bag)
+{
+    NumReadErrors++;
+    // We shift some of the rarer checks here to
+    // avoid overloading ReadCheck().
+    if (REGION(bag)->alt_owner == GetTLS())
+        return;
+    if (TLS(DisableGuards))
+        return;
+    SetGVar(&LastInaccessibleGVar, bag);
+#ifdef DEBUG_GUARDS
+    SetGuardErrorStack();
+#endif
+    ErrorMayQuit(
+        "Attempt to read object %i of type %s without having read access",
+        (Int)bag, (Int)TNAM_OBJ(bag));
+}
+
+void HandleWriteGuardError(Bag bag)
+{
+    NumWriteErrors++;
+    // We shift some of the rarer checks here to
+    // avoid overloading ReadCheck().
+    if (REGION(bag)->alt_owner == GetTLS())
+        return;
+    if (TLS(DisableGuards))
+        return;
+    SetGVar(&LastInaccessibleGVar, bag);
+#ifdef DEBUG_GUARDS
+    SetGuardErrorStack();
+#endif
+    ErrorMayQuit(
+        "Attempt to write object %i of type %s without having write access",
+        (Int)bag, (Int)TNAM_OBJ(bag));
+}
 
 #endif
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -32,6 +32,10 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
+#ifdef HAVE_SYS_RESOURCE_H
+#include <sys/resource.h>
+#endif
+
 #ifdef USE_BOEHM_GC
 # ifdef HPCGAP
 #  define GC_THREADS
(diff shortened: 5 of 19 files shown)


fingolfin:
This PR does not actually change the optimization level to -O3, wasn't that expected to be required for optimal performance? Where the benchmark numbers to mentioned earlier done with the default -O2, or with -O3 (perhaps manually inserted into the build system)?

lib/filter.g
```diff
@@ -509,8 +509,10 @@ BIND_GLOBAL( "RankFilter", function( filter )
     else
       all := WITH_IMPS_FLAGS(flags);
     fi;
-    for i  in TRUES_FLAGS(all)  do
-        rank := rank + RANK_FILTERS[i];
+    atomic readonly FILTER_REGION do
+      for i  in TRUES_FLAGS(all)  do
+	  rank := rank + RANK_FILTERS[i];
```
fingolfin: Nitpick: please don't use tabs

rbehrends: Fixed, and updated my vimrc to leave them out for GAP files, too.

rbehrends:
For now, I'm leaving the optimization level at -O2, as -O3 doesn't give me a significant speedup and slows down compilation. This has to be benchmarked more extensively, of course.

ChrisJefferson:
make for unward produces:

```
g++ -g -O2 -DADLIB_DEBUG=0  -I. -c -o build/analyzer.o src/analyzer.cc
make: *** No rule to make target 'src/args.cc', needed by 'build/args.o'. Stop.
```

rbehrends:
@ChrisJefferson Sorry, I did forget to mention re2c. It's possible to package unward with the generated C++ sources (and the Makefile should work in that case, too), but I hadn't made such a package yet.

rbehrends:
The current status is that HPC-GAP should compile and run up to the prompt (and it doesn't blow up for at least the basic operations).

As an additional featured, if you build with `-DDEBUG_GUARDS`, then the GAP variable `GUARD_ERROR_STACK` will contain a C stack backtrace if your system supports the `backtrace()` functionality. This can make it easier to troubleshoot problematic guard errors without having to look at them in the debugger.

markuspf:
So, with the instructions above one still has to have ward installed and the source is processed through ward; is this intentional/needed?

ChrisJefferson:
I don't want to slow things down, but at some point it would be good to have an explanation on where and why UNSAFE calls are put -- is it for correctness or efficiency (or both?)

fingolfin:
@rbehrends I am surprised to read that regarding -O2 vs -O3, as in previous discussions, you emphasized that -O3 was necessary in order for the `pure` functions optimization trick to be fully effective. So now you are saying this is not necessary after all (or perhaps the other way around: even -O3 is not enough to optimize the guards use in e.g. loops sufficiently)?

@markuspf I think this is because this PR simply contains no build system changes at all. As far as I understood, the `ward` support could/should simply be removed completely.

@ChrisJefferson no worries re "slowing down" things; this is a prototype, and it's only here because I badgered @rbehrends into putting it here. I hope he'll find time to finish it, but since he has many other obligations, I wanted to make sure that at least his partial work is available, so that we can see and discuss the general ideas at least, and so that in the worst case, somebody else can pick up the work...

fingolfin:
I had a lengthy comment here but my laptop ran out of power and crashed :/. But roughly it said that I think several `WARD_ENABLED` sections cover too much, resp. the code unward produces needs to be audited: e.g. many string accesses in `integer.c` and `src/hpc/aobjects` should not really be made UNSAFE.

Anyway, @rbehrends FYI I also pushed two commits here which constify more stuff. More could/should be done on that front.

ChrisJefferson:
I the CONST bits were pulled out, I'd be happy to merge them now (well, assuming packages compile and stuff, obviously).

src/gasman.h
```diff
@@ -299,13 +304,64 @@ static inline UInt SIZE_BAG_CONTENTS(const void *ptr) {
...
+        return 1;
+    if (region->readers[TLS(threadID)])
+        return 1;
+    return 0;
+}
+
+static inline PURE_FUNC int WriteCheck(Bag bag)
+{
+    Region *region;
+    if (!IS_BAG_REF(bag))
```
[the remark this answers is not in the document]
rbehrends: Yes, this is placeholder code, simply copied and pasted from the old guard code. My focus was on getting GAP and the basic tests to run first, then we can clean up and refine the code.

rbehrends:
Okay, right now I've got `tst/testinstall.g` at least running without errors. There are still some parts that aren't ... pretty, but this gives us a more complete basis to work from.

fingolfin:
@ChrisJefferson done, see PR #2872

fingolfin:
I rebased this, and also added a hack commit that automatically runs `unward` on Travis in HPC-GAP builds (let's see if it works as intended). That should result in more meaningful Travis output. I also disabled most Travis and AppVeyor builds, to get quicker CI feedback.

Looking at the errors in `testinstall` [here](https://travis-ci.org/gap-system/gap/jobs/436709889), there aren't so many left anymore. Nice.

fingolfin:
In order to debug a test failure, I entered this on the GAP prompt:
```
atomic readonly Z_MOD_NZ do pos:= Position( Z_MOD_NZ[1], n ); Print(pos,"\n"); od;
```
That worked fine on the main thread. But entering it on another thread leads to a segfault, with this backtrace:
```
(lldb) bt
* thread #2: tid = 0x41a083, 0x0000000100c48459 libgc.1.dylib`GC_clear_stack_inner + 41, stop reason = EXC_BAD_ACCESS (code=2, address=0x112009f48)
  * frame #0: 0x0000000100c48459 libgc.1.dylib`GC_clear_stack_inner + 41
    frame #1: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #2: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #3: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #4: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #5: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #6: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #7: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #8: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #9: 0x0000000100c48499 libgc.1.dylib`GC_clear_stack_inner + 105
    frame #10: 0x0000000100c48586 libgc.1.dylib`GC_clear_stack + 150
    frame #11: 0x0000000100c42e18 libgc.1.dylib`GC_generic_malloc_many + 792
    frame #12: 0x0000000100c432b7 libgc.1.dylib`GC_malloc_many + 55
    frame #13: 0x00000001001eb942 gap`AllocateBagMemory(gc_type=-1, type=18, size=168) + 306 at boehm_gc.c:242
    frame #14: 0x00000001001ebbc6 gap`NewBag(type=18, size=160) + 118 at boehm_gc.c:488
    frame #15: 0x00000001001acc66 gap`NewLVarsBag(slots=17) + 150 at vars.c:127
    frame #16: 0x0000000100094e8d gap`SwitchToNewLvars(func=0x0000000103661210, narg=1, nloc=16) + 61 at vars.h:204
    frame #17: 0x0000000100094b0d gap`DoExecFunc(func=0x0000000103661210, narg=1, arg=0x000000011200ef40) + 125 at funcs.c:496
    frame #18: 0x000000010009304d gap`DoExecFunc1args(func=0x0000000103661210, a1=0x000000010bba09c0) + 61 at funcs.c:525
    frame #19: 0x000000010003a57c gap`DoWrap2args(self=0x0000000103661210, arg1=0x000000010bba09e0, arg2=0x000000010bba09d0) + 140 at calls.c:201
    frame #20: 0x0000000100083fd2 gap`CallErrorInner(msg="Attempt to read object %i of type %s without having read access", arg1=4351990880, arg2=4297217474, justQuit=0, mayReturnVoid=0, mayReturnObj=0, lateMessage=0x000000010bba0a00, printThisStatement=1) + 882 at error.c:388
    frame #21: 0x0000000100083999 gap`ErrorMayQuit(msg="Attempt to read object %i of type %s without having read access", arg1=4351990880, arg2=4297217474) + 121 at error.c:543
    frame #22: 0x00000001001fc442 gap`ReadGuardError(o=0x0000000103661c60) + 98 at thread.c:1282
    frame #23: 0x00000001000dcb20 gap`ReadGuard(bag=0x0000000103661c60) + 160 at guards.h:83
    frame #24: 0x00000001000dc9d9 gap`FuncLENGTH(self=0x0000000102164bd0, list=0x0000000103661c60) + 25 at lists.c:154
    frame #25: 0x0000000100096bed gap`EvalOrExecCall(ignoreResult=0, nr=1, call=160) + 2013 at funcs.c:166
    frame #26: 0x0000000100095fc3 gap`EvalFunccall1args(call=160) + 35 at funcs.c:321
    frame #27: 0x0000000100086149 gap`EvalGt(expr=184) + 217 at exprs.c:395
    frame #28: 0x0000000100085364 gap`EvalAnd(expr=208) + 436 at exprs.c:182
    frame #29: 0x0000000100085289 gap`EvalAnd(expr=296) + 217 at exprs.c:174
...
    frame #10337: 0x00007fff9c1c5351 libsystem_pthread.dylib`thread_start + 13
```

I think the problem is that an error is triggered which caused a guard check to fail, which triggers an error, which causes a guard check to fail, etc., until the stack is full

fingolfin:
Just to clarify: I entered this with `n` undefined! So on the main thread, it causes an error, while on another thread, a crash.

rbehrends:
I've rebased and reorganized the changes, which are now down to <200 lines of code.

The stack overflow if an error occurs outside the main thread still needs to be fixed.

src/finfield.c
```diff
@@ -1474,6 +1475,9 @@ Obj FuncLOG_FFE_DEFAULT (
 **  the smallest integer <i> such that '<i> * <z>^0 = <z>'.
 */
 Obj IntFF;
+#ifdef HPCGAP
+Int NumFF;
+#endif
```
fingolfin: Note: the changes in here might conflict with changes in PR #2997 resp. PR by @stevelinton and @jdebeule -- we need to find a way forward there. 

I just looked a the two PRs side by side, and it *seems* to be that they do not actually overlap, so we might be fine, but it'd be good to verify this.

rbehrends: Yeah, I'm aware of that PR. This is actually some code that I'd like to simplify, anyway. We have data structures that get basically initialized once and then don't get changed anymore; it should be possible to remove many the hoops I'm currently jumping through to present a consistent picture to all threads.

src/hpc/thread.c
```diff
@@ -1283,4 +1286,75 @@ void ReadGuardError(Obj o)
...
+Int NumWriteErrors = 0;
+volatile int GuardDummy;
+
+#ifdef DEBUG_GUARDS
+
+#include <execinfo.h>
+
+#define BACKTRACE_DEPTH 128
+
+void SetGuardErrorStack(void) {
```
fingolfin: This code needs to be clang-formatted

src/hpc/threadapi.c
```diff
@@ -2011,8 +2015,14 @@ MigrateObjects(int count, Obj * objects, Region * target, int retype)
                 return 0;
         }
     }
-    for (i = 0; i < count; i++)
-        SET_REGION(objects[i], target);
+    for (i = 0; i < count; i++) {
+        Obj obj = objects[i];
+        switch (TNUM_OBJ(obj)) {
+        case T_PREC:
+          SortPRecRNam(obj, 0);
```
fingolfin: Is this meant to be "future proof", or why isn't this just a simple `if (TNUM_OBJ(obj) == T_PREC)` check?

If it stays as a switch/case, I'd insert a `break`; here

src/hpc/thread.c
rbehrends: Got it.

src/hpc/threadapi.c
rbehrends: This is a relic of an earlier version, I think, and should be fixed.

ssiccha:
I get warnings and an `error: ‘UNSAFE_PTR_BAG’ was not declared in this scope` when I try to compile gap after running unward.

This is my output from `make`.
```
    * new C compiler flags
    * new preprocessor flags
In file included from src/ariths.h:17:0,
                 from src/ariths.c:14:
src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/ariths.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/blister.c => obj/src/blister.lo
In file included from src/bool.h:16:0,
                 from src/blister.h:23,
                 from src/blister.c:73:
src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/blister.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/bool.c => obj/src/bool.lo
In file included from src/bool.h:16:0,
                 from src/bool.c:17:
src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/bool.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/hpc/c_oper1.c => obj/src/hpc/c_oper1.lo
In file included from ./src/ariths.h:17:0,
                 from ./src/compiled.h:17,
                 from src/hpc/c_oper1.c:3:
./src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
./src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
./src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
./src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
./src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
./src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
./src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
./src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/hpc/c_oper1.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/hpc/c_type1.c => obj/src/hpc/c_type1.lo
In file included from ./src/ariths.h:17:0,
                 from ./src/compiled.h:17,
                 from src/hpc/c_type1.c:3:
./src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
./src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
./src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
./src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
./src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
./src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
./src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
./src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/hpc/c_type1.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/calls.c => obj/src/calls.lo
In file included from src/calls.h:49:0,
                 from src/calls.c:37:
src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/calls.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
   C       src/code.c => obj/src/code.lo
In file included from src/code.h:19:0,
                 from src/code.c:16:
src/objects.h: In function ‘UNSAFE_ADDR_OBJ’:
src/objects.h:404:12: warning: implicit declaration of function ‘UNSAFE_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h:404:5: warning: nested extern declaration of ‘UNSAFE_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_PTR_BAG(obj);
     ^
src/objects.h:404:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_PTR_BAG(obj);
            ^
src/objects.h: In function ‘UNSAFE_CONST_ADDR_OBJ’:
src/objects.h:415:12: warning: implicit declaration of function ‘UNSAFE_CONST_PTR_BAG’ [-Wimplicit-function-declaration]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/objects.h:415:5: warning: nested extern declaration of ‘UNSAFE_CONST_PTR_BAG’ [-Wnested-externs]
     return UNSAFE_CONST_PTR_BAG(obj);
     ^
src/objects.h:415:12: warning: return makes pointer from integer without a cast [-Wint-conversion]
     return UNSAFE_CONST_PTR_BAG(obj);
            ^
src/code.c: At top level:
cc1: warning: unrecognized command line option ‘-Wno-implicit-fallthrough’
cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’
    * new C++ compiler flags
   CXX     src/collectors.cc => obj/src/collectors.lo
In file included from src/bool.h:16:0,
                 from src/collectors.cc:31:
src/objects.h: In function ‘UInt*** UNSAFE_ADDR_OBJ(Obj)’:
src/objects.h:404:30: error: ‘UNSAFE_PTR_BAG’ was not declared in this scope
     return UNSAFE_PTR_BAG(obj);
                              ^
src/objects.h: In function ‘UInt** const* UNSAFE_CONST_ADDR_OBJ(Obj)’:
src/objects.h:415:36: error: ‘UNSAFE_CONST_PTR_BAG’ was not declared in this scope
     return UNSAFE_CONST_PTR_BAG(obj);
                                    ^
At global scope:
cc1plus: warning: unrecognized command line option ‘-Wno-cast-function-type’
Makefile.rules:341: recipe for target 'obj/src/collectors.lo' failed
make: *** [obj/src/collectors.lo] Error 1
```

rbehrends:
@ssiccha Are you using the `hpcgap-noward` branch referenced in this PR? Because that's the branch where `UNSAFE_PTR_BAG()` etc. are defined.

ssiccha:
@rbehrends Oops, no I only used `unward`. I'll try it out with the correct branch next. ;)

fingolfin:
I just rebased this.

fingolfin:
@rbehrends it seems that when you last touched this PR you (accidentally?) discarded many of the changes I made. I now restored the commit which automatically runs unward on Travis, and the one which disables most Travis jobs for quicker turn around. 

I've also extracted some code to PR #3382.

fingolfin:
I've rebased this. Unfortunately, the unward tool doesn't seem to work anymore: Running it has no effect. Not even an error message is printed.

rbehrends:
The unward problem occurs because the syntax of inline functions has changed, with the `EXPORT_INLINE` macro being used in lieu of `static inline`. This should be relatively easy to fix.

rbehrends:
I'v updated `unward` and things seem to be in proper working order again.

fingolfin:
This is now part of PR #3502

rbehrends:
Reopening, because we separated these changes from PR [#3502](https://github.com/gap-system/gap/pull/3502).

fingolfin:
The last commit message is "Make HPC-GAP guards work without Ward" -- but HPC-GAP already works without ward, and has done so for months. Perhaps "Change HPC-GAP to use the unward tool" or something like that could be used instead.

src/system.h
```diff
@@ -63,6 +63,12 @@ GAP_STATIC_ASSERT(sizeof(void *) == SIZEOF_VOID_P, "sizeof(void *) is wrong");
 #endif
 #endif
 
+// If we are not running HPC-GAP, disable read and write guards.
+
+#ifndef HPCGAP
+#undef USE_HPC_GUARDS
```
fingolfin: Why is this needed?

src/gasman.h
```diff
@@ -293,18 +298,71 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr) {
 **  the application  must inform {\Gasman}  that it  has changed  the bag, by
 **  calling 'CHANGED_BAG(old)' in the above example (see "CHANGED_BAG").
 */
+#ifdef HPCGAP
+EXPORT_INLINE int ReadCheck(Bag bag)
```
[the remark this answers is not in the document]
fingolfin: I am also generally wondering what the performance impact of this PR is... Like, with something trivial, as running testinstall with and without guards, how does this change the timings (that's a very crude and bad benchmark, of course, but better than nothing)

src/system.h
ChrisJefferson: I really don't like this. If this variable should always be undefined when not using HPCGAP, perhaps put in an #error if HPCGAP is undefined and USE_HPC_GUARDS is defined?

src/hpc/thread.c
```diff
@@ -1309,73 +1309,82 @@ Region * CurrentRegion(void)
...
-    PrintGuardError(buffer, "write", o, file, line, func, expr);
-    ErrorMayQuit("%s", (UInt)buffer, 0L);
-}
+#ifdef DEBUG_GUARDS
+extern GVarDescriptor GUARD_ERROR_STACK;
+#endif
+
+// These are temporary debugging functions.
+
+static Int NumReadErrors = 0;
```
ChrisJefferson: What is "temporary" here? I guess these two variables (as they aren't ever read), and the stuff in DEBUG_GUARDS? Might be useful to be a bit clearer what we expect to keep/lose around here.

etc/ci-prepare.sh
```diff
@@ -40,8 +40,22 @@ then
...
 
-# configure and make GAP
+# configure
 time "$SRCDIR/configure" --enable-Werror $CONFIGFLAGS
+
+if [[ $HPCGAP = yes ]]
+then
+  git clone https://github.com/rbehrends/unward
+  cd unward
+  ./configure CC=gcc CXX=g++ CFLAGS=-O2 CXXFLAGS=-O2
```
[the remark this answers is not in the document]
rbehrends: This is because the build fails due to gcov flags at link time otherwise, so I'm building the tool with fixed compiler/flags settings.

src/gasman.h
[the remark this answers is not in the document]
rbehrends: I removed the `PURE_FUNC` annotations, because it made no noticeable difference on performance.

src/system.h
rbehrends: We can remove this for now. It's left over from when the guards defaulted to enabled in the configure script. If we flip the default, we also have to turn this off, however.

src/hpc/thread.c
rbehrends: I would actually like to leave them around and will document them. I had introduced them to deal with a large bulk of guard errors that occurred after introducing the new mechanism. `NumReadErrors` and `NumWriteErrors` allowed me to track progress on making them, the `#ifdef DEBUG_GUARDS` define allows you to inspect the C stack at the point of error from within GAP without having to fire up a debugger first.

src/gasman.h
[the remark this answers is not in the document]
rbehrends: Oh, and when running `tst/testinstall.g` without having a full GC after each test (to better simulate normal behavior), guard overhead is currently down to a bit less than 10%. There's too much noise to pin the overhead down better, but it's actually an improvement over the original Ward version (where it was around 15%). I suspect that this is largely because modern C compilers do a much better job at optimizing away unnecessary checks than the fairly basic dataflow analysis of Ward did (which, for starters, didn't allow for interprocedural analysis).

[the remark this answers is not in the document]
fingolfin: So did you have a look at the disassembly to see whether the compiled hoisted guard checks out of loops at all?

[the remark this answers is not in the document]
fingolfin: So did you have a look at the disassembly to see whether the compiled hoisted guard checks out of loops at all?

src/system.h
```diff
@@ -63,6 +63,14 @@ GAP_STATIC_ASSERT(sizeof(void *) == SIZEOF_VOID_P, "sizeof(void *) is wrong");
 #endif
 #endif
 
+// If we are not running HPC-GAP, guards should be disabled
+
+#ifndef HPCGAP
+#ifdef USE_HPC_GUARDS
+#error Do not use --enable-guards without --enable-hpcgap.
```
fingolfin: It would be better to check this in `configure.ac`

src/hpc/thread.c
```diff
@@ -1309,73 +1309,91 @@ Region * CurrentRegion(void)
...
-    ImpliedReadGuard(o);
-    if (TLS(DisableGuards))
-        return;
-    SetGVar(&LastInaccessibleGVar, o);
-    PrintGuardError(buffer, "write", o, file, line, func, expr);
-    ErrorMayQuit("%s", (UInt)buffer, 0);
-}
+// Kernel debugging information for guards.
+//
+// When compiled with -DDEBUG_GUARDS, the GAP variable GUARD_ERROR_STACK
```
fingolfin:
```suggestion
// When DEBUG_GUARDS is #defined, the GAP variable GUARD_ERROR_STACK
```

src/hpc/thread.c
```diff
@@ -1309,73 +1309,91 @@ Region * CurrentRegion(void)
...
+// will contain the C stack after an error, allowing us to located where
```
fingolfin:
```suggestion
// is set to a GAP plain list describing the C stack after an error, allowing us to located where
```
Actually, might be better to rephrase: "When DEBUG_GUARDS is #defined and an error occurs, then the GAP variable ..."

src/hpc/thread.c
fingolfin: @ChrisJefferson Reimer improved the comments; better now?

src/hpc/guards.h
```diff
@@ -11,108 +11,72 @@
...
-    ReadGuardError(bag
-#ifdef VERBOSE_GUARDS
-    , file, line, func, expr
-#endif
-    );
-  return bag;
 }
 
-
 static ALWAYS_INLINE Bag ImpliedReadGuard(Bag bag)
```
fingolfin: Can't this be completely removed? If not, what's the purpose of this? A comment should be added to explain it.

src/hpc/guards.h
```diff
@@ -11,108 +11,72 @@
...
-    WriteGuardError(bag
-#ifdef VERBOSE_GUARDS
-    , file, line, func, expr
-#endif
-    );
-  return bag;
 }
 
+
 EXPORT_INLINE Bag ImpliedWriteGuard(Bag bag)
```
fingolfin: Can't this be completely removed? If not, what's the purpose of this? A comment should be added to explain it.

src/lists.c
```diff
@@ -146,7 +146,6 @@ static Obj AttrLENGTH(Obj self, Obj list)
     /* internal list types                                                 */
 #ifdef HPCGAP
     ReadGuard(list);
```
fingolfin: Why is this `ReadGuard` needed, though?

rbehrends: Yes, that can be removed, too, now I look at it. It used to be paired with an `ImpliedWriteGuard()` in order to inform data flow analysis in Ward. On its own, it doesn't serve a purpose anymore (and given how the code has changed, probably hasn't for a while).

src/gasman.h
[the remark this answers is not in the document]
fingolfin: I just checked myself, by looking at `EqRange`, which is rather simple and ideally only would need to `ReadGuard` calls, but naively requires six. And sadly, what I am seeing are the full six calls. Here is disassembly in plain GAP with your PR:
```asm
$ otool -p _EqRange -x -V ./gap
./gap:
(__TEXT,__text) section
_EqRange:
0000000100107940	pushq	%rbp
0000000100107941	movq	%rsp, %rbp
0000000100107944	movq	(%rdi), %rax
0000000100107947	movq	(%rsi), %rcx
000000010010794a	movq	(%rcx), %rdx
000000010010794d	xorq	(%rax), %rdx
0000000100107950	cmpq	$0x3, %rdx
0000000100107954	ja	0x100107977
0000000100107956	movq	0x8(%rcx), %rdx
000000010010795a	xorq	0x8(%rax), %rdx
000000010010795e	cmpq	$0x3, %rdx
0000000100107962	ja	0x10010797b
0000000100107964	movq	0x10(%rcx), %rcx
0000000100107968	xorq	0x10(%rax), %rcx
000000010010796c	xorl	%eax, %eax
000000010010796e	cmpq	$0x4, %rcx
0000000100107972	setb	%al
0000000100107975	popq	%rbp
0000000100107976	retq
0000000100107977	xorl	%eax, %eax
0000000100107979	popq	%rbp
000000010010797a	retq
000000010010797b	xorl	%eax, %eax
000000010010797d	popq	%rbp
000000010010797e	retq
000000010010797f	nop
```

And here is what I get in HPC-GAP with your PR:
```asm
$ otool -p _EqRange -x -V ./gap
./gap:
(__TEXT,__text) section
_EqRange:
0000000100156730	pushq	%rbp
0000000100156731	movq	%rsp, %rbp
0000000100156734	pushq	%r15
0000000100156736	pushq	%r14
0000000100156738	pushq	%rbx
0000000100156739	pushq	%rax
000000010015673a	movq	%rsi, %r14
000000010015673d	movq	%rdi, %r15
0000000100156740	movq	0x8(%rdi), %rax
0000000100156744	testq	%rax, %rax
0000000100156747	je	0x10015676b
0000000100156749	movq	%rbp, %rcx
000000010015674c	andq	$-0x100000, %rcx
0000000100156753	cmpq	%rcx, 0x28(%rax)
0000000100156757	je	0x10015676b
0000000100156759	movslq	(%rcx), %rcx
000000010015675c	cmpb	$0x0, 0x50(%rax,%rcx)
0000000100156761	jne	0x10015676b
0000000100156763	movq	%r15, %rdi
0000000100156766	callq	_HandleReadGuardError
000000010015676b	movq	(%r15), %rax
000000010015676e	movq	(%rax), %rbx
0000000100156771	movq	0x8(%r14), %rax
0000000100156775	testq	%rax, %rax
0000000100156778	je	0x10015679c
000000010015677a	movq	%rbp, %rcx
000000010015677d	andq	$-0x100000, %rcx
0000000100156784	cmpq	%rcx, 0x28(%rax)
0000000100156788	je	0x10015679c
000000010015678a	movslq	(%rcx), %rcx
000000010015678d	cmpb	$0x0, 0x50(%rax,%rcx)
0000000100156792	jne	0x10015679c
0000000100156794	movq	%r14, %rdi
0000000100156797	callq	_HandleReadGuardError
000000010015679c	movq	(%r14), %rax
000000010015679f	xorq	(%rax), %rbx
00000001001567a2	cmpq	$0x3, %rbx
00000001001567a6	ja	0x100156885
00000001001567ac	movq	0x8(%r15), %rax
00000001001567b0	testq	%rax, %rax
00000001001567b3	je	0x1001567d7
00000001001567b5	movq	%rbp, %rcx
00000001001567b8	andq	$-0x100000, %rcx
00000001001567bf	cmpq	%rcx, 0x28(%rax)
00000001001567c3	je	0x1001567d7
00000001001567c5	movslq	(%rcx), %rcx
00000001001567c8	cmpb	$0x0, 0x50(%rax,%rcx)
00000001001567cd	jne	0x1001567d7
00000001001567cf	movq	%r15, %rdi
00000001001567d2	callq	_HandleReadGuardError
00000001001567d7	movq	(%r15), %rax
00000001001567da	movq	0x8(%rax), %rbx
00000001001567de	movq	0x8(%r14), %rax
00000001001567e2	testq	%rax, %rax
00000001001567e5	je	0x100156809
00000001001567e7	movq	%rbp, %rcx
00000001001567ea	andq	$-0x100000, %rcx
00000001001567f1	cmpq	%rcx, 0x28(%rax)
00000001001567f5	je	0x100156809
00000001001567f7	movslq	(%rcx), %rcx
00000001001567fa	cmpb	$0x0, 0x50(%rax,%rcx)
00000001001567ff	jne	0x100156809
0000000100156801	movq	%r14, %rdi
0000000100156804	callq	_HandleReadGuardError
0000000100156809	movq	(%r14), %rax
000000010015680c	xorq	0x8(%rax), %rbx
0000000100156810	cmpq	$0x3, %rbx
0000000100156814	ja	0x100156885
0000000100156816	movq	0x8(%r15), %rax
000000010015681a	testq	%rax, %rax
000000010015681d	je	0x100156841
000000010015681f	movq	%rbp, %rcx
0000000100156822	andq	$-0x100000, %rcx
0000000100156829	cmpq	%rcx, 0x28(%rax)
000000010015682d	je	0x100156841
000000010015682f	movslq	(%rcx), %rcx
0000000100156832	cmpb	$0x0, 0x50(%rax,%rcx)
0000000100156837	jne	0x100156841
0000000100156839	movq	%r15, %rdi
000000010015683c	callq	_HandleReadGuardError
0000000100156841	movq	(%r15), %rax
0000000100156844	movq	0x10(%rax), %rbx
0000000100156848	movq	0x8(%r14), %rax
000000010015684c	testq	%rax, %rax
000000010015684f	je	0x100156873
0000000100156851	movq	%rbp, %rcx
0000000100156854	andq	$-0x100000, %rcx
000000010015685b	cmpq	%rcx, 0x28(%rax)
000000010015685f	je	0x100156873
0000000100156861	movslq	(%rcx), %rcx
0000000100156864	cmpb	$0x0, 0x50(%rax,%rcx)
0000000100156869	jne	0x100156873
000000010015686b	movq	%r14, %rdi
000000010015686e	callq	_HandleReadGuardError
0000000100156873	movq	(%r14), %rax
0000000100156876	xorq	0x10(%rax), %rbx
000000010015687a	xorl	%eax, %eax
000000010015687c	cmpq	$0x4, %rbx
0000000100156880	setb	%al
0000000100156883	jmp	0x100156887
0000000100156885	xorl	%eax, %eax
0000000100156887	addq	$0x8, %rsp
000000010015688b	popq	%rbx
000000010015688c	popq	%r14
000000010015688e	popq	%r15
0000000100156890	popq	%rbp
0000000100156891	retq
0000000100156892	nopw	%cs:(%rax,%rax)
000000010015689c	nopl	(%rax)
```


Of course that particular function could (should?) easily be optimized, in fact it could simply call `memcmp`. I only chose it because it's easy to see with a glance what is going on, even if one is not familiar with assembly.

src/objects.h
```diff
@@ -841,7 +841,13 @@ EXPORT_INLINE Obj TYPE_COMOBJ(Obj obj)
 */
 EXPORT_INLINE void SET_TYPE_COMOBJ(Obj obj, Obj val)
 {
+#ifdef HPCGAP
+    MEMBAR_WRITE();
+    // Only require a read guard
+    ((Obj *) CONST_ADDR_OBJ(obj))[0] = val;
```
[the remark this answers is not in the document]
rbehrends: A minor problem here is that `UNSAFE_ADDR_OBJ()` is generated by unward and not available otherwise. This currently needs an additional `#ifdef USE_HPC_GUARDS`, or we can use `UNSAFE_PTR_BAG()` instead, which returns the same value. But yes, I can definitely change that.

src/objects.h
```diff
@@ -884,7 +890,13 @@ EXPORT_INLINE Obj TYPE_POSOBJ(Obj obj)
 */
 EXPORT_INLINE void SET_TYPE_POSOBJ(Obj obj, Obj val)
 {
+#ifdef HPCGAP
+    MEMBAR_WRITE();
+    // Only require a read guard
+    ((Obj *) CONST_ADDR_OBJ(obj))[0] = val;
```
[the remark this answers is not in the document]
rbehrends: I think the path forward here is indeed to handle `T_DATOBJ` consistently with the rest rather than preserving the existing implementation. I'll look into how to best to clean this up.

src/gasman.h
```diff
@@ -293,18 +298,71 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr) {
...
+extern void HandleWriteGuardError(Bag);
+#endif // HPCGAP
+
 EXPORT_INLINE Bag *PTR_BAG(Bag bag)
 {
     GAP_ASSERT(bag != 0);
+#ifdef USE_HPC_GUARDS
+    if (!WriteCheck(bag))
+      HandleWriteGuardError(bag);
+#endif
```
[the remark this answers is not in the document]
rbehrends: We are dealing with two cases here. One is `TNUM_OBJ()`, which we don't want to burden with additional guard checks, as it's used all the time. Changes are generally either harmless or in the rare cases where they do matter, handled explicitly. The biggest unpleasantness here is `KTNumPlist()`, which has an explicit guard check for this reason.

You are correct that with resizing read-only objects, we run into a problem. I believe the code has assumed that the functions that resize objects (such as `GROW_PLIST()` would fail naturally when changing the length word in the object). However, it looks like sometimes the change happens rather late.

I'd still simply add a write guard to `ResizeBag()`, where it causes minimal overhead, rather than making all accesses to the header more expensive.

rbehrends:
As for the code size problem, I realize now where this is coming from. I had erroneously assumed that when I saw no differences in performance for removing the "pure" attribute trick that the optimizer had managed to handle the cases regardless. However, it seems it doesn't do that, because it has to handle the case where the error handling functions can return and it has to assume that they may change arbitrary memory.

I propose a cleaner solution than using that hack again, however. I'll split the `HandleReadGuardError()` and `HandleWriteGuardError()` in two functions, one of which is actually pure (and marked as such) and contains the extra checks, while the actual error handler is properly marked with the noreturn attribute. This allows the optimizer to treat the checks as common subexpressions or to hoist them out of loops. Importantly, they use the function attributes accurately.

src/gasman.h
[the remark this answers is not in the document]
fingolfin: Just to clarify: the above code is from Apple clang version 11.0.0 (clang-1100.0.33.17) compiled with our default -O2 optimization, and even the plain GAP version obviously is not optimal. @rbehrends pointed out to me that GCC does a better job. Indeed, here is what I got with GCC 9.2.0.

Plain GAP:
```asm
$ otool -p _EqRange -t -V ./gap
./gap:
(__TEXT,__text) section
_EqRange:
00000001000e5040	movq	(%rdi), %rcx
00000001000e5043	xorl	%r8d, %r8d
00000001000e5046	movq	(%rsi), %rsi
00000001000e5049	movq	(%rcx), %rax
00000001000e504c	movq	(%rsi), %rdx
00000001000e504f	sarq	$0x2, %rax
00000001000e5053	sarq	$0x2, %rdx
00000001000e5057	cmpq	%rax, %rdx
00000001000e505a	je	0x1000e5060
00000001000e505c	movq	%r8, %rax
00000001000e505f	retq
00000001000e5060	movq	0x8(%rsi), %rdx
00000001000e5064	movq	0x8(%rcx), %rax
00000001000e5068	sarq	$0x2, %rdx
00000001000e506c	sarq	$0x2, %rax
00000001000e5070	cmpq	%rax, %rdx
00000001000e5073	jne	0x1000e505c
00000001000e5075	movq	0x10(%rsi), %rdx
00000001000e5079	xorl	%r8d, %r8d
00000001000e507c	movq	0x10(%rcx), %rax
00000001000e5080	sarq	$0x2, %rdx
00000001000e5084	sarq	$0x2, %rax
00000001000e5088	cmpq	%rax, %rdx
00000001000e508b	sete	%r8b
00000001000e508f	movq	%r8, %rax
00000001000e5092	retq
00000001000e5093	nopw	%cs:(%rax,%rax)
00000001000e509d	nopl	(%rax)
```

And with HPC-GAP, this (clearly much more complicated, but with only two guard `HandleReadGuardError` calls):
```asm
$ otool -p _EqRange -t -V ./gap
./gap:
(__TEXT,__text) section
_EqRange:
0000000100129d30	pushq	%rbp
0000000100129d31	movq	%rsp, %rbp
0000000100129d34	pushq	%r13
0000000100129d36	movq	%rsi, %r13
0000000100129d39	pushq	%r12
0000000100129d3b	movq	%rdi, %r12
0000000100129d3e	pushq	%rbx
0000000100129d3f	subq	$0x8, %rsp
0000000100129d43	movq	0x8(%rdi), %rax
0000000100129d47	testq	%rax, %rax
0000000100129d4a	je	0x100129d73
0000000100129d4c	movq	%rbp, %rdx
0000000100129d4f	andq	$-0x100000, %rdx
0000000100129d56	cmpq	%rdx, 0x28(%rax)
0000000100129d5a	je	0x100129d73
0000000100129d5c	movslq	(%rdx), %rdx
0000000100129d5f	cmpb	$0x0, 0x50(%rax,%rdx)
0000000100129d64	jne	0x100129d73
0000000100129d66	callq	_ExtendedGuardCheck
0000000100129d6b	testl	%eax, %eax
0000000100129d6d	je	0x100129e15
0000000100129d73	movq	(%r12), %rbx
0000000100129d77	movq	0x8(%r13), %rax
0000000100129d7b	movq	(%rbx), %r12
0000000100129d7e	sarq	$0x2, %r12
0000000100129d82	testq	%rax, %rax
0000000100129d85	je	0x100129dad
0000000100129d87	movq	%rbp, %rdx
0000000100129d8a	andq	$-0x100000, %rdx
0000000100129d91	cmpq	%rdx, 0x28(%rax)
0000000100129d95	je	0x100129dad
0000000100129d97	movslq	(%rdx), %rdx
0000000100129d9a	cmpb	$0x0, 0x50(%rax,%rdx)
0000000100129d9f	jne	0x100129dad
0000000100129da1	movq	%r13, %rdi
0000000100129da4	callq	_ExtendedGuardCheck
0000000100129da9	testl	%eax, %eax
0000000100129dab	je	0x100129e0d
0000000100129dad	movq	(%r13), %rdx
0000000100129db1	xorl	%r8d, %r8d
0000000100129db4	movq	(%rdx), %rax
0000000100129db7	sarq	$0x2, %rax
0000000100129dbb	cmpq	%r12, %rax
0000000100129dbe	je	0x100129dd0
0000000100129dc0	addq	$0x8, %rsp
0000000100129dc4	movq	%r8, %rax
0000000100129dc7	popq	%rbx
0000000100129dc8	popq	%r12
0000000100129dca	popq	%r13
0000000100129dcc	popq	%rbp
0000000100129dcd	retq
0000000100129dce	nop
0000000100129dd0	movq	0x8(%rdx), %rcx
0000000100129dd4	movq	0x8(%rbx), %rax
0000000100129dd8	sarq	$0x2, %rcx
0000000100129ddc	sarq	$0x2, %rax
0000000100129de0	cmpq	%rax, %rcx
0000000100129de3	jne	0x100129dc0
0000000100129de5	movq	0x10(%rbx), %rax
0000000100129de9	xorl	%r8d, %r8d
0000000100129dec	movq	0x10(%rdx), %rdx
0000000100129df0	sarq	$0x2, %rax
0000000100129df4	sarq	$0x2, %rdx
0000000100129df8	cmpq	%rax, %rdx
0000000100129dfb	sete	%r8b
0000000100129dff	addq	$0x8, %rsp
0000000100129e03	popq	%rbx
0000000100129e04	movq	%r8, %rax
0000000100129e07	popq	%r12
0000000100129e09	popq	%r13
0000000100129e0b	popq	%rbp
0000000100129e0c	retq
0000000100129e0d	movq	%r13, %rdi
0000000100129e10	callq	_HandleReadGuardError
0000000100129e15	movq	%r12, %rdi
0000000100129e18	callq	_HandleReadGuardError
0000000100129e1d	nopl	(%rax)
```

src/gasman.h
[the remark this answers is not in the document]
fingolfin: Indeed, I don't want to burden TNUM_BAG, SIZE_BAG, TEST_OBJ_FLAGS with guards; but they *can* be written to, and so at least RetypeBage / ResizeBag and perhaps also the SET/CLEAR_OBJ_FLAGS functions should have guards... I always wondered why they didn't.

src/objects.h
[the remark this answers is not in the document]
fingolfin: I am fine with using UNSAFE_PTR_BAG

src/boehm_gc.c
```diff
@@ -456,6 +456,7 @@ UInt ResizeBag(Bag bag, UInt new_size)
     CollectBags(0, 0);
 #endif
 
+    WriteGuard(bag);
```
fingolfin: Doesn't `RetypeBag` need this as well?

Also, shouldn't it be inside an `#ifdef HPCGAP`?

rbehrends: Sorry, this push went through prematurely. I've been experimenting with various checks on the header functions. `ResizeBag()` seems to not cause any problems; but other functions, such as `SIZE_BAG()`, do, when you add guards to them. This seems to be mostly related to some internal kernel objects, though, for reasons that I don't know yet.

If I add checks to `BAG_HEADER()` directly, this seems to cause an overhead on the order of 10%, possibly a bit less (with clang).

PR #2846, opened 2018-09-20
Large(r) Internally represented FFE's

jdebeule:
The changes in the pull request allow internally represented finite field of order up to 2^24. Changes in the .h and .c (and ffe.g) files are commented. 

Some tests are broken (obviously): those where an just above the old limit (2^16) are chosen and the field of that order is constructed: the results are not expected to be in the internal representation.


markuspf:
The size of `ffdata.c` (26MB) is defninitely a problem now, in my opinion.

We can make pre-computation of finite fields data part of the compilation process (which makes compilation take longer...). Additionally would it be sensible to make the size of finite fields whose elemetns are represented as immediates a compile-time-switch.
We'll have to make sure tests are adapted as well. 

Or we allow pre-computation of some of these values on the fly at runtime...

markuspf:
The pre-computed data was added for HPC-GAP to avoid awkward problems.

hulpke:
As for the running GAP job this seems to be <10MB (which does not worry me). The source file seems to be large mainly because of the inefficiency of storing decimals. Can one improve this (e..g gzip reduces to 6MB)?
Also, it would be helpful to understand why this table (which presumably is created by program) could not be created on the fly when starting GAP. @markuspf pointed to HPCGAP, but there should be no parallelity issue if the data is created before calculations start.

stevelinton:
As an experiment I tried a very simplistic, single-threaded C programme to populate these arrays. It takes about 200ms on my laptop. I'm pretty sure that could be enhanced by a significant factor with a little care. So I can't see any reason not to do this during startup.

markuspf:
In HPC-GAP we could even be daring and do the coputation in a separate thread at startup ;)

stevelinton:
I'd hate to see this change lost. I think it's a really good idea. 

It occurs to me that the main issue we are having comes from the fact that we want to number the finite fields consecutively in a thread-safe way, which, itself, comes from the fact that Martin wanted to do fields up to 2^16 in 32 bit ints and that didn't leave enough bits to just put the field order in the representation of the elements.  (16 (field size)+16 (Zech log) +2 tag bits > 32). We on the other hand don't have this problem in a 64 bit world. (24+24+3 < 64). So we can adopt a much simpler solution.

If we put q itself in the immediate FF representation and then have a thread-safe write-only hash table giving the characteristics and degrees (or something of the kind) then we should avoid all our race condition problems without having to precompute big data structures.

stevelinton:
On reflection, the best solution seems to be to leave the PR more or less exactly as it is, except to 
actually construct `ffdata.c` during the build process.  There should be no problem including a small C utility to construct it, and simply building the file in  `gen` when needed.  There is scope to save some memory by reducing two big arrays from `unsigned long` to `UInt4` saving half their memory.

stevelinton:
Turns out we already have `etc/ffgen.c`. So I think all that needs to be done is to make that a bit more flexible and include it in the build process by default, rather than having it as an optional task that can be performed by developers

jdebeule:
Hi all,

thanks a lot so far for the feedback, suggestions and improvements. Past weeks were too busy for me to work on this. I will make time for this again after the Aachen workshop next week. 

Jan

fingolfin:
@jdebeule note that @stevelinton by now made PR #2997 and you might want to look there, to avoid duplicate work

PR #2851, opened 2018-09-20
Merge PRINT_OR_APPEND_TO and its ..._STREAM variant

ssiccha:
These commits first make the two variants more similar. Then they are merged into a function called
`PRINT_OR_APPEND_TO_FILE_OR_STREAM`.

`PRINT_TO` is changed a bit in a separate commit first:

### kernel: Tidy up PRINT_OR_APPEND_TO

Make PRINT_OR_APPEND_TO more similar to PRINT_OR_APPEND_TO_STREAM:
- Removes the check for whether `filename` is a string. The function
  PRINT_OR_APPEND_TO is called only from the GAP level by PrintTo and
  AppendTo. Both of these already check whether the corresponding
  parameter is a string.

- Instead of first trying to print to stderr and then to stdout, the
  function now uses `Panic`.

- Move the TRY_IF_NO_ERROR to encompass the whole block where the
  printing happens  since PrintFunction may also execute GAP code.
  Also this makes the printing block exactly the same as in
  PRINT_OR_APPEND_TO_STREAM.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -780,17 +780,10 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
     /* try to open the file for output                                     */
     i = append ? OpenAppend( CONST_CSTR_STRING(filename) )
                : OpenOutput( CONST_CSTR_STRING(filename) );
+
     if ( ! i ) {
-        if (strcmp(CONST_CSTR_STRING(filename), "*errout*") == 0) {
-            // When trying to print an error opening *errout* failed,
-            // We exit GAP after trying to print an error.
-            // First try printing an error to stderr
-            int ret = fputs("gap: panic, could not open *errout*!\n", stderr);
-            // If that failed, try printing to stdout
-            if(ret == EOF) {
-                fputs("gap: panic, could not open *errout*!\n", stdout);
-            }
-            SyExit(1);
+        if (strcmp(CSTR_STRING(filename), "*errout*") == 0) {
+            Panic("Failed to open *errout*!");
         }
         ErrorQuit( "%s: cannot open '%g' for output",
                    (Int)funcname, (Int)filename );
@@ -800,29 +793,31 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
     /* print all the arguments, take care of strings and functions         */
     for ( i = 2;  i <= LEN_PLIST(args);  i++ ) {
         arg = ELM_LIST(args,i);
-        if ( IS_PLIST(arg) && 0 < LEN_PLIST(arg) && IsStringConv(arg) ) {
-            PrintString1(arg);
-        }
-        else if ( IS_STRING_REP(arg) ) {
-            PrintString1(arg);
-        }
-        else if ( TNUM_OBJ(arg) == T_FUNCTION ) {
-            PrintFunction( arg );
-        }
-        else {
-            memcpy( readJmpError, STATE(ReadJmpError), sizeof(syJmp_buf) );
 
-            /* if an error occurs stop printing                            */
-            TRY_IF_NO_ERROR {
-                PrintObj( arg );
+        /* if an error occurs stop printing                            */
+        memcpy(readJmpError, STATE(ReadJmpError), sizeof(syJmp_buf));
+        TRY_IF_NO_ERROR
+        {
+            if (IS_PLIST(arg) && 0 < LEN_PLIST(arg) && IsStringConv(arg)) {
+                PrintString1(arg);
             }
-            CATCH_ERROR {
-                CloseOutput();
-                memcpy( STATE(ReadJmpError), readJmpError, sizeof(syJmp_buf) );
-                ReadEvalError();
+            else if (IS_STRING_REP(arg)) {
+                PrintString1(arg);
             }
-            memcpy( STATE(ReadJmpError), readJmpError, sizeof(syJmp_buf) );
+            else if (IS_FUNC(arg)) {
+                PrintFunction(arg);
+            }
+            else {
+                PrintObj(arg);
+            }
         }
+        CATCH_ERROR
+        {
+            CloseOutput();
+            memcpy( STATE(ReadJmpError), readJmpError, sizeof(syJmp_buf) );
+            ReadEvalError();
+        }
+        memcpy(STATE(ReadJmpError), readJmpError, sizeof(syJmp_buf));
     }
 
     /* close the output file again, and return nothing                     */
--- a/src/streams.c
+++ b/src/streams.c
@@ -791,6 +791,11 @@ static Obj PRINT_OR_APPEND_TO_FILE_OR_STREAM(Obj args, int append, int file)
         }
     }
     else {
+        if (CALL_1ARGS(IsOutputStream, destination) != True) {
+            ErrorQuit("%s: <outstream> must be an output stream",
+                      (Int)funcname, 0L);
+            return 0;
+        }
         i = OpenOutputStream(destination);
         if (!i) {
             ErrorQuit("%s: cannot open stream for output", (Int)funcname, 0L);
--- a/src/streams.c
+++ b/src/streams.c
@@ -765,36 +765,44 @@ Obj FuncPrint (
     return 0;
 }
 
-static Obj PRINT_OR_APPEND_TO(Obj args, int append)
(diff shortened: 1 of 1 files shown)


src/streams.c
```diff
@@ -779,27 +779,13 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
 
     /* first entry is the filename                                         */
     filename = ELM_LIST(args,1);
-    while ( ! IsStringConv(filename) ) {
-        filename = ErrorReturnObj(
-            "%s: <filename> must be a string (not a %s)",
-            (Int)funcname, (Int)TNAM_OBJ(filename),
-            "you can replace <filename> via 'return <filename>;'" );
-    }
```
fingolfin: You write that `PrintTo` already checks that `filename` is a string. But in `lib/streams.gd:1044` only `IsString` is checked -- but that is not enough, it really needs to be in `IsStringRep`. So you risk a crash there.

src/streams.c
```diff
@@ -779,27 +779,13 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
...
-            // When trying to print an error opening *errout* failed,
-            // We exit GAP after trying to print an error.
-            // First try printing an error to stderr
-            int ret = fputs("gap: panic, could not open *errout*!\n", stderr);
-            // If that failed, try printing to stdout
-            if(ret == EOF) {
-                fputs("gap: panic, could not open *errout*!\n", stdout);
-            }
-            SyExit(1);
+            Panic("gap: failed to open *errout*!\n");
```
fingolfin: I agree wth using `Panic` here, but please remove the trailing `!\n`, and the leading `gap: `

src/streams.c
```diff
@@ -809,29 +795,29 @@ static Obj PRINT_OR_APPEND_TO(Obj args, int append)
...
-        }
-        else {
-            memcpy( readJmpError, STATE(ReadJmpError), sizeof(syJmp_buf) );
 
-            /* if an error occurs stop printing                            */
-            TRY_IF_NO_ERROR {
-                PrintObj( arg );
+        /* if an error occurs stop printing                            */
+        memcpy( readJmpError, STATE(ReadJmpError), sizeof(syJmp_buf) );
+        TRY_IF_NO_ERROR {
```
fingolfin: It would be nice if the commit message also explained the motivation for moving  `TRY_IF_NO_ERROR`. I assume it is because `PrintFunction` also can execute GAP code, and thus could trigger a `longjmp`?

ssiccha:
@fingolfin I adressed your remarks. From my POV this can be merged once tests run through.

src/streams.c
```diff
@@ -769,104 +769,49 @@ Obj FuncPrint (
...
-            (Int)funcname, (Int)TNAM_OBJ(filename),
-            "you can replace <filename> via 'return <filename>;'" );
+    /* first entry is the file or stream                                   */
+    destination = ELM_LIST(args,1);
+
+    /* try to open the output                                              */
+    if ( file ) {
+        /* the calling functions check that destination is a GAP string,
+        ** but it may not yet be in string representation                  */
+        ConvString(destination);
```
fingolfin: This is better than the code before, but it'll still crash if the first argument is not a string. What is the motivation for removing the error checking? Personally, I always prefer to have kernel functions that can't be used to crash GAP (and if we ever want to even remotely consider security and safety, the first thing we'll have to do is to add argument validation to all kernel functions anyway)

ssiccha: I've updated this to use `IsStringConv` again. Instead of removing the check, the commit msg now reads:
```
- Changes the error handling if the destination filename is not a
  string. In that case, the function prints an error message and
  quits. Previously the user could give another filename.
```

src/streams.c
```diff
@@ -765,99 +765,47 @@ Obj FuncPrint (
...
-            if(ret == EOF) {
-                fputs("gap: panic, could not open *errout*!\n", stdout);
-            }
-            SyExit(1);
-        }
-        ErrorQuit( "%s: cannot open '%g' for output",
-                   (Int)funcname, (Int)filename );
-        return 0;
+    /* try to open the output                                              */
+    if ( file ) {
```
fingolfin: Please run `git clang-format` on your commits.

src/streams.c
```diff
@@ -765,99 +765,47 @@ Obj FuncPrint (
...
+        RequireStringRep(funcname, destination);
+        i = append ? OpenAppend( CONST_CSTR_STRING(destination) )
+                   : OpenOutput( CONST_CSTR_STRING(destination) );
     }
+    else
+        i = OpenOutputStream(destination);
```
fingolfin: You could also add input validation here, i.e., based on this copied from another function:
```C
    if (CALL_1ARGS(IsOutputStream, outstream) != True) {
        ErrorQuit("READ_STREAM_LOOP: <outstream> must be an output stream", 0, 0);
    }
```

src/streams.c
```diff
@@ -765,99 +765,47 @@ Obj FuncPrint (
...
-        else if ( TNUM_OBJ(arg) == T_FUNCTION ) {
-            PrintFunction( arg );
+    /* handle failures                                                     */
+    if ( ! i ) {
+        if ( file ) {
+            if (strcmp(CSTR_STRING(destination), "*errout*") == 0) {
+                Panic("Failed to open *errout*!");
+            }
+            ErrorQuit( "%s: cannot open '%g' for output",
+                       (Int)funcname, (Int)destination );
```
fingolfin: I think I'd find it easier to follow the code if the error handling was move up, to directly after the `Open*` calls, so that you don't have another `if (file) ... else ...` here.

src/streams.c
```diff
@@ -765,99 +765,47 @@ Obj FuncPrint (
...
-        ErrorQuit( "%s: cannot open stream for output", (Int)funcname, 0L );
         return 0;
     }
 
     /* print all the arguments, take care of strings and functions         */
     for ( i = 2;  i <= LEN_PLIST(args);  i++ ) {
         arg = ELM_LIST(args,i);
 
-        /* if an error occurs stop printing                                */
+        /* if an error occurs stop printing                            */
```
fingolfin: Revert this changed line?

src/streams.c
```diff
@@ -866,7 +814,7 @@ static Obj PRINT_OR_APPEND_TO_STREAM(Obj args, int append)
             else if ( IS_STRING_REP(arg) ) {
                 PrintString1(arg);
             }
-            else if ( TNUM_OBJ( arg ) == T_FUNCTION ) {
+            else if ( TNUM_OBJ(arg) == T_FUNCTION ) {
```
fingolfin: If you change the formatting, then go the full way. As it is, I'd either remove this change, or go directly to using `IS_FUNC`, and clang-formatting.

ssiccha:
Arf, I should stop doing stuff in the middle of the night. I must have accidentally introduced the formatting changes when I did an interactive rebase. Weird.

I'll take care of the other things. The test fail was that the `toricvarieties` package could not be loaded, it loads fine on my machine though. I hope restarting the job fixes that.

src/streams.c
ssiccha: I figured out where this diff comes from. I re-used the code from `PRINT_OR_APPEND_TO_FILE`. Now github matches this to the line in `PRINT_OR_APPEND_TO_STREAM` (which is slightly different) since it shows the diff over two commits.

I guess I'll use git clang-format on the commit then.

src/streams.c
ssiccha: Am I correct that this should then also be wrapped in a `TRY_IF_NO_ERROR` block?

src/streams.c
ssiccha: Done

ssiccha:
I've updated the branch. I haven't fully understood the usage of `TRY_IF_NO_ERROR`. Is it correct to also wrap the input validation for `destination` (when it's supposed to be a stream) into a `TRY_IF_NO_ERROR` block?

ssiccha:
One thing that wasn't so clear to me, but I just copied it from older versions of this file. Is it necessary to explicitly have a `return 0;` after each `ErrorQuit` and `ErrorMayQuit`?  IIRC I also saw code where that is not the case.

ChrisJefferson:
You don't need a `return 0;`, as those two functions are marked as not returning. As you say people often have put a return in previously. I've removed a few over the years.

PR #2853, opened 2018-09-20
WIP: Redirect `*errout*`

ssiccha:
With this code I can redirect `*errout*` to any other file. Do you think this is too hacky?

Once I can also redirect `*errout*` to a GAP object, maybe it would be possible to get rid of the `LockCurrentOutput` function and the associated variables (which are used by the test-suite to send the error output into a string).

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -156,7 +156,6 @@ struct IOModuleState {
     //
     TypOutputFile * IgnoreStdoutErrout;
 
-
     // The file identifier of the current input logfile. If it is not 0 the
     // scanner echoes all input from the files '*stdin*' and '*errin*' to
     // this file.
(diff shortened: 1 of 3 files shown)


markuspf:
So I don't like this at all. We've recently *removed* a function like this.

I wrote a patch some time ago that stores a `TypOutputFile` or a `TypInputFile` in a GAP object and allows swapping out input/output with those. The code is here https://github.com/markuspf/gap/tree/inout-rename and here https://github.com/markuspf/gap/tree/inout-rewrite

ssiccha:
I'm happy I got something working after all. But I completely see the point that this is not the nicest of solutions. :)

I'll see if I can have a look at your branches tomorrow.

ssiccha:
If I have the time some day I'll try to create a kernel function `StringExpr` which does the same as `PrintExpr` but stores the result in a string. Then one could write error messages into a single string and then cleanly print that error message into a stream with a single `PrintTo` on the GAP level.

PR #2856, opened 2018-09-21
Make AtExit a stack and handle DirectoryTemporary better

ChrisJefferson:
This is kind of 2 PRs merged together, can split if required.

Firstly, the error message when cleaning up a DirectoryTemporary if it was already deleted wasn't useful, and other temporary directories wouldn't be deleted after the first failing one. This could easily occur if a temporary directory was made, then IO_fork was called.

Secondly, while debugging this I was having some problems which turned out to be caused by the fact the AtExit functions are run in-order. This is against how exit handlers normally work, as it means the first handlers GAP installs (which do things like flush buffers) get run before later, or user-installed handlers.

I pop the list of AtExit handlers, rather than just calling `Reverse` on the list, in case new handlers are installed as GAP is closing.

### DIFF
--- a/lib/session.g
+++ b/lib/session.g
@@ -24,9 +24,10 @@ end);
 BIND_GLOBAL("PROGRAM_CLEAN_UP", function()
     local f;
     if IsBound( GAPInfo.AtExitFuncs ) and IsList( GAPInfo.AtExitFuncs ) then
-        for f in GAPInfo.AtExitFuncs do
+        while not IsEmpty(GAPInfo.AtExitFuncs) do
+            f := Remove(GAPInfo.AtExitFuncs);
             if IsFunction(f) then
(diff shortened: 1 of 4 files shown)


lib/session.g
```diff
@@ -24,9 +24,10 @@ end);
 BIND_GLOBAL("PROGRAM_CLEAN_UP", function()
     local f;
     if IsBound( GAPInfo.AtExitFuncs ) and IsList( GAPInfo.AtExitFuncs ) then
-        for f in GAPInfo.AtExitFuncs do
+        while not IsEmpty(GAPInfo.AtExitFuncs) do
+            f := Remove(GAPInfo.AtExitFuncs);
             if IsFunction(f) then
-                CALL_WITH_CATCH(f,[]);        # really should be CALL_WITH_CATCH here
+                CALL_WITH_CATCH(f,[]);
```
fingolfin: Hehe, nice comment :)

PR #2858, opened 2018-09-21
WIP: Exchange PrintExpr by StringExpr

ssiccha:
Adds a kernel function `StringExpr` which can return "the" string
of an expression. It either creates a new string or appends this
to a given string.

Introduces a new dispatch table called StringExprFuncs which
StringExpr dispatches to.

PrintExpr can be replaced by doing `Pr(StringExpr(..), ..);`.

The PrintExpr calls in PRINT_CURRENT_STATEMENT could then be replaced
by `PrintTo(.., StringExpr(..));`. This in turn makes it possible
to remove the current `OpenOutput(); ... CloseOutput();` hack.

Also makes the function AppendBufToString extern instead of static.

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,0 +1,9 @@
+
+PrintStatFuncs
+    - table, instead use
+    const char *    (* CStringsForStats[256] ) ( Stat stat );
+void InstallPrintStatFunc(Int pos, void (*stat)(Stat))
+
+What are these?
+PrintSeqStat
+OriginalPrintStatFuncsForHook
--- a/src/exprs.c
+++ b/src/exprs.c
@@ -33,6 +33,7 @@
 #include "precord.h"
 #include "range.h"
 #include "records.h"
+#include "scanner.h"
 #include "stringobj.h"
 #include "vars.h"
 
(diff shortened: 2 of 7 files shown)


ssiccha:
I defined `StringTrueExpr` to return "true test" and can now do the following:
```
gap> f := {} -> true;
function(  ) ... end
gap> Print(f);
function (  )
    return true test;
end
```

ssiccha:
Something special would have to be done for `for` statements and the like, to not build up huges strings when printing functions.

ssiccha:
Is this a sensible approach? Do you think this would be useful to have?

ChrisJefferson:
As you mention, you don't want to build up huge strings.

I feel a better option would be a function which took a GAP stream and printed to it. Then you could print to stdout if you wanted to send to the screen, or an OutputTextString if you wanted a string?

stevelinton:
Can the string be significantly bigger than the original source code of the function? If not, the largest GAP file in the library (apart from a data file) is about 200KB, so this string is never really going to be big enough to cause a problem. It's all future garbage collection, but I don't see why you'd be doing this in a loop, and if you were, all you have to do is reuse the same string.  So I wouldn't worry about that aspect of it.

ChrisJefferson:
The problem with gluing strings together is it's easy (if you aren't careful) to hit O(n^2) behaviour. For example, if you implemented String on `+` as something like `Concatenation(String(x), "+", String(y))`, then a statement with 100,000 print statements takes about 20 seconds to build on my machine. Of course, that's (as you say) on the extreme end for GAP, but some generated data files could be the kind of things which can hit this horrible behaviour. Probably we never nest operators stupidly deeply, so as long as the printers for list and series of statements were written without hitting O(n^2) behaviour, we'd be OK in practice.

ssiccha:
Making `PrintExprTo` versions of `PrintExpr` with an additional stream argument also would make it possible to do something like in #2824 on the kernel level.

ssiccha:
I'm gonna open a proper version of this PR, so I'll close this now. I'll see whether I can avoid the `n^2` behaviour.

PR #2859, opened 2018-09-21
Another attempt to fix coverage for io.c and combinat.gi

fingolfin:


fingolfin:
I think ` --coverage` sneaks back in via the GAP_CFLAGS *sigh*

PR #2865, opened 2018-09-25
Fix sorting bug in RECALCULATE_ALL_METHOD_RANKS

fingolfin:
`RECALCULATE_ALL_METHOD_RANKS` failed to properly sort if there were more than two methods, and the 3rd or any later method needed to be moved to the first slot -- it only moved it to the second slot in that case.

Also add `CHECK_ALL_METHOD_RANKS` helper to verify the method ordering is correct.

### DIFF
--- a/lib/oper.g
+++ b/lib/oper.g
@@ -1933,6 +1933,33 @@ BIND_GLOBAL("MethodsOperation", function(oper, nargs)
     return result;
 end );
 
+
+#############################################################################
+##
+#F  CHECK_ALL_METHOD_RANKS
+##
+##  Debugging helper which checks that all methods are sorted correctly
+##
+BIND_GLOBAL( "CHECK_ALL_METHOD_RANKS", function()
+    local  oper, n, meths, i, result;
+
+    result := true;
+    for oper in OPERATIONS do
+        for n in [0..6] do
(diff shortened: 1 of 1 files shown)


lib/oper.g
```diff
@@ -2021,6 +2048,8 @@ BIND_GLOBAL( "RECALCULATE_ALL_METHOD_RANKS", function()
             fi;
         od;
     od;
+
+    Assert(2, CHECK_ALL_METHOD_RANKS());
```
fingolfin: BTW, with all packages loaded, `CHECK_ALL_METHOD_RANKS` takes about 60 milliseconds, while `RECALCULATE_ALL_METHOD_RANKS` takes 600 ms; so we could just always call `CHECK_ALL_METHOD_RANKS` here.

Also note that `RECALCULATE_ALL_METHOD_RANKS` spends 90% of its time in `RankFilter`, or rather, in `WITH_HIDDEN_IMPS_FLAGS`.

ChrisJefferson: I think leaving it in an assert is fine. It does occur to me (not for now) there isn't a way I know of to run Asserts at high level during GAP's loading... which might be useful for testing.

PR #2866, opened 2018-09-25
Fix IsNilpotentGroup: infinite does not imply "not p-group"

fingolfin:
This should be backported to 4.10.

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -414,7 +414,7 @@ InstallMethod( IsNilpotentGroup,
     elif s = 1 then
         SetIsPGroup( G, true );
         return true;
-    else
+    elif s <> infinity then
         SetIsPGroup( G, false );
     fi;
     TryNextMethod();


fingolfin:
@stevelinton does this seem sensible to you?

fingolfin:
Backported to 4.10 in 738817050f4675e9c0f68279c3e1eb39f1480221

@hulpke It is "clear" in so far as some time ago, we (well, mostly we) audited the library and all deposited packages for uses of `IsPGroup`, and added explicit `IsFinite` checks when relevant. But I neglected to check for code which explicitly sets `IsPGroup` to `false`. When creating this PR, I check for other places doing it, but the on modified in this PR was the only I found. So, to the best of knowledge and ability, no *public* code is left that assumes `IsPGroup` implies finite; but of course I still may have missed some, and then there is non-public code... But on the upside, I am not yet aware of any code in GAP providing infinite p-groups, so until that happens, no problems should arise, giving us some more time for the transition.

hulpke:
@fingolfin Excellent. Thanks.

PR #2867, opened 2018-09-25
Fix more test failures due to method re-ordering

fingolfin:
This PR fixes more test failures due to method re-ordering, see issue #2818.

These are the fixes which I hope will be uncontroversial. I have a related upcoming PR which may not be.

### DIFF
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -315,5 +315,6 @@ end)();
 # This method must rank below Random(SomeRandomSource, IsList)
 # for any random source SomeRandomSource, to avoid an infinite loop.
 InstallMethodWithRandomSource( Random, "for a random source and a (finite) collection",
-    [ IsRandomSource, IsCollection and IsFinite ], -8,
+    [ IsRandomSource, IsCollection and IsFinite ],
+    {} -> -RankFilter(IsCollection and IsFinite),
     {rs, C} -> RandomList(rs, Enumerator( C ) ) );
--- a/tst/testinstall/mgmring.tst
+++ b/tst/testinstall/mgmring.tst
@@ -18,8 +18,8 @@ gap> IsGroupRing( rm );
 true
 gap> centre:= Centre( rm );
 <algebra-with-one of dimension 3 over GF(3)>
-gap> GeneratorsOfAlgebra( centre );
-[ (Z(3)^0)*(), (Z(3)^0)*(), (Z(3)^0)*(2,3)+(Z(3)^0)*(1,2)+(Z(3)^0)*(1,3), 
+gap> AsSet( GeneratorsOfAlgebra( centre ) );
+[ (Z(3)^0)*(), (Z(3)^0)*(2,3)+(Z(3)^0)*(1,2)+(Z(3)^0)*(1,3), 
   (Z(3)^0)*(1,2,3)+(Z(3)^0)*(1,3,2) ]
 gap> membrm:= Embedding( m, rm );;
 gap> img:= Image( membrm, (1,2) );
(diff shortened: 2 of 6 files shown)


lib/random.gi
```diff
@@ -315,5 +315,6 @@ end)();
 # This method must rank below Random(SomeRandomSource, IsList)
 # for any random source SomeRandomSource, to avoid an infinite loop.
 InstallMethodWithRandomSource( Random, "for a random source and a (finite) collection",
-    [ IsRandomSource, IsCollection and IsFinite ], -8,
+    [ IsRandomSource, IsCollection and IsFinite ],
+    {} -> -RankFilter(IsCollection and IsFinite),
```
fingolfin: IIRC, then me, @ChrisJefferson and others discussed this particular rank adjustment at length. The current static adjustment by -8 is not enough with all packages loaded. I think it's appropriate to rank this down to where the rank of the second argument is effectively zero. When GAP starts, this amounts to `-4` (so the rank actually of this method is actually a bit higher than it was before). Once all packages are loaded, it is `-126`... (but once the `automgrp` package is fixed to not have an implication `IsFinite => IsAmenable`, it'll be down to -6).

Alternatively, we could also use `{} -> -RankFilter(IsCollection and IsFinite)-RankFilter(IsRandomSource)`, where `RankFilter(IsRandomSource) = 2`, to effectively get the rank of this method down to 0, but that seems like overkill.

tst/testinstall/opers/FrattiniSubgroup.tst
```diff
@@ -40,8 +40,8 @@ gap> FrattiniSubgroup(CyclicGroup(IsPermGroup, 9));
 Group([ (1,4,7)(2,5,8)(3,6,9) ])
 gap> FrattiniSubgroup(CyclicGroup(IsPcGroup, 3));
 Group([  ])
-gap> FrattiniSubgroup(CyclicGroup(IsPcGroup, 9));
-Group([ f2 ])
+gap> AsSet( FrattiniSubgroup(CyclicGroup(IsPcGroup, 9)) );
```
fingolfin: Two different methods are called here depending on which packages (and thus implications are loaded), but both really are fine.

tst/testinstall/mgmring.tst
```diff
@@ -18,8 +18,8 @@ gap> IsGroupRing( rm );
 true
 gap> centre:= Centre( rm );
 <algebra-with-one of dimension 3 over GF(3)>
-gap> GeneratorsOfAlgebra( centre );
-[ (Z(3)^0)*(), (Z(3)^0)*(), (Z(3)^0)*(2,3)+(Z(3)^0)*(1,2)+(Z(3)^0)*(1,3), 
+gap> AsSet( GeneratorsOfAlgebra( centre ) );
+[ (Z(3)^0)*(), (Z(3)^0)*(2,3)+(Z(3)^0)*(1,2)+(Z(3)^0)*(1,3),
```
fingolfin: Two different methods are called here depending on which packages (and thus implications are loaded), but both really are fine.

tst/testinstall/pgroups.tst
```diff
@@ -101,16 +101,12 @@ gap> K := Group((1,3),(1,2,3,4),(5,6,7));
...
-gap> HasIsPGroup(K);
-true
 gap> IsPGroup(K);
 false
 gap> L := Group((2,4), (1,2,3,4));
 Group([ (2,4), (1,2,3,4) ])
 gap> IsNilpotentGroup(L);
 true
-gap> HasIsPGroup(L) and HasPrimePGroup(L);
-true
```
fingolfin: These two `HasIsPGroup` are meant to check that the "right" method for `IsNilpotentGroup` is called (and that methods has a static rank adjustment of +20)... but that seems pointless, so let's just make the tests less specific.

PR #2873, opened 2018-09-26
Move GAP kernel version to configure script

ChrisJefferson:
This lifts the GAP kernel version to the configure script, so it can be used during configure and creating the Makefile. The kernel version is used in two new places:

1) We put it in sysinfo.gap, so it's easier for packages to get it
2) We put it in the build directory (so mine is now `x86_64-pc-linux-gnu-default64-abi4`, note the `abi4`)

2) is useful when using thee same package directory with multiple versions of GAP (which will be particularly useful in a world where more people use the new package manager...)

1) is useful for languages (in my case Rust) where we want to create packages and need to know the kernel version, but don't want to have to dig through C header files to find it.

There are two broad things I'm interested in:

A) Does putting something like the major kernel version in the gaparch seem like a good idea, to make it easier for GAPs to co-exist?

B) Does this seem like a sensible way to do it?

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -674,6 +674,9 @@ sysinfo.gap: config.status $(srcdir)/Makefile.rules cnf/GAP-CFLAGS cnf/GAP-CPPFL
 	@echo "GAP_HPCGAP=$(HPCGAP)" >> $@
 	@echo "GAP_ADDGUARDS2=$(ADDGUARDS2)" >> $@
 	@echo "" >> $@
+	@echo "GAP_KERNEL_MAJOR_VERSION=$(GAP_KERNEL_MAJOR_VERSION)" >> $@
+	@echo "GAP_KERNEL_MINOR_VERSION=$(GAP_KERNEL_MINOR_VERSION)" >> $@
+	@echo "" >> $@
 	@echo "GAP_BIN_DIR=\"$(abs_builddir)\"" >> $@
 	@echo "GAP_LIB_DIR=\"$(abs_srcdir)\"" >> $@
 	@echo "" >> $@
--- a/configure.ac
+++ b/configure.ac
@@ -575,13 +575,25 @@ AS_IF([test "x$with_gc" = xboehm],
   ]
 )
 
+dnl
+dnl Define kernel version
+dnl
+
+
+gap_kernel_major_version=5
+gap_kernel_minor_version=0
+AC_SUBST([gap_kernel_major_version])
+AC_SUBST([gap_kernel_minor_version])
+
+AC_DEFINE_UNQUOTED([GAP_KERNEL_MAJOR_VERSION], $gap_kernel_major_version, [The major version of the kernel ABI])
+AC_DEFINE_UNQUOTED([GAP_KERNEL_MINOR_VERSION], $gap_kernel_minor_version, [The minor version of the kernel ABI])
 
 dnl
 dnl User setting: Compatibility mode (on by default)
 dnl
 AS_IF([test "x$enable_hpcgap" = xyes],
-  [GAPARCH="$host-hpcgap${ABI}"],
-  [GAPARCH="$host-default${ABI}"])
+  [GAPARCH="$host-hpcgap${ABI}-kv${gap_kernel_major_version}"],
+  [GAPARCH="$host-default${ABI}-kv${gap_kernel_major_version}"])
 AS_IF([test "x$ARCHEXT" != "x"],
   [GAPARCH="$GAPARCH-$ARCHEXT"])
 AS_IF([test "x$ARCH" != "x"],
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -28,6 +28,10 @@ GC_SOURCES = @GC_SOURCES@
 COMPAT_MODE = @COMPAT_MODE@
 GAPARCH = @GAPARCH@
 
+# GAP kernel version
+GAP_KERNEL_MINOR_VERSION = @gap_kernel_minor_version@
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -6,6 +6,8 @@
 /cnf/GAP-LDFLAGS
 /cnf/GAP-LIBS
 /cnf/GAP-OBJS
+/cnf/GAP-KERNEL_MAJOR_VERSION
+/cnf/GAP-KERNEL_MINOR_VERSION
```
fingolfin: What is the purpose of these two files?

ChrisJefferson: I was following the pattern of existing lines in sysinfo.gap -- we make a file which contains the variable, use some little trick to make sure we only touch the file if it changes, and then make sysinfo.gap depend on those new files.

fingolfin: That's not what those are for: `sysinfo.gap` already depends on `config.status`, so if the kernel version is changed in `configure.ac`, we'll automatically regenerate `configure`, then re-run it, then `config.status` is called to regenerate `sysinfo.gap`

Those extra rules for CFLAGS, OBJS etc. exist because they can be changed by a change to the `Makefile` alone: e.g. adding or remove a source file can change `OBJS`, and the user might do `make CFLAGS=-O3` to recompile GAP with different compiler flags.

But we certainly don't want to allow users to do a `make KERNEL_MAJOR_VERSION=42`.

Hence, unless I am missing something, I think these files and all the related code should be removed.

configure.ac
```diff
@@ -575,13 +575,25 @@ AS_IF([test "x$with_gc" = xboehm],
...
+dnl
+
+
+gap_kernel_major_version=4
+gap_kernel_minor_version=0
+AC_SUBST([gap_kernel_major_version])
+AC_SUBST([gap_kernel_minor_version])
+
+AC_DEFINE_UNQUOTED([GAP_KERNEL_MAJOR_VERSION], $gap_kernel_major_version, [The major version of the kernel ABI])
+AC_DEFINE_UNQUOTED([GAP_KERNEL_MINOR_VERSION], $gap_kernel_minor_version, [The minor version of the kernel ABI])
```
fingolfin: If we do this, we might as well start using this to version the `libgap` shared library, by passing.Not at all saying this should be done in this PR (it's a bit tricky to get semantically right, though not much if one takes the time to read the explanations on what the arguments mean), just wanted to mention it -- we could file an issue for that idea, then somebody (possibly me) can one day look into it.

configure.ac
```diff
@@ -575,13 +575,25 @@ AS_IF([test "x$with_gc" = xboehm],
...
 
 dnl
 dnl User setting: Compatibility mode (on by default)
 dnl
 AS_IF([test "x$enable_hpcgap" = xyes],
-  [GAPARCH="$host-hpcgap${ABI}"],
-  [GAPARCH="$host-default${ABI}"])
+  [GAPARCH="$host-hpcgap${ABI}-abi${gap_kernel_major_version}"],
+  [GAPARCH="$host-default${ABI}-abi${gap_kernel_major_version}"])
```
fingolfin: That seems like a good idea to me. Although the "ABI" vs. "abi" bit are a bit confusing, and I worry that down the road it'll lead to confusion in discussions. Perhaps we should rename `ABI` to `ARCHABI`. And/or change `-abi` to `-kernel`, `-kv`, etc.?

Makefile.rules
```diff
@@ -665,7 +665,7 @@ ifeq ($(COMPAT_MODE),yes)
 
 # regenerate sysinfo.gap if necessary
 all: sysinfo.gap
-sysinfo.gap: config.status $(srcdir)/Makefile.rules cnf/GAP-CFLAGS cnf/GAP-CPPFLAGS cnf/GAP-LDFLAGS cnf/GAP-LIBS
+sysinfo.gap: config.status $(srcdir)/Makefile.rules cnf/GAP-CFLAGS cnf/GAP-CPPFLAGS cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-KERNEL_MAJOR_VERSION cnf/GAP-KERNEL_MINOR_VERSION
```
fingolfin: This change is unnecessary (see above).

Makefile.rules
```diff
@@ -1104,6 +1107,25 @@ cnf/GAP-OBJS: FORCE
...
+
+ESCAPED_KERNEL_MINOR_VERSION = $(subst ','\'',$(GAP_KERNEL_MINOR_VERSION))
+
+cnf/GAP-KERNEL_MINOR_VERSION: FORCE
+	@mkdir -p cnf
+	@FLAGS='$(ESCAPED_KERNEL_MINOR_VERSION)'; \
+	    if test x"$$FLAGS" != x"`cat cnf/GAP-KERNEL_MINOR_VERSION 2>/dev/null`" ; then \
+		echo >&2 "    * new kernel minor version"; \
+		echo "$$FLAGS" >cnf/GAP-KERNEL_MINOR_VERSION; \
+            fi
```
fingolfin: This whole block should go (see above)

fingolfin:
Oh, and overall, I think this PR is a very good idea, thank you!

.gitignore
ChrisJefferson: Thanks for explaining, will fix

configure.ac
ChrisJefferson: I like `-kv`. I didn't want something that looks too much like "version", because then people might expect it to be "4.9.2" or something like that

fingolfin: Just to clarify: that's an optional remark, i.e., it could be changed now, or later, or never

PR #2876, opened 2018-09-27
Make IsomorphismTransformationSemigroup return IdentityMapping for a transformation semigroup

wilfwilson:
I think it is appropriate that `IsomorphismTransformationSemigroup` for a transformation semigroup should simply return the `IdentityMapping`, and indeed this should always be the case, and so I've increased the rank by `SUM_FLAGS`. This is analogous to the situation for perm groups:
https://github.com/gap-system/gap/blob/d0fb0f5e5482750010ed337315d501eb6e5a32e2/lib/ghomperm.gi#L1928-L1933

### DIFF
--- a/lib/semitran.gi
+++ b/lib/semitran.gi
@@ -383,9 +383,8 @@ end);
 InstallMethod(IsomorphismTransformationSemigroup, 
 "for a transformation semigroup", 
 [IsTransformationSemigroup], 
-function(S)
-  return MagmaIsomorphismByFunctionsNC(S, S, IdFunc, IdFunc);
-end);
+SUM_FLAGS,
+IdentityMapping);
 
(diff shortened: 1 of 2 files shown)


lib/semitran.gi
```diff
@@ -383,9 +383,8 @@ end);
 InstallMethod(IsomorphismTransformationSemigroup, 
 "for a transformation semigroup", 
 [IsTransformationSemigroup], 
-function(S)
-  return MagmaIsomorphismByFunctionsNC(S, S, IdFunc, IdFunc);
-end);
```
markuspf: You sure you don't mean `IdentityMapping(S)`?

wilfwilson: These are the lines that I'm deleting, I'm not sure what you mean.

PR #2877, opened 2018-09-27
Improve IsPackageLoaded

### LINKED ISSUE #2862: Change `IsPackageLoaded` to not run the `inforec.AvailabilityTest` function
Right now, you get this if e.g. semigroups is not loaded:
```
gap> IsPackageLoaded("semigroups");
#I  Semigroups: the kernel module is not compiled, the package cannot be loaded.
false
```
That Info message is printed by semigroups `inforec.AvailabilityTest` function, and is undesirable here; it also is simply not necessary to call this function if the package has been determined to not be loaded at all.

This goes away if `AvailabilityTest` is not called.

fingolfin:
First off, it now aborts immediately if the package is not loaded, and
doesn't call the package's `AvailabilityTest` function (fixes #2862).

This was achieved by not using `TestPackageAvailability` anymore, but rather
`IsPackageMarkedForLoading`. The main difference to the latter now is this:
We now record whether a package fully completed loading (in particular, we
record whether its `init.g` and `read.g` executed without an error). This
information is then used by `IsPackageLoaded` after
`IsPackageMarkedForLoading` returned `true`, to determine the final result.

I would suggest backporting this to GAP 4.10.

### DIFF
--- a/lib/package.gd
+++ b/lib/package.gd
@@ -43,9 +43,10 @@
 ##  <Ref Var="GAPInfo.PackagesLoaded"/> is a mutable record,
 ##  its component names are the names of those &GAP; packages that are
 ##  already loaded.
-##  The component for each package is a list of length three, the entries
+##  The component for each package is a list of length four, the entries
 ##  being the path to the &GAP; root directory that contains the package,
-##  the package version, and the package name.
+##  the package version, the package name, and a boolean indicating whether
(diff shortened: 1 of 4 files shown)


fingolfin:
Backported to stable-4.10 as c94a3641a80a8a1ad2b1759763aa84d10d9b3763

PR #2880, opened 2018-09-28
Add AppVeyor build status badge

fingolfin:


fingolfin:
I don't think we need to run tests for this... so I am aborting them now.

PR #2881, opened 2018-09-28
buildsys: fix C -> C++ transition for out-of-tree builds

fingolfin:
This ensures that an out-of-tree build which was last compiled before the C++ changes were merged will recompile fine after this commit, with the C++ changes in.

Would be good to merge this ASAP, to ensure people have a smooth transition with their existing GAP builds.


ChrisJefferson:
travis seems backed up, and I have manually downloaded this and checked it works, so merging.

PR #2882, opened 2018-09-28
Add PCG32 random number generator

ChrisJefferson:
This random generator produces very high quality random numbers.
It also has the major advantage that it's initialisation and
state are both very small and simple, meaning it is very low cost
to create many instances of this generator.

I've been using this code in many of my own (non-GAP) projects, and found it works very well. It is extremely small, and I hope to use this in cases like #2878 , to create a quick random source in place, so we don't effect, or are not effected by, the global random source.

### DIFF
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -31,7 +31,7 @@ end);
 
 # Generic fallback methods, such that it is sufficient to install Random for 
 # lists or for pairs of integers.
-InstallMethod(Random, [IsRandomSource, IsInt, IsInt], function(rs, a, b)
+BindGlobal("FALLBACK_GENERATE_RANDOM_INT", function(rs, a, b)
   local  d, x, r, y;
   d := b - a;
   if d < 0  then
@@ -53,6 +53,8 @@ InstallMethod(Random, [IsRandomSource, IsInt, IsInt], function(rs, a, b)
     fi;
   fi;
 end);
+
+InstallMethod(Random, [IsRandomSource, IsInt, IsInt], FALLBACK_GENERATE_RANDOM_INT);
 
 InstallMethod(Random, [IsRandomSource, IsList], function(rs, list)
   return list[Random(rs, 1, Length(list))];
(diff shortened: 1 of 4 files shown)


lib/random.gi
```diff
@@ -54,6 +54,8 @@ InstallMethod(Random, [IsRandomSource, IsInt, IsInt], function(rs, a, b)
   fi;
 end);
 
+InstallMethod(Random, [IsRandomSource, IsInt, IsInt], FALLBACK_GENERATE_RANDOM_INT);
```
ChrisJefferson: I seperate this out into a named function, so I can choose to call it to help generate large integers.

src/intfuncs.c
```diff
@@ -26,6 +27,159 @@
...
+** * * * * * * * "PCG" random numbers  * * * * * * * * * * * * *
+**
+** This next function implements the stepping function of the 'PCG'
+** random number library. This library produces high quality random numbers
+** which pass many statistical tests. They also have a very small state which
+** can be quickly initalised
+**
+** The following three functions are under the following original license
+** *Really* minimal PCG32 code / (c) 2014 M.E. O'Neill / pcg-random.org
+** Licensed under Apache License 2.0 (NO WARRANTY, etc. see website)
```
fingolfin: I really like http://www.pcg-random.org , saw it before. On a technical level, I have no issues with this PR.

However, I note that the Apache License 2.0 is deemed incompatible with GPLv2 by the FSF and others. Of course GAP is licenses under "GPL 2 or later", and indeed, the Apache License 2.0 is compatible with GPLv3. But this change then kinda would make GAP "GPLv3" only "through the backdoor", which seems unfortunate. It'd be different if we changed the license explicitly. But then there are also reasons to dislike GPLv3 compared to GPLv2... *sigh*

Anyway, of course we could also ignore this. But I am pretty sure there are people (e.g. on Debian) who'd *really* dislike this.

ChrisJefferson: I'm going to email the author and see if they would be happy to dual license the code GPL v2. Have marked do not merge

fingolfin: Any news on this?

fingolfin:
@ChrisJefferson did you resolve the copyright / licensing issues? If not, is there any point in keeping this PR open?

ChrisJefferson:
I never had any response. I will keep the branch privately just in case I ever get a reply.

PR #2883, opened 2018-09-28
kernel: more CONSTification (useful for HPC-GAP / unward, and general code hygiene)

fingolfin:
This contains parts of PR #2845.

### DIFF
--- a/src/costab.c
+++ b/src/costab.c
@@ -2088,8 +2088,8 @@ UInt RelatorScan (
 {
     UInt  m,i,p,a,j;
     UInt  pa=0,pb=0;
-    UInt * rp;
-    rp=(UInt*)ADDR_OBJ(r);
+    const UInt * rp;
+    rp=(const UInt*)CONST_ADDR_OBJ(r);
     m=rp[1]; /* length is in position 1 */
     i=2;
     p=di;
--- a/src/objfgelm.cc
+++ b/src/objfgelm.cc
@@ -1157,7 +1157,7 @@ Obj FuncMULT_WOR_LETTREP (
 {
   UInt l,m,i,j,newlen,as,bs,ae,be;
   Obj n;
-  Obj *p;
+  const Obj *p;
   Obj *q;
 
   /* short check */
@@ -1243,16 +1243,14 @@ Obj FuncMULT_WOR_LETTREP (
   q++;
   j=as;
   /* a[as] position */
-  /* there must be a better way for giving this address ... */
-  p=(Obj*) &(ADDR_OBJ(a)[as]);
+  p=CONST_ADDR_OBJ(a) + as;
   while (j<=ae) {
     *q++=*p++;
     j++;
   }
   j=bs;
   /* b[bs] position */
-  /* there must be a better way for giving this address ... */
-  p=(Obj*) &(ADDR_OBJ(b)[bs]);
+  p=CONST_ADDR_OBJ(b) + bs;
   while (j<=be) {
     *q++=*p++;
     j++;
--- a/src/vecgf2.c
+++ b/src/vecgf2.c
@@ -107,7 +107,7 @@ Obj GF2Zero;
 
 static inline void AddGF2VecToGF2Vec(
   UInt *	ptS,
-  UInt *	ptV,
+  const UInt *	ptV,
   UInt		len)
 {
   register UInt ct;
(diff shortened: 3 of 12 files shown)


src/vecgf2.c
```diff
@@ -2926,10 +2932,6 @@ Obj FuncTRANSPOSED_GF2MAT( Obj self, Obj mat)
   /* create new matrix */
   for (i = 1; i <= w; i++) {
     NEW_GF2VEC( row,  typ, l );
-
-    ptr=BLOCKS_GF2VEC(row);
-    for (n=1;n<=nrb;n++){*ptr=0;}
-
```
fingolfin: This loops was buggy, repeatedly zeroing the very same memory cell (ptr is never incremented). But things worked anyway, because `NewBag` and hence `NEW_GF2VEC` gives us zero initialised allocations. So, I just removed it.

ChrisJefferson: agreed

src/vecgf2.c
```diff
@@ -3482,16 +3474,15 @@ UInt CosetLeadersInnerGF2( Obj veclis,
...
 	  BLOCK_ELM_GF2VEC(v, i) |= MASK_POS_GF2VEC(i);
 
-	  sy = revertbits(BLOCKS_GF2VEC(w)[0], lenw);
+	  sy = revertbits(CONST_BLOCKS_GF2VEC(w)[0], lenw);
 	  if ((Obj) 0 == ELM_PLIST(leaders,sy+1))
 	    {
 	      NEW_GF2VEC(vc, TYPE_LIST_GF2VEC_IMM, len);
-	      for (j = 0; j < NUMBER_BLOCKS_GF2VEC(v); j++)
-		BLOCKS_GF2VEC(vc)[j] = BLOCKS_GF2VEC(v)[j];
+	      memcpy(BLOCKS_GF2VEC(vc), CONST_BLOCKS_GF2VEC(v), NUMBER_BLOCKS_GF2VEC(v) * sizeof(UInt));
```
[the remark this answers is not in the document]
ChrisJefferson: verified correct

src/vecgf2.c
```diff
@@ -3237,16 +3232,13 @@ Int DistVecClosVec(
...
-	    vec = BLOCKS_GF2VEC(ovec);
-	    sum = BLOCKS_GF2VEC(osum);
+	    vec = CONST_BLOCKS_GF2VEC(ovec);
 	    SET_ELM_PLIST(d,di+1,cnt);
 	    CHANGED_BAG(d);
-	    chg=1;
 	  }
       }
-    AddGF2VecToGF2Vec(sum,BLOCKS_GF2VEC(ELM_PLIST(vp,i+1)),len);
+    AddGF2VecToGF2Vec(BLOCKS_GF2VEC(osum),CONST_BLOCKS_GF2VEC(ELM_PLIST(vp,i+1)),len);
```
ChrisJefferson: Just checking, what's happening here?

fingolfin: So, there is `Obj osum` (I guess the `o` in `osum` indicates that it is the "object"), and `UInt * sum` was set via `sum = BLOCKS_GF2VEC(osum);` -- and then there was logic to reset `sum` if `chg` is non-zero. And `chg` was tracked through recursion, presumably to track whether there *might* have been a GC. I decided that all this complexity is not worth the potential minor saving (I am not even sure it exists, because the code to track the `chg` state also takes up resources).

If one is concerned about optimizing this code, the first step should instead be to replace `SumInt` by  a suitably application of `ARE_INTOBJS` and `SUM_INTOBJS`.

src/costab.c
```diff
@@ -2088,8 +2088,8 @@ UInt RelatorScan (
 {
     UInt  m,i,p,a,j;
     UInt  pa=0,pb=0;
-    UInt * rp;
-    rp=(UInt*)ADDR_OBJ(r);
```
ChrisJefferson: you could pop this on one line (not a requirement of course)

src/objfgelm.cc
```diff
@@ -1243,16 +1243,14 @@ Obj FuncMULT_WOR_LETTREP (
   q++;
   j=as;
   /* a[as] position */
-  /* there must be a better way for giving this address ... */
-  p=(Obj*) &(ADDR_OBJ(a)[as]);
```
ChrisJefferson: There was a better way of getting this address :)

ChrisJefferson:
I popped through this again and had 1 very tiny comment (partly just to prove I read it all :) ), but am happy for it to be merged.

PR #2887, opened 2018-09-30
Remove code duplication between UNITE_BLIST_LIST and BLIST_LIST

ChrisJefferson:
UNITE_BLIST_LIST and BLIST_LIST were almost identical, this code unifies them.

I would suggest reading the two commits, as after the first one the functions are almost identical.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1068,21 +1068,14 @@ Obj FuncSIZE_BLIST (
 **  'BlistList' is most effective if <list> is a set, but can be used with an
 **  arbitrary list that has no holes.
 */
+
+Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub);
+
 Obj FuncBLIST_LIST (
     Obj                 self,
     Obj                 list,
     Obj                 sub )
 {
-    Obj                 blist;          /* boolean list, result            */
-    UInt  *             ptrBlist;       /* pointer to the boolean list     */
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
There is hypothetically some tiny slowdowns that could occur, as we will now read existing blocks of zeros and "or" them with the newly created blocks, but this is a tiny change which only occurs in the final case, and I can't benchmark the differences.

ChrisJefferson:
There is hypothetically some tiny slowdowns that could occur, as we will now read existing blocks of zeros and "or" them with the newly created blocks, but this is a tiny change which only occurs in the final case, and I can't detect the differences with some quick benchmarks I wrote.

src/blister.c
```diff
@@ -1168,8 +1165,6 @@ Obj FuncBLIST_LIST (
         if ( l * lenSub < 2 * lenList ) {
 
             /* allocate the boolean list and get pointer                   */
-            blist = NewBag( T_BLIST, SIZE_PLEN_BLIST( lenList ) );
```
fingolfin: Aren't the `lenSub` and `lenList` computations a few lines above redundant, too? (not that it matters :-)

ChrisJefferson: This whole function could maybe do with a nice cleanup, but to start with I wanted to do the minimum amount to convince myself they are the same, and merge them together :)

fingolfin: Of course that makes perfect sense (which is why I approved this already :-)

PR #2888, opened 2018-09-30
kernel: add {Ass,Elm,Isb,Unb}ComObj helpers

fingolfin:
This unifies the handling of (atomic) component objects across
interpreter, executor and compiler. Notable differences:

* the compiler used to use `AssARecord` instead of `SetARecordField`,
  unlike the interpreter and executor
* an unused and undocumented debugging code section inside
  `#ifdef CHECK_TL_ASSIGNS` was removed

@rbehrends it'd be great if you checked the few HPC-GAP related bits in this PR!


fingolfin:
Warning: if one of PR #2888 and #2889 is merged, the other must be rebased and updated before also being merged!

PR #2889, opened 2018-09-30
kernel: add {Ass,Elm,Isb,Unb}PosObj helpers

fingolfin:
Note the `MEMBAR_READ` call in `ElmPosObj` which is "essential" according
to a comment: this call was previously only present in the interpreter,
but missing in the executor.

Also, the code generated by the GAP compiler was more or less completely
broken in HPC-GAP, and lacked input validation in `C_ELM_POSOBJ`, which
`ElmPosObj` now provides "for free".

@rbehrends it'd be great if you checked the few HPC-GAP related bits in this PR!

### DIFF
--- a/lib/compiler.g
+++ b/lib/compiler.g
@@ -247,7 +247,6 @@ end );
 ##  "FastPlainLists"
 ##  "CheckTypes"
 ##  "CheckListElements"
-##  "CheckPosObjElements"
 ##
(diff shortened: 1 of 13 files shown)


fingolfin:
Warning: if one of PR #2888 and #2889 is merged, the other must be rebased and updated before also being merged!

PR #2894, opened 2018-10-04
kernel: refactor traversal code

fingolfin:
The TraversalState now is passed around to all functions using it explicitly. This allows us to get rid of `TLS(traversalState)`.

This makes it a lot easier to reason about the life time of any given traversal state, and also avoids potential problems caused by a traversal being interrupted (and aborted) by error handling.

Also contains on commit from PR #2845 to avoid annoying to resolve conflicts with that PR.


src/hpc/traverse.c
```diff
@@ -92,20 +85,16 @@ static void BeginTraversal(TraversalState * traversal)
     traversal->listSize = 0;
     traversal->listCapacity = 10;
     traversal->listCurrent = 0;
-    traversal->previousTraversal = currentTraversal();
-    TLS(traversalState) = traversal;
 }
 
-static void EndTraversal(void)
+static void EndTraversal(TraversalState * traversal)
```
ChrisJefferson: Is there any reason to keep EndTraversal around?

fingolfin: Hmm, not really...

ChrisJefferson: Was mainly curious if there were later plans for it. Is there any obvious "transversally finished" assertion checks that could be put in there? Doesn't matter if not.

fingolfin: Can't really think of any right now, but perhaps @rbehrends can?

Actaully, In a way, the motivation for this PR *was* to get rid od EndTraversal: because if there was any erro during a traversal, we'd abort the traversal but never called EndTraversal, leading to a leak at best, and possibly worse.

Issue #2871, opened 2018-09-26
Improve ApplicableMethod


ssiccha:
This stems from the discussion in #2860.

- Suggestion by @fingolfin: add "options" (in the sense of https://www.gap-system.org/Manuals/doc/ref/chap8.html) to ApplicableMethod that one can use instead of those magic constants 1 to 6 in the third argument, in order to adjust the output.
  - maybe get rid of the printlevel parameter in the process?
- Get rid of the messages that concern the first applicable method in the output of `ApplicableMethod( String, [Group(())], 1, 2 );`. After all we were only asking for the second method.
- Some of the line breaks in `ApplicableMethod( String, [Group(())], 1, 2 );` are really awkward.
- Improve the `Skipped: ` messages e.g. in `ApplicableMethod( IsCyclic, [Group(())], 1, "all" );`. `Skipped` seems to imply "not applicable". Maybe instead print `Applicable method no. i.`.

 Doing `ApplicableMethod( String, [Group(())], 1, 2 );` yields:
```
#I  Searching Method for String with 1 arguments:
#I  Total: 70 entries
#I  Method 2: ``String'' at /home/sergio/projects/gap/lib/teaching.g:794 , value: 
10000
#I  Skipped:
#I  Method 
10
 : ``String: for a group with generators'' at /home/sergio/projects/gap/lib/grp.gi\
:4259 , value: 36
function( G ) ... end
```

Doing `ApplicableMethod( IsCyclic, [Group(())], 1, "all" );` yields:
```
#I  Searching Method for IsCyclic with 1 arguments:
#I  Total: 7 entries
#I  Method 1: ``IsCyclic: system getter'' at /home/sergio/projects/gap/lib/grp.gd:
157 , value: 2*SUM_FLAGS+38
#I  Skipped:
#I  Method 4: ``IsCyclic'' at /home/sergio/projects/gap/lib/grp.gi:35 , value: 36
#I  Skipped:
#I  Method 
5
 : ``IsCyclic: generic method for groups'' at /home/sergio/projects/gap/lib/grp.gi:
44 , value: 35
#I  Skipped:
#I  Method 
7
 : ``IsCyclic: default method requiring categories and checking properties'' at /ho\
me/sergio/projects/gap/lib/grp.gd:157 , value: 1
#I  Skipped:
[ function( obj ) ... end, function( G ) ... end, function( G ) ... end, 
  function( <1 unnamed arguments> ) ... end ]
```

markuspf:
[...]

PR #2896, opened 2018-10-05
Add filename and line to functions when printed

ChrisJefferson:
This PR adds to functions their filename, and line, when they are printed (when the function comes from an actual file). As an example:

```
gap> Print(List);
# /home/caj/files/reps/gap/gap/lib/coll.gi:645
function ( arg... )
    local tnum, C, func, res, i, elm, l;
    ...
```

I put a `#` at the start so the printed out information is still a valid piece of GAP.

I'm interested if anyone thinks this is a good idea, or if there should be any other adjustments. I haven't gone through and done the (fairly mechanical) changes required to the tests yet.


fingolfin:
Actually, it is completly unclear to me how you plan to adjust the tests for this, given that paths will be different on every system. While I can think of a few approaches, none seems satisfying. So, what is the "mechanical" approach you have in mind?

ChrisJefferson:
Woops, you are right! My first thought was the boring bit of fixing the tests.

We've previously discussed some kind of way of marking the gap library root in filenames (something like `GAPROOT/lib/coll.gi`), which would make the path system-independant, at the cost of being harder to just cut+paste into an editor. If we didn't do that, then indeed I can't really see how to make this a valid patch.

fingolfin:
(Ignoring the technical concerns, I like the feature)

cdwensley:
It is a good feature - and is related, with the same problems, to issue #2871.

fingolfin:
One of the various way s one could address this in tests is to teach Test to transform outputs by replacing any occurrence of the GAP root path by the string GAPROOT, as we do in testspecial.

ChrisJefferson:
I could add a configuration option, which anyone can turn on/off, and we turn off when testing (I'd prefer to leave it on by default).

fingolfin:
But then we still need to ask all affected package authors to do that in their packages. But perhaps not that many are affected -- it'd be good to know how many before making any decision on that.

src/calls.c
```diff
@@ -1105,6 +1105,16 @@ void PrintFunction (
       return;
     }
 
+    Obj filename = GET_FILENAME_BODY(BODY_FUNC(func));
+    UInt sl = GET_STARTLINE_BODY(BODY_FUNC(func));
+
+    if(filename && sl) {
+        // Filenames starting '*' are special streams
+        if(CHARS_STRING(filename)[0] != '*') {
```
fingolfin:
```suggestion
        if (CHARS_STRING(filename)[0] != '*') {
```

stevelinton:
Maybe this should be part of a Display method for functions, rather than Print. That way it won't break existing tests, and it opens the way to adding more info if we think of any.

fingolfin:
In light of PR #2987 I think going with @stevelinton's suggestion might be the best course of action... ?

ChrisJefferson:
I've now moved the message to Display, moved the test to testspecial (where we already deal with filenames containing GAPROOT), and also stopped printing names that start with `*`, like `*stream*` and `*stdin*`. Obviously someone could have a filename that starts with a star, but it would be a silly name and there is a chance we already handle it badly somewhere.

alex-konovalov:
This may be useful. I am using `PageSource` for this, since it displays the location before displaying the source, but `Display` will be more convenient.

PR #2898, opened 2018-10-05
Replace GAP_ROOT_PATHS in tests with 'GAP_ROOT_PATH'

ChrisJefferson:
As suggested in #2896.

I decided to use the string `'GAP_ROOT_PATH'`, rather than `GAPROOT`, because GAP actually outputs `GAPROOT` in some places and I thought it might be better if the replacement isn't something that GAP already outputs as standard (also it more tightly ties it to the `GAP_ROOT_PATH` variable. I put it in single quotes because otherwise you get `GAP_ROOT_PATHlib/init.g`, which looks bad, but I didn't really want to start stripping slashes (maybe I should)

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -87,7 +87,7 @@ end);
 
 InstallGlobalFunction(RunTests, function(arg)
   local tests, opts, breakOnError, inp, outp, pos, cmp, times, ttime, nrlines,
-        s, res, fres, t, f, i;
+        s, res, fres, t, f, i, gaproot;
   # don't enter break loop in case of error during test
   tests := arg[1];
   opts := rec( breakOnError := false, showProgress := "some" );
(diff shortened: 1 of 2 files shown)


lib/test.gi
```diff
@@ -136,6 +136,11 @@ InstallGlobalFunction(RunTests, function(arg)
     t := Runtime() - t;
     CloseStream(fres);
     CloseStream(s);
+    # Remove explicit references to GAP's root path, so
+    # tests can be run on multiple machines
+    for gaproot in GAPInfo.KernelInfo.GAP_ROOT_PATHS do
+        res := ReplacedString(res, gaproot, "'GAP_ROOT_PATH'");
```
fingolfin: How about `GAP_ROOT_PATH/` instead? I'd find that more pleasing to the eye.

fingolfin: Note that we already use `GAPROOT` in the kernel for the same purpose, and also in `tst/testspecial`. I would really prefer to keep these consistent. I don't care much which one (GAPROOT or GAP_ROOT_PATH) is used in the end, just as long as the same is used. (I have a slight preference for GAPROOT, but I'll be happy to accept whichever you prefer)

ChrisJefferson: So, my thought process was that we shouldn't use GAPROOT, because, as you say, GAP sometimes prints GAPROOT and we might want to be able to tell if it printed GAPROOT or the actual path.

If we decide that's too subtle an issue to bother about, I'm happy to change to GAPROOT.

ChrisJefferson:
Adjusted to print GAPROOT

PR #2900, opened 2018-10-05
Teach Test() to abort test if ctrl-C is pressed

### LINKED ISSUE #1879: Improve interaction between ctrl-c and Test()
Right now, pressing ctrl-c while `Test(...)` is running does not open a break prompt, nor does it abort `Test(...)`. Instead, the user interrupt is treated like an error in the code being tested. This makes it difficult to abort certain long running tests; sometimes, if you are lucky, you can press ctrl-c quickly repeatedly to at least exit GAP, but even that can fail.

Perhaps we could teach `Test()` and friends to treat user interrupts differently -- i.e. if a user interrupt happens, either abort `Test()`, or (for bonus points) actually enter a break prompt (that might be difficult to pull of, though; it's not that important anyway).

To implement this, I imagine we'd modify (at least) the following places:
1 . `READ_TEST_OR_LOOP`, to teach it to abort upon user interrupt;
2. `FuncREAD_STREAM_LOOP`, to signal whether the operation completed normal, or was interrupted by the user (and possibly more), e.g. via its return value
3. `RunTests` in `lib/test.gi` then would track the return value of `READ_STREAM_LOOP`, and if it indicates a user interrupt, would take appropriate steps, e.g. abort itself, and possibly indicate that via a return value, too. Or perhaps better, it could just issue an `Error("user interrupt detected")` of its own, with the break loop re-enabled; then the user can choose whether to abort the test (via `quit;`), or to continue to the next test file (by doing a `return;`; then I imagine right after the `Error`, `RunTests` would return itself, so that processing the current file ends).

fingolfin:
This is a bit of a hack, but seems to work very well in practice. You can even resume the test with `return` if you want (but the interrupted line will still lead to diffs).

An arguably "better" (but much more complicated) solution would be to add a flag which the kernel uses to signal to the error handler that the error is a user interrupt; then we could support seamlessly resuming of the interrupted test. But this requires further work, e.g. re-redirecting input/output temporarily, saving and restoring `BreakOnError` and more. Not sure that this would be worth it. In the meantime, this simple solution already gets us most of the benefit.

Resolves #1879

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -136,6 +136,12 @@ InstallGlobalFunction(RunTests, function(arg)
     t := Runtime() - t;
     CloseStream(fres);
     CloseStream(s);
+    # check whether the user aborted by pressing ctrl-C
+    if StartsWith(res, "Error, user interrupt") then
+        BreakOnError := breakOnError;
+        Error("user interrupt");
+        BreakOnError := opts.breakOnError;
+    fi;
     Add(cmp, res);
     Add(times, t);
   od;


fingolfin:
Good point; comment added

fingolfin:
@ChrisJefferson changed to use `StartsWith(res, "Error, user interrupt") ` as test.

PR #2904, opened 2018-10-08
Fix task scanner to use new Julia API call

fingolfin:
With this, GAP can be used with the Julia master branch (and thus with Julia 1.1, when it comes out).

This is localized to the Julia GC code, so it should be safe to backport to GAP 4.10.


fingolfin:
Backported in 41517f5ef82765cc9114ea598658844aec3c617e

PR #2907, opened 2018-10-08
Fork improvements

ChrisJefferson:
This PR adds `SyFork`, a wrapper around for which allows functions to be called pre- and post- fork. Profiling then uses this to create a new profile output for each child, to avoid the profile output getting corrupted.

If/when this is merged, I will force the IO package to use SyFork (when it is available), so forks which happen in IO can also be tracked.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -2956,7 +2956,51 @@ void SySetErrorNo ( void )
 /****************************************************************************
 **
 *F * * * * * * * * * * * * * file and execution * * * * * * * * * * * * * * *
+**
 */
+
+enum { signalSyForkFuncsLen = 16 };
+
+static ForkObserver signalSyforkFuncs[signalSyForkFuncsLen];
+
+Int RegisterSyForkObserver(ForkObserver func)
+{
+    Int i;
+    for (i = 0; i < signalSyForkFuncsLen; ++i) {
+        if (signalSyforkFuncs[i] == func) {
+            return 1;
+        }
+        if (signalSyforkFuncs[i] == 0) {
+            signalSyforkFuncs[i] = func;
+            return 1;
+        }
+    }
+    return 0;
+}
+
+Int SyFork(void)
+{
+    for (int i = 0; i < signalSyForkFuncsLen; ++i) {
+        if (signalSyforkFuncs[i]) {
+            signalSyforkFuncs[i](PreFork);
+        }
+    }
+
+    Int ret = fork();
+    if (ret < 0)
+        return ret;
+
+    ForkState state = (ret == 0) ? PostForkChild : PostForkParent;
+
+    for (int i = 0; i < signalSyForkFuncsLen; ++i) {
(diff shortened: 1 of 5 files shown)


src/sysfiles.c
```diff
@@ -3109,7 +3153,7 @@ UInt SyExecuteProcess (
       func2 = &NullSignalHandler;
 
     /* clone the process                                                   */
-    pid = vfork();
```
fingolfin: I think the use of `vfork` here is/was meant as an optimization that preserves memory and is also easier to implement on, say, Windows, than full `fork`. This PR silently changes it to `fork`, a deoptimization. Do we really want that? Perhaps a `SyVFork` is needed, too?

src/sysfiles.h
```diff
@@ -401,6 +401,28 @@ extern void SySetErrorNo ( void );
 *F * * * * * * * * * * * * * file and execution * * * * * * * * * * * * * * *
 */
 
+typedef enum ForkState { PreFork, PostForkChild, PostForkParent } ForkState;
+typedef void (*ForkObserver)(ForkState);
+
+/*
+ *F RegisterSyForkObserver( <func> )
+ ** Register a function to be called before and after fork is called.
```
fingolfin: So, when adding documentation comments, I'd either make them match the existing ones, or radically different; but this is kind of a mishmash (only one space after `**`, no empty line after the `*F` line, no start line with `/*****`.

If you don't mind, I can amend this PR accordingly.

ChrisJefferson:
Oh, I also removed a use of `vfork`, as we are using `vfork` in a very inapproriate way (I'm assuming on modern OSes vfork and fork must be equivalent, as by the book our use of vfork shouldn't be legal I believe).

src/sysfiles.c
ChrisJefferson: The spec for `vfork` says the only functions you are allowed to call after a `vfork` is `_exit` or `exec*`. We are calling `chdir`, `open`, `fcntl` and `SyBufFileno` (which is one of our functions, but I believe even that isn't acceptable). I can't find any spec that says that's acceptable, even on Linux. However, I shouldn't have bundled this into the same PR, as if we want to keep vfork a SyVfork would be easy to add.

src/sysfiles.h
ChrisJefferson: I'm happy to adjust it. I want to look at why hpc-gap is crashing (I suspect switching vfork for fork, which suprises me, as I expected that to make hpc-gap more stable, not less.

src/sysfiles.c
ChrisJefferson: Actually, cygwin's vfork just calls fork it turns out ( https://github.com/wallento/newlib-cygwin/blob/dbe905c140d0cd347c731f63b153dca53857bdc9/newlib/libc/sys/linux/process.c#L32 ), although I think linux vfork actually does "proper" vfork.

ChrisJefferson:
closing because this was a stupid idea. Sorry.

PR #2914, opened 2018-10-11
Flush OutputtedFileList when creating new file

ChrisJefferson:
Further tweaks to improve profiling in IO_fork.

There are two changes here:

1) Make sure we reset OutputtedFilenameList when restarting the profile
2) Make sure the newly outputted file is valid, by printing out the header, and also `fflush`ing to make sure the header is definately outputted.

This is used directly in https://github.com/gap-packages/io/pull/71 to prove it works this time :) After this is merged, I'll remove the code from that PR which refers to this branch.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -164,6 +164,20 @@ struct ProfileState
 UInt profileState_Active;
 
 
+// Output information about how this profile was configured
+void outputVersionInfo(void)
+{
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
I will point out this PR, and the relevant bit in IO, suggests a deeper problem with fork and exec -- that we should flush all buffers, at both the C and GAP level, before forking, to avoid duplicated or lost output. However, that's a much deeper issue which could require serious surgery.

PR #2919, opened 2018-10-15
Replace all calls to SetFeatureObj, mark it as obsolete

fingolfin:
Various packages still use it, and most likely also code that people keep
around locally, so we cannot and should not remove it for the foreseeable
future. But at least new code should not use it.


fingolfin:
This is a rather trivial PR, it would be great if somebody could review it...

PR #2922, opened 2018-10-16
Remove most of the code dealing with (re)setting filters on internal lists

fingolfin:
This experimental PR removes support for setting and clearing "filters" for non-objects, specifically for internal lists. Most of this code never was used; only a single code path (which remains) is used in order to set the `IsSSortedList` filter for plists, ranges, strings, blists.

The main motivation for writing this was to reduce the complexity of how we handle filters, and to better understand the code -- when I started out writing it, I didn't at all know what was going on. Now that I have a feeling that I do understand, I could also leave the code in, and instead add some comments everywhere that try to give some clues on what's going on.

Alas, unless somebody plans to extends support for filters on internal lists, I don't quite see the value in keeping this extra code around. To the contrary, I am tempted to also remove `IsNSortedList` and `IS_NSORT_LIST` (which nothing in the library or packages use); instead, one can use `HasIsSortedList(x) and not IsSortedList(x)` to check for "known to be not sorted"; a small change to the kernel function `SET_FILTER_LIST` can make it do something similar, too.

Yet another road to extend this would be to replace `SetIsSSortedList` by a custom function, which calls `SET_FILTER_LIST` (so it would not be removed from the GAP level after all) for internal lists, and the original `SetIsSSortedList` for everything else. That would localize this "hack" even more,  would ensure that inputs like these result in an error, instead of being silently accepted:
```
gap> SetIsRectangularTable([[1],[]], true);
gap> SetIsNilpotentGroup([], true);
```


ChrisJefferson:
I certainly approve of this direction, I've never understood this code as well as I'd like.

fingolfin:
I guess it would be good if @stevelinton could comment, too. Likewise, perhaps @ThomasBreuer or @frankluebeck have any thoughts on this?

ThomasBreuer:
- Only a handful of filters are supported for plain lists,
  and `IsSSortedList` is the only one for which the value can be set/reset.
  Thus I agree that it is a good idea to remove the current general mechanism.

- Removing `IsNSortedList` would be fine.
  It is not documented, and it also does not work, one gets the error message `"not ready yet"`.

- As far as I understand the changes,
  the `SetIsSSortedList` method for internal lists shall simply be removed.
  However, this feature is used in several places in the GAP library.

stevelinton:
Happy with this. This code is clearly not worth the complexity it adds, even if it were ever used. I don't think `SetIsSortedList` will stop working for internal lists, it will carry on being handled inside `DoSetProperty` in the kernel. What is being removed is a method that can never be called.

fingolfin:
@ThomasBreuer regarding the remove `SetIsSSortedList` method, @stevelinton has it exactly right: that method *never* was called, as the C code deals with that case.

What one could consider, though, is removing that hack from the kernel, and instead do something like this pseudo code on the library level:
```gap
SetIsSSortedList_orig := SetIsSSortedList;
Unbind(SetIsSSortedList)
SetIsSSortedList := function(obj, val)
  if IsInternalRep(obj) and IsList(obj) then
    if val then
      SET_FILTER_LIST(l, IS_SSORT_LIST);
    else
      SET_FILTER_LIST(l, IS_NSORT_LIST);
    fi;
  else
    SetIsSSortedList_orig(obj, val);
  fi;
end;
```

As described above, this would also mean that the following command would generate an error (IMHO an improvement)
```
gap> SetIsNilpotentGroup([], true);
```

fingolfin:
@ThomasBreuer I agree it is strange, which is why also tracked down the origin of this code before creating this pull request. I am reproducing my findings here: the kernel function `SET_FILTER_LIST` was added by Frank Celler on 1997-12-15; it was then used by you just a few hours later to convert some calls like `IsSSortedList( F );` to `SetIsSSortedList( F, true );`

In contrast, the custom `SetIsSSortedList` method was added in a commit by @hulpke on 1998-07-16; that commit has the commit message "Merge." and touches over 600 files, but our repository gives no further hints on the origin of these changes. As such, I can only speculate as to who added this method and why. But it's not hard to verify that it never can be called via method selection these days (both by studying the relevant kernel code, or by experiments: insert an `Error` into it and try to trigger it). I also inspected the codebase for the commit in 1998, and double checked that already then, `DoSetProperty` in `src/opers.c` ends like this:
```C
...
    else if ( TNUM_OBJ( obj ) == T_DATOBJ ) {
        flags = (val == True ? self : TESTR_FILT(self));
        CALL_2ARGS( SET_FILTER_OBJ, obj, flags );
    }
    else if ( IS_PLIST(obj) || IS_RANGE(obj) || IS_STRING_REP(obj)
           || IS_BLIST_REP(obj) )  {
        if ( val == True ) {
            FuncSET_FILTER_LIST( 0, obj, self );
        }
    }
    else {
        ErrorReturnVoid(
            "value cannot be set for internal objects",
            0L, 0L,
            "you can return without setting it" );
    }

    /* return the value                                                    */
    return 0;
}
```
In particular, method dispatch is *not* used for lists. I should note that the commit in 1997 by Frank Celler added the `IS_PLIST(obj) || IS_RANGE(obj) || ...` case, so also before that commit, method dispatch would not have worked for `SetIsSSortedList` on e.g. plists. And this code has otherwise been essentially unchanged since the first commit in the GAP4 CVS repository on 1996-07-03 by Martin Schönert.

fingolfin:
As to the error messages: You talk about attribute setters, and what you say there sounds about right. But this is concerning a property setter, and there GAP 4 always produced errors for non-objects:
```
gap> SetIsFinite(rec(), true);
Error, Value cannot be set for internal objects
you can 'return;' without setting it
brk>
gap> SetIsFinite(fail, true);
Error, Value cannot be set for internal objects
you can 'return;' without setting it
brk>
```
In contrast:
```
gap> SetSize(rec(), 1);
gap> SetIsFinite([], true);
```

ThomasBreuer:
@fingolfin Thanks for your comments.

From my point of view, properties are special attributes,
and I think the exact behavior of attribute setters is not documented yet,
thus we can think about changing it.

Do you perhaps have a general rule like the following in mind?

- When a property setter is called for an internal object then an error is signaled,
  except if the object is a list and the setter is `SetIsSSorted`.
- When a property setter is called for a non-internal object then
  * it does nothing if the property in question is already set to the given value,
  * an error is signaled if the property in question is already set the other way,
  * it changes the type of the object such that the property gets set.
- When the setter of an attribute is called that is not a property then
  * it does nothing if a value for the attribute is already known, according to the attribute tester,
  * it delegates to the method selection otherwise;
    the default setter method does nothing, but other methods can be installed, even for internal objects.

I am not sure that the different behavior of properties for internal and non-internal objects
is a good idea.
One can think of code that suddenly runs into errors if some of the non-internal input objects
get replaced by internal objects.

(Just for curiosity:
Does it make sense at all that property setters and property testers are operations,
if their methods are not used?)

fingolfin:
Property setters methods *are* used, but only for external objects, not internal.

fingolfin:
Actually, strike that, I misread the code just now: What really happens that `SetFilterObj` is called, and is a plain function, so you are of course 100% correct, no method selection is used.

I have no strong opinion on whether method selection should be used here or not. Of course one could argue that some kind of exotic object might want to store properties in a custom way. But that would be problematic, because for method selection, we assume that we can fetch the properties as part of the flags. So, I guess I do indeed have a small preference to make them regular functions, not operations, as that would as a positiv side effect result in a helpful error if somebody tried to install a custom method for a property setter.

fingolfin:
I thought again about this, and there is a reason why property setters are operations: the flags set by the setter are stored in a part of the function bag which only exists for operations. So they have to be operations.

Anyway, other than that, I think we could merge this PR as-is for now -- it does not in any way change the existing behavior of GAP (while we discussed some ways we might change it, this PR actually does none of that). All it dos is remove some unused code which we seem to agree that it doesn't seem useful.

Or am I missing something?

Issue #1456, opened 2017-06-27
Plan for phasing out obsolete identifiers


fingolfin:
Right now we mark some stuff as obsolete, and even have a warning class `InfoObsolete` for it. But we are completely toothless about phasing out these identifiers, and indeed, recently some of them got used again in the library (see parts of PR #1452).

Part of the problem is that `InfoObsolete` default to 0, so most people never will now they are using obsoletes. To a degree, this default level makes sense because it avoids warnings about packages that still use these obsoletes.

I propose the following changes:

### Revise default `InfoObsolete` level and how we use it.

I propose we change the the default level of `InfoObsolete` to 1. Then we revise all uses of `InfoObsolete`:
*  For things were we determine that packages do not use them anymore, warn about them at level 1 (so the user will get warning about them by default
* for things which are still used in packages, use level 2 or higher
* To achieve this, we can add an extra `level` argument to `DeclareObsoleteSynonym`.
* in order to not flood the user with warnings for cases were an obsolete function is used in a loop, we could also enhance `DeclareObsoleteSynonym` in such a way that the info message is only shown once (or only shown every few seconds/minutes).

### Add warnings for access to obsolete global variables

We don't really warn about global variables being used (only about global functions, operations, attributes etc.). We could achieve this with a trick, though, by abusing the "automatic variables" feature (even though some people might want to phase it out eventually for HPC-GAP... but in HPC-GAP, we can use TLS variables for the same effect).

The idea is that an automatic variable is only actually initialized when it is actually accessed, and this is done by invoking a special function returning the value for the variable -- and that function can of course issue an `Info` warning. This warning will only be shown once, of course -- that might actually be an advantage. Indeed, come to think about it, the same  "automatic variable" trick could/should perhaps be used for obsolete functions, operations etc., too, to get the message only once. 

### Distinguish different "stages" of obsolete identifiers

We could refine `DeclareObsoleteSynonym` to distinguish two kinds of obsoletes: "obsolete alias, on by default" (as is currently the case), and also "obsolete alias, off by default"). 
Then we could refine the `-O` option which turns off obsoletes, to switch between three (instead of two) modes:  "all obsoletes on"; "all obsoletes of"; and "some obsoletes on". We might eventually make the last one the new default.

[...]

PR #2923, opened 2018-10-16
Extend obsolete to support multiple levels

ChrisJefferson:
One problem with the way we currently handle 'obsolete' is that normal users of GAP will never see an obsolete warning, until we delete the function (as they don't tend to increase InfoObsolete, or run with `-O`).

This extends 'obsolete' into two levels, 1 and 2. Level 1 messages are always printed, level 2 messages have to be explicitly enabled.

The -O option still disables any obsolete functions from being used at all.

My long-term plan is that we would put obsolete messages in a major release at level 1 before removing them, so users would actually see a warning, and have the chance to fix their code.

If changing how InfoObsolete works upsets anyone, we could add a different variable (`InfoObsoleteSoon`?) which displayed messages about things we will delete in the future.

### DIFF
--- a/doc/ref/obsolete.xml
+++ b/doc/ref/obsolete.xml
@@ -199,13 +199,16 @@ Note that <C>MultRowVector</C> was also renamed to <C>MultVectorLeft</C>.
 <InfoClass Name="InfoObsolete"/>
 <Description>
 is an info class to display warnings when an obsolete variable is used.
-By default, these warnings are switched off since the info level for 
-this class is 0. Setting it to 1 will trigger warnings if &GAP; will 
-detect that an obsolete variable is used at runtime.
+By default, the info level for this class is set to 1, which will only
+show variables which will be removed in the next major version of GAP.
+Setting it to 2 will trigger further warnings, for variables which have
+alternative names, or may be removed in future.
+
+This class can be set to 0 to disable all obsolete warnings.
 <P/>
-To check that the &GAP; code does not use obsolete variables at the 
-parsing time, and not at a runtime, use <C>-O</C> command line option,
-see <Ref Sect="Command Line Options"/>. 
+To check that the &GAP; code does not use any obsolete variables at
+parsing time, and not at a runtime, use the <C>-O</C> command line option,
+see <Ref Sect="Command Line Options"/>.
 </Description>
 </ManSection>
 
--- a/lib/obsolete.gd
+++ b/lib/obsolete.gd
@@ -55,6 +55,17 @@
 ##  Usually, old names will be available for several releases;
 ##  they may be removed when they don't seem to be used any more.
 ##  <P/>
+##  Information about obsolete names is printed by <Ref Func="Info"/> using the
+##  InfoObsolete Info class. By default InfoObsolete is set to 1. Newly
+##  obsoleted identifiers should at first be outputted at info level 2. Once they
+##  have been removed from all packages, they should then be moved to info level
+##  1, so they are visible to normal users, for at least one major release before
+##  being removed.
+##  <P/>
+##  The functions DeclareObsoleteSynonym and DeclareObsoleteSynonymAttr take
+##  an optional final paremeter, specifying the info level at which the given
+##  obsolete symbol should be reported. It defaults to 1.
+##  <P/>
 ##  The obsolete &GAP; code is collected in two library files,
 ##  <F>lib/obsolete.gd</F> and <F>lib/obsolete.gi</F>.
 ##  By default, these files are read when &GAP; is started.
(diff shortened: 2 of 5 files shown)


fingolfin:
See also issue #1456

lib/obsolete.gd
```diff
@@ -76,19 +76,29 @@
...
     value := EvalString( name_current );
     if IsFunction( value ) then
         orig_value := value;
+        printed_warning := false;
         value := function (arg)
             local res;
-            Info( InfoObsolete, 1, "'", name_obsolete, "' is obsolete.",
-                "\n#I  It may be removed in a future release of GAP.",
-                "\n#I  Use ", name_current, " instead.");
+            if not printed_warning then
```
fingolfin: It seems this change also causes each Info message to be shown only once. That seems reasonable to me, but it should be mentioned in the commit message and the PR description.

lib/obsolete.gd
```diff
@@ -76,19 +76,29 @@
 ##  <#/GAPDoc>
 ##
 
-BIND_GLOBAL( "DeclareObsoleteSynonym", function( name_obsolete, name_current )
-    local value, orig_value;
+BIND_GLOBAL( "DeclareObsoleteSynonym", function( name_obsolete, name_current, level_arg... )
+    local value, orig_value, printed_warning, level;
```
fingolfin: The new argument `level_arg` is not mentioned in the commit message nor the PR description. It seems to be meant to allow overriding the "obsoleteness level" for specific synonyms. I think this is sufficiently non-trivial that at least a brief comment should also be added to the code, perhaps also giving an example of the intended use?

I assume the idea is that over time, certain synonyms are moved to different levels, until they are finally removed (which may never happen in some cases, of course).

If so, shouldn't this argument also be added to `DeclareObsoleteSynonymAttr`?

tst/testbugfix/2013-06-14-t00300.tst
```diff
@@ -3,7 +3,7 @@ gap> foo:=function() return 42; end;
 function(  ) ... end
 gap> DeclareObsoleteSynonym("bar","foo");
 gap> oldLevel := InfoLevel(InfoObsolete);;
-gap> SetInfoLevel(InfoObsolete,1);
+gap> SetInfoLevel(InfoObsolete,2);
```
fingolfin: Non-blocker, but: it'd be nice if the new features were also tested. Like the fact that info messages are now printed only once; or the new `level_arg`

ChrisJefferson:
Hopefully all @fingolfin 's comments addressed, once tests cycle (adding a test did find a bug, where only printing the message once meant the message was never shown if the function had been called when the Info level was too low to show it).

doc/ref/obsolete.xml
```diff
@@ -199,13 +199,16 @@ Note that <C>MultRowVector</C> was also renamed to <C>MultVectorLeft</C>.
...
+By default, the info level for this class is set to 1, which will only
+show variables which will be removed in the next major version of GAP.
+Setting it to 2 will trigger further warnings, for variables which have
+alternative names, or may be removed in future.
+
+This class can be set to 0 to disable all obsolete warnings.
 <P/>
-To check that the &GAP; code does not use obsolete variables at the 
+To check that the &GAP; code does not use any obsolete variables at the
 parsing time, and not at a runtime, use <C>-O</C> command line option,
```
fingolfin: Not your fault, so feel free to ignore, but I was wondering:

I am not a native speaker, so this may be totally off, but: "shouldn't it be "at parsing time", without "the"? And then "use -O command line option" -> "use the -O command line option"?

Regardless of that, I wonder how clear the phrase "parsing time" is for the average GAP user. In fact, I am not sure *I* understand what is meant here... Perhaps "... while reading the GAP library"? But even with that... How about: 
> To check that the GAP library and any other GAP code being loaded does not reference or use obsolete variables and functions in any way, one may use the -O command line option (see REF) to completely disable the definition of any obsolete symbols. Note that this may cause various GAP packages to fail to load, or to load incompletely.

fingolfin: (but again: I don't require you to make that change, just wondering)

lib/obsolete.gd
```diff
@@ -55,6 +55,17 @@
 ##  Usually, old names will be available for several releases;
 ##  they may be removed when they don't seem to be used any more.
 ##  <P/>
+##  Information about obsolete names is printed by <Ref Func="Info"/> using the
+##  InfoObsolete InfoClass. By default InfoObsolete is set to 1. Newly
```
fingolfin: InfoClass -> info class?

lib/obsolete.gd
```diff
@@ -55,6 +55,17 @@
...
+##  obsoleted identifiers should be first be outputted at info level 2. Once they
```
fingolfin: "be first be outputted" ? Perhaps "at first be output" ("output" is its own past tense, like "read"). Though perhaps "at first be set to output at info level 2" is even better?

lib/obsolete.gd
```diff
@@ -55,6 +55,17 @@
...
+##  have been removed from all packages, they should then be moved to info level
+##  1, so they are visible to normal users, for at least one major release before
+##  being removed.
+##  </P>
+##  The functions DeclareObsoleteSynonym and DeclareObsoleteSynonymAttr take
+##  an optional final paremeter, specifying the Info level at which they print,
```
fingolfin: "specifying the info level at which the given obsolete symbol should be reported. It defaults to 1."

fingolfin:
The manual tests fail.

doc/ref/obsolete.xml
alex-konovalov: I agree with @fingolfin (and that text was written by me, IIRC).

PR #2927, opened 2018-10-17
Adjust bin/BuildPackages.sh for new simpcomp version

fingolfin:
With this tweak, one can build the currently shipping version of simpcomp;
future (fixed) releases; and also the simpcomp git repository.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -214,9 +214,11 @@ build_one_package() {
     ;;
 
     simpcomp*)
-      chmod a+x configure depcomp install-sh missing && \
-      echo_run ./configure && \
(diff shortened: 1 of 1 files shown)


alex-konovalov:
@fingolfin Is the plan that this is a temporary measure, and eventually simpcomp will comply with the standard `run_configure_and_make` ?

fingolfin:
Yeah, sure. Once there a version of simpcomp with these changes has been released, and some time has passed, we can remove the special case. But I'd wait a bit for this. In particular, let's not do that in GAP 4.10.

fingolfin:
Backported via 115894b137258917244e0c9051cd9d8c1fc4609e

PR #2931, opened 2018-10-18
kernel: when using julia_gc, call jl_atexit_hook in SyExit

fingolfin:
... as recommended by https://docs.julialang.org/en/v1/manual/embedding/index.html

This makes it possible to collect coverage data in Julia code (as that is written to disk via that atexit hook).

### DIFF
--- a/src/saveload.c
+++ b/src/saveload.c
@@ -372,11 +372,8 @@ static void LoadBagData ( void )
   flags = LoadUInt1();
   size = LoadUInt();
 
-    if (TNAM_TNUM(type) == NULL)
-      {
-        Pr("Bad type %d, size %d\n",type,size);
-        exit(1);
-      }
+  if (TNAM_TNUM(type) == NULL)
+    Panic("Bad type %d, size %d\n", (int)type, (int)size);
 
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
The changes to Panic/SyExit are sensible and good.

One random question -- why doesn't julia_gc just pass jl_atexit_hook to atexit?

src/saveload.c
```diff
@@ -374,8 +374,7 @@ static void LoadBagData ( void )
 
     if (TNAM_TNUM(type) == NULL)
       {
-        Pr("Bad type %d, size %d\n",type,size);
-        exit(1);
+        Panic("Bad type %d, size %d\n",type,size);
```
sebasguts: The compiler seems to complain about the `%d`

ChrisJefferson:
Wanting the exit code is a good enough reason, in that case I'm happy to merge it (it's not going to effect GAP without Julia anyway of course)

fingolfin:
Backported via 19aafe08e080fa23c53fd4292495f5017f8d6cbd

PR #2932, opened 2018-10-18
Replace calls to Random([a..b]) by Random(a,b)

fingolfin:
This avoids some overhead for creating ranges

Inspired by PR #2930

### DIFF
--- a/doc/ref/xtndxmpl.xml
+++ b/doc/ref/xtndxmpl.xml
@@ -918,7 +918,7 @@ gap> InstallMethod( Random,
 >    "for full collection Z/nZ",
 >    [ CategoryCollections( IsMyZmodnZObj ) and IsWholeFamily ],
 >    R -> MyZmodnZObj( ElementsFamily( FamilyObj(R) ),
->                    Random( [ 0 .. Size( R ) - 1 ] ) ) );
+>                    Random( 0, Size( R ) - 1 ) ) );
 gap> 
 gap> InstallMethod( Size,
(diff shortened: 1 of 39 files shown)


ChrisJefferson:
This probably doesn't matter, but just to mention, are you aware that Random([a..b]) and random(a,b) have different behaviour when b < a? Random([5..1]) enters a break loop, which Random(5,1) returns fail. I just mention it in case in some cases it could make a difference (obviously not whenever it's easy to see b >= a).

fingolfin:
Yes, I am aware of it.

doc/ref/xtndxmpl.xml
```diff
@@ -1329,7 +1329,7 @@ gap> InstallMethod( Random,
 >    "for full ring Z/nZ",
 >    [ CategoryCollections( IsMyZmodnZObjNonprime ) and IsWholeFamily ],
 >    R -> MyZmodnZObj( ElementsFamily( FamilyObj( R ) ),
->                    Random( [ 0 .. Size( R ) - 1 ] ) ) );
+>                    Random( 0 .. Size( R ) - 1 ) ) );
```
ChrisJefferson: I guess this is a typo? (The .. ).

fingolfin: Indeed, ouch :-(. I'll fix it, thank you very much for careful checking.

fingolfin: Fixed

PR #2933, opened 2018-10-18
Warning fix

ChrisJefferson:
This fixes two new warnings I see compiling GAP in gcc 8.2.

One of them `-Wno-cast-function-type` I am sure we should just disable, because GAP casts quite a lot of functions in strange ways, and there doesn't seem to be an easy way to hide it.

The other warning, `-Widentical-branches` (which checks if both sides of an if, or `?:` operator, are the same), I actually like. There is one place where it appears and it is hard to fix (checking `PATH_MAX`), so I add a slightly horrible hack (with comment) to handle the case where `PATH_MAX == 4096`.

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -82,8 +82,8 @@ GAP_STATIC_ASSERT(sizeof(void *) == SIZEOF_VOID_P, "sizeof(void *) is wrong");
 **  at least GAP won't be the only program to run into problems.
 */
 enum {
-#ifdef PATH_MAX
-    GAP_PATH_MAX = 4096 > PATH_MAX ? 4096 : PATH_MAX,
+#if defined(PATH_MAX) && PATH_MAX > 4096
+    GAP_PATH_MAX = PATH_MAX,
 #else
     GAP_PATH_MAX = 4096,
 #endif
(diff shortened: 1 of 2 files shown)


src/system.h
```diff
@@ -82,7 +82,9 @@ GAP_STATIC_ASSERT(sizeof(void *) == SIZEOF_VOID_P, "sizeof(void *) is wrong");
 **  at least GAP won't be the only program to run into problems.
 */
 enum {
-#ifdef PATH_MAX
+// The 'PATH_MAX != 4096' test stops compilers warning about idential values
+// on both sides of the : in the line below
+#if defined(PATH_MAX) && PATH_MAX != 4096
```
fingolfin: Hmm, can't we then go one step further and do this?
```C
#if defined(PATH_MAX) && PATH_MAX > 4096
    GAP_PATH_MAX = PATH_MAX,
#else
    GAP_PATH_MAX = 4096,
#endif
```

ChrisJefferson: That's much nicer! I'll do that.

ChrisJefferson:
I would suggest backporting this to 4.10, because at the moment GAP produces many pages of warnings when compiled on ubuntu 18.10.

fingolfin:
Backported via 16502e0c5d38ab6b7281f155067b7d2524378a23 and d4d604a3409c9d472e6c6b5cddffce591c5d63ad

PR #2935, opened 2018-10-20
Updated BuildPackages.sh

simpcomp-team:
Removed special rule for making simpcomp to comply with simpcomp version 2.1.9 from Oct 20 2018.


fingolfin:
Why is this necessary? I already adapted the rule to work with both current and past simpcomp, and it seemed to work fine; could you elaborate as to what is broken/fails?

simpcomp-team:
I tried it before and it did not compile. But just tried it again and now it worked. So please just ignore this request.

PR #2937, opened 2018-10-23
New functions in libgap-api

sebasguts:


src/libgap-api.c
```diff
@@ -60,3 +60,27 @@ Obj GAP_EvalString(const char * cmd)
     res = READ_ALL_COMMANDS(instream, False, True, viewObjFunc);
     return res;
 }
+
+//
+// Returns the GAP object containing
+// <string>
+Obj GAP_MakeString(char * string)
```
fingolfin:
```suggestion
Obj GAP_MakeString(const char * string)
```

src/libgap-api.h
```diff
@@ -17,5 +18,8 @@ void GAP_Initialize(int          argc,
 
 Obj GAP_ValueGlobalVariable(const char * name);
 Obj GAP_EvalString(const char * cmd);
+Obj GAP_MakeString(char * string);
```
```suggestion
Obj GAP_MakeString(const char * string);
```

src/libgap-api.h
```diff
@@ -4,6 +4,7 @@
 #define LIBGAP_API_H
 
 #include "gap.h"
+#include "plist.h"
```
fingolfin: Why include this here? It should be included in the C file only, I'd say, precisely to avoid exposing these internals to users of libgap, no?

sebasguts: Oh yes, this is simply the wrong file. Thank you :)

fingolfin:
I guess you could squash this into one commit (or we could squash-merge it)?

sebasguts:
You can squash-merge it (the multiple commits come from the suggestion feature). Afterwards, I can add more functions.

sebasguts:
You can squash-merge it (the multiple commits come from the suggestion feature). Afterwards, I can add more functions.

PR #2944, opened 2018-10-25
Make INT_INTOBJ always check its argument

ChrisJefferson:
This commit makes INT_INTOBJ always check it is given an INTOBJ, and produce an error if not. At the moment we assert in this case in debug mode, and produce nonsense in normal execution.

This might seem like an inefficiency, as this is a really high-use function. However, I'm having trouble measuring any measurable slowdown on any GAP code I can write. I have in this PR a special test which hits what I imagine is the very worst case (just reading loads of integers, one after the other).

```
x := List([1..10000000], y -> (y mod 100));;
TEST_SUM_LIST(x);
time;
```

This has a 7% slowdown (and as I say, this is code which does absolutely no useful work).

This will fix some bugs I found today (including `x := [1,,,3]; y := x{[2^2000]};`), and also in general deal with what has been a continuous source of difficult to track down crashes and bugs going back over the whole time I've worked on GAP. This will allow a bunch of tests to be deleted (I'm happy to go through and remove them, if we will merge this).

The `__builtin_unlikely`, if you haven't seen it before, it a piece of code built into gcc and clang which tells the compiler that this value will (in this case) probably be true, and the optimiser uses this accordingly. We could add some configure magic if anyone is worried about a compiler that doesn't support this (but gcc and clang both do).

### DIFF
--- a/src/intobj.h
+++ b/src/intobj.h
@@ -33,6 +33,7 @@ a
 #define GAP_INTOBJ_H
 
 #include "system.h"
+#include "error.h"
 
 #ifdef SYS_IS_64_BIT
 #define NR_SMALL_INT_BITS  (64 - 4)
@@ -111,12 +112,18 @@ static inline Int ARE_INTOBJS(Obj o1, Obj o2)
  * chooses to do a logical right shift. */
 static inline Int INT_INTOBJ(Obj o)
 {
-    GAP_ASSERT(IS_INTOBJ(o));
+    if(__builtin_expect(IS_INTOBJ(o),1))
+    {
 #ifdef HAVE_ARITHRIGHTSHIFT
-    return (Int)o >> 2;
+        return (Int)o >> 2;
 #else
-    return ((Int)o - 1) / 4;
+        return ((Int)o - 1) / 4;
 #endif
+    }
+    else
+    {
+        ErrorQuitIntSmall(o);
+    }
 }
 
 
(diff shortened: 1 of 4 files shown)


src/intobj.h
```diff
@@ -111,12 +112,18 @@ static inline Int ARE_INTOBJS(Obj o1, Obj o2)
...
-    return (Int)o >> 2;
+        return (Int)o >> 2;
 #else
-    return ((Int)o - 1) / 4;
+        return ((Int)o - 1) / 4;
 #endif
+    }
+    else
+    {
+        ErrorQuitIntSmall(o);
```
fingolfin: You could now use `RequireSmallInt` here, and get rid of `ErrorQuitIntSmall`.

Also: please clang-format.

ChrisJefferson:
I don't think code coverage really helps here. The most common problem tends to be having a place which accepts GMP hints, then trying to treat them as immediate mode ints. Usually we don't test every function with arguments big enough to catch this.

The question is, how many places do we think this problem occurs without validation? I'd bet lots.

I don't mind adding things like GetSmallInt (which we should make sure doesn't check twice), but I worry there are too many little used faulty functions. I bet I could find 3 without much difficulty.

fingolfin:
All in all, I am not keen on this change and the deoptimization it presents. I am simply not convinced about the potential gain vs what we sacrifice.

ChrisJefferson:
I'll keep it in my copy of GAP, if it doesn't have general interest.

PR #2947, opened 2018-10-28
Unify kernel error checking

fingolfin:
This contains PR #2946 (mainly for historical reasons: when I started writing this a couple months ago, I first only wanted to clean up some error messages, and then it kept growing... and now untangling it would have been just too much effort for no clear gain).

Some notes: `RequireArgumentCondition` and `RequireArgumentConditionMayReplace` are implemented as macros because that was super convenient when I started out. Part of them will likely stay so, because it makes it super easy to test for arbitrary conditions (as C has no lambda functions)... But I am now considering to reimplement at least the part which generates the error message as a separate function. One reason for that is that I am contemplating to replace the `(not a %s)` part of those errors by something slightly better:
* if the "bad" object is an integer, instead of `(not a integer)` (sic) let's show `(not the integer 123)`; this improves awkward messages like "most be a positive integer (not a integer)".
* if the "bad" object is a boolean, replace `(not a boolean or fail)` by one of
  * `(not the boolean true)` or `(not true)` (preferences, anybody)
  *  `(not the boolean false)` or `(not false)`,
  *  `(not fail)`
* ... anything else? well, it'll be trivial to add

Finally, note that `RequirePositiveSmallInt` takes three arguments, while e.g. `RequireSmallList` only takes two. Here's an example for the former:
```C
  RequirePositiveSmallInt("SET_SCOBJ_MAX_STACK_SIZE", size, "size");
```
For `RequireSmallList`, the second and third argument were merged. For `RequirePositiveSmallInt` this was not always possible:
```C
  RequirePositiveSmallInt("PosObj Element", pos, "position");
```
On the long run, I'd consider changing the code to use the same name for the argument as for the C variable, and then drop the 3rd argument to `RequirePositiveSmallInt` (which would be derivable from the 2nd). But I did not want to introduce even more variation to the the code than this PR already does.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1195,12 +1195,7 @@ Obj FuncPositionNthTrueBlist (
             (Int)TNAM_OBJ(blist), 0L,
             "you can replace <blist> via 'return <blist>;'" );
     }
-    while ( ! IS_POS_INTOBJ(Nth) ) {
-        Nth = ErrorReturnObj(
-            "Position: <nth> must be a positive small integer (not a %s)",
-            (Int)TNAM_OBJ(Nth), 0L,
-            "you can replace <nth> via 'return <nth>;'" );
-    }
+    RequirePositiveSmallIntMayReplace("Position", Nth, "nth");
     
     nrb = NUMBER_BLOCKS_BLIST(blist);
     if ( ! nrb )  return Fail;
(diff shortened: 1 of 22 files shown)


fingolfin:
I also have a commit which improves the error messages as described above (for inputs that are booleans or integers), but I think I'll put it into a separate PR, as that is a bigger change of behavior visible from the outside, and could even be considered relevant for the release notes.

ChrisJefferson:
While there are some possible future changes I might want to see (for example, depending on conversations, removing most occurrences of `mayReplace`, that will be much easier to do on top of this PR anyway.

PR #2948, opened 2018-10-28
Revision of ApplicableMethod

cdwensley:
This PR attempts to address issue #2871. 
The following aims have been addressed (in no particular order):  
 - stick to the existing 'magic constants' (except verbosity 6) rather than introduce 'options'; 
 - when a function, rather than an operation, is given, do not raise an error - return the function; 
 - improve the layout of the #I messages; 
 - this is a function which may well be used by non-expert users, so the Reference Manual needs a selection of examples; 
 - the file tst/testinstall/methwhy.tst required extensive revision.
Undoubtedly more testing is required!

### DIFF
--- a/lib/methwhy.g
+++ b/lib/methwhy.g
@@ -350,6 +350,12 @@ local oper,opargs,nopargs,verbos,fams,flags,i,j,methods,flag,flag2,
         Print(" at ",LocationFunc(oper));
       fi;
       Print("\n");
+      ## not sure what this test is for
+      if not ( fams=fail or CallFuncList(m.famPred,fams) ) then 
+        if verbos>2 then 
+          Print("#I   - bad family relations\n"); 
+        fi;
+      fi;
     elif verbos>1 then
       Print("#I  Method ",i); 
       Print(", value: ");  
--- a/lib/methwhy.g
+++ b/lib/methwhy.g
@@ -170,7 +170,7 @@ end);
 ##  #I  ``SplitString: for a string and two characters''
 ##  #I   at GAPROOT/lib/string.gi:545
 ##  #I   - 2nd argument needs [ "IsChar" ]
-##  [ function( string, d1, d2 ) ... end ]
+##  [ function( string, seps, d ) ... end ]
 ##  ]]></Log>
 ##  When a method returned by <Ref Func="ApplicableMethod"/> is called 
 ##  it returns either the desired result or the string
@@ -280,6 +280,12 @@ local oper,narg,args,verbos,fams,flags,i,j,methods,flag,flag2,
       fi;
       flag := flag and flag2;
     od;
+    if not ( fams=fail or CallFuncList(m.famPred,fams) ) then 
+      flag := false; 
+      if verbos>2 then 
+        Print("#I   - bad family relations\n"); 
+      fi;
+    fi;
     if flag then 
       valid[i] := true; 
       Add( applic, i );;
@@ -357,12 +363,6 @@ local oper,narg,args,verbos,fams,flags,i,j,methods,flag,flag2,
         Print(" at ",LocationFunc(oper));
       fi;
       Print("\n");
-      ## not sure what this test is for
-      if not ( fams=fail or CallFuncList(m.famPred,fams) ) then 
-        if verbos>2 then 
-          Print("#I   - bad family relations\n"); 
-        fi;
-      fi;
     elif verbos>1 then
       Print("#I  Method ",i); 
       Print(", value: ");  
--- a/lib/methwhy.g
+++ b/lib/methwhy.g
@@ -95,7 +95,7 @@ end);
 ##  Since <C>ApplicableMethod</C> returns a function, <C>Print(last);</C> 
 ##  may be used to view the code. 
 ##  <P/>
-##  If the first argument is a function, rather than an operation, 
+##  If the first argument is a function, but not an operation, 
 ##  then the function is returned. 
 ##  <P/>
 ##  The first example shows that there are 50 methods for <C>Size</C> 
@@ -105,12 +105,12 @@ end);
 ##  gap> ApplicableMethod( Size, [s4], 1, "all" );
 ##  #I  Searching Method for Size with 1 arguments:
 ##  #I  Total: 50 entries, of which 3 are applicable:
-##  #I  Method 6, valid operation number 1, value: 103
-##  #I  ``Size'' at ... some path .../lib/coll.gi:176
-##  #I  Method 10, valid operation number 2, value: 59
-##  #I  ``Size: for a permutation group'' at ... some path .../lib/grpperm.gi:483
-##  #I  Method 50, valid operation number 3, value: 2
-##  #I  ``Size: for a collection'' at ... some path .../lib/coll.gi:189
+##  #I  Method 6, applicable method number 1, value: 103
+##  #I  ``Size'' at GAPROOT/lib/coll.gi:176
+##  #I  Method 10, applicable method number 2, value: 59
+##  #I  ``Size: for a permutation group'' at GAPROOT/lib/grpperm.gi:483
+##  #I  Method 50, applicable method number 3, value: 2
+##  #I  ``Size: for a collection'' at GAPROOT/lib/coll.gi:189
 ##  [ function( C ) ... end, function( G ) ... end, function( C ) ... end ]
 ##  gap> Print( last[1], "\n" );                  
 ##  function ( C )
@@ -122,30 +122,30 @@ end);
 
 ##  ]]></Log>
 ##  The second example shows that for <C>DirectProduct</C>, 
-##  which is a function the location may be displayed and the code returned. 
-##  For <C>DirectProductOp</C> with verbosity <M>2</M> and <M>nr=2</M> 
+##  which is a function, the location may be displayed and the code returned. 
+##  For <C>DirectProductOp</C>, with verbosity <M>2</M> and <M>nr=2</M>, 
 ##  methods <M>1</M> and <M>5</M> are applicable, and all methods numbered 
 ##  <M>[2..5]</M> are displayed. 
 ##  <Log><![CDATA[
 ##  gap> ApplicableMethod( DirectProduct, [s4,s4], 1, 1 );    
 ##  #I  DirectProduct is a function, not an operation, located at:
-##  #I  ... some path .../lib/gprd.gi:17
+##  #I  GAPROOT/lib/gprd.gi:17
 ##  function( arg... ) ... end
 ##  gap> ApplicableMethod( DirectProductOp, [[s4,s4],s4], 2, 2 );
 ##  #I  Searching Method for DirectProductOp with 2 arguments:
 ##  #I  Total: 5 entries, of which 2 are applicable:
 ##  #I  Method 2, value: 66
 ##  #I  ``DirectProductOp: for a list (of pc groups), and a pc group''
-##  #I   at ... some path .../lib/gprdpc.gi:15
+##  #I   at GAPROOT/lib/gprdpc.gi:15
 ##  #I  Method 3, value: 47
 ##  #I  ``DirectProductOp: matrix groups''
-##  #I   at ... some path .../lib/gprdmat.gi:70
+##  #I   at GAPROOT/lib/gprdmat.gi:70
 ##  #I  Method 4, value: 40
 ##  #I  ``DirectProductOp: for a list of fp groups, and a fp group''
-##  #I   at ... some path .../lib/grpfp.gi:5495
-##  #I  Method 5, valid operation number 2, value: 37
+##  #I   at GAPROOT/lib/grpfp.gi:5495
+##  #I  Method 5, applicable method number 2, value: 37
 ##  #I  ``DirectProductOp: for a list (of groups), and a group''
-##  #I   at ... some path .../lib/gprd.gi:50
+##  #I   at GAPROOT/lib/gprd.gi:50
 ##  function( list, gp ) ... end
 ##  ]]></Log>
 ##  The third example shows the sort of output that can be obtained 
@@ -157,23 +157,23 @@ end);
 ##  #I  Total: 4 entries, of which 1 is applicable:
 ##  #I  Method 1, value: 15
 ##  #I  ``SplitString: for three strings''
-##  #I   at ... some path .../lib/string.gi:539
+##  #I   at GAPROOT/lib/string.gi:539
 ##  #I   - 3rd argument needs [ "IsString" ]
 ##  #I  Method 2, value: 11
 ##  #I  ``SplitString: for a string, a character and a string''
-##  #I   at ... some path .../lib/string.gi:561
+##  #I   at GAPROOT/lib/string.gi:561
 ##  #I   - 2nd argument needs [ "IsChar" ]
 ##  #I   - 3rd argument needs [ "IsString" ]
-##  #I  Method 3, valid operation number 1, value: 11
+##  #I  Method 3, applicable method number 1, value: 11
 ##  #I  ``SplitString: for two strings and a character''
-##  #I   at ... some path .../lib/string.gi:553
+##  #I   at GAPROOT/lib/string.gi:553
 ##  #I  Method 4, value: 7
 ##  #I  ``SplitString: for a string and two characters''
-##  #I   at ... some path .../lib/string.gi:545
+##  #I   at GAPROOT/lib/string.gi:545
 ##  #I   - 2nd argument needs [ "IsChar" ]
 ##  [ function( string, d1, d2 ) ... end ]
 ##  ]]></Log>
-##  When a method returned by <Ref Func="ApplicableMethod"/> is called then
+##  When a method returned by <Ref Func="ApplicableMethod"/> is called 
 ##  it returns either the desired result or the string
 ##  <C>"TRY_NEXT_METHOD"</C>, which corresponds to a call to
 ##  <Ref Func="TryNextMethod"/> in the method and means that
@@ -197,7 +197,7 @@ end);
 ##  <#/GAPDoc>
 ##
 BIND_GLOBAL("ApplicableMethodTypes",function(arg)
-local oper,opargs,nopargs,verbos,fams,flags,i,j,methods,flag,flag2,
+local oper,narg,args,verbos,fams,flags,i,j,methods,flag,flag2,
       m,nam,val,has,need,isconstructor,
       nummeth,valid,applic,numapplic,nr,first,last;
   if Length(arg)<2 or not IsList(arg[2]) or not IsFunction(arg[1]) then
@@ -206,8 +206,8 @@ local oper,opargs,nopargs,verbos,fams,flags,i,j,methods,flag,flag2,
   ## process the arguments 
   oper:=arg[1];
   isconstructor:=IS_CONSTRUCTOR(oper);
-    opargs:=arg[2];
-  nopargs:=Length(opargs);
+    args:=arg[2];
+  narg:=Length(args);
   verbos:=0;
   if Length(arg)>2 and IsPosInt(arg[3]) then
     verbos:=Minimum(arg[3],4);
@@ -220,8 +220,8 @@ local oper,opargs,nopargs,verbos,fams,flags,i,j,methods,flag,flag2,
       nr:=0;
     fi;
   fi;
-  ## accept the name of a function 
-  if not oper in OPERATIONS then 
+  ## if not an operation, just return the function 
+  if not IsOperation(oper) then 
     nam:=FilenameFunc(oper); 
     if nam=fail then 
       Print("#I  ",oper," is not recognised as the name of an operation\n"); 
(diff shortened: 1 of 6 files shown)


lib/methwhy.g
```diff
@@ -158,108 +244,172 @@ local oper,narg,args,skip,verbos,fams,flags,i,j,methods,flag,flag2,
...
+      Print("\n#I  ``",nam,"''\n");
+      if IsBound(m.location) then
+        Print("#I   at ", m.location[1], ":", m.location[2]);
+      elif LocationFunc(oper) <> "" then
+        Print(" at ",LocationFunc(oper));
       fi;
-      j:=j+1;
-    od;
-    if flag then
-      if fams=fail or CallFuncList(m.famPred,fams) then
```
hulpke: it looks as if the test for the family predicate has been removed and not been replaced properly.

lib/methwhy.g
```diff
@@ -89,12 +89,89 @@ end);
...
 ##  <P/>
+##  Since <C>ApplicableMethod</C> returns a function, <C>Print(last);</C> 
+##  may be used to view the code. 
+##  <P/>
+##  If the first argument is a function, rather than an operation, 
+##  then the function is returned. 
+##  <P/>
+##  The first example shows that there are 50 methods for <C>Size</C> 
+##  currently available, or which 3 are applicable to <C>s4</C>. 
+##  <Example><![CDATA[
```
ChrisJefferson: I would change this Example into a Log, just to avoid having to continuously update it as code moves around (Log is like Example, except it isn't checked by make manualexamples)

lib/methwhy.g
cdwensley: This test appears to be needed only for methods which are applicable.  Do you know of an example which triggers the Print("#I   - bad family relations\n"); warning?

lib/methwhy.g
cdwensley: Ah - I'm not particularly aware of "make manualexamples", but wondered whether this was liable too happen.  Will change.

cdwensley: Now that these are <Log> ... </Log>'s, is it better to use "GAPROOT" rather than "... some path ..."?

lib/methwhy.g
```diff
@@ -89,12 +89,89 @@ end);
...
-##  <Item>
-##      also the function body of the selected method(s)
-##  </Item>
 ##  </List>
 ##  <P/>
+##  Since <C>ApplicableMethod</C> returns a function, <C>Print(last);</C> 
+##  may be used to view the code. 
+##  <P/>
+##  If the first argument is a function, rather than an operation, 
+##  then the function is returned.
```
fingolfin: I wonder what the motivation for this is. Wouldn't it be better to have an error that says that ApplicableMethod may only be used on operations, but not on plain functions?

fingolfin: BTW, note that operations are also functions, so I find the current sentence rather confusing.

lib/methwhy.g
```diff
@@ -119,35 +196,48 @@ end);
...
+      nr:=0;
     fi;
-    erg:=[];
-  else
-    skip:=0;
   fi;
-  narg:=Length(args);
-
+  ## accept the name of a function 
+  if not oper in OPERATIONS then
```
fingolfin: How about using IsOperation instead.

lib/methwhy.g
```diff
@@ -119,35 +196,48 @@ end);
...
+    elif arg[4]="all" then 
+      nr:=0;
     fi;
-    erg:=[];
-  else
-    skip:=0;
   fi;
-  narg:=Length(args);
-
+  ## accept the name of a function
```
fingolfin: That suggests it accepts a string, but that's not what this code is about, I think.

lib/methwhy.g
```diff
@@ -119,35 +196,48 @@ end);
...
     fi;
-    erg:=[];
-  else
-    skip:=0;
   fi;
-  narg:=Length(args);
-
+  ## accept the name of a function 
+  if not oper in OPERATIONS then 
+    nam:=FilenameFunc(oper);
```
fingolfin: Perhaps also check for the case wher oper is not a function? And add a test for this, too?

tst/testinstall/methwhy.tst
```diff
@@ -1,8 +1,10 @@
-gap> START_TEST("methwy.tst");
+gap> START_TEST("methwhy.tst");
 
 #
 gap> ApplicableMethod();
-Error, usage: ApplicableMethod(<opr>,<arglist>[,<verbosity>[,<nr>]])
+#I  ApplicableMethod requires at least two arguments
+#I  usage: ApplicableMethod(<opr>,<arglist>[,<verbosity>[,<nr>]])
```
fingolfin: Hmm, so this function now handles argument validation differently from (almost?) any other - why?

tst/testinstall/methwhy.tst
```diff
@@ -35,52 +37,90 @@ function ( x )
 end
 gap> Display(ApplicableMethod(foobar, [ ['a'] ], 1));
 #I  Searching Method for foobar with 1 arguments:
-#I  Total: 4 entries
-#I  Method 1: ``foobar: for a string'' at stream:1 , value: 2*SUM_FLAGS+1
+#I  Total: 4 entries, of which 3 are applicable:
+#I  Method 1, valid operation number 1, value: 2*SUM_FLAGS+1
```
fingolfin: Valid = applicable? Why use another word?

tst/testinstall/methwhy.tst
```diff
@@ -35,52 +37,90 @@ function ( x )
...
+#I  ``foobar: for a string''
+#I   at stream:1
 function ( x )
     return StartsWith( x, "foo" );
 end
-gap> Display(ApplicableMethod(foobar, [ ['a'] ], 1, 2));
+gap> ApplicableMethod(foobar, 'a', 1);
```
fingolfin: Why was the original test replaced by another? It is of course fine to add new tests, but why remove the old?

tst/testinstall/methwhy.tst
```diff
@@ -35,52 +37,90 @@ function ( x )
...
-#I  Total: 4 entries
-#I  Method 4: ``foobar'' at stream:1 , value: 0
+#I  Total: 4 entries, of which 1 is applicable:
+#I  Method 4, valid operation number 1, value: 0
+#I  ``foobar''
+#I   at stream:1
 function ( x )
     return fail;
 end
-gap> Display(ApplicableMethod(foobar, [fail], 4));
```
fingolfin: Why this change?

lib/methwhy.g
cdwensley: It seems best to avoid ApplicableMethod throwing an error - just return 'fail' as appropriate. 
The occasional user may not realise that the function they are interested in (e.g. DirectProduct) is not an operation, so it is helpful to return that function so that a quick look at the listing by the user shows what operation(s) is called (e.g. DirectProductOp).
"rather than an operation" now changed to "but not an operation".

lib/methwhy.g
cdwensley: OK.

lib/methwhy.g
cdwensley: Now "if not an operation, just return the function"

lib/methwhy.g
cdwensley: The check at the start of the ApplicableMethodTypes includes "... or not IsFunction(arg[1]) ..." and an error is issued.  ApplicableMethod contains the lines: 
      elif not IsFunction(arg[1]) then 
           Print("#I  argument 1 must be the name of an operation\n" );

tst/testinstall/methwhy.tst
cdwensley: Examples with "..., 1, 3)",  "..., 1, 4)",  "..., 1, "all"). immediately follow, so the "..., 1, 2)" case seemed unnecessary.

tst/testinstall/methwhy.tst
cdwensley: The previous example: Display(ApplicableMethod(foobar, [fail], 1));  displays the function, so there seemed no need to display it again, and: ApplicableMethod(foobar, [fail], 4); gives the shortened version "function( x ) ... end".

lib/methwhy.g
```diff
@@ -119,35 +192,48 @@ end);
...
-  narg:=Length(args);
-
+  ## accept the name of a function 
+  if not oper in OPERATIONS then 
+    nam:=FilenameFunc(oper); 
+    if nam=fail then 
+      Print("#I  ",oper," is not recognised as the name of an operation\n"); 
+      return fail;
+    else
+      if verbos>0 then
```
stevelinton: Should we also be checking that `oper` is capable of accepting the given number of arguments -- so that `NumberArgumentsFunction(oper)` returns either the given number or -1? There isn't much point in returning the function if it doesn't.

cdwensley: Agreed - and added some tests at the end of methwhy.tst.

cdwensley:
Can anyone help with the errors now reported from tst/testinstall/coll.tst? 
This tests the function in tst/testinstall/testenumerator.g. 
The first error arises from the following commands: 
     gap> M2:=Magma(0, 1);;
     gap> enum := Enumerator(M2);;
     gap> TestConsistencyOfEnumeratorByFunctions(enum);
and the relevant check in test enumerator.g is: 
    filter:= IsEnumeratorByFunctions;
    elm:= enum[1];
    if not IsIdenticalObj( ApplicableMethod( \in, [ elm, enum ] ),
               ApplicableMethodTypes( \in, [ IsObject, filter ] ) ) then
      Print( "#E  wrong `\\in' method\n" );
      result:= false;
    fi;
Running these (revised) ApplicableMethods gives the following, 
where the first method for meth2 is the second method for meth1:
gap> meth1:=ApplicableMethod(\in,[elm,enum],1,"all");
#I  Searching Method for in with 2 arguments:
#I  Total: 74 entries, of which 3 are applicable:
#I  Method 2, applicable method number 1, value: 1*SUM_FLAGS+2
#I  ``in: for wrong family relation''
#I   at /Applications/gap/gapdev/lib/list.gi:227
#I  Method 69, applicable method number 2, value: 7
#I  ``in: for an enumerator that perhaps has its own membership test function''
#I   at /Applications/gap/gapdev/lib/coll.gi:506
#I  Method 73, applicable method number 3, value: 2
#I  ``in: for an object, and a list''
#I   at /Applications/gap/gapdev/lib/list.gi:239
[ function( arg... ) ... end, function( elm, enum ) ... end, 
  function( elm, list ) ... end ]
gap> meth2:=ApplicableMethodTypes(\in,[IsObject,filter],1,"all");
#I  Family predicate cannot be tested
#I  Searching Method for in with 2 arguments:
#I  Total: 74 entries, of which 2 are applicable:
#I  Method 69, applicable method number 1, value: 7
#I  ``in: for an enumerator that perhaps has its own membership test function''
#I   at /Applications/gap/gapdev/lib/coll.gi:506
#I  Method 73, applicable method number 2, value: 2
#I  ``in: for an object, and a list''
#I   at /Applications/gap/gapdev/lib/list.gi:239
[ function( elm, enum ) ... end, function( elm, list ) ... end ]

tst/testinstall/methwhy.tst
ChrisJefferson: I'm not entirely clear what you mean here. I understand returning fail when there is no method to pick, but it seems strange to return fail when the arguments just weren't valid (as then someone might get fail, because they mistyped the arguments).

cdwensley:
To my mind ApplicableMethod is unlike most functions in GAP, so an unusual behaviour is reasonable.  Hence the attempt to avoid raising an error when this function is used.  Comparing the output from your example with this PR, and without it, we get: 

```
gap> g := Group(());
Group(())
gap> ApplicableMethod(Size,g);
#I  argument 2 must be a list of arguments for the operation
#I  usage: ApplicableMethod(<opr>,<arglist>[,<verbosity>[,<nr>]])
fail
gap> Reread("methwhy.orig");
gap> ApplicableMethod(Size,g);
Error, usage: ApplicableMethod(<opr>,<arglist>[,<verbosity>[,<nr>]]) at methwhy.orig:256 called from
<function "ApplicableMethod">( <arguments> )
 called from read-eval loop at *stdin*:4
you can 'return;' to continue
brk> 
```

In the latter case a "quit;" is needed before editing the previous line, which seems unnecessary. 
Would it be a good idea to deal with the case where just two arguments are given, and the second is not a list, to replace the second argument with a one-element list and then continue, so that your example would no longer fail?

ChrisJefferson:
I'm not really sure why you think ApplicableMethod is so unusual. I mean, Group returns a group, List returns a list, ApplicableMethod returns an applicable method, and they all (at the moment) enter the break loop if you give them invalid arguments.

You could wish that GAP didn't enter the break loop so eagerly, but that would be a seperate discussion and should be done globally (you can stop GAP entering the break loop globally if you don't want it).

tst/testinstall/methwhy.tst
fingolfin: I agree with @ChrisJefferson

tst/testinstall/methwhy.tst
fingolfin: Well, the idea was to test *all* possible ways to call this function. Why would one randomly skip tests for some ways to call it?

tst/testinstall/methwhy.tst
fingolfin: But the point was precisely to test that `..., 4)` works as intended. That's why this test was there. You seem to be saying we don't need to test it, because we know that it works -- but then we could remove most tests with the same argument, couldn't we?

stevelinton:
The issue about `fail` vs break comes down to whether you see `ApplicableMethod` exclusively as a debugging tool, only ever used interactively, or as something that someone might programme with for some reason (even if only to build better debugging tools). I think the spirit of this PR is the first one, but @ChrisJefferson is at least considering the second.    

I would argue that it can quite sensibly be programmed with, and so should give an error on invalid input. Experience suggests that everything that can be programmed will be eventually. If someone really just wants to build an interactive debugging environment for GAP libraries, that anticipates and responds sensibly to erroneous interactive user input, that would be amazing, but I think `ApplicableMethod` is a lower level tool than that.

tst/testinstall/methwhy.tst
cdwensley: The removed(changed) examples have now been reinstated, and others added.

cdwensley:
Hopefully most of the recent comments have been covered by the latest commit.
(Reluctantly) ApplicableMethod now gives an Error when 0 or 1 argument is supplied. 
The example of @ChrisJefferson now gives:

```
gap> g:=Group(());;
gap> Display(ApplicableMethod(Size,g));
#I  replacing second argument arg2 by the list [arg2]
function ( C )
    if IsFinite( C ) then
        TryNextMethod();
    fi;
    return infinity;
end
```

lib/methwhy.g
```diff
@@ -119,31 +195,52 @@ end);
...
-  else
-    skip:=0;
   fi;
-  narg:=Length(args);
-
+  ## if not an operation, just return the function 
+  if not IsOperation(oper) then 
+    nam:=FilenameFunc(oper); 
+    if nam=fail then 
+      Print("#I  ",oper," is not recognised as the name of an operation\n");
```
ChrisJefferson: In general (I know not always) lines starting `#I` are produced using `Info`, not directly. This makes it easy to turn them off.

You could either (for me) make a new Info level, or use (for example) `Info(InfoWarning, 1, oper, " is not recognised as the name of an operation")`.

Also not that in general printing out arguments to a function is often not a good idea, if someone passes in a GAP function for example, printing it will print out the entire function body.

cdwensley:
Added a collection of recent GAPdev commits, which was probably a mistake. 
Extra test added to ApplicableMethod to check that the number of entries in the second parameter is not greater than 8.  This aims to address the problem reported in Issue #3116.  A test was also added.

fingolfin:
Yes, that was correct.

tst/testinstall/methwhy.tst
```diff
@@ -91,124 +136,207 @@ function ( x, y... )
...
-#I  Method 1: ``foobar: for 6 integers'' at stream:6, value: 0
-function ( x, y... )
-    return 1;
+#I  Total: 1 entries, of which 1 is applicable:
+#I  Method 1, applicable method number 1, value: 0
+#I  ``foobar: for 6 integers''
+#I   at stream:6
+function( x, y... ) ... end
+gap> ApplicableMethod(foobar, [1..7], 1);
+#I  no method found for this operation with this number of arguments
```
fingolfin: This test is failing (as can be seen in the Travis logs, or by running the `testinstall` locally), because it actually prints this error now:
```
Error, METHODS_OPERATION: <narg> must be an integer between 0 and 6 (not the integer 7)
```

lib/methwhy.g
```diff
@@ -90,13 +90,89 @@ end);
...
 ##  </Item>
-##  <Mark>6</Mark>
-##  <Item>
-##      also the function body of the selected method(s)
-##  </Item>
 ##  </List>
 ##  <P/>
-##  When a method returned by <Ref Func="ApplicableMethod"/> is called then
+##  Since <C>ApplicableMethod</C> returns a function, <C>Print(last);</C> 
+##  may be used to view the code.
```
fingolfin: Sure, but if I want to see the code for *all* the methods, then this is not a substitute. E.g. if I do this: `ApplicableMethod(ViewObj, [ "test" ], 6, 10);` then it'll print several functions.

tst/testinstall/methwhy.tst
```diff
@@ -1,9 +1,11 @@
-gap> START_TEST("methwy.tst");
+gap> START_TEST("methwhy.tst");
 
 #
 gap> ApplicableMethod();
+#I  ApplicableMethod requires at least two arguments
```
fingolfin: I really don't like this. No other function produces such "hint messages". In this case, I also don't see any value it adds over the usage error message below.

tst/testinstall/methwhy.tst
```diff
@@ -1,9 +1,11 @@
...
 Error, usage: ApplicableMethod(<opr>,<arglist>[,<verbosity>[,<nr>]])
 
+
```
fingolfin: Why is there an extra empty line now? Is this an accident, or does `ApplicableMethod` really print it? Either variant seems like a bug (or misfeature) to me, though... :/

tst/testinstall/methwhy.tst
```diff
@@ -35,52 +37,95 @@ function ( x )
...
-#I  Method 3: ``foobar: for an integer'' at stream:1, value: 1
+#I  Method 3, value: 1
+#I  ``foobar: for an integer''
+#I   at stream:1
 #I   - 1st argument needs [ "IsInt" ]
-#I  Method 4: ``foobar'' at stream:1, value: 0
-#I  Function Body:
-function ( x )
-    return fail;
-end
```
fingolfin: So it seems the `6` option really was killed. But why?

tst/testinstall/methwhy.tst
```diff
@@ -91,124 +136,207 @@ function ( x, y... )
...
 #I   - 1st argument needs [ "IsList" ]
-#I  Method 3: ``foobar: for an integer'' at stream:1, value: 1
+#I  Method 3, applicable method number 1, value: 1
+#I  ``foobar: for an integer''
+#I   at stream:1
 function ( x )
     return 1;
 end
-gap> Display(ApplicableMethod(foobar, [1,2], 5));
+gap> ApplicableMethod(foobar, [1], 4, 2);
```
fingolfin: Dropping the `Display` here and elsewhere in the tests has two major disadvantages:
* having the function printed served as another verification that we *really* show the right function; this is now lost;
* it results in a larger diff, which makes reviewing more painful; and I think the huge and hard to read diff is a major reason your PR has not received more attention (it certainly is the reason why I don't really feel like reviewing it).

tst/testinstall/methwhy.tst
```diff
@@ -91,124 +136,207 @@ function ( x, y... )
...
+#I  and is located at: stream:1
+function( x, arg... ) ... end
+gap> ApplicableMethod( foobar, [9], 1, 1 );  
+#I  foobar is a function, not an operation
+#I  and requires at least 2 arguments
+fail
+
+#
+# now not returning 'fail' when the second argument fails to be a list, 
+# but replacing arg2 by [arg2] and then carrying on:
```
fingolfin: I am conflicted about this. While I've made the mistake myself in the past, it usually was then trivial to correct. It'd be easier if there was a good error message, of course ("<args> must be a list, not an integer/a boolean/a ..."). But by allowing it, we allow users to learn that this is OK. But once they accidentally write `list` instead of `[list]`, they again get a wrong output; but now we are again in a potentially confusing and hard to recover from situation.

So I think I personally would prefer if this gave a helpful error, instead of accepting the wrong input in some cases. Anyway, I certainly won't veto this PR over this minor bit.

cdwensley:
I believe the methwhy.tst now fails because PR #3116 has been merged (though I cannot find it in the list of closed PRs) but has not been pulled to my setup.  I agree that this PR has become a mess and that the best way to proceed is to close this and start again with a set of smaller changes.  Whether or not I want to devote any more time to it remains to be seen.

PR #2951, opened 2018-10-29
set `IsSSortedList` in results of `Positions` etc.

ThomasBreuer:
- The global functions `Positions` and `PositionsBound`
  set the filter `IsSSortedList` for their result.
- The two `PositionsProperty` methods in `lib/list.gi`
  set the filter `IsSSortedList` for their result.
- The documentation of `Positions`and `PositionsProperty`
  now mentions that the result is a set.
- The documentation of `PositionsBound` now states that it is
  a global function not an operation.
- The test file `tst/testinstall/list.tst` now checks
  that `IsSSortedList` is set in these situations.

(It would be easy to set `IsSSortedList` also for the result of
`Filtered` if the given list has `IsSSortedList`.
Would that be reasonable?)

### DIFF
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -1342,10 +1342,15 @@ InstallGlobalFunction( Filtered,
           res[i]:= elm;
         fi;
       od;
-      return res;
     else
-      return FilteredOp( C, func );
+      res:= FilteredOp( C, func );
     fi;
+
+    if HasIsSSortedList( C ) and IsSSortedList( C ) then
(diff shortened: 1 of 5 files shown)


fingolfin:
Re Filtered also setting this filter: sounds ggod to me

ThomasBreuer:
@fingolfin 
As proposed, I have now changed also `Filtered` to set `IsSSortedList` in the result,
provided that the input has `IsSSortedList` set.

PR #2952, opened 2018-10-29
Add command line option --bare to start GAP without any packages

markuspf:
This requries (and includes) #2950.

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -117,6 +117,8 @@ BIND_GLOBAL( "GAPInfo", rec(
            help := [ "Disable the GAP read-evaluate-print loop (REPL)" ] ),
       rec( long := "nointeract", default := false,
            help := [ "Start GAP in non-interactive mode (disable read-evaluate-print loop (REPL) and break loop)" ] ),
+      rec( long := "bare", default := false,
+           help := [ "Attempt to start GAP without even needed packages (developer tool)" ] ),
       ,
       rec( short:= "c", default := "", arg := "<expr>", help := [ "execute the given expression"] ),
     ],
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
While it's yet another .travis.yml line, do we want to test GAP works with `--bare`? (how well does it work? Does it pass testinstall for example?)

markuspf:
Not all tests work (but a suprising number does!). I'll look at adding a "testbare" target.

hulpke:
What is the difference between --bare and -N (for noautoload)?

markuspf:
@hulpke `-N` disables hidden implications, I assume you mean `-A` (no autoloading of suggested packages? `--bare` does not load *any* packages, in particular it does not load `GAPDoc` which is currently the only `NeededPackage`.
Did `-N` use to do something different from what it does now?

alex-konovalov:
If you will look at "testbare" target, you may also start GAP there with `-O` to disable loading obsoletes. This will help to check they are not reappearing in the GAP library.

stevelinton:
What is the reasoning for this? If a package is "needed" GAP is allowed to fail arbitrarily badly without it, including not loading the library. If you need to start GAP without GAPDoc for some reason, then maybe GAPDoc is not needed, just recommended for interactive sessions or something?

stevelinton:
I'm fine with this, but as @wilfwilson says, it should *not* be user-facing, and for the same reason should not be part of the tests that get run automatically on Pull Requests unless we want to change the rules on what it means for a package to be needed. This is a specialist feature for debugging aspects of library and package loading.

wilfwilson:
@stevelinton would you feel if the text describing the `--bare` option (currently `Start GAP without loading any packages`) included some kind of disclaimer? For example:

> Start GAP without loading any packages (experimental)

Or something of that kind?

stevelinton:
Something of the kind. `experimental` suggests that it might be something we plan to roll out though. How about

`Attempt to start GAP without even needed packages (developer tool)`

wilfwilson:
I prefer that, I've made the change.

wilfwilson:
@wucas I think so. I would recommend using the title of this PR (which I just updated) as the text for the release notes.

PR #2953, opened 2018-10-29
a doc typo

dimpase:
PermutationGModule works pefectly well with any field.

### DIFF
--- a/doc/ref/meataxe.xml
+++ b/doc/ref/meataxe.xml
@@ -82,7 +82,7 @@ bases used are needed as well.
 <Func Name="PermutationGModule" Arg='G, F'/>
 
 <Description>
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Would it be possible to add a test to make sure these work (just to make sure it's intentional they work, as opposed to a missing test to forbid them)?

dimpase:
But why? There should not have been any finiteness test in the 1st place...

In fact there used to be a test for finiteness of the field (there and in few other functions from MTX), completely unnecessary, and I had to remove it in computations I did that time.
Certainly MTX needs field to be finite to do MTXing, but it does not need it for things like this, or constructing the tensor product, say.

So it was an issue on an old tracker about it, or even before it existed, and someone took my patches and applied them.

PR #2955, opened 2018-10-30
Improve some kernel error messages

fingolfin:
This continues the work begun in PR #2947 and implements the change I promised there: Instead of "(not a integer)" we now say for example "(not the integer 5)".

Also adds and uses `RequireStringRep`, `RequireInt`, `RequireSmallInt`

### DIFF
--- a/src/error.h
+++ b/src/error.h
@@ -163,6 +163,42 @@ extern Obj RACErrorHelper(int          mayReturn,
 
 /****************************************************************************
 **
+*F  RequireInt
+*/
+#define RequireInt(funcname, op, argname) \
+    RequireArgumentCondition(funcname, op, argname, IS_INT(op), \
+        "must be an integer")
+
+
+/****************************************************************************
+**
+*F  RequireIntMayReplace
+*/
+#define RequireIntMayReplace(funcname, op, argname) \
+    RequireArgumentConditionMayReplace(funcname, op, argname, IS_INT(op), \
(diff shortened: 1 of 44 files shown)


ChrisJefferson:
Picky, I don't really like the messages `(not true)` and `(not false)`, as in context it could be (I think) confusing, consider:

`<x> must be a small list (not true)` -- the `not true` could be referring to the fact the statement is false :) Therefore I'd prefer something else, perhaps `not the value 'true'`, or `not the boolean 'true'` (`value` has the advantage that `not the boolean `fail`, while a valid statement about GAP, also looks silly :) )

src/error.h
```diff
@@ -145,15 +154,49 @@ ErrorReturnVoid(const Char * msg, Int arg1, Int arg2, const Char * msg2);
...
+        "must be an integer")
+
+
+/****************************************************************************
+**
+*F  RequireSmallInt
+*/
+#define RequireSmallInt(funcname, op, argname) \
+    RequireArgumentCondition(funcname, op, argname, IS_INTOBJ(op), \
+        "must be an integer")
```
ChrisJefferson: Should the refer to the "smallness" of the required integer somehow?

fingolfin: Ouch, indeed, that's a copy&past mistake. I'll change it to "must be a small integer"

fingolfin:
How about `(not 'true')`?

PR #2960, opened 2018-10-31
Implement List(iter,func)

fingolfin:
@mohamed-barakat asked for it during the last GAP Days in September 2018, and now again @markuspf, so I polished up the code I wrote for @mohamed-barakat back then and added some tests.

### DIFF
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -836,15 +836,32 @@ InstallOtherMethod( Iterator,
 ##  
 ##  Does not change the iterator.
 ##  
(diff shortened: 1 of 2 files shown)


fingolfin:
Regarding documentation, also `List` for a collection is not correctly documented: it omits the possibility to specify a function which is applied to all elements of the collection. Because `List` has two manual entries, one for lists and one for collections. This is unlike `Filtered`, `Sum`, etc., which all merge this into a single reference manual entry.

markuspf:
I'll just merge this, everything else can be done in later PRs anyway.

PR #2964, opened 2018-11-03
remove broken shifting of gasman stats for package mode

RussWoodroofe:
Gasman supplies information automatically to xgap.  The logic that is currently there appears to be trying to express the numbers in mb rather than kb if larger than 2048mb.  However, it actually gets it backwards and expresses them in bytes.  It also tries to append a 'k' after the closing null character.

This patch removes the shifting altogether.  I'd also be happy to edit to match the shifting done elsewhere more closely (and include the units correctly) if anyone would prefer.

### DIFF
--- a/src/sysmem.c
+++ b/src/sysmem.c
@@ -142,9 +142,6 @@ void SyMsgsBags (
     }
     /* package (window) mode full garbage collection messages              */
     if ( phase != 0 ) {
-      shifted =   3 <= phase && nr >= (1 << 21);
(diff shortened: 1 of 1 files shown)


RussWoodroofe:
@stevelinton, yes, I think I know the code in xgap that you're talking about.  What do you think about patching xgap to always display in MB (but with 1 decimal point)?  That matches more closely what Gap.app is doing, and is probably more useful generally.  I'll go submit a patch over there now!

PR #2965, opened 2018-11-06
Add new plist helper functions

ChrisJefferson:
This adds some new plist helper functions.

1) Functions to make an empty plist, and an empty immutable plist.

2) Functions to make a plist based on a C array, or a variable argument list.

These are then used in various places in the library.

### DIFF
--- a/src/plist.c
+++ b/src/plist.c
@@ -60,6 +60,14 @@
 
 /****************************************************************************
 **
+*V  ImmutableEmptyPlist . . . . . . . . . . . . an immutable empty plain list
+**
+**  'ImmutableEmptyPlist' is an immutable empty plist.
+*/
+Obj ImmutableEmptyPlist;
+
+/****************************************************************************
+**
 *F  GROW_PLIST(<list>,<plen>) . . . .  make sure a plain list is large enough
 **
 */
@@ -1272,7 +1280,7 @@ Obj             ElmsPlist (
 
     /* select no element                                                   */
     if ( LEN_LIST(poss) == 0 ) {
-        elms = NEW_PLIST( T_PLIST_EMPTY, 0 );
+        elms = NewEmptyPlist();
     }
 
     /* general code                                                        */
@@ -1396,7 +1404,7 @@ Obj             ElmsPlistDense (
 
     /* select no element                                                   */
     if ( LEN_LIST(poss) == 0 ) {
-        elms = NEW_PLIST( T_PLIST_EMPTY, 0 );
+        elms = NewEmptyPlist();
     }
 
     /* general code                                                        */
@@ -3433,6 +3441,9 @@ static Int InitKernel (
     /* GASMAN marking functions and GASMAN names                           */
     InitBagNamesFromTable( BagNames );
 
+    InitGlobalBag(&ImmutableEmptyPlist, "src/plist.c:ImmutableEmptyPlist");
+
+
     for ( t1 = T_PLIST;  t1 < T_PLIST_FFE ;  t1 += 2 ) {
         InitMarkFuncBags( t1                     , MarkAllButFirstSubBags );
         InitMarkFuncBags( t1 +IMMUTABLE          , MarkAllButFirstSubBags );
@@ -3844,6 +3855,8 @@ static Int InitLibrary (
     /* init filters and functions                                          */
     InitGVarFiltsFromTable( GVarFilts );
     InitGVarFuncsFromTable( GVarFuncs );
+
+    ImmutableEmptyPlist = NEW_PLIST_IMM(T_PLIST_EMPTY, 0);
 
     /* return success                                                      */
     return 0;
--- a/src/plist.h
+++ b/src/plist.h
@@ -291,6 +291,22 @@ static inline Obj PopPlist(Obj list)
     return val;
 }
 
+
+/****************************************************************************
+**
+*F  NewEmptyPlist() . . . . . . . . . . . . . . create a new empty plain list
+*V  ImmutableEmptyPlist . . . . . . . . . . . . an immutable empty plain list
+**
+**  ImmutableEmptyPlist is a variable rather than a function, as we can
+**  reuse the same immutable empty plist.
+*/
+static inline Obj NewEmptyPlist(void)
+{
+    return NEW_PLIST(T_PLIST_EMPTY, 0);
+}
+
+extern Obj ImmutableEmptyPlist;
+
 /****************************************************************************
 **
 *F  AssPlistEmpty( <list>, <pos>, <val> ) . . . . .  assignment to empty list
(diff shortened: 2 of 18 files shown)


src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+*F  NewImmutableEmptyPlist() . . . . . . . Create a new immutable empty Plist
+*/
+static inline Obj NewEmptyPlist(void)
+{
+    return NEW_PLIST(T_PLIST_EMPTY, 0);
+}
+
+static inline Obj NewImmutableEmptyPlist(void)
+{
+    return NEW_PLIST_IMM(T_PLIST_EMPTY, 0);
```
fingolfin: How about adding a singleton immutable empty plist, and return that from here, instead of creating lots of new ones? Or rather, since the name `NewImmutableEmptyPlist` then would be misleading, just expose that global here, as `extern Obj ImmutableEmptyPlist;` or so?

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+}
+
+/****************************************************************************
+**
+*F  NewPlistFromArray(<list>,<length>) . . . . .Create a Plist from a C array
```
fingolfin:
```suggestion
*F  NewPlistFromArray(<list>,<length>) . . create a plain list from a C array
```

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
     return val;
 }
 
+
+/****************************************************************************
+**
+*F  NewEmptyPlist() . . . . . . . . . . . . . . . .  Create a new empty Plist
```
fingolfin:
```suggestion
*F  NewEmptyPlist() . . . . . . . . . . . . . . create a new empty plain list
```

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+*F  NewImmutableEmptyPlist() . . . . . . . Create a new immutable empty Plist
```
fingolfin:
```suggestion
*F  NewImmutableEmptyPlist() . . . . . . return an immutable empty plain list
```

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+    return NEW_PLIST_IMM(T_PLIST_EMPTY, 0);
+}
+
+/****************************************************************************
+**
+*F  NewPlistFromArray(<list>,<length>) . . . . .Create a Plist from a C array
+*/
+static inline Obj NewPlistFromArray(Obj * list, Int length)
+{
+    if(length == 0) {
```
fingolfin:
```suggestion
    if (length == 0) {
```
Perhaps run `clang-format` to be sure?

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+static inline Obj NewImmutableEmptyPlist(void)
+{
+    return NEW_PLIST_IMM(T_PLIST_EMPTY, 0);
+}
+
+/****************************************************************************
+**
+*F  NewPlistFromArray(<list>,<length>) . . . . .Create a Plist from a C array
+*/
+static inline Obj NewPlistFromArray(Obj * list, Int length)
```
fingolfin:
```suggestion
static inline Obj NewPlistFromArray(const Obj * list, Int length)
```

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+
+    Obj o = NEW_PLIST(T_PLIST, length);
+    SET_LEN_PLIST(o, length);
+    memcpy(BASE_PTR_PLIST(o), list, length * sizeof(Obj));
+    return o;
+}
+
+/****************************************************************************
+**
+*F  NewPlistFromArgs(<args...>) . . . Create a Plist from a list of arguments
```
fingolfin:
```suggestion
*F  NewPlistFromArgs(<args...>) .  create a plain list from list of arguments
```

src/plist.h
```diff
@@ -291,6 +291,46 @@ static inline Obj PopPlist(Obj list)
...
+*/
+#define NewPlistFromArgs(...)                                                \
+    NewPlistFromArray((Obj[]){ __VA_ARGS__ },                                \
+                      (sizeof((Obj[]){ __VA_ARGS__ }) / sizeof(Obj)))
```
fingolfin: This use C99 variadic macros as well as C99 compound literals, both features not everybody working on the GAP kernel may be familiar with. I think it would be helpful to add a comment above which briefly explains that?

ChrisJefferson: agreed

src/plist.h
ChrisJefferson: Bah, I could have sworn I clang-formatted this (but I definitely didn't as indeed this is wrong).

I haven't clang-formatted the whole PR, as it moves lots of the lines which call these new functions, making them look strange in their local context

src/plist.c
```diff
@@ -58,6 +58,14 @@
 #endif
 
 
+/****************************************************************************
+**
+*V  ImmutableEmptyPlist  . . . . . . . . . . . . . . .  Immutable Empty Plist
```
fingolfin:
```suggestion
*V  ImmutableEmptyPlist . . . . . . . . . . . . an immutable empty plain list
```

src/plist.h
```diff
@@ -291,6 +291,54 @@ static inline Obj PopPlist(Obj list)
     return val;
 }
 
+
+/****************************************************************************
+**
+*F  NewEmptyPlist() . . . . . . . . . . . . . . create a new empty plain list
+*F  ImmutableEmptyPlist . . . . . . . . . . . . an immutable empty plain list
```
fingolfin:
```suggestion
*V  ImmutableEmptyPlist . . . . . . . . . . . . an immutable empty plain list
```

src/plist.h
```diff
@@ -291,6 +291,54 @@ static inline Obj PopPlist(Obj list)
...
+**
+*F  NewPlistFromArgs(<args...>) .  create a plain list from list of arguments
+**
+**  This macro turns a variable-length list of macro arguments into an array,
+**  which is then passed to NewPlistFromArray.
+**
+**  __VA_ARGS__ contains the list of arguments given to this macro.
+**  (Obj[]){ __VA_ARGS__ } creates an array of Obj containing the elements
+**  of __VA_ARGS__.
+**  sizeof(A)/sizeof(Obj) gives us the number of elements in the array A.
```
fingolfin: Minor nitpick, no change needed: This would be self-explanatory if you used  the `ARRAY_SIZE` macro, wouldn't it?

ChrisJefferson: Hadn't noticed we had one :)

PR #2966, opened 2018-11-06
Remove option to return values from break loop in many places

ChrisJefferson:
This removes the option to return a value from the break loop in many places (in particular, places Max just cleaned up).

This isn't all places, but I thought I'd make this simple pass first, then work through the other cases in further PRs.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -85,8 +85,8 @@
 #include "set.h"
 
 
-#define RequireBlistMayReplace(funcname, op, argname) \
-    RequireArgumentConditionMayReplace(funcname, op, argname, IsBlistConv(op), \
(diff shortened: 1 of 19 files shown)


ChrisJefferson:
I wasn't sure if those functions might end up being wanted at some point in packages, or the library. I'm happy to delete them.

fingolfin:
Well, I see it like this: if somebody *really* wants those `*MayReplace` variants, they better have a really strong convincing argument; and then we can add them back. In the meantime we want to remove (and hence also discourage) this style of coding, don't we? Hence I'd remove them. Alas, that certainly can be done in another PR. No need to delay this one for it.

PR #2968, opened 2018-11-06
kernel: add MakeObjInt helper

fingolfin:
This is highly useful for kernel extensions that interface GAP with other systems using bigints (typically GMP `mpz_t`), such as NormalizInterface, SingularInterface, GAPJulia and more.

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -563,7 +563,7 @@ Obj ObjInt_UInt8( UInt8 i )
 #endif
 }
 
-/**************************************************************************
+/****************************************************************************
 **
 ** Convert GAP Integers to various C types -- see header file
 */
@@ -684,7 +684,7 @@ Obj MakeObjInt(const UInt * limbs, int size)
     else if (size == -1)
         obj = ObjInt_UIntInv(limbs[0]);
     else {
-        UInt tnum = (size > 0 : T_INTPOS : T_INTNEG);
+        UInt tnum = (size > 0 ? T_INTPOS : T_INTNEG);
         if (size < 0) size = -size;
         obj = NewBag(tnum, size * sizeof(mp_limb_t));
         memcpy(ADDR_INT(obj), limbs, size * sizeof(mp_limb_t));
--- a/src/integer.c
+++ b/src/integer.c
@@ -15,7 +15,7 @@
 **  Each integer has a unique representation, e.g., an integer that can be
 **  represented as 'T_INT' is never represented as 'T_INTPOS' or 'T_INTNEG'.
 **
-**  In the following, let 'N' be the number of bits in an mp_limb_t (so 32 or
+**  In the following, let 'N' be the number of bits in an UInt (so 32 or
 **  64, depending on the system). 'T_INT' is the type of those integers small
 **  enough to fit into N-3 bits. Therefore the value range of this small
 **  integers is $-2^{N-4}...2^{N-4}-1$. Only these small integers can be used
@@ -46,7 +46,8 @@
 **
 **  These large integers values are represented as low-level GMP integer
 **  objects, that is, in base 2^N. That means that the bag of a large integer
-**  has the following form:
+**  has the following form, where the "digits" are also more commonly referred
+**  to as "limbs".:
 **
 **      +-------+-------+-------+-------+- - - -+-------+-------+-------+
 **      | digit | digit | digit | digit |       | digit | digit | digit |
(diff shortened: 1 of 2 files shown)


src/integer.c
```diff
@@ -343,26 +345,11 @@ static Obj GMPorINTOBJ_FAKEMPZ( fake_mpz_t fake )
...
-    memcpy(ADDR_INT(obj), v->_mp_d, size * sizeof(mp_limb_t));
-
-    // FIXME: necessary to normalize and reduce???
-    obj = GMP_NORMALIZE( obj );
-    obj = GMP_REDUCE( obj );
-  }
-  return obj;
+        return MakeObjInt((const UInt *)v->_mp_d, -size, 1);
+    else
+        return MakeObjInt((const UInt *)v->_mp_d, size, 0);
```
sebasguts: Is there a specific reason to distinguish between positive and negative sizes here and not in `MakeObjInt`?

fingolfin: Encoding the sign inside the size is something specific to GMP, and I wanted this API to be generic. Of course as long as all client code is using GMP, it doesn't matter... so if people prefer it, I can also drop the `isNegative` argument, and encode the sign in the size, no problem at all.

sebasguts: No, I just wanted to know what the reason for this is. I simply was not sure. This is fine to me. Currently this can only be used with GMP's anyway, right?

fingolfin: Well, my hope is that it could be used with other bigint implementations, but most software I interact with is using GMP... though IIRC, Bill Hart has some other bigint code...

markuspf:
So, great, I wanted this for the `PARIInterface` package anyway, PARI has its own bigint implementation (this is the fickliest thing I still have to do to make a beta-release of PARIInterface).

Can we settle questions such as "what is a limb?" somewhere? is this documented in GMP? Can we at least add a pointer to that?

What about the opposite direction? Can I just use a GAP intobj (when it's not an immediate integer) somehow? Should we offer an export to a GMP `mpz_t` as the "de-facto standard`?

fingolfin:
Yes, what a `limb` is is documented in the extensive GMP manual. But it was also already kinda documented in `src/integer.c`; I just tweaked that text to make it explicit, though.

I also changed the API to be more closely to GMP; I figured that it actually is quite easy to use that way, even if your bigint implementation of choice does it differently; but it does simplify at least 4 implementations of GMP-to-GAP-integers that I know of.

Regarding the other direction: No, you usually cannot "just use a GAP large integer" with some 3rd party API, because those usually expect some particular memory layout. You *can* use them with the GMP low-level APIs (i.e. `mpn_*`) if you are careful. You *cannot* use them in place of `mpz_t` directly (well, if one is very careful , one can still use them with some tricks and relatively low overhead as *inputs* for some of the `mpz_*` high level APIs, see the `fake_mpz_t` in added to `src/integer.c`).

So that converting a GAP integer to a `mpz_t`, no, we don't offer a function for that right now. We could but note that I don't want to pollute `src/integer.h` with any GMP specifics; I worked hard on getting all of them out there. So it would have to be abstracted in some way. And then I am not sure how helpful it would be for your PARI interface, if that is not using GMP. That said, it is not very difficult to convert a GAP integer object into an `mpz_t` either, using `ADDR_INT` resp. `INT_INTOBJ`, but I am not sure how helpful that really would be for your purposes, as the main work required for outputting `mpz_t` is dealing with some quirks of GMP (how zero is encoded, how the sign is encoded, etc.).

While you are here, let me point out something regarding terminology: right now in our code base, the term `intobj` *strictly* refers to immediate integers, so there is no such thing as a "GAP intobj which is not immediate". This is codified by the macros `INTOBJ_INT` and `INT_INTOBJ` as well as `IS_INTOBJ`. Now, of course we could decide that we want to refer to all three types of GAP integer objects as "intobj" because its short and convenient, and then take about "small (or immediate) and large intobjs". But I'd only support that if we also renamed all the places that use "intobj" or "INTOBJ" in the current sense (and thus also all packages using them in kernel extensions), because otherwise, things will be confusing as hell... Not sure that's realistic, though, hence I propose we look for some other names. Anyway, I already tried to improve our code comments to use somewhat more consistent terminology; and so I currently write "integer object" for the genric thing, and try to avoid the term "intobj", instead talking about "small integers" resp. "immediate integers", vs. "large integers". (BTW, almost everybody else calls these "big integers" or "bigints"; I thought about changing our terminology to that, but the use of "large integer" is quite pervasive in the GAP code base, so I thought it was not worth it).

fingolfin:
@markuspf if you can tell us some details about the bigint format used by PARI, perhaps we can derive from that a good plan for a GAP API to "export" a GAP integer in a way that is useful for both GMP and PARI.

fingolfin:
On little endian systems, the word sizes don't really matter much (of course you must make sure to calculate the number of limbs / bytes to be transferred correctly). On big endian, it is more complicated, but these days I'd just ignore this (or put in an `#ifdef BIGENDIAN #error Not yet supported` or so).

fingolfin:
BTW, how does one compile `PARIInterface`? There is no build system and the README doesn't say anything about it either.

fingolfin:
Anyway, can we merge this PR now? Is there any change request people are having? @markuspf is the documentation for limbs clear enough for you now?

markuspf:
All good.

PR #2969, opened 2018-11-07
Fix task scanning for the Julia GC

rbehrends:
This is a tracking issue to discuss and track changes for scanning Julia task stacks.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -362,7 +362,6 @@ static jl_module_t *   Module;
 static jl_datatype_t * datatype_mptr;
 static jl_datatype_t * datatype_bag;
 static jl_datatype_t * datatype_largebag;
-static Bag *           StackBottomBags;
 static UInt            StackAlignBags;
 static jl_ptls_t       JuliaTLS, SaveTLS;
 static size_t          max_pool_obj_size;
@@ -474,15 +473,32 @@ static void TryMark(void * p)
     }
 }
 
+static void TryMarkRangeReverse(void * start, void * end)
+{
+    if (lt_ptr(end, start)) {
+        SWAP(void *, start, end);
+    }
+    char * p = (char *)(align_ptr(start));
+    char * q = (char *)end - sizeof(void *);
+    while (!lt_ptr(q, p)) {
+        void * addr = *(void **)q;
+        if (addr)
+            TryMark(addr);
+        q -= StackAlignBags;
+    }
+}
+
 static void TryMarkRange(void * start, void * end)
 {
     if (lt_ptr(end, start)) {
         SWAP(void *, start, end);
     }
-    char * p = align_ptr(start);
+    char * p = (char *)align_ptr(start);
     char * q = (char *)end - sizeof(void *) + StackAlignBags;
     while (lt_ptr(p, q)) {
-        TryMark(*(void **)p);
+        void * addr = *(void **)p;
+        if (addr)
+            TryMark(addr);
         p += StackAlignBags;
     }
 }
@@ -502,6 +518,14 @@ void GapRootScanner(int full)
     // mark our Julia module (this contains references to our custom data
     // types, which thus also will not be collected prematurely)
     JMark(Module);
+    jl_task_t * task = JuliaTLS->current_task;
+    size_t      size;
+    int         tid;
+    // We figure out the end of the stack from the current task. While
+    // `stack_bottom` is passed to InitBags(), we cannot use that if
+    // current_task != root_task.
+    char * stackend = (char *)jl_task_stack_buffer(task, &size, &tid);
+    stackend += size;
 
     // allow installing a custom marking function. This is used for
     // integrating GAP (possibly linked as a shared library) with other code
@@ -514,7 +538,7 @@ void GapRootScanner(int full)
     syJmp_buf registers;
     sySetjmp(registers);
     TryMarkRange(registers, (char *)registers + sizeof(syJmp_buf));
-    TryMarkRange((char *)registers + sizeof(syJmp_buf), StackBottomBags);
+    TryMarkRange((char *)registers + sizeof(syJmp_buf), stackend);
 
     // mark all global objects
     for (Int i = 0; i < GlobalCount; i++) {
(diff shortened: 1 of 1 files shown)


src/julia_gc.c
```diff
@@ -474,6 +474,20 @@ static void TryMark(void * p)
     }
 }
 
+static void TryMarkRangeReverse(void * start, void * end)
+{
+    if (lt_ptr(end, start)) {
+        SWAP(void *, start, end);
+    }
+    char * p = (char *)(align_ptr(start));
```
```suggestion
    char * p = align_ptr(start);
```
to match `TryMarkRange`

src/julia_gc.c
```diff
@@ -474,6 +474,20 @@ static void TryMark(void * p)
...
+    char * q = (char *)end - sizeof(void *);
+    while (!lt_ptr(q, p)) {
+        void *addr = *(void **)q;
+        if (addr) TryMark(addr);
```
fingolfin: If `TryMark` require non-null arguments, then perhaps that check should be done in `TryMark`? Perhaps `TryMark` also should be marked as `inline`?

src/julia_gc.c
```diff
@@ -502,6 +517,11 @@ void GapRootScanner(int full)
     // mark our Julia module (this contains references to our custom data
     // types, which thus also will not be collected prematurely)
     JMark(Module);
+    jl_task_t *task = JuliaTLS->current_task;
+    size_t size;
+    int tid;
+    StackBottomBags = (Bag *)jl_task_stack_buffer(task, &size, &tid);
+    StackBottomBags = (Bag *)((char *)StackBottomBags + size);
```
fingolfin: Since we now always recompute `StackBottomBags` here anyway, is there any reason left to keep it as a global?

Also, I wonder: Why do we need to use `TryMarkRangeReverse` in `GapTaskScanner`, but here we get away with using `TryMarkRange`? Is that some kind of hidden assumption about `current_task` here, e.g. that it equals `root_task`? Some comments explaining this would be great!

src/julia_gc.c
```diff
@@ -531,7 +551,19 @@ void GapTaskScanner(jl_task_t * task, int root_task)
...
     void * stack = jl_task_stack_buffer(task, &size, &tid);
     if (stack && tid < 0) {
-        TryMarkRange(stack, (char *)stack + size);
+        if (task->copy_stack) {
+            // We know the actually used part of the task stack, so
+            // shorten the range we have to scan.
+            stack = (void *)((char *) stack + size - task->copy_stack);
+            size = task->copy_stack;
+        }
+        volatile jl_jmp_buf *old_safe_restore = JuliaTLS->safe_restore;
```
fingolfin: Please insert a brief comment here, perhaps like this:
> // Since we don't know the exact size of the stack, we might run into a read protected page while scanning it, triggering a segfault. Julia catches this, and longjmp's to `JuliaTLS->safe_restore`; we use this mechanism to abort the stack scanning as soon as we hit a protected page. For this to work, we must scan the stack from top to bottom.

src/julia_gc.c
```diff
@@ -531,7 +551,19 @@ void GapTaskScanner(jl_task_t * task, int root_task)
     int    tid;
     void * stack = jl_task_stack_buffer(task, &size, &tid);
```
```suggestion
    char * stack = jl_task_stack_buffer(task, &size, &tid);
```

src/julia_gc.c
```diff
@@ -531,7 +551,19 @@ void GapTaskScanner(jl_task_t * task, int root_task)
...
     if (stack && tid < 0) {
-        TryMarkRange(stack, (char *)stack + size);
+        if (task->copy_stack) {
+            // We know the actually used part of the task stack, so
+            // shorten the range we have to scan.
+            stack = (void *)((char *) stack + size - task->copy_stack);
```
```suggestion
            stack = stack + size - task->copy_stack;
```

src/julia_gc.c
```diff
@@ -531,7 +551,19 @@ void GapTaskScanner(jl_task_t * task, int root_task)
...
+            size = task->copy_stack;
+        }
+        volatile jl_jmp_buf *old_safe_restore = JuliaTLS->safe_restore;
+        jl_jmp_buf exc_buf;
+        if (!jl_setjmp(exc_buf, 0)) {
+            JuliaTLS->safe_restore = &exc_buf;
+            TryMarkRangeReverse(stack, (char *)stack + size);
```
```suggestion
            TryMarkRangeReverse(stack, stack + size);
```

src/julia_gc.c
```diff
@@ -531,7 +551,19 @@ void GapTaskScanner(jl_task_t * task, int root_task)
     int    tid;
     void * stack = jl_task_stack_buffer(task, &size, &tid);
     if (stack && tid < 0) {
-        TryMarkRange(stack, (char *)stack + size);
+        if (task->copy_stack) {
+            // We know the actually used part of the task stack, so
+            // shorten the range we have to scan.
+            stack = (void *)((char *) stack + size - task->copy_stack);
+            size = task->copy_stack;
+        }
```
fingolfin: Please insert a brief comment here, perhaps like this:
> // Since we don't know the exact size of the stack, we might run into a read protected page while scanning it, triggering a segfault. Julia catches this, and longjmp's to `JuliaTLS->safe_restore`; we use this mechanism to abort the stack scanning as soon as we hit a protected page. For this to work, we must scan the stack from top to bottom.

src/julia_gc.c
rbehrends: It doesn't require non-null arguments, this is simply an optimization.

src/julia_gc.c
rbehrends: I wasn't yet sure if it is always good.

We use `TryMarkRange()` here because we *know* both ends of the stack for certain, not just one.

fingolfin: What about my question regarding the hidden assumption about `current_task`?

rbehrends: Oops, let me be clear. `current_task` is the task that is currently being executed and therefore has the stack that we are currently using. This allows us to get the "bottom" of the stack from the task state rather from what `InitBags()` was called with.

src/julia_gc.c
fingolfin: @rbehrends You can accept these proposed changes with a single click into your PR -- try it?

src/julia_gc.c
```diff
@@ -474,15 +473,30 @@ static void TryMark(void * p)
...
+    }
+}
+
 static void TryMarkRange(void * start, void * end)
 {
     if (lt_ptr(end, start)) {
         SWAP(void *, start, end);
     }
-    char * p = align_ptr(start);
+    char * p = (char *)align_ptr(start);
```
fingolfin: Why add this cast?

src/julia_gc.c
fingolfin: Alternatively, either apply the change manually, or comment here that it is undesirable, and mark this conversation as resolved?

src/julia_gc.c
```diff
@@ -529,9 +551,29 @@ void GapTaskScanner(jl_task_t * task, int root_task)
...
+            // shorten the range we have to scan.
+            stack = stack + size - task->copy_stack;
+            size = task->copy_stack;
+        }
+        volatile jl_jmp_buf *old_safe_restore =
+            (volatile jl_jmp_buf *)JuliaTLS->safe_restore;
+        jl_jmp_buf exc_buf;
+        if (!jl_setjmp(exc_buf, 0)) {
+            // The bottom of the stack may be protected with
+            // guard pages, accessing which results in segmentation
```
```suggestion
            // guard pages, accessing these results in segmentation
```

src/julia_gc.c
```diff
@@ -529,9 +551,29 @@ void GapTaskScanner(jl_task_t * task, int root_task)
...
+            // faults. Julia catches those segmentation faults and
+            // longjmps to JuliaTLS->safe_restore; we use this
+            // mechamism to abort stack scanning when a protected
+            // page is hit. For this to work, we must scan the stack
+            // from top to bottom, so we see any guard pages last.
```
fingolfin: Thanks, that's helpful!

src/julia_gc.c
rbehrends: I actually kept this an explicit cast and made the other one an explicit cast, too.

fingolfin:
This is a bit unusual, but I'd like to backport this PR to stable-4.10, for inclusion in GAP 4.10.1. The rationale being that (a) this code is not executed by any regular user, (b) it means that people will be able to use a stock GAP 4.10.1 with our ongoing GAP-Julia integration work, instead of having to wait for GAP 4.11 for that.

Any objections to that plan? @ChrisJefferson @markuspf @alex-konovalov everybody?

alex-konovalov:
Backported in b0ee11069b652241b167c759680eccee6cd1d73e

Issue #2776, opened 2018-09-04
`IsAssociativeElement`: Documentation wrong? Bug?


markuspf:
Documentation
```
31.15-1 IsAssociativeElement
  
  ‣ IsAssociativeElement( obj )  Category
  ‣ IsAssociativeElementCollection( obj )  Category
  ‣ IsAssociativeElementCollColl( obj )  Category
  
  An  element  obj  in the category IsAssociativeElement knows that the multiplication of any elements in the family of obj is associative. For example, all permutations lie in this category, as well as those ordinary matrices (see IsOrdinaryMatrix (24.2-2)) whose entries
  are also in IsAssociativeElement.
```

GAP `master`
```
gap> M := [[Z(4)]];                                                   
[ [ Z(2^2) ] ]
gap> IsOrdinaryMatrix(M);
true
gap> IsAssociativeElementCollection(M);
false
gap> IsAssociativeElementCollection( [ M ] );
false
```

and also

```
gap> M := Matrix(GF(4), [[Z(4)]]);
[ [ Z(2^2) ] ]
gap> IsOrdinaryMatrix(M);
true
gap> IsAssociativeElementCollection(M);
false
```

fingolfin:
`IsAssociativeElementCollection(M) = false;` seems correct to me. Moreover, note that `IsAssociativeElement(M)=true` holds.

I agree that `IsAssociativeElementCollection( [ M ] ) = false` is surprising. However, I do not strictly see an error in the documentation you quote. Could you specify which sentence in it you consider wrong?

But let's look at the technical reasons for this, and what we might be able to do about it. You give two examples, but in `master`, both `M` are really the same, even that returned by `Matrix` (and there are reasons for that, but I'll get back to that below).

The problem is how we compute types of lists (and hence whether they are in a given filter or not). Clearly it is desirable to have this fast. For proper objects, this is no problem, as the type is stored. But for lists, the type has to be recomputed each time we ask for it.

So to make it fast, for homogenous lists, we essentially only take the families of the elements in to account. Consider this snippet:

```
gap> v:=[Z(4)];;  # a row vector
gap> M:=[v];; # = [[Z(4)]],  a "rectangular table" of ring elements
gap> L:=[M];;
gap>
gap> FamilyObj(v) = CollectionsFamily(FamilyObj(Z(4)));
true
gap> FamilyObj(M) = CollectionsFamily(FamilyObj(v));
true
[...]

Issue #2148, opened 2018-02-02
Add new "true" VectorObj / MatrixObj implementation which do not lie in `IsList`


fingolfin:
The existing "vectors-as-lists" and "matrices-as-lists-of-lists" paradigm in GAP causes many subtle and not so subtle problems. Recently, one of them struck again, namely that of vectors of length 0, which are represented by empty lists; but the empty (plain) list does not have the "right" family for a vector space, so one gets confusing behaviour like this:
```
gap> V:=GF(3)^0;;
gap> Zero(V) in V;
false
gap> Zero(V) in Elements(V);
true
gap> Zero(V);
[  ]
```
And also this:
```
gap> W:=GF(5)^0;;
gap> Zero(W) = Zero(V);
true
```

Using `VectorObj` types seems like the natural solution. At first glance, the existing compressed vectors seem to solve the problem (at least for the small fields were they are available):
```
gap> v:=[];; CONV_VEC8BIT(v,3); v;
< mutable compressed vector length 0 over GF(3) >
gap> v in V;
true
```
So one might think about just using a compressed empty vector -- but (a) there are technical problems with that (zero length compressed vectors are poorly supported, see #2121), and (b) that doesn't actually solve that much, because compressed vector for historical reasons still lie in the filter `IsList`, and that means they have to adopt all kinds of undesirable behavior. Like this:
```
gap> v:=[];; CONV_VEC8BIT(v,3); v;
< mutable compressed vector length 0 over GF(3) >
gap> w:=[];; CONV_VEC8BIT(w,5); w;
< mutable compressed vector length 0 over GF(5) >
gap> v = w; v = []; w = [];
false
true
true
```
So transitivity of `=` is violated, and the reason is that `v=w` is buggy: since both are supposed to be lists, and lists are equal if they have equal elements, and both lists are empty, well: `v=w` should return `true`.

To overcome this, we need new VectorObj (and MatrixObj) implementations which are deliberately *not* in the filter `IsList`. We'd also need these for larger finite fields, and for characteristic zero. Then, we'd have to slowly switch the library to use these everywhere (or at least in as many places as possible), while simultaneously getting package authors to adapt their packages, too...

As to how to get there, I'll quote @stevelinton from PR #2125:
[...]

PR #2973, opened 2018-11-07
Overhaul IsPlistMatrixRep and IsPlistVectorRep

fingolfin:
Progress towards issue #2148 : This PR is an incomplete proof of concept (see #2776 for its genesis), I am putting it out here to make sure people are aware of this possible direction, and in the vague hope that somebody else might help out with it.

The idea is to tune `IsPlistMatrixRep` as much as possible so that it becomes a viable replacement for lists-of-lists (= lol) "matrices". Indeed, with this hackish POC, I already benchmarked some cases where the `IsPlistMatrixRep` code is *faster* than lists-of-lists, probably in parts because of avoided method selection overhead; another part is that the optimized code benefits from kernel optimizations for multiplication of lol matrices

To give an idea about the performance of `IsPlistMatrixRep` before and after this PR, I run some mini benchmarks, using the [benchmarking code in this gist](https://gist.github.com/fingolfin/b81101cc304d6062d195fa80fae32ab4), plus the following code:

```
M1:=IdentityMatrix(GF(4),10);
M2:=IdentityMat(10)*Z(2); # "decompress" the matrix
M3:=IdentityMatrix(IsPlistMatrixRep, GF(4),10);
I1:=IdentityMat(10);
I2:=IdentityMatrix(IsPlistMatrixRep,Integers,10);

inputs := [
   [ M1, GF(4), Is8BitMatrixRep ],
   [ M2, GF(4), IsPlistRep ],
   [ M3, GF(4), IsPlistMatrixRep ],
   [ I1, Integers, IsPlistRep ],
   [ I2, Integers, IsPlistMatrixRep ],
];

f := function(mat) local i, x; for i in [1..10000] do x:=mat*mat; od; end;;
g := function(mat) local i, x; for i in [1..10000] do x:=mat^5+mat; od; end;;
opts := rec(mintime:=1000);

for test in [ f, g ] do
    Print("-----\n");
    Print("Testing function ", NameFunction(test), "\n");
    Print("-----\n");

    for input in inputs do
        Print("... on ", NameFunction(input[3]), " matrix over ", input[2], "...\n");
        Assert(0, input[3](input[1])); # verify filter matches
        Benchmark(function() test(input[1]); end, opts);;
        Print("\n");
    od;

    Print("\n");
od;
```

The results of this are as follows with `master` (using commit 5b7263cc69a93f0dc9a067d5df90c4e756d46bbb from 2019-07-11), on a Ubuntu server in Gießen:
```
-----
Testing function f
-----
... on Is8BitMatrixRep matrix over GF(2^2)...
Performed 128 iterations, taking 1008. milliseconds.
average: 7.87 +/- 0.02 (+/- 0.%)
median: 7.870849609375

... on IsPlistRep matrix over GF(2^2)...
Performed 5 iterations, taking 1517. milliseconds.
average: 303.45 +/- 0.12 (+/- 0.%)
median: 303.48095703125

... on IsPlistMatrixRep matrix over GF(2^2)...
Performed 5 iterations, taking 5801. milliseconds.
average: 1160.23 +/- 0.89 (+/- 0.%)
median: 1159.994140625

... on IsPlistRep matrix over Integers...
Performed 18 iterations, taking 1001. milliseconds.
average: 55.62 +/- 0.07000000000000001 (+/- 0.%)
median: 55.6290283203125

... on IsPlistMatrixRep matrix over Integers...
Performed 5 iterations, taking 1900. milliseconds.
average: 380.09 +/- 0.33 (+/- 0.%)
median: 380.011962890625


-----
Testing function g
-----
... on Is8BitMatrixRep matrix over GF(2^2)...
Performed 29 iterations, taking 1010. milliseconds.
average: 34.84 +/- 0.02 (+/- 0.%)
median: 34.8369140625

... on IsPlistRep matrix over GF(2^2)...
Performed 5 iterations, taking 4797. milliseconds.
average: 959.4400000000001 +/- 0.21 (+/- 0.%)
median: 959.5048828125

... on IsPlistMatrixRep matrix over GF(2^2)...
Performed 5 iterations, taking 15796. milliseconds.
average: 3159.28 +/- 1.58 (+/- 0.%)
median: 3159.17724609375

... on IsPlistRep matrix over Integers...
Performed 5 iterations, taking 1014. milliseconds.
average: 202.89 +/- 0.09 (+/- 0.%)
median: 202.889892578125

... on IsPlistMatrixRep matrix over Integers...
Performed 5 iterations, taking 6131. milliseconds.
average: 1226.19 +/- 0.49 (+/- 0.%)
median: 1226.10791015625
```

With this PR (commit cb79cdf58b85ae83b3fbf218332e36a16970cdfb), the numbers for `Is8BitMatrixRep` and `IsPlistRep` stay essentially the same, but for `IsPlistMatrixRep` improve to more or less match those for `IsPlistRep` (they are a bit slower, but not that much):
```
-----
Testing function f
-----
... on Is8BitMatrixRep matrix over GF(2^2)...
Performed 127 iterations, taking 1001. milliseconds.
average: 7.88 +/- 0.06 (+/- 1.%)
median: 7.870849609375

... on IsPlistRep matrix over GF(2^2)...
Performed 5 iterations, taking 1513. milliseconds.
average: 302.69 +/- 0.09 (+/- 0.%)
median: 302.649169921875

... on IsPlistMatrixRep matrix over GF(2^2)...
Performed 5 iterations, taking 1582. milliseconds.
average: 316.49 +/- 0.06 (+/- 0.%)
median: 316.454833984375

... on IsPlistRep matrix over Integers...
Performed 18 iterations, taking 1004. milliseconds.
average: 55.77 +/- 0.06 (+/- 0.%)
median: 55.7733154296875

... on IsPlistMatrixRep matrix over Integers...
Performed 15 iterations, taking 1008. milliseconds.
average: 67.19 +/- 0.16 (+/- 0.%)
median: 67.14599609375


-----
Testing function g
-----
... on Is8BitMatrixRep matrix over GF(2^2)...
Performed 29 iterations, taking 1003. milliseconds.
average: 34.6 +/- 0.03 (+/- 0.%)
median: 34.587158203125

... on IsPlistRep matrix over GF(2^2)...
Performed 5 iterations, taking 4775. milliseconds.
average: 955. +/- 0.5600000000000001 (+/- 0.%)
median: 955.198974609375

... on IsPlistMatrixRep matrix over GF(2^2)...
Performed 5 iterations, taking 4858. milliseconds.
average: 971.5700000000001 +/- 0.7000000000000001 (+/- 0.%)
median: 971.367919921875

... on IsPlistRep matrix over Integers...
Performed 5 iterations, taking 1015. milliseconds.
average: 202.91 +/- 0.36 (+/- 0.%)
median: 202.7880859375

... on IsPlistMatrixRep matrix over Integers...
Performed 5 iterations, taking 1140. milliseconds.
average: 228.01 +/- 0.09 (+/- 0.%)
median: 227.99169921875
```

Of course proper benchmarks would include a range of matrix sizes and ground fields; more interesting / realistic operations; and graph all of that nicely...

### DIFF
--- a/lib/matobjplist.gd
+++ b/lib/matobjplist.gd
@@ -59,34 +59,48 @@ DeclareRepresentation( "IsPlistVectorRep",
 ##
 ##  <Description>
 ##  An object <A>obj</A> in <Ref Filt="IsPlistMatrixRep"/> describes
-##  a matrix object (see <Ref Filt="IsMatrixObj"/>) that behaves similar to
-##  a list of its rows, in the sense defined in
-##  Section <Ref Sect="Operations for Row List Matrix Objects"/>.
-##  It is internally represented as a positional object
-##  (see <Ref Filt="IsPositionalObjectRep"/> that stores four entries:
-##  <Enum>
-##  <Item>
-##    its base domain
-##    (see <Ref Attr="BaseDomain" Label="for a matrix object"/>),
-##  </Item>
-##  <Item>
-##    the number of rows
-##    (see <Ref Attr="NumberRows" Label="for a matrix object"/>), and
-##  </Item>
-##  <Item>
-##    the number of columns
-##    (see <Ref Attr="NumberColumns" Label="for a matrix object"/>), and
-##  </Item>
-##  <Item>
-##    a plain list (see <Ref Filt="IsPlistRep"/> of its rows,
-##    each of them being an object in <Ref Filt="IsPlistVectorRep"/>.
-##  </Item>
-##  </Enum>
+##  a matrix object (see <Ref Filt="IsMatrixObj"/>) that internal stores its
+##  entries as a classic GAP matrix, that is as a plain list (see
+##  <Ref Filt="IsPlistRep"/>) of plain lists. Therefore any such GAP matrix
+##  can be represented as a <Ref Filt="IsPlistMatrixRep"/>, making this
+##  representation very versatile, and a good place to start if one wants to
+##  adapt code which previously produced such classic GAP matrices to instead
+##  produce matrix objects.
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
+##  Implementation note: a matrix in IsPlistMatrixRep is internally
+##  represented as a positional object that stores four entries:
+##  - its base domain
+#   - the number of rows
+##  - the number of columns
+##  - a plain list of its rows, each also a plain list
 DeclareRepresentation( "IsPlistMatrixRep",
+        IsListMatrix and IsPositionalObjectRep
+    and IsNoImmediateMethodsObject
+    and HasNumberRows and HasNumberColumns
+    and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
+    [] );
+
+
+#############################################################################
+##
+##  <#GAPDoc Label="IsRowPlistMatrixRep">
+##  <ManSection>
+##  <Filt Name="IsRowPlistMatrixRep" Arg='obj' Type="representation"/>
+##
+##  <Description>
+##  An object <A>obj</A> in <Ref Filt="IsRowPlistMatrixRep"/> describes
+##  a matrix object (see <Ref Filt="IsMatrixObj"/>) that behaves similar to
+##  a list of its rows, in the sense defined in
+##  Section <Ref Sect="Operations for Row List Matrix Objects"/>.
+##  Its rows can be accessed as objects in <Ref Filt="IsPlistVectorRep"/>,
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareRepresentation( "IsRowPlistMatrixRep",
         IsRowListMatrix and IsPositionalObjectRep
     and IsNoImmediateMethodsObject
     and HasNumberRows and HasNumberColumns
--- a/lib/matobjplist.gd
+++ b/lib/matobjplist.gd
@@ -29,7 +29,7 @@
 ##  in a row list matrix
 ##  (see Section <Ref Subsect="Operations for Row List Matrix Objects"/>).
 ##  It is internally represented as a positional object
-##  (see <Ref Filt="IsPositionalObjectRep"/> that stores 2 entries:
+##  (see <Ref Filt="IsPositionalObjectRep"/> that stores two entries:
 ##  <Enum>
 ##  <Item>
 ##    its base domain
@@ -63,14 +63,15 @@ DeclareRepresentation( "IsPlistVectorRep",
 ##  a list of its rows, in the sense defined in
 ##  Section <Ref Sect="Operations for Row List Matrix Objects"/>.
 ##  It is internally represented as a positional object
-##  (see <Ref Filt="IsPositionalObjectRep"/> that stores 4 entries:
+##  (see <Ref Filt="IsPositionalObjectRep"/> that stores four entries:
 ##  <Enum>
 ##  <Item>
 ##    its base domain
 ##    (see <Ref Attr="BaseDomain" Label="for a matrix object"/>),
 ##  </Item>
 ##  <Item>
-##    an empty vector in the representation of each row, 
+##    the number of rows
+##    (see <Ref Attr="NumberRows" Label="for a matrix object"/>), and
 ##  </Item>
 ##  <Item>
 ##    the number of columns
@@ -94,24 +95,16 @@ DeclareRepresentation( "IsPlistMatrixRep",
 
 
 # Some constants for matrix access:
-BindGlobal( "BDPOS", 1 );
-BindGlobal( "EMPOS", 2 );
-BindGlobal( "RLPOS", 3 );
-BindGlobal( "ROWSPOS", 4 );
+# TODO rename these so that one can quickly see that they belong to IsPlist*Rep
+BindConstant( "BDPOS", 1 );
+BindConstant( "NUM_ROWS_POS", 2 );
+BindConstant( "NUM_COLS_POS", 3 );
+BindConstant( "ROWSPOS", 4 );
 
 # For vector access:
-#BindGlobal( "BDPOS", 1 );   # see above
-BindGlobal( "ELSPOS", 2 );
+#BindConstant( "BDPOS", 1 );   # see above
+BindConstant( "ELSPOS", 2 );
 
 # Two filters to speed up some methods:
 DeclareFilter( "IsIntVector" );
 DeclareFilter( "IsFFEVector" );
-
-############################################################################
-# Constructors:
-############################################################################
-
-#T Should this be documented?
-#T It seems to be just an auxiliary function for the documented constructors.
-DeclareGlobalFunction( "MakePlistVectorType" );
-
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -750,14 +750,9 @@ InstallMethod( Display, "for a plist matrix", [ IsPlistMatrixRep ],
     local i;
     Print("<");
     if not IsMutable(m) then Print("immutable "); fi;
-    Print(NrRows(m),"x",NrCols(m),"-matrix over ",m![BDPOS],":\n");
+    Print(NrRows(m),"x",NrCols(m),"-matrix over ",m![BDPOS],": [\n");
     for i in [1..NrRows(m)] do
-        if i = 1 then
-            Print("[");
-        else
-            Print(" ");
-        fi;
-        Print(m![ROWSPOS][i],"\n");
+        Print(" ",m![ROWSPOS][i],"\n");
     od;
     Print("]>\n");
   end );
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -18,10 +18,9 @@
 # Constructors:
 ############################################################################
 
-InstallMethod( NewVector, "for IsPlistVectorRep, a semiring, and a list",
-  [ IsPlistVectorRep, IsSemiring, IsList ],
-  function( filter, basedomain, l )
-    local typ;
+BindGlobal( "MakeIsPlistVectorRep",
+  function( basedomain, list )
+    local filter, typ;
     filter := IsPlistVectorRep and IsMutable;
     if HasCanEasilyCompareElements(Representative(basedomain)) and
        CanEasilyCompareElements(Representative(basedomain)) then
@@ -33,14 +32,23 @@ InstallMethod( NewVector, "for IsPlistVectorRep, a semiring, and a list",
         filter := filter and IsFFEVector;
     fi;
     typ := NewType(FamilyObj(basedomain), filter);
-    return Objectify(typ, [ basedomain, ShallowCopy(l) ]);
+    return Objectify(typ, [ basedomain, list ]);
+  end );
+
+
+InstallMethod( NewVector, "for IsPlistVectorRep, a semiring, and a list",
+  [ IsPlistVectorRep, IsSemiring, IsList ],
+  function( filter, basedomain, list )
+    list := ShallowCopy( list );
+    return MakeIsPlistVectorRep( basedomain, list );
   end );
 
 InstallMethod( NewZeroVector, "for IsPlistVectorRep, a semiring, and an int",
   [ IsPlistVectorRep, IsSemiring, IsInt ],
-  function( filter, basedomain, l )
-    l := ListWithIdenticalEntries(l,Zero(basedomain));
-    return NewVector( filter, basedomain, l );
+  function( filter, basedomain, len )
+    local list;
+    list := ListWithIdenticalEntries(len, Zero(basedomain));
+    return MakeIsPlistVectorRep( basedomain, list );
   end );
 
 InstallMethod( NewMatrix,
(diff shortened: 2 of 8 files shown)


lib/matobjplist.gi
```diff
@@ -639,120 +641,120 @@ InstallMethod( Matrix, "for a list and a plist matrix",
...
+#   [ IsPlistMatrixRep ],
+#   function( m )
+#     return List(m![ROWSPOS]);
+#   end );
+# 
+# InstallMethod( ListOp, "for a plist matrix and a function",
+#   [ IsPlistMatrixRep, IsFunction ],
+#   function( m, f )
+#     return List(m![ROWSPOS],f);
+#   end );
```
fingolfin: MatrixObj are not lists, so remove these. If one wants to extract the rows or columns as VectorObj, then we should add `GetColumnVectors` / `GetRowVectors` operations (better names pending...)

lib/matobjplist.gi
```diff
@@ -639,120 +641,120 @@ InstallMethod( Matrix, "for a list and a plist matrix",
...
+# 
+# InstallMethod( Unpack, "for a plist matrix",
+#   [ IsPlistMatrixRep ],
+#   function( m )
+#     return List(m![ROWSPOS],v->ShallowCopy(v![ELSPOS]));
+#   end );
```
fingolfin: This is part of the MatrixObj API as it is declared in `lib/matobj2.gd`. We certainly can implement it -- but I wonder what the point behind this operation is. The only use for this I can find is in `ChangedBaseDomain` methods.

lib/matobjplist.gi
```diff
@@ -929,19 +928,8 @@ InstallMethod( \*, "for two plist matrices",
...
-            v := a![ROWSPOS][i];
-            w := ZeroVector(b![RLPOS],b![EMPOS]);
-            for j in [1..a![RLPOS]] do
-                AddRowVector(w,b![ROWSPOS][j],v[j]);
-            od;
-            l[i] := w;
-        fi;
-    od;
+    l := a![ROWSPOS]*b![ROWSPOS];
+    #l := PROD_LIST_SCL_DEFAULT(a![ROWSPOS],b![ROWSPOS]);
```
fingolfin: Here we use the optimized matrix multiplication functions from the kernel, which is part of the reason for the performance improvements.

lib/matobjplist.gi
```diff
@@ -1176,68 +1156,6 @@ InstallMethod( Randomize, "for a mutable plist matrix",
...
-        l := m![ROWSPOS][i]![ELSPOS];
-        for j in [1..i-1] do
-            if not IsZero(l[j]) then
-                return false;
-            fi;
-        od;
-    od;
-    return true;
-  end );
-
```
fingolfin: I did not benchmark the removed methods against the default ones -- might be interesting to do, perhaps there is a reason to keep these specialized ones after all, although on the surface, they seem to do almost exactly the same work (perhaps even are slower, due to the use of `IsZero`).

lib/matrix.gi
```diff
@@ -168,7 +168,6 @@ InstallOtherMethod( IsDiagonalMat,
     [ IsMatrix ],
     function( mat )
    local  i, j,z;
-    if IsEmpty(mat) then return true;fi;
     z:=Zero(mat[1,1]);
```
fingolfin: This, of course, assumes that the all matrix entries have the "same" zero. In some scenarios this may not be the case, though -- at least @frankluebeck mentioned the idea of having, say, a matrix of  polynomials over the rationals, but using integers or rationals as entries instead of the corresponding constant polynomials, to save space and have more efficient arithmetic operations.

In that case, the above optimization won't work correctly.

fingolfin:
I think this is getting close to being mergeable. I am sure there are bugs in my changes, but to flush those out, we should write more and better systematic tests for the interfaces.

lib/matobjplist.gi
```diff
@@ -18,108 +18,90 @@
...
-      if nd < 2 or nd mod 2 = 1 then
-        if Length(l) mod rl <> 0 then
-          Error( "NewMatrix: Length of l is not a multiple of rl" );
+    # of lists of length 'ncols'.
+    if Length(l) = 0 then
+      # empty matrix
+      m := [];
+    elif IsVectorObj(l[1]) then
+      # list of vectors
+      # TODO: convert each IsVectorObj to a plist
```
ssiccha, line 58: That sounds to me like you want to turn each entry of `l` into `IsPlistRep`. That can't be right, or am I misunderstanding something?

lib/matobjplist.gi
```diff
@@ -1202,36 +869,17 @@ InstallMethod( \*, "for a plist vector and a plist matrix",
...
-    res := ZeroVector(m![RLPOS],m![EMPOS]);
-    for i in [1..Length(v![ELSPOS])] do
-        s := v![ELSPOS][i];
-        if not IsZero(s) then
-            AddRowVector(res,m![ROWSPOS][i],v![ELSPOS][i]);
-        fi;
-    od;
+    # TODO: should we verify that Length(v) = NrRows(m)?
+    res := ZeroVector(NrCols(m),v);
+    res![ELSPOS] := v![ELSPOS] * m![ROWSPOS];
```
ThomasBreuer, line 874: I would say it depends on the definition of arithmetic operations for vector and matrix objects:
If we say that the result is undefined if the dimensions do not match then no check is necessary.
If we promise an error message if the dimensions do not match then we have to check.
(Without check, the result object is at least "internally consistent", that is, the claimed number of columns equals the length of the stored list.)

lib/matobjplist.gi
```diff
@@ -204,42 +189,20 @@ InstallMethod( Length, "for a plist vector", [ IsPlistVectorRep ],
...
   [ IsList, IsPlistVectorRep ],
   function( l, t )
     local v;
-    v := ShallowCopy(l);
-    if IsGF2VectorRep(l) then
-        PLAIN_GF2VEC(v);
-    elif Is8BitVectorRep(l) then
-        PLAIN_VEC8BIT(v);
-    fi;
+    v := PlainListCopy(l);
```
ssiccha, line 205: So I take it `PlainListCopy` respects the GF2 and 8Bit vector reps? (This isn't tested accordingn to codecov)

lib/matobjplist.gi
```diff
@@ -595,138 +529,26 @@ InstallMethod( ZeroMatrix, "for two integers and a plist matrix",
...
-    m![ROWSPOS]{pp} := n![ROWSPOS];
-  end );
-
-InstallMethod( Append, "for two plist matrices",
-  [ IsPlistMatrixRep and IsMutable, IsPlistMatrixRep ],
-  function( m, n )
-    Append(m![ROWSPOS],n![ROWSPOS]);
-  end );
-
+# FIXME: does ShallowCopy even make sense for a matrixobj???
```
ThomasBreuer, line 546: In `matobj2.gd`, we define `ShallowCopy` for objects in `IsRowListMatrix`.

ssiccha (commented):
I've had a look up until line 711 and left a few comments.

lib/matobjplist.gi
```diff
@@ -1180,19 +844,22 @@ InstallMethodWithRandomSource( Randomize,
...
-        n[i] := v;
-    od;
-    return Objectify(TypeObj(m),[m![BDPOS],m![EMPOS],Length(m![ROWSPOS]),n]);
+    local trans;
+    # FIXME/TODO: implement this generic, or optimized for IsPlistMatrixRep?
+    # Right now, we do the latter:
+    trans := TransposedMatMutable(m![ROWSPOS]);
+
+ 
+    # FIXME: using TypeObj(m) below is probably wrong, as it
```
ThomasBreuer, line 853: I think there are many places in this file where a `TypeObj` value of the argument is used in `Objectify`, and where this type may be wrong.

lib/matobjplist.gi
```diff
@@ -1028,132 +783,41 @@ InstallMethod( ZeroMutable, "for a plist matrix",
...
     if n = fail then return fail; fi;
-    n := Matrix(n,Length(n),m);
-    if not IsMutable(m) then
-        MakeImmutable(n);
-    fi;
-    return n;
+    # FIXME: if the base domain is not a field, e.g. Integers,
+    # then the inverse we just computed may not be defined over
+    # that base domain. Should we detect this here? Or how else
+    # will we deal with this in general?
```
ThomasBreuer, line 819: I think this is first of all a question about what the operation `NewMatrix` promises. Perhaps we need `NewMatrix` and `NewMatrixNC`? (And then the question is whether we call `NewMatrix` or `NewMatrixNC` here.)

ThomasBreuer (commented):
I have added some specific comments in the code.
Two more general remarks:
Several `v![BDPOS]` have been changed to `BaseDomain(v)`. If this is a general suggestion then should it be applied everywhere?
The documentation of `IsPlistMatrixRep` states that a plain list of `IsPlistVectorRep` objects is stored, should this be changed?

fingolfin:
@ssiccha @ThomasBreuer thank you for your feedback

@ThomasBreuer you wrote:
> The documentation of IsPlistMatrixRep states that a plain list of IsPlistVectorRep objects is stored, should this be changed?

Yes, thank you. In fact, there is more in it that needs to be changed: it implies that `IsPlistMatrixRep` is a row list matrix, which was the case before, but now isn't / shouldn't be... I'll try to explain my train of thought here (and this definitely should be explained in the commit message(s), too): for optimal performance (and I run a lot of benchmarks back then), we really don't want to store a `IsPlistMatrixRep` as a list of `IsPlistVectorRep`; this just adds a ton of overhead. It's much better to simply wrap an old-style `IsMatrix and IsPlistRep` matrix, which this PR does. Also, for testing purposes, it is good to have an `IsMatrixObj` implementation which is *not* an `IsRowListMatrix` (@danielrademacher asked for this, too). So that's a useful application for `IsPlistMatrixRep`.

HOWEVER another useful application for `IsPlistMatrixRep` would be to use it as drop-in replacement for any old-style `IsMatrix and IsPlistRep`; i.e., one may want to change code producing these to instead produce `IsPlistMatrixRep` (not all at once, but one at a time, if/when it seems useful and safe). But for this application, one wants those matrices to work as closely as possible to the old-style matrices; so in particular, here we want an `IsRowListMatrix`.

So here we have a clash of goal. So my thought here now is to introduce `IsRowPlistMatrixRep` (or something like that) which specializes `IsPlistMatrixRep` and is also in `IsRowListMatrix`. I think all that is needed to make that happen is to install suitable `[]` and `[]:=` methods for `IsRowPlistMatrixRep` (and perhaps a few more, dunno). Then `mat[i]` could simply return an `IsVectorObj` which wraps the list `mat![ROWSPOS][i]` (and *not* a shallow copy; it really must point to exactly that list). This way, `mat[i][j] := 1` will still work. Of course this incurs a performance penalty; I think that at least the `[]` method should contain a `Info(InfoPerformance, ...)` message. If we want to optimize it a bit, we could add a `WeakPointerObj` as cache for these row wrapper objects (some care needs to be done: replacing a row via `[]:=` should invalidate the respective cache entry).

fingolfin:
Err, I've started to add `IsRowPlistMatrixRep`, but completing it will be tedious, as I need to once again implement `NewZeroMatrix`, `NewIdentityMatrix`, ... -- we ought to resolve #4366

lib/matobjplist.gd
```diff
@@ -45,7 +45,7 @@
 ##  <#/GAPDoc>
 ##
 DeclareRepresentation( "IsPlistVectorRep",
-   IsVectorObj and IsPositionalObjectRep, [] );
+   IsVectorObj and IsPositionalObjectRep and IsNoImmediateMethodsObject, [] );
```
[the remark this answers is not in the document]
fingolfin: Sure, why not. Open an issue for that idea?

tst/testinstall/MatrixObj/NewMatrix.tst
```diff
@@ -0,0 +1,30 @@
+# IsPlistMatrixRep
+gap> NewMatrix(IsPlistMatrixRep, Integers, 3, [ ] ); Display(last);
+<0x3-matrix over Integers>
+<0x3-matrix over Integers:
+]>
```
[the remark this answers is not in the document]
fingolfin: Indeed, fixed now by revised printing.

lib/matobjplist.gi
fingolfin: A primary point of this PR is to change that, i.e.: do not store a list of vector objects, but rather store an old style matrix. This avoids a ton of overhead in both memory usage and performance. Before this PR, `IsPlistMatrixRep` was hopelessly slow compared to old-style matrices.

lib/matobjplist.gi
```diff
@@ -18,108 +18,90 @@
...
-            m[i] := NewVector( filterVectors, basedomain, l[i] );
-        fi;
-    od;
-    e := NewVector(filterVectors, basedomain, []);
-    m := [basedomain,e,rl,m];
-    filter2 := IsPlistMatrixRep and IsMutable;
+    # FIXME/TODO: should the following test be always performed
+    # or only at a higher assertion level?
+    Assert(0, ForAll(m, row -> Length(row) = ncols));
+    Assert(0, ForAll(m, row -> ForAll(row, x -> x in basedomain)));
```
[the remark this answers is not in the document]
fingolfin: Yes, this is ultimately a general question. As are many of the other TODOs. Note that similar questions arise e.g. for `\+` for matrices (should we check and enforce that the dimensions agree? or do we want to allow one matrix to be "smaller" than the other? Or do we not want to check at all and rely on the user? 

For other binary operations, should we check that the base domains are identical? Or possibly equal (slower, but more permissive)? Or perhaps we should allow adding matrices with different base domains as long as there is a common parent of the base domains (so adding a GF(4) and a GF(8) matrix might produce a GF(64) matrix). If so, what mechanism should be used to determine that "parent" domain? 

And if we do perform any such checks, do we need an NC version of "+"? And how would one name that?

There's more. I am also pretty sure we already listed some of these somewhere.

lib/matobjplist.gi
fingolfin: What do you mean with "respects" them? It returns a new plain list with the content of the given list, and it works for arbitrary lists, including GF2 and 8bit vector reps.

lib/matobjplist.gi
fingolfin: Spot on. Thing is, in my mind, `IsPlistMatrixRep` should not have been in `IsRowListMatrix`, but of course it still was when you wrote that review remark. I simply forgot to make that change... now done, and `IsRowPlistMatrixRep` added. And then I changed this `ShallowCopy` method to require `IsRowPlistMatrixRep`.

lib/matobjplist.gd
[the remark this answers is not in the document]
ThomasBreuer: See #4529.

fingolfin:
Superseded by PR #6322

PR #2975, opened 2018-11-07
Add GetSmalIint

ChrisJefferson:
A very common pattern in GAP (which became more obvious when @fingolfin recently added RequireSmallInt) is to check if something is a small int, and then call INT_INTOBJ to get the small integer out.

This PR introduces GetSmallInt, which combines this pattern. This (in my mind) makes code smaller and easier to read, and also helps get rid of occurrences of the unsafe INT_INTOBJ function (although all removed uses here were fine).


src/error.h
```diff
@@ -187,6 +189,17 @@ extern Obj RACErrorHelper(int          mayReturn,
...
         "must be a small integer")
 
+/****************************************************************************
+**
+*F  GetSmallInt
+*/
+static inline Int GetSmallInt(const char * funcname, Obj op, const char * argname)
+{
+    RequireArgumentCondition(funcname, op, argname, IS_INTOBJ(op),
+                             "must be a small integer");
```
fingolfin: Any reason not to use `RequireSmallInt` here?

src/error.h
```diff
@@ -13,6 +13,8 @@
 #define GAP_ERROR_H
 
 #include "system.h"
+#include "integer.h"
```
fingolfin: Why is integer.h included?

src/error.h
ChrisJefferson: Nope, changed

src/error.h
ChrisJefferson: for some reason I thought I needed it for INT_INTOBJ, which obviously isn't true. removed.

PR #2976, opened 2018-11-08
Remove returning values from the break loop in blister.c

ChrisJefferson:
This removes ErrorReturnObj from blister.c

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -459,22 +459,18 @@ Obj ElmsBlist (
 
             /* get <position>                                              */
             Obj p = ELMW_LIST(poss, i);
-            while (!IS_INTOBJ(p)) {
-                p = ErrorReturnObj("List Elements: position is too large for "
-                                   "this type of list",
-                                   0L, 0L,
-                                   "you can supply a new position <pos> via "
-                                   "'return <pos>;'");
+            if (!IS_INTOBJ(p)) {
+                ErrorMayQuit("List Elements: position is too large for "
+                             "this type of list",
+                             0L, 0L);
             }
             pos = INT_INTOBJ(p);
 
             /* select the element                                          */
             if ( lenList < pos ) {
-                ErrorReturnVoid(
+                ErrorMayQuit(
                     "List Elements: <list>[%d] must have an assigned value",
-                    pos, 0L,
-                    "you can assign a value and 'return;'" );
-                return ELMS_LIST( list, poss );
+                    pos, 0L);
             }
 
             /* assign the element into <elms>                              */
@@ -1117,12 +1113,7 @@ Obj FuncLIST_BLIST (
     RequireSmallList("ListBlist", list);
     /* get and check the second argument                                   */
     RequireBlist("ListBlist", blist, "blist");
-    while ( LEN_LIST( list ) != LEN_BLIST( blist ) ) {
-        blist = ErrorReturnObj(
-            "ListBlist: <blist> must have the same length as <list> (%d)",
-            LEN_LIST( list ), 0L,
-            "you can replace <blist> via 'return <blist>;'" );
-    }
+    CheckSameLength("ListBlist", "blist", "list", blist, list);
 
     /* compute the number of 'true'-s                                      */
     n = SizeBlist(blist);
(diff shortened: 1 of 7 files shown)


src/blister.c
```diff
@@ -1117,11 +1113,10 @@ Obj FuncLIST_BLIST (
     RequireSmallList("ListBlist", list);
     /* get and check the second argument                                   */
     RequireBlist("ListBlist", blist, "blist");
-    while ( LEN_LIST( list ) != LEN_BLIST( blist ) ) {
-        blist = ErrorReturnObj(
+    if (LEN_LIST(list) != LEN_BLIST(blist)) {
+        ErrorMayQuit(
             "ListBlist: <blist> must have the same length as <list> (%d)",
```
fingolfin: You could also use `CheckSameLength` here (and elsewhere). The main challenge is that it uses `ErrorQuit` right now, not `ErrorMayQuit`. But I think changing `CheckSameLength` to use the latter would actually be an improvement. It is only u sed in 11 places. All of them are covered by tests, and relatively easy to trigger manually, so I'd be confident in making that change.

ChrisJefferson: I like it! Changing.

src/blister.c
```diff
@@ -1218,12 +1209,7 @@ Obj FuncIS_SUB_BLIST (
     /* get and check the arguments                                         */
     RequireBlist("IsSubsetBlist", list1, "blist1");
     RequireBlist("IsSubsetBlist", list2, "blist2");
-    while ( LEN_BLIST(list1) != LEN_BLIST(list2) ) {
-        list2 = ErrorReturnObj(
-        "IsSubsetBlist: <blist2> must have the same length as <blist1> (%d)",
-            LEN_BLIST(list1), 0L,
-            "you can replace <blist2> via 'return <blist2>;'" );
-    }
+    CheckSameLength("IsSubsetBlist", "blist1", "list2", list1, list2);
```
fingolfin:
```suggestion
    CheckSameLength("IsSubsetBlist", "blist1", "blist2", list1, list2);
```

src/blister.c
```diff
@@ -1264,12 +1250,7 @@ Obj FuncUNITE_BLIST (
     /* get and check the arguments                                         */
     RequireBlist("UniteBlist", list1, "blist1");
     RequireBlist("UniteBlist", list2, "blist2");
-    while ( LEN_BLIST(list1) != LEN_BLIST(list2) ) {
-        list2 = ErrorReturnObj(
-           "UniteBlist: <blist2> must have the same length as <blist1> (%d)",
-            LEN_BLIST(list1), 0L,
-            "you can replace <blist2> via 'return <blist2>;'" );
-    }
+    CheckSameLength("UniteBlist", "blist1", "list2", list1, list2);
```
fingolfin:
```suggestion
    CheckSameLength("UniteBlist", "blist1", "blist2", list1, list2);
```

src/blister.c
```diff
@@ -1315,11 +1296,10 @@ Obj FuncUNITE_BLIST_LIST (
...
     RequireBlist("UniteBlistList", blist, "blist");
-    while ( LEN_BLIST(blist) != lenList ) {
-        blist = ErrorReturnObj(
-          "UniteBlistList: <blist> must have the same length as <list> (%d)",
-            lenList, 0L,
-            "you can replace <blist> via 'return <blist>;'" );
+    if (LEN_BLIST(blist) != lenList) {
+        ErrorMayQuit("UniteBlistList: <blist> must have the same length as "
+                     "<list> (%d)",
+                     lenList, 0L);
```
fingolfin: Perhaps use `CheckSameLength` here, too? The overhead of computing the list length again should be negligible, shouldn't it?

src/blister.c
```diff
@@ -1531,12 +1511,7 @@ Obj FuncINTER_BLIST (
     /* get and check the arguments                                         */
     RequireBlist("IntersectBlist", list1, "blist1");
     RequireBlist("IntersectBlist", list2, "blist2");
-    while ( LEN_BLIST(list1) != LEN_BLIST(list2) ) {
-        list2 = ErrorReturnObj(
-       "IntersectBlist: <blist2> must have the same length as <blist1> (%d)",
-            LEN_BLIST(list1), 0L,
-            "you can replace <blist2> via 'return <blist2>;'" );
-    }
+    CheckSameLength("IntersectBlist", "blist1", "list2", list1, list2);
```
fingolfin:
```suggestion
    CheckSameLength("IntersectBlist", "blist1", "blist2", list1, list2);
```

tst/testinstall/kernel/blister.tst
```diff
@@ -123,7 +123,7 @@ Error, ListBlist: <list> must be a small list (not the value 'fail')
 gap> ListBlist([1,2,3], fail);
 Error, ListBlist: <blist> must be a boolean list (not the value 'fail')
 gap> ListBlist([1,2,3], [true,false]);
-Error, ListBlist: <blist> must have the same length as <list> (3)
+Error, ListBlist: <blist> must have the same length as <list> (not 2 and 3)
```
fingolfin: Not your fault, but rather mine (and thus not required for this PR), but I wonder if "(not 2 and 3)" could be improved... e.g. "(not 2 versus 3)"; or "(lengths are 2 and 3)" or something?

PR #2979, opened 2018-11-08
Add doublecoset benchmark

alex-konovalov:
The test from tst/testextra/doublecoset.tst, added in #2741 by @hulpke, requires more memory than in the default `make teststandard` setting, so it is moved to benchmarks, with two more shorter tests added.

### DIFF
--- a/benchmark/doublecoset/doublecoset1.tst
+++ b/benchmark/doublecoset/doublecoset1.tst
@@ -7,13 +7,12 @@
 ##
 ##  This  file  tests for double coset calculations
 ##
-gap> START_TEST("doublecoset.tst");
-gap> g:=SimpleGroup("Co3");;
+gap> START_TEST("doublecoset2.tst");
+gap> g:=SimpleGroup("J2");;
 gap> m:=MaximalSubgroupClassReps(g);;
-gap> u:=First(m,x->Index(g,x)=17931375);;
+gap> u:=First(m,x->Index(g,x)=10080);;
 gap> dc:=DoubleCosetRepsAndSizes(g,u,u);;
 gap> Length(dc);Sum(dc,x->x[2])=Size(g);
-913
+191
 true
-
-gap> STOP_TEST( "doublecoset.tst", 1);
+gap> STOP_TEST( "doublecoset1.tst", 1);
--- a/benchmark/doublecoset/doublecoset2.tst
+++ b/benchmark/doublecoset/doublecoset2.tst
@@ -1,0 +1,18 @@
+#############################################################################
+##
+#W  doublecoset.tst                                     Alexander Hulpke
+##
+##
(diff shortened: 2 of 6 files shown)


benchmark/doublecoset/doublecoset2.tst
```diff
@@ -0,0 +1,18 @@
+#############################################################################
+##
+#W  doublecoset.tst                                     Alexander Hulpke
+##
+##
+#Y  Copyright (C)  1997,  Lehrstuhl D für Mathematik,  RWTH Aachen,  Germany
+##
+##  This  file  tests for double coset calculations
+##
+gap> START_TEST("doublecoset.tst");
```
hulpke: Shouldn't this be `doublecoset2` (though I don't see the merit in splitting the tests up so much

benchmark/doublecoset/doublecoset1.tst
```diff
@@ -0,0 +1,18 @@
...
+##
+##
+#Y  Copyright (C)  1997,  Lehrstuhl D für Mathematik,  RWTH Aachen,  Germany
+##
+##  This  file  tests for double coset calculations
+##
+gap> START_TEST("doublecoset.tst");
+gap> g:=SimpleGroup("J2");;
+gap> m:=MaximalSubgroupClassReps(g);;
+gap> u:=First(m,x->Index(g,x)=10080);;
```
hulpke: This isn't really a benchmark for double cosets, as the index is small enough to even calculate orbits on cosets. (In fact the whole calculation on a group of order barely 600k is a very easy case)
The issue that the new code addresses is with subgroups of large index that do not have satisfactory intermediate subgroups (which is a given if they are maximal), so that a Leiterspiel with up-step is required.

One could try to have larger examples without Leiterspiel (just requiring to find intermediate subgroups), but as long as indices are not in the millions this is all childs play.

alex-konovalov:
@hulpke Thanks - I will adjust the names of the tests before merging this.

The reason to have three tests is that all other benchmarks also have three tests, and it's easy to have a common script in Jenkins which is parametrised by the name of the directory:
```
cd benchmark/$BENCHMARK
../../bin/gap.sh -o 1g test1.g
../../bin/gap.sh -o 1g test2.g
../../bin/gap.sh -o 2g test3.g 
```
and then what I see is

screen shot 2018-11-09 at 11 56 12

The idea is that the three tests are of increasing difficulty. If you could contribute two examples, one faster than Co3, and another either faster or slower, I will be happy to use them as more meaningful!

alex-konovalov:
I've updated test files - now I'd rather merge this and changed examples in benchmarks later, to get it out of the way of `make teststandard` which fails in Travis because of the memory demands of `tst/testextra/doublecoset.tst`.

PR #2985, opened 2018-11-09
List object improvements

ChrisJefferson:
This is a selection of fixes which I came up with while trying to implement an object which acts as a mutable list.

* SortParallel requires IsDenseList. Sort and SortBy only require IsList and work fine, so unify with them.

* IsSSortedList requires the list is homogenous. In the distant past this was required for all lists, and then the requirement was removed for plists, without removing it for plists.

* Make sure we don't call _FILT_LIST on non-plists. To be honest I'm not totally happy with this and considered rewriting this bit of code to make it less fragile, but decided I didn't want to go down that rabbithole. This certainly makes things better than they were.

### DIFF
--- a/src/lists.c
+++ b/src/lists.c
@@ -1348,17 +1348,19 @@ Int IsSSortListDefault (
         return 2L;
     }
 
-    /* a list must be homogeneous to be strictly sorted                    */
-    if ( ! IS_HOMOG_LIST(list) ) {
+    /* get the first element                                               */
+    elm1 = ELM0_LIST(list, 1);
+
+    if (!elm1) {
         return 0L;
     }
 
-    /* get the first element                                               */
-    elm1 = ELMW_LIST( list, 1 );
-
     /* compare each element with its precursor                             */
     for ( i = 2; i <= lenList; i++ ) {
-        elm2 = ELMW_LIST( list, i );
+        elm2 = ELM0_LIST(list, i);
+        if (!elm2) {
+            return 0L;
+        }
         if ( ! LT( elm1, elm2 ) ) {
             return 0L;
         }
--- a/src/lists.h
+++ b/src/lists.h
@@ -19,6 +19,7 @@
 #define GAP_LISTS_H
 
 #include "error.h"
+#include "io.h"
 #include "objects.h"
 
 /****************************************************************************
@@ -814,7 +815,7 @@ extern  Obj             TYPES_LIST_FAM (
 *F * * * * * * * * * * * * * important filters  * * * * * * * * * * * * * * *
 */
 
-enum {
+typedef enum {
     /** filter number for 'IsSSortedList' */
     FN_IS_SSORT,
 
@@ -840,7 +841,7 @@ enum {
     FN_IS_RECT,
 
     LAST_FN = FN_IS_RECT
-};
+} FilterNumber;
 
 
 /****************************************************************************
@@ -854,6 +855,11 @@ enum {
 **
 **  The macro  'SET_FILT_LIST' is  used  to  set  the filter  for a  list  by
 **  changing its type number.
+**
+**  Two values are treated specially:
+**    0 : The default. This tnum should not change.
+**   (UInt)-1 : It is an error to apply this filter (for example,
+**              marking an empty list as not sorted)
 */
 extern UInt SetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
 
@@ -862,17 +868,18 @@ extern UInt SetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
 **
 *F  SET_FILT_LIST( <list>, <fnum> ) . . . . . . . . . . . . . .  set a filter
 */
-#define SET_FILT_LIST(list,fn) \
-  do { \
-    UInt     n; \
-    n = SetFiltListTNums[TNUM_OBJ(list)][fn]; \
-    if ( n != (UInt)-1 ) \
-      RetypeBagIfWritable( list, n ); \
-     else { \
-      Pr( "#E  SET_FILT_LIST[%s][%d] in ", (Int)TNAM_OBJ(list), fn ); \
-      Pr( "%s:%d\n", (Int)__FILE__, (Int)__LINE__); \
-      }  \
-  } while (0)
+static inline void SET_FILT_LIST(Obj list, FilterNumber fn)
+{
+    UInt n = SetFiltListTNums[TNUM_OBJ(list)][fn];
+    if (n == 0) {
+        return;
+    }
+    if (n != (UInt)-1)
+        RetypeBagIfWritable(list, n);
+    else {
+        Pr("#E  SET_FILT_LIST[%s][%d]\n", (Int)TNAM_OBJ(list), fn);
+    }
+}
 
 /****************************************************************************
 **
@@ -891,6 +898,8 @@ extern Obj FuncSET_FILTER_LIST ( Obj self, Obj list, Obj filter );
 **
 **  The macro 'RESET_FILT_LIST' is used  to  set  the filter  for a  list  by
 **  changing its type number.
+**
+**  The same special values are used as SetFiltListTNums.
 */
 extern UInt ResetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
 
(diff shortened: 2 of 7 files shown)


src/lists.c
```diff
@@ -1348,17 +1348,19 @@ Int IsSSortListDefault (
         return 2L;
     }
 
-    /* a list must be homogeneous to be strictly sorted                    */
-    if ( ! IS_HOMOG_LIST(list) ) {
+    /* get the first element                                               */
+    elm1 = ELM0_LIST( list, 1 );
```
fingolfin: I am slightly puzzled by this change; couldn't you simply replace `IS_HOMOG_LIST` by `IS_DENSE_LIST`?

src/listfunc.c
```diff
@@ -640,8 +640,10 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
 #define SORT_COMP(v, w) CALL_2ARGS(func, v, w) == True
 /* list is not necc. sorted wrt. \< (any longer) */
 #define SORT_FILTER_CHECKS() \
-  RESET_FILT_LIST(list, FN_IS_SSORT); \
-  RESET_FILT_LIST(list, FN_IS_NSORT);
+  if(IS_PLIST(list)) {\
```
fingolfin:
```suggestion
  if (IS_PLIST(list)) {\
```
... or just `clang-format`

fingolfin: Or don't make this change, but simply fix `RESET_FILT_LIST`, it's a trivial change?

src/listfunc.c
```diff
@@ -711,9 +713,13 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
     /* if list was ssorted, then it still will be,
        but, we don't know anything else any more */
 #define SORT_FILTER_CHECKS() \
-  RESET_FILT_LIST(list, FN_IS_NSORT); \
-  RESET_FILT_LIST(shadow, FN_IS_SSORT); \
-  RESET_FILT_LIST(shadow, FN_IS_NSORT);
+  if(IS_PLIST(list)) {\
```
fingolfin:
```suggestion
  if (IS_PLIST(list)) {\
```

src/listfunc.c
```diff
@@ -711,9 +713,13 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
...
+    RESET_FILT_LIST(list, FN_IS_NSORT); \
+  } \
+  if(IS_PLIST(shadow)) {\
```
fingolfin:
```suggestion
  if (IS_PLIST(shadow)) {\
```

src/listfunc.c
```diff
@@ -754,10 +760,14 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
 #define SORT_COMP(v, w) CALL_2ARGS( func, v, w ) == True
 /* list is not necc. sorted wrt. \< (any longer) */
 #define SORT_FILTER_CHECKS() \
+  if(IS_PLIST(list)) {\
```
fingolfin:
```suggestion
  if (IS_PLIST(list)) {\
```

src/listfunc.c
```diff
@@ -754,10 +760,14 @@ Obj             FuncPOSITION_FIRST_COMPONENT_SORTED (
...
     RESET_FILT_LIST(list, FN_IS_SSORT); \
     RESET_FILT_LIST(list, FN_IS_NSORT); \
+  } \
+  if(IS_PLIST(shadow)) {\
```
fingolfin:
```suggestion
  if (IS_PLIST(shadow)) {\
```

src/lists.h
```diff
@@ -840,7 +841,7 @@ enum {
     FN_IS_RECT,
 
     LAST_FN = FN_IS_RECT
-};
+} EnumFnIs;
```
fingolfin:
```suggestion
} FilterNumber;
```

src/lists.h
```diff
@@ -862,17 +863,16 @@ extern UInt SetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
...
-      RetypeBagIfWritable( list, n ); \
-     else { \
-      Pr( "#E  SET_FILT_LIST[%s][%d] in ", (Int)TNAM_OBJ(list), fn ); \
-      Pr( "%s:%d\n", (Int)__FILE__, (Int)__LINE__); \
-      }  \
-  } while (0)
+static inline void SET_FILT_LIST(Obj list, EnumFnIs fn)
+{
+    UInt n = SetFiltListTNums[TNUM_OBJ(list)][fn];
+    GAP_ASSERT(n != 0);
```
fingolfin: How about replacing the `GAP_ASSERT` with his: `if (n == 0) return;`.  Then you are free to call `SET_FILT_LIST` on arbitrary objects:

src/lists.h
```diff
@@ -899,18 +899,16 @@ extern UInt ResetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
...
-    else  { \
-      Pr( "#E  RESET_FILT_LIST[%s][%d] in ", (Int)TNAM_OBJ(list), fn ); \
-      Pr( "%s:%d\n", (Int)__FILE__, (Int)__LINE__); \
-      }  \
-  } while (0)
-
+static inline void RESET_FILT_LIST(Obj list, EnumFnIs fn)
+{
+    UInt n = ResetFiltListTNums[TNUM_OBJ(list)][fn];
+    GAP_ASSERT(n != 0);
```
fingolfin: See above: just replace this by
```c
    if (n == 0)
        return;
```

src/lists.h
```diff
@@ -944,17 +942,14 @@ extern UInt ClearFiltsTNums [ LAST_REAL_TNUM ];
...
-      RetypeBag( list, n ); \
-/*    else if ( n < 0 ) { \
-      Pr( "#E  CLEAR_FILTS_LIST[%s] in ", (Int)TNAM_OBJ(list), 0 ); \
-      Pr( "%s:%d\n", (Int)__FILE__, (Int)__LINE__); \
-      } */ \
-  } while (0)
+static inline void CLEAR_FILTS_LIST(Obj list)
+{
+    UInt n = ClearFiltsTNums[TNUM_OBJ(list)];
+    GAP_ASSERT(n != 0);
```
fingolfin: ... and again, see above (only this time, you can simply drop the assert)

tst/testinstall/objlist.tst
```diff
@@ -0,0 +1,82 @@
+gap> DeclareCategory("IsTestListObj", IsSmallList);;
```
fingolfin: Could you pleas add a brief comment here which explains what is going on below? Something like (assuming I guess right):
> In this test, we set up a custom small list implementation and use it to test that various GAP features work correctly with non-internal lists.

src/lists.h
ChrisJefferson: I'm not 100% clear on what all this code does, but there are clearly some objects we aren't supposed to call this function on, the ones where `n = (UInt)-1`, as we print a mandatory warning message in those cases. Therefore I was worried about starting to accept values which shouldn't be passed in.

src/lists.c
ChrisJefferson: because IsDenseList is a category

ChrisJefferson: Also (sorry for multiple comments) it's not clear to me it's worth using IS_DENSE_LIST here, because if we don't already know it we'll have to scan the list twice, and we are already going to pass the list once anyway.

src/listfunc.c
ChrisJefferson: made change

src/lists.h
ChrisJefferson: did it anyway

ChrisJefferson:
Fixes made. I'm sure a more complete overhall of lots of list based stuff could be made (and reading comments in the source, many things were never finished in the first place), but this adds enough functionality for many basic cases, including the cases I want to work.

src/lists.h
```diff
@@ -862,17 +863,18 @@ extern UInt SetFiltListTNums [ LAST_REAL_TNUM ] [ LAST_FN + 1 ];
...
-      Pr( "#E  SET_FILT_LIST[%s][%d] in ", (Int)TNAM_OBJ(list), fn ); \
-      Pr( "%s:%d\n", (Int)__FILE__, (Int)__LINE__); \
-      }  \
-  } while (0)
+static inline void SET_FILT_LIST(Obj list, FilterNumber fn)
+{
+    UInt n = SetFiltListTNums[TNUM_OBJ(list)][fn];
+    if (n == 0) {
+        return;
+    }
```
fingolfin: The commit message still says "Add Assertions to *_FILT_LIST macros" and should be rewritten

src/plist.c
```diff
@@ -2179,7 +2178,7 @@ Int             IsSSortPlist (
 	if (isHom)
 	  SET_FILT_LIST( list, FN_IS_HOMOG);
 	else
-	  SET_FILT_LIST( list, FN_IS_HOMOG);
+	  SET_FILT_LIST( list, FN_IS_NHOMOG);
```
fingolfin: The commit message here just says "Fix typo in IsSSortPlist" which IMHO makes it sound completely harmless, but it really is an actual bug that's being fixed here, now? Perhaps "Fix bug in IsSSortPlist" or "Fix typo bug in IsSSortPlist" or so?

tst/testinstall/objlist.tst
```diff
@@ -0,0 +1,84 @@
+# Create a list-like object, to test functionality of GAP for Objects which implement
```
fingolfin:
```suggestion
# Create a list-like object, to test functionality of GAP for objects which implement
```

src/plist.c
ChrisJefferson: You are right, this is a bug (although I believe caused by a typo, but many bugs are.

PR #2986, opened 2018-11-09
Fix readline crash

ChrisJefferson:
This stops readline crashing if set colored-completion-prefix on
is in ~/.inputrc. However, it doesn't make that option work, as
(I believe) the first thing passed to rl_display_match_list should
be the current partially written word.

I am not going to further improve this PR, but I am happy or someone else to, or for it to be merged as is so at least we don't crash.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -2118,8 +2118,15 @@ int GAP_rl_func(int count, int key)
          /* display matches */
          if (!IS_LIST(data)) return 0;
          dlen = LEN_LIST(data);
-         char **strs = (char**)malloc((dlen+1) * sizeof(char*));
+         // +2, so we have a null at the end, as this must be
+         // in 'argv' format
+         char **strs = (char**)calloc(dlen+2, sizeof(char*));
          max = 0;
+         // TODO: Make this the current partially typed
+         // identifier which is, I think, the first argument to
+         // rl_display_match_list
+         char partial[] = {'\0'};
+         strs[0] = partial;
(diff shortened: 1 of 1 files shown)


src/sysfiles.c
```diff
@@ -2118,8 +2118,13 @@ int GAP_rl_func(int count, int key)
          /* display matches */
          if (!IS_LIST(data)) return 0;
          dlen = LEN_LIST(data);
-         char **strs = (char**)malloc((dlen+1) * sizeof(char*));
+         // +2, so we have a null at the end, as this must be
+         // in 'argv' format
+         char **strs = (char**)calloc(dlen+2, sizeof(char*));
          max = 0;
+         // TODO: Make this the partially typed word
```
fingolfin: that's rather cryptic. Perhaps a tiny bit more context could be given? E.g. perhaps mentioning `rl_display_match_list` would be helpful?

ChrisJefferson: done. I've also made clear that, to be honest, I'm not 100% sure. I deduced what there should be by reading the source of rl_display_match_list in readline 7, as this function seems to have no (useful) documentation I can find.

fingolfin:
This code was originally written by @frankluebeck so perhaps he has further comments or insights? Otherwise, I'd just merge this next week.

frankluebeck:
See "Fl readline fix" (#2991) for a comment and an alternative pull request.

ChrisJefferson:
closed because #2991 is better

PR #2988, opened 2018-11-11
Add some functions to libgap API

fingolfin:
Also make `GAP_CSTR_STRING` easier to use (by not crashing for non-string inputs), and tweak testlibgap

This wraps some functions for which I have seen used "in the wild" in various places (mostly SageMath, and GAPJulia). Of course this is not yet enough, and some things should be tweaked, but I think it gets us closer.

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -2,9 +2,11 @@
 
 #include "libgap-api.h"
 
+#include "ariths.h"
 #include "bool.h"
 #include "opers.h"
 #include "calls.h"
+#include "gap.h"
 #include "gapstate.h"
 #include "gvars.h"
 #include "lists.h"
@@ -15,39 +17,26 @@
 //
 // Setup and initialisation
 //
-void GAP_Initialize(int          argc,
-                    char **      argv,
-                    char **      env,
-                    CallbackFunc markBagsCallback,
-                    CallbackFunc errorCallback)
+void GAP_Initialize(int              argc,
+                    char **          argv,
+                    char **          env,
+                    GAP_CallbackFunc markBagsCallback,
+                    GAP_CallbackFunc errorCallback)
 {
     InitializeGap(&argc, argv, env);
     SetExtraMarkFuncBags(markBagsCallback);
     STATE(JumpToCatchCallback) = errorCallback;
+
+    GAP_True = True;
+    GAP_False = False;
+    GAP_Fail = Fail;
 }
 
 
-// Combines GVarName and ValGVar. For a given string, it returns the value
-// of the gvar with name <name>, or NULL if the global variable is not
-// defined.
-Obj GAP_ValueGlobalVariable(const char * name)
-{
-    UInt gvar = GVarName(name);
-    // TODO: GVarName should never return 0?
-    if (gvar != 0) {
-        return ValGVar(gvar);
-    }
-    else {
-        return NULL;
-    }
-}
+////
+//// program evaluation and execution
+////
 
-//
-// Evaluate a string of GAP commands
-//
-// To see an example of how to use this function
-// see tst/testlibgap/basic.c
-//
 Obj GAP_EvalString(const char * cmd)
 {
     Obj instream;
@@ -62,26 +51,210 @@ Obj GAP_EvalString(const char * cmd)
     return res;
 }
 
-//
-// Returns the GAP object containing
-// <string>
+
+////
+//// variables
+////
+
+Obj GAP_ValueGlobalVariable(const char * name)
+{
+    UInt gvar = GVarName(name);
+    // TODO: GVarName should never return 0?
+    if (gvar != 0) {
+        return ValGVar(gvar);
+    }
+    else {
+        return NULL;
+    }
+}
+
+
+////
+//// arithmetic
+////
+
+int GAP_EQ(Obj a, Obj b)
+{
+    return EQ(a, b);
+}
+
+int GAP_LT(Obj a, Obj b)
+{
+    return LT(a, b);
+}
+
+int GAP_IN(Obj a, Obj b)
+{
+    return IN(a, b);
+}
+
+Obj GAP_SUM(Obj a, Obj b)
+{
+    return SUM(a, b);
+}
+
+Obj GAP_DIFF(Obj a, Obj b)
+{
+    return DIFF(a, b);
+}
+
+Obj GAP_PROD(Obj a, Obj b)
+{
+    return PROD(a, b);
+}
+
+Obj GAP_QUO(Obj a, Obj b)
+{
+    return QUO(a, b);
+}
+
+Obj GAP_LQUO(Obj a, Obj b)
+{
+    return LQUO(a, b);
+}
+
+Obj GAP_POW(Obj a, Obj b)
+{
+    return POW(a, b);
+}
+
+Obj GAP_COMM(Obj a, Obj b)
+{
+    return COMM(a, b);
+}
+
+Obj GAP_MOD(Obj a, Obj b)
+{
+    return MOD(a, b);
+}
+
+
+////
+//// booleans
+////
+
+Obj GAP_True;
+Obj GAP_False;
+Obj GAP_Fail;
+
+
+////
+//// calls
+////
+
+Obj GAP_CallFuncList(Obj func, Obj args)
+{
+    return CallFuncList(func, args);
+}
+
+Obj GAP_CallFuncArray(Obj func, UInt narg, Obj args[])
+{
+    Obj result;
+    Obj list;
+
+    if (TNUM_OBJ(func) == T_FUNCTION) {
+
+        // call the function
+        switch (narg) {
+        case 0:
+            result = CALL_0ARGS(func);
+            break;
+        case 1:
+            result = CALL_1ARGS(func, args[0]);
+            break;
+        case 2:
+            result = CALL_2ARGS(func, args[0], args[1]);
+            break;
+        case 3:
+            result = CALL_3ARGS(func, args[0], args[1], args[2]);
+            break;
+        case 4:
+            result = CALL_4ARGS(func, args[0], args[1], args[2], args[3]);
+            break;
+        case 5:
+            result =
+                CALL_5ARGS(func, args[0], args[1], args[2], args[3], args[4]);
+            break;
+        case 6:
+            result = CALL_6ARGS(func, args[0], args[1], args[2], args[3],
+                                args[4], args[5]);
+            break;
+        default:
+            list = NewPlistFromArray(args, narg);
+            result = CALL_XARGS(func, list);
+        }
+    }
+    else {
+        list = NewPlistFromArray(args, narg);
+        result = DoOperation2Args(CallFuncListOper, func, list);
+    }
+
+    return result;
+}
+
+
+////
+//// lists
+////
+
+int GAP_IsList(Obj obj)
+{
+    return obj && IS_LIST(obj);
+}
+
+UInt GAP_LenList(Obj obj)
+{
+    return LEN_LIST(obj);
+}
+
+void GAP_AssList(Obj list, UInt pos, Obj val)
+{
+    if (val)
+        ASS_LIST(list, pos, val);
+    else
+        UNB_LIST(list, pos);
+}
+
+Obj GAP_ElmList(Obj list, UInt pos)
+{
+    if (pos == 0)
+        return 0;
+    return ELM0_LIST(list, pos);
+}
+
+Obj GAP_NewPlist(Int capacity)
+{
+    return NEW_PLIST(T_PLIST_EMPTY, capacity);
+}
+
+
+////
+//// strings
+////
+
+int GAP_IsString(Obj obj)
+{
+    return obj && IS_STRING_REP(obj);
+}
+
+UInt GAP_LenString(Obj obj)
+{
+    return GET_LEN_STRING(obj);
+}
+
 Obj GAP_MakeString(const char * string)
 {
     return MakeString(string);
 }
 
-//
-// Returns a pointer to the
-// contents of the GAP string object <string>
-char * GAP_CSTR_STRING(Obj string)
+Obj GAP_MakeImmString(const char * string)
 {
-    return CSTR_STRING(string);
+    return MakeImmString(string);
 }
 
-//
-// Returns a new empty plain list
-// with capacity <capacity>
-Obj GAP_NewPlist(Int capacity)
+char * GAP_CSTR_STRING(Obj string)
 {
-    return NEW_PLIST(T_PLIST_EMPTY, capacity);
+    if (!IS_STRING_REP(string))
+        return 0;
+    return CSTR_STRING(string);
 }
--- a/src/libgap-api.h
+++ b/src/libgap-api.h
@@ -1,24 +1,151 @@
-// LibGAP API - API for using GAP as shared library.
+//// LibGAP API - API for using GAP as shared library.
 
 #ifndef LIBGAP_API_H
 #define LIBGAP_API_H
 
-#include "gap.h"
+#include "system.h"
 
-typedef void (*CallbackFunc)(void);
 
-// Initialisation and finalization
+////
+//// Setup and initialisation
+////
 
-void GAP_Initialize(int          argc,
-                    char **      argv,
-                    char **      env,
-                    CallbackFunc markBagsCallback,
-                    CallbackFunc errorCallback);
+typedef void (*GAP_CallbackFunc)(void);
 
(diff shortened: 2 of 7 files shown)


src/libgap-api.c
```diff
@@ -42,46 +68,135 @@ Obj GAP_ValueGlobalVariable(const char * name)
...
+{
+    return obj && IS_LIST(obj);
+}
+
+Int GAP_LenList(Obj obj)
+{
+    return LEN_LIST(obj);
+}
+
+void GAP_AssList(Obj list, UInt pos, Obj val)
```
fingolfin: Here, I change the names, from e.g. `ASS_LIST` to `AssList`, based on the rule that for other GAP symbols, the `SNAKE_CASE` ones are low-level, while the `CamelCase` ones are ready for use in higher-level code.

Is that a good or a bad idea? You decide -- of course I can also change it to `GAP_ASS_LIST` etc.

src/libgap-api.c
```diff
@@ -42,46 +68,135 @@ Obj GAP_ValueGlobalVariable(const char * name)
...
+{
+    if (val)
+        ASS_LIST(list, pos, val);
+    else
+        UNB_LIST(list, pos);
```
fingolfin: Here, I though that unbinding (a rare operation, I believe) could be transparently achieved by the same function as is used for assignment. Good/bad idea?

src/libgap-api.c
```diff
@@ -42,46 +68,135 @@ Obj GAP_ValueGlobalVariable(const char * name)
...
+}
+
+Obj GAP_ElmList(Obj list, UInt pos)
+{
+    return ELM0_LIST(list, pos);
```
fingolfin: Note that we use ELM0_LIST here, not ELM_LIST, as that makes it easier to write code using it, w/o having to add range check and `IsBound` checks (indeed, one would have to export IsBound in the first place... still might want to do that later, but for now, it's not there, as I never had reason to use it in an interface myself).

BTW, I just realize that there either should be a check for `pos == 0`, or it should be documented what happens if one passes in 0. Same for `GAP_AssList`

src/libgap-api.h
```diff
@@ -1,24 +1,113 @@
...
+extern int GAP_IN(Obj a, Obj b);
+
+extern Obj GAP_SUM(Obj a, Obj b);
+extern Obj GAP_DIFF(Obj a, Obj b);
+extern Obj GAP_PROD(Obj a, Obj b);
+extern Obj GAP_QUO(Obj a, Obj b);
+extern Obj GAP_LQUO(Obj a, Obj b);
+extern Obj GAP_POW(Obj a, Obj b);
+extern Obj GAP_COMM(Obj a, Obj b);
+extern Obj GAP_MOD(Obj a, Obj b);
```
fingolfin: TODO: the above should get documentation comments. Also, there is still stuff from ariths.h missing which might be of interest -- or not... But hey, it's a start.

src/libgap-api.h
```diff
@@ -1,24 +1,113 @@
...
+////
+//// program evaluation and execution
+////
+
+// Evaluate a string of GAP commands
+//
+// To see an example of how to use this function see tst/testlibgap/basic.c
+extern Obj GAP_EvalString(const char * cmd);
+
+
```
fingolfin: One major thing missing here is a wrapper for `CallFuncList`. Actually, two variants could be interesting (and perhaps both should be provided): one is a straight wrapper for `CallFuncList`, which hence takes a plist of arguments as input. The other is a variant `CallFuncListWithArgsArray(Obj func, UInt narg, Obj * args)` which gets a pointer to an array of args. The latter can be more convenient and efficient in practice. Depends a bit on the client, though... Wonder what SageMath people would prefer, if at all.

fingolfin:
Also CC: @nthiery

nthiery:
@jdemeyer, @dimpase: from your ongoing work on libgap in Sage you may have insight  on some of the questions raised above by @fingolfin about the desirable API for GAP.

src/libgap-api.c
markuspf: I have an aesthetic reason to approve the CamelCase (which is probably the reason why `SNAKE_CASE` is used for internal functions, it's really grating)

src/libgap-api.c
markuspf: We discussed such behaviour before, where code like
```
list[5] := SomeFunction("foo");
```
was supposed to unbind `list[5]` if `SomeFunction` did not return a value (i.e. 0). I don't particularly like it.

src/libgap-api.h
markuspf: That wrapper was in muliple versions of the API, but without the efficiency improvement with `Obj * args`. Arguably this is *the* most important API function because once one can create primitive values and have access to global variables one can do everything with GAP.

src/libgap-api.c
```diff
@@ -42,46 +68,135 @@ Obj GAP_ValueGlobalVariable(const char * name)
...
+    return MakeImmString(string);
 }
 
 //
-// Returns a new empty plain list
-// with capacity <capacity>
-Obj GAP_NewPlist(Int capacity)
+// Returns a pointer to the contents of the GAP string object <string>,
+// or returns 0 if <string> is not a GAP string object.
+char * GAP_CSTR_STRING(Obj string)
```
markuspf: Isn't this very dangerous? If anything causes a garbage collection after one retrieved this pointer, it will just point *somewhere*. Also the memory area pointed to can be modified in whichever way the user wants.

Maybe sometimes this might be exactly what one wants (when?), but one might just copy the string out and not have an API function haphazardly grant access to GAP objects; 

I had a bit of a fiddle with code that one passes a buffer to (for which one might need to determine the size beforehand using a `GAP_Length` function). It feels awkward as well, the question here really is which usecases one has.

dimpase: It is as dangerous as having `Obj *bla` in the code calling libgap, but this kind of use is to be combined with a GC callback, which would go over such pointers in the code and mark bags they point to. (This is what Sagemath interface does: it has a ref-counted store of "handles" of GAP objects, and a GASMAN callback set to mark bags of stored objects).

So I'm OK with it. 

For one thing, such a function to make bags should be in libgap API.

src/libgap-api.c
fingolfin: I don't think that's the same discussion at all. What you are talking about is a change on the GAP language level; and I agree that we rejected that idea, and I agree with that rejection.

There is no reason why this must dictate what the C API does, though. In practice, unbinding *is* the same as assigning NULL. Just because we don't expose this on the GAP language level does not justify that we must contort ourselves on the C level. That said, there may be valid reasons to reject this, but I don't see how what you mention is one. Perhaps if you elaborate as to what the actual problems you are concerned about are...?

src/libgap-api.h
fingolfin: Agreed. The `Obj * args*` is the most likely for use in GAPJulia's `JuliaInterface -- subject to benchmarking, we might also stay with just using the non-official API, like any other GAP kernel extension (for obvious reasons, I am personally not very concerned about using some kind of "unofficial" API...). So I would use this `CallFuncList` if the measured overhead over our current solution is negligible. But I'd be willing to provide this API anyway, but only if somebody else is interested in using it... E.g. you, or the SageMath folks, or somebody else.

src/libgap-api.c
fingolfin: First off, `GAP_CSTR_STRING` is not something that *I* added, I only moved the comment to `.h` file, made it not crash on invalid argument, and documented the error behavior.

I am actually not that happy with `GAP_CSTR_STRING` myself, e.g. I don't like the name; but since it is now in a released version of GAP, I did not want to change it, which would be a breaking change after all. This is precisely the risk I warned about when I said that I'd rather be slow in adding APIs, and only want to do so after being sure somebody would, or at least might reasonably, use them; this is the case for all the APIs I added here, I created them based on inspecting the GAPJulia and SageMath code. Note that the 4.11 development cycle has only begun now, so we can still remove any and all of these new APIs before the final release, so I am not quite as concerned as with the stuff in 4.10, or possibly being backported there.

Secondly, I don't think it is any more dangerous than almost any other code directly accessing the body of an allocation in a C code base using a GC. E.g. in Julia, the C kernel also has to be aware of it. So, I think this is fine, though of course the limitations need to be documented, I'll do that ASAP.
That said, one could also implement a copying API, but that only makes sense if we also export a function to query the length of a string (which we must anyway -- actually, I thought I added one, but it seems I forgot). One thing this can't do effectively, though, is to e.g. quickly scan the content of a string for something -- though indeed it is not that clear to me whether this is all that useful in practice, but at the same time, I don't really see a practical problem with allowing it?

@dimpase That GC marking does not help with (and in fact, is irrelevant for) the issue @markuspf is thinking about (I believe): namely, that the GAP GC is a *moving* GC. So if a GC happens, the pointer returned by `GAP_CSTR_STRING` may no longer be valid.

ChrisJefferson: One thing I will say, is that holding char* pointers to GAP strings has been a source of many bugs in the GAP Kernel (i went through and fixed many such bugs some time ago) -- this is one of the few commonly used GAP functions which returns a pointer which must not be held over a GC.

src/libgap-api.c
ChrisJefferson: I agree with camelcase as well

src/libgap-api.c
ChrisJefferson: I agree we definately need to check for pos == 0 (I do that all the time when writing extensions). I'm not sure if ELM0_LIST should catch that, or it should be caught here (I'm tempted to say ELM0_LIST should catch it)

src/libgap-api.c
dimpase: oops, thanks for reminding about GC moving... 😰 
It might partially explain trouble I am having with Sage’s interface to libgap.

src/libgap-api.c
fingolfin: Hrm... Elm0Plist, ElmPlist, ElmPlistDense, all check that the index does not exceed the length, but none of them (nor their counterparts for blists, strings, ranges). This might or might not be a relevant performance optimization...

As such, I will for now simply add that check in `libgap-api.c`. I am tempted to say that in this case, 0 should be returned, as for all other "out of bounds" accesses (note that if one accidentally passes a negative index to this function, it'll be treated as a big positive integer, which then leads to 0 being returned, too). Of course it'd be trivial to instead call `ErrorMayQuit`, if people think that's better.

fingolfin:
@ChrisJefferson your push made this PR fail, and also was redundant. I will replace it with my next force push.

src/libgap-api.h
fingolfin: Ah man, I caved in and added `GAP_CallFuncList` and `GAP_CallFuncArray`.

src/libgap-api.c
ChrisJefferson: Sorry, I didn't mean to push it, I meant to just leave a suggestion (which wasn't a good suggestion, as it didn't compile of course..)

I would prefer const char*, because in my experience of GAP internals, over the years we have been very bad at keeping these char*s too long. If they are const at least the worst is a corrupted reading of a string, instead of randomly corrupting a GAP object which the pointer points to.

ChrisJefferson: On a more minor note, I wonder if it would be better to put no docs in the .c file, rather than partial docs when we probably want people to go read the full doc in the .h file.

embray: @dimpase FWIW there is a bug in your current code related to that, or possibly in my updates to your code.  In the error handler I made sure when using `CSTR_STRING` to immediately copy the referenced C string (our `char_to_str` function does this implicitly).  So I call `char_to_str` immediately, while the GAP string still has a reference to it.  Afterwards it may go away.

markuspf: I am perfectly happy to change the name btw, I don't think we have committed to a `libgap` API at all yet.

fingolfin: @ChrisJefferson re comment, I agree, In fact I actually already removed all non-grouping comments in the .c file (well, except for one I overlooked; now also removed)

fingolfin:
To clarify the intent of this PR: I really hope that SageMath (ping @dimpase @embray @nthiery ) can and will use some of these. Now, I realize that at the moment, you really want to first get 4.10.0 working with Sage, which makes sene -- but after you are done with that, or perhaps even during/instead of that, you could also look into adapting some of the new APIs added in this PR. And also suggest further ones (also taking https://hackmd.io/s/SJzZVOLem into account, possibly editing it as things progress). If those new functions turn out to be useful and "good"/"right"/"appropriate" for your use, we could then look into backporting them to 4.10.1; if they are not as useful, we could look into improving them or removing them again, based on your *constructive* feedback.

Note: Normally, I'd not consider a non-bug fix backport to the stable branch; but since this is completely isolated code which affects no other part of GAP, I'd be willing to make an exception, *if it helps somebody* -- but of course if nobody uses it, then it doesn't make sense to backport. Also, it is of course not just up to me to decide this, and perhaps @markuspf @ChrisJefferson or others have objections to that idea?

markuspf:
I am open to backporting (as you say, if it helps someone), but I will mention that we'll probably start making changes to signal handling or inputs/outputs, or error handling to make GAP remotely useful as a library. 
Those changes will just be too cumbersome to backport. Assuming that GAP 4.11 will be in around 6 months time, I'd not invest too much time into backporting.

src/libgap-api.h
```diff
@@ -1,24 +1,151 @@
...
+// If <val> is zero, then this unbinds the list entry.
+// If <list> is not a GAP list, an error may be raised.
+extern void GAP_AssList(Obj list, UInt pos, Obj val);
+
+// Returns the element at the position <pos> in the list <list>.
+// Returns 0 if there is no entry at the given position.
+// Also returns 0 if <pos> is out of bounds, i.e., if <pos> is zero,
+// or larger than the length of the list.
+// If <list> is not a GAP list, an error may be raised.
+extern Obj GAP_ElmList(Obj list, UInt pos);
```
embray: Perhaps relevant here: An annoyance I've discovered is that if you try to access an element of a plist that has not actually been assigned to, then it raises a breaking error in GAP.  I don't think this should happen, at least for the API.  It should just return `NULL` or something. 

This happens when trying to process the list returned from `GAP_EvalString`.  If an expression that was evaluated returns void, then the `READ_ALL_COMMANDS` never actually assigns to the value entry in the result list

```
 182
 183                 if (evalResult) {
 184                     AssPlist(result, 2, evalResult);
 185                 }
 186
```

So if I later call something like `ELM_LIST(result, 2)` it actually results in an error.  But if I'm looking at the results of a `GAP_EvalString` the last thing I want to do is have my code jump back into the GAP interpreter which is exactly what happens here.  I can avoid this by using `ELM_PLIST` directly instead, but that isn't really desirable for the API.

fingolfin: @embray this works exactly as designed and documented. DON'T USE `ELM_LIST` if yo don't want this behvarious. Use `ELM0_LIST`. And perhaps read the comments in `lists.h`, which explicit specify and document this behavior.

Well, or use `GAP_ElmList`, which does what you want, too (by wrapping `ELM0_LIST`)

markuspf: Note that (as the comment says!) the problem you describe does *not* happen with the API as implemented in this PR.

I also note that, while fixing the problem here, one'll have to do more generic error handling *some way* in the end. You can deactivate the break loop, so you'll not end up in it (`--nointeract` deactivates the break loop), and avoid this `is unbound` error, but you'll always have the potential that GAP code runs into an error, and there is currently no good way to handle this.

embray: I should add, it would help if the `CATCH_ERROR` macro cleared--if appropriate--the `STATE(ReadJmpError)` buffer (and that it were not jumped to if it's unset of course).  This results in errors that occur in functions like `ELM_LIST` causing longjmps to wherever the last `TRY_IF_NO_ERROR` was; often totally arbitrary.

embray: @fingolfin I see now that `GAP_ElmList` works as I would want it to.  I didn't see `ELM0_LIST` (ofc, it's not obvious what that should even mean).

markuspf:
Don't know why travis notifications got their underwear in a twist. 

I am going to merge this now, as further improvements can be done in separate PRs (and discussion about features should be had in separate PRs or better forums such as the mailing list or slack)

embray:
Where would be the best place to report additional needs I've run across?  In this PR?  In a new one?
I could also submit a PR of course, but I don't want to step on any toes.

dimpase:
A new PR is probably the best, one against the master branch.

fingolfin:
Both as issue or via a PR would be fine. Thanks!

PR #2989, opened 2018-11-11
Add new IsSimpleSemigroup method in lib/reesmat.gi

wilfwilson:
The method installed in `lib/reesmat.gi` for `IsSimpleSemigroup` for `IsReesMatrixSubsemigroup and HasUnderlyingSemigroup` is definitely the best method for this type of argument. However, depending on the rank of `RankFilter(IsFinite)`, this method is sometimes beaten by the Semigroups package method for `IsSimpleSemigroup` for `IsSemigroup and IsFinite`. This was causing a test in `tst/teststandard/reesmat.tst` to run out of memory when all packages were loaded, since the Semigroups package was being used when it shouldn't have been used.

This issue is resolved by adding an identical method for `IsSimpleSemigroup` that additionally requires the filter `IsFinite`.

This resolves the problem reported in https://github.com/gap-system/gap/issues/2818#issuecomment-437592480 by @alex-konovalov.

### DIFF
--- a/lib/reesmat.gi
+++ b/lib/reesmat.gi
@@ -123,6 +123,19 @@ InstallMethod(IsSimpleSemigroup,
 [IsReesMatrixSubsemigroup and HasUnderlyingSemigroup],
 R-> IsSimpleSemigroup(UnderlyingSemigroup(R)));
 
+# This next method for `IsSimpleSemigroup` additionally requires the filter
+# `IsFinite`, but is otherwise identical.  In the Semigroups package, there are
+# some more general methods installed for `IsSimpleSemigroup` which include the
+# filter `IsFinite`. When the rank of `IsFinite` is sufficiently large, these
+# methods can beat the above method. The above method is a more specific method
+# and should always be the one chosen for Rees matrix subsemigroups with known
+# underlying semigroup, whether finite or infinite.
+
+InstallMethod(IsSimpleSemigroup, 
+"for finite subsemigroup of a Rees matrix semigroup with underlying semigroup", 
+[IsReesMatrixSubsemigroup and HasUnderlyingSemigroup and IsFinite],
+R -> IsSimpleSemigroup(UnderlyingSemigroup(R)));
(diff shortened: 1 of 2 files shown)


stevelinton:
Would it be better to simply install the method again for `IsReesMatrixSubsemigroup and HasUnderlyingSemigroup and IsFinite`? It avoids any sensitivity to the rank of `IsFinite` (which may change as new implications are installed) or to implications of `IsReesMatrixSemigroup and IsFinite` for instance.

wilfwilson:
Yeah that's a better idea, especially as it is a very simple method.

lib/reesmat.gi
```diff
@@ -123,6 +123,11 @@ InstallMethod(IsSimpleSemigroup,
 [IsReesMatrixSubsemigroup and HasUnderlyingSemigroup],
 R-> IsSimpleSemigroup(UnderlyingSemigroup(R)));
 
+InstallMethod(IsSimpleSemigroup,
```
fingolfin: The problem with this is that if somebody sees this code in the future, they may wonder "huh, why is this duplicated?" and remove the new method again. I see two courses to protected against that:
1. Add a comment explaining the reason,
2. don't add this new method, instead adjust the rank of the existing method, by using a dynamic rank adjustment via  `{} -> RankFilter(IsFinite)`.

wilfwilson: This PR initially implemented item 2: I added the line `{} -> RankFilter(IsFinite)`. After @stevelinton's comment https://github.com/gap-system/gap/pull/2989#issuecomment-437803909 I changed it to how it is now. Therefore I'll implement your suggestion 1.

wilfwilson:
@fingolfin I added an explanatory comment.

alex-konovalov:
@wilfwilson yes, thanks - test pass now, e.g. https://travis-ci.org/gap-system/gap-docker-master-testsuite/jobs/455982635

PR #2991, opened 2018-11-11
Fl readline fix

frankluebeck:
This is an alternative for the pull request "Fix readline crash" #2986 by @ChrisJefferson.

@ChrisJefferson: Thanks for the report! I very rarely had crashes from completions before (also without the 'colored-completion-prefix' switch of readline, but could never reproduce them.

I agree that your proposed fix (#2986) should avoid the crashes, but as you already mentioned, the 'colored-completion-prefix'  switch does not yet work. This should be fixed by the present patch.

A second commit also fixes the problem that GAP did not read a '$if GAP' section of '~/.inputrc' on startup.


ChrisJefferson:
Thanks, this is much better than my PR.

alex-konovalov:
Backported in 4c26a9cb67608a0cdcd43423e8713614c99531d5 and 4191f41513a46bf970b93b03fdcc982e1ea99c26

PR #2992, opened 2018-11-12
Officially document RationalCanonicalFormTransform and SimplexMethod; some minor tweaks

hulpke:
Put `RationalCanonicalFormTransform` and `SimplexMethod` into manual

Skip unneccessary tests or calculations that can become expensive in large examples. (Too large to be suitable for tests.)

Recalculate parent depth entries if they do not yet exist. (Cannot reconstruct example that lead to change.)

### DIFF
--- a/doc/ref/matrix.xml
+++ b/doc/ref/matrix.xml
@@ -429,6 +429,7 @@ See also chapter <Ref Chap="Integral matrices and lattices"/>
 <Heading>Matrices as Linear Mappings</Heading>
 
 <#Include Label="CharacteristicPolynomial">
+<#Include Label="RationalCanonicalFormTransform">
 <#Include Label="JordanDecomposition">
 <#Include Label="BlownUpMat">
 <#Include Label="BlownUpVector">
@@ -613,6 +614,15 @@ grease and level <A>b</A> blocking.
 <#Include Label="AsBlockMatrix">
 <#Include Label="BlockMatrix">
 <#Include Label="MatrixByBlockMatrix">
+
+</Section>
+
+
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<Section Label="Linear Programming">
+<Heading>Linear Programming</Heading>
+
+<#Include Label="SimplexMethod">
 
 </Section>
 </Chapter>
--- a/lib/fitfree.gi
+++ b/lib/fitfree.gi
@@ -108,7 +108,11 @@ local ffs,hom,U,rest,ker,r,p,l,i,depths;
   # transfer the IndicesEANormalSteps
   if ffs.pcgs<>ipcgs and HasIndicesEANormalSteps(ffs.pcgs) then
     U:=IndicesEANormalSteps(ffs.pcgs);
-    r:=ipcgs!.depthsInParent;
+    if IsBound(ipcgs!.depthsInParent) then
+      r:=ipcgs!.depthsInParent;
+    else
+      r:=List(ipcgs,x->DepthOfPcElement(ffs.pcgs,x));
+    fi;
     l:=[];
     for i in U{[1..Length(U)-1]} do # last one is length+1
       p:=PositionProperty(r,x->x>=i);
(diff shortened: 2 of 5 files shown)


lib/fitfree.gi
```diff
@@ -108,7 +108,11 @@ local ffs,hom,U,rest,ker,r,p,l,i,depths;
   # transfer the IndicesEANormalSteps
   if ffs.pcgs<>ipcgs and HasIndicesEANormalSteps(ffs.pcgs) then
     U:=IndicesEANormalSteps(ffs.pcgs);
-    r:=ipcgs!.depthsInParent;
+    if IsBound(ipcgs!.depthsInParent) then
+      r:=ipcgs!.depthsInParent;
+    else
+      r:=List(ipcgs,x->DepthOfPcElement(ffs.pcgs,x));
+    fi;
```
fingolfin: Hmm, so didn't the old code sometimes run into errors? Are there "cheap" examples that show this? In any case, this sounds like a bug fix to me, so perhaps relevant for the release notes after all? Unless it is in code that is undocumented and not normally triggered by user inputs, perhaps... ? I can't judge that right now.

lib/matrix.gd
```diff
@@ -2019,6 +2019,78 @@ DeclareOperation("DefaultScalarDomainOfMatrixList",[IsListOrCollection]);
...
+##  gap> A:=[[3,1,1,4],[1,-3,2,3],[2,1,3,-1]];;
+##  gap> b:=[12,7,10];;c:=[2,4,3,1];;
+##  gap> SimplexMethod(A,b,c);
+##  [ [ 0, 52/5, 0, 2/5 ], 42 ]
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction( "SimplexMethod" );
```
fingolfin: Is there a deeper reason that this manual entries were moved? It makes it very difficult to see if you also made changes to the text, so if possible, it would be nice if they could be moved back to their original position a few lines down.

lib/matrix.gd
```diff
@@ -2019,6 +2019,78 @@ DeclareOperation("DefaultScalarDomainOfMatrixList",[IsListOrCollection]);
...
+##
+#O  RationalCanonicalFormTransform( <A> )
+##
+##  <#GAPDoc Label="Frobenius Normal Form">
+##  <ManSection>
+##  <Func Name="RationalCanonicalFormTransform" Arg='A'/>
+##
+##  <Description>
+
+##  </Description>
```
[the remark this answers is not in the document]
hulpke: Ah -- I was fighting with gapdoc not finding the section and ended up pasting the code into a copied function header that worked. I've tried to clean this (and other relics of the editing) up.

lib/fitfree.gi
hulpke: The code relied on a side-effect, a `depthsInParent` entry, being bound which is not always the case. The example I had was with a matrix group over Z/composite number and the matgrp package.

lib/norad.gi
```diff
@@ -552,7 +552,7 @@ local sus,ser,len,factorhom,uf,n,d,up,mran,nran,mpcgs,pcgs,pcisom,nf,ng,np,sub,
 	# work in quotient modules first
 	module:=GModuleByMats(Concatenation(ngm,npm),f);
 	sumos:=Reversed(MTX.BasesCompositionSeries(module));
-  Info(InfoFitFree,2,"module layers ",List(sumos,Length));
+        Info(InfoFitFree,2,"module layers ",List(sumos,Length));
```
fingolfin:
```suggestion
  Info(InfoFitFree,2,"module layers ",List(sumos,Length));
```

doc/ref/matrix.xml
```diff
@@ -614,6 +615,15 @@ grease and level <A>b</A> blocking.
...
 <#Include Label="MatrixByBlockMatrix">
 
+</Section>
+
+
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<Section Label="Linear Programming">
+<Heading>Linear Programming</Heading>
+
+<#Include Label="SimplexMethod">
```
fingolfin: This PR changes two functions from undocumented to documented; that definitely warrants a release notes entry

PR #2993, opened 2018-11-12
Add quick nonsolvability test before trying permutation Pcgs

hulpke:
In larger degree permutation groups, `IsSolvableGroup` (or the test of calculating a Pcgs for stabilizer chain) can become expensive. Added a quick probabilistic check of iterated commutators beyond maximal derived length that will recognize such nonsolvable cases quickly.
(This is already testes extensively by existing examples.)

### DIFF
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1248,7 +1248,7 @@ local som,elvth,fct,gens,new,l,i,j,a,b,bound;
 
   elvth:=One(G);
   fct:=function(G) 
-    elvth:=elvth*Product(List([1..Random([1..5])],x->Random(GeneratorsOfGroup(G))^Random([-3..3])));
+    elvth:=elvth*Product([1..Random(1,5)],x->Random(GeneratorsOfGroup(G))^Random(-3,3));
     return elvth;
   end;
 
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1243,7 +1243,7 @@ local som,elvth,fct,gens,new,l,i,j,a,b,bound;
   som:=MovedPoints(G);
   bound:=Int(LogInt(Length(som)^5,3)/2); #Dixon Bound
   if Length(som)>100 then
-    som:=som{List([1..100],x->Random([1..Length(som)]))};
+    som:=som{List([1..100],x->Random(1, Length(som)))};
   fi;
 
   elvth:=One(G);
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1232,9 +1232,72 @@ end );
 ##
 #M  IsSolvableGroup( <G> )  . . . . . . . . . . . . . . . .  solvability test
 ##
+
+# fast test for finding iteratively nontrivial commutators in a permutation group.
+# if this goes on too long the group may not be solvable. This can be much faster
+# for large degree groups than to use the full pcgs machinery (which uses the same
+# idea but builds a pcgs on the way).
+BindGlobal("QuickUnsolvabilityTestPerm",function(G)
+local som,elvth,fct,gens,new,l,i,j,a,b,bound;
+  # a few moved points
+  som:=MovedPoints(G);
+  bound:=Int(LogInt(Length(som)^5,3)/2);
+  if Length(som)>100 then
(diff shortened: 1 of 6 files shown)


lib/grpperm.gi
```diff
@@ -1232,9 +1232,72 @@ end );
...
+# if this goes on too long the group may not be solvable. This can be much faster
+# for large degree groups than to use the full pcgs machinery (which uses the same
+# idea but builds a pcgs on the way).
+BindGlobal("QuickUnsolvabilityTestPerm",function(G)
+local som,elvth,fct,gens,new,l,i,j,a,b,bound;
+  # a few moved points
+  som:=MovedPoints(G);
+  bound:=Int(LogInt(Length(som)^5,3)/2);
+  if Length(som)>100 then
+    som:=som{List([1..100],x->Random([1..Length(som)]))};
```
fingolfin:
```suggestion
    som:=som{List([1..100],x->Random(1, Length(som)))};
```

lib/grpperm.gi
```diff
@@ -1232,9 +1232,72 @@ end );
...
+  fi;
+
+  elvth:=One(G);
+  fct:=function(G) 
+    elvth:=elvth*Product(List([1..Random([1..5])],x->Random(GeneratorsOfGroup(G))^Random([-3..3])));
```
fingolfin:
```suggestion
    elvth:=elvth*Product([1..Random(1,5)],x->Random(GeneratorsOfGroup(G))^Random(-3,3));
```

lib/grpperm.gi
```diff
@@ -1232,9 +1232,72 @@ end );
...
+
+# fast test for finding iteratively nontrivial commutators in a permutation group.
+# if this goes on too long the group may not be solvable. This can be much faster
+# for large degree groups than to use the full pcgs machinery (which uses the same
+# idea but builds a pcgs on the way).
+BindGlobal("QuickUnsolvabilityTestPerm",function(G)
+local som,elvth,fct,gens,new,l,i,j,a,b,bound;
+  # a few moved points
+  som:=MovedPoints(G);
+  bound:=Int(LogInt(Length(som)^5,3)/2);
```
fingolfin: Is there a deeper mathematical reason for this bound, or is it just a heuristic?

lib/grpperm.gi
```diff
@@ -1232,9 +1232,72 @@ end );
...
+  if Length(som)>100 then
+    som:=som{List([1..100],x->Random([1..Length(som)]))};
+  fi;
+
+  elvth:=One(G);
+  fct:=function(G)
```
fingolfin: `fct` seems to be a way to quickly generate pseudo random elements of `G`? If so, why can't `PseudoRandom(G)` be used?

lib/grpperm.gi
hulpke: This is (written slightly awkwardly, as we don't have a root) bound for solvable length of a permutation group due to Dixon. I add a comment.

lib/grpperm.gi
hulpke: In some examples I tried PseudoRandom took too long, as it aims to mix better. Thus, using PseudoRandom, the supposedly quick first test took longer than the proper test using the Sims algorithm.
This (pathetic, but quick) set of products was good enough to show in the experiment cases that the group is not solvable, while in some of these cases the Sims algorithm took very long, while building a partial stabilizer chain.

fingolfin: OK, makes sense, thank you for the explanation.

lib/grpperm.gi
```diff
@@ -1232,9 +1232,72 @@ end );
...
+
+# fast test for finding iteratively nontrivial commutators in a permutation group.
+# if this goes on too long the group may not be solvable. This can be much faster
+# for large degree groups than to use the full pcgs machinery (which uses the same
+# idea but builds a pcgs on the way).
+BindGlobal("QuickUnsolvabilityTestPerm",function(G)
+local som,elvth,fct,gens,new,l,i,j,a,b,bound;
+  # a few moved points
+  som:=MovedPoints(G);
+  bound:=Int(LogInt(Length(som)^5,3)/2); #Dixon Bound
```
fingolfin: Thanks for the added comment, that at least gives a hint where to start looking.

PR #2994, opened 2018-11-12
kernel: some cleanup for the various GC implementations

fingolfin:
This adds a new kernel file `src/bags.c` and moves some common code there.

My long-term vision is to also rename `src/gasman.h` to `src/bags.h`, and remove all remaining GASMAN specifics from it to `gasman_intern.h`, then rename the latter to `gasman.h`. Does that sound OK?

Also, more (ideally: all) HPC-GAP specific code should be removed from `boehm_gc.c`; my hope is that on the long run, we can also use the Julia GC with HPC-GAP, and this is easier if the HPC-GAP specifics are cleanly decoupled from the Boehm GC specific code. Same also applies to `weakptr.c`, of course.


fingolfin:
Yes, this was also tested with Julia; @sebasguts and me have been extensively working on https://github.com/oscar-system/GAPJulia and in fact the changes in here were indirectly motivated by the addition of the Julia GC (I've been mulling about adding `src/bags.{c,h}` for months now, in order to factor out common code -- now that we have three GCs, I've got a feeling that I can tell a bit better what is really "common" code resp. API and what is not.

PR #2997, opened 2018-11-13
WIP Larger Finite Fields, take 2

stevelinton:
This is a rebased and somewhat reworked version of  #2846. The biggest addition is that the lookup tables are computed at build time using `etc/ffgen.c` which has been somewhat updated. The decision about what size of internal finite fields to use on what architectures is made entirely in `etc/ffgen.c`, in principle it should be possible to go to 23 bits or 25 bits just by editing that programme.

I've done quite a bit of refactoring of the finite fields code in the process, mostly introducing inline functions and removing some duplication.

Still to do is updates to tests, documentation and comments.

I'd welcome @fingolfin looking over this, especially the changes to the makefiles, and @jdebeule to confirm I haven't done anything too stupid to his code.


stevelinton:
Working to clean up the test failures. Mostly straightforward, but there is a problem with tests in `hpcgap/tst/testinstall/comprvec.tst` like
```
gap> F:=GF(41^3);;
gap> v:=Filtered(F,x -> x in GF(41));;
gap> IS_VECFFE(v);
false
gap> IsFFECollection(v);
true
gap> z:=CopyToVectorRep(v,41); 
< mutable compressed vector length 41 over GF(41) >
gap> z=v;                      
true
gap> IsIdenticalObj(z,v);   
false
```
This fails now because `Z(41,3)` is internally represented. To avoid that I'd have to increase 3 to 5 and then the `Filtered` line takes a *long* time. I don't understand the purpose of these tests well enough to know what is a sensible change. Can anyone advise?

Also, there is a problem building normaliz-interface. It uses an m4 script `ac_find_gap.m4` which doesn't take `GAP_CPP_FLAGS` from `sysinfo.gap` but instead rolls its own and doesn't include the `gen` directory, resulting n `ffdata.h` not being found (which is included indirectly by `compiled.h`). Not sure what the best way to fix this is.

stevelinton:
If this is merged, we might want to add a virtual enumerator for Galois Fields instead of making a plain list.

lib/ffe.g
```diff
@@ -14,10 +14,11 @@
 
 #############################################################################
 ##
-
 #V  MAXSIZE_GF_INTERNAL . . . . . . . . . . . . maximal size of internal ffes
 ##
-BIND_GLOBAL( "MAXSIZE_GF_INTERNAL", 2^16 );
+##  Set by the kernel, we just need to tidy up
```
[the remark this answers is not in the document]
stevelinton: Agreed. Will fix

etc/ffgen.c
```diff
@@ -37,96 +44,139 @@ void make_ff()
...
     else {
-        printf("#include \"ffdata.h\"\n");
-        printf("\n");
-        printf("/* Entries are ordered by value of p^d; can use binary search\n");
-        printf(" * to find them. Indices start at 1.\n");
-        printf(" */\n");
-        printf("\n");
-        printf("unsigned char DegrFF[NUM_SHORT_FINITE_FIELDS+1] = {\n");
-        printf(" %3d,", 0);
+        fprintf( dest,"#include \"system.h\"\n");
```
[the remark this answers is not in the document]
stevelinton: Is there an easy way to run it on the whole PR without altering unchanged parts of files. I most admit I thought I had done this file, but I will fix.

src/finfield.c
```diff
@@ -333,6 +253,20 @@ FF              FiniteField (
...
 
+FF FiniteField (
+    UInt                p,
+    UInt                d )
+{
+    UInt q,i;
+    /* calculate size of field */
+    q = 1;
+    for ( i = 1; i <= d; i++ )
+        q *= p;
```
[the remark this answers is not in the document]
stevelinton: Will fix and try and add a test.

src/finfield.c
```diff
@@ -333,6 +253,20 @@ FF              FiniteField (
     return ff;
 }
 
+FF FiniteField (
+    UInt                p,
+    UInt                d )
```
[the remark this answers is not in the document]
stevelinton: In the kernel, just in `vec8bit.c`, but it's an external C function that could quite sensibly
be used in package code.

src/finfield.c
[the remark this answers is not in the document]
stevelinton: There was already a check in `FuncZ2`, so `Z(4,2)` fails

hulpke:
@stevelinton 

> Working to clean up the test failures. Mostly straightforward, but there is a problem with tests in `hpcgap/tst/testinstall/comprvec.tst` like

This looks as if it tries taking making larger GF-elements that lie in the small prime field (presumably under the assumption that they are stored differently internally), and checks that compression into the small-field vector format works.

To make the exponent changeable, replace the `Filtered` with

    gap> p:=PrimitiveElement(F)^((41^3-1)/(41-1));;
    gap> Concatenation([0*p],List([0..39],x->p^x));

which writes down the same list without enumerating the large field.

lib/ffe.gd
```diff
@@ -30,15 +31,19 @@
 ##  <P/>
 ##  &GAP; can represent elements of all finite fields
 ##  <C>GF(<A>p^d</A>)</C> such that either
-##  (1) <A>p^d</A> <M>&lt;= 65536</M> (in which case an extremely efficient
-##      internal representation is used);
+##  (1) <A>p^d</A> <M>&lt;=</M> <C>MAXSIZE_GF_INTERNAL</C> (in which case an 
+##      extremely efficient internal representation is used);
```
fingolfin: Super nitpick (and I know this is old text you just reformatted, so feel free to ignore): Is the word "extremely" really appropriate? Maybe "highly"?

src/finfield.h
```diff
@@ -13,8 +13,10 @@
 **
 **  Finite fields  are an  important   domain in computational   group theory
 **  because the classical matrix groups are defined over those finite fields.
-**  In GAP we  support  small  finite  fields  with  up  to  65536  elements,
-**  larger fields can be realized  as polynomial domains over smaller fields.
+**  The GAP kernel supports elements of finite fields up to some fixed size 
+**  limit, typically 2^16 on 32 bit systems and 2^24 on 64 bit systems.
+**  To change the limits, edit etc/ffgen.c. To access the current limits use 
+**  MAXSIZE_GF_INTERNAL. Larger fields are supported in the library
```
fingolfin:
```suggestion
**  MAXSIZE_GF_INTERNAL. Support for larger fields is implemented by the library.
```

etc/ffgen.c
```diff
@@ -37,97 +46,149 @@ void make_ff()
...
-        printf("    NUM_SHORT_FINITE_FIELDS = %d\n", num_ff);
-        printf("};\n");
-        printf("\n");
-        printf("extern unsigned long SizeFF[NUM_SHORT_FINITE_FIELDS+1];\n");
-        printf("extern unsigned char DegrFF[NUM_SHORT_FINITE_FIELDS+1];\n");
-        printf("extern unsigned long CharFF[NUM_SHORT_FINITE_FIELDS+1];\n");
-        printf("\n");
-        printf("#endif // GAP_FFDATA_H\n");
+        fprintf(dest, "/* This file is generated by etc/ffgen.c during build"
+                      " do not edit */\n");
```
fingolfin: So the changes in this file are difficult to review, because of those `printf(` -> `fprintf(dest, ` changes. I wonder: are those really necessary? What's the problem with the old approach, where `ffgen` simply was called twice, and its output was piped into the relevant files? Is it perhaps too slow now?

lib/ffe.gd
```diff
@@ -30,15 +31,19 @@
...
 ##      using the machinery of residue class rings
 ##      (see section&nbsp;<Ref Sect="Residue Class Rings"/>) or
 ##  (3) if the Conway polynomial of degree <A>d</A> over the field with
 ##      <A>p</A> elements is known, or can be computed
 ##      (see <Ref Oper="ConwayPolynomial"/>).
 ##  <P/>
+##
+##  <C>MAXSIZE_GF_INTERNAL</C> may depend on the word size of your computer
+##  and the version of &GAP; but will typically be either <M>2^{16}</M> or
+##  <M>2^{24}</M>.<P/>
```
fingolfin: Maybe insert another "empty" line (i.e., `##`) after this, to make the paragraph break more visible if one reads the comment directly, not via GAPDoc?

src/finfield.h
```diff
@@ -64,10 +68,10 @@
 **
 **  Small finite fields are represented by an  index  into  a  global  table.
 **
-**  Since there are  only  6542 (prime) + 93 (nonprime)  small finite fields,
-**  the index fits into a 'UInt2' (actually into 13 bits).
+**  Depending on the configuration it may be UInt2 or UInt4. The definition
+**  is in `ffdata.h` and is calculated by etc/ffgen.c
 */
-typedef UInt2       FF;
```
fingolfin: Is there are any reason why `FF` couldn't / shouldn't simply always be `UInt4`?

src/finfield.h
```diff
@@ -128,8 +132,8 @@ extern  Obj             SuccFF;
 **  'TYPE_FF' returns the type of elements of the small finite field <ff>.
 **  'TYPE_FF0' returns the type of the zero of <ff>
 **
-**  Note that  'TYPE_FF' is a macro, so  do not call  it  with arguments that
-**  have side effects.
+**  Note that  'TYPE_FF' and TypeFF0 are macros, so  do not call them
```
fingolfin: I assume you mean `TYPE_FF0` here?

src/finfield.h
```diff
@@ -128,8 +132,8 @@ extern  Obj             SuccFF;
 **  'TYPE_FF' returns the type of elements of the small finite field <ff>.
 **  'TYPE_FF0' returns the type of the zero of <ff>
 **
-**  Note that  'TYPE_FF' is a macro, so  do not call  it  with arguments that
-**  have side effects.
+**  Note that  'TYPE_FF' and TypeFF0 are macros, so  do not call them  
+**  with arguments that have side effects.
```
fingolfin: Optional remark: They are macros which are used in precisely one place, inside function `TypeFFE` in `finfield.c`.  Moreover, `TypeFF` and `TypeFF0` are only used in that `.c` file, too Perhaps it would thus be best to remove them from the header, and turn the variables into static ones.

src/finfield.h
```diff
@@ -148,17 +152,38 @@ extern  Obj             TypeFF0;
...
+**  the  same finite field.  If you  want to multiply  two elements where one
+**  lies in a subfield of the other use 'ProdFFEFFE'.
+**
+**  If one of the values is 0 the product is 0.
+**  If $a+b <= o$ we have $a * b ~ z^{a-1} * z^{b-1} = z^{(a+b-1)-1} ~ a+b-1$
+**  otherwise   we   have $a * b ~ z^{(a+b-2)-(o-1)} = z^{(a+b-o)-1} ~ a+b-o$
 */
-typedef UInt2           FFV;
+
+static inline FFV PROD_FFV(FFV a, FFV b, const FFV * f)
```
fingolfin: Is there a reason this (and its comment) were moved from before QUO_FFV to before SUM_FFV? This makes the diff bigger and harder to see what changed.

src/finfield_conway.h
```diff
@@ -0,0 +1,6957 @@
+/**************************************************
+ * Table of Conway polynomials for fields up order
+ * 2^32. The polynomial for q = p^r 
+ * will be of the form x^r + \sum{a_i x^i} is represented
+ * by two entries in PolsFF: q and \sum{Int(a_i) q^i}
```
fingolfin: I assume this was generated by a script? Can we include that somewhere?

stevelinton: I don't have it, although it would be easy to write such a script in GAP.

stevelinton:
The remaining test failures arise from interactions between `factint` and HPC-GAP and from the problems building `normaliz-interface` already mentioned.

stevelinton:
Now just down to the `normaliz-interface` issues. Thinking about that, though, I wondered about where `ffdata.c` and `ffdata.h` should really go. At the moment they are in `gen` but they are architecture dependent (just on 32 vs 64 bits, not anything else) so should they actually be in `bin/$(GAPARCH)`? 

The only difference I can see it making is that they might not need to be rebuilt if you switch back and forth between architectures and rebuild GAP each time, but I'm not sure of that.

fingolfin:
Putting those files into `gen` is exactly the right place. There is no need for arch dependenat dirs, because now every build dir corresponds to precisely one build config; if one wants to build another binary with another config "in parallel" to the current one, one needs to use an out-of-tree build; or a separate clone.

However, the `ffgen` make target is not setup sufficiently for that, as it does not correctly track the configuration, and thus may not get rebuilt when it should be. A fix for that might be as simple as changing `ffgen: etc/ffgen.c` to `ffgen: obj/etc/ffgen.lo`, but I have not yet tried to do that; a slight variation might be necessary.

I have "looking at `NormalizInterface` issues with new FF code" on my TODO list, along with some general lovin' for `NormalizInterface` and possibly a release (CC: @sebasguts ) but it might take me a bit to get there, as I have a bunch of other GAP related tasks on it for today (such as fixing more of the fallout of the method reordering changes...)

stevelinton:
@fingolfin Thanks a lot. Take your time. Although, before this PR gets too old, we need to decide whether to include it in 4.11, forcing the change on `NormalizInterface` or add some workaround, or just keep it on ice (with attendant risks of bitrot).

fingolfin:
I'd like to avoid having it sit around for too long. I hope @sebasguts and me can make a NormalizInterface release soon. Once that has happened and was picked up, we can hopefully proceed with this PR. It would be nice to get some feedback from @jdebeule (whether he's OK with what was done to his PR ;-) first, too.

src/finfield_conway.h
jdebeule: I used indeed a simple GAP script. This is something I played with about two years ago, now it seems to give errors under gap4r10. But here is the code I used:
ConwayPolNumber := function(p,h)
local list,number;
list := ConwayPol(p,h);
number := Sum(List([1..Length(list)-1],i->list[i]*p^(i-1)));
end;

ConwayPolForCCodeFlat := function(p,h)
local list, number;
list := ConwayPol(p,h);
number := Sum(List([1..Length(list)-1],i->list[i]*p^(i-1)));
return Concatenation(String(p^h),", ",String(number),",\n");
end;


ConwayPolForCCodeNice := function(p,h)
local list,substring,co,sub,i,lc,lpp,empty,idx,a,stop;
empty := Concatenation(List([1..40],x->" "));
list := ConwayPol(p,h);
substring := [String(p^h)];
Add(substring,", ");
for i in [1..Length(list)-1] do
    co := list[i];
    if co = 0 then
        lpp := Length(String(p^(i-1)));
        sub := empty{[1..2+lpp]};
    else
        if i = 1 then
            sub := Concatenation(String(co)," ");
        else
            sub := Concatenation("+",String(co),"*",String(p^(i-1)));
        fi;
    fi;
    Add(substring,sub);
od;
idx := Length(substring);
stop := false;
while not stop do
    a := substring[idx];
    NormalizeWhitespace(a);
    if a = "" then
        Unbind(substring[idx]);
        idx := idx - 1;
    else
        stop := true;
    fi;
od;
Add(substring,",\n");
return Concatenation(substring);
end;

MAX := 2^32;

polynomials_as_strings_of_numbers := [];
p := 2;
h := 2;
while p^h <= MAX do
    while p^h <= MAX do
        Add(polynomials_as_strings_of_numbers,ConwayPolForCCodeNice(p,h));
        h := h+1;
    od;
    h := 2;
    p := NextPrimeInt(p);
od;

polynomials_as_strings_of_flat_numbers := [];
p := 2;
h := 2;
while p^h <= MAX do
    while p^h <= MAX do
        Add(polynomials_as_strings_of_flat_numbers,ConwayPolForCCodeFlat(p,h));
        h := h+1;
    od;
    h := 2;
    p := NextPrimeInt(p);
od;


SizeScreen([256,40]);
LogTo("~/output");
for p in polynomials_as_strings_of_numbers do
Print(p);
od;

SizeScreen([256,40]);
LogTo("~/output2");
for p in polynomials_as_strings_of_flat_numbers do
Print(p);
od;

jdebeule:
Since I had no time to work further on the original PR, I am very grateful that this was picked up! So I am ok with what has been done to the original PR.

fingolfin:
I have now rebased this atop of PR #3137 (which should be merged first, and then we can rebase this PR once again). I also added the code by @jdebeule for regenerating the table of conway polynomials, though it is not yet fully hooked up (see the commit message of the WIP commit adding it for more details). Also, some of the changes in this PR were squashed by me into a single "WIP" commit -- I plan to break this up further later on, but for now, I wanted to make sure my changes so far are backed up and tested on GitHub ;-).

src/finfield_conway.h
fingolfin: I added a tweaked version of this as `etc/gen_conway.g`. TODO: add a comment to src/finfield_conway.h which mentions etc/gen_conway.g

fingolfin:
This has now been reduced to a relatively small set of changes (except for the about 7000 new generated lines in `src/finfield_conway.h`, and the removal of `src/ffdata.c` with another 2000 lines). I still need to clean up the remaining commits. Perhaps also break out the addition of `etc/gen_conway.g` into its own PR, and hook it into the build system so that regenerating that table becomes trivial.

One thing that concerns me a bit about this PR is that the generated file `gen/ffdata.c` in 64 builds is 26 MB. And the GAP executable grows from 2.3 MB (1.8 MB stripped) to 12 MB (11 MB stripped), i.e., growing by a factor of 6. Now of course this is still no problem on current machines, but I nevertheless find it a bit troubling.

stevelinton:
@fingolfin Given that, as you say, it is not actually a problem on modern machines, I'm not inclined to worry about this. The root problem is that we need to number all the prime powers up to whatever our limit is, in a way that does not introduce race conditions, and allows reasonably quick lookup. It's not essential that the numbering be entirely dense, but we don't want to use more bits than we have to. Calculating it at startup (before we have multiple threads) is a bit slower than we'd like. Doing it on demand (so that the tables would be populated as far as the largest q so far considered) would be possible, but adds complexity. Loading the data  at run-time (for instance via mmap) would be possible, but really is just hiding the problem. 

One thing we might want to do is export this data to GAP level in some way, so that it is available for factoring, primality testing, etc. 

In the somewhat unlikely event that someone did want to build a 64 bit version of GAP with minimal memory footprint then (a) their main job will be to minimize how much of the library they load and (b) they can easily configure it to use fewer than 24 bits worth of finite fields.

PR #2999, opened 2018-11-14
Added functions GetChar and CharValue to libgap api

sebasguts:


src/libgap-api.c
```diff
@@ -258,3 +258,16 @@ char * GAP_CSTR_STRING(Obj string)
...
     return CSTR_STRING(string);
 }
+
+UChar GAP_CharValue(Obj obj)
+{
+    if (TNUM_OBJ(obj) != T_CHAR) {
+        ErrorQuit("GAP_CharValue: argument is not a char", 0, 0);
+    }
+    return CHAR_VALUE(obj);
+}
```
ChrisJefferson: I don't know if some people might not want ErrorQuit calling. In that case you could return an Int, which is an unsigned char, or -1 if the value is not a char.

src/libgap-api.c
```diff
@@ -258,3 +258,16 @@ char * GAP_CSTR_STRING(Obj string)
...
+
+Obj GAP_GetChar(UChar obj)
```
ChrisJefferson: Maybe GetCharObj, just so this isn't confused with the C function getchar (which reads a char from the terminal).

src/libgap-api.h
```diff
@@ -147,5 +147,11 @@ extern Obj GAP_MakeString(const char * string);
 // terminated C string.
 extern Obj GAP_MakeImmString(const char * string);
 
+// Returns the value of the GAP char <obj>.
```
```suggestion
// Returns the value of the GAP character object <obj>.
```

src/libgap-api.h
```diff
@@ -147,5 +147,11 @@ extern Obj GAP_MakeString(const char * string);
...
+// If <obj> is not a GAP char, an error is raised.
```
```suggestion
// If <obj> is not a GAP character object, an error is raised.
```

src/libgap-api.h
```diff
@@ -147,5 +147,11 @@ extern Obj GAP_MakeString(const char * string);
...
+extern UChar GAP_CharValue(Obj obj);
+
+// Returns the char Obj corresponding to <obj>
```
```suggestion
// Returns the GAP character object with value <obj>.
```

src/libgap-api.c
fingolfin: Maybe `GAP_ValueOfChar` and `GAP_CharWithValue` ?

fingolfin:
@ChrisJefferson are you OK with this now?

ChrisJefferson:
yup, happy

PR #3000, opened 2018-11-14
kernel: change argument validation for OBJ_{SET,MAP} to use RequireArgumentCondition

fingolfin:


src/objset.c
```diff
@@ -844,9 +835,13 @@ static Obj FuncCLEAR_OBJ_SET(Obj self, Obj set) {
...
  */
 
-static Obj FuncOBJ_SET_VALUES(Obj self, Obj set) {
-  CheckArgument("OBJ_SET_VALUES", set, T_OBJSET, -1);
-  return ObjSetValues(set);
+static Obj FuncOBJ_SET_VALUES(Obj self, Obj set)
+{
+    RequireArgumentCondition("OBJ_SET_VALUES", set, "objset",
+                             TNUM_OBJ(set) == T_OBJSET,
+                             "must be a mutable object set");
```
fingolfin: @rbehrends is there any reason this function requires mutable objset?

src/objset.c
```diff
@@ -967,9 +982,13 @@ static Obj FuncCLEAR_OBJ_MAP(Obj self, Obj map) {
...
  */
 
-static Obj FuncOBJ_MAP_VALUES(Obj self, Obj map) {
-  CheckArgument("OBJ_MAP_VALUES", map, T_OBJMAP, -1);
-  return ObjMapValues(map);
+static Obj FuncOBJ_MAP_VALUES(Obj self, Obj map)
+{
+    RequireArgumentCondition("OBJ_MAP_VALUES", map, "objmap",
+                             TNUM_OBJ(map) == T_OBJMAP,
+                             "must be a mutable object map");
```
fingolfin: @rbehrends is there any reason this function requires mutable objmap?

src/objset.c
```diff
@@ -980,9 +999,13 @@ static Obj FuncOBJ_MAP_VALUES(Obj self, Obj map) {
...
  */
 
-static Obj FuncOBJ_MAP_KEYS(Obj self, Obj map) {
-  CheckArgument("OBJ_MAP_KEYS", map, T_OBJMAP, -1);
-  return ObjMapKeys(map);
+static Obj FuncOBJ_MAP_KEYS(Obj self, Obj map)
+{
+    RequireArgumentCondition("OBJ_MAP_KEYS", map, "objmap",
+                             TNUM_OBJ(map) == T_OBJMAP,
+                             "must be a mutable object map");
```
fingolfin: @rbehrends is there any reason this function requires mutable objmap?

ChrisJefferson:
I'm happy with this, we can always change the requirements on mutability later, if/when we find out more.

src/objset.c
rbehrends: No, this and the other checks are probably the result of copy-and-paste errors.

fingolfin:
Since @rbehrends clarified now, I'll update this PR accordingly.

PR #3005, opened 2018-11-15
buildsys: disable 'make install', as people keep ignoring the warning

fingolfin:
This should be backported to GAP 4.10.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -471,21 +471,22 @@ clean:
 LTINSTALL=$(LIBTOOL) --mode=install $(INSTALL)
 
 install: install-libgap install-headers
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Backported to stable-4.10 in 62cf084d7a8719da7ed654c7b6c7aca1a14bb560

jdemeyer:
What's the rationale for disabling `install-headers`? We were using that in Sage for GAP-4.10.0 without problems. What is the new recommended way to install GAP headers?

jdemeyer:
Same for `install-libgap`.

jdemeyer:
I just saw that https://github.com/gap-system/gap/pull/3294 reverts this, so I guess it makes sense for Sage to use that patch.

PR #3007, opened 2018-11-15
Added GAP_ValueMacFloat to libgap-api

sebasguts:
I decided to use `double` instead of `Double` to not having to include `macfloats.h` into `libgap-api.h`

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -10,6 +10,7 @@
 #include "gvars.h"
 #include "integer.h"
 #include "lists.h"
+#include "macfloat.h"
 #include "opers.h"
 #include "plist.h"
 #include "streams.h"
@@ -191,6 +192,29 @@ Obj GAP_CallFuncArray(Obj func, UInt narg, Obj args[])
     }
 
     return result;
+}
+
+
+////
+//// floats
+////
+
+Int GAP_IsMacFloat(Obj obj)
+{
+    return IS_MACFLOAT(obj);
+}
+
+double GAP_ValueMacFloat(Obj obj)
+{
+    if (!IS_MACFLOAT(obj)) {
+        ErrorMayQuit("<obj> is not a MacFloat", 0, 0);
+    }
+    return (double)VAL_MACFLOAT(obj);
+}
+
+Obj GAP_NewMacFloat(double x)
+{
+    return NEW_MACFLOAT(x);
 }
 
 
--- a/src/libgap-api.h
+++ b/src/libgap-api.h
@@ -78,8 +78,23 @@ extern Obj GAP_Fail;
 extern Obj GAP_CallFuncList(Obj func, Obj args);
 
 // Call the GAP object <func> as a function with arguments given
-// as an array <args> with <narg> entries
+// as an array <args> with <narg> entries.
 extern Obj GAP_CallFuncArray(Obj func, UInt narg, Obj args[]);
+
+
+////
+//// floats
+////
+
+// Returns 1 if <obj> is a GAP machine float, 0 if not.
+extern Int GAP_IsMacFloat(Obj obj);
+
+// Returns the value of the GAP machine float object <obj>.
+// If <obj> is not a machine float object, an error is raised.
+extern double GAP_ValueMacFloat(Obj obj);
+
+// Returns a new GAP machine float with value <x>.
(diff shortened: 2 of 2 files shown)


src/libgap-api.c
```diff
@@ -313,3 +314,11 @@ Obj GAP_CharWithValue(UChar obj)
 {
     return ObjsChar[obj];
 }
+
+double GAP_ValueMacFloat(Obj obj)
+{
+    if (!IS_MACFLOAT(obj)) {
+        ErrorQuit("GAP_ValueMacFloat: <obj> is not a MacFloat",0,0);
```
fingolfin: Please clang-format this

src/libgap-api.h
```diff
@@ -197,4 +197,8 @@ extern Int GAP_ValueOfChar(Obj obj);
 // Returns the GAP character object with value <obj>.
 extern Obj GAP_CharWithValue(UChar obj);
 
+// Returns the value of the GAP machine float object <obj>.
+// If <obj> is not a machine float object, an Error is raised.
```
```suggestion
// If <obj> is not a machine float object, an error is raised.
```

fingolfin: In order to be able to safely use this, we also need a GAP_IsMacFloat function

src/libgap-api.c
```diff
@@ -313,3 +314,11 @@ Obj GAP_CharWithValue(UChar obj)
 {
     return ObjsChar[obj];
 }
+
+double GAP_ValueMacFloat(Obj obj)
+{
+    if (!IS_MACFLOAT(obj)) {
```
ChrisJefferson: On the assumption that interface users might not like ErrorQuits, how about an interface:

`Int GAP_ValueMacFloat(Obj obj, double* d);`

Write value of obj into d, return 0 on failure, 1 on success, inspired by scanf and friends.

sebasguts:
I hopefully adressed all the comments, and added `GAP_NewMacFloat` and `GAP_IsMacFloat`.

src/libgap-api.h
```diff
@@ -197,4 +197,15 @@ extern Int GAP_ValueOfChar(Obj obj);
 // Returns the GAP character object with value <obj>.
 extern Obj GAP_CharWithValue(UChar obj);
 
+// Assigns the value of the GAP machine float object <obj>
```
fingolfin: Nitpick: There should be a section break comment before this, i.e.,
```C
////
//// floats
////
```
and ideally this section then should be sorted between the "calls" and "integers" section.

src/libgap-api.c
fingolfin: I am not so fond of the idea. Now we've replaced one source of mistakes with another: what if you pass in an invalid pointer? OK, we can at least add a check `x != 0`, but that won't catch references to stack values that already went out of range and silliness like that.

All code I know which would call something like (GAPJulia, SingularInterface, NormalizInterface, SageMath interface for GAP) actually looks roughly like the following anyway, rendering the extra check unnecessary (and in fact, inconvenient):
```C
  if (GAP_IsInt(obj)) {
    return convert_gap_int_to_our_format(obj);
  }
  else if (GAP_IsMacFloat(obj)) {
    return convert_double_to_our_format( GAP_ValueMacFloat(obj) );
  }
  ...
```

But if we really think we should use this kind of indirection here, then I think we should be consistent and do it for all the other getters we already defined resp. will define in the future.

sebasguts:
@fingolfin The problem for GAPJulia is even bigger, though.

I tried a lot of things yesterday, and it seems this way you need to create a `Float64` Array, pass its data pointer to the function, and then extract its first value. It seems impossible to pass a pointer to a single `Float64`.

Anyway, what would you suggest instead?

fingolfin:
I would suggest to change back to `double GAP_ValueMacFloat(Obj obj)`, as it was before.

ChrisJefferson:
Sorry to keep changing things.

If it's hard to make a double* in julia, then I wouldn't mind changing back. In general, I would like to make it as hard as possible to break type-safety, but only with interfaces which are actually usable.

sebasguts:
@ChrisJefferson Do not worry here. I am not completely sure if I am right with the array, it just appeared that way.

@fingolfin Do you suggest removing the typechecks alltogether?

sebasguts:
@fingolfin okay, thanks a lot :)

PR #3008, opened 2018-11-15
build: modernize build system to full autoreconf/automake

jengelh:
* add AC_CONFIG_MACRO_DIRS as suggested by autoreconf
* make use of implied libtoolize run to generate a lot of .m4 files,
  hereafter deleted from the repo
* make use of implied automake run to generate e.g. install-sh,
  hereafter deleted from the repo
* move to automake
* move -L of GMP_LDFLAGS into GMP_LIBS (as is a normal thing for automake)

This is yet a draft. I have not tested or been able to test every aspect of this like documentation generation or cygwin builds, but being able to (still) build e.g. curlInterface externally makes this look like it could be a start.


fingolfin:
You really could have saved you all this work if you had talked to us first. Anyway, this change does not seem to solve any actual problems, but introduces a tons of new ones.

jengelh:
>No way. GNU automake is an abomination from hell.

So is autoconf and libtool, has nobody told you that yet?
(Calling tools names is rather childish; it would have been more than sufficient to express it as a simple personal dislike.) The issue AM would solved is the deduplication of libtool and compiler command lines, dependency tracking. The diffstat, even discounting the m4 boilerplate, would speak for itself.
Guess I'm not going to fixup gac then.

PR #3009, opened 2018-11-15
More error cleanups

ChrisJefferson:
Here is a few more error cleanups, mainly moving from `Require*` functions to `Get*` functions where possible, which helps reduce the number of INT_INTOBJ calls

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -821,19 +821,15 @@ Obj PowIntPerm(Obj opL, Obj opR)
 {
     Int                 img;            /* image (result)                  */
 
+    GAP_ASSERT(TNUM_OBJ(opL) == T_INTPOS || TNUM_OBJ(opL) == T_INT);
+
     /* large positive integers (> 2^28-1) are fixed by any permutation     */
     if ( TNUM_OBJ(opL) == T_INTPOS )
         return opL;
 
-    /* permutations do not act on negative integers                        */
-    img = INT_INTOBJ( opL );
-    if ( img <= 0 ) {
-        opL = ErrorReturnObj(
-            "Perm. Operations: <point> must be a positive integer (not %d)",
-            (Int)img, 0L,
-            "you can replace <point> via 'return <point>;'" );
-        return POW( opL, opR );
-    }
+    img = INT_INTOBJ(opL);
+    RequireArgumentCondition("PowIntPerm", opL, "point", img > 0,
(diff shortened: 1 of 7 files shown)


src/permutat.cc
```diff
@@ -826,14 +826,7 @@ Obj PowIntPerm(Obj opL, Obj opR)
...
     /* permutations do not act on negative integers                        */
-    img = INT_INTOBJ( opL );
-    if ( img <= 0 ) {
-        opL = ErrorReturnObj(
-            "Perm. Operations: <point> must be a positive integer (not %d)",
-            (Int)img, 0L,
-            "you can replace <point> via 'return <point>;'" );
-        return POW( opL, opR );
-    }
+    img = GetPositiveSmallInt("PermImage", opL, "point");
```
fingolfin: There is a reason why I did not just replace the above with `RequirePositiveSmallInt` in my earlier rounds: The error message after this change is somewhat misleading: `opL` only needs to be a positive int, not necessarily small; the large int case simply is handled before. Of course one can argue that one does not care about this very minor regression, but one should at least be aware of it (and perhaps add a comment pointing this out?)

tst/testinstall/perm.tst
```diff
@@ -208,7 +208,7 @@ gap> List(permAll, x -> x^-3);
 # PowIntPerm
 #
 gap> 0^(1,2);
-Error, Perm. Operations: <point> must be a positive integer (not 0)
+Error, PermImage: <point> must be a positive small integer (not the integer 0)
 gap> n:=10^30;;
 gap> ForAll(permAll, g -> n^g = n);
```
fingolfin: Note the contradiction between this test, and the error message above (10^30 is not a positive small integer).

ChrisJefferson:
Now improved permutation error message.

tst/testinstall/perm.tst
```diff
@@ -223,7 +223,8 @@ gap> List([1,2,3,4,1000],n->List(permBig, g->n^g));
 # QuoIntPerm
 #
 gap> 0/(1,2);
-Error, Perm. Operations: <point> must be a positive integer (not 0)
+Error, PermQuotient: <point> must be a positive small integer (not the integer\
```
fingolfin: The same problem appears here..

ChrisJefferson: woops, you are right. Now fixed, with better warnings

Issue #2996, opened 2018-11-12
Warning message from GRAPE 4.8.1 about variable `Stabiliser' on startup


Stefan-Kohl:
In GAP 4.10.0, when GRAPE is loaded on startup, a warning message is displayed about the variable 'Stabiliser'. The remedy for this appears to be to replace 'Stabiliser' by `Stabilizer' in line 3963 in gap-4.10.0/pkg/grape-4.8.1/lib/grape.g.

alex-konovalov:
No such warnings is seen in GRAPE's tests https://travis-ci.org/gap-system/gap-docker-pkg-tests/jobs/453290410 - likely because the tests load it at a later stage. Which exactly setup do you use to make it being loaded on startup?

Anyhow, GRAPE has a repository on GitHub and I will now test the new GitHub beta-feature to transfer this issue there.

alex-konovalov:
... and it does not work across different organisations. So, you can suggest the change to @lhsoicher by submitting a pull request to https://github.com/gap-packages/grape/.

wilfwilson:
Thanks for the report @Stefan-Kohl. This is a duplicate of an issue that I raised two months ago: https://github.com/gap-system/gap/issues/2828

Although this particular issue could be resolved by adding a workaround to GRAPE (changing the `Stabiliser` to `Stabilizer`; roughly, changing from the British spelling to American), I don't feel like this is the proper solution. I feel like this is instead a problem with GAP or its documentation.

It seems like the problem is: when a package is autoloaded via the `gap.ini` file, and that package uses synonyms defined in the `lib/transatl.g` file, then an `unbound global variable` warning occurs. Presumably, the `lib/transatl.g` file is being read *after* packages are autoloaded from `gap.ini`. And so at the point that GRAPE is autoloaded, `Stabiliser` has not yet been declared as a synonym to `Stabilizer`.

I see several more proper solutions.

Best solution to this issue:
* `lib/transatl.g` should be read earlier in the GAP loading process, in particular, before reading the `gap.ini` file.

If that is not possible for some good reason:
* It should be documented that autoloading of packages should be done via the `gaprc` file rather than the `gap.ini` file. As things stand, I feel like loading GRAPE by setting it to autoload in my `gap.ini` file is appropriate behaviour.

Or (I don't like this idea):
* It should be documented that packages cannot use synonyms declared in `lib/translat.g`. If this decision is made, then the behaviour in GRAPE is a bug.

I will attempt to experiment with the first solution.

Issue #2828, opened 2018-09-19
Stabiliser unbound when loading Grape: is 'Transatlantic' being run at the appropriate time?


wilfwilson:
In GAP master branch, if I load Grape separately after startup, then everything is fine:
```
$ gap.sh -A
 ┌───────┐   GAP 4.10dev-1163-g94ead70 of today
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin17.7.0-default64
 Configuration:  gmp 6.1.2
 Loading the library and packages ...
 Packages:   GAPDoc 1.6.1.dev, PrimGrp 3.3.1, SmallGrp 1.3, TransGrp 2.0.4
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap> LoadPackage("grape");
─────────────────────────────────────────────────────────────────────────────
Loading  GRAPE 4.8 (GRaph Algorithms using PErmutation groups)
by Leonard H. Soicher (http://www.maths.qmul.ac.uk/~lsoicher/).
Homepage: http://www.maths.qmul.ac.uk/~lsoicher/grape/
─────────────────────────────────────────────────────────────────────────────
true
```
But if Grape autoloads, then there is a complaint about an unbound global variable:
```
$ gap.sh
 ┌───────┐   GAP 4.10dev-1163-g94ead70 of today
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin17.7.0-default64
 Configuration:  gmp 6.1.2
 Loading the library and packages ...
Syntax warning: Unbound global variable in /Users/Wilf/gap/pkg/grape/lib/grape\
.g:3963
      newdelta:=InducedSubgraph(delta,A,Stabiliser(delta.group,c,OnSets));
                                                  ^
#I  method installed for Matrix matches more than one declaration
 Packages:   Digraphs 0.13.0, GAPDoc 1.6.1.dev, genss 1.6.5, GRAPE 4.8,
             IO 4.5.3, NautyTracesInterface 0.2, orb 4.8.1, PrimGrp 3.3.1,
             Semigroups 3.0.19, SmallGrp 1.3, Smallsemi 0.6.11,
             TransGrp 2.0.4
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap>
```
[...]

PR #3010, opened 2018-11-16
Read lib/transatl.g before reading gap.ini

wilfwilson:
I have made this PR as a result of what I think would be the best solution to #2828 and #2996.

Currently, `lib/transatl.g` is read in `init.g` a bit later than `gap.ini`. This means that if a package is autoloaded from `gap.ini` and the package uses a name declared in `lib/transatl.g`, then an `unbound global variable` warning is given. This happens with GRAPE 4.8.1 in particular.

This PR moves the reading of `lib/transatl.g` to the point directly before reading `gap.ini`.

Anyway, I've created this PR as a discussion point. There might be very good reasons for why this is a bad idea; in this case, I will be keen to hear about them!

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -696,6 +696,14 @@ function()
     fi;
 end);
 
+
+############################################################################
+##
+#X  Name Synonyms for different spellings
+##
+ReadLib("transatl.g");
+
+
 CallAndInstallPostRestore( function()
     READ_GAP_ROOT( "gap.ini" );
 end );
@@ -789,13 +797,6 @@ CallAndInstallPostRestore( function()
     # user preference `UseColorPrompt'
     ColorPrompt( UserPreference( "UseColorPrompt" ) );
 end );
-
-
-############################################################################
-##
-#X  Name Synonyms for different spellings
-##
-ReadLib("transatl.g");
 
 
 #############################################################################


fingolfin:
Can you please edit the issue message to say `Resolve #XYZ` for all (both?) issues that would be resolved by merging this? Thank you!

fingolfin:
AFAICT, @hulpke originally wrote this code, so he might have additional insights. And perhaps some of the other people who've been around longer, such as @stevelinton @frankluebeck @ThomasBreuer ?

wilfwilson:
@fingolfin Ah, sorry, I used `Closes` rather than `Resolve` – now I think about it, I suppose this closes the issue automatically *without* resolving it?

wilfwilson:
Oh wait, no, there is no separate notion of closed or resolved. Is it okay how it is @fingolfin?

alex-konovalov:
The consideration could be that e.g. before calling `Transatlantic(IsPSolvable)` one should have `IsPSolvable` declared already. Reading it earlier like in this PR does not break this. So it looks good for me then.

fingolfin:
@wilfwilson it is OK as it is -- I didn't see that you have the `Closes` lines in the commit messages, only looked at the PR description (where I usually put them, to make them more visible -- this is not a hidden request to change that here, BTW, just an explanation for my thought process :)

ThomasBreuer:
I have no string preferences about this topic, but since I have been asked:

- An argument for not reading `transatl.g` at all would be that due to the synonyms,
  tab completion stops before the end of the name even if the function is unique.

- An argument for reading `transatl.g` as late as possible would be
  that then it would be easier to find occurrences of function names with `grep`.

- The selection of function names for which transatlantic variants are available looks quite arbitrary.
  For example attribute testers and setters are not handled.
  The fact that this has not been observed indicates that the feature is used rarely.
  (The code would be able to deal with `DeclareSynonymAttr`, this is just not used.
  And of course one could simplify the code.)

lhsoicher:
Thanks @wilfwilson.  I agree with you. That said, I am happy to change any "Stabiliser" to "Stabilizer" for the next release of GRAPE.

PR #3014, opened 2018-11-16
Align gac compiler error checks with rest of GAP

fingolfin:
That is, use some of the Require* macros for most `CHECK_*` macros used by C code generated by gac. Also remove some `ErrorQuitFOO` helpers only used by `CHECK_`; and unify error messages for function calls with too few arguments between compiled and interpreted code.


src/compiled.h
```diff
@@ -87,32 +87,35 @@ typedef UInt    RNam;
 
 /* checks, should go into 'gap.c'  * * * * * * * * * * * * * * * * * * * * */
 
-#define CHECK_BOUND(obj,name) \
- if ( obj == 0 ) ErrorQuitBound(name);
+#define CHECK_BOUND(obj, name)                                               \
```
ChrisJefferson: This change doesn't need to be made, but I wonder if all these could just be functions instead of macros?

PR #3015, opened 2018-11-16
Misc kernel changes

fingolfin:
This is a bunch of mostly unrelated kernel changes, although three of them were created with the aim to reduce the places that use `IMMUTABLE` (I am still very slowly plodding towards turning `IMMUTABLE` into an object flag).

If people find this to annoying to review, I can of course also split it into multiple PRs.

### DIFF
--- a/src/lists.h
+++ b/src/lists.h
@@ -480,6 +480,11 @@ extern void UnbListDefault( Obj list, Int  pos );
 static inline void UNB_LIST(Obj list, Int pos)
 {
     GAP_ASSERT(pos > 0);
+    UInt tnum = TNUM_OBJ(list);
+    if (FIRST_LIST_TNUM <= tnum && tnum <= LAST_LIST_TNUM &&
+        (tnum & IMMUTABLE)) {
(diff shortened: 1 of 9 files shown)


src/objects.h
```diff
@@ -528,6 +528,13 @@ static inline void SET_TYPE_OBJ(Obj obj, Obj type)
 */
 extern void MakeImmutable( Obj obj );
 
+static inline void MakeImmutableNoRecurse(Obj obj)
```
ChrisJefferson: I'm a bit worried that this method does quite a lot less than `MakeImmutable`, for example MakeImmutable does things to the various object types.

Is it intended that this should be called on such objects, and not do anything?

ChrisJefferson:
I'm happy with everything except MakeImmutableNoRecurse, where I was just wondering if we either want a better comment, or a GAP_ASSERT to forbid it being used in things like posobj, if we don't want it calling on such objects.

src/objects.h
fingolfin: Just to add: thanks for looking carefully. I should have thought about adding that comment myself

PR #3020, opened 2018-11-17
Fix `ci-gather-coverage.sh`

### LINKED ISSUE #3019: Coverage reporting for GAP code broken
At some point very recently, coverage reported by codecov went up from 83.8 to 85.2 % simply because it stopped reporting anything on GAP coverage. It now only shows coverage for files in src.

Perhaps one of my recent changes to the travis / ci setup are to blame; or perhaps this is an issue with the profiling package. Perhaps @ChrisJefferson or @alex-konovalov have any insights?

markuspf:
`--nointeract` prevents a session to be started, and hence input on stdin has
no effect.

Closes #3019

### DIFF
--- a/etc/ci-gather-coverage.sh
+++ b/etc/ci-gather-coverage.sh
@@ -18,7 +18,7 @@ fi
 SRCDIR=${SRCDIR:-$PWD}
 
 # Make sure any Error() immediately exits GAP with exit code 1.
-GAP="bin/gap.sh --quitonbreak --nointeract --alwaystrace -q"
+GAP="bin/gap.sh --quitonbreak --alwaystrace -q"
 
 # change into BUILDDIR (creating it if necessary), and turn it into an absolute path
 if [[ -n "$BUILDDIR" ]]
@@ -46,11 +46,17 @@ for f in DirectoryContents(d) do
     Add(covs, f);
     Print("  ", f, "\n");
 od;
-Print("Merging ", Length(covs), " coverage files...\n");
-r := MergeLineByLineProfiles(covs);;
-Print("Outputting JSON\n");
-OutputJsonCoverage(r, "gap-coverage.json");;
-QUIT_GAP(0);
+if Length(covs) > 0 then
+    Print("Merging ", Length(covs), " coverage files...\n");
+    r := MergeLineByLineProfiles(covs);;
+    Print("Outputting JSON\n");
+    OutputJsonCoverage(r, "gap-coverage.json");;
+else
+    # Don't error, because we might want to gather
+    # gcov coverage, so just inform that we didn't find
+    # GAP coverage data
+    Print("No coverage files found...\n");
+fi;
 GAPInput
 
 # generate kernel coverage reports by running gcov


markuspf:
Willing to drop either the commit handling empty coverage directories or setting `NO_COVERAGE=1` for some tests.

ChrisJefferson:
Is the NO_COVERAGE=1 just about speeding things up by doing less coverage of tests we run more than once?

.travis.yml
```diff
@@ -34,7 +34,7 @@ matrix:
     # digraphs and other kernel extensions using C++ code (due to GCC's
     # `profiler_trace.h` calling `atexit`; but that fails in loadable modules
     # for technical reasons)
-    - env: TEST_SUITES=testpackages CFLAGS="-O2" LDFLAGS=
+    - env: TEST_SUITES=testpackages NO_COVERAGE=1 CFLAGS="-O2" LDFLAGS=
```
fingolfin: I don't understand, what has this change to do with the description of the PR? And what is its motivation?

markuspf:
I believe `NO_COVERAGE` was introduced by @fingolfin to reduce load on codecov.io, because apparently the number of lines we uploaded was pushing the limits of what they were willing to do. Since testpackages is run without coverage anyway, there's no need to scan for coverage files.

But also if there are no coverage files, should tests fail, after all it might be a good thing to notice if you expect coverage to be produced and it isn't.

Incidentally we might even want to make `ci-gather-coverage` fail if there's no coverage.json produced.

.travis.yml
markuspf: If you just take the first commit, `testpackages` now fails because no coverage output is produced, and `MergeLineByLineProfiles` errors.

Two fixes are possible: Don't error even if the list of profiles to merge is empty (which means we also do not detect when there wasn't any profiling data produced erroneously), or explicitly don't expect coverage.

Or, if we do expect coverage for `testpackages`, then there is a bug in how these tests are run as they don't produce coverage.

fingolfin:
Actually, `git log -G NO_COVERAGE .travis.yml etc` shows that `NO_COVERAGE` was added by @ChrisJefferson  in 4a3da99bd7ac810888b717b9d43664a7c20a4c0c to allow us to test things with`ReproducibleBehaviour` turned off (we don't want this with coverage, as it causes random fluctuations in coverage). Then later I used it reduce load on codecov.io.

.travis.yml
fingolfin: OK, so no I understand why `NO_COVERAGE` is used here. I'd greatly appreciate if the comment above this was extended to mention this, too (just like it explains about `--coverage` being dropped).

.travis.yml
```diff
@@ -74,7 +74,7 @@ matrix:
     #  compiler: clang
 
     # test creating the manual
-    - env: TEST_SUITES=makemanuals
+    - env: TEST_SUITES=makemanuals NO_COVERAGE=1
```
fingolfin: I assume the `NO_COVERAGE=1` here is only added for performance reasons? Regardless of that, a comment above would again be appreciated; it's already difficult enough to navigate and understand our `.travis.yml` :-)

.travis.yml
fingolfin: Perhaps like this:
> Since no coverage data is produced anyway, set `NO_COVERAGE=1` to avoid calling the coverage submission code.

markuspf:
Squashed and updated commit message.

For completeness: I removed the `NO_COVERAGE=1` "hack" because the situation was a (as usual) tad more complicated: sometimes we wanted to gather C-only coverage. Since I was in no mood to re-engineer `ci-gather-coverage.sh`, the current solution was the simplest that restores GAP code coverage.

PR #3027, opened 2018-11-19
Prof improve weird lines

ChrisJefferson:
This will hopefully deal with some weird lines appearing in profiles from variables not being correctly reset (going to look at the result of this PR)

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -500,6 +500,7 @@ UInt OpenInput (
         GAP_ASSERT(IS_CHAR_PUSHBACK_EMPTY());
         IO()->Input->ptr = STATE(In);
         IO()->Input->symbol = STATE(Symbol);
+        IO()->Input->interpreterStartLine = STATE(InterpreterStartLine);
     }
 
     /* enter the file identifier and the file name                         */
@@ -521,6 +522,7 @@ UInt OpenInput (
     STATE(In) = IO()->Input->line;
     STATE(In)[0] = STATE(In)[1] = '\0';
     STATE(Symbol) = S_ILLEGAL;
+    STATE(InterpreterStartLine) = 0;
     IO()->Input->number = 1;
 
     /* indicate success                                                    */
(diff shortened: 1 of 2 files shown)


fingolfin:
Unfortunately Codecov did not add a coverage report comment here. Nor did it for PR #3023. This may be Codecov being flaky again as usual; or perhaps there is still something flaky with coverage uploading from Travis to Codecov ... but then again,  https://codecov.io/gh/gap-system/gap/commit/2820560658a999701014792d75130da307972086/build  shows Travis and AppVeyor data just fine, and we can see coverage for this PR on https://codecov.io/gh/gap-system/gap/tree/2820560658a999701014792d75130da307972086

Judging from that, the spurious code markers in comments are gone.

There are still `fi;` lines marked as "code that was not executed, though. See e.g. https://codecov.io/gh/gap-system/gap/src/2820560658a999701014792d75130da307972086/lib/memusage.gd#L46 and https://codecov.io/gh/gap-system/gap/src/2820560658a999701014792d75130da307972086/lib/random.g#L71 -- but I assume those are simply not in scope for this PR.

ChrisJefferson:
Yup, I'll look at those `fi;` seperately. In general those just have to be special-cased, because in GAP those types of things often are technically read but never "executed", but that obviously isn't helpful :)

fingolfin:
Well, it would also be perfectly fine to me if `fi` was simply "gray", i.e., not considered to be code. But marking it as unexecuted code is annoying.

ChrisJefferson:
Oh I agree, it needs fixing. Just need to decide which way to handle it, write some tests, that kind of thing :)

alex-konovalov:
@ChrisJefferson I am puzzled that the label backport-4.10 has been removed, but this is still assigned to the milestone 4.10.1. I think that the initial intention was to backport it, but then it was decided to not to do it - in this case, both the label and the milestone should be updated.

PR #3031, opened 2018-11-20
Add ConjugacyClassesForSolvableGroup

fingolfin:
.. and use it in two ConjugacyClasses methods: the generic one
for finite groups, and the one for permutation groups. Also merge
the two existing "generic" methods for finite groups (they had the
same rank, which is rather brittle).

This speeds up the computation of ConjugacyClasses for e.g.

    G:=WreathProduct(CyclicGroup(IsPermGroup,12), DihedralGroup(IsPermGroup,12));

from 110 seconds to 53 seconds on my laptop.

### DIFF
--- a/lib/clas.gi
+++ b/lib/clas.gi
@@ -433,40 +433,33 @@ InstallGlobalFunction(ConjugacyClassesForSmallGroup,function(G)
   fi;
 end);
 
+InstallGlobalFunction( ConjugacyClassesForSolvableGroup,
+  function( G )
+  local   cls,  cl,  c;
 
-InstallMethod( ConjugacyClasses, "for groups: try random search",
+  cls := [  ];
+  for cl  in ClassesSolvableGroup( G, 0 )  do
+    c := ConjugacyClass( G, cl.representative, cl.centralizer );
+    Assert(2,Centralizer(G,cl.representative)=cl.centralizer);
+    Add( cls, c );
+  od;
+  Assert(1,Sum(cls,Size)=Size(G));
+  return cls;
+end );
+
+InstallMethod( ConjugacyClasses, "for groups: try solvable method and random search",
   [ IsGroup and IsFinite ],
 function(G)
 local cl;
   cl:=ConjugacyClassesForSmallGroup(G);
   if cl<>fail then
     return cl;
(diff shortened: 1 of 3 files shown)


lib/clashom.gi
```diff
@@ -3061,6 +3061,8 @@ local cl;
   cl:=ConjugacyClassesForSmallGroup(G);
   if cl<>fail then
     return cl;
+  elif IsSolvableGroup( G ) and CanEasilyComputePcgs(G) then
```
fingolfin: I added the "raw" `IsSolvableGroup` check here because (a) it is reasonably efficient for permutation groups, and its result can be useful in many subsequent computations, too, and (b) I felt that the existing check for `IsSimpleGroup` later on is not really cheaper either. But of course I might be mistaken -- but I rely on @hulpke to point that out  :-)

lib/clas.gi
```diff
@@ -433,41 +433,34 @@ InstallGlobalFunction(ConjugacyClassesForSmallGroup,function(G)
...
+end );
+
+InstallMethod( ConjugacyClasses, "for groups: try solvable method and random search",
   [ IsGroup and IsFinite ],
 function(G)
 local cl;
   cl:=ConjugacyClassesForSmallGroup(G);
   if cl<>fail then
     return cl;
+  elif IsSolvableGroup( G ) and CanEasilyComputePcgs(G) then
```
fingolfin: Here perhaps we should instead start with `HasIsSolvableGroup( G ) and ...`, as this code could be called on arbitrary groups. The only reason I didn't do it was that the existing method didn't do it either.

BTW, `CanEasilyComputePcgs` is a category, and set by some implications once the group "knows" that it is solvable. So the order of the check here is crucial.

lib/clashom.gi
hulpke: For permutation groups `IsSimpleGroup` is not as expensive as it might seem, as it uses tools from the composition series mechanism. In most cases order alone bails out quickly.
IsSolvableGroup should be cheap enough, as the code running afterwards effectively would discover that the group is solvable.

hulpke:
(I think the time overhead in the code is purely with storing informations in the centralizers about how they fit with the chosen series -- no reason to not make this change but also not to worry about the other code. It is an artifect of the huge number of classes)

PR #3032, opened 2018-11-20
Fix marking 'fi;' as executed in profiles

ChrisJefferson:
This fixes 'fi;' in profiles. There are no tests, I will add tests to the 'profiling' package (but by inspection, it seems to produce the right answer).

### DIFF
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -89,10 +89,14 @@
 // generation and execution. Otherwise, we always mark the line as
 // read, and mark as executed if STATE(IntrReturning) and STATE(IntrIgnoring)
 // are both false.
-#define INTERPRETER_PROFILE_HOOK()                                           \
+//
+// IgnoreLevel gives the highest value of IntrIgnoring which means this
+// statement is NOT ignored (this is usually, but not always, 0)
+#define INTERPRETER_PROFILE_HOOK(ignoreLevel)                                \
     if (!STATE(IntrCoding)) {                                                \
         InterpreterHook(GetInputFilenameID(), STATE(InterpreterStartLine),   \
-                        STATE(IntrReturning) || STATE(IntrIgnoring));        \
+                        STATE(IntrReturning) ||                              \
+                            (STATE(IntrIgnoring) > ignoreLevel));            \
     }                                                                        \
     STATE(InterpreterStartLine) = 0;
 
@@ -100,7 +104,7 @@
 // Put the profiling hook into SKIP_IF_RETURNING, as this is run in
 // (nearly) every part of the interpreter, avoid lots of extra code.
 #define SKIP_IF_RETURNING()                                                  \
-    INTERPRETER_PROFILE_HOOK();                                              \
+    INTERPRETER_PROFILE_HOOK(0);                                             \
     SKIP_IF_RETURNING_NO_PROFILE_HOOK();
 
 // Need to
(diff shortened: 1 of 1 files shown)


fingolfin:
Awesome. Now I wonder would the `od` in `for` and `while` loops possibly be affected, too?

ChrisJefferson:
They seem to be fine, because in `while` and `repeat` the interiors are (a) coded, and (b) never ignored. There might be a similar bug if someone wrote a multi-line `Info` or `Assert` statement, where we also try to ignore bits of code, but long multi-line statements sometimes slip up profiling already.

src/intrprtr.c
```diff
@@ -3921,8 +3928,12 @@ void            IntrInfoEnd( UInt narg )
...
 
     /* ignore or code                                                      */
-    SKIP_IF_RETURNING();
-    if ( STATE(IntrIgnoring)  > 1 ) { STATE(IntrIgnoring)--; return; }
+     INTERPRETER_PROFILE_HOOK(1);
+     SKIP_IF_RETURNING_NO_PROFILE_HOOK();
+
+     if (STATE(IntrIgnoring) > 1) {
+         STATE(IntrIgnoring)--;
+         return; }
```
fingolfin: I'd either keep this on one line, as it was; or else fully clang-format it (and get that brace on its own line)

src/intrprtr.c
```diff
@@ -659,15 +663,18 @@ Int            IntrIfEndBody (
...
+    SKIP_IF_RETURNING_NO_PROFILE_HOOK();
+
     if ( STATE(IntrIgnoring)  > 1 ) { STATE(IntrIgnoring)--; return; }
 
     // if one branch was executed (ignoring the others), reset IntrIgnoring
     if ( STATE(IntrIgnoring) == 1 ) {
         STATE(IntrIgnoring) = 0;
     }
 
+
```
fingolfin: Remove extra empty line?

src/intrprtr.c
```diff
@@ -4023,8 +4034,11 @@ void             IntrAssertAfterCondition ( void )
 void             IntrAssertEnd2Args ( void )
 {
       /* ignore or code                                                      */
-    SKIP_IF_RETURNING();
-    if ( STATE(IntrIgnoring)  > 2 ) { STATE(IntrIgnoring) -= 2; return; }
+      INTERPRETER_PROFILE_HOOK(2);
+      SKIP_IF_RETURNING_NO_PROFILE_HOOK();
+      if (STATE(IntrIgnoring) > 2) {
+          STATE(IntrIgnoring) -= 2;
+          return; }
```
fingolfin: see above

src/intrprtr.c
```diff
@@ -4023,8 +4034,11 @@ void             IntrAssertAfterCondition ( void )
 void             IntrAssertEnd2Args ( void )
 {
       /* ignore or code                                                      */
-    SKIP_IF_RETURNING();
-    if ( STATE(IntrIgnoring)  > 2 ) { STATE(IntrIgnoring) -= 2; return; }
+      INTERPRETER_PROFILE_HOOK(2);
```
fingolfin: Huh, the indentation looks off, at least compared to what was there before?!?

src/intrprtr.c
ChrisJefferson: Weirdly, this was clang-formatted. However, clang-format seems to have not reached a fixed point, and re-clang-formatting changes it again.

src/intrprtr.c
ChrisJefferson: the old formatting was 2 space (which nothing else in this file is), that's clang-format moving it to 4 spaces.

ChrisJefferson:
Now made all formatting (hopefully) reasonable, and locally consistent.

alex-konovalov:
@ChrisJefferson I am puzzled that the label backport-4.10 has been removed, but this is still assigned to the milestone 4.10.1. I think that the initial intention was to backport it, but then it was decided to not to do it - in this case, both the label and the milestone should be updated.

fingolfin:
This is not needed in the release notes, because it fixes a bug in a feature that was newly added (i.e., does not exist in 4.10)

PR #3033, opened 2018-11-20
Simplify Reread, Re(re)adAndCheckFunc

fingolfin:
Drop support for unspecified extra arguments, which are not used anywhere.

Also reset `REREADING` to false in `OnQuit`, to fix a bug where an erroneous
use of e.g. `RereadGrp` would leave `REREADING = true` permanently (resp.
until the next succesful call to a `Reread*` function).


fingolfin:
Yes, Reread and RereadLib are tested. The former by a test I explicitly wrote for that purpose, the latter by a bugfix test.

PR #3034, opened 2018-11-20
kernel: fix compiler error with GCC 4.4.7

### LINKED ISSUE #3026: Compilation problem with 4.10
### Observed behaviour

when compiling I get an error

    C       src/read.c => obj/src/read.lo
    src/read.c: In function ‘ReadVar’:
    src/read.c:619: error: unknown field ‘var’ specified in initializer
    src/read.c:619: warning: missing braces around initializer
    src/read.c:619: warning: (near initialization for ‘ref.<anonymous>’)
    src/read.c:619: error: unknown field ‘nest0’ specified in initializer
    At top level:
    cc1: warning: unrecognized command line option "-Wno-cast-function-type"
    cc1: warning: unrecognized command line option "-Wno-implicit-fallthrough"

### Expected behaviour

GAP should compile. (I can hack around it, but others might not be able to) 

gcc version 4.4.7 20120313 (Red Hat 4.4.7-23) (GCC)

(This is a system with lots of memory I have access to but no root access. For stability reasons it is still using an older RedHat, so "update the compiler" is not an option for me)

fingolfin:
Fixes #3026

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -616,7 +616,7 @@ static void ReadReferenceModifiers(TypSymbolSet follow)
 */
 static LHSRef ReadVar(TypSymbolSet follow)
 {
-    LHSRef ref = { .type = R_INVALID, .var = 0, .nest0 = 0 };
+    LHSRef ref = { R_INVALID, {0}, {0} };
 
     Obj  nams;                      // list of names of local vars.
     Obj  lvars;                     // environment


ChrisJefferson:
Just to be super-safe, I'll let the tests cycle, then whoever sees the green ticks first can merge

alex-konovalov:
Backported to stable-4.10 in 986d08a9387941988e7bec9379b041c57d37cb2d

PR #3036, opened 2018-11-20
Add helper function to get Stats from function bags

markuspf:
This will be used in #2628 and could also be used in the GAP to C compiler.


fingolfin:
I don't think this is the way to go. Is there any point in keeping this PR open?

Issue #3040, opened 2018-11-21
New `#@if` syntax in test files not supported by `rewriteToFile` option


fingolfin:
In the past I often used commands like  `TestDirectory("tst/testinstall", rec(rewriteToFile:=true));` or variations to rewrite all tests in `testinstall`. 

This, however, fails badly with test files using the new `#@if` syntax, as those `#@if` and `#@else` etc. commands get discarded.

It would be great if this could be fixed. Perhaps @ChrisJefferson would be willing to look into it?

ChrisJefferson:
Ooo, good point! I didn't consider that case.

Bla, this could get awkward. Let me see if I can think of something useful. The previous `#@skipif` ,which skips a whole file, would actually have worked in these situations. Might have to go back to that.

ChrisJefferson:
Might have a good idea.

fingolfin:
I would "just" write out the `#@if` etc. lines as they were; and likewise for the block we did not take. All the data is there, it's just a bit fiddly to access it at the right moment, I think (then again, I might be mistaken, I haven't actually looked at the code yet)

ChrisJefferson:
I have tigtened up the definition of `#@if`, and now it's easy. The problem was you could have `#@if` anywhere, and then we don't even know how much of the new output to put before, inside, or after the `#@if`.

By saying `#@if` can only wrap complete tests (by which I mean a block of input followed by a block out of the expected output, not a whole .tst file), it becomes easy to fix things.

#3041

fingolfin:
Fixed by PR #3041

PR #3041, opened 2018-11-21
Fix rewriteToFile handling with #@if

ChrisJefferson:
This limits #@if / #@else / #@endif to only appear between tests (just before a 'gap>', or at the end of the file). This makes it easy to support rewriteToFile (which is also fixed).

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -12,7 +12,7 @@
 ##  end;
 
 InstallGlobalFunction(ParseTestInput, function(str, ignorecomments)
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Fixes #3040

fingolfin:
I think the "fixes" line must be in the PR description to have any effect

fingolfin:
@ChrisJefferson As I suspected, the "fixes" line only works if it is in the PR description.

PR #3042, opened 2018-11-22
buildsys: always generate sysinfo.gap

fingolfin:
By accident, sysinfo.gap was only generated if the "compatibility mode"
of the build system was enabled. This is currently still on by default,
but I'd like to change that in a future release of GAP. This is a first
step towards this goal.

This PR was actually motivated by the build issue with `NormalizInterface` in @stevelinton's PR #2997 -- while looking at it, I noticed that compat mode still is on, and I never formulated a roadmap for phasing it out.

I'd like to backport this minor change to GAP 4.10, if nobody minds.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -658,15 +658,6 @@ distclean-extern:
 	rm -rf extern/install
 
 
-########################################################################
-# Compatibility mode
-#
-# If enabled, we prepare the environment to look like it did with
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Backported in aa080f11e82968b83c0032ef2d928d017f2d0555

PR #78, opened 2015-03-18
WeakPointer: Check for exclusive write access in LengthWPObj

markuspf:
* As per Reimer's suggestion check for exclusive write access in
  LengthWPObj.
* If we have write access, we discard objects at the end of the
  list, otherwise we just return the stored length.

PR #3043, opened 2018-11-22
Add support for writing to ERROR_OUTPUT from kernel code

embray:
#2824 added a global variable called `ERROR_OUTPUT` to `lib/error.g`, which can be manipulated to control the file/stream that error output is sent to by the error library.  This works for most cases; however, code in the kernel that outputs error messages ought also to write to whatever `ERROR_OUTPUT` points to, rather than defaulting to `"*errout*".

The one case I found this to be an issue was in printing syntax errors, though there may be other cases.

This is my first attempt to work around it, and my first addition of new code to GAP, so please be kind--I've tried my best to observe and follow local conventions, etc.

One problem with this PR I'm aware of currently is that when using `ImportGVarFromLibrary`, it sets that global variable as read-only by default, which was not previously the case for `ERROR_OUTPUT`.  I'm not exactly sure why it does this, as it is still possible to set the variable read-write afterwards.  This may be a good change though. It would prevent accidental clobbering of `ERROR_OUTPUT`: In order to change it you really have to mean it.  I've seen this pattern in use in a couple of packages when overriding `ErrorInner` as well, so there is precedent for it.  But I don't know whether or not that's considered an anti-pattern.

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -33,12 +33,48 @@
 
 
 static Obj ErrorInner;
+static Obj ERROR_OUTPUT = NULL;
 
 
 /****************************************************************************
 **
 *F * * * * * * * * * * * * * * error functions * * * * * * * * * * * * * * *
 */
+
+/****************************************************************************
+**
+*F  OpenErrorOutput()  . . . . . . . open the file or stream assigned to the
+**                                   ERROR_OUTPUT global variable defined in
+**                                   error.g, or "*errout*" otherwise
+*/
+UInt OpenErrorOutput( void )
+{
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    UInt ret = 0;
+
+    if (ERROR_OUTPUT != NULL) {
+        if (IsStringConv(ERROR_OUTPUT)) {
+            ret = OpenOutput(CONST_CSTR_STRING(ERROR_OUTPUT));
+        }
+        else {
+            ret = OpenOutputStream(ERROR_OUTPUT);
+        }
+    }
+
+    if (!ret) {
+        /* It may be we already tried and failed to open *errout* above but
+         * but this is an extreme case so it can't hurt to try again
+         * anyways */
+        if ((ret = OpenOutput("*errout*"))) {
+            Pr("failed to open error stream\n", 0, 0);
+        }
+        else {
+            Panic("failed to open *errout*");
+        }
+    }
+
+    return ret;
+}
 
 
 /****************************************************************************
@@ -581,6 +617,7 @@ static Int InitKernel(StructInitInfo * module)
     InitHdlrFuncsFromTable(GVarFuncs);
 
     ImportFuncFromLibrary("ErrorInner", &ErrorInner);
+    ImportGVarFromLibrary("ERROR_OUTPUT", &ERROR_OUTPUT);
 
     // return success
     return 0;
--- a/src/error.c
+++ b/src/error.c
@@ -65,7 +65,8 @@ UInt OpenErrorOutput( void )
         /* It may be we already tried and failed to open *errout* above but
          * but this is an extreme case so it can't hurt to try again
          * anyways */
-        if ((ret = OpenOutput("*errout*"))) {
+        ret = OpenOutput("*errout*");
+        if (ret) {
             Pr("failed to open error stream\n", 0, 0);
         }
         else {
--- a/src/error.c
+++ b/src/error.c
@@ -34,6 +34,7 @@
 
 static Obj ErrorInner;
 static Obj ERROR_OUTPUT = NULL;
+static Obj IsOutputStream;
 
 
 /****************************************************************************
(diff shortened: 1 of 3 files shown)


src/error.c
```diff
@@ -33,13 +33,49 @@
...
+        else {
+            ret = OpenOutputStream(ERROR_OUTPUT);
+        }
+    }
+
+    if (!ret) {
+        /* It may be we already tried and failed to open *errout* above but
+         * but this is an extreme case so it can't hurt to try again
+         * anyways */
+        if ((ret = OpenOutput("*errout*"))) {
```
fingolfin: Please don't assign inside the `if`, i.e., first assign, then test.

src/error.c
```diff
@@ -33,13 +33,49 @@
...
+{
+    /* Try to print the output to stream. Use *errout* as a fallback. */
+    UInt ret = 0;
+
+    if (ERROR_OUTPUT != NULL) {
+        if (IsStringConv(ERROR_OUTPUT)) {
+            ret = OpenOutput(CONST_CSTR_STRING(ERROR_OUTPUT));
+        }
+        else {
+            ret = OpenOutputStream(ERROR_OUTPUT);
```
fingolfin: We started to add more validation, and here, one could verify that ERROR_OUTPUT actually is a stream; see e.g. `PRINT_OR_APPEND_TO_FILE_OR_STREAM`.

On that note, perhaps it would be "better" to put this new function into `streams.c` or `io.c`...; but I have no strong conviction in either direction, and in any case, it's not hard to move a function to another file later on, should we feel the need for it :-).

embray: I wanted to check that it was actually a stream, but I couldn't quickly find the "right" way to check that.  Is there a TNUM for stream objects?  I'll look at those functions you mentioned.

I agree this function might make more sense in a `io.c`.  I put it in `error.c` because that was already loading another global variable from `error.g`.  I don't know if that entirely matters though: The connection between the C-based modules and the GAP language library modules is one that I haven't deeply explored yet and is unclear to me.

ChrisJefferson: In general, TNUMs only cover a very small, hard-wired list of types with special representations in the GAP kernel (things like integers, booleans, permutations). For things like streams, you want to use the more general `IsOutputStream`. As Max mentioned, you can see how to get this GAP level variable, and copy a reference to it into C, in PRINT_OR_APPEND_TO_FILE_OR_STREAM

embray: Yes, clearly I did not even test my last commit.  I'm just going to delete it.

embray:
I noticed that the JupyterKernel package already relies on setting `ERROR_OUTPUT` in order to capture error output.  So the fact that `ImportGVarFromLibrary` sets that global variable to read-only by default would break that.

I still think it's reasonable behavior that it does that.  But if this change is accepted we should either updated the JupyterKernel code, or just keep this variable read/write by default (and in that case, what would be the best way to do that after `ImportGVarFromLibrary`?)

ssiccha:
Hi @embray. I like this. :) Since the new function is equivalent to what happens at the beginning of `FuncPRINT_CURRENT_STATEMENT`, we can then also delete that code and call the new function there.

embray:
@ssiccha 
> Since the new function is equivalent to what happens at the beginning of FuncPRINT_CURRENT_STATEMENT

It isn't exactly equivalent, since that looks at an arbitrary stream object (in practice it is often the `ERROR_INPUT` which is passed here by `ErrorInner` in `error.g`, but that need not be the case for sure.

But I could refactor a bit more to eliminate that duplicate code.  I don't know how much you all care about moving around GAP kernel internal functions that aren't directly accessible from GAP code, given that there is not officially an "API" as such (outside of the the new one that's being added...)

markuspf:
> One problem with this PR I'm aware of currently is that when using `ImportGVarFromLibrary`, it sets that global variable as read-only by default, which was not previously the case for `ERROR_OUTPUT`. I'm not exactly sure why it does this, as it is still possible to set the variable read-write afterwards. 

I think this is in some sense entirely coincidental. `ERROR_OUTPUT` is introduced in the library by assignment (i.e. `ERROR_OUTPUT := "*errout*"`), which makes it a read-write global. `ImportGVarFromLibrary` just enters the global variable into a list of global variables which are by default read only.
One could view this as a bug in the library which should really have used `BindGlobal("ERROR_OUTPUT", "*errout*");` to introduce this variable as a proper global.

> This may be a good change though. It would prevent accidental clobbering of `ERROR_OUTPUT`: In order to change it you really have to mean it. I've seen this pattern in use in a couple of packages when overriding `ErrorInner` as well, so there is precedent for it. But I don't know whether or not that's considered an anti-pattern.

There might even be a point in introducing a setter function for error output (later). It might indeed be a good idea to have `ERROR_OUTPUT` read only by default.

I'd just like someone (@ssicha?) to make a matching PR for `JupyterKernel` as to not break it, some people seem to be using it now.

ssiccha:
Also I opened a corresponding PR gap-packages/JupyterKernel#85.

markuspf:
I just merged #78 on `JupyterKernel` so I am fine with merging this.

alex-konovalov:
Did we agree to backport this to stable-4.10 branch?

fingolfin:
Backported via de7b9b9d6abd28e84a57bdedea78c39ea49b8989

PR #3045, opened 2018-11-22
kernel: more error handling tweaking

fingolfin:


sebasguts:
My review was based on the fact that @fingolfin said it is trustworthy and the tests pass.

PR #3053, opened 2018-11-24
Preserve more properties in DirectProduct

wilfwilson:
The soon-to-be-released Semigroups 3.1.0 is going to prove much greater support for creating direct products of semigroups. It does this by installing methods for `DirectProductOp` for certain kinds of semigroups, which uses the `DirectProduct` function in the GAP library. When this library function has created a direct product, it tries to see whether it can set some properties that are preserved by direct products and that are known for the factors. So I thought it made sense to increase this list of properties for when the arguments are semigroups.

What do people think? Is this going to slow things down noticeably for groups, where all of these properties are irrelevant? Is there a better way of doing this? Or is it a bad idea?

### DIFF
--- a/lib/gprd.gi
+++ b/lib/gprd.gi
@@ -28,7 +28,9 @@ local d, prop;
 
   # test/set a few properties and attributes from factors
 
-  for prop in [IsFinite, IsNilpotentGroup, IsAbelian, IsSolvableGroup] do
+  for prop in [IsFinite, IsNilpotentGroup, IsAbelian, IsSolvableGroup, IsBand,
+               IsInverseSemigroup, IsRegularSemigroup, IsIdempotentGenerated,
+               IsLeftZeroSemigroup, IsRightZeroSemigroup, IsZeroSemigroup] do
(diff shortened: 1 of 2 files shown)


wilfwilson:
I agree @fingolfin, I'm all for tests adding tests and I agree it would be nice to have comprehensive tests for the stuff I've added. The difficulty here is that `DirectProduct` is declared in the GAP library, while the methods for `DirectProductOp` for constructing direct products of semigroups will be in the Semigroups package. So it will not be possible to test the new things that I have added. I can add such tests in the Semigroups package. Unfortunate, yes, but I don't see a way around this at the moment.

To make you a bit happier I have added a couple of tests relating to the properties that were already implemented for groups :)

wilfwilson:
I've added another commit because I had an idea of how to improve this further. So I suggest that this this should be reviewed again.

Explanation of change: the properties of direct products that are implemented all hold of the direct product if and only if they hold for all of the factors. Therefore, to set the property to be `false` for the direct product, we only need to know that the property is `false` for one of the factors.

PR #3057, opened 2018-11-26
Move `stats.c`'s state GAPState into `stats.c` 

markuspf:


src/stats.c
```diff
@@ -49,6 +49,12 @@ inline UInt EXEC_STAT(Stat stat)
     return (*STATE(CurrExecStatFuncs)[ tnum ]) ( stat );
 }
 
+inline Obj EXEC_STAT_RETURN_OBJ(Stat stat)
+{
+    UInt status = EXEC_STAT(stat);
+    GAP_ASSERT((status == 1) || (status == 2));
```
markuspf: Not sure this is necessary, the functions in `funcs.c` did not do anything with the return value of `EXEC_STAT`, anyway.

src/stats.c
```diff
@@ -41,14 +41,45 @@
...
 
 inline UInt EXEC_STAT(Stat stat)
 {
     UInt tnum = TNUM_STAT(stat);
     SET_BRK_CALL_TO(stat);
-    return (*STATE(CurrExecStatFuncs)[ tnum ]) ( stat );
+    return (*StatsState()->CurrExecStatFuncs[ tnum ]) ( stat );
 }
 
+inline Obj EXEC_STAT_RETURN_OBJ(Stat stat)
```
fingolfin: Note that in all calls, `stat == OFFSET_FIRST_STAT`. So you could also remove the argument, and call this something like `EXEC_CURR_FUNC` or so... I think that would make for a nice documentation comment, too.

markuspf:
Sorry, I didn't see your comment. I decided to back out the stats module state changes because they break with HPC-GAP, but I will make a new PR with your comment incorporated!

PR #3060, opened 2018-11-27
kernel: change IsPossList to require *small* ints (DO NOT MERGE)

fingolfin:
I am conflicted about this change, but I had it sitting around from some earlier refactoring, and I thought that instead of just letting it rot, or outright discarding it, I might as well put it here for some additional scrutiny by others:

Right now, the kernel defines `IsPossList` to mean: "list of positive integers, including big ones". But all actual users of it in the end seem to either require small integers in it, or else treat big positive integers trivially (e.g. permutations fix those, instead of erroring as for negative integers -- but I doubt that this is very useful in practice).

Anyway, as you can see, this change then simplifies some error checking in `trans.c` a lot, which was the original motivation for it. But that's hardly a strong reason on its own; we could add a separat dedicated function to perform these checks, or simply leave the current checking code as it is.

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -475,7 +475,7 @@ void ErrorMayQuit(const Char * msg, Int arg1, Int arg2)
 void CheckIsPossList(const Char * desc, Obj poss)
 {
     if ( ! IS_POSS_LIST( poss ) ) {
-        ErrorMayQuit("%s: <positions> must be a dense list of positive integers",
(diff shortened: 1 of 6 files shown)


ChrisJefferson:
I'm a bit concerned about changing the permutations, as I've personally often described GAP permutations as being defined on "the natural numbers", so it would be a significant change in theory to say they only work on the "small integers".

fingolfin:
That's a fair point. Also, tests fail. I think I'll just ditch this.

PR #3064, opened 2018-11-28
Remove some obsolete code from compiler.c

markuspf:
This code relates to an obsolete way references to local variables used to be
coded as, which has been removed in 8d2ca938.


ChrisJefferson:
woops

PR #3065, opened 2018-11-28
Backport libgap api to 4.10

fingolfin:
This is a backport of all the libgap-api changes done on master, so that these will appear in GAP 4.10.1. I hope this will help folks like @embray who are trying to use this.

Of course we can delay merging this in case people plan to submit more libgap-api changes. Or we just make additional PRs for that.

Note that I had to make some minor modifications to `libgap-api.c`: We don't have `RequireInt` in stable-4.10, nor `NewPlistFromArray`. For the latter, I just added it to `libgap-api.c`; for the former, I replaced the call by (almost) equivalent code using `ErrorMayQuit`. For future updates, one needs to retain those (but that's easy enough).

Note: I did not cherry-pick individual PRs modifying libgap-api, as that would have been more work with no clear gain. Nor did I cherry-pick the commit which added `NewPlistFromArray`, as it also modified several .c files, and while those changes should be harmless, I'd rather go for minimal changes that are guaranteed to not affect code outside `libgap-api.c`.

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -15,7 +15,7 @@
 **  Each integer has a unique representation, e.g., an integer that can be
 **  represented as 'T_INT' is never represented as 'T_INTPOS' or 'T_INTNEG'.
 **
-**  In the following, let 'N' be the number of bits in an mp_limb_t (so 32 or
+**  In the following, let 'N' be the number of bits in an UInt (so 32 or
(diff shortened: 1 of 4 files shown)


embray:
Thanks--I have a couple other patches I've been meaning to propose but we can handle those in a followup.

Issue #3068, opened 2018-11-28
Tabs on continuation lines break GAP's testing


markuspf:
if one uses tab characters on continuation lines in a test file, this confuses GAP:

```
gap> f := function()
>\tlocal a;
>\tif a = 0 then
>\t\tError("a is zero");
>\tfi;
> end;;
```

```gap> Test("tab.tst");
########> Diff in tab.tst:1         
# Input is:
f := function()
	local a;
		Error("a is zero");
end;;
# Expected output:
>	if a = 0 then
>	fi;
# But found:
########
false
```

ChrisJefferson:
There is another bug here, which I think is making this one harder to spot. GAP allows continuation lines in the middle of the output, and puts them as part of the input.

```
gap> x := 2 +
>	cheese
> 2;
3
```

gives:

```
########> Diff in t.tst:1           
# Input is:
x := 2 +
2;
# Expected output:
>	cheese
3
# But found:
4
########
false
```

ChrisJefferson:
This issue has been improved somewhat. Further improves were unfortunately difficult due to the fact that some package rely on not this specific case, but similar cases where tests are formatted strangely.

PR #3069, opened 2018-11-28
Improve error messages in tests

ChrisJefferson:
This deals with GAP allowing `> ` in the middle of tests, and appending the line into the input of the test. While #3068 is not fixed by this, it does make the error (in my opinion) much easier to understand.

Also, this overhauls the errors produced by tests to give a line number, and add some basic tests (which have to be in testspecial, because they involve test files which fail).


tst/testspecial/broken-test.g.out
```diff
@@ -0,0 +1,50 @@
...
+ ); at GAPROOT/lib/test.gi:124 called from
+ParseTestInput( full, opts.ignoreComments 
+ ) at GAPROOT/lib/test.gi:558 called from
+<function "Test">( <arguments> )
+ called from read-eval loop at *stdin*:2
+type 'quit;' to quit to outer loop
+brk> quit;
+gap> Test("broken-test-3.tst");
+Error, Invalid data at line 5 of test file at GAPROOT/lib/test.gi:24 called from
+testError( "Invalid data" ); at GAPROOT/lib/test.gi:130 called from
```
fingolfin: Apparently this line causes a diff; perhaps you need to run the regenerate script once again?

lib/test.gi
```diff
@@ -12,14 +12,17 @@
...
+  local lines, inp, pos, outp, ign, commands, i, skipstate, checkifelseendif, foundcmd, testError;
   lines := SplitString(str, "\n", "");
   inp := [];
   pos := [];
   outp := [];
   ign := [];
   commands := [];
   i := 1;
+  testError := function(s)
+    ErrorNoReturn(Concatenation(s, " at line ", String(i)), " of test file");
```
fingolfin:
```suggestion
    ErrorNoReturn(s, " at line ", i, " of test file");
```

fingolfin: This is still a bit unsatisfactory, it'd be nice if the name of the test file also appeared. Could we simply pass `fnam` to `ParseTestInput`, and then do something similar to what `Test` does, say,
```gap
             if IsStream(fnam) then
               Print(s, " in test stream, line ", line, ":\n");
             else
               Print(s, "in ", fnam, ":", line, "\n");
             fi;
```

tst/testspecial/broken-test.g.out
ChrisJefferson: Something weird is going on, I am getting the test working on my machine. I tried making the line length longer and that didn't fix it either... investigating

ChrisJefferson: I have no idea why this isn't working. I checked the screen width isn't getting set too small. Any suggestions welcome.

fingolfin:
That's odd. If I take your PR, remove your `TEMPCOMMITFORTESTING` commit, and run `make && cd tst/testspecial && GAPDIR= ./regenerate_tests.sh` I do get a diff matching what we see on Travis:
```patch
diff --git a/tst/testspecial/broken-test.g.out b/tst/testspecial/broken-test.g.out
index b624c7853..2f3f723a3 100644
--- a/tst/testspecial/broken-test.g.out
+++ b/tst/testspecial/broken-test.g.out
@@ -20,7 +20,8 @@ type 'quit;' to quit to outer loop
 brk> quit;
 gap> Test("broken-test-3.tst");
 Error, Invalid data at broken-test-3.tst:5 at GAPROOT/lib/test.gi:27 called from
-testError( "Invalid data" ); at GAPROOT/lib/test.gi:134 called from
+testError( "Invalid data" 
+ ); at GAPROOT/lib/test.gi:134 called from
 ParseTestInput( full, opts.ignoreComments, fnam 
  ) at GAPROOT/lib/test.gi:562 called from
 <function "Test">( <arguments> )
```

ChrisJefferson:
Now, hopefully, all fine and tests passing.

tst/testspecial/broken-test.g.out
```diff
@@ -0,0 +1,54 @@
+gap> Test("broken-test-1.tst", rec(width := 800));
+Error, Cannot put a '> ' continuation in the middle of test output at broken-test-1.tst:3 at GAPROOT/lib/test.gi:27 called from
```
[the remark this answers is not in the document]
ChrisJefferson: I overhauled the errors in another way, to also make them more consistent with each other

PR #3071, opened 2018-11-29
kernel: fix string copying logic

rbehrends:
This pull request fixes two bugs in `src/stringobj.c`.

1. Calling `CopyToStringRep()` with a string whose length had been shortened by `SET_STR_LEN()` or which for some other reason leaves one or more words at the end unused could overwrite random memory, as `memcpy()` used `SIZE_OBJ()` for its length.
2. Part of the logic in `ImmutableString()` was inverted, copying the string if it was immutable rather than if it was mutable.

### DIFF
--- a/src/stringobj.c
+++ b/src/stringobj.c
@@ -1243,7 +1243,8 @@ Obj CopyToStringRep(
     copy = NEW_STRING(lenString);
 
     if ( IS_STRING_REP(string) ) {
-        memcpy(ADDR_OBJ(copy), CONST_ADDR_OBJ(string), SIZE_OBJ(string));
+        memcpy(CHARS_STRING(copy), CONST_CHARS_STRING(string),
+            GET_LEN_STRING(string));
         /* XXX no error checks? */
     } else {
         /* copy the string to the string representation                     */
@@ -1267,7 +1268,7 @@ Obj CopyToStringRep(
 */
 Obj ImmutableString(Obj string)
 {
-    if (!IS_STRING_REP(string) || !IS_MUTABLE_OBJ(string)) {
+    if (!IS_STRING_REP(string) || IS_MUTABLE_OBJ(string)) {
         string = CopyToStringRep(string);
         MakeImmutableString(string);
     }


src/stringobj.c
```diff
@@ -1243,7 +1243,8 @@ Obj CopyToStringRep(
     copy = NEW_STRING(lenString);
 
     if ( IS_STRING_REP(string) ) {
-        memcpy(ADDR_OBJ(copy), CONST_ADDR_OBJ(string), SIZE_OBJ(string));
+        memcpy(CHARS_STRING(copy), CONST_CHARS_STRING(string),
+            GET_LEN_STRING(string));
```
fingolfin: Good catch! This goes back to a255816 from Nov 2014, so it affects 4.10. On the up side, I imagine this problem cannot cause issues with GASMAN, only with Julia GC or Boehm, right? So it's not so bad (and less surprising that we never noticed).
Still, perhaps it should be backported to 4.10.

src/stringobj.c
```diff
@@ -1267,7 +1268,7 @@ Obj CopyToStringRep(
 */
 Obj ImmutableString(Obj string)
 {
-    if (!IS_STRING_REP(string) || !IS_MUTABLE_OBJ(string)) {
+    if (!IS_STRING_REP(string) || IS_MUTABLE_OBJ(string)) {
```
fingolfin: This one is much more recently, from 742f1a4 this September (and also my fault... :-/).  On the upside, it is not in GAP 4.10, so not much harm done :-).

src/stringobj.c
rbehrends: With GASMAN, the overflow should indeed generally only hit unused memory, except possibly in the case of 32-bit systems, where it can possibly overflow the reserved area.

I did track it down on HPCGAP with the Boehm GC, where the change to `ImmutableString()` triggered it reliably during testing.

This may also be one candidate for the spurious crashes that we have seen with the Julia GC.

src/stringobj.c
rbehrends: It is also not a critical issue. `ImmutableString()` is only used for the name field of functions and methods, meaning that in the worst case, that name was still mutable. Again, I ran into it during testing HPCGAP, where it led to names being in a thread-local region and thus inaccessible from other threads.

fingolfin:
Backported the relevant bits to stable-4.10 via 872d4937bac8a30529260fcab8b996a5cb614b0d

PR #3073, opened 2018-11-30
ci: fast fail AppVeyor and Travis outdated jobs

fingolfin:
This code is taken from resp. based on what Julia does.

The etc/travis_fastfail.sh script is run as part of Travis jobs, and
auto-cancels "outdated" jobs for all PR builds, or builds for branches other
than master and stable*.

The difference to Travis' builtin auto-cancel feature is that Travis only
cancels full builds, but not individual jobs. That is, if a job of a build
already started, Travis will not cancel this build, so all jobs of that build
are run. With this PR, all subsequent jobs of an "outdated" build auto-cancel
as they are started.

In the future, perhaps we can rewrite this in Python, to get rid of the
dependency on jq; and perhaps we can also adapt this to work with AppVeyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,7 +127,7 @@ matrix:
 
 script:
   - set -e
-  - gcov --version
+  - etc/travis_fastfail.sh
   - bash etc/ci-prepare.sh
   - bash etc/ci.sh
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -33,6 +33,11 @@ install:
 
 # scripts that run after cloning repository
 build_script:
+# If there's a newer build queued for the same PR, cancel this one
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -128,6 +128,7 @@ matrix:
 script:
   - set -e
   - gcov --version
+  - etc/travis_fastfail.sh
```
markuspf: nitpicky: could go before the `gcov --version`

fingolfin: True, but I actually placed it there because I thought sometimes the extra info might come in handy. But actually, I think the `gcov --version` should just be moved into `etc/ci-prepare.sh`, where we already have some other debug output. I'll do that now.

PR #3076, opened 2018-12-02
Add `IsAutoGlobal` for testing whether a variable was declared using `DeclareAutoreadableVariables`

ChrisJefferson:
At the moment there is no way to find if a variable is an "auto" (which means a function will be run when you access it). This is a problem in the (rare) cases where one wants to iterate over every variable in GAP, as just reading the value of an auto variable can cause lots of code to run.

Fixes (hopefully) a bug in the SYNTAX_TREE test

### DIFF
--- a/tst/testinstall/syntaxtree.tst
+++ b/tst/testinstall/syntaxtree.tst
@@ -16,7 +16,7 @@ Error, SYNTAX_TREE: <function> must be a plain GAP function (not a function)
 # Just try compiling all functions we can find in the workspace
 # to see nothing crashes.
 gap> for n in NamesGVars() do
->        if IsBoundGlobal(n) then
+>        if IsBoundGlobal(n) and not IsAutoGlobal(n) then
 >            v := ValueGlobal(n);
 >            if IsFunction(v) and not IsKernelFunction(v) then
 >                SYNTAX_TREE(v);


lib/global.gd
```diff
@@ -72,13 +72,30 @@ DeclareGlobalFunction("ValueGlobal");
...
 
+#############################################################################
+##
+#F  IsAutoGlobal( <name> )  . . . . . check if a global is automatic
+##
+##  <ManSection>
+##  <Func Name="IsAutoGlobal" Arg='name'/>
+##
+##  <Description>
+##  IsAutoGlobal ( <A>name</A> ) returns true if there is a global variable
```
fingolfin: I do realize you just copied the style used elsewhere in this file, but this inconsistent with what we do elsewhere. The following three alternatives come to mind (and we sadly use all of them in some place or another; perhaps we ought to come up with some style guidelines for our docs, and try to enforce them globally?)
```suggestion
##  returns true if there is a global variable
```
or
```suggestion
##  Returns true if there is a global variable
```
or 
```suggestion
##  <Ref Func="IsAutoGlobal"/> returns true if there is a global variable
```

lib/global.gi
```diff
@@ -97,6 +97,21 @@ InstallGlobalFunction( IsBoundGlobal,
...
+#############################################################################
+##
+#M  IsAutoGlobal ( <name> )  . . . .  check if a global is automatic
+## 
+
+
+InstallGlobalFunction( IsAutoGlobal, 
+        function (name)
+    local isauto;
+    CheckGlobalName( name );
```
fingolfin: Is there are benefit to calling `CheckGlobalName` here?

lib/global.gi
```diff
@@ -97,6 +97,21 @@ InstallGlobalFunction( IsBoundGlobal,
...
+    isauto := IS_AUTO_GVAR(name);
+    Info( InfoGlobal, 3, 
+          "IsAutoGlobal: called for ", name, " returned ", isauto);
```
fingolfin: What's the motivation for this `Info`? Only to imitate what's done elsewhere in this file? Is there any scenario this might be useful?

src/gvars.c
```diff
@@ -1085,6 +1085,21 @@ Obj FuncISB_GVAR (
 #endif
 }
 
+/****************************************************************************
+**
+*F  FuncIS_AUTO_GVAR( <self>, <gvar> )  . . check if a global variable is auto
```
fingolfin:
```suggestion
*F  FuncIS_AUTO_GVAR( <self>, <gvar> ) . . check if a global variable is auto
```

src/gvars.c
```diff
@@ -1085,6 +1085,21 @@ Obj FuncISB_GVAR (
...
+*/
+
+Obj FuncIS_AUTO_GVAR (
+    Obj                 self,
+    Obj                 gvar )
+{
+    // check the argument
+    RequireStringRep("IS_AUTO_GVAR", gvar);
+
+    return ExprGVar(GVarName( CONST_CSTR_STRING(gvar) ) ) ? True : False;
```
fingolfin: This will misreported TLS vars in HPC-GAP, which are also using `ExprGVar`, putting an `INTOBJ` into it. So more like this:
```C
  Obj expr = ExprGVar(GVarName( CONST_CSTR_STRING(gvar) ) );
  return (expr && !IS_INTOBJ(expr)) ? True : False;
```

lib/global.gi
ChrisJefferson: For consistency with IsBoundGlobal

lib/global.gi
ChrisJefferson: Yes, for consistency with IsBoundGlobal, and others. we could consistently simplify this whole file of course.

lib/global.gd
ChrisJefferson: I'm not opposed to these changes, but I would also prefer if we did try to keep the various `*Global` functions consistent with each other, so change them all, or none of them.

src/gvars.c
ChrisJefferson: ooo, good catch. Will fix.

ChrisJefferson:
I've fixed the bug and added a test for hpcgap (thanks @fingolfin). I'd prefer to not change the docs here, but make larger consistent changes in another PR.

lib/global.gd
fingolfin: OK, then let's do those with another PR someday in the future.

PR #3077, opened 2018-12-02
Allow appending to the command line history

ChrisJefferson:
Also, some minor cleanups to reading command line history, including correctly setting 'Pos' after the history has been read.

It is possible no-one other than me will care about this. I have been using this patch myself for a while, as I find attach read command line history to the end more useful, as I tend to read command line history because I want to access it!

### DIFF
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -135,7 +135,7 @@ GAPInfo.UseReadline := true;
 ##  <ManSection >
 ##  <Func Arg="[fname], [app]" Name="SaveCommandLineHistory" />
 ##  <Returns><K>fail</K> or number of saved lines</Returns>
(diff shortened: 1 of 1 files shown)


fingolfin:
I don't understand what difference this makes. Could you explain with an example?

ChrisJefferson:
Assume your current history is:

```
f(1);
g(1);
```

And you load the file:

```
x := 3;
```

Then the original code treats the read commands as if they were the oldest things typed:

```
x := 3;
f(1);
g(1);
```

And the new option treats the read commands as if they were the newest things typed:

```
f(1);
g(1);
x := 3;
```

In practice, the most important detail is that if your history is already full, then `ReadCommandLineHistory(file)` doesn't do anything, because it tries to put the history in the oldest position, but the history is already full. The new option will throw away existing history to store the stuff read from the file.

PR #3079, opened 2018-12-03
Fix NumberFFVector documentation, add an example

fingolfin:
The documentation of NumberFFVector made it sound as if a value from 1 to q^n is returned, but it really returns a value from 0 to q^n-1.

Also fix another unrelated typo in listcoef.gd.

This could be backported.


fingolfin:
Backported via 782bf1816ba2da398e530fc11e7cb723943492ac

Issue #1580, opened 2017-08-13
No method found when trying to compute conjugacy classes of finitely presented group


markuspf:
This happens on today's `master` and on `stable-4.8`:

```
gap> f := FreeGroup("x", "y");;
gap> g := f / ParseRelators(GeneratorsOfGroup(f), "x^2,y^2,(x*y)^8");
<fp group on the generators [ x, y ]>
gap> ConjugacyClasses(g);
Error, no method found! For debugging hints type ?Recovery from NoMethod\
Found
Error, no 3rd choice method found for `ConjugacyClasses' on 1 arguments at /home/makx/dev/gap/lib/methsel2.g:241 called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
 called from read-eval loop at *stdin*:7
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> 
gap> Size(g);
16
gap> ConjugacyClasses(g);
[ <identity ...>^G, x^G, y^G, (y*x)^2^G, (x*y)^4^G, x*y^G, (y*x)^3^G ]
gap> 
```

Obviously knowing that `G` is finite triggers the correct method to be selected, and I am not saying that this is wrong behaviour, it just tripped up a user just now.

fingolfin:
I guess we could add a very low ranked (last resort) method, which just consists of something like this:
```gap
  Error("no method found; if you think this group is finite, try again after calling `IsFinite` on it");
```
Alternatively, we could of course also run `IsFinite` as the very last resort (possibly using `RedispatchOnCondition`).

Of course there are then tons of similar methods with similar "problems"... so I am not sure the latter is a good idea: There is only so far the automatism in GAP will ever get you, at some point you'll have to learn the nitty gritty details of how things in computational group theory work...

Perhaps @hulpke has some thoughts on this, too?

fingolfin:
Also, it's not quite clear to me what this "issue" is. I assumed this is a request for enhancement, asking for a more helpful behaviour? But the issue text does not state so.

alex-konovalov:
An error message like that would be helpful, but I am afraid that in this case there will be too many methods to add. Indeed, if we add it for `ConjugacyClasses`, why not to add it to other group-theoretic operations?

[...]

Issue #596, opened 2016-02-07
Fp-group handling


hungaborhorvath:
I played around with fp-groups a bit, and found some weird behaviour in GAP. I am not sure if these are bugs or neccessities due to the presentation of the group.

I played with the following group: 
````
gap> F := FreeGroup("x", "y", "z");;
gap> x := F.1;; y := F.2;; z := F.3;;
gap> G := F/[x*y*x^(-1)*y^(-1), x*z*x^(-1)*z^(-1), z*y*z^(-1)*y^(-1), (x*y)^180, (x*y^5)^168];;
````
First of all, ````IsAbelian(G)```` did not give me any results in a reasonable time (couple minutes). If I invert the relations then GAP immediately recognizes them as commutators, and ````IsAbelian(G)```` is instantenous. 

Second, ````AbelianInvariants```` is instantenous. Then, however, ````IndependentGeneratorsOfAbelianGroup```` takes again an immensely long time. Without having ````IsAbelian```` set to true, it seems that it is stuck with trying to check some properties to determine which method to use. If I set ````IsAbelian```` to true, then it runs ````lib/grpfp.gi:6041```` for again a very long time (85 seconds). 

I checked this method and ran all but the last two lines of it. These all ran instantenous. The last two lines are 
````
  base := List( base, gen -> MappedWord( gen, gens, GeneratorsOfGroup( G ) ) );
  return base;
````
The interesting thing is, that if I simply ask GAP to print  ````base````, it already takes an enormous time (around 40 seconds). 

If I TraceMethod it, then this is what uses the most time: 
````
#I  Setter(IndependentGeneratorsOfAbelianGroup): system setter
````

This seems to mean that the algorithm itself is fine, the problem seems to lie in how the elements of an fp-group are represented for GAP. 

Is that really the case and can this be helped in any way?

hungaborhorvath:
I tried this experiment with ````CyclicGroup(IsFpGroup, Factorial(100));````

Again all but the last two lines ran instantenous, then when I ask for ````base```` I get this error: 
````
gap> base; 
Error, Range: <first> must be an integer less than 2^60 (not a integer (>= 2^60)) in
  [ for i  in [ n, n - 1 .. 1 ]  do
        list[i] := obj;
    od; at /home/ghorvath/work/gap/lib/list.gi:3541 called from 
ListWithIdenticalEntries( n, g 
     ) at /home/ghorvath/work/gap/lib/wordrep.gi:1329 called from
[...]

PR #3080, opened 2018-12-03
Add HallSubgroup fallback method

fingolfin:
This allows computing Hall subgroups for arbitrary finite groups
via an isomorphism into a permutation group.

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -3210,6 +3210,22 @@ InstallMethod( HallSubgroupOp,
 
 #############################################################################
 ##
+#M  HallSubgroupOp( <G>, <pi> ) . . . . . . . . . . . . .  for a finite group
+##
+InstallMethod( HallSubgroupOp,
+    "fallback method for a finite group",
+    [ IsGroup and IsFinite, IsList ],
+    function( G, pi )
+    local iso, H;
+
+    iso := IsomorphismPermGroup( G );
+    H := HallSubgroup( ImagesSource( iso ), pi );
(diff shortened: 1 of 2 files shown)


fingolfin:
I don't quite see why adjusting the rank by -1 would be useful, it seems rather arbitrary to me. After all, why would any future "generic" method for finite groups necessarily be better than this one? Instead, I'd think that any such method would need to be evaluated against the one added in this PR; and then the two should either be merged, using a heuristic to choose which approach to use; or if the new one is strictly superior to this method here, then the new should simply replace this one. And if a package wants to add a better method, they can adjust the rank by +1 instead.

(That said, I won't resist if people really think it's necessary to adjust the rank by -1, it just seems pointless from my vantage point).

As to other operations being affected: sure, you are of course absolutely right. See also issues #1580 and #596.

As discussed on PR #1580, instead of adding this method, we could certainly also try to give the user a more helpful method other than "method not found", something like "... perhaps try converting your group to another representation first".  It just is note quite clear where to add that.

hulpke:
@fingolfin 
Ranking at -1 would be an indication (both in looking at the code, and in execution preference) that this is a method which is there as convenience to users (to make the command work in every representation) but makes no claims of being effective. (For example, for matrix groups, if the matgrp package is loaded and `FittingFreeLiftSetup` has been called on the group, I would expect that method to be better.)
Of course, if you have examined the problem and feel confident in that this method is as good as generic can be, then add a comment about this and rank it at 0.

PR #3081, opened 2018-12-03
Remove IS_NSORT / IsNSortedList

fingolfin:
They never really supported or used by anything. And it is unclear how well
they worked in practice. On the other hand, one can always check whether
`IsSSortedList` returns false; for internal lists which store whether they
are strictly sorted in their TNUM, this is already handled efficiently.


markuspf:
Maybe the commit message should reflect that this was never supported/used?

fingolfin:
@markuspf done

PR #3082, opened 2018-12-04
kernel: convert tabs to spaces

fingolfin:
Also remove some trailing spaces.

I keep stumbling over these. I know we don't want to do large scale reformatting, but this is IMHO different.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -527,7 +527,7 @@ Obj ElmsBlist (
                     bit = 1;
                 }
             }
-	}
+        }
     }
     /* return the result                                                   */
     return elms;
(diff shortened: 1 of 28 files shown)


ChrisJefferson:
I see why these might annoy you, but I'm not enthusiastic about so much churn. Of course fixing these when one changes the relevant lines anyway is fine.

fingolfin:
This completely removes all tabs in the kernel (and thus completes the "tab-to-space" conversion work on the kernel I started in 2012 but regrettably did not finish back then).

ChrisJefferson:
I looked over my personal branches, and this conflicted with a few, but nothing too hard to fix. I do worry we seem to keep churning bits of code, some of which require small fixes to keep up to date with.

PR #3085, opened 2018-12-04
kernel: cleanup src/integer.{c,h}

fingolfin:
* tweak the `RequireInt` macro to automatically deduce the argument name
* rename function arguments from gmpL -> opL, gmpR -> opR, gmp -> op
  especially in user visible places
* remove other gratuitous references to "GMP integer"
* sync comments in integer.h and integer.c

... and then also did some `git clang-format`ing of the changed lines, to remove various whitespace etc. -- all of those can be ignored by viewing this PR with the "Hide whitespace changes" option enabled.

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -133,8 +133,6 @@ static Obj ObjInt_UIntInv( UInt i );
 
 #define SIZE_INT_OR_INTOBJ(obj) (IS_INTOBJ(obj) ? 1 : SIZE_INT(obj))
 
-#define REQUIRE_INT_ARG(funcname, argname, op) RequireInt(funcname, op, argname)
-
 #define RequireNonzero(funcname, op, argname) \
     do { \
       if (op == INTOBJ_INT(0) ) { \
@@ -154,9 +152,9 @@ static Obj IsIntFilt;
 
 /****************************************************************************
 **
-*F  TypeInt(<gmp>)  . . . . . . . . . . . . . . . . . . .  type of integer
+*F  TypeInt(<val>) . . . . . . . . . . . . . . . . . . . . .  type of integer
 **
-**  'TypeInt' returns the type of the integer <gmp>.
+**  'TypeInt' returns the type of the integer <val>.
 **
 **  'TypeInt' is the function in 'TypeObjFuncs' for integers.
 */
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
This patch seems to change many more lines than are required, I don't really see the benefit of renaming many of the arguments, teaming white space all over the place, etc.

src/integer.c
```diff
@@ -1746,23 +1738,23 @@ Obj ModInt(Obj opL, Obj opR)
...
     if ( !(i & (i-1)) ) {
       c = VAL_LIMB0(opL) & (i-1);
     }
-    
+
     /* otherwise use the gmp function to divide                            */
     else {
       c = mpn_mod_1( (mp_srcptr)CONST_ADDR_INT(opL), SIZE_INT(opL), i );
     }
-
```
fingolfin: I will drop these "remove trailing spaces" changes

src/integer.c
```diff
@@ -1112,114 +1108,115 @@ Obj FuncINT_STRING ( Obj self, Obj string )
...
+**  'EqInt' returns 1 if the two integer arguments <opL> and <opR> are equal
+**  and 0 otherwise.
 */
-Int EqInt ( Obj gmpL, Obj gmpR )
+Int EqInt(Obj opL, Obj opR)
 {
-  CHECK_INT(gmpL);
-  CHECK_INT(gmpR);
+    CHECK_INT(opL);
+    CHECK_INT(opR);
```
fingolfin: Note that here (and also in a few other functions), almost every single line was touched by the argument name change, hence it made sense to simply fully clang-format the function.

src/integer.c
```diff
@@ -1112,114 +1108,115 @@ Obj FuncINT_STRING ( Obj self, Obj string )
 
 /****************************************************************************
 **
-*F  EqInt( <gmpL>, <gmpR> ) . . . . . . . . .  test if two integers are equal
+*F  EqInt( <opL>, <opR> ) . . . . . . . . . .  test if two integers are equal
 **
-**  
-**  'EqInt' returns 1  if  the two integer   arguments <intL> and  <intR> are
-**  equal and 0 otherwise.
+**  'EqInt' returns 1 if the two integer arguments <opL> and <opR> are equal
```
fingolfin: Here the old comments did not even make sense, as they referred to outdated argument names.

ChrisJefferson:
I looked over the history of this file, and agree it's mostly been changed fairly recently. If you want to rename, that's fine.

PR #3087, opened 2018-12-04
Remove MinimalNormalSubgroups method using conj. classes

### LINKED ISSUE #164: MinimalNormalSubgroups is slower than NormalSubgroups
While testing some ideas I came across some examples where MinimalNormalSubgroups is significantly slower than NormalSubgroups. For example, this is the case with the group of Rubik's cube (3x3x3): 

cube := Group( ( 1, 3, 8, 6)( 2, 5, 7, 4)( 9,33,25,17)(10,34,26,18)(11,35,27,19), ( 9,11,16,14)(10,13,15,12)( 1,17,41,40)( 4,20,44,37)( 6,22,46,35), (17,19,24,22)(18,21,23,20)( 6,25,43,16)( 7,28,42,13)( 8,30,41,11), (25,27,32,30)(26,29,31,28)( 3,38,43,19)( 5,36,45,21)( 8,33,48,24), (33,35,40,38)(34,37,39,36)( 3, 9,46,32)( 2,12,47,29)( 1,14,48,27), (41,43,48,46)(42,45,47,44)(14,22,30,38)(15,23,31,39)(16,24,32,40) );

`NormalSubgroups` computes in 12 seconds on my laptop (with CRISP), and afterwards `MinimalNormalSubgroups` computes immediately. But if I call `MinimalNormalSubgroups` without calling `NormalSubgroups` first, then it computed for minutes before I decided to shut it down. However, `Socle` gave an answer in 13 seconds. (And it happens to be Abelian, but note that cube is non-solvable.)

A similar phenomenon occurred when trying it for the direct product of S10 with itself:  `MinimalNormalSubgroups` takes 26 seconds, `NormalSubgroups` computes in 2.5 seconds (and afterwards `MinimalNormalSubgroups` is instantaneous). Interestingly, `Socle` gave an answer in 170 ms, which is significantly faster than any of the above. 

Is it possible to increase efficiency for `MinimalNormalSubgroups`? It would help factoring groups into direct products. (E.g. the number of direct factors is bounded by the number of minimal normal subgroups <=> number of direct factors of socle.) Of course, as an alternative, I can always call Socle for solvable groups and then simply decompose it into direct factors, but this may not work very well for nonsolvable groups.

fingolfin:
In all examples I tried, this was inferior to the method using NormalSubgroups.

But perhaps @hulpke or somebody else knows about examples where the old method might be (substantially) better?

Fixes #164


ChrisJefferson:
I'm happy with this PR from a correctness point of view, don't know about performance.

Issue #2703, opened 2018-08-15
Call GroupByGenerators/GroupWithGenerators with a collection?


ThomasBreuer:
@alex-konovalov had noticed a test failure,
see https://github.com/gap-system/gap/issues/1619#issuecomment-412834301.

This failure is a side-effect of the changes from pull request #2522.

The operations `GroupByGenerators` and `GroupWithGenerators` are documented
for a *list* of generators.
However, their methods accept a *collection* as the first argument,
and call `AsList` in order to create a list from it if necessary.
With the abovementioned changes, a new function `MakeGroupyType` was introduced
that assumes the generators to be a list, and this is where the error happens.

Thus the following works in GAP 4.9 but not in the master branch.
```
GroupByGenerators( Group( (1,2) ) )
```

I see several possibilities to fix this problem.

1. Change the method installations for `GroupByGenerators` and `GroupWithGenerators`
   such that only a list & collection or an empty list is accepted as the first argument.
   This fits to the documentation,
   the problem is that existing code may be broken that uses the undocumented feature.

2. Change the method installations for `GroupByGenerators` and `GroupWithGenerators`
   such that a non-list collection given as the first argument gets replaced by a list in the beginning.

3. Proceed like in 1. but add new methods requiring a collection as the first argument and then
   delegating to the methods for lists;
   add comments to these methods that they might become obsolete at some time.

What do others think?

markuspf:
Is calling `AsList` on a group a concern?

A user might use `GroupByGenerators( biggroup )`, and then wonder why its slow.

ThomasBreuer:
@markuspf First we should decide whether we want to admit a non-list collection as an argument at all.
If yes then we might think about how to turn a given non-list collection into a correct list of generators.
For the case of `GroupWithGenerators`, I think there is no choice,
but `GroupByGenerators` could of course do something more sensible.

markuspf:
I think we shold allow collections (that are not lists), because code like above  ( `GroupByGenerators( Group( (1,2) ) )`) looks like a fairly natural thing to do.

Now, converting just any collection to a list throws up my issue with turning things into lists that really shouldn't be; what collections would people want to use to generate groups from?

ThomasBreuer:
@markuspf If we want to allow collections then I would add a special `GroupByGenerators` treatment
[...]

Issue #1619, opened 2017-08-23
CTblLib: make sure that tests pass


alex-konovalov:
Below is the output of CTblLib tests with GAP 4.8.8 release candidate when GAP is loaded with default packages. It will be nice to clean up the diffs, which are of several kinds:

* different formatting

* different output (caused by SpinSym package?)

* tests can be performed because of TERM = "dumb" (cf. #1617)

In addition, it's not possible to detect whether tests passes or not automatically. There is an easy fix for that: `PackageInfo.g` specifies the testfile `tst/testauto.g`, but that testfile only calls one test `tst/docxpl.tst`. In this case, one could just specify `tst/docxpl.tst` in `PackageInfo.g` instead of `tst/testauto.g`. Otherwise, for the compliance with automated failure detection one could follow #1617 or use `TestDirectory`.

```
============================OUTPUT START==============================

#I  RunPackageTests("ctbllib", "1.2.2", "tst/testauto.g", auto);
 ���������������������������   GAP 4.8.8, 20-Aug-2017, build of 2017-08-20 19:40:04 (BST)
 ���  GAP  ���   https://www.gap-system.org
 ���������������������������   Architecture: x86_64-pc-linux-gnu-gcc-default64
 Libs used:  gmp, readline
 Loaded workspace: wsp.g
 Components: trans 1.0, prim 2.1, small* 1.0, id* 1.0
 Packages:   AClib 1.2, Alnuth 3.0.0, AtlasRep 1.5.1, AutPGrp 1.8, 
             Browse 1.8.7, CRISP 1.4.4, Cryst 4.1.12, CrystCat 1.1.6, 
             CTblLib 1.2.2, FactInt 1.5.4, FGA 1.3.1, GAPDoc 1.6, IO 4.4.6, 
             IRREDSOL 1.4, LAGUNA 3.7.0, Polenta 1.3.7, Polycyclic 2.11, 
             RadiRoot 2.7, ResClasses 4.6.0, Sophus 1.23, SpinSym 1.5, 
             TomLib 1.2.6, Utils 0.46
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
########> Diff in /data/gap-jenkins/workspace/GAP-minor-release-test/GAPCOPTS/\
64build/GAPGMP/gmp/GAPTARGET/packages/label/kovacs/gap4r8/pkg/ctbllib/tst/docx\
pl.tst, line 34:
# Input is:
AllCharacterTableNames( Size, 120 );
[...]

PR #3091, opened 2018-12-05
call `GroupByGenerators`, `GroupWithGenerators` with domain

ThomasBreuer:
change the default methods for `GroupByGenerators`
and `GroupWithGenerators` such that the problem gets fixed
that had been introduced with pull request #2522
and discussed in issue #2703

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -4398,55 +4398,48 @@ end);
 InstallMethod( GroupWithGenerators,
     "generic method for collection",
     [ IsCollection ],
-function( gens )
-local G,typ;
+    function( gens )
+    local typ;
 
-  typ:=MakeGroupyType(FamilyObj(gens),
+    gens:= AsList( gens );
+    typ:= MakeGroupyType( FamilyObj( gens ),
           IsGroup and IsAttributeStoringRep 
           and HasIsEmpty and HasGeneratorsOfMagmaWithInverses,
           gens,false,true);
 
-  G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
-
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, gens );
 end );
 
 InstallMethod( GroupWithGenerators,
     "generic method for collection and identity element",
-    IsCollsElms, [ IsCollection, IsMultiplicativeElementWithInverse ],
-function( gens, id )
-local G,typ;
+    IsCollsElms,
+    [ IsCollection, IsMultiplicativeElementWithInverse ],
+    function( gens, id )
+    local typ;
 
-  typ:=MakeGroupyType(FamilyObj(gens),
+    gens:= AsList( gens );
+    typ:= MakeGroupyType( FamilyObj( gens ),
           IsGroup and IsAttributeStoringRep 
             and HasIsEmpty and HasGeneratorsOfMagmaWithInverses and HasOne,
             gens,id,true);
 
-  G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
-                          One,id);
-
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, gens, One, id );
 end );
 
-InstallMethod( GroupWithGenerators,"method for empty list and element",
-  [ IsList and IsEmpty, IsMultiplicativeElementWithInverse ],
-  function( empty, id )
-local G,fam,typ;
+InstallMethod( GroupWithGenerators,
+    "method for empty list and element",
+    [ IsList and IsEmpty, IsMultiplicativeElementWithInverse ],
+    function( empty, id )
+    local typ;
 
-  fam:= CollectionsFamily( FamilyObj( id ) );
+    typ:= NewType( CollectionsFamily( FamilyObj( id ) ),
+              IsGroup and IsAttributeStoringRep and
+              HasGeneratorsOfMagmaWithInverses and HasOne and IsTrivial );
 
-  typ:=IsGroup and IsAttributeStoringRep
-        and HasGeneratorsOfMagmaWithInverses and HasOne and IsTrivial;
-  typ:=NewType(fam,typ);
-
-  G:= rec();
-  ObjectifyWithAttributes( G, typ,
-                            GeneratorsOfMagmaWithInverses, empty,
-                            One, id );
-
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, empty, One, id );
 end );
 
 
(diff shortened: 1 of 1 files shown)


lib/grp.gi
```diff
@@ -4456,14 +4449,45 @@ end );
...
-    "delegate to `GroupWithGenerators'",
+    "replace domain by suitable list, and call `GroupWithGenerators'",
     [ IsCollection ],
+    function( gens )
+    if HasGeneratorsOfMagmaWithInverses( gens ) then
+      gens:= GeneratorsOfMagmaWithInverses( gens );
+    elif HasGeneratorsOfMagmaWithOne( gens ) then
+      gens:= GeneratorsOfMagmaWithOne( gens );
+    elif HasGeneratorsOfMagma( gens ) then
+      gens:= GeneratorsOfMagma( gens );
```
fingolfin: Hmm. If we do this for GroupByGenerators, shouldn't we also add similar methods to all other `*ByGenerators*` functions? But there are dozens of these. So I am rather inclined to go the other direction, and say we shouldn't do this at all...

I found these *ByGenerator operations:
```
AdditiveGroupByGenerators
AdditiveMagmaByGenerators
AdditiveMagmaWithInversesByGenerators
AdditiveMagmaWithZeroByGenerators
AlgebraByGenerators
AlgebraWithOneByGenerators
BiAlgebraModuleByGenerators
DefaultFieldByGenerators
DefaultRingByGenerators
DivisionRingByGenerators
DomainByGenerators
FLMLORByGenerators
FLMLORWithOneByGenerators
FieldByGenerators
FieldOverItselfByGenerators
GroupByGenerators
IdealByGenerators
InducedPcgsByGenerators
InducedPcgsByGeneratorsNC
InverseMonoidByGenerators
InverseSemigroupByGenerators
IsomorphismFpGroupByGenerators
IsomorphismFpGroupByGeneratorsNC
LeftAlgebraModuleByGenerators
LeftIdealByGenerators
LeftMagmaIdealByGenerators
LeftModuleByGenerators
MagmaByGenerators
MagmaIdealByGenerators
MagmaWithInversesByGenerators
MagmaWithOneByGenerators
MonoidByGenerators
NaturalHomomorphismByGenerators
NearAdditiveGroupByGenerators
NearAdditiveMagmaByGenerators
NearAdditiveMagmaWithInversesByGenerators
NearAdditiveMagmaWithZeroByGenerators
OrbitStabilizerListByGenerators
RepresentativeTomByGenerators
RepresentativeTomByGeneratorsNC
RightAlgebraModuleByGenerators
RightIdealByGenerators
RightMagmaIdealByGenerators
RingByGenerators
RingWithOneByGenerators
SemigroupByGenerators
SemigroupIdealByGenerators
SemiringByGenerators
SemiringWithOneAndZeroByGenerators
SemiringWithOneByGenerators
SemiringWithZeroByGenerators
TwoSidedIdealByGenerators
```

lib/grp.gi
```diff
@@ -4398,55 +4398,48 @@ end);
...
           and HasIsEmpty and HasGeneratorsOfMagmaWithInverses,
           gens,false,true);
 
-  G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
-
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, gens );
 end );
```
fingolfin: This completely changes formatting, but (as far as I could tell) really only inserts a single, line, `gens:= AsList( gens );`, and then replaces one `AsList(gens)` later by `gens`.

I'd really prefer if this changes was kept to the minimum, for ease of reviewing.

lib/grp.gi
```diff
@@ -4398,55 +4398,48 @@ end);
...
             gens,id,true);
 
-  G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
-                          One,id);
-
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, gens, One, id );
 end );
```
fingolfin: Same as above.

lib/grp.gi
```diff
@@ -4398,55 +4398,48 @@ end);
...
-                            GeneratorsOfMagmaWithInverses, empty,
-                            One, id );
+    typ:= NewType( CollectionsFamily( FamilyObj( id ) ),
+              IsGroup and IsAttributeStoringRep and
+              HasGeneratorsOfMagmaWithInverses and HasOne and IsTrivial );
 
-  return G;
+    return ObjectifyWithAttributes( rec(), typ,
+               GeneratorsOfMagmaWithInverses, empty, One, id );
 end );
```
fingolfin: I am not even sure what has changed in this function. Was anything changed, other than rearranging the code and reformatting it?

lib/grp.gi
ThomasBreuer: Perhaps the discussion of issue #2703 was not explicit enough:
The point is just to fix a side-effect (which I would call a bug) of pull request #2522,
which breaks functionality in GAP 4.10 that was available in earlier versions.

lib/grp.gi
ThomasBreuer: Note that `gens` is used by `MakeGroupyType`.
This is where the error happens in the old version.

alex-konovalov:
I am thinking whether it will be easier and more consistent to adapt CTblLib to call `GroupByGenerators(GeneratorsOfGroup(...))`. 

First of all, as we can see from the list collected by @fingolfin, doing this consistently would require changing more methods. 

Second, while it seemed a good idea to me initially, it's not any more. The meaning of the names `GroupByGenerators` and `GroupByGenerators` seems intuitively clear to me as returning the structure whose generators are given as its argument. Following this logic, if it's called with a domain, then its list of generators is the list of *all* elements of the domain. Doing some reduction "under the hood" (and not in all "StructureByGenerators" cases) may be more confusing IMHO.

ThomasBreuer:
@alex-konovalov I disagree completely.
Pull request #2522 has broken functionality of GAP.
According to your comments about issue #2703, you were aware of this fact,
and still you have released GAP 4.10.
When GAP users upgrade to version 4.10 (which fortunately happens slowly)
then they will run into errors, due to the broken functionality.
I think that the bug simply has to be fixed,
independent of considerations how a better world could look like.

ThomasBreuer:
@alex-konovalov I disagree completely.
- When you refer to the documentation then you just cite the text;
  however, both the operation declarations and the method installations state
  that the first argument is expected to be a list or collection.
- I have deliberately not proposed a change of the documentation,
  because I think that the discussion is not yet finished
  concerning what we want in the long run.
- You give the impression as if the changed behavior in GAP 4.10 was intended.
  This is of course wrong, as one can see from the code and the discussion.

lib/grp.gi
fingolfin: But the old code never did something like this. It just took the set of elements of the given domain / collection as generators.

alex-konovalov:
@ThomasBreuer I agree that both of us may disagree. I for example disagree with calling this a bug in GAP - IMHO it's a side effect of another change, which happen to reveal some inefficiencies in some other code.

Nevertheless, I agree to go for #3095 which has lesser scope than this PR, just restoring the previous behaviour, and still keeping it undocumented. I hope that this will fix several diffs in CTblLib package for now, but not all of them - can I just remind you since you may not have seen updates in https://github.com/gap-system/gap/issues/1619 that a new release of CTblLib with cleanly passing tests (as well as a new release of AtlasRep) will be most welcomed!

I intend to add a warning (for master branch, to appear in GAP 4.11) to be displayed when GroupWithGenerators is called on something which is not a list - if that will be accepted, it will cause warnings for CTblLib tests in their current state.

PR #3092, opened 2018-12-05
Add new copyright header to kernel and library sources

fingolfin:
The current copyright header has various problems, which I am trying to
overcome. Here's the notable ideas behind the ones suggested here:

* don't mention the filename: doing so serves no purpose, but can get out of sync with the actual filename
* don't mention the "author" of a file -- again, this information quickly gets outdated (and is severely outdated in many places already); the git repository resp. the old Mercurial and CVS repositories, are more useful sources for that kind of information
* don't refer to "the GAP Group", and esp. don't claim copyright for it: this is a fictitious entity at best (and thus assigning copyright to it is meaningless); at worst, there is a potential for conflict with actual existing legal entities with that name
* don't hardcode copyright dates, which never get updated (resp. require pointless updates once a year)
* instead, move all copyright statements into a new file COPYRIGHT, and only references that from file headers
* however, do explicitly state the license of the code (GPL2 or later) in the source file header
* optionally, the header can be followed by a description of the content of the file, as we already do right now.
* drop legal headers from .tst files


lib/error.g
```diff
@@ -1,9 +1,24 @@
 #############################################################################
 ##
-#W  error.g                    GAP library                 Steve Linton
+##  GAP, a computer algebra system
```
alex-konovalov: I'd prefer replacing `computer` by `computational`.

fingolfin: Fine by me. But perhaps it should instead be "system for computational discrete algebra "?

On the website, we say:
> GAP - Groups, Algorithms, Programming - a System for Computational Discrete Algebra

We also say this on the website and in the README:
> GAP is a system for computational discrete algebra

The reference manual says:
> GAP is a free, open and extensible software package for computation in discrete abstract algebra

alex-konovalov: any of the three fine w/me, sure.

fingolfin:
We might also want to include a [SPDX](https://spdx.org/) tag, like `# SPDX-License-Identifier: GPL-2.0-or-later`. I see no harm in it, and it might be useful to some people

fingolfin:
Updates:
* added the `COPYRIGHT` file with a list of authors, based on the (extended/private) git repository history
* replaced "This program" by "GAP" in comments
* added SPDX identifier
* added alternate shorter license headers which omit the three paragraphs of GPL legalese, based on the idea that (a) people prefer shorter headers, and (b) the SPDX identifier should be sufficient. (Arguably, this is slightly less "safe", legally speaking, but then again we had no GPL info *at all* in most files before, so it's still an improvement.)

lib/error.g
```diff
@@ -1,9 +1,25 @@
...
+##  option) any later version.
+##
+##  GAP is distributed in the hope that it will be useful, but WITHOUT
+##  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+##  or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+##  License for more details.
+##
+##  You should have received a copy of the GNU General Public License
+##  along with GAP; if not, write to the Free Software Foundation, Inc.,
+##  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
```
fingolfin: This is the long header, 22 lines

lib/init.g
```diff
@@ -1,13 +1,11 @@
...
-#W                                                          & Martin Schönert
+##  This file is part of GAP, a system for computational discrete algebra
 ##
+##  SPDX-License-Identifier: GPL-2.0-or-later
 ##
-#Y  Copyright (C)  1997,  Lehrstuhl D für Mathematik,  RWTH Aachen,  Germany
-#Y  (C) 1998 School Math and Comp. Sci., University of St Andrews, Scotland
-#Y  Copyright (C) 2002 The GAP Group
+##  Copyright of GAP belongs to its developers, whose names are too numerous
+##  to list here. Please refer to the COPYRIGHT file for details.
```
fingolfin: This is the short header variant: only 8 lines (instead of 22)

src/hpc/tls.h
```diff
@@ -1,3 +1,27 @@
...
+//
+//  SPDX-License-Identifier: GPL-2.0-or-later
+//
+//  Copyright of GAP belongs to its developers, whose names are too numerous
+//  to list here. Please refer to the COPYRIGHT file for details.
+//
+//  GAP is free software; you can redistribute it and/or modify it under
+//  the terms of the GNU General Public License as published by the Free
+//  Software Foundation; either version 2 of the License, or (at your
+//  option) any later version.
```
fingolfin: For C/C++ files, we could also consider using inline comments `//` instead of multi-line comments. But really, it seems simpler (and more consistent) to keep the current formatting.

tst/testinstall/alghom.tst
```diff
@@ -1,12 +1,29 @@
...
 ##
-#Y  Copyright (C)  1998,  Lehrstuhl D für Mathematik,  RWTH Aachen,  Germany
+##  GAP is distributed in the hope that it will be useful, but WITHOUT
+##  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+##  or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+##  License for more details.
 ##
+##  You should have received a copy of the GNU General Public License
+##  along with GAP; if not, write to the Free Software Foundation, Inc.,
+##  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
```
fingolfin: We probably don't need copyright / license headers in `.tst` files, but they also won't hurt... I'd have a small preference for just removing them everywhere.

lib/init.g
ChrisJefferson: I'm a big fan of the shorter variant, unless anyone strongly feels it isn't complete enough. I don't think anyone could reasonable misunderstand it.

COPYRIGHT
```diff
@@ -0,0 +1,98 @@
+GAP, a system for computational discrete algebra
```
fingolfin:
```suggestion
GAP, a system for computational discrete algebra.
```

lib/init.g
fingolfin: I agree.

alex-konovalov:
Could you please check the list at https://www.gap-system.org/Contacts/People/authors.html which has names preceding the (extended/private) git repository history, and add them to the list? I hope then we will have a fairly complete list of authors.

fingolfin:
@alex-konovalov done

fingolfin:
I've now also scanned all library files for lines starting with `#W`, and extracted some additional author names from there, which are now also in the COPYRIGHT file.

ChrisJefferson:
While I'm not sure of the best way to keep it, I have found in the past the names at the top of files can be useful to have a first person to call on, when problems arise.

Other than the loss of names, I'm 100% happy with everything else.

COPYRIGHT
```diff
@@ -0,0 +1,124 @@
...
+* Soley Jonsdottir
+* Stefan Kohl
+* Steve Linton
+* Susanne Keitemeier
+* Thomas Bischops
+* Thomas Breuer
+* Thomas Merkwitz
+* Udo Polis
+* Ute Schiffer
+* VInay Wagh
```
wilfwilson:
```suggestion
* Vinay Wagh
```

alex-konovalov:
I am sympathetic to the "loss of names" remark by @ChrisJefferson, but I have a feeling that the further it goes, the less up-to-date that information is, because nobody updates that lists while making changes. OTOH digging in the git history may take slightly longer, but will give you more precise information.

COPYRIGHT
alex-konovalov: VInay -> Vinay

hulpke:
Let me add that I find the filename useful when editing multiple files in the same editor window.

wilfwilson:
I agree with the changes that you have made @fingolfin, except, as with Chris, my concern relates to the information contained in the names at the top of the files. I feel that by removing these names, there is some useful information being lost, and perhaps there should be a way of retaining that information (perhaps somewhere else).

The problem with relying on version control history to deduce authorship arises when someone's contributions pre-date GAP's use of version control. And it requires knowing how those tools work (I admit I have never got round to learning how to use git to do such things...)

I don't have a concrete suggestion for how this information could or should be stored better, however. I recognise that keeping that information up to date requires effort!

fingolfin:
I don't like the list of "authors" in the headers at all, for the following reasons:

* at least in the kernel, they are completely outdated in >90% of the files (more likely 100%)
* they give the wrong incentives, IMHO, establishing fiefdoms in the code where some code "belongs" to some people. I know that people actually *wanted* this in the past; but I think it is fundamentally opposite to the idea of "open development"
* *if* we keep "author" (maybe: "contributor" is the better term?) names in there, then IMHO we need to develop rules who can be aded there; i.e. how much of a file do I have to modify, and in which way, before I can be added? If I fix a bug leading to incorrect mathematical results or a crash; or if I speed up a computation 10fold by inserting a single character, does that qualify? When I add a single 20 line method to a file with 4000 lines, does that qualify? Who will write up these rules? Who will decide about them in case of disagreement?
* and also, who will make sure to keep those lists up-to-date?
* when I want to know who wrote code, I never look at the header, instead, I always look at the revision history -- but then, I have the full proper git history, will all author names in it, not the crippled version we made online (which not just dropped author names, but much worse, squashed all commits together into a bunch of "release commits"; and even that sadly was botched up, rendering that part of our repository unusable). Hence I'd much rather have us make that history public! With names of people who want credit and/or don't mind this being shown; while the names of all others would be changed to, say "anonymous".

alex-konovalov:
To remind, we have https://www.gap-system.org/Contacts/People/modules.html which was an attempt to keep an up-to-date summary of the contributions to the GAP library. How up to date is that?

mtorpey:
I agree that if we keep names at the tops of the files, there ought to be clearer guidelines for when names need to be added or removed.  I've modified files a number of times and been unsure what to do with the header.

I also agree that it'd be great to have the full version control history available publicly.  This would make modifying the library much easier for me, as I'd know who changed individual lines when, a big improvement on the current system of looking at the name at the top of the file and hoping it's not out of date.  Anonymising individual authors on request also seems reasonable, and would still allow the history to be useful.

fingolfin:
To be clear: I can of course add back those "author" names (although quite reluctantly), but I am quite serious when I say that then I'd like a discussion and a policy on how we decide who is listed there.

@hulpke wrote:

> The file headers are the one public place which I can point to (even to people who do not want to learn how to read git logs) as listing my past effort in contributing to the system.

I wonder: did that ever happen, or is it just hypothetical? Would somebody care more about this information than e.g. what is on https://www.gap-system.org/Contacts/People/modules.htm ?l

> (Besides, we removed names in commit messages when transitioning from CVS, so that is not really available in github any longer.)

Just to clarify: when we transitioned away from CVS, we retained *all* names; in fact I spent a great deal of effort to restore names based on signatures in commit messages. So they were all in Mercurial, and also in my initial conversion of that to git. Only when we made the repository public did we drop all these commits, essentially squashing them into one. But I still have the old repository, and using so-called "grafts", I am using it as part of the regular git repository. Super helpful.

> Let me add that I find the filename useful when editing multiple files in the same editor window.

Of course I respect that it *is* helpful for you to have those filenames, but purely out of curiosity, how does that work for you -- you have an editor that does not allow you to see the filename, yet it somehow allows you to the top of the file despite even when being scrolled to the middle of it? So it's not emacs, nor vim (`ctrl-g`), nor any GUI editor?

fingolfin:
Can we at least agree to drop the names from the kernel files? Because those really are mostly nonsense.

COPYRIGHT
```diff
@@ -0,0 +1,124 @@
...
+* Chris Jefferson
+* Chris Wensley
+* Christian Sievers
+* Christopher Russell
+* Claus Fieker
+* Colva M. Roney-Dougal
+* Craig A. Struble
+* David Joyner
+* Derek F. Holt
+* Dima Pasechnik
```
dimpase: you can remove Dima Pasechnik, and leave Dmitrii Pasechnik only there. 
Dima is a short form of Dmitrii, nothing more, and I am known under both of them.

wilfwilson:
I think I'd be happy with an updated/well-maintained/perhaps expanded version of https://www.gap-system.org/Contacts/People/modules.html to replace names at the top of files. Especially if this is linked to prominently, perhaps by including a link in the new header...

hulpke:
@fingolfin I have no opinion, nor reservation, about the kernel files.

COPYRIGHT
fingolfin: Oops, indeed, I overlooked it. Thanks, will be fixed

alex-konovalov:
For modules, PRs to https://github.com/gap-system/GapWWW/blob/master/Contacts/People/modules.mixer are welcome!

fingolfin:
If we preserve names in the library, how could that look? Here is a first suggestion, please feel free to suggest alternatives.
```gap
##
##  This file is part of GAP, a system for computational discrete algebra.
##  Contains code by Emma Smith, John Doe
##
##  Copyright of GAP belongs to its developers, whose names are too numerous
##  to list here. Please refer to the COPYRIGHT file for details.
##
##  SPDX-License-Identifier: GPL-2.0-or-later
```

hulpke:
> I wonder: did that ever happen, or is it just hypothetical? Would somebody care more about this information than e.g. what is on https://www.gap-system.org/Contacts/People/modules.html ?l

I have (when having to document impact of my work) made claims about authorship of significant parts of GAP and taken the library headers as being easily accessible support of this claim. 

> Just to clarify: when we transitioned away from CVS, we retained _all_ names; in fact I spent a great deal of effort to restore names based on signatures in commit messages. So they were all in Mercurial, and also in my initial conversion of that to git. Only when we made the repository public did we drop all these commits, essentially squashing them into one. But I still have the old repository, and using so-called "grafts", I am using it as part of the regular git repository. Super helpful.

Yes, but how likely is it that others will be able to get this information. (I would not know how to do it, even if the files are somewhere available to me.)

> Of course I respect that it _is_ helpful for you to have those filenames, but purely out of curiosity, how does that work for you -- you have an editor that does not allow you to see the filename, yet it somehow allows you to the top of the file despite even when being scrolled to the middle of it? So it's not emacs, nor vim (`ctrl-g`), nor any GUI editor?

it is the pathetic "go to line 1", but in part it is even at the start of editing when both files are open at line 1.

hulpke:
> If we preserve names in the library, how could that look? Here is a first suggestion, please feel free to suggest alternatives.
> 

> ##  This file is part of GAP, a system for computational discrete algebra.
> ##  Contains code by Emma Smith, John Doe

Fine with me, if you you prefer reformatting over keeping the existing lines.

fingolfin:
The problem with keeping the existing lines is that they also contain the filename, and often another bit like "GAP library" or "GAP Library"; also, the additional lines are not uniformly converted. So keeping the existing lines is not an option if one wants to get rid of the filenames. And even if one does not want that: then one still has to deal with the problem that one now has an even harder job of identifying which portion of the file contains "legal notices", and then to replace them via a script. So all in all, "keeping the existing lines" is not less work.

ChrisJefferson:
@ThomasBreuer : I think the intention is people should be reading the COPYRIGHT file in the git repostiory (or the copy of GAP they downloaded if they downloaded a release).

frankluebeck:
I have no objections to the proposed changes. The lists of names given in the files are indeed often quite outdated. Finding rules for these lists and maintaining them for hundreds of files is in my opinion not worth the effort.

I'm not happy with the repeated mentioning of additional information in non-public no longer used repositories. When the present public repository was created there was a long discussion about which information from former non-public repositories will be copied over (you can get the state of the repository for every historical release of GAP  4). Please, respect this decision.

wilfwilson:
> ```gap
> ##  This file is part of GAP, a system for computational discrete algebra.
> ##  Contains code by Emma Smith, John Doe
> ```
Fine. Another suggestion:
```
##  This file is part of GAP, a system for computational discrete algebra.
##  This file's authors include Emma Smith, John Doe
```

sebasguts:
From my point of view the changes are fine :)

ssiccha:
@frankluebeck this is rather a stupid question than a suggestion and has probably already been discussed: Would it be ok if the names were replaced by anonymous aliases? (I can't imagine that someone who doesn't have access of the old history anyways would go through the process of trying to match the aliases with the real names of the contributors, just to be able to point a finger at someone)

ssiccha:
And I'm fine with the changes too.

fingolfin:
Updated the PR to add back file author names to the library (lib, grp, hpcgap/lib, ...). Please check if it seems reasonable now.

lib/grplatt.gd
```diff
@@ -1,12 +1,12 @@
 #############################################################################
 ##
-#W  grplatt.gd                GAP library                   Martin Schönert,
-#W                                                          Alexander Hulpke
+##  This file is part of GAP, a system for computational discrete algebra.
+##  This files's authors include Martin Schönert,, Alexander Hulpke.
```
alex-konovalov: double comma

lib/grplatt.gi
```diff
@@ -1,12 +1,12 @@
 #############################################################################
 ##
-#W  grplatt.gi                GAP library                   Martin Schönert,
-#W                                                          Alexander Hulpke
+##  This file is part of GAP, a system for computational discrete algebra.
+##  This files's authors include Martin Schönert,, Alexander Hulpke.
```
alex-konovalov: double comma

grp/ree.gi
```diff
@@ -1,9 +1,13 @@
 #############################################################################
 ##
-#W  ree.gi                     GAP library
+##  This file is part of GAP, a system for computational discrete algebra.
+##  This files's authors include
```
alex-konovalov: empty list

ChrisJefferson:
I am now happy with this PR. We can have further discussions on the best way of handling authors of files.

hulpke:
I am happy with it now. Alexander

Stefan-Kohl:
I think having the names of the main authors listed on top of each file is useful information. While formalizing who gets mentioned or not is impractical, my guess is in practice the simple rule would work that everybody who has made a significant non-trivial / not purely janitorial contribution to a file should add their name to the list. So far it was basically that way, and I don't remember this has ever caused much controversy. If people remember where they have made significant contributions or look it up in the repository, then bringing the information up-to-date should be pretty easy.

As to copyright information, I'd certainly not go beyond a brief reference to the file with the license.

fingolfin:
Fixed the double-comma glitch @alex-konovalov reported; squashed all into one commit; and update the commit message for that to not claim anymore that author names are removed. 

If tests pass, I think this could be merged.

alex-konovalov:
It remains to fix one test because line numbers changed: https://travis-ci.org/gap-system/gap/jobs/466543837

PR #3093, opened 2018-12-05
`IsMatrixObj` objects as elements of GAP's matrix groups

ThomasBreuer:
GAP's default mechanism for dealing with small matrix groups
is to compute an isomorphic permutation group
via the action on orbits of vectors.

With the changes from this pull request,
this mechanism works for groups of objects in `IsMatrixObj`
that are not plain lists of lists.
Of course this is just one step, more changes in this area will follow.
(A preliminary test case is contained in the `GAPJulia` package.)

Some changes are hopefully not controversial,
such as replacing `Length` calls by calls of `NumberRows`,
and replacing matrix entry access `m[i][j]` by `m[i,j]`.

For other changes, there would be alternatives.
- Instead of introducing a new attribute `ListOfRowsOfMatrix`,
  one could declare `ListOp( <matobj> )` for `<matobj>` in `IsMatrixObj`
  (and not only for `<matobj>` in `IsRowListMatrix`).
  From the viewpoint that the use of `ListOp` for `IsMatrixObj` matrices
  is not recommended, according to `lib/matobj2.gd`,
  I prefer `ListOfRowsOfMatrix`.

- In the case of `DefaultScalarDomainOfMatrixList`,
  one could argue that also lists of `IsMatrixObj` objects with different `BaseDomain` can make sense,
  but I think that showing an error message is more in the spirit of `IsMatrixObj`.

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -4031,31 +4031,40 @@ end);
 ##
 #M  DefaultScalarDomainOfMatrixList
 ##
-InstallMethod(DefaultScalarDomainOfMatrixList, "generic: form ring",
-  [IsListOrCollection],
-function(l)
-local i,j,k,fg,f;
-  # try to find out the field
-  if Length(l)=0 or ForAny(l,i->not IsMatrix(i)) then
-    Error("<l> must be a list of matrices");
-  fi;
-  fg:=[l[1][1,1]];
-  if Characteristic(fg)=0 then
-    f:=DefaultField(fg);
-  else
-    f:=DefaultRing(fg);
-  fi;
-  for i in l do
-    for j in i do
-      for k in j do
-        if not k in f then
-          Add(fg,k);
-          f:=DefaultRing(fg);
-        fi;
+InstallMethod( DefaultScalarDomainOfMatrixList,
+    "generic: form ring",
+    [ IsList ],
+    function(l)
+    local B, i,j,k,fg,f;
+    # try to find out the field
+    if Length( l ) = 0 or ForAny( l, i -> not IsMatrixObj( i ) ) then
+      Error( "<l> must be a nonempty list of matrices" );
+    elif ForAll( l, HasBaseDomain ) then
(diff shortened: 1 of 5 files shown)


fingolfin:
Great, thanks for working on this!

I agree that we should not have `ListOp` for `MatrixObj`, instead `ListOfRowsOfMatrix` seems like a better idea -- although I would suggest to rename it to just `RowsOfMatrix`, or is there a good reason not to?

lib/matrix.gi
```diff
@@ -4031,31 +4031,39 @@ end);
...
+            fi;
+          od;
+        od;
       od;
-    od;
-  od;
-  return f;
+      return f;
+    fi;
+    Error( "all in <l> must either be lists of lists or have the same BaseDomain" );
```
fingolfin:
```suggestion
    Error( "all elements in <l> must either be lists of lists or have the same BaseDomain" );
```

hulpke:
Are the proposed changes in accessing a matrix 
a) stable and available in 4.10?
b) recommended already for other matrix group code ?
If so, is there a list of these changes (as I should modify my own code accordingly).

Thanks!

   Alexander

fingolfin:
Accessing matrix entries via `mat[i,j]` is in GAP 4.10 and documented, and often is faster than `mat[i][j]`. Also NumberRows / NumberColumns (resp. their aliases NrRows / NrCols) are in 4.10.

The overall MatrixObj code is not yet properly documented, this is something we badly must work on, and I hope we can make at least partial progress on it for GAP 4.11. This will then on the long run allow us to overcome all kinds of issues, and help improve performance, reduce memory usage, etc. etc.

ThomasBreuer:
@fingolfin Thanks for your comments, I have now updated the pull request.
Concerning the name `ListOfRowsOfMatrix`, my idea was to emphasize that the result is a list
(which is not obvious for objects that arise in the context of `IsMatrixObj`);
but let us try `RowsOfMatrix`.

PR #2000, opened 2017-12-06
Revise kernel #includes

fingolfin:
This PR revises all kernel #includes, stripping them down to the (for now) minimum. Key ideas and benefits:

1. each header file should #include any other header files that need to be #included before it; at the very least, `system.h`, which defines core types and includes `config.h`; this way, if you need a header, you can just #include it w/o worrying what else it in turn needs.
2. each C file should only include headers it really needs; this keeps compile times to a minimum when changing between revisions and branches, when editing a header etc.
3. inside a C file, the #includes are sorted alphabetically, to avoid having to think were one should put a new one; some exceptions:
  * I put all the `hpc/` includes to the end
   * actually, system headers come at the *very end*
   * the header matching the current C file comes first (so `#include <src/foo.h>` is first in `src/foo.c`). The idea here being that this way, each header is included first at least once, which helps to ensure goal 1 keeps being satisfied
4. No comments after each include what it is about. This is documented inside the headers, and could change, and it's just too much work to maintain; and I have yet to find any benefit in it.

Some TODOs for the future, after this PR has been merged:
* perhaps change the "central" header file from `system.h` to something else. 
* split `system.h` into the part providing essential types, defines and includes; and the part defining tons of variabls
* split `gap.h` (say into an error.h, plus more); indeed, I have some commits for that already queued up, but didn't want to mix them into this PR, which is already quite big
*  various FIXMEs I added to #includes should be resolved
* I'd like to clean up the file headers (the comments at the top of almost every .c and .h file), to a uniform format, which states a sane copyright, and probably drops pointless things (like the name of the file, which is useless, and gets out-of-sync if the file is renamed; or the "author" of the file, which does not reflect reality for almost all of those files)

PR #3094, opened 2018-12-06
Clean  up macros

ChrisJefferson:
This cleans up some more macros. In the process it adds some basic asserts and simplifies one function (it had a parameter which we always passed `True` to, so I removed it).

### DIFF
--- a/src/exprs.h
+++ b/src/exprs.h
@@ -16,7 +16,9 @@
 #ifndef GAP_EXPRS_H
 #define GAP_EXPRS_H
 
+#include "code.h"
 #include "system.h"
+#include "vars.h"
 
 /****************************************************************************
 **
@@ -25,10 +27,16 @@
 **  'OBJ_REFLVAR'  returns  the value of  the reference  to a  local variable
 **  <expr>.
 */
-#define OBJ_REFLVAR(expr)       \
-                        (OBJ_LVAR( LVAR_REFLVAR( expr ) ) != 0 ? \
-                         OBJ_LVAR( LVAR_REFLVAR( expr ) ) : \
-                         ObjLVar( LVAR_REFLVAR( expr ) ) )
+static inline Obj OBJ_REFLVAR(Expr expr)
+{
+    Int lvar = LVAR_REFLVAR(expr);
+    if (OBJ_LVAR(lvar) != 0) {
+        return OBJ_LVAR(lvar);
+    }
+    else {
+        return ObjLVar(lvar);
+    }
+}
 
 /****************************************************************************
 **
@@ -43,7 +51,23 @@
 **  (immediate) integer values having the same representation.
 */
 
-#define OBJ_INTEXPR(expr)   ((Obj)(expr))
+static inline Obj OBJ_INTEXPR(Expr expr)
+{
+    GAP_ASSERT(IS_INTOBJ((Obj)expr));
+    return (Obj)expr;
+}
+
+
+/****************************************************************************
+**
+*V  EvalExprFuncs[<type>]  . . . . . evaluator for expressions of type <type>
+**
+**  'EvalExprFuncs'  is the dispatch table   that contains for  every type of
+**  expressions a pointer  to the  evaluator  for expressions of this   type,
+**  i.e., the function that should be  called to evaluate expressions of this
+**  type.
+*/
+extern Obj (*EvalExprFuncs[256])(Expr expr);
 
 
 /****************************************************************************
@@ -61,22 +85,31 @@
 **  Note that 'EVAL_EXPR' does not use 'TNUM_EXPR', since it also handles the
 **  two special cases that 'TNUM_EXPR' handles.
 */
-#define EVAL_EXPR(expr) \
-                        (IS_REFLVAR(expr) ? OBJ_REFLVAR(expr) : \
-                         (IS_INTEXPR(expr) ? OBJ_INTEXPR(expr) : \
-                          (*EvalExprFuncs[ TNUM_STAT(expr) ])( expr ) ))
+
+static inline Obj EVAL_EXPR(Expr expr)
+{
+    if (IS_REFLVAR(expr)) {
+        return OBJ_REFLVAR(expr);
+    }
+    else if (IS_INTEXPR(expr)) {
+        return OBJ_INTEXPR(expr);
+    }
+    else {
+        return (*EvalExprFuncs[TNUM_STAT(expr)])(expr);
+    }
+}
 
 
 /****************************************************************************
 **
-*V  EvalExprFuncs[<type>]  . . . . . evaluator for expressions of type <type>
+*V  EvalBoolFuncs[<type>] . . boolean evaluator for expression of type <type>
 **
-**  'EvalExprFuncs'  is the dispatch table   that contains for  every type of
-**  expressions a pointer  to the  evaluator  for expressions of this   type,
-**  i.e., the function that should be  called to evaluate expressions of this
-**  type.
+**  'EvalBoolFuncs'  is  the dispatch table that  contains  for every type of
+**  expression a pointer to a boolean evaluator for expressions of this type,
+**  i.e., a pointer to  a function which  is  guaranteed to return a  boolean
+**  value that should be called to evaluate expressions of this type.
 */
-extern  Obj             (* EvalExprFuncs [256]) ( Expr expr );
+extern Obj (*EvalBoolFuncs[256])(Expr expr);
 
 
 /****************************************************************************
@@ -90,20 +123,10 @@ extern  Obj             (* EvalExprFuncs [256]) ( Expr expr );
 **  'EVAL_BOOL_EXPR' returns the  value of <expr> (which  is either 'true' or
 **  'false').
 */
-#define EVAL_BOOL_EXPR(expr) \
-                        ( (*EvalBoolFuncs[ TNUM_EXPR( expr ) ])( expr ) )
-
-
-/****************************************************************************
-**
-*V  EvalBoolFuncs[<type>] . . boolean evaluator for expression of type <type>
-**
-**  'EvalBoolFuncs'  is  the dispatch table that  contains  for every type of
-**  expression a pointer to a boolean evaluator for expressions of this type,
-**  i.e., a pointer to  a function which  is  guaranteed to return a  boolean
-**  value that should be called to evaluate expressions of this type.
-*/
-extern  Obj             (* EvalBoolFuncs [256]) ( Expr expr );
+static inline Obj EVAL_BOOL_EXPR(Expr expr)
+{
+    return (*EvalBoolFuncs[TNUM_EXPR(expr)])(expr);
+}
 
 
 /****************************************************************************
--- a/src/opers.h
+++ b/src/opers.h
@@ -448,9 +448,8 @@ static inline Obj SAFE_ELM_FLAGS(Obj list, UInt pos)
 **  <list> to True.  <pos> must be a positive integer less than or
 **  equal to the length of <hdList>.
 */
-static inline void SET_ELM_FLAGS(Obj list, UInt pos, Obj val)
+static inline void SET_ELM_FLAGS(Obj list, UInt pos)
 {
-    GAP_ASSERT(val == True);
     GAP_ASSERT(TNUM_OBJ(list) == T_FLAGS);
     GAP_ASSERT(pos <= LEN_FLAGS(list));
     BLOCKS_FLAGS(list)[(pos - 1) >> LBIPEB] |= MASK_POS_FLAGS(pos);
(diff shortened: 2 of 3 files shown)


src/exprs.h
```diff
@@ -61,22 +84,31 @@
...
+    }
+    else {
+        return (*EvalExprFuncs[TNUM_STAT(expr)])(expr);
+    }
+}
 
 
 /****************************************************************************
 **
-*V  EvalExprFuncs[<type>]  . . . . . evaluator for expressions of type <type>
```
ChrisJefferson: This bit looks a little strange as I had to rearrange the order of a couple of things, now we are using functions rather than macros.

src/exprs.h
```diff
@@ -17,6 +17,8 @@
 #define GAP_EXPRS_H
 
 #include "system.h"
+#include "vars.h"
+#include "code.h"
```
fingolfin: Please keep all #includes strictly alphabetically sorted.

Also, the need to add those includes here now is the main reason I did not convert those macros so far, as I was trying hard to avoid inter-header dependencies, to cut down (re)compile times. So, I am a bit sad to see this added here, but of course there are good reasons (and I can try to get rid of them again in the future through careful refactoring ;-)

src/exprs.h
```diff
@@ -25,10 +27,15 @@
...
-                         OBJ_LVAR( LVAR_REFLVAR( expr ) ) : \
-                         ObjLVar( LVAR_REFLVAR( expr ) ) )
+static inline Obj OBJ_REFLVAR(Expr expr)
+{
+    if (OBJ_LVAR(LVAR_REFLVAR(expr)) != 0) {
+        return OBJ_LVAR(LVAR_REFLVAR(expr));
+    }
+    else {
+        return ObjLVar(LVAR_REFLVAR(expr));
+    }
```
fingolfin: So I understand that this is the most straightforward implementation, but since we are no in a function, we could also untangle the code to make it more palatable, e.g. like this:
```C
    Int lvar = LVAR_REFLVAR(expr);
    Obj obj = OBJ_LVAR(lvar);
    return obj ? obj : ObjLVar(lvar);
```

src/opers.h
```diff
@@ -16,7 +16,7 @@
 
 #include "system.h"
 #include "calls.h"
-
+#include "bool.h"
```
fingolfin: Please keep #includes sorted.

But in this case, I'd rather avoid the include; see below for more on this.

src/opers.h
```diff
@@ -386,41 +416,53 @@ static inline UInt NRB_FLAGS(Obj flags)
...
-#define C_ELM_FLAGS(list, pos)                                               \
-    ((BLOCK_ELM_FLAGS(list, pos) & MASK_POS_FLAGS(pos)) ? 1 : 0)
+static inline Int C_ELM_FLAGS(Obj list, UInt pos)
+{
+    if (BLOCK_ELM_FLAGS(list, pos) & MASK_POS_FLAGS(pos)) {
+        return 1;
+    }
+    else {
+        return 0;
+    }
```
fingolfin: How about this instead?
```C
    return (BLOCK_ELM_FLAGS(list, pos) & MASK_POS_FLAGS(pos)) != 0;
```

src/opers.h
```diff
@@ -386,41 +416,53 @@ static inline UInt NRB_FLAGS(Obj flags)
...
 static inline Int SAFE_C_ELM_FLAGS(Obj flags, UInt pos)
 {
     return (pos <= LEN_FLAGS(flags)) ? C_ELM_FLAGS(flags, pos) : 0;
 }
 
-#define SAFE_ELM_FLAGS(list, pos) (SAFE_C_ELM_FLAGS(list, pos) ? True : False)
+static inline Obj SAFE_ELM_FLAGS(Obj list, UInt pos)
+{
+    return SAFE_C_ELM_FLAGS(list, pos) ? True : False;
+}
```
fingolfin: I would suggest to get rid of `SAFE_ELM_FLAGS` and `ELM_FLAGS`; and instead rename `C_ELM_FLAGS` to `ELM_FLAGS`, and `SAFE_C_ELM_FLAGS` to `SAFE_ELM_FLAGS`.  Then there is no need for including `bool.h`. And (hypothetical) external code accessing a flags list usually will be better of getting 0/1 return values anyway.

If you agree with this idea, I can also implement it and push that as an addition to this PR.

ChrisJefferson: I don't mind it in principle, but there are quite a few places using ELM_FLAGS and SAFE_ELM_FLAGS, and I worry they are going to all look messy if we start explictly converting 0/1 to True/False.

src/exprs.h
ChrisJefferson: I more feel these headers were "always needed" (as we were defining macros which couldn't be used), but we just didn't realise before.

ChrisJefferson: I point out .clang-format says `SortIncludes: false`, and we don't really have a coding guide written anywhere I know of. I will rearrange them, but I don't know how I could have known to sort them.

fingolfin: The convention is this: all headers are sorted, *except* that inside of `foo.c`, we always include `foo.h` first. This is why we can't let clang-format sort them for us, otherwise I'd have enabled this. This ensures that headers don't have accidentally missing dependencies, i.e., no "to use foo.h you must always include it after bar.h". 

As to how you could have known about this: Well, I did not mean to say you should have known, I merely pointed it out so it could be fixed. Although of course in theory, you could also have remembered it from your review of PR #2000, where this was explained :-). But again, I did not really expect that.

I don't know where I could document such a rule, though *shrug*. And to be honest, even if I documented it -- would you or anybody check that document before making a PR ;-). People don't run clang-format all the time, too *shrug*. Anyway, if you have a good idea where to place this, I can write it up there!

src/opers.h
fingolfin: There are exactly three each :)

fingolfin: Here is the full patch eliminating all uses of them. As I said, I could adjust the PR to this end, or I can make another PR where I make the changes (but that would then of course conflict with this PR, albeit in a minor way.
```patch
diff --git a/src/opers.c b/src/opers.c
index b3d23fdfb..612ab8a9b 100644
--- a/src/opers.c
+++ b/src/opers.c
@@ -1082,7 +1082,7 @@ Obj DoSetFilter (
     flags = FLAGS_TYPE( type );
     
     /* return the value of the feature                                     */
-    if ( val != SAFE_ELM_FLAGS( flags, flag1 ) ) {
+    if ( (val == True) != SAFE_C_ELM_FLAGS( flags, flag1 ) ) {
         ErrorReturnVoid(
             "value feature is already set the other way",
             0L, 0L,
@@ -1125,10 +1125,7 @@ Obj DoFilter (
     flags = FLAGS_TYPE( type );
     
     /* return the value of the feature                                     */
-    val = SAFE_ELM_FLAGS( flags, flag1 );
-    
-    /* return the value                                                    */
-    return val;
+    return SAFE_C_ELM_FLAGS( flags, flag1 ) ? True : False;
 }
 
 
@@ -2478,7 +2475,7 @@ Obj DoTestAttribute (
     flags = FLAGS_TYPE( type );
 
     /* return whether the value of the attribute is already known          */
-    return SAFE_ELM_FLAGS( flags, flag2 );
+    return SAFE_C_ELM_FLAGS( flags, flag2 ) ? True : False;
 }
 
 
@@ -2865,7 +2862,7 @@ Obj DoSetProperty (
 
     /* if the value of the property is already known, compare it           */
     if ( SAFE_C_ELM_FLAGS( flags, flag2 ) ) {
-        if ( val == ELM_FLAGS( flags, flag1 ) ) {
+        if ( (val == True) == C_ELM_FLAGS( flags, flag1 ) ) {
             return 0;
         }
         else {
@@ -2933,7 +2930,7 @@ Obj DoProperty (
 
     /* if the value of the property is already known, simply return it     */
     if ( SAFE_C_ELM_FLAGS( flags, flag2 ) ) {
-        return ELM_FLAGS( flags, flag1 );
+        return C_ELM_FLAGS( flags, flag1 ) ? True : False;
     }
 
     /* call the operation to compute the value                             */
@@ -2989,7 +2986,7 @@ Obj DoVerboseProperty (
 
     /* if the value of the property is already known, simply return it     */
     if ( SAFE_C_ELM_FLAGS( flags, flag2 ) ) {
-        return ELM_FLAGS( flags, flag1 );
+        return C_ELM_FLAGS( flags, flag1 ) ? True : False;
     }
 
     /* call the operation to compute the value                             */
```

src/exprs.h
ChrisJefferson: In another project I work on, we use the fact clang-format won't reorder headers if you leave a blank line between them to group headers into sets which can then be automatically ordered.

src/exprs.h
ChrisJefferson: fixed (well, partially, I'd prefer keeping the if)

src/exprs.h
ChrisJefferson: fixed

src/opers.h
ChrisJefferson: I haven't done this, I'll be honest, I think this makes the code quite a bit harder to read.

ChrisJefferson:
Several cleanups made

src/exprs.h
fingolfin: Ah, I didn't know that. Excellent, let's enable it then, we already separate that "primary" header from the others with a newline.

src/opers.h
fingolfin: Still not sorted...  :-(

src/opers.h
```diff
@@ -358,11 +385,13 @@ static inline UInt NRB_FLAGS(Obj flags)
 **  <pos>  must be a positive  integer  less than or  equal  to the length of
 **  <list>.
 **
-**  Note that 'BLOCK_ELM_FLAGS' is a macro, so do not call it  with arguments
```
fingolfin: Ideally the `**` line above would also be removed.

src/opers.h
```diff
@@ -372,10 +401,11 @@ static inline UInt NRB_FLAGS(Obj flags)
 **  '(<pos>-1) % BIPEB',
 **  useful for accessing the <pos>-th element of a 'FLAGS' list.
 **
-**  Note that 'MASK_POS_FLAGS'  is a macro, so  do not call it with arguments
```
fingolfin: Ideally the `**` line above would also be removed.

src/opers.h
```diff
@@ -386,41 +416,48 @@ static inline UInt NRB_FLAGS(Obj flags)
 **  is either 'true' or 'false'.  <pos> must  be a positive integer less than
 **  or equal to the length of <hdList>.
 **
-**  Note that 'ELM_FLAGS' is a macro, so do not call it  with arguments  that
```
fingolfin: Ideally the `**` line above would also be removed.

src/opers.h
fingolfin: Note that we use this pattern (`val ? True : False` already in countless places... I don't share the view that it's hard to read, though I guess one could come up with a macro/helper function to hide it away... But frankly, I'd then probably find *that* harder to read, as to understand that, one would have to lookup what that macros does.

Anyway, I can make a PR with the changes I have in mind separately, and we can discuss it further there (or not)

src/opers.h
```diff
@@ -386,41 +416,48 @@ static inline UInt NRB_FLAGS(Obj flags)
...
 **
 *F  SET_ELM_FLAGS( <list>, <pos>, <val> ) . .  set an element of a flags list
 **
 **  'SET_ELM_FLAGS' sets  the element at position <pos>   in the flags list
-**  <list> to the value <val>.  <pos> must be a positive integer less than or
-**  equal to the length of <hdList>.  <val> must be either 'true' or 'false'.
+**  <list> to True.  <pos> must be a positive integer less than or
+**  equal to the length of <hdList>.
 **
-**  Note that  'SET_ELM_FLAGS' is  a macro, so do not  call it with arguments
```
fingolfin: Ideally the `**` line above would also be removed.

ChrisJefferson:
The ? True : False don't bother me so much as the `(y == True) == C_ELM_FLAGS`, as I find the bracketed double equals confusing, and also this has a different meaning in C, as the old version `(y == ELM_FLAGS)` would always reject `y` is `Fail`, but the new one combines `false` and `fail`. Of course, this might not matter, or might even be "better", but it's not just a simple rename / removal of a function.

ChrisJefferson:
Hopefully all issues fixed!

fingolfin:
I do see your point on `(y == True) == C_ELM_FLAGS`.

fingolfin:
Thank you! Do you want to get rid of the `fixup!` commits before we merge?

(I'd plan to wait for the master test to finish before merging anyway, but feel free to merge earlier, too -- my approval from before stands :)

ChrisJefferson:
Woops forgot the --autosquash! Will do it when I'm next in front of a computer

PR #3095, opened 2018-12-06
Change GroupWithGenerators to accept collections again

### LINKED ISSUE #2703: Call GroupByGenerators/GroupWithGenerators with a collection?

fingolfin:
In GAP <= 4.9, this worked:

    gap> GroupByGenerators( Group( (1,2) ) );
    Group([ (), (1,2) ])

In GAP 4.10, this was broken and instead lead to a "method not found" error.
While strictly speaking this was never documented behavior, we restore it
to avoid regressions in code that relied on this undocumented behavior.

Resolves #2703 

This is a less intrusive (and thus hopefully less controversial) alternative to PR #3091. Closes #3091

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -4401,13 +4401,14 @@ InstallMethod( GroupWithGenerators,
 function( gens )
 local G,typ;
 
+  gens:=AsList(gens);
   typ:=MakeGroupyType(FamilyObj(gens),
           IsGroup and IsAttributeStoringRep 
           and HasIsEmpty and HasGeneratorsOfMagmaWithInverses,
           gens,false,true);
 
   G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens));
+  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,gens);
 
   return G;
 end );
@@ -4418,13 +4419,14 @@ InstallMethod( GroupWithGenerators,
 function( gens, id )
 local G,typ;
 
+  gens:=AsList(gens);
   typ:=MakeGroupyType(FamilyObj(gens),
           IsGroup and IsAttributeStoringRep 
             and HasIsEmpty and HasGeneratorsOfMagmaWithInverses and HasOne,
             gens,id,true);
 
   G:=rec();
-  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,AsList(gens),
+  ObjectifyWithAttributes(G,typ,GeneratorsOfMagmaWithInverses,gens,
                           One,id);
 
   return G;
--- a/tst/testbugfix/2018-12-06-GroupWithGenerators.tst
+++ b/tst/testbugfix/2018-12-06-GroupWithGenerators.tst
@@ -1,0 +1,5 @@
+# These are undocumented, but should still work
+gap> GroupByGenerators( Group( (1,2) ) );
+Group([ (), (1,2) ])
+gap> GroupWithGenerators( Group( (1,2) ) );
+Group([ (), (1,2) ])


ChrisJefferson:
I don't like the idea that is something doesn't break any library code, that's fine. There are many thousands of GAP users out in the world, all with their own code. If (as in this case) it is reasonable for us not to break existing code, which has worked for many years, then I think we should try to do so -- particularly when the error message is hard to read.

There are many places in GAP where you can put objects which can be enumerated instead of an explicit list, and in any of those places `SymmetricGroup(20)` would be a bad idea.

ChrisJefferson:
I don't like the idea that is something doesn't break any library code, that's fine. There are many thousands of GAP users out in the world, all with their own code. If (as in this case) it is reasonable for us not to break existing code, which has worked for many years, then I think we should try to do so -- particularly when the error message is hard to read.

There are many places in GAP where you can put objects which can be enumerated instead of an explicit list, and in any of those places `SymmetricGroup(20)` would be a bad idea.

alex-konovalov:
@ChrisJefferson there is still an alternative:
> The other idea suggested elsewhere was to check if the argument is a domain which knows a list of its generators, and in this case call GroupByGenerators on that list. 

I would be happy with one of the following:
- document the alternative behaviour as above, or 
- make a better error message when it's called with a domain, or
- accept this PR with the following modification: if it called on a domain, then it does everything like it does now (i.e. calls `AsList`) but prints a warning:

    #I Calling GroupByGenerators on a domain is very inefficient, use the list of generators instead.

Anyone in favour?

fingolfin:
@ThomasBreuer ah, thanks for pointing out the extra `AsList` calls; they are now removed

@alex-konovalov this only modifies two methods for `GroupWithGenerators`, hence I only mention that in the commit message (well, now at least, I had `GroupByGenerators` in there and in the title of this PR before by accident). Of course this then also affects any code calling or delegating to `GroupWithGenerators`, such as `GroupByGenerators`.

Of course `GroupByGenerators(SymmetricGroup(20))` is not sensible, but so what? This just restores behaviour that was present in GAP 4.9, while not affecting anything else. In particular, any "sane" code which only passes in regular list  of generators sees no change in behavior.

fingolfin:
@ThomasBreuer ah, thanks for pointing out the extra `AsList` calls; they are now removed

@alex-konovalov this only modifies two methods for `GroupWithGenerators`, hence I only mention that in the commit message (well, now at least, I had `GroupByGenerators` in there and in the title of this PR before by accident). Of course this then also affects any code calling or delegating to `GroupWithGenerators`, such as `GroupByGenerators`.

Of course `GroupByGenerators(SymmetricGroup(20))` is not sensible, but so what? This just restores behaviour that was present in GAP 4.9, while not affecting anything else. In particular, any "sane" code which only passes in regular list  of generators sees no change in behavior.

Put another way: any code that currently is "correct" and "adheres to the documentation" will work unchanged. All this PR does is to restore a behavior of GAP that was present for decades (I just checked, GAP 4.4 supports it). I see zero harm in it. Note that I am *not* suggesting to document this behavior, either, but I don't see how we benefit from gratuitously breaking code that was working before?

alex-konovalov:
@fingolfin This will not fix #2703 because that one goes into the break loop because of `GroupByGenerators`.

fingolfin:
@alex-konovalov while I wouldn't mind that warning, IMHO it should be added in a separate PR. This PR is purely about restoring behaviour that was present in GAP for the past 20+ years. To this end, I strived for a minimal set of changes. Any further changes and optimizations, documentation etc. are IMHO beyond the scope of this.

alex-konovalov:
@fingolfin This will not fix #2703 because that one goes into the break loop because of `GroupByGenerators`.

fingolfin:
@alex-konovalov while I wouldn't mind that warning, IMHO it should be added in a separate PR. This PR is purely about restoring behaviour that was present in GAP for the past 20+ years. To this end, I strived for a minimal set of changes. Any further changes and optimizations, documentation etc. are IMHO beyond the scope of this.

And as I said before: `GroupByGenerators` delegates to `GroupWithGenerators`. I even added a .tst file which verifies this.

fingolfin:
Backported via ac82a5083dd148a17755a011b2a0a10c943f9d04

PR #3098, opened 2018-12-07
Remove 'end of file' comments

fingolfin:


lib/quotsys.gd
```diff
@@ -198,11 +198,6 @@ DeclareOperation( "GroupByQuotientSystem", [IsQuotientSystem] );
 #O  TraceDefinition . . . . . . trace a generator back to defining generators
 ##
 DeclareOperation( "TraceDefinition", [IsQuotientSystem, IsPosInt] );
-
```
alex-konovalov: This is not an end-of-file comment

lib/pquot.gi
```diff
@@ -1929,11 +1929,6 @@ InstallMethod( TraceDefinition,
     Add( trace, g );
     return Reversed( trace );
 end );
-
```
alex-konovalov: This is not an EOF comment.

lib/quotsys.gd
fingolfin: Oops, thanks!

alex-konovalov:
New copyright headers and removing EOF comments is perhaps something that some package authors may be willing to adopt too.

ChrisJefferson:
Before merging this, and the tab-to-space other commit, I thought we had previously agreed not to make large-scale formatting changes which didn't provide any significant gain, to avoid breaking people's forks (I have one, I know other people have them too).

fingolfin:
Personally, I don't consider this PR to be a formatting change; I don't quite see how it could conflict with any other change either, unless that change for somevreason was modifying one of those end markers - or am I overlooking something?

As to the tab changes, I would prefer to discuss that on the relevant PR.

alex-konovalov:
A merge conflict may happen when someone added next code at the end of the file, just above the end-of-file comment:
```
<<<<<<< HEAD

Let us put more text before the end of the file.

#####################
#End of file comments
##


=======
>>>>>>> <commit-id>
```
This is not hard to resolve, and it will only happen if the new code is appended at the end, though.

ChrisJefferson:
Ok, j just aren't sure I feel the cost/benefit of this PR to be worth it, but then random comments on ends of files and similar things don't bother me at all. Others may have a different opinion.

PR #3099, opened 2018-12-07
Show a warning when GroupWithGenerators called on a domain

alex-konovalov:
In GAP <= 4.9, the following undocumented use of `GroupWithGenerators`
was possible:

    gap> GroupWithGenerators( Group( (1,2) ) );
    Group([ (), (1,2) ])

This does not work in GAP 4.10.0, and #3095 restored this never documented
behavior to avoid regressions in code that used to work previously.

This PR adds a warning when such use of `GroupWithGenerators` is detected.

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -4401,6 +4401,13 @@ InstallMethod( GroupWithGenerators,
 function( gens )
 local G,typ;
 
+  if IsGroup(gens) then
+    Info( InfoPerformance, 1,
+      "Calling `GroupWithGenerators' on a group usually is very inefficient.");
+    Info( InfoPerformance, 1,
+      "Use the list of generators of the group instead.");
+  fi;
+
   gens:=AsList(gens);
   typ:=MakeGroupyType(FamilyObj(gens),
           IsGroup and IsAttributeStoringRep 
@@ -4418,6 +4425,13 @@ InstallMethod( GroupWithGenerators,
     IsCollsElms, [ IsCollection, IsMultiplicativeElementWithInverse ],
 function( gens, id )
 local G,typ;
+
+  if IsGroup(gens) then
+    Info( InfoPerformance, 1,
+      "Calling `GroupWithGenerators' on a group usually is very inefficient.");
+    Info( InfoPerformance, 1,
+      "Use the list of generators of the group instead.");
+  fi;
 
   gens:=AsList(gens);
   typ:=MakeGroupyType(FamilyObj(gens),


lib/grp.gi
```diff
@@ -4401,6 +4401,12 @@ InstallMethod( GroupWithGenerators,
 function( gens )
 local G,typ;
 
+  if IsDomain(gens) then
+    Info( InfoPerformance, 1,
+    "Calling `GroupWithGenerators' on a domain is very inefficient.\n",
```
fingolfin: It is only inefficient for large domains. So perhaps "... on a domain usually is very inefficient"?

lib/grp.gi
```diff
@@ -4401,6 +4401,12 @@ InstallMethod( GroupWithGenerators,
...
+    "#I  Use the list of generators of the domain instead.");
```
fingolfin: This hardcodes an assumption about how the Info message is printed. It's better to use two `Info` statements instead, or else put everything in one line.

alex-konovalov: Sure. Updated to resolve both comments.

alex-konovalov:
@markuspf what do you mean by "why?"? The commit message explains the reasons, and you may also see the discussion in #3095 and related PRs/issues. 

`InfoPerformance` sounds like an optimal info level for such messages, and it's default value is 1 when GAP starts, but is set to 0 in regression tests, so this message will not show up there.

ChrisJefferson:
Are you sure this mesage is true for all domains? I think could make a domain where taking its generators and making a group from them would produce a different answer (although in such a case the user is doing something very strange...)

alex-konovalov:
@ChrisJefferson Do you think it should only check for `IsGroup` instead of `IsDomain`? Indeed, the way how the domain is generated by its generators may be different.

ChrisJefferson:
Group feels like an easy case to catch -- as previously discussed we could even then just special case groups, but then we are going in a loop :) I don't know enough about how domains are defined to know if there is a good way of catching and warning about exactly the right things.

alex-konovalov:
Indeed,`GeneratorsOfDomain` for a domain which is not a group may be generators w.r.t. different operation(s), so for that case it might have sense to use `GroupWithGenerators` on the whole domain. I've pushed the new version which uses `IsGroup` instead of `IsDomain`.

lib/grp.gi
```diff
@@ -4401,6 +4401,13 @@ InstallMethod( GroupWithGenerators,
 function( gens )
 local G,typ;
 
+  if IsGroup(gens) then
+    Info( InfoPerformance, 1,
+      "Calling `GroupWithGenerators' on a group usually is very inefficient.");
+    Info( InfoPerformance, 1,
+      "Use the list of generators of the group instead.");
```
fingolfin: But isn't the actual problem that using a large number of generators is usually inefficient? Then, shouldn't the check instead be on the *size* of `gens`? And then, shouldn't it apply equally to lists as to domains or any other kind of collection?

alex-konovalov: But how then would you know when the number is large? I assume that the number of generators is usually smaller than the number of elements in the group (unless it is already given using a very inefficient list of generators).

fingolfin: Sure, any number we'd pick as "large" would be arbitrary. Just like it is arbitrary to allow `Group(AsList(G))` as generators, but not `Group(G)`. So my personal preference would be to do neither, i.e., exactly as it is right now...

alex-konovalov: Can't think of any scenario when calling `Group(Group(...))` is more efficient than `Group(GeneratorsOfGroup(...))` - that's what I am suggesting to catch. Not other inefficiencies, as that would be more ambiguous...

PR #3101, opened 2018-12-08
Removing Section 87.2-5 in the Reference Manual

wucas:
The method for avoiding multiplication of permutations described in Section 87.2-5 does not work. 
It seems that the code providing the data structure used (which is the rgtObj and lftObj as described in Section 87.2-5) is commented out. Thus the section should be removed.


ChrisJefferson:
Thanks! Indeed this section of the documentation has been wrong for.. about 20 years or so!

alex-konovalov:
@wucas could you please point out to the location of the commented out code - may be those line could now be deleted completely too.

ChrisJefferson:
I'd prefer the commented on line stay -- it is in the partition backtrack code where it is useful to be able to see the history.

markuspf:
Just for reference the code is here: https://github.com/gap-system/gap/blob/72d2c365f3201b20c9ffdc1d108fafeaf19a355a/lib/stbcbckt.gi#L1458 and I agree with @ChrisJefferson that for the moment it is more useful that the commented code is there for understanding what the algorithm is supposed to do and to not confuse the "casual" reader more about the intention of the code.

wucas:
@alex-konovalov I get the intention of keeping a trace of this feature. But I'm not sure how such an explanation would fit in the documentation. It seems at bit out of place to suddenly explain something about code that was commented out years ago. For the casual reader i would argue that as long as the functionality is undocumented i wouldn't worry about it. I only stumbled over it because it was documented. Otherwise I'd never had the idea to use it in the first place. 
Also I personally am not a big fan of commenting the section out in `doc/ref/stbchain.xml` just because it would leave the code in a bigger mess than I found it in.
If someone tries to understand the code in `gap/lib/stbcbckt.gi` I at least wouldn't look at the source code of the documentation. If it is about helping someone reading the code I think the best would be to explain it briefly in the code itself by adding a comment there.

alex-konovalov:
@wucas agree - I think it will suffice if you will add to this PR a comment at `gap/lib/stbcbckt.gi` saying something like  
```
# Here and below the code that refers to `rgtObj` was used to avoid multiplication
# of permutations. It has been commented out for a long time, but accidentally remained
# documented in `doc/ref/stbchain.xml` until its withdrawal in 2018.
```

wucas:
@alex-konovalov thanks for the suggestion. I added the comment at the first instance of `rgtObj` in the partition backtrack function in `gap/lib/stbcbckt.gi`.

alex-konovalov:
Made a final look at the removed XML - and the only example in that section used `Listing` element (since it does not have a runnable code), so it was excluded from tests all the time ... :-(
```
<Listing><![CDATA[	
return PartitionBacktrack( G,	
  [ g, g,	
  OnPoints,	
  c -> c!.lftObj = c!.rgtObj ],	
  false, R, [ Pi_0, g ], L, R );	
]]></Listing>	
```

Issue #3028, opened 2018-11-19
Possible infinite recursion in PRINT_OR_APPEND_TO_FILE_OR_STREAM


embray:
If a stream that is being used to print errors cannot be opened (e.g. an error elsewhere prevents streams from being closed properly) this can result in an infinite recursion when reporting that error.

The specific problem is in the error handling logic here: https://github.com/gap-system/gap/blob/efd36263527f76bf3cc1cf33c5799a9bbc8734a6/src/streams.c#L799

If an `ErrorQuit` (for example) occurs elsewhere, and the stream assigned to `ERROR_OUTPUT` cannot be opened for whatever reason, this results in another call to `ErrorQuit` with the same broken stream assigned and so on.  Perhaps this should do something else: Try to open `*errout*` as a last resort, `Panic()`, etc.

markuspf:
See #2487

fingolfin:
I believe we resolved this some time ago (and the fixes are already in GAP 4.11). If you still can trigger an issue and/or believe there is one based on reading the code, please don't hesitate to reopen this and/or file a new issue.

embray:
I can believe it's been fixed but I'm not sure since the code I originally implicated hasn't been updated, but it might be dealt with elsewhere.  I admit I don't recall how the issue was originally triggered.  It might have shown up while developing the workaround I had to add for the Python libgap interface to capture error output: https://gitlab.com/sagemath/sage/-/blob/1e728ac0d36514f52871268588d38fe1392a5e5a/src/sage/libs/gap/util.pyx#L198

fingolfin:
There is a new dedicated function for opening errout in `src/error.c` with fallbacks in case there is an error doing so. See https://github.com/gap-system/gap/blob/9723019120f605fe89c590d86aa40d95c735563c/src/error.c#L58

Issue #2487, opened 2018-05-24
Abuse of TRY_READ for general error handling can lead to crashes and other problems


markuspf:
Consider the following code:
```gap
DeclareRepresentation( "IsMyOutputStreamRep",
                       IsComponentObjectRep,
                       ["kernel", "socket", "format"] );

MyOutputStreamType := NewType(
    StreamsFamily,
    IsOutputTextStream and IsMyOutputStreamRep );

InstallMethod( WriteAll,
    "output text string",
    [ IsOutputTextStream and IsMyOutputStreamRep,
      IsString ],
function( stream, string )
    local str, out;
    str := "";
    out := OutputTextString(str, false);
    PrintTo(out, string);
    return true;
end );

errstream := Objectify(MyOutputStreamType, rec( format := false ));

OutputLogTo(errstream);

x := 8
```

with `gap-4.9.1`:

```
  ┌───────┐   GAP 4.9.1 of 05-May-2018
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64
 Configuration:  gmp 6.1.0, readline
 Loading the library and packages ...
 Packages:   AClib 1.3, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.9, Browse 1.8.7, Carat 2.2.2, CRISP 1.4.4, Cryst 4.1.17, CrystCat 1.1.8, CTblLib 1.2.2, 
             FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.1, IO 4.5.1, IRREDSOL 1.4, LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.13.1, PrimGrp 3.3.1, RadiRoot 2.8, 
             ResClasses 4.7.1, SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.54
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
Syntax error: ; expected in example.g:31
fish: “./gap example.g” terminated by signal SIGSEGV (Address boundary error)
```

with `master`

```
 ┌───────┐   GAP 4.10dev-714-gd6d1b20 of today
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64
 Configuration:  gmp 6.1.0, readline
 Loading the library and packages ...
[...]

PR #3102, opened 2018-12-11
prevent infinite recursions in echoandcheck SyWriteandcheck

embray:
This can occur in cases where the stdio streams (specifically stderr, but it can also happen with just stdout) cannot be written to for some reason, resulting in EIO errors on `write()` calls to their file descriptors.

The recursions occur because if an error occurs on writing to the stream, `ErrorQuit` is called which may in turn attempt to write to the same stream.    This is similar to, but a somewhat different manifestation of #3028.

I'm not really happy with this fix since it's more of a band-aid; what is really needed is broader refactoring to error handling, as discussed partly in #2487.  But that's a longer-term effort.  In the meantime, this is needed for libgap, preferably in **4.10.1**.

To demonstrate the problem this fixes, one easy way is to just run:

```
$ ./gap > /dev/full
$ ./gap > /dev/full
Error, Cannot write to file descriptor 1, see 'LastSystemError();'
 in
  Print( " ", btop, "   ", gap, " ", GAPInfo.BuildVersion, " of ", sysdate, "\n", " ", vert, "  GAP  ", vert,
 "   https://www.gap-system.org\n", " ", bbot, "   Architecture: ", GAPInfo.Architecture, "\n"
 ); at /home/embray/src/gap-system/gap/lib/init.g:510 called from
ShowKernelInformation(  ); at /home/embray/src/gap-system/gap/lib/init.g:565 called from
func(  ); at /home/embray/src/gap-system/gap/lib/system.g:215 called from
 called from read-eval loop at /home/embray/src/gap-system/gap/lib/init.g:567
Error, Cannot write to file descriptor 1, see 'LastSystemError();'

< ... some more errors ... >

<function "SESSION">( <arguments> )
 called from read-eval loop at *stdin*:1
Syntax error: ; expected

Segmentation fault (core dumped)
```

With the patch, the same example will result in a few error messages as well, but then wait at the prompt (which just isn't displayed).

This is just one example.  A more realistic use case is one where GAP is started as a subprocess with stdout and stderr redirected to pipes, and then one of those pipes is closed.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -153,19 +153,24 @@ ssize_t echoandcheck(int fid, const char *buf, size_t count) {
   int ret;
   if (syBuf[fid].type == gzip_socket) {
       ret = gzwrite(syBuf[fid].gzfp, buf, count);
-      if (ret < 0)
+      if (ret < 0) {
           ErrorQuit(
               "Could not write to compressed file, see 'LastSystemError();'\n",
               0L, 0L);
+      }
   }
   else {
       ret = write(syBuf[fid].echo, buf, count);
-      if (ret < 0)
-          ErrorQuit("Could not write to file descriptor %d, see "
-                    "'LastSystemError();'\n",
-                    syBuf[fid].fp, 0L);
+      if (ret < 0) {
+          if (syBuf[fid].fp == fileno(stdout) || syBuf[fid].fp == fileno(stderr)) {
+              Panic("Could not write to stdout/stderr.");
+          } else {
+              ErrorQuit("Could not write to file descriptor %d, see "
+                        "'LastSystemError();'\n",
+                        syBuf[fid].fp, 0L);
+          }
+      }
   }
-
   return ret;
 }
 
@@ -1540,17 +1545,23 @@ static ssize_t SyWriteandcheck(Int fid, const void * buf, size_t count)
     int ret;
     if (syBuf[fid].type == gzip_socket) {
         ret = gzwrite(syBuf[fid].gzfp, buf, count);
-        if (ret < 0)
+        if (ret < 0) {
             ErrorQuit(
                 "Cannot write to compressed file, see 'LastSystemError();'\n",
                 0L, 0L);
+        }
     }
     else {
         ret = write(syBuf[fid].fp, buf, count);
-        if (ret < 0)
-            ErrorQuit("Cannot write to file descriptor %d, see "
-                      "'LastSystemError();'\n",
-                      syBuf[fid].fp, 0L);
+        if (ret < 0) {
+            if (syBuf[fid].fp == fileno(stdout) || syBuf[fid].fp == fileno(stderr)) {
+                Panic("Could not write to stdout/stderr.");
+            } else {
+                ErrorQuit("Cannot write to file descriptor %d, see "
+                          "'LastSystemError();'\n",
+                          syBuf[fid].fp, 0L);
+            }
+        }
     }
 
     return ret;


ChrisJefferson:
While I agree with the direction of this, I don't think this is the way to fix it. There are two main problems:

1) this isn't thread safe for hpcgap.

2) I worry we could longjmp out of this function without decreasing depth.

ChrisJefferson:
I also wonder what the best thing to do in this case is.. if you can't print error messages, terrible things are happening. While I understand a segfault is bad, I'm not sure what's better. I wouldn't want the error message just being thrown away for example.

ChrisJefferson:
For example, I'd we committed this, we could end up silently dropping writes to files, without any feedback.

embray:
I had the same concerns, but this patch has been applied, in some form or other, in Sage's GAP package for quite a long time without any concerns.  That was before multi-threaded GAP though: Perhaps it should instead it should use a thread-local variable for this in one of the state slots: Something specifically for detecting recursion into I/O functions while trying to handle error I/O...

> if you can't print error messages, terrible things are happening.

That's an exaggeration.  Perfectly normal things may be happening and I gave such an example: Where stderr is replaced with a pipe and one end of the pipe is closed.  This should not cause a segfault.

> While I understand a segfault is bad, I'm not sure what's better. I wouldn't want the error message just being thrown away for example.

In most cases where this might happen it would be during process shutdown anyways. This might even happen if GAP would have otherwise exited with a normal exit code.  Instead it results in the GAP process hanging for a long time in a loop until it segfaults.  I think it would be better in this case to just exit silently.

embray:
Out of curiosity I looked into how Python handles this sort of case in the roughly equivalent parts of its code, and it mostly just ignores such write errors and doesn't make a big deal out of them.  It will still work as best it can.  Even if you run the Python REPL like `python > /dev/full` it will *mostly* work.  

Typing input still echos characters if readline is enabled, since it will still enable terminal echo if at least stdin is a TTY.  Same on GAP with this patch--at least enough to type `quit;`.

embray:
Also,
> I worry we could longjmp out of this function without decreasing depth.

That's not really a problem, since that only occurs if there is an error writing to the error stream; you would basically never recover in this case and it would stop trying to go into `ErrorQuit` where it can't display errors.

In other words, it's try once, and once an error has occurred trying to output to some stream it won't try to raise an error about it again.

Other approaches would be to either not raise an error in this case at all, at least if `fid == 3` or maybe if `fid == 1`, or to somehow handle this from within the error-handling code, but that seems fragile.

ChrisJefferson:
I wouldn't worry if we exited cleanly, instead of segfaulting

ChrisJefferson:
I pressed a bad button. Ignore.

embray:
> I'm worried about continuing to compute and produce answers, with error messages never shown to the user

 I wouldn't worry too much about that. Again, we're talking about situations that are not likely to occur for normal users of GAP in an interactive context.  There's no reason any process *needs* to be able to output anything to standard I/O, such as a process that is only writing to some other files.  If, for some reason, you can't write to stderr then you just can't write to stderr and that's that.

ChrisJefferson:
I feel there is a difference in philosophy here -- my personal preference is, in a system where people want to trust the results, if we reach a state where we can't tell the user information (like errors), then the best option is to quit loudly and quickly (we could try sending a useful message to C's stdout/stderr and then call exit/abort rather than a stack overflow of course, or try and recover and redirect errors to stderr if the user has redirected them elsewhere).

What exactly is the need for this in libgap? I realise there is the bigger picture (someone closes stderr), but that doesn't (to me) seem like a super-urgent problem, so I assume this is coming up more often in Sage for some reason?

ChrisJefferson:
Also, the best fix for GAP (when someone has redirected output) would probably be to force  error output back to stderr, and if that disappears then just abort(), but that's probably not what Sage/libgap would want.

jdemeyer:
As @embray said, this is mostly meant to support a clean shutdown of a process which is already going to shut down anyway.

In detail: in Sage, we run GAP inside a pseudo-tty (pty). When we shut down that pty, the stdin/stdout/stderr streams are all closed. Whenever that happens, GAP will exit anyway because stdin is closed. Now if GAP tries to write anything to stdout/stderr before actually shutting down, we end up in this infinite error loop.

jdemeyer:
> What exactly is the need for this in libgap?

This PR is unrelated to libGAP.

jdemeyer:
Some history: this bug was first noticed in 2012 (long before anything like libGAP existed!) when upgrading Sage to GAP-4.5.6

There is a long discussion thread in the Sage tracker starting at https://trac.sagemath.org/ticket/13211#comment:58 for why this patch was needed.

Supposedly, this bug was already reported as http://tracker.gap-system.org/issues/125 (but that link no longer seems to work)

embray:
I haven't gotten around to it yet but I planned to today: I have a simpler version of this patch that will simply not go into the default error-handling code if an error occurs in writing to stdout or stderr since that will obviously result in an unwanted recursion.

Instead I believe it should just fail silently, though there is no "right" answer there.  But it would help to get out of the mindset of GAP being a command-line REPL first, but rather have low-level functionality not assume interactivity, but build interactivity on top of lower-level functionality.  The problem here is that low-level I/O code can jump directly into an interactive error-handling break loop.

For the sake of backwards-compatibility we should keep that behavior for now until we've had time to do deeper refactoring.  But an explicit case to prevent infinite recursion is still needed here.

embray:
Here's a simplified alternate fix to this, which does not explicitly rely on any recursion detection, and just focuses narrowly on the affected use cases where standard I/O cannot be successfully written to.

With this fix the example from the issue description works like:

```
$ ./gap > /dev/full
Panic in src/sysfiles.c:1558: Could not write to stdout/stderr.
$ echo $?
1
```

The fact that it causes a "panic" is fine for the time being, since the point is to cause the program to exit cleanly.  `Panic()` as currently implemented is a problem from the perspective of the libgap API since it forces the program to exit, but right now the use-case this issue is concerned about is just spawning gap as a subprocess.

If stderr is also redirected to a broken file, then the Panic message won't be displayed either, but that's sort of a given:

```
$ ./gap > /dev/full 2>&1
$ echo $?
1
```

ChrisJefferson:
I think this is fine for merging, the x is just to coveralls confusion.

embray:
@ChrisJefferson Can you please add backport-4.10 to this as well?

fingolfin:
Backported to stable-4.10 in aef2214c886bcd69a6c43d9404f93cd41563e3c1

jdemeyer:
I'm not convinced by this patch because there are still plenty of ways how error handling could run into an infinite loop (given that `ErrorInner` is actually a GAP function which is customizable and can generate an error in plenty of ways). I prefer the original version with the explicit check for recursion.

fingolfin:
@jdemeyer I have no idea what the original version of this PR looked like, and I don't think there is a way to let GitHub show it to me. So it'd be best if you opened a fresh PR (with a references to this one) for further discussion. Or, if that's non-trivial work, at least open a new issue (this PR is closed and most people won't see a discussion here), with a link to some version of the patch you are talking about.

ChrisJefferson:
@jdemeyer : I don't mind looking at different ones to this, but I am fundementally opposed to any patch which would discard errors, if that is the PR you want to bring back (of course, I don't have solo veto power, but I really don't like it).

embray:
@jdemeyer It is well known that there are lots of ways you can get infinite loops in GAP's error handling.  It's also been discussed on other issues that what is really needed here is a more thorough overhaul of GAP's error handling (for which I have some ideas, as I'm sure do you).

The reason we didn't use the original version of this patch is precisely because even it was misleading in the extent to which it purposed to fix infinite loops in GAP's error handling (it doesn't).  It also wasn't thread safe in any way.

Issue #3100, opened 2018-12-08
Buggy GroupHomomorphismByImages for f.p. groups


laurentbartholdi:
GAP can create a homomorphism, check that it's bijective, but the inverse map is not a homomorphism.

This seems to be related to the complexity of the groups involved. If I replace `q` by `SimplifiedFpGroup(q)` then everything works.

```
gap> f := FreeGroup(18);;
gap> AssignGeneratorVariables(f);
#I  Assigned the global variables [ f1, f2, f3, f4, f5, f6, f7, f8, f9, f10, f11, f12, f13, f14, f15, f16, f17, f18 ]
gap> q := f / [ f1*f3*f7, f2*f6*f8, f4*f18*f14, f5*f13*f9, f10*f12*f16, f11*f15*f17, f1*f6, f2*f7, f3*f18, f4*f13, f5*f8, f9*f12, f10*f15, f11*f16, f14*f17 ];
<fp group of size infinity on the generators [ f1, f2, f3, f4, f5, f6, f7, f8, f9, f10, f11, f12,
  f13, f14, f15, f16, f17, f18 ]>
gap> AssignGeneratorVariables(q);
#I  Global variable `f1' is already defined and will be overwritten
#I  Global variable `f2' is already defined and will be overwritten
#I  Global variable `f3' is already defined and will be overwritten
#I  Global variable `f4' is already defined and will be overwritten
#I  Global variable `f5' is already defined and will be overwritten
#I  Global variable `f6' is already defined and will be overwritten
#I  Global variable `f7' is already defined and will be overwritten
#I  Global variable `f8' is already defined and will be overwritten
#I  Global variable `f9' is already defined and will be overwritten
#I  Global variable `f10' is already defined and will be overwritten
#I  Global variable `f11' is already defined and will be overwritten
#I  Global variable `f12' is already defined and will be overwritten
#I  Global variable `f13' is already defined and will be overwritten
#I  Global variable `f14' is already defined and will be overwritten
#I  Global variable `f15' is already defined and will be overwritten
#I  Global variable `f16' is already defined and will be overwritten
#I  Global variable `f17' is already defined and will be overwritten
#I  Global variable `f18' is already defined and will be overwritten
#I  Assigned the global variables [ f1, f2, f3, f4, f5, f6, f7, f8, f9, f10, f11, f12, f13, f14, f15, f16, f17, f18 ]
[...]

PR #3103, opened 2018-12-11
Two fixes for fp group homomorphisms and Improvement for Fitting free matrix group setup

### LINKED ISSUE #3097: Too general "Order" for automorphisms
```
gap> f := FreeGroup(2);;
gap> x := GroupHomomorphismByImages(f,f,[f.1,f.2],[f.2,f.1]);;
gap> Order(x); # expects 2
Error, no 3rd choice method found for `CompositionSeries' on 1 arguments at /usr/local/src/gap/lib/methsel2.g:250 called from
CompositionSeries( U ) at /usr/local/src/gap/lib/grpprmcs.gi:2690 called from
ChiefSeries( d ) at /usr/local/src/gap/lib/morpheus.gi:43 called from
```

the code in lib/morpheus.gi assumes that the source of the automorphism is a finite group. Presumably the method should give up when its source doesn't have the filter `IsFinite`.

hulpke:
This fixes #3097, #3100 and a nuisance error when working with matrix groups under the fitting free setup (avoid `RestrictedMapping` which runs into trouble).

### DIFF
--- a/lib/fitfree.gi
+++ b/lib/fitfree.gi
@@ -88,7 +88,11 @@ local cache,ffs,pcisom,rest,it,kpc,k,x,ker,r;
 	    serdepths:=List(ffs.depths,y->First([1..Length(r)],x->r[x]>=y))
 	    );
   Add(cache,[ffs,r]); # keep
-  SetSize(U,Product(RelativeOrders(k))*Size(Image(rest)));
+  if Length(k)=0 then
+    SetSize(U,Size(Image(rest)));
+  else
+    SetSize(U,Product(RelativeOrders(k))*Size(Image(rest)));
+  fi;
   return r;
 
 end);
--- a/lib/fitfree.gi
+++ b/lib/fitfree.gi
@@ -31,7 +31,21 @@ local cache,ffs,pcisom,rest,it,kpc,k,x,ker,r;
 
   pcisom:=ffs.pcisom;
 
-  rest:=RestrictedMapping(ffs.factorhom,U);
+  #rest:=RestrictedMapping(ffs.factorhom,U);
+  if IsPermGroup(U) and AssertionLevel()>1 then
+    rest:=GroupHomomorphismByImages(U,Range(ffs.factorhom),GeneratorsOfGroup(U),
+      List(GeneratorsOfGroup(U),x->ImagesRepresentative(ffs.factorhom,x)));
+  else
+    RUN_IN_GGMBI:=true; # hack to skip Nice treatment
+    rest:=GroupHomomorphismByImagesNC(U,Range(ffs.factorhom),GeneratorsOfGroup(U),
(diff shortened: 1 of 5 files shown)


laurentbartholdi:
You write 2x
"
gi:=MappingGeneratorsImages(hom)[2]{ListPerm(PermList(hom!.genpositions)^-1,
Length(hom!.genpositions))};". .
I think it would be much more efficient as

gi{hom!.genpositions} := MappingGeneratorsImages (hom)[2];

I.e. without creating extra permutations and lists.

lib/fitfree.gi
```diff
@@ -28,7 +28,17 @@ local cache,ffs,pcisom,rest,it,kpc,k,x,ker,r;
 
   pcisom:=ffs.pcisom;
 
-  rest:=RestrictedMapping(ffs.factorhom,U);
+  #rest:=RestrictedMapping(ffs.factorhom,U);
+  if IsPermGroup(U) and AssertionLevel()>1 then
```
fingolfin: The commit message says:
>  using `RestrictedMapping` as it will not work for certain matrix group homomorphisms

So, does that mean that there is a bug in `RestrictedMapping`? Or is this more a general problem where e.g. `RestrictedMapping` doesn't have enough information to do something sensible in all cases? Or perhaps a third option?

Also, what exactly is a "matrix group homomorphisms" in this context? A hom with a matrix group as source? Range? Both? Either? And, based on that, I'd have expected that only these kinds of homs would be treated separately here, and the rest handled by `RestrictedMapping` anyway, but that's not the case. I am sure there is a good reason, but what is it?

In any case, IMHO there should be a comment here that briefly explains these things.

lib/fitfree.gi
```diff
@@ -28,7 +28,17 @@ local cache,ffs,pcisom,rest,it,kpc,k,x,ker,r;
...
+    rest:=GroupHomomorphismByImages(U,Range(ffs.factorhom),GeneratorsOfGroup(U),
+      List(GeneratorsOfGroup(U),x->ImagesRepresentative(ffs.factorhom,x)));
+  else
+    RUN_IN_GGMBI:=true; # hack to skip Nice treatment
+    rest:=GroupHomomorphismByImagesNC(U,Range(ffs.factorhom),GeneratorsOfGroup(U),
+      List(GeneratorsOfGroup(U),x->ImagesRepresentative(ffs.factorhom,x)));
+    RUN_IN_GGMBI:=false;
+  fi;
+  if rest=fail then Error("can't build homomorphism"); fi;
```
fingolfin: So when would this ever fail? Can we perhaps give a slightly more helpful error message?

hulpke: I've replaced the error with an assertion.

lib/fitfree.gi
hulpke: Changed text.

hulpke:
@laurentbartholdi 
Thank you -- changed.

PR #3104, opened 2018-12-12
generalize `IntegratedStraightLineProgram`

ThomasBreuer:
- extended `IntegratedStraightLineProgram` to the situation
  that some of the input programs return lists of elements,

- adjusted the documentation of `IntegratedStraightLineProgram`
  accordingly,

- mention in the documentation of `StraightLineProgram`
  that the given list of lines or string must be nonempty
  (the code has never allowed empty input data)


ChrisJefferson:
I'm always tempted to ask "could there be some tests". While I realise there is an example, does it cover all the new cases?

I realise this request might seem annoying for fairly small change.

PR #3109, opened 2018-12-14
kernel: RequireArguments: make <...> in argnames explicit; add RequireFilter to opers.c 

fingolfin:
The first commit changes calls like:

    RequireArgument("CopyToStringRep", obj, "string", "must be a string");

to this:

    RequireArgument("CopyToStringRep", obj, "<string>", "must be a string");

This change makes `RequireArgument` and the derived macros and functions more
flexible: in cases where there is no clear argument name (say inside a method
for addition), we can instead of `"<arg>"` write something more expressive, like
`"the left operand"`.

The second commit adds a new `RequireFilter` helper to `opers.c`, with a tighter check than we used before. Previously, we accepted this code, now it results in an error (again, as it
did in GAP 4.9):

    SymmetricGroupCons(SymmetricGroupCons, 2);

The second commit makes use of the increased flexibility introduced by the first commit.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -84,9 +84,9 @@
 #include "set.h"
 
 
-#define RequireBlist(funcname, op, argname) \
-    RequireArgumentCondition(funcname, op, argname, IsBlistConv(op), \
-        "must be a boolean list")
+#define RequireBlist(funcname, op, argname)                                  \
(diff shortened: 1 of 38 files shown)


fingolfin:
I have now rebased this to deal with code drift since this PR was made, and to re-run all the tests.

Issue #3055, opened 2018-11-25
Rare no-method-found error in sanity.g in stable-4.10 branch


alex-konovalov:
Happened yesterday in https://gap-ci.cs.st-andrews.ac.uk/view/GAP-weekly/job/GAP-stable-sanity/GAPCOPTS=32build,label=kovacs/11/console (link only accessible from St Andrews):
```
testing group t20n971
conjugacy classes
rational classes
automorphism group
testing group Group( [ f1, f2, f3, f4, f5, f6, f7 ] )
subgroups
normal subgroups
conjugacy classes
rational classes
automorphism group
20,971,192,136
testing group t20n972
conjugacy classes
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `+' on 2 arguments
The 1st argument is 'fail' which might point to an earlier problem
 at /circa/scratch/gap-jenkins/workspace/GAP-stable-sanity/GAPCOPTS/32build/la\
bel/kovacs/GAP-stable-snapshot/lib/methsel2.g:250 called from
Build step 'Execute shell' marked build as failure
Sending e-mails to: alexander.konovalov@st-andrews.ac.uk
Finished: FAILURE
```
Was able to reproduce this locally in stable-4.10 branch, starting GAP with `-r -A` and calling
```
Read("tst/sanity.g");
for i in [1..100] do Print(i,"\n");TestsForGroup(TransitiveGroup(20,972));od;
```
but it took several minutes and 59 iterations:
```
59
testing group t20n972
conjugacy classes
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `+' on 2 arguments
The 1st argument is 'fail' which might point to an earlier problem
 at /Users/alexk/GITREPS/gap-stable-4.10/lib/methsel2.g:250 called from
First( dsz, function ( i )
        return (i >= minlen);
    end ) - minlen at /Users/alexk/GITREPS/gap-stable-4.10/lib/clashom.gi:742 called from
ConjugacyClassesSubwreath( F, FM, n, S1, Action( FM, components[1] ), T1, 
 components, emb, proj 
 ) at /Users/alexk/GITREPS/gap-stable-4.10/lib/clashom.gi:1339 called from
ConjugacyClassesFittingFreeGroup( G 
 ) at /Users/alexk/GITREPS/gap-stable-4.10/lib/clashom.gi:2363 called from
ConjugacyClassesViaRadical( G 
[...]

PR #3110, opened 2018-12-14
Workaround for error #3055 and fix for GQuotient

hulpke:
This is a bug that arises very rarely, maybe based on random
selections. In this case restart calculations.

(The bug was reported in #3055)

Also fix for GQuotient running slow in certain cases of groups with infinite quotients.

### DIFF
--- a/tst/testbugfix/2018-12-17-gquotient.tst
+++ b/tst/testbugfix/2018-12-17-gquotient.tst
@@ -1,0 +1,7 @@
+# reported by Giles Gardam. The code for eliminating element orders can run astray if 
+# a quotent by a generator power is cyclic, but also has cyclic subgroups of infinite
+# index.
+gap> F := FreeGroup("a", "b", "c");;
+gap> G := F/ParseRelators(F,"a2b5a2b2C2,a5c3B2");;
+gap> Length(GQuotients(G,AlternatingGroup(5)));
+1
--- a/lib/clashom.gi
+++ b/lib/clashom.gi
@@ -738,6 +738,11 @@ local clT,	# classes T
 			  dsz:=DivisorsInt(scj);
 			fi;
 			while not minlen in dsz do
+                          # workaround rare problem -- try again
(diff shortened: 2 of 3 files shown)


alex-konovalov:
Thanks @hulpke! I've edited the description to add #3055 to the text - otherwise there is no visible link  to click to navigate to it (one should either do it manually or click trough the commit message).

tst/testbugfix/2018-12-17-gquotient.tst
```diff
@@ -0,0 +1,7 @@
+# reported by Giles Gardam. The code for eliminating element orders can run astray if 
+# a quotent by a generator power is cyclic, but also has cyclic subgroups of infinite
+# index.
+gap> F := FreeGroup("a", "b", "c");;
+gap> G := F/ParseRelators(F,"a2b5a2b2C2,a5c3B2");;
+gap> Length(GQuotients(G,AlternatingGroup(5)));
```
fingolfin: This tests was running for several minutes on my machine, then run out of memory.

That's not acceptable for tst/testbugfix, whch we run as part of our CI on every pull request. It could be added to `tst/testextra`, though (perhaps in a new dir `tst/testextra/bugfix`)

hulpke: Let me check. This example should run in under a second. Probably it is the assertions again...

hulpke: Dear @fingolfin 
That is really strange. I just rebased the branch, recompiled and ran the file with `Test`, and it took indeed under a second. Also, in the travis log I get

    testing: /home/travis/build/gap-system/gap/tst/testbugfix/2018-12-17-gquotient.tst
     923 ms (24 ms GC) and 28.7MB allocated for 2018-12-17-gquotient.tst

which is the same magnitude.
Is it possible that you are loading something else that interferes? Could it be the Julia issue Thomas was observing?

PR #3111, opened 2018-12-14
kernel: remove `environ` arguments, remove sysenviron

fingolfin:
This is an alternative to PR #3106 by @embray based on the discussion there.

Closes #3106

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -35,6 +35,7 @@
 #include "stats.h"   // for ClearError
 #include "streams.h"
 #include "stringobj.h"
+#include "sysenv.h"
 #include "sysfiles.h"
 #include "sysmem.h"
 #include "sysopt.h"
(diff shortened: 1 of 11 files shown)


embray:
Ah, thanks for taking care of this.  It's fine by me; I will shed no tears if we just do away with the environ passing entirely.

fingolfin:
I was motivated to do this because we just run into an issue with the GAPJulia bridge due to us being lazy and not passing the right `environ` value to libgap... fixing that would of course have been possible, but still rather annoying. So I took this as yet another sign that we should not pass environ via the API...

embray:
Indeed: Most of the points I've raised so far are in no way specific to Sage or even Python.  The segfault we found could have happened to anyone using `GAP_Initialize()`.  Whether or not it happens is mostly a matter of luck.

embray:
It just as easily might not have. This code path is already technically covered by the scant existing libgap tests. Like I said, whether or not this problem actually occurs depends on some very particular circumstances.

fingolfin:
Backported via 58670125c36257b9743cbb4fb9ea584198159f8b

PR #3122, opened 2018-12-18
Double escape backslash in make_doc.in

frankluebeck:
The backslash needs double escape because of shell. (Bad title page noticed by Bill Allombert.)


alex-konovalov:
Backported in 663a51a77c4ea888201be3284bcf93e22a72c668

PR #3129, opened 2018-12-20
extend the default `BrauerTableOp` method

ThomasBreuer:
- extended the default `BrauerTableOp` method to the case
  of cyclic defect such that all Brauer characters lift
  to characteristic zero (this case can be detected from the
  ordinary character table)

- extended the documentation accordingly

- extended the tests accordingly

### DIFF
--- a/lib/ctbl.gd
+++ b/lib/ctbl.gd
@@ -3102,10 +3102,33 @@ DeclareOperation( "CharacterTable", [ IsString ] );
 ##  returns its <A>p</A>-modular
 ##  character table if &GAP; can compute this table, and <K>fail</K>
 ##  otherwise.
-##  The <A>p</A>-modular table can be computed for <A>p</A>-solvable groups
-##  (using the Fong-Swan Theorem) and in the case that <A>ordtbl</A> is a
-##  table from the &GAP; character table library for which also the
-##  <A>p</A>-modular table is contained in the table library.
+##  <P/>
+##  The <A>p</A>-modular table can be computed in the following cases.
+##  <P/>
+##  <List>
+##  <Item>
+##    The group is <A>p</A>-solvable (see <Ref Oper="IsPSolvable"/>,
+##    apply the Fong-Swan Theorem);
+##  </Item>
+##  <Item>
+##    the Sylow <A>p</A>-subgroup of <A>G</A> is cyclic,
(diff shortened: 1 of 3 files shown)


lib/ctbl.gd
```diff
@@ -3102,10 +3102,33 @@ DeclareOperation( "CharacterTable", [ IsString ] );
...
+##  The <A>p</A>-modular table can be computed in the following cases.
+##  <P/>
+##  <List>
+##  <Item>
+##    The group is <A>p</A>-solvable (see <Ref Oper="IsPSolvable"/>,
+##    apply the Fong-Swan Theorem);
+##  </Item>
+##  <Item>
+##    the Sylow <A>p</A>-subgroup of <A>G</A> is cyclic,
+##    and all <A>p</A> modular Brauer characters of <A>G</A>
```
alex-konovalov: Missing `-` in `<A>p</A>-modular`

lib/ctbl.gi
```diff
@@ -4202,6 +4203,52 @@ InstallMethod( BrauerTableOp,
...
       fi;
+    else
+      ppart:= 1;
+      n:= Size( tbl );
+      while n mod p = 0 do
+        ppart:= ppart * p;
+        n:= n / p;
+      od;
+      if ppart in OrdersClassRepresentatives( tbl ) then
+        # The Sylow 'p'-subgroup is cyclic.
```
alex-konovalov: I suggest to remove quotes around `p` to ease reading.

alex-konovalov: @ThomasBreuer just to explain why: I read that first as `p'-subgroup`, then reread to figure out what does that mean. The style you use here is not consistent with other styles used in the GAP library: for example we have

```
grp.gi:      # compute generators for the torsion Omega p-subgroups of the center
grpprmcs.gi:##  O_p(G), the p-core of G, is the maximal normal p-subgroup
grpprmcs.gi:                        # abelian normal p-subgroup of G
grpprmcs.gi:    # p-subgroup of G; taking commutator subgroups, find abelian normal
grpprmcs.gi:    # p-subgroup of G.
grpprmcs.gi:    # p-subgroup; kernel is abelian p-group. Take image at this action, and
```
and also one case of
```
ctblsolv.gi:    # Choose a normal elementary abelian `p'-subgroup `N',
```

Issue #739, opened 2016-04-15
Differences in `Immutable()` semantics between GAP and HPC-GAP.


rbehrends:
The current implementation of `Immutable()` in HPC-GAP will actually create copies of subobjects that are mathematically immutable, but aren't physically immutable (i.e. can be written by the current thread). This does not happen for the main object, but it can happen for (e.g.) the immutable members of a mutable list if one were to call `Immutable()` on that list.

While the fix for this would be relatively straightforward (only the boundary condition for the search in `IsWritableOrImmutable()` in `traverse.c` would have to be updated), it appears to me that we may want to support two different functionalities for this:

* The original `Immutable()` semantics.
* A version of `Immutable()` where the result is in the current region (i.e. largely what HPCGAP does now) to ensure that mathematically immutable, but physically mutable objects are accessible.

fingolfin:
Sounds good to me. To say it explicitly: for (non-HPC) GAP, the new version of `Immutable()` would just be an alias to `Immutable()`.

So what should the new `Immutable` be called? 

Will there also be a sibling to `MakeImmutable()`?

rbehrends:
This would not make sense for `MakeImmutable()`, because `MakeImmutable()` does not copy objects.

As for naming, I would like it if we could establish a general terminology for "mathematically immutable" and "having an immutable representation", as this problem crops up in multiple contexts. In computer science, one usually contrasts "abstract state" with "concrete state" and you could then meaningfully talk about "immutable abstract state" and "immutable concrete state"; however, that's quite a mouthful, so a more compact term for "immutable concrete state" would be desirable, one that allows us to keep using "immutable" as *the* term for "immutable abstract state".

Issue #502, opened 2016-01-20
MakeReadOnly doesn't throw an error accessing objects wo read permission


ssiccha:
MakeReadOnly should throw an error here:
```gap
gap> list := [ ];;
gap> list := ShareObj( list );;
gap> MakeReadOnly(list);
<protected object in shared region 0x3a4cd80 (id: 41664768)>
```

rbehrends:
The reason why it doesn't result in an error is that the traversal routine ends at the boundaries of regions that the current thread can write to. This is normally  a feature (so that it works on objects with shared or immutable sub-objects), but maybe this should not happen when the primary object is already ineligible.

Issue #58, opened 2015-03-17
Error, MakeImmutable: Argument has inaccessible subobjects


mohamed-barakat:
MakeImmutable breaks the backwards compatibility of HPCGAP with GAP4:

```
gap> LoadPackage( "MatricesForHomalg" );
true
gap> ZZ := HomalgRingOfIntegers( );
Z
gap> m := HomalgInitialMatrix( 2, 2, ZZ );
<An initial 2 x 2 matrix over an internal ring>
gap> InfoOfObject( m );
rec( attributes := rec( NrColumns := 2, NrRows := 2 ), components := rec( ring := Z ),
  object := <An initial 2 x 2 matrix over an internal ring>, properties := rec( IsEmptyMatrix := false, IsInitialMatrix := true ) )
gap> MakeImmutable( m );
Error, MakeImmutable: Argument has inaccessible subobjects
not in any function at line 5 of stream
```

fingolfin:
Actually, `m` in this situation is already immutable. And then `MakeImmutable` should be an idempotent. But it isn't, due to checks performed in the kernel. Specifically, `PreMakeImmutableCheck` checks whether we have exclusive write access to the object (which we don't, as it lives in the public region, which is fine, given that it is immutable).

I guess this means the kernel should first check whether the object already is immutable; if it is, do nothing and return.

fingolfin:
One fix would be to modify `CheckedMakeImmutable` (which generates the error we are seeing here) to call `IS_MUTABLE_OBJ`, like it sibling `MakeImmutable` does. But I am not sure if this would have bad side effects... :-/

@rbehrends Any comments?

rbehrends:
The problem is slightly more complicated. `PreMakeImmutableCheck()` already tests if the top-level object is immutable, and `CheckedMakeImmutable()` then just calls `MakeImmutable()` if the check passes.

What we're seeing seems to happen (I think) because `m` is an atomic component object and mutability here implies that `RESET_FILTER_OBJ()` needs to be called (same for atomic positional objects). We need to special-case these tnums both in the traversal logic and in the `MakeImmutable()` table (so that the filter update is actually thread-safe in case two threads try to do this simultaneously).

A related problem that remains is how to handle `MakeImmutable()` for `T_DATOBJ`; we've discussed this before: a possible solution would be to use `MakeReadOnly()` underneath, but that would require some additional hackage in the kernel.

stevelinton:
[...]

Issue #47, opened 2015-03-16
Matrix corrupted by MakeReadOnly


alex-konovalov:
The following currently happens in HPC-GAP:
```
gap> m:=[[1,2],[3,4]];
[ [ 1, 2 ], [ 3, 4 ] ]
gap>  IsMatrix(m);
true
gap>  MakeReadOnly(m);
[ [ 1, 2 ], [ 3, 4 ] ]
gap>  IsMatrix(m);
false
```

stevelinton:
I think we've seen this before. It's not strictly corruption, but a clash between the mechanisms that work out the type of plain lists and the fact that the top-level list of this matrix is read-only.

stevelinton:
I understand the problem now. The kernel code that checks the type of a plain list has to allow for the possibility that the list is loopy -- that is that it appears as a (nested) element of itself. To do this it has to "mark" the list and its elements as it explores them to know if it has seen them before. Otherwise thre traversal might be inifnite. This is impossible for readonly lists, so instead we don't do the traversal at all, but just rely on whatever stored information we have about type type of the list. Since this list has mutable elements we can't stpre that it is table.

Probably the best fix is to rewrite the plain list typing machinery using Reimers object traversal mechanisms. Also be aware in library programming that matrices may not know that they are matrices, which is true anyway,

stevelinton:
Following further discussion with Reimer, we can do some fixes, but checking IsMatrix for some lists will be slow (and slow every time not just the first time).  The solution to that is to use Matrix Objects.

PR #3136, opened 2018-12-21
MakeReadOnlyObj makes objects immutable in GAP

### LINKED ISSUE #3132: In GAP MakeReadOnlyObj doesn't make objects immutable
Note that this is not in HPC-GAP.

In GAP I can do:
```
gap> l := [];;
gap> MakeReadOnlyObj(l);;
gap> l[1] := 1;;
gap> l;
[ 1 ]
```

I would find it more sensible if `MakeReadOnlyObj` makes objects immutable in GAP. At least until there is a proper `MakeReadOnlyObj` function in GAP, if we ever want one.

ssiccha:
Then code that is supposed to work both under GAP and HPC-GAP can
directly use MakeReadOnlyObj.

Fixes #3132.

### DIFF
--- a/lib/hpc/thread1.g
+++ b/lib/hpc/thread1.g
@@ -20,7 +20,7 @@
 # Mock functions from thread1.g
 
 BIND_GLOBAL("MakeThreadLocal", ID_FUNC);
-BIND_GLOBAL("MakeReadOnlyObj", ID_FUNC);
+BIND_GLOBAL("MakeReadOnlyObj", MakeImmutable);
 BIND_GLOBAL("MakeReadOnlyRaw", ID_FUNC);
 BIND_GLOBAL("MakeReadOnlySingleObj", ID_FUNC);
 


lib/hpc/thread1.g
```diff
@@ -20,7 +20,8 @@
 # Mock functions from thread1.g
 
 BIND_GLOBAL("MakeThreadLocal", ID_FUNC);
-BIND_GLOBAL("MakeReadOnlyObj", ID_FUNC);
+# HACK: improve if at some point there exists proper read-only support in GAP
```
fingolfin: I find this comment more confusing than helpful. Perhaps it meant to say: "remove this if the read-only support in HPC-GAP is ported to GAP"? But then this is obvious.

So unless you intended some other meaning (in that case: please elaborate),  I'd just remove this comment.

ssiccha: No, this is exactly what I meant. I added the comment since I wanted to make clear that this is a hack since read-only objects in HPC-GAP work differently than immutable ones in GAP. That makes more sense in the commit message, so I'll put it there.

lib/hpc/thread1.g
```diff
@@ -20,7 +20,7 @@
 # Mock functions from thread1.g
 
 BIND_GLOBAL("MakeThreadLocal", ID_FUNC);
-BIND_GLOBAL("MakeReadOnlyObj", ID_FUNC);
+BIND_GLOBAL("MakeReadOnlyObj", MakeImmutable);
```
fingolfin: I have some doubts whether this is correct. From the documentation:
> The MakeReadOnlyObj function migrates obj and all its subobjects that are within the same region as obj to the read-only region.

This suggests that if there are subobjects from different regions, not all will be made read only.

Also, does this command in HPC-GAP really make objects immutable, too? That would not fit with this text in the manual either.

I have not yet checked what the code actually does, though...

ssiccha: > This suggests that if there are subobjects from different regions, not all will be made read only.

If you interpret it so that GAP has only one region anyways, that should be fine.

fingolfin: But "immutable" implies "all subobjects are immutable", while read-only does not imply that. Hence read-only is not stricter than immutable!

Yes, one could argue that in GAP, there is only one region. But still, the effects of MakeImmutable and MakeReadOnly are not the same, and I am not sure whether it's really a good idea to mix them up like this: while it might be useful in some cases, it might be super confusing in others.

Can you give an example of two of code where this would be useful to you?

ssiccha: > But "immutable" implies "all subobjects are immutable", while read-only does not imply that. Hence read-only is not stricter than immutable!
You're right! I missed that. I was thinking only about calling `MakeReadOnlyObj` when I was talking about "making an object read only". Since `MakeReadOnlyObj` also makes all of its sub-objects read only, calling `MakeReadOnlyObj` is stricter than calling `MakeImmutable`.

An example of code are the recent commits in the repository https://github.com/lbfm-rwth/GaussPar/ e.g. commit https://github.com/lbfm-rwth/GaussPar/commit/de3a42ef9193f23e0d604c6362e74b48e799ee1a (the implementation was mostly a proof of concept and we're now trying to clean everything up). We want this code to work both in GAP and in HPC-GAP. I had the impression that the best way to ensure that nobody can "mess" with a variable is to call `MakeReadOnlyObj(MakeImmutable(..))` on it, since `MakeReadOnlyObj` currently does nothing in GAP.

I think there are two alternatives:
- `MakeReadOnlyObj` stays as it is. Then we'll have to stick with `MakeReadOnlyObj(MakeImmutable(..))`
- `MakeReadOnlyObj` calls `MakeImmutable` in GAP, but `MakeReadOnlySingleObj` would have to stay with the trivial mock-up it has now.


I hope I'm not opening a can of worms here, but could something like `BindConstant` be used to implement the HPC-GAP "read only" concept in GAP? (I have now idea how `BindConstant` works in the kernel)

fingolfin: > Since `MakeReadOnlyObj` also makes all of its sub-objects read only

But that's simply not correct, as I pointed out before. And now I had a look at the code, and verified that indeed only subobjects in the same region are treated. As a consequence: if the object is already in the read-only region, then `MakeReadOnlyObj` exits immediately, without checking any subobjects.

But OK, sure, you can argue that "in GAP, there is only one region, so this is true". But the fact remains: "read only" objects need not be immutable! Indeed, I just verified this in HPC-GAP:
```
gap> foo:=[1,2,3];
[ 1, 2, 3 ]
gap> IsMutable(foo);
true
gap> MakeReadOnlyObj(foo);
[ 1, 2, 3 ]
gap> IsMutable(foo);
true
```

Thus in your example, IMHO the right approach is to use `MakeReadOnlyObj(MakeImmutable(..))`. If you only use `MakeReadOnlyObj(..)`, then with this PR, you'll get different behavior between GAP and HPC-GAP.

Now, for *external* objects, GAP contains code which ensures that "read-only" implies "immutable". Perhaps this implication should extended to *all* objects. If that was done, this PR becomes somewhat less problematic; but then we'd still have a discrepancy when it comes to objects with subobjects in different regions (and of course then we run into various other problems, see e.g. #47, #58, #502, #739)

ssiccha: > But OK, sure, you can argue that "in GAP, there is only one region, so this is true". 

Yes sorry, what I wrote is nonsense in general. I meant "MakeReadOnlyObj makes all of its sub-objects read only if they lie in the same region, which I would argue always is the case in GAP". But then letting `MakeReadOnlyObj` in GAP make objects immutable would only work as long as it is not possible to make an object, but not its subobjects, read-only via `MakeReadOnlySingleObj`.


> Now, for external objects, GAP contains code which ensures that "read-only" implies "immutable". Perhaps this implication should extended to all objects. 

This would be incompatible with the following use case of `MakeReadOnlySingleObj`:
```
list := List([1..10], x -> AtomicList([]));
MakeReadOnlySingleObj(list);
```
On the one hand, if `list` became immutable then it would be impossible to write into the `AtomicList([])` subobjects. On the other hand `MakeImmutable(AtomicList([]));` yields `Error, MakeImmutable: Argument has inaccessible subobjects`. (Now that I think about it, in this example it doesn't matter whether we use `MakeReadOnlySingleObj` or `MakeReadOnlyObj` since the atomic lists live in a different, namely the public, region.) So I think it wouldn't be a good idea to let read-only imply immutable in HPC-GAP.


> Thus in your example, IMHO the right approach is to use MakeReadOnlyObj(MakeImmutable(..)). If you only use MakeReadOnlyObj(..), then with this PR, you'll get different behavior between GAP and HPC-GAP.

I see, the differing behaviour would be that with this PR in GAP `IsMutable` would always return false while in HPC-GAP it would return true for e.g. `IsMutable(MakeReadOnlyObj([1,2]))`. But with this PR people will get used to the fact that "read-only implies immutable" and will be confused when this doesn't hold in HPC-GAP.


All in all, thanks for the nice discussion :) and I think instead of merging this PR there should rather be a proper implementation of the read-only functionality in GAP (if this were wanted and also only if HPC-GAP survives at all).

Should I close this PR now?

PR #3133, opened 2018-12-21
Enable Coveralls submissions

fingolfin:
* disable a bunch of travis tests for quicker feedback
* still need to add C/C++ coverage upload
* results will be visible on https://coveralls.io/github/gap-system/gap

PR #3138, opened 2018-12-21
Add gcov support for coveralls.io as well

markuspf:
This builds on #3133 and adds coveralls submissions for the C and C++ code (generated by gcov). Work in progress as I am fighitng with travis.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,10 +12,11 @@ environment:
   matrix:
     - CYG_ARCH: x86
       CYG_ROOT: C:\cygwin
+      PACKAGES: "-P python-pip"
       ABI: 32
     - CYG_ARCH: x86_64
       CYG_ROOT: C:\cygwin64
-      PACKAGES: "-P libgmp-devel,zlib-devel"
+      PACKAGES: "-P libgmp-devel,zlib-devel,python-pip"
       CFLAGS: "--coverage"
       LDFLAGS: "--coverage"
 # FIXME: HPC-GAP build on AppVeyor disabled for now, as it crashes
@@ -39,6 +40,10 @@ build_script:
         Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
         throw "There are newer queued builds for this pull request, failing early." }
   - SET "PATH=%CYG_ROOT%\bin;%PATH%"
+  # HACK: test to see whether this prevents segfaults on 32 bit builds.
+  - bash -lc "python -m pip install --upgrade pip"
+  # gcovr install done here so it uses cygwin python we installed
+  - bash -lc "python -m pip install gcovr"
   - bash -lc "cd $APPVEYOR_BUILD_FOLDER && ./etc/ci-prepare.sh"
 
 test_script:
--- a/etc/ci-coveralls-merge.py
+++ b/etc/ci-coveralls-merge.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+
+# Hacked up merger for testing
+
+import json
+import os
+import os.path
+import sys
+
+coverage_files = [ 'c-coveralls.json' ]
+
+print("Merging coveralls json coverage") 
+
+# Special-cased gap-coveralls.json, because we rely
+# on GAP to set the correct service_name, pull-request number
+# etc
+print "file: gap-coveralls.json", # python2 avoiding line break.
+if os.path.isfile('gap-coveralls.json'):
+    with open('gap-coveralls.json', 'r') as f:
+        merged = json.load(f)
+    print(" done.")
+else:
+    print("WARNING: could not find gap-coveralls.json, quitting")
+    sys.exit(0)
+
+if not 'source_files' in merged:
+    print("warning: source_files not in JSON dictionary read from gap-coveralls.json")
+    merged['source_files'] = []
+
+for fn in coverage_files:
+    print "file: %s" % (fn,), # python2 avoiding line break.
+    if os.path.isfile(fn):
+        with open(fn, 'r') as f:
+            cover = json.load(f)
+            merged['source_files'].extend(cover['source_files'])
+        print(" done.")
+    else:
+        print(" not found.")
+
+if len(merged['source_files']) > 0:
+    print("Merged:")
+    for k in merged.keys():
+        if (k != 'source_files'):
+            print("%s: %s" % (k, merged[k]))   
+    print("Writing merged profiles to merged-coveralls.json")
+    with open('merged-coveralls.json', 'w+') as dump:
+        json.dump(merged, dump)
+else:
+    print("No coverage found.")
+print("Done. Bye.")
--- a/etc/ci-gather-coverage.sh
+++ b/etc/ci-gather-coverage.sh
@@ -60,10 +60,8 @@ if Length(covs) > 0 then
     Print("Outputting JSON for Codecov...\n");
     OutputJsonCoverage(r, "gap-coverage.json");;
 
-    if IsBound(GAPInfo.SystemEnvironment.TRAVIS_JOB_ID) then
-        Print("Outputting JSON for Coveralls...\n");
-        OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix);
-    fi;
+    Print("Outputting JSON for Coveralls...\n");
+    OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix);
 else
     # Don't error, because we might want to gather
     # gcov coverage, so just inform that we didn't find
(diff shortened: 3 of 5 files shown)


etc/ci-gather-coverage.sh
```diff
@@ -82,3 +82,6 @@ if [[ -f gap-coveralls.json ]]
 then
   curl -F json_file=@gap-coveralls.json "https://coveralls.io/api/v1/jobs"
 fi
+
+# upload C/C++ coverage to coveralls.io
+coveralls --follow-symlinks -n -e pkg/ -e bin/ -e etc/ -e extern/ -e gen/ -e tst/
```
fingolfin: we'll also need to exclude `julia*/` for the Julia builds. Well, or I guess we could try to place the Julia binary in some other directory to hide it, like so (untested):
```patch
diff --git a/etc/ci-prepare.sh b/etc/ci-prepare.sh
index 605444ebdf..15d92ae2a1 100644
--- a/etc/ci-prepare.sh
+++ b/etc/ci-prepare.sh
@@ -30,11 +30,12 @@ fi
 if [[ $JULIA = yes ]]
 then
   # TODO: once Julia 1.1 is released, switch to stable Julia versions here?
+  pushd extern
   wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/1.1/julia-b55b85cd99-linux64.tar.gz
   #wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz
   tar xvf julia-*-linux64.tar.gz
   rm julia-*-linux64.tar.gz
-  pushd julia-*
+  cd julia-*
   JULIA_PATH=$(pwd)
   popd
   CONFIGFLAGS="--with-gc=julia --with-julia=$JULIA_PATH $CONFIGFLAGS"
```

markuspf:
Probably best to script that coveralls stuff ourselves, the cpp-coveralls thing is clsoe to useless...

markuspf:
Also noticed that `make coverage` does not actually work...

.appveyor.yml
```diff
@@ -30,6 +30,7 @@ cache:
 
 install:
   - '%CYG_ROOT%\setup-%CYG_ARCH%.exe -qnNdO -R %CYG_ROOT% -s http://cygwin.mirror.constant.com -l %CYG_ROOT%/var/cache/setup %PACKAGES%'
+  - "python.exe -m pip install gcovr"
```
fingolfin: This is not really necessary, since we can't use coveralls on AppVeyor for now anyway (starting with the limitation in the profiling package, which hard codes "travis" as CI provider right now)

fingolfin:
Fascinating: we get the same nonsense coverage results for `io.c` on Coveralls as on Codecov, see e.g. https://coveralls.io/builds/20839674/source?filename=src/io.c
 :-(

markuspf:
welp. look at `io.c` coverage now.

markuspf:
Made a token commit to kick appveyor to run the tests too (and, more crucially, pass).

fingolfin:
Nice, https://coveralls.io/builds/21000227/source?filename=src/io.c now indeed shows sensible coverage. Although none for HPC-GAP, and also various functions that ought to be covered (as some tests execute them) are not, so something is still a bit odd. But of course that's not a problem with this PR. Probably we are still not using Coveralls quite right or something like that (we still get those multiple Coveralls PR comments, too, so...)

BTW, I tried something similar (excluding extern, pkg) with Travis in the past; but never managed to get it to work.  I also carefully inspected all the coverage data uploaded to travis (by downloading it from Travis), and the maddening aspect was that the data in there did not match with the "bad" output Travis showed. *sigh*

markuspf:
To debug coverage I now run coverage locally: `gcovr` can produce html file coverage (`gcovr -r . --output out.html --exclude-directories pkg/ --html-details`), at least I can then see what kind of file is supposedly being sent to the coverage services without having to wait for travis to turn round every time...

etc/ci-coveralls-merge.py
```diff
@@ -0,0 +1,33 @@
...
+print("Merging coveralls json coverage") 
+merged = { 'service_name': 'travis-ci'
+         , 'service_job_id': os.environ['TRAVIS_JOB_ID']
+         , 'parallel': 'true'
+         , 'source_files': [] }
+
+for fn in ['gap-coveralls.json', 'c-coveralls.json']:
+    print "file: %s" % (fn,),
+    if os.path.isfile(fn):
+        f = open(fn, 'r')
```
fingolfin: It would be more "pythonesque" to use a `with` statement, here, i.e.,
```python
  with open(fn, 'r') as f:
    cover = json.load(f)
    merged['source_files'].extend(cover['source_files'])
```
And then there is no need for `f.close()`

etc/ci-coveralls-merge.py
```diff
@@ -0,0 +1,33 @@
...
+# Hacked up merger for testing
+
+import json
+import os
+import os.path
+
+print("Merging coveralls json coverage") 
+merged = { 'service_name': 'travis-ci'
+         , 'service_job_id': os.environ['TRAVIS_JOB_ID']
+         , 'parallel': 'true'
```
fingolfin: I guess we need to do even more here, and of course also something special for Travis... See also https://github.com/lemurheavy/coveralls-ruby/blob/master/lib/coveralls/configuration.rb and https://github.com/gap-packages/profiling/pull/65

Perhaps we could also do this instead (once the JSON header generated by profiling works as we desire, the above profiling PR is merged): Start with the `gap-coveralls.json` file, which should have the "correct" service keys; then "merge" c-coveralls.json into that. That should be easy enough, right?

etc/ci-coveralls-merge.py
```diff
@@ -0,0 +1,33 @@
...
+        f.close()
+        print(" done.")
+    else:
+        print(" not found.")
+	
+if len(merged['source_files']) > 0:
+    print("Writing merged profiles to merged-coveralls.json")
+    dump = open('merged-coveralls.json', 'w+')
+    json.dump(merged, dump)
+    dump.close()
```
fingolfin: again, a `with` statement would be more idiomatic:
```python
   with open('merged-coveralls.json', 'w+') as dump:
    json.dump(merged, dump)
```

etc/ci-gcovr-coverage.py
```diff
@@ -0,0 +1,642 @@
...
+                pass
+
+        files.append( { "name": options.root_filter.sub('', file)
+                    , "source_digest": md5(file)
+                    , "coverage": coverage } )
+
+    report["service_name"] = "travis-ci"
+    report["service_job_id"] = os.environ.get("TRAVIS_JOB_ID", "")
+    report["source_files"] = files
+    report["parallel"] = True
```
fingolfin: The above is insufficient for our needs. But we are ignoring these keys in the merge script anyway, so perhaps just remove the previous 4 lines, to keep things simple?

etc/ci-gather-coverage.sh
```diff
@@ -74,8 +74,21 @@ else
 fi;
 GAPInput
 
+# TODO/FIXME: for now we only support coveralls on Travis, but not e.g. on AppVeyor
+# (mainly because the coveralls support in the profiling package hardcodes Travis
+# into the JSON it generates)
+if [[ -n "$TRAVIS_JOB_ID" ]]
```
fingolfin: So, is this still valid? I guess (hope?) not, then we could instead check here for the existence of `gap-coveralls.json`

etc/ci-gather-coverage.sh
```diff
@@ -74,8 +72,21 @@ else
 fi;
 GAPInput
 
-# upload to coveralls.io
+# TODO/FIXME: for now we only support coveralls on Travis, but not e.g. on AppVeyor
+# (mainly because the coveralls support in the profiling package hardcodes Travis
+# into the JSON it generates)
```
fingolfin: This comment is also outdated, isn't it?

markuspf:
Last thing to solve for coveralls/appveyor:

```
+ [[ -f merged-coveralls.json ]]
+ curl -F json_file=@merged-coveralls.json https://coveralls.io/api/v1/jobs
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 3019k  100    72  100 3019k     72  3019k  0:00:01  0:00:01 --:--:-- 2360k
{"message":"Couldn't find a repository matching this job.","error":true}
```

markuspf:
I see we need a token in appveyor.yml...

fingolfin:
I recently switched the AppVeyor integration to the new style, so now any org admin should be able to get a token

markuspf:
I need a *coveralls* token and have so far not signed up to coveralls beacuse they want read and write access to all my repositories including the private ones...

fingolfin:
I sent you the token via riot

fingolfin:
Regarding the stupid write-to-all-repos perm, see also 
https://github.com/lemurheavy/coveralls-public/issues/550. Totally sucks (also is the reason I don't want to use Azure Pipelines for now)

markuspf:
Squashed.

Coveralls upload on Appveyor still doesn't work, and it's nigh impossible to find out why as @coveralls documentation and error messages just plain suck.

I think we should rather merge this and fix the Appveyor Coveralls uploads (or, if we want to abandon coveralls for appveyor remove the code for appveyor).

PR #3142, opened 2018-12-22
Fix Int() and Rat() for float nan, inf, -inf

### LINKED ISSUE #3135: Int(nan) attempts to create string of negative length
We can produce an object called `nan` by, for example, calling `Log10` on a negative number.  This is to be expected, but what is not expected is what happens when we put `nan` into `Int`:
```
gap> x := Log10(-1.3);
nan
gap> Int(x);
Error, NEW_STRING: Cannot create string of negative length -2147483645
not in any function at *stdin*:6
```
It seems that on the C level, we're trying to use the absolute value of `nan`, which isn't well-defined.  I'm not sure what's causing this, but probably there should be a more sensible error message.

fingolfin:
Fixes #3135

### DIFF
--- a/lib/float.gi
+++ b/lib/float.gi
@@ -513,6 +513,10 @@ InstallOtherMethod( Rat, "for floats", [ IsFloat ],
 
     local  M, a_i, i, sign, maxdenom, maxpartial;
 
+    if not IsFinite(x) then
+        Error("cannot convert float ", x, " to rational");
+    fi;
+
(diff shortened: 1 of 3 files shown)


fingolfin:
huh, why did coveralls add 5 comments? bad...

markuspf:
I suspect this is because `parallel` isn't set in the coveralls json that is submitted.

markuspf:
cf here: https://docs.coveralls.io/parallel-build-webhook

PR #3144, opened 2018-12-27
ci: enable parallel build webhook for Coveralls

fingolfin:
This should ensure that coverage results are merged by Coveralls after all
have completed.

See <https://docs.coveralls.io/parallel-build-webhook> for details.

These changes are also contained in PR #3138, but it would be useful to have them in effect *now*, while it is not yet clear when that PR will be ready.


fingolfin:
One motivation for this PR is of course the hope that this will stop coveralls from more than one coverage comments to each PR. Alas, I think we will only be able to tell if that works once we actually merge this PR.

PR #3146, opened 2018-12-28
kernel: change `static inline` to `EXPORT_INLINE`

fingolfin:
... which is #defined to `inline` in `system.h` by default, unless it is
already defined differently. And we do define it differently in one place,
namely in debug.c, where we set it to `extern inline`.

This ensures all these functions have external linkage, and are emitted at
least once as an actual function (even if all calls to them are inlined). They
thus can be referenced from clients of libgap, which so far had to manually
write wrappers for any of our static inline functions.

It also prevents the compiler from emitting multiple instances of the same
function, which can confuse lldb when debugging (as a result, I was no longer
able to use functions like TNAM_OBJ, IS_INTOBJ etc. in lldb, which made
debugging MUCH harder than it used to be; the immediate cause for this was
that we now also compile some code as C++, which resulted in two slightly
different versions of these functions to be emitted, namely one with the
original name, and one with a C++ mangled name).

Finally, we can get rid of the `debug_func_pointers` array in debug.c, whose
maintenance was a bit of a nuisance.


fingolfin:
Yes: That commit comes from PR #3125 and without it, this PR gails due to build warnings.

sebasguts:
So I suppose we should merge #3125 before?

fingolfin:
@sebasguts Indeed. Which I've done now, and I also rebased this PR.

src/system.h
```diff
@@ -45,6 +45,24 @@
...
+// wanted to, a lot more work (and planning) would have to be invested into
+// that anyway.
+#ifndef EXPORT_INLINE
+#ifdef HPCGAP
+#define EXPORT_INLINE static inline
+#else
+#define EXPORT_INLINE inline
+#endif
+#endif
+
```
fingolfin: This is one of the few changes which are not a result of "search and replace `static inline`".

src/debug.c
```diff
@@ -8,6 +8,12 @@
 **  SPDX-License-Identifier: GPL-2.0-or-later
 */
 
+// set EXPORT_INLINE to 'extern inline' to force an instance of each
+// function marked with EXPORT_INLINE to be actually emitted, so that
+// it can e.g. be accessed from gdb or lldb, or from software using
+// libgap as a shared library
+#define EXPORT_INLINE extern inline
+
```
fingolfin: This is one of the few changes which are not a result of "search and replace `static inline`".

src/debug.c
```diff
@@ -67,345 +74,3 @@ void InstallBacktraceHandlers(void)
...
-    (VoidFunc)ATOMIC_ELM_PLIST,
-    (VoidFunc)ATOMIC_SET_ELM_PLIST_ONCE,
-    (VoidFunc)ATOMIC_SET_ELM_PLIST,
-    (VoidFunc)GetTLS,
-    (VoidFunc)LCKS_FUNC,
-    (VoidFunc)REGION,
-    (VoidFunc)SET_LCKS_FUNC,
-    (VoidFunc)SET_REGION,
-#endif
-};
```
fingolfin: This is one of the few changes which are not a result of "search and replace `static inline`".

fingolfin:
So now it fails to link on AppVeyor, with this error:
```
...
libtool: link: gcc -shared  obj/src/.libs/ariths.o obj/src/.libs/bags.o obj/src/.libs/blister.o obj/src/.libs/bool.o obj/src/.libs/c_oper1.o obj/src/.libs/c_type1.o obj/src/.libs/calls.o obj/src/.libs/code.o obj/src/.libs/collectors.o obj/src/.libs/compiler.o obj/src/.libs/compstat.o obj/src/.libs/costab.o obj/src/.libs/cyclotom.o obj/src/.libs/debug.o obj/src/.libs/dt.o obj/src/.libs/dteval.o obj/src/.libs/error.o obj/src/.libs/exprs.o obj/src/.libs/ffdata.o obj/src/.libs/finfield.o obj/src/.libs/funcs.o obj/src/.libs/gap.o obj/gen/.libs/gap_version.o obj/src/.libs/gvars.o obj/src/.libs/hookintrprtr.o obj/src/.libs/integer.o obj/src/.libs/intfuncs.o obj/src/.libs/intrprtr.o obj/src/.libs/io.o obj/src/.libs/iostream.o obj/src/.libs/libgap-api.o obj/src/.libs/listfunc.o obj/src/.libs/listoper.o obj/src/.libs/lists.o obj/src/.libs/macfloat.o obj/src/.libs/modules.o obj/src/.libs/modules-builtin.o obj/src/.libs/objcftl.o obj/src/.libs/objects.o obj/src/.libs/objfgelm.o obj/src/.libs/objpcgel.o obj/src/.libs/objset.o obj/src/.libs/opers.o obj/src/.libs/permutat.o obj/src/.libs/plist.o obj/src/.libs/pperm.o obj/src/.libs/precord.o obj/src/.libs/profile.o obj/src/.libs/range.o obj/src/.libs/rational.o obj/src/.libs/read.o obj/src/.libs/records.o obj/src/.libs/saveload.o obj/src/.libs/scanner.o obj/src/.libs/sctable.o obj/src/.libs/set.o obj/src/.libs/stats.o obj/src/.libs/streams.o obj/src/.libs/stringobj.o obj/src/.libs/syntaxtree.o obj/src/.libs/sysfiles.o obj/src/.libs/sysmem.o obj/src/.libs/system.o obj/src/.libs/tietze.o obj/src/.libs/trans.o obj/src/.libs/vars.o obj/src/.libs/vec8bit.o obj/src/.libs/vecffe.o obj/src/.libs/vecgf2.o obj/src/.libs/vector.o obj/src/.libs/weakptr.o obj/src/.libs/gasman.o   -lgmp -lz  --coverage -Wl,--stack -Wl,16777216   -o .libs/cyggap-0.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker .libs/libgap.dll.a
obj/src/.libs/debug.o:debug.c:(.text+0xae7): multiple definition of `TNAM_OBJ'
obj/src/.libs/collectors.o:/cygdrive/c/projects/gap/src/objects.h:353: first defined here
collect2: error: ld returned 1 exit status
make: *** [Makefile.rules:364: libgap.la] Error 1
```

ChrisJefferson:
Yep, now passing (ignoing stupid coveralls), feel free to squash/merge.

PR #3147, opened 2018-12-28
buildsys: fix --enable-Werror for C++ code

fingolfin:


fingolfin:
Turns out we have a pesky warning in our code, specifically in the HPC-GAP code... and we already try to suppress it: from `src/hpc/atomic.h`:
```C
// disable -Wundef temporarily, to avoid warnings about AO_AO_TS_T
// inside of libatomic_ops' header files.
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wundef"
#include <atomic_ops.h>
#pragma GCC diagnostic pop
```

Unfortunately, it seems that those pragmas really only work in gcc, but not in g++ (i.e. gcc in C++ mode). Indeed, here is the relevant GCC bug report from 2012 (still unresolved in GCC 8): https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431

PR #3149, opened 2018-12-28
Fix IsMonomialMatrix for compressed matrices

### LINKED ISSUE #3148: IsMonimialMatrix does not work for compressed matrices
When applying  `IsMonomialMatrix` to a matrix from SL(n,q), GAP throws the error

```
Error, Panic: cannot convert <list> (is a object (positional)) to a plain list in
found := BlistList( M, [  ] ); at /usr/lib/gap/lib/matrix.gi:2580 called from
<function "last">( <arguments> )
```
Converting the entries of a matrix from SL(n,q) via `IntFFE` and then calling `IsMonomialMatrix` works
fine.

As @ssiccha pointed out, one can reproduce the Error via

```
gap> mm := [[0,1,0], [1,0,0]] * Z(3);;
gap> ConvertToMatrixRep(mm);
3
gap> BlistList(mm, []);
Error, Panic: cannot convert <list> (is a object (positional)) to a plain list
not in any function at *stdin*:26`
```

fingolfin:
Note that for MatrixObj implementations which are *not* lists, doing `for row in M do` may not work anymore (unless we decide to officially define that `Iterator(matobj)` shall return a row iterator).

Fixes #3148

We could backport this, the change should be safe enough; but since this is an ancient bug, there is no urgent need to do it either.

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -2574,7 +2574,7 @@ InstallMethod( IsMonomialMatrix,
     if Length( M ) <> len  then
         return false;
     fi;
-    found:= BlistList( M, [] );
+    found:= ListWithIdenticalEntries( len, false );
     for row  in M  do
         j := PositionNonZero( row );
         if len < j or found[j]  then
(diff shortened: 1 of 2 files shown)


fingolfin:
I have no submitted a separate issue #3155 for the `BlistList` and `UniteBlistList` issue. I will also submit another PR which optimizes these two functions for the case where the `sub` list is empty.

fingolfin:
Backported to stable-4.10 via 18edc7fd1f2e1e63056866f990192aaae95bc5fc

PR #3150, opened 2018-12-28
Use NrRows, NrCols, {One,Zero}OfBaseDomain in more places

fingolfin:
This may miss some spots, as I tried to be conservative in what I change.

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -131,11 +131,11 @@ InstallMethod( IsGeneralizedCartanMatrix,
 
     local n, i, j;
 
-    if Length( A ) <> Length( A[1] ) then
+    if NrRows( A ) <> NrCols( A ) then
       Error( "<A> must be a square matrix" );
     fi;
 
-    n:= Length( A );
+    n:= NrRows( A );
     for i in [ 1 .. n ] do
       if A[i,i] <> 2 then
         return false;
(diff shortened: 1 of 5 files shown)


lib/matrix.gi
```diff
@@ -370,8 +370,8 @@ end );
 ##
 BindGlobal( "Matrix_OrderPolynomialInner", function( fld, mat, vec, vecs)
     local d, w, p, one, zero, zeroes, piv,  pols, x, t;
-    Info(InfoMatrix,2,"Order Polynomial Inner on ",Length(mat[1]),
-         " x ",Length(mat)," matrix over ",fld," with ",
+    Info(InfoMatrix,2,"Order Polynomial Inner on ",NrCols(mat),
```
ThomasBreuer: As far as I understand, the matrix must be square,
but still it is irritating that the message talks about a `NCols x NRows matrix`.

lib/matrix.gi
```diff
@@ -451,16 +451,16 @@ BindGlobal( "Matrix_CharacteristicPolynomialSameField",
     function( fld, mat, ind)
     local i, n, ords, base, imat, vec, one,cp,op,zero,fam;
     Info(InfoMatrix,1,"Characteristic Polynomial called on ",
-         Length(mat[1])," x ",Length(mat)," matrix over ",fld);
+         NrCols(mat)," x ",NrRows(mat)," matrix over ",fld);
```
ThomasBreuer: same comment as above

lib/matrix.gi
```diff
@@ -1130,7 +1130,7 @@ InstallMethod( IsZero,
     local ncols,  # number of columns
           row;    # loop over rows in 'obj'
 
-    ncols:= DimensionsMat( mat )[2];
+    ncols:= NrCols( mat )[2];
```
ThomasBreuer: Write just `NrCols( mat )`.

PR #3153, opened 2018-12-29
Avoid fluctuations in stdtasks tests

fingolfin:
The test for WaitAnyTask suffered from random fluctuations in the resulting
code coverage, due to the two tasks being executed running almost exactly the
same time; thus which one finished first depended on the machine running the
tests and its load. This lead to minuscule fluctuations in the coverage
reports, which then could result in PRs being incorrectly flagged as
increasing or decreasing coverage.

This commit tries to change that, by having more stable tests for WaitAnyTask

### DIFF
--- a/tst/testinstall/hpc/stdtasks.tst
+++ b/tst/testinstall/hpc/stdtasks.tst
@@ -40,15 +40,31 @@ gap> TaskResult(t);
 25
 
 #
-gap> task1 := DelayTask(x->SortedList(x), [3,2,1]);;
-gap> task2 := DelayTask(x->SortedList(x), [6,5,4]);;
-gap> which := WaitAnyTask(task1, task2);;
-gap> which in [ 1, 2 ];
-true
+gap> task1 := DelayTask(SortedList, [3,2,1]);;
+gap> task2 := DelayTask(SortedList, [2^15, 2^15-1 .. 1]);;
+gap> which := WaitAnyTask(task1, task2);
+1
+gap> WaitTasks(task1, task2);
 gap> TaskResult(task1);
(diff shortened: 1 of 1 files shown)


fingolfin:
Weird. The coverage report still says that lines 249 & 250 of `hpcgap/lib/hpc/stdtasks.g` are *not* executed / covered. Yet if I tests this locally, they are executed each time I run the .tst file (specifically: 3 times for each invocation of that test file).

What am I missing? Perhaps there is a problem here because multiple threads are involved? Perhaps @ChrisJefferson has some insights on that?

ChrisJefferson:
Not entirely sure what you mean by "are executed", do you mean the profiling showed they are executed, or you checked they are run?

The biggest limitation in HPC-GAP is we only profile the thread where ProfileLineByLine is run. Generating coverage of all threads wouldn't be too hard, but doing the profiling, tracking entering and leaving functions and such like, would require a fairly major rewrite.

fingolfin:
I checked they are run. But the profiler (as visible here) does not pick it up.

ChrisJefferson:
I found on my computer your test was only hitting that line about one time in 4. I've tried pushing a change which on my computer reliably seems to hit the line. Let's see if that fixes it.

ChrisJefferson:
I'll see what coverage this gets, but I think I have a very much simpler patch which achieves the same coverage.

ChrisJefferson:
Now pushed a simpler way of getting coverage.

fingolfin:
Well, at least Codecov seems to see the lines hit now, so I'd be fine with merging it. But somebody (e.g. @ChrisJefferson) needs to approve it first ;-)

Issue #3140, opened 2018-12-21
Recursion depth trap for due to level 3 assertion in characteristic polynomial of matrix code


jackschmidt:
There is an issue in the characteristic polynomial of matrix routine as called by the meat-axe as called by the conjugacy classes code for permutation groups. This is in master and 4.8.8, but requires assertions turned on, to level 3, so I'm not sure if it is considered a real issue. I think the fix is simple though.

### Reproducer:

```
SetAssertionLevel(3);
ConjugacyClasses(WreathProduct(SymmetricGroup(2),SymmetricGroup(5)));
```

### Line with the problem (polynomial versus coefficient list):

Maybe this line is wrong (`lib/matrix.gi` line 466 in 4.8.8, line 479 in 4.10.0, line 476 in master):
```
    Assert(3, IsZero(Value(cp,imat)));
```

```
brk> cp;
[ Z(2)^0 ]
brk> imat;
<an immutable 1x1 matrix over GF2>
```

`cp` is not a polynomial. Maybe it is a coefficient list of a polynomial?

### More log:

The functions before the loop are:
```
Value( f, x, One( x ) ) at /home/jack/gap/lib/ratfunul.gi:1154 called from
Value( cp, imat ) at /home/jack/gap/lib/matrix.gi:476 called from
CharacteristicPolynomialMatrixNC( DefaultFieldOfMatrix( mat ), mat, 1 ) at /home/jack/gap/lib/matrix.gi:671 called from
CharacteristicPolynomial( mm ) at /home/jack/gap/lib/matrix.gi:4412 called from
x ^ f[2] at /home/jack/gap/lib/ratfunul.gi:1119 called from
Value( f, x, One( x ) ) at /home/jack/gap/lib/ratfunul.gi:1154 called from
Value( cp, imat ) at /home/jack/gap/lib/matrix.gi:476 called from
CharacteristicPolynomialMatrixNC( F, M, 1 ) at /home/jack/gap/lib/meataxe.gi:913 called from
SMTX.SMCoRaEl( matrices, ngens, newgenlist, dim, F ) at /home/jack/gap/lib/meataxe.gi:1036 called from
SMTX.IrreducibilityTest( module ) at /home/jack/gap/lib/meataxe.gi:1233 called from
SMTX.IsIrreducible( cmod ) at /home/jack/gap/lib/meataxe.gi:1955 called from
SMTX.CollectedFactors( m ) at /home/jack/gap/lib/meataxe.gi:3059 called from
[...]

PR #3154, opened 2018-12-30
FIX: Fitting free data use in clashom

hulpke:
If the group has been obtained as subgroup from a fitting free computation,
the data will be inherited and might not guarantee that the factor group
really is fitting free. Test/resolve this.

This fixes gap-system#3139

Also recode assertion to avoid #3140

### DIFF
--- a/lib/clashom.gi
+++ b/lib/clashom.gi
@@ -2398,8 +2398,15 @@ local r,	#radical
       hom:=ser.factorhom;
       ntrihom:=true;
       f:=Image(hom);
+      # if lift setup is inherited, f might not be trivial-fitting
+      if Size(RadicalGroup(f))>1 then
+        # this is proper recursion
+        cl:=ConjugacyClasses(f:onlysizes:=false);
+        cl:=List(cl,x->[Representative(x),Centralizer(x)]);
+      else
       # we need centralizers
-      cl:=ConjugacyClassesFittingFreeGroup(f:onlysizes:=false);
+        cl:=ConjugacyClassesFittingFreeGroup(f:onlysizes:=false);
+      fi;
       fants:=Filtered(NormalSubgroups(f),x->Size(x)>1 and Size(x)<Size(f));
     else
       if IsPermGroup(G) then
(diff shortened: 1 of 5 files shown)


fingolfin:
Backported to stable-4.10 in commits 4f6aa5c3b23d875ffaefe83c6b63519efaa5d0d0 fe8ebaeccf5478d3b44c962046fab093f592b6e7 3f85d6281af52c7392bf2ce5ae103c0a7ea2f989

alex-konovalov:
@hulpke for the release notes, could you please give a proper description of the fix? Thanks.

hulpke:
@alex-konovalov I think the secription is as good as I can make it. For the release notes 
Resolved an error in computing conjugacy classes of a subgroup
should be fine, but it also just cvould go under a general headline of numerous bugs were fixed, as the error was not dangerous.

alex-konovalov:
I am particularly concerned about "fitting free computation" - seems this part could have been made clearer.

hulpke:
@alex-konovalov 
Fitting free/solvable radical is a class of algorithms that build the potentially problematic data structure. Are you concerned about  the release notes (for which I suggested a line, and which do not need this detail), or the code base (for which I think the description in the PR is plausible?

alex-konovalov:
About the style of the text used for release notes.

alex-konovalov:
Let me explain in more details why I am assuming that for a side reader this may be cryptic.

First of all, it's now clear which is the behaviour before and which is after. Is what you're describing a bug, or a behaviour after a bug? Next,

> If the group has been obtained as subgroup from a fitting free/solvable radical computation,

what is actually computed? A subgroup or a solvable radical? I would struggle to parse this

> the data will be inherited and might not guarantee that the factor group

"will be" (after the fix) or "was" (before it) or actually "is" since this did  not change at all?

really is fitting free. Test/resolve this.

and Fitting is a surname to be spelled as Fitting.

Would you be happy with this, or it's incorrect:

- [#3154](https://github.com/gap-system/gap/pull/3154) If the group has been obtained as subgroup from a Fitting free/solvable radical computation, the data is inherited and might not guarantee that the factor group really is Fitting free. Added a check and an assertion to catch this situation.

alex-konovalov:
I've put that version in https://github.com/gap-system/gap/pull/3503/ - if you have any corrections, please leave them as comments there.

PR #3158, opened 2019-01-05
Experiment: Try always compiling statements

ChrisJefferson:
This is an experiment with always compiling code, rather than using the interpreter. This would greatly simplify GAP, by allowing us to delete the interpreter.

The time taken to start GAP is very similar to normal GAP. There are 2 known issues:

1) There are numerous warnings printed. These come from inside `if` statements, where variables are declared, and then used in a new function. As we now parse the entire `if` statement before executing it, the interpreter does not know these variables will be declared before they are then used in a function.

2) Expressions (like `x` or `2`) aren't currently handled yet, as they aren't statements. This isn't that hard to fix, but will require some fiddling.

The main question is, assuming these issues can be fixed, is it worth the amount of code which would be deleted to not interpret GAP?

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -1342,6 +1342,13 @@ void CodeReturnObj ( void )
 }
 
 
+void CodeReturnMaybeObj ( void )
+{
+    if(CS(CountExpr) >= 1) {
+        CodeReturnMaybeObj();
+    }
+}
+
 /****************************************************************************
 **
 *F  CodeReturnVoid()  . . . . . . . . . . . . . .  code return-void-statement
--- a/lib/random.gi
+++ b/lib/random.gi
@@ -251,7 +251,7 @@ function(low, high)
 end );
 
 
-(function()
+tmpfunc := function()
     local func;
     func := function(installType)
         return function(args...)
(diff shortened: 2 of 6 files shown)


ChrisJefferson:
Update: I've been looking into remove the interpreter. The main problem is GAP expressions like: this: `(function() if x then return 1; fi; end)();`. This expression might return a value depending on the value of `x`. If we want to wrap this in a "fake function expression", we have to decide in advance if we want a value returning or not, GAP does not have a way of expressing a function where it may or may not return a value, where we want the value returned if it exists.

This wouldn't be that hard to do (it's been discussed before), but it would be a bit more surgery than I hoped for.

src/code.c
```diff
@@ -1342,6 +1342,13 @@ void CodeReturnObj ( void )
 }
 
 
+void CodeReturnMaybeObj ( void )
+{
+    if(CS(CountExpr) >= 1) {
+        CodeReturnMaybeObj();
```
fingolfin: So this runs an infinite recursion, by calling itself?

stevelinton:
The other traditional concern with always compiling was very large immediate expressions, in files. It's not unreasonable to have a file containing something like 

```
x := <multi-gigabyte list, record or permutation literal>;
```
Such a thing will compile to a very large function body indeed, in at least some cases, whereas in immediate interpretation, each entry of the data structure can be evaluated straight away and might reduce to an immediate integer or FFE, using a relatively small amount of memory.

fingolfin:
We discussed the arguments against doing this before, and it seems that it turned out to be more work than expected anyway.

So, is there a point in keeping this PR open? Are further experiments planned / hoped for, or possibly discussions?

ChrisJefferson:
I'll keep the code, but it seems no-one is really interested in getting rid of immediate mode.

PR #3160, opened 2019-01-06
Fix HasAvailableBytes for gzipped files

### LINKED ISSUE #3145: GAP 4.10.0 crashes when reading a gzipped file
When reading gzipped files, both GAP 4.10.0 and the master branch sometimes crash.
This did not happen up to GAP 4.9.

I can reproduce this problem with the following recipe.

1. Download [this file](http://brauer.maths.qmul.ac.uk/Atlas/spor/HS/mtx/HSG1-f49r896aB0.m1).

2. Compress the file using `gzip`.

3. Start GAP with the `-A` option.

4. Enter the following lines.

```
filename:= "HSG1-f49r896aB0.m1";;
file:= InputTextFile( filename );;
repeat
  line:= ReadLine( file );
until line = fail;
```

GAP 4.10.0 reads several lines of the file and then crashes with the message
`*** buffer overflow detected ***: /.../gap terminated`.

This problem does not occur when one sets `filename:= "HSG1-f49r896aB0.m1.gz";`
and then reads the file line by line.

ChrisJefferson:
Fixes #3145


src/sysfiles.c
```diff
@@ -1961,6 +1961,11 @@ Int HasAvailableBytes( UInt fid )
     }
 
 #ifdef HAVE_SELECT
+  // All sockets other than raw sockets are always ready
+  if(syBuf[fid].type != raw_socket) {
```
fingolfin: Optional edit:
```suggestion
  if (syBuf[fid].type != raw_socket) {
```

src/sysfiles.c
```diff
@@ -1961,6 +1961,11 @@ Int HasAvailableBytes( UInt fid )
...
+      return 1;
+  }
+  else
```
fingolfin: Hmmmm... would it perhaps be better to fall through to the call to `SyIsEndOfFile` below? I.e. just insert this line here:
```C
  if (syBuf[fid].type == raw_socket)
```

ChrisJefferson:
Improved. As you say, this is hard to test. It seems on some OSes, depending on who knows what, passing negative numbers as the fd to select causes horrible problems -- but it's hard to reproduce. Whatever we do, it's certainly wrong behaviour!

fingolfin:
Backported to stable-4.10 in 0b704ae4ac0e2238b0c1b8f700a87ec5ea2fd0de

PR #3163, opened 2019-01-07
travis: use Julia 1.1.0-rc1 for tests

fingolfin:
This should be backported to stable-4.10 simply so that tests there will also be based on Julia 1.1, and not some random old nighly build.

Once Julia 1.1 is out, I'll update this once more to target that.


fingolfin:
Backported via 6d46c38e25a4b7d50924cf1c05c53fa94271ce62

PR #3165, opened 2019-01-07
More tests and minor bug fix for PrintToFormatted; fix bug in READ_ALL_FILE on Cygwin

ChrisJefferson:
This fixes a minor bug (when trying to print an Error, GAP produces a different Error).

While fixing this, I decided to write a bunch of other tests to improve coverage.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -1579,6 +1579,7 @@ Obj FuncREAD_ALL_FILE (
 {
     Char            buf[20000];
     Int             len;
+    // Length of string read this loop (or negative for error)
     UInt            lstr;
     Obj             str;
     UInt            csize;
@@ -1598,6 +1599,7 @@ Obj FuncREAD_ALL_FILE (
  getmore:
 #endif
     while (ilim == -1 || len < ilim ) {
+      lstr = 0;
       if ( len > 0 && !HasAvailableBytes(ifid))
           break;
       if (SyBufIsTTY(ifid)) {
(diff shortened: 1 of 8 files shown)


ChrisJefferson:
Something is up with the C `select` function and Cygwin. *sigh*. I'll investigate.

ChrisJefferson:
AppVeyor is failing due to a hang, as the test is hanging on Cygwin.

ChrisJefferson:
Now with added little cygwin fix for asking to read more bytes than a file contained (which previously entered an infinite loop)

src/streams.c
```diff
@@ -1650,7 +1652,8 @@ Obj FuncREAD_ALL_FILE (
     }
     len = j;
     SET_LEN_STRING(str, len);
-    if (ilim != -1 && len < ilim)
+    // If we have not read enough, and read at least one character this loop
```
fingolfin:
```suggestion
    // If we have not yet read enough, and have read at least one character this loop, then read more
```

ChrisJefferson:
Now all fixed up i hope

tst/testinstall/read.tst
```diff
@@ -77,6 +79,12 @@ gap> FileString( Filename(dir, "test.g.gz"), "\037\213\b\b0,\362W\000\ctest.g\00
 32
 gap> StringFile( Filename(dir, "test.g") ) = "1+1;\n" or ARCH_IS_WINDOWS(); # works only when Cygwin installed with gzip
 true
+gap> StringFile( "/" );
+Error, in StringFile: Is a directory (21)
+
+gap> FileString("/dev/full", "outdata", false);
+Error, in FileString: No space left on device (28)
```
fingolfin: Unfortunately, this test fails on OS X (and I am kinda surprised it works on cygwin).

PR #3167, opened 2019-01-09
in configure, run Julia without `startup.jl`

ThomasBreuer:
I had put a `print` statement into my `startup.jl` file,
for test purposes, and then GAP's `configure` ran into an error
because the printed message was regarded as the value of `JL_SHARE` etc.

I think it is better not to read `startup.jl` at all in this situation.


fingolfin:
Backported to stable-4.10 in a3e156ce0419cf3ab4338c551bda7fba031ae567

PR #3168, opened 2019-01-10
Allow input and output to be mixed in Test

ChrisJefferson:
This was previously forbidden as it can lead to confusing errors,
but it also broke some valid Tests, as GAP can begin outputting
before a complete statement has been inputted.

In particular, this previous change broke the "Browse" package.


tst/testinstall/interpreter.tst
```diff
@@ -248,6 +248,25 @@ gap> Unbind(l![fail]);
 Error, PosObj Assignment: <position> must be a positive small integer (not the\
  value 'fail')
 
+#
+# Statements where input + output are mixed
+# Checks test can handle output directly cute+pasted from GAP's output
```
fingolfin:
```suggestion
# Checks test can handle output directly cut+pasted from GAP's output
```

fingolfin: Also, why is this in interpreter.c? It's a test for `Test`, not the interpreter, isn't it? Note that interpreter.tst is meant to be in sync with coder.tst -- we already violate this right now (my fault) but I was recently working on fixing this again, and would prefer not to drift more.

That said, I'll survive if we add it here, I'll just move it elsewhere when I get to finish that sync I just mentioned.

PR #3169, opened 2019-01-10
kernel: avoid use of compiler extensions in syntaxtree.c

fingolfin:
Calling a vararg macro with an empty list if variadic arguments leads
to various problems, which prompted all kind of compiler extensions in
GCC and other compilers, and also a revision of varargs in e.g. C++2a.

Here, we are bitten by the fact that strictly speaking, ISO C forbids
empty initializer braces, so `{ __VA_ARGS__ }` is invalid if there
are no varargs. A similar problem then affects COMPILER_ARITY.

Also, not specifying the size of the Compilers static array, also in its
prototype, is not supported by ISO C.

### DIFF
--- a/src/syntaxtree.c
+++ b/src/syntaxtree.c
@@ -49,15 +49,24 @@ typedef struct {
 
 // We put compilers for statements and expressions into the same static array,
 // assuming that the set of their respective ids are disjoint.
-static const CompilerT Compilers[];
+static const CompilerT Compilers[256];
 #define COMPILER_ARITY(...) (sizeof((ArgT[]){ __VA_ARGS__ }) / sizeof(ArgT))
 #define COMPILER(tnum, compiler, ...)                                        \
     [tnum] = {                                                               \
         tnum, compiler, #tnum, COMPILER_ARITY(__VA_ARGS__), { __VA_ARGS__ }  \
     }
 
+#define COMPILER0(tnum, compiler)                                            \
+    [tnum] = {                                                               \
+        tnum, compiler, #tnum, 0, { { 0 } }  \
+    }
+
+
 #define COMPILER_(tnum, ...)                                                 \
     COMPILER(tnum, SyntaxTreeDefaultCompiler, __VA_ARGS__)
+
+#define COMPILER0_(tnum)                                                     \
+    COMPILER0(tnum, SyntaxTreeDefaultCompiler)
 
 #define ARG(name, func)                                                      \
     {                                                                        \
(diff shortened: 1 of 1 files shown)


markuspf:
Your description text makes no sense to me.

fingolfin:
OK, which bit of it? The first paragraph is merely some background rambling and could be dropped (I elaborate that calling a variadic macro with an empty list of varargs is known to run into all kinds of corner and edge cases, which prompted various compilers to implement different kinds of compiler extensions, and also lead to standardization committees to try to improve the status quo.)

Then I go on to explain that in particular using `{ __VA_ARGS__ }` as an initialization list is problematic if there are no var args, as that evaluates to `{ }`, but using an empty initialization list for a non-empty array is strictly speaking not allowed.

To illustrate, consider this simplified version of the relevant code:
```C
extern int array[];
int array[3] = {};

extern int size;
int size = sizeof((int[]){  }) / sizeof(int);
```
This leads to the following warnings with GCC...
```
$ gcc-8 -Wall -Wpedantic -c test.c -o /tmp/test.o
test.c:2:16: warning: ISO C forbids empty initializer braces [-Wpedantic]
 int array[3] = {};
                ^
test.c:5:26: warning: ISO C forbids empty initializer braces [-Wpedantic]
 int size = sizeof((int[]){  }) / sizeof(int);
                          ^
```
resp. with clang:
```
$ clang-7 -Weverything -c test.c -o /tmp/test.o
test.c:2:16: warning: use of GNU empty initializer extension [-Wgnu-empty-initializer]
int array[3] = {};
               ^
test.c:5:26: warning: use of GNU empty initializer extension [-Wgnu-empty-initializer]
int size = sizeof((int[]){  }) / sizeof(int);
                         ^
test.c:5:26: warning: zero size arrays are an extension [-Wzero-length-array]
3 warnings generated.
```

markuspf:
While I see how this addresses some implementation-defined behaviour which just happened to work on my compiler of the day, I am a bit unhappy with this.

I was already not too happy with the `COMPILER` and `COMPILER_`, but now we have `COMPILER0` and `COMPILER0_`, too, which are duplicating macro "code".

At the very least one might've to rename `COMPILER_` to `COMPILER_DEFAULT` (which I did not want to do previously, because the name becomes so long)

Maybe I should be looking at some C++ template stuff that can make this look slightly less dodgy? Get rid of macros full stop?

fingolfin:
I agree that it is not nice. Of course we can also just keep it as-is for now, as it seems to work fine with both GCC and clang, and if ever somebody runs into problems because of it, we can easily address it then.

As to `COMPILER` vs. `COMPILER_` (or `COMPILER_DEFAULT`, or whatever): to be honest, I would just abandon this; spelling out that one argument explicitly each time is not that much work, and it removes one layer that somebody who wants to understand the code has to understand.

PR #3170, opened 2019-01-10
kernel: avoid hack in collector to extract word length

fingolfin:
While looking at another issue in the collector code, I stumbled over some weird code that was taking an UInt1/2/4 pointer, and casting it to an Obj pointer and expecting to find an intobj there... I eventually realized what the code was actually doing: retrieving `NPAIRS_WORD(word)` from a pointer returned by `DATA_WORD(word)` resp. `CONST_DATA_WORD(word)`.

This patch rewrites the code in question to not resort to such a hack, which should make it a bit easier to understand, and also gets rid of a nasty implicit dependency of the internal data layout of a word object.

### DIFF
--- a/src/collectors.cc
+++ b/src/collectors.cc
@@ -79,8 +79,7 @@ static Obj TYPE_KERNEL_OBJECT;
 **    push a generator <gen>  with exponent <exp> onto the stack.
 **
 **  SC_PUSH_WORD( word, exp )
-**    push <word>  with global exponent <exp>  into the stack, the macro uses
-**    <word> and <exp> only once.
+**    push <word>  with global exponent <exp>  into the stack.
 **
 **  SC_POP_WORD()
 **    remove topmost word from stack
@@ -91,7 +90,7 @@ static Obj TYPE_KERNEL_OBJECT;
         return -1; \
     } \
     *++nw = CONST_DATA_WORD(word); \
-    *++lw = *nw + (INT_INTOBJ((((const Obj*)(*nw))[-1])) - 1); \
+    *++lw = *nw + NPAIRS_WORD(word) - 1; \
     *++pw = *nw; \
     *++ew = (**pw) & expm; \
     *++ge = exp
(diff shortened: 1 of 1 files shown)


src/collectors.cc
```diff
@@ -519,7 +521,7 @@ Int SingleCollectWord(Obj sc, Obj vv, Obj w)
             /* collect a whole word exponent pair                          */
             else if( *pw == *nw && INT_INTOBJ(avc[gn]) == gn ) {
               gn = SAddWordIntoExpVec(
-                   v, *pw, *ge, ebits, expm, ro, pow, lpow  );
+                   v, *nw, *lw, *ge, ebits, expm, ro, pow, lpow  );
```
fingolfin: Note that by the check in the `else if` above, `*pw == *nw`; and in `SC_PUSH_WORD` we set `*lw` to point at the end of the word, as required for this call.

src/collectors.cc
```diff
@@ -79,8 +79,7 @@ static Obj TYPE_KERNEL_OBJECT;
 **    push a generator <gen>  with exponent <exp> onto the stack.
 **
 **  SC_PUSH_WORD( word, exp )
-**    push <word>  with global exponent <exp>  into the stack, the macro uses
-**    <word> and <exp> only once.
+**    push <word>  with global exponent <exp>  into the stack.
```
fingolfin: I've adjusted this comment as we now evaluated `word` twice; and my guess is that part of the reason for that nasty hack here was to ensure we evaluate it only once.
After inspecting all callers to this macro, it should be perfectly safe to evaluate `word` twice.

stevelinton: Maybe add an explicit comment that you do evaluate it twice? Is there some reason not to make this an inline function?

fingolfin: I thought about changing it into an inline function, but decided against it (for now), for two reasons: (a) I wanted to keep the changes in this commit as small as possible; (b) implementation would actually be somewhat nasty, since we modify nw, lw, pw, ew, ge ; so we'd have to pass pointers to these in (so pointers to pointers to pointers to UInt...), and then replace `*++nw` by `*++*nw`, and I am sure that I'd win some kind of price for that (but not a good one)....

Well, I just realize this is in a C++ file, so we could also use a C++ reference, which would at least cut down on the number of `*` we have to use. But then people editing this would need to know about C++ references. Which aren't terribly complicated to learn about, but we agreed to be careful about using C++ features, and that we didn't want to have more and more of them creep in silently over time, so I am reluctant to do that...

PR #3173, opened 2019-01-10
Exclude pkg and extern from codecov coverage gathering script

markuspf:


.travis.yml
```diff
@@ -135,7 +135,7 @@ script:
 after_script:
   - set -e
   - bash etc/ci-gather-coverage.sh
-  - bash <(curl -s https://codecov.io/bash)
+  - bash <(curl -s https://codecov.io/bash) -f "!./pkg/*" -f "!./extern/*"
```
fingolfin: Ah, I did not try that. If this doesn't work, I wonder if the `-g GLOB` option might also be useful?

fingolfin:
Unfortunately there is still garbage in the coverage for `io.c`, see https://codecov.io/gh/gap-system/gap/src/59778277d4b8ece2aee79af6741e25769aa39b3d/src/io.c

markuspf:
yes. it comes from the gmp build...

markuspf:
also the exclusion doesn't work on appveyor yet of course

.travis.yml
```diff
@@ -135,7 +135,7 @@ script:
 after_script:
   - set -e
   - bash etc/ci-gather-coverage.sh
-  - bash <(curl -s https://codecov.io/bash)
+  - bash <(curl -s https://codecov.io/bash) -f "!./pkg/*" -f "!./extern/*" -f "!./build/*"
```
fingolfin: Ahh, I bet I forgot to exclude `build` in past attempts, too, which might explain some things.

markuspf:
aaand we got liftoff. The coverage of `io.c` still doesn't seem quite right, but it is now more sensible.

markuspf:
now for some wrestling with bash... because why not.

fingolfin:
Things are indeed improving. `io.c` coverage seems mostly OK, in that only lines that really are code seem to be marked as such (and not anymore stuff in the middle of comments etc.). But something is still off -- lots of code is not marked at all, even code that is preceded and followed by code marked as executable.

PR #3182, opened 2019-01-16
stable-4.10: backport two buildsys changes to the GAPARCH

fingolfin:
This PR backports two commits from master that change the GAPARCH string:
* one commit by @ChrisJefferson from PR #2873 which moves GAP the kernel version to the configure script; the only change is that I adapted it to preserve the kernel version used by 4.10; this commit also modifies the GAPARCH, by inserting the kernel version into it
* one commit from PR #2905 (out of three, the other two were already backported) which changes the GAPARCH to distinguish between "default" builds, HPC-GAP and Julia builds.

These changes are useful when one wants to debug HPC-GAP and Julia builds of GAP together with packages, as the different GAPARCHs ensure that rebuilding the same package for each build variant leaves separate binaries, so the package then can be used by those different builds (i.e., no need to keep separate pkg dirs, one can use out-of-tree builds for the different build configs).

The main concern (and the reason I did not push these backports directly) is that perhaps some code somewhere could be broken by the modified GAPARCH... We already tried hard to ensure that all deposited packages keep working, and I can't really think of problems, but I want that (a) tests for this change are run, and (b) would like to know if e.g. @ChrisJefferson can think of any potential issues

### DIFF
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -28,6 +28,10 @@ GC_SOURCES = @GC_SOURCES@
 COMPAT_MODE = @COMPAT_MODE@
 GAPARCH = @GAPARCH@
 
+# GAP kernel version
+GAP_KERNEL_MINOR_VERSION = @gap_kernel_minor_version@
+GAP_KERNEL_MAJOR_VERSION = @gap_kernel_major_version@
+
 # autoconf package metadata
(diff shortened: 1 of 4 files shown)


alex-konovalov:
I started a test of this PR on Jenkins - hopefully will finish tomorrow.

ChrisJefferson:
This looks ok to me

PR #3183, opened 2019-01-16
kernel: add 'extern' to func decls in header files, and also drop some inappropriate uses of 'extern'

fingolfin:
Closes #3202

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -3040,7 +3040,7 @@ void CodeIsbComObjExpr ( void )
 **
 */
 
-extern void CodeEmpty( void )
+void CodeEmpty(void)
 {
   Stat stat;
   stat = NewStat(T_EMPTY, 0);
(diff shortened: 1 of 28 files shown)


fingolfin:
I have dropped some changes from this PR, in order to make it easier to review. So it's now down to one commit.

ChrisJefferson:
I think I said before, I don't really like sticking extern on function prototypes, as it doesn't really mean / do anything, and (personally) confuses me as I have to remember extern doesn't mean anything on a function prototype.

fingolfin:
I mean, I am also OK with dropping extern everywhere. I just dislike the current mish mash, and adding it to the few places it is missing is less work than removing it everywhere (can't just search & replace, as we need to keep it for variables)

fingolfin:
To qualify this a bit, before this PR, `git grep -w extern src/*h src/hpc/*.h | wc -l` reports 999 hits, with this PR, there are 1106 hits. Trying (crudely) to only find hits involving function prototypes, one might use `git grep "extern.*(" src/*h src/hpc/*.h  | wc -l` which reports 834 hits before, and 940 after.

Anyway, as I said, if we can agree to *remove* `extern` from all function prototypes (at least in headers), that's also fine by me.

PR #3185, opened 2019-01-17
Make (Un)HideGlobalVariables obsolete

fingolfin:
These functions were never documented, and so far in all cases where they
appear, there seem to be better, more robust alternatives, such as reading and
evaluating external code inside a function context with suitable locals set,
or by adapting the (interface to) the external code.

IMHO, their existence promotes bad code design, which is why I'd like to make them obsolete and eventually remove them.


ChrisJefferson:
I agree this has always been super weird.

Issue #3187, opened 2019-01-17
Segfault in tst/testinstall/syntaxtree.tst


alex-konovalov:
This happens in the master branch when all packages are loaded. See e.g. on Travis here: https://travis-ci.org/gap-infra/gap-docker-master-testsuite/jobs/480427212. Started to happen two days ago.

alex-konovalov:
Incidentally, on the same day Semigroups and Utils tests started to fail too:
- https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master/jobs/480676994
- https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master/jobs/480677009
for the same versions of these packages for which they used to work a day before.

alex-konovalov:
The error happening in Semigroups does not sound right at all:
```
######## > Diff in:
/home/gap/inst/gap-master/pkg/semigroups-3.0.20/tst/standard/congrms.tst:104
# Input is:
IsSubrelation(cong2, cong);
# Expected output:
Error, Semigroups: IsSubrelation: usage,
congruences must be defined over the same semigroup,
# But found:
Error, Z(3)^7Z(3)^7
########
#I  Errors detected while testing
```

ChrisJefferson:
EDIT: No, that's old code.

ChrisJefferson:
Updated problem (I think)

`GET_VALUE_FROM_CURRENT_BODY` isn't wired up properly to the syntaxtree (I think), so getting constants isn't producing the right answers.

fingolfin:
`GET_VALUE_FROM_CURRENT_BODY` is fine, the problem is that I introduced a silly GC bug, by keeping a pointer into an object across an allocation. Should be fixed in PR #3191

PR #3189, opened 2019-01-17
Fix `IsHomogeneousList` for empty string (should return true, returned false)

fingolfin:
Also split a .tst file (see commit message for explanation)


fingolfin:
The test failure is due to bug #3187 and thus fixed by PR #3191

fingolfin:
Not sure which "level" of "why" you are referring. Do you mean: why did anybody ever think the empty string should not be homogeneous? (Likely answer: because humans make mistakes?).

PR #3190, opened 2019-01-17
buildsys: only use -Wextra-semi for C++ code

### LINKED ISSUE #3188: Warning about `-Wextra-semi`
I get the following on ubuntu 18.10, gcc 8.2.0, on all C source files:

```
   C       src/vars.c => obj/src/vars.lo
cc1: warning: command line option ‘-Wextra-semi’ is valid for C++/ObjC++ but not for C
```

fingolfin:
Fixes #3188


ChrisJefferson:
I think this is the wrong way around -- the flag should be in C++, and not C.

fingolfin:
Indeed, I misread that error message in your issue.

fingolfin:
@ChrisJefferson should be better now; this time I also made sure to verify it with GCC 8.2 locally

PR #3191, opened 2019-01-17
kernel: fix GC crash in PushValue()

### LINKED ISSUE #3187: Segfault in tst/testinstall/syntaxtree.tst

fingolfin:
The NEW_PLIST call could trigger a GC, which then rendered the value of header
invalid, possibly leading to arbitrary memory corruption, and also the values
pointer of the function body was then not set correctly.

Fixes #3187

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -517,9 +517,13 @@ void PushBinaryOp (
 static Int PushValue(Obj val)
 {
     BodyHeader * header = (BodyHeader *)STATE(PtrBody);
-    if (!header->values)
-        header->values = NEW_PLIST(T_PLIST, 4);
-    return PushPlist(header->values, val);
+    Obj values = header->values;
+    if (!values) {
+        values = NEW_PLIST(T_PLIST, 4);
+        header = (BodyHeader *)STATE(PtrBody);
+        header->values = values;
+    }
+    return PushPlist(values, val);
 }
 
 


src/code.c
```diff
@@ -516,10 +516,10 @@ void PushBinaryOp (
 */
 static Int PushValue(Obj val)
 {
-    BodyHeader * header = (BodyHeader *)STATE(PtrBody);
-    if (!header->values)
-        header->values = NEW_PLIST(T_PLIST, 4);
-    return PushPlist(header->values, val);
+    Obj values = ((BodyHeader *)STATE(PtrBody))->values;
+    if (!values)
+        values = ((BodyHeader *)STATE(PtrBody))->values = NEW_PLIST(T_PLIST, 4);
```
ChrisJefferson: This line looks too clever to me (doing two assignments), also isn't PtrBody a raw pointer, so it's important if GAP calls STATE(PtrBody) before or after the RHS?

Maybe assign to `values` in this line, then assign to PtrBody in a new second line?

fingolfin: You're absolutely right. Patch revised.

PR #3193, opened 2019-01-18
kernel: fix bad prototypes for zero arg funcs

fingolfin:


markuspf:
Is the list of places where the fixes are applied "exhaustive" (motivated by compiler warnings, etc?)

fingolfin:
Almost -- I used `-Wstrict-prototypes` and fixed all I could; but it also warns about this in `system.h`:
```C
typedef Obj (* ObjFunc) (/*arguments*/);
```
While we could think about replacing that by a union or something, but there are at least two packages using `ObjFunc` (float and profiling), so we have to be a bit careful, and I decided to punt on that for now.

(Actually, I was experimenting with `-Weverything` in clang, which motivated several of my recent PRs)

PR #3197, opened 2019-01-18
Speed up some of our slowest tst files; merge "libgap" job on travis into previous job

fingolfin:
My hope is that this will reduce Travis turnaround times measurably. We'll see.


fingolfin:
Thanks for looking at this change, and pointing out the problem. I think this proves once again how important it is to document such implicit requirements in code comments. Right now, there was no way to know about what "the point" of the test is by just looking at it; it wasn't mentioned in the commit message adding it either.

I have reverted the change to the test, and tried to add a comment explaining that the "11" is crucial. @hulpke It'd be great if you could briefly read that comment, and suggestion changes if it is too off base. Thanks!

Issue #3089, opened 2018-12-05
libgap API: Problems with GASMAN and StackBottom


embray:
I have brought this issue up previously on the Sage Trac, but this is a generic problem with embedding libgap in larger applications, and it is a multi-faceted problem which I thought should be clearly summarized somewhere.

The core of the problem is that the GAP kernel, as designed, assumes that all code which uses GAP objects (tracked through the GASMAN GC), is going to run in a context with either a call to `InitializeGap` as a parent stack frame, or at least have a common parent with the call to `InitializeGap`.  This is illustrated in the fact that `InitializeGap` uses a pointer to some main function's `argc` variable to set the `StackBottom` variable in the GC (this itself is a problem which I'll list later).

This assumption for setting `StackBottom` is safe for GAP itself, but not for embedding in another application with libgap.  Consider the following scenario, where `GAP_Initialize` is called by some program embedding GAP, where GAP initialization is delayed until necessary, and possibly not called until deep into some other function call (forgive my ASCII art):

```
#0 main()
#1  |-<some func>
#2      |- <some func 2>
             .
             .
#42           |- <some func 3>   // 42 stack frames deep (or any arbitrary large N)
#43                |- GAP_Initialize()  // We need to use GAP now for some computation
                                         // StackBottom is set *here*
```

Now imagine that later in the application there is some other function called which uses GAP, but less deeply nested than the one that happened to call `GAP_Initialize`; it also happens to result in some nested GAP function calls:

```
#0 main()
#1  |-<some func>
#2      |- <some func 2>
             .
             .
#10           |- <some func 3>
#11                |- GAP_EvalString(....)
                       .
                       .
#22                    |- CALL_NARGS(....)  // Some objects are allocated on the stack here
#23                         |- CALL_NARGS(...)  // "top" of stack is here; garbage collection occurs
.  
.
.
#42 (old)          // StackBottom is still down here somewhere...
```

Now we have (on x86 architectures) `StackTop > StackBottom`.  `GenStackFuncBags` does handle the case where  `top > bottom`, but that "seems" to be more in order to handle different architectures where the stack grows in different directions, than some deliberate attempt to handle this case.  Because now there are stack frames possibly containing young bags that haven't been swept yet which are above the current "top" of the stack. 

[...]

PR #3199, opened 2019-01-20
Julia GC: fix stack marking

### LINKED ISSUE #3114: Error when computing irreducibles of a permutation group
The following happens in GAP 4.10.0 as well as in the current master branch.

```
gap> g:= AtlasGroup( "U3(8).3^2" );;
gap> Irr( g );;
Error, List Element: <list>[19] must have an assigned value in
  M[s][t] := M[s][t] + T[2][i]; at /.../ctblgrp.gi:1790 called from
D.ClassMatrixColumn( D, M, bestMat, col
 ); at /.../ctblgrp.gi:716 called from
SplitStep( D, bsm ); at /.../ctblgrp.gi:2065 called from
DixonSplit( D ) at /.../ctblgrp.gi:2171 called from
IrrDixonSchneider( G, rec(
     ) ) at /.../ctblgrp.gi:2152 called from
IrrDixonSchneider( G ) at /.../ctblgrp.gi:2194 called from
...  at *stdin*:2
```

(The error messages are not exactly the same in the two versions,
in the sense that they point to different positions in `<list>`.)

GAP 4.9.0 does not have this problem.
The same holds for GAP 4.7.6.
(And the computation in GAP 4.9.0 is much faster than that in GAP 4.7.6.)

rbehrends:
This pull request fixes two separate bugs, related to https://github.com/gap-system/gap/issues/3114.

1. The processor stack was not being marked correctly, as Julia
   made wrong assumptions about the end of the processor stack;
   we now use the value of `stack_bottom` that is being supplied
   to `InitBags()`.
2. `CHANGED_BAG()` was not being called on `CurrLVars` when a collection
   was initiated, unlike for the GASMAN implementation, where this
   happens in `VarsBeforeCollectBags()`.

Fixes #3114

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -364,6 +364,7 @@ static jl_datatype_t * datatype_mptr;
 static jl_datatype_t * datatype_bag;
 static jl_datatype_t * datatype_largebag;
 static UInt            StackAlignBags;
+static Bag *           GapStackBottom;
 static jl_ptls_t       JuliaTLS, SaveTLS;
 static size_t          max_pool_obj_size;
 static size_t          bigval_startoffset;
@@ -527,6 +528,9 @@ void GapRootScanner(int full)
     // current_task != root_task.
     char * stackend = (char *)jl_task_stack_buffer(task, &size, &tid);
     stackend += size;
+    if (JuliaTLS->tid == 0 && JuliaTLS->root_task == task) {
+        stackend = (char *)GapStackBottom;
+    }
 
     // allow installing a custom marking function. This is used for
     // integrating GAP (possibly linked as a shared library) with other code
@@ -589,6 +593,13 @@ static void PreGCHook(int full)
     // a thread-local variable.
     SaveTLS = JuliaTLS;
     JuliaTLS = jl_get_ptls_states();
+    // This is the same code as in VarsBeforeCollectBags() for GASMAN.
+    // It is necessary because ASS_LVAR() and related functionality
+    // does not call CHANGED_BAG() for performance reasons. CHANGED_BAG()
+    // is only called when the current lvars bag is being changed. Thus,
+    // we have to add a write barrier at the start of the GC, too.
+    if (STATE(CurrLVars))
+        CHANGED_BAG(STATE(CurrLVars));
     /* information at the beginning of garbage collections                 */
     SyMsgsBags(full, 0, 0);
     memset(MarkCache, 0, sizeof(MarkCache));
@@ -638,6 +649,7 @@ static uintptr_t JMarkBag(jl_ptls_t ptls, jl_value_t * obj)
 void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
 {
     // HOOK: initialization happens here.
+    GapStackBottom = stack_bottom;
     for (UInt i = 0; i < NUM_TYPES; i++)
         TabMarkFuncBags[i] = MarkAllSubBags;
     // These callbacks need to be set before initialization so


src/julia_gc.c
```diff
@@ -589,6 +593,13 @@ static void PreGCHook(int full)
     // a thread-local variable.
     SaveTLS = JuliaTLS;
     JuliaTLS = jl_get_ptls_states();
+    // This is the same code as in VarsBeforeCollectBags() for GASMAN.
+    // It is necessary because ASS_LVAR() and related functionality
+    // does not call CHANGED_BAG() for performance reasons. CHANGED_BAG()
+    // is only called when the current lvars bag is being changed. Thus,
+    // we have to add a write barrier at the start of the GC, too.
+    if (STATE(CurrLVars))
+        CHANGED_BAG(STATE(CurrLVars));
```
fingolfin: It's OK to have this in here for now, to get a quick fix; but on the long run, I prefer to fix this properly, by providing a generic "pre-gc hook" in both GASMAN (which already has one) and Julia GC, and then setting that to `VarsBeforeCollectBags`. Also, this "optimization" for lvars needs to be documented in a better place. But neither of these are change request for this PR; me or somebody else can take care of it in a future PR.

src/julia_gc.c
```diff
@@ -527,6 +528,9 @@ void GapRootScanner(int full)
     // current_task != root_task.
     char * stackend = (char *)jl_task_stack_buffer(task, &size, &tid);
     stackend += size;
+    if (JuliaTLS->tid == 0 && JuliaTLS->root_task == task) {
+        stackend = (char *)GapStackBottom;
```
fingolfin: Hmm, but note that this only will work right if GAP is the main process, loading Julia later. If, however, one is starting Julia and from there loading GAP, then `GapStackBottom ` may not be at the base of the stack frame.

This is closely related to (part of) the discussion on issue #3089 ;  see also PR #3096

src/julia_gc.c
```diff
@@ -545,7 +549,7 @@ void GapRootScanner(int full)
     for (Int i = 0; i < GlobalCount; i++) {
         Bag p = *GlobalAddr[i];
         if (IS_BAG_REF(p)) {
-            JMark(p);
+            TryMark(p);
```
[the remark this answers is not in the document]
rbehrends: Oops, that's actually a leftover from debugging and should be removed. I'll get rid of it.

src/julia_gc.c
rbehrends: I agree. My goal here was to get the fix out ASAP. As the code is identical to `VarsBeforeCollectBags()`, this should be unified.

src/julia_gc.c
rbehrends: Yes, and `GapTaskScanner()` may also need an update. That said, it should properly be handled so that Julia gets to know the proper end of the stack for task switching.

fingolfin:
I can confirm that this fixes #3114 for me; I hope @ThomasBreuer can confirm that as well (and then check which of the other crashes he was getting are also fixed; hopefully all.... ;-)

src/julia_gc.c
embray: ISTM there needs to be a `SetStackBottomBags` (as in #3096) for Julia GC as well.

fingolfin: Possible, but this code is very much in flux, so it doesn't really matter at this stage.

fingolfin:
Backported to stable-4.10 in af3bb41d34326ec2771ef71f1738a81c88cfab85

PR #3204, opened 2019-01-20
kernel: add 'static' keyword where possible

fingolfin:
This PR is somewhat dual to PR #3183 resp. #3202: it adds `static` to as many functions and variables as possible. This helps to find dead code (see e.g. PR #3203), may help compilers to do better optimizations, and also is a step towards defining a larger "libgap" API, by "hiding" away stuff that we do not explicitly want to export (this makes it harder to "cheat" and call those functions anyway; now people who want to do that hopefully will instead talk to us to get us to export them, giving us a chance to define a proper API).

The second commit is optional and clang-formats the first commit, based on the idea that if we touch those lines anyway, we might as well reformat them, too.

### DIFF
--- a/src/ariths.c
+++ b/src/ariths.c
@@ -44,7 +44,7 @@ ArithMethod1 ZeroFuncs [LAST_REAL_TNUM+1];
 */
 static Obj ZEROOp;
 
-Obj ZeroObject (
+static Obj ZeroObject (
     Obj                 obj )
 
 {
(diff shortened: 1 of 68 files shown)


ChrisJefferson:
Did you check this didn't break compiling / loading any packages?

fingolfin:
@ChrisJefferson the `testpackages` test on Travis verifies that all distributed packages still compile and load. This testing really works: it even found one issue in an earlier revision of this PR: the `cvec` package unfortunately has `extern unsigned long PolsFF[];` so I had to drop `static` from `PolsFF` -- at the same time, this is a nice example for exactly the kind of thing I'd like to prevent in the future: I'll work on fixing cvec to not require this (possibly by adding a suitable new API to the GAP kernel). Actually, I was already aware of this before, see https://github.com/gap-packages/cvec/issues/17. I'll try to get it addressed in time for GAP 4.11

PR #3208, opened 2019-01-21
Add missing CHANGED_BAG in PushValue

ChrisJefferson:
This adds a missing CHANGED_BAG, and also adds a couple of asserts which might fail in advanced testing (as how some of these variables get set is a little tricky to follow). Let's see.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -520,8 +520,12 @@ static Int PushValue(Obj val)
     Obj values = header->values;
     if (!values) {
         values = NEW_PLIST(T_PLIST, 4);
+        // Recalculate header in case NEW_PLIST caused a GC
         header = (BodyHeader *)STATE(PtrBody);
         header->values = values;
+        GAP_ASSERT(STATE(PtrBody) == (Stat *)PTR_BAG(BODY_FUNC(CURR_FUNC())));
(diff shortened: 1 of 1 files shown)


src/code.c
```diff
@@ -520,8 +520,12 @@ static Int PushValue(Obj val)
     Obj values = header->values;
     if (!values) {
         values = NEW_PLIST(T_PLIST, 4);
+        // Recalculate header in case NEW_PLIST caused a GC
         header = (BodyHeader *)STATE(PtrBody);
         header->values = values;
+        GAP_ASSERT(STATE(PtrBody) == (Stat*)PTR_BAG(BODY_FUNC(CURR_FUNC())));
```
fingolfin:
```suggestion
        GAP_ASSERT(STATE(PtrBody) == (Stat *)PTR_BAG(BODY_FUNC(CURR_FUNC())));
```

ChrisJefferson:
A test is failing, I've pushed a change to (hopefully) show backtraces so I can see what's failing.

ChrisJefferson:
The failing test has disappeared. I think there are still some other deep problems here, but this seems like an improvement, and we can continue searching for other problems.

PR #3209, opened 2019-01-21
Enable Backtraces with --enable-debug

ChrisJefferson:
This enables backtraces when GAP crashes, when GAP is compiled with --enable-debug. This will be (I hope!) particularly useful when GAP crashes on travis.

At the same time I made it so GAP will only print a backtrace when it actually crashes, not just because it was killed, or was terminated.


fingolfin:
How does this interact with gdb or lldb?

Issue #3212, opened 2019-01-22
`Gcd` no longer returns standard associates?


ThomasBreuer:
The following happens in GAP 4.10.0 as well as in the current master branch,
with or without packages loaded.

```
gap> x:= Indeterminate( Rationals, "x" );;
gap> g:= Gcd( x-2, x^2-2*x );
-1/2*x+1
gap> StandardAssociate( g );
x-2
```

Up to GAP 4.9, the value returned by `Gcd` was the standard associate.
Thanks to Meinolf Geck for this observation.

I think this changed behavior is a bug, the documentation of `Gcd` says:

> Gcd  returns  the standard associate (see StandardAssociate (56.5-5)) of the
> greatest common divisors.

fingolfin:
This regression was introduced in commit fb0cdaf59f532c8739a1c729d3c1cac1f2228c9c by @hulpke

alex-konovalov:
The fix has been backported to stable-4.10 in commit 9a6034d79f17a529f99cb9379922d98c63d11d9d

Issue #3205, opened 2019-01-21
Regression: when printing code, long literals are wrapped incorrectly


fingolfin:
Printing of function bodies with a long literals got broken somewhere between GAP 4.8 and GAP 4.9. 

Using `git bisect`, I determined that this is caused by commit c78bb4c4d89a8c17bb931675df6d39a87b9e32d3 added by @stevelinton on 2017-08-17.

With GAP 4.8, we see this:
```
gap> SizeScreen([80,100]);;
gap> Display(f->2.12345678901234567890123456789012345678901234567890123456789012345678901234567890);
function ( f )
    return
     2.12345678901234567890123456789012345678901234567890123456789012345678901\
234567890;
end
gap> Display(f->212345678901234567890123456789012345678901234567890123456789012345678901234567890);
function ( f )
    return
     2123456789012345678901234567890123456789012345678901234567890123456789012\
34567890;
end
gap> Display(f->"212345678901234567890123456789012345678901234567890123456789012345678901234567890");
function ( f )
    return
     "212345678901234567890123456789012345678901234567890123456789012345678901\
234567890";
end
```
With GAP 4.9 or master, we see this:
```
gap> SizeScreen([80,100]);;
gap> Display(f->2.12345678901234567890123456789012345678901234567890123456789012345678901234567890);
function ( f )
    return
     2.12345678901234567890123456789012345678901234567890123456789012345678901\
\
234567890;
end
gap> Display(f->212345678901234567890123456789012345678901234567890123456789012345678901234567890);
function ( f )
    return
     2123456789012345678901234567890123456789012345678901234567890123456789012\
\
34567890;
end
function ( f )
    return
     "212345678901234567890123456789012345678901234567890123456789012345678901\
\
234567890";
end
```

My guess is that this can be fixed by also adding a `Display` method to complement the `DisplayString` method.

fingolfin:
That simple fix seems to work... I'll make a PR

PR #3214, opened 2019-01-22
Add helper script for bisecting regressions

fingolfin:
This script makes it quite convenient to track down regression as in issue #3212 or #3205.

### DIFF
--- a/etc/bisect.sh
+++ b/etc/bisect.sh
@@ -1,0 +1,114 @@
+#!/bin/sh
+#
+#  This script can be used together with `git bisect run` to track down
+#  regressions. In a nutshell, `git bisect` performs a binary search on a
+#  range of commits to find the one which introduced a change in behavior
+#  (when branches and merge commits are involved, things are considerably more
+#  complicated; for further information, please read `git help bisect`).
+#
+#  To use this script, you need a .tst file that tests the regression.
+#  Moreover, you need to know at least one commit in which the bug is not
+#  (yet) present, and one where it is present. Finally, you must make
+#  a copy of this script, preferably outside the GAP directory. For example,
+#  copy it to /tmp/gap-bisect.sh.
+#
+#  Then, execute the following commands from the GAP root directory; note that
+#  the last command will usually checkout a different commit, so make sure
+#  you have a clean working tree.
+#       git bisect start
+#       git bisect bad KNOWN-BAD-COMMIT
+#       git bisect good KNOWN-GOOD-COMMIT
+#
+#  Now you normally need to test the commit selected by `git bisect` and then
+#  report whether it is "good" or "bad". This is where this script comes in:
+#  it can automate this for you using a .tst file, which means that you can
+#  drink a coffee while it does the bisecting work for you. To use it, issue
+#  the following command next:
+#
+#       git bisect run /PATH/TO/COPY/OF/bisect.sh YOUR_TEST_FILE.tst
+#
+#  Then git will start to check out revisions, and run this script, which in
+#  turn compiles GAP, and then runs the test file given to it; depending on
+#  the outcome of the test, the current commit is marked as bad or good, and
+#  so on, until the commit introducing the change is found.
+#
+#  Some usage hints and caveats:
+#
+#  For a known bad commit, typically you will use `master`, but sometimes also
+#  `stable-4.XYZ` may be useful.
+#
+#  For a known good commit, you can typically try to use `stable-4.X` for the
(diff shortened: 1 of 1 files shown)


fingolfin:
@ThomasBreuer Thanks for the feedback. I will add `git bisect start` to the instructions; I didn't add it because git offers to do it for me (and AFAICT, it always does that)
```
$ git bisect bad master
You need to start by "git bisect start"
Do you want me to do it for you [Y/n]?
```

And yes, all refs (such as `master` or `stable-4.9`) that one passes to `git bisect` must exist, just like with any other git command.

Regarding the "a mege base must be tested": I'll add a comment that `git bisect` will checkout all kinds of different revisions, and that one should read the `git bisect` man page before doing any of this.

Finally, on the error about `etc/bisect.sh` not being found: oops, that's of course a crucial point; I didn't consider that when I used it before. I guess I'll have to add instructions to first make a copy of the script somewhere.

fingolfin:
I tried to improve the text, please take another look and let me know what you think!

There is another issue that should be addressed, though: there is a range of ~200 commits (from f3da6b3812a293ca252e4a3b7a52f37007dabafb to d2ea52ef86a1c68b619f4004c586f07512c39c1d) in which GAP does not start for me, because it already requires `transpkg`, but is not compatible with the current version of `transpkg`: I get this error when starting GAP:
```
...
Error, BIND_GLOBAL: variable `TRANSREGION' must be unbound at ./lib/global.g:137 called from
BIND_GLOBAL( name, value ); at ./lib/global.gi:259 called from
<function "BindGlobal">( <arguments> )
 called from read-eval loop at ./pkg/transgrp/lib/trans.gd:79
you can 'return;' to continue
brk>
```
I need to come up with a way to workaround this: either by somehow convincing GAP to start anyway (e.g. apply a temporary fix to `lib/read4.g` which removes the offending lines -- but that seems fragile); or else by finding a way to detect this brokenness (unfortunately, the error happens before the custom `OnQuit` function is set), and then instruct git to skip the commit (but then git will just switch to the "next" commit... and since there are ~200 commits in a row that are broken, that is highly undesirable). 

Another idea for making GAP start with option `-A` for the commits in the named ranges: Perhaps the script can set up another fake GAPROOT dir in some tmp dir, then copy `read4.g` there, patch it, and instruct GAP to use that GAP dir as first root dir... Hrm, sounds rather complicated...

Any other (better) ideas?

fingolfin:
Seems to work fine now. As a concrete example, this is how one can reproduce my manual bisection of issue #3205: go to the GAP repository root dir, add a file `literals.tst` with this content:
```
gap> SizeScreen([80,100]);;
gap> Display(f->212345678901234567890123456789012345678901234567890123456789012345678901234567890);
function ( f )
    return 
     2123456789012345678901234567890123456789012345678901234567890123456789012\
34567890;
end
```
Then, initiate the bisection (you may need to run `git fetch --tags` first to get the `bisect-after-new-buildsys` tag):
```
cp etc/bisect.sh .
git bisect start
git bisect bad master
git bisect good bisect-after-new-buildsys
git bisect run ./bisect.sh literals.tst
```

This runs for a while, and eventually ends this way:
```
...
c78bb4c4d89a8c17bb931675df6d39a87b9e32d3 is the first bad commit
commit c78bb4c4d89a8c17bb931675df6d39a87b9e32d3
Author: Steve Linton <steve.linton@st-andrews.ac.uk>
Date:   Thu Aug 17 16:06:27 2017 +0100

    Add String and DisplayString methods for functions
    A bit of a hack as it uses string streams rather than modifying the kernel
    Includes tests

:040000 040000 5573f8f6f728d34e80a3af86056e27779e0fc549 11ac47f8465cf180a013fae17020de06b454c049 M	lib
:040000 040000 424c39553c16341e5f338a841f4f86939875e504 10ab5039a5eec6a041c3e05b54afa954ef697244 M	tst
bisect run success
$
```

ThomasBreuer:
@fingolfin I am still running into error messages when my configure options include
something like `with-gc=julia`, which is not allowed in older versions.
Perhaps `etc/bisect.sh` should mention that one should switch to a setup without such options
before running the script?

fingolfin:
@ThomasBreuer running `make` will indeed re-run `configure`, with the options last used, and so yes, this won't work if you use options that are not supported by one of the tested commits. I can mention this in the documentation comment, but in the end, there are tons of caveats that one could add, and I cannot possible cover them all.

fingolfin:
Added a warning to the doc comment about rebuilding and configure options

PR #3216, opened 2019-01-22
FIx overlapping memcpy in APPEND_LIST

ChrisJefferson:
I found this while trying to figure out why XModAlg crashes in startup sometimes.

I don't think this can be the problem... but it technically an error, which is triggered when loading XModAlg.


src/listfunc.c
```diff
@@ -266,8 +266,9 @@ Obj             FuncAPPEND_LIST_INTR (
         GROW_STRING(list1, len1 + len2);
         SET_LEN_STRING(list1, len1 + len2);
         CLEAR_FILTS_LIST(list1);
-        // copy data, including terminating zero byte
-        memcpy(CHARS_STRING(list1) + len1, CONST_CHARS_STRING(list2), len2 + 1);
+        // copy data, but not terminating byte, in case list1==list2
+        memcpy(CHARS_STRING(list1) + len1, CONST_CHARS_STRING(list2), len2);
+        CSTR_STRING(list1)[len1 + len2] = '\0';
```
fingolfin: Good catch. I guess alternatively we could use `SyMemmove`?

fingolfin: (I am not saying we should, just mumbling :)

ChrisJefferson: That's probably better, done.

fingolfin:
Backported to stable-4.10 in 67e51c02c933446221cafd80cee01a2316c75d16

PR #3217, opened 2019-01-22
Move from INTOBJ_INT and SET_ELM_PLIST to safer methods

ChrisJefferson:
This changes some occurrences of INTOBJ_INT and SET_ELM_PLIST to the safer ObjInt_Int and ObjInt_UInt. These are cases where we aren't in a tight loop so it isn't critical.

As part of this I switched the PushPlist rather than SET_ELM_PLIST, mainly as this correctly deals with CHANGED_BAG, which might now be required.


src/calls.c
```diff
@@ -1388,7 +1388,7 @@ Obj FuncNARG_FUNC (
     Obj                 func )
 {
     if ( TNUM_OBJ(func) == T_FUNCTION ) {
-        return INTOBJ_INT( NARG_FUNC(func) );
+        return ObjInt_Int(NARG_FUNC(func));
```
fingolfin: That really can't matter:
```C
EXPORT_INLINE Int NARG_FUNC(Obj func)
{
    return INT_INTOBJ(CONST_FUNC(func)->nargs);
}
```
If you are deeply concerned about it, perhaps this instead?
```suggestion
        return CONST_FUNC(func)->nargs;
```

src/calls.c
```diff
@@ -1586,7 +1586,7 @@ Obj FuncSTARTLINE_FUNC(Obj self, Obj func)
     if (BODY_FUNC(func)) {
         UInt sl = GET_STARTLINE_BODY(BODY_FUNC(func));
         if (sl)
-            return INTOBJ_INT(sl);
+            return ObjInt_UInt(sl);
```
fingolfin: Similar to the above, we have:
```C
UInt GET_STARTLINE_BODY(Obj body)
{
    Obj line = BODY_HEADER(body)->startline_or_location;
    return IS_POS_INTOBJ(line) ? INT_INTOBJ(line) : 0;
}
```
So this code here could be:
```
  Obj sl = BODY_HEADER(body)->startline_or_location;
  if (IS_POS_INTOBJ(sl))
    return sl;
```

src/calls.c
```diff
@@ -1598,7 +1598,7 @@ Obj FuncENDLINE_FUNC(Obj self, Obj func)
     if (BODY_FUNC(func)) {
         UInt el = GET_ENDLINE_BODY(BODY_FUNC(func));
         if (el)
-            return INTOBJ_INT(el);
+            return ObjInt_UInt(el);
```
fingolfin: and again the same story

src/gap.c
```diff
@@ -535,19 +535,18 @@ Obj FuncRETURN_NOTHING (
...
   res = NEW_PLIST(T_PLIST, 4);
-  SET_LEN_PLIST(res, 4);
-  SET_ELM_PLIST(res, 1, INTOBJ_INT( SyTime() ));
-  SET_ELM_PLIST(res, 2, INTOBJ_INT( SyTimeSys() ));
-  SET_ELM_PLIST(res, 3, INTOBJ_INT( SyTimeChildren() ));
-  SET_ELM_PLIST(res, 4, INTOBJ_INT( SyTimeChildrenSys() ));
+  PushPlist(res, ObjInt_UInt(SyTime()));
+  PushPlist(res, ObjInt_UInt(SyTimeSys()));
+  PushPlist(res, ObjInt_UInt(SyTimeChildren()));
+  PushPlist(res, ObjInt_UInt(SyTimeChildrenSys()));
```
fingolfin: Note that the code using `RUNTIMES` relies on the various parts being in a specific position each. So I kind of like the explicit indices, as it makes it easy to compare this code with its high-level wrapper in the GAP library and verify that they match. Perhaps if this code was changed to simply use `ASS_LIST` instead of `SET_ELM_PLIST`? That should be as safe or "safer" than `PushPlist`, but retain the explicit enumeration.

Of course it's only four entries here, so it's not a bridge I am going to die on, but I feel that it needlessly increases the cognitive load here.

src/gap.c
```diff
@@ -1159,9 +1156,9 @@ Obj FuncMASTER_POINTER_NUMBER(Obj self, Obj o)
     }
 #ifdef USE_GASMAN
     if ((void **) o >= (void **) MptrBags && (void **) o < (void **) MptrEndBags) {
-        return INTOBJ_INT( ((void **) o - (void **) MptrBags) + 1 );
+        return ObjInt_UInt(((void **)o - (void **)MptrBags) + 1);
     } else {
-        return INTOBJ_INT( 0 );
+        return ObjInt_UInt(0);
```
fingolfin: I really dislike this particular change: `INTOBJ_INT(0)` is a compile time constant, and I deliberately use that all over the code.

src/intfuncs.c
```diff
@@ -461,7 +462,7 @@ Obj FuncHASHKEY_BAG(Obj self, Obj obj, Obj seed, Obj offset, Obj maxlen)
       n = imaxlen;
   }
 
-  return INTOBJ_INT(HASHKEY_BAG_NC(obj, (UInt4)s, offs, (int)n));
+  return ObjInt_Int(HASHKEY_BAG_NC(obj, (UInt4)s, offs, (int)n));
```
fingolfin: But we deliberately wrote `HASHKEY_MEM_NC` (and thus code using it, like `HASHKEY_BAG_NC`) to return an integer that fits into an immediate integer object. And there is code that explicitly relies on `HASHKEY_BAG` returning an immediate integer value. I feel this change obscures this. At the very least, we'd have to do somthing lke this:
```C
  Obj result = ObjInt_Int(HASHKEY_BAG_NC(...));
  if (!IS_INTOBJ(result)) Panic("this must never happen");
  return result;
```
But it seems better to me to just leave it alone, but add comment saying something like
```C
// HASHKEY_BAG_NC is guaranteed to return a value that fits into an immediate integer
```
and then perhaps we should also insert such comments into `HASHKEY_BAG_NC` and `HASHKEY_MEM_NC`

src/modules.c
```diff
@@ -359,7 +359,6 @@ Obj FuncSHOW_STAT(Obj self)
 
     /* make a list of modules with crc values                              */
     modules = NEW_PLIST(T_PLIST, 2 * im);
-    SET_LEN_PLIST(modules, 2 * im);
 
     for (k = 0, im = 1; CompInitFuncs[k]; k++) {
```
fingolfin:
```suggestion
     for (k = 0; CompInitFuncs[k]; k++) {
```
... and then get rid of the declaration of `im`, too.

src/calls.c
ChrisJefferson: As a general argument, I'd prefer that we use `ObjInt_Int` (and friends) except in places where it really matters -- there is an issue of philosophy here. I increasingly feel we should move towards using methods which are guaranteed not to lead to corrupted / broken states, except in a small number of places where we can carefully verify correctness.

src/gap.c
ChrisJefferson: I'm happy with using ASS_LIST instead

src/intfuncs.c
ChrisJefferson: I agree this makes it harder to read, but I'd also feel happier if we checked the answer did fit in an immediate integer (and in this case, I don't think the overhead is painful). I could look again into a CheckedINTOBJ_INT

src/calls.c
fingolfin: I have a somewhat differing view: in many places, I'd agree that `ObjInt_Int` is better (including many covered by this PR). But I feel that sometimes it just obscures what is happening. And this is a prime example. It feels like we are building an onion out of abstraction layers here, and I'd feel much better if we just cut them away and wrote code that avoided pointless back and forth conversions altogether -- which is precisely what the change I suggested here does.

src/intfuncs.c
fingolfin: I don't mind the overhead of the check; I mind that we first go to great length to create a value that fits, and then apply a bludgeon to it "just in case". If we are concerned that we got the code to create the values wrong, then we should add assertions and/or `if (...) Panic(...)` checks there, right after that. Then change `HASHKEY_MEM_NC` and `HASHKEY_BAG_NC` to return an immediate integer as `Obj`. Then adapt the few callers to that (it's trivial in all cases, I think -- no need to convert back there with `INT_INTOBJ` or `Int_ObjInt`)... except that unfortunately, the `datastructures` package uses `HASHKEY_MEM_NC` and `HASHKEY_WHOLE_BAG_NC`, so we can't just change the API *sigh*. But we could "deprecate" them, and add new variants which return `Obj`...

ChrisJefferson:
So, I think all issues are fixed now. It might be the case we could do a more significant rewrite of some parts, but to be honest my preference is on simple changes which definitely improve correctness, rather than more significant changes which might break things.

src/intfuncs.c
```diff
@@ -16,6 +16,7 @@
 #include "bool.h"
 #include "calls.h"
 #include "error.h"
+#include "integer.h"
```
fingolfin: I guess this can go now?

src/calls.c
fingolfin: I agree with all of this PR *except* for the function body related changes, where I'd prefer the alternative changes I suggested, or no change at all.

PR #3220, opened 2019-01-23
kernel: fix wrong result bugs in partial perms code

fingolfin:
Various computations involving T_PERM2 objects of degree 65536 returned
incorrect results. The reason was that the degree, which is always an UInt,
was stored in an UInt2. But for degree 65536, this overflows to zero, hence
the code effectively treated the permutation as the identity.

The fix is to always store the degree as UInt, not as UInt2 or UInt4 (the
latter case actually is probably fine, but let's do it right anyway).

The fix in PowPPerm2Perm2 is a bit nasty and constitutes a deoptimization; a
proper fix at this stage would involve duplicating and tweaking a bunch of
code; but I plan to rewrite this file to use C++ templates anyway, at which
point the proper fix will come almost for free. So instead of wasting time on
a "proper" fix now, let's live with this workaround for now.

### DIFF
--- a/src/pperm.cc
+++ b/src/pperm.cc
@@ -2843,8 +2843,8 @@ Obj ProdPPerm22(Obj f, Obj g)
     GAP_ASSERT(TNUM_OBJ(f) == T_PPERM2);
     GAP_ASSERT(TNUM_OBJ(g) == T_PPERM2);
 
-    UInt   deg, degg, i, j, rank;
-    UInt2 *ptf, *ptg, *ptfg, codeg;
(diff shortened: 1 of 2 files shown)


fingolfin:
I suggest we backport this to 4.10, chance for regression should be negligible.

@james-d-mitchell ping (just want to make sure you are aware of this PR)

james-d-mitchell:
Thanks for the ping @fingolfin I'm also happy with this change.

fingolfin:
Backported to stable-4.10 via ab4dc53ec3ad316a451ce611eb01ee406b98ad68

PR #3221, opened 2019-01-23
Handle stack overflow in Attributes

### LINKED ISSUE #1286: Crash with perverse method
The following (slightly perverse) method causes a crash rather than an error:

```
gap> DeclareAttribute("XXX",IsGroup);
gap> InstallMethod(XXX,[IsGroup],XXX);
gap> XXX(SymmetricGroup(5));
Segmentation fault: 11
```

ChrisJefferson:
The main purpose of this patch is to fix stack-depth-operation.g.
Rather than handle that specially, move where we handle stack
depth, to consistently handle all problems.

Fixes #1286

If anyone can think of another way we could cause a stack overflow, let me know here and I'll add it as a test.


src/funcs.c
```diff
@@ -406,10 +406,8 @@ void RecursionDepthTrap( void )
 
 #define CHECK_RECURSION_BEFORE \
             HookedLineIntoFunction(func); \
-            CheckRecursionBefore();
 
 #define CHECK_RECURSION_AFTER \
-            DecRecursionDepth(); \
             HookedLineOutFunction(func);
```
fingolfin: Nitpick, but: the names of those macros now are kinda wrong...

fingolfin:
Of course this probably introduces a small performance penalty, but I think that not crashing by accident in the middle of a long a and complicated computation beats that.

Though I do wonder how bad the overhead is in practice? It should only affect non-GAP functions, i.e., kernel functions. Perhaps we can do some micro benchmarking to test it? Hopefully that would show it is negligible. (I'm not insisting on this, I already approved the PR; just wondering out loud).

ChrisJefferson:
I did some very basic benchmarking and couldn't measure it. If we wanted to speed things up, it would probably be worth pulling the functions I'm calling into a header, so they can be inlined.

PR #3222, opened 2019-01-23
kernel: add TrimPerm helper

fingolfin:


src/permutat.cc
```diff
@@ -2303,23 +2280,32 @@ Obj Array2Perm (
...
+    }
+    else if (m <= 65536) {
+        UInt2 * ptr2 = ADDR_PERM2(perm);
+        UInt4 * ptr4 = ADDR_PERM4(perm);
+        for (UInt k = 1; k <= m; k++) {
             ptr2[k-1] = ptr4[k-1];
         };
-        RetypeBag( perm, T_PERM2 );
+        RetypeBag(perm, T_PERM2);
+        GAP_ASSERT(m <= DEG_PERM4(perm));
```
ChrisJefferson: Minor, but I feel like this line should be before 2296 -- we might want to forbid calling DEG_PERM4 on T_PERM2 bags at some point?

fingolfin: Agreed. And I guess I could make ptr4 const.

Issue #3166, opened 2019-01-09
Add GAP interface to setenv


gaehler:
By analogy programming, I managed to write (in package Carat) a dynamical kernel module to set an environment variable in the main GAP process. The idea is to teach CARAT programs the current location of their data tables, overriding the value determined at compile time.

It works, but I first got a warning that a dynamic module combined with a static zlib is not portable. After I installed zlib from the distribution (Ubuntu 18.04), the warning went away - but why is the bundled zlib linked statically, if this is not portable?

Also, I wondered whether a GAP interface to getenv and expecially setenv would not be of general interest, rather than implementing this in a package.

I'm using:
GAP 4.10.0 of 01-Nov-2018
Architecture: x86_64-pc-linux-gnu-default64
Configuration:  gmp 6.1.2, readline

ChrisJefferson:
Can you give a list of instructions which cause the error about a static zlib not being portable? I've never seen it before.

markuspf:
Only marginally related: If you're calling `setenv`, I suggest adding this syscall to the `io` package.

gaehler:
@ChrisJefferson Remove Ubuntu package zlib1g-dev, configure and make GAP again, and then, in the GitHub version of GAP-package Carat:
```
./configure <path-to-GAP-root>
make dynmod
```
This produces:
```
/homes/gaehler/gap/gap-current/gac -d -o bin/x86_64-pc-linux-gnu-default64/setcaratdir.so src/setcaratdir.c
/bin/bash /homes/gaehler/gap/gap-4.10.0/libtool --mode=compile gcc -g -O2 -o /tmp/gaclYGsNef/7607_setcaratdir.lo -I/homes/gaehler/gap/gap-4.10.0/gen -I/homes/gaehler/gap/gap-4.10.0/src -I/homes/gaehler/gap/gap-4.10.0 -DHAVE_CONFIG_H -I/homes/gaehler/gap/gap-4.10.0/extern/install/zlib/include -c src/setcaratdir.c
libtool: compile:  gcc -g -O2 -I/homes/gaehler/gap/gap-4.10.0/gen -I/homes/gaehler/gap/gap-4.10.0/src -I/homes/gaehler/gap/gap-4.10.0 -DHAVE_CONFIG_H -I/homes/gaehler/gap/gap-4.10.0/extern/install/zlib/include -c src/setcaratdir.c  -fPIC -DPIC -o /tmp/gaclYGsNef/.libs/7607_setcaratdir.o
[...]

PR #3225, opened 2019-01-23
Switch builtin zlib to use dynamic linking

### LINKED ISSUE #3049: Builtin zlib: Fix incompatbility with libgap.la
When GAP is built, it wants to link against zlib. If we can't find headers and libraries for zlib, we build our own version of that. But for simplicity, we only build a static linked version.

Unfortunately, this is a problem when linking `libgap.la`, which now triggers on most (all?) systems a warning or even an error. And of course in `master`, we now build `libgap.la` as part of the default target, making this more prominent

Several fixes come to mind:
1. re-enable building of zlib as a dynamic library (possibly with new custom build rules, e.g. using our `libtool` executable)
2. Don't build `libgap.la` if we use our builtin zlib (or at least don't build it as part of the default target)
3. Replace zlib with https://github.com/richgel999/miniz (or something similar) and link that as part of the GAP binary, not an external library, and make sure not to export any of it symbols; that way, it remains fully transparent to the outside world. UPDATE: Unfortunately, miniz does not provide gzopen, gzread etc., so that's not as trivial a change as I originally hoped.

ChrisJefferson:
As in title.

Fixes #3049 and #3166

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -404,15 +404,7 @@ AS_IF([test $BUILD_ZLIB = no],
 AS_IF([test x$BUILD_ZLIB = xyes],[
   BUILD_ZLIB=yes
   ZLIB_CPPFLAGS='-I${abs_builddir}/extern/install/zlib/include'
-  AS_CASE([$host_os],
-    [*cygwin*],[
-      ZLIB_LDFLAGS='-L${abs_builddir}/extern/install/zlib/lib/ -lz'
-    ],
-    [*],[
-      ZLIB_LDFLAGS='${abs_builddir}/extern/install/zlib/lib/libz.a'
-    ]
-  )
-
+  ZLIB_LDFLAGS='-L${abs_builddir}/extern/install/zlib/lib/ -Wl,-rpath -Wl,${abs_builddir}/extern/install/zlib/lib/ -lz'
   ZLIB_LIBS=
 ])
 


ChrisJefferson:
Personally I hate dynamic linking, as it makes it hard to move programs to another directory, or computer. But this seems like the least work option to make various things work, and I'll just static link my own GAP.

ChrisJefferson:
Now I remember why I didn't dynamically link -- because zlib doesn't distribute a libtool file, and it's a pain in the ass to make libtool just use the library I tell it to.

configure.ac
```diff
@@ -404,15 +404,7 @@ AS_IF([test $BUILD_ZLIB = no],
...
-  AS_CASE([$host_os],
-    [*cygwin*],[
-      ZLIB_LDFLAGS='-L${abs_builddir}/extern/install/zlib/lib/ -lz'
-    ],
-    [*],[
-      ZLIB_LDFLAGS='${abs_builddir}/extern/install/zlib/lib/libz.a'
-    ]
-  )
-
+  ZLIB_LDFLAGS='-L${abs_builddir}/extern/install/zlib/lib/ -Wl,-rpath -Wl,${abs_builddir}/extern/install/zlib/lib/ -lz'
```
fingolfin: The one issue with that is that it won't work with `make install` at all. But that is of course a general problem when building with our "builtin zlib". It is less of a problem for our built-in GMP, as that also is using libtool, and so we can  implement a `make install` which properly re-links all libraries and binaries during installation with relatively little effort.

Anyway, I suspect the best long-term fix would be to simply replace the zlib build system with a custom one written by us; I might look into that at some point.

But anyway, if this change fixes a real world issue for somebody/something, let's merge it; in the very worst case, we could still revert it anyway.

Issue #338, opened 2015-11-04
`last` can be used as a variable without a warning


james-d-mitchell:
You can do the following in GAP:

```
gap> last := x -> x;
```

without an error being given. Is this the correct behaviour? 

Another example:

```
gap> 1;
1
gap> 2;
2
gap> last2 := 10;
10
gap> last2;
2
```

I think this could be rather confusing for non-expert users of GAP who are perhaps not aware that `last`, `last2`, and `last3` are "special" variables.

fingolfin:
The behavior is correct, but perhaps not desirable. We could make those variables read-only, but then the code modifying them needs to be updated.

Of course users could then still modify them, it's just more work. Indeed:
```
gap> MakeReadWriteGVar("fail");
gap> fail:=1;; fail;
1
gap> MakeReadWriteGVar("infinity");
gap> infinity:=42;; infinity;
42
```

stevelinton:
I would support these being ReadOnly globals. While people can still modify them if they have a reason to, it would reduce the risk of accidentally overwriting them, or having your own value overwritten by the system. The same would presumably apply to ~ and time.

ChrisJefferson:
This would break demo.g, and packages which cut+pasted it, but it would be easy to add an explicit "clearLasts" function to replace their usage.

fingolfin:
Indeed, demo.g should be rewritten anyway; I greatly dislike that it tries to "emulate" the main loop, including the `last`/`last2`/`last3` variables and `time`. If we could refactor the code in a way that these things are done in a central place, then future change to this area would be easier. 
The mere fact that people have to copy & edit it for their packages indicates something could be improved; and lastly the dreaded use of `SuPeRfail` should be gotten rid of ... ;-).

Note that `Shell()` in gap.c is the other place modifying these. And the code for `MakeReadOnlyGlobal` currently explicitly does this:
```
    if name in ["time", "last", "last2", "last3", "~"] then
        Error("Making ",name," read-only is not a good idea!");
    fi;
```

PR #3226, opened 2019-01-23
Make "last" (and similar variables) read-only

ChrisJefferson:
This stops users writing to `last`, `time`, and some other variables, by making them read-only.

It adds a method to write to these variables, `UpdateStat`, mainly for use in `demo.g`.

I realise this might break some code. However, I did have someone had a had-to-track down bug where they were using `last` as a variable themselves, and were then confused when they got stupid answers out, so I think it is worth changing.

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1606,7 +1606,7 @@ static Int PostRestore (
     StructInitInfo *    module )
 {
     /* construct the `ViewObj' variable                                    */
-    ViewObjGVar = GVarName( "ViewObj" ); 
+    ViewObjGVar = GVarName( "ViewObj" );
 
     /* construct the last and time variables                               */
     Last              = GVarName( "last"  );
@@ -1614,8 +1614,7 @@ static Int PostRestore (
     Last3             = GVarName( "last3" );
     Time              = GVarName( "time"  );
     MemoryAllocated   = GVarName( "memory_allocated"  );
-    AssGVar_NoReadOnlyCheck(Time, INTOBJ_INT(0));
-    AssGVar_NoReadOnlyCheck(MemoryAllocated, INTOBJ_INT(0));
+
     QUITTINGGVar      = GVarName( "QUITTING" );
     
     /* return success                                                      */
@@ -1642,6 +1641,13 @@ static Int InitLibrary (
 #else
     AssConstantGVar( GVarName( "IsHPCGAP" ), False );
 #endif
+
+    AssReadOnlyGVar(GVarName("last"), Fail);
+    AssReadOnlyGVar(GVarName("last2"), Fail);
+    AssReadOnlyGVar(GVarName("last3"), Fail);
+    AssReadOnlyGVar(GVarName("time"), INTOBJ_INT(0));
+    AssReadOnlyGVar(GVarName("memory_allocated"), INTOBJ_INT(0));
+
 
     /* return success                                                      */
     return PostRestore( module );
--- a/lib/init.g
+++ b/lib/init.g
@@ -95,7 +95,6 @@ Ignore := function ( arg )  end;
 ##
 SetFilterObj := "2b defined";
 infinity := "2b defined";
-last:="2b defined";
 
 
 #############################################################################


lib/demo.g
```diff
@@ -47,11 +47,11 @@ BindGlobal( "Demonstration", function( file )
...
         result:=READ_COMMAND_REAL( input, true ); # Executing the command.
-        time := Runtime()-storedtime;
+        UpdateStat("time", Runtime()-storedtime);
         if Length(result) = 2 then
-            last3 := last2;
-            last2 := last;
-            last := result[2];
+            UpdateStat("last3", last2);
+            UpdateStat("last2", last);
+            UpdateStat("last", result[2]);
```
fingolfin: Since you are touching this code anyway, perhaps you'd also be willing to teach it to update `memory_allocated`, too? Also perfectly fine if you don't want to, of course!

src/gap.c
```diff
@@ -283,8 +278,9 @@ Obj Shell ( Obj context,
         
     /* stop the stopwatch                                          */
     if (setTime) {
-      AssGVar( Time, INTOBJ_INT( SyTime() - time ) );
-      AssGVar(MemoryAllocated, ObjInt_Int8(SizeAllBags - mem));
+        AssGVar_NoReadOnlyCheck(Time, INTOBJ_INT(SyTime() - time));
```
fingolfin:
```suggestion
        AssGVar_NoReadOnlyCheck(Time, ObjInt_Int(SyTime() - time));
```

src/gap.h
```diff
@@ -40,6 +40,16 @@ extern UInt Last2;
 */
 extern UInt Last3;
 
+/****************************************************************************
+**
+*V  UpdateLast
```
fingolfin:
```suggestion
*F  UpdateLast(<newLast>, <lastDepth>)
```

src/gvars.c
```diff
@@ -307,6 +307,20 @@ static Obj * ELM_COPS_PLIST(Obj cops, UInt i)
...
+        ErrorMayQuit("Variable: '%g' is read only", (Int)NameGVar(gvar), 0);
+    }
+
+    AssGVar_NoReadOnlyCheck(gvar, val);
+}
+
+// This is a kernel-only variant of AssGVar which will change read-only
+// variables, which is used for constants like:
+// Time, MemoryAllocated, last, last2, last3
+void AssGVar_NoReadOnlyCheck(UInt gvar, Obj val)
```
fingolfin: Maybe `ForceAssGVar`?

src/scanner.c
```diff
@@ -845,7 +845,6 @@ static void GetChar(void)
 {
   /* skip '\''                                                           */
   Char c = GET_NEXT_CHAR();
-
```
fingolfin: ?

src/streams.c
```diff
@@ -396,15 +396,13 @@ static void READ_TEST_OR_LOOP(Obj context)
         type = ReadEvalCommand(context, &evalResult, &dualSemicolon);
 
         /* stop the stopwatch                                              */
-        AssGVar( Time, INTOBJ_INT( SyTime() - oldtime ) );
+        AssGVar_NoReadOnlyCheck(Time, INTOBJ_INT(SyTime() - oldtime));
```
fingolfin:
```suggestion
        AssGVar_NoReadOnlyCheck(Time, ObjInt_Int(SyTime() - oldtime));
```

src/gap.c
```diff
@@ -1484,6 +1480,42 @@ Obj FuncTHREAD_UI(Obj self)
...
+        AssGVar_NoReadOnlyCheck(Last2, ValGVarTL(Last));
+    if (lastDepth >= 1)
+        AssGVar_NoReadOnlyCheck(Last, newLast);
+}
+
+void FuncUpdateStat(Obj self, Obj name, Obj newStat)
+{
+    if (!IsStringConv(name) || !IS_STRING_REP(name)) {
+        ErrorMayQuit("UpdateStat: First argument must be a string", 0, 0);
+    }
```
fingolfin: Maybe use `RequireStringRep`, i.e.,
```C
    RequireStringRep("UpdateStat", name);
```

src/gap.c
```diff
@@ -1484,6 +1480,42 @@ Obj FuncTHREAD_UI(Obj self)
...
+        AssGVar_NoReadOnlyCheck(Last2, newStat);
+    }
+    else if (strcmp(cname, "last3") == 0) {
+        AssGVar_NoReadOnlyCheck(Last3, newStat);
+    }
+    else if (strcmp(cname, "memory_allocated") == 0) {
+        AssGVar_NoReadOnlyCheck(MemoryAllocated, newStat);
+    }
+    else {
+        ErrorMayQuit("UpdateStat: Unknown Stat '%g'", (Int)name, 0);
```
fingolfin:
```suggestion
        ErrorMayQuit("UpdateStat: unsupported <name> value '%g'", (Int)name, 0);
```

src/gap.c
```diff
@@ -1581,16 +1616,20 @@ static Int PostRestore (
...
     /* construct the last and time variables                               */
     Last              = GVarName( "last"  );
     Last2             = GVarName( "last2" );
     Last3             = GVarName( "last3" );
     Time              = GVarName( "time"  );
     MemoryAllocated   = GVarName( "memory_allocated"  );
-    AssGVar(Time, INTOBJ_INT(0));
-    AssGVar(MemoryAllocated, INTOBJ_INT(0));
+
+    AssReadOnlyGVar(GVarName("last"), Fail);
```
fingolfin:
```suggestion
    AssReadOnlyGVar(Last Fail);
```

src/gap.c
```diff
@@ -1581,16 +1616,20 @@ static Int PostRestore (
...
+    AssReadOnlyGVar(GVarName("last2"), Fail);
```
fingolfin:
```suggestion
    AssReadOnlyGVar(Last2, Fail);
```

src/gap.c
```diff
@@ -1581,16 +1616,20 @@ static Int PostRestore (
...
+    AssReadOnlyGVar(GVarName("last3"), Fail);
```
fingolfin:
```suggestion
    AssReadOnlyGVar(Last3, Fail);
```

src/gap.h
```diff
@@ -40,6 +40,16 @@ extern UInt Last2;
...
+**
+**  Updates Last, Last2 and Last3 by the new result 'newLast'.
+**  lastDepth chooses any how of these variables should be updated, starting
```
fingolfin:
```suggestion
**  <lastDepth> determines how many of these variables should be updated, starting
```

doc/ref/mloop.xml
```diff
@@ -161,6 +161,24 @@ Var="time"/>. This is an integer that holds the number of
...
+<ManSection>
+<Heading>UpdateStat</Heading>	
+<Func Name="UpdateStat" Arg='name, value'/>
+
+<Description>
+<Ref Func="UpdateStat"/> assigns the variable with name <A>name</A> the
+value <A>value</A>. Valid values for <A>name</A> in this version of GAP are
+<C>"last"</C>, <C>"last2"</C>, <C>"last3"</C>, <C>"time"</C> and
+<C>"memory_allocated"</C>. This does not disable these values being
+automatically updated at the end of each executed statement.
```
fingolfin: Perhaps like this?
> This does not disable the automatic updating of these variables performed by the interpreter,
> e.g. in interactive sessions.

Also, I think it would be a good idea to add a blurb that explains why one would want to use this function (namely for demos, and ideally one should not use it for anything else...

That said, I am not even sure that we really should document this function. Right now, demo code needs is very low level and needs to access various other low level and undocumented functions. Perhaps it would be better to hold off documenting this, until we had time to revise the Demo code -- if done right, this might get rid of the need for 3rd parties to roll their own variants of that. After all, this is a rather low level aspect of the system, and we don't document many other aspects of it, e.g. `SHELL`...

src/scanner.c
ChrisJefferson: I often add a printf there to see where GAP has got to when something bad happens during startup. Did that to find `last := "2b defined"`, then haven't cleaned up properly :)

src/gap.c
```diff
@@ -1484,6 +1480,42 @@ Obj FuncTHREAD_UI(Obj self)
...
+{
+    if (lastDepth >= 3)
+        AssGVar_NoReadOnlyCheck(Last3, ValGVarTL(Last2));
+    if (lastDepth >= 2)
+        AssGVar_NoReadOnlyCheck(Last2, ValGVarTL(Last));
+    if (lastDepth >= 1)
+        AssGVar_NoReadOnlyCheck(Last, newLast);
+}
+
+void FuncUpdateStat(Obj self, Obj name, Obj newStat)
```
fingolfin:
```suggestion
static void FuncUpdateStat(Obj self, Obj name, Obj newStat)
```

src/gvars.c
ChrisJefferson: The problem with force is this removes the 'ReadOnly' check, but not the 'Constant' check (because we use that to improve parsing, and we really can't change those, ever).

lib/demo.g
ChrisJefferson: I've never actually used this code, and don't even know how it should be used, so I'd prefer to leave it. I've just noticed it has no tests, no documentation, and isn't linked to from anywhere?

doc/ref/mloop.xml
ChrisJefferson: I all-caps it, and moved a (briefer) version of the docs inline

ChrisJefferson:
I think all issues are now fixed.

lib/demo.g
fingolfin: It is used like this: you take a file and put a bunch of inputs in it, each on its own line. Then you run `Demonstration` on this; whenever you press a key, the next input line is read and executed. The result then looks something like this (where each `demo>` appeared only after a key was pressed):
```
gap> Demonstration("demo.g");

Start of demonstration.

demo> 1+1;
2
demo> Factorial(10);
3628800
demo>
End of demonstration.

gap>
```

Updating last, last2, etc. only is of interest if the demo code references them. So if one wants to write a demo that showcases those, and also `memory_allocated`, then it'd be nice to have `memory_allocated`. But this is nothing this PR needs to do.

I've long wanted to give this code some love, and improve it; e.g. instead of always going to the next command upon any key press, I think only "return" on an empty line should do that; otherwise, it'd be nice if one could show the next command w/o executing it, and also edit it; and other stuff. I think other people implemented such demo modes for other languages, and they can be very useful for classes, presentations during talks etc.

src/gvars.c
fingolfin: My concern about `AssGVar_NoReadOnlyCheck` is that the underscore breaks our naming conventions. So I'd prefer `AssGVarNoReadOnlyCheck` or `AssGVarIgnoreReadOnly` or whatever. I also still think `ForceAssGVar` is good, as to me at least it is kinda clear that it only "forces" / ignores "optional" restrictions, but won't violate an important system invariant like constness. But of course that's just my view, if others (you) find this misleading, then we should avoid it.

ChrisJefferson: I've gone for `AssGVarWithoutReadOnlyCheck`. My main problem with `Force` is just I could imagine a bunch of things I could force.

ChrisJefferson:
Fixes #338 ( mentioned here so it will get closed)

PR #3228, opened 2019-01-23
kernel: split GVAR_FILTER into GVAR_{ATTR,FILT,PROP}

fingolfin:
While on a purely technical level, there is no difference between a C
function implementing a GAP function vs. a filter vs. a property vs. an
attribute, it can nevertheless sometimes be very helpful to see at a glance
what a given function is used for.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -931,9 +931,9 @@ static UInt SizeBlist(Obj blist)
 
 /****************************************************************************
 **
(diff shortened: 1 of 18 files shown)


fingolfin:
@sebasguts I updated the PR now with a minor tweak to ensure that `GVAR_PROP` can only be used in the right place (by adding a typecast to the intended target struct). So now we can be pretty sure I did it all right, otherwise the compiler would already complain :) (and yeah, I did verify that this works as intended). Let's see if the tests still pass, though.

PR #3229, opened 2019-01-24
Fix Unbound global variable warning

ChrisJefferson:
This fixes the fact that the ^^^^ markers on unbound globals would point to the wrong place.

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -37,8 +37,10 @@ static UInt NextSymbol(void);
 **  Helper function used by 'SyntaxError' and 'SyntaxWarning'.
 **
 */
-static void SyntaxErrorOrWarning(const Char * msg, UInt error)
+static void
+SyntaxErrorOrWarning(const Char * msg, UInt error, Int tokenoffset)
 {
+    GAP_ASSERT(tokenoffset >= 0 && tokenoffset <= 2);
     // do not print a message if we found one already on the current line
     if (STATE(NrErrLine) == 0) {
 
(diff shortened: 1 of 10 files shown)


src/scanner.c
```diff
@@ -920,8 +934,12 @@ static UInt NextSymbol(void)
...
     }
 
-  STATE(SymbolStartLine) = GetInputLineNumber();
-  STATE(SymbolStartPos) = GetInputLinePosition();
+    STATE(SymbolStartLine)[2] = STATE(SymbolStartLine)[1];
+    STATE(SymbolStartPos)[2] = STATE(SymbolStartPos)[1];
+    STATE(SymbolStartLine)[1] = STATE(SymbolStartLine)[0];
+    STATE(SymbolStartPos)[1] = STATE(SymbolStartPos)[0];
+    STATE(SymbolStartLine)[0] = GetInputLineNumber();
+    STATE(SymbolStartPos)[0] = GetInputLinePosition();
```
fingolfin: Perhaps this should be moved into its own helper function, say, `static void RecordInputPosition(void)` or so?

tst/testinstall/weakptr-badargs.tst
```diff
@@ -4,7 +4,7 @@ gap> START_TEST("weakptr-badargs.tst");
 gap> w := WeakPointerObject([1,,3,4]);
 Syntax warning: Unbound global variable in stream:1
 w := WeakPointerObject([1,,3,4]);
-                      ^
+     ^^^^^^^^^^^^^^^^^
 Error, Variable: 'WeakPointerObject' must have a value
```
fingolfin: I know this is not really part of this PR, but... Why is this test even here?

ChrisJefferson: Dunno :)

src/scanner.c
ChrisJefferson: Yes, once I had to add 2 levels, I should have abstracted out.

ChrisJefferson: I also added some more comments, as it's a bit subtle what's going on around here.

src/scanner.c
```diff
@@ -887,6 +897,26 @@ static void GetHelp(void)
     STATE(ValueObj) = AppendBufToString(string, buf, i);
 }
 
+
+/****************************************************************************
+**
+*F  StoreSymbolPosition()
+**
+**  Store the current position in the input stream, for use in error and
+**  warning messages. A sequnce of positions is stored, which record the start
```
fingolfin:
```suggestion
**  warning messages. A sequence of positions is stored, which record the start
```

tst/testinstall/weakptr-badargs.tst
fingolfin: Well, I thought you might now, as you added it in 35267ae :-)

ChrisJefferson: If I had to guess, I'd suspect it's just a one line too long cut+paste of a manually written test :)

PR #3230, opened 2019-01-24
Start using C++ templates for partial perms kernel code

fingolfin:
These changes were carefully prepared: usually, I took a group of 2 or 4 functions to be merged into a single function, put them into a separate document, then used regex search and replace to replace common patterns by their C++ counterpart (e.g  replace `2(f)` by `<TF>(f)` to convert things like `ADDR_PPERM2(f)` to `ADDR_PPERM<TF>(f)`). After a bunch of such scripted conversions, a few manual changes (typically for `NEW_PPERM` calls and other output related calls) were performed. Afterwards, ideally the 2 or 4 variants of the code already were identical; if not, then any minor cosmetic differences (spaces, extra parens/braces/etc.) were aligned, until the matched.

Still, mistakes can happen, which is why I usually convert only 1-2 functions per commit, to make it easier to bisect any regressions introduced by these changes.

### DIFF
--- a/src/pperm.cc
+++ b/src/pperm.cc
@@ -106,6 +106,23 @@ static inline UInt DEG_PPERM(Obj f)
 }
 
 
+//
+// The 'ResultType' template is used by functions which take two partial
+// permutations as argument to select the type of the output they produce: by
+// default, a T_PPERM4, whose entries are stored as UInt4. But if both inputs
+// are T_PPERM2, then as a special case the output is also a T_PPERM2, whose
+// entries are stored as UInt2.
+//
+template <typename TL, typename TR>
+struct ResultType {
+    typedef UInt4 type;
+};
+template <>
+struct ResultType<UInt2, UInt2> {
+    typedef UInt2 type;
+};
+
+
 #define MAX(a, b) (a < b ? b : a)
 #define MIN(a, b) (a < b ? a : b)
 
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Just noticed we aren't getting coverage of .cc files. I guess this is because we only put `--coverage` in CFLAGS at the top of .travis.yml?

ChrisJefferson:
While fixing this isn't required for this PR, it would be nice to check the coverage of the new code/functions

ChrisJefferson:
Also, I acknowledge there is some duplication between this and `permutat.cc`, including methods which do kind-of but not quite the same thing. However, I there was always duplication between permutat.c, pperm.c and trans.c, and I think nicer merging can be done later.

fingolfin:
Re C++ coverage: oops, that's indeed bad. It should be fixed, either in this PR or another. 

Re code duplication between this and `permutat.cc`: I already plan to reduce it, but I decided it is easier to focus on getting things done, then do this duplication reduction next. In particular because I may want to slightly tweak the templates in `permutat.cc`, based on what I did slightly differently this time around. My tentative plan is to add a new header `permutat_intern.hh` or so, which contains the shared permutation templates, and which then can be included from `pperm.cc` and `trans.cc`. Better ideas welcome!

james-d-mitchell:
Looks good to me!

fingolfin:
Coverage of course drops, as most of the removed code had excellent coverage

fingolfin:
@wilfwilson I wouldn't expect anybody to review this in depth, which is why I tried to explain that (and how) most of it was generated "safely" (?) in a mechanical fashion. Also test coverage of the rewritten code is excellent, which bolsters my confidence in the changes. And of course I'll work on fixing any regressions we might encounter down the road. As explained, I even split the changed to into multiple commits to help track down anybsuch regressions via bisecting.

wilfwilson:
@fingolfin Yes, I really appreciated your description at the top of the PR - it was useful. And splitting it into small commits was a good idea. I just wanted to record that I’ve had a bit of a look :) And everything I saw was positive

fingolfin:
So, if anybody is willing to approve this, we can merge it

PR #3231, opened 2019-01-24
Added more specialised function for testing conjugacy of subgroups of Sn

DominikBernhardt:
Directly using `SubgpConjSymmgp` as a specialized method  when testing two subgroups of a natural symmetric group for conjugacy in the full symmetric group often yields a speedup as for example in the last case in the test file.

### DIFF
--- a/tst/testinstall/ConjNatSym.tst
+++ b/tst/testinstall/ConjNatSym.tst
@@ -1,0 +1,14 @@
+# The following two test use the new version for natural symmetric group
+
+gap> IsConjugate( SymmetricGroup(5), Group((1,2)), Group((3,4)));
+true
+gap> IsConjugate( SymmetricGroup(5), Group((1,2)), Group((3,4,5)));
+false
+
+# This runs into the TryNextMethod case
+gap> IsConjugate( SymmetricGroup(200),PrimitiveGroup(200,4), PrimitiveGroup(200,3));
+false
+
+# Here, using SubgpConjSymmgp yields a significant speedup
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
Looks good. I'll just let all the tests cycle.

tst/testinstall/ConjNatSym.tst
```diff
@@ -0,0 +1,14 @@
+# The following two test use the new version for natural symmetric group
```
fingolfin: "the new version" of what/? Just looking at this file doesn't make that clear. How about something more like this:
> This file tests an`IsConjugate` method for natural symmetric groups, and its various cases.

tst/testinstall/ConjNatSym.tst
```diff
@@ -0,0 +1,14 @@
...
+gap> IsConjugate( SymmetricGroup(5), Group((1,2)), Group((3,4)));
+true
+gap> IsConjugate( SymmetricGroup(5), Group((1,2)), Group((3,4,5)));
+false
+
+# This runs into the TryNextMethod case
+gap> IsConjugate( SymmetricGroup(200),PrimitiveGroup(200,4), PrimitiveGroup(200,3));
+false
+
+# Here, using SubgpConjSymmgp yields a significant speedup
```
fingolfin: ... compared to what?

fingolfin:
Thanks, this is an improvement. However, the test file comments could have been better... :) When writing comments, never forget that a person reading a comment will usually *not* have the commit which added them as context, nor do they know anything about the motivation for writing the code in the first place.

PR #3234, opened 2019-01-25
kernel: rewrite SCR_SIFT_HELPER using C++ templates

fingolfin:
This also fixes a bug where it called DEG_PERM2 on a T_PERM4. I do not know if this caused user visible problems.

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -2571,150 +2571,101 @@ static Obj FuncMappingPermListList(Obj self, Obj src, Obj dst)
 /* end ); */
 
 
-static Obj FuncSCR_SIFT_HELPER(Obj self, Obj S, Obj g, Obj n)
+static UInt RN_stabilizer = 0;
+static UInt RN_orbit = 0;
+static UInt RN_transversal = 0;
+
+template <typename TG, typename Res>
+static Obj SCR_SIFT_HELPER(Obj stb, Obj g, UInt nn)
 {
-  Obj stb = S;
-  static UInt RN_stabilizer = 0;
-  static UInt RN_orbit = 0;
-  static UInt RN_transversal = 0;
-  UInt nn = INT_INTOBJ(n);
-  UInt useP2;
-  Obj result;
-  Obj t;
-  Obj trans;
-  int i;
+    int  i;
+    Obj  result = NEW_PERM<Res>(nn);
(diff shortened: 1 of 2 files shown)


src/permutat.cc
```diff
@@ -2649,150 +2649,109 @@ static Obj FuncMappingPermListList(Obj self, Obj src, Obj dst)
...
+
+    /* Copy g into the buffer */
+    Res * ptR = ADDR_PERM<Res>(result);
+    const TG * ptG = CONST_ADDR_PERM<TG>(g);
+    if (sizeof(TG) == sizeof(Res)) {
+        memcpy(ptR, ptG, sizeof(TG) * dg);
     }
     else {
-        const UInt4* ptrResult = CONST_ADDR_PERM4(result);
-        UInt degResult = DEG_PERM2(result);
```
fingolfin: @stevelinton this is the incorrect `DEG_PERM2` call.

Also, I wonder: isn't the value returned by DEG_PERM4 here equal to `nn` anyway?

src/permutat.cc
```diff
@@ -2649,150 +2649,109 @@ static Obj FuncMappingPermListList(Obj self, Obj src, Obj dst)
...
+}
+
+static Obj FuncSCR_SIFT_HELPER(Obj self, Obj stb, Obj g, Obj n)
+{
+    if (!IS_PREC(stb))
+        RequireArgument("SCR_SIFT_HELPER", stb, "must be a plain record");
+    RequirePermutation("SCR_SIFT_HELPER", g);
+    UInt nn = GetSmallInt("SCR_SIFT_HELPER", n);
+
+    if (!RN_stabilizer)
```
ChrisJefferson: Seems a bit weird to put these in here, instead of SCR_SIFT_HELPER, which actually needs/uses them?

fingolfin: This is shared code, so not putting it in there reduces the size of the generated code. Also, IMO it feels a bit less cluttered... In fact, strictly speaking, I'd prefer to init `RN_stabilizer` in `PostRestore`... I'll adjust the PR.

ChrisJefferson:
I already approved, but just for confirmation, I reapprove :)

fingolfin:
The test failure exposed by this PR is fixed by #3245

Issue #1256, opened 2017-04-17
CI coverage reports again experience random fluctuations


fingolfin:
Consider PR #1254, which touches no source files, yet still coverage changes. Namely in these files: (UPDATE: added more files, marked resolved files)

* [x] hpcgap/src/funcs.c
* [x] src/hpc/threadapi.c
* [x] hpcgap/src/vars.c
* [x] src/compiler.c
* [ ] hpcgap/lib/stdtasks.g
* [x] stc/stats.c

It would be good if we could avoid these fluctuations, as they lead to spurious coverage failures.

ChrisJefferson:
The `src/compiler.c` issues are to do with us testing compiling mode. There are several layers of problems here:

1) The `StoreString` method doesn't work properly when we use it to handle `-U` (compiler options) -- it puts a pointer into a `char [256]`, which we then try to parse as a string (that's where the fluctuations come from).

2) Even if we fix it, it turns out we ignore the argument `-U` anyway, as we manually set the flags in `InitKernel` anyway.

3) Relatedly, the `-i` option (set GAP root file) has been broken since at least 4.7.2 (last version I have that compiles), for the same reason.

I suggest, seeing as `-U` and `-i` haven't done anything useful for years and no-body cared, just removing them (for the compiler, in practice we know the best options to set, and can just continue hard-wiring them, as always)

ChrisJefferson:
This first issue (`-U`) is fixed in #1265. Once that gets merged, we can look to see if there are other issues as well.

fingolfin:
@ChrisJefferson Great find! I traced this in the logs of the old repository, and the current command line parsing code was introduced 2002-10-23. So, yeah, it should be safe to remove (although at least `-i` would also be trivial to fix -- but I see no point in that, and freeing up one-letter command line args to me is a good thing).

fingolfin:
And I just verified that at least `-i` was already broken in GAP 4.4.12:
```
~/Documents/gap4r4$ bin/gap.sh  -i lib/init.g
gap: hmm, I cannot find '�_�' maybe use option '-l <gaproot>'?
 If you ran the GAP binary directly, try running the 'gap.sh' or 'gap.bat' script instead.gap>
gap>
```

PR #3236, opened 2019-01-27
Add tests for slow IO

ChrisJefferson:
This adds some tests for when processes are slow at returning input to GAP.

As well as being (hopefully) generally useful, this is also one of the remaining sources of variation in coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
     - libreadline-dev
     - python
     - python-pip
+    - perl
 
 #
 # The following test jobs are roughly sorted by duration, from longest to
--- a/tst/teststandard/processes/streamio.tst
+++ b/tst/teststandard/processes/streamio.tst
@@ -1,0 +1,53 @@
+#@local scriptdir, write, checkPartialRead, process, c
+#
+gap> START_TEST("streamio.tst");
+gap> scriptdir := DirectoriesLibrary( "tst/teststandard/processes/" );;
+gap> write := Filename(scriptdir, "slowwrite.sh");;
+
+# Read the output of 'prog', using 'readfunc'. The output should
+# come in two parts, 'firstread' and 'secondread'.
+# If GAP is too slow, allow 'firstread' and 'secondread' to be read
+# in one step, but then return 'false'.
+# We run this function multiple times to avoid very occasional failures
+# due to OS scheduling.
+gap> checkPartialRead := function(prog, readfunc, firstread, secondread)
+> local process, l;
+> process := InputOutputLocalProcess(DirectoryCurrent(), prog, ["prog", firstread, secondread]);
+> l := ReadLine(process);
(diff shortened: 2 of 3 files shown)


fingolfin:
Those fluctuations in `lib/streams.gi` and `src/iostream.c` indeed are among the last, great. Perhaps we can close issue #1256 with this? Argh, but I just see that `hpcgap/lib/hpc/stdtasks.g` is still (again) acting up, too...

tst/teststandard/processes/streamio.tst
```diff
@@ -0,0 +1,53 @@
...
+> local process, l;
+> process := InputOutputLocalProcess(DirectoryCurrent(), prog, ["prog", firstread, secondread]);
+> l := ReadLine(process);
+> if l = Concatenation(firstread, secondread) then
+>   if ReadLine(process) <> fail then Error("Invalid end - type 1"); fi;
+>   CloseStream(process);
+>   return false;
+> else
+>   if ReadLine(process) <> secondread then Error("Missing second"); fi;
+>   if ReadLine(process) <> fail then Error ("Invalid end - type 2"); fi;
```
fingolfin:
```suggestion
>   if ReadLine(process) <> fail then Error("Invalid end - type 2"); fi;
```

tst/teststandard/processes/slowwrite.pl
```diff
@@ -0,0 +1,11 @@
...
+use strict;
+use warnings;
+
+use Time::HiRes;
+
+print $ARGV[1];
+select()->flush();
+Time::HiRes::sleep(1);
+
+print $ARGV[2];
```
fingolfin: I am not so happy in now also adding perl scripts. Is there a strong reason for this? Can't the above also be done with an equally short bash/GAP/python script? All it seems to do is to print from argv, then flush and sleep. Shouldn't e.g. this do the job as well?
```sh
#!/bin/sh
echo $1
sleep 1
echo $2
```

ChrisJefferson: I mainly used perl, because there was already a perl script in the same directory

ChrisJefferson: That other script is perl because I needed to be able to ignore sigint and sigterm, and wasn't sure how to do that in any other language (well, other than C of course).

ChrisJefferson: However (should have done this before I started), now changed to bash. Ended up being a little trickier, as I need to control carefully where newlines appear, so I used printf instead of echo.

.travis.yml
```diff
@@ -17,6 +17,7 @@ addons:
     - libreadline-dev
     - python
     - python-pip
+    - perl
```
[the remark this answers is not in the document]
ChrisJefferson: Didn't think it hurt to be clear about things, in case the defaults ever change in future. I'm happy to remove it.

PR #3237, opened 2019-01-27
Make compiler aware of hvars, and the lvar bag cache

ChrisJefferson:
GAP has the concept of lvars bags, which can become hvar bags if references to them are captured. This allows caching and reusing used lvars bags.

This is entirely ignored by the compiler -- it neither marks bags as hvar bags, or allows bags to be reused. I think these two things together cancel out, meaning this can't lead to any bag corruption, but I'm not 100% positive. I'd certainly be happier to fix it.

This PR makes the lvar bag cache much more useful -- without this PR it has about a 40% hit rate during startup, with this PR this goes up to about 95%.

### DIFF
--- a/src/compiler.c
+++ b/src/compiler.c
@@ -5137,6 +5137,9 @@ static void CompFunc(Obj func)
     /* emit the code to switch to a new frame for outer functions          */
     Emit( "\n/* allocate new stack frame */\n" );
     Emit( "SWITCH_TO_NEW_FRAME(self,%d,0,oldFrame);\n",NHVAR_INFO(info));
+    if (NHVAR_INFO(info) > 0) {
+        Emit("MakeHighVars(STATE(CurrLVars));\n");
+    }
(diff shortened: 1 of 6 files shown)


src/compiler.c
```diff
@@ -5137,6 +5137,9 @@ static void CompFunc(Obj func)
     /* emit the code to switch to a new frame for outer functions          */
     Emit( "\n/* allocate new stack frame */\n" );
     Emit( "SWITCH_TO_NEW_FRAME(self,%d,0,oldFrame);\n",NHVAR_INFO(info));
+    if(NHVAR_INFO(info) > 0) {
```
fingolfin:
```suggestion
    if (NHVAR_INFO(info) > 0) {
```

ChrisJefferson: Thought I'd formatted, now clang-formatted throughout (which makes a couple of other tiny changes)

PR #3238, opened 2019-01-28
Fix usage of InfoObsolete

ChrisJefferson:
InfoObsolete isn't a function. Fixes resclasses (and maybe others).


fingolfin:
Luckily this does not affect 4.10 (hence also no need for backporting or release notes)

PR #3239, opened 2019-01-28
kernel: better error checks in vec8bit.c, vecgf2.c

fingolfin:
Add GetNonnegativeSmallInt for this.

Note that the input validation check in `FuncADD_GF2VEC_GF2VEC_SHIFTED` was bogus and accepted too many values.

### DIFF
--- a/src/error.h
+++ b/src/error.h
@@ -227,6 +227,7 @@ GetSmallIntEx(const char * funcname, Obj op, const char * argname)
 #define GetSmallInt(funcname, op)                                            \
     GetSmallIntEx(funcname, op, NICE_ARGNAME(op))
 
+
 /****************************************************************************
 **
 *F  GetPositiveSmallIntEx, GetPositiveSmallInt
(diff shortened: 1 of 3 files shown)


src/vecgf2.c
```diff
@@ -3621,23 +3594,13 @@ static Obj
...
-                     "integer not a %s",
-                     (Int)TNAM_OBJ(off), 0L);
-    off1 = INT_INTOBJ(off);
-    if (off1 < 0) {
-        ErrorMayQuit("ADD_GF2VEC_GF2VEC_SHIFTED: <offset> must be a "
-                     "non-negative integer",
-                     0, 0);
-    }
-    len2a = INT_INTOBJ(len2);
-    while (len2a < 0 && len2a <= LEN_GF2VEC(vec2)) {
```
fingolfin: This check was wrong, len2a must be "less than the actual length of the vector"

src/error.h
```diff
@@ -227,6 +227,7 @@ GetSmallIntEx(const char * funcname, Obj op, const char * argname)
 #define GetSmallInt(funcname, op)                                            \
     GetSmallIntEx(funcname, op, NICE_ARGNAME(op))
 
+
```
ChrisJefferson: ? (but not a serious problem)

PR #3243, opened 2019-01-28
Specify a julia binary instead of a julia prefix

sebasguts:
and let julia collect the information about all other paths


sebasguts:
@fingolfin I thought about making this backwards-compatible, but I decided against because I wanted to set a default.

`--with-gc=julia` kind of implies `--with-julia=` to be set, because the configure will fail. We could have a better error message, though. Or did you mean `--with-julia` should imply `--with-gc=julia`?

fingolfin:
Backported to stable-4.10 in 3c74918908000ba06eb9c828930f96a74ae5a447

configure.ac
```diff
@@ -457,13 +457,35 @@ AC_SUBST([READLINE_LIBS])
 
 dnl Find julia
 AC_ARG_WITH([julia],
-  [AS_HELP_STRING([--with-julia@<:@=PREFIX@:>@],
-    [specify a prefix where julia can be found])],
-  [],[with_julia=no]
-)
+  [AS_HELP_STRING([--with-julia@<:@=PATH@:>@],
+    [specify a prefix for a julia binary or a julia binary])],
```
alex-konovalov: @sebasguts do you really mean `[specify a prefix for a julia binary or a julia binary]`?

sebasguts: Yes, I mean it that way.

You can either directly specify a julia binary, or its prefix, e.g., `/usr/local`

alex-konovalov: Ah, now I got it. Parsing plain text without parentheses or commas failed. Maybe 

    specify a prefix for a julia binary, or a julia binary

or
 
    specify a julia binary or a prefix for a julia binary 

would be easier to read.

sebasguts: I think I understand the problem now, sorry. I will provide a PR for that.

alex-konovalov:
See also #3316 for an updated comment.

PR #3245, opened 2019-01-28
kernel/gac: fix CompUnbComObjExpr

fingolfin:
CompUnbComObjExpr was calling CompSetUseRNam but should not; this lead to the value of random RNams being emitted to the generated C code.

This lead to a test failure in #3234.


ChrisJefferson:
So is this just leading to random variables that don't get used? (just trying to understand the bug)

fingolfin:
Basically yes. It takes the result of `CompExpr` and erroneously treats it as an rnam, which then is passed to `CompSetUseRNam`, where it is marked as "in use"; and we then generate a variable in the C code

PR #3248, opened 2019-01-29
Export Julia CFLAGS, LDFLAGS, and LIBS to sysinfo.gap

sebasguts:


fingolfin:
Backported to stable-4.10 in c67b4f0a999f9098525d6114e6d94c849d540021

PR #3257, opened 2019-02-01
kernel: pass func obj explicitly to PrintStat, PrintExpr

fingolfin:
... instead of implicitly assuming that we are printing the "current" function. This is a small step towards decoupling the code for generating, executing, and printing function bodies. The long term goal is that coding passes the T_BODY object around explicitly, and `STATE(PtrBody)` is used only by the executor. It thus becomes read-only and can be made `const`.

This change also means we can print function bodies w/o first switching to the function.

### DIFF
--- a/src/compiler.c
+++ b/src/compiler.c
@@ -975,14 +975,18 @@ static CVar CompFunccall0to6Args(Expr expr)
     CVar                args [8];       /* arguments                       */
     Int                 narg;           /* number of arguments             */
     Int                 i;              /* loop variable                   */
+    Expr                funcExpr;
+
+    funcExpr = READ_EXPR(expr, 0);
+    narg = SIZE_STAT(expr) / sizeof(Expr) - 1;
 
     /* special case to inline 'Length'                                     */
     if ( CompFastListFuncs
-      && TNUM_EXPR( FUNC_CALL(expr) ) == T_REF_GVAR
-      && READ_EXPR( FUNC_CALL(expr), 0 ) == G_Length
-      && NARG_SIZE_CALL(SIZE_EXPR(expr)) == 1 ) {
+      && TNUM_EXPR(funcExpr) == T_REF_GVAR
+      && READ_EXPR(funcExpr, 0) == G_Length
+      && narg == 1 ) {
         result = CVAR_TEMP( NewTemp( "result" ) );
-        args[1] = CompExpr( ARGI_CALL(expr,1) );
+        args[1] = CompExpr(READ_EXPR(expr, 1));
         if ( CompFastPlainLists ) {
             Emit( "C_LEN_LIST_FPL( %c, %c )\n", result, args[1] );
         }
@@ -998,18 +1002,17 @@ static CVar CompFunccall0to6Args(Expr expr)
     result = CVAR_TEMP( NewTemp( "result" ) );
 
     /* compile the reference to the function                               */
-    if ( TNUM_EXPR( FUNC_CALL(expr) ) == T_REF_GVAR ) {
-        func = CompRefGVarFopy( FUNC_CALL(expr) );
+    if (TNUM_EXPR(funcExpr) == T_REF_GVAR) {
+        func = CompRefGVarFopy(funcExpr);
     }
     else {
-        func = CompExpr( FUNC_CALL(expr) );
+        func = CompExpr(funcExpr);
         CompCheckFunc( func );
     }
 
     /* compile the argument expressions                                    */
-    narg = NARG_SIZE_CALL(SIZE_EXPR(expr));
     for ( i = 1; i <= narg; i++ ) {
-        args[i] = CompExpr( ARGI_CALL(expr,i) );
+        args[i] = CompExpr(READ_EXPR(expr, i));
     }
 
     /* emit the code for the procedure call                                */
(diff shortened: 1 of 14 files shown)


ChrisJefferson:
So, my main worry about this PR is that it looks incomplete (for example, there are some still READ_EXPR not turned into READ_EXPR_IN_BODY). Therefore, I assume, at the moment it's important users keep the 'body' argument in still in sync with the global 'STATE(PtrBody)'?

If that's the case, adding some kind of comment somewhere (near READ_EXPR/READ_EXPR_IN_BODY perhaps) describing the current state might be a good idea, just in case Google offers you a giant pile of money and someone is trying to figure out what's going on in 10 years or so :)

fingolfin:
I just removed all but the first commit -- the two removed commits indeed appeared to be incomplete (my mistake: I revived this PR last night, after not having looked at it for almost two months, and it was completely done).

So now it is back to only modifying printing (as in the original version of this PR), and there, no READ_STAT or READ_EXPR calls should be left. I just double checked that, but of course I am fallible -- so, if you can spot any, please let me know: that's a bug then and should indeed be fixed. 

Once this is in, I can continue work on the update for the compiler; and independently from that, the syntax tree compiler code can be changed, too (CC @sebasguts).

src/compiler.c
```diff
@@ -3735,7 +3739,7 @@ static void CompIf(Stat stat)
     /* print a comment                                                     */
     if ( CompPass == 2 ) {
         Emit( "\n/* if " );
-        PrintExpr(READ_EXPR(stat, 0));
```
ChrisJefferson: Why is this line (for example) READ_EXPR, when other lines are READ_EXPR_IN_BODY?

fingolfin: I was confused by that comment as that line actually is removed :-). But I assume you meant the next line, so I'll comment there.

src/compiler.c
```diff
@@ -3735,7 +3739,7 @@ static void CompIf(Stat stat)
...
+        PrintExpr(BODY_FUNC(CURR_FUNC()), READ_EXPR(stat, 0));
```
fingolfin: I think this is fine because `READ_EXPR` reads from `BODY_FUNC(CURR_FUNC())`, and it is very explicit here that this is precisely the body we pass to `PrintExpr`... inside `PrintExpr`, though, no use is made of `BODY_FUNC(CURR_FUNC())` or `STATE(PtrBody)` or anything.

Of course next the compiler code should be made to not use these globals -- that was partially done in the commits I removed from this PR, but it wasn't quite complete.

stevelinton:
This looks fine to me, but a test is failing.

fingolfin:
The error is caused by the `PrintFunccall1` which uses `FUNC_CALL`, `NARG_SIZE_CALL`, `ARGI_CALL`. I am tempted to say that these macros must go and be replaced with a better mechanism. Similar problem with `PrintInfo` and `NARG_SIZE_INFO`, `ARGI_INFO` I guess. 

Anyway, that's something for next week or so.

src/vars.c
```diff
@@ -367,19 +367,19 @@ static Obj EvalIsbHVar(Expr expr)
 **
 **  'PrintAssHVar' prints the higher variable assignment statement <stat>.
 */
-static void PrintAssHVar(Stat stat)
+static void PrintAssHVar(Obj body, Stat stat)
 {
     Pr( "%2>", 0L, 0L );
-    Pr("%H", (Int)NAME_HVAR(READ_STAT(stat, 0)), 0L);
+    Pr("%H", (Int)NAME_HVAR(READ_STAT_IN_BODY(body, stat, 0)), 0L);
```
fingolfin: Oops, I just realized this call to `NAME_HVAR` here and in a few other places is a serious problem: it requires the environment of the function object; and that is *not* available from the body alone.... *sigh*.

So it seems we either need to pass in the `T_FUNCTION` bag, not just the `T_BODY` bag; (which is a bit sad, because we only need the body in all other print functions, as far as I can tell). Or, we need to redesign how hvar references work: instead of walking about the lvars chain, we could also store any relevant `T_LVARS` resp. `T_HVARS` inside the `values` plist of the body; then adjust the code writing to HVARS to go through that. I am not quite sure whether that's a terrific or terrible idea, though...

stevelinton: The lexical environment of a function body (ie the enclosing function bodies) should logically be an attribute of the body itself, not the function and is enough to find names, though of course not values, of higher variables (debug variables are whole  other problem). 

If we adjusted our data structures to reflect this, then a function would lead to two chains of pointers -- one from functions to the local variables bags of the function calls in which they were created, and one from bodies to the lexically enclosing bodies. This is redundant, with all the usual risks of inconsistency, but without it, a function body with higher variable references is meaningless when taken out of context.

Another option is just to cache the names of the higher variables in the function body as a convenience.

src/vars.c
```diff
@@ -196,16 +196,16 @@ static Obj EvalIsbLVar(Expr expr)
 **
 **  'PrintAssLVar' prints the local variable assignment statement <stat>.
 */
-static void PrintAssLVar(Stat stat)
+static void PrintAssLVar(Obj body, Stat stat)
 {
     Pr( "%2>", 0L, 0L );
     Pr("%H", (Int)NAME_LVAR(READ_STAT(stat, 0)), 0L);
```
fingolfin: Actually, of course `NAME_LVAR` is *also* broken, and really requires a pointer to the function object, as only that contains the names of the lvars, in form of the `Obj nams` plist. Well, or the body would have to contain a pointer to the `nams` plist... but I am worried a bit about potential synchronization issues. What if a `T_BODY` is shared by multiple functions (I think this happens with higher order functions: we create a single `T_BODY`, stored in the function expression object; and then each instantiation of the `fexs` into an actual function uses the same existing `T_BODY`.

And in the future, I was envisioning a version of `PrintFunction` which also prints out the values of all `HVARs` used in the function (that would be super handy during debugging). Though I guess that's OK for `PrintFunction`, as it actually can access the function object...

Hrrm to all.

(Also, clearly I overlooked the `READ_STAT` here, and in a few other places... bad)

stevelinton: The nams list logically belongs with the body. We have to distinguish between 

- a body (which is essentially  a piece of code with its lexical context (ie the function body that textually surrounds it); 
- a function (or a closure in many languages) which is a GAP object with a local and possibly some higher variables bags associated with it; and
- a stack frame, which is an actual function invocation with values for the parameters, a currently executing statement calling routine and so on.  

the relationships are one-to-many down the page and any piece of information should belong at one of the three levels, but I don't think GAP is entirely clean about this splitting.  It would be good to make all of this tidier.

fingolfin: I am not so sure how that could / should be made tidier, though (not in the sense that I dispute the need or the possibility, just that it is not yet clear to me how to get there).

For example, we also need to take kernel function and gac compiled functions (which also may reference HVARs) into account. The former have no T_BODY at all. If we did the minimal thing, and just moved the list of argument+local names (currently named `namesOfArgsAndLocals`, but I am considering to rename it back to `nams`...) into the T_BODY, then all kernel functions would need to get a T_BODY, just to store the argument names. But OK, we can do that.

Next is the number of arguments `nargs` (BTW, I am thinking about renaming that back to `narg`). Should that also be in the T_BODY? I guess not, as there is no need for it there, and we don't distinguish between arguments and locals on the level of the body; at most it is interesting to know which local "slots" are expected to be pre-filled... The same probably holds for `nloc`, the number of locals (indeed, it is not even clear to me whether we even need that -- we should always have `nloc + narg = LEN_LIST(nams)`, no?

For `prof` (profiling information), `envi` (the environment which started this discussion), `locks` (for HPC-GAP), arguably I'd think they belong with the function, not the body... Hm

PR #3258, opened 2019-02-02
Replace FindPRec, the function for looking up members of a precord.

ChrisJefferson:
Short version: There was a function call `FindPRec`. I replaced it with a slightly faster, slightly cleaner method called `PositionPRec`. I left a implementation of `FindPRec` using `PositionPRec` because it is used by the semigroups package, hopefully this can be eventually removed.

Longer:

While investigating another bug, I thought the problem was in FindPRec, and I did find a bug in FindPRec (when it returns 0, it sometimes fails to set the pointer `pos`).

However, then it turned out if FindPRec returns 0, no-one uses pos anyway.

I decided to rewrite the function to be (in my opinion) easier to read. I also carefully benchmarked a number of micro-optimisations using valgrind, and believe I have found a local optima. This produces a small but measurable improvement on record-heavy benchmarks.

I would like to do some other cleanups (The return value of GET_RNAM_PREC is really a signed integer, but we treat it as an unsigned integer. In general I'd like to be more careful about this signed/unsigned business). However, I that PR will be much larger and much simpler, so I thought I'd do this first.

### DIFF
--- a/src/hpc/aobjects.c
+++ b/src/hpc/aobjects.c
@@ -603,8 +603,7 @@ static void PrintTLRecord(Obj obj)
   for (i = 0; i < deftable[AR_CAP].atom; i++) {
     UInt key = deftable[AR_DATA+2*i].atom;
     Obj value = deftable[AR_DATA+2*i+1].obj;
-    UInt dummy;
-    if (key && (!record || !FindPRec(record, key, &dummy, 0))) {
+    if (key && (!record || !PositionPRec(record, key, 0))) {
       if (comma)
         Pr("%2<, %2>", 0L, 0L);
       Pr("%H", (Int)(NAME_RNAM(key)), 0L);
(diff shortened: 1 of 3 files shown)


src/hpc/aobjects.c
```diff
@@ -994,7 +993,7 @@ Obj GetTLRecordField(Obj record, UInt rnam)
   contents = GetTLInner(record);
   table = ADDR_OBJ(contents);
   tlrecord = table[TLR_DATA+TLS(threadID)];
-  if (!tlrecord || !FindPRec(tlrecord, rnam, &pos, 1)) {
+  if (!tlrecord || !(pos = PositionPRec(tlrecord, rnam, 1))) {
```
fingolfin: I dislike assignments in conditions. Could this perhaps be rewritten to something like this?
```C
  pos = tlrecord ? PositionPRec(tlrecord, rnam, 1) : 0;
  if (pos) {
```

src/precord.c
```diff
@@ -264,12 +259,12 @@ UInt FindPRec( Obj rec, UInt rnam, UInt *pos, int cleanup )
...
              * fall back to binary search: */
-            i = high;
+            UInt i = high;
             while (i >= 1) {
-                rnam2 = (Int) (GET_RNAM_PREC(rec,i));
+                Int rnam2 = (Int)(GET_RNAM_PREC(rec, i));
                 if (rnam == rnam2) {
-                    *pos = i;
-                    return 1;
+                    GAP_ASSERT(i != 0);
```
fingolfin: Why this assertion? The loop condition proves that `i >= 1` holds, hence `i != 0` always must be satisfied.

ChrisJefferson: It was more just a silly brute-force, I put those in at all return values which have to return a non-zero value, as the rnam was found.

src/hpc/aobjects.c
ChrisJefferson: I agree. I don't like it either -- but I also didn't like turning it into a double level if. I'll try something like that.

PR #3261, opened 2019-02-04
Run on Travis job with GCC 4.7, to verify we stay compatible with it

### LINKED ISSUE #3141: Compiler warning in the master branch
New compiler warning in the master branch:
```
libtool: compile:  gcc -std=gnu99 -MQ obj/src/c_oper1.lo -MMD -MP -MF obj/src/.deps/c_oper1.d -m64 -Werror -fno-strict-aliasing -Wall -Wextra -Wundef -Wwrite-strings -Wpointer-arith -Wredundant-decls -Wno-unused-parameter -Wmissing-field-initializers -Wformat=2 -Wformat-nonliteral -Wformat-security -Wno-sign-compare -Wstrict-aliasing -Wshadow -Winline -Wpacked -Wmissing-format-attribute -Winit-self -Wmissing-include-dirs -Wunused-but-set-variable -Warray-bounds -Wreturn-type -Wno-inline -Wlogical-op -Wdouble-promotion -Wno-cast-function-type -Wnested-externs -Wno-implicit-fallthrough -Wimplicit-function-declaration -Wold-style-definition -Wjump-misses-init -Wno-error=unused-parameter -Wno-error=sign-compare -Wno-error=inline -I./gen -I./src -DHAVE_CONFIG_H -c src/c_oper1.c  -fPIC -DPIC -o obj/src/.libs/c_oper1.o
In file included from src/system.h:32:0,
                 from src/gasman.h:39,
                 from src/objects.h:20,
                 from src/ariths.h:16,
                 from src/compiled.h:21,
                 from src/c_oper1.c:3:
src/finfield.h: In function ���POW_FFV���:
src/debug.h:43:33: error: typedef ���static_assertion_291��� locally defined but not used [-Werror=unused-local-typedefs]
     typedef char _intern_CONCAT(static_assertion_, __LINE__)[(cond)? 1 : -1]
                                 ^
src/debug.h:40:32: note: in definition of macro ���_intern_CONCAT_���
 #define _intern_CONCAT_(X, Y)  X ## Y
                                ^
src/debug.h:43:18: note: in expansion of macro ���_intern_CONCAT���
     typedef char _intern_CONCAT(static_assertion_, __LINE__)[(cond)? 1 : -1]
                  ^
src/finfield.h:290:5: note: in expansion of macro ���GAP_STATIC_ASSERT���
     GAP_STATIC_ASSERT(sizeof(UInt) >= 2 * sizeof(FFV),
     ^
src/c_oper1.c: At top level:
cc1: error: unrecognized command line option "-Wno-implicit-fallthrough" [-Werror]
cc1: error: unrecognized command line option "-Wno-cast-function-type" [-Werror]
cc1: all warnings being treated as errors
make: *** [obj/src/c_oper1.lo] Error 1
Build step 'Execute shell' marked build as failure
```

fingolfin:
Note that 4.7 is the oldest GCC version readily available on Travis.

As a side effect, I was able to reproduce #3141, and added a workaround for it to this PR, too. Thus this PR also fixes #3141.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,10 +117,15 @@ matrix:
     # The '--enable-valgrind' checks that GAP builds and runs correctly when
     # compiled with valgrind support. We do not actually run any tests using
     # valgrind, as it is too slow.
-    - env: TEST_SUITES="docomp testinstall" NO_COVERAGE=1 ABI=64 BUILDDIR=build CONFIGFLAGS="--enable-valgrind"
+    #
+    # Also change the compiler to GCC 4.7, to ensure we stay compatible
+    # with that older version.
+    - env: TEST_SUITES="docomp testinstall" NO_COVERAGE=1 ABI=64 BUILDDIR=build CONFIGFLAGS="--enable-valgrind CC=gcc-4.7 CXX=g++-4.7"
       addons:
         apt_packages:
-            valgrind
+          - gcc-4.7
+          - g++-4.7
+          - valgrind
 
     # same as above, but in 32 bit mode, also turn off debugging (see elsewhere in this file for
     # an explanation)
--- a/src/finfield.h
+++ b/src/finfield.h
@@ -143,6 +143,8 @@ extern  Obj             SuccFF;
 */
 typedef UInt2           FFV;
 
+GAP_STATIC_ASSERT(sizeof(UInt) >= 2 * sizeof(FFV),
+                  "Overflow possibility in POW_FFV");
 
 /****************************************************************************
 **
@@ -287,8 +289,6 @@ EXPORT_INLINE FFV POW_FFV(FFV a, UInt n, const FFV * f)
 {
     GAP_ASSERT(a <= f[0]);
     GAP_ASSERT(n <= f[0]);
-    GAP_STATIC_ASSERT(sizeof(UInt) >= 2 * sizeof(FFV),
-                      "Overflow possibility in POW_FFV");
     if (!n)
         return 1;
     if (!a)
--- a/src/read.c
+++ b/src/read.c
@@ -259,20 +259,22 @@ enum REFTYPE {
 };
 
 typedef struct {
-    enum REFTYPE type;
+    UInt1 type;
 
     union {
-        UInt var;
-        UInt narg;
-        UInt rnam;
+        UInt2 nest0;
+        UInt2 level;
     };
 
     union {
-        UInt nest0;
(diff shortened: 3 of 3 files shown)


fingolfin:
First off, I don't see how supporting GCC 4.9 is any easier/better than GCC 4.7 -- any particular reason you picked that version?

Next, we have users (e.g. @hulpke) who for various reasons out of their direct control may need to use GAP on machines with even older compilers. So if we drop support for an older GCC version resp. increase our minimal required version, at least we should have a good reason to do so. Right now, we don't have any, as GAP compiles perfectly fine on these older compilers -- with the exception of some harmless warnings which this PR fixes, for the sole reason that we run our CI tests with `-Werror`.

So, yes, I do think we should enforce compatibility with GCC 4.7 for the time being (and ideally, also older versions; as explained above, I only chose 4.7 as that's the oldest that is easily available on Travis, AFAIK). I'd only revise this if somebody has a compelling reason why we need to drop it, and right now, I don't see one.  GCC 4.7 even supports most (?) of C11 and C++11, and we already exploit that by using static assertions if available. I'd not want to use any newer C features unless we have *super good* reasons, and I am reluctant about C++11 (and against C++14) use, as we agreed to keep our usage of C++ super simple to understand for C coders who are new to C++, and I don't see how anything newer C++ standards bring fits with that.

sebasguts:
if there are reasons to keep this compatibility I am not against it.

The main reason for `4.9` is the support for `decltype(auto)` and pure generics in C++, i.e., functions like
```
auto foo(auto X){ return X.bar(); }
```
I realize this is not a necessary feature, but is nice to use. But I guess as long as nobody actually wants to use it, it is fine.

fingolfin:
I am actually absolutely against using `auto` in our C++ code for the time being, or `decltype`. While convenient, they can be deeply magical, and would remove our C++ code even more from C than it already is.

ChrisJefferson:
I think supporting 4.7 as long as is easy, we should do. As we our standard method of distributing Gap is by source, we need to make it easy to compile.

fingolfin:
Unfortunately, the Julia test keeps failing. I wonder if the added `volatile` could somehow be responsible?! We are passing an `LHSRef` struct there as an argument, which takes up several registers and/or needs spilling on the stack.... but then again, this does not involve any object references, so I don't see how this could cause a GC crash... Perhaps @rbehrends has an idea?

PR #3263, opened 2019-02-04
kernel: refactor bitfields code to avoid (ab)using FEXS_FUNC and NLOC_FUNC

fingolfin:
We achieve this by instead allocating a slightly larger bag for the function, and then store the two extra values in the additional space. Note that we must store them as `Obj` and not as plain `UInt`, because the GC marking function for function bags expects that.

However, distinguish between operations and "normal" functions by the bag size. This still works if we compared the bag size against `sizeof(OperBag)`, as done by `IS_OPERATION`, but not if we compared it against `sizeof(FuncBag)` as done in `SaveFunction` and `LoadFunction`. So we change the latter to use `IS_OPERATION` to resolve that problem. (I plan to provide a PR in the future which replaces this size based detection by something more robust.)

Motivation for this, besides cleanliness / avoiding hacks that make the code harder to understand, is a desire to get rid of FEXS (PR for that upcoming).

Note that JuliaInterface and PARIInterface currently also abuse FEXS_FUNC in a similar fashion (CC @sebasguts and @markuspf ). But we can fix them in a similar way. If nobody beats me to it, I'll try to provide PRs for them doing this.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -1577,7 +1577,7 @@ static ObjFunc LoadHandler( void )
 static void SaveFunction(Obj func)
 {
   const FuncBag * header = CONST_FUNC(func);
-  for (UInt i = 0; i <= 7; i++)
+  for (UInt i = 0; i < ARRAY_SIZE(header->handlers); i++)
     SaveHandler(header->handlers[i]);
   SaveSubObj(header->name);
   SaveSubObj(header->nargs);
@@ -1587,7 +1587,7 @@ static void SaveFunction(Obj func)
   SaveSubObj(header->body);
   SaveSubObj(header->envi);
   SaveSubObj(header->fexs);
-  if (SIZE_OBJ(func) != sizeof(FuncBag))
+  if (IS_OPERATION(func))
     SaveOperationExtras( func );
 }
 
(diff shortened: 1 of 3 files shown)


sebasguts:
So, as far as I understood is that the two additional values are now marked by the GC.

In JuliaInterface we rely on the the fact that `FEXS` is not marked. How does that fit together?

fingolfin:
Simple, your second claim is wrong :-). The converse is true: we in fact rely on fexs being marked.

sebasguts:
Ah, I actually remembered that wrong. For JuliaInterface, we do not want the `FEXS` being marked (as we store a raw C function pointer in them. We just allow for that pointer being marked by encapsulating it in a Julia `Ptr{Cvoid}` object.

sebasguts:
So, follow-up questions: Could we have a non-marked member? Or should we continue wrapping and unwrapping the pointer? That is not a huge problem btw, as the operation takes almost no time.

fingolfin:
@sebasguts allowing for non-marked members would require a much more complex marking functions, and a way to specify which parts of the bag to mark and which not. E.g. there could be a way to specify "don't mark the last N bytes of the bag" or so. (BTW, I plan to add another non-marked 8 byte "field" at the *start* of the bag, in which we could store `nloc` and `nargs` compactly (both can be stored in 2 bytes), plus a "subtype" (to allow distinguishing "coded" functions, regular kernel functions, operations, and "custom" functions, as used by JuliaInterface and PARIInterface; and finally a "bitflags" field for various purposes (for starters, it could be used to store the bits currently stored in the `extra` field of an `OperBag`).

For now, I don't think this complexity is worth it (and it certainly is nothing I'd implement for this PR). But we can certainly look at it again in the future.

PR #3265, opened 2019-02-04
kernel: get rid of FEXS_FUNC

fingolfin:
This contains PRs #3262 and #3263, and will need to be rebased once those are merged. Only the last commit is "new" compared to those two PRs.

This breaks JuliaInterface and PARIInterface, but those can be fixed as described in PR #3263; in fact, we may want to first merge PR #3263, then fix those two packages, before merging this PR here.

### DIFF
--- a/src/calls.c
+++ b/src/calls.c
@@ -1581,7 +1581,6 @@ static void SaveFunction(Obj func)
   SaveSubObj(header->nloc);
   SaveSubObj(header->body);
   SaveSubObj(header->envi);
-  SaveSubObj(header->fexs);
   if (IS_OPERATION(func))
     SaveOperationExtras( func );
 }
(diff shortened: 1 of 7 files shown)


stevelinton:
This code looks fine, but is there an issue or PR somewhere that explains why you want to get rid of FEXS_FUNC?

fingolfin:
@stevelinton to quote the commit message:
> This is rather natural, as the function expressions logically belong to the body, not the function object. It also fits in perfectly with the intended usage of the `values` list in each BodyBag.

fingolfin:
@stevelinton I've now rephrased the commit message as follows, perhaps that makes it clearer.

> kernel: remove FEXS_FUNC
>
> Function expressions logically belong to the function body, not the
> function object. Hence FEXS_FUNC should be in the body, not the
> function. But in the body, we already have the `values` plist, which is
> used to so to store all other kinds of expression objects. Hence we can
> simply get rid of FEXS_FUNC, and store function expressions in the
> `values` plist.

fingolfin:
@stevelinton If you think that this is OK now, could  you please approve, so it can be merged? Thanks!

fingolfin:
Note: We might still want to hold off with actually merging it, though, to first get PARIInterface and JuliaInterface changed to not use FEXS. I'll prepare PRs and/or issues for them now.

PR #3267, opened 2019-02-05
Change StructureDescription for infinite cyclic group

fingolfin:
It used to be C0, now we refer to it as Z (as in: the integers), which
should be understood by more people without consulting the manual. Also,
the short name "0" is likewise replace by 0.

This is also consistent with what the RCWA package does in its custom
StructureDescription method.


Stefan-Kohl:
I think this change of notation for the infinite cyclic group looks o.k..

PR #3270, opened 2019-02-06
Further minor precord cleanups

ChrisJefferson:
This is just some minor cleanups to precords, as I was exploring around the code.

I thought there was a bug relating to inplace/non-inplace sorting, but I've decided there isn't a problem. There was (I believe) originally a plan to use inplace sorting, but it was never used due to a typo. It was then accidentally used in HPC-GAP. I've removed it altogther.

The only reason not to remove the extra argument is that the function is called from the 'datastructures' package. At some point we could do something to remove the extra argument.

### DIFF
--- a/src/hpc/aobjects.c
+++ b/src/hpc/aobjects.c
@@ -586,7 +586,7 @@ static void PrintTLRecord(Obj obj)
   if (record) {
     for (i = 1; i <= LEN_PREC(record); i++) {
       Obj val = GET_ELM_PREC(record, i);
-      Pr("%H", (Int)NAME_RNAM(labs((Int)GET_RNAM_PREC(record, i))), 0L);
+      Pr("%H", (Int)NAME_RNAM(labs(GET_RNAM_PREC(record, i))), 0L);
       Pr ("%< := %>", 0L, 0L);
       if (val)
         PrintObj(val);
--- a/src/hpc/serialize.c
+++ b/src/hpc/serialize.c
@@ -626,7 +626,7 @@ static Obj DeserializeRecord(UInt tnum)
         Obj el = DeserializeObj();
         SET_ELM_PREC(result, i, el);
     }
-    SortPRecRNam(result, 1);
+    SortPRecRNam(result, 0);
     if (tnum == T_PREC + IMMUTABLE)
         RetypeBag(result, tnum);
     return result;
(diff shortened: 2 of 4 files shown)


fingolfin:
@ChrisJefferson you write "the extra argument" but don't say which function you are talking about. Could you clarify?

ChrisJefferson:
Sorry, `SortPRecRNam`, before this commit, takes a second argument to say if the sorting should be "In place". That argument must now always be 0, as we don't support inplace sorting any more (as it was slower, and never needed)

But, I can't just remove the argument as that would break datastructures.

fingolfin:
@ChrisJefferson Ah, I see. Hm, actually, looking at how `datastructures` uses it now... in `BasicRecursiveHashForPRec`. Ugh. Actually, I am strongly tempted to say we should remove `BasicRecursiveHashForPRec` from `datastructures`, make a release, and then we can change it here.

The reason why I am strongly considering to remove `BasicRecursiveHashForPRec` is that its results are very volatile and can change between different runs of GAP, depending on the order in which rnams are introduced (e.g. the order in which certain packages are used). While technically, that doesn't matter while inside a single session of GAP, this introduces substantial randomness and "non-determinism" (only superficially, of course, for a user who is not familiar with the C implementation of the hash function). And if somebody dares serialize that has, they could be in trouble.

But perhaps I am just too negative?

ChrisJefferson:
It annoys me exactly the same amount that `<` on records is also non-deterministic, but I've wanted to change that for years and felt like most people thought it shouldn't be changed.

ChrisJefferson:
Actually, I was considering a change for hashing, which wouldn't substansally slow things down -- we have to hash the actual name of the record, but we don't have to sort the name, we can hash pairs [key,value], and then combine those hashes with (for example) `+`.

Implementing `<` in a way it is non-determinstic I think just requires doing a sort of the keys of the record, which is considerably more expensive that the current `<`, but then again we might just have to pay the cost.

ChrisJefferson:
I decided to fix the hash :) ( https://github.com/gap-packages/datastructures/pull/111 ), but we should still wait a while before removing the extra parameter, to give time for a new datastructures to be released, and circulate (although we won't have to give that long, as I expect datastructures users to be prepared to update fairly often).

fingolfin:
Agreed on `\<` vs. hashing; and yay on the datastructures update.

PR #3272, opened 2019-02-06
kernel: add / document some options to the Julia GC code

fingolfin:
* DISABLE_BIGVAL_TRACKING
* REQUIRE_PRECISE_MARKING
* STAT_MARK_CACHE
* MARKING_STRESS_TEST

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -31,12 +31,41 @@
 #include "julia.h"
 #include "julia_gcext.h"
 
+
+/****************************************************************************
+**
+**  Various options controlling special features of the Julia GC code follow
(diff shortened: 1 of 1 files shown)


fingolfin:
Yes, the stress test is actually due to @ChrisJefferson who pointed out that it can be used to quickly trigger GC crashes, which is "good" in that it should help us debug them... Since I was preparing this PR anyway, I thought it best to just put it in with the rest.

fingolfin:
Backported to stable-4.10 in 02297fc53c18d9cdbc2b2537c23c1cb0ed15e36a

PR #3276, opened 2019-02-07
Support custom size function bags in InstallGlobalFunction

fingolfin:
Packages like JuliaInterface and PARIInterface allow defining special
GAP wrapper functions, which are implemented as T_FUNCTION bags with
some additional data in them. Right now, they abuse the `fexs` field for
that. In the future, they will use slightly larger function bags to
achieve their goal.

This patch makes it possible to use InstallGlobalFunction on such
wrapper functions.

@sebasguts This affects JuliaInterface. If we want JuliaInterface to be compatible with the next GAP 4.10 release, we need to backport this change.

### DIFF
--- a/src/opers.c
+++ b/src/opers.c
@@ -3025,9 +3025,7 @@ static void InstallGlobalFunction(Obj oper, Obj func)
     Obj name = NAME_FUNC(oper);
 
     // clone the function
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to 4.10 in commit 140a37a04b6d43b71798838cc79fbd7f129274b4

PR #3278, opened 2019-02-10
Add new method for IsSolvableGroup

hungaborhorvath:
For checking solvability, if group knows its size, then it should check for the Feit-Thompson and for the Burnside criteria first.
The Feit-Thompson check is already implemented as an ````ImmediateMethod````.

Was thinking of adding it as ````ImmediateMethod````, but ````IsPrimePowerInt```` can be more resource hungry than most of the immediate methods. Maybe the 2-group check could be implemented, though, as an immediate method?

Kind of surprised the Burnside theorem was never implemented before.


hungaborhorvath:
testmanuals fail for whatever reason. Anyone has any suggestions how I can fix it?

tst/testinstall/opers/IsSolvableGroup.tst
```diff
@@ -53,4 +59,13 @@ true
 gap> G := F/[ a^2, b^2, c^2, d^2, (a*b)^3, (b*c)^3, (c*d)^3, (a*c)^2, (a*d)^2, (b*d)^2 ];;
 gap> not IsSolvable(G) and HasIsAbelian(G) and not IsAbelian(G);
 true
+gap> G := F/[ a^2, b^2, c^2, d^2, (a*b)^3, (b*c)^3, (c*d)^3, (a*c)^2, (a*d)^2, (b*d)^2 ];; Size(G);;
+gap> not IsSolvable(G) and HasIsAbelian(G) and not IsAbelian(G);
+true
+gap> F := FreeGroup("a", "x");; a := F.1;; x := F.2;;
+gap> G := F/[x^2*a^8, a^16, x*a*x^(-1)*a];; Size(G);;
+gap> not HasIsSolvableGroup(G) and IsSolvableGroup(G) and HasIsPGroup(G) and IsPGroup(G) and HasPrimePGroup(G) and HasIsNilpotentGroup(G) and IsNilpotentGroup(G);
+true
```
fingolfin: We have by now removed most of "tests" like this, they are simply far too fragile. I see why you did it, but I think we need a better system to test individual methods.

fingolfin:
For the failing test, the manual examples will have to be adjusted. Looking at <https://api.travis-ci.org/v3/job/491232815/log.txt>, you can search for "but found". It will tell you above that were the mismatch occurred.

hungaborhorvath:
@fingolfin I adjusted the new tests. I checked with ````Tracemethods```` that it runs the new method, anyway. 

I saw that there are a lot of similar (````HasIsAbelian````, etc.) tests in the old version of ````IsSolvableGroup.tst````. Should I rewrite those, as well? 

As for the manuals, I am really puzzled what these couple lines managed to mess up. I checked what you suggested, there were 5 problems. 1 in the tutorial, 2 about homomorphisms, 2 about ````TransformingPermutationsCharacterTables````. Are we sure that they were running with no error before I merged my changes? In any case, the examples there are rather fragile.

hulpke:
Is this actually addressing any concrete problem (that is, are there examples of groups that can determine the size cheaply, but for which the solvability test is more expensive), or is this just a tehoretical addition for hypothetical cases?

At least for permutation groups and matrix groups (and thus also for FpGroups) I expect that testing solvability is cheaper than determining the size (and will determine the size on the way), using e.g. Sims' solvable Stabilizer chain algroithm, respectively composition tree. Thus I'm not sure what this method actually achieves.

(Also, typically one will want not just a statement that a group is solvable, but actually a corresponding data structure, e.g. a Pcgs.)

hungaborhorvath:
@hulpke The idea came when the other day someone asked on the forum about the normal subgroups of the unit group of F_4D_8. For me the solvability finished in more than 90 seconds (checked only for due diligence, at the time I wasn't willing to wait for it), but size could have been determined instantenously, and therefore it was clear that the group is solvable. (It was even nilpotent, as @Stefan-Kohl determined). I have to admit, though, that ````CanComputeSize```` was ````false```` for that group, even though determining size took only 4 ms. 

Would you prefer if I replaced ````CanComputeSize```` with ````HasSize````?

hungaborhorvath:
Can someone help me what I need to do about the failing testinstall suite?

ChrisJefferson:
The issue with the 'JULIA=yes' test you can ignore, it's not a problem you have caused.

hulpke:
@hungaborhorvath Using `HasSize` would at least avoid serious performance problems. (E.g. for certain larger nonsolvable permutation groups where a quick nonsolvability test is desired *before* calculating the size.) Thus I would be strongly in favor of such a change. Note that this might revert some of the examples that changed.

`CanComputeSize` is just about capability to do so -- e.g. a finitely resenetd group would not know it a priori.

(I also think it is not good policy to use very much nonstandard examples -- such as this group of units in F4D8, which cannot do anything but element arithmetic -- to design general functionality.)

hungaborhorvath:
@hulpke Replaced ````CanComputeSize```` by ````HasSize````.

lib/grp.gi
```diff
@@ -573,6 +577,32 @@ InstallImmediateMethod( IsSolvableGroup, IsGroup and HasSize, 10,
...
     end );
 
+InstallMethod( IsSolvableGroup,
+    "if group size is known and is 2m or p^a q^b",
+    [ IsGroup and HasSize ], 25,
+    function( G )
+    local size;
+    size := Size( G );
+    if IsInt( size ) then
+      if size mod 4 <> 0 then
```
hulpke: In view of the immediate metjod just above: Will this case ever arise?

lib/grp.gi
```diff
@@ -573,6 +577,32 @@ InstallImmediateMethod( IsSolvableGroup, IsGroup and HasSize, 10,
...
+        return true;
+      else
+        size := size/4;
+        while size mod 2 = 0 do
```
hulpke: This will only ever catch 2^a q^b, but not p^aq^b as claimed

hungaborhorvath: If a group has size p^a q^b for odd p and q, then it is already odd, hence is already caught by the Feit-Thompson test. Otherwise one of the primes is 2, and the size looks like 2^a q^b.

lib/grp.gi
hungaborhorvath: As long as the Immediate method is in place, it won't. 

However, I would still prefer this to be left here, in case the immediate method is ever removed (in which case this method would not need adjusting). 

Originally, I was thinking about changing the immediate method, but finally decided against it, because felt that a possible check for being prime power, even though it is reasonably cheap, is not as cheap as a modulo 4 check is.

lib/grp.gi
hulpke: Thanks. The installation text had confused me.

tst/testinstall/opers/IsSolvableGroup.tst
hungaborhorvath: Rewrote the tests.

lib/grp.gi
hungaborhorvath: Good point. Rewrote the text in the method.

PR #3281, opened 2019-02-12
Fix declaration of sparse action homomorphisms.

### LINKED ISSUE #3279: problem with TestIdentityAction
The following happens in the current master branch as well as in released versions
(at least 4.10.0, 4.9.1, 4.7.6).
```
gap> g:= GL(2,3);;
gap> SparseActionHomomorphism( g, [ [ Z(3), Z(3) ] ] );
Error, Action not well-defined. See the manual section
``Action on canonical representatives''. at /.../gapmaster/gap/lib/oprt.gd:441 called from
TestIdentityAction( acts, pnt, act 
 ) at /.../gapmaster/gap/lib/oprt.gd:840 called from
...
```
The function `SparseActionHomomorphism` is called in a generic `NiceMonomorphism` method for matrix groups, try for example `Size( Group( GeneratorsOfGroup( GL( 2, Integers mod 6 ) ) ) )`.
The arguments of the (undocumented) function `TestIdentityAction` are the list of group generators, the list of seed vectors, and the action function (here the default `OnPoints`),
and the idea behind  `TestIdentityAction` is to check whether the `One` of the first generator fixes each seed vector w.r.t. the action function.
In the situations where `SparseActionHomomorphism` is called in GAP library code,
this does not run into an error because then the second argument is the (list of rows of) the identity element in the matrix group.
```
gap> g:= GL(2,3);;
gap> SparseActionHomomorphism( g, One( g ) );
<action epimorphism>
```
Note that in this case, `TestIdentityAction` tests in fact whether the `One` of the first generator *conjugates* the list of seed vectors to itself, in the sense that `<pnt>^<one>` is computed as `<one> * <pnt> * <one>`, which is fine only if `<pnt>` is square.
If one specifies the action function `OnRight` instead of `OnPoints` then the above example works.
(There is even a manual example that shows this situation.)
```
gap> g:= GL(2,3);;
gap> SparseActionHomomorphism( g, [ [ Z(3), Z(3) ] ], OnRight );
<action epimorphism>
```
I think that the check in `TestIdentityAction` must apply the identity element to each seed vector separately, since we cannot assume that the action function knows how to deal with a list of seed vectors.
Perhaps the cleanest solution would be to replace the `TestIdentityAction` call by an appropriate `Assert` call; note that `TestIdentityAction` seems to be not used in other places.
The only argument against this change is that currently the list of seed vectors gets replaced by the result of `TestIdentityAction` (why?), but on the other hand it seems to be allowed to skip `TestIdentityAction` via the option `NoTestAction`.

hulpke:
Sparse Action homomorphisms cannot use `OrbitishFO`  but need to use their own creator function (as point seed is list of points and  thus need to apply `TestIdentityAction` differently.)

 This fixes #3279

### DIFF
--- a/lib/oprt.gd
+++ b/lib/oprt.gd
@@ -1082,7 +1082,7 @@ local str, orbish, func;
 
     # Create the wrapper function.
     func := function( arg )
-    local   G,  D,  pnt,  gens,  acts,  act,  xset,  p,  attrG,  result,le;
+    local   G,  D,  pnt,  gens,  acts,  act,   p,  attrG,  result,le;
 
       # Get the arguments.
       if 2 <= Length( arg ) then
@@ -1090,7 +1090,6 @@ local str, orbish, func;
 	  G := arg[ 1 ];
 	  if IsFunction( arg[ le ] )  then
 	      act := arg[ le ];
-	      le:=le-1;
 	  else
 	      act := OnPoints;
 	  fi;
@@ -1111,8 +1110,7 @@ local str, orbish, func;
 	      acts := arg[ p + 2 ];
 	  fi;
       else
-	Error( "usage: ", name, "(<xset>,<pnt>)\n",
-	      "or ", name, "(<G>[,<Omega>],<pnt>[,<gens>,<acts>][,<act>])" );
+	Error( "usage: ", name, "(<G>[,<Omega>],<pnts>[,<gens>,<acts>][,<act>])" );
       fi;
       
       if not IsBound( gens )  then
@@ -1131,12 +1129,7 @@ local str, orbish, func;
       if IsBound( D )  then
 	  result := orbish( G, D, pnt, gens, acts, act );
       else
-
-	  # The following line is also executed  when `Blocks(<G>, <Omega>, <act>)'
-	  # is called to compute blocks with no  seed, but then <pnt> is really
-	  # <Omega>, i.e., the operation domain!
 	  result := orbish( G, pnt, gens, acts, act );
-
       fi;
 
       return result;
--- a/lib/oprt.gd
+++ b/lib/oprt.gd
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-OrbitishFO( "SparseActionHomomorphism", OrbitishReq,
-               IsIdenticalObj, false,false );
-OrbitishFO( "SortedSparseActionHomomorphism", OrbitishReq,
-               IsIdenticalObj, false,false );
+
+BindGlobal( "SparseActHomFO", function( name, reqs, famrel )
+local str, orbish, func;
+
+    # Create the operation.
+    str:= SHALLOW_COPY_OBJ( name );
+
+    APPEND_LIST_INTR( str, "Op" );
+    orbish := NewOperation( str, reqs );
+    BIND_GLOBAL( str, orbish );
+
+    # Create the wrapper function.
+    func := function( arg )
+    local   G,  D,  pnt,  gens,  acts,  act,  xset,  p,  attrG,  result,le;
+
+      # Get the arguments.
+      if 2 <= Length( arg ) then
+	  le:=Length(arg);
+	  G := arg[ 1 ];
+	  if IsFunction( arg[ le ] )  then
+	      act := arg[ le ];
+	      le:=le-1;
+	  else
+	      act := OnPoints;
+	  fi;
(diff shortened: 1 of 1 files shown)


lib/oprt.gd
```diff
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
...
+    func := function( arg )
+    local   G,  D,  pnt,  gens,  acts,  act,  xset,  p,  attrG,  result,le;
+
+      # Get the arguments.
+      if 2 <= Length( arg ) then
+	  le:=Length(arg);
+	  G := arg[ 1 ];
+	  if IsFunction( arg[ le ] )  then
+	      act := arg[ le ];
+	      le:=le-1;
```
ThomasBreuer: The local variable `le` is not used later on.

lib/oprt.gd
```diff
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
...
+	  else
+	      p := 2;
+	  fi;
+	  pnt := Immutable(arg[ p ]);
+	  if Length( arg ) > p + 1  then
+	      gens := arg[ p + 1 ];
+	      acts := arg[ p + 2 ];
+	  fi;
+      else
+	Error( "usage: ", name, "(<xset>,<pnt>)\n",
```
ThomasBreuer: The case of `<xset>` is not supported by the functions created by this function.

lib/oprt.gd
```diff
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
...
+	      "or ", name, "(<G>[,<Omega>],<pnt>[,<gens>,<acts>][,<act>])" );
+      fi;
+      
+      G := PreOrbishProcessing(G);
```
ThomasBreuer: Just for curiosity: Is `PreOrbishProcessing` used anywhere?
The only method provided in the library returns its argument.

lib/oprt.gd
```diff
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
...
+        TestIdentityAction(acts,p,act);
+      od;
+
+      if IsBound( D )  then
+	  result := orbish( G, D, pnt, gens, acts, act );
+      else
+
+	  # The following line is also executed  when `Blocks(<G>, <Omega>, <act>)'
+	  # is called to compute blocks with no  seed, but then <pnt> is really
+	  # <Omega>, i.e., the operation domain!
```
ThomasBreuer: The comment on `Blocks` is not necessary here.

lib/oprt.gd
```diff
@@ -1075,10 +1075,85 @@ DeclareGlobalFunction("MultiActionsHomomorphism");
...
+	      p := 2;
+	  fi;
+	  pnt := Immutable(arg[ p ]);
+	  if Length( arg ) > p + 1  then
+	      gens := arg[ p + 1 ];
+	      acts := arg[ p + 2 ];
+	  fi;
+      else
+	Error( "usage: ", name, "(<xset>,<pnt>)\n",
+	      "or ", name, "(<G>[,<Omega>],<pnt>[,<gens>,<acts>][,<act>])" );
```
ThomasBreuer: `<pnt>` should perhaps be replaced by `<pnts>`.

lib/oprt.gd
hulpke: Thanks -- will clean up

lib/oprt.gd
hulpke: Ditto

lib/oprt.gd
hulpke: I'm not sure what it is intended for. I will remove this operation.

lib/oprt.gd
hulpke: Done

lib/oprt.gd
hulpke: Done

hulpke:
Thanks @ThomasBreuer 

I agree with your remark concerning the `FO` function. I think a (related) question is how much the "external set" option (which is not really used anywhere) needs to be supported. Maybe this is best moved into a sepaarte issue.

alex-konovalov:
I think this should this be backported to stable-4.10 and mentioned in release notes.

fingolfin:
Backported to stable-4.10 in commits 7abb7eb4edd34b3ae64cb0591ccbb3e80f4e7bef, 151b4a3fd63884e3d7d4875974585e2a4c991bf3, 8fa683039be18004bc727eef5e0302d18877a99f

PR #3285, opened 2019-02-12
buildsys: add `make check`

### LINKED ISSUE #1439: Add "make check" target
The virtual standard for invoking the test suite of `make` based build systems is `make check`. We also should support that. Multiple ways to do that:
* could be analias for e.g. `testinstall`
* could simply print a list of the actual test targets and what they do
* ...?

fingolfin:
People familiar with automake based build systems expect this to work, and it also saves me some typing ;-)

Resolves #1439

I'd like to backport this to 4.10, too.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -1014,6 +1014,10 @@ testbugfix: all
           ReadGapRoot( "tst/testbugfix.g" );' | $(TESTGAP) | \
             tee `date -u +dev/log/testbugfix2_%Y-%m-%d-%H-%M` )
 
+#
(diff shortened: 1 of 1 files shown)


alex-konovalov:
@fingolfin Do you want to add some command line options, like `-r -A`?

fingolfin:
@alex-konovalov `TESTGAP` already contains `-r -A`

fingolfin:
Backported to 4.10 in eef7f9931295b2503efa90001724ca350ead27d3

PR #3287, opened 2019-02-13
Towards `make install` support: tweak usage of `config.h`

fingolfin:
One major obstacle for `make install` target is our `config.h` file, which is currently included by all other GAP headers. But installing it unchanged is a big no-go for various reasons:

First off, if our headers are included by some other code which also uses autoconf, and also has a `config.h`, there can be a clash of `#define` names. In fact, there is *always* a clash for `PACKAGE_NAME`, `PACKAGE_STRING`, etc., and indeed, several GAP packages contain workaround for this.

One "solution" for this is to prefix all `#defines` in `config.h` with `GAP_`, which we can perhaps achieve with the [ax_prefix_config_h extension for autoconf](https://www.gnu.org/software/autoconf-archive/ax_prefix_config_h.html) or some similar hack. 

But even if we did that, this is not quite enough: some of these `#define`s are for compiler specific properties, but 3rd party code using our headers may be using a different compiler (think gcc vs. clang; or different gcc versions; etc.; on otherwise the same machine); and this is perfectly legal and valid. But if for example the compiler used to compile GAP supports the `__builtin_popcountl` intrinsic, we `#define HAVE___BUILTIN_POPCOUNTL 1` or possibly with the above extension `#define GAP_HAVE___BUILTIN_POPCOUNTL 1`. But if 3rd party code uses a compiler *without* that intrinsic, we'd now run into a compiler error, if the same `config.h` as previously generated when building GAP is used.

To fix this, we must not blindly rely on macros for compiler specific features in *headers*. We need to either avoid them, or code them defensively.

This PR moves us a step closer to solving this, and thus moves us a step closer to proper `make install` support.

### DIFF
--- a/src/intobj.h
+++ b/src/intobj.h
@@ -220,31 +220,29 @@ EXPORT_INLINE int diff_intobjs(Obj * o, Obj l, Obj r)
 **  The product itself is stored in <o>.
 */
 
+// check for __builtin_mul_overflow support
+#if defined(__has_builtin)
+  // clang >= 3.8 supports it, but better to check with __has_builtin
+  #if __has_builtin(__builtin_mul_overflow)
+  #define HAVE___BUILTIN_MUL_OVERFLOW 1
+  #endif
+#elif defined(__INTEL_COMPILER)
+  // icc >= 19.0 supports it; but already version 18.0 claims to be GCC 5
+  // compatible, so we must perform this check before that for __GNUC__
+  #if __INTEL_COMPILER >= 1900
+  #define HAVE___BUILTIN_MUL_OVERFLOW 1
+  #endif
+#elif defined(__GNUC__) && (__GNUC__ >= 5)
+  // GCC >= 5 supports it
+  #define HAVE___BUILTIN_MUL_OVERFLOW 1
+#endif
 
-#if SIZEOF_VOID_P == SIZEOF_INT && defined(HAVE_ARITHRIGHTSHIFT) &&          \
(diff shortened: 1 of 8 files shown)


src/intobj.h
```diff
@@ -109,14 +109,13 @@ EXPORT_INLINE Int ARE_INTOBJS(Obj o1, Obj o2)
 /* Note that the C standard does not define what >> does here if the
  * value is negative. So we have to be careful if the C compiler
  * chooses to do a logical right shift. */
+GAP_STATIC_ASSERT((-1) >> 1 == -1, "right shifts are not arithmetic");
+GAP_STATIC_ASSERT((-2) >> 1 == -1, "right shifts are not arithmetic");
```
sebasguts: Why do you need the second test? I think the first should already distinguish arithmetic from logical shift.

fingolfin: It's there for paranoia: in practice, >> could also do something else than arithmetic or logical shift, which then by chance might work for -1 but not -2... Of course we can never test all cases, but I figured, it wouldn't hurt to test one more...

sebasguts: If you apply this logic here, shouldn't we make sure the left shifts also work? Or is there no reason to believe that, say, it is a circular shift?

fingolfin: The C standard guarantees how left shifts work, and how right shifts for non-negative numbers work. Only right shifts of negatives are implementation specific. Presumably this is to enable efficient operations on both systems with twos complement and ones complement, or with a sign bit..

sebasguts: Okay :)

PR #3293, opened 2019-02-13
kernel: misc changes

fingolfin:


src/permutat.cc
```diff
@@ -1363,7 +1363,7 @@ static Obj FuncCYCLE_STRUCT_PERM(Obj self, Obj perm)
 **  'OrderPerm' returns the  order  of  the  permutation  <perm>,  i.e.,  the
 **  smallest positive integer <n> such that '<perm>\^<n>' is the identity.
 **
-**  Since the largest element in S(65536) has oder greater than  10^382  this
```
stevelinton: Apart from the typo fixed here, this comment implicitly  assumes that permutations are limited to degree 65536. Maybe just delete this couple of lines, or something like `Since permutations can have order exponentially larger than their degree`

fingolfin: Actually, I am reading this more as saying ", look, even if the permutation is a T_PERM2, it can have a huge order; and of course for T_PERM4, it can be even larger."

PR #3295, opened 2019-02-14
kernel: move CopyBits and some related code to new header bits_intern.h

fingolfin:
Partial motivation: we should not be using ALWAYS_INLINE in header
files, as that makes it harder to install these headers properly as
part of `make install`


ChrisJefferson:
I'd don't understand this PR.. you say we shouldn't use ALWAYS_INLINE in headers, so you move some ALWAYS_INLINE from one header, to another header?

I'm guessing you plan for _intern to have a "special meaning". Can you discuss (and maybe even document?) that fact?

fingolfin:
Perhaps, I don't really know yet. We could also not install them at all; but perhaps some would be required by packages

fingolfin:
@ChrisJefferson right, I should clarify that in the commit message. In short, `ALWAYS_INLINE` should not be used in "public" headers that packages might sensible include... The plan is to eventually stop including "config.h" from any header files. Then change `bits_intern.h` (and possibly other "internal" headers) to verify "config.h" was included before them. Then finally, include config.h from all (?) .c and .cc files, as the last header (well, except for `bits_intern.h`, which would come after `config.h`)

fingolfin:
@ChrisJefferson does that make sense / is it clearer? If so, I can update the commit message.

fingolfin:
@ChrisJefferson ping? I will gladly augment the commit message, but would appreciate if you let me know if that explanation seems reasonable (otherwise it's a waste of time). Thank you!

ChrisJefferson:
That sounds ok

fingolfin:
Updated the commit message.

PR #3301, opened 2019-02-15
kernel: refactor vec8bit field info code

fingolfin:
This induces a memory usage overhead of ~27 MB in the worst case; details can be found in the commit message of the third commit.

### DIFF
--- a/src/vec8bit.c
+++ b/src/vec8bit.c
@@ -538,7 +538,8 @@ static void RewriteVec8Bit(Obj vec, UInt q)
     // UInt mut = IS_MUTABLE_OBJ(vec);
     UInt mult;
 
-    UInt1 *       gettab1, byte1;
+    const UInt1 * gettab1;
+    UInt          byte1;
     const UInt1 * ptr1;
     const UInt1 * settab;
     UInt1 *       ptr;
@@ -962,7 +963,7 @@ void PlainVec8Bit(Obj list)
     UInt    q;
     UInt    elts;
     Obj     info;
-    UInt1 * gettab;
+    const UInt1 * gettab;
     Obj     fieldobj;
     Char *  startblank;
     Char *  endblank;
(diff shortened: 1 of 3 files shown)


src/vec8bit.c
```diff
@@ -273,10 +272,6 @@ static void MakeFieldInfo8Bit(UInt q)
     UInt val0;
     UInt elt, el1, el2;    // used to build up some answers
     const FFV * succ;
-    UInt1 * setelt_info;    // Cache a value, mainly to get around a bug in
```
ChrisJefferson: You are removing some code which worked around a bug in older versions of xcode and clang. I'm not entirely sure when this bug was eventually fixed. Have you checked?

ChrisJefferson: The bug was "the compiled code was wrong with optimisation turned on". Exactly why it broke, I do not know. Please mention the removal in the commit message, so a future programmer will be able to see if the bug was reintroduced.

fingolfin: So I tracked down the discussion for this on the GAP devel mailing list. Turns out the bug was that the compiled code segfaulted, and e.g. `testinstall` already caused that. So I am pretty sure if the bug was to reappear, we'd notice, instead of corrupt data, which is good. Also, the analysis of the problem was that hoisting the `SETELT_FIELDINFO_8BIT(info)` computation out of the loop solved the crash. With this PR, this is not changed, which is why I am confident that we don't re-trigger the crash.

As to updating the commit message: OK, I can, but I honestly don't know what to write. Perhaps this?
> Removed a workaround for a compiler bug in Xcode 5.0 that could lead to a compiler crash

However, that is actually not strictly true, as the workaround was not so much "removed", as the code in which it occurred was completely rewritten. Whether that causes the compiler bug to take effect again, or not, or possibly triggers a different compiler bug in a different compiler, is hard to tell. The same holds for any refactoring.

fingolfin: Oops, that should be:
> Removes a work workaround for a compiler bug in Xcode 5.0 that could lead to a crash in the resulting binary
or so

fingolfin: (I augmented the commit message with something in this vein now)

ChrisJefferson: That's OK. Hopefully whoever is running an old compiler won't find the bug has returned.

PR #3303, opened 2019-02-15
kernel: mark more global vars as static, and add a warning that will report if new ones are introduced

fingolfin:


Makefile.rules
```diff
@@ -71,7 +71,7 @@ SOURCES += src/listoper.c
 SOURCES += src/lists.c
 SOURCES += src/macfloat.c
 SOURCES += src/modules.c
-SOURCES += src/modules-builtin.c
```
ChrisJefferson: why rename a file, and break git log and everything else?

fingolfin: This is the only file with a dash instead of an underscore. The log works fine for me. Not sure what you mean with "everything else"?

ChrisJefferson: git log doesn't follow the rename. git blame doesn't if you give a commit. Between renames and reformats, I'm finding it increasingly difficult to track the history of code.

PR #3304, opened 2019-02-15
kernel: remove redundant break and return statements

fingolfin:


fingolfin:
We are already testing with GCC 4.7. On older compilers, which don't support the noreturn attribute, there might still be warnings -- but no more, and those can be turned off; just like we are currently turning off (or not turning on) the warnings which warn about unreachable code (which prompted me to make this PR).

That said, I am not particular sad if we don't merge this either *shrug*

fingolfin:
Oh, and I actually tried to make sure to only remove returns which are followed by more code, so that (I think) even compilers w/o the noreturn attribute won't warn about this. I left in those were the return was the last line of the function, and hence would trigger a warning on an "old" compiler if removed.

stevelinton:
OK Fine with this

PR #3308, opened 2019-02-18
More api tests

markuspf:


markuspf:
Some code to make sure that the API runs; there are not many tests yet that for example establish that objects that are created via the API are the objects we want, or whether they still work after a GC, etc.

markuspf:
Might still add the remaining tests for floats and characters, but should then be merged (and extended later).

PR #3314, opened 2019-02-23
kernel: remove more uses of SYS_IS_64_BIT from headers

fingolfin:


src/gasman.h
```diff
@@ -81,12 +81,8 @@ typedef UInt * *        Bag;
 typedef struct {
     uint8_t type : 8;
     uint8_t flags : 8;
-#ifdef SYS_IS_64_BIT
-    uint64_t size : 48;
-#else
-    uint16_t reserved : 16;
-    uint32_t size : 32;
-#endif
+    uint16_t : (sizeof(UInt) == 8) ? 0 : 16;
```
ChrisJefferson: Can you add a comment here, just to explain what's going on? I realise 'reserved' wasn't well explained, but at least it had a name :)

fingolfin: The immediate reason for this is the following I get if I give the field a name:
> error: named bit-field 'reserved' has zero width

So why not drop the field? Because I'd like to make sure that the size field is aligned optimally on a 32bit build... I guess I could also achieve this by moving it before type and flags... ?

I agree a comment for this would be good, but I have to think about what to write. Suggestions are of course also welcome :)

ChrisJefferson: I think just "add padding to ensure the structure is optimally aligned on 32-bit architectures" is sufficient, just to make sure it doesn't look like an "accident" or something.

fingolfin: I've now added a comment

PR #3313, opened 2019-02-22
kernel: export COUNT_TRUES_BLOCK(S) again

fingolfin:
This is used by the Digraphs packages. We probably don't want to keep
COUNT_TRUES_BLOCK(S) in compiled.h, but this was the quickest thing I
could think of to just restore behavior. We can work on a better fix
later, but right now I just want to unbreak Digraphs.

Resolves #3312

PR #3317, opened 2019-02-26
Move COUNT_TRUE_BLOCKS and COUNT_TRUE_BLOCK back to a public header

ChrisJefferson:
This just gets digraphs working again.


ChrisJefferson:
Alternative to #3313, just to get GAP building digraphs again for now.

PR #3318, opened 2019-02-27
Move COUNT_TRUES_BLOCK(S) out of header

ChrisJefferson:
This moves COUNT_TRUES_BLOCK(S) back out of the header.

After investigation, I decided to just make them normal non-inlined functions. COUNT_TRUES_BLOCKS should take long enough function overhead is not significant, and COUNT_TRUES_BLOCK can be inlined with link-time optimisation. If we enable link-time optimisation, lots of other little functions can be inlined away, and nowadays most modern compilers support it, and it's easy to use (I wouldn't want to use link time optimsiation every time I compile, just because it leads to long compile times even when you only make a small change).

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -806,6 +806,57 @@ void ConvBlist (
     SET_LEN_BLIST( list, len );
 }
 
+/****************************************************************************
+**
+*F  COUNT_TRUES_BLOCK( <block> ) . . . . . . . . . . .  count number of trues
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Just as a datapoint, GAP's binary contains about 34,000 function calls with normal optimisation, of which about 4,000 are inlined with link time optimisation.

PR #3319, opened 2019-02-28
Collection of localized cleanups

hulpke:
These fix problems that I noted over the last months. They arose in larger calculations, thus there is not always testing code.

### DIFF
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -1060,8 +1060,9 @@ function ( fgens, grels, fsgens )
   if Length(fgens)=0 then
     return [];
   fi;
-  # call the TC plugin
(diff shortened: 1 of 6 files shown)


ChrisJefferson:
While this isn't a requirement for this PR, I'd be happy to see more tests which take hours (as long as they take less than a day say), to really test gap. Of course, such tests would preferably test some maths result, and not be effected by things like changes in random behaviour.

If you have any such tests, please post them somewhere and I will organise a test suite to run them.

hulpke:
@ChrisJefferson Thank you. Part of the size is not the actual calculation time, but setting up the situation in which the problem arises. In future I'll think about whether I can make it testable in a format less cumbersome than the tst files.

Issue #3306, opened 2019-02-18
Displaying package loading errors in a CI environment


zickgraf:
For CI purposes I look for a way to disable banners and still display package loading errors (i.e. package not found, dependent package not found, dependent package not available with the required version) without displaying anything if there are no errors.

I would expect this to be be possible by setting
```
SetInfoLevel( InfoPackageLoading, PACKAGE_ERROR );
```
or

```
SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
```

However, if I execute
```
SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
LoadPackage( "PackageDependingOnNonExistentPackage" );;
```

with banners disabled, I get no output.

Is this the intended behaviour? Is there a good way to achieve my aim?

fingolfin:
I don't see how this relates to banners. Here is what I get with banners ON; notice how with `PACKAGE_WARNING` it doesn't really give any useful information about what caused the failure, or how to resolve it (and that indeed seems to not match what the manual says); only with `PACKAGE_INFO` do I get that info:
```
gap> SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
gap> LoadPackage("PackageDependingOnNonExistentPackage");
#I  PackageDependingOnNonExistentPackage package is not available. To see further details, enter
#I  SetInfoLevel(InfoPackageLoading,4); and try to load the package again.
fail
gap> SetInfoLevel( InfoPackageLoading, PACKAGE_INFO );
gap> LoadPackage("PackageDependingOnNonExistentPackage");
#I  PackageDependingOnNonExistentPackage: PackageAvailabilityInfo for version 1.9.4dev
#I  foobar: PackageAvailabilityInfo: no installed version fits
#I  PackageDependingOnNonExistentPackage: PackageAvailabilityInfo: dependency 'foobar' is not satisfied
#I  PackageDependingOnNonExistentPackage: PackageAvailabilityInfo: no installed version fits
fail
```

But that latter information is also available if banners are turned OFF via the `-b` command line option:
```
gap> SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
gap> LoadPackage("PackageDependingOnNonExistentPackage");
fail
gap> SetInfoLevel( InfoPackageLoading, PACKAGE_INFO );
gap> LoadPackage("PackageDependingOnNonExistentPackage");
[...]

PR #3322, opened 2019-03-02
Improve info if needed dep for a package does not exist

fingolfin:
Print the "dependency XYZ is not satisfied" info message with
warning level PACKAGE_WARNING instead of PACKAGE_INFO. This seems
to be more in line with what the manual promises, namely:

> PACKAGE_WARNING should be used whenever GAP has detected a reason why a
> package cannot be loaded, and where the message describes how to solve
> this problem, for example if a package binary is missing.

So if a package cannot be loaded because one of its needed dependencies
is not the name of any known package, that certainly is something the
user would like to know.

Before this change:

    gap> SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
    gap> LoadPackage("pkg_with_missing_dep");
    fail

After this change:

    gap> SetInfoLevel( InfoPackageLoading, PACKAGE_WARNING );
    gap> LoadPackage("pkg_with_missing_dep");
    #I  pkg_with_missing_dep: PackageAvailabilityInfo: dependency 'missing_pkg' is not satisfied
    fail

Motivated by issue #3306

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -875,7 +875,7 @@ InstallGlobalFunction( PackageAvailabilityInfo,
           if testpair = false then
             # This dependency is not satisfied.
             test:= false;
-            LogPackageLoadingMessage( PACKAGE_INFO,
+            LogPackageLoadingMessage( PACKAGE_WARNING,
                 Concatenation( "PackageAvailabilityInfo: dependency '",
                     name2, "' is not satisfied" ), inforec.PackageName );
             if not checkall then


alex-konovalov:
So if A has B as suggested and B has C as needed, then loading A will trigger a warning saying that B can not be loaded, since C is missing, isn't it?

fingolfin:
Ah, indeed. But I think that can be fixed with some moderate effort.

fingolfin:
OK, I had a closer look at this whole package dependency resolution code, and it is a bit ... complex. Not sure how much of that is necessary complexity, and how much not; I'd rather reserve judgement on that for now, as I am pretty sure there are more pitfalls and corner cases I am not actively aware of which might necessitate that complexity.

For the error / warning message at hand, here is what I think when a user should or shouldn't see it by "default". Meaning: at least on on info level PACKAGE_WARNING; but actually, I'd rather see *more* of these at the default info level; I know this was a controversial topic so far, but I think that's mostly because the current system makes it hard to show those messages only "when appropriate". So I am hoping that by improving this, we can simultaneously resolve the issue at hand, but also make it more palatable to lower some of these warning levels...

So, when should the "X could not be loaded dependency 'Y' is not satisfied"  message be shown?
- when using PACKAGE_DEBUG, of course always
- when using PACKAGE_WARNING (or even on the default level):
  - always if the user tried to load X directly (via LoadPackage("X"))
  - NOT if the user only queried whether the package is available
  - always if the user tried to load package Z directly, where Z needs X
  - NOT if the user tried to load package Z' directly, where Z only suggests X
  - always if the user loads package W, which needs both Z and Z'
  - always *but only once* if the user loads package W', which needs two or more packages which in turn need X

Some ideas on how to tackle that: 
- we should keep track if we already printed a given warning/error for a package, in case we revisit it
  - for this we could employ a "cache" (e.g. implemented as a record, mapping package names to a "status") which we init to empty at the start of `LoadPackage` (resp. other entry points for loading), and in which we keep track of these things
- during dependency resolution, we should keep track of whether we are really trying to load something (then show more warnings), or just trying to find out if we *could* load something (then perhaps show fewer warnings, or only on a higher level)
  - rationale for this: if the user does `LoadPackage`, they'll expect helpful error messages if that fails;
  - while if a package X checks if it can load a package Y, in order to safely and properly deal with Y not being available, then this should not trigger such a warning or error messages
    - of course another way to tackle this could be to introduce a mechanism for detecting whether `LoadPackage` was entered interactively by the user, or from a package. That requires some hacking, but overall shouldn't be too hard (famous last words). Not sure it'd be worth it, though?
- we should keep track if a package is scheduled for loading because it is "needed", or just because it was "suggested";
  - this then needs to be tracked recursively: if package A suggests B, and B needs C, then loading A with C missing should not warn about C missing
  - but note that as we proceed through the dependency graph, this status may change: if we are trying to load A which suggests B; but A also needs D; and D needs B; then of course overall, B is needed. This status change then must be handled appropriately

wilfwilson:
@fingolfin Do you want this PR to be merged, or do you want to wait for a more thorough solution, along the lines you indicated above? I'm happy to merge it if you are.

fingolfin:
@wilfwilson problem is that while this PR improves one situations, it (arguably) makes another worse, namely the one @alex-konovalov described...

PR #3323, opened 2019-03-02
Misc kernel changes

fingolfin:


ChrisJefferson:
The commit message for 'kernel: turn PtrBody into void pointer ' seems to just stop mid-sentence (or is that github?)

src/iostream.c
```diff
@@ -62,14 +62,14 @@
 typedef struct {
   int childPID;   /* Also used as a link to make a linked free list */
   int ptyFD;      /* GAP reading from external prog */
-  UInt inuse;     /* we need to scan all the "live" structures when we have
```
ChrisJefferson: What is clang complaining about here? Seems a bit weird to turn unsigned things into signed?

fingolfin: Most of these struct members are actually booleans (which is hard to see from the definition, of course; it would have been nicer with my PR that introduced a `BOOL` typedef, but that was rejected).

Anyway, this means it doesn't matter whether we use UInt or int or whatever, we only need to distinguish "zero" from "non-zero".

The strict warning (enabled by `-Wpadded` resp. `-Weverything` here actually is that padding is introduced here (UInt is 8 bytes, int only 4). Since we only need one bit, it seems better to use the smaller data type for all. Of course I could have used UInt4 or Int4, but really, the code seems simpler to me if we just use `int` here. Of course we could also switch to a bitfield; e.g. like this:
```C
typedef struct {
  int childPID;   /* Also used as a link to make a linked free list */
  int ptyFD;      /* GAP reading from external prog */
  int status;     /* status from wait3 -- meaningful only if changed is 1 */
  unsigned inuse : 1;     /* we need to scan all the "live" structures when we have
                     had SIGCHLD so, for now, we just walk the array for
                     the ones marked in use */
  unsigned changed : 1;   /* set non-zero by the signal handler if our child has
                     done something -- stopped or exited */
  unsigned blocked : 1;   /* we have already reported a problem, which is still there */
  unsigned alive : 1;     /* gets set after waiting for a child actually fails
                     implying that the child has vanished under our noses */
  unsigned padding : 28;
} PtyIOStream;
```
Anyway, if you prefer `unsigned int` or `UInt4`, or the bitfield, or whatever, that's all also fine by me.

fingolfin:
Oops, sorry about the cutoff commit message, fixed now. Thanks for checking carefully, much appreciated!

src/iostream.c
ChrisJefferson: I'm happy with any, just wondered if the warning was related to the sign changing. I don't really like bitfields unless we really need the space, as they sometimes have slightly weird behaviour rules.

I don't really like `-Wpadded` (as it's really not an error), but if `-Weverything` includes it, then I don't mind getting rid of it just to stop people wondering what it's about.

fingolfin: Yeah, I also don't think -Wpadded is an important warning for us, and yeah my main motivation for this and several other changes in the past weeks has been to get `-Weverything` passing. However, in the end I had to disable a bunch of pointless warnings, so if I have to also add `-Wno-padding`, that's also fine. 

However, I thought this particular change is still nice-to-have, as with it, `sizeof(PtyIOStream)` goes from 48 down to 28 bytes (on 64bit systems). With the bitfield, it's even down to 16 bytes (also if one uses `char` for those 4 boolean entries). Since we have 64 of these structs, that's a reduction from 3kb to 1.8kb (resp. 1kb), which is not super critical (we waste far much more on path buffers), but hey, it's always nice to get rid of some pointless cruft ;-).

PR #3325, opened 2019-03-04
libgap: fix name of GAP_EnterDebugMessage_

fingolfin:
Without this, GAP_ENTER_DEBUG is broken.

Note sure if we plan to make a GAP 4.10.2 before GAP 4.11, but just in case we might want to backport this to stable-4.10.

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -412,7 +412,7 @@ void GAP_LeaveStack_(void)
     EnterStackCount--;
 }
 
-void GAP_EnterDebugMessage(char * message, char * file, int line)
+void GAP_EnterDebugMessage_(char * message, char * file, int line)
 {
     fprintf(stderr, "%s: %d; %s:%d\n", message, EnterStackCount, file,
             line);


fingolfin:
Also: ping @embray

dimpase:
Sage's libgap interface in Sage's 8.7.beta* is not using the affected function.

embray:
Thanks; I remember running into this at some point too but just assuming I had already fixed it elsewhere.

@dimpase Technically it is if libgap is compiled with some debug flags enabled, but it's not used normally.

dimpase:
@embray I meant to say Sage does not use it directly, neither in the interface nor in patches.

embray:
It's never used directly.  It's only used if `GAP_ENTER_DEBUG` is defined non-zero: https://github.com/gap-system/gap/blob/bf24ecdbedf86558c10b2ed66d689cb79ec43419/src/libgap-api.h#L26

fingolfin:
@dimpase well, if you might still approve this PR (Erik agrees that it is right), then we can merge it... :)

alex-konovalov:
This is mentioned in release notes for GAP 4.11.0 but seems to be included in GAP 4.10.2?

PR #3327, opened 2019-03-06
Enable -Wstrict-prototypes

fingolfin:


cnf/m4/ax_compiler_flags_cflags.m4
```diff
@@ -80,41 +80,42 @@ AC_DEFUN([AX_COMPILER_WARNING_FLAGS],[
...
             -Wextra dnl
-            -Wundef dnl
-            -Wwrite-strings dnl
-            -Wpointer-arith dnl
-            dnl -Wmissing-declarations dnl
-            -Wredundant-decls dnl
-            -Wno-unused-parameter dnl
-            -Wmissing-field-initializers dnl
-            -Wformat=2 dnl
+            -Warray-bounds dnl
```
ChrisJefferson: I can't be bothered reading all this, but it doesn't remotely fit the title of the PR, there are over 20 changed options here.

fingolfin: Commit message of the second commit says:
> buildsys: sort warning flags
> 
> By sorting the warning flags alphabetically, it becomes easier to discover if
some specific flag is in the list or not. It also helps to find duplicates
(e.g. we actually listed both -Winline and -Wno-inline).
8601fd2

fingolfin: I.e., no option was changed there, they only were sorted alphabetically (with `no-` prefix being ignored). Which revealed that we used both `-Winline` and `-Wno-inline`, so I think it's a good idea to do.

PR #3331, opened 2019-03-07
Fix finiteness condition in `IsomorphismGroups`

zickgraf:
Currently, the arguments of `IsomorphismGroups` are not handled symmetrically:

```
gap> F := FreeGroup(1);
<free group on the generators [ f1 ]>
gap> IsFinite( F );
false
gap> S := SymmetricGroup( 3 );
Sym( [ 1 .. 3 ] )
gap> IsFinite( S );
true
gap> IsomorphismGroups( S, F );
fail
gap> IsomorphismGroups( F, S );
Error, cannot test isomorphism of infinite groups at /usr/lib/gap/lib/morpheus.gi:2151 called from
<function "IsomorphismGroups">( <arguments> )
 called from read-eval loop at *stdin*:5
you can 'return;' to continue
brk>
```

After this commit, in both cases `fail` is returned.

### DIFF
--- a/lib/morpheus.gi
+++ b/lib/morpheus.gi
@@ -2151,8 +2151,11 @@ local m;
     IsFinite(G);
     IsFinite(H);
   fi;
-  if not IsFinite(G) and IsFinite(H) then
+  if not IsFinite(G) and not IsFinite(H) then
     Error("cannot test isomorphism of infinite groups");
+  fi;
+  if IsFinite(G) <> IsFinite(H) then
+    return fail;
   fi;
 
   #AH: Spezielle Methoden ?
--- a/tst/testbugfix/2019-03-08-IsomorphismGroups.tst
+++ b/tst/testbugfix/2019-03-08-IsomorphismGroups.tst
@@ -1,0 +1,9 @@
+# see https://github.com/gap-system/gap/pull/3331
+gap> F := FreeGroup( 1 );;
+gap> IsFinite( F );;
+gap> S := SymmetricGroup( 3 );;
+gap> IsFinite( S );;
+gap> IsomorphismGroups( F, S );
+fail
+gap> IsomorphismGroups( S, F );
+fail


ChrisJefferson:
This does improve things, thanks!

Two small things:

1) While it is a simple case, it would be nice to add a .tst file.

2) I wonder if, just after this check, it would be worth adding a check `if IsFinite(G) <> IsFinite(H) then return fail;`. Obviously in the case you give in your example this is being handled, but I wonder if an explicit test would be worthwhile (seeing as we already know if the groups are finite anyway).

lib/morpheus.gi
```diff
@@ -2151,7 +2151,7 @@ local m;
     IsFinite(G);
     IsFinite(H);
   fi;
-  if not IsFinite(G) and IsFinite(H) then
+  if not IsFinite(G) and not IsFinite(H) then
```
```suggestion
  if not IsFinite(G) or not IsFinite(H) then
```

fingolfin: Thinking some more about it: if one group is finite and the other is not, of course we know the answer, namely then the group is infinite... So how about this:
```gap
if IsFinite(G) <> IsFinite(H) then
  return fail;
fi;
if not IsFinite(G) then
  Error(...)
fi;
```

fingolfin:
Yes, please add a test! And I am surprised that this change made your two tests pass -- I don't see how, given that it only triggered if both inputs were infinite?! Maybe I am misunderstanding something?

lib/morpheus.gi
wilfwilson: I agree with @fingolfin's second suggestion here.

zickgraf:
@fingolfin My change worked because later in the code there is the following check:
```
  if Size(G)<>Size(H) then
    return fail;
```

What I do not like about @fingolfin's suggestion is that one cannot easily read off in which cases the error is triggered exactly. That's why I'm in favor of @ChrisJefferson's second suggestion:
```
if not IsFinite(G) and not IsFinite(H) then
  Error("cannot test isomorphism of infinite groups");
fi;
if IsFinite(G) <> IsFinite(H) then
  return fail;
fi;
```
Do you agree or shall I implement @fingolfin's suggestion?

wilfwilson:
Both pieces of code give the same answer in all cases...
```gap
if IsFinite(G) <> IsFinite(H) then
  return fail;
fi;
if not IsFinite(G) then
  Error(...)
fi;
```
```gap
if not IsFinite(G) and not IsFinite(H) then
  Error("cannot test isomorphism of infinite groups");
fi;
if IsFinite(G) <> IsFinite(H) then
  return fail;
fi;
```
...so I don't think it matters which one you choose.

*(If `G` and `H` are both finite, they both do nothing.
If `G` and `H` have different finiteness, then the first bit of code immediately returns `fail`; the second bit of code skips the first condition, and returns `fail` on the second.
If `G` and `H` are both infinite, the first bit of code skips the first condition, and gives an error on the second; the second bit of code immediately gives the error.
In other words, the two `if` statements "commute".)*

zickgraf:
My current commit also gives the same answer in all cases because of the existing check `Size(G)<>Size(H)`, so of course this is only a matter of code style.

ChrisJefferson:
The only reason to bail earlier is there could be cases where we know group G is infinite, group H is finite (for example, H could be a permutation group), but we don't actually know H's size, and calcaulating it is expensive.

However, this is just an optimisation, and fixing the bug is obviously more important.

zickgraf:
I have pushed a test file and my suggestion from above. I hope that I did put the test file in the correct directory.

fingolfin:
I disagree with this solution, because before the `Size` check, there is an `IsAbelian` check, and that can easily run into an infinite loop for infinite groups.

fingolfin:
Where "this solution" referred to "leave it to the Size check later on". But it seems you now did what I suggested anyway, so that should be fine.

fingolfin:
Could you please squash this into a single commit?

zickgraf:
Done!

PR #3334, opened 2019-03-11
Travis: re-enable OS X builds

fingolfin:
It seems OS X builds got better on Travis recently, so I figured we could give it another try. Would also be nice to get automated tests with clang back (but of course we can / should also do that with one of the Linux builds).


fingolfin:
This seems to work well, and reasonably fast, too.

PR #3337, opened 2019-03-12
kernel: generalize SyntaxTreeDefaultCompiler

fingolfin:
With this, most custom syntax tree "compiler" functions can be removed.

Also remove some redundant code in SyntaxTreeIf, and switch from SET_ELM_PLIST to PushPlist.

We plan to work on a "reverse compiler" next, and with the changes in here, that becomes simpler as well, as we plan to use the same data table for it as for the current syntax tree "compiler".

### DIFF
--- a/src/syntaxtree.c
+++ b/src/syntaxtree.c
@@ -65,6 +65,8 @@ static const CompilerT Compilers[];
     }
 #define ARG_(name) ARG(name, SyntaxTreeCompiler)
 
+#define ARGS(name) ARG(name, 0)
+
 static Obj SyntaxTreeFunc(Obj result, Obj func);
 
 static inline Obj NewSyntaxTreeNode(const char * type)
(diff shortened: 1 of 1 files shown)


src/syntaxtree.c
```diff
@@ -65,6 +65,9 @@ static const CompilerT Compilers[];
     }
 #define ARG_(name) ARG(name, SyntaxTreeCompiler)
 
+// use ARGS as last "arg" which collect all the following args together
```
ChrisJefferson: I've read this comment a couple of times, and I don't know what it means

fingolfin: Indeed, that was a leftover, I've now adapted it to the commenting style of the rest of the code in this file (i.e., I removed it). I plan to add some proper comments to this file, explaining how everything works, at a future point, now that @sebasguts and me worked it out (we think).

fingolfin: Main reason that I am not adding more comments right now, besides it being late, is that @sebasguts is working on a PR to add code that converts the output of the "SyntraxTree" functions back into GAP functions, and that code is based on this PR, so I'd prefer not to modify it to heavily, at least not until I have a chance to sync it with @sebasguts work.

PR #3342, opened 2019-03-18
kernel: simplify code which uses NAME_RNAM

fingolfin:
Since NAME_RNAM returns immutable strings in string rep, there is no
need for us to copy them around. We can just use the original strings.

### DIFF
--- a/src/precord.c
+++ b/src/precord.c
@@ -543,23 +543,16 @@ static Obj InnerRecNames(Obj rec)
 {
     Obj                 list;           /* list of record names, result    */
     UInt                rnam;           /* one name of record              */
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Not safe I'm afraid.

Various parts of the kernel assume the NAME_RNAMs are CSTRs, but if we give them out to users they could get PLAIN_LIST called on them, making them plists.

fingolfin:
@ChrisJefferson is of course right. (And it once again makes me wish I could travel back in time and stop this "silent conversion behind the scenes" madness, but that ship has sailed a long time ago, I am afraid).

PR #3343, opened 2019-03-18
kernel: add tests for syntaxtree code, fix some issues with it, some tweaks

fingolfin:


fingolfin:
The T_ATOMIC test failed in HPC-GAP, I forgot to update the test output, will fix that now.

Issue #3321, opened 2019-03-01
bug in ClassPositionsOfLowerCentralSeries


ThomasBreuer:
Frieder Ladisch had reported the following bug:
```
gap> g:= SmallGroup( 3^5, 22 );;
gap> LowerCentralSeries(g);
[ C27 : C9, Group([ f3, f5 ]), Group([ f5 ]), Group([ <identity> of ... ]) ]
gap> ct:= CharacterTable(g);;
gap> ClassPositionsOfLowerCentralSeries(ct);
[ [ 1 .. 35 ], [ 1, 4, 6, 12, 15 ] ]
```
Looking at the code, I found related issues:

- `ElementaryAbelianSeries` runs into an error if the argument is a nonsolvable group.
  According to the documentation, `fail` should be returned.
  I think the function should be adjusted to the documentation.
  (The corresponding function `ClassPositionsOfElementaryAbelianSeries` returns `fail`
  for character tables of nonsolvable groups.)
- `UpperCentralSeries` returns a *descending* series,
  whereas `ClassPositionsOfUpperCentralSeries` returns an *ascending* series.
  The documentation of `UpperCentralSeries` coincides with the return value.
  Here I am not sure what to do.
- In order to run a reasonable amount of tests (on small groups) that compare the
  `ClassPositionsOf...` functions with the corresponding functions for groups,
  it makes sense to improve the code of some `ClassPositionsOf...` methods.

I will provide a pull request.

fingolfin:
Fixed by #3347

PR #3347, opened 2019-03-19
fix a bug in ClassPositionsOfLowerCentralSeries

ThomasBreuer:
This pull request is intended to deal with issue #3321.

The proposed changes are as follows.

1. fixed errors in
   `ClassPositionsOfLowerCentralSeries`, `ClassPositionsOfSolvableResiduum`, `ClassPositionsOfSupersolvableResiduum`, which may lead to wrong results

2. avoid calling `ClassPositionsOfNormalSubgroups` in situations where this may lead to a huge list
   (the worst cases are large elementary abelian groups),
   in `IsSimpleCharacterTable`, `ClassPositionsOfMinimalNormalSubgroups`, `ClassPositionsOfFittingSubgroup`, `ClassPositionsOfSolvableRadical`, `ClassPositionsOfSolvableResiduum`, `ClassPositionsOfSupersolvableResiduum`, `IsPSolvableCharacterTableOp`

3. added extensive tests for the `ClassPositionsOf...` functions, which would not be reasonable without the changes under 2

4. fixed an `ElementaryAbelianSeries` method such that `fail` is returned in the case of nonsolvable groups, instead of calling `Error`; note that the documentation promises `fail` in this case

5. as a consequence of 4., `GroupSeriesByNiceMonomorphism` is now aware of `fail` results

### DIFF
--- a/lib/ctbl.gi
+++ b/lib/ctbl.gi
@@ -1263,10 +1263,19 @@ InstallMethod( IsPerfectCharacterTable,
 ##
 #M  IsSimpleCharacterTable( <tbl> ) . . . . . for an ordinary character table
 ##
+##  Avoid computing all normal subgroups of abelian groups and p-groups.
+##
 InstallMethod( IsSimpleCharacterTable,
     "for an ordinary character table",
     [ IsOrdinaryTable ],
-    tbl -> Length( ClassPositionsOfNormalSubgroups( tbl ) ) = 2 );
+    function( tbl )
+    if IsAbelian( tbl ) then
+      return IsPrimeInt( Size( tbl ) );
+    else
+      return not IsPrimePowerInt( Size( tbl ) ) and
+             Length( ClassPositionsOfNormalSubgroups( tbl ) ) = 2;
+    fi;
+    end );
 
 
 #############################################################################
@@ -1943,12 +1952,10 @@ InstallMethod( ClassPositionsOfMinimalNormalSubgroups,
           minimal,   # list of minimal kernels
           k;         # one kernel
 
-    # Every normal subgroup is an intersection of kernels of characters,
-    # so maximal normal subgroups are kernels of irreducible characters.
-    normal:= Set( ClassPositionsOfNormalSubgroups( tbl ) );
+    normal:= Set( List( [ 2 .. NrConjugacyClasses( tbl ) ],
+                        i -> ClassPositionsOfNormalClosure( tbl, [ i ] ) ) );
 
     # Remove non-minimal kernels
-    RemoveSet( normal, [ 1 ] );
     Sort( normal, function(x,y) return Length(x) < Length(y); end );
     minimal:= [];
     for k in normal do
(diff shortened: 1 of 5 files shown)


lib/ctbl.gi
```diff
@@ -1943,12 +1952,10 @@ InstallMethod( ClassPositionsOfMinimalNormalSubgroups,
           minimal,   # list of minimal kernels
           k;         # one kernel
 
-    # Every normal subgroup is an intersection of kernels of characters,
-    # so maximal normal subgroups are kernels of irreducible characters.
-    normal:= Set( ClassPositionsOfNormalSubgroups( tbl ) );
+    normal:= Set( List( [ 2 .. NrConjugacyClasses( tbl ) ],
```
fingolfin: Minor remark: `Set( List( l, f ) )` is equivalent to (but slower than) `Set( l, f )`

lib/ctbl.gi
```diff
@@ -2185,70 +2204,64 @@ InstallMethod( ClassPositionsOfSolvableRadical,
...
-    # Compute the structure constants $a_{x,\overline{x},g}$ with $g$ and $x$
-    # in $K_2$.
+    # Compute the structure constants $a_{g,x,x}$ with $g$ and $x$ in $K_2$.
     # Put them into a matrix, the rows indexed by $g$, the columns by $x$.
-    inv:= InverseClasses( tbl );
     mat:= List( K, x -> [] );
     for i in [ 2 .. Length( K ) ] do
       for j in K do
-        mat[i][j]:= ClassMultiplicationCoefficient( tbl, K[i], j, inv[j] );
+        mat[i,j]:= ClassMultiplicationCoefficient( tbl, K[i], j, j );
```
fingolfin: I have no idea whether this is correct or not, so I'll just trust that you know what you are doing :-).

lib/ctbl.gi
```diff
@@ -2185,70 +2204,64 @@ InstallMethod( ClassPositionsOfSolvableRadical,
...
       if Length( new ) = Length( K ) then
-        running:= false;
+        break;
       else
-        mat:= mat{ new };
-        K:= K{ new };
-        Add( series, new );
+        mat:= mat{ List( new, i -> Position( K, i ) ) };
+        K:= new;
+        Add( series, K );
```
fingolfin: Again, I'll trust that you know what you are doing.

lib/ctbl.gi
ThomasBreuer: @fingolfin Interesting, the two argument version of `Set` (resp. `SSortedList`) seems to be undocumented. Should we document it?

fingolfin: Ah, I was not aware that it is undocumented. Yes, I think it should be documented!

fingolfin:
Backported to stable-4.10 in commits 565e12fd4fb48091e65f16c9a3652462bbe4121d, 0ebdb142e58593aef1eddc71dc8389790f7e2b63, 55fba348c8185aa6ee5e58f5517cdfbb06ac8b59

PR #3350, opened 2019-03-19
Change PrintCSV to not use SizeScreen

### LINKED ISSUE #2545: Change PrintCSV to not use SizeScreen
Instead it should use something like `SetPrintFormattingStatus`.

For testing, I'd also consider changing `SizeScreen([4096,sz[2]]);` in it temporarily to, say `SizeScreen([40,sz[2]]);` to force an undesirable wrapping; with `SetPrintFormattingStatus` in effect, if all went right, this should be gone.

AnnaKDS:
- instead of using a filename (string) this functions now uses
output streams
- if the first argument is a filename then this function now creates
an output stream
- PrintFormattingStatus is set to false on the given output stream,
at the end it is set to true again

Fixes #2545

### DIFF
--- a/lib/string.gi
+++ b/lib/string.gi
@@ -943,9 +943,15 @@ local nohead,file,sep,f, line, fields, l, r, i,s,t,add,dir;
 end);
 
 InstallGlobalFunction(PrintCSV,function(arg)
-  local file,l,printEntry, rf, r, i, j,sz;
+  local stream,l,printEntry, rf, r, i, j, oldStreamFormattingStatus;
 
-  file:=arg[1];
+  if IsString(arg[1]) then
+    stream:=OutputTextFile(arg[1],false);
+  elif IsOutputStream(arg[1]) then
+    stream:=arg[1];
+  else
+    Error("PrintCSV: filename must be a string or an output stream");
+  fi;
   l:=arg[2];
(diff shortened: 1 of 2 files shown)


lib/string.gi
```diff
@@ -1016,29 +1021,29 @@ InstallGlobalFunction(PrintCSV,function(arg)
...
       fi;
       if IsBound(i.(rf[j])) then
         printEntry(i.(rf[j]));
       fi;
     od;
-    AppendTo(file,"\n");
+    AppendTo(stream,"\n");
   od;
-  SizeScreen(sz);
+  SetPrintFormattingStatus(stream,true);
```
fingolfin: It would be better to restore the previous value, instead of blindly reseting to `true` here. I.e. use `PrintFormattingStatus` at the start of the function to store the original value, then restore here to that.

AnnaKDS: Yes, that sounds like a good idea. I'll change that and add a test as you described. Thanks for the feedback!

fingolfin: Good! Could you add that example as a test into a .tst file? Feel free to ask if you need to know details.

AnnaKDS: I did it.

fingolfin:
Thank you!

PR #3351, opened 2019-03-19
kernel: refactor ON_KERNEL_ANTI_ACTION

fingolfin:
... to reduce code duplication.

PR best viewed with "ignore whitespace" diff setting enabled.

### DIFF
--- a/src/trans.cc
+++ b/src/trans.cc
@@ -1669,80 +1669,57 @@ static Obj FuncON_KERNEL_ANTI_ACTION(Obj self, Obj ker, Obj f, Obj n)
         return FuncFLAT_KERNEL_TRANS_INT(self, f, n);
     }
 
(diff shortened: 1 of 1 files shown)


fingolfin:
@james-d-mitchell @wilfwilson does this look OK to you? If you view the changes with "Hide whitespace changes" enabled, you should see that I essentially deleted "half the code", and moved some error handling to happen at the beginning.

fingolfin:
That said, I also noticed this comment before the function; so perhaps you'd prefer to rewrite this function anyway?
> // FIXME this should just always return a flat kernel of length <n>, the
> // special case should be removed, and [0] should be replaced by [1 .. n] in
> // the Semigroup package.

wilfwilson:
Sorry @fingolfin, this page has been an open tab on my computer all day. But I keep getting distracted. I'll attempt to look now :)

PR #3352, opened 2019-03-19
kernel/gac: don't hardcode TNUM values in generated C code

fingolfin:
In particular, output `T_INTPOS` instead of `1` (in theory, we also
output `T_INTNEG`, but in practice, GAP currently never generates
negative integer constants; but this might change in the future.


frankluebeck:
This is a reasonable change, creates more robust and more readable code.

PR #3353, opened 2019-03-19
kernel: refactor access to TmpTrans

fingolfin:
This also fixes a bug in FuncPOW_KER_PERM: the initial three slots of
TmpTrans, which are supposed to be bag references, were overwritten by
arbitrary data.

### DIFF
--- a/src/trans.cc
+++ b/src/trans.cc
@@ -149,11 +149,20 @@ static inline UInt DEG_TRANS(Obj f)
 static ModuleStateOffset TransStateOffset = -1;
 
 typedef struct {
+    // TmpTrans is essentially the same as TmpPerm
     Obj TmpTrans;
 } TransModuleState;
 
-// TmpTrans is the same as TmpPerm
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
This certainly looks like an improvement and should be merged. What actually is the fix? I couldn't see it.

src/trans.cc
```diff
@@ -3471,7 +3471,7 @@ static Obj FuncPOW_KER_PERM(Obj self, Obj ker, Obj p)
     SET_LEN_PLIST(out, len);
 
     ResizeTmpTrans(2 * len);
-    ptcnj = (UInt4 *)ADDR_OBJ(TmpTrans);
+    ptcnj = AddrTmpTrans();
```
fingolfin: This is the fix

fingolfin:
Still needs to be approved by a member of `gap-system`, e.g. @wilfwilson or @ChrisJefferson -- for some reason, @james-d-mitchell isn't (anymore) (although I'd be happy to change that again)

Issue #2727, opened 2018-08-22
Allow packages to use ISO 8601 dates in their PackageInfo.g


fingolfin:
It would be nice if packages could also specify the release date in ISO 8601 format, i.e., as `YYYY-MM-DD`. This is less ambiguous than the current `DD/MM/YYYY` (which e.g. Americans may confuse with `MM/DD/YYY`).

Besides needing to modify `lib/package.gi` and the documentation, it may also be necessary to modify the package distribution.

This might be a good "first issue"?

alex-konovalov:
Cf. also https://github.com/gap-system/gap-distribution/issues/20. 

I think we can implement this in a way that both formats will be permitted concurrently, and that `ValidatePackageInfo` will display an advice to switch to ISO 8601 (maybe at some info level, otherwise this will show up too often if #2726 is merged). 

Indeed, not a problem to detect whether the date is in `YYYY-MM-DD` or `DD/MM/YYYY` format.

cdwensley:
Date := "28/08/2018"  is converted by GAPDoc to "28 August 2018" in a manual front page, whereas 
Date := "2018-08-28" remains unchanged, which is not as nice.  Perhaps @frankluebeck can comment?

fingolfin:
Indeed, AFAIK does nothing with title pages. But we can certainly add support to ISO dates in GAPDoc, too. Alternatively, we could simply convert the date from ISO 8601 to the "old" date format on the fly while loading, thus preserving full compatibility with any code that might for whatever reason try to parse these dates.

wucas:
I'm working on a PR to change `ValidatePackageInfo` so it accepts both formats.

PR #3355, opened 2019-03-19
Allow packages to use ISO 8601 dates in their PackageInfo.g

wucas:
If the date format is dd/mm/yyyy and InfoLevel InfoPackageLoading is set to at least 2 it prints a hint to change the format as was suggested in #2727. This pull request is intended to fix issue #2727.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -213,7 +213,7 @@ end );
 #F  AddPackageInfo( files )
 ##
 BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
-    local file, record, pkgname, version;
+    local file, record, pkgname, version, date, dd, mm;
     for file in files do
       # Read the `PackageInfo.g' file.
       Unbind( GAPInfo.PackageInfoCurrent );
@@ -241,6 +241,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
           elif IsRecord( record.PackageDoc ) then
             record.PackageDoc:= [ record.PackageDoc ];
           fi;
+
+          # Normalize the format of 'Date', i.e. if it is the format yyyy-mm-dd
+          # then we change it to dd/mm/yyyy. When other tools have adapted to
+          # the yyyy-mm-dd format we can normalize to that format and at some
+          # point in the future get rid of this code.
+          if record.Date{[5,8]} = "--" then
+            date := List( SplitString( record.Date, "-" ), Int);
+            date := Permuted(date, (1,3)); # date = [dd,mm,yyyy]
+            # generate the day and month strings
+            # if the day has only one digit we have to add a 0
+            if date[1] < 10 then
+              dd := Concatenation("0", String(date[1]));
+            else
+              dd := String(date[1]);
+            fi;
+            # if the month has only one digit we have to add a 0
+            if date[2] < 10 then
+              mm := Concatenation("0", String(date[2]));
+            else
+              mm := String(date[2]);
+            fi;
+            record.Date := Concatenation(dd, "/", mm, "/", String(date[3]));
+          fi;
+
           if IsHPCGAP then
             # FIXME: we make the package info record immutable, to
             # allow access from multiple threads; but that in turn
@@ -2130,7 +2154,7 @@ InstallGlobalFunction( DeclareAutoreadableVariables,
 InstallGlobalFunction( ValidatePackageInfo, function( info )
     local record, pkgdir, i, IsStringList, IsRecordList, IsProperBool, IsURL,
           IsFilename, IsFilenameList, result, TestOption, TestMandat, subrec,
-          list;
+          list, CheckDateValidity;
 
     if IsString( info ) then
       if IsReadableFile( info ) then
@@ -2197,10 +2221,34 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
     TestMandat( record, "Version",
         x -> IsString(x) and 0 < Length(x) and x[1] <> '=',
         "a nonempty string that does not start with `='" );
-    TestMandat( record, "Date",
-        x -> IsString(x) and Length(x) = 10 and x{ [3,6] } = "//"
-                 and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ),
-        "a string of the form `dd/mm/yyyy'" );
+
+    # check if the date is valid
+    CheckDateValidity := function(x)
+      local date;
+
+      if not ( IsString(x) and Length(x) = 10
+        and ( ( ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar )
+        and  x{ [3,6] } = "//" ) or ( ForAll( x{ [1,2,3,4,6,7,9,10] }, IsDigitChar ) and x{ [5,8] } = "--" ) ) ) then
+        return false;
+      elif x{ [3,6] } = "//" then # for the old format split at '/'
+        date := List( SplitString( x, "/" ), Int);
+      elif x{ [5,8] } = "--" then # for the yyyy-mm-dd format split at '-'
+        date := List( SplitString( x, "-" ), Int);
+        date := date{ [3,2,1] }; # sort such that date=[dd,mm,yyyy]
+      fi;
+      return date[2] in [1..12] and date[3] >= 1999 # GAP 4 appeared in 1999
+          and date[1] in [1..DaysInMonth( date[2], date[3] )];
+    end;
+
+    TestMandat( record, "Date", CheckDateValidity, Concatenation( "a string of the form yyyy-mm-dd or dd/mm/yyyy",
+    " that represents a date since 1999") );
+
+    # If the date is in the format `dd/mm/yyyy` a message is printed
+    # code to be used after an adaption period for packages
+    #if IsBound( record.Date ) and  CheckDateValidity( record.Date ) and record.Date{ [3,6] } = "//" then
+    #   Info( InfoPackageLoading, 2, Concatenation( record.PackageName, ": Please be advised to change the date format to `yyyy-mm-dd`") );
+    #fi;
+
     TestOption( record, "License",
         x -> IsString(x) and 0 < Length(x),
         "a nonempty string containing an SPDX ID" );
--- a/tst/testinstall/package.tst
+++ b/tst/testinstall/package.tst
@@ -202,7 +202,8 @@ gap> ValidatePackageInfo(rec());
 #E  component `Subtitle' must be bound to a string
 #E  component `Version' must be bound to a nonempty string that does not start\
  with `='
-#E  component `Date' must be bound to a string of the form `dd/mm/yyyy'
+#E  component `Date' must be bound to a string of the form yyyy-mm-dd or dd/mm\
+/yyyy that represents a date since 1999
 #E  component `ArchiveURL' must be bound to a string started with http://, htt\
 ps:// or ftp://
 #E  component `ArchiveFormats' must be bound to a string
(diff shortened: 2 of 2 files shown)


lib/package.gi
```diff
@@ -2186,9 +2186,16 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
         x -> IsString(x) and 0 < Length(x) and x[1] <> '=',
         "a nonempty string that does not start with `='" );
     TestMandat( record, "Date",
-        x -> IsString(x) and Length(x) = 10 and x{ [3,6] } = "//"
-                 and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ),
-        "a string of the form `dd/mm/yyyy'" );
+        x -> IsString(x) and Length(x) = 10
+          and ( ( x{ [3,6] } = "//" and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ) )
+          or ( x{ [5,8] } = "--" and ForAll( x{ [1,2,3,4,6,7,9,10] }, IsDigitChar ) ) ) ,
+        "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy`." );
```
alex-konovalov: Here and in the next message do not use `.` at the end to keep consistent style.

lib/package.gi
```diff
@@ -2186,9 +2186,16 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+
+    #If the date is in the format `dd/mm/yyy` a message is printed
```
alex-konovalov: one `y` is missing in `yyy`

fingolfin:
For some reason, there is a conflict in `lib/package.gi` -- which is a bit odd, that file was last changed in February... @wucas it seems you started your work from an old version of the GAP master branch (December 2018). I recommend that you first update the master branch of your GAP clone to the latest, then rebase you work. I'll be happy to assist with the latter, if desired.

wucas:
@alex-konovalov I submitted a pull request to change this in gap-distribution too. Also `ValidatePackageInfo` validates the date now.

alex-konovalov:
Thanks @wucas - I will look. A cross-reference to that PR from here would be useful though. Saves time when navigating and connects things belonging together.

wucas:
The test in `tst/testinstall/package.tst` failed because the validity check of the date outputs something if the date is invalid and because the output of the format check also has a different output now. I changed this in the test file. The tests run on my computer now.

lib/package.gi
```diff
@@ -2203,9 +2203,40 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+          or ( x{ [5,8] } = "--" and ForAll( x{ [1,2,3,4,6,7,9,10] }, IsDigitChar ) ) ) ,
+        "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy`" );
+
+    #If the date is in the format `dd/mm/yyyy` a message is printed
+    if IsBound( record.Date ) and  record.Date{ [3,6] } = "//" then
+       Info( InfoPackageLoading, 2, Concatenation(record.PackageName, ": Please be adviced to change the date format to `yyyy-mm-dd`"));
+    fi;
+
+    #check if the date is valid
+    CheckDateValidity := function(x)
```
alex-konovalov: Wouldn't it be better to parse the date into components `dd`, `mm` and `yyyy`, and then do a single check instead of duplicating those lines dependently on the list layout?

wucas: Yes that's better. I'd suggest the following which only makes the distinction for the split
```suggestion
    #check if the date is valid
    CheckDateValidity := function(x)
      local date;
      if x{ [3,6] } = "//" then #for the old format split at '/'
        date := List( SplitString( x, "/" ), Int);
      elif x{ [5,8] } = "--" then #for the yyyy-mm-dd format split at '-'
        date := List( SplitString( date, "-" ), Int);
        Permuted(date, (1,3)); #sort such that date=[dd,mm,yyyy]
      fi;
      if IsBound(date) and (not (date[1] in [1..31] and
                   date[2] in [1..12] and
                   date[3] > 1999 and # GAP 4 appeared in 1999
                   date[1] <= DaysInMonth( date[2], date[3] ))) then
              return false;
      fi;
    return true;
    end;
```

wucas: If that way is okay, I'll also change this in the PR  [#78](https://github.com/gap-system/gap-distribution/pull/78) in gap-distribution.

fingolfin: looks sensible to me

fingolfin: you might want to add an `else Error("date must be in form yyyy-mm-dd or dd/mm/yyyy");` or so

wucas: As far as I understand it this is checked a couple of lines above:
```
TestMandat( record, "Date",
        x -> IsString(x) and Length(x) = 10
          and ( ( x{ [3,6] } = "//" and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ) )
          or ( x{ [5,8] } = "--" and ForAll( x{ [1,2,3,4,6,7,9,10] }, IsDigitChar ) ) ) ,
        "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy`" );
```

alex-konovalov: Yes, but what if the test above passed, so no `yyyy-mm-dd` or `dd/mm/yyyy` suggestion was displayed?

The code "which only makes the distinction for the split" looks good to me, please update the PR.

wucas:
@alex-konovalov you're right I overlooked the issue with the ambiguity of the dd/mm/yyyy format. I added a Mandat in the CheckDateValidity function. That way seemed most easy to me and I think with the comment it is clear what the code is supposed to do. I return true if that's the case so the "invalid Date" message is not printed. 
To make sure that this is caught in the future I added a test for this.

alex-konovalov:
It just occurred to me that this PR will not fully resolve the problem. In fact, it could even make it more difficult. If `PackageInfo.g` files with dates in both formats will pass validation, we will eventually have a mix of two different  formats in `GAPInfo.PackagesInfo`, and any tool that accesses dates from these record for some purposes (report, generated content etc). will have to adapt to handle two formats (for example, the "healthcheck" report, see https://github.com/gap-system/gap-distribution/issues/49).

So, I suggest that in addition to adjusting validation we should also change `SetPackageInfo` so that when loading package metadata it will convert all dates to ISO 8601, and will possibly accept `dd/mm/yyyy` dates forever for backwards compatibility.

fingolfin:
Yes, please make sure that after loading, dates are in a uniform format. I'd indeed prefer ISO 8601 for that *in the long run*, but for the time being, we must make sure no tooling is broken. Right now, that might affect [AutoDoc](https://github.com/gap-packages/AutoDoc), which parses the Date for package manuals ([see here](https://github.com/gap-packages/AutoDoc/blob/fcf9e7f6416bea7ada4648f87899875577bc3423/gap/AutoDocMainFunction.gi#L283) and also [the code for `AUTODOC_FormatDate`](https://github.com/gap-packages/AutoDoc/blob/fcf9e7f6416bea7ada4648f87899875577bc3423/gap/ToolFunctions.gi#L322)).

So perhaps it'd be better to "normalize" everything to the dd/mm/yyyy format; it'll then be trivial to switch over to the ISO format at any time, and also buys us time to adapt AutoDoc and whatever else might need it. And in case there is a problem, it is trivial to roll back, too.

alex-konovalov:
@fingolfin wrote

> So perhaps it'd be better to "normalize" everything to the dd/mm/yyyy format;

Ok, I thought initially about more radical change and switch to ISO 8601 _now_, but OK, let's do it gradually - that will allow to make a small adjustment and merge this PR then.

wucas:
@fingolfin I looked at this again since I'd like to finish this up during the GAP Days next week. 
So the date is stored in `date` which is a list of the form `[dd, mm, yyyy]` so isn't it "normalized"? Of course this is not the date as it is in the `PackageInfo.g` so should I normalize it there? And if so how can I change this file? Or am I totally misunderstanding the situation?

alex-konovalov:
@wucas unless I miss something, I see only

    Permuted(date, (1,3)); #sort such that date=[dd,mm,yyyy]

but then `date` is discarded. If you will inspect `GAPInfo.PackagesInfo`, it will still have a mix of formats - it's there where we would like to use a single format.

wucas:
I have added a normalization routine. As far as I can't tell now GAPInfo.PackagesInfo.pkgname[1].Date has the format "dd/mm/yyyy" even if in the respective PackageInfo.g the date has the format "yyyy-mm-dd". @alex-konovalov does this address @fingolfin s request?

lib/package.gi
```diff
@@ -2203,9 +2227,43 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
-                 and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ),
-        "a string of the form `dd/mm/yyyy'" );
+        x -> IsString(x) and Length(x) = 10
+          and ( ( x{ [3,6] } = "//" and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar ) )
+          or ( x{ [5,8] } = "--" and ForAll( x{ [1,2,3,4,6,7,9,10] }, IsDigitChar ) ) ) ,
+        "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy`" );
+
+    #If the date is in the format `dd/mm/yyyy` a message is printed
+    if IsBound( record.Date ) and  record.Date{ [3,6] } = "//" then
+       Info( InfoPackageLoading, 2, Concatenation(record.PackageName, ": Please be adviced to change the date format to `yyyy-mm-dd`"));
```
wilfwilson:
```suggestion
       Info( InfoPackageLoading, 2, Concatenation(record.PackageName, ": Please be advised to change the date format to `yyyy-mm-dd`"));
```

wilfwilson:
@wucas In future, please you could make PRs by pushing the branch to [your own fork of gap](https://github.com/wucas/gap), and then making the PR from your fork into this fork?

As things are, you're pushing your branch directly toe `gap-packages/gap`, which means every time you push, the tests are being run twice on Travis (which takes ages): once because a new push has been made to a branch on `gap-packages/gap`, and once because a new push has been made to a PR.

lib/package.gi
```diff
@@ -246,6 +246,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
             elif IsRecord( record.PackageDoc ) then
               record.PackageDoc:= [ record.PackageDoc ];
             fi;
+
+            #Normalize the format of 'Date', i.e. if it is the format yyyy-mm-dd
```
wilfwilson: Could you please put a space between the `#` character and the words in this PR, including in this line? I find it hard to read like this.

lib/package.gi
```diff
@@ -2203,9 +2227,43 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+      if IsBound(date) and (not (date[1] in [1..31] and
+                   date[2] in [1..12] and
+                   date[3] > 1999 and # GAP 4 appeared in 1999
+                   date[1] <= DaysInMonth( date[2], date[3] ))) then
+              return false;
+      fi;
+      return true;
+    end;
+    TestMandat( record, "Date", 
+      x -> IsString(x) and CheckDateValidity(x), "a valid date");
```
wilfwilson: I think you should show more information in the warning:
```suggestion
      x -> IsString(x) and CheckDateValidity(x), "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy` representing a date since 1999");
```

lib/package.gi
```diff
@@ -2203,9 +2227,43 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+            x -> false, "a string of the form `yyyy-mm-dd` or `dd/mm/yyyy`" );
+          return true;
+        fi;
+      elif x{ [5,8] } = "--" then #for the yyyy-mm-dd format split at '-'
+        date := List( SplitString( x, "-" ), Int);
+        date := Permuted(date, (1,3)); #sort such that date=[dd,mm,yyyy]
+      fi;
+      if IsBound(date) and (not (date[1] in [1..31] and
+                   date[2] in [1..12] and
+                   date[3] > 1999 and # GAP 4 appeared in 1999
```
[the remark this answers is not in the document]
wucas: Yes absolutely overlooked that. Thanks!

wucas:
@wilfwilson this is now essentially the code you suggested. I'm quite happy with how things look now. Thanks! If things stabilize here I will incorporate your suggestions to #78 in gap-distribution.

tst/testinstall/package.tst
```diff
@@ -202,7 +202,9 @@ gap> ValidatePackageInfo(rec());
 #E  component `Subtitle' must be bound to a string
 #E  component `Version' must be bound to a nonempty string that does not start\
  with `='
-#E  component `Date' must be bound to a string of the form `dd/mm/yyyy'
+#E  component `Date' must be bound to a string of the form `yyyy-mm-dd` or `dd\
+/mm/yyyy`
+#E  component `Date' must be bound to a valid date
```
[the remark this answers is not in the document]
wilfwilson: @alex-konovalov This is sorted out now.

lib/package.gi
```diff
@@ -2202,10 +2226,33 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+
+      if not ( IsString(x) and Length(x) = 10
+        and ForAll( x{ [1,2,4,5,7,8,9,10] }, IsDigitChar )
+        and ( x{ [3,6] } = "//" or x{ [5,8] } = "--" ) ) then
+        return false;
+      elif x{ [3,6] } = "//" then # for the old format split at '/'
+        date := List( SplitString( x, "/" ), Int);
+      elif x{ [5,8] } = "--" then # for the yyyy-mm-dd format split at '-'
+        date := List( SplitString( x, "-" ), Int);
+        date := Permuted( date, (1,3) ); # sort such that date=[dd,mm,yyyy]
```
fingolfin: I'd rather do it like this:
```suggestion
        date := date{[3,2,1]};
```

lib/package.gi
```diff
@@ -2202,10 +2226,33 @@ InstallGlobalFunction( ValidatePackageInfo, function( info )
...
+      return date[2] in [1..12] and date[3] >= 1999 # GAP 4 appeared in 1999
+          and date[1] in [1..DaysInMonth( date[2], date[3] )];
+    end;
+
+    TestMandat( record, "Date", CheckDateValidity, Concatenation( "a string of the form yyyy-mm-dd or dd/mm/yyyy",
+    " that represents a date since 1999") );
+
+    # If the date is in the format `dd/mm/yyyy` a message is printed
+    if IsBound( record.Date ) and  CheckDateValidity( record.Date ) and record.Date{ [3,6] } = "//" then
+       Info( InfoPackageLoading, 2, Concatenation( record.PackageName, ": Please be advised to change the date format to `yyyy-mm-dd`") );
```
fingolfin: Do we really want to print this right now? Perhaps better to comment that code out now, at least until we are sure all relevant tools have been updated. Things to check include:
- [ ] the package distribution
- [x] AutoDoc (already updated by me for ISO dates)
- [ ] GitHubPagesForGAP -- the `update.g` script parses the date (I *think* I already prepared that for ISO dates some time ago, but (a) this is untested and (b) some packages might still be using an older version of the `update.g` script

tst/testinstall/package.tst
```diff
@@ -222,6 +223,35 @@ gap> info := rec(
 >     PackageName := "pkg",
 >     Subtitle := "desc",
 >     Version := "0",
+>     Date := "01/20/2015",
```
fingolfin: Shouldn't there also be a test with an ISO date?

alex-konovalov:
This would be good to have in 4.11.0 - added it to the milestone to ensure that its stays on the radar.

lib/package.gi
wucas: I've made a PR for the package distro although I don't know if that is all which has to be changed. 
I can comment it out, but I already have it only for InfoLevel 2 so that it doesn't get printed all the time or am I misunderstanding the concept here?

tst/testinstall/package.tst
wucas: You're absolutely right especially since there is something wrong that I only spotted because I did the test. Sorry should have already done that!

lib/package.gi
fingolfin: Are you referring to https://github.com/gap-system/gap-distribution/pull/78 ? I commented on that last week :-)

DominikBernhardt:
This was added to the 4.11 Milestone and @alex-konovalov requested changes. From what I understand @wucas changed code according to @wilfwilson requests. What is the status of this?

lib/package.gi
wucas: Ah sorry I wasn't notified by GitHub that something happened there. I'll take care of it thanks! (also GitHub didn't tell me that you made the comment above telling me about the comment in the other repo. Sometimes I'm a bit confused about what it will notify me of and what not)

alex-konovalov:
This waits for me to review in conjunction with https://github.com/gap-system/gap-distribution/pull/78

alex-konovalov:
Unfortunately, moving this to GAP 4.12 milestone.

fingolfin:
It's rather frustrating that @wucas opened this comparatively simple change request (which is by no means to belittle the effort it took to get there) over a year ago and it is still not merged, and now has a conflict, so a rebase is necessary. I don't understand why we can't merge this right away (or could have done so in September) -- adding the ability to GAP to load packages with ISO dates before teaching the website and package distribution about ISO dates seems perfectly reasonable to me, esp. as long as we don't accept such packages for distribution.

I will attempt to rebase this now, and I really hope we can then merge it ASAP. And perhaps next week the website counterpart can be taken care of. But even if not, I think we should just merge this!

fingolfin:
Just realized the PR is not over a year, but *exactly* one year old. Happy birthday then ;-). And I managed to rebase it, too.

fingolfin:
@alex-konovalov can we merge this now?

lib/package.gi
```diff
@@ -241,6 +241,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
...
+            date := List( SplitString( record.Date, "-" ), Int);
+            date := Permuted(date, (1,3)); # date = [dd,mm,yyyy]
+            # generate the day and month strings
+            # if the day has only one digit we have to add a 0
+            if date[1] < 10 then
+              dd := Concatenation("0", String(date[1]));
+            else
+              dd := String(date[1]);
+            fi;
+            # if the month has only one digit we have to add a 0
```
alex-konovalov, line 259: If month has only one digit, this will never be caught by `record.Date{ [5,8]} = "--" ` condition, so a date line `2020-4-21` will not be changed to `21/04/2020` at all.

lib/package.gi
```diff
@@ -213,7 +213,7 @@ end );
 #F  AddPackageInfo( files )
 ##
 BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
-    local file, record, pkgname, version;
+    local file, record, pkgname, version, date, dd, mm;
```
alex-konovalov, line 216: So, we call `ValidatePackageInfo` from `LoadPackage` but not from `AddPackageInfos`. I wonder if we should do that in `AddPackageInfos` too...

lib/package.gi
ChrisJefferson: Couldn't be just do `'-' in record.Date`?

lib/package.gi
```diff
@@ -241,6 +241,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
           elif IsRecord( record.PackageDoc ) then
             record.PackageDoc:= [ record.PackageDoc ];
           fi;
+
+          # Normalize the format of 'Date', i.e. if it is the format yyyy-mm-dd
+          # then we change it to dd/mm/yyyy. When other tools have adapted to
+          # the yyyy-mm-dd format we can normalize to that format and at some
+          # point in the future get rid of this code.
+          if record.Date{ [5,8]} = "--" then
```
fingolfin, line 249:
```suggestion
          if record.Date{[5,8]} = "--" then
```

lib/package.gi
fingolfin: A date like `2020-4-21` is not a valid ISO date and thus should be rejected, as this PR does.

We still need this code here because above, we convert the parts of the day into integers. A simple implementation of course would be to not do that, and just keep the strings as they are. The calls to `Int` of course validate that we are looking at numbers; but a better check for that might be to change `if record.Date{ [5,8]} = "--" then` by adding something like `... and ForAll(record.Date{1,2,3,4,6,7,9,10}, IsDigit)` . But this starts to feel a bit like bikeshedding ;-)

lib/package.gi
fingolfin: Perhaps, but that goes beyond the purpose of this PR, doesn't it?

fingolfin:
I fixed one minor whitespace typo and rebased this PR to see whether the tests pass. I also dismissed @alex-konovalov "request for changes" review, as I believe I refuted his "bug report" and consider his "concern" a feature request that goes beyond this PR.

If I hear no new concerns about this PR, I'll merge it next week.

lib/package.gi
ChrisJefferson: My worry was if someone tried to use the new date format, and messed it up, GAP would get confused by something like 2020-4-21 and try to interpret it as an old format date.

However, on inspection ValidatePackageInfo is checking properly, and we already let any old rubbish in the Date field.

alex-konovalov:
Well, to do this properly, one should check:
- this PR
- instructions for date in https://github.com/gap-packages/example/blob/master/PackageInfo.g
- https://github.com/gap-system/gap-distribution/pull/78
- https://github.com/gap-system/GapWWW (scripts that generate package-related pages).

As there is no pressing needs for this (except that this is one year old, and it would be nice to merge it), I don't plan to look at these before the middle of the next week...

alex-konovalov (approved):

alex-konovalov:
@wucas @fingolfin thanks & let's do it.

PR #3356, opened 2019-03-19
MatrixObj: remove Fold / Unfold

fingolfin:
These operations are not really needed or used; and the name "fold"
already has another [well-established meaning in functional programming](https://en.wikipedia.org/wiki/Fold_(higher-order_function)),
and its use here is very confusing for me.

Indeed, I'd like to add a `Fold` that does what "usually" is expected under that name.

Since this was never documented, there should be no issue removing it.

### DIFF
--- a/hpcgap/lib/vec8bit.gi
+++ b/hpcgap/lib/vec8bit.gi
@@ -1165,20 +1165,6 @@ InstallOtherMethod( KroneckerProduct, "for two 8bit matrices",
     return kroneckerproduct;
   end );
 
-InstallMethod( Fold, "for an 8bit vector, a positive int, and an 8bit matrix",
-  [ IsVectorObj and Is8BitVectorRep, IsPosInt, Is8BitMatrixRep ],
-  function( v, rl, t )
-    local rows,i,tt,m;
-    m := [];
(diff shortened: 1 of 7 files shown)


fingolfin:
Fixed HPC-GAP error. There are still errors in the package tests, though, as cvec installs a Fold method. I'll take care of that.

frankluebeck:
I agree with removing `Fold` and `Unfold`. I guess that the idea for them came from matrix input methods in other systems of the kind `Matrix(ring, n, m, entries)`, where entries is a list or vector of entries of length n x m. We could still support that syntax and maybe use `Concatenation` for the converse, so that we don't need extra operations.

ThomasBreuer:
@frankluebeck `Concatenation` is not an operation.

fingolfin:
I agree that it is useful to have a way to convert a VectorObj into a matrix (typically a n x 1 or 1 x n matrix, but I am also fine with converting it to an arbitrary shape). In addition, of course constructors of MatrixObjs may take flat lists of coefficients, that's fine. In the end, it's really only the names I am opposed to (although I'd prefer if we had concrete applications for the code in mind before adding it back, but that's not a hill I plan to die on either ;-)

PR #3357, opened 2019-03-19
Add `FoldLeft`, `FoldLeftX`; and `ForAllX`, `ForAnyX`, `FilteredX`, `NumberX`, `PerformX` to complement `ListX`, `SetX`, `SumX`, `ProductX`

fingolfin:
This PR adds a new operation `FoldLeft` and a more general helper `FoldLeftX`,
then uses `FoldLeftX` to simplify the implementation of the existing
`ListX`, `SumX`, and `ProductX` functions and to add several further
`X`-style functions:
`ForAllX`, `ForAnyX`, `FilteredX`, `NumberX`, and `PerformX`.

`FoldLeft` is implemented as an operation with early methods for built-in
lists and regular methods for general collections. In particular, lists with
holes are handled consistently by iterating past holes, just as in ordinary
`for`-loops.

`FoldLeftX` takes a list of generators and filters describing nested loops and
conditions, together with an accumulator update function and an initial value.
It also supports an optional stopping value for efficient short-circuiting,
which is used by `ForAllX` and `ForAnyX`.

One idea for a future follow-up would be to broaden the scope and allow
`ListX` and related functions to accept iterators directly. But this PR
has already been here since 2019 (and the underlying code even longer),
so I'd rather not tackle this now.

AI disclosure: I used Codex for reviewing the branch, implementing the final
polish, extending tests, and completing the documentation.


ChrisJefferson:
I think this looks very reasonable.

For the shortcutting, I'm tempted to suggest an optional single argument which is a object which is compared using IsIdenticalObj. My reasoning is:

1) Nice and efficient.
2) The values I expect most commonly to be tested are 0, true, false and fail, where this will work.
3) If we make an otherwise unused bag, we can use that value when nothing was given by the user (I thought about using null, but that might mean we short-circut early when given a function which doesn't return, which might be surprising).

fingolfin:
@ChrisJefferson I like that idea! We could simply pass `""` instead of a special "unused bag", as that would be a completely fresh object guaranteed to be non-identical to any other (of course that's a slight inefficiency, too, but I'd imagine not measurable in practice).

wilfwilson:
I like this project.

lib/coll.gd
```diff
@@ -2400,6 +2400,38 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
 DeclareGlobalFunction( "Elements" );
 
 
+#############################################################################
+##
+##  TODO: document FoldLeft
```
fingolfin: TODO: add documentation

lib/coll.gd
```diff
@@ -2400,6 +2400,38 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+##
+DeclareGlobalFunction( "FoldLeft" );
+
+
+#############################################################################
+##
+##  TODO: document FoldLeftOp
+##
+DeclareOperation( "FoldLeftOp", [ IsListOrCollection, IsFunction ] );
```
fingolfin: TODO: should there also be a declaration for `FoldLeftOp` with three arguments, the third being the initial value?

lib/coll.gi
```diff
@@ -1675,497 +1675,233 @@ InstallMethod( ForAnyOp,
...
-                SetXHelp( result, gens, i+1, vals, 3 );
-            od;
-            Unbind( vals[3] );
-            return;
+InstallGlobalFunction( FoldLeft,
+    function( arg )
+    local tnum, C, func, result, i, l;
+    l := Length( arg );
+    if l < 2 or l > 3 or not IsFunction(arg[2]) then
+      Error( "usage: FoldLeft( <C>, <func>[, <init>] )" );
```
fingolfin: TODO: should `FoldLeft` also get an abort mechanism, like `FoldLeftX`? How should it work? The one for `FoldLeftX` is a bit adhoc: it works great for the few applications we have, but it's not very general...

What do other languages do for this, if they do anything at all?

lib/coll.gd
```diff
@@ -2400,6 +2400,38 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+
+#############################################################################
+##
+##  TODO: document the following
+##
+DeclareGlobalFunction( "ForAllX" );
+DeclareGlobalFunction( "ForAnyX" );
+DeclareGlobalFunction( "FilteredX" );
+DeclareGlobalFunction( "NumberX" );
+DeclareGlobalFunction( "PerformX" );
```
fingolfin: TODO: also add `FirstX` and `LastX`?

lib/coll.gi
```diff
@@ -1675,497 +1675,233 @@ InstallMethod( ForAnyOp,
...
-    return result;
+
+#############################################################################
+##
+#M  ForAnyX(<obj>,...)
+##
+InstallGlobalFunction( ForAnyX, function ( arg )
+    local f;
+    f := Remove(arg);
+    # TODO: implement early abort
```
fingolfin:
```suggestion
```

src/listfunc.c
```diff
@@ -1524,6 +1527,112 @@ static Obj FuncLIST_WITH_IDENTICAL_ENTRIES(Obj self, Obj n, Obj obj)
...
+                         Obj * acc,
+                         Obj   abortValue,
+                         Obj   args,
+                         int   genIndex,
+                         int   valIndex)
+{
+    while (genIndex <= LEN_PLIST(gens)) {
+        Obj gen = ELM_PLIST(gens, genIndex);
+        if (IS_FUNC(gen))
+            gen = FastCallFuncList(gen, args);
```
fingolfin: TODO: replace `FastCallFuncList ` by just `CallFuncList`? It was kind of an attempt to bypass call overhead, but now I wonder if that wasn't a premature optimization. Doing some timing measurements probably would be a good idea: comparing performance with ` `FastCallFuncList ` versus `CallFuncList`, but also that of `ListX`, `SumX`, `ProductX` before and after this PR.

src/listfunc.c
```diff
@@ -1460,6 +1463,121 @@ static Obj FuncLIST_WITH_IDENTICAL_ENTRIES(Obj self, Obj n, Obj obj)
...
+                    continue; // skip holes
+                AssPlist(args, valIndex, elm);
+                if (FoldLeftXHelp(gens, foldFunc, acc, abortValue, args,
+                                  genIndex + 1, valIndex + 1))
+                    return 1;
+            }
+            UNB_LIST(args, valIndex);
+            return 0;
+        }
+        else if (CALL_1ARGS(IsListOrCollection, gen) == True) {
```
fingolfin, line 1527: actually here is a chance to generalize `ListX` and friends to also accept iterators: we just need to accept objects in the filter `IsIterator` here, that should do it. Note that `List` and friends already (mostly?) support iterators.

lib/coll.gd
fingolfin: The "problem" with `FirstX` is the same as for `ForAnyX` and `ForAllX`: we want to be able to short circuit and abort the search as soon as we know the answer.  The mechanism for doing this in this PR is the ugliest part (IMHO) of this PR by far (there is an optional `abortValue` that one can specify, and for which we check by identity comparison... once encountered, the search aborts. Works, is simply, but... ugly?

lib/coll.gd
```diff
@@ -2416,6 +2430,78 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+##  </Description>
+##  </ManSection>
+##
+DeclareOperation( "FoldLeftOp", [ IsListOrCollection, IsFunction ] );
+DeclareOperation( "FoldLeftOp", [ IsListOrCollection, IsFunction, IsObject ] );
+
+
+#############################################################################
+##
+#O  FoldLeftX( <gens>, <func>, <init>[, <abortValue>] )
```
fingolfin, line 2477: TODO: replace `abortValue` by `stopValue` or `stoppingValue`?

lib/coll.gi
```diff
@@ -1682,496 +1682,230 @@ InstallMethod( ForAnyOp,
...
-            for val2  in gen  do
-                ListXHelp2( result, gens, i+1, val1, val2 );
-            od;
-            return;
+#############################################################################
+##
+#F  FoldLeft( <coll>, <func> )
+#F  FoldLeft( <coll>, <func>, <init> )
+##
+InstallGlobalFunction( FoldLeft,
```
fingolfin, line 1701: Should be redone using `InstallEarlyMethod`

lib/coll.gi
```diff
@@ -1682,496 +1682,230 @@ InstallMethod( ForAnyOp,
...
+      C:= arg[1];
+      func:= arg[2];
+      if l = 2 then
+        if IsEmpty( C ) then
+          Error("folding an empty collection without initial value is not supported");
         else
-            Error( "gens[",i+1,"] must be a collection, a list, a boolean, ",
-                   "or a function" );
+          result:= C[1];
+          for i in [ 2 .. Length( C ) ] do
```
fingolfin, line 1718: Contrasting this with the code path used when `init` is given (which uses `for i in C do`) I think we get different behavior if `C` has holes. We probably don't want that difference, do we?

src/listfunc.c
fingolfin: I don't want to do that in this PR, it has already been stalled for so many years.

Might be a nice future project, though.

lib/coll.gd
```diff
@@ -2478,6 +2478,120 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+##
+##  <#GAPDoc Label="FoldLeftX">
+##  <ManSection>
+##  <Func Name="FoldLeftX" Arg='gens, func, init[, abortValue]'/>
+##
+##  <Description>
+##  <Ref Func="FoldLeftX"/> is a generalization of <Ref Func="FoldLeft"/>
```
ThomasBreuer:
```suggestion
##  <Ref Func="FoldLeftX"/> is a generalization of <Ref Oper="FoldLeft"/>
```
This should fix the test failure.

(`doc/ref/testconsistency.g` defines some additional rules about the GAPDoc syntax of the manual.
Some of the code and the output generated by running the code in this file is simply wrong.
For the rest of it, I think it would be much better to leave the decision about correct GAPDoc syntax to GAPDoc, perhaps via an optional argument that causes a more strict check of references etc. This way, such an official "strict mode" could be used also by package manuals based on GAPDoc, if one really wants this.)

lib/coll.gd
```diff
@@ -2478,6 +2478,120 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+##  This operation can be used to express reductions similar to
+##  <Ref Func="Sum"/> or <Ref Func="Product"/>, but it is more general
+##  because the accumulator is updated by an arbitrary binary function.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> FoldLeft( [ 1 .. 10 ], \+ );
+##  55
+##  gap> FoldLeft( [ 1 .. 4 ], \*, 2 );
+##  48
+##  gap> FoldLeft( [ 1,, 3 ], function( a, b ) return a + b; end );
```
ThomasBreuer:
```suggestion
##  gap> FoldLeft( [ 1,, 3 ], { a, b } -> a + b );
```
Since the whole pull request is about fancy syntax, we could advertise also the fancy function syntax.

lib/coll.gd
```diff
@@ -2939,6 +3053,116 @@ DeclareGlobalFunction( "SumX" );
...
+##
+##  <#GAPDoc Label="FilteredX">
+##  <ManSection>
+##  <Func Name="FilteredX" Arg='arg1, arg2, ... func'/>
+##
+##  <Description>
+##  <Ref Func="FilteredX"/> returns the tuples of loop-variable values for
+##  which the last argument returns <K>true</K>.
+##  In other words, it keeps precisely those tuples that would pass the
+##  filters when calling <Ref Func="ListX"/> with the same generators.
```
ThomasBreuer: generators -> arguments?

lib/coll.gd
```diff
@@ -2939,6 +3053,116 @@ DeclareGlobalFunction( "SumX" );
...
+##
+#O  NumberX( <arg1>, <arg2>, ... <func> )
+##
+##  <#GAPDoc Label="NumberX">
+##  <ManSection>
+##  <Func Name="NumberX" Arg='arg1, arg2, ... func'/>
+##
+##  <Description>
+##  <Ref Func="NumberX"/> returns the number of tuples selected by the last
+##  argument, using the same generators as <Ref Func="ListX"/>.
```
ThomasBreuer: generators -> arguments?

lib/coll.gd
```diff
@@ -2478,6 +2478,120 @@ DeclareOperation( "AsSSortedListNonstored", [IsListOrCollection] );
...
+##  <Item>
+##      this introduces a new for-loop in the sequence of nested
+##      for-loops and if-statements, where the loop-range depends on
+##      the values of the outer loop-variables; or
+##  </Item>
+##  <Mark>a function returning <K>true</K> or <K>false</K></Mark>
+##  <Item>
+##      this introduces a new if-statement in the sequence of nested
+##      for-loops and if-statements.
+##  </Item>
```
ThomasBreuer, line 2559: It is not said what the arguments of the functions in `gens` are.
(The description corresponds to that if `ListX`, there it is also not defined what the arguments are.)

ThomasBreuer (approved):
Nice.
Besides my few comments, perhaps a cross-reference from `ListX` to `FoldLeftX` could be added.
First I thought that manual examples would be nice but also `SetX` and `SumX` have no examples, they just refer to `ListX`.

limakzi (approved):

wilfwilson:
Great work Max and Codex and reviews, for getting this sorted!

PR #3361, opened 2019-03-20
buildsys: resolve some issues if path contains spaces

fingolfin:
Properly quote paths in a few more places to ensure the affected
scripts work inside directories with spaces in their name.

Note that even though a few GMP files also are fixed, this is not
enough to actually allow GMP to build in the described situation:
GMP even contains a check that aborts compilation if there are spaces
in the PWD (but these checks actually only are reached with the
fixes in this PR...). The main reason I added the GMP fixes here is
that perhaps they can be contributed back to upstream, in case
upstream is not completely against trying to address this issue.

Overall, building GAP in a path with whitespaces seems to work fine; but various packages may have issues with it.

### DIFF
--- a/cnf/build-extern.sh
+++ b/cnf/build-extern.sh
@@ -12,7 +12,7 @@ pkg=$1; shift
 src=$1; shift # directory with package sources -- must be an absolute path
 
 builddir=extern/build/$pkg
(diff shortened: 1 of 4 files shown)


fingolfin:
ping @JHahnHRO

PR #3365, opened 2019-03-20
Add First(list), Last(list) and Last(list,func)

fingolfin:
Another PR I had sitting around for a couple years...


fingolfin:
The error is due to the qaos package which defines a global `Last` (and then never uses it). I already removed that in the qaos repository, and thought about making a release... but unfortunately it seems we are going to have to remove qaos from the distribution anyway, see https://github.com/gap-packages/qaos/issues/13

PR #3368, opened 2019-03-21
Removed floating-point constants from StringOfMemoryAmount

laurentbartholdi:


fingolfin:
@stevelinton wrote this code so he may want to comment, but I think this change is pretty much an obvious improvement .

PR #3370, opened 2019-03-21
Added pragmas

sebasguts:
Functions can contain pragmas, i.e., comments
which are stored inside the "compiled" function
as a string.

This is a reimplementation of #1811 

Closes #1811

PR #3371, opened 2019-03-21
Add a coder to reread syntaxtrees

sebasguts:
See [here](https://github.com/fingolfin/gap/pull/114) for the initial discussion and history

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -250,7 +250,7 @@ Obj GET_VALUE_FROM_CURRENT_BODY(Int ix)
     return ELM_PLIST(values, ix);
 }
 
-Stat NewStatWithProf (
+Stat NewStatOrExpr (
     UInt                type,
     UInt                size,
     UInt                line)
@@ -284,7 +284,7 @@ Stat NewStatWithProf (
 
 static Stat NewStat(UInt type, UInt size)
 {
-    return NewStatWithProf(type, size, GetInputLineNumber());
+    return NewStatOrExpr(type, size, GetInputLineNumber());
 }
 
 
@@ -1442,7 +1442,7 @@ void CodeReturnVoidWhichIsNotProfiled ( void )
 
     /* allocate the return-statement, without profile information          */
 
-    stat = NewStatWithProf( T_RETURN_VOID, 0 * sizeof(Expr), 0 );
+    stat = NewStatOrExpr( T_RETURN_VOID, 0 * sizeof(Expr), 0 );
 
     /* push the return-statement                                           */
     PushStat( stat );
--- a/src/code.c
+++ b/src/code.c
@@ -833,7 +833,7 @@ void CodeFuncExprBegin (
     assert( stat1 == OFFSET_FIRST_STAT );
 }
 
-void CodeFuncExprEnd(UInt nr)
+Expr CodeFuncExprEnd(UInt nr, UInt pushExpr)
 {
     Expr                expr;           /* function expression, result     */
     Stat                stat1;          /* single statement of body        */
@@ -853,10 +853,18 @@ void CodeFuncExprEnd(UInt nr)
     }
     else {
         stat1 = PopStat();
-        PushStat( stat1 );
-        if ( TNUM_STAT(stat1) != T_RETURN_VOID
-          && TNUM_STAT(stat1) != T_RETURN_OBJ )
-        {
+        PushStat(stat1);
+        //  If we code a function where the body is already packed into nested
+        //  sequence statements, e.g., from reading in a syntax tree, we need
+        //  to find the last `real` statement of the last innermost sequence
+        //  statement to determine if there is already a return or not.
+        while (T_SEQ_STAT <= TNUM_STAT(stat1) &&
+               TNUM_STAT(stat1) <= T_SEQ_STAT7) {
+            UInt size = SIZE_STAT(stat1) / sizeof(Stat);
+            stat1 = READ_STAT(stat1, size - 1);
+        }
+        if (TNUM_STAT(stat1) != T_RETURN_VOID &&
+            TNUM_STAT(stat1) != T_RETURN_OBJ) {
             CodeReturnVoidWhichIsNotProfiled();
             nr++;
         }
@@ -900,7 +908,10 @@ void CodeFuncExprEnd(UInt nr)
         len = PushValue(fexp);
         expr = NewExpr( T_FUNC_EXPR, sizeof(Expr) );
         WRITE_EXPR(expr, 0, len);
-        PushExpr( expr );
+        if (pushExpr) {
+            PushExpr(expr);
+        }
+        return expr;
     }
 
     // otherwise, make the function and store it in 'CS(CodeResult)'
@@ -908,6 +919,7 @@ void CodeFuncExprEnd(UInt nr)
         CS(CodeResult) = MakeFunction(fexp);
     }
 
+    return 0;
 }
 
 
--- a/src/code.c
+++ b/src/code.c
@@ -1898,7 +1898,7 @@ static UInt CheckForCommonFloat(const Char * str)
         return 0;
 }
 
-static void CodeLazyFloatExpr(Obj str)
+Expr CodeLazyFloatExpr(Obj str, UInt pushExpr)
 {
     UInt ix;
 
@@ -1912,7 +1912,10 @@ static void CodeLazyFloatExpr(Obj str)
     WRITE_EXPR(fl, 1, PushValue(str));
 
     /* push the expression */
-    PushExpr(fl);
+    if (pushExpr) {
+        PushExpr(fl);
+    }
+    return fl;
 }
 
 static void CodeEagerFloatExpr(Obj str, Char mark)
@@ -1947,7 +1950,7 @@ void CodeFloatExpr(Obj s)
         CodeEagerFloatExpr(s, mark);
     }
     else {
-        CodeLazyFloatExpr(s);
+        CodeLazyFloatExpr(s, 1);
     }
 }
 
--- a/src/code.c
+++ b/src/code.c
@@ -540,7 +540,7 @@ static void PushBinaryOp(UInt type)
 }
 
 
-Int PushValue(Obj val)
+Int AddValueToBody(Obj val)
 {
     BodyHeader * header = (BodyHeader *)STATE(PtrBody);
     Obj values = header->values;
@@ -905,7 +905,7 @@ Expr CodeFuncExprEnd(UInt nr, UInt pushExpr)
     /* if this was inside another function definition, make the expression */
     /* and store it in the function expression list of the outer function  */
     if (STATE(CurrLVars) != CS(CodeLVars)) {
-        len = PushValue(fexp);
+        len = AddValueToBody(fexp);
         expr = NewExpr( T_FUNC_EXPR, sizeof(Expr) );
         WRITE_EXPR(expr, 0, len);
         if (pushExpr) {
@@ -1610,7 +1610,7 @@ void CodeIntExpr(Obj val)
     else {
         GAP_ASSERT(TNUM_OBJ(val) == T_INTPOS || TNUM_OBJ(val) == T_INTNEG);
         expr = NewExpr( T_INT_EXPR, sizeof(UInt) );
-        Int ix = PushValue(val);
+        Int ix = AddValueToBody(val);
         WRITE_EXPR(expr, 0, ix);
     }
 
@@ -1809,7 +1809,7 @@ void CodeStringExpr (
     GAP_ASSERT(IS_STRING_REP(str));
 
     Expr string = NewExpr( T_STRING_EXPR, sizeof(UInt) );
-    Int ix = PushValue(str);
+    Int ix = AddValueToBody(str);
     WRITE_EXPR(string, 0, ix);
     PushExpr( string );
 }
@@ -1819,7 +1819,7 @@ void CodePragma(Obj pragma)
     GAP_ASSERT(IS_STRING_REP(pragma));
 
     Expr pragmaexpr = NewStat(T_PRAGMA, sizeof(UInt));
-    Int  ix = PushValue(pragma);
+    Int  ix = AddValueToBody(pragma);
     WRITE_EXPR(pragmaexpr, 0, ix);
     PushStat(pragmaexpr);
 }
@@ -1909,7 +1909,7 @@ Expr CodeLazyFloatExpr(Obj str, UInt pushExpr)
     if (!ix)
         ix = getNextFloatExprNumber();
     WRITE_EXPR(fl, 0, ix);
-    WRITE_EXPR(fl, 1, PushValue(str));
+    WRITE_EXPR(fl, 1, AddValueToBody(str));
 
     /* push the expression */
     if (pushExpr) {
@@ -1923,8 +1923,8 @@ static void CodeEagerFloatExpr(Obj str, Char mark)
     /* Eager case, do the conversion now */
     Expr fl = NewExpr(T_FLOAT_EXPR_EAGER, sizeof(UInt) * 3);
     Obj v = CALL_2ARGS(CONVERT_FLOAT_LITERAL_EAGER, str, ObjsChar[(Int)mark]);
-    WRITE_EXPR(fl, 0, PushValue(v));
-    WRITE_EXPR(fl, 1, PushValue(str));  // store for printing
+    WRITE_EXPR(fl, 0, AddValueToBody(v));
+    WRITE_EXPR(fl, 1, AddValueToBody(str));  // store for printing
     WRITE_EXPR(fl, 2, (UInt)mark);
     PushExpr(fl);
 }
--- a/src/code.c
+++ b/src/code.c
@@ -134,7 +134,7 @@ static Stat * ADDR_STAT(Stat stat)
     return (Stat *)STATE(PtrBody) + stat / sizeof(Stat);
 }
 
-static void WRITE_EXPR(Expr expr, UInt idx, UInt val)
+void WRITE_EXPR(Expr expr, UInt idx, UInt val)
 {
     ADDR_EXPR(expr)[idx] = val;
 }
@@ -250,20 +250,7 @@ Obj GET_VALUE_FROM_CURRENT_BODY(Int ix)
     return ELM_PLIST(values, ix);
 }
 
-
-/****************************************************************************
-**
-*F  NewStat( <type>, <size> ) . . . . . . . . . . .  allocate a new statement
-**
-**  'NewStat'   allocates a new   statement memory block  of  type <type> and
-**  <size> bytes.  'NewStat' returns the identifier of the new statement.
-**
-**  NewStatWithProf( <type>, <size>, <line>, <file> ) allows the line number
-**  and fileid of the statement to also be specified, else the current line
-**  and file when NewStat was called is used. line=0, file=0 is used
-**  to denote a statement which should not be tracked.
-*/
-static Stat NewStatWithProf (
+Stat NewStatWithProf (
     UInt                type,
     UInt                size,
     UInt                line)
@@ -337,7 +324,7 @@ static inline UInt CapacityStatStack(void)
     return SIZE_BAG(CS(StackStat)) / sizeof(Stat) - 1;
 }
 
-static void PushStat (
+void PushStat (
     Stat                stat )
 {
     /* there must be a stack, it must not be underfull or overfull         */
@@ -553,14 +540,7 @@ static void PushBinaryOp(UInt type)
 }
 
 
-/****************************************************************************
-**
-*F  PushValue( <val> ) . . . . . . . . . . . . . . store value in values list
-**
-**  'PushValue' pushes a value into the value list of the body, and returns
-**  the index at which the value was inserted.
-*/
-static Int PushValue(Obj val)
+Int PushValue(Obj val)
 {
     BodyHeader * header = (BodyHeader *)STATE(PtrBody);
     Obj values = header->values;
--- a/src/code.h
+++ b/src/code.h
@@ -692,12 +692,13 @@ Expr CodeFuncExprEnd(UInt nr, UInt pushExpr);
 
 /****************************************************************************
 **
-*F  PushValue( <val> ) . . . . . . . . . . . . . . store value in values list
+*F  AddValueToBody( <val> ) . . . . . . . . . . .  store value in values list
 **
-**  'PushValue' pushes a value into the value list of the body, and returns
-**  the index at which the value was inserted.
+**  'AddValueToBody' adds a value into the value list of the body of the
+**  function currently being coded, and returns the index at which the value
+**  was inserted. This function must only be called while coding a function.
 */
-Int PushValue(Obj val);
+Int AddValueToBody(Obj val);
 
 /****************************************************************************
 **
--- a/src/code.h
+++ b/src/code.h
@@ -1105,6 +1105,13 @@ void CodePragma(Obj pragma);
 
 /****************************************************************************
 **
+*F  CodeLazyFloatExpr(<str>,<pushExpr>) . . . . .  code lazy float expression
+*/
+Expr CodeLazyFloatExpr(Obj str, UInt pushExpr);
+
+
+/****************************************************************************
+**
 *F  CodeFloatExpr(<str>) . . . . . . . . . .  code literal float expression
 */
 void CodeFloatExpr(Obj str);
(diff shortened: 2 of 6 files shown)


fingolfin:
The first commit "WIP: add SyntaxTreeDefaultCoder" should be removed.

src/syntaxtree.c
```diff
@@ -236,15 +333,66 @@ static Obj SyntaxTreeRecExpr(Obj result, Expr expr)
...
+    Obj string = GET_VALUE_FROM_CURRENT_BODY(READ_EXPR(expr, 1));
     AssPRec(result, RNamName("value"), string);
     return result;
 }
 
+static Obj SyntaxTreeFloatEager(Obj result, Expr expr)
+{
+    Obj  value = GET_VALUE_FROM_CURRENT_BODY(READ_EXPR(expr, 0));
+    Obj  string = GET_VALUE_FROM_CURRENT_BODY(READ_EXPR(expr, 1));
+    UInt char_val = (UInt)READ_EXPR(expr, 2);
```
fingolfin:
```suggestion
    UChar chr = (UChar)READ_EXPR(expr, 2);
```
Otherwise an out of bounds access could happen a few lines later on.

And `char_val` violates our naming conventions.

src/syntaxtree.c
```diff
@@ -236,15 +333,66 @@ static Obj SyntaxTreeRecExpr(Obj result, Expr expr)
...
+    AssPRec(result, RNamName("value"), value);
+    AssPRec(result, RNamName("string"), string);
+    AssPRec(result, RNamName("mark"), ObjsChar[char_val]);
```
fingolfin:
```suggestion
    AssPRec(result, RNamName("mark"), ObjsChar[chr]);
```

src/code.c
```diff
@@ -1909,7 +1893,10 @@ static void CodeLazyFloatExpr(Obj str)
     WRITE_EXPR(fl, 1, PushValue(str));
 
     /* push the expression */
-    PushExpr(fl);
+    if(pushExpr) {
+        PushExpr(fl);
+    }
+    return fl;
```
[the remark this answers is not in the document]
sebasguts: I really do not understand this comment. All modifications in `CodeLazyFloatExpr` are made in `0a00459bd2a3eb87df8d0ef22561c1d7a2d770e5`, and then the next commit uses it.

sebasguts:
@fingolfin I think I adressed all your comments. I also added sensible error messages for checking if a needed entry in a syntax tree node record is bound.

src/code.c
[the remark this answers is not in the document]
sebasguts: I now understand this comment. Apparently, Github displays commits by date, not in the order they appear on the branch . . . stupid rebasing. Should I make a rebase with new dates?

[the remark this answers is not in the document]
ChrisJefferson: adding `--ignore-date` to rebasing will make it set all the dates to the current date, which gets all your commits in the "right order" again.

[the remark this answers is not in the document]
sebasguts: Yeah, but I tent to not do that, for reasons that are probably not really logical. Will rebase this one with new dates.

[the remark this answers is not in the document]
sebasguts: Did it, now they all have the same date and still appear in the wrong order . . .

src/syntaxtree.c
```diff
@@ -31,50 +31,92 @@
...
 
 static Obj SyntaxTreeFunc(Obj result, Obj func);
 
+static UInt1 FIND_TNUM_FROM_STRING(Obj typestr)
+{
+    UInt rnam = RNamObj(typestr);
+    if (IsbPRec(typeRec, rnam)) {
+        return (UInt1)UInt_ObjInt(ElmPRec(typeRec, rnam));
+    }
+    ErrorQuit("Unrecognized syntax tree node type %s",
```
fingolfin:
```suggestion
    ErrorQuit("Unrecognized syntax tree node type %g",
```

src/syntaxtree.c
```diff
@@ -31,50 +31,92 @@
...
+              (Int)CSTR_STRING(typestr), 0);
```
fingolfin:
```suggestion
              (Int)typestr, 0);
```

src/syntaxtree.c
```diff
@@ -31,50 +31,92 @@
...
+}
+
+static UInt1 GetTypeTNum(Obj node)
+{
+    UInt type = RNamName("type");
+    if (!IsbPRec(node, type)) {
+        ErrorQuit("while coding: <type> field is not present", 0, 0);
+    }
+    return FIND_TNUM_FROM_STRING(ELM_REC(node, type));
```
fingolfin:
```suggestion
    return FIND_TNUM_FROM_STRING(ElmPRec(node, type));
```
You just called `IsbPRec` a few lines before, and use `ElmPRec`, so might as well use that here. Or, if we truly want to support general records (in case somebody ever adds some), we should be using `ISB_REC`, `UNB_REC`, `ELM_REC` everywhere. But I think it's OK to restrict to precs

sebasguts:
This PR is now based on #3370

src/code.h
```diff
@@ -109,6 +109,25 @@ EXPORT_INLINE Obj VALUES_BODY(Obj body)
...
+**
+**  'NewStat'   allocates a new   statement memory block  of  type <type> and
+**  <size> bytes.  'NewStat' returns the identifier of the new statement.
+**
+**  NewStatWithProf( <type>, <size>, <line>, <file> ) allows the line number
+**  and fileid of the statement to also be specified, else the current line
+**  and file when NewStat was called is used. line=0, file=0 is used
+**  to denote a statement which should not be tracked.
+*/
+Stat NewStatWithProf(UInt type, UInt size, UInt line);
```
fingolfin: Better comment:
```
**
*F  NewStatOrExpr(<type>,<size>,<line>,<file>) . . . allocate a new statement
**
**  'NewStatOrExpr' allocates a new statement or expressions memory block of
**  type <type> and with <size> bytes. It also records the line number <line>
**  and file id <file> as source of the statement for profiling. It returns
**  the offset of the new statement.
**
**  Callers may pass zero for both <line> and <file> to denote a statement
**  which should not be tracked by the profiling code.
*/
```
and then rename `NewStatWithProf` to `NewStatOrExpr` globally (could be in a commit of its own, don't care).

src/code.h
```diff
@@ -661,12 +684,25 @@ void CodeFuncCallEnd(UInt funccall, UInt options, UInt nr);
...
 
-void CodeFuncExprEnd(UInt nr);
+Expr CodeFuncExprEnd(UInt nr, UInt pushExpr);
+
+/****************************************************************************
+**
+*F  AddValueToBody( <val> ) . . . . . . . . . . .  store value in values list
+**
+**  'AddValueToBody' adds a value into the value list of the body of the
+**  function currently being coded, and returns the index  at which the
```
fingolfin:
```suggestion
**  function currently being coded, and returns the index at which the value
```

src/code.h
```diff
@@ -661,12 +684,25 @@ void CodeFuncCallEnd(UInt funccall, UInt options, UInt nr);
...
+**  value was inserted. This function must only be called while coding a
```
fingolfin:
```suggestion
** was inserted. This function must only be called while coding a function.
```

src/code.h
```diff
@@ -661,12 +684,25 @@ void CodeFuncCallEnd(UInt funccall, UInt options, UInt nr);
...
+**  function.
```
fingolfin:
```suggestion
```

src/code.h
```diff
@@ -661,12 +684,25 @@ void CodeFuncCallEnd(UInt funccall, UInt options, UInt nr);
...
+**
```
fingolfin:
```suggestion
```

src/code.h
```diff
@@ -1069,6 +1105,16 @@ void CodeListExprEnd(UInt nr, UInt range, UInt top, UInt tilde);
 */
 void CodeStringExpr(Obj str);
 
+void CodePragma(Obj pragma);
+
+
+/****************************************************************************
+**
+*F  CodeLazyFloatExpr(<str>,<pushExpr>) . . .  code literal string expression
```
fingolfin:
```suggestion
*F  CodeLazyFloatExpr(<str>,<pushExpr>) . . . . .  code lazy float expression
```

src/scanner.c
```diff
@@ -746,6 +746,30 @@ static void GetStr(void)
...
 
+
+static void GetPragma(void)
+{
+    Obj  string = 0;
+    Char buf[1024];
+    UInt i = 0;
+    Char c = PEEK_CURR_CHAR();
+
+    while ( c != '\n' && c != '\r' && c != '\f' && c != '\t' && c != '\377') {
```
fingolfin:
```suggestion
    while (c != '\n' && c != '\r' && c != '\f' && c != '\t' && c != '\377') {
```

src/scanner.h
```diff
@@ -63,6 +63,8 @@ enum SCANNER_SYMBOLS {
     S_STRING            = (1UL<<11)+3,
     S_TILDE             = (1UL<<11)+4,
     S_HELP              = (1UL<<11)+5,
+    S_PRAGMA            = (1UL<<11)+6,
+
```
fingolfin:
```suggestion
```

src/syntaxtree.c
```diff
@@ -272,6 +453,57 @@ static Obj SyntaxTreeIf(Obj result, Stat stat)
...
+    UInt1 tnum = GetTypeTNum(node);
+    Obj   branches = ElmRecST(tnum, node, "branches");
+    UInt  len = LEN_PLIST(branches);
+    Expr  ifexpr = NewStatWithProf(tnum, 2 * len * sizeof(Expr), 0);
+    for (int i = 0; i < len; i++) {
+        Obj  condbodypair = ELM_PLIST(branches, i + 1);
+        Obj  conditionnode = ElmRecST(tnum, condbodypair, "condition");
+        Obj  bodynode = ElmRecST(tnum, condbodypair, "body");
+        Expr condition = SyntaxTreeDefaultExprCoder(conditionnode);
+        Expr body = SyntaxTreeDefaultStatCoder(bodynode);
```
fingolfin:
```suggestion
        Stat body = SyntaxTreeDefaultStatCoder(bodynode);
```

src/syntaxtree.c
```diff
@@ -236,15 +366,66 @@ static Obj SyntaxTreeRecExpr(Obj result, Expr expr)
     return result;
 }
 
-static Obj SyntaxTreeFloatLazy(Obj result, Expr expr)
+static Expr SyntaxTreeCodeRecExpr(Obj node)
 {
-    Obj string;
+    UInt1 tnum = GetTypeTNum(node);
+    Obj   keyvalue = ElmRecST(tnum, node, "keyvalue");
+    UInt  len = LEN_LIST(keyvalue);
```
fingolfin:
```suggestion
    UInt  len = LEN_PLIST(keyvalue);
```

sebasguts:
I think I addressed all your comments, and also added the checks we discussed today.

Should we try to add the context object, too, or would be a different PR a better place.

src/syntaxtree.c
ChrisJefferson: I might be going the other way to @fingolfin, but I was a bit worried at the number of uses of the PLIST functions in this, which don't check they are PLISTs. Using the LIST functions instead, while a little slower, would at least make sure GAP didn't crash/corrupt.

sebasguts: I wouldn't mind using either the one or the other.

src/syntaxtree.c
```diff
@@ -279,6 +482,49 @@ static Obj SyntaxTreeCompilePragma(Obj result, Stat stat)
     return result;
 }
 
+static Expr SyntaxTreeCodeIf(Obj node)
+{
+    UInt1 tnum = GetTypeTNum(node);
+    Obj   branches = ElmRecST(tnum, node, "branches");
+    UInt  len = LEN_PLIST(branches);
```
fingolfin: So @ChrisJefferson pointed out in another comment (now on "outdated" code, so not immediately visible), where I suggested to change a `LEN_LIST` to `LEN_PLIST`, that we could go the other way, and change all PLIST functions to generic ones, for safety. I am also fine with this -- the alternative being to add lost of "is this really a pist" checks everywhere. What I do not like is code which calls `LEN_LIST` and then immediately `ELM_PLIST`, as this code used to do, and as unfortunately still quite some other kernel code does. It combines the worst of both worlds: inferior performance plus a risk of crashes...

Anyway, performance shouldn't matter much here, so I'd be fine if we replaced `LEN_PLIST` with `LEN_LIST`, `ELM_PLIST` with `ELM_LIST` (resp. `ELM0_LIST` in the few cases where it is needed), `SET_ELM_PLIST` with `ASS_LIST`, etc.

src/syntaxtree.c
```diff
@@ -31,50 +31,101 @@
...
 
+static Obj typeStrings;    // plain list mapping statement tnums to strings
+                           // like "T_INTEXPR"
+static Obj typeRec;    // plain record mapping strings like "T_INTEXPR" to the
+                       // corresponding statement tnum
+
+static Obj ElmRecST(UInt1 tnum, Obj node, const char * name)
+{
+    UInt rnam = RNamName(name);
+    if (!IsbPRec(node, rnam)) {
```
fingolfin: We never verify that the object we get is a precord, do we? So this would crash if I created malicious input (respectively: if I try to create a valid syntax tree object but screwed it up).

As with lists, we can either insert `RequirePRecord` tests everywhere (after creating such a function). Or we just use `ISB_REC`, `ELM_REC`, etc.

It might also be instructive to write malicious input just to see how it crashes and burns w/o these error checks. :-).

src/syntaxtree.c
```diff
@@ -31,50 +31,101 @@
...
+#define ARG(name, func, coder, isStat)                                       \
     {                                                                        \
-        name, func                                                           \
+        name, func, coder, isStat                                            \
     }
-#define ARG_(name) ARG(name, SyntaxTreeCompiler)
 
-#define ARGS(name) ARG(name, 0)
+#define ARG_EXPR(name, func, coder) ARG(name, func, coder, 0)
+// The corresponding ARG_STAT is not used
```
fingolfin: Huh? Perhaps you meant this?
> There is no ARG_STAT, because it is never used

That surprise me, though: T_ATOMIC contains a single body statement, plus a variable list of expressions for locks... Hmm, looking at your code for T_ATOMIC, it seems to be incorrect, which would explain why there is no ARG_STAT here.

src/syntaxtree.c
```diff
@@ -31,50 +31,101 @@
...
-#define ARG(name, func)                                                      \
+#define ARG(name, func, coder, isStat)                                       \
     {                                                                        \
-        name, func                                                           \
+        name, func, coder, isStat                                            \
     }
-#define ARG_(name) ARG(name, SyntaxTreeCompiler)
 
-#define ARGS(name) ARG(name, 0)
+#define ARG_EXPR(name, func, coder) ARG(name, func, coder, 0)
```
fingolfin: Now would be a good time to also add some comments that explain what these macros do... What is `ARG_EXPR` about? Perhaps like this:
> // ARG_EXPR indicates an "expression argument", i.e., that the corresponding slot of the
> // statement or expression type it belongs to is meant to reference an expression.
> // This information is used for validation when coding an AST back into a T_BODY object.
> // The expression is stored in resp. read from the record entry <name>.

And then later:
> // ARG_STAT is like ARG_EXPR, except that it indicates that the corresponding slot should contain
> // a statement instead of an expression.

It's not great, but perhaps better than nothing. And it could be adapted to A

src/syntaxtree.c
```diff
@@ -31,50 +31,101 @@
...
+// The corresponding ARG_STAT is not used
+
+#define ARG_STAT_(name)                                                      \
+    ARG(name, SyntaxTreeCompiler, SyntaxTreeDefaultStatCoder, 1)
+#define ARG_EXPR_(name)                                                      \
+    ARG(name, SyntaxTreeCompiler, SyntaxTreeDefaultExprCoder, 0)
 
+#define ARGS_STAT(name) ARG(name, 0, 0, 1)
+#define ARGS_EXPR(name) ARG(name, 0, 0, 0)
```
fingolfin: > // ARGS_EXPR can be used instead of ARG_EXPR as the last argument to the COMPILER_
> // macro, and indicates that a variable number of arguments follow; all are supposed to be
> // expressions; and all of them are to be stored into a plist which is stored in resp. read
> // from the record entry <name>.

src/syntaxtree.c
```diff
@@ -31,50 +31,101 @@
...
 
-static Obj typeStrings;
 
+static Obj typeStrings;    // plain list mapping statement tnums to strings
+                           // like "T_INTEXPR"
+static Obj typeRec;    // plain record mapping strings like "T_INTEXPR" to the
```
fingolfin: We usually put such comments before the variable, not after it, as you do here.

src/syntaxtree.c
sebasguts: The single body statement is handles by `ARG_STAT_` since it is just a `T_SEQ...` I think. So we do not need `ARG_STAT` which allows for a custom coder/compiler.

sebasguts: Anyway, the code was indeed wrong. Fixed that.

sebasguts:
@fingolfin I think I adressed all of your comments

sebasguts:
@ChrisJefferson @fingolfin ping?

src/code.h
```diff
@@ -109,6 +109,25 @@ EXPORT_INLINE Obj VALUES_BODY(Obj body)
...
+/****************************************************************************
+**
+*F  NewStatOrExpr(<type>,<size>,<line>,<file>) . . . allocate a new statement
+**
+**  'NewStatOrExpr' allocates a new statement or expressions memory block of
+**  type <type> and with <size> bytes. It also records the line number <line>
+**  and file id <file> as source of the statement for profiling. It returns
+**  the offset of the new statement.
+**
+**  Callers may pass zero for both <line> and <file> to denote a statement
```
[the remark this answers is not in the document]
ChrisJefferson: I agree. This is an old bug, at some point the file argument disappeared and the comment wasn't fixed, but it would be reasonable to fix it now.

src/syntaxtree.c
```diff
@@ -31,56 +31,127 @@
...
+// read from the record entry <name>.
+#define ARG_EXPR(name, func, coder) ARG(name, func, coder, 0)
+
+// ARG_STAT is like ARG_EXPR, except that it indicates that the corresponding
+// slot should contain a statement instead of an expression. There is no
+// ARG_STAT, because it is never used.
+
+
+// ARG_EXPR_ can be used instead of ARG_EXPR if the default compiler
+// and coder should be used to compile the corresponding expression or code
```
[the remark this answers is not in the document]
sebasguts: I am fine with using any of those terms. `Code` is currently used in the interface, `Compile` is not. So changing compile to `decode` is the easiest. But we can safely assume that there is nobody using it right now anyway, so we can just change it to whatever we want.

sebasguts:
I think I adressed all the comments now :)

ChrisJefferson:
So, possibly stupid question, is this supposed to be complete, or progress towards a rereader?

I tried extending the test which tests SYNTAX_TREE on every function, by the following horrible patch:

```
@@ -25,11 +25,17 @@ gap> for n in NamesGVars() do
 >            v := ValueGlobal(n);
 >            if IsFunction(v) and not IsKernelFunction(v) then
 >                SYNTAX_TREE(v);
+>                        if not test_tree(v) then
+>                           Print("failed round trip: ",v,"\n");
+>                        fi;
 >            elif IsOperation(v) then
 >                for i in [1..6] do
 >                    for x in METHODS_OPERATION(v, i) do
 >                        if IsFunction(x) and not IsKernelFunction(v) then
 >                        SYNTAX_TREE(x);
+>                        if not test_tree(x) then
+>                           Print("failed round trip: ",v,"\n");
+>                        fi;
 >                        fi;
 >                    od; 
 >                od;
```

And GAP segfaults.

sebasguts:
@ChrisJefferson It does not work on operations, no. But I should really check that when calling `SyntaxTree` The segfault is really bad . . .

sebasguts:
Wait, this is a different issue. Sorry.

ChrisJefferson:
This is enough to upset it:

```
gap> f := function() return [,[]]; end;
function(  ) ... end
gap> SYNTAX_TREE_CODE(SYNTAX_TREE(f));
```

sebasguts:
I will fix that. I am not sure how that slipped. Thank you.

Anyway, there is a thing that I am aware of that does not work yet, namely `HVAR` references that do not point to the "read" function context anymore, but outside.

I plan to overcome this by adding a `T_REF_VALUE` (and `T_ASS_VALUE`) to GAP, so we can store those references not as `HVAR`s anymore, but as values directly. However, this of course may change the overall semantics of a function . . .

ChrisJefferson:
(I don't know if this is reasonable), if in cases where SYNTAX_TREE_CODE failed it just returned 'fail', then we could obviously test for when it works, or returns fail, and then still loop over all the functions? (I don't know if that's reasonable given the way it works internally)

sebasguts:
@ChrisJefferson Currently, they throw an error, which I would guess is more reasonable. But I could do a `no_error` version for testing, if you would prefer?

The problem rereading those `HVAR` functions is that rereading them works, but their result is arbitrary, see here (which is worse than a segfault)):
```
gap> xx := (i -> ( j-> i + j ) )(3);
function( j ) ... end
gap> SYNTAX_TREE(SYNTAX_TREE_CODE(SYNTAX_TREE(xx)))=SYNTAX_TREE(xx);
true
gap> SYNTAX_TREE_CODE(SYNTAX_TREE(xx))(3);
Z(3)
```

ChrisJefferson:
I was just thinking about if it would be reasonable to merge something like the diff I made, to test a large set of functions. If we are still too far away from that, we can also merge the current partial code, and work on improving it later.

sebasguts:
No, absolutely reasonable. I will include it.

sebasguts:
@ChrisJefferson I added the full cycle test, and they pass
@fingolfin I made additional changes to `CodeFuncExprEnd`, please have a new look at the changes there.

tst/testinstall/syntaxtree.tst
```diff
@@ -10,19 +10,30 @@ gap> SyntaxTree(x -> x);
...
+>                if not test_tree(v) then
+>                   Print("failed round trip: ",n,"\n");
+>                fi;
 >            elif IsOperation(v) then
 >                for i in [1..6] do
 >                    for x in METHODS_OPERATION(v, i) do
 >                        if IsFunction(x) and not IsKernelFunction(v) then
->                        SYNTAX_TREE(x);
+>                            if not test_tree(v) then
+>                                Print("failed round trip: ",n,"\n");
```
fingolfin: Would be more useful like this:
```suggestion
>                                Print("METHODS_OPERATION(", n, ",", i, ") failed round trip\n");
```

src/code.c
```diff
@@ -874,11 +854,21 @@ void CodeFuncExprEnd(UInt nr)
     else {
         stat1 = PopStat();
         PushStat( stat1 );
-        if ( TNUM_STAT(stat1) != T_RETURN_VOID
-          && TNUM_STAT(stat1) != T_RETURN_OBJ )
-        {
-            CodeReturnVoidWhichIsNotProfiled();
-            nr++;
+        while( 1 ){
```
fingolfin: Code formatting!
```suggestion
        while (1) {
```

fingolfin: Also, this code *really* needs to be commented. Both what it is doing (see my suggestion below), but also more importantly, *why* it is doing it. Which I don't know right now...

fingolfin: Why doesn't it suffice to insert the following into the existing code, to achieve the same effect?
```
  // find the last statement
  while (T_SEQ_STAT <= TNUM_STAT(stat1) && TNUM_STAT(stat1) <= T_SEQ_STAT7) {
    TNUM_STAT(stat1) <= T_SEQ_STAT7) {
    UInt size = SIZE_STAT(stat1) / sizeof(Stat);
    stat1 = READ_STAT(stat1, size - 1);
  }
```

sebasguts: Well, logically, this does exactly the same, except it is not intertwined with the addition of the `return void` to the statement.

sebasguts: About the why: The thing is that if we reread a syntax tree, the statements are already nested, i.e., if there are more than 7 statements in the body, the last one is already a new `SEQ_STAT*`. This ensures that the `return;` is only added if there is really no `return` in the last statement.

When a function is coded, the statements are packed into nested statements later.

I discovered this problem when I realized that "long" functions get an additional `return;`

PR #3372, opened 2019-03-21
Make some interpreter / reader errors a bit nicer

### LINKED ISSUE #1117: Unhelpful error message (?)
When playing with the parser I get this

```
gap> function()  
> 1+1;
Syntax error: end expected
1+1;
^
```

fingolfin:
Previously, we produced errors like this:

    gap> function() 123; end;
    Syntax error: inside a function, statement or 'end' expected
    function() 123; end;
               ^^^

This now is slightly more helpful and reads as follows.

    gap> function() 123; end;
    Syntax error: while parsing a function: statement or 'end' expected
    function() 123; end;
               ^^^

Similar adjustments are made for loops, atomic blocks and if statements.

Resolves #1117

### DIFF
--- a/tst/testbugfix/2012-09-06-t00253.tst
+++ b/tst/testbugfix/2012-09-06-t00253.tst
@@ -13,6 +13,7 @@ gap> Read(s);
 Syntax error: ) expected in stream:4
     v := rec(a := [];);
                     ^
-Syntax error: end expected in stream:5
+Syntax error: while parsing a function: statement or 'end' expected in stream:\
+5
 od;
 ^^
--- a/src/read.c
+++ b/src/read.c
@@ -1338,7 +1338,7 @@ static void ReadFuncExpr (
     ReadFuncExprBody(follow, 0, nloc, args, startLine);
 
     /* 'end'                                                               */
-    Match( S_END, "end", follow );
+    Match(S_END, "while parsing a function: statement or 'end'", follow);
 }
 
 
@@ -1994,7 +1994,7 @@ static void ReadIf (
     }
 
     /* 'fi'                                                                */
-    Match( S_FI, "fi", follow );
+    Match(S_FI, "while parsing an 'if' statement: statement or 'fi'", follow);
     TRY_IF_NO_ERROR { IntrIfEnd( nrb ); }
 }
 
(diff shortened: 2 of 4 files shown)


tst/testbugfix/2012-09-06-t00253.tst
```diff
@@ -13,6 +13,6 @@ gap> Read(s);
 Syntax error: ) expected in stream:4
     v := rec(a := [];);
                     ^
-Syntax error: end expected in stream:5
+Syntax error: inside a function, statement or 'end' expected in stream:5
```
fingolfin: reading this again today, perhaps a `semicolon` would make this a bit clearer, as here (note that the "expected" is added automatically at the end, so we have to contort ourselves a bit to fit the sentence structure in there)
```suggestion
Syntax error: inside a function; statement or 'end' expected in stream:5
```

Thoughts?

ChrisJefferson: While ; might be more mathematically correct, I feel that in code I always read it as "end of statement".

How about "Syntax error while parsing a function: statement or 'end' expected" (so push the new stuff before the semicolon)

fingolfin: The "Syntax error:" prefix and "expected" suffix are fixed, I can only fill in stuff between them, so the best I can offer w/o deeper changes is this:
> Syntax error: while parsing a function: statement or 'end' expected

wilfwilson: What do you think about either of these options, which break the error into two sentences with a full stop? (It also adds commas around the "or" option; the difference between the two options is the presence of the word "was")

> Syntax error: inside a function. A statement, or 'end', was expected in stream:5

> Syntax error: inside a function. A statement, or 'end', expected in stream:5

However I don't feel particularly strongly about any of the suggestions put forth in this PR/discussion, so I don't really mind.

fingolfin: Now updated

fingolfin: @wilfwilson sorry, I did not actually see your comment before I updated this PR, i.e., I did not deliberately ignore it! Your variant is also fine by me. I'll think about it again tomorrow, after some sleep. Perhaps others have an opinion, too

fingolfin:
I think I'll just merge this as is: it is an improvement over the current status quo, and a truly better message will require more flexibility in the code which generates those messages (to not force the trailing "expected" etc.).

PR #3375, opened 2019-03-23
Give more helpful warning when using tabs in continuations

### LINKED ISSUE #3068: Tabs on continuation lines break GAP's testing

ChrisJefferson:
This rejects continuations using 'tab' inside tests. They can
be confusing for users because GAP does not treat them as
continuations, but instead a line of output.

They are rejected, rather than treated as continuations, because
we do not want different parts of GAP to handle continuations
differently, or really start changing the definition of something
so fundamental.

This does break any tests which start an output line with '>\t',
but no such tests seem to exist and would be easily confused with
a continuation.

Fixes #3068 (by giving a nice error) while hopefully not breaking any tests (unlike my previous attempt at fixing this, which broke Browse).

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -101,23 +101,23 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments, fnam)
         Add(outp[1], '\n');
       fi;
     elif Length(lines[i]) = 0 and ignorecomments = true and i < Length(lines) 
-         and Length(lines[i+1]) > 0 and lines[i+1][1] = '#' then
+         and StartsWith(lines[i+1], "#") then
       # ignore an empty line followed by comment lines
       Add(ign, i);
       i := i+1;
-      while i <= Length(lines) and Length(lines[i]) > 0 and
-            lines[i][1] = '#' and not StartsWith(lines[i], "#@") do
+      while i <= Length(lines) and StartsWith(lines[i], "#") and
+            not StartsWith(lines[i], "#@") do
         Add(ign, i);
         i := i+1;
       od;
-    elif Length(lines[i]) > 4 and lines[i]{[1..5]} = "gap> " then
+    elif StartsWith(lines[i], "gap> ") then
       foundcmd := false;
       Add(outp, "");
       Add(inp, lines[i]{[6..Length(lines[i])]});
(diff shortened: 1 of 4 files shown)


lib/test.gi
```diff
@@ -124,6 +124,8 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments, fnam)
       Append(inp[Length(inp)], lines[i]{[3..Length(lines[i])]});
       Add(inp[Length(inp)], '\n');
       i := i+1;
+    elif Length(lines[i]) > 1 and lines[i]{[1..2]} = ">\t" then
```
fingolfin: Maybe it is just me, but wouldn't this make the code a bit easier to understand?
```suggestion
    elif StartsWith(lines[i], ">\t") then
```

tst/testspecial/broken-test-6.tst
```diff
@@ -0,0 +1,6 @@
+gap> f := function()
```
fingolfin:
```suggestion
gap> # continuation prompt followed by a tab leads to an error
gap> f := function()
```

lib/test.gi
```diff
@@ -124,6 +124,8 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments, fnam)
...
+        testError("Invalid test file: Continuation '> ' uses a tab instead of a space");
```
fingolfin: When reading error messages like this, I am always confused whether they tell me what *is*, or what *should be*. Perhaps this could be made clearer? How about this?
```suggestion
        testError("Invalid test file: Continuation prompt '> ' followed by a tab, expected a regular space");
```

ChrisJefferson:
All cleaned up. Also, did a quick cleanup using `StartsWith` consistently.

PR #3376, opened 2019-03-23
Add SortedList method with function

### LINKED ISSUE #1002: FEATURE: Add `SortedList(list, cmpFunc)`
It would be nice to have a two-argument variant of `SortedList`, mirroring the two-argument version of `Sort`. It would do roughly this:
```gap
SortedList := (list, cmp)
  list := ShallowCopy(list);
  Sort(list, cmp);
  return list;
end;
```
While at it, one could also add a `SortedListBy` variant which uses `SortBy` instead of `Sort`.

This request is inspired by http://tracker.gap-system.org/issues/402

james-d-mitchell:
This PR adds a method for `SortedList` that takes a function as second argument. 

Resolves #1002.

### DIFF
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -2097,7 +2097,7 @@ DeclareOperation( "ListOp", [ IsListOrCollection, IsFunction ] );
 ##
 ##  <#GAPDoc Label="SortedList">
 ##  <ManSection>
-##  <Oper Name="SortedList" Arg='listorcoll'/>
+##  <Oper Name="SortedList" Arg='listorcoll[, func]'/>
 ##
 ##  <Description>
 ##  <Ref Oper="SortedList"/> returns a new mutable and dense list <A>new</A>.
@@ -2107,7 +2107,8 @@ DeclareOperation( "ListOp", [ IsListOrCollection, IsFunction ] );
 ##  <C>Length( <A>new</A> )</C> is the number of elements of
 ##  <A>listorcoll</A>,
 ##  and <A>new</A> contains the elements in sorted order,
-##  w.r.t.&nbsp;<C>&lt;=</C>.
+##  w.r.t.&nbsp;<C>&lt;</C> or <A>func</A> if it is specified.
+##  For details, please refer to <Ref Oper="Sort"/>.
 ##  <C><A>new</A>[<A>pos</A>]</C> executes in constant time
 ##  (see&nbsp;<Ref Filt="IsConstantTimeAccessList"/>),
 ##  and the size of <A>new</A> in memory is proportional to its length.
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
I’m pretty sure that if the appveyor tests are re run this will pass, the failure looks like it happened because I force pushed almost immediately after opening the PR

lib/coll.gd
```diff
@@ -2107,7 +2107,7 @@ DeclareOperation( "ListOp", [ IsListOrCollection, IsFunction ] );
 ##  <C>Length( <A>new</A> )</C> is the number of elements of
 ##  <A>listorcoll</A>,
 ##  and <A>new</A> contains the elements in sorted order,
-##  w.r.t.&nbsp;<C>&lt;=</C>.
+##  w.r.t.&nbsp;<C>&lt;=</C> or <A>func</A> if it is specified.
```
fingolfin: I realize this was here before, but: why does this say `<=` and not `<` as in the documentation for `Sort` etc? That may matter in case the list contains two copies of the same element... is `func` then supposed to return true or false? Hmm, seeing that this actually calls `Sort`, it seems the current documentation is actually incorrect!

Overall, this text doesn't really make it clear what `func` is about. The quickest way to make this clearer might be to reference the documentation for `Sort` here; alternatively, one could also copy text from there, but seeing as the implementation just calls `Sort`, I think it would make a lot of sense to just link to the Sort documentation, this way, they'll always be automatically in sync.

Perhaps like this:
```suggestion
##  w.r.t.&nbsp;<C>&lt;</C> or <A>func</A> if it is specified.
##  For details, please refer to <Ref Oper="Sort"/>.
```

tst/testinstall/coll.tst
```diff
@@ -154,6 +154,12 @@ gap> SetEnumeratorSorted(M0, []);
 gap> RepresentativeSmallest(M0);
 Error, <C> must be nonempty to have a representative
 
+# SortedList 
+gap> SortedList([2, 1, 3]);
+[ 1, 2, 3 ]
+gap> SortedList([2, 1, 3], {x, y} -> y < x);
+[ 3, 2, 1 ]
```
fingolfin: Minor (completely optional) suggestion: Even better would be to also test that the input is not modified. E.g. by passing an immutable list to the function.

fingolfin:
(and feel free to squash any suggested changes by me, I do not need "credit" for them in any form either ;-). And don't worry about AppVeyor failing, that is simply broken right now (but hopefully I can get it fixed via PR #3377)

fingolfin:
I took the liberty of adding `Resolves #1002.` to the issue description, so that this issue gets closed when we merge this PR.

PR #3377, opened 2019-03-24
Fix AppVeyor

fingolfin:
Let's see if this fixes AppVeyor builds (w/o breaking coverage and what not else)


fingolfin:
That did *not* help. Now it just dies in the next line:
```
SET "PATH=%CYG_ROOT%\bin;%PATH%"
bash -lc "python -m pip install gcovr"
Command exited with code 127
```
Perhaps the path is wrong now, or something else in that vein?

fingolfin:
At least part of the problem is that there are two different Python's installed, one by cygwin and one by AppVeyor, and they mix in a bad way...

PR #3384, opened 2019-03-27
performance enhancements in subgroup lattice

hulpke:
With this, GAP can calculate the subgroups of HNM4 on 1920 points -- before runs out of memory.

### DIFF
--- a/lib/grppclat.gi
+++ b/lib/grppclat.gi
@@ -259,7 +259,7 @@ end);
 ##
 InstallGlobalFunction(ActionSubspacesElementaryAbelianGroup,function(arg)
 local P,g,op,act,a,pcgs,ma,mat,d,f,i,j,new,newmat,id,p,dodim,compldim,compl,
-      dims,Pgens,Pcgens,Pu,Pc,perms,one,par,ker,kersz;
+      dims,Pgens,Pcgens,Pu,Pc,perms,one,par,ker,kersz,pcelm,pccache,asz;
 
   P:=arg[1];
   if IsModuloPcgs(arg[2]) then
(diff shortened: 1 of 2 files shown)


lib/grppclat.gi
```diff
@@ -432,15 +432,32 @@ local P,g,op,act,a,pcgs,ma,mat,d,f,i,j,new,newmat,id,p,dodim,compldim,compl,
...
   new:=[];
   for i in ma do
     #a:=SubgroupNC(Parent(g),List(i,j->Product([1..d],k->pcgs[k]^j[k])));
+    asz:=kersz*p^Length(i);
     if kersz=1 then
-      a:=SubgroupNC(par,List(i,j->PcElementByExponentsNC(pcgs,j)));
+      a:=SubgroupNC(par,List(i,j->pcelm(j)));
     else
-      a:=ClosureGroup(ker,List(i,j->PcElementByExponentsNC(pcgs,j)));
+      #a:=ClosureGroup(ker,List(i,j->pcelm(j)):knownClosureSize:=asz);
```
wilfwilson: Why did you replace the `ClosureGroup` with the `SubgroupNC`? Is it faster? And given the replacement, why have you kept the `ClosureGroup` line as a comment?

lib/grpperm.gi
```diff
@@ -736,6 +736,8 @@ BindGlobal("DoClosurePrmGp",function( G, gens, options )
             fi;
         od;
     else
+        o:=ValueOption("knownClosureSize");
```
wilfwilson: What do you intend for the `knownClosureSize` value option: if you're not using it in the code in the rest of the PR, and it's not documented, is there any point in it being there?

hulpke:
@wilfwilson This is a bit complicated:
The underlying problem, `Closure(Sub)group` taking significant time pops up in a number of solvable radical type calculations. In the case of subgroup lattice I now observed cases where it became so extreme that it required to be resolved.

The "knownClosureSize" was the first attempt, it helped, but not enough. I suspect there will be other cases where it should be used, which is the reason for keeping the code in.
The reason for keeping the commented line is that "morally" this should be a closure computation, not a new subgroup. I'm still slightly puzzled why the stabilizer chains computed by `Closure` take up so much memory, and I hope in the long run to have a "cleaner" fix, preferrably one that actually still can use `Closure`.

wilfwilson:
Thanks for the explanation 🙂 @hulpke

PR #3385, opened 2019-03-28
Add Flip/SetAll/ClearAllBlist

james-d-mitchell:
This PRs adds three functions for blists, `FlipBlist`, `SetAllBitsBlist`, and `ClearAllBitsBlist`. I found that I wanted these three functions in various pieces of code, and so I thought I'd add them. I'm not particularly attached to these particular names.

### DIFF
--- a/lib/list.g
+++ b/lib/list.g
@@ -875,3 +875,108 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
 ##  <#/GAPDoc>
 ##
 DeclareSynonym( "SubtractBlist", SUBTR_BLIST );
+
+#############################################################################
+##
+#F  FlipBlist( <blist> )
+##
+##  <#GAPDoc Label="FlipBlist">
+##  <ManSection>
+##  <Func Name="FlipBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> that equals <K>true</K> to <K>false</K>
+##  and vice versa. If <C>blist1</C> and <C>blist2</C> are boolean lists with
+##  equal length and every value in <C>blist2</C> is <K>true</K>,
+##  then <C>FlipBlist( blist1 )</C> is equivalent to
+##  <C>SubtractBlist( blist2, blist1 ); blist1 := blist2;</C>
+##  but <C>FlipBlist</C> is faster, and simpler to type.
+##  <P/>
+##  <Ref Func="FlipBlist"/> returns nothing, it is only called to change
+##  <A>blist</A> in-place.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> blist1 := [ true, true, true, true ];;
+##  gap> blist2 := [ true, false, true, false ];;
+##  gap> SubtractBlist( blist1, blist2 );
+##  gap> blist1;
+##  [ false, true, false, true ]
+##  gap> FlipBlist( blist2 );
+##  gap> blist2;
+##  [ false, true, false, true ]
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareSynonym( "FlipBlist", FLIP_BLIST );
+
+#############################################################################
+##
+#F  ClearAllBlist( <blist> )
+##
+##  <#GAPDoc Label="ClearAllBlist">
+##  <ManSection>
+##  <Func Name="ClearAllBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> to <K>false</K>.
+##  If <C>blist1</C> and <C>blist2</C> are boolean lists with
+##  equal length and every value in <C>blist2</C> is <K>false</K>,
+##  then <C>ClearAllBlist( blist1 )</C> is equivalent to
+##  <C>IntersectBlist( blist1, blist2 );</C> but is faster, and simpler to
+##  type.
+##  <P/>
+##  <Ref Func="ClearAllBlist"/> returns nothing, it is only called to change
+##  <A>blist</A> in-place.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> blist1 := [ true, true, true, true ];;
+##  gap> blist2 := [ true, false, true, false ];;
+##  gap> ClearAllBlist( blist1 );
+##  gap> blist1;
+##  [ false, false, false, false ]
+##  gap> ClearAllBlist(blist2);
+##  gap> blist2;
+##  [ false, false, false, false ]
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareSynonym( "ClearAllBlist", CLEAR_ALL_BLIST );
+
+#############################################################################
+##
+#F  SetAllBlist( <blist> )
+##
+##  <#GAPDoc Label="SetAllBlist">
+##  <ManSection>
+##  <Func Name="SetAllBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> to <K>true</K>.
+##  If <C>blist1</C> and <C>blist2</C> are boolean lists with
+##  equal length and every value in <C>blist2</C> is <K>true</K>,
+##  then <C>SetAllBlist( blist1 )</C> is equivalent to
+##  <C>UniteBlist( blist1, blist2 );</C> but is faster, and simpler to
+##  type.
+##  <P/>
+##  <Ref Func="SetAllBlist"/> returns nothing, it is only called to change
+##  <A>blist</A> in-place.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> blist1 := [ true, true, true, true ];;
+##  gap> blist2 := [ true, false, true, false ];;
+##  gap> SetAllBlist( blist1 );
+##  gap> blist1;
+##  [ true, true, true, true ]
+##  gap> SetAllBlist(blist2);
+##  gap> blist2;
+##  [ true, true, true, true ]
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareSynonym( "SetAllBlist", SET_ALL_BLIST );
--- a/src/blister.c
+++ b/src/blister.c
@@ -1577,6 +1577,102 @@ static Obj FuncMEET_BLIST(Obj self, Obj list1, Obj list2)
     return False;
 }
 
+/****************************************************************************
+**
+*F  FuncFLIP_BLIST( <self>, <list> ) . . .
+**
+**  'FuncFLIP_BLIST' implements the internal function 'FlipBlist'.
+**
+**  'FlipBlist( <list> )'
+**
+**  'FlipBlist' changes every value in the blist <list> from true to false, and
+**  vice versa.
+*/
+
+static Obj FuncFLIP_BLIST(Obj self, Obj list)
+{
+    // get and check the arguments
+    RequireBlist("FlipBlist", list, "blist");
+
+    if (LEN_BLIST(list) == 0) {
+        return 0;
+    }
+
+    UInt * ptr = BLOCKS_BLIST(list);
+    for (UInt i = NUMBER_BLOCKS_BLIST(list); 0 < i; i--) {
+        *ptr = ~(*ptr);
+        ptr++;
+    }
+    // If the logical length of the boolean list is not a multiple of BIPEB the
+    // last block will contain unused bits, which are then zero.
+    UInt mask =
+        ~(UInt)0 >> ((BIPEB * NUMBER_BLOCKS_BLIST(list)) - LEN_BLIST(list));
+    ptr = BLOCK_ELM_BLIST_PTR(list, LEN_BLIST(list));
+    *ptr &= mask;
+    return 0;
+}
+
+/****************************************************************************
+**
+*F  FuncCLEAR_ALL_BLIST( <self>, <list> ) . . .
+**
+**  'FuncCLEAR_ALL_BLIST' implements the internal function 'ClearAllBlist'.
+**
+**  'ClearAllBlist( <list> )'
+**
+**  'ClearAllBlist' changes every value in the blist <list> to false.
+*/
+
+static Obj FuncCLEAR_ALL_BLIST(Obj self, Obj list)
+{
+    // get and check the arguments
+    RequireBlist("ClearAllBitsBlist", list, "blist");
+
+    if (LEN_BLIST(list) == 0) {
+        return 0;
+    }
+
+    UInt * ptr = BLOCKS_BLIST(list);
+    for (UInt i = NUMBER_BLOCKS_BLIST(list); 0 < i; i--) {
+        *ptr++ = 0;
+    }
+
+    return 0;
+}
+
+/****************************************************************************
+**
+*F  FuncSET_ALL_BLIST( <self>, <list> ) . . .
+**
+**  'FuncSET_ALL_BLIST' implements the internal function 'SetAllBlist'.
+**
+**  'SetAllBlist( <list> )'
+**
+**  'SetAllBlist' changes every value in the blist <list> to true.
+*/
+
+static Obj FuncSET_ALL_BLIST(Obj self, Obj list)
+{
+    // get and check the arguments
+    RequireBlist("SetAllBitsBlist", list, "blist");
+
+    if (LEN_BLIST(list) == 0) {
+        return 0;
+    }
+
+    UInt * ptr = BLOCKS_BLIST(list);
+    for (UInt i = NUMBER_BLOCKS_BLIST(list); 0 < i; i--) {
+        *ptr++ = ~(UInt)0;
+    }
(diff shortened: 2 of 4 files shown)


lib/list.g
```diff
@@ -875,3 +875,107 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
...
+##  gap> blist2;
+##  [ false, false, false, false ]
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareSynonym( "SetAllBitsBlist", Set_ALL_BITS_BLIST );
+
+#############################################################################
```
ChrisJefferson: I'm guessing one of these should be "ClearAllBitsBlist"?

ChrisJefferson:
The kernel documentation for blister.c says that if there are unused bits in the last word they will be set to zero. This is required for blists to work properly. This means handling the last word in flip and set will require a special case.

lib/list.g
```diff
@@ -875,3 +875,107 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
...
+##
+#F  FlipBlist( <blist> )
+##
+##  <#GAPDoc Label="FlipBlist">
+##  <ManSection>
+##  <Func Name="FlipBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> that equals <K>true</K> to <K>false</K>
+##  and vice versa. If <C>blist1</C> and <C>blist2</C> are boolean list with
```
wilfwilson: list -> lists

lib/list.g
```diff
@@ -875,3 +875,107 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
...
+##
+#F  SetAllBitsBlist( <blist> )
+##
+##  <#GAPDoc Label="SetAllBitsBlist">
+##  <ManSection>
+##  <Func Name="SetAllBitsBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> to <K>false</K>.
+##  If <C>blist1</C> and <C>blist2</C> are boolean list with
```
wilfwilson: list -> lists

lib/list.g
```diff
@@ -875,3 +875,107 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
...
+##
+#F  SetAllBitsBlist( <blist> )
+##
+##  <#GAPDoc Label="SetAllBitsBlist">
+##  <ManSection>
+##  <Func Name="SetAllBitsBlist" Arg='blist'/>
+##
+##  <Description>
+##  Changes every entry in <A>blist</A> to <K>true</K>.
+##  If <C>blist1</C> and <C>blist2</C> are boolean list with
```
wilfwilson: list -> lists

lib/list.g
```diff
@@ -875,3 +875,107 @@ DeclareSynonym( "IntersectBlist", INTER_BLIST );
...
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareSynonym( "FlipBlist", FLIP_BLIST );
+
+#############################################################################
+##
+#F  SetAllBitsBlist( <blist> )
```
wilfwilson: As Chris has pointed out, this documentation and synonym declaration should be done all in terms of `ClearAllBits`.

lib/list.g
fingolfin: @wilfwilson you can also make code suggestions like so, which then the PR submitter can apply directly, thus saving some time (and often it also makes it clearer what is being suggested).
```suggestion
##  and vice versa. If <C>blist1</C> and <C>blist2</C> are boolean lists with
```

fingolfin:
Just to clarify: in general I support the idea of adding functions with the functionality described here. What I am objecting is putting "bits" into the function names; the names should "fit" with the existing blist API.

james-d-mitchell:
As I said in my original PR I'm not particularly attached to the names, any specific suggestions?

james-d-mitchell:
I don't really agree with your comments @fingolfin regarding the "fit" of these functions, but I do agree that the names don't fit.  I don't have the time to argue the point, nor do I think it will serve any purpose to do so. 

I've updated the PR to:

* set the "trailing" bits to 0, as per @ChrisJefferson 's comments.
* fix the documentation, and the synonyms in `list.g`
*  the names of the functions are `SetAllBlist`, `ClearAllBlist`, and `FlipBlist` i.e. no mention of bits. If you prefer different names, then please let me know what they are, and I'll be happy to update the PR accordingly.

I would find these 3 functions useful, which is why I made this PR, if you decide that you don't want them in GAP, then that's ok with me, please feel free to close this PR.

ChrisJefferson:
I wrote some more tests, I have cut+pasted them here, as that's probably the easiest thing to do. These currently crash. I think the problem is probably lists of length 0.

```
# FuncFLIP_BLIST
gap> FLIP_BLIST(fail);
Error, FlipBlist: <blist> must be a boolean list (not the value 'fail')
gap> x:= [false,true,true,false];;
gap> FLIP_BLIST(x);
gap> x;
[ true, false, false, true ]
gap> FLIP_BLIST(x);
gap> x;
[ false, true, true, false ]
gap> for i in [0..200] do
> f1 := List([1..i], x -> false);
> f2 := List([1..i], x -> false);
> t1 := List([1..i], x -> true);
> t2 := List([1..i], x -> true);
> FLIP_BLIST(f1); FLIP_BLIST(t1);
> if f1 <> t1 or f2 <> t2 then Print("Broken FLIP_BLIST\n"); fi;
> od;

# FuncSET_ALL_BLIST
gap> SET_ALL_BLIST(fail);
Error, SetAllBitsBlist: <blist> must be a boolean list (not the value 'fail')
gap> x:= [false,true,true,false];;
gap> SET_ALL_BLIST(x);
gap> x;
[ true, true, true, true ]
gap> SET_ALL_BLIST(x);
gap> x;
[ true, true, true, true ]
gap> for i in [0..200] do
> f1 := List([1..i], x -> false);
> t1 := List([1..i], x -> true);
> SET_ALL_BLIST(f1);
> if f1 <> t1 then Print("Broken SET_ALL_BLIST\n"); fi;
> od;

# FuncCLEAR_ALL_BLIST
gap> CLEAR_ALL_BLIST(fail);
Error, ClearAllBitsBlist: <blist> must be a boolean list (not the value 'fail'\
)
gap> x:= [false,true,true,false];;
gap> CLEAR_ALL_BLIST(x);
gap> x;
[ false, false, false, false ]
gap> CLEAR_ALL_BLIST(x);
gap> x;
[ false, false, false, false ]
gap> for i in [0..200] do
> f1 := List([1..i], x -> false);
> t1 := List([1..i], x -> true);
> CLEAR_ALL_BLIST(t1);
> if f1 <> t1 then Print("Broken CLEAR_ALL_BLIST\n"); fi;
> od;
```

james-d-mitchell:
Thanks @ChrisJefferson, I didn't account for the list of length 0 case. I don't think your tests are valid though, you flip `f1 := List([1..i], x -> false);` and `t1 := List([1..i], x -> true);`, so I wouldn't expect them to be the same. Did you mean to compare `f1` and `t2`, and `t1` and `f2`, instead?

ChrisJefferson:
Woops, yes, you are right :) That's what I was aiming for (I'm using the fact that I know that = compares the whole block of memory, to check we set/clear bits off the end correctly).

james-d-mitchell:
I've updated the PR to handle the 0 length blist case, and to add @ChrisJefferson tests.

james-d-mitchell:
I'm going to close this because there are no comments, or further requests for changes in the past 7 days.

ChrisJefferson:
Obviously, you can shut the PR (I won't force reopen it), but I did generally like the work. All my issues with it were fixed, but I know some other people had issues so I was waiting to see if they commented. If you reopen, I'll accept the PR and we can merge it (unless of course someone does still have an issue).

james-d-mitchell:
I got the impression that this isn't a desirable feature in the end, I'll re-open if I got the wrong impression. If it's not though, I'll move it into one of our packages, rather than leaving it as an open PR for longer.

wilfwilson:
The only person to say anything particularly critical was @fingolfin, and even then his clarification:
> Just to clarify: in general I support the idea of adding functions with the functionality described here.

was pretty clear to me. I had the impression that this would be merged, so I think you've got the wrong impression. I know it's annoying to make changes and wait a week and hear nothing, but that's the nature of GAP, where most contributors have to fit their work on GAP around (usually) much more pressing work issues.

james-d-mitchell:
@wilfwilson, I am not annoyed by waiting, nor by anything @fingolfin or @ChrisJefferson wrote. @fingolfin's first and second comments seem contradictory to me, so I wasn't sure what to think.

wilfwilson:
Fair enough, glad you weren't annoyed. @fingolfin has started at a new university this week, so it might be a bit longer before he'll comment further on this, if he intends to. I'll reopen it to give it some more time for comments, I hope you don't mind. In particular I'll most likely give it an 'Approving' review, if and when I have time 🙂

fingolfin:
I was indeed (and still am) super busy with the new job, where I also had to start teaching on the very first day, organize furniture, deal with tons of  bureaucracy and more.

And just to put things in perspective: I have plenty of examples where a PR of mine got no review at all for more than a month. Of course that's not ideal, but this is just what happens if all contributors essentially only can work on GAP as a side job. As such I am grateful that recently some more people started to help with reviews; before that, in the past year or so, it was mostly Chris and me who took care of the bulk of reviews.

src/blister.c
```diff
@@ -219,7 +219,7 @@ static Obj DoCopyBlist(Obj list, Int mut)
 
     /* return the copy                                                     */
     return copy;
-  
+
```
fingolfin: I've been told before on numerous occasions by @ChrisJefferson that pure white spaces changes are annoying and should be avoided. It seems he is fine with that here, and I'd normally not care about it myself, *but* here they really conflict with a WIP PR of mine, so it'd be nice if these could be removed from this PR. I'd be happy to perform the required changes with your permission (by pushing a commit with them to this PR), as this is trivial to do with a little git magic (using `git reset` plus `git add -p` resp. `git checkout -p` etc.). Of course you can also do them yourself.

fingolfin:
As to this:
> @fingolfin's first and second comments seem contradictory to me

I am not sure what to make of that: my second comment specifically stated that it was meant to clarify the first comment, in that I pointed out that I was objecting to the fit of the names, not the fit of functions (which I had indeed phrased a bit carelessly in my initial comment). So, to me, it's pretty clear that the second comment clarifies and hence supersedes the first. I really don't know how to make it more clear.

src/blister.c
```diff
@@ -1569,14 +1569,110 @@ static Obj FuncMEET_BLIST(Obj self, Obj list1, Obj list2)
...
+    // get and check the arguments
+    RequireBlist("SetAllBitsBlist", list, "blist");
+
+    if (LEN_BLIST(list) == 0) {
+        return 0;
+    }
+
+    UInt * ptr = BLOCKS_BLIST(list);
+    for (UInt i = NUMBER_BLOCKS_BLIST(list); 0 < i; i--) {
+        *ptr++ = (UInt)~0;
```
fingolfin: Very strictly and pedantically speaking, `(UInt)~0` (also elsewhere in this PR) should be `~(UInt)0` for full portability, although I assume all machines built today and probably also in the future will be using two's complement, so it probably doesn't matter... (some of the code in this file also uses `~0UL`).

Anyway, I won't complain if this is left as-is, we probably rely on two's complement elsewhere anyway *shrug*

tst/testinstall/kernel/blister.tst
```diff
@@ -209,5 +209,59 @@ true
...
+gap> x;
+[ true, false, false, true ]
+gap> FLIP_BLIST(x);
+gap> x;
+[ false, true, true, false ]
+gap> for i in [0..200] do
+> f1 := List([1..i], x -> false);
+> f2 := List([1..i], x -> false);
+> t1 := List([1..i], x -> true);
+> t2 := List([1..i], x -> true);
```
fingolfin: Note a request for changes, just a general remark: in real code, I'd always suggest to write `ListWithIdenticalEntries(i, true);` instead, which is optimized and directly produces a blist (unlike the above code, which produces a plist).

src/blister.c
james-d-mitchell: Ooops, this was a mistake. I'm happy to remove it.

src/blister.c
james-d-mitchell: Happy to change this also

tst/testinstall/kernel/blister.tst
james-d-mitchell: Yup, me too, I think it was @ChrisJefferson that wrote this particular line.

src/blister.c
fingolfin: Thank you! Please post a comment once you are done updating this, so we'll see it -- as you know, unfortunately (or fortunately? ;-)it'd be nice if at least one could choose...) GitHub does not send notifications when a PR is updated.

james-d-mitchell:
> As to this:
> 
> > @fingolfin's first and second comments seem contradictory to me
> 
> I am not sure what to make of that: my second comment specifically stated that it was meant to clarify the first comment, in that I pointed out that I was objecting to the fit of the names, not the fit of functions (which I had indeed phrased a bit carelessly in my initial comment). So, to me, it's pretty clear that the second comment clarifies and hence supersedes the first. I really don't know how to make it more clear.

It wasn't clear to me if the names changes were what you were aiming at, or if you would only accept a version of the method along the lines of the `SetBlistEntries` that you wrote about in your first comment, or indeed if there were further issues with the PR.  I thought there was a recent request to close PR/issues without comments after some time, but apparently I got that wrong.

tst/testinstall/kernel/blister.tst
ChrisJefferson: I keep forgetting about ListWithIdenticalEntries, there is no good reason not to use it.

src/blister.c
ChrisJefferson: Weirdly I didn't see them when I reviewed it, maybe they got added later. They do generally annoy me, just because they make reviewing harder

wilfwilson:
> I thought there was a recent request to close PR/issues without comments after some time, but apparently I got that wrong.

On this topic:

At the GAP days in Halle, a group of us spoke about ways that we could manage issues and PRs more effectively. The aim was to make the job easier for those who review PRs and attempt to solve issues, rather than adding hoops for PR/issue creators to jump through. I've been writing up something about this to help contributors, but I'm not finished yet.

The main thrust is to use labels more effectively, which seems to be going well in my opinion. There's no particular aim to close issues or PRs (beyond actually resolving them/merging them in the proper way), however we have introduced a small measure along these lines:

Since #3366 was merged, issues/PRs that are given the label `status: awaiting response` will be automatically closed after 14 days if the original author does not reply. The idea is to close issues/PRs where there is not enough information from the author to proceed, and where the author is not forthcoming with the required information (perhaps 14 days should be increased?). I think this is a fairly small use case, and so far nothing has been closed in this way (although #2386 is close).

src/blister.c
wilfwilson: Perhaps you have "Hide whitespace changes" selected in the diff settings when reviewing a PR (and if you're not using this option, maybe using it will help you get less annoyed!)

ChrisJefferson:
Sorry to kick up one more problem!

This isn't checking the blists are mutable. You could either fix this with some code a bit like:

```
 if (!IS_MUTABLE_OBJ(list1))
        ErrorMayQuit("Append: <list1> must be a mutable", 0, 0);
```

Or, wait until (hopefully)  #3391 is merged, at which point you will be able to write:

`RequireMutable(list1);`

fingolfin:
@ChrisJefferson good catch on the mutability -- but to be fair, most other functions in this file also lack this :-/, e.g. `FuncUNITE_BLIST`. So I'd also be fine with accepting this "bug" in this PR for now, and then follow it up soon with another PR which adds the missing `RequireMutable` calls to all `blister.c` functions, together with test cases. And then we put that as a "wrong result" bug fix into the release notes, or so.

Here one of the instances of this bug in live action:
```
gap> l1:=MakeImmutable(ListWithIdenticalEntries(10,true));
[ true, true, true, true, true, true, true, true, true, true ]
gap> l2:=MakeImmutable(ListWithIdenticalEntries(10,false));
[ false, false, false, false, false, false, false, false, false, false ]
gap> UniteBlist(l2, l1);
gap> l2;
[ true, true, true, true, true, true, true, true, true, true ]
```

ChrisJefferson:
I agree, ignire the mutablely issue, we will do a general cleanup later.

ChrisJefferson:
Just for clarity, I'm happy to merge this as soon as @james-d-mitchell tells me he's made any final changes he wants to make (no rush, and no requirement for changes, that's just why I'm not pressing merge now)

james-d-mitchell:
I believe I've addressed all of the issues (except the mutability one), please let me know if there is anything else.

james-d-mitchell:
Thanks @wilfwilson, @fingolfin and @ChrisJefferson  for the comments, and apologies for the confusion!

PR #3390, opened 2019-04-02
Fix Bug in 64Bit SetCyclotomicsLimit

### LINKED ISSUE #3389: Bug in SetCyclotomicsLimit
Benjamin Sambale reported the following issue on the GAP support mailing list: On 64 bit systems, `SetCyclotomicsLimit` accepts `2^32` as input, but this is then internally put into a 32bit integer, which ends up as equal to zero, leaving GAP in an inconsistent state.

Moreover, the manual section 18.6 says:
> The maximum value of the limit is 2^28-1 on 32 bit systems, and 2^32 on 64 bit systems.

Here is the bug in action:
```
gap> SetCyclotomicsLimit(2^32-1); GetCyclotomicsLimit();
4294967295
gap> SetCyclotomicsLimit(2^32); GetCyclotomicsLimit();
0
gap>
gap> SetCyclotomicsLimit(2^32+1); GetCyclotomicsLimit();
Error, Cyclotomic field size limit must be less than 2^32
```

The error message in the final example says how it should be. So, we need to (a) fix the range check in `FuncSetCyclotomicsLimit`, and (b) fix the documentation.

DominikBernhardt:
Fixes #3389.

### DIFF
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -854,7 +854,7 @@ static Obj FuncSetCyclotomicsLimit(Obj self, Obj newlimit)
                      CyclotomicsLimit, 0);
     }
 #ifdef SYS_IS_64_BIT
-    if (ulimit > (1L << 32)) {
+    if (ulimit >= (1L << 32)) {
         ErrorMayQuit("Cyclotomic field size limit must be less than 2^32", 0,
                      0);
     }
--- a/tst/testinstall/cyclotom.tst
+++ b/tst/testinstall/cyclotom.tst
@@ -276,6 +276,10 @@ he integer 0)
 gap> SetCyclotomicsLimit(100);
 Error, SetCyclotomicsLimit: <newlimit> must not be less than old limit of 1000\
 000
+#@if 8*GAPInfo.BytesPerVariable = 64
+gap> SetCyclotomicsLimit(2^32);
+Error, Cyclotomic field size limit must be less than 2^32
+#@fi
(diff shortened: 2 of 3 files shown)


src/cyclotom.c
```diff
@@ -854,7 +854,7 @@ static Obj FuncSetCyclotomicsLimit(Obj self, Obj newlimit)
                      CyclotomicsLimit, 0);
     }
 #ifdef SYS_IS_64_BIT
-    if (ulimit > (1L << 32)) {
+    if (ulimit >= (1L << 32)) {
```
fingolfin: Could you please also add a test case, in `tst/testinstall/cyclotom.tst` around line 279 would be fitting

tst/testinstall/cyclotom.tst
```diff
@@ -276,6 +276,8 @@ he integer 0)
 gap> SetCyclotomicsLimit(100);
 Error, SetCyclotomicsLimit: <newlimit> must not be less than old limit of 1000\
 000
+gap> SetCyclotomicsLimit(2^32);
+Error, Cyclotomic field size limit must be less than 2^32
```
fingolfin: Thanks. Unfortunately, this now fails on 32bit systems. To solve that, you can use `#@if 8*GAPInfo.BytesPerVariable = 64`... See for example `tst/test32bit/2017-06-20-restrictedperm.tst`  (sorry, I should have thought about that and warned you)

PR #3391, opened 2019-04-06
Add RequireMutable

ChrisJefferson:
This adds the C function "RequireMutable".

This took more work than I planned, as I found errors of the type:

"<list> must be mutable (not a ...)" a bit confusing, as for this error I found the "not a ..." bits misleading.

### DIFF
--- a/src/error.h
+++ b/src/error.h
@@ -239,6 +239,14 @@ Obj RequireArgumentEx(const char * funcname,
     RequireArgumentCondition(funcname, op, op == True || op == False,        \
                              "must be true or false")
 
+/****************************************************************************
+**
+*F  RequireMutable
+*/
+#define RequireMutable(funcname, op, type)                                   \
+    RequireArgumentCondition(funcname, op, IS_MUTABLE_OBJ(op),               \
+                             "must be a mutable " type)
+
 
 /****************************************************************************
 **
--- a/src/listfunc.c
+++ b/src/listfunc.c
@@ -221,8 +221,7 @@ static Obj FuncAPPEND_LIST_INTR(Obj self, Obj list1, Obj list2)
     Int                 i;              /* loop variable                   */
 
     /* check the mutability of the first argument */
-    if (!IS_MUTABLE_OBJ(list1))
-        ErrorMayQuit("Append: <list1> must be a mutable list", 0, 0);
+    RequireMutable("Append", list1, "list");
 
 
     /* handle the case of strings now */
--- a/src/vecgf2.c
+++ b/src/vecgf2.c
@@ -1845,11 +1845,7 @@ static Obj FuncELMS_GF2VEC(Obj self, Obj list, Obj poss)
 static Obj FuncASS_GF2VEC(Obj self, Obj list, Obj pos, Obj elm)
 {
     // check that <list> is mutable
-    if (!IS_MUTABLE_OBJ(list)) {
-        ErrorReturnVoid("List Assignment: <list> must be a mutable list", 0L,
-                        0L, "you can 'return;' and ignore the assignment");
-        return 0;
-    }
+    RequireMutable("List Assignment", list, "list");
 
     // get the position
     UInt p = GetSmallInt("ASS_GF2VEC", pos);
@@ -1913,11 +1909,7 @@ static Obj FuncPLAIN_GF2MAT(Obj self, Obj list)
 static Obj FuncASS_GF2MAT(Obj self, Obj list, Obj pos, Obj elm)
 {
     // check that <list> is mutable
-    if (!IS_MUTABLE_OBJ(list)) {
-        ErrorReturnVoid("List Assignment: <list> must be a mutable list", 0L,
-                        0L, "you can 'return;' and ignore the assignment");
-        return 0;
-    }
+    RequireMutable("List Assignment", list, "list");
 
     // get the position
     UInt p = GetSmallInt("ASS_GF2MAT", pos);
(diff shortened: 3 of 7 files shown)


src/vecgf2.c
```diff
@@ -1845,11 +1845,7 @@ static Obj FuncELMS_GF2VEC(Obj self, Obj list, Obj poss)
 static Obj FuncASS_GF2VEC(Obj self, Obj list, Obj pos, Obj elm)
 {
     // check that <list> is mutable
-    if (!IS_MUTABLE_OBJ(list)) {
-        ErrorReturnVoid("List Assignment: <list> must be a mutable list", 0L,
```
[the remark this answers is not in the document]
fingolfin: We agreed some time ago (also on the gap mailing list) that we essentially want to remove all uses of `ErrorReturnVoid` and `ErrorReturnObj`.

src/error.h
```diff
@@ -100,11 +100,22 @@ void ErrorReturnVoid(const Char * msg, Int arg1, Int arg2, const Char * msg2);
 **
 **  If funcname is 0, then 'funcname: ' is omitted from the message.
 **  If argname is 0, then '<argname> ' is omitted from the message.
+**
+**  RequireArgumentExNoArgType is the same as RequireArgumentEx, except
+**  it omits '(not a %s').
```
fingolfin:
```suggestion
**  it omits '(not a %s)' from the message.
```

src/listfunc.c
```diff
@@ -221,8 +221,7 @@ static Obj FuncAPPEND_LIST_INTR(Obj self, Obj list1, Obj list2)
     Int                 i;              /* loop variable                   */
 
     /* check the mutability of the first argument */
-    if (!IS_MUTABLE_OBJ(list1))
-        ErrorMayQuit("Append: <list1> must be a mutable list", 0, 0);
+    RequireMutable("Append", list1);
```
fingolfin: As @wilfwilson pointed out, it's a bit sad that the conveyed information ("mutable *list*") is reduced. Perhaps if there was another optional argument to `RequireMutable`, like so:  `RequireMutable("Append", list1, "list");` to convey the "type"?

To implement that, I'd replace `Int showArgType` by `const char * alternateArgType`; if `alternateArgType` is NULL, then the usual `(not a %s)` is shown; otherwise `alternateArgType` is used instead (perhaps as another special case, if `alternateArgType` is an empty string, then also omit the space normally inserted before it).

src/error.c
```diff
@@ -579,19 +580,22 @@ Obj RequireArgumentEx(const char * funcname,
...
-        strlcat(msgbuf, " (not the value 'true')", sizeof(msgbuf));
-    else if (op == False)
-        strlcat(msgbuf, " (not the value 'false')", sizeof(msgbuf));
-    else if (op == Fail)
-        strlcat(msgbuf, " (not the value 'fail')", sizeof(msgbuf));
-    else {
-        strlcat(msgbuf, " (not a %s)", sizeof(msgbuf));
-        arg1 = (Int)TNAM_OBJ(op);
+
+    if (showArgType) {
```
fingolfin: With my `alternateArgType` from below, this could be replaced by the following (and no code would need to be reindented):
```C
    if (alternateArgType) {
        if (*alternateArgType) { // not the empty string?
            strlcat(msgbuf, " ", sizeof(msgbuf));
            strlcat(msgbuf, alternateArgType, sizeof(msgbuf));
        }
    } else if (IS_INTOBJ(op)) {
...
```

src/listfunc.c
fingolfin: Thinking some more about it, of course just providing `"list"` there is not enough;  so I think instead, an invocation could like this: `RequireMutable("Append", list1, "a mutable list");`, and `showArgType` could be left as-is...

src/vecgf2.c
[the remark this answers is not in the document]
wilfwilson: Thanks for the information.

src/error.h
fingolfin: At least the typo here should be fixed

wilfwilson: Oh sorry, yes, I agree.

src/error.c
ChrisJefferson: I'll have another go. If I keep the full type (i.e. mutable list) , then also keeping the (not a...) becomes more reasonable too.

ChrisJefferson:
This patch got massively simplified :) But I think the messages are much better. Thanks for all feedback.

src/listfunc.c
ChrisJefferson: I went with this kind of route in the end. For simplicity I assumed we will always start the message "a mutable ", but that could of course be changed later if ever needed. Sorry for the long new function, which ended up getting removed!

src/vecgf2.c
```diff
@@ -2027,11 +2015,7 @@ static Obj FuncUNB_GF2VEC(Obj self, Obj list, Obj pos)
 static Obj FuncUNB_GF2MAT(Obj self, Obj list, Obj pos)
 {
     // check that <list> is mutable
-    if (!IS_MUTABLE_OBJ(list)) {
-        ErrorReturnVoid("List Unbind: <list> must be a mutable list", 0L, 0L,
-                        "you can 'return;' and ignore the unbind");
-        return 0;
-    }
+    RequireMutable("List Unbind", list, "list");
```
fingolfin: Perhaps "matrix" instead of "list"?

PR #3392, opened 2019-04-06
Prevent blist functions from modifying immutable blists (and several other kernel tweaks leading up to this)

### LINKED ISSUE #3393: Some functions in blister.c ignore mutability
_(Resolving this issue should wait until #3385 and #3391  are merged, so that the whole file can be done at once by adding the appropriate `RequireMutable` calls.)_

As pointed out initially by @fingolfin in https://github.com/gap-system/gap/pull/3385#issuecomment-480506296:

> ...most other functions in [blister.c] also lack [mutability checks] :-/, e.g. FuncUNITE_BLIST. So I'd also be fine with accepting this "bug" in this PR for now, and then follow it up soon with another PR which adds the missing RequireMutable calls to all blister.c functions, together with test cases. And then we put that as a "wrong result" bug fix into the release notes, or so.

> Here one of the instances of this bug in live action:
```
gap> l1:=MakeImmutable(ListWithIdenticalEntries(10,true));
[ true, true, true, true, true, true, true, true, true, true ]
gap> l2:=MakeImmutable(ListWithIdenticalEntries(10,false));
[ false, false, false, false, false, false, false, false, false, false ]
gap> UniteBlist(l2, l1);
gap> l2;
[ true, true, true, true, true, true, true, true, true, true ]
```

fingolfin:
This conflicts with PR #3385, so that PR should be merged first; after which I can rebase and update this PR.

But I wanted to put it out now, to make sure I don't leave it bitrotting here for another half year ;-).

Resolves #3393

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1181,32 +1181,32 @@ static Obj FuncPositionNthTrueBlist(
 
 /****************************************************************************
 **
-*F  FuncIsSubsetBlist( <self>, <list1>, <list2> ) . . . . . . . . subset test
+*F  FuncIsSubsetBlist( <self>, <blist1>, <blist2> ) . . . . . . . subset test
 **
 **  'FuncIsSubsetBlist' implements the internal function 'IsSubsetBlist'.
 **
-**  'IsSubsetBlist( <list1>, <list2> )'
+**  'IsSubsetBlist( <blist1>, <blist2> )'
 **
-**  'IsSubsetBlist' returns 'true' if the boolean list <list2> is a subset of
-**  the  boolean  list <list1>, which must  have  equal length.  <list2> is a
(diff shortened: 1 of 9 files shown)


wilfwilson:
I'll add the `do not merge` label and the new `status: awaiting a different PR` label until #3385 is merged. Once that happens and you've updated, I'll be happy to review again if you want.

wilfwilson:
#3385 is now merged, so there's nothing stopping this PR from taking its purest form 🙂

fingolfin:
I have now rebased this PR. I am not sure about the "purest form", though, it now also adds the mutability checks... I tried to break everything into "clean" commits, but let me know if you think it is too much and I should split it into two or more PRs.

PR #3394, opened 2019-04-07
Give the plist TNUMS more human-readable names

ChrisJefferson:
We are increasingly printing out the TNUMs of types, and I had someone at GAP days ask me what "list (plain, nhom)" meant. A reasonable question!

This replaces the names with longer human-readable names. I'm happy for tweaks to the names (I find non-strictly-sorted long. I thought about writing "set" in that case, or non-strictly sorted?). This does make them much longer.

I notice we seem to use both homogeneous and homogenous, I went with the first, as it is (I believe) correct, and we use it much more.

### DIFF
--- a/src/plist.c
+++ b/src/plist.c
@@ -2542,65 +2542,65 @@ static Obj FuncIsRectangularTablePlist(Obj self, Obj plist)
 *V  BagNames  . . . . . . . . . . . . . . . . . . . . . . . list of bag names
 */
 static StructBagNames BagNames[] = {
-  { T_PLIST,                                "list (plain)" },
-  { T_PLIST            +IMMUTABLE,          "list (plain,imm)" },
+  { T_PLIST,                                "plain list" },
+  { T_PLIST            +IMMUTABLE,          "immutable plain list" },
 
-  { T_PLIST_NDENSE,                         "list (plain,ndense)" },
-  { T_PLIST_NDENSE     +IMMUTABLE,          "list (plain,ndense,imm)" },
+  { T_PLIST_NDENSE,                         "non-dense plain list" },
+  { T_PLIST_NDENSE     +IMMUTABLE,          "immutable non-dense plain list" },
 
-  { T_PLIST_DENSE,                          "list (plain,dense)" },
-  { T_PLIST_DENSE      +IMMUTABLE,          "list (plain,dense,imm)" },
+  { T_PLIST_DENSE,                          "dense plain list" },
+  { T_PLIST_DENSE      +IMMUTABLE,          "immutable dense plain list" },
 
-  { T_PLIST_DENSE_NHOM,                     "list (plain,dense,nhom)" },
-  { T_PLIST_DENSE_NHOM +IMMUTABLE,          "list (plain,dense,nhom,imm)" },
+  { T_PLIST_DENSE_NHOM,                     "dense non-homogeneous plain list" },
+  { T_PLIST_DENSE_NHOM +IMMUTABLE,          "immutable dense non-homogeneous plain list" },
 
-  { T_PLIST_DENSE_NHOM_SSORT,               "list (plain,dense,nhom,ssort)" },
-  { T_PLIST_DENSE_NHOM_SSORT +IMMUTABLE,    "list (plain,dense,nhom,ssort,imm)" },
+  { T_PLIST_DENSE_NHOM_SSORT,               "dense non-homogeneous strictly-sorted plain list" },
+  { T_PLIST_DENSE_NHOM_SSORT +IMMUTABLE,    "immutable dense non-homogeneous strictly-sorted plain list" },
 
-  { T_PLIST_DENSE_NHOM_NSORT,               "list (plain,dense,nhom,nsort)" },
-  { T_PLIST_DENSE_NHOM_NSORT +IMMUTABLE,    "list (plain,dense,nhom,nsort,imm)" },
+  { T_PLIST_DENSE_NHOM_NSORT,               "dense non-homogeneous non-strictly-sorted plain list" },
+  { T_PLIST_DENSE_NHOM_NSORT +IMMUTABLE,    "immutable dense non-homogeneous non-strictly-sorted plain list" },
 
-  { T_PLIST_EMPTY,                          "list (plain,empty)" },
-  { T_PLIST_EMPTY      +IMMUTABLE,          "list (plain,empty,imm)" },
+  { T_PLIST_EMPTY,                          "empty plain list" },
+  { T_PLIST_EMPTY      +IMMUTABLE,          "immutable empty plain list" },
 
-  { T_PLIST_HOM,                            "list (plain,hom)" },
-  { T_PLIST_HOM        +IMMUTABLE,          "list (plain,hom,imm)" },
+  { T_PLIST_HOM,                            "homogeneous plain list" },
+  { T_PLIST_HOM        +IMMUTABLE,          "immutable homogeneous plain list" },
 
-  { T_PLIST_HOM_NSORT,                      "list (plain,hom,nsort)" },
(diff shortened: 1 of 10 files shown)


src/plist.c
```diff
@@ -2542,65 +2542,65 @@ static Obj FuncIsRectangularTablePlist(Obj self, Obj plist)
...
+  { T_PLIST_TAB_NSORT +IMMUTABLE,           "non-strictly-sorted immutable plain list (table)" },
 
-  { T_PLIST_TAB_SSORT,                      "list (plain,table,ssort)" },
-  { T_PLIST_TAB_SSORT +IMMUTABLE,           "list (plain,table,ssort,imm)" },
+  { T_PLIST_TAB_SSORT,                      "strictly-sorted plain list (table)" },
+  { T_PLIST_TAB_SSORT +IMMUTABLE,           "strictly-sorted immutable plain list (table)" },
 
-  { T_PLIST_TAB_RECT,                       "list (plain,rect table)" },
-  { T_PLIST_TAB_RECT       +IMMUTABLE,      "list (plain,rect table,imm)" },
+  { T_PLIST_TAB_RECT,                       "plain list (rect table)" },
```
wilfwilson: What are anyone's thoughts about even expanding `rect` to `rectangular`? Since everything else is being unabbreviated, perhaps this should be too. I think I'd vote for that anyway.

src/plist.c
```diff
@@ -2542,65 +2542,65 @@ static Obj FuncIsRectangularTablePlist(Obj self, Obj plist)
...
+  { T_PLIST_TAB_RECT       +IMMUTABLE,      "immutable plain list (rect table)" },
 
-  { T_PLIST_TAB_RECT_NSORT,                 "list (plain,rect table,nsort)" },
-  { T_PLIST_TAB_RECT_NSORT +IMMUTABLE,      "list (plain,rect table,nsort,imm)" },
+  { T_PLIST_TAB_RECT_NSORT,                 "non-strictly-sorted plain list (rect table)" },
+  { T_PLIST_TAB_RECT_NSORT +IMMUTABLE,      "non-strictly-sorted immutable plain list (rect table)" },
```
wilfwilson: Question for my own curiosity: does `non-strictly-sorted` mean that the list is definitely not strictly sorted, or just that it's not known to be strictly sorted?

tst/testinstall/bitfields.tst
```diff
@@ -60,7 +60,7 @@ Error, Field getter: <data> must be a small integer (not a ffe)
 gap> bf.setters[1](1, (1,2));
 Error, Field Setter: <val> must be a small integer (not a permutation (small))
 gap> bf.setters[1]([],1);
-Error, Field Setter: <data> must be a small integer (not a list (plain,empty))
+Error, Field Setter: <data> must be a small integer (not a empty plain list)
```
wilfwilson: And here is the unfortunate side effect of your nice changes. "not a empty..." (or "not a immutable..." or whatever), when the correct English would be "not an empty..." (or "not an immutable..."). As an English speaker, this is something that mildly upsets me, but I can live with this. Do we try to make sure that indefinite articles (a/an) are correct in GAP? I'd guess not. I can imagine that it would be a pain to make this always work properly, so I'm not going to suggest that we should.

src/plist.c
ChrisJefferson: I was thinking about the table and rect cases, wasn't sure how to handle them most nicely. They should be adjusted however.

src/plist.c
ChrisJefferson: Yes, lists can be in one of 3 states: known sorted, known non-sorted, or no comment one way or the other (which means unknown).

tst/testinstall/bitfields.tst
ChrisJefferson: It probably wouldn't be that hard to fix :) We could have an investigate.

ChrisJefferson:
I meant to give a consistent ordering to the words, but obviously slipped up. I've moved all the mutables to the front, and I've ordered the other things by their order in the TNUM name, as it seems like an easy way to get a consistent ordering.

tst/testinstall/bitfields.tst
fingolfin: Actually, making this right was (for me) one of the driving motivations (yes, seriously :-) for introducing those `Require*` helpers (changed `not a integer` to `not an integer` and then later `not the integer %d`). So, I'd really want to improve this, at least when using the `Require` macros. One trick option here would be to modify `RequireArgumentEx`. Like this:
```patch
diff --git a/src/error.c b/src/error.c
index ff156757a..f7b071f35 100644
--- a/src/error.c
+++ b/src/error.c
@@ -590,8 +590,19 @@ Obj RequireArgumentEx(const char * funcname,
     else if (op == Fail)
         strlcat(msgbuf, " (not the value 'fail')", sizeof(msgbuf));
     else {
-        strlcat(msgbuf, " (not a %s)", sizeof(msgbuf));
-        arg1 = (Int)TNAM_OBJ(op);
+        const char * tnam = TNAM_OBJ(op);
+        // check for starting vowel
+        // TODO: move this into a separate helper function?
+        // TODO: improve this, and e.g. use "a" before "un"; "an hour"; ...
+        //  see also <https://github.com/rossmeissl/indefinite_article/blob/master/lib/indefinite_article.rb>
+        //  alternatively, could enhance the TNAM records to (optionally) allow indicating whether 'an' or 'a'
+        //  should be used....
+        if (strchr("aeiou", tnam[0]))
+            strlcat(msgbuf, " (not an", sizeof(msgbuf));
+        else
+            strlcat(msgbuf, " (not a", sizeof(msgbuf));
+        strlcat(msgbuf, tnam, sizeof(msgbuf));
+        strlcat(msgbuf, ")", sizeof(msgbuf));
     }
 
     ErrorMayQuit(msgbuf, arg1, arg2);
```

src/plist.c
```diff
@@ -2542,65 +2542,65 @@ static Obj FuncIsRectangularTablePlist(Obj self, Obj plist)
 *V  BagNames  . . . . . . . . . . . . . . . . . . . . . . . list of bag names
 */
 static StructBagNames BagNames[] = {
-  { T_PLIST,                                "list (plain)" },
-  { T_PLIST            +IMMUTABLE,          "list (plain,imm)" },
+  { T_PLIST,                                "plain list" },
+  { T_PLIST            +IMMUTABLE,          "immutable plain list" },
```
fingolfin: Not very important, but: I kinda disagree with the commit message / title which refers to "more human-readable names". To me, they are equally readable; the difference is in how "understandable" / "intelligible" / "comprehensible" they are.

src/plist.c
```diff
@@ -2542,65 +2542,65 @@ static Obj FuncIsRectangularTablePlist(Obj self, Obj plist)
...
 
-  { T_PLIST_DENSE,                          "list (plain,dense)" },
-  { T_PLIST_DENSE      +IMMUTABLE,          "list (plain,dense,imm)" },
+  { T_PLIST_DENSE,                          "dense plain list" },
+  { T_PLIST_DENSE      +IMMUTABLE,          "immutable dense plain list" },
 
-  { T_PLIST_DENSE_NHOM,                     "list (plain,dense,nhom)" },
-  { T_PLIST_DENSE_NHOM +IMMUTABLE,          "list (plain,dense,nhom,imm)" },
+  { T_PLIST_DENSE_NHOM,                     "dense non-homogeneous plain list" },
+  { T_PLIST_DENSE_NHOM +IMMUTABLE,          "immutable dense non-homogeneous plain list" },
```
fingolfin: Re "homogeneous" versus "homogenous": both are correct. But we use the former 657 and the latter just 29 times in the GAP repository (according to `git grep -i`), so I support this change, and am tempted to follow it up with another commit / PR that replaces those 29 uses of "homogenous".

fingolfin:
Seems not all test files were updated, there are diffs in `tst/testinstall/list.tst` and `tst/testinstall/listindex`

PR #3395, opened 2019-04-08
Fix 'PrintObj(1.)'

### LINKED ISSUE #3369: `PrintObj(1.)` prints "1" instead of "1." or "1.0"
This is in current GAP master:
```
gap> 1.;
1.
gap> ViewObj(1.);
1.
gap> PrintObj(1.);
1
```

DominikBernhardt:
Fixes #3369 .

I received help for this PR from @sebasguts and @fingolfin


src/macfloat.c
```diff
@@ -70,6 +70,10 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
     }
     else {
         snprintf(buf, bufsize, "%.*" PRINTFFORMAT, precision, val);
+        if(!strchr(buf, '.'))
```
fingolfin: Please, if possible use `git clang-format` on your pull request. If you need help with that, let me know. Alternatively, just adjust to the "desired" formatting:
```suggestion
        if (!strchr(buf, '.')) {
```

src/macfloat.c
```diff
@@ -70,6 +70,10 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
...
+        {
```
fingolfin:
```suggestion
```

src/macfloat.c
```diff
@@ -70,6 +70,10 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
...
+          strcat(buf, ".");
```
fingolfin: Please do not use `strcat`, it is a potential source of buffer overflows. use `strlcat` or `strxcat` instead.

tst/testinstall/float.tst
```diff
@@ -179,12 +179,12 @@ gap> 1.5e10;
 gap> -1.5e0;
 -1.5
 gap> 0.7e-10;
-7.e-11
+7e-11.
```
fingolfin: Hmmm, that's not a valid floating point :/. 
```
gap> 7e14;
Error, Variable: '7e14' must have a value
not in any function at *stdin*:14
gap> 7e14.;
Error, Variable: '7e14' must have a value
not in any function at *stdin*:15
Syntax error: Record component name expected
7e14.;
     ^
gap> 7.e14;
7.e+14
```

I guess that means the logic in your patch needs to be a bit more complicated. I guess looking at `MACFLOAT_STRING_DOTTIFY` to see what it does, and then adapt it in the C code, would do the trick.

src/macfloat.c
DominikBernhardt: Sure, I will stick to that.

src/macfloat.c
DominikBernhardt: Changed it.

fingolfin:
For the record, I verified that no distributed package is using `MACFLOAT_STRING_DOTTIFY`

src/macfloat.c
```diff
@@ -70,6 +71,16 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
     }
     else {
         snprintf(buf, bufsize, "%.*" PRINTFFORMAT, precision, val);
+        if (strlen(buf) < bufsize) {
```
fingolfin: This check is incorrect. Correct one follows, together with a comment that explains it (a good idea in such delicate code anyway).
```suggestion
        // check if a period is in the output; this is not always the case,
        // e.g. if the value is an integer
        if (strchr(buf, '.'))
            return; // everything is fine
        // we need to insert a '.'; either at the end, or before an exponent
        // (e.g. "7e10" -> "7.e10"). For this we need 1 extra byte of storage,
        // plus of course 1 byte for the string terminator; check if the
        // buffer is big enough
        if (strlen(buf) + 2 <= bufsize) {
```

fingolfin: Of course in reality, our buffers are 1024 bytes and won't overflow, but I always prefer to code defensively... and better no bounds check than an incorrect one... of course even better to have a correct one ;-)

src/macfloat.c
```diff
@@ -70,6 +71,16 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
...
+            char * loc = strchr(buf, 'e');
+            if (loc && !strchr(buf, '.')) {
+                memmove(loc + 1, loc, strlen(loc) + 1);
+                loc[0] = '.';
+            }
+            else if (!strchr(buf, '.')) {
```
fingolfin:
```suggestion
            else {
```

src/macfloat.c
```diff
@@ -70,6 +71,16 @@ static void PrintMacfloatToBuf(char *buf, size_t bufsize, Double val, int precis
     }
     else {
         snprintf(buf, bufsize, "%.*" PRINTFFORMAT, precision, val);
+        if (strlen(buf) < bufsize) {
+            char * loc = strchr(buf, 'e');
+            if (loc && !strchr(buf, '.')) {
```
fingolfin:
```suggestion
            if (loc) {
```

src/macfloat.c
sebasguts: No, that if is correct. The string might be `1.7`

fingolfin: But then we never get here, because we already check for the `.` a few lines above (with my suggested change)

sebasguts: Argh, I see now, okay :). I was confused by the long change, and thought it was a comment.

PR #3399, opened 2019-04-10
Allow floating point values in memory command line options

ChrisJefferson:
So, I always assumed that `-o 2.5G` did what I expected (allocated 2.5GB of memory), but turns out the `.5` was silently ignored. Switch to using `atof` so we parse a floating point number.

We still silently ignore badly written constants, but I think this is a useful improvement.

Release notes suggestion:

* Support floating point numbers when specifying how much memory GAP should use, for example "-o 2.5G".

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -868,7 +868,7 @@ static sizeMultiplier memoryUnits[]= {
 
 static UInt ParseMemory( Char * s)
 {
-  UInt size  = atol(s);
+  double size = atof(s);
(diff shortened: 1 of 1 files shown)


fingolfin:
Should we also document this?

ChrisJefferson:
I felt this was already "documented", in that the docs don't say you can't give a floating point number, and personally I've been writing things like `-o 2.5G` for ages, not realising the .5 it wasn't doing anything useful.

PR #3404, opened 2019-04-11
RFC: kernel: remove FUNC_CALL, ARGI_{CALL,INFO}, NARG_SIZE_{CALL,INFO}, SIZE_NARG_{CALL,INFO}

fingolfin:
This is a subset of PR #3257, and quite possibly controversial. After all, why remove access helper functions and replace them with raw access? Well, the immediate reason in PR #3257 is the following: Right now, any access to function bodies (i.e., `T_BODY` kernel objects), be it to write them (during code), execute them, print them, or turn them into a GAP readable form via `SyntaxTree`, requires that one sets the global variable `STATE(PtrBody)` to point to the `T_BODY` (via `SWITCH_TO_NEW_LVARS`); then perform the access; then restore`STATE(PtrBody)` (via `SWITCH_TO_OLD_LVARS`). This is rather awkward. So, why is it done? Well, for the *executor*, this is an optimization: it allows us to avoids lots of calls to `ADDR_OBJ`/`PTR_BAG`. That the other kinds of access go through it, though, is not necessary at all; and so my goal is to change this. The hope is that ultimately, the code accessing the `T_BODY` objects will get simpler and thus easier to maintain, and to learn about for people interested in them. And at the same time, it might enable further optimizations for the executor (for starters, we then could turn `STATE(PtrBody)` into a const pointer).

The problem now is that the new code, which does not set `STATE(PtrBody)`, cannot use `ARGI_CALL` etc. anymore, as those implicitly go through `STATE(PtrBody)`.

This now leads me to my request for comments: I see the following alternatives, and I am interested to learn what others thinks; perhaps you have additional ideas; and anyway, it's good for myself to spell these out. Ultimately, I'd prefer an approach that also allows us to cover other types of statements and expressions, though.

1. Just live without these accessors (i.e., what this PR does right now); after all, we do that for all other statement types, whose structure often is of similar complexity, or even more complex, and not documented explicitly anywhere either. (Of course "other code is not documented well" is not exactly a great argument; but here it still has *some* merit, I'd argue. But I hope we can do better...)

2. Leave `FUNC_CALL`, `ARGI_CALL`, etc. in, but add in alternates, say, `FUNC_CALL_IN_BODY`, which also gets the `Obj body` as an argument, and switch code to use those as it transitions from using `STATE(PtrBody)` to passing the body object explicitly. (We can reduce code duplication by adding `FUNC_CALL_PTR` etc. which works similar to the next suggestion:)

3. Replace `FUNC_CALL(call)` by a helper function `Expr FUNC_CALL(void *ptrBody, Expr call)`; the first argument then either is `STATE(PtrBody)`, or else `ADDR_OBJ(body)`, depending on the use case.

4. Add C `struct`s which describe the layout of a function call, then use these for access. Something like this:
```C
struct CodedFuncCall {
   Expr func; // the function to call
   Expr args[]; // the arguments, as a variable length array
};
```
   However, this does not actually solve the whole problem, as we'd still need something like `NARG_SIZE_CALL` and `SIZE_NARG_CALL`. Moreover, flexible array members (i.e.,. `Expr args[];`) are not supported by C++, making this change very potentially problematic for kernel extensions with C++ code, and also possible future conversions of select GAP source files from C to C++.

5. Use a lightweight C++ class to abstract this (in a sense, that seems like the nicest solution to me, but earlier we agreed that we want to be very, very careful and restrictive about adding more C++ code, esp. code using C++ features we are not using yet...). In our case, this could work as follows (of course there is also a lot of potential for inheritance here, but I'll try to keep it simple; also, this code is untested and incomplete, and just meant to give an idea):
```C++
class CodedFuncCall {
  void * ptr;
public:
  // constructors for reading
  CodedFuncCall(Stat stat) : ptr(STATE(PtrBody) + stat) { }
  CodedFuncCall(Obj body, Stat stat) : ptr(ADDR_OBJ(body) + stat) { }
  // constructor for writing
  CodedFuncCall(int narg) {
    Stat stat = NewStat(T_FUNCCALL+narg, (1+narg)*sizeof(Expr));
    ptr = STATE(PtrBody) + stat;
  }
  
  // these could be moved to a parent class
  int size() const { return ((const StatHeader *)ptr - 1)->size; }
  int type() const { return ((const StatHeader *)ptr - 1)->type; }
  
  // reading (with const qualifier, so can be called on const instances)
  int narg() const { return size() / sizeof(Expr) - 1; }
  Expr func() const { return ptr[0]; }
  Expr arg(int i) const {
     GAP_ASSERT(1 <= i && i <= narg());
     return ptr[i];
  }

  // writing (for coding)
  void setFunc(Expr expr) { ptr[0] = expr; }
  Expr setArg(int i, Expr expr) {
     GAP_ASSERT(1 <= i && i <= narg());
     ptr[i] = expr;
  }
};
...

  const CodedFuncCall theCall(body, stat); // for reading only, hence const
  for (i = 1; i <= theCall.nargs(); i++) {
    Expr arg = theCall.arg(i);
    ... now do somethingn with the argument...
    theCall.setFunc(0); // will lead to a compiler error, as theCall is const
  }
```

6. ... some other ideas... ?

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -155,10 +155,6 @@ void SET_VISITED_STAT(Stat stat)
 }
 
 
-#define SET_FUNC_CALL(call,x)   WRITE_EXPR(call, 0, x)
-#define SET_ARGI_CALL(call,i,x) WRITE_EXPR(call, i, x)
-
-
 static inline void PushOffsBody( void ) {
     GAP_ASSERT(CS(OffsBodyCount) < MAX_FUNC_EXPR_NESTING);
     CS(OffsBodyStack)[CS(OffsBodyCount)++] = CS(OffsBody);
@@ -720,19 +716,21 @@ void CodeFuncCallEnd (
     UInt                i;              /* loop variable                   */
     Expr                opts = 0;       /* record literal for the options  */
     Expr                wrapper;        /* wrapper for calls with options  */
+    UInt                size;
 
     /* allocate the function call                                          */
+    size = (nr + 1) * sizeof(Expr);
     if ( funccall && nr <= 6 ) {
-        call = NewExpr( EXPR_FUNCCALL_0ARGS+nr, SIZE_NARG_CALL(nr) );
+        call = NewExpr(EXPR_FUNCCALL_0ARGS + nr, size);
     }
     else if ( funccall /* && 6 < nr */ ) {
-        call = NewExpr( EXPR_FUNCCALL_XARGS,    SIZE_NARG_CALL(nr) );
+        call = NewExpr(EXPR_FUNCCALL_XARGS, size);
     }
     else if ( /* ! funccall && */ nr <=6 ) {
-        call = NewExpr( STAT_PROCCALL_0ARGS+nr, SIZE_NARG_CALL(nr) );
+        call = NewExpr(STAT_PROCCALL_0ARGS + nr, size);
     }
     else /* if ( ! funccall && 6 < nr ) */ {
-        call = NewExpr( STAT_PROCCALL_XARGS,    SIZE_NARG_CALL(nr) );
+        call = NewExpr(STAT_PROCCALL_XARGS, size);
     }
 
     /* get the options record if any */
@@ -742,12 +740,12 @@ void CodeFuncCallEnd (
     /* enter the argument expressions                                      */
     for ( i = nr; 1 <= i; i-- ) {
         arg = PopExpr();
-        SET_ARGI_CALL(call, i, arg);
+        WRITE_EXPR(call, i, arg);
     }
 
     /* enter the function expression                                       */
     func = PopExpr();
-    SET_FUNC_CALL(call, func);
+    WRITE_EXPR(call, 0, func);
 
     /* wrap up the call with the options */
     if (options)
--- a/src/code.c
+++ b/src/code.c
@@ -157,7 +157,6 @@ void SET_VISITED_STAT(Stat stat)
 
 #define SET_FUNC_CALL(call,x)   WRITE_EXPR(call, 0, x)
 #define SET_ARGI_CALL(call,i,x) WRITE_EXPR(call, i, x)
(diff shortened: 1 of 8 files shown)


ChrisJefferson:
I'll admit, I'm going to pick on one little thing, as I don't think I fully get the big picture.

I don't really understand how removing things like: `#define SET_FUNC_CALL(call,x)   WRITE_EXPR(call, 0, x)` and `#define FUNC_CALL(call)  READ_EXPR(call, 0)` helps -- this just seems to wrap around the idea that the '0' slot is special. You appear to be mostly just inlining these macros, which doesn't seem to help, or hinder, the bigger picture?

fingolfin:
> I'll admit, I'm going to pick on one little thing, as I don't think I fully get the big picture.
> 
> I don't really understand how removing things like: `#define SET_FUNC_CALL(call,x) WRITE_EXPR(call, 0, x)` and `#define FUNC_CALL(call) READ_EXPR(call, 0)` helps -- this just seems to wrap around the idea that the '0' slot is special. You appear to be mostly just inlining these macros, which doesn't seem to help, or hinder, the bigger picture?
> 
> I mean, whatever you do, you are going to need to `READ_EXPR(call, 0)` and `WRITE_EXPR(call, 0, x)`, so I'd prefer to not remove them.

I thought I explained this pretty extensively above...?

To say it again with different words: The problem with `FUNC_CALL` and the other macros is that they implicitly read from / write to the function body pointed at by `PtrBody`. But in the future, this will only make sense in the executor, not in the coder, the syntax tree code or when printing functions. There are various ways to deal with this, described above. For example, one would be to add a new `FUNC_CALL_IN_BODY` macro which makes the function body an explicit argument; and then of the 18 calls to the `FUNC_CALL` macro, all but one would be converted to `FUNC_CALL_IN_BODY`.

Various other possibilities exist, described above.

PR #3405, opened 2019-04-12
Add tests for viewing/printing nested self-referential lists/records, and cleanup related code

fingolfin:
Another old change I had sitting on my hard disk for far too long:

At some point I tried to understand how the recursive printing/viewing of self-referential objects worked, and to help me, I very mildly refactored the code, and added some test cases that trigger this code.

I actually came to the conclusion that a subtle problem exists with this code, albeit in such a remote fringe case that I am not surprised if nobody ever run into it or will ever run into it (unless they deliberately try to trigger it): if printing is interrupted by a break loop, then we reset the "PrintObj stack" in `Shell()`, and later "restore" it, like this:
```C
  oldPrintDepth = STATE(PrintObjDepth);
  STATE(PrintObjDepth) = 0;
...
  STATE(PrintObjDepth) = oldPrintDepth;
```
But obviously this is *not* sufficient to actually save/restore the PrintObj stack! Thus, if we print a self-referential list; which is interrupted by a break loop; then print another or the same self-referential list during the break loop; and then return; then the PrintObj stack will be corrupt, and all kinds of weird stuff might happen: most likely, garbage is printed or errors are triggered for OOB access to lists or records. From a cursory glance, I don't think it would crash, but I wouldn't want to rule that out either without digging deeper.

If somebody wants to try to reproduce this, they could then also look into fixing this: I'd try to add functions `SavePrintObjStack` and `RestorePrintObjStack` which get called from `Shell`, and which copy/restore the state to some auxiliary object (probably a plist, or something slightly more complex), and restores from there as wel.

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -101,7 +101,6 @@ typedef struct GAPState {
                                    // that use GAP as a library to handle errors
 
     /* From objects.c */
-    Int PrintObjIndex;
     Int PrintObjDepth;
 
     /* From info.c */
--- a/src/lists.c
+++ b/src/lists.c
@@ -1759,7 +1759,7 @@ static void PrintListDefault(Obj list)
         if ( elm != 0 ) {
             if (1 < i)
                 Pr("%<,%< %2>", 0L, 0L);
-            STATE(PrintObjIndex) = i;
+            SetPrintObjIndex(i);
             PrintObj( elm );
         }
         else {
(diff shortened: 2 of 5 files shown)


fingolfin:
@DominikBernhardt yes, tracking it in an issue might be a good idea. Although it's a fairly low priority one, I guess...

PR #3412, opened 2019-04-15
Temporary fix for the Julia GC.

rbehrends:
This addresses a problem where the Julia GC during a partial sweep frees
some, but not all objects of an unreachable data structure. If an
address to a remaining object of such a data structure is still randomly
hit during conservative stack scanning, it may erroneously try to mark
the deallocated objects, too.

This temporary fix works by validating each pointer before it is marked.
Because this is potentially expensive, the eventual solution should fix
the root cause, which is conservative stack scanning resurrecting
pointers to dead objects that have been freed during a partial
collection.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -18,6 +18,7 @@
 #include "funcs.h"
 #include "gapstate.h"
 #include "gasman.h"
+#include "julia_gc.h"
 #include "objects.h"
 #include "plist.h"
 #include "sysmem.h"
@@ -482,6 +483,31 @@ static inline int JMark(void * obj)
         abort();
 #endif
     return jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj);
+}
+
+void MarkJuliaObjSafe(void * obj)
+{
+    if (!obj)
+        return;
+    // Validate that `obj` is still allocated and not on a
+    // free list already. We verify this by checking that the
+    // type is a pool object of type `jl_datatype_type`.
+    jl_value_t * ty = jl_typeof(obj);
+    if (jl_gc_internal_obj_base_ptr(ty) != ty)
+        return;
+    if (!jl_typeis(ty, jl_datatype_type))
+        return;
+    if (jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj))
+        YoungRef++;
+}
+
+
+void MarkJuliaObj(void * obj)
+{
+    if (!obj)
+        return;
+    if (JMark(obj))
+        YoungRef++;
 }
 
 
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -37,10 +37,14 @@
 **  Various options controlling special features of the Julia GC code follow
 */
 
-// if DISABLE_BIGVAL_TRACKING is defined, we don't track the location of
-// large bags; this speeds up some things, but may expose bugs in GAP code
-// which incorrectly holds pointers into bags over a GC.
-// #define DISABLE_BIGVAL_TRACKING
+// if SCAN_STACK_FOR_MPTRS_ONLY is defined, stack scanning will only
+// look for references to master pointers, but not bags themselves. This
+// should be safe, as GASMAN uses the same mechanism. It is also faster
+// and avoids certain complicated issues that can lead to crashes, and
+// is therefore the default. The option to scan for all pointers remains
+// available for the time being and should be considered to be
+// deprecated.
+#define SCAN_STACK_FOR_MPTRS_ONLY
 
 // if REQUIRE_PRECISE_MARKING is defined, we assume that all marking
 // functions are precise, i.e., they only invoke MarkBag on valid bags,
@@ -53,6 +57,12 @@
 
 // if MARKING_STRESS_TEST is defined, we stress test the TryMark code
 // #define MARKING_STRESS_TEST
+
+// if VALIDATE_MARKING is defined, the program is aborted if we ever
+// encounter a reference during marking that does not meet additional
+// validation criteria. These tests are compararively expensive and
+// should not be enabled by default.
+// #define VALIDATE_MARKING
 
 
 /****************************************************************************
@@ -125,7 +135,7 @@ static void JFinalizer(jl_value_t * obj)
         TabFreeFuncBags[tnum]((Bag)&contents);
 }
 
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
 
 /****************************************************************************
 **
@@ -197,7 +207,7 @@ static inline void * align_ptr(void * p)
     return (void *)u;
 }
 
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
 
 typedef struct treap_t {
     struct treap_t *left, *right;
@@ -404,10 +414,10 @@ static UInt            StackAlignBags;
 static Bag *           GapStackBottom;
 static jl_ptls_t       JuliaTLS, SaveTLS;
 static size_t          max_pool_obj_size;
-#if !defined(DISABLE_BIGVAL_TRACKING)
-static size_t          bigval_startoffset;
-#endif
 static UInt            YoungRef;
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
+static size_t bigval_startoffset;
+#endif
 
 
 #ifndef NR_GLOBAL_BAGS
@@ -449,10 +459,31 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
     TabMarkFuncBags[type] = mark_func;
 }
 
-static inline int JMark(void * obj)
+static inline int JMarkGapObjSafe(void * obj)
 {
+    // only traverse objects internally used by GAP
+    void * ty = jl_typeof(obj);
+    if (ty != datatype_mptr && ty != datatype_bag &&
+        ty != datatype_largebag && ty != jl_weakref_type)
+        return 0;
     return jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj);
 }
+
+static inline int JMark(void * obj)
+{
+#ifdef VALIDATE_MARKING
+    // Validate that `obj` is still allocated and not on a
+    // free list already. We verify this by checking that the
+    // type is a pool object of type `jl_datatype_type`.
+    jl_value_t * ty = jl_typeof(obj);
+    if (jl_gc_internal_obj_base_ptr(ty) != ty)
+        abort();
+    if (!jl_typeis(ty, jl_datatype_type))
+        abort();
+#endif
+    return jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj);
+}
+
 
 // Overview of conservative stack scanning
 //
@@ -478,12 +509,19 @@ static inline int JMark(void * obj)
 // the stack (including Julia task stacks) that points to a master pointer
 // or the contents of a bag (including a location after the start of the
 // bag) indicates a valid reference that needs to be marked.
+//
+// One additional concern is that Julia may opportunistically free a subset
+// of unreachable objects. Thus, with conservative stack scanning, it is
+// possible for a pointer to resurrect a previously unreachable object,
+// from which freed objects are then marked. Hence, we add additional checks
+// when traversing GAP master pointer and bag objects that this happens
+// only for live objects.
 
 static void TryMark(void * p)
 {
     jl_value_t * p2 = jl_gc_internal_obj_base_ptr(p);
     if (!p2) {
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
         // It is possible for p to point past the end of
         // the object, so we subtract one word from the
         // address. This is safe, as the object is preceded
@@ -514,7 +552,12 @@ static void TryMark(void * p)
 #endif
     }
     if (p2) {
-        JMark(p2);
+#ifdef SCAN_STACK_FOR_MPTRS_ONLY
+        if (jl_typeis(p2, datatype_mptr))
+            JMark(p2);
+#else
+        JMarkGapObjSafe(p2);
+#endif
     }
 }
 
@@ -574,9 +617,12 @@ void CHANGED_BAG(Bag bag)
 
 static void GapRootScanner(int full)
 {
-    // mark our Julia module (this contains references to our custom data
-    // types, which thus also will not be collected prematurely)
-    JMark(Module);
+    // Mark our Julia module (this contains references to our custom data
+    // types, which thus also will not be collected prematurely).
+    // We call jl_gc_mark_queue_obj() directly here, because we know that
+    // Module is a valid Julia object at this point, so further checks
+    // in JMark() can be skipped.
+    jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)Module);
     jl_task_t * task = JuliaTLS->current_task;
     size_t      size;
     int         tid;
@@ -688,7 +734,15 @@ static uintptr_t JMarkMPtr(jl_ptls_t ptls, jl_value_t * obj)
 {
     if (!*(void **)obj)
         return 0;
-    if (JMark(BAG_HEADER((Bag)obj)))
+    void * header = BAG_HEADER((Bag)obj);
+    // The following check ensures that the master pointer does
+    // indeed reference a bag that has not yet been freed. See
+    // the comments on conservative stack scanning for an in-depth
+    // explanation.
+    void * ty = jl_typeof(header);
+    if (ty != datatype_bag && ty != datatype_largebag)
+        return 0;
+    if (JMark(header))
         return 1;
     return 0;
 }
@@ -714,7 +768,7 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     }
     // These callbacks need to be set before initialization so
     // that we can track objects allocated during `jl_init()`.
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
     jl_gc_set_cb_notify_external_alloc(alloc_bigval, 1);
     jl_gc_set_cb_notify_external_free(free_bigval, 1);
     bigval_startoffset = jl_gc_external_obj_hdr_size();
@@ -827,7 +881,7 @@ Bag NewBag(UInt type, UInt size)
     if (size == 0)
         alloc_size++;
 
-#if defined(DISABLE_BIGVAL_TRACKING)
+#if defined(SCAN_STACK_FOR_MPTRS_ONLY)
     bag = jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
     SET_PTR_BAG(bag, 0);
 #endif
@@ -839,7 +893,7 @@ Bag NewBag(UInt type, UInt size)
     header->size = size;
 
 
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
     // allocate the new masterpointer
     bag = jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
     SET_PTR_BAG(bag, DATA(header));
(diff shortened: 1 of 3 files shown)


src/julia_gc.c
```diff
@@ -54,6 +54,8 @@
 // if MARKING_STRESS_TEST is defined, we stress test the TryMark code
 // #define MARKING_STRESS_TEST
 
+#define DISABLE_BIGVAL_TRACKING
```
fingolfin: The correct way would be to uncomment the `#define` in line 43

src/julia_gc.c
```diff
@@ -450,6 +452,10 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
 
 static inline int JMark(void * obj)
 {
+    void *ty = jl_typeof(obj);
+    if (ty != datatype_mptr && ty != datatype_bag
+        && ty != datatype_largebag && ty != jl_weakref_type)
+        return 0;
```
fingolfin: Hm, so that's quite a bit different from the changes we discussed last. So, this is a hard coded list of types -- so we are making the hard assumption here that GAP objects will *never* directly reference other Julia objects. But that is not only fragile, it is also quite simply wrong, because JuliaInterface of course does so.

Hence, I don't understand how this can work at all.

I also don't quite understand why it is now sufficient to perform this check in `JMark` alone -- is it safe to increment `YoungRef` in `MarkBag` when dealing with invalid bags?

rbehrends: Hmm, I hadn't realized that JuliaInterface goes through `JMark()` also. This is something that needs to be fixed indeed. I understand the fragility concern, but this is a hot path, where we need to look up these types fast.

The `MarkBag()` case is safe. We have at this point already validated that `p` is a valid address. If on a freelist or if unmarked, then the lowest bit will be zero and we will not increment `YoungRef`.

I am still going to rewrite this for better readability (and, of course, to remove the `JMark()` problem.

src/julia_gc.c
rbehrends: Will do.

src/julia_gc.c
```diff
@@ -685,7 +701,7 @@ static void PostGCHook(int full)
 // is called by the Julia GC whenever it marks a GAP master pointer object)
 static uintptr_t JMarkMPtr(jl_ptls_t ptls, jl_value_t * obj)
 {
-    if (!*(void **)obj)
+    if (!*(void **)obj || !jl_typeis(obj, datatype_mptr))
```
fingolfin: But `obj` does not point at the start of the mptr, `BAG_HEADER((Bag)obj)` does.

src/julia_gc.c
```diff
@@ -575,7 +591,7 @@ static void GapRootScanner(int full)
 {
     // mark our Julia module (this contains references to our custom data
     // types, which thus also will not be collected prematurely)
-    JMark(Module);
+    jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *) Module);
```
fingolfin: Why this change?

src/julia_gc.c
```diff
@@ -513,7 +524,12 @@ static void TryMark(void * p)
 #endif
     }
     if (p2) {
-        JMark(p2);
+#ifdef DISABLE_BIGVAL_TRACKING
+        JMarkGapObjSafe(p2);
+#else
+        if (jl_typeis(p2, datatype_mptr))
+            JMark(p2);
```
fingolfin: Why not always use this? Why is the `DISABLE_BIGVAL_TRACKING` doing something else?

src/julia_gc.c
fingolfin: Ah, I somehow overlooked that `MarkBag` already validates the pointer (which it indeed does).

src/julia_gc.c
rbehrends: This allowed me to put validation logic in `JMark()` to verify that bad pointers indeed don't get traversed. I should actually put that in.

src/julia_gc.c
rbehrends: Yes, that was actually an unnecessary test, too.

src/julia_gc.c
rbehrends: The two cases were flipped around, or alternatively, an `#ifndef` should have been used. So the `DISABLE_BIGVAL_TRACKING` was actually more restrictive.

fingolfin: OK, but that doesn't really answer my question: why are the two cases doing something different? (Perhaps a code comment should explain it?)

src/julia_gc.c
```diff
@@ -605,7 +634,7 @@ static void GapRootScanner(int full)
     for (Int i = 0; i < GlobalCount; i++) {
         Bag p = *GlobalAddr[i];
         if (IS_BAG_REF(p)) {
-            JMark(p);
+            JMarkGapObjSafe(p);
```
fingolfin: Why does this need to use `JMarkGapObjSafe`? It seems to be me that we even could use `jl_gc_mark_queue_obj` directly here, no?

src/julia_gc.c
rbehrends: Okay, technically the `DISABLE_BIGVAL_TRACKING` should actually have a different name, namely that we only mark master pointers conservatively from the stack. Without the test, it would also mark small objects. I think the right thing to do here is to rename the define.

src/julia_gc.c
rbehrends: It probably could. I'll have a look at that.

src/julia_gc.c
```diff
@@ -37,10 +37,13 @@
 **  Various options controlling special features of the Julia GC code follow
 */
 
-// if DISABLE_BIGVAL_TRACKING is defined, we don't track the location of
-// large bags; this speeds up some things, but may expose bugs in GAP code
-// which incorrectly holds pointers into bags over a GC.
-// #define DISABLE_BIGVAL_TRACKING
+// if SCAN_STACK_FOR_MPTRS_ONLY is defined, stack scanning will only
+// look for references to master pointers, but not bags themselves.
+// This should be safe, as Gasman uses the same mechanism, but it is
```
```suggestion
// This should be safe, as GASMAN uses the same mechanism, but it is
```

src/julia_gc.c
```diff
@@ -37,10 +37,13 @@
...
+// at least theoretically possible that an optimizing compiler will
+// remove the last master pointer reference before the last reference
+// to the bag itself.
```
fingolfin: As discussed several times before: such an optimization would necessarily also affect GASMAN in the very same way! Except perhaps if the Julia GC was running concurrently to the thread GAP is running in; but then we might have a serious problem anyway.

src/julia_gc.c
```diff
@@ -196,7 +205,7 @@ static inline void * align_ptr(void * p)
     return (void *)u;
 }
 
-#if !defined(DISABLE_BIGVAL_TRACKING)
+#if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
```
fingolfin: ... seems it's a good thing I insisted on putting my `DISABLE_BIGVAL_TRACKING` mod in there... ;-)

src/julia_gc.c
```diff
@@ -448,11 +457,51 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
...
+#endif
     return jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj);
 }
 
+void MarkJuliaObjSafe(void * obj)
+{
+    if (!obj)
+        return;
+    jl_value_t *ty = jl_typeof(obj);
+    if (jl_gc_internal_obj_base_ptr(ty) != ty)
```
fingolfin: So we call `jl_gc_internal_obj_base_ptr` on the type -- why don't we just call `jl_gc_internal_obj_base_ptr` on the object, i.e., `jl_gc_internal_obj_base_ptr(obj)`?

src/julia_gc.c
```diff
@@ -448,11 +457,51 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
...
+        return;
+    if (!jl_typeis(ty, jl_datatype_type))
+        return;
```
fingolfin: Please add comments that explain the checks above

src/julia_gc.c
fingolfin: Same question of course applies to the debug code in `JMark`

src/julia_gc.c
```diff
@@ -448,11 +457,51 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
     TabMarkFuncBags[type] = mark_func;
 }
 
+static inline int JMarkGapObjSafe(void * obj)
+{
+    // only traverse objects internally used by GAP
+    void *ty = jl_typeof(obj);
+    if (ty != datatype_mptr && ty != datatype_bag
+        && ty != datatype_largebag && ty != jl_weakref_type)
```
fingolfin: There are three calls to `JMarkGapObjSafe`. Two are in `MarkBag` -- but for that caller, we only need to accept `datatype_mptr` and `jl_weakref_type`. And if we added a custom `MarkBagWeakly` function (to mirror the name used in GASMAN) which only accepts `jl_weakref_type`, we could reduce the check to `jl_typeis(obj, datatype_mptr)` which is more efficient.

And `MarkBagWeakly` could potentially be considerably simpler than `MarkBag`, as (a) it is very rarely used, and (b) we control where it is called from tightly, and so we'd not have to validate as much, and could skipping the `MarkCache`. So this might suffice:
```C
void MarkBagWeakly(Bag bag)
{
    if (bag == 0)
        return;
    jl_value_t * p = (jl_value_t *)bag;
    GAP_ASSERT(jl_typeis(p, jl_weakref_type));
    if (jl_gc_mark_queue_obj(JuliaTLS, p))
        YoungRef++;
}
```


Anyway, that leaves the third call to `JMarkGapObjSafe`, which is in `TryMark`. Note that in the `SCAN_STACK_FOR_MPTRS_ONLY` case, it also only checks `jl_typeis(obj, datatype_mptr)`.

So, let's change `JMarkGapObjSafe` to only check for `datatype_mptr`, and then replace the one remaining call to it with a copy of the code that's now here. And then we could rename this function to something more specific... I want to say `JMarkMPtr`, but we already use that name (which, BTW, I find rather confusing -- `JMarkMPtr` is a Julia marking function, while `JMark` is a helper function for use in lots of places

src/julia_gc.c
```diff
@@ -687,7 +741,11 @@ static uintptr_t JMarkMPtr(jl_ptls_t ptls, jl_value_t * obj)
 {
     if (!*(void **)obj)
         return 0;
-    if (JMark(BAG_HEADER((Bag)obj)))
+    void *header = BAG_HEADER((Bag) obj);
```
```suggestion
    void *header = BAG_HEADER((Bag)obj);
```

src/julia_gc.c
```diff
@@ -448,11 +457,51 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
...
+    jl_value_t *ty = jl_typeof(obj);
+    if (jl_gc_internal_obj_base_ptr(ty) != ty)
+        abort();
+    if (!jl_typeis(ty, jl_datatype_type))
+        abort();
+#endif
     return jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj);
 }
 
+void MarkJuliaObjSafe(void * obj)
```
sebasguts: Since we need those functions (`MarkJuliaObjSafe` and `MarkJuliaObj`) in JuliaInterface, could you expose them in a header?

And, out of curiosity, is there a reason that the argument type is `void*` and not `jl_value_t*`?

fingolfin: Because `void *` is much easier to expose in a header (doesn't force all GAP code to include Julia headers)

sebasguts: Makes sense, thank you for the clarification.

src/julia_gc.c
rbehrends: At this point we already know that `obj` points to a proper object, which may be either a pool or large object. What the check does is verifying that it actually is a proper Julia object and has not been freed; if freed, the type tag would be a freelist pointer and fail that check.

src/julia_gc.c
rbehrends: In `TryMark()`, `JMarkGapObjSafe()` is not called if `SCAN_STACK_FOR_MPTRS_ONLY` is enabled. Otherwise, it's used in lieu of the MPtr type check and needs to cover all cases.

I did not specialize the type check in `JMarkGapObjSafe()` further for other use cases, because it will succeed on the first hit for almost all cases with `SCAN_STACK_FOR_MPTRS_ONLY` enabled and is probably an unnecessary optimization.

src/julia_gc.c
rbehrends: Will do.

src/julia_gc.c
rbehrends: Yes, we can definitely put that in a header.

src/julia_gc.c
rbehrends: I am on board with renaming the functions, though. That thought had occurred to me, too.

src/julia_gc.c
fingolfin: To be clear: With that remark I meant to say that I think your comment is misleading, and should be adjusted. Either removed, or qualified.

src/julia_gc.c
```diff
@@ -687,7 +747,11 @@ static uintptr_t JMarkMPtr(jl_ptls_t ptls, jl_value_t * obj)
 {
     if (!*(void **)obj)
         return 0;
-    if (JMark(BAG_HEADER((Bag)obj)))
+    void *header = BAG_HEADER((Bag)obj);
+    void *ty = jl_typeof(header);
+    if (ty != datatype_bag && ty != datatype_largebag)
```
fingolfin: Please add a brief comment here which explains why this check is performed.

src/julia_gc.c
```diff
@@ -575,7 +636,7 @@ static void GapRootScanner(int full)
 {
     // mark our Julia module (this contains references to our custom data
     // types, which thus also will not be collected prematurely)
-    JMark(Module);
+    jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *) Module);
```
fingolfin: Please add a brief comment here which explains why this does not call `JMark` resp. `MarkJuliaObj` (probably something like this: 
```suggestion
    // We call jl_gc_mark_queue_obj directly here instead of one of our wrappers
    // because we know that <Module> is a valid Julia object here; so we can and
    // should bypass any validation checks.
    jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)Module);
```
Although this does not yet explain why `YoungRef` is not relevant here, which should also be stated.

src/julia_gc.h
```diff
@@ -0,0 +1,35 @@
...
+**
+**  This file declares functions of the Julia GC interface.
+*/
+
+#ifndef GAP_JULIA_GC_H
+#define GAP_JULIA_GC_H
+
+/****************************************************************************
+**
+*F  MarkJuliaObj(<obj>)  . . . . . . . . . . . . . . . . mark Julia object
```
fingolfin:
```suggestion
*F  MarkJuliaObj(<obj>) . . . . . . . . . . . . . . . . . . mark Julia object
```

src/julia_gc.h
```diff
@@ -0,0 +1,35 @@
...
+**
+**  'MarkJuliaObjSafe' marks a Julia object; the argument can be NULL.
```
fingolfin:
```suggestion
**  'MarkJuliaObj' marks a Julia object; the argument may be NULL. No further
**  checks are performed. If <obj> is not a Julia object, a crash may result.
```

src/julia_gc.h
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+void MarkJuliaObj(void * obj);
+
+/****************************************************************************
+**
+*F  MarkJuliaObjSafe(<obj>)  . . . . . . . . . . . . . . mark Julia object
+**
+**  'MarkJuliaObjSafe' marks a Julia object, verifying that it is a valid
+**  object.
```
fingolfin:
```suggestion
```

src/julia_gc.h
```diff
@@ -0,0 +1,35 @@
...
+**  'MarkJuliaObjSafe' marks a Julia object; the argument can be NULL.
+*/
+
+void MarkJuliaObj(void * obj);
+
+/****************************************************************************
+**
+*F  MarkJuliaObjSafe(<obj>)  . . . . . . . . . . . . . . mark Julia object
+**
+**  'MarkJuliaObjSafe' marks a Julia object, verifying that it is a valid
```
fingolfin:
```suggestion
**  'MarkJuliaObjSafe' marks a Julia object; the argument may be NULL.
**  Extra validation steps are performed to determine whether <obj> is a valid
**  Julia object pointer; if not, it is silently ignored.
```

src/julia_gc.h
rbehrends: This does not seem to be correct?

src/julia_gc.c
```diff
@@ -30,17 +30,20 @@
...
 **  Various options controlling special features of the Julia GC code follow
 */
 
-// if DISABLE_BIGVAL_TRACKING is defined, we don't track the location of
-// large bags; this speeds up some things, but may expose bugs in GAP code
-// which incorrectly holds pointers into bags over a GC.
-// #define DISABLE_BIGVAL_TRACKING
+// if SCAN_STACK_FOR_MPTRS_ONLY is defined, stack scanning will only
+// look for references to master pointers, but not bags themselves.
+// This should be safe, as GASMAN uses the same mechanism and is therefore
```
fingolfin: Wouldn't this be more accurate?
```suggestion
// This should be safe, as GASMAN uses the same mechanism. It is also faster
// and avoids certain complicated issues that can lead to crashes, and is therefore
```

src/julia_gc.h
```diff
@@ -0,0 +1,48 @@
...
+**  'MarkJuliaObjSafe' marks a Julia object; the argument may be NULL.
+**  Extra validation steps are performed to determine whether <obj> is
+**  a valid Julia object. If not, it is silently ignored.
+*/
+
+void MarkJuliaObjSafe(void * obj);
+
+/****************************************************************************
+**
+*F  MarkJuliaWeakRef(<ref>)  . . . . . . . . . .  mark Julia weak reference
```
fingolfin:
```suggestion
*F  MarkJuliaWeakRef(<ref>) . . . . . . . . . . . . mark Julia weak reference
```

src/julia_gc.h
```diff
@@ -0,0 +1,48 @@
...
+**
+**  This file declares functions of the Julia GC interface.
+*/
+
+#ifndef GAP_JULIA_GC_H
+#define GAP_JULIA_GC_H
+
+/****************************************************************************
+**
+*F  MarkJuliaObj(<obj>)  . . . . . . . . . . . . . . . . . mark Julia object
```
fingolfin:
```suggestion
*F  MarkJuliaObj(<obj>) . . . . . . . . . . . . . . . . . . mark Julia object
```

src/julia_gc.h
```diff
@@ -0,0 +1,48 @@
...
+**  'MarkJuliaObjSafe' marks a Julia object; the argument can be NULL. No
+**  further checks are performed. If <obj> is not a Julia object, a crash
+**  may result.
+*/
+
+void MarkJuliaObj(void * obj);
+
+/****************************************************************************
+**
+*F  MarkJuliaObjSafe(<obj>)  . . . . . . . . . . . . . . . mark Julia object
```
fingolfin:
```suggestion
*F  MarkJuliaObjSafe(<obj>) . . . . . . . . . . . . . . . . mark Julia object
```

src/julia_gc.c
```diff
@@ -448,11 +457,55 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
     TabMarkFuncBags[type] = mark_func;
 }
 
+static inline int JMarkTyped(void * obj, jl_datatype_t * ty)
```
fingolfin: OK, that seems fine!

But, if we use this name with `JMark` prefix here now, please let's at least rename `JMarkMPtr` and `JMarkBag` -- e.g. to `JuliaMarkFunctionForMPtrs` and `JuliaMarkFunctionForBags`.

src/julia_gc.c
```diff
@@ -448,11 +457,55 @@ void InitMarkFuncBags(UInt type, TNumMarkFuncBags mark_func)
...
+    if (jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj))
+        YoungRef++;
+}
+
+
+void MarkJuliaObj(void * obj)
+{
+    if (!obj)
+        return;
+    if (jl_gc_mark_queue_obj(JuliaTLS, (jl_value_t *)obj))
```
fingolfin: Should we perhaps call `JMark` here so that when `VALIDATE_MARKING` is enabled, we validate here, too?

src/julia_gc.c
rbehrends: I picked `MPtrMarkFunc` and `BagMarkFunc`, respectively. Those should be both clear and distinct enough.

src/julia_gc.c
rbehrends: Good point, will put that in.

PR #3414, opened 2019-04-15
kernel: merge ListExpr{1,2} into EvalList{Tilde,}Expr

fingolfin:
The result is not much longer, but this way, the code is easier to understand and adjust.

I did this because of some other work which modifies these functions among other things, which lead to this refactoring.

### DIFF
--- a/src/exprs.c
+++ b/src/exprs.c
@@ -811,22 +811,52 @@ static Obj EvalPermExpr(Expr expr)
 **
 **  'EvalListExpr'  evaluates the list   expression, i.e., not  yet evaluated
 **  list, <expr> to a list value.
-**
-**  'EvalListExpr'  just  calls 'ListExpr1'  and  'ListExpr2' to evaluate the
-**  list expression.
 */
-static Obj  ListExpr1(Expr expr, Int tildeInUse);
-static void ListExpr2(Obj list, Expr expr, Int tildeInUse);
-
 static Obj EvalListExpr(Expr expr)
 {
-    Obj                 list;         /* list value, result                */
+    Obj                 list;           /* list value, result              */
+    Obj                 sub;            /* value of a subexpression        */
+    Int                 len;            /* logical length of the list      */
+    Int                 i;              /* loop variable                   */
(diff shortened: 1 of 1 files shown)


src/exprs.c
```diff
@@ -811,165 +811,117 @@ static Obj EvalPermExpr(Expr expr)
...
+    // handle the subexpressions
+    for (i = 1; i <= len; i++) {
+        Expr subExpr = READ_EXPR(expr, i - 1);
 
-    /* assign the list to '~'                                              */
-    STATE(Tilde) = list;
+        // skip holes
+        if (subExpr == 0) {
+            // there is a hole, hence the list is not dense (note that list
+            // expressions never contain holdes at the end, so we do not have
```
fingolfin: Typo : holdes

src/exprs.c
```diff
@@ -811,165 +811,117 @@ static Obj EvalPermExpr(Expr expr)
...
-                                           next full position causes
-                                           the list to be made
-                                           non-dense */
+    // remember the old value of '~'
+    tilde = STATE(Tilde);
 
-    /* get the length of the list                                          */
-    len = SIZE_EXPR(expr) / sizeof(Expr);
+    // assign the list to '~'
+    STATE(Tilde) = list;
```
fingolfin: Huh... I just thought about all this, and now I have serious doubts whether this PR is correct -- or rather, whether the code it refactors was correct; I now think we must *always* use the "slow" path; or else we need a third expression type T_LIST_NESTED_IN_TILDE_EXPR or so (and all these changes are also needed for records).

fingolfin: OK, when I wrote the above, I thought the following might happen: while inside a `T_LIST_EXPR` which is nested inside a `T_TILDE_LIST_EXPR`, we can use `~` to access subobjects of the parent list. I thought we could use that to modify the nested list as it was being constructed. However, this is not true, because we only store a reference to the nested list into the parent list after the nested list has been fully constructed; any access to it after that, however, is fine.

So, false alarm. Phew :-)

PR #3415, opened 2019-04-15
Minor tweaks: remove T_NO_STAT; update an example in dev manual; whitespace & comment fixes

fingolfin:


doc/dev/kernel.xml
```diff
@@ -863,9 +863,10 @@ $ gap -b
 GAP4, Version: 4.dev of today, ....
 gap> LoadDynamicModule("./foo.so");
 gap> Print(foo);
-function ( <<arg-1>> )
-    <<compiled code>>
-endgap> foo(3);
+function ( x )
```
PaulaHaehndel: Sorry, I can't check whether gap actually outputs "function ( x )" or "function ( <<arg-1>> )"

doc/dev/kernel.xml
```diff
@@ -863,9 +863,10 @@ $ gap -b
...
+    <<compiled GAP code>> from foo.g:1
+end
+gap> foo(3);
```
PaulaHaehndel: I'm fine with the other changes. They are reasonable.

src/iostream.c
```diff
@@ -131,7 +131,7 @@ static void KillChild(UInt stream)
 
 /****************************************************************************
 **
-*F  GetMasterPty( <fid> ) . . . . . . . . .  open a master pty (from "xterm")
+*F  OpenPty( <master>, <slave> ) . . . . . . . . open a pty master/slave pair
```
PaulaHaehndel: This now seems to fit the code. So thank you.

src/hpc/threadapi.c
```diff
@@ -185,11 +185,11 @@ static void WaitThreadSignal(void)
 {
     int id = TLS(threadID);
     if (!UpdateThreadState(id, TSTATE_RUNNING, TSTATE_BLOCKED))
-        HandleInterrupts(1, T_NO_STAT);
+        HandleInterrupts(1, 0);
```
PaulaHaehndel: Why is it replaced by 0 and not -1 as the value was beforehand?
Does this cause any problems by calling in thread.c:
`SET_BRK_CALL_TO(0);`
instead of 
`SET_BRK_CALL_TO(-1);`?

Sorry, my understanding is not good enough to judge this on my own.

ChrisJefferson: This is a very reasonable question, I'm interested too

doc/dev/kernel.xml
fingolfin: Actually, I did it by following the instructions right there in the file -- when you say you can't check it, do you mean those instructions didn't work for you, or that you simply did not run them (which is fine, of course! Just in case there were problems with the instructions, I'd like to address them, too)

src/hpc/threadapi.c
fingolfin: That's of course a reasonable question, thank you for checking closely! I was worried by your question at first, but now I think all is fine: The call to `SET_BRK_CALL_TO` in `InterruptCurrentThread` is never reached in either case. Before this PR, the code does `if (stat == T_NO_STAT) return;`, with this PR it does `if (stat == 0) return;`

doc/dev/kernel.xml
PaulaHaehndel: No the instruction where fine, but for some reasons that I didn't want to invastigate my gap-master-version gives an error when I try to `LoadDynamicModule` and my gap-10.0-version still returns `function ( <<arg-1>> )`.

src/hpc/threadapi.c
PaulaHaehndel: Ah, right, thank you. And `InterruptCurrentThread` is never called somewhere else with `state`<1?

fingolfin: Yes. I just did `git grep InterruptCurrentThread` to verify that `InterruptCurrentThread` is called in exactly one place:
```
$ git grep -n InterruptCurrentThread
src/hpc/thread.c:859:static void InterruptCurrentThread(int locked, Stat stat)
src/hpc/thread.c:904:        InterruptCurrentThread(locked, stat);
```

PaulaHaehndel: Then this is fine. Thank you.

PR #3419, opened 2019-04-23
gac: emit semicolon after all CHECK_* calls; turn CHECK_{BOOL,FUNC} into functions

fingolfin:
Main motivation for this is to ensure sensible error message in gac compiled code which match those produced by the interpreter and coder. A new test verifies this for `CHECK_BOOL`.

I was planning to add a similar test for `CHECK_FUNC`, but that revealed bugs in gac related to that, so that test along with a fix will be in a separate PR.


ChrisJefferson:
I'll close this, and merge the other one.

PR #3420, opened 2019-04-23
Give human-readable names to functions in InstallMethod

ChrisJefferson:
@markusbaumeister pointed out that many library functions show up as "nameless" in profiling output. This turned out to be because InstallMethod does not give the functions a useful name. This is a shame, as it's easy to give them a very informative name in InstallMethod. This PR does that.

This shouldn't cause any problems, apart from using a small amount of extra memory on strings. I decided to also give a useful name to the functions which calculate priority. While these functions are very small, it's still useful to know what they are in a trace.

## Text for release notes 

* Give more library methods human-readable names. These are used when profiling.

### DIFF
--- a/lib/oper1.g
+++ b/lib/oper1.g
@@ -389,7 +389,7 @@ BIND_GLOBAL( "INSTALL_METHOD",
           rank,
           method,
           oreqs,
-          req, reqs, match, j, k, imp, notmatch, lk;
+          req, reqs, match, j, k, imp, notmatch, lk, funcname;
 
     if IsHPCGAP then
         # TODO: once the GAP compiler supports 'atomic', use that
@@ -599,6 +599,31 @@ BIND_GLOBAL( "INSTALL_METHOD",
         od;
 
       fi;
+    fi;
+
+    if IS_FUNCTION(method) and IsBound(HasNameFunction) and
+      IsBound(TYPE_FUNCTION_WITH_NAME) and IsBound(TYPE_OPERATION_WITH_NAME) and
+      not VAL_GVAR("HasNameFunction")(method) then
+        funcname := SHALLOW_COPY_OBJ(NAME_FUNC(opr));
+        APPEND_LIST_INTR(funcname, " ");
+        if info <> false then
+            APPEND_LIST_INTR(funcname, info);
+        else
+            APPEND_LIST_INTR(funcname, "method");
+        fi;
+        SET_NAME_FUNC(method, funcname);
+    fi;
+
+    if IS_FUNCTION(rank) and IsBound(HasNameFunction) and
+       IsBound(TYPE_FUNCTION_WITH_NAME) and IsBound(TYPE_OPERATION_WITH_NAME) and
+       not VAL_GVAR("HasNameFunction")(rank) then
+        funcname := "Priority calculation for ";
+        APPEND_LIST_INTR(funcname, NAME_FUNC(opr));
+        if info <> false then
+            APPEND_LIST_INTR(funcname, " ");
+            APPEND_LIST_INTR(funcname, info);
+        fi;
+        SET_NAME_FUNC(rank, funcname);
     fi;
 
     # Install the method in the operation.
--- a/src/c_oper1.c
+++ b/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include "compiled.h"
-#define FILE_CRC  "127678438"
+#define FILE_CRC  "72241890"
 
 /* global variables used in handlers */
 static GVar G_REREADING;
@@ -12,10 +12,16 @@ static GVar G_PRINT__OBJ;
 static Obj  GC_PRINT__OBJ;
 static GVar G_GAPInfo;
 static Obj  GC_GAPInfo;
+static GVar G_TYPE__FUNCTION__WITH__NAME;
+static Obj  GC_TYPE__FUNCTION__WITH__NAME;
+static GVar G_TYPE__OPERATION__WITH__NAME;
+static Obj  GC_TYPE__OPERATION__WITH__NAME;
 static GVar G_IS__FUNCTION;
 static Obj  GF_IS__FUNCTION;
 static GVar G_NAME__FUNC;
 static Obj  GF_NAME__FUNC;
+static GVar G_SET__NAME__FUNC;
+static Obj  GF_SET__NAME__FUNC;
 static GVar G_NARG__FUNC;
 static Obj  GF_NARG__FUNC;
 static GVar G_IS__OPERATION;
@@ -32,6 +38,8 @@ static GVar G_IS__STRING__REP;
 static Obj  GF_IS__STRING__REP;
 static GVar G_Error;
 static Obj  GF_Error;
+static GVar G_VAL__GVAR;
+static Obj  GF_VAL__GVAR;
 static GVar G_TYPE__OBJ;
 static Obj  GF_TYPE__OBJ;
 static GVar G_IMMUTABLE__COPY__OBJ;
@@ -132,6 +140,8 @@ static GVar G_NamesFilter;
 static Obj  GF_NamesFilter;
 static GVar G_Ordinal;
 static Obj  GF_Ordinal;
+static GVar G_HasNameFunction;
+static Obj  GC_HasNameFunction;
 static GVar G_INSTALL__METHOD__FLAGS;
 static Obj  GF_INSTALL__METHOD__FLAGS;
 static GVar G_LENGTH__SETTER__METHODS__2;
@@ -1682,6 +1692,7 @@ static Obj  HdlrFunc6 (
  Obj l_imp = 0;
  Obj l_notmatch = 0;
  Obj l_lk = 0;
+ Obj l_funcname = 0;
  Obj t_1 = 0;
  Obj t_2 = 0;
  Obj t_3 = 0;
@@ -1714,6 +1725,7 @@ static Obj  HdlrFunc6 (
  (void)l_imp;
  (void)l_notmatch;
  (void)l_lk;
+ (void)l_funcname;
  Bag oldFrame;
  
  /* allocate new stack frame */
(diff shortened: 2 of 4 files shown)


lib/oper1.g
```diff
@@ -601,6 +601,25 @@ BIND_GLOBAL( "INSTALL_METHOD",
       fi;
     fi;
 
+    if IS_FUNCTION(method) and not HAS_NAME_FUNC(method) then
+        funcname := SHALLOW_COPY_OBJ(NAME_FUNC(opr));
+        if info <> false then
+            APPEND_LIST_INTR(funcname, " ");
+            APPEND_LIST_INTR(funcname, info);
```
stevelinton: Is there any merit in appending " method" or something similar for the case where info = false? Not sure.

fingolfin:
@ChrisJefferson please assign labels to your PRs.

lib/function.g
```diff
@@ -159,6 +159,31 @@ BIND_GLOBAL( "TYPE_OPERATION",
 ##
 DeclareOperationKernel("NameFunction", [IS_OBJECT], NAME_FUNC);
 
+#############################################################################
+##
+#O  NameFunction( <func> )  . . . . . . . . . . . . . . .  name of a function
```
fingolfin: This is wrong

lib/function.g
```diff
@@ -159,6 +159,31 @@ BIND_GLOBAL( "TYPE_OPERATION",
...
+##  true
+##  gap> HasNameFunction(x->x);
+##  true
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+##
+DeclareOperationKernel("HasNameFunction", [IS_OBJECT], HAS_NAME_FUNC);
```
fingolfin: Huhm... So along with `SetNameFunction`, this "operation" now kind of is an attribute... Perhaps we should turn it into one (a custom one, with custom storage, of course) ? Then people who want to implement custom objects just need to implement that attribute in some way. It would perhaps also remove the need to document `HasNameFunction` (and `SetNameFunction`)?

Of course on a technical level, actually doing that might be hard. So I am not asking that this should be done in this PR; I am just thinking aloud. Perhaps @stevelinton has some insights on this, too?

ChrisJefferson: I'll be honest, kernel attributes/methods is one of those things I do not really understand. I'm happy if someone can point me to something similar I can cut+paste

ChrisJefferson: Also it's a bit weird, as we return "unknown" even if HasNameFunction is false.

ChrisJefferson: Also it's a bit weird, as we return "unknown" even if HasNameFunction is false.

lib/function.g
```diff
@@ -159,6 +159,31 @@ BIND_GLOBAL( "TYPE_OPERATION",
 ##
 DeclareOperationKernel("NameFunction", [IS_OBJECT], NAME_FUNC);
 
+#############################################################################
+##
+#O  NameFunction( <func> )  . . . . . . . . . . . . . . .  name of a function
+##
+##  <#GAPDoc Label="HasNameFunction">
```
alex-konovalov: So you add `<#GAPDoc Label="HasNameFunction">` but this is not linked from anywhere in the manual - is this intended?

lib/function.g
```diff
@@ -159,6 +159,31 @@ BIND_GLOBAL( "TYPE_OPERATION",
...
+##
+##  <Description>
+##  returns if a function has an assigned name.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> NameFunction(SylowSubgroup);
+##  "SylowSubgroup"
+##  gap> HasNameFunction(SylowSubgroup);
+##  true
+##  gap> HasNameFunction(x->x);
```
alex-konovalov: Maybe I miss something, but how does this work?

alex-konovalov:
@ChrisJefferson I find the description for release notes "Give more library functions descriptive names when profiling" to be misleading. This PR does not seem to change any names - it intends to add ability to do this, and apparently adds new functions to do this, which should be then mentioned in release notes (and linked in their GAPDoc version to appropriate mansections).

lib/function.g
fingolfin: That's of course a valid point. But such behavior is not completely unheard of -- one can disable (both temporarily and permanently) "attribute storing" for select attributes or objects...

Of course I'd also be happy to change it to trigger an error if `NameFunction` is called when there is no name set, but then we'll likely have to adjust a bunch of code for that...

fingolfin: Anyway, this attribute would also have to be a "mutable attribute", seeing as the name can be changed later on.

I've not yet had time to look deeper into this, but a patch could start out like this (completely untested, and clearly incomplete, as e.g. `TYPE_FUNCTION_WITH_NAME` still needs to be defined in the library):
```patch
diff --git a/lib/function.g b/lib/function.g
index a48381e29..de02482c9 100644
--- a/lib/function.g
+++ b/lib/function.g
@@ -157,7 +157,7 @@ BIND_GLOBAL( "TYPE_OPERATION",
 ##  <#/GAPDoc>
 ##
 ##
-DeclareOperationKernel("NameFunction", [IS_OBJECT], NAME_FUNC);
+DeclareAttributeKernel("NameFunction", IS_OBJECT, NAME_FUNC);
 
 
@@ -174,7 +174,7 @@ DeclareOperationKernel("NameFunction", [IS_OBJECT], NAME_FUNC);
 ##  </Description>
 ##  </ManSection>
 ##
-DeclareOperationKernel( "SetNameFunction", [IS_OBJECT, IS_STRING], SET_NAME_FUNC );
+InstallMethod( "SetNameFunction", [IsFunction, IS_STRING], SET_NAME_FUNC );
 
 
diff --git a/src/calls.c b/src/calls.c
index d2694d39b..f45959de2 100644
--- a/src/calls.c
+++ b/src/calls.c
@@ -965,14 +965,19 @@ Obj ArgStringToList(const Char *nams_c) {
 */
 static Obj TYPE_FUNCTION;
 static Obj TYPE_OPERATION;
+static Obj TYPE_FUNCTION_WITH_NAME;
+static Obj TYPE_OPERATION_WITH_NAME;
 
 static Obj TypeFunction(Obj func)
 {
-    return ( IS_OPERATION(func) ? TYPE_OPERATION : TYPE_FUNCTION );
+    // TODO: return correct type depending on whether NAME_FUNC is set or not
+    if (NAME_FUNC(func) == 0)
+        return (IS_OPERATION(func) ? TYPE_OPERATION : TYPE_FUNCTION);
+    else
+        return (IS_OPERATION(func) ? TYPE_OPERATION_WITH_NAME : TYPE_FUNCTION_WITH_NAME);
 }
 
 
-
 **
 *F  PrintFunction( <func> )   . . . . . . . . . . . . . . .  print a function
@@ -1224,12 +1229,12 @@ static Obj FuncCALL_FUNC_LIST_WRAP(Obj self, Obj func, Obj list)
 
 **
-*F  FuncNAME_FUNC( <self>, <func> ) . . . . . . . . . . .  name of a function
+*F  AttrNAME_FUNC( <self>, <func> ) . . . . . . . . . . .  name of a function
 */
-static Obj NAME_FUNC_Oper;
+static Obj NameFuncAttr;
 static Obj SET_NAME_FUNC_Oper;
 
-static Obj FuncNAME_FUNC(Obj self, Obj func)
+static Obj AttrNAME_FUNC(Obj self, Obj func)
 {
     Obj                 name;
 
@@ -1659,6 +1664,18 @@ static StructGVarFilt GVarFilts [] = {
 };
 
 
+**
+*V  GVarAttrs . . . . . . . . . . . . . . . . .  list of attributes to export
+*/
+static StructGVarAttr GVarAttrs [] = {
+
+    GVAR_ATTR(NAME_FUNC, "func", &NameFuncAttr),
+    { 0, 0, 0, 0, 0 }
+
+};
+
+
 **
 *V  GVarOpers . . . . . . . . . . . . . . . . .  list of operations to export
@@ -1722,6 +1739,8 @@ static Int InitKernel (
     /* install the type functions                                          */
     ImportGVarFromLibrary( "TYPE_FUNCTION",  &TYPE_FUNCTION  );
     ImportGVarFromLibrary( "TYPE_OPERATION", &TYPE_OPERATION );
+    ImportGVarFromLibrary( "TYPE_FUNCTION_WITH_NAME",  &TYPE_FUNCTION_WITH_NAME  );
+    ImportGVarFromLibrary( "TYPE_OPERATION_WITH_NAME", &TYPE_OPERATION_WITH_NAME );
     TypeObjFuncs[ T_FUNCTION ] = TypeFunction;
 
     /* init filters and functions                                          */
```

lib/function.g
ChrisJefferson: How does what work?

ChrisJefferson:
Improved code (although I haven't switched to an attribute). Linked up docs, and also linked up the unused doc for SetNameFunction.

I put two things in the release notes, as I feel the better names are important (and this PR does add better names, that's the point). Adding "HasNameFunction" doesn't feel like something most people would care about, so it should probably be mentioned in some kind of "minor" section (or it can be put with all the other release notes, don't mind).

lib/function.g
alex-konovalov: That `x->x` has a name? It's not set in this example. Or it is autogenerated? What it will be in this case?

ChrisJefferson: Ah, there was a typo in the test, now fixed :)

lib/function.g
ChrisJefferson: While I agree this might have been a better idea from the start, I'd prefer not to change NameFunction to start returning errors, when previously it's always returned a valid string.

fingolfin: Actually, that's not necessary at all: Even right now, calling `FuncNAME_FUNC` actually sets the name of the function explicitly -- i.e., this works just like regular attributes work: if no value is set, we compute one on the fly and store it.

ChrisJefferson:
I'm happy to rebase this, is this PR fine as it is? (I don't mind if someone else wants to add #3430 on top later).

fingolfin:
@ChrisJefferson as @alex-konovalov noted, the title of this PR is misleading, and unsuitable for release notes.

lib/oper1.g
```diff
@@ -601,6 +601,27 @@ BIND_GLOBAL( "INSTALL_METHOD",
...
+
+    if IS_FUNCTION(rank) and not HAS_NAME_FUNC(rank) then
+        funcname := "Priority calculation for ";
+        APPEND_LIST_INTR(funcname, NAME_FUNC(opr));
+        if info <> false then
+            APPEND_LIST_INTR(funcname, " ");
+            APPEND_LIST_INTR(funcname, info);
+        fi;
+        SET_NAME_FUNC(rank, funcname);
+    fi;
```
fingolfin: Aha, I overlooked that you added this code. I'll have to read and understand that now, then.

fingolfin: OK, so this makes sense, I guess -- but there really should be tests that demonstrate this new feature, and also ensure it works and keeps working.

Also, PR #3430 is ready, and I kinda prefer that over the hackish addition of `HasNameFunction`, but of course I am biased. Perhaps we can get some other people to chime in with reviews.

fingolfin:
OK, so this now obviously needs to be rebased... sorry for the hassle.

ChrisJefferson:
I can't fix this, as I can't call HasNameFunction in oper.g.

fingolfin:
@ChrisJefferson `TESTER_FILTER(NAME_FUNC)`

ChrisJefferson:
Nope, segfaults

fingolfin:
Ahh, of course: As long as `HasNameFunction` is not yet defined, also the type objects of functions ( `TYPE_FUNCTION`, `TYPE_FUNCTION_WITH_NAME`, `TYPE_OPERATION` and `TYPE_OPERATION_WITH_NAME`) are not yet defined. But to test the filter, we need the type object. Hence this can't work.

fingolfin:
So one quick fix would be to check for `IsBound(TYPE_FUNCTION)` or `IsBound(HasNameFunction)`, and only call the tests in that case. This way, only very early method installations won't be covered (we are talking about a single digit number of them), but everything else later one should work.

ChrisJefferson:
Nope, still segfaults (see my pushed version, I check for `IsBound(HasNameFunction) and IsBound(TYPE_FUNCTION) and IsBound(TYPE_FUNCTION_WITH_NAME) and IsBound(TYPE_OPERATION_WITH_NAME) and VAL_GVAR("HasNameFunction")(method)`)

fingolfin:
Interesting. I can't reproduce a segfault here with your branch. I do get a strange error, though:
```
$ ./gap -A
#W Static module lib/oper1.g has CRC mismatch, ignoring
 ┌───────┐   GAP 4.10dev-2036-gae0e26a of today
Error, Record Element: '<rec>.GeneratorsOfMagmaWithInverses' must have an assigned value in
  GeneratorsOfGroup( F ) at /Users/mhorn/Projekte/GAP/gap.github/lib/grpfp.gi:1583 called from
<function "/ for a free group and an empty list of relators">( <arguments> )
 called from read-eval loop at /Users/mhorn/Projekte/GAP/gap.github/lib/grpfp.gi:5666
brk>
```

fingolfin:
I guess the cause for the error is that w/o your patch, I get this:
```
gap> NameFunction(GeneratorsOfMagmaWithInverses);
"GeneratorsOfMagmaWithInverses"
```
but with it, I get this, which then breaks stuff:
```
brk> NameFunction(GeneratorsOfMagmaWithInverses);
"FreeGeneratorsOfFpGroup for a free group"
```

I don't have time to track it down further right now, sorry.

ChrisJefferson:
Hopefully now all updated + cleaned up.

lib/oper1.g
```diff
@@ -601,6 +601,31 @@ BIND_GLOBAL( "INSTALL_METHOD",
       fi;
     fi;
 
+    if IS_FUNCTION(method) and IsBound(HasNameFunction) and
+      IsBound(TYPE_FUNCTION_WITH_NAME) and IsBound(TYPE_OPERATION_WITH_NAME) and
+      not VAL_GVAR("HasNameFunction")(method) then
```
fingolfin: So, I am in principle fine with keeping this as is, but here's an alternative that might perhaps be a bit easier to read, and which would ensure that the handful of methods which are installed before `HasNameFunction` is defined properly get names, too: Right before this function, insert `HasNameFunction := ReturnFalse;`, possibly with a comment explaining what's going on. Then `Unbind(HasNameFunction);` right before `HasNameFunction` is actually defined. And here, we can then simply return to the original form, i.e.:
```
     if IS_FUNCTION(method) and HasNameFunction(method) then
```
and of course similar for the rank function

PR #3422, opened 2019-04-24
kernel: rename expression and statement "TNUMs" to avoid confusion with actual TNUMs

fingolfin:
(This PR is motivated by repeated experiences trying to explain to people how GAP encodes function bodies. It is meant to make this code slightly more accessible.)

A long, long time ago, the statements and expressions which make up GAP function bodies were each coded as a separate GAP object. This is very inefficient for various reasons, so it was changed; now expressions and statements were simply a bunch of bytes within a single GAP object of type `T_BODY`, with a specific format. Yet they all retained their "identity" in the form of a  type number, such as `T_WHILE` to indicate a statement representing a while loop. To these day, the kernel code refers to these as "TNUMs".

This can be rather confusing, as those TNUMs have nothing to do anymore with the GAP object TNUMs; yet they do not only still have the same name, but also they all share the same `T_` prefix. E.g. we have `T_STRING` (an object tnum) and `T_STRING_EXPR` (an expression tnum).

We discussed this on slack some time ago. The consensus seemed to be that we should rename these, turning the `T_` into either `STAT_` or `EXPR_`, depending on the type.
One alternative would be to use the same prefix for both, as there really is no difference between the two on a purely technical (low) level; the main importance for distinguishing the two cases is in validation of the coded function's structure, resp. whether in knowing which dispatch table to look when evaluating them. But for that it's not necessary to give them different names...
Alas, since a lot of code now tries hard to distinguish precisely between the two, I think it makes sense to go with the split here for now; if one day in the future we decided to unify expressions and statement terminology again, doing it for these "tnums" will be about the easiest part anyway ;-). 

To perform the actual change, I wrote a little shell script, which I put into this PR. That makes it trivial to change things around: if we want another pair of prefixes: no problem. Also, if we merge some other PRs which will conflict with this PR (e.g. @sebasguts's PR #3371 which definitely should be merged before this PR), then it is trivial for me to recreate this PR.

Before this can be merged, some steps should be taken:
- [ ] make sure everybody is OK with this change (obviously)
- [x] merge other conflicting PRs, such as PR #3371
- [ ] come up with a better term than "statement or expression tnum" and use that consistently (perhaps "statement type" and "expression type" 
- [ ] to reduce confusion further, possibly rename a few existing identifiers in the kernel code which already have the `STAT_` prefix (there are none with `EXPR_` prefix), namely:
  - [ ] `STAT_HEADER`
  - [x] `STAT_MARK_CACHE`
  - [ ] `STAT_LVARS`
  - [ ] `STAT_LVARS_PTR`

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -135,7 +135,7 @@ UInt HaveInterrupt(void)
 **  then 0 is returned.
 **
 **  A statement sequence with <n> statements is represented by  a bag of type
-**  'T_SEQ_STAT' with  <n> subbags.  The first  is  the  first statement, the
+**  'STAT_SEQ_STAT' with  <n> subbags.  The first  is  the  first statement, the
 **  second is the second statement, and so on.
 */
 static ALWAYS_INLINE UInt ExecSeqStatHelper(Stat stat, UInt nr)
@@ -204,7 +204,7 @@ static UInt ExecSeqStat7(Stat stat)
 **  tell the  calling executor that a  leave-statement was executed).   If no
 **  leave-statement is executed, then 0 is returned.
 **
-**  An if-statement with <n> branches is represented by  a bag of type 'T_IF'
+**  An if-statement with <n> branches is represented by  a bag of type 'STAT_IF'
 **  with 2*<n> subbags.  The first subbag is  the first condition, the second
 **  subbag is the  first body, the third subbag  is the second condition, the
 **  fourth subbag is the second body, and so  on.  If the if-statement has an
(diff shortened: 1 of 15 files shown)


sebasguts:
I would be totally happy with this PR, and also distinguishing stats and exprs is valid imho.

I will try to fix #3371 soon (hopefully today or tomorrow) so we can get on with this. I assume you will leave the scripts in the PR, so I can apply them to other code I have already written using syntax trees? If not, it is also fine.

src/stats.c
```diff
@@ -330,7 +330,7 @@ static UInt ExecIfElifElse(Stat stat)
 **  then 0 is returned.
 **
 **  A for-loop with <n> statements  in its body   is represented by a bag  of
-**  type 'T_FOR' with <n>+2  subbags.  The first  subbag is an assignment bag
+**  type 'STAT_FOR' with <n>+2  subbags.  The first  subbag is an assignment bag
```
fingolfin: There are still lots of these comments which claim that statements and expressions are "represented by a bag", and which talk about "subbags", and so on. I think it is out of scope for this PR, but it would be kinda nice to improve that.

fingolfin:
From my point of view, this could be merged now.

sebasguts:
I would be very happy if this can get merged. It conflicts with #3441 which I can update afterwards

PR #3424, opened 2019-04-24
kernel: introduce struct ScannerState

fingolfin:
(This is another PR I've had sitting on my HD for over a year now, so instead of letting it bitrot, I figured I might as well see if I can get some feedback on it).

This is a step towards fully modularizing the scanner, making it reusable
from external code (e.g. to implement a code formatting tool, for use in
future advanced versions of AutoDoc and what not else).

The vision is that ultimately, there won't be a ScannerState member of struct
GAPState; rather, code that needs to use a scanner will instantiate one
locally, and release it after use.

More work is needed to get there, most importantly, a clean interface with
the input code needs to be defined. This is desirable for other reasons, too;
e.g. if we can move the line buffering into the scanner, then we can benefit
from inlining of e.g. GET_NEXT_CHAR, which may result in a noticeable speedup
in GAP's startup time.

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -17,6 +17,8 @@
 
 #include "system.h"
 
+#include "scanner.h"
+
 #if defined(HPCGAP)
 #include "hpc/tls.h"
 #endif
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
Looks good. I have only one comment. At the definition of ScannerState, it might be worth having a comment that says we don't support (or at least, don't test) creating more than one ScannerState yet, just in case someone comes along and tries making one.

fingolfin:
@ChrisJefferson that's a good suggestion. I also added a TODO comment into `gapstate.h`.

fingolfin:
@ChrisJefferson Everything OK now?

ChrisJefferson:
yup

Issue #2194, opened 2018-02-19
"\200" and "\377" show as "?" on SPARC Solaris 11


dimpase:
### Observed behaviour
```
gap> x:="\0xFF";
"�"
gap> "\377";
"�"
gap> "\200";
"�"
```

### Expected behaviour
```
gap> x:="\0xFF";
"\377"
gap> "\377";
"\377"
gap> "\200";
"\200"
```

```
 *********   GAP 4.8.8-5085-gb6d4e8e-dirty of today
 *  GAP  *   https://www.gap-system.org
 *********   Architecture: sparc-sun-solaris2.11-default64
 Configuration:  gmp 6.1.2, KernelDebug
 Loading the library and packages ...
 Packages:   AClib 1.2, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.8, CRISP 1.4.4, Cryst 4.1.13, CrystCat 1.1.6, CTblLib 1.2.2, FactInt 1.6.1, FGA 1.3.1,
             GAPDoc 1.6.1, IRREDSOL 1.4, LAGUNA 3.8.0, Polenta 1.3.8, Polycyclic 2.11, PrimGrp 3.3.0, RadiRoot 2.7, ResClasses 4.7.1, smallgrp 1.2,
             Sophus 1.23, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.53
```
this is master git of 19.02.2018, built with gcc 5.4.0.
(same with 4.9 beta of 01st of Feb).

-----------------------
`\0x7E` and `\0x7F` (i.e. octal 176 and 177) still work, but `\0x80` (i.e. octal 200) does not.
So this looks like that only rightmost 17 bits work, and setting any bit to the left of
the rightmost 17 does not work correctly.  

Note that the **chars** `'\200'`, etc work as they should!

------------------------
In case, here are locale settings:
```LANG=en_GB.UTF-8
LC_ALL=
LC_COLLATE=
LC_CTYPE=
LC_MESSAGES=
LC_MONETARY=
LC_NUMERIC=
LC_TIME=
```

-----------------------
Also, here is what I see running tests in `testinstall` for the 4.9 beta.
```
Diff in /datapool/dima/Sage/gap-4.9.0/tst/testinstall/strings.tst:44
# Input is:
x:="\0xFF";
# Expected output:
"\377"
# But found:
"<FF>"
########
[...]

PR #3428, opened 2019-04-29
Fix GNU readline support on OpenBSD, and make the configure test for it more robust

### LINKED ISSUE #3426: openbsd support: readline treatment
### Observed behaviour
the default readline on openbsd is version 4 (the last LGPL version, I guess), and it lacks `rl_bind_keyseq`.
Thus, linking bails out with an undefined ref to it.

### Expected behaviour
A minimal guard against this should have a test for `rl_bind_keyseq`. To actually get it all
working, one should use `libereadline` from `/usr/local`. 
I'll send a PR to fix this, or at least to test for `rl_bind_keyseq`.

dimpase:
# Description

this fixes #3426.

## Text for release notes 

Better handling of incompatible with GAP readline libraries.
Enable building of GAP on OpenBSD (with ereadline fixed as in 
https://marc.info/?l=openbsd-ports&m=155646111509306&w=2) 

## Further details

Do not forget to run ./autogen + ./configure before testing this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -537,31 +537,56 @@ AC_ARG_WITH([readline],
 dnl Now check if we can find GNU readline. We go to some extra efforts to
 dnl ensure it is GNU readline, and not e.g. BSD editline wrappers for
 dnl readline, which do not suffice for GAP.
+dnl
+dnl note that OpenBSD installs modern (v6+) GNU readline into /usr/local under name libereadline
+dnl and its headers into /usr/local/include/ereadline.
+dnl
 AS_IF([test "x$with_readline" != xno],[
-  READLINE_LIBS="-lreadline"
-  AS_CASE([x"$with_readline"],
-    [xyes|x],[
-      READLINE_CPPFLAGS=
-      READLINE_LDFLAGS=
-    ],
-    [*],[
-      READLINE_CPPFLAGS="-I$with_readline/include"
-      READLINE_LDFLAGS="-L$with_readline/lib"
-    ]
-  )
-
-  dnl We check for the symbol rl_ding to distinguish GNU readline
-  dnl from other readline implementation.
-  AX_CHECK_LIBRARY([READLINE], [readline/readline.h], [readline], [rl_ding], [], [], [])
-
-  AS_IF([test $ax_cv_have_READLINE = yes],[
-    AC_DEFINE([HAVE_LIBREADLINE], [1], [Define if you have libreadline])
-  ],[
-    AS_CASE([x"$with_readline"],
+ dnl Checking for rl_bind_keyseq rules out very old readline
+ dnl ereadline is the name of GNU readline library on OpenBSD
+ AS_CASE([x"$with_readline"],dnl set up needed -L flags
+    [xyes|x], [READLINE_LDFLAGS=],
+    [*], [READLINE_LDFLAGS="-L$with_readline/lib"])
+ save_LDFLAGS="$LDFLAGS"
+ LDFLAGS="$LDFLAGS $READLINE_LDFLAGS"
+ save_LIBS="$LIBS"
+ AC_SEARCH_LIBS([rl_bind_keyseq], [readline ereadline], [
+   AS_IF([test x$ac_cv_search_rl_bind_keyseq != "xnone required"],
+    [READLINE_LIBS=$ac_cv_search_rl_bind_keyseq], [READLINE_LIBS=])
+   AS_CASE([x"$with_readline"],dnl libereadline is a special case
+    [xyes|x], [
+      AS_IF([test x$ac_cv_search_rl_bind_keyseq = "x-lereadline"], [
+	READLINE_CPPFLAGS="-I/usr/local/include/ereadline"
+	], [
+	READLINE_CPPFLAGS=])
+    ], [*],[
+      AS_IF([test x$ac_cv_search_rl_bind_keyseq = "x-lereadline"], [
+	READLINE_CPPFLAGS="-I$with_readline/include/ereadline"
+	], [
+	READLINE_CPPFLAGS="-I$with_readline/include"])
+      ]
+   )
+  dnl We check that the corresponding header is present
+   save_CPPFLAGS="$CPPFLAGS"
+   CPPFLAGS="$READLINE_CPPFLAGS $CPPFLAGS"
+   AC_CHECK_HEADERS([readline/readline.h], [
+        AC_DEFINE([HAVE_LIBREADLINE], [1], [Define if you have lib(e)readline])
+	], [
+	AS_CASE([x"$with_readline"],
+            [x], [READLINE_CPPFLAGS= READLINE_LDFLAGS= READLINE_LIBS=],
+            [*], [AC_MSG_FAILURE([--with-readline was given, but further tests for readline failed])])
+        ]
+   )
+   CPPFLAGS="$save_CPPFLAGS"
+   break
+   ],[
+   AS_CASE([x"$with_readline"],
       [x], [READLINE_CPPFLAGS= READLINE_LDFLAGS= READLINE_LIBS=],
       [*], [AC_MSG_FAILURE([--with-readline was given, but test for readline failed])]
-    )
-  ])
+   )
+ ])dnl AC_SEARCH_LIBS([rl_bind_keyseq] ...) ends here
+ LDFLAGS="$save_LDFLAGS"
+ LIBS="$save_LIBS"
 ])
 
 AC_SUBST([READLINE_CPPFLAGS])


configure.ac
```diff
@@ -421,16 +425,23 @@ AS_IF([test "x$with_readline" != xno],[
...
   dnl We check for the symbol rl_ding to distinguish GNU readline
   dnl from other readline implementation.
-  AX_CHECK_LIBRARY([READLINE], [readline/readline.h], [readline], [rl_ding], [], [], [])
-
+  case "$ac_cv_search_rl_bind_keyseq" in
+     *ereadline)
+       AX_CHECK_LIBRARY([READLINE], [ereadline/readline/readline.h], [ereadline], [rl_ding], [], [], [])
+       ;;
+     *)
+       AX_CHECK_LIBRARY([READLINE], [readline/readline.h], [readline],  [rl_ding], [], [], [])
```
fingolfin: I don't really like that now we are mixing `AC_SEARCH_LIBS` and `AX_CHECK_LIBRARY`, that kind of defeats the purpose of the latter, and also adds a redundant linker check (though perhaps if you always check for `rl_ding`, resp. always check for `rl_bind_keyseq`, autoconf's caching might take care of that...).

One alternative would be to simply use `AC_CHECK_HEADER` here, as we already performed the linker check.

configure.ac
```diff
@@ -421,16 +425,23 @@ AS_IF([test "x$with_readline" != xno],[
...
   AS_IF([test $ax_cv_have_READLINE = yes],[
     AC_DEFINE([HAVE_LIBREADLINE], [1], [Define if you have libreadline])
   ],[
     AS_CASE([x"$with_readline"],
       [x], [READLINE_CPPFLAGS= READLINE_LDFLAGS= READLINE_LIBS=],
-      [*], [AC_MSG_FAILURE([--with-readline was given, but test for readline failed])]
+      [*], [AC_MSG_FAILURE([--with-readline was given, but further tests for readline failed])]
     )
   ])
+ ], [AC_MSG_FAILURE([--with-readline was given, but first test for readline failed])])
```
fingolfin: Hmm, is that true? Looking at the code now, it seems to me this check also is run if `--with-readline` is not given; the only way to bypass it is an explicit `--without-readline` resp. `--with-readline=no`. So that would mean we now force everybody who does not have GNU readline installed to pass one of those explicitly, which is not what we want.

But perhaps I am misreading the code -- have you tested this scenario?

dimpase: oops, I've forgotten to add the condition, to bail out only if readline is requested. Fixed by latest commit.

configure.ac
dimpase: I suppose one can simply get rid of these calls, and do everything needed (calling `AC_DEFINE` and setting variables) within the scope of the `AC_SEARCH_LIBS` call.
I can do this change if you like.

configure.ac
```diff
@@ -407,7 +406,12 @@ dnl Now check if we can find GNU readline. We go to some extra efforts to
 dnl ensure it is GNU readline, and not e.g. BSD editline wrappers for
 dnl readline, which do not suffice for GAP.
 AS_IF([test "x$with_readline" != xno],[
-  READLINE_LIBS="-lreadline"
+ dnl Checking for rl_bind_keyseq rules out very old readline
+ dnl as well as editline pretending to be readline (yes, looking at you, OSX!)
+ dnl ereadline is the name of GNU readline library on OpenBSD
+ AC_SEARCH_LIBS([rl_bind_keyseq], [readline ereadline], [
```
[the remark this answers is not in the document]
dimpase: yes, AC_SEARCH_LIBS does prepend the 1st found library to `LIBS`, but isn't it what we want all along?

configure.ac
```diff
@@ -407,7 +406,12 @@ dnl Now check if we can find GNU readline. We go to some extra efforts to
 dnl ensure it is GNU readline, and not e.g. BSD editline wrappers for
 dnl readline, which do not suffice for GAP.
 AS_IF([test "x$with_readline" != xno],[
-  READLINE_LIBS="-lreadline"
+ dnl Checking for rl_bind_keyseq rules out very old readline
+ dnl as well as editline pretending to be readline (yes, looking at you, OSX!)
```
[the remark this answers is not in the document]
dimpase: OK, I agree that this does not affect OSX. I was confused here. I've removed the OSX comment in the update.

fingolfin:
I adjusted the title of this PR to be more suitable for release notes; please re-edit if you disagree; you may also want to change the commit message of your first commit, as it still reference OSX; and squash the second commit into the first.

configure.ac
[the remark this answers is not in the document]
fingolfin: No, we don't want this in LIBS, we want it in READLINE_LIBS.

configure.ac
fingolfin: Yes, that sounds better. If you were willing to do that, would be great!

DominikBernhardt:
@dimpase Ping.

dimpase:
yep, I should get back to this.

configure.ac
dimpase: It appears the check for `rl_ding` (appeared in readline 4.2) is strictly weaker than the check for  `rl_bind_keyseq` (appeared in readline 5.0). So I'll just check for headers, instead of calling  `AC_SEARCH_LIBS` or `AX_CHECK_LIBRARY`.

dimpase: We overlooked that due to non-standard header location on OpenBSD one should add
`-I/usr/local/include/ereadline` to `READLINE_CPPFLAGS` for this case. 
And also make sure that `AC_CHECK_HEADER` picks the right place by adding `READLINE_CPPFLAGS` to `CFLAGS` before calling it.

configure.ac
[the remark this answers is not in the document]
dimpase: all done today, pushed a complete reworking of this PR

configure.ac
dimpase: done

dimpase:
Ready for review

fingolfin:
@dimpase thanks for the update, I'll take a look. But the commit message still is odd; it says "also, do not wrongly accuse OSX of wrongdoing" but the only other reference to OSX is in the first line of the commit message? So maybe just remove both mentions -- or does OSX have any relevance for this PR after all?

dimpase:
> commit message still is odd

It probably referred to some already removed comment in configure.ac. I'll edit the commit message, no problem. Needless to say, this change works on OSX if `--with-readline=/foo/bar` is provided, and readline installed in `/foo/bar`.

It also should be mentioned that this fixes a bug: testing for `rl_ding` did not identify the needed readline version, so now the test for a newer `rl_...` finction.

dimpase:
imroved commit message, no other changes

configure.ac
```diff
@@ -537,31 +536,56 @@ AC_ARG_WITH([readline],
...
+ save_LDFLAGS="$LDFLAGS"
+ LDFLAGS="$LDFLAGS $READLINE_LDFLAGS"
+ save_LIBS="$LIBS"
+ AC_SEARCH_LIBS([rl_bind_keyseq], [readline ereadline], [
+   AS_IF([test x$ac_cv_search_rl_bind_keyseq != "xnone required"],
+    [READLINE_LIBS=$ac_cv_search_rl_bind_keyseq], [READLINE_LIBS=])
+   AS_CASE([x"$with_readline"],dnl libereadline is a special case
+    [xyes|x], [
+      AS_IF([test x$ac_cv_search_rl_bind_keyseq = "x-lereadline"], [
+	READLINE_CPPFLAGS="-I/usr/local/include/ereadline"
```
fingolfin: So, just to double-check: that means they have the header file in `/usr/local/include/ereadline/readline/readline.h` (three times (e)readline in a row) ?

I assume this was tested on an actual OpenBSD system?

configure.ac
```diff
@@ -537,31 +536,56 @@ AC_ARG_WITH([readline],
...
+   )
+   CPPFLAGS="$save_CPPFLAGS"
+   break
+   ],[
+   AS_CASE([x"$with_readline"],
       [x], [READLINE_CPPFLAGS= READLINE_LDFLAGS= READLINE_LIBS=],
-      [*], [AC_MSG_FAILURE([--with-readline was given, but test for readline failed])]
-    )
-  ])
+      [*], [AC_MSG_FAILURE([--with-readline was given, but first test for readline failed])]
```
fingolfin:
```suggestion
      [*], [AC_MSG_FAILURE([--with-readline was given, but test for readline failed])]
```
No need to say "first" here (and "further" above, as `configure` shows before that line which test failed.

configure.ac
```diff
@@ -527,7 +527,6 @@ AC_SUBST([ZLIB_CPPFLAGS])
 AC_SUBST([ZLIB_LDFLAGS])
 AC_SUBST([ZLIB_LIBS])
 
-
```
fingolfin: Perhaps add that line back to reduce the gratuitous diff?

configure.ac
dimpase: Yes, this is correct. Tested on an actual OpenBSD x86_64 system too. I don't know why they came to the idea of `include/ereadline`, perhaps they also have some package that installs /usr/local/include/readline/readline.h (the system readline, an old LGPL GNU readline 4, has header in /usr/include)

configure.ac
```diff
@@ -537,31 +536,56 @@ AC_ARG_WITH([readline],
 dnl Now check if we can find GNU readline. We go to some extra efforts to
 dnl ensure it is GNU readline, and not e.g. BSD editline wrappers for
 dnl readline, which do not suffice for GAP.
+dnl
+dnl note that OpenBSD installs modern readline into /usr/local under name libereadline
```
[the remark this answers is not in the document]
dimpase: modern, i.e. version 5+, GNU readline. (GAP doesn't fly with GNU readline 4).

configure.ac
dimpase: will do

configure.ac
dimpase: OpenBSD 6.6, to be precise. I beleive 6.4 had underlinked (e)readline, e.g. one also needed to link with  `-ltinfo` or something like this.

configure.ac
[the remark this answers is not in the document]
fingolfin: OK, that makes sense. Then I'd suggest this change to make this clearer (possibly rewrap)
```suggestion
dnl note that OpenBSD installs modern GNU readline (i.e., >= 5.x) into /usr/local under the name libereadline
```

configure.ac
fingolfin: Cool, thanks

configure.ac
dimpase: cf 4df77b2

configure.ac
dimpase: fixed by 4df77b2

configure.ac
dimpase: added more detail on this in 4df77b2

dimpase:
@kaashif - perhaps this is now ready for making an OpenBSD GAP package.

dimpase:
squashed. Thanks for taking care of this.

fingolfin:
@ChrisJefferson @alex-konovalov I'd like to backport to this to stable-4.11, OK?

ChrisJefferson:
That's fine with me

kaashif:
@dimpase Maybe. I'll make another pass on the weekend.

fingolfin:
For packaging, I recommend talking with/to us -- we can often provide hints how to overcome certain problems. Even if you get stuck and decide to abandon the effort, we are grateful for any constructive feedback, to help improve things (In the end, lack of time/resources/contributors sadly is always a limiting factor)

dimpase:
@kaashif so to build on OpenBSD, one needs to `pkg_add`  `gmp` and `ereadline` and run
```
./autogen.sh
CC=clang CXX=clang++ ./configure --with-readline=/usr/local --with-gmp=/usr/local
gmake
```
installation should be tested with `LC_CTYPE=en_US.UTF-8` - otherwise there are weird UTF-8 problems  - see #2194

fingolfin:
Backported to stable-4.11 as f67d98af43dfdd282d4f94f6dd0edd558b87d58e

PR #3429, opened 2019-04-29
kernel: fix off-by-one error in FuncREM_SET

### LINKED ISSUE #3427: Suspicious code in FuncREM_SET
(The following is partly an info-dump, so it isn't lost)

Commit `3c36b2241b` contains the code:

```
@@ -527,11 +522,8 @@ static Obj FuncREM_SET(Obj self, Obj set, Obj obj)
     /* remove the element from the set if it is there                      */
     if ( pos <= len && EQ( ELM_PLIST(set,pos), obj ) ) {
 
-        ptr = PTR_BAG(set) + pos;
-        for ( i = pos; i < len; i++ ) {
-            *ptr = *(ptr+1);
-            ptr ++;
-        }
+        Obj * ptr = ADDR_OBJ(set) + pos;
+        SyMemmove(ptr, ptr + 1, sizeof(Obj) * (len - pos + 1));
         SET_ELM_PLIST( set, len, 0 );
```

I think this has an off-by-one error in it. valgrind certainly flags it up.

fingolfin:
This was in no released version of GAP, hence does not need a release notes entry.

Fixes #3427.

### DIFF
--- a/src/set.c
+++ b/src/set.c
@@ -523,7 +523,7 @@ static Obj FuncREM_SET(Obj self, Obj set, Obj obj)
     if ( pos <= len && EQ( ELM_PLIST(set,pos), obj ) ) {
 
         Obj * ptr = ADDR_OBJ(set) + pos;
-        SyMemmove(ptr, ptr + 1, sizeof(Obj) * (len - pos + 1));
+        SyMemmove(ptr, ptr + 1, sizeof(Obj) * (len - pos));
         SET_ELM_PLIST( set, len, 0 );
         SET_LEN_PLIST( set, len-1 );
 


src/set.c
```diff
@@ -523,7 +523,7 @@ static Obj FuncREM_SET(Obj self, Obj set, Obj obj)
     if ( pos <= len && EQ( ELM_PLIST(set,pos), obj ) ) {
 
         Obj * ptr = ADDR_OBJ(set) + pos;
-        SyMemmove(ptr, ptr + 1, sizeof(Obj) * (len - pos + 1));
+        SyMemmove(ptr, ptr + 1, sizeof(Obj) * (len - pos));
```
fingolfin: To illustrate why this is right (and was wrong before): If `pos == len`, then we are removing the last element, and 0 bytes need to be moved. If `pos == len-1`, then we need to move `len - pos == 1` object pointers; and so on.

ChrisJefferson:
Thanks. I'm not sure this actually causes any problems, but it's worth fixing. I'll merge it once tests cycle.

PR #3430, opened 2019-04-29
Turn NameFunction into an attribute

fingolfin:
This is a potential alternative to PR #3420. It needs some polishing and tests, but at least for some quick tests, seems to work reasonably well. I won't have time to work more on it for now, so if somebody else wants to continue with this, feel free to do so.

UPDATE: I have now improved this PR, it should be ready for merging.

### DIFF
--- a/lib/function.g
+++ b/lib/function.g
@@ -77,6 +77,48 @@ DeclareCategoryKernel( "IsOperation",
 
 #############################################################################
 ##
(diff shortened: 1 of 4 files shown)


fingolfin:
I just updated this PR, cleaning it up a bit, and adding some tests to ensure that the kernel attribute works as intended.

PR #3441, opened 2019-05-08
Added VALUE_EXPR

sebasguts:
to store an arbitrary value in an expression

Thank You!

# Description

Adds `VALUE_EXPR` expressions, which can store an arbitrary value. They can only coded using `SYNTAX_TREE_CODE` at the moment.

## Further details

- [ ] proper formatting

- [ ] runnable tests


sebasguts:
Thanks for the comments. I addressed both of them in a sense:

I changed the name from `string` to `obj`. However, I decided to remove the copying completely, as we probably do not even want to do a shallow copy here (if we want to use this for constants). I instead make sure that if a value is added to the syntax tree, an immutable copy is added.

src/exprs.c
```diff
@@ -1063,6 +1063,20 @@ static Obj EvalFloatExprEager(Expr expr)
 }
 
 
+/****************************************************************************
+**
+*F  EvalValueExpr(<expr>)                 eval value expressions to its value
```
fingolfin:
```suggestion
*F  EvalValueExpr(<expr>) . . . . . . . . eval value expressions to its value
```

src/exprs.c
```diff
@@ -1063,6 +1063,20 @@ static Obj EvalFloatExprEager(Expr expr)
...
+**
+**  'EvalValueExpr' evaluates the value  expression  <expr>  to its value.
```
fingolfin:
```suggestion
**  'EvalValueExpr' evaluates the value expression <expr> to its value.
```

src/exprs.c
```diff
@@ -1585,6 +1599,19 @@ static void PrintFloatExprEager(Expr expr)
     }
 }
 
+/****************************************************************************
+**
+*F  PrintValueExpr(<expr>) . . . . . . . . . . . . print a value expression
```
fingolfin:
```suggestion
*F  PrintValueExpr(<expr>) . . . . . . . . . . . . . print a value expression
```

fingolfin:
Added labels. This does not (yet) need to be in the release notes, as there are no user facing changes other than in the syntax tree code, which is new anyway.

sebasguts:
I added your suggestions, thank you :)

src/code.h
```diff
@@ -488,6 +488,8 @@ enum EXPR_TNUM {
     T_FLOAT_EXPR_EAGER,
     T_FLOAT_EXPR_LAZY,
 
+    T_VALUE,
```
fingolfin: Please rebase this PR and rename this to EXPR_VALUE

sebasguts:
Done :)

fingolfin:
@sebasguts the title of this PR and also the commit message should also be adjusted... :-)

sebasguts:
@fingolfin done.

src/syntaxtree.c
```diff
@@ -549,7 +549,7 @@ static Expr SyntaxTreeCodeValue(Obj node)
     UInt1 tnum = GetTypeTNum(node);
     Obj   value = ElmRecST(tnum, node, "value");
     Expr  expr = NewStatOrExpr(tnum, sizeof(UInt), 0);
-    Int   ix = AddValueToBody(value);
+    Int   ix = AddValueToBody(CopyObj(value, 0));
```
fingolfin: I am really torn on this whole concept. On the one hand, it is quite powerful and the implementation is very easy (thanks to the recent addition of `AddValueToBody` and the whole code driving it).
But at the same time, it's a major departure from how things worked so far. Previously, you could print a function, and that gave you code text that could be read in to essentially reproduce the function body. This is now now the case anymore. Perhaps that's not a problem, though (and strictly speaking, this already could happen if you printed a nested function containing hvar references; but there, at least you could have printed the parent function to "fix" it.

Now, I could provide as value, say, an immutable string. Printing the resulting function would give something that indistinguishable from the same function with a string expression, but they'd behave differently. Or, even worse, I could use a *mutable* string. Then it'd look the same, but would work radically differently. E.g the function might modify that string value, and after each call to it, printing the function could show something else.

Don't get me wrong, I am not intending to shoot this down, but as always I'd like us to think through some of the long-term consequences first... and perhaps come up with some ideas how they might be addressable. That is also tightly related to the question how to make these "values" accessible without using the syntax tree (de)coding.

One idea for that would be to add some new syntax to mark an expression, which says: "when coding, evaluate this expression immediately and insert a reference to the resulting value". So one could write something very roughly in the vein of this (before @sebasguts gets me wrong on this, too: Of course I am not saying these things must be implemented before this PR can be merged; what I want is that we at least verify that we see a way forward, before possibly ourselves into a corner!):
```
gap> f := x -> x + EvalImmediately( 2^30 );;
gap> Print(f);
function ( x )
    return x + EvalImmediately( 1073741824 );
end
gap> g := function(x) local y; y := EvalImmediately( [] ); Add(y, x); return y; end;;
gap> Display(g);
function ( x )
    local y;
    y := EvalImmediately( [  ] );
    Add( y, x );
    return y;
end
gap> g(1);
[ 1 ]
gap> g(2);
[ 1, 2 ]
gap> Display(g);
function ( x )
    local y;
    y := EvalImmediately( [ 1, 2 ] );
    Add( y, x );
    return y;
end
```
To clarify how this is meant to work: `EvalImmediately` here would be a syntactic element, not a function, i.e., similar to how `IsBound` and `Unbind` work. It would only be valid in coding mode (though I guess we could allow it in interpreted mode, as a no-op). When reading `EvalImmediately( expr )`, then `expr` is evaluated immediately when the final `)` is read; thus no references to hvars (arguments or local variables) of the surrounding functions are possible, as those are not bound at parsing time. Once the expression is evaluated, it is inserted as a `EXPR_VALUE` into the function body. And printing an `EXPR_VALUE` would be done by printing `EvalImmediately( VALUE )` instead of just `VALUE`. 

I think that's relatively straight forward to explain and implement. But of course it's still far from perfect, i.e., it does not capture the full power of `EXPR_VALUE` -- so we may want to make it even more general, or restrict what we allow as `EXPR_VALUE` ...
For example, it cannot reflect mutability:
```
gap> h1 := {} -> EvalImmediately( MakeImmutable("string") );;
gap> h2 := {} -> EvalImmediately( "string" );;
gap> Display(h1);
function (  )
    return EvalImmediately( "string" );
end
gap> Display(h2);
function (  )
    return EvalImmediately( "string" );
end
```
So `h1` and `h2` are printed identically, and both return always the same strings; but in one case its mutable, in the other immutable. Oh, and of course, I can do this: `MakeImmutable(h2());` and afterwards the string returned by `h2` is immutable, too...

The specific problem of immutability could be addressed, of course, in various ways (say by adding a special case to the printing code to output `EvalImmediately( MakeImmutable( VALUE ) )` instead of just `EvalImmediately( VALUE )`; but of course that's not enough as there are many objects where the printed form is not enough to reconstruct the object.

BTW, the syntax tree (de)coding also cannot capture the full power of this concept due to the `CopyObj` call you inserted above; and also printing and re-reading the resulting text in general won't produce identical results. Here is an example for that:
```
gap> list := [];
[  ]
gap> func := function(x) Add( EvalImmediately( list ), x ); return EvalImmediately( list ); end;;
gap> Display(func);
function ( x )
    Add( EvalImmediately( [ ] ), x );
    return EvalImmediately( list );
end
gap> func := function(x) Add( EvalImmediately( list ), x ); return EvalImmediately( list ); end;;
gap> func(1);
[ 1 ]
gap> list;
[ 1 ]
gap> Display(func);
function ( x )
    Add( EvalImmediately( list ), x );
    return EvalImmediately( list );
end
```
That kind of code is at the same time pretty cool and pretty horrific (in terms of long-term maintainability). Perhaps one should prevent it, by explicitly specifying that `EvalImmediately` inserts a *copy* of the evaluated object?

ChrisJefferson:
Sorry, shoved a do not merge on, as maxs comment is important and I'm on my phone so didn't know how to just remove my positive review

ChrisJefferson:
Another option - - what is this saving in principle over a global variable? Obviously it saves you having to make up some kind of random name that won't be used elsewhere, and a little efficiency, and nicer printing. Is there some specific use case we have for t_value? 

We could limit t_value to only be a fixed list of types, which we know we can print and read well?

src/syntaxtree.c
sebasguts: You are really right about this concerns. Let me address some of them with comments (as I have absolutely no solution, just additional comments)

* For clarification: You do not want to make the functions readable from their display string using this `EvalImmediately`, right? You just want to indicate that this value is somehow special. Because, if you want to make it fully evaluateable again, one needs to store the input string, too, and make it completely static (i.e., not depending on anything global). Furthermore, when working with Syntax Trees, one might not want to even create such an input string, as it might be between hard and up to impossible. However, Just printing the value would certainly be fine for me.

* About the copying: We can just drop the copying when coding the syntax tree. I did not had any special thoughts in mind when doing so, so not making copies at all is not a problem.

* For copying semantics of `EvalImmediately`: We could probably make it produce a `ShallowCopy`, since it already looks like a function call anyway. One probably does not want to have a `DeepCopy` in there always, since this could just be added manually if somebody really wants to copy the value, but if it is the default, it again removes possibilities.

sebasguts:
I am actually against restricting the `T_VALUE` to certain types. I am not sure how much benefit the "rereadability" of a function really has. I am also not sure how much benefit this has.

Of course, this can be achieved by creating global variables, too, but then, this also does not let you copy printed out function code between gap sessions.

ChrisJefferson:
The problem (and this is a long-running problem), is many, many types can't be read back in from their printed out form, such as most of the small and transitive groups libraries.

sebasguts:
@ChrisJefferson I understand the problem, and I guess we will never reach a point where all objects can print out their full information that is necessary to read them back in (even in the library, not to speak about the packages).

I just wanted to mention that storing values in a global variable does not really change this problem either.

ChrisJefferson:
Sorry, I should be more clear with my aim for using global variables.

Indeed global variables aren't readable back in in another GAP session. However, using a global variable does produce valid GAP code, where it is (hopefully) obvious why the code doesn't work in another GAP -- missing variable. On the other hand, trying to put variables printed out into functions can produce what the GAP parser would view as nonsense, there are groups which print out things like '2[x]2' and such like.

So, it depends on what the goal is. I think trying to make functions with arbitary GAP objects embedded in them, which can then be printed out, just isn't going to work without confusing people for years to come. Therefore it would be useful to know what the aim here is -- I'm assuming this is mainly for syntax tree?

sebasguts:
My main use was some kind of constant folding. So yeah, coding it via Syntax Trees. That is why I am not really concerned about the possibility to read them back in, as they would only appear in somehow `generated` functions.  My main goal was to have the possibility to write something like
```
#@ eval immediately
x := Foo(bar);
```
in functions, which some function can then compile to
```
x := <Return value of Foo(bar)>;
```
using `VALUE_EXPR`.

ChrisJefferson:
In that case, we could go the other way, make clear these functions will never read back in, and then people can just accept optimised functions aren't readable by GAP from their printed out form?

Trying to think of some nice notation, which definitely isn't valid GAP. Perhaps:

`< The constant ... >`, where ... is the result of the printing? I believe anywhere a constant could go, you couldn't have a `<`?

fingolfin:
I don't think this will go anywhere, esp. with the main author gone ;-) but also in general... We may of course revive this or the ideas in here in the future if we see a need.

PR #3444, opened 2019-05-08
Remove '-a', other small input cleanups

ChrisJefferson:
This performs some small cleanups to GAP's command line arguments. The main visible difference is removing `-a`, which didn't actually do anything useful on modern OSes (it allocated, then freed, a bunch of memory).

For release notes:

* Remove obsolete '-a' option

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -990,8 +990,6 @@ static Int enableMemCheck(Char ** argv, void * dummy)
 #endif
 
 
-static Int preAllocAmount;
-
 /* These are just the options that need kernel processing. Additional options will be 
    recognised and handled in the library */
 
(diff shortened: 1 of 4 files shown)


src/system.c
```diff
@@ -1006,7 +1004,6 @@ static struct optInfo options[] = {
   { 'L', "", storeString, &SyRestoring, 1}, /* must be handled in kernel  */
   { 'M', "", toggle, &SyUseModule, 0}, /* must be handled in kernel */
   { 'R', "", unsetString, &SyRestoring, 0}, /* kernel */
-  { 'a', "", storeMemory, &preAllocAmount, 1 }, /* kernel -- is this still useful */
```
fingolfin: Please also remove the documentation for this in `doc/ref/run.xml`

fingolfin: Also, `INSTALL.md` mentions `-a`.

ChrisJefferson: Woops, should have searched harder. Will fix!

ChrisJefferson: That install.md comment is just wrong (in my opinion), the run.xml does at least give me some idea why this option originally existed, but I checked the main OSes we run on and none of them seem to require -a anymore.

src/system.c
```diff
@@ -1024,7 +1021,7 @@ static struct optInfo options[] = {
   { 'P', "", storePosInteger, &SyNumProcessors, 1 }, /* Thread UI */
   { 'G', "", storePosInteger, &SyNumGCThreads, 1 }, /* Thread UI */
 #endif
-  /* The following three options must be handled in the kernel so they happen early enough */
+  /* The following options must be handled in the kernel so they are setup before loading the library */
```
fingolfin:
```suggestion
  /* The following options must be handled in the kernel so they are set up before loading the library */
```

PR #3447, opened 2019-05-10
kernel: rename T_PRAGMA -> STAT_PRAGMA

fingolfin:
Also add comment to PrintPragma and change its argument from `Expr expr`
to `Stat stat`, as pragmas are statements.

I forgot to adjust #3422 with this before merging it, my bad.


sebasguts:
Looks good, thank you :). I should probably have seen this missing in #3422 and left a comment, I forgot.

PR #3448, opened 2019-05-14
lib: add AsSSortedList method ...

ssiccha:
... for collections that already are strictly sorted lists.

This ensures that `AsSet([1..n])` returns `[1..n]` instead of creating a
new plain list.

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -646,6 +646,12 @@ InstallOtherMethod(AsSSortedList,
      [ IsList ],
      l -> AsSSortedListList( AsPlist( l ) ) );
 
+InstallMethod( AsSSortedList,
+    "for a strictly sorted list",
+    [ IsSSortedList ],
+    SUM_FLAGS,
+    Immutable );
+
 
 #############################################################################
 ##
--- a/tst/testinstall/coll.tst
+++ b/tst/testinstall/coll.tst
@@ -338,6 +338,8 @@ gap> res:=List([AsList,AsSortedList,AsSet], f -> f(Magma(1)));
 [ [ 1 ], [ 1 ], [ 1 ] ]
 gap> List(res,IsMutable);
 [ false, false, false ]
+gap> AsSet([1..5]);
+[ 1 .. 5 ]
 
 #############################################################################
 #
--- a/tst/testinstall/list.tst
+++ b/tst/testinstall/list.tst
@@ -472,5 +472,26 @@ gap> PositionsBound( [ 1,, 2 ] );
 gap> HasIsSSortedList( PositionsBound( [ 1,, 2 ] ) );
 true
 
+# AsSet
+gap> l := ["a", "b", "c"];
+[ "a", "b", "c" ]
+gap> MakeImmutable(l);;
+gap> IsSet(l);
+true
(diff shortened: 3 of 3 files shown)


ChrisJefferson:
Stupid question -- why does this have to be "IsCollection and IsSSortedList", and not just "IsSSortedList"?

ssiccha:
Good question. Since there is a method installed for `IsCollection and IsConstantTimeAccessList` I was afraid that `RankFilter` may be to small then, but I didn't check. :) Actually we have
```
gap> RankFilter(IsSSortedList);
8
gap> RankFilter(IsCollection and IsConstantTimeAccessList);
6
```

I have updated the PR. I have added a "manual rank increment" such that the new functions should always be called when the argument already is a (immutable or mutable) `IsSSortedList`.

lib/list.gi
```diff
@@ -646,6 +646,18 @@ InstallOtherMethod(AsSSortedList,
      [ IsList ],
      l -> AsSSortedListList( AsPlist( l ) ) );
 
+InstallMethod( AsSSortedList,
+    "for a mutable strictly sorted list",
+    [ IsSSortedList and IsMutable ],
+    100
```
fingolfin: Please don't use static rank adjustments like that except as a very, very last resort.

In this particular case, it would probably be fine to just use `SUM_FLAGS`. Alternatively, one could use a dynamic  rank adjustment via a rank function, which has the added benefit that if done right, it is "self-documenting", i.e., makes it clear which other filter you are trying to beat. But I think `SUM_FLAGS` is fine here.

lib/list.gi
```diff
@@ -646,6 +646,18 @@ InstallOtherMethod(AsSSortedList,
...
+    Immutable );
+
+InstallMethod( AsSSortedList,
+    "for a strictly sorted list",
+    [ IsSSortedList ],
+    100
```
fingolfin: Ditto.

ssiccha:
I'll have a look into why tests fail.

lib/list.gi
ssiccha: Will do! Thanks.

ssiccha: Done. I also updated the commit message accordingly.

lib/list.gi
```diff
@@ -646,6 +646,18 @@ InstallOtherMethod(AsSSortedList,
      [ IsList ],
      l -> AsSSortedListList( AsPlist( l ) ) );
 
+InstallMethod( AsSSortedList,
+    "for a mutable strictly sorted list",
+    [ IsSSortedList and IsMutable ],
+    SUM_FLAGS,
+    Immutable );
+
+InstallMethod( AsSSortedList,
```
ChrisJefferson: I understand why we might have this optimisation (ID_FUNC), but Immutable is fast (and doesn't copy) objects which are already immutable, and I never like functions which require another function to be correct.

ChrisJefferson: Also, maybe chuck in some more tests (I'd be tempted to chuck in a sorted list of booleans, and a sorted string) as they are "non-standard" list types. Probably not required in this case, but I personally use them whenever I'm sorting lists.

ChrisJefferson:
Thanks

PR #3450, opened 2019-05-14
Some GASMAN related refactoring; allow multiple before/after GC callbacks

fingolfin:
This is needed to get SingularInterface to work again. That package previously copied the old values of `BeforeCollectFuncBags` and `AfterCollectFuncBags` and then installed its own callbacks, which called the original ones. But then we made these two variables static, which broke that hack. So the easiest fix would have been to export them again. But it seems better to keep them static, and instead provide a slightly more generic API. As a bonus, we can actually make use of that ourselves in a minor way.

Also cleanup some other stuff (mostly comments) I noticed while looking at `gasman.h` to implement this.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -984,21 +984,31 @@ void            InitFreeFuncBag (
 
 /****************************************************************************
 **
-*F  InitCollectFuncBags(<bfr-func>,<aft-func>) . install collection functions
-**
-**  'InitCollectFuncBags' is really too simple for an explanation.
 */
-static TNumCollectFuncBags BeforeCollectFuncBags;
+static struct {
+    UInt nrBefore;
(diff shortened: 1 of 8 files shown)


ChrisJefferson:
There are still some issues with this, the travis failures are "real", but in general the PR looks sensible.

fingolfin:
@ChrisJefferson yeah those were embarrassing omissions of mine. I am glad that you added testing for the memcheck code to Travis! Anyway, I *think* everything is fine now, but Travis will tell us if I am right on that :-)

fingolfin:
@ChrisJefferson now this is fixed for real.

fingolfin:
Typo fixed, Travis running :)

PR #3451, opened 2019-05-14
Performance improvements

### LINKED ISSUE #3464: Errors in `testextra/grpperm.tst` in master branch
See e.g. [this St Andrews-only link](https://gap-ci.cs.st-andrews.ac.uk/view/GAP-release/job/GAP-major-release-test/129/GAPCOPTS=64build,GAPTARGET=extra,label=kovacs/console):

```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-major-release-test/\
GAPCOPTS/64build/GAPTARGET/extra/label/kovacs/gap-4.11.0/tst/testextra/grpperm\
.tst:260
# Input is:
coh.cohomology;
# Expected output:
[ <an immutable GF2 vector of length 336>, <an immutable GF2 vector of length
    336> ]
# But found:
[ <an immutable GF2 vector of length 224>, <an immutable GF2 vector of length 
    224> ]
########
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-major-release-test/\
GAPCOPTS/64build/GAPTARGET/extra/label/kovacs/gap-4.11.0/tst/testextra/grpperm\
.tst:264
# Input is:
p:=List(e,x->FpGroupCocycle(coh,x,true));
# Expected output:
[ <fp group of size 61440 on the generators [ F1, F2, F3, F4, F5, F6, F7, F8,
      m1, m2, m3, m4 ]>, <fp group of size 61440 on the generators
    [ F1, F2, F3, F4, F5, F6, F7, F8, m1, m2, m3, m4 ]>,
  <fp group of size 61440 on the generators [ F1, F2, F3, F4, F5, F6, F7, F8,
      m1, m2, m3, m4 ]>, <fp group of size 61440 on the generators
    [ F1, F2, F3, F4, F5, F6, F7, F8, m1, m2, m3, m4 ]> ]
# But found:
Error, Assertion failure
########
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-major-release-test/\
GAPCOPTS/64build/GAPTARGET/extra/label/kovacs/gap-4.11.0/tst/testextra/grpperm\
.tst:271
# Input is:
p:=List(p,x->Image(IsomorphismPermGroup(x)));
# Expected output:
[ <permutation group of size 61440 with 12 generators>,
  <permutation group of size 61440 with 12 generators>,
  <permutation group of size 61440 with 12 generators>,
  <permutation group of size 61440 with 12 generators> ]
# But found:
Error, Variable: 'p' must have a value
########
```
In these three diffs we have:
1. Different dimensions
2. Assertion failure
3. Caused by failed (2)

hulpke:
# Description

Improvements that mostly relate to computing double cosets, in particular  in symmetric groups, and to permutation representations of extensions:

- Conjugation of subgroups into other subgroup uses permutation group orbits first
- coding improvements in double coset routine
- Try to use good presentation if group is Sn/An. Fix for wrong assumption in complement code this throwed up.
- RepresentativeAction method for OnTuplesSets
- Coding improvement in finding permrep. for extensions.
- Resolve #3464 

## Text for release notes 

Performance of double coset calculations, in particular in symmetric and alternating groups, has been improved.


lib/gpprmsya.gi
```diff
@@ -272,15 +272,39 @@ function ( G )
...
+local hom;
+  if IsPermGroup(a) then
+    hom:=SmallerDegreePermutationRepresentation(a);
+    if Image(hom)=b then return hom;
+    elif NrMovedPoints(Range(hom))<NrMovedPoints(a) then
+      return hom*IsomorphismGroups(Image(hom),b);
+    else
+      return IsomorphismGroups(a,b);
+    fi;
+  fi;
```
fingolfin: So if the first argument is not a perm group, this function suddenly returns nothing. That seems wrong. Perhaps it should return `fail`, or call `Error`?

lib/gpprmsya.gi
```diff
@@ -272,15 +272,39 @@ function ( G )
...
+end);
+
 InstallMethod( IsomorphismFpGroup, "alternating group", true,
-    [ IsNaturalAlternatingGroup ], 10, # override `IsSimpleGroup' method
+    [ IsAlternatingGroup ], 
+    40, # override `IsSimple...' and `IsNatural...' method
```
fingolfin: I don't understand, which `IsNatural...` method are we overriding here? Does this *replace* the `IsNaturalAlternatingGroup` method?

lib/gpprmsya.gi
```diff
@@ -272,15 +272,39 @@ function ( G )
...
+    35, # override `IsSimple...' and `IsNatural...' method
+function(G,str)
+local H;
+  H:=AlternatingGroup(AlternatingDegree(G));
+  return CheapIsomSymAlt(G,H)*IsomorphismFpGroup(H,str);
+end);
+
+InstallOtherMethod( IsomorphismFpGroup, "natural alternating group,name",
     true,
     [ IsNaturalAlternatingGroup, IsString ],
```
fingolfin: Why is this method kept, isn't the method above always invoked over it?

lib/csetgrp.gi
```diff
@@ -80,6 +80,29 @@ local c,i;
   fi;
 end );
 
+
```
fingolfin: Typo in one commit message: "permuattion"

lib/csetgrp.gi
```diff
@@ -988,9 +1052,13 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
 	fi;
       end;
 
-      for i in TryMaximalSubgroupClassReps(G:cheap) do
-	if Index(G,i)<maxidx(c) and Index(G,i)<=avoidlimit then
+      r:=Filtered(TryMaximalSubgroupClassReps(G:cheap),
+        x->Index(G,x)<=5*avoidlimit);
+      SortBy(r,a->-Size(a));
+      for i in r do
+	if Index(G,i)<maxidx(c) then
```
fingolfin: Something is off about the indentation (tabs vs. spaces perhaps?)

lib/csetgrp.gi
```diff
@@ -988,9 +1052,13 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
...
 	  p:=Intersection(a,i);
+          AddSet(doneidx,Index(a,p));
```
fingolfin: indentation

lib/csetgrp.gi
```diff
@@ -1045,7 +1113,12 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
   stabs:=[b];
   dcs:=[];
 
+  # Do we want to keep for smaller group (as cheaper fuse is possible
```
fingolfin: I don't understand this sentence, "Do we want to keep for smaller group" -- keep what?

lib/csetgrp.gi
```diff
@@ -1119,39 +1193,49 @@ local c, flip, maxidx, refineChainActionLimit, cano, tryfct, p, r, t,
...
       while bsz>0 and cnt>0 do
 	cnt:=cnt-1;
 
 	# compute orbit and stabilizers for the next step
         # own Orbitalgorithm and stabilizer computation
 
-	ps:=Position(blist,false);
+	#ps:=Position(blist,false);
+        while blist[posi] do posi:=posi+1;od;
+        ps:=posi;
```
fingolfin: Or you could use `ps:=Position(blist,false,posi-1);` -- this is shorter and uses a kernel function (probably doesn't matter here, though).
Also, I think that if you change `posi:=1` to `posi:=0` above, you could also use `ps:=Position(blist,false,posi);`

lib/gpprmsya.gi
hulpke: No, It still ranks higher. I welcome suggestions to make this cleaner.

lib/csetgrp.gi
hulpke: Thanks.

lib/gpprmsya.gi
hulpke: Changed,.

lib/csetgrp.gi
hulpke: changed.

lib/csetgrp.gi
hulpke: I've globally replaced tabs and now its OK.

lib/gpprmsya.gi
hulpke: No. This is an unfortunate side effect of a greedy `diff`.

There is a new method inserted for alternating groups that are not natural, and the old method having the comment string "alternating group" instead of "natural alternating group"

fingolfin: OK, but still, you want this method to be ranked higher than that for `IsNaturalAlternatingGroup`? That seems strange at first glance. Doesn't that mean that the `IsNaturalAlternatingGroup` is never going to be run anyway, given that `IsNaturalAlternatingGroup` implies `IsAlternatingGroup ` ?

lib/gpprmsya.gi
fingolfin: I don't understand why it ranks higher? I am seeing this:
```
gap> RankFilter(IsNaturalAlternatingGroup);
65
gap> RankFilter(IsAlternatingGroup);
37
gap> RankFilter(IsAlternatingGroup) + 40 > RankFilter(IsNaturalAlternatingGroup);
true
```

hulpke: I think it is implied filters (but I did not check that). I checked with `ApplicableMethod` and got (excerpt)

    #I  Method 2 : ``IsomorphismFpGroup: natural alternating group,name'' at /Users/hulpke/gap4/lib/gpprmsya.gi:306, value: 78
    #I   - 1st argument needs
    [ "IsNaturalAlternatingGroup", "HasIsNaturalAlternatingGroup" ]    
    #I  Method 3: ``IsomorphismFpGroup: alternating group,name'' at     /Users/hulpke/gap4/lib/gpprmsya.gi:296, value: 77
    #I   - 1st argument needs [ "IsAlternatingGroup", "HasIsAlternatingGroup" ]
    #I  Method 6: ``IsomorphismFpGroup: natural symmetric group,name'' at /Users/hulpke/gap4/lib/gpprmsya.gi:1787, value: 68
    #I   - 1st argument needs
    [ "IsNaturalSymmetricGroup", "HasIsNaturalSymmetricGroup" ]
    #I  Method 7: ``IsomorphismFpGroup: symmetric group,name'' at /Users/hulpke/gap4/lib/gpprmsya.gi:1777, value: 67
    #I   - 1st argument needs [ "IsSymmetricGroup", "HasIsSymmetricGroup" ]

So the ordering is OK. One could go down in the value (I think up to 9) without getting into problems with other methods.

lib/gpprmsya.gi
hulpke: No, it ranks lower, see next comment.

lib/gpprmsya.gi
fingolfin: The problem is that the difference is not stable -- it is 1 for you, but it is -12 on my system: the difference is heavily impacted by which packages are loaded.

One way to fix this would be to replace the static rank with a dynamic one, such as `{}->RankFilter(IsNaturalAlternatingGroup)-RankFilter(IsAlternatingGroup)+1`

hulpke: Ok -- I did not realize that packages could influence the library that way.

I've changed the code to just have a single method in either case, this should avoid all ambiguity and will also catch the hypothetical case of a group knowing it is alternating but not yet knowing its naturally alternating.

alex-konovalov:
REMINDER - to all PR authors and reviewers: if a PR adds tests to `tst/testextra`, do not assume that it's tested with Travis CI, for it is not. Please test it manually (this PR added a broken test to `tst/testextra`, which had been fixed in #3501).

PR #3453, opened 2019-05-15
Clarify documentation of GeneratorsOfDomain and add an example

ssiccha:
A domain comes with an associated `operational structure`. The documentation for GeneratorsOfDomain states:
> For  a  domain D, GeneratorsOfDomain returns a list containing all elements of D,
  perhaps  with repetitions.

I'm changing this to:
> For  a  domain D, GeneratorsOfDomain returns a list containing generators of D with respect to its operational structure.
> The returned list may contain repetitions.


fingolfin:
Of course we may now argue that this is not what it should be; but then the question is, *what* should it do? How do we decide which kind of "generators" are supposed to be returned? Those of a magma? A semigroup? A monoid? A group? A ring? A ring-with-one? etc. ... Or perhaps "The most specific one?" -- assuming that is always well-defined?

ssiccha:
Ah, I've interpreted the following section in the documentation as if there were a unique operational structure attached to a domain.
```
One important difference between the operational structure and the properties of
a  domain  is  that  the  operational  structure  is  fixed  when  the domain is
constructed,  whereas  properties  can  be discovered later.
```
But since domains are modelled via filters there can't be a unique "most specific with respect to inclusion" one.

In #3460 I suggest a category `IsSetAsDomain`. Would it make sense to add a documented synonym for `GeneratorsOfDomain` by the name of e.g. `GeneratorsOfSetAsDomain`?

Would it be problematic if installations of `GeneratorsOfDomain` assume another operational structure than the trivial one? Note that permutation groups don't have any method installed for `GeneratorsOfDomain`.

ssiccha:
~Note that permutation groups don't have any method installed for `GeneratorsOfDomain`.~ That's not true. Actually we have
```
gap> G := Group([(1,2), (3,4)]);
Group([ (1,2), (3,4) ])
gap> GeneratorsOfDomain(G);
[ (), (3,4), (1,2), (1,2)(3,4) ]
```

So `GeneratorsOfDomain` returns generators with respect to the trivial operational structure, that is it returns the domain as a set, as @fingolfin said. I will update my PR to clarify this in the documentation.

ssiccha:
Nevermind my previous suggestions. I just updated this PR and I think the new version is a lot clearer.

PR #3454, opened 2019-05-15
Improve error handling for `Image`, `Images`, `PreImage`, `PreImages`

ssiccha:
..., PreImage, and PreImages.

If one of these functions figures out that the second argument is not an
element or subset of the Range or the Source, it now tells the user.

### DIFF
--- a/hpcgap/lib/mapping.gi
+++ b/hpcgap/lib/mapping.gi
@@ -158,8 +158,12 @@ InstallGlobalFunction( Image, function ( arg )
     local   map,        # mapping <map>, first argument
             elm;        # element <elm>, second argument
 
+    if Length(arg) > 0 and not IsGeneralMapping( arg[1] ) then
+      ErrorNoReturn( "<map> must be a general mapping" );
+    fi;
+
     # image of the source under <map>, which may be multi valued in this case
-    if   Length( arg ) = 1 and IsGeneralMapping(arg[1]) then
+    if Length( arg ) = 1 then
 
         return ImagesSource( arg[1] );
 
@@ -168,17 +172,21 @@ InstallGlobalFunction( Image, function ( arg )
       map := arg[1];
       elm := arg[2];
 
+      # image of a single element <elm> under the mapping <map>
       if     FamSourceEqFamElm( FamilyObj( map ), FamilyObj( elm ) ) then
-
         if not IsMapping( map ) then
-          Error( "<map> must be a mapping" );
-        elif elm in Source( map ) then
-          return ImageElm( map, elm );
+          ErrorNoReturn( "<map> must be single-valued and total" );
+        elif not elm in Source( map ) then
+          ErrorNoReturn( "<elm> must be an element of Source(<map>)" );
         fi;
+        return ImageElm( map, elm );
 
-      # image of a set or list of elments <elm> under the mapping <map>
-      elif     CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) )
-           and IsSubset( Source( map ), elm ) then
+      # image of a collection of elments <elm> under the mapping <map>
+      elif CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) ) then
+        if not IsSubset( Source( map ), elm ) then
+          ErrorNoReturn( "the collection <elm> must be contained in ",
+                         "Source(<map>)" );
+        fi;
 
         if IsDomain( elm ) or IsSSortedList( elm ) then
           if HasSource(map) and IsIdenticalObj(Source(map),elm) then
@@ -195,9 +203,15 @@ InstallGlobalFunction( Image, function ( arg )
 
         return [];
 
+      else
+        ErrorNoReturn( "the families of the element or collection <elm> ",
+                       "and Source(<map>) don't match, ",
+                       "maybe <elm> is not contained in Source(<map>) or ",
+                       "is not a homogeneous list or collection" );
       fi;
     fi;
-    Error( "usage: Image(<map>), Image(<map>,<elm>), Image(<map>,<coll>)" );
+    ErrorNoReturn( "usage: Image(<map>), Image(<map>,<elm>), ",
+                   "Image(<map>,<coll>)" );
 end );
 
 
@@ -212,6 +226,10 @@ InstallGlobalFunction( Images, function ( arg )
     local   map,        # mapping <map>, first argument
             elm;        # element <elm>, second argument
 
+    if Length(arg) > 0 and not IsGeneralMapping( arg[1] ) then
+      ErrorNoReturn( "<map> must be a general mapping" );
+    fi;
+
     # image of the source under <map>
     if Length( arg ) = 1  then
 
@@ -222,19 +240,19 @@ InstallGlobalFunction( Images, function ( arg )
         map := arg[1];
         elm := arg[2];
 
-        if not IsGeneralMapping( map ) then
-          Error( "<map> must be a general mapping" );
-        fi;
-
         # image of a single element <elm> under the mapping <map>
-        if     FamSourceEqFamElm( FamilyObj( map ), FamilyObj( elm ) )
-           and elm in Source( map ) then
-
+        if     FamSourceEqFamElm( FamilyObj( map ), FamilyObj( elm ) ) then
+          if not elm in Source( map ) then
+            ErrorNoReturn( "<elm> must be an element of Source(<map>)" );
+          fi;
           return ImagesElm( map, elm );
 
-        # image of a set or list of elments <elm> under the mapping <map>
-        elif     CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) )
-             and IsSubset( Source( map ), elm ) then
+        # image of a collection of elments <elm> under the mapping <map>
+        elif CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) ) then
+          if not IsSubset( Source( map ), elm ) then
+            ErrorNoReturn( "the collection <elm> must be contained in ",
+                           "Source(<map>)" );
+          fi;
 
           if IsDomain( elm ) or IsSSortedList( elm ) then
             return ImagesSet( map, elm );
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
Could you write tests to hit all these new cases, just to sanity check them?

lib/mapping.gi
```diff
@@ -165,17 +165,22 @@ InstallGlobalFunction( Image, function ( arg )
       map := arg[1];
       elm := arg[2];
 
-      if     FamSourceEqFamElm( FamilyObj( map ), FamilyObj( elm ) ) then
+      if not IsMapping( map ) then
+        ErrorNoReturn( "<map> must be a mapping" );
```
fingolfin: But why? Before this, one could have done `Image(generalMap, set)` to return the set of (generalized) images for the generalized map `generalMap`. I think that was intentionally supported.

ssiccha:
I'll definitely set up some tests as @ChrisJefferson suggested and have a closer look at your inline comment @fingolfin.

lib/mapping.gi
ssiccha: I think you're right. I updated the PR. The next thing to do now is to write some tests.

ssiccha:
I'm updating the PR right now. I have added tests and deleted the
```
      else
          ErrorNoReturn( "<coll> must be a domain, a set, ",
                         "or a homogeneous list" );
```
line from Image. This "else" branch can't be executed since the function compares the families of the source and of the argument `elm`. The source must be a domain and thus will always have a different family in the situation that `elm` is not a homogeneous list.

ssiccha:
PR is updated.

ssiccha:
And again some HPCGAP stuff.

hpcgap/lib/mapping.gi
```diff
@@ -195,9 +202,15 @@ InstallGlobalFunction( Image, function ( arg )
 
         return [];
 
+      else
+        ErrorNoReturn( "couldn't match families of <elm> or <coll> and ",
+                       "Source(<map>),\n",
+                       "maybe <elm> or <coll> are not contained in ",
+                       "Source(<map>) or\naren't homogeneous lists," );
```
fingolfin: Why is that `\n` in the middle of a line, not at the end? Perhaps format it like so:
```gap
        ErrorNoReturn( "couldn't match families of <elm> or <coll> and Source(<map>),\n",
                       "maybe <elm> or <coll> are not contained in Source(<map>) or\n",
                       "aren't homogeneous lists," );
```
Or if you are concerned about line width:
```gap
        ErrorNoReturn(
            "couldn't match families of <elm> or <coll> and Source(<map>),\n",
            "maybe <elm> or <coll> are not contained in Source(<map>) or\n",
            "aren't homogeneous lists," );
```

That said, I am not even sure it's such a good idea to insert manual line breaks... ???

hpcgap/lib/mapping.gi
```diff
@@ -195,9 +202,15 @@ InstallGlobalFunction( Image, function ( arg )
 
         return [];
 
+      else
+        ErrorNoReturn( "couldn't match families of <elm> or <coll> and ",
```
fingolfin: Why "couldn't match families" and not "families of ... don't match"?

hpcgap/lib/mapping.gi
```diff
@@ -247,9 +263,15 @@ InstallGlobalFunction( Images, function ( arg )
 
           return [];
 
+        else
+          ErrorNoReturn( "couldn't match families of <elm> or <coll> and ",
+                         "Source(<map>),\n",
+                         "maybe <elm> or <coll> are not contained in ",
+                         "Source(<map>) or\naren't homogeneous lists," );
```
fingolfin: Same remarks as above.

hpcgap/lib/mapping.gi
```diff
@@ -298,10 +326,15 @@ InstallGlobalFunction( PreImage, function ( arg )
 
           return [];
 
+        else
+          ErrorNoReturn( "couldn't match families of <elm> or <coll> and ",
+                         "Range(<map>),\n",
+                         "maybe <elm> or <coll> are not contained in ",
+                         "Source(<map>) or\naren't homogeneous lists," );
```
fingolfin: Same as above.

hpcgap/lib/mapping.gi
ssiccha: Yep, you're right. I'll drop the manual line breaks. It feels a bit weird.

hpcgap/lib/mapping.gi
ssiccha: I'll use "families of .. don't match". Sounds better. :+1:

ssiccha:
I've addressed your remarks.

hpcgap/lib/mapping.gi
```diff
@@ -168,17 +172,20 @@ InstallGlobalFunction( Image, function ( arg )
...
+          ErrorNoReturn( "<elm> must be an element of Source(<map>)" );
         fi;
+        return ImageElm( map, elm );
 
       # image of a set or list of elments <elm> under the mapping <map>
-      elif     CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) )
-           and IsSubset( Source( map ), elm ) then
+      elif CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) ) then
+        if not IsSubset( Source( map ), elm ) then
+          ErrorNoReturn( "<coll> must be a subset of Source(<map>)" );
```
wilfwilson: Why is this `<coll>`? I expect, in this context the error message to be telling me the name of a variable here. But the variable here is called `<elm>`. So I think either you should say something like `The collection <elm>...`, or in this case (that `elm` is actually a collection), you could define a new variable `coll := elm` so that you can actually refer to `coll` in a proper way.

hpcgap/lib/mapping.gi
```diff
@@ -195,9 +202,15 @@ InstallGlobalFunction( Image, function ( arg )
...
+      else
+        ErrorNoReturn( "families of <elm> or <coll> and Source(<map>) ",
+                       "don't match, ",
+                       "maybe <elm> or <coll> are not contained in ",
+                       "Source(<map>) or aren't homogeneous lists," );
       fi;
     fi;
-    Error( "usage: Image(<map>), Image(<map>,<elm>), Image(<map>,<coll>)" );
+    ErrorNoReturn( "usage: Image(<map>), Image(<map>,<elm>), ",
+                   "Image(<map>,<coll>)" );
```
wilfwilson: However, in my opinion, it is fine to use `<coll>` here. Here you're only giving a name for a potential argument, and in doing so suggesting that it could be a collection. You're not implying that a variable called `coll` already exists.

hpcgap/lib/mapping.gi
```diff
@@ -195,9 +202,15 @@ InstallGlobalFunction( Image, function ( arg )
 
         return [];
 
+      else
+        ErrorNoReturn( "families of <elm> or <coll> and Source(<map>) ",
```
wilfwilson: Again, in this error message, the variable `<coll>` that it's referring to doesn't exist, so I think you should use words instead. Something like `...the families of the element or collection <elm> and Source(<map>)...`

hpcgap/lib/mapping.gi
```diff
@@ -222,19 +239,18 @@ InstallGlobalFunction( Images, function ( arg )
...
+            ErrorNoReturn( "<elm> must be an element of Source(<map>)" );
+          fi;
           return ImagesElm( map, elm );
 
         # image of a set or list of elments <elm> under the mapping <map>
-        elif     CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) )
-             and IsSubset( Source( map ), elm ) then
+        elif CollFamSourceEqFamElms( FamilyObj( map ), FamilyObj(elm) ) then
+          if not IsSubset( Source( map ), elm ) then
+            ErrorNoReturn( "<coll> must be a subset of Source(<map>)" );
```
wilfwilson: Same comments as above, I won't repeat the text but my comments will apply to all the other places where you have written `coll` in the ways that I have complained about.

hpcgap/lib/mapping.gi
```diff
@@ -276,16 +302,18 @@ InstallGlobalFunction( PreImage, function ( arg )
 
         # preimage of a single element <img> under <map>
         if     FamRangeEqFamElm( FamilyObj( map ), FamilyObj( img ) ) then
-          if not (     IsGeneralMapping( map ) and IsInjective( map )
-                   and IsSurjective( map ) ) then
-            Error( "<map> must be an inj. and surj. mapping" );
-          elif img in Range( map ) then
-            return PreImageElm( map, img );
+          if not ( IsInjective( map ) and IsSurjective( map ) ) then
+            ErrorNoReturn( "<map> must be an inj. and surj. mapping" );
```
wilfwilson: I know this isn't your message, but I recommend removing the abbreviations, to make things as clear as possible.
```suggestion
            ErrorNoReturn( "<map> must be an injective and surjective mapping" );
```

fingolfin: Actually, why not just say "bijective" (and use `IsBijective`)?

wilfwilson: As well as `IsInjective` and `IsSurjective`, `IsBijective` I implies `IsTotal` and `IsSingleValued`. I don’t know, but perhaps there’s a reason we don’t want to require one or the other of those things here

fingolfin: Ugh, yeah, a map from a non-trivial group to a trivial group which is only defined at `1` is an injective and surjective mapping, even single valued, but it is not total. (That's -1 points for GAP for making a very surprising change in the meaning of very well established mathematical terms, with surprising consequences *sigh*).

hpcgap/lib/mapping.gi
ssiccha: I'm rewriting this to `<elm>`. I'll also change the wording from `a subset of Source..` to `contained in Source..`. Talking about "subsets" feels like it implies that `elm` and `Source..` are sets.

ssiccha: Done.

hpcgap/lib/mapping.gi
ssiccha: Done.

ssiccha:
I adopted the suggested improvements.

I put the changes into a separate commit so you can see more easily what I've changed. I'll turn the new commits into a single one after the PR is accepted.

hpcgap/lib/mapping.gi
ssiccha: Yupp. I'll keep it that way.

ssiccha:
.. and I rebased onto master.

ssiccha:
Squashed.

PR #3455, opened 2019-05-15
Add and document new function `DirectProductFamily`

ssiccha:
lib: add operation DirectProductFamily

This declares the new operation DirectProductFamily and installs a
method for it. Also adds documentation and tests.

In the reference manual, the section "IsDirectProductElement (Filter)"
is renamed into "Direct Products and their Elements" to also accomodate
the documentation of DirectProductFamily and future categories and
operations.

UnderlyingRelation is adapted to use the new method, which makes it
easier to read.

### DIFF
--- a/lib/mapping.gi
+++ b/lib/mapping.gi
@@ -1259,12 +1259,11 @@ InstallMethod( UnderlyingRelation,
     true,
     [ IsGeneralMapping ], 0,
     function( map )
-    local rel;
-    rel:= Objectify( NewType( CollectionsFamily(
-          DirectProductElementsFamily( [ ElementsFamily( FamilyObj( Source( map ) ) ),
-                          ElementsFamily( FamilyObj( Range( map  ) ) ) ] ) ),
-                              IsDomain and IsAttributeStoringRep ),
-                     rec() );
+    local type, rel;
+    type:= NewType( DirectProductFamily( [ FamilyObj( Source( map ) ),
+                                           FamilyObj( Range( map ) ) ] ),
+                    IsDomain and IsAttributeStoringRep );
+    rel:= Objectify( type, rec() );
     SetUnderlyingGeneralMapping( rel, map );
     return rel;
     end );
--- a/lib/tuples.gd
+++ b/lib/tuples.gd
@@ -178,3 +178,52 @@ direct product elements families" );
 DeclareOperation( "DirectProductElement", [ IsList ]);
 DeclareOperation( "DirectProductElementNC",
     [ IsDirectProductElementFamily, IsList ]);
+
+
+#############################################################################
+##
+##
+##  <#GAPDoc Label="DirectProductFamily">
+##  <ManSection>
+##  <Func Name="DirectProductFamily" Arg='args'/>
+##
+##  <Description>
+##  <A>args</A> must be a dense list of <Ref Attr="CollectionsFamily"/>
+##  families, otherwise the function raises an error.
+##  <P/>
+##  <Ref Func="DirectProductFamily"/> returns <C>fam</C>, a collections
+##  family of <Ref Filt="IsDirectProductElement"/> objects.
+##  <P/>
+##  <C>fam</C> is the <Ref Filt="CollectionsFamily"/> of
+##  <Ref Filt="IsDirectProductElement"/> objects
+##  whose <C>i</C>-th component is in <C>ElementsFamily(args[i])</C>.
+##  <P/>
+##  Note that a collection in <C>fam</C> may not itself be a
+##  direct product; it just is a subcollection of a direct product.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> D8 := DihedralGroup(IsPermGroup, 8);;
+##  gap> FamilyObj(D8) = CollectionsFamily(PermutationsFamily);
+##  true
+##  gap> fam := DirectProductFamily([FamilyObj(D8), FamilyObj(D8)]);;
+##  gap> ComponentsOfDirectProductElementsFamily(ElementsFamily(fam));
+##  [ <Family: "PermutationsFamily">, <Family: "PermutationsFamily"> ]
+##  ]]></Example>
+##  Also note that not all direct products in &GAP; are created via these
+##  families. For example if the arguments to <Ref Func="DirectProduct"/>
+##  are permutation groups, then it returns a permutation group as well, whose
+##  elements are not <Ref Filt="IsDirectProductElement"/> objects.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> fam = FamilyObj(DirectProduct(D8, D8));
+##  false
+##  gap> D4 := DihedralGroup(IsPcGroup, 4);;
+##  gap> fam2 := DirectProductFamily([FamilyObj(D8), FamilyObj(D4)]);;
+##  gap> fam2 = FamilyObj(DirectProduct(D8, D4));
+##  true
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+DeclareGlobalFunction( "DirectProductFamily",
+                       "for a dense list of collection families" );
--- a/lib/tuples.gi
+++ b/lib/tuples.gi
@@ -506,3 +506,17 @@ InstallOtherMethod( \*,
     fi;
     return DirectProductElement( List( dpelm, entry -> nonlist * entry ) );
     end );
+
+
+#############################################################################
+##
+##
+InstallGlobalFunction( DirectProductFamily,
+    function( args )
+    if not IsDenseList(args) or not ForAll(args, IsCollectionFamily) then
+        ErrorNoReturn("<args> must be a dense list of collection families");
+    fi;
+    return CollectionsFamily(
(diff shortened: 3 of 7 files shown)


ssiccha:
I have added the `release notes not needed tag` since this function is not documented. Or should I document it?

DominikBernhardt:
Several tests failed due to a `No method found` error. You should check that.

ssiccha:
Thanks for the heads-up.

I fixed the issue. The filter `IsCollection` was (in hindsight obviously) too restrictive.

ChrisJefferson:
Do you want to be able to call this in your own code? If so, we should document it, so we know what it is supposed to do.

lib/tuples.gi
```diff
@@ -506,3 +506,19 @@ InstallOtherMethod( \*,
...
+##
+##
+InstallMethod( DirectProductFamily,
+    "for a collection (of families)",
+    fam -> fam = CollectionsFamily(FamilyOfFamilies),
+    [ IsDenseList ],
+    function( args )
+    return CollectionsFamily(
+        DirectProductElementsFamily(
+            List( args, x -> ElementsFamily( x ) )
```
fingolfin:
```suggestion
            List( args, ElementsFamily )
```

DominikBernhardt:
Tests for `DirectProductFamily` checking your added code would be nice.

lib/tuples.gi
```diff
@@ -506,3 +506,19 @@ InstallOtherMethod( \*,
     fi;
     return DirectProductElement( List( dpelm, entry -> nonlist * entry ) );
     end );
+
+
+#############################################################################
+##
+##
```
fingolfin: Do you expect a need for additional methods? Otherwise, this could be just a global function...

stevelinton:
In documenting this, (and `DirectProductElement`) it is probably worth pointing out that not *all* direct products are made this way. For instance:
```
gap> g := SymmetricGroup(4);
Sym( [ 1 .. 4 ] )
gap> DirectProduct(g,g);    
Group([ (1,2,3,4), (1,2), (5,6,7,8), (5,6) ])
```

Arguably, perhaps they should be, with `IsomorphismPermGroup` available to those who want a nicer form, but that's a question for another time.

ssiccha:
Good point! Thanks

ssiccha:
I was thinking about doing a function `DirectProductGroupInProductAction`. But that's low priority for me atm.

lib/tuples.gi
ssiccha: :+1:

lib/tuples.gi
ssiccha: I've converted it into a global function.

ssiccha:
I've updated the PR!

ssiccha:
Added the method to `hpcgap/lib/tuples.gi`

DominikBernhardt:
Looks good to me.

lib/tuples.gd
```diff
@@ -178,3 +178,46 @@ direct product elements families" );
...
+##  <ManSection>
+##  <Func Name="DirectProductFamily" Arg='args'/>
+##
+##  <Description>
+##  <A>args</A> must be a dense list of <Ref Attr="CollectionsFamily"/>
+##  families, otherwise the function raises an error.
+##  <P/>
+##  <Ref Func="DirectProductFamily"/> returns a collections family <C>fam</C>
+##  with the following property:
+##  Each collection <C>coll</C> in <C>fam</C> is a direct product
```
fingolfin: Wait, is that true: is each such collection a direct product? Isn't it just a subset of a direct product in general?

ssiccha: Yes, that's true. I realize now that my name for this family is misleading. A better name that comes to my mind is `DirectProductElementCollectionsFamily`. In #3477 I introduce a `IsDirectProductDomain`category, where the objects in this family actually are direct products.

ssiccha: I corrected and rewrote the documentation. I hope this is also clearer than the first version.

lib/tuples.gd
```diff
@@ -178,3 +178,43 @@ direct product elements families" );
...
+##  objects.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> D8 := DihedralGroup(IsPermGroup, 8);;
+##  gap> FamilyObj(D8) = CollectionsFamily(PermutationsFamily);
+##  true
+##  gap> fam := DirectProductFamily([FamilyObj(D8), FamilyObj(D8)]);;
+##  gap> ComponentsOfDirectProductElementsFamily(ElementsFamily(fam));
+##  [ <Family: "PermutationsFamily">, <Family: "PermutationsFamily"> ]
+##  ]]></Example>
```
wilfwilson: There's nothing wrong with this example, but as stated elsewhere the direct product of permutation groups is again a permutation group, in the permutations family. In my opinion, it would be nice to _also_ have an example where the family of the resulting direct product is the `DirectProductFamily`, eg. something along the lines of:
```
gap> D8 := DihedralGroup(IsPermGroup, 8);;
gap> D4 := DihedralGroup(IsPcGroup, 4);;
gap> fam := DirectProductFamily([FamilyObj(D8), FamilyObj(D4)]);;
gap> fam = FamilyObj(DirectProduct(D8, D4));
true
```
Is that a reasonable request? Do you think it makes thing less confusing, or more confusing?

lib/tuples.gd
```diff
@@ -178,3 +178,43 @@ direct product elements families" );
...
+##  families, otherwise the function raises an error.
+##  <P/>
+##  <Ref Func="DirectProductFamily"/> returns <C>fam</C>, a collections
+##  family of <Ref Filt="IsDirectProductElement"/> objects.
+##  <P/>
+##  <C>fam</C> is the <Ref Filt="CollectionsFamily"/> of
+##  <Ref Filt="IsDirectProductElement"/> objects
+##  whose <C>i</C>-th component is in <C>ElementsFamily(args[i])</C>.
+##  <P/>
+##  Note that a collection in <C>fam</C> must thus not necessarily be a
```
fingolfin:
```suggestion
##  Note that a collection in <C>fam</C> may not itself be a
```

lib/tuples.gd
```diff
@@ -178,3 +178,43 @@ direct product elements families" );
...
+##  direct product.
```
fingolfin:
```suggestion
##  direct product; it just is a subcollection of a direct product.
```

lib/tuples.gd
```diff
@@ -178,3 +178,43 @@ direct product elements families" );
...
+##  Also note that not all direct products in &GAP; are created via these
+##  families. For example the elements of a permutation group created via
```
fingolfin: This requires the reader to know how to create permutation group "via DirectProduct", resp. to at least know this is possible. So perhaps like this:
```suggestion
##  families. For example if the arguments to <Ref Func="DirectProduct"/> are permutation groups,
```
cont'd below

lib/tuples.gd
```diff
@@ -178,3 +178,43 @@ direct product elements families" );
...
+##  <Ref Func="DirectProduct"/> are not <Ref Filt="IsDirectProductElement"/>
```
fingolfin: cont'd from above
```suggestion
##  then it returns a permutation group as well, whose elements are not <Ref Filt="IsDirectProductElement"/>
```

lib/tuples.gd
fingolfin: I think that makes a lot of sense.

lib/tuples.gd
ssiccha: Oh, I accidentally used the german version of the meaning of `must`. Thanks.

lib/tuples.gd
ssiccha: Ooh, nice. I wasn't aware that you could do that. I'll add that.

I'll put the explanation about the direct product of permutation groups being a permutation group and your example into a new little paragraph. I hope that will make the documentation clearer.

ssiccha:
I rebased onto master and put the new changes into a separate commit. Once this PR is accepted I'll squash everything into a single commit.

ssiccha:
Squashed.

lib/mapping.gi
```diff
@@ -1259,12 +1259,11 @@ InstallMethod( UnderlyingRelation,
...
-    rel:= Objectify( NewType( CollectionsFamily(
-          DirectProductElementsFamily( [ ElementsFamily( FamilyObj( Source( map ) ) ),
-                          ElementsFamily( FamilyObj( Range( map  ) ) ) ] ) ),
-                              IsDomain and IsAttributeStoringRep ),
-                     rec() );
+    local type, rel;
+    type:= NewType( DirectProductFamily( [ FamilyObj( Source( map ) ),
+                                           FamilyObj( Range( map ) ) ] ),
+                    IsDomain and IsAttributeStoringRep );
+    rel:= Objectify( type, rec() );
```
fingolfin: I am afraid there also is `hpcgap/lib/mapping.gi` which needs this change, too.

ssiccha: Huh, I thought I did that. I'm gonna check it tomorrow. Thanks!

ssiccha: Done.

PR #3459, opened 2019-05-17
Add ShowDeclarationsOfOperation helper 

fingolfin:
When debugging issues with method installations for operations (including
attributes and properties), it can be helpful to find out what all
declarations of a given operation are. This helper function does just that.

The placement of this function in the manual could perhaps be better, I am open to suggestions, but generally feel that this could and should be improved in that whole chapter

This is a new feature and hence should be listed in the release notes.

### DIFF
--- a/lib/methwhy.g
+++ b/lib/methwhy.g
@@ -388,6 +388,52 @@ BIND_GLOBAL("ShowImpliedFilters",function(filter)
   fi;
 end);
 
+
+#############################################################################
+##
+#F  ShowDeclarationsOfOperation( <oper> )
+##
+##  <#GAPDoc Label="ShowDeclarationsOfOperation">
+##  <ManSection>
+##  <Func Name="ShowDeclarationsOfOperation" Arg='oper'/>
+##
+##  <Description>
+##  Displays information about all declarations of the operation <A>oper</A>,
+##  including the location of each declaration and the argument filters.
+##  <Log><![CDATA[
+##  gap> ShowDeclarationsOfOperation(IsFinite);
+##  Available declarations for operation <Property "IsFinite">:
+##    1: GAPROOT/lib/coll.gd:1451 with 1 arguments, and filters [ IsListOrCollection ]
+##    2: GAPROOT/lib/float.gd:212 with 1 arguments, and filters [ IsFloat ]
+##    3: GAPROOT/lib/ctbl.gd:1195 with 1 arguments, and filters [ IsNearlyCharacterTable ]
+##  ]]></Log>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+BIND_GLOBAL("ShowDeclarationsOfOperation",function(oper)
+    local locs, reqs, i, r;
+    if not IsOperation(oper) then
+        Error("<oper> must be an operation");
+    fi;
+    Print("Available declarations for operation ", oper, ":\n");
+    locs := GET_DECLARATION_LOCATIONS(oper);
+    if locs = fail then
+        return;
+    fi;
+    reqs := GET_OPER_FLAGS(oper);
+    for i in [1.. Length(locs)] do
+        r := List(reqs[i], r -> JoinStringsWithSeparator(NamesFilter(r), " and \c"));
+        Print(String(i, 3), ": ", locs[i][1], "\c:", locs[i][2], "\c",
+              " with ", Length(reqs[i]), "\c",
+              " arguments, and filters [ ", "\c",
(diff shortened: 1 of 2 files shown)


DominikBernhardt:
Thanks, this is indeed a very helpful function!
I don't have anything against the placement in the manual, although one could also think about putting it closer to the explanation of operations, i.e. `5.5-2` after `IsOperation`? It's not ideal either, but maybe a user would rather search there?

DominikBernhardt:
I agree with @markusbaumeister that the header `Function Types` itself does not feel perfect. But we have already established that there is no perfect _chapter_ name for this function, so I would rather put it near to `IsOperation` and imho `5.5-2` fits this. Ultimately, this is a matter of taste I guess.

fingolfin:
Yeah, I also find the section title "Global Variables in the Library" super confusing and rather inappropriate. And there are indeed some key concepts related to these operations/filters/categories/ etc. which don't seem to be explained in that chapter (or anywhere in the reference manual that I could find now).

I'll think some more about it...

lib/methwhy.g
```diff
@@ -388,6 +388,49 @@ BIND_GLOBAL("ShowImpliedFilters",function(filter)
...
+    Print("Available declarations for operation ", oper, ":\n");
+    locs := GET_DECLARATION_LOCATIONS(oper);
+    reqs := GET_OPER_FLAGS(oper);
+    for i in [1.. Length(locs)] do
+       r := List(reqs[i], r -> JoinStringsWithSeparator(NamesFilter(r), " and "));
+       Print(String(i, 3), ": ", locs[i][1], ":", locs[i][2],
+             " with ", Length(reqs[i]),
+             " arguments, and filters [ ",
+             JoinStringsWithSeparator(r, ", "),
+             " ]\n"
```
PaulaHaehndel: Because of the way the strings are joined together, when a line in the output is too long it breaks in an unintuitive way that is hard to read.
Therefore I suggest first build a complete string and then print it (or something similar).
An example where this happens is by invoking
`ShowDeclarationsOfOperation(IsPrimitive);`
The result is (for example depending on the width of the terminal):
```Available declarations for operation <Property "IsPrimitive">:
  1: /mnt/c/gap/lib/oprt.gd:1879 with 1 arguments, and filters [ IsExternalSet ]
  2: /mnt/c/gap/lib/oprt.gd:1879 with
5
  arguments, and filters [ IsMagmaWithInverses and IsAssociative and HasIsAssociative, IsListOrCollect\
ion, IsList, IsList, IsFunction ]```

For instance something like the following might work:
```Print(JoinStringsWithSeparator([String(i, 3), ": ", locs[i][1], ":", locs[i][2],
             " with ", Length(reqs[i]),
             " arguments, and filters [ ",
             JoinStringsWithSeparator(r, ", "),
             " ]\n"],"")
             );```

**Comment by PaulaHaehndel** [comment]
I also wouldn't expect it in the current place and more likely would search for it in `7 Debugging and Profiling Facilities`.

**Comment by fingolfin** [comment]
@PaulaHaehndel yes, it breaks if you give it `IsOperation`. Just like it breaks when you give it a string, the integer 42, or anything else that's not an operation. So, what are you saying -- is that an implicit request for argument validatioj?

File: lib/methwhy.g (hunk 1 above)
Comment (fingolfin) [reply lib/methwhy.g]: I've tweaked the output now by adding some `\c` into the string being printed, so now one gets this in your example:
```
gap> SizeScreen([80,100]);
[ 80, 100 ]
gap> ShowDeclarationsOfOperation(IsPrimitive);
Available declarations for operation <Property "IsPrimitive">:
  1: /Users/mhorn/Projekte/GAP/gap.spielwiese/lib/oprt.gd:1831 with 1 arguments, and filters [ IsExternalSet ]
  2: /Users/mhorn/Projekte/GAP/gap.spielwiese/lib/oprt.gd:1831 with 5 arguments, and filters [ IsMagmaWithInverses and IsAssociative and HasIsAssociative, IsListOrCollection\
, IsList, IsList, IsFunction ]]
```

fingolfin: Anyway, this is a debugging tool; while I am willing to make trivial improvements, there is only so far I'd want to with polishing it, as the returns are quickly diminishing.

fingolfin:
@PaulaHaehndel regarding the error with `IsOperation` as input: I stand corrected, `IsOperation` *is* an operation, I badly mixed this up with something else. My apologies! I've now added a check for that.

fingolfin:
I'll merge this now, and then provide another PR to improve the arrangement of things a bit. Thank you all for your feedback!

PR #3460, opened 2019-05-18
WIP lib: add SetAsDomain

ssiccha:
TODO
- move new code into domain.g?
- add tests!!
- text for release notes

# Description
Currently there is no filter that specifies explicitly that a domain is created as a _set with no operational structure_. This PR introduces the filter `IsSetAsDomain`. The motivation is to be able to use e.g. `SetAsDomain([1..n])` as a domain for `MappingByFunction`:
```
gap> MappingByFunction(SetAsDomain([1..5]), SetAsDomain([1..5]), x -> x);
MappingByFunction( SetAsDomain([ 1 .. 5 ]), SetAsDomain([ 1 .. 5 ]), function( x ) ... end )
gap> Image(last);
[ 1, 2, 3, 4, 5 ]
```

One thing that is not entirely clear to me yet: As `IsPermGroup` already implies `IsDomain`, should it also imply `IsSetAsDomain` or not? I think it should. Methods for e.g. `IsPermGroup` would then outrank methods for `IsSetAsDomain` anyways.

### DIFF
--- a/lib/set.gi
+++ b/lib/set.gi
@@ -222,3 +222,50 @@ InstallMethod( SubtractSet,
       RemoveSet( set, obj );
     od;
     end );
+
+
+#############################################################################
+##
+InstallMethod(SetAsDomain,
+"for a set & collection",
+[IsSet and IsCollection],
+function(set)
+    local copy, type, domain;
+    copy := Immutable(set);
+    type := NewType(FamilyObj(copy),
+                    IsSetAsDomain and IsAttributeStoringRep);
+    return ObjectifyWithAttributes(rec(), type,
+                                   GeneratorsOfDomain, copy,
+                                   AsList, copy,
+                                   AsSet, copy);
+end);
+
+# Provide family of objects `fam`. Then the resulting domain lives in the same
+# family as a domain created from a non-empty set of objects of `fam`.
+InstallOtherMethod(SetAsDomain,
+"for a family and an empty list",
+[IsFamily, IsEmpty and IsList],
+function(family, set)
+    local copy, type;
+    copy := Immutable([]);
+    type := NewType(CollectionsFamily(family),
+                    IsSetAsDomain and IsAttributeStoringRep);
+    return ObjectifyWithAttributes(rec(), type,
+                                   GeneratorsOfDomain, copy,
+                                   AsList, copy,
+                                   AsSet, copy);
+end);
+
+InstallMethod(\in,
+"for an IsSetAsDomain",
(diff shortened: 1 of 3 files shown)


lib/set.gi
```diff
@@ -222,3 +222,50 @@ InstallMethod( SubtractSet,
...
+                    IsSetAsDomain and IsAttributeStoringRep);
+    return ObjectifyWithAttributes(rec(), type,
+                                   GeneratorsOfDomain, copy,
+                                   AsList, copy,
+                                   AsSet, copy);
+end);
+
+InstallMethod(\in,
+"for an IsSetAsDomain",
+[IsObject, IsSetAsDomain],
```
fingolfin: But `IsSetAsDomain` is merely a synonym for `IsDomain`; so now you really have installed a generic `\in` method for arbitrary domains, which tries to use `AsList`. That makes me a bit nervous, though I can't put my finger on it right now.

lib/set.gi
```diff
@@ -222,3 +222,50 @@ InstallMethod( SubtractSet,
...
+function(elm, set)
+    return elm in AsList(set);
+end);
+
+InstallMethod(PrintObj,
+"for an IsSetAsDomain",
+[IsSetAsDomain],
+function( setAsDomain )
+    Print("SetAsDomain(", AsList(setAsDomain), ")");
```
fingolfin: Don't you need a different function for empty domains?

ssiccha:
Oh sorry. This was a bit chaotic since I also had like 5 other PRs going on. I first thought I would introduce a new Category `IsSetAsDomain`. Then I decided against that and unfortunately left the PR in a broken state which doesn't make any sense.

Here are my current thoughts on what I actually want to do:
For constructing the normalisers of socles of primitive groups I want to use the `IsMapping` objects in GAP. If I define a mapping of "sets" in the current master though I get
```
gap> MappingByFunction(Domain([1..5]), Domain([1..5]), x -> x);
MappingByFunction( <object>, <object>, function( x ) ... end )
```
which I find very ugly. At the very least I want users to be able to see for a map that I defined what its domain and range are. So I think there are basically two options
1. Install a `PrintObj` method for `IsDomain and HasGeneratorsOfDomain`
1. Create a new category `IsSetAsDomain` which models that our domain is an actual set. After all a collection of families could also be turned into a domain and this is neither a set in the mathematical sense nor in the `IsSSortedList` sense.

While I'm not sure whether there are people who would find `2.` useful I'm gonna go with `1.` atm since that's also a lot less effort.

PR #3462, opened 2019-05-18
Add two temporary travis tests, to try to track down bug

ChrisJefferson:
I am making this a PR in the hope it will fail, and give us more evidence about what is causing an occasional crash we are seeing.

If we can't find the crash soon, I suggest merging this, to see if it helps us narrow down the crash (once found, it could be removed again)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,6 +125,22 @@ matrix:
           - g++-4.7
           - valgrind
 
+    # The following two configurations are temporary, to try to track down
+    # a GAP crash in the previous test.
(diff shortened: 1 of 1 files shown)


fingolfin:
Unfortunately this seems to have passed. But of course you could try to just re-run the tests a couple times (by restarting them in the Travis web UI) to see if they eventually fail.

fingolfin:
Closing this: @ChrisJefferson seems to have found the issue (and in any case, we can always reopen this).

PR #3463, opened 2019-05-18
Fix buffer overflow in TransformationListListNC

ChrisJefferson:
Fix a buffer overflow in TransformationListList. This bug is in 4.10 (might be in earlier releases too)

## Text for release notes 

* Fix bug in TransformationListList which could cause a crash.

### DIFF
--- a/src/trans.cc
+++ b/src/trans.cc
@@ -476,8 +476,12 @@ static Obj FuncTransformationListListNC(Obj self, Obj src, Obj ran)
             ptf2[i] = i;
         }
         for (i = LEN_LIST(src); 1 <= i; i--) {
-            ptf2[INT_INTOBJ(ELM_LIST(src, i)) - 1] =
-                INT_INTOBJ(ELM_LIST(ran, i)) - 1;
+            s = INT_INTOBJ(ELM_LIST(src, i));
+            r = INT_INTOBJ(ELM_LIST(ran, i));
+            // deg may be smaller than s if s = r
+            if (s != r) {
+                ptf2[s - 1] = r - 1;
+            }
         }
     }
     else {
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
@james-d-mitchell : I realise this is a busy time of year, just wanted to link you to this in case you had any opinion about it / I'd done anything stupid.

src/trans.cc
```diff
@@ -476,8 +476,11 @@ static Obj FuncTransformationListListNC(Obj self, Obj src, Obj ran)
             ptf2[i] = i;
         }
         for (i = LEN_LIST(src); 1 <= i; i--) {
-            ptf2[INT_INTOBJ(ELM_LIST(src, i)) - 1] =
-                INT_INTOBJ(ELM_LIST(ran, i)) - 1;
+            s = INT_INTOBJ(ELM_LIST(src, i));
+            r = INT_INTOBJ(ELM_LIST(ran, i));
+            if(s != r) {
```
fingolfin:
```suggestion
            if (s != r) {
```

src/trans.cc
```diff
@@ -487,8 +490,11 @@ static Obj FuncTransformationListListNC(Obj self, Obj src, Obj ran)
             ptf4[i] = i;
         }
         for (i = LEN_LIST(src); 1 <= i; i--) {
-            ptf4[INT_INTOBJ(ELM_LIST(src, i)) - 1] =
-                INT_INTOBJ(ELM_LIST(ran, i)) - 1;
+            s = INT_INTOBJ(ELM_LIST(src, i));
+            r = INT_INTOBJ(ELM_LIST(ran, i));
+            if(s != r) {
```
fingolfin:
```suggestion
            if (s != r) {
```

ChrisJefferson:
woops, done some clang-formatting.

The problem (I've tweaked the description to hopefully help) is that at the beginning of the function when we figure out the size of the transformation we need, we basically take the max of the values in the two arrays, EXCEPT we skip any index i where the two lists are equal. This makes sense, because if the transformation maps x to x for any x, we don't need to explicitly store that fact, as that's the default for transformations.

Except, if (say) the original arrays were [[1,1000], [2,1000]], we would decide we need a transformation of size 2, but then later try to fill in that the transformation maps 1000 to 1000, just writing into the memory past the end of the transformation.

This almost always won't be into another allocated bag (because we've only just allocated this memory), but it will make some future allocated bag be allocated with non-zero values, which is (in practice) basically as bad.

ChrisJefferson:
also chucked a comment in

james-d-mitchell:
This makes sense to me @ChrisJefferson, well spotted!

PR #3465, opened 2019-05-20
Reduce varnames.tst sensitivity to the list of loaded packages

alex-konovalov:
Loading recog package made the output containing some more global
variables, see https://github.com/gap-packages/recog/issues/61


alex-konovalov:
Backported to stable-4.10 in cbe24fa88e3e9f1d0671a97e7b4593f87131313f

PR #3466, opened 2019-05-21
Change http to https for many web links

ChrisJefferson:
I made a general pass through the codebase, looking for places to change http: to https:

I made sure these links actually work in https (quite a few servers have broken certificates it turns out).

As well as being more secure, at some point I think some servers might start turning off http.


alex-konovalov:
Thank you @ChrisJefferson! Are there any `http://` left after this?

Perhaps we need to advise the same to package authors...

ChrisJefferson:
Remaining ones are:

1) I didn't edit anything in any external packages (zlib, gmp, gc).

2) I didn't edit licenses, just because I don't like editing licenses (technically, I think it might change the license?)

3) Some servers have broken certificates (like www.math.sci.hiroshima-u.ac.jp and www.math.rwth-aachen.de )

fingolfin:
Re point 2: I think it is fine to edit HTTP -> HTTPs in "licenses". But in any case, I am not even sure which licenses you are referring to here?

There also seems various places this PR does not yet cover, perhaps you'll want to address them, too (but we can also do this in a future PR, of course). E.g. these (ignoring stuff generated by autotools, and in historic "changes" documentation):
```
.appveyor.yml:33:  - '%CYG_ROOT%\setup-%CYG_ARCH%.exe -qnNdO -R %CYG_ROOT% -s http://cygwin.mirror.constant.com -l %CYG_ROOT%/var/cache/setup %PACKAGES%'
Makefile.rules:286:# http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/
README.buildsys.md:155:<http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/>
configure.ac:517:  # Boehm garbage collector, see http://www.hboehm.info/gc/
```

ChrisJefferson:
The cygwin path has an invalid certificate, so won't work if you make it https. The 'make.mad-scientist' was complaining when I first did this PR but now seems to work.

I mostly didn't change things in anything relating to the build script, as I'm never positive what is recreated automatically. With regards licenses, there are copies of the GPL, which refer to http:// in lots of files related to the build system, which once again I decided not to edit.

fingolfin:
> With regards licenses, there are copies of the GPL, which refer to http:// in lots of files related to the build system, which once again I decided not to edit.

Where, I don't see any? Or perhaps you are talking about these texts?
```
# You should have received a copy of the GNU General Public License
# along with this program; if not, see <http://www.gnu.org/licenses/>.
```
But those are not licenses -- these are just pointers *to* licenses, and of course one may (and should!) update them to work. Well, except here, as I stated above, I wouldn't as those files are either generated by autotools or otherwise shipped from upstream sources, so I wouldn't modify them, just like I wouldn't modify the GMP code in `extern`.

So, all in all, I think this PR is now ready to be merged :-).

PR #3467, opened 2019-05-21
Add missing CHANGED_BAG call + assorted cleanups

ChrisJefferson:
Add a missing CHANGED_BAG call.

This PR also contains some assorted cleanups found while tracking down this problem. The 'valgrind' one is the most serious, the other two are just tidyups.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -136,11 +136,17 @@ static Stat * ADDR_STAT(Stat stat)
 
 void WRITE_EXPR(Expr expr, UInt idx, UInt val)
 {
+    GAP_ASSERT(expr / sizeof(Expr) + idx <
+               SIZE_BAG(BODY_FUNC(CURR_FUNC())) / sizeof(Expr));
+
     ADDR_EXPR(expr)[idx] = val;
 }
 
 static void WRITE_STAT(Stat stat, UInt idx, UInt val)
 {
+    GAP_ASSERT(stat / sizeof(Stat) + idx <
+               SIZE_BAG(BODY_FUNC(CURR_FUNC())) / sizeof(Stat));
+
     ADDR_STAT(stat)[idx] = val;
 }
 
--- a/src/plist.h
+++ b/src/plist.h
@@ -320,6 +320,7 @@ EXPORT_INLINE Obj NewPlistFromArray(const Obj * list, Int length)
     Obj o = NEW_PLIST(T_PLIST, length);
     SET_LEN_PLIST(o, length);
     memcpy(BASE_PTR_PLIST(o), list, length * sizeof(Obj));
+    CHANGED_BAG(o);
     return o;
 }
 
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -472,9 +472,7 @@ static void CANARY_ALLOW_ACCESS_BAG(Bag bag)
 
     BagHeader * header = BAG_HEADER(bag);
     VALGRIND_MAKE_MEM_DEFINED(
-        (char *)header + sizeof(header->memory_canary_padding1),
-        sizeof(*header) - sizeof(header->memory_canary_padding1) -
-            sizeof(header->memory_canary_padding2));
+        header, sizeof(*header) - sizeof(header->memory_canary_padding));
 }
 
 // Reverse CANARY_ALL_ACCESS_BAG, making the masterpointer, bag contents and
@@ -488,9 +486,7 @@ static void CANARY_FORBID_ACCESS_BAG(Bag bag)
 
     BagHeader * header = BAG_HEADER(bag);
     VALGRIND_MAKE_MEM_NOACCESS(
-        (char *)header + sizeof(header->memory_canary_padding1),
-        sizeof(*header) - sizeof(header->memory_canary_padding1) -
-            sizeof(header->memory_canary_padding2));
+        header, sizeof(*header) - sizeof(header->memory_canary_padding));
 }
 
 // Mark all bags as accessible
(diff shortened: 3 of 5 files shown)


src/code.c
```diff
@@ -119,6 +119,8 @@ Obj NewFunctionBody(void)
 */
 static Expr * ADDR_EXPR(Expr expr)
 {
+    GAP_ASSERT(STATE(PtrBody) == PTR_BAG(BODY_FUNC(CURR_FUNC())));
```
fingolfin: Doesn't this add major overhead to method execution?

Also, I don't really see how this assertion could ever be violated, given that we update `PtrBody` in a special callback, `VarsAfterCollectBags`.

src/code.c
```diff
@@ -131,16 +133,22 @@ static Expr * ADDR_EXPR(Expr expr)
...
 static Stat * ADDR_STAT(Stat stat)
 {
+    GAP_ASSERT(STATE(PtrBody) == PTR_BAG(BODY_FUNC(CURR_FUNC())));
+
     return (Stat *)STATE(PtrBody) + stat / sizeof(Stat);
 }
 
 void WRITE_EXPR(Expr expr, UInt idx, UInt val)
 {
+    GAP_ASSERT(expr/sizeof(Stat) + idx < SIZE_BAG(BODY_FUNC(CURR_FUNC())) / sizeof(Expr));
```
fingolfin: Why mix Expr and Stat like that? And why not simply this:
```suggestion
    GAP_ASSERT(expr + idx * sizeof(Expr) < SIZE_BAG(BODY_FUNC(CURR_FUNC())));
```

src/code.c
```diff
@@ -131,16 +133,22 @@ static Expr * ADDR_EXPR(Expr expr)
...
+
     ADDR_EXPR(expr)[idx] = val;
 }
 
 static void WRITE_STAT(Stat stat, UInt idx, UInt val)
 {
+    GAP_ASSERT(stat/sizeof(Stat) + idx < SIZE_BAG(BODY_FUNC(CURR_FUNC())) / sizeof(Expr));
```
fingolfin:
```suggestion
    GAP_ASSERT(stat + idx * sizeof(Stat) < SIZE_BAG(BODY_FUNC(CURR_FUNC())));;
```

src/code.c
ChrisJefferson: Technically that's different if expr doesn't divide sizeof(Stat), and what I wrote lines up more closely with what the code actually does. The Stat/Expr is just a cut+paste problem.

src/code.c
ChrisJefferson: took this one out. I mainly added it because I suspected something dodgy was going on around here, and checking everything (I now think it's fine)

ChrisJefferson:
I think now improved and cleaned up (and with a CHANGED_BAG).

src/code.c
fingolfin: Honestly, I don't see how it lines up more closely with how that code works. I'd say to the contrary that what I wrote corresponds more closely to how it works. My version also will catch the case where due to a bug, `expr` is not a multiple of `sizeof(Expr)` (e.g. because somebody passed in an immediate int expression), which one doesn't see if one divides `expr` by `sizeof(Expr)`.

src/plist.h
```diff
@@ -320,6 +320,7 @@ EXPORT_INLINE Obj NewPlistFromArray(const Obj * list, Int length)
     Obj o = NEW_PLIST(T_PLIST, length);
     SET_LEN_PLIST(o, length);
     memcpy(BASE_PTR_PLIST(o), list, length * sizeof(Obj));
+    CHANGED_BAG(o);
```
fingolfin: Aha, indeed, good catch! Let's hope this fixes our recent problems (I made `gac` use `NewPlistFromArgs`, which uses `NewPlistFromArray`, so it's plausible that this has now become a (more) noticeable problem.

src/code.c
```diff
@@ -136,11 +136,17 @@ static Stat * ADDR_STAT(Stat stat)
 
 void WRITE_EXPR(Expr expr, UInt idx, UInt val)
 {
+    GAP_ASSERT(expr / sizeof(Expr) + idx <
+               SIZE_BAG(BODY_FUNC(CURR_FUNC())) / sizeof(Expr));
+
```
fingolfin: As I wrote in another comment (which is not visible here, which is why I am repeating it here):

Honestly, I don't see how it lines up more closely with how that code works. I'd say to the contrary that what I wrote corresponds more closely to how it works. My version also will catch the case where due to a bug, `expr` is not a multiple of `sizeof(Expr)` (e.g. because somebody passed in an immediate int expression), which one doesn't see if one divides `expr` by `sizeof(Expr)`. 

So, I'd still suggest this instead:
```C
    GAP_ASSERT(expr + idx * sizeof(Expr) < SIZE_BAG(BODY_FUNC(CURR_FUNC())));
```
and similar below.

src/code.c
ChrisJefferson: The code for ADDR_EXPR/ADDR_STAT does `(Expr *)STATE(PtrBody) + expr / sizeof(Expr);`, so it always divides by sizeof(Expr), and would work fine if it wasn't a multiple.

If we want to require expr is a multiple of sizeof(Expr), we could of course also check that property.

src/code.c
ChrisJefferson: (repeating myelf, so sorry if you see this twice):

I tried to line up with what the code does, and ADDR_EXPR does: `(Expr *)STATE(PtrBody) + expr / sizeof(Expr);`, and doesn't require expr % sizeof(Expr) is 0 (although we could add that as a requirement too).

fingolfin: Ahhh, now I understand!

Well, in any case, this is better than no check, so, whatever :-)

PR #3468, opened 2019-05-21
Add a `PrintObj` method for general domains which know their `GeneratorsOfDomain`

ssiccha:
Supersedes #3460.

This enables nicer printing of mappings between sets:
```
gap> MappingByFunction(Domain([1..5]), Domain([1..5]), x -> x);
MappingByFunction( Domain([ 1 .. 5 ]), Domain([ 1 .. 5 ]), function( x ) ... end )
```

### DIFF
--- a/lib/domain.gi
+++ b/lib/domain.gi
@@ -207,6 +207,18 @@ InstallMethod( GeneratorsOfDomain,
 
 #############################################################################
 ##
+##  PrintObj
+##
+InstallMethod( PrintObj,
+    "for a domain with GeneratorsOfDomain",
+    [ HasGeneratorsOfDomain and IsDomain ],
+    function( dom )
+    Print( "Domain(", GeneratorsOfDomain( dom ), ")" );
+    end );
+
+
+#############################################################################
+##
 #M  AsList( <D> ) . . . . . . . . . . . . . . .  list of elements of a domain
 #M  Enumerator( <D> ) . . . . . . . . . . . . .  list of elements of a domain
 ##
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
This looks reasonable -- the failing tests is because the manual needs adjusting to the new nicer output.

lib/domain.gi
```diff
@@ -389,3 +401,4 @@ InstallGlobalFunction( InstallAccessToGenerators,
         end );
 
     end );
+
```
fingolfin: Why add this line?
```suggestion
```

lib/domain.gi
```diff
@@ -205,6 +205,18 @@ InstallMethod( GeneratorsOfDomain,
     AsList );
 
 
+#############################################################################
+##
+##  PrintObj
+##
+InstallMethod( PrintObj,
+    "for a domain with HasGeneratorsOfDomain",
+    [HasGeneratorsOfDomain and IsDomain],
```
fingolfin:
```suggestion
    [ HasGeneratorsOfDomain and IsDomain ],
```

lib/domain.gi
ssiccha: Was done by accident. I've removed it now.

ssiccha:
Ah, there were tests in the manuals that I missed. I'll have a look.

ssiccha:
Updated. It should work now.

lib/domain.gi
```diff
@@ -205,6 +205,18 @@ InstallMethod( GeneratorsOfDomain,
     AsList );
 
 
+#############################################################################
+##
+##  PrintObj
+##
+InstallMethod( PrintObj,
+    "for a domain with HasGeneratorsOfDomain",
```
fingolfin:
```suggestion
    "for a domain with GeneratorsOfDomain",
```
would be my preference, but it's not really that important.

ssiccha: updated :+1:

PR #611, opened 2016-02-11
WIP: Constructors

stevelinton:
So far this PR fixes the immediate problem reported in #601 that method selection for constructors can miss applicable methods.

There are still a number of things to do in this area:
1. Work out what method should be preferred for a constructor when several are applicable? 
   For instance what kind of group should `AbelianGroupCons(IsGroup,[0])` return?
   and make the method selection do it.
2. Document constructors 
3. Possible add a more efficient `IS_CONSTUCTOR` (might wait until ObjSets are merged because that would be a perfect solution.
4. Simplify code that works around issue #601, such as `AbelianGroup`

PR #3469, opened 2019-05-21
document DeclareCategoryCollections; document constructors;  improve reference manual; add NameOfCategoryCollections helper

fingolfin:
The commit adding an explanation for constructors is stolen from @stevelinton's PR #611, albeit in slightly modified form. If this PR here is merged, I'll update PR #611 accordingly.

### DIFF
--- a/doc/ref/domain.xml
+++ b/doc/ref/domain.xml
@@ -605,8 +605,8 @@ The functions <Ref Func="InstallSubsetMaintenance"/>,
 <Ref Func="InstallFactorMaintenance"/> are used to tell &GAP;
 under what conditions an attribute is maintained under taking subsets,
 or forming factor structures or isomorphic domains.
-This is used only when a new attribute is created,
-see&nbsp;<Ref Sect="Creating Attributes and Properties"/>.
+This is used only when a new attribute or property is created,
+see&nbsp;<Ref Func="NewAttribute"/> and <Ref Func="NewProperty"/>.
 For the attributes already available,
 such as <Ref Prop="IsFinite"/> and <Ref Prop="IsCommutative"/>,
 the maintenances are already notified.
@@ -680,8 +680,7 @@ and <Ref Filt="IsLieMatrix"/>.
 
 The following categories of elements are to be understood mainly as
 categories for all objects in a family,
-they are usually used as third argument of <C>NewFamily</C>
-(see&nbsp;<Ref Sect="Creating Families"/>).
+they are usually used as third argument of <Ref Func="NewFamily"/>.
 The purpose of each of the following categories is then to guarantee that
 each collection of its elements automatically lies in its collections
 category (see&nbsp;<Ref Func="CategoryCollections"/>).
--- a/doc/ref/methsel.xml
+++ b/doc/ref/methsel.xml
@@ -76,6 +76,77 @@ and then set the attribute tester for the object to <K>true</K>.
 
 
 <!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<Section Label="Constructors">
+<Heading>Constructors</Heading>
+
+Constructors are a special type of operation used to make new objects. The key
+difference compared to regular operations is that method selection works
+slightly differently for them: The first argument in a call to a constructor
+must always be a filter <Ref Sect="Filters"/>. The result of a method is
+expected to lie in that filter. Moreover, while normally method selection
+matches on the type of each argument, for a constructor the first argument is
+treated differently: instead of matching its type, the argument itself (which
+is a filter) must be a subset of the filter specified by the method which is
+being tested for match. In other words, the argument filter must imply the
+method filter.
+<P/>
+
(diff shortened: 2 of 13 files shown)


doc/ref/domain.xml
```diff
@@ -605,8 +605,8 @@ The functions <Ref Func="InstallSubsetMaintenance"/>,
 <Ref Func="InstallFactorMaintenance"/> are used to tell &GAP;
 under what conditions an attribute is maintained under taking subsets,
 or forming factor structures or isomorphic domains.
-This is used only when a new attribute is created,
-see&nbsp;<Ref Sect="Creating Attributes and Properties"/>.
+This is used only when a new attribute or property is created,
+see&nbsp;<Ref Sect="NewAttribute"/> and <Ref Sect="NewProperty"/>.
```
markusbaumeister: Is "Sect" the correct reference here? I thought "NewAttribute" was a method so it should be called by "Subsect", but that might not apply in this particular case.

doc/ref/domain.xml
```diff
@@ -680,8 +680,7 @@ and <Ref Filt="IsLieMatrix"/>.
 
 The following categories of elements are to be understood mainly as
 categories for all objects in a family,
-they are usually used as third argument of <C>NewFamily</C>
-(see&nbsp;<Ref Sect="Creating Families"/>).
+they are usually used as third argument of <Ref Func="NewFamily"/>.
```
markusbaumeister: I would leave the name "NewFamily" in there, like
"... third argument of <C>NewFamily<C> (<Ref Func="NewFamily">)"
Otherwise it looks strange (e.g. "third argument of 12.32").

doc/ref/methsel.xml
```diff
@@ -68,6 +68,71 @@ Depending on the type of the object,
...
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<Section Label="Constructors">
+<Heading>Constructors</Heading>
+
+Constructors are a special type of operation used to make new
+objects. The key difference is that the first argument in a call to a
+constructor is a filter <Ref Sect="Filters"/> rather than an
+object. This signifies filters in which the constructed object must
+lie and method selection is based on the value of this filter, rather
+than its type.
```
markusbaumeister: "rather than its type" is very hard to understand since "type" is unclear. From the first line of the paragraph I gather that "Constructor" is a type but this is probably not meant. I suggest replacing "than its type" by "than the types of its inputs" or something along those lines to make it clearer. Then, I would probably change the "type" from the first line into "kind".

doc/ref/methsel.xml
```diff
@@ -68,6 +68,71 @@ Depending on the type of the object,
...
+gap> InstallMethod(XCons, [IsSemigroup, IsInt], function(t,x) return FullTransformationMonoid(x); end);
+gap> XCons(IsGroup,3);
+<pc group of size 3 with 1 generators>
+gap> XCons(IsPermGroup,3);
+Sym( [ 1 .. 3 ] )
+gap> XCons(IsSemigroup,4);
+<full transformation monoid of degree 4>
+]]></Example>
+
+The example above shows sonme basic examples (usually a constructor
```
markusbaumeister: sonme -> some

doc/ref/types.xml
```diff
@@ -385,7 +502,14 @@ attribute value, and just removing the value might leave the data
 structures in an inconsistent state. If necessary, a new object can be
 constructed.
 <P/>
-Several attributes have methods for more than one argument.  For example
+Each method that is installed for an attribute via <Ref Func="InstallMethod"/>
```
markusbaumeister: In this paragraph I would quote the method names before their reference (4 instances).

doc/ref/domain.xml
fingolfin: It should be `Func`, thank you for catching this.

doc/ref/domain.xml
fingolfin: Actually, no, it is rendered like this:
> they are usually used as third argument of NewFamily (13.1-2).

doc/ref/methsel.xml
fingolfin: No, "Constructor" is not a type. You are right, this is hard to understand. What is really meant that, while method selection for operations depends on the types of the arguments, this is not the case for constructors... Anyway, I'll try to write a better text.

doc/ref/types.xml
fingolfin: I don't understand what you mean with that.

doc/ref/methsel.xml
fingolfin: Fixed.

doc/ref/methsel.xml
fingolfin: I rewrote that text, please have another look.

doc/ref/types.xml
markusbaumeister: This was a question concerning the rendering of links which does not warrant further change (I misinterpreted the rendering).

doc/ref/methsel.xml
markusbaumeister: Yes, it is much clearer now.

PR #3472, opened 2019-05-24
Document `IsRangeRep` and improve the documentation of `ConvertToRangeRep`, `IsRange`, and the section `Ranges`

ssiccha:
Documents `IsRangeRep`.

Restructures `ConvertToRangeRep`:
- remove some redundant statements
- move some explanations to the `IsRangeRep` doc

Adds examples to `IsRange` and the info that assigning to a range turns it
into a plain list.

Improves the example in the introduction of the `Ranges` section,

### DIFF
--- a/doc/ref/lists.xml
+++ b/doc/ref/lists.xml
@@ -1738,6 +1738,8 @@ Of course, unless you assign
 <C><A>last</A> + <A>second</A> - <A>first</A></C> to the entry
 <C><A>range</A>[ Length( <A>range</A> ) + 1 ]</C>,
 the resulting list will no longer be a range.
+Note that assigning to an entry of <A>range</A> will convert it back into
+a plain list.
 <P/>
 <Example><![CDATA[
 gap> r := [10..20];
@@ -1748,8 +1750,15 @@ gap> r[3];
 12
 gap> 17 in r;
 true
-gap> r[12] := 25;; r;  # r is no longer a range
-[ 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 25 ]
+gap> # r still is a range but is now represented as a plain list
+gap> r[1] := 10;; r;
+[ 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20 ]
+gap> IsRange(r);
+true
+gap> # r is no longer a range
+gap> r[12] := 25;;
+gap> IsRange(r);
+false
 gap> r := [1,3..17];
 [ 1, 3 .. 17 ]
 gap> Length( r );
@@ -1789,6 +1798,7 @@ and also a row vector (see&nbsp;<Ref Chap="Row Vectors"/>),
 because it contains no holes and all elements are integers.
 <P/>
 <#Include Label="IsRange">
+<#Include Label="IsRangeRep">
 <#Include Label="ConvertToRangeRep">
 
 </Section>
--- a/lib/list.g
+++ b/lib/list.g
@@ -345,10 +345,11 @@ InstallMethod( ASS_LIST,
 ##  tests if the object <A>obj</A> is a range, i.e. is a dense list of
 ##  integers that is also a range
 ##  (see&nbsp;<Ref Sect="Ranges"/> for a definition of <Q>range</Q>).
-##  <!-- shouldn't this better be a property?-->
 ##  <Example><![CDATA[
 ##  gap> IsRange( [1,2,3] );  IsRange( [7,5,3,1] );
 ##  true
+##  true
+##  gap> IsRange( [1 .. 3] );
 ##  true
 ##  gap> IsRange( [1,2,4,5] );  IsRange( [1,,3,,5,,7] );
 ##  false
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
 ##
 #R  IsRangeRep( <obj> )
 ##
+##  <#GAPDoc Label="IsRangeRep">
 ##  <ManSection>
 ##  <Filt Name="IsRangeRep" Arg='obj' Type='Representation'/>
 ##
 ##  <Description>
-##  For internally represented ranges, there is a special representation
-##  which requires only a small amount of memory.
+##  Tests whether <A>obj</A> is represented as a range,
+##  that is by internally storing only the first value, the in- or decrement,
+##  and the last value of the range.
+##  <P/>
+##  To test whether a list is a range in the mathematical sense see <Ref
+##  Filt="IsRange"/>.
+##  <P/>
+##  Lists created by the syntactic construct
+##  <C>[ <A>first</A>, <A>second</A>  .. <A>last</A> ]</C>,
+##  see <Ref Sect="Ranges"/>, are in <Ref Filt="IsRangeRep"/>.
+##  <P/>
+#   Note that if you modify an <Ref Filt="IsRangeRep"/> object by assigning to
+#   one of its entries, or by using <Ref Oper="Add"/> or <Ref Oper="Append"/>,
+##  then the range may be converted into a plain list, even
+##  though the resulting list may still be a range, mathematically.
+##  <P/>
+##  <Example><![CDATA[
+##  gap> IsRangeRep( [1 .. 3] );
+##  true
+##  gap> IsRangeRep( [1, 2, 3] );
+##  false
+##  gap> l := [1..3];;
+##  gap> l[1] := 1;;
+##  gap> l;
+##  [ 1, 2, 3 ]
+##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
+##  <#/GAPDoc>
 ##
 DeclareRepresentationKernel( "IsRangeRep",                                   
     IsInternalRep, [], IS_OBJECT, IS_RANGE_REP );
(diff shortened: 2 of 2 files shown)


doc/ref/lists.xml
```diff
@@ -1738,6 +1738,8 @@ Of course, unless you assign
 <C><A>last</A> + <A>second</A> - <A>first</A></C> to the entry
 <C><A>range</A>[ Length( <A>range</A> ) + 1 ]</C>,
 the resulting list will no longer be a range.
+Note that in any case assigning to <A>range</A> will convert it back into
```
fingolfin: It's not assigning to `range` that convert it, but rather assigning to one of its entries. I'd also drop the "in any case". Maybe like this?
```suggestion
Note that assigning to an entry of <A>range</A> will convert it back into
```

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
 ##  <Filt Name="IsRangeRep" Arg='obj' Type='Representation'/>
 ##
 ##  <Description>
-##  For internally represented ranges, there is a special representation
-##  which requires only a small amount of memory.
+##  Tests whether <A>obj</A> is represented as a range,
+##  that is by internally storing only the first value, the in- or decrement,
+##  and the last value of the range.
+##  <P/>
+##  To test whether a list is an arithmetic pro- or degression in the
```
fingolfin: I find this reference to "pro- or degression" confusing, can't we just refer to "range" here?

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
+##  mathematical sense see <Ref Filt="IsRange"/>.
+##  <P/>
+##  Lists created by the syntactic construct
+##  <C>[ <A>first</A>, <A>second</A>  .. <A>last</A> ]</C>,
+##  see <Ref Sect="Ranges"/> , are in <Ref Filt="IsRangeRep"/>.
```
fingolfin:
```suggestion
##  see <Ref Sect="Ranges"/>, are in <Ref Filt="IsRangeRep"/>.
```

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
+##  <P/>
+##  Note that if you change <Ref Filt="IsRangeRep"/> object by assignment,
```
fingolfin:
```suggestion
##  Note that if you modify an <Ref Filt="IsRangeRep"/> object by assigning to one of its entries,
```

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
+##  <Ref Oper="Add"/> or <Ref Oper="Append"/>,
```
fingolfin:
```suggestion
##  or by using <Ref Oper="Add"/> or <Ref Oper="Append"/>,
```

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
+##  the range will be converted into a plain list, even if the resulting list
```
fingolfin:
```suggestion
##  the range may be converted into a plain list, even if the resulting list
```
If we write "will", then we essentially promise this will always be the case, thus prevent any future changes to this behavior.

lib/list.g
```diff
@@ -392,53 +419,32 @@ DeclareRepresentationKernel( "IsRangeRep",
...
-##  representation.
+##  the representation of <A>list</A> to <Ref Filt="IsRangeRep"/>.
 ##  A call of <Ref Func="ConvertToRangeRep"/> for a list that is not a range
 ##  is ignored.
 ##  <P/>
-##  If you change a mutable range that is represented in the compact way,
-##  by assignment, <Ref Oper="Add"/> or <Ref Oper="Append"/>,
-##  the range will be converted to the ordinary representation, even if the
-##  change is such that the resulting list is still a proper range.
+##  Note that in general it is safer to directly construct <A>list</A> as a
```
fingolfin: In which way is it "safer"?

lib/list.g
```diff
@@ -392,53 +419,32 @@ DeclareRepresentationKernel( "IsRangeRep",
...
-##  represented in the ordinary way and that you now want to convert it to
-##  the compact representation to save space.
-##  Then you should call <Ref Func="ConvertToRangeRep"/> with that list as an
-##  argument.
-##  You can think of the call to <Ref Func="ConvertToRangeRep"/> as a hint
-##  to &GAP; that this list is a proper range.
+##  Note that if you change a range in <Ref Filt="IsRangeRep"/> by assignment,
+##  <Ref Oper="Add"/> or <Ref Oper="Append"/>,
+##  the range will be converted into a plain list, even if the
+##  resulting list is still a range.
```
fingolfin: See above (but is it really necessary to repeat this here?)

doc/ref/lists.xml
ssiccha: :+1:

lib/list.g
ssiccha: yepp, that's simpler. Done.

lib/list.g
ssiccha: :+1:

lib/list.g
ssiccha: :+1:

lib/list.g
ssiccha: :+1:

lib/list.g
ssiccha: Unfortunately this already was stated in the documentation of `ConvertToRangeRep`.
```
##  If you change a mutable range that is represented in the compact way,
##  by assignment, <Ref Oper="Add"/> or <Ref Oper="Append"/>,
##  the range will be converted to the ordinary representation, even if the
##  change is such that the resulting list is still a proper range.
```

ssiccha: Granted, it was not very discoverable there. So should we still change the wording here (and also in the `doc/ref/list.xml`) to say "may"?

lib/list.g
ssiccha: I'll drop it. It's now mentioned both in `IsRangeRep` and in the introductory section about `Ranges`.

lib/list.g
ssiccha: Do you mean I should add more explanation or that I should drop the comment?

I mean in the sense of:
If `list` is also known to another function `f` then e.g. `f` could afterwards assign to an entry of `list` and thus convert it back into a plain list. Or, `f` could assume that `list` is a plain list and call special kernel functions. In this case GAP could crash or - even worse - produce wrong results.

What do you think?

ssiccha:
> The commit message claims: "Restructures ConvertToRangeRep:" but I think this should be "Restructures the ConvertToRangeRep documentation:"
Done

lib/list.g
fingolfin: I'd prefer to change it to `may` in both places. But perhaps others (@stevelinton @ChrisJefferson ...) have differing views.

lib/list.g
fingolfin: I meant the question directly like that: It was completely unclear to me what you meant with "safer" here. And frankly, I am still not sure I follow after your explanation. How do you think one can trigger that crash here, from within GAP (of course from kernel code, it is always trivial to crash).

So, I'd either remove, replace, or explain this sentence, but it's not OK in the current form.

ChrisJefferson: In anything with isn't kernel documentation, then users shouldn't really worry about the representation as much as possible, and certainly shouldn't think anything is "safer".

ssiccha: > So, I'd either remove, replace, or explain this sentence, but it's not OK in the current form.

I agree. In hindsight implying that `ConvertToRangeRep` is in some way "unsafe" is way over the top.

My thought was to mention that (at least from my POV) one should avoid converting or changing objects that you didn't create yourself or that other functions may depend on. But the documentation of `ConvertToRangeRep` is hardly the place to do that. I'm gonna remove this sentence from my PR. Just a thought: maybe a comment along these lines may be more appropriate, _if at all_, in the documentation of `MakeImmutable`.

ssiccha: I removed the sentence.

lib/list.g
```diff
@@ -369,14 +370,40 @@ DeclareCategoryKernel( "IsRange",
...
+##  To test whether a list is a range in the mathematical sense see <Ref
+##  Filt="IsRange"/>.
+##  <P/>
+##  Lists created by the syntactic construct
+##  <C>[ <A>first</A>, <A>second</A>  .. <A>last</A> ]</C>,
+##  see <Ref Sect="Ranges"/>, are in <Ref Filt="IsRangeRep"/>.
+##  <P/>
+#   Note that if you modify an <Ref Filt="IsRangeRep"/> object by assigning to
+#   one of its entries, or by using <Ref Oper="Add"/> or <Ref Oper="Append"/>,
+##  the range will be converted into a plain list, even if the resulting list
```
wilfwilson:
```suggestion
##  then the range may be converted into a plain list, even
##  though the resulting list may still be a range, mathematically.
```

lib/list.g
ssiccha: @wilfwilson @fingolfin Nobody spoke up against it, so I'll change it.

lib/list.g
ssiccha: I like it. Thanks :)

lib/list.g
```diff
@@ -369,14 +370,41 @@ DeclareCategoryKernel( "IsRange",
...
+##  <P/>
+##  Lists created by the syntactic construct
+##  <C>[ <A>first</A>, <A>second</A>  .. <A>last</A> ]</C>,
+##  see <Ref Sect="Ranges"/>, are in <Ref Filt="IsRangeRep"/>.
+##  <P/>
+#   Note that if you modify an <Ref Filt="IsRangeRep"/> object by assigning to
+#   one of its entries, or by using <Ref Oper="Add"/> or <Ref Oper="Append"/>,
+##  then the range may be converted into a plain list, even
+##  though the resulting list may still be a range, mathematically.
+##  is still a range.
```
wilfwilson: This line needs to go!

ssiccha: ooh, yes. on it.

ssiccha: Done.

ssiccha:
I just pushed the newest version.

ssiccha:
Rebased onto master.

PR #3473, opened 2019-05-24
Improve an AsList method for domains with stored GeneratorsOfDomain

ssiccha:
The new method ensures that no copy of GeneratorsOfDomain is created if
GeneratorsOfDomain already is a duplicate free list.

### DIFF
--- a/lib/domain.gi
+++ b/lib/domain.gi
@@ -244,12 +244,19 @@ InstallImmediateMethod( Enumerator,
 InstallMethod( AsList,
     "for a domain with stored domain generators",
     [ IsDomain and HasGeneratorsOfDomain ],
-    D -> DuplicateFreeList( GeneratorsOfDomain( D ) ) );
+    function( D )
+        if HasIsDuplicateFreeList( GeneratorsOfDomain( D ) )
+                and IsDuplicateFreeList( GeneratorsOfDomain( D ) ) then
+            return GeneratorsOfDomain( D );
+        else
+            return DuplicateFreeList( GeneratorsOfDomain( D ) );
+        fi;
+    end );
 
 InstallMethod( Enumerator,
     "for a domain with stored domain generators",
     [ IsDomain and HasGeneratorsOfDomain ],
-    D -> DuplicateFreeList( GeneratorsOfDomain( D ) ) );
+    AsList );
 
 
 #############################################################################
--- a/tst/testinstall/domain.tst
+++ b/tst/testinstall/domain.tst
@@ -27,5 +27,21 @@ Domain([ 1 .. 5 ])
 gap> Domain(FamilyObj(1), []);
 Domain([  ])
 
+# AsList and Enumerator for domains which know their GeneratorsOfDomain
+gap> r := Immutable([1..3]);;
+gap> d := Domain(r);;
+gap> IsIdenticalObj(AsList(d), r);
+true
+gap> IsIdenticalObj(Enumerator(d), r);
+true
+gap> r := Immutable([1,2,3,1]);;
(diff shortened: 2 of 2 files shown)


ssiccha:
Let's see whether any of the manual tests fail due to new printing of e.g. `AsList(Domain([1 .. 3]))`.

lib/domain.gi
```diff
@@ -232,12 +232,18 @@ InstallImmediateMethod( Enumerator,
...
+        else
+            return DuplicateFreeList( GeneratorsOfDomain( D ) );
+        fi;
+    end );
 
 InstallMethod( Enumerator,
     "for a domain with stored domain generators",
     [ IsDomain and HasGeneratorsOfDomain ],
-    D -> DuplicateFreeList( GeneratorsOfDomain( D ) ) );
+    D -> AsList(D) );
```
fingolfin: Do we have any tests for either of these methods, which compare the before/after situation?

ssiccha: I have checked and I didn't find any. Do you want me to add some?

I'd also be fine to remove the `AsList` here and instead duplicate the function from above.

fingolfin:
```suggestion
    AsList );
```
It's fine for me this way, but I'd slightly simplify the code, as suggested here.

ssiccha: of course. thanks :+1:

ssiccha: of course. thanks :+1:

ssiccha:
I resolved the merge conflict.

lib/domain.gi
```diff
@@ -244,12 +244,18 @@ InstallImmediateMethod( Enumerator,
 InstallMethod( AsList,
     "for a domain with stored domain generators",
     [ IsDomain and HasGeneratorsOfDomain ],
-    D -> DuplicateFreeList( GeneratorsOfDomain( D ) ) );
+    function( D )
+        if IsDuplicateFreeList( GeneratorsOfDomain( D ) ) then
+            return GeneratorsOfDomain( D );
+        else
+            return DuplicateFreeList( GeneratorsOfDomain( D ) );
```
fingolfin: Thinking about this some more, I am not really sure anymore whether this is an "improvement", as the description of this PR claims: `AsList` is an attribute. So we are talking about a one-time cost here. Now, with the old code, we'd just call `DuplicateFreeList`, which performs an `O(n^2)` algorithm.
With this "improvement", we instead first run `IsDuplicateFreeList`, which also is `O(n^2)`. And if it fails, we still run `DuplicateFreeList`. So while there may be scenarios where this saves something (namely memory, but not performance), there are also scenarios where it results in a slow-down.

Are there any real-world applications that motivate this change?

ssiccha: Oh, I'm completely fine with changing the condition to
`HasIsDuplicateFreeList( GeneratorsOfDomain( D ) ) and IsDuplicateFreeList( GeneratorsOfDomain( D ) )`.

ssiccha: > Are there any real-world applications that motivate this change?
I want `AsList(Domain([1..3]))` to return the range `[1..3]`. At the moment gap returns the plain list `[1,2,3]`. As the range `[1..3]` knows that it is a `IsDuplicateFreeList` in this situation there should be no reason to copy it and create a new plain list.

ssiccha: I changed the condition to
`HasIsDuplicateFreeList( GeneratorsOfDomain( D ) ) and IsDuplicateFreeList( GeneratorsOfDomain( D ) )`

fingolfin: With `HasIsDuplicateFreeList` it is of course fine.

PR #3476, opened 2019-05-25
Fix #3474  wrong result from \< for equal but not identical large integers

### LINKED ISSUE #3474: Wrong result from LT for rational numbers!
Arising from a bug that Leonard Soicher had, we have a reproducible serious bug in rational number arithmetic
### Observed behaviour
```
gap> x := -9303829819079165/4076863488;
-9303829819079165/4076863488
gap> x < x; 
false
gap> y := -9303829819079165/4076863488;
-9303829819079165/4076863488
gap> x < y;
true
gap> x = y;
true
gap> y < x;
true
```
### Expected behaviour
`x < y` and `y < x` should return false.

```
[sal@MCC02XF17NJGH8 crypting-0.9 (master)]$ gap4
 ┌───────┐   GAP 4.10dev-2057-g8e281e4 of today
 Packages:   AClib 1.3.1, Alnuth 3.1.0, AtlasRep 1.5.1, AutoDoc 2018.09.20, AutPGrp 1.10, CRISP 1.4.4, Cryst 4.1.18, CrystCat 1.1.8, 
             CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.2, IRREDSOL 1.4, LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.14, 
             PrimGrp 3.3.2, RadiRoot 2.8, ResClasses 4.7.1, SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.7, TransGrp 2.0.4, 
             utils 0.59
```

I'm looking into this now.

stevelinton:
# Description
Fix issue #3474 with \< for equal but not identical large integers

## Text for release notes 
This release fixes a **dangerous** bug in the comparison of large negative integers. If 'x' and 'y' are
equal, but not identical, large negative numbers then `x < y` returned `true` instead of `false`. This error was introduced in GAP 4.10.1 on ????? 

## Further details

Also introduces tests.

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -1141,6 +1141,58 @@ static Obj FuncINT_STRING(Obj self, Obj string)
 
 /****************************************************************************
 **
+*F  CmpInt( <opL>, <opR> ) . . . . . . . . . . . . . . . compare two integers
+**
+**  'CmpInt' return a negative, zero or positive integer according as
+**  <opL> is less than, equal to, or greater than <opR>.
+**
+**  It is used to implement EqInt and LtInt. Inlining and optimisation
+**  can reasonably be expected to take care of any minor inefficiencies
+**  resulting from this code sharing.
+*/
+static Int CmpInt(Obj opL, Obj opR)
+{
+    CHECK_INT(opL);
+    CHECK_INT(opR);
+
+    /* compare two small integers */
+    if (ARE_INTOBJS(opL, opR))
+        return (Int)opL - (Int)opR;
+
+    /* Distinct types */
+    if (TNUM_OBJ(opL) != TNUM_OBJ(opR)) {
+
+        /* There are just three possible types here and we know we
+           have two distinct ones, so there are six
+           possibilities. Nevertheless the code below which appears to
+           look at just four conditions, is correct, as if
+           IS_INTNEG(opL) is true (for instance) it doesn't matter
+           whether opR is immediate or large and positive */
+        if (IS_INTNEG(opL) || IS_INTPOS(opR))
+            return -1;
+        if (IS_INTNEG(opR) || IS_INTPOS(opL))
+            return 1;
+
+        /* should never get here */
+        GAP_ASSERT(0);
+    }
+
+    Int  ucmp; /* comparison of unsigned values */
+    UInt sl = SIZE_INT(opL);
+    UInt sr = SIZE_INT(opR);
+    if (sl < sr)
+        ucmp = -1;
+    else if (sr < sl)
+        ucmp = 1;
+    else
+        ucmp = mpn_cmp((mp_srcptr)CONST_ADDR_INT(opL),
+                       (mp_srcptr)CONST_ADDR_INT(opR), sl);
+
+    return IS_INTNEG(opL) ? -ucmp : ucmp;
+}
+
+/****************************************************************************
+**
 *F  EqInt( <opL>, <opR> ) . . . . . . . . . .  test if two integers are equal
 **
 **  'EqInt' returns 1 if the two integer arguments <opL> and <opR> are equal
@@ -1148,26 +1200,7 @@ static Obj FuncINT_STRING(Obj self, Obj string)
 */
 Int EqInt(Obj opL, Obj opR)
 {
-    CHECK_INT(opL);
-    CHECK_INT(opR);
-
-    /* compare two small integers */
-    if (ARE_INTOBJS(opL, opR))
-        return opL == opR;
-
-    /* a small int cannot equal a large int */
-    if (IS_INTOBJ(opL) != IS_INTOBJ(opR))
-        return 0;
-
-    /* compare the sign and size */
-    if (TNUM_OBJ(opL) != TNUM_OBJ(opR) || SIZE_INT(opL) != SIZE_INT(opR))
-        return 0;
-
-    if (mpn_cmp((mp_srcptr)CONST_ADDR_INT(opL),
-                (mp_srcptr)CONST_ADDR_INT(opR), SIZE_INT(opL)) == 0)
-        return 1;
-    else
-        return 0;
+    return CmpInt(opL, opR) == 0 ? 1 : 0;
 }
 
 /****************************************************************************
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Put `,x1,x2` at the top of the test at the end of the last #@local line, to make the test pass (just in case you haven't looked at this before).

src/integer.c
```diff
@@ -1141,33 +1141,62 @@ static Obj FuncINT_STRING(Obj self, Obj string)
...
+        return opL == opR ? 0 : (Int)opL < (Int)opR ? -1 : 1;
+
+    /* Distinct types */
+    if (TNUM_OBJ(opL) != TNUM_OBJ(opR)) {
+        if (IS_INTNEG(opL) || IS_INTPOS(opR))
+            return -1;
+        if (IS_INTNEG(opR) || IS_INTPOS(opL))
+            return 1;
+        /* should never get here */
+        GAP_ASSERT(0);
```
fingolfin: This is correct, because at least one of the two arguments must be a large integer, positive or negative, and those cases are all handled by the two preceding `if` clauses.

Meta note: Does the compiler detect that all those calls to `TNUM_OBJ` that are hidden behind those checks can be coalesced into a single check?

In fact, we could even do this, by exploiting that the TNUMs are almost ordered as we want them; we just `^ 1` to switch T_INT and T_INTPOS; after this trick, the modified tnum value can be used as a proxy for comparing the original inputs; since the order is "T_INTPOS^1 == 0  < T_INT^1 == 1 < T_INTNEG^1 == 3`, we have to switch left and right in the comparion.
```
    GAP_STATIC_ASSERT(T_INT == 0, "T_INT != 0");
    GAP_STATIC_ASSERT(T_INTPOS == 1, "T_INTPOS != 1");
    GAP_STATIC_ASSERT(T_INTNEG == 2, "T_INTNEG != 2");

    Int tnumL = TNUM_OBJ(opL);
    Int tnumR = TNUM_OBJ(opR);
    // trick: use the tnum values directly to compare ordering
    if (tnumL != tnumR)
        return (tnumR^1) - (tnumL^1);
```

src/integer.c
```diff
@@ -1141,33 +1141,62 @@ static Obj FuncINT_STRING(Obj self, Obj string)
...
+    }
 
-    /* a small int cannot equal a large int */
-    if (IS_INTOBJ(opL) != IS_INTOBJ(opR))
-        return 0;
+    UInt sl = SIZE_INT(opL);
+    UInt sr = SIZE_INT(opR);
```
fingolfin: If you declare `sl` and `sr` here, why not also `ucmp`?

src/integer.c
```diff
@@ -1141,33 +1141,62 @@ static Obj FuncINT_STRING(Obj self, Obj string)
...
-        return 0;
+/****************************************************************************
+**
+*F  EqInt( <opL>, <opR> ) . . . . . . . . . .  test if two integers are equal
+**
+**  'EqInt' returns 1 if the two integer arguments <opL> and <opR> are equal
+**  and 0 otherwise.
+*/
+
+
```
fingolfin:
```suggestion
```

src/integer.c
```diff
@@ -1141,33 +1141,62 @@ static Obj FuncINT_STRING(Obj self, Obj string)
...
-    else
-        return 0;
+/****************************************************************************
+**
+*F  EqInt( <opL>, <opR> ) . . . . . . . . . .  test if two integers are equal
+**
+**  'EqInt' returns 1 if the two integer arguments <opL> and <opR> are equal
+**  and 0 otherwise.
+*/
+
```
fingolfin:
```suggestion
```

src/integer.c
```diff
@@ -1177,40 +1206,10 @@ Int EqInt(Obj opL, Obj opR)
 **  'LtInt' returns 1 if the integer <opL> is strictly less than the
 **  integer <opR> and 0 otherwise.
 */
+
```
fingolfin:
```suggestion
```

src/integer.c
```diff
@@ -1177,40 +1206,10 @@ Int EqInt(Obj opL, Obj opR)
...
-    else
-        res = mpn_cmp((mp_srcptr)CONST_ADDR_INT(opL),
-                      (mp_srcptr)CONST_ADDR_INT(opR), SIZE_INT(opL)) < 0;
-
-    /* if both arguments are negative, flip the result */
-    if (IS_INTNEG(opL))
-        res = !res;
-
-    return res;
+    return CmpInt(opL, opR) == -1 ? 1 : 0;
```
fingolfin: That does not match the documentation of `CmpInt`! Correct (and also otherwise better, IMHO) would be this:
```suggestion
    return CmpInt(opL, opR) < 0;
```

src/integer.c
stevelinton: Nothing in this code is writing via pointers, so the C compiler should be able to avoid multiple reads safely. I had a look at the assembler output, but it's been reorganised so far from the C that it's hard to tell. In any event it would be an L1 cache hit.

src/integer.c
stevelinton: I've basically done this. I left in the redundant  ? 1 : 0 because I think it makes clear that we are into 1 for true rather than non-zero for true territory, which can be useful.

src/integer.c
fingolfin: I had a look at the disassembly generated by Apple `clang` . First off, you are right, reading the TNUM seems to be optimized into a single memory access for each input object, which is good.

However, `CmpInt` is not inlined unless I add an `ALWAYS_INLINE`, so there is no specialization (and thus, optimization) for `EqInt`. And even if I do add `ALWAYS_INLINE`, wildly inefficient code is the result.

All in all, this makes me wonder if it's worth it to completely rewrite this code *again*, when this far simpler change would also fix the issue, while retaining the efficiency of the code:
```patch
diff --git a/src/integer.c b/src/integer.c
index b34d6c382..f35b2cae4 100644
--- a/src/integer.c
+++ b/src/integer.c
@@ -1199,18 +1199,18 @@ Int LtInt(Obj opL, Obj opR)
 
     /* signs are equal; compare sizes and absolute values */
     if (SIZE_INT(opL) < SIZE_INT(opR))
-        res = 1;
+        res = -1;
     else if (SIZE_INT(opL) > SIZE_INT(opR))
-        res = 0;
+        res = +1;
     else
         res = mpn_cmp((mp_srcptr)CONST_ADDR_INT(opL),
-                      (mp_srcptr)CONST_ADDR_INT(opR), SIZE_INT(opL)) < 0;
+                      (mp_srcptr)CONST_ADDR_INT(opR), SIZE_INT(opL));
 
     /* if both arguments are negative, flip the result */
     if (IS_INTNEG(opL))
-        res = !res;
+        res = -res;
 
-    return res;
+    return res < 0;
 }
 
 
```

PR #3479, opened 2019-05-30
kernel: simplify / optimize EqInt and LtInt

fingolfin:
This saves a few machine instructions for each, at least on x86; alas, overall, this is of course still on the slow path, so I am not sure it's worth it *shrug*

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -1151,23 +1151,18 @@ Int EqInt(Obj opL, Obj opR)
     CHECK_INT(opL);
     CHECK_INT(opR);
 
-    /* compare two small integers */
-    if (ARE_INTOBJS(opL, opR))
+    // if at least one input is a small int, a naive equality test suffices
+    if (IS_INTOBJ(opL) || IS_INTOBJ(opR))
         return opL == opR;
 
-    /* a small int cannot equal a large int */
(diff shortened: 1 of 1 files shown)


src/integer.c
```diff
@@ -1151,23 +1151,18 @@ Int EqInt(Obj opL, Obj opR)
...
-    /* a small int cannot equal a large int */
-    if (IS_INTOBJ(opL) != IS_INTOBJ(opR))
-        return 0;
-
-    /* compare the sign and size */
-    if (TNUM_OBJ(opL) != TNUM_OBJ(opR) || SIZE_INT(opL) != SIZE_INT(opR))
+    // compare the sign and size
+    // note: at this point we know that opL and opR are proper bags, so
+    // we can use TNUM_BAG instead of TNUM_OBJ
+    if (TNUM_BAG(opL) != TNUM_BAG(opR) || SIZE_INT(opL) != SIZE_INT(opR))
```
fingolfin: @ssiccha  The difference between `TNUM_OBJ` and `TNUM_BAG` is that the former first checks for immediate integers, and immediate FFE elements. But at this point, we already know that `opL` and `opR` are neither, so we can avoid those redundant checks. Granted, this is a minuscule optimization.

PR #3480, opened 2019-05-30
BuildPackages.sh: "make clean" before full build

fingolfin:
This way, if a user re-runs BuildPackages.sh in order to re-build packages,
there is a higher chance that this works as expected, instead of mixing old
compilation results with new ones.

For GAP packages with non-autoconf based build systems, we re-run configure
after the `make clean`, as for many of them, the latter removes their
generated Makefile (in violation of well-established UNIX conventions).

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -168,7 +168,10 @@ run_configure_and_make() {
     if grep Autoconf ./configure > /dev/null
     then
       echo_run ./configure --with-gaproot="$GAPROOT" $CONFIGFLAGS
+      echo_run "$MAKE" clean
     else
+      echo_run ./configure "$GAPROOT"
+      echo_run "$MAKE" clean
       echo_run ./configure "$GAPROOT"
     fi
     echo_run "$MAKE"


ChrisJefferson:
I seem to remember doing this in the past, and finding some packages, on make clean, remove their documentation, and then make not breaking it back.

ChrisJefferson:
I'm trying to find the package which did this, and I can't seem to find it. Maybe it was an earlier problem and has been fixed.

ChrisJefferson:
If I comparing building the packages, to doing "build, clean, build" then the following files are missing (of note, there is the odd minor file here and there)

gbnp has some files in doc/examples and some test files

GeneralizedMorhismsForCAP, LinearAlgebraForCAP, ModulePresentationsForCAP all don't regenerate their html docs, and also remove a file called 'maketest.g'

fingolfin:
OK, then I'll bug the CAP folks (@sebasguts and @sebastianpos) about fixing their Makefiles then :-).

fingolfin:
@ChrisJefferson huh, but looking closer, I don't understand: those three CAP packages don't have a `configure` script, so this PR should not even affect them?!

ChrisJefferson:
I didn't notice you don't run 'make clean' on packages without a configure. It looks like all packages which feature problematic 'make clean' happen to not have a configure script either.

fingolfin:
So, I just did a fresh package bootstrap; then used `cd pkg && git init . && git add . && git commit -m import` to turn `pkg` into a git repository. Then I run this modified build script. Afterwards, `git status` reported this (plus of course a ton of untracked files created by the build process):
```
    deleted:    FPLSA-1.2.2/doc/manual.pdf
    deleted:    FPLSA-1.2.2/doc/manual.six
    deleted:    anupq-3.2.1/doc/ANUPQ.toc
    modified:   guava-3.14/src/leon/aclocal.m4
    modified:   guava-3.14/src/leon/configure
    modified:   guava-3.14/src/leon/install-sh
    modified:   guava-3.14/src/leon/missing
```

So, it seems that FPLSA is affected (and I'll make sure to fix that there), but nothing else, at least nothing in the distribution.

fingolfin:
I just released fplsa 1.2.3 to improve this.

alex-konovalov:
Should this be mentioned in release notes?

fingolfin:
Regarding mentioning this in the release notes:
pro: it's a change that could in theory affect user's, and so it might be helpful to mention it. Also, technically it's an "enhancement"...
con: it probably won't affect any users, and even if, I think the fraction of those who'd be then helped by mentioning this in the release notes is close to or equal zero

All in all, I am completely neutral on whether to mention it or not

alex-konovalov:
Also it may affect package authors... It will make no harm to mention it, so let's do that.

PR #3481, opened 2019-05-31
Close streams which are created in PrintCSV

ChrisJefferson:
Code which calls PrintCSV many times with a filename will eventually start failing, as PrintCSV did not close files after it was finished with them.

### DIFF
--- a/tst/testinstall/stringobj.tst
+++ b/tst/testinstall/stringobj.tst
@@ -3,7 +3,7 @@
 ##  This file tests various aspects of strings in IsStringRep
 ##
 #@local OldCopyToStringRep,a2000,a3000,at2000,at3000,cp1,cp2,cp3
-#@local ret2000,ret3000,s,tmp,tmpdir,fname,dir,filename,sstream,fstream,t,u
+#@local ret2000,ret3000,s,tmp,tmpdir,fname,dir,filename,sstream,fstream,t,u,i
 gap> START_TEST("stringobj.tst");
 
 # RemoveCharacters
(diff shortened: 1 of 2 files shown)


fingolfin:
It's really a shame that we don't have finalizers to take care of this!

tst/testinstall/stringobj.tst
```diff
@@ -119,10 +119,11 @@ gap> s := ReadCSV( filename{[1..Length(filename)-4]}, ";" );
...
   rec( f := 11, f_2 := 12 ), rec( f := "yy", f1 := "x", f_2 := "zzz" ) ]
 gap> tmpdir := DirectoryTemporary();;
-gap> fname := Filename(tmpdir, "output.csv");;
-gap> PrintCSV( fname, s );
-gap> s = ReadCSV( fname );
-true
+gap> for i in [1..1000] do
+>   fname := Filename(tmpdir, "output.csv");;
+>   PrintCSV( fname, s );
+>   if s <> ReadCSV( fname ) then Print("failed loop", i,"\n"); fi;
```
fingolfin: If I look at this test in a year, I won't have a clue what it is about. I'd have to dig through the commit history. Perhaps a brief comment to explain that could be added?

ChrisJefferson: Good idea, now fixed

ChrisJefferson:
I agree the lack of finalisers is annoying, in particular this is still "bugged" if we exit due to an error early, but in current GAP that's basically unfixable.

PR #3483, opened 2019-06-01
Switch from 2GB memory space to 3/4 of physical memory

ChrisJefferson:
The current 2GB memory limit was set in 2012, and memory has grown since then.

On 64-bit OSes, this switches to a default of 3/4 of physical memory, or 2GB, whichever is larger. 32-bit OSes are left at their previous 1GB (just because I don't want to think about issues of filling up the memory space on 32-bit OSes, and most people won't be using them any more).

Tested on linux, mac and cygwin.

For release notes:

* Changed the pre-set memory limit default from 2GB to 3/4 of physical memory. Use the '-o' option if you want to change this limit.

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -1063,6 +1063,14 @@ void InitSystem (
 #ifdef SYS_IS_64_BIT
     SyStorMin = 128 * 1024L;
     SyStorMax = 2048*1024L;          /* This is in kB! */
+#if defined(HAVE_SYSCONF)
+#if defined(_SC_PAGESIZE) && defined(_SC_PHYS_PAGES)
+    // Set to 3/4 of memory size (in kB), if this is larger
+    Int SyStorMaxFromMem =
+        (sysconf(_SC_PAGESIZE) * sysconf(_SC_PHYS_PAGES) * 3L) / 4 / 1024;
+    SyStorMax = SyStorMaxFromMem > SyStorMax ? SyStorMaxFromMem : SyStorMax;
+#endif
+#endif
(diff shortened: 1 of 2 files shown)


src/system.c
```diff
@@ -1063,6 +1063,12 @@ void InitSystem (
 #ifdef SYS_IS_64_BIT
     SyStorMin = 128 * 1024L;
     SyStorMax = 2048*1024L;          /* This is in kB! */
+#if defined(HAVE_SYSCONF)
+#if defined(_SC_PAGESIZE) && defined(_SC_PHYS_PAGES)
```
wilfwilson: Not a nitpick, genuine question: is there a reason why these `#if`s are split across two lines, not not just one with another `&&`?

src/system.c
```diff
@@ -1063,6 +1063,12 @@ void InitSystem (
...
+    // Set to 3/4 of memory size (in kB)
+    SyStorMax = (sysconf(_SC_PAGESIZE) * sysconf(_SC_PHYS_PAGES) * 3L) / 4 / 1024;
```
wilfwilson: Your description says
> On 64-bit OSes, this switches to a default of 3/4 of physical memory, or 2GB, whichever is larger

but I don't see how it's doing the 'whichever is larger' part. Isn't this just setting it to a default of ¾ of physical memory in any case?

ChrisJefferson: That's a mistake, will fix

src/system.c
ChrisJefferson: Just the line will be very long, and it looks weird when preprocessor lines are split

PR #3485, opened 2019-06-04
Fix a "method not found" error in `SubdirectProduct`

### LINKED ISSUE #3431: Problem with SubdirectProducts of FP Groups (inter alia)

fingolfin:
This is an alternative to PR #3437, with a less invasive change, which IMHO makes it more appropriate for backporting to GAP 4.10. The remaining changes in PR #3437 could of course still be added to GAP 4.11, although I remain somewhat sceptical about that.

Fixes #3431

### DIFF
--- a/lib/gprd.gi
+++ b/lib/gprd.gi
@@ -544,6 +544,11 @@ local gc,hc,S,info;
   fi;
   gh:=InverseGeneralMapping(gc)*gh;
   hh:=InverseGeneralMapping(hc)*hh;
+  # the ...Op is installed for `IsGroupHomomorphism'. So we have to enforce
+  # the filter to be set.
+  if not IsGroupHomomorphism(gh) or not IsGroupHomomorphism(hh) then
+    Error("mappings are not homomorphisms");
+  fi;
   S:=SubdirectProductOp(Image(gc,G),Image(hc,H),gh,hh);
   info:=rec(groups:=[G,H],
 	    homomorphisms:=[gh,hh],
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -4187,6 +4187,8 @@ local s, a, hom;
   hom:=EpimorphismSolvableQuotient(G,s);
   if Size(Image(hom))<>s then
     Error("group is not solvable");
+  else
+    SetIsInjective(hom, true);
   fi;
   if a<>fail then
     hom:=a*hom;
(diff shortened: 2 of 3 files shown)


fingolfin:
Backported to `stable-4.10` in the commits feb142c6d47c0d39443cf0a5ae7c5682bd8480fb, 27a86780574afec93f3e4aeec8bd4622194fde5f, 0b5d176f784231eb9b66f85a829c1cb62b3292c2

PR #3490, opened 2019-06-05
kernel: remove C_NEW_STRING_DYN and MakeImmutableString, deprecate C_NEW_STRING

fingolfin:
The only reason I am not outright removing `C_NEW_STRING` is that Digraphs recently started to use it; but see also https://github.com/gap-packages/Digraphs/pull/198

For details, please also refer to the commit messages.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -192,7 +192,7 @@ Obj GET_FILENAME_BODY(Obj body)
 void SET_FILENAME_BODY(Obj body, Obj val)
 {
     GAP_ASSERT(IS_STRING_REP(val));
-    MakeImmutableString(val);
+    MakeImmutable(val);
     BODY_HEADER(body)->filename_or_id = val;
 }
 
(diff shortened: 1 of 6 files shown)


fingolfin:
For the record: no package ever used `MakeImmutableString`, as far as I can tell (looked back several versions). The profiling package used to use `C_NEW_STRING_DYN` until about version 2.0.1 or so, but the current version does not. Nothing else ever used `C_NEW_STRING_DYN`.

For `C_NEW_STRING`, the situation is a bit worse: Of course `digraphs` currently still uses it (but soon won't anymore). Beyond that, also `io`, `json`, `NormalizInterface` used to use it (and during the GAP 4.5 days also `fr`;  well, and also pargap used to use it, but we dropped that from distribution). Mind you, all these package stopped using it in their latest versions; but even in GAP 4.10.1, they do use it. So I think that's another reason to wait a bit longer with removing it.

PR #3491, opened 2019-06-05
Fix passing invalid rnams to records

### LINKED ISSUE #3489: Segmentation fault on wrong access to record components
The following input causes a segmentation fault (already in very old GAP versions).
```
gap> r:= rec( a:= 1 );;
gap> \.( r, "a" );
```
I think GAP should better give an appropriate error message.

In addition to that, I think that the documentation is not clear enough in this respect. 
The section "Record Access Operations" in the GAP Reference Manual says that record component names are encoded by numbers,
and that the correspondence is implemented by `NameRNam` and `RNamObj`.
It does not say what this means for non-infix calls to the operation `\.`
and to the methods installed for this operation.
In particular it does not show an example.
The following example would already be helpful.
```
gap> G:= SymmetricGroup( 4 );;
gap> G.1;
(1,2,3,4)
gap> \.( G, RNamObj( 1 ) );
(1,2,3,4)
gap> meth:= ApplicableMethod( \., [ G, 4711 ] );;
gap> meth( G, RNamObj( 1 ) );
(1,2,3,4)
```
If you agree that extending the documentation this way would be useful
then I can provide a pull request for such a change.

ChrisJefferson:
Fixes #3489 

This could be backported to 4.10, but seeing how long it has been around for, I don't think that is required.

## Text for release notes

Fix crashes when passing invalid arguments to functions for records: \\., IsBound\\., Unbind\\. and \\.\\:\\=

### DIFF
--- a/tst/testinstall/kernel/records.tst
+++ b/tst/testinstall/kernel/records.tst
@@ -24,10 +24,12 @@ a boolean or fail)
 gap> ForAll([1,-1,"abc"], x -> NameRNam(RNamObj(x)) = String(x));
 true
 gap> NameRNam(-1);
-Error, NameRName: <rnam> must be a positive small integer (not the integer -1)
+Error, NameRNam: <rnam> must be a positive small integer (not the integer -1)
 gap> NameRNam(fail);
-Error, NameRName: <rnam> must be a positive small integer (not the value 'fail\
-')
+Error, NameRNam: <rnam> must be a positive small integer (not the value 'fail'\
+)
+gap> NameRNam(2^26);
+Error, NameRNam: <rnam> must be a valid rnam (not the integer 67108864)
 
 # ElmRecHandler
 gap> ELM_REC(r, RNamObj(1));
--- a/tst/testinstall/recordname.tst
+++ b/tst/testinstall/recordname.tst
@@ -71,4 +71,99 @@ Syntax error: Identifiers in GAP must consist of at most 1023 characters. in s\
 tream:2
 aaaxyz := 1; # 1024 chars
  ^^^^^
+
+##
+gap> r := rec(y := 2);
+rec( y := 2 )
+gap> \.(r, RNamObj("y"));
+2
+gap> \.(r, RNamObj("x"));
+Error, Record Element: '<rec>.x' must have an assigned value
+gap> \.(r, 0);
+Error, Record Element: <rnam> must be a positive small integer (not the intege\
+r 0)
+gap> \.(r, -1);
+Error, Record Element: <rnam> must be a positive small integer (not the intege\
+r -1)
+gap> \.(r, 2^1000);
+Error, Record Element: <rnam> must be a positive small integer (not a large po\
+sitive integer)
+gap> \.(r, "a");
+Error, Record Element: <rnam> must be a positive small integer (not a list (st\
+ring))
+gap> \.(r, 1000000);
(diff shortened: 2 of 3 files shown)


src/error.c
```diff
@@ -512,6 +512,19 @@ void ErrorMayQuit(const Char * msg, Int arg1, Int arg2)
...
+/****************************************************************************
+**
+*F  GetValidRNam( <funcname>, <rnam> ) . check if <rnam> is a valid prec rnam
+*/
+UInt GetValidRNam(const char * funcname, Obj rnam)
+{
+    RequirePositiveSmallInt(funcname, rnam, "<rnam>");
+    UInt val = INT_INTOBJ(rnam);
+    RequireArgumentCondition(funcname, rnam, IS_VALID_RNAM(val),
+                             "must be a valid rnam");
```
fingolfin: Do we expect this check to be needed in other files? Otherwise, how about just defining it as `static` inside `records.c`? We do similar things for several other specialized argument checker functions.

And then you don't need to export `IS_VALID_RNAM` (and in fact, don't need to define it in the first place).

tst/testinstall/recordname.tst
```diff
@@ -71,4 +71,97 @@ Syntax error: Identifiers in GAP must consist of at most 1023 characters. in s\
 tream:2
 aaaxyz := 1; # 1024 chars
  ^^^^^
+gap> r := rec(y := 2);
```
fingolfin: I would suggest to separate the new tests from the existing ones; i find that this makes `.tst` files a lot easier to read.
```suggestion

# tests for issue #3489 
gap> r := rec(y := 2);
```

tst/testinstall/recordname.tst
```diff
@@ -71,4 +71,97 @@ Syntax error: Identifiers in GAP must consist of at most 1023 characters. in s\
...
+gap> Unbind\.(r, "a");
+Error, Record Unbind: <rnam> must be a positive small integer (not a list (str\
+ing))
+gap> Unbind\.(r, 1000000);
+Error, Record Unbind: <rnam> must be a valid rnam (not the integer 1000000)
+
+##
+gap> r;
+rec(  )
+gap> \.\:\=(r,RNamObj("y"), 2);
```
fingolfin: You could also use `ASS_REC` instead of `\.\:\=`, that might be a bit easier to grok. Same for `ISB_REC`, `ELM_REC`, `UNB_REC`, of course, though there it's perhaps less of a concern.

ChrisJefferson: I'd prefer to use \.\:\= and friends, are those are what we actually document.

ChrisJefferson:
Cleaned up. In the cleaning up I find one more function (NameRNam) where I could use my new test, so I did.

tst/testinstall/kernel/records.tst
```diff
@@ -28,6 +28,8 @@ Error, NameRName: <rnam> must be a positive small integer (not the integer -1)
 gap> NameRNam(fail);
 Error, NameRName: <rnam> must be a positive small integer (not the value 'fail\
 ')
+gap> NameRNam(2^29);
+Error, NameRName: <rnam> must be a valid rnam (not the integer 536870912)
```
fingolfin: This error message (and several before it) still has the old, incorrect `NameRName` in it, which is why the Travis tests failed.

ChrisJefferson: sorry, been away from my computer. Now hopefully fixed and poked.

PR #3494, opened 2019-06-07
Improvements to finding permutation representations and to automorphism group computations

hulpke:
# Description
Improvements (smaller degree, faster) for permutation representation for factor groups.

Performance improvements for automorphism group computations

## Text for release notes 
The routine for calculating automorphism groups has been improved.
The code for finding factor groups of  permutation groups performs better in speed and result.


fingolfin:
This PR now contains the fix for issue #3496 as well as completely separate new features / enhancements. That's not so nice, as it means the bug fix can't be applied right away, one has to review and merge the (much larger) other unrelated changes.

So the bugfix commit really should be in a separate PR. If @hulpke could do that, that'd be great. If not, I can also do it, it's a matter of 2 minutes after all.

(CC: @alex-konovalov @ChrisJefferson )

hulpke:
@fingolfin The fix is part of this PR, as it interacts with changes (at least manual example changes) that the other code provides. I'm not sure why keeping it together creates extra reviewing burden.

alex-konovalov:
@hulpke this problem causes a big backlog of tests in Jenkins, due to massive timeouts in three types of tests simultaneously, and takes resources needed for GAP 4.10.2 preparation. Reviewing a shorter PR is easier and can be done quickly. To go ahead with #3501, there is only one manual example update that needs to be borrowed from this PR.

hulpke:
Dear @wilfwilson Thank you for the comments. Yes, this is somewhat complicated code in that (a large part of) it combines a number of heuristics in the aim of getting a small-degree permutation representation at a moderate cost. The risk is less in returning a wrong result (basically in each case it is checked that the kernel is as desired), but in spending too long on attempts (as this is code used inside other routines).
I would not expect anyone to spend lots of time in going to the details, and the set-up of requiring  an  "approving review" is rather unfortunate here, as I'm not sure who, but the author, would be able to make confident statements about the code.

lib/factgrp.gi
```diff
@@ -758,7 +826,7 @@ totalcnt, interupt, u, nu, cor, zzz,bigperm,perm,badcores,max,i;
         max:=Filtered(max,x->IndexNC(G,x)<knowi and IsSubset(x,N)); 
         for i in max do
 	  cor:=Core(G,i);
-	  AddNaturalHomomorphismsPool(G,cor,i,Index(G,i));
+	  AddNaturalHomomorphismsPool(G,cor,i,IndexNC(G,i));
```
fingolfin: This line of code is not tested by the Travis test suite.

lib/factgrp.gi
```diff
@@ -773,13 +841,16 @@ totalcnt, interupt, u, nu, cor, zzz,bigperm,perm,badcores,max,i;
     cnt:=cnt-1;
 
     if cnt=0 and zzz>badi then
-      Info(InfoWarning,2,"index unreasonably large, iterating");
       badi:=Int(badi*12/10);
+      Info(InfoWarning+InfoFactor,2,
+        "index unreasonably large, iterating ",badi);
       cnt:=20;
+      totalcnt:=0;
+      interupt:=false;
```
fingolfin: This block of code is not tested by the Travis test suite.

lib/factgrp.gi
```diff
@@ -1113,7 +1183,7 @@ local gimg,img,dom,b,improve,bp,bb,i,k,bestdeg,subo,op,bc,bestblock,bdom,
 	    fi;
 	  else
 	    k:=KernelOfMultiplicativeGeneralMapping(op);
-	    SetSize(Range(op),Index(gimg,k));
+	    SetSize(Range(op),IndexNC(gimg,k));
```
fingolfin: This line of code (and the code surrounding it) is not tested by the Travis test suite.

lib/factgrp.gi
```diff
@@ -1160,6 +1230,86 @@ function(G,N)
   return fail;
 end);
 
+BindGlobal("FactPermRepMaxDesc",function(g,n,maxlev)
+local lim,deg,all,c,recurse,use;
+  if ValueOption("infactorpermrep")=true then return false;fi;
+  lim:=RootInt(IndexNC(g,n),3);
```
fingolfin: This line of code is not tested by the Travis test suite, i.e., the function never gets beyond the `return false` one line above.

lib/factgrp.gi
```diff
@@ -1169,20 +1319,56 @@ InstallMethod(FindActionKernel,"perm",IsIdenticalObj,
...
+    dom:=ChiefSeriesThrough(G,[oo,N]);
+    dom:=Filtered(dom,x->IsSubset(oo,x) and IsSubset(x,N));
+
+    i:=2;
+    while i<=Length(dom) do
+      j:=i;
+      while j<Length(dom)
+        #and HasElementaryAbelianFactorGroup(dom[i-1],dom[j+1]) 
+        and IndexNC(dom[i-1],dom[j+1])<=2000 do
+        j:=j+1;
```
fingolfin: This line of code is not tested by the Travis test suite.

lib/factgrp.gi
```diff
@@ -1169,20 +1319,56 @@ InstallMethod(FindActionKernel,"perm",IsIdenticalObj,
...
+          return v[PositionCanonical(v,rep^g)];
+          end);
+        for k in oo do
+          if Length(k)>1 then
+            u:=Stabilizer(G,k[1],function(x,g)
+              return v[PositionCanonical(v,x^g)];
+            end);
+            repeat
+              if not IsNormal(G,u) then
+                AddNaturalHomomorphismsPool(G,Core(G,u),u,IndexNC(G,u));
```
fingolfin: This line of code is not tested by the Travis test suite.

lib/factgrp.gi
```diff
@@ -1332,26 +1518,30 @@ local pool, dom, bestdeg, blocksdone, o, s, badnormals, cnt, v, u, oo, m,
...
 	blocksdone:=true;
       fi;
-      Info(InfoFactor,1,"  Random search found ",
-           DegreeNaturalHomomorphismsPool(G,N));
-    #if (bestdeg>500 and Index(G,o)<5000) or Index(G,o)<bestdeg then
-    #  # tell 'IODBB' not to doo too much blocksearch
-    #  o:=ImproveActionDegreeByBlocks(G,o,N,bestdeg<Index(G,o));
-    #  Info(InfoFactor,1,"  Blocks improve to ",Index(G,o),"\n");
-    #fi;
+      bestdeg:=DegreeNaturalHomomorphismsPool(G,N);
```
fingolfin: This line of code is not tested by the Travis test suite.

wilfwilson:
I'm pretty sure that the manual examples were passing yesterday, which would mean the failures were introduced by whatever was pushed 11 hours ago. But perhaps I remember wrongly. In any case, yes they need updating.

In case this is helpful to anyone reviewing, the are two commits that contain the bulk of the changes. The rest are:
- The commit `PERFORMANCE: Use SmallerDegree.. with cheap option`, which is obviously safe.
- Of course the `FIX whitespace issue in test` commit is fine too.
- The commit `ENHANCE: iterator wont jump at calculating lattice` is just a one-line change that as far as I can tell doesn't affect that correctness of the code.

lib/factgrp.gi
hulpke: I'm puzzled that it ever gets even to the first line. The option tested is to avoid infinite recursion and is only set by this function. That is, the function is never called by the tiny test examples. Is supect codecov got confused.

hulpke:
The manual issues were all trivial consequences of a rather harmless addition (call `SmallerDegree` with cheap option). This is resolved now.

Yes, not all code is tested by Travis,( and I find it of little help to admonish a line in which an `Index` call changed to `IndexNC`). This is because Travis imposes time limits, and these changes enable larger calculations. (Nor would I take a passed Travis test to be more than a check for stupid mistakes.) I have added an explicit test for a utility function and a more meaty test to testextra (that indeed exercises the newer chunks). It tests isomorphism between two group of order 245760, runs about 12 minutes with no assertions turned on, and was completely hopeless before.

This is somewhat complicated code in that (a large part of) it combines a number of heuristics in the aim of getting a small-degree permutation representation at a moderate cost. The risk is less in returning a wrong result (basically in each case it is checked that the kernel is as desired, and maps are homomorphisms as they arise from actions), but in spending too long on attempts (as this is code used inside other routines). I have run extensive tests of this as part of a larger calculation.

I consider it unreasonable to expect from anyone else to assert correctness of the code, and think the required “approving review” in cases as this ought to merely affirm the absence of obvious problems.

PR #3497, opened 2019-06-12
Compatibility with GAP.jl

sebasguts:
# Description

* Added installation of GAPTypes.jl to configure when compilation
  with Julia is requested
* Made MPtr, Bag, and LargeBag Julia types instances of GapObj

Thank You!

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -497,6 +497,9 @@ AS_IF([test "x$with_julia" != xno ],[
   AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to obtain JULIA_LIBS from julia-config.jl])])
   JULIA_LIBS=${JULIA_LIBS//\'/}
   AC_MSG_RESULT([${JULIA_LIBS}])
+  AC_MSG_NOTICE([installing GAPTypes.jl])
+  ${JULIA} -e 'import Pkg; Pkg.add("GAPTypes")'
+  AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to install GAPTypes.jl])])
 ],
 [
   AS_IF( [ test "x$with_gc" = xjulia ],
--- a/doc/changes/changes-4.10.xml
+++ b/doc/changes/changes-4.10.xml
@@ -667,6 +667,11 @@ Specify the Julia binary instead of the Julia prefix
 Export Julia <C>CFLAGS</C>, <C>LDFLAGS</C>, and <C>LIBS</C> to <C>sysinfo.gap</C>
 (<URL><LinkText>&Hash;3248</LinkText><Link>https://github.com/gap-system/gap/pull/3248</Link></URL>).
 </Item>
+<Item>
+Change <C>MPtr</C> Julia type of &GAP; objects to be a subtype of the abstract Julia <C>GapObj</C> type
+provided by the Julia package <C>GAPTypes.jl</C>
+(<URL><LinkText>&Hash;3497</LinkText><Link>https://github.com/gap-system/gap/pull/3497</Link></URL>).
+</Item>
 </List>
 
 Improved and extended functionality:
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -817,6 +817,25 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     max_pool_obj_size = jl_gc_max_internal_obj_size();
     jl_gc_enable_conservative_gc_support();
     jl_init();
+
+    // Import GAPTypes module to have access to GapObj abstract type.
+    // Needs to be done before setting any GC states
+    jl_eval_string("import GAPTypes");
+    if (jl_exception_occurred()) {
+        Panic("could not import GAPTypes module into Julia");
+    }
+    // Get GapObj abstract julia type
+    jl_module_t * gaptypes_module =
+        (jl_module_t *)jl_get_global(jl_main_module, jl_symbol("GAPTypes"));
+    if (jl_exception_occurred()) {
+        Panic("Could not read global GAPTypes variable from Julia");
+    }
+    jl_datatype_t * gapobj_type =
+        (jl_datatype_t *)jl_get_global(gaptypes_module, jl_symbol("GapObj"));
+    if (jl_exception_occurred()) {
+        Panic("could not read GapObj variable from Julia");
+    }
+
     JuliaTLS = jl_get_ptls_states();
     // These callbacks potentially require access to the Julia
     // TLS and thus need to be installed after initialization.
@@ -828,14 +847,18 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
 
     Module = jl_new_module(jl_symbol("ForeignGAP"));
     Module->parent = jl_main_module;
+
+    // Import GapObj type into ForeignGAP module
+    jl_module_use(Module, gaptypes_module, jl_symbol("GapObj"));
+
     jl_set_const(jl_main_module, jl_symbol("ForeignGAP"),
                  (jl_value_t *)Module);
     datatype_mptr = jl_new_foreign_type(
-        jl_symbol("MPtr"), Module, jl_any_type, MPtrMarkFunc, NULL, 1, 0);
-    datatype_bag = jl_new_foreign_type(jl_symbol("Bag"), Module, jl_any_type,
+        jl_symbol("MPtr"), Module, gapobj_type, MPtrMarkFunc, NULL, 1, 0);
+    datatype_bag = jl_new_foreign_type(jl_symbol("Bag"), Module, gapobj_type,
                                        BagMarkFunc, JFinalizer, 1, 0);
     datatype_largebag =
-        jl_new_foreign_type(jl_symbol("LargeBag"), Module, jl_any_type,
+        jl_new_foreign_type(jl_symbol("LargeBag"), Module, gapobj_type,
                             BagMarkFunc, JFinalizer, 1, 1);
 
     // export datatypes to Julia level


src/julia_gc.c
```diff
@@ -817,6 +817,12 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     max_pool_obj_size = jl_gc_max_internal_obj_size();
     jl_gc_enable_conservative_gc_support();
     jl_init();
+
+    // Import GAPTypes module to have access
+    // to GapObj abstract type.
+    // Needs to be done before setting any GC states
+    jl_eval_string("import GAPTypes");
```
fingolfin: Shouldn't we perform at least some minimal error checking, and abort if this fails, with a helpful error message?

Conversely: Have you tried what happens if this fails (e.g. because the user removed `GAPTypes` manually)

configure.ac
```diff
@@ -497,6 +497,9 @@ AS_IF([test "x$with_julia" != xno ],[
   AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to obtain JULIA_LIBS from julia-config.jl])])
   JULIA_LIBS=${JULIA_LIBS//\'/}
   AC_MSG_RESULT([${JULIA_LIBS}])
+  AC_MSG_NOTICE([installing GAPTypes.jl])
+  ${JULIA} -e 'import Pkg; Pkg.add(Pkg.PackageSpec(url="https://github.com/oscar-system/GAPTypes.jl", rev="master"))'
```
fingolfin: Is this meant to be the permanent form of the patch? Or is it meant to just do `Pkg.add("GAPType.jl")` eventually? I guess there are some merits to either... But I am not exactly comfortable with referencing `master` here, at least on the long run. Of course for now, while we develop stuff, we can do that; but ultimately, I'd expect this to reference a specific tag? Although now that I think about it, maybe not, so that we can hotfix things? Maybe a branch???

fingolfin: We should probably explain whatever the plan is in a brief comment here.

src/julia_gc.c
```diff
@@ -826,16 +832,26 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     jl_gc_set_cb_post_gc(PostGCHook, 1);
     // jl_gc_enable(0); /// DEBUGGING
 
+    // Get GapObj abstract julia type
+    jl_module_t * gaptypes_module =
+        (jl_module_t *)jl_get_global(jl_main_module, jl_symbol("GAPTypes"));
+    jl_datatype_t * gapobj_type =
+        (jl_datatype_t *)jl_get_global(gaptypes_module, jl_symbol("GapObj"));
+
     Module = jl_new_module(jl_symbol("ForeignGAP"));
```
fingolfin: We have `GAPTypes.jl`, `GAP.jl`, `ForeignGAP` vs. `GapObj` ... Should we rename the latter to `GAPObj` to make things more consistent?

src/julia_gc.c
```diff
@@ -826,16 +832,26 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     jl_gc_set_cb_post_gc(PostGCHook, 1);
     // jl_gc_enable(0); /// DEBUGGING
 
+    // Get GapObj abstract julia type
+    jl_module_t * gaptypes_module =
+        (jl_module_t *)jl_get_global(jl_main_module, jl_symbol("GAPTypes"));
+    jl_datatype_t * gapobj_type =
+        (jl_datatype_t *)jl_get_global(gaptypes_module, jl_symbol("GapObj"));
```
fingolfin: For both of those, I'd recommend that we test what happens if either symbol is unbound -- do they throw exceptions? Or just return a null pointer? In the latter case, we should handle that.

Also, perhaps we can at least add checks that the values we go there actually represent a module resp. a type object?

src/julia_gc.c
sebasguts: No, I have not checked it. But error checking would be a good idea.

configure.ac
sebasguts: This is meant to be replaced once the release of GAPTypes.jl is accepted.

src/julia_gc.c
sebasguts: I do not really care tbh. `GAPObj` looks more consistent, but less Julia . . .

src/julia_gc.c
sebasguts: Yep. I will add the error checks.

sebasguts:
I added error checking, and also updated this PR since GAPTypes is a released package now. If tests pass, I think this could be merged.

src/julia_gc.c
fingolfin: I am pretty sure this will throw an exception (see also `JULIAINTERFACE_EXCEPTION_HANDLER`). We probably should catch that exception and then call `Panic` if it occurs. And of course we should have tested that at least one (by forcefully removing GAPTypes from `~/.julia` before starting GAP; and once again when loading `GAP.jl`).

src/julia_gc.c
fingolfin: We should either decide between `GapObj` and `GAPObj` now; or else, be willing to potentially carry both of them around for backwards compatibility (which is fine, IMHO, just mentioning it). In either case, I am not worried :-)

src/julia_gc.c
```diff
@@ -817,6 +817,26 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
     max_pool_obj_size = jl_gc_max_internal_obj_size();
     jl_gc_enable_conservative_gc_support();
     jl_init();
+
+    // Import GAPTypes module to have access
+    // to GapObj abstract type.
```
fingolfin: Weird wrapping, perhaps that comment could go into a single line? (Clearly, this is *very* important ;-)

sebasguts:
@fingolfin @alex-konovalov I think I adressed both comments now.

src/julia_gc.c
sebasguts: I am totally happy with both spellings, too.

doc/changes/changes-4.10.xml
```diff
@@ -667,6 +667,9 @@ Specify the Julia binary instead of the Julia prefix
 Export Julia <C>CFLAGS</C>, <C>LDFLAGS</C>, and <C>LIBS</C> to <C>sysinfo.gap</C>
 (<URL><LinkText>&Hash;3248</LinkText><Link>https://github.com/gap-system/gap/pull/3248</Link></URL>).
 </Item>
+<Item>
+Make <C>MPtr</C> Julia type of GAP objects depend on abstract Julia <C>GapObj</C> type
```
alex-konovalov: Please use `&GAP;` for GAP

sebasguts: Done.

doc/changes/changes-4.10.xml
```diff
@@ -667,6 +667,9 @@ Specify the Julia binary instead of the Julia prefix
 Export Julia <C>CFLAGS</C>, <C>LDFLAGS</C>, and <C>LIBS</C> to <C>sysinfo.gap</C>
 (<URL><LinkText>&Hash;3248</LinkText><Link>https://github.com/gap-system/gap/pull/3248</Link></URL>).
 </Item>
+<Item>
+Make <C>MPtr</C> Julia type of &GAP; objects depend on abstract Julia <C>GapObj</C> type
+(<URL><LinkText>&Hash;3497</LinkText><Link>https://github.com/gap-system/gap/pull/3497</Link></URL>).
```
alex-konovalov: Missing </Item>

sebasguts: Sry. Fixed now

configure.ac
```diff
@@ -497,6 +497,9 @@ AS_IF([test "x$with_julia" != xno ],[
   AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to obtain JULIA_LIBS from julia-config.jl])])
   JULIA_LIBS=${JULIA_LIBS//\'/}
   AC_MSG_RESULT([${JULIA_LIBS}])
+  AC_MSG_NOTICE([installing GAPTypes.jl])
+  ${JULIA} -e 'import Pkg; Pkg.add("GAPTypes")'
+  AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to install GAPTypes.jl])])
```
fingolfin: Did you verify that this works as intended? (E.g. by changing the previous line to use `Pkg.add("foobarquux")` to force an error?

doc/changes/changes-4.10.xml
```diff
@@ -667,6 +667,10 @@ Specify the Julia binary instead of the Julia prefix
 Export Julia <C>CFLAGS</C>, <C>LDFLAGS</C>, and <C>LIBS</C> to <C>sysinfo.gap</C>
 (<URL><LinkText>&Hash;3248</LinkText><Link>https://github.com/gap-system/gap/pull/3248</Link></URL>).
 </Item>
+<Item>
+Make <C>MPtr</C> Julia type of &GAP; objects depend on abstract Julia <C>GapObj</C> type
```
fingolfin: "Depends" sounds wrong to me. "Derive from" ? "Be a subtype of"
```suggestion
Change <C>MPtr</C> Julia type of &GAP; objects to be a subtype of the abstract Julia <C>GapObj</C> type
```

fingolfin: "Depends" sounds wrong to me. "Derive from" ? "Be a subtype of"
```suggestion
Change <C>MPtr</C> Julia type of &GAP; objects to be a subtype of the abstract Julia <C>GapObj</C> type
provided by the Julia package <C>GAPTypes.jl</C>
```

fingolfin: "Depends" sounds wrong to me. "Derive from" ? "Be a subtype of"
```suggestion
Change <C>MPtr</C> Julia type of &GAP; objects to be a subtype of the abstract Julia <C>GapObj</C> type
provided by the Julia package <C>GAPTypes.jl</C>
```

configure.ac
sebasguts: Yep.

sebasguts:
Addressed the last comments. Feel free to merge it.

fingolfin:
Backported in commit 212556f947991e6ee707183d5f0c062d70f28d25

fingolfin:
We forgot to add this to the release notes for 4.10.2, but it's not so important, so I decided to omit this from the 4.11 release notes, too.

PR #3500, opened 2019-06-13
Two small cleanups

ChrisJefferson:
* Use `cat` instead of `rev` in a test, as `rev` sometimes crashes (due to a bug in some versions of `rev`).

* Remove two members of GAPInfo which have never been used since their introduction


ChrisJefferson:
Sorry, made an edit, but now (hopefully) figured out why this test keeps failing (I managed to reproduce the failure in a VM, it is documented behaviour)

PR #3502, opened 2019-06-14
HPC-GAP: Allow configuration of TLS and improve native TLS performance on macOS

rbehrends:
This is a tracking PR for progress on HPC-GAP changes. Aside from older changes in [this PR](https://github.com/gap-system/gap/pull/2845), we also aim at improving our implementations of thread-local storage, specifically so that HPC-GAP can be used as a library.

Our current implementation requires HPC-GAP to be run as the main program, which in particular prevents its use as a library. Alternative portable implementations are considerably slower on systems that do not properly support efficient thread-local storage (such as macOS and CygWin).

We're attempting to address this problem in two ways.

One issue is implementations that rely on `pthread_getspecific()` directly or indirectly do not treat such function calls as pure functions and thus do not get used for common subexpression elimination or hoisted out of loops. As those systems all rely on gcc/clang, we can tag `pthread_getspecific()` as `__attribute__((pure))`, which will immediately lead to a measurable performance improvement.

On macOS, we can go one step further. While there is no guaranteed stable ABI, `pthread_getspecific()` has consistently been implemented by `movq %gs:(,%rdi, 8), %rax`. We therefore can verify that `pthread_getspecific()` actually uses this implementation (`memcmp()` against the machine code) and then use a direct assembly language implementation that has nearly the same performance characteristics as our current implementation.

In addition, we store the key in a static variable that is initialized with an `__attribute__((constructor)__` function. This is generally slightly faster than putting it in a non-static variable on macOS in particular, as we're saving one level of indirection.

Also, in the worst case, this code functions with just `pthread_getspecific()` and no other mechanism available, it is just slower.

Todo:
* [x] Configure check for `__attribute__((pure))` and definition of a macro for it.
* [x] Configure check for `__attribute__((constructor))` and definition of a macro for it.
* [x] Configure check for it being acceptable to define `__pthread_getspecific()` as pure.
* [x] Configure check for macOS supporting this particular implementation.
* [x] Potentially distinguish between building HPC-GAP as a library and a standalone program?
* [ ] Efficient Cygwin implementation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -113,6 +113,8 @@ AX_GCC_FUNC_ATTRIBUTE([always_inline])
 AX_GCC_FUNC_ATTRIBUTE([format])
 AX_GCC_FUNC_ATTRIBUTE([noinline])
 AX_GCC_FUNC_ATTRIBUTE([noreturn])
+AX_GCC_FUNC_ATTRIBUTE([constructor])
+AX_GCC_FUNC_ATTRIBUTE([pure])
 
 dnl compiler builtins
 AC_DEFUN([CHECK_COMPILER_BUILTIN],
@@ -201,6 +203,135 @@ AS_CASE([$with_gc],
     [*],        [AC_MSG_ERROR([Invalid gc method '$with_gc', use default|none|gasman|boehm|julia])],
     )
 AC_MSG_RESULT([$with_gc])
+
+dnl
+dnl User setting: native thread-local storage (off by default)
+dnl See src/hpc/tls.h for more details on thread-local storage options.
+dnl
+
+AC_ARG_ENABLE([native-tls],
+    [AS_HELP_STRING([--enable-native-tls],
+        [use native thread-local storage implementation])],
+        [enable_native_tls=$enableval],
+        [enable_native_tls=no])
+AC_MSG_CHECKING([whether to use native tls])
+dnl
+dnl Distinguish between cases where we support __thread declarations
+dnl and situations where we use a pthread_getspecific() implementation.
+dnl Right now, we only do the latter for 64-bit macOS. See src/hpc/tls.h
+dnl for details.
+dnl
+enable_macos_tls_asm=default
+AS_IF([[test "x$enable_native_tls" == "xyes"]], [
+        case "$host" in
+            x86_64-apple-darwin*)
+                AC_DEFINE([USE_PTHREAD_TLS], [1], [define as 1 if using pthread_getspecific])
+                dnl
+                dnl Test if we can optimize pthread_getspecific() calls via
+                dnl inline assembly on macOS.
+                dnl
+                AC_RUN_IFELSE(
+                    [AC_LANG_SOURCE([[
+// The following code also occurs in src/hpc/thread.c and both need to be
+// kept in sync.
+#include <pthread.h>
+#include <string.h>
+
+#define OFFS 0x100
+#define END (-1)
+
+int cmpOpCode(unsigned char *code, int *with) {
+    int result = 0;
+    while (*with >= 0) {
+        if (*with == OFFS) {
+            result = *code;
+        } else {
+            if (*code != *with)
+                return -1;
+        }
+        code++;
+        with++;
+    }
+    return result;
+}
+
+int main() {
+    // This is an idea borrowed from Mono. We test if the implementation
+    // of pthread_getspecific() uses the assembly code below. If that is
+    // true, we can replace calls to pthread_getspecific() with the
+    // matching inline assembly, allowing a significant performance boost.
+#if defined(__APPLE__) && defined(__x86_64__)
+    // There are two possible implementations.
+    static int asm_code[] = {
+        // movq %gs:[OFFS](,%rdi,8), %rax
+        // retq
+        0x65, 0x48, 0x8b, 0x04, 0xfd, OFFS, 0x00, 0x00, 0x00, 0xc3, END
+    };
+    static int asm_code2[] = {
+        // pushq  %rbp
+        // movq   %rsp, %rbp
+        // movq   %gs:[OFFS](,%rdi,8),%rax
+        // popq   %rbp
+        // retq
+        0x55, 0x48, 0x89, 0xe5, 0x65, 0x48, 0x8b, 0x04, 0xfd, OFFS,
+        0x00, 0x00, 0x00, 0x5d, 0xc3, END
+    };
+    if (cmpOpCode((unsigned char *)pthread_getspecific, asm_code) >= 0) {
+        return 0;
+    }
+    if (cmpOpCode((unsigned char *)pthread_getspecific, asm_code2) >= 0) {
+        return 0;
+    }
+    return 1;
+#else
+#error FAIL
+#endif
+}
+                    ]])],
+                    dnl => Test succeeded
+                    [AC_DEFINE([USE_MACOS_PTHREAD_TLS_ASM], [1],
+                        [define as 1 if macOS assembly implementation supported])
+                    AC_MSG_RESULT([[yes (macOS assembly)]])
+                    enable_macos_tls_asm=yes
+                    ],
+                    dnl => Test failed
+                    [AC_MSG_RESULT([[yes (pthread-based)]])
+                    enable_macos_tls_asm=no
+                    ],
+                    dnl => Cross-compilation, test impossible
+                    [AC_MSG_RESULT([[yes (pthread-based)]])
+                    enable_macos_tls_asm=no
+                    ])
+            ;;
+            *)
+                dnl => Any other OS, --with-native-tls specified
+                AC_DEFINE([USE_NATIVE_TLS], [1], [define as 1 if using native TLS])
+                AC_MSG_RESULT([yes])
+            ;;
+        esac
+    ],
+    [
+        dnl => --without-native-tls
+        AC_MSG_RESULT([no])
+    ]
+)
+dnl
+dnl Test if pthread_getspecific() can be overridden as
+dnl __attribute__((pure)).
+dnl
+AC_MSG_CHECKING([[whether pthread_getspecific() can be made pure]])
+AC_COMPILE_IFELSE(
+    [AC_LANG_SOURCE([[
+#include <pthread.h>
+__attribute__((pure))
+void * pthread_getspecific(pthread_key_t key);
+]])], [
+    AC_MSG_RESULT(yes)
+    AC_DEFINE([ALLOW_PURE_PTHREAD_GETSPECIFIC], [1],
+        [define as 1 if pthread_getspecific() can be declared pure])
+], [
+    AC_MSG_RESULT(no)
+])
 
 dnl
 dnl User setting: Debug mode (off by default)
--- a/src/hpc/guards.h
+++ b/src/hpc/guards.h
@@ -104,7 +104,6 @@ static ALWAYS_INLINE Bag ImpliedReadGuard(Bag bag)
   return bag;
 }
 
-
 static ALWAYS_INLINE int CheckReadAccess(Bag bag)
 {
   Region *region;
--- a/src/hpc/thread.c
+++ b/src/hpc/thread.c
@@ -103,10 +103,95 @@ void UnlockThreadControl(void)
 #define MAP_ANONYMOUS MAP_ANON
 #endif
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
+
+#ifdef USE_PTHREAD_TLS
+static int InitTLSKey;
+static pthread_key_t TLSKey;
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+static UInt TLSOffset;
+
+// The following code also occurs in configure.ac, and both need to be
+// kept in sync.
+
+#define OFFS 0x100
+#define END (-1)
+
+int cmpOpCode(unsigned char *code, int *with) {
+    int result = 0;
+    while (*with >= 0) {
+        if (*with == OFFS) {
+            result = *code;
+        } else {
+            if (*code != *with)
+                return -1;
+        }
+        code++;
+        with++;
+    }
+    return result;
+}
+
+void FindTLSOffset() {
+    // This is an idea borrowed from Mono. We test if the implementation
+    // of pthread_getspecific() uses the assembly code below. If that is
+    // true, we can replace calls to pthread_getspecific() with the
+    // matching inline assembly, allowing a significant performance boost.
+    // There are two possible implementations.
+    static int asm_code[] = {
+        // movq %gs:[OFFS](,%rdi,8), %rax
+        // retq
+        0x65, 0x48, 0x8b, 0x04, 0xfd, OFFS, 0x00, 0x00, 0x00, 0xc3, END
+    };
+    static int asm_code2[] = {
+        // pushq  %rbp
+        // movq   %rsp, %rbp
+        // movq   %gs:[OFFS](,%rdi,8),%rax
+        // popq   %rbp
+        // retq
+        0x55, 0x48, 0x89, 0xe5, 0x65, 0x48, 0x8b, 0x04, 0xfd, OFFS,
+        0x00, 0x00, 0x00, 0x5d, 0xc3, END
+    };
+    TLSOffset = cmpOpCode((unsigned char *)pthread_getspecific, asm_code);
+    if (TLSOffset >= 0)
+        return;
+    TLSOffset = cmpOpCode((unsigned char *)pthread_getspecific, asm_code2);
+    if (TLSOffset >= 0)
+        return;
+    Panic("Unable to find macOS thread-local storage offset");
+}
+#endif
+
+static void CreateTLSKey(void)
+{
+    pthread_key_create(&TLSKey, NULL);
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+    FindTLSOffset();
+#endif
+    InitTLSKey = 1;
+}
+
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+UInt GetTLSOffset(void)
+{
+    if (!InitTLSKey) {
+        CreateTLSKey();
+    }
+    return (UInt)TLSKey * sizeof(void *) + TLSOffset;
+}
+#endif
+pthread_key_t GetTLSKey(void)
+{
+    if (!InitTLSKey) {
+        CreateTLSKey();
+    }
+    return TLSKey;
+}
+#endif /* USE_PTHREAD_TLS */
 
 void * AllocateTLS(void)
 {
+#ifndef USE_PTHREAD_TLS
     void * addr;
     void * result;
     size_t pagesize = getpagesize();
@@ -126,6 +211,14 @@ void * AllocateTLS(void)
     mprotect((char *)result + tlssize, pagesize, PROT_NONE);
 #endif
     return result;
+#else
+    void * result = pthread_getspecific(GetTLSKey());
+    if (!result) {
+        result = malloc(sizeof(GAPState));
+        pthread_setspecific(GetTLSKey(), result);
+    }
+    return result;
+#endif /* USE_PTHREAD_TLS */
 }
 
 void FreeTLS(void * address)
@@ -137,7 +230,7 @@ void FreeTLS(void * address)
 #endif
 }
 
-#endif /* HAVE_NATIVE_TLS */
+#endif /* USE_NATIVE_TLS */
 
 #ifndef DISABLE_GC
 void AddGCRoots(void)
@@ -157,7 +250,7 @@ static void RemoveGCRoots(void)
 }
 #endif /* DISABLE_GC */
 
-#ifndef HAVE_NATIVE_TLS
+#if !defined(USE_NATIVE_TLS) && !defined(USE_PTHREAD_TLS)
 
 /* In order to safely use thread-local memory on the main stack, we have
  * to work around an idiosyncracy in some virtual memory systems. These
@@ -191,7 +284,7 @@ static NOINLINE void GrowStack(void)
 
 static NOINLINE void SetupTLS(void)
 {
-#ifndef HAVE_NATIVE_TLS
+#if !defined(USE_NATIVE_TLS) && !defined(USE_PTHREAD_TLS)
     GrowStack();
 #endif
     InitializeTLS();
@@ -212,7 +305,7 @@ void RunThreadedMain(int (*mainFunction)(int, char **),
                      int     argc,
                      char ** argv)
 {
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
 #ifdef STACK_GROWS_UP
 #error Upward growing stack not yet supported
 #else
@@ -355,9 +448,8 @@ static void * DispatchThread(void * arg)
 Obj RunThread(void (*start)(void *), void * arg)
 {
     ThreadData * result;
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
     void * tls;
-    size_t         pagesize = getpagesize();
 #endif
     pthread_attr_t thread_attr;
     LockThreadControl(1);
@@ -370,7 +462,7 @@ Obj RunThread(void (*start)(void *), void * arg)
     }
     result = thread_free_list;
     thread_free_list = thread_free_list->next;
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
     if (!result->tls)
         result->tls = AllocateTLS();
     tls = result->tls;
@@ -395,7 +487,8 @@ Obj RunThread(void (*start)(void *), void * arg)
     result->thread_object = NewThreadObject(result - thread_data);
     /* set up the thread attribute to support a custom stack in our TLS */
     pthread_attr_init(&thread_attr);
-#ifndef HAVE_NATIVE_TLS
+#if !defined(USE_NATIVE_TLS) && !defined(USE_PTHREAD_TLS)
+    size_t         pagesize = getpagesize();
     pthread_attr_setstack(&thread_attr, (char *)tls + pagesize * 2,
                           TLS_SIZE - pagesize * 2);
 #endif
@@ -411,7 +504,7 @@ Obj RunThread(void (*start)(void *), void * arg)
         thread_free_list = result;
         UnlockThreadControl();
         pthread_attr_destroy(&thread_attr);
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
         FreeTLS(tls);
 #endif
         return (Obj)0;
@@ -424,7 +517,7 @@ int JoinThread(int id)
 {
     pthread_t pthread_id;
     void (*start)(void *);
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
     void * tls;
 #endif
     if (id < 0 || id >= MAX_THREADS)
@@ -432,7 +525,7 @@ int JoinThread(int id)
     LockThreadControl(1);
     pthread_id = thread_data[id].pthread_id;
     start = thread_data[id].start;
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
     tls = thread_data[id].tls;
 #endif
     if (thread_data[id].joined || start == NULL) {
@@ -451,7 +544,7 @@ int JoinThread(int id)
     */
     thread_data[id].start = NULL;
     UnlockThreadControl();
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
     FreeTLS(tls);
 #endif
     return 1;
--- a/src/hpc/tls.c
+++ b/src/hpc/tls.c
@@ -9,8 +9,9 @@
 */
 
 #include "gapstate.h"
+#include "hpc/thread.h"
 
-#ifdef HAVE_NATIVE_TLS
+#ifdef USE_NATIVE_TLS
 
 __thread GAPState TLSGAPtate;
 __thread ThreadLocalStorage *TLSInstance;
@@ -19,9 +20,12 @@ __thread ThreadLocalStorage *TLSInstance;
 
 void InitializeTLS(void)
 {
-#ifdef HAVE_NATIVE_TLS
+#ifdef USE_NATIVE_TLS
     // FIXME: is this the right place to do this?
     TLSInstance = &(TLSGAPtate.tls);
+#endif
+#ifdef USE_PTHREAD_TLS
+    AllocateTLS();
 #endif
     memset(ActiveGAPState(), 0, sizeof(GAPState));
 }
--- a/src/hpc/tls.h
+++ b/src/hpc/tls.h
@@ -17,6 +17,49 @@
 #error This header is only meant to be used with HPC-GAP
 #endif
 
+// This contains our implementation of thread-local storage (TLS). TLS
+// is a feature that is only to varying degrees supported by compilers
+// and operating systems. As it is critical for the performance of
+// HPC-GAP, we allow the implementation to choose between various
+// options.
+//
+// By default, we use a low-level stack-based approach. Each thread
+// stack is allocated on a memory segment whose boundaries are a
+// multiple of a fixed power of 2 (see tlsconfig.h). The pages with
+// the lowest address in that memory range are reserved for TLS,
+// the rest for the stack itself. This allows us to get the base
+// address for TLS by taking the stack or frame pointer (or any address
+// on the stack) and masking out the lowest bits. This is very fast
+// and on many platforms actually faster than global variables.
+//
+// However, this option only works when GAP is run as an application,
+// as it needs to control the creation of threads in order to force
+// thread stacks on the required boundaries. When used as a library,
+// we need a native implementation. These can be chosen by passing
+// the --enable-native-tls option to the configure script.
+//
+// By default, native TLS support will default to declaring TLS as
+// storage type __thread. This is simple, but inefficient on some
+// platforms. In particular, Cygwin and macOS require a function call
+// to a dynamically loaded library to access __thread storage, which
+// furthermore inhibits compiler optimizations.
+//
+// For these platforms, we use a different approach. As a basic
+// alternative, on platforms that support it, we use
+// pthread_getspecific() to get the base address of TLS and we
+// define pthread_getspecific() as a pure function. This is still
+// not faster than __thread by itself, but declaring
+// pthread_getspecific() as pure allows compiler optimizations such
+// as common subexpression elimination and code hoisting.
+//
+// As a further improvement, on macOS (and in the future, on Cygwin),
+// we use an assembly-based alternative implementation of
+// pthread_getspecific(). This assembly code can be treated like
+// a normal variable access. In order to check that this is safe,
+// we verify both during configuration and at runtime that the
+// assembly code for pthread_getspecific() is implemented in a way
+// consistent with such code.
+
 #include "hpc/tlsconfig.h"
 
 typedef struct Region Region;
@@ -55,28 +98,97 @@ typedef struct ThreadLocalStorage
   UInt LocksContended;
 } ThreadLocalStorage;
 
-#ifdef HAVE_NATIVE_TLS
+#ifdef USE_NATIVE_TLS
 extern __thread ThreadLocalStorage *TLSInstance;
 #endif
 
-#if defined(VERBOSE_GUARDS)
-static inline ThreadLocalStorage *GetTLS(void)
+#ifdef USE_PTHREAD_TLS
+
+#include <pthread.h>
+
+#ifdef ALLOW_PURE_PTHREAD_GETSPECIFIC
+// pthread_getspecific() is not defined as pure by default; redeclaring
+// it as such works for gcc/clang and allows the compiler to hoist calls
+// to pthread_getspecific() out of loops and perform constant
+// subexpression elimination on expressions containing TLS values.
+PURE_FUNC void * pthread_getspecific(pthread_key_t);
+#endif
+
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+static UInt TLSOffsetCopy;
+UInt GetTLSOffset(void);
 #else
-static ALWAYS_INLINE ThreadLocalStorage *GetTLS(void)
+static pthread_key_t TLSKeyCopy;
+pthread_key_t GetTLSKey(void);
+#endif
+
+// This constructor function will initialize a static variable to
+// address thread-local storage via pthread_getspecific() for each C
+// module (TLSOffset or TLSKeyCopy, depending on how thread-local
+// storage is being addressed). Making the variable static (instead of
+// having it global and initialized once) allows optimizations that a
+// global non-static variable would not permit. On some architectures,
+// it is also faster to access a static than a global non-static
+// variable, avoiding one level of indirection.
+//
+// Constructor functions are executed upon program start/library load.
+#ifdef HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+CONSTRUCTOR_FUNC static void InitTLSOffset()
+{
+    TLSOffsetCopy = GetTLSOffset();
+}
+#else
+CONSTRUCTOR_FUNC static void InitTLSKey()
+{
+    TLSKeyCopy = GetTLSKey();
+}
+#endif // USE_MACOS_PTHREAD_TLS_ASM
+#endif // HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR
+#endif // USE_PTHREAD_TLS
+
+#if defined(VERBOSE_GUARDS)
+static inline ThreadLocalStorage * GetTLS(void)
+#else
+static ALWAYS_INLINE ThreadLocalStorage * GetTLS(void)
 #endif
 {
-#ifdef HAVE_NATIVE_TLS
+#if defined(USE_NATIVE_TLS)
     return TLSInstance;
+#elif defined(USE_PTHREAD_TLS)
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+    long   offset;
+    void * p;
+    offset = (long)TLSOffsetCopy;
+    // The following inline assembly code is the same as that in
+    // pthread_getspecific(). That pthread_getspecific() is actually
+    // implemented in that way has been verified in the configure
+    // script and resulted in defining USE_MACOS_PTHREAD_TLS_ASM.
+    // If that test fails, we fall back to the standard implementation
+    // of pthread_getspecific().
+    asm("movq %%gs:(%1), %0"
+        : "=r"(p)  /* output = %0 */
+        : "r"(offset) /* input = %1 */
+        : /* clobber = none */);
+    return (ThreadLocalStorage *)p;
 #else
-    void *stack;
-  #ifdef __GNUC__
+#ifdef HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR
+    return (ThreadLocalStorage *)pthread_getspecific(TLSKeyCopy);
+#else
+    return (ThreadLocalStorage *)pthread_getspecific(GetTLSKey());
+#endif
+#endif /* USE_MACOS_PTHREAD_TLS_ASM */
+#else
+    // use stack mask approach
+    void * stack;
+#ifdef __GNUC__
     stack = __builtin_frame_address(0);
-  #else
+#else
     int dummy[1];
     stack = dummy;
-  #endif
-    return (ThreadLocalStorage *) (((uintptr_t) stack) & TLS_MASK);
 #endif
+    return (ThreadLocalStorage *)(((uintptr_t)stack) & TLS_MASK);
+#endif /* USE_NATIVE_TLS, USE_PTHREAD_TLS */
 }
 
 // Convenience helper for accessing TLS members
--- a/src/hpc/tlsconfig.h
+++ b/src/hpc/tlsconfig.h
@@ -18,7 +18,7 @@
 #endif
 
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
 
 enum {
     TLS_SIZE = (sizeof(UInt) == 8) ? (1L << 20) : (1L << 18),
@@ -28,6 +28,6 @@ enum {
 GAP_STATIC_ASSERT((TLS_SIZE & (TLS_SIZE - 1)) == 0,
                   "TLS_SIZE must be a power of 2");
 
-#endif // HAVE_NATIVE_TLS
+#endif // USE_NATIVE_TLS
 
 #endif // GAP_TLSCONFIG_H
--- a/src/hpc/thread.h
+++ b/src/hpc/thread.h
@@ -34,7 +34,7 @@
 
 extern int PreThreadCreation;
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
 void *AllocateTLS(void);
 void FreeTLS(void *address);
 #endif
--- a/src/objfgelm.cc
+++ b/src/objfgelm.cc
@@ -120,7 +120,7 @@ Obj NewWord(Obj type, UInt npairs)
     ADDR_OBJ(word)[1] = INTOBJ_INT(npairs);
     SetTypeDatObj(word, type);
 #ifdef HPCGAP
-    MakeBagReadOnly(word);
+    MakeBagPublic(word);
 #endif
     return word;
 }
--- a/src/system.h
+++ b/src/system.h
@@ -119,6 +119,19 @@ enum {
 #define NORETURN
 #endif
 
+#ifdef HAVE_FUNC_ATTRIBUTE_PURE
+#define PURE_FUNC __attribute__((pure))
+#else
+#define PURE_FUNC
+#endif
+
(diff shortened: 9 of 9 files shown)


fingolfin:
My understanding is that this contains PR #2845, correct? If so, it really would be nice if we could get *that* PR finished and merged, before building new stuff atop of it...

rbehrends:
Strictly speaking, this PR does not actually depend on the other one, but is committed on top of that branch for convenience.

I am already working on cleaning up the other PR as part of my current work to make multi-threaded Oscar a possibility, so that we can move on from that.

configure.ac
```diff
@@ -202,6 +202,61 @@ AS_CASE([$with_gc],
...
+    [AS_HELP_STRING([--with-native-tls],
+        [use native thread-local storage implementation])],
+        [],
+        [with_native_tls=no])
+AC_MSG_CHECKING([whether to use native tls])
+AS_IF([test "x$with_native_tls" == "xyes"], [
+        AC_COMPILE_IFELSE(
+            [AC_LANG_SOURCE([[
+#if defined(__APPLE__) && defined(__x86_64__)
+#error FAIL
```
fingolfin: If you just want to check for macOS, then you can do it much simpler, as we do elsewhere in this file:
```
case "$host_os" in
  *darwin*)
    ...
    ;;
esac
```
Also, do you really want to limit this to x86_64 ? So x86_32, and in the future ARM, are OK?

configure.ac
```diff
@@ -202,6 +202,61 @@ AS_CASE([$with_gc],
...
+dnl
+dnl User setting: native thread-local storage (off by default)
+dnl
+
+AC_ARG_WITH([native-tls],
+    [AS_HELP_STRING([--with-native-tls],
+        [use native thread-local storage implementation])],
+        [],
+        [with_native_tls=no])
+AC_MSG_CHECKING([whether to use native tls])
```
fingolfin: "whether to use native tls... pthread" is an odd message. Perhaps this:
```suggestion
AC_MSG_CHECKING([which TLS implementation to use])
```

Then down below, replace `AC_MSG_RESULT([yes])` by `AC_MSG_RESULT([native])`

configure.ac
```diff
@@ -202,6 +202,61 @@ AS_CASE([$with_gc],
...
+        return 1;
+    }
+    return 0;
+#else
+#error FAIL
+#endif
+            ]])],
+            [AC_DEFINE([USE_MACOS_PTHREAD_TLS_ASM], [1],
+                [define as 1 if macOS assembly implementation supported])],
+            [],
```
fingolfin: I think there should be a `AC_MSG_RESULT` here -- with the other change I suggested above, it could e.g. say `AC_MSG_RESULT([macOS x86_64 assembly])` or something like that

configure.ac
```diff
@@ -202,6 +202,61 @@ AS_CASE([$with_gc],
...
+            [AC_MSG_NOTICE(["Don't support cross-compiling for HPC_GAP with native TLS"])])
```
fingolfin: Shouldn't this only be printed if cross compilation is being used? Also, the message is a bit backwards, how about this:
```suggestion
            [AC_MSG_NOTICE(["Cross compilation of HPC_GAP with native TLS is not supported"])])
```

But actually, if I read your code right, isn't this only executed if `with_native_tls` is *not* set to `yes`? I.e. we don't use "native" TLS?! Though then the macOS assembly check also seems misplaced?!? Maybe I skipped the wrong number of opening and closing brackets... Perhaps you could briefly summarize, in words, what the detection logic here is supposed to do, then I could make a concrete suggestion how to tweak the code to make it clearer.

etc/ci-prepare.sh
```diff
@@ -40,8 +40,22 @@ then
...
+  cd unward
+  ./configure CC=gcc CXX=g++ CFLAGS=-O2 CXXFLAGS=-O2
+  make
+  cd ..
+  unward/bin/unward --inplace src
+  # commit the result to prevent the docomp test from triggering
+  git commit -m "Unward" src
+fi
+
+# build GAP
```
fingolfin: By the way, is this PR supposed to supersede PR #2845? Or will that PR be updated as well?

src/hpc/thread.c
```diff
@@ -103,10 +103,24 @@ void UnlockThreadControl(void)
...
+static int init_tls_key = 0;
+static pthread_key_t TLSKey;
+
+pthread_key_t GetTLSKey(void) {
+  if (!init_tls_key) {
+    pthread_key_create(&TLSKey, NULL);
+    init_tls_key = 1;
+  }
+  return TLSKey;
+}
```
fingolfin: please `clang-format` resp. `git clang-format`

src/hpc/thread.c
```diff
@@ -191,7 +213,19 @@ static NOINLINE void GrowStack(void)
...
 static NOINLINE void SetupTLS(void)
 {
-#ifndef HAVE_NATIVE_TLS
+#if defined USE_MACOS_PTHREAD_TLS_ASM
+    static char asm_code[] = {
+        // movq %gs:(,%rdi,8), %rax
+        // retq
+        0x65, 0x48, 0x8b, 0x04, 0xfd, 0x00, 0x00, 0x00, 0x00, 0xc3
+    };
+    if (memcmp((void *)pthread_getspecific, asm_code, sizeof(asm_code)) != 0) {
```
fingolfin:
```suggestion
    if (memcmp(pthread_getspecific, asm_code, sizeof(asm_code)) != 0) {
```

src/hpc/thread.c
```diff
@@ -191,7 +213,19 @@ static NOINLINE void GrowStack(void)
...
+        fprintf(stderr, "internal error: %s\n",
+            "inlining pthread_getspecific() assembly code failed");
+        SyExit(1);
```
fingolfin: Please consider using `Panic` here

src/hpc/thread.c
```diff
@@ -191,7 +213,19 @@ static NOINLINE void GrowStack(void)
 
 static NOINLINE void SetupTLS(void)
 {
-#ifndef HAVE_NATIVE_TLS
+#if defined USE_MACOS_PTHREAD_TLS_ASM
```
fingolfin: So `USE_MACOS_PTHREAD_TLS_ASM` is not "native" TLS, but also not `USE_PTHREAD_TLS`? Yet it shares the name with the latter, and a lot of the code with the former (e.g. in `AllocateTLS`) ? This is rather confusing, isn't it? Or maybe I already got it wrong (which, however, would kinda underline my point ;-).

src/hpc/thread.c
```diff
@@ -1285,4 +1323,78 @@ void ReadGuardError(Obj o)
...
+        char * p = calls[i] + 1;
+        char * q = p;
+        while (*p) {
+            if (p[-1] == ' ' && p[0] == ' ')
+                p++;
+            else
+                *q++ = *p++;
+        }
+        *q++ = 0;
+        SET_ELM_PLIST(stack, i, MakeImmString(calls[i]));
```
fingolfin: This is missing a `CHANGED_BAG` call, isn't it? Of course for Boehm GC that doesn't matter, but in the future we might be using the Julia GC here, no? Then it would matter.

src/hpc/threadapi.c
```diff
@@ -2042,8 +2046,13 @@ MigrateObjects(int count, Obj * objects, Region * target, int retype)
                 return 0;
         }
     }
-    for (i = 0; i < count; i++)
-        SET_REGION(objects[i], target);
+    for (i = 0; i < count; i++) {
+        Obj obj = objects[i];
+        if (TNUM_OBJ(obj) == T_PREC) {
+            SortPRecRNam(obj, 0);
```
fingolfin: Please add a comment that explain what is going on here.

src/hpc/tls.c
```diff
@@ -19,9 +19,13 @@ __thread ThreadLocalStorage *TLSInstance;
 
 void InitializeTLS(void)
 {
-#ifdef HAVE_NATIVE_TLS
+#ifdef USE_NATIVE_TLS
     // FIXME: is this the right place to do this?
     TLSInstance = &(TLSGAPtate.tls);
+#endif
+#ifdef USE_PTHREAD_TLS
+    void *AllocateTLS(void);
```
fingolfin: No, just: no! Don't declare external references inline like that. Add them properly to a header, or if such a header already exists (as in this case: `src/hpc/thread.h`), just #include that header.

src/hpc/tls.h
```diff
@@ -55,19 +55,40 @@ typedef struct ThreadLocalStorage
...
+// out of loops and perform constant subexpression elimination on
+// expressions containing TLS values.
+__attribute__((pure)) void * pthread_getspecific(pthread_key_t);
+#endif
+
+static pthread_key_t TLSKeyCopy;
+pthread_key_t GetTLSKey(void);
+
+__attribute__((constructor))
+static void InitTLSKey() {
```
fingolfin: Please clang-format

src/hpc/tls.h
```diff
@@ -55,19 +55,40 @@ typedef struct ThreadLocalStorage
...
+#ifdef USE_NATIVE_TLS
 extern __thread ThreadLocalStorage *TLSInstance;
 #endif
 
+#ifdef USE_PTHREAD_TLS
+
+#include <pthread.h>
+
+#ifdef __GNUC__
+// This trick allows gcc/clang to hoist calls to pthread_getspecific()
```
fingolfin: Which "trick" does this refer to? Let's be explicit! Is it the use of `__attribute__((pure)) `? Then let's say that: "Using __attribute__((pure)) here is a trick that allows ..."

src/hpc/tls.h
```diff
@@ -55,19 +55,40 @@ typedef struct ThreadLocalStorage
...
+// out of loops and perform constant subexpression elimination on
+// expressions containing TLS values.
+__attribute__((pure)) void * pthread_getspecific(pthread_key_t);
+#endif
+
+static pthread_key_t TLSKeyCopy;
+pthread_key_t GetTLSKey(void);
+
+__attribute__((constructor))
```
fingolfin: Why is this attribute here (needed) ? Let's either add a comment explaining that, or get rid of it.

src/hpc/tls.h
```diff
@@ -55,19 +55,40 @@ typedef struct ThreadLocalStorage
...
 static inline ThreadLocalStorage *GetTLS(void)
 #else
 static ALWAYS_INLINE ThreadLocalStorage *GetTLS(void)
 #endif
 {
-#ifdef HAVE_NATIVE_TLS
+#ifdef USE_NATIVE_TLS
     return TLSInstance;
 #else
+#ifndef USE_PTHREAD_TLS
```
fingolfin: I'd prefer to use non-negated conditions; that's of course subjective, but I find it easier to understand the code flow here. Also, here we could actually avoid the nesting, and do this instead:
```C
#if defined(USE_NATIVE_TLS)
 ...
#elif defined(USE_MACOS_PTHREAD_TLS_ASM)
...
#elif defined(USE_PTHREAD_TLS)
...
#else
#error no support TLS method detected
#endif
```

src/hpc/tlsconfig.h
```diff
@@ -18,7 +18,7 @@
 #endif
 
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USEE_NATIVE_TLS
```
fingolfin:
```suggestion
#ifndef USE_NATIVE_TLS
```

src/hpc/tlsconfig.h
```diff
@@ -28,6 +28,6 @@ enum {
 GAP_STATIC_ASSERT((TLS_SIZE & (TLS_SIZE - 1)) == 0,
                   "TLS_SIZE must be a power of 2");
 
-#endif // HAVE_NATIVE_TLS
+#endif // USEE_NATIVE_TLS
```
fingolfin:
```suggestion
#endif // USE_NATIVE_TLS
```

configure.ac
rbehrends: I'm actually special-casing this for 64-bit Darwin, yes. That is intended.

configure.ac
rbehrends: Sounds good.

configure.ac
rbehrends: Done.

configure.ac
rbehrends: This is the branch of `AC_RUN_IFELSE()` that is taken if cross-compiling is being done, in which case we can't do the test (it's the third branch after the "then" and "else" branches). That said, I've modified the output to simply add a message result rather than dissecting it further for now.

configure.ac
```diff
@@ -202,6 +202,61 @@ AS_CASE([$with_gc],
...
+#include <pthread.h>
+#include <string.h>
+]],[[
+#if defined(__APPLE__) && defined(__x86_64__)
+    static char asm_code[] = {
+        // movq %gs:(,%rdi,8), %rax
+        // retq
+        0x65, 0x48, 0x8b, 0x04, 0xfd, 0x00, 0x00, 0x00, 0x00, 0xc3
+    };
+    if (memcmp((void *)pthread_getspecific, asm_code, sizeof(asm_code)) != 0) {
```
ChrisJefferson: Woo, there needs to be a MUCH bigger comment around here. I would never guess what this was doing.

What is the improvement in performance got from this? I really don't like the idea of doing this super low-level thing, particularly becuase it is super-fragile to future changes to pthread_getspecific, even if they produce an equivalent future.

etc/ci-prepare.sh
rbehrends: Right now it exists on top of the other PR, but yes, I expect to just turn both into one next week.

src/hpc/thread.c
rbehrends: Done.

src/hpc/thread.c
rbehrends: That code is actually now unnecessary that it's been moved to the configure script. That said, I remember it giving me a warning so I would rather leave the cast in there.

src/hpc/thread.c
rbehrends: See above, I'm going to remove it entirely.

src/hpc/thread.c
rbehrends: Correct, but also gone now.

src/hpc/thread.c
rbehrends: Fixed.

src/hpc/threadapi.c
rbehrends: Added, this is about unsorted records being sorted if they are moved to a region where they cannot be mutated (or only if you have write access).

src/hpc/tls.c
rbehrends: Oops, fixed.

src/hpc/tls.h
rbehrends: Done. I'll probably also add checks for `__attribute__((constructor))` and `__attribute__((pure))` to the configure script rather than relying on ifdefs.

src/hpc/tls.h
rbehrends: Reworded.

src/hpc/tls.h
rbehrends: Explanation added. Briefly, the constructor attribute results in execution of the code when the program is started/library is loaded in order to initialize the static variable.

src/hpc/tls.h
rbehrends: Restructured that, though the logic looks different. `USE_MACOS_PTHREAD_TLS_ASM` has to be nested under the `USE_PTHREAD_TLS` case.

src/hpc/tlsconfig.h
rbehrends: Oops, fixed.

src/hpc/tlsconfig.h
rbehrends: Ditto.

configure.ac
rbehrends: I moved this to the configure script, where it's getting its own comment.

The improvement in performance is pretty significant, runs around 20%-30% slower without or so, I seem to recall. Note that this is basically taken from the mono (open source .NET compiler/runtime) implementation (not the code, just the idea). It's a very macOS-specific hack, because Apple refuses to define that part of the ABI.

But this is also why it needs to go to the configure script in order to avoid potential fragility. I'm actually thinking about adding a couple of additional sanity checks there to ensure that it works as advertised.

ChrisJefferson: A comment linking to Mono might be a good idea, particularly because if this breaks here, the Mono people might well fix it and we can take their fix :)

rbehrends: Yes, I added a reference to that. Note that it "breaking" means falling back to the slower portable version, so it would be a performance regression.

configure.ac
fingolfin: Still shouldn't use a compilation check for that, just match the host string against `x86_64-apple-darwin`

configure.ac
fingolfin: Adding a comment line like `dnl the following checks are only executed when cross compiling` would also help a lot in making this code more understandable.

src/hpc/thread.c
fingolfin: It might give a warning in C++, but never in C.

src/hpc/tls.h
fingolfin: You should be able to use AX_GCC_FUNC_ATTRIBUTE which we already use for other attributes.

.gitignore
```diff
@@ -90,9 +90,16 @@ doc/gapmacrodoc.idx
...
-/tst/testlibgap/wsload.out
+/tst/testlibgap/*.out
+
+/tst/testkernel/dstruct
+/tst/testkernel/*.out
+
+# emacs backup and lock files
+*~
+\#*\#
+.\#*
```
fingolfin: Huh... I think you ought to rebase on lastest master now, you are several months behind.

configure.ac
```diff
@@ -202,6 +202,67 @@ AS_CASE([$with_gc],
...
+#else
+#error FAIL
+#endif
+            ]])],
+            [AC_DEFINE([USE_MACOS_PTHREAD_TLS_ASM], [1],
+                [define as 1 if macOS assembly implementation supported])
+            AC_MSG_RESULT([[yes (macOS assembly)]])
+            ],
+            [AC_MSG_RESULT([[yes (pthread-based)]])],
+            [AC_MSG_NOTICE([[yes (pthread-based)]])])
```
fingolfin: This is confusing to read, perhaps partially because of the deep nesting... but why is there both a AC_MSG_RESULT  and a AC_MSG_NOTICE here?

src/hpc/threadapi.c
```diff
@@ -2042,8 +2046,15 @@ MigrateObjects(int count, Obj * objects, Region * target, int retype)
                 return 0;
         }
     }
-    for (i = 0; i < count; i++)
-        SET_REGION(objects[i], target);
+    // If we're migrating records to a region where they become immutable,
```
fingolfin:
```suggestion
    // If we are migrating records to a region where they become immutable,
```

src/precord.c
```diff
@@ -44,6 +44,7 @@
 #ifdef HPCGAP
 #include "hpc/aobjects.h"
 #include "hpc/traverse.h"
+#include "hpc/guards.h"
```
fingolfin: need to be moved up one line (clang-format sorts headers alphabetically)

configure.ac
rbehrends: That makes sense. I'll look at that.

configure.ac
rbehrends: Good point. I'll probably do another pass to break down the logic more cleanly, as there's a lot of nesting going on.

.gitignore
rbehrends: Huh, I thought I had copied that from a freshly pulled master, because I was getting merge issues (which is what the last commit is all about). Weird, I'll have to look at it again.

configure.ac
rbehrends: I think auto-complete tripped me up, should indeed be `AC_MSG_RESULT()` also.

src/precord.c
rbehrends: I have to do that for the headers alone, if I reformat all of `src/precord.c`, then it adds a lot of spurious changes (as the file does not currently match the clang-format style).

src/hpc/threadapi.c
rbehrends: Done.

configure.ac
fingolfin: @ChrisJefferson Reimer added a comment, please take a look to see if that is good enough for you :-)

.gitignore
fingolfin: This stuff is definitely in master. But it is not in the old HPC-GAP PR on which you based this PR; that one was last updated in June. So a `git rebase master` should get rid of the above.

configure.ac
ChrisJefferson: Could you just point to where mono does this? A link to some place in the code would be good.

ChrisJefferson: My general worry is, in the event none of the people in this PR are around when this needs fixing (which I'm sure it will, when new version of Mac OS X come out), then fixing it back up again will be a major undertaking.

Also, should we maybe have some way of noticing this is failing? (I suppose we can hope someone notices the configure test is failing, or notices the slowdown).

rbehrends: You can find the matching code in mono [here](https://github.com/mono/mono/blob/6bc988b0870b7d4d56a73c13b45fe5158f81c64d/mono/mini/mini-amd64.c#L4125). The ABI has remained unchanged ever since Apple started using Intel processors, so I don't expect this to change from version to version.

Actually, it looks like while the basic code hasn't changed in forever, it's slightly different if the executable is launched from the XCode GUI. While this is unlikely to happen here, I might just as well add a check for it, too (the semantics of the two functions are necessarily the same).

And yes, it would probably be a good idea to have a way to notice when it's failing. Just not sure how to keep it from being drowned in configure spam.

rbehrends:
This is now a revised version that mirrors the Mono logic. It can also now be more easily extended to equally support  Windows implementations.

configure.ac
```diff
@@ -202,6 +204,114 @@ AS_CASE([$with_gc],
     )
 AC_MSG_RESULT([$with_gc])
 
+dnl
+dnl User setting: native thread-local storage (off by default)
```
fingolfin: Not saying it should be different, but just to understand better: why is this off by default? What's the pro/con for turning it on (or off) by default?

configure.ac
```diff
@@ -202,6 +204,114 @@ AS_CASE([$with_gc],
...
+        [with_native_tls=no])
+AC_MSG_CHECKING([whether to use native tls])
+dnl
+dnl Distinguish between cases where we support __thread declarations
+dnl and situations where we use a pthread_getspecific() implementation.
+dnl Right now, we only do the latter for 64-bit macOS.
+dnl
+enable_macos_tls_asm=default
+AS_IF([[test "x$with_native_tls" == "xyes"]], [
+        case "$host_cpu-$host_vendor-$host_os" in
```
fingolfin:
```suggestion
        case "$host" in
```

configure.ac
```diff
@@ -202,6 +204,114 @@ AS_CASE([$with_gc],
...
+AC_ARG_WITH([native-tls],
+    [AS_HELP_STRING([--with-native-tls],
+        [use native thread-local storage implementation])],
+        [],
+        [with_native_tls=no])
+AC_MSG_CHECKING([whether to use native tls])
+dnl
+dnl Distinguish between cases where we support __thread declarations
+dnl and situations where we use a pthread_getspecific() implementation.
+dnl Right now, we only do the latter for 64-bit macOS.
```
fingolfin: This is better than what was there before, but still lacks information about the reasons: *why* don't we use `__thread` on macOS, resp. conversely: why do we  prefer it elsewhere over `pthread_getspecific()`.

All this should be explained once and in total in some places; other places then can refer to it: ".. for details on the reasons for this, please look at file XYZ"

configure.ac
```diff
@@ -202,6 +204,114 @@ AS_CASE([$with_gc],
...
+    if (cmpOpCode((unsigned char *)pthread_getspecific, asm_code2) >= 0) {
+        return 0;
+    }
+    return 1;
+#else
+#error FAIL
+#endif
+}
+                    ]])],
+                    dnl => Test succeeded
```
fingolfin: thanks those comments are really helpful!

etc/ci-prepare.sh
```diff
@@ -40,8 +40,22 @@ then
...
+if [[ $HPCGAP = yes ]]
+then
+  git clone https://github.com/rbehrends/unward
+  cd unward
+  ./configure CC=gcc CXX=g++ CFLAGS=-O2 CXXFLAGS=-O2
+  make
+  cd ..
+  unward/bin/unward --inplace src
+  # commit the result to prevent the docomp test from triggering
+  git commit -m "Unward" src
```
fingolfin: So this means people who want to use the code in this PR still need to have unward, and run it, and then have lots of modifications in their tree. That's IMHO *the* major blocker for merging this PR, and until we resolve it, there is no way this can be merged. So I see two ways forward:

1. We run `unward` and commit the modifications once and for all, and move on (my understanding always was that this is the plan?)
2. We determine that we will have to keep using `unward` for the foreseeable. In that case, we will have to use it like `ward`, and thus need to restore the parts of the build system which transparently handled applying `ward` (now: `unward`) to sources followed by compilation (instead of working inlaces, the results would be placed into `gen/src/`; and build rules would ensure `gen/src/FOO.c` gets rebuilt whenever `src/FOO.c` or one of its header dependencies is touched.

Thing is, we need people to use this code so it gets tested, so keeping it in a PR indefinitely is not a good way forward.

src/gasman.h
```diff
@@ -293,13 +298,59 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr) {
 **  the application  must inform {\Gasman}  that it  has changed  the bag, by
 **  calling 'CHANGED_BAG(old)' in the above example (see "CHANGED_BAG").
 */
+#ifdef HPCGAP
+EXPORT_INLINE PURE_FUNC int ReadCheck(Bag bag)
```
fingolfin: I really think we need to add an explanation here as to why we declare this function as a `PURE_FUNC`, and why this is "legal". That explanation should be written in one place, and other places using PURE_FUNC (at least in this header) should then refer to that: "For more information about the use of PURE_FUNC here, please read the documentation comment for ReadCheck" or something like that.

src/gasman.h
```diff
@@ -293,13 +298,59 @@ EXPORT_INLINE UInt SIZE_BAG_CONTENTS(const void *ptr) {
...
+#endif // HPCGAP
+
 EXPORT_INLINE Bag *PTR_BAG(Bag bag)
+{
+    GAP_ASSERT(bag != 0);
+#ifdef HPCGAP
+    if (!WriteCheck(bag))
+      GuardDummy = HandleWriteGuardError(bag);
+#endif
+    return *(Bag**)bag;
```
fingolfin:
```suggestion
    return *(Bag **)bag;
```

`git clang-format master` ?

configure.ac
rbehrends: Yes, that's actually what I'm shooting for. I'm looking at putting a full explanation into tls.h, and then refer to that.

configure.ac
rbehrends: Performance and portability. The stack-based TLS approach is the fastest *and* the most portable one, but it cannot be used with arbitrary thread stacks (i.e. when using it GAP as a library). One could make a good argument that it should be the other way round; in either case, there should be an explanation.

etc/ci-prepare.sh
rbehrends: So, this is something that I don't have a good solution for myself. My plan was to not put the unward changes in for a transition period. The reason is that if we need to change something about how they work (e.g. include an option for disabling guards for performance reasons), they mess up diffs and git blame, as they're pretty extensive.

But that leads to the problem that if you build HPC-GAP without unward, you get a near immediate crash, as an unnecessary read or write guard fails (one that gets removed by unward).

Possible solution: add a `--without-guards` configure option. That allows builds to (unsafely) work without unward for the transition period *and* would also later allow HPC-GAP to be built without guards for performance reasons if performance is critical.

If we were to put unward into the build system, it already supports a `-o <dir>` option to place the transformed sources (and all untransformed ones, too) into a separate directory, say `gen/`. This takes less than a second, is incremental (it doesn't overwrite files that didn't get changed and therefore doesn't change their time stamps) and thus make-friendly, and is in principle possible.

But this would again introduce an external dependency, which I'm not very enthusiastic about (even though unward, unlike ward, is itself dependency-free and only requires a C++-98 or later compiler). A large part of the raison d'être for unward is, after all, that it can be dropped from the build system eventually and does not continue to be a maintenance burden.

src/gasman.h
rbehrends: Good point. Will do that.

etc/ci-prepare.sh
fingolfin: This sounds as if you are aiming for option 1 eventually. But the big question then is: when would we be ready for it? What's the hard criterium for making that decision? I keep coming back to this very same questions since when you first started talking about unward. You keep saying it's meant to be used once, but also keep saying that it may need to be changed, and then re-run. Unless there is a clear goal as to when it's "done", I don't see a solution out of this dilemma, other than to accept that the notion of "unward will be run once and then discarded" is hopeless. Hence option 2.

At the same time, it's also totally unclear to me what major changes to unward output you think might be needed?

src/hpc/tls.h
```diff
@@ -55,28 +55,96 @@ typedef struct ThreadLocalStorage
...
+        : /* clobber = none */);
+    return (ThreadLocalStorage *)p;
+#else
+#ifdef __GNUC__
+    return (ThreadLocalStorage *)pthread_getspecific(TLSKeyCopy);
 #else
-    void *stack;
-  #ifdef __GNUC__
+    return (ThreadLocalStorage *)pthread_getspecific(GetTLSKey());
+#endif
```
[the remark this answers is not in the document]
rbehrends: This is explained in a comment further up. That said, I may actually want to change that to a configure test to be absolutely on the safe side.

src/gasman.h
rbehrends: I'm not sure I understand. This is the same format as it currently exists on the master branch?

etc/ci-prepare.sh
rbehrends: There are two possible changes that I'm contemplating right now and that may require changes in the output:

1. Allow disabling of guards at compile time.
2. Add `#ifdef HPCGAP` around the inline functions, so that they don't get compiled when not built with HPC-GAP. This was unnecessary when they were still static, but now they're not static anymore and should not show up in the code regardless.

My goal with unward was to keep the choice open-ended. This is why I specifically supported both the following options:

1. Make the process idempotent, so that you can rerun unward if necessary.
2. Add a `-o` option to allow for use as a preprocessor in the build system (i.e. like ward).

This means that we wouldn't be forced by design to pick one or the other.

As for the process, I was hoping to basically get the following done:

1. Add the current PR (cleaned up) to GAP.
2. Ask a couple of people to test that it works with unward (there aren't many who are using HPC-GAP, but there are a couple at least).
3. If that's the case, then after a short delay (couple of weeks) add the unward changes in a separate patch.

I am totally fine with rolling them into the current PR, too, if you think that's the best way forward. In the end, you probably have a better view of the big picture as far as GAP is concerned.

src/gasman.h
fingolfin: The old code is indeed not (yet) consistently clang-formatted; but you are adding new code here, which you may as well clang-format. Anyway, this is hardly important, feel free to ignore :-)

src/gasman.h
rbehrends: I've actually removed the `PURE_FUNC` stuff. It does not seem to make any performance difference anymore.

fingolfin:
@rbehrends the commit here is titled "Make HPC-GAP work without Ward." but that's not really what this is about, given that we already are not using ward...

src/hpc/tls.h
```diff
@@ -17,6 +17,49 @@
...
+// storage type __thread. This is simple, but inefficient on some
+// platforms. In particular, Cygwin and macOS require a function call
+// to a dynamically loaded library to access __thread storage, which
+// furthermore inhibits compiler optimizations.
+//
+// For these platforms, we use a different approach. As a basic
+// alternative, on platforms that support it, we use
+// pthread_getspecific() to get the base address of TLS and we
+// define pthread_getspecific() as a pure function. This is still
+// not faster than __thread by itsefl, but declaring
```
fingolfin:
```suggestion
// not faster than __thread by itself, but declaring
```

src/hpc/tls.h
```diff
@@ -55,28 +98,96 @@ typedef struct ThreadLocalStorage
...
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+static UInt TLSOffset;
+UInt GetTLSOffset(void);
+#else
+static pthread_key_t TLSKeyCopy;
+pthread_key_t GetTLSKey(void);
+#endif
+
+// This constructor function will initialize the static variable for
+// each C module. Making the variable static allows optimizations that
```
fingolfin: Which variable is "the variable" which is being referred to here?

src/hpc/tls.h
```diff
@@ -55,28 +98,96 @@ typedef struct ThreadLocalStorage
...
+// variable, avoiding one level of indirection.
+//
+// Constructor functions are executed upon program start/library load.
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+CONSTRUCTOR_FUNC static void InitTLSOffset()
+{
+    TLSOffset = GetTLSOffset();
+}
+#elif defined(HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR)
+CONSTRUCTOR_FUNC static void InitTLSKey()
```
fingolfin: You already used `CONSTRUCTOR_FUNC` above, so I think the `#ifdef` logic needs to be changed a bit:
```C
#ifdef HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR
  #ifdef USE_MACOS_PTHREAD_TLS_ASM
  ...
  #else
  ...
  #endif
#endif
```

src/hpc/tls.h
```diff
@@ -55,28 +98,96 @@ typedef struct ThreadLocalStorage
...
+{
+    TLSKeyCopy = GetTLSKey();
+}
+#endif
+#endif // USE_PTHREAD_TLS
+
+#if defined(__APPLE__) && defined(__x86_64__)
+#define PTHREAD_TLS_ASM
+#endif
```
fingolfin: As far as I can tell, `PTHREAD_TLS_ASM` is not used, and the above three lines can be removed?

configure.ac
```diff
@@ -202,6 +204,149 @@ AS_CASE([$with_gc],
...
+	    result = *code;
+	} else {
+	    if (*code != *with)
+	        return -1;
+	}
+	code++;
+	with++;
+    }
+    return result;
+}
```
fingolfin: Something is off in the indentation of this function, at least when rendered on GitHub. Maybe a matter of tabs-vs-spaces? Should only use spaces here.

src/hpc/guards.h
```diff
@@ -11,109 +11,73 @@
...
-      !region->readers[TLS(threadID)] && region->alt_owner != GetTLS())
-    ReadGuardError(bag
-#ifdef VERBOSE_GUARDS
-    , file, line, func, expr
-#endif
-    );
-  return bag;
 }
 
-
```
fingolfin: Why are you inserting an empty new line before `ImpliedWriteGuard`, but removing one before `ImpliedReadGuard`? I mean, I could understand if one wanted to tweak things so that always the same number of empty lines is between functions (i.e., always 1 or always 2); but instead, you seem to just switching the inconsisten spacing around?

src/hpc/thread.c
```diff
@@ -103,10 +103,93 @@ void UnlockThreadControl(void)
...
+    }
+    return result;
+}
+
+void FindTLSOffset() {
+    // This is an idea borrowed from Mono. We test if the implementation
+    // of pthread_getspecific() uses the assembly code below. If that is
+    // true, we can replace calls to pthread_getspecific() with the
+    // matching inline assembly, allowing a significant performance boost.
+    // There are two possible implementations.
```
fingolfin: This code also appears in `configure.ac`, and I think it would be wise to always keep the two copies in sync. Ideal would be if there was really only one copy, but pulling that off might be more of a hassle than it's worth it, so let's forget about that.

Next best is to add a comment that reminds the reader about this. Something like:  "This check exists in src/hpc/thread.c and configure.ac, and the two copies should always be kept in sync", added in both places.

src/hpc/thread.c
```diff
@@ -103,10 +103,93 @@ void UnlockThreadControl(void)
 #define MAP_ANONYMOUS MAP_ANON
 #endif
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
+
+#ifdef USE_PTHREAD_TLS
+static int init_tls_key = 0;
```
fingolfin: Global vars (even static ones) should start with an upper case letter, and use CamelCase. Also, no need to explicitly init to 0, the standard mandates that for globals anyway.
```suggestion
static int InitTLSKey;
```

src/hpc/thread.c
```diff
@@ -103,10 +103,93 @@ void UnlockThreadControl(void)
...
+static pthread_key_t TLSKey;
+#ifdef USE_MACOS_PTHREAD_TLS_ASM
+static int tls_offset;
```
fingolfin:
```suggestion
static int TLSOffset;
```

src/hpc/thread.c
```diff
@@ -103,10 +103,93 @@ void UnlockThreadControl(void)
...
+    init_tls_key = 1;
+}
+
+pthread_key_t GetTLSKey(void)
+{
+    if (!init_tls_key) {
+        CreateTLSKey();
+    }
+    return TLSKey;
+}
```
fingolfin: I believe this function should be inside an `#ifndef USE_MACOS_PTHREAD_TLS_ASM`; resp, move it into an `#else` block after the implementation of `GetTLSOffset` (that would match the structure in the header file)

etc/ci-prepare.sh
fingolfin: @rbehrends (CC: @stevelinton @ChrisJefferson ) regarding your first two points, it seems you already added `--enable-guards` / `--disable-guards` (which is perfectly fine by me). Note sure around which inline functions you want(ed?) to add `#ifdef HPCGAP`, but I guess it's fine, too.


Regarding how to proceed: My main concern about merging this PR is how it would affect people who are using HPC-GAP right now, but who do not yet know about unward and thus are not using it. Will they get a horribly broken HPC-GAP, or just one that's more or less on the same level as what we have in `master` right now anyway?

If it was the latter, there is indeed nothing major stopping this from being merged (my remaining remarks are all very minor nitpicks and should be easy to resolve).

However, it seems to be the former (starting GAP built from your branch w/o running unward causes an immediate segfault due to violated guards). That makes me somewhat more reserved about the idea of merging this as-is. It'd be still OK by me if there was a very concrete timeline for it, and also a plan B should we determined that result of the merge is unusable. But my main concern is who the "couple of people" would be we'd ask to test it, and what makes us think they'd be more likely to test it then, than they would be to test this PR / your branch if we asked them explicitly to test it?

Until we come up with a good strategy to deal with that, it might indeed be best to split this commit into two: a non-controversial commit we could merge right away (which would contain the TLS changes, the renaming of `HAVE_NATIVE_TLS` to `USE_NATIVE_TLS` and a bunch of other minor changes), and the rest (mainly: the guards changes).

rbehrends:
I'll address the remaining comments tomorrow, but about how to go forward with a merge: we could flip the default for `--disable-guards` around to have guards disabled by default; this would preserve the status quo and you would have to `./configure --enable-hpcgap --enable-guards` and run `unward` to get guards semantics. Once we merge the unward changes, we could then enable guards by default.

src/hpc/tls.h
rbehrends: Done.

src/hpc/tls.h
rbehrends: Elaborated on that.

src/hpc/tls.h
rbehrends: Good catch. This is actually a problem with code that can only possibly be compiled on non-gcc/clang compilers in that it's difficult to test for that eventuality (especially as icc and others also include the gcc/clang extensions).

src/hpc/tls.h
rbehrends: Yes, that's a leftover from before.

configure.ac
rbehrends: Yes, some tabs snuck in from a copy and paste, I think. Fixed.

src/hpc/guards.h
rbehrends: This was by accident, I think. Before squashing the commits, there was code that got deleted, apparently in one case without combining the lines. Fixed.

src/hpc/thread.c
rbehrends: I added comments in both places.

src/hpc/thread.c
rbehrends: Fixed the case. However, I've left the initialization, since it documents that this variable has a default value.

src/hpc/thread.c
rbehrends: Also changed.

src/hpc/thread.c
rbehrends: You are correct. Fixed.

src/hpc/thread.c
```diff
@@ -103,10 +103,96 @@ void UnlockThreadControl(void)
 #define MAP_ANONYMOUS MAP_ANON
 #endif
 
-#ifndef HAVE_NATIVE_TLS
+#ifndef USE_NATIVE_TLS
+
+#ifdef USE_PTHREAD_TLS
+static int InitTLSKey = 0;
```
fingolfin: @rbehrends I honestly don't see how inserting `= 0` "documents that this variable has a default value" (quoted from a previous comment). As I stated back then, all global values default to 0, so this assignment is superfluous. If you indeed want to "document" something here, I would suggest to actually, well, "document" it in the form of a comment that says whatever you want to say.

fingolfin:
The suggestion to flip the default for `--disable-guards` for now sounds good to me

configure.ac
```diff
@@ -202,6 +204,151 @@ AS_CASE([$with_gc],
     )
 AC_MSG_RESULT([$with_gc])
 
+dnl
+dnl User setting: disable guards (race protection)
+dnl
+
+AC_ARG_ENABLE([guards],
+    [AS_HELP_STRING([--enable-guards],
+        [disable race protection in HPC-GAP])],
```
fingolfin: Hmm, why don't the name of the option and the description not match? I'd expect it either to be called `--enable-race-protection`, or else a description which actually matches the option.
Also, the option says "--enable-guard", but the description talks about **disable** ?

 How about:
```suggestion
        [disable guards for race protection in HPC-GAP])],
```
or if the verb was wrong, then
```suggestion
        [enable guards for race protection in HPC-GAP])],
```

configure.ac
```diff
@@ -202,6 +204,151 @@ AS_CASE([$with_gc],
...
+    [enable_guards=$enableval],
+    [enable_guards=no])
+
+AS_IF([[test "x$enable_guards" == "xyes"]], [
+    AC_DEFINE([USE_HPC_GUARDS], [1], [define as 1 if guards should be checked])
+])
+AC_MSG_CHECKING([whether to use race protection in HPC-GAP])
```
fingolfin: Again: the option is `--enable-guards`; the `AC_DEFINE` says "guards should be checked". So I'd refer to "guards" here, too, so that when reading the message in a configure log, it immediately is clear what it relates to.
```suggestion
AC_MSG_CHECKING([whether to use guards for race protection in HPC-GAP])
```

configure.ac
rbehrends: Oops, I flipped the semantics, but not the description. Fixed.

configure.ac
rbehrends: Also fixed.

src/hpc/thread.c
rbehrends: Okay, this is no big deal for me either way. I've changed it.

fingolfin:
Still has guards failures here:
```
gs:=[ExtraspecialGroup( 27, 3 ), DihedralGroup(32)];
G:=DirectProduct(gs);
A:=AutomorphismGroup(G);
```
gives me
```
Error, Attempt to read object 4666793776 of type plain list without having read access in
  AddDictionary( dict, img, Length( orb ) ); at GAPROOT/lib/grpmat.gi:306 called from
DoSparseLinearActionOnFaithfulSubset( grp, OnRight, sort ) at GAPROOT/lib/grpmat.gi:494 called from
NicomorphismOfGeneralMatrixGroup( G, false, false ) at GAPROOT/lib/grpmat.gi:535 called from
...
```
Using `lldb` to figure out where that plist comes from revealed that it's `cacheBag`; specifically, the failure is here:
```
   1874
   1875	#ifdef HPCGAP
   1876	    // reset the method cache if necessary
-> 1877	    if (ELM_PLIST(cacheBag, 1) != methods) {
   1878	        Obj * cache = BASE_PTR_PLIST(cacheBag);
   1879	        cache[0] = methods;
   1880	        memset(cache + 1, 0, SIZE_OBJ(cacheBag)-2*sizeof(Obj));
```

The method cache is per-thread, so the access should be fine; but I think we never set a region on those lists, hence the access error. @rbehrends can you take a look?

rbehrends:
@fingolfin Yes, I'm working on that. There are a couple more (and oddly enough, one I see only locally, not on Travis).

fingolfin:
There is still an error in the same place as before (the `AutomorphismGroup` example I pasted above), but now for another reason; it only happens if `autpgrp` is loaded. When this happens, for some reason, we run into this code inside the GAP library, in `lib/pquot.gi:505`:
```gap
    wt := qs!.collector![SCP_WEIGHTS];
    wt{n+[1..nhwg]} := [1..nhwg] * 0 + class;  # <- guard is violated here
```
The guard is violated because `wt` is in the read-only region. I have not yet figured out why it is read-only (as I am about to change trains now, so no time)

fingolfin:
(This also means there is an easy workaround, but we should first better understand what's going on, so we don't break multi threaded stuff)
```patch
diff --git a/lib/pquot.gi b/lib/pquot.gi
index e30e26d0c..6c39cf5c8 100644
--- a/lib/pquot.gi
+++ b/lib/pquot.gi
@@ -502,8 +502,9 @@ UpdateWeightInfo := function( qs )
     class := class + 1;
     qs!.collector![SCP_CLASS] := class;
 
-    wt := qs!.collector![SCP_WEIGHTS];
+    wt := ShallowCopy(qs!.collector![SCP_WEIGHTS]);
     wt{n+[1..nhwg]} := [1..nhwg] * 0 + class;
+    qs!.collector![SCP_WEIGHTS] := MakeImmutable(wt);
 
     avc2 := [1..n]+0;
     for g in [1..n] do
```

rbehrends:
There is also a problem with the FactInt package in `tst/testinstall/hpc/tasks.tst` in lines 35 and 37. The calls to `CallAsTask( NF, 7, [ 1 ] )` and `CallAsTask( NF, 7, [ 1, 2 ] )` error out because the function `IntegerFactorization` in `FactInt-1.6.2/lib/general.gi` accesses a global cache (`FACTINT_SMALLINTCACHE`), which is created in the main thread as thread-local data.

fingolfin:
@rbehrends some optimization opportunities:
-   `CURR_FUNC` accesses only data and objects which are thread local, so perhaps we can turn off ward in it?
- likewise, `NARG_FUNC`, `CALL_0ARGS`, etc. all shouldn't need ward (and hence no UNSAFE variants) because function and function bodies are essentially immutable after being created. Well, with some caveats:
  - for `T_FUNCTION`, all members of `struct FuncBag` with exception of the `name` are constant after creation; for operations, a few of the additional members (such as the methods cache, and the `extra` field) are not, but that's easy to take care of.
  - for `T_BODY`, the content is constant *except* that the profiling code wants to set bits in statements/expressions once they are executed for the first time. But surely we can take care of that with an atomic operation?

My suggestion  here would be to change T_FUNCTION and T_BODY to *not* be in the public region by default; instead, we create them thread local, and then once we are done setting them up, change them to the readonly region (resp. for T_BODY, we might even be ok marking them as read-only, with the caveat that atomically setting those profiling bits must still work)

rbehrends:
This is a general optimization opportunity for all objects that are immutable. Currently, we have these tagged via `MakeBagTypePublic()` and that means that any read or write guard check exits early via the fast path. However, even the fast path incurs *some* overhead, so we should definitely excise the unnecessary checks.

I would, however, propose to postpone such optimizations until everything is working and integrated, unless you think there are good reasons to include them in this PR?

fingolfin:
No, there is no need to include those in this PR.

BTW, there are more reasons for this than optimization: also safety / correctness. Right now, we create a lot of bags in the public region, then modify them extensively. This is OK because those objects are not visible to the outside while we are modifying them; and then later on, we never modify them (or only in a "safe") fashion. Or so we claim, but there is no good way to actually check that. But if we instead created those objects thread local, and only made them public once the setup is complete, then we could (at least in principle) forbid `ADDR_OBJ` / `PTR_BAG` for objects in the public region (raising a guard error or so), only allowing `CONST_ADDR_OBJ` / `CONST_PTR_BAG` and `UNSAFE_ADDR_OBJ` / `UNSAFE_PTR_BAG`. Of course the devil is in the details, but that would be the direction I'd try to follow.

fingolfin:
Reminder: once this PR is merged, we need to update https://github.com/gap-system/pkg-ci-scripts/blob/master/build_gap.sh so that `unward` is run when testing against HPC-GAP.

PR #3504, opened 2019-06-16
buildsys: teach package boostrap to use curl if available

### LINKED ISSUE #1866: Teach "make bootstrap" to use `curl` if `wget` is not available
On macOS, `curl` is available but `wget` is not.

fingolfin:
This allows people to do a package bootstrap on macOS without installing wget,
as curl is installed there by default but wget is not.

Resolves #1866


sebasguts:
looks good to me :)

PR #3505, opened 2019-06-17
Switch COPYING to use T_DATOBJ, as PLISTs should not contain arbitary values

ChrisJefferson:
Semigroups and digraphs (for example https://github.com/gap-packages/Semigroups/issues/542 ) have seen occasional warning when doing SaveWorkspace.

I have tracked this down to COPYING, which makes a PLIST with an arbitrary value as the first thing in the PLIST. This isn't legal, PLISTs should only contain valid GAP Objs.

This PR switches from using PLISTS to DATOBJs.

I don't think this could ever have caused an error, and is fairly hard to trigger, as these objects are usually GCed after copying finishes.

## Text for release notes 

* Fix occasional warning about bad bags when saving a workspace.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -425,13 +425,14 @@ void PrepareCopy(Obj obj, Obj copy)
     // between original and copy (e.g. for objects, they point to the type;
     // if making an immutable copy of a mutable object, the types will
     // differ).
+    // Store in a DATOBJ, as CONST_ADDR_OBJ(obj)[0] could be anything,
+    // including an arbitary C integer.
     // Likewise, the TNUM of the copy and the original can and will differ;
     // e.g. for a weak pointer list, the copy can be a plist.
-    Obj tmp = NEW_PLIST(T_PLIST, 3);
-    SET_LEN_PLIST(tmp, 3);
-    SET_ELM_PLIST(tmp, 1, CONST_ADDR_OBJ(obj)[0]);
-    SET_ELM_PLIST(tmp, 2, copy);
-    SET_ELM_PLIST(tmp, 3, INTOBJ_INT(TNUM_OBJ(obj)));
+    Obj tmp = NewBag(T_DATOBJ, 4 * sizeof(Obj *));
+    ADDR_OBJ(tmp)[1] = CONST_ADDR_OBJ(obj)[0];
+    ADDR_OBJ(tmp)[2] = copy;
+    ADDR_OBJ(tmp)[3] = INTOBJ_INT(TNUM_OBJ(obj));
 
     // insert the forwarding pointer
     GAP_ASSERT(SIZE_OBJ(obj) >= sizeof(Obj));
@@ -460,7 +461,7 @@ Obj COPY_OBJ(Obj obj, Int mut)
         Obj fpl = CONST_ADDR_OBJ(obj)[0];
 
         // return pointer to the copy
-        copy = ELM_PLIST(fpl, 2);
+        copy = ADDR_OBJ(fpl)[2];
     }
     else if (!IS_MUTABLE_OBJ(obj)) {
         copy = obj;
@@ -488,11 +489,11 @@ void CLEAN_OBJ(Obj obj)
     Obj fpl = CONST_ADDR_OBJ(obj)[0];
 
     // remove the forwarding pointer
-    ADDR_OBJ(obj)[0] = ELM_PLIST(fpl, 1);
+    ADDR_OBJ(obj)[0] = ADDR_OBJ(fpl)[1];
     CHANGED_BAG(obj);
 
     // restore the tnum
-    UInt tnum = INT_INTOBJ(ELM_PLIST(fpl, 3));
+    UInt tnum = INT_INTOBJ(ADDR_OBJ(fpl)[3]);
     RetypeBag(obj, tnum);
 
     // invoke type specific cleanup, if any
@@ -509,8 +510,12 @@ static void MarkCopyingSubBags(Obj obj)
     // mark the forwarding pointer
     MarkBag(fpl);
 
+    // Mark stored subobjects
+    MarkBag(ADDR_OBJ(fpl)[1]);
+    MarkBag(ADDR_OBJ(fpl)[2]);
+
     // mark the rest as in the non-copied case
-    UInt tnum = INT_INTOBJ(ELM_PLIST(fpl, 3));
+    UInt tnum = INT_INTOBJ(ADDR_OBJ(fpl)[3]);
     TabMarkFuncBags[tnum](obj);
 }
 


src/objects.c
```diff
@@ -425,13 +425,14 @@ void PrepareCopy(Obj obj, Obj copy)
...
     // Likewise, the TNUM of the copy and the original can and will differ;
     // e.g. for a weak pointer list, the copy can be a plist.
-    Obj tmp = NEW_PLIST(T_PLIST, 3);
-    SET_LEN_PLIST(tmp, 3);
-    SET_ELM_PLIST(tmp, 1, CONST_ADDR_OBJ(obj)[0]);
-    SET_ELM_PLIST(tmp, 2, copy);
-    SET_ELM_PLIST(tmp, 3, INTOBJ_INT(TNUM_OBJ(obj)));
+    Obj tmp = NewBag(T_DATOBJ, 4 * sizeof(Obj *));
+    ADDR_OBJ(tmp)[1] = CONST_ADDR_OBJ(obj)[0];
+    ADDR_OBJ(tmp)[2] = copy;
```
fingolfin: But this is the only reference to `copy` and GASMAN cannot see it here. So unless I miss something, this makes it worse...

src/objects.c
```diff
@@ -425,13 +425,14 @@ void PrepareCopy(Obj obj, Obj copy)
     // between original and copy (e.g. for objects, they point to the type;
     // if making an immutable copy of a mutable object, the types will
     // differ).
+    // Store in a DATOBJ, as CONST_ADDR_OBJ(obj)[0] could be anything,
+    // including an arbitary C integer.
     // Likewise, the TNUM of the copy and the original can and will differ;
     // e.g. for a weak pointer list, the copy can be a plist.
-    Obj tmp = NEW_PLIST(T_PLIST, 3);
-    SET_LEN_PLIST(tmp, 3);
-    SET_ELM_PLIST(tmp, 1, CONST_ADDR_OBJ(obj)[0]);
```
fingolfin: For which kind of object is this not a valid object reference?

ChrisJefferson: It's hard to tell exactly what object, as I'm just seeing the objects later. I think it's C_STRINGs

src/objects.c
ChrisJefferson: I actually have a different way of handling this, which might be better / safer.

ChrisJefferson: So the problem isn't GASMAN, the problem is SaveWorkspace, which walks along the plist and expects to find a valid GAP Obj at every location. GASMAN is actually happy with getting random pointers (as it has to be, as they might be found on the stack).

ChrisJefferson:
I'm thinking this is might not be a good idea, because now if we someone ended up doing SaveWorkspace with a partial COPY operation complete, we wouldn't save the objects I've put in the DATOBJ. Of course, that should never happen, but we shouldn't crash in that case.

ChrisJefferson:
Alternative Suggestion: Clear the PLISTs once copying is finished. This will be slightly slower, and still means we have "bad plists" during copying, but should stop warnings in SaveWorkspace.

src/objects.c
fingolfin: Of course GASMAN is fine with this in the current form in master. But I don't think it is OK with your T_DATOBJ variant, as no subbags / slots other than that at offset 0 are marked.

ChrisJefferson: I added this to MarkCopyingSubbags, which already special-cases the things it copies.

fingolfin: But  how does the work? If the only reference to `copy` is in that `T_DATOBJ`, then on a full GC, we never mark `copy`, and hence never invoke `MarkCopyingSubbags` on it -- or am I missing something?

src/objects.c
fingolfin: OK, indeed, for `T_STRING` the first word is the length of the string as a raw integer. I just went through all calls to `PrepareCopy` to check for each type using it what is going on:
* blist: first slot contains length as an intobj
* posobj, comobj, datobj: first slot contains ref to type object
* plist: first slot contains length as an intobj
* prec: first slot is always 0
* range: first slot contains length as an intobj
* string: first slot contains length as a **raw integer** (!!!)
* weak pointer object: first slot contains length as a **raw integer** (!!!)

So strings and wpobjs are the oddballs. I think for wpobj, it is an absolute nobrainer to switch to storing the length as an INTOBJ. This leaves strings. They also used to store the length as an intobj, but at some point this was changed (IIRC by Frank) to allow larger strings. This is only relevant on 32bit systems, though. Hence, at least on 64bit systems, we could change this to an intobj, too.

I'd thus suggest this fix: Leave this code here as it is. Change wpobj to use an intobj length. Change strings also to store the integer as an object, but with a special case for 32bit systems, where instead of `INTOBJ_INT` and `INT_INTOBJ`, we use the more general `ObjInt_UInt` resp. `UInt_ObjInt`.

wilfwilson:
From my point of view, I don't mind what the fix is, but just to make the motivation for this issue clear:

It would be great if this kind of thing:
```
########> Diff in save-workspace.tst:155
# Input is:
SaveWorkspace(Concatenation(SEMIGROUPS.PackageDir, "/tst/workspaces/test-output.w"));
# Expected output:
true
# But found:
#W bad bag id -1632527300 found, 0 saved
true
########
Semigroups package: workspaces/save-workspace.tst
msecs: 1220
false
```
didn't happen any more, and randomly make Travis jobs fail (note that it happens most often in 32-bit jobs). In particular, saving a workspace shouldn't involve reading raw integers from objects that haven't been garbage collected for whatever reason.

src/objects.c
ChrisJefferson: I don't think Frank changed them to be raw integers, I think he added the fact we stored the length at all, instead of assuming strings are always the same size as the bag they are contained in.

ChrisJefferson: I don't like the idea of using ObjInt_Uint, as that means that setting the length of a string could cause a GC, so we'd have to check everywhere in both the kernel, and packages, to make sure no-one is keeping a C pointer (which is a common thing to do when playing with strings) in such cases.

If we just switched to using INTOBJ_INT (and checked the maximum size of strings) that would limit us to 256MB strings on a 32-bit OS, which doesn't sound like an unreasonable requirement, but is possible to hit.

fingolfin: Here is an alternate suggestion: just add another argument to `PrepareCopy`, a boolean indicating whether slot 0 contains raw data or an object reference. Store that boolean in the plist, too; and then, depending on its value, store slots 0 in the plist either verbatim, as we do now (if the boolean indicates it is an object ref), or else using `ObjInt_UInt` (if the boolean indicates it is raw data).

ChrisJefferson: Thanks for the suggestions.

I'm actually going to go and do some investigations. I've been poking both strings and blists, and it turns out neither of them work properly if you actually get larger than a value that fits in an INTOBJ, for example if you can make a blist and string of size >= 2^28, then `l[2^28]` doesn't work, and I can easily get GAP to crash by asking various questions about these 2^28 length blists and strings.

ChrisJefferson: To be concrete, I think your original idea is best - - make the lengths intobjs, as larger values are currently borken and that is the easier and more stable fix, in my mind.

ChrisJefferson: To be concrete, I think your original idea is best - - make the lengths intobjs and add error checking for lengths, as larger values are currently borken anyway.

ChrisJefferson:
This problem is being fixed in a different way, but I'm keeping it here as I want to make sure I address all the comments raised.

fingolfin:
@ChrisJefferson can this be closed now?

ChrisJefferson:
Yep, I checked (by saving workspace after disabling GC) and we no long create any bags which SaveWorkspace complains about.

PR #3507, opened 2019-06-18
Skip test involving fork() on windows

### LINKED ISSUE #3506: teststandard/processes/children.tst unusable in 64-bit Windows build
This issue arises when GAP is build under cygwin and cygwin64, and then copied over to a cygwin-free machine to test it there. 

The cygwin-free machine is a Windows10 VM that can use 4 GB RAM. 

The test of the 32-bit build works there just fine. It takes 13-14 hours to run twice `testinstall.g` and twice `teststandard.g`, with `-A` and with default packages loaded. 

However, the test of the 32-bit build stuck in `teststandard/processes/children.tst` forever, and was terminated after 20 hours. 

```
    3594 ms (718 ms GC) and 419MB allocated for pperm.tst
testing: /proc/cygdrive/C/gap-4.10.2-x86_64/tst/teststandard/processes/childre\
n.tst
Build timed out (after 1,200 minutes). Marking the build as aborted.
Build was aborted
[WS-CLEANUP] Deleting project workspace...
[WS-CLEANUP] Deferred wipeout is used...
[WS-CLEANUP] done
Finished: ABORTED
```

I logged in and tried commands from the test interactively, and it works:
```
gap> d := DirectoryCurrent();;
gap> scriptdir := DirectoriesLibrary( "tst/teststandard/processes/" );;
gap> checkpl := Filename(scriptdir, "check.pl");;
gap> check.pl;
Error, Variable: 'check' must have a value
not in any function at *stdin*:6
gap> checkpl;
"/proc/cygdrive/C/gap-4.10.2-x86_64/tst/teststandard/processes/check.pl"
gap> if IsBoundGlobal("IO_RestoreSIGCHLDHandler") then
> ValueGlobal("IO_RestoreSIGCHLDHandler")();
> fi;
gap> runChild := function(ms, ignoresignals)
>    local signal;
>    if ignoresignals then signal := "1"; else signal := "0"; fi;
>    return InputOutputLocalProcess(d, checkpl, [ String(time), signal]);
>  end;;
gap> reps:=1;
1
gap> runChild(1,false);
< input/output stream to check.pl >
gap> runChild(1,true);
< input/output stream to check.pl >
gap> Perform([last,last2],CloseStream);
```

but then trying to call

    children := List([1..20], x -> runChild(Random([1..2000]), Random([false,true])));;

makes computer very slow to response. I started resource monitor and run a new attempt, and it seems like it starts heavy swapping:
 
cygwin64

Shall we have two versions of this test perhaps? A heavy one in `testextra`, which I do not run on Windows, and a lightweight (with just two `runChild` calls as above) in `teststandard` so that we do run this test on Windows to test that at least `InputOutputLocalProcess` works?

fingolfin:
This otherwise is far too slow on 64 bit windows.

Fixes #3506

### DIFF
--- a/tst/teststandard/processes/children.tst
+++ b/tst/teststandard/processes/children.tst
@@ -12,6 +12,7 @@ gap> runChild := function(ms, ignoresignals)
 >    return InputOutputLocalProcess(d, checkpl, [ String(time), signal]);
 >  end;;
 gap> reps:=200;;
+gap> if ARCH_IS_WINDOWS() then reps:=0; fi;  # FIXME: this test is too slow under windows
 gap> if IsHPCGAP then reps:=0; fi;  # FIXME: this test is broken in HPC-GAP
 gap> for i in [1..reps] do
 > children := List([1..20], x -> runChild(Random(1,2000), Random([false,true])));;


alex-konovalov:
Thanks, but this is not enough. As #3506 says, even the first iteration hangs it. You need also to tweak parameters in

    children := List([1..20], x -> runChild(Random([1..2000]), Random([false,true])));;

What worked for me was simple `runChild(1,false)` and `runChild(1,true)`.

fingolfin:
I've decided to simply don't run this part of the test file at all on Windows. It's IMHO just not worth the hassle to try to just launch 1-2 child processes: we already have other tests which do that, so we don't gain anything from it.

alex-konovalov:
Ok! I guess I still can test #3513 where this test is not modified?

fingolfin:
@alex-konovalov does your approval of this PR mean that you tested it on Windows / Cygwin and it resolved the issue there?

alex-konovalov:
No - it's pretty obvious that this test should be excluded from Windows, so my approval is just based on inspecting the diff.

fingolfin:
Argh, I confused the PRs, my bad :-(

alex-konovalov:
Backported to stable-4.10 in 2dbf56e42d6155a2f6da7169d19e0b9f1171e609

PR #3512, opened 2019-06-19
kernel: change STRING_LIST_DIR to return a plist

fingolfin:
This actually results in simpler kernel code, and means that GAP does not have to split a string into a list.

### DIFF
--- a/lib/files.gd
+++ b/lib/files.gd
@@ -299,7 +299,7 @@ DeclareOperation( "ReadAsFunction", [ IsString ] );
 ##  the directory <A>dir</A>. The argument <A>dir</A> can either be given as 
 ##  a string indicating the name of the directory or as a directory object
 ##  (see <Ref Filt="IsDirectory"/>).
-##  It is an error, if such a directory does not exist. 
+##  If the specified directory does not exist, <K>fail</K> is returned.
 ##  <P/>
 ##  The ordering of the list entries can depend on the operating system.
 ##  <P/>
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
It looks to me like Browse calls this function, and would get broken by this PR?

fingolfin:
@ChrisJefferson good catch *sigh*. Though I guess "you get what you ask for" -- calling undocumented internal functions always is likely to break things.

So it would be good if @frankluebeck and @ThomasBreuer could update Browse to not use `STRING_LIST_DIR`. It seems they cloned `DirectoryContents` to avoid an error output in it. I guess one substitute for that would be to change `DirectoryContents` to return fail for missing dirs; and/or add an optional argument to it to force the behavior change.

frankluebeck:
As far as I remember I first implemented 'STRING_LIST_DIR' as a substitute of an external call to 'find ... -print0' (private code). But the proposed change looks fine to me. I would:

  - just rename 'FuncSTRING_LIST_DIR' to 'FuncDirectoryContents'
  - remove the code of 'DirectoryContents' in the library
  - adjust the documentation of 'DirectoryContents' (to return 'fail' instead of raising an error, similar to, e.g., 'CrcFile' or 'RemoveFile')
  - maybe add a one-line compatibility function 'STRING_LIST_DIR' in 'lib/obsolete.g*' for a short time to avoid inconsistencies between the library and Browse

We will then change Browse to use 'DirectoryContents' directly.

fingolfin:
@frankluebeck that sounds like a good plan to me. I'll work on it

fingolfin:
PR now updated to more or less follow what @frankluebeck suggested. (I kept `DirectoryContents` as a GAP function, as that makes it easier to handle `IsDirectoryRep` inputs).

Issue #1889, opened 2017-11-10
Plan: add more "caching" abstractions


fingolfin:
## Motivation
We recently added `lib/cache.gi` and `lib/cache.gd`, currently with only a single function `MemoizePosIntFunction` in them.

Our plan is to add more functionality in there for common caching needs. One motivation is that this will help us to further reduce the number of files in `hpcgap/lib/`, as quite a few of them are in there simply because they are adding locking code to caches. If we replace the caching code in them by calls to generic caching solutions, we just have to provide HPC-GAP variants of these, but can keep the code using the caches uniform between GAP and HPC-GAP.

Another motivation is that writing an efficient cache can be tricky, and by providing some generally useful abstractions, we will encourage more people to use robust caching solutions in there code.

(I should mention that @sebasguts also worked on caches of a certain kind for the CAP project, and there is certainly some overlap between his work and what I describe here! So perhaps we can share code or collaborate in other ways, or at least benefit from his experiences).

In this issue, I want to make a list of types of caches I think we'll need, with different needs. To do so, I analyzed the diffs between `lib/` and `hpcgap/lib/`, which I think is quite representative. However, since I limited myself to these files, it is likely that there are additional places that use caching logic in the library which are not covered below.

## Weak caches vs. flushable caches

The first differentiation criterion for caches is what purpose they serve:
1. Some caches are needed to ensure multiple calls to a function with equal resp. identical inputs will result in *identical* outputs. Typical use cases are functions which compute types or families.
2. Other caches are meant to improve performance, by remembering the result of an expensive computation. GAP attributes and properties are examples of that, but also e.g. the `Bernoulli` function

In my experience, these two use cases seem to be fully disjoint. (**QUESTION**: Are they? Can anybody think of a counter example? Resp. a third category of caches not covered by these two?)
For the former, the cache contents are often (always?) cheap to recompute, and so it makes sense to keep the cached values into weak pointer lists, so that GAP can garbage collect them if they are not needed anymore. Summed up, the main feature of this type of cache is that it ensures **uniqueness** and may keep **weak** pointers.

For the latter kind of cache, one does not want to let go of the computed result, so weak pointers are usually (always?) not an option. Instead, one may use `InstallFlushableValue` to give the user an exit hatch if they need to recover memory (albeit a rather coarse one: either flushing everything, or nothing). Note that it's not an option to flush the values of the first type of cache. To summarize, these caches about **performance**, they trade memory for speed, and are **flushable**

[...]

PR #3515, opened 2019-06-21
Add GET_FROM_SORTED_CACHE helper; use it to unify hpcgap/lib/ffe.gi, hpcgap/lib/pcgsperm.gi, hpcgap/lib/zmodnz.gi

fingolfin:
Note that this changes how the undocumented variables `GALOIS_FIELDS` and `CYCLICACHE` work. But I verified that no package access them (as it should be). I also think that the new caching approach is not really worse than the old in both cases (the memory overhead is slightly larger for `CYCLICACHE`).

This could be considered a small step towards resolving issue #1889, but of course the primitive data structure employed here is not exactly ideal. But I decided that instead of waiting for a hypothetical better implementation that may never come, I wanted to have some progress now, so that for fixing issues in `zmodnz.gi`, I don't have to remember editing two files.


lib/cache.gi
```diff
@@ -70,3 +70,62 @@ function(func, extra...)
...
+    pos:= POSITION_SORTED_LIST( cache[1], key );
+    if pos <= Length( cache[1] ) and cache[1][pos] = key then
+      return cache[2][ pos ];
+    fi;
+
+    # Compute new value.
+    val := maker();
+
+    # Store the value.
+    Add( cache[1], key, pos );
```
ChrisJefferson: AddSet? (and similarly above, and one line below)?

PR #3516, opened 2019-06-21
Performance fix for the Julia GC integration.

rbehrends:
This is both a performance fix and an attempt to refactor part of the Julia GC integration.

On the performance side, this avoids rescanning unused task stacks unnecessarily. See [this related GAP.jl issue](https://github.com/oscar-system/GAP.jl/issues/266).

I've also attempted a migration of the GC integration to C++. The motivation for that was the need to have clean data types, which were messy to handle in C (especially the generics). This probably requires some testing.

Also, while GAP currently already uses some C++ code, it is not linked against the C++ stdlib, which means that things such as `new` and `delete` won't properly work. Therefore, the C++ is written to work without any such features and is primarily used for abstraction. Care was also taken to only require C++98 features (build with `CXXFLAGS="-g -O2 -std=c++98"` to verify).

There's currently a short test, build with `CXXFLAGS=-DTEST_JULIA_GC_INTERNALS` to run it instead of GAP. This is only temporary and meant to be removed before this PR is being merged.

Note that the C++ stdlib issues could be resolved by using `CXX` instead of `CC` for the `LINK` command in `Makefile.rules`, but I am not sure if that is desirable.


src/julia_gc.cc
```diff
@@ -138,30 +525,12 @@ static void JFinalizer(jl_value_t * obj)
...
-**  balanced trees (B-trees being a likely suitable candidate).
-*/
-
 // Comparing pointers in C without triggering undefined behavior
 // can be difficult. As the GC already assumes that the memory
 // range goes from 0 to 2^k-1 (region tables), we simply convert
 // to uintptr_t and compare those.
 
-static inline int cmp_ptr(void * p, void * q)
+int cmp_ptr(void * p, void * q)
```
fingolfin: Why?

rbehrends: That's a problem with C++98/C++03, which does not allow you to pass static functions as template parameters. Using them as template parameters allows inlining (as opposed to storing a function pointer and the overhead that comes with that). I'm not happy with it, either, and have been thinking about workarounds, such as using inheritance instead or classes that function as comparators. Or, alternatively, have a function with a more unique name.

rbehrends: I've done some code reorganization and put potentially conflicting symbols in a namespace of their own. And `cmp_ptr` is back to being a static inline function regardless.

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
 
-static treap_t * treap_free_list;
 
-treap_t * alloc_treap(void)
-{
-    treap_t * result;
-    if (treap_free_list) {
-        result = treap_free_list;
-        treap_free_list = treap_free_list->right;
+class StdAlloc {
```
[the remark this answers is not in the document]
fingolfin: That said, this class `StdAlloc` feels like over-engineering to me. Can't we just have the global `alloc` template function call `malloc`? Or even simpler, just overload operator `new` and `delete`?
```C++
void* operator new(std::size_t sz)
{
    void *result = std::malloc(sz);
    if (!result)
        abort();
}
void operator delete(void* ptr) noexcept
{
    std::free(ptr);
}
```

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
+static const UInt MaxTreeDepth = 2 * (sizeof(UInt) * 8);
+static int        height_to_size_init = 0;
+// For scapegoat trees with balance factor alpha:
+// height_to_size: d -> (1/alpha) ^ d
+static Int height_to_size[MaxTreeDepth];
+
+template <typename T, typename Cmp = T>
+class BalancedTree : public StdAlloc {
+    // A scapegoat tree. It is an easy to implement balanced
+    // tree, with amortized O(log n) insertion and deletion
```
fingolfin: Could you briefly explain why we need to replace the treap code with yet another new data structure?

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
+                w /= (double)alpha_lo;
+                height_to_size[d] = (Int)w;
+            }
+        }
+    }
+    void deinit()
+    {
+        remove_all();
+    }
+    void remove_all()
```
fingolfin:
```suggestion
    void clear()
```
to match STL containers?

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
+    }
+    T * find(T item)
+    {
+        return find_aux(_root, item);
+    }
+    void remove(T item)
+    {
+        remove_aux(_root, item);
+    }
+    Int depth_aux(Node * node)
```
fingolfin: Should this be a private / protected method?

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
+struct CmpAddr {
+    static int cmp(void *a, void *b) {
+        return cmp_ptr(a, b);
+    }
+};
+
+
+} // namespace JuliaGC
+
+using namespace JuliaGC;
```
fingolfin: Why do we need this local namespace if we then immediately do a `using` on it? This should be explained by a comment here.

src/julia_gc.cc
```diff
@@ -703,28 +988,33 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
...
     char * stack = (char *)jl_task_stack_buffer(task, &size, &tid);
+    // If it is the current task, it has been scanned by GapRootScanner()
+    // already.
+    if (task == JuliaTLS->current_task)
+        return;
+    int rescan = 1;
+    if (!FullGC) {
+        // This is a temp hack to work around a problem with the
+        // generational GC. Basically, task stacks are being scanned
+        // regardless of whether they are old or new. In order to avoid
```
fingolfin: Is this a bug in Julia, resp. something there we don't understand? If so, can you please provide a link to the relevant Julia issue report, resp. the relevant discussion on the Julia discourse?

src/julia_gc.cc
```diff
@@ -1030,7 +1322,7 @@ void SwapMasterPoint(Bag bag1, Bag bag2)
 
 // HOOK: mark functions
 
-inline void MarkBag(Bag bag)
+void MarkBag(Bag bag)
```
fingolfin: Why?

src/julia_gc.cc
```diff
@@ -1089,3 +1381,61 @@ void MarkJuliaWeakRef(void * p)
     if (JMarkTyped(p, jl_weakref_type))
         YoungRef++;
 }
+
+
+#ifdef TEST_JULIA_GC_INTERNALS
```
fingolfin: When are these tests run? I don't see them being hooked up anywhere.

src/julia_gc.cc
```diff
@@ -1089,3 +1381,61 @@ void MarkJuliaWeakRef(void * p)
...
+        if (i == N / 2 && btree->count() != N / 2)
+            abort();
+        btree->remove(j);
+    }
+    if (btree->count() != 0 || btree->depth() != 0)
+        abort();
+    btree->remove(0);
+    BalancedTree<int, IntCmp>::destroy(btree);
+    printf("Scapegoat tree passed tests.\n");
+    exit(0);
```
fingolfin: There are no tests for the `ArrayList`, it seems?

src/julia_gc.cc
```diff
@@ -1089,3 +1381,61 @@ void MarkJuliaWeakRef(void * p)
...
+        if (!btree->find(i))
+            abort();
+    }
+    for (int i = 1; i < N; i++) {
+        btree->remove(i);
+    }
+    if (btree->count() != 1 || btree->depth() != 1)
+        abort();
+    btree->remove(0);
+    for (int i = 0, j = 0; i < N; i++, j = (5 * j + 1) & (N - 1)) {
```
ChrisJefferson: These tests (to me) don't seem complete enough -- they don't test that after each insertion and removal the tree is correct, they don't try removing a value from the tree which isn't present, or adding a value more than once. They also only seem to test 'find' returning true.

src/julia_gc.cc
```diff
@@ -210,204 +126,471 @@ static inline void * align_ptr(void * p)
...
+#ifndef REQUIRE_PRECISE_MARKING
+
+#define MARK_CACHE_BITS 16
+#define MARK_CACHE_SIZE (1 << MARK_CACHE_BITS)
+
+#define MARK_HASH(x) (FibHash((x), MARK_CACHE_BITS))
+
+// The MarkCache exists to speed up the conservative tracing of
+// objects. While its performance benefit is minimal with the current
+// API functionality, it can significantly reduce overhead if a slower
```
ChrisJefferson: I don't want to sound bad, but if this provides very little benefit at present why add it, as it seems to be making things much more complicated.

src/julia_gc.cc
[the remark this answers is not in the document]
rbehrends: I tried custom `operator new` and `operator delete`, but that requires that the C++ runtime is present. This is currently not the case, because the code is linked with gcc instead of g++. If I could rely on the runtime, that would simplify things enormously, as I also wouldn't have to hack around missing constructor/destructor calls.

The existence of `StdAlloc` is in part a leftover from trying to have those operators, that I then reengineered to just encapsulate the allocation mechanism.

src/julia_gc.cc
rbehrends: The treap code technically relies on a random number generation; while it is very unlikely to fail, I prefer deterministic behavior. Scapegoat trees are the easiest to implement deterministic balanced tree. I could have used `std::set` and `std::map`, but they again rely on the missing C++ runtime, so I plugged in some code that I had already written for something else.

src/julia_gc.cc
rbehrends: Yes. Will fix.

src/julia_gc.cc
rbehrends: non-static inline functions can produce linker problems and they do in this case. We had a similar issue a while ago (with a hash implementation in another file), and as in practice, `inline` is just treated as a recommendation by optimizers, I removed the keyword.

src/julia_gc.cc
rbehrends: They are run as `__attribute__((constructor)` functions. See the PR description for how to pass the flag. This is a hack in order to be able to run some simple tests.

src/julia_gc.cc
rbehrends: No. I copied these tests over from my original implementation, which did not have any for `ArrayList`, which is a pretty barebones dynamic array.

src/julia_gc.cc
rbehrends: Yeah, the tests are incomplete. At the moment, I wasn't sure whether using C++ was even acceptable, so I didn't want to invest too much time without being certain that I wasn't going to waste it.

src/julia_gc.cc
rbehrends: Yeah, the performance benefits vary. When I originally wrote it, I got like a 5%-10% speedup in some cases (for overall code, not just the GC part), but it seems to be compiler/architecture dependent as to whether it's actually a win. I'm fine with stripping that out in a separate PR if there aren't any performance benefits on common setups.

src/julia_gc.cc
rbehrends: Basically, to not pollute the global linker namespace with generic names such as `BalancedTree` or `ArrayList`. I'll add a comment.

src/julia_gc.cc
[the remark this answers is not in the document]
rbehrends: As for C vs. C++, it's perfectly possible to redo this in C. The downside is that generics generally do not result in pretty or maintainable C code. You can either use macros to emulate generics or you can instantiate the generics from a template, e.g. as in `src/sortbase.h`. It's no big deal to rewrite it this way (probably something like a day of coding/testing), though, it's just a concern about code maintenance.

src/julia_gc.cc
rbehrends: I'm not sure if its a bug or an inefficiency. Basically, as I understand it, the task stacks can be roots themselves without being referenced from elsewhere. So, in parts, this is most likely intentional and simply hasn't created any performance issues, as task stacks tend to be short.

src/julia_gc.cc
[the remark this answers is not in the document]
fingolfin: I wasn't aware that operator new and delete need the C++ runtime, though I guess it makes some sense. Anyway, it is absolutely fine by me to not use that, I was just wondering... Thanks for the explanation.

As to redoing this in C: Actually, personally I am with keeping this in C++. Esp. if the code already existed before and is somewhat tested; a rewrite is just a chance to introduce new regressions :-).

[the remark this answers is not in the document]
fingolfin: That said, I'd still recommend to get rid of `StdAlloc` as a base class, i.e., no inheritance; just provide global `alloc` / `dealloc` template functions which always call `malloc` / `free`

src/julia_gc.cc
fingolfin: OK, that explains why the shift away from a treap. I also agree that we shouldn't use `std::set` and `std::map`.

src/julia_gc.cc
fingolfin: To be clear: this PR does *not* add the MarkCache. It only moves it around. Unfortunately, GitHub's diff displaying logic fails to properly detect that a big chunk of code was inserted... But you can search for `#define MARK_CACHE_BITS` on this page to find the old place of the code (around line 80)

src/julia_gc.cc
fingolfin: Yeah, but since `TEST_JULIA_GC_INTERNALS` never is `#defined`, that constructor is also never run -- except perhaps on you personal computer, where you added `-DTEST_JULIA_GC_INTERNALS` to your `CPPFLAGS`.

But tests really should be run on Travis CI all the time, to catch future regressions...

src/julia_gc.cc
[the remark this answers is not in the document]
rbehrends: Heh. I just went and redid it in C. As the rewrite was mostly just mechanical, I'm not too worried about regressions. That said, I'm personally fine either way.

src/julia_gc.cc
rbehrends: I totally agree that this is a stopgap measure. I would love to plug them into a testing framework, but as far as I can tell, we only have one for GAP code, not for C code. I could of course register a kernel function for those tests, but that also feels like a workaround.

In short, I'd be delighted to use whatever approach you think is ideal for testing, just let me know what you'd prefer.

src/julia_gc.cc
rbehrends: Okay, I've also added some array tests.

fingolfin:
@rbehrends this version doesn't compile cleanly, at least on Travis.

rbehrends:
@fingolfin Yeah, I saw it, for some reason clang incorrectly accepted a `static const` array size (a C++ leftover), but gcc complained (appropriately so).

src/dynarray.h
```diff
@@ -0,0 +1,192 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+#endif
+
+#define Array JOIN(ELEM_TYPE, Array)
+
+#define FN(sym) JOIN(Array, sym)
+
+#define JOIN(s1, s2) JOIN2(s1, s2)
+#define JOIN2(s1, s2) s1##s2
```
fingolfin: These two macros also appear in `sortbase.h`. And in `debug.h`, we have them under a different name:
```
#define _intern_CONCAT_(X, Y)  X ## Y
#define _intern_CONCAT(X, Y)  _intern_CONCAT_(X,Y)
```
Perhaps we should just give in and move this to, say, `src/gaputils.h` and document it properly (this is not a change request for this PR, just a general thought. @ChrisJefferson ?)

rbehrends:
I've factored out the tests and made `testkernel` a separate Makefile target under the assumption that more could be added to it later and kept conceptually different from libgap tests, even though the same framework is used. Is that reasonable or should I put them in with the libgap tests?

Makefile.rules
```diff
@@ -1090,7 +1090,24 @@ testlibgap: ${LIBGAPTESTS}
...
+	rm -f ${KERNELTESTS}
+
+# run all the tests in tst/testkernel
+testkernel: ${KERNELTESTS}
+	$(foreach v,$^, \
+		echo "Running $(v) ..." && \
+		$(v) -A -l $(top_srcdir) -q -T --nointeract >$(v).out && \
+		diff $(top_srcdir)/$(v).expect $(v).out && ) :
+
+.PHONY: testlibgap testkernel
```
fingolfin: Nice that you added `testkernel`, but this also should be hooked up to the CI, else those tests still never are run by anybody. A minimal patch for that cold look like this:
```patch
diff --git a/.travis.yml b/.travis.yml
index b11834b54..f04b91c7f 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,7 +136,7 @@ matrix:
           - libreadline-dev:i386
 
     # test error reporting and compiling as well as libgap
-    - env: TEST_SUITES="testspecial test-compile testlibgap"
+    - env: TEST_SUITES="testspecial test-compile testlibgap testkernel"
 
     # test Julia integration
     - env: TEST_SUITES="testinstall" JULIA=yes CONFIGFLAGS="--disable-Werror"
diff --git a/etc/ci.sh b/etc/ci.sh
index 009b241ce..b75d4c52c 100644
--- a/etc/ci.sh
+++ b/etc/ci.sh
@@ -167,6 +167,10 @@ GAPInput
     make testlibgap
     ;;
 
+  testkernel)
+    make testlibgap
+    ;;
+
   *)
     if [[ ! -f  $SRCDIR/tst/${TEST_SUITE}.g ]]
     then
```

src/julia_gc.c
```diff
@@ -30,11 +30,13 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <julia.h>
 #include <julia_gcext.h>
 
 
+
```
```suggestion
```

src/julia_gc.c
```diff
@@ -600,19 +425,95 @@ static void TryMarkRangeReverse(void * start, void * end)
...
+#define ELEM_TYPE TaskInfo
+#define COMPARE CmpTaskInfo
+
+#include "baltree.h"
+
+static TaskInfoTree * task_stacks = NULL;
+
+static void
+SafeScanTaskStack(PtrArray *stack, void * start, void * end)
```
```suggestion
static void SafeScanTaskStack(PtrArray * stack, void * start, void * end)
```

src/julia_gc.c
```diff
@@ -600,19 +425,95 @@ static void TryMarkRangeReverse(void * start, void * end)
...
+}
+
+#define ELEM_TYPE TaskInfo
+#define COMPARE CmpTaskInfo
+
+#include "baltree.h"
+
+static TaskInfoTree * task_stacks = NULL;
+
+static void
```
```suggestion
```

src/julia_gc.c
```diff
@@ -1089,3 +999,5 @@ void MarkJuliaWeakRef(void * p)
     if (JMarkTyped(p, jl_weakref_type))
         YoungRef++;
 }
+
+
```
```suggestion
```

src/julia_gc.c
```diff
@@ -1089,3 +999,5 @@ void MarkJuliaWeakRef(void * p)
     if (JMarkTyped(p, jl_weakref_type))
         YoungRef++;
 }
+
```
```suggestion
```

src/julia_gc.c
```diff
@@ -1030,7 +940,7 @@ void SwapMasterPoint(Bag bag1, Bag bag2)
 
 // HOOK: mark functions
 
-inline void MarkBag(Bag bag)
+void MarkBag(Bag bag)
```
fingolfin: Is this still necessary, now that we are back to C code?

src/julia_gc.c
```diff
@@ -991,7 +901,7 @@ UInt ResizeBag(Bag bag, UInt new_size)
             alloc_size++;
 
         // allocate new bag
-        header = AllocateBagMemory(header->type, alloc_size);
+        header = (BagHeader *)AllocateBagMemory(header->type, alloc_size);
```
fingolfin: Is this still necessary, now that we are back to C code?

src/julia_gc.c
```diff
@@ -959,7 +869,7 @@ Bag NewBag(UInt type, UInt size)
 
 #if !defined(SCAN_STACK_FOR_MPTRS_ONLY)
     // allocate the new masterpointer
-    bag = jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
+    bag = (Bag)jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
```
fingolfin: Is this still necessary, now that we are back to C code?

src/julia_gc.c
```diff
@@ -946,11 +856,11 @@ Bag NewBag(UInt type, UInt size)
         alloc_size++;
 
 #if defined(SCAN_STACK_FOR_MPTRS_ONLY)
-    bag = jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
+    bag = (Bag)jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
     SET_PTR_BAG(bag, 0);
 #endif
 
-    BagHeader * header = AllocateBagMemory(type, alloc_size);
+    BagHeader * header = (BagHeader *)AllocateBagMemory(type, alloc_size);
```
fingolfin: Is this still necessary, now that we are back to C code?

src/julia_gc.c
```diff
@@ -946,11 +856,11 @@ Bag NewBag(UInt type, UInt size)
         alloc_size++;
 
 #if defined(SCAN_STACK_FOR_MPTRS_ONLY)
-    bag = jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
+    bag = (Bag)jl_gc_alloc_typed(JuliaTLS, sizeof(void *), datatype_mptr);
```
fingolfin: Is this still necessary, now that we are back to C code?

PR #3518, opened 2019-06-24
Switch PLISTS and WeakPointerObjs to use IntObj to store the size

ChrisJefferson:
This switches PLISTS and WeakPointerObjs to use an IntObj to store their size.

This is part of a cleanup towards making all the builtin data types use IntObj to store their size.

The only notable problem is that now a compeletely zeroed block of memory isn't a valid Plist, as it needs to store INTOBJ_INT(0) as it's size.

There are two reasons to do this:

* It helps make clearer these datastructures can only be as big as an intobj (which was already true anyway in practice)
* It is required to get rid of a bug in saveworkspace (in a neat way; see #3505).

### DIFF
--- a/src/hpc/threadapi.c
+++ b/src/hpc/threadapi.c
@@ -374,10 +374,10 @@ static pthread_mutex_t KeepAliveLock;
 
 Obj KeepAlive(Obj obj)
 {
-    Obj newKeepAlive = NewBag(T_PLIST, 4 * sizeof(Obj));
+    Obj newKeepAlive = NEW_PLIST(T_PLIST, 4);
     SET_REGION(newKeepAlive, NULL);    // public region
     pthread_mutex_lock(&KeepAliveLock);
-    ADDR_OBJ(newKeepAlive)[0] = (Obj)3; /* Length 3 */
+    SET_LEN_PLIST(newKeepAlive, 3);
     KEPTALIVE(newKeepAlive) = obj;
     PREV_KEPT(newKeepAlive) = LastKeepAlive;
     NEXT_KEPT(newKeepAlive) = (Obj)0;
--- a/src/weakptr.c
+++ b/src/weakptr.c
@@ -80,7 +80,7 @@ static inline void FORGET_WP(Obj wp, UInt pos)
 */
 static inline void STORE_LEN_WPOBJ(Obj wp, Int len)
 {
-    ADDR_OBJ(wp)[0] = (Obj)len;
+    ADDR_OBJ(wp)[0] = INTOBJ_INT(len);
 }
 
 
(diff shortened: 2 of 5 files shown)


ChrisJefferson:
Note that I have some tests to ensure that these behave well, and there are no problems are 2^28, but I'm not sure I want to commit them and have them regularly tested, as they can only be run on 32-bit machines, needs lots of memory, and also they require running lots of small test files in seperate GAPs, as sometimes GASMAN fails to collect the huge bags I'm making and GAP runs out of memory.

fingolfin:
I just grepped through packages for `NewBag.*T_PLIST` and it seems none of them do that (unless they put a newline after `NewBag` and before `T_PLIST`... but then, shame on them for producing unreadable code). Good! I'll properly review this PR after I am done teaching.

src/hpc/threadapi.c
```diff
@@ -374,10 +374,10 @@ static pthread_mutex_t KeepAliveLock;
 
 Obj KeepAlive(Obj obj)
 {
-    Obj newKeepAlive = NewBag(T_PLIST, 4 * sizeof(Obj));
+    Obj newKeepAlive = NEW_PLIST(T_PLIST, 4);
```
fingolfin: I like this. I think we should actually add even more `NEW_FOO` functions, and try to call `NewBag` directly as little as possible.

src/weakptr.c
```diff
@@ -636,7 +636,7 @@ static Obj CopyObjWPObj(Obj obj, Int mut)
         ADDR_OBJ(copy)[0] = CONST_ADDR_OBJ(obj)[0];
     }
     else {
-        copy = NewBag( T_PLIST+IMMUTABLE, SIZE_OBJ(obj) );
+        copy = NEW_PLIST( T_PLIST+IMMUTABLE, LengthWPObj(obj) );
         SET_LEN_PLIST(copy,LengthWPObj(obj));
```
fingolfin: Note that the value of `LengthWPObj(obj)` could change between the two previous lines. And the loop a few lines later again uses `SIZE_OBJ(obj)`...

src/weakptr.c
```diff
@@ -636,7 +645,7 @@ static Obj CopyObjWPObj(Obj obj, Int mut)
         ADDR_OBJ(copy)[0] = CONST_ADDR_OBJ(obj)[0];
     }
     else {
-        copy = NewBag( T_PLIST+IMMUTABLE, SIZE_OBJ(obj) );
+        copy = NEW_PLIST(T_PLIST + IMMUTABLE, LengthWPObj(obj));
```
fingolfin:
```suggestion
        copy = NEW_PLIST_IMM(T_PLIST, LengthWPObj(obj));
```
I was trying in the past to get rid of as many uses of `IMMUTABLE` as possible, in the hope that at some point we turn it into an object flag (also Markus worked hard towards that goal).

PR #3521, opened 2019-06-25
Add more IsSimpleGroup-related implication; cleanup

fingolfin:
In GAP, IsAlmostSimpleGroup only applies to *nonabelian* simple groups,
whereas IsSimpleGroup also is true for abelian simple groups.

Instead, we add two weaker implications from IsSporadicSimpleGroup
resp. from IsSimpleGroup and IsPerfectGroup, to IsAlmostSimpleGroup.

UPDATE: the original issue got fixed in PR #3522; this PR now adds a test for it, and performs some related cleanup. Best to review this on a per-commit basis.

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -624,11 +624,17 @@ InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
 ##  <#/GAPDoc>
 ##
 DeclareProperty( "IsSimpleGroup", IsGroup );
-InstallTrueMethod( IsGroup, IsSimpleGroup );
+InstallTrueMethod( IsGroup and IsNonTrivial, IsSimpleGroup );
 DeclareSynonymAttr( "IsNonabelianSimpleGroup", IsSimpleGroup and IsPerfectGroup );
-InstallTrueMethod( IsSimpleGroup, IsNonabelianSimpleGroup );
+InstallTrueMethod( HasIsAbelian, IsNonabelianSimpleGroup );
 
 InstallIsomorphismMaintenance( IsSimpleGroup, IsGroup, IsGroup );
+
+# abelian simple groups are cyclic p-groups and not perfect
+InstallTrueMethod( IsCyclic, IsSimpleGroup and IsAbelian );
+InstallTrueMethod( IsPGroup, IsSimpleGroup and IsAbelian );
+InstallTrueMethod( HasIsPerfectGroup, IsSimpleGroup and IsAbelian );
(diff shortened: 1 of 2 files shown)


lib/grp.gd
```diff
@@ -709,8 +710,10 @@ InstallTrueMethod( IsSimpleGroup, IsSporadicSimpleGroup );
 DeclareProperty( "IsAlmostSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup, IsAlmostSimpleGroup );
 
-# a simple group is almost simple
-InstallTrueMethod( IsAlmostSimpleGroup, IsSimpleGroup );
+# only a nonabelian simple group is almost simple
+#InstallTrueMethod( IsAlmostSimpleGroup, IsSimpleGroup ); # WRONG!
+InstallTrueMethod( IsAlmostSimpleGroup, IsSporadicSimpleGroup );
```
wilfwilson: Technically this line is redundant, since a known sporadic simple group is simple and perfect via the above `InstallTrueMethod`s, and is therefore almost simple via the below `InstallTrueMethod`. Could be removed, although also no harm in keeping it I suppose.

fingolfin: Good point, done.

fingolfin:
I actually added a few more implications. But I now think we should first merge @hulpke's PR #3522, then I can rebase this PR and adapt it to use `IsNonabelianSimpleGroup`

fingolfin:
@wilfwilson This PR has changed substantially since you reviewed it, hence I dismissed your review, as I'd consider it unfair if I (ab)used your approval of essentially a completely different PR.

lib/grp.gd
```diff
@@ -711,12 +714,11 @@ InstallTrueMethod( IsSimpleGroup, IsSporadicSimpleGroup );
 ##  <#/GAPDoc>
 ##
 DeclareProperty( "IsAlmostSimpleGroup", IsGroup );
-InstallTrueMethod( IsGroup, IsAlmostSimpleGroup );
+InstallTrueMethod( IsGroup and IsNonTrivial, IsAlmostSimpleGroup );
 
-# a simple group is almost simple
+# nonabelian simples group are almost simple
```
wilfwilson:
```suggestion
# nonabelian simple group are almost simple
```

wilfwilson: GitHub has assumed that my neutral review was an approving reviewing. Wouldn't be terrible to merge with this typo, but why not fix it ;)

fingolfin: Oops, good catch, thanks!

PR #3522, opened 2019-06-25
ENHANCE: Added `IsNonabelianSimpleGroup`

hulpke:
and use it in library wherever this is intended in place of `IsSimpleGroup`.

There were places in the code which uses `IsSimpleGroup` intending nonabelian simple. While thy did not cause obvious errors, it is cleaner to use a dedicated filter and will also inherit the `IsAlmostSimpleGroup` implication.

Release Notes:
Clarified definition and use of `IsSimpleGroup` to allow for abelian simple groups and introduced `IsNonabelianSimpleGroup`.

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -636,23 +636,29 @@ InstallIsomorphismMaintenance( IsSporadicSimpleGroup,
 #############################################################################
 ##
 #P  IsSimpleGroup( <G> )
+#P  IsNonabelianSimpleGroup( <G> )
 ##
 ##  <#GAPDoc Label="IsSimpleGroup">
 ##  <ManSection>
 ##  <Prop Name="IsSimpleGroup" Arg='G'/>
+##  <Prop Name="IsNonabelianSimpleGroup" Arg='G'/>
 ##
 ##  <Description>
 ##  A group is <E>simple</E> if it is nontrivial and has no nontrivial normal
-##  subgroups.
+##  subgroups. A <E>nonabelian simple</E> group is simple and not abelian.
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
 DeclareProperty( "IsSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup, IsSimpleGroup );
+DeclareSynonymAttr( "IsNonabelianSimpleGroup", IsSimpleGroup and IsPerfectGroup );
+InstallTrueMethod( IsSimpleGroup, IsNonabelianSimpleGroup );
 
 InstallIsomorphismMaintenance( IsSimpleGroup,
     IsGroup and IsSimpleGroup, IsGroup );
+
+InstallIsomorphismMaintenance( IsPerfectGroup, IsGroup, IsGroup );
 
 InstallTrueMethod( IsSimpleGroup, IsSporadicSimpleGroup );
 
@@ -710,7 +716,7 @@ DeclareProperty( "IsAlmostSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup, IsAlmostSimpleGroup );
 
 # a simple group is almost simple
-InstallTrueMethod( IsAlmostSimpleGroup, IsSimpleGroup );
+InstallTrueMethod( IsAlmostSimpleGroup, IsNonabelianSimpleGroup );
 
 
 #############################################################################
@@ -1877,7 +1883,7 @@ DeclareAttribute( "MinimalNormalSubgroups", IsGroup );
 ##  <Example><![CDATA[
 ##  gap> g:=SymmetricGroup(4);;NormalSubgroups(g);
 ##  [ Sym( [ 1 .. 4 ] ), Alt( [ 1 .. 4 ] ),
-##    Group([ (1,4)(2,3), (1,2)(3,4) ]), Group(()) ]
+##    Group([ (1,4)(2,3), (1,3)(2,4) ]), Group(()) ]
 ##  ]]></Example>
 ##  <P/>
 ##  The algorithm for the computation of normal subgroups is described in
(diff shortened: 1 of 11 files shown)


lib/grp.gi
```diff
@@ -504,6 +504,13 @@ InstallMethod( IsSimpleGroup,
     return true;
     end );
 
+#############################################################################
+##
+#M  IsNonabelianSimpleGroup( <G> )
+##
+InstallMethod( IsNonabelianSimpleGroup,"generic",[IsGroup],
+  G->(not IsAbelian(G)) and IsSimpleGroup(G));
```
fingolfin: This would then be redundant.

lib/grpnice.gi
```diff
@@ -539,6 +539,9 @@ PropertyMethodByNiceMonomorphism( IsPerfectGroup,
 PropertyMethodByNiceMonomorphism( IsSimpleGroup,
     [ IsGroup ] );
 
+PropertyMethodByNiceMonomorphism( IsNonabelianSimpleGroup,
```
fingolfin: This would also be redundant

lib/grp.gi
fingolfin: This method can (should) be removed.

lib/grp.gd
```diff
@@ -711,6 +718,7 @@ InstallTrueMethod( IsGroup, IsAlmostSimpleGroup );
 
 # a simple group is almost simple
 InstallTrueMethod( IsAlmostSimpleGroup, IsSimpleGroup );
```
fingolfin: Could you also remove this bad implication? Then we could merge your PR before my PRs #3521 and #3525 , which I will rebase and adapt to use `IsNonabelianSimpleGroup`.
```suggestion
```

lib/grpnice.gi
hulpke: Tganks -- I've done these changes.

lib/grp.gd
fingolfin: On second thought, you can also leave it in and I'll remove it in my PRs. Whichever you prefer.

lib/grplatt.gi
```diff
@@ -923,7 +923,8 @@ InstallGlobalFunction(LatticeViaRadical,function(arg)
       HN:=Image(hom,H);
       c:=LatticeByCyclicExtension(f,
 	  [u->IsSubset(HN,u),u->IsSubset(HN,u)])!.conjugacyClassesSubgroups;
-    elif select<>fail and (select=IsPerfectGroup  or select=IsSimpleGroup) then
+    elif select<>fail and (select=IsPerfectGroup  or select=IsAlmostSimpleGroup)
```
fingolfin: This change goes beyond what the description of this PR. Is it intentional? If so, I of course trust that it is the right thing; I just wanted to verify, as it differed from the other changes.

lib/grp.gd
hulpke: I've just taken it out :-)

lib/grplatt.gi
hulpke: Thanks for noticing.  Yes, it needs to to be `NonabelianSimple`. Fixed.

lib/grplatt.gi
```diff
@@ -923,7 +923,8 @@ InstallGlobalFunction(LatticeViaRadical,function(arg)
       HN:=Image(hom,H);
       c:=LatticeByCyclicExtension(f,
 	  [u->IsSubset(HN,u),u->IsSubset(HN,u)])!.conjugacyClassesSubgroups;
-    elif select<>fail and (select=IsPerfectGroup  or select=IsSimpleGroup) then
+    elif select<>fail and (select=IsPerfectGroup  or select=IsNonabelianSimpleGroup)
```
fingolfin: `IsNonabelianSimpleGroup` implies `IsPerfectGroup` now, though, so this would be redundant.
```suggestion
    elif select=IsPerfectGroup then
```

lib/grplatt.gi
```diff
@@ -923,7 +923,8 @@ InstallGlobalFunction(LatticeViaRadical,function(arg)
...
+      then
```
```suggestion
```

lib/morpheus.gi
```diff
@@ -492,7 +492,8 @@ v, val, o, i, comb, best,actbase,action;
 	  end);
       fi;
 
-      if hom=fail and not IsAbelian(Socle(g)) and IsSimpleGroup(Socle(g)) then
+      if hom=fail and not IsAbelian(Socle(g))
+        and IsNonabelianSimpleGroup(Socle(g)) then
```
fingolfin: Isn't the `IsAbelian` check redundant now? Alternatively, the switch from `IsSimpleGroup` to `IsNonabelianSimpleGroup` is redundant. (Not that it's important, the code is correct either way)

lib/grplatt.gi
hulpke: No, this is not a check, but a parameter check for particular selection functions given to the routine (to only calculate subgroups satisfying some condition). If the user only wants perfect or nonabelian almost simple subgroups, it is sufficient to initialize with perfect subgroups in the radical factor.

lib/morpheus.gi
hulpke: Indeed it is. When I went through the changes I tried to not do anything nontrivial with the logic (so it could be understood easily).
Indeed here the `IsAbelian` can go now, will do.

lib/grplatt.gi
fingolfin: ahhhh, OK, I see, my bad!

fingolfin: Can't this be simplified? (I am not actually requesting that you do that, though, just wondering; it is fine to merge this PR as-is):
```suggestion
    elif select=IsPerfectGroup or select=IsNonabelianSimpleGroup then
```

fingolfin:
I fixed the diff in the manual, tests are running again

wilfwilson:
The Travis build has passed but GitHub doesn't seem to have realised, so I can't merge.

lib/grplatt.gi
hulpke: Yes, it could be simplified. (I think it is a relic from an older construct). I'll do so.

hulpke:
I've included the last suggestion and rebases, lets hope now Travis gets it.

PR #3524, opened 2019-06-26
Switch strings to store their length as an intobj

ChrisJefferson:
Switch strings to use an intobj as their length.

This should not be a visible change, as no kernel programmer should be accessing the length directly, and if anyone was creating strings on a 32-bit machine with length longer than fits in an intobj, GAP often crashed anyway.

### DIFF
--- a/src/compiler.c
+++ b/src/compiler.c
@@ -199,19 +199,19 @@ typedef UInt4           LVar;
 
 #define INFO_FEXP(fexp)         PROF_FUNC(fexp)
 #define SET_INFO_FEXP(fexp,x)   SET_PROF_FUNC(fexp,x)
-#define NEXT_INFO(info)         PTR_BAG(info)[0]
-#define NR_INFO(info)           (*((Int*)(PTR_BAG(info)+1)))
-#define NLVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+2)))
-#define NHVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+3)))
-#define NTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+4)))
-#define NLOOP_INFO(info)        (*((Int*)(PTR_BAG(info)+5)))
-#define CTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+6)))
-#define TNUM_LVAR_INFO(info,i)  (*((Int*)(PTR_BAG(info)+7+(i))))
+#define NEXT_INFO(info)         PTR_BAG(info)[1]
+#define NR_INFO(info)           (*((Int*)(PTR_BAG(info)+2)))
+#define NLVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+3)))
+#define NHVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+4)))
+#define NTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+5)))
+#define NLOOP_INFO(info)        (*((Int*)(PTR_BAG(info)+6)))
+#define CTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+7)))
+#define TNUM_LVAR_INFO(info,i)  (*((Int*)(PTR_BAG(info)+8+(i))))
 
 #define TNUM_TEMP_INFO(info,i)  \
-    (*((Int*)(PTR_BAG(info)+7+NLVAR_INFO(info)+(i))))
+    (*((Int*)(PTR_BAG(info)+8+NLVAR_INFO(info)+(i))))
 
-#define SIZE_INFO(nlvar,ntemp)  (sizeof(Int) * (8 + (nlvar) + (ntemp)))
+#define SIZE_INFO(nlvar,ntemp)  (sizeof(Int) * (1 + 8 + (nlvar) + (ntemp)))
 
 #define W_UNUSED                0       /* TEMP is currently unused        */
 #define W_HIGHER                (1L<<0) /* LVAR is used as higher variable */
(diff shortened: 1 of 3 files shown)


src/compiler.c
```diff
@@ -199,19 +199,19 @@ typedef UInt4           LVar;
...
-#define CTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+6)))
-#define TNUM_LVAR_INFO(info,i)  (*((Int*)(PTR_BAG(info)+7+(i))))
+#define NEXT_INFO(info)         PTR_BAG(info)[1]
+#define NR_INFO(info)           (*((Int*)(PTR_BAG(info)+2)))
+#define NLVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+3)))
+#define NHVAR_INFO(info)        (*((Int*)(PTR_BAG(info)+4)))
+#define NTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+5)))
+#define NLOOP_INFO(info)        (*((Int*)(PTR_BAG(info)+6)))
+#define CTEMP_INFO(info)        (*((Int*)(PTR_BAG(info)+7)))
+#define TNUM_LVAR_INFO(info,i)  (*((Int*)(PTR_BAG(info)+8+(i))))
```
fingolfin: Thanks. I made a similar change independently yesterday, but luckily saw your PR here before I completely finished it, and you version seems slightly nicer anyway :-).

src/compiler.c
```diff
@@ -5254,8 +5254,8 @@ Int CompileFunc (
     compilerMagic2 = magic2;
 
     /* create 'CompInfoGVar' and 'CompInfoRNam'                            */
-    CompInfoGVar = NewBag( T_STRING, sizeof(UInt) * 1024 );
-    CompInfoRNam = NewBag( T_STRING, sizeof(UInt) * 1024 );
+    CompInfoGVar = NewBag( T_DATOBJ, sizeof(UInt) * 1024 );
+    CompInfoRNam = NewBag( T_DATOBJ, sizeof(UInt) * 1024 );
```
fingolfin: This part of the change conflicts with PR #3527 in a trivial way.

fingolfin: I also find the end of the commit message a bit confusing. It says:
> The values stored (RNams and GVars) are never 0, so there are no
problems with overwriting the 0th element (where the type of a
DATOBJ is stored).

I think this only refers to `CompInfoGVar` and `CompInfoRNam`, not `info`, but that's IMHO not so clear if one looks at this commit w/o prior knowledge, and then is mildly confusing. But even then it is IMHO confusing, because we do not store the RNam and GVar; but rather, we use them, treated as UInt, and then use them as an offset into `(UInt *)ADDR_OBJ(CompInfoGVar)` (resp. RNam).. And these offsets always happen to be at least 1, so we never overwrite slot 0.

Perhaps just drop these two lines from the commit, and that paragraph from the PR; then our two PRs don't conflict anymore?

src/stringobj.c
```diff
@@ -415,10 +415,14 @@ static Obj FuncREVNEG_STRING(Obj self, Obj val)
 Obj NEW_STRING ( Int len )
 {
   Obj res;
-  if (len < 0)
-       ErrorQuit(
-           "NEW_STRING: Cannot create string of negative length %d",
-           (Int)len, 0L);
+  if (len < 0) {
+      ErrorQuit("NEW_STRING: Cannot create string of negative length %d",
```
fingolfin: Minor consistency remark: most other error messages of this style continue in lower case after the colon:
```suggestion
      ErrorQuit("NEW_STRING: cannot create string of negative length %d",
```
Now, I realize that you only tweaked an existing message here, but you also added new errors which then of course should be consistent with this one, too...

src/stringobj.c
```diff
@@ -415,10 +415,14 @@ static Obj FuncREVNEG_STRING(Obj self, Obj val)
...
+                (Int)len, 0L);
+  }
+  if (len > INT_INTOBJ_MAX) {
+      ErrorQuit("NEW_STRING: Size must be a small integer", 0L, 0L);
```
fingolfin:
```suggestion
      ErrorQuit("NEW_STRING: length must be a small integer", 0L, 0L);
```
We say "length" everywhere else, don't we?

src/stringobj.c
```diff
@@ -441,13 +445,20 @@ Int             GrowString (
     UInt                len;            /* new physical length             */
     UInt                good;           /* good new physical length        */
 
+    if (need > INT_INTOBJ_MAX)
+        ErrorMayQuit("GrowString: String size too large", 0, 0);
```
fingolfin: Perhaps this?
```suggestion
        ErrorMayQuit("GrowString: new string length is too large", 0, 0);
```
Also, why use `ErrorMayQuit` here, but `ErrorQuit` in `NEW_STRING`?

src/stringobj.c
```diff
@@ -441,13 +445,20 @@ Int             GrowString (
...
+
     /* find out how large the data area  should become                     */
     good = 5 * (GET_LEN_STRING(list)+3) / 4 + 1;
+    if (good > INT_INTOBJ_MAX)
+        good = INT_INTOBJ_MAX;
+
```
fingolfin:
```suggestion
```

src/stringobj.c
```diff
@@ -441,13 +445,20 @@ Int             GrowString (
...
 
     /* but maybe we need more                                              */
     if ( need < good ) { len = good; }
     else               { len = need; }
 
+
```
fingolfin:
```suggestion
```

src/stringobj.c
fingolfin: Looking at this function again inspired me to write PR #3528

wilfwilson:
Sorry, there conflicts now, since I merged #3527.

ChrisJefferson:
I think everything is cleaned up now (let's wait for the tests to pass)

PR #3527, opened 2019-06-26
kernel: add NewKernelBuffer and MakeRat helpers

fingolfin:
This slightly overlaps with PR #3524 but the conflict is trivial to resolve, I think, in either direction.

### DIFF
--- a/src/objects.h
+++ b/src/objects.h
@@ -920,6 +920,19 @@ void SetTypeDatObj(Obj obj, Obj type);
 
 /****************************************************************************
 **
+*F  NewKernelBuffer( <size> )  . . . . . . . . . . return a new kernel buffer
+**
+**  Return a new T_DATOBJ of the specified <size>, with its type set to the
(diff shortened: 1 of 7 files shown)


src/objects.h
```diff
@@ -918,6 +918,15 @@ EXPORT_INLINE void SET_TYPE_DATOBJ(Obj obj, Obj val)
 void SetTypeDatObj(Obj obj, Obj type);
 
 
+/****************************************************************************
+**
+*F  NewKernelBuffer( <size> )  . . . . . . . . . . return a new kernel buffer
+**
+**  TODO: document this
```
wilfwilson: Do you intend to write this documentation as part of this PR?

fingolfin: Oops, that was a mistake. Thanks, I'll fix it ASAP.

fingolfin: Fixed now, thanks again.

PR #3528, opened 2019-06-27
kernel: clean up NEW_STRING

fingolfin:
The bag returned by NewBag is zero initialized, so we don't
need to insert null terminators. Also reformat.

### DIFF
--- a/src/stringobj.c
+++ b/src/stringobj.c
@@ -412,24 +412,15 @@ static Obj FuncREVNEG_STRING(Obj self, Obj val)
 **  character and "first behind last" set to zero
 **
 */
-Obj NEW_STRING ( Int len )
+Obj NEW_STRING(Int len)
 {
-  Obj res;
-  if (len < 0) {
-      ErrorQuit("NEW_STRING: cannot create string of negative length %d",
-                (Int)len, 0L);
-  }
-  if (len > INT_INTOBJ_MAX) {
-      ErrorQuit("NEW_STRING: length must be a small integer", 0L, 0L);
-  }
-
-  res = NewBag( T_STRING, SIZEBAG_STRINGLEN(len)  ); 
-  SET_LEN_STRING(res, len);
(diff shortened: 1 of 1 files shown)


fingolfin:
This PR together together with #3518 probably warrant an increment of the kernel version, no?

src/stringobj.c
```diff
@@ -412,20 +412,14 @@ static Obj FuncREVNEG_STRING(Obj self, Obj val)
...
-  res = NewBag( T_STRING, SIZEBAG_STRINGLEN(len)  ); 
-  SET_LEN_STRING(res, len);
-  /* it may be sometimes useful to have trailing zero characters */
-  Char *p = CSTR_STRING(res);
-  p[0] = '\0';
-  p[len] = '\0';
-  return res;
+    if (len < 0)
+        ErrorQuit("NEW_STRING: cannot create string of negative length %d",
+                  (Int)len, 0);
```
fingolfin: Actually, thinking about it again, I wonder why we even call `ErrorQuit` here, and not `GAP_ASSERT`, as in other similar functions. And then make it `static inline` and move it into the header file...

ChrisJefferson: I agree that I can't think why anyone should be calling such a low-level function with a negative length.

fingolfin:
Rebased to resolve conflict and switch to `GAP_ASSERT`. Of course now the other length check could also be changed to an `GAP_ASSERT` -- that would match what we do in `NEW_PLIST`. On the other hand, I wonder if getting a too large value here might be something that could actually happen in code... So we should always check for it, but `GAP_ASSERT` only has an effect a debug builds. So perhaps we actually should change `NEW_PLIST` to always check for overflows, not in debug builds?!?!? I am really torn on this *sigh*

ChrisJefferson:
So, I do think NEW_STRING should check for overflow, as there are a few places where we decide to make a large string, and don't check the length externally.

If the same check should also exist in NEW_PLIST I'm not sure.

PR #3537, opened 2019-07-02
Further improvements to factor group and automorphisms

hulpke:
Avoid/workaround for certain bad cases.

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -168,12 +168,83 @@ BindGlobal("AGSRAutomLift",function(ocr,nat,fhom,miso)
 end);
 
 
-# step through intermediate subgroups to find subgroup that satisfies
-# condition
-BindGlobal("SubgroupConditionAbove",function(G,cond,S)
-local c,hom,q,a,t,int,bad,have,ups,up,new,u;
+# Find a larger subgroup that satisfies a condition, when testing
+# the condition can become expensive.
+# First try `SubgroupProperty`, but when it stalls attempt to find
+# minimal supergroups and prove that none of them satisfies.
+InstallGlobalFunction(SubgroupConditionAbove,function(G,cond,S)
+local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort;
 
-  if ForAll(GeneratorsOfGroup(G),cond) then return G;fi;
+  Info(InfoMorph,2,"SubgroupAbove ",IndexNC(G,S));
+
+  # first try, how far `SubgroupProperty` goes
+  good:=[];
+  bad:=[];
+  b:=0;
+  abort:=false;
+  c:=SubgroupProperty(G,
+    function(elm)
+      if abort then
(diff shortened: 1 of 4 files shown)


lib/autsr.gi
```diff
@@ -170,10 +170,74 @@ end);
...
+InstallGlobalFunction(SubgroupConditionAbove,function(G,cond,S)
+local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good;
+
+  Info(InfoMorph,2,"SubgroupAbove ",IndexNC(G,S));
+
+  # first try, how far `SubgroupProperty` goes
+  good:=[];
+  bad:=[];
+  b:=0;
+  a:=false;
```
fingolfin: could perhaps `a` be renamed to `abort` or `bailout` or something?

lib/autsr.gi
```diff
@@ -666,7 +730,7 @@ condcnt:=0;
     #if IsNormal(sub,Aperm) and IndexNC(sub,Aperm)<=50000 then
     #  sub:=SubgroupConditionAbove(sub,cond,Aperm);
     #else
```
fingolfin: So I guess that means the commented out code above can go away?

hulpke: Yes.

lib/autsr.gi
hulpke: But then I'm using a further variable (since `a` is re-used later) :-)
(Yes, I've never made the step of realizing that I have more than 8 registers to work with). Changed.

fingolfin: Heh

lib/autsr.gi
```diff
@@ -170,10 +170,74 @@ end);
...
+  if abort=false then 
+    # we actually found the subgroup
+    Info(InfoMorph,2,"SubgroupProperty finds ",IndexNC(G,c));
+    return c;
+  fi; 
+
+  Info(InfoMorph,2,"intermediate improvement ",IndexNC(G,S));
+
+  good:=false;
+  if IndexNC(G,S)<=10^5 then
```
wilfwilson: @hulpke _If_ you think it would be possible to give some brief explanation for why you have chosen the 'magic constants' that you have in this PR, in a way that could make future maintenance easier, then I would appreciate that. There is this line, the `b*2000` above, and the updated `lim` below, that seem to be things of this kind.

I envisage something like: `with 10^5, this typically takes no more than a few seconds`. That way, if in the future `10^6` would take just a few seconds, a future maintainer could choose to increase the limit.

If you don't think that's a good idea/not useful/don't want to, then that's okay.

hulpke: I've added some text explaining the numbers. The issue of the 10^5 is more memory than time. The 2000 is 1/20 percent as both establishing a "really small success rate" versus not trying in vain for too long.

wilfwilson:
Thanks for those comments @hulpke

PR #3543, opened 2019-07-05
Add PositionSortedBy

fingolfin:
One thing to point out is that the interface is a bit different from
PositionSorted, which might be confusing: the object you pass in is
*not* a (potential) element of the list, but rather the result of
applying the "by func" to an object (or at least something that looks
like such a result). The reason: first off, replacing the element by its
func value would be the very first thing done by any implementation
anyway. And a typical application of this function is to search for an
object matching a a certain criterion -- say, the first item which is a
list whose first entry is 42. With the current model, one can write

    PositionSortedBy([ "a", "ab" ], 4, Length);

and is not forced to first construct an actual object, like here:

    PositionSortedBy([ "a", "ab" ], "abcd", Length);

In this example, that's a pretty small saving, but in bigger examples,
it can add up. In particular, this is motivated by real world usage
of this function.

---

By the way, this is perhaps my oldest still active branch for GAP: I wrote
it in early 2014, around the time I discovered `PositionFirstComponent`
and tried to get rid of that (which finally happened this year). I then never
could get myself to clean this up, but today I wanted to have such a function
in `recog`, so I finally had a good enough reason to dig in and write documentation
and tests and all that ;-)

### DIFF
--- a/doc/ref/lists.xml
+++ b/doc/ref/lists.xml
@@ -1405,6 +1405,7 @@ The latter can be done also using <Ref Func="ListWithIdenticalEntries"/>.
 <#Include Label="PositionCanonical">
 <#Include Label="PositionNthOccurrence">
 <#Include Label="PositionSorted">
+<#Include Label="PositionSortedBy">
 <#Include Label="PositionSet">
 <#Include Label="PositionMaximum">
 <#Include Label="PositionProperty">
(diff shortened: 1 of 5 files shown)


fingolfin:
@wilfwilson I tried to improve it, please have another look

PR #3550, opened 2019-07-10
kernel: rename ELM2_LIST -> ELM_MAT, etc.

fingolfin:
Also fixed a minor bug in the syntax tree code for `STAT_ASS_MAT` (formerly known as `STAT_ASS2_LIST`).

I think these new names are much easier to grasp for somebody looking at the kernel code then the old ones.

I have further PRs lined up which build on this, to add libgap APIs for matrices; to restrict the row/column arguments to positive integers; and to split the "matrix" access functions into separate operations (so instead of installing a method for `ELM_LIST` resp. `\[\]`, you'd install it for `ELM_MAT` resp. `\[\,\]`;  and `MatElm` would become a synonym for that; similar for `ASS_LIST`/`ASS_MAT`/`\[\,\]\:\=`/`SetMatElm`). But those all may be more controversial; I hope this PR isn't.

Note that this doesn't affect any packages as far as I know, as no kernel extension uses `ELM2_LIST` or `ASS2_LIST`.

I also renamed `ISB2_LIST` to `ISB_MAT` and `UNB2_LIST` to `UNB_MAT`, but actually I am seriously considering that we might want to drop support for unbinding matrix entries; matrices should never have unbound entries, after all. But that, too, is a rather bigger and potentially controversial change, so...

### DIFF
--- a/src/lists.h
+++ b/src/lists.h
@@ -197,7 +197,7 @@ EXPORT_INLINE Int ISB_LIST(Obj list, Int pos)
 
 Int ISBB_LIST(Obj list, Obj pos);
 
-Int ISB2_LIST(Obj list, Obj pos1, Obj pos2);
+Int ISB_MAT(Obj list, Obj row, Obj col);
 
 
 /****************************************************************************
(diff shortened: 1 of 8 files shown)


src/lists.h
```diff
@@ -324,14 +324,14 @@ EXPORT_INLINE Obj ELM_LIST(Obj list, Int pos)
 
 /****************************************************************************
 **
-*F  ELM2_LIST( <list>, <pos1>, <pos2> ) . . . . select an element from a list
+*F  ELM_MAT( <list>, <row>, <col> ) . . . . select an element from a list
```
ThomasBreuer: write `select an element from a matrix`

PR #3551, opened 2019-07-11
Add new kernel operations ELM_MAT, ASS_MAT

fingolfin:
This continues the work begun in PR #3550.

The kernel functions with the same name now delegate to these operations instead of `ELM_LIST`, `ASS_LIST`.

Also turned `MatElm`, `SetMatElm` into synonyms for `ELM_MAT`, `ASS_MAT`.

Code which already used MatrixObj for a long time and thus installed methods for `MatElm`, `SetMatElm` will thus work unchanged. Code which adapted to the recommendation in GAP 4.10 and thus instead installed methods for `ELM_LIST`, `ASS_LIST` with 3 resp. 4 arguments now should be adapted to (again) install methods for `MatElm`, `SetMatElm` for maximal compatibility. To ensure this code
works after this change, though, we provide low ranked fallback methods in `ELM_MAT`, `ASS_MAT` which delegate to `ELM_LIST`, `ASS_LIST`. At the same time, we remove the low-ranked delegations from `ELM_LIST`, `ASS_LIST` to `MatElm`, `SetMatElm` to ensure we don't end up delegating back and forth endlessly.

As far as I could tell, the only package which is (mildly) impacted by this is `cvec`; I'll shortly release an update of that which addresses all those concerns, though, and which should work seamlessly with and without this PR. Of course there might be further code out there which is not in deposited packages, and which I thus can't test, but I am confident it'll be OK, too -- though of course I can't be sure without being able to look at that code.

### DIFF
--- a/hpcgap/lib/vecmat.gi
+++ b/hpcgap/lib/vecmat.gi
@@ -586,11 +586,6 @@ InstallOtherMethod( \[\], "for GF2 matrix",
       IsPosInt ],
     ELM_GF2MAT );
 
-InstallMethod( \[\], "for GF2 matrix",
-    [ IsGF2MatrixRep,
-      IsPosInt, IsPosInt ],
-    MAT_ELM_GF2MAT );
-
 
 #############################################################################
 ##
@@ -609,13 +604,6 @@ InstallOtherMethod( \[\]\:\=,
       IsPosInt,
       IsObject ],
     ASS_GF2MAT );
-
-InstallMethod( \[\]\:\=,
-    "for GF2 matrix",
-    [ IsGF2MatrixRep,
-      IsPosInt, IsPosInt,
-      IsObject ],
-    SET_MAT_ELM_GF2MAT );
 
 
 #############################################################################
@@ -2383,7 +2371,7 @@ InstallMethod( MatElm, "for a gf2 matrix and two integers",
   [ IsGF2MatrixRep, IsPosInt, IsPosInt ],
   MAT_ELM_GF2MAT );
 InstallMethod( SetMatElm, "for a gf2 matrix, two integers, and a ffe",
-  [ IsGF2MatrixRep, IsPosInt, IsPosInt, IsFFE ],
+  [ IsGF2MatrixRep and IsMutable, IsPosInt, IsPosInt, IsFFE ],
   SET_MAT_ELM_GF2MAT );
 
 InstallMethod( Matrix, "for a list of vecs, an integer, and a gf2 mat",
--- a/lib/mat8bit.gi
+++ b/lib/mat8bit.gi
@@ -70,7 +70,7 @@ end);
 #M  Length( <mat> )
 ##
 
-InstallOtherMethod( Length, "For a compressed MatFFE",
+InstallOtherMethod( Length, "for a compressed MatFFE",
         true, [IsList and Is8BitMatrixRep], 0, m->m![1]);
 
 #############################################################################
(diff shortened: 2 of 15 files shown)


hpcgap/lib/vecmat.gi
```diff
@@ -586,11 +586,6 @@ InstallOtherMethod( \[\], "for GF2 matrix",
       IsPosInt ],
     ELM_GF2MAT );
 
-InstallMethod( \[\], "for GF2 matrix",
-    [ IsGF2MatrixRep,
-      IsPosInt, IsPosInt ],
-    MAT_ELM_GF2MAT );
```
fingolfin: This is now redundant as we already do `InstallMethod(MatElm, ..., MAT_ELM_GF2MAT);` later in the file.

hpcgap/lib/vecmat.gi
```diff
@@ -610,13 +605,6 @@ InstallOtherMethod( \[\]\:\=,
       IsObject ],
     ASS_GF2MAT );
 
-InstallMethod( \[\]\:\=,
-    "for GF2 matrix",
-    [ IsGF2MatrixRep,
-      IsPosInt, IsPosInt,
-      IsObject ],
-    SET_MAT_ELM_GF2MAT );
```
fingolfin: This is now redundant as we already do `InstallMethod(SetMatElm, ..., SET_MAT_ELM_GF2MAT);` later in the file.

lib/mat8bit.gi
```diff
@@ -88,7 +88,7 @@ InstallOtherMethod( \[\],  "For a compressed MatFFE",
 #M  <mat> [ <pos1>, <pos2> ]
 ##
 
-InstallMethod( \[\],  "For a compressed MatFFE",
+InstallMethod( \[\,\],  "For a compressed MatFFE",
```
fingolfin: We could of course also use `MatElm` here or `ELM_MAT`... *shrug*

lib/matobjplist.gi
```diff
@@ -826,19 +826,6 @@ InstallMethod( SetMatElm, "for a plist matrix, two positions, and an object",
...
-  [ IsPlistMatrixRep, IsPosInt, IsPosInt ],
-  function( m, row, col )
-    return m![ROWSPOS][row]![ELSPOS][col];
-  end );
-
-InstallMethod( \[\]\:\=, "for a plist matrix, two positions, and an object",
-  [ IsPlistMatrixRep and IsMutable, IsPosInt, IsPosInt, IsObject ],
-  function( m, row, col, ob )
-    m![ROWSPOS][row]![ELSPOS][col] := ob;
-  end );
```
fingolfin: Redundant, as we already (resp.: still) install identical methods for `MatElm` and `SetMatElm` above

fingolfin:
One question to ask here is: What about ISB_MAT and UNB_MAT? Should they also become kernel operations, with aliases `IsBound\[\,\]` and `Unbind\[\,\]`?

Thing is, I am not convinced these operations are useful at all. I wouldn't include them in the MatrixObj specification at all. I am in fact tempted to deprecate the syntax `IsBound(mat[i,j])` and `Unbind(mat[i,j])`, or even outright abolish it. I just can't think of any good use for matrices; and for more general tables, well, you still can do Unbind(table[i][j])` ... Thoughts, comments?

lib/mat8bit.gi
```diff
@@ -111,7 +111,7 @@ InstallOtherMethod( \[\]\:\=,  "For a compressed MatFE",
 #M  <mat> [ <pos1>, <pos2> ] := <val>
 ##
 
-InstallMethod( \[\]\:\=,  "For a compressed MatFE",
+InstallMethod( \[\,\]\:\=,  "For a compressed MatFE",
```
ThomasBreuer: write `"For a compressed MatFFE"` (occurs twice)

fingolfin: Done.

fingolfin:
@ThomasBreuer re GAPJulia: ah good -- I didn't check that. I did, however, check `MethodsOperation` after doing a `LoadAllPackage` with the standard set of distributed packages. But of course that misses any packages which are not distributed, as well as not yet released versions.

PR #3556, opened 2019-07-11
Make more lib/matrix.gi code compatible with MatrixObj

fingolfin:
In particular powering of MatrixObjs now is fast, and we have generic methods for `IsDiagonalMat`, `IsUpperTriangularMat`, `IsLowerTriangularMat`

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -160,12 +160,11 @@ InstallMethod( IsGeneralizedCartanMatrix,
 ##
 #M  IsDiagonalMat(<mat>)
 ##
-InstallOtherMethod( IsDiagonalMat,
+InstallMethod( IsDiagonalMat,
     "for a matrix",
-    [ IsMatrix ],
+    [ IsMatrixObj ],
     function( mat )
-   local  i, j,z;
-    if IsEmpty(mat) then return true;fi;
+    local  i, j, z;
     z:=ZeroOfBaseDomain(mat);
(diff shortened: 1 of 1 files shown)


fingolfin:
True... actually, that made me wonder: Can non-square matrices to be diagonal or triangular? If so, what is the definition? For triangular, it seems our code is a bit asymmetric; e.g. `IsUpperTriangularMat` accepts any zero matrix provided its number of columns is >= the number of rows minus 1. Which seems rather arbitrary.

ThomasBreuer:
Concerning the *definitions*, the Reference Manual says:

`IsDiagonalMat( mat )` returns `true` if `mat` has only zero entries off the main diagonal, `false` otherwise.

`IsUpperTriangularMat( mat )` returns  `true` if `mat` has only zero entries below the main diagonal, `false` otherwise.

`IsLowerTriangularMat( mat )` returns `true` if `mat` has only zero entries below the main diagonal, `false` otherwise.

(Of course the second "below" has to be replaced by "above".)

Concerning the *code*, `IsUpperTriangularMat` runs into problems if the matrix has too few columns.

I will provide a pull request (also for correcting some more details of the documentation).

fingolfin:
I already provided such a pull request yesterday

fingolfin:
That is, PR #3559 fixes `IsUpperTriangularMat`, but I did not make any correction to the documentation.

PR #3558, opened 2019-07-11
 Remove some redundant methods for IsPlistMatrixRep objects

fingolfin:


fingolfin:
In fact they are arguably wrong, as the documentation does not state that only square matrices are supported...

wilfwilson:
Good point.

PR #3559, opened 2019-07-11
Fix IsUpperTriangularMat for non-square matrices

fingolfin:
Also add some tests.

### DIFF
--- a/tst/testinstall/matrix.tst
+++ b/tst/testinstall/matrix.tst
@@ -1,0 +1,47 @@
+#
+gap> IsDiagonalMat(NullMat(3, 3));
+true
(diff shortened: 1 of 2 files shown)


tst/testinstall/matrix.tst
```diff
@@ -0,0 +1,47 @@
+#
+gap> IsDiagonalMat(NullMat(3, 3));
+true
+gap> IsDiagonalMat(NullMat(1, 3));
+true
```
PaulaHaehndel: Why do we want this to be this way?

PaulaHaehndel: After consulting the manual again it is at least documented, including a reason.

Issue #1371, opened 2017-05-30
Problem when an attribute storing object is mutable.


fingolfin:
While looking at the MatrixObj code, we realized that it is in principle possible that an attribute storing object is mutable.

So if you had a matrix object which is mutable, and also attribute storing, and if `RankMat` is an attribute, then we would store the rank; yet if you later modify the matrix (thus possibly changing the rank), this would not be reflected by the attribute.

It is not quit clear to us how "likely" or easy it is to get a mutable attribute storing object, short of using `SetFilterObj(obj, IsMutable)`; still, there is some concern here about accidental bugs.

So perhaps the default setter code for `IsAttributeStoringRep` could at least check whether the `IsMutable` filter is set, and if so, either print a warning, or error out.

This is not currently done, as visible in this example:
```
gap> G:=Group(  (1,2,3), (1,2) );;
gap> HasSize(G);
false
gap> SetFilterObj(G, IsMutable);
gap> IsMutable(G);
true
gap> Size(G);
6
gap> HasSize(G);
true
```

Note that we cannot just change `IsAttributeStoringRep` to require immutability, because there is no filter `IsImmutable`, just a filter `IsMutable`.

stevelinton:
It is possible in principle that one might have a Mutable object that could still safely store some attributes, because none of the operations that mutate it can change the values of those attributes, but this should be done with explicit setter methods for those attributes.  The generic behaviour introduced by IsAttributeStoringRep is, I think, always the wrong thing for mutable objects. 

I think the best solution would be for the default system setter to simply do nothing when the object is mutable. That way you could create an object with IsMutable and IsAttributeStoringRep  and if it were later made Immutable, it would immediately start storing attributes.

stevelinton:
> I think the best solution would be for the default system setter to simply do nothing when the object is mutable. That way you could create an object with IsMutable and IsAttributeStoringRep and if it were later made Immutable, it would immediately start storing attributes.

I tried this, and it causes numerous test failures. I dug into the first one, and the problem essentially is that when creating a mutable Presentation (for Tietze) the code calls SetOne. (tietze.gi line 381). Since the object has IsPresentationDefaultRep which implies IsAttributeStoringRep the value is stored and later relied upon.

The programmer knows that this is safe, because although the presentation may mutate, the stored attribute value will remain correct. 

[...]

PR #3560, opened 2019-07-12
some more fixes and changes for matrix(obj) functions

ThomasBreuer:
- renamed `IsDiagonalMat`, `IsUpperTriangularMat`, `IsLowerTriangularMat`,
  `DiagonalOfMat`
  to `IsDiagonalMatrix`, `IsUpperTriangularMatrix`, `IsLowerTriangularMatrix`,
  `DiagonalOfMatrix`;
  the old names are available via synonyms

  (This is how I interpret the "new plan" statement in `matobj2.gd`.
  Or should the old names better just be kept but become undocumented?
  In that case, the declarations of synonyms should be moved to
  `obsolete.gd`, and test files and cross-references in the documentation
  must be adjusted.)

- in the declarations of `IsDiagonalMatrix`, `IsUpperTriangularMatrix`,
  `IsLowerTriangularMatrix`,
  expect the first argument to be in `IsMatrixObj` not in `IsMatrix`

- fixed wrong documentation of `IsLowerTriangularMatrix`

- fixed GAPDoc syntax for `IsDiagonalMatrix`,
  `IsUpperTriangularMatrix`, `IsLowerTriangularMatrix`

- fixed `UpperSubdiagonal` for non-square matrices, added manual examples

(Concerning "release notes: not needed":
I regard this pull request as a test case for more changes for matrix objects.
In the end, the release notes should mention the relevant changes as a whole.)

### DIFF
--- a/lib/matrix.gd
+++ b/lib/matrix.gd
@@ -64,13 +64,13 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
 
 #############################################################################
 ##
-#O  IsDiagonalMatrix( <mat> )
-#O  IsDiagonalMat( <mat> )
+#P  IsDiagonalMatrix( <mat> )
+#P  IsDiagonalMat( <mat> )
 ##
 ##  <#GAPDoc Label="IsDiagonalMat">
 ##  <ManSection>
-##  <Oper Name="IsDiagonalMatrix" Arg='mat'/>
-##  <Oper Name="IsDiagonalMat" Arg='mat'/>
+##  <Prop Name="IsDiagonalMatrix" Arg='mat'/>
+##  <Prop Name="IsDiagonalMat" Arg='mat'/>
 ##
 ##  <Description>
 ##  return <K>true</K> if the matrix <A>mat</A> has only zero entries
@@ -87,20 +87,20 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareOperation( "IsDiagonalMatrix", [ IsMatrixObj ] );
+DeclareProperty( "IsDiagonalMatrix", IsMatrixObj );
 
 DeclareSynonym( "IsDiagonalMat", IsDiagonalMatrix );
 
 
 #############################################################################
 ##
-#O  IsUpperTriangularMatrix( <mat> )
-#O  IsUpperTriangularMat( <mat> )
+#P  IsUpperTriangularMatrix( <mat> )
+#P  IsUpperTriangularMat( <mat> )
 ##
 ##  <#GAPDoc Label="IsUpperTriangularMat">
 ##  <ManSection>
-##  <Oper Name="IsUpperTriangularMatrix" Arg='mat'/>
-##  <Oper Name="IsUpperTriangularMat" Arg='mat'/>
+##  <Prop Name="IsUpperTriangularMatrix" Arg='mat'/>
+##  <Prop Name="IsUpperTriangularMat" Arg='mat'/>
 ##
 ##  <Description>
 ##  return <K>true</K> if the matrix <A>mat</A> has only zero entries below
@@ -117,20 +117,20 @@ DeclareSynonym( "IsDiagonalMat", IsDiagonalMatrix );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareOperation( "IsUpperTriangularMatrix", [ IsMatrixObj ] );
+DeclareProperty( "IsUpperTriangularMatrix", IsMatrixObj );
 
 DeclareSynonym( "IsUpperTriangularMat", IsUpperTriangularMatrix );
 
 
 #############################################################################
 ##
-#O  IsLowerTriangularMatrix( <mat> )
-#O  IsLowerTriangularMat( <mat> )
+#P  IsLowerTriangularMatrix( <mat> )
+#P  IsLowerTriangularMat( <mat> )
 ##
 ##  <#GAPDoc Label="IsLowerTriangularMat">
 ##  <ManSection>
-##  <Oper Name="IsLowerTriangularMatrix" Arg='mat'/>
-##  <Oper Name="IsLowerTriangularMat" Arg='mat'/>
+##  <Prop Name="IsLowerTriangularMatrix" Arg='mat'/>
+##  <Prop Name="IsLowerTriangularMat" Arg='mat'/>
 ##
 ##  <Description>
 ##  return <K>true</K> if the matrix <A>mat</A> has only zero entries above
@@ -147,7 +147,7 @@ DeclareSynonym( "IsUpperTriangularMat", IsUpperTriangularMatrix );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareOperation( "IsLowerTriangularMatrix", [ IsMatrixObj ] );
+DeclareProperty( "IsLowerTriangularMatrix", IsMatrixObj );
 
 DeclareSynonym( "IsLowerTriangularMat", IsLowerTriangularMatrix );
 
--- a/lib/matrix.gd
+++ b/lib/matrix.gd
@@ -64,77 +64,120 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
 
 #############################################################################
 ##
+#O  IsDiagonalMatrix( <mat> )
 #O  IsDiagonalMat( <mat> )
 ##
 ##  <#GAPDoc Label="IsDiagonalMat">
 ##  <ManSection>
+##  <Oper Name="IsDiagonalMatrix" Arg='mat'/>
 ##  <Oper Name="IsDiagonalMat" Arg='mat'/>
 ##
 ##  <Description>
-##  returns true if mat has only zero entries off the main diagonal, false
-##  otherwise.
+##  return <K>true</K> if the matrix <A>mat</A> has only zero entries
+##  off the main diagonal, and <K>false</K> otherwise.
+##  <Example><![CDATA[
+##  gap> IsDiagonalMatrix( [ [ 1 ] ] );
+##  true
+##  gap> IsDiagonalMatrix( [ [ 1, 0, 0 ], [ 0, 1, 0 ] ] );
+##  true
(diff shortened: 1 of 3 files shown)


lib/matrix.gd
```diff
@@ -64,77 +64,120 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
...
+##  false
+##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareOperation("IsDiagonalMat",[IsMatrix]);
+DeclareOperation( "IsDiagonalMatrix", [ IsMatrixObj ] );
+
+DeclareSynonym( "IsDiagonalMat", IsDiagonalMatrix );
```
fingolfin: Perhaps also remove the declarations in `lib/matobj2.gd` around line 1130 ?

fingolfin:
I am guessing that one problem with this PR is that also homalg defines `IsDiagonalMatrix` etc.; and in fact defines them as a property, not an operation (which has some merit, I think... we probably should make them properties, too, no?)) I think we'll have to carefully coordinate with @mohamed-barakat and the rest of the homalg team if we want that.

I would suggest that we stick with the old names for now, and defer introducing the new names to a later date, and in the meantime, "somebody" should submit PRs to homalg to ensure their declarations only take affect if the GAP library does not introduce them. Then, once a homalg releae with these changes has been released, and added to the package distribution, we can revisit this change.

ThomasBreuer:
@fingolfin Thanks for your thoughts. I will change the pull request for the moment.

Besides that:
Yes, eventually we want to have properties like `IsDiagonalMatrix` and attributes like `DiagonalOfMatrix` for `IsMatrixObj` objects.
This should be not problematic for immutable objects, but we also want *mutable* `IsMatrixObj` objects.

The following behaviour of GAP does not really support this point of view.
Consider the following session.
```GAP
gap> DeclareProperty( "IsGood", IsObject );
gap> DeclareAttribute( "Good", IsObject );
gap> InstallMethod( IsGood, [ IsObject ], x -> true );
gap> InstallMethod( Good, [ IsObject ], x -> "good" );
gap> type:= NewType( NewFamily( "Fam" ), IsAttributeStoringRep and IsMutable );;
gap> obj:= Objectify( type, rec() );;
gap> HasIsGood( obj );
false
gap> IsGood( obj );
true
gap> HasIsGood( obj );
false
gap> SetIsGood( obj, true );
gap> HasIsGood( obj );
true
gap> HasGood( obj );
false
gap> Good( obj );
"good"
gap> HasGood( obj );
true
```
We see that for mutable objects, even if they are in `IsAttributeStoringRep`, the *property setter* is not called automatically when the property gets called and returns `true`.
On the other hand, *attribute setters* are called automatically for objects in `IsAttributeStoringRep`, no matter whether they are mutable or not.

(See the code in `src/opers.c` for the reason: `DoProperty` checks for mutability, `DoAttribute` doesn't.)

Isn't this inconsistent?

With the current behaviour, it would be safe to turn non-properties like `IsDiagonalMat` into properties,
but turning `DiagonalOfMat(rix)` into an attribute would cause problems.

stevelinton:
@ThomasBreuer 

> We see that for mutable objects, even if they are in IsAttributeStoringRep, the property setter is not called automatically when the property gets called and returns true.
On the other hand, attribute setters are called automatically for objects in IsAttributeStoringRep, no matter whether they are mutable or not.
Isn't this inconsistent?

I would say *yes*.  I wonder if the original idea was simply never to have mutable objects in `IsAttributeStoringRep`?  But once we had `MakeImmutable` it made sense to have a mutable object that might start storing attributes when it became immutable (which is basically what we want here).  Maybe we should open this as a separate issue? Also do some experiments and see what happens if we add the test in the kernel.

stevelinton:
Incidentally, it would be nice to have `IsNormedMatrix` (first non-zero entry in each row if any is a 1 and they occur in different columns for the non-zero rows) and `IsSemiEchelonMatrix` (same but the columns are also ordered left-to-right as you read down the matrix and the zero rows are at the bottom). These concepts are used in the meataxe interface.

ThomasBreuer:
@stevelinton Thanks for your remarks. I will open a separate issue, as proposed. Note that issue #1371 shows a similar discussion.

stevelinton:
@ThomasBreuer This is exactly issue #1371, isn't it. Maybe we should just return to that one.

ThomasBreuer:
@fingolfin I have updated my pull request according to your comments.
(I will come back to the originally proposed renamings as soon as #3588 or #3592 gets merged.)

lib/matrix.gd
```diff
@@ -71,13 +71,22 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
 ##  <Oper Name="IsDiagonalMat" Arg='mat'/>
 ##
 ##  <Description>
-##  returns true if mat has only zero entries off the main diagonal, false
-##  otherwise.
+##  return <K>true</K> if the matrix <A>mat</A> has only zero entries
```
fingolfin:
```suggestion
##  returns <K>true</K> if the matrix <A>mat</A> has only zero entries
```
We use "returns" not "return" in the vast majority of entries; I always read this as "IsDiagonalMat returns true...". One could also write "Return true" (with an upper case "R"), I guess, if one does not want to indicate that this continues a sentence that started elsewhere. But then the text is addressing the function ("ordering" it to do something), not the reader; which I find a bit odd for documentation (it's fine by me in code comments).

lib/matrix.gd
```diff
@@ -71,13 +71,22 @@ DeclareProperty( "IsGeneralizedCartanMatrix", IsMatrix );
...
+##  off the main diagonal, and <K>false</K> otherwise.
```
fingolfin: Not an issue of this PR, as the text is old, but: I always found this text confusing and usually have to read it at least twice. And in light of the possibility of unbound entries in a classical matrix (list of lists), one could even misinterpret this as saying: The diagonal entries are all unbound, and any bound entry outside the diagonal must be zero. Which of course is nonsense, nobody would read it that way, but still, it irks me.

How about this instead: "returns true if all off-diagonal entries of mat are zero, and false otherwise." ?

lib/matrix.gd
```diff
@@ -88,13 +97,22 @@ DeclareOperation("IsDiagonalMat",[IsMatrix]);
 ##  <Oper Name="IsUpperTriangularMat" Arg='mat'/>
 ##
 ##  <Description>
-##  returns true if mat has only zero entries below the main diagonal, false
-##  otherwise.
+##  return <K>true</K> if the matrix <A>mat</A> has only zero entries below
```
fingolfin:
```suggestion
##  returns <K>true</K> if the matrix <A>mat</A> has only zero entries below
```

lib/matrix.gd
```diff
@@ -88,13 +97,22 @@ DeclareOperation("IsDiagonalMat",[IsMatrix]);
...
+##  the main diagonal, and <K>false</K> otherwise.
```
fingolfin: Same gripe as above. Since it is not your "fault", I am not expecting you do anything about it... still, I'd find this text better:
> returns true if all entries of <mat> below the main diagonal are zero, and false otherwise.

And of course similarly for the `IsLowerTriangularMat` command.

lib/matrix.gd
```diff
@@ -105,36 +123,47 @@ DeclareOperation("IsUpperTriangularMat",[IsMatrix]);
 ##  <Oper Name="IsLowerTriangularMat" Arg='mat'/>
 ##
 ##  <Description>
-##  returns true if mat has only zero entries below the main diagonal, false
-##  otherwise.
+##  return <K>true</K> if the matrix <A>mat</A> has only zero entries above
```
fingolfin:
```suggestion
##  returns <K>true</K> if the matrix <A>mat</A> has only zero entries above
```

lib/matrix.gd
```diff
@@ -105,36 +123,47 @@ DeclareOperation("IsUpperTriangularMat",[IsMatrix]);
...
-#O  DiagonalOfMat( <mat> )
+#F  DiagonalOfMat( <mat> )
 ##
 ##  <#GAPDoc Label="DiagonalOfMat">
 ##  <ManSection>
 ##  <Func Name="DiagonalOfMat" Arg='mat'/>
 ##
 ##  <Description>
-##  returns the diagonal of the matrix <A>mat</A>. If <A>mat</A> is not a
+##  return the diagonal of the matrix <A>mat</A>. If <A>mat</A> is not a
```
fingolfin:
```suggestion
##  returns the diagonal of the matrix <A>mat</A>. If <A>mat</A> is not a
```

lib/matrix.gd
```diff
@@ -1221,16 +1250,20 @@ DeclareOperation( "TriangulizeMat", [ IsMatrix and IsMutable ] );
 ##
 ##  <Description>
 ##  returns a mutable list containing the entries of the <A>pos</A>th upper
-##  subdiagonal of <A>mat</A>.
+##  subdiagonal of the matrix <A>mat</A>.
```
fingolfin: Offtopic (and not something I'd expect this PR to address!), but: what is a "subdiagonal"? Do we define this anywhere? I wouldn't be able to figure out what this function does from just that text, I'd have to look at examples (luckily there are some) and possibly the code.

And why don't we offer a "LowerSubdiagonal" command? (Or do we? If so, shouldn't its text be updated, too?)

stevelinton:
#1371 is now resolved by #3592

wilfwilson:
I agree with @fingolfin about turning 'return' into 'returns' at various places.

As @stevelinton said, since #3592 is now merged, you can "come back to the originally proposed renamings" @ThomasBreuer.

ThomasBreuer:
Concerning the comments by @fingolfin and @wilfwilson :
- The "return" vs. "returns" problem arose because with the originally proposed renamings, there are *two* operations in the manual sections in question, and then "return" is correct.
- The whole manual chapter on matrices has to be revised, which concerns not only the text but also the examples. I do not think that local changes make sense, except for bugfixes.
- Apparently `LowerSubdiagonal` has never been asked for before, and I do not think that `LowerSubdiagonal` or `UpperDiagonal` are something valuable.
  The point is that `UpperSubdiagonal` is documented, and then the code should at least be correct.

I have now tried to take out the commit 5fd922c , i.e., return to the original version of this pull request.

wilfwilson:
Okay, fine about the return vs returns.

PR #3564, opened 2019-07-14
Add WhereWithVars - a more verbose Where

ChrisJefferson:
# Description

This adds an extended version of `Where`, called `WhereWithVars`, which adds the values of all arguments and locals to the backtrace given in a breakloop. This is a tidied up version of some code I've had use privately for a while.

I didn't just add an argument to 'Where', as in some places 'Where' is replaced.

## Text for release notes 

Add 'WhereWithVars',  an extended version of `Where` which prints the values of all arguments and locals.

### DIFF
--- a/doc/ref/mloop.xml
+++ b/doc/ref/mloop.xml
@@ -739,14 +739,16 @@ is an example in the &GAP; code where the idea is actually used.
 <P/>
 <ManSection>
 <Func Name="Where" Arg='nr'/>
-
+<Func Name="WhereWithVars" Arg='nr'/>
 <Description>
 <Index Subkey="GAP3 name for Where">Backtrace</Index>
 <Index>Stack trace</Index>
 shows the last <A>nr</A> commands on the execution stack during whose execution
 the error occurred. If not given, <A>nr</A> defaults to 5. (Assume, for the
 following example, that after the last example <Ref Func="OnBreak"/>
-has been set back to its default value.)
+has been set back to its default value.). <Ref Func="WhereWithVars"/> acts the
+same as <Ref Func="Where"/> while also showing the arguments and local
+variables of each function.
 <P/>
 <Log><![CDATA[
 gap> StabChain(SymmetricGroup(100)); # After this we typed ^C  
--- a/lib/error.g
+++ b/lib/error.g
@@ -59,8 +59,49 @@ end);
 
 ErrorLVars := fail;
 
-BIND_GLOBAL("WHERE", function( context, depth, outercontext)
-    local   bottom,  lastcontext,  f;
+# In this method the line hint changes in every PrintTo are balanced, because at the moment
+# PrintTo(ERROR_OUTPUT,...) resets the indentation level every time it is called.
+# If/when this is fixed, the indendation in this function could be simplified
+BIND_GLOBAL("PRETTY_PRINT_VARS", function(context)
+    local vars, i, argcount, val;
+    vars := ContentsLVars(context);
+    if not IsRecord(vars) then
+        return;
+    fi;
+
+    argcount := ABS_RAT(NumberArgumentsFunction(vars.func));
+    PrintTo(ERROR_OUTPUT, "\>\>\n\<\<arguments:");
+    if argcount = 0 then
+        PrintTo(ERROR_OUTPUT, " <none>");
+    else
+        for i in [1..argcount] do
+            if IsBound(vars.values[i]) then
+                val := vars.values[i];
+            else
+                val := "<unassigned>";
+            fi;
+            PrintTo(ERROR_OUTPUT, "\>\>\>\>\n", vars.names[i], " :=\>\> ", val, "\<\<\<\<\<\<");
+        od;
+    fi;
+
+    PrintTo(ERROR_OUTPUT, "\>\>\n\<\<local variables:");
+    if argcount = Length(vars.names) then
+        PrintTo(ERROR_OUTPUT, " <none>");
+    else
+        for i in [argcount+1..Length(vars.names)] do
+            if IsBound(vars.values[i]) then
+                val := vars.values[i];
+            else
+                val := "<unassigned>";
+            fi;
+            PrintTo(ERROR_OUTPUT, "\>\>\>\>\n", vars.names[i], " :=\>\> ", val, "\<\<\<\<\<\<");
+        od;
+    fi;
+    PrintTo(ERROR_OUTPUT,"\n");
+end);
+
+BIND_GLOBAL("WHERE", function(context, depth, outercontext, showlocals)
+    local bottom, lastcontext, f, args;
     if depth <= 0 then
         return;
     fi;
@@ -68,6 +109,10 @@ BIND_GLOBAL("WHERE", function( context, depth, outercontext)
     lastcontext := outercontext;
     while depth > 0  and context <> bottom do
         PRINT_CURRENT_STATEMENT(ERROR_OUTPUT, context);
+        if showlocals then
+            PRETTY_PRINT_VARS(context);
+        fi;
+
         PrintTo(ERROR_OUTPUT, " called from\n");
         lastcontext := context;
         context := ParentLVars(context);
@@ -83,6 +128,26 @@ BIND_GLOBAL("WHERE", function( context, depth, outercontext)
 end);
 
 
+BIND_GLOBAL("WHERE_INTERNAL", function(depth, showlocals)
+    if ErrorLVars = fail or ErrorLVars = GetBottomLVars() then
+        PrintTo(ERROR_OUTPUT, "not in any function ");
+    else
+        WHERE(ParentLVars(ErrorLVars), depth, ErrorLVars, showlocals);
+    fi;
+    PrintTo(ERROR_OUTPUT, "at ", INPUT_FILENAME(), ":", INPUT_LINENUMBER(), "\n");
+end);
+
+BIND_GLOBAL("WhereWithVars", function(arg)
+    local   depth;
+    if LEN_LIST(arg) = 0 then
+        depth := 5;
+    else
+        depth := arg[1];
+    fi;
+
+    WHERE_INTERNAL(depth, true);
+end);
+
 BIND_GLOBAL("Where", function(arg)
     local   depth;
     if LEN_LIST(arg) = 0 then
@@ -90,14 +155,10 @@ BIND_GLOBAL("Where", function(arg)
     else
         depth := arg[1];
     fi;
-    
-    if ErrorLVars = fail or ErrorLVars = GetBottomLVars() then
-        PrintTo(ERROR_OUTPUT, "not in any function ");
-    else
-        WHERE(ParentLVars(ErrorLVars),depth, ErrorLVars);
-    fi;
-    PrintTo(ERROR_OUTPUT, "at ",INPUT_FILENAME(),":",INPUT_LINENUMBER(),"\n");
+
+    WHERE_INTERNAL(depth, false);
 end);
+
 
 OnBreak := Where;
 
--- a/tst/testspecial/backtrace.g.out
+++ b/tst/testspecial/backtrace.g.out
@@ -15,6 +15,14 @@ brk> Where();
 l[[ 1 .. 3 ]] := 1; at *stdin*:5 called from
 <function "f">( <arguments> )
  called from read-eval loop at *errin*:1
+brk> WhereWithVars();
+l[[ 1 .. 3 ]] := 1; at *stdin*:5
+  arguments: <none>
+  local variables:
+    l := [ 0, 0, 0, 0, 0, 0 ]
+ called from
+<function "f">( <arguments> )
+ called from read-eval loop at *errin*:2
 brk> quit;
 gap> 
 gap> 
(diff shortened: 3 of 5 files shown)


tst/testspecial/backtrace.g.out
```diff
@@ -109,6 +135,19 @@ for i in 1 do
...
+for i in 1 do
+    return 1;
+od; at *stdin*:24
+with local state:
+-Arguments:
+-Locals:
+   i: <unassigned>
+
+
+ called from
```
fingolfin: I think I'd prefer if there was only one or even no empty line above this, and instead some more indentation. But of course that's subjective and a matter of taste.

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
 
 ErrorLVars := fail;
 
-BIND_GLOBAL("WHERE", function( context, depth, outercontext)
-    local   bottom,  lastcontext,  f;
+BIND_GLOBAL("PRETTY_PRINT_VARS", function(context)
+    local vars, i;
+    vars := ContentsLVars(context);
+    if IsRecord(vars) then
```
fingolfin: A matter of taste, but there'd be less need for indentation if you used `if not IsRecord(vars) then return; fi;`.

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
+        PrintTo(ERROR_OUTPUT, "-Arguments:\n");
```
fingolfin:
```suggestion
        PrintTo(ERROR_OUTPUT, " - arguments:\n");
```

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
+        for i in [1..ABS_RAT(NumberArgumentsFunction(vars.func))] do
+            PrintTo(ERROR_OUTPUT, "   ",vars.names[i],": ");
+            if IsBound(vars.values[i]) then
+                PrintTo(ERROR_OUTPUT, vars.values[i], "\n");
+            else
+                PrintTo(ERROR_OUTPUT, "<unassigned>\n");
+            fi;
+        od;
+        PrintTo(ERROR_OUTPUT, "-Locals:\n");
```
fingolfin:
```suggestion
        PrintTo(ERROR_OUTPUT, " - locals:\n");
```

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
 ErrorLVars := fail;
 
-BIND_GLOBAL("WHERE", function( context, depth, outercontext)
-    local   bottom,  lastcontext,  f;
+BIND_GLOBAL("PRETTY_PRINT_VARS", function(context)
+    local vars, i;
+    vars := ContentsLVars(context);
+    if IsRecord(vars) then
+        PrintTo(ERROR_OUTPUT, "-Arguments:\n");
+        for i in [1..ABS_RAT(NumberArgumentsFunction(vars.func))] do
```
fingolfin: `ABS_RAT(NumberArgumentsFunction(vars.func))` is used twice, perhaps store it in a local var `numArgs` or so?

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
     if depth <= 0 then
         return;
     fi;
     bottom := GetBottomLVars();
     lastcontext := outercontext;
     while depth > 0  and context <> bottom do
         PRINT_CURRENT_STATEMENT(ERROR_OUTPUT, context);
-        PrintTo(ERROR_OUTPUT, " called from\n");
+        if showlocals then
+            PrintTo(ERROR_OUTPUT,"\nwith local state:\n");
```
fingolfin:
```suggestion
            PrintTo(ERROR_OUTPUT,"\n with local state:\n");
```
To match the indentation of "called from". I wonder if indenting it even more would be even better, though (and then of course also indent the output of `PRETTY_PRINT_VARS` to match)

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
 
-BIND_GLOBAL("WHERE", function( context, depth, outercontext)
-    local   bottom,  lastcontext,  f;
+BIND_GLOBAL("PRETTY_PRINT_VARS", function(context)
+    local vars, i;
+    vars := ContentsLVars(context);
+    if IsRecord(vars) then
+        PrintTo(ERROR_OUTPUT, "-Arguments:\n");
+        for i in [1..ABS_RAT(NumberArgumentsFunction(vars.func))] do
+            PrintTo(ERROR_OUTPUT, "   ",vars.names[i],": ");
```
fingolfin:
```suggestion
            PrintTo(ERROR_OUTPUT, "   ", vars.names[i], ": ");
```

lib/error.g
```diff
@@ -59,16 +59,47 @@ end);
...
+            if IsBound(vars.values[i]) then
+                PrintTo(ERROR_OUTPUT, vars.values[i], "\n");
+            else
+                PrintTo(ERROR_OUTPUT, "<unassigned>\n");
+            fi;
+        od;
+        PrintTo(ERROR_OUTPUT, "-Locals:\n");
+        for i in [ABS_RAT(NumberArgumentsFunction(vars.func))+1..Length(vars.names)] do
+            PrintTo(ERROR_OUTPUT, "   ",vars.names[i],": ");
```
fingolfin:
```suggestion
            PrintTo(ERROR_OUTPUT, "   ",vars.names[i], ": ");
```

doc/ref/mloop.xml
```diff
@@ -739,14 +739,15 @@ is an example in the &GAP; code where the idea is actually used.
...
+<Func Name="WhereWithVars" Arg='nr'/>
 <Description>
 <Index Subkey="GAP3 name for Where">Backtrace</Index>
 <Index>Stack trace</Index>
 shows the last <A>nr</A> commands on the execution stack during whose execution
 the error occurred. If not given, <A>nr</A> defaults to 5. (Assume, for the
 following example, that after the last example <Ref Func="OnBreak"/>
-has been set back to its default value.)
+has been set back to its default value.). <Ref Func="WhereWithVars"/> also
+shows the function arguments and local variables in each stack frame
```
fingolfin:
```suggestion
shows the function arguments and local variables in each stack frame.
```
The term "stack frame" is not actually defined in the manual, is it? In that case, it might be better to either define it, or avoid it here.

tst/testspecial/backtrace.g.out
```diff
@@ -109,6 +135,19 @@ for i in 1 do
 od; at *stdin*:24 called from
 <function "f">( <arguments> )
  called from read-eval loop at *errin*:1
+brk> WhereWithVars();
+for i in 1 do
+    return 1;
+od; at *stdin*:24
+with local state:
+-Arguments:
+-Locals:
```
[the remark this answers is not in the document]
ChrisJefferson: I played around a little with the formatting, I'm happy for other suggestions, and couldn't come up with anything I loved.

A nicer handling of 'none' definitely seems like a nice idea.

ChrisJefferson:
i've tried overhauling the output based on printing records. it looks a bit awkward as PrintTo(ERROR_OUTPUT,...) doesn't preserve the indentation level.

tst/testspecial/backtrace.g.out
```diff
@@ -128,47 +163,156 @@ for i in true do
 od; at *stdin*:27 called from
 <function "f">( <arguments> )
  called from read-eval loop at *errin*:1
+brk> WhereWithVars();
+for i in true do
+    return 1;
+od; at *stdin*:27
+  arguments: <none>
+
```
fingolfin: Can we get rid of the empty line here when there are no arguments?

Otherwise: much better :)

doc/ref/mloop.xml
```diff
@@ -739,14 +739,16 @@ is an example in the &GAP; code where the idea is actually used.
...
+<Func Name="WhereWithVars" Arg='nr'/>
 <Description>
 <Index Subkey="GAP3 name for Where">Backtrace</Index>
 <Index>Stack trace</Index>
 shows the last <A>nr</A> commands on the execution stack during whose execution
 the error occurred. If not given, <A>nr</A> defaults to 5. (Assume, for the
 following example, that after the last example <Ref Func="OnBreak"/>
-has been set back to its default value.)
+has been set back to its default value.). <Ref Func="WhereWithVars"/> acts the
+same as <Ref Func="Break"/> while also showing the arguments and local
```
fingolfin: Did you mean this?
```suggestion
same as <Ref Func="Where"/> while also showing the arguments and local
```

lib/error.g
```diff
@@ -59,16 +59,58 @@ end);
...
 ErrorLVars := fail;
 
-BIND_GLOBAL("WHERE", function( context, depth, outercontext)
-    local   bottom,  lastcontext,  f;
+BIND_GLOBAL("PRETTY_PRINT_VARS", function(context)
+    local vars, i, argcount, val;
+    vars := ContentsLVars(context);
+    if IsRecord(vars) then
+        argcount := ABS_RAT(NumberArgumentsFunction(vars.func));
+        PrintTo(ERROR_OUTPUT, "\>\>\n\<\<arguments:");
```
fingolfin: That doesn't make sense to me. Why would you decrease the hint level with `\>\>` at the start, and then after the newline increase it again? Is that to counteract some line break hints in the caller? But that would seem rather fragile, and at least should be documented.

lib/error.g
```diff
@@ -59,16 +59,58 @@ end);
...
+                if IsBound(vars.values[i]) then
+                    val := vars.values[i];
+                else
+                    val := "<unassigned>";
+                fi;
+                # This must all be on one line as PrintTo(ERROR_OUTPUT,...) resets indentation on every call
+                PrintTo(ERROR_OUTPUT, "\>\>\>\>\n", vars.names[i], " :=\>\> ", val, "\<\<\<\<\<\<");
+            od;
+        fi;
+        PrintTo(ERROR_OUTPUT,"\n");
```
fingolfin: This leaves the line break hints unbalanced, which we shouldn't do (see also issue #2677)

lib/error.g
```diff
@@ -59,16 +59,58 @@ end);
...
+    fi;
+end);
+
+BIND_GLOBAL("WHERE", function( context, depth, outercontext, showlocals)
```
fingolfin: Optional suggestion:
```suggestion
BIND_GLOBAL("WHERE", function(context, depth, outercontext, showlocals)
```

lib/error.g
```diff
@@ -59,16 +59,58 @@ end);
...
     bottom := GetBottomLVars();
     lastcontext := outercontext;
     while depth > 0  and context <> bottom do
         PRINT_CURRENT_STATEMENT(ERROR_OUTPUT, context);
-        PrintTo(ERROR_OUTPUT, " called from\n");
+        if showlocals then
+            PRETTY_PRINT_VARS(context);
+        fi;
+
+        PrintTo(ERROR_OUTPUT," called from\n");
```
fingolfin:
```suggestion
        PrintTo(ERROR_OUTPUT, " called from\n");
```
This shrinks the diff

lib/error.g
```diff
@@ -83,22 +125,38 @@ BIND_GLOBAL("WHERE", function( context, depth, outercontext)
 end);
 
 
-BIND_GLOBAL("Where", function(arg)
+BIND_GLOBAL("WHERE_INTERNAL", function(depth, showlocals)
+    if ErrorLVars = fail or ErrorLVars = GetBottomLVars() then
+        PrintTo(ERROR_OUTPUT, "not in any function ");
+    else
+        WHERE(ParentLVars(ErrorLVars),depth, ErrorLVars, showlocals);
```
fingolfin: Optional suggestion
```suggestion
        WHERE(ParentLVars(ErrorLVars), depth, ErrorLVars, showlocals);
```

lib/error.g
```diff
@@ -83,22 +125,38 @@ BIND_GLOBAL("WHERE", function( context, depth, outercontext)
...
+    fi;
+    PrintTo(ERROR_OUTPUT, "at ",INPUT_FILENAME(),":",INPUT_LINENUMBER(),"\n");
```
fingolfin: Optional suggestion
```suggestion
    PrintTo(ERROR_OUTPUT, "at ", INPUT_FILENAME(), ":", INPUT_LINENUMBER(), "\n");
```

lib/error.g
ChrisJefferson: So, \>\> increases the hint level (at least, it causes an indent). The reason I write this strangely is because (as I mention below, but I should pull the comment up) there is a long-standing bug in `PrintTo(ERROR_OUTPUT,` (and a related places in printing) where the indentation level gets reset on each print call.

Therefore I'm making sure I do all the indentation I want in a single call to PrintTo is balanced (which I believe I've achieved, unless I'm missing something), so that at the moment I end up with the output I want, and if PrintTo is ever fixed, the output won't change.

ChrisJefferson: I'll write a longer comment pointing this out at the top of the function

lib/error.g
ChrisJefferson: I think they are balanced? (in fact, every PrintTo is balanced)

lib/error.g
fingolfin: Thanks for the explanation, I mixed up the meaning of `\<` and `\>`.

And: Ah yeah, I remember that bug: the problem is that for each such `PrintTo` call, we "open" stderr, print to it, and close it -- but the output state for the stdout and stderr stream are not shared, but arguably should be, at least if both are known to map to the same device. We probably should have an issue for that (perhaps we already do, I didn't check).

lib/error.g
fingolfin: My apologies, Mi miscounted. This is all fine!

PR #3565, opened 2019-07-14
Allow SyAllocPool to be smaller than SyStorMax

### LINKED ISSUE #3509: GAP can't fork on systems without overcommit
For the purposes of this issue, there are two important memory values in GAP:

* SyAllocPool, set with '-s', sets how much memory space GAP asks the OS for.
* SyStorMax, set with '-o', sets at what amount of memory we warn about excess memory usage and print a warning.

At the moment GAP always sets SyAllocPool is >= SyStorMax.

This gives a problem since we changed SyStorMax to be 3/4 of physical memory, which sets SyAllocPool to 3/4 of physical memory.

This doesn't mean we actually use 3/4 of physical memory, just we tell the OS we might use it. Then when we fork, we are technically using 1.5x physical memory, except again we aren't because (a) most of it is identical and the OS doesn't duplicate identical pages, and (b) most of it is unused.

HOWEVER, you an enable an option which ensures the OS never promises memory it can't give out later, and if you enable that GAP can't fork, which means it can't run external processes.

Travis seems to have some kind of limit against excessive overcommit, from what we can tell, which is causing some travis tests to fail.

My tempted fix to this is to stop making SyAllocPool be >= SyStorMax, and set it back by default to 2g, which was it's default in previous versions. This should, I believe, not make any significant difference to users.

However, I thought I'd write down the issue and see if anyone had any comments.

ChrisJefferson:
Fixes #3509 

In practice, this sets SyAllocPool back to the value it had before I recently increased SyStorMax. This should make no difference to 64-bit users, except GAP will appear to reserve less memory up front.

These two variables handle difference cases -- SyStorMax
sets the amount of memory at which a warning is printed,
while SyAllocPool sets how much memory we ask the OS
for at startup. If SyAllocPool is more than half of
physical memory, then we cannot fork on machines with
overcommit turned off.

A low SyAllocPool can only cause problems of GASMAN
tries to allocate more memory and fails. This should
never be a problem on 64-bit OSes as we place memory
at a high memory address

This (in my opinion) doesn't need adding to the release notes, as it is part of the earlier "increase GAP's default memory limit", and trying to explain exactly how the various ways GAP measures memory interact is understood by very few people.

### DIFF
--- a/src/sysmem.h
+++ b/src/sysmem.h
@@ -66,6 +66,20 @@ extern Int SyStorMin;
 /****************************************************************************
 **
 *V  SyAllocPool
+**
+**  'SyAllocPool' is the size of the OS memory block which Gasman is using
+**  to store its workspace.
+**
+**  Gasman's workspace must be a single continuous block, and can only be
+**  extended. Extending this memory block after GAP has been running for a
+**  while requires the OS does not allocate any memory immediately after the
+**  current location of the workspace. On 64-bit systems using mmap this is
+**  usually possible as GAP's workspace starts at memory location 16TB. On
+**  32-bit systems it may not be possible to extend so this acts as the
+**  maximum workspace size. The main reason extending the workspace on 32-bit
+**  fails is if code calls malloc (or new in C++), which many packages do.
+**
+**  This option can be changed with -s.
 */
 extern UInt SyAllocPool;
 
--- a/src/system.c
+++ b/src/system.c
@@ -1188,10 +1188,6 @@ void InitSystem (
                             SyAllocPool > 1024 * SyStorKill ) {
         SyAllocPool = SyStorKill * 1024;
     }
-    /* fix pool size if it is given and lower than SyStorMax */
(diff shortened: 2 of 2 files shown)


ChrisJefferson:
I did add a comment, seeing as I bothered to figure out what this is.

Basically, it's the amount of space we mmap.

src/sysmem.h
```diff
@@ -66,6 +66,19 @@ extern Int SyStorMin;
...
 **
 *V  SyAllocPool
+**
+**  'SyAllocPool' is the size of the OS memory block which Gasman is using
+**  to store its workspace.
+**
+**  Gasman's workspace must be a single continuous block, and can only be
+**  extended. Therefore extending this after GAP has been executing requires
+**  the OS does not allocate any memory immediately after the existing
+**  workspace. On 64-bit systems sing mmap this is usually possible as
```
wilfwilson: Is 'sing' meant to be 'using' here? Otherwise I'm not sure what it means.

ChrisJefferson: yup, thanks

PR #3566, opened 2019-07-14
Special case identity permutations and improve printing of permutations

### LINKED ISSUE #3547: Special casing identity in permutation operations
As an experiment, I decided to check how many of the times someone operates on permutations, one of the permutations is the identity, when running testinstall

p*q : 19% of 3410082
Quo: 6% of 690495
Conj: 5% of 30276
p^i for integer i: 2% of 53155

So, given that I think it's certainly worth special casing p*q to handle p or q being the identity. It's probably worth it for quo and conj as well, as we have calculate the length of both permutations, so all we are doing is a comparison on something we already have in a register anyway.

ChrisJefferson:
Resolves #3547 

This adds special cases for identity permutations in various methods. These do get called fairly often, and are very cheap.

EDIT: This now fixes another problem found along the way. GAP has the internal concept of the 'degree' of a permutation, which can be larger than the LargestMovedPoint. For example, `p=(1,2)` has DEG_PERM=2, but `q=(2^19,2^20)*(1,2)*(2^19,2^20)` has DEG_PERM=2^20, as GAP doesn't try hard to reduce this value, but just uses a weak guess.

This value is used in printing to decide how many digits we need when printing, so the permutations `p` and `q` print differently, even though from the GAP language there is no way to tell them apart. This could break tests in some packages, who have made permutations with a large DEG_PERM

```
 gap> p := (1,2);
(1,2)
gap> q := (1,2)*(2^19,2^20)*(2^19,2^20);
(1,2)
gap> Print(p);
(1,2)
gap> Print(q);
(    1,    2)
gap> p=q;
true
```

## Text for release notes 

* Operations involving identity permutations optimised.
* Improve printing of permutations


fingolfin:
Is there a strong reason to combine a potentially breaking change with an enhancement in one PR? I think we might want to merge one immediately, but for the other take some more time to first let Alex run all package tests and determine if we break some packages, and then give the package authors a chance to adapt (or even help them fix their packages)?

src/permutat.cc
```diff
@@ -355,6 +359,14 @@ static Obj ProdPerm(Obj opL, Obj opR)
     degP = degL < degR ? degR : degL;
     prd  = NEW_PERM<Res>( degP );
 
+    // Deal with trivial cases
+    if (degL == 0) {
+        return opR;
+    }
+    if (degR == 0) {
+        return opL;
+    }
```
fingolfin: Perhaps move these test *before* creating the now useless temporary object `prd`?

src/permutat.cc
```diff
@@ -424,6 +436,14 @@ static Obj LQuoPerm(Obj opL, Obj opR)
     degM = degL < degR ? degR : degL;
     mod = NEW_PERM<Res>( degM );
 
+    // handle trivial cases
+    if (degL == 0) {
```
fingolfin: As above, I'd move these before the `NEW_PERM` call.

src/permutat.cc
```diff
@@ -860,6 +886,15 @@ static Obj PowPerm(Obj opL, Obj opR)
     degC = degL < degR ? degR : degL;
     cnj = NEW_PERM<Res>( degC );
 
+    // Handle trivial cases
```
fingolfin: As above, I'd move these before the `NEW_PERM` call.

src/permutat.cc
```diff
@@ -860,6 +886,15 @@ static Obj PowPerm(Obj opL, Obj opR)
...
+    if (degL == 0) {
+        return IdentityPerm;
```
fingolfin: This is fine, but I wonder: above in `PowPermInt`, you returned `opL` in a comparable situation; I'd expect the code to do the same in both cases (I don't care which of the two it does, though)

src/permutat.cc
```diff
@@ -909,6 +944,10 @@ static Obj CommPerm(Obj opL, Obj opR)
     degC = degL < degR ? degR : degL;
     com = NEW_PERM<Res>( degC );
 
+    if (degL == 0 || degR == 0) {
```
fingolfin: As above, I'd move these before the `NEW_PERM` call.

tst/testinstall/trans.tst
```diff
@@ -2295,7 +2295,9 @@ gap> (f * (1, 2)) * (1, 2, 3);
 Transformation( [ 3, 2, 1 ] )
 gap> (1, 2, 3) * (f * (1, 2));
 Transformation( [ 1, 3, 2 ] )
-gap> p := AS_PERM_TRANS(f) * (1,2);;
+
+# Large values force this to be a Perm4Rep
+gap> p := AS_PERM_TRANS(f) * (1,2)(2^16,2^16+1) * (2^16,2^16+1);;
```
fingolfin: Why is this change necessary?

Also, I don't really understand what the comment is trying to tell me. Is the *goal* here to force a Perm4Rep, to ensure some tests later on are meaningful? Then perhaps the comment should state it that way, and also give the reason?

ChrisJefferson:
I'll have a rearrange. The two issues ended up merged because I only discovered the problem with printing when testing this.

src/permutat.cc
ChrisJefferson: Yes, the ebeing after the new perm defeats most of the purpose. Should have been more careful.

ChrisJefferson:
So, I looked at separating this, but I think it might be best to test them together, as they can both produce visible differences in tests, so if we separate we just have to make two passes over the packages.

src/permutat.cc
```diff
@@ -857,6 +885,16 @@ static Obj PowPerm(Obj opL, Obj opR)
     /* compute the size of the result and allocate a bag                   */
```
```suggestion
```

src/permutat.cc
```diff
@@ -857,6 +885,16 @@ static Obj PowPerm(Obj opL, Obj opR)
...
+    // Handle trivial cases
+    if (degL == 0) {
+        return IdentityPerm;
+    }
+
+    if (degR == 0) {
+        return opL;
+    }
+
     degC = degL < degR ? degR : degL;
```
```suggestion
    // compute the size of the result and allocate a bag
    degC = degL < degR ? degR : degL;
```

src/permutat.cc
```diff
@@ -349,9 +353,18 @@ static Obj ProdPerm(Obj opL, Obj opR)
     const TR *          ptR;            /* pointer to the right operand    */
     UInt                p;              /* loop variable                   */
 
-    /* compute the size of the result and allocate a bag                   */
     degL = DEG_PERM<TL>(opL);
     degR = DEG_PERM<TR>(opR);
+
+    // Deal with trivial cases
```
```suggestion
    // handle trivial cases
```
I prefer uniformity between the various functions whenever possible, that IMHO highlights the real differences.

src/permutat.cc
```diff
@@ -517,6 +539,12 @@ static Obj PowPermInt(Obj opL, Obj opR)
 
     /* get the operands and allocate a result bag                          */
     deg = DEG_PERM<T>(opL);
+
+    // Handle trivial perm
```
```suggestion
    // handle trivial permutations
```

src/permutat.cc
```diff
@@ -517,6 +539,12 @@ static Obj PowPermInt(Obj opL, Obj opR)
 
     /* get the operands and allocate a result bag                          */
```
```suggestion
```

src/permutat.cc
```diff
@@ -517,6 +539,12 @@ static Obj PowPermInt(Obj opL, Obj opR)
...
     deg = DEG_PERM<T>(opL);
+
+    // Handle trivial perm
+    if (deg == 0) {
+        return IdentityPerm;
+    }
+
     pow = NEW_PERM<T>(deg);
```
```suggestion
    // allocate a result bag
    pow = NEW_PERM<T>(deg);
```

src/permutat.cc
```diff
@@ -857,6 +885,16 @@ static Obj PowPerm(Obj opL, Obj opR)
     /* compute the size of the result and allocate a bag                   */
     degL = DEG_PERM<TL>(opL);
     degR = DEG_PERM<TR>(opR);
+
+    // Handle trivial cases
```
```suggestion
    // handle trivial cases
```

src/permutat.cc
```diff
@@ -903,9 +941,14 @@ static Obj CommPerm(Obj opL, Obj opR)
     const TR *          ptR;            /* pointer to the right operand    */
     UInt                p;              /* loop variable                   */
 
-    /* compute the size of the result and allocate a bag                   */
     degL = DEG_PERM<TL>(opL);
     degR = DEG_PERM<TR>(opR);
+
+    if (degL == 0 || degR == 0) {
```
```suggestion
    // handle trivial cases
    if (degL == 0 || degR == 0) {
```

src/permutat.cc
```diff
@@ -418,9 +431,18 @@ static Obj LQuoPerm(Obj opL, Obj opR)
...
-    /* compute the size of the result and allocate a bag                   */
     degL = DEG_PERM<TL>(opL);
     degR = DEG_PERM<TR>(opR);
+
+    // handle trivial cases
+    if (degL == 0) {
+        return opR;
+    }
+    if (degR == 0) {
+        return INV(opL);
```
fingolfin: We could slightly optimize this to:
```suggestion
        return InvPerm<T>(opL);
```
I think it will also need a declaration for `InvPerm` at the top.

PR #3571, opened 2019-07-14
Fix NrCols for empty matrices in IsMatrix

### LINKED ISSUE #3570: NrCols enters break loop for empty matrix
```
gap> M := EmptyMatrix(2);
EmptyMatrix( 2 )
gap> IsMatrixObj(M);
true
gap> NrRows(M);
0
gap> NrCols(M);
Error, no 1st choice method found for `[]' on 2 arguments at /Users/Wilf/gap/lib/methsel2.g:249 called from
return Length( m[1] ); at /Users/Wilf/gap/lib/matobj.gi:635 called from
<function "NumberColumns for a plist matrix">( <arguments> )
```
This should also give `0`.

wilfwilson:
Closes #3570.

## Text for release notes

Fix `NrCols` and `NumberColumns` for empty matrices in `IsMatrix`.

### DIFF
--- a/lib/algmat.gi
+++ b/lib/algmat.gi
@@ -1474,23 +1474,23 @@ InstallGlobalFunction( EmptyMatrix, function( char )
     fi;
 
     # Construct the matrix.
-    mat:= Objectify( NewType( Fam,
-                                  IsList
-                              and IsEmpty
-                              and IsOrdinaryMatrix
-                              and IsZero
-                              and IsAssociativeElement
-                              and IsCommutativeElement
-                              and IsJacobianElement
-                              and IsAttributeStoringRep ),
-                     rec() );
-
-    SetName( mat, Concatenation( "EmptyMatrix( ", String( char ), " )" ) );
+    mat := ObjectifyWithAttributes( rec(),
+             NewType( Fam,
+                          IsList
+                      and IsEmpty
+                      and IsOrdinaryMatrix
+                      and IsZero
+                      and IsAssociativeElement
+                      and IsCommutativeElement
+                      and IsJacobianElement
+                      and IsAttributeStoringRep ),
+             Name,          Concatenation("EmptyMatrix( ", String(char), " )" ),
+             DimensionsMat, [ 0, 0 ],
+             Length,        0,
+             NrRows,        0,
+             NrCols,        0) ;
     SetInverse( mat, mat );
     SetAdditiveInverse( mat, mat );
-    SetDimensionsMat( mat, [ 0, 0 ] );
-    SetLength( mat, 0 );
-
     return mat;
 end );
 
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -629,10 +629,16 @@ InstallMethod( RowsOfMatrix,
     Immutable );
 
 
-InstallOtherMethod( NumberRows, "for a plist matrix",
+InstallOtherMethod( NumberRows, "for a matrix",
   [ IsMatrix ], Length);
-InstallOtherMethod( NumberColumns, "for a plist matrix",
-  [ IsMatrix ], m -> Length(m[1]) );
+InstallOtherMethod( NumberColumns, "for a matrix",
+  [ IsMatrix ],
+  function(m)
+    if Length(m) = 0 then
+      return 0;
+    fi;
+    return Length(m[1]);
+  end );
 
 #
 # Compatibility code: Generic methods for IsMatrixObj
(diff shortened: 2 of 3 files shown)


lib/matobj.gi
```diff
@@ -632,7 +632,13 @@ InstallMethod( RowsOfMatrix,
 InstallOtherMethod( NumberRows, "for a plist matrix",
   [ IsMatrix ], Length);
 InstallOtherMethod( NumberColumns, "for a plist matrix",
-  [ IsMatrix ], m -> Length(m[1]) );
+  [ IsMatrix ],
+  function(m)
+    if Length(m) = 0 then
+      return 0;
+    fi;
+    return Length(m[1]);
```
fingolfin: This fix and the test don't really fit together: The tests looks at `EmptyMatrix(2)`, but the result of that is *not* a plist, but rather a special type. Take a look at the source of `EmptyMatrix` for details. So we still need a  proper test for this fix (e.g. `NrCols( [] )`), and possibly should also fix / improve the `EmptyMatrix` function (say by adding `SetNumberColumns` and `SetNumberRows` calls to it).

wilfwilson: Oh, I'm confused now. The result of `EmptyMatrix(2)` is not a plist, I believe you, but the method that I changed _is_ the one that applies to it:
```
ApplicableMethod(NrCols, [EmptyMatrix(2)], true);
(...removed...)
4: ``NumberColumns: for a plist matrix'' at /Users/Wilf/gap/lib/matobj.gi:
634, value: 34
#I  Function Body:
function ( m )
    return Length( m[1] );
endfunction( m ) ... end
```
So what's going on here? The string here says `"for a plist matrix"`, but the filter is only for `IsMatrix`. I (as a noob) would expect this to mean that this method is very general and should apply to _any_ matrix (since surely any matrix satisfies `IsMatrix`?). Is it true that for any GAP matrix `m` (i.e. object in `IsMatrix`), either it has `Length(m) = 0`, or `Length(m[1])` gives you the number of columns?

So perhaps this assumption is wrong, or it's correct and the string "for a plist rep" should be made more general.

Or perhaps I don't know enough about matrices in GAP to contribute properly on this topic. In particular, I'm not sure what the correct thing is for me to do here.

(I'm also not sure how I create a "plist" matrix with no rows and columns - matrices in `IsPlistMatrixRep` have their own methods so it can't be one of those.)

fingolfin: So, you are right, that test is indeed valid for this method -- the problem then is (a) the PR description / commit message, and (b) the info string of that method, which claims "for a plist matrix" but really should be "for a matrix".

However, with the MatrixObj project, as a matter fact it will *not* be true that every matrix will lie in the `IsMatrix` filter, because, you see, the definition of `IsMatrix` is too limited:
```gap
DeclareSynonym( "IsMatrix", IsRingElementTable );
```
So a matrix really is a "list of lists containing ring elements". But with MatrixObj, we want to allow for matrices which are not lists. Which is why `IsMatrixObj` was introduced (and similarly `IsVectorObj`). We currently have an implication from `IsMatrix` to `IsMatrixObj`, but not vice versa.

Anyhow, the method you are editing in this PR really makes sense for any `IsMatrix`, but doesn't make sense for general `IsMatrixObj`

lib/algmat.gi
```diff
@@ -1490,6 +1490,8 @@ InstallGlobalFunction( EmptyMatrix, function( char )
     SetAdditiveInverse( mat, mat );
     SetDimensionsMat( mat, [ 0, 0 ] );
     SetLength( mat, 0 );
+    SetNumberRows( mat, 0 );
+    SetNumberColumns( mat, 0 );
```
fingolfin: Thanks. BTW, one could improve this code even more by replacing `Objectify` with `ObjectifyWithAttributes`, which would avoid immediate method overhead for setting all these attributes. Alas, that can be done in a later PR.

lib/matobj.gi
```diff
@@ -632,7 +632,13 @@ InstallMethod( RowsOfMatrix,
 InstallOtherMethod( NumberRows, "for a plist matrix",
```
fingolfin:
```suggestion
InstallOtherMethod( NumberRows, "for a matrix",
```
And then in other methods using `IsMatrixObj`, we could write "for a matrix object"

lib/matobj.gi
fingolfin: BTW, `EmptyMatrix` certainly is a hack itself. With MatrixObj, we can have quite reasonably matrices of dimension 0 x n or n x 0, and even with any specific base domain.

lib/algmat.gi
wilfwilson: I couldn't work out how to do `Set{Additive}Inverse(mat, mat)` inside of an `ObjectifyWithAttributes`, but I did the others.

fingolfin: Beautiful, thank you so much!

PR #3575, opened 2019-07-15
Clear stored inverse when calling TrimPerm

### LINKED ISSUE #3574: Hang at assertion level 2 when computing with a certain group of perms made by PermLeftQuoTransformation
This problem only appears in the GAP `master` branch after https://github.com/gap-system/gap/commit/8c9f355aee4bce42c061d66f779442ee0678082f (I'm not saying that there's anything wrong with that commit, only that that commit brought this problem to light).
```
gap> SetAssertionLevel(2);
gap> p := PermLeftQuoTransformation(IdentityTransformation, Transformation([2, 1]));
(1,2)
gap> G := Group([(1,2,3), p]);; # SymmetricGroup(3)
gap> H := MaximalSubgroupClassReps(G)[2];
Group([ (2,3) ])
gap> Pcgs(H);
[HANG]
```
@ChrisJefferson has been looking into this so far, and has said:
> "It's something deeper and crazy, a perm4 is ended up with a stored perm2 inverse, which breaks lots of things. I don't know how that is happening"

ChrisJefferson:
Fixes #3574 , although this is actually an old bug which was only recently discovered by an interaction between semigroups and a recent change.

This is a minimal fix, so it can be backported (if another release of 4.10 is made). I would like to investigate removing/changing TrimPerm/TRIM_PERM, because it's very easy to abuse and break things, but that's for another PR.

The problem we definitely have: If TrimPerm turns a Perm4 into a Perm2, we need to remove the stored inverse, as a permutation and it's store inverse must have the same TNUM. I think the only place this actually causes a problem is calculating x/p for an integer x and permutation p, after p has been trimmed, but there could be other places.

The reason this PR always throw away the stored inverse: If TrimPerm was used to produce a logically different permutation (say trimming (1,2)(3,4) to just (1,2)), then we also have to remove the inverse. It's unclear to me if that is supposed allowed by TrimPerm. It seems like a terrible idea as permutations are immutable, so I think should be banned.

* Fix bug in calculating x/p for an integer x and permutation p, if p has been 'trimmed'.

### DIFF
--- a/src/permutat.h
+++ b/src/permutat.h
@@ -76,7 +76,12 @@ EXPORT_INLINE const UInt4 * CONST_ADDR_PERM4(Obj perm)
 
 EXPORT_INLINE Obj STOREDINV_PERM(Obj perm)
 {
-    return ADDR_OBJ(perm)[0];
+    Obj inv = ADDR_OBJ(perm)[0];
+    /* Check inv has the same TNAM as perm
+      This is checked below in SET_STOREDINV_PERM, but could
+      be invalidated if either perm or inv is changed in-place */
+    GAP_ASSERT(!inv || (TNUM_OBJ(perm) == TNUM_OBJ(inv)));
+    return inv;
 }
 
 /* SET_STOREDINV_PERM should only be used in neither perm, nor inv has
@@ -95,6 +100,18 @@ EXPORT_INLINE void SET_STOREDINV_PERM(Obj perm, Obj inv)
         CHANGED_BAG(perm);
         ADDR_OBJ(inv)[0] = perm;
         CHANGED_BAG(inv);
+    }
+}
+
+/* Clear the stored inverse. This is required if 'perm' changes TNUM.
+   Also clears the stored inverse of the stored inverse (which should be
+   perm). */
+EXPORT_INLINE void CLEAR_STOREDINV_PERM(Obj perm)
+{
+    Obj inv = ADDR_OBJ(perm)[0];
+    if (inv) {
+        ADDR_OBJ(inv)[0] = 0;
+        ADDR_OBJ(perm)[0] = 0;
     }
 }
 
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -2238,6 +2238,7 @@ Obj Array2Perm (
 
 void TrimPerm(Obj perm, UInt m)
 {
+    CLEAR_STOREDINV_PERM(perm);
     if (TNUM_OBJ(perm) == T_PERM2) {
         GAP_ASSERT(m <= DEG_PERM2(perm));
         ResizeBag(perm, SIZEBAG_PERM2(m));
--- a/tst/testbugfix/2019-07-15-StoredInv.tst
+++ b/tst/testbugfix/2019-07-15-StoredInv.tst
@@ -1,0 +1,40 @@
+# When TRIM_PERM causes a permutation 'p' to change TNAM, the stored inverse
+# must be cleared, as the stored inverse of 'p' must have the
+# same TNAM as 'p'
+
+# Make p a Perm4
+gap> p := (1,2,3,4)*(2^16,2^16+1)*(2^16,2^16+1);
+(1,2,3,4)
+gap> IsPerm4Rep(p);
+true
+
+# Force inverse calculation
+gap> q := p^-1;
+(1,4,3,2)
+gap> IsPerm4Rep(q);
+true
+
+# Now trim
+gap> TRIM_PERM(p, 4);
+gap> IsPerm2Rep(p);
+true
+
+# Check inverse is also a perm2
+gap> IsPerm2Rep(p^-1);
+true
+
+# But this has not changed q
+gap> IsPerm4Rep(q);
+true
+
+# and it's inverse is the correct type
+gap> IsPerm4Rep(q^-1);
+true
+
+# Check some calculations that use the inverse
+gap> List([1..5], x -> x/p);
+[ 4, 1, 2, 3, 5 ]
+
+# And on q (to ensure it's inverse is not still 'p')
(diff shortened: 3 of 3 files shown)


wilfwilson:
Another option (obviously more work) would be, whenever you trim a perm `p`, to also trim its stored inverse (if it has one) to the same length. Just wondered whether you'd considered this.

ChrisJefferson:
My plan for this PR, given it might be backported, was to do the smallest (hopefully) clearly correct thing. Although I've just realised I failed at that goal, because as well as stopping `p` pointing at it's inverse, I need to stop it's inverse pointing at `p` :(

wilfwilson:
Whoops! Okay, sure, I'll must make a note on #3576.

ChrisJefferson:
Now with bonus fix!

wilfwilson:
And unfortunately Travis is failing because <https://www.gap-system.org> is down 🙁

wilfwilson:
(I can't restart the AppVeyor build; maybe you can @ChrisJefferson, otherwise I guess push again to make the tests pass.)

src/permutat.h
```diff
@@ -98,6 +103,18 @@ EXPORT_INLINE void SET_STOREDINV_PERM(Obj perm, Obj inv)
     }
 }
 
+/* Clear the stored inverse. This is required if 'perm' changes TNUM.
+   Also clears the stored inverse of the stored inverse (which should be
+   perm). */
+EXPORT_INLINE void CLEAR_STOREDINV_PERM(Obj perm)
+{
+    Obj inv = ADDR_OBJ(perm)[0];
+    if(inv) {
```
fingolfin:
```suggestion
    if (inv) {
```

src/permutat.h
```diff
@@ -98,6 +103,18 @@ EXPORT_INLINE void SET_STOREDINV_PERM(Obj perm, Obj inv)
...
+        ADDR_OBJ(inv)[0] = 0;
+    }
+    ADDR_OBJ(perm)[0] = 0;
```
fingolfin: This could be moved inside the `if`, too

fingolfin:
I think one could change to be parameterless and just compute the "largest moved point" and use that. It'll be a minor cost. Or perhaps it could take the maximum of this and the argument...

That said, I'd greatly prefer if we didn't change permutations in place at all, and got rid of `TRIM_PERM` completely. However, the `orb` package uses it for a hash function. But I think we have a "safer" such hashing function in `datastructures`, so perhaps it could be ported to `orb`? Then I'd make a release of `orb`, and once that is in the distro, we could get rid of `TRIM_PERM` in the master branch.

fingolfin:
I took the liberty of updating this PR with the suggestions by Wilf and me.

ChrisJefferson:
I mentioned in the original message why we have to always remove the inverse -- because if we trim `(1,2)(3,4)` to length 2, then we have a logically different permutation, and therefore a different inverse.

fingolfin:
That usage of TRIM_PERM is simply invalid. While it works by chance to call `TRIM_PERM( (1,2)(3,4), 2 )`, this already creates a broken permutation: `TRIM_PERM( (1,2)(3,4), 3 )`. We really should just bite the bullet and always compute the largest moved point in `TRIM_PERM`, IMHO.

I've added a commit which does just that, including test cases.

ChrisJefferson:
The whole reason (I guess) that TRIM_PERM took an argument, instead of just calculating the LargestMovedPoint, is that it was designed to be a very cheap. Actually calculating the LargestMovedPoint makes the function much more expensive.

Also, while I agree TRIM_PERM(p,n) is invalid unless the permutation stabilises [1..n], I wasn't sure if someone might be using it for that purpose (where it would work fine, as long as there were no other references to the permutation).

fingolfin:
We can easily enumerate all uses of `TRIM_PERM` in the library and in packages (only `orb` is affected), and none of them use this to truncate a permutation (and if they did, that would be a serious bug in that code, IMNSHO). And if we further inspect each existing call to `TRIM_PERM`, we see that all but one of them actually call `LARGEST_MOVED_POINT_PERM` or `LargestMovedPoint` right before. The one exception is `TrimStabChain`. That function in turn has a single call across the whole library and all package: `TrimStabChain(S,LargestMovedPoint(Range(hom)));`

This suggests to me that that our best option is to modify `TRIM_PERM` as I do here, and then modify all callers to avoid their `LargestMovedPoint` calls if possible. (It is not quite possible for the hash function in `lib/dicthf.gi:215` but there it will be called only very rarely, and not anymore after the permutation has been trimmed at least once.

Another variant would be to simply trim permutations as a side effect of `LargestMovedPoint`, but that might confuse some tests. It shouldn't break any actual code, though -- although that is a somewhat tenuous "shouldn't".

fingolfin:
OK, small correction: the call `TRIM_PERM( img, LargestMovedPoint( Range( hom ) ) );` in the library actually asks the range *group* for its largest moved point which may (a) be cached and (b) be actually larger than `LargestMovedPoint(img)` (whether that is good or bad or irrelevant is another matter). I am somewhat confused as to why `ImagesRepresentative` call this at all, though -- at least if `TrimStabChain` was called on the stab chain, it shouldn't be necessary?

fingolfin:
Anyhow, this requires some extra work, so I've now restored the `CLEAR_STOREDINV_PERM()` to be the way @ChrisJefferson wants it (though I am still not convinced that it's going to be useful, because in instances where it matters the user already did something wrong; and in all existing calls to TRIM_PERM, it never matters). I also dropped my commit, and will add it to a separate PR, so as to not hold up this important bug fix further!

fingolfin:
By the way, once a permutation has been trimmed, `LargestMovedPointPerm` will be very quick for it, too

Issue #3577, opened 2019-07-16
'Length' became considerably slower


frankluebeck:
### Observed behaviour
In gap-4.10.2 (x86_64):
```
gap> for i in [1..100000000] do od; time;
719
gap> l:=[2..100]+0;;
gap> for i in [1..100000000] do len:=Length(l); od; time;
2063
gap> s:="fhklaheirifnieufhawifhweifhwifuhawifuhwefiuwhf";;
gap> for i in [1..100000000] do len:=Length(s); od; time;
2016
```

Current GAP master (same machine, same memory):
```
gap> for i in [1..100000000] do od; time;
894
gap> l:=[2..100]+0;;
gap> for i in [1..100000000] do len:=Length(l); od; time;
3088
gap> s:="fhklaheirifnieufhawifhweifhwifuhawifuhwefiuwhf";;
gap> for i in [1..100000000] do len:=Length(s); od; time;
3098
```

### Expected behaviour
The empty loop should not be 25% slower and the calls of 'Length' should not be 70% slower than in 4.10.2.

### Comments
Is this connected to the change that lengths of plists and strings are now stored as GAP integers instead of C integers?

ChrisJefferson:
So, I agree this is a problem and needs looking at.

However, the amount of time actually spent in ElmPlist is basically 0. A similar problem is visible, with almost identical slowdown, with (picking a random function) ABS_RAT and an integer:

In GAP master:
```
gap> l := [2..100]+0;; c := 2;;
gap> for i in [1..100000000] do len:=Length(l); od; time; for i in [1..100000000] do ABS_RAT(c); od; time;
3897
2785
```

In 4.10:
```
gap> for i in [1..100000000] do len:=Length(l); od; time; for i in [1..100000000] do ABS_RAT(c); od; time;
2568
1880
```

ChrisJefferson:
Update: we are doing something weird, we are making many more calls to gasman. We should probably do better at benchmarking.

ChrisJefferson:
Update update: that was a mistake.

One big problem I think is checking recursion depth. This got more expensive for 3 reasons:

1) we do it in more places, to try to catch more problems
2) it was moved out of the header so can't be inlined
3) accessing global state was made more general, and in the process got more expensive.

ChrisJefferson:
Update, it was largely caused by a commit of mine, in category (1), where I started checking depth in many more places to try to simplify code and fix a bug. Commit  579456e

[...]

PR #3578, opened 2019-07-16
Speed up recursion checking

ChrisJefferson:
This PR does two things:

1) Removes 579456e2 (except the tests which still pass). This reintroduces #1286, but at preent I cannot find a way of fixing that issue which does not produce an unacceptable speed loss.

2) Performs a simple optimisation to CheckRecursionBefore, which speeds up it's usage, and improves the performance of code which recurses. This can lead to a 6% improvement in very trivial functions (too trivial to do anything useful...)


fingolfin:
Could you explain the motivation for this? It seems to be related to speed, but you never quite say how so?

ChrisJefferson:
Sorry, the motivation is that #3577 observed that a number of fairly simple pieces of GAP code have gotten significantly slower than they were in 4.10. On profiling investigation, the code to check the recursion depth show up much higher than they used to do, to the point where in some cases they are over 50% of the profile, given very small functions.

579456e introduced many more increments and decrements of the recursion depth, to try to catch more places where we could enter an infinite loop. While it fixed a bug, it turns out to be quite expensive, so I think for now we should consider reverting the commit.

While profiling this code I found an easy wasy of speeding up the recursion depth check, which I also put in this PR.

fingolfin:
There are some further test failures in `tst/testspecial/stack-depth-rec.g` and `tst/testspecial/copy-weakpointerobj.g`

ChrisJefferson:
I've broken this PR while fiddling with it. Will report back once I've fixed it.

ChrisJefferson:
Now working.

I have disabled a test I previously wrote. On deep inspection, with this PR it ends up always failing, as a value ends up left on the stack which stops GASMAN collecting a particular bag. While I could poke the test until randomly changes in the stack made it pass again, I decided it was easier to just remove it. This test tested what happened when an operation occurred mid-GC.

If we decided we really cared about this, having a more reliable way of forcing a GC to occur at a particular time, and clean up particular bags.

fingolfin:
You say that you disabled "a test", but two tests were removed. I just wanted to double check that was intentional then, and you meant that you had removed two tests?

ChrisJefferson:
So, I think there are two different test removals.

The first commit removes a test which has been around for a long time, but kept failing with this PR. The PR isn't directly related to this test, it just caused random stack fluctuations which caused the test to fail.


When reverting the old commit at the end of the PR, two of the tests are removed as this unmerging broke them.

PR #3579, opened 2019-07-16
Speed up writing to global variables

ChrisJefferson:
This PR is based around the observation that when we assign a global variable, we have to check iExprsGVar, FopiesGVar or CopiesGVar for that global variable. However, for most global variables they are set to their default values. This PR adds a single flag which tells us, for each gvar, us if any of these 3 arrays have a non-default value.

To avoid introducing another array, we take WriteGVars (which stores one of three values, if a gvar is mutable, read-only or constant), and make it into FlagsGVars, which now stores this original information, and an extra bit that tells us if any of Exprs,FOpies or Copies is not default. As we have to read WriteGVars anyway, whenever we assign a global, this means there are no extra memory lookups, just the unpacking of the flags.

For simplicity, this new bit is never cleared once it is set. This means when it is true we don't know for certain one of exprsgvar, fopiesgvar or copiesgvar is set to a non-default value, but if it is false we know they are are default.

In a tight loop assigning a global, this PR speeds things up by about 40%

* Release Notes: Speed up writing to global variables

### DIFF
--- a/src/gvars.c
+++ b/src/gvars.c
@@ -773,7 +773,6 @@ void MakeReadOnlyGVar (
         ErrorMayQuit("Variable: '%g' is constant", (Int)NameGVar(gvar), 0L);
     }
     SetGVarWriteState(gvar, GVarReadOnly);
-    CHANGED_GVAR_LIST( FlagsGVars, gvar );
 }
 
 /****************************************************************************
@@ -789,7 +788,6 @@ void MakeConstantGVar(UInt gvar)
             (Int)NameGVar(gvar), 0L);
     }
     SetGVarWriteState(gvar, GVarConstant);
-    CHANGED_GVAR_LIST(FlagsGVars, gvar);
 }
 
 
--- a/src/gvars.c
+++ b/src/gvars.c
@@ -186,14 +186,14 @@ inline Obj ValGVar(UInt gvar) {
 /****************************************************************************
 **
 *V  NameGVars . . . . . . . . . . . . . . . . . . . names of global variables
-*V  WriteGVars  . . . . . . . . . . . . .  writable flags of global variables
+*V  FlagsGVars  . . . . . . . . . . . . flags of global variables (see below)
 *V  ExprGVars . . . . . . . . . .  expressions for automatic global variables
 *V  CopiesGVars . . . . . . . . . . . . . internal copies of global variables
 *V  FopiesGVars . . . . . . . .  internal function copies of global variables
 */
 #ifdef USE_GVAR_BUCKETS
 static Obj             NameGVars[GVAR_BUCKETS];
-static Obj             WriteGVars[GVAR_BUCKETS];
+static Obj             FlagsGVars[GVAR_BUCKETS];
 static Obj             ExprGVars[GVAR_BUCKETS];
 static Obj             CopiesGVars[GVAR_BUCKETS];
 static Obj             FopiesGVars[GVAR_BUCKETS];
@@ -210,7 +210,7 @@ static Obj             FopiesGVars[GVAR_BUCKETS];
 #else   // USE_GVAR_BUCKETS
 
 static Obj             NameGVars;
-static Obj             WriteGVars;
+static Obj             FlagsGVars;
 static Obj             ExprGVars;
 static Obj             CopiesGVars;
 static Obj             FopiesGVars;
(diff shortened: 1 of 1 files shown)


src/gvars.c
```diff
@@ -226,6 +226,67 @@ static Obj             FopiesGVars;
...
+static GVarWriteFlag GetGVarWriteState(Int gvar)
+{
+    UInt val = INT_INTOBJ(ELM_GVAR_LIST(FlagsGVars, gvar));
+    return val & 3;
+}
+
+typedef struct {
+    GVarWriteFlag gvarWriteFlag;
+    Int           hasExprCopiesFopies;
+} GVarFlagInfo;
```
fingolfin: So this struct takes 16 bytes on a 64 bit system to store 3 bits of information, which strikes me as a tad wasteful :-). How about this instead:
```C
typedef struct {
    unsigned gvarWriteFlag : 2;
    unsigned hasExprCopiesFopies : 1;
} GVarFlagInfo;
```
And then one could simplify `GetGVarWriteState`, `GetGVarFlagInfo`, etc. ... Well, maybe it wouldn't really be simpler? I must admit I haven't quite thought this through... will do it tomorrow, after some sleep :-)

src/gvars.c
```diff
@@ -226,6 +226,67 @@ static Obj             FopiesGVars;
...
+//         present, so it can be set to 1 for a gvar, while these three arrays
+//         all have their default value.
+
+// Below are some helper methods for reading and writing FlagsGVars. They
+// should always be used to simplifying adding more flags in future.
+
+typedef enum {
+    GVarReadOnly = 0,
+    GVarMutable = 1,
+    GVarConstant = 2
```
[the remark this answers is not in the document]
ChrisJefferson: Just kept the default 0 while making changes, but no need to keep it now

src/gvars.c
ChrisJefferson: I wondered if I could just cast a struct last this to/from an int, but I suspect by the C standard that isn't legal.

fingolfin: That's what we have `union` for, though ;-). So either
```C
  union {  Int intval;  GVarFlagInfo flags; };
```
or directly inside `GVarFlagInfo`:
```C
typedef union {
    int intval;
    struct {
        unsigned gvarWriteFlag : 2;
        unsigned hasExprCopiesFopies : 1;
    };
} GVarFlagInfo;
```

fingolfin: Alternatively, on can of course also cheat and do  `Int val = ...;  return *(GVarFlagInfo *)&val;` but that seems more fragile.

src/gvars.c
```diff
@@ -226,6 +226,67 @@ static Obj             FopiesGVars;
...
+    UInt         val = INT_INTOBJ(ELM_GVAR_LIST(FlagsGVars, gvar));
+    GVarFlagInfo info;
+    info.gvarWriteFlag = val & 3;
+    info.hasExprCopiesFopies = val & 4;
+    return info;
+}
+
+static void SetGVarWriteState(Int gvar, GVarWriteFlag w)
+{
+    UInt val = INT_INTOBJ(ELM_GVAR_LIST(FlagsGVars, gvar));
```
fingolfin: When this is called the very first time, `ELM_GVAR_LIST(FlagsGVars, gvar)` returns `0`, and thus an assertion in `INT_INTOBJ` is triggered. So we need to deal with that case explicitly. Also in the other places accessing this.
Alternatively, we could just change the initialisation of `FlagsGVars` to set it to `INTOBJ_INT(0)` at first.

src/gvars.c
```diff
@@ -639,7 +718,7 @@ UInt GVarName (
         SET_ELM_GVAR_LIST( ValGVars,    numGVars, 0 );
         SET_ELM_GVAR_LIST( NameGVars,   numGVars, string );
         CHANGED_GVAR_LIST( NameGVars,   numGVars );
-        SET_ELM_GVAR_LIST( WriteGVars,  numGVars, INTOBJ_INT(1) );
```
fingolfin: So if we kept this line (possibly changing the 1 to 0), and leave in the `SetGVarWriteState` in the next line, all should be fine, I guess?

src/gvars.c
```diff
@@ -226,6 +226,67 @@ static Obj             FopiesGVars;
 
 #endif
 
+// FlagsGVars contains information about global variables as a single INTOBJ
+// Once turned into a C int, this information is:
+//
+// Bits 0 and 1 - A value of type GVarWriteFlag which denotes if the variable
+// is ReadOnly, Mutable or Constant
```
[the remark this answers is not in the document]
fingolfin: So I am mostly happy with this PR now, except for the use of the term "mutable", which I feel is already overloaded in GAP with other meanings. But please don't hesitate to argue otherwise :-)

src/gvars.c
ChrisJefferson: So in the end I went with memcpy, while I admit it looks stupid, in my experience it seems like the most reliable long-term solution to doing this kind of type changing.

src/gvars.c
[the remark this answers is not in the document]
ChrisJefferson: I could have sworn we use 'mutable' for gvars, but having grepped the code that is not true. changed.

ChrisJefferson:
I ended up merging the commits in this PR together, because both of them had been changed so that the middle one wasn't actually buildable any more, and I didn't think it was worth fixing.

src/gvars.c
```diff
@@ -226,6 +226,75 @@ static Obj             FopiesGVars;
...
 
+// FlagsGVars contains information about global variables.
+// Once cast to a GVarFlagInfo struct, this information is:
+//
+// gvarWriteFlag: A value of type GVarWriteFlag which denotes if the variable
+// is Assignable, ReadOnly, or Constant.
+// hasExprCopiesFopies: If the variable has ever had a non-default value
+// assigned to ExprGVars, CopiesGVars or FopiesGVars. Note that this value is
+// never cleared at present, so it can be set to 1 while these three arrays
+// all have their default value, but if it is 0 these arrays definately have
```
fingolfin:
```suggestion
// all have their default value, but if it is 0 these arrays definitely have
```

fingolfin:
(And by "fine by me" I obviously meant: Very nice improvement, thank you!! Obviously. Ahem. :-)

PR #3580, opened 2019-07-16
Fix potential infinite loop or recursion when computing size of infinite cyclic groups

fingolfin:
This method needs to test whether a group element is the identity; but this
can lead to an infinite recursion if the input is an fp group, as then this
equality check may end up trying to compute the group size. This currently
actually happens when all packages are loaded (with only the default set of
filters, another Size method is selected for cyclic fp groups).

We fix this by adding the `CanEasilyCompareElements` filter to the argument
filters of the method. Strictly speaking, we'd also want to check for
something like `CanEasilyComputerElementOrder`, but that doesn't exist.

### DIFF
--- a/tst/testbugfix/2019-07-17-cyclic-size.tst
+++ b/tst/testbugfix/2019-07-17-cyclic-size.tst
@@ -1,0 +1,10 @@
+# Fix a bug that only occurs if all packages are loaded
+gap> F:=FreeGroup(2);;
+gap> F:=F/[F.1/F.2];;
+gap> H:=Group(F.1);;
+gap> IsCyclic(H);
+true
+gap> HasIsFinite(H);
+false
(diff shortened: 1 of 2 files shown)


tst/testbugfix/2019-07-17-cyclic-size.tst
```diff
@@ -0,0 +1,13 @@
+# Fix a bug that only occurs if all packages are loaded
+gap> F:=FreeGroup(2);
+<free group on the generators [ f1, f2 ]>
+gao> F:=F/[F.1/F.2];
```
DominikBernhardt: Maybe a stupid question, but why does it say gao here?

fingolfin: That's not a stupid question, that's a stupid PR submitter *sigh*

PR #3582, opened 2019-07-17
Fix warnings that pop up with some GCC versions

fingolfin:
With GCC 5.4 on an Ubuntu machine, I get annoying warnings like these:
```
In file included from src/julia_gc.c:246:0:
src/julia_gc.c:243:19: warning: ‘PtrArrayDelete’ defined but not used [-Wunused-function]
 #define ELEM_TYPE Ptr
                   ^
src/dynarray.h:33:23: note: in definition of macro ‘JOIN2’
 #define JOIN2(s1, s2) s1##s2
                       ^
src/dynarray.h:30:17: note: in expansion of macro ‘JOIN’
 #define FN(sym) JOIN(Array, sym)
                 ^
src/dynarray.h:32:22: note: in expansion of macro ‘JOIN2’
 #define JOIN(s1, s2) JOIN2(s1, s2)
                      ^
src/dynarray.h:28:15: note: in expansion of macro ‘JOIN’
 #define Array JOIN(ELEM_TYPE, Array)
               ^
src/dynarray.h:28:20: note: in expansion of macro ‘ELEM_TYPE’
 #define Array JOIN(ELEM_TYPE, Array)
                    ^
src/dynarray.h:30:22: note: in expansion of macro ‘Array’
 #define FN(sym) JOIN(Array, sym)
                      ^
src/dynarray.h:61:13: note: in expansion of macro ‘FN’
 static void FN(Delete)(Array * arr)
             ^
src/julia_gc.c:243:19: warning: ‘PtrArrayClone’ defined but not used [-Wunused-function]
 #define ELEM_TYPE Ptr
                   ^
src/dynarray.h:33:23: note: in definition of macro ‘JOIN2’
 #define JOIN2(s1, s2) s1##s2
                       ^
...
```

This PR makes them go away.

### DIFF
--- a/src/baltree.h
+++ b/src/baltree.h
@@ -26,11 +26,6 @@
 // #define ALLOC allocation function (optional)
 // #define DEALLOC deallocation function (optional)
 
-#ifdef __GNUC__
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-function"
-#endif
-
 #define Tree JOIN(ELEM_TYPE, Tree)
 #define Node JOIN(ELEM_TYPE, Node)
 
@@ -61,7 +56,7 @@ static Int height_to_size[MaxTreeDepth];
 static const Int alpha_hi = 3;
 static const Int alpha_lo = 2;
 
-static void InitBalancedTrees(void)
+static inline void InitBalancedTrees(void)
 {
     if (!height_to_size_init) {
         height_to_size_init = 1;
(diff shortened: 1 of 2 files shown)


src/baltree.h
```diff
@@ -61,7 +61,7 @@ static Int height_to_size[MaxTreeDepth];
 static const Int alpha_hi = 3;
 static const Int alpha_lo = 2;
 
-static void InitBalancedTrees(void)
+static inline void InitBalancedTrees(void)
```
ChrisJefferson: is this what EXPORT_INLINE is for?

fingolfin: No. That is for exporting symbols to the outside world. We do not want this here, to the contrary.

rbehrends:
This is odd. This problem *should* have been handled by the `#pragma GCC` directive, but apparently it wasn't.

fingolfin:
@rbehrends good point, have done so now

PR #3586, opened 2019-07-21
kernel: put GASMAN specific code inside #ifdef USE_GASMAN

fingolfin:
Some code in `sysmem.c` is only really used if GASMAN is used. Disable it otherwise. This makes it a bit easier to understand the code, and to see what's going on.

Perhaps it would be even better if those parts of `sysmem.c` were moved into a new file `gasman_intern.c`. I am happy to do that, if people feel that's the right way to go about it.

### DIFF
--- a/src/stats.c
+++ b/src/stats.c
@@ -33,7 +33,9 @@
 #include "records.h"
 #include "stringobj.h"
 #include "sysfiles.h"
+#ifdef USE_GASMAN
 #include "sysmem.h"
+#endif
 #include "vars.h"
 
 #ifdef HPCGAP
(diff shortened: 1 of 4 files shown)


fingolfin:
@frankluebeck yes, I'd prefer to have all the GASMAN relevant bits in a separate file, too, but there are some complications with that (I forgot what it was, though :/), and figured I could make this PR in the meantime, and clean it up properly later (e.g. by moving all GASMAN specific into a new file `gasman_intern.c`; or moving anything not GASMAN specific out of `sysmem.c`; whatever works better).

PR #3589, opened 2019-07-23
Undocument second argument for `DeclareGlobalFunction"

stevelinton:
# Description

The header of the manual entry for `DeclarelGlobalFunction` suggests it takes a second argument `info`
This is not mentioned in the remainder of the entry, and if passed, is discarded. This PR adjusts the documentation header.

## Text for release notes 
not needed

### DIFF
--- a/lib/oper.g
+++ b/lib/oper.g
@@ -1794,12 +1794,12 @@ end );
 
 #############################################################################
 ##
-#F  DeclareGlobalFunction( <name>, <info> ) . .  create a new global function
+#F  DeclareGlobalFunction( <name> ) . . . . . .  create a new global function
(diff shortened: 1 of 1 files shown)


fingolfin:
Your commit message talks about `InstallGlobalFunction`, but you actually are talking about `DeclareGlobalFunction`. The commit message should also ideally have a first line that's no longer than 60 chars.

BTW, not so long ago, a lot of packages actually did use that second argument, though I made a dozen or so PRs to get them to drop it, with the long term plan to deprecate it.

stevelinton:
Cleaned stuff up. This PR just makes the header of the manual section match the contents. It doesn't block actual use of the second argument.

PR #3591, opened 2019-07-23
Improve documentation of parts of chapter 41 in the reference manual

PaulaHaehndel:
I tried to update some (not all) of the functions in the context of group actions (documented in chapter 41 in the reference manual) such that the documentation matches the behaviour of GAP.

While doing so I included some more examples.

### DIFF
--- a/lib/oprt.gd
+++ b/lib/oprt.gd
@@ -613,7 +613,7 @@ BindGlobal( "OrbitsishOperation", function( name, reqs, usetype, NewAorP )
           gens:= GeneratorsOfGroup( G );
         fi;
         if IsDomain( D ) then
-	   if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
+          if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
         fi;
         return op( G, D, gens, gens, OnPoints );
         end );
@@ -649,7 +649,7 @@ BindGlobal( "OrbitsishOperation", function( name, reqs, usetype, NewAorP )
           gens:= GeneratorsOfGroup( G );
         fi;
         if IsDomain( D ) then
-	   if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
+          if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
         fi;
         return op( G, D, gens, gens, act );
         end );
@@ -679,7 +679,7 @@ BindGlobal( "OrbitsishOperation", function( name, reqs, usetype, NewAorP )
         [ IsGroup, IsObject, IsList, IsList ], 0,
         function( G, D, gens, acts )
         if IsDomain( D ) then
-	   if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
+          if IsFinite( D ) then D:= AsSSortedList( D ); else D:= Enumerator( D ); fi;
         fi;
         return op( G, D, gens, acts, OnPoints );
         end );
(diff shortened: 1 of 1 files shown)


wilfwilson:
Does this resolve any open issues? If so, please could you edit your first post in this PR (I.e. the description) to automatically close those issues (see https://help.github.com/en/articles/closing-issues-using-keywords)

PaulaHaehndel:
As far as I'm aware there are no open issues that will be closed by this PR.

lib/oprt.gd
```diff
@@ -1334,18 +1334,27 @@ OrbitishFO( "Orbit", OrbitishReq, IsCollsElms, false, false );
...
+##  <Attr Name="Orbits" Arg='G'
+##   Label="for a permutation group"/>
 ##  <Attr Name="Orbits" Arg='xset' Label="attribute"/>
 ##
 ##  <Description>
 ##  returns a duplicate-free list of the orbits of the elements in
 ##  <A>seeds</A> under the action <A>act</A> of <A>G</A> or under
 ##  <Ref Func="OnPoints"/> if no action function is given.
 ##  <P/>
+##  For a permutation group <A>G</A>, one may also invoke this as
```
DominikBernhardt: Maybe you could use this shorter version:
_For a permutation group <A>G</A>, <C>Orbits(<A>G</A>)</C> computes the orbits on the moved points of <A>G</A>._

lib/oprt.gd
```diff
@@ -1378,6 +1390,10 @@ OrbitsishOperation( "Orbits", OrbitsishReq, false, NewAttribute );
 ##  The domain <A>Omega</A> must be closed under the action of <A>G</A>,
 ##  otherwise an error can occur.
 ##  <P/>
+##  For a permutation group <A>G</A>, one may also invoke this as
```
DominikBernhardt: If you decide to with my suggestion above, this could be changed accordingly.

lib/oprt.gd
```diff
@@ -1430,18 +1446,26 @@ OrbitishFO( "OrbitLength", OrbitishReq, IsCollsElms, false, false );
...
 ##  <Oper Name="OrbitLengths" Arg='G, seeds[, gens, acts][, act]'
 ##   Label="for a group, a set of seeds, etc."/>
+##  <Attr Name="OrbitLengths" Arg='G' Label="for a permutation group"/>
 ##  <Attr Name="OrbitLengths" Arg='xset' Label="for an external set"/>
 ##
 ##  <Description>
 ##  computes the lengths of all the orbits of the elements in <A>seeds</A>
 ##  under the action <A>act</A> of <A>G</A>.
+##  <P/>
+##  For a permutation group <A>G</A>, one may also invoke this as
```
DominikBernhardt: See above.

DominikBernhardt:
This is essentially the same comment for the changes with your formulation, so I did not make them everywhere. If you want, you can change this, but I don't insist on it.

lib/oprt.gd
```diff
@@ -1587,12 +1626,21 @@ OrbitsishOperation( "ExternalOrbitsStabilizers", OrbitsishReq,
 ##  <P/>
 ##  An action is <E><M>k</M>-transitive</E> if every <M>k</M>-tuple of points
 ##  can be mapped simultaneously to every other <M>k</M>-tuple.
+##  <P/>
+##  For a permutation group <A>G</A>, one may also invoke this as
+##  <C>Transitivity(<A>G</A>)</C>, which returns the degree of transitivity
+##  of the group with respect to its natural action on the points moved by
+##  it.
+##  For example the group <M>\langle (2,3,4),(2,3) \rangle</M>
+##  is 3-transitive on the three points 2, 3, 4.
```
DominikBernhardt: Technically, on usually defines a group to act 3-fold transitive on a set, not on points. So maybe:
_is 3-transitive on the set {2,3,4}._?

DominikBernhardt:
I would suggest that you squash you commits into one commit when you're done and change the commit message.

lib/oprt.gd
PaulaHaehndel: I changed it now in all bits, but it now feels inconsistent with the sentence before.

lib/oprt.gd
PaulaHaehndel: This is the outcome of another conversation in [PR 3363](https://github.com/gap-system/gap/pull/3363#discussion_r267409263). I think both of the two options are fine. So I leave it as it is for now, unless you or someone else has strong objections.

DominikBernhardt: Ok, I wasn;t aware of the discussion in #3363 .

DominikBernhardt:
As @wilfwilson said, this chapter should probably be rewritten, but you changes are a step in the right direction, so thank you for the improvement! I agree with the changes.

One more thing @PaulaHaehndel and @wilfwilson : This PR now consists of 6 commits. Maybe you could squash it into one and write a new commit message? I would then be happy with having this merged.

wilfwilson:
Whoever merges it could instead do a “Squash and merge” which would squash them into one commit automatically while merging. I don’t think it matters so much what happens - I’ll leave it up to @PaulaHaehndel

PR #3592, opened 2019-07-24
Don't automatically call attribute setters on mutable objects

stevelinton:
# Description
This is an alternative to #3588. Instead of stopping the system setters storing in mutable objects, we stop attribute calls from calling the setter altogether in this situation.

This allows for calling the setter directly, but means that even if you have installed a custom setter, it is not called by default.

See also #1371 and #3560 

## Text for release notes 

The values of computed Attributes will no longer be stored automatically in mutable attribute-storing
objects, since they may become out of date as the object mutates. If you wish to store such a value and know that it is safe to do so, then you can address this by calling the corresponding attribute setter directly.

### DIFF
--- a/doc/ref/types.xml
+++ b/doc/ref/types.xml
@@ -485,7 +485,7 @@ except if the attribute had been specially constructed as
 <Q>mutable attribute</Q>.
 <P/>
 It depends on the representation of an object (see&nbsp;<Ref Sect="Representation"/>)
-which attribute values it stores.  An object in the representation
+which attribute values it stores.  An immutable object in the representation
 <C>IsAttributeStoringRep</C> stores <E>all</E> attribute values once they are
 computed.  Moreover, for an object in this representation, subsequent
 calls to an attribute will return the <E>same</E> object; this is achieved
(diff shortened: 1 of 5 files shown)


stevelinton:
Assuming this round of tests pass, I'm now proposing this for inclusion. It seems the better of the reasonably simple alternatives for fixing #1371 and the issues that arose with #3560.

wilfwilson:
There's a failing test https://travis-ci.org/gap-system/gap/jobs/565395421#L2106

stevelinton:
@wilfwilson The test fails because of #3596

stevelinton:
I've adjusted the test to work around #3596. I think this is now ready.

doc/ref/types.xml
```diff
@@ -568,7 +575,8 @@ of the object <A>obj</A> is known.
 <Description>
 For an attribute <A>attr</A>, <C>Setter(<A>attr</A>)</C> 
 is called automatically when the attribute value has been
-computed for the first time.
+computed for an immutable object which does not already have stored
+value for <A>attr</A>.
```
wilfwilson: There's a missing article here, I think 'have a value stored' would make it right.

stevelinton: Fixed, thanks.

PR #3602, opened 2019-08-16
MatrixObj: Add IsEmptyMatrix property

### LINKED ISSUE #3567: MatrixObj: Add IsEmptyMatrix property?
I suggest that we add
```
DeclareProperty( "IsEmptyMatrix", IsMatrixObj );
...
InstallMethod( IsEmptyMatrix, [ IsMatrixObj ], mat -> NrRows(mat) = 0 or NrCols(mat) = 0);
```

DominikBernhardt:
Adds an `IsEmptyMatrix` property for `MatrixObj` objects.
As always I'm open to suggestions. 

Resolves #3567.

### DIFF
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -1289,6 +1289,24 @@ DeclareOperation( "^", [ IsVectorObj, IsMatrixObj ] );
 # AddCoeffs
 
 
+#############################################################################
+##
+##  IsEmptyMatrix( <matobj> )
+##
+##  <#GAPDoc Label="MatObj_IsEmptyMatrix">
+##  <ManSection>
+##    <Prop Name="IsEmptyMatrix" Arg='M' Label="for matrices"/>
+##    <Returns>A boolean</Returns>
+##    <Description>
+##      Is <K>true</K> if <A>M</A> either has zero columns or zero rows and 
+##      <K>false</K> otherwise.
+##      In other words, a matrix is empty if it has no entries.
+##    </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareProperty( "IsEmptyMatrix", IsMatrixObj );
+
 
 # while we are at it also make the naming of following more uniform ???
 
--- a/doc/ref/matrix.xml
+++ b/doc/ref/matrix.xml
@@ -305,6 +305,7 @@ see&nbsp;<Ref Sect="Mutability and Copyability"/>.
 <#Include Label="DeterminantMat">
 <#Include Label="DeterminantMatDestructive">
 <#Include Label="DeterminantMatDivFree">
+<#Include Label="MatObj_IsEmptyMatrix">
 <#Include Label="IsMonomialMatrix">
 <#Include Label="IsDiagonalMat">
 <#Include Label="IsUpperTriangularMat">
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -645,3 +645,12 @@ InstallOtherMethod( NumberColumns, "for a matrix",
 #
 InstallOtherMethod( DimensionsMat, "for a matrix in IsMatrixObj",
   [ IsMatrixObj ], m -> [ NumberRows( m ), NumberColumns( m ) ] );
+
+############################################################################
+##
+#M  IsEmptyMatrix( <matobj> )
+##
+InstallMethod( IsEmptyMatrix,
+  [ IsMatrixObj ],
+  mat -> NrRows(mat) = 0 or NrCols(mat) = 0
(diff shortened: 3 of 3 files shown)


lib/matobj2.gd
```diff
@@ -1289,6 +1289,22 @@ DeclareOperation( "^", [ IsVectorObj, IsMatrixObj ] );
...
+##
+#P  IsEmptyMatrix( <matobj> )
+##
+##  <#GAPDoc Label="IsEmptyMatrix">
+##  <ManSection>
+##  <Prop Name="IsEmptyMatrix" Arg='M'/>
+##
+##  <Description>
+## is <K>true</K> if <A>M</A> either has zero columns or zero rows.
+DeclareProperty( "IsEmptyMatrix", IsMatrixObj );
```
fingolfin: That line should be after the GAPDoc comment

lib/matobj2.gd
```diff
@@ -1289,6 +1289,22 @@ DeclareOperation( "^", [ IsVectorObj, IsMatrixObj ] );
...
+#############################################################################
+##
+#P  IsEmptyMatrix( <matobj> )
+##
+##  <#GAPDoc Label="IsEmptyMatrix">
+##  <ManSection>
+##  <Prop Name="IsEmptyMatrix" Arg='M'/>
+##
+##  <Description>
+## is <K>true</K> if <A>M</A> either has zero columns or zero rows.
```
fingolfin:
```suggestion
##  is <K>true</K> if <A>M</A> either has zero columns or zero rows.
```

Right now this property will not appear in the manual. Is that intentional?

stevelinton:
Why is this a property, not a category? Is it possible that a matrix will not know at creation time that it is empty? 

Actually, should the category be `IsNonEmptyMatrix` implying that `m[1,1]` must exist. That seems more useful for method selection.

DominikBernhardt:
@fingolfin Thanks, I adressed your comments.
@stevelinton Tbh I just followed the suggestion of @fingolfin  from Issue #3567 . I am open to changes but I am not sure what the benefits of turning this into a category would be?

fingolfin:
A category won't work reliably (except with kernel changes plus utmost care by every single MatrixObj implementation, which is unrealistic): every matrix constructor creating matrix objects would have to set it; operations adding or removing rows or columns would have to clear or set it; the kernel would need to be made aware in order to set it for plist-of-list matrices; etc.

And then there are lazy matrices (e.g. in homalg) for which determining whether they are empty requires a nontrivial operation (e.g. to compute a Gröbner basis).

All in all, I think this needs to be a property. If that's not possible, then perhaps an operation; but category makes no sense to me.

stevelinton:
@fingolfin OK. I hadn't realised that MatrixObj objects could change their dimensions. In that case it's clearly a Property.

fingolfin:
@stevelinton as long as we keep the implication `IsMatrix => IsMatrixObj` (which we may not want to do forever....), this can indeed change. But as I explained, even if we forbade this change, there is legitimate interest in allow matrices to lazily compute their properties, including the dimension.

lib/matobj2.gd
```diff
@@ -1289,6 +1289,25 @@ DeclareOperation( "^", [ IsVectorObj, IsMatrixObj ] );
...
+##      Is <K>true</K> if <A>M</A> either has zero columns or zero rows and 
+##      <K>false</K> otherwise.
+##    </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareProperty( "IsEmptyMatrix", IsMatrixObj );
+
+
+
```
fingolfin: I usually tru not to put more than two empty lines in a row. Not a reason to delay this PR, though :-)

DominikBernhardt: Thanks, your right. I changed it. ;)

lib/matobj2.gd
```diff
@@ -1289,6 +1289,23 @@ DeclareOperation( "^", [ IsVectorObj, IsMatrixObj ] );
...
+##
+##  IsEmptyMatrix( <matobj> )
+##
+##  <#GAPDoc Label="MatObj_IsEmptyMatrix">
+##  <ManSection>
+##    <Prop Name="IsEmptyMatrix" Arg='M' Label="for matrices"/>
+##    <Returns>A boolean</Returns>
+##    <Description>
+##      Is <K>true</K> if <A>M</A> either has zero columns or zero rows and 
+##      <K>false</K> otherwise.
```
wilfwilson: I'd like it if you could add another sentence:

> In other words, a matrix is empty if it has no entries.

PR #3603, opened 2019-08-18
Various fixes to IsAssociate, StandardAssociate(Unit), ... and some tests

fingolfin:
Fixes #3510

PR #3605, opened 2019-08-19
Replace 'for ring' with 'for a ring'

DominikBernhardt:
This originates from PR #3603. It replaces `for ring [...]` with `for a ring [...]`.


wilfwilson:
App Ryle has messed up.

Issue #3569, opened 2019-07-14
Add a convenient way to display (or even access!) bindings inside of closures


fingolfin:
Consider this:
```
gap> f := ApplicableMethod( OrbitsDomain, [ SymmetricGroup(5), [1..5] ] );;
gap> Display(f);
function ( G, D )
    if D = MovedPoints( G ) then
        return op( G );
    else
        TryNextMethod();
    fi;
    return;
end
```

Quick, what is "op" standing for here?

...

Well, of course again for `OrbitsDomain`.

Code that generates code by using closures can be super helpful and nice, but a nighmare to debug, largely because it can be so tricky to figure out the bindings inside a closure. A more direct example would be this:
```
gap> makeFun:=n -> x -> x + n;;
gap> f:=makeFun(42);;
gap> Display(f);
function ( x )
    return x + n;
end
```
If you hadn't seen the definition of `f` here, you'd have no way to know what `n` is.

This should be different. I suggest that we add a way to view the local bindings of a function; at least the ones that are actually used in it. We could then show it when "Displaying" the function, or even allow the user to query the actual objects. E.g. 
gap> Display(f);
# function closure with bindings:
#   n = 42
function ( x )
    return x + n;
end
gap> BindingsOfClosure(f);
rec( n := 42 )
```
Of course the "bindings" are strictly speaking all locals (including arguments) of the parent function of the given function (and this can potentially go back recursively). Coming back to the initial `OrbitsDomain` example, the parent function is `OrbitsishOperation` and it starts like this:
```gap
BindGlobal( "OrbitsishOperation", function( name, reqs, usetype, NewAorP )
    local op;

    # Create the attribute or property.
    op:= NewAorP( name, IsExternalSet );
...
```
So the "bindings" in my initial example actually also include `name`, `reqs`, `usetype` and `NewAorP`. But I think that's usually actually quite helpful.

The first step towards implementing this idea would be to add a lowlevel function for getting the `ENVI_FUNC` of a function, which points to the `lvars` of the parent function, if there is any. (Note, however, that a few places in the kernel abuse `ENVI_FUNC` to store other things in there, in particular for attributes getter/setter functions; but this can be dealt with with a simple TNUM check).

Once one has this, defining `BindingsOfClosure` should be easy, and once one has that, better printing also is no problem.

wilfwilson:
I updated the issue in light of #3606 being created and merged.

fingolfin:
[...]

PR #3606, opened 2019-08-19
Add BindingsOfClosure helper

fingolfin:
Partially addresses #3569.

Well, or maybe it addresses it completely, though one still might want to enhance the output of `Display(func)` if there are bindings? Somebody else could do that based on this PR, though.

I also guess it would be nice to have documentation for `BindingsOfClosure`...

### DIFF
--- a/src/vars.c
+++ b/src/vars.c
@@ -2008,13 +2008,16 @@ static Obj FuncParentLVars(Obj self, Obj lvars)
 
 static Obj FuncContentsLVars(Obj self, Obj lvars)
 {
+  if (!IS_LVARS_OR_HVARS(lvars)) {
+      RequireArgument("ContentsLVars", lvars, "must be an lvars");
+  }
   Obj contents = NEW_PREC(0);
   Obj func = FUNC_LVARS(lvars);
   Obj nams = NAMS_FUNC(func);
   UInt len = (SIZE_BAG(lvars) - 2*sizeof(Obj) - sizeof(UInt))/sizeof(Obj);
   Obj values = NEW_PLIST_IMM(T_PLIST, len);
   if (lvars == STATE(BottomLVars))
-    return False;
+    return Fail;
   AssPRec(contents, RNamName("func"), func);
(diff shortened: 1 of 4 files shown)


src/vars.c
```diff
@@ -2008,13 +2008,16 @@ static Obj FuncParentLVars(Obj self, Obj lvars)
...
+      RequireArgument("ParentLVars", lvars, "must be an lvars");
+  }
   Obj contents = NEW_PREC(0);
   Obj func = FUNC_LVARS(lvars);
   Obj nams = NAMS_FUNC(func);
   UInt len = (SIZE_BAG(lvars) - 2*sizeof(Obj) - sizeof(UInt))/sizeof(Obj);
   Obj values = NEW_PLIST_IMM(T_PLIST, len);
   if (lvars == STATE(BottomLVars))
-    return False;
+    return Fail;
```
PaulaHaehndel: I assume then the test (what travis tests) should be changed accordingly.

fingolfin: Updated

wilfwilson:
@fingolfin Do you intend to document `BindingsOfClosure` in this PR?

fingolfin:
I don't have the time and energy to work on documentation for `BindingsOfClosure` right now. I am not sure we'd want to document it as such globally, either... so there are multiple paths forward:
1. We merge this now, then for a few people `BindingsOfClosure` already is useful; and update #3569 to mention it, suggesting that somebody else submit a PR to enhance `Display` for closures using it (and also that they could consider documenting `BindingsOfClosure` as well).
2. We merge it until somebody has time and energy to write some documentation, and tack it onto this PR.

I think 1. is better because it means this code won't bitrot, and people can benefit from it already (even if the number of people doing so would be small, as it is not documented). But either is fine by me.

wilfwilson:
I'm happy for us to do 1. I thought that _perhaps_ your sentence
> I also guess it would be nice to have documentation for `BindingsOfClosure`...

suggested that you were maybe thinking of creating documentation, which is why I asked about it.

PR #3610, opened 2019-08-20
Invalid use of / on ZmodnZObjs (e.g. dividing a unit by a zero divisor) produces an error, instead of returning fail

fingolfin:
If you want `fail`, use `Quotient`, which was made for that.


DominikBernhardt:
@fingolfin I think you forgot to adjust a few tests as they still expect `fail` as an output.

PR #3611, opened 2019-08-20
Fix StandardAssociateUnit for polynomial rings

fingolfin:
By definition in the documentation, the returned unit must be an element of the ring.

The drawback of this is that it makes `StandardAssociateUnit` a bit less efficient: converting the unit into a polynomial introduces overhead in terms of memory and runtime. And for typical applications of this (e.g. implementing row reductions in a matrix), it doesn't matter, as all we really need is that for inputs `a` and `b` we get a "unit" `u` such that `a  * u = b` and `a = b * u^-1` hold.

So an alternative to this PR would be to adjust the documentation, and also `checkStandardAssociate`.

Thoughts, anybody?

### DIFF
--- a/hpcgap/lib/ringpoly.gi
+++ b/hpcgap/lib/ringpoly.gi
@@ -635,7 +635,7 @@ InstallMethod(StandardAssociateUnit,
 function(R,f)
   local c;
   c:=LeadingCoefficient(f);
-  return StandardAssociateUnit(CoefficientsRing(R),c);
+  return StandardAssociateUnit(CoefficientsRing(R),c) * One(R);
 end);
 
 InstallMethod(FunctionField,"indetlist",true,[IsRing,IsList],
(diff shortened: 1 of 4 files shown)


wilfwilson:
> So an alternative to this PR would be to adjust the documentation...

What would the documentation say in the case? Something like
> ...returns a unit in _some_ ring whose elements can be multiplied with those of `R` with the property that..."?

I think it would also need an explanation for why it was this way, otherwise it would seem a bit weird.

To be honest I don't really mind, I've never used this function directly and don't think I would do anytime soon. From my mathematical point of view, I think I prefer the behaviour and documentation as this PR makes things.

I understand the inefficiency side of things. Could this be perhaps accommodated with another function with a different name?

fingolfin:
I suggest we backport this to 4.11, simply because several other fixes for `StandardAssociateUnit` are already in there. However, if people prefer to defer this to 4.12, I don't mind either.

hulpke:
Yes, please backport.

PR #3612, opened 2019-08-20
Clarify, fix and test semantics of Quotient(R,a,b)

fingolfin:
This improves the documentation of `Quotient` to make it sensible for non-commutative rings, and rings with zero divisors; it also fixes some methods for `Quotient` which violated the documentation of `Quotient` (before and after my changes to it).

In particular `Quotient(R, x, Zero(R))` should return fail, not run into an error.

Also add some tests for various `Quotient` methods. It turns out that those for
polynomials still are broken, which should be fixed in the future.

### DIFF
--- a/tst/testinstall/opers/Quotient.tst
+++ b/tst/testinstall/opers/Quotient.tst
@@ -1,0 +1,128 @@
+gap> START_TEST("Quotient.tst");
+
+#
+gap> Quotient(2, 1);
+2
+gap> Quotient(1, 2);
+fail
+gap> Quotient(1, 0);
+fail
+gap> Quotient(Integers, 2, 1);
+2
+gap> Quotient(Integers, 1, 2);
+fail
+gap> Quotient(Integers, 1, 0);
+fail
+gap> Quotient(GaussianIntegers, 2, 1);
+2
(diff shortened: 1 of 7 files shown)


tst/testinstall/opers/Quotient.tst
```diff
@@ -0,0 +1,130 @@
...
+Z(7)^4
+gap> Quotient(R, t, 0*t);
+fail
+
+#
+gap> R:=PolynomialRing(Integers mod 6, "t"); t:=R.1;;
+<monoid>[t]
+gap> Quotient(2*t, t);
+ZmodnZObj(2,6)
+gap> #Quotient(t, 2*t); # FIXME: should return fail
```
wilfwilson: Do you intend to fix this `FIXME` in this PR?

fingolfin: @wilfwilson Nope, that would require substantial changes to the polynomial code, which isn't something I plan to look at any time soon.

fingolfin: I tried to clarify the comment to make it clearer what is going on. Also tried to fix the test failure.

wilfwilson: Thanks

PR #3615, opened 2019-08-21
document basic representations of objects

ThomasBreuer:
- documented `IsInternalRep`, `IsDataObjectRep`, `IsComponentObjectRep`,
  `IsPositionalObjectRep`, `IsAttributeStoringRep`, `IsPlistRep`

  Up to now, these representations were either not documented at all
  or just via `<Index>` entries in some general text.
  I would like to refer to some of these representations in the documentation for `IsMatrixObj` and `IsVectorObj`; for that, they should be properly documented.

- adjusted the relevant cross-references:
  Use `<Ref...>` instead of `<C>` elements.

### DIFF
--- a/lib/list.gd
+++ b/lib/list.gd
@@ -59,8 +59,7 @@ BIND_GLOBAL( "ListsFamily", NewFamily( "ListsFamily", IsList ) );
 ##
 ##  <Description>
 ##  &GAP; lists created by entering comma separated values in square brackets
-##  are represented internally as so-called <E>plain lists</E>,
-##  see <Ref Filt="IsPlistRep"/>.
+##  are usually represented internally as so-called <E>plain lists</E>.
 ##  Other representations of lists are <Ref Filt="IsBlistRep"/>,
 ##  <Ref Filt="IsRangeRep"/>, <Ref Filt="IsStringRep"/>,
 ##  or the ones that are chosen for implementing enumerators,
--- a/lib/list.gd
+++ b/lib/list.gd
@@ -53,12 +53,34 @@ BIND_GLOBAL( "ListsFamily", NewFamily( "ListsFamily", IsList ) );
 ##
 #R  IsPlistRep  . . . . . . . . . . . . . . . . representation of plain lists
 ##
+##  <#GAPDoc Label="IsPlistRep">
 ##  <ManSection>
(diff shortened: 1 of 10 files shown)


lib/list.gd
```diff
@@ -53,12 +53,34 @@ BIND_GLOBAL( "ListsFamily", NewFamily( "ListsFamily", IsList ) );
...
 #R  IsPlistRep  . . . . . . . . . . . . . . . . representation of plain lists
 ##
+##  <#GAPDoc Label="IsPlistRep">
 ##  <ManSection>
-##  <Filt Name="IsPlistRep" Arg='obj' Type='Representation'/>
+##  <Filt Name="IsPlistRep" Arg='obj' Type='representation'/>
 ##
 ##  <Description>
+##  &GAP; lists created by entering comma separated values in square brackets
+##  are represented internally as so-called <E>plain lists</E>,
```
fingolfin: Note that this is not 100% correct: for boolean lists, GAP returns a blist rep:
```
gap> IsPlistRep( [ true, false ] );
false
gap> IsBlistRep( [ true, false ] );
true
```

lib/type.g
```diff
@@ -314,14 +367,52 @@ DeclareRepresentation( "IsAtomicPositionalObjectRep",
...
+##  1
+##  false
+##  ]]></Example>
+##  <!--
 ##  The name of the  component that holds  the value of  an attribute is  the
 ##  name of the attribute, with the first letter turned to lower case.
-##  <!-- This will be changed eventually, in order to avoid conflicts between-->
-##  <!-- ordinary components and components corresponding to attributes.-->
+##  This will be changed eventually, in order to avoid conflicts between
+##  ordinary components and components corresponding to attributes.-->
```
fingolfin: Perhaps we should just remove those last 4-5 lines?

ThomasBreuer: Yes.
Note that the rule for the component names had not been commented out before, but still it was not part of the documentation since the GAPDoc "chunk" was not requested by any of the XML files.

lib/list.gd
ThomasBreuer: Good point. I will fix this.

ThomasBreuer:
@fingolfin I have updated my pull request according to your comments.

PR #3616, opened 2019-08-21
Irreducible Modules and Permutation Representations

hulpke:
Performance improvement (count modules and stop if all are found, split tensor square) for computing irreducible modules.

Better determination of (factor) permutation representations,  in particular for group extensions.

Added more test examples.

### DIFF
--- a/lib/factgrp.gd
+++ b/lib/factgrp.gd
@@ -374,7 +374,7 @@ DeclareGlobalFunction("CloseNaturalHomomorphismsPool");
 
 #############################################################################
 ##
-#F  PullBackNaturalHomomorphismsPool(<hom>]) . . transfer nathoms of image
+#F  PullBackNaturalHomomorphismsPool(<hom>) . . transfer nathoms of image
 ##
 ##  <ManSection>
 ##  <Func Name="PullBackNaturalHomomorphismsPool" Arg='hom'/>
@@ -386,6 +386,23 @@ DeclareGlobalFunction("CloseNaturalHomomorphismsPool");
 ##  </ManSection>
 ##
 DeclareGlobalFunction("PullBackNaturalHomomorphismsPool");
+
+#############################################################################
+##
+#F  TryQuotientsFromFactorSubgroups(<hom>,<ker>,<bound>) 
+##
+##  <ManSection>
+##  <Func Name="TryQuotientsFromFactorSubgroups" Arg='hom,ker,bound'/>
+##
+##  <Description>
+##  For a homomorphism <A>hom</A>, this command iterates through subgroups
+##  of the image, up to index <A>bound</A>,
+##  trying to find derived subgroups that expose more of the
+##  factor modulo <A>ker</A>.
+##  </Description>
+##  </ManSection>
+##
+DeclareGlobalFunction("TryQuotientsFromFactorSubgroups");
 
 #############################################################################
 ##
--- a/lib/factgrp.gi
+++ b/lib/factgrp.gi
@@ -1584,7 +1584,7 @@ BindGlobal("FACTGRP_TRIV",Group([],()));
 InstallMethod(NaturalHomomorphismByNormalSubgroupOp,
   "search for operation",IsIdenticalObj,[IsGroup,IsGroup],0,
 function(G,N)
-local h;
+local h,pool;
 
   # catch the trivial case N=G 
   if CanComputeIndex(G,N) and IndexNC(G,N)=1 then
@@ -1602,11 +1602,17 @@ local h;
   fi;
 
   # check, whether we already know a factormap
+  pool:=NaturalHomomorphismsPool(G);
+  h:=PositionSet(pool.ker,N);
+  if h<>fail and IsGeneralMapping(pool.ops[h]) then
+    return GetNaturalHomomorphismsPool(G,N);
+  fi;
+
+
   DoCheapActionImages(G);
   if HasRadicalGroup(G) and N=RadicalGroup(G) then
     h:=GetNaturalHomomorphismsPool(G,N);
   fi;
-
 
   h:=DegreeNaturalHomomorphismsPool(G,N);
   if h<>fail and RootInt(h^3,2)<IndexNC(G,N) then
@@ -1729,6 +1735,45 @@ end);
 
 #############################################################################
 ##
+#F  TryQuotientsFromFactorSubgroups(<hom>,<ker>,<bound>) 
+##
+InstallGlobalFunction(TryQuotientsFromFactorSubgroups,function(hom,ker,bound)
+local s,p,k,it,u,v,d,ma,mak,lev,sub,low;
+  s:=Source(hom);
+  p:=Image(hom);
+  k:=KernelOfMultiplicativeGeneralMapping(hom);
+  it:=DescSubgroupIterator(p:skip:=4);
+  repeat
+    u:=NextIterator(it);
+    Info(InfoExtReps,2,"Factor subgroup index ",Index(p,u));
+    v:=PreImage(hom,u);
+    d:=DerivedSubgroup(v);
+    if not IsSubset(d,k) then
+      d:=ClosureGroup(ker,d);
+      if not IsSubset(d,k) then
+        ma:=NaturalHomomorphismByNormalSubgroup(v,d);
+        mak:=Image(ma,k);
+        lev:=0;
+        sub:=fail;
+        while sub=fail do
+          lev:=lev+1;
+          low:=ShallowCopy(LowLayerSubgroups(Range(ma),lev));
+          SortBy(low,x->-Size(x));
+          sub:=First(low,x->not IsSubset(x,mak));
+        od;
+        sub:=PreImage(ma,sub);
+        Info(InfoExtReps,2,"Found factor permrep ",IndexNC(s,sub));
+        d:=Core(s,sub);
+        AddNaturalHomomorphismsPool(s,d,sub);
+        k:=Intersection(k,d);
+        if Size(k)=Size(ker) then return;fi;
+      fi;
+    fi;
+  until IndexNC(p,u)>=bound;
+end);
+
+#############################################################################
+##
 #M  UseFactorRelation( <num>, <den>, <fac> )  . . . .  for perm group factors
 ##
 InstallMethod( UseFactorRelation,
@@ -1748,3 +1793,4 @@ InstallMethod( UseFactorRelation,
    fi;
    TryNextMethod();
    end );
+
--- a/lib/grplatt.gi
+++ b/lib/grplatt.gi
@@ -3335,57 +3335,252 @@ local c,n,deg,ind,core,i,j,sum;
 
 end);
 
+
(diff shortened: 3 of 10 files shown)


lib/factgrp.gd
```diff
@@ -387,6 +387,23 @@ DeclareGlobalFunction("CloseNaturalHomomorphismsPool");
...
 
+#############################################################################
+##
+#F  TryQuotientsFromFactorSubgroups(<hom>,<ker>,<bound>) 
+##
+##  <ManSection>
+##  <Func Name="TryQuotientsFromFactorSubgroups" Arg='hom,ker,bound'/>
+##
+##  <Description>
+##  If <A>hom</a> is a homomorphism, this command iterates through subgroups
```
fingolfin: Nitpick but: this sounds as if ho does not *need* to be a homomorphism?

lib/factgrp.gi
```diff
@@ -1727,6 +1733,45 @@ local s,r,nat,k;
...
 end);
 
+#############################################################################
+##
+#F  TryQuotientsFromFactorSubgroups(<hom>,<ker>,<bound>) 
+##
+InstallGlobalFunction(TryQuotientsFromFactorSubgroups,function(hom,ker,bound)
+local s,p,k,it,u,v,d,ma,mak,lev,sub,low;
+  s:=Source(hom);
+  p:=Image(hom);
```
fingolfin: Observation: `TryQuotientsFromFactorSubgroups ` is never executed by the Travis test suite, which include testinstall and teststandard. Is it only used for gigantic example or something like that?

lib/twocohom.gi
```diff
@@ -1533,15 +1586,66 @@ local hom,mats,mode,m,min,i,j,mo,bas,a,l,ugens,gi,r,cy,act,k,it,p;
   until false;
 end);
 
+BindGlobal("RandomSubgroupNotIncluding",function(g,n,time)
+local best,i,u,v,start,cnt;
+  start:=Runtime();
+  best:=TrivialSubgroup(g);
+  cnt:=0;
+  while Runtime()-start<time or cnt<100 do
```
fingolfin: Could you please use `GET_TIMER_FROM_ReproducibleBehaviour` so that the code results in reproducible behavior when requested?

lib/twocohom.gi
```diff
@@ -1533,15 +1586,66 @@ local hom,mats,mode,m,min,i,j,mo,bas,a,l,ugens,gi,r,cy,act,k,it,p;
...
-      it,hom;
+local r,z,ogens,n,gens,str,dim,i,j,f,rels,new,quot,g,p,collect,m,e,fp,old,sim,
+      it,hom,trysy,prime,mindeg,fps,ei,mgens,mwrd,nn,newfree,mfpi,mmats,sub,
+      tab,tab0,evalprod,gensmrep,invsmrep,zerob,step;
+
+  # function to evaluate product (as integer list) in gens (and their
+  # inverses invs) with corresponding action mats
+  evalprod:=function(w,gens,invs,mats)
+  local new,i,a;
+    new:=[[],zerob];
```
fingolfin: Observation: `evalprod` is never executed by the Travis test suite.

lib/twocohom.gi
```diff
@@ -1590,20 +1697,221 @@ local r,z,ogens,n,gens,str,dim,i,j,f,rels,new,quot,g,p,lay,m,e,fp,old,sim,
...
+              50
+              # the rewriting seems to be sufficiently spiffy that we don't
+              # need to worry about this more involved process.
+              then
+
+              # Rewriting produces a bad presentation. Rather rebuild a new
+              # fp group using rewriting rules, finds its abelianization,
+              # and then lift that quotient
+              sub:=PreImage(quot,m);
+              tab0:=AugmentedCosetTableInWholeGroup(sub);
```
fingolfin: Observation: `this new code starting at line 1732 ` is never executed by the Travis test suite

lib/twocohom.gi
```diff
@@ -1590,20 +1697,221 @@ local r,z,ogens,n,gens,str,dim,i,j,f,rels,new,quot,g,p,lay,m,e,fp,old,sim,
...
+                    str:=Stabilizer(nn,ei[i],OnSets);
+                    if Size(Core(nn,str))=1 then
+                      j:=ei[i];
+                    fi;
+                    i:=i+1;
+                  od;
+                  if j<>false then
+                    Info(InfoExtReps,4,"deg. improved by blocks ",Size(j));
+                    # action on blocks
+                    i:=ActionHomomorphism(nn,Orbit(nn,j,OnSets),OnSets);
```
fingolfin: Observation: The code in this `if` statement is never executed by the Travis test suite.

lib/twocohom.gi
```diff
@@ -1590,20 +1697,221 @@ local r,z,ogens,n,gens,str,dim,i,j,f,rels,new,quot,g,p,lay,m,e,fp,old,sim,
...
+                  NrMovedPoints(e!.quot)/NrMovedPoints(nn));
+
+                e:=SubgroupOfWholeGroupByQuotientSubgroup(FamilyObj(fps),nn,
+                  Stabilizer(nn,1));
+              fi;
+            elif e=fail and IndexNC(p,m)>trysy then
+              trysy:=Size(fp); # never try again
+              # can the Sylow subgroup get us something?
+              m:=SylowSubgroup(p,prime);
+              e:=PreImage(quot,m);
```
fingolfin: Observation: The code in this `elif` block is never executed by the Travis test suite.

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
...
+  fi;
+
+  # make elementary
+  p:=Factors(IndexNC(G,d))[1];
+  ma:=NaturalHomomorphismByNormalSubgroup(G,d);
+  a:=Image(ma,G);
+  ab:=AbelianInvariants(a);
+  if Length(Set(ab))>1 then
+    e:=LogInt(Exponent(a),p);
+    b:=fail;
```
fingolfin: Observation: The code in this `if` statement (lines 3373-3394) is never executed by the Travis test suite.

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
...
+                SortBy(l,Size);
+              fi;
+
+              if Size(l[Length(l)])<divs[Length(divs)] then
+                # new size. 
+
+                if Length(process)>0 
+                 and Size(process[Length(process)])<=limit then
+                  # switch to lattice
+                  a:=Size(l[Length(l)]);
```
fingolfin: Observation: The code in this `if` statement is never executed by the Travis test suite.

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
...
+  elif IsPrimeInt(ind) then return d;fi;
+
+  # make a p group
+  ind:=List(Collected(Factors(ind)),x->x[1]^x[2]);
+  Sort(ind);
+  if Length(ind)>1 then 
+    i:=0;
+    while i<Length(ind) and ind[i+1]<=limit do
+      i:=i+1;
+    od;
```
fingolfin: Isn't the above just the same as `i := PositionSorted(ind, limit);` ? (Ah well, maybe not if limit exceeds all entries in `ind`... then the current code sets `i` to `Length(ind)`, while `PositionSorted` would set it to `Length(ind)+1`.

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
...
+BindGlobal("BoundedIndexAbelianized",function(P,G,limit)
+local d,ind,i,ma,ab,a,p,b,c,e,n;
+  d:=DerivedSubgroup(G);
+  ind:=IndexNC(G,d);
+  if ind=1 then return G;
+  # derived index small enough
+  elif ind<=limit then return d;
+  elif IsPrimeInt(ind) then return d;fi;
+
+  # make a p group
```
```suggestion
  # make a p-group
```

tst/teststandard/opers/IsomorphismGroups.tst
```diff
@@ -10,7 +10,7 @@ gap> Length(CharacteristicSubgroups(g));
 5
 gap> Length(CharacteristicSubgroups(h));
 5
-gap> IsomorphismGroups(g,h:forcetest)<>fail;
+gap> IsomorphismGroups(g,h)<>fail;
```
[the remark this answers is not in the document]
hulpke: The `forcetest` option makes the algorithm immediately go into a costly step of applying the p-quotient algorithm to what will be a Sylow subgroup (so that that code will be tested). By default this only happens much later if the (faster) standard heuristics do not make progress.
The hope was that this way the default test would have checked that code, alas it turned out that travis timed out (and I had to remove the test again).

lib/grplatt.gi
hulpke: I think it would have to be `PositionSorted(ind,limit+1)` to get the first position where `ind` is larger. Still, I don't think the code is hard to read or slow in practice, so I would prefer to keep it boring.

lib/grplatt.gi
hulpke: Yes. Once subgroups of size <=limit are required, it is worth top do the whole lattice rather than do iterated maximals. I have not been able to find an example of this case that is small enough that I would not have to worry about the test not timing out. (I have run this in larger examples by hand.)

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
 
 end);
 
+
+# utility function: Find subgroup of $G$, containing $G'$ of index up to
+# limit which is normal in the normalizer of G in P.
```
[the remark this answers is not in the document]
hulpke: I've rephrased.

lib/grplatt.gi
```diff
@@ -3335,57 +3335,254 @@ local c,n,deg,ind,core,i,j,sum;
...
+
+  # make a p group
+  ind:=List(Collected(Factors(ind)),x->x[1]^x[2]);
+  Sort(ind);
+  if Length(ind)>1 then 
+    i:=0;
+    while i<Length(ind) and ind[i+1]<=limit do
+      i:=i+1;
+    od;
+    ind:=List(ind{Difference([1..Length(ind)],[i])},x->Factors(x)[1]);
```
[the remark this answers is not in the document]
hulpke: I'm taking the largest prime power dividing $[G:G']$ that is not larger than limit (since then I can factor out all other Sylow subgroups and get a characteristic subgroup of small index).
Yes, your code would be equivalent.

lib/twocohom.gi
hulpke: Done, Thanks.

lib/grplatt.gi
hulpke: Changed.

fingolfin:
Backported to stable-4.11 in commit 9bf9dac66d6276c35afbe57cc5758fe61e5eb2ce

PR #3617, opened 2019-08-22
kernel: remove a bunch of useless default functions; optimize Unbind for blists, ranges

fingolfin:
Also fix some comments and add tests.

### DIFF
--- a/src/range.c
+++ b/src/range.c
@@ -1190,6 +1190,10 @@ static Int InitKernel (
     IsHomogListFuncs[ T_RANGE_NSORT +IMMUTABLE ] = AlwaysYes;
     IsHomogListFuncs[ T_RANGE_SSORT            ] = AlwaysYes;
     IsHomogListFuncs[ T_RANGE_SSORT +IMMUTABLE ] = AlwaysYes;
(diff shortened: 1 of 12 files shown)


src/range.c
```diff
@@ -404,6 +404,27 @@ static Obj ElmsRange(Obj list, Obj poss)
...
+**  This is to avoid unpacking of the range to a plain list when <pos> is
+**  larger or equal to the length of <list>.
+*/
+static void UnbRange(Obj list, Int pos)
+{
+    GAP_ASSERT(IS_MUTABLE_OBJ(list));
+    const Int len = GET_LEN_RANGE(list);
+    if (len == pos && len > 2) {
+        SET_LEN_RANGE(list, len - 1);
+    }
```
wilfwilson: Was going to ask whether you can do something similar with `pos == 1 && len > 2`  But then I remembered what unbinding meant. 

Why is it `len > 2` and not `len >= 2` - are ranges of length 1 forbidden?

ChrisJefferson: Yep, you can't make ranges of length 1:

```gap> IsRangeRep([1..1]);
false
```

You could, hypothetically, optimise this case very slightly by directly building the length 1 plist, but I don't think that's worth it.

PR #3618, opened 2019-08-22
Refactor the 'classical' random number generator

fingolfin:
So far, the old random number generator using a single global state in the global variables `R_N` and `R_X` was still present in the GAP library, and was used to power the `GlobalRandomSource`. In addition, there was a type `IsGAPRandomSource` of random sources, which also emulated the way the old RNG worked.

This PR removes the old RNG, and replaces the `GlobalRandomSource` by one of type `IsGAPRandomSource`. The undocumented function `RANDOM_LIST` is also removed (no distributed package uses it). To ensure backwards compatibility, we provide an implementation of `RANDOM_SEED` which uses `GlobalRandomSource` (only two packages, `recog` and `gbnp`, call `RANDOM_SEED`, and only in their package tests). Finally, dummies for `R_N` and `R_X` are retained to help detect any ancient internal code people might throw at GAP.

WIP: I noticed that this introduces a discrepancy in HPC-GAP, apparently the random seed counter is incremented once too often, or in the wrong order or something; I simply did not yet have a chance to look into it.

UPDATE: I added a fix / workaround for the random state discrepancy.

Another reason this shouldn't be merged now is that it contains some risk to break things, and is not super important, so if `stable-4.11` is branched soon, we may want to wait till after that.


ChrisJefferson:
In the past I was worried user code might access R_X and R_N -- they were documented for many years (although not any more), so I wasn't convinced it was worth "cleaning them up" when they could just be left instead.

fingolfin:
I have a hard time believing that R_X and R_N ever were used by external code, documented or not. But I see no harm in providing read-only globals with these names just in case such code does exist after all.

As to the "worth" of performing this cleanup: I did it because in the current setup, some of the low-level `Random` methods have really strange interdependencies and unwritten implicit "rules" about which has to be ranked how, which makes it rather difficult to understand, and to debug (in fact, I think @ChrisJefferson had quite some "fun" with those when working on improvements for `Random` and random sources in the past, too). With the refactored code, I believe this gets a bit simpler.

fingolfin:
@hulpke thanks for that historical insight! Well, then it is a good thing we are "protecting" `R_X` and `R_N` now in this PR, I guess

PR #3619, opened 2019-08-22
changed the definition of `Randomize`

### LINKED ISSUE #2018: `Randomize` is declared and implemented in the GAP library, but documentation is in `orb` package
I suspect `Randomize` used to be in the `orb` package but then was migrated to the GAP library -- except the documentation was not moved.

So most likely the fix is to move the documentation, too.

ThomasBreuer:
As proposed in `lib/matobj2.gd`:
For a random source `rs`,
document `Randomize( rs, obj )` instead of `Randomize( obj, rs )`,
and then use `InstallMethodWithRandomSource` to install methods.

The release notes should mention this change mainly because `Randomize` is used in a few packages.
Another aspect is that after the change, results of `Randomize` from older GAP sessions may not be reproducible because former calls of the form `Random( C )` are now executed as `Random( GlobalMersenneTwister, C )`.

Details:

- added an `<Oper Name="Random" .../>` declaration for
  random source and collection, to be used in cross-references;
  note that the `DeclareOperation` calls had covered this already.

- added `Randomize` to the documented operations for which
  the function `InstallMethodWithRandomSource` is intended

- used `InstallMethodWithRandomSource` in order to replace
  two method installations by just one (as was proposed in
  `matobj2.gd`) for all installations of `Randomize` methods
  in the GAP library

- make sure that all `Randomize` methods return the changed object

- adjusted outputs of the testfile
  `testinstall/MatrixObj/Randomize.tst`
  (one instance of the changed outputs mentioned above)

Resolves #2018

### DIFF
--- a/hpcgap/lib/vecmat.gi
+++ b/hpcgap/lib/vecmat.gi
@@ -2338,19 +2338,11 @@ InstallMethod( Vector, "for a list of gf2 elements and a gf2 vector",
     ConvertToVectorRep(r,2);
     return r;
   end );
-InstallMethod( Randomize, "for a mutable gf2 vector",
-  [ IsGF2VectorRep and IsMutable ],
-  function( v ) 
-    local i;
-    MultVector(v,0);
-    for i in [1..Length(v)] do 
-        if Random(0,1) = 1 then v[i] := Z(2); fi;
-    od;
-    return v;
-  end );
-InstallMethod( Randomize, "for a mutable gf2 vector and a random source",
-  [ IsGF2VectorRep and IsMutable, IsRandomSource ],
-  function( v, rs ) 
+
+InstallMethodWithRandomSource( Randomize,
+    "for a random source and a mutable gf2 vector",
+    [ IsRandomSource, IsGF2VectorRep and IsMutable ],
+  function( rs, v )
     local i;
     MultVector(v,0);
     for i in [1..Length(v)] do 
@@ -2358,6 +2350,7 @@ InstallMethod( Randomize, "for a mutable gf2 vector and a random source",
     od;
     return v;
   end );
+
 InstallMethod( MutableCopyMat, "for a gf2 matrix",
   [ IsGF2MatrixRep ],
   function( m )
@@ -2457,19 +2450,12 @@ end );
 
 
 
-InstallMethod( Randomize, "for a mutable gf2 matrix",
-  [IsGF2MatrixRep and IsMutable],
-  function( m )
+InstallMethodWithRandomSource( Randomize,
+    "for a random source and a mutable gf2 matrix",
+    [ IsRandomSource, IsGF2MatrixRep and IsMutable ],
+  function( rs, m )
     local v;
-    for v in m do Randomize(v); od;
-    return m;
-  end );
-
-InstallMethod( Randomize, "for a mutable gf2 matrix, and a random source",
-  [IsGF2MatrixRep and IsMutable, IsRandomSource],
-  function( m, rs )
-    local v;
-    for v in m do Randomize(v,rs); od;
+    for v in m do Randomize( rs, v ); od;
     return m;
   end );
 
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -538,26 +538,30 @@ InstallMethod( CopySubVector,
     CopySubVector(dst,dcols,src,scols);
 end );
 
-InstallMethod( Randomize,
-  "generic method for a vector",
-  [ IsVectorObj and IsMutable ],
-  function(vec)
-    local basedomain, i;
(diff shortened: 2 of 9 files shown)


hpcgap/lib/vecmat.gi
```diff
@@ -2338,26 +2338,19 @@ InstallMethod( Vector, "for a list of gf2 elements and a gf2 vector",
...
-  function( v, rs ) 
+
+InstallMethodWithRandomSource( Randomize,
+    "for a random source and a mutable gf2 vector",
+    [ IsRandomSource, IsGF2VectorRep and IsMutable ],
+  function( rs, v )
     local i;
     MultVector(v,0);
     for i in [1..Length(v)] do 
         if Random(rs,0,1) = 1 then v[i] := Z(2); fi;
```
fingolfin: Not relevant for this PR, but: this is something which probably would be several orders of magnitudes faster if done in C, as we could just generate random data in larger chunks (e.g. 64 bits at a time) and initialize the memory with it as needed, instead of iterating over each bit.

lib/matobj.gi
```diff
@@ -538,26 +538,30 @@ InstallMethod( CopySubVector,
...
-  "generic method for a vector and a random source",
-  [ IsVectorObj and IsMutable, IsRandomSource ],
-  function(vec, rs)
-    local basedomain, i;
-    basedomain := BaseDomain( vec );
-    for i in [ 1 .. Length( vec ) ] do
-        vec[ i ] := Random( rs, basedomain );
+InstallMethodWithRandomSource( Randomize,
+  "for a random source and a matrix object",
+  [ IsRandomSource, IsMatrixObj and IsMutable ],
```
fingolfin: The code this replaces was using `IsVectorObj`, and now here is `IsMatrixObj`? Am I missing something? Aha, perhaps this is simply a new function that was not present before, and the diff is therefore misleading? In that case, it would be nice if the commit message / PR description explicitly mentioned this new functionality (but it's not super important either, it's just a suggestion).

lib/matobj.gi
```diff
@@ -538,26 +538,30 @@ InstallMethod( CopySubVector,
...
+  function( rs, mat )
+    local basedomain, i, j;
+    basedomain := BaseDomain( mat );
+    for i in [ 1 .. NrRows( mat ) ] do
+      for j in [ 1 .. NrCols( mat ) ] do
+        mat[i][j]:= Random( rs, basedomain );
```
fingolfin:
```suggestion
        mat[i,j]:= Random( rs, basedomain );
```

lib/matobj.gi
```diff
@@ -654,3 +658,14 @@ InstallMethod( IsEmptyMatrix,
...
+
+
+#
+# Compatibility code: old variants of arguments (to become obsolete)
+#
+
+InstallOtherMethod( Randomize,
+    "backwards compatibility: swap arguments",
+    [ IsObject and IsMutable, IsRandomSource ],
+    { obj, rs } -> Randomize( rs, obj ) );
```
fingolfin: This compatibility method ensures that existing code *using* `Randomize` will keep working, and can thus be adapted incrementally -- very good. However, existing *implementations* will need to be adapted, else they won't get called... I don't see a good alternative to this, anyway (other than installing another method which swaps the arguments in another order, and then perhaps also a horrible hack which prevents us from doing an infinite recursion in case no specialized method is provided). Luckily, only three packages provide such methods: `cvec` and `recog` (which I'll both adapt now, and make releases), and also `meataxe64`, but that's not yet deposited and I am sure @stevelinton will be happy to adapt it, resp. we can provide a PR. (And by adapting, I mean I'll install methods for both argument orders, this way the packages will work with both old and new GAP versions).

lib/matobj2.gd
```diff
@@ -562,28 +562,38 @@ DeclareGlobalFunction( "MakeVector" );
...
+##    <Oper Arg="[Rs,]m" Name="Randomize" Label="for a matrix object"/>
 ##    <Description>
-##      Replaces every entry in <A>V</A> with a random one from the base
-##      domain. If given, the random source <A>Rs</A> is used to compute the
-##      random elements. Note that in this case, the random function for the
-##      base domain must support the random source argument.
+##      Replaces every entry in the mutable vector object <A>v</A>
+##      or matrix object <A>m</A>, respectively, with
+##      a random one from the base domain of <A>v</A> or <A>m</A>,
+##      respectively, and returns the argument.
```
fingolfin: Note that the documentation in `orb` says that nothing is returned. I prefer your variant, though, it is useful in some situations. However, this is one more thing that needs to be adapted in `recog`, `cvec` and `meataxe64`. (I'll look into doing that right now).

Issue #3538, opened 2019-07-02
Add a nice Display method for pc group


fingolfin:
It would be nice if there was an easy way to "see" the pc relation of a pc group. My suggestion would to add a `Display` (and/or `DisplayString`) method to do the job. Somewhat similar to what the `polycyclic` package offers with the `PrintPcpPresentation` command.

To illustrate what I mean, compare this:
```
gap> g:=SymmetricGroup(IsPcGroup, 4);
<pc group of size 24 with 4 generators>
gap> Display(g);
Group( [ f1, f2, f3, f4 ] )
```

with this:
```
gap> h:=SymmetricGroup(IsPcpGroup, 4);
Pcp-group with orders [ 2, 3, 2, 2 ]
gap> PrintPcpPresentation(h);
g1^2 = id
g2^3 = id
g3^2 = id
g4^2 = id
g2 ^ g1 = g2^2
g3 ^ g1 = g4
g3 ^ g2 = g4
g4 ^ g1 = g3
g4 ^ g2 = g3 * g4
```

Here is a suggestion what the display output could look like:
```
gap> Display(g)
pc-group with relations:
  g1^2 = id
  g2^3 = id
  g3^2 = id
  g4^2 = id
  g2 ^ g1 = g2^2
  g3 ^ g1 = g4
  g3 ^ g2 = g4
  g4 ^ g1 = g3
  g4 ^ g2 = g3 * g4
```

(Incidentally, I think that `polycylic` also should have custom `Display` method which shows the presentation).

wucas:
Can this be closed since #3621 git merged?

DominikBernhardt:
I'll close this as #3621 is merged.

PR #3621, opened 2019-08-23
Add Display method for pc groups

flyingapfopenguin:
# Description / Text for release notes

Add Display method for pc groups, refering #3538.

### DIFF
--- a/lib/grppc.gd
+++ b/lib/grppc.gd
@@ -521,6 +521,12 @@ DeclareGlobalFunction( "GapInputPcGroup" );
 
 #############################################################################
 ##
+#F  PrintPcPresentation( <grp>, <commBool> )
+##
+DeclareGlobalFunction( "PrintPcPresentation" );
+
+#############################################################################
+##
 #O  CanonicalSubgroupRepresentativePcGroup( <G>, <U> )
 ##
 ##  <ManSection>
--- a/lib/grppc.gi
+++ b/lib/grppc.gi
@@ -1852,6 +1852,96 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
 
 end );
 
+#############################################################################
+##
+#F  PrintPcPresentation( <grp>, <commBool> )
+##
+##  Display the pc relations of a pc group.
+##  If <commBool> is true, then the commutator presentation is printed,
+##  otherwise the power presentation.
+##  Trivial commutators / powers are not printed.
+##  The generators are named "g<i>".
+##  The returned boolean indicates if there are commutating generators.
+##
+InstallGlobalFunction( PrintPcPresentation, function(G, commBool)
+    local pcgs, n, F, gens, i, pis, exp, t, h, rel, commPower, j, trivialCommutators;
+
+    pcgs:=Pcgs(G);
+    n:=Length(pcgs);
+    F    := FreeGroup( n, "g" );
+    gens := GeneratorsOfGroup( F );
+    pis  := RelativeOrders( pcgs );
+
+    # compute the orders of the pc-generators
+    for i in [1..n] do
+        exp := ExponentsOfRelativePower( pcgs, i ){[i+1..n]};
+        t   := One( F );
+        for h in [i+1..n] do
+            t := t * gens[h]^exp[h-i];
+        od;
+        if IsOne( t ) then
+            t := "id";
+        fi;
+        Print(gens[i], "^", pis[i], " = ", t, "\n");
+    od;
+
+    # compute the commutators / conjugation
+    # of all pairs of pc-generators
+    trivialCommutators := false;
+    for i in [1..n] do
+        for j in [i+1..n] do
+            if pcgs[j] * pcgs[i] = pcgs[i] * pcgs[j] then
+                trivialCommutators := true;
+                continue;
+            fi;
+            if commBool then
+                commPower := Comm( pcgs[j], pcgs[i] );
+            else
+                commPower := pcgs[j]^pcgs[i];
+            fi;
+            exp := ExponentsOfPcElement( pcgs, commPower ){[i+1..n]};
+            t   := One( F );
+            for h in [i+1..n] do
+                t := t * gens[h]^exp[h-i];
+            od;
+            if commBool then
+                Print("[", gens[j], ",", gens[i] , "]");
+            else
+                Print(gens[j], "^", gens[i]);
+            fi;
+            Print(" = ", t, "\n");
+        od;
+    od;
+
+    return trivialCommutators;
+end );
+
+#############################################################################
+##
+#M  Display( <grp> )
+##
+InstallMethod( Display,
+    "for a pc group",
+    [ IsPcGroup ],
+    function( G )
+        local n, trivialCommutators;
+        if IsTrivial(G) then
+            Print("trivial pc-group\n");
+            return;
+        fi;
+        n := Size(Pcgs(G));
+        if IsOne(n) then
+            Print("cyclic pc-group with 1 pc-generator and the relation:\n");
+        else
+            Print("pc-group with ", Size(Pcgs(G)), " pc-generators and relations:\n");
+        fi;
+        trivialCommutators := PrintPcPresentation( G, false );
+        if IsAbelian(G) then
+          Print("all generators commute, the groups is abelian\n");
+        elif trivialCommutators then
+          Print("all other pairs of generators commute\n");
+        fi;
+    end );
 
 #############################################################################
 ##
--- a/tst/testinstall/grppc.tst
+++ b/tst/testinstall/grppc.tst
@@ -1,16 +1,83 @@
 #@local F,G,S,c,cl,f,g,g1,g10,g3,gens,h,hh,i,m,n,pcgs,r,rws,sys,u,v,x,y
 gap> START_TEST("grppc.tst");
+gap> Display(TrivialGroup(IsPcGroup));
+trivial pc-group
+gap> Display(CyclicGroup(IsPcGroup, 10));
+pc-group with 2 pc-generators and relations:
+g1^2 = g2
+g2^5 = id
+all generators commute, the groups is abelian
+gap> Display(SymmetricGroup(IsPcGroup, 3));
+pc-group with 2 pc-generators and relations:
+g1^2 = id
+g2^3 = id
+g2^g1 = g2^2
 gap> h:=Group((1,2,3,4),(1,2));;
 gap> m:=IsomorphismPcGroup(h);;
 gap> hh:=Image(m,h);;
+gap> Display(hh);
+pc-group with 4 pc-generators and relations:
+g1^2 = id
+g2^3 = id
+g3^2 = id
+g4^2 = id
+g2^g1 = g2^2
+g3^g1 = g4
+g4^g1 = g3
+g3^g2 = g4
+g4^g2 = g3*g4
+all other pairs of generators commute
 gap> pcgs:=Pcgs(hh);;
 gap> ForAll(pcgs,i->PreImagesRepresentative(m,i) in h);
 true
 gap> g:=WreathProduct(Group((1,2,3),(1,2)),Group((1,2,3,4,5,6,7)));;
 gap> i:=IsomorphismPcGroup(g);;
 gap> g:=Range(i);;
+gap> Display(g);
+pc-group with 15 pc-generators and relations:
+g1^7 = id
+g2^2 = id
+g3^2 = id
+g4^2 = id
+g5^2 = id
+g6^2 = id
+g7^2 = id
+g8^2 = id
+g9^3 = id
+g10^3 = id
+g11^3 = id
+g12^3 = id
+g13^3 = id
+g14^3 = id
+g15^3 = id
+g2^g1 = g8
+g3^g1 = g2
+g4^g1 = g3
+g5^g1 = g4
+g6^g1 = g5
+g7^g1 = g6
+g8^g1 = g7
+g9^g1 = g15
+g10^g1 = g9
+g11^g1 = g10
+g12^g1 = g11
+g13^g1 = g12
+g14^g1 = g13
+g15^g1 = g14
+g9^g2 = g9^2
+g10^g3 = g10^2
+g11^g4 = g11^2
+g12^g5 = g12^2
+g13^g6 = g13^2
+g14^g7 = g14^2
+g15^g8 = g15^2
+all other pairs of generators commute
 gap> u:=Subgroup(g,GeneratorsOfGroup(g){[2..15]});;
 gap> n:=Subgroup(g,[g.1]);;
+gap> Display(n);
+cyclic pc-group with 1 pc-generator and the relation:
+g1^7 = id
+all generators commute, the groups is abelian
 gap> v:=Normalizer(u,n);;
 gap> IsSubgroup(u,v);
 true
@@ -42,6 +109,27 @@ gap> g:=Group((1,15,8,4,14,9)(2,16,7,3,13,10)(5,18,12)(6,17,11),
 > (1,3,6)(2,4,5)(7,9,12)(8,10,11)(13,15,18)(14,16,17),
 > (5,6)(7,8)(9,10)(13,14)(15,16),(1,2)(7,8)(13,14),(1,2)(3,4)(5,6),
 > (7,8)(9,10)(11,12),(13,14)(15,16)(17,18));;
+gap> Display(Image(IsomorphismPcGroup(g)));
+pc-group with 8 pc-generators and relations:
+g1^2 = id
+g2^2 = id
+g3^3 = id
+g4^2 = id
+g5^2 = id
+g6^3 = id
+g7^2 = id
+g8^2 = id
+g6^g1 = g6*g8
+g3^g2 = g3^2
+g4^g2 = g5
+g5^g2 = g4
+g4^g3 = g4*g5*g7*g8
+g5^g3 = g4
+g6^g4 = g6*g8
+g6^g5 = g6*g8
+g7^g6 = g8
+g8^g6 = g7*g8
+all other pairs of generators commute
 gap> cl:=ConjugacyClasses(Image(IsomorphismPcGroup(g),g));;
 gap> G := Group( ( 1, 2, 3, 4, 5, 6, 7, 8, 9,10,11,12,13,14), (15,16) );;
 gap> sys := SylowSystem( G );;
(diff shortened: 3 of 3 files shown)


stevelinton:
Is there anything to be said for omitting relations of the form `g17^g15 = g17` from the printout and instead appending `all other pairs of generators commute`.  It is quite common to have presentations which are "sparse" in the sense that most pairs of generators commute.

flyingapfopenguin:
@stevelinton Thanks for the reply. I added the message to the output of the Display function.

wilfwilson:
The tests need to be updated to account for the new line that you added.

Also, I think that the line `All other pairs of generators commute.` should only be shown if there _are_ actually such pairs of generators that commute. (I don't know anything about pc groups really, so perhaps it is possible that there are always some generators that commute).

fingolfin:
@flyingapfopenguin you added the message, but did not update the tests (which you can do by running `Test("tst/PATH/TO/TEST.tst", rec(rewriteToFile:=true));`

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
 
 end );
 
+#############################################################################
+##
+#F  PrintPcPresentation( <grp>, <commBool> )  . . . . . . .  gap input string
+##
+##  Display the pc relations of a pc group.
+##  If commBool the commutator presentation is printed,
```
fingolfin:
```suggestion
##  If <commBool> is true, then the commutator presentation is printed,
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+##  otherwise the power presentation.
+##  Trivial commutators / powers are not printed.
+##  The generators  will  be  named "g<i>".
```
fingolfin:
```suggestion
##  The generators are named "g<i>".
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+#############################################################################
+##
+#M  Display( <grp> )
+##
+InstallMethod( Display,
+    "for a pc group",
+    [ IsPcGroup ],
+    function( G )
+        if IsTrivial(G) then
+            Print("Trivial Group\n");
```
fingolfin: See above for notes on the message

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+            return;
+        fi;
+        Print("pc-group with relations:\n");
+        PrintPcPresentation( G, false );
+        Print("All other pairs of generators commute.\n");
```
fingolfin: Why are these two `Print` calls are not part of `PrintPcPresentation`? (I am not saying they should be, I honestly wonder about the reasoning).

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+InstallMethod( Display,
+    "for a pc group",
+    [ IsPcGroup ],
+    function( G )
+        if IsTrivial(G) then
+            Print("Trivial Group\n");
+            return;
+        fi;
+        Print("pc-group with relations:\n");
+        PrintPcPresentation( G, false );
```
fingolfin: Some code in GAP uses spaces after opening parens, some not; I don't care much either way, but I'd recommend to at least stay consistent within the code of a single functions. So I'd suggest to adapt this call to the other code in this function:
```suggestion
        PrintPcPresentation(G, false);
```

tst/testinstall/grppc.tst
```diff
@@ -1,16 +1,70 @@
...
 gap> h:=Group((1,2,3,4),(1,2));;
 gap> m:=IsomorphismPcGroup(h);;
 gap> hh:=Image(m,h);;
+gap> Display(hh);
+pc-group with relations:
+g1^2 = id
+g2^3 = id
+g3^2 = id
+g4^2 = id
+g2 ^ g1 = g2^2
```
fingolfin: Why `g2 ^ g1` and not `g2^g1`?  Or conversely: Why `g1^2` and not `g1 ^ 2`?

tst/testinstall/grppc.tst
```diff
@@ -1,16 +1,70 @@
 #@local F,G,S,c,cl,f,g,g1,g10,g3,gens,h,hh,i,m,n,pcgs,r,rws,sys,u,v,x,y
 gap> START_TEST("grppc.tst");
+gap> Display(TrivialGroup());
```
fingolfin: This implicitly relies on `TrivialGroup` defaulting to create a pc group, which could change. So better to do this instead:
```suggestion
gap> Display(TrivialGroup(IsPcGroup));
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+    gens := GeneratorsOfGroup( F );
+    pis  := RelativeOrders( pcgs );
+
+    # the power
+    for i in [1..n] do
+        exp := ExponentsOfRelativePower( pcgs, i ){[i+1..n]};
+        t   := One( F );
+        for h in [i+1..n] do
+            t := t * gens[h]^exp[h-i];
+        od;
```
fingolfin: I think (but have not verified) that the previous four lines can be replaced by just this: `t := PcElementByExponentsNC( pcgs, exp );`. Similar elsewhere in this function.

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+        if t = One( F ) then
```
fingolfin:
```suggestion
        if IsOne( t ) then
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+            Print(gens[i], "^", pis[i], " = id\n");
+        else
+            Print(gens[i], "^", pis[i], " = ", t, "\n");
+        fi;
```
fingolfin: You could simplify the above like this:
```
  if IsOne(t) then t := "id"; fi;
  Print(gens[i], "^", pis[i], " = ", t, "\n");
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,87 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+            od;
+            if commBool then
+                if t = One( F ) then
+                    continue;
+                fi;
+                Print("[ ", gens[j], ", ", gens[i] , " ] = ", t, "\n");
+            else
+                if t = gens[j] then
+                    continue;
+                fi;
```
fingolfin: Instead of effectively checking this twice, you could insert the following at the start of the inner loop:
```gap
  if pcgs[j] * pcgs[i] = pcgs[i] * pcgs[j] then
    continue;
  fi;
```

flyingapfopenguin:
@wilfwilson @fingolfin Thanks for your constructive response. I applied (most of) your suggestions.

lib/grppc.gi
flyingapfopenguin: I simply tested this and its not correct for e.g. C10.
From the test output:

> ########> Diff in tst/testinstall/grppc.tst:5
> # Input is:
> Display(CyclicGroup(IsPcGroup, 10));
> # Expected output:
> pc-group with relations:
> g1^2 = g2
> g2^5 = id
> All other pairs of generators commute.
> # But found:
> pc-group with relations:
> g1^2 = f1
> g2^5 = id
> g2 ^ g1 = f1
> ########

lib/grppc.gi
flyingapfopenguin: The variable t is an element of the free group and not a string.

lib/grppc.gi
flyingapfopenguin: Sorry, I don't get it.

tst/testinstall/grppc.tst
flyingapfopenguin: It's "g2^2" on the right side because gap prints terms of generators this way. (So this is not easy to change.)
The output of the left is constructed manually - the way @fingolfin wanted to have it, see #3538.

lib/grppc.gi
flyingapfopenguin: I just wanted to create more reusable code by writing on function that - only - prints the non trivial relations.

lib/grppc.gi
fingolfin: That comment of mine was simply nonsense, sorry.

lib/grppc.gi
fingolfin: Sure. But we are free to change it to anything, including the string `"id"`, to achieve our goals.

lib/grppc.gi
fingolfin: Seems a comment got lost: I originally commented on the other place where you print `Trivial Group` that this looses valuable information, namely that one is looking at a pc group. Also, the capitalization is off, we use lowercase elsewhere. So I'd suggest `trivial pc-group` as message instead.

Also, I find it odd that this message is printed twice, once here and once inside `PrintPcPresentation` -- why the duplication?

lib/grppc.gi
fingolfin: In that case, it should print nothing if given a trivial group, instead of "Trivial Group"

tst/testinstall/grppc.tst
```diff
@@ -52,6 +139,11 @@ gap> List(sys,i->Length(AsList(i)));
 gap> G := SmallGroup( 144, 183 );;
 gap> F := FittingSubgroup( G );;
 gap> S := SylowSubgroup( F, 2 );;
+gap> Display(Image(IsomorphismPcGroup(S)));
+pc-group with relations:
+g1^2 = id
+g2^2 = id
+All other pairs of generators commute.
```
fingolfin: Note that no pairs of generators were printed, making this message in this particular situation somewhat confusion (even though it is of course formally correct). Perhaps one could refine the message a bit, by doing something like  `if IsAbelian(grp) then Print("all generators commute, the groups is abelian\n") else ...` or so. 

Of course this is super nitpicky. I am also OK with it if it stays as it is.

lib/grppc.gi
```diff
@@ -1852,6 +1852,92 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
 
 end );
 
+#############################################################################
+##
+#F  PrintPcPresentation( <grp>, <commBool> )  . . . . . . .  gap input string
```
fingolfin: "gap input string" makes no sense here...
```suggestion
#F  PrintPcPresentation( <grp>, <commBool> )
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,92 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+##
+##  Display the pc relations of a pc group.
+##  If <commBool> is true, then the commutator presentation is printed,
+##  otherwise the power presentation.
+##  Trivial commutators / powers are not printed.
+##  The generators are named "g<i>".
+##  The returned boolean indicate if there are commutating generators.
```
DominikBernhardt:
```suggestion
##  The returned boolean indicates if there are commutating generators.
```

lib/grppc.gi
```diff
@@ -1852,6 +1852,92 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+    n:=Length(pcgs);
+    if n=0 then
+        Print("Trivial Group\n");
+        return false;
+    fi;
+    F    := FreeGroup( n, "g" );
+    gens := GeneratorsOfGroup( F );
+    pis  := RelativeOrders( pcgs );
+
+    # the power
```
DominikBernhardt: Could you maybe improve this comment so it says more about the following code?

DominikBernhardt:
One general remark: You print the message `All other pairs of generators commute` but the user never sees the generators. Should one maybe print the _number of generators_ for the pc group as well?

lib/grppc.gi
flyingapfopenguin: I'm going to take this over.

The duplication, which is just a duplication in code and not in any output, was implemented to avoid an empty output of `PrintPcPresentation(TrivialGroup(), true);`. This is now the case but I can live with that.

tst/testinstall/grppc.tst
flyingapfopenguin: I'm just going to skip the spaces on the left side of the equations.

lib/grppc.gi
flyingapfopenguin: For avoiding a single "else" statement I will not implement a type mismatch.

lib/grppc.gi
flyingapfopenguin: I tried to improve it a little. Do you want more?

lib/grppc.gi
fingolfin: What do you mean with "implement a type mismatch"?

Anyway, the idea of course as to get rid of the code duplication, and the subtle bugs that tend to creep in because of it. But this code is trivial enough, so I won't loose sleep over it being either way :-)

lib/grppc.gd
```diff
@@ -519,6 +519,18 @@ DeclareGlobalFunction(
...
 
+#############################################################################
+##
+#F  PrintPcPresentation( <grp>, <commBool> )
+##
+##  <#GAPDoc Label="PrintPcPresentation">
+##  <ManSection>
+##  <Func Name="PrintPcPresentation" Arg='grp, commBool'/>
+##  </ManSection>
+##  <#/GAPDoc>
```
fingolfin: Note that this GAPDoc comment won't appear in the manual, as it would have to be referenced there explicitly. But that's fine, as we do not want to explicitly document this function. I just wanted to point this out -- the above 5 lines could be safely deleted.

But *if* we keep them, it is weird that the actual content which would normally go inside the `<Description>` of this `<ManSection>` is not here, but rather in the `.gi` file. That seems to be neither here nor there. So I'd either remove the GAPDoc here; or else move the actual description text from the `.gi` file into this comment.

lib/grppc.gi
```diff
@@ -1852,6 +1852,96 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+##  otherwise the power presentation.
+##  Trivial commutators / powers are not printed.
+##  The generators are named "g<i>".
+##  The returned boolean indicates if there are commutating generators.
+##
+InstallGlobalFunction( PrintPcPresentation, function(G, commBool)
+    local pcgs, n, F, gens, i, pis, exp, t, h, rel, commPower, j, trivialCommutators;
+
+    pcgs:=Pcgs(G);
+    n:=Length(pcgs);
```
fingolfin: This is not a request for change, just wanted to point something out: the above two lines clearly come from another source than the following three lines, based on their formatting... Overall, this function mixes several code formatting styles (e.g. `Foo( arg )` vs. `Foo(arg)`)

lib/grppc.gi
```diff
@@ -1852,6 +1852,96 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+    [ IsPcGroup ],
+    function( G )
+        local n, trivialCommutators;
+        if IsTrivial(G) then
+            Print("trivial pc-group\n");
+            return;
+        fi;
+        n := Size(Pcgs(G));
+        if IsOne(n) then
+            Print("cyclic pc-group with one pc-generator and the relation:\n");
```
fingolfin: Unimportant nitpick: So you print "one pc-generator" but "2 pc-generators" ? maybe "one" -> "1" for consistency?

I am not sure I see the importance of adding `cyclic`. It might even be misleading, because many cyclic pc-groups won't get it (e.g. C_4 or C_6 if the pcgs is refined).

lib/grppc.gi
```diff
@@ -1852,6 +1852,96 @@ InstallGlobalFunction( GapInputPcGroup, function(U,name)
...
+        else
+            Print("pc-group with ", Size(Pcgs(G)), " pc-generators and relations:\n");
+        fi;
+        trivialCommutators := PrintPcPresentation( G, false );
+        if IsAbelian(G) then
+          Print("all generators commute, the groups is abelian\n");
```
fingolfin: Again unimportant, but: If you handle the 1 generator case separately above, you might want to consider doing so here as well, as "all generators" then refers to just 1 generator.

lib/grppc.gi
flyingapfopenguin: I explicitly learned to not code that way. As you as an GAP export don't care that doesn't seem to be conflicting with the guidelines used in this project. So I'm going to change this.

lib/grppc.gd
flyingapfopenguin: Thanks for the replay. I have no idea how GAPDoc works and just tried to do the "necessary" documentation by (educated) copy and paste. I'm just deleting these lines.

lib/grppc.gi
flyingapfopenguin: @fingolfin What is the GAP coding style? As I didn't knew I copied code without changing its formating and coded the rest like it seems to make sense to me.

lib/grppc.gi
flyingapfopenguin: That's right, but all is one in this case.

lib/grppc.gi
flyingapfopenguin: I'm going to change "one" to "1" for consistency.

About "cyclic": I don't think this is important in any way but i'm going to leave this.

flyingapfopenguin:
> Overall seems fine to me now, though I had some further minor nitpicks. The PR should also be squashed.

Should I simply squash all commits together and do a force push or how is this usually done?
If I anyhow do a force push, should I also rebase on the current master?

fingolfin:
@flyingapfopenguin yes, please squash, rebase on latest master and then force push.

flyingapfopenguin:
> @flyingapfopenguin yes, please squash, rebase on latest master and then force push.

done

DominikBernhardt:
One test failed due to a download error, I restarted it.

lib/grppc.gi
fingolfin: There is no official coding style (and trying to set one is one of these things were I expect a lot of pain and blood for little gain). Many people seem to take the formatting which GAP itself uses when printing a function as reference, but usually with some adjustments (e.g. aligning some things across multiple lines, as below).

Anyway, I am not asking for a specific style to be followed -- just that the code within a given function be self-consistent. That's note the case here.

flyingapfopenguin: I'm going to fix this tomorrow by adding a new commit, so that you can do a reintegration merge afterwards.

fingolfin: I already merged this PR, didn't want to hold it hostage over some whitespace differences :-) I only pointed this formatting bit out again to make you aware of it; as it is, lots of GAP code is less well formatted than what is in this PR...

wucas:
Is there a reason why this PR does not close Issue #3538?

DominikBernhardt:
Is `refering` maybe not the right keyword? Could be that one has to write `resolves` or `fixes`? I can close the original issue manually this time.

PR #3623, opened 2019-08-24
ENHANCE: Work around ErrorQuit in abelianized Rrs

hulpke:
The Abelianized Rrs uses kernel routines that are hard-coded for
small integers. If an integer becomes larger, they used to exit with
ErrorQuit, thus not making it possible to recover from this situation in a
larger calculation. This commit changes this behavior:

1) Instead of callign `ErrorQuit` a failure value is returned by the inner
kernel routines, the outer kernel routines accordingly have `ApplyRel2`
(which used to return no value) returns `true` or `false` accordingly.
2) The code for abelianized Rrs catches this case and changes to a (slow)
library calculation based on the nun-abelianiszed Rrs. Not great, but at
least no error
3) The "cheap" option skips such hard calculation (with a return value of
`fail`). This is recognized in `LargerQuotientBySubgroupAbelianization` and in
turn by `FpGroupCocycle` to continue a candidate search, even if
encountering such a situation.

I have an (rather convoluted since the circumstances are difficult) example, but the file is os size 100k and the example will run for hours to utilize the new code. I welcome suggestions on whether this should be included (and if so in which way).


lib/sgpres.gi
```diff
@@ -1434,6 +1437,30 @@ InstallGlobalFunction( RelatorMatrixAbelianizedSubgroupRrs, function ( G, H )
...
     aug.subgroupRelators := RewriteAbelianizedSubgroupRelators( aug,
                              aug.groupRelators);
+    if aug.subgroupRelators=fail then
+      if ValueOption("cheap")=true then 
+        return fail;
+      fi;
+      # the abelianized rewriting in the kernel failed because the
+      # coefficients were to large.
+      Info(InfoWarning,1,"exponent too large, abelianized MTC aborted");
+      Info(InfoWarning,1,"calcuilation will be slow");
```
wilfwilson:
```suggestion
      Info(InfoWarning,1,"calculation will be slow");
```

src/costab.c
```diff
@@ -1028,6 +1030,7 @@ static void AddCosetFactor2 (
             leng  = LEN_PLIST(tmp);
             for ( i = 1;  i <= leng;  i++ ) {
                 if ( ! SUM_INTOBJS( sum, ptWord[i], ptFac[i] ) ) {
+                    return 1;
                     ErrorQuit(
```
wilfwilson: Maybe delete the `ErrorQuit`, and turn its message string into a comment? The same applies about 15 lines below.

src/costab.c
```diff
@@ -1008,8 +1008,10 @@ static Int TreeEntryC ( void )
 **  variables, and 'treeType' is assumed to be either 0 or 2,
 **
 **  Warning: 'factor' is not checked for being zero.
+**
+**  it returns 0 if everything worked, and 1 if an error arose.
```
wilfwilson: Maybe write ‘problem’ instead of ‘error’, like you did below. ('Error' to me kind of suggests that something _incorrect_ has happened, whereas this is more that something annoying has happened, if I understand correctly.).
```suggestion
**  it returns 0 if everything worked, and 1 if a problem arose.
```

Issue #1191, opened 2017-03-16
Add a "strict" mode


fingolfin:
Let's add a strict mode (like e.g. perl has) which warns resp. errors on extra things:

* if a function returns sometimes an obj and sometimes nothing (see issue #466)
* invalid backslash escaped chars inside strings (see issue #619)
* ...

This mode could be enabled or disabled via a command line switch.

fingolfin:
So to elaborate a bit, here is what I'd do:
* add a global boolean flag to the kernel, say `StrictMode`, which defaults to 0 / false.
* add a command line option `--strict` which changes the default to 1 / true (can be added in `src/system.c` next to the other command line options)
* then look at the code in the two pull requests referenced above; try to adapt their patches so that their warnings are only printed if `StrictMode` is 1 / true.

fingolfin:
Another possible feature: warn if a function has unused locals (i.e. local variables that are never read from; so completely unused, or only ever assigned to)

fingolfin:
I also have code that prints a warning when a function contains an unused local.

wilfwilson:
This sounds nice.

PR #3628, opened 2019-08-28
Add Info statement when assigning already assigned Attributes

ChrisJefferson:
An occasional source of subtle GAP bugs is that assignments to attributes are silently ignored, after they are assigned:

```
gap> g := Group(());
gap> SetTransitivity(g, 4);
gap> Transitivity(g);
4
gap> SetTransitivity(g, 3);
gap> Transitivity(g);
4
```

Now, for properties this is checked always, and for specifically `Size` this is checked when the assertionlevel is high. However, it isn't checked in general.

This PR adds checking when the AssertionLevel is 2. I don't add it in general because it could be the case someone assigns an attribute to a value which is the same, but where = is not implemented, or extremely slow.

If anyone is interested, I has a look at when attributes are assigned the same value (as this could be considered inefficient). It turns out it happens quite a bit for what seem to me to be reasonable reasons (sometimes the same simple property, like IsTrivial, can be assigned for a number of different reasons. It would be more expensive to check if it was already set).

### DIFF
--- a/lib/attr.gd
+++ b/lib/attr.gd
@@ -36,10 +36,14 @@
 ##  <InfoClass Name="InfoAttributes"/>
 ##
 ##  <Description>
-##  This info class (together with <Ref InfoClass="InfoWarning"/> is used
-##  for messages about attribute storing  being  disabled  (at  level  2)  or
-##  enabled (level 3). It may be  used  in  the  future  for  other  messages
-##  concerning changes to attribute behaviour.
+##  This info class (together with <Ref InfoClass="InfoWarning"/>) is used
+##  for messages about attributes. Messages are shown under the following circumstances:
+##  <List>
+##    <Item> <Ref Func="EnableAttributeValueStoring"/> is used (level 2).</Item>
+##    <Item> <Ref Func="DisableAttributeValueStoring"/> is used (level 3).</Item>
+##    <Item> When trying to assign to non-mutable attribute which already is set to a different value (level 3).</Item>
+##    <Item> When the test filter for an attribute (i.e., <C>HasFOO</C>) is set, but no value is assigned (level 3).</Item>
+##  </List>
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
--- a/lib/attr.gi
+++ b/lib/attr.gi
@@ -37,3 +37,13 @@ InstallGlobalFunction(DisableAttributeValueStoring, function( attr )
     Info(InfoAttributes + InfoWarning, 2, "Disabling value storing for ",NAME_FUNC(attr));
     SET_ATTRIBUTE_STORING( attr, false);
 end);
+
+CHECK_REPEATED_ATTRIBUTE_SET := function(obj, name, val)
+    if InfoLevel(InfoAttributes) >= 3 then
+        if not IsBound(obj!.(name)) then
+            Info(InfoAttributes, 3, "Attribute ", name, " of ", obj, " is marked as assigned, but it has no value");
+        elif obj!.(name) <> val then
+            Info(InfoAttributes, 3, "Attribute ", name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
+        fi;
+    fi;
+end;
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -3105,13 +3105,7 @@ InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
 function(obj,sz)
 local filt;
   if HasSize(obj) and Size(obj)<>sz then
-    if AssertionLevel()>2 then
-      # Make this an ordinary error (not ErrorNoReturn as suggested) to
-      # preserve all debugging options -- even use `return` to investigate
-      # what would have happened before this methods was introduced.
-      Error("size of ",obj," already set to ",Size(obj),
-        ", cannot be changed to ",sz);
-    fi;
+    CHECK_REPEATED_ATTRIBUTE_SET(obj, "Size", sz);
     return;
   fi;
   if sz=0 then filt:=IsEmpty;
--- a/lib/oper1.g
+++ b/lib/oper1.g
@@ -1090,3 +1090,7 @@ InstallMethod( ViewObj,
     [ IS_OBJECT ],
     0,
     PRINT_OBJ );
+
+# A dummy version of this function is installed here, the full version
+# is defined in attr.gi, after Info-related functionality is set up.
+CHECK_REPEATED_ATTRIBUTE_SET := function(obj, name, val) end;
--- a/src/opers.c
+++ b/src/opers.c
@@ -1385,6 +1385,7 @@ static UInt RNamIsVerbose;
 static UInt RNamIsConstructor;
 static UInt RNamPrecedence;
 static Obj  HANDLE_METHOD_NOT_FOUND;
+static Obj  CHECK_REPEATED_ATTRIBUTE_SET;
 
 static void HandleMethodNotFound(Obj   oper,
                                  Int   nargs,
@@ -3283,12 +3284,15 @@ static Obj DoSetterFunction(Obj self, Obj obj, Obj value)
     flag2  = INT_INTOBJ( FLAG2_FILT(tester) );
     type   = TYPE_OBJ_FEO(obj);
     flags  = FLAGS_TYPE(type);
+
+    UInt rnam = (UInt)INT_INTOBJ(ELM_PLIST(tmp, 1));
+
     if ( SAFE_C_ELM_FLAGS(flags,flag2) ) {
+        CALL_3ARGS(CHECK_REPEATED_ATTRIBUTE_SET, obj, NAME_RNAM(rnam), value);
         return 0;
     }
 
     /* set the value                                                       */
-    UInt rnam = (UInt)INT_INTOBJ(ELM_PLIST(tmp,1));
 #ifdef HPCGAP
     if (atomic)
       SetARecordField( obj, rnam, CopyObj(value,0) );
@@ -3771,6 +3775,9 @@ static Int InitKernel (
 
     ImportFuncFromLibrary("HANDLE_METHOD_NOT_FOUND",
                           &HANDLE_METHOD_NOT_FOUND);
+
+    ImportFuncFromLibrary("CHECK_REPEATED_ATTRIBUTE_SET",
+                          &CHECK_REPEATED_ATTRIBUTE_SET);
 
 #ifdef GASMAN
     ImportGVarFromLibrary( "IsType", &IsType );
--- a/tst/testinstall/attribute.tst
+++ b/tst/testinstall/attribute.tst
@@ -1,4 +1,6 @@
 gap> START_TEST("attribute.tst");
+gap> attributeinfo := InfoLevel(InfoAttributes);;
+gap> SetInfoLevel(InfoAttributes, 3);
 gap> DeclareAttribute();
 Error, Function: number of arguments must be at least 2 (not 0)
 gap> DeclareAttribute("banana");
@@ -60,6 +62,19 @@ gap> HasSize(foo);
 true
 gap> Size(foo);
 17
+gap> SetSize(foo, 16);
+#I  Attribute Size of <object> already set to 17, cannot be changed to 16
+gap> Size(foo);
+17
+gap> HasDimension(foo);
+false
+gap> SetDimension(foo, 3);
+gap> HasDimension(foo);
+true
+gap> SetDimension(foo, 4);
+#I  Attribute Dimension of <object> already set to 3, cannot be changed to 4
+gap> Dimension(foo);
+3
 gap> InstallMethod(FavouriteFruit, [HasSize], x-> "pear");
 gap> FavouriteFruit(foo);
 "pear"
@@ -70,5 +85,22 @@ gap> FavouriteFruit(foo);
 "pear"
 gap> HasFavouriteFruit(foo);
 true
+gap> SetFavouriteFruit(foo, "apple");
+#I  Attribute FavouriteFruit of <object> already set to pear, cannot be changed to apple
+gap> FavouriteFruit(foo);
+"pear"
+gap> Unbind(foo!.FavouriteFruit);
+gap> SetFavouriteFruit(foo, "apple");
+#I  Attribute FavouriteFruit of <object> is marked as assigned, but it has no value
 #@fi
+
+# Check a mutable attribute
+gap> grp := Group(());;
+gap> SetStabChainMutable(grp, [1,2]);
+gap> StabChainMutable(grp);
+[ 1, 2 ]
+gap> SetStabChainMutable(grp, [3,4]);
+gap> StabChainMutable(grp);
+[ 3, 4 ]
+gap> SetInfoLevel(InfoAttributes, attributeinfo);
 gap> STOP_TEST("attribute.tst", 1);
--- a/lib/module.gi
+++ b/lib/module.gi
@@ -415,7 +415,7 @@ InstallOtherMethod(SetDimension,true,[IsObject and IsAttributeStoringRep,IsObjec
 function(obj,dim)
 local filt;
   if HasDimension(obj) and IsBound(obj!.Dimension) then
-    Assert(2, Dimension(obj) = dim);
+    CHECK_REPEATED_ATTRIBUTE_SET(obj, "Dimension", dim);
     return;
   fi;
   if dim=0 then
--- a/src/c_oper1.c
+++ b/src/c_oper1.c
@@ -1,7 +1,7 @@
 #ifndef AVOID_PRECOMPILED
 /* C file produced by GAC */
 #include "compiled.h"
-#define FILE_CRC  "-77878372"
+#define FILE_CRC  "-76837766"
 
 /* global variables used in handlers */
 static GVar G_REREADING;
@@ -24,6 +24,7 @@ static GVar G_SET__NAME__FUNC;
 static Obj  GF_SET__NAME__FUNC;
 static GVar G_NARG__FUNC;
 static Obj  GF_NARG__FUNC;
+static GVar G_CHECK__REPEATED__ATTRIBUTE__SET;
 static GVar G_IS__OPERATION;
 static Obj  GF_IS__OPERATION;
 static GVar G_AINV;
@@ -181,7 +182,7 @@ static RNam R_CommandLineOptions;
 static RNam R_N;
 
 /* information for the functions */
-static Obj  NameFunc[19];
+static Obj  NameFunc[20];
 static Obj FileName;
 
 /* handler for function 2 */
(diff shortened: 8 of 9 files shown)


ChrisJefferson:
I'd be interested in seeing this run over the packages, to see if anything fails.

ChrisJefferson:
I'm looking at the instances where this is failing (there are a bunch of such cases, I am looking into special casing, or removing, them.

fingolfin:
I like the idea, but having it on assertion level 2 might be too aggressive, as test files using `START_TEST` are on that assertion level, and so it'd automatically affect all packages. Of course if we can find and fix all problematic tests in all packages, and get those packages to release new versions with those changes, we could do it...

But on level 2, this definitely shouldn't be merged before GAP 4.11 is branched

src/opers.c
```diff
@@ -3856,6 +3860,7 @@ static Int InitKernel (
     ImportGVarFromLibrary( "TYPE_FLAGS", &TYPE_FLAGS );
     TypeObjFuncs[ T_FLAGS ] = TypeFlags;
 
+
```
fingolfin:
```suggestion
```

src/opers.c
```diff
@@ -3389,12 +3390,15 @@ static Obj DoSetterFunction(Obj self, Obj obj, Obj value)
     flag2  = INT_INTOBJ( FLAG2_FILT(tester) );
     type   = TYPE_OBJ_FEO(obj);
     flags  = FLAGS_TYPE(type);
+
+    UInt rnam = (UInt)INT_INTOBJ(ELM_PLIST(tmp,1));
+
     if ( SAFE_C_ELM_FLAGS(flags,flag2) ) {
+        CALL_3ARGS(CHECK_REPEATED_ATTRIBUTE_SET, obj, NAME_RNAM(rnam), value);
```
fingolfin: Just wondering: This code isn't run for mutable attributes, is it?

lib/oper1.g
```diff
@@ -1090,3 +1090,13 @@ InstallMethod( ViewObj,
     [ IS_OBJECT ],
     0,
     PRINT_OBJ );
+
+BIND_GLOBAL( "CHECK_REPEATED_ATTRIBUTE_SET", function(obj, name, val)
+    if CurrentAssertionLevel >= 2 then
+        if not IsBound(obj!.(name)) then
+            ErrorNoReturn("Attribute ", name, " of ", obj, " is marked as assigned, but it has no value");
+        elif obj!.(name) <> val then
+            ErrorNoReturn(name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```
fingolfin: I wouldn't print the old and new values: doing so can lead to errors itself; and it can also simply flood the screen, or cause all kinds of other problems. Inspecting these values if desired is what we have the break loop for. I'd instead give the relevant values suggestive names and indicate them in the error message, e.g. like so:
```suggestion
            ErrorNoReturn("Attribute ", name, " of <obj> already set to <oldval>, cannot be changed to <val>);
```
Then of course a local variable `oldval` should be added and set to `obj!.(name)`. 

(I'd also add the `"Attribute ", ` lead to the message, to match the other message)

src/opers.c
ChrisJefferson: Nope, they are handled in a different function (but I'll add a test, just to demonstrate this)

ChrisJefferson:
This certainly shouldn't go in 4.11.

I'm currently further investigating when this is being hit -- in some cases (like ParentAttr), we explictly say in the documentation it can be set multiple times, and late calls will be ignored, and that then happens in several places.

I'm thinking this might have to be weakened to an Info level, at least initially.

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
 ##
 ##  <Description>
 ##  This info class (together with <Ref InfoClass="InfoWarning"/> is used
```
fingolfin: Optional suggestion:
```suggestion
##  This info class (together with <Ref InfoClass="InfoWarning"/>) is used
```
Missing closing parenthesis (not you fault, of course, but in case other changes need to be made to this PR, this could also be addressed.

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
...
-##  for messages about attribute storing  being  disabled  (at  level  2)  or
-##  enabled (level 3). It may be  used  in  the  future  for  other  messages
+##  for messages about attributes. The current warnings are:
+##  <List>
+##    <Item> Attribute storing  being  disabled  (level  2) </Item>
```
fingolfin: I don't understand this. What is being warned about and when? I don't have the code at hand right now; perhaps the warning is issued if one uses an attribute setter on an object which does not store attributes resp. for which attribute storing is disabled? Or perhaps this is for attributes for which attribute storing is disabled (does such a thing exist? I seem to vaguely recall it, but can't right now think of an application)

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
...
+##    <Item> Atribute storing being enabled (level 3)</Item>
```
fingolfin: Again, I don't understand. When is this shown? Perhaps by attribute setters but when the previous message is not shown?!

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
...
+##    <Item> Trying to assign a non-mutable attribute a different value (level 3)</Item>
```
fingolfin: At least in German, an enumeration should either consist of complete sentences (starting in upper case, ending with a period/question mark/exclamation mark), or an enumeration as part of one big sentence (then starting in lower case, and usually ending with a comma, semicolon, or possibly a conjoining word like "and", "or). In the latter case it should also fit together with the start of the sentence before the list. Not sure how much of that applies in English, but here's a suggestion for this line (I'd also make suggestions for the first two lines, but I don't understand them at all, so I cannot do that :/).

So if we were to start the sentence with "The current warnings are shown under the following circumstances:" as I suggest above, we could have this list item read like this:

```suggestion
##    <Item>when trying to assign to non-mutable attribute which already is set to a different value (level 3);</Item>
```

Of course we could also do it completely differently. It's also not really the style or grammar I am concerned about, but rather comprehensibility of the text.

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
 ##
 ##  <Description>
 ##  This info class (together with <Ref InfoClass="InfoWarning"/> is used
-##  for messages about attribute storing  being  disabled  (at  level  2)  or
-##  enabled (level 3). It may be  used  in  the  future  for  other  messages
+##  for messages about attributes. The current warnings are:
```
fingolfin:
```suggestion
##  for messages about attributes. The current warnings are shown under the following circumstances:
```

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
...
+##  <List>
+##    <Item> Attribute storing  being  disabled  (level  2) </Item>
+##    <Item> Atribute storing being enabled (level 3)</Item>
+##    <Item> Trying to assign a non-mutable attribute a different value (level 3)</Item>
+##    <Item> An attribute which is marked as assigned, but has no value (level 3)</Item>
```
fingolfin:
```suggestion
##    <Item>when the test filter for an attribute (i.e., <C>HasFOO</C>) is set, but no value is assigned (level 3)</Item>
```

lib/attr.gd
```diff
@@ -37,8 +37,14 @@
...
+##  </List>
+##  It may be  used  in  the  future  for  other  messages
 ##  concerning changes to attribute behaviour.
```
fingolfin: That last sentence seems redundant to me. Of course we may add new Info messages in the future, that's true for every info class, isn't it? 

So, optional change: perhaps remove it?

lib/attr.gi
```diff
@@ -37,3 +37,11 @@ InstallGlobalFunction(DisableAttributeValueStoring, function( attr )
     Info(InfoAttributes + InfoWarning, 2, "Disabling value storing for ",NAME_FUNC(attr));
     SET_ATTRIBUTE_STORING( attr, false);
 end);
+
+CHECK_REPEATED_ATTRIBUTE_SET := function(obj, name, val)
+    if not IsBound(obj!.(name)) then
+        Info(InfoAttributes + InfoWarning, 3, "Attribute ", name, " of ", obj, " is marked as assigned, but it has no value");
+    elif obj!.(name) <> val then
```
fingolfin: Note that checking for equality can potentially be very expensive or even run into an infinite loop. So we should be careful to *only* check this if we are going to print a message anyway!

I.e., this function should probably at the very start check if the info level is right, and if not, return immediately.

lib/oper1.g
```diff
@@ -1090,3 +1090,7 @@ InstallMethod( ViewObj,
     [ IS_OBJECT ],
     0,
     PRINT_OBJ );
+
+# An dummy version of this function is installed here, the full version
```
fingolfin:
```suggestion
# A dummy version of this function is installed here, the full version
```

lib/oper1.g
```diff
@@ -1090,3 +1090,7 @@ InstallMethod( ViewObj,
...
+# is installed in attr.gd, after Info-related functionality is set up.
```
fingolfin:
```suggestion
# is defined in attr.gi, after Info-related functionality is set up.
```

src/opers.c
```diff
@@ -3283,12 +3284,15 @@ static Obj DoSetterFunction(Obj self, Obj obj, Obj value)
     flag2  = INT_INTOBJ( FLAG2_FILT(tester) );
     type   = TYPE_OBJ_FEO(obj);
     flags  = FLAGS_TYPE(type);
+
+    UInt rnam = (UInt)INT_INTOBJ(ELM_PLIST(tmp,1));
```
fingolfin:
```suggestion
    UInt rnam = (UInt)INT_INTOBJ(ELM_PLIST(tmp, 1));
```

lib/attr.gi
```diff
@@ -37,3 +37,11 @@ InstallGlobalFunction(DisableAttributeValueStoring, function( attr )
...
+        Info(InfoAttributes + InfoWarning, 3, name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```
fingolfin: I am somewhat wary about printing `obj` as well as the old/new value here. This can lead to pages of output, or trigger errors of its own, or even recursively run into this very same info message.

Which in turn makes me wonder if this should be an info message at all; perhaps it should be a call to `Error` instead (only called if the assertion level is higher than some threshold, and/or controlled by info levels as it is right now). Because thinking about how I'd deal with my code triggering these warning, the first thing I'd do is to edit this function to replace `Info` by `Error`, so that I get a backtrace to find out where the issue actually happened; in contrast, in my experience the printed object or value usually won't be helpful for me.

I guess in the end, it depends a bit on "where do we want to go with this"? Is our goal to stop people from setting non-mutable attributes to new, different values? Then wouldn't an Error (only at high enough level) be the way to go?
This reminds me of my idea of adding a "strict" mode in which all kinds of checks are stricter (see <https://github.com/gap-system/gap/issues/1191>)...


Anyway: please don't get me wrong, I don't want to hold back this PR over some bike shedding, and in the end, I am fine with this PR as it is as well (well, with the performance concern above addressed); I just wonder what the rationale for Info messages vs. Errors is, as naively, I'd expect the Error to be more useful -- but unlike you, I have not yet actually tried this PR, so my naive thoughts here might be totally offbase!

fingolfin:
Oh, and one more thing: I believe that in HPC-GAP, we kind of rely on being allowed to set an attribute twice, in case two threads race for computing the value of an attribute. Of course that should mean that the threads end up setting the attribute to "equal" values; but that also means that for HPC-GAP we'd have to be careful about turning these new checks on by default, as we'd then add those (possibly expensive) equality tests. Well, unless we can come up with a way to distinguish these "N threads raced for a computation" cases from others?

lib/attr.gd
ChrisJefferson: So, this, along with quite a few of your other comments, are text which already existed here. I'm happy to change it of course.

lib/attr.gi
ChrisJefferson: Yes, this is a serious problem, which I will fix.

lib/attr.gi
ChrisJefferson: I originally used Error in this PR, but there are lots of places where we do try to change attributes to different values. There are places (Like ?SetParent) where we document that trying to assign a different value will be ignored, and it is done in a bunch of places, so we can't just blindly change it. Therefore, it can't really be put at any (reasonable) assertion level, as it is valid and (reasonably) common behaviour to do these reassignments.

However, there are cases where looking at this uncovers bugs (I already reported one, which was fixed), and we could consider moving towards forbidding / reducing how many occur eventually.

I'd also prefer an error, but I'd have to make up a new kind of thing (like a Strict mode as you mentioned), so I thought an Info was better than nothing, and at least means we know in future where we have to "hook" to get more useful information about.

lib/attr.gd
fingolfin: Yeah, that's why I said this optional -- I am not expecting you to fix text that was already broken, although of course it is much appreciated if you do :-)

lib/attr.gi
fingolfin: Of course we could also just check for identity, using `IsIdenticalObj`, but then there might be more false warnings... hm

lib/attr.gi
fingolfin: BTW, I actually thought that we do document that calling an attribute setter again always has no effect (which of course is contradicted by your leading example). I can't find this in the manual now, though, but I thought that e.g. HPC-GAP relies on this principle in some parts -- but then perhaps this is simply different in HPC-GAP?

Hmmm.. but actually, I cannot reproduce your example at all. I get this:
```
gap> g := Group(());
Group(())
gap> SetTransitivity(g, 4);
gap> Transitivity(g);
4
gap> SetTransitivity(g, 3);
gap> Transitivity(g);
4
```
which is in line with what I thought was happening.

Now I wonder why you see something different?

lib/attr.gd
```diff
@@ -36,10 +36,14 @@
...
 ##  <Description>
-##  This info class (together with <Ref InfoClass="InfoWarning"/> is used
-##  for messages about attribute storing  being  disabled  (at  level  2)  or
-##  enabled (level 3). It may be  used  in  the  future  for  other  messages
-##  concerning changes to attribute behaviour.
+##  This info class (together with <Ref InfoClass="InfoWarning"/>) is used
+##  for messages about attributes. Messages are shown under the following circumstances:
+##  <List>
+##    <Item> <Ref Func="EnableAttributeValueStoring"/> is used (level 2).</Item>
+##    <Item> <Ref Func="DisableAttributeValueStoring"/> is used (level 3).</Item>
```
fingolfin: Ahhhh! That's much more helpful, thanks!

lib/attr.gi
```diff
@@ -37,3 +37,13 @@ InstallGlobalFunction(DisableAttributeValueStoring, function( attr )
     Info(InfoAttributes + InfoWarning, 2, "Disabling value storing for ",NAME_FUNC(attr));
     SET_ATTRIBUTE_STORING( attr, false);
 end);
+
+CHECK_REPEATED_ATTRIBUTE_SET := function(obj, name, val)
+    if InfoLevel(InfoAttributes + InfoWarning) >= 3 then
+        if not IsBound(obj!.(name)) then
+            Info(InfoAttributes + InfoWarning, 3, "Attribute ", name, " of ", obj, " is marked as assigned, but it has no value");
+        elif obj!.(name) <> val then
+            Info(InfoAttributes + InfoWarning, 3, name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```
fingolfin: Perhaps align this message to the one two lines above, as follows?
```suggestion
            Info(InfoAttributes + InfoWarning, 3, "Attribute ", name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```

lib/rwsgrp.gi
```diff
@@ -335,41 +335,7 @@ end );
...
-
-    # check the true methods
-    if HasIsFinite( rws ) then
-      SetIsFinite( grp, IsFinite( rws ) );
-    fi;
-    if IsPolycyclicCollector( rws ) then
-      defpcgs:=DefiningPcgs( ElementsFamily(FamilyObj(grp)) );
-      SetFamilyPcgs( grp, defpcgs);
-      SetHomePcgs( grp, defpcgs);
+#M  GroupByRwsNC( InfSetHomePcgs( grp, defpcgs);
```
fingolfin: Oops?

lib/attr.gi
```diff
@@ -37,3 +37,13 @@ InstallGlobalFunction(DisableAttributeValueStoring, function( attr )
     Info(InfoAttributes + InfoWarning, 2, "Disabling value storing for ",NAME_FUNC(attr));
     SET_ATTRIBUTE_STORING( attr, false);
 end);
+
+CHECK_REPEATED_ATTRIBUTE_SET := function(obj, name, val)
+    if InfoLevel(InfoAttributes) >= 3 then
+        if not IsBound(obj!.(name)) then
+            Info(InfoAttributes, 3, "Attribute ", name, " of ", obj, " is marked as assigned, but it has no value");
+        elif obj!.(name) <> val then
+            Info(InfoAttributes, 3, name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```
fingolfin: Perhaps align this message to the one two lines above, as follows?
```suggestion
            Info(InfoAttributes, 3, "Attribute ", name, " of ", obj, " already set to ", obj!.(name), ", cannot be changed to ", val);
```

lib/attr.gi
fingolfin: @ChrisJefferson can you explain the discrepancy between the leading example in this PR versus what I observe instead on master?

ChrisJefferson: The example at the top was a terrible typo, which went against the whole point :)

There are some setters which document you can reassign (SetParent). Some setters have an explicit check already to see if the value is changed and error if that happen (SetSize). Also properties won't let you set them the other way (but you can assign the same value multiple times):

```
gap> g := Group(());                                                  
Group(())
gap> SetIsAbelian(g, true);
gap> SetIsAbelian(g, true);
gap> SetIsAbelian(g, false);
Error, property is already set the other way
not in any function at *stdin*:8
```

ChrisJefferson:
Just to say (here at the bottom), my personal prefered long-term goal for attributes is:

1) Users can always assign an attribute the same value multiple times
2) When the AssertionLevel is < 2, assigning an attribute a different value is silently ignored (as present)
3) When the AssertionLevel is >= 2, then assigning an attribute a different value causes an Error (this would be new).

But, it turns out we are further from (3) than I thought, and furthermore even document the current behaviour in some places.

lib/attr.gi
fingolfin: Note that `Parent` is not an attribute, so `SetParent` is not a good example.

fingolfin:
I think investigating these is worthwhile. I believe I spotted already at least one actual bug using this PR. I'll report tomorrow, after some sleep :)

lib/coll.gi
```diff
@@ -3105,13 +3105,7 @@ InstallOtherMethod(SetSize,true,[IsObject and IsAttributeStoringRep,IsObject],
...
 local filt;
   if HasSize(obj) and Size(obj)<>sz then
-    if AssertionLevel()>2 then
-      # Make this an ordinary error (not ErrorNoReturn as suggested) to
-      # preserve all debugging options -- even use `return` to investigate
-      # what would have happened before this methods was introduced.
-      Error("size of ",obj," already set to ",Size(obj),
-        ", cannot be changed to ",sz);
-    fi;
+    CHECK_REPEATED_ATTRIBUTE_SET(obj, "Size", sz);
```
fingolfin: Perhaps a similar change should be applied to `SetDimension` in `lib/module.gi:413`?

tst/testinstall/attribute.tst
```diff
@@ -70,5 +76,13 @@ gap> FavouriteFruit(foo);
 "pear"
 gap> HasFavouriteFruit(foo);
 true
+gap> SetFavouriteFruit(foo, "apple");
+#I  Attribute FavouriteFruit of <object> already set to pear, cannot be changed to apple
```
fingolfin: should there perhaps be a similar test for *mutable* attributes which verifies that no warning is printed, and that the value is changed successfully?

lib/coll.gi
ChrisJefferson: Well spotted!

PR #3630, opened 2019-08-29
ENHANCE: Faster MaximalAbelianQuotient for subgroups of fp groups

hulpke:
# Description

Use abelianized rewriting rather than go though IsomorphismFpGroup.
Also added heuristic relator matrix reduction (not full HNF)
to use before SNF to speed up.

Also use in case that Abelian Rrs (in the kernel) fails.

This also includes a cleanup for #3634 
## Text for release notes 

The performance of AbelianInvariants and MaximalAbelianQuotient for subgroups of fp groups has been improved.

### DIFF
--- a/lib/matint.gi
+++ b/lib/matint.gi
@@ -1000,3 +1000,133 @@ InstallOtherMethod( AbelianInvariantsOfList,
     [ IsList and IsEmpty ],
     list -> [] );
 
+
+# Reduce a list of abelianized relations: Heuristic reduction without
+# making big vectors, iterate three times. Does not aim to do full HNF
+InstallGlobalFunction(ReducedRelationMat,function(mat)
+local n,zero,nv,new,pip,piv,i,v,p,w,g,nov,pin,now,rat,extra,clean,assign,try;
+
+  nv:=v->v*SignInt(v[PositionNonZero(v)]);
+  assign:=function(p,v)
+  local a,i,w,wn;
+    a:=v[p];
+    for i in [1..Length(pip)] do
+      if i<>p and IsInt(pip[i]) and mat[pip[i]][p]<>0 then
+        w:=mat[pip[i]]-QuoInt(mat[pip[i]][p],a)*v;
+        wn:=w*w;
+        if wn<=rat*pin[i] then
+          mat[pip[i]]:=nv(w);
+          pin[i]:=wn;
+        fi;
+      fi;
+    od;
+    mat[pip[p]]:=v;
+    # also try to reduce extra vectors
+    for i in [1..Length(extra)] do
+      w:=extra[i];
+      if not IsZero(extra[i]) then
+        wn:=w*w;
+        w:=w-QuoInt(w[p],a)*v;
+        if w*w<=rat*wn then
+          extra[i]:=w;
+        fi;
+      fi;
+    od;
+  end;
+
+  n:=NrCols(mat);
+  rat:=2; # growth ratio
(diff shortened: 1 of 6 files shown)


lib/matint.gi
```diff
@@ -1000,3 +1000,135 @@ InstallOtherMethod( AbelianInvariantsOfList,
...
+        fi;
+      fi;
+    od;
+  end;
+
+  n:=Length(mat[1]);
+  rat:=2; # growth ratio
+  zero:=ListWithIdenticalEntries(n,0);
+  mat:=Filtered(mat,x->not IsZero(x));
+  new:=Set(List(mat,nv)); # kill duplicates
```
fingolfin:
```suggestion
  new:=Set(mat,nv); # kill duplicates
```

lib/matint.gi
```diff
@@ -1000,3 +1000,135 @@ InstallOtherMethod( AbelianInvariantsOfList,
...
+        wn:=w*w;
+        w:=w-QuoInt(w[p],a)*v;
+        if w*w<=rat*wn then
+          extra[i]:=w;
+        fi;
+      fi;
+    od;
+  end;
+
+  n:=Length(mat[1]);
```
fingolfin:
```suggestion
  n:=NrCols(mat);
```

lib/matint.gi
```diff
@@ -1000,3 +1000,135 @@ InstallOtherMethod( AbelianInvariantsOfList,
...
+  pin:=[];
+  mat:=[];
+  extra:=[];
+
+  # we once reduce and then go over the remainders again in case they were
+  # nice and short
+  for try in [1..3] do
+    w:=List(new,x->x*x); # norm sort
+    SortParallel(w,new);
+    new:=Reversed(new);
```
fingolfin: It seems `w` is not used again later? If so, the above three lines could be replaced by just this:
```suggestion
    SortBy(new, x -> - x*x);
```

lib/matint.gi
```diff
@@ -1000,3 +1000,135 @@ InstallOtherMethod( AbelianInvariantsOfList,
...
+      fi;
+    od;
+    # now we've processed all. Clean out extra
+    new:=List(Filtered(Set(extra),x->not IsZero(x)),nv);
+    #Print("\n\nAfter ",try,": ",Length(extra)," to ",Length(new)," new ones\n");
+    extra:=[];
+
+  od;
+
+  return Filtered((Concatenation(mat,new)),
```
fingolfin:
```suggestion
  return Filtered(Concatenation(mat,new),x->not IsZero(x));
```

lib/matint.gi
```diff
@@ -1000,3 +1000,135 @@ InstallOtherMethod( AbelianInvariantsOfList,
...
+    x->not IsZero(x));
```
fingolfin:
```suggestion
```

lib/matint.gi
hulpke: I wasn't sure whether `SortBy' calculates the values once, or each time there is a comparison. If they are computed only once, that is of course fine.

lib/matint.gi
hulpke: Thanks -- relic of older.

lib/matint.gi
fingolfin: It does (which the documentation also guarantees)

hulpke: Thanks!

PR #3632, opened 2019-08-30
Cleanup and refactor in error.g

fingolfin:
Thank You!

# Description

## Text for release notes 

## Further details

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/lib/error.g
+++ b/lib/error.g
@@ -27,11 +27,10 @@ ERROR_OUTPUT := "*errout*";
 
 #############################################################################
 ##
-#F  OnQuit( )                                   currently removes all options
+#F  OnQuit( )
 ##
-Unbind(OnQuit);         # OnQuit is called from the kernel so we take great
-BIND_GLOBAL( "OnQuit",  # care to ensure it always has a definition. - GG
-        function()
+Unbind(OnQuit);
+BIND_GLOBAL( "OnQuit", function()
     if not IsEmpty(OptionsStack) then
       repeat
         PopOptions();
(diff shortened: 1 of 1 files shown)


ssiccha:
I don't have time to check this but might this break anything that hooks into the undocumented parts of the error handling? I think the JupyterKernel package should still work. What about sage etc?

ChrisJefferson:
I suspect this test failing is for exactly the reason you mention. Also (although this type of thing has been said before), might be worth waiting for 4.11 to branch, if this is going to mess up sage and friends.

fingolfin:
We don't test against SageMath, so that can't be the cause of any test failures. Nor do we run any package tests...

But in any case: SageMath does not reference `ErrorInner` (and even if it did, there would be no reason to expect why it would require a version of ErrorInner at the level of GAP 4.4?). Anyway, I also did check all distributed packages for using `ErrorInner`, and they are safe. 

All in all, I very much doubt that any code exists which is broken by this PR: In the first commit we change `ErrorInner` from variadic to two arguments. But all that does is change an explicit error "ErrorInner: new format takes exactly two arguments" into a generic error which you get when calling a function with the wrong number of arguments. The second commit removes `WHERE_INTERNAL`, but that did not exist in GAP 4.10 either. The last commit only contains formatting changes and removes some commented out code.

What happened is that the `debug-var.g` test in `testspecial` crashed, because I forgot a `return` statement in the second commit (ironically I remember explicitly thinking that I'd have to add that `return;` but then managed to forget it anyway :/). Should be fixed now.

PR #3641, opened 2019-09-02
re-added a method to compute `IsDiagonalMat` for `[]`

### LINKED ISSUE #3640: IsDiagonalMat for empty list / break cryst package tests
The `cryst` test suite recently started to fail, see e.g. [here](https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master-staging/jobs/579543535). Drilling down leads to this test in particular:
```
gap> List( List( [1..17], i -> SpaceGroupIT( 2, i ) ),
>   g -> Length( SpaceGroupsByPointGroup( PointGroup( g ) ) ) );
Error, no 1st choice method found for `IsDiagonalMatrix' on 1 arguments at GAPROOT/lib/methsel2.g:249 called from
IsDiagonalMat( M ) at GAPROOT/pkg/cryst/gap/zass.gi:151 called from
SolveHomEquationsModZ( mat ) at GAPROOT/pkg/cryst/gap/zass.gi:306 called from
ZassFunc( grp, [  ], false, CrystGroupDefaultAction = LeftAction ) at GAPROOT/pkg/cryst/gap/zass.gi:379 called from
SpaceGroupsByPointGroup( PointGroup( g ) ) at *stdin*:4 called from
func( C[i] ) at GAPROOT/lib/coll.gi:665 called from
...  at *stdin*:4
brk> M;
[  ]
brk>
```

So clearly, `IsDiagonalMat` is being called on an empty list. Up to GAP 4.10, this returned `true`.

It would be good to have this fixed before GAP 4.11 is released.

Ping @ThomasBreuer

ThomasBreuer:
I had replaced the method in question by an implication because GAP had told me that would be appropriate. However, this is not true in this situation because plain lists do not benefit from implications.
Thus I have now re-added the old method.

(This is intended to fix #3640.)

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -181,6 +181,24 @@ InstallTrueMethod( IsDiagonalMatrix, IsMatrixObj and IsEmptyMatrix );
 
 #############################################################################
 ##
+##  Note that an empty list is not a matrix,
+##  but 'IsDiagonalMat(rix)' used to return 'true' for it.
+##
+##  Note also that there was no such hack for other operations such as
+##  'IsUpperTriangularMat(rix)' or 'IsLowerTriangularMat(rix)'.
+##
+##  And note that installing an implication from 'IsList and IsEmpty' to
+##  'IsDiagonalMatrix' does not work because the type of an empty plain list
+##  cannot store the information.
+##  Furthermore, we cannot simply install 'ReturnTrue' as a method because
+##  then the method installation would complain that one should just install
+##  an implication.
+##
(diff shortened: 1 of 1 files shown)


wilfwilson:
Please could you add a test somewhere that `IsDiagonalMatrix([]) = true`.

ThomasBreuer:
@wilfwilson The aim of this pull request is to keep the behaviour for GAP 4.11 that people apparently have (mis)used up to now.
I hope this behaviour is not documented anywhere, and adding a test would make it "official".
Instead, I prefer to eventually get rid of it.  Note that the empty list `[]` claims to be a diagonal matrix but not a matrix.

PR #3649, opened 2019-09-06
kernel: turn some macros into static inline functions

### LINKED ISSUE #3648: Compilation issue on older Linux version
I am trying to compile the development version on a machine which (for reasons of "robustness" is using somewhat older versions of compiler etc, I have no way to override that policy. However I think this is not a completely untypical setup) The compiler is

     gcc version 4.4.7 20120313 (Red Hat 4.4.7-23) (GCC)

and I had to install autoconf on my own, as the version provided is below 2.68

When trying to compile, compilation fails with

    cc1: warning: unrecognized command line option "-Wno-implicit-fallthrough"
    cc1: warning: unrecognized command line option "-Wno-cast-function-type"
       CXX     src/objfgelm.cc => obj/src/objfgelm.lo
    src/objfgelm.cc:1332: error: insufficient contextual information to determine type

(also on following lines) and 

    src/objfgelm.cc:1384: error: too many initializers for ‘StructGVarFunc’

I'm reporting it as issue, as it might affect compilability of 4.11 (4.10 compiled on the machine) with the new C++ files

fingolfin:
Resolves #3648 (well, at least it ought to, but I can't test it, so let's wait to see what @hulpke reports)

### DIFF
--- a/src/objfgelm.cc
+++ b/src/objfgelm.cc
@@ -130,11 +130,8 @@ Obj NewWord(Obj type, UInt npairs)
 **
 *F  FuncNBits_Equal( <self>, <l>, <r> )
 */
-#define Func8Bits_Equal FuncNBits_Equal<UInt1>
-#define Func16Bits_Equal FuncNBits_Equal<UInt2>
-#define Func32Bits_Equal FuncNBits_Equal<UInt4>
 template <typename UIntN>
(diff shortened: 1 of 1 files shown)


fingolfin:
Good. And the rest?

hulpke:
The rest is that compilation works, but linking failed. (I put it into another issue, as it is a different problem, see #3651 )

Thanks again for the quick fix of this issue!

PR #3652, opened 2019-09-08
buildsys: use the C++ compiler for linking

### LINKED ISSUE #3651: Compilation/linking issue
[sorry for filling the issue space:]
still teh same machine as #3648. Thanks to #3649 compilation works, but linking gave an error

       LINK    gap
    obj/src/.libs/collectors.o:(.data.DW.ref.__gxx_personality_v0[DW.ref.__gxx_personality_v0]+0x0): undefined reference to `__gxx_personality_v0'

Searching found a recommendation:

If I change `GNUmakefile` in the line

    LIBS = -lm -ldl -lutil -lstdc++

(added the `-lstdc++` bit) then compilation succeeds. Alas I don't understand whether:

a) I can do so without harm for other systems

b) How to do it actually (as the line is auto-created). 

c) Is it just an artifect of me compiling a new autoconf and not compiling a new gcc?

fingolfin:
This ensures any required C++ runtime dependencies are added. While this
doesn't seem to matter for current versions of clang and gcc (as we
don't actually use anything from the C++ standard library), it does
matter for older GCC versions, e.g. GCC 4.4 seems to need it.

Hopefully fixes #3651

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,6 +18,7 @@ environment:
       CYG_ROOT: C:\cygwin64
       PACKAGES: "-P libgmp-devel,zlib-devel,python27,python27-pip"
       CFLAGS: "--coverage"
+      CXXFLAGS: "--coverage"
       LDFLAGS: "--coverage"
 # FIXME: HPC-GAP build on AppVeyor disabled for now, as it crashes
 #    - CYG_ARCH: x86_64
--- a/.travis.yml
+++ b/.travis.yml
@@ -140,6 +140,7 @@ matrix:
 
     # test Julia integration
     - env: TEST_SUITES="testinstall" JULIA=yes CONFIGFLAGS="--disable-Werror"
+      dist: bionic # need GCC 6 to avoid linker error
 
 script:
   - set -e
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -358,7 +358,7 @@ obj/%.lo: %.c cnf/GAP-CFLAGS cnf/GAP-CPPFLAGS libtool
 # Linker rules for gap executable
 ########################################################################
 
-LINK=$(LIBTOOL) --mode=link $(CC) -export-dynamic
+LINK=$(LIBTOOL) --mode=link $(CXX) -export-dynamic
 
 libgap.la: $(OBJS) cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-OBJS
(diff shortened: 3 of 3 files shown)


fingolfin:
This breaks the cygwin build, because we use an alternate linker command:
```
	$(QUIET_LINK)$(LINK) $(GAP_LDFLAGS) $(srcdir)/src/gapw95.c $(GAP_LIBS) libgap.la -o $@
```
and this now causes the C file gapw95.c to be compiled as C++, which introduced unwanted name mangling. I have several fixes in mind, let's see if I get one to work.

fingolfin:
Unfortunately, this change also breaks 32bit builds on Travis / Linux :-(.

ChrisJefferson:
I get the same problem on my machine, if I configure with ABI=32 (I have to type `make`, not just `make gap`, as it is building libgap which fails.

Googling around, other people seem to have problems with libtool picking up the wrong libraries when using -m32. Not sure why behaviour is changing with c++ however...

ChrisJefferson:
I think it's failing because libtool passes -nostdlib to the C++ compiler, then tries (and fails) to find the standard library itself.

ChrisJefferson:
I'm not saying this is right, but this seems to fix ABI=32:

```
 libgap.la: $(OBJS) cnf/GAP-LDFLAGS cnf/GAP-LIBS cnf/GAP-OBJS
-       $(QUIET_LINK)$(LINK) -no-undefined -rpath $(libdir) $(GAP_LDFLAGS) $(OBJS) $(GAP_LIBS) -o $@
+       $(QUIET_LINK)$(LINK) -no-undefined  $(GAP_LDFLAGS) $(OBJS) $(GAP_LIBS) -o $@
```

Seeing as, at present, we aren't installing GAP, then I think removing rpath shouldn't be a problem (and maybe it needs setting to something else on multilib?)

fingolfin:
I have a solution already but am busy bringing the kid to bed

fingolfin:
@ChrisJefferson I think I figured out the last remaining issue on Travis, in the Julia tests (needed a newer GCC for stupid reasons -- thank you StackOverflow and Google for helping me solve this); the other was on AppVeyor in 64 bit mode, and was because we didn't set `CXXFLAGS=--coverage`, and that ended up somehow producing a linker error in libgcov or so... Whatever, the change is obviously correct, mirrors what we do in Travis, and solved the issue, so I am happy :-). Now just the remaining Travis tests need to pass again.

fingolfin:
Backported to stable-4.11 in commit 1b6ea5ce4aedf1f1f52c983a1538cad295b68c9f

PR #3653, opened 2019-09-09
Typo: 'This files' -> 'This file'

fingolfin:
This is a trivial change, but that typo was really annoying and once you see it, you can't unsee it ;-).


ChrisJefferson:
I'm happy with this. I wonder if it should be merged into 4.11, just because it touches so many files (for example, it changes every generated file), so could create some annoying merges in future (not a requirement, just a thought).

fingolfin:
agreed, backporting seems sensible

PR #3654, opened 2019-09-11
Fix BuildPackages.sh

rbehrends:
Add the "-A" option to the gap invocation for determining the package
name. Otherwise, output from automatically loaded packages can create
problems.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -168,7 +168,7 @@ run_configure_and_make() {
   then
     if grep Autoconf ./configure > /dev/null
     then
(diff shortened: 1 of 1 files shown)


rbehrends:
This showed up on OSCAR CI because the IO package wasn't built properly. Some other package printed a "\r" character during loading, which confused the build script.

fingolfin:
Browse prints the \r, that's unfortunately a known issue in there

fingolfin:
Backported in commit 6a68156d1cb0bd91ab99b89aae2e3a6b9c023683

PR #3655, opened 2019-09-11
Avoid setting attributes twice

fingolfin:
This fixes a few places which cause Info messages when using PR #3628 and a high enough InfoAttributes level. In the case of GeneralRestrictedMapping, it avoids a warning for an attribute where HasFOO is set (here: HasSource and HasRange), but no actual value is stored in the object: for mappings, we do this intentionally, as the Source/Range are stored in the family object.

This touches Pcgs and mappings code. Many further things could be fixed, but these seem fairly "obvious" to me.

### DIFF
--- a/lib/mapprep.gi
+++ b/lib/mapprep.gi
@@ -1694,29 +1694,26 @@ InstallMethod( IsSurjective,
 
 #############################################################################
 ##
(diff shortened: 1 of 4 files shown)


ChrisJefferson:
Thanks for looking at these, at least one is a clear bug, and the others are definately improvements.

PR #3656, opened 2019-09-11
Fix NanosecondsSinceEpoch to use monotonic clock on linux

markuspf:
Including `time.h` in `src/system.c` makes CLOCK_MONOTONIC available, thus
activating the correct code for `NanosecondsSinceEpoch`.

I don't know how long this was broken for.


fingolfin:
Backported to stable-4.11 in commit 063d30124571d940742a0964bbcec04a2d0790ba

PR #3657, opened 2019-09-12
buildsys: change how ABI_CFLAGS are handled; export GAP_CXX in sysinfo.gap

fingolfin:
This is a subset of PR #3652, which I split out for two reasons:
1. Running this on Travis helps me narrow down the remaining issues with PR #3652 (kind of like bisecting the issue...)
2. I think the changes in here are independently useful, and so it's useful to merge this ASAP, so it gets more testing, while completing PR #3652 might take longer (or I might even have to go with a completely different approach for it in the end).

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -721,6 +721,7 @@ sysinfo.gap: config.status $(srcdir)/Makefile.rules cnf/GAP-CFLAGS cnf/GAP-CPPFL
 	@echo "GAP_LIB_DIR=\"$(abs_srcdir)\"" >> $@
 	@echo "" >> $@
 	@echo "GAP_CC=\"$(CC)\"" >> $@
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Just checking @fingolfin , are you now happy for this to be merged? I am, and the tests pass, just wanted to sanity check.

fingolfin:
Yes, merged it. There was an issue with HPC-GAP in 32 bit failing, but that actually seems to be independent of this PR. I still would have fixed it, but could not reproduce it locally.

fingolfin:
Backported to stable-4.11 in commits f1ce52fbb8f94baa8531c813debdebbb87ff40c0 and 8cee4d21e2b0447ab8c1de2c830cc2124ec74d4d

PR #3659, opened 2019-09-16
Fix random fluctuations in direct_factors.tst; simplify UnionIfCanEasilySortElements

### LINKED ISSUE #3658: Unstable DirectFactorsOfGroup test
This is happening in the master branch when no packages are loaded (see https://travis-ci.org/gap-infra/gap-docker-master-testsuite/jobs/585051188). The output needs to be sorted to enforce the ordering.
```
########> Diff in /home/gap/inst/gap-master/tst/teststandard/direct_factors.ts\
t:45
# Input is:
List(DirectFactorsOfGroup(G), IdGroup);
# Expected output:
[ [ 6, 1 ], [ 6, 1 ], [ 6, 1 ], [ 2, 1 ], [ 2, 1 ], [ 2, 1 ] ]
# But found:
[ [ 2, 1 ], [ 6, 1 ], [ 2, 1 ], [ 6, 1 ], [ 2, 1 ], [ 6, 1 ] ]
########
########> Diff in /home/gap/inst/gap-master/tst/teststandard/direct_factors.ts\
t:48
# Input is:
List(DirectFactorsOfGroup(G), IdGroup);
# Expected output:
[ [ 6, 1 ], [ 6, 1 ], [ 2, 1 ], [ 2, 1 ] ]
# But found:
[ [ 2, 1 ], [ 6, 1 ], [ 2, 1 ], [ 6, 1 ] ]
########
```

fingolfin:
Fixes #3658

Do you want this to be backported?

### DIFF
--- a/lib/grpnames.gi
+++ b/lib/grpnames.gi
@@ -81,16 +81,12 @@ InstallMethod( IsTrivialNormalIntersection,
 InstallGlobalFunction( UnionIfCanEasilySortElements,
 
   function( arg )
-    local i, N;
 
-    for i in [1..Length(arg)] do
-      for N in arg[i] do
-        if not CanEasilySortElements(N) then
-          return Concatenation(arg);
-        fi;
-      od;
-    od;
-    return Union(arg);
+    if ForAll(arg, CanEasilySortElements) then
+      return Union(arg);
+    else
+      return Concatenation(arg);
+    fi;
   end);
 
 #############################################################################
--- a/tst/teststandard/direct_factors.tst
+++ b/tst/teststandard/direct_factors.tst
@@ -42,11 +42,11 @@ true
 gap> G:=SmallGroup(16,7);; C := Center(G);; NormalComplement(G,C);
 fail
 gap> G := DirectProduct(D, D, D);;
-gap> List(DirectFactorsOfGroup(G), IdGroup);
-[ [ 6, 1 ], [ 6, 1 ], [ 6, 1 ], [ 2, 1 ], [ 2, 1 ], [ 2, 1 ] ]
+gap> SortedList(List(DirectFactorsOfGroup(G), IdGroup));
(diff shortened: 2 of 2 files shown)


lib/grpnames.gi
```diff
@@ -81,16 +81,12 @@ InstallMethod( IsTrivialNormalIntersection,
...
 
-    for i in [1..Length(arg)] do
-      for N in arg[i] do
-        if not CanEasilySortElements(N) then
-          return Concatenation(arg);
-        fi;
-      od;
-    od;
-    return Union(arg);
+    if ForAll(arg, CanEasilySortElements) then
```
wilfwilson: The code as previously written was literally:
```suggestion
    if ForAll(arg, N -> ForAll(N, CanEasilySortElements)) then
```
Is what you've written equivalent to this? Just wanted to check. (Sorry - I'm ignorant of the details of `CanEasilySortElements`).

fingolfin: That's a very sensible question, of course! So, there is this method:
```gap
InstallMethod(CanEasilySortElements,"homogeneous list",
  [IsHomogeneousList],
function(l)
  return Length(l)=0 or CanEasilySortElements(l[1]);
end);
```
So my version should be fine, as long as all list over which we take the union are homogenous, which they are in the usescases we apply it. *But* thinking about this again, of course `UnionIfCanEasilySortElements`, while undocumented, could in principle be called of any kind of input, so perhaps we should be defensive here?

OTOH, the worst that can go wrong is that we take a concatenation when we could have perform a union. Also, `AddSetIfCanEasilySortElements(list, obj)` already seems to rely on this behavior, as it calls `CanEasilySortElements(list)` (and also not `CanEasilySortElements(obj)`)

fingolfin:
Backported the test fix (but not the refactoring) to stable-4.11 in commit fbd979b27a167e114661baebdf016a0405bc0d9f

PR #3661, opened 2019-09-17
Kernel cleanup related to associative words / pc words

fingolfin:


lib/rwspcgrp.gi
```diff
@@ -400,7 +416,7 @@ function( sc )
     fam!.8BitsType := NewType( fam, Is8BitsPcWordRep );
 
     # copy the assoc word type
-    for i  in [ AWP_FIRST_ENTRY .. AWP_FIRST_FREE-1 ]  do
+    for i in [ AWP_FIRST_ENTRY+1 .. AWP_LAST_ENTRY ] do
         fam!.8BitsType![i] := sc![SCP_DEFAULT_TYPE]![i];
```
fingolfin: Note that AWP_FIRST_ENTRY+1 = AWP_PURE_TYPE, and that entry is set directly after this loop. So the insertion of `+1` just ensures that we don't set that entry twice, and syncs us with the corresponding HPC-GAP code.

hpcgap/lib/rwspcgrp.gi
```diff
@@ -400,7 +416,7 @@ function( sc )
     fam!.8BitsType := NewType( fam, Is8BitsPcWordRep );
 
     # copy the assoc word type
-    for i  in [ AWP_FIRST_ENTRY+1 .. AWP_FIRST_FREE-1 ]  do
```
hulpke: Is this fixing a bug, or did the size of the data type change?

fingolfin: I think our comments overlapped in time :-); here is what I wrote above:

Note that AWP_FIRST_ENTRY+1 = AWP_PURE_TYPE, and that entry is set directly after this loop. So the insertion of +1 just ensures that we don't set that entry twice, and syncs us with the corresponding HPC-GAP code.

hulpke: Thanks!

PR #3662, opened 2019-09-17
FIX: Bug in MaximalSubgroupClassReps

hulpke:
# Description

The code used to compute a special pcgs in a pc group image and transfer
some cached information -- that does not seem to work properly, and also not
be worth the effort of doing. Replace by direct computation.

This fixes a bug reported by Andries Brouwer (example is in test file)

## Text for release notes 

A bug in `MaximalSubgroupClassReps` that could lead to a wrong result was fixed.


fingolfin:
Backported to stable-4.11 in commit a0b9798aac226d603a2618ab02d039a1fc175027

PR #3664, opened 2019-09-18
Tweak Rees matrix semigroups code

fingolfin:
This removes the Matrix attribute; instead, methods for the Matrix *operation*
are installed, which delegate suitably to MatrixOfReesMatrixSemigroup resp.
MatrixOfZeroReesMatrixSemigroup, which now are independent attributes (instead
of both being synonyms for Matrix).

Also change Rees matrix semigroup code which access the attributes Rows and
Columns to instead access RowsOfReesZeroMatrixSemigroup, etc., explicitly, to
mirror the changes for Matrix, and to prepare for the eventuality that we
might want to convert Rows and Columns into plain operations, too.

Finally, tweak some manual entries for Rees matrix semigroups: mansections
referencing themselves should still use `<Ref>` for that.

The motivation for this is that we may want to allow using single-arg `Matrix` as a convenient way to construct MatrixObj instances; say `Matrix(old_style_mat)` could take a classic list-of-lists, and return an equivalent MatrixObj. That said, we might not actually do this; but simply the fact that Matrix is already an operation for 2, 3, ... args means that it is somewhat awkward to have it also be an attribute. So I'd rather avoid that.

Perhaps we should do the same for Rows and Columns, just in case we want to provide similar operations for MatrixObj? Not sure about that, though. And even if we over such APIs, we could call them, say, RowsOfMatrix; and then, perhaps we'd want an iterator anyway, so it'd be RowIterator/ColumnIterator; so I am not sure it's worth the hassle to even think about this more at this stage? The main motivation to change them, too, at this stage, would be symmetry; i.e., trying to not surprise people using Rees matrix semigroups by odd differences between Matrix on the one hand, and Rows/Columns on the other.

This PR should NOT be merged before Semigroups merges https://github.com/gap-packages/Semigroups/pull/618 and then has a release.


doc/ref/reesmat.xml
```diff
@@ -125,14 +125,14 @@ gap> S:=ReesZeroMatrixSemigroup(DihedralGroup(IsPermGroup, 8), mat);
     <Oper Name="ReesZeroMatrixSubsemigroup" Arg="R, I, U, J"/>
     <Returns>A Rees matrix or 0-matrix subsemigroup.</Returns>
     <Description>
-      The arguments of <C>ReesMatrixSubsemigroup</C> should be a Rees matrix
+      The arguments of <Ref Oper="ReesMatrixSubsemigroup"/> should be a Rees matrix
       semigroup <A>R</A>, subsets <A>I</A> and <A>J</A> of the rows and columns
       of <A>R</A>, respectively, and a subsemigroup <A>S</A> of the underlying
```
ThomasBreuer: `S` does not occur among the arguments.

doc/ref/reesmat.xml
```diff
@@ -360,16 +360,20 @@ false]]></Example>
 </ManSection>
 
   <ManSection>
-    <Attr Name="Matrix" Arg="R"/>
+    <Oper Name="Matrix" Arg="R"/>
```
ThomasBreuer: As soon as the operation `Matrix` becomes documented in the MatrixObj context,
we will need labels to distinguish the different declarations.
We can add the `Label="for Rees matrix semigroups"` attribute already now,
and use it in the `<Ref>` elements below.

lib/reesmat.gd
```diff
@@ -81,10 +81,8 @@ DeclareOperation( "ReesZeroMatrixSubsemigroup",
   [IsReesZeroMatrixSubsemigroup, IsList, IsSemigroup, IsList]);
 DeclareGlobalFunction("ReesZeroMatrixSubsemigroupNC");
 
-DeclareAttribute("Matrix", IsReesMatrixSubsemigroup);
-DeclareAttribute("Matrix", IsReesZeroMatrixSubsemigroup);
-DeclareSynonymAttr("MatrixOfReesMatrixSemigroup", Matrix);
-DeclareSynonymAttr("MatrixOfReesZeroMatrixSemigroup", Matrix);
+DeclareAttribute("MatrixOfReesMatrixSemigroup", IsReesMatrixSubsemigroup);
+DeclareAttribute("MatrixOfReesZeroMatrixSemigroup", IsReesZeroMatrixSubsemigroup);
```
ThomasBreuer: Up to now, there was just one attribute `Matrix` with two synonyms `MatrixOfReesMatrixSemigroup`, `MatrixOfReesZeroMatrixSemigroup`.
Do we really need/want two independent attributes `MatrixOfReesMatrixSemigroup`, `MatrixOfReesZeroMatrixSemigroup`?

wilfwilson: What would be an alternative way of doing this, so that there were not two independent attributes? We could have one attribute called something like `MatrixOfReesOrReesZeroMatrixSemigroup` with the other names being synonyms for it, but that feels a bit weird to me.

wilfwilson:
Thanks for pointing out the mutability difference. I would be surprised if the immutable result causes anything to break, but we'll see.

lib/reesmat.gd
ThomasBreuer: Since there was just one attribute with different names `MatrixOfReesMatrixSemigroup`, `MatrixOfReesZeroMatrixSemigroup` up to now, the situation seems to be that either no object lies in both `IsReesMatrixSubsemigroup` and `IsReesZeroMatrixSubsemigroup` or the attribute values for objects in both filters are equal.
In the former case, introducing two independent attributes is on the one hand safer (one gets an error if one calls the wrong attribute), but on the other hand `Matrix` methods for the two cases make the distinction disappear, and from this point of view just one filter is wasted.
In the latter case, introducing two independent attributes may cause that the same value has to be computed twice, if one asks for the values of both attributes.

wilfwilson: The former case is the true one: `IsReesMatrixSemigroupElement` and `IsReesZeroMatrixSemigroupElement` (from which `IsReesMatrixSubsemigroup` and `IsReesZeroMatrixSubsemigroup` are derived) are different categories, and so a GAP object cannot lie in both.

Is it a problem to 'waste' filters?

fingolfin:
> Besides the specific comments in the code, I would like to emphasize that after the change, `Matrix` for Rees matrix semigroups will return an _immutable_ result, whereas the result has been _mutable_ up to now (due to a declaration of `Matrix` as a mutable attribute, which is read before the declaration in `reesmat.gd`).

Good point, I actually forgot about that! I'll update this PR to also remove that "mutable attribute Matrix".

doc/ref/reesmat.xml
fingolfin: Indeed, I believe `U` was meant. Will fix

doc/ref/reesmat.xml
fingolfin: Thanks, this will be fixed in the next update of this PR.

lib/reesmat.gd
fingolfin: I asked @wilfwilson beforehand to make sure that the filters are mutually exclusive, and hence split this into two attributes. The alternative would IMHO not be to introduce `MatrixOfReesOrReesZeroMatrixSemigroup`, but rather I'd just turn one `DeclareAttribute` back into `DeclareSynonymAttr`.

But I really think it's best the way I now did it. Yes, we loose one filter bit. I don't see an issue with that, we have plenty bits left ;-).

lib/reesmat.gi
```diff
@@ -557,27 +558,31 @@ end);
...
 [IsReesZeroMatrixSubsemigroup and HasGeneratorsOfSemigroup], 
 function(R)
   if not IsReesZeroMatrixSemigroup(R) then 
     return fail;
   fi;
-  return Matrix(ParentAttr(R));
+  return MatrixOfReesZeroMatrixSemigroup(ParentAttr(R));
 end);
 
+# for backwards compatibility
```
fingolfin: perhaps this should say "for convenience and backwards compatibility"?

ThomasBreuer: Yes, since `Matrix` is documented for these situations.

james-d-mitchell:
Thanks for these changes and comments all, I will double check that everything still works in Semigroups today or tomorrow, then I’ll happy “approve”.

fingolfin:
@james-d-mitchell thank you! In any case, we won't merge this before a new Semigroups release with the changes has been made, and picked up into the packages distribution.

wilfwilson:
@james-d-mitchell made a new release (thanks!), hopefully it will be picked up soon by the package distribution system.

fingolfin:
Excellent. So once @alex-konovalov resp. the package distribution picked up Semigroups 3.1.5, we could merge this, I guess. Well, of course only after rebasing it and rerunning the tests to make sure they then pass again.

fingolfin:
Tests still fail, because the YangBaxter package does this: `DeclareAttribute("Matrix", IsCycleSet);`

wilfwilson:
Which, for ease of keeping track, @fingolfin has made a PR to address at https://github.com/gap-packages/YangBaxter/pull/64.

wilfwilson:
Seeing as YangBaxter 0.9.0 has been released, I’m hoping that the tests will now pass.

wilfwilson:
@fingolfin Could you rebase this please? Hopefully that will make the tests pass, and then we can merge 🙂

Issue #3663, opened 2019-09-17
Problem with several root directories containing the same version of a package


ThomasBreuer:
@mohamed-barakat had observed the following problem with GAP's package loading mechanism (see https://github.com/oscar-system/GAP.jl/issues/287).
When a GAP package can be found in several root directories, with the same version number, then GAP ignores all except the first instance.
It may happen that the chosen instance cannot be loaded because a compiled module is missing or has been compiled with configuration parameters that do not fit to the GAP in question (with or without Julia, say), whereas another instance would be loadable.

One could argue that GAP should load the first instance that fits, and then the current behaviour could be called a bug.

No matter whether we regard this as a bug or not, I think it would be not difficult to deal with the situation sketched above.
For that,
- GAP must not refuse different package installations with the same version number; this concerns the function `AddPackageInfos`,
- functions such as `DirectoriesPackagePrograms` and `DirectoriesPackageLibrary` must use the installation path instead of the version number as their search criterion (which is anyhow an improvement),
- `InitializePackageInfoRecords` must sort the info records stably, in order to achieve that the loadable instance gets chosen in the first possible root directory.

Do you think this is sensible?

fingolfin:
Seem fine to me.

mohamed-barakat:
For me too.

ThomasBreuer:
O.k., I will create a pull request for the proposed change.

DominikBernhardt:
@ThomasBreuer For some reason this wasn't closed by merging #3668 . Should this remain open?

ThomasBreuer:
@DominikBernhardt Thanks for spotting this. No, the issue should be closed.

mohamed-barakat:
Thank you very much Thomas 👍

Issue #2568, opened 2018-06-20
Immutable GAPInfo.PackagesInfo records: Problems for `JuliaInterface` and `float`


fingolfin:
When I merged `hpcgap/lib/package.gi` into `lib/package.gi`, see 42789e61c61440f24663e08731b5351ef1ae2b6b, I accidentally broke the `float` package, which modifies its package record after it has been set. It does that to modify its banner string, to indicate which parts of it where actually compiled and loaded. That is indeed quite useful information, and it makes sense to display that as part of the banner string.

One way to fix this is to stop making the package records immutable, at least in GAP. But they'll stay immutable in HPC-GAP, so `float` would be broken there, unless we can stop  making the package record immutable there, too... or perhaps we can delay doing so until the package was fully loaded (i.e. both init.g and read.g have been read)? 

So my preferred solution would be to work together with @laurentbartholdi to change float to *not* modify the package record during a late stage anymore. But right now, I have not yet made my mind up on what I think would be a good solution; also, perhaps changing the immutability of the package info record at a later stage can be implemented, then no change to float would be necessary.

This must be resolved before GAP 4.10.

alex-konovalov:
Why can't we make them atomic records in HPC-GAP instead of being immutable?

fingolfin:
Well, then the content of the records (i.e. strings, lists, other records) still would not be accessible from all threads. Of course that may not be a major problem in the end -- I did not extensively test this, I just copied changes that others applied there in the past (e.g. @ChrisJefferson it seems, based on my reading of `git log`, which may flawed, though).

laurentbartholdi:
Well, I get punished for hacking, and that's only fair.

What if we rather change the format of PackageInfo, and say that
BannerString can be either a string or a function with 0 arguments and
returning a string? That would avoid all problems, be clean and
backwards-compatible. It's a single-line change to the package loading code.

On Jun 20, 2018 16:25, "Alexander Konovalov" <notifications@github.com>
wrote:

If you do atomic list of immutable things, then Float can simply replace
one immutable banner by another immutable banner.

Or, maybe Float should simply change the way it modifies its record? Make a
copy, modify it, make it immutable and replace the old record. Needs some
experimenting.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/gap-system/gap/issues/2568#issuecomment-398768741>, or mute
the thread

fingolfin:
[...]

PR #3668, opened 2019-09-19
support several package instances with same version

ThomasBreuer:
- `AddPackageInfos` does no longer ignore package installations for which one instance with the same version number has already been found.

- `InitializePackagesInfoRecords` sorts the info records stably, in order to achieve that the loadable instance gets chosen in the first possible root directory.

- `DirectoriesPackagePrograms` and `DirectoriesPackageLibrary` now search according to the installation path, not to the version number.

This pull request is intended to fix issue #3663.

*Two related questions:*
- The manual example for `DirectoriesPackagePrograms` shows an output of length larger than 1.
  ```
    gap> DirectoriesPackagePrograms( "nq" );
    [ dir("/home/gap/4.0/pkg/nq/bin/x86_64-unknown-linux-gnu-gcc/64-bit/"),
      dir("/home/gap/4.0/pkg/nq/bin/x86_64-unknown-linux-gnu-gcc/") ]
  ```
  Has this ever been possible?
- The code of `AddPackageInfos` contains a `FIXME` comment that mentions the (closed) issue #2568. Should we remove the `FIXME`, or should `MakeImmutable` perhaps be called in general, not just in the `IsHPCGAP` situation? 

## Text for release notes 
If several instances of a package, with the same version number, are available in the root directories, and if the first such instance cannot be loaded (because a compiled module is missing or does not fit) then the other instances are considered for being loaded.
In earlier versions, only the first instance of a given package version was considered.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -1189,26 +1189,26 @@ InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
 #F  DirectoriesPackagePrograms( <name> )
 ##
 InstallGlobalFunction( DirectoriesPackagePrograms, function( name )
-    local info, ppath;
+    local info, installationpath;
 
     # We are not allowed to call
     # `InstalledPackageVersion', `TestPackageAvailability' etc.
     info:= PackageInfo( name );
     if IsBound( GAPInfo.PackagesLoaded.( name ) ) then
       # The package is already loaded.
-      ppath:= GAPInfo.PackagesLoaded.( name )[1];
+      installationpath:= GAPInfo.PackagesLoaded.( name )[1];
     elif IsBound( GAPInfo.PackageCurrent ) then
       # The package is currently going to be loaded.
-      ppath:= GAPInfo.PackageCurrent.InstallationPath;
+      installationpath:= GAPInfo.PackageCurrent.InstallationPath;
     elif 0 < Length( info ) then
       # Take the installed package with the highest version
       # that has been found first in the root paths.
-      ppath:= info[1].InstallationPath;
+      installationpath:= info[1].InstallationPath;
     else
       # This package is not known.
       return [];
     fi;
-    return [ Directory( Concatenation( ppath, "/bin/",
+    return [ Directory( Concatenation( installationpath, "/bin/",
                             GAPInfo.Architecture, "/" ) ) ];
 end );
 
@@ -1218,7 +1218,7 @@ end );
 #F  DirectoriesPackageLibrary( <name>[, <path>] )
 ##
 InstallGlobalFunction( DirectoriesPackageLibrary, function( arg )
-    local name, path, info, ppath, tmp;
+    local name, path, info, installationpath, tmp;
 
     if IsEmpty(arg) or 2 < Length(arg) then
         Error( "usage: DirectoriesPackageLibrary( <name>[, <path>] )" );
@@ -1240,19 +1240,19 @@ InstallGlobalFunction( DirectoriesPackageLibrary, function( arg )
     info:= PackageInfo( name );
     if IsBound( GAPInfo.PackagesLoaded.( name ) ) then
       # The package is already loaded.
-      ppath:= GAPInfo.PackagesLoaded.( name )[1];
+      installationpath:= GAPInfo.PackagesLoaded.( name )[1];
     elif IsBound( GAPInfo.PackageCurrent ) then
       # The package is currently going to be loaded.
-      ppath:= GAPInfo.PackageCurrent.InstallationPath;
+      installationpath:= GAPInfo.PackageCurrent.InstallationPath;
     elif 0 < Length( info ) then
       # Take the installed package with the highest version
       # that has been found first in the root paths.
-      ppath:= info[1].InstallationPath;
+      installationpath:= info[1].InstallationPath;
     else
       # This package is not known.
       return [];
     fi;
-    tmp:= Concatenation( ppath, "/", path );
+    tmp:= Concatenation( installationpath, "/", path );
     if IsDirectoryPath( tmp ) = true then
       return [ Directory( tmp ) ];
     fi;
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -225,36 +225,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
         NormalizeWhitespace( pkgname );
         version:= record.Version;
 
-        # If we have this version already then leave it out.
-        if ForAll( GAPInfo.PackagesInfo,
-                    r ->    r.PackageName <> record.PackageName
-                        or r.Version <> version ) then
-
-          # Check whether GAP wants to reset loadability.
-          if     IsBound( GAPInfo.PackagesRestrictions.( pkgname ) )
-              and GAPInfo.PackagesRestrictions.( pkgname ).OnInitialization(
-                      record ) = false then
-            Add( GAPInfo.PackagesInfoRefuseLoad, record );
-          elif pkgname in ignore then
-            LogPackageLoadingMessage( PACKAGE_DEBUG,
-                Concatenation( "ignore package ", record.PackageName,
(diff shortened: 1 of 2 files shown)


lib/package.gi
```diff
@@ -225,36 +225,30 @@ BindGlobal( "AddPackageInfos", function( files, pkgdir, ignore )
         NormalizeWhitespace( pkgname );
         version:= record.Version;
 
-        # If we have this version already then leave it out.
-        if ForAll( GAPInfo.PackagesInfo,
-                    r ->    r.PackageName <> record.PackageName
-                        or r.Version <> version ) then
```
fingolfin: Personally, I prefer code which is not deeply nested, makes it easier for me to keep the state in mind. Here, that could be achieved by negating the condition and using `continue;` in the body of the `if`, no? Then the diff would also be minimized.

But that's in the end a matter of taste.

lib/package.gi
```diff
@@ -1190,30 +1189,27 @@ InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
...
 
-    arch := GAPInfo.Architecture;
-    dirs := [];
     # We are not allowed to call
     # `InstalledPackageVersion', `TestPackageAvailability' etc.
     info:= PackageInfo( name );
     if IsBound( GAPInfo.PackagesLoaded.( name ) ) then
       # The package is already loaded.
-      version:= GAPInfo.PackagesLoaded.( name )[2];
+      ppath:= GAPInfo.PackagesLoaded.( name )[1];
```
fingolfin: What does ppath stand for? "package path"? I first thought it was a typo. So, optional suggestion: revert to `path`, or use a more expressive name, like `pkg_path`?

lib/package.gi
```diff
@@ -1190,30 +1189,27 @@ InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
...
     elif IsBound( GAPInfo.PackageCurrent ) then
       # The package is currently going to be loaded.
-      version:= GAPInfo.PackageCurrent.Version;
+      ppath:= GAPInfo.PackageCurrent.InstallationPath;
     elif 0 < Length( info ) then
-      # Take the installed package with the highest version.
-      version:= info[1].Version;
+      # Take the installed package with the highest version
+      # that has been found first in the root paths.
+      ppath:= info[1].InstallationPath;
```
fingolfin: So, that's a semantical change now, isn't it? Before we took all packages with the same version, and for each took its path. Now, you only take the first one, and ignore the rest.

Is that intentional? Why? Also, is this covered by the PR description? Maybe it is, but I don't quite understand all in there, either. E.g. what does this mean:
> DirectoriesPackagePrograms and DirectoriesPackageLibrary now search according to the installation path, not to the version number

I think this confused me because "search" is ambiguous here. When reading that comment, I am thinking about loading packages, and then "search" immediately leads me to think about "searching the pkg dirs in the GAP roots for PackageInfo.g files".

Anyway, I really had to look at the code to understand what is going on, and even now, I am not sure I fully understand. I need to look at that code in a wider context, I think (i.e., open up the file, and read)

fingolfin:
Yes, in older GAP versions, two paths in `DirectoriesPackagePrograms` like in the example you quote were possible, because there was a second "architecture" `GAPInfo.ArchitectureBase`. Actually, that code was added by you 2011-04-21, and removed by me 2017-09-21 in commit c67e018ce745489f0b8b1552dae6597deb60ef7a -- clearly I overlooked those examples, which should be adjusted.

However, I actually always thought that the *intention* behind the code in `DirectoriesPackagePrograms` was all along to support multiple copies of the same version of a package, by simply returning a list containing all the `bin` dirs of each copy -- well, at least each that contains a matching `bin/ARCH` subdir...

lib/package.gi
```diff
@@ -1239,29 +1235,28 @@ InstallGlobalFunction( DirectoriesPackageLibrary, function( arg )
...
-    return dirs;
+      # Take the installed package with the highest version
+      # that has been found first in the root paths.
+      ppath:= info[1].InstallationPath;
+    else
+      # This package is not known.
+      return [];
+    fi;
+    tmp:= Concatenation( ppath, "/", path );
+    if IsDirectoryPath( tmp ) = true then
```
fingolfin: I notice that we check `IsDirectoryPath` here, but not in `DirectoriesPackagePrograms`... I wonder why?

Also, perhaps the two functions could share some code in a common helper (I am not meaning to request that you do this in this PR, just thinking out loud).

ThomasBreuer: Yes, `DirectoriesPackageLibrary` checks the path, and `DirectoriesPackagePrograms` doesn't.
Users may rely on this behaviour, and I had no reason to change this.

lib/package.gi
ThomasBreuer: I wanted the code of `DirectoriesPackageLibrary` and `DirectoriesPackagePrograms` to be similar, and `path` was already used in one of them.
I can choose another variable name, but I think this discussion is not useful.

lib/package.gi
ThomasBreuer: I do not understand this comment.
The point is that this `if` disappears completely.

ThomasBreuer:
@fingolfin Concerning the remark

> However, I actually always thought that the intention behind the code in DirectoriesPackagePrograms was all along to support multiple copies of the same version of a package, by simply returning a list containing all the bin dirs of each copy -- well, at least each that contains a matching bin/ARCH subdir...

Multiple instances of the same version of a package were not supported up to now,
and I think that it would be asking for trouble if GAP could collect pieces of data from different instances.

lib/package.gi
ThomasBreuer: I do not understand the comment.
The list `info` contains all info records for the package in question, sorted by decreasing version number (and in the new version, where several records for the same version can exist, also sorted according to the ordering of the root paths where they were found).
In the old version, we fetch the version number and later look up the (unique) info record with this version number in order to fetch its path.
In the new version, we fetch the path directly.
(Perhaps the reason for the confusion is the `for` loop in the old version, which did not find more than one entry to add because there was only one package for the given version.)

lib/package.gi
fingolfin: For some reason the comment ended up in the wrong line. I'll move it.

fingolfin: Ah no, in fact, it was indeed just nonsense, I had turned on "ignore whitespace" and then apparantly develop a temporary case of blindness / stupidity :-). Sorry for the confusion

lib/package.gi
fingolfin: Yes, the confusion is: *why* is there only one record for each given version? My (apparently wrong) impression was that the plan with this patch was to add *all* copies of a given package we find to our records, even those with identical version; but keep them sorted in the order we find them; and then when looking for binaries, we search all copies of the latest version. But it seems I misunderstood the intentions here?

And I still think that the phrase "search according to the installation path" is completely unclear. You now wrote "sorted according to the ordering of the root paths where they were found" which makes much more senes to me.

ThomasBreuer: @fingolfin Thanks for the explanations.
As I had written in the description, the idea is to admit instances of a package with the same version numbers, in the sense that the package is regarded as loadable if there is one instance which is loadable.
This means in particular that it is not supported to "distribute the parts of a package to different root directories", for example to have a valid package binary only in one directory, some GAP functions of the package only in a second directory, and some package data only in a third directory.

(Concerning the test failure due to decreased coverage:
The coverage decreases when one removes some useless but covered code lines.
And when one changes or moves an uncovered line then apparently this counts as the introduction of a new uncovered line.
Very helpful.)

fingolfin: The coverage failure does not block merging, only actual test suite failures do (this is configured in the "branch protection rules"). The PR status says "Merging is blocked" not because of the "failing" coverage, but because there is no approving review yet.

mohamed-barakat:
Thanks to both of you @ThomasBreuer @fingolfin.

PR #3669, opened 2019-09-19
julia: added a new way to load GAPTypes

sebasguts:
by getting module and object from a variable

# Description

This fixes the bug [here](https://github.com/homalg-project/HomalgProject.jl/issues/2) together with a patch for GAP.jl submitted [here](https://github.com/oscar-system/GAP.jl/pull/289)

Does not need to appear in release notes.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -728,61 +728,45 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
 
     // Import GAPTypes module to have access to GapObj abstract type.
     // Needs to be done before setting any GC states
-    UInt            gapjl_already_loaded = 1;
-    jl_module_t *   gapjl_module;
+    jl_module_t *   parent_module;
     jl_module_t *   gaptypes_module;
     jl_datatype_t * gapobj_type;
-    // Check if we can get the GAP.jl module
-    if (jl_boundp(jl_main_module, jl_symbol("__JULIAGAPMODULE"))) {
-        jl_value_t * gapjl_value =
-            jl_get_global(jl_main_module, jl_symbol("__JULIAGAPMODULE"));
-        if (!jl_is_module(gapjl_value)) {
+
+    parent_module = (jl_module_t *)jl_get_global(
+        jl_main_module, jl_symbol("__JULIAGAPMODULE"));
+    if (parent_module) {
+        if (!jl_is_module(parent_module)) {
             Panic("__JULIAGAPMODULE is set in julia main module, but does "
                   "not point to a module");
         }
-        gapjl_module = (jl_module_t *)gapjl_value;
     }
     else {
-        gapjl_already_loaded = 0;
-    }
-
-    if (gapjl_already_loaded) {
-        if (!jl_boundp(gapjl_module, jl_symbol("GAPTypes"))) {
-            Panic("GAPTypes is not available in already loaded GAP.jl Julia "
-                  "module");
-        }
-        jl_value_t * gaptypes_module_value =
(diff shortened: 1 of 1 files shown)


src/julia_gc.c
```diff
@@ -728,20 +728,36 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
...
+        go_to_gaptypes_fallback = 1;
     }
-    // Get GapObj abstract julia type
-    jl_module_t * gaptypes_module =
-        (jl_module_t *)jl_get_global(jl_main_module, jl_symbol("GAPTypes"));
-    if (jl_exception_occurred()) {
-        Panic("Could not read global GAPTypes variable from Julia");
+    else {
+        gaptypes_module = (jl_module_t *)jl_get_global(
+            jl_main_module, jl_symbol("__GapObj_MODULE_VAR"));
```
fingolfin: So, how about we instead get the GAP.jl module, from the already existing global var `__JULIAGAPMODULE`. Then, if it exists, it will also have the GAPTypes module as a member, and from that we can get GapObj, no?

src/julia_gc.c
```diff
@@ -728,20 +728,36 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
...
+        if (jl_exception_occurred()) {
+            go_to_gaptypes_fallback = 1;
```
fingolfin: Have you tested this? is there really an exception here. From the quick glance at the Julia code, my impression was that  `jl_get_global` returns NULL if the binding does not exist. Also, `jl_exception_occurred` w/o an explicit exception handler only works for a very, very small set of APIs explicitly designated for this purpose, and I am pretty sure `jl_get_global` is not one of them

src/julia_gc.c
fingolfin: Probably also wouldn't hurt to verify we are really getting a module, via `jl_is_module`

fingolfin:
Thanks for working on this.

Ideally, we should finish this ASAP and then backport it to 4.11

sebasguts:
I think I addressed your comments, and added a lot of checks. I will close the corresponding GAP.jl now.

src/julia_gc.c
```diff
@@ -728,20 +728,61 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
...
-    if (jl_exception_occurred()) {
-        Panic("could not import GAPTypes module into Julia");
+    UInt            gapjl_already_loaded = 1;
+    jl_module_t *   gapjl_module;
+    jl_module_t *   gaptypes_module;
+    jl_datatype_t * gapobj_type;
+    // Check if we can get the GAP.jl module
+    if (jl_boundp(jl_main_module, jl_symbol("__JULIAGAPMODULE"))) {
+        jl_value_t * gapjl_value =
+            jl_get_global(jl_main_module, jl_symbol("__JULIAGAPMODULE"));
```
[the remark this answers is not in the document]
fingolfin: Actually, I just verified it (by running the code): it returns NULL, and does *not* return an exception.

I will update this PR with an alternative implementation for further discussion.

fingolfin:
@sebasguts I refactored your code, please have a look. Also, perhaps @rbehrends could review this PR as well?

sebasguts:
I think this should be merged. However, since this is my PR, I cannot approve it.

fingolfin:
Backported to stable-4.11 in commits 4203986c0483705323524728b586a06370775aac and 5a72f726369bfa5d02d322cd8b8376ab3b0fa2da

PR #3671, opened 2019-09-20
Travis: switch to Julia 1.2

fingolfin:
This only affects the Julia job on Travis, nothing else (and it should not really have any effect there, either, just wanted to make sure we test against the version of Julia that is actually relevant for our GAP-Julia integration these days).


fingolfin:
Backported to stable-4.11 in commit 4203986c0483705323524728b586a06370775aac

PR #3673, opened 2019-09-20
Use dynamic rank for ViewObj/PrintObj on identity/zero maps

fingolfin:
This fixes a problem in the RCWA package test suite, where the following
happens: the package installed a custom PrintObj method for its own type
IsRcwaMappingOfZ, with SUM_FLAGS as rank adjustment. But loading certain other
packages causes the rank of `IsGeneralMapping and IsOne` to grow a lot, and
grow above that of IsRcwaMappingOfZ. This causes the PrintObj method from the
GAP library to be selected unexpectedly.

This should fix the RCWA tests on https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master-staging

### DIFF
--- a/lib/mapprep.gi
+++ b/lib/mapprep.gi
@@ -1343,7 +1343,11 @@ InstallMethod( PreImagesRepresentative,
 InstallMethod( ViewObj,
     "for identity mapping",
     true,
-    [ IsGeneralMapping and IsOne ], SUM_FLAGS,
+    [ IsGeneralMapping and IsOne ],
+    # rank up, but just to exactly SUM_FLAGS, so that mappings in a special
+    # representation with a custom printing method still get that, even if
+    # the rank of IsGeneralMapping and IsOne happens to be increased a lot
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.11 in commit 4bea1d8094ab870216803f9b5390d4c415e17a3e

alex-konovalov:
RCWA tests still failing but the reason seems to be different: https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master-staging/jobs/600105776
```
########> Diff in /home/gap/inst/gap-master/pkg/rcwa-4.6.4/tst/integral.tst:
1103
# Input is:
phi := IsomorphismPermGroup(G);
# Expected output:
[ <rcwa permutation of Z with modulus 2, of order 2>, 
  <rcwa permutation of Z with modulus 3, of order 3> ] -> 
[ (1,6)(2,3)(4,5), (1,5,6)(2,3,4) ]
# But found:
Error, no 1st choice method found for `RightCayleyDigraph' on 1 arguments
########
```
and then 5 more diffs likely caused by this.

PR #2758, opened 2018-08-28
 Revert "Downrank MatrixObj methods to unbreak Semigroups"

markuspf:
This reverts  "Downrank MatrixObj methods to unbreak Semigroups" again and serves as a reminder that we need to take care of this issue after 4.10 is released.

PR #3674, opened 2019-09-23
Revert some Semigroups-related downrankings of Matrix methods

wilfwilson:
Three downrankings of `Matrix` methods were added in advance of GAP 4.10, along with one for `BaseDomain`, in order to make sure that GAP 4.10 did not break the Semigroups package. This was supposed to be temporary, and #2758 exists to revert this change.

In fact, only one of these downrankings is required to keep Semigroups happy. This commit reverts the others. Once it is merged, I will update #2758 accordingly.

At some point (separately from this PR) I will try to see about removing the remaining downranking of a `Matrix` method.

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -184,9 +184,6 @@ InstallMethod( Matrix,
 #
 InstallMethod( Matrix,
   [IsSemiring, IsList, IsInt],
-  # FIXME: Remove this downranking, it was introduced to prevent
-  #        Semigroups from breaking ahead of the 4.10 release
-  -SUM_FLAGS,
   function( basedomain, list, nrCols )
     local rep;
     rep := DefaultMatrixRepForBaseDomain(basedomain);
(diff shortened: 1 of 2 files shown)


wilfwilson:
There is an additional FIXME that I didn't consider - I'll check it out shortly.

wilfwilson:
I've also reverted the `BaseDomain` downranking that also existed (thanks @PaulaHaehndel for pointing it out to me). Doing so doesn't seem to affect the Semigroups package, so I don't think there can be any harm in removing it.

@fingolfin As for backporting, I don't mind. These downrankings didn't seem to be having any consequences, so I don't know whether backporting would make any noticeable difference to anything. Please add the backport tag if you think it's appropriate.

wilfwilson:
Just to clarify, I ran the GAP `testinstall`, `teststandard`, `testbugfix` test files with this branch of GAP and with the latest released version of the Semigroups package loaded, and I also ran the Semigroups package tests in this version of GAP. Everything passed as normal.

PR #3678, opened 2019-09-27
kernel: fix bug when printing objects is interrupted

### LINKED ISSUE #3677: Bug in print/view
This bug requires some setup. Firstly do:

```
gap> x := []; for i in [1..10000] do x := [x]; od;
gap> x;
```

Then press ctrl+c to enter the break loop while `x` is printing, then type 'quit;' to exit the break loop.

Now, the `Print` function is broken, and will sometimes print `~`.

```
gap> l := [1,2,3];
[ 1, 2, 3 ]
gap> l;
[ 1, 2, 3 ]
gap> Print(l,"\n");
~
gap> l;
[ 1, 2, 3 ]
gap> Print(l,"\n");
~
gap> Print(l,"\n");
~
gap> 2;
2
gap> Print(l,"\n");
[ 1, 2, 3 ]
gap> Print(l,"\n");
~
```

This is an interaction of a few different things:

* The main bug. The variable `LastPV` is used to record if we are already "in" a `Print` or `View` statement. If we exit while inside a `Print` or `View` statement, this variable is never reset, and we forever think we are still "inside" the `View` which was printing `x`.

* We leave stale values in os->PrintObjThis and os->PrintObjIndex, which get reused when we think we are inside a `View` statement.

fingolfin:
... by a break loop, for deeply recursive objects. In that case, the global
kernel variable LastPV, which normally should be 0 when not printing, could be
left in a non-zero state. If it ever is left in a non-zero state (which can
happen if printing is interrupted at an inconvenient moment), then it is
virtually impossible to reset it to 0, leaving one in a broken state.

There was also another issue regarding to LastPV: as it was not thread local,
printing recursive objects in HPC-GAP on two different threads could have been
badly broken (didn't verify this experimentally, but seems plausible).

Fixes #3677

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -196,8 +196,7 @@ static Obj Shell(Obj    context,
       ErrorQuit("SHELL: can't open infile %s",(Int)inFile,0);
     }
   
-  oldPrintDepth = STATE(PrintObjDepth);
-  STATE(PrintObjDepth) = 0;
+  oldPrintDepth = SetPrintObjState(0);
 
   while ( 1 ) {
 
@@ -210,7 +209,7 @@ static Obj Shell(Obj    context,
     /* read and evaluate one command                                   */
     STATE(Prompt) = prompt;
     ClearError();
-    STATE(PrintObjDepth) = 0;
+    SetPrintObjState(0);
     ResetOutputIndent();
     SetRecursionDepth(0);
       
@@ -291,7 +290,7 @@ static Obj Shell(Obj    context,
 
   }
   
-  STATE(PrintObjDepth) = oldPrintDepth;
+  SetPrintObjState(oldPrintDepth);
   CloseInput();
   CloseOutput();
   STATE(BaseShellContext) = oldBaseShellContext;
--- a/src/objects.c
+++ b/src/objects.c
@@ -839,8 +839,8 @@ static inline UInt IS_ON_PRINT_STACK(const ObjectsModuleState * os, Obj obj)
   if (!(FIRST_RECORD_TNUM <= TNUM_OBJ(obj)
         && TNUM_OBJ(obj) <= LAST_LIST_TNUM))
     return 0;
-  for (i = 0; i < os->PrintObjDepth-1; i++)
-    if (os->PrintObjThiss[i] == obj)
+  for (i = 1; i < os->PrintObjDepth; i++)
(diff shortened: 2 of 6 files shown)


src/gap.c
```diff
@@ -291,7 +290,7 @@ static Obj Shell(Obj    context,
 
   }
   
-  STATE(PrintObjDepth) = oldPrintDepth;
+  SetPrintObjState(oldPrintDepth);
```
fingolfin: Now that I pushed this, I see that there is yet another bug lurking here: we reset the print stack *depth*, but not its contents. So that could potentially lead to issues, and even crashes, as the objects "on" that stack could be changed now.

To fix this, I think we'd have to save and restore the whole printobj stack with all its content. Ugh.

My next thought was, "OK, let's just set the depth to 0 here, then while resulting in a botched printing (the one we interrupted by a break loop), otherwise we should be safe". But that's of course not true, as the kernel C functions `ViewObj` and `PrintObj` unconditionally decrement `PrintObjDepth` after the printing. So we'd end up with an underflow. OK, so next we can of course insert a check for that: if `PrintObjDepth == 0` before the point were we decrement it, then immediately abort the printing.

That's probably the quickest way to "fix" that issue. For a proper fix, I think we should just move the printobj state into a GAP back (or a bunch of them.. perhaps a bunch of plists), then it'll be easy to store and reset them when entering / leaving `Shell`.

src/objects.c
```diff
@@ -829,7 +839,7 @@ static inline UInt IS_ON_PRINT_STACK(const ObjectsModuleState * os, Obj obj)
   if (!(FIRST_RECORD_TNUM <= TNUM_OBJ(obj)
```
ChrisJefferson: While here, maybe make the `UInt i` on line 828 be `Int i`, so we correctly handle `os->PrintObjDepth-1` being 0. This function shouldn't get called when the stack is empty, but that is one of the things that this bug caused.

fingolfin: I addressed this now, albeit in a slightly different way.

fingolfin:
Backported to stable-4.11 in commits b5b151afd744d4d2c98be6935ebffa22a65aaedc and cb9bee9035c29b3212aad86fdf1409926f1284e1

fingolfin:
I don't think any user will ever notice this very obscure fringe case, so no need to put this in the release notes.

PR #3679, opened 2019-09-27
kernel: refactor PrintObj/ViewObj

fingolfin:
This PR contains PR #3678. Best to view this commit by commit, I think.


src/objects.c
```diff
@@ -835,14 +835,14 @@ static Obj FuncMakeImmutable(Obj self, Obj obj)
...
 {
-  UInt i;
-  if (!(FIRST_RECORD_TNUM <= TNUM_OBJ(obj)
-        && TNUM_OBJ(obj) <= LAST_LIST_TNUM))
+    UInt i;
+    if (!(FIRST_RECORD_TNUM <= TNUM_OBJ(obj) &&
+          TNUM_OBJ(obj) <= LAST_LIST_TNUM))
+        return 0;
+    for (i = 0; i < os->PrintObjDepth; i++)
+        if (os->PrintObjThiss[i] == obj)
```
fingolfin: Note that thanks to the refactoring, we can rewrite this loop to be sane again :-). And the "fix" from 712a56f1138ad5a78db97d4f3a12dafe838267d8 becomes unnecessary.

ChrisJefferson:
I put some effort into entering and leaving break loops with this patch (except I had to push the limit back up, so I had time to press ctrl+c), and couldn't get anything bad to happen.

PR #3684, opened 2019-10-04
kernel: strict variants of GVAR_FUNC and GVAR_OPER

fingolfin:
This PR adds strict variants of the GVAR_FUNC and GVAR_OPER macros which verify that the given C function has the correct signature. The price to achieve that portably is that the new macros have the number of arguments as part of the macro name.

This uncovered (only :-) one function in the kernel which had a bad signature, which the first commit in this PR fixes.


src/modules.h
```diff
@@ -134,6 +134,62 @@ struct init_info {
...
+{
+    return 0;
+}
+static inline ObjFunc VALIDATE_FUNC_HELPER_6(ObjFunc_6ARGS f)
+{
+    return 0;
+}
+
+#define VALIDATE_FUNC_0ARGS(func)                                            \
+    (0 ? VALIDATE_FUNC_HELPER_0(func) : (ObjFunc)func)
```
ChrisJefferson: Not wanting to be too picky, would this be simpler with a simple identity function, which passed it's argument back? I'd expect it to get optimised away as well.

fingolfin: I tried this at first, but it doesn't work. Well, it *would* work in C++ code, and perhaps some C compilers would allow it as an extension, but in general, this results in a compiler error (in clang: `error: initializer element is not a compile-time constant`; in GCC: `error: initializer element is not constant`).

You can verify this with e.g. https://godbolt.org and this input, where the definition of `val` is accepted and that of `val2` produces a compiler error:
```C
static inline int func(int x) { return x; }

int val = 0 ? func(0) : 1;
int val2 = func(0);

int main(void)
{
    return val;
}
```
Perhaps I should explain that in a code comment? OTOH, if somebody tries to refactor the code, they'll quickly notice, I think...

Issue #3373, opened 2019-03-22
Change `PLAIN_LIST` (and similar in-place converters) to only work on mutable objects, reject immutable ones


fingolfin:
Motivation: all kinds of problems can crop up if e.g. an immutable string object is silently converted to a plain list. It also means that all places in the kernel that currently expose immutable string objects to the user are in danger, as user's could cause those strings to be turned into plists (e.g. as in issue #815), which then later leads to crashes when some other kernel code accesses those string, expecting them to be still string objects.

Therefore, I think `PLAIN_LIST` should *not* accepted immutable objects, but rather trigger an error for them. Of course it's not a as as that: we'd also have to audit all calls to `PLAIN_LIST` to ensure they only are performed for immutable inputs. Some approaches that can be used to this end:

* change the code calling `PLAIN_LIST` to switch from plist APIs like ELM_PLIST to ELM_LIST etc.;
* make two variants of the calling code, one using ELM_PLIST etc. and one using ELM_LIST etc. (using C++ templates, this could be achieved with little code duplication);
* instead of calling `PLAIN_LIST`, we could instead call a function which creates a new plist with the content of the old list (at least for immutable inputs)
* we could outright reject (immutable) non-plists, and get people to fix their code calling into the kernel (of course that's a backward breaking change, so one should be very careful about doing that).

All of this could of course be applied to other in-place conversion functions; to this end, it'd be interesting to collect an overview of them (in particular, I am not sure if any exist that do something to string objects?).

ChrisJefferson:
I think my preference would be a mixture of (1) and (2), in many cases I think `PLAIN_LIST` is an optimisation. If we could write a C++ template list wrapper, it would be easy to optimise many algorithms for plists vs normal lists.

fingolfin:
Related issue: #3155

fingolfin:
Also somewhat related: #815

stevelinton:
This is about the definition of immutability. At the moment immutable objects are explicitly allowed to change representation, add stored attributes, etc. as long as they do not cease to be equal to their old values.  So at the moment, kernel code that assumes that this will not happen is actually buggy.

We could change our mind on this and (for instance) forbid immutable objects from changing representation in-place. That would forbid switching permutations from perm2 to perm4 format, for instance. We could just impose that rule on lists. 

Alternatively, we could just try to reduce the number occasions on which these changes happen, but then the kernel code is still buggy, but harder to detect.

[...]

Issue #3155, opened 2018-12-30
BlistList (and UniteBlistList and perhaps more) docs claim it supports arbitrary list inputs, but it doesn't


fingolfin:
This comes from issue #3148, but is a separate (bigger) problem, as @ThomasBreuer pointed out on PR #3149: The documentation for `BlistList` says:
> returns  a  new  boolean  list  that  describes  the  list  sub as a sublist of the dense list list

This suggest that any dense list may be passed in as first argument. But this is not true in practice:
```
gap> vec:=ImmutableVector(GF(2),[Z(2),Z(2)]);;
gap> BlistList(vec, [Z(2)]);
Error, Panic: cannot convert <list> (is a object (data)) to a plain list
not in any function at *stdin*:8
```
The underlying problem is that internally, `BlistList(list, sub)` (resp. `BLIST_LIST`) calls `UniteBlistList(list, new_all_false_blist, sub)` (resp. `UNITE_BLIST_LIST`), which ultimately ends up calling `PLAIN_LIST( list )` in one way or another -- but that fails (which is a good thing).

To me, the real culprit here is `PLAIN_LIST( list )`, which is "neat" to simplify a lot of kernel code (which now only has to implement a plist version), but the silent conversions it performs lead to all kinds of unexpected nasty surprises (and are a big no-no for HPC-GAP, too). So, my ideal fix for this issue would be to get rid of all uses of `PLAIN_LIST` in the kernel, and thus of `PLAIN_LIST` itself. Alas, that may not be feasible, as some code may implicitly rely on the implicit conversions?

Another fix would be to explicitly reject non-internal lists in `BlistList` and friends (and adjust the documentation to say that), but of course that'd be sad.

In this particular case, a minor optimization would help, in where we detect that the final argument is an empty list, and then don't bother to do anything... But that of course only fixes on minor variant of the real problem.

ThomasBreuer:
@fingolfin Yes, converting a given list into a plain list is a bad idea.

A natural alternative to `PLAIN_LIST` would be a function that takes a list
and returns a plain list with its entries and its holes at the same positions as in the given list.
In all situations where the list to work with needs not be identical with the given one
(as in the `BlistList` case), we would get the abovementioned simplifications
but the code would be safe.

Of course is may be inefficient to create a new plain list in situations as for `BlistList`;
here one solution could be to turn `BlistList` into an operation,
then a good method can be provided, depending on the types of the arguments.

fingolfin:
[...]

Issue #1095, opened 2017-01-19
UniteSet makes immutable arguments mutable, and removes string rep.


fingolfin:
A weird test failure on PR #1072 lead me to the following discovery:
```gap
gap> not IsMutable(CHARS_LALPHA); IsStringRep(CHARS_LALPHA);
true
true
gap> UniteSet("abc",CHARS_LALPHA);
gap> not IsMutable(CHARS_LALPHA); IsStringRep(CHARS_LALPHA);
false
false
```
So, the globale variable `CHARS_LALPHA` is immutable and in string rep before the call to `UniteSet`. Afterwards, it is neither.

fingolfin:
Wow, this must be an anicent bug (?!).

Turns out that the kernel function `IsSet` invokes `PLAIN_LIST` on anything that is not a plist but a small list (such as a string in stringrep...) and sorted.

And since `IsSet` is invoked on both arguments of `UniteSet`...

So to fix this, we could change `IsSet` to not modify its arguments. But that will have a ripple effect, we'd have to carefully inspect everything call it. For `FuncUNITE_SET`, it's not so bad because it already calls `SetList` on its second argument if it is not a set. But for the first argument, things are different.

I guess we might need two variants of `IsSet` (e.g. by giving it a second argument, a boolean which indicates whether it should attempt to convert its argument in-place or not. And even for the version that performs in-place conversion, we should consider honoring the immutable flag -- in this case, that would mean checking `IS_MUTABLE_OBJ` *before* `IsSet`

Moreover, we now may want to provide versions of `UniteSet` etc. for strings, and/or arbitrary small lists. The code would be almost identical, of course, just `ELM_PLIST` be replaced by `GET_ELM_STRING` or `ELM_LIST`. So we could use some preprocessor magic to reduce the required effort... (And if we were allowed to use C++, then some basic template programming would very easily take care of this, too).

Finally, if we fix this, it might affect some library code which implicitly expects that the arguments of `UniteSet` (and other set functions) are turned into plain lists.

fingolfin:
Seems this bug has been around since the old repository was started :)

fingolfin:
So `PlainString` was not correctly preserving mutability, that fixes half of the problem.

After having looked longer at the relevant code, I guess it is intentional that the representation of the string is changed. I still find it questionable to do this, but there is a lot of code in the kernel of that nature, so trying to change it might be quite monumental...

alex-konovalov:
[...]

Issue #815, opened 2016-06-02
`IsSubsetSet` (and others) change lists in-place


ChrisJefferson:
Various functions in GAP (including `IsSubset`) can change functions in place, making them not a string rep, and mutable. This is probably a bad idea, and will certainly be a bad idea in HPC-GAP. I think the best fix is probably to change `PLAIN_LIST` to do a copy instead of an in-place modify (or at least, not in-place modify immutable variables).

```
gap> g := Immutable("E");
"E"
gap> IsMutable(g);
false
gap> IsStringRep(g);
true
gap> IsSubset("A", g);
false
gap> IsStringRep(g);
false
gap> IsMutable(g);
true
```

alex-konovalov:
@ChrisJefferson The current behaviour is as follows:
```
gap> g := Immutable("E");
"E"
gap> IsMutable(g);
false
gap>  IsStringRep(g);
true
gap>  IsSubset("A", g);
false
gap> IsStringRep(g);
false
gap> IsMutable(g);
false
gap> 
```
so g is still immutable, but changes representation.

fingolfin:
Yes, this minor improvement is due to b6cc57f8546c6ab64690f9b44fcbbaa6e242dee5

fingolfin:
Here is another example, taken from #1095 and motivated by actual code that run into problems due to this bug:
```gap
gap> not IsMutable(CHARS_LALPHA); IsStringRep(CHARS_LALPHA);
true
true
gap> UniteSet("abc",CHARS_LALPHA);
gap> not IsMutable(CHARS_LALPHA); IsStringRep(CHARS_LALPHA);
true
false
```

alex-konovalov:
This also causes problems for FR, see https://github.com/gap-packages/fr/issues/19

fingolfin:
I fixed this not so long ago! Now, basically all places performing this inplace conversion via the kernel function `PLAIN_LIST` only do so for mutable lists, immutable lists effectively have "locked" representation...

... except for the kernel functions underlying `ConvertToVectorRep` and `ConvertToMatrixRep` which still call it on immutable objects -- and of course also the other way around, i.e. they convert stuff into compressed vectors/matrices even if it is immutable. This is a major design flaw of these functions, and one of the reasons I introduced `ImmutableVector` and improved `ImmutableMatrix`... I'd really like to purge / replaces all uses of these functions and related ones.

Anyway, I believe this issue can be closed now.

PR #3685, opened 2019-10-04
kernel: add PLAIN_LIST_COPY helper

fingolfin:
The function PLAIN_LIST converts any other type of internal lists (blists, strings, ranges) into plain lists. This is a powerful function that can lead to very surprising and problematic results (see e.g. issues #815, #1095; also relevant: #3155, #3373 , hence it would be good to minimize or even remove its usage.

It turns out that in quite some places, we first make a shallow copy of a list and then apply PLAIN_LIST to that. So here we introduce a new helper which combines the two tasks into one and avoids the in-place conversion altogether. This let's the remaining uses of PLAIN_LIST in the kernel stand out more.

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -269,8 +269,7 @@ Obj CALL_WITH_CATCH(Obj func, volatile Obj args)
         RequireArgument("CALL_WITH_CATCH", args, "must be a list");
 #ifdef HPCGAP
     if (!IS_PLIST(args)) {
-        args = SHALLOW_COPY_OBJ(args);
-        PLAIN_LIST(args);
+        args = PLAIN_LIST_COPY(args);
     }
 #endif
 
--- a/src/lists.c
+++ b/src/lists.c
@@ -1614,6 +1614,23 @@ static void PlainListError(Obj list)
         (Int)TNAM_OBJ(list), 0L );
 }
 
+Obj PLAIN_LIST_COPY(Obj list)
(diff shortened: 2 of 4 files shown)


ChrisJefferson:
I do however like any attempt to reduce the use of PLAIN_LIST

fingolfin:
There was a stupid off-by-one bug in `PLAIN_LIST_COPY` (I didn't copy the last element...) and another bug in `FuncOnPosIntSetsTrans` (it used `IS_PLIST_MUTABLE` on something which is not a plist, which I now changed to use `IS_MUTABLE_OBJ`). Other than that, though, `FuncOnPosIntSetsTrans` at least seems to pass tests. It certainly may crash for other inputs, but then a lot of functions in `trans.cc` and `pperm.cc` do that -- my impression was on purpose, i.e., they often do minimal input validation or only use assertion checks (i.e.: upon invalid input, crash in debug builds, and do weird things in regular builds).

fingolfin:
@ChrisJefferson I filed an issue https://github.com/gap-packages/Semigroups/issues/623 -- but is there anything else *concrete* that you'd like me to change in this PR right now?

ChrisJefferson:
Nope, I'm happy for this to be merged.

fingolfin:
@ChrisJefferson but merging is blocked because you requested changes... :-)

PR #3686, opened 2019-10-05
Add descriptive message to declared but not defined global functions

### LINKED ISSUE #2926: Unhelpful printing of declared but not installed functions
This just confused a user; it's probably "just" a bug in function viewing/printing.

```
gap> DeclareGlobalFunction("Cheese");
gap> Cheese;
function( args... ) ... end
gap> Print(Cheese);
function ( args... )
    <<kernel or compiled code>>
end
gap> Cheese();
Error, Cheese: function is not yet defined
not in any function at *stdin*:7
```

ChrisJefferson:
Fixes #2926

Basically adds a location to declared but not defined functions, to help users know what they have. This is a bit of an abuse, but anything else would require more major surgery I think.

### DIFF
--- a/src/opers.c
+++ b/src/opers.c
@@ -18,6 +18,7 @@
 #include "blister.h"
 #include "bool.h"
 #include "calls.h"
+#include "code.h"
 #include "error.h"
 #include "gapstate.h"
 #ifdef USE_GASMAN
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Generally cleaned up and (hopefully) improved.

tst/testinstall/declarefunction.tst
```diff
@@ -0,0 +1,45 @@
...
+end
+gap> InstallGlobalFunction(testfunctionA, x -> x);
+gap> testfunctionA;
+function( x ) ... end
+gap> Print(testfunctionA,"\n");
+function ( x )
+    return x;
+end
+gap> name := List([1..100], x -> 'a');;
+gap> DeclareGlobalFunction(name);
```
fingolfin: Minor nitpick (it is fine to ignore this!):

The one issue with this test is that re-running it in the same GAP session won't work, as then these global functions are already defined. Granted, that's not a serious issue most of the time, although I did spend some effort in the past trying to avoid such things.

If one wanted to, one could avoid it here by directly calling `NEW_GLOBAL_FUNCTION( name )`.

src/opers.c
```diff
@@ -2918,10 +2919,28 @@ static Obj NewGlobalFunction(Obj name, Obj nams)
...
+    // We set the location to a description, to make clear the function
+    // hasn't been defined yet
+    const char label[] = "the global function \"%s\" is not yet defined";
+
+    // As the '%s' in 'label' will be replaced with 'namobj', there is
+    // no need for an extra character to store the end-of-string null.
+    Obj    filename = NEW_STRING(strlen(label) + GET_LEN_STRING(namobj));
+    char * buf = CSTR_STRING(filename);
+    sprintf(buf, label, CONST_CSTR_STRING(namobj));
+    SET_LEN_STRING(filename, strlen(buf));
```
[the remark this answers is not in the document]
ChrisJefferson: fixed. I didn't want to try to figure out the answer "hard-wired", but now I don't even have to pay for the strlen

tst/testinstall/declarefunction.tst
ChrisJefferson: fixed

PR #3688, opened 2019-10-06
kernel: remove use of PLAIN_LIST in FuncUNITE_BLIST_LIST

fingolfin:
We achieve this by using LEN_LIST, ELM_LIST resp. ELM0_LIST in a few select
places. The performance impact should be negligible for all our current
applications.

We also switch in a few places to using ELM_PLIST to access plist entries,
instead of peeking through CONST_ADDR_OBJ. This gives the same result (on the
machine code level) in an optimized release build, but affords us better
checks in a debug builds, and easier to read code.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1341,7 +1341,7 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
 
         /* loop over <sub> and set the corresponding entries to 'true'     */
         s = INT_INTOBJ( GET_ELM_RANGE( list, 1 ) );
-        for ( l = 1; l <= LEN_LIST(sub); l++ ) {
+        for (l = 1; l <= lenSub; l++) {
             if ( ptrSub[l] != 0 ) {
 
                 /* if <sub>[<l>] is an integer it is very easy             */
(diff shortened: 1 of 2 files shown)


src/blister.c
```diff
@@ -1443,28 +1436,28 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
...
         for ( l = 1; l <= lenList; l++ ) {
-
+            prev = elm;
+            elm = ELM_LIST(list, l);
             /* test if <list>[<l>] is in <sub>                             */
-            if ( l == 1 || LT(CONST_ADDR_OBJ(list)[l-1],CONST_ADDR_OBJ(list)[l]) ){
-                while ( k <= lenSub
-                     && LT(CONST_ADDR_OBJ(sub)[k],CONST_ADDR_OBJ(list)[l]) )
+            if (l == 1 || LT(prev, elm)) {
+                while (k <= lenSub && LT(ELM_PLIST(sub, k), elm))
```
fingolfin: Note that this code was broken if `list` contains holes, as then we were passing NULL to `LT`, which would result in nonsense or crashes. Now we get a proper error, due to the use of `ELM_LIST`. (I guess I should add a test case for that, too). 

The new code will still do weird things if `sub` contains holes, but that means it is not worse than it was before.

ChrisJefferson: Is there a good reason not to just remove all the `PLAIN_LIST`s and make everything ELM_LIST variants? (or are you working through in phases)?

fingolfin: Good question, I forgot to remove the last one. But I corrected that, so all *direct* `PLAIN_LIST` calls in that function are gone. Mind you, it is still indirectly called via `IsSet`; changing that is

PR #3689, opened 2019-10-06
Fix BlistList for two ranges

fingolfin:
Several results were wrong, and had been wrong since at least GAP 4.4.
E.g. before this commit:

    gap> BlistList([0..3], [-3..-1]);
    [ true, true, true, false ]

Now we get the correct result

    gap> BlistList([0,1,2,3], [-3..-1]);
    [ false, false, false, false ]

One major issue was the use of unsigned variables to store signed data. Two
variables were only of type `long` instead of `Int`, but on 64bit, this could
lead to further issues.

The logic handling the intersection was broken and overly complicated. The
rewritten logic should be simpler.

Finally, there is a tiny optimization enabled by switching from 1-based
indexing to 0-based.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1286,9 +1286,9 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
     UInt                bit;            /* one bit of block                */
     Int                 lenList;        /* logical length of the list      */
     const Obj *         ptrSub;         /* pointer to the sublist          */
-    UInt                lenSub;         /* logical length of sublist       */
-    UInt                i, j, k = 0, l;     /* loop variables                  */
-    long                s, t;           /* elements of a range             */
+    Int                 lenSub;         /* logical length of sublist       */
+    Int                 i, j, k, l;     /* loop variables                  */
+    Int                 s, t;           /* elements of a range             */
 
     RequireSmallList("UniteBlistList", list);
     RequireBlist("UniteBlistList", blist);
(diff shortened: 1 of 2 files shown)


src/blister.c
```diff
@@ -1313,19 +1313,22 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
...
         t = INT_INTOBJ( GET_ELM_RANGE( sub, 1 ) );
-        if ( s <= t )  i = t - s + 1;
-        else           i = 1;
 
-        if ( i + lenSub - 1 <= lenList )  j = i + lenSub - 1;
-        else                              j = lenList;
+        // compute bounds
+        i = t - s;
+        j = lenSub + i;
+        if (i < 0)
```
fingolfin: Note that it is crucial to compute i and j *before* capping the bounds; clearly the previous two lines do not commute; yet in the old code, they were in the wrong order.

src/blister.c
```diff
@@ -1286,9 +1286,9 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
     UInt                bit;            /* one bit of block                */
     Int                 lenList;        /* logical length of the list      */
     const Obj *         ptrSub;         /* pointer to the sublist          */
-    UInt                lenSub;         /* logical length of sublist       */
-    UInt                i, j, k = 0, l;     /* loop variables                  */
-    long                s, t;           /* elements of a range             */
+    Int                 lenSub;         /* logical length of sublist       */
+    Int                 i, j, k, l;     /* loop variables                  */
+    Int                 s, t;           /* elements of a range             */
```
fingolfin: s, t were just of type `long`, so in 64 bit mode, with compilers where long is only 32 bit (basically anywhere except some Microsoft / Windows compilers), there could be overflow here. I didn't add tests for that, though.

src/blister.c
```diff
@@ -1313,19 +1313,22 @@ static Obj FuncUNITE_BLIST_LIST(Obj self, Obj list, Obj blist, Obj sub)
...
-        for ( ; k+BIPEB <= j; k += BIPEB )
-            ptrBlist[(k-1)/BIPEB] = ~(UInt)0;
-        for ( ; k <= j; k++ )
-            ptrBlist[(k-1)/BIPEB] |= (1UL << (k-1)%BIPEB);
+        for ( k = i; k < j && k%BIPEB != 0; k++ )
+            ptrBlist[k/BIPEB] |= (1UL << k%BIPEB);
+        for ( ; k+BIPEB < j; k += BIPEB )
+            ptrBlist[k/BIPEB] = ~(UInt)0;
+        for ( ; k < j; k++ )
+            ptrBlist[k/BIPEB] |= (1UL << k%BIPEB);
```
fingolfin: Note that `k-1` changed to `k` by adjusting the bounds. I hope the optimizer would have produced identical results in both cases; but the result is certainly slightly easier to read.

fingolfin:
Backported to stable-4.11 in 8761c8f6f3ceb5abd5b0253a6c773caf5be2c39d

PR #3690, opened 2019-10-06
Fix NullspaceModQ and renamed it to NullspaceModN

### LINKED ISSUE #3614: Bug in NullspaceModQ
The following was reported on https://math.stackexchange.com/questions/3329100:
```
gap> NullspaceModQ([[2]],8);
[ [ 0 ] ]
```
But it really should return the list `[ [0], [4] ]`.

fingolfin:
Fixes #3614 and adds a manual example demonstrating the fixed behavior.
The old name NullspaceModQ is kept as an alias for backwards compatibility.
Also updated the documentation for NullspaceModN and BasisNullspaceModN.

Should be OK to backport to stable-4.11


fingolfin:
Backported to stable-4.11 in a1e58ee0e6f37c0a80ad6fbe5eb0313dc07a1163

Issue #3400, opened 2019-04-10
Try harder to use the correct indefinite article (a/an) in messages to the user


wilfwilson:
Sorry that this is a vague issue, perhaps the scope should be narrowed, feel free to edit this to something more actionable. But I wanted to make an issue while I still remembered. Some messages to the user in GAP, such as the following one (produced by `RequireArgumentEx`), use the wrong article:
https://github.com/gap-system/gap/blob/bd34ff4d5f54789fef7ff17faeae75649ce441c6/tst/testinstall/bitfields.tst#L63
To be clear, 'a empty' should be 'an empty'. Although it's not a huge problem, it still doesn't give the best impression of GAP, so I think we should attempt to improve the situation.

@fingolfin suggested a (partial?) solution in [a comment on](https://github.com/gap-system/gap/pull/3394/files#r272914689) #3994. What does anyone think? I can think of the following steps:
* Identify where this might occur in GAP (Perhaps `RequireArgument` is the only place where this happens now? @fingolfin said that he's been trying to solve this problem).
* Decide how far we are willing to go to choose the most appropriate article.
* Implement that.

ChrisJefferson:
Looking around, there are quite a few place where we print "a %s", where %s is set to TNAM_OBJ (usually, but not always as part of a "(not a %s)".

Therefore I'm tempted to suggest a new function, TNAM_OBJ_WITH_ARTICLE (that's a horrible name), which returns TNAMs with the article included.

ChrisJefferson:
A better option might be to add a modifier to '%s', say '%as', which makes it attach 'a/an' as a prefix.

Chris

wilfwilson:
Yes, perhaps introducing `%as` and putting the logic behind there would be a nice solution.

fingolfin:
I am not so sure about `%as` being a good solution at this point. Perhaps if we identify more places where this would be used. But note that providing a fully generic solution that can be used to print arbitrary stuff is highly non-trivial, as there are a ton of exceptions for the use of "a" vs. "an". To get an idea of what is needed, see e.g. here: <https://github.com/rossmeissl/indefinite_article/blob/master/lib/indefinite_article.rb>

Whereas for TNUMs, we have a fixed list of <= 256 entries, and we only need to make sure the articles are correct for those. That's a much simpler problem.

So I'd really start with that, and if we then determine that we really need article inflection in more places, *then* we can look into a more generic solution.

fingolfin:
[...]

PR #3691, opened 2019-10-06
kernel: tweak error messages to say "an object"

fingolfin:
... instead of "a object", and similar.

This makes some progress on issue #3400 and is otherwise about the least import PR we had in a long time, I guess ;-).

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -568,7 +568,6 @@ Obj RequireArgumentEx(const char * funcname,
 {
     char msgbuf[1024] = { 0 };
     Int  arg1 = 0;
-    Int  arg2 = 0;
 
     if (funcname) {
         strlcat(msgbuf, funcname, sizeof(msgbuf));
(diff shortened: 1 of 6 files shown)


ChrisJefferson:
I understand the reason someone would want this, but it's going to break the tests of lots of packages, and make it hard to easily write tests with error messages which work in versions of GAP both before and after this patch.

I don't have a good idea on how to fix this. A configuration option feels overkill.

fingolfin:
I actually do not believe this will break a single package test: 
1. These errors concern validation of GAP kernel functions; no package test ought to trigger any of those.
2. To verify this, I grepped through all deposited packages plus a ton of package git repositories I have on my computer, using `rg -F '(not a '`, and as far as I could tell, none of them trigger any of the affected kernel error messages. There were only a few hits anyway; those in `.tst` files either were inside comments (for `atlasrep`), or came from package functions (for `datastructures`).
3. In GAP 4.11, we completely overhauled these errors (by adding those `Require*` macros, and also by changing many TNAMs); it's very likely that those changes would have broken packages which this PR here would break. But none were broken, which seems to support my point 2.

ChrisJefferson:
Yep, I was too cautious.

PR #3693, opened 2019-10-07
Improve ViewString so Strings are correctly escaped

### LINKED ISSUE #2884: ViewString does not correctly escape quotes inside strings
### Observed behaviour

`ViewString` does not correctly escape quotes inside strings.

```
gap> Print(ViewString("fo\"o"));
"fo"o"
```

### Expected behaviour

The same behavior one gets from directly evaluating the string in the console.
```
gap> "fo\"o";
"fo\"o"
```

```
 ┌───────┐   GAP 4.9.2 of 04-Jul-2018
 └───────┘   Architecture: x86_64-apple-darwin17.7.0-default64
 Configuration:  gmp 6.1.2
 Packages:   AClib 1.3, Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.9, Browse 1.8.8, CRISP 1.4.4, Cryst 4.1.17, CrystCat 1.1.8, CTblLib 1.2.2,
             FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.1, IO 4.5.1, IRREDSOL 1.4, LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.14, PrimGrp 3.3.1,
             RadiRoot 2.8, ResClasses 4.7.1, SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.6, TransGrp 2.0.2, utils 0.54
```

ChrisJefferson:
Fixes #2884

This fixes ViewString not correctly escaping strings. This is done by making the function which prints strings to the terminal generic, so we can print a string (with escape characters) to a buffer.

Note that, to make the kernel level stuff easier, I print to a plist of strings which have to be concatenated together, which then actually gets done back at the GAP level. If this offends anyone I could make a growing string and do this in one go, but this seemed easier.

### DIFF
--- a/src/stringobj.c
+++ b/src/stringobj.c
@@ -57,6 +57,7 @@
 #include "error.h"
 #include "gaputils.h"
 #include "io.h"
+#include "listfunc.h"
 #include "lists.h"
 #include "modules.h"
 #include "opers.h"
@@ -529,8 +530,11 @@ static Obj CopyString(Obj list, Int mut)
 /****************************************************************************
 **
 *F  PrintString(<list>) . . . . . . . . . . . . . . . . . . .  print a string
+*F  FuncCHUNK_ESCAPE_STRING(<list>) . . . . . . . escape a string as a string
 **
 **  'PrintString' prints the string with the handle <list>.
+**  'CHUNK_ESCAPE_STRING' returns a list of strings which, when concatenated,
+**  produce a String containing what PrintString outputs.
 **
 **  No linebreaks are  allowed, if one must be inserted  anyhow, it must
 **  be escaped by a backslash '\', which is done in 'Pr'.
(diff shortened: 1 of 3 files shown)


src/stringobj.c
```diff
@@ -529,8 +530,11 @@ static Obj CopyString(Obj list, Int mut)
 /****************************************************************************
 **
 *F  PrintString(<list>) . . . . . . . . . . . . . . . . . . .  print a string
+*F  FuncCHUNK_ESCAPE_STRING(<list>) . . . . . . . escape a string as a string
 **
 **  'PrintString' prints the string with the handle <list>.
+**  'CHUNK_ESCAPE_STRING' returns a list of strings which, when concatenated,
+**  produce a String containing what PrintString outputs.
```
fingolfin: What I don't understand: Why not instead provide a function `VIEW_STRING_FOR_STRING` which returns the already concatenated result? Do you have any applications in mind where one would need these separate "chunks"?

src/stringobj.c
```diff
@@ -542,14 +546,36 @@ static Obj CopyString(Obj list, Int mut)
...
+
+// Output to a list of Strings
+void ToStringOutputter(void * data, char * strbuf, UInt len)
+{
+    Obj list = (Obj)data;
+
+    strbuf[len++] = '\0';
+    AddPlist(list, MakeImmString(strbuf));
+    CHANGED_BAG(list);
+}
```
fingolfin: To switch from `CHUNK_ESCAPE_STRING` to `VIEW_STRING_FOR_STRING`, I think almost the only thing we would have to do is to change this function to instead append strbuf to a T_STRING given via `data`, i.e.,:
```C
void ToStringOutputter(void * data, const char * buf, UInt lenbuf)
{
    Obj str = (Obj)data;
    Obj lenstr = GET_LEN_STRING(str);
    GROW_STRING(str, lenstr + lenbuf);
    memcpy(CHARS_STRING(str) + lenstr, buf, lenbuf);
    CHARS_STRING(str)[lenstr + lenbuf] = '\0';
}

File: src/stringobj.c [hunk 3]
```diff
@@ -606,12 +632,32 @@ void PrintString(Obj list)
...
+{
+    if (!IS_STRING(string)) {
+        RequireArgument("ConvString", string, "must be a string");
+    }
+
+    if (!IS_STRING_REP(string)) {
+        string = CopyToStringRep(string);
+    }
+
+    Obj list = NEW_PLIST(T_PLIST, 1);
```
fingolfin: ... well and here of course we'd have to use `NEW_STRING`.

fingolfin:
Seems this needs to be rebased now, too. Also, labels need to be set (at the very least a "release notes" label)

src/stringobj.c
ChrisJefferson: Nope, just generally trying to make the C-level code as simple as possible

src/stringobj.c
ChrisJefferson: I tried this, but it is producing the wrong answer. I haven't figured out why yet.

fingolfin: Well, I forgot the `SET_LEN_STRING` in my sample code... I took the liberty of updating this PR with a commit of mine which makes the proposed changes. If you are happy with them, feel free to squash them into your initial commit

fingolfin:
@ChrisJefferson I updated my commit to fix a comment; otherwise, this is ready to merge from my POV, but since I modified your PR, I do not want to merge this w/o your explicit approval (in particular, if you do not approve, then I don't want to merge it).

ChrisJefferson:
Sorry, I was ignoring it while the tests still failed, but I notice this is just jenkins/gap-system being badly behaved. I'm happy with this.

PR #3694, opened 2019-10-07
Deprecate undocumented extra arguments for DeclareGlobalFunction and InstallGlobalFunction

fingolfin:
The extra argument was never documented for `InstallGlobalFunction`. In
contrast, for `DeclareGlobalFunction`, the documentation for a long time
claimed that it required exactly (!) two arguments, although all
examples only showed one argument, and the documentation never said what
the other argument was supposed to do. Various people then made
differing guesses as to the supposed meaning of the extra arguments:
some put in an info string; some put in a list of input filters; some
just a single filter. This PR removes several of these from the GAP
library itself.

Various packages still use these extra arguments, and loading them will
trigger the warnings. Some already fixed it in their repositories and
just need to make releases. When the time comes for releasing GAP 4.12,
we can see if all packages are adapted, and then depending on that
either turn the warning into an error; or conversely, increase the
warning level to 2, to hide it from regular users who might be confused
by the message.

### DIFF
--- a/lib/dict.gd
+++ b/lib/dict.gd
@@ -467,7 +467,7 @@ DeclareRepresentation( "IsDenseHashRep",
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareGlobalFunction( "DenseHashTable", [] );
+DeclareGlobalFunction( "DenseHashTable" );
 
 
 #############################################################################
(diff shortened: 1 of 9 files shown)


alex-konovalov:
@fingolfin for most packages this should only break _their_ tests, but there is a case when it broke the bugfix test for the core GAP system when GAP is loaded with `-A` option: https://travis-ci.org/gap-infra/gap-docker-master-testsuite/jobs/599991524. I have requested new Crisp release in https://github.com/bh11/crisp/issues/4

fingolfin:
? Fixing that requires no crisp release, just some trivial modifications to that tst file to tweak info levels

alex-konovalov:
Submitted #3726 to silence the warning in the bugfix test.

PR #3696, opened 2019-10-07
kernel: use Require* macros more; and refactor some code in trans.cc

fingolfin:
Thank You!

# Description

## Text for release notes 

## Further details

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/src/trans.cc
+++ b/src/trans.cc
@@ -1482,10 +1482,10 @@ static Obj FuncSMALLEST_IDEM_POW_TRANS(Obj self, Obj f)
 ** GAP level functions for regularity of transformations
 *******************************************************************************/
 
-// Returns True if the transformation or list <t> is injective on the list
+// Returns True if the transformation or list <obj> is injective on the list
 // <list>.
 
-static Obj FuncIsInjectiveListTrans(Obj self, Obj list, Obj t)
+static Obj FuncIsInjectiveListTrans(Obj self, Obj list, Obj obj)
 {
     UInt    n, i, j;
     const UInt2 * ptt2;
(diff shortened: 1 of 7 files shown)


tst/testinstall/trans.tst
```diff
@@ -749,25 +749,26 @@ false
...
+Error, IsInjectiveListTrans: <list> must be a small list (not a transformation\
+ (large))
 gap> IsInjectiveListTrans([1, 2], [2,3]);
-Error, the second argument must consist of positive integers in the range [1 .\
-. 2]
+Error, <obj> must be a list of positive small integers in the range [1 .. 2]
 gap> IsInjectiveListTrans([1, []], f);
-Error, the entries of the first argument must be positive integers (not a empt\
-y plain list)
+Error, <list> must be a list of positive small integers (not a empty plain lis\
```
wilfwilson: I was under the impression that the "not a empty plain list" article inflection was resolved by #3691. I guess this must be triggering some different code than used by #3691, right?

fingolfin: Indeed. There are still some places in the kernel which don't use `RequireArgumentEx` to print error messages involving `(not a %s)`, most of them in `trans.cc`. This PR actually replaces some more of these(`git grep -F '(not a ' src | wc -l` give me 40 on `master` and 27 with this PR). and then there are a few where there is a good reason for that. E.g. in some of them, the error message contains an other `%d` print formatter, and that currently can't be handed by `RequireArgumentEx`. Example from `trans.cc`:
```C
        if (!IS_INTOBJ(ELM_LIST(ran, i))) {
            ErrorQuit("TransformationListListNC: <ran>[%d] must be a small "
                      "integer (not a %s)",
                      (Int)i, (Int)TNAM_OBJ(ELM_LIST(ran, i)));
        }
```
I guess we could add another helper to deal with this particular use case, though I think it's really only a very small number, so may not be worth it.

src/trans.cc
```diff
@@ -1486,38 +1382,32 @@ static Obj FuncSMALLEST_IDEM_POW_TRANS(Obj self, Obj f)
...
+    if (TNUM_OBJ(obj) == T_TRANS2) {
+        ptt2 = CONST_ADDR_TRANS2(obj);
+        for (i = LEN_LIST(list); i >= 1; i--) {
+            val = ELM_LIST(list, i);
             if (!IS_POS_INTOBJ(val)) {
                 ErrorQuit(
-                    "the entries of the first argument must be positive "
-                    "integers (not a %s)",
+                    "<list> must be a list of positive small integers (not a %s)",
                     (Int)TNAM_OBJ(val), 0L);
```
fingolfin: Hmm, I think I may have accidentally made the error message worse. To fix this, I should either drop the `(not a %s)`, as the messages makes it sound as if this refers to the overall list; or perhaps it could be something like this:
```C
  ErrorQuit("IsInjectiveListTrans: <list>[%d] must be a positive small integer (not a %s)", i, (Int)TNAM_OBJ(val));
```

tst/testinstall/trans.tst
wilfwilson: I was just curious, thanks for the explanation 🙂 As far as I'm concerned I agree, it's probably not worth it.

src/trans.cc
fingolfin: I have a new PR ready which fixes this. I think it would be easiest to just merge this PR here as it is, even though that error above is flawed; and then I'd immediately submit a new PR which replaces this bit code with an IMHO much better solution.

wilfwilson: Fine by me.

PR #3700, opened 2019-10-11
Increment kernel major version

fingolfin:
This way, code which wants to check for GAP 4.12 kernel features
can differentiate it from GAP 4.11

This is motivated by PR #3699 resp. https://github.com/gap-packages/Semigroups/pull/625

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -556,7 +556,7 @@ dnl Define kernel version
 dnl
 
 
-gap_kernel_major_version=7
+gap_kernel_major_version=8
 gap_kernel_minor_version=0
 AC_SUBST([gap_kernel_major_version])
 AC_SUBST([gap_kernel_minor_version])


fingolfin:
@alex-konovalov  you just blocked merging this, so your comment doesn't really make sense to me... anyway I hope that we'll get rid of those scripts during the GAP days in the spring, replacing them with something less fragile. Even if not, there will be plenty of time before GAP 4.12, don't you think?

alex-konovalov:
Sorry blocking was accidental. The comment though makes perfect sense - in the past I was taken by surprise several times when missing the kernel version update.

fingolfin:
@alex-konovalov The comment makes sense *now* -- it didn't make sense (to me) in conjunction with blocking this PR. But since you said now this was a mistake, it's clearer....

... except that I still think that if your scripts require manual action each time the kernel version is changed, then those scripts are bad and should be rewritten to not require that. In principle, I would like to change the kernel version much more often, and being held back for this kind of work because some tooling is insufficient is a bad reason, IMHO. THis is just one of the many things we really need to automate. It's a waste if you or any other human has to spend time on dealing with things that could be dealt with automatically.

fingolfin:
@alex-konovalov I made https://github.com/gap-system/gap-distribution/pull/89 for this. Let me know if there are more places that hardcode the GAP arch string.

fingolfin:
@alex-konovalov ping, this was merged, please update your scripts (or possibly instead review and merge my PR that should hopefully get rid of the need for that -- if they don't, I'd be happy to work with you on refining them for the future!)

PR #3703, opened 2019-10-14
kernel: remove last uses of ErrorReturnObj

fingolfin:
A few package still use `ErrorReturnObj`, though:
- Browse
- io (see https://github.com/gap-packages/io/commit/2130f337e930d12bf74d42ab5480f32ac8a2a654) 
- float (see https://github.com/gap-packages/float/pull/50)
- img (see https://github.com/gap-packages/img/pull/20)

I will send patches to the respective package owner (or, in the case of io, apply that myself).

This PR also tweaks `FuncSizeScreen` to use `GetSmallIntEx`

### DIFF
--- a/doc/ref/debug.xml
+++ b/doc/ref/debug.xml
@@ -1007,8 +1007,8 @@ brk> GetRecursionDepth();
 0
 brk> return;
 gap> SetRecursionTrapInterval(-1);
-SetRecursionTrapInterval( <interval> ): <interval> must be a non-negative smal\
-l integer
+Error, SetRecursionTrapInterval: <interval> must be a small integer greater than 5 (n\
+ot the integer -1)
 not in any function
 Entering break read-eval-print loop ...
 you can 'quit;' to quit to outer loop, or
--- a/src/gap.c
+++ b/src/gap.c
@@ -643,24 +643,18 @@ static Obj FuncSizeScreen(Obj self, Obj args)
   }
   else {
     elm = ELMW_LIST(size,1);
-    if (!IS_INTOBJ(elm)) {
-        ErrorMayQuit("SizeScreen: <x> must be an integer", 0, 0);
-    }
-    len = INT_INTOBJ( elm );
+    len = GetSmallIntEx("SizeScreen", elm, "<x>");
     if ( len < 20  )  len = 20;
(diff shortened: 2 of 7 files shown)


src/vecffe.c
```diff
@@ -738,8 +720,6 @@ static Obj FuncADD_ROWVECTOR_VECFFES_3(Obj self, Obj vecL, Obj vecR, Obj mult)
 **
 */
 
-static Obj MultVectorLeftOp; /* BH changed to static */
```
fingolfin: Funny story: we actually failed to initialise `MultVectorLeftOp`, so the code path using it (deleted by this PR) would have crashed.

src/gap.c
```diff
@@ -643,24 +643,18 @@ static Obj FuncSizeScreen(Obj self, Obj args)
...
-    }
-    len = INT_INTOBJ( elm );
+    len = GetSmallIntEx("SizeScreen", elm, "<x>");
     if ( len < 20  )  len = 20;
     if ( MAXLENOUTPUTLINE < len )  len = MAXLENOUTPUTLINE;
   }
 
   /* extract the number                                                  */
-  if ( LEN_LIST(size) < 2 || ELM0_LIST(size,2) == 0 ) {
+  elm = ELM0_LIST(size, 2);
```
fingolfin: Note that ELM0_LIST is explicitly callable with arbitrary positive indices; if the index is out of bounds, it returns 0. This allows us to simplify this code, and avoid the call to `ELMW_LIST` later on.

doc/ref/debug.xml
```diff
@@ -1007,8 +1007,9 @@ brk> GetRecursionDepth();
 0
 brk> return;
 gap> SetRecursionTrapInterval(-1);
-SetRecursionTrapInterval( <interval> ): <interval> must be a non-negative smal\
-l integer
+Error, SetRecursionTrapInterval: <interval> must be a small integer greater \
+than 5 (not the integer -1)
+an 5 (not the value 'fail')
```
ChrisJefferson: what's happening here?

fingolfin: That was wrong (and I even noticed and fixed it yesterday, but then failed to force push *sigh*).

PR #3704, opened 2019-10-14
Two workarounds for old issues.

### LINKED ISSUE #3533: Regression in grpauto and doublecoset benchmarks (memory usage)
This used to work in the master branch, but not anymore:
```
cd benchmark/grpauto
../../bin/gap.sh -o 1g test1.g
```
results in
```
# line 63, input:
IsomorphismGroups(gp1,gp2)<>fail;
########> Diff in permiso.tst:63
# Input is:
IsomorphismGroups(gp1,gp2)<>fail;
# Expected output:
true
# But found:
```
This is a weekly test, it passed last time successfully on June 22, but now fails.

In addition,
```
cd benchmark/doublecoset
../../bin/gap.sh -o 2g test3.g
```
also fails in 32-bit build on mac:
```
# line 5, input:
START_TEST("doublecoset3.tst");
# line 6, input:
g:=SimpleGroup("Co3");;
# line 7, input:
m:=MaximalSubgroupClassReps(g);;
# line 8, input:
u:=First(m,x->Index(g,x)=17931375);;
# line 9, input:
dc:=DoubleCosetRepsAndSizes(g,u,u);;
Panic in src/sysmem.c:666: cannot extend the workspace any more!!!
```
The St Andrews-only link to Jenkins is [here](https://gap-ci.cs.st-andrews.ac.uk/view/GAP-weekly/job/GAP-master-benchmarks/).

hulpke:
as the default assertions completely distort the cost -- all time is wasted
on checking homomorphism properties again and again.

This resolves #3533

### DIFF
--- a/benchmark/grpauto/permiso.tst
+++ b/benchmark/grpauto/permiso.tst
@@ -2,6 +2,8 @@
 ##
 ## Warning: most of the later tests need more than the default memory allocation
 
+gap> SetAssertionLevel(0);;
+
 #
 # Too hard work for permiso
 #
--- a/lib/clashom.gi
+++ b/lib/clashom.gi
@@ -1054,7 +1054,7 @@ local clT,	# classes T
     Info(InfoHomClass,1,"fused to ",Length(newreps)," classes");
   od;
   
-  Assert(2,Sum(clout,i->Index(F,i[2]))=Size(F)-Size(M));
+  if Sum(clout,i->Index(F,i[2]))<>Size(F)-Size(M) then return fail;fi;
 
   Info(InfoHomClass,2,Length(clin)," inner classes, total size =",
         Sum(clin,i->Index(F,i[2])));


ChrisJefferson:
Just as a sanity check, is replacing an Assert with a `return fail` going to mess anything up, or is it handled somewhere sensible?

hulpke:
@ChrisJefferson 
The code already handles the rare situation that the function does not discover all classes, in this case the calculation is restarted . (In the long run this means probably a rewrite). So returning `fail`will be handled properly, even though this PR does not do so.
(The intention was to have a minimal workaround that has no chance of introducing other problems.)

ChrisJefferson:
Great, thanks

alex-konovalov:
This should be backported to 4.11.

fingolfin:
Backported to stable-4.11 in ed958f08a5ef5c1caf76d1aff6e65f6069e4e8c0

PR #3702, opened 2019-10-13
Make TmpNameAllArchs obsolete by enhancing TmpName

fingolfin:
Back in 2011, a new library function `TmpNameAllArchs` was added which wraps the kernel function `TmpName` and adds some Windows specific sauce to it. This PR enhances `TmpName` so that `TmpNameAllArchs` can be turned into a simple synonym for `TmpName`. It does so by aligning the code of `FuncTmpName` and `FuncTmpDirectory` more closely. (They still differ in that the latter honors the the `TMPDIR` env var, the former does not; we might want to change that, but I didn't want to make too big a change in one go.

There is another difference between the old and new code: the old used `C:/WINDOWS/Temp/`, the new uses `/cygdrive/c/WINDOWS/Temp/`. I *hope* that doesn't matter or even is an improvement, *but* it would be really good if somebody with Windows access could verify that (beyond what little the tests in our test suite do).

Resolves #3328

The other change in this PR is the removal of the library function `ShortFileNameWindows` which @hulpke added in 2009 but which hasn't ever been used by GAP or any package, as far as I can tell. It's also undocumented, and it seems not so useful to access 8.3 filenames these days. But perhaps I am mistaken, and there are usecases (in particular, I have no idea what the motivation for adding it back then was).

PR #3707, opened 2019-10-18
Two corrections in recently changed code

hulpke:
These are two fixes of (stupid, and they are mine) bugs in recently changed (4.11) code, both of which cause subsequent code to give error messages. While I ran into both I don't have easy examples to test the situation. The errors are:

a) In BoundedIndexAbelianized the test for being elementary was wrong.
b) `ReducedRelationMat` could delete the whole matrix (if it was 0), resulting in a subsequent error for lack of columns.
c) Also changed is an example imn testextra that was not spotted before (but now was noted in #3702)


fingolfin:
Backported to `stable-4.11` in 4da0315bed2dedd2cc7b0acce9c6d10aeb6524d6 + a21c3bbed9011bf6e0f52d7685b5d703e9db1c07 + d91899d3e9f1d41f3868cee11aafbbf40f77da75

PR #3710, opened 2019-10-21
kernel: include hpc/guards.h in compiled.h

fingolfin:
... so that packages such as ZeroMQInterface can use guards in their
kernel extensions without including additional headers.

This unbreaks `ZeroMQInterface` compiled against HPC-GAP, and should be backported to 4.11

### DIFF
--- a/src/compiled.h
+++ b/src/compiled.h
@@ -78,6 +78,7 @@ extern "C" {
 
 #ifdef HPCGAP
 #include "hpc/aobjects.h"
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Thanks - fix worked for me, and Travis builds passed too.

alex-konovalov:
Backported to stable-4.11 in db7dc597c402557566d240f95235657a0f9f18f9

PR #3711, opened 2019-10-21
Adjust TestPackage to return info about test result

fingolfin:
... at least where possible; the value `true` indicates success, `false` means
that tests were run but failed, and `fail` that there was an error, e.g. no
tests were found or it was not possible tor run them.

This contains code which is contained in several repositories under https://github.com/gap-infra/ such as here: https://github.com/gap-infra/gap-docker-pkg-tests-master-devel/blob/master/ci.g

We could backport this to stable-4.11 -- or are there concerns about that?

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -980,14 +980,14 @@ InstallGlobalFunction( "TestPackage", function(pkgname)
 local testfile, str;
 if not IsBound( GAPInfo.PackagesInfo.(pkgname) ) then
     Print("#I  No package with the name ", pkgname, " is available\n");
-    return;
+    return fail;
 elif LoadPackage( pkgname ) = fail then
     Print( "#I ", pkgname, " package can not be loaded\n" );
-    return;
+    return fail;
 elif not IsBound( GAPInfo.PackagesInfo.(pkgname)[1].TestFile ) then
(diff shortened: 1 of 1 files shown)


lib/test.gi
```diff
@@ -980,36 +980,46 @@ InstallGlobalFunction( "TestPackage", function(pkgname)
...
         Print( "#I Test file `", testfile, "' for package `", pkgname,
         " version ", GAPInfo.PackagesInfo.(pkgname)[1].Version, " is not readable\n" );
+        return fail;
+    else
+        # At this point, the READ succeeded, but we have no idea what the
+        # outcome of that test was. Hopefully, that file printed a message of
+        # its own and then terminated GAP with a suitable error code (e.g. by
+        # using TestDirectory with exitGAP:=true); in that case we never get
+        # here an all is fine.
+        return "UNKNOWN";
```
ChrisJefferson: I'd probably just go `fail` here again, as we don't know the result, and people trying to test all packages might get confused by the sudden string. Also "here an all is fine" -> "here and all is fine".

fingolfin: Agreed, and done / fixed

fingolfin:
Backported to `stable-4.11` in 0f9cef2882d4c04399ae240dc0d4f56201db84d9

PR #3713, opened 2019-10-23
Add tracing and counting of built-in operations

ChrisJefferson:
This is some code I've had a branch of a long time, cleaned up.

Sometimes, for papers and general research it is interesting to know things like how many times + is applied to integers, or * or ^ is applied to permutations.

This can't be done by any of GAP's normal tracing functionality at present. Also, printing a message every time one of these functions is called would be far too much.

This adds `TraceInternalMethods`, which produces a record which contains how many times each operation is applied to each tnum (or each pair of tnums for 2-argument methods).

This doesn't try to catch when kernel functions add integers, only when these operations are performed at the GAP level.

This still needs tests, but before I put more work into polishing I wanted to see if other people found it useful.

### DIFF
--- a/lib/oper.g
+++ b/lib/oper.g
@@ -615,6 +615,93 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
     fi;
 end );
 
+#############################################################################
+##
+##
+##  <#GAPDoc Label="TraceInternalMethods">
+##  <ManSection>
+##  <Func Name="TraceInternalMethods" Arg=''/>
+##  <Func Name="UntraceInternalMethods" Arg=''/>
+##  <Func Name="GetTraceInternalMethodsCounts" Arg=''/>
+##  <Func Name="ClearTraceInternalMethodsCounts" Arg=''/>
+##
+##  <Description>
+##  <Ref Func="TraceInternalMethods"/> enables tracing for all internal methods.
+##  Internal methods are methods which implement many fundamental operations in GAP.
+##  In this version of GAP, the internal methods which can be traced are:
+##  <List>
+##  <Mark>Zero, ZeroMut</Mark><Item>Mutable and Immutable <Ref Attr="Zero"/></Item>
+##  <Mark>AInv, AInvMut</Mark><Item>Mutable and Immutable <Ref Attr="AdditiveInverse"/></Item>
+##  <Mark>One, OneMut</Mark><Item>Mutable and Immutable <Ref Attr="One"/></Item>
+##  <Mark>Inv, InvMut</Mark><Item>Mutable and Immutable <Ref Attr="Inverse"/></Item>
+##  <Mark>Sum</Mark><Item>The operator <Ref Oper="\+"/></Item>
+##  <Mark>Diff</Mark><Item>The operator <C>-</C> operator</Item>
+##  <Mark>Prod</Mark><Item>The operator <Ref Oper="\*"/></Item>
+##  <Mark>Quo</Mark><Item>The operator <Ref Oper="\/"/></Item>
+##  <Mark>LQuo</Mark><Item>The left-quotient operator</Item>
+##  <Mark>Pow</Mark><Item>The operator <Ref Oper="\^"/></Item>
+##  <Mark>Comm</Mark><Item>The operator <Ref Oper="Comm"/></Item>
+##  <Mark>Mod</Mark><Item>The operator <Ref Oper="\mod"/></Item>
+##  </List>
+##  <P/>
+##  <Ref Func="UntraceInternalMethods"/> turns tracing off.
+##  As these methods can be called hundreds of thousands of times in simple GAP
+##  code, there isn't a statement printed each time one is called. Instead, the
+##  method <Ref Func="GetTraceInternalMethodsCounts"/> returns how many times
+##  each operation has been applied to each type of variable (the type of a
+##  variable can be found with the <C>TNAM_OBJ</C> method).
+##  The return value for two argument operators is a record of records <C>r</C>, where
+##  <C>r.op</C> stores information about operator <C>op</C>. For one argument operators
+##  <C>r.op.i</C> stores how many times <C>op</C> was called with an argument of type
+##  <C>i</C>, while for two argument operators <C>r.op.i.j</C> stores how many times
+##  <C>op</C> was called with arguments of type <C>i</C> and <C>j</C>.
+##  <Log><![CDATA[
+## gap> TraceInternalMethods();
+## true
+## gap> 2+3+4+5+6;;
+## gap> 2.0+2.0;;
+## gap> 3^(1,2,3);;
+## gap> GetTraceInternalMethodsCounts();
+## rec( Pow := rec( integer := rec( ("permutation (small)") := 1 ) ),
+##  Sum := rec( integer := rec( integer := 4 ),
+##      macfloat := rec( macfloat := 1 ) ) )
+## # 'macfloat' is a floating point number
+## gap> UntraceInternalMethods();
+##  ]]></Log>
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+
+# The return type here is stored as a record, as 0 is a valid TNUM.
+BIND_GLOBAL("GetTraceInternalMethodsCounts", function()
+    local ret, type, i, j, counts,member, nicename;
+    counts := GET_TRACED_INTERNAL_METHODS_COUNTS();
+    ret := rec();
+    for type in REC_NAMES(counts) do
+        # Drop the 'Funcs' part
+        nicename := type{[1..LENGTH(type)-5]};
+        ret.(nicename) := rec();
+        member := counts.(type);
+        for i in [1..LENGTH(member)] do
+            if IsBound(member[i]) then
+                if IS_LIST(member[LENGTH(member)]) then
+                    # Is a 2D array
+                    ret.(nicename).(GET_TNAM_FROM_TNUM(i-1)) := rec();
+                    for j in [1..LENGTH(member[i])] do
+                        if IsBound(member[i][j]) then
+                            ret.(nicename).(GET_TNAM_FROM_TNUM(i-1)).(GET_TNAM_FROM_TNUM(j-1)) := member[i][j];
+                        fi;
+                    od;
+                else
+                    # Is a 1D array
+                    ret.(nicename).(GET_TNAM_FROM_TNUM(i-1)) := member[i];
+                fi;
+            fi;
+        od;
+    od;
+    return ret;
+end);
 
 #############################################################################
 ##
@@ -2068,3 +2155,4 @@ BIND_GLOBAL( "RECALCULATE_ALL_METHOD_RANKS", function()
 
     Assert(2, CHECK_ALL_METHOD_RANKS());
 end );
+
--- a/src/modules_builtin.c
+++ b/src/modules_builtin.c
@@ -21,6 +21,7 @@
 #include "objset.h"
 #include "profile.h"
 #include "syntaxtree.h"
+#include "tracing.h"
 #include "vec8bit.h"
 #include "vecffe.h"
 #include "vecgf2.h"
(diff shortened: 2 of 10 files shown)


ChrisJefferson:
Thanks for the corrections. I've just put an explicit list of the "internal methods" which are traced.

lib/oper.g
```diff
@@ -615,6 +615,85 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+##  method <Ref Func="GetTracedInternalMethodsCounts"/> returns how many times
+##  each operation has been applied to each type of variable (the type of a
+##  variable can be found with the <C>TNUM_OBJ</C> method).
+##  <Log><![CDATA[
+## gap> TraceInternalMethods();
+## true
+## gap> 2+3+4+5+6;;
+## gap> 2.0+2.0;
+## gap> GetTracedInternalMethodsCounts();
+## rec( Sum := rec( 0 := rec( 0 := 4 ), 6 := rec( 6 := 1 ) ) )
```
DominikBernhardt: To be honest I am not sure if it is clear how to interpret this output. So, in your example
`GetTracedInternalMethodsCounts().sum;` yields `rec( 0 := rec( 0 := 4 ), 6 := rec( 6 := 1 ) )`. Why are the entries of this themselves records again? The first entry concerns integers (TNUM type 0), and it is a record with entry `0:=4`. Maybe this could be made clearer in the documentation? If I'm the only one that is not happy with this output I'm happy to remove this comment though.

ChrisJefferson: I can certainly explain better. There are two parts:

Sum is a 2-argument operator, so Sum.0.0 tells you how many times sum got called with 2 arguments, the first of TNUM 0 and the second of TNUM 0.

Why records instead of lists? Because a list of lists here can end up being really big, with mostly unbound values (or full of zeroes, depending on how you treat operators which never get called), which is annoying to read. However, it might still be better to just use a list of lists, as that is a more natural structure.

tst/testinstall/trace.tst
```diff
@@ -46,4 +46,16 @@ gap> Size(a);
 gap> TraceImmediateMethods(false);
 
 #
```
DominikBernhardt: These tests don't cover all of the methods that can be traced. I am happy with leaving the tests as they are now though, I am just not sure if we have a policy for these situations?

lib/oper.g
```diff
@@ -615,6 +615,92 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+##  <C>op</C> was called with arguments of type <C>i</C> and <C>j</C>.
+##  <Log><![CDATA[
+## gap> TraceInternalMethods();
+## true
+## gap> 2+3+4+5+6;;
+## gap> 2.0+2.0;;
+## gap> 3^(1,2,3);;
+## gap> GetTracedInternalMethodsCounts();
+## rec( Pow := rec( integer := rec( ("permutation (small)") := 1 ) ),
+##      Sum := rec( integer := rec( integer := 4 ), macfloat := rec( macfloat := 1 ) ) )
```
DominikBernhardt: In the compiled documentation, the last `)` is outside of the rectangular box in which Examples are printed. Thus, it looks like a overfull hbox. Maybe you could insert another line break?

lib/oper.g
```diff
@@ -615,6 +615,85 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+##  <ManSection>
+##  <Func Name="TraceInternalMethods" Arg=''/>
+##  <Func Name="UntraceInternalMethods" Arg=''/>
+##  <Func Name="GetTracedInternalMethodsCounts" Arg=''/>
+##  <Func Name="ClearTraceInternalMethodsCounts" Arg=''/>
+##
+##  <Description>
+##  <Ref Func="TraceInternalMethods"/> enables tracing for all internal methods.
+##  Internal methods are methods which implement many fundamental operations in GAP.
+##  In this version of GAP, the internal methods which can be traced is:
```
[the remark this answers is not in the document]
DominikBernhardt: Ddi you leave this unchanged on purpose or did you just forget about it?

DominikBernhardt:
Thanks @ChrisJefferson . Overall, I am happy with this now. If you could address the small points above I am happy to approve this and to get in merged.

lib/oper.g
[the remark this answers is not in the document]
wilfwilson: Yes this should be `are` as @DominikBernhardt suggests, not `is`.

lib/oper.g
wilfwilson: If this is tested by Travis etc, then Chris doesn't have any choice about how the output should look; if he adds arbitrary line breaks, then the tests will fail when the manual examples are tested. However I'm not sure where this _is_ ever tested by Travis.

DominikBernhardt: Oh ok, I see. I guess I won't take the risk of a manual formatting now because it might be tested in the future and thing could break?

tst/testinstall/trace.tst
ChrisJefferson: I will add some more tests

lib/oper.g
ChrisJefferson: THis formatting was actually done by me, so I'll shorten/neaten it

tst/testinstall/trace.tst
ChrisJefferson: I've now added a bunch more tests. In many cases the output may be surprising, but that seems to be what GAP uses internally, there isn't a simple mapping to the underlying methods (which does make this function harder to use, I'll admit.

DominikBernhardt: Thanks!

DominikBernhardt:
@ChrisJefferson Tests are failing now. Maybe you could have a look at that? Also, while you're at it, could you change the documentation and change the `is` to `are` as suggested above?

lib/oper.g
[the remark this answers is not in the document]
ChrisJefferson: fixed

lib/oper.g
```diff
@@ -615,6 +615,93 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+##  <Mark>LQuo</Mark><Item>The left-quotient operator</Item>
+##  <Mark>Pow</Mark><Item>The operator <Ref Oper="\^"/></Item>
+##  <Mark>Comm</Mark><Item>The operator <Ref Oper="Comm"/></Item>
+##  <Mark>Mod</Mark><Item>The operator <Ref Oper="\mod"/></Item>
+##  </List>
+##  <P/>
+##  <Ref Func="UntraceInternalMethods"/> turns tracing off.
+##  As these methods can be called hundreds of thousands of times in simple GAP
+##  code, there isn't a statement printed each time one is called. Instead, the
+##  method <Ref Func="GetTracedInternalMethodsCounts"/> returns how many times
```
DominikBernhardt: In the documentation, `GetTracedInternalMethodsCounts` is an undefined reference resulting in a (**??**) after it is mentioned.

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+        Obj list = NEW_PLIST(T_PLIST, 0);
+        ASS_REC(RecordedStats, name, list);
+    }
+    Obj list = ELM_REC(RecordedStats, name);
+
+    UInt tnam = TNUM_OBJ(op) + 1;
+    Obj  val = ELM0_LIST(list, tnam);
+    if (!val) {
+        val = INTOBJ_INT(0);
+    }
```
fingolfin: Could also use `Obj val = ELM_DEFAULT_LIST(list, tnam, INTOBJ_INT(0))`

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+
+    UInt8 intval = Int8_ObjInt(val);
+    intval++;
+    val = ObjInt_Int8(intval);
```
fingolfin: or just `val = SUM(val, INTOBJ_INT(1))` ? Of course there is a performance penalty; but the above code is far from optimal either, and this replacement at least is easy to understand. If performance truly is a concern (there are some good arguments to be made to this end, of course), then I'd instead restrict to INTOBJ, and then one can perform this much faster. But then `ReportWrappedOperation1` is an extern function, and can't be inlined, so this whole setup is far from being efficient anyway. I mean, just accessing a record for each of these operations is a severe performance hit. So, if we say "this simply makes things quite a bit slower, accept it, I'd prefer the simpler code. If we say "we should try to make this as fast as possible", I'd ditch the whole code, and go completely different about it: create C int arrays storing these counts for each unary/binary operation (as 
part of the `DEFINE_OP_WRAPPER1/2` macros, and store everything in those.

src/tracing.h
```diff
@@ -0,0 +1,82 @@
...
+        return Wrap##Array[TNUM_OBJ(op)](op);                                \
+    }                                                                        \
+    void Array##HookActivate(void)                                           \
+    {                                                                        \
+        for (int i = 0; i < LAST_REAL_TNUM; ++i) {                           \
+            Wrap##Array[i] = Array[i];                                       \
+            Array[i] = Wrap##Array##Func;                                    \
+        }                                                                    \
+    }                                                                        \
+                                                                             \
```
fingolfin: Either remove this "empty" line, or add a similar one before `void Array##HookActivate(void)`

src/tracing.h
```diff
@@ -0,0 +1,82 @@
...
+**
+**  Copyright of GAP belongs to its developers, whose names are too numerous
+**  to list here. Please refer to the COPYRIGHT file for details.
+**
+**  SPDX-License-Identifier: GPL-2.0-or-later
+**
+**  This file defines the functions for tracing operations.
+*/
+
+#include "objects.h"
```
fingolfin: This header is missing include guards.

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+
+/****************************************************************************
+**
+*F  InitInfoStats() . . . . . . . . . . . . . . . . . table of init functions
+*/
+static StructInitInfo module = {
+    // init struct using C99 designated initializers; for a full list of
+    // fields, please refer to the definition of StructInitInfo
+    .type = MODULE_BUILTIN,
+    .name = "stats",
```
fingolfin: wrong! now we have two internal modules with identical name (I am actually surprised this didn't trigger an error; we ought to add a check for that; in the meantime, it can potentially break savestates

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+    ASS_LIST(pos, tnam2, val);
+}
+
+typedef struct {
+    voidfunc activate;
+    voidfunc deactivate;
+} voidfuncs;
+
+// Store the list of operators which can have tracing enabled and disabled
+static voidfuncs controllers[64];
```
fingolfin: Where is this 64 coming from? is it just an arbitrary limit?

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+    }
+    int pos = 0;
+    while (pos < 64 && controllers[pos].deactivate != 0) {
+        controllers[pos].deactivate();
+        pos++;
+    }
+    tracking_active = 0;
+    return 0;
+}
+static Obj FuncGET_TRACED_INTERNAL_METHODS_COUNTS(Obj self)
```
fingolfin: Add empty line above

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+typedef struct {
+    voidfunc activate;
+    voidfunc deactivate;
+} voidfuncs;
+
+// Store the list of operators which can have tracing enabled and disabled
+static voidfuncs controllers[64];
+static int       tracking_active;
```
fingolfin: we use `CamelCase` and not `snake_case` (nor `lowerCamelCase`) for global variables (almost) everywhere else...

src/tracing.c
```diff
@@ -0,0 +1,204 @@
...
+
+void InstallOpWrapper(voidfunc activate, voidfunc deactivate)
+{
+    int pos = 0;
+    while (pos < 64 && controllers[pos].activate != 0) {
```
fingolfin: Instead of hardcoding 64 here and elsewhere, it would be better to use `ARRAYSIZE(controllers)`. Or use a global enum for the size. Or don't use a size at all, and instead use a PLIST to store this data.

src/tracing.h
```diff
@@ -0,0 +1,82 @@
...
+
+
+void ReportWrappedOperation1(const char *, Obj op);
+void ReportWrappedOperation2(const char *, Obj op1, Obj op2);
+void InstallOpWrapper(voidfunc, voidfunc);
+
+
+// These are macros which make it simple to wrap
```
fingolfin: This comment looks as if it is incomplete? In any case, it is useless in the current form. Would be nice to have at least some comments explaining what those macros are and how to use them.

src/tracing.h
```diff
@@ -0,0 +1,82 @@
...
+**
+**  This file defines the functions for tracing operations.
+*/
+
+#include "objects.h"
+
+
+void ReportWrappedOperation1(const char *, Obj op);
+void ReportWrappedOperation2(const char *, Obj op1, Obj op2);
+void InstallOpWrapper(voidfunc, voidfunc);
```
fingolfin: Some comments explaining these functions and how/when to use them would be nice

lib/oper.g
```diff
@@ -615,6 +615,93 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+## true
+## gap> 2+3+4+5+6;;
+## gap> 2.0+2.0;;
+## gap> 3^(1,2,3);;
+## gap> GetTraceInternalMethodsCounts();
+## rec( Pow := rec( integer := rec( ("permutation (small)") := 1 ) ),
+##  Sum := rec( integer := rec( integer := 4 ),
+##      macfloat := rec( macfloat := 1 ) ) )
+## # 'macfloat' is a floating point number
+## gap> UntraceInternalMethods();
```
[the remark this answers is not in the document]
ChrisJefferson: One (usually) wants to call `UntraceInternalMethods` before doing anything else, to make sure the counts don't get inflated. `UntraceInternalMethods` does promise not to clear the counts, but `TraceInternalMethods` does clear them. I'll clean up docs around here.

src/modules_builtin.c
```diff
@@ -44,95 +45,57 @@ const InitInfoFunc InitFuncsBuiltinModules[] = {
...
-    InitInfoCode,
-    InitInfoVars,       /* must come after InitExpr and InitStats */
-    InitInfoFuncs,
-    InitInfoOpers,
-    InitInfoInfo,
-    InitInfoIntrprtr,
+    InitInfoIO, InitInfoScanner, InitInfoRead, InitInfoCalls, InitInfoExprs,
+    InitInfoStats, InitInfoCode,
+    InitInfoVars, /* must come after InitExpr and InitStats */
+    InitInfoFuncs, InitInfoOpers, InitInfoInfo, InitInfoIntrprtr,
```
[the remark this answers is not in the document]
ChrisJefferson: I'd be happy to turn off formatting here / reformat. I'd come around to the idea that we should reformat code around changes, particularly when git-clang-format wants to change it.

lib/oper.g
```diff
@@ -615,6 +615,93 @@ BIND_GLOBAL( "TraceImmediateMethods", function( arg )
...
+## gap> TraceInternalMethods();
+## true
+## gap> 2+3+4+5+6;;
+## gap> 2.0+2.0;;
+## gap> 3^(1,2,3);;
+## gap> GetTraceInternalMethodsCounts();
+## rec( Pow := rec( integer := rec( ("permutation (small)") := 1 ) ),
+##  Sum := rec( integer := rec( integer := 4 ),
+##      macfloat := rec( macfloat := 1 ) ) )
+## # 'macfloat' is a floating point number
```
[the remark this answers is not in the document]
ChrisJefferson: No sorry, it's a comment I've put in to help the reader, but that's not obvious.

src/tracing.c
ChrisJefferson: Actually, it turns out this it's 3! (stats.c, profile.c and tracing.c all use .name = "stats")...

src/tracing.c
ChrisJefferson: There's a much more boring reason: SUM uses SumFuncs, which is traced, which would cause an infinite loop. I'll add a comment to say why I'm avoiding SUM here.

src/tracing.c
ChrisJefferson: Yep, just a random(ish) number

src/tracing.c
ChrisJefferson: I started switching to PLIST, but it doesn't really work as we have to store C function pointers

PR #3715, opened 2019-10-24
Allow HPC-GAP to run as a forkable server process.

rbehrends:
This change is prompted by trying to make HPC-GAP work for Jupyter. Note that it changes the meaning of the `-S` command line option and therefore is labeled as WIP. I've put up the pull request so that people can test the functionality of Jupyter + HPC-GAP, even if details still needs to be settled. Note also that so far I've only tested this on macOS.

This commit includes the following changes:

1. The Boehm GC is configured to properly handle fork().
2. The -S command line option now starts HPC-GAP up in single-threaded
   mode without any other threads running.
3. Normally, signals are handled by a separate thread; this is now being
   done in the function `InstallHPCGAPSignalHandling`. If started with
   -S, this function needs to be called explicitly in order to enable
   proper handling of SIGINT etc.

Thank You!

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1391,9 +1391,19 @@ static Obj FuncBREAKPOINT(Obj self, Obj arg)
 
 static Obj FuncTHREAD_UI(Obj self)
 {
-  return ThreadUI ? True : False;
+    return (ThreadUI && !SingleThreadStartup) ? True : False;
 }
 
+/****************************************************************************
+**
+*F  FuncSINGLE_THREAD_STARTUP . . whether to start up in single-threaded mode
+**
+*/
+
+static Obj FuncSINGLE_THREAD_STARTUP(Obj self)
+{
+    return SingleThreadStartup ? True : False;
+}
 
 #endif
 
@@ -1480,6 +1490,7 @@ static StructGVarFunc GVarFuncs[] = {
     GVAR_FUNC_0ARGS(KERNEL_INFO),
 #ifdef HPCGAP
     GVAR_FUNC_0ARGS(THREAD_UI),
+    GVAR_FUNC_0ARGS(SINGLE_THREAD_STARTUP),
 #endif
     GVAR_FUNC_1ARGS(MASTER_POINTER_NUMBER, ob),
     GVAR_FUNC_1ARGS(BREAKPOINT, integer),
--- a/hpcgap/lib/hpc/thread.g
+++ b/hpcgap/lib/hpc/thread.g
@@ -83,18 +83,26 @@ BindGlobal("NewInterruptID", function()
   od;
 end);
 
-CreateThread(function()
-  local handlers;
-  handlers := rec(
-    SIGINT := DEFAULT_SIGINT_HANDLER,
-    SIGCHLD := DEFAULT_SIGCHLD_HANDLER,
-    SIGVTALRM := DEFAULT_SIGVTALRM_HANDLER,
-    SIGWINCH := DEFAULT_SIGWINCH_HANDLER
-  );
-  while true do
-    SIGWAIT(handlers);
-  od;
+BindGlobal("InstallHPCGAPSignalHandling", function()
+  if not IsBound(SignalHandlerThread) then
+    BindGlobal("SignalHandlerThread", CreateThread(function()
+      local handlers;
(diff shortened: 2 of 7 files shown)


alex-konovalov:
Thanks @rbehrends - this seems to work, we will do some further testing now.

alex-konovalov:
@rbehrends maybe new command line option (`-J` for Jupyter?)

rbehrends:
Something like this is what I'm thinking, except that this is only going to be used programmatically, so we don't have to waste a single letter command line option on it. Something like `--single-thread` or `--server-mode`?

alex-konovalov:
@rbehrends let's spare a single-letter option indeed, but both ` --single-thread` or `--server-mode` could be equally confusing (e.g. understood as ways to disable threading completely, or to run an SCSCP server). What about `--jupyter-mode`? This will also be reusable outside HPC-GAP if we will discover that we need it later e.g. in GAP 4.12.

rbehrends:
The thing is that it's not really "Jupyter mode". What it does is it making sure that GAP starts up with only a single thread running. This is more of a fork-compatible mode. Jupyter is just one application for that.

rbehrends:
I've added a `--single-thread` option and reverted the semantics of `-S`; the name is a placeholder for now, but with the logic in place, it can easily be renamed.

alex-konovalov:
@rbehrends thanks - then we will update our setup for Jupyter notebooks HPC-GAP demo.

src/gap.c
```diff
@@ -1391,9 +1391,19 @@ static Obj FuncBREAKPOINT(Obj self, Obj arg)
 
 static Obj FuncTHREAD_UI(Obj self)
 {
-  return ThreadUI ? True : False;
+  return (ThreadUI && !SingleThreadStartup) ? True : False;
 }
 
+/****************************************************************************
+**
+*F FuncSINGLE_THREAD_STARTUP  ... Whether to start up in single-threaded mode.
```
fingolfin:
```suggestion
*F  FuncSINGLE_THREAD_STARTUP . . whether to start up in single-threaded mode
```

src/gap.c
```diff
@@ -1391,9 +1391,19 @@ static Obj FuncBREAKPOINT(Obj self, Obj arg)
...
+**
+*/
+
+static Obj FuncSINGLE_THREAD_STARTUP(Obj self)
+{
+  return SingleThreadStartup ? True : False;
```
fingolfin:
```suggestion
    return SingleThreadStartup ? True : False;
```

src/gap.c
```diff
@@ -1391,9 +1391,19 @@ static Obj FuncBREAKPOINT(Obj self, Obj arg)
 
 static Obj FuncTHREAD_UI(Obj self)
 {
-  return ThreadUI ? True : False;
+  return (ThreadUI && !SingleThreadStartup) ? True : False;
```
fingolfin:
```suggestion
    return (ThreadUI && !SingleThreadStartup) ? True : False;
```

rbehrends:
@fingolfin See my comment above. I reverted the changes to `-S` in a later commit and introduced a separate command line option. I'll fix the formatting later tonight.

rbehrends:
Yeah, it's WIP because I basically just made up a name for the command line option and wanted to first check that this is fine with everyone.

rbehrends:
Also, another thing with `--single-thread` is that if you want signal handling, you need to do that after startup/forking by calling `InstallHPCGAPSignalHandling()`. This is currently not documented, nor is it clear whether we want that name or a more generic name like `FinishHPCStartup()` if we later need to put additional functionality in there.

alex-konovalov:
@rbehrends shall we merge this?

rbehrends:
If it's okay this way, I'll squash the commits, update the commit message, and then we can, yes.

rbehrends:
Update: I've squashed the commits as noted above, if the tests work, it can be merged.

alex-konovalov:
@rbehrends thanks!

fingolfin:
Shall this be backported to 4.11?

alex-konovalov:
@fingolfin I think so!

fingolfin:
Backported to stable-4.11 in commit ac4b89802

PR #3717, opened 2019-10-28
Revert "buildsys: use the C++ compiler for linking"

fingolfin:
This essentially reverts PR #3652, which was meant to fix issue #3651 for @hulpke which was caused by a bug in GCC 4.4.7.

The problem with that fix is that it causes severe linker issues when using GAP with Julia on Ubuntu 16.04, due to that system using GCC 5 (more specifically any other Linux installation with GCC <= 6.0 as its primary compiler is affected, only GCC >= 6.1 are safe to use). I myself can workaround that there by updating the compiler to GCC 6, but that is really problematic for people who can't modify their systems like that for some reason or another. The only fix I could come up with so far is switching back to linking with the C compiler by default, which *ought* to work as we don't use the standard C++ library.

Of course this would open up issue #3651 again, but I hope that we can workaround it in another way. Specifically, I hope that @hulpke can work around his problems there with this:

    ./configure LIBS=-lstdc++    # ... plus any other configure options you want/need
    make

If @hulpke can confirm that this also fixes the issue for him (with this PR merged, I mean), then I think we can safely go ahead with merging this, and backporting it to stable-4.11

CC @wbhart

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,7 +18,6 @@ environment:
       CYG_ROOT: C:\cygwin64
       PACKAGES: "-P libgmp-devel,zlib-devel,python27,python27-pip"
       CFLAGS: "--coverage"
-      CXXFLAGS: "--coverage"
       LDFLAGS: "--coverage"
 # FIXME: HPC-GAP build on AppVeyor disabled for now, as it crashes
 #    - CYG_ARCH: x86_64
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,6 +18,7 @@ environment:
       CYG_ROOT: C:\cygwin64
       PACKAGES: "-P libgmp-devel,zlib-devel,python27,python27-pip"
       CFLAGS: "--coverage"
+      CXXFLAGS: "--coverage"
       LDFLAGS: "--coverage"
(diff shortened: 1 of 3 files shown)


wbhart:
I don't know what the Cygwin failure is all about. I'm a little surprised we have cygwin anywhere, as I understood Julia didn't support it. But the error looks like a problem with the stack unwinding model?

Could this be related to your change Max?

fingolfin:
@hulpke oh that's great news

@wbhart the failing Cygwin tests on AppVeyor are *not* using Julia.

The error indicates missing symbol `__gxx_personality_seh0` is from libstdc++; so I could fix that by adding `-libstdc++`. But this shouldn't be needed at all (and the 32bit cygwin build indeed doesn't need it, nor does any of our other builds). This might be a compiler bug again...  This symbol specifically is for C++ exception handling, and we don't use exceptions, nor classes, nor anything but C++ templates. And I can't just add libstdc++ back, because that caused the troubles I am trying to avoid with this PR *sigh*

fingolfin:
Backported to `stable-4.11` in 2351a8c7b62607842000d55e56f21b8abe21a319

PR #3721, opened 2019-10-30
Add optimised SmallestMovedPoint implementation

ChrisJefferson:
This just adds a SmallestMovedPoint C implementation, as I had a case where the time taken mattered.

We leave a GAP-level implementation for permutations which aren't PERM2s or PERM4s, as SmallestMovedPoint is used on permutations which aren't internal rep.

Note (if interested) we don't have a GAP-level implementation for LargestMovedPoint, as there isn't a sensible algorithm for arbitrary permutation implementations (as we don't have a "largest candidate" to start from, as we do with internal permutations).

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -1125,6 +1125,40 @@ UInt LargestMovedPointPerm(Obj perm)
         return LargestMovedPointPerm_<UInt4>(perm);
 }
 
+/****************************************************************************
+**
+*F  SmallestMovedPointPerm( <perm> ) smallest point moved by perm
+**
+**  'SmallestMovedPointPerm' implements 'SmallestMovedPoint' for internal
+**  permutations.
+*/
+
+// Import 'Infinity', as a return value for the identity permutation
+static Obj Infinity;
+
+template <typename T>
+static inline Obj SmallestMovedPointPerm_(Obj perm)
+{
+    const T * ptPerm = CONST_ADDR_PERM<T>(perm);
+    UInt      deg = DEG_PERM<T>(perm);
+
+    for (UInt i = 1; i <= deg; i++) {
+        if (ptPerm[i - 1] != i - 1)
+            return INTOBJ_INT(i);
+    }
+    return Infinity;
+}
+
+static Obj SmallestMovedPointPerm(Obj perm)
+{
+    GAP_ASSERT(TNUM_OBJ(perm) == T_PERM2 || TNUM_OBJ(perm) == T_PERM4);
+
+    if (TNUM_OBJ(perm) == T_PERM2)
+        return SmallestMovedPointPerm_<UInt2>(perm);
+    else
+        return SmallestMovedPointPerm_<UInt4>(perm);
+}
+
 
 /****************************************************************************
 **
@@ -1141,6 +1175,18 @@ static Obj FuncLARGEST_MOVED_POINT_PERM(Obj self, Obj perm)
     return INTOBJ_INT(LargestMovedPointPerm(perm));
 }
 
+/****************************************************************************
+**
+*F  FuncSMALLEST_MOVED_POINT_PERM( <self>, <perm> )
+**
+**  GAP-level wrapper for 'SmallestMovedPointPerm'.
+*/
+static Obj FuncSMALLEST_MOVED_POINT_PERM(Obj self, Obj perm)
+{
+    RequirePermutation("SmallestMovedPointPerm", perm);
+
+    return SmallestMovedPointPerm(perm);
+}
 
 /****************************************************************************
 **
@@ -2736,6 +2782,7 @@ static StructGVarFunc GVarFuncs [] = {
 
     GVAR_FUNC_1ARGS(PermList, list),
     GVAR_FUNC_1ARGS(LARGEST_MOVED_POINT_PERM, perm),
+    GVAR_FUNC_1ARGS(SMALLEST_MOVED_POINT_PERM, perm),
     GVAR_FUNC_2ARGS(CYCLE_LENGTH_PERM_INT, perm, point),
     GVAR_FUNC_2ARGS(CYCLE_PERM_INT, perm, point),
     GVAR_FUNC_1ARGS(CYCLE_STRUCT_PERM, perm),
@@ -2783,6 +2830,9 @@ static Int InitKernel (
 #endif
 
     ImportGVarFromLibrary("PERM_INVERSE_THRESHOLD", &PERM_INVERSE_THRESHOLD);
+
+    /* needed for SmallestMovedPoint                                       */
+    ImportGVarFromLibrary("infinity", &Infinity);
 
     /* install the type functions                                           */
     ImportGVarFromLibrary( "TYPE_PERM2", &TYPE_PERM2 );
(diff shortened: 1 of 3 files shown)


src/permutat.cc
```diff
@@ -1125,6 +1125,40 @@ UInt LargestMovedPointPerm(Obj perm)
         return LargestMovedPointPerm_<UInt4>(perm);
 }
 
+/****************************************************************************
+**
+*F  SmallestMovedPointPerm( <perm> ) smallest point moved by perm
+**
+**  'SmallestMovedPointPerm' returns  the  smallest  positive  integer that is
```
wucas: But `Infinity` is not an integer. Shouldn't it say something like: `'SmallestMovedPointPerm' returns  the  smallest  positive  integer that is moved by the permutation <perm> or Infinity if <perm> is the identity.`

src/permutat.cc
```diff
@@ -1125,6 +1125,40 @@ UInt LargestMovedPointPerm(Obj perm)
...
+**  moved by the permutation <perm>.
+*/
+
+// Import 'Infinity', as return value for identity permutation
```
wucas: as the return value for the/an identity pmerutation/permutations (not sure how one would say it in english)

tst/testinstall/perm.tst
```diff
@@ -314,12 +314,38 @@ gap> PermList(Concatenation([1..70000],[70001,1,70003]));
...
 
 #
-# LARGEST_MOVED_POINT_PERM
-#
-gap> LARGEST_MOVED_POINT_PERM((2,3));
-3
-gap> LARGEST_MOVED_POINT_PERM((2,70000));
-70000
+# SmallestMovedPoint, LargestMovedPoint, MovedPoints, NrMovedPoints
+#
```
wucas: Not to say that this makes sense but in the spirit of testing also a bit more unusual inputs, why not add a test for a permutation like (5,7,1) or just (10,1) that is not already sorted?

src/permutat.cc
fingolfin: Indeed. Could also take the text used in the manual for SmallestMovedPoint. Or, for that matter, just say
> This function implements the operation `SmallestMovedPoint` for internal permutations.

Also, don't use those double spaces, please :-)

src/permutat.cc
```diff
@@ -1141,6 +1175,21 @@ static Obj FuncLARGEST_MOVED_POINT_PERM(Obj self, Obj perm)
...
+/****************************************************************************
+**
+*F  FuncSMALLEST_MOVED_POINT_PERM( <self>, <perm> )
+*F  Smallest point moved by perm
+**
+**  GAP-level wrapper for 'SmallestMovedPointPerm'.
+*/
+static Obj FuncSMALLEST_MOVED_POINT_PERM(Obj self, Obj perm)
+{
+
```
```suggestion
```

src/permutat.cc
```diff
@@ -1141,6 +1175,21 @@ static Obj FuncLARGEST_MOVED_POINT_PERM(Obj self, Obj perm)
     return INTOBJ_INT(LargestMovedPointPerm(perm));
 }
 
+/****************************************************************************
+**
+*F  FuncSMALLEST_MOVED_POINT_PERM( <self>, <perm> )
+*F  Smallest point moved by perm
```
fingolfin: The `*F` is only for the function name, not for description.
```suggestion
**
**  Return the smallest point moved by <perm>, or Infinity.
```

src/permutat.cc
```diff
@@ -1141,6 +1175,21 @@ static Obj FuncLARGEST_MOVED_POINT_PERM(Obj self, Obj perm)
...
+**
+**  GAP-level wrapper for 'SmallestMovedPointPerm'.
```
fingolfin: I don't think this adds any useful information, as the 2 line implementation makes this info clear anyway. So I'd just remove it.
```suggestion
```

tst/testinstall/perm.tst
```diff
@@ -314,12 +314,38 @@ gap> PermList(Concatenation([1..70000],[70001,1,70003]));
 fail
 
 #
-# LARGEST_MOVED_POINT_PERM
-#
-gap> LARGEST_MOVED_POINT_PERM((2,3));
-3
-gap> LARGEST_MOVED_POINT_PERM((2,70000));
-70000
```
fingolfin: The reason for these tests was to make 100% sure we test those kernel function, and not some other installed methods.

I guess these days, I'd move those tests to `tst/testinstall/kernel/permutat.tst`

tst/testinstall/perm.tst
fingolfin: That won't make a difference, because the differences vanishes the moment the permutation object is created, i.e., well before the permutation is passed to those functions.

ChrisJefferson:
Thank you both for your comments.they mostly look accurate but I may not be able to address them until Sunday.

ChrisJefferson:
Yes, sorry, I'm trying to do no non-required programming until a paper gets finished :)

src/permutat.cc
```diff
@@ -1125,6 +1125,40 @@ UInt LargestMovedPointPerm(Obj perm)
...
+    UInt      deg = DEG_PERM<T>(perm);
+
+    for (UInt i = 1; i <= deg; i++) {
+        if (ptPerm[i - 1] != i - 1)
+            return INTOBJ_INT(i);
+    }
+    return Infinity;
+}
+
+Obj SmallestMovedPointPerm(Obj perm)
```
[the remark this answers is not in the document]
ChrisJefferson: While I'm fixing various issues, many of the issues (including this one) are "I mirrored LargestMovedPointPerm".

src/permutat.cc
```diff
@@ -1125,6 +1125,40 @@ UInt LargestMovedPointPerm(Obj perm)
...
+static Obj Infinity;
+
+template <typename T>
+static inline Obj SmallestMovedPointPerm_(Obj perm)
+{
+    const T * ptPerm = CONST_ADDR_PERM<T>(perm);
+    UInt      deg = DEG_PERM<T>(perm);
+
+    for (UInt i = 1; i <= deg; i++) {
+        if (ptPerm[i - 1] != i - 1)
```
wilfwilson: Why don't you instead do the loop from `i = 0; i < deg` so that you don't have to do `i - 1` twice on this line for every iteration of the loop? Then on the following line you'd just do
```
return INTOBJ_INT(i + 1);
```

ChrisJefferson: The boring answer to your question is "LargestMovedPointPerm" does that, and I kept the two as close as possible, as any differences might raise questions of "why are they different".

Of course, the other option is to clean them both up (which I think I'll do, in another commit).

ChrisJefferson: Actually, looking through, basically everywhere in permutat.cc uses this 'ptr[i-1] = i-1' style, so it would seem sensible to clean them all up, instead of just one, and I'd do that in a future PR.

ChrisJefferson:
I'll add a test like #3738 which tests SMALLEST_MOVED_POINT_PERM once that PR is merged (I want that one to PR to be merged in first, so it cleanly merges back to 4.11).

src/permutat.cc
wilfwilson: Happy for that to be done separately. I made #3739 to remind you/us to do it.

PR #3725, opened 2019-11-04
Fix power maps in result of CharacterTableIsoclinic

ThomasBreuer:
The following bug gets fixed that had been introduced in commit 2013da192c1b6bea588e1803a67e72bda1db6c22:
The result of `CharacterTableIsoclinic` in the case of a central subgroup of order four (a rare situation ...) can store wrong power maps.

This bug was not present in released GAP versions (up to 4.10.2);
if the fix makes it into GAP 4.11 then it need not be mentioned in release notes.

### DIFF
--- a/lib/ctbl.gi
+++ b/lib/ctbl.gi
@@ -6226,7 +6226,7 @@ InstallMethod( CharacterTableIsoclinic,
 
         for i in [ 1 .. p-1 ] do
           # Deal with the classes in the 'i'-th coset.
-          ypos:= PowerMap( tbl, ( k * QuoInt( i * q, p ) ) mod p, xpos );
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
@fingolfin Thanks for the remark about the label.
Concerning a test that shows the bug, I had added one to `tst/teststandard/ctblisoc.tst`; is this not enough?

fingolfin:
@ThomasBreuer it wasn't clear to me that this used to trigger the bug before :-). Thanks for clarifying.

fingolfin:
Backported to `stable-4.11` in 3f4378936549dbc7be9993c6fa468ecdb4932926

PR #3727, opened 2019-11-06
ENHANCE: Performance improvements for certain cases in 2-cohomology

hulpke:
These changes add a number of performance improvements for the 2-cohomology, to avoid certain cases of unreasonably bad performance:

- Use a special (Coxeter) rewriting system for symmetric group (not one built as A_n.2)
- Ensure special methods for simple groups are used when computing a presentation
- Avoid unneeded (and complicating) use of Tietze trnsformations when using rewriting presentation.

They still should go into 4.11 and will not have impact outside the new 2-cohomology code.

The changes in twocohom.gi look larger as they are due to whitespace/ indentation when moving existing code into an `if` clause.

### DIFF
--- a/lib/ghomperm.gi
+++ b/lib/ghomperm.gi
@@ -266,7 +266,7 @@ InstallGlobalFunction( RelatorsPermGroupHom, function ( hom, gensG )
           index, inv0, iso, j, map, ndefs, next, ngens, ngens2, ni, orbit,
           order, P, perm, perms, range, regular, rel, rel2, rels, relsG,
           relsGen, relsH, relsP, S, sizeS, stabG, stabS, table, tail, tail1,
-          tail2, tietze, tzword, undefined,
+          tail2, tietze, tzword, undefined,w,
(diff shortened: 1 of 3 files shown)


lib/gpfpiso.gi
```diff
@@ -320,6 +320,7 @@ function(g,str,N)
       od;
 
       # we know sf is simple
+      IsSimpleGroup(sf);
```
fingolfin: Hmm, what's the point of this? Doesn't this even slow down the code in some cases, because it might end up performing a simplicity check?
Also note that `IsNonabelianSimpleGroup` already implies `IsSimpleGroup`, so the next line would set the filter, too.

hulpke: Thanks. Indeed redundant now (and I think the result of merging older changes in, after other things happened. Will remove

fingolfin:
Could you please rebase, to fix the CI failures (as explained on the other PR).

fingolfin:
Backported to stable-4.11 in commits 5bfdc14d9 and b4e888962

PR #3728, opened 2019-11-06
Performance improvements to normal subgroups

hulpke:
Performance improvement if sub-case is elementary abelian.

Should be able to go into 4.11 without issue, but not the end of the world if not.

### DIFF
--- a/lib/grplatt.gi
+++ b/lib/grplatt.gi
@@ -1900,7 +1900,7 @@ local G,	# group
 	  Info(InfoLattice,2,"factorsize=",Index(j,N),"/",Index(M,N));
 
 	  # reasons not to go complements
-	  if (HasAbelianFactorGroup(j,N) and
+	  if (HasElementaryAbelianFactorGroup(j,N) and
 	    p^(Length(mpcgs)*LogInt(Index(j,M),p))>100)
 	    then
             Info(InfoLattice,3,"Set l to fail");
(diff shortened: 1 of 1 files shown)


lib/grplatt.gi
```diff
@@ -2068,6 +2068,23 @@ local G,	# group
...
 	      # all groups obtained will have a solvable factor
 	      l:=[];
+            elif HasElementaryAbelianFactorGroup(j,N) then
+              #Error("invar");
+              r:=ModuloPcgs(j,N);
+              jg:=RelativeOrders(r)[1];
+              l:=MTX.BasesSubmodules(GModuleByMats(LinearActionLayer(G,r),
+                GF(jg)));
+	      Info(InfoLattice,2,"found ",Length(l)," submodules");
+	      idx:=LogInt(Index(j,M),jg);
```
fingolfin: Something is funny about the indentation here; tabs vs. spaces? Also again in lines 2085-2087

lib/grplatt.gi
```diff
@@ -1900,7 +1900,7 @@ local G,	# group
 	  Info(InfoLattice,2,"factorsize=",Index(j,N),"/",Index(M,N));
 
 	  # reasons not to go complements
-	  if (HasAbelianFactorGroup(j,N) and
+	  if (HasElementaryAbelianFactorGroup(j,N) and
```
fingolfin: I think I understand the new code below; but I don't understand what the change in this line is about. Could you briefly explain?

lib/grplatt.gi
```diff
@@ -2068,6 +2068,23 @@ local G,	# group
...
+	      idx:=LogInt(Index(j,M),jg);
+              C:=List(GeneratorsOfGroup(M),x->ExponentsOfPcElement(r,x))*Z(jg)^0;
+              C:=Filtered(TriangulizedMat(C),x->not IsZero(x));
+              l:=Filtered(l,x->Length(x)=idx 
+                and RankMat(Concatenation(x,C))=Length(r));
+              l:=List(l,x->ClosureGroup(N,List(x,
+                y->PcElementByExponents(r,y))));
+	      l:=Filtered(l,i->IsNormal(G,i));
+	      Info(InfoLattice,1,Length(l)," of these normal");
+	      nnt:=Concatenation(nnt,l);
```
```suggestion
	      Append(nnt,l);
```

lib/grplatt.gi
hulpke: It probably still is tabs (of which the file has many), while the new line only includes spaces. To fix this I think I would have to replace tabs by spaces overall (which will make the diff look large but otherwise is harmless)?

lib/grplatt.gi
hulpke: The code (at some point) calculates complements. If this is in an abelian group it can be hard (which is why the code used to delegate this to `SubgroupsSolvableGroup`, finding invariant subgroups. In the meantime the code got changed to use action on the cohomology, so the old diversion to subgroups code is not worth doing any longer (generically), but only the special case diversion for the case of modules. (I.e. this change should have happened long ago, but I did not notice the performance difference.)

PR #3731, opened 2019-11-07
lib: make GroupByQuotientSystem more robust

fingolfin:
This fixes a bug in HPC-GAP, where this command triggered a
guard error:

    AutomorphismGroup(SmallGroup(864, 603));

This is a part of PR #3502 which is ready for merging.


fingolfin:
Backported to stable-4.11 in commit 879238707

PR #3732, opened 2019-11-07
HPC-GAP: Fix CloseInput() in threads without stdin

fingolfin:
This fixes a bug where a task (which has no stdin) tries to use Read():
This works, until the point where GAP tries to close the file it just
read, at which point the check modified in this patch triggered an
error.

This commit is originally by @rbehrends and taken from PR #3502, but since it is of independent value, I figured we could merge it right now, and perhaps even backport it.


fingolfin:
Backported to `stable-4.11` in 202a0fa54

PR #3733, opened 2019-11-08
fixed ConstituentsOfCharacter for Brauer character

ThomasBreuer:
# Description
The two bugs mentioned in the text for release notes (see below) are now fixed,
some tests for the variants of arguments are available,
and the documentation states which cases are admissible:
ordinary or modular characters,
ordinary or modular virtual characters.

## Text for release notes 
The result of <code class="code">ConstituentsOfCharacter</code>, when called with a Brauer character as its only argument, was not reliable.
(This bug has been reported by Gabriel Navarro.)
Also, calling <code class="code">ConstituentsOfCharacter</code> with a Brauer character table and a virtual Brauer character caused error messages.


fingolfin:
Backported to stable-4.11 in commit d394a71f7

PR #3734, opened 2019-11-11
WIP: Making HPC-GAP work with Julia

rbehrends:
This is still very raw and has a number of hacks (such as using `calloc()` for `AllocateMemoryBlock()`). The PR is currently an FYI/progress tracker, as a lot of the implementation is still subject to change. It also relies on previous HPC-GAP related PRs (namely https://github.com/gap-system/gap/pull/3502 and https://github.com/gap-system/gap/pull/2845).

Aside from making HPC-GAP work with Julia, it also aims at allowing HPC-GAP to be used as a library (which is necessary for GAP.jl, but also allows other code to link to it).

It currently runs only with `gap --single-thread`, as `CreateThread()` is disabled due to limitations in Julia. (It's also not a priority, as we are currently more concerned with calling GAP from Julia rather than having yet another GC for GAP.)

Configuration needs to happen with:

```
./configure --with-gc=julia --with-julia=/path/to/julia --enable-hpcgap --enable-native-tls
```

Remaining issues that I am working on are:

* [ ] Proper initialization of thread-local state. This involves refactoring of `RunThread()` and `DispatchThread()`.
* [ ] Thread creation/initialization for the libgap version, building on the previous bullet point. This is necessary for GAP.jl.
* [ ] Fixing errors in root scanning. Currently, we're seemingly missing out on some global variables, leading to eventual crashes.
* [ ] Clean implementation of `AllocateMemoryBlock()`; this may require an internal API change for that function to permit a type parameter. `AllocateMemoryBlock()` presumes that a conservative GC is being used, which may not be the best decision, anyway.
* [ ] Make `CreateThread()` work via the task system (Julia only recognizes its own preallocated threads, all other thread activity needs to be mapped to Julia tasks that are then being run by threads. This is how the Julia macro `@threads` works, for example.).
* [ ] Insert GC safepoints around I/O (the Julia GC requires that for multi-threaded GC to work). This is necessary primarily for console I/O, which can block indefinitely while computations are running (I am currently not so much worried about a pause while reading a file, as GAP tends to read files only once, so a pause is more acceptable).

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -64,9 +64,7 @@ SOURCES += src/intfuncs.c
 SOURCES += src/intrprtr.c
 SOURCES += src/io.c
 SOURCES += src/iostream.c
-ifeq ($(HPCGAP),no)             # we don't support a kernel API in HPC-GAP atm
 SOURCES += src/libgap-api.c
-endif
 SOURCES += src/listfunc.c
 SOURCES += src/listoper.c
 SOURCES += src/lists.c
--- a/configure.ac
+++ b/configure.ac
@@ -221,6 +221,13 @@ AC_MSG_CHECKING([whether to use guards for race protection in HPC-GAP])
 AC_MSG_RESULT([$enable_guards])
 
 dnl
+dnl Check: --enable-guards requires --enable-hpcgap
+dnl
+
+AS_IF([[test "x$enable_guards" == "xyes" && test "x$enable_hpcgap" == "xno"]],
+  [AC_MSG_ERROR([Option --enable-guards requires --enable-hpcgap])])
+
+dnl
 dnl User setting: native thread-local storage (off by default)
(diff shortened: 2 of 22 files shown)


fingolfin:
The last commit message is "Make HPC-GAP guards work without Ward" -- but HPC-GAP already works without ward, and has done so for months. Perhaps "Change HPC-GAP to use the unward tool" or something like that could be used instead.

rbehrends:
I think a bit of subtlety got lost here. I had changed the message to "Make HPC-GAP *guards* work without Ward" (emphasis on *guards*); I had inserted "guards" as a result of your earlier suggestion to change the commit message. HPC-GAP did work without Ward, guards didn't.

If you think that this still doesn't describe the commit well, I think adding a longer explanation would be better than trying to squeeze an adequate description into a single line.

PR #3737, opened 2019-11-12
Computation of Omega groups of arbitrary p groups is supported.

DominikBernhardt:
The manual claims GAP can only compute Omega-groups of abelian p groups. In my opinion, this is not true, see `lib/grppc.gi`. This PR removes the corresponding line in the manual.
Also, a part of the comment in the code was redundant: Any p-group with nilpotency class at most `p` is regular, so I clarified this part in the documentation.

I put Release notes: To be added. Feel free to change this if you want to.

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -1972,7 +1972,6 @@ DeclareAttribute( "NrConjugacyClasses", IsGroup );
 ##  \{ g \in <A>G</A> \mid g^{{<A>p</A>^{<A>n</A>}}} = 1 \}</M>.
 ##  The default value for <A>n</A> is <C>1</C>.
 ##  <P/>
-##  <E>@At the moment methods exist only for abelian <A>G</A> and <A>n</A>=1.@</E>
 ##  <Example><![CDATA[
 ##  gap> h:=SmallGroup(16,10);
 ##  <pc group of size 16 with 4 generators>
--- a/tst/testinstall/grppc.tst
+++ b/tst/testinstall/grppc.tst
@@ -166,6 +166,11 @@ gap> RepresentativeAction(G,x,x^2)<>fail;
 true
 gap> RepresentativeAction(G,x,x^2);
 f1*f2
+gap> g := SmallGroup(243,27);;
+gap> AsSet(Omega(g,3,1)) = Set(Filtered(g, g -> IsOne(g^3)));
+true
+gap> AsSet(Omega(g,3,2)) = Set(Filtered(g, g -> IsOne(g^9)));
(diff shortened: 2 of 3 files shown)


lib/grp.gd
```diff
@@ -1972,7 +1972,6 @@ DeclareAttribute( "NrConjugacyClasses", IsGroup );
 ##  \{ g \in <A>G</A> \mid g^{{<A>p</A>^{<A>n</A>}}} = 1 \}</M>.
 ##  The default value for <A>n</A> is <C>1</C>.
 ##  <P/>
-##  <E>@At the moment methods exist only for abelian <A>G</A> and <A>n</A>=1.@</E>
```
fingolfin: OK, I tried this and it seemed to work:
```
gap> grps:=AllSmallGroups(32);;
gap> Collected(List(grps,g->Size(Omega(g,2))));
[ [ 2, 2 ], [ 4, 14 ], [ 8, 10 ], [ 16, 13 ], [ 32, 12 ] ]
gap> Collected(List(grps,g->Size(Omega(g,2,2))));
[ [ 4, 1 ], [ 8, 3 ], [ 16, 9 ], [ 32, 38 ] ]
gap> Collected(List(grps,g->Size(Omega(g,2,3))));
[ [ 8, 1 ], [ 16, 2 ], [ 32, 48 ] ]
```
So it seems you are right. But perhaps add a minimal test to verify this? Anyway, that's optional.

fingolfin:
I don't think this needs to go into the release notes; it just removes an incorrect restriction in the manual, which did log a bit odd to start with. The code itself has been in there (apparently written by Jack Schmidt and added in by @hulpke )

DominikBernhardt:
@fingolfin Added one test.

tst/testinstall/grppc.tst
```diff
@@ -166,6 +166,12 @@ gap> RepresentativeAction(G,x,x^2)<>fail;
 true
 gap> RepresentativeAction(G,x,x^2);
 f1*f2
+gap> g := SmallGroup(243,27);
+<pc group of size 243 with 5 generators>
+gap> Omega(g,3,1);
```
fingolfin: The way the output is printed could change, and it's not even clear to me that it is correct now. How about this instead (and similar below):
```suggestion
gap> AsSet(Omega(g,3,1)) = Set(Filtered(g, g -> IsOne(g^3)));
true
```

DominikBernhardt: I was just stupid there, sorry. I changed the test.

PR #3738, opened 2019-11-12
Fix bug in CycleStructurePerm for a single cycle of length 2^16

ChrisJefferson:
CYCLE_STRUCT_PERM overflowed a UInt2 when given a cycle of length
exactly 2^16, so store the size-1 (which is what we output anyway)

This caused both wrong answers, and memory corruption.

This function could, in principle, be overhauled/simplified to use less fancy buffers, but I'm convinced by this fix and wanted something small to go into 4.11 at least.

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -1372,14 +1372,14 @@ static inline Obj CYCLE_STRUCT_PERM(Obj perm)
     max = 0;
     for (pnt = 0; pnt < deg; pnt++) {
         if (clr[pnt] == 0) {
-            len = 1;
+            len = 0;
             clr[pnt] = 1;
             for (p = ptPerm[pnt]; p != pnt; p = ptPerm[p]) {
                 clr[p] = 1;
                 len++;
             }
 
-            if (len > 1) {
+            if (len > 0) {
                 offset[cnt] = (T)len;
                 cnt++;
                 if (len > max) {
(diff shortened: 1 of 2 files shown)


src/permutat.cc
```diff
@@ -1346,27 +1346,21 @@ static inline Obj CYCLE_STRUCT_PERM(Obj perm)
     ende = cnt;
 
     /* create the list */
-    list = NEW_PLIST(T_PLIST, max - 1);
-    SET_LEN_PLIST(list, max - 1);
+    list = NEW_PLIST(T_PLIST, max);
+    SET_LEN_PLIST(list, max);
```
fingolfin: Until here, it seems you just changed `len` and `max` to be 1 smaller than they were before, yes? (This is of course fine, I am just trying to verify I understand your changes correctly)

src/permutat.cc
```diff
@@ -1346,27 +1346,21 @@ static inline Obj CYCLE_STRUCT_PERM(Obj perm)
...
     for (cnt = 0; cnt < ende; cnt++) {
         pnt = (UInt)offset[cnt];
-        pnt--;
-        ptList[pnt] = (Obj)((UInt)ptList[pnt] + 1);
-    }
-
-    for (pnt = 1; pnt < max; pnt++) {
-        if (ptList[pnt] != 0) {
-            ptList[pnt] = INTOBJ_INT((UInt)ptList[pnt]);
+        if(ptList[pnt] == 0) {
```
```suggestion
        if (ptList[pnt] == 0) {
```

src/permutat.cc
```diff
@@ -1346,27 +1346,21 @@ static inline Obj CYCLE_STRUCT_PERM(Obj perm)
...
+    SET_LEN_PLIST(list, max);
     ptList = ADDR_OBJ(list);
 
     /* Recalculate after possible GC */
     scratch = ADDR_TMP_PERM<T>();
     offset = (T *)((UInt)scratch + (bytes));
 
-    for (pnt = 1; pnt < max; pnt++) {
-        ptList[pnt] = 0;
-    } /* clean out */
```
fingolfin: We zero initialize new objects, and `list` is new, so this is redundant, OK.

tst/testinstall/kernel/permutat.tst
```diff
@@ -0,0 +1,103 @@
...
+gap> permprops(PermList(Concatenation([2^16-1], [1..2^16-2])));
+LARGEST_MOVED_POINT_PERM:65535
+ORDER_PERM:65535
+SIGN_PERM:1
+CYCLE_SHORT_PERM (short output): [ [ 65534, 1 ] ]
+gap> permprops(PermList(Concatenation([2^16-2], [1..2^16-3])));
+LARGEST_MOVED_POINT_PERM:65534
+ORDER_PERM:65534
+SIGN_PERM:-1
+CYCLE_SHORT_PERM (short output): [ [ 65533, 1 ] ]
```
fingolfin: How about having at least one test for a permutation with more than one cycle?

src/permutat.cc
ChrisJefferson: One day I'll remember to run clang-format

tst/testinstall/kernel/permutat.tst
ChrisJefferson: Some more tests added

ChrisJefferson:
I believe all issues are now fixed

fingolfin:
Backported to stable-4.11 in commit 7d39a545f

PR #3740, opened 2019-11-12
Speed up printing of large permutations

ChrisJefferson:
GAP is very slow at printing large permutations (say on the order of 2^20), it can take several minutes. This is because it uses a silly O(n^2) algorithm.

This improves the two places where we print permutations to instead use an O(n) algorithm.

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -163,10 +163,8 @@ static Obj InvPerm(Obj perm);
 **  uses the degree to print all points with same width, which  looks  nicer.
 **  Linebreaks are prefered most after cycles and  next  most  after  commas.
 **
-**  It does not remember which points have already  been  printed.  To  avoid
-**  printing a cycle twice each is printed with the smallest  element  first.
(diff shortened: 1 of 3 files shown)


src/permutat.cc
```diff
@@ -163,10 +163,8 @@ static Obj InvPerm(Obj perm);
 **  uses the degree to print all points with same width, which  looks  nicer.
 **  Linebreaks are prefered most after cycles and  next  most  after  commas.
 **
-**  It does not remember which points have already  been  printed.  To  avoid
-**  printing a cycle twice each is printed with the smallest  element  first.
-**  This may in the worst case, for (1,2,..,n), take n^2/2 steps, but is fast
-**  enough to keep a terminal at 9600 baud busy for all but the extrem cases.
```
fingolfin: lol

fingolfin:
Oops, needs to be rebased now

ChrisJefferson:
Hopefully all issues fixed (once the tests cycle)

PR #3741, opened 2019-11-13
A bunch of fixes and tests for `src/hpc/threadapi.c`

fingolfin:
These tests are mostly not great, they just call functions with silly arguments to trigger errors. But that's enough to get some extra coverage (gameification, yay), and already revealed several crashes, deadlocks etc. 

I'd like to get the coverage report to see if I missed some "obvious" stuff.

### DIFF
--- a/src/hpc/threadapi.c
+++ b/src/hpc/threadapi.c
@@ -2048,11 +2048,11 @@ static Obj FuncFORCE_MAKE_PUBLIC(Obj self, Obj obj)
 
 static Obj FuncSHARE_NORECURSE(Obj self, Obj obj, Obj name, Obj prec)
 {
-    Region * region = NewRegion();
     if (name != Fail && !IsStringConv(name))
         return ArgumentError(
             "SHARE_NORECURSE: Second argument must be a string or fail");
     Int p = GetSmallInt("SHARE_NORECURSE", prec);
+    Region * region = NewRegion();
     region->prec = p;
     if (!MigrateObjects(1, &obj, region, 0))
         return ArgumentError("SHARE_NORECURSE: Thread does not have "
@@ -2106,11 +2106,11 @@ static Obj FuncCLONE_DELIMITED(Obj self, Obj obj)
 
 static Obj FuncNEW_REGION(Obj self, Obj name, Obj prec)
 {
-    Region * region = NewRegion();
     if (name != Fail && !IsStringConv(name))
         return ArgumentError(
             "NEW_REGION: Second argument must be a string or fail");
     Int p = GetSmallInt("NEW_REGION", prec);
+    Region * region = NewRegion();
     region->prec = p;
     if (name != Fail)
         SetRegionName(region, name);
@@ -2125,14 +2125,14 @@ static Obj FuncREGION_PRECEDENCE(Obj self, Obj regobj)
 
 static Obj FuncSHARE(Obj self, Obj obj, Obj name, Obj prec)
 {
-    Region * region = NewRegion();
-    Obj      reachable;
     if (name != Fail && !IsStringConv(name))
         return ArgumentError(
             "SHARE: Second argument must be a string or fail");
     Int p = GetSmallInt("SHARE", prec);
+    Region * region = NewRegion();
     region->prec = p;
-    reachable = ReachableObjectsFrom(obj);
+
+    Obj reachable = ReachableObjectsFrom(obj);
     if (!MigrateObjects(LEN_PLIST(reachable), ADDR_OBJ(reachable) + 1, region,
                         1))
         return ArgumentError(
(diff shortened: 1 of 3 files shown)


src/hpc/threadapi.c
```diff
@@ -374,13 +374,9 @@ static int GetThreadID(const char * funcname, Obj thread)
...
-
-
-#define AS_STRING(s) #s
-    const char *msg = "must be thread object or an integer between 0 and "
-                AS_STRING(MAX_THREADS-1);
-#undef AS_STRING
-    RequireArgumentEx(funcname, thread, NICE_ARGNAME(thread), msg);
+    RequireArgumentEx(funcname, thread, NICE_ARGNAME(thread),
+                      "must be a thread object or an integer between 0 and "
+                      "MAX_THREADS - 1");
```
ChrisJefferson: Is MAX_THREADS a gap level variable people can query? If not, might be nice to put something here people can check.

rbehrends: It is currently only defined in the kernel, but could be exposed easily if needed:

```
Obj FuncMAX_THREADS()
{
    return INTOBJ_INT(MAX_THREADS);
}
```

fingolfin: I added `ExportAsConstantGVar(MAX_THREADS);` now, as `MAX_THREADS` is a compile time constant. That said, I am not sure putting it into the global namespace is a good idea? One alternative would be to put into `GAPInfo.KernelInfo`.

BTW, note before my change, the error message also said `MAX_THREADS`, i.e., it never printed a number (it was meant to do so, but the code was buggy and untested. Oh, and of course it was written by me *sigh*).

tst/testinstall/kernel/hpc/threadapi.tst
```diff
@@ -0,0 +1,397 @@
...
+
+#
+gap> StartBarrier(b, 2);
+gap> b;
+<barrier with 0 of 2 threads arrived>
+gap> tmp:=false;;
+gap> t:=CreateThread(function() WaitBarrier(b); repeat MicroSleep(10); until tmp; end);;
+gap> MicroSleep(50);
+gap> b;
+<barrier with 1 of 2 threads arrived>
```
fingolfin: @rbehrends Either I am misunderstanding something (which wouldn't be a surprise, given that barriers in HPC-GAP seem to be completely undocumented), or there is a bug: but it seems that sometimes, the first thread created above does not "arrive" so we get "0 of 2" printed above. Even if I change the `MicroSleep` above to wait for a full second.

tst/testinstall/kernel/hpc/threadapi.tst
```diff
@@ -0,0 +1,397 @@
...
+nd MAX_THREADS - 1 (not the value 'fail')
+gap> ResumeThread(fail);
+Error, ResumeThread: <thread> must be a thread object or an integer between 0 \
+and MAX_THREADS - 1 (not the value 'fail')
+
+#
+gap> t := CreateThread(function() MicroSleep(100); end);;
+gap> PauseThread(t);
+gap> ResumeThread(t);
+gap> WaitThread(t); # FIXME: sometimes deadlocks
```
fingolfin: @rbehrends it seems `WaitThread` sometimes (???) ends up blocking the calling thread if the thread `t` has terminated before. Is that right and intentional? If so, how can one use `WaitThread` safely and without a race condition? Actually, assuming that thread ids are reused, I don't see how one can use this safely at all unless every thread one wants to wait for never "just exits", but instead uses careful control logic (e.g. a semaphore or so) to block its exit until it receives "permission to die" from its controller. Whereas for pthreads, to my understanding this problem does not exist, because each `pthread_t` is unique and cannot be reused until explicitly reclaimed via `pthread_destroy`. Well, there is also `pthread_detach` but HPC-GAP doesn't allow calling that.

Am I missing something? What?

The one way to solve this I can think of is to *require* use of thread objects (drop support for specifying threads via a raw thread ID, which IMHO is problematic anyway). Then, `FuncWaitThread` could pass a given `T_THREAD` object `obj` to `JoinThread`, which then could validate `obj` by comparing it against `thread_data[ThreadID(obj)].thread_object`.

rbehrends: There shouldn't be a problem here. Note that while there may be multiple thread objects that share the same thread ID, at most one will not have `THREAD_JOINED` set. This means that `JoinThread()` will never be called more than once for the thread associated with a given thread object.

tst/testinstall/kernel/hpc/threadapi.tst
rbehrends: HPC-GAP barrier semantics follow the semantics POSIX barriers fairly precisely. I.e. after calling `StartBarrier(b, n)`, `WaitBarrier(b)` will block until it has been called exactly `n` times on `b` (this also necessitates that there will be `n` distinct threads doing that).

There is an issue with printing barrier objects in that the numbers may be wrong. I've added a `hpcgap-barriers` branch on https://github.com/rbehrends/gap that should fix the printing issue.

Changing the second `MicroSleep()` above to be longer should properly report a change in state, though (I have also not been able to reproduce the issue doing that). Unless you are talking about the `MicroSleep()` inside the thread function, which should typically have no effect on the barrier state at that point.

tst/testinstall/kernel/hpc/threadapi.tst
rbehrends: I'll add that I am wondering if it's actually the `PauseThread()` and `ResumeThread()` that directly or indirectly cause `WaitThread()` to block. They do manipulate the exec table in `SetInterrupt()` in particular, and a bug there could disrupt subsequent thread execution.

PR #3743, opened 2019-11-14
Avoid call of RunImmediateMethods if IGNORE_IMMEDIATE_METHODS = false

zickgraf:
# Description

Analogous to https://github.com/gap-system/gap/blob/3111e67d9e2425a1f886d4b4ba15d20d932211d8/lib/type1.g#L663

This saves some 100ms when creating a really large number of objects.

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/lib/type1.g
+++ b/lib/type1.g
@@ -613,7 +613,8 @@ BIND_GLOBAL( "Objectify", function ( type, obj )
     elif IS_REC( obj )  then
         SET_TYPE_COMOBJ( obj, type );
     fi;
-    if not IsNoImmediateMethodsObject(obj) then
+    if not ( IGNORE_IMMEDIATE_METHODS
+             or IsNoImmediateMethodsObject(obj) ) then
       RunImmediateMethods( obj, type![POS_FLAGS_TYPE] );
     fi;
     if IsHPCGAP then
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
Hi, if you run `make docomp`, then commit the result, it will remake some files which will then let more tests pass (I guess when you run GAP it tells you some CRCs are wrong?).

We don't do this automatically (sorry!), as running `make docomp` requires a working GAP, and it generates C code so if it goes wrong it can break your build.

lib/type1.g
```diff
@@ -613,7 +613,8 @@ BIND_GLOBAL( "Objectify", function ( type, obj )
     elif IS_REC( obj )  then
         SET_TYPE_COMOBJ( obj, type );
     fi;
-    if not IsNoImmediateMethodsObject(obj) then
+    if not ( IGNORE_IMMEDIATE_METHODS
+             or IsNoImmediateMethodsObject(obj) ) then
```
fingolfin: You need to run `make docomp` (once in a regular GAP build, and once in a build with `hpcgap` enabled) so that the files `src/c_type1.c` and `src/hpc/c_type1.c` are updated.

(This also work with out-of-tree builds, which is the most convenient way to update both the regular and the HPC version with a single clone)

zickgraf:
I was aware of `make docomp` and verified the performance improvement using it. However, I was not aware that the C-files are tracked in git. Should be fixed now :-)

PR #3744, opened 2019-11-14
Add Travis tests on arm64, ppc64le, s390x

### LINKED ISSUE #2362: Test, or block, big-endian CPUs
#2292 shows we have a bug generating random numbers on big-endian CPUs.

I'm going to suggest a major change. Unless someone is willing to run regular tests, including package tests, then I suggest we print a big warning, or maybe even refuse to build, on big-endian machines. At the moment we can't have confidence we are producing correct answers, which is worrying for software like GAP.

fingolfin:
Note that `s390x` is big endian which would be really nice in light of issue #2362.

And testing on arm64 might help with debugging issue #3919.

Resolves #2362.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ addons:
     - zlib1g-dev
     - python
     - python-pip
-    - perl
+    - jq    # for etc/travis_fastfail.sh
 
 #
 # The following test jobs are roughly sorted by duration, from longest to
@@ -149,17 +149,38 @@ matrix:
     # test Julia integration
     - env: TEST_SUITES="testinstall" JULIA=yes CONFIGFLAGS="--disable-Werror"
 
+    # build on non-x86 platform
+    - env: TEST_SUITES="docomp testinstall"
+      arch: arm64
+      dist: bionic
+      before_install:
+        # work around cache dir owned by root (see https://travis-ci.community/t/7822/6)
+        - sudo chown -fR $USER:$GROUP ~/.cache/pip/wheels
+
+    # build on non-x86 platform
+    - env: TEST_SUITES="docomp testinstall"
+      arch: ppc64le
+      dist: bionic
+
+    # build on non-x86 platform; big endian!
+    - env: TEST_SUITES="docomp testinstall"
+      arch: s390x
+      dist: bionic
+      before_install:
+        # work around cache dir owned by root (see https://travis-ci.community/t/7822/6)
+        - sudo chown -fR $USER:$GROUP ~/.cache/pip/wheels
+
+# use travis_terminate below to ensure travis immediately aborts upon error,
+# and also to work around timeouts (see https://travis-ci.community/t/7659)
 script:
-  - set -e
-  - etc/travis_fastfail.sh
-  - pip install --user gcovr==4.1
-  - bash etc/ci-prepare.sh
-  - bash etc/ci.sh
+  - etc/travis_fastfail.sh || travis_terminate $?
+  - python -m pip install --user gcovr==4.1 || travis_terminate $?
+  - bash etc/ci-prepare.sh || travis_terminate $?
+  - bash etc/ci.sh || travis_terminate $?
 
 after_script:
-  - set -e
-  - bash etc/ci-gather-coverage.sh
-  - bash etc/ci-run-codecov.sh
+  - bash etc/ci-gather-coverage.sh || travis_terminate $?
+  - bash etc/ci-run-codecov.sh || travis_terminate $?
 
 notifications:
   email:
--- a/etc/ci-coveralls-merge.py
+++ b/etc/ci-coveralls-merge.py
@@ -2,6 +2,8 @@
 
 # Hacked up merger for testing
 
+from __future__ import print_function
+
 import json
 import os
 import os.path
@@ -14,12 +16,13 @@ print("Merging coveralls json coverage")
 # Special-cased gap-coveralls.json, because we rely
 # on GAP to set the correct service_name, pull-request number
 # etc
-print "file: gap-coveralls.json", # python2 avoiding line break.
+print("file: gap-coveralls.json", end="")
 if os.path.isfile('gap-coveralls.json'):
     with open('gap-coveralls.json', 'r') as f:
         merged = json.load(f)
     print(" done.")
 else:
+    print()
     print("WARNING: could not find gap-coveralls.json, quitting")
     sys.exit(0)
 
@@ -28,7 +31,7 @@ if not 'source_files' in merged:
     merged['source_files'] = []
 
 for fn in coverage_files:
-    print "file: %s" % (fn,), # python2 avoiding line break.
+    print("file: %s" % (fn,), end="")
     if os.path.isfile(fn):
         with open(fn, 'r') as f:
             cover = json.load(f)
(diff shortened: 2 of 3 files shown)


ChrisJefferson:
I had a friend try reproducing the aarch64 failure. They didn't see it with normal configure, configure --enable-debug, or by adding --coverage (as we do when configuring ci).

They are running Ubuntu 18.04 with gcc 7.4.0. This is what they see when loading GAP:

```
~/gap$ ./gap
 ┌───────┐   GAP 4.12dev-117-gb5b9381 of today
 └───────┘   Architecture: aarch64-unknown-linux-gnu-default64-kv8
 Configuration:  gmp 6.1.2, GASMAN, KernelDebug
 Packages:   AClib 1.3.1, Alnuth 3.1.1, AtlasRep 2.1.0, AutoDoc 2019.09.04, 
             AutPGrp 1.10.1, CRISP 1.4.5, Cryst 4.1.21, CrystCat 1.1.9, 
             CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.3, 
             IRREDSOL 1.4, LAGUNA 3.9.3, Polenta 1.3.9, Polycyclic 2.15.1, 
             PrimGrp 3.3.2, RadiRoot 2.8, ResClasses 4.7.2, SmallGrp 1.4.1, 
             Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.4, 
             utils 0.67
gap> 
```

fingolfin:
While the aarch64 and ppc64le crashes are puzzling, in the end I only really care about s390x, as it is big endian, and there it runs fine and gets to the test suite (I guess @dimpase will be happy ;-). But then we see an actual issues, due to the hash function (as mentioned on issue #2362). I see no way to get identical hashes here, though (because the hash function does not know whether our data consists of 16bit or 32bit words). So that leaves the following options:
1. Just drop those tests and pretend nothing happened ;-)
2. we already have different hashes for 32 vs 64 bit archs; so I simply add two more values, for BE.
3. we decide that we rather have non-reproducible hashes, and always scramble them on each GAP startup with some kind of global random seed (no idea how we'd obtain that, but that's an irrelevant detail for now). I think globally I like this most, *but* it's a rather major break from what we do right now everywhere else (where we strongly try to keep RNG states reproducible across GAP launches)

Other suggestions?

hulpke:
If you want an opinion, I would chose 2.  It is not a big extra cost (adding two more definitions), and unlikely to force further cases in the forseeable future.

fingolfin:
BTW, there is a second test failure seen on both ppc64le and s390x which I right now can't explain, caused by different results given by `OPERS_CACHE_INFO()` right after calling `CLEAR_CACHE_INFO();`
```
########> Diff in /home/travis/build/gap-system/gap/tst/testinstall/kernel/opers.tst:266
# Input is:
opcheck{[1..11]};
# Expected output:
[ 0, 0, 0, 7, 0, 0, 2, 0, 0, 0, 0 ]
# But found:
[ 0, 0, 0, 6, 0, 0, 2, 0, 0, 0, 0 ]
```
I have absolutely no idea right now what might be causing this. At first I thought the hash function for the method selection cache is not BE safe, but then I noticed the issue also pops up on ppc64le. Huh.

fingolfin:
Good news: the AArch64 builds don't crash anymore (no idea why... but OK).

Bad news: all builds end with a timeout that I can't explain:
```
#I  Errors detected while testing

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

fingolfin:
Some new observations:
- I've now hacked QUIT_GAP to always map to FORCE_QUIT_GAP
- and FORCE_QUIT_GAP prints a message, and that message *is* visible in the Travis logs, so we get there
- I also added `echo` statements to `etc/ci.sh` before and after calling GAP, and they show nicely before GAP runs; but not afterwards, so we never really "leave" GAP
- so it is really something in the process cleanup that locks up, but it's not in our code
- so now I am strongly suspecting that writing the coverage data is what is broken (or too slow, or whatever)
- I will thus now try with `NO_COVERAGE=1` to disable coverage (I'd like to have coverage, though, but one step at a time)

fingolfin:
Aargh, this makes no sense! Even with `NO_COVERAGE`, and calling `_exit` directly does not help.  The only vague idea I have left is that we are waiting for some child process somewhere or so?!?
I guess I could try calling `abort`?

fingolfin:
Progress! Sometimes some of the tests now pass... If this eventually stabilizes, I'd like to try and backport this to 4.11, to help with e.g. understanding the GC issue in #3919

fingolfin:
So I dug down on the weird difference in the output of `OPERS_CACHE_INFO`, where different `OperationHit` values are shown (7 on x86, but 6 on these three architectures), with the most genius debugging tool invented by mankind, `printf`. I set up things so that whenever `OperationHit` is incremented, its new value is printed as well as the name of the operation for which the cache was searched. I also logged the GAP call tree, so I know that all hits are inside the GAP library function `RunTests`

On aarch64 and s390x, I got this (for powerpc64le the Travis had another issue, but I am confident it'll be the same:
```
OperationHit = 1 for ReadLine
OperationHit = 2 for CloseStream
OperationHit = 3 for CloseStream
OperationHit = 4 for InputTextString
OperationHit = 5 for OutputTextString
OperationHit = 6 for PrintFormattingStatus
OperationHit = 7 for ReadLine
OperationHit = 8 for CloseStream
```

In contrast, on my Mac on x86, I got this:
```
OperationHit = 1 for ReadLine
OperationHit = 2 for CloseStream
OperationHit = 3 for CloseStream
OperationHit = 4 for Int          <--------- this is new
OperationHit = 5 for InputTextString
OperationHit = 6 for OutputTextString
OperationHit = 7 for PrintFormattingStatus
OperationHit = 8 for ReadLine
OperationHit = 9 for CloseStream
```

I inserted some more code to get a backtrace of the GAP code leading to this, and I found out that the `Int` call in question is the one in `lib/test.gi:213`; i.e., in this code:
```gap
      Print("\r# line ", pos[i],
            " of ", nrlines,
            " (", Int(pos[i] / nrlines * 100), "%)",
            "\c");
```

I still don't have a clue why it is in the method cache the one time but not the other. But OK, some minor difference in the environment might lead to a different path being taken somewhere and then an `Int` call is made differently, resulting in a slightly different cache content... So now I have at least hope that this might *not* be due to random memory corruption.

It also suggests a simple workaround: insert a call `Int(1/2)` before the problematic test. Let's see if that works.

ChrisJefferson:
Wow, thanks for battling this. Let's see how stable it is long term, but I'm happy to merge it (assuming all the tests pass of course)

fingolfin:
For the backport, see https://github.com/gap-system/gap/pull/3966

fingolfin:
Backported via 4d3e4919dbe7517d1badc5aad19b987c7cd9ac1d from PR #3966.

dimpase:
as Travis is gone for GAP (?), what's the status of this? No more s390 tests?

PR #3745, opened 2019-11-14
kernel: add new header gap_all.h

fingolfin:
So far, most package kernel extensions `#include "compiled.h"` in order
to conveniently get access to (almost) all source headers of GAP at once
(which is nice for us because it means we are free to rename headers or
move stuff around between them without breaking kernel extensions).

But the result is rather counterintuitive: if one is not familiar with
this, seeing `#include "compiled.h"` in code is not at all clear. Also,
the original of `compiled.h` was to be included by C code generated by
`gac`, and it contains stuff we'd rather not encourage others to use.

Thus, add a new header `gap_all.h` containing all `#include` statements
formerly found in `compiled.h` -- this resolves all concerns.

---

Of course it will take time for kernel extensions to switch to this new header, especially if they care about staying compatible with older GAP versions. But the sooner we add this, the sooner such a switch could happen. Hence it would be nice to backport this (any objections).

### DIFF
--- a/src/compiled.h
+++ b/src/compiled.h
@@ -18,71 +18,7 @@
 extern "C" {
 #endif
 
-#include "ariths.h"
-#include "blister.h"
-#include "bool.h"
-#include "calls.h"
-#include "code.h"
-#include "collectors.h"
-#include "compiler.h"
-#include "compstat.h"
-#include "costab.h"
-#include "cyclotom.h"
-#include "dt.h"
-#include "dteval.h"
-#include "error.h"
-#include "exprs.h"
-#include "finfield.h"
-#include "funcs.h"
-#include "gap.h"
-#include "gapstate.h"
-#include "gasman.h"
-#include "gvars.h"
-#include "info.h"
-#include "integer.h"
-#include "intrprtr.h"
-#include "io.h"
-#include "iostream.h"
-#include "listfunc.h"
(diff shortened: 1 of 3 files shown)


stevelinton:
No objections. It might be worth looking for cases where packages use the other functionality from `compiled.h`. Some cases might be functionality that should be moved elsewhere, others might be considered errors to fix in the package.

DominikBernhardt:
@fingolfin It seems to me we might want to add this to the release notes so people who write kernel extensions are actually aware of this change?

frankluebeck:
The change looks ok to me. I have checked that the `Browse` and `EDIM` packages work as usual with including `gap_all.h` instead of `compiled.h`.

fingolfin:
@stevelinton I am pretty sure we can easily adapt any package which uses something from `compiled.h` that is not in `gap_all.h`; and even if not, there is no immediate plan to remove `compiled.h`, at least not on my side, so package would still keep working.

@DominikBernhardt agreed, so I added the label

@frankluebeck thanks for the test!

stevelinton:
@fingolfin Agreed. My thinking was just that it might point us to things in`compiled.h` which ought to be somewhere else because they are generally useful.

fingolfin:
Backported to stable-4.11 in commit 35ffd5f63

PR #3746, opened 2019-11-18
Insert GAP_VERSION into sysinfo.gap

fingolfin:
This is useful for package build systems which want to verify the GAP version.

This is a minor change but potentially quite useful for packages, yet clearly won't interfere with anything badly, so I'd really like to backport it. At the same time, it is relevant for package authors, so I'd like to mention it in the release notes, together with other stuff relevant for package authors.


fingolfin:
Backported to stable-4.11 as f1f5dc553485638d36329b442b0b0386b55fffa7

PR #3747, opened 2019-11-19
Fix unexpected error when testing permutation groups for conjugacy

### LINKED ISSUE #3446: Error when testing permutation groups for conjugacy
The file
[conjbug.txt](https://github.com/gap-system/gap/files/3163647/conjbug.txt)
defines three permutation groups `a`, `b`, and `c` of degree 448. Then

    gap> RepresentativeAction(a,b,c);
    Error, no 1st choice method found for `^' on 2 arguments called from
    OnTuples( sub, subs.reps[o]
     ) at /Volumes/BigDisk/oldgap/gap4r8/lib/stbcbckt.gi:724 called from

Clearly this should work.
I have been able to trace this back to GAP 4.7, but did not have an older runneable version to test on.

hulpke:
This is a work-around for the problem encountered in #3446, that is the code accessing `subs.reps[o]` is not executed, if it is not in the expected form (`true` or a list).

In the cases I encountered this resolves the error (and leads to a correct result).

I however do not understand the code (and parameters) well enough to assure that this is simply an omission in the original code, and not a reflection of a deeper problem. Thge code thus issues a warning, if this happens.
I have been running my own calculations with this modification for half a year now and did not encounter any problems, I thus want to offer it to the general system.

Resolves #3446


ChrisJefferson:
This is also my best guess as to how to fix this too (and I'm also unsure)

fingolfin:
Thank you!
I have absolutely no idea what's going on in that code, but if both of you think it's plausibel, I fully trust that (and should there be any unexpected regression, we'll deal with that as usual *shrug*).

fingolfin:
Should this be backported? Normally I'd say "no, to risky, let's first test it for a while in `master`", but since @hulpke has been using it for half a year anyway, it seems not so bad... Thoughts?

hulpke:
@fingolfin 
As for backporting, it fixes an issue that has never arisen in 20+ years (and I think I am the only one who encountered it so far) and it gives an error, not a wrong result.

I thus think it is safe to not backport it -- we can always issue a fix later.

PR #3751, opened 2019-11-20
kernel: replace unsafe uses of 1L

fingolfin:
On some 64bit systems (most notably, Windows), sizeof(long) == 4, and so
expressions like `1L << 60` overflow. Moreover, an enum strictly
speaking is only guaranteed to be able to represent an int, and so using
it to represent 64bit constants is not safe; we hence turn
INT_INTOBJ_MIN and INT_INTOBJ_MAX into global static constants.

Note that there are other issues with compiling GAP on native Windows (i.e., without cygwin).

### DIFF
--- a/src/bits_intern.h
+++ b/src/bits_intern.h
@@ -34,7 +34,7 @@
 /* constructs a mask that selects bits <from> to <to> inclusive of a UInt */
 static inline UInt MaskForCopyBits(UInt from, UInt to)
 {
-    return ((to == BIPEB - 1) ? 0 : (1L << (to + 1))) - (1L << from);
+    return ((to == BIPEB - 1) ? 0 : ((UInt)1 << (to + 1))) - ((UInt)1 << from);
 }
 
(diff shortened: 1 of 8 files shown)


ChrisJefferson:
While I'm happy with the general direction of this fix, I'm also happy with making the constants #defines, if you can find the right combination of static and const to make them work that's fine, but this seems to be annoyingly difficult.

fingolfin:
I went for the `#define`, as there simply is no other portable way to do this in C (in C++11, apparently one can do `enum Foo : int64_t { val = (int64_t)1 << 60 };` do get enums based on 64 bit systems; but in C, it's always `sizeof(enum) <= sizeof(int)` empirically at least. Anything else is a non-portable extension (on which we relied so far, but which extended warnings did warn about, we just don't have those enabled).

fingolfin:
Backported to stable-4.11 in a20c8f40883c8656a240317d732229dbe7c3b5ab

PR #3752, opened 2019-11-20
Remove uses of "long constants" such as 0L, 1L

fingolfin:
In all instances, this turned out to be pointless: either we were immediately (usually implicitly) converting/coercing the value to a specific type anyway; or a long was not necessary to start with; or it was an insufficient attempt to get a 64bit constant (see PR #3751).

This is a big PR 99% of which was made with a script. The changes in here actually helped me make PR #3751, because after removing the useless uses, just a handful remained which I either put in that PR, or took care of in the final (hand crafted) commit in this PR.


src/error.c
```diff
@@ -147,10 +147,10 @@ static Obj FuncUpEnv(Obj self, Obj args)
         depth = INT_INTOBJ(ELM_PLIST(args, 1));
     }
     else {
-        ErrorQuit("usage: UpEnv( [ <depth> ] )", 0L, 0L);
```
ChrisJefferson: I'm guessing changes like this (the change of `( [` to `([` are some kind of scripting error?

fingolfin: Ouch, indeed: I think my regex was bad (I wrote something like `(Error.*)\( *([^;]+), *0L *, *0L* \);` for the capturing -- clearly the first group should have been something like `(Error[A-Za-z]*)`... *sigh*. I'll fix this. Thank you for looking carefully.

fingolfin: Seems in addition to UpEnv and DownEnv, also QUIT and FORCE_QUITE were affected, but that's it (determined by grepping the diff between this PR and master). Will push a fix.

PR #3758, opened 2019-11-22
kernel: refactor and cleanup sysmem.c; fix some obscure bugs

fingolfin:
This PR got a bit bigger than I intended. I recommend to look at it commit by commit, and to enable GitHub's "ignore whitespace" option.

The original motivation was to understand the code, then I noticed some minor weird differences, then I tried to minimize differences, and then I went overboard... :-)

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -1109,7 +1109,7 @@ static void MaybeMoveBags(void)
 
 void FinishBags( void )
 {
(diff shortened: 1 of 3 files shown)


fingolfin:
Indeed. I did make sure to test with `-s 0` as well as in plain mode; and on both Linux and macOS. I tried to split this into not too complex commits, in the hope that if we find a regression, we can bisect it to one not too big commit

PR #3759, opened 2019-11-25
Fix an issue with error reporting in parallel threads.

rbehrends:
Without making the `ERROR_OUTPUT` constant immutable in `lib/error.g`, it belongs to the main thread and cannot be used for error reporting in concurrent threads.


fingolfin:
Ah, I didn't see this yesterday, otherwise I would have asked that the tests this is supposed to fix are re-enabled... I'll submit another PR for that, then.

PR #3760, opened 2019-11-25
Change Concatenation to StringFormatted

ChrisJefferson:
Personally, I prefer `StringFormatted` to `Concatenation` for joining multiple strings together. However, I don't know if other like it so before I make lots of changes I thought I'd make a small PR demonstrating the change.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -182,21 +182,21 @@ BindGlobal( "FindPackageInfosInSubdirectories", function( pkgdir, name )
     fi;
 
     file:= Filename( [ pkgdir ],
-                      Concatenation( name, "/PackageInfo.g" ) );
+                      StringFormatted( "{}/PackageInfo.g", name ) );
     if file = fail then
       files := [];
       # Perhaps some subdirectories contain `PackageInfo.g' files.
       for subdir in Set( DirectoryContents( pkgpath ) ) do
         if not subdir in [ ".", ".." ] then
           pkgpath:= Filename( [ pkgdir ],
-                              Concatenation( name, "/", subdir ) );
+                              StringFormatted( "{}/{}", name, subdir ) );
           if pkgpath <> fail and IsDirectoryPath( pkgpath )
                               and not subdir in [ ".", ".." ] then
             file:= Filename( [ pkgdir ],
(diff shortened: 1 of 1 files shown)


lib/package.gi
```diff
@@ -182,21 +182,21 @@ BindGlobal( "FindPackageInfosInSubdirectories", function( pkgdir, name )
     fi;
 
     file:= Filename( [ pkgdir ],
-                      Concatenation( name, "/PackageInfo.g" ) );
+                      StringFormatted( "{}/PackageInfo.g", name ) );
```
fingolfin: Personally, I find the old code both shorter and more readable.

stevelinton: I would agree with @fingolfin. I would say that if is really just joining strings together then `Concatenation` is more natural, but if it also involves converting other data to a string, as in the not uncommon:
```
Concatenation("foo :",String(foo)," etc.\n");
```
then `StringFormatted` is better.

ThomasBreuer: I would use `StringFormatted` only if nontrivial formatting is involved.
For the situation mentioned by @stevelinton (concatenation after conversion of objects to strings if necessary), I would regard the currently undocumented `STRINGIFY` as more appropriate; perhaps `STRINGIFY` could become documented, with a nicer name.

frankluebeck:
I find the calls of `Concatenation` shorter and easier to understand (essentially self-explaining). (But I can understand that people who regularly use similar functions in C, python or other languages also like or even prefer `StringFormatted`.)

And another important point against the proposed changes is SPEED:

```
gap> name:="/usr/local/gap4/pkg";;  subdir:="mypkg";;
gap> for i in [1..100000] do
>   a := Concatenation( name, "/", subdir );
> od; time;
24
gap> for i in [1..100000] do
>   a := StringFormatted( "{}/{}", name, subdir);
> od; time;
379
```
Seeing this example I have difficulties to think of any case where I would use `StringFormatted`.
Turning `STRINGIFY`, mentioned by @ThomasBreuer, into a documented function sounds more useful.

ChrisJefferson:
Increasing speed is a thing which could be done by porting to C, but if in general people prefer Concatenation, then I won't start changing the library for no good reason.

PR #3761, opened 2019-11-25
Fix bugs in HPC-GAP serialization code and improve its tests

fingolfin:


rbehrends:
Looks good, minus the problem that test output for object maps/sets depends on ordering?

src/objset.c
```diff
@@ -595,14 +616,16 @@ void ClearObjMap(Obj map) {
  *  This function returns all values from the map.
  */
 
-Obj ObjMapValues(Obj set) {
-  UInt len = CONST_ADDR_WORD(set)[OBJSET_USED];
-  UInt size = CONST_ADDR_WORD(set)[OBJSET_SIZE];
+Obj ObjMapValues(Obj (map))
```
ChrisJefferson: some weird stuff is going on here, and two lines lower, with double brackets.

tst/testinstall/hpc/serialize.tst
```diff
@@ -214,4 +218,23 @@ gap> DeserializeNativeString("\000\377");
...
+#
+# verify that stuff that gets serialized twice is
+# deserialized into the same object each time
+#
+gap> CheckRepeatedSerialization := function(y)
+>   CheckSerializationGeneric([y, y], x->IsIdenticalObj(x[1], x[2]));
+>   CheckSerializationGeneric(rec(a:=y,b:=y), x->IsIdenticalObj(x.a, x.b));
+> end;;
+gap> CheckRepeatedSerialization("abc");
+gap> CheckRepeatedSerialization(2^100);
```
ChrisJefferson: Maybe pop in a comment saying this difference is fine, as these are large integers (assuming you plan on leaving the test the same)

src/hpc/serialize.c
```diff
@@ -589,8 +589,8 @@ static void SerializeRecord(Obj obj)
     len = LEN_PREC(obj);
     WriteImmediateObj(INTOBJ_INT(len));
     for (i = 1; i <= len; i++) {
-        UInt rnam = GET_RNAM_PREC(obj, i);
-        Obj  rnams = NAME_RNAM(rnam);
+        Int rnam = GET_RNAM_PREC(obj, i);
+        Obj rnams = NAME_RNAM(rnam >= 0 ? rnam : -rnam);
```
ChrisJefferson: Might be getting comment-heavy, but maybe add a comment here saying we save all rnams as positive, so the record will be sorted when loaded back in (I had to go check a few files to convince myself this was the right thing to do, a hint might be nice for future readers)

fingolfin: Will do.

src/objset.c
fingolfin: Oops, will fix

src/hpc/serialize.c
fingolfin: Actually, we don't store the `rnam` at all, as they can change across sessions, we store the *name* of the RNAM, and for that we must take its absolute value. I'll add comments to clarify this.

tst/testinstall/hpc/serialize.tst
fingolfin: Actually, this is IMHO not fine, so I'll add a FIXME just like with the other cases. Actually looking at this problem was what caused me to write this PR (as I kept discovering more and more issues); ironically, I've not yet managed to fix the initial problem ;-).

src/hpc/serialize.c
ChrisJefferson: Ah yes, I misread the method entirely, sorry! However, it's still worth a comment :)

fingolfin: I totally agree. Please let me know if you like what I wrote, or some hints what you'd consider better.

ChrisJefferson:
I'm now happy to merge, but I'll let you squash the fixup! commit first

PR #3766, opened 2019-12-01
Some kernel refactoring

fingolfin:
The commits are independent from each other, so best reviewed in isolation. I can also split this PR if people want me to.

### DIFF
--- a/src/modules.c
+++ b/src/modules.c
@@ -209,34 +209,31 @@ static Int SyLoadModule(const Char * name, InitInfoFunc * func)
 */
 static Obj FuncLOAD_DYN(Obj self, Obj filename, Obj crc)
 {
-    StructInitInfo * info;
-    Obj              crc1;
-    Int              res;
-    InitInfoFunc     init;
-
     /* check the argument                                                  */
     RequireStringRep("LOAD_DYN", filename);
     if (!IS_INTOBJ(crc) && crc != False) {
         RequireArgument("LOAD_DYN", crc, "must be a small integer or 'false'");
     }
 
-    /* try to read the module                                              */
-#ifdef HAVE_DLOPEN
-    res = SyLoadModule(CONST_CSTR_STRING(filename), &init);
-    if (res == 1)
-        ErrorQuit("module '%g' not found", (Int)filename, 0);
-    else if (res == 3)
-        ErrorQuit("symbol 'Init_Dynamic' not found", 0, 0);
-#else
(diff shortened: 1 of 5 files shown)


src/modules.c
```diff
@@ -311,14 +304,10 @@ static Obj FuncLOAD_STAT(Obj self, Obj filename, Obj crc)
 
     /* check the crc value                                                 */
     if (crc != False) {
-        crc1 = ObjInt_Int(info->crc);
-        if (!EQ(crc, crc1)) {
+        if (INT_INTOBJ(crc) != info->crc) {
```
ChrisJefferson: Just checking, on a 32bit system do we know the crcs fit into an INTOBJ?

fingolfin: I actually wondered about this, too, *but* the code explicitly requires INTOBJs (the function checks its arguments), and the function `LOAD_STAT` has done that (require an INTOBJ) since at least GAP 4.4.12. 

So I dug a bit more, and finally arrived at `SyGAPCRC`, and it turns out that function deliberately throws away the four least significant bits of the CRC to make them fit into an INTOBJ (note that it actually discards one bit beyond what it'd have to discard).

In `FuncGAP_CRC` we call `ObjInt_Int(SyGAPCRC(...))`, but that code used to use `INTOBJ_INT`. Turns out @ChrisJefferson changed that in 3f25ca2e0ed087db04a778d1d609723da8d42a9f (and I approved and merge that *g*), probably assuming that `SyGAPCRC` returns something that might not fit into an INTOBJ (an assumption which I still find quite reasonable, one really has to dig into the code to learn the truth *sigh*)

Anyway, we could easily change that to not discard any bits, as (a) we don't need to on 64bit systems which most people use, and (b) even on 32bit systems, using `ObjInt_Int` and `Int_ObjInt` makes dealing with full 32bit ints rather easy and convenient. And the code dealing with CRC checking is not exactly a hotpath either ;-).

That said, such a change could be easily done later, atop the changes in this PR, as we'd essentially have to change only the line under discussion in a trivial way (back to `if (!EQ(crc, ObjInt_Int(info->crc)))`, and also the `Require*` check at the start of the function.

ChrisJefferson: My use of `ObjInt_Int` is mostly I use it anywhere that is not on a fast-path (and checking a CRC can't be that fast, as you have to calculate the CRC!), just because it's error checked and handles more cases well. But in this case I'm happy to use INTOBJ_INT. Thanks.

PR #3768, opened 2019-12-02
Fix some warnings reported by clang static analyzer

fingolfin:
See https://clang-analyzer.llvm.org

I had this sitting around for some time now, as it was not complete. But I figure it's better to get it merged, complete or not, to avoid wasted or duplicate efforts :-)

### DIFF
--- a/src/cyclotom.c
+++ b/src/cyclotom.c
@@ -1172,24 +1172,17 @@ static Obj ProdCycInt(Obj opL, Obj opR)
 
     /* otherwise multiply every coefficent                                 */
     else {
-        hdP = NewBag( T_CYC, SIZE_CYC(opL) * (sizeof(Obj)+sizeof(UInt4)) );
-        SET_NOF_CYC(hdP, NOF_CYC(opL));
+        len = SIZE_OBJ(opL);
+        hdP = NewBag(T_CYC, len);
+        memcpy( ADDR_OBJ(hdP), CONST_ADDR_OBJ(opL), len);
         len = SIZE_CYC(opL);
-        cfs = CONST_COEFS_CYC(opL);
-        exs = CONST_EXPOS_CYC(opL,len);
         cfp = COEFS_CYC(hdP);
-        exp = EXPOS_CYC(hdP,len);
         for ( i = 1; i < len; i++ ) {
-            CHANGED_BAG( hdP );
-            prd = PROD( cfs[i], opR );
-            cfs = CONST_COEFS_CYC(opL);
-            exs = CONST_EXPOS_CYC(opL,len);
+            prd = PROD( cfp[i], opR );
             cfp = COEFS_CYC(hdP);
-            exp = EXPOS_CYC(hdP,len);
             cfp[i] = prd;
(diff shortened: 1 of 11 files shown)


src/cyclotom.c
```diff
@@ -1172,24 +1172,17 @@ static Obj ProdCycInt(Obj opL, Obj opR)
 
     /* otherwise multiply every coefficent                                 */
     else {
-        hdP = NewBag( T_CYC, SIZE_CYC(opL) * (sizeof(Obj)+sizeof(UInt4)) );
```
ChrisJefferson: Can you give me a feeling what's changed here, the code seems to have entirely changed.

ChrisJefferson:
This all seems fine, although (as commented) one change seems more significant, and I wasn't clear what's happening.

src/cyclotom.c
fingolfin: So the static analyzer complained about the values `exs` and `exp` in the old lines 1179 and 1181. I at first just removed them, but then had a closer look at the code. It turns out that it copies half the data; and the other half are coefficients which are multiplied by a fixed scalar. So I changed the code to first copy everything via a single `memcpy`, and then update the scalars in-place. This way, we only need to compute one bag pointer per iteration (instead of 4), and it also is conceptually easier to understand, IMHO. I think this all gets clearer if one doesn't use the "unified diff" view on GitHub, but rather compares old and new code side-by-side.

src/cyclotom.c
```diff
@@ -1172,24 +1172,17 @@ static Obj ProdCycInt(Obj opL, Obj opR)
...
         for ( i = 1; i < len; i++ ) {
-            CHANGED_BAG( hdP );
-            prd = PROD( cfs[i], opR );
-            cfs = CONST_COEFS_CYC(opL);
-            exs = CONST_EXPOS_CYC(opL,len);
+            prd = PROD( cfp[i], opR );
             cfp = COEFS_CYC(hdP);
-            exp = EXPOS_CYC(hdP,len);
             cfp[i] = prd;
-            exp[i] = exs[i];
```
fingolfin: In this new code, this copy statement is taken care of by the single `memcpy` call at the top

src/cyclotom.c
```diff
@@ -1172,24 +1172,17 @@ static Obj ProdCycInt(Obj opL, Obj opR)
...
+        hdP = NewBag(T_CYC, len);
+        memcpy( ADDR_OBJ(hdP), CONST_ADDR_OBJ(opL), len);
         len = SIZE_CYC(opL);
-        cfs = CONST_COEFS_CYC(opL);
-        exs = CONST_EXPOS_CYC(opL,len);
         cfp = COEFS_CYC(hdP);
-        exp = EXPOS_CYC(hdP,len);
         for ( i = 1; i < len; i++ ) {
-            CHANGED_BAG( hdP );
-            prd = PROD( cfs[i], opR );
```
fingolfin: In the old code, input and output bag are different; in the new code, we work in-place in a single bag.

PR #3769, opened 2019-12-02
kernel: avoid using vfork

fingolfin:
Calling chdir in the child process after vfork can corrupt the
state of the parent process.

Reported by clang static analyzer <https://clang-analyzer.llvm.org>.

I don't think this actually ever caused problems, but then I can't be sure. But I *am* sure the code can't cause correctness regressions (at worst: a minor performance impact on some OSes, but I doubt it; there is no impact on Linux, where `fork` is blindingly fast anyway). I don't think it needs to be mentioned in release notes: if there was a problem, then it was around for years w/o anybody ever reporting it, or anything that sounded remotely like it could be related to this.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -3106,7 +3106,7 @@ UInt SyExecuteProcess (
       func2 = &NullSignalHandler;
 
     /* clone the process                                                   */
-    pid = vfork();
+    pid = fork();
     if ( pid == -1 ) {
         return -1;
     }


ChrisJefferson:
I don't know if cygwin special cases vfork vs fork (or if we even use this code path in cygwin), but fork is definately better, so let's see what happens.

fingolfin:
For Cygwin, we actually don't use this code, but instead an alternate implementation using `spawnve`. And super crazy, but: that variant of `SyExecuteProcess` actually calls `chdir` on the parent process first, and never bothers to restore the current dir... Huurmpf.

PR #3771, opened 2019-12-03
INSTALL.md: some tweaks

fingolfin:
This is based on some recent Slack discussion. Should be backported, I think.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -291,8 +291,6 @@ complete in about a minute. You will get a large number of lines with
 output about the progress of the tests, for example:
 
     gap> Read( Filename( DirectoriesLibrary( "tst" ), "testinstall.g" ) );
-    You should start GAP4 using `gap -A -x 80 -r -m 100m -o 1g -K 2g'.
-
     Architecture: SOMETHING-SOMETHING-gcc-default64
 
     testing: ..../gap-4.X.Y/tst/testinstall/alghom.tst
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -44,8 +44,8 @@ operating system. In particular, it does not include an automatic installer
 program.
 
 Installing the GAP distribution with all the packages and full data libraries
-takes about 1.6 GB of disk space and (except on Windows) will require both a C
-and a C++ compiler (gcc is recommended) to be installed on your system.
+will require (except on Windows) both a C and a C++ compiler (gcc or clang
+is recommended) to be installed on your system.
 To get maximum benefit from GAP and from various packages, we recommend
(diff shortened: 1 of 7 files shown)


INSTALL.md
```diff
@@ -44,8 +44,8 @@ operating system. In particular, it does not include an automatic installer
 program.
 
 Installing the GAP distribution with all the packages and full data libraries
-takes about 1.6 GB of disk space and (except on Windows) will require both a C
-and a C++ compiler (gcc is recommended) to be installed on your system.
+and (except on Windows) will require both a C and a C++ compiler (gcc or clang
```
alex-konovalov: check this sentence

wucas:
Sorry for the long comment, none of it means the PR cannot be merged. 
Regarding the command line option with which one apparently is supposed to start gap when running these tests I just noticed that this is even shown when I do start GAP with these options which might be confusing. Or at least I just double checked the options I used because I interpreted this as that I did not use these options. But that might just be me reading this wrong.

fingolfin:
Many good comments, I'll address them later. On describing the output of `Test` resp. `TestDirectory`: Yes, that should be done, but not in the README but rather in the documentation for those functions (and then the README can point to that).

INSTALL.md
```diff
@@ -44,8 +44,8 @@ operating system. In particular, it does not include an automatic installer
...
+are recommended) to be installed on your system.
```
wilfwilson: I feel pretty sure that this should still be 
```suggestion
is recommended) to be installed on your system.
```
Of course it might be my peculiar interpretation of the English language, but with an 'or' statement, I like to use the same verb conjugation that I would use if I 'expanded' the statement:
> gcc or clang is recommend → gcc is recommended or clang is recommended

I don't mind if you disagree and want to keep the sentence as is. Otherwise, no objections to this PR 🙂

fingolfin:
Thanks everybody for the feedback. I tried to fix the language issues pointed out by @alex-konovalov and @wilfwilson, please have a look. I also updated another TestDirectory output example inside the reference manual to match what is in  `INSTALL.md` (and fixed the title of this Pr and the commit to reference `INSTALL.md`, not `README.md`... ooops). I also took into account what @wucas wrote, and (a) created issue #3776, (b) added a commit which removes the messages in the tests that recommends users to start GAP with certain options - I *never* heed those anyway, and I agree they are confusing. For people resp. occasions where the might matter (e.g. could be in CI builds), we can still set those options, of course.

fingolfin:
Backported to stable-4.11 in 7ae90d6f1

Issue #3772, opened 2019-12-05
Master branch broken due to factint / HPC-GAP interaction


ChrisJefferson:
The master branch is currently broken in HPC-GAP, due to factint

This problem has existed previously, and was handled by removing factint when testing HPC-GAP. I would go back to this "fix".

ChrisJefferson:
This isbecause FactInt was upgraded from 1.6.2 to 1.6.3, but I haven't looked at exactly how that is causing problems.

ChrisJefferson:
Found problem, investigating fix.

PR #3773, opened 2019-12-05
Add default IsBound on small lists for large integers

ChrisJefferson:
This provides an implementation of IsBound for small lists when they are given an integer which is too large to index the list.

This adds support for IsBound to a few list types, including non-dense lists and atomic lists. The atomic list support fixes the master branch ( #3772 ). The reason this works in non-HPC gap is atomic lists are just implemented as plists, and dense plists support operations like `IsBound(x[2^100])` already.

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -3006,6 +3006,20 @@ InstallMethod( IsBound\[\],
     return index <= Length( list );
     end );
 
+#############################################################################
+##
+#M  IsBound( list[i] ) . . . . . IsBound for small lists with large arguments
+##
+InstallMethod( IsBound\[\],
+    "for a small list and large positive integer",
+    [ IsSmallList, IsPosInt ],
+    function( list, index )
+    if IsSmallIntRep(index) then
+        TryNextMethod();
+    else
+        return false;
+    fi;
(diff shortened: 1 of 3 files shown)


wilfwilson:
I propose that we merge this quickly so that Travis is working again.

wilfwilson:
I've just spoken to @ChrisJefferson. To get Travis working again, I'm going to merge this.

However, he's open to alternative ways to fix this problem, and isn't going to be offended if people want to propose alternative PRs that change this commit to do it in a nicer way.

fingolfin:
Backported to stable-4.11 in commit ab1eb4066dc9ab0d459339e3a727b826ece69114

fingolfin:
I don't think we need or should mention this in the release notes, as (a) the issue it resolves is not visible to people who use stable GAP versions, and (b) this is a really technical change, which a regular user won't be able to appreciate ("you added what? and why should I care???") without a length explanation... in the end, this just makes "things work as they should" (which is good! but not that newsworthy).

But please don't hesitate to disagree vocally :-)

wilfwilson:
Agreed 🙂

PR #3778, opened 2019-12-09
Tweak README.md and INSTALL.md instructions some more

fingolfin:
- suggest `apt-get` over `apt` as the latter is not present by default
  (at least on older systems)
- make sure users who follow the instructions in INSTALL.md are made
  aware of some of the information in README.md that pertains to them,
  even if they are not installing GAP from the git repository.
- don't mention the obsolete BOB tool
- list zlib as an optional dependency
- clarify that GMP is optional, too

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,7 +24,6 @@ Alternative installation methods which aim to simplify the installation
 mostly by offering precompiled binaries are:
 
 * GAP installer for Homebrew (package manager for macOS)
-* BOB - a tool to download and build GAP and its packages from source
 * Docker image for GAP and most of the packages
 * the rsync-based binary distribution for Linux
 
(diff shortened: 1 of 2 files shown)


dimpase:
IMHO one has to go to a museum to find a Debian apt-less system, but well :-)

fingolfin:
@dimpase Then you are working with different people than I do, I guess

dimpase:
... in 1996 in l'Aquila I installed GAP on a computer running an old, at that time, UNIX. It took a while to download a tarball, as the university was connected to the outside world via a couple of phone line modems :-)

ChrisJefferson:
Experience is that autoconf has lots of versions, and there is a chance someone might have an incompatible version, while that tends not to happen with the other libraries (sometimes Readline).

Autoconf is also unusual in that there is a good chance any given Linux distro has multiple versions you could install, while other libraries tend to just have a single version.

fingolfin:
Backported to stable-4.11 in 08b160b1b

Issue #3774, opened 2019-12-05
Computation of automorphism group of very large group can cause crash to desktop


Frightning:
I have been able to cause my install of GAP 4.10.2 (32-bit via .exe installer on Windows 10) to crash straight to the desktop without any error message or warning.
To do this, I did the following series of computations (computing the automorphism series of a finite group for a few steps, eventually constructing a very large and then attempting to construct automorphism group). After only a few minutes of running time, GAP crashes. No error message, memory warning, etc. I have GAP set to warn me if it tries to use more than 2gb of ram (and have 16gb of system ram, so there's no lack of physical memory).

Instructions to reproduce:
I start with a group in the SmallGroups library, then compute it's automorphism group. Which happens to be of order 1536. Because of this, IdSmallGroup is not available, and the only way to continue efficiently is to convert to a PermutationGroup (or alternatively PcGroup), I chose the former via the NiceObject command. I then iteratively compute automorphism group and then nice object until reaching a very large (of size~2 quadrillion!). The computation of A4 and G4 took at least a minute combined, so expect it to take a while. Attempting to compute A5 as shown however will crash GAP to desktop after a few minutes.
```
gap> G:=SmallGroup(384,17948);
<pc group of size 384 with 8 generators>
gap> A1:=AutomorphismGroup(G);
<group of size 1536 with 7 generators>
gap> G1:=NiceObject(A1);
<permutation group of size 1536 with 7 generators>
gap> A2:=AutomorphismGroup(G1);
<group of size 6144 with 8 generators>
gap> G2:=NiceObject(A2);
<permutation group of size 6144 with 8 generators>
gap> A3:=AutomorphismGroup(G2);
<group of size 196608 with 11 generators>
gap> G3:=NiceObject(A3);
<permutation group of size 196608 with 11 generators>
gap> A4:=AutomorphismGroup(G3);
<group of size 2078076976496640 with 7 generators>
gap> G4:=NiceObject(A4);
<permutation group of size 2078076976496640 with 7 generators>
gap> A5:=AutomorphismGroup(G4);
```
This behavior is consistently reproducible for me. I would expect an error message of some kind at the very least, if not entering the break loop so I can at least abandon the attempt if it has somehow failed. The group G4 is not solvable, so maybe this has something to do with it?

hulpke:
Dear @Frightning 
Would you be willing to tell us your name and where you're from?

Thank you,

  Alexander Hulpke (Colorado State)
[...]

PR #3783, opened 2019-12-12
Improvements to Automorphism group/Group isomorphism

hulpke:
In particular concerning permutation representations.

These changes aid in calculations for automorphism groups or group isomorphism test in harder cases (that were impossible before), such as #3774.

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -232,6 +232,7 @@ BindGlobal("AGSRAutomLift",function(ocr,nat,fhom,miso)
       for i in [1..Length(genimages)] do
 	genimages[i]:=genimages[i]*s[i];
       od;
+
       # later use NC version
       hom:=GroupHomomorphismByImagesNC(ocr.group,ocr.group,
 	      ocr.generators,genimages);
@@ -248,33 +249,35 @@ end);
 # the condition can become expensive.
 # First try `SubgroupProperty`, but when it stalls attempt to find
 # minimal supergroups and prove that none of them satisfies.
-InstallGlobalFunction(SubgroupConditionAbove,function(G,cond,S)
-local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort;
+BindGlobal("SubgroupConditionAboveAux",function(G,cond,S1,avoid)
+local S,c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort,clim,worked,pp,
+  cnt,locond,tstcnt;
 
+  S:=S1;
   Info(InfoMorph,2,"SubgroupAbove ",IndexNC(G,S));
 
   # first try, how far `SubgroupProperty` goes
-  good:=[];
-  bad:=[];
   b:=0;
+  tstcnt:=0;
   abort:=false;
+  # if less than 1/100 percent of elements succeed, assume close
+  # to the subgroup has been found, and rather aim to prove there
+  # will not be more.
+  clim:=Maximum(QuoInt(IndexNC(G,S),10000),1000);
   c:=SubgroupProperty(G,
     function(elm)
       if abort then
         return true; # are we bailing out since it behaves too badly?
-      elif cond(elm) then
-        Add(good,elm);
+      fi;
+      tstcnt:=tstcnt+1;
+      if cond(elm) then
         S:=ClosureGroup(S,elm); # remember
         Info(InfoMorph,3,"New element ",IndexNC(G,S));
         b:=0;
         return true;
       else
-        Add(bad,elm);
         b:=b+1;
-        # if less than 1/50 percent of elements succeed, assume close
-        # to the subgroup has been found, and rather aim to prove there
-        # will not be more.
-        if b*5000>IndexNC(G,S) then
+        if b>clim then
           abort:=true;
         fi;
         return false;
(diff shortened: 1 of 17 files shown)


hulpke:
These are three examples of groups that were impossible before but now work in a few GB.

[benfieldaut.txt](https://github.com/gap-system/gap/files/3958166/benfieldaut.txt)
[isomtestdec19_2.txt](https://github.com/gap-system/gap/files/3958167/isomtestdec19_2.txt)
[isomtestdec19.txt](https://github.com/gap-system/gap/files/3958168/isomtestdec19.txt)

fingolfin:

lib/grpperm.gd
```diff
@@ -529,6 +529,30 @@ DeclareGlobalFunction( "ReducedPermdegree" );
...
 DeclareGlobalFunction("MovedPointsPerms");
 
+
+#############################################################################
+##
+#A  OrbitsMovedPoints( <G> )
+##
+##  <#GAPDoc Label="AllBlocks">
+##  <ManSection>
+##  <Attr Name="AllBlocks" Arg='G'/>
```
ChrisJefferson: I'm guessing this reference to `AllBlocks` (and two lines above) is a cut+paste error?

lib/grpperm.gi
```diff
@@ -466,6 +466,48 @@ InstallMethod( NrMovedPoints,
...
+  # avoid new stabilizer chains if they are not computed yet
+  if OrbitsMovedPoints(G)<>OrbitsMovedPoints(H) then
+    return false;
+  fi;
+  if not (HasGeneratorsOfGroup(G) and HasGeneratorsOfGroup(H)) then
+    TryNextMethod();
+  fi;
+  if IsEqualSet( GeneratorsOfGroup( G ), GeneratorsOfGroup( H ) ) then return true;fi;
+
+  if Size(G)<>Size(H) then return false;fi;
```
ChrisJefferson: This probably wants to be `HasSize(G) and HasSize(H) and Size(G)<>Size(H)`, because otherwise GAP will just go and build stabilizer chains for both `G` and `H`, to find their size, which makes the next part pointless.

lib/grpperm.gi
```diff
@@ -466,6 +466,48 @@ InstallMethod( NrMovedPoints,
...
+
+
+#############################################################################
+##
+#M  \=( <G>, <H> )  . .  . . . . . . . . .  test if two perm groups are equal
+##
+InstallMethod( \=, "perm groups", IsIdenticalObj, [ IsPermGroup, IsPermGroup ],
+function ( G, H )
+  # avoid new stabilizer chains if they are not computed yet
+  if OrbitsMovedPoints(G)<>OrbitsMovedPoints(H) then
```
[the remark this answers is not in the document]
ChrisJefferson: I agree that I didn't know that a permutation group could lack `GeneratorsOfGroup` (can it happen just when it's the identity group? Or would it still be an empty list?)

lib/grpperm.gi
```diff
@@ -466,6 +466,48 @@ InstallMethod( NrMovedPoints,
...
+
+
+
+#############################################################################
+##
+#M  \=( <G>, <H> )  . .  . . . . . . . . .  test if two perm groups are equal
+##
+InstallMethod( \=, "perm groups", IsIdenticalObj, [ IsPermGroup, IsPermGroup ],
+function ( G, H )
+  # avoid new stabilizer chains if they are not computed yet
```
ChrisJefferson: If you want another quick pre-test, to avoid even calculating `OrbitsMovedPoints`, you could just check the largest moved points of both `G` and `H`. Of course, depending on your application that always be the same.

lib/grpperm.gi
hulpke: This size comparison comes just before the backtrack. Presumably there should be an extra check with HasSize at the start. will change.

lib/grpperm.gi
hulpke: There isn't a current example, but theoretically there could be groups wiuthout generators (defined as stabilizer of a structure).

The application I have is a group acting on subgroups. There the largest point does not distinguish much, but then again its cost is minimal.

lib/grpperm.gi
```diff
@@ -466,6 +466,48 @@ InstallMethod( NrMovedPoints,
...
 
+#############################################################################
+##
+#M  OrbitsMovedPoints( <G> )
+##
+InstallMethod(OrbitsMovedPoints,"for a permutation group",true,[IsPermGroup],0,
+function(G)
+local o;
+  o:=Orbits(G,MovedPoints(G));
+  o:=List(o,x->Immutable(Set(x)));
```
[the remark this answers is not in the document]
hulpke: The `Perform` will not work, as `Orbits` returns immutable lists. The `List` and `Set` do the equivalent of `ShallowCopy`.

lib/autsr.gi
```diff
@@ -287,44 +289,155 @@ local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort;
...
+    until good=false;
+
+  fi;
+
+
+
+
+  Info(InfoMorph,3,"Did ",tstcnt," tests");
+  return S;
+  Error();
```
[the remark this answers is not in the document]
hulpke: Yes. This old code is obsolete now (but it wasn't clear initially that it would be completely obsolete). I have put it in comments now  (and will delete it in a while).

lib/grpperm.gi
```diff
@@ -466,6 +466,48 @@ InstallMethod( NrMovedPoints,
...
+  Sort(o);
```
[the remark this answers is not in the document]
hulpke: Thank you -- is changed. The application is is using the attribute value for binary search in a list of groups, so sorting and canonical form is important.

lib/grpperm.gi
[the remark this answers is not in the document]
hulpke: Yes, the attribute holds memory of size equivalent to one group element. But is this significant if the groups are large? What follows is a stabilizer chain which will cost the same memory many times over. So I don't see this as detrimental extra cost (but if this continues to be a concern I'm happy to add a degree bound).

As for the generator question, a group without generators (say defined as abstract stabilizer) might have its own method for the moved points, but that is at the moment purely academic. I move the `Generators` test earlier.

lib/grpperm.gi
ChrisJefferson: I don't know if you left this line in accidentally when you moved the Size check to the top. I think it will be better to remove this and let the StabChain stuff, as that is designed to only require a stab chain (and therefore Size) on only one of the two groups.

lib/grpperm.gi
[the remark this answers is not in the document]
fingolfin: You have of course an excellent point about `OrbitsMovedPoints` taking about the same amount of memory as a few random group elements (it's actually 2 in the "degree 1 million" case, not 1, but that doesn't change the validity of your argument). So, the question to my answer "is that a good idea" is "yes" :-).

lib/grp.gd
```diff
@@ -3317,6 +3317,32 @@ KeyDependentOperation( "PRump", IsGroup, IsPosInt, "prime" );
...
+#A  StructuralSeriesOfGroup( <G> )
+##
+##  <#GAPDoc Label="StructuralSeriesOfGroup">
+##  <ManSection>
+##  <Attr Name="StructuralSeriesOfGroup" Arg='G'/>
+##
+##  <Description>
+##  The structural series of a finite group<A>G</A> is a descending series of characteristic
+##  subgroups which goes throught the derived series of the solvable radical of $G$, refined into
+##  elementary abelian factors, as well as the socle and <A>Pker</A> of the radical factor.
```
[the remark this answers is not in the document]
hulpke: The kernel of the permutation action on the socle factors. This is the standard naming when working with solvable radical/trivial fitting

lib/autsr.gi
```diff
@@ -287,44 +290,155 @@ local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort;
...
+  # aim for zuppos that cannot intersect avoid
+  c:=ConjugacyClasses(q);
+  c:=Filtered(c,x->IsPrimePowerInt(Order(Representative(x))) and
+    not Representative(x) in ups);
+  # not order-p power in avoid
+  c:=Filtered(c,x->not Representative(x)^
+    (Order(Representative(x))/SmallestPrimeDivisor(Order(Representative(x))))
+    in ups);
+
+  # this also implies prime powers after primes
```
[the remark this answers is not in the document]
hulpke: prime powers after their respective primes.

lib/grpperm.gi
hulpke: Done.

lib/autsr.gi
```diff
@@ -287,44 +290,155 @@ local c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort;
...
-        S:=ClosureGroup(S,t[c]);
-        good:=true;
+  hom:=NaturalHomomorphismByNormalSubgroupNC(G,u);
+  q:=Image(hom,G);
+  ups:=Image(hom,avoid);
+  # aim for zuppos that cannot intersect avoid
+  c:=ConjugacyClasses(q);
+  c:=Filtered(c,x->IsPrimePowerInt(Order(Representative(x))) and
+    not Representative(x) in ups);
+  # not order-p power in avoid
```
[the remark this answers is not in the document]
hulpke: `avoid` is a subgroup. No power of prime order in that subgroup.

lib/clas.gd
```diff
@@ -437,3 +437,6 @@ DeclareGlobalFunction( "RationalClassesSolvableGroup" );
 ##
 
 DeclareGlobalFunction( "CentralizerSizeLimitConsiderFunction" );
+
+# not yet documented, as not generically implemented
+DeclareAttribute("CanonicalConjugacyClassElement",IsConjugacyClassGroupRep);
```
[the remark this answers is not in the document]
hulpke: Then lets just use CanonicalRepresentativeOfExternalSet.

lib/oprt.gi
```diff
@@ -997,7 +997,7 @@ local orb,d,gen,i,p;
   for p in orb do
     for gen in acts do
       i:=act(p,gen);
-      MakeImmutable(i);
+      if not IsHPCGAP then MakeImmutable(i);fi;
```
fingolfin: Huh, I wondered about this and then found this in a commit message:
> The HPCGAP test otherwise gives a cryptic error message:
> Error, MakeImmutable: Argument has inaccessible subobjects

So that message means that subobject of `i` was either created in a different thread, or for some other reason the current thread has no access to it. Since I can't see from just looking at this what kind of object `i` is when the error arose, it's a bit difficult to analyze further. But in any case, not calling `MakeImmutable` likely just defers the problem. Would be better to get to the root cause.

Is there a simple input one can use to trigger that error (after changing this line and/or the other back) ?

fingolfin: (Just to clarify: I am not expecting @hulpke to investigate this HPC-GAP specific issue, but perhaps @ChrisJefferson or me could have a look, once we know how to trigger it, so I am hoping for a hint on that by @hulpke)

hulpke: @fingolfin It is tests in the standard test suite using automorphism group calculations. I have removed the commit again, so the test will show them up. (I will update this comment later with teh concrete command triggering the error.)
The change triggering the error message is an orbit algorithm on objects that are lists , one of whose entries is a conjugacy class.

hulpke: (ping @fingolfin) Any update on this? I would like to get this chunk of changes merged to not have too much of a discrepancy between my main branch and work branch. Could we keep the HPCGAP workaround for merging and then have the debugging in HPCGAP?

fingolfin: Sorry, I missed your reply :-(.

I had a look. In the quick way to reproduce (thanks, that helped a lot!), the variable `i` is a plain list containing permutations. The problem is that this list is in the "public region", i.e. all threads can *read* it -- but that means we are not allowed to write to it, and we need to write to it to make it immutable (even though that really just means flipping a bit here).

However, it's really odd that a mutable plain list is in the public region. That seems like a bug to me. I'll try to figure out where those come from, it's odd.

fingolfin: Tracked it down, fix is in https://github.com/gap-system/gap/pull/3848. This is a really old bug which just started to have a more visible bad effect in HPC-GAP. All in all, I am actually glad you run into this and we persisted in digging it out :-). Sorry for taking so long to complete this, though.

CC @rbehrends @ChrisJefferson

hulpke: @fingolfin Thank you for the fix. Yes, its good this bug was found.

lib/grp.gd
```diff
@@ -3317,6 +3317,32 @@ KeyDependentOperation( "PRump", IsGroup, IsPosInt, "prime" );
...
+##  <#GAPDoc Label="StructuralSeriesOfGroup">
+##  <ManSection>
+##  <Attr Name="StructuralSeriesOfGroup" Arg='G'/>
+##
+##  <Description>
+##  The structural series of a finite group <A>G</A> is a descending series
+##  of characteristic subgroups which goes throught the derived series of
+##  the solvable radical of $G$, refined into elementary abelian factors, as
+##  well as the socle and the <A>Pker</A> (kernel of the action on socle
+##  compoents) of the radical factor
```
fingolfin:
```suggestion
##  components) of the radical factor
```

lib/morpheus.gi
```diff
@@ -367,300 +367,655 @@ end);
...
+    SetNiceMonomorphism(au,hom);
+    SetIsHandledByNiceMonomorphism(au,true);
+  end;
+
+  # avoid storing list of class elements anew
+  claselms:=function(clas)
+    if HasAsSSortedList(clas) then
+      return AsSSortedList(clas);
+    else
+      return Immutable(Set(Orbit(g,Representative(clas))));
```
```suggestion
      return MakeImmutable(Set(Orbit(g,Representative(clas))));
```
This avoids making a copy of the set, instead the set is marked as immutable.

lib/morpheus.gi
```diff
@@ -367,300 +367,655 @@ end);
...
+    SetIsFinite(au,true);
+  fi;
 
-  elif IsFpGroup(g) then
-    # no sane person should work with automorphism groups of fp groups, but
-    # if someone really does this is a shortcut that avoids canonization
-    # issues.
+  if IsFpGroup(g) then
+    # no sane person should work with automorphism groups of fp groups, as
+    # this is doubly inefficient,but if someone really does, this is a
```
```suggestion
    # this is doubly inefficient, but if someone really does, this is a
```

lib/morpheus.gi
```diff
@@ -367,300 +367,655 @@ end);
...
+    #shortcut that avoids canonization issues.
```
```suggestion
    # shortcut that avoids canonization issues.
```

lib/morpheus.gi
```diff
@@ -367,300 +367,655 @@ end);
...
+    if Length(c)>0 
+      # no need to process classes that will not improve
+      and Size(c[1])<bestdeg-offset then
+      if Length(c)>1 then
+        if Size(c[1])<250 and SIZE_OBJ(Representative(c[1]))*Size(c[1])*Length(c)<10^7 then
+          c:=List(c,claselms);
+          Sort(c);
+          Info(InfoMorph,4,"Element sets");
+          if actbase<>fail then
+            of:=Orbits(auo,c,OnSets); # not necc. domain
```
fingolfin: What's "necc."?

lib/morpheus.gi
```diff
@@ -367,300 +367,655 @@ end);
...
+    r:=[Order(Representative(cl[i])),Size(cl[i])];
+    u:=Filtered([1..Length(cl)],
+      x->[Order(Representative(cl[x])),Size(cl[x])]=r and not Representative(cl[x]) in ser[pos]);
+    c:=cl{u};
+    cl:=cl{Difference([1..Length(cl)],u)};
+    if Length(c)>0 
+      # no need to process classes that will not improve
+      and Size(c[1])<bestdeg-offset then
+      if Length(c)>1 then
+        if Size(c[1])<250 and SIZE_OBJ(Representative(c[1]))*Size(c[1])*Length(c)<10^7 then
```
fingolfin: Huh, why is the kernel function SIZE_OBJ used here?

tst/teststandard/opers/AutomorphismGroup.tst
```diff
@@ -39,6 +39,8 @@ gap> StructureDescription(H);
 #
 gap> g:=PerfectGroup(IsPermGroup,30720,1);;
 gap> h:=g^(1,153);;
+gap> AutomorphismGroup(g);;
+gap> AutomorphismGroup(h);;
```
fingolfin: So, these just test that the computations do not run into an error? But we don't even test whether the result is `fail` or a group. Can't one perhaps at least output, say, the size of the resulting groups (if they are known), or test somehow else that the two computed groups are plausible? If that's not easily possible (e.g. computing the sizes takes too long), perhaps at least  this could be changed to `IsGroup(AutomorphismGroup(g));`?

hulpke: No. These are intended to pullout calculations of the subsequent `IsomorphismTest` earlier ,to reduce the chance of a timeout for the automated tests.

lib/morpheus.gi
hulpke: SIZE_OBJ might not be the right operation. I want to store elements if it does not require too much memory.

lib/morpheus.gi
hulpke: neccessarily

lib/morpheus.gi
hulpke: Thanks. Will change

tst/teststandard/opers/AutomorphismGroup.tst
hulpke: @fingolfin I've replaced it with `MemoryUsage`.

lib/morpheus.gi
fingolfin: Ah, I see. Well `SIZE_OBJ` only measures the immediate storage of the object - which is fine if you know they are e.g. permutations. But for e.g. matrices it underestimates the memory usage -- for that `MemoryUsage` can be used (which goes recursively through all subobjects)

fingolfin: Ah I just noticed that you already changed it to `MemoryUsage`, great :-)

fingolfin:
Let's wait till the current Travis tests have cycled through, then from my POV we can merge this PR (assuming @hulpke is happy to do so?)

PR #3787, opened 2019-12-19
Brauer table of a permuted isoclinic table

ThomasBreuer:
Consider an ordinary character table t that arises from a table t1
by a class permutation pi, say, and assume t1 arises
as an isoclinic table of another table t2.
If the p-modular table of t2 is known then the p-modular table
of t can be derived from it.

Up to now, GAP ran into an error when the permutation pi did not
act on the set of p-regular classes.
(Since nobody has complained about this, apparently it did not happen.)


fingolfin:

PR #3788, opened 2019-12-19
Travis: add libcdd-dev to deps for testpackages

fingolfin:
This is meant to unbreak Travis. If/once it works, we should backport it to stable-4.11

EDIT: but *PLEASE* don't be too quick in merging this! I'd rather merge PR #3789 first (as explained there), then PR #3777, and only then this PR here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ matrix:
           - libreadline-dev
           - zlib1g-dev
           - libboost-dev            # for NormalizInterface
+          - libcdd-dev              # for CddInterface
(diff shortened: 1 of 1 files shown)


wilfwilson:
This fixes the `testpackages` jobs. Happy to hold off on merging as per Max's plan (although I couldn't merge anyway, since the Travis tests will fail thanks to the Mac job).

fingolfin:
Backported as part of commit ceb7d4912a16347aa87027ac3252ab1f8499783b

PR #3790, opened 2019-12-19
Improvements to simple group generation/iteration

hulpke:
#Description

Added generators for 3D4(q), Order formula for G(q)
Better handling of outer automorphism structure for D(q) in DataAboutSimpleGroups
Cleanup in parsing names of type O+-

## Text for release notes 

Added further information to the library of simple groups

### DIFF
--- a/grp/simple.gi
+++ b/grp/simple.gi
@@ -233,7 +233,44 @@ local p,f,z,G,o;
       [0,0,0,0,1,0,0]]);
   fi;
 
+  SetSize(G,q^6*(q^6-1)*(q^2-1));
   return G; 
+end);
+
+# generators from 
+# Howlett, R. B.(5-SYD-SM); Rylands, L. J.; Taylor, D. E.(5-SYD-SM)
+# Matrix generators for exceptional groups of Lie type. 
+# J. Symbolic Comput. 31 (2001), no. 4, 429–445. 
+# Note that Magma uses slightly different generators
+BindGlobal("Chevalley3D4",function(q)
+local f,mu,m1,m2,x,n,o;
+  f:=GF(q^3);
+  o:=One(f);
+  mu:=PrimitiveRoot(f);
+  m1:=DiagonalMat([mu^(q^2),mu^(-q^2),mu^(q+1),mu^(q-1),
+                   mu^(-q+1),mu^(-q-1),mu^(q^2),mu^(-q^2)]);
+  x:=IdentityMat(8,f);
+  x[1,2]:=o;
+  x[3,4]:=o;
+  x[3,5]:=o;
+  x[3,6]:=o;
+  x[4,6]:=o;
+  x[5,6]:=o;
+  x[7,8]:=o;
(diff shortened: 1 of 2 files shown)


grp/simple.gi
```diff
@@ -233,9 +233,46 @@ local p,f,z,G,o;
       [0,0,0,0,1,0,0]]);
   fi;
 
+  SetSize(G,q^6*(q^6-1)*(q^2-1));
   return G; 
 end);
 
+# generators from
```
wilfwilson: Thanks for the comment, it's handy.

grp/simple.gi
```diff
@@ -233,9 +233,46 @@ local p,f,z,G,o;
       [0,0,0,0,1,0,0]]);
   fi;
 
+  SetSize(G,q^6*(q^6-1)*(q^2-1));
```
wilfwilson: Just curious: Is this a 'well known' thing?

grp/simple.gi
```diff
@@ -860,6 +904,7 @@ local nam,e,EFactors,par,expo,prime,result,aut,i;
 	return dd;
       else
 	Error("mixed triality not yet done");
+        return false;
```
wilfwilson: You could also do `ErrorNoReturn` here instead of the `Error` and then `return false`. There's no problem with this way, though. (Unless I'm missing something.)

wilfwilson:
(Note that the Travis tests will currently fail (see #3788) so we'll eventually ask you to rebase on `master` once the problems in the `master` branch are fixed.)

grp/simple.gi
hulpke: @wilfwilson Yes. (Though I guess the "well" is relative, given that you ask). The order of every Chevalley group is a polynomial in the field order. This follows from the general theory and is proven e.g. in Carter's (simple groups) book. I copied the formula from a table in the ATLAS preface.

wilfwilson: Thanks for the answer, that’s good to know. Despite my interest in finite simple groups, I’m not at all well-versed in their details, so the fact that I didn’t know shouldn’t weigh in on the “well-known”-ness of this fact. My algebra education focused more on semigroups, rings, fields, and infinite groups, and not on the finite ones. 🙂 I still lack quite a few of the basics.

fingolfin:

fingolfin:
Actually, there is also a test "failure", which however is easy enough (and satisfying ;-) to fix;
```
testing: /cygdrive/c/projects/gap/tst/testinstall/opers/SimpleGroup.tst
########> Diff in /cygdrive/c/projects/gap/tst/testinstall/opers/SimpleGroup.tst:292
# Input is:
SimpleGroup("3D(4,4)");
# Expected output:
Error, Can't do yet
# But found:
3D4(4)
########
    9360 ms (205 ms GC) and 129MB allocated for opers/SimpleGroup.tst
```

fingolfin:
Backported to stable-4.11 in commit c4832e174fb457e52348e7cd5c36d260b1067ed6

PR #3793, opened 2019-12-20
Handle out of memory errors better

### LINKED ISSUE #3297: GAP terminates session when trying to calculate large coset space
The calculation
```G:=SymmetricGroup(14);
Sym( [ 1 .. 14 ] )
gap> H:=Group((1,2),(2,3));
Group([ (1,2), (2,3) ])
gap> RightCosets(G,H);
```
causes GAP to terminate with the message `gap: cannot extend the workspace any more!` and to delete the history of the current session.

### LINKED ISSUE #3786: System Panic when working with large solvable group
[This was reported by Heiko Dietrich]

### Observed behaviour

    gap> code:= rec( code := 4036976711286048149259,
    >                extdim := [  ], isFrattiniFree := true,
    >                isUnique := true,
    >             order := 4037839681938311097109 );
    gap> F := PcGroupCodeRec( code );
    <pc group of size 4037839681938311097109 with 2 generators>
    gap> Factors(4037839681938311097109);
    [ 4679, 862970652262943171 ]
    gap> IrreducibleModules(F,GF(4679),1);
     #G  FULL  691935/  68330kb live  4606596/ 263344kb dead  Panic in src/sysmem.c:654: cannot extend the workspace any more!

### Expected behaviour

A clearer error message, possibly in break loop

ChrisJefferson:
Fixes #3297
Fixes #3786 

This is a WIP PR to try to produce useful error messages when GAP runs out of memory because of a single large allocation.

I'm interested if anyone can make GAP do "bad" things with this PR loaded.

Example:

```
gap> l := [];
[  ]
gap> l[2^59] := 1;
Error, Cannot allocate 4611686018427387912 bytes: cannot extend the workspace any more!!!!
not in any function at *stdin*:3
brk> 
Lgap> List(SymmetricGroup(15));
Error, Cannot allocate 10461394944008 bytes: cannot extend the workspace any more!! in
  Length( list ) at /home/caj/files/reps/gap/gap/lib/list.gi:560 called from 
ShallowCopy( Enumerator( C ) ) at /home/caj/files/reps/gap/gap/lib/coll.gi:689 called from
CallFuncList( ListOp, arg ) at /home/caj/files/reps/gap/gap/lib/coll.gi:677 called from
<function "List">( <arguments> )
```


ChrisJefferson:
From #3786 :

```
gap> code:= rec( code := 4036976711286048149259,
>                extdim := [  ], isFrattiniFree := true,
>                isUnique := true,
>             order := 4037839681938311097109 );
rec( code := 4036976711286048149259, extdim := [  ], isFrattiniFree := true, isUnique := true, order := 4037839681938311097109 
 )
gap> F := PcGroupCodeRec( code );
<pc group of size 4037839681938311097109 with 2 generators>
gap> Factors(4037839681938311097109);
[ 4679, 862970652262943171 ]
gap> IrreducibleModules(F,GF(4679),1);
Error, Cannot allocate 6903765218103545376 bytes: cannot extend the workspace any more!! in
  EmptyPlist( Length( C ) ) at /home/caj/files/reps/gap/gap/lib/coll.gi:659 called from 
List( [ 1 .. r ], function ( x )
      return f ^ (x - 1);
  end ) at /home/caj/files/reps/gap/gap/lib/grppcrep.gi:419 called from
InitAbsAndIrredModules( r, F, dim ) at /home/caj/files/reps/gap/gap/lib/grppcrep.gi:572 called from
AbsAndIrredModules( G, F, dim ) at /home/caj/files/reps/gap/gap/lib/grppcrep.gi:617 called from
<function "IrreducibleModules generic method for groups with pcgs">( <arguments> )
 called from read-eval loop at *stdin*:9
brk> 
gap> 
```

src/gasman.c
```diff
@@ -1099,6 +1099,27 @@ static void MaybeMoveBags(void)
 
 #endif
 
+/****************************************************************************
+**
+*F  OutOfMemoryPanic(size) . . . . . . . . . .  handle running out of memory
```
fingolfin:
```suggestion
*F  OutOfMemoryPanic(size, reason) . . . . . . . handle running out of memory
```

fingolfin: The comment is also in the wrong spot, i.e., not directly in front of the function

src/gasman.c
```diff
@@ -1099,6 +1099,27 @@ static void MaybeMoveBags(void)
...
+
+void OutOfMemoryPanic(UInt size, const char * reason)
+{
+    // If we could not satisfy an allocation, and that allocation was large
+    // (128MB at present), then GAP can reasonably continue.
+    if (size > (1LL << 27) && TooLargeAllocFunc) {
+        TooLargeAllocFunc(size, reason);
+    }
+    Panic("%s", reason);
+}
```
fingolfin:
```suggestion
}

```

src/gasman.c
```diff
@@ -1099,6 +1099,27 @@ static void MaybeMoveBags(void)
...
+
+void OutOfMemoryPanic(UInt size, const char * reason)
+{
+    // If we could not satisfy an allocation, and that allocation was large
+    // (128MB at present), then GAP can reasonably continue.
+    if (size > (1LL << 27) && TooLargeAllocFunc) {
+        TooLargeAllocFunc(size, reason);
+    }
+    Panic("%s", reason);
```
fingolfin: Note that Panic is a macro that is careful to record the site it was called at. It would be nice if `OutOfMemoryPanic` did that as well (it can then proceed to call `Panic_`). I.e., rename it to `OutOfMemoryPanic_` and then add a macro `OutOfMemoryPanic` similar to the macro `Panic`.

Then we can also do away with the "!!!" hack in the various panic messages.

src/gasman.c
```diff
@@ -1099,6 +1099,27 @@ static void MaybeMoveBags(void)
...
+**
+*/
+
+static TooLargeAllocationHandlerFunc TooLargeAllocFunc;
+void SetTooLargeAllocationHandler(TooLargeAllocationHandlerFunc f)
+{
+    TooLargeAllocFunc = f;
+}
+
+void OutOfMemoryPanic(UInt size, const char * reason)
```
fingolfin: Perhaps a batter name would be just `HandleOutOfMemory` or `SignalOutOfMemory` or just `OutOfMemory`, given that the function does *not* `Panic` in general, only if we really are out of memory...?

src/system.h
```diff
@@ -422,6 +422,21 @@ void Panic_(const char * file, int line, const char * fmt, ...) NORETURN
 #define Panic(...) \
     Panic_(__FILE__, __LINE__, __VA_ARGS__)
 
+void OutOfMemoryPanic(UInt size, const char * reason);
+
+/****************************************************************************
+**
+*F SetTooLargeAllocationHandler
+**
+** Pass a function to be called when a block of memory which is too large
```
fingolfin:
```suggestion
**  Pass a function to be called when a block of memory which is too large
```

src/system.h
```diff
@@ -422,6 +422,21 @@ void Panic_(const char * file, int line, const char * fmt, ...) NORETURN
 #define Panic(...) \
     Panic_(__FILE__, __LINE__, __VA_ARGS__)
 
+void OutOfMemoryPanic(UInt size, const char * reason);
```
fingolfin: There should be a comment before this function, too! In fact you have one near the implementation. Personally, I think it's more important to have these documentation comments in the header file. But regardless of what one considers better, it's odd to have the comment for one function (`OutOfMemoryPanic`) only in the .c file, and the comment for the next one (`SetTooLargeAllocationHandler`) only in the `.h` file.

src/system.h
```diff
@@ -422,6 +422,21 @@ void Panic_(const char * file, int line, const char * fmt, ...) NORETURN
...
+
+/****************************************************************************
+**
+*F SetTooLargeAllocationHandler
```
fingolfin:
```suggestion
*F  SetTooLargeAllocationHandler( handler )
```

src/system.h
```diff
@@ -422,6 +422,21 @@ void Panic_(const char * file, int line, const char * fmt, ...) NORETURN
...
+**
+** Pass a function to be called when a block of memory which is too large
+** is requested.
```
fingolfin:
```suggestion
**  is requested.
```

src/system.h
fingolfin: In any case, I wonder: Do we really need this `SetTooLargeAllocationHandler` stuff at all? It seems like generalization just for its own sake. Without it, this PR becomes substantially simpler, and I can't think of any applications for it right now. At the same time, it's one more API that in principle now *all* our GCs would have to support (this PR only deals with GASMAN).

tst/testspecial/run_gap.sh
```diff
@@ -14,7 +14,7 @@ gfile="$2"
 #    so the output is usable on other machines
 # 4) Set lower and upper memory limits, for consistency
 GAPROOT=$(cd ../..; pwd)
-( echo "LogTo(\"${outfile}.tmp\");" ; cat "$gfile" ; echo "QUIT;" ) |
+( echo "LogTo(\"${gfile}.tmp\");" ; cat "$gfile" ; echo "QUIT;" ) |
     "$gap" -r -A -b -m 256m -o 512m -x 800 2>/dev/null >/dev/null
-sed "s:${GAPROOT//:/\\:}:GAPROOT:g" < "${outfile}.tmp"
-rm "${outfile}.tmp"
+sed "s:${GAPROOT//:/\\:}:GAPROOT:g" < "${gfile}.tmp"
+rm "${gfile}.tmp"
```
fingolfin: Why this change?

src/sysmem.h
```diff
@@ -33,6 +33,9 @@
 */
 extern Int SyStorMax;
 extern Int SyStorOverrun;
+// A constant which can be assigned to SyStorOverrun to denote the overrun
+// has already been printed.
+enum { SY_STOR_OVERRUN_HANDLED = 999 };
```
fingolfin: So, `SyStorOverrun` has three valid values: 0, -1 and 999. For one of them, an enum is defined. If you already go for an enum (a good idea), why not define all allowed values, and then turn `SyStorOverrun` into a variable of that enum type? I.e.
```
enum StorOverrunType { ... };
extern enum StorOverrunType SyStorOverrun;
```

src/gasman.c
```diff
@@ -2141,7 +2165,8 @@ UInt CollectBags (
 
     // Check if this allocation would even fit into memory
     if (SIZE_MAX - (size_t)(sizeof(BagHeader) + size) < (size_t)AllocBags) {
-        return 0;
+        ret = 0;
+        goto cleanup;
```
fingolfin:
```
src/gasman.c: In function ‘CollectBags’:
src/gasman.c:2169:9: error: jump skips variable initialization [-Werror=jump-misses-init]
         goto cleanup;
         ^~~~
src/gasman.c:2315:1: note: label ‘cleanup’ defined here
 cleanup:
 ^~~~~~~
src/gasman.c:2173:11: note: ‘stopBags’ declared here
     Bag * stopBags = AllocBags + WORDS_BAG(sizeof(BagHeader)+size);
           ^~~~~~~~
```

src/gasman.c
```diff
@@ -2302,8 +2331,8 @@ UInt CollectBags (
 
     GAP_ASSERT(SanityCheckGasmanPointers());
 
-    /* return success                                                      */
-    return 1;
+    /* success                                                      */
```
fingolfin: Perhaps you meant to remove the word "success"? :-). In any case, I'd just remove the whole pointless comment
```suggestion
```

src/system.h
fingolfin: Isn't `OutOfMemoryPanic` strictly specific to GASMAN? It should be inside an `#ifdef USE_GASMAN` then -- and perhaps even be declared in `gasman.h` or `sysmem.h` or even `gasman_intern.h`, no?

tst/testspecial/run_gap.sh
ChrisJefferson: Oh, this makes run_gap threadsafe (outfile isn't actually set to anything, so we just use the filename .tmp). I tend to run run_gap.sh in parallel.

src/sysmem.h
ChrisJefferson: Yes, this generally needs cleaning up. This was just the start of a hack when I realised I needed a way of telling SyInterrupt that I'd already handled the memory overflow (as I thought that was the easiest way of handling it). It might be I can just clear it instead, then no need this.

src/gasman.c
ChrisJefferson: I think I just somehow broke the comment

ChrisJefferson:
In terms of the "API", the big picture reason for it is to get around the fact that, at the moment, GASMAN includes a bare minimum of GAP headers, so it can't call Error easily. Overall, I quite like this, as it stops calls tol GAP functions sneaking into gasman.c (which would in most cases be a horrible thing to do). 

I'm happy with suggestsions on the bestbest way to let gasman call back into GAP.

tst/testspecial/low-mem-plist-1.g.out
```diff
@@ -0,0 +1,17 @@
+gap> l := [];
+[  ]
+gap> l[2^50] := 1;
+Error, Cannot allocate 9007199254741000 bytes: cannot extend the workspace any more!!!!
```
fingolfin: Won't this test break in 32bit mode, because GAP only allows indices which are immediate integers? At least I get this in my 64bit GAP:
```
gap> l:=[];;l[2^100]:=1;
Error, List Assignment: <pos> must be a positive small integer (not a large positive integer)
```

ChrisJefferson:
Thanks for the review. This PR is certainly in a terrible mess (I was mainly figuring out if solving the problem was even possible, and if GAP would go crazy and crash). If people seem to approve of the general principle, then I'm happy to clean it up.

tst/testspecial/low-mem-plist-1.g.out
ChrisJefferson: Yes, testing in 32-bit is going to be a pain, and probably require splitting testspecial so we can have special 32-bit tests.

src/stats.c
```diff
@@ -1073,10 +1073,13 @@ static UInt ExecIntrStat(Stat stat)
     /* and now for something completely different                          */
 #ifdef USE_GASMAN
     if ( SyStorOverrun != 0 ) {
-      SyStorOverrun = 0; /* reset */
-      ErrorReturnVoid(
-  "reached the pre-set memory limit\n(change it with the -o command line option)",
-        0, 0, "you can 'return;'" );
+        Int SyStorValue = SyStorOverrun;
+        SyStorOverrun = 0; /* reset */
+        if (SyStorValue != SY_STOR_OVERRUN_HANDLED) {
```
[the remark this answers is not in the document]
ChrisJefferson: So, here is the general picture (as I see it)

At present, IntrExecStat always prints something, either "out of memory" (if SyStorOverrun is non-zero), or "ctrl+c pressed" (if SyStorOverrun is 0).

It's hard to avoid having SyStorOverrun set before we realise our allocation request was too big, so the interrupt is set. At this point we have two choices:

1) Unset to interrupt before it is ever called
2) Let it be called, but don't print anything.

(1) seems fiddly, and I'm worried about "unsetting" it incorrectly, so I thought doing (2) seemed like a better plan.

ExecIntrStat is more the function I'm interested in, because if I try printing an Error, that error ends up being interuppted by ExecIntrStat as soon as it starts.

However, now I've written this all down, I think I have a better way of handling all this :)

ChrisJefferson:
So, I think this PR is now ready for reviewing. I'm not going to remove the "do not merge" yet, because it's been massively changed since the last review, so it should probably get checked.

src/gasman.c
```diff
@@ -2302,8 +2332,8 @@ UInt CollectBags (
...
+    return ret;
```
fingolfin: With https://github.com/gap-system/gap/pull/3797 you'd only have to remove the `if (done == 2)` check, and instead change the return here to `return (done == 1)` or `return done != 2` or something like that. And no need for `ret`, and most of the other changes in `CollectBags` (except for the change to that one `SyAllocBags` call, of course)

src/error.c
```diff
@@ -622,6 +623,15 @@ void AssertionFailure(void)
     ErrorReturnVoid("Assertion failure", 0, 0, "you may 'return;'");
 }
 
+// This function is called from gasman.c, when a large allocation fails
+void TooLargeAllocError(UInt bytes, const char * reason)
+{
+    // If GAP is planning to print a SyStorOverrun message, mark it as
```
fingolfin: GAP cannot plan. How about this:
```suggestion
    // If a SyStorOverrun message is scheduled, mark it as already printed
```

src/sysmem.h
```diff
@@ -32,7 +32,16 @@
 **  This is used in the function 'SyAllocBags' below.
 */
 extern Int SyStorMax;
-extern Int SyStorOverrun;
+
+// SyStorOverrun is used for reporting to the user if more than SyStorMax
+// memory has been allocated. It is used as there is more than once place
+// this can be reported to the use.
```
fingolfin: I don't understand the final sentence. Actually, re-reading, the first sentence also is a bit odd... SyStorOverrun is not reporting to the user itself after all. How about something like this instead:
> SyStorOverrun tracks whether an allocation exceeding the memory limit specified by
> SyStorMax is exceeded. Several places in the GAP code base check its values and
> will take appropriate actions if it is set to a value different from SY_STOR_OVERRUN_CLEAR.

tst/testspecial/run_all.sh
```diff
@@ -6,9 +6,20 @@ set -ex
 # (so for out-of-tree builds, builddir and not srcdir)
 GAPDIR=${GAPDIR:-../..}
 
-retvalue=0
 gap="$GAPDIR/bin/gap.sh"
-for gfile in *.g; do
+
+
+if ./run_gap.sh "${gap}" "support/check_bytes_per_variable.g" | grep "8"; then
```
fingolfin: Alternatively could get rid of `check_bytes_per_variable.g` again and do this::
```suggestion
if "${gap}" -A -b -c 'QUIT_GAP(GAPInfo.BytesPerVariable - 8);'; then
```

tst/testspecial/run_all.sh
```diff
@@ -17,5 +28,5 @@ for gfile in *.g; do
         rm -f "${gfile}.bad"
     fi;
 done;
-exit ${retvalue}
```
fingolfin: Why remove this? If we really want to loose it, please also remove `retvalue`, which is now unused.

src/stats.c
```diff
@@ -1127,10 +1130,15 @@ void ClearError ( void )
...
-          Pr("the maximum is now enlarged to %d kB.\n", (Int)SyStorMax, 0);
+        if (SyStorOverrun != SY_STOR_OVERRUN_CLEAR) {
+            SyStorEnum value = SyStorOverrun;
+            SyStorOverrun = SY_STOR_OVERRUN_CLEAR; /* reset */
+            if (value == SY_STOR_OVERRUN_TO_REPORT) {
+                Pr("GAP has exceeded the permitted memory (-o option),\n", 0,
+                   0);
+                Pr("the maximum is now enlarged to %d kB.\n", (Int)SyStorMax,
+                   0);
+            }
```
fingolfin: Or, just do this:
```C
if (SyStorOverrun == SY_STOR_OVERRUN_TO_REPORT) {
  Pr(...);
}
SyStorOverrun = SY_STOR_OVERRUN_CLEAR;
```

src/stats.c
```diff
@@ -1072,11 +1072,14 @@ static UInt ExecIntrStat(Stat stat)
...
-    if ( SyStorOverrun != 0 ) {
-      SyStorOverrun = 0; /* reset */
-      ErrorReturnVoid(
-  "reached the pre-set memory limit\n(change it with the -o command line option)",
-        0, 0, "you can 'return;'" );
+    if (SyStorOverrun != SY_STOR_OVERRUN_CLEAR) {
+        SyStorEnum value = SyStorOverrun;
+        SyStorOverrun = SY_STOR_OVERRUN_CLEAR; /* reset */
+        if (value == SY_STOR_OVERRUN_TO_REPORT) {
+            ErrorReturnVoid("reached the pre-set memory limit\n(change it "
```
fingolfin: Can we do this, as in: does `clang-format` accept it?
```suggestion
            ErrorReturnVoid("reached the pre-set memory limit\n"
```

src/stats.c
```diff
@@ -1072,11 +1072,14 @@ static UInt ExecIntrStat(Stat stat)
...
+                            "with the -o command line option)",
```
fingolfin:
```suggestion
                            "(change it with the -o command line option)",
```

DominikBernhardt:
Just a small remark: Maybe you could mention #3297 here? I would be fine with closing this issue at some point.

src/stats.c
ChrisJefferson: My reason for doing it the way I've done it was because I was worried about calling other functions before I'd set SyStorOverrun. However, Pr doesn't call into GAP code, so that shouldn't be a problem.

fingolfin:
@ChrisJefferson this now needs to be rebased due to the `gasman.c` changes, but it shouldn't be too hard. I can also do that, if you want me to (after all, it's my "fault")

ChrisJefferson:
I'd already started rebasing, I think all the issues are now fixed (but should obviously let all the tests cycle)

src/gasman.c
```diff
@@ -114,6 +114,7 @@
 #include "gasman.h"
 #include "gasman_intern.h"
 
+#include "error.h"
```
fingolfin: But if you include `error.h` now, why do we still need `TooLargeAllocError` declared in a separate source file?

src/gasman.c
```diff
@@ -1100,6 +1101,22 @@ static void MaybeMoveBags(void)
...
+**
+*F  ReportOutOfMemory(size, reason) . . . . . .  handle running out of memory
+**
+*/
+// ReportOutOfMemory is a macro so the correct filename/line is passed to
+// panic. The '1ll << 27' check ensures we only try to recover when a large
+// allocation fails.
+#define ReportOutOfMemory(size, reason)                                      \
+    {                                                                        \
+        if (size > (1LL << 27)) {                                            \
```
fingolfin:
```suggestion
        if (size > (UInt)1 << 27)) {                                            \
```

src/gasman.c
```diff
@@ -1100,6 +1101,22 @@ static void MaybeMoveBags(void)
 
 #endif
 
+/****************************************************************************
+**
+*F  ReportOutOfMemory(size, reason) . . . . . .  handle running out of memory
+**
+*/
+// ReportOutOfMemory is a macro so the correct filename/line is passed to
```
fingolfin: It's a bit jarring to see this mix of comment styles. Why not just turn the `//` into `**` and move the `*/` below, then it fits? Or else, just use `//` for the whole comment.

src/stats.c
```diff
@@ -1127,10 +1130,14 @@ void ClearError ( void )
         }
 #ifdef USE_GASMAN
         /* and check if maximal memory was overrun */
-        if ( SyStorOverrun != 0 ) {
-          SyStorOverrun = 0; /* reset */
-          Pr("GAP has exceeded the permitted memory (-o option),\n", 0, 0);
-          Pr("the maximum is now enlarged to %d kB.\n", (Int)SyStorMax, 0);
+        if (SyStorOverrun != SY_STOR_OVERRUN_CLEAR) {
```
fingolfin: Can't you just remove this line?
```suggestion
```
Well, and then also the matching `}`, and the indentation needs to be adjusted.

src/gasman.c
```diff
@@ -2302,8 +2320,7 @@ UInt CollectBags (
 
     /* * * * * * * * * * * * * * * check phase * * * * * * * * * * * * * * */
     done = CollectBags_Check(size, nrBags);
-    if (done == 2)
-        return 0;
+    // done: 0 - finished, 1 - not finished, 2 - out of memory
```
fingolfin: That's wrong, correct:
```suggestion
    // done: 0 - not finished, 1 - finished, 2 - out of memory
```
or even better:
```suggestion
    // done: 0 - not done, 1 - done with allocation, 2 - done because out of memory
```

But really, isn't the name `done` self-explanatory? In PR #3803 I refactored this loop as well, replacing the `goto` construct by a loop: `do { ... } while (!done)` which says it quite well, IMHO: "repeat while not done"

src/gasman.c
```diff
@@ -1478,7 +1495,8 @@ UInt ResizeBag (
         // check that enough storage for the new bag is available
         if (SpaceBetweenPointers(EndBags, CONST_PTR_BAG(bag)) < WORDS_BAG(new_size)
               && CollectBags( new_size-old_size, 0 ) == 0 ) {
-            Panic("cannot extend the workspace any more!!!!!");
+            ReportOutOfMemory(new_size,
+                             "cannot extend the workspace any more!!!");
```
wilfwilson: Are the different numbers of exclamation marks in the three versions of “cannot extend the workspace any more” chosen for a deliberate reason? I can see that it could help us debug. Note that there is no test coverage (I think) for this 3 exclamation mark version.

fingolfin: Yes, that's why they are there (although with `ReportOutOfMemory` we now print the line number where the warning is triggered, so this trick should be less useful.

PR #3794, opened 2019-12-20
kernel: delete all 'return success' comments

fingolfin:
They are completely pointless


fingolfin:
@ChrisJefferson is this OK by you?

ChrisJefferson:
Yes, I downmark students when they write these kinds of comments :)

PR #3796, opened 2019-12-20
Remove -B option for overriding the architecture

fingolfin:
The documentation for it claimed that "[u]nder rare circumstances, it
may be necessary to override this name" but I simply can't come up
with any that are not contrived and highly dangerous, and which are not
more easily resolved by renaming or copying a few `bin` subdirectories
inside packages.

That of course just means my imagination might be limited. I wonder if
some of the people who were around (?) when this was added might
know the original motivation and whether there could be circumstances
now where this would still be useful.

### DIFF
--- a/doc/ref/run.xml
+++ b/doc/ref/run.xml
@@ -80,16 +80,6 @@ which can be useful for debugging or testing.
 The needed packages (and their needed packages, and so on)
 are loaded in any case.
 </Item>
-<Mark><Index Key="-B"><C>-B</C></Index>
-<C>-B </C><A>architecture</A></Mark>
-<Item>
-Executable binary files that form part of &GAP; or of  a &GAP;  package
-are kept in a subdirectory of the <F>bin</F> directory within  the  &GAP; or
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
I have half a memory of discussing this years ago with @frankluebeck , so I will mention him here just so he is notified, in case he wants to comment.

Many apologises if I misremember!

fingolfin:
Yeah I already requested reviews from @frankluebeck @ThomasBreuer @stevelinton in the hopes that they can shed light on this.

I think the option may have been more useful with the old build system that was in use until GAP 4.8 ?

fingolfin:
@frankluebeck @ThomasBreuer @stevelinton perhaps you missed this because of Christmas: this PR removes the `-B` option for overriding the architecture. Any concerns about this?

frankluebeck:
I vaguely remember that the -B option was used in earlier times in network installations in inhomogeneous networks (machines with different architectures, different optimizations for some machines,...). Later I had introduced a mechanism using the CONFIGNAME variable.
With the current build system both are probably no longer needed.

PR #3797, opened 2019-12-20
gasman: split CollectBags into several functions

fingolfin:
By splitting this huge function into several smaller ones, it
gets a bit easier to reason about each phase of the GC on its own.

This is motivated by a tiny part of PR #3793 which has to modify `CollectBags` to always perform the "cleanup" part. After this PR here, that modifications becomes trivial. That also means that the two PRs conflict, but I hope it'll be trivial to resolve that. And I think this refactoring is a win on its own.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -1838,66 +1838,12 @@ to ensure that IsWeakDeadBag works correctly.
 static Bag * NewWeakDeadBagMarker = (Bag *)(1000*sizeof(Bag) + 1);
 static Bag * OldWeakDeadBagMarker = (Bag *)(1001*sizeof(Bag) + 1);
 
-
-
-UInt CollectBags (
-    UInt                size,
-    UInt                full )
+static UInt CollectBags_Mark(void)
 {
     Bag                 first;          /* first bag on a linked list      */
-    Bag *               p;              /* loop variable                   */
-    Bag *               dst;            /* destination in sweeping         */
-    Bag *               src;            /* source in sweeping              */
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Just to say, I'm not ignoring this, I'm just reviewing it and testing it carefully. That takes a while.

fingolfin:
No need to hurry, take your time :-). I really appreciate you double checking.

In case it helps, here is how I created the PR; it started as an experiment, and I worked in a rather simple minded fashion: I took the first phase, and moved the code entirely into a new `void foo(void)` function, and inserted a call to the new function in lieu of the (re)moved code. Then I fixed the compiler errors -- mostly that meant *moving* variable declarations (thus proving that no hidden data dependency was lost). In a few cases, I had to *copy* variable declarations (e.g. `first`, `i`); there I tried to verify manually that there were no hidden data dependencies.  I added function parameters and return values as needed.

Once the first phase was done, I repeated this separately for each phase. In the end, I squashed all together because I realized that this actually resulted in a fairly small diff. I briefly thought about moving some more code in CollectBags to new functions (the code before / after the phases) to minimize the diff even more, but that felt rather artificial in the end, plus I think `CollectBags` is small enough now to be easy to understand and reason about.

This all started out as a crazy experiment to figure out the interdependencies of the phases; I didn't actually expect it to work out *this* well :-)

ChrisJefferson:
It seems good, I'm just so nervous about anything to do with gasman I like to run all the tests, with all the various super-debugging flags on, just to see if anything falls out (nothing yet).

PR #3798, opened 2019-12-21
kernel: add common.h header containing parts of system.h

fingolfin:
This PR moves parts of `system.h` into a new header file `common.h`. The idea is that ultimately, `common.h` will only contain things that are "invariant" and do not depend on `config.h`, while all truly system and config specific things are in `system.h`.

This in turn is partially motivated by the desire to implement proper `make install` support: one limiting factor for that is that one shouldn't just install a `config.h` header generated by autoconf, at least not without carefully considering what is in it.

Anyway, this PR is not quite there yet, because right now `common.h` still needs to include `config.h` because `debug.h` needs it. I plan to change that, and introduce a separate `gap_config.h` header which only contains GAP specific "configuration" (e.g. it will contain `GAP_KERNEL_MAJOR_VERSION` and `USE_GASMAN`; but not `HAVE_VFORK` nor `PACKAGE_BUGREPORT`).

This is a PR I had sitting on my disk for close to a year now, progress has been stalled. I am also oscillating between thinking this is a good idea or a bad idea or simply doesn't matter. So feel free to shoot it down ;-).


src/common.h
```diff
@@ -0,0 +1,199 @@
...
+**
+*/
+enum { BIPEB = sizeof(UInt) * 8, LBIPEB = (BIPEB == 64) ? 6 : 5 };
+
+
+/****************************************************************************
+**
+*T  StructInitInfo  . . . . . . . . . . . . . . . . . module init information
+**
+**  This is just a forward declarations so that it can be used in various
```
DominikBernhardt:
```suggestion
**  This is just a forward declaration so that it can be used in various
```

src/scanner.h
```diff
@@ -258,7 +258,7 @@ typedef struct {
 
 EXPORT_INLINE int IsIdent(char c)
 {
-    return IsAlpha(c) || c == '_' || c == '@';
```
ChrisJefferson: Unless your plan is to remove IsAlpha (which doesn't seem to be part of this PR?), I'd prefer we consistently use IsAlpha. Unless you have another plan?

flyingapfopenguin:
Compiling your code gives the following error:
`src/sysfiles.h:42:27: Fehler: »GAP_PATH_MAX« ist hier (außerhalb einer Funktion) nicht deklariert;`

`GAP_PATH_MAX` was (and is still) defined in system.h, but sysfiles.h now includes common.h instead of system.h.
Please not that this definition can't be moved simply to common.h since `GAP_PATH_MAX` is required at different positions of the code, e.g. in system.c.

fingolfin:
The definition of `GAP_PATH_MAX` certainly could be moved to `common.h`, but it's not supposed to, because its value is system dependant, hence it'll remain in `system.h`. The compiler error indeed was  stupid mistake (caused by working on a change to remove all uses of `GAP_PATH_MAX` from headers, and then doing insufficient testing on my side before pushing). It is now fixed, I hope.

src/common.h
fingolfin: Fixed, thanks

src/scanner.h
fingolfin: Longterm I'd like to remove `IsAlpha` simply because it seems pointless to "wrap" a perfectly fine ISO C function. But this PR is not meant to do that; I restored the code, but moved `IsIdent` to `sysstr.h`, where it seems to fit in better.

ChrisJefferson:
Would merge, but conflicting

fingolfin:
@ChrisJefferson rebased, let's see if CI tests still pass

PR #3799, opened 2019-12-21
Split system.h and system.c, move parts of their content to new sys*.{c,h} files

fingolfin:
This is complementing PR #3798 in that together, they reduce system.h to a tiny reminder.

The advantage of this split is that system.h is currently included by *every* kernel source file, but most of them don't need the majority of the functionality in it. Indeed, most only need what is in common.h (see #3798). A very few need the time/string/setjmp functionality which this PR splits away. It is my hope that by reducing system.h, it'll get easier to do further refactoring.


src/tracing.h
```diff
@@ -19,7 +19,7 @@
 // so the macros below function correctly
 void ReportWrappedOperation1(const char *, Obj op);
 void ReportWrappedOperation2(const char *, Obj op1, Obj op2);
-void InstallOpWrapper(voidfunc, voidfunc);
+void InstallOpWrapper(void (*activate)(void), void (*deactivate)(void));
```
DominikBernhardt: Sorry if this is a stupid question, but does this change have anything to do with the above changes or is it independent?

fingolfin: Yes, voidfunc is a typedef in system.h and it felt silly to include that just for this trivial typedef

DominikBernhardt: Oh right, thanks for explaining! :)

fingolfin:
This PR is not really related to `make install` support, it just helps to reduce header (inter-)dependencies.

PR #3800, opened 2019-12-21
Change the use of ", say"

DominikBernhardt:
This PR originated from discussions in #3780 .
It aims at replacing formulations containing `, say` as an implicit way of defining variables in explanations.

### DIFF
--- a/doc/ref/gappkg.xml
+++ b/doc/ref/gappkg.xml
@@ -236,8 +236,8 @@ we provide guidelines for the release preparation and its distribution.
 <Heading>Structure of a GAP Package</Heading>
 
 <Index Subkey="for a GAP package">home directory</Index>
-A &GAP; package should have an alphanumeric name (<A>packagename</A>, say);
-mixed case is fine, but there should  be  no  whitespace characters. 
+A &GAP; package should have an alphanumeric name;
+mixed case is fine, but there should be no whitespace characters.
 All files of a &GAP; package <A>packagename</A> must be collected in a 
 single directory <A>packagedir</A>, where <A>packagedir</A> should  be
 just <A>packagename</A> optionally converted to lowercase and optionally 
@@ -923,7 +923,7 @@ before the package is loaded so it will be silently overwritten.
 <Index>needed package</Index>
 <Index>suggested package</Index>
 <Index Subkey="for a GAP package">dependencies</Index>
-It is possible for one &GAP; package <C>A</C>, say,
+It is possible for one &GAP; package <C>A</C>
 to require another package <C>B</C>.
 For that, one simply adds the name and the (least) version number of the
 package <C>B</C> to the <C>NeededOtherPackages</C> component of the
@@ -1106,7 +1106,7 @@ part and implementation part does in general <E>not</E> allow one to actually
 For example,
 in the case of a <Q>cyclic dependency</Q> as in the second example above,
 suppose that <C>B</C> provides a new function <C>f</C> or a new global record
-<C>r</C>, say, which are declared in the declaration part of <C>B</C>.
+<C>r</C> which are declared in the declaration part of <C>B</C>.
 Then the code in the implementation part of <C>A</C> may contain
 calls to the functions defined in the declaration part of <C>B</C>.
 However, the implementation part of <C>A</C> may be read
@@ -1286,7 +1286,7 @@ This means that the binary is not regarded as essential for this
 package.
 <P/>
 You might also have to cope with the situation that external binaries will
-only run under UNIX (and not, say, under Windows), or may not compile with
+only run under UNIX (and not e.g. under Windows), or may not compile with
 some compilers or default compiler options.
 See&nbsp;<Ref Sect="Testing for the System Architecture"/>
 for information on how to test for the architecture.
--- a/doc/ref/grphomom.xml
+++ b/doc/ref/grphomom.xml
@@ -390,8 +390,8 @@ to form groups of automorphisms.
 <Heading>Calculating with Group Automorphisms</Heading>
 
 Usually the best way to calculate in a group of automorphisms is to
-translate all calculations to an isomorphic group in a representation, for
-which better algorithms are available, say a permutation group. This
+translate all calculations to an isomorphic group in a representation for
+which better algorithms are available, such as a permutation group. This
 translation can be done automatically using <Ref Attr="NiceMonomorphism"/>.
 <P/>
 Once a group knows to be a group of automorphisms (this can be achieved
--- a/doc/ref/grplib.xml
+++ b/doc/ref/grplib.xml
@@ -573,7 +573,7 @@ just consists of the <M>&QQ;</M>-classes of i.m.f.&nbsp;subgroups of <M>GL_n(&QQ
 <P/>
 Secondly, there is the set <M>Q_2(n)</M> of the  <M>&QQ;</M>-classes of the remaining
 i.m.f.&nbsp;subgroups of <M>GL_n(&ZZ;)</M>, i.e.,  of   those which are  not  maximal
-finite subgroups of <M>GL_n(&QQ;)</M>. For any such group  <M>G</M>, say, there is at
+finite subgroups of <M>GL_n(&QQ;)</M>. For any such group <M>G</M>, there is at
 least one class <M>C \in Q_1(n)</M> such that <M>G</M> is conjugate under <M>&QQ;</M> to a
 proper subgroup of some   group <M>H \in C</M>.   In  fact, the class <M>C</M>   is
 uniquely determined for any group   <M>G</M> occurring in the library  (though
@@ -587,7 +587,7 @@ of i.m.f.&nbsp;subgroups  of <M>GL_4(&ZZ;)</M>   with representative subgroups  
 \ldots,  G_6</M> of isomorphism types  <M>G_1 \cong W(F_4)</M>, <M>G_2 \cong D_{12}
 \wr C_2</M>, <M>G_3 \cong  G_4 \cong C_2 \times  S_5</M>, <M>G_5 \cong W(B_4)</M>, and
 <M>G_6  \cong (D_{12}  </M><C>Y</C><M>  D_{12})   \!:\! C_2</M>.  The    corresponding
-<M>&QQ;</M>-classes, <M>R_1, \ldots, R_6</M>, say, are pairwise different except that
+<M>&QQ;</M>-classes, which we denote <M>R_1, \ldots, R_6</M>, are pairwise different except that
 <M>R_3</M> coincides  with <M>R_4</M>.   The  groups <M>G_1</M>,  <M>G_2</M>, and  <M>G_3</M>  are
 i.m.f.&nbsp;subgroups of <M>GL_4(&QQ;)</M>, but <M>G_5</M>  and <M>G_6</M> are not because they
 are  conjugate under <M>GL_4(&QQ;)</M> to proper  subgroups  of <M>G_1</M> and <M>G_2</M>,
--- a/doc/ref/lists.xml
+++ b/doc/ref/lists.xml
@@ -1815,7 +1815,7 @@ An <E>enumerator</E> is an immutable list that need not store its elements
 explicitly but knows, from a set of basic data,
 how to determine the <M>i</M>-th element and the position of a given object.
 A typical example of this is a vector space over a finite field with <M>q</M>
-elements, say, for which it is very easy to enumerate all elements
+elements for which it is very easy to enumerate all elements
 using <M>q</M>-adic expansions of integers.
 <P/>
 Using this enumeration can be even quicker than a binary search in a sorted
--- a/doc/ref/matrix.xml
+++ b/doc/ref/matrix.xml
@@ -140,7 +140,7 @@ or the product of a finite field element and an integer matrix.
 returns the product of the row vector <A>vec</A> and the matrix <A>mat</A>.
 Probably the most usual situation is that <A>vec</A> and <A>mat</A> have the same
 lengths and are defined over a common field,
-and that all rows of <A>mat</A> have the same length <M>m</M>, say;
+and that all rows of <A>mat</A> have some common length <M>m</M>;
 in this case the product is a new row vector of length <M>m</M> over the same
 field which is the sum of the scalar multiples of the rows of <A>mat</A> with the
 corresponding entries of <A>vec</A>.
@@ -173,7 +173,7 @@ This form evaluates to the (Cauchy) product of the two matrices <A>mat1</A> and
 Probably the most usual situation is that the number of columns of <A>mat1</A>
 equals the number of rows of <A>mat2</A>,
 and that the elements of <A>mat</A> and <A>vec</A> lie in a common field;
-if <A>mat1</A> is a matrix with <M>m</M> rows and <M>n</M> columns, say, and <A>mat2</A> is a
+if <A>mat1</A> is a matrix with <M>m</M> rows and <M>n</M> columns and <A>mat2</A> is a
 matrix with <M>n</M> rows and <M>o</M> columns, the result is a new matrix with <M>m</M>
 rows and <M>o</M> columns.
 The element in row <M>i</M> at position <M>j</M> of the product is the sum of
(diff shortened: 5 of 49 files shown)


doc/ref/gappkg.xml
```diff
@@ -1286,8 +1287,8 @@ This means that the binary is not regarded as essential for this
 package.
 <P/>
 You might also have to cope with the situation that external binaries will
-only run under UNIX (and not, say, under Windows), or may not compile with
-some compilers or default compiler options.
+only run under UNIX (and not, for example, under Windows), or may not compile
```
fingolfin: You could also use `e.g.` here and in another few places as a replacement for `say` that has almost the same length and thus causes less of a ripple effect of other changes.

doc/ref/grphomom.xml
```diff
@@ -391,7 +391,7 @@ to form groups of automorphisms.
 
 Usually the best way to calculate in a group of automorphisms is to
 translate all calculations to an isomorphic group in a representation, for
```
fingolfin: OK, unrelated to this PR, but that comma really stumped me, and makes no sense to me. But perhaps native speakers like @wilfwilson or @ChrisJefferson might disagree?
```suggestion
translate all calculations to an isomorphic group in a representation for
```

doc/ref/gappkg.xml
```diff
@@ -1105,8 +1106,8 @@ part and implementation part does in general <E>not</E> allow one to actually
 <E>call</E> functions from a package when the implementation part is read.
 For example,
 in the case of a <Q>cyclic dependency</Q> as in the second example above,
-suppose that <C>B</C> provides a new function <C>f</C> or a new global record
-<C>r</C>, say, which are declared in the declaration part of <C>B</C>.
```
fingolfin: Note that "say" here referred to both `f` and `B`, while your replacement only refers to `B`.

But actually, here in several other places, I'd just drop the "say" instead of inserting any replacement. Because IMHO, mathematicians are trained to read expressions involving an indefinite article, such as "*a* KIND-OF-OBJECT NAME", as saying that from here on, "NAME" refers to some arbitrary instance of KIND-OF-OBJECT. As such, the sentence is clear enough. 

But perhaps my humble opinion is wrong, so it'd be good to hear others pitch in and say how they view this.

doc/ref/gappkg.xml
```diff
@@ -236,8 +236,9 @@ we provide guidelines for the release preparation and its distribution.
 <Heading>Structure of a GAP Package</Heading>
 
 <Index Subkey="for a GAP package">home directory</Index>
-A &GAP; package should have an alphanumeric name (<A>packagename</A>, say);
-mixed case is fine, but there should  be  no  whitespace characters. 
+A &GAP; package should have an alphanumeric name (let us say
+<A>packagename</A>);
```
fingolfin: Here I would simple remove the complete text between `(` and `)`, as it is irrelevant for that sentence; in fact I think it has a negative contribution in that it might confuse (it could be read as being an example for an alphanumeric name, which it of course is; but it really is meant to be the placeholder name for a hypothetical package the rest of the text talks about). 

I'd just remove it, as the next sentence "introduces" it sufficiently on its own, IMHO.

doc/ref/gappkg.xml
```diff
@@ -923,7 +924,7 @@ before the package is loaded so it will be silently overwritten.
 <Index>needed package</Index>
 <Index>suggested package</Index>
 <Index Subkey="for a GAP package">dependencies</Index>
-It is possible for one &GAP; package <C>A</C>, say,
+It is possible for one &GAP; package, let us say <C>A</C>,
```
fingolfin: Or just drop the ", say" without replacement?

doc/ref/gappkg.xml
wucas: I think you're right there or at least I'd read it the same way.

doc/ref/gappkg.xml
wucas: If it makes sense to remove it I would. I think that makes the text smoother to read.

doc/ref/gappkg.xml
wucas: I'm perfectly fine with either `e.g.` or `for example` but to make the endeavor maybe a bit smaller I'd suggest to leave the `say`s in situations where there is actually something after it like in this case. So every time it can be replaced with `for example` I'd leave it because using it that way is not that unusual. But in the places where it can just be removed using it is more unusual and also disturbs the flow.

doc/tut/opers.xml
```diff
@@ -169,7 +169,7 @@ of other  attribute   values. Details  of   this <E>method selection</E>   are
 described in chapter&nbsp;<Ref Chap="Method Selection" BookName="ref"/>.
 <P/>
 We  only present the following  rule  of  thumb here:
-Each installed method for an attribute, say <Ref Func="Size" BookName="ref"/>,
```
wucas: As I said before not changing it would be fine too.

doc/ref/gappkg.xml
wilfwilson: Agreed.

doc/ref/gappkg.xml
wilfwilson: Agreed.

doc/ref/gappkg.xml
wilfwilson: Agreed.

doc/ref/gappkg.xml
wilfwilson: ‘for example’ or ‘e.g.’ are both fine here, although I can’t quite work out the best way to put it in the parenthetical clause.

> and not e.g. under Windows

> and not under e.g. Windows

> and e.g. not under Windows

Sticking with the current ordering is probably best.

I don’t think the commas are needed in any case.

doc/ref/grphomom.xml
wilfwilson: Agree that the comma should go.

doc/ref/grphomom.xml
```diff
@@ -391,7 +391,7 @@ to form groups of automorphisms.
...
-which better algorithms are available, say a permutation group. This
+which better algorithms are available, for example a permutation group. This
```
wilfwilson:
```suggestion
which better algorithms are available, such as a permutation group. This
```

doc/ref/grplib.xml
```diff
@@ -587,7 +587,7 @@ of i.m.f.&nbsp;subgroups  of <M>GL_4(&ZZ;)</M>   with representative subgroups
 \ldots,  G_6</M> of isomorphism types  <M>G_1 \cong W(F_4)</M>, <M>G_2 \cong D_{12}
 \wr C_2</M>, <M>G_3 \cong  G_4 \cong C_2 \times  S_5</M>, <M>G_5 \cong W(B_4)</M>, and
 <M>G_6  \cong (D_{12}  </M><C>Y</C><M>  D_{12})   \!:\! C_2</M>.  The    corresponding
-<M>&QQ;</M>-classes, <M>R_1, \ldots, R_6</M>, say, are pairwise different except that
+<M>&QQ;</M>-classes, call them <M>R_1, \ldots, R_6</M>, are pairwise different except that
```
wilfwilson: I don’t like the “call them” formulation here, to me it’s weirder than the “say” formulation.

My suggestions of something that sounds a bit better:

```suggestion
<M>&QQ;</M>-classes, which we denote <M>R_1, \ldots, R_6</M>, are pairwise different except that
```
```suggestion
<M>&QQ;</M>-classes, let us call them <M>R_1, \ldots, R_6</M>, are pairwise different except that
```
```suggestion
<M>&QQ;</M>-classes, which we call <M>R_1, \ldots, R_6</M>, are pairwise different except that
```
```suggestion
<M>&QQ;</M>-classes, <M>R_1, \ldots, R_6</M>, are pairwise different except that
```

doc/ref/lists.xml
```diff
@@ -1815,8 +1815,8 @@ An <E>enumerator</E> is an immutable list that need not store its elements
 explicitly but knows, from a set of basic data,
 how to determine the <M>i</M>-th element and the position of a given object.
 A typical example of this is a vector space over a finite field with <M>q</M>
-elements, say, for which it is very easy to enumerate all elements
-using <M>q</M>-adic expansions of integers.
+elements, where <M>q</M> is a prime power, for which it is very easy to
+enumerate all elements using <M>q</M>-adic expansions of integers.
```
wilfwilson: This is one of those points where I show my mathematical ignorance:

I would have guessed that q-adic expansions are the same as p-adic expansions, and I would have guessed that the p of p-adic is required to be a prime.

Is that the case? Basically: does it make sense to talk about a (prime power)-adic expansion?

If not, this sentence should be adjusted to say that q must be a prime/the finite field must be prime.

doc/ref/matrix.xml
```diff
@@ -140,7 +140,7 @@ or the product of a finite field element and an integer matrix.
 returns the product of the row vector <A>vec</A> and the matrix <A>mat</A>.
 Probably the most usual situation is that <A>vec</A> and <A>mat</A> have the same
 lengths and are defined over a common field,
-and that all rows of <A>mat</A> have the same length <M>m</M>, say;
+and that all rows of <A>mat</A> have the same length, let us say <M>m</M>;
```
wilfwilson:
```suggestion
and that all rows of <A>mat</A> have some common length <M>m</M>;
```

doc/ref/reesmat.xml
```diff
@@ -38,9 +38,9 @@
   semigroup elements is created from a specific Rees matrix semigroup, which
   contains the whole family of its elements.  So, it is not possible to
   multiply or compare elements belonging to distinct Rees matrix semigroups, 
-  since they belong to different families. This situation is similar to, say,
-  free groups, and different to, say, permutations, which belong to a single
-  family, and where arbitrary permutations can be compared and multiplied
+  since they belong to different families. For example, this situation is
+  similar to free groups, and different to permutations, which belong to a
```
wilfwilson: I think the current formulation doesn’t make clear enough that the “belong to a single family” applies to perms rather than free groups. Perhaps:
```suggestion
  similar to free groups, but it is different to permutations, which belong to a
```
In fact this whole sentence (original) is not ideal and could do with restructuring, but that’s outside the scope of this PR.

doc/ref/pres.xml
```diff
@@ -88,8 +88,8 @@ numbering of generators in a family of associative words.
 <Section Label="Printing Presentations">
 <Heading>Printing Presentations</Heading>
 
-Whenever you create a presentation <M>P</M>, say, or assign it to a variable,
-&GAP; will respond by printing <M>P</M>.
+Whenever you create a presentation, let us call it <M>P</M>, or assign it to a
```
wilfwilson: I think basically removing the “say” is the best option here. Especially as it is the beginning of a section so there is not way that someone could think that P is a pre-existing variable.
```suggestion
Whenever you create a presentation <M>P</M>, or assign <M>P</M> to a variable,
```

doc/ref/lists.xml
DominikBernhardt: You're right, I'm sorry, I didn't pay enough attention there. Will fix it.

doc/ref/gappkg.xml
```diff
@@ -923,7 +923,7 @@ before the package is loaded so it will be silently overwritten.
 <Index>needed package</Index>
 <Index>suggested package</Index>
 <Index Subkey="for a GAP package">dependencies</Index>
-It is possible for one &GAP; package <C>A</C>, say,
+It is possible for one &GAP; package <C>A</C>,
```
fingolfin:
```suggestion
It is possible for one &GAP; package <C>A</C>
```

doc/ref/gappkg.xml
```diff
@@ -1286,8 +1286,8 @@ This means that the binary is not regarded as essential for this
 package.
 <P/>
 You might also have to cope with the situation that external binaries will
-only run under UNIX (and not, say, under Windows), or may not compile with
-some compilers or default compiler options.
+only run under UNIX (and not e.g. under Windows), or may not compile
+with some compilers or default compiler options.
```
fingolfin: Optional and unimportant, but: could move "with" back to the previous line, to reduce the diff

doc/ref/lists.xml
```diff
@@ -1815,8 +1815,8 @@ An <E>enumerator</E> is an immutable list that need not store its elements
 explicitly but knows, from a set of basic data,
 how to determine the <M>i</M>-th element and the position of a given object.
 A typical example of this is a vector space over a finite field with <M>q</M>
-elements, say, for which it is very easy to enumerate all elements
-using <M>q</M>-adic expansions of integers.
+elements, where <M>q</M> is a prime, for which it is very easy to
```
fingolfin: Why is `q` limited to a prime here, as opposed to being a non-trivial prime power? Or is that an accident? If so, instead of fixing that insertion, perhaps just undo it? Because I'd expect most if not all readers to know that the size of a finite field always is a prime power, and even if not, pointing that out is not really relevant for the example, is it? After all, q-adic expansion is also define for q which is not a prime power.

doc/ref/matrix.xml
```diff
@@ -140,7 +140,7 @@ or the product of a finite field element and an integer matrix.
 returns the product of the row vector <A>vec</A> and the matrix <A>mat</A>.
 Probably the most usual situation is that <A>vec</A> and <A>mat</A> have the same
 lengths and are defined over a common field,
-and that all rows of <A>mat</A> have the same length <M>m</M>, say;
+and that all rows of <A>mat</A> have some common length <M>m</M>;
```
fingolfin: Personally I find the phrasing with "the same" (but with "say" removed) clearer. But that may just be me. Not a blocker.

doc/ref/pres.xml
```diff
@@ -88,8 +88,8 @@ numbering of generators in a family of associative words.
 <Section Label="Printing Presentations">
 <Heading>Printing Presentations</Heading>
 
-Whenever you create a presentation <M>P</M>, say, or assign it to a variable,
-&GAP; will respond by printing <M>P</M>.
+Whenever you create a presentation <M>P</M>, or assign it to a
+variable, &GAP; will respond by printing <M>P</M>.
```
fingolfin: Optional and unimportant, but: could move "variable, " back to the previous line, to reduce the diff

DominikBernhardt:
@fingolfin My idea was to go through all the formulations using the `,say` construction. I thought I might all put them in this PR? So I left the WIP as there are many more places where this is used and I'm currently working on changing all of them.

doc/ref/matrix.xml
DominikBernhardt: I followed a suggestion by @wilfwilson in one of his above comments. I quite like this, so I went with it.

doc/ref/lists.xml
DominikBernhardt: Maybe this is completely ignorant of me, but I thought `p`-adic extension is defined for prime numbers in the standard literature? @wilfwilson pointed this out to me above. Originally, I put `where <M>q</M> is a prime power`. Of course, a field can have prime-power order, but my impression was that `q`-adic expansion always refers to a prime number.

fingolfin: g-adic expansion is defined for any integer g >= is 2; for g=10 you use it every day to write numbers ... :-) See also the GAP function `CoefficientsQadic`.

fingolfin:
I didn't realize there were many more uses of "say". So the "Unless" in "Unless you plan further major changes" is triggered -- fine by me :-)

doc/ref/lists.xml
DominikBernhardt: Oh my.... Can I please delete this thread? 🤦‍♂

wilfwilson: Don’t worry @DominikBernhardt - it’s fine for us to be ignorant of parts of mathematics and terminology (it’s a big subject 😉).

DominikBernhardt:
The status of this is the following:
There are still a few `, say`-constructions in files in `extern/` and `hpcgap/extern/`. Is it worth changing it there?
Apart from that, I either changed, removed, reformulated (and rarely left) the uses of `, say`.

fingolfin:
Please don't modify files in `extern` and `extern/hpcgap`; they are imported from upstream "vendors" and if we ever need to upgrade (e.g. to a newer GMP), we'll have to very carefully comb through any patches we applied and port them. That's a lot of work and should be kept to a minimum. Besides, those files should normally never affect anything visible to GAP users.

DominikBernhardt:
@fingolfin This is what I thought, I just thought I'd check with you.

wilfwilson:
Yes, you may as well squash, and update title and labels. Thanks.

DominikBernhardt:
Updated and squashed.

PR #3803, opened 2019-12-24
kernel: misc changes

fingolfin:
These are all minor tweaks.

<strike>
The 1UL -> (UInt)1 changes are done because on some 64bit systems, "unsigned long" is actually just 32bit, so something like `1UL << 44` won't work right (it'll be zero) while `(UInt)1 << 44` will work.

I am not super happy about the name `ALL_BITS_UINT`, better suggestions welcome.

Lastly, as always I am happy to split this PR if it helps, though it should be possible to review it commit-by-commit.</strike>

I moved parts of this PR to #3812 and #3813

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -627,13 +627,11 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
     // try to look up the variable on the stack of local variables
     const UInt countNams = LEN_PLIST(ReaderState()->StackNams);
     for (nest = 0; nest < countNams; nest++) {
-#ifndef SYS_IS_64_BIT
         if (nest >= MAX_FUNC_EXPR_NESTING) {
             Pr("Warning: abandoning search for %s at %dth higher frame\n",
                (Int)s->Value, MAX_FUNC_EXPR_NESTING);
             break;
         }
-#endif
         nams = ELM_PLIST(ReaderState()->StackNams, countNams - nest);
         indx = findValueInNams(nams, s->Value, 1, LEN_PLIST(nams));
         if (indx != 0) {
@@ -664,14 +662,12 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
             }
             lvars = ENVI_FUNC(FUNC_LVARS(lvars));
             nest++;
-#ifndef SYS_IS_64_BIT
             if (nest >= MAX_FUNC_EXPR_NESTING) {
                 Pr("Warning: abandoning search for %s at %dth higher "
                    "frame\n",
                    (Int)s->Value, MAX_FUNC_EXPR_NESTING);
                 break;
             }
-#endif
         }
         lvars0 = PARENT_LVARS(lvars0);
         nest0++;
--- a/src/read.c
+++ b/src/read.c
@@ -1319,7 +1319,7 @@ static UInt ReadLocals(ScannerState * s, TypSymbolSet follow, Obj nams)
             }
             nloc += 1;
             PushPlist(nams, MakeImmString(s->Value));
-            if (LEN_PLIST(nams) >= 65536) {
+            if (LEN_PLIST(nams) >= MAX_FUNC_LVARS) {
                 SyntaxError(s, "Too many function arguments and locals");
             }
         }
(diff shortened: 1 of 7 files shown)


src/read.c
```diff
@@ -626,13 +626,11 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
     // try to look up the variable on the stack of local variables
     const UInt countNams = LEN_PLIST(ReaderState()->StackNams);
     for (nest = 0; nest < countNams; nest++) {
-#ifndef SYS_IS_64_BIT
```
DominikBernhardt: Out of curiosity, what is the reason for changing this now or why was this only enforced in 32bit systems in the first place?

DominikBernhardt:
In the description of this PR you're writing:

> so something like 1UL << 44 won't work right (it'll be zero) while (UInt)1 << 4 will work.

Do you mean `(UInt)1 << 44` instead?

src/read.c
fingolfin: The limit used to be enforced uniformly but with a incorrect error message up to and including GAP 4.6:
```C
...
	if (nest >= 65536)
	  {
	    Pr("Warning: abandoning search for %s at 1024th higher frame\n",
	       (Int)Value,0L);
	    break;
	  }
      }
      lvars0 = PTR_BAG( lvars0 )[2];
      nest0++;
	if (nest0 >= 65536)
	  {
	    Pr("Warning: abandoning search for %s 4096 frames up stack\n",
	       (Int)Value,0L);
	    break;
	  }
...
```
In GAP 4.7 we fixed the messages to say it 65536 in both cases.

in GAP 4.9 the code underwent a major refactoring. The code became clearer, and it became obvious were the restriction to 16 bits came from: to make sure the values `nest` and `indx` fit together into a 32bit value. But on 32bit systems, `nest` actually could take 48=64-16 bits, which is for all practical purposes unlimited. Hence the check was limited to happen on 32bit systems.

However, later on (basically in GAP 4.11), I changed struct `LHSRef` and limited the storage for this combined value back to 32bit (for various reasons, IMHO good ones, but doesn't matter for now). Anyway, this means we should limit the value on 64 bit systems, too!

There is also a non-technical argument: it seems bad to allow code to work on 64bit that won't work on 32bit. Not that it is likely: code that breaks this limit would be extremely weird (and slow).

fingolfin: I'll update the commit message for this change, and also for the other thing you pointed out.

fingolfin:
Updated two commit messages based on @DominikBernhardt's good feedback, and also added one final commit replacing a use of the magic constant 65536 by the named constant `MAX_FUNC_LVARS`

src/read.c
wilfwilson: Since you’ve put this comment into a commit message, it may as well be correct, so I offer the following suggestions:

- ‘A incorrect’ → ‘an incorrect’
- ‘Obvious were’ → ‘obvious where’
- ‘But on 32bit systems’ → ‘but on 64bit systems’

I also think that 32- and 64-bit should be written with hyphens, but that’s a totally inconsequential nitpick, and so I’m not asking you to change your usage!

src/gasman.c
```diff
@@ -2266,50 +2263,44 @@ UInt CollectBags (
...
+        nrBags += CollectBags_Sweep(full);
 
-    /* * * * * * * * * * * * * * * sweep phase * * * * * * * * * * * * * * */
-    nrBags += CollectBags_Sweep();
+        // check phase
+        done = CollectBags_Check(size, full, nrBags);
 
-    /* * * * * * * * * * * * * * * check phase * * * * * * * * * * * * * * */
-    done = CollectBags_Check(size, nrBags);
-    if (done == 2)
```
wilfwilson: (Why) have you discarded this case?

src/read.c
fingolfin: @wilfwilson thanks, all fixed and some more :-)

src/gasman.c
fingolfin: Oops, good catch, that was an accident. Well, the plan *is* to remove that, but then also the final `return` needs to be changed. But that latter change is missing here; and in any case, this change of how `done == 2` is handled should not be sneaked in silently like this as part of refactoring. I've now put it into its own commit, properly and with a commit message/

Note that this change will conflict with PR #3793, though resolving the conflict ought to be trivial.

fingolfin: I've now moved this into its own PR #3812

src/read.c
DominikBernhardt: Thanks for explaining @fingolfin !

PR #3804, opened 2019-12-27
kernel: add UpdateTime helper

fingolfin:
Also turn Last, Last2, Last3, Time into static vars

### DIFF
--- a/src/gap.h
+++ b/src/gap.h
@@ -18,30 +18,6 @@
 
 /****************************************************************************
 **
-*V  Last  . . . . . . . . . . . . . . . . . . . . . . global variable  'last'
-**
-**  'Last',  'Last2', and 'Last3'  are the  global variables 'last', 'last2',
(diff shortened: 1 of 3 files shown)


src/gap.h
```diff
@@ -16,30 +16,6 @@
...
-*V  Last2 . . . . . . . . . . . . . . . . . . . . . . global variable 'last2'
-*/
-extern UInt Last2;
-
-
-/****************************************************************************
-**
-*V  Last3 . . . . . . . . . . . . . . . . . . . . . . global variable 'last3'
-*/
-extern UInt Last3;
```
fingolfin: @wucas in reply to your question: yes, you guessed right; removing the above lines from `gap.h` in a sense is the central purpose of this PR. Reducing the "surface" of the kernel makes it easier to reason about, easier to use it as a library, easier to make future changes. Of course this PR is just a tiny contribution to this overall goal.

PR #3807, opened 2019-12-29
kernel: move IsIdent to sysstr.h

fingolfin:
This was/is part of PR #3798 but is somewhat independent of that PR, so it could be merged/rejected on its own. I also kind of ended up needing the same set of #include changes in PR #3805, so it seemed sensible to put this into its own PR, in the hopes that it can be merged faster this way...

### DIFF
--- a/src/sysstr.h
+++ b/src/sysstr.h
@@ -37,6 +37,16 @@
 
 /****************************************************************************
 **
+*F  IsIdent( <ch> )
+*/
+EXPORT_INLINE int IsIdent(char ch)
+{
+    return IsAlpha(ch) || ch == '_' || ch == '@';
+}
+
+
+/****************************************************************************
+**
 *F  strlcpy( <dst>, <src>, <len> )
 **
 **  Copy <src> to buffer <dst> of size <len>. At most <len>-1 characters will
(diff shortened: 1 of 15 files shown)


DominikBernhardt:
Just something I've noticed while looking at this (in principle unrelated, but I thought I might mention it here):
You will find an explanatory text at the beginning of most `.h` files in `src/` starting with 

> This file declares [...]

and the corresponding `.c` file has something similar.
`sysstr.{c,h}` does not contain such a text. Is this something we would like to add?

src/sysstr.h
```diff
@@ -35,6 +35,16 @@
 #define IsDigit(ch) (isdigit((unsigned int)ch))
 
 
+/****************************************************************************
+**
+*F  IsIdent( <ch> )
+*/
```
DominikBernhardt: Seeing that all other functions in here have (at least a short) explanation on what they do, maybe you could add one here?

fingolfin: This should be added at some point, but since I just moved this code around, and since two other PRs depend on this PR, I'd rather do this in a separate PR, and merge this PR here right away.
Same for the comment at the file start -- note that these are either missing or wrong for basically all `sys*.{c,h}` files at this point...

PR #3809, opened 2019-12-29
kernel: tweak sysstr.h, IsIdent()

fingolfin:
- move #include <ctype.h> from system.h to sysstr.h
- add #include <string.h> to sysstr.h (for strlcpy, strlcat on BSD and macOS)
- IsIdent now also accept digits, by using isalnum() instead of IsAlpha; this
  is sightly more efficient. In a few places were we are using IsIdent, we
  strictly speaking don't want to accept digits; but there, digits never can
  be the input anyway, so all is good
- add a file description comment to the top of sysstr.c and sysstr.h

Partially motivated by PR #3807

### DIFF
--- a/src/sysstr.h
+++ b/src/sysstr.h
@@ -7,6 +7,8 @@
 **
 **  SPDX-License-Identifier: GPL-2.0-or-later
 **
+**  This file declares low-level functions dealing with C strings and
+**  characters.
 */
 
 #ifndef GAP_SYSSTR_H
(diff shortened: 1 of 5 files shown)


src/sysstr.h
```diff
@@ -37,11 +41,15 @@
 
 /****************************************************************************
 **
-*F  IsIdent( <ch> )
+*F  IsIdent( <ch> ) . . . . . . . .  is a character valid in a GAP identifier
+**
+**  'IsIdent' returns 1 if its its character argument can be used unquoted
```
wilfwilson: Repeated its

fingolfin: Fixed, thanks

PR #3811, opened 2019-12-29
kernel: replace some uses of Char by char

fingolfin:
... to make debugging with LLDB a bit easier: that debugger for some reason
does not treat Char and char pointers the same; the latter are printed as C
strings, the former as arrays. With this change, one can e.g. print any
`ScannerState` instance within LLDB and get a nice representation, instead of
flooding the screen with hundreds of lines, including one line for each entry
of `char Value[1024]`.

Actually, I wonder why we even bother to `typedef char Char` in the first place. I am tempted to just replace *all* occurrences of `Char` in the kernel by `char`, except for that typedef (which I'd keep for now, to ensure compatibility with a few packages whose kernel extensions use `Char`). Can anybody think of a reason not to do that (the only one I can think of is "symmetry with `UChar`", but that seems like a weak argument to me; and in any case, my answer would be "so let's get rid of `UChar` as well ;-).

Actually, I can think of one other hypothetical reason: the C standard does not define whether `char` is signed and unsigned, and in the past, implementations doing either existed; this could sometimes cause issues if code relied on `char` being signed. So perhaps the intention was to do `typedef signed char Char`? But fact is: that's not what we do, and never was. So that can't be a reason for this either.

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -70,9 +70,9 @@ typedef struct GAPState {
     // begin interpreted, so the current line is outputted when profiling
     UInt InterpreterStartLine;
 
-    const Char * Prompt;
+    const char * Prompt;
 
-    Char * In;
+    char * In;
 
     /* From stats.c */
     Obj  ReturnObjStat;
--- a/src/io.c
+++ b/src/io.c
@@ -63,7 +63,7 @@ typedef struct {
     Int file;
 
     // the name of the file; this is only used in error messages
-    Char name[256];
+    char name[256];
 
     //
     UInt gapnameid;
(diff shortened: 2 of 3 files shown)


ChrisJefferson:
I honestly thought (having never checked obviously!) that Char was unsigned char. Turns out if it is just char, there is no reason to keep it (unless it is in packages). Perhaps it dates from sufficiently long ago there was no char in some compilers? in the early prehistory? (not that that matters now)

fingolfin:
@ChrisJefferson there are a few packages using `Char`, e.g. json and profiling ... ;-). Actually, the only other is `Browse`, and there it's a single use in a cast that should be removed anyway.

PR #3813, opened 2019-12-30
kernel: add ALL_BITS_UINT; avoid 1UL, 0UL

fingolfin:
On some 64bit systems, "unsigned long" is actually just 32bit, so something like `1UL << 44` won't work right (it evaluates to zero) while `(UInt)1 << 44` will work.

I am not super happy about the name `ALL_BITS_UINT`, better suggestions welcome.

This change used to be part of PR #3803, but I moved it to its own PR as it kind of drowned out all the other changes in that PR.


src/scanner.h
```diff
@@ -30,95 +30,95 @@
...
-    S_RETURN            = (1UL<<29)+1,
-    S_QUIT              = (1UL<<29)+2,
-    S_QQUIT             = (1UL<<29)+3,
-    S_CONTINUE          = (1UL<<29)+4,
-
-    S_SEMICOLON         = (1UL<<30)+0,
-    S_DUALSEMICOLON     = (1UL<<30)+1,
-
-    S_EOF               = (1UL<<31),
+    S_ILLEGAL           = 0,
```
DominikBernhardt: I hope this is not a stupid question, but why is this equivalent to ` S_ILLEGAL = 0UL,`? In the above changes, you always replace the `UL` construction with `(UInt)`, here you just drop it.

fingolfin: Not a stupid question at all! I dropped it here because all those values fit into 32bit, and the resulting enum values will be stored as 32bit, too, so there is no point in using `(UInt)1` or `1UL`.

Side remark: the C standard only guarantees that an enum must be able to represent any `int` value, which means 32bit these days -- we are thus careful not to use more than 31 bits here (we can't use the sign bit). GCC actually supports 64bit enum values, but there are C compilers which don't, and the clang static analyzer warns about it.

DominikBernhardt: Ah that is a good point, thanks!

PR #3814, opened 2019-12-30
kernel: introduce BOOL, TRUE, FALSE

fingolfin:
This is a resurrected version of PR #1883.

This adds `typedef UChar BOOL`; the resulting C type `BOOL` can be used to
indicate that the return type of a function is supposed to be a boolean.
Matching integer constants TRUE=1 and FALSE=0 are also added.

The reason we are not using type `bool` from stdbool.h and the constants
`true` and `false`, all available in C99, is the following: in C99, a pointer
is automatically converted to type bool if necessary. That means that if one
by accident writes `return False` instead of `return false` in a function with
return type `BOOL`, then no compiler warning is issued. Instead, a bug is
introduced, as `False` is a non-NULL pointer, and thus is cast to the `bool`
value `true`.

With this commit, we get warnings similar to the following if we mix up
TRUE/True or FALSE/False; while if using `bool`, we'd only get the latter two
warnings, but not the first one.

    src/objects.c:897:12: error: incompatible pointer to integer conversion
          returning 'Obj' (aka 'unsigned long **') from a function with result
          type 'BOOL' (aka 'unsigned char') [-Werror,-Wint-conversion]
        return False;
               ^~~~~

    src/objects.c:1185:39: error: pointer/integer type mismatch in conditional
          expression ('int' and 'Obj' (aka 'unsigned long **'))
          [-Werror,-Wconditional-type-mismatch]
        return (TNUM_OBJ(obj) == T_COMOBJ ? TRUE : False);
                                          ^ ~~~~   ~~~~~

    src/objects.c:1337:16: error: incompatible integer to pointer conversion
          returning 'int' from a function with result type 'Obj' (aka
          'unsigned long **') [-Werror,-Wint-conversion]
            return TRUE;
                   ^~~~

### DIFF
--- a/src/blister.h
+++ b/src/blister.h
@@ -27,7 +27,7 @@
 **
 *F  IS_BLIST_REP( <list> )  . . . . .  check if <list> is in boolean list rep
 */
-EXPORT_INLINE Int IS_BLIST_REP(Obj list)
+EXPORT_INLINE BOOL IS_BLIST_REP(Obj list)
 {
     return T_BLIST <= TNUM_OBJ(list) &&
            TNUM_OBJ(list) <= T_BLIST_SSORT + IMMUTABLE;
@@ -47,6 +47,7 @@ EXPORT_INLINE Int SIZE_PLEN_BLIST(Int plen)
     GAP_ASSERT(plen >= 0);
     return sizeof(Obj) + (plen + BIPEB - 1) / BIPEB * sizeof(UInt);
 }
+
 
 /****************************************************************************
 **
--- a/src/calls.h
+++ b/src/calls.h
@@ -223,7 +223,7 @@ EXPORT_INLINE void SET_LCKS_FUNC(Obj func, Obj locks)
 **  'IsKernelFunction' returns 1 if <func> is a kernel function (i.e.
 **  compiled from C code), and 0 otherwise.
 */
-Int IsKernelFunction(Obj func);
+BOOL IsKernelFunction(Obj func);
 
 
 EXPORT_INLINE ObjFunc_0ARGS HDLR_0ARGS(Obj func)
@@ -271,7 +271,7 @@ EXPORT_INLINE ObjFunc_1ARGS HDLR_XARGS(Obj func)
 **
 *F  IS_FUNC( <obj> )  . . . . . . . . . . . . . check if object is a function
 */
-EXPORT_INLINE int IS_FUNC(Obj obj)
+EXPORT_INLINE BOOL IS_FUNC(Obj obj)
 {
     return TNUM_OBJ(obj) == T_FUNCTION;
 }
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -442,7 +442,7 @@ UInt                    NrHalfDeadBags;
 **
 *F  IS_BAG_ID -- check if a value looks like a masterpointer id
 */
-static inline UInt IS_BAG_ID(void * ptr)
+static inline BOOL IS_BAG_ID(void * ptr)
 {
     return (((void *)MptrBags <= ptr) && (ptr < (void *)MptrEndBags) &&
             ((UInt)ptr & (sizeof(Bag) - 1)) == 0);
@@ -452,7 +452,7 @@ static inline UInt IS_BAG_ID(void * ptr)
 **
 *F  IS_BAG_BODY -- check if value like a pointer to a bag body
 */
-static inline UInt IS_BAG_BODY(void * ptr)
+static inline BOOL IS_BAG_BODY(void * ptr)
 {
     return (((void *)OldBags <= ptr) && (ptr < (void *)AllocBags) &&
             ((UInt)ptr & (sizeof(Bag) - 1)) == 0);
@@ -606,17 +606,17 @@ static inline Bag MARKED_HALFDEAD(Bag x)
     return (Bag)((UInt)x | HALFDEAD);
 }
 
-static inline Int IS_MARKED_DEAD(Bag x)
+static inline BOOL IS_MARKED_DEAD(Bag x)
 {
     return LINK_BAG(x) == MARKED_DEAD(x);
 }
 
-// static inline Int IS_MARKED_ALIVE(Bag x)
+// static inline BOOL IS_MARKED_ALIVE(Bag x)
 // {
 //     return LINK_BAG(x) == MARKED_ALIVE(x);
 // }
 
-static inline Int IS_MARKED_HALFDEAD(Bag x)
+static inline BOOL IS_MARKED_HALFDEAD(Bag x)
 {
     return LINK_BAG(x) == MARKED_HALFDEAD(x);
 }
@@ -685,12 +685,12 @@ void MarkBagWeakly(Bag bag)
     }
 }
 
-Int IsWeakDeadBag(Bag bag)
+BOOL IsWeakDeadBag(Bag bag)
 {
     CANARY_DISABLE_VALGRIND();
-    Int isWeakDeadBag = (((UInt)bag & (sizeof(Bag) - 1)) == 0) &&
-                        (Bag)MptrBags <= bag && bag < (Bag)MptrEndBags &&
-                        (((UInt)*bag) & (sizeof(Bag) - 1)) == 1;
+    BOOL isWeakDeadBag = (((UInt)bag & (sizeof(Bag) - 1)) == 0) &&
+                         (Bag)MptrBags <= bag && bag < (Bag)MptrEndBags &&
+                         (((UInt)*bag) & (sizeof(Bag) - 1)) == 1;
     CANARY_ENABLE_VALGRIND();
     return isWeakDeadBag;
 }
@@ -708,7 +708,7 @@ void CallbackForAllBags(void (*func)(Bag))
 {
     for (Bag bag = (Bag)MptrBags; bag < (Bag)MptrEndBags; bag++) {
         CANARY_DISABLE_VALGRIND();
-        Int is_bag = IS_BAG_BODY(*bag);
+        BOOL is_bag = IS_BAG_BODY(*bag);
         CANARY_ENABLE_VALGRIND();
         if (is_bag) {
             (*func)(bag);
--- a/src/hpc/traverse.c
+++ b/src/hpc/traverse.c
@@ -35,7 +35,7 @@ struct TraversalState {
     UInt                    hashBits;
     Region *                region;
     int                     border;
-    int (*traversalCheck)(TraversalState *, Obj);
+    BOOL (*traversalCheck)(TraversalState *, Obj);
 };
 
 static Obj NewList(UInt size)
@@ -159,7 +159,7 @@ void QueueForTraversal(TraversalState * traversal, Obj obj)
 
 static void TraverseRegionFrom(TraversalState * traversal,
                         Obj              obj,
-                        int (*traversalCheck)(TraversalState *, Obj))
+                        BOOL (*traversalCheck)(TraversalState *, Obj))
 {
     GAP_ASSERT(IS_BAG_REF(obj));
     GAP_ASSERT(REGION(obj) != NULL);
@@ -204,24 +204,24 @@ static void TraverseRegionFrom(TraversalState * traversal,
 //   return CheckReadAccess(obj);
 // }
 
-static int IsSameRegion(TraversalState * traversal,Obj obj)
+static BOOL IsSameRegion(TraversalState * traversal, Obj obj)
 {
     return REGION(obj) == traversal->region;
 }
 
-static int IsMutable(TraversalState * traversal, Obj obj)
+static BOOL IsMutable(TraversalState * traversal, Obj obj)
 {
     return CheckReadAccess(obj) && IS_MUTABLE_OBJ(obj);
 }
 
-static int IsWritableOrImmutable(TraversalState * traversal, Obj obj)
+static BOOL IsWritableOrImmutable(TraversalState * traversal, Obj obj)
 {
     int writable = CheckExclusiveWriteAccess(obj);
     if (!writable && IS_MUTABLE_OBJ(obj)) {
         traversal->border = 1;
-        return 0;
+        return FALSE;
     }
-    return 1;
+    return TRUE;
 }
 
 Obj ReachableObjectsFrom(Obj obj)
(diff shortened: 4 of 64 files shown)


src/lists.h
```diff
@@ -592,7 +592,7 @@ EXPORT_INLINE Int IS_TABLE_LIST(Obj list)
 *F  IS_SSORT_LIST(<list>) . . . . . . . . . .  test for strictly sorted lists
 *V  IsSSortListFuncs[<type>]  .  table of strictly sorted list test functions
 **
-**  'IS_SSORT_LIST' returns 2 if the list <list> is  a strictly  sorted  list
+**  'IS_SSORT_LIST' returns 1 if the list <list> is  a strictly  sorted  list
```
fingolfin: Note that I changed this to always return 0 or 1. Many (most?) of the functions installed in `IsSSortListFuncs` already return 0 or 1, but a few returned 0 or 2 as documented. Luckily, all callers (there are not many) only care whether the return value is zero or non-zero, so this didn't matter.

By changing the return value to `BOOL` and using `return TRUE` and `return FALSE`, I think this code got clearer.

src/blister.h
```diff
@@ -48,6 +48,7 @@ EXPORT_INLINE Int SIZE_PLEN_BLIST(Int plen)
     return sizeof(Obj) + (plen + BIPEB - 1) / BIPEB * sizeof(UInt);
 }
 
+
```
DominikBernhardt: Why did you add this line? This is not consistently done in the GAP kernel. Do we want to have two empty lines before we habe a `/***[...]` line?

src/calls.h
```diff
@@ -271,7 +271,7 @@ EXPORT_INLINE ObjFunc_1ARGS HDLR_XARGS(Obj func)
 **
 *F  IS_FUNC( <obj> )  . . . . . . . . . . . . . check if object is a function
```
DominikBernhardt:
```suggestion
*F  IS_FUNC( <obj> )  . . . . . . . . . . . . . check if object <obj> is a function
```
Just a minor remark.

src/gasman.c
```diff
@@ -772,10 +772,8 @@ void InitGlobalBag (
 }
 
 
-static Int IsLessGlobal (
-    const Char *        cookie1,
-    const Char *        cookie2,
-    UInt                byWhat )
+static BOOL
+IsLessGlobal(const Char * cookie1, const Char * cookie2, UInt byWhat)
```
DominikBernhardt: A general question here: Above, you changed return values `0` and `1` to `FALSE` and `TRUE`, below in lines 783 and following (I can't mark that line as there is no change in it), you left the return values `0`, `1` and `-1`. Why did you change the return value of the function to BOOL then?

src/hpc/traverse.c
```diff
@@ -204,17 +204,17 @@ static void TraverseRegionFrom(TraversalState * traversal,
...
 }
 
-static int IsMutable(TraversalState * traversal, Obj obj)
+static BOOL IsMutable(TraversalState * traversal, Obj obj)
 {
     return CheckReadAccess(obj) && IS_MUTABLE_OBJ(obj);
 }
 
-static int IsWritableOrImmutable(TraversalState * traversal, Obj obj)
+static BOOL IsWritableOrImmutable(TraversalState * traversal, Obj obj)
```
DominikBernhardt: Same as in a comment above: This function still returns `0` and `1`. Maybe I don't understand something, but it seems to me like this should be `FALSE` and `TRUE`?

src/integer.h
```diff
@@ -33,7 +33,7 @@ EXPORT_INLINE Int IS_LARGEINT(Obj obj)
 **  'IS_INT' returns 1 if 'obj' is either a large or an immediate integer
```
DominikBernhardt: So the return value of the function now is `BOOL`, should this comment be updated to 
```suggestion
**  'IS_INT' returns TRUE if 'obj' is either a large or an immediate integer
```

DominikBernhardt: There are a bunch of places where this is the case. I don't point them all out now because I first would like to hear your opinion on this.

src/lists.c
```diff
@@ -77,14 +77,14 @@ static Int IsListObject(Obj obj)
...
-Int             (*IsSmallListFuncs [LAST_REAL_TNUM+1]) ( Obj obj );
+BOOL (*IsSmallListFuncs[LAST_REAL_TNUM + 1])(Obj obj);
 
 static Obj IsSmallListFilt;
 static Obj HasIsSmallListFilt;
 static Obj LengthAttr;
 static Obj SetIsSmallList;
 
-static Int IsSmallListObject(Obj obj)
+static BOOL IsSmallListObject(Obj obj)
```
DominikBernhardt: Same as above, some return values are still `1` and `0`.

src/objects.h
```diff
@@ -562,8 +562,8 @@ void CheckedMakeImmutable(Obj obj);
 **  'IS_MUTABLE_OBJ' returns   1 if the object  <obj> is mutable   (i.e., can
 **  change due to assignments), and 0 otherwise.
 */
-extern Int (*IsMutableObjFuncs[LAST_REAL_TNUM+1]) ( Obj obj );
-EXPORT_INLINE Int IS_MUTABLE_OBJ(Obj obj)
+extern BOOL (*IsMutableObjFuncs[LAST_REAL_TNUM + 1])(Obj obj);
+EXPORT_INLINE BOOL IS_MUTABLE_OBJ(Obj obj)
```
DominikBernhardt: This also returns `0`.

src/blister.h
fingolfin: I generally aim for two empty lines, but this is not codified anywhere (and I am not keen on trying to codify it explicitly or officially either, as I am afraid that will just lead down a road of endless bikeshedding). In this particular case, the functions before and after this all used two empty lines (later on in the file it is again inconsistent). I guess I inserted the empty line (two years ago...) out of reflex. I can drop it, but I think it doesn't hurt either.

src/calls.h
fingolfin: Out of 3083 such comments ("short function description"), just 136 reference a variable in this way. 

I really think the comment is fine as it is (it conveys its meaning well, doesn't it?). In any case, if it is to be changed, I'd suggest to change it in more places, and in a PR dedicate to it.

src/gasman.c
fingolfin: This was a mistake, thank you for catching it. I'll restore the return type, and also rename the function (IMHO anything starting with `Is...` like this should return a BOOL)

fingolfin: Actually, looking at the code in question closer, I think it simply is buggy (probably due to a past refactoring?) -- indeed, its caller clearly expects it to return a BOOL.

fingolfin: I refactored the code in question in PR #3815

src/hpc/traverse.c
fingolfin: I'll change this particular function to return TRUE/FALSE now; but I never intended for this PR to be the final word on BOOL/TRUE/FALSE, but rather the beginning; in particular, there are lots of places that still use 0/1 not FALSE/TRUE. And I think that's fine overall (after all, comparisons like `foo < bar` also return 0 and 1), though it won't hurt to change it to TRUE/FALSE over time. But it's not necessary to do it all at once.

src/integer.h
fingolfin: I am reluctant to touch them because to preserve the line width limit one has to rewrap most of these comments, and it didn't seem worth the pain. But I wouldn't object if somebody else wanted to do this in another PR once this one is merged  :-)

src/lists.c
fingolfin: I changed this now, too, but as explained above, I don't really want to change all such places in this PR; I'd rather leave this to a follow up PR (or multiple)

src/objects.h
fingolfin: See above

src/blister.h
DominikBernhardt: My comment was more of a question than an intend to get you to change it back. I'm fine with it, I just wanted to know what the general philosophy was.

src/calls.h
DominikBernhardt: I guess it is also a personal preference. Yes, you're right, this I probably clear enough and we can leave it as is.

src/hpc/traverse.c
DominikBernhardt: Ok, fair enough. You're right, one could do that in future PRs.

fingolfin:
Turns out the semigroups package uses #define BOOL... this breaks all. So either we rename BOOL, or semigroups needs to change

DominikBernhardt:
Oh bummer. We could replace it with something like `GAPBOOL` or even `GAP_BOOL`, although both are longer and look a bit clumsier..

DominikBernhardt:
@fingolfin Am I stupid, but when I grep for `BOOL` in semigroups, I only find `BOOL_MAT`?

fingolfin:
It seems to be in libsemigroup, see Travis build logs

fingolfin:
I've added a workaround for the conflict with Semigroups (namely, I added `#undef BOOL` to `common.h`), and also submitted https://github.com/gap-packages/Semigroups/pull/635

src/system.c
```diff
@@ -42,6 +42,9 @@
 #include <sys/stat.h>
 
 #ifdef SYS_IS_DARWIN
+// Workaround for the fact that TRUE / FALSE area also
```
DominikBernhardt:
```suggestion
// Workaround for the fact that TRUE / FALSE are also
```

DominikBernhardt:
Thanks! Now that [gap-packages/Semigroups#635](https://github.com/gap-packages/Semigroups/pull/635) is merged, are you planning to remove the workaround in this PR?

fingolfin:
I can only remove the workaround once a new Semigroups release has happened *and* the package distribution has it. Anyway, I think we might want to keep the workaround for some time (perhaps 1-2 full release cycles), as it allows people to keep using their existing semigroups installation.

src/system.c
fingolfin: Oops, fixed, thanks

fingolfin:
One thing we did not discuss is the name of that type. One might argue that the name `Bool` would fit better than `BOOL` (more similar to `Int`, `Char`, `Obj`, `Bag`, ...).  I think it would also conflict with fewer external sources! The reason I chose `BOOL` was that it seemed to fit better with `TRUE` and `FALSE`. And those were chosen because `True` and `False` already exist, of type `Obj` . I was concerned that it might be confusing that `True` is not of type `Bool` but `TRUE` is. While `TRUE` is `BOOL` but `True` is `Obj` seems more plausible.

fingolfin:
@wucas `bool` is a type provided by C++, and also by C99 (if one includes `stdbool.h`), see e.g. <https://en.cppreference.com/w/c/types/boolean>. The nice thing about using `bool` would be that it is standardized, and well-established in C++ code bases. The drawback for our purposes however is that a lot of things are automatically treated as `bool` by the C compiler. This is a feature in general, but undesirable for us: The feature is meant to ensure that things like `if (ptr) { ... }` just "work": `if(expr)` expects an expression evaluating to `bool`; but a `ptr` is not a `bool`; but in C, there is the implicit assumption that a null pointer is `false` and anything else is `true`, to ensure this kind of check (which is ubiquitous in C code) keeps working as expected. But for us, that's a problem, because we have global constants `True` and `False`, which are pointers to GAP objects representing `True` and `False`. So it would be quite easy to accidentally mixup `True` and `False` with `true` and `false`; and due to the above "feature" of C, we can only catch a subset of these wrong uses, even with the strictest compiler warnings enabled.

This is why this PR does not use the standard type `bool`, but rather invents its own type, `BOOL`.

wucas:
>@wucas `bool` is a type provided by C++, and also by C99 (if one includes `stdbool.h`), see e.g. <https://en.cppreference.com/w/c/types/boolean>. The nice thing about using `bool` would be that it is standardized, and well-established in C++ code bases. The drawback for our purposes however is that a lot of things are automatically treated as `bool` by the C compiler. This is a feature in general, but undesirable for us: The feature is meant to ensure that things like `if (ptr) { ... }` just "work": `if(expr)` expects an expression evaluating to `bool`; but a `ptr` is not a `bool`; but in C, there is the implicit assumption that a null pointer is `false` and anything else is `true`, to ensure this kind of check (which is ubiquitous in C code) keeps working as expected. But for us, that's a problem, because we have global constants `True` and `False`, which are pointers to GAP objects representing `True` and `False`. So it would be quite easy to accidentally mixup `True` and `False` with `true` and `false`; and due to the above "feature" of C, we can only catch a subset of these wrong uses, even with the strictest compiler warnings enabled. ...

@fingolfin thank you very much for your answer I think I understand it now. Also my approval still stands.

wilfwilson:
I think `BOOL` is probably the best choice at the moment but don’t care that much.

ChrisJefferson:
I'm happy with this. Seeing as we can't compile-time check if BOOL, TRUE and FALSE are used consistently/correctly, I'm not positive how much gain we will get, but I don't think it makes anything worse.

fingolfin:
@ChrisJefferson For me the main thing to gain is that it now is obvious that certain functions expect or return a boolean value, and that certain variables only take on boolean values (see also the recent question by @wucas on PR #3823); I also don't have to fret whether to make boolean variables be of type int, Int, UInt, char, ... and don't have to fret over code where these are all mixed. So for me, the main advantage is that it actually simplifies for me, mentally. Of course that may be highly subjective; but as long as nobody here feels actively *hindered* by this, I am hoping it'll be net positive, albeit indeed a small one.

fingolfin:
That said, I wish somebody would make an C variant where there is a type `bool` and it is a non-integral type, with no implicit conversion between it and `int`, pointers, whatever. No other change to the language. Then I could adapt all my C code to be compliant with it, by turning things like `if (ptr)` into `if (ptr != 0)` etc. If one does this right, the resulting code would still be valid C code and compile warning free, but I'd be sure I didn't accidentally mix up booleans and non-booleans.

I think there are a few other similar minor changes one could make; restrictions to the language where the resulting code still would be valid C code, and thus usable in lots of places, but one removes a "feature" that perhaps once seemed convenient, but now is burdensome. Unfortunately, people who look into such things apparently can't resist the urge to start making bigger and bigger changes until they arrive at a completely different language, like D. Ah well :-)

PR #3816, opened 2019-12-30
ENHANCE: maintain information about tail indices to relations

hulpke:
Export information about tails as corresponding to relations in the confluent rewriting system, thus easing interfacing to the data.


fingolfin:
Backported in commit 3e11c3018

PR #3817, opened 2019-12-31
Another assortment of kernel tweaks

fingolfin:


src/sysfiles.h
```diff
@@ -7,8 +7,8 @@
 **
 **  SPDX-License-Identifier: GPL-2.0-or-later
 **
-**  The  file 'system.c'  declares  all operating system  dependent functions
-**  except file/stream handling which is done in "sysfiles.h".
+**  This file declares operating system dependent functions dealing with file
+**  and stream operations,
```
wilfwilson: Full stop rather than comma?

src/streams.c
```diff
@@ -427,7 +427,7 @@ static void READ_TEST_OR_LOOP(Obj context)
 Int READ_GAP_ROOT ( const Char * filename )
 {
     TypGRF_Data         result;
-    Int                 res;
+    int                 res;
```
DominikBernhardt: What his the reason for changing this?

fingolfin: That's to avoid a compiler warning in the `Panic` call, where `%d` is used which should indicate an `int` argument; on most 64bit systems, `Int` is a bigger type. Anyway, this part of the PR actually conflicts with PR #3818, so I'll remove this here.

PR #3818, opened 2019-12-31
kernel: avoid using GAP_PATH_MAX in headers; add sysroots.{c,h}

fingolfin:
Since the definition of `GAP_PATH_MAX` is system dependent, avoiding its use in headers helps to stabilize the ABI of GAP, and removes one (very minor) obstacle for `make install`. That said, we probably could have gotten this also by simply always defining `GAP_PATH_MAX` to be 4096 or so, but I think there is independent value in these changes:
- consolidating code dealing with `SyGapRootPaths` in `sysroots.c` allows us to make `SyGapRootPaths` static and thus hide an implementation detail, allow for future refactoring there
- similarly it is good that `DotGapPath` is now hidden
- the changes to `SyCompileInput` etc. simplify the code and avoid pointless copying around of data in buffers, and also nicely sidesteps the problem of using appropriate buffer sizes.

### DIFF
--- a/src/sysroots.c
+++ b/src/sysroots.c
@@ -1,0 +1,225 @@
+/****************************************************************************
+**
+**  This file is part of GAP, a system for computational discrete algebra.
+**
+**  Copyright of GAP belongs to its developers, whose names are too numerous
+**  to list here. Please refer to the COPYRIGHT file for details.
+**
+**  SPDX-License-Identifier: GPL-2.0-or-later
+**
+*/
+
+#include "sysroots.h"
+
+#include "gaputils.h"
+#include "plist.h"
+#include "stringobj.h"
+#include "sysfiles.h"
+#include "sysstr.h"
+
+
+/****************************************************************************
+**
+*V  SyGapRootPaths  . . . . . . . . . . . . . . . . . . . array of root paths
+**
+**  'SyGapRootPaths' conatins the  names   of the directories where   the GAP
+**  files are located.
+**
+**  It is modified by the command line option -l.
+**
+**  It is copied into the GAP variable 'GAPInfo.RootPaths' and used by
+**  'SyFindGapRootFile'.
+**
+**  Each entry must end  with the pathname seperator, eg.  if 'init.g' is the
+**  name of a library file 'strcat( SyGapRootPaths[i], "lib/init.g" );'  must
+**  be a valid filename.
+*/
+enum { MAX_GAP_DIRS = 16 };
+static Char SyGapRootPaths[MAX_GAP_DIRS][GAP_PATH_MAX];
+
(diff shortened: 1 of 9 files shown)


src/sysroots.c
```diff
@@ -0,0 +1,230 @@
...
+**  'SyFindGapRootFile'.
+**
+**  Each entry must end  with the pathname seperator, eg.  if 'init.g' is the
+**  name of a library file 'strcat( SyGapRootPaths[i], "lib/init.g" );'  must
+**  be a valid filename.
+**
+**  In addition we store the path to the users ~/.gap directory, if available,
+**  in 'DotGapPath'.
+**
+**  Put in this package because the command line processing takes place here.
```
fingolfin: Fixme

src/sysroots.c
```diff
@@ -0,0 +1,230 @@
...
+**
+**  It is copied into the GAP variable 'GAPInfo.RootPaths' and used by
+**  'SyFindGapRootFile'.
+**
+**  Each entry must end  with the pathname seperator, eg.  if 'init.g' is the
+**  name of a library file 'strcat( SyGapRootPaths[i], "lib/init.g" );'  must
+**  be a valid filename.
+**
+**  In addition we store the path to the users ~/.gap directory, if available,
+**  in 'DotGapPath'.
```
fingolfin: FIXME

src/sysroots.c
```diff
@@ -0,0 +1,230 @@
...
+        }
+        strxcat(SyGapRootPaths[n - 1], "hpcgap/",
+                sizeof(SyGapRootPaths[n - 1]));
+        n++;
+#endif
+    }
+
+    // replace leading tilde ~ by HOME environment variable
+    // TODO; instead of iterating over all entries each time, just
+    // do this for the new entries
```
fingolfin: FYI I left this code this way as it was the minimal change to achieve the goal, with least risk of introducing a regression.

Overall it would be nicer to store all these paths as GAP objects (strings and plain lists), but GASMAN is not yet initialized when this function is called... because it is called when we parse command line options, and those include settings for GASMAN... one solution to that would be to parse command line options twice, once for GASMAN options, then again for everything else...

src/sysroots.h
```diff
@@ -0,0 +1,53 @@
...
+**  enough, then <buf> is set to an empty string and NULL is returned.
+*/
+Char * SyFindGapRootFile(const Char * filename, Char * buf, size_t size);
+
+
+/****************************************************************************
+**
+*F  SyGetGapRootPaths()
+*/
+Obj SyGetGapRootPaths(void);
```
fingolfin: TODO add explanation

PR #3819, opened 2020-01-01
kernel: move ReaderState out of GAPState and onto the stack

fingolfin:
This way, saving and restoring it becomes unnecessary. It also gets easier to reason about the behaviour of the code.

Note: the second commit in this PR should be removed before merging; it contains a gross `#define` hack which I added to make reviewing easier, by reducing the noise in which the actual changes are otherwise lost. It's still quite a bit, but hopefully manageable.


src/intrprtr.c
```diff
@@ -657,14 +657,14 @@ void            IntrIfEnd (
...
-void IntrForBegin ( void )
+void IntrForBegin(Obj stackNams)
 {
     /* ignore                                                              */
     SKIP_IF_RETURNING();
     SKIP_IF_IGNORING();
 
     if (STATE(IntrCoding) == 0)
-        StartFakeFuncExpr(0);
+        StartFakeFuncExpr(stackNams, 0);
```
fingolfin: Note that `StartFakeFuncExpr` previously accessed the ReaderState, to be able to access `StackNams`. But since `StartFakeFuncExpr ` is called from the interpreter, which doesn't have access to the ReaderState (and I'd like to keep it that way), we can't just pass the ReaderState to `StartFakeFuncExpr`. The solution then is to just pass `StackNams` to a few select interpreter functions.

All in all, I am not really satisfied with this solution; but then, I've grappled with this very problem (which boils down to the interpreter calling back to the reader, which otherwise doesn't happen) for quite some time (1-2 years?) and couldn't come up with a satisfactory solution so far. The only other approach (besides the one I used here) which I didn't hate completely was to add return values to `IntrForBegin` and friends to indicate to the reader that it should call  `StartFakeFuncExpr`. resp. `FinishAndCallFakeFuncExpr`. But this relies on `CodeForBegin` etc. to not do anything (which is true, and probably *safe* to rely on, but still not "nice"; but OK, I could live with that). In any case, it felt to me that it just made it even harder to reason about what's going on... 

That said if somebody has an idea how to do this "better" / "differently", I am eager to hear, so please don't hesitate to speak up!

src/read.c
```diff
@@ -37,6 +37,8 @@
 
 struct ReaderState {
 
+ScannerState s;
```
fingolfin: The ScannerState becomes part of the ReaderState (I've intended to do it this way for quite some time now - finally it happens, yay!)

One way to make this perhaps slightly nicer, and avoid some of the diffs in this PR permanently (i.e. not just with my crappy HACK commit) would be to switch `read.c` to C++, and then make `ReaderState` into a "subclass" of `ScannerState`. This way, a `ReaderState *` can be passed to any function expecting a `ScannerState`. 

But I have doubts whether that's enough to justify converting the whole file to C++...  ?!

src/read.c
```diff
@@ -2555,11 +2551,7 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
     STATE(Prompt) = SyQuiet ? "" : "> ";
 
     /* remember the old reader context                                     */
-    stackNams   = rs->StackNams;
-    readTop     = rs->ReadTop;
-    readTilde   = rs->ReadTilde;
     tilde       = STATE(Tilde);
-    currLHSGVar = rs->CurrLHSGVar;
```
fingolfin: Here something "real" happens: we avoid the need to backing up parts of the ReaderState (and we also avoid questions about why we just save *those* parts of the Reader, but not others... Now we "save" all of them, automatically.)

src/read.c
```diff
@@ -2788,10 +2759,6 @@ void ReadEvalError(void)
 
 
 struct SavedReaderState {
-  Obj                 stackNams;
-  UInt                readTop;
-  UInt                readTilde;
-  UInt                currLHSGVar;
   UInt                userHasQuit;
   syJmp_buf           readJmpError;
   UInt                intrCoding;
```
fingolfin: Hmm, I guess we should rename `SavedReaderState` to something else... 

Note that I am thinking about moving `STATE(IntrCoding)` and friends into a new `InterpreterState` (which ultimately also would be stack allocated...)

fingolfin: This was done in PR #3823 which also removed `SavedReaderState`

src/read.c
```diff
@@ -83,12 +85,7 @@ UInt LoopNesting;
 
 };
 
-static ModuleStateOffset ReaderStateOffset = -1;
-
-extern inline struct ReaderState * ReaderState(void)
-{
-    return (struct ReaderState *)StateSlotsAtOffset(ReaderStateOffset);
-}
+typedef struct ReaderState ReaderState;
```
fingolfin: Here we can see that there was a (per-thread) instance of  `struct ReaderState` inside the global GAPState (accessed here via `StateSlotsAtOffset`), and now isn't anymore. `ReaderState` used to refer to a helper function to access it; now `ReaderState` is an alias for `struct ReaderState`, and we instead pass a stack-allocated instance of `struct ReaderState` around via function arguments

src/read.c
```diff
@@ -98,32 +95,30 @@ extern inline struct ReaderState * ReaderState(void)
 **  is <Statements>. The functions 'ReadExpr' and 'ReadStats' must therefore
 **  be declared forward.
 */
-static void ReadExpr(ScannerState * s, TypSymbolSet follow, Char mode);
+static void ReadExpr(ReaderState * rs, TypSymbolSet follow, Char mode);
```
fingolfin: From here on, we see a lot of changes identical to this: instead of just a `ScannerState`, we now pass around a `ReaderState` (which contains a `ScannerState`).

src/read.c
```diff
@@ -98,32 +95,30 @@ extern inline struct ReaderState * ReaderState(void)
...
     rs->CurrentGlobalForLoopDepth++;
 }
 
-static void PopGlobalForLoopVariable(void)
+static void PopGlobalForLoopVariable(ReaderState * rs)
 {
-    GAP_ASSERT(ReaderState()->CurrentGlobalForLoopDepth);
-    ReaderState()->CurrentGlobalForLoopDepth--;
+    GAP_ASSERT(rs->CurrentGlobalForLoopDepth);
+    rs->CurrentGlobalForLoopDepth--;
```
fingolfin: We also see a lot of changes identical to this: instead of using `ReaderState()`, we now use a `ReaderState` instance named `rs` which the callers passes down to us.

src/read.c
```diff
@@ -179,7 +178,7 @@ static void ReadFuncCallOption(ScannerState * s, TypSymbolSet follow)
     }
     else if (s->Symbol == S_LPAREN) {
         Match(s, S_LPAREN, "(", S_COMMA | follow);
-        ReadExpr(s, follow, 'r');
+        ReadExpr(rs, follow, 'r');
```
fingolfin: Aaaand we also see a lot of  these: instead of passing a `ScannerState` named `s`, we pass on a `ReaderState` called `rs`.

src/read.c
```diff
@@ -161,6 +156,10 @@ static UInt findValueInNams(Obj nams, const Char * val, UInt start, UInt end)
 */
 
 
+// HACK
+#define s (&rs->s)
```
fingolfin: This HACK is part of the second commit (which is meant to be dropped before merging this), and only is here for review purposes, to reduce the number of meaningless diffs: where before we directly accessed a `ScannerState` named `s`, we now instead need to get the `ScannerState` from inside the `ReaderState` called `rs`. To hide this, we use this `#define`.

src/read.c
```diff
@@ -2466,22 +2468,16 @@ static void RecreateStackNams( Obj context )
...
 #endif
 
-    struct ReaderState * volatile rs = ReaderState();
-
-    ScannerState scanner;
-    ScannerState * volatile s = &scanner;
-    memset(s, 0, sizeof(ScannerState));
+    ReaderState reader;
+    ReaderState * volatile rs = &reader;
+    memset(rs, 0, sizeof(ReaderState));
```
fingolfin: Up to here, basically nothing happened besides the things I explicitly pointed out; 99% of the changes are change `ScannerState * s` -> `ReaderState * rs` and related changes.

Here now we allocate a `ReaderState` on the stack, and zero it.

src/read.c
```diff
@@ -2573,11 +2565,7 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
 
     /* switch back to the old reader context                               */
     memcpy( STATE(ReadJmpError), readJmpError, sizeof(syJmp_buf) );
-    rs->StackNams      = stackNams;
-    rs->ReadTop        = readTop;
-    rs->ReadTilde      = readTilde;
     STATE(Tilde)       = tilde;
-    rs->CurrLHSGVar    = currLHSGVar;
```
fingolfin: Real change: since we freshly allocated the `ReaderState` on the stack, there is no need to "restore" its content before returning.

src/read.c
```diff
@@ -2701,11 +2674,7 @@ UInt ReadEvalFile(Obj *evalResult)
     if (TLS(CurrentHashLock))
       HashUnlock(TLS(CurrentHashLock));
 #endif
-    rs->StackNams    = stackNams;
-    rs->ReadTop      = readTop;
-    rs->ReadTilde    = readTilde;
     STATE(Tilde)     = tilde;
-    rs->CurrLHSGVar  = currLHSGVar;
```
fingolfin: Real change: since we freshly allocated the `ReaderState` on the stack, there is no need to "restore" its content before returning.

src/read.c
```diff
@@ -2663,13 +2641,8 @@ UInt ReadEvalFile(Obj *evalResult)
...
     /* read the statements                                                 */
-    {
-        UInt oldLoopNesting = rs->LoopNesting;
-        rs->LoopNesting = 0;
-        nr = ReadStats(s, S_SEMICOLON | S_EOF);
-        rs->LoopNesting = oldLoopNesting;
-    }
-
+    GAP_ASSERT(rs->LoopNesting == 0);
+    nr = ReadStats(rs, S_SEMICOLON | S_EOF);
```
fingolfin: Real change: since we freshly allocated the `ReaderState` on the stack, there is no need to "restore" its `LoopNesting` member here.

src/read.c
```diff
@@ -2629,11 +2611,7 @@ UInt ReadEvalFile(Obj *evalResult)
     STATE(Prompt) = SyQuiet ? "" : "> ";
 
     /* remember the old reader context                                     */
-    stackNams   = rs->StackNams;
-    readTop     = rs->ReadTop;
-    readTilde   = rs->ReadTilde;
     tilde       = STATE(Tilde);
-    currLHSGVar = rs->CurrLHSGVar;
```
fingolfin: Real change: since we freshly allocated the `ReaderState` on the stack, there is no need to "preserve" its "previous" content.

src/read.c
```diff
@@ -2611,11 +2595,9 @@ UInt ReadEvalFile(Obj *evalResult)
...
 #endif
 
-    struct ReaderState * volatile rs = ReaderState();
-
-    ScannerState scanner;
-    ScannerState * volatile s = &scanner;
-    memset(s, 0, sizeof(ScannerState));
+    ReaderState reader;
+    ReaderState * volatile rs = &reader;
+    memset(rs, 0, sizeof(ReaderState));
```
fingolfin: Real change: we allocate a `ReaderState` on the stack and zero it.

src/read.c
```diff
@@ -2760,10 +2723,6 @@ static void SaveReaderState(struct SavedReaderState *s) {
 }
 
 static void ClearReaderState(void ) {
-  ReaderState()->StackNams   = NEW_PLIST( T_PLIST, 16 );
-  ReaderState()->ReadTop     = 0;
-  ReaderState()->ReadTilde   = 0;
-  ReaderState()->CurrLHSGVar = 0;
```
fingolfin: No need to preserve or restore these, as we stack allocate the `ReaderState` for each use freshly on the stack.

src/read.c
```diff
@@ -2774,10 +2733,6 @@ static void ClearReaderState(void ) {
 static void RestoreReaderState(const struct SavedReaderState *s) {
   memcpy( STATE(ReadJmpError), s->readJmpError, sizeof(syJmp_buf) );
   STATE(UserHasQuit) = s->userHasQuit;
-  ReaderState()->StackNams   = s->stackNams;
-  ReaderState()->ReadTop     = s->readTop;
-  ReaderState()->ReadTilde   = s->readTilde;
-  ReaderState()->CurrLHSGVar = s->currLHSGVar;
```
fingolfin: No need to preserve or restore these, as we stack allocate the `ReaderState` for each use freshly on the stack.

src/read.c
```diff
@@ -2875,9 +2830,6 @@ Obj Call1ArgsInNewReader(Obj f,Obj a)
 static Int InitKernel (
     StructInitInfo *    module )
 {
-#if !defined(HPCGAP)
-    InitGlobalBag(&ReaderState()->StackNams, "src/read.c:StackNams");
```
fingolfin: `StackNams` now is in a `ReaderState` instance on the stack, and hence our stack scanning code ensure it is not garbage collected. So this line can go.

src/read.c
```diff
@@ -2886,11 +2838,6 @@ static Int InitKernel (
 static Int InitModuleState(void)
 {
     STATE(ErrorLVars) = (UInt **)0;
-    ReaderState()->StackNams = NEW_PLIST(T_PLIST, 16);
-    ReaderState()->ReadTop = 0;
-    ReaderState()->ReadTilde = 0;
-    ReaderState()->CurrLHSGVar = 0;
-    ReaderState()->CurrentGlobalForLoopDepth = 0;
```
fingolfin: There is no per-thread `ReaderState` as part of the `GAPState` anymore, so the above can go.

src/read.c
```diff
@@ -2907,8 +2854,6 @@ static StructInitInfo module = {
     .name = "read",
     .initKernel = InitKernel,
 
-    .moduleStateSize = sizeof(struct ReaderState),
-    .moduleStateOffsetPtr = &ReaderStateOffset,
```
fingolfin: There is no per-thread `ReaderState` as part of the `GAPState` anymore, so the above can go.

fingolfin:
I've now left a bunch of comments that try to explain the changes in-depth. Perhaps that's enough to make some people dare to look at it, e.g. @DominikBernhardt @wucas @flyingapfopenguin @ChrisJefferson

src/read.c
ChrisJefferson: Could we do reader = {0} instead?

fingolfin: I actually tried this first, but it caused compiler warnings on Travis. I couldn't reproduce those on macOS or on my Ubuntu, probably because my versions of GCC were too new.

ChrisJefferson: No problem. Stupid travis, but it's not worth adding code which will have warnings.

PR #3820, opened 2020-01-01
kernel: move most system #includes out of common.h

fingolfin:
... and into the files that actually need them. That minimizes the risk of header conflicts, and also may be desirable for who code linking against GAP which may for some reason or another not want to #include certain system headers.


ChrisJefferson:
This should be checked it doesn't break any packages which are assuming those headers are included by gap?

fingolfin:
That's among the things Travis tests for

fingolfin:
So at the very least, the `io` and `profiling` packages are affected by this, both are in need of an `#include <stdio.h>`

fingolfin:
I added a HACK commit to workaround the issue by adding `#include <stdio.h>` into `src/compiled.h`. This shouldn't be merged as is, but will allow me to find out which others packages might be affected.

ChrisJefferson:
Will merge when conflicts are fixed.

fingolfin:
Done, but let's see if tests still pass

PR #3822, opened 2020-01-01
Make the assertion level (which controls `Assert` statements) thread local when using HPC-GAP

### LINKED ISSUE #3821: Make CurrentAssertionLevel thread local
Right now, `CurrentAssertionLevel` is global. I think this is problematic in HPC-GAP (ping @rbehrends @stevelinton). So I suggest we make it thread local instead.

The easiest way to do that would be to remove it as a GAP variable, and instead put it into the kernel, as a member of the thread local `GAPState`. That would also require moving `SetAssertionLevel` and `AssertionLevel` from the library to the kernel, but that's easy. I can't think of any downsides?

fingolfin:
... by moving `CurrentAssertionLevel` to the kernel, into the GAPState.

While at it, also validate the level argument given to Assert(), and error out if it is not a non-negative small integer.

Also fix a potential issue in gac where INT_INTOBJ was called on inputs without validation; we now use Int_ObjInt instead, which produces an error if the input is not a small integer.

Resolves #3821


ChrisJefferson:
One first comment, That INT_INTOBJ -> Int_ObjInt change I like, but I'd like it in a separate commit (either in this PR, or seperately).

Last time I edited this code I left the GAP-level CurrentAssertionLevel, because I was worried someone might be using it. Any code which was using it will now silently just change a GAP global variable without doing anything, which slightly worries me.

Could we instead just make the variable thread-local at the GAP level?

fingolfin:
@ChrisJefferson split into two commits (good idea, thanks). Making `CurrentAssertionLevel` into a GAP visible TLS variable would have substantially complicated the code. Instead, I chose to `BindConstant` it to the value `false`. This way, any attempts to modify it, or to compare it integers, will raise an error.

src/gap.c
```diff
@@ -1387,6 +1387,20 @@ static Obj FuncUPDATE_STAT(Obj self, Obj name, Obj newStat)
...
+{
+    RequireNonnegativeSmallInt("SetAssertionLevel", level);
+    STATE(CurrentAssertionLevel) = INT_INTOBJ(level);
+    return 0;
+}
+
+
+static Obj FuncAssertionLevel(Obj self)
+{
+    return INTOBJ_INT(STATE(CurrentAssertionLevel));
```
fingolfin: BTW, placing these into `gap.c` was somewhat arbitrary. If somebody has a preference for another location, let me know.

tst/testinstall/kernel/stats.tst
```diff
@@ -12,6 +12,8 @@ gap> f('x');
 Error, Range: <last> must be a small integer (not a character)
 
 # test Assert with two arguments
+gap> function() Assert(fail, 0); end();
+Error, Assert: <lev> must be a small integer (not the value 'fail')
```
fingolfin: For the record, I am using `<lev>` here as that matches what the documentation for `Assert` calls that argument (that's my general rule for the name of the arguments in error messages). I am fine with changing it here, e.g. to `level`, but IMHO it then should be changed in the documentation as well.

PR #3823, opened 2020-01-02
kernel: refactor interpreter state handling

fingolfin:
This continues refactoring work begun in PRs #3805 and #3819.

Moving the interpreter state into a dedicated struct which is allocated on the C stack has multiple advantages; the latter commits in this PR exploit several of them:
- we can get rid of `ExecBegin`/`ExecEnd`
- we can simplify `IntrBegin`/`IntrEnd`
- we can reduce the number of places calling `IntrBegin`/`IntrEnd` from 5 down to 2, both in `read.c` and in places where they clearly belong
- I suspect we did not actually save/restore the interpreter state quite completely/correctly in a few places; these are now taken care of completely automatically by virtue of the interpreter state being on the stack
- ...

UPDATE: I've removed all but one commit from this PR, to keep it simple. I'll add the other back in a separate PR (or more) once this PR here has been merged.

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -35,11 +35,6 @@ typedef struct GAPState {
 #endif
 
     /* From intrprtr.c */
-    UInt IntrIgnoring;
-    UInt IntrReturning;
-    UInt IntrCoding;
-    Obj  IntrState;
-    Obj  StackObj;
     Obj  Tilde;
 
     // The current assertion level for use in Assert
--- a/src/hpc/threadapi.c
+++ b/src/hpc/threadapi.c
@@ -432,6 +432,7 @@ static GVarDescriptor GVarTHREAD_EXIT;
 
 static void ThreadedInterpreter(void * funcargs)
 {
+    IntrState intr = { 0, 0, 0, 0 };
     Obj tmp, func;
     int i;
 
@@ -439,7 +440,7 @@ static void ThreadedInterpreter(void * funcargs)
     STATE(NrError) = 0;
     STATE(ThrownObject) = 0;
 
-    IntrBegin(STATE(BottomLVars));
+    IntrBegin(&intr, STATE(BottomLVars));
     tmp = KEPTALIVE(funcargs);
     StopKeepAlive(funcargs);
     func = ELM_PLIST(tmp, 1);
@@ -461,13 +462,13 @@ static void ThreadedInterpreter(void * funcargs)
         exit = GVarOptFunction(&GVarTHREAD_EXIT);
         if (exit)
             CALL_0ARGS(exit);
-        PushVoidObj();
+        PushVoidObj(&intr);
         /* end the interpreter */
-        IntrEnd(0, NULL);
+        IntrEnd(&intr, 0, NULL);
     }
     CATCH_ERROR
     {
-        IntrEnd(1, NULL);
+        IntrEnd(&intr, 1, NULL);
         ClearError();
     }
 }
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -47,58 +47,20 @@
 #include "hpc/guards.h"
 #endif
 
-/****************************************************************************
-**
-*V  IntrReturning   . . . . . . . . . . .  interpreter is currently returning
-**
-**  If 'IntrReturning' is  non-zero, the interpreter is currently  returning.
-**  The interpreter switches  to this mode when  it finds a return-statement.
-**  If it interprets a return-value-statement, it sets 'IntrReturning' to 1.
-**  If it interprets a return-void-statement,  it sets 'IntrReturning' to 2.
-**  If it interprets a quit-statement, it sets 'IntrReturning' to 8.
-*/
-/* TL: UInt IntrReturning; */
-
-
-/****************************************************************************
-**
-*V  IntrIgnoring  . . . . . . . . . interpreter is currently ignoring actions
-**
-**  If 'IntrIgnoring'  is  non-zero,  the interpreter  is  currently ignoring
-**  actions.  The interpreter switches to this mode for  the right operand of
-**  'or' and 'and'  constructs where the  left operand already determines the
-**  outcome.
-**
-**  This mode is also used in Info and Assert, when arguments are not printed.
-*/
-/* TL: UInt IntrIgnoring; */
-
-
-/****************************************************************************
-**
-*V  IntrCoding  . . . . . . . . . . . interpreter is currently coding actions
-**
-**  If 'IntrCoding' is non-zero, the interpreter is currently coding actions.
-**  The interpreter  switches  to this  mode for  constructs  that it  cannot
-**  directly interpret, such as loops or function bodies.
-*/
-/* TL: UInt IntrCoding; */
-
 // INTERPRETER_PROFILE_HOOK deals with profiling of immediately executed
 // code.
-// If STATE(IntrCoding) is true, profiling is handled by the AST
+// If intr->coding is true, profiling is handled by the AST
 // generation and execution. Otherwise, we always mark the line as
-// read, and mark as executed if STATE(IntrReturning) and STATE(IntrIgnoring)
+// read, and mark as executed if intr->returning and intr->ignoring
 // are both false.
 //
 // IgnoreLevel gives the highest value of IntrIgnoring which means this
 // statement is NOT ignored (this is usually, but not always, 0)
-static void INTERPRETER_PROFILE_HOOK(int ignoreLevel)
+static void INTERPRETER_PROFILE_HOOK(IntrState * intr, int ignoreLevel)
 {
-    if (!STATE(IntrCoding)) {
+    if (!intr->coding) {
         InterpreterHook(GetInputFilenameID(), STATE(InterpreterStartLine),
-                        STATE(IntrReturning) ||
-                            (STATE(IntrIgnoring) > ignoreLevel));
+                        intr->returning || (intr->ignoring > ignoreLevel));
     }
     STATE(InterpreterStartLine) = 0;
 }
@@ -107,16 +69,19 @@ static void INTERPRETER_PROFILE_HOOK(int ignoreLevel)
 // Put the profiling hook into SKIP_IF_RETURNING, as this is run in
 // (nearly) every part of the interpreter, avoid lots of extra code.
 #define SKIP_IF_RETURNING()                                                  \
-    INTERPRETER_PROFILE_HOOK(0);                                             \
+    INTERPRETER_PROFILE_HOOK(intr, 0);                                       \
     SKIP_IF_RETURNING_NO_PROFILE_HOOK();
 
 // Need to
 #define SKIP_IF_RETURNING_NO_PROFILE_HOOK()                                  \
-    if (STATE(IntrReturning) > 0) {                                          \
+    if (intr->returning > 0) {                                               \
         return;                                                              \
     }
 
-#define SKIP_IF_IGNORING()  if ( STATE(IntrIgnoring)  > 0 ) { return; }
+#define SKIP_IF_IGNORING()                                                   \
+    if (intr->ignoring > 0) {                                                \
+        return;                                                              \
+    }
 
 
 /****************************************************************************
@@ -140,20 +105,13 @@ static void INTERPRETER_PROFILE_HOOK(int ignoreLevel)
 **
 **  'PopVoidObj' returns the  top element from the values  stack and pops it.
 **  It is an error if the stack is empty but not if the top element is void.
-**
-**  Since interpreters  can nest, there can   be more than one  values stack.
-**  The bottom  element of each values stack is the 'StackObj' which was
-**  active when the current interpreter was started and which will be made
-**  active again when the current interpreter will stop.
 */
-/* TL: Obj             IntrState; */
-
 /* TL: Obj             StackObj; */
 
-static void PushObj(Obj val)
+static void PushObj(IntrState * intr, Obj val)
 {
-    assert( val != 0 );
-    PushPlist( STATE(StackObj), val );
+    GAP_ASSERT(val != 0);
+    PushPlist(intr->StackObj, val);
 }
 
 /* Special marker value to denote that a function returned no value, so we
@@ -165,32 +123,32 @@ static void PushObj(Obj val)
  * so it will not see this magic value. */
 static Obj VoidReturnMarker;
 
-static void PushFunctionVoidReturn(void)
+static void PushFunctionVoidReturn(IntrState * intr)
 {
-    PushPlist( STATE(StackObj), (Obj)&VoidReturnMarker );
+    PushPlist(intr->StackObj, (Obj)&VoidReturnMarker);
 }
 
-void PushVoidObj(void)
+void PushVoidObj(IntrState * intr)
 {
-    PushPlist( STATE(StackObj), (Obj)0 );
+    PushPlist(intr->StackObj, 0);
 }
 
-static Obj PopObj(void)
+static Obj PopObj(IntrState * intr)
 {
-    Obj val = PopPlist( STATE(StackObj) );
+    Obj val = PopPlist(intr->StackObj);
 
     if (val == (Obj)&VoidReturnMarker) {
         ErrorQuit("Function call: <func> must return a value", 0, 0);
     }
 
     // return the popped value (which must be non-void)
-    assert( val != 0 );
+    GAP_ASSERT(val != 0);
     return val;
 }
 
-static Obj PopVoidObj(void)
+static Obj PopVoidObj(IntrState * intr)
 {
-    Obj val = PopPlist( STATE(StackObj) );
+    Obj val = PopPlist(intr->StackObj);
 
     // Treat a function which returned no value the same as 'void'
     if (val == (Obj)&VoidReturnMarker) {
@@ -202,9 +160,9 @@ static Obj PopVoidObj(void)
 }
 
 
-static void StartFakeFuncExpr(Obj stackNams, Int startLine)
+static void StartFakeFuncExpr(IntrState * intr, Obj stackNams, Int startLine)
 {
-    assert(STATE(IntrCoding) == 0);
+    GAP_ASSERT(intr->coding == 0);
 
     // switch to coding mode now
     CodeBegin();
@@ -232,9 +190,9 @@ static void StartFakeFuncExpr(Obj stackNams, Int startLine)
 }
 
 
-static void FinishAndCallFakeFuncExpr(Obj stackNams)
+static void FinishAndCallFakeFuncExpr(IntrState * intr, Obj stackNams)
 {
-    assert(STATE(IntrCoding) == 0);
+    GAP_ASSERT(intr->coding == 0);
 
     // code a function expression (with one statement in the body)
     CodeFuncExprEnd(1, 1);
@@ -252,7 +210,7 @@ static void FinishAndCallFakeFuncExpr(Obj stackNams)
     CALL_0ARGS(func);
 
     // push void
-    PushVoidObj();
+    PushVoidObj(intr);
 }
 
 
@@ -277,28 +235,23 @@ static void FinishAndCallFakeFuncExpr(Obj stackNams)
 **  return-void-statement was interpreted. If 'IntrEnd' returns 'STATUS_QUIT',
 **  then a quit-statement was interpreted.
 */
-void IntrBegin ( Obj frame )
+void IntrBegin(IntrState * intr, Obj frame)
 {
-    /* remember old interpreter state                                      */
-    if (!STATE(IntrState))
-        STATE(IntrState) = NEW_PLIST(T_PLIST, 16);
-    PushPlist(STATE(IntrState), STATE(StackObj));
-
     /* allocate a new values stack                                         */
-    STATE(StackObj) = NEW_PLIST( T_PLIST, 64 );
+    intr->StackObj = NEW_PLIST(T_PLIST, 64);
 
     /* must be in immediate (non-ignoring, non-coding) mode                */
-    assert( STATE(IntrIgnoring) == 0 );
-    assert( STATE(IntrCoding)   == 0 );
+    GAP_ASSERT(intr->ignoring == 0);
+    GAP_ASSERT(intr->coding == 0);
 
     /* no return-statement was yet interpreted                             */
-    STATE(IntrReturning) = 0;
+    intr->returning = 0;
 
     /* start an execution environment                                      */
     ExecBegin(frame);
 }
 
-ExecStatus IntrEnd(UInt error, Obj *result)
+ExecStatus IntrEnd(IntrState * intr, UInt error, Obj * result)
 {
     UInt                intrReturning;  /* interpreted return-statement?   */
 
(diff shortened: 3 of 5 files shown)


src/gap.h
```diff
@@ -51,8 +51,8 @@ void ViewObjHandler(Obj obj);
 **                      subroutines, explaining why evaluation, or execution
 **                      has terminated.
 **
-**  Values are powers of two, although I do not currently know of any
-**  cirumstances where they can get combined
+**  Values are powers of two; this is used to test a a given status for
```
wilfwilson:
```suggestion
**  Values are powers of two; this is used to test a given status for
```

src/gapstate.h
```diff
@@ -35,11 +35,6 @@ typedef struct GAPState {
 #endif
 
     /* From intrprtr.c */
-    UInt IntrIgnoring;
-    UInt IntrReturning;
-    UInt IntrCoding;
-    Obj  IntrState;
-    Obj  StackObj;
```
fingolfin: These variables are removed from the global GAPState and put into a stack allocated `IntrState`.

WEll, all except `STATE(IntrState)` which itself became superfluous.

src/hpc/threadapi.c
```diff
@@ -432,14 +432,15 @@ static GVarDescriptor GVarTHREAD_EXIT;
 
 static void ThreadedInterpreter(void * funcargs)
 {
+    IntrState intr = { 0, 0, 0, 0 };
```
fingolfin: Allocate `IntrState` on the stack and use it below.

src/read.c
```diff
@@ -2771,27 +2760,26 @@ static void RestoreReaderState(const struct SavedReaderState *s) {
 Obj Call0ArgsInNewReader(Obj f)
 
 {
-  /* for the new interpreter context: */
-/*  ExecStatus          type; */
   struct SavedReaderState s;
+  IntrState intr = { 0, 0, 0, 0 };
```
fingolfin: Allocate `IntrState` on the stack and use it below.

src/read.c
```diff
@@ -2809,9 +2797,8 @@ Obj Call0ArgsInNewReader(Obj f)
 Obj Call1ArgsInNewReader(Obj f,Obj a)
 
 {
-  /* for the new interpreter context: */
-/*ExecStatus          type; */
   struct SavedReaderState s;
+  IntrState intr = { 0, 0, 0, 0 };
```
fingolfin: Allocate `IntrState` on the stack and use it below.

src/read.c
```diff
@@ -39,6 +39,8 @@ struct ReaderState {
 
 ScannerState s;
 
+IntrState intr;
```
fingolfin: Add an instance of `IntrState` into `ReaderState`; it'll thus be stack-allocated as part of that in `ReadEvalCommand` below

src/read.c
```diff
@@ -2729,35 +2730,23 @@ void ReadEvalError(void)
 struct SavedReaderState {
   UInt                userHasQuit;
   syJmp_buf           readJmpError;
-  UInt                intrCoding;
-  UInt                intrIgnoring;
-  UInt                intrReturning;
```
fingolfin: We don't need to save/restore these here anymore, as the interpreter state is on the stack -- yay!

src/intrprtr.c
```diff
@@ -47,58 +47,21 @@
...
-/****************************************************************************
-**
-*V  IntrCoding  . . . . . . . . . . . interpreter is currently coding actions
-**
-**  If 'IntrCoding' is non-zero, the interpreter is currently coding actions.
-**  The interpreter  switches  to this  mode for  constructs  that it  cannot
-**  directly interpret, such as loops or function bodies.
-*/
-/* TL: UInt IntrCoding; */
-
```
fingolfin: The above comments are all already in the header file, so they can be removed here.

src/intrprtr.c
```diff
@@ -47,58 +47,21 @@
...
+// If intr->IntrCoding is true, profiling is handled by the AST
 // generation and execution. Otherwise, we always mark the line as
-// read, and mark as executed if STATE(IntrReturning) and STATE(IntrIgnoring)
+// read, and mark as executed if intr->IntrReturning and intr->IntrIgnoring
 // are both false.
 //
 // IgnoreLevel gives the highest value of IntrIgnoring which means this
 // statement is NOT ignored (this is usually, but not always, 0)
-static void INTERPRETER_PROFILE_HOOK(int ignoreLevel)
+static void INTERPRETER_PROFILE_HOOK(IntrState * intr, int ignoreLevel)
```
fingolfin: The majority of the changes below either add an `IntrState * intr` argument to functions; or pass `intr` on to other functions that need them; or replace `STATE(IntrFoo)` by `intr->foo`.

From here on, a long time nothing else happens... I think for the rest of the whole file `intrprtr.c`

src/intrprtr.c
```diff
@@ -277,28 +233,23 @@ static void FinishAndCallFakeFuncExpr(Obj stackNams)
...
-    PushPlist(STATE(IntrState), STATE(StackObj));
-
     /* allocate a new values stack                                         */
-    STATE(StackObj) = NEW_PLIST( T_PLIST, 64 );
+    intr->StackObj = NEW_PLIST( T_PLIST, 64 );
 
     /* must be in immediate (non-ignoring, non-coding) mode                */
-    assert( STATE(IntrIgnoring) == 0 );
-    assert( STATE(IntrCoding)   == 0 );
+    GAP_ASSERT(intr->ignoring == 0);
```
fingolfin: Oh yeah, I changed a few `assert` to `GAP_ASSERT` since I was touching those lines anyway

src/intrprtr.c
```diff
@@ -277,28 +233,23 @@ static void FinishAndCallFakeFuncExpr(Obj stackNams)
 **  return-void-statement was interpreted. If 'IntrEnd' returns 'STATUS_QUIT',
 **  then a quit-statement was interpreted.
 */
-void IntrBegin ( Obj frame )
+void IntrBegin(IntrState * intr,  Obj frame)
 {
-    /* remember old interpreter state                                      */
-    if (!STATE(IntrState))
-        STATE(IntrState) = NEW_PLIST(T_PLIST, 16);
-    PushPlist(STATE(IntrState), STATE(StackObj));
```
fingolfin: Real change: we don't need to push `StackObj` onto a stack here and restore it later, as the interpreter state contains `StackObj` and is itself allocated on the C calling stack.

src/intrprtr.c
```diff
@@ -140,20 +103,13 @@ static void INTERPRETER_PROFILE_HOOK(int ignoreLevel)
 **
 **  'PopVoidObj' returns the  top element from the values  stack and pops it.
 **  It is an error if the stack is empty but not if the top element is void.
-**
-**  Since interpreters  can nest, there can   be more than one  values stack.
-**  The bottom  element of each values stack is the 'StackObj' which was
-**  active when the current interpreter was started and which will be made
-**  active again when the current interpreter will stop.
 */
-/* TL: Obj             IntrState; */
```
fingolfin: As explained above, we don't need `STATE(IntrState)` anymore, so this comment on its predecessor (it used to be a simple global variable) can be removed, too.

src/intrprtr.c
```diff
@@ -330,34 +281,31 @@ ExecStatus IntrEnd(UInt error, Obj *result)
...
+        intr->ignoring = 0;
+        intr->coding   = 0;
 
         /* dummy result value (probably ignored)                           */
         if (result)
             *result = 0;
     }
 
-    // switch back to the old state
-    STATE(StackObj) = PopPlist(STATE(IntrState));
```
fingolfin: As explained above, we don't need `STATE(IntrState)` anymore.

src/intrprtr.c
fingolfin: Note that we did not save IntrIgnoring, IntrCoding, IntrReturning here, but relied on the caller to do so.

src/intrprtr.c
```diff
@@ -379,22 +328,21 @@ void IntrAbortCoding(Obj lvars)
...
 {
     /* ignore or code                                                      */
     SKIP_IF_RETURNING();
     SKIP_IF_IGNORING();
-    if ( STATE(IntrCoding)    > 0 ) { CodeFuncCallBegin(); return; }
-
+    if (intr->coding > 0) {
+        CodeFuncCallBegin();
+        return;
+    }
```
fingolfin: I accidentally clang-formatted this `if`. But that makes me wonder: since I am touching all these `if (IntrCoding)` etc. checks anyway, is there any harm in running `git clang-format` on them? Doesn't make the `git blame` experience worse here, does it? Just reviewing would be slightly more annoying, but that could be alleviated by putting those reformatting changes into a second commit.

src/intrprtr.c
```diff
@@ -584,101 +530,99 @@ void            IntrIfBegin ( void )
...
     /* ignore or code                                                      */
     SKIP_IF_RETURNING();
     SKIP_IF_IGNORING();
-    if ( STATE(IntrCoding)    > 0 ) { CodeIfElse(); return; }
+    if ( intr->coding    > 0 ) { CodeIfElse(); return; }
 
 
     /* push 'true' (to execute body of else-branch)                        */
-    PushObj( True );
+    PushObj(intr, True);
```
fingolfin: Another note: almost all changes which add `IntrState * intr` as a parameter, resp. pass `intr` as an argument, could be removed again if we turned `IntrState` into a simple C++ class, and turned `IntrIfElse` etc. into member functions of that.

src/intrprtr.c
```diff
@@ -1937,20 +1876,19 @@ void            IntrCharExpr (
 */
 static Obj GetFromStack(Obj cycle, Int j)
 {
-    return PopObj();
+    IntrState * intr = (IntrState *)cycle;
+    return PopObj(intr);
```
fingolfin: Another "real" change: we need to pass `intr` to `PopObj`. Luckily, `ScanPermCycle` (which calls `GetFromStack` as a callback) already allows us to pass in a pointer of our choice, namely `cycle`, so we abuse that to pass in `intr`.

src/intrprtr.c
```diff
@@ -1959,29 +1897,28 @@ void            IntrPermCycle (
...
-        const UInt countObj = LEN_PLIST(STATE(StackObj));
-        m = INT_INTOBJ( ELM_LIST( STATE(StackObj), countObj - nrx ) );
-        perm = ELM_LIST( STATE(StackObj), countObj - nrx - 1 );
+        const UInt countObj = LEN_PLIST(intr->StackObj);
+        m = INT_INTOBJ( ELM_LIST( intr->StackObj, countObj - nrx ) );
+        perm = ELM_LIST( intr->StackObj, countObj - nrx - 1 );
     }
 
-    m = ScanPermCycle(perm, m, 0, nrx, GetFromStack);
+    m = ScanPermCycle(perm, m, (Obj)intr, nrx, GetFromStack);
```
fingolfin: Real change: pass in `intr` to be used as `Cycle` argument for `GetFormStack`.

src/intrprtr.c
```diff
@@ -2346,53 +2272,55 @@ void            IntrRecExprEnd (
...
     UInt                rnam;           /* record name                     */
 
     /* ignore or code                                                      */
     SKIP_IF_RETURNING();
     SKIP_IF_IGNORING();
-    if ( STATE(IntrCoding)    > 0 ) { CodeFuncCallOptionsBeginElmExpr(); return; }
+    if (intr->coding > 0) {
+        CodeFuncCallOptionsBeginElmExpr();
+        return;
+    }
```
fingolfin: Hrm, another place I accidentally reformatted.

src/intrprtr.c
```diff
@@ -3972,8 +3866,6 @@ void             IntrAssertEnd3Args ( void )
 static Int InitKernel (
     StructInitInfo *    module )
 {
-    InitGlobalBag( &STATE(IntrState),  "src/intrprtr.c:IntrState"  );
-    InitGlobalBag( &STATE(StackObj),   "src/intrprtr.c:StackObj"   );
```
fingolfin: `STATE(IntrState)` and `STATE(StackObj)` are gone, so we can (and indeed, must) remove the above.

src/intrprtr.c
```diff
@@ -3989,16 +3881,6 @@ static Int InitKernel (
 }
 
 
-static Int InitModuleState(void)
-{
-    STATE(IntrCoding) = 0;
-    STATE(IntrIgnoring) = 0;
-    STATE(IntrReturning) = 0;
```
fingolfin: The caller is now responsible for passing a zeroed `struct IntrState` to `IntrBegin`.

src/intrprtr.c
```diff
@@ -4009,11 +3891,9 @@ static StructInitInfo module = {
     .type = MODULE_BUILTIN,
     .name = "intrprtr",
     .initKernel = InitKernel,
-
-    .initModuleState = InitModuleState,
```
fingolfin: Since `InitModuleState` is gone, we can and must remove this.

I believe I've explained all changes in `intrprtr.c` now in one way or another.

src/intrprtr.h
```diff
@@ -22,40 +22,34 @@
...
+    // directly interpret, such as loops or function bodies.
+    UInt coding;
+
+    // If 'IntrReturning' is non-zero, the interpreter is currently exiting
+    // statements enclosing a return statement. Actions from these statements
+    // are ignored.
+    ExecStatus returning;
+
+    // 'StackObj' is the stack of values.
+    Obj StackObj;
```
fingolfin: The above four variables were moved from `GAPState` into this new `struct IntrState`. I reformatted the comments suitable, and stripped irrelevant bits from them.

src/intrprtr.h
```diff
@@ -78,9 +72,9 @@
 **  return-void-statement was interpreted. If 'IntrEnd' returns 'STATUS_QUIT',
 **  then a quit-statement was interpreted.
 */
-void IntrBegin(Obj frame);
+void IntrBegin(IntrState * intr, Obj frame);
```
fingolfin: Most (all?) of the remaining changes are about adding `IntrState * intr` as first parameter to `IntrFoobar` functions.

src/read.c
```diff
@@ -170,31 +172,31 @@ static void ReadFuncCallOption(ReaderState * rs, TypSymbolSet follow)
     if (rs->s.Symbol == S_IDENT) {
         rnam = RNamName(rs->s.Value);
         Match(&rs->s, S_IDENT, "identifier", S_COMMA | follow);
-        TRY_IF_NO_ERROR { IntrFuncCallOptionsBeginElmName(rnam); }
+        TRY_IF_NO_ERROR { IntrFuncCallOptionsBeginElmName(&rs->intr, rnam); }
```
fingolfin: Most of the changes below are about adding `&rs->intr` as first argument to `IntrFoobar` calls.

src/read.c
```diff
@@ -265,55 +267,55 @@ GAP_STATIC_ASSERT(sizeof(LHSRef) <= 8, "LHSRef is too big");
 /****************************************************************************
 **
 */
-static UInt EvalRef(const LHSRef ref, Int needExpr)
+static UInt EvalRef(ReaderState * rs, const LHSRef ref, Int needExpr)
```
fingolfin: `EvalRef` now needs the `ReaderState * rs` so that it can extract the `IntrState` from it and pass that on.

src/read.c
```diff
@@ -324,49 +326,49 @@ static UInt EvalRef(const LHSRef ref, Int needExpr)
     return 0;
 }
 
-static void AssignRef(const LHSRef ref)
+static void AssignRef(ReaderState * rs, const LHSRef ref)
```
fingolfin: Same as `EvalRef`.

src/read.c
```diff
@@ -378,92 +380,92 @@ static void AssignRef(const LHSRef ref)
     }
 }
 
-static void UnbindRef(ScannerState * s, const LHSRef ref)
+static void UnbindRef(ReaderState * rs, const LHSRef ref)
```
fingolfin: Same as `EvalRef`; note that the `ScannerState` is contained in the `ReaderState`, so there is no need for us to pass both.

src/read.c
```diff
@@ -378,92 +380,92 @@ static void AssignRef(const LHSRef ref)
...
         case R_FUNCCALL_OPTS:
         default:
-            SyntaxError(s, "Illegal operand for 'Unbind'");
+            SyntaxError(&rs->s, "Illegal operand for 'Unbind'");
         }
     }
 }
 
-static void IsBoundRef(ScannerState * s, const LHSRef ref)
+static void IsBoundRef(ReaderState * rs, const LHSRef ref)
```
fingolfin: Same as `EvalRef`; note that the `ScannerState` is contained in the `ReaderState`, so there is no need for us to pass both.

src/read.c
```diff
@@ -713,7 +715,7 @@ static void CheckUnboundGlobal(ReaderState * rs, LHSRef ref)
         return;
 
     // don't warn if we are skipping/ignoring code
-    if (STATE(IntrIgnoring))
+    if (rs->intr.ignoring)
```
fingolfin: Replace `STATE(IntrIgnoring)`

src/read.c
```diff
@@ -803,31 +805,31 @@ static void ReadCallVarAss(ReaderState * rs, TypSymbolSet follow, Char mode)
...
             TRY_IF_NO_ERROR {
-                IntrFuncCallEnd(0, ref.type == R_FUNCCALL_OPTS, ref.narg);
+                IntrFuncCallEnd(&rs->intr, 0, ref.type == R_FUNCCALL_OPTS, ref.narg);
             }
         }
         else {
             Match(&rs->s, S_ASSIGN, ":=", follow);
             UInt currLHSGVar = rs->CurrLHSGVar;
-            if ( LEN_PLIST(rs->StackNams) == 0 || !STATE(IntrCoding) ) {
+            if ( LEN_PLIST(rs->StackNams) == 0 || !rs->intr.coding ) {
```
fingolfin: Replace `STATE(IntrCoding)`

src/intrprtr.c
ChrisJefferson: Maybe add an extra commit which does the formatting, just so it will be easy to spot if looking through history/blame?

ChrisJefferson:
Just as a sanity check, this doesn't effect the speed does it? I'm sure it doesn't (or if anything, would slightly speed things up, as it avoid accessing globals), but it might be worth just sanity checking.

src/intrprtr.c
fingolfin: It's out own macro, under our own full control; it does what we want, not the compiler or OS vendor. For now it is an alias for `assert`, but only if kernel debug mode was activated by configure

src/intrprtr.h
```diff
@@ -483,17 +478,17 @@ void IntrNot(void);
...
-void IntrLt(void);
+void IntrLt(IntrState * intr);
 
-void IntrGe(void);
+void IntrGe(IntrState * intr);
 
-void IntrGt(void);
+void IntrGt(IntrState * intr);
 
-void IntrLe(void);
```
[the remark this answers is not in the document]
ChrisJefferson: So, (void) means no arguments. In C++, both (void) and () mean the same thing.

In C, () doesn't mean no arguments! It means "unknown number and types of arguments". However, no-one should ever write this, and most people assume () means "no arguments". To see this, gcc will compile this without complaint:

```
void f() {}

int main(void)
{ f(2,3,4); }
```

src/intrprtr.h
fingolfin: There is no such type in GAP yet, that PR is still nit merged

src/intrprtr.c
ChrisJefferson: The problem in general reformatting is that it can make it slightly harder to see which bits of the code have changed (although in this case, it doesn't really make any difference, as you say)

fingolfin: I still think it gets slightly more annoying because the diff is bigger; and when a single line is just modified, you can see easily in the diff what was inserted/removed; while if I reformat it into multiple lines, you can't see that quite as easily; it might not even be easy to match old and new code. Anyway, it shouldn't be *that* bad; I'll put it into a second commit, so we can also drop it if we change our minds.

fingolfin:
@ChrisJefferson I measured `RereadGrp("clasmax.grp");` which takes about 30 milliseconds. Let it run 50 times through my benchmark function (which runs the GC etc. before taking timings), and could not see any significant difference in terms of performance. If you have a better idea for a "benchmark", let me know, I can run that then, too.

fingolfin:
I now also went to a Linux server (less fluctuation than on my laptop), and used perf tool to measure GAP's startup time (since the whole library has to be read, it's quite sensitive to changes in the reader and interpreter. I measures the head commits of the three stable branches for 4.9, 4.10 and 4.11, as well as this PR (the GASMAN fix from PR #3840 does not seem to have an effect). The file `quit.g` just contains `QUIT;`; no packages are loaded, no `gap.ini` or `.gaprc` are in effect.

## stable-4.9
```
 Performance counter stats for './gap -A -b ../quit.g' (10 runs):

            802.84 msec task-clock:u              #    0.999 CPUs utilized            ( +-  0.27% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
            34,966      page-faults:u             #    0.044 M/sec                    ( +-  0.00% )
     2,482,203,480      cycles:u                  #    3.092 GHz                      ( +-  0.28% )
     3,163,169,668      instructions:u            #    1.27  insn per cycle           ( +-  0.00% )
       693,667,281      branches:u                #  864.016 M/sec                    ( +-  0.01% )
        10,447,508      branch-misses:u           #    1.51% of all branches          ( +-  0.34% )

           0.80333 +- 0.00219 seconds time elapsed  ( +-  0.27% )
```


## stable-4.10
```
 Performance counter stats for './gap -A -b ../quit.g' (10 runs):

            833.74 msec task-clock:u              #    0.999 CPUs utilized            ( +-  0.22% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
            35,064      page-faults:u             #    0.042 M/sec                    ( +-  0.00% )
     2,583,316,151      cycles:u                  #    3.098 GHz                      ( +-  0.15% )
     3,565,799,484      instructions:u            #    1.38  insn per cycle           ( +-  0.00% )
       795,139,389      branches:u                #  953.702 M/sec                    ( +-  0.00% )
        10,690,165      branch-misses:u           #    1.34% of all branches          ( +-  0.27% )

           0.83419 +- 0.00181 seconds time elapsed  ( +-  0.22% )
```

## stable-4.11
```
 Performance counter stats for './gap -A -b ../quit.g' (10 runs):

            975.68 msec task-clock:u              #    1.000 CPUs utilized            ( +-  0.30% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
            37,200      page-faults:u             #    0.038 M/sec                    ( +-  0.00% )
     3,061,431,840      cycles:u                  #    3.138 GHz                      ( +-  0.22% )
     4,186,947,880      instructions:u            #    1.37  insn per cycle           ( +-  0.00% )
       914,773,888      branches:u                #  937.575 M/sec                    ( +-  0.00% )
        11,666,942      branch-misses:u           #    1.28% of all branches          ( +-  0.50% )

           0.97611 +- 0.00294 seconds time elapsed  ( +-  0.30% )
```

## This PR
```
 Performance counter stats for './gap -A -b ../quit.g' (10 runs):

            973.42 msec task-clock:u              #    1.000 CPUs utilized            ( +-  0.23% )
                 0      context-switches:u        #    0.000 K/sec
                 0      cpu-migrations:u          #    0.000 K/sec
            37,172      page-faults:u             #    0.038 M/sec                    ( +-  0.00% )
     3,047,728,129      cycles:u                  #    3.131 GHz                      ( +-  0.27% )
     4,213,028,737      instructions:u            #    1.38  insn per cycle           ( +-  0.03% )
       917,797,785      branches:u                #  942.859 M/sec                    ( +-  0.04% )
        11,572,378      branch-misses:u           #    1.26% of all branches          ( +-  0.73% )

           0.97388 +- 0.00219 seconds time elapsed  ( +-  0.23% )
```

What we can see is that this PR has no direct impact, but we *do* see a regression of the startup time, it got slower with each release :-(. One suspect for this I have in mind are th changes to decouple scanner and IO code; I'll investigate if those indeed are (part of) the cause. If they are, I already was planning to work on that for quite some time, it just wasn't high on my priority list.

fingolfin:
I used `git bisect` to track down the slowdown between GAP 4.10 and 4.11, and discovered what I should have known: this was caused by 1762043ea3bc02f8e65a91fadeca3c6f7dc38e23 from PR #2773, i.e., the method reordering.

ChrisJefferson:
Thanks, that's more than enough testing!

PR #3824, opened 2020-01-02
kernel: only compile workspace support if GASMAN is used

fingolfin:
We don't support workspaces for any other GC right now. This patch makes this explicit.

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1587,9 +1587,11 @@ void InitializeGap (
     InitGlobalBag(&POST_RESTORE, "gap.c: POST_RESTORE");
     InitFopyGVar( "POST_RESTORE", &POST_RESTORE);
 
+#ifdef GAP_ENABLE_SAVELOAD
+    if ( SyRestoring ) {
+
 #ifdef COUNT_BAGS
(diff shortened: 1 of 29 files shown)


src/gap.c
```diff
@@ -1608,37 +1610,33 @@ void InitializeGap (
...
           }
         }
-    }
 
+        return;
+    }
+#endif // GAP_ENABLE_SAVELOAD
 
     /* otherwise call library initialisation                               */
-    else {
```
fingolfin: Note that it is safe to remove this `else {` because we added a `return;` a few lines above; and that in turn is OK because `SyLoadSystemInitFile` is always zero when `SyRestoring` is non-zero; so the only thing we "loose" is the call to `ModulesCheckInit();`. While it would have been trivial enough to add this before the removed `return`, I chose not to, because it just performs some sanity checks, which are pointless to run *after* we finished execution (i.e., after `POST_RESTORE` returns). Besides, those checks are mostly meant to catch errors in constant internal tables, so any errors it detects would be caught by us (as they'd show up on every launch) and thus not included in a released GAP version.

PR #3826, opened 2020-01-03
kernel: merge SyTmpname, SyTmpdir into callers

fingolfin:
The "portability abstraction" provided by sysfiles.c doesn't help
us these days, it just makes it harder to reason about this code.

This is an hopefully uncontroversial part of PR #3702, which has stalled; I hope this bit can be merged now, so that PR #3702 can be reduced to its essential / controversial bits, to facilitate a resolution there.

UPDATE: while the first commit just moves code, I just added a second commit which also fixes an obvious issue when using these functions in multi threaded code (I replaced two static buffers). This was a trivial fix thanks to the refactoring in the first commit.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -1036,14 +1036,10 @@ static Obj FuncREAD_GAP_ROOT(Obj self, Obj filename)
 */
 static Obj FuncTmpName(Obj self)
 {
-    Char *              tmp;
-    Obj                 name;
-
-    tmp = SyTmpname();
-    if ( tmp == 0 )
-        return Fail;
-    name = MakeString(tmp);
-    return name;
+    static char name[1024];
+    strxcat(name, "/tmp/gaptempfile.XXXXXX", sizeof(name));
+    close(mkstemp(name));
+    return MakeString(name);
 }
 
 
@@ -1053,14 +1049,25 @@ static Obj FuncTmpName(Obj self)
 */
 static Obj FuncTmpDirectory(Obj self)
 {
-    Char *              tmp;
-    Obj                 name;
+    static char name[1024];
+    char *      env_tmpdir = getenv("TMPDIR");
+    if (env_tmpdir != NULL) {
+        strxcpy(name, env_tmpdir, sizeof(name));
+        strxcat(name, "/", sizeof(name));
+    }
+    else {
+#ifdef SYS_IS_CYGWIN32
+        strxcpy(name, "/cygdrive/c/WINDOWS/Temp/", sizeof(name));
+#else
+        strxcpy(name, "/tmp/", sizeof(name));
+#endif
+    }
+    strxcat(name, "gaptempdirXXXXXX", sizeof(name));
 
-    tmp = SyTmpdir("tm");
-    if ( tmp == 0 )
+    char * tmp = mkdtemp(name);
+    if (tmp == 0)
         return Fail;
-    name = MakeString(tmp);
-    return name;
+    return MakeString(tmp);
 }
 
 
--- a/src/streams.c
+++ b/src/streams.c
@@ -1036,8 +1036,7 @@ static Obj FuncREAD_GAP_ROOT(Obj self, Obj filename)
 */
 static Obj FuncTmpName(Obj self)
 {
-    static char name[1024];
-    strxcat(name, "/tmp/gaptempfile.XXXXXX", sizeof(name));
+    char name[] = "/tmp/gaptempfile.XXXXXX";
     close(mkstemp(name));
     return MakeString(name);
 }
@@ -1049,8 +1048,8 @@ static Obj FuncTmpName(Obj self)
 */
 static Obj FuncTmpDirectory(Obj self)
 {
-    static char name[1024];
-    char *      env_tmpdir = getenv("TMPDIR");
+    char   name[GAP_PATH_MAX];
+    char * env_tmpdir = getenv("TMPDIR");
     if (env_tmpdir != NULL) {
         strxcpy(name, env_tmpdir, sizeof(name));
         strxcat(name, "/", sizeof(name));
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -3217,60 +3217,6 @@ Obj SyIsDir ( const Char * name )
 *F * * * * * * * * * * * * * * * directories  * * * * * * * * * * * * * * * *
 */
 
-
-/****************************************************************************
-**
-*F  SyTmpname() . . . . . . . . . . . . . . . . . return a temporary filename
-**
-**  'SyTmpname' creates and returns  a new temporary name.  Subsequent  calls
-**  to 'SyTmpname'  should  produce different  file names  *even* if no files
-**  were created.
-*/
-Char *SyTmpname ( void )
-{
-  static char name[1024];
-  strxcpy(name, "/tmp/gaptempfile.XXXXXX", sizeof(name));
-  close(mkstemp(name));
-  return name;
-}
-
-
-/****************************************************************************
-**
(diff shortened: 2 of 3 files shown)


ChrisJefferson:
While I'm sure we are breaking this all over the place :) I thought part of the point of sysfiles. Cwas to keep OS specific code (like Windows) in one place. Just to officially check, are we going to stop doing that? Or do we have a different rule for what should go in sysfiles?

fingolfin:
@ChrisJefferson first off, I think that's a very good and fair question you are asking. I'll reply extensively below; I realize this turned a bit into a rant, but that rant is not directed against you or anybody :-)

Perhaps a long long time ago, all system specific code really was in `sys*.c` files. And back then, there certainly was a *lot* of that, and many functions had two more variants, as there was support for DOS, OS/2, Atari TOS, MacOS (classic), various Unix variants, etc. But almost all of that is gone. 

The next question then is: What exactly is "OS specific code" ? The border seems extremely blurry.

Finally, the IMHO most important question is: What's the point of the "rule" here (whatever the "rule" might be, exactly)?

In the end, *we* make the rules, and we make them to be useful; if a rule is not useful, let's not follow it for its own sake, but rather revise or drop it.

Here, in the specific case of this PR, ignoring that "rule" shrinks the code from 92 to 21 lines while retaining identical functionality. The resulting code then becomes much easier to understand and hence easier to fix and otherwise modify.

Granted, a lot of the saved "code" is lengthy comments -- comments that try to codify an internal "sys API", but in the end are insufficient for that. In the revised code, we just call `mkstemp` and `mkdtemp`, and *those* functions have extensive specifications in POSIX, so it's clear what they are doing. Indeed, think about it: in order to implement a non-POSIX alternative of, say, `SyRmdir`, I need to know the exact semantics of `SyRmdir` (in particular: how to deal with errors). But in truth there is just a single function calling it, namely `FuncRemoveDir`. In truth, what we care about are the semantics of `FuncRemoveDir`. Looking at it, it is completely unclear to me what we gain from keeping `FuncRemoveDir` and `SyRmdir` separate? Anybody have an idea? Note that there never was a second implementation of `SyRmdir`.

Now, one day, I hope we'll move POSIX. Be it that we get a proper Windows port (not using Cygwin), or that the world just comes up with a better OS with saner APIs. Then we may need to rework a lot of that code. Maybe it'll even make sense to re-introduce some intermediate wrappers to be able to write code that works on POSIX and that new OS API as well. But IMHO we can only decide how that new intermediate wrapper should look like once we know the target.

Let's look at our sys*.c files and what they do:
- `system.c`: contains `Panic_` (to be moved to `error.c`?) and `SyExit`; and otherwise basically only stuff related to command line parsing -> perhaps the file should be rename to `sysopt.c` soon?
- `system.h`: not much remains here; most either should be moved into a future `gap_config.h` file (`SYS_IS_64_BIT`, `GAP_PATH_MAX`, ...) or to `error.h` (`Panic`?), or to `sysopt.h` (`InitSystem`, after renaming to `ParseArguments`/`ParseOptions`/...?). Only thing I am not sure where else to put is `SyExit`; a year ago I'd have voted to remove it, but now it's useful for the Julia integration (it calls `jl_atexit_hook` there). Perhaps rename it to `GapExit` and move to `gap.h` / `gap.c`?
- `sysmem.{c,h}`:  contains a few command line settings variables (which perhaps should be moved back to system.c/sysopt.c?), and otherwise just gasman specific "system specific" bits (with one annoying exception: `SyMsgsBags` resp. `SyGasmanNumbers` is also used by the Julia GC, so that our test suites produce slightly nicer output; I'd like to replace this with a nicer API that's not so closely tied to GASMAN). Anyway, this file should perhaps be renamed to `gasman_intern.c` ?
- `sysroots.{c,h}`: contains virtually nothing system specific; possible better name: `gaproots.{c,h}
- `sysstr.{c,h}`: These files contain basically no system dependant code (unless you count ISO C functions among those?), but it ...
  - contains our own wrappers for ctype.h (remove those, use isalpha/isdigit directly? and move `IsIdent` back to e.g. `scanner.h`?);
  - and also `strlcpy`, `strlcat` (these are BSD system APIs; I now wonder if we should just rename them to `gap_strlcpy` / `gap_strlcat` to avoid complications that arise because I don't notice on macOS (which has native `strlcpy`, `strlcat`) when those headers are by accident not included); 
  - and also `strxcpy` / `strxcat` -- which were a great quick&dirty way to replace a lot of unsafe uses of strcpy, strcat, strncpy, strncat, but now I'd prefer to just get rid of those completely...
- `systime.{c,h}`: I am actually happy with those :-)
- `sysjmp.{c,h}`: also OK, though really; `setjmp` are certainly system dependent; though one could argue that this is also about the `RegisterSyLongjmpObserver` code, which is an abstraction built *atop* this internal API, so strictly speaking should be elsewhere; indeed, on the long run, the better thing to do would be to move this all in our own "exception_handler.{c,h}" or so... or maybe convert the whole code to C++ and use C++ exceptions?!? anyway, for now, it can stay where it is
- `sysenv.h` is about `environ`, which is part of POSIX -- but sadly a very shady part, where it's unclear to me why they didn't come up with a successor solution 20 years ago (of course they would have had to keep `environ` around for now, but the hope is that after a few decades it could be deprecated and eventually removed; as it is, this migration has not yet been started, or even planned *sigh*). Anyhow, this file rightfully has `sys` in its name :-)
- `sysfiles.{c,h}`: this should probably be split in 2 or 3... parts of it are indeed dealing with system dependant functions, but a lot of it also is not...
  - I already moved the bits dealing with GAP roots out of it recently.
  - a big chunk in there is about the `SyBuf*` stuff. This is not so much about being system independent (that code uses ISO C resp. POSIX C APIs through and through), but simply a wrapper that implement things like our pseudo file names `*stdout*`, support for "window handlers" . I think this part of the code should perhaps be moved to a `sysbuf.c` or 
  - then there is the GNU readline "interface", which is also not really fit for a `sys*` file, but it sits there because our use of readline essentially is a big hack where we replace `fgets` by a function which calls readline (at least under certain conditions). This, too, should go elsewhere (this is part of my motivation to refactor the whole scanner/reader/interpreter/... code)
  - finally there are wrappers for "file system" interaction, like `SyIsReadableFile`, `SyIsDir` which arguably are indeed suitable for a `sys*.c`. 
   - there is also `SyExecuteProcess` in there...
- `iostream.c` also contains a lot of system dependant stuff, for launching and controlling subprocesses. In a "pure" model, parts of it probably should be moved into a new `sys*.c` file (and/or into `sysfiles.c`). But I am not sure what we'd gain by this. I don't think it'd make the code easier to understand, maintain, or extend (for new, non-POSIX APIs). Well, *maybe* the last bit (support for new APIs) would be handy, but I don't think this can be done sensibly before actually knowing that hypothetical POSIX alternative...
- `streams.c` also contains "system dependant" stuff. It also interfaces closely with (parts of) `sysfiles.c`. I think this should perhaps be split into three files:
   - leave in it `READ` and friends
   - move the functions dealing with log files to a new `log.c`
   - move the functions dealing with files (`FuncIsExistingFile`) into a new file (`files.c` ?) and then consider merging them with their `sysfiles.c` counterparts (this PR here is doing just this for two of those functions).

fingolfin:
Let me try a short summary of my thoughts: I am all for carefully restricting the places where we call APIs beyond ISO C, including POSIX stuff not in ISO C, such as `fork`, etc. This makes future porting efforts easier and is a good idea in general. However, I don't think the current `sys*.c` approach does that all too well. In many cases, I think the level at which we encapsulate is not chosen well. E.g. it requires us to work with fixed size buffers and manually byte shuffling, only for the calling function to then copy all that data into GAP lists and strings. So, let's move away from dogma and towards a more pragmatic encapsulation.

Issue #2566, opened 2018-06-20
HPC-GAP: add proper support for module state / `InitModuleState` in kernel extensions


fingolfin:
In HPC-GAP, GAP kernel modules may have thread local state, which is initialized via the module function `InitModuleState`.

Kernel extensions may also make use of this -- although so far, I think none do, but this could change in the future. If it does change, we ought to resolve this issue: Right now, when loading a kernel extension, we call its `InitModuleState` method only once, on the currently active thread. But we ought to call it on all threads resp. somehow ensure that on each thread, it is called before any other functions of that kernel extension are used on that thread.

fingolfin:
On a tangential point, we may also need to add a second per-thread callback to `StructInitInfo`: right now, we first call `InitKernel` then `InitLibrary`, then `InitModuleState` -- all from the main thread. In the future, we might call `InitModuleState` from each thread.

But I think we also need another callback run from each thread which is invoked after `InitKernel` but before `InitLibrary`, which is used to add GC roots.

PR #3828, opened 2020-01-06
kernel tweaks: fix name of IO kernel module; add/move/update some comments; remove dead declarations; etc.

fingolfin:


src/io.c
```diff
@@ -2174,7 +2174,7 @@ static StructInitInfo module = {
     // init struct using C99 designated initializers; for a full list of
     // fields, please refer to the definition of StructInitInfo
     .type = MODULE_BUILTIN,
-    .name = "scanner",
+    .name = "io",
```
DominikBernhardt: Just so I understand this better:
The documentation for `StructInitInfo` under `.name` says: 

> name of the module: filename with ".c" or library filename    

There is both a `scanner.c` and `io.c`, so that's fine. Clearly we want to refer to `io` here, I understand that. But why doesn't this trigger problems if the name corresponds to the wrong `.c` file?

src/code.c
```diff
@@ -3230,10 +3230,8 @@ static Int InitKernel (
     MakeBagTypePublic(T_BODY);
 #endif
 
-#if !defined(HPCGAP)
```
DominikBernhardt: Here and in the following three files: Why can we drop `#if !defined(HPCGAP)`? If this is obvious, please ignore my question, I don't understand much of HPCGAP. Ping @wilfwilson to have a look at this.

src/code.c
```diff
@@ -1809,6 +1809,11 @@ void CodeStringExpr (
     PushExpr( string );
 }
 
+
+/****************************************************************************
+**
+*F  CodePragma(<pragme>)
```
DominikBernhardt: Maybe 
```suggestion
*F  CodePragma(<pragma>)
```
?

src/code.h
```diff
@@ -1100,6 +1110,11 @@ void CodeListExprEnd(UInt nr, UInt range, UInt top, UInt tilde);
 */
 void CodeStringExpr(Obj str);
 
+
+/****************************************************************************
+**
+*F  CodePragma(<pragme>)
```
DominikBernhardt:
```suggestion
*F  CodePragma(<pragma>)
```
?

src/io.c
fingolfin: Because we don't have code that verifies that no module name is used twice.

But there are functions which lookup modules by name, and if you used those with the name `"scanner"`, they would return whichever module comes first in the list of modules; and that actually is `io`, not `scanner`. Luckily, this doesn't matter at all, because those functions are not actually called that way (and they might never be). So this had no real effect, for now. But it might one day...

For now, the only "visible" effect is when using debugging tools, like the `-D` option, which currently gives this output during startup:
```$ ./gap -D
#I  InitInfo(builtin gvars)
#I  InitInfo(builtin objects)
#I  InitInfo(builtin profile)
#I  InitInfo(builtin hookintrprtr)
#I  InitInfo(builtin tracing)
#I  InitInfo(builtin scanner)
#I  InitInfo(builtin scanner)
#I  InitInfo(builtin read)
#I  InitInfo(builtin calls)
...
```
Note how `scanner` is listed twice.

This is also how I noticed this.

src/code.c
fingolfin: Indeed

src/io.c
DominikBernhardt: Thank you!

src/code.c
fingolfin: `InitGlobalBag` simply does nothing in HPC-GAP; or more specifically: with the Boehm garbage collector. My guess is that at some stage of the HPC-GAP development, there was a move to disable all `InitGlobalBag` calls for HPC-GAP. But as you can see already a few lines below this block, that never went far. (@rbehrends or @stevelinton may know the background better).

In any case, when we merged HPC-GAP back into GAP, instead of removing those calls, we put them into `#ifdef HPCGAP`. But this serves no purpose in the end, it just makes the code confusing ("huh, why is this `InitGlobalBag` disabled for HPCGAP? And why isn't the next?")

rbehrends: Yes, this should be removed, I think, especially as we want it to work with the Julia GC, which uses the same roots as GASMAN. I think it may actually be a good idea to have the Boehm GC at some time declare its global roots explicitly (it normally defaults to scanning the entire data segment, which is both inefficient and may hide bugs). That's not a pressing issue, though, just something to keep in mind.

fingolfin: Note that for HPC-GAP, these calls could not stay in `InitKernel`, as `InitKernel` is only called once, from the main thread. So we'd need something like a "per-thread-InitKernel". See also #2566

PR #3830, opened 2020-01-07
kernel: move Start/FinishAndCallFakeFuncExpr to intrprtr.c

fingolfin:
This removes the only call to the coder from read.c. It also makes sense since
`StartFakeFuncExpr` and `FinishAndCallFakeFuncExpr` are only called from
functions in `intrprtr.c` anyway, so we can now make them `static`.

This was/is part of PR #3819; I am putting it into its own PR in an attempt to simplify #3819 as much as possible, to increase my chance of finding a reviewer for that.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1185,60 +1185,6 @@ static ArgList ReadFuncArgList(ScannerState * s,
 }
 
 
-void StartFakeFuncExpr(Int startLine)
-{
-    assert(STATE(IntrCoding) == 0);
-
-    // switch to coding mode now
-    CodeBegin();
-
-    // code a function expression (with no arguments and locals)
-    Obj nams = NEW_PLIST(T_PLIST, 0);
-
-    // If we are in the break loop, then a local variable context may well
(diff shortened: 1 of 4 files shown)


src/read.c
```diff
@@ -1185,60 +1185,6 @@ static ArgList ReadFuncArgList(ScannerState * s,
...
-    // Without this, access to variables defined in the existing local
-    // variable context will be coded as LVAR accesses; but when we then
-    // execute this code, they will not actually be available in the current
-    // context, but rather one level up, i.e., they really should have been
-    // coded as HVARs.
-    //
-    // If we are not in a break loop, then this would be a waste of time and
-    // effort
-    if (LEN_PLIST(ReaderState()->StackNams) > 0) {
-        PushPlist(ReaderState()->StackNams, nams);
```
fingolfin: The only change made to the two functions when copying them was to replace `ReaderState()->StackNams` by an explicit argument to `StartFakeFuncExpr` resp. `FinishAndCallFakeFuncExpr`

PR #3833, opened 2020-01-07
Request required storage in one allocation in GASMAN

ChrisJefferson:
Previously, when we failed to allocate enough memory we would
still extend the GASMAN memory space to the maximum possible size,
which would lead to massive memory usage until the next GC.

To see this, in the current master branch run the following code. The memory usage of GAP will grow massively. With this PR, the memory used by GASMAN does not grow after the failed allocation.

This PR is designed to not change things significantly, in particular it still allocates in a multiple of 512K, just because this is what has always been done.

```
gap> l := [];;
gap> l[2^50] := 1;
Error, Cannot allocate 9007199254741000 bytes: cannot extend the workspace any more!!!!
not in any function at *stdin*:2
brk> quit;
gap> while true do List([1..10000], x -> x^2); od;
```

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -2173,9 +2173,17 @@ static Int CollectBags_Check(UInt size, UInt FullBags, UInt nrBags)
 
 
         /* get the storage we absolutely need                              */
-        while ( EndBags < stopBags
-             && SyAllocBags(512,0) )
-            EndBags += WORDS_BAG(512*1024L);
+        if (EndBags < stopBags) {
+            size_t bytes = (char *)stopBags - (char *)EndBags;
+            // Increment in blocks of 512K
+            size_t blocks = bytes / 1024 / 512;
+            if (blocks * 1024 * 512 < bytes) {
+                blocks++;
+            }
+            if (SyAllocBags(blocks * 512, 0)) {
+                EndBags += WORDS_BAG(blocks * 512 * 1024);
+            }
+        }
 
         /* if not enough storage is free, fail                             */
         if ( EndBags < stopBags )


src/gasman.c
```diff
@@ -2173,9 +2173,14 @@ static Int CollectBags_Check(UInt size, UInt FullBags, UInt nrBags)
 
 
         /* get the storage we absolutely need                              */
-        while ( EndBags < stopBags
-             && SyAllocBags(512,0) )
-            EndBags += WORDS_BAG(512*1024L);
+        if (EndBags < stopBags) {
+            size_t bytes = (char *)stopBags - (char *)EndBags;
+            // Increment in blocks of 512K, rounded up.
+            size_t blocks = bytes / 1024 / 512 + 1;
```
wilfwilson: What’s the thinking with the rounding up? Perhaps you shouldn’t be adding the 1 if `bytes` divides exactly into 512K?

ChrisJefferson: The rounding up is to emulate the previous behaviour, which is we always allocate and deallocate in blocks of 512K. The dividing exactly bit is actually what we used to do, and also it's not really a problem the allocate a little more capacity.

However, I do admit it looks a bit strange, so I should maybe at least say that I'm doing this just because we always used to.

wilfwilson: I’m looking on my phone in a train station, so hard to think properly, but it looks to me that if `stopBags = EndBags + (512 * 1024)` exactly, then the old code allocated 512K extra memory, whereas the new code allocated twice as much. I am right, or have I got got it wrong? (Very possible.)

ChrisJefferson: You are right. Allocating extra memory isn't a problem, allocating less would be (as we need at least enough memory to satisfy the request). This kind of thing is annoying, so I think I'll write a better comment.

wilfwilson: Cool.  Yes, I understand that 512K extra memory is not a problem these days.  I just had the assumption that you wanted to keep the behaviour _exactly_ as before (with the exception that you don’t allocate any new memory if it’s going to fail). But if you’re happy for that not to be the case, then no worries.

wilfwilson: (To get the behave exactly as before, you’d only add the `+1` when `bytes` is _not_ divisible exactly by `512 * 1024`.)

ChrisJefferson: (hopefully) improved

src/gasman.c
```diff
@@ -2173,9 +2173,17 @@ static Int CollectBags_Check(UInt size, UInt FullBags, UInt nrBags)
...
-            EndBags += WORDS_BAG(512*1024L);
+        if (EndBags < stopBags) {
+            size_t bytes = (char *)stopBags - (char *)EndBags;
+            // Increment in blocks of 512K
+            size_t blocks = bytes / 1024 / 512;
+            if (blocks * 1024 * 512 < bytes) {
+                blocks++;
+            }
+            if (SyAllocBags(blocks * 512, 0)) {
+                EndBags += WORDS_BAG(blocks * 512 * 1024L);
```
fingolfin:
```suggestion
                EndBags += WORDS_BAG(blocks * 512 * 1024);
```
There is no point in using `1024L` (and we just removed a bunch of stuff)

PR #3834, opened 2020-01-07
kernel: fix `for` loop regression

### LINKED ISSUE #3831: Regression: `for` loop with undeclared index variable doesn't give a warning anymore
The following used to happen:
```
gap> f := function() for unknownvarname in [1..10] do od; end;
f := function() for unknownvarname in [1..10] do od; end;
                                    ^
function(  ) ... end
```
But with current master, there is no warning anymore. This regression was introduced by PR #3808.

I will look into a fix later today/tonight, but that shouldn't stop others from looking into it.

fingolfin:
... namely print warnings if the index variable `i` in a statement
of the form `for i in list do` is an undefined global.

To do this, move the code performing this check into a new helper function `CheckUnboundGlobal`, and also use this opportunity to refactor that check and add more comments to it.

Fixes #3831

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -693,6 +693,55 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
     return ref;
 }
 
+// Helper function to be called after `ReadVar`, before any further tokens
+// have been consumed via calls to `Match`.
+static void CheckUnboundGlobal(ScannerState * s, LHSRef ref)
+{
+    // only warn if we are accessing a global variable
+    if (ref.type != R_GVAR)
+        return;
+
+    // only warn if inside a function
+    if (LEN_PLIST(ReaderState()->StackNams) == 0)
+        return;
+
+    // allow use in left hand side (LHS) of an assignment
+    if (ref.var == ReaderState()->CurrLHSGVar)
+        return;
+
+    // only warn if the global variable does not exist ...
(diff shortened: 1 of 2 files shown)


src/read.c
```diff
@@ -693,6 +693,55 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
...
+{
+    // only warn if we are accessing a global variable
+    if (ref.type != R_GVAR)
+        return;
+
+    // only warn if inside a function
+    if (LEN_PLIST(ReaderState()->StackNams) == 0)
+        return;
+
+    // allow use in left hand side (LHS) of of assignment
```
wilfwilson:
```suggestion
    // allow use in left hand side (LHS) of an assignment
```

src/read.c
```diff
@@ -693,6 +693,55 @@ static LHSRef ReadVar(ScannerState * s, TypSymbolSet follow)
...
+
+    // ... and isn't an auto var
+    if (ExprGVar(ref.var) != 0)
+        return;
+
+    // don't warn if we are skipping/ignoring code
+    if (STATE(IntrIgnoring))
+        return;
+
+    // of the global was used as loop variable in an enclosing for loop, that
```
wilfwilson:
```suggestion
    // if the global was used as loop variable in an enclosing for loop, that
```

PR #3838, opened 2020-01-08
FIX: Made ConfluentMonoidPresentationForGroup an attribute

hulpke:
to ensure that always the same presentation is used in subsequent cohomology
calculations.

### DIFF
--- a/lib/twocohom.gd
+++ b/lib/twocohom.gd
@@ -323,3 +323,26 @@ DeclareGlobalFunction("FpGroupCocycle");
 ##
 DeclareGlobalFunction("CompatiblePairOrbitRepsGeneric");
 
+
+#############################################################################
+##
+#A  ConfluentMonoidPresentationForGroup( <G> )
+##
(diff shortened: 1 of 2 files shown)


wilfwilson:
Do you want the new documentation to appear in the manual? If so, then I believe you'll have to include the line
```
<#Include Label="ConfluentMonoidPresentationForGroup">
```
somewhere, perhaps somewhere like here
https://github.com/gap-system/gap/blob/a4a37643cac16d117271198711e958b872950418/doc/ref/grppc.xml#L466
in `doc/ref/grppc.xml`.

hulpke:
@wilfwilson 
I hadn't planned for it to go in the manual yet, as the interface might change and since there is not enough documentation/examples to make it easily usable by others, but I was trying to be nice to not just add a `DeclareAttribute` but give a bit of explanation what it is for.

wilfwilson:
@wilfwilson Thanks for being nice!

PR #3839, opened 2020-01-09
kernel: move InterpreterStartLine from GAPState to `struct IntrState`

fingolfin:
This further decouples the scanner/reader/interpreter from the IO code.

<s>This is based on PR #3823, which is the first commit of this PR. Thus if anybody wants to review this PR already now, you may wish to focus on the second commit.

I am posting it already now because I want to see how my change affects code coverage.</s>

PR #3823 has been merged, so this can be reviewed.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -130,10 +130,26 @@ static UInt GlobalComesFromEnclosingForLoop(ReaderState * rs, UInt var)
     return 0;
 }
 
-// match either a semicolon or a dual semicolon
-static void MatchSemicolon(ScannerState * s, TypSymbolSet skipto)
+// `Match_` is a thin wrapper around the scanner's Match() function, in which
+// we can track the start line of each interpreter "instruction". This
+// information is then used for profiling.
+static void Match_(ReaderState * rs,
+           UInt           symbol,
+           const Char *   msg,
+           TypSymbolSet   skipto)
 {
-    Match(s, s->Symbol == S_DUALSEMICOLON ? S_DUALSEMICOLON : S_SEMICOLON,
+    if (rs->intr.startLine == 0 && symbol != S_ILLEGAL) {
+        rs->intr.startLine = rs->s.SymbolStartLine[0];
+    }
+
+    Match(&rs->s, symbol, msg, skipto);
+}
+
+
+// match either a semicolon or a dual semicolon
+static void MatchSemicolon(ReaderState * rs, TypSymbolSet skipto)
+{
+    Match_(rs, rs->s.Symbol == S_DUALSEMICOLON ? S_DUALSEMICOLON : S_SEMICOLON,
           ";", skipto);
 }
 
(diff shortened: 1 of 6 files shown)


src/read.c
```diff
@@ -128,10 +130,23 @@ static UInt GlobalComesFromEnclosingForLoop(ReaderState * rs, UInt var)
     return 0;
 }
 
+static void Match_(ReaderState * rs,
```
fingolfin: If somebody has a better idea for a good name for this static wrapper function, please let me know

fingolfin: TODO: should add a documentation comment for this function

wilfwilson:
@fingolfin Is this ready to be reviewed? The PR description, the TODO comment, and the failing CI might suggest that it’s possibly a WIP.

fingolfin:
@wilfwilson this has been ready for quite some time. The TODO comment was already addressed, I've now made that clearer. I've also rebased this, to see if test failures persist.

The main "issue" (?!?) with this PR is that it affects code coverage for interpreted code. But it's IMHO not quite clear cut whether the change is a bug/regression or a feature/improvement. This then lead to me submitting issue #3841, where I was (and am) hoping to come to some kind of agreement on what correct coverage for interpreted code *should* look like. Then we can use that to decide whether the change in this PR is beneficial (gets us closer to the agreement), or not (of course it could also be the one in some regards and the other in others).

For this, it would also be nice if @ChrisJefferson had a look here resp. at #3841, where I made some *concrete* suggestions as to how I think code coverage should work (that's meant as a baseline for the conversation; I am quite flexible in changing my views on this :-).

wilfwilson:
I’ve had a look through and the code seems reasonable, and it seems to do what you say.

I get that this PR affects code coverage for interpreted code, and that (in #3841) we should decide our specification for what we want to be considered as code for the code coverage.

The relevant question I have here is: do the efforts in this PR restrict what specification we will be able to implement?

If the moving of `InterpreterStartLine` from `GAPState` to `struct IntrState` is basically independent of our ability to be able to call things “code” or not, then I think we should just get on and merge this, and decide and implement a consistent coverage specification separately. (Even if the coverage changes in this PR are contrary to what we will ultimately want.)

fingolfin:
@wilfwilson I believe this PR does not fundamentally affect our ability to make changes to the profiling of the interpreted code. Actually I think that in order to make those changes, we may have to radically change how exactly we track the profiling information; right now we "cheat" by hooking into `Match` / `Match_`, but I suspect this won't actually do it; we may need a different approach. But if that's true, then it doesn't matter whether this PR here is merged or not.

fingolfin:
@ChrisJefferson you are the canonical person to review this, and also to comment on the coverage changes this introduces  (those prompted me to file issue #3841). I realize you are on strike now and so perhaps don't want to review this right now, but if at some later point in the future you could, that'd be great!

ChrisJefferson:
Could you give this branch a rebase, just so I can compare the coverage with current master?

fingolfin:
@ChrisJefferson rebased

ChrisJefferson:
I think this helps clean up code. Let's merge it.

PR #3840, opened 2020-01-09
Move MarkBag* funcs into GCs so they can inline MarkBag

ChrisJefferson:
While tracking down a slowness, I found that splitting functions involved in marking bags between different files created a significant slowdown in gasman. This patch pulls these functions into a header which should be included into each GC. I'm not totally happy with how this is done, but it feels like the best way to do things without duplicating code.

The ".inc" extension is something I've used in other systems for headers which must only be included in one file. I'm open to alternative suggestions. Another (horrible) option would be to put this stuff into gasman.h, guarded by a macro (IN_GC?), which would be #defined in each GC file...

On master:

```
gap> l := List([1..10000], x -> ListWithIdenticalEntries(10000, 0));; # make some stuff
gap> GASMAN("collect");
gap> time;
649
gap> GASMAN("collect");
gap> time;
501
gap> GASMAN("collect");
gap> time;
453
```

With this patch:

```
gap> l := List([1..10000], x -> ListWithIdenticalEntries(10000, 0));; # make some stuff
gap> GASMAN("collect");
gap> time;
396
gap> GASMAN("collect");
gap> time;
273
gap> GASMAN("collect");
gap> time;
295
```


ChrisJefferson:
I am aware this needs some cleaning up, but I'm interested on opinions on the general layout (in particular having the Inc file at all)

fingolfin:
Wow, that's massive. And my fault, I guess (never expect inlining to be so vital in this case, but I should have measured back then *sigh*).

I am not "hot" for those `.inc` files, but I don't really see a better alternative. Of course we could also just return to the old way and duplicate the code to each of the three GC implementations. Honestly, both of these options seem equally appealing to me.

src/bags_funcs.inc
```diff
@@ -0,0 +1,62 @@
...
+**
+**  This file is part of GAP, a system for computational discrete algebra.
+**
+**  Copyright of GAP belongs to its developers, whose names are too numerous
+**  to list here. Please refer to the COPYRIGHT file for details.
+**
+**  SPDX-License-Identifier: GPL-2.0-or-later
+**
+** This file includes functions which are required by all GCs. For efficiency
+** it is important these are defined in the same function as "MarkBag", so
```
fingolfin:
```suggestion
** it is important these are defined in the same file as "MarkBag", so
```

src/bags_funcs.inc
```diff
@@ -0,0 +1,62 @@
...
+** this file should be included in each GC.
```
fingolfin:
```suggestion
** this file should be included in each GC implementation.
```

src/julia_gc.c
```diff
@@ -27,6 +27,8 @@
 #include "sysmem.h"
 #include "vars.h"
 
+#include "bags_funcs.inc"
```
fingolfin: Shouldn't `boehm_gc.c` also get this?

fingolfin: Since HPC-GAP failed on Travis, I guess the answer is "yes" ;-)

fingolfin:
I just checked, and luckily, this regression is not present in GAP 4.10, but of course it *is* present in the (as of yet unreleased) GAP 4.11. As such, I'd like to backport this to GAP 4.11, if possible (CC @alex-konovalov).

fingolfin:
Backported in 8b3ec23e534e1d98d1d471047aa8689487f3f2de

PR #3844, opened 2020-01-10
kernel: avoid Intr{Begin,End} calls, remove Exec{Begin,End}

fingolfin:
Also remove FuncsState()->ExecState. We can instead store the old lvars on the
stack resp. in the interpreter state.

### DIFF
--- a/src/hpc/threadapi.c
+++ b/src/hpc/threadapi.c
@@ -19,7 +19,6 @@
 #include "funcs.h"
 #include "gapstate.h"
 #include "gvars.h"
-#include "intrprtr.h"
 #include "io.h"
 #include "lists.h"
 #include "modules.h"
@@ -31,6 +30,7 @@
 #include "set.h"
 #include "stats.h"
 #include "stringobj.h"
+#include "vars.h"
 
 #include "hpc/guards.h"
 #include "hpc/misc.h"
@@ -432,15 +432,15 @@ static GVarDescriptor GVarTHREAD_EXIT;
 
 static void ThreadedInterpreter(void * funcargs)
 {
-    IntrState intr = { 0, 0, 0, 0 };
-    Obj tmp, func;
+    Obj tmp, func, oldLVars;
     int i;
 
-    /* initialize everything and begin an interpreter                       */
+    // initialize everything and begin a fresh execution context
     STATE(NrError) = 0;
     STATE(ThrownObject) = 0;
+    oldLVars = STATE(CurrLVars);
+    SWITCH_TO_OLD_LVARS(STATE(BottomLVars));
 
-    IntrBegin(&intr, STATE(BottomLVars));
     tmp = KEPTALIVE(funcargs);
     StopKeepAlive(funcargs);
     func = ELM_PLIST(tmp, 1);
@@ -462,15 +462,12 @@ static void ThreadedInterpreter(void * funcargs)
         exit = GVarOptFunction(&GVarTHREAD_EXIT);
         if (exit)
             CALL_0ARGS(exit);
-        PushVoidObj(&intr);
-        /* end the interpreter */
-        IntrEnd(&intr, 0, NULL);
     }
     CATCH_ERROR
     {
-        IntrEnd(&intr, 1, NULL);
         ClearError();
     }
+    SWITCH_TO_OLD_LVARS(oldLVars);
 }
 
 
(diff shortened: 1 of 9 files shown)


src/read.c
```diff
@@ -2718,112 +2718,79 @@ void ReadEvalError(void)
...
+
 /****************************************************************************
 **
 *F  Call1ArgsInNewReader(Obj f,Obj a) . . . . . . . . . . call a GAP function
 **
 **  The current reader context is saved and a new one is started.
 */
-Obj Call1ArgsInNewReader(Obj f,Obj a)
-
+Obj Call1ArgsInNewReader(Obj f, Obj a)
```
ChrisJefferson: I don't like the complexity and similarity of Call0ArgsInNewReader and Call1ArgsInNewReader now SaveReaderState has been removed.

I'd prefer it either get reintroduced, or these two functions get a common implementation (which shouldn't be too hard, as they only differ on one line?)

src/modules.c
```diff
@@ -155,9 +156,10 @@ Int ActivateModule(StructInitInfo * info)
         if (info->initLibrary) {
             // Start a new executor to run the outer function of the module in
             // global context
-            ExecBegin(STATE(BottomLVars));
+            Bag oldLvars = STATE(CurrLVars);
+            SWITCH_TO_OLD_LVARS(STATE(BottomLVars));
             res = res || info->initLibrary(info);
-            ExecEnd(res);
+            SWITCH_TO_OLD_LVARS(oldLvars);
```
fingolfin: Note that I am not even sure we need to save&restore `CurrLVars` here; but historically we do. But I really wonder what a situation would be were we need to do this, yet don't need to save&restore some of the other global state...?

src/read.c
fingolfin: Yeah, this part of the PR definitely is unfinished. I might indeed revive `SaveReaderState`. (For some background, the single commit in this PR is the result of squashing several commits, minus some changes that were already merged otherwise, plus some extra tweaks. In the original series of commits, I simply did incremental changes as they became "obvious", the end result was not clear to me. As it is now, the minimal diff would just move the `CurrLVars` saving/restoring to `SaveReaderState`. Though I also wonder about the name of that function. And also whether perhaps we ought to save&restore more stuff, e.g. `STATE(Tilde)`. In the end, shouldn't `ReadEvalCommand` and the code here save&restore about the same things? Hmmm

fingolfin: Updated

src/hpc/threadapi.c
```diff
@@ -432,15 +432,15 @@ static GVarDescriptor GVarTHREAD_EXIT;
...
     int i;
 
-    /* initialize everything and begin an interpreter                       */
+    // initialize everything and begin a fresh execution context
     STATE(NrError) = 0;
     STATE(ThrownObject) = 0;
+    oldLVars = STATE(CurrLVars);
+    SWITCH_TO_OLD_LVARS(STATE(BottomLVars));
 
-    IntrBegin(&intr, STATE(BottomLVars));
```
fingolfin: To understand why here and elsewhere, we can get rid of `IntrBegin` and `IntrEnd`, consider what its code:
```C
void IntrBegin(IntrState * intr, Obj frame)
{
    /* allocate a new values stack                                         */
    intr->StackObj = NEW_PLIST(T_PLIST, 64);

    /* must be in immediate (non-ignoring, non-coding) mode                */
    GAP_ASSERT(intr->ignoring == 0);
    GAP_ASSERT(intr->coding == 0);

    /* no return-statement was yet interpreted                             */
    intr->returning = 0;

    /* start an execution environment                                      */
#if BEFORE_THIS_PR
    ExecBegin(frame);
#else
    // remember the old execution state
    intr->oldLVars = STATE(CurrLVars);

    // start an execution environment
    SWITCH_TO_OLD_LVARS(frame);    
#endif
}
```

So all it really does is save the value of `STATE(CurrLVars);` and then invoke `SWITCH_TO_OLD_LVARS`; and as you see, we now exactly this here, instead of calling `IntrBegin`.

Similar for `IntrEnd`: All it does here, effectively (given that we even touch the `IntrState intr`, other than passing it to `IntrBegin` and `IntrEnd`, is to call `SWITCH_TO_OLD_LVARS`. Which we also do here (see new line 470 below).

PR #3847, opened 2020-01-13
FIX: Printing of assoc. words if there are iterated powers

### LINKED ISSUE #3845: bug in view for words in a free group
The following happens in GAP 4.10.2 as well as in the current master branch.

```
gap> f:= FreeGroup( ["x","y"] );;
gap> x:= f.1;;  y:= f.2;;
gap> w1:= y^7*x*y*x*y^3*x*y*x;
y^4*(y^2*(y*x)^2)^2
gap> w2:= y^-7*x*y^16;
y^-7*x*y^16
gap> w:= w1 * w2;
y^4*(y^2*y^16^2)^2*y^-7*x*y^16
```

(The way how `w1` and `w2` are viewed is correct, but the way `w` is viewed is wrong.)

hulpke:
Re-indexing with temporary letters went wrong if there were already powers
stored.

close #3845

## Text for release notes 

An error in printing certain words in free groups (if subexpressions occur as powers iteratedly) was fixed

### DIFF
--- a/lib/wordrep.gi
+++ b/lib/wordrep.gi
@@ -219,9 +219,9 @@ local a,n,t,
    or (IsInt(PRINTWORDPOWERS) and Length(l)<PRINTWORDPOWERS)) and
      ValueOption("printnopowers")<>true then
     if Length(l)>0 and n=infinity then
-      n:=2*(AbsInt(Maximum(l))+1); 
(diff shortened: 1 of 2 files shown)


fingolfin:
Backported in commit d93cd7851

PR #3848, opened 2020-01-13
kernel: fix OnSets to preserve immutability

fingolfin:
OnSets would sometimes return a mutable list when it should have returned an
immutable one; in HPC-GAP this got worse, as the final object was a mutable
and *public* list, which lead to inconsistencies down the road.

Also change `CheckedMakeImmutable` to trigger a break loop when there is an
error, so that "Argument has inaccessible subobjects" errors are easier to
track down.

This is mildly annoying in GAP, but really bad in HPC-GAP, as a critical invariant is violated by this. The issue has been present since "forever", i.e., at least since GAP 4.4.

### DIFF
--- a/src/listfunc.c
+++ b/src/listfunc.c
@@ -1105,10 +1105,10 @@ static Obj FuncOnSets(Obj self, Obj set, Obj elm)
         break;
         
       case 1:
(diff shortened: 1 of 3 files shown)


fingolfin:
Backported in commit 09fd08215a36ac0e02f4a7cd72b0727c76271576

PR #3849, opened 2020-01-13
kernel: prevent immutable -> mutable conversion

fingolfin:
This is inspired by PR #3848, and should help catch similar issues resp. prevent us from regressing in this area.

This imposes a small performance overhead for `RetypeBag`; however, we perform this rather rarely, so I kinda hope it doesn't matter, but I have not yet performed any benchmarks. If there turns out to be a performance issue, one could move the `RetypeBag` into a header and mark it as `inline`; and/or restrict this check to kernel debug mode (now that I think about it, I probably want to do the latter)...

But for now, I just wanted this PR out there to get some initial traction and see if it uncovers any issues in the test suite.

### DIFF
--- a/src/bags.c
+++ b/src/bags.c
@@ -24,14 +24,58 @@ TNumInfoBags InfoBags[NUM_TYPES];
 UInt8 SizeAllBags;
 
 
+#ifdef GAP_KERNEL_DEBUG
+void PrecheckRetypeBag(Bag bag, UInt new_type)
+{
+    // If the new type contains an (im)mutability bit, verify that this does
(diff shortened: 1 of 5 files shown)


fingolfin:
Seems I poked a bee hive; there are a bunch of things I need to untangle to get this ready.

fingolfin:
@ChrisJefferson @wilfwilson I've overhauled this now to work around the crashes -- this required adding some hacks, which I tried to document. I'd appreciate if you could look at that again, just to make sure what I wrote there is clear enough (or if not, to use your feedback to improve it).

PR #3850, opened 2020-01-15
Fix assertion in Factors for polynomials over cyclotomics

fingolfin:
The function at the start splits of all factors x^k, and `pol` is set to the
resulting polynomial. But `factors` contains all factors of the original
polynomial; so the assertion check would then fail.

### DIFF
--- a/lib/cyclotom.gi
+++ b/lib/cyclotom.gi
@@ -1941,7 +1941,7 @@ InstallMethod( Factors,
     fi;
 
     # Store the factorization.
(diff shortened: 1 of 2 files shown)


hulpke:
Thank you -- yes, this never got changed properly.

fingolfin:
It triggered in the manual tests of the `corelg` package, that's how I came across it :-)

fingolfin:
Backported in commit 5214aea8e

PR #3851, opened 2020-01-15
A bunch of minor changes, fixes, tweaks

fingolfin:
While looking into making PR #3849 work, I discovered a bunch of things I should change. Which in turn  lead me to make a bunch of sort of unrelated tweaks, which this PR collects. With these in place, the following PR(s) will be tidier.

The main thing I realized while working on PR #3849 is that I should also try to detect when public objects in HPC-GAP get their type changed; which resulted in me realizing that we sometimes `Objectify` immutable objects, which shouldn't be allowed, and isn't for records, but due to insufficient input validation, we let it slip through for lists for a long time. This PR fixes a bunch of `Objectify` calls. A couple more are in the `corelg` package, see https://github.com/gap-packages/corelg/pull/8); and I won't be surprised if we discover more, although grepping for `Objectify.*Immutable` through all packages didn't reveal anything (but that doesn't mean too much, e.g. `Objectify` and `Immutable`/`MakeImmutable` could be on different lines.

Anyway, I hope we can change the code to issue a warning if given an immutable plain list; and perhaps one distant day, we can even turn it into an error.

Another thing I came upon once again is that I really want to move `IMMUTABLE` out of the TNUM, and turn it into an object flag. Once of the minor changes needed for that is to come up with a convention for naming object flags, so I introduce one, and rename the two existing object flags.

As to the rest, please just look at the commits and the commit messages.


dev/listarith.tex
```diff
@@ -161,14 +161,14 @@ \section{Determining the Type of Plists}
 **
 **  TypePlist works with KTnumPlist to determine the type of a plain list
 **  Considerable care is needed to deal with self-referential lists. This is
-**  basically achieved with the TESTING flag in the Tnum. This must be set in
+**  basically achieved with the OF_TESTING flag in the Tnum. This must be set in
```
wilfwilson: Does the `OF` in `OF_TESTING` stand for `object flag`? I keep trying to pronounce it in my head as the English with "of", resulting in "of testing", which is weird.

I think it would be nice if you explicitly wrote somewhere that `OF_` things are object flags.

fingolfin: Yes it is meant to stand for "object flags". We already have three related functions: `TEST_OBJ_FLAG`, `SET_OBJ_FLAG`, `CLEAR_OBJ_FLAG`. So of course we could also use `OBJ_FLAG_TESTING` or `TESTING_OBJ_FLAG`, but that seemed a bit longish. Anyway, I'd be happy to change this, too.

wilfwilson: I'm not necessarily asking you to change it; `OBJ_FLAG_TESTING` is certainly longish. Another suggestion is `O_F_TESTING`, which is a compromise in terms of length, and avoids suggesting that `OF` is a word, but is also uglyish...

Perhaps @ChrisJefferson has an opinion.

fingolfin: My personal preference is `OF_TESTING` > `OBJ_FLAG_TESTING` > `O_F_TESTING`, although I do see the potential for confusion with `OF_` and the word "of". I am happy to settle on `OBJ_FLAG_TESTING` then. But sure, let's see if @ChrisJefferson has thoughts. In the end, though, we can quite easily repaint the bikeshed later on ;-)

fingolfin: I've now switched this PR to `OBJ_FLAG_TESTING` resp. `OBJ_FLAG_TESTED`.

ChrisJefferson: I think I also have slight preference against `OF`, the new words are perfectly fine :)

PR #3852, opened 2020-01-15
kernel: refactor SET_TYPE_OBJ, add some input validation and tests

fingolfin:
This touches upon a particular dark corner (IMHO) of the code, were I always wondered why certain checks were done one way in one place, and a different way in another place; and why there was no input validation at all.

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -1394,14 +1394,17 @@ Bag NewBag (
 void RetypeBagIntern(Bag bag, UInt new_type)
 {
     BagHeader * header = BAG_HEADER(bag);
+    UInt        old_type = header->type;
+
+    // exit early if nothing is to be done
+    if (old_type == new_type)
+        return;
 
 #ifdef COUNT_BAGS
     /* update the statistics      */
     {
-          UInt                old_type;       /* old type of the bag */
           UInt                size;
 
-          old_type = header->type;
           size = header->size;
           InfoBags[old_type].nrLive   -= 1;
           InfoBags[new_type].nrLive   += 1;
--- a/src/objects.c
+++ b/src/objects.c
@@ -1540,10 +1540,7 @@ static Obj FuncSET_TYPE_DATOBJ(Obj self, Obj obj, Obj type)
     ReadGuard( obj );
 #endif
     SET_TYPE_DATOBJ(obj, type);
-#ifdef HPCGAP
-    if (TNUM_OBJ(obj) != T_DATOBJ)
-#endif
-      RetypeBag( obj, T_DATOBJ );
+    RetypeBag( obj, T_DATOBJ );
     CHANGED_BAG( obj );
     return obj;
 #endif
--- a/src/objects.c
+++ b/src/objects.c
@@ -230,7 +230,10 @@ void SET_TYPE_OBJ(Obj obj, Obj type)
         break;
 
     default:
-        if (IS_PLIST(obj)) {
+        if (IS_STRING_REP(obj)) {
+            // FIXME/TODO: Hap calls Objectify on a string...
+        }
+        else if (IS_PLIST(obj)) {
 #ifdef HPCGAP
             MEMBAR_WRITE();
 #endif
@@ -1337,7 +1340,10 @@ static Obj FuncSET_TYPE_POSOBJ(Obj self, Obj obj, Obj type)
     case T_POSOBJ:
         break;
     default:
-        if (!IS_PLIST(obj)) {
+        if (IS_STRING_REP(obj)) {
+            // FIXME/TODO: Hap calls Objectify on a string...
+        }
+        else if (!IS_PLIST(obj)) {
             ErrorMayQuit("You can't make a positional object from a %s",
                          (Int)TNAM_OBJ(obj), 0);
         }
--- a/src/objects.c
+++ b/src/objects.c
@@ -239,7 +239,7 @@ void SET_TYPE_OBJ(Obj obj, Obj type)
             CHANGED_BAG(obj);
         }
         else {
-            ErrorQuit("cannot change type of a %s", (Int)TNAM_OBJ(obj), 0);
+            ErrorMayQuit("cannot change type of a %s", (Int)TNAM_OBJ(obj), 0);
         }
         break;
     }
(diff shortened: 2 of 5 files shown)


src/objects.c
```diff
@@ -201,28 +199,44 @@ void SET_TYPE_OBJ(Obj obj, Obj type)
         break;
 #endif
     case T_PREC:
+#ifdef HPCGAP
+        MEMBAR_WRITE();
+#endif
```
fingolfin: @rbehrends based on the discussion we had earlier on Slack, I understood you as saying this `MEMBAR_WRITE` and several afterwards in function `SET_TYPE_OBJ`, could go. Could you please confirm or correct this?

Also, can't we likewise remove `ReadGuard` statements?

src/objects.c
```diff
@@ -1347,30 +1325,24 @@ static Obj FuncIS_POSOBJ(Obj self, Obj obj)
...
-        SET_TYPE_POSOBJ(obj, type);
-        RetypeBag( obj, T_POSOBJ );
-        CHANGED_BAG( obj );
+    default:
+        if (!IS_PLIST(obj)) {
+            ErrorMayQuit("You can't make a positional object from a %s",
+                         (Int)TNAM_OBJ(obj), 0);
+        }
+        // TODO: we should also reject immutable plists, but that risks
+        // breaking existing code
```
fingolfin: Hopefully this can be addressed in the future. I contemplated printing a warning, but it's too soon for that (see also PR #3851 and https://github.com/gap-packages/corelg/pull/8)

src/objects.c
rbehrends: I've looked at it again. The guards and memory barriers were put in under the assumption that the type could be changed concurrently. This should not normally happen, but if we (as we did back then for T_DATOBJ) put them in the public region, it is technically still possible. The guards can go regardless, but if we want to permit changing the type again, we may still need memory barriers.

fingolfin: But in all the cases in `SET_TYPE_OBJ`, we end up writing to the object, and for that call `ADDR_OBJ` -- shouldn't that indeed automatically insert even a `WriteGuard` ? How can we then ever change anything concurrently in this function? Maybe my understanding of how guards are supposed to work simply is wrong?

fingolfin: For that matter, I also don't understand how the handling of atomic objects works a few lines above this: sure, we take a hash lock for those objects. But regardless of this, shouldn't a guard violation be triggered the moment we invoke `ADDR_OBJ` (at least once guards work again, with PR #2845) ? 

Clearly I am missing something fundamental.

rbehrends: The specific case we are looking at here is objects in the public region. Guard checks will always succeed for the public region. (This is also why atomic objects never trigger guard errors and the relevant code is actually bracketed with `#ifndef WARD_ENABLED` to excise even those checks for performance reasons.)

Now, with `T_DATOBJ` we have the problem that the code as it exists makes the objects public. Alternatively, we can make them readonly, and there is a setting for that (`ReadOnlyDatObjs`). However, that is something we had disabled by default, because it caused a slew of errors.

But as `T_DATOBJ` objects are now public to avoid these errors, they can in theory be modified again and concurrently by `SET_TYPE_DATOBJ()`, so memory barriers may be needed in case somebody does that.

For `T_COMOBJ` and `T_POSOBJ`, both guards or memory barriers should not be needed anymore. As I recall, this was an unrelated artifact caused by how type setting functions originally worked, bypassing guard checks, and with `unward`, these are now protected by guards by default.

fingolfin: @rbehrends you say "the relevant code is actually bracketed with `#ifndef WARD_ENABLED`" -- well, there are calls to `ADDR_OBJ` for atomic objects just a few lines above the line we are discussing right now, and I see no `#ifndef WARD_ENABLED` there? Nor were there any in the old location of that code. And in `ADDR_OBJ` resp. `PTR_BAG` resp. the guard, there isn't any code  do anything special to handle atomic objects -- or is there? I tried to find such a check in PR #2845 but couldn't find anything. Did I miss it?

fingolfin:
@wilfwilson I fixed those typos, thank you!

src/gasman.c
```diff
@@ -1330,14 +1330,17 @@ void            RetypeBag (
     UInt                new_type )
 {
     BagHeader * header = BAG_HEADER(bag);
+    UInt        old_type = header->type;
```
ChrisJefferson: Is this check worth doing, as all we do is assign header->type=new_type?

fingolfin: It's irrelevant in single threaded mode. In multi-threaded code on multiple cores, it can invalidate cache lines for no good reason and thus degrade performance.

src/objects.c
rbehrends: I'm sorry, I meant to write "the relevant code in `src/hpc/aobjects.c`", because that's where the hot paths are to be found. Normally, a successful guard check on public objects is (by design) very cheap, as it only compares the region pointer to `NULL`. The cost is generally low enough to not justify polluting every location where an access of an atomic object happens with an `#ifndef WARD_ENABLED`, which is also potentially error-prone. But in `aobjects.c`, these accesses are all over the place, so they have been turned off for the performance-critical parts.

src/gasman.c
ChrisJefferson: At first I thought gcc/clang would optimise this away, but on further thought they aren't allowed to (as the memory could be read-only). It's fine anyway.

src/objects.c
```diff
@@ -1350,30 +1331,27 @@ static Obj FuncIS_POSOBJ(Obj self, Obj obj)
...
+    case T_POSOBJ:
         break;
-      default:
-        MEMBAR_WRITE();
-        SET_TYPE_POSOBJ(obj, type);
-        RetypeBag( obj, T_POSOBJ );
-        CHANGED_BAG( obj );
+    default:
+        if (IS_STRING_REP(obj)) {
+            // FIXME/TODO: Hap calls Objectify on a string...
```
fingolfin: Had to add this workaround; see https://github.com/gap-packages/hap/pull/27

src/objects.c
```diff
@@ -204,28 +202,47 @@ void SET_TYPE_OBJ(Obj obj, Obj type)
...
+        CHANGED_BAG(obj);
         break;
     case T_DATOBJ:
         SetTypeDatObj(obj, type);
         break;
 
     default:
-        if (IS_PLIST(obj)) {
+        if (IS_STRING_REP(obj)) {
+            // FIXME/TODO: Hap calls Objectify on a string...
```
fingolfin: Had to add this workaround; see https://github.com/gap-packages/hap/pull/27

fingolfin:
@ChrisJefferson I had to add a slightly workaround for Hap. Would be great if you had time to very briefly glance at it (it's trivial, just want to make sure you are aware of it; the workaround is in its own commits, so we can easily revert it, should we want to).

ChrisJefferson:
I'm kind of surprised that "works", but this is the right workaround for now.

PR #3853, opened 2020-01-15
kernel: simplify FuncUNITE_SET

fingolfin:


src/set.c
```diff
@@ -561,36 +559,26 @@ static Obj FuncUNITE_SET(Obj self, Obj set1, Obj set2)
...
         else {
-            lenr++;
-            SET_ELM_PLIST( TmpUnion, lenr, e2 );
-            CHANGED_BAG( TmpUnion );
+            PushPlist( TmpUnion, e2 );
             i2++;
         }
     }
     while ( i1 <= len1 ) {
         e1 = ELM_PLIST( set1, i1 );
```
wilfwilson: How about
```suggestion
         e1 = ELM_PLIST( set1, i1++ );
```
and then removing the separate `i1++` below (and similarly for the next block of code)? Just an idea.

src/set.c
```diff
@@ -561,36 +559,26 @@ static Obj FuncUNITE_SET(Obj self, Obj set1, Obj set2)
...
```
wilfwilson: My main 'worry' about this function is that `e1` and `e2` are both re-accessed on every iteration of this loop, even though 'typically' only one of `e1` and `e2` has changed. I didn't immediately come up with a concise way of only accessing when necessary. Perhaps there's basically no cost to doing `ELM_PLIST` anyway?

fingolfin: There *is* some cost to it, although it's low. So I agree there are further improvements possible in this function! My main goal with this patch was to make the code easier to understand (I actually created this months ago, while looking at something else -- forgot what -- and struggled a bit verifying the copying loop in the end. The new code actually might be a (very small) de-optimization, but IMHO here clarity is more important -- we already had some subtle bugs in code related to computing unions...

src/set.c
fingolfin: well, if we change this, wouldn't it be even more natural to switch from a `while` loop to a `for` loop, like `for ( ; i1 <= len1; i1++ ) {`?

wilfwilson: Oh. Of course! Silly me. Do so if you feel like it.

src/set.c
wilfwilson: Sure. I agree that simplicity is a good thing to aim for (and you were clear that that was your goal).

wilfwilson: (I'm happy for the PR to be merged as-is.)

PR #3859, opened 2020-01-17
Travis: use curl instead of wget for bootstrap

fingolfin:
Supposedly curl also retries after DNS resolution failures, so this might help with the intermittent failures we are seeing on Travis.

Or maybe not. The only way to find out is to try, I guess. If this passes, and we merge it, and the issue doesn't pop up again, great; if not, at least if have some fresh data.


wilfwilson:
(I’m assuming my correction to the title is actually correct; my apologies if not. If it’s right you maybe want to update your commit message.

fingolfin:
Ooops, silly me, thanks!

PR #3861, opened 2020-01-18
ENHANCE: Extended range of simple groups

hulpke:
(based on work of Stefan Kohl)

This is -- as basic first option -- simply extending the precomputed list of simple groups of "small" orders from 10^18 to about 10^27

## Text for release notes 

The range of `SimpleGroupsIterator` has been extended by 9 magnitudes


grp/simple.gi
```diff
@@ -1165,7 +1179,7 @@ local a;
   if n<168 then return 5;fi; 
   a:=Filtered(SIMPLEGPSNONL2,x->x[1]=n);
   # we have degree data up to order 2^55
-  if n<=10^55 and Length(a)=0 then
+  if n<=2^55 and Length(a)=0 then
```
fingolfin: The documentation for `SimpleGroupsIterator` should also be updated. It currently says that "up to order 10^18" is supported, which clearly is outdated.

Ideally, also the documentation and code for `SmallSimpleGroup` and  `AllSmallNonabelianSimpleGroups ` should be adjusted: their documentation says the limit is 10^6, their code enforces the limit 10^18. Perhaps the limit (10^18, 10^55, 2^55, whatever it may be) could be put into a global variable or global constant, and then used from there? (well, perhaps it can't be used in the documentation this way, but at least the other places; and then a comment next to that global variable could say something like "Note: when changing this value, please also update the documentation for X, Y, Z."

@hulpke would you perhaps willing to do that? If you lack time, of course we can do it in another PR.

fingolfin:
Also: nice patch, thank you :-)

grp/simple.gi
hulpke: I'll change documentation.

This is actually separate, fixing a stupid typo. For groups up to order 2^55 the smallest permutation degree is stored. (This is used in the library to decide whether a permutation representation is of sufficiently small degree.)

grp/simple.gi
```diff
@@ -703,16 +703,30 @@ local NextL2PrimePowerInt;
...
 local a,l,pos,g,b;
   if it!.done then return fail;fi;
   a:=it!.b;
-  if a>1259903 then
-    # 1259903 is the last prime power whose L2 order is <10^18
-    Error("List of simple groups is only available up to order 10^18");
+  if a>=1316848669 then
+    # 1316848669 is the first prime power whose L2 order is beyond the
+    # simple2 list
+    Error("List of simple groups only available up to ",
```
fingolfin: Did you drop "order" intentionally? If not:
```suggestion
    Error("List of simple groups only available up to order",
```

grp/simple.gi
```diff
@@ -775,8 +789,8 @@ InstallGlobalFunction(SimpleGroupsIterator,function(arg)
     stack:=a[3];
     a:=a[2];
   until SizeL2Q(b)>=start;
-  pos:=First([1..Length(SIMPLEGPSNONL2)],x->SIMPLEGPSNONL2[x][1]>=start);
-
+  if start>=10^18 then LOADSIMPLE2(); fi;
+  pos:=First([1..Length(SIMPLEGPSNONL2)],x->SIMPLEGPSNONL2[x][1]>=start);
```
fingolfin: Optional alternative (should be double checked ;-):
```suggestion
  pos:=PositionProperty(SIMPLEGPSNONL2,x->x[1]>=start); 
```

Stefan-Kohl:
Why 10^27? -- This bound seems a bit arbitrary to me. -- Up to order 10^30, there are 2171 non-PSL2's, up to order 10^40, there are 22709 non-PSL2's, up to order 10^50, there are 298023 non-PSL2's and up to order 10^60, there are 4263813 non-PSL2's. -- Of course listing all the latter in the way you do it now would take hundreds of megabytes of text. -- But that's the reason for the iterator code, instead of using just a list.

hulpke:
@Stefan-Kohl Yes, 10^27 is arbitary. I don't claim this is the best and end-of-all. If you want to offer your (better) general iterator code as PR, please do so, and I'd be happy to have my code replaced.

PR #3862, opened 2020-01-20
Rename QUIT_GAP, FORCE_QUIT_GAP and GAP_EXIT_CODE to QuitGap, ForceQuitGap and GapExitCode (the old names remain available as synonyms)

ChrisJefferson:
A usual gap rule is CamelCase functions are for users, and ALL_CAPS are internal.

I think enough people are now using QUIT_GAP, FORCE_QUIT_GAP and GAP_EXIT_CODE that they should be moved to CamelCase (obviously leaving the old ALL_CAPS versions as obsolete).

### DIFF
--- a/doc/ref/mloop.xml
+++ b/doc/ref/mloop.xml
@@ -986,14 +986,19 @@ or <C>brk></C> or <C>brk_<A>nn</A>></C> prompt.
 </ManSection>
 
 <ManSection>
-<Func Name="GapExitCode" Arg='ret'/>
+<Func Name="GapExitCode" Arg='[ret]'/>
 
 <Description>
-A <C>GapExitCode</C> sets the return value which will be used when
-&GAP; exits. This may be an integer, or a boolean (where
-<K>true</K> is interpreted as 0, and
-<K>false</K> is interpreted as 1.
+<Ref Func="GapExitCode"/> sets the exit value which is returned
+to the operating system (or parent process) when &GAP; exits.
+This may be an integer in the range [-128..127] (other values
+are reduced modulo 256), or a boolean. <K>true</K> corresponds
+to the return value 0, which by convention is treated as "success".
+<K>false</K> corresponds to the return value 1, which by convention
+is treated as "failure". The exit value is not changed if no argument
+is given.
 <P/>
+The <E>previous</E> exit code is returned.
 </Description>
 </ManSection>
 
@@ -1001,9 +1006,10 @@ A <C>GapExitCode</C> sets the return value which will be used when
 <Func Name="QuitGap" Arg='[ret]'/>
 
 <Description>
-A <C>QuitGap</C> acts similarly to the keyword <C>quit</C>. It exits
-&GAP; cleanly, calling any function installed using <C>InstallAtExit</C>.
-The optional argument will be passed to <C>GapExitCode</C>.
+<Ref Func="QuitGap"/> acts similarly to the keyword <C>QUIT</C>, except
+<C>QUIT</C> can not be called from a function. It exits
+&GAP; cleanly, calling any function installed using <Ref Func="InstallAtExit"/>.
+The optional argument <A>ret</A> will be passed to <Ref Func="GapExitCode"/>.
 <P/>
 </Description>
 </ManSection>
@@ -1012,10 +1018,11 @@ The optional argument will be passed to <C>GapExitCode</C>.
 <Func Name="ForceQuitGap" Arg='[ret]'/>
 
 <Description>
-A <C>ForceQuitGap</C> is similar to <C>QuitGap</C>, except it ignores any
-functions installed with <C>InstallAtExit</C>, or any other functions
-normally run at GAP exit, and exits GAP immediately.
-The optional argument will be passed to <C>GapExitCode</C>.
+<Ref Func="ForceQuitGap"/> is similar to <Ref Func="QuitGap"/>, except it ignores any
+functions installed with <Ref Func="InstallAtExit"/>, or any other functions
+normally run at GAP exit, such as flushing any partially outputted lines
+to both the screen and files, and exits GAP immediately.
+The optional argument <A>ret</A> will be passed to <Ref Func="GapExitCode"/>.
 <P/>
 </Description>
 </ManSection>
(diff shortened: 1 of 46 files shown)


ChrisJefferson:
We could decide that QUIT_GAP -> QuitGap is OK, but leave FORCE_QUIT_GAP as all caps, as that function is "less user safe"...

lib/test.gi
```diff
@@ -931,9 +931,9 @@ InstallGlobalFunction( "TestDirectory", function(arg)
 
   if opts.exitGAP then
     if testTotalFailures = 0 then
-      QUIT_GAP(0);
+      QuitGap(0);
     else
-      QUIT_GAP(1);
+      QuitGap(1);
```
fingolfin: Independent of this PR: looking at this PR, I discovered `SetExitValue` and learned that one could rewrite the above code as `QuitGap(testTotalFailures = 0);`. But I am not aware of *any* code using that.  So, is it worth keeping this complication? If so, shouldn't we then actually use it here and elsewhere (perhaps not as part of this PR)?
On the other hand, it seems to me that code like `QuitGap(testTotalFailures = 0);`

src/gap.c
```diff
@@ -1122,32 +1122,32 @@ static int SetExitValue(Obj code)
...
-static Obj FuncQUIT_GAP(Obj self, Obj args)
+static Obj FuncQuitGap(Obj self, Obj args)
 {
   if ( LEN_LIST(args) == 0 ) {
     SystemErrorCode = 0;
   }
   else if ( LEN_LIST(args) != 1 
             || !SetExitValue(ELM_PLIST(args, 1) ) ) {
-    ErrorQuit( "usage: QUIT_GAP( [ <return value> ] )", 0, 0);
+    ErrorQuit( "usage: QuitGap( [ <return value> ] )", 0, 0);
```
fingolfin: BTW, I am not aware of any code calling `QuitGap` or `ForceQuitGap` without an exit code. Are you? What would be a use case?

ChrisJefferson: I don't think I've done it myself. I suppose someone might want to just quit gap, but asking for a return value doesn't feel unreasonable.

lib/test.gi
ChrisJefferson: So, I have used SetExitValue(1), so if GAP exits in the middle of my program (say because of an error), the exit value ends up set to a non-zero value.

Using QUIT_GAP(boolean), I don't remember ever seeing. We could probably remove it.

ChrisJefferson: Looking at packages, there are a couple of `QUIT_GAP()`, but no passing in of booleans I can see.

fingolfin: I agree that `SetExitValue` is useful to have, yes.

lib/obsolete.gd
```diff
@@ -760,3 +760,8 @@ BindGlobal("STRING_LIST_DIR", function(dirname)
       return JoinStringsWithSeparator( list, "\000" );
     fi;
 end);
+
+
+DeclareObsoleteSynonym("GAP_EXIT_CODE", "GapExitCode", 2);
```
fingolfin, line 765: Please also add a comment here that explains why these are here, like we have for everything else in here. It really helps to have these.

src/gap.c
```diff
@@ -1134,32 +1134,32 @@ static int SetExitValue(Obj code)
 
 /****************************************************************************
 **
-*F  FuncGAP_EXIT_CODE() . . . . . . . . Set the code with which GAP exits.
+*F  FuncGapExitCode() . . . . . . . . Set the code with which GAP exits.
 **
 */
 
-static Obj FuncGAP_EXIT_CODE(Obj self, Obj code)
+static Obj FuncGapExitCode(Obj self, Obj code)
```
fingolfin, line 1141: So tangentially related to this PR (which BTW needs to be rebased, then it could be merged): I just had need to query the exit code, and there simply is no way to do this. However, the natural name for a function to do this is taken... Can we either rename this to `SetGapExitCode` (and then later we can can add `GapExitCode` or `GetGapExitCode` to query it); or else change this function to return the exit code, and then also to allow calling it with 0 args, to just get the exit code w/o changing it?

fingolfin:
I took the liberty of rebasing this PR.

doc/ref/mloop.xml
```diff
@@ -986,36 +986,41 @@ or <C>brk></C> or <C>brk_<A>nn</A>></C> prompt.
...
 
 <Description>
-A <C>FORCE_QUIT_GAP</C> is similar to <C>QUIT_GAP</C>, except it ignores any
+A <C>ForceQuitGap</C> is similar to <C>QuitGap</C>, except it ignores any
 functions installed with <C>InstallAtExit</C>, or any other functions
-normally run at GAP exit, and exits GAP immediately.
-The optional argument will be passed to <C>GAP_EXIT_CODE</C>.
+normally run at GAP exit, such as flushing any partially outputted lines
+to both the screen and files, and exits GAP immediately.
+The optional argument <A>ret</A> will be passed to <C>GapExitCode</C>.
```
fingolfin, line 1023:
```suggestion
The optional argument <A>ret</A> will be passed to <Ref Func="GapExitCode"/>.
```

doc/ref/mloop.xml
```diff
@@ -986,36 +986,41 @@ or <C>brk></C> or <C>brk_<A>nn</A>></C> prompt.
...
-<Func Name="QUIT_GAP" Arg='[ret]'/>
+<Func Name="QuitGap" Arg='[ret]'/>
 
 <Description>
-A <C>QUIT_GAP</C> acts similarly to the keyword <C>quit</C>. It exits
+A <C>QuitGap</C> acts similarly to the keyword <C>QUIT</C>, except
+<C>QUIT</C> can not be called from a function. It exits
 &GAP; cleanly, calling any function installed using <C>InstallAtExit</C>.
-The optional argument will be passed to <C>GAP_EXIT_CODE</C>.
+The optional argument <A>ret</A> will be passed to <C>GapExitCode</C>.
```
fingolfin, line 1010:
```suggestion
The optional argument <A>ret</A> will be passed to <Ref Func="GapExitCode"/>.
```

PR #3865, opened 2020-01-22
FIX: Induced fitting free setup requires own pcisom

### LINKED ISSUE #3864: error in ConjugacyClasses computation
Gabriel Navarro has reported the following problem.
It happens in GAP 4.10.2 as well as in the current master branch.

```
gap> g:= Group( (1,2,3)(7,8)(12,13), (1,3)(2,4,5)(6,7,9,12,11,8,10,13), 
>               (1,4,3)(6,7)(8,11)(9,13)(10,12) );;
gap> max:= MaximalSubgroupClassReps( g );;
gap> ConjugacyClasses( max[1] );
Error, should not happen nonslv stabstack at /.../gapmaster/gap/lib/fitfree.gi:662 called from
OrbitsRepsAndStabsVectorsMultistage( o.stabradgens, 
 List( o.stabradgens, function ( x )
        return bas * npcgsact( x ) * basinv;
    end ), pcisom, o.stabrsubsz, solvtriv, o.stabfacgens, 
 List( o.stabfacgens, function ( x )
        return bas * npcgsact( x ) * basinv;
    end ), o.stabfacimgs, hom, o.subsz, OnRight, el 
 ) at /.../gapmaster/gap/lib/clashom.gi:2148 called from
LiftClassesEATrivRep( G, mpcgs, i, fants, hom, pcisom, solvtriv 
 ) at /.../gapmaster/gap/lib/clashom.gi:2478 called from
ConjugacyClassesViaRadical( G 
 ) at /.../gapmaster/gap/lib/clashom.gi:3087 called from
<function "ConjugacyClasses perm group">( <arguments> )
 called from read-eval loop at *stdin*:4
you can 'return;' to continue
```

hulpke:
If we simply copy the pcisom, some routines that use the source get the
wrong group. This causes problem is using such algorithms on subgroups
obtained the same way.

This fixes #3864

## Text for release notes 

A bug in setting up a subgroup data structure by the solvable radical method has been fixed.

### DIFF
--- a/lib/clashom.gi
+++ b/lib/clashom.gi
@@ -2474,7 +2474,7 @@ local r,	#radical
 	Info(InfoHomClass,3,"central step");
 	new:=LiftClassesEANonsolvCentral(G,mpcgs,i,hom,pcisom,solvtriv,fran);
       elif Length(fants)>0 and Order(i[1])=1 then
-	# special case for trivial representetive
+	# special case for trivial representative
 	new:=LiftClassesEATrivRep(G,mpcgs,i,fants,hom,pcisom,solvtriv);
 	if new=fail then
 	  new:=LiftClassesEANonsolvGeneral(G,mpcgs,i,hom,pcisom,solvtriv,fran);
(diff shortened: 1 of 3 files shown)


fingolfin:
Backported in commit e41efcd65

PR #3866, opened 2020-01-23
julia: don't Panic in RetypeBag

fingolfin:
There is no need to bring down the whole session for this, it is perfectly safe to call `ErrorQuit` here instead.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -877,7 +877,7 @@ void RetypeBagIntern(Bag bag, UInt new_type)
         // different type is something better to be avoided anyway. So instead
         // of supporting a feature nobody uses right now, we error out and
         // wait to see if somebody complains.
-        Panic("cannot change bag type to one requiring a 'free' callback");
+        ErrorQuit("cannot change bag type to one requiring a 'free' callback", 0, 0);
     }
     header->type = new_type;
 }


fingolfin:
How would this lead to a resource leak?

rbehrends:
Unless I'm missing something, this error occurs if we're changing from a bag type that does not have an associated finalizer to one that has. The `RetypeBag()` then doesn't occur (and if it did, there would be no way to add a finalizer). But presumably the content of the bag has already been updated to something that requires finalization, so there's a possible resource leak, as the finalizer won't be called.

I'm not sure if that's more than a theoretical concern, and even if it happens, there is no reason to bring down the entire program for the error, as you observed.

Note that this was more of an aside; I think the change is perfectly good, I just wanted to make a note of this.

ChrisJefferson:
I worry some code might not be designed for retypeBag to fail, so you end up with a bag with "broken" contents, which will cause a segfault when later accessed?

fingolfin:
@rbehrends ahhh, excellent point, I hadn't thought about that.

@ChrisJefferson if the user just goes on after that, you are right. But really, this situation is super exceptional. If this error or panic is ever triggered, it almost definitely hints at a serious bug in some kernel code, and will be fixed ASAP.

But that said, it convinces me that we don't need this PR after all: In practice, this panic *never* happens, and if it happens after all, you better redesign your kernel code before you let it use on people.

PR #3867, opened 2020-01-23
julia: remove the need for GAPTypes.jl

fingolfin:
... by doing something moderately evil: If GAP is loaded before Julia, we
create the Julia type `ForeignGAP.MPtr` *without* a super type. Then once
Julia is loaded, and the type `GAP.GapObj` has been created, GAP.jl calls the
new function `GAP_register_GapObj` with this type as argument, and we patch
`ForeignGAP.MPtr` to have `GAP.GapObj` as supertype.

This seems to work well in practice and ought to be safe, as prior to the
moment were we change the super type, no Julia code beyond the Julia GC should
have ever "seen" `ForeignGAP.MPtr` or any instances of it, and hence no
internal data structures of the Julia runtime that might get into an invalid
state due to our change (e.g. anything related to method dispatch) should
reference it.

Also, *don't* set `GapObj` as supertype for `ForeignGAP.Bag` and
`ForeignGAP.LargeBag`, doing that was IMHO simply a mistake. In practice it
should not have mattered so far, as no regular Julia code ever touches
instances of these types directly. This might change in the future though, if
we introduce Julia wrapper objects for GAP objects that directly access their
underlying bits, instead of going through libgap APIs.

---

I tried to write this code so that old versions of GAP.jl will still work with it; the Travis tests will tell whether this works as intended (at least in one direction).

Once this PR here is merged and backported, I'll update GAP.jl to use `GAP_register_GapObj`

----

UPDATE: actually slightly different plan:
- once this PR is approved, I'll merge https://github.com/oscar-system/GAP.jl/pull/327 
- then release a new version of `GAP.jl` which will work with GAP both with and without this PR here
- once the release is public, this PR can be merged
- once this PR has been backported to stable-4.11, I can merge https://github.com/oscar-system/GAP.jl/pull/322 (which means GAP.jl will only work with GAP versions including this PR here, not with older ones
- then I make another GAP.jl release

But before all this, we should determine whether the approach in this PR is really workable. To thi end, I [posted to the Julia discourse](https://discourse.julialang.org/t/swizzling-the-super-type-of-a-foreign-julia-type-or-how-evil-must-i-be/33733), let's see if I get any useful replies.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -653,9 +653,6 @@ AS_IF([test "x$with_julia" != xno ],[
   AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to obtain JULIA_LIBS from julia-config.jl])])
   JULIA_LIBS=${JULIA_LIBS//\'/}
   AC_MSG_RESULT([${JULIA_LIBS}])
-  AC_MSG_NOTICE([installing GAPTypes.jl])
-  ${JULIA} -e 'import Pkg; Pkg.add("GAPTypes")'
-  AS_IF([ test $? != 0 ], [AC_MSG_ERROR([failed to install GAPTypes.jl])])
 ],
 [
   AS_IF( [ test "x$with_gc" = xjulia ],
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
I'm also happy with this, while not understanding Julia

fingolfin:
I just released GAP.jl 0.2.4 so now I'll merge this PR

fingolfin:
Backported in commit 0f20ca1da

PR #3870, opened 2020-01-27
kernel: simplify FuncIS_SUBSET_SET

fingolfin:
Half the code was never used, because before the `IsSet(set2)`, we already had ensured that `set2` refers to a set... This gets rid of yet another `PLAIN_LIST` call. (`PLAIN_LIST` is IMHO evil, like all in-place conversions in GAP, and a regular source of bugs and confusion; while we can't get completely get rid of it, we can at least try to reduce it to a minimum).

### DIFF
--- a/src/set.c
+++ b/src/set.c
@@ -281,67 +281,34 @@ static Obj FuncIS_SUBSET_SET(Obj self, Obj set1, Obj set2)
     UInt                i2;             /* index into the right set        */
     Obj                 e1;             /* element of left  set            */
     Obj                 e2;             /* element of right set            */
-    UInt                pos;            /* position                        */
 
     RequireSmallList("IsSubsetSet", set1);
     RequireSmallList("IsSubsetSet", set2);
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Does the !IsSet and SetList enforce we have PLISTs? Wasn't sure why you don't need the PLAIN_LIST any more.

fingolfin:
So `IsSet` only ever returns 1 / TRUE if the input is a plain list; or if it converted it to a plain list.  If it return 0 / FALSE, then we call `SetList`; but that always returns a plist. (Please double check my claims!)

So in summary, `if ( ! IsSet( set1 ) )  set1 = SetList( set1 );` always ends up with `set1` being a plain list.

fingolfin:
@ChrisJefferson once this is merged, I'll add a PR which essentially removes `IsSet` other than the call in macro `RequireMutableSet` defined in file `set.c`

Issue #3868, opened 2020-01-25
Error, repeated further fixpoint in sanity check


alex-konovalov:
A new error occurs in the sanity check in the master branch of GAP:
```
12,295,64,150
testing group [A(6)^2]2=A(6)wr2
conjugacy classes
rational classes
automorphism group
Error, repeated further fixpoint -- ambiguity at /circa/scratch/gap-jenkins/workspace/GAP-master-sanity/GAPCOPTS/64build/la\
bel/kovacs/GAP-master-snapshot/lib/morpheus.gi:2415 called from
Build step 'Execute shell' marked build as failure
```
This worked a week ago.

alex-konovalov:
@hulpke could this be coming from your new code?

hulpke:
I don;'t think it should be the new code, as this is completely separate. It might be a side-effect of recent changes that generically tried less aggressive permutation degree reduction, but here such reduction might be needed. I have a fix.

PR #3872, opened 2020-01-28
Three fixes for accidental breakages in recent changes

hulpke:
Fix for #3868: Force maximal degree reduction of automorphism group of fitting free (if needed)

Redid fix for #3864:

The previous fix (#3865) created a homomorphism object that is expensive to create for matrix groups. (matgrp package , https://github.com/hulpke/matgrp/issues/12 )

One thus must be more careful in making this map.

(Test is with matgrp package)

Fixed construction of simple groups whose name string includes a +/- 

All problems were only in dev, never in release versions

### DIFF
--- a/grp/simple.gi
+++ b/grp/simple.gi
@@ -184,20 +184,20 @@ local p,f,z,G,o;
   # first generator differs for q=2,p=3
   if q=2 then 
     G:=Group(
-	o*[[1,1,0,0,0,0,0],
-	[0,1,0,0,0,0,0],
-	[0,0,1,1,1,0,0],
-	[0,0,0,1,0,0,0],
-	[0,0,0,0,1,0,0],
-	[0,0,0,0,0,1,1],
-	[0,0,0,0,0,0,1]],
-	o*[[0,0,1,0,0,0,0],
-	[1,0,0,0,0,0,0],
-	[0,0,0,0,0,1,0],
-	[0,0,0,1,0,0,0],
-	[0,1,0,0,0,0,0],
-	[0,0,0,0,0,0,1],
-	[0,0,0,0,1,0,0]]);
+        o*[[1,1,0,0,0,0,0],
+        [0,1,0,0,0,0,0],
+        [0,0,1,1,1,0,0],
+        [0,0,0,1,0,0,0],
+        [0,0,0,0,1,0,0],
+        [0,0,0,0,0,1,1],
+        [0,0,0,0,0,0,1]],
+        o*[[0,0,1,0,0,0,0],
+        [1,0,0,0,0,0,0],
+        [0,0,0,0,0,1,0],
+        [0,0,0,1,0,0,0],
+        [0,1,0,0,0,0,0],
+        [0,0,0,0,0,0,1],
+        [0,0,0,0,1,0,0]]);
   elif p=3 then 
     G:=Group(
     o*[[z^2,0,0,0,0,0,0],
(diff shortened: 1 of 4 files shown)


lib/morpheus.gi
```diff
@@ -2374,8 +2374,20 @@ InstallGlobalFunction(AutomorphismGroupFittingFree,function(g)
 
   aup:=Normalizer(d,Image(emb,g));
 
-  #reduce degree
-  s:=SmallerDegreePermutationRepresentation(aup:cheap);
+  #reduce degree -- avoid redundatnt fixedpoints
```
```suggestion
  #reduce degree -- avoid redundant fixed points
```

lib/morpheus.gi
```diff
@@ -2374,8 +2374,20 @@ InstallGlobalFunction(AutomorphismGroupFittingFree,function(g)
...
+  if count>1 then
+    p:=aup;
+    s:=IdentityMapping(p);
+    repeat
+      i:=SmallerDegreePermutationRepresentation(p);
+      if NrMovedPoints(Image(i))<NrMovedPoints(p) then
+        p:=Image(i,p);
```
fingolfin: Perhaps like this for consistency?
```suggestion
        p:=Image(i);
```
Could also use `ImagesSource(i)` here and in the line above

lib/morpheus.gi
```diff
@@ -2395,13 +2407,14 @@ InstallGlobalFunction(AutomorphismGroupFittingFree,function(g)
...
 
-  # produce data to map fro au to aup:
+  # produce data to map from au to aup:
   lemb:=MovedPoints(aup);
   stbs:=[];
   orb:=Orbits(aup,MovedPoints(aup));
   base:=BaseStabChain(StabChainMutable(aup));
   newbas:=[];
   for i in orb do
+
```
```suggestion
```

lib/morpheus.gi
hulpke: The `Image(i,p)` was deliberate, to ensure that -- even if the homomorphism is defined on a larger set, which I cannot prove will never happen -- the image group is the same. But the first `Image` could be the range.

grp/simple.gi
```diff
@@ -329,15 +329,15 @@ local brg,str,p,a,param,g,s,small,plus,sets;
...
-      if a[p+1]='1' and a[p+2]=',' then
+      elif Length(a)>=p+1 and a[p+1]='1' and a[p+2]=',' then
 	# gave O(+1,8,2) or so
 	plus:=fail;
       fi;
       if plus<>fail then
 	if p=1 then
 	  # leading +-, possibly with comma
 	  a:=a{[2..Length(a)]};
+          if a="1" then a:="";fi;
```
fingolfin: odd indentation (not important)

hulpke:
Added test for the broken case (which only involved the O+/- cases). The syntax for other classes is alreday tested through the simple groups iterator.

grp/simple.gi
hulpke: Seems to be a tab issue. Will change

fingolfin: Thanks

fingolfin:
@hulpke in order to backport this, I'll have to backport PR #3790 as well (the alternative would be to untangle the code, but that doesn't seem to be worth the effort)

fingolfin:
Backported to stable-4.11 in commit fdc4f1b9cca052406804ac9af2037e8ad9539b97

hulpke:
@fingolfin Sorry -- I had added it because I though #3790 had been backported already. Thanks in any case.

PR #3873, opened 2020-01-28
kernel: don't use PLAIN_LIST in OnTuples, OnSets

fingolfin:
... by changing OnSets{Perm,Trans,PPerm} resp. OnTuples{Perm,Trans,PPerm} to accept non-plain list arguments, which then are copied, and modified in place

### DIFF
--- a/tst/testinstall/perm.tst
+++ b/tst/testinstall/perm.tst
@@ -1,4 +1,4 @@
-#@local checklens,n,permAll,permBig,permSml,x,y,moved,p
+#@local checklens,n,permAll,permBig,permSml,x,y,moved,p,r,t,l
 gap> START_TEST("perm.tst");
 
 # Permutations come in two flavors in GAP, with two TNUMs: T_PERM2 for
@@ -471,9 +471,9 @@ gap> OnTuples([(1,2),(1,3)],(1,2,70000));
 
 #
 gap> OnTuples([,1],());
-Error, OnTuples for perm: list must not contain holes
+Error, OnTuples: <tup> must not contain holes
 gap> OnTuples([,1],(70000,70001));
-Error, OnTuples for perm: list must not contain holes
+Error, OnTuples: <tup> must not contain holes
 
 #
 # OnSets for permutations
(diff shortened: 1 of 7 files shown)


ChrisJefferson:
@hulpke : In the long term, it would be nice if we could say that no immutable GAP object changed type. At the moment, in HPC-GAP, if we called PLAIN_LIST on a string being used in another thread, things would break, as most functions aren't prepared for the object they are reading to change type mid-function. Also, if one really tries you can break GAP by doing crazy things like sorting a list with a comparator which changes the type of the list mid-sort.

Also, strings which are intended to be immutable (things like names of record members, of global functions), sometimes get turned into PLISTS because they get passed to a function which calls PLAIN_LIST on them, and this is creates weird issues.

ChrisJefferson:
Just as an example of the kind of thing the various in-place conversion functions can cause (I'm not suggesting we try and fix this in the short term -- I think the only way to fix it would be to (eventually) remove `ConvertToStringRep`)

```
gap> p := ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h'];
"abcdefgh"
gap> Sort(p, function(x,y) ConvertToStringRep(p); return x < y; end);
gap> p;
"\377\000(\000\000\000\000\000"
```

hulpke:
@ChrisJefferson Thank you for the explanation (I hope I did not keep you from talk preparation :-)

fingolfin:
>  [...] apart from forbidding OnSets for objects that claim to be lists [...]

@hulpke actually, this PR does not such thing (or if it does, then that's a bug, and I'd appreciate hints where it happens so I can fix it). To the contrary: one new tests shows that one can now provide objects which pretend to be lists as input to `OnSets` and it'll work, while before an error would be raised.

hulpke:
@fingolfin Thank you -- then I misunderstood what the tests do.

tst/testinstall/perm.tst
```diff
@@ -487,6 +487,19 @@ gap> OnSets([(3,4),(1,2)],(1,2,70000));
...
+#
+gap> r:=NewCategory("ListTestObject",IsSmallList);;
+gap> t:=NewType(ListsFamily, r and IsMutable and IsPositionalObjectRep);;
+gap> InstallMethod(ELM0_LIST,[r,IsPosInt],{l,i}->l![i]);
+gap> InstallMethod(Length,[r],l->3); # hard code length
+gap> l:=Objectify(t,[1,2,4]);;
+gap> OnTuples(l, (1,3));
+[ 3, 2, 4 ]
+gap> OnSets(l, (1,3));
+Error, Panic: cannot convert <list> (is a positional object) to a plain list
```
fingolfin: OK, that's a bug, oops, will fix

fingolfin: Ah, this is caused by `IsSet` being called by `FuncOnSets`. But that is fixed in PR #3871; so let's first merge that PR, and then also PR #3877, after which I can rebase this PR and adjust the test output above, and also remove the references to `ELM0_LIST`

Issue #3854, opened 2020-01-16
AssociatedPartition in GAP4.10.1


rtange:
Minor point:
AssociatedPartition([]); gives an error message, but it should just give [] (as in GAP3).
Please delete or close this issue if this has already been corrected in GAP4.10.2.
```
My GAP banner:
 ┌───────┐   GAP 4.10.1 of 23-Feb-2019
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: i686-pc-cygwin-default32-kv3
 Configuration:  gmp 6.1.2, readline
 Loading the library and packages ...

```

wucas:
I can reproduce this with GAP 4.10.2:
```
 ┌───────┐   GAP 4.10.2 of 19-Jun-2019
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin18.6.0-default64-kv3
 Configuration:  gmp 6.1.2
 Loading the library and packages ...
 Packages:   AClib 1.3.1, Alnuth 3.1.1, AtlasRep 2.1.0, AutoDoc 2019.05.20, AutPGrp 1.10, Browse 1.8.8, Carat 2.2.3, CRISP 1.4.4, Cryst 4.1.19, CrystCat 1.1.9, CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, 
             Forms 1.2.5, GAPDoc 1.6.2, genss 1.6.5, IO 4.6.0, IRREDSOL 1.4, LAGUNA 3.9.3, orb 4.8.2, Polenta 1.3.8, Polycyclic 2.14, PrimGrp 3.3.2, RadiRoot 2.8, recog 1.3.2, ResClasses 4.7.2, 
             SmallGrp 1.3, Sophus 1.24, SpinSym 1.5.1, TomLib 1.2.8, TransGrp 2.0.4, utils 0.63
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap> AssociatedPartition([]);
Error, List Element: <list>[1] must have an assigned value in
  for j in [ 1 .. lambda[1] ] do
    if j <= lambda[k] then
        res[j] := k;
    else
        k := k - 1;
        while j > lambda[k] do
            k := k - 1;
        od;
        res[j] := k;
    fi;
od; at /Users/lucas/Projects/gap-4.10.2/lib/combinat.gi:2478 called from 
<function "AssociatedPartition">( <arguments> )
[...]

Issue #3764, opened 2019-11-29
Some MeatAxe issues


ThomasBreuer:
Rudolf Tange has sent me the following suggestions concerning MeatAxe functionality.

- There are two implementations of `MTX.BasisRadical`, one based on `SMTX.BasesMaximalSubmodules` and one based on `SMTX.DualModule` and `SMTX.BasisSocle`; the first one gets overwritten by the second one.

- The result of `SMTX_BasisSocle` for an irreducible module is mutable, an `ImmutableMatrix` call is missing.

- The result of `MTX.InducedActionFactorModule` is not a module record if `sub` spans the whole module; the documentation of `MTX.InducedActionFactorModule` does not mention this.
(Analogously, `MTX.InducedActionSubmodule` returns a list in the case of a zero submodule.)

- In `MTX.BasesMaximalSubmodules` and `MTX.BasisRadical`, it is not necessary to assume that the matrices are invertible; one could use just the transposed matrices instead of their inverses.

rtange:
One thing I forgot to  mention:
It would be useful to have a function SMTX.RandomMaximalSubGModule.
This is extremely simple to implement: it can be done completely analogously to the derivation of MTX.BasisRadical from MTX.BasisSocle (you derive it from SMTX.RandomIrreducibleSubGModule only using transposes, no inverses).
For example, sometimes you know there is a unique maximal submodule, and then this method will be quicker than using MTX.BasisRadical.

PR #3879, opened 2020-01-29
Resolve some reported minor issues

hulpke:
This fixes AssociatedPartition for the empty partition	#3854 
and some of the MeatAxe issues in #3764

### DIFF
--- a/lib/combinat.gi
+++ b/lib/combinat.gi
@@ -2466,6 +2466,7 @@ InstallGlobalFunction(AssociatedPartition,function(lambda)
   local res, k, j;
   res := [];
   k := Length(lambda);
+  if k=0 then return res;fi; # empty partition
   for j in [1..lambda[1]] do
     if j <= lambda[k] then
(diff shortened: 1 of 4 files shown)


lib/combinat.gi
```diff
@@ -2466,6 +2466,7 @@ InstallGlobalFunction(AssociatedPartition,function(lambda)
   local res, k, j;
   res := [];
   k := Length(lambda);
+  if k=0 then return res;fi; # empty partition
```
fingolfin: A minimal test which just executes `AssociatedPartition([]);` would be nice.

hulpke:
I think I leave the backport question to those who would have to do the work. There shouldn't be a problem with it, but it is an extra step. I'll wait with merging for a day or two so that whoever wants can add the backport tag.

fingolfin:
Backported to stable-4.11 in commit a66ec381e701f2cd92f25c245710cfe781a453bd

PR #3881, opened 2020-02-03
kernel: simplify OnPosIntSetsPartialPerm

fingolfin:
This function seems to do the same as OnSets, except that it allows for the
input value `[ 0 ]` (contrary to its name), and it omits all kinds of argument
validations.

We change it to delegate in most cases to OnSetsPPerm. This adds a tiny bit of
overhead in the core loop(s), but in general I would hope this is not much of
an issue, while the extra input validation ought to help in finding and
preventing certain bugs.

@james-d-mitchell this really should be reviewed by you and not somebody else. In particular, I wonder if I missed some subtle difference. Also, while in "small" examples, the performance penalty seems to be negligible, I do wonder if you have computations where it is not.

The main point behind this PR is that I want to get rid of the `PLAIN_LIST` call here. If this PR is not good for merging, I can provide an alternate one which changes less code, but still gets rid of `PLAIN_LIST`.

### DIFF
--- a/src/pperm.cc
+++ b/src/pperm.cc
@@ -3658,67 +3658,16 @@ static Obj FuncOnPosIntSetsPartialPerm(Obj self, Obj set, Obj f)
     RequireSmallList("OnPosIntSetsPartialPerm", set);
     RequirePartialPerm("OnPosIntSetsPartialPerm", f);
 
-    UInt2 *     ptf2;
-    UInt4 *     ptf4;
-    UInt        deg;
-    const Obj * ptset;
-    Obj *       ptres, tmp, res;
-    UInt        i, k, reslen;
+    const UInt len = LEN_LIST(set);
 
-    if (LEN_LIST(set) == 0)
+    if (len == 0)
         return set;
 
-    if (LEN_LIST(set) == 1 && ELM_LIST(set, 1) == INTOBJ_INT(0)) {
+    if (len == 1 && ELM_LIST(set, 1) == INTOBJ_INT(0)) {
         return FuncIMAGE_SET_PPERM(self, f);
     }
 
-    PLAIN_LIST(set);
-    res = NEW_PLIST_WITH_MUTABILITY(IS_PLIST_MUTABLE(set), T_PLIST_CYC_SSORT,
-                                    LEN_PLIST(set));
-
-    /* get the pointer                                                 */
-    ptset = CONST_ADDR_OBJ(set) + LEN_LIST(set);
-    ptres = ADDR_OBJ(res) + 1;
-    reslen = 0;
-
-    if (TNUM_OBJ(f) == T_PPERM2) {
-        ptf2 = ADDR_PPERM2(f);
-        deg = DEG_PPERM2(f);
-        /* loop over the entries of the tuple                              */
-        for (i = LEN_LIST(set); 1 <= i; i--, ptset--) {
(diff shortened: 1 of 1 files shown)


src/pperm.cc
```diff
@@ -3550,16 +3550,17 @@ Obj OnSetsPPerm(Obj set, Obj f)
             }
         }
     }
-    if (reslen == 0) {
-        RetypeBagSM(res, T_PLIST_EMPTY);
-        return res;
-    }
+
     SET_LEN_PLIST(res, reslen);
     SHRINK_PLIST(res, reslen);
```
fingolfin: Moving the code above downwards is a tiny optimization in case a huge input set is mapped to an empty list. I don't expect this to matter for general real world inputs

src/pperm.cc
```diff
@@ -3657,67 +3658,16 @@ static Obj FuncOnPosIntSetsPartialPerm(Obj self, Obj set, Obj f)
...
-    ptset = CONST_ADDR_OBJ(set) + LEN_LIST(set);
-    ptres = ADDR_OBJ(res) + 1;
-    reslen = 0;
-
-    if (TNUM_OBJ(f) == T_PPERM2) {
-        ptf2 = ADDR_PPERM2(f);
-        deg = DEG_PPERM2(f);
-        /* loop over the entries of the tuple                              */
-        for (i = LEN_LIST(set); 1 <= i; i--, ptset--) {
-            k = INT_INTOBJ(*ptset);
```
fingolfin: Note how this code does not verify that `*ptset` is an immediate integer, nor that it is a *positive* immediate integer. With this PR, we effectively add an `IS_POS_INTOBJ` (by calling `OnSetsPPerm`, which has that check). I imagine that if we execute this loop a couple million times, one might be able to measure this; but I didn't conduct any experiments, as I have no idea what a typical use case would be. Anyway, personally I'd take the safety over speed here any day, but I don't feel it's my decision to make for this code!

src/pperm.cc
```diff
@@ -3657,67 +3657,16 @@ static Obj FuncOnPosIntSetsPartialPerm(Obj self, Obj set, Obj f)
...
-    UInt4 *     ptf4;
-    UInt        deg;
-    const Obj * ptset;
-    Obj *       ptres, tmp, res;
-    UInt        i, k, reslen;
+    const UInt len = LEN_LIST(set);
 
-    if (LEN_LIST(set) == 0)
+    if (len == 0)
         return set;
```
fingolfin: Note that this is another dangerous compared to `OnSets`: if `set` is an empty *mutable* set, then `OnSets` returns a new (non-identical) empty list, which helps avoid all kinds of nasty bugs. I didn't touch it here for now as I wasn't sure if this was intentional or not...

wilfwilson:
(But the failure suggests that there is a technical subtlety to the existing code that the Semigroups package relies on.)

fingolfin:
Actually I suspect the failure is caused by the issued fixed in PR #3882, which used to be part of this PR. I've now rebased this PR, let's see if that helps.

fingolfin:
Tests passed just fine after rebasing.

fingolfin:
I'll wait for now to see if @james-d-mitchell has time and interest to review this. It's not urgent after all

fingolfin:
@james-d-mitchell do you think you might have a chance to review this? Or should we just merge it as-is?

PR #3882, opened 2020-02-03
kernel: fix OnSets for pperms producing an empty list

fingolfin:
We sometimes ended up producing an "empty plist of length > 0"; that is, on
the GAP level, length 0 was shown thanks to `LenPlistEmpty` being called; but
the actual *stored* length still was 1. As a result, comparing such a broken
"empty" plist again an actual empty plist returned false.

We fix this by always setting the plist length in OnSetsPPerm.

Fixes https://github.com/gap-packages/orb/issues/52

### DIFF
--- a/src/plist.c
+++ b/src/plist.c
@@ -1035,6 +1035,7 @@ static Int LenPlist(Obj list)
 
 static Int LenPlistEmpty(Obj list)
 {
+    GAP_ASSERT(LEN_PLIST(list) == 0);
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


ChrisJefferson:
Could we also chuck a GAP_ASSERT in LenPlistEmpty, to try to catch this type of issue (won't always catch it of course)?

fingolfin:
Excellent idea. I'd also like to add a ChangeToEmptyPlist helper func and replace RetypeBag calls by that.

fingolfin:
@ChrisJefferson I've now added the suggested assertion.

fingolfin:
Something is wrong with AppVeyor, the tests passed there but somehow GitHub did not get notified. I’ll just merge it anyway.

PR #3887, opened 2020-02-06
kernel: ensure our Julia types always have a super type

fingolfin:
See https://github.com/oscar-system/GAP.jl/issues/337

This should be backported to stable-4.11.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -794,7 +794,7 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
         }
     }
     else {
-        gapobj_type = 0;
+        gapobj_type = jl_any_type;
     }
 
     JuliaTLS = jl_get_ptls_states();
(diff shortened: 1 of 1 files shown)


ThomasBreuer:
I have checked the changes from this pull request, and still get the segmentation fault.
The messages printed by Julia are similar to the ones reported by @fingolfin ,
and they are almost the same with and without the changes.
(The only differences in the two situations are a different ip for the second reported `unknown function`, and a different number of allocations in the end.)

fingolfin:
Please add the full output (including the GAP banner) of both cases to your issue at the GAP.jl repository.

In the meantime, this PR corrects a bug in the code (regardless of whether it is related to the crashes you are seeing or not), so it’d be good to be approved & merged. It also fixes the issue for me, so I’ll need further information about the problem in your case in order to debug it (such as the logs I requested above, which will give me info about the involved GAP version etc.)

fingolfin:
Backported to stable-4.11 in commit 9466a696e

PR #3893, opened 2020-02-12
gac: add rudimentary assembler support

fingolfin:
This is potentially useful for meataxe64, as with this it could use `gac` instead of autotools.

A more sophisticated implementation would add dedicated ASFLAGS / ASMFLAGS, and perhaps also use `$(AS)` instead of the C compiler. But for now, this works quite well.

I'd like to backport this to `stable-4.11`

### DIFF
--- a/cnf/gac.in
+++ b/cnf/gac.in
@@ -327,7 +327,7 @@ while [ $# -gt 0 ]; do
 
     -L|--addlibs)         shift; c_addlibs="${c_addlibs} $1";;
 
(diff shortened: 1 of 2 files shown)


stevelinton:
No problem with this patch. An additional complication with using assembler though is that OS X and Linux have different linkage conventions (whether there is an _ on the start of globals). I have a script that rewrites Richard's assembler for OS X and I need Make to run that.

fingolfin:
Backported in aa0ba380de7b75af2c6af4c9de819f769dad77e5

PR #3896, opened 2020-02-13
Fix and minor modification for Solvable-Radical algorithms

hulpke:
The solvable-radical based conjugacy test (which so far will not be used in the library) ran into a data structure error when standardizing two elements simultaneously. This is fixed.

Included are also two small strategy changes that help in calculating group isomorphism/automorphism groups for larger groups. (Explicit examples are too large for standardized tests, but this should not affect coverage.)


fingolfin:
the CI issues are finally resolved; can you please rebase this so the tests get re-run?

hulpke:
@fingolfin  Is rebased (and tests fine)

fingolfin:
@hulpke I noticed this has a backport label. Is this just for the "FIX" commit in here? Or also for the enhancements?

hulpke:
@fingolfin I had thought about both, as the changes are small and the desire to not support multiple code bases outweighs the tiny risk that they would introduce other problems.

fingolfin:
@hulpke I backported the commits 1 and 3 to stable-4.11 (as 30d3609d4fc9ac6825ae2e05298e6f89ccfcc0a1 and d00e1903e3d6af95cdc54d0401eebe54af37c9b2), but commit 2 ("ENHANCE: Prune series for automorphism group") had a conflict: the code it edits does not exist in 4.11. So I skipped that.

PR #3899, opened 2020-02-18
Provide a default `Unpack` method for lists

ThomasBreuer:
The default `Unpack` method we had up to now delegates to `ShallowCopy` and requires `IsRowVector and IsPlistRep`.
However, this does not cover the rows of the matrix objects that are implemented in the `matrgrp` package, since these objects are not in `IsPlistRep`.

I think it is safe to install `ShallowCopy` as a really generic `Unpack` method that requires only `IsList`.

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -625,8 +625,8 @@ InstallMethod( Unpack,
             ## but avoids code duplication.
 
 InstallOtherMethod( Unpack,
-  "generic method for plain lists",
-  [ IsRowVector and IsPlistRep ],
+  "generic method for lists",
+  [ IsList ],
   ShallowCopy );
 
 InstallMethod( \{\},


ThomasBreuer:
Two Travis tests fail with the message
```
Error, found no Singular executable in PATH while searching the following list:
[ "Singular" ]
 at /home/travis/build/gap-system/gap/pkg/IO_ForHomalg-2019.09.02/gap/IO_ForHo\
malg.gi:309 called from
```
I do not think that this problem arises because of the changes proposed in the pull request.

ThomasBreuer:
@hulpke Thanks for approving, but I cannot merge as long as these tests fail.

fingolfin:
the CI issues are finally resolved; can you please rebase this so the tests get re-run?

PR #3903, opened 2020-02-25
Fixed bug with Cycles((),[])

laurentbartholdi:
# Description

There is a case of Emptyophobia with CyclesOp:

in the previous version, 'Cycles((),[])' triggers an error. It now correctly returns '[]'.

### DIFF
--- a/lib/oprt.gi
+++ b/lib/oprt.gi
@@ -2188,8 +2188,12 @@ InstallMethod( CyclesOp, true, [ IsObject, IsList, IsFunction ], 1,
     IsSSortedList(D);
     blist := BlistList( [ 1 .. Length( D ) ], [  ] );
     orbs := [  ];
-    next := 1;
-    while next <> fail do
+    next := 0;
+    while true do
+        next := Position( blist, false, next );
+        if next = fail then
+            return Immutable( orbs );
+        fi;
         pnt := D[ next ];
         orb := CycleOp( g, D[ next ], act );
         Add( orbs, orb );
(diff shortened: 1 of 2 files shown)


lib/oprt.gi
```diff
@@ -2188,7 +2188,7 @@ InstallMethod( CyclesOp, true, [ IsObject, IsList, IsFunction ], 1,
     IsSSortedList(D);
     blist := BlistList( [ 1 .. Length( D ) ], [  ] );
     orbs := [  ];
-    next := 1;
+    next := Position( blist, false, 0 ); # 1, except false if D is empty
```
fingolfin: Not "false" but "fail". Could be further simplified.
```suggestion
    next := Position( blist, false ); # fail is D is empty, otherwise 1
```
That said, perhaps simple is better here: instead of explaining in a comment that we start at the start unless the list is empty, how about instead using code that makes this explicit and is self-explanatory without any comment: We could just insert something like this at the start of the function:
```gap
    if IsEmpty(D) return MakeImmutable( [] ); fi;
```

laurentbartholdi: OK -- I avoided the inelegant double-call to Position. (I try to avoid handling special cases such as empty data with a separate test, in this case it's easy to do). I also added test cases.

fingolfin: So instead of inserting a single line, you prefer to modify 10 lines? Ah well, "elegance" surely is a very subjective thing :-).

PR #3906, opened 2020-02-28
lib: add warning to BIND_GLOBAL

ssiccha:
# Description
This PR adds a warning to `BIND_GLOBAL`. A warning is printed if `BIND_GLOBAL(name, val)` is called with a `name` which starts with a whitespace. 

It is possible to do `BIND_GLOBAL(" x", 1);` (notice the whitespace) but this variable won't be (?) accessible from the command line. Or is there a way to access such a variable?

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/lib/global.gi
+++ b/lib/global.gi
@@ -57,12 +57,29 @@ end);
 ##
 
 CheckGlobalName := function( name )
+    local pos;
     if not IsString( name ) then
-      Error("CheckGlobalName: the argument must be a string");
+        Error("CheckGlobalName: the argument must be a string");
     fi;
-    if ForAny(name, l -> not l in IdentifierLetters) then
-        Info(InfoWarning + InfoGlobal, 2, 
-             "suspicious global variable name ", name);
+    pos := 1;
+    # This function is used early in the initialization process, so we need to
+    # do PositionProperty manually.
+    while pos <= Length(name) do
+        if name[pos] in IdentifierLetters then
+            pos := pos + 1;
+        else
+            break;
+        fi;
+    od;
+    if pos <= Length(name) then
+        Info(InfoWarning + InfoGlobal, 1,
+             "suspicious global variable name \"", name, "\", ",
+             "non-identifier character found at position ", pos);
+    fi;
+    if Length(name) = 0 then
+        Info(InfoWarning + InfoGlobal, 1,
+             "suspicious global variable name \"", name, "\", ",
+             "name is the empty string");
     fi;
 end;
 
--- a/tst/testinstall/global.tst
+++ b/tst/testinstall/global.tst
@@ -1,0 +1,6 @@
+# test CheckGlobalName
+gap> CheckGlobalName(" ");
+#I  suspicious global variable name " ", non-identifier character found at position 1
+gap> CheckGlobalName("");
+#I  suspicious global variable name "", name is the empty string
+gap> CheckGlobalName("name");


ChrisJefferson:
I'm happy with this warning. For information, such variables can be accessed, by escaping the space, but I can't imagine a good use for this, other than perhaps a cunning way of hiding private data.

```
gap> BIND_GLOBAL(" x", 3);
3
gap> Print(\ x);
3
```

ssiccha:
@ChrisJefferson Ah, thanks. :)

I just added a test.

stevelinton:
There are plenty of other strange variable names which can be used. For instance:
```
gap> foo := function() \2 := 1; Print(\2," ",2,"\n"); end;
function(  ) ... end
gap> Print(foo);                                          
function (  )
    2 := 1;
    Print( 2, " ", 2, "\n" );
    return;
end
gap> foo();
1 2
gap> 
```

While we're adding warnings, are there any others we should add?

ssiccha:
> While we're adding warnings, are there any others we should add?

Hmm, maybe there should be a warning if the first character is not an alpha-numerical character? Or is that too strict? Is there a way with e.g. jupyter to bind to unicode characters like one can do in julia? (I really have no idea)

I've just been thinking about whether `BIND_GLOBAL` might be called in any code that needs to be performant. I guess that's not the case, right? If we would go for warnings with non alpha-numerical characters as first characters we _could_ also check the whole string for such weird characters.

> Actually, does the above show a more serious issue with printing functions, which we should actually fix?

If someone happens to print such a function with a strange variable name it might be super confusing. So if it's easy to fix let's do it, otherwise not since it's a really weird corner case.

fingolfin:
Two remarks:

1. @ssiccha and @stevelinton Please create branches in your personal fork of the gap repository, not here in `gap-system/gap` (For Steve, I am talking about branch `func-auts` in PR #3901). 

2. `BIND_GLOBAL` is a low-level function. I don't think it should perform this kind of check. Rather, `BindGlobal` should perform such checks -- and indeed it does perform them, and produces an `Info` message if either `InfoGlobal` or `InfoWarning` is set to level 2. I suppose we could lower this to 1, then the message would be shown by default. Also, one might want to tweak the message slightly: put quotes around the variable name when printing it, to make things like a leading whitespace more visible, and/or be more specific about what exactly is suspicious (e.g. "non-identifier character ' ' found at position 1")
```
gap> BindGlobal(" x", 1);
#I  suspicious global variable name  x
```

lib/global.g
```diff
@@ -133,6 +133,10 @@ MAKE_READ_ONLY_GLOBAL("REREADING");
 ##
   
 BIND_GLOBAL := function( name, val)
+    if IsBound(name[1]) and name[1] in [' ', '\t', '\r', '\n'] then
+        Print("#W BIND_GLOBAL: variable starts with a whitespace and is thus",
+              " not accessible\n");
```
fingolfin: That message is inaccurate. The variable *is* accessible:
```
gap> BindGlobal(" x", 1);
#I  suspicious global variable name  x
gap> ValueGlobal(" x");
1
gap> \ x;
1
```

stevelinton:
@fingolfin Sorry aboput putting the branch in the wrong place -- typo.

ssiccha:
@fingolfin 
> Please create branches in your personal fork of the gap repository

Oops, will do in the future.

I've updated the PR. Now the `tst/testinstall/syntaxtree.tst` fails:
```
########> Diff in /home/sergio/projects/gap/tst/testinstall/syntaxtree.tst:23
# Input is:
for n in NamesGVars() do
       if IsBoundGlobal(n) and not IsAutoGlobal(n) then
           v := ValueGlobal(n);
           if IsFunction(v) and not IsKernelFunction(v) then
               if not test_tree(v) then
                  Print("failed round trip: ",n,"\n");
               fi;
           elif IsOperation(v) then
               for i in [1..6] do
                   for x in METHODS_OPERATION(v, i) do
                       if IsFunction(x) and not IsKernelFunction(v) then
                           if not test_tree(v) then
                               Print("METHODS_OPERATION(", n, ",", i, ") faile\
d round trip\n");
                           fi;
                       fi;
                   od; 
               od;
           fi;
       fi;
od;;
# Expected output:
# But found:
#I  suspicious global variable name "*", non-identifier character found at pos\
ition 1
#I  suspicious global variable name "*", non-identifier character found at pos\
ition 1
#I  suspicious global variable name "+", non-identifier character found at pos\
ition 1
...
```

This is turning out to be a lot more complicated than I thought so I'm gonna stop for now. Does someone have an idea whether the above can be fixed easily?

Also, originally I wanted to include `CheckGlobalName` into `DeclareGlobalVariable` and `DeclareGlobalFunction`. But these functions seem to be loaded so early in the initialization process that I can't even add
```
if IsBound(CheckGlobalName) then
    CheckGlobalName(name);
fi;
```
to `DeclareGlobalFunction` without getting an error:
```
failed to open error stream
Syntax warning: Unbound global variable in /home/sergio/projects/gap/lib/oper.g:19\
51
        CheckGlobalName( name );
        ^^^^^^^^^^^^^^^
```

Also, I'll reopen this PR with the branch from my fork once the discussion here is somewhat finished.

fingolfin:
Well, there simply *are* global symbols with names like "*", "\\in", "+" etc. -- they are actively used. Rejecting them is not an option...

fingolfin:
Taking a step back from implementation details, perhaps let's start by discussing what the *motivation* for this PR is. What are the problems you want to combat? Somebody inserting a space into a string by accident? OK, fine, this now you catch. But you don't catch if I e.g. swap to letters or make a typo, or use uppercase instead of lowercase or vice versa, or.... So why is catching an extra whitespace so important?

ssiccha:
@fingolfin Originally, I was testing a `sed` script which accidentally turned code of the form `BindGlobal("X")` into `BindGlobal(" Y")` instead of `BindGlobal("Y")`. I don't recall exactly anymore but I think it took me at least 20 minutes to figure out why everything was breaking. When I first checked the files I think my brain chose to ignore the extra whitespace.

Then I thought, it should be trivial to add a corresponding test to `BindGlobal` which checks for a leading whitespace. So that's where this PR comes from.

But I think we shouldn't add such a check only to `BindGlobal` and not have it in `DeclareGlobalVariable` and `DeclareGlobalFunction`. That to me feels too inconsistent and thus also confusing.

ssiccha:
This is not so important IMO so I'll close this PR.

PR #3907, opened 2020-02-29
Do not put a link to HPC-GAP manual

alex-konovalov:
It is only available if GAP is compiled with HPC-GAP support,
otherwise the cross-reference is unresolved.


alex-konovalov:
Will merge this now and trigger release build again.

alex-konovalov:
Backported to stable-4.11 in 0cf2f1a68786d0b6667a2ecf4fd782006c70d65e

PR #3912, opened 2020-03-02
Update GMP to version 6.2.0

fingolfin:
We had 6.1.2 before, released in December 2016. This new version was released January 2020.


stevelinton:
There may be a problem with Macs. The GMP web page says:

```
Issues with GMP 6.2.0:
MacOS Xcode 11 prior to 11.3 miscompiles GMP, leading to crashes and miscomputation.
Issues with GMP 6.1.2:
MacOS Xcode 11 prior to 11.3 miscompiles GMP, leading to crashes and miscomputation.
```
The affected versions of Xcode came out between Sep and Dec 2019. Anyone who got a new Mac or upgraded their OS in that interval and hasn't install the (inevitably huge) upgrade to Xcode will be affected.

fingolfin:
But as you write, this affects both version 6.1.2 which we are using right now, and also the new version 6.2.0. So this problem is orthogonal to this PR, isn't it?

Somebody might still want to implement a change to our buildsystem to check for the problematic Xcode versions...

stevelinton:
@fingolfin Fair point. The bug may well be in parts of GMP we don't use.

PR #3914, opened 2020-03-05
Add InnerAutomorphismGroup for groups

fingolfin:
If the automorphism group is known, calls InnerAutomorphismsAutomorphismGroup;
otherwise creates the InnerAutomorphismGroup from scratch.

I suspect that there is a reason `InnerAutomorphismGroup` does not exist right now (i.e. this PR might be too naive); hopefully @hulpke knows about that. So please don't merge this before he had a chance to comment.

### DIFF
--- a/lib/morpheus.gi
+++ b/lib/morpheus.gi
@@ -2589,6 +2589,28 @@ InstallMethod( InnerAutomorphismsAutomorphismGroup,
 
 #############################################################################
 ##
+#M  InnerAutomorphismGroup( <G> )
+##
+InstallMethod( InnerAutomorphismGroup,
+    "for groups",
+    true,
+    [ IsGroup and HasGeneratorsOfGroup ], 0,
+    function( G )
+    local gens, A;
+    if HasAutomorphismGroup( G ) or IsTrivial( G ) then
+        return InnerAutomorphismsAutomorphismGroup( AutomorphismGroup( G ) );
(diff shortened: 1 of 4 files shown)


lib/morpheus.gi
```diff
@@ -2587,6 +2587,26 @@ InstallMethod( InnerAutomorphismsAutomorphismGroup,
...
+    [ IsGroup and HasGeneratorsOfGroup ], 0,
+    function( G )
+    local gens;
+    if HasAutomorphismGroup( G ) or IsTrivial( G ) then
+        return InnerAutomorphismsAutomorphismGroup( AutomorphismGroup( G ) );
+    fi;
+    gens:= GeneratorsOfGroup( G );
+    # get the non-central generators
+    gens:= Filtered( gens, i -> not ForAll( gens, j -> i*j = j*i ) );
+    return Group( List( gens, i -> InnerAutomorphism( G, i ) ) );
```
hulpke: You want to call
`SetIsGroupOfAutomorphismsFiniteGroup(autgrp,true)` on this group, so GAP knows that this is a group of group automorphisms, and thus uses appropriate `Nice` methods to work with the group. (This is a pain to do, but as far as I know there is no way to deduce this information automatically from the type.)

fingolfin: Thanks, done!

PR #3916, opened 2020-03-11
Enhancements to creating and using rewriting systems.

hulpke:
This will speed up some calls, enhance the capabilities of computing 2-cohomology, and allow for cleaner interfacing.

There is some reorganization between functions/files that the `diff` is not clever enough to recognize, which makes the change set look larger than it really is.

### DIFF
--- a/lib/gpfpiso.gi
+++ b/lib/gpfpiso.gi
@@ -245,7 +245,7 @@ function(g,str,N)
   local ser, ab, homs, gens, idx, start, pcgs, hom, f, fgens, auts, sf, orb,
   tra, j, a, ad, lad, n, fg, free, rels, fp, vals, dec, still, lgens, ngens,
   nrels, nvals, p, dodecomp, decomp, hogens, di, i, k, l,
-  m,abelianlimit,locallim,abpow,needgens;
+  m,abelianlimit,locallim,abpow,needgens,fampcgs;
 
   abelianlimit:=ValueOption("abelianlimit");
   if abelianlimit=fail then
@@ -270,6 +270,11 @@ function(g,str,N)
     fi;
     ser:=Filtered(ser,i->IsSubset(i,N));
   fi;
+  if IsSolvableGroup(g) then
+    fampcgs:=SpecialPcgs(g);
+  else
+    fampcgs:=fail;
+  fi;
   ab:=[];
   homs:=[];
   gens:=[];
@@ -279,7 +284,12 @@ function(g,str,N)
     start:=Length(gens)+1;
     if HasAbelianFactorGroup(ser[i-1],ser[i]) then
       ab[i-1]:=true;
-      pcgs:=ModuloPcgs(ser[i-1],ser[i]);
+      if fampcgs<>fail then
+        pcgs:=CanonicalPcgs(InducedPcgs(fampcgs,ser[i-1])) mod
+        CanonicalPcgs(InducedPcgs(fampcgs,ser[i]));
+      else
+        pcgs:=ModuloPcgs(ser[i-1],ser[i]);
+      fi;
       homs[i-1]:=pcgs;
       Append(gens,pcgs);
       f:=pcgs;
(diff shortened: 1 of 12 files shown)


lib/gpfpiso.gi
```diff
@@ -888,5 +964,992 @@ function( G, series )
...
+      iso:=GroupHomomorphismByImagesNC(G,H,P,GeneratorsOfGroup(H));
+    fi;
+    return iso;
+  end;
+
+  # try to identify Weyl series
+  n:=0;
+  repeat
+    n:=n+1;
+    fn:=Factorial(n);
```
fingolfin: It probably doesn't matter for any group size we see in practice, but you could also init `fn:=1` before the loop and then write here:
```suggestion
    fn:=fn*n;
```

lib/gpfpiso.gi
```diff
@@ -888,5 +964,992 @@ function( G, series )
...
+    bigcount:=0;
+    sz:=fn*(n+1);
+    if Size(G)<sz then bigcount:=bigcount+1;
+    elif Size(G)=sz and IsSymmetricGroup(G) then
+      # try An
+      if HasIsomorphismFpGroup(G) then
+        G:=Group(GeneratorsOfGroup(G));
+        SetSize(G,sz);
+        IsSymmetricGroup(G);
```
fingolfin: Perhaps this?
```suggestion
        SetIsSymmetricGroup(G,true);
```

lib/gpfpiso.gi
```diff
@@ -888,5 +964,992 @@ function( G, series )
...
+  repeat
+    n:=n+1;
+    fn:=Factorial(n);
+    bigcount:=0;
+    sz:=fn*(n+1);
+    if Size(G)<sz then bigcount:=bigcount+1;
+    elif Size(G)=sz and IsSymmetricGroup(G) then
+      # try An
+      if HasIsomorphismFpGroup(G) then
+        G:=Group(GeneratorsOfGroup(G));
```
fingolfin: Why discard all info about `G`? Just in case it already has an `IsomorphismFpGroup` stored? Perhaps a comment could explain this?

lib/gpfpiso.gi
```diff
@@ -888,5 +964,992 @@ function( G, series )
...
+    if iso<>fail then return iso;fi;
+  elif Size(G)=192*Factorial(10) then
+    iso:=tryweyl("E",8);
+    if iso<>fail then return iso;fi;
+  fi;
+
+  return fail;
+end);
+
+# rewriting systems for simple groups based on BN pairs, following Schmidt
```
fingolfin: Can you give a more explicit reference than just "following Schmidt"? Like, is there a paper or a thesis, or something?

I may be missing something, but, aren't these "standard" ideas when working with BN-groups, and present in e.g. Cohen-Murray-Taylor?

lib/gpfpiso.gi
```diff
@@ -888,5 +964,992 @@ function( G, series )
...
+BindGlobal("BNnyRewritingPresentation",function(group,borel,weyl)
+local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
+    lev,ord,monb,mons,gp,trawo,hom,tst,dc,dcreps,act,decomp,ranb,ranw,
+    nofob,nofow,reduce,pcgs,can,pri,stb,addrule,invmap,jj,wo;
+
+  if Size(ClosureGroup(borel,weyl))<Size(group) then return fail;fi;
```
fingolfin: I don't know how expensive the above is in practical examples you want to tackle; one could in principle have a possibly faster (assuming the sizes are already known) pre-test:
```
if Size(borel)^2*Size(weyl)<Size(group) then return fail;fi;
```

hulpke: This is if it is a proper Borel. But I want to use the code also for some sporadic groups in which borel and weyl are just some subngroups, abusing the notation. (The resulting RWS are still better than with random generators)

lib/gpfpiso.gi
hulpke: Done

lib/gpfpiso.gi
hulpke: Added remark

lib/gpfpiso.gi
hulpke: It is basically Schmidt's only paper (and his thesis). But I see the point that there might be another person names Schmidt.
Added a referecne.

Issue #3624, opened 2019-08-26
Parallelise BuildPackages.sh


stevelinton:
Please use the following template to submit an issue
(you may delete lines which are not used). Thank You!

### Observed behaviour

BuildPackages takes ages, even on a machine with fast disks and lots of cores.

### Expected behaviour

It would be nice if it was quicker. It could use make -j on packages for which ithat works, and/or build multiple packages in parallel.

fingolfin:
(General caveat: I am not sure how far I want to go in extending this script which always was meant as an interim hack until we have something better. On the other hand, it's not clear we ever will get something better. Well, perhaps https://github.com/gap-packages/PackageManager will get there eventually).

A simple way to get this to happen is to use the `MAKEFLAGS` environment variable, like so: `MAKEFLAGS=-j8 ../bin/BuildPackages.sh`. But that will "only" parallelize within a single build (and only if it uses `make`). Another orthogonal approach is to indeed build multiple packages in parallel. On systems with `xargs` installed (so for most Linux distros and macOS), one can do e.g. this (for 8 parallel jobs):  `ls | xargs -n1 -P8 ../bin/BuildPackages.sh` (this is a bit of hack, though, instead of `ls` it would be better to generate a list of package directory names; but it works OK enough). The big advantage of that is that it also parallelizes `configure` runs, which can take up some time (this effect is much more prominent on macOS, and even more so on Windows, due to their very slow process startup times, which dominates the performance of GNU `configure` scripts).

Here are some timings from an somewhat older Ubuntu quad-core server (i7-3820 CPU @ 3.60GHz). For a quadcore, usually 8 jobs is a good compromise that prevents IO and/or memory saturation, for other machines one may of course want to use different parameters.

Without any parallelization:
```
$ time ../bin/BuildPackages.sh
real	6m32.992s
user	5m23.264s
sys	0m12.540s
```

With `MAKEFLAGS=-j8`
```
$ time MAKEFLAGS=-j8 ../bin/BuildPackages.sh
real	4m40.908s
user	5m12.872s
sys	0m11.308s
```

With `xargs -P8`:
```
$ time ls | xargs -n1 -P8 ../bin/BuildPackages.sh
real	3m28.195s
user	5m16.572s
sys	0m12.852s
```

Of course one can also combine the two techniques, but I'd recommend to not oversaturate the computer But e.g. using `MAKEFLAGS=-j3` and `xargs -P3` might be a good compromise. Let's try that
```
[...]

PR #3922, opened 2020-03-24
Add --parallel to BuildPackages.sh

ChrisJefferson:
This PR adds a --parallel to BuildPackages.sh. It just builds all packages in parallel at the same time.

It would be better to build packages with a more configurable limit, but that would add a lot more code, and probably require some extra dependancy which not all OSes would have (gnu parallel has to be installed, many machines don't have an xargs with -P yet).

On my machine, this speeds up running BuildPackages.sh from 3 minutes 49 seconds to 58 seconds.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -34,6 +34,7 @@ CURDIR="$(pwd)"
 GAPROOT="$(cd .. && pwd)"
 COLORS=yes
 STRICT=no       # exit with non-zero exit code when encountering any failures
+PARALLEL=no
 PACKAGES=()
 
 # If output does not go into a terminal (but rather into a log file),
@@ -45,6 +46,7 @@ while [[ "$#" -ge 1 ]]; do
   case "$option" in
     --with-gaproot)   GAPROOT="$1"; shift ;;
     --with-gaproot=*) GAPROOT=${option#--with-gaproot=}; ;;
+    --parallel)       PARALLEL=yes; ;;
 
     --no-color)       COLORS=no ;;
     --color)          COLORS=yes ;;
@@ -57,6 +59,10 @@ while [[ "$#" -ge 1 ]]; do
     *)                PACKAGES+=("$option") ;;
   esac
 done
+
+if [ "x$PARALLEL" = "xyes" ]; then
+  export MAKEFLAGS="${MAKEFLAGS:--j3}"
+fi;
 
 # If user specified no packages to build, build all packages in subdirectories.
 if [[ ${#PACKAGES[@]} == 0 ]]
@@ -241,7 +247,8 @@ build_one_package() {
 
 date >> "$LOGDIR/fail.log"
 for PKG in "${PACKAGES[@]}"
-do
+do 
+  ( 
   # cut off the ending slash (if exists)
   PKG="${PKG%/}"
   # cut off everything before the first slash to only keep the package name
@@ -278,7 +285,19 @@ do
     echo
     warning "$PKG does not seem to be a package directory, skipping"
   fi
+  ) &
+  if [ "x$PARALLEL" != "xyes" ]; then
+    # If more than 4 background jobs are running, wait for one to finish
+    if [[ $(jobs -r -p | wc -l) -gt 4 ]]; then
+        wait -n
+    fi
+  else
+    # wait for this package to finish building
+    wait
+  fi;
 done
+# Wait until all packages are built, if in parallel
+wait
 
 echo "" >> "$LOGDIR/fail.log"
 echo ""


ChrisJefferson:
@wucas pointed out this is a brute-force approach to fix https://github.com/gap-system/gap/issues/3624

ChrisJefferson:
Now using some parts of #3624 -- adding a MAKEFLAGS, and also running no more than 4 packages in parallel (these could be made configurable).

The parallelisation limiter uses `jobs` -- something like xargs or gnu parallel would be better, but that would be something else to require.

wucas:
I tried it on an older Thinkpad I have laying around (latest Arch Linux, i5-2520M 2core 2.5GHz, 8GB RAM) and it worked finde there (nothing crashes or explodes). Only I noticed that `CaratInterface`, `ZeroMQInterface-0.12`, `CddInterface-2020.01.01`, `curlInterface-2.1.1` and `NormalizInterface-1.1.0` failed. With a normal GAP 4.11.0 install only `CddInterface-2020.01.01` and `NormalizInterface-1.1.0` failed. But I just copied the `pkg` folder from the GAP 4.11.0 I had so not sure if thats the right move there. 
I also tried it on my MacBook Pro 2018 and there it worked as well. There `simpcomp
NormalizInterface-1.0.2
xgap-4.29
anupq-3.2
profiling-2.0.1` but those fail with the standard script as well.

wucas:
I can try it on a Raspberry Pi 3 too later if that helps.

ChrisJefferson:
Thanks, that sounds promising! I think I've adjusted it so it shouldn't go crazy and kill things now.

wucas:
I've tried it on my raspi. Works as well as in the raps still lives. With `--parallel` the following packages fail: 
```
json-2.0.1
semigroups-3.2.3
anupq-3.2.1
simpcomp
CddInterface-2020.01.01
nq-2.5.4
DeepThought-1.0.2
io-4.7.0
profiling-2.2.1
float-0.9.1
digraphs-1.1.1
ferret-1.0.2
ZeroMQInterface-0.12
crypting-0.10
Browse
EDIM-1.3.5
Gauss-2019.09.02
curlInterface-2.1.1
datastructures-0.2.5
orb-4.8.3
cvec-2.7.4
NormalizInterface-1.1.0
xgap-4.30
```

And with the standard 4.11.0 install only 
```
profiling-2.2.1
xgap-4.30
nq-2.5.4
ferret-1.0.2
ZeroMQInterface-0.12
curlInterface-2.1.1
io-4.7.0
simpcomp
float-0.9.1
digraphs-1.1.1
CddInterface-2020.01.01
json-2.0.1
NormalizInterface-1.1.0
semigroups-3.2.3
anupq-3.2.1
DeepThought-1.0.2
```
 fail.

fingolfin:
@ChrisJefferson maybe take a look at https://github.com/gap-system/gap/issues/3624 for some inspiration -- in particular, I found that for me, a tiny number of packages dominate the overall build time, and it therefore helps to not just build multiple packages in parallel, but also to let packages use more than one CPU. The trick bit is to coordinate this.

That said, I really am sad that we still have to put lipstick on this pig of a script. I hope we can soon work towards moving this functionality into PackageManager, and improving it there, and using that to install and build packages... and perhaps even eventually discard the monolithic package distribution...

fingolfin:
In particular, one can use `make -jNUM` to parallelize things -- we already expect people to have GNU make to build GAP and many packages, after all.

ChrisJefferson:
Sorry, didn't push an update. Now I use MAKEFLAGS, and do a slightly cheap thing to limit parallelisation too.

As you say, it would be nice to replace this altogether.

ChrisJefferson:
I'm happy with this as is. It could obviously be better, but it's (for me at least) better for now.

ssiccha:
Ok. Then someone (@fingolfin ?) needs to press the button since I'm not authorized to merge things. :)

fingolfin:
Ah I just saw the most recent comment by @ChrisJefferson somehow missed it before. Will merge it then.

Of course one could make the job numbers (3 for make and 4 globally) adjustable, but for now this seem fine.

fingolfin:
Backported to stable-4.11 in commit 0b8e4398417ee09fe28119dce7da0f9348aa5bf0

bin/BuildPackages.sh
```diff
@@ -278,7 +285,19 @@ do
     echo
     warning "$PKG does not seem to be a package directory, skipping"
   fi
+  ) &
+  if [ "x$PARALLEL" != "xyes" ]; then
```
fingolfin, line 289: errr, wait, this check is the wrong way around, isn't it?

PR #3923, opened 2020-03-24
Make curl follow redirects, fixes 'make bootstrap-pkg-*'

ChrisJefferson:
When rearranging files on www.gap-system.org, we have introduced some 301 redirects. Curl needs an explicit flag (-L) to make it follow these redirects. wget follows them by default.


ChrisJefferson:
Updated links, will just check if if tests cycles through correctly.

fingolfin:
Backported to stable-4.11 in 341314c6cdb84969eb8300a91fad0592dcb67a56 and a71443d983a48e93a4dcc7356fccc723faa2b4f8

PR #3925, opened 2020-03-25
Add the missing perfect groups of order up to a million

hulpke:
Added perfect groups of orders that had been missing in the Holt/Plesken book (newly computed). This increases the number of groups in the perfect groups library by a factor close to 6.

Also added 5 groups that were found missing in the existing lists.


fingolfin:
This needs to be rebased on latest master: package bootstrap was temporarily broken because we moved file downloads to its own subdomain; now, of course we also set up a redirect, but it turns out we didn't configure `curl` to follow redirects... This was quickly fixed on master, but now this PR needs to be rebased to get the fix.

Sorry for the inconvenience. Also, 10% or so of Travis builds currently randomly fail because the downloads from gap-system.org alternate between blazing fast (5 seconds to download package tarball) to incredible slow (> 40 minutes do download). We are working during the ongoing virtual reality GAP Days on addressing this in some way.

grp/perf.gd
```diff
@@ -129,28 +129,10 @@ DeclareGlobalFunction("SizesPerfectGroups");
...
 DeclareSynonym("NrPerfectGroups",NumberPerfectGroups);
+DeclareSynonym("NumberPerfectLibraryGroups",NumberPerfectGroups);
+DeclareSynonym("NrPerfectLibraryGroups",NumberPerfectGroups);
 
 
-#############################################################################
-##
-#F  NumberPerfectLibraryGroups( <size> )  . . . . . . . . . . . . . . . . . .
-##
-##  <#GAPDoc Label="NumberPerfectLibraryGroups">
```
fingolfin: Note that the reference manual still references these, so you need to update `doc/ref/grplib.xml` if you remove this section.

However, instead of outright dropping these from the manual, I think it would be nice if we kept at least a minimal reference to them, something like this:
> `NumberPerfectLibraryGroups` and `NrPerfectLibraryGroups` are synonyms for `NumberPerfectGroups` which only exist for historical reasons.
That could be part of the `NumberPerfectGroups` manual entry, and we could add suitable GAPDoc commands to ensure `NrPerfectLibraryGroups` and `NumberPerfectLibraryGroups` are still in the index and point to that manual entry.

hulpke: Thank you -- yes, I will do so (and add other documentation)

hulpke:
> Sorry for the inconvenience. Also, 10% or so of Travis builds currently randomly fail because the downloads from gap-system.org alternate between blazing fast (5 seconds to download package tarball) to incredible slow (> 40 minutes do download). We are working during the ongoing virtual reality GAP Days on addressing this in some way.

Thank you for the explanation. I observed this weird behavior and was wondering whether anyone was aware of it. I suspect you have good ideas what to do about it and won't need me to second-guess possible workarounds.

fingolfin:
At the core, it seems that the VM the University of St Andrews gives us to host the GAP web server is flaky (for many months now, it has been offline for a few seconds or minutes, multiple times per day); and recently, the bandwidth breaks down frequently, too. I have no idea what exactly is the cause, but in the end, it doesn't matter.

Right now, we are considering to just move the "default" downloads to the GitHub release system, which has been very reliable for us so far, and use the St Andrews only as backup. In addition, I may be able to offer hosting in Kaiserslautern, but maybe only after the Corona pandemic, as right now it is difficult to coordinate this with the computer admins.

doc/ref/grplib.xml
```diff
@@ -297,42 +297,30 @@ It returns <K>fail</K> if no such group exists in the library.
...
 be noted that  their book contains a  lot of further information for many
 of the library groups.  So we would like  to recommend  it to any  &GAP;
 user who is interested in the groups.
+Two additional groups omitted from the book <Q>Perfect Groups</Q> have also
+been included.
+<P/>
+The library of these has been brought into &GAP; format by Volkmar Felsch.
+<P/>
+The groups of the orders listed above have been enumerated by Alexander
+Hulpke. They are stired directly and provide less construction information
```
fingolfin, line 321:
```suggestion
Hulpke. They are stored directly and provide less construction information
```

doc/ref/grplib.xml
```diff
@@ -297,42 +297,30 @@ It returns <K>fail</K> if no such group exists in the library.
...
 available to the &GAP; community  by contributing their files. It should
 be noted that  their book contains a  lot of further information for many
 of the library groups.  So we would like  to recommend  it to any  &GAP;
 user who is interested in the groups.
+Two additional groups omitted from the book <Q>Perfect Groups</Q> have also
+been included.
+<P/>
+The library of these has been brought into &GAP; format by Volkmar Felsch.
+<P/>
+The groups of the orders listed above have been enumerated by Alexander
```
fingolfin, line 320: I am not sure what is meant here with "the orders". Does this refer to the groups which were missing from the work by Holt and Plesken? Or *all* groups (i.e. you recomputed all)? Or something else? If is the first case, how about this to clarify it:
```suggestion
The perfect groups of size less than <M>10^6</M>which were missing in the work of Holt
and Plesken have been enumerated by Alexander
```

grp/perf.gd
```diff
@@ -116,41 +116,29 @@ DeclareGlobalFunction("SizesPerfectGroups");
...
 ##
 ##  <Description>
 ##  returns the number of non-isomorphic perfect groups of size <A>size</A> for
 ##  each positive integer  <A>size</A> up to <M>10^6</M> except for the eight  sizes
 ##  listed at the beginning  of  this section for  which the number is not
 ##  yet known. For these values as well as for any argument out of range it
 ##  returns <K>fail</K>.
+##  <A>NumberPerfectLibraryGroups</A> and <A>NrPerfectLibraryGroups</A> 
+##  are synonyms for NumberPerfectGroups which only exist for historical
+##  reasons.
```
fingolfin, line 131:
```suggestion
##  <A>NrPerfectGroups</A> is a synonym for <Ref Func="NumberPerfectGroups"/>.
##  Moreover <A>NumberPerfectLibraryGroups</A> and <A>NrPerfectLibraryGroups</A> 
##  are synonyms for <Ref Func="NumberPerfectGroups"/> which only exist for historical
##  reasons.
```

fingolfin:
@hulpke this still has the "do not merge" label; on purpose?

hulpke:
Yes, the "Do not merge" is since I still want to run correctness tests.

doc/ref/grplib.xml
```diff
@@ -297,42 +297,31 @@ It returns <K>fail</K> if no such group exists in the library.
...
+<M>n = 61440</M>, 122880, 172032, 245760, 344064, 491520,
+688128, or 983040 were omitted.
 <P/>
 We are grateful to Derek Holt and Wilhelm Plesken for making their groups
 available to the &GAP; community  by contributing their files. It should
 be noted that  their book contains a  lot of further information for many
 of the library groups.  So we would like  to recommend  it to any  &GAP;
 user who is interested in the groups.
+Two additional groups omitted from the book <Q>Perfect Groups</Q> have also
+been included.
```
ThomasBreuer, line 316: @hulpke My understanding of this statement is that including these two groups fixes a bug in the book.
In this case, it would be very interesting to state which groups had been missing (group order, structure).
Note that there may be published results based on searches in the library of perfect groups, which rely on the nonexistence of groups with certain properties.

hulpke: @ThomasBreuer This fix was there before me, the text might show up because the whole section needed some reformatting. I'll see whether I can find out what was added when.

ThomasBreuer: @hulpke Thanks for the clarification, and thanks for trying to find a precise description of the fix.
It would have been good if the documentation of the library would have stated this earlier,
but the current extension of the library makes it even more relevant to say what the differences are and when they arose.

hulpke: I seem to remember that Jack Schmidt found tghese groups, probably some time between 2006 and 2012, but I can't find anything about it in my email. I think @fingolfin has a copy of the old commit history that is more complete. Is there any commit to `perf0.g` durring this period that changes numbers?

fingolfin:
These seem to be the relevant commits:
```
commit db7a8df486e0ca4a1f59b351f7340b4927be1008
Author: Jack Schmidt <jack@ms.uky.edu>
Date:   2005-11-24 18:51:28 +0000

    Added missing perfect group of order 450000, A6 2^1 # 5^4. Restricts nicely to A5 2^1 # 5^2, which is its minimal index corefree subgroup. JS
```
and
```
commit 2d5f94f342bd412cff4b95fb33fe3fd599a73e67
Author: Jack Schmidt <jack@ms.uky.edu>
Date:   2005-09-10 17:53:48 +0000

    First draft of adding in the perfect group of order 962280. JS
```

hulpke:
@fingolfin Perfect (pun intened). I'll add a remark.
[So it was 2005, which actually predates my current email archive...]

ThomasBreuer:
@hulpke One more minor remark:
The changed behaviour of `NumberPerfectLibraryGroups` may break existing code, in the sense that the old variant allows one to run over the available perfect groups of order `n` via
```
List( [ 1 .. NumberPerfectLibraryGroups( n ) ], i -> PerfectGroup( IsPermGroup, i );
```
whereas the new variant forces one to check for `fail` before forming the range;
try for example
```
[ 1 .. NumberPerfectLibraryGroups( 1451520 ) ];
```

hulpke:
@ThomasBreuer 
I look at this as more of a hypothetical issue, but would be happy to reconsider if someone can point to concrete code out there that does this.

The issue of "available perfect groups" was much more a concern with the holes in the library. OTOH, the old convention of `NumberPerfectLibraryGroups` has the risk of a user misinterpreting it as `NumberPerfectGroups` and making wrong conclusions about existence of such groups.

The new behavior also is more consistent with the other group libraries, none of which distinguishes "Groups" from "LibraryGroups".

What probably however should be done with this change is to move `NumberPerfectLibraryGroups` on the way to obsolescence, e.g. have it issue a warning that points to `NrPerfectGroups`.

ThomasBreuer:
@hulpke Well, this is how I found the hypothetical problem with the changed behaviour of `NumberPerfectLibraryGroups`:
I wanted to include the IDs of the relevant new groups from the perfect groups library in the character table library, and ran into an error (see `ctbllib/gap4/ctdbattr.g`, attribute `"perf"`, function `create`).

fingolfin:
Great job! That's definitely going to be a top tier in the release notes for 4.12 ;-)

PR #3928, opened 2020-03-26
add libgap versioning, and use it to set GAP kernel version, i.e. gap_kernel_major/minor_version

dimpase:
as discussed on Slack. Note that automatically divining cur:rev:age seems to be pretty much impossible, so it's all manual increment.


ChrisJefferson:
I don't really like the idea of a third version type (we already have kernel major version and minor version, and GAP version).

The way libtool does versioning does seem annoying. We could just make the version kernelmajorversion*100+kernelminorversion, and the age kernelminorversion. That would lead to big versions of course.

dimpase:
This would throw away the compatibility window idea provided by "age", no?

dimpase:
assuming  the kernel getting re-implemented on top of libgap, this number of different versions would go down...

ChrisJefferson:
No, when we increment kernelminorversion, it's backwards compatable (hopefully I've set the "age" in the right way), and when we increment kernelmajorversion, the kernel isn't backwards compatible.

It could be that sometimes we break backwards compatability for kernel extensions, but not libgap, but I don't think we want to try to seperate those two concepts.

ChrisJefferson:
I'm not aware of any plans to reimplement gap on top of libgap. Of course if someone wanted to put the work in, and it didn't make GAP worse for existing GAP users and developers, I wouldn't have a problem with it

fingolfin:
Chris, there are good reasons why the libtool versioning works like that, it makes it possible to support different OSes which have very different ways of versioning shared libraries (think: Linux, macOS, *BSD, Windows -- they all do this differently).

Still, i also don't want "yet another version". Which is why I explained on the mailing list and again on slack that the kernel version should then be derived from the libtool version, as follows:
- kernel_major = libtool_current - libtool_age
- kernel_minor = libtool_age
There is no need for libtool_revision, it can just stay 0 forever.

ChrisJefferson:
@fingolfin : That looks sensible. I don't mind if the libtool or the kernelversion are the "official" one, and we derive the other -- I think we have the same derivation in both directions

dimpase:
OK, I'll implement this.

ChrisJefferson:
Just for when you hit it, in Makefile.rules there is the line ` cp .libs/cyggap-0.dll $@  # FIXME: HACK to support kernel extensions` which is going to break. Could I suggest just changing it to `.libs/cyggap-*.dll` (unless it is easy to put the right version in here). This is just a hack for cygwin. Technically this will break if someone builds to versions in the same cygwin, but I imagine that's very uncommon.

dimpase:
> Could I suggest just changing it to `.libs/cyggap-*.dll` (unless it is easy to put the right version in here).

well, I'll do this, but I have my doubts - e.g. Sagemath builds libgap on Cygwin, and this cyggap-0.dll
sees to be a part of GAP executable, i.e. "another" libgap.

I'd appreciate this tested on Cygwin by running `make install-libgap`...

fingolfin:
We test Cygwin with each PR on AppVeyor CI, and that passed.

configure.ac
```diff
@@ -735,6 +736,15 @@ AC_SUBST([gap_kernel_minor_version])
 AC_DEFINE_UNQUOTED([GAP_KERNEL_MAJOR_VERSION], $gap_kernel_major_version, [The major version of the kernel ABI])
 AC_DEFINE_UNQUOTED([GAP_KERNEL_MINOR_VERSION], $gap_kernel_minor_version, [The minor version of the kernel ABI])
 
+dnl libgap_current/revision/age must be set using the GNU Libtool rules at
+dnl https://www.gnu.org/software/libtool/manual/libtool.html#Libtool-versioning
+dnl which uses current:revision:age format.
+dnl Set it up as discussed on GAP's PR 3928
+libgap_current=$(( gap_kernel_major_version + gap_kernel_minor_version ))
```
fingolfin: Nope, that won't do it. With that setup,  kernel  3.1 and 4.0 would have the same `current` value.

It really has to be as I described it, not the other way around.

dimpase: It is the way you described it, execept that the libtool numbers are expressed in kernel numbers,
and not the other way around. An invertible linear transformation, how can this get different?

E.g. for kernel 4.0 you'll get Linux libraries libgap.so.4.0.0, and for kernel 3.1 you will get libgap.so.3.1.0. The beauty of libtool versioning :-)

fingolfin: That reasoning is flawed, I am afraid, but you are not the first to fall in the trap (I was there, too).

If it was just a matter of identifying kernel versions with library versions, we could use Libtool's `-version-number major[:minor[:revision]]` interface and be done with it! But they caution against it, and there is a reason: it is difficult to map such triples correctly to all different operating systems, which use very different schemes for how to indicate shared library compatibility. The advantage of the current/revision/age approach is that one gets a monotonic increasing "version number" current, which can be compared like that, and be fitted into any system.


We are considering the following twospaces:
1. K := N x N,  containing pairs (kernel major, kernel minor)
2. L := { (c,a) \in N x N |  c >= a } ,  containing pairs (libtool current, libtool age)
and the maps
 f: L -> K,  (c,a) -> (c-a, a)    (mine)
 g: K -> L,  (M,m) -> (M+m, m)     (yours)

Now, our maps must preserve the *semantics* of the versions.  In your scheme, as I said, the kernel versions 3.1 and 4.0 are mapped to libtool current.age pairs  (4,1) and (4,0).  That indicates that the two are (a) binary compatible (same "current") version, and that the first is *newer* than the second. But that's wrong!

The map f of course also would have this issue, except it does *not* if one follows the rules for dealing with libtool current/age pairs: whenever current is incremented, then either age is incremented, too (meaning we stay backwards compatible), or else it is reset to 0 (meaning a breaking change, not backwards compatible).
Then   we might have a progression like this (left c/a pairs, right side their image under f):
  1,0 -> 1.0
  2,1 -> 1.1
  3,2 -> 1.2
  4,0 -> 4.0
  4,1 -> 4.1

And this preserves ordering semantics in the images, too.

Ah well. I guess in the time it took me to write all these explanations, I could have just implemented this myself *sigh*

dimpase: well, either this, or https://github.com/gap-system/gap/pull/3928#issuecomment-604420578, is wrong. Summing up the equations you wrote in the latter one comes to the very expression for libgap_current you object against.

ChrisJefferson: I can quite belive I was wrong, sorry.

dimpase: It boils down to the rule one needs to use to increment version. With K, whenever backward compatibility after kernel version (x,y) is to be broken, the next should be (x+y+1,0). E.g. after (x,y)=(3,1) the next kernel should be (5,0), and not (4,0).

I can either add this in comments, or indeed, re-implement the way you prefer, i.e. use `f` rather than `g`.
Please shout your preference.

fingolfin: @dimpase no, the comment you quote is correct -- it describes the use of the function `f`. As I explained above, this is NOT equivalent to function `g`, which your PR currently uses.

And yes, of course we can implement semantic versioning for the kernel version on its own. That's what we are already doing. But then there is NO WAY to map this to libtool versions. Your way (using function `g`) is incorrect, as it breaks the ordering. This is not a matter of taste, but one of wrong vs right.

dimpase: I don't understand, I think that's an elementary maths that's a matter of taste, and result 
is the same. (you probably didn't notice the extra rule on kernel version bumping that needs to be imposed to achieve equivalence).

Claim: using g with the rule (D) that (x,y) in K is bumped either to (x,y+1) [compatibility case] or to (x+y+1,0), is equivalent to using  f with the rule (M, aka libtool) that (c,a) in L is bumped to either (c+1,a+1) [compatibility case] or to (c+1,0). 

Indeed, f=[[1,-1],[0,1]] and g=[[1,1],[0,1]] are inverses of each other on RxR, we only need to check that the bumping leads to the same thing. If (c,a) is bumped to (c+1,a+1) [compatibility case] then the kernel version f(c,a)=(c-a,a) is bumped to  f(c+1,a+1)=(c-a,a+1), as'd be by (D).
If (x,y) in K is bumped to (x,y+1) then g(x,y+1)=(x+y+1,y+1), as'd be by (M) - we bump
(x+y,y) to (x+y+1,y+1).
 
If (c,a) is bumped to (c+1,0) then the kernel version f(c,a)=(c-a,a) is bumped to
f(c+1,0)=(c+1,0)=((c-a)+a+1,0), as it should by (D).
If (x,y) in K is bumped to (x+y+1,0) then g(x+y+1,0)=(x+y+1,0), as it'd be by (M) - we bump (x+y,y) to (x+y+1,0).     QED.

ChrisJefferson:
Extra comment -- however we end up defining the ABI, at the moment how these values should be updated is documented in src/modules.h , around line 37 (I'm not claiming that is the best place for it to be). If you end up changing how the values should be updated, be sure to update the docs (and feel free to move it to their definition, but at least it should be correct).

fingolfin:
Yes with an artificial rule for maintaining the kernel versions it would also work. But we don't want that. I already explained three times now how we would like it.

dimpase:
well, the question is whether you like to introduce one extra rule for exising pair of parameters, or a complete new (not to you and me, but to some other devs) pair of parameters, which need to be bumped up using an extra rule, too.
 
The former might be less invasive.

fingolfin:
I stated what I want . I will not continue this pointless discussion. I already wasted more time on it than it would take to implement it right from scratch

dimpase:
I think it is a good idea to ask whether people mind switching to new versioning scheme, in particular as libgap is not really the main GAP part, and it's not clear why its version shoud drive the kernel version.

ChrisJefferson:
In practice, I think only me and @fingolfin update these numbers, and as long as I only have to edit one version, and not two independent versions, I'm happy. I also trust @fingolfin 's opinion on what's best for libtool.

dimpase:
@ChrisJefferson: In practice, there are dozens of downstream projects which would curse whoever make a switch to a new versioning scheme, they'd have to figure out, and adapt their work to. It's not the first time GAP would be making the life of downstream harder than necessary, for reasons not really convincing. 

Why not play it nice?

ChrisJefferson:
I don't really see what we are "breaking". I don't think anyone else cares about GAP's kernel version outside kernel developers (and to them it will look the same anyway), and we don't current have a libgap version at all.

dimpase:
> I don't really see what we are "breaking".
GAP kernel packages? Linux distributions that package GAP (and libgap)?

ChrisJefferson:
Wait, now I'm totally confused.

Based on what Max suggested, as I understand it we are still going to have gap_kernel_major_version and gap_kernel_minor_version. The only difference is now they will be derived from the libtool versions, which will start(in stable-4.11 as current=7, age=0, revision=0 and in master as current=8, age=0, revision=0.

They will be derived by:

    gap_kernel_major_version = libtool_current - libtool_age
    gap_kernel_minor_version = libtool_age


The gap_kernel_major_version will possibly skip some numbers, but that doesn't matter, as long as it is monotonically increasing, and goes up whenever we break backwards compatability, which will happen because whenever we do that we will increment libtool_current by 1 and reset libtool_age to 0.

dimpase:
The difference between what Max proposes and what I implement is that Max proposes to switch to new pair of inputs (libtool's `current` and `age`), and my implementation sticks with the old ones - both schemes result in the same `gap_kernel_major_version` and `gap_kernel_minor_version`.

dimpase:
an update to switch to libgap c:r:a versioning is to be added shortly.

dimpase:
The last commit changes the versioning to use libtool's `(c:r:a)` numbers as input.

By the way, different sources say different things about incrementing `c, r, a`, e.g. https://autotools.io/libtool/version.html says that `r` must be incremented always.

configure.ac
dimpase: fixed

PR #3933, opened 2020-03-29
Add libtool versioning to libgap shared library

fingolfin:
We set up a libtool library version using the `current:revision:age`
format employed by libtool (see the libtool manual for details). We
currently do not use the `revision` value and use 0 instead. To avoid an
ever increasing number of versions, the kernel version is derived from
the libtool versioning info as follow:

    kernel_major := current - age
    kernel_minor := age

This provides the right semantics, provided we diligently follow the
rules for adjusting the values of `current` and `age` during releases,
as outlined in a comment in `configure.ac`.

Closes #3928

We could backport this to stabe-4.11, but that creates a somewhat iffy situation, as technically it would break the ABI going from 4.11.0 to 4.11.1; something we promised not to do. Yet one of the people who specifically asked whether we guarantee ABI stability in the 4.11 series, namely Bill Allombert from Debian, apparently is suggesting on the GAP mailing list that we do just that (however, I should qualify that this is just my understanding of what he said -- I may have misunderstood and don't want to misrepresent him). While that is all good and fine, I am not sure how other people packaging GAP would like this (e.g. @jamesjer working Fedora packages). Right now I think I'd prefer if we did not backport this, and just tell upstream packagers who want to know that GAP 4.12 will use the library version 8, so they can set any lower value for GAP 4.11, should they want to.

### DIFF
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -27,9 +27,13 @@ GC_SOURCES = @GC_SOURCES@
 COMPAT_MODE = @COMPAT_MODE@
 GAPARCH = @GAPARCH@
 
-#
+# GAP version and release date
 GAP_VERSION = @GAP_VERSION@
 GAP_RELEASEDAY = @GAP_RELEASEDAY@
+
+# libtool library version
+GAP_LIBTOOL_CURRENT = @GAP_LIBTOOL_CURRENT@
+GAP_LIBTOOL_AGE = @GAP_LIBTOOL_AGE@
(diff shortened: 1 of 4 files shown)


jamesjer:
I can cope with the situation either way.  If you merge it, I will make the necessary package adjustments.  If you don't ... well, then there is nothing for me to do until 4.12 is released.  Thank you for letting me know about this so I can plan ahead.

PR #3935, opened 2020-03-30
FIX for #3934

### LINKED ISSUE #3934: Problem with IsomorphismGroups
The error described below occurs in the master branch but not in GAP 4.11.

First prepare two groups.
```
LoadPackage( "grpconst", false );;
g:= PerfectGroup( IsPermGroup, 322560, 6 );;
ext:= CyclicExtensions( g, 2 );;
genslist:= List( ext{ [ 17, 18 ] }, GeneratorsOfGroup );;
```
In order to demonstrate that the problem depends only on the groups, not on the above session history, save the generators, start a new GAP session, and load the generators.
Then enter the following.
```
gap> IsomorphismGroups( Group( genslist[1] ), Group( genslist[2] ) );;
Error, Action not well-defined. See the manual section
``Action on canonical representatives''. at /.../gap/lib/oprt.gi:3050 called from
ImageElmActionHomomorphism( hom, gen
 ) at /.../gap/lib/oprt.gi:890 called from
[...]
```

hulpke:
Make sure an `autactbase' is not inherited wrongly. Can happen when using the old automorphism group code within an isomorphism test.

Also prune small steps in elementary abelian series to make example run through in reasonable time. No bugfix file provided, as the full test takes a few minutes.

This does not affect 4.11 or earlier, thus no release note needed.

This fixes #3934

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -612,6 +612,22 @@ local ff,r,d,ser,u,v,i,j,k,p,bd,e,gens,lhom,M,N,hom,Q,Mim,q,ocr,split,MPcgs,
     SortBy(d,Size); # in case reversed order....
   fi;
 
+  # now go up in series if elementary abelian to avoid too many tiny steps
+  u:=1; # last group in the series to be used
+  i:=2;
+  while i<Length(d) do
+    p:=IndexNC(d[i+1],d[u]);
+    # should we skip group i?
+    if p<100 and IsPrimePowerInt(p) and
+       HasElementaryAbelianFactorGroup(d[i+1],d[u]) then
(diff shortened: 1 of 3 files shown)


hulpke:
@ThomasBreuer Thank you for the remarks and approval. The reason for combining with the enhancement was to have the example actually being able to run through in reasonable time. Without it it takes extremely long. (Alas its still too much for a bugfix test).

I have added a check to catch cases (as here) where `autactbase` is clearly completely wrong. It is not a user-documented option, so I think the danger is quite small.

The reason for having it as an option is to allow setting it through a call of `NiceMonomorphism`. The alteranative would be to duplicate quite a bit of code to provide a special function`NiceMonomorphismWithAutactbase`  whose purpose would be solely to hand through the extra information.

PR #3938, opened 2020-03-31
Remove sySetjmp / Syjmp_buf (was Enable _POSIX_C_SOURCE)

ChrisJefferson:
This is required to build NormalizInterface on Cygwin, but could be useful in other situations.

I explictly mentioned the Cygwin case in the comment, just because someone finding this in 5 years might wonder what the heck it is there for. If it annoys people, I could put it in the commit message instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -903,24 +903,6 @@ AC_SEARCH_LIBS([cos], [m], [], [
 dnl check for math functions
 AC_CHECK_FUNCS([log2 log10 log1p exp2 expm1 exp10 trunc])
 
-dnl sigsetjmp is a macro on some platforms, cannot use AC_CHECK_FUNCS
-AC_MSG_CHECKING([for sigsetjmp])
-AC_LINK_IFELSE(
-  [AC_LANG_PROGRAM([[#include <setjmp.h>]],
-      [[sigjmp_buf t; sigsetjmp(t, 0)]])],
-  [AC_MSG_RESULT([yes])
-   AC_DEFINE([HAVE_SIGSETJMP], [1], [Define to 1 if you have the `sigsetjmp' function.])],
-  [AC_MSG_RESULT([no])])
-
-dnl _setjmp is a macro on some platforms, cannot use AC_CHECK_FUNCS
-AC_MSG_CHECKING([for _setjmp])
-AC_LINK_IFELSE(
-  [AC_LANG_PROGRAM([[#include <setjmp.h>]],
-      [[sigjmp_buf t; _setjmp(t, 0)]])],
-  [AC_MSG_RESULT([yes])
-   AC_DEFINE([HAVE__SETJMP], [1], [Define to 1 if you have the `_setjmp' function.])],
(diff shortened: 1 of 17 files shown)


src/common.h
```diff
@@ -13,6 +13,12 @@
 
 #include "config.h"
 
+// GAP uses various POSIX functions, such as siglongjmp
+// This is required when compiling with -std=c++11 on Cygwin
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 1
```
fingolfin: It seems we require newer POSIX: snprintf for example:
```suggestion
#define _POSIX_C_SOURCE 200112L
```
But even then it breaks on macOS, because now `strlcat` and `strlcpy` are hidden, yet `configure` reported their presence *sigh*. We can of course work around that, too (for starters, I now regret that I used these BSD extension functions under their BSD names - it's OK to use them, but I just should have called them `gap_strlcpy` and `gap_strlcat` and never even bothered to check for a system implementation).

But that "easy" change is spiraling quite a bit out of control now...

fingolfin: Oh and I also worry that we might be breaking yet moreL by #defining this in common.h., it'll affected every GAP kernel extension, and then those might be broken because features they rely on are "lost".

So perhaps we should just ask all GAP kernel extensions to not use the "strict" C99 / C++11 / etc. modes? Ah well

ChrisJefferson:
I hoped declaring it globally might make things better... In the long term, we could try to remove autoconf things from headers we use in kernel extensions, but thats a long term plan obviously.

fingolfin:
Actually, if you check closely you'll find that I already worked hard during the last half year or so to remove virtually all uses of things that depend on `config.h` from the headers!

But this one is a trick one, because sysjmp.h *has* to be included from `gapstate.h` and `libgap-api.h`, at least considering how things work now.

However, I wonder if we couldn't get away with just using plain `setjmp`/`longjmp`... Is using `siglongjmp` anything other than a tiny and irrelevant optimization?

ChrisJefferson:
I thought siglongjmp does more, but looking it, it lets you control if save signal information, while longjmp might, or might not.

However, there should be basically no GAP code where we are time dependant on longjmp, and we certainly can't rely on siglongjmp for correctness as we always pass it 0. I'll try doing a quick sanity check benchmark.

ChrisJefferson:
If I make a C program whose only job is to setjmp/longjmp, then siglongjmp is about 4% faster. However, I can also do 100,000,000 in 0.7 seconds, so I don't think it's measurable.

ChrisJefferson:
Right, this (now poorly named) branch is now killing sySetjmp instead :)

fingolfin:
@ChrisJefferson on second thought, this is an ABI breaking change, so I don't think we should backport it after all... At least not w/o clearing it with at least Debian and Fedora...

ChrisJefferson:
I agree. I'm not expecting to switch to my "new installer" on 4.11.x, as it does introduce several small  user-visible changes, so backporting isn't a major requirement.

PR #3939, opened 2020-04-01
Cygwin improvements for new release scripts

ChrisJefferson:
This improves the Cygwin release in a few ways. ~~This should only be merged once we have decided we are definitely moving to my new Windows release, because this breaks the current release.~~

The improvements are:

~~1) Use 'cygstart' to open webpages. This deals with unix vs windows filenames. It doesn't let us handle anchors, but after some googling this appears impossible in windows, as `#` can appear in filenames.~~

~~2) always use `sh` to spawn external programs (`sh` can still run windows programs)~~

3) Don't try to hard-wire the temp directory (and on many computers the guess is wrong). Instead we reconfigure cygwin to tell it to bind `/tmp` to the correct temp directory (this is done by the installer).

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -517,13 +517,20 @@ end );
 ##  <Func Name="ARCH_IS_WINDOWS" Arg=''/>
 ##
 ##  <Description>
-##  tests whether &GAP; is running on a Windows system in Cygwin.
+##  tests whether &GAP; is running on a Windows system without
+##  standard POSIX tools available (such as a shell).
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
 BIND_GLOBAL("ARCH_IS_WINDOWS",function()
-  return POSITION_SUBSTRING (GAPInfo.Architecture, "cygwin", 0) <> fail;
+  # Exit early is we are not in Cygwin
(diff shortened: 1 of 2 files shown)


fingolfin:
Also, could you explain in how far this breaks the current release?

ChrisJefferson:
So, the changes are:

1) Use the program 'cygstart'. We don't currently package that, and if we did we wouldn't set up the paths correctly.

2) Similarly, run exteral commands via 'sh', we don't distribute that either.

3) Use `/tmp` for temp files -- at the moment that won't work because we don't set up standard unix directories. Further, in the new installer we make cygwin point `/tmp` at the windows temp directory.

ChrisJefferson:
Hi @fingolfin or @wilfwilson , this very old PR is now ready to merge, it should not effect other releases

wilfwilson (approved):
@ChrisJefferson and I have just been talking about it this PR, and associated Cygwin-related stuff. I've crossed out some text in the previous parts of this discussion, because they're no longer applicable.

Given our discussion, I think this should be merged.

wilfwilson:
In more detail, @ChrisJefferson said this PR doesn't break the existing GAP release system for Windows. It also does two things:

* The changes to the temporary directory that Chris mentioned above
* This redefines `ARCH_IS_WINDOWS` to mean 'is running in Cygwin but not in a full Cygwin installation' (see the changed documentation).

PR #3944, opened 2020-04-04
Fix error checking in PartialPerm

### LINKED ISSUE #3943: bus error resulting from a short GAP function
as reported by Alex Degtyarev on support list (see below for the original post): 

### Observed behaviour
```
gap> shift_perm :=
> function(perm, delta)
>  perm := ListPerm(perm);
>  return AsPermutation(PartialPerm([1..Length(perm)] + delta, perm + delta));
> end;
function( perm, delta ) ... end
gap> p := shift_perm((2,3), -1);
(1,2)
gap> p^-1;
Bus error
```

### Expected behaviour

to be found 

```
┌───────┐   GAP 4.12dev-355-gee918e7-dirty built on 2020-03-27 08:20:52+0800
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2019.09.04, 
             AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, 
             CTblLib 1.2.2, FactInt 1.6.3, FGA 1.4.0, Forms 1.2.5, 
             GAPDoc 1.6.3, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4, LAGUNA 3.9.3, 
             orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, PrimGrp 3.4.0, 
             RadiRoot 2.8, recog 1.3.2, ResClasses 4.7.2, SmallGrp 1.4.1, 
             Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.4, 
             utils 0.69
```
(also reported to crash on Windows, and with different GAP versions.)

### the original report 

```
Date: Sat, 4 Apr 2020 04:50:54 +0300
From: Alex Degtyarev <degt@fen.bilkent.edu.tr>
To: support@gap-system.org
Subject: [GAP Support] bug report

[...]

Here is a GAP issue that appears to be a bug. Yes, it is my programming
mistake, however, the code below, instead of raising an Error at some point
or returning fail, just miserably crashes (under MS PowerShell on Windows
10; both 4.10.1 32 bit and 4.11.0 64 bit). Note also that mathematically the
operation is quite meaningful and, well, I just forgot how I programmed it
long ago and tried to use in the unforeseen direction :)

Best regards,
Alex Degtyarev

P.S. Accidentally, what is the "right" way to do what I'm trying to achieve?

shift_perm :=
function(perm, delta)
 perm := ListPerm(perm);
 return AsPermutation(PartialPerm([1..Length(perm)] + delta, perm + delta));
end;
p := shift_perm((2,3), -1);
p^-1;
```

ChrisJefferson:
Fixes #3943 , corruption caused by invalid inputs (numbers < 1) to PartialPerm

### DIFF
--- a/lib/pperm.gi
+++ b/lib/pperm.gi
@@ -518,8 +518,8 @@ function(arg)
                     "duplicate-free,");
     fi;
   elif Length(arg) = 2 then
-    if IsSSortedList(arg[1]) and ForAll(arg[1], IsSmallIntRep)
-        and IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsSmallIntRep)
+    if IsSSortedList(arg[1]) and ForAll(arg[1], IsPosInt and IsSmallIntRep)
+        and IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsPosInt and IsSmallIntRep)
         and Length(arg[1]) = Length(arg[2]) then
       return SparsePartialPermNC(arg[1], arg[2]);
     else
--- a/tst/testinstall/pperm.tst
+++ b/tst/testinstall/pperm.tst
@@ -3333,8 +3333,40 @@ gap> MultiplicativeZeroOp(x);
 gap> MultiplicativeZero(x);
 <empty partial perm>
 
-# Test PartialPerm (for sparse incorrect arg)
+# Test PartialPerm
 gap> PartialPerm([1,2,8],[3,4,1,2]);
+Error, usage: the 1st argument must be a set of positive integers and the 2nd \
+argument must be a duplicate-free list of positive integers of equal length to\
+ the first
+gap> PartialPerm([0,1],[2,3]);
+Error, usage: the 1st argument must be a set of positive integers and the 2nd \
+argument must be a duplicate-free list of positive integers of equal length to\
+ the first
+gap> PartialPerm([2,3],[0,1]);
+Error, usage: the 1st argument must be a set of positive integers and the 2nd \
+argument must be a duplicate-free list of positive integers of equal length to\
+ the first
+gap> PartialPerm([-2,-3],[5,6]);
+Error, usage: the 1st argument must be a set of positive integers and the 2nd \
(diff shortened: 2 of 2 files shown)


lib/pperm.gi
```diff
@@ -518,8 +518,8 @@ function(arg)
                     "duplicate-free,");
     fi;
   elif Length(arg) = 2 then
-    if IsSSortedList(arg[1]) and ForAll(arg[1], IsSmallIntRep)
-        and IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsSmallIntRep)
+    if IsSSortedList(arg[1]) and ForAll(arg[1], IsPosInt and IsSmallIntRep)
+        and IsDuplicateFreeList(arg[2]) and ForAll(arg[2], IsPosInt and IsSmallIntRep)
         and Length(arg[1]) = Length(arg[2]) then
       return SparsePartialPermNC(arg[1], arg[2]);
```
fingolfin, line 524: Strictly speaking I'd feel better if we improved the input validation in `SparsePartialPermNC` instead or in addition to the above, but as this PR fixes the immediate issue, I'd be happy to see it merged as-is, too.

fingolfin: The `small pos int` and "sorted list" tests can be done very cheaply in the kernel; duplicate free might be more work, though perhaps even that could be done as a side effect of the kernel code. Anyway, as long as there is no performance issue here... *shrug*

ChrisJefferson: I agree porting this all into the kernel might be netter, most of this comes almost for free while you are building the partial permutation (except DuplicateFreeList), but that would be a bigger change, and this little fix feels better for backporting.

fingolfin:
Backported to stable-4.11 in aac1175f06e4643cb65913598b7d88354ff5ff8c

PR #3945, opened 2020-04-06
kernel: only #include sysjmp.h where necessary

fingolfin:


fingolfin:
Travis PR actually passed after restarting failed jobs, but the state is stuck here...

@ssiccha no, no hidden pitfalls in here (that I am aware of)

Anyway, if @ChrisJefferson has no objections, I guess we could just merge this.

PR #3947, opened 2020-04-09
Use travis-ci caching for package file

ChrisJefferson:
This adds caching of the results of bootstrap-pkg-*, to (hopefully) reduce the number of failures we see.

I also added a timeout of 10 minutes, to try to deal with very slow downloads timing out.

### DIFF
--- a/etc/ci-download.sh
+++ b/etc/ci-download.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+#
+# This script expects a single argument, which should be an URL pointing to a
+# file for download; it then tries to download that file, into a local file
+# with the same name as the remote file.
+# If a file of the same name exists in ~/.gap-package-cache, then this file
+# is used if it has the same hash
+set -e
+
+PACKAGEDIR=~/.gap-package-cache
+
+# Get the name of the file we are downloading (everything after the final /)
+name=${*##*/}
+
+# Check if file already exists and has right hash
+if [ -f ${PACKAGEDIR}/"${name}" ] ; then
+    command -v sha256sum > /dev/null 2>&1 &&
+    {
+        hash1=$(curl --retry 5 --retry-delay 5 --max-time 10 -L "$*.sha256")
+        hash2=$(sha256sum ${PACKAGEDIR}/"${name}" | cut -d' ' -f1)
+        if [ "x$hash1" == "x$hash2" ] ; then
+            echo "Getting file from cache"
+            cp ${PACKAGEDIR}/"${name}" .
+            exit 0
+        fi
+    }
+fi
+
+if curl -L --retry 5 --retry-delay 5 --max-time 120 -O "$*" ; then
+    if [ -d ${PACKAGEDIR} ]; then
+        cp "${name}" ${PACKAGEDIR}
+    fi
+    exit 0
+fi
+exit 1
--- a/etc/ci-prepare.sh
+++ b/etc/ci-prepare.sh
@@ -44,9 +44,8 @@ fi
 time "$SRCDIR/configure" --enable-Werror $CONFIGFLAGS
 time make V=1 -j4
 
-# download packages; instruct curl to retry several times if the
-# connection is refused, to work around intermittent failures
-DOWNLOAD="curl -L --retry 5 --retry-delay 5 --max-time 120 -O"
+# Use alternative downloader which retries on failure and uses the Travis cache
+DOWNLOAD="$SRCDIR/etc/ci-download.sh"
 if [[ $(uname) == Darwin ]]
 then
     # Travis OSX builders seem to have very small download bandwidth,
--- a/etc/download.sh
+++ b/etc/download.sh
@@ -1,4 +1,4 @@
-#/bin/sh
+#!/bin/sh
 #
 # This script expects a single argument, which should be an URL pointing to a
 # file for download; it then tries to download that file, into a local file
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@
 dist: xenial
 language: c
 
+# Store downloaded package archives in ~/.gap-package-cache
+cache:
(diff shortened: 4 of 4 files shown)


etc/download.sh
```diff
@@ -1,12 +1,40 @@
...
 
 # check whether curl is available, and if so, delegate to it
-command -v curl >/dev/null 2>&1 && exec curl -L -O "$@"
+command -v curl >/dev/null 2>&1 &&
+{
+    name=${@##*/}
+    # Check if file already exists and has right hash
+    if [ -e ~/.gap-package-cache/${name} ] ; then
+        command -v sha256sum > /dev/null 2>&1 &&
+        {
```
fingolfin, line 17: Could we perhaps put this code into a separate script, say `etc/ci-download.sh`? Because it is also executed by users which execute a `make bootstrap-pkg-full` and I don't feel comfortable stuffing hundred of MB of data into a directory the user may not be aware exists.

They might also not want to retry; and they might be on a slow connection and not be happy if we abort the download after 10 minutes (download might take 12 minutes on their system...)

etc/download.sh
```diff
@@ -1,12 +1,40 @@
...
+            # Retry multiple times, including if download is very slow (>10 minutes)
+            hash1=$(curl --retry 5 --retry-delay 5 --max-time 600 -L "$@.sha256")
+            hash2=$(sha256sum ~/.gap-package-cache/${name} | cut -d' ' -f1)
+            if [ "x$hash1" == "x$hash2" ] ; then
+                echo "Getting file from cache"
+                cp ~/.gap-package-cache/${name} .
+                exit 0
+            fi;
```
fingolfin, line 25: I don't think the semicolon is needed here, is it?
```suggestion
            fi
```

etc/download.sh
```diff
@@ -1,12 +1,40 @@
...
+        }
+    fi;
+
+    if curl -L --retry 5 --retry-delay 5 -O "$@" ; then
```
fingolfin, line 29: BTW why the timeout when download the SHA256 file, but not here?

etc/download.sh
ChrisJefferson: I'm happy to move things around. My intention is that if you don't have a directory ~/.gap-package-cache none of this stuff will do anything, but that might be too subtle.

etc/ci-download.sh
```diff
@@ -0,0 +1,33 @@
+#/usr/bin/env bash
+#
+# This script expects a single argument, which should be an URL pointing to a
+# file for download; it then tries to download that file, into a local file
+# with the same name as the remote file.
+# If a file of the same name exists in ~/.gap-package-cache, then this file
+# is used if it has the same hash
+set -e
+
```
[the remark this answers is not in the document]
ChrisJefferson: One reason not to create the directory, is if it doesn't exist I assume there is no caching (travis creates it automatically)

etc/ci-prepare.sh
```diff
@@ -44,9 +44,8 @@ fi
 time "$SRCDIR/configure" --enable-Werror $CONFIGFLAGS
 time make V=1 -j4
 
-# download packages; instruct curl to retry several times if the
-# connection is refused, to work around intermittent failures
-DOWNLOAD="curl -L --retry 5 --retry-delay 5 --max-time 120 -O"
+# Use alternative downloader which retries on failure and uses the Travis cache
+DOWNLOAD="$SRCDIR/etc/ci-download.sh"
```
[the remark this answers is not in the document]
ChrisJefferson: So, at the moment appveyor (should) be no worse, it just won't use the cache

etc/ci-download.sh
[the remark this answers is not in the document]
ChrisJefferson: But I did make a variable

etc/ci-download.sh
```diff
@@ -0,0 +1,33 @@
...
+PACKAGEDIR=~/.gap-package-cache
+
+name=${*##*/}
```
fingolfin, line 12: I have no idea what this does.. I can of course look it up, but perhaps you could add a brief comment explaining it?

ChrisJefferson:
This is (I hope) now OK to merge -- we could do better, but I'd prefer to see if this seems to improve things in travis before trying to figure out how to get appveyor's caching to work (or just move to github actions / azure pipelines).

I ran shellcheck and fixed a couple of small issues it raised (including a typo in download.sh)

etc/ci-download.sh
```diff
@@ -0,0 +1,33 @@
...
+        hash2=$(sha256sum ${PACKAGEDIR}/"${name}" | cut -d' ' -f1)
+        if [ "x$hash1" == "x$hash2" ] ; then
+            echo "Getting file from cache"
+            cp ${PACKAGEDIR}/"${name}" .
+            exit 0
+        fi
+    }
+fi
+
+if curl -L --retry 5 --retry-delay 5 --max-time 120 -O "$@" ; then
```
fingolfin, line 27: Why $@ here but $* above

ChrisJefferson: Honestly, shellshock told me using "$@.sha256" was bad as it was combining an array and a string, so I did what I was told (I had a look at it seemed reasonable). I can change this one too. Certainly it should probably be $* now, as I don't claim this script would work downloading multiple files any more (if it ever did)

etc/ci-prepare.sh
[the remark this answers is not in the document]
fingolfin: OK

etc/download.sh
```diff
@@ -1,4 +1,4 @@
-#/bin/sh
+#!/bin/sh
```
fingolfin, line 1: Oops! My bad

fingolfin:
Backported to stable-4.11 in commits bdfd9bc9c and d805c7b8a

PR #3948, opened 2020-04-09
WIP: minimal makedocrel.g without pdf compilation

ssiccha:
This adds the file `doc/ref/makedocrelminimal.g`. It creates GAPDoc
documentation for the reference manual, but does not compile the latex
files.
## Further details

On my machine, compiling the documentation this way takes 30s instead of
3 minutes.  I think this is useful if one wants to work on the
documentation and compile it several times.

- [ ] proper formatting
- [ ] runnable tests


ChrisJefferson:
I haven't traced through how `make doc` works, but could a `make doc-minimal` target also be added, else I worry this file will get lost inside the doc/ref directory, and people won't find it.

ssiccha:
Maybe then one could also add something like `make doc-ref-minimal`. I very seldomly use `make doc` since it also compiles the `dev`, `hpc` and `tut` manuals.

fingolfin:
You should not need a new file for this, `make html` also doesn't. The main thing you are doing here is not building all manuals, just a single one, right?

ssiccha:
Ooh, I didn't even know there was a `make html`. I'd make a "copy" of that make target and call it `make doc-html` to make it more discoverable? (I have no idea how makefiles work) And then I'll figure out how to only build one manual. Shouldn't be too hard.

alex-konovalov:
if there is a `make html` already, do we need this PR? Building all manuals helps to get cross-references right. Most of the time is spent on building PDF, not html versions, and the other manuals are much smaller than the reference manual, and building their html versions will not create a large overhead.

ssiccha:
I'll figure out how much overhead it is in my machine. I'd introduce at least another make target like `doc-html-only` since simply `html` is not very descriptive. :)

fingolfin:
Several remarks:
- of course *just* building the ref manual (and building it only once) will of course still be faster
- I am a bit fuzzy right now as to what this will do to references: for the PDF using LaTeX, we need to run pdflatex twice to get those right. I am actually not sure right now what happens with the HTML: does this generate its own refs? If so, are those right with just a single run? Or do these refs rely on the pdflatex output somehow (and thus are broken if use `nopdf` mode)? I guess I should scour the GAPDoc manual and/or code to figure it out, but I am short on time. Perhaps @frankluebeck has time to quickly chime in and clarify this?
- target names could indeed be better, and easier to discover, if they all started with `doc`; then for me, tab-completion on `make doc` would reveal them. So perhaps `doc-html`, and for symmetry also `doc-pdf` ? And of course you could also have `doc-ref-html` and `doc-tut-pdf` and so on... But I am not sure how fine grained we really would want this.

frankluebeck:
The make targets `make manuals` and `make html` are already minimal to yield the four main manuals with correct cross references in all situations. (For cross references it is necessary to build all manuals once and then to build all of them again). 

The pdf-version of the manuals need 4 runs of pdflatex to get everything right. To avoid these on many machines time consuming runs or to allow running GAPDoc when no LaTeX is installed on a machine, there is now an option `"nopdf"` for `MakeGAPDocDoc`. This option is used by `make html`.

With  `"nopdf"` the text- and html-versions of the manuals are correctly built, `pdflatex` is not needed. If `bibtex` is available then it is used to produce the same labels in the bibliography as in the pdf-version (otherwise they can differ slightly).

For a quick check of changes in the documentation to a specific manual it would often be enough to run `MakeGAPDocDoc` once for this manual, provided this and all referenced manuals were built before. But this is really just for people who know what they are doing - so I would not create new make targets for this task.

I'm myself using the approach suggested in this pull request (that is copying the file `makedocrel.g` in a manual directory and change it as I like). I'm not sure if it is of suffiently general interest to put such alternative versions into the repository. If yes, I would change the pull request a little bit:

 - call the additional file `makedocrelnopdf.g` which is more selfexplaining
 - add the "MathJax" option (producing the additional HTML version takes very little time)
 - add a corresponding new file for any of the four main manuals

So, in `doc/ref` I would add a file `makedocrelnopdf.g` with content
```
Read( "makedocreldata.g" );

MakeGAPDocDoc( GAPInfo.ManualDataRef.pathtodoc,
               GAPInfo.ManualDataRef.main,
               GAPInfo.ManualDataRef.files,
               GAPInfo.ManualDataRef.bookname,
               GAPInfo.ManualDataRef.pathtoroot,
               "MathJax", 
               "nopdf");;

CopyHTMLStyleFiles(".");
```
On my computer `make manuals` takes about 2 minutes, `make html` about 70 seconds and `gap makedocrelnopdf.g` within the `doc/ref` directory about 15 seconds.

ssiccha:
I've also tried running `make html`. On my machine that takes 130s vs 30s for only the ref manual.

Also, I've checked the make file and it doesn't even use the `makedocrel.g` file. So I think there's two alternatives:
- get rid of the `makedocrel.g` files. I don't like this since they are quite useful for e.g. building only one manual
- make `make` use the `makedocrel.g` files and teach the `makedocrel.g` files to be able to not create pdf manuals if asked to

ssiccha:
I'm almost done. Currently only making the reference manual works, though adapting the files for the other manuals should be easy.

fingolfin:
This fails on Travis:
```
Error, file "makedocreldata.g" must exist and be readable at /home/travis/build/gap-system/gap/lib/files.gi:210 called from
Makefile.rules:803: recipe for target 'doc' failed
```

ssiccha:
rebased onto master

fingolfin:
@ssiccha any plans to "finish" this?

ssiccha:
Yepp, definitely. I was thinking about working on this and #4151 after Gap days.

PR #3950, opened 2020-04-13
Silence `gcc -fsanitize=undefined` warnings

fingolfin:
Reported by Bill Allombert on the GAP mailing list: Variable length arrays with length 0 are not permitted by the C standard. Change `DoOperationNArgs` to avoid them.

Note however that these VLAs are never accessed if n=0, so no problem in
practice should exist. Except that of course C compilers increasingly abuse
undefined behavior to mean "do whatever you want to look good on benchmarks,
e.g. produce NOPs or format the hard drive". So best to avoid this.

### DIFF
--- a/src/opers.c
+++ b/src/opers.c
@@ -1831,8 +1831,8 @@ static ALWAYS_INLINE Obj DoOperationNArgs(Obj  oper,
                  Obj  arg5,
                  Obj  arg6)
 {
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.11 in 569d9bc7a8f606789d5e55cdd3e201509a2de117

PR #3953, opened 2020-04-13
BuildPackages.sh: run prerequisites.sh if present

fingolfin:
This script is used by Semigroups. While it doesn't do anything for
released versions of the package (as far as I can tell), it seems
sensible to run it here. This way then, we can e.g. simplify some
of the code used in gap-infra.

Also check that autogen.sh resp. configure are executable,
not just merely files.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -160,11 +160,11 @@ build_fail() {
 run_configure_and_make() {
   # We want to know if this is an autoconf configure script
   # or not, without actually executing it!
-  if [[ -f autogen.sh && ! -f configure ]]
+  if [[ -x autogen.sh && ! -x configure ]]
   then
     ./autogen.sh
   fi
-  if [[ -f "configure" ]]
+  if [[ -x configure ]]
   then
     if grep Autoconf ./configure > /dev/null
     then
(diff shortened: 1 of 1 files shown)


alex-konovalov:
Yes, `prerequsites.sh` is documented at https://github.com/gap-infra/gap-docker-pkg-tests-master-devel. Why you're so sure that "it doesn't do anything for
released versions of the package" though?

fingolfin:
@alex-konovalov I am so sure because I tried it? For Semigroups, of course, the only package currently using this AFAIK. 

And, no offense, but I'd mentioning `prerequsites.sh` at an obscure location like https://github.com/gap-infra/gap-docker-pkg-tests-master-devel really doesn't count as "documented" for me. I'd consider it "documented" if we actually mentioned it in documentation, i.e., in our ref manual, or at least in the Example package documentation...

fingolfin:
Background: I am considering to switch NormalizInterface to use this, perhaps other packages, but that's not so useful if BuildPackage.sh doesn't know about it. Which is why I'd like to backport it, too, so that we can benefit from it in 4.11.1 already.

It'd also be handy for the GAP-Julia interface, which provides a wrapper around @mtorpey's PackageManager

fingolfin:
Yeah, it's useful. I am considering to use this for NormalizInterface.

fingolfin:
Backported to stable-4.11 in commit 77f179c169d285259f331c9a8c35dbc671f60a3f

PR #3954, opened 2020-04-13
Travis: add a timeout to curl downloads

fingolfin:
Normally Travis downloads the bootstraps in under 20 seconds; we give it 2 minutes before we retry

This should hopefully help with all those runaway Travis jobs that timeout after 50 minutes because curl is stuck. With this PR, ideally builds should complete (because I expect that if a slow download is retried, it becomes fast). At worst, we'll get a timeout after ~12 minutes ((5+1) times 2 minutes), so the Travis job will end quicker.

Note that PR #3947 also does something similar, but also does a lot more; I hope that this PR is uncontroversial and can be merged soon.

### DIFF
--- a/etc/ci-prepare.sh
+++ b/etc/ci-prepare.sh
@@ -46,7 +46,7 @@ time make V=1 -j4
 
 # download packages; instruct curl to retry several times if the
 # connection is refused, to work around intermittent failures
-DOWNLOAD="curl -L --retry 5 --retry-delay 5 -O"
+DOWNLOAD="curl -L --retry 5 --retry-delay 5 --max-time 120 -O"
 if [[ $(uname) == Darwin ]]
 then
     # Travis OSX builders seem to have very small download bandwidth,


fingolfin:
Another benefit is that builds in general should be fast: I just noticed a Travis job where the download completed after 36 minutes. The job passed overall, but took 35 minutes longer than necessary. With this PR, I hope it'll be "just" 2 minutes wasted (as I'd expect the first retry to go through). We'll see.

fingolfin:
OK, so I just saw a job where it tried 6 times in a row and always got slow speeds. So it looks as if indeed the problem is with (certain) Travis machines... So the issue persists; but at least now the jobs fail faster, which is beneficial (they proceed quicker to building the next PR).

fingolfin:
Also, jobs which in the past barely passed (e.g. run 45 minutes instead of usual 5) now fail. Which I personally consider an improvement, but others might disagree, it's a bit subjective ;-)

alex-konovalov:
@fingolfin I may want the same in the setup for packages too, if we will see here that it helps.

fingolfin:
backported to stable-4.11 in 6d8731d5b456854d719eb2df15875b3e375b7163

PR #3955, opened 2020-04-14
buildsys: add maintainer mode

fingolfin:
Maintainer mode (on by default) controls whether our build system
automatically regenerates `configure` if `configure.ac` or any of its other
sources changes, by running `autoconf`. Similarly, it regenerates
`src/config.h.in` if necessary by running `autoheader`.

While this is very useful when developing GAP, it can be problematic if a user
wants to compile a release version of GAP but does not have autoconf
installed; or if for some reason the GAP source code is readonly and they want
to run an out-of-tree build. Thus we allow turning this off.

Note: I'd like to backport this to 4.11 because it would greatly help resolving an issue in the Julia package [GAP.jl](https://github.com/oscar-system/GAP.jl).

Note 2: Right now this mode is always on by default; we could change it so that it is on by default if the version ends with `dev` and off otherwise (so that release tarballs have it off)


fingolfin:
backported to stable-4.11 in 523a034b289084c17b553850c35d147275720012

PR #3956, opened 2020-04-14
Fix test suite to work on big endian systems (for stable-4.11)

fingolfin:
Submitted by Bill Allombert via the gap mailing list.

See also https://github.com/gap-system/gap/pull/3744 which contains a similar commit for the master branch (where the patched test has changed), and which also adds a Travis job running on a big endian system.


ssiccha:
Is it really ok, that the hash function can return different values depending on the architecture gap is run on? See my comment in PR #3958.

ssiccha:
And I've just read the discussion in #3744. Nevermind my previous comment. :)

PR #3961, opened 2020-04-15
Fix new plist

ChrisJefferson:
This adds a C++-compatible version of NewPlistFromArgs. I couldn't come up with anything which is both valid C and valid C++. Also I have to do a really horrible "exit from extern C".

I use the function in some places in some C++ files, just to prove it works.


src/plist.h
```diff
@@ -314,35 +314,54 @@ EXPORT_INLINE Obj NewImmutableEmptyPlist(void)
...
+**  which is then passed to NewPlistFromArray.
+**
+**  __VA_ARGS__ contains the list of arguments given to this macro.
+**  There is no (nice) way to convert this list of arguments to an array
+**  in both C and C++, so we need two implementations
 */
-EXPORT_INLINE Obj NewPlistFromArray(const Obj * list, Int length)
+
+#ifdef __cplusplus
+} // escape extern "C"
```
fingolfin, line 328: Not for this PR, but: wrapping `#includes` in `extern "C"` was a quick&dirty hack I did; on the long run, it'd probably be better if each of our headers did this instead (after its own `#includes`):
```C++
#ifdef __cplusplus
extern "C" {
#endif 
... original content ...
#ifdef __cplusplus
}
#endif 
```

src/plist.h
```diff
@@ -314,35 +314,54 @@ EXPORT_INLINE Obj NewImmutableEmptyPlist(void)
 /****************************************************************************
 **
 *F  NewPlistFromArray(<list>,<length>) . . create a plain list from a C array
+*F  NewPlistFromArgs(<args...>) .  create a plain list from list of arguments
+**
+**  This macro turns a variable-length list of macro arguments into an array,
```
fingolfin, line 319: "this macro" when you are taking about two things? Also, one isn't a macro anymore? Or is it?

src/plist.h
```diff
@@ -314,35 +314,54 @@ EXPORT_INLINE Obj NewImmutableEmptyPlist(void)
...
+**  in both C and C++, so we need two implementations
 */
-EXPORT_INLINE Obj NewPlistFromArray(const Obj * list, Int length)
+
+#ifdef __cplusplus
+} // escape extern "C"
+template <unsigned int length>
+EXPORT_INLINE Obj NewPlistFromArrayCpp(const Obj (&list)[length])
 {
     if (length == 0) {
```
fingolfin, line 332: Note that strictly spealing, length zero arrays are not allowed, so the code two lines above might give an error. Also, `__VA_ARGS__` is not allowed to be empty according to the standard (although GNU C and some other compilers support it as an extension)

ChrisJefferson:
I didn't realise any code called NewPlistFromArray. I've tweaked how I do this, to make the difference between the C and C++ implementations smaller.

src/plist.h
ChrisJefferson: Have (hopefully) clarified what's going on

src/plist.h
ChrisJefferson: That's very hard to reasonably fix (although it would be easy to fix in C++11, so let's consider that if it ever becomes a problem)

src/plist.h
ChrisJefferson: Now handling this in a different way (as for some reason this didn't work in cygwin -- not sure why)

src/plist.h
```diff
@@ -314,35 +314,46 @@ EXPORT_INLINE Obj NewImmutableEmptyPlist(void)
...
+
+extern "C++"
+template <unsigned int length>
+EXPORT_INLINE Obj NewPlistFromArrayCppWrap(const Obj (&list)[length])
+{
+    return NewPlistFromArray(list, length);
+}
+
+#define NewPlistFromArgs(...)                                                \
+    NewPlistFromArrayCppWrap({ __VA_ARGS__ })
```
fingolfin, line 349: I am afraid this requires C++11:
```
src/pperm.cc: In function ‘UInt** FuncINDEX_PERIOD_PPERM(Obj, Obj)’:
src/plist.h:347:30: error: extended initializer lists only available with -std=c++11 or -std=gnu++11 [-Werror]
     NewPlistFromArrayCppWrap({ __VA_ARGS__ })
                              ^
```

But if we use C++11, we might as well use a nicer implementation using variadic templates:
```C++
template<typename... Ts>
Obj NewPlistFromArgs(Ts... args)
{
    const int size = sizeof...(args);
    Obj res[size] = {args...};
    return NewPlistFromArray(res, size);
}
```

ChrisJefferson: Woops, I should have checked if this worked with C++03 explictly!

fingolfin: Also, we'd then have to decide that we really want to require C++11. Might then also require C11. Both standards are 8 years old... I am still not sure whether we could convince everybody? I guess somewhat older compilers supporting partial (pre-)version fo C++11 would also be OK? According to  <https://en.cppreference.com/w/cpp/compiler_support>, variadic templates arrived in GCC 4.4 (first released April 2009) resp. 4.3 (first released March 2008, but only supports an earlier version of the spec, which may or may not be sufficient for our needs).

Of course initializer lists also were introduced in GCC 4.4 (cf same website). I am not sure what "extended initializer lists" refers to, though. 

In any case, we'd have to use something like https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html(possibly tweaked to accept `-std=gnu++1x` or so?)

fingolfin: I guess https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html is the macro to use -- and looking at it, it seems to accept `c++0x` as a fallback, so perhaps we'd be OK with that -- using a 10 years old standard ought to be fine. Right? Right?

ChrisJefferson: I'd think so -- I've been using C++0x in some packages and other programs, and haven't had any complains for a few years.

fingolfin: I know that @hulpke sometimes is using a cluster which "only" has gcc version 4.4.7 20120313, but that one has perfectly fine support for either implementation of this PR, so should work fine.

hulpke: Also, the machine giot finally updated, so this gcc issue seems to be gone. But thanks for thinking of me!

src/plist.h
```diff
@@ -314,35 +314,42 @@ EXPORT_INLINE Obj NewImmutableEmptyPlist(void)
...
 *F  NewPlistFromArray(<list>,<length>) . . create a plain list from a C array
+*F  NewPlistFromArgs(<args...>) .  create a plain list from list of arguments
+**
+**  NewPlistFromArgs turns a variable-length list of arguments into
+**  an array, which is then passed to NewPlistFromArray.
+**
+**  There is no (nice) single implementation of NewPlistFromArgs that works in
+**  both C and C++, so there are two seperate implementations.
 */
+
```
fingolfin, line 325:
```suggestion
```

fingolfin:
With my tweaks, this PR now passes almost all CI tests -- except the one where we use gcc 4.7: apparently that one does not understand the `-std=gnu++0x` or `-std=c++0x` switches, based on the autoconf switches. But it should support the feature based on https://en.cppreference.com/w/cpp/compiler_support ?!

Actually, I just checked and https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gcc/Standards.html#Standards explicitly names this flag. Huh?

ChrisJefferson:
The problem is that ax_cxx_compile_stdcxx.m4 tries compiling a giant pile of C++11 code, some of which wasn't yet supported by g++ 4.7. We could either just check if the flag works somehow, or I'm happy to hack the bits out of ax_cxx_compile_stdcxx.m4 we don't need, but I don't know if you'd want to edit this file (does it ever get auto-updated or anything?)

ChrisJefferson:
I've added a fix to  ax_cxx_compile_stdcxx.m4 which fixes g++ 4.7. I didn't go and figure out exactly what was wrong, but I think it's a combination of constexpr and auto (neither of which we currently use anyway)

PR #3963, opened 2020-04-15
Implement compression/decompression of filenames ending .gz

ChrisJefferson:
This implements automatic compression/decompression of files ending .gz.

The documentation claims we do this... the functionality was mostly added but the final code was not added to SyFopen.

There are a number of tests which make sure everything works correctly.

One note: If a file ends in .gz, but is not a compressed file, it is still opened correctly. This means anyone who has tried using compression will not find they have unreadable files (this was always intended behaviour, to not break any strange old code).

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -616,6 +616,9 @@ Int SyFopen (
     Char                namegz [1024];
     int                 flags = 0;
 
+    Char * terminator = strrchr(name, '.');
+    BOOL endsgz = terminator && (strcmp(terminator, ".gz") == 0);
+
     /* handle standard files                                               */
     if ( strcmp( name, "*stdin*" ) == 0 ) {
         if ( strcmp( mode, "r" ) != 0 )
@@ -677,8 +680,11 @@ Int SyFopen (
 #endif
 
     /* try to open the file                                                */
-    syBuf[fid].fp = open(name,flags, 0644);
-    if ( 0 <= syBuf[fid].fp ) {
+    if (endsgz && (syBuf[fid].gzfp = gzopen(name, mode))) {
+        syBuf[fid].type = gzip_socket;
+        syBuf[fid].fp = -1;
+        syBuf[fid].bufno = -1;
+    } else if (0 <= (syBuf[fid].fp = open(name, flags, 0644))) {
         syBuf[fid].type = raw_socket;
         syBuf[fid].echo = syBuf[fid].fp;
         syBuf[fid].bufno = -1;
--- a/tst/testinstall/compressed.tst
+++ b/tst/testinstall/compressed.tst
@@ -1,0 +1,169 @@
+#@local dir,fname,isGzippedFile,stream,str
+gap> START_TEST("compressed.tst");
+gap> dir := DirectoryTemporary();;
+gap> fname := Filename(dir, "test.g.gz");;
+
+# Let us check when we have written a compressed file by checking the gzip header
+gap> isGzippedFile := function(dir, name)
+>    local out, str,prog;
+>    str := "";
+>    out := OutputTextString(str, true);
+>    Process(dir, Filename(DirectoriesSystemPrograms(),"cat"), InputTextNone(), out, [name]);
+>    return str{[1..2]} = "\037\213";
+>  end;;
+gap> str := "hello\ngoodbye\n";;
+
+# Write a compressed file
+gap> FileString( fname, str ) = Length(str);
+true
+
+# Check file really is compressed
+gap> isGzippedFile(dir, "test.g.gz");
+true
+
+# Check reading compressed file
+gap> StringFile( fname ) = str;
+true
+
+# Check gz is added transparently
+gap> StringFile( Filename(dir, "test.g") ) = str;
+true
+
+# Test reading/seeking in a gzip compressed file
+gap> stream := InputTextFile(fname);;
+gap> ReadLine(stream);
+"hello\n"
+gap> ReadLine(stream);
+"goodbye\n"
+gap> ReadLine(stream);
+fail
+gap> SeekPositionStream(stream, -1);
+fail
+gap> SeekPositionStream(stream, 0);
+true
+gap> ReadLine(stream);
+"hello\n"
+gap> ReadLine(stream);
+"goodbye\n"
+gap> ReadLine(stream);
+fail
+gap> SeekPositionStream(stream, 2);
+true
+gap> PositionStream(stream);
+2
+gap> ReadLine(stream);
+"llo\n"
+gap> ReadLine(stream);
+"goodbye\n"
+gap> SeekPositionStream(stream, 0);
+true
+gap> ReadAll(stream) = str;
+true
+gap> SeekPositionStream(stream, 0);
+true
+gap> PositionStream(stream);
+0
+gap> ReadAll(stream) = str;
+true
+gap> CloseStream(stream);
+
+# Test multiple writes
+gap> stream := OutputTextFile( fname, false );;
+gap> PrintTo( stream, "1");
+gap> AppendTo( stream, "2");
+gap> PrintTo( stream, "3");
+gap> CloseStream(stream);
+gap> stream;
+closed-stream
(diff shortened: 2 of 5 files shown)


fingolfin:
I am guessing this should be backported, right?

src/sysfiles.c
```diff
@@ -677,8 +680,13 @@ Int SyFopen (
 #endif
 
     /* try to open the file                                                */
-    syBuf[fid].fp = open(name,flags, 0644);
-    if ( 0 <= syBuf[fid].fp ) {
+
```
fingolfin, line 683: why two empty lines after that comment?

src/sysfiles.c
```diff
@@ -677,8 +680,13 @@ Int SyFopen (
...
+
+    if(endsgz && (syBuf[fid].gzfp = gzopen(name, mode))) {
```
fingolfin, line 685:
```suggestion
    if (endsgz && (syBuf[fid].gzfp = gzopen(name, mode))) {
```

src/sysfiles.c
```diff
@@ -677,8 +680,13 @@ Int SyFopen (
...
+        syBuf[fid].type = gzip_socket;
+        syBuf[fid].fp = -1;
+        syBuf[fid].bufno = -1;
+    } else if ( 0 <= ( syBuf[fid].fp = open(name,flags, 0644))) {
```
fingolfin, line 689:
```suggestion
    } else if (0 <= (syBuf[fid].fp = open(name, flags, 0644))) {
```

tst/testinstall/compressed.tst
```diff
@@ -0,0 +1,169 @@
...
+>    local out, str,prog;
+>    str := "";
+>    out := OutputTextString(str, true);
+>    Process(dir, Filename(DirectoriesSystemPrograms(),"cat"), InputTextNone(), out, [name]);
+>    return str{[1..2]} = "\037\213";
+>  end;;
+gap> str := "hello\ngoodbye\n";;
+
+# Write a compressed file
+gap> FileString( Filename(dir, "test.g.gz"), str ) = Length(str);
```
fingolfin, line 17:
```suggestion
gap> FileString( fname, str ) = Length(str);
```

tst/testinstall/compressed.tst
```diff
@@ -0,0 +1,169 @@
...
+true
+
+# Check file really is compressed
+gap> isGzippedFile(dir, "test.g.gz");
+true
+
+# Check reading compressed file
+gap> StringFile( Filename(dir, "test.g.gz") ) = str;
```
fingolfin, line 25:
```suggestion
gap> StringFile( fname ) = str;
```

tst/testinstall/compressed.tst
```diff
@@ -0,0 +1,169 @@
...
+true
+
+# Check gz is added transparently
+gap> StringFile( Filename(dir, "test.g") ) = str;
```
fingolfin, line 29:
```suggestion
gap> StringFile( fname ) = str;
```

tst/testinstall/compressed.tst
```diff
@@ -0,0 +1,169 @@
...
+gap> START_TEST("compressed.tst");
+gap> dir := DirectoryTemporary();;
+gap> fname := Filename(dir, "test.g.gz");;
+
+# Let us check when we have written a compressed file by checking the gzip header
+gap> isGzippedFile := function(dir, name)
+>    local out, str,prog;
+>    str := "";
+>    out := OutputTextString(str, true);
+>    Process(dir, Filename(DirectoriesSystemPrograms(),"cat"), InputTextNone(), out, [name]);
```
fingolfin, line 11: Hrm, that's yucky. Can you please at least add a comment that you are not simply using `StringFile` here because that would decompress the data?

I guess what we'd need to make this less ugly is an `InputRawFile` or so which does not perform any transparent decompression. Or perhaps an optional flag to `InputTextFile` which allows turning off the decompression?

tst/testinstall/compressed.tst
```diff
@@ -0,0 +1,169 @@
...
+
+# Check reading compressed file
+gap> StringFile( Filename(dir, "test.g.gz") ) = str;
+true
+
+# Check gz is added transparently
+gap> StringFile( Filename(dir, "test.g") ) = str;
+true
+
+# Test reading/seeking
```
fingolfin, line 32:
```suggestion
# Test reading/seeking in a gzip compressed file
```

tst/testinstall/compressed.tst
ChrisJefferson: Yes, I didn't like this. I'm happy with either of those suggestions, if anyone has a strong prference.

tst/testinstall/compressed.tst
ChrisJefferson: This one isn't the same (doesn't have the .gz)

tst/testinstall/compressed.tst
ChrisJefferson: Thinking a little more, the problem with introducing some kind of InputRawFile, or anything else, is it is another thing to document which (I think) no-one is ever going to want to use.

We could use the IO package to grab raw access, or introduce some undocumented kernel function to get raw access?

src/sysfiles.c
```diff
@@ -677,8 +680,11 @@ Int SyFopen (
 #endif
 
     /* try to open the file                                                */
-    syBuf[fid].fp = open(name,flags, 0644);
-    if ( 0 <= syBuf[fid].fp ) {
+    if (endsgz && (syBuf[fid].gzfp = gzopen(name, mode))) {
+        syBuf[fid].type = gzip_socket;
+        syBuf[fid].fp = -1;
+        syBuf[fid].bufno = -1;
+    } else if (0 <= (syBuf[fid].fp = open(name, flags, 0644))) {
```
fingolfin, line 687: Hm, so we are patching the most low-level function in the kernel for opening a file to transparently access .gz files. That makes it really tough to bypass this, one has to either bypass `SyFopen` or else copy or rename the file in question to look at its "true" content.

I think it is OK to do this now in this PR, but I'd like us to revisit this. I notice that `SyFopen` still takes an old clunky "mode" string as second argument. That argument could be augmented to indicate that "raw" or "binary" mode (no automatic (de)compressing) is wanted. Indeed, we already supported on Cygwin using `"rb"` and `"wb"` modes (for savefiles).

But I'd also suggest to change it from a `const char *` to an integer bit mask (basically similar to the POSIX `O_RDONLY` etc. flags for `open`, just out own portable version of those).

Anyway, either way it'd be trivial to add a simple kernel helper function which basically does the same as `StringFile` but using this "raw" mode (`StringFile` is a GAP function but one that's trivial to rewrite in a <10 lines kernel function, and then only the argument to `SyFopen` in there needs to be changed)

tst/testinstall/compressed.tst
fingolfin: Fine by me; see comments above for some ideas.

src/sysfiles.c
fingolfin: Of course the idea of changing the mode string to an int (or enum) is separate from the rest; this is just ramblings, don't feel obliged to implement any of this.

tst/testinstall/compressed.tst
fingolfin: Just to be clear: It's also OK by me to merge this now as-is; we can always revisit this later. Perhaps amend 1-2 of the minor optional remarks I made, if you feel like it; otherwise just merge it as-is *shrug*

src/sysfiles.c
ChrisJefferson: I'm not opposed to how to fix this, but I'm a bit unsure about implementing it at all. I don't really like adding stuff just to support testing. If anyone is upset they can't access raw gzip files I'm very happy to fix it -- although at the moment we don't really support binary file access at all (on Windows we always do line ending conversion for example, so we'd probably corrupt raw gzipped files anyway).

fingolfin:
Backported to stable-4.11 in commit c69cda343

PR #3964, opened 2020-04-16
BuildPackages.sh: add --with-gap option

fingolfin:
We call gap to parse PackageInfo.g files. To do so, so far we called
$GAPROOT/gap. But that does not work reliably with out-of-tree builds; instead
$GAPROOT/bin/gap.sh should be called (which we now do by default) But even
that might not be enough for special applications. Hence we added a --with-gap
option to allow users to override this, if they should really need to.

I'd like to backport this to stable-4.11, for use in the GAP-Julia bridge.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -32,6 +32,7 @@ fi
 
 CURDIR="$(pwd)"
 GAPROOT="$(cd .. && pwd)"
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.11 in commit e89f016dc4213431b72d7132c314eda60c9fb488

PR #3965, opened 2020-04-16
Fix potential garbage collector crashes on 64bit ARM systems

fingolfin:
Thanks to Jerry James @jamesjer (Fedora) and Bill Allombert (Debian) for help tracking this down.

Fixes #3919

For the backport, only the second commit is needed.

Issue #3919, opened 2020-03-16
GC problems in 4.11.0


jamesjer:
I am trying to build GAP 4.11.0 for Fedora Rawhide.  I am having issues with the garbage collector.  The initial build used the default gasman collector.  That worked fine on all architectures except aarch64 (the 64-bit ARM architecture).  GAP 4.10.2 did not have any issues on aarch64.  With 4.11.0, the executable and documentation built successfully, but many tests failed like this:
```
$ make check
/builddir/build/BUILD/gap-4.11.0/bin/gap.sh --quitonbreak -b -m 100m -o 1g -q -x 80 -r -l /builddir/build/BUILD/gap-4.11.0 -A  tst/testinstall.g
You should start GAP4 using `gap -A -x 80 -r -m 100m -o 1g -K 2g'.
Architecture: aarch64-redhat-linux-gnu-default64-kv7
testing: /builddir/build/BUILD/gap-4.11.0/tst/testinstall/ConjNatSym.tst
     148 ms (28 ms GC) and 23.8MB allocated for ConjNatSym.tst
testing: /builddir/build/BUILD/gap-4.11.0/tst/testinstall/DirectProductElement\
.tst
########> Diff in /builddir/build/BUILD/gap-4.11.0/tst/testinstall/DirectProdu\
ctElement.tst:2
# Input is:
START_TEST("DirectProductElement.tst");
# Expected output:
# But found:
Error, Length: <list> must be a list (not a function body bag)
########
     106 ms (91 ms GC) and 655KB allocated for DirectProductElement.tst
testing: /builddir/build/BUILD/gap-4.11.0/tst/testinstall/ElmsBlist.tst
########> Diff in /builddir/build/BUILD/gap-4.11.0/tst/testinstall/ElmsBlist.t\
st:1
# Input is:
START_TEST("ElmsBlist.tst");
# Expected output:
# But found:
Error, Length: <list> must be a list (not a integer)
########
     127 ms (82 ms GC) and 4.33MB allocated for ElmsBlist.tst
```

All of the failures refer to START_TEST.  A little experimentation shows that the problem happens at this line in START_TEST: `GASMAN( "collect" );`.  I can try this week to get access to aarch64 hardware to debug the problem, but would appreciate any ideas on what might be going wrong.

[...]

PR #3966, opened 2020-04-16
Travis: test builds on arm64, ppc64le, s390x (for stable-4.11)

fingolfin:
Besides just adding three jobs to the Travis build matrix, we have to
deal with some issues with the VMs for the three new platforms:

- <s>Switch to using Python 3 instead of Python 2 to workaround an issue where `python` is `python2.7` but `pip` is `pip3.6`.</s>
  Always call the right pip matching the `python` binary; and prepare for `python` being Python 3.
  For details, see <https://travis-ci.community/t/8165>

- Work around the fact that the pip cache dir is owned by root.
  For details, see <https://travis-ci.community/t/7822/6>

- Use `travis_terminate` liberally in `.travis.yml` to workaround another issue with the three new platforms; as a side benefit, we also get Travis to reliably and early on abort CI builds with failures.
  For details, see <https://travis-ci.community/t/7659/>.

- Adjust a test for `OPERS_CACHE_INFO` to pass on the three new platforms.

This is a backport of PR #3744. I am hoping that we see the crash from issue #3919 so that we can then test PR #3965.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ addons:
     - zlib1g-dev
     - python
     - python-pip
-    - perl
+    - jq    # for etc/travis_fastfail.sh
 
 #
 # The following test jobs are roughly sorted by duration, from longest to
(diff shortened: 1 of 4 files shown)


fingolfin:
Nice: the arm64 build failed exactly as predicted, i.e. we can reproduce the reports by @jamesjer and Bill. I'll now add the fix to this PR to see if it really helps.

fingolfin:
And now the fix makes it pass. So I guess this PR is simultaneously backports PR #3744 and PR #3965

PR #3967, opened 2020-04-18
buildsys: fix install-gaproot target for out-of-tree builds

fingolfin:
The file sysinfo.gap is generated, and not in the srcdir.


fingolfin:
Backported to stable-4.11 in commit 04ba0d2aabc2580473c4f5c2efb0dac939e8f537

PR #3968, opened 2020-04-18
BuildPackage.sh: prepare for NormalizInterface change

fingolfin:
... namely for it to rename its `build-normaliz.sh` script to
`prerequisites.sh`. For this we need to pass the GAPROOT to
`prerequisites.sh` (which doesn't hurt existing versions of
`prerequisites.sh` in other packages, as those expect no arguments).
Also, only call `build-normaliz.sh` if no `prerequisites.sh` is present.


fingolfin:
Backported to stable-4.11 in commit 33fd229ecf88328ec1a3301a270b74828f5add2e

PR #3969, opened 2020-04-18
Fix broken references in the reference manual

fingolfin:
This is a patch taken from the Fedora package by @jamesjer; see
https://src.fedoraproject.org/rpms/gap/blob/master/f/gap-ref.patch

### DIFF
--- a/doc/ref/function.xml
+++ b/doc/ref/function.xml
@@ -34,10 +34,8 @@ functions.
 
 
 <!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
-<Section Label="Calling a function with a list argument that is interpreted as
-several arguments">
-<Heading>Calling a function with a list argument that is interpreted as
-several arguments</Heading>
+<Section Label="Calling a function with a list argument that is interpreted as several arguments">
+<Heading>Calling a function with a list argument that is interpreted as several arguments</Heading>
 
 <#Include Label="CallFuncList">
 
--- a/doc/ref/pperm.xml
+++ b/doc/ref/pperm.xml
@@ -441,8 +441,7 @@ gap> EmptyPartialPerm();
 
   <ManSection><Heading>RandomPartialPerm</Heading>
     <Func Name="RandomPartialPerm" Arg="n" Label="for a positive integer"/>
-    <Func Name="RandomPartialPerm" Arg="set" Label="for a set of positive
-      integers"/>
+    <Func Name="RandomPartialPerm" Arg="set" Label="for a set of positive integers"/>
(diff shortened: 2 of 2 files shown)


fingolfin:
Hmm, I thought these were meant to fix `Ref`s to these sections. But yeah I don't see any, and it is not clear to me that the newlines actually matter. (I guess I shouldn't file PRs at 1:42 ;-).

Perhaps @jamesjer has time to clarify why that PR is necessary in Fedora, and/or @frankluebeck whether he thinks it has any effect.

jamesjer:
This was a response to lines such as these in the irredsol build log:
```
Ignored line: \makelabel{ref:RandomPartialPerm for a set of positive
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
Ignored line:    integers}{54.2.7}{X7E6ADC8583C31530}
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
Ignored line: \makelabel{ref:AsPartialPerm for a permutation and a set of
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
Ignored line:   positive integers}{54.4.1}{X81B32CB182489ACA}
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
Ignored line: \makelabel{ref:LQUO for a permutation or partial permutation
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
Ignored line:     and partial permutation}{54.5}{X848CD855802C6CE1}
... in /builddir/build/BUILD/irredsol-1.4.1/doc/./../../../doc/ref/manual.lab
```
I don't know if any GAP packages actually refer to those labels, but I "fixed" them just in case, to avoid broken cross-references.

I think the patch is just a workaround, which is why I never submitted it.  Really, convert.pl should be made to handle embedded newlines.  Sadly, my rudimentary perl skills are not up to the task.

alex-konovalov:
@jamesjer thanks, that makes sense to me now! So, IRREDSOL (and some other packages - list at #639) are using old `gapmacro.tex`-based  manual format, which is no longer supported, but luckily still works. I don't think anyone is interested in fixing `convert.pl`, so if this PR fixes the problem, I am happy to merge it. But there may be more packages then having similar problems, and in the future such new lines may appear again...

alex-konovalov:
See also #3977.

frankluebeck:
The error shown by @jamesjer above is not correctly fixed by this pull request. 

A (hopefully) correct fix is proposed in "Fix line breaks in manual.lab files" (#3977).

fingolfin:
Thanks @frankluebeck ! @jamesjer if you find the time (I know you are busy right now, but perhaps later), could you verify Frank's patch resolves the issue for you?

Or perhaps somebody else here could try to reproduce the issue and then see if Frank's patch helps. I am unfortunately unable to do it myself right now.

PR #3972, opened 2020-04-20
Fix BuildPackages.sh --parallel option

fingolfin:
This fixes the code just merged via PR #3922 (ouch).

Should of course be backported, too.


fingolfin:
Backported to stable-4.11 in commit 7f33ebc0568d46e18b26932682c6ce289e88786c

PR #3974, opened 2020-04-21
buildsys: overhaul dependency tracking

fingolfin:
In the past for a source file `src/foo.c` we generated the files
`obj/src/foo.lo` and `obj/src/.deps/foo.d`.

Now we generate `gen/obj/src/foo.c.lo` and `gen/deps/src/foo.c.d`.

Let's unravel this:

- we use `gen` instead of `obj`, matching what the new package
  buildsystem does: there seems to be no point in having two directories
  containing generated content, so let's merge them
- no more hidden `.deps` directories, instead this is out in the open.
- different .d files for C and C++ sources: this removes annoyances
  introduced when converting source files from C to C++; afterwards, a
  `git bisect` (or a switch to a different branch where the conversion
  did not take place) could trigger annoying build errors (which `make
  clean` resolved, but that's of course still annoying)
- likewise: different .lo files for C and C++ sources; while I am not
  aware of any issue this fixes, it allows to simplify the build system
  logic, and also means that adding more file extensions in the future
  is trivial.

Moreover, we now explicitly inform `make` that the .lo and .d files are
generated "together"; and we make `gap` and `libgap` depend on the .d
files, too. This way, if a .d file is lost, we force a rebuild of the
corresponding .lo file, and vice versa. Without this, it can sometimes
happen (e.g. when switching between commits during a `git bisect`) that
these files get "out of sync" and an incorrect build results.

In the future, we might want to rename `gen` to something nice, e.g.
`generated`, `build`, `products`, ... Be my guest in bikeshedding this


ChrisJefferson:
Sanity check, should that be `gen/...` for the `.d` file? If not, why not?

fingolfin:
You are right, a mistake on my part (I've been rewriting this thing today multiple times *sigh*).

Note that instead of `gen/deps/` and `gen/obj/` one could just put these all directly into `gen/`; I mainly separated this as I found it cleaner / easier to grok, but that's rather subjective. Another argument is that it's perhaps good to avoid `gen/src/` (which we'd get if I drop the `obj` and `deps` components from the path), as we have `gen` in the include path and so `#include "src/compiled.h"` would look into `gen/src/` too... but in reality, there should never be a `.h` file in there, so it doesn't matter).

All in all, this PR fixes *and* simplifies stuff by not trying to merge together different things... which to me is always a good hint that I am on the right track. But I need to really test this thoroughly before even considering this for merging; e.g. test the behavior when switching between older commits and this one; and when converting a file to C++ (and back) while on this PR etc.

fingolfin:
I am fully open to renaming `gen` to `build` or `products` or something like that. It's trivial for me to do as well.

fingolfin:
@ChrisJefferson any more comments on this PR?

ChrisJefferson:
I'm happy

PR #3976, opened 2020-04-22
Bugfix and test for IsIntegralRing

### LINKED ISSUE #3975: Wrong result in IsIntegralRing
Reported by @grahamknockillaree:
```
gap> R:=SmallRing(4,3);;
gap> IsIntegralRing(R);
true
gap> ShowMultiplicationTable(R);
*   | 0*a a   2*a -a
----+----------------
0*a | 0*a 0*a 0*a 0*a
a   | 0*a a   2*a -a
2*a | 0*a 2*a 0*a 2*a
-a  | 0*a -a  2*a a
```
The problem is in the line

https://github.com/gap-system/gap/blob/86c4135df7f1396dc19175b457c2b34581ea1b87/lib/ring.gi#L704

which should be 

    for k  in [i..Length(elms)]

to check the diagonal as well. 

Thank you for reporting it - I will make a PR tomorrow.

olexandr-konovalov:
Two bugfixes and test for `IsIntegralRing`

The first one fixes an off-by-one error causing missing the case when zero appears in the diagonal in the multiplication table.

The second one removes true method for `IsEuclideanRing` which was added in commit 0f9c104 and caused `IsIntegralRing` wrongly return `true` e.g. for `Integers mod 6`,
as reported in #3975.

Closes #3975.

### DIFF
--- a/lib/ring.gi
+++ b/lib/ring.gi
@@ -701,7 +701,7 @@ InstallMethod( IsIntegralRing,
         zero := Zero( R );
         for i  in [1..Length(elms)]  do
             if elms[i] = zero then continue; fi;
-            for k  in [i+1..Length(elms)]  do
+            for k  in [i..Length(elms)]  do
                 if elms[k] = zero then continue; fi;
                 if elms[i] * elms[k] = zero then
                     return false;
--- a/lib/zmodnz.gd
+++ b/lib/zmodnz.gd
@@ -129,13 +129,6 @@ InstallTrueMethod( IsFinite,
 
 #############################################################################
 ##
-#M  IsEuclideanRing( <R> ) . . . . . . . . . . . .  method for full ring Z/nZ
-##
-InstallTrueMethod(IsEuclideanRing, IsZmodnZObjNonprimeCollection and
-    IsWholeFamily and IsRing);
-
-#############################################################################
-##
 #V  Z_MOD_NZ
 ##
 ##  <ManSection>
--- a/tst/testinstall/ring.tst
+++ b/tst/testinstall/ring.tst
@@ -16,6 +16,10 @@ false
 gap> IsIntegralRing( SmallRing(4,3) );
 false
 
+# Zero divisors not on the diagonal
+gap> IsIntegralRing( Integers mod 6 );
+false
+
 # Integral rings
 gap> IsIntegralRing( GF(5) );
 true
--- a/tst/testinstall/ring.tst
+++ b/tst/testinstall/ring.tst
@@ -1,0 +1,30 @@
+gap> START_TEST("ring.tst");
+
+#####################################################################
+#
+# Tests for IsIntegralRing
+#
+# Trivial ring
+gap> IsIntegralRing( SmallRing(1,1) );
+false
+
+# Non-commutative ring
+gap> IsIntegralRing( SmallRing(4,7) );
+false
+
+# Zero divisors on the diagonal
+gap> IsIntegralRing( SmallRing(4,3) );
+false
+
+# Integral rings
+gap> IsIntegralRing( GF(5) );
+true
+gap> IsIntegralRing( Integers );
+true
+gap> IsIntegralRing( Rationals );
+true
+gap> IsIntegralRing( CF(4) );
+true
+
+#
+gap> STOP_TEST( "ring.tst" );


lib/ring.gd
```diff
@@ -623,8 +623,6 @@ InstallSubsetMaintenance( IsIntegralRing,
 
 InstallTrueMethod( IsIntegralRing,
     IsRing and IsMagmaWithInversesIfNonzero and IsNonTrivial );
-InstallTrueMethod( IsIntegralRing,
-    IsUniqueFactorizationRing and IsNonTrivial );
```
fingolfin, line 627: But according to the documentation for `IsUniqueFactorizationRing`, this is correct! If we make this change, we also ought to change the documentation.

But I think the true cause is elsewhere. As I commented on #3975, the commit which introduced this regression is 0f9c10479318a4ae19ac18eb494e22b6fae29527 and it adds another implication which might be the root cause:
```gap
InstallTrueMethod(IsEuclideanRing, IsZmodnZObjNonprimeCollection and
    IsWholeFamily and IsRing);
```
Indeed, I overlooked that the documentation for `IsEuclideanRing` says: 
> A  ring  R  is  called  a  Euclidean  ring  if  it is an integral ring

Oops. So *that* implication is the culprit.

That said, I think it is unfortunate that we define euclidean rings as being integral. So right now I think the quick fix would be to remove the implication I added, but the better fix would be to change the documentation and implications for `IsEuclideanRing`.

I now have a meeting but can look at this later.

alex-konovalov: Oh, no, quick fix does not work: just removing the implication causes this:
```
Error, required filters [ "IsListOrCollection", "IsCollection", "IsFinite", "HasIsFinite", "IsWholeFamily", "HasIsWholeFamily", 
  "IsDuplicateFree", "HasIsDuplicateFree", "IsExtAElement", "CategoryCollections(IsExtAElement)", 
  "CategoryCollections(IsNearAdditiveElement)", "CategoryCollections(IsNearAdditiveElementWithZero)", 
  "CategoryCollections(IsNearAdditiveElementWithInverse)", "CategoryCollections(IsAdditiveElement)", 
  "IsExtLElement", "CategoryCollections(IsExtLElement)", "IsExtRElement", "CategoryCollections(IsExtRElement)", 
  "CategoryCollections(IsMultiplicativeElement)", "CategoryCollections(IsMultiplicativeElementWithOne)", 
  "CategoryCollections(IsMultiplicativeElementWithInverse)", "CategoryCollections(IsAssociativeElement)", 
  "CategoryCollections(IsAdditivelyCommutativeElement)", "CategoryCollections(IsCommutativeElement)", 
  "IsGeneralizedDomain", "IsMagma", "IsAssociative", "HasIsAssociative", "IsCommutative", "HasIsCommutative", 
  "IsGeneratorsOfSemigroup", "HasIsGeneratorsOfSemigroup", "IsNearAdditiveMagma", "IsNearAdditiveMagmaWithZero", 
  "IsNearAdditiveGroup", "IsAdditivelyCommutative", "HasIsAdditivelyCommutative", "IsLDistributive", 
  "HasIsLDistributive", "IsRDistributive", "HasIsRDistributive", "CategoryCollections(IsZmodnZObjNonprime)" ]
for 1st argument do not match a declaration of EuclideanDegree called from
<<compiled GAP code>> from GAPROOT/lib/oper1.g:338 in function InstallMethod called from
```

hulpke: As far as I can see, `IsIntegralRing` is what is usually called integral domain (but domain  has a technical meaning in GAP). I thus do not see the problem with Euclidean rings being defined to be integral, respectively why should Z/6Z be Euclidean?

fieker:
On Wed, Apr 22, 2020 at 01:19:08PM -0700, Alexander Hulpke wrote:
> @hulpke commented on this pull request.
> 
> 
> 
> > @@ -623,8 +623,6 @@ InstallSubsetMaintenance( IsIntegralRing,
>  
>  InstallTrueMethod( IsIntegralRing,
>      IsRing and IsMagmaWithInversesIfNonzero and IsNonTrivial );
> -InstallTrueMethod( IsIntegralRing,
> -    IsUniqueFactorizationRing and IsNonTrivial );
> 
> As far as I can see, `IsIntegralRing` is what is usually called integral domain (but domain  has a technical meaning in GAP). I thus do not see the problem with Euclidean rings being defined to be integral, respectively why should Z/6Z be Euclidean?
> 
Because Z/6Z has a Euclidean function and a Euclidean division. You can
do HNF, SNF, ... over Z/nZ since it is Euclidean.
Its a matter of your taste weather you allow zero divisors in euclidean
rings or not.
Some make the distinctionbetween Euclidean Rings and Euclidean Domains
for that purpose...
Along the same lines: all ideals are principal, so it might be a
principal ideal ring...

lib/ring.gd
fingolfin: As @fieker explained below, I'd turn the question around: I see now reason why one should restrict "euclidean rings" to be integral, this is arbitrary.

hulpke:
@fieker Thanks. I think the problematic implication (and definition implying integral) comes from the textbook definition of Euclidean Domain.

I suggest we change the definition of `IsEuclideanRing` and remove these implications, but to avoid confusion also introduce `IsEuclideanDomain` or `IsEuclideanIntegralRing` (and ,if we have one of them, also a distinction between PID and PIR) and then can have the implication for the domain case. There might be similar "Domain" assumptions lurking in `EuclideanRing` routines or implications.

PR #3977, opened 2020-04-22
Fix line breaks in manual.lab files

frankluebeck:
Very long section headers could lead to unwanted line breaks in
entries in manual.lab files (only needed by old-style package
documentation to resolve references).

Closes #3969


fingolfin:
Backported to stable-4.11 in commit 20901d533

PR #3978, opened 2020-04-22
Use locale variables in accordance with POSIX requirements.

rbehrends:
Per the [POSIX spec](https://pubs.opengroup.org/onlinepubs/7908799/xbd/envvar.html), the `LC_ALL` environment variable should override all other locale settings if set. However, in `lib/init.g`, `LC_CTYPE` had priority over `LC_ALL`. This change checks these variables in the proper order.

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -381,11 +381,11 @@ CallAndInstallPostRestore( function()
       env := GAPInfo.SystemEnvironment;
     fi;
     enc := fail;
-    if IsBound(env.LC_CTYPE) then
-      enc := env.LC_CTYPE;
-    fi;
-    if enc = fail and IsBound(env.LC_ALL) then
+    if IsBound(env.LC_ALL) then
       enc := env.LC_ALL;
+    fi;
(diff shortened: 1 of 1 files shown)


fingolfin:
@rbehrends could you clarify whether this fixes a  bug? and if so, which? can a test case be added? Should this be backported to 4.11? Feel free to set relevant labels, too :-)

rbehrends:
The bug is that the current behavior does not match what POSIX says the environment variables should do, as noted in the original comment. I thought about filing a separate issue, but I figured that the change was easy enough and any discussion could take place here.

Primary concern would be that some software relies on the current behavior, e.g. where `LC_CTYPE` and `LC_ALL` are set to different values.

Adding a test case is not trivial, as this is related to what environment GAP is executed with and cannot easily be tested from within GAP.

fingolfin:
So, this is not fixing a specific issue that you run into, but rather is "just" making the code do the right thing. Which is of course perfectly fine and good, I just wanted to understand the context. Thanks!

PR #3980, opened 2020-04-22
FIX: GCD for rational polynomials

hulpke:
Use of modular results by Chinese remainder. Code was included in a wrong way
with missing argument that led to syntax error. Rare case, as this situation usually can be handled with by heuristics.

This fixes an error reported by Willem de Graaf


fingolfin:
Backported to stable-4.11 in commit e261ac3c5

PR #3986, opened 2020-04-26
Add some missing methods for machine floats (Sinh, Cosh, Tanh, Asinh, Acosh, Atanh, CubeRoot, Erf, Gamma)

### LINKED ISSUE #3983: Asinh/Acosh/Atanh have no methods for machine floats
### Observed behaviour

```gap
gap> Asinh( 0. );
Error, no 1st choice method found for `Asinh' on 1 arguments at /Users/mo/software/gap/build/gap-4.11.0/lib/methsel2.g:249 called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
 called from read-eval loop at *stdin*:11
brk>
```

### Expected behaviour
```gap
gap> Asinh( 0. );
0.
```

```gap
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 └───────┘   Architecture: x86_64-apple-darwin19.3.0-default64-kv7
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2019.09.04, AutPGrp 1.10.2, Browse 1.8.8, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, CTblLib 1.2.2,
             FactInt 1.6.3, FGA 1.4.0, Forms 1.2.5, GAPDoc 1.6.3, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4, LAGUNA 3.9.3, orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, PrimGrp 3.4.0,
             RadiRoot 2.8, recog 1.3.2, ResClasses 4.7.2, SmallGrp 1.4.1, Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.5, utils 0.69
```

fingolfin:
- A bunch of operations for floating point numbers had no implementation for
  machine floats, even though C99 provides suitable functions.

- Most of the math functions we checked for in configure are actually
  guaranteed to be there by C99. The sole exception is exp10, which is a
  non-standard extension.

- Add a bunch of super-trivial tests to ensure those new functions are
  actually callable. These are not good tests, just a bare minimum.

Resolves #3983


tst/testinstall/float.tst
```diff
@@ -628,20 +628,79 @@ gap> Sin(0.);
...
+[ 0., 0 ]
+gap> LdExp(0.,0);
+0.
+gap> Norm(0.);
+0.
+gap> SinCos(0.);
+[ 0., 1. ]
+gap> Erf(0.);
+0.
+gap> #Zeta(0.);     # TODO: not implemented for machine floats
```
fingolfin, line 701: Actually as far as I can tell, not even the `float` package implements this, for any float type -- correct, @laurentbartholdi ?

So perhaps we should just remove this, there seems little value in keeping an operation with no methods around?

laurentbartholdi: Well, I just followed the standard (I don't remember where I read it, though); so in my opinion it's best to leave it there, so as to reserve the word "Zeta".

fingolfin: @laurentbartholdi could you please clarify which "standard" you are referring to here?

laurentbartholdi:
No, sorry :)
I just remember having looked up something and copied it into the file, but
can't locate it anymore. A quick Google search brings up

Library TR1 ISO/IEC TR 19768:2007, then published as an independent ISO
standard, ISO/IEC 29124:2010

which seems never to have been properly implemented, except perhaps in
C++17.

fingolfin:
Well, C99 implements most of the functions you added to GAP, except for `zeta`. And also `gamma` was renamed to `tgamma` and with a companion `lgamma` that apparently often is more useful. See https://en.cppreference.com/w/c/numeric/math

I looked a bit around and found no standard the includes zeta, other than indeed C++17 which seems to include everything and the kitchen sink, but that's not something we can reasonably aim for :-)

PR #3989, opened 2020-04-27
Misc changes

fingolfin:
Two minor commits, see commit messages

### DIFF
--- a/src/gap_version.h
+++ b/src/gap_version.h
@@ -17,7 +17,7 @@
 **
 *V  SyKernelVersion . . . . . . . . . . . . . . . . . . kernel version number
 **
-**  'SyKernelVersion' is something like "4.12.2" or "4.12.dev".
+**  'SyKernelVersion' is something like "4.37.2" or "4.37dev".
 */
 extern const char * SyKernelVersion;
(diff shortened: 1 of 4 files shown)


src/gap_version.h
```diff
@@ -17,7 +17,7 @@
 **
 *V  SyKernelVersion . . . . . . . . . . . . . . . . . . kernel version number
 **
-**  'SyKernelVersion' is something like "4.12.2" or "4.12.dev".
+**  'SyKernelVersion' is something like "4.37.2" or "4.37dev".
```
wilfwilson, line 20: Did you deliberately remove the `.` before `dev`?

fingolfin: Good catch, but: Yes! This was actually wrong before, we are e.g. using "4.12dev" as bae version on master right now. I suppose we might consider using `4.12-dev` or so instead (that seems to better match prevalent conventions). and of course we could also improve our version compare code to support more "special" versions like `4.37-pre` or `4.37-rc1`, as described in https://semver.org). But, err, that kinda goes beyond this PR, sorry for getting distracted ;-)

PR #3990, opened 2020-04-27
kernel: restrict STATE(In) access to io.c

fingolfin:
With this, I finally reached my "secret goal" of removing all uses of `STATE` from `scanner.c`. The only remaining uses are inside `SyntaxErrorOrWarning`, to access `STATE(NrError)` and `STATE(NrErrLine)`. I am going to get those eventually, too ;-).

Of course this in turn has its own motivation: Not relying on global state makes it easier to reason about the program logic, and we made great strides in this regard already.
And then there is this pipe dream of mine, where `scanner.c` is completely stand-alone and usable in other projects. Not sure we'll ever quite get there, but we certainly are MUCH closer to that than we used to be a few years ago.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -294,7 +294,15 @@ Char PEEK_NEXT_CHAR(void)
 
 Char PEEK_CURR_CHAR(void)
 {
-    return *STATE(In);
+    Char c = *STATE(In);
+
+    // if no character is available then get one
+    if (c == '\0') {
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
This looks good to me. I wanted to check the moving around of code was OK, but I believe there won't be any way to tell the difference.

PR #3993, opened 2020-04-28
gac: *really* try to remove gactmp

fingolfin:
There can be stray files in there for whatever reason. Let's err
on the safe side here, instead or risking to run into errors in this
cleanup step.

Inspired by @jamesjr's patch at  https://src.fedoraproject.org/rpms/gap/blob/master/f/gap-gac.patch (that patch does more, but the other things it does are related to `make install` work, and will be in another PR)

### DIFF
--- a/cnf/gac.in
+++ b/cnf/gac.in
@@ -478,8 +478,6 @@ if [ $comp_howfar = "link" ]; then
 fi
 
 # Remove temporary directory.
-# We may assume it is empty at this stage.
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
Looks reasonable, I don't see any reason to do it the other way (other than paranoia)

fingolfin:
Backported to stable-4.11 in commit ba39930b5

PR #3996, opened 2020-04-28
kernel: move NrError into ScannerState

fingolfin:
UPDATE: This has now been slimmed down to a single commit (thanks to a ton of other PRs that are now merged), and is ready for review!

This is work in progress on overhauling the error handling mechanism in the GAP kernel. Here, I am trying to untangle the general error handling from the special case of handling errors in the scanner/reader (resp. interpreter/coder), by adding two new macro `GAP_TRY` and `GAP_CATCH` (might also add a `GAP_THROW` later) for use in general code; and then moving `NrError` out of the global GAP state into `ScannerState`. This makes reasoning about the code *much* easier.

But first we need to convince ourselves that this change in itself is correct and safe! This is the difficult part. And while the tests passed locally for me, I am not yet convinced that the changes are OK, as our test suite for error handling is not that comprehensive to start with. So we'll have to very carefully reason about this. I'll try to through every place I touched which uses try/catch and try to discuss for each (via comments here on this PR) why I think it is "safe". Well, after I convinced myself that it is... as I said, I am not yet convinced ;-)). I might also write some more tests.

I might also write a complementing PR or two which adds assertions to test hypothesis I have formed about this code, to verify whether they are true (at least for our test suites).

In the meantime, any "evil" ideas and suggestions for breaking what I did here are welcome ;-).

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1597,7 +1597,6 @@ void InitializeGap (
              (Bag *)(((UInt)pargc / C_STACK_ALIGN) * C_STACK_ALIGN),
              C_STACK_ALIGN);
 
-    GAP_ASSERT(STATE(NrError) == 0);
     STATE(ThrownObject) = 0;
     STATE(UserHasQUIT) = 0;
     STATE(UserHasQuit) = 0;
--- a/src/stats.h
+++ b/src/stats.h
@@ -135,11 +135,7 @@ extern  void            (* PrintStatFuncs[256] ) ( Stat stat );
  **
  *F  ClearError()  . . . . . . . . . . . . . .  reset execution and error flag
  *
- * FIXME: This function accesses NrError which is state of the scanner, so
- *        scanner should have an API for this.
- * 
  */
-
 void ClearError(void);
 
 
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -55,7 +55,6 @@ typedef struct GAPState {
     jmp_buf ReadJmpError;
 
     /* From scanner.c */
-    UInt   NrError;
     UInt   NrErrLine;
 
     char Prompt[80];
--- a/src/hpc/threadapi.c
+++ b/src/hpc/threadapi.c
@@ -438,7 +438,6 @@ static void ThreadedInterpreter(void * funcargs)
     int i;
 
     // initialize everything and begin a fresh execution context
-    GAP_ASSERT(STATE(NrError) == 0);
     STATE(ThrownObject) = 0;
     SWITCH_TO_OLD_LVARS(STATE(BottomLVars));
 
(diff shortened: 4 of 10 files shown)


rbehrends:
I like the general concept; at first I wondered if something like `if (cond) GAP_TRY { ... }` might create problems, but C does not allow a declaration in an if clause that's not a block and will throw an error, so we're good there.

I think that to test this seriously, we'd have to write some actual tests that target the new constructs as part of `tst/testlibgap` or `tst/testkernel`.

src/gap.c
```diff
@@ -1684,7 +1684,7 @@ void InitializeGap (
                     " script instead.", 0, 0);
             }
       }
-      CATCH_ERROR {
+      GAP_CATCH {
```
fingolfin, line 1687: This change is correct because in case of a failure, we `Panic`.

src/gap.c
```diff
@@ -1649,9 +1648,10 @@ void InitializeGap (
         /* Call POST_RESTORE which is a GAP function that now takes control, 
            calls the post restore functions and then runs a GAP session */
         if (POST_RESTORE != 0 && IS_FUNC(POST_RESTORE)) {
-          TRY_IF_NO_ERROR {
+          GAP_TRY {
             CALL_0ARGS(POST_RESTORE);
           }
+          GAP_CATCH {}
```
fingolfin, line 1654: This change is correct because in case of a failure, it doesn't matter whether we `NrError` to a non-zero value or: before we enter a REPL, we'll call `ClearError` anyway.

fingolfin:
@ChrisJefferson @rbehrends This has now been slimmed down to a single commit (thanks to a ton of other PRs that are now merged), and is ready for review!

src/stats.h
```diff
@@ -135,11 +135,7 @@ extern  void            (* PrintStatFuncs[256] ) ( Stat stat );
  **
  *F  ClearError()  . . . . . . . . . . . . . .  reset execution and error flag
  *
- * FIXME: This function accesses NrError which is state of the scanner, so
- *        scanner should have an API for this.
- *
```
fingolfin, line 140: It's silly but it *really* feels satisfying to get rid of this FIXME :-)

rbehrends:
>  Hence I am confident that this PR is correct.

Yeah, I understood that and actually went and worked my way through the changes. I just wanted to clarify that I *think* that, but just am not 100% sure (only 95% or so), so you don't understand my approval as "I'm completely sure that this works."

fingolfin:
@rbehrends oh sure, of course -- I'd never expert a reviewer to be "100% sure" (heck, I should never claim I am 100% sure, there's always a chance one missed something) :-)

PR #3997, opened 2020-04-28
kernel: verify some invariants in debug mode

fingolfin:
- NrError is always 0 in a fresh GAP state;
- NrError is always 0 when entering ReadEval{Command,File}
- verify some invariants in ReadEvalError instead of "enforcing" them

This is motivated by PR #3996: I want to test some hypothesis I formed about the kernel code. Locally this passes the test suite. Of course that doesn't *prove* I am right, but it at least supports my theory. Let's see if it still works with the full test run here on Travis.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -274,7 +274,10 @@ Stat NewStatOrExpr (
         bodySize = CS(OffsBody);
     while (bodySize < CS(OffsBody))
         bodySize *= 2;
+    GAP_ASSERT(STATE(PtrBody) == PTR_BAG(body));
     ResizeBag(body, bodySize);
+    // resize a bag can change its address, even without a GC taking place;
+    // so we must update PtrBody here
     STATE(PtrBody) = PTR_BAG(body);
 
     /* enter type and size                                                 */
--- a/src/read.c
+++ b/src/read.c
@@ -2514,6 +2514,10 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
     int                 lockSP;
 #endif
 
+    // invariant: before any call to ReadEvalCommand, there must have been
+    // a call to ClearError, and thus NrError == 0 holds
+    GAP_ASSERT(STATE(NrError) == 0);
+
     ReaderState reader;
     ReaderState * volatile rs = &reader;
     memset(rs, 0, sizeof(ReaderState));
@@ -2633,6 +2637,10 @@ UInt ReadEvalFile(Obj * evalResult)
     volatile int        lockSP;
 #endif
 
+    // invariant: before any call to ReadEvalFile, there must have been
+    // a call to ClearError, and thus NrError == 0 holds
+    GAP_ASSERT(STATE(NrError) == 0);
+
     ReaderState reader;
     ReaderState * volatile rs = &reader;
     memset(rs, 0, sizeof(ReaderState));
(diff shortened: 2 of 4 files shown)


src/read.c
```diff
@@ -2514,6 +2514,10 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
     int                 lockSP;
 #endif
 
+    // invariant: before any call to ReadEvalCommand, there must have been
+    // a call to ClearError, and thus NrError == 0 holds
+    GAP_ASSERT(STATE(NrError) == 0);
```
fingolfin, line 2519: Note that I verified this by looking at all callsites for `ReadEvalCommand`.

src/read.c
```diff
@@ -2633,6 +2637,10 @@ UInt ReadEvalFile(Obj * evalResult)
     volatile int        lockSP;
 #endif
 
+    // invariant: before any call to ReadEvalFile, there must have been
+    // a call to ClearError, and thus NrError == 0 holds
+    GAP_ASSERT(STATE(NrError) == 0);
```
fingolfin, line 2642: There is exactly one place calling `ReadEvalFile` and there this is true.

src/read.c
```diff
@@ -2723,8 +2731,8 @@ UInt ReadEvalFile(Obj * evalResult)
 */
 void ReadEvalError(void)
 {
-    STATE(PtrBody)  = PTR_BAG(BODY_FUNC(CURR_FUNC()));
-    STATE(PtrLVars) = PTR_BAG(STATE(CurrLVars));
+    GAP_ASSERT(STATE(PtrBody) == PTR_BAG(BODY_FUNC(CURR_FUNC())));
+    GAP_ASSERT(STATE(PtrLVars) == PTR_BAG(STATE(CurrLVars)));
```
fingolfin, line 2735: Actually, I just saw that `CURR_FUNC` actually contains this check:
```C
EXPORT_INLINE Obj CURR_FUNC(void)
{
    GAP_ASSERT(IS_LVARS_OR_HVARS(STATE(CurrLVars)));
    GAP_ASSERT(STATE(PtrLVars) == PTR_BAG(STATE(CurrLVars)));
    return FUNC_LVARS_PTR(STATE(PtrLVars));
}
```
So clearly at least the second assertion must have always been satisfied.

Moreover, we do this in `vars.c`:
```
static void VarsAfterCollectBags(void)
{
  if (STATE(CurrLVars))
    {
      STATE(PtrLVars) = PTR_BAG( STATE(CurrLVars) );
      STATE(PtrBody)  = PTR_BAG( BODY_FUNC( CURR_FUNC() ) );
    }
}
```
This is called by a GC, so that can't mess up `PtrBody`. So really the only way `PtrBody` could be "wrong" is if we modify it then call `ReadEvalError` *before* any GC has a chance of happening. So where is `PtrBody` changed? First, by `SWITCH_TO_*_LVARS` but those three functions preserve the invariants. That leaves a single other place modifying it, in `NewStatOrExpr`:
```C
...
    Obj body = BODY_FUNC(CURR_FUNC());
    UInt bodySize = SIZE_BAG(body);
    if (bodySize == 0)
        bodySize = CS(OffsBody);
    while (bodySize < CS(OffsBody))
        bodySize *= 2;
    ResizeBag(body, bodySize);
    STATE(PtrBody) = PTR_BAG(body);
...
```
As you can see the invariant is preserved, and in fact the assignment unnecessary. I'll change that into an assertion, too.

src/code.c
```diff
@@ -275,7 +275,7 @@ Stat NewStatOrExpr (
     while (bodySize < CS(OffsBody))
         bodySize *= 2;
     ResizeBag(body, bodySize);
-    STATE(PtrBody) = PTR_BAG(body);
+    GAP_ASSERT(STATE(PtrBody) == PTR_BAG(body));
```
fingolfin, line 278: OK, so here I was wrong: This does *not* hold...

fingolfin: It took me some time to understand this; but of course `ResizeBag` can change the bag pointer *even if no GC takes place*, simply because there is not enough space left at the current location of the object. Stupid me *facepalm*

I moved the assertion now to before the ResizeBag, and restored the assignment here.

src/read.c
fingolfin: Actually the assignment is *not* unnecessary, simply because ResizeBag can of course move a bag even without a GC taking place. But I now added an assertion *before* the `ResizeBag` to verify the invariant.

PR #3998, opened 2020-04-30
kernel: avoids some uses of TRY_IF_NO_ERROR

fingolfin:
In `FuncPrint`, we catch an exception only to immediately re-throw it.
Clearly this is pointless.

Three places using `TRY_IF_NO_ERROR` simply want to perform a call to a
GAP function but ignoring any error this might raise; we change them to
use `Call0ArgsInNewReader` instead. Note that this arguably even might
fix bugs in some corners, as we now call `ClearError` which the previous
code did not.

@rbehrends please take a look at the HPC specific bits, in case I missed some subtlety about them in my analysis?

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -1649,9 +1649,7 @@ void InitializeGap (
         /* Call POST_RESTORE which is a GAP function that now takes control, 
            calls the post restore functions and then runs a GAP session */
         if (POST_RESTORE != 0 && IS_FUNC(POST_RESTORE)) {
-          TRY_IF_NO_ERROR {
-            CALL_0ARGS(POST_RESTORE);
-          }
+          Call0ArgsInNewReader(POST_RESTORE);
         }
 
         return;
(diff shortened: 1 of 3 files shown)


rbehrends:
I'm not entirely sure if the HPC-GAP parts should be changed (though these functions are not currently used from the GAP standard library, so it's not like that matters much). Hash locks that are still locked when an error occurs are cleared in `CALL_WITH_CATCH()`, along with region locks, but not when caught by `TRY_IF_ERROR`. The purpose here is to ensure that the locks are properly unlocked in case some other code needs a hash lock before the current lock is cleared. I don't think that happens, but I wonder if that may still not make the code more brittle.

fingolfin:
@rbehrends I don't understand your comment at all. I assume you are talking about `FuncHASH_SYNCHRONIZED` and `FuncHASH_SYNCHRONIZED_SHARED`? Note that *existing* code already uses `TRY_IF_ERROR`, and so does the replacement code -- only that the `TRY_IF_ERROR` call is now hidden inside `Call0ArgsInNewReader`. Both in the old and new code, we then call `HashUnlock` and `HashUnlockShared` after the call, regardless of whether it succeeded or not. So my claim is that there is actually no semantic change. You seem to disagree, but I don't understand where you suspect an issue, so I can't really say more until you clarify. (I also don't see how you think `CALL_WITH_CATCH` factor in here -- I am not using that?)

rbehrends:
Oops, I missed that `Call0ArgsInNewReader()` does a `TRY_IF_ERROR` already. My mistake. In that case, no, I don't see a problem.

PR #4000, opened 2020-04-30
NormalClosure: accept list of normal generators as second argument

fingolfin:


lib/grp.gi
```diff
@@ -2914,6 +2914,12 @@ function(G,U)
   return U;
 end);
 
+InstallOtherMethod( NormalClosure, "generic method for a list of normal generators",
```
PaulaHaehndel, line 2917: This option of calling the method then is undocumented.
And I first was confused by the words "normal generators".

fingolfin: Documenting this definitely is a good idea.

fingolfin: @PaulaHaehndel done

PR #4002, opened 2020-05-02
Improve explanation of -c option

ssiccha:


ssiccha:
Feel free to suggest improvements of my formulation. :)

lib/system.g
```diff
@@ -113,7 +113,10 @@ BIND_GLOBAL( "GAPInfo", rec(
       rec( long := "bare", default := false,
            help := [ "Attempt to start GAP without even needed packages (developer tool)" ] ),
       ,
-      rec( short:= "c", default := "", arg := "<expr>", help := [ "execute the given expression"] ),
+      rec( short:= "c", default := "", arg := "<expr>",
+      help := [ "Execute the given expression. The order in which files are
+                read and expressions are executed is the order in which they
+                are given on the command line."] ),
```
fingolfin, line 119: I think it would be more appropriate to put this into the manual; and/or a man page for GAP (perhaps one could use that from Debian or Fedora as basis)

fingolfin: Oh, and this code is syntactically wrong, so it won't even load.

fingolfin:
@ssiccha ping?

ssiccha:
Sorry, I'd get back to this if I have the time to do it properly. I'm closing this for now.

PR #4004, opened 2020-05-03
kernel: make ViewObjHandler more robust

fingolfin:
READ_TEST_OR_LOOP used to invoke SWITCH_TO_OLD_LVARS after invoking
ViewObjHandler, but I don't see how that could ever be necessary; and if
it is necessary, why it wouldn't be required for the other call to
ViewObjHandler the main REPL.

So let's replace it by an assertion that verifies it is not necessary,
then let's run some more tests.

If this PR passes all tests, let's not merge it; instead, we can simply remove the checks in question.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -374,14 +374,17 @@ static void READ_TEST_OR_LOOP(Obj context)
     UInt                type;
     UInt                oldtime;
     UInt                dualSemicolon;
+    UInt                oldPrintObjState;
 
     /* get the starting time                                               */
     oldtime = SyTime();
+    oldPrintObjState = SetPrintObjState(0);
 
     /* now do the reading                                                  */
     while ( 1 ) {
 
         /* read and evaluate the command                                   */
+        SetPrintObjState(0);
         ClearError();
         Obj evalResult;
         type = ReadEvalCommand(context, &evalResult, &dualSemicolon);
@@ -397,9 +400,7 @@ static void READ_TEST_OR_LOOP(Obj context)
 
             /* print the result                                            */
             if ( ! dualSemicolon ) {
-                Bag currLVars = STATE(CurrLVars); /* in case view runs into error */
                 ViewObjHandler( evalResult );
-                SWITCH_TO_OLD_LVARS(currLVars);
             }
         }
 
@@ -416,6 +417,7 @@ static void READ_TEST_OR_LOOP(Obj context)
 
     }
 
+    SetPrintObjState(oldPrintObjState);
     ClearError();
 }
 
--- a/src/gap.c
+++ b/src/gap.c
@@ -128,13 +128,12 @@ UInt ViewObjGVar;
 
 void ViewObjHandler ( Obj obj )
 {
-  volatile Obj        func;
-
-  /* get the functions                                                   */
-  func = ValAutoGVar(ViewObjGVar);
+  // save some values in case view runs into error
+  volatile Bag  currLVars   = STATE(CurrLVars);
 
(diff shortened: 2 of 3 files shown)


src/streams.c
```diff
@@ -396,9 +396,11 @@ static void READ_TEST_OR_LOOP(Obj context)
 
             /* print the result                                            */
             if ( ! dualSemicolon ) {
+#ifdef GAP_KERNEL_DEBUG
                 Bag currLVars = STATE(CurrLVars); /* in case view runs into error */
+#endif
                 ViewObjHandler( evalResult );
-                SWITCH_TO_OLD_LVARS(currLVars);
+                GAP_ASSERT(currLVars == STATE(CurrLVars));
```
fingolfin, line 403: so, I am still worried about this... There is similar code in `CALL_WITH_CATCH`. If I modify that in a similar way, I *can* trigger the assertion:
```
gap> CALL_WITH_CATCH(x->x,[1..3]);
Error, Function: number of arguments must be 1 (not 3)
brk>
Assertion failed: (currLVars == (ActiveGAPState()->CurrLVars)), function CALL_WITH_CATCH, file ../../src/error.c, line 27
...
(lldb) p currLVars
(Obj) $0 = 0x0000100000010fe8
(lldb) p MainGAPState.CurrLVars
(Bag) $1 = 0x00001000005d1f28
(lldb) p PARENT_LVARS(MainGAPState.CurrLVars)
(Obj) $2 = 0x0000100000010fe8
```
So the active lvars at that point have the original ones as parent. I think this come from the break loop? Makes me wonder if `SHELL` should save & restore lvars... but are there other places this could happen?

Perhaps if a ViewObj method triggers an Error, could this assertion here be triggered? I *think* I tried that, but I should try again to be sure?

Hrm

fingolfin: Aha. Turns out I tested this, but only on a break loop, not in a `.tst` file. And it turns out, I can trigger the assertion then: if I put this into a file `foo.tst`:
```
gap> InstallMethod(ViewObj, [IsStringRep], 100,
>     function(s)
>         if s="foobar" then Error("oops"); fi;
>         TryNextMethod();
>     end);
gap> "foobar";
Error, oops at *stdin*:1
```
Then I see this:
```
gap> Test("foo.tst");
# line 6 of 6 (100%)Assertion failed: (currLVars == (ActiveGAPState()->CurrLVars)), function READ_TEST_OR_LOOP, file ../../src/streams.c, line 403.
```

fingolfin: So we must not merge this as-is. Now I tried and patched the kernel function `Shell` to perform the same assertion check, and see, and can trigger the assertion there with the same code snippet (now pasted into a REPL).

Need to analyze this further to see why not all hell breaks loose. Probably we are restoring CurrLVars elsewhere, but where? Need to work on other stuff now, though.

fingolfin:
Update this PR: added a test case, and modified ViewObjHandler to be "safe"

fingolfin:
The errors in CI look as if the new tests reveal actual breakage -- we are perhaps still not resetting enough. Good! I mean, not the breakage, but that we discovered it this way -- once more, it pays off to question any code in the kernel I understand, and refactor it until I feel I understand it (better).

Tests hopefully are fixed now and this PR is ready for review.

fingolfin:
I discovered there is at least one more variant of this issue: not just the `ViewObj` which the REPL calls can cause issues, but also direct calls to ViewObj or PrintObj, at least in a "test REPL" for `.tst` files. I added a fix for that, and also refined the tests so that `tst/testinstall/kernel/gap.tst` on its own triggers the error.

PR #4007, opened 2020-05-03
kernel: refactor CALL_WITH_CATCH to use GAP_TRY

fingolfin:
Also get rid of the `GAP_ASSERT(currStat == BRK_CALL_TO())` left over from a previous refactoring (I think we have established beyond doubt at this point that this invariant holds)

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -30,6 +30,7 @@
 #include "stringobj.h"
 #include "sysjmp.h"
 #include "sysstr.h"
+#include "trycatch.h"
 #include "vars.h"
 
 #ifdef HPCGAP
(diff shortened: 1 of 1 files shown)


src/error.c
```diff
@@ -250,37 +249,15 @@ Obj CALL_WITH_CATCH(Obj func, volatile Obj args)
...
     Region * savedRegion = TLS(currentRegion);
 #endif
-    if (setjmp(STATE(ReadJmpError))) {
-        SET_LEN_PLIST(res, 2);
-        SET_ELM_PLIST(res, 1, False);
-        SET_ELM_PLIST(res, 2, STATE(ThrownObject));
-        CHANGED_BAG(res);
-        STATE(ThrownObject) = 0;
-        SWITCH_TO_OLD_LVARS(currLVars);
-        GAP_ASSERT(currStat == BRK_CALL_TO());
```
fingolfin, line 273: This is the removed assertion.

PR #4009, opened 2020-05-03
kernel: move time related code to new files gaptime.{c.h}

fingolfin:
This also gets rid of some pointless calls to Panic: even if getrusage
fails, we can recover, so no reason to crash the whole program.


src/gaptime.c
```diff
@@ -0,0 +1,390 @@
...
+**  'SyTime' returns the number of milliseconds spent by GAP so far.
+**
+**  Should be as accurate as possible,  because it  is  used  for  profiling.
+*/
+UInt SyTime(void)
+{
+    struct rusage buf;
+
+    if (getrusage(RUSAGE_SELF, &buf)) {
+        Panic("'SyTime' could not get time");
```
ChrisJefferson, line 55: Did you mean to leave this Panic?

fingolfin: fixed

ChrisJefferson:
Thanks. Just to confirm, I'm happy for this to be merged once the tests cycle.

PR #4012, opened 2020-05-06
BuildPackages.sh: adjust default GAP to GAPROOT

fingolfin:
The default value for GAP was based on the default value for GAPROOT; now it is
based on the actual GAPROOT we use in the script; so if --with-gaproot is used,
then this also adjusts GAP.

This is necessary for out-of-tree builds (I hope this will be the final patch needed for that ... *sigh*)

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -32,7 +32,6 @@ fi
 
 CURDIR="$(pwd)"
 GAPROOT="$(cd .. && pwd)"
-GAP="$GAPROOT/bin/gap.sh"
 COLORS=yes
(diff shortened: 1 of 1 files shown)


rbehrends:
So, to clarify that I'm understanding this correctly: if `$GAP` is set, then it will default to `$GAP`, otherwise it will default to the script relative to `$GAPROOT`?

fingolfin:
@rbehrends exactly, that's the idea. One can use `--with-gap=VALUE` to override `GAP`, too. Or have it exported to the environment.

fingolfin:
Backported to stable-4.11 in commit 8c6a4ee39

PR #4013, opened 2020-05-06
buildsys: rename gen/ dir to build/

fingolfin:


pedritomelenas:
Agree.

fingolfin:
Let's not merge this for now -- first PR #4026 should land; then this PR here should be adjusted to update `etc/ci.sh` (which will use `gen` in PR #4026)

PR #4015, opened 2020-05-07
Add IsWritablePathString (GAP) and SyIsWritablePath (C)

stevelinton:
# Description

Adds functions which attempt to check whether a pathname is suitable
for writing to -- either an existing writable file (not a directory)
or a non-existing name in a writable directory.

They are not perfect, they might be fooled by symlinks, and there is always a race condition, 
in that things may change between checking the files and using them. There is, I guess thus an argument that tests like this one are a bad idea and you should always deal with these issues
by checking the return codes when you try and open the file. However the motivation in this case is for a package that links to an upstream library (meatxe64) which basically has no error recovery pathways and segfaults if file creation fails (see a recent issue on github). 

## Text for release notes 

Add a new function `IsWritablePathString` which attempts to check whether the given pathname is either an existing a writable file, or a path where a new file could be created.

- [ X] proper formatting

- [ x] usage of relevant labels

- [x ] runnable tests

All the `IsXxxxxFile` and similar functions could use better tests.

- [x ] adequate pull request title
- [x ] well formulated text for release notes
- [ x] relevant documentation updates
- [ x] sensible comments in the code


doc/ref/files.xml
```diff
@@ -265,6 +265,19 @@ resp. <Ref Func="IsExecutableFile"/> to check such permissions.
...
 
+<ManSection>
+<Func Name="IsWritablePath" Arg='pathname'/>
+
+<Description>
+<Ref Func="IsWritablePath"/>
+returns <K>true</K> if either there is a writable file (not a directory) with
+the filename <A>pathname</A>
+or there is no such file, but one could be created by writing to it
+(ie the path names a non-existent file in a writable directory)
```
fingolfin, line 276:
```suggestion
(i.e., the path names a non-existent file in a writable directory)
```

src/streams.c
```diff
@@ -1232,6 +1232,21 @@ static Obj FuncIsDirectoryPathString(Obj self, Obj filename)
     return res == -1 ? False : True;
 }
 
+/****************************************************************************
+**
+*F  FuncIsWritablePath( <self>, <name> ) . . . .  is file <name> a directory
```
fingolfin, line 1237: that description is wrong, and the variable `name` does not match `filename` below

src/sysfiles.h
```diff
@@ -380,11 +380,22 @@ Int SyIsExecutableFile(const Char * name);
...
 **
-**  'SyIsDirectoryPath' returns 1 if the  file <name>  is a directory  and  0
+**  'SyIsDirectoryPath' returns 0 if the  file <name>  is a directory  and -1
 **  otherwise. <name> is a system dependent description of the file.
 */
 Int SyIsDirectoryPath(const Char * name);
 
+/****************************************************************************
+**
+*F  SyIsWritablePath( <name> ) . . . . . . . . .is <name> a valid output path
```
fingolfin, line 390:
```suggestion
*F  SyIsWritablePath( <name> ) . . . . . . . .  is <name> a valid output path
```

src/sysfiles.h
```diff
@@ -339,7 +339,7 @@ UInt SyExecuteProcess(Char * dir, Char * prg, Int in, Int out, Char * args[]);
 **
 *F  SyIsExistingFile( <name> )  . . . . . . . . . . . does file <name> exists
 **
-**  'SyIsExistingFile' returns 1 if the  file <name> exists and 0  otherwise.
+**  'SyIsExistingFile' returns 0 if the  file <name> exists and -1  otherwise.
```
fingolfin, line 342: Arrgh! OK, first off: good catch.

Secondly: This is a case were I very much think the documentation is right and code is wrong. Reading code that does `if (SyIsExistingFile(name)) { ... do stuff that assumes the files does NOT exist ... }` is just begging for somebody to get confused and screw it up.

So, I'd prefer if you didn't "fix" the comments here, and instead we fix those function (and all code calling them) to have sane return values which match the name of the function. In fact, I think we should use `BOOL`, `TRUE` and `FALSE` for them.

(To be clear, I am not asking you to make that fix which goes beyond what you wanted to do in this PR)

src/sysfiles.h
```diff
@@ -380,11 +380,22 @@ Int SyIsExecutableFile(const Char * name);
...
 **  otherwise. <name> is a system dependent description of the file.
 */
 Int SyIsDirectoryPath(const Char * name);
 
+/****************************************************************************
+**
+*F  SyIsWritablePath( <name> ) . . . . . . . . .is <name> a valid output path
+**
+**  'SyIsWritablePath' returns 0 if the  file <name>  is an existing writable
+*file
```
fingolfin, line 393: ?

src/sysfiles.c
```diff
@@ -3139,6 +3138,32 @@ Int SyIsDirectoryPath ( const Char * name )
...
+        return 0;
+    if (0 == SyIsExistingFile(name))
+        return -1;
+    Char * slash = rindex(name, '/');
+    if (!slash)
+        return SyIsWritableFile(".");
+    Char basename[slash - name + 1];
+    memcpy(basename, name, slash - name);
+    basename[slash - name] = '\0';
+    return SyIsWritableFile(basename);
```
fingolfin, line 3164: So I guess that means `SyIsWritableFile` allows checking whether a directory is writeable? Interesting, I would not have expected that from its name (but looking at its implementation, it's clear why this is).

But we don't guarantee this in the specs of this function, do we?

src/sysfiles.h
fingolfin: Looking closer, it's actually worse, as the return value of `SyIsExistingFile` actually does not allow distinguishing between "the file does not exist" and "some error occurred".

Honestly, I wonder why we even bother with these "abstraction layers". They don't have well-defined semantics, and each typically has only a single caller (`SyIsReadableFile` being the exception). So a hypothetical porting effort wouldn't really benefit from their existence... I'll think about submitting a PR to just merge them into their callers.

stevelinton:
I have split this into two commits. One adds the new features and the other adjusts the comments on existing functions to match the code. We could drop the second and add an issue to fix the code instead.

src/sysfiles.c
stevelinton: Given that it just wraps the POSIX function `access` on all systems I don't see much harm in just documenting this. Obviously we could have a separate `IsWriteableDirectory` or just call `access` directly.

Issue #3973, opened 2020-04-20
Makefile.gappkg improvements


fingolfin:
Just to make sure I won't forget about these:

- [ ] add support for user-supplied CFLAGS, CXXFLAGS, LDFLAGS (this may fix coverage reporting via Travis&Codecov for packages using it)
- [ ] add a version number into `Makefile.gappkg` which `configure` scripts can use to decide if the  copy of `Makefile.gappkg` in GAP is actually *newer* than the one bundled with the package (to this end, it should be a simple integer that is easy to extract with grep or awk or so)
- [ ] some packages combine `Makefile.gappkg` with `autoconf`; this requires some extra code right now in their `Makefile.in` files; think about whether we can somehow move this into `Makefile.gappkg` in a generic way
- [ ] think about ways to add support for `make install` support (this is a big one)
- [ ] update all packages that use `Makefile.gappkg` to the latest and greatest

fingolfin:
I actually implemented the second point (adding a version number & automatic selfupdate), but I think I won't submit it as a PR: The more I think about it, the more I believe it is better to do this manually. The one point going for the autoupdate is that it'll "happen" to package maintainers, who those perhaps will be urged (annoyed?) to update to new versions (otherwise the `release-gap-package` script won't run through). While that seems appealing on the surface (right now I need to prod package authors / submit lots of PRs to get an update out), I don't think it warrants the efforts, complexity, and possible source of serious annoyance (like, if a new version of the file is broken for some package, then that package shouldn't have to switch to it). And in general, it's OK to keep using an older version of it; if a new version adds compelling new features (e.g. `make install` support), that should be enough to either convince package authors to update by themselves, or us to submit PRs to them to get them to update.

PR #4016, opened 2020-05-07
etc/Makefile.gappkg: various improvements

fingolfin:
- built-in autoconf support for packages that need a complex configure script;
  this adds rules which re-run autoconf or autoheader as needed, and to regenerate
  various files automatically; sometimes this can be undesired, then it can be
  turned off by setting the (environment) variable MAINTAINER_MODE to "no"
- support assembler ".s" source files (via C compiler)
- honor user supplied CFLAGS, CXXFLAGS, LDFLAGS
- little hack to ensure that when using GAP 4.9, the GAP src directory is in
  the list of include paths (as it is automatically with GAP 4.10 and later),
  so that `#include "compiled.h"` works

This covers #3973 partially.

This should be backported to stable-4.11.

I'll also submit PRs to (more) packages using this file to convince them to update

### DIFF
--- a/etc/Makefile.gappkg
+++ b/etc/Makefile.gappkg
@@ -3,8 +3,8 @@
 # The build rules in this file are intended for use by GAP packages that
 # want to build a simple GAP kernel extensions. They are based on the
 # GAP build system, and require GNU make. To use this in your GAP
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.11 in commit bbaab827d

PR #4022, opened 2020-05-10
Remove locale specific tests

### LINKED ISSUE #2194: "\200" and "\377" show as "?" on SPARC Solaris 11

### LINKED ISSUE #3979: GAP tests are locale-dependent.
Several tests fail when run under the `C` locale.

Example:

```
LC_ALL=C LC_CTYPE=C ./gap tst/testinstall.g
```

This results in errors like this:

```
testing: /Users/behrends/develop/gap-dev/master/tst/testinstall/stringobj.tst
########> Diff in /Users/behrends/develop/gap-dev/master/tst/testinstall/strin\
gobj.tst:41
# Input is:
List("abcxyzABCXYZ\n\t !019?\377\000$", LowercaseChar);
# Expected output:
"abcxyzabcxyz\n\t !019?\377\000$"
# But found:
"abcxyzabcxyz\n\t !019?�\000$"
########
########> Diff in /Users/behrends/develop/gap-dev/master/tst/testinstall/strin\
gobj.tst:43
# Input is:
List("abcxyzABCXYZ\n\t !019?\377\000$", UppercaseChar);
# Expected output:
"ABCXYZABCXYZ\n\t !019?\377\000$"
# But found:
"ABCXYZABCXYZ\n\t !019?�\000$"
########
########> Diff in /Users/behrends/develop/gap-dev/master/tst/testinstall/strin\
gobj.tst:45
# Input is:
UppercaseString("abcxyzABCXYZ\n\t !019?\377\000$");
# Expected output:
"ABCXYZABCXYZ\n\t !019?\377\000$"
# But found:
"ABCXYZABCXYZ\n\t !019?�\000$"
########
########> Diff in /Users/behrends/develop/gap-dev/master/tst/testinstall/strin\
gobj.tst:47
# Input is:
LowercaseString("abcxyzABCXYZ\n\t !019?\377\000$");
# Expected output:
"abcxyzabcxyz\n\t !019?\377\000$"
# But found:
"abcxyzabcxyz\n\t !019?�\000$"
########
     372 ms (0 ms GC) and 24.4MB allocated for stringobj.tst
```

A similar problem could be [observed in the Homebrew tests](https://github.com/gap-system/homebrew-gap/issues/2).

fingolfin:
These caused issues in various places, but test nothing useful.

Resolves #2194
Resolves #3979

### DIFF
--- a/tst/testinstall/stringobj.tst
+++ b/tst/testinstall/stringobj.tst
@@ -38,14 +38,14 @@ gap> s;
 "Ilovepiespiespies"
 
 # Lowercase - Uppercase
-gap> List("abcxyzABCXYZ\n\t !019?\377\000$", LowercaseChar);
-"abcxyzabcxyz\n\t !019?\377\000$"
-gap> List("abcxyzABCXYZ\n\t !019?\377\000$", UppercaseChar);
-"ABCXYZABCXYZ\n\t !019?\377\000$"
-gap> UppercaseString("abcxyzABCXYZ\n\t !019?\377\000$");
-"ABCXYZABCXYZ\n\t !019?\377\000$"
-gap> LowercaseString("abcxyzABCXYZ\n\t !019?\377\000$");
-"abcxyzabcxyz\n\t !019?\377\000$"
+gap> List("abcxyzABCXYZ\n\t !019?\000$", LowercaseChar);
+"abcxyzabcxyz\n\t !019?\000$"
+gap> List("abcxyzABCXYZ\n\t !019?\000$", UppercaseChar);
+"ABCXYZABCXYZ\n\t !019?\000$"
+gap> UppercaseString("abcxyzABCXYZ\n\t !019?\000$");
+"ABCXYZABCXYZ\n\t !019?\000$"
+gap> LowercaseString("abcxyzABCXYZ\n\t !019?\000$");
+"abcxyzabcxyz\n\t !019?\000$"
 gap> UppercaseString("");
 ""
 gap> LowercaseString("");
--- a/tst/testinstall/strings.tst
+++ b/tst/testinstall/strings.tst
@@ -2,7 +2,7 @@
 ##
 ##  This file tests output methods (mainly for strings)
 ##
-#@local x, str
+#@local x, str, len
 gap> START_TEST("strings.tst");
 
 # FFE
@@ -40,24 +40,22 @@ gap> PrintString(x);
 "abc"
 gap> String(x);
 "abc"
-gap> x:="\0xFF";
-"\377"
-gap> PrintString(x);
-"\377"
-gap> ViewString(x);
-"\"\\377\""
+gap> "\0x4a";
+"J"
+gap> x:="\0x4A";
+"J"
 gap> x:="\0x42\0x23\0x10\0x10\0x10";
 "B#\020\020\020"
 gap> PrintString(x);
 "B#\020\020\020"
 gap> ViewString(x);
 "\"B#\\020\\020\\020\""
-gap> x:="A string with \0xFF Hex stuff \0x42 in it";
-"A string with \377 Hex stuff B in it"
+gap> x:="A string with Hex stuff \0x42 in it";
+"A string with Hex stuff B in it"
 gap> PrintString(x);
-"A string with \377 Hex stuff B in it"
+"A string with Hex stuff B in it"
 gap> ViewString(x);
-"\"A string with \\377 Hex stuff B in it\""
+"\"A string with Hex stuff B in it\""
 gap> x := "\n\t\c\\\"'";
 "\n\t\c\\\"'"
 gap> PrintString(x);
@@ -185,11 +183,19 @@ gap> PrintString(x);
 "'a'"
 gap> String(x);
 "'a'"
-gap> x:='\0x42';
+gap> '\0x42';
 'B'
-gap> x:='\0xFF';
+gap> '\102';
+'B'
+gap> '\0xFF';
 '\377'
-gap> x:='\0xab';
+gap> '\0xff';
+'\377'
+gap> '\0xFf';
+'\377'
+gap> '\0xfF';
+'\377'
+gap> '\0xab';
 '\253'
 
 # Huge strings


dimpase:
I am not 100% sure it is wise to remove these tests - I would have kept
them, but only tested that running them doesn't lead to a crash.

fingolfin:
@dimpase and why is that? What do you believe is tested by the remover parts of the tests which is not tested by the remaining tests?

dimpase:
I meant to say that other tests don't test printing "unprintable"/locale-dependent  characters. I do not know the underlying code, but it is easy to imagine an implementation that does not do a correct thing.

Testing platforms such as Python's doctests allow to test for wildcards in the output (this is `...` in [doctests](https://docs.python.org/3/library/doctest.html)), perhaps GAP should do this too.

fingolfin:
@dimpase GAP has no special code or any kind of special behavior for "unprintable"/locale-dependent characters. Thus, it is unclear what you think this "tests" that is not tested by printing regular printable letters like "a". It is "easy to imagine" a lot of things, but I am interested in cold hard facts, not fantasy :-).

I did at first think about coming up with ways to allow us to leave in those `\0xff` chars, but then I realized that "testing" for them really serves no purpose I can think of...

dimpase:
I'm saying that by not testing for invalid input you decrease the coverage of the testsuite (@coveralls agrees withg me :-)). One reason for testing for invalid input is to test the environment GAP runs in, which might have quirks beyond anyone's imagination, leading to segfaults.

It's much easier to deal with the sutuation where a user reports a test failure, than with a user report saying "help, I get a segfault, while all the GAP tests pass".

fingolfin:
> I'm saying that by not testing for invalid input you decrease the coverage of the testsuite

There is no "invalid input" in the tests touched by this PR.

> (@coveralls agrees withg me :-)).

You do have a point there: I must admit that code coverage was so badly broken for a long time (tons of random fluctuations even in PRs which don't change any code) that I started to ignored it. But I just had a look and it is reasonable now (at least on coveralls; codecov seem to just have stopped working for us; which is a pity as I find its UX much better than that of coveralls)

Anyway, it turns out that we *do* have a special case for ViewObj of chars and strings equal to/ containing `0xFF` (and other similar chars), just not in the kernel, but in the library; and that moreover, the one for strings is replaced by a method in GAPDoc (which of course is always loaded, so the overloaded method is never called). And GAPDoc is what brings in the locale dependency... 

All the (limited) unicode handling in GAP is implemented in GAPDoc, so a core component of the system is maintained outside of it; I find this problematic. But for now, this is how it is; testing GAPDoc is out of scope for these CI tests (it would make sense to test it here, but since it is developed externally to this repository, writing tests for it here raises all kinds of concerns, so I'd rather not deal with it right this moment).

I have reinstated the tests for `IsChar`, though, as here our functions *are* called (as a GAP char is a single byte, not a UTF-8 multi-byte "character"). There is no way to test the GAP library ViewObj method for strings, though, as explained above.


> One reason for testing for invalid input is to test the environment GAP runs in, which might have quirks beyond anyone's imagination, leading to segfaults.

Agreed (which is why I added tons of tests for invalid inputs over the past couple years, as did Markus and Chris; before us, these were almost never tested).

But also irrelevant for this PR, as it does not test "invalid inputs". OK, perhaps it tests "unusual" inputs. But look, testing always is a delicate balance -- there is an essentially infinite number of inputs you "should" test, but only a finite amount of time for actual tests. So we have to weigh tests by probability. Yes, that's flawed, but so is testing in general: In theory I should test printing of *every* integer, but i can't. So instead we limit ourselves to a certain set of numbers which we think are "interesting" in that they consist of edge cases etc. which we *think* are more likely to trigger broken behavior than "general" numbers. But by doing so, of course we preselected the tests, and may overlook bugs that would have been uncovered if we just had tested "every" integer -- but we can't.

Thus, just saying there "might" be inputs that lead to segfaults is not enough. You need to analyze the situation and give reasonable arguments why certain inputs ought to be included in the tests.

A reasonable argument here is that analysis of the coveralls report (the changed coverage percentage alone is meaningless) reveals that the removed test tested parsing of character in hex encoding with upper case hex letters. So I added that back, and even more (I added mixed-case hex constants -- in the current code that doesn't matter, but I can think of bugs where it might, so i am adding it proactively). But I am e.g. *not* adding printing of every hex constant from 00 to 255/FF (even though this would actually be feasible in this particular case). Of course we may eventually learned that printing `'0xf3'` is broken -- so be it, I am willing to risk that.

> 
> It's much easier to deal with the sutuation where a user reports a test failure, than with a user report saying "help, I get a segfault, while all the GAP tests pass".

True, but irrelevant for this PR. There is no segfault here, or any reason to suspect that one could be triggered. If we followed your argument, we'd be back to "let's test printing of every single integer".

fingolfin:
Backported to stable-4.11 in c18b0c4215b5212a2cc4f305e2d5b94ba716bee8

PR #4023, opened 2020-05-12
kernel: improve float literal parsing, fix some inconsistencies in it

fingolfin:
I had a look at the code coverage for the kernel function `GetNumber` which parses integers and floats. There was some code I could not get coverage for, and after some thinking about, determined that it was dead. Yay! This PR first adds some test to increase coverage for the float parsing code (they pass on their own). It then contains a bunch of commits which refactor the code gradually, in an attempt to make it "obvious" that the refactoring is correct (but the reviewers of the PR be the judge of whether I managed that ;-). Please read the commit messages, they are meant to explain my reasoning at each step.

The result is much simple and also more consistent (as the changes to tests added at the start in one of the later commits highlight).

### DIFF
--- a/tst/testinstall/kernel/scanner.tst
+++ b/tst/testinstall/kernel/scanner.tst
@@ -233,29 +233,25 @@ stream:1
 gap> 1.e0i;
 "std:1.e0i"
 gap> 1.e0ix;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.e0ix;
-     ^
+^^^^^
 gap> 1.e0_;
 "std:1.e0"
 gap> 1.e0_y;
 "eager:1.e0"
 gap> 1.e0_yx;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.e0_yx;
-      ^
+^^^^^^
 gap> 1.e0i_;
-Syntax error: ; expected in stream:1
-1.e0i_;
-     ^
+"std:1.e0i"
 gap> 1.e0i_y;
-Syntax error: ; expected in stream:1
-1.e0i_y;
-     ^^
+"eager:1.e0i"
 gap> 1.e0i_yx;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.e0i_yx;
-     ^^^
+^^^^^^^
 
 #
 gap> 1.0i;
@@ -327,29 +323,25 @@ stream:1
 gap> 1.0e0i;
 "std:1.0e0i"
 gap> 1.0e0ix;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.0e0ix;
-      ^
+^^^^^^
 gap> 1.0e0_;
 "std:1.0e0"
 gap> 1.0e0_y;
 "eager:1.0e0"
 gap> 1.0e0_yx;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.0e0_yx;
-       ^
+^^^^^^^
 gap> 1.0e0i_;
-Syntax error: ; expected in stream:1
-1.0e0i_;
-      ^
+"std:1.0e0i"
 gap> 1.0e0i_y;
-Syntax error: ; expected in stream:1
-1.0e0i_y;
-      ^^
+"eager:1.0e0i"
 gap> 1.0e0i_yx;
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.0e0i_yx;
-      ^^^
+^^^^^^^^
 
 # restore float handlers
 gap> Unbind(EAGER_FLOAT_LITERAL_CONVERTERS.y);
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Do you believe this changes the parsing of what are valid floats?

fingolfin:
@ChrisJefferson If so then only in that it fixes currently broken behavior, and accepts something that was not accepted before but I think should have... That said, I just realized I mad a mistake in the initial version of this PR (which is now fixed, though)

So here is what changed (and it would be great if @stevelinton and/or @laurentbartholdi could check this over): We allow for these trailing letters and underscores in float literals, to affect conversion of literals to actual float objects (this is sadly underspecified and difficult to find in the documentation). There was three copies of the code responsible for parsing these, each differing from the other two slightly. One of the copies was never executed and is removed in one of the first commits in this PR. Of the remaining two, one was called on float literals without an exponent, and one was called for float literals *with* an exponent. The latter did not accept all combinations the former accepted..

So, as far I understand, all of the following should be valid float literals; here `x` could be e.g. something like `i` to indicate an imaginary value (I am not aware of anything implementing this, but the [GAP manual chapter on `Floats` mentions this possibility](https://www.gap-system.org/Manuals/doc/ref/chap19.html#X7E8F6EFB87A65F78) -- though only for `i`, but the code accept an arbitrary letter from `[a-zA-Z]`. This optional indicator may be followed by an underscore which activates immediate or "eager conversion of the float literal. This then may optionally be followed by another letter, which can be used to request eager conversion by a specific float handler (e.g. `l` is predefined and requests that a GAP machine integer be created).

- 1.0_
- 1.0x
- 1.0_y
- 1.0x_
- 1.0x_y

But these should also be equally valid with an exponent:

- 1.0e0_
- 1.0e0x
- 1.0e0_y
- 1.0e0x_    → this lead to a syntax error before this PR
- 1.0e0x_y  → this lead to a syntax error before this PR

You can look at commit 47b6f54eb1598ae6f4d91440deaf84a2d2a2196d to see the change -- it actually still rejects these (simply because I have no float handler which actually accepts literals like `1.0i`), but not with a syntax error, just with a conversion error.

Conversely, we tried to parse some float literals despite a syntax error.

All in all, several inputs in the past lead to both a syntax error and a "Error, failed to convert float literal" -- now this doesn't happen anymore, you get exactly one of the two.

laurentbartholdi:
This looks great, Max! I won't comment much more on it because the code was written by @stevelinton . You may want to extend the tests with 
gap> EAGER_FLOAT_LITERAL_CONVERTERS.x := x->Concatenation("x:",x);;
gap> 1.0_x;
"x:1.0"
gap> FLOAT_STRING := x->Concatenation("std:",x);;
gap> 1.0i;
"std:1.0i"

and so on.

fingolfin:
@laurentbartholdi ah, that's a good hint, thanks! I'll have to make sure to restore the normal behavior after the tests, of course, but that should be doable. I'll look into doing this later, but in the meantime, I have some other things to attend to, and perhaps @stevelinton and @ChrisJefferson are interested in looking at the code changes already now. Again: while the overall PR might look a bit overwhelming, I think each commit on its own is manageable (provided that you trust me that when I write that I swapped two blocks of code, that that's really all I did -- I pinky swear that it is so, modulo honest mistakes)

stevelinton:
I agree this looks like a great improvement. I can see no obvious problems.

fingolfin:
While we are at it, could someone explain to me why GAP also accepts `d` or `q` instead of `e` for exponents? The resulting literals cannot be parsed by `strtod` (which, however, supports `p` for hexdecimal exponents...)

fingolfin:
That said, `CONVERT_FLOAT_LITERAL` takes care of converting any of `dDqQ` to `e`; but I still wonder: why?!

stevelinton:
@fingolfin I seem to recall that D is used in some contect for double precision (64 bit) and q for quad precision (128 bit).

fingolfin:
@stevelinton I tried to find a reference for that and couldn't find anything; not in IEEE-754 2019 and also not in the GAP manual...

fingolfin:
Anyway, I followed @laurentbartholdi's advice to install custom conversion handlers, and now I think the answer to @ChrisJefferson question is much clearer visible, in commit a8db26328aa4004cc237752cd47cf941b1d1e400 one sees how e.g. `1.0e0i_` used to produce a syntax error and now works "as expected".

laurentbartholdi:
I'm the culprit. Fortran sometimes prints floateans with D or Q, depending
on the precision. It sounded harmless to accept them.

fingolfin:
"It sounded harmless to accept them." -- I totally get that; the problem with these things is that you basically cannot take them back, as people start relying on them... Though here it might even be possible, because that feature seems to be 100% undocumented? Anyway it's not that I have plans to change this, I simply wanted to understand what this is all about -- now I do, thanks for the explanations :-)

tst/testinstall/kernel/scanner.tst
```diff
@@ -200,29 +200,25 @@ stream:1
...
-Syntax error: ; expected in stream:1
+Syntax error: Badly formed number in stream:1
 1.e0_yx;
-      ^
+^^^^^^
 gap> 1.e0i_;
-Syntax error: ; expected in stream:1
-1.e0i_;
-     ^
+"std:1.e0i"
```
fingolfin, line 215: E.g. this syntax error was wrong, and now returns a working element

fingolfin:
@stevelinton @laurentbartholdi @ChrisJefferson so if one of you dares to approve this (don't worry, if there are regressions caused by this, the blame is on me, not on the reviewer(s) who approved ;-)

PR #4026, opened 2020-05-13
buildsys: only include .d files that exist

### LINKED ISSUE #4025: buildsys: out-of-tree builds regression related to config.h
My recent buildsystem overhaul broke something; it seems out-of-tree builds are not regenerating config.h correctly anymore. This is likely due to vpath trickery. I will fix this, but don't want to forget about it in the meantime, hence this issue.

fingolfin:
Fixes #4025

Of course ideally we'd also have tests for this... Here is one way we could do that: we could have a Travis test which uses an out-of-tree build where we do `make clean` *before* `make` but after configure. That'll trigger the bug as I fixed it.

We could of course do more thing: 
- after the `make`, remove one file *.d file then run `make` and verify it is being recreated
- the same but for an .lo file
- add in a "bad" .d file which emulates a leftover from a renamed file (with invalid syntax in it) and verify it is *not* included (as in: `make` doesn't run into an error)
- verify that `make ; make` does not regenerate gap or anything (say, by testing that the second `make` has no output)
- test mixing an in-tree and an out-of-tree build to verify the OOT build does not accidentally use files from the in-tree build (say by first doing the in-tree build, then replacing all .d and .lo files with garbage, then doing the OOT build -- if it picks up one of the wrong files, it'll run into an error)
- ... more  can be added over time. That would have caught some of the regressions I accidentally introduced. Of course it won't catch all problems, but we can at least give it a try


fingolfin:
I added a bunch of tests. They involving using the name `gen`, so there is an indirect conflict with PR #4013 here. I suggest this PR here is merged *first* (as I think merging PR #4013 will trigger the issue this PR here is meant to fix for everybody working on GAP!). Once this PR here is merged, I'l update PR #4013.

Makefile.rules
```diff
@@ -290,7 +290,7 @@ endif
 DEPFILES = $(patsubst %,gen/deps/%.d,$(SOURCES))
 
 # Include the dependency tracking files, skip any missing ones
```
fingolfin, line 292: Note that the comment already that we "skip any missing ones", but that wasn't true anymore with my previous refactoring of the buildsystem; this PR restores the behavior described in the comment

Issue #3937, opened 2020-03-31
Coverage of testspecial is not tracked on coveralls


zickgraf:
For an example see https://travis-ci.org/github/gap-system/gap/jobs/668660257 on line 1619 of the output:
```
> > > gap> gap> Scanning for coverage data...
gap> gap> > > > > > gap> > > > > > > > > > > > > > > > > > > > > No coverage files found...
```

PR #4031, opened 2020-05-17
CI: send testspecial coverage data to Coveralls.io, too

fingolfin:


fingolfin:
Excluding `extern` from the coveralls reports works! But somehow I still don't see the coverage which should result from `testspecial` there. So this needs more digging. But in the meantime, I think the changes here are still a strict improvement and should be merged.

zickgraf:
@fingolfin Does this close #3937? From reading the PR title I assume it does :-)

PR #4033, opened 2020-05-18
Travis: investigate diskquota failures (DO NOT MERGE OR REVIEW)

fingolfin:
See also https://travis-ci.community/t/s390x-adding-apt-sources-no-usable-temporary-directory-found-tmp-enospc/8635/2

Perhaps we can figure out the source of the disk being full and implement a counter; that would help us avoid the barrage of temporary failures of the ppc64le/arm64/s390x failures on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,124 +33,124 @@ addons:
 matrix:
   include:
     # general test suite (subsumes testinstall tests, too)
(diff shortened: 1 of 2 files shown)


fingolfin:
Travis fixed this for ARM64 it seems but not s390x. But it's just too sporadic to be caught by this, it seems, so I am giving this up for now.

PR #4036, opened 2020-05-19
Add GitHub Actions CI tests

fingolfin:
Build results can be seen at https://github.com/fingolfin/gap/actions
- [ ] verify that failed tests are properly detected
- [ ] add "all" tests performed by Travis
- [ ] investigate whether we can come up with an Little Endian test again, to replace the Travis one (some people seem to be using qemu with GitHub Actions successfully) -- e.g. look at https://github.com/uraimo/run-on-arch-action
- [ ] restore 32bit tests
- [ ] get Codecov integration to work
- [ ] get Coveralls integration to work
- [ ] cache the package tarballs, by using that we can download their corresponding sha256 tarballs
- ...

### DIFF
--- a/etc/ci-gather-coverage.sh
+++ b/etc/ci-gather-coverage.sh
@@ -71,7 +71,40 @@ fi;
 # Always output coveralls JSON, as ci-coveralls-merge.py relies
 # on it being present and containing all relevant metadata...
 Print("Outputting JSON for Coveralls...\n");
-OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix);
+
+env := GAPInfo.SystemEnvironment;;
+if IsBound(env.GITHUB_ACTIONS) then
+    opt := rec(
+        service_name := "github",
+        # The build number. Will default to chronological numbering from builds on repo
+        service_number := env.GITHUB_RUN_NUMBER,
+        # A unique identifier of the job on the service specified by service_name
+        # FIXME: there seems to be no unique per-job id; the GITHUB_RUN_ID is
+        # shared by all jobs in a single build :-(
+        #service_job_id := env.GITHUB_RUN_ID,
+        service_branch := env.GITHUB_REF{[Length("refs/heads/")..Length(env.GITHUB_REF)]},
+        commit_sha := env.GITHUB_SHA,
+#         service_build_url := Concatenation(
+#             "https://ci.appveyor.com/project/",
+#             env.APPVEYOR_REPO_NAME,
+#             "/build/",
+#             env.APPVEYOR_BUILD_VERSION),
+    );
+
+    # GITHUB_REF has the form refs/pull/12345/merge
+    if IsBound(env.GITHUB_REF) and StartsWith(env.GITHUB_REF, "refs/pull/") then
+        tmp := SplitString(env.GITHUB_REF, "/");
+        if Length(tmp) = 4 and tmp[4] = "merge" and Int(tmp[3]) <> fail then
+            # The associated pull request ID of the build. Used for updating the status and/or commenting
+            opt.service_pull_request := tmp[3];
+        fi;
+    fi;
+
+    OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix, opt);
+
+else
+    OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix);
+fi;
 GAPInput
 
 if [[ -f gap-coveralls.json ]]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,207 @@
+name: CI
+
+# Trigger the workflow on push or pull request
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: ${{ matrix.test-suites }} - ${{ matrix.extra }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # base test: fast first test
+        os: [ubuntu-latest]
+        test-suites: ["testinstall"]
+        extra: [""]
+
+        # add a few extra tests
+        include:
+          - os: ubuntu-latest
+            test-suites: "docomp teststandard"
+            extra: ""
+
+#          - os: ubuntu-latest
+#            test-suites: "docomp teststandard"
+#            extra: "ABI=32 CONFIGFLAGS=\"\""
+
+          # FIXME: we used to run `teststandard` for HPC-GAP under Travis CI, but
+          # somehow when running on GitHub Actions, it takes almost 4 hours (!) to
+          # complete instead of 25 minutes. So for now we just run testinstall.
+          - os: ubuntu-latest
+            test-suites: "docomp testinstall"
+            extra: "HPCGAP=yes ABI=64"
+
+          # compile packages and run GAP tests
+          # don't use --enable-debug to prevent the tests from taking too long
+          - os: ubuntu-latest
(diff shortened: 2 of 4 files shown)


fingolfin:
Tests work but coverage integratiom (Codecov, Coveralls) requires work, possibly also in the `profiling` package.

fingolfin:
I've already disabled a bunch of tests (e.g. anything testing 32bit builds) which seems non-trivial to getting working again. Right now, the test-packages job still fails. Perhaps that can be fixed, otherwise we could just disable it as well, and merge this partially, just so that we get at least *some* Linux+Max based CI feedback on our PRs (in addition to the Cygwin tests running on AppVeyor; thank god we still have at least those).

fingolfin:
I've added libcurl4-openssl-dev hopefully that'll cure the `testpackage` failure. But really, `testpackage` should provide more helpful output, just "PASS" or "FAIL" are really poor for debugging

wilfwilson (commented):
I've had a look through what you've done. I support merging what you've got working so far, which seems to me to be quite a lot.

I see that in the most recent build [the testspecial job](https://github.com/fingolfin/gap/runs/1379200164) is newly failing. Do you think that's just a quirk? And related: does GitHub Actions have a facility for re-running particular jobs without having to push a whole new commit? (Obviously I don't have access to do that on your fork of gap, but I'm wondering how it'll work once it's merged into the main repo).

I also can't quite tell how far you've got with the code coverage integration. It certainly seems like reports are getting uploaded to Codecov. Are you happy with that bit, or is it not there yet?

Also: thank you. Getting CI to work properly - and keep working - for the Digraphs and Semigroups packages in past years left a bitter taste in my mouth when it comes to CI. It's so frustrating to debug a system when you don't even have interactive access to the system!

ssiccha:
@fingolfin What do you think should be fixed before this can be merged? Is it mostly fixing the `testspecial` test and replicating this to the [pkg-ci-scripts repo](https://github.com/gap-system/pkg-ci-scripts) for the packages or am I missing something? I guess we don't need to get the Little Endian and 32-bit tests working _right now_, do we?

Codecov seems to work, so coveralls doesn't seem thaat important for now. At a first glance the codecov pages generated by [your branch](https://codecov.io/gh/fingolfin/gap/tree/fc7ad89d32229710ac380c5ad3ae38d868a6f029) and e.g. an older [PR of mine](https://codecov.io/gh/gap-system/gap/tree/d08499592c87b0b1d7d6ae9eaa4cfb8a2f01bcb7) don't look too different.

The testspecial failure is due to the WIP commit fc7ad89d. It's called WIP after all, so no big surprise. At least reverting it on my machine makes testspecial pass, leaving the commit in makes it fail in the same way as on the action run.

ssiccha:
I've hacked together a [first draft PR](https://github.com/gap-packages/recog/pull/202) for using GH actions with the recog package, see [the actions on my fork](https://github.com/ssiccha/recog/actions) for the first run. It built GAP so I must have done something right. ATM it fails when building the manuals but that might be due to recog being a bit iffy there. I'll have a look tomorrow. :)

fingolfin:
@wilfwilson wwrote:

> I've had a look through what you've done. I support merging what you've got working so far, which seems to me to be quite a lot.

It's quite a bit, but some things are still missing, e.g. 32bit tests. I'll leave it to someone else to figure those out -- or we just drop official 32bit support (I am not talking about removing it, just about: we don't test there and we don't make promises it'll work)

> 
> I see that in the most recent build [the testspecial job](https://github.com/fingolfin/gap/runs/1379200164) is newly failing. Do you think that's just a quirk?

I hope so; it passed again in the latest build... not sure what went wrong there.

> And related: does GitHub Actions have a facility for re-running particular jobs without having to push a whole new commit? (Obviously I don't have access to do that on your fork of gap, but I'm wondering how it'll work once it's merged into the main repo).

One can re-run the complete build, but unfortunately not (yet) individual jobs; this has of course already been requested by many people, somewhere there's an issue open for that.

> 
> I also can't quite tell how far you've got with the code coverage integration. It certainly seems like reports are getting uploaded to Codecov. Are you happy with that bit, or is it not there yet?

I think Codecov upload works, but Coveralls does not; The problem with Codecov is that it's been super flaky and e.g. I haven't been able to view PR reports for GAP on their website for years (they are too big for them to process "on the fly" ; try accessing e.g. https://codecov.io/gh/gap-system/gap/pull/4159). And they play possum with regards to support requests... So Coveralls would be a nice alternative. Both Codecov and Coveralls offer premade actions to upload coverage data, but one has to figure out how to make them work well with what `profiling` produces (and where etc.). It's probably not difficult but very tedious work, because you can't test locally.

> 
> Also: thank you. Getting CI to work properly - and keep working - for the Digraphs and Semigroups packages in past years left a bitter taste in my mouth when it comes to CI. It's so frustrating to debug a system when you don't even have interactive access to the system!

Indeed! That's super annoying.

etc/ci-gather-coverage.sh
```diff
@@ -71,7 +71,40 @@ fi;
 # Always output coveralls JSON, as ci-coveralls-merge.py relies
 # on it being present and containing all relevant metadata...
 Print("Outputting JSON for Coveralls...\n");
-OutputCoverallsJsonCoverage(r, "gap-coveralls.json", prefix);
+
+env := GAPInfo.SystemEnvironment;;
+if IsBound(env.GITHUB_ACTIONS) then
```
fingolfin, line 76: @wilfwilson @ChrisJefferson @ssiccha if you take a look at the hacky patch I made to etc/ci-gather-coverage.sh: that's probably the biggest thing we need, and it should be folded into the profiling package: a way to detect that we are inside a GitHub Action run, and then use that to set the correct meta data for coveralls... But perhaps there is a way to avoid all that, too, e.g. by using https://github.com/marketplace/actions/coveralls-github-action

The main reason these shenigangs are needed is to help coveralls understand which coverage results belong together / belong to a specific PR. But again, perhaps the GitHub Action I just linked to helps us with that and we can simplify the code in profiling. But I am afraid that likely we'll need *some* kind of modification to profiling (also for GAP packages)... Not sure

etc/ci-gather-coverage.sh
```diff
@@ -71,7 +71,40 @@ fi;
...
+    opt := rec(
+        service_name := "github",
+        # The build number. Will default to chronological numbering from builds on repo
+        service_number := env.GITHUB_RUN_NUMBER,
+        # A unique identifier of the job on the service specified by service_name
+        # FIXME: there seems to be no unique per-job id; the GITHUB_RUN_ID is
+        # shared by all jobs in a single build :-(
+        #service_job_id := env.GITHUB_RUN_ID,
+        service_branch := env.GITHUB_REF{[Length("refs/heads/")..Length(env.GITHUB_REF)]},
```
fingolfin, line 85: So even with https://github.com/marketplace/actions/coveralls-github-action we need to invest work to handle parallel jobs (= the default for almost everybody using GitHub Action Workflows, sheesh, thanks for making everybody work for that Coveralls :-( and kudos to Codecov for making do without.

I think the only thing we can really do here is to manually assign a "name" to each of our CI jobs, which we can then use. Actually, perhaps we can just use the existing `name: ${{ matrix.test-suites }} - ${{ matrix.extra }} - ${{ matrix.os }}` field for that.

fingolfin:
Assuming the tests pass, I suggest we merge this and then I'll open a new PR in which we can try to get coverage working fully again (and possibly some of the missing tests). Note that I am not removing Travis integration in this PR as I still hope we can get that back at least for a while.

alex-konovalov:
Curious - merged, but no checkboxes ticked in the decription. @fingolfin ?

PR #4038, opened 2020-05-24
kernel: add SELF_NAME, use SELF_NAME in RequireFOO calls where possible

fingolfin:
`SELF_NAME` is a convenience helper to access the name of GAP a kernel function as a C     string, which can be passed to `RequireArgumentEx` and its variants.

The advantage is that one does not have to duplicate function names in lots of places; and if the function is renamed, the errors are automatically adjusted. In a few places, this actually fixes error messages; in a few places, it introduces regressions, but arguably most of those hint at cases where should perhaps just rename the kernel function instead of having a kernel function with one name and an identical GAP function with another name (e.g. `NumeratorRat` vs. `NUMERATOR`). Otherwise, one could just change the affected functions back to not using `SELF_NAME`. 

BTW, we could think about exposing SELF_NAME() as a GAP function (which would have to rely on some hackery to determine the *caller function* and then get the name of *that*) to be used for similar features.

### DIFF
--- a/src/blister.c
+++ b/src/blister.c
@@ -1068,7 +1068,7 @@ static Obj FiltIS_BLIST_REP(Obj self, Obj obj)
 */
 static Obj FuncSIZE_BLIST(Obj self, Obj blist)
 {
-    RequireBlist("SizeBlist", blist);
+    RequireBlist(SELF_NAME, blist);
     return INTOBJ_INT(SizeBlist(blist));
 }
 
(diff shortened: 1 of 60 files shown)


ChrisJefferson:
Minor suggestion, stripping the 'Func' off (is present) when talking the value of __func__, to better line up with the GAP-level name

ChrisJefferson:
Although maybe every Func will also have a self, and therefore not hit this codepath?

fingolfin:
Every `FuncFOOBAR` which is to be used as handler for a GAP function must have `Obj self` as its first argument. Any `FuncFOOBAR` which is used for a different purpose won't be visible under the name `FOOBAR` in GAP. Unless you do something really wonky (I mean, of course I could write `Obj FuncQUUX(Obj self) {  FuncFOOBAR();  return 0; }` or so, then export `QUUX` to the GAP level, do `FOOBAR := QUUX;` there and document `FOOBAR`... but really, if somebody insists on doing something idiotic like that, I'd in return insist that `SELF_NAME` was not made for them :-)

PR #4039, opened 2020-05-24
Travis: make the `chown` workaround more resilient

fingolfin:
It recently started to fail on arm64 (but not on s390x) jobs, which
might mean the underlying issue is fixed there, or it just might be that
it got even more random -- hard to tell w/o any kind of feedback from
Travis engineers. In any case, by checking whether ~/.cache/pip/wheels
exists before trying to change its owner & group, we can avoid troubles.


fingolfin:
Backported to stable-4.11 in commit f9e7e4b45a7443abce7dadbc56bf5df68d930d6f

PR #4041, opened 2020-05-27
buildsys: fix 'make check' in out-of-tree builds

fingolfin:


fingolfin:
Backported to stable-4.11 in commit ac24f986ca4148292a116ad7f0a05e7d6f1ca868

PR #4044, opened 2020-06-02
Improve sysinfo.gap generation and content

fingolfin:
In particular define GAP and GAC variables in `sysinfo.gap` and use them in `bin/BuildPackages.sh` and `etc/Makefile.gappkg`.

This strengthens the purpose of `sysinfo.gap` (as I perceive it): that in order to build a package (or anything that wants to interact with GAP), you just point it at the sysinfo.gap of the target GAP installation, and it extracts all it needs from there, including the location of a GAP and GAC executable.

Note that `sysinfo.gap` can be directly included from POSIX shell scripts and Makefiles, and it is also fairly easy to parse for other systems.

Hopefully `PackageManager` will also eventually benefit from this.

This doesn't cover all situations and setups in every case, but that's OK; if people have a non-standard setup (involving their own modified `gap.sh` scripts etc.) then that's how it is, period. (We can help those cases a bit more by adding ways to query a GAP executable for its settings. Some ways that come to mind:
- add a command line option to `gap` which prints the path to the `sysinfo.gap` file
- add a command line option to `gap` which prints the content of the `sysinfo.gap` file
- add a GAP command which returns that path of the `sysinfo.gap` file 
- then teach BuildPackage.sh that if given the name of a GAP executable but not a GAPROOT, it should query the GAP command for the "gaproot" (i.e. the path in which sysinfo.gap is)
- also PackageManager could take advantage of these (see also https://github.com/gap-packages/PackageManager/pull/55/)


fingolfin:
Backported to stable-4.11 in commit 4b8aebd97e00026ab97ddf1551e3fd6d5add1f78

PR #4045, opened 2020-06-03
kernel: refactor IntrBegin / IntrEnd and their callers

fingolfin:


src/read.c
```diff
@@ -2632,23 +2638,22 @@ ExecStatus ReadEvalCommand(Obj context, Obj *evalResult, UInt *dualSemicolon)
...
 
-        /* check for dual semicolon */
-        if (dualSemicolon)
-            *dualSemicolon = (rs->s.Symbol == S_DUALSEMICOLON);
-    }
-    CATCH_ERROR {
-        IntrEnd(&rs->intr, 1, evalResult);
-        type = STATUS_ERROR;
+    // restore the execution environment
+    SWITCH_TO_OLD_LVARS(oldLVars);
```
ChrisJefferson, line 2646: Not a requirement, and we could do it later...

I feel like the "SWITCH_TOO_OLD_LVARS" should come after the IntrEnd. I realise this lines up more with the current code, but it feels really weird. I don't think it would break anything (famous last words)

fingolfin: You raise a good point here. It is arguably "more correct" to do it the way you suggest (and I just changed it accordingly!).
Of course as you already deduced, I did it the other way around before to minimize changes in behavior compared to before my PR. But `IntrEnd` actually may call `CodeEnd`, and *that* also can call `SWITCH_TO_OLD_LVARS`... I am not 100% sure why this works! I assume the fact that most `ReadEvalCommand` invocations uses the `BottomLVars` as start context anyway, so we keep reseting to that, and so `CodeEnd` switches to the right LVars by luck.

Or perhaps some other factor that I don't see right now (I am rather tired and my brain is foggy) actually makes this correct after all? But I still think the new variant is better.

PR #4053, opened 2020-06-17
Fix the logic for scanning tasks in the Julia GC.

rbehrends:
This fixes the test that tries to avoid scanning the stack of the current thread twice.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -646,7 +646,7 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
         if (tag->bits.gc & 2)
             rescan = 0;
     }
-    if (stack && tid < 0) {
+    if (stack) {
         if (task->copy_stack) {
             // We know which part of the task stack is actually used,
             // so we shorten the range we have to scan.


src/julia_gc.c
```diff
@@ -646,7 +646,7 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
         if (tag->bits.gc & 2)
             rescan = 0;
     }
-    if (stack && tid < 0) {
+    if (stack) {
```
fingolfin, line 649: This seems like the right thing to do....

In fact `tid < 0` means `tid == -1` means that the task is not the current task of any thread. Why was it ever OK to skip scanning the stack of such tasks? It might have been the previously active task on a thread, right? And then skipping it could lead to us missing a GC root and hence a GC crash, no?

Or is it not that bad? Then what am I missing this time?

rbehrends: You aren't missing anything, that was a bug.

fingolfin:
Backported to stable-4.11 in 1450c23d0a5539d68258883b59af0d89eacd2c46

PR #4054, opened 2020-06-17
Enhancements around SmallGeneratingSet

hulpke:
These are functionally equivalent cleanups and speedups of the operations `SmallGeneratingSet` (and `LogPerm` that is uses) and `SmallerDegreePermutationRepresentation` (which now uses the former).

These operations are used throughout the library, so no extra tests needed.

Also added a safety finiteness test that had been requested for `AllHomomorphisms`.

### DIFF
--- a/doc/ref/grpperm.xml
+++ b/doc/ref/grpperm.xml
@@ -92,7 +92,7 @@ gap> Length(MovedPoints(p));
 720
 gap> q:=SmallerDegreePermutationRepresentation(p);;
 gap> NrMovedPoints(Image(q));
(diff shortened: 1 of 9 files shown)


hulpke:
> With this PR, `tst/teststandard/opers/AutomorphismGroup.tst` line 44 now takes > 10 minutes, leading to a Ci failure

Thank you. I'm puzzled -- the test runs through on my computer in <40 seconds, and I can't detect anything that takes disproportionally longer than before. I've changed the example to a smaller one.

PR #4057, opened 2020-06-19
Travis: switch most builds to bionic

fingolfin:
The motivation for this is a hope that as a weird side effect it might fix the s390x builds (but honestly I don't see how, it's just a random try).

See https://travis-ci.community/t/s390-xenial-builds-fail-with-an-error-occurred-while-generating-the-build-script/9039

If this doesn't work, we'll have to drop the s390x builds, which would be sad, as they are our only big endian CI builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-# Force Ubuntu 16.04 "Xenial" to get newer GCC, binutils etc.
-dist: xenial
+dist: bionic
 language: c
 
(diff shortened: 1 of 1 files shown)


alex-konovalov:
@fingolfin something went wrong:
```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install gcc-4.7 g++-4.7 valgrind" failed and exited with 100 during .
```

PR #4058, opened 2020-06-24
Refine the logic for scanning Julia stacks.

rbehrends:
This fixes how we handle scanning the root task of the main thread. It also improves the comments that explain stack scanning logic.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -199,6 +199,7 @@ static jl_datatype_t * datatype_largebag;
 static UInt            StackAlignBags;
 static Bag *           GapStackBottom;
 static jl_ptls_t       JuliaTLS, SaveTLS;
+static jl_task_t *     RootTaskOfMainThread;
 static size_t          max_pool_obj_size;
 static UInt            YoungRef;
 static int             FullGC;
@@ -387,6 +388,38 @@ void MarkJuliaObj(void * obj)
 // from which freed objects are then marked. Hence, we add additional checks
 // when traversing GAP master pointer and bag objects that this happens
 // only for live objects.
+//
+// We use "bottom" to refer to the origin of the stack, and "top" to describe
+// the current stack pointer. Confusingly, on most contemporary architectures,
+// the stack grows "downwards", which means that the "bottom" of the stack is
+// the highest address and "top" is the lowest. The stack is contained in a
+// stack buffer, which has a start and end (and the end of the stack buffer
+// coincides with the bottom of the stack).
+//
+//   +------------------------------------------------+
+//   | guard |    unused area          | active stack |
+//   | pages |  <--- growth ---        | frames       |
+//   +------------------------------------------------+
+//   ^                                 ^              ^
+//   |                                 |              |
+// start                              top         bottom/end
+//
+// All stacks in Julia are associated with tasks and we can use
+// jl_task_stack_buffer() to retrieve the buffer information (start and size)
+// for that stack. That said, in a couple of cases we make adjustments.
+//
+// 1. The stack buffer of the root task of the main thread, when started
+//    from GAP can extend past the point where Julia believes its bottom is.
+//    Therefore, for that stack, we use GapBottomStack instead.
+// 2. For the current task of the current thread, we know where exactly the
+//    top is and do not need to scan the entire stack buffer.
+//
+// As seen in the diagram above, the stack buffer can include guard pages,
+// which trigger a segmentation fault when accessed. As the extent of
+// guard pages is usually not known, we intercept segmentation faults and
+// scan the stack buffer from its end until we reach either the start of
+// the stack buffer or receive a segmentation fault due to hitting a guard
+// page.
 
 static void TryMark(void * p)
 {
@@ -485,13 +518,16 @@ static void SafeScanTaskStack(PtrArray * stack, void * start, void * end)
         (volatile jl_jmp_buf *)JuliaTLS->safe_restore;
     jl_jmp_buf exc_buf;
     if (!jl_setjmp(exc_buf, 0)) {
-        // The bottom of the stack may be protected with
-        // guard pages; accessing these results in segmentation
-        // faults. Julia catches those segmentation faults and
-        // longjmps to JuliaTLS->safe_restore; we use this
-        // mechamism to abort stack scanning when a protected
-        // page is hit. For this to work, we must scan the stack
-        // from top to bottom, so we see any guard pages last.
+        // The start of the stack buffer may be protected with guard
+        // pages; accessing these results in segmentation faults.
+        // Julia catches those segmentation faults and longjmps to
+        // JuliaTLS->safe_restore; we use this mechanism to abort stack
+        // scanning when a protected page is hit. For this to work, we
+        // must scan the stack from the end of the stack buffer towards
+        // the start (i.e. in the direction in which the stack grows).
+        // Note that this will by necessity also scan the unused area
+        // of the stack buffer past the stack top. We therefore also
+        // optimize scanning for areas that contain only null bytes.
         JuliaTLS->safe_restore = &exc_buf;
         FindLiveRangeReverse(stack, start, end);
     }
@@ -591,22 +627,23 @@ static void GapRootScanner(int full)
     // 2. The stack of the current task is that of the root task of the
     //    main thread (which has thread id 0).
     //
-    // The reason is that if Julia is being initialized from GAP, it
-    // cannot always reliably find the top of the stack for that task,
-    // so we have to fall back to GAP for that.
-    if (!IsUsingLibGap() && jl_threadid() == 0 &&
-        JuliaTLS->root_task == task) {
+    // The reason is that when called from GAP, jl_init() does not
+    // reliably know where the bottom of the initial stack is. However,
+    // GAP does have that information, so we use that instead.
+    if (task == RootTaskOfMainThread) {
         stackend = (char *)GapStackBottom;
     }
 
-    // allow installing a custom marking function. This is used for
+    // Allow installing a custom marking function. This is used for
     // integrating GAP (possibly linked as a shared library) with other code
     // bases which use their own form of garbage collection. For example,
     // with Python (for SageMath).
     if (ExtraMarkFuncBags)
         (*ExtraMarkFuncBags)();
 
-    // scan the stack for further object references, and mark them
+    // We scan the stack of the current task from the stack pointer
+    // towards the stack bottom, ensuring that we also scan any
+    // references stored in registers.
     jmp_buf registers;
     setjmp(registers);
     TryMarkRange(registers, (char *)registers + sizeof(jmp_buf));
@@ -647,13 +684,34 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
             rescan = 0;
     }
     if (stack) {
-        if (task->copy_stack) {
+        // task->copy_stack is 0 if the COPY_STACKS implementation is
+        // not used or 1 if the task stack does not point to valid
+        // memory. If it is neither zero nor one, then we can use that
+        // value to determine the actual top of the stack.
+        switch (task->copy_stack) {
+        case 0:
+            // do not adjust stack.
+            break;
+        case 1:
+            // stack buffer is not valid memory.
+            return;
+        default:
             // We know which part of the task stack is actually used,
             // so we shorten the range we have to scan.
             stack = stack + size - task->copy_stack;
             size = task->copy_stack;
         }
-        ScanTaskStack(rescan, task, stack, stack + size);
+        char * stackend = stack + size;
+        if (task == RootTaskOfMainThread) {
+            stackend = (char *)GapStackBottom;
+        }
+
+        // Unlike the stack of the current task that we scan in
+        // GapRootScanner, we do not know the stack pointer. We
+        // therefore use a separate routine that scans from the
+        // stack bottom until we reach the other end of the stack
+        // or a guard page.
+        ScanTaskStack(rescan, task, stack, stackend);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
 // from which freed objects are then marked. Hence, we add additional checks
 // when traversing GAP master pointer and bag objects that this happens
 // only for live objects.
+//
+// All stacks in Julia are associated with tasks and we can use
+// jl_task_stack_buffer() to retrieve the buffer information for that stack.
+// That said, in a couple of cases these need adjustments.
+//
+// 1. The stack buffer of the root task of the main thread, when started from
+//    GAP can extend past the point where Julia believes it ends. Therefore,
```
fingolfin, line 397: Shouldn't this be:
```suggestion
//    GAP can extend past the point where Julia believes it starts. Therefore,
```
or
```suggestion
//    GAP can extend past the point where Julia believes its origin is. Therefore,
```
or so?

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+//    for that stack, we use GapBottomStack instead.
+// 2. If it is the current task of the current thread, we know where exactly
+//    the lowest valid address of the stack is and do not need to scan the
```
fingolfin, line 400:
```suggestion
//    the top (i.e.,  on all currently popular architectures: lowest valid address) of the stack is and do not need to scan the
```

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+// In addition, stack buffers can include guard pages, which trigger a
+// segmentation fault when accessed. We intercept such segmentation faults
+// and terminate stack scanning early for those.
+//
+// In GAP terminology, we use "bottom" to describe the end of the stack
+// where it starts and "top" to describe the current stack pointer. This
+// means that the "bottom" of the stack is the highest address on most
+// modern architectures and "top" is the lowest. When we treat the stack
+// as an abstract buffer, "start" and "end" refer to its lowest and
+// highest address, respectively.
```
fingolfin, line 412: That's not just "GAP terminology". I found this use of top and bottom in virtually all CS references I checked. Also, I'd bring these definitions earlier, then the text before here could use them instead of talking about origin/start/end/lowest address etc.

Also, I think "the end ... where it starts" is a really confusing choice of words. Perhaps: This:

```suggestion
// We use "bottom" to refer to the origin of the stack, and "top" to describe
// the current stack pointer. Confusingly, on most contemporary architectures,
// the stack grows "downwards", which means that the "bottom" of the stack
// is the highest address and "top" is the lowest. When we treat the stack
// as an abstract buffer, "start" and "end" refer to its lowest and
// highest address, respectively.
```
Also, perhaps instead of "start" and "end" it would then be better to talk about "low" and "high" ?

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+//
+//   +------------------------------------------------+
+//   | guard |  unused area            | active stack |
+//   | pages |                         | frames       |
```
fingolfin, line 416: How about adding in the growth direction?
```suggestion
//   | pages |     <--- growth ---     | frames       |
```

src/julia_gc.c
```diff
@@ -653,7 +685,16 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
             stack = stack + size - task->copy_stack;
             size = task->copy_stack;
         }
-        ScanTaskStack(rescan, task, stack, stack + size);
+        char * stackend = stack + size;
+        if (task == RootTaskOfMainThread) {
+            stackend = (char *)GapStackBottom;
+        }
+        // Unlike the stack of the current task that we can in
```
fingolfin, line 692:
```suggestion
        // Unlike the stack of the current task that we scan in
```

src/julia_gc.c
```diff
@@ -765,6 +806,22 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
...
     jl_init();
 
+    JuliaTLS = jl_get_ptls_states();
+    // These callbacks potentially require access to the Julia
+    // TLS and thus need to be installed after initialization.
+    jl_gc_set_cb_root_scanner(GapRootScanner, 1);
+    jl_gc_set_cb_task_scanner(GapTaskScanner, 1);
+    jl_gc_set_cb_pre_gc(PreGCHook, 1);
+    jl_gc_set_cb_post_gc(PostGCHook, 1);
+    // jl_gc_enable(0); /// DEBUGGING
```
fingolfin, line 816: Why was this moved up?

src/julia_gc.c
```diff
@@ -765,6 +806,22 @@ void InitBags(UInt initial_size, Bag * stack_bottom, UInt stack_align)
...
+
+    // If we are embedding Julia in GAP, remember the root task
+    // of the main thread. The extent of the stack buffer of that
+    // task is calculated a bit differently than for other tasks.
+
```
fingolfin, line 821:
```suggestion
```

src/julia_gc.c
rbehrends: These calls were originally directly after `jl_init()` (see comment). In between then and now, code was reorganized that moved them down and can in theory (though not currently in practice) trigger a GC for which they would be needed. This change is putting them back to where they were before.

src/julia_gc.c
rbehrends: I like the second option best.

src/julia_gc.c
rbehrends: This would unfortunately be inaccurate. The top of the stack is the current stack pointer, not the beginning of the stack buffer (which is an implementation detail).

src/julia_gc.c
rbehrends: It's conventional terminology when dealing with stacks as abstract data types; when dealing with hardware stacks, terminology is unfortunately inconsistent. Julia in particular uses the opposite terminology for its task stacks ([example](https://github.com/JuliaLang/julia/blob/6cd329c371c1db3d9876bc337e82e274e50420e8/src/task.c#L1058)). This is why I tried to distinguish between Julia and GAP terminology.

However, I think I still like your wording better, as it's overall clearer and more to the point, so I'm good with the change.

src/julia_gc.c
rbehrends: Good idea.

src/julia_gc.c
rbehrends: Oops.

src/julia_gc.c
rbehrends: Done.

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+//   +------------------------------------------------+
+//   | guard |    unused area          | active stack |
+//   | pages |  <--- growth ---        | frames       |
+//   +------------------------------------------------+
+//   ^                                 ^              ^
+//   |                                 |              |
+// start                              top         bottom/end
+//
+// All stacks in Julia are associated with tasks and we can use
+// jl_task_stack_buffer() to retrieve the buffer information for that stack.
```
fingolfin, line 408: With the definitions above, we could make this more precise, right?
```suggestion
// jl_task_stack_buffer() to retrieve the start and end pointer for that stack.
```

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+// That said, in a couple of cases these need adjustments.
+//
+// 1. The stack buffer of the root task of the main thread, when started
+//    from GAP can extend past the point where Julia believes its bottom is.
+//    Therefore, for that stack, we use GapBottomStack instead.
+// 2. If it is the current task of the current thread, we know where exactly
```
fingolfin, line 414:
```suggestion
// 2. For the current task of the current thread, we know where exactly
```

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+//   | guard |    unused area          | active stack |
+//   | pages |  <--- growth ---        | frames       |
+//   +------------------------------------------------+
+//   ^                                 ^              ^
+//   |                                 |              |
+// start                              top         bottom/end
+//
+// All stacks in Julia are associated with tasks and we can use
+// jl_task_stack_buffer() to retrieve the buffer information for that stack.
+// That said, in a couple of cases these need adjustments.
```
fingolfin, line 409:
```suggestion
// That said, in a couple of cases we make adjustments:
```
Note that you give one case where adjustments are needed, and one were it is an optimization.

src/julia_gc.c
```diff
@@ -387,6 +388,36 @@ void MarkJuliaObj(void * obj)
...
+//
+// 1. The stack buffer of the root task of the main thread, when started
+//    from GAP can extend past the point where Julia believes its bottom is.
+//    Therefore, for that stack, we use GapBottomStack instead.
+// 2. If it is the current task of the current thread, we know where exactly
+//    the top is and do not need to scan the entire stack buffer.
+//
+// In addition, stack buffers can include guard pages, which trigger a
+// segmentation fault when accessed. We intercept such segmentation faults
+// and terminate stack scanning early for those.
```
fingolfin, line 419: From this, it does not become clear (a) where the guard pages are (your sketch above of course shows this, but perhaps mention it again), but more importantly (b) it is unclear why Julia would point <start> beyond the guard page... And it's not clear to me if it really *does* that? I thought the problem was only for root tasks, as their stack is not allocated by Julia? Or what? Sorry, I am still confused about this :-/

src/julia_gc.c
```diff
@@ -491,7 +522,7 @@ static void SafeScanTaskStack(PtrArray * stack, void * start, void * end)
         // longjmps to JuliaTLS->safe_restore; we use this
         // mechamism to abort stack scanning when a protected
         // page is hit. For this to work, we must scan the stack
-        // from top to bottom, so we see any guard pages last.
+        // from bottom to top, so we see any guard pages last.
```
fingolfin, line 525: this doesn't fit the terminology above anymore
```suggestion
        // from start to end (as defined above), so we see any guard pages last.
```

src/julia_gc.c
```diff
@@ -653,7 +685,16 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
             stack = stack + size - task->copy_stack;
             size = task->copy_stack;
         }
-        ScanTaskStack(rescan, task, stack, stack + size);
+        char * stackend = stack + size;
+        if (task == RootTaskOfMainThread) {
+            stackend = (char *)GapStackBottom;
+        }
```
fingolfin, line 691:
```suggestion
        }

```

rbehrends: Done.

src/julia_gc.c
rbehrends: Umm, I don't see it or maybe I'm misunderstanding something? The guard pages are at the start of the buffer. I agree that "to top" is misleading and should be "towards the start of the stack buffer" or something like that. But the direction is high -> low address?

src/julia_gc.c
rbehrends: The Julia stack buffer information includes the guard pages. The information that Julia internally maintains is the start and end of the stack buffer, including guard pages. For some stacks, Julia may not even know where the guard pages end. This can in principle (with `COPY_STACKS`) be any task; in addition, it may change in future implementations.

I will try and spell out more clearly what we're doing here.

src/julia_gc.c
fingolfin: The thing is that in here, we don't necessary scan till the top (as we don't know where it is). As to whether we scan from "end to start" or "start to end" , I forgot that we scan "downwards"; so we could say "from end downwards to the start". But it would indeed nice to say "from bottom to top"; alas, again, I am concerned that "top" is not accurate here in general?

We could also think about calling `start` something else, like `max_top` or `absolute_top` or so?

src/julia_gc.c
rbehrends: Hmm. I was using adjustments as a general term (synonym for changes) that included optimizations. In any event, I've altered the comment to include the new phrasing.

src/julia_gc.c
rbehrends: Done.

src/julia_gc.c
rbehrends: It's actually start and size that we get, so how about:

```
// All stacks in Julia are associated with tasks and we can use
// jl_task_stack_buffer() to retrieve the buffer information (start and size)
// for that stack.
```

src/julia_gc.c
rbehrends: So, I've rephrased comments to refer to the start/end of the *stack buffer* that contains the stack, and the top/bottom of the *stack* and also did some rewording to describe the actual process.

src/julia_gc.c
fingolfin: fine by me

src/julia_gc.c
```diff
@@ -647,13 +684,25 @@ static void GapTaskScanner(jl_task_t * task, int root_task)
             rescan = 0;
     }
     if (stack) {
-        if (task->copy_stack) {
+        // task->copy_stack, if not zero or one, is an upper bound for the
+        // size of the stack.
+        if (task->copy_stack > 1) {
```
fingolfin, line 689: I am not sure this is quite right now -- the value `task->copy_stack == 1` should probably still be treated separately? I think it can only occur for "fresh" tasks which don't have a stack allocation yet; so I assume we could just do `if (task->copy_stack ==1) return;` or so?

rbehrends: Hmm, I'll go and have a look at that.

rbehrends: So, `task->copy_stack == 1` implies `task->bufsz == 0`. This means that the logic is accidentally correct, but the comment would have to change. But I think that making control flow more explicit by aborting stack scanning for `task->copy_stack == 1` is much clearer, so I'll go with that.

src/julia_gc.c
```diff
@@ -485,13 +518,16 @@ static void SafeScanTaskStack(PtrArray * stack, void * start, void * end)
...
-        // guard pages; accessing these results in segmentation
-        // faults. Julia catches those segmentation faults and
-        // longjmps to JuliaTLS->safe_restore; we use this
-        // mechamism to abort stack scanning when a protected
-        // page is hit. For this to work, we must scan the stack
-        // from top to bottom, so we see any guard pages last.
+        // The start of the stack buffer may be protected with guard
+        // pages; accessing these results in segmentation faults.
+        // Julia catches those segmentation faults and longjmps to
+        // JuliaTLS->safe_restore; we use this mechamism to abort stack
```
fingolfin, line 524:
```suggestion
        // JuliaTLS->safe_restore; we use this mechanism to abort stack
```

fingolfin:
Done in 74cfb2a8f66a31370aaa7cfd9bd05dbc563ccea6

PR #4061, opened 2020-06-24
Fix recovering with fg after Ctrl-z

frankluebeck:
At some stage GAP changed its behaviour when the user pressed  `Ctrl-z` (which stops GAP and allows for entering shell commands) and wanted to get back into the GAP session with `fg`. There was no GAP prompt, and input was echoed but not read.

This is fixed by this pull request. Now `fg` reenters the GAP session as expected. I have tested this for quite a while and did not experience any problems.

Checking the readline documentation indicates that the setting `rl_catch_signals = 1;` is correct. I don't know why former versions of GAP (which had this variable set to `0`) worked as expected.

## Text for release notes 
Pressing Ctrl-z in a GAP session and fg in the shell brings the user back into the interrupted GAP session.

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -2062,7 +2062,7 @@ static void initreadline(void)
   /* this should pipe signals through to GAP  */
   rl_already_prompted = 1 ;
 
-  rl_catch_signals = 0;
+  rl_catch_signals = 1;
   rl_catch_sigwinch = 1;
   /* hook to read from other channels */
   rl_event_hook = 0;


ThomasBreuer:
I believe that the proposed change is correct, but I cannot reproduce the problem.
Interrupting with `Ctrl-z` and then re-entering with `fg` works for me In GAP 4.11.0 and in the GAP master branch, with GASMAN or Julia GC.

frankluebeck:
It is interesting that @ThomasBreuer  cannot reproduce this. I tried several computers (different versions of Debian or Ubuntu), tried a larger number of terminal applications and different shells. But in no combination I can recover a GAP session with gap-4.10.2 or newer. There is no problem with gap-4.7.6 or older (also without the patch proposed here).

fingolfin:
I cannot reproduce "this" either, on several computers, including macOS and Linux machines (Debian, Ubuntu, Arch, ...). And I regularly use ctrl-Z to interrupt GAP sessions, and have been doing so fo many years. This is with GAP 4.10, 4.10.2, 4.11.0 and master.

Here with "this" I mean that I cannot reproduce that GAP echos but ignores input. For me it echos and handles the input. What it doesn't do (and never did, for me) was to redraw the prompt as it was so far . That would actually be nice to have (Julia does it). But I don't think this PR does it: I see no change in behavior with this PR at all.

frankluebeck:
Comparing with  @ThomasBreuer  I could now identify what is different between our setups. 

It turned out that the bad behaviour I have reported vanishes if I throw out the following line from my `~/.inputrc` (which is read when readline is initialized in a program):
```
$if GAP                                                                                                           
"\C-j": self-insert                                                                                               
$endif
```
I don't understand why this line causes a problem with `Ctrl-z`, but the problem is fixed by the present pull request.
(I want to keep the line in my `.inputrc` which allows for input lines in GAP which use several physical lines of the terminal.)

Actually, the problem also existed in former versions of GAP. The old version I had available for a quick test (4.7.6) had another bug: it did not read the `~/.inputrc` file on startup.

src/sysfiles.c
```diff
@@ -2062,7 +2062,7 @@ static void initreadline(void)
   /* this should pipe signals through to GAP  */
   rl_already_prompted = 1 ;
 
-  rl_catch_signals = 0;
+  rl_catch_signals = 1;
```
fingolfin, line 2065: So I am slightly concerned about other consequences this change might have with respect to signal handling, and interaction with other system; in particular for the integration of GAP into Julia.  I also don't really understand why/how this fixes the issue you are facing?

fingolfin:
I played with this in march but forgot to record my results here or really finish up... but basically while I could never reproduce the "There was no GAP prompt, and input was echoed but not read." what I always do see is that there is no GAP prompt, and input is being read, and echoed (so in a sense reverse to what Frank describes). Alas, this may be due to difference in the OS (macOS vs Linux) and Terminal emulators...

Anyway, I played some more and noticed that this patch did have some behaviour changes after all for me: Without it, I see this (on macOS):
```
gap> ^Z
[1]+  Stopped                 ./gap
$ fg
./gap      # <- I think this is printed by the shell?
1+2;       # <- this is what I type; I do get echo but no prompt
1+2;       # <- this and all afterward is printed after I press return
3
gap>
```
With this patch, I get the following; note that the second `1+2;` is gone.
```
gap> ^Z
[1]+  Stopped                 ./gap
$ fg
./gap
1+2;
3
gap>
```
So in this sense, this patch constitutes a minor improvement for me, too.

At the same time, I am still a little bit concerned about the issues when integrating this code in Julia (but this could be solved, e.g. by disabling the signals again via a preprocessor check for whether we are building for Julia). 

That said, even with the patch, there is no prompt for me after returning to the foreground, and sometimes editing behaves weirdly... 

I tried to solve the missing prompt by adding
```C
static int gap_signal_event_hook(void)
{
    rl_clear_visible_line();
    rl_redraw_prompt_last_line();
    rl_keep_mark_active();
    return 0;
}
```
and then in `initreadline`, do
```C
  rl_signal_event_hook = gap_signal_event_hook;
```
This kinda works, but it misses colors (i.e., the escape sequences for coloring the prompt and input don't seem to be resent). Also, unfortunately there seems to be no way in the signal hook to distinguish which signal triggered it. That means that if I e.g. press Ctrl-C, the prompt is also redrawn -- and in the wrong color (at least if as by default the prompt is blue and the remaining input red; then after ctrl-C the prompt is also red; probably because no escape codes for the colors are (re)sent, I did not yet try to dig deeper).

Issue #4052, opened 2020-06-17
Document what `^` does for group elements, link it in the index, and fix ^ index entries


fingolfin:
Several closely related issues regarding the `^` operator:

- for a beginner it is needlessly difficult to find out what `g^h` does when `g` and `h` are group elements. Sure, for group theory people it is usually "clear" that this must be h^{-1}*g*h`, but there are books and communities where it is different, so I think it'd be helpful to state this explicitly *somewhere* in the manual

- ideally that place also would have an index entry for `^` so that `?^` would find it

- in https://www.gap-system.org/Manuals/doc/ref/chapInd.html I see two entries for `^` but sorted into different positions; the same happens for other operators such as `\<`:
```
...
\XYZ 27.2 
^, for class functions 72.4 
\< 31.11-1 
    for associative words 37.3-2 
    for nonassociative words 36.2-2 
    for pcwords 46.2-1 
    for permutations 42.2-1 
    for two elements in a f.p. group 47.3-2 
    for two strings 27.6-2 
\* 31.12-1 
    for pcwords 46.2-2 
\*, for transformations 53.4 
...
\<, for transformations 53.4 
...
\= 31.11-1 
    for associative words 37.3-1 
    for nonassociative words 36.2-1 
    for pcwords 46.2-1 
    for permutations 42.2-1 
\=, for transformations 53.4 
\=, for two elements in a f.p. group 47.3-1 
    for two elements in a f.p. semigroup 52.3-1 
    for two strings 27.6-1 
...
\^ 31.12-1 
    for a field and a pair of integers 61.9-5 
    for a field and an integer 61.9-4 
\^, for a positive integer and a transformation 53.4 
    for a transformation and a permutation 53.4 
...
\{\} 21.3-1 
\{\}\:\= 21.4-1 
^ 4.13 
abelian number field 60.2-3 
...
```

ThomasBreuer:
Yes, the meaning of "conjugation" is apparently not defined explicitly in the GAP manual (contrary to the meaning of "commutator").

If one asks for `?^` then one finds two exact matches in the GAP Reference Manual.

- The one for `^` belongs to the section about "Arithmetic Operators" in the "Language" chapter, and the paragraph about the operator `^` could be extended by the definition for the case of two group elements.
- The one for `\^` belongs to the section about "Arithmetic Operations for Elements" in the "Domains" chapter, and gives the advice to consult the index entries for `\^`; thus an index entry with the key "^ for two group elements" should be created.

[...]

PR #4063, opened 2020-06-26
Update docs for operators on transformations & pperms

fingolfin:
This switches those docs to use ManSections, which results in
a nicer index.

This resolves the remaining index issues discussed in issue #4052 .


frankluebeck:
@ThomasBreuer : Yes, there was also a problem with sorting index entries which contained certain special characters in GAPDoc (HTML and LaTeX versions); see my comment in #4052.

PR #4071, opened 2020-07-10
Make the Julia GC threadsafe when used from GAP.jl.

rbehrends:
When running from multi-threaded Julia, we have to consider the case
that GAP can be called from more than one thread. In those cases, the
variable JuliaTLS may be incorrect. We therefore update it in NewBag(),
where it is depended upon for allocations. It had already been set when
a GC was called from a separate thread.

In order to avoid unnecessary overhead in the single-threaded case, we
do not change JuliaTLS for allocations when there is only one thread, as
determining the current thread-local state pointer can be somewhat
costly (which is why we cache it in JuliaTLS in the first place).

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -199,10 +199,16 @@ static jl_datatype_t * datatype_largebag;
 static UInt            StackAlignBags;
 static Bag *           GapStackBottom;
 static jl_ptls_t       JuliaTLS, SaveTLS;
+static BOOL            is_threaded;
 static jl_task_t *     RootTaskOfMainThread;
 static size_t          max_pool_obj_size;
 static UInt            YoungRef;
 static int             FullGC;
+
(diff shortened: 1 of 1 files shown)


src/julia_gc.c
```diff
@@ -199,11 +199,17 @@ static jl_datatype_t * datatype_largebag;
 static UInt            StackAlignBags;
 static Bag *           GapStackBottom;
 static jl_ptls_t       JuliaTLS, SaveTLS;
+static int             is_threaded;
```
fingolfin, line 202:
```suggestion
static BOOL            is_threaded;
```

rbehrends: Changed it.

fingolfin:
Backported to stable-4.11 in 27ebdf9e6bb7614d119d47dda20804a3c406ba30

PR #4073, opened 2020-07-13
Replace "sanity" by "consistency" in the GAP test suite

alex-konovalov:
There was a `sanity.g` file which is used for a weekly tests using Jenkins instance in St Andrews. I have renamed it to `consistency.g` and also improved some other wording in `tst` files, following the [Google style guide on writing inclusive documentation](https://developers.google.com/style/inclusive-documentation).

### DIFF
--- a/tst/teststandard/sort.tst
+++ b/tst/teststandard/sort.tst
@@ -135,7 +135,7 @@ gap> for i in [0..100] do
 >      od;
 >    od;
 
-# Just sanity check I really am making string reps
+# Just consistency check I really am making string reps
 gap> IsStringRep(lowAlpha{[1..0]}) and IsStringRep(lowAlpha{[1..10]});
 true
 gap> for i in [0..26] do
--- a/tst/teststandard/stablesort.tst
+++ b/tst/teststandard/stablesort.tst
@@ -148,7 +148,7 @@ gap> for i in [0..100] do
 >      od;
 >    od;
 
-# Just sanity check I really am making string reps
+# Just consistency check I really am making string reps
 gap> IsStringRep(lowAlpha{[1..0]}) and IsStringRep(lowAlpha{[1..10]});
 true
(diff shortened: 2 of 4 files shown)


tst/teststandard/sort.tst
```diff
@@ -135,7 +135,7 @@ gap> for i in [0..100] do
 >      od;
 >    od;
 
-# Just sanity check I really am making string reps
+# Just consistency check I really am making string reps
```
wilfwilson, line 138:
```suggestion
# Just check I really am making string reps
```

tst/teststandard/stablesort.tst
```diff
@@ -148,7 +148,7 @@ gap> for i in [0..100] do
 >      od;
 >    od;
 
-# Just sanity check I really am making string reps
+# Just consistency check I really am making string reps
```
wilfwilson, line 151:
```suggestion
# Just check I really am making string reps
```

tst/teststandard/sort.tst
fingolfin: Sorry @wilfwilson I didn't see your comment when I merged this. I agree with your suggestions and add a change for (using a completely different text) to https://github.com/gap-system/gap/pull/4082

fingolfin:
Do we really want to / need to backport this minor changes to 4.11? It doesn't seem worth the effort. Anyway, if somebody wants to do it, perhaps also pick up my follow-up commit 5204f4b7e7791d66e5796035623819e26a210959

fingolfin:
LOL, OK I just noticed that I already backported this in 419fd130001c04e9254455fb698231598a8ec8d2 ;-) OK, well, then why not...

alex-konovalov:
I wanted to backport this myself, for my Jenkins tests, but you did it earlier - thanks:)

PR #4077, opened 2020-07-16
Enhancements for Rewriting Systems and Cohomology

hulpke:
2-Cohomology code now deals better with relations that eliminate generators.
Better methods (calculation, lookup) for rewriting systems for simple groups (in particular of smallish order)


lib/gpfpiso.gi
```diff
@@ -1851,8 +2053,120 @@ local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
...
+    TryNextMethod();
+  fi;
+  iso:=ShallowCopy(a.idSimple);
+  if not IsBound(iso.parameter) then
+    iso.parameter:=iso.shortname;
+  fi;
+
+  # is it a pre-stored presentation?
+  f:=First(List(GAPInfo.RootPaths,x->Concatenation(x,"grp/simplerew.grp")),
+       IsExistingFile);
```
fingolfin, line 2156:
```suggestion
  f:=Filename(DirectoriesLibrary("grp"), "simplerew.grp");
```

fingolfin: Food for thought, not meant to hold up the PR:

It seems wasteful to parse this file every time you use it (which takes ~160ms on my 2018 MBP with very fast SSD), creating ~2MB of data, most of which is discarded again. Sure, 160ms is not much if you need to run computations that are much slower; but it adds up?

There are several ways to address this (as you are of course well aware):
- we could just always load the data at the start of GAP (but of course then we slow down GAP startup, and increase the base memory usage, for data the user might never use)
- we could load it the first time it is used, but "cache" it by storing it some variable (we'd still waste those 2 MB afterwards, if it is only used once, but this could be mitigated by using `InstallFlushableValue`)
- the data could be stored in multiple files: e.g. all data for each series in one file; or even one file per group: something like `simplerew-C-2-3.grp` for series S, parameters [2,3].
- on the long run, having a simple fast builtin database would be nice... I've been thinking about doing just such a thing in OSCAR for some time (but of course as always it's difficult to make something that fits everyone's needs...

lib/gpfpiso.gi
```diff
@@ -1712,6 +1892,24 @@ local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
...
+    k:=[];
+    for i in [1..Length(a)] do
+      k[i]:=[];
+      for j in [0..LogInt(QuoInt(pri[i],2),5)] do
+        k[i]:=Union(k[i],5^j*[-2..2]);
+      od;
+      k[i]:=Intersection(k[i],[-QuoInt(pri[i],2)..QuoInt(pri[i],2)]);
+      k[i]:=List(k[i],x->a[i]^x);
+    od;
+    borelelm:=Set(List(Cartesian(k),x->Product(x)));
```
fingolfin, line 1909:
```suggestion
    borelelm:=Set(Cartesian(k),x->Product(x));
```
or even
```suggestion
    borelelm:=Set(Cartesian(k),Product);
```

lib/gpfpiso.gi
```diff
@@ -1851,8 +2053,120 @@ local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
   return iso;
 end);
 
+# creator function for data -- for documentation
```
fingolfin, line 2056: Does this refer to `grp/simplerew.grp`, i.e. did this code generate `grp/simplerew.grp`? If so it'd be really helpful if
the comment said that. Even better if `grp/simplerew.grp` also said that the code used to generate it can be found here / is in function ConfluentRewritingData :-)

lib/grp.gi
```diff
@@ -5134,6 +5132,9 @@ local  i, U, gens,test;
   return gens;
 end);
 
+InstallMethod(SmallGeneratingSet,"generators subset",
+  [IsGroup and HasGeneratorsOfGroup],SMALLGENERATINGSETGENERIC);
```
fingolfin, line 5136: What's the motivation for this change? Do you plan to use `SMALLGENERATINGSETGENERIC` directly elsewhere?

lib/twocohom.gi
```diff
@@ -863,27 +864,34 @@ local field,fp,fpg,gens,hom,mats,fm,mon,kb,tzrules,dim,rules,eqs,i,j,k,l,o,l1,
...
-      # awkward.
-      if formalinverse[r[1][1]]>r[1][1] then
-        Error("generator vanishes?");
+      # Length of r[1] is 1. That is, this generator is not used!
+      m:=First(RelationsOfFpMonoid(mon),x->List(x,LetterRepAssocWord)=r);
+      m:=List(m,x->PreImagesRepresentative(fm,ElementOfFpMonoid(FamilyObj(One(mon)),x)));
+      m:=List(m,UnderlyingElement); # free group elements/words
+
+      if not IsOne(m[1]*Subword(m[2],1,1)) then
+        # Does the relation make a generator redundant (by expressing it iun the
```
fingolfin, line 889:
```suggestion
        # Does the relation make a generator redundant (by expressing it in the
```

lib/gpfpiso.gi
```diff
@@ -1402,23 +1453,130 @@ local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
...
+          i:=j+1;
+        fi;
+      else
+        if i<Length(w) then i:=i+2;fi;
+      fi;
+    until i>=Length(w);
+    return w;
+  end;
+
+#  oreduce:=function(wo,rules)
```
fingolfin, line 1523: What's this commented code about?

lib/gpfpiso.gi
```diff
@@ -1402,23 +1453,130 @@ local isob,isos,iso,gens,u,a,rels,l,i,j,bgens,cb,cs,b,f,k,w,monoid,
...
+
+#  pcgs:=SpecialPcgs(borel);
+#  borel:=Group(GeneratorsOfGroup(borel));
+#  SetPcgs(borel,pcgs);
+#  b:=List(LGFirst(pcgs),x->SubgroupNC(borel,pcgs{[x..Length(pcgs)]}));
+#  SetChiefSeries(borel,b);
+#
+#  cb:=ConfluentMonoidPresentationForGroup(borel);
+#
+#  isob:=cb.fphom;
```
fingolfin, line 1473: Any reason to keep this old code in commented out form?
```suggestion
```

lib/grp.gi
hulpke: Yes. (So far privately, but it might at some point become library code.)

lib/gpfpiso.gi
hulpke: Added.

lib/gpfpiso.gi
hulpke: @fingolfin 
I've thought about the issue of (re)loading, but in the end decided on the current solution as being most simple (and thus least prone to errors or in need of testing) and as not hogging up memory in permanence. 

The reason for doing so is that the only place this is going to be used is in calculating 2-cohomology (and extensions) for groups with larger simple composition factors. The data base will reduce really long KB runs (sometimes hours, a week in the case of M22) to a lookup. Even if that takes 10 seconds, it will be magnitudes faster.
Thus if this data base is used more than once, it will be in an outer loop iterating over many groups. For each of these groups the cohomology calculation still needs to iterate over all pairs of rules. This will take in the magnitude of minutes or more. A loading time of seconds will not make a meaningful difference.

There does not seem to be a "canned" solution (all data bases do their own thing), but of course if there was, this could easily be changed. (Though I don't think it would be a poster child for infrastructure either.)

fingolfin (approved):

Issue #1796, opened 2017-10-21
Check documentation is linked into Manual


ChrisJefferson:
After realising the documentation in some places isn't linked into the manual correctly, I wrote a small horrible script to find examples where there is some GAPDoc which (appears) to not be linked into the manual -- that is a `ManSection` which is not within a `#GAPDoc` section.

You can see the results of this script here: https://gist.github.com/ChrisJefferson/042d22644eed15a0ff4a28e77495d761 , or if you want to run the script yourself, here it in (run this from inside `lib`):

```
#!/bin/bash
for i in *; do
    echo $i
    awk '/#GAPDoc/{flag=0}/#\/GAPDoc/{flag=1;next}flag' $i | awk '/\/ManSection/{flag=0}/<ManSection/{flag=1}flag'
done;
```

It would be nice to go over each of these, and see what's going on. I suspect they fall into 5 categories

* My script got confused
* You find they are in the documentation. That means someone has probably cut+pasted it directly into an XML file in the 'doc' directory. It's possible the stuff in 'doc' (or 'ref') and 'lib' have drifted apart. Decide which is better, put that in the 'lib', and then link the 'lib' version into the xml file.
* There is good documentation, just not in the manual. Just link it in!
* There is horrible documentation. Improve it, then link it into the manual!
* The function is commented out, or shouldn't be called by users. Decide what we should do with the function (or ask!) and then we'll clean up one way or another.

This will require some knowledge of GAPDoc. My advice is to learn from GAPDoc documentation, but the important bits for our purposes:

* Build the documentation with `make manuals`
* A section of XML surrounded with `<#GAPDoc Label="bla">` <#/GAPDoc>` with be inserted into the manual where `<#Include Label="bla">` appears.

alex-konovalov:
Discovered that @ChrisJefferson already raised this question in #855, and I mentioned some ideas in #538.

fingolfin:
I somewhat disagree with the 5 categories @ChrisJefferson provides above, or rather the proposed actions. Some things are *intentionally* not documented in the reference manual, because they are  for internal use. But of course we still document them for developers!

Problem is: there is no good way to express that. Perhaps we can add a way to mark a function as "for internal use only" (say... by using a prefix for all and every internal function, such as `_` -- there was an issue suggesting that, but I can't find it right now). In my dream world, every function would either be documented, or marked as intentionally undocumented)

[...]

PR #4078, opened 2020-07-20
Fixed several typos in the documentation

ThomasBreuer:
... and removed some statements that are not true (anymore).

By the way:
Several chunks of documentation are currently not included in the Reference Manual.
(One gets an overview if one sets the `InfoGAPDoc` level to at least 2.)
This may be good in some cases, but I would prefer to include for example the entry for the function `Last`.


fingolfin:
`Last` absolutely should be in the manual, that's a bad oversight (PR to fix it welcome!).

As to other parts of the documentation not included in the reference manual, see also https://github.com/gap-system/gap/issues/1796

PR #4080, opened 2020-07-21
Allow registering global handlers for GAP_TRY/GAP_CATCH.

rbehrends:
This change allows packages to register global handlers in order to be notified when a `GAP_TRY` statement is started, when it succeeds, or when it results in an error.

The main motivation here is the GAP-Julia integration, where this change would allow us to handle GAP errors without creating unnecessary overhead by wrapping lots of short calls into GAP in `GAP_TRY/GAP_CATCH` sections.

Such a feature might also be useful in other situations.

### DIFF
--- a/src/sysjmp.c
+++ b/src/sysjmp.c
@@ -28,10 +28,9 @@ enum { signalSyLongjmpFuncsLen = 16 };
 
 static voidfunc signalSyLongjmpFuncs[signalSyLongjmpFuncsLen];
 
-Int RegisterSyLongjmpObserver(voidfunc func)
+int RegisterSyLongjmpObserver(voidfunc func)
 {
-    Int i;
-    for (i = 0; i < signalSyLongjmpFuncsLen; ++i) {
+    for (int i = 0; i < signalSyLongjmpFuncsLen; ++i) {
         if (signalSyLongjmpFuncs[i] == func) {
             return 1;
         }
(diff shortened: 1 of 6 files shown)


src/trycatch.h
```diff
@@ -18,6 +18,33 @@
...
+**
+*T  TryCatchMode
+*T  TryCatchHandler
+*F  RegisterTryCatchHandler()
+*F  InvokeTryCatchHandler()
+**
+**  The function RegisterTryCatchObserver() allows the installation of
+**  global exception handlers that are being called whenever GAP_TRY
+**  or CALL_WITH_CATCH() code is executed. The mode parameter signals
+**  whether it is has been called at the beginning of the section,
```
fingolfin, line 31:
```suggestion
**  whether it has been called at the beginning of the section,
```

src/sysjmp.c
```diff
@@ -43,6 +43,32 @@ Int RegisterSyLongjmpObserver(voidfunc func)
     return 0;
 }
 
+enum { tryCatchFuncsLen = 16 };
+
+static TryCatchHandler tryCatchFuncs[tryCatchFuncsLen];
+
+Int RegisterTryCatchHandler(TryCatchHandler func)
+{
+    Int i;
```
fingolfin, line 52:
```suggestion
    int i;
```

src/sysjmp.c
```diff
@@ -43,6 +43,32 @@ Int RegisterSyLongjmpObserver(voidfunc func)
...
+            tryCatchFuncs[i] = func;
+            return 1;
+        }
+    }
+    return 0;
+}
+
+void InvokeTryCatchHandler(TryCatchMode mode)
+{
+    Int i;
```
fingolfin, line 67:
```suggestion
    int i;
```

src/sysjmp.c
```diff
@@ -43,6 +43,32 @@ Int RegisterSyLongjmpObserver(voidfunc func)
...
+    for (i = 0; i < tryCatchFuncsLen; ++i) {
+        if (tryCatchFuncs[i] == func) {
+            return 1;
+        }
+        if (tryCatchFuncs[i] == 0) {
+            tryCatchFuncs[i] = func;
+            return 1;
+        }
+    }
+    return 0;
```
fingolfin, line 62: The meaning of the return value of this function should be documented. Could also change it to use BOOL/TRUE/FALSE.

rbehrends: The code is basically a carbon copy of `RegisterSyLongJmpObserver()`. I can easily incorporate all the changes; should I update `RegisterSyLongJmpObserver()` then too, while I'm at it?

rbehrends:
> Seems sensible enough to me. Of course a a test, e.g. in `tst/testlibgap`, would be super nice... In the meantime, at least a comment somewhere stating that this is (or rather: will be) used by the GAP-Julia interface would be appropriate. Else this might get remove again for being unused...

This is easily doable, yes, and makes good sense.

src/sysjmp.c
ChrisJefferson: Just to say, I wrote RegisterSyLongJmpObserver, and I agree these are sensible changes to make there as well, if you are happy to do so.

rbehrends: Okay, will do. :)

fingolfin:
Unfortunately we cannot easily backport this to GAP 4.11, as GAP_TRY & GAP_CATCH are the result of a substantial kernel refactoring in the `master` branch.

PR #4081, opened 2020-07-23
libgap: fix GAP_ValueGlobalVariable for automatic vars

fingolfin:
Automatic variables are relatively rare (not used by the GAP library, only by a few packages), but it makes sense to properly support them.


fingolfin:
Backport to stable-4.11 in commit 4a1e4580c5bb6603058d69aae0e34ec0db5cce45

PR #4084, opened 2020-07-24
kernel: simplify {GET,PEEK}_NEXT_CHAR and more

fingolfin:
... by getting back of the `Pushback` buffer and `RealIn`. This is
achieved by reserving the first byte of the `Input->line` buffer;
thus there is always one available byte before the current position
stored in `STATE(In)`, which we can use for pushback.

I hope this will enable some further tweaks and perhaps a minor optimization in startup speed (we spend a big part of that in GET_NEXT_CHAR)

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -327,9 +327,7 @@ const Char * GetInputLineBuffer(void)
 Int GetInputLinePosition(void)
 {
     GAP_ASSERT(IO()->Input);
-    // first byte of Input->line is reserved for the pushback buffer, so add 1
-    // subtract 1 from STATE(In) because TODO/FIXME
-    return (STATE(In) - 1) - (IO()->Input->line + 1);
+    return STATE(In) - GetInputLineBuffer();
 }
 
 UInt GetInputFilenameID(void)
--- a/src/io.c
+++ b/src/io.c
@@ -73,6 +73,8 @@ typedef struct {
     // a buffer that holds the current input line; always terminated
     // by the character '\0'. Because 'line' holds only part of the line for
     // very long lines the last character need not be a <newline>.
+    // The actual line data starts in line[1]; the first byte line[0]
+    // is reserved for the "pushback buffer" used by PEEK_NEXT_CHAR.
     char line[32768];
 
     // the next line from the stream as GAP string
@@ -187,9 +189,6 @@ struct IOModuleState {
     TypOutputFile OutputLogFileOrStream;
 
     Int NoSplitLine;
-
-    Char   Pushback;
-    Char * RealIn;
 };
 
 // for debugging from GDB / lldb, we mark this as extern inline
@@ -211,20 +210,9 @@ void LockCurrentOutput(Int lock)
 **  'GET_NEXT_CHAR' returns the next character from  the current input file.
 **  This character is afterwards also available as '*In'.
 */
-
-
-static inline BOOL IS_CHAR_PUSHBACK_EMPTY(void)
-{
-    return STATE(In) != &IO()->Pushback;
-}
-
 Char GET_NEXT_CHAR(void)
 {
-    if (STATE(In) == &IO()->Pushback) {
-        STATE(In) = IO()->RealIn;
-    }
-    else
-        STATE(In)++;
+    STATE(In)++;
 
     // handle line continuation, i.e., backslash followed by new line; and
     // also the case when we run out of buffered data
@@ -264,11 +252,7 @@ Char GET_NEXT_CHAR(void)
 // current line, when handling comment lines.
 Char GET_NEXT_CHAR_NO_LC(void)
 {
-    if (STATE(In) == &IO()->Pushback) {
-        STATE(In) = IO()->RealIn;
-    }
-    else
-        STATE(In)++;
+    STATE(In)++;
 
     if (!*STATE(In))
         GetLine();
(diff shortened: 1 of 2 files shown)


src/io.c
```diff
@@ -278,18 +260,18 @@ Char GET_NEXT_CHAR_NO_LC(void)
...
 
     // read next character
     GET_NEXT_CHAR();
 
-    // fake insert the previous character
-    IO()->RealIn = STATE(In);
-    STATE(In) = &IO()->Pushback;
-    return *IO()->RealIn;
+    // insert c back into the line buffer
+    --STATE(In);
```
ChrisJefferson, line 270: While it should never matter, maybe just add a sanity check GAP_ASSERT(STATE(In) >= IO->Input->line) after this?

src/io.c
```diff
@@ -531,6 +500,7 @@ UInt OpenInput (
     // start with an empty line
     STATE(In) = IO()->Input->line;
     STATE(In)[0] = STATE(In)[1] = '\0';
+    STATE(In)++;
```
ChrisJefferson, line 503: I'm not positive why we used to set the first two elements to 0 here (for look ahead?), so maybe we now need to set 3 elements to 0?

src/io.c
```diff
@@ -1269,18 +1239,19 @@ static Char GetLine(void)
     }
 
     /* bump the line number                                                */
-    if (IO()->Input->line < STATE(In) && *(STATE(In) - 1) == '\n') {
+    if (STATE(In)[-1] == '\n') {
```
ChrisJefferson, line 1242: This feels like it has different semantics to before, but I can't find any issues with testing

src/io.c
fingolfin: Yeah I also was wondering about that. But I couldn't come up with any reason. Note that this code was already exactly like this (well, modulo the `STATE` bit) in GAP 4.4, and that predates the look ahead, which is only used inside the floating point parsing code, which was added in GAP 4.5.

I don't mind clearing three bytes, but I'd prefer to understand *why* that might be necessary?

src/io.c
fingolfin: Well, this looks at the previous character, and in the past, that was only possible if the current character did not happen to be the first on the line. Otherwise, we'd perform an out of bound access. But now that byte 0 is reserved for "pushed back" chars after a peek, that's not an issue anymore...

That said, there are all kinds of corner cases to consider (like: what if the pushed back char is a newline? well, it can't be, because we only ever call `PEEK_NEXT_CHAR` if the current char is a dot `.` ; but of course that's not the kind of knowledge that is "obvious" when looking at this code.  Also, in theory, someone might want to use `PEEK_NEXT_CHAR` for another syntax extension; and then whoever did that most likely would not be aware of the need to check this code... Well, perhaps if we left a really good comment which mentions `PEEK_NEXT_CHAR` ? Hrm.

So perhaps the check should indeed be reinstated, just to be on the safe side?

src/io.c
ChrisJefferson: I also couldn't find any sensible reason for this either

ChrisJefferson: There are places where we access STATE(In)[1] (and even STATE(In)[2]), but they are only triggered if *STATE(In) is `\`, so they wouldn't be used in this situation, but maybe it was just for safety.

src/io.c
fingolfin: I added the check back, but reversed the first condition, as I found it easier to understand this way.

src/io.c
fingolfin: I've added this and a few more asserts. I also added some comments to explain a bit better what's going on.

src/io.c
```diff
@@ -328,24 +320,16 @@ Int GetInputLineNumber(void)
...
-            pos = 0;
-        return pos;
-    }
-    else {
-        return STATE(In) - IO()->Input->line - 1;
-    }
+    GAP_ASSERT(IO()->Input);
+    // first byte of Input->line is reserved for the pushback buffer, so add 1
+    // subtract 1 from STATE(In) because TODO/FIXME
+    return (STATE(In) - 1) - (IO()->Input->line + 1);
```
fingolfin, line 332: I wanted to explain what's going on here, but then realized I don't understand it anymore 😂 I'll have to carefully step through the code, I guess... but if anybody can explain it right now, feel free to tell

ChrisJefferson: I'm not sure either :)

src/io.c
fingolfin: Yeah I think that's actually a reasonable explanation: there was/is some code which looks ahead to (1) deal with line continuations and (b) deal with CR/LF vs. CR vs LF. But the latter is mostly done by GetLine2 or code even lower level; and the former (in its current form) is easily enough seen as being correct, as in: it won't access `In[1]` if `In[0]` is 0. And IMHO that's a good invariant to rely on: i.e., the above change is correct not because of some obscure code analysis, but rather because it simply adhere to the basic rule that a C string is terminated by a single null byte. If we have any code in there left which violates that, by expecting a second null byte (and otherwise producing wrong results), then *that* code is buggy and ought to be fixed.

I therefore feel confident to make this change; even if it causes a regression down the line, then this just means it helped us unearth a bug which was covered up before.

src/io.c
fingolfin: I gave up trying to understand it -- I think it simply was nonsense! IMHO it was a bug / misfeature. So I now dropped the -1, and instead adjusted `SyntaxErrorOrWarning` to cope with the changed result. Indeed, this now also gives values that make sense. For example:
```
gap> local;
Syntax error: expression expected
local;
^^^^^
```
now gives `startPos=0` and `pos=5` in `SyntaxErrorOrWarning`, while before it was -1 and 4.

fingolfin:
@ChrisJefferson happy with this now?

ChrisJefferson:
yep

PR #4085, opened 2020-07-26
Makefile.rules: don't install 'tst/'

fingolfin:
There is no need to install that; running the tests for the installed version can be done by using the tst from the source.


fingolfin:
Backport to stable-4.11 in commit f6dea835ac5113b336b0409835a638400538f1fb

PR #4086, opened 2020-07-30
julia: avoid access into Julia structs in Julia >= 1.6

fingolfin:
We currently access members of various structs provided by Julia (of type jl_task_t resp. jl_tls_t). This introduces an unwanted binary dependency on the precise Julia version GAP was compiled against; i.e. because of these, a binary compiled against Julia 1.3 cannot be used with Julia 1.5

To avoid this, the Julia folks merged some new APIs:
- `jl_active_task_stack` allows us to avoid access to `task->copy_stack` (see https://github.com/JuliaLang/julia/pull/36823)
- `jl_get_safe_restore` and `jl_set_safe_restore` allow us to avoid direct access to `tls->safe_restore` (see https://github.com/JuliaLang/julia/pull/36064)

In order to simultaneously support older and newer Julia versions, we use two tricks:
- we don't link during compile time against the new APIs; instead we use `dlsym` to query at runtime if they are available, and only then call them (this introduces as small overhead at the call sites, but this is negligible)
- for `task->copy_stack` this is enough as its offset has been unchanged between Julia 1.3 and Julia 1.6
- for `tls->safe_restore`, the offset *did* change once between Julia 1.4 and 1.5. Luckily, the change is the same on all 64 bit platforms: the offset shifted by 8. To adjust for that, we compare the Julia version used to compile GAP against that which runs GAP to compute a suitable offset for adjusting access to `tls->safe_restore`.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -14,6 +14,8 @@
 **  and gasman.c for two other garbage collector implementations.
 **/
 
+#define _GNU_SOURCE
+
 #include "julia_gc.h"
 
 #include "common.h"
@@ -29,6 +31,7 @@
 
 #include "bags.inc"
 
+#include <dlfcn.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


src/julia_gc.c
```diff
@@ -518,10 +521,64 @@ static void MarkFromList(PtrArray * arr)
...
+static void SetupSafeRestoreHandlers(void)
+{
+    get_safe_restore = dlsym(RTLD_DEFAULT, "jl_get_safe_restore");
+    set_safe_restore = dlsym(RTLD_DEFAULT, "jl_set_safe_restore");
+
+    // if the new API is available, just use it!
+    if (get_safe_restore && set_safe_restore)
+        return;
+
+    GAP_ASSERT(!set_safe_restore && !Get_safe_restore);
```
rbehrends, line 559: I think that `get_safe_restore` got capitalized by accident here. It only results in an error with `--enable-debug`, though.

fingolfin: Oops. That also made me realize we don't have `--enable-debug` on the Julia tests, which would have caught that. I've added a commit to fix that, too.

fingolfin:
This needs to be updated to reflect the changes in https://github.com/JuliaLang/julia/pull/36823

fingolfin:
Rebased and adapted to match the final version of `jl_active_task_stack ` which was merged into Julia in PR https://github.com/JuliaLang/julia/pull/36823. Assuming this passes, we can finally merge this PR and then backport it to stable-4.11

@ThomasBreuer ping just FYI

rbehrends (approved):

fingolfin:
Backported to stabled-4.11 in ef2ad9864bb51845491375af471ce77436e3899b

PR #4090, opened 2020-08-04
buildsys: tweak GAP options for `make check`

fingolfin:
In particular drop -x80 and the memory restrictions.


fingolfin:
Yes, I did test it locally :-)

PR #4091, opened 2020-08-04
kernel: remove STATE(In)

fingolfin:
Instead, we can simply always use `IO()->Input->ptr`. This further decouples the input code from global state.

### DIFF
--- a/src/gapstate.h
+++ b/src/gapstate.h
@@ -59,8 +59,6 @@ typedef struct GAPState {
 
     char Prompt[80];
 
-    char * In;
-
     /* From stats.c */
     Obj  ReturnObjStat;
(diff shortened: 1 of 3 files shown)


fingolfin:
Note that fetching `IO()->Input->ptr` involves strictly more work than fetching `STATE(In)`. However, I run benchmarks (measuring the time to start GAP) and saw no regression in performance. Nevertheless, the next step I am looking at involves pulling out the ``IO()->Input` and passing that into `GET_NEXT_CHAR` and friends as an argument.

ChrisJefferson:
Can you maybe try some benchmarking? My only worry is if this increases cost, as we are (I think) adding another indirection, and also IO() is probably more expensive than STATE(In). Of course, it might not actually matter.

PR #4094, opened 2020-08-05
Fix output redirection for firefox other browsers

### LINKED ISSUE #4087: Using the "firefox" help viewer on Linux creates empty file named "1"
### Observed behaviour

After executing `SetHelpViewer("firefox");` on Linux and then using the help viewer (which works fine) an empty file named "1" appears in the current working directory.

### Expected behaviour

No files are created in the current directory.

### Cause and fix

The templates for the shell commands used to start Firefox and several other browsers on Linux https://github.com/gap-system/gap/blob/c7696ae2bd7f2e7ac4848057de0229a0b17bf68f/lib/helpview.gi#L224 have the redirect `2>1` which redirects standard error into the file named "1", instead it should be `2>&1` which redirects stderr to the same destination as stdout.

```
% gap
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv7
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2019.09.04, AutPGrp 1.10.2, Browse 1.8.8, CaratInterface 2.3.3, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, CTblLib 1.2.2, FactInt 1.6.3, 
             FGA 1.4.0, Forms 1.2.5, GAPDoc 1.6.3, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4, LAGUNA 3.9.3, orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, PrimGrp 3.4.0, RadiRoot 2.8, recog 1.3.2, ResClasses 4.7.2, 
             SmallGrp 1.4.1, Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.5, utils 0.69
```

fingolfin:
... when they are used as help viewers. In the old version of the code,
output from stderr was redirect into a file named "1"; now it is redirected
to stdout, as likely was the intention.

It is still not clear to me how useful that is: Will the user see the output
on stdout? If so, won't they also see the output to stderr? So why do we
need to redirect it? But I have no way to test any of this right now, so
I went with the minimal "obvious" fix.

Perhaps @frankluebeck has some insight into this?

Fixes #4087

### DIFF
--- a/lib/helpview.gi
+++ b/lib/helpview.gi
@@ -221,7 +221,7 @@ else # UNIX but not Mac OS X
   HELP_VIEWER_INFO.firefox := rec(
   type := "url",
   show := function(url)
-    Exec(Concatenation("firefox \"file://", url,"\" >/dev/null 2>1 &"));
+    Exec(Concatenation("firefox \"file://", url,"\" >/dev/null 2>&1 &"));
   end
   );
 
@@ -229,7 +229,7 @@ else # UNIX but not Mac OS X
   HELP_VIEWER_INFO.chrome := rec(
   type := "url",
   show := function(url)
-    Exec(Concatenation("chromium-browser \"file://", url,"\" >/dev/null 2>1 &"));
+    Exec(Concatenation("chromium-browser \"file://", url,"\" >/dev/null 2>&1 &"));
   end
   );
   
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
I assume the plan here is that with the >/dev/null, we are redirecting all output to dev/null?

However, I agree with this as a required fix of course

PR #4095, opened 2020-08-06
Minor fixes and performance enhancements

### LINKED ISSUE #4089: SemidirectProduct error if group has no generators
### Observed behaviour
```
gap> G:=SymmetricGroup(3);;
gap> H:=SubgroupProperty(G,x->Order(x)=1);
Group(())
gap> T:=Group(());
Group(())
gap> H=T;
true
gap> GeneratorsOfGroup(H);
[  ]
gap> GeneratorsOfGroup(T);
[ () ]
gap> T=Group(GeneratorsOfGroup(T));
true
gap> H=Group(GeneratorsOfGroup(H));
Error, usage: Group(<gen>,...), Group(<gens>), Group(<gens>,<id>) at /proc/cygdrive/C/gap-4.10.0-x86_64/lib/grp.gi:4563 called from
<function "Group">( <arguments> )
 called from read-eval loop at *stdin*:9
you can 'return;' to continue
brk> quit;
```
### Expected behaviour
```
gap> H=Group(GeneratorsOfGroup(H));
true
```
```
 ┌───────┐   GAP 4.10.0 of 01-Nov-2018
 └───────┘   Architecture: x86_64-unknown-cygwin-default64
 Configuration:  gmp 6.1.0, readline
 Packages:   AClib 1.3.1, Alnuth 3.1.0, AtlasRep 1.5.1, AutoDoc 2018.09.20, AutPGrp 1.10, Browse 1.8.8, CRISP 1.4.4,
             Cryst 4.1.18, CrystCat 1.1.8, CTblLib 1.2.2, FactInt 1.6.2, FGA 1.4.0, GAPDoc 1.6.2, IO 4.5.4,
             IRREDSOL 1.4, LAGUNA 3.9.0, Polenta 1.3.8, Polycyclic 2.14, PrimGrp 3.3.2, RadiRoot 2.8,
             ResClasses 4.7.1, SmallGrp 1.3, Sophus 1.24, SpinSym 1.5, TomLib 1.2.7, TransGrp 2.0.4, utils 0.59
```

hulpke:
Catch case of semi/subdirect products with trivial factors and no generators (resolve  #4089)

Performance improvements to automorphism groups (reduce chance of rare case of being caught in bad random choices).
[A test for this is a repeated call of
```
g:=PerfectGroup(IsPermGroup,491520, 559);a:=AutomorphismGroup(g);
```
but it would be too slow for automated tests]

Option to create unimodular matrices with larger entries (not just -10..10)

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -55,7 +55,7 @@ local C,M,p,all,gens,sub,q,hom,fp,rels,new,pre,sel,i,free,cnt;
 end);
 
 BindGlobal("AGSRPrepareAutomLift",function(G,pcgs,nat)
-local ocr, fphom, fpg, free, len, dim, tmp, L0, S, R, rels, mat, r, RS, i, g, v;
+local ocr,fphom,fpg,free,len,dim,tmp,L0,S,R,rels,mat,r,RS,i,g,v,cnt;
 
   ocr:=rec(group:=G,modulePcgs:=pcgs);
   fphom:=IsomorphismFpGroup(G);
@@ -72,13 +72,18 @@ local ocr, fphom, fpg, free, len, dim, tmp, L0, S, R, rels, mat, r, RS, i, g, v;
   ocr.module:=GModuleByMats(
     LinearActionLayer(G,ocr.generators,ocr.modulePcgs),ocr.field);
   ocr.moduleauts:=MTX.ModuleAutomorphisms(ocr.module);
+
   if Size(ocr.moduleauts)>
       # Finding the relations comes at a cost that needs to be plausible
       # with searching multiple times through the automorphism group. This
       # order bound is a heuristic that seems to be OK by magnitude.
       321
     then
-    ocr.trickrels:=AGSRFindRels(nat);
+    cnt:=0;
+    repeat
+      ocr.trickrels:=AGSRFindRels(nat);
+      cnt:=cnt+1;
+    until ocr.trickrels<>fail or 2^cnt>10*Size(ocr.moduleauts);
   else
     ocr.trickrels:=fail;
   fi;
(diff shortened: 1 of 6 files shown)


lib/gprd.gi
```diff
@@ -560,7 +560,7 @@ local gc,hc,S,info;
 	    homomorphisms:=[gh,hh],
 	    projections:=[Projection(S,1)*InverseGeneralMapping(gc),
 			  Projection(S,2)*InverseGeneralMapping(hc)]);
-  S:=Group(GeneratorsOfGroup(S));
+  S:=Group(GeneratorsOfGroup(S),One(S));
```
fingolfin, line 563: Do we have a test case for those semidirect product fixes?

lib/matrix.gd
```diff
@@ -1778,11 +1778,17 @@ DeclareGlobalFunction( "RandomMat" );
 ##  returns a new random mutable <A>m</A><M>\times</M><A>m</A> matrix with integer
 ##  entries that is invertible over the integers.
 ##  Optionally, a random source <A>rs</A> can be supplied.
+##  If the option <A>domain</A> is given, random seection is made from <A>domain</A>, otherwise
```
fingolfin, line 1781:
```suggestion
##  If the option <A>domain</A> is given, random selection is made from <A>domain</A>, otherwise
```

lib/autsr.gi
```diff
@@ -924,10 +924,10 @@ local ff,r,d,ser,u,v,i,j,k,p,bd,e,gens,lhom,M,N,hom,Q,Mim,q,ocr,split,MPcgs,
     # try to find some further generators
     if Size(AQP)/Size(Aperm)>100 then
       for j in SpecialPcgs(RadicalGroup(AQP)) do
-	if precond(j) then cond(j);fi;
+	cond(j);
```
fingolfin, line 927: The commit "REBASE: precond does not always work" could/should perhaps be folded into the commit which it modifies? I.e., "ENHANCE: More efforts on automorphism lifting".

I also don't understand what the "REBASE: " prefix is about. If at all, this seems to be a "fixup", nothing is being rebased?

hulpke: REBASE (for me) means exactly what you write. This commit should be squashed it into another commit. (It is only kept separately initially to make it easy to undo if it would cause problems.)

lib/gprd.gi
hulpke: I did not prepare one (as this line is anyhow run through by the tests, and I did not have test cases for all of the situations potentially caught).

PR #4097, opened 2020-08-07
kernel: reduce arguments of FuncSHELL

fingolfin:
... and simplify a bunch of code using that, e.g. replace `THREAD_SESSION` by `SESSION`.

### DIFF
--- a/lib/error.g
+++ b/lib/error.g
@@ -174,8 +174,7 @@ Unbind(ErrorInner);
 BIND_GLOBAL("ErrorInner", function(options, earlyMessage)
     local   context, mayReturnVoid,  mayReturnObj,  lateMessage,
             x,  prompt,  res, errorLVars, justQuit, printThisStatement,
-            printEarlyMessage, printEarlyTraceback, lastErrorStream,
-            shellOut, shellIn;
+            printEarlyMessage, printEarlyTraceback, lastErrorStream;
 
     context := options.context;
     if not IsLVarsBag(context) then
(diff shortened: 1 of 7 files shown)


lib/error.g
```diff
@@ -349,16 +348,8 @@ BIND_GLOBAL("ErrorInner", function(options, earlyMessage)
     else
         prompt := "brk> ";
     fi;
-    shellOut := "*errout*";
-    shellIn := "*errin*";
```
fingolfin, line 353: Ah, I am stupid. Of course we must use errout/errin here. So I need to add back one argument, something like `BOOL isBreakPrompt` or so.

ChrisJefferson (approved):

ChrisJefferson:
I point out this looks like it might break JupyterKernl, which calls SHELL -- anyone who wants to fix that could use NumberArgumentsFunction to tell which SHELL they are calling.

fingolfin:
Hmm, I explicitly searched for packages using `SHELL`, and found none. I just had another look at `JupyterKernel` and still couldn't find anything. Am I missing something?

ChrisJefferson:
False alarm, I wasn't up to date. It used to use it, but a git pull removed it.

src/gap.c
```diff
@@ -177,7 +177,22 @@ static Obj Shell(Obj    context,
...
   STATE(ErrorLLevel) = 0;
   Int oldRecursionDepth = GetRecursionDepth();
-  
+
+  if (breakLoop) {
+    inFile = "*errin*";
+    outFile = "*errout*";
+  }
+#ifdef HPCGAP
+  if (ThreadUI) {
```
fingolfin, line 186:
```suggestion
  else if (ThreadUI) {
```

PR #4099, opened 2020-08-11
Passing a `.tst` file as argument to `gap` will now invoke `Test` on it, instead of trying to `Read` it (which will fail)

fingolfin:
... instead of Read / READ_NORECOVERY. This is purely a convenience
feature: now one can for example run

    gap tst/testinstall/perm.tst

which is equivalent to

    gap -c 'Test("tst/testinstall/perm.tst");'

but shorter and easier to enter via shell command completion.

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -937,6 +937,9 @@ InstallAndCallPostRestore( function()
         f := GAPInfo.InitFiles[i];
         if IsRecord(f) then
             status := READ_NORECOVERY(InputTextString(f.command));
+        elif EndsWith(f, ".tst") then
+            Test(f);
+            status := true;
         else
             status := READ_NORECOVERY(f);
         fi;


ChrisJefferson:
I do this at least once a month, so I like the change

lib/init.g
```diff
@@ -937,6 +937,9 @@ InstallAndCallPostRestore( function()
         f := GAPInfo.InitFiles[i];
         if IsRecord(f) then
             status := READ_NORECOVERY(InputTextString(f.command));
+        elif EndsWith(f, ".tst") then
+            Test(f);
+            status := true;
```
fingolfin, line 942: Of course one could try to make this more robust: e.g. first check if `f` refers to an existing and readable file, and if not... dunno, what? Right now, one gets a break loop, which says "type 'quit;' to quit to outer loop" but if one does that, GAP actually quits (because the main REPL has not yet been started). Hmm

alex-konovalov:
Amazing!

PR #4100, opened 2020-08-14
WIP: Normalisers and automorphisms.

DominikBernhardt:
WIP


lib/morpheus.gd
```diff
@@ -586,3 +586,24 @@ DeclareOperation("GQuotients",[IsGroup,IsGroup]);
...
+## 
+## <Description>
+## This function returns representatives for each element of the outer 
+## automorphism group of the finite group <A>grp</A>.<P/>
+## The optional argument <A>aut</A>, if present, must be the automorphism
+## group of <A>grp</A>. In this case, the function does not need to 
+## comupute the automorphism group of <A>grp</A>. Note that it is
+## not checked whether <A>aut</A> is indeed the automorphism group
+## of <A>grp</A>. <P/>
+## Note: This Algorithm might become slow if the automorphism group
```
ssiccha, line 607:
```suggestion
## Note: This algorithm is (currently) only useful if the outer automorphism group is very small.
```

lib/gpprmsya.gd
```diff
@@ -159,6 +159,25 @@ DeclareAttribute("OrbitStabilizingParentGroup",IsPermGroup);
 
 DeclareGlobalFunction("NormalizerParentSA");
 
+#############################################################################
+##
+#F NormalizerSimpleGroupInParentSymmetricViaAutomorphisms( <grp>, [,<aut>]) )
```
ssiccha, line 164: Now that I think about it, the name is already quite long so we could also call it
`NormalizerSimpleGroupInParentSymmetricGroupViaAutomorphisms`.

lib/gpprmsya.gi
```diff
@@ -1430,6 +1430,42 @@ local P;
   fi;
 end);
 
+
+BindGlobal( "NormalizerSimpleGroupInParentSymmetricViaAutomorphisms",
+function( T, aut... ),
```
ssiccha, line 1435: I think we need `T` to be transitive.

lib/morpheus.gi
```diff
@@ -3001,3 +3001,42 @@ local cl,cnt,bg,bw,bo,bi,k,gens,go,imgs,params,emb,clg,sg,vsu,c,i;
   Info(InfoMorph,1,Length(emb)," found -> ",Length(cl)," homs");
   return cl;
 end);
+
+# TODO: Maybe use Babai-Szemeredi Cube-method, see Permutation Group 
+# Algorithms by Seress, Section 4.4.
+BindGlobal("RepresentativesOuterAutomorphismGroup", 
+function(T, args...)
+  local order_out, out_reps, candidate, rep, i;
+  if Length(aut) = 1 then
```
ssiccha, line 3010: `aut` -> `args`

lib/morpheus.gi
```diff
@@ -3001,3 +3001,42 @@ local cl,cnt,bg,bw,bo,bi,k,gens,go,imgs,params,emb,clg,sg,vsu,c,i;
   Info(InfoMorph,1,Length(emb)," found -> ",Length(cl)," homs");
   return cl;
 end);
+
+# TODO: Maybe use Babai-Szemeredi Cube-method, see Permutation Group 
+# Algorithms by Seress, Section 4.4.
+BindGlobal("RepresentativesOuterAutomorphismGroup", 
+function(T, args...)
+  local order_out, out_reps, candidate, rep, i;
```
ssiccha, line 3009: update this

lib/morpheus.gd
```diff
@@ -586,3 +586,24 @@ DeclareOperation("GQuotients",[IsGroup,IsGroup]);
...
+## of <A>grp</A> is large.
+DeclareGlobalName("RepresentativesOuterAutomorphismGroup");
```
ssiccha, line 609:
```suggestion
DeclareGlobalName("RepresentativesOuterAutomorphismGroupViaRandomSearch");
```
I'd specify the way we search for the representatives. Why? I feel without the suffix `ViaRandomSearch` the name of the function suggests that it is *the* GAP function to compute representatives of the elements of the outer automorphism group. But the function (at least in its current form) is rather specialized, i.e. only applicable where the outer automorphism group is really small.

hulpke:
What one could do (and I have done in similar cases) is to represent the outer automorphism group as (say) permutation group (it is an iterated semidirect product of cyclic groups, and e.g. `DataAboutSimpleGroup` gives you a list of its subgroups in the component `allExtensions`, and one could even have it return an isomorphic group) and use `SubgroupProperty` to search for the subgroup that is induced by the normalizer. This avoids redundant tests.
Of course what this will not help with is the work to write down at least generators  as outer automorphisms.

lib/gpprmsya.gd
DominikBernhardt: @fingolfin said in the slack channel that we could just call it `NormalizerOfSimpleGroupInSymmetricGroup`, which I would support. Is it clear enough that this refers to the normaliser on the `MovedPoints`?

DominikBernhardt: Also, I don't know what the additional information `ViaAutomorphisms` might help the user? Why do we have it in the function name?

lib/gpprmsya.gi
DominikBernhardt: So we assume that `T` is finite non-abelian and simple. _If_  the action is faithful, it is either transitive or trivial. As we compute the normalizer in the symmetric group on the moved points, I guess we want to assume that the action of `T` on its moved points is faithful? If so, the only thing we need to check is that the action is not trivial. What do you think @ssiccha ?

lib/morpheus.gd
DominikBernhardt: Updated

lib/morpheus.gi
DominikBernhardt: Updated and changed the subsequent calls to `Length(aut)` to `Length(args)` as well.

lib/morpheus.gi
DominikBernhardt: Done

lib/morpheus.gd
DominikBernhardt: I updated this for now.

lib/gpprmsya.gi
DominikBernhardt: I don't think `ConjugatorIsomorphism` requires `T` to be transitive, so this should not cause any problems.

lib/morpheus.gd
DominikBernhardt: We added a TODO for now pointing to Alex' suggestion below.

ssiccha:
@hulpke Thanks a lot for the suggestion! We'll look into it. :)

ssiccha:
@hulpke Are there situations in which GAP can compute the automorphism group, but `DataAboutSimpleGroup` will not return all extensions? Because if not, we could remove `RepresentativesOuterAutomorphismGroup` entirely or at least parts of it.

ssiccha:
@hulpke 
> and one could even have it return an isomorphic group

@DominikBernhardt has looked into `DataAboutSimpleGroup` and for the `PSL(2,7)` the component `allExtensions` is `[ [ 2, "2" ] ]`. How could we get a permutation group that is isomorphic to the outer automorphism group from that?

hulpke:
> @hulpke Are there situations in which GAP can compute the automorphism group, but `DataAboutSimpleGroup` will not return all extensions? Because if not, we could remove `RepresentativesOuterAutomorphismGroup` entirely or at least parts of it.

No, there should not be any such situations (the one that existed was a bad bug). In fact the code for normalizer in Sn uses the information to see whether a simople group can have normalizing elements outside the centralizer.

And with the new code in, it would not be hard to always return a faithful representation of the outer automorphism group as well. (The main difficulty is to match the naming of the extension to the ones used in the character table library.)

ssiccha:
@hulpke I'm a bit at a loss how to actually do that :sweat_smile: 

> And with the new code in, it would not be hard to always return a faithful representation of the outer automorphism group as well. (The main difficulty is to match the naming of the extension to the ones used in the character table library.)

Can you give me a starting point for where to look into that? Or would that be easier and quicker to explain in a short video call?

hulpke:
If you'd find it useful, I can add a component to `DataAboutSimpleGroup` that holds the image of such a representation. Let me know.

ssiccha:
@hulpke That would be awesome! But, don't we also need the homomorphism?

hulpke:
Yes, the homomorphism from Aut(T) to Out(T) would be more useful (but will be much harder, and I cannot offer that).
It would obviously also require something like constructive identification.
Still, knowledge of the outer automorphisms (and where they come from) should be of help, as they indicate when one can stop searching.

hulpke:
Very stupid question: The function `NormalizerParentSA` already has some such tests using the automorphism group in. How does your code compare? (OR are you interested in imprimitive cases?)

flyingapfopenguin (approved):

fingolfin:
@hulpke I guess the fact that `NormalizerParentSA` has zero comments explaining what it is for, what it computes (what does "SA" stand for), or how it works, may be a major reason why nobody has discussed how it compares so far... :-)

The only thing I could find out was from the commit log (the "secret" internal one which is not generally available on GitHub) is that the initial version of that function was added by you 2002-05-08 with this commit message:

> Dedicated `StabChain' method for Sn.
> Special methods for normalizer and conjugating element in Sn (first map/fix
> orbits, block systems; then take normalizers of the respective actions and
> combine to wreath.) AH

and it is called by `DoNormalizerSA` which in turn is installed as `NormalizerOp` method for `IsNaturalSymmetricGroup` and `IsNaturalAlternatingGroup` -- so I am guessing "SA" stands for "Symmetric or Alternating"?

@DominikBernhardt do you have a reply?

DominikBernhardt:
Sorry @hulpke  and @fingolfin , I forgot to reply. 
I tested both functions on my machine under the assumption that the first argument of `NormalizerParentSA` is the containing symmetric group of the second argument of that function and that the second argument is the group we want to compute the normaliser of. It seems, to my surprise, that both functions are comparable in their runtime on groups such as `PSL(2,41)`, `PSL(3,3)`, `M11`, `M24` and `PSU(4,2)`. 
@ssiccha and my motivation to implement this dedicated version of normaliser computation was that sometimes computing normaliser of simple groups of Lie-type in a permutation representation that is not the smallest permutation representation is slow. However, when running into these problems, I always called `Normaliser` and not `NormalizerParentSA` so maybe in some cases dispatching to `NormalizerParentSA` takes a while? I can try to recreate this situation, but I can't quite recall which groups I looked at when first running into this issue.

hulpke:
Well, it did not help that we expunge the comment messages. I'll add a comment to the file.
(The reason for the function were a number of papers in ISSAC that made preposterous claims about the normalizer in GAP being bad, based on computing in the full symmetric group)

SA is ``symmetric or alternating''. First parameter is Sn or An, second is the subgroup. What is computed is a subgroup that:

- Stabilizes orbits and block systems
- If primitive:
   = Normalize socle
   = Use the primitive groups library to find possible normalizing overgroups, in particular find whether group must be self-normalizing
   = If nonabelian socle, find which automorphisms could be Sn-induced.

ssiccha:
@hulpke So, does that mean that if the group `G` we're searching the normalizer of is simple and imprimitive, then `NormalizerParentSA`
1. doesn't do much or
2. recurses to a primitive version of `G` and calls itself again?

hulpke:
@ssiccha Ah -- I did not realize that your focus was imprimitive simple groups.

If an imprimitive (may be simple) group is given, and the block system is ``unique'' so that it must be styabilizerd by the normalizer, it will return a wreath product A\wr B where A normalizes the block stabilizers action on the block and B normalizes the action on the blocks (Note that for simple groups the action on the blocks is (ultimately) primitive, so this recursive call will use in the primitive case.) viz:

```
gap> g:=MathieuGroup(11);
Group([ (1,2,3,4,5,6,7,8,9,10,11), (3,7,11,8)(4,10,5,6) ])
gap> g:=MathieuGroup(11);;
gap> s:=Stabilizer(g,[1,2],OnTuples);;
gap> a:=Image(FactorCosetAction(g,s));
<permutation group of size 7920 with 2 generators>
gap> NrMovedPoints(a);
110
gap> s:=SymmetricGroup(110);
Sym( [ 1 .. 110 ] )
gap> p:=NormalizerParentSA(s,a);
<permutation group of size 285348072390194626560 with 2 generators>
gap> AllBlocks(p);  # Note it does not chose the blocks of size 10, as a has 2 such block systems!
[ [ 1, 102 ] ]
gap> Size(Action(p,Orbit(p,[1,102],OnSets),OnSets));
7920
gap> Size(Normalizer(p,a));time;
15840
58
```
In the imprimitive case, it does not try to use automorphisms per se.

If a primitive (again may be simple) group with nonabelian socle is given, the routine either tries to find possible larger groups from the primitive groups library, or tries to find which automorphisms are induced by the normalizer.

If the normalizer is not computes in a symmetric or alternating group it oprobably is not worth to go through NormalizerParentSA.

Yes, it might be useful to move the ``find normalizer by automorphisms'' into a helper function. Or NormalizerParentPA could call your routine in the case of a simple socle.

PR #4103, opened 2020-08-19
Small corrections

hulpke:
Fix two reported bugs, one in the lattice for generic direct products, one with units in nonassociative algebras.

No test file for the first error, as this example runs very long, but making it smaller avoids
the error and  does not trigger the error.


wilfwilson (approved):

fingolfin:

PR #4105, opened 2020-08-24
admit an optional third argument in `OrderMod` ...

ThomasBreuer:
... that is a multiple of the order one wants to compute.
Note that the current function computes `Lambda( m )` as the "default multiple", and then divides this number by its prime divisors until the order is found.
If one knows a reasonable multiple in advance then the perhaps expensive computation of `Lambda( m )` can be skipped.

## Text for release notes 
The function <code class="code">OrderMod</code> now admits an optional third argument that is a multiple of the order one wants to compute.

### DIFF
--- a/lib/numtheor.gd
+++ b/lib/numtheor.gd
@@ -161,11 +161,11 @@ DeclareOperation( "Lambda", [ IsObject ] );
 
 #############################################################################
 ##
-#F  OrderMod( <n>, <m> )  . . . . . . . .  multiplicative order of an integer
+#F  OrderMod( <n>, <m>[, <bound>] ) . . .  multiplicative order of an integer
 ##
 ##  <#GAPDoc Label="OrderMod">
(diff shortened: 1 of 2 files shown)


PaulaHaehndel (approved):
Thanks for implementing this.

One may could add a sentence in the documentation about what happens if the given bound is incorrect.
I also would assume that release notes are needed.

ThomasBreuer:
@PaulaHaehndel thanks for your suggestions.

By the way:
One of the tests failed, the last lines of output are as follows.
```
testing: /home/travis/build/gap-system/gap/tst/testinstall/semigrp.tst
# line 186 of 506 (36%)cannot allocate memory for thread-local data: ABORT
```
(I think it is not very likely that the code change is the reason for this failure.)

PaulaHaehndel (approved):

wilfwilson (approved):

PR #4106, opened 2020-08-28
Fix test by removing output

james-d-mitchell:
This PR removes a line of output from a test in a file that was apparently added on 19/08/2020, which cases the Semigroups travis-ci job to fail:

https://travis-ci.org/github/gap-packages/Semigroups/jobs/721632136

There are probably other solutions, I'm not invested in this particular one, happy to revise if necessary.

### DIFF
--- a/tst/testbugfix/2020-08-19-Uniz.tst
+++ b/tst/testbugfix/2020-08-19-Uniz.tst
@@ -1,5 +1,4 @@
 # Units in nonassociative algebra # 4096
 gap> a:=OctaveAlgebra(GF(2));
 <algebra of dimension 8 over GF(2)>
-gap> u:=Units(a);
-<magma with 5 generators>
+gap> u:=Units(a);;


james-d-mitchell:
Aha so this is a bug in the Semigroups package then and not GAP. I’ll close this PR and try to fix whatever is wrong in Semigroups itself. Thanks @hulpke

fingolfin:
@james-d-mitchell in case you didn't spot it already, the bug is visible here:
```
gap> R:=OctaveAlgebra(GF(2));;
gap> ApplicableMethod(MagmaByGenerators,[[One(R)]],6);
#I  Searching Method for MagmaByGenerators with 1 arguments:
#I  Total: 2 entries
#I  Method
1
 : ``MagmaByGenerators: for a finite multiplicative element collection'' at GAPROOT/pkg/semigroups-3.2.3/gap/semigroups/semigrp.gi:184, value: 9
#I  Function Body:
<Operation "SemigroupByGenerators"><Operation "SemigroupByGenerators">
```

So a method for `MagmaByGenerators` is installed which is accepting non-associative elements as generators, it seems.

james-d-mitchell:
Thanks @fingolfin, I did find this already, and I'm happy to report there's an easy fix.

alex-konovalov:
@james-d-mitchell is this fixed in the released versions of Semigroups?

james-d-mitchell:
@alex-konovalov I believe it is fixed in (at least the most recent) release version of Semigroups.

alex-konovalov:
@james-d-mitchell It seems that it still see 

```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-master-test/GAPCOPT\
S/64build/GAPTARGET/bugfix/label/kovacs/GAP-master-snapshot/tst/testbugfix/202\
0-08-19-Uniz.tst:4
# Input is:
u:=Units(a);
# Expected output:
<magma with 5 generators>
# But found:
<semigroup of size 120, with 5 generators>
########
```

which is the same as https://travis-ci.org/github/gap-packages/Semigroups/jobs/721632136, with Semigroups 3.4.0 of 28/08/2020.

james-d-mitchell:
Ah, right you are @alex-konovalov looking at the commit history, this was resolved in:

https://github.com/semigroups/Semigroups/commit/89c718600d55b63d266e973bfd20551ae38c82d0

but this was immediately after the release of 3.4.0. I will try to make a release that fixes this asap, sorry for the delay!

PR #4111, opened 2020-09-07
Add option `generatorNames` to allow changing the names of the generators of a group being constructed

hulpke:
Introduced an option that allows to change the names of the free group generators through an intermediate function.

Also some manual fix for the tutorial.

### DIFF
--- a/doc/ref/grplib.xml
+++ b/doc/ref/grplib.xml
@@ -99,6 +99,31 @@ calculations.
 <#Include Label="SuzukiGroup">
 <#Include Label="ReeGroup">
 
+<Subsection Label="Generator Names">
+<Heading>Generator Names</Heading>
+For groups created as finitely presented groups, including polycyclic groups, the
+generators are labelled, by default, with a letter and a number. Using the
+option <C>generatorNames</C> it is possible to influcence this naming. If
+this option holds a string, the generators are named with this string and
+numbers, if a list of strings of sufficient length is given, the generator
+names are taken from this list of strings.
+<P/>
+<Example><![CDATA[
+gap> GeneratorsOfGroup(AbelianGroup([5,7]));
+[ f1, f2 ]
+gap> GeneratorsOfGroup(AbelianGroup([5,7]:generatorNames:="a"));
+[ a1, a2 ]
+gap> GeneratorsOfGroup(AbelianGroup([5,7]:generatorNames:=["u","v","w"]));
+[ u, v ]
+gap> AsSSortedList(DihedralGroup(12:generatorNames:="a"));
+[ <identity> of ..., a1, a2, a3, a1*a2, a1*a3, a2*a3, a3^2, a1*a2*a3,
+  a1*a3^2, a2*a3^2, a1*a2*a3^2 ]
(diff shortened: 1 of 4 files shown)


doc/ref/grplib.xml
```diff
@@ -99,6 +99,31 @@ calculations.
 <#Include Label="SuzukiGroup">
 <#Include Label="ReeGroup">
 
+<Subsection Label="Generator Names">
+<Heading>Generator Names</Heading>
+For groups created as finitely presented groups of polycyclic groups, the
```
fingolfin, line 104: Not sure I understand... perhaps this was meant?
```suggestion
For groups created as finitely presented groups or polycyclic groups, the
```

Optional remark: This might then be slightly confusing to *some* people: mathematically, of course polycyclic groups are also finitely presented groups. Perhaps `FpGroups` vs `PcGroups` is meant here? Of course that's a bit technical. Perhaps like this:
```suggestion
For groups created as finitely presented groups, including polycyclic groups, the
```

doc/ref/grplib.xml
```diff
@@ -99,6 +99,31 @@ calculations.
...
+generators are labelled, by default, with a letter and a number. Using the
+option <C>generatorNames</C> it is possible to influcence this naming. If
+this option holds a string, the generators are named with this sting and
+numbers, if a list of strings of sufficient length is given, the generator
```
fingolfin, line 108:
```suggestion
this option holds a string, the generators are named with this string and
numbers, if a list of strings of sufficient length is given, the generator
```
Optional:
```suggestion
this option holds a string, the generators names are formed by concatenating
 this string with the numbers 1, 2, 3, ...; if instead a list of strings
 of sufficient length is given, the generator
```

doc/tut/domain.xml
```diff
@@ -311,7 +312,8 @@ true
 ]]></Example>
 <P/>
 Many functions return subdomains of their arguments, for example
-the result of <C>SylowSubgroup( <A>G</A> )</C> is a group with parent group
+the result of <C>SylowSubgroup( <A>G</A>, <A>prime</A> )</C>
+is a group with parent group
 <A>G</A>.
```
fingolfin, line 317: Optional (but IMHO makes the XML more readable)
```suggestion
the result of <C>SylowSubgroup( <A>G</A>, <A>prime</A> )</C>
is a group with parent group <A>G</A>.
```

doc/tut/group.xml
```diff
@@ -340,7 +341,8 @@ for example would the default domain of <C>Group(  (2,3,4)  )</C> be
 To avoid confusion, all action  functions  require that you
 specify  the domain of action.
 If we had specified <C>[ 1 .. 113 ]</C> in the
-primitivity test above,  point&nbsp;113  would have been  a fixpoint  (and the
+primitivity test above,  point&nbsp;113  would have been  a fixed point
+(and the
 action would not even have been transitive).
```
fingolfin, line 346: Optional
```suggestion
primitivity test above,  point&nbsp;113  would have been  a fixed point
(and the action would not even have been transitive).
```

fingolfin (commented):
Some typos, otherwise looks good to me.

PR #4118, opened 2020-09-29
kernel: move BottomLVars out of TLS

fingolfin:
A single global BottomLVars shared by all TLS suffices. Note that really only
the address stored in BottomLVars matters, the actual object behind it is
essentially never accessed.


rbehrends (approved):
Nice find. :)

PR #4122, opened 2020-10-02
kernel: remove input stream stack

fingolfin:
Instead of pre-allocating a fixed number of TypInputFile instances on each
thread, we allocate the required storage dynamically on the stack. This
arguably makes it easier to reason about the global state of GAP.

It also enables future simplifications and improvements, e.g. further
decoupling the input file code from the line buffering and potential
optimizations resulting from that.

Next step: minimize or ideally eliminate usage of `IO()->Input` resp. `GetCurrentInput`.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -205,7 +205,7 @@ typedef const struct init_info StructInitInfo;
 *T  TypInputFile  . . . . . . . . . .  structure of an open input file, local
 **
 **  This is a forward declaration so that TypInputFile can be used in header
-**  files. The actual declaration is in io.c and is private.
+**  files. The actual declaration is in io.h.
 */
 typedef struct TypInputFile TypInputFile;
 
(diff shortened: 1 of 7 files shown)


ChrisJefferson:
Maybe mention in the commit message we are removing the (apparently unused) ToggleEcho? I guess at some point someone wanted it -- no need to mention it in the release notes.

ChrisJefferson (approved):

fingolfin:
Aha, good point, I forgot about the removal of `ToggleEcho` -- actually, I could just not remove it, too (in an earlier version of this PR I was still trying to get rid of the complete global `IO()->Input` pointer, but I gave up on that when I realized that GAP code actively makes use of that). Well, and `ToggleEcho` was not used by anything as far as I could tell...

fingolfin:
I've added `ToggleEcho` back. Seems better to avoid potential conflicts of interest here; it could still be removed in another PR.

PR #4124, opened 2020-10-04
set AlwaysPrintTracebackOnError to false inside Test

### LINKED ISSUE #4121: AlwaysPrintTracebackOnError
The (undocumented) global variable `AlwaysPrintTracebackOnError` seems to be intended for situations where GAP is integrated with other systems.
In the GAP/Julia integration, I have set the value to `true` in order to get better error messages, see oscar-system/GAP.jl/pull/520.

One problem is that `Test` apparently expects that `AlwaysPrintTracebackOnError` has the value `false`;
If errors occur in a GAP testfile then differences are reported because the output looks differently.
As an example, try the following GAP input.
```
gap> AlwaysPrintTracebackOnError:= true;;
gap> Test( Filename( DirectoriesLibrary( "tst" ), "testinstall/zmodnz.tst" ) );
```
One way to fix this problem would be to set `AlwaysPrintTracebackOnError` to `false` during tests.
Another way would be to change the comparison of error messages inside `Test`.
What do you think?

ThomasBreuer:
This fixes #4121.

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -164,11 +164,14 @@ InstallGlobalFunction(ParseTestFile, function(arg)
 end);
 
 InstallGlobalFunction(RunTests, function(arg)
-  local tests, opts, breakOnError, inp, outp, pos, cmp, times, ttime, nrlines,
-        s, res, fres, t, f, i, gaproot, localbag;
+  local tests, opts, breakOnError, alwaysPrintTracebackOnError, inp, outp,
+        pos, cmp, times, ttime, nrlines, s, res, fres, t, f, i, gaproot,
+        localbag;
   # don't enter break loop in case of error during test
   tests := arg[1];
-  opts := rec( breakOnError := false, showProgress := "some", localdef := false );
+  opts := rec( breakOnError := false, alwaysPrintTracebackOnError:= false,
+               showProgress := "some", localdef := false );
+
   if not IS_OUTPUT_TTY() then
     opts.showProgress := false;
   fi;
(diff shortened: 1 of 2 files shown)


fingolfin:
We would like this backported for 4.11, so we can benefit from it in the Julia integration.

fingolfin:
This PR breaks a few tests in `tst/testspecial` for the trivial reason that it changes a few line numbers and those tests currently are very simple and need to be adjusted in that case. That can usually be done by changing into that directory and running `regenerate_tests.sh`.

fingolfin (approved):

fingolfin:
I'll leave this open a little bit longer in case @alex-konovalov or @ChrisJefferson would like to comment

ChrisJefferson (approved):

fingolfin:
@ThomasBreuer the reason these files are kept in git and not (re)generated when we build GAP is that that would defeat the purpose: we want to be able to catch if certain outputs of GAP change, and that requires a stable reference. However, we could (and should) certainly try to avoid those pesky line numbers, e.g. by replacing them by a generic placeholder; the main issue is finding all instances (but perhaps just finding *some* instances would help; here clearly finding and replacing `GAPROOT/lib/test.gi:572` by say `GAPROOT/lib/test.gi:LINE` would be easy enough.

fingolfin:
Backported to stable-4.11 in b417302adfcdf24b982da729d960b736cbba7581 and e658aa032e7af39980927a875326b74b2a4b32b9

PR #4125, opened 2020-10-06
Misc kernel tweaks: ease debugging of module state sizes; fix bug in OpenAppend; use TRUE/FALSE some more

fingolfin:
I don't have a testcase to actually trigger undefined behavior in OpenAppend, but I suspect it exists.

Should this be in the release notes? I am tempted to say "no" as probably nobody ever hit this, but... *shrug*


wilfwilson (approved):
Let's not bother with adding it to the release notes.

PR #4126, opened 2020-10-07
Add 'gap --version'

### LINKED ISSUE #3735: Add `--version` command line option
Add `--version` command line option which prints out the GAP version and exits. This command should be fast, i.e., be handled by the kernel before the library is loaded.

This would be helpful for scripts that just want to test if the installed GAP version is new enough.

fingolfin:
Resolves #3735

I guess we could try to add a test, but I am not quite sure what to test for... Perhaps a non-zero exit code, and then match the output against a regex?

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -54,6 +54,7 @@ BIND_GLOBAL( "GAPInfo", rec(
     # for those options is correct
     CommandLineOptionData := [
       rec( short:= "h", long := "help", default := false, help := ["print this help and exit"] ),
+      rec( long := "version", default := false, help := ["print the GAP version and exit"] ),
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):
How would you even go about adding a test? Would it have to go in `etc/ci.sh` somehow? Anyway, if you feel like doing it, it would be nice to have, but I'm satisfied without a test.

fingolfin:
Yeah, in `etc/ci.sh` -- indeed, not very much to test, though one could test that the exit code is as desired, and that the output consist of a single line matching a certain pattern (as opposed to an error message or so). But I am not sure how useful that'd really be... so unless somebody reaaaallly wants it, I'd rather not invest the effort :-)

PR #4110, opened 2020-09-04
Add TextInputFileRaw and TextOutputFileRaw

ChrisJefferson:
Add the ability to access files without compression, for packagemanager (and other users who might need this).

Needs some more tests before merging, checking if the naming / general design is reasonable / fixes the problem

PR #4128, opened 2020-10-08
Add OutputGzipFile ( was TextInputFileRaw and TextOutputFileRaw )

ChrisJefferson:
Continues #4110, as I stupidly put that branch in gap-system/gap, and shouldn't do that. All suggested fixes from that patch have been applied here.

### DIFF
--- a/lib/streams.gd
+++ b/lib/streams.gd
@@ -726,14 +726,15 @@ DeclareOperation( "OutputTextString", [ IsList, IsBool ] );
 ##  <#GAPDoc Label="OutputTextFile">
 ##  <ManSection>
 ##  <Oper Name="OutputTextFile" Arg='filename, append'/>
+##  <Oper Name="OutputGzipFile" Arg='filename, append'/>
 ##
 ##  <Description>
 ##  <C>OutputTextFile( <A>filename</A>, <A>append</A> )</C> returns an output stream in the
 ##  category <C>IsOutputTextFile</C> that writes received characters to the file
 ##  <A>filename</A>.  If <A>append</A> is <K>false</K>, then the file is emptied first,
 ##  otherwise received characters are added at the end of the file.
-##  If <A>filename</A> ends in <C>.gz</C> then the file will be
-##  written with gzip compression.
+##  <C>OutputGzipFile</C> acts identically to <C>OutputTextFile</C>, except it compresses
+##  the output with gzip.
 ##  <P/>
 ##  <Example><![CDATA[
 ##  gap> # use a temporary directory
@@ -766,6 +767,7 @@ DeclareOperation( "OutputTextString", [ IsList, IsBool ] );
 ##  <#/GAPDoc>
 ##
 DeclareOperation( "OutputTextFile", [ IsString, IsBool ] );
+DeclareOperation( "OutputGzipFile", [ IsString, IsBool ] );
 
 
 #############################################################################
--- a/lib/streams.gi
+++ b/lib/streams.gi
@@ -1084,14 +1084,14 @@ fi;
 #M  OutputTextFile( <str>, <append> )
 ##
 InstallMethod( OutputTextFile,
-    "output text stream from file",
+    "output text stream to file",
     [ IsString,
       IsBool ],
 function( str, append )
     local   fid;
     str := UserHomeExpand(str);
 
-    fid := OUTPUT_TEXT_FILE( str, append );
+    fid := OUTPUT_TEXT_FILE( str, append, false );
     if fid = fail  then
         return fail;
     else
@@ -1108,6 +1108,37 @@ InstallOtherMethod( OutputTextFile,
         -SUM_FLAGS, # as low as possible
         function( str )
     Error("Usage OutputTextFile( <fname>, <append> )");
+end );
+
+#############################################################################
+##
+#M  OutputGzipFile( <str>, <append> )
+##
+InstallMethod( OutputGzipFile,
+    "output gzipped text to file",
+    [ IsString,
+      IsBool ],
+function( str, append )
+    local   fid;
+    str := UserHomeExpand(str);
+
+    fid := OUTPUT_TEXT_FILE( str, append, true );
+    if fid = fail  then
+        return fail;
+    else
+        atomic OutputTextFileStillOpen do
+            AddSet( OutputTextFileStillOpen, fid );
+        od;
+        return Objectify( OutputTextFileType, [fid, Immutable(str), true] );
+    fi;
+end );
+
+InstallOtherMethod( OutputGzipFile,
+        "error catching method, append not given",
+        [ IsString ],
+        -SUM_FLAGS, # as low as possible
+        function( str )
+    Error("Usage OutputGzipFile( <fname>, <append> )");
 end );
 
 #############################################################################
--- a/src/io.c
+++ b/src/io.c
@@ -360,7 +360,7 @@ UInt OpenInput(TypInputFile * input, const Char * filename)
 #endif
 
     /* try to open the input file                                          */
-    file = SyFopen( filename, "r" );
+    file = SyFopen(filename, "r", TRUE);
     if ( file == -1 )
         return 0;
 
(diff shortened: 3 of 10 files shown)


wilfwilson:
Which PR are you referring to? #4410 doesn't seem to exist on `gap-system/gap` (which is here).

ChrisJefferson:
Sorry, meant #4110 (which I've corrected in the first message, in case anyone else comes along)

ChrisJefferson:
There was a failing test which is, I believe, unrelated to this PR but I have fixed here anyway -- ReadLine can get only a partial line when reading from a process which hasn't yet finished outputting a whole line.

src/sysfiles.c
```diff
@@ -585,7 +585,7 @@ void SyBufSetEOF(Int fid)
 
 /****************************************************************************
 **
-*F  SyFopen( <name>, <mode> ) . . . . . . . .  open the file with name <name>
+*F  SyFopen( <name>, <mode>, <transcompress> ) open the file with name <name>
```
wilfwilson, line 588: Please make the same change in the header file (and add the sentence about `transcompress`).

lib/streams.gi
```diff
@@ -1102,6 +1124,25 @@ function( str, append )
...
 
+InstallMethod( OutputTextFileRaw,
+    "output raw text stream from file",
+    [ IsString,
+      IsBool ],
+function( str, append )
+    local   fid;
+    str := UserHomeExpand(str);
+
+    fid := OUTPUT_TEXT_FILE_RAW( str, append );
```
fingolfin, line 1135: So, I don't want to object to the merging of this PR, but I still wonder this: There's a lot of code duplication in this PR. But I wonder, couldn't we just add an extra argument `BOOL useRawMode` or so to `INPUT_TEXT_FILE` and `OUTPUT_TEXT_FILE` ? And then add an optional second/third argument to `InputTextFile` and `OutputTextFile`?  (granted, the latter gets a bit messy; I wish GAP had keyword arguments; abusing options for that isn't so nice because of unintended consequences due to their recursive nature)

ChrisJefferson: for some reason I thought some packages were calling these methods directly, but now I don't know why I thought that as I can't find any packages which do. I'll merge.

fingolfin:
Since I've run into the PackageManager issue myself now (paging @mtorpey), I've thought about this PR. First off, for PackageManager, I suggest that it directly downloads into a file when possible, sidestepping the double compression issue (see also https://github.com/gap-packages/PackageManager/issues/59).

But overall, I am scared that the double compression  bug will hit more people; while the automatic gzip compression is a cool feature, it's also a trap inviting regressions and weird bugs. 

This PR attempts to mitigate it by at least adding a way to bypass the issue (so an opt-out solution). But this strikes me as the wrong approach: It still breaks existing code which needs to be adapted (and then the adaption means that old code won't work with old GAP versions). Shouldn't we rather return the default behavior to what it was (i.e., *no* automatic compression), and then add a way to *opt-in* to compression? E.g. via a `GzipOutputStream`, which then also would work with arbitrary extensions?

fingolfin:
At the very least, we should remove the automatic compression from the `stable-4.11` branch, so that if we ever make a 4.11.1 release it doesn't regress existing code; if we decide we want the regression, it should be in 4.12.0.

src/sysfiles.h
```diff
@@ -114,7 +114,7 @@ void SyBufSetEOF(Int fid);
 **  Right now GAP does not read nonascii files, but if this changes sometimes
 **  'SyFopen' must adjust the mode argument to open the file in binary mode.
 */
```
wilfwilson, line 116: Just to take it match up with what's now in `sysfiles.c`:
```suggestion
**
**  If <transcompress> is TRUE, files with names ending '.gz' will be
**  automatically compressed/decompressed using gzip.
*/
```

ChrisJefferson: thanks (done in another rearrange)

ChrisJefferson:
Now significantly changed based on feedback (although the internals are similar).

Remove all 'Raw' methods. Add 'OutputGzipFile' (to match OutputTextFile), which gzips anything. OutputTextFile now never compresses.

InputTextFile still always uncompresses (as it has for many years, although now it does it more situations). Even if this was disabled, you still couldn't read gzipped files with InputTextFile, because we have always (and we have no way to disable currently) rewriten line endings in windows, which would break binary files.

lib/streams.gd
```diff
@@ -726,14 +726,16 @@ DeclareOperation( "OutputTextString", [ IsList, IsBool ] );
...
 ##  <ManSection>
 ##  <Oper Name="OutputTextFile" Arg='filename, append'/>
+##  <Oper Name="OutputGzipFile" Arg='filename, append'/>
 ##
 ##  <Description>
 ##  <C>OutputTextFile( <A>filename</A>, <A>append</A> )</C> returns an output stream in the
 ##  category <C>IsOutputTextFile</C> that writes received characters to the file
 ##  <A>filename</A>.  If <A>append</A> is <K>false</K>, then the file is emptied first,
 ##  otherwise received characters are added at the end of the file.
 ##  If <A>filename</A> ends in <C>.gz</C> then the file will be
```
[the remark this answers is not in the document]
ChrisJefferson: Sorry, your message is too brief for me to guess the problem?

My aim is now OutputTextFile has it's 4.10 behaviour -- it never gzips, and you use OutputGzipFile explicitly if you want to output with gzip compression.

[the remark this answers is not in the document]
ChrisJefferson: I see the problem, yes, docs are still wrong.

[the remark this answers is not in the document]
ChrisJefferson: Also, I noticed I was still compressing things like `OpenLog`. Maybe we want that, but for now let's do the minimal.

The only thing I'm now compressing if it ends .gz without prompting is SaveWorkspace, because it's just a binary file most people wouldn't understand anyway, and it compresses really well.

src/sysfiles.c
```diff
@@ -596,21 +596,22 @@ void SyBufSetEOF(Int fid)
...
+**  'SyFopen( "*errin*", "r", ..)' returns 2, the brk loop input file.
+**  'SyFopen( "*errout*","w", ..)' returns 3, the error messages file.
 **
 **  If it is necessary  to adjust the filename  this should be done here, the
 **  filename convention used in GAP is that '/' is the directory separator.
 **
 **  Right now GAP does not read nonascii files, but if this changes sometimes
 **  'SyFopen' must adjust the mode argument to open the file in binary mode.
+**
+**  If <transcompress> is TRUE, files with names ending '.gz' will be
```
fingolfin, line 610: transcompress = "transparent_compress"?

lib/streams.gi
```diff
@@ -1084,14 +1084,33 @@ fi;
...
+        return fail;
+    else
+        atomic OutputTextFileStillOpen do
+            AddSet( OutputTextFileStillOpen, fid );
+        od;
+        return Objectify( OutputTextFileType, [fid, Immutable(str), true] );
+    fi;
+end );
+
+InstallMethod( OutputGzipFile,
```
fingolfin, line 1105: Could you perhaps move this below the second `OutputTextFile` and add a comment like we use for other stuff? ```suggestion
##
 #M  OutputGzipFile( <str>, <append> )
 ##
InstallMethod( OutputGzipFile,
```
This would preserve grouping, and also the diff here is easier to read. I am trying to understand why the CI tests fail and it is difficult to grok what happens.

Alternatively, I'd be happy to make that change, I just don't want to make it "behind your back"

File: src/io.c [hunk 7]
```diff
@@ -360,7 +360,7 @@ UInt OpenInput(TypInputFile * input, const Char * filename)
 #endif
 
     /* try to open the input file                                          */
-    file = SyFopen( filename, "r" );
+    file = SyFopen(filename, "r", FALSE);
```
fingolfin, line 363: Shouldn't transparent decompression upon reading always be enabled? Also elsewhere... that might explain the test failures?
```suggestion
    file = SyFopen(filename, "r", TRUE);
```

ChrisJefferson: Yep, cleaning up

ChrisJefferson:
Sorry, just cleaning up and pushing some changes. I am having one consistent problem, which is this patch seems to stop HPCGAP from even getting to the gap prompt, which is strange and I'm investigating.

ChrisJefferson:
EDIT: Found the issue, was simple (needed to fix HPC-GAP's console code to use new functions)

fingolfin (approved):
LGTM!
The testmanuals check failed to download the bootstrap archive; I've restarted the checks to make sure it passes then (unfortunately one can only restart *all* checks at once...)

PR #4129, opened 2020-10-09
kernel: merge OpenAppend into OpenOutput

fingolfin:
This reduces code duplication, and in fact aligns the behavior of these two functions which really should behave virtually the same.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -420,12 +420,12 @@ static UInt OpenDefaultOutput(void)
     return OpenOutputStream(stream);
   func = GVarOptFunction(&DEFAULT_OUTPUT_STREAM);
   if (!func)
-    return OpenOutput("*stdout*");
+    return OpenOutput("*stdout*", FALSE);
   stream = CALL_0ARGS(func);
   if (!stream)
     ErrorQuit("DEFAULT_OUTPUT_STREAM() did not return a stream", 0, 0);
   if (IsStringConv(stream))
-    return OpenOutput(CONST_CSTR_STRING(stream));
+    return OpenOutput(CONST_CSTR_STRING(stream), FALSE);
   TLS(DefaultOutput) = stream;
   return OpenOutputStream(stream);
 }
@@ -919,15 +919,19 @@ UInt CloseOutputLog ( void )
 **  'OpenOutput' passes  those  file names to 'SyFopen'  like any other name,
 **  they are just a convention between the main and the system package.
 **
-**  It is not neccessary to open the initial output file, 'InitScanner' opens
-**  '*stdout*' for that purpose.  This  file  on the other hand   can not  be
-**  closed by 'CloseOutput'.
+**  The function does nothing and returns success for '*stdout*' and
+**  '*errout*' when 'LockCurrentOutput(1)' is in effect (used for testing
+**  purposes).
+**
+**  It is not neccessary to open the initial output file; '*stdout'* is
+**  opened for that purpose during startup. This file on the other hand  can
+**  not be closed by 'CloseOutput'.
+**
+**  If <append> is set to true, then 'OpenOutput' does not truncate the file
+**  to size 0 if it exists.
 */
-UInt OpenOutput (
-    const Char *        filename )
+UInt OpenOutput(const Char * filename, BOOL append)
 {
-    Int                 file;
-
     // do nothing for stdout and errout if caught
     if (IO()->Output != NULL && IO()->IgnoreStdoutErrout == IO()->Output &&
         (streq(filename, "*errout*") || streq(filename, "*stdout*"))) {
@@ -947,7 +951,7 @@ UInt OpenOutput (
 #endif
 
     /* try to open the file                                                */
-    file = SyFopen( filename, "w" );
+    Int file = SyFopen(filename, append ? "a" : "w");
     if ( file == -1 )
         return 0;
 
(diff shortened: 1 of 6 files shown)


src/io.c
```diff
@@ -919,15 +919,19 @@ UInt CloseOutputLog ( void )
 **  'OpenOutput' passes  those  file names to 'SyFopen'  like any other name,
 **  they are just a convention between the main and the system package.
 **
-**  It is not neccessary to open the initial output file, 'InitScanner' opens
-**  '*stdout*' for that purpose.  This  file  on the other hand   can not  be
-**  closed by 'CloseOutput'.
+**  The function does nothing and returns success for '*stdout*' and
+**  '*errout*' when 'LockCurrentOutput(1)' is in effect (used for testing
+**  purposes).
```
fingolfin, line 924: Note: the above paragraph was copied from the corresponding comment in `io.h`, i.e., what I did here was to sync the comment in io.c and io.h

fingolfin: Actually, I just checked: even when `InitScanner` existed, it didn't do what was claimed here since about 1997 :-)

src/io.c
```diff
@@ -1052,53 +1056,6 @@ UInt CloseOutput ( void )
...
-{
-    Int                 file;
-
-    /* fail if we can not handle another open output file                  */
-    if (IO()->OutputStackPointer == MAX_OPEN_FILES)
-        return 0;
-
-#ifdef HPCGAP
-    if (streq(filename, "*defout*"))
-        return OpenDefaultOutput();
```
fingolfin, line 1077: Just for those checking this PR very carefully, you might wonder that this code differs from the corresponding code in `OpenOutput` and whether that is correct: well, I think the code here in `OpenAppend` actually was not sufficient and the code in `OpenOutput` is correct, so this is kind of a bugfix (for a hypothetical bug); it just never mattered in practice because nobody ever tries to `OpenAppend` errout (it would/should have done the same as `OpenOutput` in plain GAP anyway; only in HPC-GAP would there have been a potential issue).

Similarly, one may wonder about the extra checking involving `IgnoreStdoutErrout` in `OpenOutput` that wasn't present in `OpenAppend`: again, adding this check in the append case is strictly speaking a bug fix, which again wasn't triggered because nobody though of writing code strange enough to do so. 

In summary, I really think that `OpenAppend` should be identical to `OpenOutput` except for the append vs. truncate mode question. This is confirmed by looking at the initial commit in the GAP 4 CVS repository, where the two really were almost identical (it's actually fascinating to me to contemplate how this stuff changed over time):
```C
**
*F  OpenOutput( <filename> )  . . . . . . . . . open a file as current output
**
**  'OpenOutput' opens the file  with the name  <filename> as current output.
**  All subsequent output will go  to that file, until either   it is  closed
**  again  with 'CloseOutput' or  another  file is  opened with 'OpenOutput'.
**  The file is truncated to size 0 if it existed, otherwise it  is  created.
**  'OpenOutput' does not  close  the  current file, i.e., if  <filename>  is
**  closed again, output will go again to the current output file.
**
**  'OpenOutput'  returns  1 if it  could  successfully  open  <filename> for
**  writing and 0 to indicate failure.  'OpenOutput' will fail if  you do not
**  have  permissions to create the  file or write   to it.  'OpenOutput' may
**  also   fail if you   have  too many files   open  at once.   It is system
**  dependent how many are too many, but 16 files should work everywhere.
**
**  You can open '*stdout*'  to write  to the standard output  file, which is
**  usually the terminal, or '*errout*' to write  to the standard error file,
**  which is the terminal  even   if '*stdout*'  is  redirected to   a  file.
**  'OpenOutput' passes  those  file names to 'SyFopen'  like any other name,
**  they are just a convention between the main and the system package.
**
**  It is not neccessary to open the initial output file, 'InitScanner' opens
**  '*stdout*' for that purpose.  This  file  on the other hand   can not  be
**  closed by 'CloseOutput'.
*/
UInt            OpenOutput (
    Char *              filename )
{
    Int                 file;

    /* fail if we can not handle another open output file                  */
    if ( Output+1==OutputFiles+(sizeof(OutputFiles)/sizeof(OutputFiles[0])) )
        return 0;

    /* in test mode keep printing to test output file for breakloop output */
    if ( TestInput != -1 && ! SyStrcmp( filename, "*errout*" ) )
        return 1;

    /* try to open the file                                                */
    file = SyFopen( filename, "w" );
    if ( file == -1 )
        return 0;

    /* put the file on the stack, start at position 0 on an empty line     */
    Output++;
    Output->file    = file;
    Output->line[0] = '\0';
    Output->pos     = 0;
    Output->indent  = 0;

    /* variables related to line splitting, very bad place to split        */
    Output->spos    = 0;
    Output->sindent = 666;

    /* indicate success                                                    */
    return 1;
}


**
*F  CloseOutput() . . . . . . . . . . . . . . . . . close current output file
**
**  'CloseOutput' will  first flush all   pending output and  then  close the
**  current  output  file.   Subsequent output will  again go to the previous
**  output file.  'CloseOutput' returns 1 to indicate success.
**
**  'CloseOutput' will  not  close the  initial output file   '*stdout*', and
**  returns 0 if such attempt is made.  This  is  used in 'Error' which calls
**  'CloseOutput' until it returns 0, thereby closing all open output files.
**
**  Calling 'CloseOutput' if the corresponding 'OpenOutput' call failed  will
**  close the current output file, which will lead to very strange behaviour.
**  On the other  hand if you  forget  to call  'CloseOutput' at the end of a
**  'PrintTo' call or an error will not yield much better results.
*/
UInt            CloseOutput ( void )
{
    /* refuse to close the initial output file '*stdout*'                  */
    if ( Output == OutputFiles )
        return 0;

    /* refuse to close the test output file                                */
    if ( Output->file == TestOutput )
        return 0;

    /* flush output and close the file                                     */
    Pr( "%c", (Int)'\03', 0L );
    SyFclose( Output->file );

    /* revert to previous output file and indicate success                 */
    Output--;
    return 1;
}


**
*F  OpenAppend( <filename> )  . . open a file as current output for appending
**
**  'OpenAppend' opens the file  with the name  <filename> as current output.
**  All subsequent output will go  to that file, until either   it is  closed
**  again  with 'CloseAppend' or  another  file is  opened with 'OpenOutput'.
**  Unlike 'OpenOutput' 'OpenAppend' does not truncate the file to size 0  if
**  it exists.  Appart from that 'OpenAppend' is equal to 'OpenOutput' so its
**  description applies to 'OpenAppend' too.
*/
UInt            OpenAppend (
    Char *              filename )
{
    Int                 file;

    /* fail if we can not handle another open output file                  */
    if ( Output+1==OutputFiles+(sizeof(OutputFiles)/sizeof(OutputFiles[0])) )
        return 0;

    /* in test mode keep printing to test output file for breakloop output */
    if ( TestInput != -1 && ! SyStrcmp( filename, "*errout*" ) )
        return 1;

    /* try to open the file                                                */
    file = SyFopen( filename, "a" );
    if ( file == -1 )
        return 0;

    /* put the file on the stack, start at position 0 on an empty line     */
    Output++;
    Output->file    = file;
    Output->line[0] = '\0';
    Output->pos     = 0;
    Output->indent  = 0;

    /* variables related to line splitting, very bad place to split        */
    Output->spos    = 0;
    Output->sindent = 666;

    /* indicate success                                                    */
    return 1;
}


**
*F  CloseAppend() . . . . . . . . . . . . . . . . . close current output file
**
**  'CloseAppend' will  first flush all   pending output and  then  close the
**  current  output  file.   Subsequent output will  again go to the previous
**  output file.  'CloseAppend' returns 1 to indicate success.  'CloseAppend'
**  is exactely equal to 'CloseOutput' so its description applies.
*/
UInt            CloseAppend ( void )
{
    /* refuse to close the initial output file '*stdout*'                  */
    if ( Output == OutputFiles )
        return 0;

    /* refuse to close the test output file                                */
    if ( Output->file == TestOutput )
        return 0;

    /* flush output and close the file                                     */
    Pr( "%c", (Int)'\03', 0L );
    SyFclose( Output->file );

    /* revert to previous output file and indicate success                 */
    Output--;
    return 1;
}
```

fingolfin: The fact that these functions diverged so much over time (the differences actually used to be even bigger, we just already removed some of them) nicely shows why code duplication can be problematic

ChrisJefferson (approved):

PR #4132, opened 2020-10-10
kernel: remove output stream stack

### LINKED ISSUE #1308: `SetPrintFormattingStatus( "*stdout*", false );` does not quite work
This happens with master and originally comes from here: http://stackoverflow.com/questions/43833333

If one puts
```gap
SetPrintFormattingStatus( "*stdout*", false );
Print( Primes{[1..30]}, "\n" );
```
into a file `test.g` and then runs it via `gap test.g`, then output formatting is not actually ignored, and so line breaks are inserted.

fingolfin:
Instead of pre-allocating a fixed number of TypOutputFile instances on each thread, we allocate the required storage dynamically on the stack. This arguably makes it easier to reason about the global state of GAP.

It also enables future simplifications and improvements, e.g. to fix issues were switching between stdout and errout breaks tracking of the output state, which leads to GAP expecting resp. inserting linebreaks in the wrong spots.

<s>This includes PR #4122 which should be reviewed and merged first.</s> DONE

Fixes #1308

Text for release notes:
> Fix a long standing bug which prevented `SetPrintFormattingStatus( "*stdout*", false );` from working as expected

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -44,34 +44,6 @@
 #include <limits.h>
 
 
-/****************************************************************************
-**
-*T  TypOutputFiles  . . . . . . . . . structure of an open output file, local
-**
-**  'TypOutputFile' describes the information stored for open  output  files:
-**  'file' holds the file identifier which is  received  from  'SyFopen'  and
-**  which is passed to  'SyFputs'  and  'SyFclose'  to  identify  this  file.
-**  'line' is a buffer that holds the current output line.
-**  'pos' is the position of the current character on that line.
-*/
-/* the maximal number of used line break hints */
-#define MAXHINTS 100
-typedef struct {
-    BOOL isstream;
-    BOOL isstringstream;
-    Obj  stream;
-    Int  file;
-
-    char line[MAXLENOUTPUTLINE];
-    Int  pos;
-    BOOL format;
-    Int  indent;
-
-    /* each hint is a tripel (position, value, indent) */
-    Int hints[3 * MAXHINTS + 1];
-} TypOutputFile;
-
-
 static Char GetLine(TypInputFile * input);
 static void PutLine2(TypOutputFile * output, const Char * line, UInt len);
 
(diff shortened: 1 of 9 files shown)


src/io.c
```diff
@@ -1915,7 +1871,12 @@ static Obj FuncINPUT_LINENUMBER(Obj self)
 
 static Obj FuncSET_PRINT_FORMATTING_STDOUT(Obj self, Obj val)
 {
-    IO()->OutputStack[1]->format = (val != False);
+    TypOutputFile * output = IO()->Output;
+    if (!output)
+        ErrorMayQuit("SET_PRINT_FORMATTING_STDOUT called while no output is opened\n", 0, 0);
+    while (output->prev)
+        output = output->prev;
+    output->format = (val != False);
```
fingolfin, line 1879: As an accidental byproduct, this change also fixes #1308

src/io.h
```diff
@@ -78,6 +78,47 @@ struct TypInputFile {
...
+    BOOL isstringstream;
+    Obj  stream;
+    Int  file;
+
+    char line[MAXLENOUTPUTLINE];
+    Int  pos;
+    BOOL format;
+    Int  indent;
+
+    /* each hint is a tripel (position, value, indent) */
```
wilfwilson, line 117: I know it's not your own typo, but we may as well fix it:
```suggestion
    /* each hint is a triple (position, value, indent) */
```

wilfwilson (approved):
Looks fine except perhaps addressing that pre-existing typo.

src/io.h
fingolfin: Done, plus a few more (except what you already fixed in #4139)

wilfwilson: Nice! The typos that you fixed also exist in `io.c`, I think, but I'm totally happy for them to be fixed some other time.

fingolfin: Also fixed in `io.c`. I hate it when those comments drift apart -- should probably get rid of more of the duplicates. Thanks for catching this one, too

wilfwilson (approved):

PR #4137, opened 2020-10-14
Fix bug where ? or a pragma comment acted as a semicolon

### LINKED ISSUE #2017: Regression: parser ignores anything after `?`
This regression was introduced by PR #1886, which introduced `S_HELP` to the parser. 

First off, if `?` is not the first non-space thing on an input line, it eats up all input after it (so almost like `#`, except the semicolon `;` is still handled). Example:
```
gap> 1?hello world, this is so much fun!!!;
1
gap> 1??;
1
```
For reference, this is what I get in GAP 4.8.8:
```
gap> 1?hello world, this is so much fun!!!;
Syntax error: ; expected
1?hello world, this is so much fun!!!;
 ^
Error, Variable: 'hello' must have a value
not in any function at line 6 of *stdin*
gap> 1??;
Syntax error: ; expected
1??;
 ^
```

I guess we'll want to raise an error in `GetHelp()` if the `?` did not occur at the start of the line. So we could just check `GetLinePosition() == 0`. Of course that would mean it is not anymore possible to add spaces before a help query; but then, this is also how it was in GAP 4.8, so it would be an acceptable loss, IMHO. That said, of course it would be nicer to allow the user to add those spaces in front of the help query.

fingolfin:
Fixes #2017

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2627,8 +2627,14 @@ ExecStatus ReadEvalCommand(Obj            context,
     // the top level than inside a function, if/else or loop
     case S_QUIT:      ReadQuit(rs,    S_SEMICOLON|S_EOF      ); break;
     case S_QQUIT:     ReadQUIT(rs,    S_SEMICOLON|S_EOF      ); break;
-    case S_HELP:      ReadHelp(rs,    S_SEMICOLON|S_EOF      ); break;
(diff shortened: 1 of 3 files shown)


wilfwilson (approved):
Seems like a suitable solution 🙂

fingolfin:
Yeah, well, in hindsight it seems almost obvious... but it took me three years to figure it out 😂

PR #4140, opened 2020-10-15
Fix compatibility with some old glibc versions

### LINKED ISSUE #3918: Checking for presence of O_DIRECTORY and O_CLOEXEC.
Trying to build GAP on [ARCHER](https://www.archer.ac.uk/) with @raemarina @IrynaRaievska and @marioa, and observing this:

```
src/iostream.c: In function 'StartChildProcess':
src/iostream.c:396: error: 'O_DIRECTORY' undeclared (first use in this function)
src/iostream.c:396: error: (Each undeclared identifier is reported only once
src/iostream.c:396: error: for each function it appears in.)
src/iostream.c:396: error: 'O_CLOEXEC' undeclared (first use in this function)
make: *** [obj/src/iostream.lo] Error 1
```

As @fingolfin said, that means that default glibc is more than 10 years old and 
>  that's a regression we should fix in GAP: we detect the presence of `posix_spawn` and want to use it, but don't check for presence of `O_DIRECTORY` and `O_CLOEXEC`. .. In the meantime, a workaround might be to remove the test for `posix_spawn` from `configure.ac`.

fingolfin:
On some old glibc versions, `posix_spawn` is present but not `O_DIRECTORY` nor
`O_CLOEXEC`, leading to a compilation error.

Also avoid another compilation error that happens if `popen` is not available.

Fixes #3918.

@alex-konovalov @raemarina  can you confirm that this helps?

### DIFF
--- a/src/iostream.c
+++ b/src/iostream.c
@@ -316,8 +316,22 @@ static Obj FuncDEFAULT_SIGCHLD_HANDLER(Obj self)
 #undef HAVE_POSIX_SPAWN
 #endif
 
+// if neither posix_spawn_file_actions_addchdir nor O_CLOEXEC are available,
+// our posix_spawn code should not be used. This affects e.g. Linux systems
+// with an old glibc, see https://github.com/gap-system/gap/issues/3918.
+#if !defined(HAVE_POSIX_SPAWN_FILE_ACTIONS_ADDCHDIR) && !defined(O_CLOEXEC)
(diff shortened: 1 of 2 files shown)


ChrisJefferson (approved):

alex-konovalov:
Thanks @fingolfin! For ARCHER, @raemarina can hopefully check, if she continues to use GAP there.

PR #4117, opened 2020-09-28
Fix --disable-X for some options in configure

ChrisJefferson:
Fixes #4116 -- some options I added to configure were enabled both for --enable-X and --disable-X, as I didn't understand how autoconf works.

PR #4141, opened 2020-10-15
buildsys: fix --enable-debug regression

fingolfin:
A missing comma introduced in PR #4117 caused enable-debug to default to being active.

Thanks to @wilfwilson for catching the resulting performance regression on Travis CI; without him we might not have spotted this for weeks or months to come!


wilfwilson (approved):
Ah ha! This also explains why I noticed `KernelDebug` in my GAP banner this evening (which I promptly ignored).

ChrisJefferson:
woops, thanks!

PR #4146, opened 2020-10-19
Fix encoding of BibEntry(nam) and Cite

frankluebeck:
`BibEntry` produces a snippet of XML without processing instruction
about its encoding. Therefore the result should be in the default
encoding "UTF-8" for our XML parser.

Instead `Cite` should use the setting in GAPInfo.TermEncoding for
its printing.

Before this fix `Cite` ran into an error in a terminal with, say, latin1 encoding.
The fix was tested in terminals with utf8 and with latin1 encoding. I don't see how to add a test for this in the `tst` directory.

# Description
see above

## Text for release notes 
not needed

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -2638,6 +2638,8 @@ InstallGlobalFunction( BibEntry, function( arg )
       fi;
     fi;
 
+    # Make sure that output of BibEntry is in UTF-8 encoding (also
+    # if PackageInfo.g is in latin1 encoding)
     ps:= function( str )
       local uni;
 
@@ -2645,7 +2647,7 @@ InstallGlobalFunction( BibEntry, function( arg )
       if uni = fail then
         uni:= Unicode( str, "ISO-8859-1" );
       fi;
-      return Encode( uni, GAPInfo.TermEncoding );
+      return Encode( uni, "UTF-8" );
     end;
 
     # According to <Cite Key="La85"/>,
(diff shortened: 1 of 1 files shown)


lib/package.gi
```diff
@@ -2778,83 +2780,96 @@ InstallGlobalFunction( BibEntry, function( arg )
...
+    name:="GAP";
+  else
+    name := NormalizedWhitespace(arg[1]);
+  fi;
+  if LowercaseString(name) = "gap" then 
+    name:="GAP"; 
+  else
+    pkginfo := InstalledPackageVersion(name);
+    pkginfo := First(PackageInfo(name), r-> r.Version = pkginfo);
+    name := pkginfo.PackageName;
```
fingolfin, line 2799: This code is also new and not mentioned in the commit message resp. PR description as far as I can tell? My guess is that it's there to "normalize" the spelling (in terms of case) of the package name?

lib/package.gi
```diff
@@ -2778,83 +2780,96 @@ InstallGlobalFunction( BibEntry, function( arg )
...
+    Print("If you are not using BibTeX, here is the bibliography entry produced \n",
+          "by BibTeX (in bibliography style `alpha'):\n\n",
+          "\\bibitem[GAP]{GAP4}\n", 
+          "\\emph{GAP -- Groups, Algorithms, and Programming}, ",
+          "Version ", GAPInfo.Version, ",\n", 
+          "The GAP~Group (", year, "), \\verb+https://www.gap-system.org+.\n\n");
+    Print(
+    "If you have (predominantly) used one or more particular GAP packages,\n", 
+    "please cite these packages in addition to GAP itself (either check the\n", 
+    "the package documentation for the suggestions, or use a scheme like:\n\n",
```
ThomasBreuer, line 2856: Perhaps remove one of the two `the` after `check`.

ThomasBreuer (approved):
Very good.
(I have inserted one comment concerning an old typo which could be fixed now.)

frankluebeck:
I have added a comment, fixed the typo as suggested, and added a remark to the commit message.

fingolfin (approved):
Thanks!

wilfwilson (approved):

PR #4149, opened 2020-10-21
Better syntax error for an expr when expecting a statement

### LINKED ISSUE #822: Misleading error message when missing `()` from function call
### Observed behaviour

```
gap> f := function() Stabilizer; end;
Syntax error: := expected
f := function() Stabilizer; end;
```
### Expected behaviour

Perhaps `:= or () expected`? We could also expect `.`, `[]`, but I think the function call or brackets are what we want.

fingolfin:
Resolves #822

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -896,7 +896,7 @@ static void ReadCallVarAss(ReaderState * rs, TypSymbolSet follow, Char mode)
             }
         }
         else {
-            Match_(rs, S_ASSIGN, ":=", follow);
+            Match_(rs, S_ASSIGN, "found an expression when a statement was", follow);
             UInt currLHSGVar = rs->CurrLHSGVar;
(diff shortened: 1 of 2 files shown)


fingolfin:
I am a bit concerned about the precise phrasing and would appreciate if a native speaker or two could take a look at this.

ChrisJefferson (approved):
I would say "found an expression..."(or more technically, you could say 'parsed an expression...') instead of "got an expression..."

stevelinton (approved):

wilfwilson (approved):

ruthhoffmann (approved):
Not a (fully) native speaker but agree with @ChrisJefferson

fingolfin:
I've changed this to `found...` right after @ChrisJefferson pointed it out (it immediately "felt" better to me as well). So I assume everybody is happy with this now :-).

PR #4151, opened 2020-10-22
Fix the user preference SaveAndRestoreHistory

ssiccha:
Fixes the user preference `SaveAndRestoreHistory` by extending `InstallAtExit` to pass arguments and by fixing the behaviour of `SaveCommandLineHistory`.

The latter promises to either append to or overwrite the line history file with the line history of the current session. Instead it took both the old line history and the line history of the current session, and wrote into or appended it to the line history file.

GAP by default used the overwriting version of `SaveCommandLineHistory`. This led to incorrect history files if two or more gap sessions were started simultaneously. After closing all sessions only the line history of the most recently closed session ended up in the line history file.

### DIFF
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -700,28 +700,38 @@ BindKeysToGAPHandler("\022");
 BindGlobal("SaveCommandLineHistory", function(arg)
   local  fnam, append, hist, out;
 
-  if Length(arg) > 0 then
+  if Length(arg) = 1 then
+    if IsString(arg[1]) then
+      fnam := arg[1];
+      append := false;
+    elif arg[1] in [false, true] then
+      append := arg[1];
+    fi;
+  elif Length(arg) = 2 then
     fnam := arg[1];
-  else
+    append := arg[2];
+  elif Length(arg) > 2 then
+    ErrorNoReturn("Usage: SaveCommandLineHistory([fname, ][app])");
+  fi;
+  if not IsBound(fname) then
     if IsExistingFile(GAPInfo.UserGapRoot) then
       fnam := Concatenation(GAPInfo.UserGapRoot, "/history");
     else
       fnam := UserHomeExpand("~/.gap_hist");
     fi;
   fi;
-  if true in arg then
-    append := true;
-  else
-    append := false;
-  fi;
   hist := GAPInfo.History.Lines;
+  hist := hist{[GAPInfo.History.StartPosition + 1
+                .. Length(GAPInfo.History.Lines)]};
   out := OutputTextFile(fnam, append);
   if out = fail then
     return fail;
   fi;
   SetPrintFormattingStatus(out, false);
   WriteAll(out,JoinStringsWithSeparator(hist,"\n"));
-  WriteAll(out,"\n");
+  if Length(hist) > 0 then
+    WriteAll(out,"\n");
+  fi;
   CloseStream(out);
   return Length(hist);
 end);
--- a/lib/init.g
+++ b/lib/init.g
@@ -763,8 +763,8 @@ end );
 ##  
 CallAndInstallPostRestore( function()
   if UserPreference("SaveAndRestoreHistory") = true then
-    InstallAtExit(SaveCommandLineHistory);
-    ReadCommandLineHistory();
+    InstallAtExit(SaveCommandLineHistory, [true]);
+    GAPInfo.History.StartPosition := ReadCommandLineHistory();
   fi;
 end );
 
--- a/doc/ref/mloop.xml
+++ b/doc/ref/mloop.xml
@@ -1029,12 +1029,20 @@ The optional argument <A>ret</A> will be passed to <Ref Func="GapExitCode"/>.
 
 
 <ManSection>
-<Func Name="InstallAtExit" Arg='func'/>
+<Func Name="InstallAtExit" Arg='func, [args]'/>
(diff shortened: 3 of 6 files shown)


lib/cmdledit.g
```diff
@@ -700,28 +700,38 @@ BindKeysToGAPHandler("\022");
...
   local  fnam, append, hist, out;
 
-  if Length(arg) > 0 then
+  if Length(arg) = 1 then
+    if IsString(arg[1]) then
+      fnam := arg[1];
+      append := false;
+    elif arg[1] in [false, true] then
+      append := arg[1];
+    fi;
```
ChrisJefferson, line 709: you probably need an error case here, for when you don't understand arg1?

ssiccha: Oh yeah, thanks!

lib/init.g
```diff
@@ -751,8 +751,8 @@ end );
 ##  
 CallAndInstallPostRestore( function()
   if UserPreference("SaveAndRestoreHistory") = true then
-    InstallAtExit(SaveCommandLineHistory);
-    ReadCommandLineHistory();
+    InstallAtExit(SaveCommandLineHistory, [true]);
```
fingolfin, line 754: Instead of complicating `InstallAtExit`, why not just use a closure here?
```suggestion
    InstallAtExit({} -> SaveCommandLineHistory(true));
```

fingolfin (changes requested):
I would strongly prefer not to change `InstallAtExit`. That would reduce this patch substantially, too.

frankluebeck:
This pull request does three things:

1. Extend `InstallAtExit` to deal with functions with arguments

This part really complicates the simple mechanism for exit functions and it looks totally unnecessary to me.
All of this is not needed. In your example application you could instead just install the function
```function() SaveCommandLineHistory(true); end```

2. Changes the behaviour of `SaveCommandLineHistory`

The code now allows to call `SaveCommandLineHistory(true);` without giving a file name which is then assumed to be the default history file. I don't think that this is sensible behaviour for the default history file because it would grow indefinitely.

3. Changes the meaning of the user preference `"SaveAndRestoreHistory"` when set to `true`

With the proposed change GAPs default history file would grow forever and ignore the associated user preference `"HistoryMaxLines"`. (And GAP would need to read a bigger and bigger history file at startup.) This is never a sensible behaviour.

==========================
So far GAP does the following when `"SaveAndRestoreHistory"` is set to true: On startup it reads the default history file and stores the lines in  its history. Then lines are added during the new session; note that old lines are deleted if needed such that the history does never contain more than `"HistoryMaxLines"` lines. When GAP is regularly quit it substitutes its default history file by the history of the current session.

At least if one runs GAP sessions one after another this seems the correct and intended behaviour to me.

Of course, one can ask what should be done if often several GAP sessions are run in parallel since they all use the same history file. I don't see an obvious answer. Currently, the history file contains the history of the GAP session which was closed last. This sounds fine to me, and it works fine for me in practice (I use the save and restore history mechanism a lot).

As long as the history file is short one could imagine to reread the current history file, prepend it to the history, shorten it to `HistoryMaxLines` and then to save the history (at exit). But after a short while the history will always contain `HistoryMaxLines` lines; then this strategy would lead to the current behaviour.

ssiccha:
Argh, I missed the totally obvious easier version!

ssiccha:
@frankluebeck
> 2. Changes the behaviour of SaveCommandLineHistory
> The code now allows to call SaveCommandLineHistory(true); without giving a file name which is then assumed to be the default history file. I don't think that this is sensible behaviour for the default history file because it would grow indefinitely.

I did not realize that my change now makes it possible for `SaveCommandLineHistory` to grow the line history file beyond `HistoryMaxLines`. That was not on purpose and I'll definitely change that.

And I just realized that I completely misunderstood the documentation. I thought that `SaveCommandLineHistory( fname, false )` would only write the "command line history" _of the current session_, that is excluding the part that was read in from the history file at the beginning, while `SaveCommandLineHistory( fname, true )` keeps the content of the history file and appends the "command line history" _of the current session_. I think that should be clarified because I find it super counter-intuitive that the append mode essentially doubles the size of the history file (as long as we're below `HistoryMaxLines`).

Changing the behaviour of `SaveCommandLineHistory` is out of the question IMO.

> Of course, one can ask what should be done if often several GAP sessions are run in parallel since they all use the same history file. I don't see an obvious answer. Currently, the history file contains the history of the GAP session which was closed last. This sounds fine to me, and it works fine for me in practice (I use the save and restore history mechanism a lot).

I would suggest the following: if a GAP session exits regularly, then it takes the history file at the moment that it exits and appends the "new lines" which were entered during that session. It makes sure that the history file does not exceed `MaxHistoryLines` lines by deleting lines from the beginning accordingly.

Could there in principle appear problems if two sessions exit such that they try to write into the history file at the same time? Or does `OutputTextFile` make sure that we don't have garbled writes?

---

I usually have a second session open which I use almost exclusively to read the documentation. And whenever I close that session last, I use the history from the session where I've actually been doing stuff.

Note for myself: I need to
- [ ] repair my version of `SaveCommandLineHistory`
- [ ] when lines from history get removed (due to exceeding `HistoryMaxLines`) adjust `StartPosition`

fingolfin:
@ssiccha left academia some time ago and I don't think he'll come back to "finish" this PR (whatever that would mean), so I am going to close this.

PR #4155, opened 2020-10-25
Performance Enhancements for rewriting

hulpke:
Performance Enhancements (mostly motivated by work with extensions) for larger cases. They are too minor to merit special mention in the release notes:

- Avoid unneccessary checks when computing a RWS for a group
- Improvements in computing induced/modulo PCGS if the groups are abelian of high (hundreds) rank
- Minor speedup of cohomology (if there are many generators)

I would like them to still go into 4.12, as they will help with examples for a paper -- thus listed as high priority

### DIFF
--- a/lib/dict.gi
+++ b/lib/dict.gi
@@ -305,7 +305,7 @@ end);
 #F  NewDictionary(<objcoll>,<look>)
 ##
 InstallGlobalFunction(NewDictionary,function(arg)
-local hashfun,obj,dom,lookup,maxblist;
+local hashfun,obj,dom,lookup,maxblist,forcesort;
   obj:=arg[1];
   lookup:=arg[2];
   if Length(arg)>2 then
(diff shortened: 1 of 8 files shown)


fingolfin:
There is no need to worry -- right now, it is completely unclear when the next GAP release will happen. Sadly, I have little hope that it'll be before 2021. I simply don't have the time and energy to organize a complete new release process, and the old one (which hinged in great parts on @alex-konovalov ) simply is not working anymore, due to Alex having a different job and hence naturally (!) different priorities. The fact that our public CI just evaporated makes the whole situation worse.

Note that I do have a plan for a revised release process that involves far less work by automating most of it, thus allowing us to have frequent and timely release in the coming years again; but this still (just like the new CI setup) requires a one-time push by some people to get it working. I initially hoped to get this done in the GAP Days Spring 2020, but in the end that was only enough for the great people participating there to rescue the website setup (a prerequisite for the release process). I wanted to organize GAP Days Fall 2020, but the lack of volunteers to co-organize plus our new baby prevented that. So, I am hoping that we can have GAP Days Spring 2021 (after the lecture period is over) and get this settled for good.

fingolfin (approved):
Luckily the CI tests for this still went through. Code looks OK on a cursory check (not that I expected differently!), so as usual I'll just trust that Alexander knows what he's doing :-).

PR #4159, opened 2020-10-30
Added comments and improvements to src/baltree.h

rbehrends:
This change does three things:

* Add comments that explain how scapegoat trees work and the details of our implementation.
* Change the rebalancing operation to work in place instead of allocating extra memory.
* Add a defensive programming safeguard to protect against potential performance degradation.

The last change in particular is not needed for correctness and is currently a no-op in practice. However, it protects against potential performance degradations in case future changes misconstruct the `min_nodes_for_height` table. The rationale is that such edge cases are difficult to detect through tests. We could insert an assertion instead, but that assertion may only trigger in rare edge cases and not show up in tests, but only in production runs.

As scapegoat trees are required to only be loosely height-balanced, we therefore simply adjust the implementation to deal with such a situation, should it arise.

Note: this change still needs to be tested in the CI environment for Oscar, as this code is currently exclusively being used for the Julia GC. Setting up these tests to use a non-standard GAP version is a bit involved, so it has not yet been done.

### DIFF
--- a/src/baltree.h
+++ b/src/baltree.h
@@ -61,11 +61,16 @@
 #define BALANCED_TREE_INIT
 
 // Scapegoat trees are height-balanced trees. With respect to a
-// parameter alpha, a binary tree is height-balanced, if
+// parameter alpha, with 0.5 < alpha < 1, a binary tree is
+// height-balanced, if
 //
-//   height(tree) <= log_1/alpha(size(tree))
+//   height(tree) <= log_{1/alpha}(size(tree))
 //
-// where size(tree) is the number of nodes that a tree contains.
+// where size(tree) is the number of nodes that a tree contains. The
+// closer alpha is to 0.5, the closer a height-balanced tree is to a
+// balanced tree. With alpha = 1, a linked list would be
+// height-balanced. However, the smaller alpha is, the more time will be
+// wasted on rebalancing operations.
 //
 // A binary tree is called weight-balanced iff for any subtree `tree`,
 //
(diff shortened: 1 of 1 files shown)


src/baltree.h
```diff
@@ -45,26 +60,76 @@
...
+//   height(tree) <= log_1/alpha(size(tree))
+//
+// where size(tree) is the number of nodes that a tree contains.
+//
+// A binary tree is called weight-balanced iff for any subtree `tree`,
+//
+//   size(left) <= alpha * size(tree) and
+//   size(right) <= alpha * size(tree)
+//
+// where `left` and `right` are the left and right subtrees of `tree`.
```
fingolfin, line 75: So `size(left)+size(right)+1 = size(tree)` -- I guess that means alpha actually must be between 0.5 and 1 or so?

src/baltree.h
```diff
@@ -45,26 +60,76 @@
...
+// This array contains the minimum number of nodes that
+// a binary tree of a given height can contain while still
+// being height-balanced. Any tree with fewer nodes is too
+// sparse to be height balanced.
+static Int min_nodes_for_height[MaxTreeDepth];
 
 // alpha = alpha_lo / alpha_hi
+// We use integers to allow for more efficient arithmetic.
 static const Int alpha_hi = 3;
 static const Int alpha_lo = 2;
```
fingolfin, line 122: if efficiency is a concern, perhaps `int` instead of `Int` would be better? Not that I checked the actual machine code, but that's the rule of thumb, isn't it?

(I guess in the context of GAP, I'd actually make this an `enum { ALPHA_NUM = 3, ALPHA_DEN = 2 };` but this code is outside the scope of this PR, so this does not have to be touched by this PR. I.e.: feel free to ignore, this is just me rambling on ;-)

fingolfin (commented):
Just some quick remarks: the added/changed comments look fine. But I've not yet had a chance to look at the rewritten *code*, and you write that it should still be tested more thoroughly... perhaps we can sit together next week (possibly in VR) to discuss it?

src/baltree.h
```diff
@@ -45,26 +60,76 @@
 #ifndef BALANCED_TREE_INIT    // non-generic part shared by all versions
 #define BALANCED_TREE_INIT
 
+// Scapegoat trees are height-balanced trees. With respect to a
+// parameter alpha, a binary tree is height-balanced, if
+//
+//   height(tree) <= log_1/alpha(size(tree))
```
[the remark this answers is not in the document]
rbehrends: It's `log_{1/alpha}(size(tree))`, and yes, it's the same as the negative logarithm. I was sticking to the terminology of the paper. The reason is that it's `log_2(size(tree))` for a perfectly balanced tree, so this is where the derivation comes from. Alpha needs to be in the interval (0.5, 1.0) and I've added a comment for that.

src/baltree.h
rbehrends: `Int` vs `int` doesn't matter here, as they're constants, but I have made them enums to be consistent with the rest of GAP.

fingolfin:
@rbehrends you wrote:
> Note: this change still needs to be tested in the CI environment for Oscar

What's the status of that? Let me know once you tested it, then we can merge it.

rbehrends:
Oscar seems to work. There are test failures, but they are the same that show up on normal tests. The GC (which is the part affected by these changes) appears to work fine.

fingolfin (approved):

PR #4160, opened 2020-10-30
support a new command line option `systemfile`

ThomasBreuer:
... which allows one to read a given file during GAP's initialization,
after the file `lib/system.g` has been read.

The idea is that at this point, the command line options have been
evaluated, and it is a suitable place for customizing the GAP session
beyond command line options.
The motivation for this proposal is oscar-system/GAP.jl/issues/518,
see the discussion there for the background.

P.S.:
Each of the files `lib/system.g` and `src/system.c` claims that the
command line options listed there must be kept in sync with those listed
in the other file.
Apparently they are aleady out of sync.
If the claims are still relevant then of course
the new command line option should be mentioned also in `src/system.c`.

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -574,7 +574,10 @@ end);
 ##
 #T really ???
 
-if GAPInfo.CommandLineOptions.systemfile <> "" and
-   IsReadableFile( GAPInfo.CommandLineOptions.systemfile ) then
-  READ( GAPInfo.CommandLineOptions.systemfile );
+if GAPInfo.CommandLineOptions.systemfile <> ""
+   and not READ( GAPInfo.CommandLineOptions.systemfile ) then
+  PRINT_TO( "*errout*", "Could not read file \"",
+            GAPInfo.CommandLineOptions.systemfile,
+            "\" (command line option --systemfile).\n" );
+  QuitGap(1);
 fi;
--- a/lib/system.g
+++ b/lib/system.g
@@ -111,6 +111,8 @@ BIND_GLOBAL( "GAPInfo", rec(
            help := [ "Disable the GAP read-evaluate-print loop (REPL)" ] ),
       rec( long := "nointeract", default := false,
            help := [ "Start GAP in non-interactive mode (disable REPL", "and break loop)" ] ),
+      rec( long := "systemfile", default := "",
+           help := [ "Read this file after 'lib/system.g'" ] ),
       rec( long := "bare", default := false,
            help := [ "Attempt to start GAP without even needed packages", "(developer tool)" ] ),
       ,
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
If this is useful to Julia, I'm generally in favour of it.

My only thought is I would either delete the "IsReadableFile" check, or at least put in an error message if the file isn't readable, else at the moment GAP will silently continue if you mispell the filename (which is probably not what we want).

lib/system.g
```diff
@@ -571,3 +573,8 @@ end);
 ##  </ManSection>
 ##
 #T really ???
+
+if GAPInfo.CommandLineOptions.systemfile <> "" and
+   IsReadableFile( GAPInfo.CommandLineOptions.systemfile ) then
+  READ( GAPInfo.CommandLineOptions.systemfile );
```
fingolfin, line 579:
```suggestion
if GAPInfo.CommandLineOptions.systemfile <> "" and not READ( GAPInfo.CommandLineOptions.systemfile ) then
    Error( "file \"", name, "\" must exist and be readable" );
```

In view of what @ChrisJefferson suggested (and I agree), how about doing something like the above?

fingolfin (commented):
Thanks! I left a suggestion for a minor improvement.

Regarding the "keep in sync", I think one has to read these two kernel comments together:
```
/* These are just the options that need kernel processing. Additional options will be 
   recognised and handled in the library */

/* These options must be kept in sync with those in system.g, so the help output
   is correct */
```

So I think that the list in the kernel simply ought to be a subset of the list in the library (unless some command line option is intended to be kept "secret").

ChrisJefferson (approved):

lib/system.g
```diff
@@ -571,3 +573,10 @@ end);
 ##  </ManSection>
 ##
 #T really ???
+
+if GAPInfo.CommandLineOptions.systemfile <> ""
+   and not READ( GAPInfo.CommandLineOptions.systemfile ) then
+  Print( "Could not read file \"", GAPInfo.CommandLineOptions.systemfile,
+         "\" (command line option --systemfile)\n" );
+  QuitGap();
```
fingolfin, line 581:
```suggestion
  QuitGap(1);
```
... so that we can detect from scripts that there was an error.

lib/system.g
```diff
@@ -571,3 +573,10 @@ end);
 ##  </ManSection>
 ##
 #T really ???
+
+if GAPInfo.CommandLineOptions.systemfile <> ""
+   and not READ( GAPInfo.CommandLineOptions.systemfile ) then
+  Print( "Could not read file \"", GAPInfo.CommandLineOptions.systemfile,
```
fingolfin, line 579:
```suggestion
  PRINT_TO( "*errout*", "Could not read file \"", GAPInfo.CommandLineOptions.systemfile,
```

fingolfin (changes requested):
OK, `Error` doesn't work here, but let's still print to errout and set an exit code.

fingolfin (approved):

fingolfin:
Backported to stable-4.11 in cabff6af41cf0ed4c5b460f5b274d621bb3a3476

PR #4165, opened 2020-11-07
Disable automatic compression of .gz files in stable-4.11 again

fingolfin:
This caused a regression in PackageManager and possibly elsewhere; so it is not suitable for the stable-4.11 branch. To clarify: this feature was *not* in GAP 4.11.0, we just backported it there. This PR adjusts that backport: we retain the improved to reading .gz files, but not the new behavior for automatically compression every file whose name ends with .gz.

See also discussion in PR #4128

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -851,7 +851,7 @@ Int SyFopen (
 #endif
 
     /* try to open the file                                                */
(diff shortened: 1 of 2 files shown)


fingolfin:
Luckily we still have AppVeyor set up, so we are not completely without CI, even if this is incomplete. Anyway I also tested this change locally and it should be simple and safe enough...

wilfwilson (approved):

PR #4168, opened 2020-11-11
improve the documentation concerning GASMAN; add CollectGarbage

ThomasBreuer:
GAP can use several garbage collectors (GASMAN, the Julia garbage collector, and the Boehm garbage collector).
Up to now, the GAP documentation mentions only GASMAN.
Several GAP functions make sense only if GASMAN is running, and several command line options make sense only in this case.

The proposed changes try to adjust the documentation to situations where GASMAN is not used.
(More changes in this direction will be needed, below is a list of open questions.)

**Details of the changes**:

- In `debug.xml`, section "Global Memory Information",
  moved the introductory part to a new subsection "Garbage Collection"
  that can be cross-referenced from places where garbage collection
  is mentioned;
  mention in this subsection that several garbage collectors
  can be used by GAP.

- In the functions `GasmanStatistics`, `GasmanMessageStatus`,
  `GasmanLimits`, mention that they work only if GASMAN is running.

- In `run.xml`, section "Saving and Loading a Workspace",
  mention that workspaces can be saved/loaded only if GASMAN is running.

- Added a cross-reference from `SaveOnExitFile` to `SaveWorkspace`.

- Changed the description of GAP's command line options:
  - In `run.xml`, mention that the options `-g`, `-L`, `-o`, `-s`
    are available only if GASMAN is used.
  - Do the same in `system.g`, variable `GAPInfo.CommandLineOptionData`.
    (I had thought about omitting the not supported options from the
    list, but such a filtered list might confuse users who want to
    understand why their options did not work as expected.)

- Call `GASMAN( "message" )` in `SetGasmanMessageStatus`
  only if GASMAN is running, in order to avoid an error message.

- In chapter "Weak Pointers", removed the word GASMAN in the introduction
  and added a reference to "Garbage Collection".
  (Weak pointer objects work also with other garbage collectors than
  GASMAN.)

**Open questions**:

- The command line options `-K` and `-m` are evaluated also when the
  Julia GC is running, but are their values really used?

- The introduction of the section "Command Line Options" refers to a file
  https://www.gap-system.org/Download/INSTALL which exists but seems to
  be outdated --which URL should be given here?
  (The corresponding file in the GAP distribution is `INSTALL.md`.)

- If GASMAN is not running, a file given with the `-L` command line option
  is currently ignored, because `LoadWorkspace` is not called when
  `GAP_ENABLE_SAVELOAD` is not defined.
  Should we better show an error message telling that no workspace
  could be loaded?
  (In earlier versions, there was such a message.)

- The function `GASMAN` is undocumented.
  (The Development manual contains a section "Garbage collection in GAP",
  where GASMAN is said to be "the GAP memory manager".
  This seems to be the right place to add documentation for alternative
  garbage collectors.
  On the other hand, this manual is not part of the GAP distribution.)

  Calling `GASMAN( "collect" )` and `GASMAN( "partial" )` work also
  with the Julia GC; if this is intentional then we should choose
  a better name for the function.
  (`GASMAN( "message" )` works only with the GASMAN GC.)

  The function `GASMAN` is shown in manual examples about weak pointer
  objects. Thus a function for forcing a garbage collection should be
  documented.

- The section "The GASMAN Interface for Weak Pointer Objects"
  in chapter "Weak Pointers" looks GASMAN specific (not only the title).
  How to reformulate it?

### DIFF
--- a/doc/ref/run.xml
+++ b/doc/ref/run.xml
@@ -54,7 +54,7 @@ and to <C>gap -b -q -b -q</C> etc.
 &GAP; for UNIX will distinguish between upper and lower case options.
 <P/>
 As described in the &GAP; installation instructions (see the 
-<F>INSTALL</F> file in the &GAP; root directory, or at
+<F>INSTALL.md</F> file in the &GAP; root directory, or at
 <URL>https://www.gap-system.org/Download/INSTALL</URL>),
 usually you will not execute &GAP; directly. Instead  you  will
 call a (shell) script, with the name <C>gap</C>, which in turn executes  &GAP;.
@@ -155,6 +155,8 @@ toggle; you must use <C>-n</C> to disable line editing.
 <Item>
 tells &GAP; to print a message every  time  a  full  garbage
 collection is performed.
+(This is available only if the <C>GASMAN</C> garbage collector is used,
+see <Ref Subsect="Garbage Collection"/>.)
 <P/>
 <Log><![CDATA[
 #G  FULL 44580/2479kb live   57304/4392kb dead   734/4096kb free
@@ -169,8 +171,10 @@ reclaimed by it, and that 734 kilobytes from a  total  allocated  memory  of
 <C>-g -g</C></Mark>
 <Item>
 If you give the option <C>-g</C> twice, &GAP; prints  a  information  message
-every time a  partial  or  full  garbage  collection  is  performed.  The
-message,
+every time a  partial  or  full  garbage  collection  is  performed.
+(This is available only if the <C>GASMAN</C> garbage collector is used,
+see <Ref Subsect="Garbage Collection"/>.)
+The message,
 <P/>
 <Log><![CDATA[
 #G  PART 9405/961kb+live   7525/1324kb+dead   2541/4096kb free
@@ -202,6 +206,8 @@ switched off. You have to set it explicitly when you want to enable it.
 <Item>
 The option <C>-L</C> tells &GAP; to load a saved workspace. See
 section&nbsp;<Ref Sect="Saving and Loading a Workspace"/>.
+(This is available only if the <C>GASMAN</C> garbage collector is used,
+see <Ref Subsect="Garbage Collection"/>.)
 </Item>
 <Mark><Index Key="-l"><C>-l</C></Index>
 <C>-l </C><A>path_list</A></Mark>
@@ -530,6 +536,10 @@ which GAP will not recognize.
 
 &GAP; workspace files are binary files that contain the data of a &GAP;
 session.
+Currently saving and loading workspace files are supported only when the
+<C>GASMAN</C> garbage collector is used,
+see Section <Ref Subsect="Garbage Collection"/>.
+<P/>
 One can produce a workspace file with <Ref Func="SaveWorkspace"/>,
 and load it into a new &GAP; session using the <C>-L</C> command line option,
 see Section <Ref Sect="Command Line Options"/>.
--- a/doc/ref/run.xml
+++ b/doc/ref/run.xml
@@ -54,8 +54,7 @@ and to <C>gap -b -q -b -q</C> etc.
 &GAP; for UNIX will distinguish between upper and lower case options.
 <P/>
 As described in the &GAP; installation instructions (see the 
-<F>INSTALL.md</F> file in the &GAP; root directory, or at
-<URL>https://www.gap-system.org/Download/INSTALL</URL>),
+<F>INSTALL.md</F> file in the &GAP; root directory),
 usually you will not execute &GAP; directly. Instead  you  will
 call a (shell) script, with the name <C>gap</C>, which in turn executes  &GAP;.
 This  script sets some options which are necessary  to  make  &GAP;
--- a/lib/system.g
+++ b/lib/system.g
@@ -66,14 +66,14 @@ BIND_GLOBAL( "GAPInfo", rec(
       rec( short:= "y", long := "lines", default := "", arg := "<num>", help := ["set number of lines"] ),
       ,
       rec( short:= "g", long := "gasinfo", default := 0,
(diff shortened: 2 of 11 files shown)


doc/ref/run.xml
```diff
@@ -54,7 +54,7 @@ and to <C>gap -b -q -b -q</C> etc.
 &GAP; for UNIX will distinguish between upper and lower case options.
 <P/>
 As described in the &GAP; installation instructions (see the 
-<F>INSTALL</F> file in the &GAP; root directory, or at
+<F>INSTALL.md</F> file in the &GAP; root directory, or at
 <URL>https://www.gap-system.org/Download/INSTALL</URL>),
```
fingolfin, line 58: IIRC on the GAP website, we changed links to that file to point to the repository instead:
```suggestion
<URL>https://github.com/gap-system/gap/blob/master/INSTALL.md</URL>),
```
However, there is risk that this results in a mismatch (when this file changed in master in a way that is incompatible with the version the user is trying to install). 

So the other option is to just drop this link. After all, the `INSTALL.md` file is there, so what's the point in having this link???

Finally, that outdated INSTALL file should indeed be removed, see also https://github.com/gap-system/GapWWW/issues/206

lib/system.g
```diff
@@ -66,14 +66,14 @@ BIND_GLOBAL( "GAPInfo", rec(
       rec( short:= "y", long := "lines", default := "", arg := "<num>", help := ["set number of lines"] ),
       ,
       rec( short:= "g", long := "gasinfo", default := 0,
-           help := ["show GASMAN messages (full/all/no garbage","collections)"] ),
+           help := ["show GASMAN messages (full/all/no garbage","collections)", "(available only if GASMAN is running)"] ),
```
fingolfin, line 69: I think this could be misunderstood, as in "perhaps GASMAN is some software that I need to install and run on my computer for this to work? but where do I get it). Perhaps this is slightly clearer in this regard:
```suggestion
           help := ["show GASMAN messages (full/all/no garbage","collections)", "(only available if GAP uses GASMAN)"] ),
```

Another alternative would be to not show this text if GASMAN is used; and otherwise show a text more like "not available in this GAP build"; or perhaps "not available with the active GAP memory manager" and then at some point print the memory manager as part of the output of this usage text (i.e. at the start or end, say something like "GAP was compiled using the FOOBAR memory manager")

fingolfin (commented):
Thanks, that's a great improvement! A few minor remarks

fingolfin:
> * The command line options `-K` and `-m` are evaluated also when the
>   Julia GC is running, but are their values really used?

These set the C kernel variables `SyStorKill` resp. `SyStorMin`. The former is honored by GASMAN and Boehm GC, but not but by Julia GC. The latter is only honored by GASMAN.

The main reason I didn't limit them further in the GAP kernel is that their values are returned by `GASMAN_LIMITS`, which in turn is used by `GasmanLimits`. Perhaps they should simply be changed to return `infinity` when not used. Then we could disable them also in Julia GC resp. Boehm GC, on the kernel level...

> * The introduction of the section "Command Line Options" refers to a file
>   https://www.gap-system.org/Download/INSTALL which exists but seems to
>   be outdated --which URL should be given here?
>   (The corresponding file in the GAP distribution is `INSTALL.md`.)

I left an inline comment on this.

> * If GASMAN is not running, a file given with the `-L` command line option
>   is currently ignored, because `LoadWorkspace` is not called when
>   `GAP_ENABLE_SAVELOAD` is not defined.
>   Should we better show an error message telling that no workspace
>   could be loaded?
>   (In earlier versions, there was such a message.)

Agreed, showing an error message would be better.

> * The function `GASMAN` is undocumented.
>   (The Development manual contains a section "Garbage collection in GAP",
>   where GASMAN is said to be "the GAP memory manager".
>   This seems to be the right place to add documentation for alternative
>   garbage collectors.
>   On the other hand, this manual is not part of the GAP distribution.)
>   Calling `GASMAN( "collect" )` and `GASMAN( "partial" )` work also
>   with the Julia GC; if this is intentional then we should choose
>   a better name for the function.
>   (`GASMAN( "message" )` works only with the GASMAN GC.)
>   The function `GASMAN` is shown in manual examples about weak pointer
>   objects. Thus a function for forcing a garbage collection should be
>   documented.

Indeed, there should be a function for garbage collection, with a neutral name, say `CollectGarbage()` which takes a boolean as argument.  However, we should make it clear that in normal situation, there should be no need to invoke it.

Also, we still might want to mention something like "in older GAP versions, one can use `GASMA("collect")` for this"; or even something like:
> The function `CollectGarbage` was introduced in GAP 4.12. If your code needs to be compatible with older GAP versions as well, you can use helper code like this:
```GAP
if not CompareVersionNumbers( GAPInfo.Version, "4.12" ) then
  CollectGarbage := function(full)
   if full then
     GASMAN( "collect" );
   else
     GASMAN( "partial" );
   fi;
  end;
```

> * The section "The GASMAN Interface for Weak Pointer Objects"
>   in chapter "Weak Pointers" looks GASMAN specific (not only the title).
>   How to reformulate it?

That section IMHO simply does not belong into the reference manual, it should be in the dev manual instead.

doc/ref/run.xml
ThomasBreuer: I have dropped the link in the updated version of the pull request.

lib/system.g
ThomasBreuer: I have changed the formulation as proposed.
Leaving out the not supported options from the help list may be irritating for users --suppose I have read somewhere how to start GAP, including some fancy command line options, thenI get an error message, look at the help screen, and the relevant options are not listed at all.

fingolfin (approved):
Looks good to me, but I'd appreciate if @ChrisJefferson could also have at least a quick peek at this.

ChrisJefferson (approved):

PR #4170, opened 2020-11-12
Use colors in output of `Test`

fingolfin:
...  and some other test code, to make it easier to see and understand diffs in there.

Consider this test file:
```
gap> 1+1;
2 
gap> 2+2;
4
```
This will actually cause a test failure, due to whitespace, which makes it annoying to discover. Currently, `Test` will produce something like this:

Screenshot 2020-11-12 at 11 48 31

With this PR, you'll get this:
Screenshot 2020-11-12 at 11 55 41

This change is IMHO useful beyond better understanding whitespace diffs. For example when I run the GAP test suite and there are multiple failures, it can be challenging to visually parse what is being displayed. With this PR, I find it much easier to understand what is going on.

Here is a more complex example:
```
gap> 1+1;
2 
gap> 2+2;
4
# comment in a bad place

gap> 3+3;
6
gap> Print("1\n2\n3\n");
1
2
3
gap> true
true
```
Which currently produces this, which I find hard to make sense out (basically, I resorted to trying to fix the first one and then hope I can repeat this; or I use the `rewriteFile:=true` option):
Screenshot 2020-11-12 at 12 16 35

With this PR (the output could still be nicer, but it's a start):
Screenshot 2020-11-12 at 12 17 15

```

### DIFF
--- a/dev/ci.sh
+++ b/dev/ci.sh
@@ -31,6 +31,7 @@ do
   # restore current directory before each test suite
   cd "$BUILDDIR"
 
+  echo "Running test suite $TEST_SUITE"
   case $TEST_SUITE in
   testspecial | test-compile)
     cd $SRCDIR/tst/$TEST_SUITE
@@ -97,9 +98,9 @@ GAPInput
             Print("-----------------------------------------------------\n");
             Print("Loading $pkg ... \n");
             if LoadPackage("$pkg",false) = true then
-              Print("PASS: $pkg\n\n");
+              Print(TextAttr.2, "PASS: $pkg\n\n", TextAttr.reset);
             else
-              Print("FAIL: $pkg\n\n");
+              Print(TextAttr.1, "FAIL: $pkg\n\n", TextAttr.reset);
               AppendTo("fail.log", "Loading failed : ", "$pkg", "\n");
             fi;
 GAPInput
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -499,6 +499,44 @@ end);
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##  
+DeclareGlobalName("TextAttr"); # from GAPDoc
+DeclareGlobalName("DefaultReportDiffColors"); # initialized in Test() or by the user
+BindGlobal("DefaultReportDiff", function(inp, expout, found, fnam, line, time)
+  if UserPreference("UseColorsInTerminal") = true then
+    Print(DefaultReportDiffColors.message);
+    Print("########> Diff in ");
+    if IsStream(fnam) then
+      Print("test stream, line ",line,":");
+    else
+      Print(fnam,":",line);
+    fi;
+    Print(TextAttr.reset, "\n", DefaultReportDiffColors.message);
+    Print("# Input is:", TextAttr.reset, "\n");
+    Print(DefaultReportDiffColors.input);
+    Print(inp);
+    Print(TextAttr.reset, TextAttr.delline, DefaultReportDiffColors.message);
+    Print("# Expected output:", TextAttr.reset, "\n");
+    Print(DefaultReportDiffColors.expected);
+    Print(expout);
+    Print(TextAttr.reset, TextAttr.delline, DefaultReportDiffColors.message);
+    Print("# But found:", TextAttr.reset, "\n");
+    Print(DefaultReportDiffColors.actual);
+    Print(found);
+    Print(TextAttr.reset, TextAttr.delline, DefaultReportDiffColors.message);
+    Print("########", TextAttr.reset, "\n");
+  else
+    Print("########> Diff in ");
+    if IsStream(fnam) then
+      Print("test stream, line ",line,":\n");
+    else
+      Print(fnam,":",line,"\n");
+    fi;
+    Print("# Input is:\n", inp);
+    Print("# Expected output:\n", expout);
+    Print("# But found:\n", found);
+    Print("########\n");  fi;
+end);
+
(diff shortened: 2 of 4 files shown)


ChrisJefferson:
Does this make the output hard to read in less, and similar programs, by filling them with colour codes? (does GAPDoc already handle this by dropping them based on where stdout is going?)

fingolfin:
I've never ever tried to send the output of `Test()` into a pager. But in general, `less` can handle color escape codes just fine with the right settings (namely `-r` or `-R`).

So I just tried `./gap -c 'Test("simple.tst"); QUIT;' | less -r` and it worked fine. But without the `-r`, less indeed prints the escape sequences. But that's the normal thing for `less`, isn't it?

wilfwilson (commented):
I do like the idea of colours in the output. However, I use a terminal with a white background, and the yellow text on a white background is unpleasant to read.

Screen Shot 2020-11-14 at 09 09 51

Could you find a solution for the currently-yellow text that has high contrast with both light and dark terminal backgrounds? Or perhaps only colour the `#######` bit (which doesn't actually need to be read), and leave the other yellow text as it was?

fingolfin:
Yeah colors could be tweak; I also wonder about people who are red-green blind.

alex-konovalov:
Can you deliberately break some more tests for testing this PR, so that we can see how GitHub actions will render colours.

fingolfin:
@wilfwilson there are some more colors we could pick from. Also, background color can be used, and of course any combination for foreground and background color could be used. But of course most terminals also allow overriding these colors, so there won't be a single "optimal" choice, ever... julia has a `--color=yes/no` command line option to be able to disable or enable use of colors. In GAP, we have `ColorPrompt` and I guess we could also add a test option to force disabling/enabling colors.

Screenshot 2020-11-20 at 01 15 45

@alex-konovalov GitHub Actions renders the colors just fine, see e.g

wilfwilson (approved):
Actually I'm happy enough with it as it is.

fingolfin:
@alex-konovalov GH Actions handles the colors just fine. But I introduced a failure so you can look for yourself, e.g. at https://github.com/gap-system/gap/pull/4170/checks?check_run_id=1427795100

@wilfwilson well, I am really open to change the colors, I just don't know what a good compromise might be. Perhaps one could explicitly set foreground *and* background color to increase the chance of good contrast? Playing with the code to change the colors is relatively easy.

wilfwilson:
I had thought about it a bit, and didn't feel that I could suggest an improvement. However I have thought about it a bit more.

I think combining text and bg colours is a good way to make sure that the text looks decent in light and dark terminal windows. And this is my preference (not strongly held opinion though):
* Text with green background should be 'black'
* Text with red background should be 'white'
* The currently yellow text should be given a black background, or this should be inverted to black text on a yellow background.

Here are those styles on white/black terminal backgrounds:

Screen Shot 2020-11-20 at 12 37 29

I think they're all quite clear.

fingolfin:
That's fine by me; but of course it won't be perfect in all cases: someone might have a terminal theme with a very bright red and a super dark green... There are terminals which allow specifying 24bit RGB values... but I guess that's overkill ;-).

Anyway, I used your color suggestions (with yellow-on-black), and also made these potentially configurable via a global record `DefaultReportDiffColors` (undocumented for now).

wilfwilson:
Yes, but although it might look ugly in some bizarre terminal setups, hopefully it'll all still be readable 🙂 

Thanks. I'm happy with it. The record is a good idea.

frankluebeck:
A few comments, since I have experimented a bit with colors in terminals:
  * avoid yellow (in standard terminal settings), it is difficult to read on many terminals 
  * also avoid using background/foreground combinations, these are unpleasant to read (I think I have removed all background settings in GAPDoc)
  * colors should not be used if the user preference "UseColorsInTerminal" is set to false
  * you could use a user preference to enable users to change the colors (e.g., GAPDoc has the user preference "TextTheme"); any setting without yellow is probably a reasonable default for most users

fingolfin:
> A few comments, since I have experimented a bit with colors in terminals:
> 
> * avoid yellow (in standard terminal settings), it is difficult to read on many terminals

I changed it to blue now, which looks OK for me in various terminal theme choices (but of course you can always find places where any choice is problematic, e.g. if someone really loves a blue background for their terminal.... though at least in my terminal, I would then also change the "blue color" to be something else)

> * also avoid using background/foreground combinations, these are unpleasant to read (I think I have removed all background settings in GAPDoc)

While I am not a fan of them, but *for this purpose* I think they make sense, because they make it possible to "see" whitespace issues, which often are an annoying source of test diffs. 

> * colors should not be used if the user preference "UseColorsInTerminal" is set to false

Good point, and added.

> * you could use a user preference to enable users to change the colors (e.g., GAPDoc has the user preference "TextTheme"); any setting without yellow is probably a reasonable default for most users

I'll need to look into the userpref mechanism to see whether I can do it / whether to put it in this PR or a follow-up.

wilfwilson:
I'm happy with this.

fingolfin:
Here is a fresh side-by-side comparison screenshot. The output in the left is basically impossible for me to read correctly. The one on the right with colors still is nasty, but I can manage.
 
Screenshot 2021-02-18 at 00 35 12

fingolfin:
I'd rather not delay this for adding UserPreference: the colors already *are* customizable by tweaking `DefaultReportDiffColors` in `gap.ini`, and if a userpref is desired, we can add one later; but I'd like to benefit from the colors *now*...

PR #4171, opened 2020-11-12
GitHub Actions: re-enable coveralls integration (WIP)

fingolfin:
This is work in progress to restore the Coveralls integration using GitHub Actions. Also, to switch codecov reporting to use the official codecov action.

See also https://github.com/marketplace/actions/coveralls-github-action


fingolfin:
May want to make use of new `profiling` function `OutputLcovCoverage` which was added in https://github.com/gap-packages/profiling/issues/85

fingolfin:
This is obsolete (see PR #4435 where I removed coveralls integration)

PR #4175, opened 2020-11-13
Makefile.gappkg: some defensive changes

fingolfin:
- cope with GAP or GAC not being defined in sysinfo.gap (as in older GAP versions)
- don't implicitly rely on gac resp. libtool creating output directories


wilfwilson (approved):

fingolfin:
Backported in dfef9c8bc12044e72865c2481ddd3069b9511135

PR #4177, opened 2020-11-14
init.g: export function ProcessInitFiles

fingolfin:
Give the function which processes GAPInfo.InitFiles a name so that it can be
accessed from elsewhere. Together with the --systemfile argument, this enables
delayed processing of GAPInfo.InitFiles. in the Julia interface.

I'd like to backport this to stable-4.11 so that I can start using it ASAP in GAP.jl.

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -931,10 +931,10 @@ fi;
 
 ResumeMethodReordering();
 
-InstallAndCallPostRestore( function()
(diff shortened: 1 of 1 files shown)


wilfwilson (approved):

fingolfin:
Backported to stable-4.11 in b61b69af7be985b8eeafb3b35cb040733a438817

PR #4178, opened 2020-11-16
Fix bugs in RestrictedPerm

ChrisJefferson:
The code in RestrictedPerm for handling ranges had a few bugs (I found issue (2) in some code I was working on, when I went to investigate I noticed issue (1))

1) If the list contained negative numbers, () was always returned
2) Ranges with steps which were not step = 1 were mishandled -- but (I believe) this would never return a wrong answer, just an Error about invalid input, when the input was value.

Rather than try to fix the old code, which was both buggy and confusing, I just wrote a (hopefully) much simpler loop, which I believe will be just as efficient (as the only one wasn't trying to do "clever skipping" anyway.

### DIFF
--- a/src/permutat.cc
+++ b/src/permutat.cc
@@ -1748,19 +1748,13 @@ static inline Obj RESTRICTED_PERM(Obj perm, Obj dom, Obj test)
       len = GET_LEN_RANGE(dom);
       p = GET_LOW_RANGE(dom);
       inc = GET_INC_RANGE(dom);
-      while (p<1) {
-        p+=inc;
-        len=-1;
+      if (p < 1 || p + inc * (len - 1) < 1) {
+          return Fail;
       }
-      i=p+(inc*len)-1;
-      while (i>deg) {
-        i-=inc;
-      }
-      p-=1;
-      i-=1;
-      while (p<=i) {
-        ptRest[p]=ptPerm[p];
-        p+=inc;
+      for (i = p; i != p + inc * len; i += inc) {
+          if (i <= deg) {
+              ptRest[i - 1] = ptPerm[i - 1];
+          }
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/permutat.cc
```diff
@@ -1748,19 +1748,10 @@ static inline Obj RESTRICTED_PERM(Obj perm, Obj dom, Obj test)
...
-      while (i>deg) {
-        i-=inc;
-      }
-      p-=1;
-      i-=1;
-      while (p<=i) {
-        ptRest[p]=ptPerm[p];
-        p+=inc;
+      for ( i = p; i != p+inc*len; i += inc) {
+          if(i >= 1 && i <= deg) {
```
wilfwilson, line 1752: You could combine these if you wanted to into something like:
```suggestion
      for ( i = Maximum(1,p); i != p+inc*len && i <= deg; i += inc) {
```
(Not sure how you do `Maximum`)

wilfwilson (approved):
Good catch.

src/permutat.cc
ChrisJefferson: Be aware that inc can be positive or negative, and p could start higher than deg (that's why I do i != p+inc*len, rather than the more usual i < p+inc*len).

ChrisJefferson:
Just to say, we could also change RestrictedPerm to just reject inputs which contain non-positive integers (it doesn't really make sense to restrict a perm on negative inputs), but that would be a change from current behaviour on non-ranges (which don't currently have any bugs I know of).

src/permutat.cc
wilfwilson: Oops, silly me, you're quite right. Please continuing ignoring my wrong suggestion ;-)

fingolfin, line 1752:
```suggestion
      for (i = p; i != p + inc * len; i += inc) {
          if (i >= 1 && i <= deg) {
```

fingolfin (approved):

ChrisJefferson:
Code now reformatted and simplified somewhere, I now (correctly) just reject negative inputs. Added (possibly too many) tests, to make sure all the various boundary conditions are hit properly.

wilfwilson (approved):
Great.

src/permutat.cc
ChrisJefferson: fixed by clang-formatting, thanks

fingolfin (approved):

fingolfin:

PR #4183, opened 2020-11-19
GH Actions: enable ABI=32 tests

fingolfin:
This also unbreaks the 64bit `testpackages` job (by always calling `apt-get update`)


wilfwilson (approved):
Very nice 🙂

PR #4184, opened 2020-11-23
Update bundled GMP to 6.2.1

fingolfin:
Changes between GMP version 6.2.0 and 6.2.1

  BUGS FIXED
  * A possible overflow of type int is avoided for mpz_cmp on huge operands.

  * Overflows are more carefully detected and reported for mpz_pow_ui.

  * A bug in longlong.h for aarch64 sub_ddmmss, not affecting GMP, was healed.

  * mini-gmp: mpz_out_str and mpq_out_str now correctly handle out of
    range bases.

  FEATURES
  * C90 compliance.

  * Initial support for Darwin on arm64, and improved portability.

  * Support for more processors.

  SPEEDUPS
  * None, except indirectly through recognition of new CPUs.

PR #4185, opened 2020-11-23
CI: Test building GAP with bundled GMP and zlib

fingolfin:
While working on PR #4184 I noticed that we don't actually test building with the bundled copy of GMP in GitHub Actions right now; so this PR tries to restore that part of the test coverage. 

Once we've verified this PR does what it promises, we should merge it before PR #4184, then re-run that PR with the tests here (even though I *did* test PR #4184 on my computer).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,21 +15,20 @@ jobs:
         # base test: fast first test
         os: [ubuntu-latest]
         test-suites: ["testinstall"]
-        extra: [""]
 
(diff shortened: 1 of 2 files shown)


fingolfin:
This accidentally removes gmp&zlib for too many jobs. I'll submit a fix when I am back at a computer.

Also there is an issue compiling GMP in 32bits. Haven't looked into that yet (but it used to work...)

wilfwilson (approved):
This seems to be working properly now.

PR #4189, opened 2020-11-26
kernel: restrict compilation of GASMAN  specific code further

fingolfin:


wilfwilson (approved):
This all looks sensible to me.

PR #4190, opened 2020-12-01
kernel: fix syntax error not being reported on REPL

### LINKED ISSUE #4188: REPL silently accepts garbage input
Look at this: notice the two typos in the first input; the first (in PseudoRandom) I fix, but the `:` instead of `;` is left -- and bam, I get no syntax error (but GAP also doesn't do anything).

```
gap> List([1..300], i->Order(PseudoRanom(grp))):
List([1..300], i->Order(PseudoRanom(grp))):
                        ^^^^^^^^^^^
Error, Variable: 'PseudoRanom' must have an assigned value in
  return Order( PseudoRanom( grp ) ); at *stdin*:45 called from
func( C[i] ) at /Users/mhorn/Projekte/GAP/gap.github/lib/coll.gi:665 called from
<function "List">( <arguments> )
 called from read-eval loop at *stdin*:45
brk> quit;
Syntax error: ; expected
List([1..300], i->Order(PseudoRanom(grp))):
                                          ^
gap> List([1..300], i->Order(PseudoRandom(grp))):
gap>
```

This doesn't happen in GAP 4.11.0, so I suspect this is caused by one of my interpreter refactoring sprints.

fingolfin:
Fixes #4188

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1052,6 +1052,7 @@ static Char GetLine(TypInputFile * input)
     input->line[0] = '\0';    // init the pushback buffer
     input->line[1] = '\0';    // empty line buffer
     input->ptr = input->line + 1;
+    input->lastErrorLine = 0;
 
     /* try to read a line                                              */
     if (!GetLine2(input)) {
--- a/tst/testinstall/kernel/scanner.tst
+++ b/tst/testinstall/kernel/scanner.tst
@@ -393,6 +393,7 @@ gap> Read(InputTextString("\"\"\"123"));
 Syntax error: String must end with """ before end of file in stream:1
 gap> obj := """
 Syntax error: String must end with """ before end of file in stream:2
+Syntax error: ; expected in stream:2
 
 #
 # test EOF inside a pragma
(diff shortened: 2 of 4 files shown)


fingolfin:
Unfortunately this revives another bug:
```
# Input is:
obj := """
# Expected output:
Syntax error: String must end with """ before end of file in stream:2
# But found:
Syntax error: String must end with """ before end of file in stream:2
Syntax error: ; expected in stream:2
```

wilfwilson:
> Unfortunately this revives another bug, see `tst/testinstall/kernel/scanner.tst`:
> 
> ```
> # Input is:
> obj := """
> # Expected output:
> Syntax error: String must end with """ before end of file in stream:2
> # But found:
> Syntax error: String must end with """ before end of file in stream:2
> Syntax error: ; expected in stream:2
> ```

Is this a bug? I looks to me like both syntax warnings are valid.

fingolfin:
It's not really a bug; it's a extraneous warning; but after thinking again about it, it's there because the invalid input `"""`  actually ends with a new line, so, yeah, it's OK.

So from my POV this could be merged.

wilfwilson (approved):
I agree.

PR #4193, opened 2020-12-02
Reduce diff between DoLogModRho & DoLogFFERho, add comments

fingolfin:
Also add a couple test cases for their respective helper functions `LogModRhoIterate` and `FFECONWAY.LogFFERhoIterate`.

Moreover, a couple failing tests based on issue #3784 are being added.

This issue is not (yet?) fixed by this PR, but it hopefully makes the code easier to understand; it also aligns the two implementation of Pollard rho for logarithms, to make it easier to spot accidental differences, if any.

I'll try to understand the code better later tonight, but now have appointments.

### DIFF
--- a/hpcgap/lib/ffeconway.gi
+++ b/hpcgap/lib/ffeconway.gi
@@ -1203,10 +1203,9 @@ end);
 # Adapted from the integer code by AH and Sean Gage
 #
 
-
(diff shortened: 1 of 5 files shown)


fingolfin:
This is a dead end in light of PR #5495

PR #4194, opened 2020-12-02
Replace Set(List(l,f)) by Set(l,f)

fingolfin:
This is more compact and more efficient.

(Recursive regex patterns are really useful for these kinds of transformations :-)).


ChrisJefferson (approved):
I checked the first dozen, I did not check every one but I can't imagine how they could cause an issue, assuming tests pass. This can be much more memory efficient, so is a good idea.

wilfwilson (approved):
Excellent.

PR #4195, opened 2020-12-02
Restructure the output of --help

fingolfin:
... by grouping it in various categories; this makes it arguably easier to find something in the list (at least it does help *me*, but YMMV). Also remove mentions of command line options that are not actually supported anymore, like `-a` (which @ChrisJefferson removed some time ago, but we forgot one mention in `system.g`) and `-z` (for which I could not figure what it ever did -- or perhaps it still does something and I just couldn't find it?).

To better show what's going on, here is the output of `gap --help` with this PR:
```
usage: gap [OPTIONS] [FILES]
       run the Groups, Algorithms and Programming system, Version 4.12dev

Startup:
  -h, --help                     print this help and exit
      --version                  print the GAP version and exit
      --print-gaproot            print the primary GAP root and exit
  -b, --banner                   disable/enable the banner
  -c                   <expr>    execute the expression <expr>
      --systemfile     <file>    read this file after 'lib/system.g'

Input/output:
  -q, --quiet                    enable/disable quiet mode
  -e                             disable/enable quitting on <ctrl>-D
  -f                             force line editing
  -n                             prevent line editing
  -E, --readline                 disable/enable use of readline library (if
                                   possible)
  -x, --width          <num>     set line width
  -y, --lines          <num>     set number of lines
  -p                             enable/disable package output mode
                                   (for use by XGAP and similar interfaces)

Memory:
  (may use postfix 'k' = *1024, 'm' = *1024*1024,
                   'g' = *1024*1024*1024):

  -g, --gasinfo                  show GASMAN messages (full/all/no garbage
                                   collections)
                                   (only available if GAP uses GASMAN)
  -m, --minworkspace   <mem>     set the initial workspace size
  -o, --maxworkspace   <mem>     set workspace size where GAP will warn about
                                   excessive memory usage (GAP may allocate more)
                                   (only available if GAP uses GASMAN)
  -K, --limitworkspace <mem>     set maximal workspace size (GAP never
                                   allocates more)
  -s                   <mem>     set the initially mapped virtual memory
                                   (only available if GAP uses GASMAN)

Roots:
  -l, --roots          <paths>   set or modify the GAP root paths
                                   Directories are separated using ';'.
                                   Putting ';' on the start/end of list appends
                                   directories to the end/start of existing list
                                   of root paths
  -r                             disable/enable user GAP root dir
                                   GAPInfo.UserGapRoot

Loading:
  -A                             disable/enable autoloading of suggested
                                   GAP packages
  -D                             enable/disable debugging the loading of files
  -M                             disable/enable loading of compiled modules

Error handling, REPL:
  -T, --nobreakloop              disable/enable break loop and error traceback
      --alwaystrace              always print error traceback
                                   (overrides behaviour of -T)
      --quitonbreak              quit GAP with non-zero return value instead
                                   of entering break loop
      --norepl                   Disable the GAP read-evaluate-print loop (REPL)
      --nointeract               Start GAP in non-interactive mode (disable REPL
                                   and break loop)

Workspaces:
  (only available if GAP uses GASMAN)
  -L                   <file>    restore a saved workspace
  -R                             prevent restoring of workspace (ignoring -L)

Profiling:
      --prof           <file>    Run ProfileLineByLine(<file>) on GAP start
      --memprof        <file>    Run ProfileLineByLine(<file>) on GAP start
                                   with recordMem := true
      --cover          <file>    Run CoverageLineByLine(<file>) on GAP start

Internal developer tools
  -N                             do not use hidden implications
  -O                             disable/enable loading of obsolete files
      --bare                     Attempt to start GAP without even needed
                                   packages

  Boolean options toggle the current value each time they are called.
  Default actions are indicated first.
```

For reference, this is what it looks like right now:
```
usage: gap [OPTIONS] [FILES]
       run the Groups, Algorithms and Programming system, Version 4.12dev

 -h, --help                     print this help and exit
     --version                  print the GAP version and exit
     --print-gaproot            print the primary GAP root and exit
 -b, --banner                   disable/enable the banner
 -q, --quiet                    enable/disable quiet mode
 -e                             disable/enable quitting on <ctrl>-D
 -f                             force line editing
 -n                             prevent line editing
 -E, --readline                 disable/enable use of readline library (if
                                  possible)
 -x, --width          <num>     set line width
 -y, --lines          <num>     set number of lines

 -g, --gasinfo                  show GASMAN messages (full/all/no garbage
                                  collections)
                                  (only available if GAP uses GASMAN)
 -m, --minworkspace   <mem>     set the initial workspace size
 -o, --maxworkspace   <mem>     set workspace size where GAP will warn about
                                  excessive memory usage (GAP may allocate more)
                                  (available only if GAP uses GASMAN)
 -K, --limitworkspace <mem>     set maximal workspace size (GAP never
                                  allocates more)
 -s                   <mem>     set the initially mapped virtual memory
                                  (available only if GAP uses GASMAN)
 -a                   <mem>     set amount to pre-malloc-ate
                                  postfix 'k' = *1024, 'm' = *1024*1024,
                                  'g' = *1024*1024*1024

 -l, --roots          <paths>   set or modify the GAP root paths
                                  Directories are separated using ';'.
                                  Putting ';' on the start/end of list appends
                                  directories to the end/start of existing list
                                  of root paths
 -r                             disable/enable user GAP root dir
                                  GAPInfo.UserGapRoot
 -A                             disable/enable autoloading of suggested
                                  GAP packages
 -D                             enable/disable debugging the loading of files
 -M                             disable/enable loading of compiled modules
 -N                             do not use hidden implications
 -O                             disable/enable loading of obsolete files
 -T, --nobreakloop              disable/enable break loop and error traceback
     --alwaystrace              always print error traceback
                                  (overrides behaviour of -T)
     --quitonbreak              quit GAP with non-zero return value instead
                                  of entering break loop

 -L                   <file>    restore a saved workspace
                                  (available only if GAP uses GASMAN)
 -R                             prevent restoring of workspace (ignoring -L)

 -p                             enable/disable package output mode
     --prof           <file>    Run ProfileLineByLine(<file>) on GAP start
     --memprof        <file>    Run ProfileLineByLine(<file>) on GAP start
                                  with recordMem := true
     --cover          <file>    Run CoverageLineByLine(<file>) on GAP start
     --norepl                   Disable the GAP read-evaluate-print loop (REPL)
     --nointeract               Start GAP in non-interactive mode (disable REPL
                                  and break loop)
     --systemfile     <file>    Read this file after 'lib/system.g'
     --bare                     Attempt to start GAP without even needed
                                  packages (developer tool)

 -c                   <expr>    execute the expression <expr>

  Boolean options toggle the current value each time they are called.
  Default actions are indicated first.
```

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -53,10 +53,16 @@ BIND_GLOBAL( "GAPInfo", rec(
     # These options must be kept in sync with those in system.c, so the help output
     # for those options is correct
     CommandLineOptionData := [
+      rec( section:= ["Startup:"] ),
       rec( short:= "h", long := "help", default := false, help := ["print this help and exit"] ),
       rec( long := "version", default := false, help := ["print the GAP version and exit"] ),
       rec( long := "print-gaproot", default := false, help := ["print the primary GAP root and exit"] ),
       rec( short:= "b", long := "banner", default := false, help := ["disable/enable the banner"] ),
+      rec( short:= "c", default := "", arg := "<expr>", help := [ "execute the expression <expr>"] ),
+      rec( long := "systemfile", default := "", arg := "<file>",
+           help := [ "read this file after 'lib/system.g'" ] ),
+      ,
+      rec( section:= ["Input/output:"] ),
       rec( short:= "q", long := "quiet", default := false, help := ["enable/disable quiet mode"] ),
       rec( short:= "e", default := false, help := ["disable/enable quitting on <ctrl>-D"] ),
       rec( short:= "f", default := false, help := ["force line editing"] ),
@@ -65,19 +71,23 @@ BIND_GLOBAL( "GAPInfo", rec(
            help := ["disable/enable use of readline library (if", "possible)"] ),
       rec( short:= "x", long := "width", default := "", arg := "<num>", help := ["set line width"] ),
       rec( short:= "y", long := "lines", default := "", arg := "<num>", help := ["set number of lines"] ),
+      rec( short:= "p", default := false, help := ["enable/disable package output mode", "(for use by XGAP and similar interfaces)"] ),
+      ,
+      rec( section:= ["Memory:",
+                      "  (may use postfix 'k' = *1024, 'm' = *1024*1024,",
+                      "                   'g' = *1024*1024*1024):"] ),
       ,
       rec( short:= "g", long := "gasinfo", default := 0,
            help := ["show GASMAN messages (full/all/no garbage","collections)", "(only available if GAP uses GASMAN)"] ),
       rec( short:= "m", long := "minworkspace", default := "128m", arg := "<mem>",
            help := ["set the initial workspace size"] ),
       rec( short:= "o", long := "maxworkspace", default := "2g", arg := "<mem>",
-           help := [ "set workspace size where GAP will warn about", "excessive memory usage (GAP may allocate more)", "(available only if GAP uses GASMAN)"] ),
+           help := [ "set workspace size where GAP will warn about", "excessive memory usage (GAP may allocate more)", "(only available if GAP uses GASMAN)"] ),
       rec( short:= "K", long := "limitworkspace", default := "0", arg := "<mem>",
            help := [ "set maximal workspace size (GAP never", "allocates more)"] ),
-      rec( short:= "s", default := "4g", arg := "<mem>", help := [ "set the initially mapped virtual memory", "(available only if GAP uses GASMAN)" ] ),
-      rec( short:= "a", default := "0",  arg := "<mem>",help := [ "set amount to pre-malloc-ate",
-             "postfix 'k' = *1024, 'm' = *1024*1024,", "'g' = *1024*1024*1024"] ),
+      rec( short:= "s", default := "4g", arg := "<mem>", help := [ "set the initially mapped virtual memory", "(only available if GAP uses GASMAN)" ] ),
       ,
+      rec( section:= ["Roots:"] ),
       rec( short:= "l", long := "roots", default := [], arg := "<paths>",
            help := [ "set or modify the GAP root paths",
                      "Directories are separated using ';'.",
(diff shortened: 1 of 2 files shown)


lib/system.g
```diff
@@ -87,36 +97,34 @@ BIND_GLOBAL( "GAPInfo", rec(
...
       rec( short:= "T", long := "nobreakloop", default := false, help := ["disable/enable break loop and error traceback"] ),
       rec( long := "alwaystrace", default := false, help := ["always print error traceback", "(overrides behaviour of -T)"] ),
       rec( long := "quitonbreak", default := false, help := ["quit GAP with non-zero return value instead", "of entering break loop"]),
       ,
-      rec( short:= "L", default := "", arg := "<file>", help := [ "restore a saved workspace", "(available only if GAP uses GASMAN)"] ),
+      rec( section:= ["Workspaces:", "  (only available if GAP uses GASMAN)"] ),
+      rec( short:= "L", default := "", arg := "<file>", help := [ "restore a saved workspace"] ),
       rec( short:= "R", default := false, help := ["prevent restoring of workspace (ignoring -L)"] ),
       ,
-      rec( short:= "p", default := false, help := ["enable/disable package output mode"] ),
```
fingolfin, line 99: This is for use by XGAP and similar tools, I don't think it makes sense to mention it in the `--help` output (just like we don't mention various gac specific options like `-C`)

lib/system.g
```diff
@@ -87,36 +97,34 @@ BIND_GLOBAL( "GAPInfo", rec(
...
-      rec( short := "E", default :=false ),
-      rec( short := "s", default := "4g" ),
```
fingolfin, line 101: -E and -s are already mentioned elsewhere, and with a default value, so these two lines didn't do anything... I think... ???

lib/system.g
```diff
@@ -87,36 +97,34 @@ BIND_GLOBAL( "GAPInfo", rec(
...
-      rec( short := "z", default := "20" ),
```
fingolfin, line 102: According to the manual, `-z` used to be used in some way by `gac` but I have no idea for what exactly....

ChrisJefferson:
Just to say, I'm in general in favour of both fixing the output, and cleaning up unused options -- I did some small things some time ago, but I think a bigger change is reasonable.

frankluebeck:
I guess that not many people are using this output (or are even aware of it). Of course, making things easier to read it never bad.

How about removing some of the more exotic options from this output and adding a hint to the help section
```
?Command Line Options
```
instead?

ThomasBreuer:
@frankluebeck
I thought that a `--help` option is a common way to describe the valid command line options of a program.
And I got the (perhaps wrong) impression from some discussion that the output of the call with `--help` may be turned automatically into man pages or something similar.
When the `--help` output is used to find out which options make sense or why some options do not work as expected, I think it is better if the overview does not get restricted to a few important options, or to those options that are valid in the current GAP installation.
(And if one calls GAP with `--help` because the script used for starting GAP does not work in the given situation then a hint to GAP's online help is not helpful.)

fingolfin:
I only ever use `--help` resp. `gap -h`, which is pretty much the norm for command line tools, and I think I am not alone, given that fact that manual section "Command Line Options" regularly is forgotten to be updated when people change the list of available options...

And yes, there are tools which try to parse the output and generate man pages from it, although that's not a great way to do; but we could of course try to write our own tool which uses our internal list of options on the GAP level to emit a man page (but then the various "optional" arguments for HPC-GAP or different garbage collectors would have to be treated with some care). Or perhaps even better, generate it from the GAPDoc sources for the manual section "Command Line Options" somehow...

fingolfin:
I've updated this PR now and resolved all my internal TODO comments.

I agree there is more one could/should do, but it seems to me all of these ideas go beyond the scope of this PR? Are there any concerns with the general idea of this PR? Or perhaps specific concerns (like the chosen groups, their names, their order, ...)? I am open to suggestions there, I don't claim these are great, but they are already helpful to me.

wucas (approved):
This is a great improvement! I often use (as I was taught to do when I learned Linux) `gap --help` as I often forget the options and having them grouped makes it much, much easier to find what I'm looking for. There might be better names but tbh I cannot think of any and I like them. 
I wouldn't remove options that might be relevant from this output though. I'm always pretty annoyed when a CLI doesn't tell me options that I would have needed. Mentioning the help page is of course good if it is up to date then.

PR #4202, opened 2020-12-30
Add option to skip guard pages for Julia GC.

rbehrends:
This is an experimental option to allow us to skip the scanning of guard pages rather than relying on intercepting segmentation faults if the task stack scanner encounters one.

It is currently enabled to facilitate testing, but should probably be disabled by default and only be used if changes in the Julia implementation necessitate it, as intercepting segmentation faults relies on fewer assumptions about stack layout.

It also needs more testing, especially in the context of actual GAP.jl uses.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -48,6 +48,10 @@
 **  Various options controlling special features of the Julia GC code follow
 */
 
+// if SKIP_GUARD_PAGES is set, stack scanning will attempt to determine
+// the extent of any guard pages and skip them if needed.
+// #define SKIP_GUARD_PAGES
+
 // if SCAN_STACK_FOR_MPTRS_ONLY is defined, stack scanning will only
 // look for references to master pointers, but not bags themselves. This
 // should be safe, as GASMAN uses the same mechanism. It is also faster
@@ -80,6 +84,10 @@
 // can be difficult. As the GC already assumes that the memory
 // range goes from 0 to 2^k-1 (region tables), we simply convert
 // to uintptr_t and compare those.
+//
+#ifdef SKIP_GUARD_PAGES
+#include <pthread.h>
+#endif
 
 static inline int cmp_ptr(void * p, void * q)
 {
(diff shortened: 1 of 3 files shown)


src/julia_gc.c
```diff
@@ -597,8 +607,31 @@ static void SetupSafeRestoreHandlers(void)
...
+
+static size_t guardpages_size;
+
+void SetupGuardPagesSize(void) {
+    // This is a generic implementation that assumes that all threads
+    // have the default guard pages. This should be correct for the
+    // current Julia implementation.
+    pthread_attr_t attr;
+    pthread_attr_init(&attr);
+    if (pthread_attr_getguardsize(&attr, &guardpages_size) < 0) {
```
fingolfin, line 620: unfortunately this causes a linker error in the CI:
```
undefined reference to symbol 'pthread_attr_getguardsize@@GLIBC_2.2.5'
```
I guess we need to pass a suitable linker flag, something like `-pthreads` or so?

rbehrends: Ah yes, that's a Linux thing where you need to use `-pthread` to compile and link. I think this is something that I should add to the configure script for the `--with-gc=julia` case (Julia already requires pthreads, so it's just a matter of getting compilation and linking right).

ChrisJefferson (approved):

ChrisJefferson:
While I'm not sure exactly what this is up to, I'm happy to approve it, as it does not effect any part of GAP outside the Julia integration.

fingolfin:
I am trying to revive this PR and get it merged; because it is *super* helpful for debugging (with the regular GC code, one gets constant segfaults as part of stack scanning, which are handled, but basically make debugging via gdb / lldb impossible).

fingolfin:
My plan is to enable this option in the Julia `GAP_jll` package (possibly by having multiple `libgap` version in there: one with this one and one with this off; and possibly other build variations. We'll see.

fingolfin (approved):

PR #4206, opened 2021-01-06
fix DirectoriesPackageLibrary/Programs

ThomasBreuer:
... for the case that *another* package is the current package to be loaded at the time when the function gets called.

Apparently this did not happen until now, otherwise it would have been observed earlier.

## Text for release notes

Suppose that one of the functions `DirectoriesPackageLibrary`, `DirectoriesPackagePrograms` is called with first argument `"a"`, *and* that this call happens while another package with name `"b"`, say, gets loaded, in the sense that `GAPInfo.CurrentPackage` is bound to an info record for `"b"`.
Before the fix, the directory path in question for `"b"` was returned. After the fix, the path for `"a"` is returned.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -1198,8 +1198,9 @@ InstallGlobalFunction( DirectoriesPackagePrograms, function( name )
     if IsBound( GAPInfo.PackagesLoaded.( name ) ) then
       # The package is already loaded.
       installationpath:= GAPInfo.PackagesLoaded.( name )[1];
-    elif IsBound( GAPInfo.PackageCurrent ) then
-      # The package is currently going to be loaded.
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Makes sense! Sorry for reviewing this so later, somehow I didn't notice this PR :-(

Out of curiosity, where / how did you encounter this?

ThomasBreuer:
@fingolfin There will be code in the next version of the AtlasRep package which checks for possible "shortcuts":
If some other GAP package provides a data extension which is distributed together with that package then AtlasRep decides to prefer the locally available files to remote ones --one need not download these files (on demand) since they are already there.
In order to decide about the availability, the code checks for the presence of the relevant subdirectory of the other package.
And this check may happen during the time when AtlasRep gets loaded.

PR #4210, opened 2021-01-09
Update Julia integration

fingolfin:
- remove obsolete `GAP_register_GapObj`
- refactor initialisation of ForeignGAP module
- add `GAP_InitJuliaMemoryInterface`: This allows the GAP_jll Julia package to initialize the GAP-Julia memory interface *before* the actual GAP initialization. That is, it will allow resolving https://github.com/oscar-system/GAP.jl/issues/591.

I've already been testing a version of this locally for some time.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -895,23 +895,6 @@ static uintptr_t BagMarkFunc(jl_ptls_t ptls, jl_value_t * obj)
     return YoungRef;
 }
 
-//
-// This function is called from GAP.jl to set the super type our custom Julia
-// types to GAP.GapObj
-//
-void GAP_register_GapObj(jl_datatype_t * gapobj_type)
-{
-    if (!gapobj_type || !jl_is_datatype(gapobj_type)) {
-        Panic("GAP_register_GapObj: GapObj is not a datatype");
-    }
-
-    // DO THE GREAT MAGIC HACK (yes we are bad citizens :/) and
-    // change the super type of ForeignGAP.MPtr to GapObj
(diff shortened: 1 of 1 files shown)


rbehrends:
This looks mostly like a reorganization, so I don't see any obvious problems. One thing that I am not entirely sure about is whether initializing `StackAlignBags` to `sizeof(void *)` is 100% safe. For a different GC, I recall having had [to actually use half-word steps](https://github.com/nim-lang/Nim/pull/2671) for scanning registers saved via `setjmp()` on some architecture (Linux or macOS, I don't recall which), because pointers weren't saved on word-aligned boundaries. I'll note that this seems to have been triggered by gcc versions that are basically not in use anymore, so I'm not sure how serious of a concern that would be in the first place.

Note: I don't want to talk up what may be a non-issue; the previous paragraph is long simply because the issue is obscure, not because it is necessarily important.

fingolfin:
No GAP object should have been allocated between this "early initialization " and the latter real initialization; but a GC of course could happen. This I init StackAlignBags to be non-zero (to avoid endless looping) and also hopefully safe in archs which don't permit unaligned reads.

fingolfin:
It still fails CI, though :-(. I'll investigate later.

rbehrends:
> No GAP object should have been allocated between this "early initialization " and the latter real initialization; but a GC of course could happen. This I init StackAlignBags to be non-zero (to avoid endless looping) and also hopefully safe in archs which don't permit unaligned reads.

Okay, then there should be no issue.

fingolfin:
Backported to stable-4.11 in PR #4216

PR #4212, opened 2021-01-12
inside `HELP`, give access to the original search string

ThomasBreuer:
`HELP` calls `HELP_SHOW_MATCHES`, which calls `HELP_GET_MATCHES`, which calls the `SearchMatches` function of the help book handler.

Currently a case insensitive search string is entered in these calls, because `HELP` calls `SimpleString` on its input.
This is not a problem because the help requests to GAP manuals are case insensitive.

However, although it is not difficult to access the help system of Julia via GAP's help system, one runs into the problem that Julia's help is case sensitive.
With the changes proposed here, it will be possible to fetch help for Julia functions and types via inputs of the form `?Julia:sqrt` or `?Julia:Base.size`.
(See oscar-system/GAP.jl/pull/566 for details about the Julia side.)

The idea is as follows:

- In order to give the help book handler access to the (case sensitive) original search string, including the prefix that specifies the book, `HELP` now sets the global option `HELP_TOPIC` when it calls `HELP_SHOW_MATCHES`.
(I would prefer to carry the original search string from `HELP` to the help book handler function as a proper argument, but this does not work already because `Browse` replaces `HELP_SHOW_MATCHES` by a custom function.)

- In order to make the inputs `?-` and `?+` work, a new list `HELP_ORIG_TOPIC_RING` containing the original search strings
has been introduced.

### DIFF
--- a/lib/helpbase.gi
+++ b/lib/helpbase.gi
@@ -1172,6 +1172,11 @@ BindGlobal("HELP_TOPIC_RING", ListWithIdenticalEntries( HELP_RING_SIZE,
 if IsHPCGAP then
   LockAndMigrateObj(HELP_TOPIC_RING,HELP_REGION);
 fi;
+BindGlobal("HELP_ORIG_TOPIC_RING", ListWithIdenticalEntries( HELP_RING_SIZE,
+                                             "welcome to gap" ));
+if IsHPCGAP then
+  LockAndMigrateObj(HELP_ORIG_TOPIC_RING,HELP_REGION);
+fi;
 # here we store the last shown topic, initialized with 0 (leading to
 # show "Tutorial: Help", see below)
 BindGlobal("HELP_LAST", rec(MATCH := 0, BOOK := 0, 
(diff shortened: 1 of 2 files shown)


lib/helpbase.gi
```diff
@@ -1172,6 +1172,11 @@ BindGlobal("HELP_TOPIC_RING", ListWithIdenticalEntries( HELP_RING_SIZE,
 if IsHPCGAP then
   LockAndMigrateObj(HELP_TOPIC_RING,HELP_REGION);
 fi;
+BindGlobal("HELP_ORIG_TOPIC_RING", ListWithIdenticalEntries( HELP_RING_SIZE,
+                                             "welcome to gap" ));
+if IsHPCGAP then
+  LockAndMigrateObj(HELP_ORIG_TOPIC_RING,HELP_REGION);
+fi;
```
fingolfin, line 1179: Please apply the same changes to `hpcgap/lib/helpbase.gi` (can be done relatively pain-free with `git diff` + `patch` with suitable options.

Alternatively, we can also first merge `hpcgap/lib/helpbase.gi` into `lib/helpbase.gi`, at the cost of a couple of `atomic HELP_REGION do ... end` and `AtomicRecord` and `MigrateObj` calls being inserted...

fingolfin:
Seems sensible to me overall, thank you!

lib/helpbase.gi
ThomasBreuer: O.K.

fingolfin (approved):
Looks good to me

frankluebeck (commented):
This is an elegant solution, given that any need for the original, case sensitive, search string was not anticipated when the HELP mechanism was designed.

fingolfin:
Backported to stable-4.11 in commit 7a2dfbec79d3d20a4ac4d5f174eaf35aa34387d7

PR #4213, opened 2021-01-12
fixed `HELP_SEARCH_ALTERNATIVES`

ThomasBreuer:
... for the case that there is exactly one character behind the pattern:

Before the change:
```
gap> HELP_SEARCH_ALTERNATIVES( "Centralizer" );
[ "Centralise", "Centralize" ]
```

After the change:
```
gap> HELP_SEARCH_ALTERNATIVES( "Centralizer" );
[ "Centraliser", "Centralizer" ]
```

### DIFF
--- a/lib/helpbase.gi
+++ b/lib/helpbase.gi
@@ -214,7 +214,7 @@ if Length(positions) > 0 then # matches found
     begin := Minimum( patterns[i].finish+1, Length(topic) );
   od;
 
-  if begin < Length( topic ) then
+  if begin <= Length( topic ) then
     Add( chop, [ topic{[begin..Length(topic)]} ] );
   fi;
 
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

lib/helpbase.gi
```diff
@@ -214,7 +214,7 @@ if Length(positions) > 0 then # matches found
     begin := Minimum( patterns[i].finish+1, Length(topic) );
   od;
 
-  if begin < Length( topic ) then
+  if begin <= Length( topic ) then
```
fingolfin, line 217: Please apply the same fix to `hpcgap/lib/helpbase.gi`

fingolfin (changes requested):

lib/helpbase.gi
ThomasBreuer: O.K.

fingolfin (approved):

Issue #3030, opened 2018-11-20
libgap API: How much (if any) of GASMAN to expose?


embray:
Currently none of the garbage collection related functions are (intentionally) exposed in libgap-api.h.  At the very least least the `markBagsCallback` that currently gets passed to `GAP_Initialize` needs to be able to call `MarkBag` on objects being tracked by the user.

If you didn't want to expose GASMAN internals at all (e.g. maybe for working with GAP built to use a different GC?) an alternative might be to change the `markBagsCallback` to simply return a pointer to an array of Bags (or some more opaque object pointer type) that the user wishes to be marked as active by the GC.  So instead of the current implementation of

```
    /* allow installing a custom marking function. This is used for integrating
       GAP (possibly linked as a shared library) with other code bases which use
       their own form of garbage collection. For example, with Python (for
       SageMath) or Julia. */
    if (ExtraMarkFuncBags) {
        (*ExtraMarkFuncBags)();
    }
```

something roughly more like

```
if (ExtraMarkFuncBags) {
    Bag* tmp = ExtraMarkFuncBags();
    for (tmp; tmp != NULL; tmp++) {
        MarkBag(*tmp);
    }
}
```

just a thought, in case you would prefer to keep GC-related details more private.

fingolfin:
I have no problem with exposing `MarkBag` as `GAP_MarkBag`, it is supported by all three GC implementations, too, and if we ever add more, they'll also have it. BTW, you can also see that we (well, I at least) consider this acceptable by looking at https://hackmd.io/s/SJzZVOLem . So if you want to submit a PR adding it, we'll merge it.

embray:
Okay, fine by me!  It would be good to have `CollectBags` as well, which I also see in your list of "functions which are probably OK for the API".

fingolfin:
Yeah, `CollectBags` is fine, too. I would recommend only exposing the `full` argument, though, so e.g. `void GAP_ CollectBags(int full)`. Also, feel free to come up with a name where it is "easier to guess what it means"; perhaps `GAP_CollectGarbage`? I find it hard to judge this, as of course I am intimately familiar with the internal GAP "API", and so for me, the closer the new name is to the internal name, the "easier"; but people who just want to use the API might appreciate some fresh air.

embray:
[...]

PR #4215, opened 2021-01-18
libgap API: expose minimal interface to garbage collector

embray:
Add libgap interfaces to the garbage collector.  As discussed in #3030 
these are probably the minimal interfaces that need to be exposed, and
are compatible with all current GC implementations.

As suggested at https://github.com/gap-system/gap/issues/3030#issuecomment-440259263
GAP_CollectBags only takes a `full` argument, but not `size`.

Copy-editing may be needed on the documentation the header.

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -75,7 +75,7 @@ void GAP_MarkBag(Obj obj)
     MarkBag(obj);
 }
 
-UInt GAP_CollectBags(UInt full)
+void GAP_CollectBags(BOOL full)
 {
     CollectBags(0, full);
 }
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -68,6 +68,20 @@ void GAP_Initialize(int              argc,
 
 
 ////
+//// Garbage collector interface
+////
+void GAP_MarkBag(Obj obj)
+{
+    MarkBag(obj);
+}
+
+UInt GAP_CollectBags(UInt full)
+{
+    CollectBags(0, full);
+}
+
+
+////
 //// program evaluation and execution
 ////
 
--- a/src/libgap-api.h
+++ b/src/libgap-api.h
@@ -175,10 +175,10 @@ void GAP_MarkBag(Obj obj);
 // Manually run the garbage collector.
 // A <full> collection checks all previously allocated objects, including those
 // that have survived at least one previous garbage collection.
-// A partial collection will attempt to clean up only recently-allocated
+// A partial collection will attempt to clean up only recently allocated
 // objects which have not been garbage-collected yet, and is hence normally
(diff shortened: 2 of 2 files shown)


src/libgap-api.c
```diff
@@ -67,6 +67,20 @@ void GAP_Initialize(int              argc,
...
 
+////
+//// Garbage collector interface
+////
+void GAP_MarkBag(Obj obj)
+{
+    MarkBag(obj);
+}
+
+UInt GAP_CollectBags(UInt full)
```
fingolfin, line 78:
```suggestion
UInt GAP_CollectBags(BOOL full)
```

src/libgap-api.h
```diff
@@ -161,6 +161,26 @@ void GAP_Initialize(int              argc,
...
+// GAP_Initialize.
+void GAP_MarkBag(Obj obj);
+
+// Manually run the garbage collector.
+// A <full> collection checks all previously allocated objects, including those
+// that have survived at least one previous garbage collection.
+// A partial collection will attempt to clean up only recently-allocated
+// objects which have not been garbage-collected yet, and is hence normally
+// a faster operation.
+UInt GAP_CollectBags(UInt full);
```
fingolfin, line 181:
```suggestion
UInt GAP_CollectBags(BOOL full);
```

src/libgap-api.h
```diff
@@ -161,6 +161,26 @@ void GAP_Initialize(int              argc,
...
+// a GAP object to be long-lived beyond the context of the stack frame where
+// it was created, it is necessary to call GAP_MarkBag on the object when
+// the garbage collector is run by the markBagsCallback function passed to
+// GAP_Initialize.
+void GAP_MarkBag(Obj obj);
+
+// Manually run the garbage collector.
+// A <full> collection checks all previously allocated objects, including those
+// that have survived at least one previous garbage collection.
+// A partial collection will attempt to clean up only recently-allocated
```
fingolfin, line 178:
```suggestion
// A partial collection will attempt to clean up only recently allocated
```

fingolfin (approved):
Looks good to me but it'd be nice if @ChrisJefferson also had a look

fingolfin (approved):
I took the liberty of applying some tweaks, to make the tests pass and to address some minor nitpicks of mine. Hope that was OK, of course we can still revise!

Still would like a second opinion by @ChrisJefferson . But in general this seems to me to be OK to merge

ChrisJefferson:
This is fine to merge with me.

I point out the existence of InitGlobalBag, it might be places where you are using MarkBag you could be using InitGlobalBag if you only set some things up once. I wouldn't bother adding a wrapper for it until someone wants it.

PR #4217, opened 2021-01-19
Add GasmanDisableMemoryLimits

ChrisJefferson:
This gives a way in GAP code of disabling GAP entering the break loop when memory is full, or stopping altogether if the upper memory limit is reached.

This might seem weird, but I find it useful for two reasons.

1) When writing code to be run automatically it stops me having to figure out upper limits to give.

2) It is useful in Windows, where it is hard to add command line parameters. I have had this requested by at least 3 windows users over time, particularly the "don't enter the break loop" part.

Of course, GAP will still eventually run out of memory, but most other languages just allocate memory until the OS refuses.


src/sysmem.h
```diff
@@ -59,10 +61,20 @@ extern SyStorEnum SyStorOverrun;
...
 **
-**  This is used in the function 'SyAllocBags' below.
+**  This is used in the function 'SyAllocBags' below and can be
+**  disabled using SyGasmanDisableMemoryLimits.
 */
 extern Int SyStorKill;
 
+/****************************************************************************
+**
+*F  SyGasmanDisableMemoryLimits() . . . . .  Disable limits on Workspace size
```
fingolfin, line 71:
```suggestion
*F  SyGasmanDisableMemoryLimits() . . . . .  Disable limits on workspace size
```

src/sysmem.h
```diff
@@ -59,10 +61,20 @@ extern SyStorEnum SyStorOverrun;
...
+**
+**  By default GASMAN limits the size of the workspace using SyStorMax and
+**  SyStorKIll. This disables those checks.
```
fingolfin, line 74:
```suggestion
**  SyStorKill. This disables those checks.
```

src/gap.c
```diff
@@ -815,11 +816,17 @@ static Obj FuncGASMAN(Obj self, Obj args)
             SyMsgsFlagBags = (SyMsgsFlagBags + 1) % 3;
         }
 
+        /* or disable workspace size limits                                */
+        else if (streq(CONST_CSTR_STRING(cmd), "disablelimit")) {
+            SyGasmanDisableMemoryLimits();
```
fingolfin, line 821: Personal opinion (not a blocker for this PR, just a remark): I am not very fond of adding more subcommands to `GASMAN`. Why not just add this as a separate function, say `DISABLE_MEMORY_LIMITS`? On the long run, I'd prefer to remove more GASMAN subcommands, too (or at least add alternatives, and only keep the subcommands for backwards compatibility. E.g. instead of the `collect`/`partial` subcommands, one should use the `CollectGarbage` function; and perhaps we should just turn that into a kernel function (so that it doesn't go through `GASMAN`, as it does right now).

lib/gasman.gi
```diff
@@ -27,6 +27,16 @@ InstallGlobalFunction( CollectGarbage, function( full )
     fi;
 end );
 
+#############################################################################
+##
+#F  GasmanDisableMemoryLimits( )
+##
+##  Disable limits on GASMAN workspace size.
+##
+InstallGlobalFunction("GasmanDisableMemoryLimits", function()
```
fingolfin, line 36: I think Boehm GC also honors `SyStorKill`, and in the future, perhaps also the Julia GC might. So how about calling this just `DisableMemoryLimits`, and talk about "workspace size", not "GASMAN workspace size"?

lib/gasman.gd
```diff
@@ -49,6 +49,25 @@
...
+##
+##  <#GAPDoc Label="GasmanDisableMemoryLimits">
+##  <ManSection>
+##  <Func Name="GasmanDisableMemoryLimits" Arg=''/>
+##  <Returns>
+##  nothing.
+##  </Returns>
+##  <Description>
+##  Disables limits on the size of GAP's workspace. These are usually
+##  set using the --maxworkspace and --limitworkspace command line options.
```
fingolfin, line 64: Wait, I didn't know about `--limitworkspace` ?!? Grepping for it reveals one hit in `lib/system.g`, where it is an alias for `-K` . But in `system.c`, we set `--maximal-workspace` as alias for `-K`. That seems to be a bug, dating back to commit 2b115dc?

fingolfin (changes requested):

lib/gasman.gd
ChrisJefferson: Will check this

lib/gasman.gi
ChrisJefferson: I wasn't positive what other GCs might do, I'm happy to make this a global option

lib/system.g
```diff
@@ -67,11 +67,11 @@ BIND_GLOBAL( "GAPInfo", rec(
...
       rec( short:= "g", long := "gasinfo", default := 0,
            help := ["show GASMAN messages (full/all/no garbage","collections)", "(only available if GAP uses GASMAN)"] ),
-      rec( short:= "m", long := "minworkspace", default := "128m", arg := "<mem>",
+      rec( short:= "m", long := "initial-workspace", default := "128m", arg := "<mem>",
            help := ["set the initial workspace size"] ),
-      rec( short:= "o", long := "maxworkspace", default := "2g", arg := "<mem>",
+      rec( short:= "o", long := "warn-workspace", default := "2g", arg := "<mem>",
            help := [ "set workspace size where GAP will warn about", "excessive memory usage (GAP may allocate more)", "(available only if GAP uses GASMAN)"] ),
-      rec( short:= "K", long := "limitworkspace", default := "0", arg := "<mem>",
+      rec( short:= "K", long := "maximal-workspace", default := "0", arg := "<mem>",
```
fingolfin, line 74: OK, thanks for fixing this.

Orthogonal to this PR: perhaps we should document those long names in the manual, too (see <https://www.gap-system.org/Manuals/doc/ref/chap3.html#X782751D5858A6EAF>) ? Perhaps we should file an issue to remind us about it... tomorrow...

doc/ref/debug.xml
```diff
@@ -516,6 +516,18 @@ in microseconds.
...
   </ManSection>
 
+  <ManSection>
+  <Func Name="DisableMemoryLimits" Arg=''/>
+  <Returns>
+  nothing.
+  </Returns>
+  <Description>
+  Disables limits on the size of GAP's workspace. These are usually
+  set using the --warn-workspace and --maximal-workspace command line options.
```
fingolfin, line 526:
```suggestion
  set using the <C>--warn-workspace<C> and </C>--maximal-workspace</C> command line options.
```
Since we don't document these in the manual right now, perhaps it would be better to refer to them instead or additionally by their short names, which some users may be more familiar with anyhow?

src/system.c
```diff
@@ -653,15 +653,15 @@ static const struct optInfo options[] = {
...
+  { 'm', "initial-workspace", storeMemory2, &SyStorMin, 1 }, /* kernel */
 #endif
   { 'r', "", toggle, &IgnoreGapRC, 0 }, /* kernel */
 #ifdef USE_GASMAN
   { 's', "", storeMemory, &SyAllocPool, 1 }, /* kernel */
 #endif
   { 'n', "", forceLineEditing, 0, 0}, /* prob library */
 #ifdef USE_GASMAN
-  { 'o', "", storeMemory2, &SyStorMax, 1 }, /* library with new interface */
+  { 'o', "warn-workspace", storeMemory2, &SyStorMax, 1 }, /* library with new interface */
```
fingolfin, line 664: off topic: looking at this, I really wonder what the sort order of this struct is? In parts it looks as if it is sorted alphabetically by the short name, upper case before lower case; but that's not really true, as e.g. `E` comes between `f` and `l`... Huh.

Seems to be me we should either sort like that, or else "group logically" / "group as in system,g" ?

Again: this is not a change request for this PR, I am just wondering out loud

src/gap.c
```diff
@@ -897,6 +897,32 @@ static Obj FuncTotalMemoryAllocated(Obj self)
     return ObjInt_UInt8(SizeAllBags);
 }
 
+/****************************************************************************
+**
+*F  SyDisableMemoryLimits() . . . . .  Disable limits on workspace size
```
fingolfin, line 902:
```suggestion
*F  SyDisableMemoryLimits() . . . . . . . . .  Disable limits on workspace size
```

src/gap.c
```diff
@@ -897,6 +897,32 @@ static Obj FuncTotalMemoryAllocated(Obj self)
...
+#endif
+
+#if defined(USE_GASMAN) || defined(USE_BOEHM_GC)
+    SyStorKill = max_mem;
+#endif
+
+#if defined(USE_GASMAN)
+    SyStorMax = max_mem;
+#endif
+
```
fingolfin, line 922:
```suggestion
#if defined(USE_GASMAN) || defined(USE_BOEHM_GC)

#ifdef SYS_IS_64_BIT
    int64_t max_mem = INT64_MAX;
#else
    int32_t max_mem = INT32_MAX;
#endif

    SyStorKill = max_mem;

#if defined(USE_GASMAN)
    SyStorMax = max_mem;
#endif

#endif // defined(USE_GASMAN) || defined(USE_BOEHM_GC)

```
Otherwise we'll get compiler warnings about `max_mem` being unused when building for Julia (I think)

fingolfin:
@ChrisJefferson any plans to revise this to be ready for merging?

ChrisJefferson:
will do, once it gets up my list to the top again :)

fingolfin:
@ChrisJefferson this has been sitting without update for 3 years. Any plans to finish it? Or should we just close it for now (you can always resubmit in the future)

PR #4218, opened 2021-01-21
Julia: fix usage in libgap mode; more GC safety

fingolfin:
In order to be able to safely call GAP_InitJuliaMemoryInterface *before*
InitBags, move code which relies on IsUsingLibGap (which may not be set up
yet) from the former to the latter.

Also make a paranoia change to the foreign types, and store them immediately
into the module, as hypothetically, allocating one of the later types might
cause a GC which could then prematurely collect one of the types we just
created but did not yet store. (In practice, this should not matter, as Julia
apparently does not GC types right now, but better safe than sorry.)

Clearly it would be better if we also tested the Julia integration in libgap mode.
I will look into that eventually, but for now that's only a problem for me, so....

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -934,12 +934,6 @@ void GAP_InitJuliaMemoryInterface(jl_module_t *   module,
     jl_gc_set_cb_post_gc(PostGCHook, 1);
     // jl_gc_enable(0); /// DEBUGGING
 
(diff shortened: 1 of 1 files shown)


fingolfin:
Backported to stable-4.11 in commit 42b0b58709ccb7272de7ad467ecd6f05f570c817

PR #4219, opened 2021-01-21
Performance additions to generic 2-cohomology and Automorphism group/Isomorphism test

### LINKED ISSUE #4240: conjugate a subgroup into a given one
(The following holds for the master branch as well as for released GAP versions.)

Consider the situation that we are given a group G, say, and two subgroups A and B,
and that we want to know an element g in G such that B^g is contained in A.
The GAP function `ContainedConjugates` can be used for that task, but it computes all solutions up to conjugacy, which is more than what we need.
The name of the (undocumented) function `DoConjugateInto` sounds promising, and one can call this function in a way such that one solution is returned once it has been found.
However, `ContainedConjugates` performs better than `DoConjugateInto` in a few quick experiments.
Is there a reason why `ContainedConjugates` does not admit a shortcut (via a global option, say) in order to return just one solution?

hulpke:
Changes for Release notes:

- `IntermediateGroup` might have returned `fail` if the index was large but subgroups exist  (instead of issuing an error message or warning about a hard calculation)

Changes that do not require release notes:

- Speed up collection process for 2-cohomology
- If factor permrep is large, do first try for stabilizer
- when computing permdegree, avoid the bold first attempt to be too bad.

- Better composition series choice in `SubgroupConditionAbove` -- move known parts on bottom.
- Introduces `CompositionSeriesThrough` as a tool
- Improvements in normalizers in `SubgroupConditionAbove`. Use NormalizerViaRadical if there is huge radical bit.
- Delay bottom permrep, if bounded orbit algorithms will do the job.
- Allow forcing old isomorphism test
- Redo generators when searching for trick relators
- Require to stabilize sets of normal subgroups (may reduce in factor)
- Also added option for `ContainedConjugates` (does not merit mentioning in release notes)

Also Typo fix/remove outdated comment in gpfpiso

Resolves #4240.

### DIFF
--- a/lib/autsr.gi
+++ b/lib/autsr.gi
@@ -85,7 +85,7 @@ end;
 # case the values of the relators on pre-images in G do not depend on choice
 # of representatives and can be used to deduce the module automorphism
 # belonging to a factor group automorphism.
-BindGlobal("AGSRFindRels",function(nat)
+BindGlobal("AGSRFindRels",function(nat,newgens)
 local C,M,p,all,gens,sub,q,hom,fp,rels,new,pre,sel,i,free,cnt;
   M:=KernelOfMultiplicativeGeneralMapping(nat);
   C:=Centralizer(Source(nat),M);
@@ -94,13 +94,24 @@ local C,M,p,all,gens,sub,q,hom,fp,rels,new,pre,sel,i,free,cnt;
   fi;
   p:=SmallestPrimeDivisor(Size(M));
   all:=[];
-  gens:=SmallGeneratingSet(Image(nat));
+  if newgens=true then
+    # so generators new
+    sub:=TrivialSubgroup(Image(nat));
+    while Size(sub)<Size(Image(nat)) do
+      sub:=ClosureGroup(sub,ImagesRepresentative(nat,Random(C)));
+    od;
+  else
+    sub:=Image(nat,C);
+  fi;
+
+  gens:=SmallGeneratingSet(sub);
+
   free:=FreeGroup(Length(gens));
   sub:=TrivialSubgroup(M);
   cnt:=0;
   while Size(sub)<Size(M) do
     q:=Group(gens);
-    SetSize(q,Size(Image(nat)));
+    SetSize(q,Size(Image(nat,C)));
     # use `ByGenerators` to force more random relators
     hom:=IsomorphismFpGroupByGenerators(q,gens);
     fp:=Range(hom);
@@ -144,13 +155,14 @@ local ocr,fphom,fpg,free,len,dim,tmp,L0,S,R,rels,mat,r,RS,i,g,v,cnt;
       # Finding the relations comes at a cost that needs to be plausible
       # with searching multiple times through the automorphism group. This
       # order bound is a heuristic that seems to be OK by magnitude.
-      321
+      100
     then
     cnt:=0;
     repeat
-      ocr.trickrels:=AGSRFindRels(nat);
+      ocr.trickrels:=AGSRFindRels(nat,cnt>3);
       cnt:=cnt+1;
-    until ocr.trickrels<>fail or 2^cnt>10*Size(ocr.moduleauts);
+    until ocr.trickrels<>fail or 2^cnt>100*Size(ocr.moduleauts);
+  if ocr.trickrels=fail then Info(InfoMorph,1,"trickrels fails");fi;
   else
     ocr.trickrels:=fail;
   fi;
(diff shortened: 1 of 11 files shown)


lib/autsr.gi
```diff
@@ -491,30 +539,148 @@ local S,c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort,clim,worked,pp,
...
+
+  if nr>=0 then # succeeded (small index case)
+    return S;
+  elif Length(u)>0 then
+    S:=ClosureGroup(Sorig,u);
+  else
+    S:=Sorig;
+  fi;
+
+  # now build along coposition series
```
wilfwilson, line 578: Obviously not important but you may as well fix if you're touching this file again...
```suggestion
  # now build along composition series
```

lib/autsr.gi
```diff
@@ -491,30 +539,148 @@ local S,c,hom,q,a,b,i,t,int,bad,have,ups,up,new,u,good,abort,clim,worked,pp,
...
+
+  rad:=RadicalGroup(G);
+  # composition series through perfect residuum seems to work better
+  # Possible reason: Simple composition factors arise as inner
+  # automorphisms. Moving the perfect residuum (if smaller) in the series,
+  # thus increases the chance that much of the subgroup is found early (and
+  # does not need to be found through search through complements
```
wilfwilson, line 585: ...and here...
```suggestion
  # does not need to be found through search through complements)
```

wilfwilson (approved):
Everything seems alright to me, nice. It would be good if you could provide test cases for the brand new functionality, i.e. the 4-argument `ContainedConjugates` and the `CompositionSeriesThrough`. Otherwise I made two incredibly minor comments on your comments... 😬

tst/teststandard/grpprmcs.tst
```diff
@@ -312,6 +312,10 @@ Group
 Group
 gap> List( ChiefSeriesOfGroup( g ), Size );
 [ 1215506, 607753, 31987, 1103, 1 ]
+gap> u:=NormalClosure(g,SylowSubgroup(g,2));;
+gap> cs:=ChiefSeriesThrough(g,[u]);;
+gap> List(cs,Size);
+[ 1215506, 63974, 2206, 1103, 1 ]
```
wilfwilson, line 318: Thanks @hulpke. Sorry, I didn't notice that your new four-argument `ContainedConjugates` was already tested by the manual example you added. 🙂 

As for this test, since it's `CompositionSeriesThrough` rather than `ChiefSeriesThrough` that is added by this pull request, would you be happy to accept this code suggestion? Unsurprisingly the series are identical, but from my quick look at things, only the latter uses your new `CompositionSeriesThrough` code. Please correct me if I'm mistaken.
```suggestion
[ 1215506, 63974, 2206, 1103, 1 ]
gap> cs:=CompositionSeriesThrough(g,[u]);;
gap> List(cs,Size);
[ 1215506, 63974, 2206, 1103, 1 ]
```

hulpke: Thank you, @wilfwilson . Indeed it should have been `CompositionSeriesThrough` and I mistyped and did not spot it (since it returned a composition series). Will add this.

lib/grp.gi
```diff
@@ -939,6 +939,64 @@ InstallMethod( CompositionSeries,
...
+local cs,i,j,pre,post,c,new,rev;
+  cs:=CompositionSeries(G);
+  # find normal subgroups not yet in
+  normals:=Filtered(normals,x->not x in cs);
+  # do we satisfy by sheer dumb luck?
+  if Length(normals)=0 then return cs;fi;
+
+  SortBy(normals,x->-Size(x));
+  # check that this is a valid series
+  Assert(0,ForAll([2..Length(normals)],i->IsSubset(normals[i-1],normals[i])));
```
fingolfin, line 954: Just out of curiosity, is there a specific reason to use `IsSubset` instead of `IsSubgroup` here?

lib/grp.gi
```diff
@@ -939,6 +939,64 @@ InstallMethod( CompositionSeries,
...
+      # so from pre to post-1 needs to be changed
+      new:=cs{[1..pre-1]};
+
+      rev:=[j];
+      i:=post-1;
+      repeat
+        if not IsSubset(rev[Length(rev)],cs[i]) then
+          c:=ClosureGroup(cs[i],j);
+          if Size(c)>Size(rev[Length(rev)]) then
+            # proper down step
```
fingolfin, line 974: Why is adding a larger group a "down step"?

lib/grp.gi
```diff
@@ -939,6 +939,64 @@ InstallMethod( CompositionSeries,
...
+  SortBy(normals,x->-Size(x));
+  # check that this is a valid series
+  Assert(0,ForAll([2..Length(normals)],i->IsSubset(normals[i-1],normals[i])));
+
+  # Now move series through normals by closure/intersection
+  for j in normals do
+    # first in cs that does not contain j
+    pre:=PositionProperty(cs,x->not IsSubset(x,j));
+    # first contained in j. 
+    post:=PositionProperty(cs,x->Size(j)>=Size(x) and IsSubset(j,x));
```
fingolfin, line 961: Minor optimization possible?
```suggestion
    post:=PositionProperty(cs,x->(Size(j) mod Size(x) = 0) and IsSubset(j,x));
```

lib/grp.gi
fingolfin: Perhaps there should also be a second assertion (perhaps run only at a higher assertion level) which verifies that `normals` truly consists of normal subgroups; and/or add a warning to the documentation that passing non-normal subgroups may result in wrong output?

lib/grplatt.gd
```diff
@@ -449,21 +449,29 @@ DeclareGlobalFunction("LowLayerSubgroups");
...
 ##  <ManSection>
-##  <Oper Name="ContainedConjugates" Arg='G, A, B'/>
+##  <Oper Name="ContainedConjugates" Arg='G, A, B [,onlyone]'/>
 ##
 ##  <Description>
 ##  For <M>A,B \leq G</M> this operation returns representatives of the <A>A</A>
 ##  conjugacy classes of subgroups that are conjugate to <A>B</A> under <A>G</A>.
 ##  The function returns a list of pairs of subgroup and conjugating element.
+##  If the optional fourth argument <A>onlyone</A> is given as <A>true</A>,
+##  then only one pair (or <A>fail</A> if none exists) is returned.
```
fingolfin, line 463:
```suggestion
##  If the optional fourth argument <A>onlyone</A> is given as <K>true</K>,
##  then only one pair (or <K>fail</K> if none exists) is returned.
```

lib/grplatt.gi
```diff
@@ -3187,21 +3187,22 @@ local act,offset,G,lim,cond,dosub,all,m,i,j,new,old;
...
-##
-InstallMethod(ContainedConjugates,"finite groups",IsFamFamFam,[IsGroup,IsGroup,IsGroup],0,
-function(G,A,B)
-local l,N,dc,gens,i;
+DoContainedConjugates:=function(arg)
+local G,A,B,onlyone,l,N,dc,gens,i;
+  G:=arg[1];
+  A:=arg[2];
+  B:=arg[3];
+  if Length(arg)>3 then onlyone:=arg[4]; else onlyone:=false;fi;
```
fingolfin, line 3195: Could also do this:
```suggestion
DoContainedConjugates:=function(G,A,B,onlyone...)
local l,N,dc,gens,i;
  if Length(onlyone)>0 then onlyone:=onlyone[1]; else onlyone:=false;fi;
```

fingolfin: Or this
```suggestion
DoContainedConjugates:=function(G,A,B,onlyone)
local l,N,dc,gens,i;
```
and then change one InstallMethod below to use the following function instead: `{G,A,B}->DoContainedConjugates(G,A,B,false)`

fingolfin:
Argh, terrible timing :-)

I just was about to submit my review when you merged :-(.

Sorry, i was extremely busy the past couple weeks and missed this one.

hulpke:
@fingolfin 
Sorry -- if you know how to unmerge, I will go through your suggestions.

fingolfin:
No, you can't "unmerge", and I think it's fine. I am sorry for being late to the game; and there was nothing terribly important. If you think any of my comments deserve an adjustment, you can just open a new PR. Or ignore them

PR #4220, opened 2021-01-21
Performance improvements motivated by hybrid quotient

hulpke:
- Speed up Exponents wrt a canonical PCGS, as these are just a subset of the exponents of the parent pcgs. Thus no need to divide off.
- Use bit lists to speed up subword testing in checking confluence of rewriting system

These commits are combined because of common applications and common impact on some manual examples


fingolfin (approved):
Looks good to me. The test failure is unrelated, see PR #4222

PR #4222, opened 2021-01-22
Always define DeclareObsoleteSynonym{Attr,}

fingolfin:
... even when the `-O` command line option is used. This way,
packages can safely use it.

Should fix failing `testmanuals` CI tests which use `-O`, and which now fail due to the new smallgrp package release which indeed uses `DeclareObsoleteSynonym`.

Should be backported to stable-4.11 to fix the same CI failure there.


ChrisJefferson (approved):

fingolfin:

PR #4224, opened 2021-01-22
Do not name functions with automatically generated names

### LINKED ISSUE #4223: Functions named "LastReadValue"
### Observed behaviour

There is an undocumented global variable `LastReadValue` that is assigned the result of every call of the internal function `READ()`.  If that result happens to be a function, the function's name will become `LastReadValue`.

Example:

```
gap> stream := InputTextString("function() end;");
InputTextString(0,15)
gap> READ_STREAM(stream);
true
gap> LastReadValue;
function(  ) ... end
gap> NameFunction(LastReadValue);
"LastReadValue"
```

### Expected behaviour

The function's name should not be assigned to in this case.  Perhaps there should be away to assign to a global variable *without* invoking the behavior of setting a function's name.

It's mostly harmless, but confusing.

```
 ┌───────┐   GAP 4.11.0 of today
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv7
```

ChrisJefferson:
Functions are usually named using the first variable they are
assigned to. Disable this for automatically assigned variables
like Last and LastReadValue

Added some brief documentation to some internal functions at the same time.

Fixes #4223

### DIFF
--- a/src/gvars.c
+++ b/src/gvars.c
@@ -387,7 +387,15 @@ static Obj * ELM_COPS_PLIST(Obj cops, UInt i)
     return (Obj *)val;
 }
 
-static void AssGVarInternal(UInt gvar, Obj val, Int hasExprCopiesFopies)
+// Assign 'val' to name 'gvar'. 'hasExprCopiesFopies' can be false if this
+// variable has a non-default value assigned to ExprGVars, CopiesGVars or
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Looks good to me

wilfwilson (approved):

PR #4225, opened 2021-01-23
Build 32-bit GAP on Ubuntu 18.04, as latest is missing many 32-bit packages

ChrisJefferson:
Fixes the current failures in github CI.

Basically, building all packages on 32-bit GAP in Ubuntu 20.04 isn't possible, as many 32-bit versions of packages have been removed.

Let's see how long 18.04 hangs around for in github actions before we decide how (or if) to make this work in 20.04

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,7 +56,7 @@ jobs:
           # compile packages and run GAP tests in 32 bit mode
           # it seems profiling is having trouble collecting the coverage data
           # here, so we use NO_COVERAGE=1
-          - os: ubuntu-latest
+          - os: ubuntu-18.04
             test-suites: "testpackages testinstall-loadall"
             extra: "ABI=32 NO_COVERAGE=1"
             packages: "
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,15 +13,15 @@ jobs:
       fail-fast: false
       matrix:
         # base test: fast first test
-        os: [ubuntu-latest]
+        os: [ubuntu-18.04]
         test-suites: ["testinstall"]
 
         # add a few extra tests
         include:
-          - os: ubuntu-latest
+          - os: ubuntu-18.04
             test-suites: "docomp teststandard"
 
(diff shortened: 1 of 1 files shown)


wilfwilson:
Have you actually had any failing jobs yet because of this? I’m running the CI of the current `master` on my fork and it’s still using 18.04 (and seemingly it’s giong to pass) https://github.com/wilfwilson/gap/actions/runs/509003806

I don't mind using 18.04 in preparation for when 20.04 is getting used everywhere, but it'd be good to actually see it fail first :)

ChrisJefferson:
If you look at: https://github.com/ChrisJefferson/gap/runs/1754323341?check_suite_focus=true

You will see the 'install dependancies' step fails with a bunch of i386 package failures. However, I thought I saw these on master too and now can't see them.

If you look here https://packages.ubuntu.com/bionic/singular you will see bionic (18.04) has i386 singular, while https://packages.ubuntu.com/focal/singular (20.04) doesn't have i386 singular.

18.04 does have support until 2023 (don't know how long github actions will support it of course)

wilfwilson (approved):
Thanks for that, and also for noticing the problem and fixing it.

.github/workflows/CI.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           # compile packages and run GAP tests in 32 bit mode
           # it seems profiling is having trouble collecting the coverage data
           # here, so we use NO_COVERAGE=1
-          - os: ubuntu-latest
+          - os: ubuntu-18.04
```
fingolfin, line 59: Actually perhaps we should run most  tests on 18.04 or an even old version (if available)? In the past I saw more issues where we broke things for old Linux versions (by using things "too new") than on new ones.

fingolfin (approved):
Looks good to me but also see my comment/question

ChrisJefferson:
I switched everything but one (picked fairly at random I'll admit) to ubuntu-18.04, left testmanuals on latest, just to make sure we test there, without adding another test.

fingolfin (approved):

PR #4228, opened 2021-01-25
Minor grammar changes to documentation and comments

wilfwilson:
There are four separate commits describing the types of changes that I've made. Three of the commits fix mistakes and should be totally non-controversial.

The fourth commit standardises on 'cannot' rather than 'can not'; although 'can not' is not incorrect English, it's far less common in both general English, and in GAP (it occurred around 20 times less frequently).

### DIFF
--- a/lib/algrep.gi
+++ b/lib/algrep.gi
@@ -1254,8 +1254,8 @@ end );
 
 #############################################################################
 ##
-#M  \/( <V>, <W> )  . . . . . . .factor of a an algebra module by a submodule
-#M  \/( <V>, <vectors> )  . . . .factor of a an algebra module by a submodule
+#M  \/( <V>, <W> )  . . . . . . .  factor of an algebra module by a submodule
+#M  \/( <V>, <vectors> )  . . . .  factor of an algebra module by a submodule
 ##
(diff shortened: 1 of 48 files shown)


lib/algrep.gi
```diff
@@ -1254,8 +1254,8 @@ end );
 
 #############################################################################
 ##
-#M  \/( <V>, <W> )  . . . . . . .factor of a an algebra module by a submodule
-#M  \/( <V>, <vectors> )  . . . .factor of a an algebra module by a submodule
+#M  \/( <V>, <W> )  . . . . . . . factor of an algebra module by a submodule
+#M  \/( <V>, <vectors> )  . . . . factor of an algebra module by a submodule
```
ThomasBreuer, line 1258: The text behind the dots could be moved one more space to the right.

ThomasBreuer (approved):
This looks good to me, I have added just one comment.
(I do not see a way how such problems could be found automatically, apart from heuristic checks as in the case of "can not".)

lib/algrep.gi
wilfwilson: Done, thanks.

PR #4230, opened 2021-01-26
fixed `HELP_SEARCH_ALTERNATIVES` again

### LINKED ISSUE #4229: On-line help fails given a query ending "-ize"
Using the latest GAP master, I've found that the online help sometimes fails in the following way:

Here the first query succeeds but the second query fails to find an appropriate match:
```
gap> ?Pluraliz
Help: Showing `Reference: Pluralize'
gap> ?Pluralize
Help: no matching entry found
Help: 'Pluralize' is currently undocumented.
      For details, try ?Undocumented Variables
```
And here the first and third queries succeed, but the second fails to find anything:
```
gap> ?Stabiliz
gap> ?Stabilize
Help: no matching entry found
gap> ?Stabilizer
```
and the same here:
```
gap> ?Diagonaliz
gap> ?Diagonalize
Help: no matching entry found
gap> ?DiagonalizeMa
Help: Showing `Reference: DiagonalizeMat'
```
The common denominator seems to be the query ending "-ize".

ThomasBreuer:
Apparently there were two errors, and #4213 had dealt only with one of them.
Now the second one is fixed, and there are also tests for the relevant cases.

(This resolves #4229.)

### DIFF
--- a/lib/helpbase.gi
+++ b/lib/helpbase.gi
@@ -211,7 +211,7 @@ if Length(positions) > 0 then # matches found
   for i in [1..Length(positions)] do
     Add( chop, [ topic{[begin..patterns[i].start-1 ]} ] );
     Add( chop, patterns[i].pattern );
-    begin := Minimum( patterns[i].finish+1, Length(topic) );
+    begin := Minimum( patterns[i].finish, Length(topic) )+1;
   od;
 
   if begin <= Length( topic ) then
--- a/hpcgap/lib/helpbase.gi
+++ b/hpcgap/lib/helpbase.gi
@@ -214,7 +214,7 @@ if Length(positions) > 0 then # matches found
   for i in [1..Length(positions)] do
     Add( chop, [ topic{[begin..patterns[i].start-1 ]} ] );
     Add( chop, patterns[i].pattern );
-    begin := Minimum( patterns[i].finish+1, Length(topic) );
+    begin := Minimum( patterns[i].finish, Length(topic) )+1;
   od;
 
   if begin <= Length( topic ) then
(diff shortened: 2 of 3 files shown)


wilfwilson (approved):
Thanks for doing this so quickly. I can confirm that the failures that I mentioned in #4229 are all fixed by this PR.

lib/helpbase.gi
```diff
@@ -211,7 +211,7 @@ if Length(positions) > 0 then # matches found
   for i in [1..Length(positions)] do
     Add( chop, [ topic{[begin..patterns[i].start-1 ]} ] );
     Add( chop, patterns[i].pattern );
-    begin := Minimum( patterns[i].finish+1, Length(topic) );
+    begin := Minimum( patterns[i].finish, Length(topic) )+1;
```
wilfwilson, line 214: Oops, as with #4213, the same change should be made to `hpcgap/lib/helpbase.gi`, please.

wilfwilson (changes requested):

wilfwilson (approved):
Thank you

fingolfin:
Just to check: does this need to be backported?

ThomasBreuer:
@fingolfin The only important point is that the pull requests #4213 and #4230 belong together --either both can be backported or none of them. I do not think that it is important to backport them.

wilfwilson:
Note that, on a quick check, it seems that the bug fixed by #4213 *does* exist in `stable-4.11`, while the bug fixed by this PR *does not* exist in `stable-4.11`:
```
gap> HELP_SEARCH_ALTERNATIVES( "Centralizer" );
[ "Centralise", "Centralize" ]
```
but this works correctly:
```
gap> ?Stabilize
```

PR #4231, opened 2021-01-27
LoadKernelExtension, IsKernelExtensionAvailable

fingolfin:
The idea is to provide "better" (?) APIs for packages to load kernel extensions. With `IsKernelExtensionAvailable` (and an upcoming kernel function to make it better), packages can simplify their `TestAvailability` functions, and at the same time make them more robust. Likewise, `LoadKernelExtension` simplifies `init.g` code.

The big win is that it'll then make it much easier for us to change how to locate and load kernel extensions; e.g. to better support `make install` of packages, or also better Julia integration.

Open questions:
1. perhaps I should rename these functions to `LoadKernelModule` and `IsKernelModuleAvailable`? Background: in the manual and various other places, we talk about "kernel modules". I (and some other places/people) use "kernel extension" (possibly a short hand for "kernel extension module") to distinguish it from built-in kernel modules; I guess one could also refer to it as "external kernel module" (but that would not be quite right for "static" kernel extension modules...).
2. In case of failure, perhaps `IsKernelExtensionAvailable` resp. `IS_LOADABLE_DYN` should return proper error codes (instead of just `false`) to indicate what exactly went wrong (file missing / file not a loadable module / file not a GAP kernel extension / kernel extension but built for wrong GAP version / ....). It could even be just a string; in any case, something that helps produce better error messages for `AvailabilityTest` in `PackageInfo.g`
3. Perhaps add an optional field `KernelModule` to the package info record, which when present would prompt GAP to automatically invoke `IsKernelExtensionAvailable` when checking the package availability; and `LoadKernelExtension` when loading the package (possible values: `KernelModule  := true` = use package name, `KernelModule := "modname"`; or even `KernelModule := ["mod1", "mod2"]` if someone wants to support multiple kernel modules in one package. Alas, that seems very fringe, so I'd probably not do it)

### DIFF
--- a/lib/files.gd
+++ b/lib/files.gd
@@ -566,12 +566,24 @@ end );
 ##
 #F  LoadDynamicModule( <filename> ) . . . . . .  try to load a dynamic module
 ##
+##  <#GAPDoc Label="LoadDynamicModule">
 ##  <ManSection>
 ##  <Func Name="LoadDynamicModule" Arg='filename'/>
 ##
 ##  <Description>
+##  To load a compiled file, the command <Ref Func="LoadDynamicModule"/> is
+##  used. This command loads <A>filename</A> as module.
+##  <P/>
+##  <Log><![CDATA[
+##  gap> LoadDynamicModule("./test.so");
+##  ]]></Log>
+##  <P/>
+##  On some operating systems, once you have loaded a dynamic module with a
+##  certain filename, loading another with the same filename will have no
+##  effect, even if the file on disk has changed.
 ##  </Description>
 ##  </ManSection>
+##  <#/GAPDoc>
 ##
 BIND_GLOBAL( "LoadDynamicModule", function( filename )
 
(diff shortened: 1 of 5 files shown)


lib/files.gd
```diff
@@ -626,6 +626,63 @@ BIND_GLOBAL( "LoadStaticModule", function( arg )
...
+        modname := modname[1];
+    else
+        Error( "usage: LoadPackageKernelExtension( <pkgname> [, <modname> ] )" );
+    fi;
+
+    if modname in SHOW_STAT() then
+        LoadStaticModule(modname);
+        return true;
+    fi;
+    fname := Filename(DirectoriesPackagePrograms(pkgname), modname);
```
ThomasBreuer, line 677: `Filename(DirectoriesPackagePrograms(pkgname), Concatenation(modname, ".so"))`?

lib/files.gd
```diff
@@ -626,6 +626,63 @@ BIND_GLOBAL( "LoadStaticModule", function( arg )
...
+    elif Length(modname) = 1 then
+        modname := modname[1];
+    else
+        Error( "usage: IsPackageKernelExtensionAvailable( <pkgname> [, <modname> ] )" );
+    fi;
+
+    if modname in SHOW_STAT() then
+        return true;
+    fi;
+    fname := Filename(DirectoriesPackagePrograms(pkgname), modname);
```
ThomasBreuer, line 647: `Filename(DirectoriesPackagePrograms(pkgname), Concatenation(modname, ".so"))`?

lib/package.gi
```diff
@@ -1210,6 +1210,18 @@ InstallGlobalFunction( DirectoriesPackagePrograms, function( name )
...
+# so that the precise combination of GAP and package are matched...
+# That is, take the precise GAP version (possibly including the git commit
+# it was compiled from), plus the package version (and possibly also the locations of both?)
+# and combine them into a unique ID (e.g. via a hash), then use that to
+# store .so binaries
+
+
+# also add a new mechanism to integrate kernel extensions into a package:
+# let the package declare in its PackageInfo.g that it has a kernel extension;
+# how to build it; what's its name; then GAP can take care of that dynamically
```
ThomasBreuer, line 1224: Can the intended new entries in `PackageInfo.g` also replace the `prerequisites.sh` files used by`BuildPackages.sh`?

ThomasBreuer (commented):
Nice idea. See my comments.

lib/files.gd
fingolfin: Indeed
```suggestion
    fname := Filename(DirectoriesPackagePrograms(pkgname), Concatenation(modname, ".so"));
```

lib/files.gd
fingolfin:
```suggestion
    fname := Filename(DirectoriesPackagePrograms(pkgname), Concatenation(modname, ".so"));
```

lib/package.gi
fingolfin: Hopefully / perhaps -- it depends a bit on what the package does in there...

fingolfin:
I've overhauled this now, and IMHO it is mostly ready, except for the lack of documentation.

Going beyond this, I wonder if we should add an optional field `KernelExtension` to the package info record. If that is present, then GAP could automatically invoke `IsKernelExtensionAvailable` when checking the package availability; and `LoadKernelExtension` when loading it. That said, perhaps this is too fringe?

ChrisJefferson:
I generally like this PR. I have one (larger) comment -- I wonder if we should separate static and dynamic modules more?

Is there any evidence we can still (or demand for?) loading extensions as GAP static modules would work? The built-in ones (where we compile some library files with gac) are not handled in the "standard way", READ_GAP_ROOT calls LookupStaticModule. Therefore we could, perhaps, simplify this code some of the general loading code by removing mentions of static modules, and (even) rename 'static' to 'library' or something.

Of course, if people want the option in future of doing a big static build, where we static compile extensions, that wouldn't work.

fingolfin:
@ChrisJefferson I am not sure what to make of your comment. It seems orthogonal to the content of this PR? If at all, this PR would help with "getting rid and/or renaming static modules": it removes the need to explicitly call into the APIs for static modules, by providing a fresh abstraction layer. So once we have this PR in, future packages could switch to using `LoadKernelExtension` and `IsKernelExtensionAvailable`. Then the package authors would not have to think about "static modules" anymore, even if they still wanted to support those.

The sooner we have this in, the sooner packages could start using it. Conversely, even if we get this into 4.12, it'll take time for this to be adopted by packages.

fingolfin:
I notice that in the manual and various other places, we talk about "kernel modules". I (and some other places/people) use "kernel extension" (possibly a short hand for "kernel extension module") to distinguish it from built-in kernel modules; I guess one could also refer to it as "external kernel module" (but that would not be quite right for "static" kernel extension modules...).

Aaaanyway: perhaps I should rename these functions then to `LoadKernelModule` and `IsKernelModuleAvailable` ?

fingolfin:
I also wonder if `IsKernelExtensionAvailable` resp. `IS_LOADABLE_DYN` should perhaps produce a proper error code (instead of just returning `false`) to indicate what exactly went wrong (file missing / file not a loadable module / file not a GAP kernel extension / kernel extension but built for wrong GAP version / ....). It could even be just a string; in any case, something that helps produce better error messages for `AvailabilityTest` in `PackageInfo.g`

fingolfin:
Perhaps we should just merge this? There are plenty more things one *could* do, but this is already an improvement...

ChrisJefferson (approved):

PR #4232, opened 2021-01-27
Improve iterators for lists

### LINKED ISSUE #4226: Possible bug in IteratorList
### Observed behaviour
~~~~
gap> tmp := [1, 2, 3, 4];
[ 1, 2, 3, 4 ]
gap> MakeImmutable(tmp);
[ 1, 2, 3, 4 ]
gap> it := IteratorList(tmp);
<iterator>
gap> for x in it do od;
gap> IsDoneIterator(it);
true
gap> NextIterator(it);
  return iter!.list[iter!.pos]; at /Users/jdm/gap/lib/list.gi:763 called from
iter!.NextIterator( iter ) at /Users/jdm/gap/lib/coll.gi:948 called from
<function "NextIterator for `IsIteratorByFunctions'">( <arguments> )
 called from read-eval loop at *stdin*:21
brk> quit;
gap> IsDoneIterator(it);
false
~~~~

### Expected behaviour

The iterator should remain **done** after `NextIterator` is called.

~~~~
 ┌───────┐   GAP 4.12dev-622-gf3e72d2 built on 2021-01-14 19:49:28+0000
 └───────┘   Architecture: x86_64-apple-darwin13.4.0-default64-kv8
 Packages:   AutoDoc 2018.11.19dev, datastructures 0.2.5, Digraphs 1.3.1,
             GAPDoc 1.6.1, genss 1.6.5, GRAPE 4.8.2, images 1.3.0, IO 4.7.0,
             orb 4.8.3, PackageManager 1.1, PrimGrp 3.3.1, Semigroups 3.4.0,
             SmallGrp 1.3, Smallsemi 0.6.12, TransGrp 2.0.4
gap>
~~~~

Fairly sure this is caused by 

~~~~
BindGlobal( "IsDoneIterator_List",
    iter -> ( iter!.pos = iter!.len ) );
~~~~

in line 741 of `list.gi`, which should probably be

~~~~
BindGlobal( "IsDoneIterator_List",
    iter -> ( iter!.pos >= iter!.len ) );
~~~~

or `iter!.pos` shouldn't be incremented after it has reached `Length(iter!.list)`.

wilfwilson:
This resolves #4226, reported by @james-d-mitchell. See the discussion there for further context.

Currently, iterators behave differently for immutable dense lists as opposed to other lists w.r.t. whether `NextIterator` throws an informative error when called on a finished iterator. Moreover, in one of the cases, calling `NextIterator` on a finished iterator unnecessarily broke the value of `IsDoneIterator`.

I tried making the error-checking behaviour consistent, but this led to an unacceptable 10% slowdown. Instead I've just removed the breaking of `IsDoneIterator`, and found one or two ways to speed up the code (mostly by reducing access to record components).

If speed of iterators is important, then I think these changes are useful. If speed is not important, then I will add reinstate to `NextIterator_DenseList`...

I've added some tests, removed some commented-out code, and rearranged a bit of code too - I'm happy to remove either of the latter two for being off-topic.

Quick experiments:
```
# Mutable list
gap> for i in [1..10] do iter := Iterator([1 .. 10000000]);; for i in iter do od; od; time;
20926 # Before this PR
17429 # After this PR

# Immutable list
gap> for i in [1..10] do iter := Iterator(Immutable([1 .. 10000000]));; for i in iter do od; od; time;
13205 # Before this PR
13158 # After this PR
```

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -756,8 +756,6 @@ BindGlobal( "NextIterator_List", function ( iter )
     return l[ p ];
     end );
 
-#BindGlobal( "IsDoneIterator_DenseList",
-#    iter -> not IsBound( iter!.list[ iter!.pos + 1 ] ) );
 
 BindGlobal( "NextIterator_DenseList", function ( iter )
     local p;
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -782,14 +782,13 @@ InstallGlobalFunction( IteratorList, function ( list )
 #T call `Immutable'?
                 pos  := 0,
                 len := Length(list),
+                IsDoneIterator := IsDoneIterator_List,
                 ShallowCopy := ShallowCopy_List );
 
     if IsDenseList( list ) and not IsMutable( list ) then
-      iter.IsDoneIterator := IsDoneIterator_List;
-      iter.NextIterator   := NextIterator_DenseList;
+      iter.NextIterator := NextIterator_DenseList;
     else
-      iter.IsDoneIterator := IsDoneIterator_List;
-      iter.NextIterator   := NextIterator_List;
+      iter.NextIterator := NextIterator_List;
     fi;
 
     return IteratorByFunctions( iter );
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Thank you, this PR is exemplary -- nicely separated clean commits make it a breeze to review. 5/5, would review again ;-)

Issue #4241, opened 2021-02-03
inconsistency concerning `arg`


ThomasBreuer:
The following happens in the current master branch.
The behaviour is perhaps not really a bug, but I think it is inconsistent.
```
gap> f1:= function( arg ) return 0; end;
function( arg... ) ... end
gap> f2:= arg -> 0;
function( arg... ) ... end
```
Up to here, the two functions look equal.
```
gap> Print( f1 );
function ( arg... )
    return 0;
end
gap> Print( f2 );
function ( arg )
    return 0;
end
```
No, here they did not look equal.

And they behave differently: One can call `f1` with several arguments, but `f2` admits exactly one argument.

fingolfin:
That's how it was forever: in the past, GAP only supported `->` for single argument functions, and the special treatment for `arg` was not implemented.

Thanks to the refactoring I did in the last 1-2 years, it should be very easy to add this; although but I wonder if this might break some existing code? Hopefully not, but we should at least grep for occurrences of `arg ->` in packages.

In the meantime, you can write `f2 := {arg...} -> 0` to get the desired result, too.

ThomasBreuer:
Thanks.
I think that the inconsistency would be fixed if the function `f2` above would be `View`ed as `function( arg ) ... end`.
The `arg...` is produced by a `ViewString` method for functions in `lib/function.gi` (and also by a `ViewObj` method in `lib/function.g`), because of the following code.
```
    nams := NAMS_FUNC(func);
    narg := NARG_FUNC(func);
    if narg < 0 then
        isvarg := true;
        narg := -narg;
    fi;
    if narg = 1 and nams <> fail and nams[1] = "arg" then
        isvarg := true;
    fi;
```
This looks like a hack that overrides the return value of `NARG_FUNC` in the case of the argument name `arg`.
Without this special treatment, the inconsistency would disappear.

PR #4248, opened 2021-02-11
changed `ViewObj` and `ViewString` for functions

ThomasBreuer:
I think that a function object knows best whether it takes a variable number of arguments or not, according to the result of `NARG_FUNC`, and independent of the argument name `arg`, see the examples in issue #4241.

(Let us see whether some tests fail after the change.)


fingolfin:
Tests pass; but that also means that no test actually checked the precise situation you modified. Perhaps you can add a test case which does? I.e. using your example in PR #4241?

fingolfin (approved):

PR #4249, opened 2021-02-14
Add ARCH_IS_WSL, to detect Windows Subsystem for Linux

ChrisJefferson:
This lets us detect the "Windows Subsystem for Linux".

I realise in some ways the naming has gotten weird now, we call Cygwin (one linux in windows) "ARCH_IS_WINDOWS", which WSL is not. While that is weird, I think it makes sense, WSL is a complete install of Linux (usually Ubuntu).

The only time we need ARCH_IS_WSL is when we want to open GUI programs -- I've added help viewers (unfortunately it doesn't seem possible to open either anchors in HTML, which makes this much less useful. I spent some time searching the internet, but everything people say seems to not work in practice, except making a new webpage which just forwards to the anchor, and I didn't want to do that as it would create lots of new temp files/directories).

I plan on using this in the 'profiling' package too, to automatically open HTML pages (I originally wrote this in that package, then thought I might be generally useful).

### DIFF
--- a/lib/helpview.gi
+++ b/lib/helpview.gi
@@ -198,48 +198,78 @@ elif ARCH_IS_MAC_OS_X() then
             return;
         end
     );
- 
+
 else # UNIX but not Mac OS X
 
-  # html version with netscape
-  HELP_VIEWER_INFO.netscape := rec(
-  type := "url",
-  show := function(url)
-    Exec(Concatenation("netscape -remote \"openURL(file:", url, ")\""));
-  end
-  );
+  # Graphical systems handled differently in WSL and standard Linux
+  if ARCH_IS_WSL() then
+    HELP_VIEWER_INFO.browser := rec(
+    type := "url",
+    show := function( url )
+      # Ignoring part of the URL after '#' since we are unable
+      # to navigate to the precise location on Windows
+      url := SplitString( url, "#" )[1];
+      Exec(Concatenation("explorer.exe \"$(wslpath -a -w \"",url, "\")\""));
+    end
+    );
 
-  # html version with mozilla
-  HELP_VIEWER_INFO.mozilla := rec(
-  type := "url",
-  show := function(url)
-    Exec(Concatenation("mozilla -remote \"openURL(file:", url, ")\""));
-  end
-  );
+    HELP_VIEWER_INFO.("pdf viewer") := rec(
+        type := "pdf",
+        show := function(file)
+          local   page;
+          # unfortunately one cannot (yet?) give a start page to windows
+          page := 1;
+          if IsRecord(file) then
+            if IsBound(file.page) then
+              page := file.page;
+            fi;
+            file := file.file;
+          fi;
+          Exec(Concatenation("explorer.exe \"$(wslpath -a -w \"",file, "\")\""));
+          Print("#  see page ", page, " in PDF.\n");
+    end
+    );
+  else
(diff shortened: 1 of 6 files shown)


src/sysfiles.c
```diff
@@ -339,6 +340,18 @@ static Obj FuncCrcString(Obj self, Obj str)
     return INTOBJ_INT(((Int4) crc) >> 4);
 }
 
+// Get OS Kernel version. Used to discover if GAP is running inside
+// 'Windows Subystem for Linux'
+Obj SyGetOsRelease(void)
+{
+    struct utsname buf;
+    if (!uname(&buf)) {
```
fingolfin, line 348: If you wrap uname, how about going the full way and return a record with all components? And then just return `fail` if there is an error?

lib/system.g
```diff
@@ -564,6 +564,33 @@ BIND_GLOBAL("ARCH_IS_UNIX",function()
...
+#F  ARCH_IS_WSL()
+##
+##  <#GAPDoc Label="ARCH_IS_WSL">
+##  <ManSection>
+##  <Func Name="ARCH_IS_WSL" Arg=''/>
+##
+##  <Description>
+##  tests whether &GAP; is running on a Windows system inside the
+##  'Windows Subsystem for Linux'. Note that in this case
+##  <Ref Func="ARCH_IS_UNIX"/> will be <C>true</C>, and in most situations
```
wilfwilson, line 578:
```suggestion
##  <Ref Func="ARCH_IS_UNIX"/> will be <K>true</K>, and in most situations
```

lib/helpview.gi
```diff
@@ -198,48 +198,78 @@ elif ARCH_IS_MAC_OS_X() then
...
-    Exec(Concatenation("netscape -remote \"openURL(file:", url, ")\""));
-  end
-  );
+  # Graphical systems handled differently in WSL and standard Linux
+  if ARCH_IS_WSL() then
+    HELP_VIEWER_INFO.browser := rec(
+    type := "url",
+    show := function( url )
+      # Ignoring part of the URL after '#' since we are unable
+      # to navigate to the precise location on Windows
```
wilfwilson, line 210: That's weird, and unfortunate.

wilfwilson (approved):
Nice. I've got no suggestions on the code but I do have a cosmetic nitpick on the doc.

fingolfin:
Just wondering, how is this related to PR #3939?

ChrisJefferson:
This is unrelated to cygwin changes (although in the long term, the plan is the two distros will come closer together. This will be one place where they differ, as they have different ways of spawning files in windows)

src/sysfiles.c
fingolfin: (to be clear, this is not a change required for this PR to be merged. You might want to apply the suggestion by Wilf, though)

fingolfin (approved):

lib/system.g
ChrisJefferson: I'm happy to fix it, although I coped what I wrote from about 20 lines higher up (I'll fix that too)

ChrisJefferson:
Hopefully all cleaned up now. I decided not to fix other <C>true</C>, I thought they could be sweeped (also <C>false</C>) in another PR.

wilfwilson:
Sure, thanks Chris. I'll merge this when the tests pass.

PR #4251, opened 2021-02-15
Initial Cygwin github action

ChrisJefferson:
Add a cygwin github action.

At the moment this is in it's own file, partially because it is huge and horrible. At the moment it doesn't try to upload coverage (I think it could).

### DIFF
--- a/.github/workflows/cygwin-windows.yml
+++ b/.github/workflows/cygwin-windows.yml
@@ -52,9 +52,6 @@ jobs:
         SET CYGROOT=C:\cygwin64
         SET CYGCACHE=%CYGROOT%\var\cache\setup
         setup-x86_64.exe -qnNdO -R %CYGROOT% -l %CYGCACHE% -s %CYGMIRROR% -P wget,git,gcc,gcc-g++,gcc-core,m4,libgmp-devel,make,automake,libtool,autoconf,autoconf2.5,zlib-devel,libreadline-devel
-        SET TARGET_ARCH=x86_64-w64-mingw32
-        SET CD_MFMT=%cd:\=/%
-        SET RESULT_INSTALLDIR_CFMT=%CD_MFMT:C:/=/cygdrive/c/%
       env:
         CYGMIRROR: "http://mirror.easyname.at/cygwin"
       shell: cmd
--- a/.github/workflows/cygwin-windows.yml
+++ b/.github/workflows/cygwin-windows.yml
@@ -29,37 +29,8 @@ jobs:
         git config --global core.eol lf
 
     - uses: actions/checkout@v2
-    - name: Download cygwin
-      run: |
-        & cmd /c 'nslookup www.cygwin.com 2>&1'
-        while ($true)
-        {
-            try
-            {
-                (New-Object Net.WebClient).DownloadFile('http://www.cygwin.com/setup-x86_64.exe', 'setup-x86_64.exe')
-                break
-            }
-            catch
-            {
-                Write-Host "There is an error during package downloading:`n $_"
-            }
-        }
-        & cmd /c 'nslookup www.cygwin.com 8.8.8.8 2>&1'
-      shell: powershell
-    - name: Install cygwin
-      run: |
-        @ECHO ON
-        SET CYGROOT=C:\cygwin64
-        SET CYGCACHE=%CYGROOT%\var\cache\setup
-        setup-x86_64.exe -qnNdO -R %CYGROOT% -l %CYGCACHE% -s %CYGMIRROR% -P wget,git,gcc,gcc-g++,gcc-core,m4,libgmp-devel,make,automake,libtool,autoconf,autoconf2.5,zlib-devel,libreadline-devel
-      env:
-        CYGMIRROR: "http://mirror.easyname.at/cygwin"
-      shell: cmd
 
-    - name: Set up cygwin profile
-      run: |
-        C:\cygwin64\bin\bash -l -c bash
-      shell: cmd
+    - uses: gap-actions/setup-cygwin-for-gap@v1
 
     # CHERE_INVOKING=1 lets us start a 'login shell' (to set paths) without changing directory
     - name: Compile GAP and download packages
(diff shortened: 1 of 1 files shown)


.github/workflows/cygwin-windows.yml
```diff
@@ -0,0 +1,79 @@
...
+    # This sets git to use Linux line endings. While GAP should allow
+    # Windows line endings in GAP files, we currently can't build GAP if the
+    # whole source has Windows line endings
+    - name: Set git to use LF
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
+
+    - uses: actions/checkout@v2
+    - name: Download cygwin
```
fingolfin, line 32: This step feels like something that could/should be in a Github Action of its own, perhaps under the gap-actions org. @ssiccha knows how to do that, in case you are interested, Chris. That would then make it easier for packages to use this, too.

fingolfin:
Runtime seems comparable to AppVeyor, but the log is better, it's in the same place as the other CI reports, *and* it is not limited to 1 concurrent runner... So overall a big win. Nice!

.github/workflows/cygwin-windows.yml
```diff
@@ -0,0 +1,79 @@
...
+        & cmd /c 'nslookup www.cygwin.com 8.8.8.8 2>&1'
+      shell: powershell
+    - name: Install cygwin
+      run: |
+        @ECHO ON
+        SET CYGROOT=C:\cygwin64
+        SET CYGCACHE=%CYGROOT%\var\cache\setup
+        setup-x86_64.exe -qnNdO -R %CYGROOT% -l %CYGCACHE% -s %CYGMIRROR% -P wget,git,gcc,gcc-g++,gcc-core,m4,libgmp-devel,make,automake,libtool,autoconf,autoconf2.5,zlib-devel,libreadline-devel
+        SET TARGET_ARCH=x86_64-w64-mingw32
+        SET CD_MFMT=%cd:\=/%
```
ssiccha, line 56: Out of curiosity, what could the FMT part stand for?

ssiccha (approved):
Cool, LGTM!

.github/workflows/cygwin-windows.yml
ChrisJefferson: that is an excellent question, googling, it looks like a bit I should have tried removing, let me try :)

ssiccha:
There now is a cygwin installation action: https://github.com/gap-actions/setup-cygwin-for-gap
For a version of this PR with that action see
- [The PR](https://github.com/ssiccha/gap/tree/cygwin-CI)
- [workflow result](https://github.com/ssiccha/gap/runs/1910071496?check_suite_focus=true)

fingolfin:
So I guess this PR could be modified to use @ssiccha's action -- @ChrisJefferson you interested in this? Or perhaps @ssiccha can suggest an edit with the required change?

ChrisJefferson:
I’m very happy to use @ssiccha 's action, I’m not sure of the best way to do that, might be easiest to make a new PR with the new action in use? If that passes tests, we can close this and merge that instead?

ssiccha:
@ChrisJefferson @fingolfin I've added a commit to this branch. The tests passed when I ran what is now in this PR on my fork, so they should pass here too. Let's see.

.github/workflows/cygwin-windows.yml
```diff
@@ -0,0 +1,47 @@
...
+      TEST_SUITES: testinstall
+
+    steps:
+    # This sets git to use Linux line endings. While GAP should allow
+    # Windows line endings in GAP files, we currently can't build GAP if the
+    # whole source has Windows line endings
+    - name: Set git to use LF
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
```
fingolfin, line 29: Is this still necessary, now that we have a `.gitattributes` file?

fingolfin:
This looks pretty clean now (thanks to @ssiccha sweeping the nasty bits under the rug 😁 ).

I wonder: would it worth the effort to also port the 32bit test? Then we could completely remove the Appveyor integration, which I'd be a big fan of : only one place to check for CI results, and more runners -- currently our GAP Days CI backlog still is big because all PRs wait on AppVeyor, which takes something like ~30 minutes per PR build, and it processes PRs sequentially

fingolfin (approved):

fingolfin:
I guess adding a 32bit version would require tweaking `gap-actions/setup-cygwin-for-gap@v1` a bit, adding a parameter to it for switching between 32 and 64 bit. And then we can use a simple job matrix in this workflow., mirroring that for AppVeyor.

.github/workflows/cygwin-windows.yml
ChrisJefferson: I believe .gitattribute is about how things are stored in the repository, not how they are checked out.

ssiccha: Removing that fails for me:
https://github.com/ssiccha/gap/runs/1925397861?check_suite_focus=true

ChrisJefferson:
The only required change for 32-bit cygwin is to use the 32-bit installer ( https://www.cygwin.com/setup-x86.exe ), but I note that nowadays cygwin strongly discourage using 32-bit nowadays (doesn't hurt to test it)

ssiccha:
I'm having a quick look

ssiccha:
Tests using 32bit cygwin can be found here:
- action: https://github.com/ssiccha/gap/blob/cygwin-CI/.github/workflows/cygwin-windows.yml
- results: https://github.com/ssiccha/gap/runs/1925397861?check_suite_focus=true

fingolfin:
Needs to be rebased

fingolfin:
I am happy to drop the 32bit binary

ssiccha:
Since @ChrisJefferson also said, that the cygwin people apparently don't want anybody to use the 32bit cygwin anymore, I'd say let's not distribute 32bit cygwin. 

I've rebased this and changed the `etc/` bits to `dev/`. Tests should pass again now. Let's see.

ssiccha:
Yepp, all tests pass. IMO this can be merged.

PR #4252, opened 2021-02-16
Add .gitattributes with `* text=auto`

fingolfin:
This normalizes line ends, and should help avoid complications caused by people editing files on windows and inserting CRLF line ends.

In the future we could add more complicated rules, see e.g. https://rehansaeed.com/gitattributes-best-practices/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


alex-konovalov (commented):
Never used this, but from the linked page and https://www.git-scm.com/docs/gitattributes sounds very appropriate.

fingolfin:
Just to clarify, we use this in various OSCAR related repositories, and since we switched to it, I haven't seen any CRLF files being committed, which happened occasionally before. But of course that's anecdotal evidence only.

alex-konovalov (approved):

fingolfin:

PR #4253, opened 2021-02-16
buildsys: overhaul `make distclean`

fingolfin:
The purpose of this target is to delete files that are not wanted in a
"distribution" / release tarball.

- do not delete `configure`, as it is required for release tarballs
- delete some auxiliary files created by LaTeX (but NOT the HTML and PDF files)

I stumbled over this while working on the new release script(s). Should be backported!

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -446,8 +446,9 @@ etags:
 # Rules for 'make clean'
 ########################################################################
 distclean: clean
-	rm -f config.log config.status libtool GNUmakefile configure
+	rm -f config.log config.status libtool GNUmakefile
 	rm -f doc/make_doc
+	rm -f doc/*/*.aux doc/*/*.bbl doc/*/*.blg doc/*/*.brf doc/*/*.idx doc/*/*.ilg doc/*/*.ind doc/*/*.log doc/*/*.out doc/*/*.pnr doc/*/*.tex doc/*/*.toc
 	rm -rf dev/log
 	rm -rf tags
 	rm -rf TAGS


alex-konovalov (commented):
@fingolfin We do have these:
```
gap-4.11.0 $ ls doc/*/*.i*
doc/hpc/main.ind	doc/ref/main.ind	doc/tut/main.ind
gap-4.11.0 $ ls doc/*/*.lab
doc/hpc/manual.lab	doc/ref/manual.lab	doc/tut/manual.lab
```
Are they needed for some gapmacro-based package manuals?

wilfwilson:
@alex-konovalov I don't know much about `*.lab` files but I do know that the GAP manual says something a bit relevant:

>   The  archive  of a GAP package should contain all files necessary for the package to work. In particular there should be a compiled documentation, which includes  the  manual.six,  manual.toc and manual.lab file in the documentation subdirectory which are created by GAPDoc while TeXing the documentation. (The  first  two  files are needed by the GAP help system, and the manual.lab file is needed if the main manuals or another package is referring to your package. Use the command GAPDocManualLab( packagename ); to create this file for your help books if you use GAPDoc.)

Perhaps @frankluebeck knows better whether we should be removing `*.ind` and `*.lab` files when making a GAP release.

fingolfin:
I blindly copied this from some other script, so I am not surprised they need to be adjusted. The failing test scripts further support this 😂

alex-konovalov:
I wonder why GAOP distribution includes {txt,html,six,lab,toc,ind,css} - see https://github.com/gap-system/gap-distribution/blob/7edd3b66a8c15b759f376a0d5cd5fc9a5161b927/DistributionUpdate/addmanualfiles#L19. Can't find out why - perhaps just in case...

fingolfin:
Updated

wilfwilson (approved):

Makefile.rules
```diff
@@ -446,8 +446,9 @@ etags:
 # Rules for 'make clean'
 ########################################################################
 distclean: clean
-	rm -f config.log config.status libtool GNUmakefile configure
+	rm -f config.log config.status libtool GNUmakefile
 	rm -f doc/make_doc
+	rm -f doc/*/*.aux doc/*/*.bbl doc/*/*.blg doc/*/*.brf doc/*/*.idx doc/*/*.ilg doc/*/*.ind doc/*/*.log doc/*/*.out doc/*/*.pnr doc/*/*.tex
```
frankluebeck, line 451: `doc/*/*.toc` are also generated by LaTeX and can be removed.

frankluebeck:
@alex-konovalov : Just now I cannot see a reason why `.ind`, `.toc` and even the `.tex` files should be distributed after GAPDoc manuals are built. 

Of course, in old style `gapmacro.tex` manuals the `.tex` and `.toc` files must  be kept since they are used by the help system.

Makefile.rules
fingolfin: Thanks, I removed .toc (not sure how I lost that, as the code where I copied the list from had it. Huh)

alex-konovalov:
@frankluebeck - thanks for confirming (and yes, we do not supply tex files for main GAP manuals in GAP 4.11.0 distribution).

fingolfin:

PR #4254, opened 2021-02-16
buildsys: rename a few helper scripts

fingolfin:
Instead of using `UPPER-CASE` use `lowercase.sh` to match our
conventions elsewhere.

This makes it trivial to delete some auxiliary files in `make distclean`.

The latter point is the main motivation, similar as to PR #4253: this simplifies scripting the GAP release process.


alex-konovalov (approved):

wilfwilson (approved):

fingolfin:

PR #4255, opened 2021-02-16
Fix some "lost" entries in the CHANGES.md

fingolfin:
Luckily, the gap.wiki is a git repository itself, so I could use
the history there to figure out what was originally intended.

To see what's wrong, search for 2955 on https://github.com/gap-system/gap/blob/master/CHANGES.md versus https://github.com/fingolfin/gap/blob/mh/fix-CHANGES.md/CHANGES.md

This should not interfere with the work by @ThomasBreuer @User141 (Daniel -- you really need a better use name 😂) @PaulaHaehndel but I am flagging them anyway just in case...


alex-konovalov (approved):

fingolfin:

PR #4263, opened 2021-02-17
Remove leftovers from Travis integration

fingolfin:
Should be backported, I guess?!

PR #4264, opened 2021-02-17
Sort the content of `dev` and `etc` directories

fingolfin:
The idea is that `etc` is shipped in GAP releases and contains extras
for users, while `dev` is not shipped and only for people interacting
with the GAP git repository. Hence move CI specific helper scripts as
well as `bisect.sh` from `etc` to `dev`.

Conversely, move some ctags related files from `dev` to `etc`, as those are
needed to let `make tags` function correctly.

Move `download.sh` from `etc` to `cnf` as it is used by the build system,
not users.

Finally, add `etc/README.md` explaining some of the files there.

(This doesn't touch or explain `etc/travis_fastfail.sh` as that is removed in #4263 which I made before realizing there should be bigger changes).


alex-konovalov (approved):
Very good! (I wonder if we need to put them in `dev/scripts` or something like that).

BTW, there is still `dev/rsyncserv` directory there - @frankluebeck do you think we need it there?

wilfwilson (approved):

frankluebeck:
@alex-konovalov : The directory  `dev/rsyncserv` is outdated and can now be deleted.
My new attempt to revive the rsync distribution looks good. If it works I will create a new git-repository for the code.

etc/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Files for GAP developers
+
+The following are tools for people working on the GAP code base itself
+
+- `ffgen.c`: used by the GAP build system
+- `tags.sh`: used by the `make tags` build target
+
+## Files for GAP package authors
+
+- `Makefile.gappkg`
```
ssiccha, line 15: What is this file for? That's missing.

fingolfin: Added a description

etc/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+The following are tools for people working on the GAP code base itself
+
+- `ffgen.c`: used by the GAP build system
+- `tags.sh`: used by the `make tags` build target
+
+## Files for GAP package authors
+
+- `Makefile.gappkg`: build rules intended for use by GAP packages with a kernel extension
+- `convert.pl`: script to convert old-style GAP manual TeX files to HTML
```
alex-konovalov, line 16: Perhaps emphasise and say "the old style gapmacro-based GAP manual TeX files"?

alex-konovalov:
@fingolfin please backport this - I can then update gap-distribution release wrapping scripts (easy fix there, and when new release tarballs will appear, we can compare results of two release processes to test).

fingolfin:
Backported to stable-4.11 in 94dfe32aa0dc2403b54b3c74131b02129e7c9f42

PR #4275, opened 2021-02-18
Enable SaveAndRestoreHistory by default

fingolfin:
I think a lot of people are missing out of this; and it should be safe for most. If this turns out to be wrong (as in: we get lots of complaints from people where it broke), we can still go back to the old default.

But perhaps I am missing other reasons that speak against it? @frankluebeck might have some suggestions.

Should IMHO be added to release notes, and even to the "top tier" news, because while being trivial and minor, I think many people will benefit from this (so many people to whom I taught this option reacted like "whaaaat??? I wish I had known about this years ago).

I am also tempted to raise the default number of lines from 1000 to 10000...

#### Save and restore input history by default

GAP has supported saving and restoring the interactive input history for a long time, but this feature was off by default. With this release, it is now on by default. That means that if you run GAP, enter some commands, then exit it, and start it again, you can reach the commands entered in the previous session by the usual means of accessing the command history, e.g. by pressing the up-arrow key. Should you prefer the old behaviour, you can get it back by setting the `SaveAndRestoreHistory ` user preference to false. For example, by entering these commands: `SetUserPreference( "SaveAndRestoreHistory", false ); WriteGapIniFile();`

In addition, the default limit on how many lines are preserved was increased from 1000 to 10000. This can be adjusted via the `HistoryMaxLines` user preference.

### DIFF
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -455,7 +455,7 @@ arbitrarily many lines.",
     "These preferences are ignored if GAP was not compiled with \
 readline support.",
     ],
-  default:= [1000, false],
+  default:= [10000, true],
   check:= function(max, save) 
     return ((IsInt( max ) and 0 <= max) or max = infinity) 
            and save in [true, false];


ssiccha:
I'm in favor of this PR. I'd also go with 10000 lines instead of a 1000. :)

FriedrichRober:
I think this is a very good idea. This should make working with GAP for the user more convenient, since many people do not know how to activate history.

ChrisJefferson:
I'm in favour of this

ChrisJefferson (approved):

ChrisJefferson:
I approve, I'd also approve raising the limit to 10,000

frankluebeck (approved):
My comment here is essentially the same as  [for the colored prompts](https://github.com/gap-system/gap/pull/4274#pullrequestreview-593200143).

The history feature is most useful for relatively recent GAP input, so I doubt that keeping much more than 1000 lines in the history is of much use. But I have no strong opinion about what the best number is.

fingolfin:
I increased the default value for `HistoryMaxLines` to 1000, and added a suggestion for the release notes text; feel free to suggest improvements.

RussWoodroofe:
Is there a convenient way to suppress adding a particular line to the command history?  Gap.app issues some commands from its GUI, most of which should probably not be added to the saved history.

fingolfin:
@RussWoodroofe hmm, I doubt it, but I am not sure (also, in general I know almost nothing about that interface Gap.app and xgap are using). I would suggest opening an issue requesting it; perhaps @frankluebeck has insights, but that's best discussed on such a new issue.

frankluebeck:
@RussWoodroofe There is no generic hook for configuring the history. But it is implemented as simple GAP code. I guess you would need to overwrite the function `AddHistory` which is in `lib/cmdledit.g`.

RussWoodroofe:
@frankluebeck  To complicating things, Gap.app loads gap without readline, so the less-tested code in cmdeditx.g is the relevant one.  But I found a workaround that I think is reasonably clean:  I replace LineEditKeyHandlers[1] with a short function that resets it back to CommandLineHistoryHandler.  (So GAP skips all end-of-line processing for exactly one command line.)  Thanks!

alex-konovalov:
Like this!

PR #4276, opened 2021-02-18
Fix optimization of writing to string-based streams

embray:
# Description

Just a little bug I noticed while trying to better understand how streams work.

The existing code failed to distinguish between IsOutputTextStringRep
and IsInputTextStringRep.

I don't know if there is a good way to test this (there are not many C tests in the test suite to begin with).  I confirmed in gdb that with this change writes to an OutputTextString stream go through the intended code path in [PutLine2](https://github.com/gap-system/gap/blob/7a57e8cf45b635423b48ad5864d1eaad45e76f51/src/io.c#L1088)

## Text for release notes 

Not sure if needed since it's not normally user-visible, but it's still a bug.

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1094,6 +1094,7 @@ static void PutLine2(TypOutputFile * output, const Char * line, UInt len)
     /* special handling of string streams, where we can copy directly */
     if (output->isstringstream) {
       str = CONST_ADDR_OBJ(output->stream)[1];
+      ConvString(str);
       lstr = GET_LEN_STRING(str);
       GROW_STRING(str, lstr+len);
       memcpy(CHARS_STRING(str) + lstr, line, len);
--- a/src/io.c
+++ b/src/io.c
@@ -49,7 +49,8 @@ static void PutLine2(TypOutputFile * output, const Char * line, UInt len);
 
 static Obj ReadLineFunc;
 static Obj WriteAllFunc;
-static Obj IsStringStream;
+static Obj IsInputStringStream;
+static Obj IsOutputStringStream;
 static Obj PrintPromptHook = 0;
 Obj EndLineHook = 0;
 static Obj PrintFormattingStatus;
@@ -407,7 +408,7 @@ UInt OpenInputStream(TypInputFile * input, Obj stream, BOOL echo)
     input->prev = IO()->Input;
     input->isstream = TRUE;
     input->stream = stream;
-    input->isstringstream = (CALL_1ARGS(IsStringStream, stream) == True);
+    input->isstringstream = (CALL_1ARGS(IsInputStringStream, stream) == True);
     if (input->isstringstream) {
         input->sline = CONST_ADDR_OBJ(stream)[2];
         input->spos = INT_INTOBJ(CONST_ADDR_OBJ(stream)[1]);
@@ -855,7 +856,7 @@ UInt OpenOutputStream(TypOutputFile * output, Obj stream)
     output->prev = IO()->Output;
     IO()->Output = output;
     output->isstream = TRUE;
-    output->isstringstream = (CALL_1ARGS(IsStringStream, stream) == True);
+    output->isstringstream = (CALL_1ARGS(IsOutputStringStream, stream) == True);
     output->stream = stream;
     output->line[0] = '\0';
     output->pos = 0;
@@ -1964,7 +1965,8 @@ static Int InitKernel (
     /* import functions from the library                                   */
     ImportFuncFromLibrary( "ReadLine", &ReadLineFunc );
     ImportFuncFromLibrary( "WriteAll", &WriteAllFunc );
-    ImportFuncFromLibrary( "IsInputTextStringRep", &IsStringStream );
+    ImportFuncFromLibrary( "IsInputTextStringRep", &IsInputStringStream );
+    ImportFuncFromLibrary( "IsOutputTextStringRep", &IsOutputStringStream );
     InitCopyGVar( "PrintPromptHook", &PrintPromptHook );
     InitCopyGVar( "EndLineHook", &EndLineHook );
     InitFopyGVar( "PrintFormattingStatus", &PrintFormattingStatus);
--- a/lib/streams.gd
+++ b/lib/streams.gd
@@ -68,6 +68,23 @@ DeclareRepresentation(
 
 #############################################################################
 ##
+#R  IsOutputTextStringRep   (used in kernel)
+##
+##  <ManSection>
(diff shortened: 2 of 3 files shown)


fingolfin (approved):

fingolfin:
The crashes are weird...

embray:
It is really weird.  I was hoping 16e0161, which was needed anyways, would fix it.  But I can't reproduce these crashes locally.

ChrisJefferson:
This is the bug, there has to be a 'ConvString' added (as below), basically in some cases a non-string ends up here (it's an empty list). ConvString is cheap when we already have a string, so (I think) we may as well check every time for safety.

Do you mind adding this to your PR and see if it fixes it?

```
    if (output->isstringstream) {
      str = CONST_ADDR_OBJ(output->stream)[1];
      ConvString(str);
      lstr = GET_LEN_STRING(str);
      GROW_STRING(str, lstr+len);
```

fingolfin:
OK, but how does this PR trigger the crash? Is there a place where we use an input stream where an output stream is expected, or vice-versa?

ChrisJefferson:
Sorry, too brief!

The bug is if someone constructs an OutputTextString with something which is not a StringRep, for example (cut down from that test:

```
 s := OutputTextString([], false);
OutputTextString(0)
gap> PrintTo(s, "abc");
```
That bug has always been there, but because we messing up checking if things were input/output/stringstreams, this code was never been run before, so we were doing the slow path which was in GAP so correctly handled the plist (see https://codecov.io/gh/gap-system/gap/src/master/src/io.c, where it is never currently executed ).

fingolfin:
> but because we messing up checking if things were input/output/stringstreams, this code was never been run before

I still don't understand this bit. The way I understand the change, it should only matter if someone uses an input string stream for output, or vice-versa. Where do we do this? In some test file?  Or am I misunderstanding this?

ChrisJefferson:
Be sure to look at the (old) variable definition (the name IsStringStream is very misleading)

`ImportFuncFromLibrary( "IsInputTextStringRep", &IsStringStream );`

So, output string streams (which are not IsInputTextStringRep) would not be marked isstringstream. So in practice, when outputting, nothing was getting the member isstringstream set to true before, so all the special code for outputting to stringstreams was never being run.

embray:
@ChrisJefferson Thank you, I'll try that.  Still not sure why I couldn't reproduce it.

embray:
By the way, I have some thoughts on how I might like to further refactor some of the code and structures in io.c to make it a bit more generic, rather than having checks everywhere like `file->isstream`, etc.  Reason being, I want to make it easier in the libgap interface to define custom I/O handlers, and I think this can be made easier with some reorganization of the existing code.  I'll submit a PR to make what I have in mind more concrete, but I want to make sure no one else is already thinking of any more major IO refactoring so that I don't end up stepping on any toes.  @fingolfin , it looks like you've been doing the most with that code lately.

ChrisJefferson:
The lack of reproduction is because this is a low-level memory corruption, so it is effected by minor changes.

If you configure GAP with --enable-debug, this turns on many extra asserts, which will catch the problem here fairly immediately.

embray:
Still one weird failure on 32-bit.  I can't tell if it is possibly related.

src/io.c
```diff
@@ -1093,6 +1094,7 @@ static void PutLine2(TypOutputFile * output, const Char * line, UInt len)
     /* special handling of string streams, where we can copy directly */
     if (output->isstringstream) {
       str = CONST_ADDR_OBJ(output->stream)[1];
+      ConvString(str);
```
fingolfin, line 1097: This is still unsafe in that `ConvString` doesn't check whether the given object is a string. So at the very least we should insert an assertion here:
```suggestion
      GAP_ASSERT(IS_STRING(str));
      ConvString(str);
```
Of course even better might be to always handle this, but given that we are in the middle of outputting text, just calling `Error` blindly may be a really bad idea... We could also just do `if (!IS_STRING(str)) Panic(...)` (with the idea that crash is better than silently corrupting data). Perhaps somewhat better is to simply fall back to the generic code in this case (but chances are it'll also just end up running into an `Error`. BTW, one can also use `IsStringConv`... so like this:
```
if (output->isstringstream) {
  str = ..
  if (IsStringConv(str)) {
    lstr = ..
    ...
    return;
  }
}
```

ChrisJefferson: I think it will get caught in ConvString eventually, but something clearer might be nice. Also, asserts are only in debug builds anyway.

I agree that IsStringConv, then heading back to the GAP level path (which will Error rather than crash) is the best option. Sorry @embray we are pushing this around so much.

fingolfin: It is true that in debug mode (with assertion on), `ConvString` will catch bad values (but indirectly: first `LEN_LIST` will assert out if it's not a list, then later `CHAR_VALUE` if its values are not vars.

But in release mode, only `LEN_LIST` will catch non-lists, but `CHAR_VALUE` will produce garbage. And if the input happens to be a list of immediate integers, it might segfault.

But in all of these cases, there won't be a very helpful error message... It might be insightful to try and triggers this on purpose, and see what happens?

Anyway, I do not yet see how any of this could explain the 32bit crash... I will try and see if I can reproduce it on an Ubuntu 20.20 box.

ChrisJefferson: Yes, that 32-bit crash I'm unable to reproduce myself (I might have another try)

ChrisJefferson: Reading the docs for OutputTextString, it even talks about "puts all received characters into the list 'list'", so a user could reasonably read that, and use OutputTextString in lists which are not strings

```
gap> x := [1];
[ 1 ]
gap> w := OutputTextString(x, false);
OutputTextString(0)
gap> PrintTo(w, "hello");
gap> x;
"hello"
gap> x := [1];
[ 1 ]
gap> w := OutputTextString(x, true);
OutputTextString(1)
gap> PrintTo(w, "hello");
gap> x;
[ 1, 'h', 'e', 'l', 'l', 'o' ]
```

fingolfin: evil :-)

fingolfin: I've restarted the jobs once again, just rule out a random fluke (which would still be worrisome, but might not be directly caused by this PR).

fingolfin (approved):
Tests pass now, huh... the title still says "WIP", can that be removed and the PR be merged? (Yes from my POV, but confirmation by @embray would be nice)

embray:
Good to go from my end as long as the tests are passing now :+1:

PR #4279, opened 2021-02-22
Add scripts for making releases / distributions

fingolfin:
These are the result of work during the GAP Days "Spring" 2021.


ssiccha (commented):
I'd say let's get at least utils.py, make_tarball.py and make_github_release.py into the main branch since they're more or less finished. Could update_website.py also be merged in?

ruthhoffmann (commented):
I agree with @ssiccha; merge what is ready. We could merge website_update as well but would have to note the manual steps in the README.

fingolfin:
I would rather merge everything and continue work here. In the meantime, I still want to go over everything myself, too (didn't really get to it last week); I just started with some first pushes

wilfwilson (approved):
Yes, I'd like to get this merged in so that we can each most sensibly continue work in separate PRs to this repo, if I want to.

fingolfin:
I've dropped the GitHub Action code from this PR again, and will open a separate PR later. 

I think we should merge this ASAP, and then make improvements via further PRs by all involved parties. OK?

Note that I already made a few tweaks, compared to our old repository, which I left in a second commit to make them easily visible.

wilfwilson:
Sounds good to me.

ruthhoffmann (commented):
Had a look over bits and bobs. Things are looking good! A couple comments/fixes in `README.md` are recommended (I don't want to mess with the PR commits);
Line 11 in Readme has a typo  "(as outline**d** in the root..."
The texts in the numbered lists of the two guides should be the same (this is a personal preference).

ruthhoffmann:
Of course I can't read. Yes, merging and then tinkering sounds like a good plan!

fingolfin:
I've merged this now; @ruthhoffmann you have valid points, a PR to fix them would be welcome, otherwise I'll try to take care of it when I have some spare time, but for now if I have any spare minute I'll try to spend it on getting closer to having an actual release out...

ruthhoffmann:
Creating that PR just now. :)

PR #4280, opened 2021-02-22
Don't freak out if a package has an empty or short Date

fingolfin:
Fixes a regression from PR #3355.

I've had a package prototype sitting around with an empty Date field.
While of course the validator rejects that, the package loading code
should not, or at least should not run into a break loop.

No test case, but I think the problem and fix are obvious.


alex-konovalov (approved):
OK, I agree!

wucas (approved):
Yes that's an unfortunate (and somewhat embarrassing) oversight on my part (not checking all assumptions of the code^^). But the fix looks sensible to me.

PR #4286, opened 2021-02-24
Further work on dev/releases/update_website.py (and related changes)

wilfwilson:
I've spent some time this evening doing working on the release scripts, especially the `update_website.py` script. I've also made a few tweaks to the `utils.py` scratch.

The main changes in this PR:

* Use PyGithub to access the information about GAP releases
* Use the release date defined in `configure.ac` of the GAP release tarball
* Most of the diffs are because I've standardised on double quotes for strings, rather than single quotes. (Sorry to any potential reviewers).

I'm aware that we'd like to release GAP 4.11.1 this week (or ASAP afterwards). I also realise that there are significant changes that we want to make to the way that the GAP website works and is updated. So hopefully the script we use fo GAP 4.12.0 will be quite different.

I'll only push things here that leave the `update_website.py` script in a working state. So it can be merged whenever, from my point of view, and still be usable for the GAP 4.11.1 release. I might have time to work on this further this week, but I might not.

The main missing bits of important functionality:
* Pushing the new branch with proper GitHub token-based authentication. (Currently, just `git push <remote> <branch>` is run.)
* Creating a PR to gap-system/GapWWW. (Currently, this has to be done manually.)

### DIFF
--- a/dev/releases/update_website.py
+++ b/dev/releases/update_website.py
@@ -9,13 +9,17 @@
 
 #import subprocess
 import argparse
-import sys
-import requests
 import datetime
-import dateutil.parser
-import tempfile
+import github
+import os
+import re
+import requests
+import subprocess
+import sys
 import tarfile
-from utils import *
+import tempfile
+import utils
+from utils import notice, error, verify_command_available, verify_git_repo, is_git_clean, verify_git_clean, working_directory, download_with_sha256
 from git import Repo
 
 
@@ -28,10 +32,11 @@ if sys.version_info < (3,6):
 ################################################################################
 # Utility functions
 def is_possible_gap4_release_tag(tag):
-    tag = tag.split('.')
-    return len(tag) == 3 and tag[0] == 'v4' and all(x.isdigit() for x in tag[1:])
+    tag = tag.split(".")
+    return len(tag) == 3 and tag[0] == "v4" and all(x.isdigit() for x in tag[1:])
 
 def mb_bytes(nrbytes):
+    # (str) bytes to (str) megabytes
     return str(round(int(nrbytes) / (1024 * 1024)))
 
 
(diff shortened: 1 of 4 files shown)


dev/releases/utils.py
```diff
@@ -155,10 +158,17 @@ def check_whether_github_release_exists(tag):
...
               + " are already initialized.")
     if token == None and "GITHUB_TOKEN" in os.environ:
         token = os.environ["GITHUB_TOKEN"]
+    if token == None:
+        temp = subprocess.run(["git", "config", "--get", "github.token"], text=True, capture_output=True)
+        if temp.returncode == 0:
+            token = temp.stdout.strip()
+    if token == None and os.path.isfile(os.path.expanduser('~') + '/.github_shell_token'):
+        with open(os.path.expanduser('~') + '/.github_shell_token', 'r') as token_file:
+            token = token_file.read().replace('\n', '')
```
wilfwilson, line 171: Note to self, this could be:
```suggestion
            token = token_file.read().strip()
```

dev/releases/update_website.py
```diff
@@ -138,124 +139,112 @@ def mb_bytes(nrbytes):
...
         else:
             assets_unix.append(asset)
 
 # Release date
 if args.use_github_date:
-    release_date = dateutil.parser.isoparse(release['published_at'])
-year  = release_date.strftime('%Y')
-month = release_date.strftime('%B')
-date  = release_date.strftime('%d %B %Y')
+    release_date = release.published_at
```
fingolfin, line 182: What type has `release_date` at this point? Is it a string? A python date object? (I have an ulterior motive for asking, but I think perhaps it would be good to record this as a comment here for the future anyhow)

dev/releases/update_website.py
```diff
@@ -138,124 +139,112 @@ def mb_bytes(nrbytes):
...
+if not args.date and not args.use_github_date:
+    with working_directory(gaproot):
+        with open("configure.ac", "r") as configure_ac: 
+            filedata = configure_ac.read()
+            try:
+                # Expect date in YYYY-MM-DD format
+                release_date = re.search('\[gap_releaseday\], \[(\d{4}-\d{2}-\d{2})\]', filedata).group(1)
+            except:
+                error("Cannot find the release date in configure.ac, consider using --date")
+            release_date = datetime.datetime.strptime(release_date, "%Y-%m-%d")
```
fingolfin, line 235: So before this, `release_date` is a YYYY-MM-DD string, right? Do we really need to parse that? Can't we work with the string below? It's trivial to extract year, month, day from it and thus also the `date` value below. That way, we avoid the dependency on `datetime`. 

It may not be worth the effort to change for now; I am mainly asking to see if there are some deeper difficulties I am missing. (Depending on the answer to that, I will perhaps add a comment here explaining it either way, to stop future me from fretting about it ;-) )

dev/releases/update_website.py
wilfwilson: Dates that you get via PyGitHub seem to always be Python `datetime` objects. In particular that is the case here.

dev/releases/update_website.py
wilfwilson: Currently, the date can get into this script in three ways:
1. via the argument `--date` (which should be checked is in YYYY-MM-DD format)
2. via configure.ac (which we can rely on being in proper YYYY-MM-DD format)
3. via the option `--use-github-date` (which gets a `datetime` object from the GitHub API)

Currently I convert them all to `datetime` objects to benefit from its validation, and to have them all in a standard format for getting the day, month, year.

For 1, converting to a `datetime` object allows us to avoid having to validate the date ourselves. But of course we could do this ourselves.
For 2, yes we could parse this directly.
For 3, I think we need the `datetime` module to parse this (but I'm not an expert). However, maybe the option`--use-github-date` should be removed anyway.

dev/releases/update_website.py
wilfwilson: I have added such a comment.

dev/releases/update_website.py
fingolfin: I think validation is a good argument for all of this. Anyway, I am 100% happy to leaves this as-is

fingolfin (approved):

PR #4288, opened 2021-02-25
buildsys: change bootstrap targets to fetch from GitHub

fingolfin:
This should be more reliable than the GAP website; and perhaps
is also more efficient for GitHub actions.


wilfwilson (approved):

fingolfin:
Backported to stable-4.11 in 3c272b6b13e59d6bdf2486abc4976c3308ce3cfe

Issue #4272, opened 2021-02-18
Add user available functions to open a URL in a web browser


fingolfin:
We already go through a lot of effort to set `HELP_VIEWER_INFO.browser` to a function like this for HTML; but I know that the very least, the [numericalsgps](https://github.com/gap-packages/numericalsgps) package by @pedritomelenas and friends needs such functionality, see https://github.com/gap-packages/numericalsgps/blob/231329432a3e6e651cfc8e719271ccb23ae973a7/gap/dot.gi#L62-L68 -- but I also think other uses might exist.

So, let's expose it through a dedicated function, say `OpenURLInBrowser(url)`. Shouldn't be too hard (most work is a writing a reference manual entry for it)

One could do the same for PDFs, too, I guess...

pedritomelenas:
Also [jupyterviz](https://www.gap-system.org/Packages/jupyterviz.html) has this functionality, since I asked @nathancarter to add it and he kindle agreed. Probably [francy](https://www.gap-system.org/Packages/francy.html) also has it. I think that both used similar code as the one @fingolfin describes in dot.gi. 
There is some "splash" functions in other packages, like in [intpic](https://www.gap-system.org/Packages/intpic.html) by @ManuelAFDelgado. 
One probably wants to splash a pdf or probably something written in graphviz dot language.

RussWoodroofe:
Suggestion:  A starting point would be to have OpenURLInBrowser look at the help viewers.  Typically, the preferred help viewer would also be the preferred browser for other purposes, and there's already some sometimes-nontrivial code to e.g. bring up a URL in Safari and go to a given anchor.

nathancarter:
It's best to defer opening a file to the operating system, which will use the user's settings for the given file type.  The relevant code is just these 7 lines:
https://github.com/nathancarter/jupyterviz/blob/master/lib/main.gi#L79-L85
```gap
if ARCH_IS_MAC_OS_X() then
    Exec( "open ", filename );
elif ARCH_IS_WINDOWS() then
    Exec( "start ", filename );
elif ARCH_IS_UNIX() then
    Exec( "xdg-open ", filename );
fi;
```

Note that since writing them, I now realize that there should probably be another entry in there for WSL, which should use the `wslview` command, but I'm not sure how to test if we're in a WSL environment...it would probably be a subcase of `ARCH_IS_UNIX()`.

ChrisJefferson:
[...]

PR #4295, opened 2021-02-27
Add OpenExternal, a function for opening files in the OS

ChrisJefferson:
Add a function "OpenExternal", which opens files in other apps in the file system. While I like the names used by existing packages (like Splash), I also don't want to clash with those existing packages.

Rather than base this on the existing code used for help, I wrote a new function. This does remove the customizability, but also means we can open any file type. Also, in my experience, nowadays the "default" operating system method of opening files is what I want.

This also does not implement the functionality provided by (for example) semigroups, where files such as '.dot' and '.tex' files are compiled before being opened.

I'm very happy for suggestions on naming / functionality / etc. I thought having a PR would give a place to discuss such issues.

Discussed in #4272

### DIFF
--- a/lib/streams.gd
+++ b/lib/streams.gd
@@ -1232,3 +1232,21 @@ DeclareGlobalFunction( "UnInstallCharReadHookFunc" );
 ##  <#/GAPDoc>
 ##
 DeclareGlobalFunction( "InputFromUser" );
+
+#############################################################################
+##
+#F  OpenExternal( <filename> )
+##
+##  <#GAPDoc Label="OpenExternal">
+##  <ManSection>
+##  <Func Name="OpenExternal" Arg='filename'/>
+##
+##  <Description>
+##  Open the file <A>filename</A> using the default application for this file
+##  in the operating system. This can be used to open files like HTML and PDF
+##  files in the GUI.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction( "OpenExternal" );
--- a/lib/streams.gi
+++ b/lib/streams.gi
@@ -1388,6 +1388,29 @@ InstallGlobalFunction( InputFromUser,
   end );
 
 
+#############################################################################
+##
+#M  OpenExternal(filename)  . . . . . . . . . . . . open file in external GUI
+##
+InstallGlobalFunction( OpenExternal, function(filename)
+    local file;
+    if ARCH_IS_MAC_OS_X() then
+      Exec(Concatenation("open \"",filename,"\""));
+    elif ARCH_IS_WINDOWS() then
+      Exec(Concatenation("cmd /c start \"",filename,"\""));
+    elif ARCH_IS_WSL() then
+      # If users pass a URL, make sure if does not get mangled.
+      if ForAny(["https://", "http://"], {pre} -> StartsWith(filename, pre)) then
+        file := filename;
+      else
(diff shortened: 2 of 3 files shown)


lib/streams.gd
```diff
@@ -1232,3 +1232,21 @@ DeclareGlobalFunction( "UnInstallCharReadHookFunc" );
...
+##
+#F  OpenExternal( <filename> )
+##
+##  <#GAPDoc Label="OpenExternal">
+##  <ManSection>
+##  <Func Name="OpenExternal" Arg='filename'/>
+##
+##  <Description>
+##  Open the file <A>filename</A> using the standard application for this file
+##  provided by the operating system. This can be used to open files like HTML and PDF
```
wilfwilson, line 1246:
```suggestion
##  Open the file <A>filename</A> using the default application for this file
##  in the operating system. This can be used to open files like HTML and PDF
```
I think "default" is a more standard word that "standard" in this context. And in my opinion, someone could interpret "provided by the OS"  to mean that it only uses applications that are _bundled_ with the OS. But maybe not, and even if so, my suggestion is not necessarily better.

wilfwilson (commented):
I like this feature, and I think `OpenExternal` is a decent name for it.

lib/streams.gd
```diff
@@ -1232,3 +1232,21 @@ DeclareGlobalFunction( "UnInstallCharReadHookFunc" );
...
+##  files in the GUI.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction( "OpenExternal" );
```
fingolfin, line 1252:
```suggestion
DeclareGlobalFunction( "OpenExternal" );

```

wilfwilson (approved):

wilfwilson:
I'd be happy to merge this, I just wanted to sanity check before doing so:
* Have you tested the code yourself on the various operating systems? I've tried it on my Mac and I can confirm that it worked as expected.
* Is it possible to add tests in GAP for `OpenExternal`? I presume such a test might be too brittle (and what would success even look like?) but if possible to do robustly, it could at least catch the case that `open`/`explorer.exe`/`xdg-open`... become renamed or whatever.

fingolfin:
I don't see a good way to test this automatically in our CI. So I'd just merge this as-is, but I'll wait to give a chance for more comments.

fingolfin (approved):

lib/streams.gi
```diff
@@ -1388,6 +1388,29 @@ InstallGlobalFunction( InputFromUser,
...
 
 
+#############################################################################
+##
+#M  OpenExternal(filename)  . . . . . . . . . . . . open file in external GUI
+##
+InstallGlobalFunction( OpenExternal, function(filename)
+    local file;
+    if ARCH_IS_MAC_OS_X() then
+      Exec(Concatenation("open \"",filename,"\""));
```
fingolfin, line 1398: OK, I don't want to whine, but I have to repeat one of my pet mantras, namely that "Exec is a really bad function, we should never have added it like that". Besides not even allowing to check the return value, it also forces us to do silly quoting tricks like the above; and this is not even complete, a filename with a `"` in it will break it. We really should think about adding an alternative to `Exec` which is a bit better (doesn't go through a shell and hence has no quoting issues, and gives access to return value, and optionally allows for suppressing or redirecting output.

fingolfin: (But of course none of this is required for this PR!)

PR #4307, opened 2021-03-02
Upload some artifacts for "nightly releases"

fingolfin:
... so that we can use them for further build steps / review them
for debugging.

Not sure if I even got the syntax right. Perhaps @ssiccha or @FriedrichRober can help finishing this.

Also, right now this mixes a bunch of things together (gap-X.Y.Z.tar.gz, gap-X.Y.Z-core.tar.gz, package-info.json, and their sha256 files) -- perhaps better to upload them as multiple separate artifacts

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,6 +56,7 @@ jobs:
           path: |
             tmp/gap-*.tar.gz*
             tmp/*json*
+            !tmp/gap-*-core.tar.gz*
 
       - name: "Make GitHub release"
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,7 @@ jobs:
         run: python -u ./dev/releases/make_archives.py
 
       - name: "Upload artifacts"
-        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') && failure() }}
         # Warning: the result is a single .zip file (so things are compressed twice)
         # To keep things from exploding, we only upload a subset of all generated files
         uses: actions/upload-artifact@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,17 @@ jobs:
         run: bash dev/ci-prepare.sh
       - name: "Make archives"
         run: python -u ./dev/releases/make_archives.py
+
+      - name: "Upload artifacts"
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        # Warning: the result is a single .zip file (so things are compressed twice)
+        # To keep things from exploding, we only upload a subset of all generated files
+        uses: actions/upload-artifact@v2
+        with:
+          path: |
+            tmp/gap-*.tar.gz*
+            tmp/*json*
+
       - name: "Make GitHub release"
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
         # TODO: we should check whether the tag triggering the release


.github/workflows/release.yml
```diff
@@ -46,6 +46,17 @@ jobs:
         run: bash dev/ci-prepare.sh
       - name: "Make archives"
         run: python -u ./dev/releases/make_archives.py
+
+      - name: "Upload artifacts"
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
```
fingolfin, line 51: Uhh, I didn't even check if this syntax is valid...

.github/workflows/release.yml
```diff
@@ -46,6 +46,17 @@ jobs:
...
+        # Warning: the result is a single .zip file (so things are compressed twice)
+        # To keep things from exploding, we only upload a subset of all generated files
+        uses: actions/upload-artifact@v2
+        with:
+          path: |
+            tmp/gap-*.tar.gz*
+            tmp/*json*
```
fingolfin, line 58: Note that one can also exclude patterns (see <https://github.com/actions/upload-artifact>), which can be useful to match `gap-X.Y.Z.tar.gz` but not `gap-X.Y.Z-core.tar.gz`

FriedrichRober:
Ich schaue es mir mal an.

.github/workflows/release.yml
FriedrichRober: The syntax should be correct according to GitHub Docs.

.github/workflows/release.yml
FriedrichRober: I think adding `!tmp/gap-*-core.tar.gz*` should do the job to exclude `gap-X.Y.Z-core.tar.gz`

FriedrichRober:
I am running the workflow on my fork and test if the artifacts upload correctly.

.github/workflows/release.yml
FriedrichRober:
```suggestion
            tmp/*json*
            !tmp/gap-*-core.tar.gz*
```

FriedrichRober:
So everything seems to work fine on my fork: [Upload-Artifacts](https://github.com/FriedrichRober/gap/actions/runs/614764414)

Changes made on my fork:
I added the suggested line for excluding the file `gap-X.Y.Z-core.tar.gz`.
I commented out line 23 in `relase.yml` so that the workflow could trigger on my branch.

FriedrichRober:
I looked into the following [Issue](https://github.com/actions/upload-artifact/issues/3) to see if we can upload artifacts separately. As far as I understand this is currently not possible. The files are uploaded separately but at the moment we can access them through the GitHub UI only as a compressed zip file all at once.

fingolfin:
Thanks @FriedrichRober

fingolfin:
I am wary about merging this as-is, as it would add hundreds of MB of data *to every single build*; so if a PR is rebuilt 5 times before it is merged, you get 5 such artifacts, and they are kept by default for 90 days. So if there is a lot of activity, the data accumulates quickly... At least that's how I understand it to  work?

Maybe more reasonable is to keep the archives only if a special flag is set. I.e., if a PR *fails* the release test, then we want to be able to get a look at the archives; perhaps this could be done by checking for a specific label on the PR? Not sure if (and how) we could access such information, though. Another alternative would be to add a manual trigger for the "release" workflow, and *that* uploads archives.

Finally, we still want to set up release builds that happen on a clock (say 24 hours), and for those we also will want to upload the archives?

ssiccha:
@fingolfin 
> Maybe more reasonable is to keep the archives only if a special flag is set. I.e., if a PR fails the release test

Ha, I've actually found what you're looking for a few days ago! We can do [this](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure) to run a step only if the job itself failed.

ssiccha:
See the commit that I just pushed.

.github/workflows/release.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         run: python -u ./dev/releases/make_archives.py
 
       - name: "Upload artifacts"
-        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') && failure() }}
```
fingolfin, line 51: nice!

Can we test this? As in, let's check that the current job does not produce an artifact; then perhaps @ssiccha can push a commit which introduces a failure somewhere to see what happens?

ssiccha: This is such a niice feature. This did not upload an artifact: 
https://github.com/gap-system/gap/pull/4307/checks?check_run_id=2021086422 
I'll now push something that fails.

ssiccha: This works. Here I introduced an error and it is currently uploading artifacts:
https://github.com/gap-system/gap/pull/4307/checks?check_run_id=2021529089

I'm removing my error commit again

ssiccha (approved):
LGTM

PR #4308, opened 2021-03-02
Make release scripts more robust

ssiccha:
Thank You!

# Description

## Text for release notes 

is suitable as title.

in the root directory.

## Further details

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,8 +48,6 @@ jobs:
         run: python -u ./dev/releases/make_archives.py
       - name: "Make GitHub release"
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
-        # TODO: we should check whether the tag triggering the release
-        # has the right name
-        run: python -u ./dev/releases/make_github_release.py
+        run: python -u ./dev/releases/make_github_release.py ${GITHUB_REF#refs/tags/} tmp/
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/dev/releases/make_github_release.py
+++ b/dev/releases/make_github_release.py
@@ -24,14 +24,18 @@ PATH_TO_RELEASE = sys.argv[2]
 if re.fullmatch( r"v[1-9]+\.[0-9]+\.[0-9]+", TAG_NAME) == None:
     utils.error("This does not look like a release version")
 
+# Initialize GITHUB_INSTANCE and CURRENT_REPO
+utils.initialize_github()
+
+# Error if the tag TAG_NAME hasn't been pushed to CURRENT_REPO yet.
+if not any(tag.name == TAG_NAME for tag in utils.CURRENT_REPO.get_tags()):
+    utils.error(f"Repository {utils.CURRENT_REPO_NAME} has no tag '{TAG_NAME}'")
+
 # make sure that TAG_NAME
 # - exists
 # - is an annotated tag
 # - points to current HEAD
 utils.check_git_tag_for_release(TAG_NAME)
-
-# Initialize GITHUB_INSTANCE and CURRENT_REPO
-utils.initialize_github()
 
 # Error if this release has been already created on GitHub
 if any(r.tag_name == TAG_NAME for r in utils.CURRENT_REPO.get_releases()):
--- a/dev/releases/make_github_release.py
+++ b/dev/releases/make_github_release.py
@@ -56,7 +56,7 @@ with utils.working_directory(PATH_TO_RELEASE):
     # Now check that TAG_NAME and the created archives belong together
     main_archive_name = "gap-" + TAG_NAME[1:] + ".tar.gz"
     if not main_archive_name in manifest:
-        error(f"Expected a file {main_archive_name} but it does not exist!")
+        utils.error(f"Expected a file {main_archive_name} but it does not exist!")
     # Upload all assets to release
     try:
         for filename in manifest:
--- a/dev/releases/make_github_release.py
+++ b/dev/releases/make_github_release.py
@@ -3,51 +3,56 @@
 # This script is intended to implement step 6 of
 # <https://hackmd.io/AWds-AnZT72XXsbA0oVC6A>, i.e.
 # it makes a github release and uploads all tar balls as assets.
-# The target repository is hardcoded in utils.py:
-#   CURRENT_REPO_NAME
-#
-# As in make_archives.py, the version of the gap release is taken from the
-# Makefile variable GAP_BUILD_VERSION.
+# The name of the target repository CURRENT_REPO_NAME is defined in utils.py.
 
 # If we do import * from utils, then initialize_github can't overwrite the
 # global GITHUB_INSTANCE and CURRENT_REPO variables.
 import utils
 import github
 import os
+import sys
 import re
 
-# Identify GAP release version
-try:
-    GAPVERSION = utils.get_makefile_var("GAP_BUILD_VERSION")
-except:
-    utils.error("Could not get GAP version")
+if len(sys.argv) != 3:
+    utils.error("usage: "+sys.argv[0]+" <tag_name> <path_to_release>")
 
-utils.notice(f"Detected GAP version {GAPVERSION}")
-if re.fullmatch( r"[1-9]+\.[0-9]+\.[0-9]+", GAPVERSION) == None:
-    utils.error("This does not look a release version")
+utils.verify_git_clean()
 
+TAG_NAME = sys.argv[1]
+PATH_TO_RELEASE = sys.argv[2]
 
+if re.fullmatch( r"v[1-9]+\.[0-9]+\.[0-9]+", TAG_NAME) == None:
+    utils.error("This does not look like a release version")
+
+# make sure that TAG_NAME
+# - exists
+# - is an annotated tag
+# - points to current HEAD
+utils.check_git_tag_for_release(TAG_NAME)
+
+# Initialize GITHUB_INSTANCE and CURRENT_REPO
 utils.initialize_github()
 
 # Error if this release has been already created on GitHub
-if utils.check_whether_github_release_exists("v"+GAPVERSION):
-    utils.error(f"Release v{GAPVERSION} already exists!")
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -48,8 +48,6 @@ jobs:
         run: python -u ./dev/releases/make_archives.py
       - name: "Make GitHub release"
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
-        # TODO: we should check whether the tag triggering the release
-        # has the right name
         run: python -u ./dev/releases/make_github_release.py
```
fingolfin, line 51: Shouldn't the tag name be passed as an argument here?

dev/releases/utils.py
```diff
@@ -152,6 +152,38 @@ def check_whether_github_release_exists(tag):
...
+# lightweight vs annotated
+# https://stackoverflow.com/questions/40479712/how-can-i-tell-if-a-given-git-tag-is-annotated-or-lightweight#40499437
+def is_annotated_git_tag(tag):
+    is_annotated = subprocess.run(["git", "for-each-ref", "refs/tags/" + tag],
+                                  check=True, capture_output=True, text=True)
+    if "tag" == is_annotated.stdout.split()[1]:
+        return True
+    if "commit" == is_annotated.stdout.split()[1]:
+        return False
+    error(f"could not check whether tag {tag} is annotated")
```
fingolfin, line 164: Why not just use this, and get rid of `does_git_tag_exist`?
```suggestion
def is_annotated_git_tag(tag):
    res = subprocess.run(["git", "for-each-ref", "refs/tags/" + tag],
                         capture_output=True, text=True)
    return res.returncode <> 0 and "tag" == res.stdout.split()[1]:
```

dev/releases/utils.py
```diff
@@ -152,6 +152,38 @@ def check_whether_github_release_exists(tag):
...
+def get_commit_of_git_tag(tag):
+    result = subprocess.run(["git", "show-ref", "--tags", "--dereference", tag],
+                            check=True, capture_output=True, text=True)
+    result = result.stdout.split('\n')
+    # TODO finish this
+    return commit
+
+def check_git_tag_for_release(tag):
+    if not does_git_tag_exist(tag):
+        error(f"There exists no tag {TAG_NAME}")
```
fingolfin, line 175:
```suggestion
    if not does_git_tag_exist(tag):
        error(f"There exists no tag {tag}")
```
but I'd rather use the "improved" `is_annotated_git_tag` I suggest above:
```suggestion
    if not is_annotated_git_tag(tag):
        error(f"There is no annotated tag {tag}")
```

dev/releases/utils.py
```diff
@@ -143,7 +143,7 @@ def check_whether_git_tag_exists(tag):
     return False
 
 # Returns a boolean
-def check_whether_github_release_exists(tag):
+def does_github_release_exist(tag):
     if CURRENT_REPO == None:
```
fingolfin, line 147: Why is `CURRENT_REPO` even a global variable? In general I would try to avoid such things, and instead let `initialize_github` return the repo object, then have the calling code store it in a local var, say `repo` and then pass it around as an argument. And when it comes down to this, this function is equivalent to `any(r.tag_name == tag for r in repo.get_releases())` so it might not even be necessary to have this as a dedicated function.

dev/releases/make_github_release.py
```diff
@@ -16,38 +12,44 @@
 import os
 import re
 
-# Identify GAP release version
-try:
-    GAPVERSION = utils.get_makefile_var("GAP_BUILD_VERSION")
-except:
-    utils.error("Could not get GAP version")
+if len(sys.argv) <> 2:
+    utils.error("usage: make_github_release.py <tag_name>")
```
fingolfin, line 16:
```suggestion
    utils.error("usage: "+sys.argv[0]+" <tag_name>")
```

dev/releases/make_github_release.py
```diff
@@ -16,38 +12,44 @@
 import os
 import re
 
-# Identify GAP release version
-try:
-    GAPVERSION = utils.get_makefile_var("GAP_BUILD_VERSION")
-except:
-    utils.error("Could not get GAP version")
+if len(sys.argv) <> 2:
```
fingolfin, line 15:
```suggestion
if len(sys.argv) != 2:
```

.github/workflows/release.yml
ssiccha: Done.

dev/releases/utils.py
ssiccha: > `any(r.tag_name == tag for r in repo.get_releases())`

Ooh, I didn't know that syntax. That's really nice. Yep, I'm getting rid of this function.

Wrt the `CURRENT_REPO`: I think we won't be changing repos in the middle of the script, so I've decided to use `CURRENT_REPO` as a global variable which we have to initialize once at the beginning of our script and then never change again. But I'm not married to that. If you want I can open a new PR and remove the global variable `CURRENT_REPO`. That should be easy.

dev/releases/utils.py
ssiccha: I'm not sure whether `git for-each-ref` is a plumbing tool, so just in case its output changes I'd like to check whether we found something different from `"tag"` or `"commit"` in our `is_annotated.stdout.split()[1]`. But I'll add the simplification of getting rid of `does_git_tag_exist()`

dev/releases/utils.py
ssiccha: Ah wait, it's just list comprehension inside an `any()` so you can drop the square brackets? :thinking: Anyways, I'm using it!

ssiccha: And I got rid of that function completely. Nice.

dev/releases/utils.py
ssiccha: Because I think that if the output of `git for-each-ref` would change, then this would be really confusing to debug.

fingolfin:
@ssiccha Since you are editing these files anyway, maybe also remove the `from getpass import getpass` from `utils.py`?

dev/releases/utils.py
fingolfin: git for-each-ref is explicitly a plumbing command, see https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git.html

dev/releases/utils.py
fingolfin: I just dislike global state, it makes it generally harder to reason about the program, and here it doesn't seem to add anything useful?

fingolfin: It also is the sole reason why you can't just do `import * from utils` as in all other of our files

dev/releases/utils.py
```diff
@@ -128,29 +128,39 @@ def safe_git_fetch_tags():
...
+    # check that tag points to HEAD
+    tag_commit = get_commit_of_git_tag(tag)
+    head = subprocess.run(["git", "rev-parse", "HEAD"],
+                          check=True, capture_output=True, text=True)
+    if tag_commit <> head:
+        error(f"The tag {tag} does not point to the current commit {head} but"
+              + f" instead points to {tag_commit}")
+    # check that tag is an annotated tag
+    if not is_annotated_git_tag(tag):
+        error(f"The tag {tag} is not annotated")
```
fingolfin, line 163:
```suggestion
```

ssiccha:
> Since you are editing these files anyway, maybe also remove the from getpass import getpass from utils.py?

Done

dev/releases/utils.py
ssiccha: > It also is the sole reason why you can't just do import * from utils as in all other of our files

Yeah, I wasn't aware of this until I tried it out. :/

dev/releases/utils.py
ssiccha: Aaah, that is where I find that information. Thanks! I have been looking at the man page of `git for-each-ref` and couldn't find anything!

ssiccha: Okay, I was just guessing which commands would be plumbing and which would be porcelain. No more guessing!

ssiccha:
I think this is finished. @fingolfin could you have another look?

fingolfin (approved):

dev/releases/utils.py
```diff
@@ -128,29 +127,34 @@ def safe_git_fetch_tags():
...
+
+def get_commit_of_git_tag(tag):
+    res = subprocess.run(["git", "show-ref", "--tags", "--dereference", tag],
+                            check=True, capture_output=True, text=True)
+    res = res.stdout.split('\n')
+    res = [obj for obj in res if obj.find(tag + "^{}") != -1]
+    if len(res) == 0:
+        error(f"Could not find tag {tag}")
+    res = res[0]
+    return res.split()[0]
```
fingolfin, line 145: This is needlessly complicated, though: you could just do `git rev-parse v4.11.1^{}` -- and you already do that below for `HEAD`

dev/releases/utils.py
```diff
@@ -128,29 +127,34 @@ def safe_git_fetch_tags():
...
+
```
fingolfin, line 146:
```suggestion
```

dev/releases/utils.py
```diff
@@ -128,29 +127,34 @@ def safe_git_fetch_tags():
...
+def check_git_tag_for_release(tag):
+    if not is_annotated_git_tag(tag):
+        error(f"There is no annotated tag {tag}")
+    # check that tag points to HEAD
+    tag_commit = get_commit_of_git_tag(tag)
+    head = subprocess.run(["git", "rev-parse", "HEAD"],
+                          check=True, capture_output=True, text=True)
+    head = head.stdout.strip()
```
fingolfin, line 154:
```suggestion
    tag_commit = subprocess.run(["git", "rev-parse", tag + "^{}"],
                          check=True, capture_output=True, text=True).stdout.strip()
    head = subprocess.run(["git", "rev-parse", "HEAD"],
                          check=True, capture_output=True, text=True).stdout.strip()
```

dev/releases/utils.py
ssiccha: I added your suggestions, thanks!

PR #4309, opened 2021-03-02
update_website.py: remove superfluous quotes

fingolfin:


wilfwilson (approved):
Thanks, I had been meaning to fix this but didn't get around to it 😉

PR #4256, opened 2021-02-16
some release notes entries for 4.11.1

ThomasBreuer:
The file `CHANGES.md` now contains the entries about the merged pull requests for GAP 4.11.1, categorized in the same way as for GAP 4.11.0.

What is now *MISSING* is the information about new and updated GAP packages in 4.11.1, compared to 4.11.0.

~~... currently unstructured, taken from the subset of pull request obtained today with the query~~

~~https://github.com/gap-system/gap/pulls?q=is%3Apr+-label%3A%22release+notes%3A+not+needed%22+-label%3A%22release+notes%3A+added%22+is%3Aclosed+merged%3A2019-09-10..2033-12-31+base%3Amaster+is%3Amerged~~

~~(It remains to deal with the entries which are still returned by this query. Those entries which are proposed here for addition to `CHANGES.md` do no longer match the query because now they have the label "release notes: added".)~~

PR #4310, opened 2021-03-02
CHANGES.md: list of package updates now is a Markdown list

fingolfin:
Otherwise it renders quite ugly on https://github.com/gap-system/gap/blob/master/CHANGES.md

UPDATE: @alex-konovalov tells me this may be deliberate. OK, also fine by me, but I hope someone can confirm :-)


ThomasBreuer (approved):
From my point of view, the list looks better, and I did not intend to show the updated packages as "running test". In fact, the corresponding list shown in #4256 is rendered like a list there, without actually being a Markdown list.

PR #4315, opened 2021-03-05
Make CI action upload manuals

ssiccha:
Thank You!

# Description

## Text for release notes 

is suitable as title.

in the root directory.

## Further details

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -189,16 +189,25 @@ jobs:
         run: ${{ matrix.extra }} bash dev/ci-prepare.sh
       - name: "Run tests"
         run: ${{ matrix.extra }} bash dev/ci.sh
-      - name: "Upload manuals"
+      - name: "Upload pdf manuals"
         if: ${{ matrix.test-suites == 'makemanuals' }}
         uses: actions/upload-artifact@v2
         with:
-          name: manuals
+          name: manuals-pdf
           path: |
             doc/dev/manual.pdf
             doc/hpc/manual.pdf
             doc/ref/manual.pdf
             doc/tut/manual.pdf
+      - name: "Upload html manuals"
+        if: ${{ matrix.test-suites == 'makemanuals' }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: manuals-html
+          path: |
+            doc/*/*.html
+            doc/*/*.css
+            doc/*/*.js
       - name: "Gather coverage"
         run: ${{ matrix.extra }} bash dev/ci-gather-coverage.sh
       - name: "Upload coverage data to codecov"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -189,6 +189,16 @@ jobs:
         run: ${{ matrix.extra }} bash dev/ci-prepare.sh
       - name: "Run tests"
         run: ${{ matrix.extra }} bash dev/ci.sh
+      - name: "Upload manuals"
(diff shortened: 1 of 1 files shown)


ssiccha:
I think this is finished now. Let's see whether the action succeeds.

ssiccha:
So the makemanuals action uploads 4 files, that is the `dev`, `hpc`, `ref` and `tut` manuals:
https://github.com/gap-system/gap/runs/2038672727?check_suite_focus=true

But I think I can't see the artifacts until all jobs have finished.

ssiccha:
I've checked the artifacts. I was able to open them all.

wilfwilson:
This seems to work as intended. I typically prefer looking at the HTML, rather than the PDF version. Is there a particular reason not to include the `.html` files too? (and the relevant `css`/`js` files)

ssiccha:
@wilfwilson only that I don't know which ones they are. :) I'm having a look.

ssiccha:
I think this should do the trick. Do we definitely need the *.js files? (I have no idea :))

wilfwilson:
I assume that we need the JS files for the HTML version of the manuals if and only if the JS files should exist at all. So: probably, yes.

I don’t have time to investigate now, but something is wrong:
```
Warning: No files were found with the provided path: doc/dev/*.{html,css,js}
doc/hpc/*.{html,css,js}
doc/ref/*.{html,css,js}
doc/tut/*.{html,css,js}. No artifacts will be uploaded.
```

fingolfin:
I am fine with uploading the HTML, but note that the user still needs to download them and the view them in a browser; the PDF is preferable for me only because I can view it on my iPad or iPhone, unlike the HTML+js+css files. But of course offering them all may be the best solution:-)

.github/workflows/CI.yml
```diff
@@ -189,6 +189,26 @@ jobs:
...
+            doc/tut/manual.pdf
+      - name: "Upload html manuals"
+        if: ${{ matrix.test-suites == 'makemanuals' }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: manuals-html
+          path: |
+            doc/dev/*.{html,css,js}
+            doc/hpc/*.{html,css,js}
+            doc/ref/*.{html,css,js}
```
fingolfin, line 210: This didn't work - I think bash-style `{...}` patterns are not supported. But OTOH you could maybe use four patterns of the form `doc/*/*.html`

wilfwilson: That kind of thing seems to work fine, see e.g. the result https://github.com/wilfwilson/gap/runs/2057072844?check_suite_focus=true of commit https://github.com/wilfwilson/gap/commit/113226445d62c82f5805d33a69594fb7c1fc9474

The log gives:
```
With the provided path, there will be 290 files uploaded
Total file count: 290 ---- Processed file #276 (95.1%)
Total size of all the files uploaded is 3459174 bytes
Finished uploading artifact manuals-html. Reported size is 3459174 bytes. There were 0 items that failed to upload
Artifact manuals-html has been successfully uploaded!
```
You'd think that would make the archive be about 3.5 MB, but it's actually 17.8 MB for some reason. Still acceptable though, I think?

ssiccha: Just updated my PR. IMO the 3.5MB vs 17.8MB is not a problem. We only do that once per run of the CI workflow.

wilfwilson (approved):

PR #4320, opened 2021-03-15
FIX: Outer automorphisms of certain Chevalley groups.

### LINKED ISSUE #4318: Error when computing automorphisms of simple orthogonal groups
### Observed behaviour

```
gap> G := PrimitiveGroup(119, 1);
O-(8, 2)
gap> S := SymmetricGroup(119);
Sym( [ 1 .. 119 ] )
gap> IsSubgroup(S, G);
true
gap> N := Normalizer(S, G);;
gap> StructureDescription(N);
"O-(8,2)"
gap> Index(N, G);
1
gap> T := TwoClosure(G);
<permutation group with 32 generators>
gap> StructureDescription(T);
"O-(8,2) : C2"
gap> Index(T, G);
2
```

### Expected behaviour

`G` is a subgroup of index 2 in `T`, so `T` must lie in the normalizer `N`.

`NormalizerPermGroup` outputs a correct answer:
```
gap> N2 := NormalizerPermGroup(S, G);;
gap> Index(N2, G);
2
```

### GAP banner
```
 ┌───────┐   GAP 4.10.0 of 01-Nov-2018
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64
 Configuration:  gmp 6.1.2, readline
 Packages:   Alnuth 3.1.0, AtlasRep 1.5.1, AutPGrp 1.10, CTblLib 1.2.2, FactInt 1.6.2, GAPDoc 1.6.2, IO 4.5.4, Polycyclic 2.14, PrimGrp 3.4.0, 
             SmallGrp 1.3, TomLib 1.2.7, TransGrp 2.0.4
```

The same problem is present in the current dev version 4.12dev-731-gc7a0c6e.

hulpke:
Instead of tables in ATLAS (that had been misinterpreted), use the
explict presentations in Bray/Holt/Roney-Dougal.

This can lead to wrong results when calculating normalizers in S_n.

The groups affected are of type O_n-(q), n>=8 even
(and to a lesser extent -- wrong outer automorphism group structure, but that is not used in other places for deductions -- S4(4^e) )

This fixes #4318

### DIFF
--- a/grp/simple.gi
+++ b/grp/simple.gi
@@ -887,9 +887,90 @@ local id;
   return DataAboutSimpleGroup(id);
 end);
 
+# Tables for outer automorphisms from Bray/Holt/Roney-Dougal, p. 36/37
+BindGlobal("OuterAutoSimplePres",function(class,n,q)
+local p,e,f,rels,gp;
+  class:=UppercaseString(class);
+  p:=SmallestPrimeDivisor(q);
+  e:=LogInt(q,p);
+  if class="L" and n=2 then
+    f:=FreeGroup("d","f");
+    rels:=StringFormatted("d{}=f{}=[d,f]=1",Gcd(q-1,2),e);
+  elif class="L" and n>=3 then
+    f:=FreeGroup("d","g","f");
+    rels:=StringFormatted("d{}=g2=f{}=[g,f]=1,d^g=D,d^f=d{}",Gcd(q-1,n),e,p);
+  elif class="U" and n>=3 then
+    f:=FreeGroup("d","g","f");
+    rels:=StringFormatted("d{}=g2=1,f{}=g,d^g=D,d^f=d{}",Gcd(q+1,n),e,p);
+  elif class="S" and n>=2 and (n<>4 or p<>2) then
+    f:=FreeGroup("d","f");
+    rels:=StringFormatted("d{}=f{}=[d,f]=1",Gcd(q-1,2),e);
+  elif class="S" and n=4 and p=2 then
+    f:=FreeGroup("g","f");
+    rels:=StringFormatted("g2=f,f{}=1",e);
+  elif (class="O" or class="O0" or class="OO") and n>=3 then
+    # special case of characteristic 2 -- copy S status
+    if p=2 then
+      f:=FreeGroup("d","f");
+      rels:=StringFormatted("d{}=f{}=[d,f]=1",Gcd(q-1,2),e);
+    else
+      f:=FreeGroup("d","f");
+      rels:=StringFormatted("d2=f{}=[d,f]=1",e);
+    fi;
+  elif class="O+" and n>=6 and IsEvenInt(n) and n<>8 and p=2 then
+    f:=FreeGroup("g","f");
+    rels:=StringFormatted("g2=f{}=[g,f]=1",e);
+  elif class="O+" and n=8 and p=2 then
+    f:=FreeGroup("t","g","f");
+    rels:=StringFormatted("t^3=g2=(gt)2=f{}=[t,f]=[g,f]=1",e);
+  elif class="O-" and n>=4 and IsEvenInt(n) and p=2 then
+    f:=FreeGroup("g","f");
+    rels:=StringFormatted("g2,f{}=g",e);
+  elif class="O+" and n=8 and IsOddInt(p) then
+    f:=FreeGroup("p","t","g","d","f");
+    rels:=StringFormatted(
+      "p2=t3=g2=(gt)2=d2=1,d^t=p,p^dp,(dg)2=p,f{}=[d,f]=[t,f]=[g,f]=1",e);
+  elif class="O+" and n>=12 and n mod 4=0 and IsOddInt(q) then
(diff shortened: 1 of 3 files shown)


grp/simple.gi
```diff
@@ -887,9 +887,101 @@ local id;
...
+    rels:=Concatenation("d",String(Gcd(q-1,2)),
+      "=f",String(e),"=[d,f]=1");
+  elif class="L" and n>=3 then
+    f:=FreeGroup("d","g","f");
+    rels:=Concatenation("d",String(Gcd(q-1,n)),
+      "=g2=f",String(e),"=[g,f]=1,d^g=D,d^f=d",String(p));
+  elif class="U" and n>=3 then
+    f:=FreeGroup("d","g","f");
+    rels:=Concatenation("d",String(Gcd(q+1,n)),
+      "=g2=1,f",String(e),"=g,d^g=D,d^f=d",String(p));
```
wilfwilson, line 907: This is not a request to make any changes, just a note in case you're not aware: there is a fairly-new function `StringFormatted` in GAP, which I now tend to use in situations like this, where I'm constructing a string from various parts. Depending on the context and the code in question, it can sometimes make the result more readable (and yes, it can sometimes make the result less readable).
```suggestion
    rels:=StringFormatted("d{}=g2=1, f{}=g, d^g=D, d^f=d{}", Gcd(q+1,n), e, p);
```

grp/simple.gi
```diff
@@ -887,9 +887,101 @@ local id;
   return DataAboutSimpleGroup(id);
 end);
 
+# Tables for outer automorphisms from Bray/Holt/Roney-Dougal, p. 36/37
```
wilfwilson, line 890: Thanks for including this line.

I have no idea how we deal with referencing inGAP, in general (and whether we have any kind of 'policy' about it): so for anyone who knows about this sort of sutff, should there also be a corresponding reference to this book in the GAP manual/bibliography/website/somewhere else?

wilfwilson (approved):
Thanks for making this fix @hulpke.

Unsurprisingly the failing 32-bit GitHub Actions job checks is nothing to do with you.

grp/simple.gi
hulpke: Thank you -- indeed that will make it much nicer. I will change this later today once I'm done with classes.

grp/simple.gi
fingolfin: In practice there is no consistent way this is dealt with, I am afraid. Of course in principle it would be good to at least add a BibTeX entry for this in our master .bib file. Whether and where to cite it is a different question, though...

grp/simple.gi
```diff
@@ -887,9 +887,90 @@ local id;
...
+BindGlobal("OuterAutoSimplePres",function(class,n,q)
+local p,e,f,rels,gp;
+  class:=UppercaseString(class);
+  p:=Factors(q)[1];
```
fingolfin, line 894:
```suggestion
  p:=SmallestPrimeDivisor(q);
```

grp/simple.gi
```diff
@@ -887,9 +887,90 @@ local id;
...
+local
+nam,e,EFactors,par,expo,prime,result,aut,i,classical,classaut,shortname;
+
+  shortname:=function(gp)
+  local s;
+    if IsCyclic(gp) then
+      return String(Size(gp));
+    elif Size(gp)<=31 or Size(gp) in [33..47] then
+      s:=StructureDescription(gp);
+      s:=Filtered(s,x->not x in "C ");
```
fingolfin, line 968: What exactly is this test about? It seems to remove all spaces and all C from the structure description? The result won't be in stringrep, though. How about this instead?
```suggestion
      s:=RemoveCharacters(s, "C ");
```

grp/simple.gi
```diff
@@ -1115,8 +1204,10 @@ local nam,e,EFactors,par,expo,prime,result,aut,i;
       e:=EFactors(Gcd(4,par[2]^par[1]-1),expo,2);
     fi;
   elif id.series="2D" then
+    classical:=["O-",2*par[1],par[2]];
     nam:=Concatenation("O",String(par[1]*2),"-(",String(par[2]),")");
-    e:=EFactors(Gcd(4,par[2]^par[1]+1),expo/2,1);
+
+    e:=EFactors(Gcd(4,par[2]^par[1]+1),2*expo,1);
```
fingolfin, line 1210: Why the change from /2 to *2 -- is that part of the bug fix?

fingolfin (approved):
Fine by me. Some minor questions

grp/simple.gi
hulpke: This is about producing names for subgroups of outer automorphism groups that is consistent with what is used in the table of marks library (to find the group there). E.g. `StructureDescription` gives `C2 x C2`, while the table of marks (and character table) libraries use .2x2

grp/simple.gi
hulpke: Yes. That was the underlying bug. However this (slightly cryptic) variant is phased out now in favour of the presentations from Bray/Holt/Roney-Dougal, and currently both are checked against each other to avoid further implementation errors.

PR #4324, opened 2021-03-19
Improve DataAboutSimpleGroup

ssiccha:
This commit does two things:
- it turns an Error into an InfoWarning
- it changes the return value of the subfunction EFactors to fail.

The latter should make EFactors be more in line with how
DataAboutSimpleGroup handles the case "simple group tom
nonidentified/not yet done", where it assigns fail to the
components `nam` and `e`.

Turning the `return false;` into `return fail;` should not break any
existing code because of the following. Previously, `e` could have the
values fail or also, if the user `return;`s on the `Error` this commit
deletes, false (or the extensions of course). Thus previously a user had
to check for both false and fail.

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/grp/simple.gi
+++ b/grp/simple.gi
@@ -1000,8 +1000,8 @@ nam,e,EFactors,par,expo,prime,result,aut,i,classical,classaut,shortname;
               [ 24, "S4" ] ];
         return dd;
       else
-        Error("mixed triality not yet done");
-        return false;
+        Info(InfoWarning, "DataAboutSimpleGroup: mixed triality not yet done");
+        return fail;
       fi;
     fi;
 


hulpke (approved):
No objections, but the cuirrent status of the function is that it was patched to correct an error. I am planning, now that there is an alternate source for the outer automorphisms that has been checked, to gut the EFactors function (and maybe just store the three numbers to compare the order and offer them foir calculation of the multiplier). This code would then probably go the way of all the earth.

fingolfin:
So... merge this or not?

Does this PR fix any specific issue? Or what other benefit does it bring?

ssiccha:
We want to use this function in #4100 and just stumbled over the fact that sometimes you need to be at the console in order to continue the computation.

PR #4325, opened 2021-03-20
fixed method ordering for `IsomorphismPermGroup`

ThomasBreuer:
When starting GAP without packages, currently I get the following.

```
gap> meths:= ApplicableMethod( IsomorphismPermGroup, [ SO(1,8,2) ], 1, "all" );;
#I  Searching Method for IsomorphismPermGroup with 1 arguments:
#I  Total: 16 entries
#I  Method 10: ``IsomorphismPermGroup: matrix group'' at /.../lib/grpmat.gi:520 , value: 55
#I  Skipped:
#I  Method 11: ``IsomorphismPermGroup: via niceomorphisms'' at /.../lib/grpnice.gi:572 , value: 54
[...]
```

Starting GAP with autoloaded packages yields the following:

```
gap> meths:= ApplicableMethod( IsomorphismPermGroup, [ SO(1,8,2) ], 1, "all" );;
#I  Searching Method for IsomorphismPermGroup with 1 arguments:
#I  Total: 16 entries
#I  Method 10: ``IsomorphismPermGroup: via niceomorphisms'' at /.../lib/grpnice.gi:572 , value: 56
#I  Skipped:
#I  Method 11: ``IsomorphismPermGroup: matrix group'' at /.../lib/grpmat.gi:520 , value: 55
[...]
```

(This difference was not present in GAP 4.10, see below.)

The comments for the `IsomorphismPermGroup` method in `lib/grpnice.gi` say that the method from `lib/grpmat.gi` is regarded as better; for example, it often yields permutation groups of smaller degree.
(One could argue that the method from `lib/grpnice` could achieve this if `NiceMonomorphism` would call `NicomorphismOfGeneralMatrixGroup` instead of the "falling back on GL" method, but this call would be more expensive.  It depends on the later computations what is better.)

The reason for this unwanted difference is as follows:

- The method from `grpnice.gi` requires `IsGroup and IsFinite`.
- The method from `grpmat.gi` does not require `IsGroup and IsFinite`.
- A package (`Polycyclic`, but this does not matter) declares a new property   and an implication from `IsGroup and IsFinite` to this new property,   with the effect that the rank of `IsGroup and IsFinite` grows by 2   (1 for the tester filter, 1 for the getter filter of the new property).

```
gap> RankFilter( IsGroup and IsFinite );
45
gap> DeclareProperty( "IsJustForRankingUp", IsGroup );
gap> InstallTrueMethod( IsJustForRankingUp, IsGroup and IsFinite );
gap> RankFilter( IsGroup and IsFinite );
47
```

- After loading packages, GAP adjusts the method installations according   to the current ranks, using `ResumeMethodReordering`.
  This has the effect that the method from `grpnice.gi` gets ranked higher than the method from `grpmat.gi`.
  (This feature was introduced in GAP 4.11.0, thus the problem was not yet present in GAP 4.10.)

### DIFF
--- a/lib/grpmat.gi
+++ b/lib/grpmat.gi
@@ -517,8 +517,29 @@ local   nice,img,module,b;
   return nice;
 end );
 
-InstallMethod( IsomorphismPermGroup,"matrix group", true,
-  [ IsMatrixGroup ], 10,
+
+#############################################################################
+##
+#M  IsomorphismPermGroup( <mat-grp> )
+##
+##  We want to use a method based on 'NicomorphismOfGeneralMatrixGroup'
+##  for the finite matrix group <mat-grp>.
+##
+##  If <mat-grp> does not know whether it is finite,
+##  there is currently no concurrent method,
+##  thus no rank shift is necessary in the first method installation.
+##
+##  If <mat-grp> knows to be finite, we want our method to beat the one for
+##  'IsGroup and IsFinite and IsHandledByNiceMonomorphism',
+##  so we install it with the same rank shift as that method.
+##  (Note that 'IsHandledByNiceMonomorphism' is implied by
+##  'IsMatrixGroup and IsFinite'.)
+##
+##  Installing just one method (with requirement 'IsMatrixGroup') would be
+##  difficult since we do not know the rank of 'IsGroup and IsFinite'.
+##
+InstallMethod( IsomorphismPermGroup,"matrix group",
+  [ IsMatrixGroup ],
 function(G)
 local map;
   if HasNiceMonomorphism(G) and IsPermGroup(Range(NiceMonomorphism(G))) then
@@ -537,6 +558,27 @@ local map;
     return map;
   fi;
 end);
+
+InstallMethod( IsomorphismPermGroup,
+    "finite matrix group",
+    [ IsMatrixGroup and IsFinite ], 5-NICE_FLAGS,
+    function( G )
+    local map;
+
+    if HasNiceMonomorphism( G ) and
+       IsPermGroup( Range( NiceMonomorphism( G ) ) ) then
+      map:= NiceMonomorphism( G );
+      if IsIdenticalObj( Source( map ), G ) then
+        return map;
+      fi;
+      return GeneralRestrictedMapping( map, G, Image( map, G ) );
+    else
+      map:= NicomorphismOfGeneralMatrixGroup( G, false, false );
+      SetNiceMonomorphism( G, map );
+      return map;
+    fi;
+end );
+
 
 #############################################################################
 ##
--- a/lib/grpmat.gi
+++ b/lib/grpmat.gi
@@ -527,19 +527,23 @@ end );
 ##
 ##  If <mat-grp> does not know whether it is finite,
 ##  there is currently no concurrent method,
-##  thus no rank shift is necessary in the first method installation.
+##  thus no rank shift is necessary for the method installation.
 ##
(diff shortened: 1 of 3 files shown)


lib/grpmat.gi
```diff
@@ -517,8 +517,29 @@ local   nice,img,module,b;
...
+##  thus no rank shift is necessary in the first method installation.
+##
+##  If <mat-grp> knows to be finite, we want our method to beat the one for
+##  'IsGroup and IsFinite and IsHandledByNiceMonomorphism',
+##  so we install it with the same rank shift as that method.
+##  (Note that 'IsHandledByNiceMonomorphism' is implied by
+##  'IsMatrixGroup and IsFinite'.)
+##
+##  Installing just one method (with requirement 'IsMatrixGroup') would be
+##  difficult since we do not know the rank of 'IsGroup and IsFinite'.
```
fingolfin, line 539: Isn't that what we have dynamic rank adjustment for? I.e., instead of a static rank adjustment like `10`, one can use a no-argument function which is used to recompute the rank adjustment each time we have to reorder methods due to new implications; e.g. `{} -> RankFilter(IsFinite and IsMatrixGroup)` is used in `lib/grpnice.gi:159`.

hulpke (approved):
Thank you for spotting and investigating this. Indeed, when the two methods were created, I checked that the ordering was as desired, but did not think about that another implication could change this.

lib/grpmat.gi
ThomasBreuer: Yes, one could use this feature.
Thus I should reformulate the sentence, because I do not think that this would be a good solution.

I think that the source of the problem is the current method with requirement `IsMatrixGroup`:
On the one hand, it has a low rank.
On the other hand, it shall be preferred to the method with requirement `IsGroup and IsHandledByNiceMonomorphism` in the situation that this is applicable, which typically happens when the argument is in `IsMatrixGroup and IsFinite`.

Upranking the `IsMatrixGroup` is not fair (compared to other methods) in the general case, and downranking the `IsGroup and IsHandledByNiceMonomorphism` method just because of the desired relation to the `IsMatrixGroup` method is also not fair.
A better solution would be to install the two methods without incremental ranks, and the method that is believed to be better in the concurrence situation should be installed explicitly for this situation.

(I did not dare to remove the extreme downranking of the `IsGroup and IsHandledByNiceMonomorphism` method. We can make sure that such a change does not mess up the code from the GAP distribution, but other GAP code might be affected.
Also, it is perhaps a mistake not to install the same method for the intersection situation `IsMatrixGroup and `IsHandledByNiceMonomorphism`. Should I change the pull request in this direction?)

lib/grpmat.gi
```diff
@@ -538,6 +559,27 @@ local map;
   fi;
 end);
 
+InstallMethod( IsomorphismPermGroup,
+    "finite matrix group",
+    [ IsMatrixGroup and IsFinite ], 5-NICE_FLAGS,
+    function( G )
+    local map;
```
fingolfin, line 566: Instead of duplicating this code via copy&paste, how about instead moving storing it as its own function (`BindGlobal("IsomorphismPermGroupForMatrixGroup", function ... end);`) and then we can call `InstallMethod` twice for it?

fingolfin (approved):
Fine by me, thanks for the explanation.

lib/grpmat.gi
ThomasBreuer: Yes. I had sketched this in an earlier comment.

wilfwilson (approved):
Thank you Thomas.

PR #4326, opened 2021-03-21
Add DirectProductOfPermGroupsWithMovedPoints helper

fingolfin:
This is basically what DirectProduct does for a set of permutation groups,
except that one can override the moved points for each factor. This is useful
if one needs to compute "sums of G-sets": If e.g. we have a trivial group
acting on 3 points, and a Sym(4), then DirectProduct just gives back the
Sym(4). But if we need the result to act on 3+4=7 points, with the Sym(4)
acting on points acting on [4..7], then this new helper allows us to do that.

For now, I left this deliberately undocumented, as it is quite niche.

### DIFF
--- a/tst/testinstall/gprd.tst
+++ b/tst/testinstall/gprd.tst
@@ -46,4 +46,11 @@ gap> HasIsFinite(d4) and not IsFinite(d4);
 true
 
 #
+gap> d1 := DirectProductOfPermGroupsWithMovedPoints([g1, g1, g1], [[1..4], [1..4], [1..4]]);;
+gap> MovedPoints(d1);
+[ 1, 2, 3, 5, 6, 7, 9, 10, 11 ]
+gap> HasIsFinite(d1) and IsFinite(d1);
+true
+
+#
 gap> STOP_TEST("gprd.tst");
--- a/lib/gprdperm.gi
+++ b/lib/gprdperm.gi
@@ -12,20 +12,12 @@
 
 #############################################################################
 ##
-#M  DirectProductOp( <grps>, <G> )  . . . . . . direct product of perm groups
+#F  DirectProductOfPermGroupsWithMovedPoints( <grps>, <pnts> )
 ##
-InstallMethod( DirectProductOp,
-    "for a list of permutation groups, and a permutation group",
-    IsCollsElms,
-    [ IsList and IsPermCollColl, IsPermGroup ], 0,
-    function( grps, G )
-    local   oldgrps,  olds,  news,  perms,  gens,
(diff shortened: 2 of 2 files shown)


ThomasBreuer (approved):
Perhaps `Immutable` should be called on the entries of `olds`, in order to have independent lists inside the (mutable) `DirectProductInfo` record.

tst/testinstall/gprd.tst
```diff
@@ -45,5 +45,12 @@ gap> d4 := DirectProduct(g1, g2, g3);;
 gap> HasIsFinite(d4) and not IsFinite(d4);
 true
 
+#
+gap> d1 := DirectProductOfPermGroupsWithMovedPoints([g1, g1, g1], [[1..4], [1..4], [1..4]]);;
+gap> MovedPoints(d1);
+[ 1, 2, 3, 5, 6, 7, 9, 10, 11 ]
+gap> HasIsFinite(g1) and IsFinite(g1);
```
wilfwilson, line 52: Is this what you meant?
```suggestion
gap> HasIsFinite(d1) and IsFinite(d1);
```

fieker (approved):
As far as I understand this, this indeed will solve my problem (not only mine I hope)
I think the IsFinite(d1) ca IsFinite(g1) is correct (but irrelevant)

tst/testinstall/gprd.tst
fingolfin: Oops, fixed

fingolfin:
@wilfwilson fixed the tests, thanks
@ThomasBreuer added a commit which makes some stuff immutable
@fieker indeed it should.

wilfwilson (approved):

ThomasBreuer:
The test failure in `gap/tst/testbugfix/2006-08-28-t00151.tst:14` looks strange.
My first idea was that garbage collections could distort the timings, but they should not need the major part of the time.
(Are such timing tests safe at all?)

wilfwilson:
@ThomasBreuer I know that @hulpke encountered the same failure while working on his PR #4320, which is why he made a change in that PR (now in 52fa8c9ab87121a3ec4b8a072194f694e4e2675a) to increase the work done by this test, in an attempt to make it more robust.

wilfwilson:
The same kind of test failure was also encountered in #4332; see https://github.com/gap-system/gap/pull/4332#issuecomment-805911954. Let's keep an eye out for this happening again.

wilfwilson:
Since this is (as @fingolfin said) a niche feature backend feature, without documentation, I don't think this needs to be mentioned in the release notes.

PR #4327, opened 2021-03-21
Fix InvariantQuadraticForm for Omega(-1, 2*d, 2^n)

### LINKED ISSUE #4323: More problems with orthogonal groups ...
I am trying to get a solution for the problem from issue #500, in the direction of what @frankluebeck, @fingolfin, and @stevelinton had suggested.
The problem seems to be that the quadratic forms stored in the groups `Omega( -1, 2m, 2^n )` are wrong.
Or do I misunderstand something?
```
gap> m:= 4;; q:= 2;;
gap> g:= Omega( -1, m, q );;
gap> gens:= GeneratorsOfGroup( g );;
gap> Q:= InvariantQuadraticForm( g ).matrix;;   # the stored form
gap> ForAll( GF(q)^m, v -> ForAll( gens, g -> v*Q*v = (v*g)*Q*(v*g) ) );
false
gap> Q:= [[0,0,0,1],[0,1,1,0],[0,0,1,0],[0,0,0,0]] * Z(2);;  # the correct form
gap> ForAll( GF(q)^m, v -> ForAll( gens, g -> v*Q*v = (v*g)*Q*(v*g) ) );
true
gap> m:= 6;; q:= 2;;
gap> g:= Omega( -1, m, q );;
gap> gens:= GeneratorsOfGroup( g );;
gap> Q:= InvariantQuadraticForm( g ).matrix;;   # the stored form
gap> ForAll( GF(q)^m, v -> ForAll( gens, g -> v*Q*v = (v*g)*Q*(v*g) ) );
false
gap> Q:= [[0,0,0,0,0,1],[0,0,0,0,1,0],[0,0,1,1,0,0],[0,0,0,1,0,0],[0,0,0,0,0,0],[0,0,0,0,0,0]] * Z(2);;  # the correct form
gap> ForAll( GF(q)^m, v -> ForAll( gens, g -> v*Q*v = (v*g)*Q*(v*g) ) );
true
```

fingolfin:
Also add proper tests to check that the quadratic forms are indeed correct
(the existing tests were too weak).

Fixes #4323 

Some background: We didn't even store the `InvariantQuadraticForm` for these groups up to GAP 4.9; it was only added in 4.10, via my PR #2577. In there, I also added tests to verify the quadratic form, but those tests were insufficient (rather obviously so, in retrospect *sigh*). This adds a test similar to what Thomas used to highlight the issue in his bug report.

### DIFF
--- a/grp/classic.gi
+++ b/grp/classic.gi
@@ -1789,10 +1789,8 @@ BindGlobal( "OmegaMinus", function( d, q )
       x[i,d-i+1] := o;
     od;
     x[m,d-m+1] := -nu - nubar;
-    if q mod 2 = 1 then
-      x[m,d-m] := -o;
-      x[m+1,d-m+1] := xi^( (q+1)/2 );
-    fi;
+    x[m,d-m] := -o;
+    x[m+1,d-m+1] := -xi;
     x:= ImmutableMatrix( f, x, true );
     SetInvariantQuadraticFormFromMatrix( g, x );
 
--- a/tst/testinstall/grp/classic-forms.tst
+++ b/tst/testinstall/grp/classic-forms.tst
@@ -21,10 +21,14 @@ gap> CheckBilinearForm := function(G)
 >               g -> g*M*TransposedMat(g) = M);
 > end;;
 gap> CheckQuadraticForm := function(G)
->   local M, Q;
+>   local M, Q, V, vecs;
 >   M := InvariantBilinearForm(G).matrix;
 >   Q := InvariantQuadraticForm(G).matrix;
->   return Q+TransposedMat(Q) = M;
+>   V := FieldOfMatrixGroup(G)^DegreeOfMatrixGroup(G);
+>   vecs:=List([1..100], i->Random(V));
+>   return (Q+TransposedMat(Q) = M) and ForAll(vecs,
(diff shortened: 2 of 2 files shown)


ThomasBreuer (approved):
Thanks.
Is there anything one can say why the matrix of the quadratic form in `OmegaMinus` can look like this? (The structure of the generators yields that one has to work only for d = 4.)
In `CheckQuadraticForm`, checking the relation between the matrices of the bilinear and the quadratic form still makes sense, it is just not sufficient in even characteristic.

hulpke:
> Thanks.
> Is there anything one can say why the matrix of the quadratic form in `OmegaMinus` can look like this? (The structure of the generators yields that one has to work only for d = 4.)

I believe the generators are taken from `Rylands, Taylor: Matrix generators for the Orthogonal Group, JSC 25 (1998), 351--360, and p. 359 (about the middle) gives the form for Omega-

fingolfin:
@ThomasBreuer I restored the check for the relation between the matrices of the bilinear and the quadratic form -- I did not mean to disable it in the first place, thanks for spotting this!

fingolfin:
@ThomasBreuer ah that's of course much nicer. Pushed it.

PR #4331, opened 2021-03-23
Speed up building stabchain from base and strong generators

ChrisJefferson:
When we already have a base + strong generating set, the function StabChainBaseStrongGenerators turns this into a GAP stabilizer chain. This function scaled surprisingly badly, because it used existing functions for adding to a stabilizer chain which have "O(n^2) in number of generator" loops.

Long term it would be good to clean those up, but in this case, where we have a list of strong generators, we can just construct the stabilizer chain.

A brief benchmark:

`sc := StabChainBaseStrongGenerators([1..1000], List([1,3..999], x -> (x,x+1)));;` is sped up from 48 seconds to 45 milliseconds (that's not a unit mistake, it is 1000 times faster).

I test this with the existing tests for stabiliser chains, which use the fact that stabiliser chains are used for enumerating groups.

This includes another fix (which I could pull out) -- the function PlainListCopy didn't actually produce a plain list for types like ranges, which I needed to fix as the internal AGEST only accepts PLISTs.

### DIFF
--- a/tst/testinstall/stabchain.tst
+++ b/tst/testinstall/stabchain.tst
@@ -1,7 +1,7 @@
 #@local TestGens,m
 gap> START_TEST("stabchain.tst");
 gap> TestGens := function(g)
-> local m;
+> local m, sc;
 > m := SortedList(List(g));
 > if Size(m) <> Size(g) or Size(Set(m)) <> Size(m) or
 >    Group(m) <> g then Print("0"); fi;
@@ -14,10 +14,19 @@ gap> TestGens := function(g)
 > ChangeStabChain(StabChainMutable(g), [1,10,11,12,13,14,15,2,3,4,5,6,7,8], false);
 > if m <> SortedList(List(g)) then Print("5"); fi;
 > if m <> SortedList(List(g, x -> x)) then Print("4"); fi;
+> sc := StabChainBaseStrongGenerators([1,10,11,12,13,14,15,2,3,4,5,6,7,8], 
+>  StrongGeneratorsStabChain(StabChainMutable(g)), ());
+> g := Group(GeneratorsOfGroup(g), ());
+> SetStabChainMutable(g, sc);
+> if m <> SortedList(List(g)) then Print(g,"6"); fi;
+> if m <> SortedList(List(g, x -> x)) then Print(g,"7"); fi;
 > return true;
 > end;;
 gap> TestGens(Group(()));;
 gap> List([2..7],
 >       x -> List([1..NrTransitiveGroups(x)],
 >         y -> TestGens(TransitiveGroup(x,y))));;
+gap> TestGens(Group((1,2,3),(4,5,6)));;
+gap> TestGens(Group((1,2,3)(4,5,6)));;
+gap> TestGens(Group((2,4,6),(1,3,5),(1,3)));;
 gap> STOP_TEST("stabchain.tst", 1);
--- a/tst/testinstall/stabchain.tst
+++ b/tst/testinstall/stabchain.tst
@@ -29,4 +29,10 @@ gap> List([2..7],
 gap> TestGens(Group((1,2,3),(4,5,6)));;
 gap> TestGens(Group((1,2,3)(4,5,6)));;
 gap> TestGens(Group((2,4,6),(1,3,5),(1,3)));;
+gap> m := StabChainBaseStrongGenerators(
+>    [1,3..1999],
+>    List([1,3..1999], x -> (x,x+1)),
+>    ());;
+gap> Log(SizeStabChain(m), 2) = 1000;
+true
 gap> STOP_TEST("stabchain.tst", 1);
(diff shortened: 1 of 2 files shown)


wilfwilson (commented):
Unsurprisingly, touching this code has had consequences! 🙈  I hope they're not too hard to iron out.

Although I'm not a stickler for this kind of stuff, I think it would be best if the `PlainListCopy` stuff was in a separate PR. But I don't mind. In any case, although `PlainListCopy` (and `PlainListCopyOp`) have some documentation in the `lib/list.gd` file, this documentation isn't linked into the manual, but in my opinion that would be helpful (probably in `Section 21.24 Plain Lists`, at the very end of `Chapter 21: Lists`). If you wouldn't mind adding that, I'd appreciate it.

I certainly support the main change you've made to `StabChainBaseStrongGenerators`, especially given the performance problems we were encountering yesterday. My main suggestion is that you should include your 1000-fold speedup example in a test file (perhaps made even bigger) to make a regression super obvious.

Note that this example includes a redundant base. Although that's not forbidden, maybe you'd prefer to give an irredundant base (i.e. use the same range in both places):
```
sc := StabChainBaseStrongGenerators([1,3..999], List([1,3..999], x -> (x,x+1)), ());;
```
Fingers crossed the tests pass this time.

lib/list.gi
```diff
@@ -3916,10 +3916,13 @@ InstallGlobalFunction(PlainListCopy, function( list )
     tnum := TNUM_OBJ(copy);
     if FIRST_LIST_TNUM > tnum or LAST_LIST_TNUM < tnum then
         copy := PlainListCopyOp( copy );
+    elif FIRST_PLIST_TNUM > tnum or LAST_PLIST_TNUM < tnum then
+        copy := PLAIN_LIST_COPY( copy );
```
fingolfin, line 3920: I think there's more wrong with `PlainListCopy`, and I am not sure what the original intention for this function were -- it's possible that the author meant "plain list" as "native GAP list", which would explain the current code. (The original authors turns out to be @stevelinton who added it 1998-08-28 in the "old repository commit b6f3672c8a5b59f7248c8b3e8d5f2402ff4bca9c").

But, OK, I agree we probably want "real" plain lists (but let's inspect existing callers first, to make sure?)

Also, looking closer at `PlainListCopyOp` methods there are, several of those seem to be *wrong*. E.g.
```gap
InstallMethod( PlainListCopyOp, "for a GF2 vector",
        true, [IsGF2VectorRep and IsSmallList ],
        0, function( v )
    PLAIN_GF2VEC(v);
    return v;
end);
```
This modifies `v` in place!

Thinking some more about it, if we truly want a generic way to copy any list to a plain list, the "obvious" one to do it is:  `List(somelist, x -> x);` but of course that may have some special methods for "optimizing" this, so better is:
```gap
  n := Length(oldlist);
  newlist := EmptyPlist(n);
  for i in [1..n] do newlist[i] := oldlist[i]; end;
```
and then we can install better methods if we really think it's worth it (but beware: using method dispatch here can be a major performance killer...)

ChrisJefferson: You are right, I definitely didn't pay this bit enough attention. I think calling PLAIN_LIST_COPY in the kernel should handle the "efficient" versions, then just doing something boring.

I think what happened is we had PLAIN_LIST in the kernel, added PLAIN_LIST_COPY as a variant of that, and never connected that to the (undocumented) GAP-level PlainListCopy.

ChrisJefferson: I'm going to pull this bit out in a separate PR.

tst/testinstall/stabchain.tst
```diff
@@ -14,10 +14,19 @@ gap> TestGens := function(g)
...
+> if m <> SortedList(List(g, x -> x)) then Print(g,"7"); fi;
 > return true;
 > end;;
 gap> TestGens(Group(()));;
 gap> List([2..7],
 >       x -> List([1..NrTransitiveGroups(x)],
 >         y -> TestGens(TransitiveGroup(x,y))));;
+gap> TestGens(Group((1,2,3),(4,5,6)));;
+gap> TestGens(Group((1,2,3)(4,5,6)));;
+gap> TestGens(Group((2,4,6),(1,3,5),(1,3)));;
```
wilfwilson, line 31:
```suggestion
gap> TestGens(Group((2,4,6),(1,3,5),(1,3)));;
gap> m := StabChainBaseStrongGenerators(
>    [1,3..1999],
>    List([1,3..1999], x -> (x,x+1)),
>    ());;
gap> Log(SizeStabChain(m), 2) = 1000;
true
```

wilfwilson (approved):
I've thought about this a fair amount, and I'm happy to approve. It would be nice if you include my suggestion below (basically your benchmark example, increased a little).

This should be merged after #4332.

fingolfin (changes requested):
Let's wait until PR #4332 is merged then rebase this, and review again

wilfwilson:
Please rebase when you're ready @ChrisJefferson 🙂

fingolfin (approved):
I took the liberty of rebasing this.

fingolfin:
I've also applied @wilfwilson's patch (I hope you don't mind, Chris -- if you do, you know how to revert it)

PR #4332, opened 2021-03-23
Use internal PLAIN_LIST_COPY to implement PlainListCopy, and document

ChrisJefferson:
After looking at uses of PlainListCopy, I decided the internal kernel function PLAIN_LIST_COPY does everything it is supposed to do, and the best idea was to delegate to it.

While PlainListCopy could (by it's docs) sometimes copy and sometimes not, none of the uses we relying on the behaviour of not copying, or changing the original.

This mainly involved removing the now unused PlainListCopyOp.

This PR is required for #4331

### DIFF
--- a/lib/list.gd
+++ b/lib/list.gd
@@ -2371,41 +2371,6 @@ DeclareGlobalFunction( "IntersectionBlist" );
 ##
 DeclareGlobalFunction( "ListWithIdenticalEntries" );
 
-
-#############################################################################
-##
-#F  PlainListCopy( <list> ) . . . . . . . .  make a plain list copy of a list
-##
-##  <ManSection>
-##  <Func Name="PlainListCopy" Arg='list'/>
-##
-##  <Description>
-##  This is intended for use in certain rare situations,
-##  such as before objectifying.
-##  Normally, <C>ConstantAccessTimeList</C> should be enough.
-##  </Description>
-##  </ManSection>
-##
-DeclareGlobalFunction("PlainListCopy");
-
-
-#############################################################################
-##
-#O  PlainListCopyOp( <list> ) . . . . . . . .return a plain version of a list
-##
-##  <ManSection>
-##  <Oper Name="PlainListCopyOp" Arg='list'/>
-##
-##  <Description>
-##  This operation returns a list equal to its argument, in a plain list
-##  representation. This may be the argument converted in place, or
-##  may be new. It is only intended to be called by <C>PlainListCopy</C>.
-##  </Description>
-##  </ManSection>
-##
-DeclareOperation("PlainListCopyOp", [IsSmallList]);
-
-
 #############################################################################
 ##
 #O  PositionNot( <list>, <val>[, <from>] )  . . . . . . . . .  find not <val>
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -3890,39 +3890,6 @@ InstallMethod( ViewObj,
     Print( " ]" );
     end );
 
-
-#############################################################################
-##
-#F  PlainListCopy( <list> ) . . . . . . . . . . make a plain list copy of
-##                                          a list
-##
-##  This is intended for use in certain rare situations, such as before
-##  Objectifying. Normally, ConstantAccessTimeList should be enough
-##
-##  This function guarantees that the result will be a plain list, distinct
-##  from the input object.
-##
-InstallGlobalFunction(PlainListCopy, function( list )
-    local tnum, copy;
-
-    if not IsSmallList( list ) then
-        Error("PlainListCopy: argument must be a small list");
-    fi;
-
-    # This is enough much of the time
-    copy := ShallowCopy(list);
-
-    # now do a cheap check on copy
-    tnum := TNUM_OBJ(copy);
-    if FIRST_LIST_TNUM > tnum or LAST_LIST_TNUM < tnum then
-        copy := PlainListCopyOp( copy );
-    fi;
-    Assert(2, not IsIdenticalObj(list,copy));
-    Assert(2, TNUM_OBJ(copy) >= FIRST_LIST_TNUM);
-    Assert(2, TNUM_OBJ(copy) <= LAST_LIST_TNUM);
-    return copy;
-end);
-
 #############################################################################
 ##
 #M  PositionNot( <list>, <obj>, <from-minus-one> ) . . . . . . default method
--- a/src/lists.c
+++ b/src/lists.c
@@ -1590,6 +1590,12 @@ Obj PLAIN_LIST_COPY(Obj list)
     return res;
 }
 
+Obj FuncPlainListCopy(Obj self, Obj list)
+{
+    RequireSmallList(SELF_NAME, list);
+    return PLAIN_LIST_COPY(list);
+}
+
 
 /****************************************************************************
 **
@@ -1856,6 +1862,7 @@ static StructGVarFunc GVarFuncs [] = {
     GVAR_FUNC_1ARGS(IS_SSORT_LIST_DEFAULT, list),
     GVAR_FUNC_1ARGS(IS_POSS_LIST_DEFAULT, list),
     GVAR_FUNC_3ARGS(POS_LIST_DEFAULT, list, obj, start),
+    GVAR_FUNC_1ARGS(PlainListCopy, list),
     { 0, 0, 0, 0, 0 }
 
 };
--- a/tst/testinstall/list.tst
+++ b/tst/testinstall/list.tst
@@ -309,6 +309,44 @@ gap> l;
 gap> TNAM_OBJ(l);
 "plain list (rectangular table)"
 
+# Check PlainListCopy
+gap> checkPlainListCopy := function(l)
+>   local copy, tnum;
+>   tnum := TNUM_OBJ(l);
+>   copy := PlainListCopy(l);
+>   return IsPlistRep(copy) and l = copy and
+>          not IsIdenticalObj(l,copy) and TNUM_OBJ(l) = tnum;
+> end;;
+gap> checkPlainListCopy([]);
+true
+gap> checkPlainListCopy([1, ,()]);
+true
+gap> checkPlainListCopy([1..5]);
+true
+gap> checkPlainListCopy([10,8..-4]);
+true
+gap> checkPlainListCopy("");
+true
+gap> checkPlainListCopy("abc");
+true
+gap> checkPlainListCopy(ListWithIdenticalEntries(3, false));
+true
+gap> checkPlainListCopy(NewZeroVector(IsGF2VectorRep, GF(2), 10));
+true
+gap> checkPlainListCopy(NewZeroVector(Is8BitVectorRep, GF(3), 10));
+true
+gap> PlainListCopy(6);
+Error, PlainListCopy: <list> must be a small list (not the integer 6)
+gap> PlainListCopy((1,2,3));
+Error, PlainListCopy: <list> must be a small list (not a permutation (small))
+#@if IsHPCGAP
+gap> PlainListCopy(Group((1,2)));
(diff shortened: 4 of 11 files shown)


ChrisJefferson:
Suggested release notes:

Add function `PlainListCopy`, which can be used to make a copy of any list in GAP's internal "plain list" representation.

lib/list.gd
```diff
@@ -2376,36 +2376,22 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
 ##
 #F  PlainListCopy( <list> ) . . . . . . . .  make a plain list copy of a list
 ##
+##  <#GAPDoc Label="PlainListCopy">
 ##  <ManSection>
 ##  <Func Name="PlainListCopy" Arg='list'/>
 ##
 ##  <Description>
+##  This operation returns a list equal to its argument, in a plain list
```
wilfwilson, line 2384: `PlainListCopy` is currently a function (although maybe it _should_ be an operation? It's not important though)
```suggestion
##  This function returns a list equal to its argument, in a plain list
```

lib/list.gd
```diff
@@ -2376,36 +2376,22 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
...
+##  representation.
 ##  This is intended for use in certain rare situations,
-##  such as before objectifying.
+##  such as before objectifying or calling some kernel functions.
 ##  Normally, <C>ConstantAccessTimeList</C> should be enough.
```
wilfwilson, line 2388: I know you didn't write this, but I can't find any other reference to `ConstantAccessTimeList` in the whole of GAP... so I'm not sure what this is referring to. Maybe just delete this line?
```suggestion
```

lib/list.gd
```diff
@@ -2376,36 +2376,22 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
...
+##  <#GAPDoc Label="PlainListCopy">
 ##  <ManSection>
 ##  <Func Name="PlainListCopy" Arg='list'/>
 ##
 ##  <Description>
+##  This operation returns a list equal to its argument, in a plain list
+##  representation.
 ##  This is intended for use in certain rare situations,
-##  such as before objectifying.
+##  such as before objectifying or calling some kernel functions.
```
wilfwilson, line 2387: Just for clarification, to rule out the interpretation "objectifying some kernel functions or calling some kernel functions"
```suggestion
##  such as before objectifying, or calling some kernel functions.
```

lib/list.gd
wilfwilson, line 2384: I think it would be good to document that the returned list is guaranteed to be an actual copy (i.e. it's equal but not `IsIdenticalObj`)

tst/testinstall/list.tst
```diff
@@ -309,6 +309,29 @@ gap> l;
 gap> TNAM_OBJ(l);
 "plain list (rectangular table)"
 
+# Check PlainListCopy
+gap> checkPlainListCopy := function(l)
+>   local copy, tnum;
+>   copy := PlainListCopy(l);
+>   tnum := TNUM_OBJ(l);
+>   return IsPlistRep(copy) and l = copy and
+>          not IsIdenticalObj(l,copy) and TNUM_OBJ(l) = tnum;
```
wilfwilson, line 318: I don't get the purpose of the `TNUM_OBJ(l) = tnum` here, since you've just defined `tnum := TNUM_OBJ(l);` two lines above.

Is it to make sure that the `TNUM` of `l` didn't change while computing `l = copy` and `IsIdenticalObj(l,copy)`? Because they're the only things involving `l` that happen in between.

wilfwilson (commented):
Thanks for separating this into a separate PR. I've got a few questions/suggestions.

tst/testinstall/list.tst
ChrisJefferson: That is a mistake, I wanted to check PlainListCopy didn't change anything about l, and put the TNUM_OBJ in the wrong place.

lib/list.gd
```diff
@@ -2376,36 +2376,21 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
 ##
 #F  PlainListCopy( <list> ) . . . . . . . .  make a plain list copy of a list
 ##
+##  <#GAPDoc Label="PlainListCopy">
 ##  <ManSection>
 ##  <Func Name="PlainListCopy" Arg='list'/>
 ##
 ##  <Description>
+##  This function returns a list equal to its argument, in a plain list
+##  representation.
```
fingolfin, line 2385: Maybe also add a ref to IsPlistRep for extra clarity?

wilfwilson, line 2385:
```suggestion
##  This function returns a new list that is equal to its argument
##  and is in the plain list representation.
```

lib/list.gd
```diff
@@ -2376,36 +2376,21 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
...
 ##  This is intended for use in certain rare situations,
-##  such as before objectifying.
-##  Normally, <C>ConstantAccessTimeList</C> should be enough.
```
fingolfin, line 2385: This exists and was a typo for ConstantTimeAccessList

wilfwilson (approved):
Thanks for making those changes. Here's a concrete suggestion for how you could implement the remaining comment from my first review.

lib/list.gi
```diff
@@ -3896,31 +3896,12 @@ InstallMethod( ViewObj,
...
-    # now do a cheap check on copy
-    tnum := TNUM_OBJ(copy);
-    if FIRST_LIST_TNUM > tnum or LAST_LIST_TNUM < tnum then
-        copy := PlainListCopyOp( copy );
-    fi;
-    Assert(2, not IsIdenticalObj(list,copy));
-    Assert(2, TNUM_OBJ(copy) >= FIRST_LIST_TNUM);
-    Assert(2, TNUM_OBJ(copy) <= LAST_LIST_TNUM);
-    return copy;
+    return PLAIN_LIST_COPY(list);
```
fingolfin, line 3904: Why not just `BindGlobal("PlainListCopy", PLAIN_LIST_COPY);`

Or even rename the kernel function?

src/lists.c
```diff
@@ -1590,6 +1590,10 @@ Obj PLAIN_LIST_COPY(Obj list)
     return res;
 }
 
+Obj FuncPLAIN_LIST_COPY(Obj func, Obj list) {
```
fingolfin, line 1593:
```suggestion
Obj FuncPLAIN_LIST_COPY(Obj func, Obj list)
{
```

tst/testinstall/list.tst
```diff
@@ -309,6 +309,29 @@ gap> l;
...
+true
+gap> checkPlainListCopy([1..5]);
+true
+gap> checkPlainListCopy([10,8..-4]);
+true
+gap> checkPlainListCopy("");
+true
+gap> checkPlainListCopy("abc");
+true
+gap> checkPlainListCopy(ListWithIdenticalEntries(3, false));
```
fingolfin, line 332: Let's also test some compressed matrices and vectors?

Also, it would be good if we inspected all callers of PlainListCopy. I can help with that later

ChrisJefferson:
I'm going to go though all the callers of PlainListCopy.

Firstly, there a number of lines of the form `x := PlainListCopy(x)` in listcoef.gi. These lines include other attempts to convert between representations. As we are assigning the return, it doesn't make any difference if the original was returned, or copied.

The function PLAIN_SL tries to convert a sparse list in place to a plain list. It does: CLONE_OBJ(sl, PlainListCopy(sl));, so once again it doesn't matter if PlainListCopy returns the original, or a copy. I checked this function does work correctly if PlainListCopy does, or doesn't return a copy. This function is very weird looking, but no need to change it today.

Finally, tuples.gi calls PlainListCopy on a newly created list. I imagine here it is because List can "be clever" and return a weird type of list.

fingolfin:
I mildly disagree that `x := PlainListCopy(x)` implies it doesn't matter whether we modified the original `x` in place or not: sure, we don't reference it anymore in `x` afterwards, but it still might be referenced somewhere else, and so the modification still be visible there... However, all calls in listcoef.gi are preceded by a check of the form `if q <> ConvertToVectorRepNC(vec,q)` so we tried inplace conversion of the vector and it failed, which *should* mean that the vector is not compressed, and so we are good. (Extra crazy weirdness could occurs if vec is compressed but cannot be stored compressed over GF(q), but that shouldn't be possible here, and if it is, I frankly don't care, then this code is bizarre and broken). So in the end, I agree with you, this part is fine.

The whole "sparse list" code is undocumented, unused, not loaded, and AFAIK incomplete. We may as well delete it (opened PR #4336).

I really don't understand why it is used in `tuples.gi` (wish there was a comment explaining it!) but I am willing to risk it ;-).

Finally, no packages seem to be using those. Good!

fingolfin (approved):
Looks good to me but perhaps review the couple comments by Wilf and myself, and decide if you want to do anything about any of them

fingolfin:
I took the liberty and adjusted the suggested release notes. Feel free to revise further.

lib/list.gd
fingolfin:
```suggestion
##  representation (see <Ref Filt="IsPlistRep"/>).
```

lib/list.gd
fingolfin: That said, I don't think this sentence was helpful, so I am fine with removing it :-)

tst/testinstall/list.tst
```diff
@@ -309,6 +309,29 @@ gap> l;
...
+true
```
fingolfin, line 333:
```suggestion
true
gap> checkPlainListCopy(NewZeroVector(IsGF2VectorRep, GF(2), 10));
true
gap> checkPlainListCopy(NewZeroVector(Is8BitVectorRep, GF(3), 10));
true
```

lib/list.gd
```diff
@@ -2376,35 +2376,19 @@ DeclareGlobalFunction( "ListWithIdenticalEntries" );
...
-##  <Description>
-##  This operation returns a list equal to its argument, in a plain list
-##  representation. This may be the argument converted in place, or
-##  may be new. It is only intended to be called by <C>PlainListCopy</C>.
+##  such as before objectifying, or calling some kernel functions.
 ##  </Description>
 ##  </ManSection>
+##  <#/GAPDoc>
 ##
-DeclareOperation("PlainListCopyOp", [IsSmallList]);
```
fingolfin, line 2406: Should we say here that it is a kernel function, in case people wonder where to find it?

fingolfin: (I am also fine with merging as-is, just wondering)

ChrisJefferson: I've now moved the documentation inline, because it was weird here -- but maybe that's a bad idea too. We could start scanning C code for GAPDoc?

fingolfin: In `doc/ref/makedocreldata.g` we already list `src/sysfiles.c` and could add more. I would in fact suggest that we change that file to not hardcode the list of files to scan, but rather to scan all files `src/*.{c,h}`, `lib/*.{gd,gi}`, etc., but of course that goes way beyond the scope of this PR ;-)

wilfwilson (approved):

wilfwilson:
The CI failures:
```
########> Diff in /home/runner/work/gap/gap/tst/testbugfix/2006-08-28-t00\
151.tst:14
# Input is:
if time1 >= 2*time2 then
Print("Bad timings for bugfix 2006/08/28 (FL): ", time1, " >= 2*", time2, "\n"\
); 
fi; # time1 and time2 should be about the same
# Expected output:
# But found:
Bad timings for bugfix 2006/08/28 (FL): 19430 >= 2*4959
########
```

fingolfin (approved):
Still looks good to me, please merge at will

PR #4334, opened 2021-03-23
added the group constructors `PGO` and `PSO`

ThomasBreuer:
... and changed `POmega` to use the same mechanism


fingolfin:
The failing tests look as if they may need PR #4333? (Not a problem, of course, we'll probably merge that soon, then this PR here can be rebased)

grp/classic.gd
```diff
@@ -993,7 +993,125 @@ DeclareSynonym( "PSp", PSP );
...
+##  <#GAPDoc Label="ProjectiveGeneralOrthogonalGroup">
+##  <ManSection>
+##  <Func Name="ProjectiveGeneralOrthogonalGroup" Arg='[filt, ][e, ]d, q'/>
+##  <Func Name="PGO" Arg='[filt, ][e, ]d, q'/>
+##
+##  <Description>
+##  constructs a group isomorphic to the projective group
+##  PGO( <A>e</A>, <A>d</A>, <A>q</A> )
+##  of GO( <A>e</A>, <A>d</A>, <A>q</A> ),
+##  modulo the centre
```
fingolfin, line 1056: A nitpick (which really is about existing documentation, which you just covered, so I don't really expect it to be addressed here): Formally, the definition is not to take the quotient modulo the center, but rather modulo scalar matrices (achieved implicitly by using a projective action on the vector). The two are related by not completely trivial mathematics. I think for a user who is not yet very experienced, this can be confusing (source: myself, many years ago ;-) -- so maybe I am just projecting here (pun not intended) and nobody else ever got confused).

Anyway, I guess what I am saying is that it might be better if we described in a separate section / paragraph what we mean by projective group (modulo scalars, equiv: acting on lines) and that this means we take a central factor, and that for many important cases ("full" groups) this in fact means factoring out the whole center. And then we can just reference that.

Thoughts? If people are not adverse to this, I could give it a go some of these days (I'd probably for now just submit an issue to not forget it...)

fingolfin (approved):
Seems sensible.

grp/classic.gd
ThomasBreuer: Such a section on projective actions would be useful. It could refer to the FinInG package. (I had expected that also the recog package contains something interesting in this respect, but I did not find this at first sight.) In the other direction, the functions `NormedVectors`, `OnLines`, `ProjectiveActionOnFullSpace`, `ProjectiveOrder` could refer to the new section.

PR #4336, opened 2021-03-23
Remove unused sparselist code

fingolfin:
Is there any point in keeping this around?


fingolfin:
No comments in march; I think this can be removed...

fingolfin:
The alternative would be to move the code into `dev/deleted-code` if someone feels that's better *shrug*

ThomasBreuer (approved):
The code is probably not known to most users/developers, thus there is no problem with removing it.

Apparently there is no read need for sparse lists in the GAP library; what would be the right place for such code, perhaps the `datastructures` package?

PR #4337, opened 2021-03-24
Partially prepare grp/classic.gi for MatrixObj

fingolfin:
- use mat[i,j] instead of mat[i][j]
- avoid using Length(mat)
- consistently write `Sp(n,q)` instead of `SP(n,q)`
- use NullMat / IdentityMat instead of tricks using `List` and `ShallowCopy`

Ideally should use ZeroMatrix / IdentityMatrix, but there are some caveats with those right now, so I am doing this intermediate step now, while working on another matobj related PR.

### DIFF
--- a/grp/classic.gi
+++ b/grp/classic.gi
@@ -29,54 +29,50 @@ InstallMethod( SymplecticGroupCons,
     f := GF(q);
     z := PrimitiveRoot( f );
     o := One( f );
-#T introduce variable d_2 = d/2 ?
 
     # if the dimension is two it is a special linear group
     if d = 2 then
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
This looks good.
There is still one instance of `List( ..., ShallowCopy )` in `grp/classic.gi`.

fingolfin:
Indeed; I removed that last use of `ShallowCopy`, and then noticed two uses of `StructuralCopy` in `WreathProductOfMatrixGroup`, and *then* I noticed that there now were a bunch of variables named `id` that got defined but never used, so I removed those as well.

Issue #1362, opened 2017-05-24
SymmetricPower undocumented


dimpase:
While one can do
```
gap> Display(SymmetricPower([[1,2],[3,4]],3));
[ [   1,   2,   4,   8 ],
  [   9,  16,  28,  48 ],
  [  27,  42,  64,  96 ],
  [  27,  36,  48,  64 ] ]
```
it is not documented. It is implemented in `lib/algrep.gi`. Once in a while people search for it in vain, including longterm GAP devs (e.g. @hulpke asked for such a function on `gap-dev` list in 2014). Even if this will be re-implemented at some point, it seems strange to keep it hidden,

hulpke:
Just as an aside:

The method your example calls is in `matrix.gi`, I added it *after* finding no such function available.

fingolfin:
See also https://github.com/gap-system/gap/issues/1796

stevelinton:
I just ran into the same problem. Also the methods run into a divide by zero over finite fields.

jackschmidt:
I've worked on this a little. I have some simple but longer code with the following properties:
* :+1: agrees with matrix.gi for square matrices
* :+1: works for all (including small) characteristics
* :+1: works for rectangular matrices (homomorphisms between symmetric powers) and agrees with M2's SchurFunctors
* :+1: is often faster than matrix.gi (10x10 one-digit to the 4th power takes 1 second compared to 10 seconds)
* :+1: is often faster than M2's SchurFunctors (9x9 over GF(13) to the 4th power takes 1.5 seconds compared to 20 seconds)
* :-1: is slower than matrix.gi for high powers (depends on matrix size, but 9x9 over GF(13) to the 6th power takes 1000 seconds compared to 500 seconds)

Should I just paste the undocumented code here, or put it in a pull request, or document and test it?

Documentation and testing will take a bit more time -- the code is meant to be easily documented, except that I seem to have misunderstood something (a missing dual space?) and so the code currently confuses me.

Algorithmic complexity also worries me: why is the Permanent method faster for higher powers? It appears to do a divide and conquer approach. Is there a similar technique to compute the symmetric power (without division)?

It was also unclear to me if it would be nice to have all the SchurFunctors (not just symmetric and exterior powers).  Would it be much more work? M2 computes a tensor product of exterior powers and then sums up some rows and columns. I don't read M2 code very well, and so was unable to port it to GAP on my first attempt.

PR #4339, opened 2021-03-24
Document `TensorProduct`, `ExteriorPower` and `SymmetricPower` (they existed since at least GAP 4.4)

fingolfin:
This is triggered by a discussion with @AnnaKDS and @wucas.

They've been around for many years but undocumented; people still use them. I see no harm in documenting them.

Related: issue #1362 which discusses a `SymmetricPower` method for matrices, which perhaps also should be documented? And related to *that* (as well as `TensorProduct`) there is `KroneckerProduct`, which already is in the manual -- but perhaps there should be references between those?

In the same issue, problems with these over finite fields are mentioned. Perhaps those should be investigated and fixed.

Finally, some tests in the test suite for these would be nice.

### DIFF
--- a/doc/ref/vspc.xml
+++ b/doc/ref/vspc.xml
@@ -183,6 +183,17 @@
 <#Include Label="CheckForHandlingByNiceBasis">
 
 </Section>
+
+
+<Section Label="Tensor Products and Exterior and Symmetric Powers">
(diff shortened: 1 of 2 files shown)


wucas (approved):

AnnaKDS (approved):

pedritomelenas (commented):
Fine.

PR #4050, opened 2020-06-11
Use `Pluralize` to give correct pluralization in many GAP library methods

wilfwilson:
This PR was originally part of #3992, so it has already been discussed there.

This PR uses the `Pluralize` function in many locations in the GAP library, so that GAP produces more grammatically correct strings - such as `<magma with 1 generator>` rather than `<magma with 1 generators>`. There is also a follow-up PR #4343.

This PR affects the `ViewString` of many objects, and so this will cause the tests/manual examples for many packages to fail. Therefore this cannot be merged until those tests/examples are adjusted, and new package versions are released. I will use this description to keep track of what needs to be done for that, initially copied from https://github.com/gap-system/gap/pull/3992#issuecomment-626346932.

### Open PRs

None.

### Pending release

None.

### Resolved

- [x] anupq – fixed in v3.2.3 (via https://github.com/gap-packages/anupq/pull/46)
- [x] circle – fixed in v1.6.4 (via https://github.com/gap-packages/circle/pull/9)
- [x] crime
- [x] crisp – incorrect pluralisations only occur in a `.tex` file
- [x] cubefree – fixed in v1.19 via https://github.com/gap-packages/cubefree/pull/5
- [x] ctbllib-1.3.2 - incorrect pluralisations only occur in `doc2/ambigfus.xml`, and the relevant test does not appear in a test file. So I don't think that merging this will cause its tests to fail.
- [x] digraphs – fixed in v1.2.1
- [x] fining – I'm pretty sure that the manual example output is automatically generated and not used in tests; test files are not affected by this PR (I ran `tst/testall.g` with this PR)
- [x] forms – one instance, but doesn't seem to affect tests
- [x] fr – one manual example affected, but those seem not to be used in tests.
- [x] laguna – fixed in v3.9.4 (via **https://github.com/gap-packages/laguna/pull/20)
- [x] lpres
- [x] nilmat – fixed in v1.4.1 via https://github.com/gap-packages/nilmat/pull/4
- [x] nq
- [x] NumericalSgps – might want to use `Pluralize` in the future
- [x] qpa-version – uses it in manual examples, but those are manually formatted anyway and not used in tests?
- [x] rds – fixed in v1.8 via https://github.com/gap-packages/rds/pull/1
- [x] RepnDecomp – fixed in v1.2.0 via https://github.com/gap-packages/RepnDecomp/pull/10
- [x] semigroups – fixed initially in v3.3.1 and again in v3.4.1
- [x] singular – fixed in v2020.12.18
[...]

Issue #2336, opened 2018-03-30
Plan: get rid of hidden implications


fingolfin:
### Short summary

I would like to get rid of hidden implications in GAP completely. They can lead to counterintuitive problems, like the order in which packages were loaded can affect ranking; and other problems, see e.g. issue #1649. 

* [ ] adding a way to turn off hidden implications temporarily, for testing: see PR #2246
* [ ] turning hidden implications in the GAP library into actual implications where possible: see PRs #2222 and #2247 
* [ ] turning hidden implications in GAP packages into actual implications where possible
* [ ] identifying the effect of the remaining hidden implications, and trying to get minimize or eliminate it (i.e.: identify changes in method ordering; determine if they matter; adjust method ranking if needed; etc.)
* [ ] finally, eliminate the remaining hidden implications and the mechanism powering them.

### Background: What are hidden implications?

Currently, GAP's method ranking relies on both implications, and so-called "hidden implications". An implication is when a filter `IsFOO` implies another filter `IsBAR` (e.g. via `InstallTrueMethod`). In that case, the rank of `IsFOO` should be larger than that of `IsBAR`, reflecting the idea that the more "properties" (used in a non-technical sense here) a filter implies, the higher it should be ranked.

This is all good and fine, but apparently was/is not quite good enough to get a good ranking system (?). Thus GAP also uses so-called "hidden" implications: A "hidden" implication from a filter `IsFOO` to a filter `IsBAR` has the same effect on the relative ranking of the two filters, but without a mathematical implication between the two.

These typical are created indirectly when you declare a property or attribute. Example: 
```gap
DeclareProperty( "IsNilpotentGroup", IsGroup );
```
This creates a hidden implication from `IsNilpotentGroup` to `IsGroup`, but not a mathematical one. Meaning that the rank of `IsNilpotentGroup` is larger than that of `IsGroup`; but in principle at least, something could be in the filter `IsNilpotentGroup` without being also in the filter `IsGroup`. Now, in this particular case, one "obviously" also intended an actual proper implication -- but none exists! One has to manually establish it, e.g. via this:
```gap
InstallTrueMethod( IsGroup, IsNilpotentGroup )
```

The reason for this is clearer if you consider this example: in polycyclic, we have this:
```
DeclareProperty( "IsTorsionFree", IsGroup );
```gap
Now, this installs a hidden implication `IsTorsionFree` to `IsGroup`. But now perhaps you are writing a package in which you deal with torsion free monoids. So you add this:
[...]

PR #4341, opened 2021-03-25
Collection of miscellaneous small changes ("cleanup")

wilfwilson:
This is a collection of several 'random' cleanup-style commits that I've made over the last year, but that have never made their way into another PR. I decided to just collect them into one PR here.

This is split into 7 commits with descriptive commit messages to help with reviewing.

### DIFF
--- a/lib/gpprmsya.gi
+++ b/lib/gpprmsya.gi
@@ -673,14 +673,8 @@ InstallMethod(IsNaturalAlternatingGroup,"knows size",true,[IsPermGroup
 end );
 
 InstallMethod(IsNaturalAlternatingGroup,"comprehensive",true,[IsPermGroup],0,
-function( grp )
-  if 0 = NrMovedPoints(grp)  then
-    return IsTrivial(grp);
-  else
-    return PermgpContainsAn(grp)=true and
-      ForAll(GeneratorsOfGroup(grp),i->SignPerm(i)=1);
-  fi;
-end );
+G -> IsTrivial(G) or
+  ForAll(GeneratorsOfGroup(G),i->SignPerm(i)=1) and PermgpContainsAn(G)=true);
 
 #############################################################################
 ##
@@ -705,14 +699,8 @@ InstallMethod(IsNaturalSymmetricGroup,"knows size",true,[IsPermGroup
 end );
 
 InstallMethod(IsNaturalSymmetricGroup,"comprehensive",true,[IsPermGroup],0,
-function( grp )
-  if 0 = NrMovedPoints(grp)  then
-    return IsTrivial(grp);
-  else
-    return PermgpContainsAn(grp)=true and
-      ForAny(GeneratorsOfGroup(grp),i->SignPerm(i)=-1);
-  fi;
-end );
+G -> IsTrivial(G) or
+  ForAny(GeneratorsOfGroup(G),i->SignPerm(i)=-1) and PermgpContainsAn(G)=true);
 
 #############################################################################
 ##
--- a/tst/testinstall/pperm.tst
+++ b/tst/testinstall/pperm.tst
@@ -4118,6 +4118,11 @@ gap> OnTuples([1], PartialPerm([])) = [];
 true
 
 #
+gap> SetUserPreference("PartialPermDisplayLimit", 1);
+gap> PartialPerm([1], [2]);
+<partial perm on 1 pts with degree 1, codegree 2>
+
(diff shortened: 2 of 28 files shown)


tst/testinstall/pperm.tst
```diff
@@ -4117,6 +4117,11 @@ true
 gap> OnTuples([1], PartialPerm([])) = [];
 true
 
+#
+gap> SetUserPreference("PartialPermDisplayLimit", 1);
+gap> PartialPerm([1], [2]);
+<partial perm on 1 pts with degree 1, codegree 2>
```
wilfwilson, line 4123: Note: if (when) this PR is merged, #4050 will need to be rebased and updated to change this line to
```suggestion
<partial perm on 1 pt with degree 1, codegree 2>
```

lib/gpprmsya.gi
```diff
@@ -673,14 +673,8 @@ InstallMethod(IsNaturalAlternatingGroup,"knows size",true,[IsPermGroup
...
-function( grp )
-  if 0 = NrMovedPoints(grp)  then
-    return IsTrivial(grp);
-  else
-    return PermgpContainsAn(grp)=true and
-      ForAll(GeneratorsOfGroup(grp),i->SignPerm(i)=1);
-  fi;
-end );
+G -> IsTrivial(G) or
+  ForAll(GeneratorsOfGroup(G),i->SignPerm(i)=1) and PermgpContainsAn(G)=true);
```
fingolfin, line 677: Perhaps also like this?
```suggestion
  ForAll(GeneratorsOfGroup(G),i->SignPerm(i)=1) and PermgpContainsAn(G));
```

lib/semitran.gd
```diff
@@ -17,8 +17,6 @@ DeclareSynonym("IsTransformationMonoid", IsMonoid and
 IsTransformationCollection);
 
 DeclareProperty("IsFullTransformationSemigroup", IsSemigroup);
-InstallTrueMethod(IsSemigroup, IsFullTransformationSemigroup);
```
fingolfin, line 20: Why is this unnecessary?

lib/gpprmsya.gi
wilfwilson: `PermgpContainsAn` can return either `true` or `false` or the return value of `PermNatAnTestDetect` which can be a set of points, so I can't make that change (as much as I'd like to).

lib/semitran.gd
wilfwilson: Perhaps I misunderstand how things work.

Since `IsFullTransformationSemigroup` is only declared for `IsSemigroup`, that means anything that could know it satisfies `IsFullTransformationSemigroup` already knows that it is a semigroup.

Is perhaps the idea that: this `InstallTrueMethod` is useful because perhaps someone in a package could (for example) do `DeclareProperty("IsFullTransformationSemigroup", IsMagma);` and then install methods for `IsMagma` (which would have to somehow test the magma for associativity, but perhaps not directly, and this could therefore mean that `IsFullTransformationSemigroup` gets set but not `IsSemigroup`)?

But actually it seems that any magma of transformations already knows that it is associative, and therefore a semigroup:
```
gap> S := Magma(GeneratorsOfMagma(FullTransformationSemigroup(3)));;
gap> HasIsSemigroup(S);
true
```
I can't think of any use for this true method, but perhaps I am not imaginative enough 🙂 

I notice that there are similar true methods for natural symmetric and alternating groups (e.g. in `lib/gpprmsya.gd`):
```
InstallTrueMethod( IsPermGroup, IsNaturalSymmetricGroup );
```
But things seem to work differently for groups:
```
gap> G := Semigroup(GeneratorsOfMagma(SymmetricGroup(4)));;
gap> IsGroup(G);
false
gap> SetIsNaturalSymmetricGroup(G, true);
gap> IsGroup(G);
true
```

fingolfin: There is no automatic implication and there cannot be, as indeed GAP cannot predict the future and so can't know whether that property might in the future also be applicable to, say, algebras or strings or graphs or... Sure, as a human you know this is unlikely because of the name, but GAP does not known this (and anyway.

Somewhat related: issue #2336

wilfwilson: > (and anyway.

Were you going to add something?

Anyway, I'll be happy to remove this commit. I'll read #2336 too.

fingolfin (approved):

PR #4342, opened 2021-03-25
Refactor PutLine2

fingolfin:
Split this into two commits to make it easier (I hope) to trace what happened. Feel free to squash when merging.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1085,28 +1085,24 @@ static Char GetLine(TypInputFile * input)
 **  knows the length of <line>, so it is not necessary to compute it again
 **  with the inefficient C- strlen.  (FL)
 */
-
 static void PutLine2(TypOutputFile * output, const Char * line, UInt len)
 {
-  Obj                     str;
-  if ( output->isstream ) {
-    /* special handling of string streams, where we can copy directly */
+    Obj str;
+
     if (output->isstringstream) {
-      str = CONST_ADDR_OBJ(output->stream)[1];
-      ConvString(str);
-      AppendCStr(str, line, len);
-      return;
+        // special handling of string streams, where we can copy directly
+        str = CONST_ADDR_OBJ(output->stream)[1];
+        ConvString(str);
+        AppendCStr(str, line, len);
     }
-
-    /* Space for the null is allowed for in GAP strings */
-    str = MakeImmStringWithLen(line, len);
-
-    /* now delegate to library level */
-    CALL_2ARGS( WriteAllFunc, output->stream, str );
-  }
-  else {
-    SyFputs( line, output->file );
-  }
+    else if (output->isstream) {
+        // delegate to library level
+        str = MakeImmStringWithLen(line, len);
+        CALL_2ARGS(WriteAllFunc, output->stream, str);
+    }
+    else {
+        SyFputs(line, output->file);
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


wilfwilson (commented):
Would it ever be possible for `output->isstringstream` to be true, but `output->isstream` to be false?

I assume not, but if the answer if 'yes' then the refactored version behaves differently in this case. In particular, in this case, the `if (output->isstringstream)` block is executed, rather than the `else` block.

Just wanted to sanity check this. Otherwise this is fine.

fingolfin:
No this is not possible (I checked it prior to submitting)

wilfwilson (approved):
Thanks Max.

wilfwilson:
The code coverage reports are weird, have you noticed this kind of thing before? Two comments and an `else` apparently do not have tests coverage! (Obviously that's not a problem with this PR).

Screenshot 2021-03-25 at 19 58 14

fingolfin:
Even for the latest master coverage report, if you look at https://codecov.io/gh/gap-system/gap/src/master/src/io.c#L1089 you'll see a bunch of comments marked as "not executed code".

We've seen this before, it was caused by some GAP package (or multiple) also having an `io.c` file, and coverage data for the two were mixed up. Quoting `dev/ci-prepare.sh:88`:
```
  # Compile io and profiling packages
  # we deliberately reset CFLAGS, CXXFLAGS, LDFLAGS to prevent them from being
  # compiled with coverage gathering, because otherwise gcov may confuse
  # IO's src/io.c with GAP's.
```
and then `dev/ci.sh:63` says
```
    # reset CFLAGS, CXXFLAGS, LDFLAGS before compiling packages, to prevent
    # them from being compiled with coverage gathering, because
    # otherwise gcov may confuse IO's src/io.c, or anupq's src/read.c,
    # with GAP kernel files with the same name
```

I've now re-inspected some of the raw coverage data (see e.g. https://codecov.io/gh/gap-system/gap/commit/f43438bfbe0e8f2f6b31d009825c40dbe3c3c6b4/build) and it seems there is also `/extern/gmp/tests/mpq/io.c` and `/extern/gmp/tests/mpz/io.c`. And there is definitely coverage data for those as well as those files in pkg.  I tried several times in the past to filter those out, and it seemed to work for a time, but I guess it's back :-(.

Oh, also in `dev/ci-run-codecov.sh` there's an attempt to resolve this; note the exclusion patterns used here: (I may very well be using them *incorrectly*... it's not as if this stuff was documented very well :-(
```
./codecov.sh -f '!./pkg/*' -f '!./extern/*' -f '!./build/*'
```

fingolfin:
Actually, I think I know part of the problem: removing the `--coverage` flag in the CI script does not quite work anymore since a lot of packages now use `Makefile.gappkg` and that in turn directly uses the compile flags of GAP as stored in sysinfo.gap, and *that* still contains that `--coverage` flag... *sigh*

fingolfin:
So I guess I can hack away that `--coverage` flag, but the real fix would be to stop Codecov (or gcov, or whatever tool is at fault) from mixing up those `io.c` files *grr*

fingolfin:
Oh yeah, I also put `pkg` and `extern` in our `.codecov.yml` back in the day, also didn't help:
```
ignore:
 - "extern"
 - "hpcgap/extern"
 - "tst"
 - "pkg"
 - "grp/*.grp"
```

PR #4345, opened 2021-03-25
Fix an implication to `IsSemiband`

wilfwilson:
...and add implications of semigroup-only properties to `IsSemigroup`.

Previously, there was a true method:
```
InstallTrueMethod(IsSemiband, IsIdempotentGenerated);
```
since if a semigroup is idempotent generated, it is a semiband (in fact, in and only if - but I'll deal with that later), and `IsIdempotentGenerated` is _currently_ only declared for `IsSemigroup`. So currently this _should_ only ever set `IsSemiband` to be `true` for idempotent generated semigroups. But someone may well want to declare it for `IsMagma`, and a non-associative magma cannot be a semiband, whether or not it is idempotent generated. That's because a semiband is a special kind of semigroup.

Currently, you can force this mathematically incorrect situation to happen:
```
gap> tab := [[4, 2, 1, 4], [2, 2, 2, 3], [2, 4, 3, 1], [4, 1, 1, 4]];;
gap> m := MagmaByMultiplicationTable(tab);;
gap> m = Submagma(m, Filtered(m, IsIdempotent));
true
gap> SetIsIdempotentGenerated(m, true);
gap> IsSemigroup(m);
false
gap> IsSemiband(m);
true
```
In this first commit of this PR, I change the true method to:
```
InstallTrueMethod(IsSemiband, IsIdempotentGenerated and IsSemigroup);
```
and added a bugfix test file.

I don't think this needs to be mentioned in the release notes (it's arguably more a potential bug than a bug...) but I'm happy to change the label and add some text for the release notes if you prefer.

In a second commit, I added `InstallTrueMethod(IsSemigroup, <property>);` for all the remaining semigroup properties in `lib/semigrp.gd` to which it mathematically applies – which was in fact all of them, except for `IsIdempotentGenerated` – even though some of them are surely redundant because of the other true methods installed at the bottom of the file.

I will review and probably clean up/add to those latter true methods in a later PR.

### DIFF
--- a/lib/semigrp.gd
+++ b/lib/semigrp.gd
@@ -655,38 +655,68 @@ DeclareAttribute("NilpotencyDegree", IsSemigroup);
 DeclareOperation("IsSubsemigroup", [IsSemigroup, IsSemigroup]);
 
 DeclareProperty("IsBand", IsSemigroup);
+
 DeclareProperty("IsBrandtSemigroup", IsSemigroup);
+InstallTrueMethod(IsSemigroup, IsBrandtSemigroup);
+
 DeclareProperty("IsCliffordSemigroup", IsSemigroup);
+InstallTrueMethod(IsSemigroup, IsCliffordSemigroup);
+
 DeclareProperty("IsCommutativeSemigroup", IsSemigroup);
+InstallTrueMethod(IsSemigroup, IsCommutativeSemigroup);
+
 DeclareProperty("IsCompletelyRegularSemigroup", IsSemigroup);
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

wilfwilson:
I pushed an update which removed the implications `IsBand => IsSemigroup` and `IsSemilattice => IsSemigroup`, since [bands](https://en.wikipedia.org/wiki/Band_(order_theory)) and [semilattices](https://en.wikipedia.org/wiki/Semilattice) aren't necessarily only about semigroups.

PR #4346, opened 2021-03-25
CI: fix coverage reporting

fingolfin:
See the discussion at https://github.com/gap-system/gap/pull/4342 for details.

Not sure if this will work, we'll see.

### DIFF
--- a/dev/ci-upload-coverage.sh
+++ b/dev/ci-upload-coverage.sh
@@ -13,4 +13,4 @@ fi
 # upload to Codecov
 curl -s https://codecov.io/bash > codecov.sh
 chmod +x codecov.sh
-./codecov.sh -f '!./pkg/*' -f '!./extern/*' -f '!./build/*'
+./codecov.sh -f '!./pkg/**' -f '!./extern/**' -g './pkg/**' -g './extern/**'


fingolfin:
Maybe also https://docs.codecov.io/docs/fixing-paths can help us, in addition to this.

wilfwilson:
Just to note, this follows on from a discussion on #4342.

fingolfin:
Seems to have worked, see https://codecov.io/gh/gap-system/gap/src/6d1762a8247ca7ae561f93209721c51cf4e2443e/src/io.c

wilfwilson:
Yes, things seem to be looking quite reasonable.

Not completely reasonable, e.g. I can't understand why there are many lines that are neither green nor red, despite being surrounded by other red/green lines, e.g. in 

Screenshot 2021-03-26 at 09 57 54

and

Screenshot 2021-03-26 at 09 58 44

wilfwilson (approved):
But I'm happy for this to be merged anyway.

fingolfin:
Unfortunately this causes `testspecial` to fail with a linker error -- it uses `gac` which in turn uses `sysinfo.gap`, and now the `--coverage` flag is missing there.... I'll look into it

fingolfin:
@wilfwilson huh, that spotty coverage is indeed wildly bizarre! For a few of them, one might be able to explain this with compiler optimizations, but overall this looks really bad.

And on https://app.codecov.io/gh/gap-system/gap/compare/4346/tree/src/io.c it looks completely broken, as no "green" lines are shown at all (but I am happy that this page loads at all -- for many years those PR specific Codecov pages just timed out for us, presumably because we uploaded too much data for them to handle...).

Anywhere, from there I can get with a few clicks first to the commit level reports and then to the latest, which now is https://codecov.io/gh/gap-system/gap/src/ff4358f9ab0d01e8e194b2651d62ee6bbd985fed/src/io.c, which looks much better -- it still has a few of these weird "gaps" were code is marked as non-code (and it also misses HPC-GAP coverage :-(). But it's much more reasonable... yet still incomplete. Weird. I wonder if this might be a problem with the compiler version and/or gcov ?

fingolfin:
Reported weirdness at https://community.codecov.io/t/coverage-report-not-showing-any-green-lines-executed-code-on-app-codecov-io-but-works-on-codecov-io/2685

wilfwilson:
:-( indeed. Thanks for reporting it.

PR #4347, opened 2021-03-25
Restore CopySubVector argument order to it original state

fingolfin:
During a MatrixObj design discussion some time (... years *sighs) ago, we
decided to change the order of the arguments of `CopySubVector` to make them
fit better with the rest of the API.

The idea was to do this in a backwards compatible way, so that existing
calling code would keep working.

However, in reality, this compatibility is flawed (as soon you operate on
integer vectors, it is ambiguous), and we only changed the documentation,
added a compatibility method, and changed 1-2 other places -- but we didn't
even complete the conversion in the GAP library, let alone in any of the
packages using this (and then of course users may have code we never will see
that also uses it).

I run into this when I wanted to use CopySubVector in some GAP library code,
and run into errors...

At this point, it seems prudent to just go back to how things were: yes,
the new argument might be "better", and if this was new code, I'd go for it;
but it doesn't seem worth the hassle to do it properly. If we really care,
we should instead come up with a new name for that function... but I'll leave
that thought for someone else or at for some other day.

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -653,8 +653,8 @@ InstallMethod( ExtractSubMatrix,
 
 InstallMethod( CopySubVector,
     "generic method for vector objects",
-  [ IsVectorObj and IsMutable, IsList, IsVectorObj, IsList ],
-  function(dst, dcols, src, scols)
+  [ IsVectorObj, IsVectorObj and IsMutable, IsList, IsList ],
+  function(src, dst, scols, dcols)
     local i;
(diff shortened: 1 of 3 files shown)


ThomasBreuer (approved):
Yes, this change makes things better.
Besides the logical problem with "backwards compatibility, I see also the compatibility with `CopySubMatrix` as an argument to return to the situation that is documented in the cvec package (for both `CopySubVector` and `CopySubMatrix`).

PR #4348, opened 2021-03-25
Use PositionNonZero in Meataxe code

fingolfin:
Besides being "cleaner" this will also be more efficient for some matrix implementations... Motivated by profiling some slow Meataxe example sent by @danielrademacher (not that this changed helped noticeably there, but still ...)

### DIFF
--- a/lib/meatauto.gi
+++ b/lib/meatauto.gi
@@ -102,7 +102,7 @@ end);
 
 
 BindGlobal("SMTX_AddEqns",function ( eqns, newmat, newvec)
-local n, zero, weights, mat, vec, NextPositionProperty, ReduceRow, t,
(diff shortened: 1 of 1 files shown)


wilfwilson (approved):

danielrademacher (approved):
Looks great. Thanks Max! I will make some performance tests today and post them here.

fingolfin:
Oh I don't think this PR will help with performance in a noticeable way. But I have one in preparation which gives about 20% for indecomposing SL(40,3)

danielrademacher:
Yes, you're right Max. In the tests for my code, I couldn't see any real difference in performance. But the code looks nicer.

Thanks! I'm already looking forward to the next PR.

PR #4349, opened 2021-03-25
Forbid installing an operation as method for itself

### LINKED ISSUE #4340: Segfault after installing a silly method for a synonym attribute
In the current `master` and `stable-4.11` branches, and in `4.11.1` (I believe), I can cause a segfault in the following way:
```
gap> DeclareAttribute("Wilf", IsGroup);
gap> DeclareSynonymAttr("Wilson", Wilf);
gap> InstallMethod(Wilf, [IsGroup], x -> fail);
gap> InstallMethod(Wilson, [IsGroup], Wilf);
gap> Wilf(SymmetricGroup(3));
Segmentation fault: 11
```
Calling `Wilson(SymmetricGroup(3));` instead on the final line also gives the a segfault.

The specific group involved, and indeed even the use of `IsGroup` as the filter rather than some other object, doesn't seem to be important.

Obviously I'm not using things correctly – the declaration as a synonym means that the line `InstallMethod(Wilson, [IsGroup], Wilf);` is completely unnecessary. But whatever the case, this shouldn't lead to a crash.

This shows that it is most likely the redundant method installation that causes the problem:
```
gap> DeclareAttribute("Wilf", IsGroup);
gap> InstallMethod(Wilf, [IsGroup], x -> fail);
gap> Wilf(SymmetricGroup(3)); # All fine
fail
gap> DeclareSynonymAttr("Wilson", Wilf);
gap> Wilf(SymmetricGroup(3)); # All fine
fail
gap> InstallMethod(Wilson, [IsGroup], Wilf);
gap> Wilf(SymmetricGroup(3));
Segmentation fault: 11
```
Note that doing the method installation the other way round (i.e. `InstallMethod(Wilf, [IsGroup], Wilson);`) also gives the segfault.

fingolfin:
Triggering such a "method" would lead to a segfault due to a stack overflow.

Of course one can still trigger this by e.g. having two operations `foo` and
`bar, and installing `foo` as method for `bar` and vice-versa. But that is far
less likely to happen by accident. Also, it seems better to fix at least some
instances of this issue than to give up because we can't fix it completely.

Essentially resolves #4340 and #1286

Blocked by the fact that at least one method (irredsol) actually installs an operation as method for itself, albeit by accident, and in a fashion that should make it impossible for this method to ever be actually triggered in "normal" situations. Still, this PR will cause this to throw an error, so we can't merge it until irredsol gets a new release with a fix, such as https://github.com/bh11/irredsol/pull/10.


wilfwilson:
IRREDSOL 1.4.2 was released with your fix, will some extra step be required for it to show up in the CI?

fingolfin:
@wilfwilson the package distribution needs to pick it up, @alex-konovalov takes care of that

alex-konovalov:
IRREDSOL picked up (https://github.com/gap-system/gap/wiki/Package-updates-status) but package archives are not yet used for our CI - wait till I update https://github.com/gap-system/gap-distribution/releases/tag/package-archives

wilfwilson:
Looks like it's listed there now.

wilfwilson:
I am struggling to build this PR (because despite the CI now using IRREDSOL, it seems that something is still installing an operation as a method for itself, and I wanted to see which one - that is not obvious from the output of the CI). Could you perhaps rebase, when convenient?

This is the problem I get:
```
   LINK    gap
   LINK    libgap.la
libtool:   error: 'build/obj/src/c_oper1.c.lo' is not a valid libtool object
make: *** [gap] Error 1
make: *** Waiting for unfinished jobs....
libtool:   error: 'build/obj/src/c_oper1.c.lo' is not a valid libtool object
make: *** [libgap.la] Error 1
```

fingolfin:
errrr... I am not sure what happened here, and why the branch was deleted and this PR closed?!? I must have screwed up something. I'll open a fresh PR (as I can't seem to restore this one)

PR #4366, opened 2021-04-06
MatrixObj: improve (New)ZeroMatrix, (New)IdentityMatrix, (New)ZeroVector

fingolfin:
This is work in progress, but I wanted to have it already visible, so that e.g. @ThomasBreuer has a chance to comment if he wants to. Also, so that I can link to this PR from issues :-).

It should actually be mostly ready, but I need to go over everything once, and there are a few things that should be cleaned up, perhaps a few more tests can be added... We'll see.

### DIFF
--- a/hpcgap/lib/vec8bit.gi
+++ b/hpcgap/lib/vec8bit.gi
@@ -1140,6 +1140,9 @@ InstallMethod( BaseField, "for a compressed 8bit vector",
 InstallMethod( NewVector, "for Is8BitVectorRep, GF(q), and a list",
   [ Is8BitVectorRep, IsField and IsFinite, IsList ],
   function( filter, f, l )
+    if not Size(f) in [3..256] then
+        Error("Is8BitVectorRep only supports base fields with 3 to 256 elements");
+    fi;
     return CopyToVectorRep(l,Size(f));
   end );
 
@@ -1147,6 +1150,9 @@ InstallMethod( NewZeroVector, "for Is8BitVectorRep, GF(q), and an int",
   [ Is8BitVectorRep, IsField and IsFinite, IsInt ],
   function( filter, f, i )
     local v;
+    if not Size(f) in [3..256] then
+        Error("Is8BitVectorRep only supports base fields with 3 to 256 elements");
+    fi;
     v := ListWithIdenticalEntries(i,Zero(f));
     ConvertToVectorRep(v,Size(f));
     return v;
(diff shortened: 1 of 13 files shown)


ThomasBreuer:
One aim seems to be to get rid of `ConvertToVectorRep` calls,
to call the responsible low level functions instead where possible,
and to call `CopyToVectorRep` (which was available only in HPCGAP up to now) otherwise.

(The same would make sense for `ConvertToMatrixRep`.)

Concerning the extended argument checks and error messages,
wouldn't it be easier to leave them to `CONV_VEC8BIT`?

fingolfin:
This is now ready for review: I've added more tests, dropped some stuff, added missing variants of ZeroVector, and grouped everything into hopefully logical commits which should make the motivation for each change clearer. To further aid with that, I add the tests in an early commit, and then for later commits containing fixes, the effect of the fix is visible in how it changes the test file.

ThomasBreuer (approved):
I do not want to request changes (thus we can merge this pull request),
but the following thoughts about details should be discussed further.

- Explicitly downranking the `BaseDomain` method for `IsRowVector` is motivated by the fact that otherwise the (better) method for `IsGF2VectorRep` has a lower rank.
I would argue that the methods for `IsGF2VectorRep` should better be installed for `IsRowVector and IsGF2VectorRep` since any object in `IsGF2VectorRep` will be also in `IsRowVector`.
One could define `IsGF2Vector:= IsRowVector and IsGF2VectorRep`, and use this in the method installations.
This "concurrence situation" will not occur for other VectorObj implementations, where the defining filter really determines the method that will be used.

- The removal of the "default method" for `NewZeroVector` is motivated by the fact that constructors cannot have default methods, see the discussion of issue #4398. The documentation which methods are mandatory for a new VectorObj/MatrixObj implementation will depend on the result of this discussion.

fingolfin:
The main reason I dealt with `BaseDomain` using `-SUM_FLAGS` was that the following methods all did the same. But I am happy to have it undone and instead tweak the filters for the `BaseDomain` methods for `Is8BitVectorRep` and `IsGF2VectorRep` as suggested. 

However, I wonder why `IsRowVector` is not automatically implied by `IsGF2VectorRep`... So I had a look at how we define it:
```gap
DeclareRepresentation(
    "IsGF2VectorRep",
    IsDataObjectRep and IsVectorObj, [],
    IsRowVector );
```
This then lead me to submit issues #4408 and #4409.

PR #4373, opened 2021-04-08
Add associativity check to `AsSemigroup` & declare `IsCommutative`/`IsAssociative` for `IsCollection`

### LINKED ISSUE #4030: Non-associative semigroup
How this is possible:
```
gap> T:=[ [ 1, 2, 6, 5, 4, 3 ], [ 2, 1, 4, 2, 2, 5 ], [ 6, 5, 1, 6, 6, 4 ], 
>   [ 5, 6, 3, 4, 1, 2 ], [ 4, 3, 2, 1, 5, 6 ], [ 3, 4, 5, 3, 3, 1 ] ];
gap> Display(T);
[ [  1,  2,  6,  5,  4,  3 ],
  [  2,  1,  4,  2,  2,  5 ],
  [  6,  5,  1,  6,  6,  4 ],
  [  5,  6,  3,  4,  1,  2 ],
  [  4,  3,  2,  1,  5,  6 ],
  [  3,  4,  5,  3,  3,  1 ] ]
gap> M:=MagmaByMultiplicationTable(T);
<magma with 6 generators>
gap> IsAssociative(M);
false
gap> S:=AsSemigroup(M);
<semigroup of size 6, with 3 generators>
gap> MultiplicationTable(S);
[ [ 1, 2, 6, 5, 4, 3 ], [ 2, 1, 4, 2, 2, 5 ], [ 6, 5, 1, 6, 6, 4 ], 
  [ 5, 6, 3, 4, 1, 2 ], [ 4, 3, 2, 1, 5, 6 ], [ 3, 4, 5, 3, 3, 1 ] ]
gap> T=MultiplicationTable(S);
true
```

wilfwilson:
See https://github.com/gap-system/gap/issues/4030.

It might not be fast, but I think it is necessary.

I thought about adding a similar check to `SemigroupByGenerators`, but that (via the documentation for `Semigroup`) is explicitly documented as not checking for associativity.

Fixes #4030.

## Text for release notes 

* Check for  associativity in `AsSemigroup`, and correctly return `fail` for non-associative collections, such as non-associative magmas.
* Declare `IsCommutative` and `IsAssociative` more generally for `IsCollection`, rather than just for `IsMagma`.

### DIFF
--- a/lib/magma.gd
+++ b/lib/magma.gd
@@ -485,14 +485,16 @@ DeclareAttribute( "TrivialSubmagmaWithOne", IsMagmaWithOne );
 
 #############################################################################
 ##
-#P  IsAssociative( <M> )  . . . . . . . . test whether a magma is associative
+#P  IsAssociative( <M> )  . . . . .  test whether a collection is associative
 ##
 ##  <#GAPDoc Label="IsAssociative">
 ##  <ManSection>
 ##  <Prop Name="IsAssociative" Arg='M'/>
 ##
 ##  <Description>
-##  A magma <A>M</A> is <E>associative</E> if for all elements
+##  A collection <A>M</A> of elements that can be multiplied via
+##  <Ref Oper="\*"/>
+##  is <E>associative</E> if for all elements
 ##  <M>a, b, c \in</M> <A>M</A> the equality
 ##  <M>(a</M><C> * </C><M>b)</M><C> * </C><M>c =
 ##  a</M><C> * </C><M>(b</M><C> * </C><M>c)</M> holds.
@@ -507,10 +509,9 @@ DeclareAttribute( "TrivialSubmagmaWithOne", IsMagmaWithOne );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareProperty( "IsAssociative", IsMagma );
+DeclareProperty( "IsAssociative", IsCollection );
 
-InstallTrueMethod( IsAssociative,
-    IsAssociativeElementCollection and IsMagma );
+InstallTrueMethod( IsAssociative, IsAssociativeElementCollection );
 
 InstallSubsetMaintenance( IsAssociative,
     IsMagma and IsAssociative, IsMagma );
--- a/lib/magma.gd
+++ b/lib/magma.gd
@@ -524,7 +524,7 @@ InstallTrueMethod( IsAssociative, IsMagma and IsTrivial );
 
 #############################################################################
 ##
-#P  IsCommutative( <M> )  . . . . . . . . test whether a magma is commutative
+#P  IsCommutative( <M> )  . . . . .  test whether a collection is commutative
 #P  IsAbelian( <M> )
 ##
 ##  <#GAPDoc Label="IsCommutative">
@@ -533,7 +533,9 @@ InstallTrueMethod( IsAssociative, IsMagma and IsTrivial );
 ##  <Prop Name="IsAbelian" Arg='M'/>
 ##
 ##  <Description>
-##  A magma <A>M</A> is <E>commutative</E> if for all elements
+##  A collection <A>M</A> of elements that can be multiplied via
+##  <Ref Oper="\*"/>
+##  is <E>commutative</E> if for all elements
 ##  <M>a, b \in</M> <A>M</A> the
 ##  equality <M>a</M><C> * </C><M>b = b</M><C> * </C><M>a</M> holds.
 ##  <Ref Prop="IsAbelian"/> is a synonym of <Ref Prop="IsCommutative"/>.
@@ -545,12 +547,11 @@ InstallTrueMethod( IsAssociative, IsMagma and IsTrivial );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareProperty( "IsCommutative", IsMagma );
+DeclareProperty( "IsCommutative", IsCollection );
 
 DeclareSynonymAttr( "IsAbelian", IsCommutative );
 
-InstallTrueMethod( IsCommutative,
-    IsCommutativeElementCollection and IsMagma );
+InstallTrueMethod( IsCommutative, IsCommutativeElementCollection );
 
 InstallSubsetMaintenance( IsCommutative,
     IsMagma and IsCommutative, IsMagma );
--- a/lib/semigrp.gd
+++ b/lib/semigrp.gd
@@ -175,6 +175,7 @@ DeclareOperation( "SemigroupByGenerators", [ IsCollection ] );
 ##
 ##  <Description>
 ##  If <A>C</A> is a collection whose elements form a semigroup
+##  under <Ref Oper="\*"/>
 ##  (see&nbsp;<Ref Filt="IsSemigroup"/>) then <Ref Oper="AsSemigroup"/>
 ##  returns this semigroup.
 ##  Otherwise <K>fail</K> is returned.
--- a/lib/semigrp.gi
+++ b/lib/semigrp.gi
@@ -544,6 +544,10 @@ InstallMethod( AsSemigroup,
     function ( D )
     local   S,  L;
 
+    if not IsAssociative( D ) then
+      return fail;
+    fi;
+
     D := AsSSortedList( D );
     L := ShallowCopy( D );
     S := Submagma( SemigroupByGenerators( D ), [] );
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -162,7 +162,7 @@ InstallMethod( IsAssociative,
 
 #############################################################################
 ##
-#M  IsCommutative( <M> )  . . . . . . . . test whether a magma is commutative
+#M  IsCommutative( <M> )  . . . . .  test whether a collection is commutative
 ##
 InstallImmediateMethod( IsCommutative,
     IsMagma and IsAssociative and HasGeneratorsOfMagma, 0,
@@ -196,6 +196,26 @@ InstallMethod( IsCommutative,true,
       TryNextMethod();
     fi;
     end );
+
+InstallMethod( IsCommutative,
+    "for a collection",
+    [ IsCollection ],
+    function( C )
+      local elms, n, x, y, i, j;
+
+      elms:= Enumerator( C );
+      n := Length( elms );
+      for i in [ 1 .. n ] do
+        for j in [ i + 1 .. n ] do
+          x := elms[ i ];
+          y := elms[ j ];
+          if x * y <> y * x then
+            return false;
+          fi;
+        od;
+      od;
+      return true;
+    end);
 
 InstallMethod( IsCommutative,
     "for a magma",
--- a/lib/magma.gi
+++ b/lib/magma.gi
@@ -133,12 +133,11 @@ InstallMethod( IsTrivial,
 
 #############################################################################
 ##
-#M  IsAssociative( <M> )  . . . . . . . . test whether a magma is associative
+#M  IsAssociative( <M> )  . . . . .  test whether a collection is associative
 ##
 InstallMethod( IsAssociative,
-    "for a magma",
-    true,
-    [ IsMagma ], 0,
+    "for a collection",
+    [ IsCollection ],
     function( M )
 
     local elms,      # list of elements
--- a/tst/testbugfix/2021-04-08-non-associative-semigroup.tst
+++ b/tst/testbugfix/2021-04-08-non-associative-semigroup.tst
@@ -1,0 +1,52 @@
+# https://github.com/gap-system/gap/issues/4030
+#
+gap> T := [
+>   [ 1, 2, 6, 5, 4, 3 ],
+>   [ 2, 1, 4, 2, 2, 5 ],
+>   [ 6, 5, 1, 6, 6, 4 ],
+>   [ 5, 6, 3, 4, 1, 2 ],
+>   [ 4, 3, 2, 1, 5, 6 ],
+>   [ 3, 4, 5, 3, 3, 1 ]
+> ];;
+gap> M := MagmaByMultiplicationTable(T);
+<magma with 6 generators>
+gap> IsAssociative(M);
+false
+gap> AsSemigroup(M);
+fail
+gap> M := MagmaByMultiplicationTable(T);
+<magma with 6 generators>
+gap> AsSemigroup(M);
+fail
+gap> AsSemigroup(Elements(M));
+fail
+
+#
+gap> T := [
+>   [ 1, 2, 3, 4, 5 ],
+>   [ 2, 2, 2, 5, 5 ],
+>   [ 3, 2, 3, 4, 5 ],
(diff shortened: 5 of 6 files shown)


wilfwilson:
The "Gather coverage data" step failed on one of the jobs. That's the second time I've witnessed this today. In this case, it was on the `docomp testinstall - BOOTSTRAP_MINIMAL=yes - macos-latest` job; see https://github.com/gap-system/gap/pull/4373/checks?check_run_id=2296430933
```
Run BOOTSTRAP_MINIMAL=yes bash dev/ci-gather-coverage.sh
+ [[ -n '' ]]
+ SRCDIR=/Users/runner/work/gap/gap
+ GAP='bin/gap.sh --quitonbreak --alwaystrace -q'
+ [[ -n '' ]]
+ BUILDDIR=/Users/runner/work/gap/gap
+ COVDIR=coverage
+ ls -l coverage
total 1048808
-rw-r--r--  1 runner  staff  536937058 Apr  8 11:56 testinstall.coverage
+ bin/gap.sh --quitonbreak --alwaystrace -q -a 500M -m 500M -q
 ┌───────┐   GAP 4.12dev built on 2021-04-08 11:49:04+0000
 └───────┘   Architecture: x86_64-apple-darwin19.6.0-default64-kv8
#I  autpgrp package is not available. Check that the name is correct
#I  alnuth package is not available. Check that the name is correct
#I  crisp package is not available. Check that the name is correct
#I  ctbllib package is not available. Check that the name is correct
#I  factint package is not available. Check that the name is correct
#I  fga package is not available. Check that the name is correct
#I  irredsol package is not available. Check that the name is correct
#I  laguna package is not available. Check that the name is correct
#I  polenta package is not available. Check that the name is correct
#I  polycyclic package is not available. Check that the name is correct
#I  resclasses package is not available. Check that the name is correct
#I  sophus package is not available. Check that the name is correct
#I  tomlib package is not available. Check that the name is correct
 Packages:   GAPDoc 1.6.4, PrimGrp 3.4.1, SmallGrp 1.4.2, TransGrp 3.0
gap> #I  profiling package is not available. To see further details, enter
#I  SetInfoLevel(InfoPackageLoading,4); and try to load the package again.
> ERROR: could not load profiling package
> 
Error: Process completed with exit code 1.
```

ChrisJefferson:
Looking at the full log ( https://pipelines.actions.githubusercontent.com/6gZBP4Gju9TDa2Fzvl09zOv8EqapU5UAfeIK6pIOqTBhECmBVy/_apis/pipelines/1/runs/1218/signedlogcontent/4?urlExpires=2021-04-08T13%3A17%3A17.1471533Z&urlSigningMethod=HMACV1&urlSignature=ZulVV8wgYvC63Ks95e48CGLr4kqD8A%2BTGRBzc6ddOGc%3D ), it's io which is failing to build. I'm not 100% sure what the issue is here:

```
==== Checking io
2021-04-08T11:49:56.3701800Z + mkdir -p gen
2021-04-08T11:49:56.3722930Z + autoconf -Wall -f
2021-04-08T11:49:56.6746330Z configure.ac:4: warning: The macro `AC_CONFIG_HEADER' is obsolete.
2021-04-08T11:49:56.6752650Z configure.ac:4: You should run autoupdate.
2021-04-08T11:49:56.6753690Z ./lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...
2021-04-08T11:49:56.6754580Z configure.ac:4: the top level
2021-04-08T11:49:56.6756010Z configure.ac:14: warning: The macro `AC_HEADER_STDC' is obsolete.
2021-04-08T11:49:56.6757030Z configure.ac:14: You should run autoupdate.
2021-04-08T11:49:56.6757980Z ./lib/autoconf/headers.m4:704: AC_HEADER_STDC is expanded from...
2021-04-08T11:49:56.6758860Z configure.ac:14: the top level
2021-04-08T11:49:56.6760160Z configure.ac:24: warning: The macro `AC_HEADER_TIME' is obsolete.
2021-04-08T11:49:56.6761140Z configure.ac:24: You should run autoupdate.
2021-04-08T11:49:56.6762080Z ./lib/autoconf/headers.m4:743: AC_HEADER_TIME is expanded from...
2021-04-08T11:49:56.6763050Z configure.ac:24: the top level
2021-04-08T11:49:56.7688140Z + autoheader -Wall -f
2021-04-08T11:50:01.4933090Z Running './configure' '--with-gaproot=/Users/runner/work/gap/gap' 
2021-04-08T11:50:01.9169440Z configure: error: cannot find required auxiliary files: config.guess config.sub
2021-04-08T11:50:01.9748500Z Running 'make' 'clean' 
2021-04-08T11:50:01.9846550Z make: *** No rule to make target `clean'.  Stop.
2021-04-08T11:50:01.9948710Z Running 'make' 
2021-04-08T11:50:02.0050840Z make: *** No targets specified and no makefile found.  Stop.
2021-04-08T11:50:02.0152060Z 
2021-04-08T11:50:02.0253320Z WARNING: Failed to build io
```

wilfwilson:
Thanks for having a look @ChrisJefferson – I'm going to make an issue here on `gap-system/gap` to keep track of this.

lib/semigrp.gi
```diff
@@ -542,7 +542,25 @@ InstallMethod( AsSemigroup,
...
+    if ApplicableMethod(IsAssociative, [D]) <> fail then
+      # In case there are better methods/known attributes for associativity
+      if not IsAssociative(D) then
+        return fail;
+      fi;
+    else
+      for i in D do
+        for j in D do
+          for k in D do
+            if (i * j) * k <> i * (j * k) then
```
fingolfin, line 557: Would it perhaps make sense to install this as a generic fallback method for  IsAssociative? Then the code in this function could be considerably simplified.

lib/semigrp.gi
```diff
@@ -542,7 +542,25 @@ InstallMethod( AsSemigroup,
     "generic method for collections",
     [ IsCollection ],
     function ( D )
-    local   S,  L;
+    local L, S, i, j, k;
+
+    # Must check associativity
+    if ApplicableMethod(IsAssociative, [D]) <> fail then
```
fingolfin, line 548: Perhaps also
```suggestion
    if HasIsAssociative(D) or ApplicableMethod(IsAssociative, [D]) <> fail then
```

wilfwilson: Is it the case that if something has `HasIsAssociative`, then it necessarily has an applicable method (i.e. the 'getter'?)

fingolfin: Indeed, I think you are right. Might be worthwhile adding this as a comment -- but actually I'd greatly prefer if we found a way to avoid the use of `ApplicableMethod` in the first place :-). But that's just taste, I won't object in the end either way.

lib/semigrp.gi
fingolfin: I think there is an argument to be made the `IsAssociative` (and also `IsCommutative`, and maybe more) also make sense for arbitrary *subsets* of magmas. If one accepts this point of view, it would make sense to install methods for both for `IsMultiplicativeElementCollection`. One could also have `InstallTrueMethod( IsAssociative, IsAssociativeElementCollection );`

However, I am not sure if this has a risk of breaking something? I can't think of anything, though...

Wonder if @ThomasBreuer has some thoughts on this...

wilfwilson: It seems sensible to me... I'll give it a go eventually 🙂

lib/semigrp.gi
fingolfin, line 548: Just to be clear, the only "concern" I have about this PR is that I am wary of using code that relies on `ApplicableMethod` to make decisions.

However, I acknowledge this is a subjective feeling. So I don't want to block this PR over it; I just wanted to make sure we consider alternatives.

fingolfin (approved):

lib/semigrp.gi
wilfwilson: I'm in no rush for this to be merged and I'll be happy to rework the change along the lines discussed above.

But this PR is currently a wrong-result bugfix, so if get close to releasing GAP 4.12.0 and I've still not made the further (non-bugfix) changes, then we should merge this as-is.

ThomasBreuer:
Yes, the associativity test is necessary.
The problem is actually worse:
```
gap> T:= [ [ 1, 2, 6, 5, 4, 3 ], [ 2, 1, 4, 2, 2, 5 ], [ 6, 5, 1, 6, 6, 4 ], 
>          [ 5, 6, 3, 4, 1, 2 ], [ 4, 3, 2, 1, 5, 6 ], [ 3, 4, 5, 3, 3, 1 ] ];;
gap> M:= MagmaByMultiplicationTable( T );
<magma with 6 generators>
gap> IsAssociative( M );
false
gap> AsSemigroup( M );
<semigroup of size 6, with 3 generators>
gap> GroupByGenerators( M ); # AsGroup( M ) returns fail, which is o.k.
<group with 6 generators>
gap> MonoidByGenerators( M ); # AsMonoid(M) runs into an error
<monoid with 6 generators>
gap> Group( Elements( M ) );
<group with 6 generators>
```

wilfwilson:
Thanks for the further information @ThomasBreuer. I'll investigate how that behaviour corresponds to the documentation, and try to add fixed to this PR as appropriate.

ThomasBreuer:
@wilfwilson I have been thinking about installing a default `IsAssociative` method for `IsCollection`, and calling this in the various places where the check is needed. It is not completely clear to me where these places are, apparently several operations are affected.

wilfwilson:
I've now updated this along the lines you suggested @fingolfin, and removed the "do not merge" label. Let's see what the tests and code coverage say.

wilfwilson:
@ThomasBreuer here's how things lie with this PR:
```
gap> T:= [ [ 1, 2, 6, 5, 4, 3 ], [ 2, 1, 4, 2, 2, 5 ], [ 6, 5, 1, 6, 6, 4 ],
>          [ 5, 6, 3, 4, 1, 2 ], [ 4, 3, 2, 1, 5, 6 ], [ 3, 4, 5, 3, 3, 1 ] ];;
gap> M:= MagmaByMultiplicationTable( T );
<magma with 6 generators>
gap> IsAssociative( M );
false
gap> AsSemigroup( M );
fail
gap> AsSubsemigroup( M, M );
fail
gap> GroupByGenerators( M );
<group with 6 generators>
gap> Group( Elements( M ) );
<group with 6 generators>
gap> MonoidByGenerators( M );
<monoid with 6 generators>
gap> SemigroupByGenerators( M );
<semigroup with 6 generators>
```
Shall we open a separate issue about this?

wilfwilson:
This PR now includes a bugfix and an enhancement; I can break them up if you'd like (the bugfix would have to be based on top of the enhancement). As it stands, I wonder if such a combination will confuse the automated release notes creation script...

lib/magma.gd
```diff
@@ -530,7 +532,8 @@ InstallTrueMethod( IsAssociative, IsMagma and IsTrivial );
 ##  <Prop Name="IsAbelian" Arg='M'/>
 ##
 ##  <Description>
-##  A magma <A>M</A> is <E>commutative</E> if for all elements
+##  A multiplicative element collection
+##  <A>M</A> is <E>commutative</E> if for all elements
```
fingolfin, line 536: This sounds a bit awkward to me, how about this instead?
```suggestion
##  A collection of multiplicative elements
##  <A>M</A> is <E>commutative</E> if for all elements
```
or
```suggestion
##  A collection <A>M</A> of multiplicative elements
##  is <E>commutative</E> if for all elements
```

lib/magma.gd
```diff
@@ -483,18 +483,21 @@ DeclareAttribute( "TrivialSubmagmaWithOne", IsMagmaWithOne );
...
 ##  <Description>
-##  A magma <A>M</A> is <E>associative</E> if for all elements
+##  A collection <A>M</A> of multiplicative elements
+##  is <E>associative</E> if for all elements
 ##  <M>a, b, c \in</M> <A>M</A> the equality
 ##  <M>(a</M><C> * </C><M>b)</M><C> * </C><M>c =
 ##  a</M><C> * </C><M>(b</M><C> * </C><M>c)</M> holds.
 ##  <P/>
+##  A collection of multiplicative elements that is closed under the
+##  multiplication is called a <E>magma</E>.
```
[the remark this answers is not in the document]
wilfwilson: I understand. Do you have a suggestion for how I could improve this? Otherwise I might just remove this new sentence - the intention was to be helpful, but now it is potentially confusing.

The real 'problem' here is that this documentation previously only applied to magmas, which is why it was (and still is) in the chapter on magmas, and why magmas didn't need to be defined in this manual entry.

Now that `IsAssociative` applies more generally, perhaps this documentation should instead appear in a different file, and perhaps in a different chapter - what do you think?

lib/semigrp.gi
```diff
@@ -540,10 +540,14 @@ InstallMethod( AsSemigroup,
 
 InstallMethod( AsSemigroup,
     "generic method for collections",
-    [ IsCollection ],
+    [ IsMultiplicativeElementCollection ],
```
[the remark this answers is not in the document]
wilfwilson: Thank you for this explanation. You are correct that this PR breaks your example.

The motivation for this change was that I generalised the declaration of `IsAssociative` from `IsMagma` to `IsMultiplicativeElementCollection`. Since this `AsSemigroup` method now uses `IsAssociative` in turn, I thought that these should agree, as otherwise the `AsSemigroup` method would give a 'no method found' error when it got to the call to `IsAssociative`.

Should I have instead declared `IsAssociative` for `IsCollection`? That means that `IsAssociative( C )` for a collection of (say) digraphs would give a 'no method found for *`' error, but I suppose that is reasonable? This seems like the right approach to me.

lib/semigrp.gd
```diff
@@ -174,8 +174,9 @@ DeclareOperation( "SemigroupByGenerators", [ IsCollection ] );
 ##  <Oper Name="AsSemigroup" Arg='C'/>
 ##
 ##  <Description>
-##  If <A>C</A> is a collection whose elements form a semigroup
-##  (see&nbsp;<Ref Filt="IsSemigroup"/>) then <Ref Oper="AsSemigroup"/>
+##  If <A>C</A> is a collection whose elements can be multiplied via
+##  <Ref Oper="\*"/> and thereby forms a semigroup
```
ThomasBreuer, line 178: Perhaps "thereby" is dangerous here: It is not clear to me whether it shall express an implication (which would be wrong). Why was the old formulation bad?

ThomasBreuer (approved):
This looks good, apart from a suspicious formulation (for me as a non-native speaker).

lib/semigrp.gd
wilfwilson: You're right that it is a suspicious formulation. My intention was to make it clear that the elements have to form a semigroup with respect to the `*` operation (as opposed to the `+` operation, say). I will change it to `whose elements form a semigroup under <Ref Oper="\*"/>`.

Issue #1478, opened 2017-07-05
Long term TODO: ensure cnf/config.{guess,sub} stay up-to-date


fingolfin:
We currently have `cnf/config.{guess,sub}` as part of our repository. The reason is that while for projects using automake, the `automake` resp. `autoreconf` scripts install them -- but `autoreconf` does *not* do that when not using `automake`.

For now we have fairly recent versions in the repository, and I added a script `cnf/update-config.sh` to fetch the latest versions. But perhaps we should somehow automate this... or add a reminder that makes us run the script every couple months?

I guess what I am really asking here is: How can ensure that this and other things that need to be checked rarely but regularly can be taken care of?

In other projects, I'd do this by adding a step to the "release checklist", but we currently don't have one (but see issue #1477 -- if we implemented that, it would immediately allow to resolve this issue).

fingolfin:
We could actually set up a GH Actions workflow which checks for new versions of these files once a month or so, and if found, open up a pull request. We already have workflows which open PRs in the GapWWW and PackageDistro repositories, which one can look at as a template.

fingolfin:
Creating PRs can be done using https://github.com/peter-evans/create-pull-request

PR #4377, opened 2021-04-09
Fix autoreconf 2.71 missing some files

ChrisJefferson:
The release notes for autoconf 2.70 ( https://lwn.net/Articles/839395/ ) mention it may be necessary to run `autoreconf --install`. This should probably be added to each package, but for now we can add it to `build_packages.sh` -- this may be too much of a sledge-hammer, let's first see if this fixes problems at all.


ChrisJefferson:
I (think) that's working now. The main problem is summed up by the paragraph below (quoted from the release notes). In order to get `autoreconf -i` to run on mac, I also had to install `automake` (reading the homebrew pages, while some bits of autoconf, including autoreconf, require automake, they don't automatically install automake if you install autoconf.

```
 As a consequence of improved support for cross compilation (see below),
  more macros now use the auxiliary scripts ‘config.sub’ and ‘config.guess’.
  If you use any of the affected macros, these scripts must be available
  when your configure script is run, even if you have no intention of
  ever cross-compiling your program.

  autoreconf will issue an error if any auxiliary scripts are needed but
  cannot be found.  (It is not currently possible to make autoconf
  itself issue this error.)

  ‘autoreconf --install’ will add ‘config.sub’, ‘config.guess’, and
  ‘install-sh’ to your source tree if they are needed.  If you are
  using Automake, scripts added to your tree by ‘autoreconf --install’
  will automatically be included in the tarball produced by ‘make dist’;
  otherwise, you will need to arrange for them to be distributed
  yourself.

  See the “Input” section of the manual for more detail, including
  where to get the auxiliary scripts that may be needed by autoconf macros.
```

ChrisJefferson:
(there is still a whole bunch of warnings about deprecated things, that might well want fixing some time, but at least right now don't seem to be blocking us running CI, or building packages.)

wilfwilson (approved):
Brilliant

.github/workflows/CI.yml
```diff
@@ -183,7 +183,7 @@ jobs:
                    sudo apt-get update
                    sudo apt-get install "${packages[@]}"
                elif [ "$RUNNER_OS" == "macOS" ]; then
-                   brew install gmp readline zlib
+                   brew install gmp readline zlib automake
```
fingolfin, line 186: No no no! We do not want to pull in automake

bin/BuildPackages.sh
```diff
@@ -186,6 +186,11 @@ run_configure_and_make() {
   if [[ -x autogen.sh && ! -x configure ]]
   then
     ./autogen.sh
+    # (Hopefully) temporary fix for Autoconf 2.71
```
fingolfin, line 189: We can try to simply update config.sub and config.guess, that's something we have to do occasionally anyway (see issue #1478). There are other changes to be made in configure.ac. I'll look into that. But all of this must keep working with autoconf 2.69, which is what almost everyone has. In a quick test of mine, it does so.

fingolfin (changes requested):
I'll submit an alternate PR.

But we still need to also deal with this for packages. I actually started looking into this yesterday, when I run into this independently for some GAP packages. I should have notified you folks to avoid duplication of effort, sorry.

bin/BuildPackages.sh
fingolfin: I think the issue was actually specific to the io package, which actually had a buggy configure.ac that happened to work in autoconf 2.69, and now 2.70 detected it. To be precise: I set it up to not check for a "canonical host type", but then used autoconf macros which actually require that; in autoconf 2.69 this sailed through (and possibly resulted in buggy behavior, though this unlikely on any modern day OS). In autoconf 2.70 it detects this requirement and computes the canonical host type, which require config.guess and config.sub ... hence the issue

fingolfin (changes requested):
I really don't think this is required (nor appropriate), but I am open to have my mind changed about this :-)

PR #4381, opened 2021-04-09
Update config.guess and config.sub

fingolfin:
This should still work with 2.69, which is the most wide spread version. We'll see.

See also  #1478

### DIFF
--- a/cnf/config.guess
+++ b/cnf/config.guess
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2019 Free Software Foundation, Inc.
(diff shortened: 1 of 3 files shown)


wilfwilson:
I'm happy to merge in the updated helper scripts, but I don't see why it should close #4377, because the purpose of that PR is to make the macOS GitHub Actions job pass, which this PR does not do.

wilfwilson (approved):
Thanks for sorting this out @fingolfin.

fingolfin:
You are right, this was a red herring. I've updated the PR title and removed the "closes XYZ"

fingolfin:
I still think it's useful to update these scripts, though, the last update was 2019-10-24

Issue #4384, opened 2021-04-09
`BuildPackages.sh --strict` no longer works


fingolfin:
_Previous title:_ `CI: failure to compile IO, profiling should cause the relevant GitHub Actions step to fail`

Right now, it doesn't: In https://github.com/gap-system/gap/pull/4381/checks?check_run_id=2305769116 it fails to compile `io` but just goes on. And that despite using `BuildPackages.sh --strict` which should give us a non-zero exit code. Perhaps that got broken, or perhaps we are not correctly passing it on, or...

wilfwilson:
It is the fault of `BuildPackages.sh`. If I deliberately break the profiling package and then call
```
../bin/BuildPackages.sh --strict profiling* && echo $?
```
from my `pkg` dir, then the final line of output is `0`. Here's the last few lines of output:
```
WARNING: Failed to build profiling

Packages which failed to build are in ./log/fail.log
0
```
Indeed, the `if [[ $STRICT ]]; then exit 1; fi` bit of code occurs immediately after the code that prints the `WARNING:` line, so that error code is only being caught internally; the script then goes on to print the "Packages which failed to build..." line and exit happily.

I'll see if I can fix it.

wilfwilson:
This was introduced in #3922, which (according to my very limited understanding of bash...) put the relevant bit in a subshell whose result was ignored - something like that.

PR #4386, opened 2021-04-09
Restore some functionality of the `--strict` option to the `bin/BuildPackages.sh` script

wilfwilson:
See https://github.com/gap-system/gap/issues/4384.

The addition of the `--parallel` option to the `bin/BuildPackages.sh` script in #3922 unfortunately broke the operation of the `--strict` option. This is because the compilation of each package was put into a background process, and its exit code was no longer being checked.

This restores the functionality of `--strict` **when `--parallel` is not used**. I am not sure how to catch the error codes
of parallel background processes – so, if would like help with this, please.

## Text for release notes 

Restore the broken `--strict` option to `BuildPackages.sh`, but only if `--parallel` is not also specified.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -23,13 +23,6 @@
 
 set -e
 
-# Is someone trying to run us from inside the 'bin' directory?
-if [[ -f gapicon.bmp ]]
-then
-  error "This script must be run from inside the pkg directory" \
-        "Type: cd ../pkg; ../bin/BuildPackages.sh"
-fi
-
 CURDIR="$(pwd)"
 GAPROOT="$(cd .. && pwd)"
 COLORS=yes
@@ -63,22 +56,6 @@ while [[ "$#" -ge 1 ]]; do
   esac
 done
 
-if [ "x$PARALLEL" = "xyes" ]; then
-  export MAKEFLAGS="${MAKEFLAGS:--j3}"
-fi;
-
-# If user specified no packages to build, build all packages in subdirectories.
-if [[ ${#PACKAGES[@]} == 0 ]]
-then
-  # Put package directory names into a bash array to avoid issues with
-  # spaces in filenames. This code will still break if there are newlines
-  # in the name.
-  old_IFS=$IFS
-  IFS=$'\n' PACKAGES=($(find . -maxdepth 2 -type f -name PackageInfo.g))
-  IFS=$old_IFS
-  PACKAGES=( "${PACKAGES[@]%/PackageInfo.g}" )
-fi
-
 # Some helper functions for printing user messages
 if [[ "x$COLORS" = xyes ]]
 then
@@ -94,6 +71,28 @@ else
   std_error() { printf "%s\n" "$@" ; }
 fi
 
+# Is someone trying to run us from inside the 'bin' directory?
+if [[ -f BuildPackages.sh ]]
+then
+  error "This script must be run from inside the pkg directory" \
+        "Type: cd ../pkg; ../bin/BuildPackages.sh"
+fi
+
+if [ "x$PARALLEL" = "xyes" ]; then
+  export MAKEFLAGS="${MAKEFLAGS:--j3}"
+fi;
+
+# If user specified no packages to build, build all packages in subdirectories.
+if [[ ${#PACKAGES[@]} == 0 ]]
+then
+  # Put package directory names into a bash array to avoid issues with
+  # spaces in filenames. This code will still break if there are newlines
+  # in the name.
+  old_IFS=$IFS
+  IFS=$'\n' PACKAGES=($(find . -maxdepth 2 -type f -name PackageInfo.g))
+  IFS=$old_IFS
+  PACKAGES=( "${PACKAGES[@]%/PackageInfo.g}" )
+fi
 
 notice "Using GAP root $GAPROOT"
 
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -78,6 +78,10 @@ then
         "Type: cd ../pkg; ../bin/BuildPackages.sh"
 fi
 
+if [ "x$PARALLEL" = "xyes" ] && [ "x$STRICT" = "xyes" ]; then
+  error "The options --strict and --parallel cannot be used simultaneously"
+fi
+
 if [ "x$PARALLEL" = "xyes" ]; then
   export MAKEFLAGS="${MAKEFLAGS:--j3}"
 fi;
@@ -268,7 +272,7 @@ build_one_package() {
 date >> "$LOGDIR/fail.log"
 for PKG in "${PACKAGES[@]}"
 do 
-  ( 
+  (  # start a background process
   # cut off the ending slash (if exists)
   PKG="${PKG%/}"
   # cut off everything before the first slash to only keep the package name
@@ -306,6 +310,7 @@ do
     warning "$PKG does not seem to be a package directory, skipping"
   fi
   ) &
+  BUILD_PID=$!
   if [ "x$PARALLEL" = "xyes" ]; then
     # If more than 4 background jobs are running, wait for one to finish (if
     # <wait -n> is available) or for all to finish (if only <wait> is available)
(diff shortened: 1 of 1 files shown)


wilfwilson:
Evidence that this works is given by the failing macOS job now actually failing at the appropriate point.

But... urgh, unfortunately the Cygwin job is now also failing, because the `profiling` package is not compiling there. With the broken `--strict` option, that wasn't a problem, because that job was never actually using `profiling` (I presume)...

ChrisJefferson:
I started fixing this, but the fix grew kept growing, and I'm not convinced it's going to work correctly on all systems (as there can be subtle differences between bash versions). Therefore my suggestion is to just remove this bit of parallel code that builds packages in parallel.

We can leave in the--parallel option and have it still turn on  "MAKEFLAGS=-j3", which still helps speed up packages like semigroups (which is where most of the time is spent), the only loss is parallelising running things like configure scripts.

Anyone who really wants to build packages in parallel can do it themselves still with something like this (which we only don't use as you have to install "parallel")

`ls | parallel ../bin/BuildPackages.sh`

wilfwilson:
I think that's a reasonable solution.

Would this also be a reasonable solution?: Accept this PR, with an additional check at the start of the script that at most one of `--parallel` and `--strict` is given, exiting with an error otherwise (and perhaps giving a message that directs the user to GNU parallel?). On their own, with this PR, the two options work as intended; it's just in combination that there's a problem.

Note that (at least my impression is that) the `BuildPackages.sh` script is somehow 'unofficial', so I think we're free to do what we want with it.

Somewhat related: is there any reason _not_ to do `MAKEFLAGS=-j3` by default?

bin/BuildPackages.sh
```diff
@@ -315,7 +316,10 @@ do
     fi
   else
     # wait for this package to finish building
-    wait
+    if ! wait $BUILD_PID && [[ $STRICT = yes ]]
```
fingolfin, line 319: So my understanding of shell job control is fairly limited, so the following is possibly "obvious" if one knows this stuff, but it isn't to me: I don't understand, why do we need `$BUILD_PID`? Why doesn't this work:
```suggestion
    if ! wait && [[ $STRICT = yes ]]
```

wilfwilson: `wait` always returns `0` if called without a process ID as argument:
```
$ help wait
wait: wait [n]
    Wait for the specified process and report its termination status.  If
    N is not given, all currently active child processes are waited for,
    and the return code is zero. [...]
```

fingolfin: Aaaah of course!

so could we adapt https://stackoverflow.com/questions/356100/how-to-wait-in-bash-for-several-subprocesses-to-finish-and-return-exit-code-0#356154 ?

wilfwilson: I'm not sure – this seems complicated, as @ChrisJefferson has found.

`BuildPackages.sh --parallel` currently runs only <= 4 jobs in the background, and (by default) uses `wait -n` (which is not supported on my version of bash, see #4291) to wait for _any one_ of those jobs to finish, before despatching one more job. (`wait -n` _does_ pass on the exit code of the job that finished... but surely you'd miss checking some exit codes, if you relied on check them this way...)

This is as opposed to not using `wait -n` but just normal `wait` and waiting for _all 4 jobs_ to terminate before spawning any more, which on first glance seems to be what would be required of that Stack Overflow answer.

Alternatively, we could run all jobs simultaneously in the background without a limit, but I presume this is undesirable (it is what @ChrisJefferson initially implemented, before moving to a limit of 4).

wilfwilson:
By the way, in my opinion we should just use `GNU parallel` in `BuildPackages.sh --parallel` if it makes things simpler:

> Note that (at least my impression is that) the BuildPackages.sh script is somehow 'unofficial', so I think we're free to do what we want with it.

Sure, the `--parallel` option won't work on everyone's computer, but then the purpose of `BuildPackages.sh` isn't to work on everyone's computer, its purpose is to be useful!

fingolfin:
I'd be happy to require GNU parallel if the script properly checks for its presence and prints a helpful error message if it isn't there.

.github/workflows/CI.yml
```diff
@@ -246,6 +246,7 @@ jobs:
 
     env:
       TEST_SUITES: testinstall
+      NO_COVERAGE: 1
```
fingolfin, line 249: Why is this here?

wilfwilson: Ah! Thank you for noticing this @fingolfin. I can't remember.

But I realised that the `test-cygwin` job runs neither of the scripts `dev/ci-gather-coverage.sh` and `dev/ci-gather-coverage.sh`, so I thought: what's the point of tracking code coverage if we then don't do anything with it? And I thought that including `NO_COVERAGE: 1` (meaning `io` and `profiling` no longer need to be compiled) would speed up the Cygwin job, in which compilation is slow.

...But of course, such a change (or indeed the opposite, actually gathering and uploading the coverage data in Cygwin) should be in a separate PR 🙂.

wilfwilson: Removed.

fingolfin (approved):
Thanks for the explanations!

So, my understanding is that strict mode is currently broken if `--parallel` is used. Can we then at least check for both being used at the start of the script, and produce an error that this is currently not supported?

fingolfin:
So this fails now in the cygwin64 job -- and this is issue existed before, and is revealed by this PR here (so yeah, it does its job). See e.g. https://github.com/gap-system/gap/pull/4373/checks#step:5:1067 for another PR where the same issue occurs, it just isn't reported.

This is the error:
```
==== Checking profiling-2.4.1
Running './configure' '/cygdrive/d/a/gap/gap' 
Using settings from /cygdrive/d/a/gap/gap/sysinfo.gap
Running 'gmake' 'clean' 
Makefile.gappkg:41: /cygdrive/d/a/gap/gap/sysinfo.gap: No such file or directory
gmake: *** No rule to make target '/cygdrive/d/a/gap/gap/sysinfo.gap'.  Stop.
Running './configure' '/cygdrive/d/a/gap/gap' 
Using settings from /cygdrive/d/a/gap/gap/sysinfo.gap
Running 'gmake' 
Makefile.gappkg:41: /cygdrive/d/a/gap/gap/sysinfo.gap: No such file or directory
gmake: *** No rule to make target '/cygdrive/d/a/gap/gap/sysinfo.gap'.  Stop.
```

This may be a hint at the "new(ish)" package build system based on `Makefile.gappkg` has issues on Windows / cygwin, but I've never seen reports about that before...

Anyway, this also means it would really be useful to have cygwin tests for packages, too, at least for "central" ones like `io` and `profiling`.

wilfwilson:
I've started to investigate Cygwin tests for IO: https://github.com/gap-packages/io/pull/97

wilfwilson:
I have pushed a change which throws an error if `--strict` and `--parallel` are used together.

I also fixed the part where the script tests that it is not being run from in GAP's `bin` directory. This was doubly broken: it was using the function `error` before it had been defined, and it was doing this by testing for a file `gapicon.bmp` which (as far as I can tell) is no longer present in GAP, certainly not in `bin` anyway.

fingolfin:
@wilfwilson wow, thanks for catching that.

Regarding https://github.com/gap-packages/io/pull/97, I left a few comments there, but overall io seems to work -- in any case, it builds fine, and so does it here. What does not build is the `profiling` package. I guess one could start by adapting your IO patch for profiling and then see if it builds there, and if not, use that as starting point for further debugging of the issue.

wilfwilson:
I was thinking the same on my bike ride this morning.

fingolfin:
The `profiling` build failure is weird. The file `/cygdrive/d/a/gap/gap/sysinfo.gap` definitely exists; indeed, `configure` verifies that it exists and is readable. But then `gmake` says "No such file or directory"....

Hmmm, which `gmake` is that, I wonder -- perhaps it's *not* the one from `cygwin`? That could maybe explain it... Indeed, looking at the cygwin package `make`, it contains `make.exe` but not `gmake.exe` (see <https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Fmake%2Fmake-4.3-1&grep=make>).

One way to deal with that would be to locate the "bad" gmake.exe and simply remove it.

wilfwilson:
@ChrisJefferson fixed the failing Cygwin job via https://github.com/gap-actions/setup-cygwin/pull/7 – I've re-run the CI and the Cygwin job passed! So when the rest of the CI has passed, I will merge this, and open a feature request issue asking for `--strict` and `--parallel` to be supported simultaneously (perhaps with `GNU parallel`).

Issue #4374, opened 2021-04-08
Sporadic failure to build IO in the macOS CI job


wilfwilson:
I first mentioned this https://github.com/gap-system/gap/pull/4372#issuecomment-815747822 and https://github.com/gap-system/gap/pull/4373#issuecomment-815739007.

Starting recently (I first noticed it today, I've seen it about four times so far) the `Gather coverage data` step sometimes fails in the macOS job (`docomp testinstall - BOOTSTRAP_MINIMAL=yes - macos-latest`), because it cannot load the profiling package. @ChrisJefferson has looked a little further and noticed that this is because `IO` fails to build. From https://github.com/gap-system/gap/pull/4373#issuecomment-815817324:
> Looking at the full log, it's io which is failing to build. I'm not 100% sure what the issue is here:
> ```
> ==== Checking io
> 2021-04-08T11:49:56.3701800Z + mkdir -p gen
> 2021-04-08T11:49:56.3722930Z + autoconf -Wall -f
> 2021-04-08T11:49:56.6746330Z configure.ac:4: warning: The macro `AC_CONFIG_HEADER' is obsolete.
> 2021-04-08T11:49:56.6752650Z configure.ac:4: You should run autoupdate.
> 2021-04-08T11:49:56.6753690Z ./lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...
> 2021-04-08T11:49:56.6754580Z configure.ac:4: the top level
> 2021-04-08T11:49:56.6756010Z configure.ac:14: warning: The macro `AC_HEADER_STDC' is obsolete.
> 2021-04-08T11:49:56.6757030Z configure.ac:14: You should run autoupdate.
> 2021-04-08T11:49:56.6757980Z ./lib/autoconf/headers.m4:704: AC_HEADER_STDC is expanded from...
> 2021-04-08T11:49:56.6758860Z configure.ac:14: the top level
> 2021-04-08T11:49:56.6760160Z configure.ac:24: warning: The macro `AC_HEADER_TIME' is obsolete.
> 2021-04-08T11:49:56.6761140Z configure.ac:24: You should run autoupdate.
> 2021-04-08T11:49:56.6762080Z ./lib/autoconf/headers.m4:743: AC_HEADER_TIME is expanded from...
> 2021-04-08T11:49:56.6763050Z configure.ac:24: the top level
> 2021-04-08T11:49:56.7688140Z + autoheader -Wall -f
[...]

PR #4387, opened 2021-04-10
CI: temporarily remove coverage from macOS job

wilfwilson:
Currently the macOS job is failing because IO is not building, which is requiring by the code coverage system, and this causes the job to fail. See #4374.

By adding `NO_COVERAGE=1` to this job, IO will no longer need to be compiled. Obviously, in the long term we want coverage on this job, and we want to be able to compile IO, but this will get the CI running for the rest of GAP in a non-controversial way in the meantime, I hope.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -86,7 +86,7 @@ jobs:
 
           - os: macos-latest
             test-suites: "docomp testinstall"
-            extra: "BOOTSTRAP_MINIMAL=yes"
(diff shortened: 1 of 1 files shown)


fingolfin:
Or wait a day or two when I'll have a chance to push out the fixes to io and any other affected package.

but if you think it's this urgent... *shrug*

wilfwilson:
_shrug_ from me too 🙂. It's not particularly urgent, but then again it took almost no effort to create this one-line PR, and now it stands ready as a backup in case those fixes end up taking longer than expected.

PR #4389, opened 2021-04-11
Automate creating Windows installers for GAP releases with GitHub Actions

wilfwilson:
_(Phew, this was a lot of work, building on what must've been a lot of work by @ChrisJefferson (who was building on a lot of work by @embray and perhaps others at Sage...) and the others involved in the creation of GAP's existing GitHub Actions release system, and I'm sure many others too. Thank you everyone! Open source software collaboration can be nice. :slightly_smiling_face:)_

This PR uses [my fork](https://github.com/wilfwilson/sage-windows)* of [@ChrisJefferson's fork](https://github.com/ChrisJefferson/sage-windows) of [SageMath's sage-windows](https://github.com/sagemath/sage-windows) repository to create Windows installers for GAP, and in particular, when a new version of GAP is released this creates an installer from the `gap-X.Y.Z.tar.gz` archive produced by the current release system, and adds it to the GitHub release.

It adds a new `cygwin` job to the `releases` workflow, along with two new short Python scripts in `dev/releases` that it uses. The existing `releases` job is renamed to `unix`. The new workflow job is heavily commented, so if you are reviewing, please read those comments carefully, in which I describe what the steps are doing, and in which I describe several possible changes/improvements to the way that it works (which would possibly remove the need for the two new Python scripts).

In summary:
* This new `cygwin` job runs _after_ each run of the `unix` job (which creates all of the GAP/package tarballs and zip files).
* In all cases except the PR, the job runs twice: once to make a 32-bit GAP installer, and once to make a 64-bit GAP installer.
* In more detail:
    * If `on: pr`, it wraps the merge commit of the GAP PR, with only the required packages, into a 64-bit installer.
    * If `on: push: tags`, it wraps the `gap-X.Y.Z.tar.gz` that it gets from the just-created GitHub release, and uploads the installers to the release.
    * Otherwise, it wraps the head of the appropriate branch, with the packages given by `make bootstrap-pkg-full`. If the trigger is `on: schedule`, then it also uploads the installers as artifacts (with a retention period of 1 day)

---
_* I have opened a PR https://github.com/ChrisJefferson/sage-windows/pull/1 to merge my fork into Chris's fork, and eventually Chris's fork will be put somewhere more appropriate._

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,26 +4,32 @@
 #
 # For builds triggered by a tag, the tag is turned into a GitHub release and
 # the produced archives are attached to that.
-name: release
+name: "Wrap releases"
 
-# Trigger the workflow on push or pull request
 on:
+  workflow_dispatch:
   pull_request:
   push:
     tags: v[1-9]+.[0-9]+.[0-9]+
     branches:
       - master
       - stable-*
+  schedule:
+    # Every day at 3:33 AM UTC
+    - cron:  '33 3 * * *'
+
+env:
+  NO_COVERAGE: 1
+  BOOTSTRAP_MINIMAL: yes
 
 jobs:
-  release:
-    name: Release
+  unix:
+    name: "Create UNIX archives and data"
     # Don't run this twice on PRs for branches pushed to the same repository
     if: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) }}
     runs-on: ubuntu-latest
-    env:
-      NO_COVERAGE: "1"
-      BOOTSTRAP_MINIMAL: "yes"
+    outputs:
+      cygwin-matrix: ${{ steps.set-cygwin-matrix.outputs.matrix }}
 
     steps:
       - uses: actions/checkout@v2
@@ -39,7 +45,7 @@ jobs:
       - name: "Set up Python"
         uses: actions/setup-python@v2
       - name: "Install Python modules"
-        run: pip3 install PyGithub requests python-dateutil
+        run: pip3 install PyGithub python-dateutil
       - name: "Install latex"
         run: sudo apt-get install texlive texlive-latex-extra texlive-extra-utils texlive-fonts-extra
       - name: "Compile GAP and download packages"
@@ -61,5 +67,184 @@ jobs:
       - name: "Make GitHub release"
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
         run: python -u ./dev/releases/make_github_release.py ${GITHUB_REF#refs/tags/} tmp/
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Decide which Cygwin installers to build next"
+        id: set-cygwin-matrix
+        run: |
+          if [[ $GITHUB_EVENT_NAME == 'pull_request' ]]; then
+            CYGWIN_MATRIX='{"arch":["x86_64"]}'
+          else
+            CYGWIN_MATRIX='{"arch":["x86_64","x86"]}'
+          fi
+          echo "Setting Cygwin matrix: $CYGWIN_MATRIX"
+          echo "::set-output name=matrix::$CYGWIN_MATRIX"
+
+  cygwin:
+    name: "Create Windows ${{ matrix.arch }} installer"
+    needs: unix
+    runs-on: windows-latest
+    env:
+      CHERE_INVOKING: 1
+
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.unix.outputs.cygwin-matrix) }}
+
+    steps:
+      # If the event is a pushed release tag v1.2.3, then VERSION=1.2.3
+      # If the event is a PR #123, then VERSION=pr-123
+      # Otherwise VERSION=branch, where <branch> is either the pushed branch, or
+      # the branch selected for workflow_dispatch, or master (for a cron job).
+      # In all cases, the GAP to be wrapped is put into gap-$VERSION/
+      #
+      # The name of this directory is currently required for the sage-windows
+      # script, which is passed the value of $VERSION (via SAGE_VERSION)
+      #
+      # If the event is a pushed release tag, then we are wrapping a version of
+      # GAP that already has its packages in place and all manuals compiled.
+      # Therefore sage-windows only needs to compile GAP and the packages.
+      #
+      # Otherwise, (currently) we are wrapping a cloned version of GAP.
+      # In none of these cases do we build GAP's manuals, because it's too slow.
+      # If the event is a PR, then we `make bootstrap-pkg-minimal` to save time
+      # (none of the required packages requires compilation).
+      # If the event is a pushed branch, then we `make bootstrap-pkg-full`.
+      # Compiling these packages takes a long time.
+      - name: "Set some environment variables according to the GitHub context"
+        shell: bash
+        run: |
+          if [[ $GITHUB_REF == refs/tags/v* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/v}
+            COMPILEGAP="make -j2"
+            GAPDEV_DIR="gap-dev"
+          else
+            if [[ $GITHUB_EVENT_NAME == 'pull_request' ]]; then
+               VERSION=${GITHUB_REF#refs/pull/}
+               VERSION="pr-${VERSION%/merge}"
+               COMPILEGAP="make -j2 && make bootstrap-pkg-minimal"
+            elif [[ $GITHUB_REF == refs/heads/* ]]; then
+              VERSION=${GITHUB_REF#refs/heads/}
+              COMPILEGAP="make -j2 && make bootstrap-pkg-full"
+            else
+              echo "Unrecognised GitHub situation"
+              exit 1
+            fi
+            GAPDEV_DIR="gap-$VERSION"
+          fi
+
+          echo "Using data: VERSION=$VERSION / GAPDEV_DIR=$GAPDEV_DIR / COMPILEGAP=$COMPILEGAP"
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+          echo "GAPDEV_DIR=$GAPDEV_DIR" >> $GITHUB_ENV
+          echo "SAGE_RUN_CONFIGURE_CMD=\"cd \$(SAGE_ROOT) && ${COMPILEGAP}\"" >> $GITHUB_ENV
+
+      # In all cases, we currently need to clone GAP for its release scripts.
+      #
+      # If the GitHub event is a pushed release tag, then we compile/wrap a GAP
+      # tarball that we download from the GitHub release that was just created
+      # and uploaded by the preceding 'unix' job.
+      #
+      # Otherwise, we compile and wrap the appropriate GAP development clone
+      # into the installer.
+      #
+      # Alternatively, we could upload the desired GAP tarball via the
+      # 'actions/upload-artifact' action in the preceding 'unix' job, and then
+      # use this in all cases in this 'cygwin' job. This would save some
+      # repeated work and hence time, and unify the code here slightly.
+      #
+      # Furthermore, this would remove the need for the
+      # download_release_archive.py script, and if we would also remove the need
+      # for the upload_file_to_github_release.py script, as described below,
+      # then we would no longer need access to the GAP release scripts in this
+      # job at all. Therefore, in all cases, we would be able to remove:
+      # * Clone GAP
+      # * Copy GAP's release scripts
+      # * Set up Python and its modules
+      #   - Although we would still need a way to create the EXE checksum files
+      # * The existence of the GAPDEV_DIR environment variable
+      # * The existence of the COMPILEGAP environment variable, because in all
+      #   cases it would be just "make -j2", since the downloaded release
+      #   artifact would/could already contain the appropriate packages.
+
+      - name: "Clone GAP"
+        uses: actions/checkout@v2
+        with:
+          path: ${{ env.GAPDEV_DIR }}
+
+      - name: "Copy GAP's release scripts to a safe place"
+        shell: bash
+        run: cp -rp ${GAPDEV_DIR}/dev/releases .
+
+      - uses: actions/setup-python@v2
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+
+      - name: "Install required Python modules"
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: pip3 install PyGithub
+
+      # We could probably use either a GitHub Action from the GitHub Marketplace
+      # to download the appropriate GAP tarball from the GitHub release, rather
+      # than creating and using the Python file.  This would probably mean that
+      # we lose the ability to verify checksums, but that is probably fine.
+      - name: "Download the appropriate GAP release tarball"
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        shell: bash
+        run: |
+          python -u ./releases/download_release_archive.py v${VERSION} gap-${VERSION}.tar.gz .
+          tar -zxf gap-${VERSION}.tar.gz
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # TODO When the PR https://github.com/ChrisJefferson/sage-windows/pull/1
+      # is merged, then should be updated to ChrisJefferson/sage-windows.
+      # Ultimately, this repository should be owned by one of the GAP-related
+      # organisations, and perhaps ultimately it could/should be re-integrated
+      # with the original sagemath/sage-windows repository.
+      - name: "Clone the Windows installer maker scripts"
+        uses: actions/checkout@v2
+        with:
+          repository: wilfwilson/sage-windows
+          ref: master
+          path: sage-windows
+
+      - uses: gap-actions/setup-cygwin-for-gap@v1
+
+      # Currently, the sage-windows/release_gap.sh script wraps the GAP
+      # contained in gap-$VERSION, and outputs its installer to
+      # sage-windows/Output/gap-${{ env.VERSION }}-$ARCH.exe
+      #
+      # TODO:
+      # * Investigate how to speed this up. e.g. if we don't need to compile
+      #   GAP's manuals then could we perhaps avoid installing TeXLive?
+      # * Investigate splitting release_gap.sh into multiple scripts so that
+      #   this big step can be split up into multiple steps
+      - name: "Compile GAP and its packages, and create the installer"
+        shell: C:\cygwin64\bin\bash.exe --login --norc -o igncr '{0}'
+        run: |
+          cd $GITHUB_WORKSPACE/sage-windows
+          bash release_gap.sh
+        env:
+          ARCH: ${{ matrix.arch }}
+          SAGE_BUILD_DOC_CMD: '"true"'
+          SAGE_VERSION: ${{ env.VERSION }}
+
+      # Artifacts live for 1 day, i.e. until the next cron job runs.
+      - name: "Upload the installer as an artifact"
+        if: ${{ github.event_name == 'schedule' }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: gap-${{ env.VERSION }}-${{ matrix.arch }}.exe
+          path: sage-windows/Output/gap-${{ env.VERSION }}-${{ matrix.arch }}.exe
+          retention-days: 1
+
+      # To reduce code, we could use an Action from the GitHub Marketplace to
+      # upload to the release, rather than use our own script.  We would still
+      # have to create the .sha256 file, and upload it here too.
+      - name: "Upload the installer to the GitHub release"
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        shell: bash
+        run: |
+          python -u ./releases/upload_files_to_github_release.py v${VERSION} sage-windows/Output/gap-${{ env.VERSION }}-${{ matrix.arch }}.exe
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/dev/releases/download_release_archive.py
+++ b/dev/releases/download_release_archive.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+
+# This script is supposed to download the release asset with name <asset> to
+# the directory <download_dir> from the GitHub release at utils.CURRENT_REPO
+# with the tag <tag_name>.
+
+# It checks that <download_dir> exists, that <tag_name> is a release, and that
+# <asset> is the name of a file attached to the release.
+
+import os
+import sys
+import utils
+
+if len(sys.argv) != 4:
+    utils.error("usage: "+sys.argv[0]+" <tag_name> <asset> <download_dir>")
+
+TAG_NAME = sys.argv[1]
+ASSET = sys.argv[2]
+DOWNLOAD_DIR = sys.argv[3]
+
+if not os.path.isdir(DOWNLOAD_DIR):
+    utils.error(f"{DOWNLOAD_DIR} seems not to be a directory")
+utils.verify_is_possible_gap_release_tag(TAG_NAME)
+utils.initialize_github()
+
+try:
+    RELEASE = utils.CURRENT_REPO.get_release(TAG_NAME)
+except:
+    utils.error(f"{utils.CURRENT_REPO_NAME} contains no release {TAG_NAME}")
+
+assets = [ x for x in RELEASE.get_assets() if ASSET == x.name ]
+if len(assets) != 1:
+    utils.error(f"No unique file {FILE} exists on the release {TAG_NAME}")
+
+utils.download_with_sha256(assets[0].browser_download_url, os.path.realpath(DOWNLOAD_DIR) + "/" + ASSET)
--- a/dev/releases/upload_files_to_github_release.py
+++ b/dev/releases/upload_files_to_github_release.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+
+# This script is supposed to upload the files at <path_to_file1> [...] to the
+# GitHub release at utils.CURRENT_REPO with the tag <tag_name>.
+
+# It checks that each <path_to_file1> (etc) is a file, that <tag_name> is a
+# release, and that there are not already any files with the same name attached
+# to the release.
+
+import os
+import sys
+import utils
+
+if len(sys.argv) < 3:
+    utils.error("usage: "+sys.argv[0]+" <tag_name> <path_to_file1> [...]")
+
+TAG_NAME = sys.argv[1]
+ASSETS = sys.argv[2:]
+
+for ASSET in ASSETS:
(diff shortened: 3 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -63,3 +63,178 @@ jobs:
...
+          tar -zxf gap-${VERSION}.tar.gz
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # TODO When the PR https://github.com/ChrisJefferson/sage-windows/pull/1
+      # is merged, then should be updated to ChrisJefferson/sage-windows.
+      # Ultimately, this repository should be owned by one of the GAP-related
+      # organisations, and perhaps ultimately it can be re-integrated with
+      # the original sagemath/sage-windows repository.
+      - name: "Clone the Windows installer maker scripts"
```
wilfwilson, line 195: This could use `actions/checkout@v2` instead.

dev/releases/make_archives.py
```diff
@@ -130,37 +141,40 @@
...
+    notice("Building GAP's manuals")
     run_with_log(["make", "doc"], "gapdoc", "building the manuals")
 
     notice("Removing unwanted version-controlled files")
     badfiles = [
     ".appveyor.yml",
+    ".clang-format",
     ".codecov.yml",
+    ".ctags",
+    ".editorconfig",
```
wilfwilson, line 174: Is it correct to remove these files from a GAP release? I was unsure, please let me know if I should remove anything from this list.

.github/workflows/release.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: "Set up Python"
         uses: actions/setup-python@v2
       - name: "Install Python modules"
-        run: pip3 install PyGithub requests python-dateutil
+        run: pip3 install PyGithub python-dateutil
```
wilfwilson, line 42: `PyGithub` requires `requests`, so we don't need to separately install it. And anyway, we don't even use the `requests` module directly in any of our release scripts, except for `update-website.py`, but that script isn't run in this workflow. So it's simpler to just remove it.

.github/workflows/release.yml
```diff
@@ -63,3 +63,178 @@ jobs:
         run: python -u ./dev/releases/make_github_release.py ${GITHUB_REF#refs/tags/} tmp/
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  cygwin:
+    name: "Create Windows ${{ matrix.arch }} installer"
+    needs: unix
+    runs-on: windows-latest
+    # TODO Do we want to skip these jobs for pull requests?  If so, add the following line, and delete the PR-specific code from the steps below.
+    # if: ${{ github.event_name != 'pull_request' }}
```
ssiccha, line 72:
```suggestion
      if: ${{ github.event_name != 'pull_request' }}
```
I'd say let's not run this on every pull request. It takes 1h 20mins compared to the CI tests' 50 mins. Maybe instead add a `workflow_dispatch:` to the triggers of the workflow? Then whenever a PR touches the release system, we could manually trigger the full workflow, that is including the windows installer jobs. (In theory, we could also let the full workflow be triggered automatically on every PR with an `topic: infrastructure` label.)

.github/workflows/release.yml
```diff
@@ -63,3 +63,178 @@ jobs:
...
+          SAGE_VERSION: ${{ env.VERSION }}
+
+      # TODO Decide the situation in which one might wish to upload the
+      # installer as an artifact.
+      - name: "Upload the installer as an artifact"
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: gap-${{ env.VERSION }}-${{ matrix.arch}}.exe
+          path: sage-windows/Output/gap-${{ env.VERSION }}-${{ matrix.arch }}.exe
```
ssiccha, line 228: Since the artifacts are huge and may stay around for a bit, I say let's not upload artifacts for every commit and every PR. 

If we want to run this workflow on every commit and PR, I'd say let's never upload artifacts. Or let's only upload them on failure, but then there probably won't be anything to upload anyways. If there are failures, there always is the "debugging via tmate" action.

If we decide to only run the full workflow when triggered manually, then in that case uploading the artifacts should be fine IMO.

ssiccha (changes requested):
First of all, this looks like it was a ton of work @wilfwilson @ChrisJefferson. 

Disclaimer: I haven't looked into the sage-windows repository for more than 30 seconds. 

I've left a few minor comments. Besides from those, everything looks fine to me.
I also added the label infrastructure.

.github/workflows/release.yml
wilfwilson: FYI, you can specify how long to keep artifacts, with a minimum of 1 day: https://github.com/actions/upload-artifact#retention-period – this might be useful on some/all of our workflows; see also https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy

wilfwilson: Note that even the logs of this workflow are BIG - 50MB or so, I think.

fingolfin:
Looking at https://github.com/wilfwilson/gap/actions/runs/738243905 it took about ~1 hour to produce the windows binaries. So I am wary about running this too often.

Looking at the x86 builds, it takes about 10 minutes of preparation before it gets to building GAP. I wonder if the GH Actions cache system could be used to help with that?

Compiling all packages: some of them fail (e.g. CddInterface), so once #4386 is merged, that may lead to failures here (?).  

I think compilation of all packages then finishes after another 30 minutes, so roughly 40 minutes in total were spent so far. The log visible on the website then ends as is too long; it says I can download the full log, but I didn't do that yet.

So I wonder if https://ccache.dev works on Windows with Cygwin; if we could use it and use the GitHub cache system on the ccache, it might reduce the compilation time. Just a thought, though, it may be too much of a hassle to get working.


BTW, in the x86 logs, I saw this (didn't look at the x86_64 logs; those numbers are line numbers, copy&paste for me always includes them, (un)fortunately); not sure if it is "normal" or we should worry about it:
```
1485
# Move the tmpdir into the final environment location
1486
Sun, 11 Apr 2021 13:29:38 GMT
mkdir -p envs
1487
Sun, 11 Apr 2021 13:29:38 GMT
mv /tmp/tmp.FemAgZRzTO envs/build-stable-cygwin-x86
1488
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/pki/ca-trust/source/ca-bundle.legacy.crt': Permission denied
1489
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/hosts': Permission denied
1490
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/protocols': Permission denied
1491
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/services': Permission denied
1492
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/networks': Permission denied
1493
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/alternatives/automake-doc': Permission denied
1494
Sun, 11 Apr 2021 13:29:38 GMT
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/alternatives/python': Permission denied
1495
Sun, 11 Apr 2021 13:29:38 GMT
```

.github/workflows/release.yml
fingolfin: Agreed, and I prefer the manual trigger over label magic (we don't use labels very consistently)

.github/workflows/release.yml
ssiccha: Ooh nice. So then instead of not uploading artifacts I'm for keeping them around for one day.

ChrisJefferson:
My general feeling is there shouldn't be need to run this more than once a release, and maybe also once a week, so I'm not too worried about the time, also, any type of caching is dangerous with cygwin because if you cache binaries built against a different version of the cygwin DLL, bad things happen.

I do want to look at the remaining packages which aren't building yet -- first I'd like to get testing of all packages working in Cygwin, so I could easily see what is / isn't working, as I suspect some packages might be building but not running.

.github/workflows/release.yml
ssiccha: As @ChrisJefferson suggested, we can also run this once a week. Of course it would be nice to have slack notifications so we don't miss when something breaks..

fingolfin:
I don't think running this just once per release is enough, we *must* run this regularly if we want to notice regressions quickly (experience tells that it then is much easier to resolve them).

Running it on each update of `master` may still be too slow, though; 2x 1 hour Windows jobs (which are much more limited than Unix jobs) could be a killer during "busy days". But we could of course try it out for now... otherwise, setting a daily cron trigger would also work.

In either case, we badly need CI failure notifications again, see issue #4390.


Another question: how much work would it be to split the "Compile GAP and its packages, and create the installer" step up? E.g. into these:
1. setup cygwin (actually... didn't we do that before? It seems to do it *again*)
2. build GAP
3. build packages
4. create the installed

That would help with pinpointing issues and generally make the GH Actions log a bit easier to navigate.

dev/releases/make_archives.py
fingolfin: Note that I intentionally left in .clang-format, .editoconfig and .ctags precisely because someone might edit the code in there and submit the result as a patch to us -- in that case, I'd prefer if they used settings similar to what we ask for.

That said, of course it's not clear whether people who don't use our git repo to make a PR would then go on to use clang-format (plus, these days that's not going to be many people, if any at all).

So, I guess then removing these is OK *shrug*

dev/releases/make_archives.py
```diff
@@ -130,37 +141,40 @@
...
-            file.write(package_infos.encode('utf-8'))
+            file.write(package_infos.stdout.encode('utf-8'))
+        manifest_list.append("package-infos.json.gz")
 
+    notice("Cleaning up the json package")
     with working_directory(path_to_json_package):
         subprocess.run(["make", "clean"], check=True)
-        subprocess.run(["rm", "-rf", "bin/", "Makefile"],
-                       check=True)
+        subprocess.run(["rm", "-rf", "bin/", "Makefile"], check=True)
```
fingolfin, line 162: Aside: we really should use `shutil.rmtree` or `os.removedirs` instead of calling `rm -rf` (and `os.remove` to delete `Makefile`)

dev/releases/make_archives.py
```diff
@@ -114,14 +120,19 @@
     notice(f"PKG_MINIMAL = {PKG_MINIMAL}")
     notice(f"PKG_FULL = {PKG_FULL}")
 
+    # FIXME Huh? Currently we just take the pkg manuals as they come...
     # Downloading, building and extracting pkgs manuals
```
fingolfin, line 124:
```suggestion
    # Download, build and extract manuals of all packages
```

dev/releases/make_archives.py
```diff
@@ -114,14 +120,19 @@
...
 
-    notice("Extract the packages")
-    with tarfile.open("../"+all_packages_tarball) as tar:
+    notice("Extracting package tarballs")
+    with tarfile.open(tmpdir+"/"+all_packages_tarball) as tar:
         tar.extractall(path="pkg")
+    with tarfile.open(tmpdir+"/"+req_packages_tarball) as tar:
+        tar.extractall(path=tmpdir+"/"+req_packages)
 
     # TODO: at this point we could generate a JSON file which collects the metadata of
```
fingolfin, line 137: This TODO comment is obsolete and could be removed/revised

dev/releases/make_archives.py
wilfwilson: Sure. They're hidden files anyway by default so they're barely 'clutter', and possibly useful. I'll remove the diff the next time I push.

.github/workflows/release.yml
```diff
@@ -63,3 +63,178 @@ jobs:
...
+      matrix:
+        arch:
+          - x86
+          - x86_64
+
+    steps:
+      - name: "Set git to use UNIX-style line endings"
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
```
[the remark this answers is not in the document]
wilfwilson: I have no idea, I've just been copy-and-pasting it around. I'll try commenting it out and we'll see what happens.

.github/workflows/release.yml
wilfwilson: I agree that there should be a `on: workflow_dispatch:` trigger (indeed I think all of our workflows should have that trigger). I think the same about having `on: schedule`  triggers too (although there we have to decide the frequency, and which branch(es) should be tested on any regular schedule).

I will respond to the other topic(s) in a separate comment in the main thread.

.github/workflows/release.yml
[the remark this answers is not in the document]
wilfwilson: Simply removing this steps causes the workflow to fail. In particular, the `release_gap.sh` script starts with:
```
:\a\gap\gap>SET CHERE_INVOKING=1 

D:\a\gap\gap>cd sage-windows 

D:\a\gap\gap\sage-windows>C:\cygwin64\bin\bash -l -c 'bash release_gap.sh' 
release_gap.sh: line 2: $'\r': command not found
<builtin>: target '.o' does not exist
cc      -o .o
cc: fatal error: no input files
compilation terminated.
make: *** [<builtin>: .o] Error 1
release_gap.sh: line 4: SAGE_GIT?=https://github.com/gap-system/gap: No such file or directory
release_gap.sh: line 5: SAGE_VERSION?=master: command not found
release_gap.sh: line 6: SAGE_MAKE_CONFIGURE_CMD?="cd $(SAGE_ROOT) && ./autogen.sh && ./configure": No such file or directory
release_gap.sh: line 7: SAGE_RUN_CONFIGURE_CMD?="cd $(SAGE_ROOT) && make && make bootstrap-pkg-full": command not found
release_gap.sh: line 8: $'\r': command not found
release_gap.sh: line 9: $'\r': command not found
release_gap.sh: line 10: $'\r': command not found
release_gap.sh: line 11: $'\r': command not found
release_gap.sh: line 12: $'\r': command not found
release_gap.sh: line 13: $'\r': command not found
release_gap.sh: line 14: SAGE_BUILD_DOC_CMD?="cd $(SAGE_ROOT) && make doc": command not found
release_gap.sh: line 15: $'\r': command not found
release_gap.sh: line 16: $'\r': command not found
release_gap.sh: line 17: $'\r': command not found
release_gap.sh: line 18: SAGEMATH_ISS?=gap.iss: command not found
release_gap.sh: line 19: $'\r': command not found
release_gap.sh: line 20: $'\r': command not found
release_gap.sh: line 21: $'\r': command not found
Error: Process completed with exit code 127.
```

According to https://github.com/actions/checkout/issues/135#issuecomment-613361104 we would need `.gitattributes` to contain `* text eol=lf` (it is currently `* text=auto`) but that fails: https://github.com/wilfwilson/gap/runs/2325655805?check_suite_focus=true 

In particular, the command `verify_git_clean()` of the `dev/releases/make_archives.py` script fails (presumably because this setting causes the line endings of some files to be changed when GAP is checked out?).

I'm not going to investigate this further.

wilfwilson:
I'm happy for there to be manually-dispatched workflow runs and scheduled workflow runs. That should be no problem, I'll implement those sometime, and we discuss the precise parameters that they should have.

---

In my opinion, the details of how the script of the `sage-windows` repository achieves its goals (i.e. the details of how `release_gap.sh` does its thing) are mostly independent of this PR.

The purpose of this PR is to implement a GitHub Actions Workflow that:
* at an appropriate point in time (schedule? manual? release? PR? branch push?)...
* gives the appropriate version of GAP to the `release_gap.sh` script...
* and appropriately handles the installers that it makes.

I think that the comments and details relating to the steps involved in the `release_gap.sh` script should be directed towards https://github.com/ChrisJefferson/sage-windows - in particular, we should primarily have those discussion on my PR to that repo https://github.com/ChrisJefferson/sage-windows/pull/1.  Note that that repository already has CI set up, so you can see over there what the `release_gap.sh` script is doing (it currently takes a couple of hours longer, because it also compiles GAP's manuals).

wilfwilson:
I will now respond on a high level to some of the discussions about resource usage and making things faster.

The summary of my initial suggestions for making this faster: Only build the 64-bit installer most of the time, and teach `ChrisJefferson/sage-windows` to skip installing LaTeX in certain cases (and possibly skip installing some other Cygwin components).

Other comments:

* (I believe) that the `release_gap.sh` script *needs* to be run from within Cygwin.  Therefore we need to set up Cygwin in this Workflow before calling the script.  Currently, we do that with `gap-actions/setup-cygwin-for-gap`, which is overkill, because we're not actually going to run GAP in this Cygwin, but this step 'only' takes two minutes and I don't have the desire to work out which subset of Cygwin is actually necessary.

* The remaining significant time taken by this Workflow is in the script `release_gap.sh`. As mentioned above, I think most discussions related to this script would be more appropriately held on https://github.com/ChrisJefferson/sage-windows/pull/1.

* The `release_gap.sh` script sets up its own new Cygwin with very specific properties (it possibly does this twice). I therefore doubt that we can re-use the Cygwin that we initially used to call the `release_gap.sh` script. In particular, this spends a lot of time installing TeX Live, which I believe we should be able to skip for the purposes of this Workflow, because (as far as I understand), we don't need to compile GAP's manuals in this workflow.

From https://github.com/gap-system/gap/pull/4389#issuecomment-817732082 (@fingolfin):
> We must run this regularly if we want to notice regressions quickly...

I agree!

From https://github.com/gap-system/gap/pull/4389#discussion_r611456157 (@ssiccha):
> I'd say let's not run this on every pull request. It takes 1h 20mins compared to the CI tests' 50 mins.

No, under the current setup a GitHub Actions run of `release.yml` on a _pull request_ takes ~42 minutes (see e.g.  https://github.com/gap-system/gap/actions/runs/738246766), of which ~18 minutes is the already-existing release scripts, and then that is followed by the parallel Windows installer jobs, which each take about ~24 minutes. Which in my opinion is a reasonable length of time. We could further reduce our Windows usage by only building on one architecture (i.e. 64-bit) on PRs.

It takes less time on a PR because there I only do `make bootstrap-pkg-minimal`, and so there are no packages to compile, whereas on a branch push I do `make bootstrap-pkg-full` and there are lots of packages that are slow to compile.

From https://github.com/gap-system/gap/pull/4389#issuecomment-817717621 (@fingolfin):
> Compiling all packages: some of them fail (e.g. CddInterface), so once #4386 is merged, that may lead to failures here (?).

#4386 only causes `BuildPackages.sh` to exit non-zero if you use the option `--strict`. But the `release_gap.sh` script uses `BuildPackages.sh --parallel`.

From https://github.com/gap-system/gap/pull/4389#issuecomment-817721840 (@ChrisJefferson):
> I do want to look at the remaining packages which aren't building yet...

Sure – that's great 🙂!

I don't mind waiting for all packages to be working in Cygwin before merging this. But at the same time, I don't think broken packages should stop this being merged, because the Workflow runs successfully even with broken packages - it just means that those packages won't work properly in the resulting Windows GAP installation.

From https://github.com/gap-system/gap/pull/4389#issuecomment-817732082 (@fingolfin):
> Another question: how much work would it be to split the "Compile GAP and its packages, and create the installer" step up? 

I agree that it would be great to have this. But I think it would be a large amount of work, although I don't know. `release_gap.sh` consists of just a single call to `make`. So the `Makefile` would need to be restructured. I have zero desire to get involved in this in the near future, and would make further significant diffs against `sagemath/sage-windows`. @ChrisJefferson will know better about this, so I suggest further discussions on this topic should move to https://github.com/ChrisJefferson/sage-windows/pull/1

wilfwilson:
Oops, forgot one. @fingolfin those lines of the form
```
mv: listing attributes of '/tmp/tmp.FemAgZRzTO/etc/services': Permission denied
```
also appear in the CI of https://github.com/ChrisJefferson/sage-windows, so they're not particular to this PR. I've got no idea of their importance.

wilfwilson:
Hmm, annoyingly, when I implemented @ssiccha's change to re-add `.ctags` into the release archives, the script now fails when extracting the `gap-4.X.Y.tar.gz` archive in the Windows bit of the Workflow (before then putting it into an installer), see e.g. https://github.com/wilfwilson/gap/runs/2326455805?check_suite_focus=true
```
tar: gap-4.12.14/.ctags: Cannot create symlink to 'etc/ctags_for_gap': No such file or directory
tar: Exiting with failure status due to previous errors
```
This is because `.ctags` is simply a symlink to `etc/ctags_for_gap` and... Windows doesn't seem to support symlinks 😱

fingolfin:
Just remove the .ctags symlink from the archive. Nobody will really miss it, or at least not enough people to justify loosing hair about it

wilfwilson:
I’ve already lost all of mine 😉

ssiccha (approved):
> No, under the current setup a GitHub Actions run of release.yml on a pull request takes ~42 minutes (see e.g. https://github.com/gap-system/gap/actions/runs/738246766), of which ~18 minutes is the already-existing release scripts, and then that is followed by the parallel Windows installer jobs, which each take about ~24 minutes. Which in my opinion is a reasonable length of time. We could further reduce our Windows usage by only building on one architecture (i.e. 64-bit) on PRs.

Ok, that sounds great! I have no idea whether our usage of windows machines may become a bottleneck during GAP days. So I'm undecided on whether to include building 32 bit releases on every PR. However, we can still make that minor change in an extra PR.

wilfwilson:
I think homebrew was having trouble yesterday, hopefully it'll work the next time I push.

.github/workflows/release.yml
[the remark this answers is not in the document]
fingolfin: Thank you!

fingolfin (approved):
Thank you very much for taking care of this!

fingolfin:
Since it's just 24 minutes for a PR build, I'd be fine with just enabling this on all PRs after all. Anyway, feel free to restore that, or merge this as-is.

wilfwilson:
Hmm, I'm not sure if I'm able to easily conditionally run only one of the two jobs specified by the 'matrix' (so that I can only create 64-bit installers for PRs).

So the solution might be a bit hacky. I could manually add an `if:` conditional to all of the steps to stop them from being run for in the case that the architecture is 32-bit and the event is a PR; or I could just add those to the steps that take more than a few seconds. There are probably other solutions.

wilfwilson:
I got the `workflow_dispatch` and `schedule` triggers to work fine (https://github.com/wilfwilson/gap/actions/workflows/release.yml?query=event%3Aschedule and https://github.com/wilfwilson/gap/actions/workflows/release.yml?query=event%3Aworkflow_dispatch).

The artifacts are currently only being uploaded on the daily scheduled job, and only living for one day (i.e. roughly until the next job runs). I might upload artifacts for `workflow_dispatch` too (`schedule` only builds the master branch; `workflow_dispatch` can build any branch that contains the workflow, I think).

I've still got PR jobs disabled for now until I work out how to best only build on architecture. I might enable them in a separate PR.

But the reason that I've still got the "DO NOT MERGE" label is that I have not actually tried running the resulting Windows installer(s). I don't currently have access to a Windows computer. But I'll work out a way, possibly ask some people to help.

ssiccha:
> Hmm, I'm not sure if I'm able to easily conditionally run only one of the two jobs specified by the 'matrix' (so that I can only create 64-bit installers for PRs).

> So the solution might be a bit hacky. I could manually add an if: conditional to all of the steps to stop them from being run for in the case that the architecture is 32-bit and the event is a PR; or I could just add those to the steps that take more than a few seconds. There are probably other solutions.

What about this: add a first step to the cygwin release job which runs `exit 0` if `github.event_name` is `pull_request` and if `arch` is `x86`?

wilfwilson:
No, that just exits that step successfully (indeed, exiting with a return code of 0 is exactly what you hope to happen in each step). It does't skip all of the remaining steps.

ssiccha:
Oh yes, woops. :joy:

ssiccha:
What about putting the cygwin release job into its own composite action? You could put it into a file `release-cygwin.yml` and call it in `release.yml` via `using: ./release-cygwin` IIRC.

Then you could add two jobs: one which always compiles the 64 bit release and one which compiles the 32 bit release only if you are not in a PR.

wilfwilson:
Thanks, but I think I'd rather not do that.

Another option (simpler, indeed it takes less time overall, but more time from start to finish) would be to build the two installers one after the other in the same job, rather than building them in parallel as two matrix jobs.

So:
* Get rid of the matrix
* The step `"Compile GAP and its packages, and create the installer"` is either duplicated (one for each `ARCH`) or it contains a `for` loop (running over the two `ARCH`s).
* Then we can easily skip building the second installer, which now just becomes a single step.

fingolfin (commented):
I am unsure what the state of this PR is now. What is left to be done, exactly? Is a decision needed on something, or is it just a matter of finding time to implement some already made decisions?

It may also help to take this in steps: e.g. some changes in this PR could probably already go in, e.g. those to `make_archives.py` look to me as if they are mostly cleanup and could mostly go in right away?

Regarding the 32 vs 64 bit issue: right now, I think we only test Cygwin 64 bit; so it would be a reasonable option if we for now just built the 64bit installer. And then open a separate PR/issue regarding the 32bit one. This way, we would reap the majority of the benefit now and postpone the icky technical discussions... I hope, at least? Of course that's just a suggestion, and it might be a bad one, I do not claim to fully understand the problem scope

wilfwilson:
Thanks for checking in. It hasn't changed much since you looked at it before (mostly the addition of the `cron` and `manual` workflow triggers), and there isn't much more to do.

There are two small things I want to do before merging:
* Only build the 64-bit installers on PRs (shouldn't be too difficult)
* Ask someone to actually install one of the Windows installers on Windows to check that it isn't at least completely broken.

I'll do at least some of this today.

.github/workflows/release.yml
```diff
@@ -63,3 +66,182 @@ jobs:
...
```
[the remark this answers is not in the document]
wilfwilson: Not necessarily, because we're also cloning and using the `sage-windows` repository. I'll try adding the same `.gitattributes` to that repository.

wilfwilson:
I think this is _very nearly_ ready to be merged.

The jobs run on PR (64-bit only), push branch/tag, on a daily schedule (on the master branch), and manually. For a release tag, it also upload the installers to the GitHub release.

You can see an up-to-date example of the workflow in this PR running:
* As a PR, here: https://github.com/gap-system/gap/actions/runs/771225157
* As a pushed branch on my fork: https://github.com/wilfwilson/gap/actions/runs/771069724
* As a pushed tag `v4.12.19` on my fork: https://github.com/wilfwilson/gap/actions/runs/770998659
    * The resulting GitHub release for `v4.12.19`, including assets on my fork: https://github.com/wilfwilson/gap/releases/tag/v4.12.19

I have asked @ChrisJefferson to install both installers from that `v4.12.9` release on his Windows computer, and to have a quick check that the GAPs load and at least superficially seem to work.

wilfwilson:
@ChrisJefferson tried this installers and (at least on a quick check!) it all seemed in order. So I'll merge!

PR #4392, opened 2021-04-12
CI: remove AppVeyor integration

### LINKED ISSUE #4382: CI: stop using AppVeyor
We now test Cygwin with Github Actions, too, and we have more parallel jobs there (AppVeyor only has 1 job in its free tier). I think it's much better to have all the CI status data in one place.

fingolfin:
Resolves #4382 

Well, in addition, we also need to remove the AppVeyor integration from https://github.com/gap-system/gap/settings/installations and perhaps more


wilfwilson (approved):

wilfwilson:
The Mac job failed because of `homebrew`; I had problems with it earlier today too (unrelated to GAP) so I think their servers might be struggling a bit.

You'll also need to remove the branch protection rules concerning the AppVeyor build.

PR #4393, opened 2021-04-13
Slack notifications for GH Actions

### LINKED ISSUE #4390: CI: restore CI failure Slack notifications
With Travis we used to get notifications for CI failures on Slack. We need this again, so that we notice when master or stable jobs start failing.

In order for this to maximize usefulness and minimize annoyance, in the past we configured this to only ping us on "edges": I.e. there was a notification when the build started failing; and one when it stopped failing. I guess for master/stable-* branches, it would also be OK to get a notification for each failure. But we do *not* want a notification for each success.

I suggest using a new Slack channel `#status` for this.

ssiccha:
Adds a job to the CI workflow which is triggered whenever the test-unix or the
test-cygwin jobs fail. That job then sends a message into the #status channel
of the gap-system slack workspace.

For a test run on my fork see:
- [a run without errors](https://github.com/ssiccha/gap/actions/runs/744321276)
- [a run where I added an error to hpcgap](https://github.com/ssiccha/gap/actions/runs/744348237)

Fixes #4390

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -274,3 +274,31 @@ jobs:
         run: |
           SET CHERE_INVOKING=1
           C:\cygwin64\bin\bash -l -c "bash dev/ci.sh"
+
+  slack-notification-on-failure:
+    name: Send slack notification on CI failure
+    needs:
+      - test-unix
+      - test-cygwin
+    if: ${{ always() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get branch name
+        id: get-branch
+        run: echo ::set-output name=branch::${GITHUB_REF#refs/*/}
+      - name: Determine whether CI status changed
+        uses: gap-actions/should-i-notify-action@v1
+        id: should_notify
+        with:
+          branch: ${{ steps.get-branch.outputs.branch }}
+          needs_context: ${{ toJson(needs) }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          notify_on_changed_status: true
+      - name: Send slack notification
+        uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
+        if: steps.should_notify.outputs.should_send_message == 'yes'
+        with:
+          status: ${{ steps.should_notify.outputs.last_status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}


wilfwilson:
Are you aware of issue #4390? Obviously if this works at all then that's great and is much bette than nothing (I haven't looked at the diff yet). But _ideally_ we'd only like Slack notifications when a branch goes from failing to passing, or vice versa.

ssiccha:
Ah damn, I didn't see the new posts to #4390 from yesterday.

wilfwilson:
How does this work with respect to forks? If I break the master branch in my fork of GAP, is that going to post in the #status channel? Or will that step fail because I don't have the correct access token set (or something like that?).

ssiccha:
> Or will that step fail because I don't have the correct access token set (or something like that?).

That's exactly what's going to happen.

ssiccha:
Ok, it looks like I only have to add the action https://github.com/marketplace/actions/should-i-notify-slack to the new job and we should be good. Cool! :)

.github/workflows/CI.yml
```diff
@@ -274,3 +274,21 @@ jobs:
         run: |
           SET CHERE_INVOKING=1
           C:\cygwin64\bin\bash -l -c "bash dev/ci.sh"
+
+  slack-notification-on-failure:
+    name: Send slack notification on CI failure
+    needs:
+      - test-unix
+      - test-cygwin
+    if: ${{ failure() && github.event_name == 'push' }}
```
wilfwilson, line 283: Would something like be sensible (and would it work)?
```suggestion
    if: ${{ failure() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
```

ssiccha: I think that makes sense, yes. I'll have to add that once I'm done testing.

ssiccha:
So _if_ what I pushed so far works, then I think we have to add similar jobs to the release workflows.

.github/workflows/CI.yml
```diff
@@ -274,3 +274,32 @@ jobs:
...
```
wilfwilson, line 283: Although I don't think we want `failure()`, because we want to also be notified when a previously-failing branch is now passing. So I *think* this should probably be `always()`?

.github/workflows/CI.yml
```diff
@@ -274,3 +274,32 @@ jobs:
...
+    # Don't run this twice on PRs for branches pushed to the same repository
+    #if: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) }}
```
wilfwilson, line 285:
```suggestion
```

.github/workflows/CI.yml
```diff
@@ -274,3 +274,32 @@ jobs:
...
+        id: should_notify
+        with:
+          branch: ${{ steps.vars.outputs.branch }}
+          needs_context: ${{ toJson(needs) }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Send slack notification
+        uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
+        if: steps.should_notify.outputs.should_send_message == 'yes'
+        with:
+          status: Failure
```
wilfwilson, line 303: It shouldn't always be failure, the status might also be success (if the failing branch has been fixed). Judging by your helpful Markdown file, it should instead look like:
```suggestion
          status: ${{ job.status }}
```

wilfwilson (changes requested):
Hmm, yes, good point - we do also want to be notified when the release scripts fail. So either this should be duplicated at some point (probably the best solution), or the two workflow files should be combined into one.

.github/workflows/CI.yml
ssiccha: Yes, totally!

.github/workflows/CI.yml
ssiccha: Hmm, good point. Although resolving that might be a bit more complicated. IIRC, if I put `job.status` there, then the status of the `slack-notification-on-failure` job is used, which always is success. I'll have to look into that later.

ssiccha:
Except for passing the correct status of the branch to the slack message, in principle everything seems to work. I made a minimal version of this slack notification job and put it here:
https://github.com/ssiccha/gap/tree/ss/minimal-slack-test
Results of the actions can be found here:
https://github.com/ssiccha/gap/actions?query=branch%3Ass%2Fminimal-slack-test++
The interesting part is the step "Send slack notification" of the job "slack test". That one gets skipped using the `Jimdo/should-i-notify-action` action.

.github/workflows/CI.yml
```diff
@@ -274,3 +274,30 @@ jobs:
...
+      - name: Determine whether CI status changed
+        uses: Jimdo/should-i-notify-action@0f48901c482c651cba9dc19a03efc068b402789b
+        id: should_notify
+        with:
+          branch: ${{ steps.vars.outputs.branch }}
+          needs_context: ${{ toJson(needs) }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Send slack notification
+        uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
+        if: steps.should_notify.outputs.should_send_message == 'yes'
```
wilfwilson, line 299: This doesn't seem to work (two failed jobs in a row at https://github.com/ssiccha/gap/actions both sent the message) - perhaps it needs this syntax?
```suggestion
        if: ${{ steps.should_notify.outputs.should_send_message == 'yes' }}
```

.github/workflows/CI.yml
wilfwilson: Cool, yeah - no rush. I'm not sure what the permitted values of `status` are; if they can be anything, then maybe something like this could be made to work?
```suggestion
          status: ${{ failure() }}
```

Otherwise maybe something like this could work?
```
(Step to determine whether to post a message)

# Step to determine which message to send
- name:
  run: |
    if ${{ failure() }} then
      # set env variable STATUS
    elif ${{ success() }}
      # set env variable STATUS to be success
    fi

# Step to post the message
- uses: (whatever) 
  ...
  with:
    status: ${{ env.STATUS }}
```

wilfwilson:
To avoid duplicating this code in the releases workflow file, would it be possible to make this new bit into its own composite action (perhaps at `gap-actions`)? Although I'm not sure if composite actions can themselves call composite actions, which this would need to do.

.github/workflows/CI.yml
ssiccha: Two succeeded jobs in a row didn't sent a message though. So in principle it works I think. OTOH the action https://github.com/Jimdo/should-i-notify-action states in its README that it always returns `yes`, if the action failed. So I guess we have to extend or change that. I've checked it out and the action itself is not too complicated. I have an appointment now though, so I'll have to get back to this another time.

ssiccha:
> would it be possible to make this new bit into its own composite action
I think that _should_ be possible. :thinking:

dev/slack-notifications-for-github-actions.md
```diff
@@ -0,0 +1,21 @@
+# Slack notifications for GitHub Actions
+
+Max Horn, Chris Jefferson, Alexander Konovalov, Sergio Siccha, and Wilf Wilson
+have access to the app `GAP GH Actions Notifications`. If you have access, then
```
fingolfin, line 4: Good idea to document this somewhere; should also (instead?) be on https://github.com/gap-system/gap/wiki/Developer-resources (which in turn also needs to be updated). Or perhaps that file in turn should be part of the GAP repo.

@alex-konovalov @ChrisJefferson thoughts?

fingolfin: I just updated the Developer-resources Wiki page a bit, but I am sure I missed things

.github/workflows/CI.yml
wilfwilson: Ah I didn't realise. Thanks for the clarification.

.github/workflows/CI.yml
ssiccha: Yes, we have to use `needs.<job>.result`, then it works! :) See 2b9c913e9ab78ee6ef

.github/workflows/CI.yml
ssiccha: I've now made a fork of the should-i-notify-action. See my new post below.

dev/slack-notifications-for-github-actions.md
ssiccha: I've removed this file and added an according wiki page.

ssiccha: And just added links to that wiki page.

.github/workflows/CI.yml
wilfwilson: Cool 🙂 So according to https://github.com/ssiccha/gap/blob/ss/minimal-slack-test/.github/workflows/test-slack.yml it should be:
```suggestion
          status: ${{ steps.get-status.outputs.status }}
```

ssiccha: Oh, yes I forgot to push that ^^

ssiccha:
@fingolfin @wilfwilson when you have time, could you have a look again whether this is OK with you?

In the meantime, I'll try to put this into its own action.

wilfwilson (approved):

wilfwilson:
Thanks for doing this @ssiccha 🙂 I'm looking forward to having this.

.github/workflows/CI.yml
```diff
@@ -274,3 +274,40 @@ jobs:
...
+          if [[ ${{ needs.test-unix.result }} = "failure" || ${{ needs.test-cygwin.result }} = "failure" ]]; then
+            status="failure"
+          fi
+          echo ::set-output name=status::$status
+      - name: Determine whether CI status changed
+        uses: gap-actions/should-i-notify-action@v1
+        id: should_notify
+        with:
+          branch: ${{ steps.vars.outputs.branch }}
+          needs_context: ${{ toJson(needs) }}
```
fingolfin, line 304: The `needs` context is used to determine whether there was a failure. You are also computing that same thin in the `get-status` step above...

Can you perhaps add an output to `gap-actions/should-i-notify-action` which indicates the status? Then you could get rid of the `get-status` step, .

.github/workflows/CI.yml
```diff
@@ -274,3 +274,40 @@ jobs:
...
+    name: Send slack notification on CI failure
+    needs:
+      - test-unix
+      - test-cygwin
+    if: ${{ always() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get branch name
+        id: get-branch
```
fingolfin, line 288: Seems this is unused? You access the output of this step via `steps.vars.outputs.branch` below (I did not check whether that works, though)

.github/workflows/CI.yml
ssiccha: I've updated the action accordingly.

.github/workflows/CI.yml
ssiccha: Yes, that must have been broken. I'm updating this. I didn't realise because I only tested my MWE version of this action. I'm doing a proper test on my fork now.

ssiccha: The four latest runs on https://github.com/ssiccha/gap/actions should now trigger all cases of this new job.

ssiccha: Both notifications were sent to #status-action-tests.

ssiccha:
I've squashed all commits. Tests on my fork also triggered both a success message and a failure message

wilfwilson (approved):
I'm going to merge this so we can start using it 🙂 We can always revisit it in future PRs if it's not entirely to our tastes, or if it breaks.

wilfwilson:
Of course we'll never actually get any benefit from this because the CI will never fail... 😉

PR #4394, opened 2021-04-13
buildsys: improve cross compilation support

### LINKED ISSUE #4154: buildsys: propagate more configure flags to GMP, zlib, boehm-gc
GAP bundles some software (GMP, zlib, boehm-gc) for convenience of the user; if the software is not found on the system where GAP is compiled, we compile the bundled copies as required.

However, this breaks down when cross compiling, as we don't pass on e.g. the `--host` flag. (Also for `make install` support one might argue that `--prefix` should be passed on; however, I have some doubts whether it's a good idea to allow installing at all when using a bundled GMP... but that's a separate issue).

Anyway: we should change that and pass on more flags, in an intelligent way.

Some people might not suggest "why not use `AC_CONFIG_SUBDIRS`. Fair question; if somebody can get it to work reliably, be my guest. But when I tried this back when I wrote the current GAP build system, I run into all kinds of troubles and limitations and ultimately gave up on it. One issue was that it basically passes on *all* configure arguments, including e.g. `--with-gmp=PATH`, which are not understood by the `configure` scripts of the software we bundled; and in some cases arguments might clash and weird errors resulted. Alas, it has been quite some time since I tried it, who knows, perhaps I screwed it up in some way. OTOH, I did test many different scenarios, and it's easy to write code which uses `AC_CONFIG_SUBDIRS` and works in simple settings, but not in more complex ones (out-of-tree build with complex arguments and extra env variables)

We could experiment also with using [`AX_SUBDIRS_CONFIGURE`](https://www.gnu.org/software/autoconf-archive/ax_subdirs_configure.html).

But I am not yet convinced that either macros is a win in the end (in terms of effort required by us, and features it gives us). For now it seems conceptually much simpler to create an accept-list of arguments to pass on, instead of curating a block-list of arguments not to pass on, as `AX_SUBDIRS_CONFIGURE` offers. Both approaches have their pros and cons, of course. For now I'd keep it simple and see how far we get with that, but of course anybody working on this is free to experiment in all directions!

fingolfin:
When building bundled libraries, pass the C/C++ compiler and info about the host/target architecture to their build systems.

A simple precursor of this patch has already been in use for several months to produce the [cross compiled GAP binaries for Julia](https://github.com/JuliaBinaryWrappers/GAP_jll.jl/releases/tag/GAP-v400.1100.1%2B0), with [build recipe here](https://github.com/JuliaPackaging/Yggdrasil/tree/master/G/GAP).

But I went further now and tried to do it "properly", to also resolve #4154 and to simply have more uniform handling of the external libraries.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -552,6 +552,13 @@ install-libgap: libgap.la
 # master target for external dependencies / subprojects
 EXTERN_FILES =
 
+# pass on some settings to external dependencies / subprojects which use
+# autoconf; this is helpful for cross compilation
+EXTERN_CONFIGFLAGS := \
+	CC="$(CC)" \
+	CXX="$(CXX)" \
+	--build="$(build)" \
+	--host="$(host)"
 
 #
 # GMP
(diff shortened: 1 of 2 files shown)


ssiccha (approved):
LGTM

PR #4397, opened 2021-04-14
Import CopyToVectorRep(NC) from HPC-GAP

### LINKED ISSUE #1643: GAP library declares `CopyToVectorRep` and `CopyToVectorRepNC`, but does not install functions
As part of the cleanup and merge of GAP and HPC-GAP @fingolfin suggests to first fix the API of `CopyToVectorRep`, and document it, before it becomes something we have to support.

This issue is mainly here so we do not forget addressing it.

fingolfin:
... as a stop-gap measure so we can make progress on the MatrixObj API without compromising HPC-GAP.

Resolves #1643

Was/is part of #4366 but I figured it would simplify review and merging of that PR to get this in a separate PR, while I sort out #4366


wilfwilson (approved):
Sure.

PR #4401, opened 2021-04-14
Remove MakeVector

fingolfin:
It is IMHO redundant with Vector.


ThomasBreuer (approved):
The comment on `MakeVector` in the list of open items already suggested that the function is not needed.
Since the function was undocumented and apparently also not used, I see no problem with removing it.

PR #4402, opened 2021-04-15
Reduce HPC-GAP diffs

fingolfin:
Some unwanted diffs between files in `hpcgap/lib/` and their counterparts in `lib/` have crept in once again -- remove them.


wilfwilson (approved):
Fine

PR #4404, opened 2021-04-15
Improve GET_FROM_SORTED_CACHE, use it to further the HPC-GAP unification; fix issue in ApproximateRoot

fingolfin:
This fixes a limitation in `GET_FROM_SORTED_CACHE`: the "maker" function now can involve recursion, which is necessary for being able to use it in `AllIrreducibleMonicPolynomialCoeffsOfDegree`. It then changes a few more places to use `GET_FROM_SORTED_CACHE`, and uses that to get rid of three more files in `hpcgap/lib/`.

@hulpke I think `ApproximateRoot` is your code... The last commit in this PR changes the caching it uses: before this PR, the cache in `ApproximateRoot` ignored the number of digits the user specified, so one got this:
```
gap> ApproximateRoot(2,2);
779770732423810/551381172667677
gap> ApproximateRoot(2,2,20);
779770732423810/551381172667677
gap> ApproximateRoot(2,2);
779770732423810/551381172667677
```

Now we get this:
```
gap> ApproximateRoot(2,2);
779770732423810/551381172667677
gap> ApproximateRoot(2,2,20);
61816429438630776653272737/43710816444795547405901687
gap> ApproximateRoot(2,2);
779770732423810/551381172667677
```
With my change, it caches only the result for digits=10 (the default). I am open to suggestions for other strategies. As it is, it seems that code in GAP calling `ApproximateRoot` never uses the third argument "digits", so hopefully it doesn't matter too much...?


hulpke (approved):
Thank you for the alert -- the change is fine.

PR #4407, opened 2021-04-16
Fix crashes when accessing `mat[i,j]` and `mat[i]` is unbound

fingolfin:
For a plist-of-plists matrix `mat`, accessing `mat[i,j]` crashed if `mat[i]`
was unbound and i in [1..Length(mat)].

Also, if `mat[i,j]` is not bound, then it now raises an error instead of
returning nothing.


ssiccha (approved):
LGTM

wilfwilson (approved):
Ooh, good catch.

PR #4410, opened 2021-04-16
Components of Wreath Product Elements

FriedrichRober:
# Description

## Notation
I am working on a package that aims to improve the performance of certain operations in wreath products.
Let $W = K \wr H$ be a wreath product and $H \leq \Sym(m)$. (I am not really interested in top actions that are not faithful or actions on an infinite set.)
We identify elements $w = (f, h) \in W$ by a tuple $(f_1, ..., f_m; h)$ where $f_i \in K$ is the $i$-th base component and $h \in H$ is the top component of $w$.

## My Package and Computational Advantages of this Tuple Notation
This generic representation of an element can be exploited to solve the following problems (hopefully efficiently) by using a *wreath cycle decomposition* and breaking the problems down to the groups $K$ and $H$:

- Given $w, v \in W$, determine if $w$ and $v$ are conjugate in $W$ and if so, compute an element $a \in W$ such that $w^a = v$.
- Given $w \in W$, enumerate the conjugacy class $w^W$ and determine its size.
- Given $W$, enumerate representatives for all conjugacy classes of $W$ and determine the number of conjugacy classes of $W$.
- Given $w \in W$, enumerate (generators of) $C_W(w)$ and determine its size.
In my package, I aim to provide implementations to solve these problems and hope that they will perform quite well in comparison with generic methods that do not exploit the wreath product structure.

[Package WreathProductElements](https://github.com/FriedrichRober/WreathProductElements)

The theory behind this will be described in a paper that is work in progress, by Alice C. Niemeyer, Sebastian Krammer, @DominikBernhardt, @FriedrichRober and @wucas.

## Interface for Wreath Product Elements in *GAP*
For my package I need an interface that allows me to work with elements of a wreath product in this tuple notation efficiently.
Note that it is easy to access $F = (f, 1)$ and $h$ via the current interface of a wreath product, namely by `h := w ^ Projection(W)` and `F := w * (h ^ (-1)) ^ Embedding(W, m + 1)`, but there is no convenient way known to me to access the components of $f$.

@hulpke suggested that this interface should be implemented on the `GAP`-side, which has the advantage that we could use the (undocumented) internal data stored in wreath products for the implementation. There are several ways how to realise such an interface, I include an implementation of such an interface in this pull request:

In analogy to `PermList` and `ListPerm`, I add two methods that translate between elements of wreath products in various representations to a list/tuple representation:
- `WreathProductElementList`
- `ListWreathProductElement`

In a follow-up pull request I suggest to document the record returned by the function `WreathProductInfo` partially and thus to provide some of the stored data that is shared by all wreath product representations to users and package authors.

I am very interested in other suggestions how such an interface can be realized and improved.

## Further details

I fixed `Projection` for perm wreath products in product action,
but there are still other issues with `Projection`, which I have to investigate further.
For example `Projection` is not implemented yet for matrix wreath products.
I also need to add tests, where the top domain or the base domain gets `compressed/renamed` by calling `WreathProduct`, for example I observe the following behavior which is expected to result in issues with my implementation:
```
gap> G := WreathProduct(SymmetricGroup(3), SymmetricGroup([11,12,13,14,15]));;
gap> Image(Projection(G));
Group([ (), (), (), (), (), (), (), (), (), (), (1,2,3,4,5), (1,2) ])
gap> Source(Embedding(G, 6));
Sym( [ 11 .. 15 ] )
```
There are also issues where generators of wreath products are inconvenient, for example we have in the above example:
```
gap> GeneratorsOfGroup(G);
[ (1,2,3), (1,2), (4,5,6), (4,5), (7,8,9), (7,8), (10,11,12), (10,11), (13,14,15), (13,14),
  (1,4,7,10,13)(2,5,8,11,14)(3,6,9,12,15), (1,4)(2,5)(3,6) ]
```
But it is only necessary to store base generators for each orbit under the top action, so in this case
`[ (1,2,3), (1,2), (1,4,7,10,13)(2,5,8,11,14)(3,6,9,12,15), (1,4)(2,5)(3,6) ]` is a more convenient generating set of `G`.
I will try to solve these issues (maybe in a separate pull request).

- [ ] proper formatting

- [ ] runnable tests

### DIFF
--- a/lib/gprd.gd
+++ b/lib/gprd.gd
@@ -566,3 +566,66 @@ InstallTrueMethod(IsGeneratorsOfMagmaWithInverses,
 
 DeclareRepresentation("IsWreathProductElementDefaultRep",
   IsWreathProductElement and IsPositionalObjectRep,[]);
+
+#############################################################################
+##
+#F  ListWreathProductElement
+#O  ListWreathProductElementNC
+##
+##  <#GAPDoc Label="ListWreathProductElement">
+##  <ManSection>
+##  <Func Name="ListWreathProductElement" Arg='G, x[, testDecomposition]'/>
+##  <Oper Name="ListWreathProductElementNC" Arg='G, x, testDecomposition'/>
+##
+##  <Description>
+##  Let <A>x</A> be an element of a wreath product <A>G</A>
+##  where <M>G = K \wr H</M> and <M>H</M> acts
+##  as a finite permutation group of degree <M>m</M>.
+##  We can identify the element <A>x</A> with a tuple <M>(f_1, \ldots, f_m; h)</M>,
+##  where <M>f_i \in K</M> is the <M>i</M>-th base component of <A>x</A>
+##  and <M>h \in H</M> is the top component of <A>x</A>.
+##  <P/>
+##  <Ref Func="ListWreathProductElement"/> returns a list <M>[f_1, \ldots, f_m, h]</M>
+##  containing the components of <A>x</A> or <K>fail</K> if <A>x</A> cannot be decomposed in the wreath product.
+##  <P/>
+##  If ommited, the argument <A>testDecomposition</A> defaults to true.
+##  If <A>testDecomposition</A> is true, <Ref Func="ListWreathProductElement"/> makes additional tests to ensure
+##  that the computed decomposition of <A>x</A> is correct,
+##  i.e. it checks that <A>x</A> is an element of the parent wreath product of <A>G</A>:
+##  <P/>
+##  If <M>K \leq \mathop{Sym}(l)</M>, this ensures that <M>x \in \mathop{Sym}(l) \wr \mathop{Sym}(m)</M>
+##  where the parent wreath product is considered in the same action as <A>G</A>,
+##  i.e. either in imprimitive action or product action.
+##  <P/>
+##  If <M>K \leq \mathop{GL}(n,q)</M>, this ensures that <M>x \in \mathop{GL}(n,q) \wr \mathop{Sym}(m)</M>.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction( "ListWreathProductElement" );
+DeclareOperation( "ListWreathProductElementNC", [HasWreathProductInfo, IsObject, IsBool] );
+
+#############################################################################
+##
+#F  WreathProductElementList
+#O  WreathProductElementListNC
+##
+##  <#GAPDoc Label="WreathProductElementList">
+##  <ManSection>
+##  <Func Name="WreathProductElementList" Arg='G, list'/>
+##  <Oper Name="WreathProductElementListNC" Arg='G, list'/>
+##
+##  <Description>
+##  Let <A>list</A> be equal to <M>[f_1, \ldots, f_m, h]</M> and <A>G</A> be a wreath product
+##  where <M>G = K \wr H</M>, <M>H</M> acts
+##  as a finite permutation group of degree <M>m</M>,
+##  <M>f_i \in K</M> and <M>h \in H</M>.
+##  <P/>
+##  <Ref Func="WreathProductElementList"/> returns the element <M>x \in G</M>
+##  identified by the tuple <M>(f_1, \ldots, f_m; h)</M>.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareGlobalFunction( "WreathProductElementList" );
+DeclareOperation( "WreathProductElementListNC", [HasWreathProductInfo, IsList] );
--- a/lib/gprd.gi
+++ b/lib/gprd.gi
@@ -898,6 +898,73 @@ end);
 
 #############################################################################
 ##
+#M  ListWreathProductElement(<G>, <x>[, <testMembership>])
+##
+InstallGlobalFunction( ListWreathProductElement,
+function(G, x, testDecomposition...)
+  local info;
+  if Length(testDecomposition) = 0 then
+    testDecomposition := true;
+  elif Length(testDecomposition) = 1 then
+    testDecomposition := testDecomposition[1];
+  elif Length(testDecomposition) > 1 then
+    ErrorNoReturn("too many arguments");
+  fi;
+  if not HasWreathProductInfo(G) then
+    ErrorNoReturn("usage: <G> must be a wreath product");
+  fi;
+  return ListWreathProductElementNC(G, x, testDecomposition);
+end);
+
+InstallMethod( ListWreathProductElementNC, "generic wreath product", true,
+ [ HasWreathProductInfo, IsWreathProductElement, IsBool ], 0,
+function(G, x, testDecomposition)
+  local info, list, i;
+  info := WreathProductInfo(G);
+  list := EmptyPlist(info!.degI + 1);
+  for i in [1 .. info!.degI + 1] do
+    list[i] := StructuralCopy(x![i]);
+  od;
+  return list;
+end);
+
+#############################################################################
+##
+#M  WreathProductElementList(<G>, <list>)
+##
+InstallGlobalFunction( WreathProductElementList,
+function(G, list)
+  local info, i;
+
+  if not HasWreathProductInfo(G) then
+    ErrorNoReturn("usage: <G> must be a wreath product");
+  fi;
+  info := WreathProductInfo(G);
+  if Length(list) <> info.degI + 1 then
+    ErrorNoReturn("usage: <list> must have ",
+                  "length 1 + <WreathProductInfo(G).degI>");
+  fi;
+  for i in [1 .. info.degI] do
+    if not list[i] in info.groups[1] then
+      ErrorNoReturn("usage: <list{[1 .. Length(list) - 1]}> must contain ",
+                    "elements of <WreathProductInfo(G).groups[1]>");
+    fi;
+  od;
+  if not list[info.degI + 1] in info.groups[2] then
+    ErrorNoReturn("usage: <list[Length(list)]> must be ",
+                  "an element of <WreathProductInfo(G).groups[2]>");
+  fi;
+  return WreathProductElementListNC(G, list);
+end);
+
+InstallMethod( WreathProductElementListNC, "generic wreath product", true,
+ [ HasWreathProductInfo, IsList ], 0,
+function(G, list)
+  return Objectify(FamilyObj(One(G))!.defaultType, StructuralCopy(list));
+end);
+
+#############################################################################
+##
 #M  PrintObj(<x>)
 ##
 InstallMethod(PrintObj,"wreath elements",true,[IsWreathProductElement],0,
--- a/lib/gprdmat.gi
+++ b/lib/gprdmat.gi
@@ -440,6 +440,65 @@ local  info, degI, dimA, zero, projFunc;
   return info.projection;
 end);
 
+#############################################################################
+##
+#M  ListWreathProductElementNC( <G>, <x> )
+##
+InstallMethod( ListWreathProductElementNC, "matrix wreath product", true,
+ [ IsMatrixGroup and HasWreathProductInfo, IsObject, IsBool], 0,
+function(G, x, testDecomposition)
+  local info, degI, dimA, h, list, i, j, k, zeroMat;
+
+  info := WreathProductInfo(G);
+  degI := info.degI;
+  dimA := info.dimA;
+
+  # The top group element
+  h := x ^ Projection(G);
+  if h = fail then
+    return fail;
+  fi;
+  list := EmptyPlist(degI + 1);
+  list[degI + 1] := h;
+  if testDecomposition then
+    # ZeroMatrix does not accept IsPlistRep
+    if IsPlistRep(x) then
+      zeroMat := NullMat(dimA, dimA, info.field);
+    else
+      zeroMat := ZeroMatrix(dimA, dimA, x);
+    fi;
+  fi;
+  for i in [1 .. degI] do
+      j := i ^ h;
+      list[i] := ExtractSubMatrix(x, [dimA * (i - 1) + 1 .. dimA * i], [dimA * (j - 1) + 1 .. dimA * j]);
+      if testDecomposition then
+        for k in [1 .. degI] do
+          if k = j then
+            continue;
+          fi;
+          if ExtractSubMatrix(x, [dimA * (i - 1) + 1 .. dimA * i], [dimA * (k - 1) + 1 .. dimA * k]) <> zeroMat then
+            return fail;
+          fi;
+        od;
+      fi;
+  od;
+  return list;
+end);
+
+#############################################################################
+##
+#M  WreathProductElementListNC(<G>, <list>)
+##
+InstallMethod( WreathProductElementListNC, "matrix wreath product", true,
+ [ IsMatrixGroup and HasWreathProductInfo, IsList ], 0,
+function(G, list)
+  local info;
+
+  info := WreathProductInfo(G);
+  # TODO: Remove `MatrixByBlockMatrix` when `BlockMatrix` supports the MatObj interface.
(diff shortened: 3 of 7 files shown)


hulpke (approved):
Do you really want to have the element test `x in G` in `ListWreathProductElement` ? This requires dividing off a stabilizer chain each time. I'd expect that the test would actually be cheaper to do once one obtained the decomposition as tuple.

FriedrichRober:
@hulpke You are right. So probably I will redo this safety check in `ListWreathProductElement` the following way:
```
First check input, but without the element test x in G
list := ListWreathProductElementOp(G, x);
Then check if list encodes an element of G
return list;
```
I still think that it might be useful if the method `ListWreathProductElement` checks if `x` is an element of `G` in some way.  If the user is really sure that he doesn't need any safety check, he can still use `ListWreathProductElementOp` that doesn't do any checks on the input.

hulpke:
OK. Then you might want to rename the `Op` operations into NC (no check)

ssiccha:
The idea sounds good! I haven't looked at your code though.

> I fixed Projection for perm wreath products in product action,
but there are still other issues with Projection, which I have to investigate further.
For example Projection is not implemented yet for matrix wreath products.

Unless you need the `Projection` function I'd suggest to just leave that as it is. Maybe instead open issues about what you found?

lib/gprd.gi
```diff
@@ -896,6 +896,68 @@ local I,n,fam,typ,gens,hgens,id,i,e,info,W,p,dom;
...
+function(G, list)
+  local info, i;
+
+  if not HasWreathProductInfo(G) then
+    ErrorNoReturn("usage: <G> must be a wreath product");
+  fi;
+  info := WreathProductInfo(G);
+  if Length(list) <> info.degI + 1 then
+    ErrorNoReturn("usage: <list> must have ",
+                  "length <WreathProductInfo(G).degI>");
```
ssiccha, line 940:
```suggestion
                  "length 1 + <WreathProductInfo(G).degI>");
```

lib/gprd.gi
```diff
@@ -896,6 +896,68 @@ local I,n,fam,typ,gens,hgens,id,i,e,info,W,p,dom;
...
+  fi;
+  return ListWreathProductElementOp(G, x);
+end);
+
+InstallMethod( ListWreathProductElementOp, "generic wreath product", true,
+ [ HasWreathProductInfo, IsWreathProductElement ], 0,
+function(G, x)
+  local info, list, i;
+  info := WreathProductInfo(G);
+  list := ListWithIdenticalEntries(info!.degI + 1, 0);
```
ssiccha, line 919:
```suggestion
  list := EmptyPlist(info!.degI + 1);
```

lib/gprdmat.gi
```diff
@@ -416,6 +416,49 @@ local  info,proj,H;
...
+  info := WreathProductInfo(G);
+  degI := info.degI;
+  dimA := info.dimA;
+  field := info.field;
+
+  base := [];
+  topImages := [];
+  for k in [1 .. degI] do
+      for l in [1 .. degI] do
+          block := x{[dimA * (k - 1) + 1 .. dimA * k]}{[dimA * (l - 1) + 1 .. dimA * l]};
```
ssiccha, line 437:
```suggestion
          block := ExtractSubMatrix(x, [dimA * (k - 1) + 1 .. dimA * k], [dimA * (l - 1) + 1 .. dimA * l]);
```
Doing `x{rows}{cols}` first creates an intermediate object `x{rows}` and then selects the columns.

lib/gprdmat.gi
```diff
@@ -416,6 +416,49 @@ local  info,proj,H;
...
+          if block <> NullMat(dimA, dimA, field) then
+              Add(base, block);
+              Add(topImages, l);
+              break;
+          fi;
+      od;
+  od;
+  top := PermList(topImages);
+  return Concatenation(base, [top]);
```
ssiccha, line 446: You could also do (and should then rename `base` to maybe `result`)
```suggestion
  Add(base, PermList(topImages));
  return base;
```

lib/gprdperm.gi
```diff
@@ -813,17 +813,80 @@ local  info,proj,H;
...
+ [ IsPermGroup and HasWreathProductInfo, IsObject ], 0,
+function(G, x)
+  local info, list, h, f, degK, i, constPoints, imageComponents, comp;
+
+  info := WreathProductInfo(G);
+  # Imprimitive Action, tuple (i, j) corresponds
+  # to point i + degK * (j - 1)
+  if IsBound(info.permimpr) and info.permimpr then
+    h := x ^ Projection(G);
+    f := x * Image(Embedding(G, info.degI + 1), h) ^ (-1);
```
ssiccha, line 842:
```suggestion
    # The top group element
    h := x ^ Projection(G);
    # The product of the base group elements
    f := x * Image(Embedding(G, info.degI + 1), h) ^ (-1);
```

lib/gprdperm.gi
```diff
@@ -813,17 +813,80 @@ local  info,proj,H;
...
+    list := ListWithIdenticalEntries(info!.degI + 1, 0);
+    for i in [1 .. info.degI] do
+        list[i] := RestrictedPermNC(f, info.components[i]) ^ info.perms[i];
+    od;
+    list[info.degI + 1] := h;
+  # Primitive Action, tuple (t_1, ..., t_degI) corresponds
+  # to point Sum_{i=1}^degI t_i * degK ^ (i - 1)
+  elif IsBound(info.productType) and info.productType then
+    h := x ^ Projection(G);
+    f := x * Image(Embedding(G, info.degI + 1), h) ^ (-1);
```
ssiccha, line 852:
```suggestion
    # The top group element
    h := x ^ Projection(G);
    # The product of the base group elements
    f := x * Image(Embedding(G, info.degI + 1), h) ^ (-1);
```

lib/gprdperm.gi
```diff
@@ -813,17 +813,80 @@ local  info,proj,H;
...
+    degK := NrMovedPoints(info.groups[1]);
+    # constPoints correspond to [1, 1, 1, ...], [2, 2, 2, ...], ...
+    constPoints := List([0 .. degK - 1], x -> Sum([0 .. info.degI - 1],
+                                                  y -> x * degK ^ y)) + 1;
+    # imageComponents = [ [1^f_1, 1^f_2, 1^f_3, ...],
+    #                     [2^f_1, 2^f_2, 2^f_3, ...], ... ]
+    imageComponents := List(OnTuples(constPoints, f) - 1,
+                            p -> CoefficientsQadic(p, degK) + 1);
```
ssiccha, line 860: Neat trick! :)

lib/gprdperm.gi
```diff
@@ -813,17 +813,80 @@ local  info,proj,H;
...
+    for i in [1 .. degK] do
+      comp := imageComponents[i];
+      Append(comp, ListWithIdenticalEntries(info.degI - Length(comp), 1));
```
ssiccha, line 863: What do you do this for? A comment would be nice.

lib/gprdperm.gi
```diff
@@ -813,17 +813,80 @@ local  info,proj,H;
...
+    od;
+    list := ListWithIdenticalEntries(info!.degI + 1, 0);
```
ssiccha, line 865:
```suggestion
    list := EmptyPlist(info!.degI + 1);
```

ssiccha (changes requested):
I've got a few minor remarks. I've marked a few lines where it's not clear to me what they do. Could you add explanations there?

lib/gprdperm.gi
FriedrichRober: Fixed Points are Zeroes in the Qadic expansion. I need to append "Trailing" Zeros.

FriedrichRober: For example if '(1, ..., 1) ^ (f_1, ..., f_m) = (1, ..., 1)`, before I append anything, I have `CoefficientsQadic(0, degK) + 1 = []`

ssiccha: That makes sense. Could you add that as a comment into the source code?

lib/gprdmat.gi
FriedrichRober: Thanks for showing me this improvement.

lib/gprdperm.gi
FriedrichRober: Yes, I add this as a comment

FriedrichRober:
Ok, I implemented all changes @ssiccha suggested and I am nearly finished with the membership test idea proposed by @hulpke. Only the `ProductAction` case is missing, where the Projection is not optimized yet and throws an error if`x` is not an element of the wreath product `G`.

FriedrichRober:
So, in my opinion the product action is constructed unintuitively. It seems to me, that the action of the top elements is renamed in reverse order via the isomorphism `Sym([1, ..., m]) -> Sym((m, ..., 1])`. The base elements act as expected. The action points are as expected translated via a qadic expansion.
```
gap> K := SymmetricGroup(3);;
gap> H := SymmetricGroup(5);;
gap> G := WreathProductProductAction(K, H);;
gap> x := (1,2) ^ Embedding(G, 1);; # corresponds to ( (1,2), (), (), (), (); () )
gap> y := (1,2) ^ Embedding(G, 6);; # corresponds to ( (), (), (), (), (); (1,2) )
gap> a := 3 ^ 0 + 1;; # corresponds to [2, 1, 1, 1, 1]
gap> b := 3 ^ 4 + 1;; # corresponds to [1, 1, 1, 1, 2]
# I expect that x acts on the first component via (1,2),
# i.e. [ p[1], p[2], ..., p[5] ] ^ x = [ p[1] ^ (1,2), p[2], ..., p[5] ]
gap> CoefficientsQadic(a ^ x - 1, degK) + 1; # corresponds to a ^ x = [1, 1, 1, 1, 1] as expected
[]
gap> CoefficientsQadic(b ^ x - 1, degK) + 1; # corresponds to b ^ x = [2, 1, 1, 1, 2], as expected
[ 2, 1, 1, 1, 2 ]
# I expect that y acts by permutation of the components,
# i.e. [ p[1], ..., p[5] ] ^ y = [ p[1 ^ (h ^ -1)], ..., p[5 ^ (h ^ -1)] ],
# but in GAP it acts on the components in reverse order,
# i.e. [ p[5], ..., p[1] ] ^ y = [ p[5 ^ (h ^ -1)], ..., p[1 ^ (h ^ -1)] ]
gap> CoefficientsQadic(a ^ y - 1, degK) + 1; # corresponds to a ^ y = [2, 1, 1, 1, 1], but expected [1, 2, 1, 1, 1] 
[ 2 ]
gap> CoefficientsQadic(b ^ y - 1, degK) + 1; # corresponds to a ^ y = [1, 1, 1, 2, 1], but expected [1, 1, 1, 1, 2] 
[ 1, 1, 1, 2 ]
```

lib/gprdperm.gi
fingolfin: I still find this confusing: you talk about missing zeros and that you need to append them, but then you append ones.

tst/testinstall/opers/ListWreathProductElement.tst
```diff
@@ -0,0 +1,141 @@
...
+gap> list := [K.1, K.2, One(K), K.2, One(K), (1,2)(3,4)];;
+gap> x := WreathProductElementList(G, list);;
+gap> list = ListWreathProductElement(G, x);
+true
+
+#
+gap> x := G.1 * G.5 * G.2 * G.5 ^ ((G.4 * G.5) ^ 2) * G.2;;
+gap> list := ListWreathProductElement(G, x);;
+gap> x = WreathProductElementList(G, list);
+true
```
fingolfin, line 141:
```suggestion
true

```

lib/gprd.gd
```diff
@@ -566,3 +566,63 @@ InstallTrueMethod(IsGeneratorsOfMagmaWithInverses,
...
   IsWreathProductElement and IsPositionalObjectRep,[]);
+
+#############################################################################
+##
+#F  ListWreathProductElement
+#O  ListWreathProductElementNC
+##
+##  <#GAPDoc Label="ListWreathProductElement">
+##  <ManSection>
+##  <Func Name="ListWreathProductElement" Arg='G, x, [testMembership]'/>
```
fingolfin, line 577:
```suggestion
##  <Func Name="ListWreathProductElement" Arg='G, x[, testMembership]'/>
```

fingolfin (commented):
Nice!

lib/gprdmat.gi
```diff
@@ -406,14 +406,88 @@ end );
...
+  for i in [1 .. degI] do
+      j := i ^ h;
+      list[i] := ExtractSubMatrix(x, [dimA * (i - 1) + 1 .. dimA * i], [dimA * (j - 1) + 1 .. dimA * j]);
+      if testDecomposition then
+        for k in [1 .. degI] do
+          if k = j then
+            continue;
+          fi;
+          if ExtractSubMatrix(x, [dimA * (i - 1) + 1 .. dimA * i], [dimA * (k - 1) + 1 .. dimA * k]) <> zeroMat then
+            return fail;
```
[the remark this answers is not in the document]
FriedrichRober: Yes, I will add some more tests

lib/gprdperm.gi
FriedrichRober: You are right. I will adjust the comment again. For example the point `1` is first translated to the qadic expansion of `1 - 1 = 0 = [0, 0, ...]_degK`, then we add `1` to each integer in the expansion and get `[1, 1, ...]` as our tuple. So the problem is indeed about missing zeroes in the qadic expansion, but I append `1` because `imageComponents` already contains the translated tuples.

FriedrichRober:
@fingolfin, I have added some new tests and applied all of your suggestions. I left the comments open that may require further discussion.

Do you know what to do about the failed tests in the environments, where `ExtractSubMatrix` is not defined?

lib/gprdperm.gi
FriedrichRober: I updated the comment. I hope that it is now less confusing.

lib/gprdmat.gi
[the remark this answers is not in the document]
FriedrichRober: I added some tests in order to catch some more cases where `fail` gets returned.

fingolfin:
@FriedrichRober sorry, didn't see your question before... Do the tests pass locally for you? Do they still pass after rebasing your PR on the latest master? Do they pass if GAP is loaded without packages (`gap -A`) ?

lib/gprdmat.gi
```diff
@@ -406,14 +406,91 @@ end );
...
+
+  # The top group element
+  h := x ^ Projection(G);
+  if h = fail then
+    return fail;
+  fi;
+  list := EmptyPlist(degI + 1);
+  list[degI + 1] := h;
+  if testDecomposition then
+    zeroMat := NullMat(dimA, dimA, info.field);
```
fingolfin, line 464: If your goal is to let `x` be any MatrixObj, then you should use something like `ZeroMatrix(dimA, dimA, x)` here (perhaps it needs a special case for `IsPlistRep`, see e.g. [here](https://github.com/gap-packages/forms/pull/4/files#diff-ad4960ae227bb883de1bf424204b3fb9d2a4b63a9d795a278b85600e211e87d4R21-R27).

FriedrichRober:
@fingolfin, the tests do not run without packages (`gap -A`). The reason is that `ExtractSubMatrix` does not have any implementation to deal with `IsPlistRep` if no packages are loaded.

FriedrichRober:
@fingolfin I think the pull request is now ready.

I changed the option `testMembership` into `testDecomposition`, so it no longer tests for membership in `G` but for membership in the parent wreath product of `G`, i.e. it checks that the decomposition of the element `x` in the wreath product `G` is correct.

I would squash the commits into three commits once you are ready with the review:
commit 1: Add `ExtractSubMatrix` for `IsPlistRep`
commit 2: Adjust `Projection` methods for wreath products
commit 3: Add `ListWreathProductElement` and `WreathProductElementList`

fingolfin (approved):

ssiccha (approved):
My comments were adressed. I haven't been able to read the newest version, though.

PR #4413, opened 2021-04-17
remove an irritating `if` statement in `MethodsOperation`

ThomasBreuer:
The component `location` in the records returned by `MethodsOperation` is always bound.
Note that the check before the definition of `MethodsOperation` guarantees that the list returned by `METHODS_OPERATION` has the right format.

### DIFF
--- a/lib/methsel.g
+++ b/lib/methsel.g
@@ -31,7 +31,6 @@ local type,fam,methods,i,j,flag,erg;
   fam:=FamilyObj(obj);
   methods:=METHODS_OPERATION(attr,1);
   for i in [1..LEN_LIST(methods)/(1+BASE_SIZE_METHODS_OPER_ENTRY)] do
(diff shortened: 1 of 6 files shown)


fingolfin (approved):
OK. This is from a time when `BASE_SIZE_METHODS_OPER_ENTRY` was in flux. You could also remove the checks for its value in oper1.g inside `INSTALL_METHOD_FLAGS`

ThomasBreuer:
Thanks for the hint, @fingolfin. I have now adjusted also `methsel.g` and `oper1.g`.

fingolfin:
I've adjusted your second commit to work with HPC-GAP. One needs to run `make docomp` twice, once in a regular GAP build, and once in a HPC-GAP build; the latter is necessary as the difference in `oper1.g` are controlled by which GAP is running, so we have no choice (with our current tools) than to do it this way.

fingolfin (approved):

PR #4415, opened 2021-04-19
`Random` for non-dense lists and empty intervals

### LINKED ISSUE #4379: Random for non-dense lists
In GAP 4.11 and in the master branch, currently we have the following.
```
gap> Random( [ 1,,,,,, 2 ] );
  Random( GlobalMersenneTwister, 1, Length( l ) 
 ) at /.../gap/lib/random.gi:240 called from 
<function "Random for an internal list">( <arguments> )
 called from read-eval loop at *stdin*:33
```

ThomasBreuer:
- Admit `Random` for non-dense lists:
  The documentation has always allowed this,
  the IO package has a correct method for non-dense lists (thus we cannot simply require `IsDenseList`),
  just the methods in the GAP library did not catch this case.

- Improved `Random( RandomSource( IsMersenneTwister ), 1, 0 )`:
  This had returned `1`, now it returns `fail`.
  Note that the documentation does not state what happens
  if the given interval is empty,
  thus I do not claim that the change is a bugfix.

(resolves #4379)

### DIFF
--- a/lib/coll.gi
+++ b/lib/coll.gi
@@ -254,7 +254,7 @@ InstallMethod( RepresentativeSmallest,
 # RandomList is not an operation to avoid the (often expensive) cost of
 # dispatch for lists
 InstallGlobalFunction( RandomList, function(args...)
-  local len, source, list;
+  local len, source, list, nr;
   len := Length(args);
   if len = 1 then
     source := GlobalMersenneTwister;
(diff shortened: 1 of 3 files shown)


ChrisJefferson (approved):

ThomasBreuer:
In a discussion with @fingolfin, we agreed that it is better to proceed in a different way, see #4438.
Sorry for the confusion.

PR #4419, opened 2021-04-20
Slack Notifications: send current status, not previous

wilfwilson:
Although the Action was sending the notification at (pretty much) the right time, it was not always (not ever?) sending the correct message. This also required that I make the following PR https://github.com/gap-actions/should-i-notify-action/pull/1 which is now merged.

I have spent today testing this on the repository here, and I have satisfied myself that it now sends the right notification at the right time.


fingolfin (approved):
I trust Wilf that this is correct.

ssiccha (approved):
Yes, how did I miss that? :sweat_smile:

Issue #4417, opened 2021-04-20
changes in `lib/oper1.g`


ThomasBreuer:
Yesterday I tried to make a harmless little change in `lib/oper1.g`.
(Of course, experienced developers know that changes in this file are *never* harmless.)

When I had created the pull request, the CI tests reported failures.

- The first one showed that also the file `src/c_oper1.c` must be updated (which can be done by calling `make docomp`). The failure happened in a strange place, a call to `bin/BuildPackages.sh`. Namely, the function `run_configure_and_make` tries to determine the name of a package by calling GAP, reading the `PackageInfo.g` file, extracting the name, and printing it. However, the first printed value was a message of the form `#W Static module lib/oper1.g has CRC mismatch, ignoring`, which is currently not a valid package name.
  Since GAP works nicely without `src/c_oper1.c`, I think the installation of packages should not fail because this file is not up to date. Perhaps GAP could simply be started with the `-M` option, in order to omit the warning message?

- The next question is why the generated file `src/c_oper1.c` must be under version control at all. It can be created when GAP gets installed. This would save some developers' work.

- After updating `src/c_oper1.c`, the HPCGAP related CI tests complained, since also the file `src/hpc/c_oper1.c` must be updated. For this update, one has to run `make docomp` for an installed HPCGAP.
  Again, the natural situation for creating `src/hpc/c_oper1.c` is installation time (this time the installation of HPCGAP), and the file would not need to be under version control.
  (Is it really expected from a GAP developer to install HPCGAP, just because of a harmless little change in `lib/oper1.g`?)

- The next failure was reported from the following test in `tst/testinstall/varargs.tst`.

      gap> Display(InstallMethod);
      function ( arg... )
          <<compiled GAP code>> from GAPROOT/lib/oper1.g:338
      end

  The aim of this test is to show an example of the `<<compiled GAP code>>` case, but the test actually checks that the code of a particular function occurs in a particular line of `lib/oper1.g`.
  Thus we are forced to adjust this testfile because of changes in `lib/oper1.g`. Is this acceptable?

  We could extend the code of `Test` such that the two outputs get modified according to some rules before they get compared, in order to admit predefined differences in tests.
  A rudimentary version of such an approach is taken in the tests for the JuliaInterface package --there are other instances where this problem arises.
  I could create a pull request for this functionality if others regard it as reasonable.

fingolfin:
[...]

Issue #4368, opened 2021-04-07
sporadic test failure in tst/testbugfix/2006-08-28-t00151.tst


ThomasBreuer:
The test from `tst/testbugfix/2006-08-28-t00151.tst:14` has recently failed in several situations.
This is mentioned in the discussion of pull request #4326, which contains links to other instances.
@wilfwilson stated: "Let's keep an eye out for this happening again."
Now I got the same failure again in [a test run](https://github.com/gap-system/gap/actions/runs/723811708).
I have still no idea what is going on here.

fingolfin:
One potential issue with this test is that it depends on the state of the garbage collector: a lot of collections are made in it, and at some point (or several) a GC is triggered; perhaps even a full (not partial) one. If it happens in the first loops vs. the second loop, that can substantially skew the timings. I'll open a PR for this.

Next, the CI tests run in VMs, and multiple VMs will run on a single real host; and so other workloads run in parallel can cause strong fluctuations in the performance of the VM. So it may happen that it slows down temporarily while one of the two loops being compared is run. This skew can be quite substantial -- looking at https://github.com/gap-system/gap/actions/workflows/CI.yml?query=branch%3Amaster I see some passing runs in 37 minutes and other in over an hour. 

To combat *that* one can employ a bit more sophisticated sampling: right now we run each of the two tests 20 times, and then compare the sums (so, in effect, the averages). It might be better to compare the medians; or to discard outliers and then compare the average; and one cold also compare the geometric mean instead, etc. etc. As it is, benchmarking is simply hard, and the approach used in that test is perhaps a bit too naive when faced with such an erratic environment as presented by the VMs our CI runs on...

wilfwilson:
The failures persist (although perhaps less frequently now?)
* https://github.com/gap-system/gap/runs/2394609749?check_suite_focus=true

ThomasBreuer:
Yes, I had mentioned this just an hour ago in a review comment for pull request #4421. (Why doesn't github automatically show a link here?)

fingolfin:
@ThomasBreuer that's indeed weird, normally GitHub shows "mentions". In any case, I think it's good that Wilf added the explicit links to failed build jobs here, as it makes it much easier to see and review them when one visits this issue.

[...]

PR #4421, opened 2021-04-20
CI: detect failure to run `make docomp` nicely again

fingolfin:
We have a test for our CI which detects if a PR should have run `make docomp`.
But we did not reach it anymore since some changes to `bin/BuildPackages.sh`
which added code to use GAP to extract the name of any given packages, by
letting GAP print it to stdout and grabbing it from there. But in this
particular failure scenario, GAP also printed a warning, which caused
`BuildPackages.sh` to fail. This patch suppresses that warning.

To verify this works, I am adding a commit which is supposed to trigger this failure mode.

UPDATE: test worked as desired, see e.g. https://github.com/gap-system/gap/runs/2394724252?check_suite_focus=true

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -194,7 +194,7 @@ run_configure_and_make() {
   then
     if grep Autoconf ./configure > /dev/null
     then
-      local PKG_NAME=$($GAP -q -T -A -r <<GAPInput
+      local PKG_NAME=$($GAP -q -T -A -r -M <<GAPInput
 Read("PackageInfo.g");
 Print(GAPInfo.PackageInfoCurrent.PackageName);
 GAPInput


ThomasBreuer (approved):
Thanks.
(I had suggested this change in issue #4417.)
The test failure is again due to `tst/testbugfix/2006-08-28-t00151.tst`, see issue #4368.
We can merge the pull request, but the test is apparently still not safe.

wilfwilson (approved):
Thanks for the demonstration that this works.

I think we should merge despite the unrelated CI job failure.

fingolfin:
Test: reference #4368 from a regular PR comment

fingolfin (commented):
Test: reference #4368 from a PR review comment

fingolfin:
Test: reference #4368 from a regular PR comment *again*

fingolfin (commented):
Test: reference #4368 from a PR review comment *again*

PR #4426, opened 2021-04-21
Remove an unstable test

### LINKED ISSUE #4368: sporadic test failure in tst/testbugfix/2006-08-28-t00151.tst

fingolfin:
... which sometimes failed, as it relied on time measurements that
are inherently unreliable.

The test is meant to verify that we do not accidentally perform
method selection as part of the implementation of `PositionSorted`.
But this is an unlikely regression; we even have a comment explaining
why both `PositionSorted` and `PositionSortedOp` exist. We should
perhaps document it even more explicitly. But we do not perform similar
tests for any other of the many operations in a similar situation,
and as mentioned, the test is unreliable.

So instead of trying to put lipstick onto a pig, let's just remove it.

Closes #4368


ThomasBreuer (approved):
O.k., this is also a solution.

alex-konovalov (approved):

PR #4427, opened 2021-04-21
Fix `QuaternionAlgebra` bug that caused `QuaternionAlgebra( [1], 2, 5 ).1` to error the second time it was invoked

### LINKED ISSUE #2940: HPC-GAP specific code for `QuaternionAlgebraData`
Just for curiosity:

I stumbled over HPCGAP specific code for dealing with the cache `QuaternionAlgebraData`,
see `lib/algsc.gi`.
The function `StoreQuaternionAlgebraData` adds the algebra and related data to this cache if necessary.
In the case that no new entry is needed, this function creates a new algebra,
which is equal (in the sense of `\=`) to the given one, and returns it.

Why is this new object necessary, and why would it be wrong to keep (in the function `QuaternionAlgebra`)
the given algebra instead of replacing it by the one returned by `StoreQuaternionAlgebraData`?

fingolfin:
Also ensure that the exact same code is used to produce the
resulting algebra regardless of whether it is created freshly
or from the cache. This revealed a bug: `A.1` only worked
in the former case, not the latter, which could lead to subtly
broken code. Before this patch, the following happened:

    gap> A:= QuaternionAlgebra( [1], 2, 5 );
    <algebra-with-one of dimension 4 over Rationals>
    gap> A.1;
    e
    gap> B:= QuaternionAlgebra( [1], 2, 5 );
    <algebra-with-one of dimension 4 over Rationals>
    gap> B.1;
    Error, illegal access to record component `obj.1'

This is now fixed.

Resolves #2940

### DIFF
--- a/lib/algsc.gi
+++ b/lib/algsc.gi
@@ -689,26 +689,9 @@ InstallAccessToGenerators( IsSCAlgebraObjCollection and IsFullSCAlgebra,
 ##
 #V  QuaternionAlgebraData
 ##
-InstallFlushableValue( QuaternionAlgebraData, [] );
+InstallFlushableValue( QuaternionAlgebraData, [ [], [] ] );
 if IsHPCGAP then
     ShareSpecialObj( QuaternionAlgebraData );
-
-    BindGlobal("StoreQuaternionAlgebraData", function( a, b, F, A )
-      local stored;
-      atomic readwrite QuaternionAlgebraData do
-        stored:= First( QuaternionAlgebraData,
-                        t -> t[1] = a and t[2] = b and
-                             IsIdenticalObj( t[3], FamilyObj( F ) ) );
-        if stored = fail then
-          # Store the data for the next call.
-          Add( QuaternionAlgebraData, MakeImmutable([ a, b, FamilyObj( F ), A ]) );
-        else
-          A:= AlgebraWithOne( F, GeneratorsOfAlgebra( stored[4] ), "basis" );
-          SetGeneratorsOfAlgebra( A, GeneratorsOfAlgebraWithOne( A ) );
-        fi;
-      od;
-      return A;
-    end);
 fi;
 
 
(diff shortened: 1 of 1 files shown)


lib/algsc.gi
```diff
@@ -780,15 +755,11 @@ InstallGlobalFunction( QuaternionAlgebra, function( arg )
...
 
-      # Store the data for the next call.
-      if IsHPCGAP then
-        A := StoreQuaternionAlgebraData( a, b, F, A );
-      else
-        Add( QuaternionAlgebraData, [ a, b, FamilyObj( F ), A ] );
-      fi;
-
-    fi;
+    A:= AlgebraWithOne( F, GeneratorsOfAlgebra( stored ), "basis" );
```
ThomasBreuer, line 761: Here a new algebra gets created; setting the filter `IsFullSCAlgebra` would be useful.

ThomasBreuer (changes requested):
Very nice, the caching technicalities are delegated to `GET_FROM_SORTED_CACHE`.
The test failures are due to a subtle change of the behaviour.
After the changes, each call of `QuaternionAlgebra` creates a new algebra, and the `\.` access to generators is not supported since the filter `IsFullSCAlgebra` is not set.
The situation was much worse with the old code: The *first* call for each coefficients family simply took the newly created algebra, in which the filter was set, whereas subsequent calls used the cache and then created a new algebra, without the filter.
Thus it depended on the GAP session whether `\.` access to generators worked.
I regard the changes as a bugfix, perhaps even an entry for the release notes would be appropriate.

lib/algsc.gi
fingolfin: Great catch, fixed now, and commit message adjusted. I agree this is a bugfix.

ThomasBreuer (approved):
Very nice.

PR #259, opened 2015-09-26
Merge First with FirstOp, Filtered with FilteredOp, etc.

fingolfin:
This pull request deals with `First`, `Number`, `ForAll`, `ForAny`, `Filtered` and their undocumented "Op" variants `FirstOp` and so on. To keep things simple, I'll just refer to `First` from now on.

Previously, `First` was a global function, which checked if its first argument had the TNUM of an internal list, and if yes, applied certain code to it; if not, it delegated to the `FirstOp` operation. This setup was introduced by Thomas Breuer in March 1998 (the commit also dealt with `Sum`, `Product`, and `List`, but I do not). No reason was given, but I assume it was to speed up things.

Fast forward today. For the datastructure package, we'd like to use `First` in a single-argument version, so I looked into turning `First` into an operation. This lead to me discovering the situation I just described. Curious, I tested what happend if I merged `First` and `FirstOp`. I wanted to know how much slower it really made things.

To my surprise, the change made no difference. Moreover, when I applied a similar change to `Number`, things actually got _faster_, at least in the micro benchmark I used! This lead to a lot more research and testing... I discovered the following:
- the method dispatch creates no measurable overhead. Or perhaps it is measurable with a far more elaborate setup, but for me it was basically drowned out in the noise generated by garbage collection and other outside factors
- the following two loops seemingly do the same thing, but in reality, they (a) do not and (b) perform quite differently, depending on the input:

```
# Loop 1
for x in C do
  ...
od;
# Loop 2
for i in [1..Length(C)] do
  x := C[i];
  ...
od;
```

But these are not at all equivalent. For starters, loop 2 only works correctly for dense lists, but let's ignore that for the moment. But in reality, the first loop actually corresponds more closely to this (as is explained in the manual entry for the `for` loop):

```
# Loop 3
i := 1;
while i <= Length(C) do
  x := C[i];
  i := i + 1;
  ...
od;
```

Now, for internal lists (such as plists, blists, ranges, strings), loop 1 is always the fastest. But for non-internal lists (such as 8-bit compressed vectors), the second loop is faster -- because it does not constantly recompute the list length. While doing that is fast, if you have millions of iterations, the overhead adds up. (The explicit loop 3 is always slowest, by the way).

[...]

PR #4428, opened 2021-04-21
docs: mention FirstOp, LastOp, ForAllOp, etc.

fingolfin:
... in the documentation entries for their non-Op counterparts First, Last,
ForAll, etc. and explain that custom methods need to be installed there.

Also explain uniformly that this is to avoid the method selection overhead
penalty for plain list inputs.

This may be helpful for some people, and it certainly would have saved me time back when I created the misguided PR #259 :-).

I am sure this could be phrased better; I am open for suggestions. I tried to write it in such a way that a regular users understand they can ignore it, yet can still have an idea what this is about. 

Finally, of course there are alternatives; e.g. we could add separate manual entries for FirstOp, ... etc.; but I thought it made sense to not do so, as they are in a sense an internal implementation detail, so it is OK to only mention them in passing (just enough so that people who stumble across them get a rough idea and developers have enough to know where to dig). But here, too, I am open to having my mind changed :-).

### DIFF
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -2051,6 +2051,10 @@ DeclareAttribute( "UnderlyingCollection", IsListOrCollection );
 ##  same as <Ref Oper="ShallowCopy"/> (see also 
 ##  <Ref Sect="Duplication of Lists"/>).
 ##  <P/>
+##  Developers who wish to adapt this for custom list or collection types need to
+##  install suitable methods for the operation <C>ListOp</C>.
+##  <Index Key="ListOp"><C>ListOp</C></Index>
+##  <P/>
 ##  <Example><![CDATA[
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):

fingolfin:
@ThomasBreuer I agree that such a section would be useful and also that there is more to the story; if someone wants to write one, or even just file an issue to suggest doing so, I'd welcome it. But as it is, I simply wanted to scratch this particular itch, also partially motivated by issue #4426 (where I removed a test that, as you know, was supposed to verify that `PositionSorted` is not affected by method dispatch overhead -- I figured we should make sure this requirement is carefully documented; then I noticed that it *is* documented there, just not everywhere else, and then this PR was the result)

PR #4429, opened 2021-04-21
Remove src/c_*.c files from repository

fingolfin:
Instead we generate them dynamically on the fly as needed.

This resolves part of issue #4417.

However, it means that we have to build GAP twice. It also adds some complexity. And I have no idea if it works under cygwin (but hopefully the CI tests will tell us). All in all, I am slightly worried I screwed something up, so I'd like to ponder the changes in here a bit before merging.


fingolfin:
Weird macOS error processing the coverage:
```
  File "dev/ci-coveralls-merge.py", line 47, in <module>
file: gap-coveralls.json done.
    source_digest = md5(name)
merging c-coveralls.json
  File "dev/ci-coveralls-merge.py", line 15, in md5
processing  build/obj/build/.libs/c_oper1.c
    with open(fname, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'build/obj/build/.libs/c_oper1.c'
```

ThomasBreuer (approved):
These changes are a big improvement. I am happy that this change of the setup is possible at all and simplifies the code. I  am sorry that I cannot help to fix the remaining problem.

fingolfin:
The removal of coveralls in #4435 resolved the build failure but something is still fishy, as it prints the warning
```
==> Running gcov in . (disable via -X gcov)
build/c_oper1.c: No such file or directory
build/c_type1.c: No such file or directory
File './src/ariths.h'
Lines executed:68.42% of 38
...
```

Then again, those files are removed from the repo, so who knows, maybe this is "normal" for generated source files (e.g. perhaps the code in question is specifically looking for files which are contained in the git rep).

Anyway, let's wait for the final coverage results and analyze those for plausibility before merging this.

fingolfin:
Something weird is going on, for some reason the HPC-GAP job did not running, for two separate CI runs. I've now rebased this PR once more, and again I don't see it in the list. Huh?????

wilfwilson:
This is possibly because the parameters of the HPC-GAP job kind of match the parameters of the job specified by the matrix, i.e. `os: ubuntu-18.04` and `test-suites: testinstall`. GitHub Actions doesn't build an `include:` job if it is already specified by the matrix.

Now, it's true that the `include:` job for HPC-GAP _also_ has the field `extra: "HPCGAP=yes ABI=64"`, but perhaps GitHub Actions doesn't take that into account, because the matrix doesn't have such a field.

wilfwilson:
Adding `extra: ''` to the matrix might fix things (or, getting rid of the matrix entirely and _only_ having "include"-style jobs).

ChrisJefferson:
I agree it's weird, and I can't see where HPC-GAP disappeared (it seems just in this PR). If it was a "things that only differ by extra", the I'd expect the testpackages testinstall-loadall with ABI=64 and ABI=32 NO_COVERAGE=1 to also have a problem...

wilfwilson:
I don't think it's simply "things that only differ by extra", but "things that differ _from a job specified in the core bit of the matrix_ by extra". You don't have this behaviour with two different "include" jobs.

The matrix is:
```
      matrix:
        os: [ubuntu-18.04]
        test-suites: ["testinstall"]
```
so my belief is that any job listed under `include:` that matches these parameters will not be run.

ChrisJefferson:
Ah, that sounds plausable, and would explain why it's started happening since @fingolfin has removed `docomp`, leaving just `testinstall` in the HPC-GAP job.

wilfwilson:
Confirmed on my fork https://github.com/wilfwilson/gap/actions/runs/792092643 that changing the matrix as follows gets the HPC-GAP job back:
```
      matrix:
        os: [ubuntu-18.04]
        test-suites: ["testinstall"]
        extra: [""]
```

ChrisJefferson (approved):

ChrisJefferson:
Ignoring testing issues, I'm generally happy with this PR, it's both less code than I expected, and doesn't significantly slow down how long GAP takes to build.

fingolfin:
Thanks folks, that makes sense! I actually thought about this cause, but thought it couldn't be the case for now clearly wrong reasons...  I'll augment the PR.

PR #4433, opened 2021-04-23
CI: split and refactor CI scripts

### LINKED ISSUE #4383: CI: split "Compile GAP and download packages" into multiple steps
It would be good to split the CI step "Compile GAP and download packages" into multiple steps. Namely:
- Compile GAP
- Download packages
- Build (some) packages

This is nicer if there are failures in there; e.g. one immediately has a better idea what kind of issue it is. It's also easier to e.g. find the part where io and profiling are compiled, right now one has to scroll through hundreds of lines compiling GAP.

fingolfin:
The main advantage here is that the logs become a bit easier to browse and failures get pinpointed better.

Resolves #4383


wilfwilson (approved):
Cool.

PR #4434, opened 2021-04-26
Add scheduled and manual triggers for the CI GitHub Actions workflow

wilfwilson:
Now that #4389 has been merged, these triggers are in place for the "Wrap releases" workflow, so we may as well do the same for the "CI" workflow, too.


fingolfin (approved):
Fine by me

PR #4436, opened 2021-04-26
CI: revise coverage gathering and uploading

fingolfin:
- use the codecov/codecov-action action
- remove the now obsolete dev/ci-upload-coverage.sh
- simplify dev/ci-gather-coverage.sh

DO NOT MERGE before we've verified that this does not disturb coverage collection.

UPDATE: coverage report seems OK...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -219,9 +219,13 @@ jobs:
             doc/*/*.js
       - name: "Gather coverage data"
         run: ${{ matrix.extra }} dev/ci-gather-coverage.sh
-      - name: "Upload coverage data"
-        run: ${{ matrix.extra }} dev/ci-upload-coverage.sh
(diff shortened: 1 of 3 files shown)


fingolfin:
The removal of coveralls in #4435 resolved the build failure but something is still fishy, as it prints the warning
```
==> Running gcov in . (disable via -X gcov)
build/c_oper1.c: No such file or directory
build/c_type1.c: No such file or directory
File './src/ariths.h'
Lines executed:68.42% of 38
...
```

Then again, those files are removed from the repo, so who knows, maybe this is "normal" for generated source files (e.g. perhaps the code in question is specifically looking for files which are contained in the git rep).

Anyway, let's wait for the final coverage results and analyze those for plausibility before merging this.

wilfwilson (approved):

PR #4437, opened 2021-04-28
Don't hardcode `lib/oper1.g` line numbers in tests

### LINKED ISSUE #4417: changes in `lib/oper1.g`

fingolfin:
Also update a kernel comment which was still referencing `src/c_oper1.c`.

Resolves #4417

### DIFF
--- a/tst/testinstall/varargs.tst
+++ b/tst/testinstall/varargs.tst
@@ -74,21 +74,9 @@ gap> function(a,b....) end;
 Syntax error: ) expected in stream:1
 function(a,b....) end;
                ^
-gap> f := function(a,b..) end;
-Syntax error: Three dots required for variadic argument list in stream:1
-f := function(a,b..) end;
-                 ^^
 gap> Display(RETURN_FIRST);
 function ( first, rest... )
     <<kernel code>> from src/gap.c:RETURN_FIRST
-end
-gap> Print(INSTALL_METHOD_FLAGS,"\n");
-function ( opr, info, rel, flags, baserank, method )
-    <<compiled GAP code>> from GAPROOT/lib/oper1.g:147
-end
-gap> Display(InstallMethod);
-function ( arg... )
-    <<compiled GAP code>> from GAPROOT/lib/oper1.g:331
 end
 gap> [1..2];
 [ 1, 2 ]
--- a/src/streams.c
+++ b/src/streams.c
@@ -435,10 +435,9 @@ Int READ_GAP_ROOT ( const Char * filename )
     // try to find compiled version of the GAP file
     if (SyUseModule) {
         // This code section covers transparently loading GAC compiled
-        // versions of GAP source files, by running code similar to
-        // that in FuncLOAD_STAT. For example, lib/oper1.g is compiled
-        // into C code which is stored in src/c_oper1.c; when reading
-        // lib/oper1.g, we instead will load its compiled version.
+        // versions of GAP source files, by running code similar to that in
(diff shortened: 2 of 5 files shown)


tst/testinstall/varargs.tst
```diff
@@ -74,22 +74,10 @@ gap> function(a,b....) end;
 Syntax error: ) expected in stream:1
 function(a,b....) end;
                ^
-gap> f := function(a,b..) end;
-Syntax error: Three dots required for variadic argument list in stream:1
-f := function(a,b..) end;
-                 ^^
```
fingolfin, line 80: I remove this test because it is tests the same thing as the test starting in line 57.

tst/testinstall/varargs.tst
```diff
@@ -74,22 +74,10 @@ gap> function(a,b....) end;
...
 gap> Display(RETURN_FIRST);
 function ( first, rest... )
     <<kernel code>> from src/gap.c:RETURN_FIRST
 end
-gap> Print(INSTALL_METHOD_FLAGS,"\n");
-function ( opr, info, rel, flags, baserank, method )
-    <<compiled GAP code>> from GAPROOT/lib/oper1.g:147
-end
```
fingolfin, line 88: This test actually didn't even make sense in this file, so I am not sure why it was here to begin with

ThomasBreuer (approved):
Very nice.

wilfwilson (approved):

Issue #4403, opened 2021-04-15
Documentation of random sources 


ThomasBreuer:
When trying to provide a new kind of random sources (see oscar-system/GAP.jl/pull/642), I noticed that the documentation of GAP's random sources leaves some open questions.
- The documentation says that any GAP random source must provide methods for the operations `State`, `Reset`, and `Init`. It should say that in the case of `Reset` and `Init`, methods for the two argument versions must be provided, and that the second argument may be an integer or a `State` value.
  In fact, the unary version of `Init` is documented but GAP does not provide any such method. I think the documentation should be adjusted to the code.
- It is not specified whether `Init( rs, state )` makes a copy of `state`. In fact, I would like to state that this is not the case in the implementation mentioned above. That is, the user must copy `state` before calling `Init( rs, state )` if changing `state` in later calls `Random( rs, ... )` is not wanted. I think the documentation should be adjusted to the code.
- The code in `lib/random.gi` shows that `Init( rs, seed )` or `Init( rs, state )` is admissible also if `rs` is an already initialized random source, and then has the same effect as `Reset` with the same arguments (except of course that `Init` returns `rs` but `Reset` returns a copy of the old `State` value). I think the documentation should be adjusted to the code.
- In my situation, negative integers aren't admissible as seeds. Perhaps the documentation should say that such restrictions may apply, depending on the type of the random source.

fingolfin:
Actually, I don't understand why `Init` works like that -- it seems odd to require the user to provide a previously made object. Shouldn't we instead tell them to use `RandomSource(filter, seed)` ? Indeed, looking at the code, the only places calling `Init` in GAP are a `RandomSource` method and the `Reset` methods for `IsGAPRandomSource` and `IsMersenneTwister`. As such, it seems more like an implementation detail; or perhaps an "internal API" for people who want to implement a random source; but not something intended for end users. 

I wonder why not instead `RandomSource` is a constructor (instead of an operation); then we could do away with `Init` and instead implement methods for `RandomSource`. This would break `IsRealRandomSource` from `io`, but otherwise I think it'd be fine? (Well, one annoyance would be that we could then not implement the one-argument variant of `RandomSource` generically (as I recently rediscovered, we can't have "generic" constructor methods). But this could be dealt with by having `RandomSource` stay as a method and instead add `DeclareConstructor("RandomSourceCons", [IsRandomSource, IsObject]);`

[...]

Issue #4424, opened 2021-04-21
Support 0-generator `FreeMagma` and `FreeSemigroup`?


wilfwilson:
This comes from @markuspf in #1385:

> Though it is documented that the rank has to be positive, but why should there not be a `FreeSemigroup(0)` (which would be empty)?

This may or may not be a good idea. Possibly we also want 0-generator `FreeMonoid` and `FreeMagmaWithOne` (which would be trivial).

fingolfin:
I am neutral on this. But the work invested in making this possible should be proportional to how useful this is; if this is just there out of completeness, clearly it's far less pressing; OTOH if somebody could really simplify their code if they had this... But to me, it seems somewhat esoteric?

I am also slightly wary whether such an empty collection might break other parts of the code which do not expect such things. Of course *that* code then could also be fixed; so what I am really wary about is that the amount of work needed to make this happen might spiral out of control.

That said, I have no fundamental objections at all, so if someone wants to make this happen, you are most welcome :-).

ThomasBreuer:
I think supporting empty domains is asking for trouble. (For example, sooner or later people will ask for maps and morphisms between them, etc., and already the question when two such maps are equal might be challenging in GAP.)

The documentation of `IsMagma` and `IsSemigroup` does not forbid empty magmas and semigroups, and `Subsemigroup( SymmetricGroup( 3 ), [] )` is printed as `<empty semigroup>`; thus there is some support for empty domains.
On the other hand, `Semigroup( [] )` and `MagmaByMultiplicationTable( [] )` run into errors.

Perhaps the term "0-generator objects" mentioned in issue #1385 is misleading, empty lists of generators should be supported if the domains themselves are not empty; for example the trivial subgroup of a group usually has an empty list of generators, and tests for this situation are important; but this is a different issue.

wilfwilson:
I've thought about this on-and-off recently, and I have settled on thinking that we should **not** support `FreeMagma(0)` and `FreeSemigroup(0)`. Although empty magmas and semigroups are mathematically defined, and possible to construct in GAP as @ThomasBreuer demonstrates, I have never seen 'empty free magmas/semigroups' being talked about as 'free' objects. Until/unless someone comes up with a use case for `FreeMagma(0)` and `FreeSemigroup(0)` (which I don't foresee), then I think we should **not** proactively add such support.

Therefore I will close this issue.

Note that `FreeMagma(0)` and `FreeSemigroup(0)` are now rejected in a consistent way:
```
[...]

PR #4438, opened 2021-04-28
improved documentation for random sources

### LINKED ISSUE #4379: Random for non-dense lists

ThomasBreuer:
- document that `Random` for a list is defined only for nonempty and
  dense lists; the declaration for arbitrary lists is kept in order
  not to break code in the IO package

- install `Random` methods for lists with the requirement `IsDenseList`
  (thus we get a "no method found" error if a non-dense list is given)

- document that `Random` for two integers is defined only if the first
  is not larger than the second

- describe the user interface for random sources in the subsection
  about `IsRandomSource`

- added a subsection on developing new kinds of random sources

- point to the user interface and the developer interface in the
  manual section about random sources

- added a manual example for `RandomList`

- added some cross-references

- call `Reset` with one argument in `START_TEST`
  (does not change the behaviour)

- document `Init` only via an index entry, since is not needed by users,
  removed the unary version of `Init`, as it is apparently not used

- removed the `Random` method for `IsMersenneTwister` and `IsList`,
  because the generic method is sufficient

- fixed `Random( IsMersenneTwister, 1, 0 )` and added a few tests;
  in general, we do not promise error messages if the arguments do not fit,
  thus we do not need tests for this situation

(resolves #4379 and #4403;
replaces #4415: in a discussion with @fingolfin, we decided better not to extend the documentation of `Random` to non-dense lists)

### DIFF
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -1613,8 +1613,8 @@ DeclareAttribute( "RepresentativeSmallest", IsListOrCollection );
 
 #############################################################################
 ##
-#O  Random( <C> ) . . . . . . . . . .  random element of a list or collection
-#O  Random( <list> )  . . . . . . . .  random element of a list or collection
+#O  Random( <C> ) . . . . . . . . . . random element of a nonempty collection
+#O  Random( <list> )  . . . . . . .  random element of a dense, nonempty list
 #O  Random( <from>, <to> )
 ##
 ##  <#GAPDoc Label="Random:coll">
@@ -1627,17 +1627,29 @@ DeclareAttribute( "RepresentativeSmallest", IsListOrCollection );
 ##  <Index Key="Random"><Ref Oper="Random" 
 ##                           Label="for a list or collection"/></Index> 
 ##  <Ref Oper="Random" Label="for a list or collection"/> returns a
-##  (pseudo-)random element of the list or collection <A>listorcoll</A>.
+##  (pseudo-)random element of the dense, nonempty list or nonempty
+##  collection <A>listorcoll</A>.
+##  The behaviour for non-dense or empty lists, and for empty collections
+##  (see <Ref Filt="IsDenseList"/>, <Ref Prop="IsEmpty"/>)
+##  is undefined.
 ##  <P/>
 ##  As lists or ranges are restricted in length (<M>2^{28}-1</M> or 
 ##  <M>2^{60}-1</M> depending on your system), the second form returns a
 ##  random integer in the range <A>from</A> to <A>to</A> (inclusive) for
 ##  arbitrary integers <A>from</A> and <A>to</A>.
+##  The behaviour in the case that <A>from</A> is larger than <A>to</A>
+##  is undefined.
+##  <P/>
+##  See Section <Ref Sect="Random Sources"/> for more about computing
+##  random elements, in particular for
+##  <Ref Oper="Random" Label="for random source and list"/> methods
+##  that take a random source as the first argument.
 ##  <P/>
 ##  The distribution of elements returned by
 ##  <Ref Oper="Random" Label="for a list or collection"/> depends
 ##  on the argument.
-##  For a list the distribution is uniform (all elements are equally likely).
+##  For a dense, nonempty list the distribution is uniform (all elements are
+##  equally likely).
 ##  The same holds usually for finite collections that are
 ##  not lists.
 ##  For infinite collections some reasonable distribution is used.
@@ -1669,6 +1681,8 @@ DeclareAttribute( "RepresentativeSmallest", IsListOrCollection );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
+# We keep the declaration for non-dense lists
+# in order not to break existing code.
 DeclareOperation( "Random", [ IsListOrCollection ] );
 DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 
@@ -1695,7 +1709,7 @@ DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 
 #############################################################################
 ##
-#F  RandomList( <list> )
+#F  RandomList( [<rs>, ]<list> )
 ##
 ##  <#GAPDoc Label="RandomList">
 ##  <ManSection>
@@ -1707,11 +1721,24 @@ DeclareOperation( "Random", [ IS_INT, IS_INT ] );
 ##  <Ref Func="RandomList"/> returns a (pseudo-)random element with equal
 ##  distribution.
 ##  <P/>
-##  The random source <A>rs</A> is used to choose a random number.
+##  The random source <A>rs</A> (see <Ref Sect="Random Sources"/>)
+##  is used to choose a random number.
 ##  If <A>rs</A> is absent,
 ##  this function uses the <Ref Var="GlobalMersenneTwister"/> to produce the
 ##  random elements (a source of high quality random numbers).
 ##  <P/>
+##  <Example><![CDATA[
+##  gap> RandomList( [ 1 .. 6 ] );
+##  3
+##  gap> elms:= AsList( Group( (1,2,3) ) );;
+##  gap> RandomList( elms );  RandomList( elms );
+##  (1,2,3)
+##  (1,3,2)
+##  gap> rs:= RandomSource( IsMersenneTwister, 1 );
+##  <RandomSource in IsMersenneTwister>
+##  gap> RandomList( rs, elms );
+##  (1,2,3)
+##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
--- a/lib/random.gd
+++ b/lib/random.gd
@@ -21,15 +21,23 @@
 ##  <Filt Name="IsRandomSource" Arg='obj' Type='Category'/>
 ##
 ##  <Description>
-##  This is the category of random source objects which are defined to have,
-##  for  an  object  <A>rs</A>  in  this  category,  methods  available  for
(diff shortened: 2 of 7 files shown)


lib/random.gd
```diff
@@ -74,47 +86,46 @@ DeclareCategory( "IsRandomSource", IsComponentObjectRep );
...
-##  data which are needed. Optionally, it should allow one to specify a 
-##  <A>seed</A> for the initial state, as explained for <Ref Oper="Reset"/>.
+##  <C>Reset( <A>rs</A>, <A>seed</A> )</C> resets the random source <A>rs</A>
+##  to a state described by <A>seed</A>, if the random source can be reset;
+##  otherwise it does nothing.
+##  Here <A>seed</A> can be an output of <Ref Oper="State"/> and then
+##  <A>rs</A> gets reset to that state.
+##  Also integers are allowed as <A>seed</A>.
+##  Without the <A>seed</A> argument the default <M><A>seed</A> = 1</M> is
+##  used.
```
fingolfin, line 123: How about changing this as follows, and not making any claims about what the default seed is?
```suggestion
##  Without the <A>seed</A> argument a fixed default seed is used.
```

lib/random.gd
```diff
@@ -74,47 +86,46 @@ DeclareCategory( "IsRandomSource", IsComponentObjectRep );
...
-##  <Ref Oper="Init"/> is the constructor of a random source, it gets an empty component 
-##  object <A>prers</A> which has already the correct type and should fill in the actual 
-##  data which are needed. Optionally, it should allow one to specify a 
-##  <A>seed</A> for the initial state, as explained for <Ref Oper="Reset"/>.
+##  <C>Reset( <A>rs</A>, <A>seed</A> )</C> resets the random source <A>rs</A>
+##  to a state described by <A>seed</A>, if the random source can be reset;
+##  otherwise it does nothing.
+##  Here <A>seed</A> can be an output of <Ref Oper="State"/> and then
+##  <A>rs</A> gets reset to that state.
+##  Also integers are allowed as <A>seed</A>.
```
fingolfin, line 121: Note that for some random source types, all seeds may be integers, so "also" is not quite appropriate. Moreover, I'd like to deemphasize use of hardcoded integer seeds.
```suggestion
##  For historical reasons, random sources accept integer values as <A>seed</A>.
##  We recommend that new code should not rely on this; always use the output
##  of a prior call to <Ref Oper="State"/> as <A>seed</A>, or omit it.
```

lib/random.gd
```diff
@@ -211,22 +225,30 @@ fi;
...
-##  <A>cat</A> is the category describing the type of the random generator, 
-##  an optional <A>seed</A> which can be an integer or a type specific data 
-##  structure can be given to specify the initial state.
+##  This operation is used to create new random sources. The first argument
+##  <A>cat</A> is the category describing the type of the random generator,
+##  for example one of the categories listed in
+##  Section <Ref Subsect="Kinds of Random Sources"/>.
+##  <P/>
+##  An optional <A>seed</A> can be given to specify the initial state.
+##  It can be an integer or a type specific data structure.
```
fingolfin, line 234: Perhaps here we should just refer to the documentation of `Reset`?
```suggestion
##  For details, see  <Ref Oper="State"/> and  <Ref Oper="Reset"/>
```

lib/random.gd
```diff
@@ -237,6 +259,70 @@ fi;
...
+##############################################################################
+##
+##  <#GAPDoc Label="RandomSource_develop">
+##  <Subsection Label="Implementing new kinds of random sources">
+##  <Heading>Implementing new kinds of random sources</Heading>
+##
+##  If one wants to implement a new kind of random sources then
+##  the first step is the declaration of a new category <C>C</C>, say,
+##  that implies <Ref Filt="IsRandomSource"/>, analogous to the categories
+##  listed in Section <Ref Subsect="Kinds of Random Sources"/>.
```
fingolfin, line 272: Perhaps we should be explicit and mention this is usually done by calling `DeclareCategory("C", IsRandomSource);`

lib/random.gd
```diff
@@ -237,6 +259,70 @@ fi;
...
+##  <P/>
+##  Then the following method installations are needed.
+##  <P/>
+##  <Index Key="Init"><C>Init</C> (initialize a random source object)</Index>
+##  <C>InstallMethod( Init, [ C, IsObject ], function( prers, seed_or_state )
+##  ... end );</C>
+##  <P/>
+##  Here <C>prers</C> is an empty component object (which has already the
+##  filter <C>C</C>), and <C>seed_or_state</C> is an integer or a state value
```
fingolfin, line 281: Why "seed_or_state"? What you call "state value" is what I'd call "seed" (which does not need to be an integer). This also matches the nomenclature in the existing user facing documentation, I think?

lib/random.gd
```diff
@@ -237,6 +259,70 @@ fi;
...
+##  If <C>rs</C> supports resetting then the function must reinitialize
+##  <C>rs</C> to the integer or <Ref Oper="State"/> value <C>seed_or_state</C>
+##  and must return the <Ref Oper="State"/> value of <C>rs</C> before these
+##  changes; if resetting is not supported then <Ref Func="ReturnNothing"/>
+##  should be installed.
+##  Reset need not copy a given state.
+##  Note that the generic unary <Ref Oper="Reset"/> method uses the default
+##  seed <C>1</C>.
+##  <P/>
+##  An example of an implementation as described here is given by the
```
fingolfin, line 319: of course the library implementations for `IsGAPRandomSource` and `IsMersenneTwister` are also examples... :-)

fingolfin (commented):
Thanks, that's a very nice improvement!

I left some comments

lib/random.gd
ThomasBreuer: O.k.

lib/random.gd
ThomasBreuer: O.k.

lib/random.gd
ThomasBreuer: The method installations for `IsMersenneTwister` fit.
Concerning `IsGAPRandomSource`, the code does not follow the recommended pattern: A `Random` method for a list gets installed, which relies on a default method that delegates the two-integer variant to the list variant.

lib/random.gd
ThomasBreuer: Good idea.

lib/random.gd
ThomasBreuer: O.k.

lib/random.gd
ThomasBreuer: O.k.

lib/random.gi
```diff
@@ -201,18 +203,18 @@ InstallMethod(Reset, [IsMersenneTwister, IsObject], function(rs, seed)
   return old;
 end);
 
-InstallMethod(Random, [IsMersenneTwister, IsList], function(rs, list)
-  return list[Random(rs, 1, Length(list))];
-end);
```
fingolfin, line 206: @ChrisJefferson I have a vague memory that when you refactored this code last, we had to keep some methods (for ... something...?) instead of generic methods due to some issues that I forgot about... I am hoping this is not relevant here (in particular, I believe the tests for Random that you wrote would hopefully catch any such issues), but perhaps you could have a quick look at the changes in this PR just in case you remember this better than I do... :-)

fingolfin (approved):

lib/coll.gd
```diff
@@ -1627,17 +1627,29 @@ DeclareAttribute( "RepresentativeSmallest", IsListOrCollection );
 ##  <Index Key="Random"><Ref Oper="Random" 
 ##                           Label="for a list or collection"/></Index> 
 ##  <Ref Oper="Random" Label="for a list or collection"/> returns a
-##  (pseudo-)random element of the list or collection <A>listorcoll</A>.
+##  (pseudo-)random element of the dense, nonempty list or nonempty
+##  collection <A>listorcoll</A>.
+##  The behaviour for non-dense or empty lists, and for empty collections
```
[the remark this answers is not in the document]
ThomasBreuer: As I mentioned in a comment for #4424, there is some support for empty semigroups.

lib/random.gi
ChrisJefferson: At some point when I worked on this I tried to make sure that GAP would produce the same sequence of random numbers on different versions -- now I think that wasn't worth the effort, as for most types we don't make that promise anyway. In some cases I found infinite loops in packages if I removed the wrong function, but if this passes all package tests then I think it's OK.

ThomasBreuer: I see the situation as follows.

We have generic `Random` methods for the cases `[ IsRandomSource, IsList and ... ]` and `[ IsRandomSource, IsInt, IsInt ]`.
They delegate to each other, with the effect that for specific kinds of random sources, one has to install only one such method. (If one does not install any `Random` method for a new kind of random sources, one runs into infinite loops, thus the generic methods are to some extent dangerous. However, we do not support "generic random sources".)
The `Random` methods for different specific kinds of random sources do not compete with each other, thus only the specific method (if available) and the generic method are applicable.

In the case of `IsMersenneTwister`, the specific method for `IsList` which I propose to remove does the same as the generic method for this case, thus the removal should not change anything at runtime, in particular the sequence of random numbers should not be affected.

ChrisJefferson (approved):

PR #4439, opened 2021-04-28
fixed `ExamplesReportDiff`

ThomasBreuer:
I am suspicious, perhaps I have overlooked something obvious.

Or is it possible that this function was never run on manual examples with wrong output?


ChrisJefferson:
I've certainly not used it myself. Does it work after this fix (the fix looks very reasonable)

ThomasBreuer:
@ChrisJefferson I have tried the function "by hand", and found the error.
However, I thought that the `testmanuals` CI test uses the function, and then this test does not work as expected.

(My situation was as follows:
Rewrite some manual examples (involving `Random` calls), create the manual anew, extract the manual examples, and produce the correct output, w.r.t. chapter-wise testing, by running the tests for the chapters in question.)

fingolfin:
Well testmanuals uses it but of course if there is no error we don't notice a bug in the function reporting errors...

fingolfin (approved):

PR #4442, opened 2021-04-29
Fix crash when incorrectly accessing a compressed vector `v` via  `v{...}[...]` or `v{...}{...}` as  if it was a matrix 

fingolfin:
For a compressed vector `v`, writing code involving `v{...}[...]` and
`v{...}{...}` would crash GAP.

A subset of this was originally reported by me in 2013 (on the old GAP issue
tracker, hence the issue number 312 you may see inside this PR is not valid
here on GitHub), and later fixed with a workaround. However, back then I did
not realize that there are in total four places that needed the fix, and so
only fixed one of them. I also thought it was just a workaround for a deeper
issue. Upon reanalyzing, I believe this was wrong; the real bug always was
lack of input validation combined with an incorrect use of a feature for
accessing slices of matrices with a vector as input instead of a matrix.

For further details, please read the comment in the test file in this patch.

### DIFF
--- a/tst/testbugfix/2013-03-12-t00285.tst
+++ b/tst/testbugfix/2013-03-12-t00285.tst
@@ -1,5 +1,58 @@
 # 2013/03/12 (MH)
-gap> v:=IdentityMat(28,GF(2))[1];
-<a GF2 vector of length 28>
-gap> v{[1..Length(v)]}{[1..5]};
-Error, List Elements: <lists> must be a list (not a data object)
+#
+# The following tests used to crash. This is because for compressed FFE
+# vectors, we install custom methods for ELMS_LIST (ELMS_VEC8BIT,
+# ELMS_GF2VEC...) which return compressed vectors again -- which are not plain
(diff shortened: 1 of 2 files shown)


ChrisJefferson (approved):

tst/testbugfix/2013-03-12-t00285.tst
```diff
@@ -1,5 +1,58 @@
 # 2013/03/12 (MH)
-gap> v:=IdentityMat(28,GF(2))[1];
-<a GF2 vector of length 28>
-gap> v{[1..Length(v)]}{[1..5]};
-Error, List Elements: <lists> must be a list (not a data object)
+#
+# The following tests used to crash. This is because for for compressed FFE
```
wilfwilson, line 3: If you happen to make further changes to this PR, then perhaps change this too. Otherwise, no problem.
```suggestion
# The following tests used to crash. This is because for compressed FFE
```

wilfwilson (approved):

PR #4443, opened 2021-04-29
CI: send Slack notifications on manual & scheduled runs

wilfwilson:
Currently, the CI only thinks about whether to send a notification if there is a push. However, there are now more triggers where it would be appropriate to send notifications: the newly-added scheduled runs, and any manually-dispatched runs. I didn't think to add them in #4434.

Indeed, the existence of these runs might cause weirdness if we _don't_ send notifications for them: I think they count as runs on the `master` branch (unless a workflow is manually dispatched for a non-`master` branch), and so the following can happen:
- A push to `master` newly fails. A failure notification is sent to Slack.
- Things are fixed in a scheduled job. No notification is sent, because it's not yet implemented.
- The next push to `master` also passes, and the CI checks the status of the previous run on the `master` branch. The previous run was the scheduled job, which also passed, so again, no notification is sent.

So in this scenario, the failure notification is sent, but the success notification is not.

(Note: ultimately we'll want to add these notifications to the `Wrap releases` workflow too, but I propose we only do that when things settle down with the initial implementation.)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -267,12 +267,12 @@ jobs:
         shell: C:\cygwin64\bin\bash.exe --login -o igncr '{0}'
         run: dev/ci.sh
 
-  slack-notification-on-failure:
-    name: Send slack notification on CI failure
+  slack-notification:
+    name: Send Slack notification on status change
     needs:
       - test-unix
       - test-cygwin
-    if: ${{ always() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
+    if: ${{ always() && github.event_name != 'pull_request' && github.repository == 'gap-system/gap' }}
     runs-on: ubuntu-latest
 
     steps:
@@ -289,7 +289,7 @@ jobs:
           notify_on_changed_status: true
       - name: Send slack notification
         uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
-        if: steps.should_notify.outputs.should_send_message == 'yes'
+        if: ${{ steps.should_notify.outputs.should_send_message == 'yes' }}
         with:
           status: ${{ steps.should_notify.outputs.current_status }}
         env:


.github/workflows/CI.yml
```diff
@@ -272,7 +272,7 @@ jobs:
     needs:
       - test-unix
       - test-cygwin
-    if: ${{ always() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
+    if: ${{ always() && github.event_name != 'pull_request' && github.repository == 'gap-system/gap' }}
```
fingolfin, line 275: Actually can't we even write this?
```suggestion
    if: ${{ github.event_name != 'pull_request' && github.repository == 'gap-system/gap' }}
```

wilfwilson: No, it is necessary, the `needs:` context implies by default that the `slack-notification-on-failure` job (which I will rename) would only run if `test-unix` and `test-cygwin` are _successful_

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-requiring-dependent-jobs-to-be-successful

fingolfin: The documentation also says that "always()" simply always returns `true`. There is nothing magical about it, as gar as I could tell

wilfwilson: Ah right, I see, thanks for the further explanation. I'll give it a go.

wilfwilson: It seems that `always()` _is_ magical.

In a branch on my fork I broke the tests and removed the `always()`. Without the `always()`, the Slack job was skipped: https://github.com/wilfwilson/gap/actions/runs/795815484 but when I changed nothing except to add back the `always()`, the Slack job was run https://github.com/wilfwilson/gap/actions/runs/795831635, as it should be.

wilfwilson: That must mean that, in the absence of an `always()`, the GitHub Actions software implicitly adds the condition `!failure() &&` to the `if:` line.

fingolfin: OK, sorry for wasting your time with this then -- and I guess the GitHub docs need to be improved :/

.github/workflows/CI.yml
```diff
@@ -267,12 +267,12 @@ jobs:
...
 
-  slack-notification-on-failure:
-    name: Send slack notification on CI failure
+  slack-notification:
+    name: Send Slack notification on status change
     needs:
       - test-unix
       - test-cygwin
-    if: ${{ always() && github.event_name == 'push' && github.repository == 'gap-system/gap' }}
+    if: always() && github.event_name != 'pull_request' && github.repository == 'gap-system/gap'
```
fingolfin, line 275: Wait, you can commit the `${{ ... }}$` ? On https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions it sounds as if that's required?! (Their docs really could be better)

fingolfin: On https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif it says

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within ${{ }} to explicitly mark it for evaluation.

Since there are operators in there, I assume we *do* need it?

wilfwilson: Well, we don't _need_ it, because I've found that it does work without the `${{ ... }}`, but according to the documentation, it is needed. I can re-add the `${{ ... }}` if you insist.

We have been using this kind of syntax already for a while without problems; see several lines further down the workflow file:
```
if: steps.should_notify.outputs.should_send_message == 'yes'
```

fingolfin: I am simply a bit wary about relying on undocumented behavior that explicitly contradicts the documentation, even if it works right now anyway...

fingolfin (approved):

PR #4453, opened 2021-05-04
Remove `Autoload` for package documentation from PackageInfo.g example

fingolfin:
It's not been in use for a long time.


alex-konovalov (approved):
yes, certainly a relic from the times when not all manual books were loaded.

PR #4454, opened 2021-05-04
Refactor SMTX_NewEqns, SMTX_NullspaceEqns

fingolfin:
Now uses MatrixObj APIs

Motivated by some performance tuning work I did on the MeatAxe some weeks ago for an example by @danielrademacher ; that will come in a separate PR, but for some reasons I also made this cleanup, and it seems a pity to discard it.

### DIFF
--- a/tst/testinstall/meatauto.tst
+++ b/tst/testinstall/meatauto.tst
@@ -1,4 +1,4 @@
-#@local F, e, M, v
+#@local F, e, M, v, N
 gap> START_TEST("meatauto.tst");
 
 #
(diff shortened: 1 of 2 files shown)


tst/testinstall/meatauto.tst
```diff
@@ -20,7 +20,7 @@ rec( dim := 3, failed := false, field := GF(2), index := [  ],
   mat := [ <a GF2 vector of length 3>, <a GF2 vector of length 3> ], 
   vec := [ 0*Z(2), Z(2)^0 ], weights := [ 1, 2 ] )
 gap> SMTX_NullspaceEqns(e);
-[ [ Z(2)^0, 0*Z(2), Z(2)^0 ] ]
+[ <a GF2 vector of length 3> ]
```
danielrademacher, line 23: Is it possible to see the "GF2 vector of length 3"? Since the tests worked, they are most probably the same.

danielrademacher (approved):
Looks good. A lot of improvements if one knows how to work with matrices in GAP. 👍

tst/testinstall/meatauto.tst
fingolfin: Good point, I've added a test to check the return value.

danielrademacher: Great, thanks!

PR #4455, opened 2021-05-04
Small changes to manual: add index entry for `\n`, and add labels to `FactorCosetAction` documentation

wilfwilson:
@RebeccaAHW noticed that the documentation for `\n` was uniquely missing an entry in the index of the reference manual, unlike say `\"`, `\\` and `\b`.

I think this is probably because, if you follow the pattern of the other characters, and simply use `\n` itself as the index key, then you get this scary message when compiling the manuals:
```
#I Writing LaTeX file, 4 x pdflatex with bibtex and makeindex,
#W There were LaTeX errors:
! Undefined control sequence.
<argument> \string \indexentry {\n
                                   @\texttt {\texttt {\symbol {92}}n}|hyperp..\
.
____________________
```
But if I use the key `\\n` instead, then everything seems to work well.

In making this change, I also noticed and added some missing labels:
```
#W There were LaTeX Warnings:
LaTeX Warning: Label `FactorCosetAction' multiply defined.

)
____________________
LaTeX Warning: There were multiply-defined labels.
```

### DIFF
--- a/doc/ref/string.xml
+++ b/doc/ref/string.xml
@@ -153,7 +153,10 @@ The following special character sequences are currently defined.
 For any other sequence starting with a backslash, the backslash is ignored.
 <P/>
 <List>
-<Mark><C>\n</C></Mark>
+<Mark>
+<Index Key='\\n'><C>\n</C></Index>
+<Index>newline character</Index>
+<C>\n</C></Mark>
 <Item>
     <E>newline character</E>.
     This is the character that, at least on UNIX systems,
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

alex-konovalov (approved):

wilfwilson:
> if I use the key `\\n` instead, then everything seems to work well.

Although this means that the `\n` isn't sorted into _quite_ the right position in the matrix (as you can see from the build artifact that was uploaded in the CI). But I can't work out how to get it in the right place. I think this is good enough (for now, at least).

PR #4458, opened 2021-05-04
Fix printing of strings inside functions

ChrisJefferson:
Previously we normalised all whitespace in printed out functions,
which would effect whitespace inside strings.

Previously we would print out functions, then to get rid of all whitespace using `NormalizeWhitespace`. However, this would also normalise whitespace inside strings.


fingolfin:
Why do we modify the whitespace at all?

ChrisJefferson:
Guessing, I assume it is because we want the String to be a single line, and concatenating the lines together would create huge spaces from indendantion.

wilfwilson (commented):
Seems to do the job nicely, thanks, but I have a question and a suggestion.

lib/function.gi
```diff
@@ -78,13 +78,21 @@ end);
...
     CloseStream(stream);
-    NormalizeWhitespace(s);
-    return MakeImmutable(s);
+    lines := SplitString(s, "\n");
+    for i in [1..Length(lines)] do
+        Chomp(lines[i]);
+        pos := 1;
+        while pos <= Length(lines[i]) and lines[i][pos] = ' ' do
+            pos := pos + 1;
+        od;
```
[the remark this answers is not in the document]
fingolfin: Or maybe just do
```gap
  lines := List(lines, l -> StripBeginEnd(l, " "));
```

lib/function.gi
```diff
@@ -78,13 +78,21 @@ end);
 
 InstallMethod(String, "for a function, with whitespace reduced", [IsFunction and IsInternalRep],
 function(fun)
-    local  s, stream;
+    local  s, stream,lines, pos, i;
     s := "";
     stream := OutputTextString(s, true);
```
fingolfin, line 83:
```suggestion
    stream := OutputTextString(s, true);
    SetPrintFormattingStatus(s, false);
```
And then add test which involve printing a function with a loooong string of whitespace, one which involves line wrapping because it is so long...

ChrisJefferson: There are other issues with this PR, but this is the most important one to fix, and I think fixing this will deal with many other issues.

ChrisJefferson:
There were several issues with this PR, but most of the code disappeared while fixing @fingolfin's comment (as removing formatting removed the indentation I needed to get rid of)

lib/function.gi
fingolfin: Interesting, I didn't expect that!

Still would be nice to have that test with a function that has a looooong whitespace string in it, just to make sure we don't regress on this in future refactorings.

ChrisJefferson: I do have a test (I make the string with concatenation, just because long tests can look awkward)

fingolfin: Ok, but that test is not yet in here, is it? At least I don't see it

fingolfin (approved):
Thanks

wilfwilson (approved):

PR #4463, opened 2021-05-05
kernel: use sigaction instead of signal to handle SIG_INT

fingolfin:
This allows us to get rid of siginterrupt which is deprecated on Linux these days, and triggers compiler warnings like these:
```
src/sysfiles.c: In function ‘syAnswerIntr’:
src/sysfiles.c:997:5: warning: ‘siginterrupt’ is deprecated: Use sigaction with SA_RESTART instead [-Wdeprecated-declarations]
  997 |     siginterrupt( SIGINT, 0 );
      |     ^~~~~~~~~~~~
```

### DIFF
--- a/src/sysfiles.c
+++ b/src/sysfiles.c
@@ -992,10 +992,6 @@ static void syAnswerIntr(int signr)
         SyExit( 1 );
     }
 
-    /* reinstall 'syAnswerIntr' as signal handler                          */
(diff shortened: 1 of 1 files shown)


fingolfin:
@ChrisJefferson it would be great if you could have a look to see if I did this right...

ChrisJefferson (approved):

PR #4464, opened 2021-05-06
buildsys: improve cross compiling experience

### LINKED ISSUE #4460: buildsys: cross compilation broken due to dynamic generation of build/c_type1.c and build/c_oper1.c
Since we generate `build/c_type1.c` and `build/c_oper1.c` dynamically now, there is a problem when cross compiling GAP: we are compiling GAP on a machine with architecture A with the desire of producing an executable that runs on architecture B. But we also need to run GAP to generate `build/c_type1.c` and `build/c_oper1.c`, and for that we must have a GAP executable that runs on architecture A.

I see three primary options to deal with this:
1. add those files back to the git repository (I don't like this one, but I list it for completeness)
2. "enhance" the build system to properly deal with cross compilation (this can be done, but it will complicate the build system further, quite a bit, for a relatively fringe use case -- however it is important for the Julia version of GAP, where I cross compile GAP for multiple architectures; and I am guessing that some Linux distros also use cross compilation to produce binaries for multiple platforms (but maybe they just use native build environments?)
3. we leave things as they are in the git version of GAP, but bundle `build/c_type1.c` and `build/c_oper1.c` with GAP releases (and ensure in the build system that these bundled files are used by default).

The advantage of option 3 is that it is relatively simple to implement (I think): We already build GAP as part of `dev/releases/make_archives.py`, so it would be trivial to retain `build/c_type1.c` and `build/c_oper1.c` from there. In order to protect them from `make clean`, we could copy them back to their old place in `src`; then the build system would have to be modified to use those (in release builds, at least) instead of trying to generate `build/c_type1.c` and `build/c_oper1.c`. 

One remaining issue would be that for HPC-GAP, a different version of those files is needed. We could either also generate those as part of the release script, or just say that cross compilation is not supported out of the box for HPC-GAP built from a GAP release tarball.

Otherwise, the required build system modifications are very simple, so I think I'll just provide a patch for this.

I also looked into option 2 last night; what appeals to me about it is that the same code would be used in git and release versions of GAP; . I've done similar things in the past, most recently I tweaked the Singular build system last year to deal with something similar. However, it seems to me that doing this requires quite a lot more effort, and will further complicate our build system for a relatively fringe use case (well, it won't be super bad, but still...). All in all I am reluctant to go down this road, although if someone wanted to work on this, I'd be happy to discuss details. Here's just a sketch:
- use [`AX_PROG_CC_FOR_BUILD`](https://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html) and [`AX_PROG_CXX_FOR_BUILD`](https://www.gnu.org/software/autoconf-archive/ax_prog_cxx_for_build.html) to detect compilers for the build system
  - for performance reasons, I'd suggest to only invoke these if `cross_compiling` is set to `yes`)
- export the variables these define, such as `CC_FOR_BUILD`, `CFLAGS_FOR_BUILD`, etc., by editing `GNUmakefile.in` suitably
- also export whether cross compilation is used (something like `AC_SUBST([CROSS_COMPILE], [$cross_compiling])` or so)
- in `Makefile.rules`, change how `OBJS_NOCOMP` is generated if `CROSS_COMPILE = yes`: say, add the target architecture as a prefix (so that instead of `build/obj/src/ariths.c.lo` we ask for `build/obj/TARGETARCH/src/ariths.c.lo`, so e.g. `build/obj/x86_64-pc-linux-gnu/src/ariths.c.lo` or whatever.
- duplicate (?) the build rules for C and C++ files and adjust them for this new pattern. Something like this could be a starting point
   - but this *really* needs further changes!!! E.g., I didn't bother to lookup what the correct variable name for `TARGETARCH is; we also need `GAP_CFLAGS_FOR_BUILD`, and soon.. Alsonote that I deliberately dropped `WARN_CFLAGS`: no need for those here, and i). 
```
build/obj/TARGETARCH/%.c.lo: %.c cnf/GAP-CFLAGS cnf/GAP-CPPFLAGS libtool build/TARGETARCH/config.h
	@$(MKDIR_P) build/obj/$(*D) build/deps/$(*D)
	$(QUIET_CC)$(LIBTOOL) --mode=compile --tag CC $(CC_FOR_BUILD) $(DEPFLAGS) $(GAP_CFLAGS) $(GAP_CPPFLAGS) -c $< -o $(OBJFILE)
	@echo "$<:" >> $(DEPFILE)
```
- there is probably more I am forgetting, but it'll become obvious when working or this (or during code review).

However, once again, I am going to go with approach 3 for now, and already have a prototype for this working.

fingolfin:
Document how to do it, and improve support for it in the build system (we need this for Julia)

Resolves #4460

This is not yet done, the part where the release archives are updated still needs to be done. But I am out of time for now and wanted to have it here for early review.


ThomasBreuer (approved):
I had assumed that the archives of released GAP versions will contain the current versions of `src/c_oper1.c` and `src/c_type1.c`, and that these files are used in the compilation when they are present; apparently this is not the only option.
Since this is the option dealt with in this pull request (option 3 in issue #4460), I approve this approach.

fingolfin:
The wrapping failure is weird... I see this in the log:
```
...
   LINK    build/gap-nocomp
   GAC     lib/type1.g => build/c_type1.c
   GAC     lib/oper1.g => build/c_oper1.c
   C       build/c_type1.c => build/obj/build/c_type1.c.lo
   C       build/c_oper1.c => build/obj/build/c_oper1.c.lo
   LINK    gap
   LINK    libgap.la
rm build/c_type1.c build/c_oper1.c
Building HPC-GAP
...
```
So why the `rm build/c_type1.c build/c_oper1.c` ? My guess is that make does it because those are just intermediate files, and so we need to mark them as `.PRECIOUS` or `.SECONDARY`... But the bit which totally confuses me is that this doesn't seem to happen in regular builds. And if I `cd` into the tmp directory created by `make_archives.py` and type `make -j8`, it generates the removed files again, and this time, it does not delete them. What might cause this change in behaviour, I wonder?!?

fingolfin:
Ready for review now.

Before this is merged, we should test the release tarballs it produces, to verify that they work as intended (and contain `src/c_oper1.c` etc.). Unfortunately I think the only way to get a release tarball out of this is if it fails. But one can run the make_archive.py script locally / manually

wilfwilson:
> So why the `rm build/c_type1.c build/c_oper1.c` ? My guess is that make does it because those are just intermediate files, and so we need to mark them as `.PRECIOUS` or `.SECONDARY`... But the bit which totally confuses me is that this doesn't seem to happen in regular builds. And if I `cd` into the tmp directory created by `make_archives.py` and type `make -j8`, it generates the removed files again, and this time, it does not delete them. What might cause this change in behaviour, I wonder?!?

This sounds like what I reported in #4467 - does this therefore now resolve that issue?

README.buildsys.md
```diff
@@ -180,6 +180,55 @@ is searched first for files, and only if no matching file is found there does
...
+## Cross compilation
+
+The GAP build system supports cross compilation by leveraging the GNU autoconf
+support for this (please consult the GNU autoconf manual for details).
+When building GAP from a release archive, everything should just work.
+
+However, for development versions of GAP built directly from its git sources,
+there is a complication: GAP uses a few C source files which are the output of
+the GAP-to-C compiler `gac`, which in turn needs a working `gap` executable.
+This requires a bootstrapping process, which is works because GAP can actually
```
wilfwilson, line 192:
```suggestion
This requires a bootstrapping process, which works because GAP can actually
```

fingolfin:
> This sounds like what I reported in #4467 - does this therefore now resolve that issue?

Indeed it does, I hadn't seen that issue when I made this PR :-)

README.buildsys.md
fingolfin: Thanks

Issue #4197, opened 2020-12-02
Reference manual should document the long forms of command line options


fingolfin:
E.g. on https://www.gap-system.org/Manuals/doc/ref/chap3.html#X782751D5858A6EAF we see `-h` but not `--help`.

PR #4465, opened 2021-05-06
added the command line option `--print-gaproot`

### LINKED ISSUE #4448: Add command line option for querying "the" (primary) GAP root directory
It would be good if there was a command line option like `--get-gaproot` or `--sysinfopath` or `--get-primary-gaproot` or `-print-gaproot` or ... which prints the path of the primary GAP root, i.e., the one which contains `sysinfo.gap`, and which one needs to pass via `--with-gaproot=BLA` to `configure` scripts of GAP packages.

This would have multiple applications
- the mentioned `configure` scripts could have a better default behaviour that using `../..` as GAPROOT...
   - they could see if there is `gap` executable in the PATH and if so, ask it for the PATH
   - they could have a `--with-gap=PATH` option which takes the name and/or path of a GAP executable and query it for the (primary) GAP root directory
- the CI scripts for packages could be simplified in some cases

It should be fairly easy to implement this; in GAP,  one can do it via `ReplacedString(Filename(DirectoriesLibrary(""),"sysinfo.gap"),"/sysinfo.gap","");` (although this code is not *quite* correct, it really should strip the `sysinfo.gap` from the end, not replace the first occurrence...).

Of course at the the time we handle command line arguments, these functions are available, but I am sure it can be done (if need be at the C/kernel level). But I guess it suffices to scan `GAPInfo.RootPaths` resp. `GAPInfo.KernelInfo.GAP_ROOT_PATHS`, and return the first of these directory paths which contain `sysinfo.gap`

Really, the hardest part seems to settle on a good name for the command line option :-P.

It would be great to add it before 4.12 so that we can start using it sooner rather than later.

ThomasBreuer:
Note that the list of command line options in the GAP Reference Manual (Section "Command Line Options") does not contain several of the options that have been added recently.
More precisely, this list does not mention options with "long form" that start with `--`, and does not list any of these options.
Should we extend the list in the manual, or should we perhaps better refer to the list printed by the `-h` (or `--help`) option?

resolves #4448

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -55,6 +55,7 @@ BIND_GLOBAL( "GAPInfo", rec(
     CommandLineOptionData := [
       rec( short:= "h", long := "help", default := false, help := ["print this help and exit"] ),
       rec( long := "version", default := false, help := ["print the GAP version and exit"] ),
+      rec( long := "print-gaproot", default := false, help := ["print the primary GAP root and exit"] ),
       rec( short:= "b", long := "banner", default := false, help := ["disable/enable the banner"] ),
       rec( short:= "q", long := "quiet", default := false, help := ["enable/disable quiet mode"] ),
       rec( short:= "e", default := false, help := ["disable/enable quitting on <ctrl>-D"] ),
@@ -483,6 +484,19 @@ CallAndInstallPostRestore( function()
        "  Boolean options toggle the current value each time they are called.\n",
        "  Default actions are indicated first.\n",
        "\n" );
+      QuitGap();
+    fi;
+
+    # Evaluate the '--print-gaproot' option.
+    if GAPInfo.CommandLineOptions.( "print-gaproot" ) then
+      for line in GAPInfo.RootPaths do
+        value:= SHALLOW_COPY_OBJ( line );
+        APPEND_LIST_INTR( value, "sysinfo.gap" );
+        if IsExistingFile( value ) then
+          PRINT_TO( "*errout*", line );
+          break;
+        fi;
+      od;
       QuitGap();
     fi;
 end );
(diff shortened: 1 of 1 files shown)


wilfwilson:
The issue of documenting the long form command line options was raised in #4197.

lib/system.g
```diff
@@ -485,6 +486,19 @@ CallAndInstallPostRestore( function()
...
       QuitGap();
     fi;
+
+    # Evaluate the '--print-gaproot' option.
+    if GAPInfo.CommandLineOptions.( "print-gaproot" ) then
+      for line in GAPInfo.RootPaths do
+        value:= SHALLOW_COPY_OBJ( line );
+        APPEND_LIST_INTR( value, "sysinfo.gap" );
+        if IsExistingFile( value ) then
+          PRINT_TO( "*errout*", line );
```
fingolfin, line 496:
```suggestion
          PRINT_TO( "*stdout*", line );
```

fingolfin (approved):
Looks good to me except I'd print the result to stdout, not stderr, as it is expected output, not an error.

The test failure is not caused by this PR and hopefully PR #4468 will fix it.

lib/system.g
fingolfin: I would print the --help output to stdout, too, yes. I didn't (consciously) realize that we print it to stderr.

However, I don't care as much in this case: `--print-gaproot` is exclusively meant for use by scripts and so having it print to stderr would be rather inconvenient.

In contrast, I don't have to process the output of `--help` from a script, so I care less. (I am not claiming nobody has need for that, just that I don't -- and before I'd parse that, I'd rather write code which prints the information in there in a machine readable format. But that's another story).

(Actually, as I write this, I realize that I might want to parse the output `gap --help` after all: to efficiently figure out if `--print-gaproot` is supported 😂 . But in practice, I think I'd rather use an invocation like `gap --print-gaproot -q -b -A -r path/to/fallback/code.g`

fingolfin: tl;dr: sure, let's change --help to print to stdout, too, but perhaps in another PR?

wilfwilson (approved):
This works for me, and the `--help` text appears nicely.

Issue #2431, opened 2018-05-01
Undocumented items in reference manual


olexandr-konovalov:
Tried to put cross-references into release overview for GAP 4.9.1 and they did not work:
```
#W WARNING: non resolved reference: rec(
  Attr := "DirectFactorsOfGroup",
  BookName := "ref" )
#W WARNING: non resolved reference: rec(
  Attr := "SemidirectDecompositions",
  BookName := "ref" )
#W WARNING: non resolved reference: rec(
  Attr := "NormalHallSubgroups",
  BookName := "ref" )
#W WARNING: non resolved reference: rec(
  BookName := "ref",
  Oper := "RootFFE" )
#W WARNING: non resolved reference: rec(
  BookName := "ref",
  Func := "IdOfFilter" )
#W WARNING: non resolved reference: rec(
  BookName := "ref",
  Func := "IdOfFilterByName" )
#I  File: manual.lab written.
```
Is there no documentation intentionally, or it's there but not hooked into the manual?

- [ ] DirectFactorsOfGroup
- [ ] SemidirectDecompositions
- [ ] NormalHallSubgroups
- [ ] RootFFE
- [ ] IdOfFilter
- [ ] IdOfFilterByName

hulpke:
I think the first three were introduced in #379 and #591

RootFFE was never really documented. In GAP 3 it was purely lookup of the Zech-logarithm, but now it does some calculations if the representation is not Zech. We probably should define what we consider as basic and as plausible operations for FFEs.

`IdOfFilter` (which sounds like Freudian analysis) is defined in filter.gi

alex-konovalov:
@hulpke thanks. `RootFFE` is also a subject of #2131.

alex-konovalov:
Avoided unresolved links and implemented two most recent @fingolfin's suggestions - should be ready for merge.

Issue #2131, opened 2018-01-30
Document RootFFE? Merge it with `NthRoot`? What about `Root`?


fingolfin:
@hulpke added a new operation `LowIndexSubgroups` in PR #683, but it is currently undocumented (well, there is a documentation comment, but it is not hooked up into the the reference manual).

I wonder if this is intentional, or whether perhaps `LowIndexSubgroups` should be documented (and then also on the `stable-4.9` branch)?

fingolfin:
I actually wonder the same about `RootFFE`, which was already added in gap-4.8 (and later revised in GAP 4.9)

alex-konovalov:
Just to remind about #855 and #1796 discussing how to find documentation which is not linked into Manuals

fingolfin:
@alex-konovalov I mention these two function specifically because `RootFFE` is mentioned in the draft of the GAP 4.9 release notes -- but if it stays undocumented, it should be removed there! For else we'd be suggesting to users that `RootFFE` is safe to use in their own code, when it isn't. (That said, I'd greatly prefer for it to become documented!)

Somewhat similar for `LowIndexSubgroups`: I wanted to improve the description for PR #683 in the release notes, and wanted to mention `LowIndexSubgroups` in there, but then realized it is undocumented...

fingolfin:
By chance, I just learned that `RootFFE` actually duplicates the functionality of `NthRoot`, for which we install a method in `lib/ffe.gi`, albeit one less efficient (I think) than what `RootFFE` does. 

Of course there is then also an operation `Root` -- but it is "broken" in that it only accepts a multiplicative element and (optional) a positive integer n, but not a domain. I guess one could fix that, and install methods for it that mirror `NthRoot` etc. ... hohum. 

Of course `Root` is also undocumented, in particular its semantics are unclear. Since it is declared right next to `Sqrt`, one might expect that `Sqrt(x) = Root(x, 2)`, but this is not the case, as the only method the GAP library installs for `Root` is `RootInt`, which returns the integral part of a root... In particular:
```
gap> Sqrt(5);
E(5)-E(5)^2-E(5)^3+E(5)^4
gap> Root(5,2);
2
gap> Sqrt(-1);
E(4)
gap> Root(-1,2);
Error, Root: <n> is negative but <k> is even in
...
```

*sigh*

fingolfin:
Let's see...

- `NthRoot`
  - used in the library: no
  - packages using it: `forms`
  - documented: no
[...]

PR #4471, opened 2021-05-07
Document RootFFE

AnnaKDS:
I came across the function `RootFFE` and noticed that it wasn't properly documented. 

This PR fixes a mistake in the documentation and adds it to the manual.

### DIFF
--- a/lib/ffe.gd
+++ b/lib/ffe.gd
@@ -668,7 +668,7 @@ DeclareAttribute( "AsInternalFFE", IsFFE);
 ##
 ##  <Description>
 ##  <Ref Func="RootFFE"/> returns a finite field element 
-##  <A>r</A> from <F> whose <A>k</A>-th power is <A>z</A>.
+##  <A>r</A> from <A>F</A> whose <A>k</A>-th power is <A>z</A>.
 ##  If no such element exists
 ##  then
 ##  <K>fail</K> is returned.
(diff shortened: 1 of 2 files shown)


fingolfin:
Have you seen issues #2131 and #2431 ?

lib/ffe.gd
```diff
@@ -668,7 +668,7 @@ DeclareAttribute( "AsInternalFFE", IsFFE);
 ##
 ##  <Description>
 ##  <Ref Func="RootFFE"/> returns a finite field element 
-##  <A>r</A> from <F> whose <A>k</A>-th power is <A>z</A>.
+##  <A>r</A> from <A>F</A> whose <A>k</A>-th power is <A>z</A>.
```
fingolfin, line 671: The documentation for `RootFFE` is, however, incomplete:  `F` is optional, and one can also instead pass in the size of the finite field (i.e., a prime power). Whether we want to document that or not is another question....?

fingolfin (commented):
Thank you. We may need to discuss a little bit, though, see comment.

Ping @hulpke

lib/ffe.gd
hulpke: I think the initial use was for the size of the field and we then realized that we need to allow different bases. I think both uses have their place, and I thus would suggest we document them both.

hulpke (approved):

fingolfin (approved):

PR #4472, opened 2021-05-07
Fix XML errors in main manuals.

frankluebeck:
These problems were detected with a new function
'ValidateGAPDoc' in the GAPDoc package.

### DIFF
--- a/doc/dev/main.xml
+++ b/doc/dev/main.xml
@@ -2,11 +2,11 @@
 <?LaTeX ExtraPreamble="\usepackage[pdftex]{graphicx}"?>
 
 <!DOCTYPE Book SYSTEM "gapdoc.dtd"
-[<!ENTITY C "<Package>C</Package>">]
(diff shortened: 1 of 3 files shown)


doc/dev/main.xml
```diff
@@ -36,8 +35,7 @@ among the developers before they are committed, called <E>proposals</E>;
 there is a concept of <E>modules</E> where  parts of the
 &GAP; system are assigned to specified maintainers.
 </Abstract>
-
-
+    <Copyright> 2005-2015, The &GAP; Group</Copyright>
```
fingolfin, line 38: Should we maybe also change that to 2021?

fingolfin (approved):
Looks good to me, thanks!

doc/dev/main.xml
frankluebeck: I have changed it such that it  is similar to the other manuals (using a `&RELEASEYEAR;` entity).

wilfwilson (approved):

PR #4473, opened 2021-05-07
Refactor CI Workflow, unifying with Cygwin CI job

wilfwilson:
Hopefully it is clear from the diff what I have done.

The primary change is that the separate Cygwin/Windows job is now unified with the main Unix jobs. This has the additional benefit that code coverage is now captured for the Cygwin job. This was made possible through the use of the setting:
```
defaults:
  run:
    shell:
```
The Unix jobs are set to always use `bash` (which is their default, anyway), but we want the Windows job to use a special `bash` (the Windows default is `cmd`).

I would be nice if the workflow files were 'cleverer', so that we could program the Linux/Mac jobs to always use `bash`, without having to repeat `shell: bash` all over the place, but I don't see how this is currently possible.

I also noticed that I could put more jobs into the 'matrix' directly, which saves some code duplication. To do so, I have:
* Changed the `ubuntu-latest` job from being the `testmanuals` to being the `testbugfix` job, which I think is harmless.
* Moved the lists of required additional packages (for the `testpackages` and the `makemanuals` jobs) into the `Install dependencies` step. I think this is more natural anyway, and again, it reduces code duplication.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,27 +15,54 @@ env:
   LDFLAGS: "--coverage"
   # default config flags: enable debug asserts
   CONFIGFLAGS: "--enable-debug"
+  # CHERE_INVOKING=1 lets us start a 'login shell' (to set paths) in Windows without changing directory
+  CHERE_INVOKING: 1
 
 jobs:
-  test-unix:
+  test:
     name: ${{ matrix.test-suites }} - ${{ matrix.extra }} - ${{ matrix.os }}
     # Don't run this twice on PRs for branches pushed to the same repository
     if: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) }}
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
     strategy:
       fail-fast: false
       matrix:
-        # base test: fast first test
         os: [ubuntu-18.04]
-        test-suites: ["testinstall"]
+        shell: [bash]
+        test-suites:
+          [
+            # base test: fast first test
+            "testinstall",
+
+            "teststandard",
+
+            # run tests contained in the manual
+            "testmanuals",
+
+            # test error reporting and compiling as well as libgap
+            "testspecial test-compile testlibgap testkernel",
+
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -178,6 +165,32 @@ jobs:
...
+                         libfplll-dev
+                         libmpc-dev
+                         libmpfi-dev
+                         libmpfr-dev
+                         libncurses5-dev
+                         libzmq3-dev
+                         pari-gp
+                         singular
+                       )
+                   fi
```
fingolfin, line 183: Maybe we should move this lengthy step into a `dev/ci-install-deps.sh` script?

wilfwilson: Perhaps. We could also just put it on one line to reduce the number of lines taken.

I worry that putting this stuff in a separate script would make the workflow a bit more opaque. But since we already use several scripts that's probably not important.

For me the ideal thing would be to know why each of these packages is here, or at least, which package(s) require each. 4ti2 and singular are obvious, some of the others less so 🙂,

fingolfin:
Typo in commit message: Refector -> Refactor (yes, a major issue 😉 )

.github/workflows/CI.yml
```diff
@@ -178,6 +165,32 @@ jobs:
...
+                         libcdd-dev
+                         libcurl4-openssl-dev
+                         libfplll-dev
+                         libmpc-dev
+                         libmpfi-dev
+                         libmpfr-dev
+                         libncurses5-dev
+                         libzmq3-dev
+                         pari-gp
+                         singular
```
fingolfin, line 181: Since you asked for it:
```suggestion
                         4ti2                   # 4ti2Interface
                         libboost-dev           # NormalizInterface
                         libcdd-dev             # CddInterface
                         libcurl4-openssl-dev   # curlInterface
                         libfplll-dev           # float
                         libmpc-dev             # float
                         libmpfi-dev            # float
                         libmpfr-dev            # float
                         libncurses5-dev        # browse
                         libzmq3-dev            # ZeroMQInterface
                         pari-gp                # alnuth
                         singular               # singular
```

fingolfin (approved):

Issue #1385, opened 2017-06-05
`FreeMagma` has unhelpful error messages, inconsistent argument checks, and can create incorrect 0-generator objects


markuspf:
First off this is a bug:
```
gap> FreeSemigroup();
Error, List Element: <list>[1] must have an assigned value in
  if IsFilter( arg[1] )  then
    lesy := arg[1];
    zarg := arg{[ 2 .. Length( arg ) ]};
else
    zarg := arg;
```

```
gap> F := FreeSemigroup([]);
<free semigroup on the generators [  ]>
gap> Size(F);
infinity
gap> e := Enumerator(F);
<enumerator of <free semigroup on the generators [  ]>>
gap> e[2];
^2
```

I would also argue that this is a bug (or at least an extremely unhelpful error message yet again)
```
gap> FreeMagma(0);
Error, usage: FreeMagma(<name1>,<name2>..),FreeMagma(<rank>) at /home/makx/dev/gap/lib/mgmfree.gi:316 called from
<function "FreeMagma">( <arguments> )
 called from read-eval loop at *stdin*:7
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>
```

Though it is documented that the rank has to be positive, but why should there not be a `FreeSemigroup(0)` (which would be empty)? 

I can fix most of this, just leaving it here for me to remember. (of course I welcome people beating me to it)

wilfwilson:
This has been resolved for `FreeSemigroup`s in #4372.

I will work on a similar fix for `FreeMagma` (and `FreeMonoid` and `FreeMagmaWithOne`, if they have similar problems). 

I have added a feature request in #4424 to see about properly supporting these objects with zero generators.

PR #4475, opened 2021-05-07
Improve argument checking and documentation for `FreeGroup`, `FreeMonoid`, `FreeSemigroup`, `FreeMagmaWithOne`, and `FreeMagma`

wilfwilson:
This PR introduces a function `FreeXArgumentProcessor` to unify the argument processing for the following functions:
* `FreeSemigroup`
* `FreeMonoid`
* `FreeMagma`
* `FreeMagmaWithOne`
* `FreeGroup`

This resolves the remaining problems raised in #1385. This also builds on #4111.

### DIFF
--- a/doc/ref/grplib.xml
+++ b/doc/ref/grplib.xml
@@ -101,12 +101,14 @@ calculations.
 
 <Subsection Label="Generator Names">
 <Heading>Generator Names</Heading>
-For groups created as finitely presented groups, including polycyclic groups, the
-generators are labelled, by default, with a letter and a number. Using the
-option <C>generatorNames</C> it is possible to influcence this naming. If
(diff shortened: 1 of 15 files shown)


wilfwilson:
Whoops, this turned into a very big PR. I hope that it is not too difficult to review.

* Many of the added lines are in test files/newly added manual examples, so don't need to be looked at too closely.
* The different pieces of documentation are very similar, so there's not much point looking at more than one in any detail.

Changes in functionality:
* You can't now use an empty string as a label of a generator. I just thought this was too silly.
* All of the functions in question now support the `generatorNames` option that @hulpke added in #4111.
* Hopefully all the known bugs related to this stuff are now fixed.

hulpke (approved):

Issue #4480, opened 2021-05-10
Fix associativity problems in domain constructions.


ThomasBreuer:
This is a follow-up of pull request  #4373, which itself fixes #4030.

Currently there are quite a few situations where GAP creates domains that claim to be associative although associativity does not hold for all triples of elements.
```
gap> T:= [ [ 1, 2, 6, 5, 4, 3 ], [ 2, 1, 4, 2, 2, 5 ], [ 6, 5, 1, 6, 6, 4 ],
>          [ 5, 6, 3, 4, 1, 2 ], [ 4, 3, 2, 1, 5, 6 ], [ 3, 4, 5, 3, 3, 1 ] ];;
gap> M:= MagmaByMultiplicationTable( T );
<magma with 6 generators>
gap> SemigroupByGenerators( M );
<semigroup with 6 generators>
gap> MonoidByGenerators( M );
<monoid with 6 generators>
gap> GroupByGenerators( M );
<group with 6 generators>
gap> Group( Elements( M ) );
<group with 6 generators>
```
It is not clear to me where the associativity checks should be inserted.
Perhaps the real question is about the documentation:
For group constructions, we have the plain function `Group` and the operations `GroupByGenerators` and `GroupWithGenerators`. The function `Group` calls `IsGeneratorsOfMagmaWithInverses` in order to check whether the given generators are allowed, and then calls `GroupByGenerators`. The two operations do not perform such checks.
(The latter is crucial for example in situations where functions such as `GL` construct groups from huge matrices, which one does not want to check for invertibility. Note that we do not have something such as `GroupNC` or `GroupByGeneratorsNC`.)
I think the associativity checks should happen where also the invertibility is checked, and the whole setup should become documented --thus now is the last chance to revise something.

wilfwilson:
Thank you for raising this issue.

I'll note that the manual entries for the functions `Semigroup` and `Monoid` explicitly state that they don't test for associativity:
> It is not checked whether the underlying multiplication is associative, use `Magma` and `IsAssociative` if you want to check whether a magma is in fact a semigroup.

> It is not checked whether the underlying multiplication is associative, use
`MagmaWithOne` and `IsAssociative` if you want to check whether a magma-with-one is in fact a monoid.

Matthias-Fresacher:
[...]

PR #4484, opened 2021-05-11
`?Group` shows first info about the library function

ThomasBreuer:
Adding an explicit `Heading` to the manual section for `Group` has the effect that `?Group` lists this section more prominently.
Note that the sections for `Monoid` and `Semigroup` had already an explicit `Heading`, and the three functions should behave similarly, see issue #4480.


wilfwilson (approved):
This is a good idea. I didn't know it was possible to change the behaviour this way.

I have often typed `?Group` and been annoyed that the function `Group` was so far down the list. This does the job nicely.

fingolfin (approved):

Issue #500, opened 2016-01-19
Omega(n,q) is not always a subgroup of SO(n,q)


dimpase:
E.g.

```
gap> G:=SO(3,9);
SO(0,3,9)
gap> G1:=Omega(3,9);
Omega(0,3,9)
gap> G1.1 in G;
false
```

This contradicts documentation (50.2-8), which says "This group has always index two in SO( e, d, q )". 
(So it stabilises a different form)...

dimpase:
Specifically, in this case, in coordinates X,Y,Z, the group SO preserves the quadratic form XY-Z^2, and Omega preserves the form XZ-Y^2.

stevelinton:
Section 50.2 of the documentation explains where all the generators come from and specifically says "Note  that  due  to  the different sources for the generators, the invariant
  forms  for the groups ohm(e,d,q) are in general different from the forms for
  SO(e,d,q) and GO(e,d,q)."

So the fix is probably just to amend 50.2-8.

fingolfin:
It would be kind of nice if they used the same form, though. Achieving that shouldn't be very hard. I can look into it (maybe not today, but...).

fingolfin:
The only concern is: Might there be existing code which breaks if we change which form is used?

dimpase:
one can make a new function, called SOmega, say, and put a new
implementation there. And deprecate Omega.

frankluebeck:
It is maybe good to have generators which can be found in the literature available in GAP. 

If someone wants to look at this: It would be nice to store the bilinear/quadratic form in the group returned by `Omega` similar to what is stored in the group by `SO` or `Sp`.

Could one use the `Forms` package to provide a variant of `SO`, `Omega`, `Sp` and so on where the form is given as an argument?

dimpase:
Forms package is broken in characteristic 2, at least. Try using it on SO(3,2)... (This bug has been discovered today by me and reported to the author...)

fingolfin:
Using the forms package to provide such variants is actually quite easy (modulo bugs, such as the one reported by Dima, of course). Here is an example:
```
gap> G:=Omega(3,3);
Omega(0,3,3)
gap> formOmega := PreservedSesquilinearForms(G)[1];
< bilinear form >
gap> Display(formOmega);
Bilinear form
Gram Matrix:
 . . 1
 . 2 .
 1 . .
gap> H:=GO(3,3);
GO(0,3,3)
[...]

PR #4489, opened 2021-05-12
admit specifying a form for classical groups

ThomasBreuer:
## Text for release notes 
The documentation and the code of the functions `GO`, `GU`, `Omega`, `SO`, `SU`, and `Sp` have been extended such that one can specify an invariant form for the group in question. The methods that do the work will be provided via the Forms package.

See gap-packages/forms/pull/4 for the proposed corresponding changes in the Forms package.
The pull request is intended to resolve issue #500.

### DIFF
--- a/grp/classic.gd
+++ b/grp/classic.gd
@@ -224,6 +224,13 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
 ##  or a form object in <Ref Filt="IsQuadraticForm" BookName="Forms"/>
 ##  or a group with stored <Ref Attr="InvariantQuadraticForm"/> value
 ##  (and then this form is taken).
+##  <P/>
+##  A given <A>form</A> determines <A>e</A> and <A>d</A>, and also <A>q</A>
+##  except if <A>form</A> is a matrix that does not store its
+##  <Ref Attr="BaseDomain" Label="for a matrix object"/> value.
+##  These parameters can be entered, and an error is signalled if they do
+##  not fit to the given <A>form</A>.
+##  <P/>
 ##  If <A>form</A> is not given then a default is chosen as described in the
 ##  introduction to Section <Ref Sect="Classical Groups"/>.
 ##  <P/>
@@ -334,6 +341,13 @@ DeclareConstructor( "GeneralUnitaryGroupCons",
 ##  or a form object in <Ref Filt="IsHermitianForm" BookName="Forms"/>
 ##  or a group with stored <Ref Attr="InvariantSesquilinearForm"/> value
 ##  (and then this form is taken).
+##  <P/>
+##  A given <A>form</A> determines <A>d</A>, and also <A>q</A>
+##  except if <A>form</A> is a matrix that does not store its
+##  <Ref Attr="BaseDomain" Label="for a matrix object"/> value.
+##  These parameters can be entered, and an error is signalled if they do
+##  not fit to the given <A>form</A>.
+##  <P/>
 ##  If <A>form</A> is not given then a default is chosen as described in the
 ##  introduction to Section <Ref Sect="Classical Groups"/>.
 ##  <P/>
@@ -521,6 +535,13 @@ DeclareConstructor( "SpecialOrthogonalGroupCons",
 ##  or a form object in <Ref Filt="IsQuadraticForm" BookName="Forms"/>
 ##  or a group with stored <Ref Attr="InvariantQuadraticForm"/> value
 ##  (and then this form is taken).
+##  <P/>
+##  A given <A>form</A> determines <A>e</A> and <A>d</A>, and also <A>q</A>
+##  except if <A>form</A> is a matrix that does not store its
+##  <Ref Attr="BaseDomain" Label="for a matrix object"/> value.
+##  These parameters can be entered, and an error is signalled if they do
+##  not fit to the given <A>form</A>.
+##  <P/>
 ##  If <A>form</A> is not given then a default is chosen as described in the
 ##  introduction to Section <Ref Sect="Classical Groups"/>.
 ##  <P/>
@@ -620,6 +641,13 @@ DeclareConstructor( "SpecialUnitaryGroupCons",
 ##  or a form object in <Ref Filt="IsHermitianForm" BookName="Forms"/>
 ##  or a group with stored <Ref Attr="InvariantSesquilinearForm"/> value
 ##  (and then this form is taken).
+##  <P/>
+##  A given <A>form</A> determines <A>d</A>, and also <A>q</A>
+##  except if <A>form</A> is a matrix that does not store its
+##  <Ref Attr="BaseDomain" Label="for a matrix object"/> value.
+##  These parameters can be entered, and an error is signalled if they do
+##  not fit to the given <A>form</A>.
+##  <P/>
 ##  If <A>form</A> is not given then a default is chosen as described in the
 ##  introduction to Section <Ref Sect="Classical Groups"/>.
 ##  <P/>
@@ -730,6 +758,13 @@ DeclareConstructor( "SymplecticGroupCons", [ IsGroup, IsPosInt, IsRing ] );
 ##  or a form object in <Ref Filt="IsBilinearForm" BookName="Forms"/>
 ##  or a group with stored <Ref Attr="InvariantBilinearForm"/> value
 ##  (and then this form is taken).
+##  <P/>
+##  A given <A>form</A> determines and <A>d</A>, and also <A>q</A>
+##  except if <A>form</A> is a matrix that does not store its
+##  <Ref Attr="BaseDomain" Label="for a matrix object"/> value.
+##  These parameters can be entered, and an error is signalled if they do
+##  not fit to the given <A>form</A>.
+##  <P/>
 ##  If <A>form</A> is not given then a default is chosen as described in the
 ##  introduction to Section <Ref Sect="Classical Groups"/>.
 ##  <P/>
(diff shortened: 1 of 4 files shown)


grp/classic.gd
```diff
@@ -227,19 +236,59 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
...
+    return GeneralOrthogonalGroupCons( IsMatrixGroup, arg[1] );
+  elif Length( arg ) = 2 and IsInt( arg[1] ) then
+    # (d, q) or (d, R)
     return GeneralOrthogonalGroupCons( IsMatrixGroup, 0, arg[1], arg[2] );
   elif Length( arg ) = 3 and IsInt(arg[1]) and IsInt(arg[2]) and
     (IsInt(arg[3]) or IsRing(arg[3])) then
+    # (e, d, q) or (e, d, R)
     return GeneralOrthogonalGroupCons( IsMatrixGroup,arg[1],arg[2],arg[3] );
+  elif Length( arg ) = 3 and IsInt(arg[1]) and
+    (IsInt(arg[2]) or IsRing(arg[2])) and
```
fingolfin, line 250: Actually, it seems the possibility to pass in a ring is undocumented for GO, SO, etc. (however, it is documented for GL and SL). That seems like an oversight to me, perhaps we should also fix that (not necessarily in this PR, of course).

grp/classic.gd
```diff
@@ -193,13 +192,17 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
 
 #############################################################################
 ##
-#F  GeneralOrthogonalGroup( [<filt>, ][<e>, ]<d>, <q> ) .  gen. orthog. group
-#F  GO( [<filt>, ][<e>, ]<d>, <q> )
+#F  GeneralOrthogonalGroup( [<filt>, ][<e>, ]<d>, <q>[, <form>] )
```
fingolfin, line 195: But doesn't the form already imply the value for `d` and `e`?  If we accept all three, how are they supposed to interact / be sanity checked?

I guess only the base field would perhaps we in need of overriding (i.e., it seems reasonably to say: "use this form given over GF(2) but create a group over GF(8)").

grp/classic.gd
```diff
@@ -227,19 +236,59 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
 ##  <#/GAPDoc>
 ##
 BindGlobal( "GeneralOrthogonalGroup", function ( arg )
```
fingolfin, line 238: I think the code below could be simplified quite a bit (roughly halving the code?) by doing something like this:
```suggestion
BindGlobal( "GeneralOrthogonalGroup", function ( arg )
  local filter;
  if IsFilter(arg[1]) then
    filter := Remove(arg, 1);
  else
    filter := IsMatrixGroup;
  fi;
```
And then using `filter` in the obvious way.

One could also do something like this (in addition):
```gap
  if IS_FORM_OR_GROUP_WITH_FORM(Last(arg)) then
     form := Remove(arg);
     # now handle the cases with form -- as I state above, I would assume at this
     # point either `arg` is empty, or else contains a ring or the size of a finite field
     ...
  fi;

**Review by fingolfin — CHANGES_REQUESTED** [review]
Thanks for working on this, I like the direction, but have some suggestions / nitpicks

File: grp/classic.gd (hunk 1 above)
Comment (ThomasBreuer) [reply grp/classic.gd:250]: I am not sure whether this was an oversight.
For `GL`, `SL`, and `Sp`, we have also methods for constructing the group in question over a residue class ring of the integers, but finite fields are the only supported rings for the orthogonal groups.

It would not be a problem to admit also `GO( 3, GF(5) )`, etc.; the documentation should then state that currently only finite fields are supported, analogous to `GL`, `SL`, and `Sp`.

File: grp/classic.gd (hunk 2 above)
Comment (ThomasBreuer) [reply grp/classic.gd:195]: My idea was the following:
One can write `GO( 3, 5, form )` in order to express that one wants the group in question, with the given form as an additional requirement. One can also write `GO( form )` (which is also a documented variant), if one wants to express that the given form determines the other parameters.
In code examples, I would probably prefer the former variant.

File: grp/classic.gd (hunk 3 above)
Comment (ThomasBreuer) [reply grp/classic.gd:238]: Yes.
(One could have done this already before the changes proposed here, but it was less striking.)

File: grp/classic.gd [hunk 4]
```diff
@@ -227,19 +236,59 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
...
     (IsInt(arg[3]) or IsRing(arg[3])) then
+    # (e, d, q) or (e, d, R)
     return GeneralOrthogonalGroupCons( IsMatrixGroup,arg[1],arg[2],arg[3] );
+  elif Length( arg ) = 3 and IsInt(arg[1]) and
+    (IsInt(arg[2]) or IsRing(arg[2])) and
+    (IsMatrixObj(arg[3]) or
+     (IsBoundGlobal( "IsQuadraticForm" ) and ValueGlobal( "IsQuadraticForm" )(arg[3])) or
+     HasInvariantQuadraticForm(arg[3])) then
+    # (d, q, form) or (d, R, form)
+    return GeneralOrthogonalGroupCons( IsMatrixGroup,0,arg[1],arg[2],arg[3] );
```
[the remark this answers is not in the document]
ThomasBreuer: Well, what should be the rule? The use of whitespace was not consistent before the changes proposed here, and it is not consistent afterwards. Should one really care about it?
(For the code line in question, one could argue that it would exceed 80 characters with the additional whitespace. For other lines, this argument would not apply.)

grp/classic.gd
```diff
@@ -227,19 +236,59 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
...
     return GeneralOrthogonalGroupCons( IsMatrixGroup, 0, arg[1], arg[2] );
   elif Length( arg ) = 3 and IsInt(arg[1]) and IsInt(arg[2]) and
     (IsInt(arg[3]) or IsRing(arg[3])) then
+    # (e, d, q) or (e, d, R)
     return GeneralOrthogonalGroupCons( IsMatrixGroup,arg[1],arg[2],arg[3] );
+  elif Length( arg ) = 3 and IsInt(arg[1]) and
+    (IsInt(arg[2]) or IsRing(arg[2])) and
+    (IsMatrixObj(arg[3]) or
+     (IsBoundGlobal( "IsQuadraticForm" ) and ValueGlobal( "IsQuadraticForm" )(arg[3])) or
+     HasInvariantQuadraticForm(arg[3])) then
```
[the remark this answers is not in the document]
ThomasBreuer: This is a good idea.

grp/classic.gd
```diff
@@ -227,19 +236,59 @@ DeclareConstructor( "GeneralOrthogonalGroupCons",
 ##  <#/GAPDoc>
 ##
 BindGlobal( "GeneralOrthogonalGroup", function ( arg )
-  if   Length( arg ) = 2 then
+  if   Length( arg ) = 1 then
+    # form (matrix, form object, or group with stored form)
+    return GeneralOrthogonalGroupCons( IsMatrixGroup, arg[1] );
+  elif Length( arg ) = 2 and IsInt( arg[1] ) then
```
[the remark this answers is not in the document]
ThomasBreuer: My aim is to add new variants of arguments. For that, I have to perform enough tests to distinguish the possible cases, and I have tried not to introduce more tests. (In the original code for the two-argument case, none of the argument types is checked.)
It was not my aim to clean the available code.

grp/classic.gd
fingolfin: We already admit `GO(3,GF(5))`, it just is undocumented

grp/classic.gd
fingolfin: But if the former is done, shouldn't we then also check that the `form` matches the given values for `d` and `q`? Or do we expect all installed methods to re-implement that check?

ThomasBreuer: The GAP library does not know about form objects, and the code dealing with this argument belongs to the Forms package.
We could check some conditions about a matrix or a group with stored form in the global functions in the GAP library, but only the code in the Forms package can decide whether the given argument makes sense.

grp/classic.gd
ThomasBreuer: Perhaps the word "admit" was not a good choice.
I wrote in a commit message:
> Concerning the possibility to enter `GF(q)` instead of `q` (or more generally a ring) as an argument:
> This was already documented and supported for `Sp`, and supported but not documented for `GO` and `SO`.
> Now it will be supported but not documented also for `Omega`.
> (For `GU` and `SU`, supporting `GF(q)` (or `GF(q^2)`?) would be asking for trouble:
> The GAP library uses `q` where the Forms package expects `GF(q^2)`.)

fingolfin (approved):
There was a new release of the forms package today (version 1.2.6). So let's move forward with this PR. As far as I understand, we can merge it now, independent of forms, right?

I'll wait with merging for (a) the CI tests which I just restarted by closing and reopening this PR, and (b) for @ThomasBreuer to comment if this PR is ready from his perspective

ThomasBreuer:
@fingolfin This pull requests depends on the new version of the Forms package in the sense that the documentation of the GAP functions in question promises that certain functionality is available if the Forms package is available.
If it is sure that the next released version of GAP will contain at least Forms 1.2.6 then we can merge the pull request.
Formally, it would be perhaps the safest to state explicitly in the abovementioned documentation that Forms >= 1.2.6 is needed for the promised functionality; shall I adjust the pull request this way?

fingolfin:
Yes, please update the PR accordingly

ThomasBreuer:
@fingolfin O.k.; I have updated the PR, the tests have passed, I will merge the PR.

PR #4493, opened 2021-05-17
Remove optional `crc` argument to `LoadDynamicModule` and `LoadStaticModule`

### LINKED ISSUE #4235: LoadDynamicModule: documentation for `crc` argument is wrong
To quote:
```
This command loads <A>filename</A> as module. If given, the CRC checksum 
<A>crc</A> must match the value of the module (see <Ref Sect="CrcFile"/>).
```
and then it goes on with this example:
```
gap> LoadDynamicModule("./test.so");
gap> CrcFile("test.so");
2906458206
gap> LoadDynamicModule("./test.so",1);
Error, <crc> mismatch (or no support for dynamic loading) called from
<function>( <arguments> ) called from read-eval-loop
Entering break read-eval-print loop ...
you can 'return;' to continue
brk> quit;
gap> LoadDynamicModule("./test.so",2906458206);
```

This suggest that the CRC of the .so file is checked. But that's wrong! Rather, the value of `crc` field in the `StructInitInfo` returned by the `Init__Dynamic` function of the module is checked.

For regular kernel extensions, this is normally always 0; the real use is for code produced by `gac`, then the CRC field is set to the CRC of the GAP source file used to generate the C code in question.

I am not even sure whether we use this `crc` test at all, though? I had a quick look and couldn't find anything that used it (or its counterpart for `LoadStaticModule`). We *do* perform CRC checks for the few files in the GAP kernel which are produced from GAP sources (`lib/oper1.g` -> `src/c_oper1.c` and `lib/type1.g` -> `src/c_type1.g`) but there the CRC check is performed by `READ_GAP_ROOT`.

In conclusion, perhaps the best way to fix the documentation is to undocumented the `crc` argument; and perhaps even remove it altogether (thus also simplifying the kernel functions `FuncLOAD_DYN` and `FuncLOAD_STAT`

fingolfin:
No caller used it, and using it correctly would have been both non-trivial,
and generally of limited use (as only the arbitrary hardcoded "crc" value in
the module's `StructInitInfo` instance is checked)

Resolves #4235


ChrisJefferson:
Is the plan to (long term) get rid of crc altogether? If so, it might be worth adding/editing the comment in modules.h, to say the crc values isn't actually used any more?

fingolfin:
The `crc` is still used for files produced by `gac`, and there we do use it, namely to detect if the source file changed since the compilation -- i.e., GAP will detect if `oper1.g` is different from what was used to generate `c_oper1.c`, and then won't use the compiled code from `c_oper1.c`.

fingolfin:
That said, it makes sense to update the comment for the `crc` field to explain just that

fingolfin:
I've now added such a change to `src/modules.h`

fingolfin:
@ChrisJefferson OK now?

ChrisJefferson (approved):

PR #4499, opened 2021-05-19
Let IsBlockMatrixRep imply IsMatrix

fingolfin:
(Note: I created this patch in late 2019 and then kinda forget about it. Right now I cleaning out old patches. I am not sure anymore this patch is the right thing to do -- I guess I had doubts back then and didn't submit it for that reason. I also forgot *why* I made it -- presumably the wrong method was called in some case(s), but unfortunately I made no note of which, nor recorded an example or test. Hrmm... Anyway, I'd like to resolve this patch: whether it gets merged or rejected; afterwards I can drop it :-) ).

This implication substantially increases the rank of IsBlockMatrixRep.

Note that a lot of code in fact only deals with IsOrdinaryMatrix and
IsBlockMatrixRep, including the constructor, which only allows creating
IsBlockMatrixRep which also are in the filter IsOrdinaryMatrix.

Presumably, one could create a block matrix in filter IsLieMatrix with
some manual intervention, but most methods for IsBlockMatrixRep
currently require IsOrdinaryMatrix for no apparent reason.

### DIFF
--- a/lib/matblock.gi
+++ b/lib/matblock.gi
@@ -43,7 +43,7 @@
 ##  \enditems
 ##
 DeclareRepresentation( "IsBlockMatrixRep",
-    IsComponentObjectRep,
+    IsComponentObjectRep and IsMatrix,
     [ "blocks", "zero", "nrb", "ncb", "rpb", "cpb" ] );
 
 


ThomasBreuer (commented):
`IsOrdinaryMatrix` implies `IsMatrix`. Thus an object in `IsBlockMatrixRep and IsOrdinaryMatrix` automatically lies in `IsMatrix`.
Since `IsBlockMatrixRep` occurs in the GAP library always in conjunction with `IsOrdinaryMatrix`, the proposed change would not change the behaviour.
Creating `IsBlockMatrixRep` objects not in `IsOrdinaryMatrix` is not allowed, according to the documentation. Thus we need not worry about complaints when someone has used `IsBlockMatrixRep` in a different way.

On the other hand, I have always tried to avoid implications from representations to other filters. The original idea was that the representation describes the internal structure of an object (the components one is allowed to use), which is independent of the mathematical meaning of the object. In principle, completely unrelated objects could use the same representation. (I am not aware of cases where this approach has been taken in practice.)

wucas:
I'm a bit surprised to learn that a representation doesn't imply the represented *thing* to be the mathematical *thing* it is. I always assumed that that's how it works. And @fingolfin comment https://github.com/gap-system/gap/pull/4519#issuecomment-852080779  suggests that other users seem to assume the same? Already from a readability point of view I wouldn't like the idea of e.g. a `BlockMatrixRep` to represent anything other than, well a block matrix which I'd assume to be a, well matrix (meaning anything that in the end implies `IsMatrix`) but maybe I miss a lot of cases and detail here.

ThomasBreuer:
@wucas Logically, the mathematical meaning of an object and the way how it is represented are independent: A matrix can be represented in different ways, and a GAP object represented by "two slots one of which is some vector" can be used to represent a lot of different mathematical objects.
Apparently one point is the (unfortunate?) naming of representations --if a representation is called `IsBlockMatrixRep` then one is tempted to conclude that all objects in this filter are block matrices, and in this situation it is natural to postulate that the implications suggested by the name hold on the level of filters.
Note that GAP's simple filters are precious in the sense that their number should be kept small; one could use the same representation filter for several kinds of objects, but apparently this opportunity has not been taken.
One could keep the old name `IsBlockMatrixRep`, define `IsBlockMatrix` as `IsBlockMatrixRep and IsMatrix`, and then use `IsBlockMatrix` in all method installations; the drawback would then be that an object represented by another implementation of block matrices would not be in `IsBlockMatrix`, which is perhaps also confusing.

PR #4500, opened 2021-05-19
Define IsGeneralizedCartanMatrix for IsMatrixObj

fingolfin:
Another old patch from Summer 2019 (rebased and updated).


ThomasBreuer (approved):
Good.

PR #4501, opened 2021-05-19
libgap: add GAP_CallFunc[0-3]Args

fingolfin:
This are purely for convenience, to avoid the need to create a C array of arguments.

@embray might be interested


embray:
This could be useful in a few places where GAP_CallFunc is used directly, though for calling an arbitrary GAP function the more general array-based interface is the most useful.  In the current gappy I found only 5 instances of `GAP_CallFuncArray` all with either 1 or 2 arguments, so this would be a minor convenience there.

ChrisJefferson (approved):

PR #4504, opened 2021-05-20
Turn `IsLexicographicallyLess` into an obsolete synonym for `\<`

fingolfin:
Because \< already compares lists lexicographically, just
more efficiently than what `IsLexicographicallyLess` did.

The one downside of this I can think of is now one could
abuse this to write `IsLexicographicallyLess(1,2)`

As far as I can tell, no package and no library code uses `IsLexicographicallyLess` after this PR. So perhaps we should mark it as obsolete, too?

### DIFF
--- a/doc/ref/lists.xml
+++ b/doc/ref/lists.xml
@@ -1658,7 +1658,6 @@ with possibly slightly different meaning for lists and non-list collections.
 <#Include Label="Flat">
 <#Include Label="Reversed">
 <#Include Label="Shuffle">
-<#Include Label="IsLexicographicallyLess">
 <#Include Label="Apply">
 <#Include Label="Perform">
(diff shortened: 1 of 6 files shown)


ThomasBreuer (approved):
Yes, `\<` is better than `IsLexicographicallyLess`. (One more advantage is that `\<` works also for non-dense lists.)
I do also second the idea to move `IsLexicographicallyLess` to the obsolete functionality, and to drop the manual entry.

ChrisJefferson (approved):

ChrisJefferson:
I also agree with making it obsolete (but this could be merged first, then that)

fingolfin:
I've just added a commit making it obsolete

wilfwilson (approved):

PR #4505, opened 2021-05-20
Fix `ShallowCopy` for `IteratorOfPartitionsSet`

fingolfin:
Found this while reviewing PR #4502.

### DIFF
--- a/lib/combinat.gi
+++ b/lib/combinat.gi
@@ -2500,7 +2500,7 @@ InstallGlobalFunction( IteratorOfPartitionsSet, function( s )
       return out;
     end;
 
-    shallowCopy := iter -> rec( next := ShallowCopy( iter!.next ) );
+    shallowCopy := iter -> rec( next := ShallowCopy( iter!.next ), s := iter!.s );
 
     isDone := iter -> iter!.next = false;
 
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):

alex-konovalov:
@fingolfin 

This worked yesterday and does not work today - perhaps related? Happens when all packages are loaded:
```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-pkg-update-master-q\
uicktest/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/standard/label/kovacs/GAP-pkg-u\
pdate-master-snapshot/tst/testinstall/combinat.tst:474
# Input is:
for s in [[], [5], [1,2,3,4], [2,5,7], ["a","b","c","d","e"], [3..13]] do
     pn1:= PartitionsSet( s );
     pn2:= List( IteratorOfPartitionsSet( s ) );
     if Length(pn1) <> Length(pn2) then
       Error( "wrong number of elements" );
     elif Set(pn1) <> Set(pn2) then
       Error( "different elements" );
     fi;
   od;
# Expected output:
# But found:
########
```

fingolfin:
Doesn't have to be in the release notes as this function is new in GAP 4.12

PR #4508, opened 2021-05-26
Fix crash in `LocationFunc` when the input has incomplete location information

### LINKED ISSUE #4507: Segfault when applying LocationFunc to a function created with SYNTAX_TREE_CODE
### Steps to reproduce

Enter the following code in a REPL:
```
LocationFunc(SYNTAX_TREE_CODE(SYNTAX_TREE(x->x)));
```

### Observed behaviour

Segfault

### Expected behaviour

Not a segfault :D

```
 ┌───────┐   GAP 4.12dev-890-gae742ec built on 2021-05-22 17:28:02+0200
 Packages:   Browse 1.8.9, GAPDoc 1.dev, IO 4.7.0dev, PrimGrp 3.4.0, SmallGrp 1.4.1, TransGrp 2.0.5
```

### Notes
1. This also occurs with GAP 4.11.1 (I assume this never worked).
2. `FilenameFunc`, `StartlineFunc`, and `EndlineFunc` do not cause a segfault.

fingolfin:
Also change it to error if the argument is not a function, and to
return fail (instead of an empty string) if the location can not
be determined.

Finally, make the underlying kernel function `GET_LOCATION_BODY` resilient
about function bodies that store no location information at all,
as produced by the code which turns syntax tree object back into
functions. This previously crashed.

Fixes #4507 

Note: we could also try to improve the syntax tree code to store "complete" location information, but I am not sure what would be appropriate, so I opted to not tackle this in this PR. But if @zickgraf is interested in having that, PRs and discussions are welcome :-)

### DIFF
--- a/lib/function.g
+++ b/lib/function.g
@@ -313,8 +313,8 @@ BIND_GLOBAL( "EndlineFunc", ENDLINE_FUNC );
 ##
 ##  <Description>
 ##  Let <A>func</A> be a function.
-##  Returns a string describing the location of <A>func</A>, or an empty
-##  string if the information cannot be found. This uses the information
+##  Returns a string describing the location of <A>func</A>, or <K>fail</K>
+##  if the information cannot be found. This uses the information
 ##  provided by <Ref Func="FilenameFunc"/> and <Ref Func="StartlineFunc"/>
 ##  <P/>
 ##  <Log><![CDATA[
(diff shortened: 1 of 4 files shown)


zickgraf:
Thanks for the fix!

> Note: we could also try to improve the syntax tree code to store "complete" location information, but I am not sure what would be appropriate, so I opted to not tackle this in this PR. But if @zickgraf is interested in having that, PRs and discussions are welcome :-)

Of course that would be nice to have, but I currently do not need it: I try to do all my plausibility checks right after calling `SYNTAX_TREE`, where I still have a reference to the original function. And once I have transformed the syntax tree, a reference to the original line would not be helpful in most cases. So basically I'm fine with any output, as long as there is no segfault :D

wilfwilson (approved):

Issue #945, opened 2016-11-09
Complete MatrixObj


fingolfin:
This issue is supposed to be the "master issue" for completing the MatrixObj code in the GAP library, which is meant to replace (or complement) our current matrix code (which represents matrices as lists of lists) with proper "matrix objects", which internally can be represented in different ways (as lists of lists in row-major *or* column-major form; as a list of `rows*columns` entries; as various kinds of sparse matrices; as compressed matrices; etc.).

I'll try to list some things that needs to be done for `MatrixObj` to be "complete". But we shouldn't necessarily add everything here, instead we can file separate issues, and also organize these into a GitHub "project", see https://github.com/gap-system/gap/projects/1

(I really have no idea whether everything should be in a separate issue, or in one big list, or... whatever, just do what seems natural. The important thing in the end is to actual code stuff anyway, not how we log the open tasks :-)

fingolfin:
Here are some things I think should be addressed. Some of these should probably be turned into full issues of their own. Others might turn out to be bad ideas. We'll see

- [ ] Finalize the MatrixObj interface (there were some minor concerns about this by @mohamed-barakat  and also myself)
  - [ ] consider writing the MatrixObj API so that it also works sanely for non-commutative base domains. E.g. `MultRowVector` should specify from which side it multiplies with scalars, and perhaps even come in both variants?
- [ ] Make it possible to create groups of MatrixObj matrices with `Group([mat1,mat2,...])`.
- [ ] Teach the `NiceMonomorphism` code for groups to support `MatrixObj`
- [ ] Teach `recog` to use `MatrixObj`
- [ ] Complete the documentation, see http://www.gap-system.org/Manuals/doc/ref/chap26.html
- [ ] go through all (well, at least as many as possible) functions in GAP that support matrices as input, and teach them to support `MatrixObj`
  - [ ] as this is a herculean effort, also think about ways to allow backwards compatibility
- [ ] allow `mat[i,j]` can be used to read and write matrix entries
  - [ ] perhaps also allow this for plain "lists of lists", to make it easier to write code which accepts both lists of lists and MatrixObj?
- [ ] consider allowing for `0 x n` and `m x 0` matrix objects. These can then properly remove existing hacks like `EmptyMatrix`, `EmptyRowVector`, ...
- [ ] also think about vectors...
  - [ ] we want both row and column vectors
  - [ ] for non-commutative base domains, row vectors should be left module elements, while column vectors should be right modules
[...]

PR #4509, opened 2021-05-26
Remove the MatrixObj todo lists

ssiccha:
These are for now put into https://hackmd.io/WAShW7z9Rp2rJRq1AVnWAQ.


fingolfin:
In principle I am for this (I suggested it as one possible course of action). I worry slightly that the HackMD can be accessed and edited by anyone with the URL (which is now public via this PR; and in any case, we probably will want to add some links to it somewhere?). That said, one has to be logged in, and if we really fear (or worse, encounter) vandalism we can restrict access to it to owners. I just hope someone will make a backup from time to time?

fingolfin:
When/if this is merged, we should e.g. add a link to the HackMD to https://github.com/gap-system/gap/issues/945, right? And remove references to the comments/file removed in this PR from that HackMD.

Moreover, https://github.com/gap-system/gap/issues/945 itself lists a bunch of TODOs and issues; those also should be merged in that big list?

ssiccha:
I added a link to a backup in #945.

fingolfin (approved):

PR #4516, opened 2021-05-27
Improve MatrixObj documentation

ssiccha:
Make section on required operations for new MatrixObj objects more
discoverable: rename it and add a reference to the introduction
of its chapter.

### DIFF
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -147,6 +147,10 @@
 ##  Therefore it may be useful to set the filter
 ##  <Ref Filt="IsNoImmediateMethodsObject"/> in the definition of new kinds
 ##  of vector and matrix objects.
+##  <P/>
+##  For information on how to implement new <Ref Filt="IsMatrixObj"/> and
+##  <Ref Filt="IsVectorObj"/> representations see Section
+##  <Ref Sect="Implementing New Vector and Matrix Objects Types"/>.
 ##  <#/GAPDoc>
 ##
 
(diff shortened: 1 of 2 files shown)


ThomasBreuer (commented):
I do not understand the change of the section heading. Also the other operations are "required", the point is that one has to provide specific methods for the ones without default methods when one implements a new kind of vector or matrix objects. The proposed heading does not express this.

ssiccha:
I agree that "required operations" does not fit here. I don't like the current name either though.

IMO, the idea behind this chapter is to list the minimal set of methods one needs to install for a new type of vector or matrix objects to make it work. So I'll change the name of the chapter to "Implementing New Vector and Matrix Objects Types". What do you think?

ThomasBreuer:
O.K.
(It is not the title of the chapter but the title of the section.)

lib/matobj2.gd
```diff
@@ -130,6 +130,10 @@
 ##  However, row list matrices behave nearly like lists of row vectors
 ##  that insist on being dense and containing only vectors of the same
 ##  length and with the same base domain.
+##  <P/>
+##  New <Ref Filt="IsMatrixObj"/> and <Ref Filt="IsVectorObj"/> representations
+##  have to implement methods for the operations listed in Section
+##  <Ref Sect="Implementing New Vector and Matrix Objects Types"/>
```
fingolfin, line 136:
```suggestion
##  <Ref Sect="Implementing New Vector and Matrix Objects Types"/>.
```

fingolfin: Hmm, but of course this is just the minimum; in practice some more will almost always be implemented for performance? But anyway, shouldn't this be explained in detail in section "Implementing New Vector and Matrix Objects Types"; and then we just put a link to that here?

ssiccha: Jepp, I've replaced this by only a link to that section.

ssiccha:
Rebased onto master.

lib/matobj2.gd
fingolfin: Will you still apply my fix (adding a missing `.`)?

ssiccha: Done.

fingolfin (approved):

PR #4517, opened 2021-05-27
Add basic operations for row/column reductions in matrix objects

### LINKED ISSUE #3962: MatrixObj: basic operations for row/column reduction
To be able to write all kinds of algorithms that work at least somewhat efficiently with proper MatrixObj implementations, we need to allow people to write code that avoids extracting and manipulating rows. That means adding a bunch of operating facilitating this. Perhaps we already have some of these, but I couldn't find them, and they are not mentioned in chapter 26 of the ref manual... The names are of course tentative, the functionality is what is important

(CC @mohamed-barakat @ThomasBreuer @User141)

- `MultMatrixRow( mat, row, scalar )`  does `mat[row] :=mat[row] * scalar`;
- `MultMatrixColumn( mat, col, scalar )`  same for columns
- to support non-commutative base domains, also need an operation which does `mat[row] :=scalar * mat[row]` and same for columns; unfortunately we can't just overload the same function, as a positive integer could denote both a scalar and a row/column index. Perhaps `MultMarixRowLeft` and `MultMarixRowRight` ? And then have `MultMatrixRow` be a synonym for `MultMatrixRowRight` (@mohamed-barakat suggestions welcome :-)

- `AddMatrixRows(mat, dstrow, srcow, scalar`) does `mat[dstrow] := mat[dstrow] + mat[srcrow] * scalar`

- same for columns; and for scalars coming from the left (I think we assume that addition is commutative in too many places for it to make sense to allow variants there, too)

- `SwapMatrixRows(mat, row1, row2)`
- `SwapMatrixColumns(mat, col1, col2)`

We might want more (e.g. one could have `PermuteMatrixRows(mat, perm)`), but those listed above are an absolute minimum. If you can think of others, chime up.

wucas:
# Description
Closes  #3962 
## Text for release notes 
Implements elementary matrix operation for MatrixObjects


lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##  <#/GAPDoc>
+##
+InstallMethod( MultMatrixColumn, "for a mutable matrix object, one column number, and an scalar",
+  [ IsMatrixObj and IsMutable, IsInt, IsObject ],
+  function( mat, column, scalar )
+    local i;
+    
+    # Checks
+    if not( 0 < column and column <= NrCols(mat) ) then
+        Print("The second argument row has to fulfill 0 < row <= NrCols(mat) ");
```
ThomasBreuer: row -> column

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##  <P/>
+##  For a matrix <A>mat</A> such that <A>mat</A> is mutable, a row index <A>i</A>  
+##  and a scalar <A>elm</A> this operation multiplies <A>elm</A> with the <A>i</A>-th row 
+##  in-place.
+##  Note that by default <A>elm</A> is multiplied from the left. 
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+InstallMethod( MultMatrixRow, "for a mutable matrix object, one row number, and an scalar",
```
ThomasBreuer: an scalar -> a scalar (also in other places)

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+    fi;
+  
+    for i in [1..NrCols(mat)] do
+        mat[row,i] := scalar * mat[row,i];
+    od;
+    
+  end );
+
+
+# TODO FOR MUTABLE MATRICES
```
ThomasBreuer: Is this (commented out) code needed?

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##  <Oper Name="MultMatrixRow" Arg='mat,i,elm' Label=""/>
+##  
+##  <Returns>nothing</Returns>
+##
+##  <Description>
+##  <P/>
+##  For a matrix <A>mat</A> such that <A>mat</A> is mutable, a row index <A>i</A>  
+##  and a scalar <A>elm</A> this operation multiplies <A>elm</A> with the <A>i</A>-th row 
+##  in-place.
+##  Note that by default <A>elm</A> is multiplied from the left.
```
ThomasBreuer, line 1590: Here you can add a cross-reference to the function that multiplies from the right.

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##  in-place.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+InstallMethod( MultMatrixRowLeft, "for a mutable matrix object, one row number, and an scalar",
+  [ IsMatrixObj and IsMutable, IsInt, IsObject ],
+  function( mat, row, scalar )
+ 
+    MultMatrixRow(mat,row,scalar);
```
ThomasBreuer, line 1673: I would put the explicit code into a method of `MultMatrixRowLeft`, and call `MultMatrixRowLeft` in the method for `MultMatrixRow`; then calling the more specific operation directly would save one indirection.

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
   end );
 
 
+############################################################################
+# Elementary matrix operations
+############################################################################
+
+############################################################################
+##
+#O  MultMatrixRow( <M>, <row>, <scalar> )
```
ThomasBreuer, line 1577: scalar -> elm

lib/matobj.gi
```diff
@@ -1568,3 +1568,572 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+    fi;
+ 
+    if row1 <> row2 then
+        
+        for i in [1..NrCols(mat)] do
+            mat[row1,i] := mat[row1,i] + scalar * mat[row2,i];
+        od;
+        
+    else
+        MultMatrixRow(mat,row1,scalar+1);
```
ThomasBreuer, line 1815: I am not sure: Can we assume that a method for `scalar + 1` is always available?

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,330 @@ InstallMethod( NewCompanionMatrix,
...
 
+############################################################################
+# Elementary matrix operations
+############################################################################
+
+############################################################################
+##
+#M  MultMatrixRow( <M>, <row>, <scalar> )
+##
+InstallMethod( MultMatrixRow, "for a mutable IsPlistMatrixRep, one row number, and an scalar",
```
ThomasBreuer: `IsPlistMatrixRep` is just one implementation, the method is installed for `IsRowListMatrix`.

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,330 @@ InstallMethod( NewCompanionMatrix,
...
+  function( mat, column, scalar )
+    local i;
+    
+    # Checks
+    if not( 0 < column and column < NrCols(mat) ) then
+        Print("The second argument row has to fulfill 0 < row < NrCols(mat) ");
+        return fail;
+    fi;
+  
+    # TODO ASK MAX
```
ThomasBreuer, line 1325: So this code should be commented out.

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,314 @@ InstallMethod( NewCompanionMatrix,
...
+    
+  end );
+  
+
+############################################################################
+##
+#M  SwapMatrixRows( <M>, <row1>, <row2> )
+##
+InstallMethod( SwapMatrixRows, "for a mutable matrix object, one row number, second row number",
+  [ IsRowListMatrix and IsMutable, IsInt, IsInt ],
```
ThomasBreuer, line 1545: The method is for `IsRowListMatrix`, the comment should say this.

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,314 @@ InstallMethod( NewCompanionMatrix,
...
+        fi;
+        if not( 0 < row2 and row2 <= NrRows(mat) ) then
+            ErrorNoReturn("the third argument row has to fulfill 0 < row <= NrRows(mat) ");
+        fi;
+        
+        for i in [1..NrCols(mat)] do
+            temp := mat[row1,i];
+            mat[row1,i] := mat[row2,i];
+            mat[row2,i] := temp;
+        od;
```
ThomasBreuer, line 1563: For a row list matrix, can't we just take the row objects, and swap them?

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,314 @@ InstallMethod( NewCompanionMatrix,
...
+    local i;
+    
+    # Checks
+    if not( 0 < column and column <= NrCols(mat) ) then
+        ErrorNoReturn("the second argument column has to fulfill 0 < column <= NrCols(mat) ");
+    fi;
+  
+    for i in [1..NrRows(mat)] do
+        mat[i,column] := scalar * mat[i,column];
+    od;
```
ThomasBreuer, line 1377: This is like in the generic method for `IsMatrixObj`. Thus we need no special method for `IsRowListMatrix` (also in other places).

ThomasBreuer (changes requested):
I have left some comments

lib/matobj.gi
```diff
@@ -1579,15 +1579,16 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
 ##  <Description>
 ##  <P/>
-##  For a matrix <A>mat</A> such that <A>mat</A> is mutable, a row index <A>i</A>  
-##  and a scalar <A>elm</A> this operation multiplies <A>elm</A> with the <A>i</A>-th row 
-##  in-place.
-##  Note that by default <A>elm</A> is multiplied from the left. 
+##  Multiplies the <A>i</A>-th row of the mutable matrix <A>mat</A> with the scalar 
+##  <A>elm</A> from the left in-place.
+##  <P/>
+##  <C>MultMatrixRowLeft</C> is a synonym for <C>MultMatrixRow</C>.
```
ThomasBreuer, line 1591: The recommended GAPDoc syntax for references of documented variables would look as follows.
`<Ref Oper="MultMatrixRowLeft"/> is a synonym of <Ref Oper="MultMatrixRow"/>.`

ThomasBreuer (changes requested):

lib/matobj.gi
ssiccha: Also inside the definition of `MultMatrixRow` and `MultMatrixRowLeft` itself?

ThomasBreuer: Yes, GAPDoc is smart enough to detect this.

lib/matobj.gi
fingolfin: In fact M,row, scalar -> mat,i,elm.

or we just delete these lines, they have no purpose?

AnnaKDS: Sergio told us that those lines are legacy and not needed any more. We‘ll go through our code and delete those everywhere.

lib/matobj.gi
```diff
@@ -1568,3 +1568,394 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##
+InstallMethod( MultMatrixRow, "for a mutable matrix object, one row number, and a scalar",
+  [ IsMatrixObj and IsMutable, IsInt, IsObject ],
+  function( mat, row, scalar )
+    local i;
+    
+    # Checks
+    if not( 0 < row and row <= NrRows(mat) ) then
+        ErrorNoReturn("the second argument <row> has to fulfill 0 < row <= NrRows(mat) ");
+    fi;
```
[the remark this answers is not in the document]
AnnaKDS: So far, we have only written tests for SwapMatrixRows and SwapMatrixColumns. We plan to also write tests for all other functions that we added.

lib/matobj.gi
```diff
@@ -1568,3 +1568,394 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+##  <#/GAPDoc>
+##
+InstallMethod( MultMatrixRow, "for a mutable matrix object, one row number, and a scalar",
+  [ IsMatrixObj and IsMutable, IsInt, IsObject ],
+  function( mat, row, scalar )
+    local i;
+    
+    # Checks
+    if not( 0 < row and row <= NrRows(mat) ) then
+        ErrorNoReturn("the second argument <row> has to fulfill 0 < row <= NrRows(mat) ");
```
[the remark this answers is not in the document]
wucas: Tbh I don't know if it does. Actually in 26.7-1 the manual says "It is not specified what happens if i is larger than the length of v" for vectors, the entry for mattrix element access in 26.11-1 doesn't mention whether it does bound checks or not.

lib/matobj.gi
AnnaKDS: I tried replacing this line. Those are the warnings I get:
`#W WARNING: non resolved reference: rec(
  Oper := "MultMatrixRowLeft" )
#W WARNING: non resolved reference: rec(
  Oper := "MultMatrixRow" )`

lib/matobj.gi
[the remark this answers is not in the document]
wucas: Looking at the code for `IsPlistMatrixRep` it doesn't look like any bound checks are performed. Am I missing something?

```
InstallMethod( MatElm, "for a plist matrix and two positions",
  [ IsPlistMatrixRep, IsPosInt, IsPosInt ],
  function( m, row, col )
    return m![ROWSPOS][row]![ELSPOS][col];
  end );
```

lib/matobj.gi
```diff
@@ -1568,3 +1568,394 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
...
+        mat[row,i] := mat[row,i] * scalar;
+    od;
+    
+  end );
+
+############################################################################
+##
+#O  MultMatrixColumnLeft( <M>, <row>, <scalar> )
+##
+##  <#GAPDoc Label="MultMatrixColumnLeft">
```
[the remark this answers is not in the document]
wucas: What exactly is the policy on that? We oriented ourselves on the code that already existed. There are quite a few documentation comments in `matobj.gi`. The same holds for the `#O` and `#M` lines. They are everywhere so we assumed they are for something we don't know about. Can we remove them then evrywhere in the matobj stuff? Or even in the whole library? Maybe in a new PR?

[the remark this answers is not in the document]
AnnaKDS: In our code I already removed all lines starting with #O or #M. Not in the entire `matobj.gi` file, though.

lib/matobj.gi
ThomasBreuer: Cross-references via `<Ref>` elements work only if the referenced element can be matched; if it has a label then the `<Ref>` element needs this label, too. From another point of view: Labels make sense only to distinguish several `<Oper>` or so elements with the same identifier, and then a `<Ref>` element can point to a specific one, via the label.

AnnaKDS: Ok, thanks for the clarification!

lib/matobj.gi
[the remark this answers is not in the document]
fingolfin: But that ends up performing a bounds check, because in GAP access to a list via `list[idx]` performs a bounds check.

lib/matobj.gi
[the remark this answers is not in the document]
fingolfin: The documentation comments for newly declared operations, attributes, etc. all should be next to the declaration of that operation, attribute, etc., not next to one (arbitrary) method. 

In contrast to this, the GAPDoc comments inside `matobj.gi` are all for specific methods, or they document concepts. E.g. the first one starting in line 21, with label "MatrixObjCompare", explains how MatrixObj instances are supposed to be compared, and to that end also "documents" the `\=` method "for two vector objects". Since this is about a specific *method* (or rather, a bunch of them), and not a general operation, it makes place to put it right next to where those methods are defined.

lib/matobj.gi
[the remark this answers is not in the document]
wucas: Ah alright. We just removed the bound checks. The only reason I still kind of like the idea of having them is the Error Message that's shown. When we remove them its a rather cryptic one that at least less experienced users might have trouble with to figure out what the actual problem was. That of course makes sense since the function that throws the error can't know in what context it was called. Our message on the other hand was quite clearly pointing at the wrong argument and told the user what range was expected.

Of course performance wise performing two bound checks doesn't make sense. So should we just remove them and live with a less clear error message or implement a NC version? I'm a bit torn what makes the most sense.
We'll remove it for now though. That solution to get a clearer error message is quite hacky and probably not the right way to do that.

lib/matobj.gi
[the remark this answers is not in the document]
danielrademacher: Tests for all functions have been added.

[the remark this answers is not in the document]
danielrademacher: It seems like GAP is always using the "IsRowListMatrix" functions. Is there a way to define a matrix which is not a "RowListMatrix"?

lib/matobj.gi
```diff
@@ -1568,3 +1568,142 @@ InstallMethod( \[\,\]\:\=, "for a matrix object, two positions, and an object",
   end );
 
 
+############################################################################
+# Elementary matrix operations
+############################################################################
+
+InstallMethod( MultMatrixRow, "for a mutable matrix object, one row index, and a scalar",
```
fingolfin, line 1575: This may just be me, but following up on my previous comment, I think I would find it easier if this said `MultMatrixRowLeft`, i.e., 
```suggestion
InstallMethod( MultMatrixRowLeft, "for a mutable matrix object, one row index, and a scalar",
```
That way it is immediate that the code is correct, I don't need to remember that `MultMatrixRow` is a synonym.

But of course this is subjective, feel free to disagree, this PR shall not be blocked over this :-)

lib/matobj2.gd
```diff
@@ -1926,3 +1926,201 @@ DeclareOperation( "[]:=", [ IsMatrixObj, IsPosInt, IsPosInt, IsObject ] );
...
+##  <Returns>nothing</Returns>
+##
+##  <Description>
+##  <P/>
+##  Swaps the <A>i</A>-th column and <A>j</A>-th column of a mutable matrix <A>mat</A>.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+DeclareOperation( "SwapMatrixColumns", [ IsMatrixObj and IsMutable, IsInt, IsInt ] );
```
fingolfin, line 2126: Please end the file in a newline
```suggestion
DeclareOperation( "SwapMatrixColumns", [ IsMatrixObj and IsMutable, IsInt, IsInt ] );

```

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,72 @@ InstallMethod( NewCompanionMatrix,
...
+############################################################################
+# Elementary matrix operations
+############################################################################
+
+############################################################################
+##
+#M  MultMatrixRow( <M>, <row>, <scalar> )
+##
+InstallMethod( MultMatrixRow, "for a mutable IsRowListMatrix, one row number, and a scalar",
+  [ IsRowListMatrix and IsMutable, IsInt, IsObject ],
```
fingolfin, line 1298: This is the wrong file; it is intended for the implementation of `IsPlistMatrix` (see also PR #2973). These methods should go into `lib/matobj.gi`.

lib/matobjplist.gi
```diff
@@ -1286,3 +1286,72 @@ InstallMethod( NewCompanionMatrix,
...
+#M  SwapMatrixRows( <M>, <row1>, <row2> )
+##
+InstallMethod( SwapMatrixRows, "for a mutable IsRowListMatrix, one row number, second row number",
+  [ IsRowListMatrix and IsMutable, IsInt, IsInt ],
+  function( mat, row1, row2 )
+    local temp;
+  
+    temp := mat[row1];
+    mat[row1] := mat[row2];
+    mat[row2] := temp;
```
fingolfin, line 1355:
```suggestion
    mat{[row1,row2]} := mat{[row2,row1]};
```

lib/matobj.gi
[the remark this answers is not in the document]
fingolfin: Good question. In the end, this is what `IsPlistMatrixRep` (or perhaps `IsStrictPlistMatrixRep`) should do; see PR #2973 . It shouldn't be hard to do that atop the mentioned PR.

lib/matobjplist.gi
fingolfin: Or perhaps add a new file `lib/matrobjrowlist.gi` (to get GAP to actually read it, edit `lib/read5.g` around line 110 to add a suitable `Read` statement).

danielrademacher (changes requested):
The commit e334988 produced an error. Therefore, the old code of the "SwapMatrixRows" function is used again.

lib/matobjplist.gi
```diff
@@ -1348,10 +1348,6 @@ InstallMethod( AddMatrixRowsRight, "for a mutable IsRowListMatrix, one row numbe
 InstallMethod( SwapMatrixRows, "for a mutable IsRowListMatrix, one row number, second row number",
   [ IsRowListMatrix and IsMutable, IsInt, IsInt ],
   function( mat, row1, row2 )
-    local temp;
-  
-    temp := mat[row1];
-    mat[row1] := mat[row2];
-    mat[row2] := temp;
+    mat{[row1,row2]} := mat{[row2,row1]};
```
danielrademacher, line 1357: This gives us an error:

Error, List Assignments: <rhss> must be a dense list (not a positional object) in
  mat{[ row1, row2 ]} := mat{[ row2, row1 ]}; at .../GAP/gap/lib/matobjplist.gi:1351 called from 
<function "SwapMatrixRows for a mutable IsRowListMatrix, one row number, second row number">( <arguments> )
 called from read-eval loop at *stdin*:6

ThomasBreuer: This problem shows a bug in GAP, see issue #4533.

lib/matobj2.gd
```diff
@@ -1926,3 +1926,201 @@ DeclareOperation( "[]:=", [ IsMatrixObj, IsPosInt, IsPosInt, IsObject ] );
...
+##  <Oper Name="MultMatrixRowLeft" Arg='mat,i,elm'/>
+##  
+##  <Returns>nothing</Returns>
+##
+##  <Description>
+##  <P/>
+##  Multiplies the <A>i</A>-th row of the mutable matrix <A>mat</A> with the scalar 
+##  <A>elm</A> from the left in-place.
+##  <P/>
+##  <Ref Oper="MultMatrixRowLeft"/> is a synonym of <Ref Oper="MultMatrixRow"/>.
```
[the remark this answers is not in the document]
fingolfin: @wucas @AnnaKDS @danielrademacher any thoughts on this?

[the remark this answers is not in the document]
danielrademacher: Sorry, for the late reply @fingolfin.I like it when it's consistent, so I agree. To work on this and the next comment have been on my list since the weekend, but I haven't gotten to it yet. I'll finish this by tomorrow evening.

lib/matobj2.gd
```diff
@@ -1813,3 +1816,339 @@ DeclareOperation( "Randomize", [ IsMatrixOrMatrixObj and IsMutable, IsRandomSour
...
+##    machinery would cause that 'Matrix( <list> )' returns an *immutable*
+##    value, which is not acceptable.
+##    The declaration as a mutable attribute in 'lib/matobj2.gd' works
+##    because this file is read before 'lib/reesmat.gd',
+##    but the consequence is that the attribute values in the semigroup
+##    context are also mutable (try the manual example for 'Matrix');
+##    this is a bug.
+##    (One could turn any attribute into a mutable one, by adding a
+##    corresponding declaration early enough.)
+##
```
fingolfin, line 2059: I think this text was added back by accident, and needs to be removed again.

lib/matrobjrowlist.gi
```diff
@@ -0,0 +1,83 @@
...
+    mat[row] := mat[row] * scalar;
+    
+  end );
+
+
+############################################################################
+##
+#M  AddMatrixRows( <M>, <row1>, <row2>, <scalar> )
+##
+InstallMethod( AddMatrixRowsLeft, "for a mutable IsRowListMatrix, one row number, second row number, and a scalar",
```
fingolfin, line 47: "one" and "second" together makes no sense; maybe "first" and "second"... but really, I'd write it as
```suggestion
InstallMethod( AddMatrixRowsLeft, "for a mutable IsRowListMatrix, two row numbers, and a scalar",
```

lib/matrobjrowlist.gi
```diff
@@ -0,0 +1,83 @@
...
+    mat[row] := scalar * mat[row];
+    
+  end );
+
+
+############################################################################
+##
+#M  MultMatrixRowRight( <M>, <row>, <scalar> )
+##
+InstallMethod( MultMatrixRowRight, "for a mutable IsRowListMatrix, one row number, and a scalar",
```
fingolfin, line 34:
```suggestion
InstallMethod( MultMatrixRowRight, "for a mutable IsRowListMatrix, a row number, and a scalar",
```

lib/matrobjrowlist.gi
fingolfin: Similar elsewhere

fingolfin (commented):
This PR is still marked as draft. What would you like to add before considering this ready for final review and merging?

lib/matobj2.gd
danielrademacher: Done

danielrademacher: I'm sorry. I think that happened during a merge.

lib/matrobjrowlist.gi
danielrademacher: Good point. Has been corrected.

lib/matrobjrowlist.gi
danielrademacher: Done.

danielrademacher:
At the moment some functions cannot be tested because we don't have a MatrixObj which is not a RowListMat. Everything else should be finished (A little bit disappointed because of the slow performance mentioned in #3962 ).

ssiccha:
> A little bit disappointed because of the slow performance mentioned in #3962

The slow performance is only for `IsMatrix` lists of lists, though. Of course it would be nice to also have a performant version for those matrices, but we can work on that later.

fingolfin:
PR #4557 may help! Once that's in, we could use it here (this actually motivated me to finally tackle `InstallEarlyMethod`)

danielrademacher (approved):

fingolfin (approved):
These are a good start and should be merged; we can improve them later on, after all (better tests, performance tuning, using `InstallEarlyMethod`, etc.)

Issue #4503, opened 2021-05-20
Replace `IsMatrix => IsMatrixObj` implication, by introducing `IsMatrixOrMatrixObj`?


fingolfin:
(This issue is closely related to issue #2972)

Currently, we have an implication `IsMatrix => IsMatrixObj`:
```gap
InstallTrueMethod( IsMatrixObj, IsMatrix );
```
The motivation for that was to make it easier to write code that works both for old-style list-of-list matrices, and new-style MatrixObj matrices.

The drawback is that it is harder to write code which *requires* new style MatrixObj, e.g. to support empty matrices, or to get accurate `BaseDomain` results.

As a compromise, we could instead introduce a new filter/category/whatever `IsMatrixOrMatrixObj` in analogy to `IsListOrCollection`:
```gap
DeclareCategory( "IsMatrixOrMatrixObj", IsObject );
InstallTrueMethod( IsMatrixOrMatrixObj, IsMatrix );
InstallTrueMethod( IsMatrixOrMatrixObj, IsMatrixObj );
```
The drawback would be that we'd have to adjust a lot of operations and methods to use this filter. But we already have to do that anyway, the question is just whether we change to `IsMatrixOrMatrixObj` or to `IsMatrix`. On the upside, we can also start to get rid of some hacks; e.g. 
```gap
InstallMethod( ZeroMutable,
    [ IsMatrixObj ],
    function( M )
    if IsPlistRep( M ) then
      TryNextMethod();
    else
      return Matrix( ZeroMutable( Unpack( M ) ), M );
    fi;
    end );
```
could become
```gap
InstallMethod( ZeroMutable,
    [ IsMatrixObj ],
    M -> Matrix( ZeroMutable( Unpack( M ) ), M ) );
```
(for that function one may then also wonder whether perhaps using `ZeroMatrix` or `NewZeroMatrix` would be more appropriate, but that's a separate concern).

Are there perhaps other negative consequences of this that I am overlooking? @ThomasBreuer might be able to think of some...

ThomasBreuer:
The question is about generic methods (and about the documentation of the MatrixObj setup).

I think the main reason why on wants to have different generic methods for an `IsPlistRep` matrix and a non-`IsPlistRep` MatrixObj is that the code for the non-`IsPlistRep` MatrixObj delegates to `IsPlistRep` objects (via `Unpack`), and using this code for an `IsPlistRep` matrix would mean infinite recursion.
[...]

PR #4518, opened 2021-05-27
introduce `IsMatrixOrMatrixObj`

ThomasBreuer:
as sketched in issue #4503.

In this first step, I have changed the documentation of the defining filters and replaced `IsMatrixObj` by `IsMatrixOrMatrixObj` in most places.

(In a next step, we can now change the method installations mentioned in issue #4503.)

### DIFF
--- a/lib/matobj1.gd
+++ b/lib/matobj1.gd
@@ -91,12 +91,13 @@ DeclareCategory( "IsVectorObj", IsVector and IsCopyable );
 ##  <Filt Name="IsMatrixOrMatrixObj" Arg='obj' Type="category"/>
 ##
 ##  <Description>
-##  This filter is the <Q>common roof</Q> for <Ref Filt="IsMatrix"/> and
-##  <Ref Filt="IsMatrixObj"/>.
+##  Several functions are defined for objects in <Ref Filt="IsMatrix"/> and
+##  objects in <Ref Filt="IsMatrixObj"/>.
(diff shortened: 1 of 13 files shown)


fingolfin:
`testlibgap` fails -- we need to either change `tst/testlibgap/api.c:78` or else adjust `GAP_IsMatrixObj` to check `IsMatrixOrMatrixObj` instead of `IsMatrixObj`...

lib/matobj1.gd
```diff
@@ -85,6 +84,58 @@
...
 
 
+#############################################################################
+##
+##  <#GAPDoc Label="IsMatrixOrMatrixObj">
+##  <ManSection>
+##  <Filt Name="IsMatrixOrMatrixObj" Arg='obj' Type="category"/>
+##
+##  <Description>
+##  This filter is the <Q>common roof</Q> for <Ref Filt="IsMatrix"/> and
```
fingolfin, line 94: I am not sure if "common roof" will be understood ? Perhaps we could borrow from the `IsListOrCollection` documentation, which starts like this:
> Several functions are defined for both lists and collections, [...]
> IsListOrCollection is a supercategory of IsList and IsCollection (that  is, all lists and collections lie in this category)

fingolfin (approved):
There are obvious things to do next (like changing a bunch of methods back to `IsMatrixObj` and simplifying their code), but I don't think we need to wait for this. From my POV this can be merged right away.

@ThomasBreuer if you agree, let's do it, and let's continue work afterwards.

Issue #4409, opened 2021-04-16
Unspecified and unused optional argument `req` in `NewRepresentation` and `DeclareRepresentation`


fingolfin:
The documentation for `NewRepresentation` and `DeclareRepresentation` mentions and optional fourth argument `req` but does not specify what it does.

Looking at the implementation, as far as I can tell it is simply ignored:
```gap
BIND_GLOBAL( "NewRepresentation", function ( arg )
    local   rep, filt;
...
    # Create the filter.
    if LEN_LIST(arg) = 3  then
        rep := NEW_FILTER( arg[1] );
    elif LEN_LIST(arg) = 4  then
        rep := NEW_FILTER( arg[1] );
    else
        Error("usage: NewRepresentation( <name>, <super>, <slots> [, <req> ] )");
    fi;
...
```
I looked back through our extended version history, and it was like that since the initial import of GAP4 into CVS on 1996-09-25. So maybe this is code that simply was never finished

Yet some code "uses" it:
```gap
DeclareRepresentation(
    "IsGF2VectorRep",
    IsDataObjectRep and IsVectorObj, [],
    IsRowVector );
```
Unfortunately it is not completely clear to me what the intention is -- how does this optional list of filters differ from the second argument `super`?

Maybe `NewFamily` gives a hint, which has argument `req` (for "required") and `imp` (for "implied"), which are also both filter lists. But the first thing it then does with them is to join them together:
```
    imp_filter := WITH_IMPS_FLAGS( AND_FLAGS( imp_filter, req_filter ) );
```
Well, that and store `req` into `family!.REQ_FLAGS` and use that for printing...

We should figure out which code uses this unused 4th argument (e.g. by removing it, then trying to load GAP and packages, and see what breaks); and then decide how to proceed. E.g. if not much uses it, we could just remove it and "fix" all affected call sites.

ThomasBreuer:
The documentation of `NewRepresentation` and `DeclareRepresentation` mentions the optional fourth argument but does not state what it means. The code ignores the fourth argument, thus it cannot be accessed from GAP's internal data. Thus I think we should first remove it from the documentation, and change the places in the GAP library where a fourth argument is given.

(I do not remember what the intended meaning of `<req>` was.)

[...]

Issue #1042, opened 2016-12-28
Third and fourth argument to NewRepresentation and DeclareRepresentation are ignored


fingolfin:
This is more a question or a "request for comments" and an issue:

Today, I was wondering what the `slots` argument to `DeclareRepresentation` and `NewRepresentation` really do es(the manual is a bit vague on that), and also what the fourth argument `req` (supposedly for "required"?) is about -- the documentation does not talk about it at all.

To my surprise, it seems that both arguments are simply ignored?!? Perhaps they were meant for future extensions that were never implemented? Or perhaps they were removed at some point?

There are some things we could do:

1. identify what those arguments were meant for, and implement that functionality
2. remove `req` from the documentation, but keep accepting it silently for backwards compatibility
3. remove both `slots` and `req` from the documentation, but keep accepting them silently for backwards compatibility
4. as a variation of 2 and 3, we could also warn if either is used (say, if `InfoObsolete` is set to a non-zero value).

Regarding 1, perhaps `slots` was meant to be used as some kind of limited type checking, where objects are not allowed to bind other "slots" than those. But in reality, this never (????) was enforced. So it is very likely that if we started to enforce it, code would break. Hence option 1 does not seem very realistic at this point -- though perhaps it could be implemented in the form of an `InfoWarning` message or so?

hulpke:
As far as I remember, these "slots" were supposed to be record components that could, respectively that would have to exist in a component object. Its purpose presumably was to keep control over what components were entered (and to force most of them into attributes) but it was never implemented and never missed, furthermore many declarations in the library only list few components.

 my preference would be for two, that is removing it from the documentation but quietly accepting them as no-op arguments for existing code.

frankluebeck:
I vote for variant 3.

fingolfin:
@hulpke Did you perhaps also mean variant 3 (and not 3)? Just wondering because what you write refers to `slots`, and you write "accepting them" which would fit 3, not 2.

hulpke:
Yes, 3. I didn’t read that far and tghought req was the general ``requirements’’ for using `DeclareRepresentation’. Sorry!

  Alexander

ThomasBreuer:
This can be closed since #4519 has been merged.

PR #4519, opened 2021-05-27
Make third argument of `DeclareRepresentation` and `NewRepresentation` optional, and document that it and the fourth argument are (and always were) unused

fingolfin:
See the commit messages for details.

See also issues #4409 and #1042

### DIFF
--- a/lib/type.g
+++ b/lib/type.g
@@ -166,11 +166,11 @@ end );
 
 #############################################################################
 ##
-#F  NewRepresentation( <name>, <super>, <slots>[, <req>] )  .  representation
+#F  NewRepresentation( <name>, <super>[, <slots>[, <req>]] )  .  representation
 ##
 ##  <#GAPDoc Label="NewRepresentation">
 ##  <ManSection>
-##  <Func Name="NewRepresentation" Arg='name, super, slots[, req]'/>
+##  <Func Name="NewRepresentation" Arg='name, super[, slots[, req]]'/>
 ##
 ##  <Description>
 ##  <Ref Func="NewRepresentation"/> returns a new representation <A>rep</A>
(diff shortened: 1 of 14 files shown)


lib/type.g
```diff
@@ -190,26 +190,13 @@ end );
...
-##  subrepresentation of it, and <A>slots</A> tells what positions of the
-##  objects in the representation <A>rep</A> may be bound.
-##  In the latter case,
-##  <A>rep</A> must be <Ref Filt="IsComponentObjectRep"/> or a
-##  subrepresentation of, and <A>slots</A> lists the admissible names of
-##  components that objects in the representation <A>rep</A> may have.
-##  The admissible positions resp. component names of <A>super</A> need not
-##  be listed in <A>slots</A>.
+##  The optional third and fourth arguments <A>slots</A> and <A>req</A>
+##  are unused (and never were) only provided for backwards compatibility.
```
ThomasBreuer, line 194: perhaps "are (and always were) unused and only provided for backwards compatibility"

ThomasBreuer (commented):
I agree with the changes, and have made just one comment concerning a formulation.
Concerning the newly inserted `# TODO` comments:
I think they might be confusing for readers in the future.
The filters in question have not been used before, in particular they were not intended to be automatically implied by the newly created filters.

lib/type.g
fingolfin: Indeed, that's much better. Fixed now

fingolfin:
The reason I added those TODO comments is because without them, it feels as if some information is lost... and yeah, I know these were not implied, but I still wonder why? I guess this is the same underlying issue as in PR #4499: Why doesn't `IsBlockMatrixRep` not imply `IsMatrix`? It is genuinely unclear to me. Is it really just "purity", or a deeper thing? The argument that one might want to use `IsBlockMatrixRep` for something that is not a matrix doesn't strike me as convincing -- I mean, it is *in the name* that this is for matrices. One argument of course is that nowadays, perhaps it should be `IsMatrixOrMatrixObj`. But other than that? 

On the other hand, we have had plenty of ranking issues (esp. with "hidden implications" disabled) because people used (and use) representations in method argument filters, unaware that these filters lack many seemingly "obvious" implications.


Anyway, I don't think this particular instance is worth arguing about a lot, so I just dropped these comments now *shrug*

ThomasBreuer (approved):

PR #4521, opened 2021-05-27
Turn InducedPcgs into a global function

fingolfin:
This way we don't have to mess with `WRAPPER_OPERATIONS` to prevent
people from installing methods for it.


ssiccha (approved):
LGTM. That Julia test needs to be restarted.

PR #4524, opened 2021-05-31
Speed up combinat.tst

### LINKED ISSUE #4515: IteratorOfPartitionsSet runs out of memory
This worked the day before yesterday and does not work since yesterday - perhaps after #4505? 

Happens when all packages are loaded:
```
########> Diff in /circa/scratch/gap-jenkins/workspace/GAP-pkg-update-master-q\
uicktest/GAPCOPTS/64build/GAPGMP/gmp/GAPTARGET/standard/label/kovacs/GAP-pkg-u\
pdate-master-snapshot/tst/testinstall/combinat.tst:474
# Input is:
for s in [[], [5], [1,2,3,4], [2,5,7], ["a","b","c","d","e"], [3..13]] do
     pn1:= PartitionsSet( s );
     pn2:= List( IteratorOfPartitionsSet( s ) );
     if Length(pn1) <> Length(pn2) then
       Error( "wrong number of elements" );
     elif Set(pn1) <> Set(pn2) then
       Error( "different elements" );
     fi;
   od;
# Expected output:
# But found:
########
```
but only in the GAP master branch - works fine in stable-4.11.

Could be of course a simple overflow due to many packages being loaded.

fingolfin:
Also update the #@local header

Should resolve #4515.

### DIFF
--- a/tst/testinstall/combinat.tst
+++ b/tst/testinstall/combinat.tst
@@ -2,7 +2,7 @@
 ##
 ##  This  file  tests  the functions that  mainly  deal  with  combinatorics.
 ##
-#@local i,iter,list,n,pn1,pn2,it
+#@local it,n,pn1,pn2,s
 gap> START_TEST("combinat.tst");
 
 #F  Factorial( <n> )  . . . . . . . . . . . . . . . . factorial of an integer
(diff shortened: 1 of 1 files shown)


alex-konovalov (commented):
Looks ok! I didn't test it with all packages though...

fingolfin:
Redundant, as PR #4502 contained the same change

PR #4525, opened 2021-05-31
Refactor some functions to be vararg with fixed arguments

fingolfin:
This arguably makes some of that code more readable.

It also revealed an oddity in `RestrictedPartitions` where `arg[2]` (a set of
integers) was compared against 0. Presumably, it meant to check whether
`arg[2]` is empty. Alas, the general code works fine in that case, so I simply
dropped that check.

### DIFF
--- a/lib/combinat.gi
+++ b/lib/combinat.gi
@@ -182,14 +182,13 @@ CombinationsK := function ( mset, m, n, k, comb, i )
 end;
 MakeReadOnlyGlobal( "CombinationsK" );
 
-InstallGlobalFunction(Combinations,function ( arg )
-    local   combs, mset;
-    if Length(arg) = 1  then
-        mset := ShallowCopy(arg[1]);  Sort( mset );
+InstallGlobalFunction(Combinations,function ( mset, arg... )
+    local   combs;
+    mset := ShallowCopy(mset);  Sort( mset );
+    if Length(arg) = 0  then
         combs := CombinationsA( mset, 1, Length(mset), [], 1 );
-    elif Length(arg) = 2  then
-        mset := ShallowCopy(arg[1]);  Sort( mset );
-        combs := CombinationsK( mset, 1, Length(mset), arg[2], [], 1 );
+    elif Length(arg) = 1  then
+        combs := CombinationsK( mset, 1, Length(mset), arg[1], [], 1 );
     else
         Error("usage: Combinations( <mset> [, <k>] )");
     fi;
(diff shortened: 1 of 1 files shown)


lib/combinat.gi
```diff
@@ -182,14 +182,13 @@ CombinationsK := function ( mset, m, n, k, comb, i )
 end;
 MakeReadOnlyGlobal( "CombinationsK" );
 
-InstallGlobalFunction(Combinations,function ( arg )
-    local   combs, mset;
-    if Length(arg) = 1  then
-        mset := ShallowCopy(arg[1]);  Sort( mset );
+InstallGlobalFunction(Combinations,function ( mset, arg... )
```
danielrademacher, line 185: What is the advantage of splitting mset from arg?

lib/combinat.gi
```diff
@@ -2284,27 +2273,24 @@ RestrictedPartitionsK := function ( n, set, m, k, part, i )
...
         else
-            if arg[2] = 0  then
-                parts := [  ];
-            else
-	    if not ForAll(arg[2],IsPosInt) then
-      Error("RestrictedPartitions: Set entries must be positive integers");
-	    fi;
-                parts := RestrictedPartitionsK(
-                             arg[1], arg[2], Length(arg[2]), arg[3], [], 1 );
+            if not ForAll(set,IsPosInt) then
```
danielrademacher, line 2289: Can we have an error here if "set = 0"?

danielrademacher (commented):
Looks good to me. 

The only thing I'm not sure about is whether it is ok to leave out the marked test.
---

if arg[2] = 0  then               
   parts := [  ];           
else

----

"ForAll(0,IsPosInt);" gives an error and there is no check for set.

lib/combinat.gi
danielrademacher: Ah, I see.
Is it faster to work with arg and to ask about the number of arguments in the GAP function or to write several GAP functions with different numbers of arguments and let GAP choose the right one?
Of course, it is more complex to maintain the same code in several functions.

lib/combinat.gi
danielrademacher: I'm sorry, forget that.

danielrademacher (approved):
Looks good to me.

lib/combinat.gi
fingolfin: To let GAP "choose" the "right" function would mean to use an operation (via `DeclareOperation`), not a plain global function. And that incurs more overhead than a simple global function. Besides, it would be a much bigger change than the simple refactoring done in this PR.

fingolfin:
@danielrademacher thank you for the review, and no worries. It is great that you ask questions, and I think trying to do a review is a great way to learn *if one asks questions*, so please don't hesitate to ask more of them here or elsewhere :-). And I definitely make mistakes, oh yes ... *sigh* so I am really grateful when people give my changes a sanity check.

PR #4526, opened 2021-05-31
Remove GAPInfo.FormattingStatusStdout

fingolfin:
... and instead use a new kernel helper function PRINT_FORMATTING_STDOUT().
This way we only track this piece of state once, in the kernel; tracking
something twice always carries the risk of getting out of sync.


fingolfin:
This is now part of PR #4540

PR #4529, opened 2021-06-01
`IsNoImmediateMethodsObject` is now a documented filter; it is intended for performance improvements.

ThomasBreuer:
and add a hint to the documentation of vector and matrix objects that setting this filter is recommended for them.
(There is no dedicated "performance hints" section for vector and matrix objects yet.)


fingolfin (approved):

fingolfin:
Should this really not be mentioned in the release notes? It might be useful to mention it there to draw attention to it ?

ThomasBreuer:
@fingolfin I have changed the title and the labels according to your suggestion.

PR #4530, opened 2021-06-01
ENHANCE: Perfect groups of orders 10^6-2*10^6.

hulpke:
Two data files for particular orders are kept separately for individual
download, as they are very large.


wilfwilson (approved):
Thanks for adding this.

These files seem to be 12.3MB and 51.7MB each, and so I think distributing them separately is a sensible approach.

grp/perf.grp
```diff
@@ -35,7 +35,23 @@ local p,pos;
...
+  while not READ_GAP_ROOT( libname ) do
+    Error("\n\n",
+    "For reasons of size, the perfect groups library for orders >10^6 is\n",
+    "not distributed fully by default. To access the group requested, get\n",
+    "the file ",name," from\n",
+    "https://github.com/hulpke/extraperfect\n",
+    "and put it in the `grp` subdirectory of your GAP installation. Then",
+    " type \n\nreturn;\n\n",
+    "to continue in this GAP session ",
+    "(which will read in the file).\n\n\n");
```
fingolfin, line 51: This is indeed reasonable for now, and we can easily replace this with something "better" (whatever that'll mean) once/if we have anything.

fingolfin (approved):
Great addition.

By how much does this PR increase the size of the `grp` directory (and hence of the GAP distribution)?

grp/perf34.grp
```diff
@@ -0,0 +1,158 @@
+#############################################################################
+##
+##  This file is part of GAP, a system for computational discrete algebra.
+##  It contains the perfect groups of order 1975680
+##  This data was computed by Alexander Hulpke
+##  It is distributed under the artistic license 2.0
+##  https://opensource.org/licenses/Artistic-2.0
```
fingolfin, line 7: Hmmm, that's a problem from a legal point of view. Right now we claim "GAP is GPL 2", period, without any differentation. And that's correct at least for core GAP (not for package). This PR changes that. This will require some extra changes. I can help with those, but this is a blocker for a release, IMHO.

wilfwilson:
It seems to be that the increase in size (for the newly included files in the range `perf24.grp` -- `perf34.grp`) is about 3.5MB total.

grp/perf.grp
jackschmidt: Yes. My first idea for "better" was NrPerfectLibraryGroups could check if the files are available. Back when the perfect groups library had holes in it, I used NrPerfectLibraryGroups to run tests or searches over the library. I agree the current message is very clear and very informative for an interactive session. I followed the instructions, used `return` and the search continued without any trouble.

jackschmidt:
FWIW, I checked that all of the groups have the claimed order and are perfect. The default method for `PerfectIdentification` needs to be updated to the new upper limit. I am planning on seeing if my old perfect identification code (never submitted) can be updated to prove non-isomorphism and provide a faster id.

PR #4531, opened 2021-06-02
added `GAP_IsMatrix` to `src/libgap-api.*`

ThomasBreuer:
and added/corrected some comments concerning functions about `GAP_IsMatrix*`

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -412,6 +412,7 @@ Obj GAP_NewPlist(Int capacity)
 ////
 
 static Obj IsMatrixOrMatrixObjFilt;
+static Obj IsMatrixFilt;
 static Obj IsMatrixObjFilt;
 static Obj NrRowsAttr;
 static Obj NrColsAttr;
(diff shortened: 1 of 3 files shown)


wucas (approved):
I cannot really say whether everything is correct (for that I lack knowledge of the C part of GAP) but the comments and code look good to me. Is this related to #4518 ?

src/libgap-api.c
```diff
@@ -422,6 +423,12 @@ int GAP_IsMatrixOrMatrixObj(Obj obj)
     return obj && DoFilter(IsMatrixOrMatrixObjFilt, obj) == True;
 }
 
+// Returns 1 if <obj> is a GAP matrix, 0 if not.
+int GAP_IsMatrix(Obj obj)
+{
+    return obj && DoFilter(IsMatrixFilt, obj) == True;
```
fingolfin, line 429: `IsMatrix` is not a plain filter, but rather an and-filter, so calling `DoFilter` is wrong and the cause for the crash in the CI tests.
```suggestion
    return obj && CALL_1ARGS(IsMatrixFilt, obj) == True;
```
I suggest we change `GAP_IsMatrixOrMatrixObj` and `GAP_IsMatrixObj` similarly -- while `DoFilter` works for them right now, it's a bit risky, as any change to their definition would accidentally break this code.

fingolfin (approved):

PR #4535, opened 2021-06-05
Speed up IsValidIdentifier

fingolfin:
Each time `IsValidIdentifier` got called, it invoked `ALL_KEYWORDS()`, which
in turned created a fresh list of keywords. Instead, use `GAPInfo.Keywordds`.
Moreover, sort that list to speed up the `str in GAPInfo.Keywords` test.


wilfwilson (approved):
Nice!

PR #4538, opened 2021-06-06
Ch.26 minor changes

cdwensley:
# Description
As a newcomer to the MatObj project I have been studying Chapter 26 of the manual.
This PR suggests a few corrections; rewordings etc. 
I will add in a later comment a number of questions which occurred to me as I read through the chapter.

### DIFF
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -1352,7 +1352,7 @@ DeclareOperation( "Matrix", [ IsList ]);
 ##  of <A>M</A> (provided that such a representation exists).
 ##  <P/>
 ##  The idea is that there should be an efficient way to 
-##  form the product <M>vM</M>.
+##  form the product <M>v</M><A>M</A>.
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -463,6 +463,7 @@ DeclareOperation( "ListOp", [ IsVectorObj, IsFunction ] );
 ##  <Oper Name="Unpack" Arg="v" Label="for a vector object"/>
 ##  <Oper Name="Unpack" Arg="M" Label="for a matrix object"/>
 ##
+##  <Returns>A plain list</Returns>
 ##  <Description>
 ##  Returns a new mutable plain list (see <Ref Filt="IsPlistRep"/>)
 ##  containing the entries of the vector object <A>v</A> or the matrix object
@@ -828,7 +829,7 @@ DeclareConstructor( "NewZeroVector", [ IsVectorObj, IsSemiring, IsInt ] );
 ##  <Ref Attr="BaseDomain" Label="for a vector object"/> values.
 ##  <P/>
 ##  Similarly, <Ref Constr="NewIdentityMatrix"/> returns a mutable identity
-##  matrix object with <A>m</A> rows and columns
+##  matrix object with <A>n</A> rows and columns
 ##  which has <A>filt</A> and <A>R</A> as
 ##  <Ref Attr="ConstructingFilter" Label="for a vector object"/> and
 ##  <Ref Attr="BaseDomain" Label="for a vector object"/> values,
@@ -931,7 +932,7 @@ DeclareOperation( "Randomize", [ IsRandomSource, IsMatrixOrMatrixObj and IsMutab
 ##  (see <Ref Oper="ExtractSubVector"/>)
 ##  to the positions <A>dcols</A> in <A>dst</A>,
 ##  but without creating an intermediate object and thus
-##  &ndash; at least in special cases &ndash;
+##  &ndash;at least in special cases&ndash;
 ##  much more efficiently.
 ##  <P/>
 ##  For certain objects like compressed vectors this might be significantly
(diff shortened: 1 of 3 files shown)


fingolfin (approved):
Looks good to me, thank you. Let's see if @ThomasBreuer and @ssiccha are happy with it, too.

ssiccha (approved):
LGTM :)

lib/matobj2.gd
```diff
@@ -1351,8 +1350,8 @@ DeclareOperation( "Matrix", [ IsList ]);
 ##  <Ref Attr="CompatibleVectorFilter" Label="for a matrix object"/> value
 ##  of <A>M</A> (provided that such a representation exists).
 ##  <P/>
-##  The idea is that there should be an efficient way to multiply <M>v</M>
-##  and <A>M</A>.
+##  The idea is that there should be an efficient way to 
+##  form the product <A>vM</A>.
```
ThomasBreuer, line 1354: The `<A>` element is intended to be used for arguments of the function in question, which are declared in the `Arg` attribute. In this case, we have `Arg="M"`, thus `<A>M</A>` is allowed. On the other hand, `v` is a "free" variable, thus `<M>v</M>` is an appropriate syntax but `<A>vM</A>` is not, and also `<A>v</A>` would not make sense.

lib/matobj2.gd
```diff
@@ -829,7 +828,7 @@ DeclareConstructor( "NewZeroVector", [ IsVectorObj, IsSemiring, IsInt ] );
 ##  <Ref Attr="BaseDomain" Label="for a vector object"/> values.
 ##  <P/>
 ##  Similarly, <Ref Constr="NewIdentityMatrix"/> returns a mutable identity
-##  matrix object with <A>m</A> rows and <A>n</A> columns
+##  matrix object with <A>m</A> rows and columns
```
ThomasBreuer, line 831: The declaration of `NewIdentityMatrix` introduces `n` in its `Arg` list, thus `<A>n</A>` should be used here.

lib/matobj2.gd
```diff
@@ -463,7 +463,6 @@ DeclareOperation( "ListOp", [ IsVectorObj, IsFunction ] );
 ##  <Oper Name="Unpack" Arg="v" Label="for a vector object"/>
 ##  <Oper Name="Unpack" Arg="M" Label="for a matrix object"/>
 ##
-##  <Returns>A plain list</Returns>
```
ThomasBreuer, line 466: The `<Returns>` element is used in several places. What is the reason to remove it here? (I would rather recommend to add such elements than to remove them.)

lib/matobj2.gd
```diff
@@ -924,15 +923,15 @@ DeclareOperation( "Randomize", [ IsRandomSource, IsMatrixOrMatrixObj and IsMutab
...
 ##  such that <A>dst</A> is mutable,
-##  and two lists <A>dcols</A> and <A>scols</A> of positions,
+##  and two lists <A>scols</A> and <A>dcols</A> of positions,
 ##  <Ref Oper="CopySubVector"/> assigns the entries
 ##  <A>src</A><C>{ </C><A>scols</A><C> }</C>
 ##  (see <Ref Oper="ExtractSubVector"/>)
 ##  to the positions <A>dcols</A> in <A>dst</A>,
 ##  but without creating an intermediate object and thus
-##  &ndash;at least in special cases&ndash;
+##  &ndash; at least in special cases &ndash;
```
ThomasBreuer, line 934: According to what I have read, the rule about inserting spaces around n-dashes or omitting them is that one should use the same behaviour throughout, thus making changes in a single file may lead to infinite loops.
(For textual parentheses, I prefer n-dashes to brackets because there are already lots of other brackets which have other meanings. I think spaces before the opening and after the closing n-dash increase the readability.)

ThomasBreuer (changes requested):
Most of the proposed changes are improvements. I have left a few comments.

lib/matobj2.gd
cdwensley: OK

lib/matobj2.gd
cdwensley: Happy to reinstate this.  It just seemed to me that "Returns: A plain list.  Returns a new mutable plain list ..." was unnecessarily repetitive.

lib/matobj2.gd
cdwensley: Happy to reinstate this.  Use of dashes as parentheses is not something I'm familiar with.

lib/matobj2.gd
cdwensley: My reason for changing "multiply v with M" to "product vM" was to remove any possibility that Mv might be understood by some readers.  I have only recently got it fixed in my mind that "vector" means "row vector".  Of course I should be more familiar with GAPDoc conventions.  Hopefully "<M>vM</M> is acceptable.

lib/matobj2.gd
cdwensley: This occurs in both 26.9.3 and 26.11.5 so reinstated twice.

lib/matobj2.gd
fingolfin: I agree that your new text is better! On the technical level, I guess maybe `<M>v</M><A>M</A>` would be correct? @ThomasBreuer ?

cdwensley: Done

ThomasBreuer: Yes, `<M>v</M><A>M</A>` is correct.

ThomasBreuer (approved):
Thanks.

PR #4539, opened 2021-06-06
kernel: move Default{In,Out}put from tls.h to io.c

fingolfin:
... and rename them to `DefaultOutputStream` resp. `DefaultInputStream`,
to avoid confusion with `IO()->DefaultOutput`.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -97,6 +97,10 @@ struct IOModuleState {
 
     TypOutputFile DefaultOutput;
 
+#ifdef HPCGAP
+    Obj DefaultOutputStream;
+    Obj DefaultInputStream;
+#endif
 
     Int NoSplitLine;
 };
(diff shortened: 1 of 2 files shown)


ssiccha (commented):
This previously was in a `ThreadLocalStorage`. So does that mean that each thread stored its own version of `DefaultIn/Outputstream`? Would making this global break something?

fingolfin:
It still is in TLS; the GAP state and all module states are stored in TLS. See https://github.com/gap-system/gap/blob/master/src/gapstate.h#L36

wilfwilson (approved):

src/io.c
```diff
@@ -97,6 +97,8 @@ struct IOModuleState {
 
     TypOutputFile DefaultOutput;
 
+    Obj DefaultOutputStream;
```
fingolfin, line 100: Actually This should be inside "ifdef HPCGAP"

fingolfin: done

PR #4540, opened 2021-06-06
kernel: better keep track of stdout/errout state

### LINKED ISSUE #4534: Error message not being wrapped to the right length
This is happening for me in the master branch, on a Mac, with readline. It seems as if the output of `Error` is being wrapped as if it didn't have the 7 characters of "`Error, `" prepended to it:
```
$ bin/gap.sh -A -r
 ┌───────┐   GAP 4.12dev-909-gca1c659 built on 2021-06-04 10:33:48+0100
 └───────┘   Architecture: x86_64-apple-darwin20.4.0-default64-kv8
 Packages:   GAPDoc 1.dev, PrimGrp 3.4.1, SmallGrp 1.4.2, TransGrp 3.3
gap> str := "this is an error message that going to be more than 80 character long although my terminal is only 80 characters wide";
"this is an error message that going to be more than 80 character long althoug\
h my terminal is only 80 characters wide"
gap> Length(str);
117
```
So the string itself wrapped as I would expect. But:
```
gap> Error(str);
Error, this is an error message that going to be more than 80 character long although\
 my terminal is only 80 characters wide called from
not in any function at *stdin*:8
```
which is wrapped too long. It seems like this only affects the first line:
```
gap> str := "this is an error message that going to be more than 160 character long although my terminal is only 80 characters wide. I will do this by adding some nonsense to the end";
"this is an error message that going to be more than 160 character long althou\
gh my terminal is only 80 characters wide. I will do this by adding some nonse\
nse to the end"
gap> Length(str);
169
gap> Error(str);
Error, this is an error message that going to be more than 160 character long althoug\
h my terminal is only 80 characters wide. I will do this by adding some nonsen\
se to the end called from
not in any function at *stdin*:15
```

## Also... 
I'm not sure if this is exactly related, but if I add this example to a test file, then GAP happily accepts the correct behaviour. Contents of test file:
```
gap> str := "this is an error message that going to be more than 80 character long although my terminal is only 80 characters wide";;
gap> Length(str);
117
gap> Error(str);
Error, this is an error message that going to be more than 80 character long a\
lthough my terminal is only 80 characters wide
```
Result of test:
```
gap> Test("width.tst");
true
```

But if I break the error message then the suggested correct version looks very weird. Contents of test file (last word changed):
```
gap> str := "this is an error message that going to be more than 80 character long although my terminal is only 80 characters wide";;
gap> Length(str);
117
gap> Error(str);
Error, this is an error message that going to be more than 80 character long a\
lthough my terminal is only 80 characters tall
```
Result of test:
```
gap> Test("width.tst");
########> Diff in width.tst:4
# Input is:
Error(str);
# Expected output:
Error, this is an error message that going to be more than 80 charac\
ter long a\
lthough my terminal is only 80 characters tall
# But found:
Error, this is an error message that going to be more than 80 charac\
ter long a\
lthough my terminal is only 80 characters wide
########
false
```

fingolfin:
Fixes #4534. See also https://github.com/gap-system/gap/pull/4496#discussion_r634701750 and https://github.com/gap-system/gap/issues/1197#issuecomment-299159780

The GAP kernel keeps track of the current output device for printing via
a stack of output "files", represented by structs of type
`TypOutputFile`. For each open file, GAP keeps track of its state (such
as the current cursor position, in such a struct.

However, the standard output is somewhat unique here in that it may be
opened multiple times (both as "stdout" and also as "errout"). But so
far GAP did never synchronize the state between the potentially multiple
structs tracking the state of stdout/errout. This leads to weird
printing effects, such as GAP printing line continuations in unexpected
places (too later or too early). This was exacerbated by the fact that
printing to errout in `lib/error.g` happens via calls to
`PrintTo(ERROR_OUTPUT, ...)` which actually opens and then immediately
again closes errout; so tracking any information on cursor position or
hints for word wrapping would not be carried forward correctly.

Moreover, flushing the output buffer via `\c` incorrectly reset the
presumed cursor position to 0; this was because the same variable `pos`
was used to track the cursor position, and the current offset into the
line buffer.

This patch fixes all that, by doing two things: First, whenever we open
or close a `TypOutputFile` referencing stdout or errout, we synchronize
it with the previous `TypOutputFile` referencing stdout or errout, if
any. Secondly, we add a new field to `TypOutputFile` which keeps track
of the cursor position separately from the line buffer state.

This revealed a bug where parts of the GAP banner suddenly were wrapped
unexpectedly. This in turn is caused by us printing some unicode
characters in the banner (to draw a nice ASCII art box around the word
GAP); but since these unicode characters each take up more than one
byte, this threw of our cursor tracking code. The reason this worked
before and broke now is that the code printing the box actually abused
the fact that flushing the line buffer by printing `\c` also reset the
tracked cursor position. But we just fixed this... So as a new
workaround, we temporarily disable output formatting for stdout.

Not sure if this warrants a release notes entry.


lib/init.g
```diff
@@ -466,10 +466,17 @@ BindGlobal( "ShowKernelInformation", function()
...
+  # possible use of unicode characters in btop and bbot which confuses the
+  # code in io.c tracking the current cursor position: those Unicode
+  # characters consist of multiple bytes each, which that code does not take
+  # into account.
+  SET_PRINT_FORMATTING_STDOUT(false);
   Print( " ",btop,"   ",gap," ", GAPInfo.BuildVersion,
          sysdate, "\n",
          " ",vert,"  GAP  ",vert,"   https://www.gap-system.org\n",
          " ",bbot,"   Architecture: ", GAPInfo.Architecture, "\n" );
+  SET_PRINT_FORMATTING_STDOUT(true);
```
fingolfin, line 479: Of course it would be better to save and restore the current state -- but that requires PR #4526.

fingolfin:
CC @zickgraf

src/io.c
```diff
@@ -1102,6 +1150,11 @@ static void PutLine2(TypOutputFile * output, const Char * line, UInt len)
     }
     else {
         SyFputs(line, output->file);
```
fingolfin, line 1152: Actually ... this ignores `len` which looks like a potential bug? Need to investigate

src/io.c
```diff
@@ -1102,6 +1150,11 @@ static void PutLine2(TypOutputFile * output, const Char * line, UInt len)
...
+        if (output->file == 1 || output->file == 3) {
+            // FIXME: this does not work right for unicode output,
+            // but we don't handle this right anyway, so...
+            output->column += output->pos;
```
fingolfin, line 1156: I guess I could (should?) drop the stdout/errout check...

also, this (although it has no effect in practice, but it may seem less surprising to anybody reading the code)
```suggestion
            output->column += len;
```

fingolfin:
The diff in testspecial looks like a real regression / opportunity for improvement:
```patch
--- a/tst/testspecial/bugfix-2019-09-27-LastPV.g.out
+++ b/tst/testspecial/bugfix-2019-09-27-LastPV.g.out
@@ -6,11 +6,10 @@ gap> SetFilterObj(badgroup, filt);
 gap> old_OnBreak:=OnBreak;;
 gap> OnBreak:=fail;; # prevent backtrace with line numbers that keep chaning
 gap> View([[badgroup]]); # <- trigger break loop
-Error, Rational operations: <divisor> must not be zero in
-  0 / 0 at *stdin*:3 called from
-type 'quit;' to quit to outer loop
-brk> quit;
-[ [
+[ [ Error, Rational operations: <divisor> must not be zero in
+          0 / 0 at *stdin*:3 called from
+        type 'quit;' to quit to outer loop
+        brk> quit;
 gap> OnBreak:=old_OnBreak;;
 gap> l := [1,2,3];
 [ 1, 2, 3 ]
```
So what happens is that `View` started to print something and then run into an error. In the past, we apparently just printed the error message over the partially printed output, and the partially printed output only appeared on the terminal after leaving the break loop. Actually that has been annoying me for some time and in the past I could not figure out why it happened. Now it is clear: It is still in the line buffer for stdout, not yet flushed; we then open `errout` to print the error. And the stdout buffer is only flushed when we resume regular operations.

One fix for this could be to force a flush of stdout/errout whenever we open a new stdout/errout (i.e. in `OpenOutput` before we call `copyOutputFileStruct` -- we want to make sure the output is correctly sent to the operating system's `stdout` or `stderr`). 

Alternatively else we could let `ErrorInner` start by checking if there is unprinted output; if so, flush it, and start a new line. Otherwise, do nothing. (We may have to add a kernel helper to achieve this).

I'll need to think more about this.

fingolfin:
Argh, this is a deeeeep rabbit hole... 

It turns out the problems go on: the GAPDoc function `PrintFormattedString` (which we also use in the GAP library to implement `Info`... *sigh*) relies on the bug I am fixing here (where `\c` resets the column counter). Since this PR also fixes `SetPrintFormattingStatus` for stdout, I can easily provide an implementation of `PrintFormattedString` which still works. But to let that have effect, I now need to monkey patch GAPDoc (i.e., I have to wait till it is loaded, then re-bind `PrintFormattedString` afterwards). Also, to make all of this work in tests, one needs a function which gets / sets the formatting status for the current output stream, whatever that is (this seems useful anyway). I'll add that, too. And then one can wonder if perhaps the print formatting status should be settable separately for `*stdout*` and `*errout*` ...?

wilfwilson:
I think this is beyond my level of ability to help/give sensible comment, sorry. Thank you for trying to deal with this though.

src/io.c
```diff
@@ -765,6 +765,29 @@ UInt CloseOutputLog ( void )
     return 1;
 }
 
+
+// Locate the previously opened stdout/stderr; return NULL if none found
+static TypOutputFile * findStdout(void)
+{
+    TypOutputFile * output = IO()->Output;
+    while (output && output->prev) {
+        if (!output->isstream && (output->file == 1 || output->file == 3))
```
fingolfin: So right now we treat `*stdout*` and `*errout*` as the same device in almost all ways; e.g. they share the same formatting status and line buffer. The only difference at this point really is whether we print to the system's `stdout` or `stderr`.

But perhaps we shouldn't do it like this; perhaps we should only sync any open instances with `*stdout*` with each other, and separately `*errout*`; the only thing to sync between *both* would be the cursor position.

But even there, we may need to be careful: what if `stdout` and `stderr` actually go to different devices? E.g. stdout might be connected to one terminal, but `stderr` to another (I just verified that this works fine).

Soooo... any syncing between `*stdout*` and `*errout*` state (whether it is all state, or just the cursor position) likely should also depend on whether the system's `stdout` and `stderr` point to the same device; that's a check we can already add to `InitSysFiles` I guess (all the relevant info is already available there).

ChrisJefferson:
I'm happy with the direction this is taking, and also think it will allow us to do other discussed things in future (like separating line wrapping + indenting). I had no idea `\c` could be use to reset the column position, and it isn't documented as doing that, so I don't think it's something we have to keep working. I agree with monkey-patching GAPDoc until it can be fixed (of course now, the function isn't so necessary anyway, with `*current*`).

src/io.c
fingolfin: No, this is OK, after inspecting the calling code (a comment should perhaps be added explaining this)

src/io.c
```diff
@@ -1929,6 +1990,28 @@ static Obj FuncPRINT_FORMATTING_CURRENT(Obj self)
     return output->format ? True : False;
 }
 
+// flush pending output on stdout/errout, and ensure we are
+// at the start of a new line. also reset any formatting hints
+static Obj FuncFLUSH_STDOUT_ERROUT(Obj self)
```
fingolfin, line 1995: This name is not quite accurate; we flush the output buffer, but we *also* move to the next line (but *only* if there is anything on the current line). If anybody has a better suggestion...

fingolfin:
From my point of view, this PR is ready. Perhaps @ChrisJefferson and @frankluebeck have time to take (another) look at it? Anybody else is welcome to do so as well, of course

wilfwilson (approved):
Re-reading all of this related stuff, and looking at the current state of the code, I do feel like I can say your PR looks like a reasonable solution and I'd be happy for it to be merged. My only questions:

* What's going on with the CI test diffs for the HPC-GAP job?
* Do you intend to address the `FIXME` and/or `HACK HACK HACK` comments that you add in this PR? If not, do you think there's any further explanation/hints that you can easily and sensibly add to these comments to help out whoever might address these things in the future? (I'm guessing the answer is that they are already clear enough - in which case fine 🙂).

frankluebeck:
Ad `\c`:
Sure, the documentation of `\c` is not very clear. But the interpretation given above is what was implemented and used for 30 years.
The change of this in this PR will break existing code in various places and packages. (For example, `Test` contains a counter similar to my example above, which is broken by this PR.) 
I repeat: I would leave the behaviour of `\c` as it always was, and support the introduction of a "true flush character".

Ad mixing `"*stdout*"` and `"*errout*"`:
If someone writes code that prints information on stdout then this code is probably written with some useful formatting of the output in mind. 
Printing to errout is likely done by some calls of `Error...` in case of an error. If this error function prints some lengthy information this is also programmed with some sensible formatting in mind. 
So, if stdout and errout go to different files/streams we can expect that both have a nice formatting. 
On the other hand, if the two go to the same stream, typically a terminal, it is to be expected that the formatting of the complete output is not optimal. I don't think the common column counter for both streams as in this PR is the desirable behaviour.

wilfwilson:
@fingolfin Do you have any thoughts about how/whether to continue with this PR?

Issue #235, opened 2015-09-21
InstallFalseMethods etc.


mohamed-barakat:
This issue is just a reminder for me. Once we have completed profiling and see how much time we spend doing logic I'll come back with more details:

@stevelinton: When I visited St Andrews some years ago I asked  if it would be possible to have InstallFalseMethods as well. I remember that you started something (even more elaborate) back then. Now that the new category theory project CAP of Sebastians is extensively using logic more than homalg it would be nice to reconsider this.

fingolfin:
There was a proposal for this on GAP devel in 2011. Steve proposed to add the following new API:
```
InstallExtendedImplication(
  <and  of flags that must be true>,
  <list of filters that must all be false>,
  <and of flags to set true>)
```
For an example:
```
InstallExtendedImplication(
    IsGroup and HasIsNilpotentGroup,
    [IsNilpotentGroup],
    HasIsAbelian)
```
Some notes:
1. The entries of the second argument likely should be simple flags, to avoid amibguities for the reader (as those filters are implicitly negated, so any "and" turns into an "or", etc. -- if you want an "or" effect on those filters, you can invoke `InstallExtendedImplication` multiple times.
2. No filters are set to "false" by it. That is sufficient, because if one e.g. sets `HasIsAbelian` to true, but does not set `IsAbelian` to true, then in fact `IsAbelian` returns false (as its corresponding "bit" in the flag has not been set).

fingolfin:
There are many cases where this would be useful. E.g. as soon as `IsFinitelyGenerated` is known to be *false* for a group, we can conclude a lot of things about it (e.g. is infinite and has infinite rank).

hungaborhorvath:
If I am not mistaken, this can already be done. 
````
InstallImmediateMethod(Size, IsGroup and HasIsFinitelyGenerated, 1000
function(G)
  if IsFinitelyGenerated(G) then
    TryNextMethod();
  else
    return infinity;
  fi;
end);
````
For me having InstallFalseMethod (or whatever name) would just make things complicated. There must be a good reason why it is done this way, but for me allowing to use the logical expressions (not only ````and```` of ````true```` expressions) would be the way to go. But I guess this would need a lot of rewriting.

fingolfin:
What you propose does not work. It would trigger a size computation as soon as  group learns that it is finitely generated. But such a computation need not terminate.

[...]

PR #4546, opened 2021-06-07
Improve property deduction for simple groups

fingolfin:
- change various constructors for simple groups to also indicate the
  constructed groups are nonabelian
- add some more automatic implications: nonabelian simple groups are
  also non-nilpotent and non-solvable

Now `IsSolvableGroup(SuzukiGroup(2^9));` returns instantly, as GAP
already knows the answer.

CC @jackschmidt

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -629,8 +629,13 @@ InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
 ##
 DeclareProperty( "IsSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup and IsNonTrivial, IsSimpleGroup );
+
 DeclareSynonymAttr( "IsNonabelianSimpleGroup", IsSimpleGroup and IsPerfectGroup );
+# ... implies not abelian, not nilpotent (also not solvable, but
+# HasIsSolvableGroup only gets defined later)
 InstallTrueMethod( HasIsAbelian, IsNonabelianSimpleGroup );
+InstallTrueMethod( HasIsNilpotentGroup, IsNonabelianSimpleGroup );
+#InstallTrueMethod( HasIsSolvableGroup, IsNonabelianSimpleGroup );
 
 InstallIsomorphismMaintenance( IsSimpleGroup, IsGroup, IsGroup );
 
@@ -812,7 +817,11 @@ InstallFactorMaintenance( IsSolvableGroup,
 InstallTrueMethod( IsSolvableGroup, IsMonomialGroup );
 InstallTrueMethod( IsSolvableGroup, IsSupersolvableGroup );
 
+# nontrivial solvable groups are not perfect
 InstallTrueMethod( HasIsPerfectGroup, IsGroup and IsSolvableGroup and IsNonTrivial );
+
+# nonabelian simple groups are not solvable
+InstallTrueMethod( HasIsSolvableGroup, IsNonabelianSimpleGroup );
 
 
 #############################################################################
--- a/grp/basicprm.gi
+++ b/grp/basicprm.gi
@@ -134,8 +134,7 @@ function( filter, dom )
         SetMovedPoints( alt, dom );
         SetNrMovedPoints( alt, Length(dom) );
         if 4 < Length(dom)  then
-            SetIsSimpleGroup(  alt, true );
-            SetIsPerfectGroup( alt, true );
+            SetIsNonabelianSimpleGroup( alt, true );
         elif 2 < Length(dom)  then
             SetIsPerfectGroup( alt, false );
         fi;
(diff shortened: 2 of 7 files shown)


lib/grp.gd
```diff
@@ -629,8 +629,13 @@ InstallTrueMethod( IsPerfectGroup, IsGroup and IsTrivial );
 ##
 DeclareProperty( "IsSimpleGroup", IsGroup );
 InstallTrueMethod( IsGroup and IsNonTrivial, IsSimpleGroup );
+
 DeclareSynonymAttr( "IsNonabelianSimpleGroup", IsSimpleGroup and IsPerfectGroup );
+# ... implies not abelian, not nilpotent (also not solvable, but
+# HasIsSolvableGroup only gets defined later)
 InstallTrueMethod( HasIsAbelian, IsNonabelianSimpleGroup );
+InstallTrueMethod( HasIsNilpotentGroup, IsNonabelianSimpleGroup );
```
wilfwilson, line 637: I have not seen this technique used before. Can you just confirm just
```
InstallTrueMethod( HasIsX, IsY );
```
means that objects `o` for which `IsY(o) = true` becomes known then immediately have the knowledge that `IsX(o) = false`?

Is this because the "default" value of a property is somehow `false`, and that setting the tester for the property to be `true` basically means that this "default" value becomes the accepted truth?

fingolfin: No, that's not what happens. If `IsY(o) = true` then it stays true.

But if that happens, you are in an inconsistent state. It typically happens because an overeager user invokes `Set IsNonabelianSimpleGroup(G,true)` on a group that is not actually nonabelian simple. As with other footguns in GAP, we don't attempt to do anything about it.

We *could* in theory at least set a flag that indicate an inconsistency. I guess that's what the `IsImpossible` attribute is meant for (off-side: why is that an attribute, not a property?!). But to use it, one currently needs to add immediate methods, which incurs a performance penalty. Or one makes a generic such method, but then gets less helpful errors (just that an inconsistency occurred, but not how)

jackschmidt:
Thanks! I don't see how to add a new file to the "suggested changes" in the review. But the silly `SetIsSolvableGroup` in:

https://github.com/gap-system/gap/blob/682f8a4fca830d0c024d8beafb474c863764b8a0/tst/teststandard/ctblmono.tst#L29

can be removed in this commit too (after rebasing it off a more recent development version). A simple:

    gap> IsMinimalNonmonomial(Sz(512));

should work now, thanks to this commit :tada:

lib/grp.gd
jackschmidt: I don't quite answer @wilfwilson 's question, but here is a demonstration that GAP does get the answer false if it doesn't have any methods:

```
DeclareProperty("IsInteresting",IsGroup);;
DeclareProperty("IsVeryInteresting",IsGroup);
DeclareProperty("IsBoring",IsGroup);

# Every very interesting group is interesting
InstallTrueMethod(IsInteresting, IsVeryInteresting);

# An interesting group is not boring
InstallTrueMethod(HasIsBoring, IsInteresting);
```

test it:

```
gap> g := SL(2,3);;
gap> SetIsVeryInteresting(g,true);
gap> HasIsBoring(g);
true
gap> IsBoring(g);
false
```

fingolfin: Ah, I see now what you were getting at, sorry, I misunderstood. So the point is that GAP stores filters as bits in a huge bitmask stored in the type of any object. So if an object is in that filter, that bit is set, otherwise it is not. 

For properties, there are two filters/flags, and hence two bits: one stores whether the value is known or not; the other stores the actual value.

Initially all flag/filter bits are zero. So when we set `HasIsFOO` to true, but do not touch `IsFOO` , we leave the latter at its default value, which is 0 / false.

It would certainly be nice to have a nicer API for this, an `InstallFalseMethod` if you will (see issue #235 where exactly this was requested in 2015 -- but @mohamed-barakat already requested this way earlier on the GAP mailing list, going back to at least 2011)

But in a nutshell, the GAP implication system is too primitive to provide a general `InstallFalseMethod` right now. But what we could do is provide a version that supports some special cases; namely `InstallFalseMethod(IsFOO, cond)` where `IsFOO` is a "simple" property, i.e., declared via `DeclareProperty`, and *not* the result if AND-ing multiple properties. Because for those, we can do `SetIsFOO(obj, false)`. We can't do that for filters like `IsAbelian and IsFinite` in general, because what should that do? It could mean the group is not abelian, or it could mean it is not finite, or both (well, one *could* handle some more cases if one wanted, by translating e.g. `InstallFalseMethod(IsAbelian and IsFinite, req);` to 
```gap
InstallTrueMethod(HasIsAbelian, req and IsFinite);
InstallTrueMethod(HasIsFinite, req and IsAbelian);
```
but getting there (and making sure one does not screw up) requires some non-trivial amount of work; not sure if it's really worth it? (It would be  useful if one also implemented the rest requested in issue #235, namely support for "extended" implications, but that's a much bigger change)

wilfwilson: Thanks for all that, those first few paragraphs are exactly the info that I was missing.

wilfwilson (approved):

PR #4552, opened 2021-06-08
Update license info for future Windows executables

### LINKED ISSUE #4543: Update licensing information for the new Windows installers
In #4389, I added a GitHub Actions workflow which uses https://github.com/ChrisJefferson/sage-windows to automatically create 32- and 64-bit Windows installers whenever a GAP release is made, and upload the installers to the GitHub release.

As far as I am aware (although perhaps I am missing something), the idea is that for GAP 4.12.0 onwards, this is the system that we will use for making Windows installers, and the old method of @alex-konovalov will stop being used.*

_*I seem to recall @ChrisJefferson mentioning that the new installers will only work on Windows 10. Is this right? If so, we should explicitly agree and announce that we are dropping supporting for older Windows versions. Without a Windows installation (new or old), I have no way of checking this myself._

Therefore, we most likely need to update the information about the Windows installers in the manual. This is what is currently written in [Section 1.4 "Copyright and License"](https://www.gap-system.org/Manuals/doc/ref/chap1.html#X7950EFA183E3F666):

> The executable of GAP for Windows that we distribute was compiled with the gcc compiler supplied with Cygwin installation (https://cygwin.com/).
>
> The GNU C compiler is
>
> Copyright © 2010 Free Software Foundation, Inc.
>
> under the terms of the GNU General Public License (GPL).
>
> The Cygwin API library is also covered by the GNU GPL. The executable we provide is linked against this library (and in the process includes GPL'd Cygwin glue code). This means that the executable falls under the GPL too, which it does anyhow.
>
> The `cyggcc_s-1.dll`, `cygncurses-10.dll`, `cygncursesw-10.dll`, `cygpanel-10.dll`, `cygpopt-0.dll`, `cygreadline7.dll`, `cygstart.exe`, `cygwin1.dll`, `libW11.dll`, `mintty.exe` and regtool.exe` are taken unmodified from the Cygwin distribution. They are copyright by RedHat Software and released under the GPL. For more information on Cygwin, see https://www.cygwin.com.

How much of this is still relevant? Does some of it need removing? Does some different information need adding? I am not able to answer these questions.

wilfwilson:
I basically adopted Chris's suggestion at https://github.com/gap-system/gap/issues/4543#issuecomment-856092232. I have one question which I'll make on the diff.

Closes #4543.

### DIFF
--- a/doc/ref/copyright.xml
+++ b/doc/ref/copyright.xml
@@ -66,16 +66,6 @@ make.
 In addition to the general copyright for &GAP; set forth above,
 the following terms apply to the versions of &GAP; for Windows.
 <P/>
-The executable of &GAP; for Windows that we distribute was compiled with
-the <C>gcc</C> compiler supplied with <Package>Cygwin</Package> installation
-(<URL>https://cygwin.com/</URL>).
-<P/>
-The GNU C compiler is
-<P/>
-<E>Copyright &copyright; 2010 Free Software Foundation, Inc.</E>
-<P/>
-under the terms of the GNU General Public License (GPL).
-<P/>
 The <Package>Cygwin</Package> API library is also covered by the GNU GPL. 
 The executable we provide is linked against this library (and in the process 
 includes GPL'd <Package>Cygwin</Package> glue code). This means that the 
--- a/doc/ref/copyright.xml
+++ b/doc/ref/copyright.xml
@@ -66,18 +66,13 @@ make.
 In addition to the general copyright for &GAP; set forth above,
 the following terms apply to the versions of &GAP; for Windows.
 <P/>
-The <Package>Cygwin</Package> API library is also covered by the GNU GPL. 
-The executable we provide is linked against this library (and in the process 
-includes GPL'd <Package>Cygwin</Package> glue code). This means that the 
-executable falls under the GPL too, which it does anyhow.
-<P/>
-The <F>cyggcc_s-1.dll</F>, <F>cygncurses-10.dll</F>, <F>cygncursesw-10.dll</F>,
-<F>cygpanel-10.dll</F>, <F>cygpopt-0.dll</F>, <F>cygreadline7.dll</F>,
-<F>cygstart.exe</F>, <F>cygwin1.dll</F>, <F>libW11.dll</F>, <F>mintty.exe</F> 
-and <F>regtool.exe</F> are taken unmodified from the 
-<Package>Cygwin</Package> distribution. They are copyright by RedHat Software
-and released under the GPL. For more information on <Package>Cygwin</Package>,
-see <URL>https://www.cygwin.com</URL>.
+&GAP; for Windows is packaged with several packages from
+<Package>Cygwin</Package> (<URL>https://cygwin.com</URL>).
+Files from <Package>Cygwin</Package> are provided under their respective
+licenses, which are included in the installation.
+The &GAP; for Windows executables that we distribute are themselves released
+under the terms of the GNU General Public License (GNU GPL);
+either version 2 of the License, or (at your option) any later version.
 <P/>
 Please contact <Email>support@gap-system.org</Email> if you need further information.
 


doc/ref/copyright.xml
```diff
@@ -66,28 +66,20 @@ make.
 In addition to the general copyright for &GAP; set forth above,
 the following terms apply to the versions of &GAP; for Windows.
 <P/>
-The executable of &GAP; for Windows that we distribute was compiled with
+&GAP; for Windows is packaged with several packages from
+<Package>Cygwin</Package> (<URL>https://cygwin.com</URL>).
+Files from <Package>Cygwin</Package> are provided under their respective
+licenses (included).
+The &GAP; for Windows executables that we distribute are released under
```
wilfwilson, line 73: Note that I've pluralised 'executable' here since we intend to offer 32-bit and 64-bit versions.

alex-konovalov: Fine, and I am happy with the plural anyway (e.g. we may have bat + exe + dll files even for a single 64-bit mode)

doc/ref/copyright.xml
```diff
@@ -66,28 +66,20 @@ make.
...
-executable falls under the GPL too, which it does anyhow.
-<P/>
-The <F>cyggcc_s-1.dll</F>, <F>cygncurses-10.dll</F>, <F>cygncursesw-10.dll</F>,
-<F>cygpanel-10.dll</F>, <F>cygpopt-0.dll</F>, <F>cygreadline7.dll</F>,
-<F>cygstart.exe</F>, <F>cygwin1.dll</F>, <F>libW11.dll</F>, <F>mintty.exe</F> 
-and <F>regtool.exe</F> are taken unmodified from the 
-<Package>Cygwin</Package> distribution. They are copyright by RedHat Software
-and released under the GPL. For more information on <Package>Cygwin</Package>,
-see <URL>https://www.cygwin.com</URL>.
+under the terms of the GNU GPL.
```
alex-konovalov, line 82: Also say which version?

doc/ref/copyright.xml
```diff
@@ -66,28 +66,20 @@ make.
 In addition to the general copyright for &GAP; set forth above,
 the following terms apply to the versions of &GAP; for Windows.
 <P/>
-The executable of &GAP; for Windows that we distribute was compiled with
+&GAP; for Windows is packaged with several packages from
+<Package>Cygwin</Package> (<URL>https://cygwin.com</URL>).
+Files from <Package>Cygwin</Package> are provided under their respective
+licenses (included).
```
alex-konovalov, line 72: What does "included" mean here?

wilfwilson: I don't know, we can ask @ChrisJefferson – this is part of the text I took from his suggestion https://github.com/gap-system/gap/issues/4543#issuecomment-856092232

doc/ref/copyright.xml
wilfwilson: I'm pretty sure it's version 3, so I've changed it to that, but I'm not 100% sure (perhaps it is v3 or later?). Since we're not including any of the gcc source code (as far as I'm aware) I don't think we actually _need_ to be talking about GCC here. That said I'm happy to mention it.

doc/ref/copyright.xml
```diff
@@ -66,28 +66,21 @@ make.
...
-executable falls under the GPL too, which it does anyhow.
-<P/>
-The <F>cyggcc_s-1.dll</F>, <F>cygncurses-10.dll</F>, <F>cygncursesw-10.dll</F>,
-<F>cygpanel-10.dll</F>, <F>cygpopt-0.dll</F>, <F>cygreadline7.dll</F>,
-<F>cygstart.exe</F>, <F>cygwin1.dll</F>, <F>libW11.dll</F>, <F>mintty.exe</F> 
-and <F>regtool.exe</F> are taken unmodified from the 
-<Package>Cygwin</Package> distribution. They are copyright by RedHat Software
-and released under the GPL. For more information on <Package>Cygwin</Package>,
-see <URL>https://www.cygwin.com</URL>.
+under the terms of the GNU GPL, version 3.
```
fingolfin, line 83: Wow, I never was aware of this text. Why is this even in there?? I see no reason, can someone elaborate?

Otherwise, let's just remove it all?

wilfwilson: No idea, I can’t see why either.

fingolfin: OK, I traced the origin of this text through history (it wasn't easy; it move through several files in multiple commits; often being added to a new file in one commit, and then removed from the old place in another, later commit). Ultimately, the oldest place introducing it I could find is in a commit from 1998-07-16 with the commit message "Merge" -- it seems a (CVS?) development branch was being merged, but no separated history was recorded. Could be due to a bug in my repository conversion back then, or perhaps the development happened outside CVS, I can't say. But I am pretty sure it means the reasons are lost in the mists of time. So I say: let's just remove it.

Of course we can also wait a bit and see if @alex-konovalov or @ChrisJefferson have anything to add

alex-konovalov: Agree, let's remove it.

doc/ref/copyright.xml
alex-konovalov: Ah, ok

doc/ref/copyright.xml
alex-konovalov: @ChrisJefferson indeed, looks to me like we include all licenses for all types represented there (?) - I am not sure we do.

doc/ref/copyright.xml
ChrisJefferson: I don't think it's necessary either -- my suggested fix was a minimal version which I felt handled the same requirement.

doc/ref/copyright.xml
ChrisJefferson: By included I meant (included in the release) -- as we now package a full cygwin distro, it includes all the license files (as I don't go out of my way to remove them).

wilfwilson: Thanks Chris, I've changed "...licenses (included)" to "...licenses, which are included in the installation." Hope that's right.

alex-konovalov (approved):

fingolfin (approved):
Thanks!

PR #4553, opened 2021-06-08
Send Slack notifications for "Wrap releases" GitHub Actions jobs, too

wilfwilson:
I think that the Slack notifications code is now stable enough to use for the 'wrap releases' script too.

Perhaps @ssiccha has a better idea, but as far as I can see, in order to achieve this, we unfortunately have to duplicate the job in each Workflow file.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -254,3 +254,33 @@ jobs:
           python -u ./releases/upload_files_to_github_release.py v${VERSION} sage-windows/Output/gap-${{ env.VERSION }}-${{ matrix.arch }}.exe
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # The following job is duplicated in CI.yml - keep the two in sync.
+  # (except for their different 'needs' components).
+  slack-notification:
+    name: Send Slack notification on status change
+    needs:
+      - unix
+      - cygwin
+    if: ${{ always() && github.event_name != 'pull_request' && github.repository == 'gap-system/gap' }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get branch name
+        id: get-branch
+        run: echo ::set-output name=branch::${GITHUB_REF#refs/*/}
+      - name: Determine whether CI status changed
+        uses: gap-actions/should-i-notify-action@v1
+        id: should_notify
+        with:
+          branch: ${{ steps.get-branch.outputs.branch }}
+          needs_context: ${{ toJson(needs) }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -254,3 +254,33 @@ jobs:
...
+          needs_context: ${{ toJson(needs) }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          notify_on_changed_status: true
+      - name: Send slack notification
+        uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
+        if: ${{ steps.should_notify.outputs.should_send_message == 'yes' }}
+        with:
+          status: ${{ steps.should_notify.outputs.current_status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
```
fingolfin, line 286: We could also put all of this into an action at https://github.com/gap-actions -- after all, these slack notifications might of interest for packages, too? Or maybe not, dunno.

fingolfin (approved):
Looks good to me but perhaps @ssiccha should also comment

ssiccha:
Instead of creating a new repository for a new action, we can also create a new action locally. We put the commands into a file `.github/actions/slack-notification.yml` and use it like [this](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow),

.github/workflows/release.yml
wilfwilson: This isn't easy because composite actions (as those on gap-actions must be) do not support the `if` components, and they do not support the `uses` component. So for example the lines:
```
uses: act10ns/slack@e4e71685b9b239384b0f676a63c32367f59c2522
if: ${{ steps.should_notify.outputs.should_send_message == 'yes' }}
```
at least would become much more difficult to achieve.

ssiccha: Ah, true, they don't support `if`... Then IMO let's just keep a duplicate of the code for now. Instead of doing a composite action we could also do one of those javascript actions. I have no idea how to do that though.

wilfwilson:
Oh, interesting – and this would be as fully-fledged workflow, as opposed to a composite action? (Which would suffer from the problems I mentioned here https://github.com/gap-system/gap/pull/4553#discussion_r648078535) This is maybe something to look in to.

ssiccha:
I think that would also only be a composite action. :(

wilfwilson:
I would _really_ like composite actions to get more of the features of normal actions, it would make many things better, especially adapting all of our `gap-actions` stuff to work with Cygwin builds. See https://github.com/actions/runner/issues/646 for discussions about this.

PR #4556, opened 2021-06-09
 use the new definition of `IsMatrixObj` in methods

ThomasBreuer:
(This is a follow-up of #4518.)

In each method installation that takes an argument in `IsMatrixOrMatrixObj`
such that the function exits with `TryNextMethod()` if this argument is in `IsPlistRep`,
restrict the requirement to `IsMatrixObj` and omit the check for `IsPlistRep`.

Also removed some ancient (from GAP 3 times) comment lines,
and added a remark to the documentation that `IsRowListMatrix` need not imply `IsList`.


lib/matobj.gi
```diff
@@ -1346,14 +1273,8 @@ InstallMethod( AdditiveInverseSameMutability,
...
-    [ IsMatrixOrMatrixObj ],
-    function( M )
-    if IsPlistRep( M ) then
-      TryNextMethod();
-    else
-      return Matrix( ZeroMutable( Unpack( M ) ), M );
-    fi;
-    end );
+    [ IsMatrixObj ],
+    M -> Matrix( ZeroMutable( Unpack( M ) ), M ) );
```
fingolfin, line 1277: Should we perhaps instead delegate to ZeroMatrix?
```suggestion
    M -> ZeroMatrix( NrRows( M ), NrCols( M ), M ) );
```

lib/matobj.gi
```diff
@@ -1366,14 +1287,8 @@ InstallMethod( ZeroSameMutability,
...
-    [ IsMatrixOrMatrixObj ],
-    function( M )
-    if IsPlistRep( M ) then
-      TryNextMethod();
-    else
-      return Matrix( OneMutable( Unpack( M ) ), M );
-    fi;
-    end );
+    [ IsMatrixObj ],
+    M -> Matrix( OneMutable( Unpack( M ) ), M ) );
```
fingolfin, line 1291: Should we perhaps instead delegate to IdentityMatrix?
```suggestion
    M -> IdentityMatrix( NrRows( M ), M ) );
```

fingolfin (approved):

lib/matobj.gi
ThomasBreuer: Yes, thanks. (As a rule of thumb, avoiding `Unpack` is an improvement.)

PR #4557, opened 2021-06-10
Add `InstallEarlyMethod` which allows installing special methods that bypass method selection (and thus its overhead)

### LINKED ISSUE #955: Method dispatch idea: InstallEarlyMethod
One long-running problem in GAP is that method dispatch for built-in lists can be very expensive.

This leads to methods like `First`, which do a pre-check before calling `FirstOp`, which does the actual dispatch. I've played around to try to remove this cost, but it isn't possible in general without a complete change to how GAP does things.

Instead, here is an idea to make it easier to work around, a new function `InstallEarlyMethod`:

`InstallEarlyMethod(opr, method)` installs `method` for `opr`. `method` is called before any other installed method is considered. Note that `method` is called even for arguments which do not satisfy the filter requirements of `opr`.

fingolfin:
Resolves #955

I only converted `First` / `FirstOp` to use it as a proof of concept. Before more is done, I thought it would be good to solicit feedback. Also, since I am a bit short on time, perhaps others would like to help and contribute by converting more `Foo` / `FooOp` pairs to use this?

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -2811,38 +2811,40 @@ InstallMethod( Permuted,
 ##
 #F  First( <C>, <func> )  . . .  find first element in a list with a property
 ##
-InstallGlobalFunction( First,
-    function ( C, func... )
+InstallEarlyMethod( First, 
+    function ( C )
     local tnum, elm;
-    if Length( func ) > 1 then
-      Error( "too many arguments" );
-    fi;
     tnum:= TNUM_OBJ( C );
     if FIRST_LIST_TNUM <= tnum and tnum <= LAST_LIST_TNUM then
-      if Length( func ) = 0 then
-        func := ReturnTrue;
-      else
-        func := func[1];
-      fi;
       for elm in C do
-          if func( elm ) then
-              return elm;
-          fi;
+        return elm;
       od;
       return fail;
-    elif Length( func ) = 0 then
-      return FirstOp( C );
-    else
-      return FirstOp( C, func[1] );
     fi;
-end );
+    TryNextMethod();
+    end );
+
+InstallEarlyMethod( First,
+    function ( C, func )
+    local tnum, elm;
+    tnum:= TNUM_OBJ( C );
+    if FIRST_LIST_TNUM <= tnum and tnum <= LAST_LIST_TNUM then
+      for elm in C do
+        if func( elm ) then
+          return elm;
+        fi;
+      od;
+      return fail;
+    fi;
+    TryNextMethod();
+    end );
 
 
 #############################################################################
 ##
-#M  FirstOp( <C>, <func> )  . .  find first element in a list with a property
+#M  First( <C>, <func> ) . . . . find first element in a list with a property
 ##
-InstallMethod( FirstOp,
+InstallMethod( First,
     "for a list or collection and a function",
     [ IsListOrCollection, IsFunction ],
     function ( C, func )
@@ -2855,7 +2857,7 @@ InstallMethod( FirstOp,
     return fail;
     end );
 
-InstallMethod( FirstOp,
+InstallMethod( First,
     "for a list or collection",
     [ IsListOrCollection ],
     function ( C )
--- a/lib/oper1.g
+++ b/lib/oper1.g
@@ -362,6 +362,48 @@ BIND_GLOBAL( "InstallOtherMethod",
 
 #############################################################################
 ##
+#F  InstallEarlyMethod( <opr>,<method> )
+##
+##  <#GAPDoc Label="InstallEarlyMethod">
+##  <ManSection>
+##  <Func Name="InstallEarlyMethod" Arg="opr,method"/>
+##
+##  <Description>
+##  installs a special "early" function method <A>method</A> for the
+##  operation <A>opr</A>. An early method is special in that it bypasses
+##  method dispatch, and is always the first method to be called when
+##  invoking the operation.
+##  <P/>
+##  This can be used to avoid method selection overhead for certain special
+##  cases, i.e., as an optimization. Overall, we recommend to use this
+##  feature very sparingly, as it is tool with sharp edges: for example, any
+##  inputs that are handled by an early method can not be intercepted by a
+##  regular method, no matter how high its rank is; this can preclude other
+##  kinds of optimizations.
+##  <P/>
+##  Also, unlike regular methods, no checks are performed on the arguments.
+##  Not even the required filters for the operation are tested, so early
+##  methods must be careful in validating their inputs.
+##  This also means that any operation can have at most one such early
+##  method for each arity (i.e., one early method taking 1 argument, one
+##  early method taking 2 arguments, etc.).
+##  <P/>
+##  If an early method determines that it is not applicable, it can resume
+##  regular method dispatch by invoking <Ref Func="TryNextMethod"/>.
+##  <P/>
+##  For an example application of early methods, they are used by
+##  <Ref Oper="First"/> to deal with internal lists, for which computing
+##  the exact type (needed for method selection) can be very expensive.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+BIND_GLOBAL( "InstallEarlyMethod", INSTALL_EARLY_METHOD );
+#TODO; store location info somehow
+#MakeImmutable([INPUT_FILENAME(), READEVALCOMMAND_LINENUMBER, INPUT_LINENUMBER()])
+
+#############################################################################
+##
 #F  INSTALL_METHOD( <arglist>, <check> )  . . . . . . . . .  install a method
 ##
 DeclareGlobalFunction( "EvalString" );
--- a/tst/testinstall/kernel/opers.tst
+++ b/tst/testinstall/kernel/opers.tst
@@ -242,6 +242,42 @@ Error, SET_METHODS_OPERATION: <narg> must be an integer between 0 and 6 (not t\
 he integer 7)
 
 #
+gap> oper := NewOperation("foobar", [IsList]);
+<Operation "foobar">
+gap> INSTALL_EARLY_METHOD(fail, fail);
+Error, INSTALL_EARLY_METHOD: <oper> must be an operation (not the value 'fail'\
+)
+gap> INSTALL_EARLY_METHOD(oper, fail);
+Error, INSTALL_EARLY_METHOD: <func> must be a function (not the value 'fail')
+gap> INSTALL_EARLY_METHOD(oper, oper);
+Error, <func> must not be an operation
+gap> InstallMethod(oper, [IsPlistRep], function(l) Print("IsPlistRep method\n"); end);
+gap> oper([1,2,3]);
+IsPlistRep method
+gap> INSTALL_EARLY_METHOD(oper, function(l) Print("early method\n"); TryNextMethod(); end);
+gap> oper([1,2,3]);
+early method
+IsPlistRep method
+gap> INSTALL_EARLY_METHOD(oper, ReturnTrue);
+Error, <func> must not be variadic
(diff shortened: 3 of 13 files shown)


tst/testinstall/kernel/opers.tst
```diff
@@ -241,6 +241,24 @@ gap> SET_METHODS_OPERATION (Size,7,[]);
...
+Error, INSTALL_EARLY_METHOD: <func> must be a function (not the value 'fail')
+gap> INSTALL_EARLY_METHOD(oper, oper);
+Error, <func> must not be an operation
+gap> InstallMethod(oper, [IsPlistRep], function(l) Print("IsPlistRep method\n"); end);
+gap> oper([1,2,3]);
+IsPlistRep method
+gap> INSTALL_EARLY_METHOD(oper, function(l) Print("early method\n"); TryNextMethod(); end);
+gap> oper([1,2,3]);
+early method
+IsPlistRep method
```
wilfwilson, line 260:
```suggestion
IsPlistRep method
gap> INSTALL_EARLY_METHOD(oper, ReturnTrue);
Error, <func> must not be variadic
gap> INSTALL_EARLY_METHOD(oper, {a,b,c,d,e,f,g} -> fail);
Error, <func> must take at most 6 arguments
gap> INSTALL_EARLY_METHOD(oper, x -> fail);
Error, early method already installed
```

wilfwilson (commented):
Looks really nice to me. Here's a bit more code coverage.

ThomasBreuer (approved):
This is very elegant.

Later on, also `ApplicableMethod` and `MethodsOperation` should return also the early methods.
(This is no problem because `MethodsOperation` is currently undocumented --the records for early methods cannot have all the components that are available for ordinary methods.)

Concerning the way how to deal with the now unnecessary `FirstOp`, I think it can be moved to the obsolescent stuff, since it is documented only via an index entry and the explanation in the documentation of `First` (which should remain).

Concerning the ordinary generic methods for `First`, one could think about removing them, and instead dealing with these cases in the early methods.

lib/list.gi
```diff
@@ -2811,38 +2811,40 @@ InstallMethod( Permuted,
 ##
 #F  First( <C>, <func> )  . . .  find first element in a list with a property
 ##
-InstallGlobalFunction( First,
-    function ( C, func... )
+INSTALL_EARLY_METHOD( First,
```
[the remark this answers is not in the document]
fingolfin: Forgot to update it - I wrote the code as a first prototype and only later added `InstallEarlyMethod` (for potential future use)

fingolfin:
I made some improvements:
- better commit message
- added a "getter" function `EARLY_METHOD(oper,narg)` so that we can improve `ApplicableMethod`  to show early methods, and perhaps also `MethodsOperation`.
- add `InstallEarlyMethod` to `etc/ctags_for_gap`
- changed a test that referenced `FirstOp` to instead reference `First`.

@ThomasBreuer wrote:
> Concerning the way how to deal with the now unnecessary FirstOp, I think it can be moved to the obsolescent stuff, since it is documented only via an index entry and the explanation in the documentation of First (which should remain).

I moved it `lib/obsolete.gd`, but I can't change it to `DeclareObsoleteSynonym` because that would break the `fr` package, which installs a method for `FirstOp`.

> Concerning the ordinary generic methods for First, one could think about removing them, and instead dealing with these cases in the early methods.

I am not sure I understand. Do you mean the methods for `IsListOrCollection`? But then code implementing special kinds of lists or collections (as e.g. the `fr` package does) wouldn't be able to add specialized methods for `First`.

ThomasBreuer:
@fingolfin Forget about the last statement in my remark above, this was nonsense.

(These early methods are dangerous, everything that is handled by them is taken away from the method selection; they are like the "GAP kernel magic" that happens for internal objects.)

Thanks for dealing with `MethodsOperation`, I had not expected that this would be done already with the current pull request.
(I will have to adjust a function in the Browse package to the changed output.)
In fact, I can think of listing also the "GAP kernel magic" methods (or something equivalent to them) in the result of `MethodsOperation` --but not in this pull request.

lib/oper1.g
```diff
@@ -360,6 +360,40 @@ BIND_GLOBAL( "InstallOtherMethod",
...
+##  <P/>
+##  For example, we install early methods for <Ref Oper="First"/> that
+##  deal with internal lists, for which computing their types (and hence
+##  method selection) can be very expensive.
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+BIND_GLOBAL( "InstallEarlyMethod", INSTALL_EARLY_METHOD );
+#TODO; store location info somehow
```
ssiccha, line 394: What does this mean?

ssiccha (commented):
Should we introduce a helper function `EarlyMethodsOfOperation` so that one can conveniently check whether early methods are installed?

(I'm gonna have a look at the rest of the PR after lunch. Edit: had to help with the kid. I'll have a look later.)

lib/oper1.g
wilfwilson: I think it is a reminder to implement the idea in https://github.com/gap-system/gap/pull/4557#issuecomment-859024184 sometime (not necessarily in this PR).

ssiccha: Ah, I see. Thanks :)

fingolfin:
@ssiccha early methods are a dangerous specialized tool. To be honest, I am not even sure yet whether it's a good idea to document it in the manual (maybe we shouldn't?). There are already two ways for accessing them: (1) via `MethodsOperations`, (2) via `EARLY_METHOD` (for low-level use only). I don't think much code will ever have to access them; rather this will be mostly useful for debugging, and `MethodsOperation` / `ApplicableMethod` cover that. As such, I don't see what we gain by adding `EarlyMethodsOfOperation`. Also, what would that return, exactly (the plural seems to suggest that it would iterate over all arities and adds any early method it finds to a list? but what's the usecase for that?)

@ThomasBreuer wrote:
> These early methods are dangerous

Agreed, they are a sharp tool. Maybe we should mention this in their documentation. And maybe we should not (yet) hook the documentation up to the reference manual?

lib/oper1.g
fingolfin, line 394: Perhaps `InstallEarlyMethod` should also take an (optional) `info` string, so that one can e.g. write something like `for plain lists and a function`.

But then that also has to be stored somewhere. One solution for that: change `OPER(oper)->earlyMethod[n]` to point to a plist which contains the function, the location of the installation, and the info string. The result will have identical overhead to what is now in this PR when there is *no* early method; and only a small extra overhead in the rare case where there is one.

Thoughts on this, anyone?

lib/oper1.g
```diff
@@ -360,6 +360,40 @@ BIND_GLOBAL( "InstallOtherMethod",
...
+##  installs a special "early" function method <A>method</A> for the
+##  operation <A>opr</A>. An early method is special in that it bypasses
+##  method dispatch, and is always the first method to be called when
+##  invoking the operation. This can be used to avoid method selection
+##  overhead for certain special cases.
+##  <P/>
+##  Unlike with regular methods, no checks are performed on the
+##  arguments. Hence any operation can have at most one such early
+##  method for each arity (i.e., one early method taking 1 argument, one
+##  early method taking 2 arguments, etc.).
```
fingolfin, line 381: Somewhere here we should point out that early methods are dangerous and if they handle something, then the user cannot override this behaviour later one with regular method installations

lib/oper1.g
ssiccha: We could also add components `earlyMethodInfoString` and `earlyMethodInstallLocation`.

ssiccha:
Re `EarlyMethodsOfOperation`: yes, I was thinking of having this as a debugging tool, but as you said @fingolfin, `ApplicableMethod` and `MethodsOperation` already do everything we need. I think though that if there is an early method installed for an operation, then `ApplicableMethod` and `MethodsOperation` should make that as obvious as possible. `MethodsOperation` does that already IMO. `ApplicableMethod` should already at printlevel `1` make clear whether a method is early.

ssiccha:
I'm in favor of documenting this in the reference manual. If someone new to GAP decides to have a look at functions like `First`, then not having `InstallEarlyMethod` documented makes this harder to understand. It should definitely come with a big fat warning though.

We could not document this in v4.12.0, test it out a bit and add "Document InstallEarlyMethod" as an issue to a v4.12.1 milestone. If we don't document it now, I imagine we'll forget about it and never put it into the reference manual.

ssiccha (approved):
I like this. :+1:

ChrisJefferson:
Just as a sanity check, I did a bit of benchmarking on this, but (as I expected) it's basically impossible to get the time taken to show up.

lib/oper1.g
ChrisJefferson: I would be happy to just call these "early method", rather than record an extra string.

ChrisJefferson (approved):

lib/oper1.g
```diff
@@ -360,6 +360,40 @@ BIND_GLOBAL( "InstallOtherMethod",
...
+##  <#GAPDoc Label="InstallEarlyMethod">
+##  <ManSection>
+##  <Func Name="InstallEarlyMethod" Arg="opr,method"/>
+##
+##  <Description>
+##  installs a special "early" function method <A>method</A> for the
+##  operation <A>opr</A>. An early method is special in that it bypasses
+##  method dispatch, and is always the first method to be called when
+##  invoking the operation. This can be used to avoid method selection
+##  overhead for certain special cases.
```
ChrisJefferson, line 376: I wondered if we should add something to clarify that this operation is called even if the arguments to not satisfy the argument filters for the method (I realise technically there was never an explicit check for these, they are enforced in `InstallMethod`, but I thought it might be worth clarifying.

lib/oper1.g
fingolfin: they currently *are* called "early method", I just thought it might be nicer to see something like "early method for a plain list and a function"; given that no predicates are shown for an early method, so without this, the only way to have any idea what it does is to inspect its code. But yeah, it's certainly a secondary feature, just like tracking the location of the `InstallEarlyMethod` call (useful for debugging, but not crucial for the feature overall).

fingolfin:
@ssiccha wrote:
> ApplicableMethod should already at printlevel 1 make clear whether a method is early.

But it does? Or what are what would you like to see added here:
```
gap> ApplicableMethod(First, [ [1,2,3] ], 1);
#I  Searching Method for First with 1 arguments:
#I  Total: 2 entries
#I  Method 1: ``early method'' at /Users/mhorn/Projekte/GAP/gap.spielwiese/lib/list.gi:2815 , value: infinity
function( C ) ... end
```
Of course one could do it nicer and e.g. remove the `value: infinity` etc., but to me that's fine tuning, and someone else who wants it can submit a PR for it after this has been merged.

lib/oper1.g
fingolfin: I've now added some text for this.

lib/oper1.g
fingolfin: I've now added some text for this.

ThomasBreuer:
I think this pull request can be merged.

ssiccha:
> @ssiccha wrote:
> 
> > ApplicableMethod should already at printlevel 1 make clear whether a method is early.
> 
> But it does? Or what are what would you like to see added here:
> 
> ```
> gap> ApplicableMethod(First, [ [1,2,3] ], 1);
> #I  Searching Method for First with 1 arguments:
> #I  Total: 2 entries
> #I  Method 1: ``early method'' at /Users/mhorn/Projekte/GAP/gap.spielwiese/lib/list.gi:2815 , value: infinity
> function( C ) ... end
> ```
> 
> Of course one could do it nicer and e.g. remove the `value: infinity` etc., but to me that's fine tuning, and someone else who wants it can submit a PR for it after this has been merged.

I must have missed your code for `ApplicableMethod`.

PR #4558, opened 2021-06-10
Fix argument order of StoreHashTable

ChrisJefferson:
The arguments were passed to this function in the wrong order --
this greatly slowed GAP as the hash table would never match

Fixes a slowdown caused by #4554

### DIFF
--- a/src/opers.cc
+++ b/src/opers.cc
@@ -644,6 +644,7 @@ static Obj FuncAND_FLAGS(Obj self, Obj flags1, Obj flags2)
 template <Int len>
 static Obj LookupHashTable(Obj ht, Int hash, Obj flags)
 {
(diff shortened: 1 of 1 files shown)


src/opers.cc
```diff
@@ -654,7 +654,7 @@ static Obj LookupHashTable(Obj ht, Int hash, Obj flags)
 }
 
 template <Int len>
-static void StoreHashTable(Obj ht, Int hash, Obj new_flags, Obj new_with)
+static void StoreHashTable(Obj ht, Int hash, Obj new_with, Obj new_flags)
```
fingolfin, line 657: Thanks. Perhaps should also add GAP_ASSERT calls to verify the types of those arguments?

fingolfin (approved):
Thanks!

wilfwilson (approved):

PR #4562, opened 2021-06-14
Revise our pull request template

fingolfin:
The old one was far too long, meaning that most people would skip
over most of it. So let's focus on the important bits, which are
a useful description and release notes.

The stuff at the end for reviewers was not used by us in practice;
instead, there should be a separate document somewhere with guidelines
for reviewers.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,58 +1,13 @@
-Please use the following template to submit a pull request, filling
-in at least the "Text for release notes" and/or "Further details".
-Thank you!
+Please provide a short summary of this PR and its purpose here. E.g., does it add a new feature, and which? Does it fix a bug, and which? If there is an associated issue, please list it here.
 
-# Description
+## Text for release notes
 
-## Text for release notes 
+We track noteworthy changes as entries in the `CHANGES.md` file in the root directory of this repository. To help us decide whether and how to describe your PR, please do one of the following:
 
-If this pull request shall **not** be mentioned in the release notes
-(to be distributed in the file `CHANGES.md`),
-please add the label `release notes: not needed`.
-
-Otherwise, please proceed in one of the following ways:
-
-- Choose a title that can serve as text for the release notes,
-  and add the label `release notes: use title`.
-
-- Put the text for the release notes **here**,
-  that is, between the markers `Text for release notes`
-  and `(End of text for release notes)`.
-
-The first variant is recommended whenever the text for release notes
-is suitable as title.
-
-In both cases, please follow the style of the GAP `CHANGES.md` file
-in the root directory.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,58 +1,19 @@
-Please use the following template to submit a pull request, filling
-in at least the "Text for release notes" and/or "Further details".
-Thank you!
+# Summary
```
fingolfin, line 1: I am not sure this headline is a good idea, we could also just drop it...

ChrisJefferson (approved):

ChrisJefferson:
I think that's good -- I understand the wish to tell people to write tests, etc, but as you say I don't think putting that in the PR template really helps.

.github/pull_request_template.md
```diff
@@ -1,58 +1,19 @@
...
-is suitable as title.
-
-In both cases, please follow the style of the GAP `CHANGES.md` file
-in the root directory.
-In particular, please surround the names of GAP functions with backquotes.
-
-## (End of text for release notes)
+1. If this PR shall **not** be mentioned in the release notes, write "none" here.
+2. If a brief note suffices, edit the title of this PR to be usable as entry in `CHANGES.md`, and write "see title" here (or if you have the required permissions, add the label `release notes: use title` to this PR and remove this section)
+3. If a longer note is needed,just write it here, ideally following the general style used in that file
```
wilfwilson, line 15:
```suggestion
3. If a longer note is needed, just write it here, ideally following the general style used in that file
```

.github/pull_request_template.md
wilfwilson: Happy to get rid of it and include 'here' in the first sentence.

wilfwilson (approved):
Perhaps there are good reasons to do this that I'm not aware of. But I think the lines of prose text should not be manually wrapped to a particular length - since this text will be viewed in a browser (which is not usually in a fixed-width terminal these days)/

alex-konovalov (approved):

fingolfin:
@wilfwilson good point on the line wrapping; done now.

fingolfin:
I am cancelling the tests for this PR, they are clearly pointless :-)

ThomasBreuer:
I just made a pull request, and noticed that the template had changed.
Weren't the markers at the beginning and the end of the text for the release notes intended to automatically extract the text?
(If yes then shouldn't the code in question be adjusted?)
And why is it an improvement to remove the hint about the `release notes: not needed` label?

fingolfin:
The scripts do not (yet) do that.

labels can only be set by people with the required permissions, not regular users. The former should get (and use) a more extensive guideline on labels. But there is no point in confusing new contributors with information about things they can not affect

PR #4563, opened 2021-06-14
Permutation representation of extensions with reducible modules

hulpke:
Fixed a bug (in updating data structure) if a permutation representation of an extension needs several steps (can only happen for reducible modules).
Also made `ClosureSubgroup` not force a triviality test if not known, and expanded an info message

### DIFF
--- a/tst/testextra/perfect.tst
+++ b/tst/testextra/perfect.tst
@@ -10,4 +10,16 @@ gap> Length(l);
 gap> l:=Practice(10752);;
 gap> Length(l);
 9
+gap> LoadPackage("atlasrep");;
+gap> perms:=AtlasGenerators("2.A5",1).generators;;
+gap> mats:=AtlasGenerators("2.A5",4).generators;;
+gap> gp:=Group(perms);;
+gap> mo:=GModuleByMats(mats,GF(2));;
+gap> coh:=TwoCohomologyGeneric(gp,mo);;
+gap> Length(coh.cohomology);
+1
+gap> gp:=FpGroupCocycle(coh,coh.cohomology[1],true);;
+gap> gp:=Image(IsomorphismPermGroup(gp));;
+gap> Size(Group(GeneratorsOfGroup(gp)));
+61440
 gap> STOP_TEST( "perfect.tst", 1);
--- a/tst/testinstall/grpfp.tst
+++ b/tst/testinstall/grpfp.tst
@@ -1,4 +1,4 @@
-#@local a,b,c2,e,f,g,iter,l,s,F,rels
+#@local a,b,c2,e,f,g,iter,l,s,F,rels,sub
 gap> START_TEST("grpfp.tst");
 gap> f:= FreeGroup( "a", "b" );;  a := f.1;;  b := f.2;;
 gap> c2:= f / [ a*b*a^-2*b*a/b, (b^-1*a^3*b^-1*a^-3)^2*a ];;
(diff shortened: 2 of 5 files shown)


tst/testextra/perfect.tst
```diff
@@ -10,4 +10,16 @@ gap> Length(l);
 gap> l:=Practice(10752);;
 gap> Length(l);
 9
+gap> LoadPackage("atlasrep");
```
jackschmidt, line 13: This line of the test fails. It just needs a second semi colon:
`gap> LoadPackage("atlasrep");;`

jackschmidt (changes requested):
(Only one change requested, but I still don't know how to edit -- missing semicolon in the test_

This does appear to fix the perfect group extension with reducible module (which worked in a fairly recent dev version, but not the last few).

The new test takes about 2 seconds but needs internet access. The new `lib/twocohom.gi` lines are in fact tested in testextra.

The new test does have a harmless typo (missing semi-colon).

The "collectgens" option is not tested or documented, but maybe it is a debug helper?

tst/testextra/perfect.tst
wilfwilson: @jackschmidt you can suggest a specific change by clicking this button:
Screenshot 2021-06-15 at 12 21 22
Which gives you some text that you edit to look like this:
Screenshot 2021-06-15 at 12 28 33

Once you've posted a comment with a specific change, you can then click a "Commit suggestion" button somewhere on the comment to create a new commit with that change, which is pushed onto the branch that the PR is made from.

You can also click here in the top-right of any file in the "Files changed" tab of a pull request and edit the file in question there, too (although in a not-very-nice plain-text interface, in my opinion). This then saves your changes into a new commit on the PR branch:
Screenshot 2021-06-15 at 12 22 13

Alternatively, you can simply push a commit to the PR branch from your local computer (i.e. in this case you could push directly to the fixes branch on @hulpke's own repository).

(If these options are available to you, you might need some GitHub permissions for `gap-system`, which would be happily given to you - but I don't know the details of this).

We normally only do the first of these three options here for reasons of etiquette, and so that the PR author can decide which changes to accept, and how to arrange the commit history of their PR.

jackschmidt, line 13:
```suggestion
gap> LoadPackage("atlasrep");;
```

So simple, thank you!

hulpke:
The use of atlasrep is laziness of mine. Would it be better to write out the elements to avoid accessing the web?

I'll think about a test -- probably just count generators, since it otherwise becomes a mess.

hulpke:
Actually, the option for generator collection is not needed, but the issue was (my fault!) a careless test for a group being not trivial. I have changed this.

tst/testinstall/grpfp.tst
```diff
@@ -108,5 +108,12 @@ gap> rels:=ParseRelators(F,"a2,b3,c4,abC");
 gap> IsomorphismSimplifiedFpGroup(F/rels);
 [ a, b, c ] -> [ c*b^-1, b, c ]
 
+# ClosureSubgroupNC will not force a triviality or membership test
+# if we do not know anything.
+gap> f:=FreeGroup(3);;
```
jackschmidt, line 113: Line 1 needs:

```
#@local a,b,c2,e,f,g,iter,l,s,F,rels
```

to 

```
#@local a,b,c2,e,f,g,iter,l,s,F,rels,sub
```

hulpke: Thank you!

jackschmidt (approved):
These all look good to me!

fingolfin (approved):

PR #4568, opened 2021-06-16
Remove references to obsolete BANNER global variable

fingolfin:


wilfwilson (approved):
Is there any reason not to delete these references too?

https://github.com/gap-system/gap/blob/e7aa60619b236827222776140a951919cec67809/lib/obsolete.gd#L182

https://github.com/gap-system/gap/blob/e7aa60619b236827222776140a951919cec67809/lib/obsolete.gd#L170

PR #4572, opened 2021-06-18
Release scripts: create and upload a file `help-links.json.gz` for use on the website

wilfwilson:
This is an adaptation of the script https://github.com/gap-system/GapWWW/blob/master/etc/LinksOfAllHelpSections.g from the [GapWWW repository](https://github.com/gap-system/GapWWW).

I think that the updated script is a bit more robust/clear. The output is now JSON rather than YAML. I think it makes sense to build this information once, here, while making all of the other archives, and then upload it to the GitHub release.

In my PR https://github.com/gap-system/GapWWW/pull/247; I have deleted `LinksOfAllHelpSections.g` and updated the website to use this new `help-links.json` file to get its information about links in the manuals.

The reason that I now additionally **format** the outputted JSON files (with Python) is that it makes it much easier to see changes (i.e. to do a diff between GAP versions), and the extra characters come at a minimal space cost.

I am working on another PR to _this_ repository which updates (and massively simplifies) `dev/releases/update_website.py` to play nicely with the new system introduced by this PR and by https://github.com/gap-system/GapWWW/pull/247.

If you're happy, I think this is ready to merge.

### DIFF
--- a/dev/releases/HelpLinks-to-JSON.g
+++ b/dev/releases/HelpLinks-to-JSON.g
@@ -1,0 +1,85 @@
+##  A utility for dumping links to all available manual sections
+##  (for use with named manual references on web pages or in search utilities).
+##
+##  The main purpose for this script is to create the help-links.json file that
+##  is used by the GAP website. This happens as part of the release process.
+##
+##  In order to be used on the GAP website, relative to a particular GAP
+##  version, this script should be run from within the corresponding released
+##  version of GAP, with exactly its bundled packages (installed in the pkg/
+##  subdirectory), and with no special user config. The GAP and package manuals
+##  are precompiled in the released version.
+
+# Temporary hack to make sure that the record components in the JSON file are
+# sorted, to increase stability in the output between GAP versions.
+# Remove once https://github.com/gap-packages/json/pull/24 is merged+released.
+InstallMethod(RecNames, [IsRecord and IsInternalRep], x -> AsSSortedList(REC_NAMES(x)));
+
+LoadPackage("json");
+out := OutputTextUser();
+r := rec();
+# load all books (without prompting the user for input)
+Perform(HELP_KNOWN_BOOKS[1], HELP_BOOK_INFO);;
+for x in NamesOfComponents(HELP_BOOKS_INFO) do
+  book := HELP_BOOKS_INFO.(x);
+
+  # We need two capitalised versions of the book name:
+  # - One with " (not loaded)" removed (we don't want this suffix, ultimately)
+  # - One with ": " added (for matching and removing this as a substring)
+  bname := ReplacedString(book.bookname, " (not loaded)", "");
+  match := Concatenation(book.bookname, ": ");
+
+  # Paths in the record <book> are full paths, which are machine-specific.
+  #
+  # For the GAP website use-case described above, for the individual items, we
+  # require paths relative to an imaginary common GAP root:
+  #   doc/{ref/tut/dev}/...    for the GAP manuals,
+  #   pkg/<pkgdir>/...         for the package manuals,
+  # where <pkgdir> agrees with the naming of the relevant bundled GAP package.
+  #
+  # If GAP is installed with packages from a release, then this can be achieved
+  # by simply removing the relevant GAP root path.
+  # We store the specific paths relative to the book directory, and we store
+  # the book directory relative to the imaginary GAP root.
+  #
+  # There seem to be two different ways to access the book directory.
+  if IsBound(book.directory) then
+    # Convert directory object to string.
+    fulldir := Filename(book.directory, "");
+  elif IsBound(book.directories) then
+    Assert(0, IsList(book.directories) and IsTrivial(book.directories));
+    # gapmacro manual: convert, and replace any trailing "/doc/" with "/htm/".
+    fulldir := Filename(book.directories[1], "");
+    if EndsWith(fulldir, "/doc/") then
+      fulldir := Concatenation(fulldir{[1 .. Length(fulldir) - 5]}, "/htm/");
+    fi;
+  else
+    Error("cannot determine the installation directory of the book: ", book);
+  fi;
+  reldir := ShallowCopy(fulldir);
+  for root in GAPInfo.RootPaths do
+    reldir := ReplacedString(reldir, root, "");
+  od;
+
+  r.(bname) := rec(directory := reldir, reference := rec());
+
+  for i in [1 .. Length(book.entries)] do
+    entry := HELP_BOOK_HANDLER.HelpDataRef(book, i);
+    name := StripEscapeSequences(entry[1]);
+    name := ReplacedString(name, match, "");
+    NormalizeWhitespace(name);
+
+    path := entry[6];
+    if path = fail then
+      path := "FAIL";
+    else
+      path := ReplacedString(path, fulldir, "");
+      path := ReplacedString(path, "_mj.html", ".html");
+    fi;
+
+    r.(bname).reference.(name) := path;
+  od;
+od;
+
+GapToJsonStream(out, r);;
+QUIT;
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,19 +62,36 @@ jobs:
         run: dev/ci-download-pkgs.sh
       - name: "Make archives"
         run: python -u ./dev/releases/make_archives.py
+      - name: "Record the GAP build version"
+        id: get-build
+        run: |
+          # Remove the leading "tmp/" and the trailing "-core.tar.gz"
+          BUILD=`ls tmp/gap-*-core.tar.gz | cut -c 5-`
+          echo "::set-output name=name::${BUILD%-core.tar.gz}"
 
-      - name: "Upload artifacts"
+      # Upload the main GAP .tar.gz file (which includes packages).
+      # We only upload this tarball in order to minimise our demand on GitHub's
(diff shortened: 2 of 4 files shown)


dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,53 @@
...
+InstallMethod(_GapToJsonStreamInternal,
+[IsOutputStream, IsObject],
+function(o, x)
+    PrintTo(o, "null");
+end);
+
+out := OutputTextUser();
+r := rec();
+# load all books (without prompting the user for input)
+Perform(HELP_KNOWN_BOOKS[1], HELP_BOOK_INFO);;
```
fingolfin, line 21: I guess the point is to ensure the list is sorted?

dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,53 @@
...
+##
+##  This is meant to be used with a plain vanilla GAP release, with
+##  only the bundled packages and no special user config.
+##
+
+LoadPackage("json");
+
+InstallMethod(_GapToJsonStreamInternal,
+[IsOutputStream, IsObject],
+function(o, x)
```
fingolfin, line 14: What is this about? Perhaps add a comment?

dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,53 @@
...
+    name := StripEscapeSequences(entry[1]);
+    name := ReplacedString(name, match, "");
+    NormalizeWhitespace(name);
+
+    path := entry[6];
+    if path = fail then
+      path := "FAIL";
+    else
+      for root in GAPInfo.RootPaths do
+        path := ReplacedString(path, root, "");
```
fingolfin, line 43: Perhaps also add a comment explaining this loop? I think it ensures that the paths do not containing anything specific to the build machine / are relative. Anyone working on this code in thwnfugure migjt appreciate being told :-)

fingolfin (approved):
Great work, thanks! I left some optional suggestions.

dev/releases/HelpLinks-to-JSON.g
wilfwilson: Ah, I presume you mean for the line below? I was confused for a minute. I'll comment about my thinking on that line specific line.

dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,53 @@
...
+[IsOutputStream, IsObject],
+function(o, x)
+    PrintTo(o, "null");
+end);
+
+out := OutputTextUser();
+r := rec();
+# load all books (without prompting the user for input)
+Perform(HELP_KNOWN_BOOKS[1], HELP_BOOK_INFO);;
+for x in Set(NamesOfComponents(HELP_BOOKS_INFO)) do # Any point doing Set?
```
wilfwilson, line 22: It would be nice to keep diffs between the `help-links.json` files of subsequent GAP versions as small as possible (although this is not essential). Therefore it would make sense for the books to be sorted in the output.

The question that I'm asking myself in this line is basically: does sorting the book names _here_ actually help to produce a sorted _output_?

The question arises because I am constructing my information as a record of records, and I have no idea whether there is any relationship between the _order_ in which I create record components, and the _order_ in which `GapToJsonStream` outputs record components to a stream.

Perhaps `GapToJsonStream` outputs record components: 
* in the order they were created (in which case the `Set` would help here, and we should also sort the entries of each book)
* randomly (in which case we can't do anything to help)
* in alphabetical order (in which case we don't need to do anything to get our desired behaviour; the `Set` should go).

I haven't yet experimented to see what it does.

Perhaps it would also be possible and sensible to get Python to sort everything when it reformats the JSON.

dev/releases/HelpLinks-to-JSON.g
wilfwilson: I will add a comment. Yes, the GAP website expects to get paths relative to a GAP root directory, this *should* be able to be achieved by removing any GAP root from the path. I copied this code from you! https://github.com/gap-system/gap/commit/efe19e5284012cfae069c0f38d73d2de95445679#diff-526d642e5b9cfac653edfcfa495f06aa6a1f4f96dea9b26e352fa6f75a221252

dev/releases/HelpLinks-to-JSON.g
wilfwilson: To be honest I don't know, and I will just delete this. For context, some of the code in this file was copied from `dev/releases.PackageInfos-to-JSON.g`. I copied this particular bit from the top of that file; it was added when the file was created in https://github.com/gap-system/gap/commit/8e158d6252d396a7dc04a70db42e7412e616232a#diff-526d642e5b9cfac653edfcfa495f06aa6a1f4f96dea9b26e352fa6f75a221252.

I guess that @ssiccha added that because a PackageInfo record could in theory contain arbitrarily weird GAP objects, and he wanted to write the script in a way that doesn't break when it encounters a crazy PackageInfo record. Therefore he installed this default method for `IsObject` (i.e. to catch unexpected things) which overwrote the JSON method.

But here, we _construct_ the record piece by piece that we feed to the JSON package; it is a record of records of strings. Therefore it's known to be safe, and we don't need such a custom fallback.

dev/releases/HelpLinks-to-JSON.g
wilfwilson: After looking at this further, I am going to remove the `Set` since it doesn't affect the output. I have made a PR to the json package to help find a better solution https://github.com/gap-packages/json/pull/24

wilfwilson:
Thanks for the review @fingolfin. I've responded individually to your comments here on GitHub, and added some more comments to the script.

I also made a new release on my fork which is the release artifacts for the current version of this PR: https://github.com/wilfwilson/gap/releases/tag/v4.12.2

dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,61 @@
...
+for x in NamesOfComponents(HELP_BOOKS_INFO) do
+  book := HELP_BOOKS_INFO.(x);
+
+  # <x> is the lowercase bookname; we want two capitalised versions:
+  # - One with " (not loaded)" removed (we don't want this string in the end)
+  # - One with ": " added (for matching and removing this as a substring)
+  bname := ReplacedString(book.bookname, " (not loaded)", "");
+  match := Concatenation(book.bookname, ": ");
+
+  r.(bname) := rec();
```
wilfwilson, line 28: Another development to make diffs smaller, and to reduce file size, could be to separate the directory name of the package from the path. Therefore, instead of `r.Digraphs` looking like:
```
"Digraphs": {
  "String": "pkg/digraphs-1.4.1/doc/chap9.html#X81FB5BE27903EC32",
  "CharacteristicPolynomial": "pkg/digraphs-1.4.1/doc/chap5.html#X87FA0A727CDB060B",
  "AsTransformation": "pkg/digraphs-1.4.1/doc/chap3.html#X7C5360B2799943F3",
  "Copyright": "pkg/digraphs-1.4.1/doc/chap0.html#X81488B807F2A1CF1",
  ...
}
```
(with lots of repeated `pkg/digraphs-1.4.1`; meaning also that every single line of every single package will change when a new version of the package is picked up), we could instead have:
```
"Digraphs": {
  "directory": "pkg/digraphs-1.4.1/"
  "reference": {
    "String": "pkg/digraphs-1.4.1/doc/chap9.html#X81FB5BE27903EC32",
    "CharacteristicPolynomial": "pkg/digraphs-1.4.1/doc/chap5.html#X87FA0A727CDB060B",
    "AsTransformation": "pkg/digraphs-1.4.1/doc/chap3.html#X7C5360B2799943F3",
    "Copyright": "pkg/digraphs-1.4.1/doc/chap0.html#X81488B807F2A1CF1",
    ...
  }
}
```
I think this would require only a small change to the website to accommodate.

wilfwilson: I've done this.

dev/releases/HelpLinks-to-JSON.g
```diff
@@ -0,0 +1,84 @@
...
+##
+##  In order to be used on the GAP website, relative to a particular GAP
+##  version, this script should be run from within the corresponding released
+##  version of GAP, with exactly its bundled packages (installed in the pkg/
+##  subdirectory), and with no special user config. The GAP and package manuals
+##  are precompiled in the released version.
+
+# Temporary hack to make sure that the record components in the JSON file are
+# sorted, to increase stability in the output between GAP versions.
+InstallMethod(RecNames, [IsRecord and IsInternalRep], x -> AsSSortedList(REC_NAMES(x)));
```
fingolfin, line 15: See https://github.com/gap-packages/json/pull/24 - I hope @ChrisJefferson will merge that and make a new release soon

fingolfin (approved):

dev/releases/HelpLinks-to-JSON.g
wilfwilson: I added a comment to the script explicitly pointing to this PR. I also added the same hack and comment to the package-infos script as well.

wilfwilson:
Thanks again. I've just been making a few tweaks; I'll merge once the CI passes again.

Here's an example of the JSON files:
* [help-links.json.gz](https://github.com/gap-system/gap/files/6686205/help-links.json.gz)
* [package-infos.json.gz](https://github.com/gap-system/gap/files/6686207/package-infos.json.gz)

PR #4574, opened 2021-06-21
Performance improvements and cleanup related to permutation representations

hulpke:
- When calculating transversals in large degree permutation groups, do not maintain large image lists
- Option to work harder on finding a smaller degree permutation representation
- Action on subgroups may use reduced degree if appropriate  (can speed up subgroup lattice)
- Clean up interface of `MinimalFaithfulPermutationDegree` (avoid use of option for getting representation)
- 
## Text for release notes

None

### DIFF
--- a/lib/csetperm.gi
+++ b/lib/csetperm.gi
@@ -158,8 +158,17 @@ end );
 InstallGlobalFunction( AddCosetInfoStabChain, function( G, U, maxmoved )
     local   orb,  pimg,  img,  vert,  s,  t,  index,
             block,  B,  blist,  pos,  sliced,  lenflock,  i,  j,
-            ss,  tt,t1,t1lim;
+            ss,  tt,t1,t1lim,found,tl,vimg;
     
+    # iterated image
+    vimg:=function(point,list)
+    local i;
+      for i in list do
+        point:=point^i;
+      od;
+      return point;
+    end;
+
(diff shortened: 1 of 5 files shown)


lib/csetperm.gi
```diff
@@ -273,17 +282,34 @@ InstallGlobalFunction( AddCosetInfoStabChain, function( G, U, maxmoved )
...
+                        fi;
+
+                        if found then
 			    U.repsStab[ t ][ s ] := true;
 			    index := index + lenflock;
 			fi;
+
 		      else
                         U.repsStab[ t ][ s ] := true;
                         index := index + lenflock;
```
wilfwilson, line 315: I was wondering why the indentation looked so weird here. And the answer is not that you're doing anything weird, it's that some of the _unchanged_ lines use tabs, and GitHub renders tabs with a different width than does your text editor (or my text editor, for that matter).

So there's no problem here 🙂

wilfwilson (approved):
All looks good to me. Thanks for the separate commits and their commit messages.

lib/csetperm.gi
wilfwilson: And in fact it looks fine on the main pull request page; it's in the "[Files changed](https://github.com/gap-system/gap/pull/4574/files)" tab where it renders weirdly.

PR #4582, opened 2021-06-25
Releases CI: name Windows installers with `GAP_BUILD_VERSION`

wilfwilson:
Sorry; yet more tinkering, but I think this will be the last for a while.

This is mostly a bit of general cleanup in the "Wrap releases" workflow. The main change is to rename how the Windows installers are named in builds on PRs and branches, in particular to include the `GAP_BUILD_VERSION` in the name. I think this is just more useful.

Therefore, rather than `gap-master-x86.exe` and so on, we will now have (for example, taken from https://github.com/wilfwilson/gap/actions/runs/969228341):
Screenshot 2021-06-25 at 09 46 14
(I have some made some experimental "releases" on my fork, which is why it shows as being after `4.12.1`.)

For a released version, the `GAP_BUILD_VERSION` is (e.g.) `4.12.1`, so there is no change there.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       cygwin-matrix: ${{ steps.set-cygwin-matrix.outputs.matrix }}
+      gap-build-version: ${{ steps.get-build.outputs.name }}
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Good idea, I like it. And thank you for your continued work on this, much appreciated!

wilfwilson:
Sorry, I've just added a do not merge because there's one thing I want to check before this gets merged.

PR #4587, opened 2021-06-26
Add CALL_WITH_STREAM helper

fingolfin:
This allows calling a function while redirect the GAP output to a given stream. This is e.g. useful for the Julia-GAP interface, but can also be used for other GAP interfaces, and probably also has all kinds of other applications.

I am not saying this is the greatest interface, we could argue a lot about that: e.g. perhaps it should instead (or in addition?) be a function available on the kernel level via `libgap-api.h`. But note that I think it's useful to be able to call this from GAP; and from there it is reachable for libgap users, albeit with some overhead for going though the GAP function call machinery. 

Since I am not yet sure this is the best way to go, I delieberately left this as an undocumented kernel function for now -- if we decide it is good, we can come up with a high level name (`CallWithStream`), and possible add some nicer variants (say `CallFuncWithListAndOuputStream`).

But this particular function has already been used as-is for some time in the GAP-Julia interface, so it seems useful to me to have to it. I'd like to have it in the kernel to decouple it from implementation details of the kernel.

CC @embray

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -506,6 +506,34 @@ Int READ_GAP_ROOT ( const Char * filename )
 
 /****************************************************************************
 **
+*F  FuncCALL_WITH_STREAM( <stream>, <func>, <args> )
+**
+**  Temporarily set the active output stream to <stream>, then call the
+**  function <func> with the arguments in the list <args>. This can for
+**  example be used to capture the output of a function into a string.
+*/
+static Obj FuncCALL_WITH_STREAM(Obj self, Obj stream, Obj func, Obj args)
+{
+    RequireOutputStream(SELF_NAME, stream);
+    RequireSmallList(SELF_NAME, args);
+
+    TypOutputFile output = { 0 };
+    if (!OpenOutputStream(&output, stream)) {
+        ErrorQuit("CALL_WITH_STREAM: cannot open stream for output", 0, 0);
+    }
+
+    Obj result = CallFuncList(func, args);
+
+    if (!CloseOutput(&output)) {
+        ErrorQuit("CALL_WITH_STREAM: cannot close output", 0, 0);
+    }
+
+    return result;
+}
+
+
+/****************************************************************************
+**
 *F  FuncCLOSE_LOG_TO()  . . . . . . . . . . . . . . . . . . . .  stop logging
 **
 **  'FuncCLOSE_LOG_TO' implements a method for 'LogTo'.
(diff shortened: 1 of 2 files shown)


fingolfin:
Upon thinking some more about this, I've actually decided to completely rewrite it, into a new function `CALL_WITH_STREAM` which is basically `CallFuncList` with an extra `stream` argument (so now the return value can be used, and any number of arguments used). The previous code used `Call1ArgsInNewReader` but without really a good justification; at the same time, it had calls to `ErrorQuit` in it -- what's the point of suppressing errors if you then trigger them anyway. (if one needs this, one can try to use `CALL_WITH_CATCH`; it also has issues, but at least its a known quantity, no need to add yet another variant into the mix).

ThomasBreuer (approved):
Yes, this can eventually replace JuliaInterface's `STREAM_CALL`.
I found one typo.

Just for curiosity: Why
`GVAR_FUNC(CALL_WITH_STREAM, 3, "stream, func, args")`
and not
GVAR_FUNC_3ARGS(CALL_WITH_STREAM, stream, func, args)`?

fingolfin:
Indeed it should have been using `GVAR_FUNC_3ARGS` (and now it is)

frankluebeck:
`CALL_WITH_STREAM` seems closely related to the documented function `AppendTo1`:
```
CWS := function(stream, func, arguments)
   AppendTo1(stream, function() CallFuncList(func, arguments); end);
end;
# or the other way round
AT1 := function(stream, func)
  CALL_WITH_STREAM(stream, func, []);
end;
```
The example
```
gap> str := "";;
gap> out := OutputTextString(str, false);;
gap> SetPrintFormattingStatus(out, false);
gap> CALL_WITH_STREAM(out, Display, [ [[1,2],[3,4]] ]);
```
could be done with
``` 
gap> str := "";;
gap> out := OutputTextString(str, false);;
gap> SetPrintFormattingStatus(out, false);
gap> AppendTo1(out, function() Display([[1,2],[3,4]]); end);
```

fingolfin:
Fascinating, I never heard of `AppendTo1` before -- and indeed, it is part of GAPDoc, not GAP. According to its documentation, it takes a filename, not a stream? Perhaps the documentation is in incomplete ?

wilfwilson (approved):

frankluebeck:
Yes, `PrintTo1` and `AppendTo1` also work with open streams, just as `PrintTo` and `AppendTo` do. And yes, the documentation could be improved in this respect (for both pairs fo functions).

PR #4589, opened 2021-06-28
Add CALL_WITH_FORMATTING_STATUS helper

fingolfin:
... and remove again (SET_)PRINT_FORMATTING_CURRENT resp.
`SetPrintFormattingStatus("*current*",false);`. Those APIs always seemed a bit
awkward to me (albeit pragmatic), e.g. there were doubts in my mind whether or
not these should just temporarily modify the formatting status (which is what
they were added for), or modify it more permanently by also calling
`SetPrintFormattingStatus` on the underlying stream object (which would be
consistent with how `SetPrintFormattingStatus` behaves otherwise)

Another advantage of this approach is that it nicely composes with the
`CALL_WITH_STREAM` added in PR #4587; I.e., if one wants to print to
a specific output and at the same time temporarily modify the formatting status
of that output, one could do it this way:

```gap
CALL_WITH_FORMATTING_STATUS(false, CALL_WITH_STREAM, [stream, func, args]);
```

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -37,6 +37,7 @@
 #include "sysfiles.h"
 #include "sysopt.h"
 #include "sysstr.h"
+#include "trycatch.h"
 
 #ifdef HPCGAP
 #include "hpc/aobjects.h"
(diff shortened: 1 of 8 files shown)


ThomasBreuer (approved):
With this pull request, we keep the achievements of #4555 and get rid of `*current*`. Very nice.

frankluebeck:
I still find it difficult to understand what `*current*` meant. Good, if it is no longer needed. 

Can you give more examples how `CALL_WITH_FORMATTING_STATUS` could be used?

In the examples
```
CALL_WITH_FORMATTING_STATUS(false, PrintTo, ["guck", [1..100]+0, "\n"]);
out := OutputTextFile("guck",false);;
CALL_WITH_FORMATTING_STATUS(false, PrintTo, [out, [1..100]+0, "\n"]);
CloseStream(out);
```

it is not so clear to which stream the given formatting status should be applied.
The output in the file `guck` is formatted. Is this intended?
Or, is there another easy way, using the new utilities, to print or append to a file (given by name) without formatting?

wilfwilson (approved):

fingolfin:
As with `*current*` before, the goal is to modify the printing status of the currently (at the time the new function is called) active output. If the function that is the called performs actions that change the active output (as in your example) then the formatting status of this different output stream is not affected.

The primary use case is to use this when calling code that uses Print/Display/ViewObj. Indeed, in the applications I have in mind, `func` will usually be `ViewObj` or `Display`.

src/io.c
```diff
@@ -1910,23 +1910,26 @@ static Obj FuncPRINT_FORMATTING_ERROUT(Obj self)
...
-static Obj FuncPRINT_FORMATTING_CURRENT(Obj self)
-{
     TypOutputFile * output = IO()->Output;
     if (!output)
-        ErrorMayQuit("PRINT_FORMATTING_CURRENT called while no output is open", 0, 0);
-    return output->format ? True : False;
+        ErrorMayQuit("CALL_WITH_FORMATTING_STATUS called while no output is open", 0, 0);
+    BOOL old = output->format;
+    output->format = (status != False);
+    Obj result = CallFuncList(func, args);
```
ChrisJefferson, line 1930: I was a bit worried about what happens if this enters the break loop, and the user exits, which leads to broken formatting (this isn't new to this PR, but maybe worth fixing anyway).

Maybe replace the line `Obj result = CallFuncList(func,args);` with something like:

    Obj result = 0;
    GAP_TRY
    {
      result = CallFuncList(func, args);
    }
    GAP_CATCH
    {
      output->format = old;
      GAP_THROW();
    }

fingolfin: Good point, done.

I also added a similar tweak to `CALL_WITH_STREAM` to restore the active output if there is an error. Of course many other kernel functions still don't do such a thing

PR #4596, opened 2021-07-01
Simplify `dev/releases/update-website.py` for the new GapWWW structure

### LINKED ISSUE #4316: Update dev/releases README
We've changed quite a bit of the release scripts after the last GAP Days. Me and @ruthhoffmann are looking into updating the README for the `make_archives.py` and `make_github_release.py`.

@wilfwilson could you maybe have a look into updating the description of the website script? Or is the way that it works still going to change a lot?

wilfwilson:
In recent weeks, I have updated the GAP website to use the fairly new `package-infos.json.gz` and `help-links.json.gz` files, and I have made the GAP website use JSON in other places too. This had the promise of making the website much simpler to update (indeed it is now automated to use the new `update-website.py` script of this PR; see https://github.com/gap-system/GapWWW/pull/241).

This PR is the result of that promise.

This PR also refactors `dev/releases/README.md`, so this closes #4316.

_(To be honest I think the 'simple' version of the README is sufficient; that is the release process that we want to happen in practice. If someone wants to know more (in order to further the develop the release process, say), they can find out how it works by looking at the GitHub Actions workflows, and at the various scripts that they call. So I suggest we don't spend too much further effort keeping the 'detailed' version up to date, and perhaps remove it when it next becomes out of date.)_

You can try out this script by checking out this branch, and then running it from the root directory of an up-to-date master branch of GapWWW. The result should be... no changes, because all GitHub release data for GAP is already known by GapWWW.

But here's an example of what you get if a new (but fake) GAP v4.12.0 exists: https://github.com/wilfwilson/GapWWW/pull/9/files (ignore the changes from `gap-system` to `wilfwilson`; that's because I'm fetching release data there from my own GAP fork).

### DIFF
--- a/dev/releases/README.md
+++ b/dev/releases/README.md
@@ -1,102 +1,102 @@
-# How To Make a GAP Release for version Z.X.Y
+# How to release GAP version X.Y.Z
  
-**This release process assumes the following**
-- A stable branch of GAP has been identified or agreed upon. This branch will be identified here as `stable-Z.X.Y` and the release notes in `CHANGES.md` have been created.
-- You run these scripts in the root directory of a clone of the [GAP repository](https://github.com/gap-system/gap)
-- You have compiled this clone of GAP
-- You have a clone of the [GapWWW repository](https://github.com/gap-system/GapWWW)
+**These instructions assume the following**
+- It has been agreed to make a release called `X.Y.Z` from the `stable-X.Y` branch.
+	- Note that most likely `X=4` for the foreseeable future.
+- All changes to be included in the release are now present in the `stable-X.Y` branch.
+	- This includes the release notes for version `X.Y.Z` being included in `CHANGES.md`.
 
-## Dependencies
-Before starting the release process, the scripts have the following dependencies. Make sure you have the following installed and up to date
+
+## The release process for GAP -- the quick guide
+
+This is the way the process should happen in practice, with the help of GitHub Actions. Instructions for achieving the same without GitHub Actions are given later (in case GitHub Actions breaks or disappears).
+
+1. Access your local clone of the GAP repository.
+2. Make sure that your clone is up to date with `gap-system/gap`.
+3. Create an annotated tag `vX.Y.Z` in your clone at the appropriate commit in the `stable-X.Y` branch, and push the tag to `gap-system/gap`. For example:
+    ```
+    git tag -m "Version X.Y.Z" vX.Y.Z stable-X.Y
+    git push origin vX.Y.Z
+    ```  
+4. Wait. Pushing the tag will trigger the “[Wrap releases](https://github.com/gap-system/gap/actions/workflows/release.yml)” GitHub Actions workflow on `gap-system/gap`, which wraps the release archives and Windows installers, and creates a release on GitHub with the archives and installers attached. This takes around 90 minutes.
+5. Once the “Wrap releases” workflow is finished, either manually dispatch the “[Sync](https://github.com/gap-system/GapWWW/actions/workflows/sync.yml)” GitHub Actions workflow on `gap-system/GapWWW`, or wait overnight for it to happen on schedule.
+	- This Workflow creates a pull request to `gap-system/GapWWW`, which updates the GAP website according to the release data hosted on `gap-system/gap`.  Check that the result is sensible.
+6. When it is time to publicise the new GAP release, merge the pull request to GapWWW.
+7. There are currently additional steps required for the changes to appear on <https://www.gap-system.org>, which could be automated but are currently not; these are described in steps 10–13 below.
+
+
+## The release process for GAP -- the more detailed guide
+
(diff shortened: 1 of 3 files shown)


dev/releases/update_website.py
```diff
@@ -14,416 +14,193 @@
...
-    except:
-        error("non-existent or ambiguous release tag name: " + args.tag)
-else:
-    # Using latest release; therefore ignore drafts and prerelease
-    release = [ x for x in utils.CURRENT_REPO.get_releases() if \
-                not x.draft and not x.prerelease and is_possible_gap4_release_tag(x.tag_name)]
+# When we find a previously unknown release, we will extract the release date
+# from the configure.ac file contained in gap-{version}-core.tar.gz.
+# This date is set by the make_archives.py script.
+def get_date_from_configure_ac():
```
fingolfin, line 64: This function does more than this comment and its name let on: it in fact downloads and extracts a tarball, which I found surprising.

I would suggest to at least update the comment. On the long run, it might be better to split this into two (or more) functions, as the fragments tend to be easier to work with / compose differently (e.g. the downloaded 
tarball resp. the extracted content, may be useful for other things. E.g. at some point, we will also want to automate updating the documentation on the website)

But if it works now, I don't want to ask you to break it now.

dev/releases/README.md
```diff
@@ -1,12 +1,36 @@
...
+2. Make sure that your clone is up to date with `gap-system/gap`.
+3. Create an annotated tag `vX.Y.Z` in your clone at the appropriate commit in the `stable-X.Y` branch, and push the tag to `gap-system/gap`. For example:
+    ```
+    git tag -m "Version X.Y.Z" vX.Y.Z stable-X.Y
+    git push origin vX.Y.Z
+    ```  
+4. Wait. Pushing the tag will trigger the “[Wrap releases](https://github.com/gap-system/gap/actions/workflows/release.yml)” GitHub Actions workflow on `gap-system/gap`, which wraps the release archives and Windows installers, and creates a release on GitHub with the archives and installers attached. This takes around 90 minutes.
+5. Once the “Wrap releases” workflow is finished, either manually dispatch the “[Sync](https://github.com/gap-system/GapWWW/actions/workflows/sync.yml)” GitHub Actions workflow on `gap-system/GapWWW`, or wait overnight for it to happen on schedule.
+	- This Workflow creates a pull request to `gap-system/GapWWW`, which updates the GAP website according to the release data hosted on `gap-system/gap`.  Check that the result is sensible.
+6. When it is time to publicise the new GAP release, merge the pull request to GapWWW.
```
fingolfin, line 24: I think this is missing one step: SSH'ing into the webserver in St Andrews and (a) installing the latest package manuals there and (b) `git pull`ing the latest GapWWW changes.

Granted, this seems to be missing in the existing README as well, but we should add it (on the long run, we should of course ideally get rid of that step / automate it too, but AFAIK this is not the case yet, is it?)

fingolfin (approved):
Looks good to me, although I did not check every single line everywhere

dev/releases/README.md
wilfwilson: Correct, the steps of:
* extracting the GAP and package manuals with GapWWW's `etc/extract-manuals.py` script
* putting them in the right place on the server
* updating the git repository on the server
* building the new Jekyll website

still need to be done manually. I'll add at least some information about that in this PR.

dev/releases/update_website.py
wilfwilson: I will at least update the comment, and I agree that splitting it would be better.

dev/releases/README.md
wilfwilson: (I also agree that we want automate these steps as much as possible 🙂)

wilfwilson: (I also agree that we want automate these steps as much as possible 🙂)

dev/releases/README.md
```diff
@@ -17,86 +41,53 @@ Before starting the release process, the scripts have the following dependencies
...
+You can generate such a token as follows (see also <https://help.github.com/articles/creating-an-access-token-for-command-line-use>).
 
 1. Go to <https://github.com/settings/tokens>.
-
 2. Click **Generate new token**.
-
 3. Select the scope "public_repo", and give your token a descriptive name.
-
 4. Click **Generate token** at the bottom of the page.
+5. Copy the token to your clipboard. For security reasons, after you navigate off the page, you will not be able to see the token again. You therefore should store it somewhere, e.g. with option 3 in the following list.
```
ssiccha, line 90: What's "option 3 in the following list"? I guess that list doesn't exist anymore.

ssiccha (approved):
I also haven't looked at every line in detail. This looks a lot simpler than what had to be done before, nice! :+1:

dev/releases/update_website.py
wilfwilson: I've split this up.

PR #4597, opened 2021-07-02
Improvements around the MTC (and two ad-hoc fixes)

hulpke:
Slight changes in how the MTC is run and how the presentation is cleaned up by Tietze transformations.
Fixed the long-time-broken (wrong result) RelatorRepresentatives.

Fixed an overeager assertion (that could blame correct cases) and included a requested example for LowLayerSubgroups

## Text for release notes

None -- the errors would not have been notable to the user and the performance improvement is small.

## Further details

### DIFF
--- a/lib/grpfp.gi
+++ b/lib/grpfp.gi
@@ -3518,7 +3518,7 @@ end );
 ##  is automatically cyclically reduced).
 ##
 InstallGlobalFunction( RelatorRepresentatives, function ( rels )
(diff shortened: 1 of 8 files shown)


fingolfin:
Looks good to me but it seems several examples in `lib/tietze.gd` need to be updated

fingolfin (approved):
Looks good to me, thank you!

PR #4599, opened 2021-07-05
ENHANCE: `PositionSublist` uses `Position` to search for first entry

hulpke:
instead of linear search in the library. This is much faster when there is little self-overlap.
This is used in some of the routines working with presentations and gives a significant speedup.

Ultimately, this should become a kernel function.

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -1769,7 +1769,7 @@ InstallMethod( PositionSublist,
     "list,sub,pos",
     [IsList,IsList,IS_INT],
     function( list,sub,start )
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Nice!

I took the liberty of pushing another commit to this PR which refactors it a bit, to reduce diffs (when hiding whitespace diffs) to the previous version of the code (and some negligible performance gains -- like, `for` loops over ranges are a bit faster than `while` loops).

wilfwilson (approved):

PR #4601, opened 2021-07-06
kernel: fix `echoandcheck` error messages

fingolfin:


src/sysfiles.c
```diff
@@ -166,12 +166,12 @@ static ssize_t echoandcheck(int fid, const char *buf, size_t count)
   else {
       ret = write(syBuf[fid].echo, buf, count);
       if (ret < 0) {
-          if (syBuf[fid].fp == fileno(stdout) || syBuf[fid].fp == fileno(stderr)) {
+          if (syBuf[fid].echo == fileno(stdout) || syBuf[fid].echo == fileno(stderr)) {
               Panic("Could not write to stdout/stderr.");
```
fingolfin, line 170: We might want to give a bit more information here and also give the value of `fid`? Same below... @ChrisJefferson ?

ChrisJefferson: Yes, I think knowing 'fid' would be useful, in the unlikely event people see this message it will give us a better idea what has gone wrong.

ChrisJefferson (approved):
Approved, but I'd also be happy for error message to be improved (as discussed in comment)

fingolfin:
I've updated the error message, OK this way, @ChrisJefferson ?

PR #4602, opened 2021-07-07
Improve documentation and argument checks for `StabChainBaseStrongGenerators`

wilfwilson:
I made a note to myself to document the 2-argument version of `StabChainBaseStrongGenerators` (which was already implemented and is probably what most people would want to use). I finally got around to it, and in the process I made a few other small improvements: added a manual example; added tests; added some argument checks.

(Can anyone think when you might want the `one` to _not_ be `()`?)

### DIFF
--- a/lib/stbc.gd
+++ b/lib/stbc.gd
@@ -216,20 +216,33 @@ DeclareGlobalName( "DefaultStabChainOptions" );
 
 #############################################################################
 ##
-#F  StabChainBaseStrongGenerators( <base>, <sgs>, <one> )
+#F  StabChainBaseStrongGenerators( <base>, <sgs>[, <one>] )
 ##
 ##  <#GAPDoc Label="StabChainBaseStrongGenerators">
 ##  <ManSection>
-##  <Func Name="StabChainBaseStrongGenerators" Arg='base, sgs, one'/>
+##  <Func Name="StabChainBaseStrongGenerators" Arg='base, sgs[, one]'/>
 ##
 ##  <Description>
 ##  Let <A>base</A> be a base for a permutation group <M>G</M>, and let
 ##  <A>sgs</A> be a strong generating set for <M>G</M> with respect to
 ##  <A>base</A>; <A>one</A> must be the appropriate identity element of
 ##  <M>G</M> (see <Ref Attr="One"/>, in most cases this will be <C>()</C>).
-##  This function constructs a stabilizer chain without the need to find
+##  This function constructs a stabilizer chain corresponding to the given
+##  base and strong generating set without the need to find
 ##  Schreier generators;
 ##  so this is much faster than the other algorithms.
+##  <P/>
+##  If <A>sgs</A> is nonempty, then the argument <A>one</A> is optional;
+##  if not given, then the <Ref Attr="One" Style="Text"/> of
+##  <C><A>sgs</A>[1]</C> is taken as the identity element.
+##  <Example><![CDATA[
+##  gap> sc := StabChainBaseStrongGenerators([1,2], [(1,3,4), (2,3,4)], ());
+##  <stabilizer chain record, Base [ 1, 2 ], Orbit length 4, Size: 12>
+##  gap> GroupStabChain(sc) = AlternatingGroup(4);
+##  true
+##  gap> StabChainBaseStrongGenerators([1,3], [(1,2),(3,4)]);
+##  <stabilizer chain record, Base [ 1, 3 ], Orbit length 2, Size: 4>
+##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
--- a/lib/stbc.gi
+++ b/lib/stbc.gi
@@ -398,12 +398,20 @@ end);
 InstallGlobalFunction(StabChainBaseStrongGenerators,function(arg)
 local   base,sgs,one,S,  T,  pnt, genlabels;
 
+    if not Length(arg) in [2, 3]
+        or not ForAll([1, 2], i -> IsHomogeneousList(arg[i])) then
+      ErrorNoReturn("usage: ",
+                    "StabChainBaseStrongGenerators(<base>, <sgs>[, <one>])");
+    fi;
     base:=PlainListCopy(arg[1]);
     sgs:=PlainListCopy(arg[2]);
     if Length(arg)=3 then
       one:=arg[3];
+    elif not IsEmpty(sgs) then
+      one:= One(sgs[1]);
     else
-      one:= One(arg[2][1]);
+      ErrorNoReturn("the identity element must be given as the third argument ",
+                    "when the second argument <sgs> is empty");
     fi;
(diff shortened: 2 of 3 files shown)


lib/stbc.gd
```diff
@@ -216,20 +216,32 @@ DeclareGlobalName( "DefaultStabChainOptions" );
...
+##  <M>G</M> (see <Ref Attr="One"/>; in most cases this will be <C>()</C>).
+##  This function constructs a stabilizer chain corresponding to the given
+##  base and strong generating set without the need to find
 ##  Schreier generators;
 ##  so this is much faster than the other algorithms.
+##  <P/>
+##  If <A>sgs</A> is nonempty, then the third argument <A>one</A> is
+##  optional; if <A>one</A> is not given in this case, then the identity
+##  element of <M>G</M> is taken to be the <Ref Attr="One" Style="Text"/>
+##  of the first entry of <A>sgs</A>.
```
ssiccha, line 238: I found the reference to `G` a bit confusing, I thought "wait, `G` is not an argument, is it?" (But that may just be me). Maybe instead do this?
```suggestion
##  optional; if <A>one</A> is not given in this case, then the
##  <Ref Attr="One" Style="Text"/> of the first entry of <A>sgs</A> is taken
## as the identity element.
```

ssiccha (approved):
LGTM. I've got a suggestion, but am not sure whether that's an improvement.

ssiccha:
And I have no idea when `one` would be something else than `()`. Could `one` also be an identity of a semi-group, which accidentally also is a group or something like that? :man_shrugging:

When `sgs` only is empty if `G` is trivial. Or am I misunderstanding the function.

lib/stbc.gi
```diff
@@ -398,12 +398,20 @@ end);
...
     base:=PlainListCopy(arg[1]);
     sgs:=PlainListCopy(arg[2]);
     if Length(arg)=3 then
       one:=arg[3];
+    elif not IsEmpty(sgs) then
+      one:= One(sgs[1]);
     else
-      one:= One(arg[2][1]);
+      ErrorNoReturn("the identity element must be given as the third argument ",
+                    "when the second argument <sgs> is empty");
```
fingolfin, line 414: But... why? Why don't we simply always do `one:= ()`?

wilfwilson: Good question, I have no idea. As I said at the top of this thread:
> (Can anyone think when you might want the one to not be ()?)

I can't think of a use case. Maybe @ChrisJefferson has an idea?

And @ssiccha – this function is documented to only apply to permutation groups, so it can't be because of semigroups.

fingolfin: Sorry, I missed that question at the top 😊

fingolfin: The only package calling `StabChainBaseStrongGenerators` is `grape` and it also passes `()` as third argument.

But in `TryPcgsPermGroup`, though, we pass `U.identity`... That gave me a clue: we have
```
lib/memory.gi:108:        S.identity := S.identity!.el;
```
So perhaps this is meant to allow supporting "permutations with memory" as produced by `GeneratorsWithMemory`?

Looking at `git blame`, the line in `TryPcgsPermGroup` was added by @hulpke 21 years ago in this commit:
```
Author:     Alexander Hulpke <hulpke@math.colostate.edu>
AuthorDate: 2000-01-25 18:50:09 +0000
Commit:     Alexander Hulpke <hulpke@math.colostate.edu>
CommitDate: 2000-01-25 18:50:09 +0000

    StabChainBaseStrongGenerators gets the appropriate `One' (this permits other
    permutation representations). SM&AH.
```
Perhaps he recalls something?

hulpke: If we use other representations (e.g. elements with memory, or straight line program elements) the identity element will be given in this representation.

fingolfin: No intention to change it -- but clearly the documentation for this function then is wrong / incomplete

wilfwilson: Thanks for the explanation @hulpke, and don't worry, my purpose of the PR is to make the documentation match the behaviour of the code by changing the documentation, rather than to change the behaviour of the code.

However, I now feel a bit out of my depth, since I don't yet know anything about other representations for permutations in GAP.

I am also now worried that #4331 broken `StabChainBaseStrongGenerators` for your private code @hulpke – could you please check? In particular, the assertion `Assert(2, S.labels[1] = ());` is perhaps invalid, if a `one` other than `()` is specified.

wilfwilson:
@ssiccha Yes a stabiliser chain can only have an empty set of strong generators if it is trivial. The 2-argument behaviour prior to this PR is:
```
gap> StabChainBaseStrongGenerators([1], []);
Error, List Element: <list>[1] must have an assigned value in
  one := One( arg[2][1] ); at /Users/Wilf/gap/lib/stbc.gi:406 called from
<function "StabChainBaseStrongGenerators">( <arguments> )
```
So I feel that a specific error message is more helpful. A stabiliser chain needs to know its corresponding identity, so we can't guess it in that case.

lib/stbc.gi
```diff
@@ -398,12 +398,20 @@ end);
 InstallGlobalFunction(StabChainBaseStrongGenerators,function(arg)
 local   base,sgs,one,S,  T,  pnt, genlabels;
 
+    if not Length(arg) in [2, 3]
+        or not ForAll([1, 2], i -> IsHomogeneousList(arg[i])) then
```
wilfwilson, line 402: Is it valid to assume that valid bases and strong generating sets will be homogeneous  lists?

lib/stbc.gi
hulpke: That should (apart from anyhow being in an assertion) be harmless as objects with memory would have to compare as equal to (). 
The reason for the different `one` is if elements get created that will be used later on. By starting with the correct `one' allos to build a proper word expression on the way.

lib/stbc.gi
fingolfin: I think it is but perhaps @hulpke and/or @ChrisJefferson may want to comment?

fingolfin:
This PR looks good to me. But I'd prefer if @hulpke could approve it, too (or point out any issues, of course!). If @ChrisJefferson also has time to look at it: even better :-)

wilfwilson:
That's fine, there's certainly no rush to get this merged.

From my point of view, this PR is doing three things:
* ✅ More accurately document the current behaviour (and I think we are now happy that the current behaviour is correct)
* ✅ Add tests (harmless)
* ❓Check that the second and optional third arguments are homogeneous lists ('list' is fine, but 'homogeneous' is possibly dubious; therefore I could remove the `Homogeneous` part to make this completely uncontroversial).

lib/stbc.gi
ChrisJefferson: I think it's reasonable

ChrisJefferson:
I'm happy with this PR, as long as it doesn't break anything @hulpke is doing (I wouldn't expect so)

PR #4604, opened 2021-07-07
Add Bray/Holt/Roney-Dougal 2013 to the GAP bibliography

wilfwilson:
This book is the source for the data used in the function `OuterAutoSimplePres` of `grp/simple.gi`.

I'm not sure if there's a good place to actually cite this book in the manual; I'm open to suggestions if anyone has any. Otherwise, it's surely a good starting point to at least put this book in our bib file.

### DIFF
--- a/doc/manualbib.xml
+++ b/doc/manualbib.xml
@@ -5538,4 +5538,24 @@
   <other type="doi">10.2307/2005583</other>
 </article></entry>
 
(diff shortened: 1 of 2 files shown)


hulpke (approved):
Excellent -- thank you.

ssiccha (approved):

PR #4605, opened 2021-07-07
Remove duplicate entry in `COPYRIGHT` file

wilfwilson:
I noticed that `E. M. Bray` and `Erik M. Bray` both appear in the `COPYRIGHT` file; I gather that they are both @embray. Perhaps @embray has a preference one way or another about which one should remain; I'm happy to change this PR if I've guessed the wrong one.


ssiccha (approved):
I'd also go for the full first name.

PR #4607, opened 2021-07-11
kernel: adapt Julia GC code for Julia 1.7

fingolfin:
The old code did not compile cleanly with Julia 1.7. This patch
corrects that; it also modifies compilation for Julia 1.6.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -40,9 +40,13 @@
 #include <julia.h>
 #include <julia_gcext.h>
 
+#if (JULIA_VERSION_MAJOR * 100 + JULIA_VERSION_MINOR) <= 106
+
 // import jl_get_current_task from julia_internal.h, which unfortunately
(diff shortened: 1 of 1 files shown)


ssiccha (commented):
In general, this looks good to me. But where can I see that this is actually tested with Julia 1.7? Searching through the log of [this job](https://github.com/gap-system/gap/runs/3041648852?check_suite_focus=true) it looks like this is tested with Julia 1.4.2?

fingolfin:
The julia integration is not really tested here, it is tested in the GAP.jl repository. (I could in principle setup a CI test here which "tests" it with several Julia versions, but those tests really only test whether compilation succeeds and GAP starts, but they have nothing to do with how this code *really* is used, so this is IMHO mostly pointless)

ssiccha (approved):
Makes sense.

PR #4608, opened 2021-07-11
kernel: move serializer state from GAPState to stack

fingolfin:
... and split the serializer and deserializer state structs into two, to avoid
nasty casting which depends on the situation. This allows us to get rid of two
GAP_TRY/GAP_CATCH/GAP_THROW combos.

All of this makes it easier to reason about what's going on

### DIFF
--- a/src/hpc/serialize.c
+++ b/src/hpc/serialize.c
@@ -27,20 +27,20 @@
 
 #include <stdio.h>
 
-static ModuleStateOffset SerializeStateOffset = -1;
-
-typedef struct SerializeModuleState {
+typedef struct SerializerState {
+    Obj                   stack;
     Obj    obj;
     UInt   index;
-    void * dispatcher;
+    SerializerInterface * dispatcher;
     Obj    registry;
-    Obj    stack;
-} SerializeModuleState;
+} SerializerState;
 
-#define SERIALIZER ((SerializerInterface *)(MODULE_STATE(Serialize).dispatcher))
-
-#define DESERIALIZER                                                         \
-    ((DeserializerInterface *)(MODULE_STATE(Serialize).dispatcher))
+typedef struct DeserializerState {
+    Obj                     stack;
+    Obj                     obj;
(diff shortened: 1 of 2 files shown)


src/hpc/serialize.c
```diff
@@ -775,41 +736,41 @@ static Obj LookupIntTag(Obj tag)
...
             break;
         case T_POSOBJ:
-            result = DeserializeObj();
+            result = DeserializeObj(state);
             if (!IS_PLIST(result))
                 DeserializationError();
             break;
         case T_PREC:
-            result = DeserializeObj();
+            result = DeserializeObj(state);
```
ssiccha, line 773: Codecov says this line and some others in this function aren't covered by tests. Would it be relatively easy to do so and, if you think it's worth the time, would you add some?

ssiccha (approved):
I haven't looked through all the `SerializeFFE`, `SerializeCyc` etc functions, but there the calls are only adjusted to the additional state argument, if I saw that correctly.

One question wrt `SerializedAlready`: It is called at the beginning of some serialization functions, and if an object was not serialized already, then it marks it as serialized. However, what happens if the actual serialization then encounters an `SerializeError`? Are we catching that case? Do we even need to catch that? 

Looks good to me.

src/hpc/serialize.c
fingolfin: No, I won't: all the existing tests were added by me (i.e. there were zero before I did that). Whenever I add tests to this code, I uncover bugs, and then I won't be able to stop myself from trying to fix them.

As it is, I only touched this code right now because I needed to understand whether the GAP_TRY/GAP_CATCH did anything "interesting". I was able to eliminate them, which is good, as that's one place less I have to worry about...

(On the long run, we should either discard this code along with the rest of HPC-GAP, or come up with a plan to start using them. The serializer code might be of interest independent of HPC-GAP, as an alternative to IO pickling that works on non-POSIX systems, and is built into GAP itself...)

PR #4609, opened 2021-07-12
kernel: better deserialization error messages

fingolfin:
This is from an old branch I had lying around and never finished; but this commit seems useful on its own.

The added tests highlight that there are still bugs in the serializer code, but I don't have time to fix them (the abandoned branch tried to do that, by folding `SerializedAlready` into `SerializeObj`. Alas, doing that naively just broke serialization; I clearly don't understand that code well enough, so I'll rather leave it alone for now)


ssiccha (approved):
Looks sensible to me.

PR #4610, opened 2021-07-12
Fix a regression in MaximalSubgroupClassReps

fingolfin:
Also add some tests, and mark another (currently unresolved) regression: the command `MaximalSubgroupClassReps (GL(2,4));` should terminate quickly but in fact runs for a long time. I did not wait for it to complete, but it definitely is MUCH slower than in previous GAP versions.

Actually, I noticed that this got a lot faster from GAP 4.8 to 4.9: in 4.8 that test file completes for me in ~350 milliseconds, in all versions of GAP after that, including master, it is more like 1100 milliseconds. I wonder what's behind that, perhaps we can bisect the cause for this.

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -1461,7 +1461,7 @@ InstallMethod(MaximalSubgroupClassReps,"default, catch dangerous options",
 function(G)
 local H,a,m,i,l;
   # use ``try'' and set flags so that a known partial result is not used
-  m:=TryMaximalSubgroupClassReps(H:
+  m:=TryMaximalSubgroupClassReps(G:
           cheap:=false,intersize:=false,inmax:=false,nolattice:=false);
   l:=[];
   for i in m do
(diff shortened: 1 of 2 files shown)


hulpke (approved):
Yes, it obviously has to be G. No idea why it happened and I did not spot it.

hulpke:
I think the issue with the GL example is:

- The nice method installation for `CalcMaximalSubgroupClassReps` is missing the `IsHandledByNiceMonomorphism` (line 689 in grpnice.gi)
- This method should call again `CalcMaximalSubgroupClassReps`, not `Try...`
With these changes it is fast.

Sorry for these problems. The problem is in how to allow for calculation (and storage) of a subset of maximal subgroups (what can be gotten cheaply) without this messing up the full list of maximal subgroups.

However these errors now all only made the system slow, not wrong.

fingolfin:
Thanks, that works and restores performance to how it was in GAP 4.8. I've added a commit to this PR with the changes you described.

PR #4611, opened 2021-07-13
Improve docs of StabChain

ssiccha:
Use the proper option instead of a `SetSize` hack.

### DIFF
--- a/doc/ref/grpperm.xml
+++ b/doc/ref/grpperm.xml
@@ -354,9 +354,11 @@ gap> Size( h );
 <P/>
 The result is still correct, and the running time is actually somewhat
 slower.
-If you give the algorithm additional information so that it can check
-its results,
-things become faster and the result is guaranteed to be correct.
+If you give the algorithm the order of the group, then it can check
+its result, and so things become faster and the result is guaranteed
+to be correct. This can be done with the <C>size</C> option
+(see&nbsp;<Ref Func="StabChain" Label="for a group (and a record)"/>),
+or by setting the size of the group beforehand with <C>SetSize</C>.
 <P/>
 <Log><![CDATA[
 gap> h:=Group( GeneratorsOfGroup( op ) );;


doc/ref/grpperm.xml
```diff
@@ -360,8 +360,7 @@ things become faster and the result is guaranteed to be correct.
 <P/>
 <Log><![CDATA[
 gap> h:=Group( GeneratorsOfGroup( op ) );;
-gap> SetSize( h, 2317591180800 );
-gap> StabChain( h );;  time;
+gap> StabChain( h, rec( limit := 2317591180800 ) );;  time;
 170
```
fingolfin, line 364: Perhaps we should take a bigger example; for me the timings are not 1410 and 170 but rather 15 and 5 😂

fingolfin (approved):
Looks good to me

hulpke:
I don't have an objection, but I would not consider `SetSize` a hack. In particular when walling `StabChain` only through other functions (say `SylowSubgroup`), it is the only way to provide size information.

hulpke (commented):
Why do you use `limit` (which is an upper bound) and not `size` ?

ssiccha:
I may have been a bit overzealous calling using `SetSize` a hack, but it is _not documented_ in `StabChain` that `Size`, if known, will be used during the computation. That probably should just be added to the docs of `StabChain`.

I use `limit` because it is a little more general. I guess `size` fits better here. In either case, a reference to the documentation of the options would make sense here.

fingolfin:
So, how to proceed here? Merge as is? Change to `size := ...`? Add some text which references the options, and says something roughly like this:

> In the example below we used the `limit` option; here we could also have used the `size` option. Also `SetSize(G, ...)` prior to the StabChain call would have worked

wilfwilson (commented):
Thanks @ssiccha for working to improve this.

I agree with the essence of what @hulpke is saying. I think `SetSize` was fine, but in any case, I think the `size` option is more appropriate here than `limit`.

If you want to go further, I suggest doing something like changing the paragraph preceding this example to following:
```
If you give the algorithm the order of the group, then it can check
its result, and so things become faster and the result is guaranteed
to be correct. This can be done with the <C>size</C> option
(see&nbsp;<Ref Func="StabChain" Label="for a group (and a record)"/>),
or by setting the size of the group beforehand with <C>SetSize</C>.
```
I don't think it's useful to mention the `limit` option here.

I second @fingolfin's suggestion of making the example more difficult, but that doesn't need to be done in this PR.

wilfwilson (approved):
I'm happy with what the new sentences say, although I'm biased since I basically wrote it 😉

PR #4612, opened 2021-07-16
Make `MappedWord` handle inverses better

### LINKED ISSUE #4606: Optionally allow passing generator inverses to `MappedWord` ?
Right now, if applying `MappedWord` to words in a free group, it (re)computes inverses of the image generators (resp. "divides" by them). This is fine if inversion / "division" is cheap (such as for words, or permutations), but not so much if it is expensive (e.g. big matrices, automorphisms, ...).

So perhaps we could add another (optional) argument to `MappedWord` which receives a list of the inverses of the generators? Or perhaps there is something better we could do?

Perhaps @hulpke has some insights about this.

hulpke:
This is in response to #4606. So far only implemented for the letter representation of associative words. One would have to add similar changes for the syllable represenation and the generic method.

The code:
a) Precomputes inverses once (if they are to be used more than once)
b) Uses inverses that are given as extra generators

For example, just taking larger matrices so we see multiplication cost
```
gap> g:=AtlasGroup("M24",IsMatrixGroup,true,Dimension,483);;
gap> epi:=EpimorphismFromFreeGroup(g);;
gap> mgi:=MappingGeneratorsImages(epi);;
gap> w:=PseudoRandom(Source(epi):radius:=100);;  # random word
gap> v0:=MappedWord(w,mgi[1],mgi[2]);;time;  # old code
4147
``` 

This is the base-line for the existing code, which uses w/gen all the time to handle inverses

Here is the new code that caches inverses:
```
gap> v1:=MappedWord(w,mgi[1],mgi[2]);;time;  # new code
2412
```

Clearly a speedup. And if we give the inverses as extra generators, a little bit extra bump

```
gap> mgj:=List(mgi,x->Concatenation(x,List(x,Inverse)));; # also give inverses as extra generators
gap> v3:=MappedWord(w,mgj[1],mgj[2]);;time; # slight speedup
2323
```

This closes #4606

### DIFF
--- a/lib/word.gd
+++ b/lib/word.gd
@@ -343,8 +343,8 @@ DeclareCategoryFamily( "IsNonassocWordWithOne" );
 ##  The lists <A>gens</A> and <A>imgs</A> must of course have the same length.
 ##  <P/>
 ##  <Ref Oper="MappedWord"/> needs to do some preprocessing to get internal
-##  generator numbers etc. When mapping many (several thousand) words, an
-##  explicit loop over the words syllables might be faster.
+##  generator numbers etc. When mapping many (several thousand) words, a
+##  dedicated loop might be faster.
 ##  <P/>
 ##  For example, if the elements in <A>imgs</A> are all
 ##  <E>associative words</E>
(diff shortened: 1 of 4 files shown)


fingolfin (approved):

fingolfin (approved):
This looks good to me. Of course ideally, the feature should also be mentioned in the manual entry for `MappedWord`, and perhaps illustrated with an example (which then would also serve as a minimal test case for the new code)

hulpke:
Thank you, @fingolfin 
Then I assume this could be considered to resolve #4606 and I'll add methods for the other two representations and a manual entry

fingolfin:
Yes, please :-)

fingolfin:
(There's also a typo in the commit message: "kip" -> "keep"

PR #4615, opened 2021-07-22
kernel: get rid of CATCH_ERROR macro

fingolfin:
It turns out all cases where it was used are now redundant: they were all about calling `IntrAbortCoding`. But since all the interpreter and coder state is on the stack now, and also `IntrEnd` invokes `IntrAbortCoding` for us if there was an error, we can just get rid of all these cases.

Oh, and once an error was signaled (meaning `rs->s.NrError` is not equal 0), it stays this way (`rs->s.NrError` is never reset / decremented until the end of `ReadEvalCommand`), all the following `TRY_IF_NO_ERROR` blocks will be skipped, thus all calls into the interpreter (and hence transitively: into the coder) will be skipped. So I see no way how this could cause issues.

My main motivation here was/is to simplify the code; I have been trying to untangle the error handling setup, and needed to understand what each "catch" does. And realized those using `CATCH_ERROR` do effectively nothing, so...

That said, I am still only about 99% certain my analysis is correct; but of course that may just be my lack of imagination :-). Make of that what you will, but I think if there are things this breaks, then (a) we should know, and add tests cases for them, and (b) the most efficient way to find out is to try with this patch and wait for someone to complain (though if anyone can *think* of a scenario were this patch "breaks" something, please let me know).

Note that anything "broken" by this would simply change from one kind of error to another (probably less helpful) error, so I think the potential negative impact from this is fairly low.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -39,22 +39,14 @@
 /****************************************************************************
 **
 *S  TRY_IF_NO_ERROR
-*S  CATCH_ERROR
 **
 **  To deal with errors found by the reader, we implement a kind of exception
-**  handling using setjmp, with the help of these two macros. See also
+**  handling using setjmp, with the help of this macro. See also
 **  GAP_TRY and GAP_CATCH in trycatch.h for two closely related macros.
 **
-**  To use these constructs, write code like this:
+**  To use this construct, write code like this:
 **    TRY_IF_NO_ERROR {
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
This seems right, and simplifying this area is always a good idea

ChrisJefferson (approved):

PR #4616, opened 2021-07-23
kernel: add missing GAP_CATCH handlers

fingolfin:
In the past, not closing an output was not too bad; now it can lead to crashes.
At least when using this code in an embedded GAP (i.e., libgap).

This patch looks big, but that's due to whitespace changes. If you instruct GitHub to hide whitespace ([click here to see the result](https://github.com/gap-system/gap/pull/4616/files?diff=unified&w=1) or better yet, look at each commit separately), you can see that this PR is actually quite small.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -157,13 +157,8 @@ static Int READ_COMMAND(TypInputFile * input, Obj *evalResult)
 */
 Obj READ_ALL_COMMANDS(Obj instream, Obj echo, Obj capture, Obj resultCallback)
 {
-    ExecStatus status;
-    UInt       dualSemicolon;
-    Obj        result, resultList;
-    Obj        copy;
-    Obj        evalResult;
-    Obj        outstream = 0;
-    Obj        outstreamString = 0;
+    volatile Obj outstream = 0;
+    volatile Obj outstreamString = 0;
 
     RequireInputStream("READ_ALL_COMMANDS", instream);
 
@@ -184,18 +179,25 @@ Obj READ_ALL_COMMANDS(Obj instream, Obj echo, Obj capture, Obj resultCallback)
         return Fail;
     }
 
-    resultList = NEW_PLIST(T_PLIST, 16);
+    volatile Obj resultList = NEW_PLIST(T_PLIST, 16);
+    BOOL rethrow = FALSE;
 
-    do {
-        if (outstream) {
-            // Clean in case there has been any output
-            SET_LEN_STRING(outstreamString, 0);
-        }
+    GAP_TRY
+    {
+        while (1) {
+            if (outstream) {
+                // Clean in case there has been any output
+                SET_LEN_STRING(outstreamString, 0);
+            }
 
-        status = ReadEvalCommand(0, &input, &evalResult, &dualSemicolon);
+            UInt dualSemicolon;
+            Obj  evalResult;
 
-        if (!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))) {
-            result = NEW_PLIST(T_PLIST, 5);
+            ExecStatus status = ReadEvalCommand(0, &input, &evalResult, &dualSemicolon);
+            if (status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))
+                break;
+
+            Obj result = NEW_PLIST(T_PLIST, 5);
             AssPlist(result, 1, False);
             PushPlist(resultList, result);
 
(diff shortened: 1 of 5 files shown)


fingolfin:
Oh and I need this for the GAP-Julia interface

fingolfin:
There is a diff in `testspecial` for the output of `tst/testlibgap/trycatch`. I addressed part of it, but one diff remains: it executes
```
    RegisterTryCatchHandler(handle_trycatch);
    test_eval("Display(CALL_WITH_CATCH(function() return 314; end, []));;");
    test_eval("Display(CALL_WITH_CATCH(function() return [][1]; end, []));;");
```


The last line should produce this:
```
gap> Display(CALL_WITH_CATCH(function() return [][1]; end, []));;
Entering GAP_TRY section
Error, List Element: <list>[1] must have an assigned value
Caught error in GAP_TRY section
[ false, 0 ]
```

But instead, with this PR, we get:
```
gap> Display(CALL_WITH_CATCH(function() return [][1]; end, []));;
Entering GAP_TRY section
Error, List Element: <list>[1] must have an assigned value
Leaving GAP_TRY section
[ false, 0 ]
```

Note the difference in the penultimate line. Right now I have no clue at all what might be causing that... perhaps @ChrisJefferson has one?

ChrisJefferson:
It's easier to see what's going on if we make the `handle_trycatch` function print every time it is run. I added "printf("L: %d %d\n", level, mode);", in master we get:

```
# Initializing GAP...
gap> OnBreak := false;;

gap> MakeReadWriteGVar("ERROR_OUTPUT");

gap> ERROR_OUTPUT := MakeImmutable("*stdout*");;

gap> Display(CALL_WITH_CATCH(function() return 314; end, []));;
L: 0 0
Entering GAP_TRY section
L: 1 1
Leaving GAP_TRY section
[ true, 314 ]

gap> Display(CALL_WITH_CATCH(function() return [][1]; end, []));;
L: 0 0
Entering GAP_TRY section
L: 1 0
L: 2 1
Error, L: 1 0
L: 2 1
List Element: <list>[1] must have an assigned valueL: 1 0

L: 2 1
L: 1 2
Caught error in GAP_TRY section
[ false, 0 ]
```

in the PR we get:

```
# Initializing GAP...
gap> OnBreak := false;;

gap> MakeReadWriteGVar("ERROR_OUTPUT");

gap> ERROR_OUTPUT := MakeImmutable("*stdout*");;

gap> Display(CALL_WITH_CATCH(function() return 314; end, []));;
L: 0 0
Entering GAP_TRY section
L: 1 0
L: 2 1
L: 1 1
Leaving GAP_TRY section
[ true, 314 ]

gap> Display(CALL_WITH_CATCH(function() return [][1]; end, []));;
L: 0 0
Entering GAP_TRY section
L: 1 0
L: 2 0
L: 3 1
Error, L: 2 0
L: 3 1
List Element: <list>[1] must have an assigned valueL: 2 0

L: 3 1
L: 2 2
L: 1 1
Leaving GAP_TRY section
[ false, 0 ]
```

ChrisJefferson:
In short, I think the 'depth' needs updating to 2, as we want to see if 'CALL_WITH_CATCH' is correctly seeing and dealing with an error, as there is now another error level (catching errors in the stream, which doesn't happen in this case). However, I'd be tempted to just change the test to print out it's value at every level -- it would make it slightly more fragile, but it wouldn't hurt to keep track of changes to streams + call_with_catch.

fingolfin:
Ohhh, I didn't realize (i.e., didn't look properly at all) that `handle_trycatch` only shows "some" events. With that, it is indeed unsurprising that its behavior changed.

And I agree that the test should just be simplified to print more info, even if that might mean it needs to be updated more often; but that is a good thing if it means that we notice certain changes explicitly.

fingolfin:
@ChrisJefferson I've improved the trycatch test

src/streams.c
```diff
@@ -185,53 +180,66 @@ Obj READ_ALL_COMMANDS(Obj instream, Obj echo, Obj capture, Obj resultCallback)
...
 
-        status = ReadEvalCommand(0, &input, &evalResult, &dualSemicolon);
+            UInt dualSemicolon;
+            Obj  evalResult;
 
-        if (!(status & (STATUS_EOF | STATUS_QUIT | STATUS_QQUIT))) {
-            result = NEW_PLIST(T_PLIST, 5);
-            AssPlist(result, 1, False);
-            PushPlist(resultList, result);
+            status = ReadEvalCommand(0, &input, &evalResult, &dualSemicolon);
```
fingolfin, line 198: After thinking about this PR again, I feel I need to clarify something: `ReadEvalCommand` should never let an exception through, it "swallows" them all and turns any error into a flag in its `status` return value. This is why I did not add `GAP_TRY` to other places in `streams.c` which use `OpenInput`

However, the call to `resultCallback` may throw an exception, and we ought to deal with that (which we now do).

ChrisJefferson (approved):

wilfwilson (approved):
Looks sensible to me too.

PR #4617, opened 2021-07-23
kernel: track try/catch nesting depth

fingolfin:
- add `STATE(TryCatchDepth)` to track how many `GAP_TRY` have been nested (and thus are "active" at the same time).
- merge `src/sysjmp.h` into `src/trycatch.h`
- rename `src/sysjmp.c` to `src/trycatch.c`
- rename `RegisterSyLongjmpObserver` to `RegisterThrowObserver`, and change the argument type: observers now take a single int argument (instead of taking no arguments at all); that argument is set to `STATE(TryCatchDepth)` (so it is merely for convenience of the observer code)

(The fourth point affects @ChrisJefferson unreleased "Debugger" package but that should be easy to resolve there.)

The motivation for tracking the depth and reporting it to the throw observer callback is the Julia backend: there, we install an observer callback which checks if it was called at depth 0 (i.e., we are about to GAP_THROW without any surrounding GAP_TRY/ GAP_CATCH pair), and if so, throws a Julia exception instead. This way, we don't have to surround all calls into the GAP kernel with GAP_TRY / GAP_CATCH -- this simplifies the interface, but it is also an optimization for the common case, namely that no error occurs.


ChrisJefferson (approved):

fingolfin:
Some more testing on the Julia side sadly revealed that I need to do some more work on this (and perhaps return some things from an older version of this patch that I stripped for the PR).

fingolfin:
Actually, all is fine after all -- I just got stumped by the fact that evaluating commands via `READ_ALL_COMMANDS` "eats" all the errors

PR #4618, opened 2021-07-24
kernel: simplify FuncREAD_COMMAND_REAL implementation

fingolfin:
... by merging the READ_COMMAND helper into it. This enables some code simplifications (which to me were surprising, i.e., I clearly did not fully understand the code before).

I believe verifying this change is not that easy, I arrived at it in incremental steps. I recommend viewing the diff in "split view mode" (not as a "unified diff") so that one can more easily compared the complete old to the complete new code.

One of the things that surprised me was the helper function would set `STATE(UserHasQuit)` resp. `STATE(UserHasQUIT)` to 1 and the main function would promptly reset them to 0. Oh, and if they were set to 1 before (e.g. before entering `FuncREAD_COMMAND_REAL`, or as a side effect of `ReadEvalCommand`), then the helper returns immediately, and then `FuncREAD_COMMAND_REAL` returns immediately; this is unchanged.

The one drawback of this change is that the similarity of the code to `READ_INNER` is reduced. But I think that's fair, esp. since that "similarity" IMHO was misleading, as discussed in the preceding paragraph.

Oh, and note the FIXME comment I added... I am not sure about that one. Frank added it 2009-09-25 in "old history commit" f9ad911331c99b6407afcbfa6ae9bd83aa120b48 with this commit message:

>  Make sure that the recursion depth counter is reset to zero when quitting a
>  break loop. Problem reported by Simon King, caused by (non-elegant) Sage code
>  that entered and left many break loops.

But back then, we did not save & restore the recursion depth counter in `GAP_TRY`/`GAP_CATCH`/`TRY_IF_NO_ERROR` (resp. their predecessor code -- those macros did not exist back then). So I kinda expect we don't need that `SetRecursionDepth(0)` call anymore (nor its sibling in `READ_COMMAND`). We should test that theory, though.

### DIFF
--- a/src/streams.c
+++ b/src/streams.c
@@ -84,33 +84,6 @@ static UInt OpenInputFileOrStream(const char *   funcname,
                       "must be a string or an input stream");
 }
 
-static Int READ_COMMAND(TypInputFile * input, Obj *evalResult)
-{
-    ExecStatus    status;
-
-    status = ReadEvalCommand(0, input, evalResult, 0);
-    if( status == STATUS_EOF )
-        return 0;
-
-    if ( STATE(UserHasQuit) || STATE(UserHasQUIT) )
-        return 0;
-    
-    /* handle return-value or return-void command                          */
-    if ( status & (STATUS_RETURN_VAL | STATUS_RETURN_VOID) ) {
-        Pr("'return' must not be used in file read-eval loop\n", 0, 0);
-    }
-
-    /* handle quit command                                 */
-    else if (status == STATUS_QUIT) {
-        STATE(UserHasQuit) = 1;
-    }
-    else if (status == STATUS_QQUIT) {
-        STATE(UserHasQUIT) = 1;
-    }
-
-    return 1;
-}
-
 /****************************************************************************
 **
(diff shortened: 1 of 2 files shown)


fingolfin:
I made up my mind how to resolve that `FIXME`, see PR #4619

fingolfin:
I had to rebase this since PR #4619 was merged. And I actually made a change that affects behavior: namely `quit` now is treated like `QUIT`, as a "failed command". I would argue that this is more in line with what `READ`, `READ_ALL_COMMANDS`, `READ_TEST_OR_LOOP` etc. do: they all treat `quit` and `QUIT` as equivalent (only `READ_GAP_ROOT` ignores `QUIT` but not `quit`, which seems like a bug to me).

Note that `READ_COMMAND_REAL` is an undocumented command called by a few packages. I checked and I believe this change does not matter for any of them. Still, if there is concern, I can also revert it.

fingolfin:
Oh and I added a test file for `READ_COMMAND_REAL` which shows this change, and tests a few other cases

ChrisJefferson (approved):

PR #4619, opened 2021-07-25
kernel: reduce calls to `SetRecursionDepth(0)`

fingolfin:
These were added 2009-09-25 to resolve issues where the recursion
depth counter would "overflow" after entering and leaving the break
loop repeatedly during recursive execution.

But these days, we save and restore the execution depth in `GAP_TRY` and
`GAP_CATCH` as well as in `TRY_IF_NO_ERROR` and `Shell`; and at the start
of each REPL iteration in `Shell`, it is set to 0 freshly.

I've verified this experimentally by printing the recursion depth in
the spots were we formerly reset it, and this seems to confirm that the
explicit reset is no longer necessary.


ChrisJefferson:
Oh, nice this got fixed as a side effect of other work.

ChrisJefferson (approved):

PR #4620, opened 2021-07-27
buildsys: generate cnf/GAP-VERSION-FILE in builddir, not srcdir

fingolfin:
Previously I thought this file could be "safely" shared between multiple GAP
builds. While that's technically true, it can be irritating if one does a
fresh build on otherwise unchanged sources that were last compiled a couple
days/weeks/months ago: the resulting GAP executable will print the build
time from that last build, not the new build, which can be confusing.

Letting each build track this on its own avoids that confusion.

Reported by @ThomasBreuer

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -1289,11 +1289,12 @@ cnf/GAP-OBJS: FORCE
 # should we need to.
 ########################################################################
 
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
This looks good.
(After `make distclean`, I can run `./configure` and `make`, and this works.)

Issue #4208, opened 2021-01-07
libgap: add GAP_ValueSmallInt or something like it


embray:
### Observed behaviour

The libgap API currently exposes a `double GAP_ValueMacFloat(Obj)` function to return a machine double from a GAP machine float.  However, no such interface exists for native ints.  `GAP_SizeInt` and `GAP_AddrInt` can be used to read the limbs of large ints, but converting a small int back to a C int requires knowledge of the internal structure of immediate ints.

### Expected behaviour

A proposed `Int GAP_ValueSmallInt(Obj)` function (or some other name if there's a better one) to return an `Int/intptr_t` from a GAP small int should be added to the API so that the internal structure of small ints does not need to be exposed.

fingolfin:
Agreed!

If you have time to add this, I'll be happy to merge the PR. Otherwise, gotta wait till I have some spare cycles.

embray:
I sort of forget about this since my workaround is working for now.  But my workaround definitely just depends on explicitly knowing GAP internals.  So it would be good to add this.   I just wonder what it should do if it's passed the wrong type of object.

PR #4621, opened 2021-07-27
Add `GAP_NewRange`, `GAP_NewObjIntFromInt`, `GAP_ValueInt` to the libgap API

fingolfin:
This is cleanup, but it will also be used in the GAP-Julia interface. Might also be useful for other libgap users.

<s>TODO: This needs some more work, in particular the libgap API functions needs to be documented.</s>

### DIFF
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -355,6 +355,16 @@ Obj GAP_MakeObjInt(const UInt * limbs, Int size)
     return MakeObjInt(limbs, size);
 }
 
+Obj GAP_NewObjIntFromInt(Int val)
+{
+    return ObjInt_Int(val);
+}
+
+Int GAP_ValueInt(Obj obj)
+{
+    return Int_ObjInt(obj);
+}
+
 Int GAP_SizeInt(Obj obj)
 {
     RequireInt("GAP_SizeInt", obj);
--- a/src/libgap-api.c
+++ b/src/libgap-api.c
@@ -33,6 +33,7 @@
 #include "opers.h"
 #include "plist.h"
 #include "precord.h"
+#include "range.h"
 #include "records.h"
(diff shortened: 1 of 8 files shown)


src/libgap-api.c
```diff
@@ -406,6 +407,21 @@ Obj GAP_NewPlist(Int capacity)
...
+static BOOL fitsInIntObj(Int i)
+{
+    return INT_INTOBJ_MIN <= i && i <= INT_INTOBJ_MAX;
+}
+
+Obj GAP_NewRange(Int len, Int low, Int inc)
+{
+    if (!fitsInIntObj(len)) return 0;
+    if (!fitsInIntObj(low)) return 0;
+    if (!fitsInIntObj(inc)) return 0;
```
ChrisJefferson, line 419: Do you just want to catch inc==0 here too? I don't think it's allowed.

fingolfin: Good point, done

fingolfin:
I've now enhanced this a bit more to resolves issue #4208 by @embray who also may have comments on this PR.

ThomasBreuer (approved):
Just for curiosity: Why `fitsInIntObj` and not `fitsInObjInt`?
(`ObjInt` is used in most situations in the `src` files.)

(And there are two instances of "an GAP integer object" in comments.)

fingolfin:
@ThomasBreuer "intobj" refers to immediate ints (cf. `IS_INTOBJ`, `INT_INTOBJ`, and many more), while objint refers to any kind of GAP integer, i.e., "an `Obj` representing an integer"; the latter is relatively rarely used, though, mostly in `ObjInt_Int` and `Int_ObjInt` (and there many variants; but the variants all derive from those two; as does `MakeObjInt`).

It is rather confusing.

PR #4622, opened 2021-07-27
Speedup precomputation in AlgebraicExtension

frankluebeck:
When creating an algebraic extension K[X]/(f(X)), say of degree n,
then GAP precomputes the reduced representatives of X^i, n <= i <= 2n-2,
modulo f(X). With this patch this is done in n-2 row operations instead
of (n-1)*(n-2)/2. This yields a significant speedup already for moderate
degrees.

## Text for release notes
'AlgebraicExtension(NC)' is now faster for larger degrees.

### DIFF
--- a/lib/algfld.gi
+++ b/lib/algfld.gi
@@ -65,7 +65,7 @@ end;
 InstallMethod(AlgebraicElementsFamily,"generic",true,
   [IsField,IsUnivariatePolynomial,IsBool],0,
 function(f,p,check)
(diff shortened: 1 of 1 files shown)


hulpke (approved):
Makes complete sense. Thank you!

fingolfin (commented):
Thank you, this looks very nice.

Would it be possible to add a test for this code, even a very simple one? As it is, the code is not at all exercised by our test suite

PR #4623, opened 2021-07-30
Fix DescribesInvariantQuadraticForm, etc

fingolfin:
We recently introduced IsMatrixOrMatrixObj and these functions (which we recently added via PR #4489) need to use it. Otherwise the forms package addons for GO, SO, GU, SU etc. won't work.


ThomasBreuer (approved):
Yes, of course. Thanks for fixing this.

PR #4625, opened 2021-08-02
Ranges: improve printing and conversion to strings

fingolfin:
- Add tests for printing ranges & converting them to strings
- Improve String/ViewString/PrintString for ranges

The diff between the first and second commit shows what changed.

Perhaps we should have similar tests for other library objects?

This patch has been motivated by the CI failure I observed here: https://github.com/gap-packages/fr/pull/38/checks?check_run_id=3160342361

### DIFF
--- a/tst/testinstall/range.tst
+++ b/tst/testinstall/range.tst
@@ -40,7 +40,7 @@ gap> TestPrintRangeRep:=function(r)
 #
 gap> TestPrintRangeRep([0..1]);
 Display: [ 0 .. 1 ]
-DisplayString: <object>
+DisplayString: [ 0 .. 1 ]
 ViewObj: [ 0, 1 ]
 ViewString: [ 0, 1 ]
 PrintObj: [ 0 .. 1 ]
@@ -48,31 +48,31 @@ PrintString: [ 0 .. 1 ]
 String: [ 0 .. 1 ]
 gap> TestPrintRangeRep([0..2]);
 Display: [ 0 .. 2 ]
-DisplayString: <object>
+DisplayString: [ 0 .. 2 ]
 ViewObj: [ 0 .. 2 ]
-ViewString: [ 0, 1, 2 ]
+ViewString: [ 0 .. 2 ]
 PrintObj: [ 0 .. 2 ]
 PrintString: [ 0 .. 2 ]
 String: [ 0 .. 2 ]
 gap> TestPrintRangeRep([0,2..4]);
 Display: [ 0, 2 .. 4 ]
-DisplayString: <object>
+DisplayString: [ 0, 2 .. 4 ]
 ViewObj: [ 0, 2 .. 4 ]
-ViewString: [ 0, 2, 4 ]
+ViewString: [ 0, 2 .. 4 ]
 PrintObj: [ 0, 2 .. 4 ]
 PrintString: [ 0, 2 .. 4 ]
 String: [ 0, 2 .. 4 ]
 gap> TestPrintRangeRep([0,-1..-2]);
 Display: [ 0, -1 .. -2 ]
-DisplayString: <object>
+DisplayString: [ 0, -1 .. -2 ]
 ViewObj: [ 0, -1 .. -2 ]
-ViewString: [ 0, -1, -2 ]
+ViewString: [ 0, -1 .. -2 ]
 PrintObj: [ 0, -1 .. -2 ]
 PrintString: [ 0, -1 .. -2 ]
 String: [ 0, -1 .. -2 ]
 gap> TestPrintRangeRep([0,-1..-1]);
 Display: [ 0, -1 .. -1 ]
-DisplayString: <object>
+DisplayString: [ 0, -1 .. -1 ]
 ViewObj: [ 0, -1 ]
 ViewString: [ 0, -1 ]
 PrintObj: [ 0, -1 .. -1 ]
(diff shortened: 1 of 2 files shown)


tst/testinstall/range.tst
```diff
@@ -1,16 +1,72 @@
...
+gap> TestPrintRangeRep([0..2]);
+ViewObj: [ 0 .. 2 ]
+PrintObj: [ 0 .. 2 ]
+Display: [ 0 .. 2 ]
+DisplayString: <object>
+ViewString: [ 0 .. 2 ]
+PrintString: [ 0 .. 2 ]
+String: [ 0 .. 2 ]
+gap> TestPrintRangeRep([0,2..4]);
+ViewObj: [ 0, 2 .. 4 ]
```
fingolfin, line 46: I guess we could go one step further and also print `[ 0, 2 .. 4 ]` as `[ 0, 2, 4 ]`...

tst/testinstall/range.tst
```diff
@@ -1,16 +1,72 @@
...
+PrintString: [ 0, -1 .. -2 ]
+String: [ 0, -1 .. -2 ]
+gap> TestPrintRangeRep([0,-1..-1]);
+ViewObj: [ 0, -1 ]
+PrintObj: [ 0, -1 ]
+Display: [ 0, -1 ]
+DisplayString: <object>
+ViewString: [ 0, -1 ]
+PrintString: [ 0, -1 ]
+String: [ 0, -1 ]
```
fingolfin, line 68: Note: perhaps we do want this to produce `[ 0, -1 .. -1 ]` in a few case; and like wise, `[ 0 .. 1 ]` instead of `[ 0, 1 ]`. But then IMHO these two should behave consistenly: either both kinds of ranges of length 2 are shown "as range", or both are shown "as lists". Also, of course `ViewString` and `ViewObj` should be consistent, and likewise `PrintString` and `PrintObj`. (Well... and `DisplayString`, but I did not even try to fix that in this OR)

wilfwilson:
I don't have time to properly review at the moment, but I do like the general idea from what I can get at a glance. However, I worry that it suffers from the same problem as #4050, i.e. breaks the tests of various packages, which will make it annoying to merge.

tst/testinstall/range.tst
ThomasBreuer: The input `[ 0, 2, 4 ]` does not produce an object in `IsRangeRep`, whereas `[ 0, 2 .. 4 ]` creates such an object.
In this sense, I would not be in favour of this change.

fingolfin: Note that the documentation does not make any claim that the *representation* is preserved by reentering the output of Print; just that an equal object is returned. So both print outputs conform to the documentation.

In any case, always printing ranges as ranges certainly is also an option; the main motivation for this PR really is to make it behave consistently in one way or the other.

wilfwilson: I would support this kind of change, if you want to make it.

tst/testinstall/reesmat.tst
```diff
@@ -270,7 +270,7 @@ false
 
 # IsReesMatrixSemigroup: for a Rees matrix subsemigroup with generators, 1
 gap> R := ReesMatrixSemigroup(SymmetricGroup(2), [[(1,2)]]);
-<Rees matrix semigroup 1x1 over Sym( [ 1 .. 2 ] )>
+<Rees matrix semigroup 1x1 over Sym( [ 1, 2 ] )>
```
wilfwilson, line 273: There are quite a few of these kinds of changes in Semigroups and Digraphs, (almost?) all of which involve `Sym( [ 1 .. 2 ] )` becoming `Sym( [ 1, 2 ] )`

wilfwilson (approved):
I've looked at this in more detail now and I do support it the changes, in particular the consistency and the showing length-two ranges like `[1..2]` as `[1,2]`. I would support further such changes in this direction (e.g. [1,3..5] being [1,3,5]).

@ThomasBreuer raises a good point about the `\>` and `\<` hints.

I agree with @ThomasBreuer that this PR should not be anyway near as disruptive as #4050, but it causes at least some problems. Running the standard tests of Digraphs and Semigroups with this PR I get:
* Digraphs: `6 failures in 2 of 16 files`
* Semigroups: `11 failures in 3 of 67 files`

It should be easy to fix such problems by suppressing output. Hopefully it will be easy enough to detect affected packages.

Note that the CI is currently not passing.

alex-konovalov (commented):
Would be nice to have - will impact package tests, but with the advent of https://github.com/gap-system/PackageDistro/ that can be then systematically addressed. Some packages may need more work to adapt examples to pass in both new and old GAP versions.

fingolfin:
I have no revised the changes in this PR, going in a somewhat different direction: instead of trying to align ViewObj with PrintObj and Display, I've opted to only "fix" `ViewString` for ranges to produce a string corresponding to the output of `ViewObj`. Also, I've now added a `DisplayString` method matching `Display`.

The result is still a bit odd because now the `View*` methods (still) behave differently from the `Display*` and `Print*` methods. But at least one inconsistency was resolved, and I don't think there will be any conflicts with packages. If someone wants to resolve the remaining inconsistency one day, be my guess, but I don't think it's all that important.

wilfwilson (approved):

pedritomelenas (commented):
I wonder if lists with three elements like [0,2..4]  should be “printed” as [0, 2, 4] instead of [0, 2 .. 4]

fingolfin:
"Should"? Dunno.... "Could"? Sure. But any such change has a cost, you need to at least update all packages whose test suites depend on the printing. IMHO it is not worth the trouble

pedritomelenas:
Sure, I should have used could.

PR #4628, opened 2021-08-17
Cleanup some GAP library code related to iostreams

fingolfin:
- lib: move ReadAllLine methods into iostream section
- lib: some iostream tweaks

### DIFF
--- a/lib/streams.gi
+++ b/lib/streams.gi
@@ -707,6 +707,17 @@ function( stream, pos )
     return SEEK_POSITION_FILE( stream![1], pos );
 end );
 
+#############################################################################
+##
+#M  FileDescriptorOfStream( <input-text-file> )
+##
+
+InstallMethod(FileDescriptorOfStream,
+        [IsInputTextStream and IsInputTextFileRep],
+        function(stream)
+    return FD_OF_FILE(stream![1]);
+end);
+
 
 #############################################################################
 ##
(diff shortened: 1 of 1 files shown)


lib/streams.gi
```diff
@@ -1518,6 +1546,8 @@ InstallOtherMethod( ReadAllLine, "iostream",
 # For an input stream that is not an input/output stream it's really
 # inappropriate to call ReadAllLine. We provide the functionality of
 # ReadLine only, in this case.
+# TODO: actually, why do we do this??? it seems better to simply produce
```
wilfwilson, line 1549: Good question. However, it's documented this way so perhaps not easy to change:
> If the first argument is an input stream but not an input/output stream then `ReadAllLine` behaves as if ReadLine (10.3-4) was called with just the first argument and any additional arguments are ignored.

lib/streams.gi
```diff
@@ -1497,6 +1519,12 @@ InstallMethod( ReadAllLine, "iostream,boolean,function",
     return line;
 end);
 
+InstallMethod( ReadAllLine, "iostream,boolean,function",
+        [ IsInputOutputStream, IsBool, IsFunction ],
+    function(iostream, nofail, IsAllLine)
+    ErrorNoReturn("not implemented");
```
wilfwilson, line 1525: Good, I like these kinds of errors.

wilfwilson (approved):

PR #4547, opened 2021-06-07
Improve chapter 4 "The Programming Language" of the reference manual

fingolfin:

PR #4632, opened 2021-08-19
Document and unify the order in which GAP evaluates function arguments and options

### LINKED ISSUE #4631: Unexpected behaviour of options
Consider the following piece of code:
```
myfunc1 := function( )
	Display( ValueOption( "myopt" ) );
	return 1;
end;

myfunc2 := function( )
	return IdFunc( myfunc1( ) : myopt := "myopt_value" );
end;

x := myfunc2( );
x := IdFunc( myfunc1( ) : myopt := "myopt_value" );
```

### Observed behaviour
The output is:
```
myopt_value
fail
```

### Expected behaviour

I would expect the output to be
```
fail
fail
```
(or at least the same output in both lines) because I do not see a reason why the option should propagate into the call of `myfunc1` in the first case but not in the second.

```
 ┌───────┐   GAP 4.12dev-1020-g1fe30f7 built on 2021-08-19 10:21:04+0200
 Packages:   Browse 1.8.9, GAPDoc 1.dev, IO 4.7.0dev, PrimGrp 3.4.0, SmallGrp 1.4.1, TransGrp 2.0.5
```
Also happens with GAP 4.11.1.

fingolfin:
Previously, this was what the immediate interpreter did, but *not*
what the executor did; the latter always evaluated options first.

Fixes #4631

### DIFF
--- a/doc/ref/language.xml
+++ b/doc/ref/language.xml
@@ -1130,15 +1130,13 @@ Remember that a variable is a location in a &GAP; program
 that points to a value. Thus for  each formal argument and for each
 formal local such a location is allocated.
 <P/>
-Next the arguments <A>arg-expr</A>s are evaluated, and the values are assigned
+Next the arguments <A>arg-expr</A>s are evaluated from left to right,
+and the values are assigned
 to the newly created variables corresponding to the formal arguments. Of
 course the first value is assigned to the new variable corresponding to
 the first formal argument, the second  value  is assigned to the new
 variable corresponding  to  the second  formal argument, and  so on.
-However, &GAP; does not make any guarantee about the order in which the
-arguments are evaluated. They might be evaluated left to right, right to
-left, or in any other order, but each argument is evaluated once. An
-exception again occurs if the last formal argument has
+An exception again occurs if the last formal argument has
 the name <C>arg</C>. In this case the values of all the actual
 arguments not assigned to the other formal parameters are
 stored in  a list and this  list is assigned to the  new variable
@@ -1200,6 +1198,8 @@ are separated from the actual arguments by a colon <C>:</C> and have much
 the same syntax as the components of a record expression. The one
 exception to this is that a component name may appear without a value,
 in which case the value <K>true</K> is silently inserted.
+<P/>
+Options are evaluated from left to right, but only after all arguments have been evaluated.
 <P/>
 The following example shows a call to <Ref Attr="Size"/> passing the options
 <C>hard</C> (with the value <K>true</K>)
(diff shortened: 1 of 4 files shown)


wilfwilson (commented):
The kernel changes seem to be doing what you want, from what I can tell.

I'm a bit worried that the behaviour one way or the other is not documented. As far as I can tell, the closest documentation that address this behaviour is `4.11-2 Function Call With Options` <https://www.gap-system.org/Manuals/doc/ref/chap4.html#X867D54987EF86D1D> which includes:
> The following example shows a call to Size (30.4-6) passing the options hard (with the value true) and tcselection (with the string "external" as value).
> ```
> gap> Size( fpgrp : hard, tcselection := "external" );
> ```
> Options supplied with function calls in this way are passed down using the global options stack described in chapter 8, so that the call above is exactly equivalent to
> ```
> gap> PushOptions( rec( hard := true, tcselection := "external") );
> gap> Size( fpgrp );
> gap> PopOptions( );
> ```

This example doesn't address the topic at issue here, but a sentence could be added, and this example modified (perhaps by constructing `fpgrp` in the one-liner too), to show the order in which arguments are evaluated versus options set.

fingolfin:
That chapter of the reference manual also says this about evaluation order:
> However, GAP does not make any guarantee about the order in which the arguments are evaluated. They might be evaluated left to right, right to left, or in any other order, but each argument is evaluated once

Based on this, one may argue that the "bug" this issue "fixes" isn't a bug at all -- it's a legitimate variation in otherwise undefined behavior.

However, I am *not* a fan of this, or of that sentence. GAP performs things in a fixed order (and typically that order is left-to-right, depth first), and it can be beneficial to users to be able to rely on that. Yes, we might in theory want to implement some kind of optimizations where it might be useful to be able to reorder these evaluations.... but does it seem likely we'll ever do this? I personally don't think so (but perhaps @ChrisJefferson disagrees?)

ChrisJefferson:
I don't think any optimisations would be worth the subtle bugs which would be introduced -- I think that both for GAP, and other programming languages, but GAP is the one where we can fix the order ourselves :)

fingolfin:
Good! I'll augment the documentation

fingolfin:
I've added a commit updating the documentation, please check it.

Editing this code made me realize that our documentation for variable length argument lists could be better, so I'll work on that next.

fingolfin:
Note that this conflicts with PR #4547 -- it'd be much easier for me if that PR was merged first, then I could easily update the PR here to match

wilfwilson (approved):
Thank you for the documentation changes, and I'm happy to also approve the code codes.

wilfwilson:
#4547 is merged so it's time to rebase this when you have a chance.

PR #4638, opened 2021-08-25
Relax correctness check of Test pragmas

### LINKED ISSUE #4635: Test pragmas are very sensitive
With current master I'm using the `#@if`, `#@else`, and `#@fi` test pragmas that I think @ChrisJefferson introduced. IMO I have found weird, unintuitive behaviour. If I use the following test file, I get `Error, Invalid test file: #@ command found in the middle of a single test`:

```
# Test case 1
#@if true
gap> d := 1;
#@else
gap> d := 2;
#@fi

gap> d := 3;
```

I think that shouldn't happen.

However, if I delete the empty line before the last line or if I put a line containing `# Test case 2` in front of the last line, then everything runs fine.

Another example, is that swapping the first two lines of the above file also results in an `invalid test file` error.

I'm aware that that's probably the documented behaviour. I'd say let's change the behaviour or provide more meaningful error messages.

fingolfin:
It turned simple whitespace issues (such as placing an empty line into
the wrong spot) into hard error saying:

> Invalid test file: #@ command found in the middle of a single test

Resolves #4635

To be clear, @ssiccha's example test files will still result in test failures, and those extra empty lines will be removed by `rewriteToFile:=true`. But that's as intended.

### DIFF
--- a/lib/test.gi
+++ b/lib/test.gi
@@ -128,7 +128,7 @@ InstallGlobalFunction(ParseTestInput, function(str, ignorecomments, fnam)
     elif StartsWith(lines[i], ">\t") then
         testError("Invalid test file: Continuation prompt '> ' followed by a tab, expected a regular space");
     elif Length(outp) > 0 then
-      if foundcmd then
+      if foundcmd and not ForAll(lines[i], c -> c = ' ' or c = '\t') then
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):
I'm fine with this.

And yes @ssiccha please do open more issues if you think there are further things that could/should be improved 🙂

Issue #4640, opened 2021-08-25
CI: mismatch between gcov and compiler version for Windows builds?


fingolfin:
In the "testinstall -- windows-latest" CI jobs, code coverage reporting prints many warnings looking like this:
```
D:/a/gap/gap/build/obj/src/.libs/ariths.c.gcno:version 'B12*', prefer 'A81*'
D:/a/gap/gap/build/obj/src/.libs/ariths.c.gcda:version 'B12*', prefer version 'A81*'
D:/a/gap/gap/build/obj/src/.libs/ariths.c.gcno:'PostRestore' has arcs to entry block
D:/a/gap/gap/build/obj/src/.libs/ariths.c.gcno:'PostRestore' has arcs from exit block
```
According to https://stackoverflow.com/a/41247820 this suggests a mismatch between the GCC version which produced those files, and the version of `gcov` used to produce them. We should audit this.

wilfwilson:
Thanks for raising this issue. These are the thoughts off the top of my head:
* The Action `gap-actions/setup-cygwin` could have a problem with installing inconsistent versions, meaning it needs to be updated
* The Action `gap-actions/setup-cygwin` might need to be called with some custom arguments
* With the Windows CI setup, there is always a tension between the Runner-provided chopped-down version of Cygwin, and the more full-featured version of Cygwin that we install ourselves. We often need to be careful about running the correct version of bash as our shell; sometimes we need the built-in one, sometimes we need our special one. They're basically incompatible.
    * It could be that the Action `codecov/codecov-action` changed in such a way that it no longer plays nicely with respect to this distinction. E.g. perhaps it is using 'our' gcc but the Runner-provided gcov - something weird like that.
    * This could be resolved by using custom code for uploading the data, or rolling back to a previous version of this Action, or trying to fix `codecov/codecov-action` directly with a PR.

ChrisJefferson:
fixed at some point -- not sure exactly what fixed it, but looks good from the output to me

PR #4642, opened 2021-08-27
CI: disable coverage gathering for windows

fingolfin:
This is a hack to get our Windows CI tests unstuck: it currently tries for over 5 hours to "upload coverage data" and then times out. This in turn *could* be caused by a version of `gcov` not matching the version of `gcc` being installed -- see issue #4640.

So perhaps this is also related to the `setup-cygwin` github action?

But I don't have time to debug this further, so this PR just tries to fix the symptom, and perhaps others can find a cure for the cause?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -236,6 +236,7 @@ jobs:
         run: ${{ matrix.extra }} dev/ci-gather-coverage.sh
       - name: "Upload coverage data to Codecov"
         uses: codecov/codecov-action@v1
+        if: ${{ runner.os != 'Windows' }} # FIXME/HACK: skip this on Windows, it runs forever
         with:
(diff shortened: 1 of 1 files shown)


wilfwilson:
Thanks for the debugging that you did @fingolfin, and this fix.

To be honest, before looking at the file, I tried to remember whether I had tried to include code coverage in the Windows builds. It looks like I was aware that I was enabling code coverage for Windows builds in #4473, but this was because it happened to 'just work' if I used the code for the existing Linux/Mac builds, not that I put any effort into making it work.

Unfortunately I'm unlikely to have the desire to spend time looking into this, either. And I'm happy to be without Windows-specific coverage data.

But I think your detective work in #4640 may well be pointing in the right direction.

PR #4644, opened 2021-08-28
Correct libgap-api.h for C++ usage

MathieuDutSik:
In order to link C libraries to C++ programs, one needs to avoid the mangling.
This is done by adding the *extern "C" {" statements as done in this PR.
This allows to use the libgap in C++ programs with success.


dimpase (approved):
looks good to me. 

One would also need to create tests for C++ code calling libgap, but it can be done in another PR.

PR #4646, opened 2021-08-28
CI: update to codecov/codecov-action@v2

### LINKED ISSUE #4645: CI: the coverage uploader codecov/codecov-action@v1 that we use will stop working soon
We use version 1 of the Codecov uploader. But as stated in its README https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1:
> On February 1, 2022, this version will be fully sunset and no longer function
>
> Due to the [deprecation](https://about.codecov.io/blog/introducing-codecovs-new-uploader/) of the underlying bash [uploader](https://github.com/codecov/uploader), the Codecov GitHub Action has released v2 which will use the new uploader. You can learn more about our deprecation plan and the new uploader on our [blog](https://about.codecov.io/blog/introducing-codecovs-new-uploader/).

Note that:
> The v2 uploader has a few breaking changes for users

I noticed this by looking a little bit into #4640. It's _possible_ that migrating to v2 will at least get rid of the warnings reported in #4640, and get code coverage working again on the Windows CI, but it might not, and even if it does, it might not address the root cause.

fingolfin:
Resolves #4645 (maybe -- we'll see if it works or not)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -250,12 +250,10 @@ jobs:
       - name: "Gather coverage data"
         run: ${{ matrix.extra }} dev/ci-gather-coverage.sh
       - name: "Upload coverage data to Codecov"
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v2
         if: ${{ runner.os != 'Windows' }} # FIXME/HACK: skip this on Windows, it runs forever
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: "!./pkg/**,!./extern/**"
-          gcov_path_exclude: "./pkg/**"
 
   # The following job is duplicated in release.yml - keep the two in sync.
   # (except for their different 'needs' components).
--- a/dev/ci-gather-coverage.sh
+++ b/dev/ci-gather-coverage.sh
@@ -9,12 +9,6 @@
 
 set -ex
 
-# If we don't care about code coverage, do nothing
-if [[ -n ${NO_COVERAGE} ]]
-then
-    exit 0
-fi
-
 SRCDIR=${SRCDIR:-$PWD}
 
 # Make sure any Error() immediately exits GAP with exit code 1.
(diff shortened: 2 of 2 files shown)


fingolfin:
"Error: No coverage files found, exiting", hrm

fingolfin:
This is working now! Well, there are some small difference in coverage for the C/C++ kernel code that I do not quite understand (for some reason, it now marks ~2000 lines as "partially" covered, something it didn't do before), but it's IMHO "good enough".

To compare the results:
- before: https://codecov.io/gh/gap-system/gap/tree/8a14583e65c8c4a895312da2a127163d43730005/src or concretely https://codecov.io/gh/gap-system/gap/src/8a14583e65c8c4a895312da2a127163d43730005/src/trycatch.c
- after: https://codecov.io/gh/gap-system/gap/tree/e7af89606a82587582d5764dd163bde5a4479a60/src or concretely https://codecov.io/gh/gap-system/gap/src/e7af89606a82587582d5764dd163bde5a4479a60/src/trycatch.c

wilfwilson (approved):
Thank you very much for persevering with this, @fingolfin.

Given that the `@v1` action ceases to function on 1st February 2022, this is very good timing.

Hopefully the existence of this PR will help to make it easier for GAP packages to be migrated from `@v1` to `@v2`.

The difference in results that you mention is interesting. In particular, the (increased?) appearance of 'partial lines' is interesting.

Some _very_ quick Googling suggests that an `if` statement will be marked as 'partial' if it is always satisfied in testing, or never satisfied. That would explain at least some of the 'partial' lines in `trycatch.c`, which you link to. This doesn't explain why these lines were not caught previously.

Previously: the uncovered `return` shows that the `if` was never satisfied, but the `if` is nevertheless marked as covered.
old
Now: the `if` is still never satisfied, but is now marked as partial:
new

__**EDIT**: I had put the images int he wrong order.__

(But in any case, I think this PR should be merged as-is!)

fingolfin:
@wilfwilson did you by chance switch the two images? Other than that, the explanation makes perfect sense; so I'll count this as a win. I still don't understand why it changed, but perhaps I've overlooked some option codecov.sh used to pass to `gcov`, or this is a different `gcov` version, or ... Whatever :-).

As to adjusting GAP packages: I can look into writing a script which will fire up a mass of PRs to update this in "all" (?) packages, at least all in the `gap-packages` org. But we may also need to adjust https://github.com/gap-actions/process-coverage/blob/main/action.yml#L36 -- I am actually surprised that that line works, given that it is using the `-o` option incorrectly (entirely my fault). Perhaps the same code as in this PR should be used, i.e.:
```
find build -type f -name '*.gcno' -exec gcov -pb {} +
```

wilfwilson:
Yes I had put my screenshots in the wrong order, whoops!

PR #4648, opened 2021-09-01
Add special cases to `Transitivity` for a permgroup with known size: natural An and Sn

wilfwilson:
Apart from trivial cases, `Transitivity(SymmetricGroup(n)) = n` and `Transitivity(AlternatingGroup(n)) = n - 2`. Easy! This currently takes longer than it should, in my opinion, because there is no special case for such groups. The existing method works quite well for natural symmetric groups (presumably because there are special methods for point stabilisers of natural symmetric groups), but it's still not amazing:
```
gap> Transitivity(SymmetricGroup(1000));; time;
214
```
But it's much worse for alternating groups (degree of just 100):
```
gap> Transitivity(AlternatingGroup(100));; time;
12084
```
So I've added a special case to this method, and now:
```
gap> Transitivity(AlternatingGroup(1000));; time;
2
```
(This breaks down once the degree gets big enough that the order of the alternating/symmetric group is not set at creation.)

I think there's possibly a case to be made for setting the value of this attribute at creation in the `SymmetricGroup`/`AlternatingGroup` methods, but this works well enough already and this should catch groups that happen to be naturally symmetric/alternating without being created by those functions.

Of course this PR isn't meant to be exhaustive. There are many places in the permgroups code where we could add special cases for natural symmetric and alternating groups: they are cheap to detect as such, but it is expensive to build stabiliser chains for such groups (which default methods often do), so we should avoid doing that when reasonably practical.

### DIFF
--- a/lib/oprtperm.gi
+++ b/lib/oprtperm.gi
@@ -1094,6 +1094,10 @@ InstallOtherMethod( Transitivity,
       # The trivial group is transitive on the empty set,
       # but has transitivity zero.
       return 0;
+    elif IsNaturalSymmetricGroup( G ) then
+      return n;
+    elif IsNaturalAlternatingGroup( G ) then
+      return n - 2;
     fi;
     t:= 0;
     size:= Size( G );
--- a/tst/testinstall/oprtperm.tst
+++ b/tst/testinstall/oprtperm.tst
@@ -1,0 +1,26 @@
+#@local
+gap> START_TEST("oprtperm.tst");
+
+# Transitivity of permgroups of known size that happen to be natural Sn/An
+# (Really we're testing performance)
+gap> Transitivity(SymmetricGroup(1000));
+1000
+gap> Transitivity(AlternatingGroup(1002));
+1000
+
+# Check small cases with moved points [1..n]
+gap> List([0 .. 10], n -> Transitivity(SymmetricGroup(n)));
+[ 0, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+gap> List([0 .. 10], n -> Transitivity(AlternatingGroup(n)));
+[ 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8 ]
+
+# Check small cases with moved points different from [1..n]
+gap> List([0 .. 10], n ->
+>      Transitivity(SymmetricGroup(Shuffle(ShallowCopy(Primes)){[1 .. n]})));
+[ 0, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+gap> List([0 .. 10], n ->
+>      Transitivity(AlternatingGroup(Shuffle(ShallowCopy(Primes)){[1 .. n]})));
+[ 0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8 ]
+
+#
+gap> STOP_TEST("oprtperm.tst", 1);


ChrisJefferson:
in this kind of case it's worth thinking if you should use `HasIsSymmetricGroup` first, to avoid paying the cost, but having done some benchmarks these functions seem very cheap, and we start doing `Size` and `Stabilizer` straight afterwards, which is much more expensive anyway (if not known).

wilfwilson:
(I think you mean `IsNaturalSymmetricGroup` etc rather than `IsSymmetricGroup` - see below for an example of why).

But just to clarify, were you suggesting that I should think about doing `if HasIsNaturalSymmetricGroup(G) and IsNaturalSymmetricGroup(G) then` rather than `if IsNaturalSymmetricGroup(G) then`? I agree that that's a good thing to think about in general, but yeah `IsNaturalSymmetricGroup` and `IsNaturalAlternatingGroup` just seem so cheap that it's worth calling it early.

Note that the `Transitivity` method that I'm altering is specifically for `IsPermGroup and HasSize`, so there is no penalty to using the size, but yes we still want to avoid constructing stabiliser chains for alternating groups etc.

The promised example:
```
gap> G := Group([ (1,3,5)(2,4,6), (1,4)(2,3)(5,6) ]);;
gap> StructureDescription(G);
"S3"
gap> IsSymmetricGroup(G);
true
gap> IsNaturalSymmetricGroup(G);
false
gap> Transitivity(G);
1
```

wilfwilson:
Can anyone help me with the manual example? It's the final step of the manual example of `FpGroupCocycle` that's failing, and I don't understand whether or not the diff is 'harmless' (I assume it's **not** harmless, since there's a comment about normal forms, and normal forms don't change!).

Specifically, in GAP master, the following code gives the final result `F1*m4`:
```
gap> g:=Group((2,15,8,16)(3,17,14,21)(4,23,20,6)(5,9,22,11)(7,13,19,25),
> (2,12,7,17)(3,18,13,23)(4,24,19,9)(5,10,25,15)(6,11,16,21));;
gap> StructureDescription(g);;
gap> mats:=[[[1,1,0,2],[2,0,0,0],[0,2,2,0],[0,1,0,0]],
> [[0,0,0,1],[1,1,2,0],[1,0,2,1],[1,0,1,0]]]*Z(3)^0;;
gap> mo:=GModuleByMats(mats,GF(3));;
gap> coh:=TwoCohomologyGeneric(g,mo);;
gap> coh.cohomology;
gap> p:=FpGroupCocycle(coh,coh.zero,true);
gap> g1:=Image(IsomorphismPermGroup(p));
gap> p:=FpGroupCocycle(coh,coh.cohomology[1],true);
gap> g2:=Image(IsomorphismPermGroup(p));
gap> Collected(List(MaximalSubgroupClassReps(g1),Size));
gap> Collected(List(MaximalSubgroupClassReps(g2),Size));
gap> p:=FpGroupCocycle(coh,coh.cohomology[1],true:normalform);;
gap> p.7*p.1;
```

Actually, re-running this block of code in the same GAP session gives different answers! So there's some randomness involved... is that to be expected? The second answer that I get is `F1*m2*m3^-1*m4`.

I'd appreciate any insight please.

fingolfin:
This is @hulpke's code. My understanding is that the normal form is "fixed" for a particular run of GAP, but it might change depending on e.g. randomization (if different stabilizer chains are used) and other factors. I think I've seen some of his PRs make similar adjustment to the tests... but I might be completely wrong, so best to hear what @hulpke has to say :-)

hulpke:
> Can anyone help me with the manual example? It's the final step of the manual example of `FpGroupCocycle` that's failing, and I don't understand whether or not the diff is 'harmless' (I assume it's **not** harmless, since there's a comment about normal forms, and normal forms don't change!).
> 
> Specifically, in GAP master, the following code gives the final result `F1*m4`:
> 
> ```
> gap> g:=Group((2,15,8,16)(3,17,14,21)(4,23,20,6)(5,9,22,11)(7,13,19,25),
> > (2,12,7,17)(3,18,13,23)(4,24,19,9)(5,10,25,15)(6,11,16,21));;
> gap> StructureDescription(g);;
> gap> mats:=[[[1,1,0,2],[2,0,0,0],[0,2,2,0],[0,1,0,0]],
> > [[0,0,0,1],[1,1,2,0],[1,0,2,1],[1,0,1,0]]]*Z(3)^0;;
> gap> mo:=GModuleByMats(mats,GF(3));;
> gap> coh:=TwoCohomologyGeneric(g,mo);;
> gap> coh.cohomology;
> gap> p:=FpGroupCocycle(coh,coh.zero,true);
> gap> g1:=Image(IsomorphismPermGroup(p));
> gap> p:=FpGroupCocycle(coh,coh.cohomology[1],true);
> gap> g2:=Image(IsomorphismPermGroup(p));
> gap> Collected(List(MaximalSubgroupClassReps(g1),Size));
> gap> Collected(List(MaximalSubgroupClassReps(g2),Size));
> gap> p:=FpGroupCocycle(coh,coh.cohomology[1],true:normalform);;
> gap> p.7*p.1;
> ```
> 
> Actually, re-running this block of code in the same GAP session gives different answers! So there's some randomness involved... is that to be expected? The second answer that I get is `F1*m2*m3^-1*m4`.
> 
> I'd appreciate any insight please.

TLDR: It is harmless (and it is one of the cases where the verbatim checking of manual examples is less than helpful.)

What happens is that a confluent rewriting system is constructed for the group `g` (which is $GL(2,5)$). This uses an isomorphism to the composition factor A_5 which uses random elements.
This rewriting system then influences the normal form. for elements of the extension -- thus the different result. depending on random choices.

Why is it called `normalform` then (BTW this is not yet in any released version -- feel free to rename): If set, arithmetic in the constructed extension (which is an FpGroup) will use the implicit rewriting system, to bring elements into normal form (the normal form however run-dependent). You can see the difference if you call

`p.1^Order(p.1);`

with or without constructing `p` with the `normalform` option.

wilfwilson:
Thanks for the explanation @hulpke. I will update the manual example.

hulpke (approved):
The test for being a natural Sn/A_n should be considered (otherwise its a bug) as much faster than calculating the order, and as basically no cost if the order is known. So this makes sense as written.

hulpke:
> Thanks for the explanation @hulpke. I will update the manual example.



> Thanks for the explanation @hulpke. I will update the manual example.

This example seems to be highly instable, I've thus put it into a non-checked example with my last PR. So once you rebase, the problem should be gone.

wilfwilson:
Thank you for doing that @hulpke, I've rebased now.

PR #4650, opened 2021-09-06
Clean up of partition backtrack code.

MathieuDutSik:
The code of the partiion backtrack has several old code that deserved to be cleanup:
* The code of lftObj and rgtObj is nor used at all within the code.
* One entry in AddRefinement did not use the global names attributed to it.
* The symmetric group code still remained outcommented.
So, this should help make things clearer for the many readers of this code.

Note that it would be helpful to complement mentions such as "See sec. 3.7  of my thesis"
by the actual title of the thesis and the author. But that is unfortunately out of my
knowledge.

### DIFF
--- a/lib/stbcbckt.gi
+++ b/lib/stbcbckt.gi
@@ -44,7 +44,7 @@ BindGlobal( "Refinements", AtomicRecord() );
 #F  IsSlicedPerm( <perm> )  . . . . . . . . . . . . . . . sliced permutations
 ##
 DeclareRepresentation( "IsSlicedPerm", IsPerm,
-                        [ "length", "word", "lftObj","opr" ] );
+                        [ "length", "word", "opr" ] );
 
 #############################################################################
 ##
@@ -52,7 +52,7 @@ DeclareRepresentation( "IsSlicedPerm", IsPerm,
 ##
 InstallGlobalFunction( UnslicedPerm@, function( perm )
     local   prm,  i;
-    
+
     if IsSlicedPerm( perm )  then
         prm := ();
         for i  in [ 1 .. perm!.length ]  do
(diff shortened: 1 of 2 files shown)


ssiccha:
Hi @MathieuDutSik, many thanks for the PR! :) I don't know much about the partition backtrack code, but maybe @ChrisJefferson, @AnnaKDS, or @wucas can have a look.

lib/stbcbckt.gi
```diff
@@ -1340,10 +1333,6 @@ InstallGlobalFunction( PartitionBacktrack,
                     image.level := rbase.lev[ d ];
                     if IsSlicedPerm( image.perm )  then
                         image.perm!.length := oldprm;
-                         # Here and below the code that refers to `rgtObj` was used to avoid multiplication
```
hulpke, line 1343: I think this is actually a useful comment (in particular if ever there will be a discrepancy to an older version.) I would leave this in.

hulpke (commented):
Besides many whitespace changes (that make it hard to find what actually changed) this is mainly deleting old code that has been disabled by commenting it out. 
I find this commented-out code helpful, as it indicates improvements that were attempted and ultimately deemed to be not worth it, or is debugging code.

lib/stbcbckt.gi
wucas: This comment was added by myself. @AnnaKDS and I had a look at the backtrack code in 2018 for a seminar and @ChrisJefferson and @markuspf helped us to understand it. There we noticed that this was still documented but not used in the code. So I removed it from the documentation but added this comment so that its easier to understand why the code is the way it is. The relevant PR is #3101

wucas (changes requested):

ssiccha:
@hulpke If you are viewing the PR on github, then under the `Files changed`-tab -> cogwheel -> _hide whitespace changes_ you can hide the whitespace changes.

fingolfin:
Thank you to @MathieuDutSik for making the effort to produce this PR. However, I don't think this is something we should merge, for the reasons already outlined by @hulpke , with which I mostly agree -- conversely, there seems little (if any?) value added by these changes. Perhaps I am missing something here, though, so feel free to elaborate as to what exactly the motivation of this change are, i.e., in how far does it help you in whatever you are actually trying to do?

MathieuDutSik:
The PR was not judged to be of sufficient value

PR #4653, opened 2021-09-12
added `IsQuasisimpleGroup`, `IsQuasisimpleCharacterTable`

ThomasBreuer:
and fixed a bug due to a wrong implication (which apparently nobody has noticed up to now)

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -3306,8 +3306,8 @@ DeclareOperation("CentralizerModulo", [IsGroup,IsGroup,IsObject]);
 ##  gap> PCentralSeries(g,2);
 ##  [ <pc group of size 12 with 3 generators>, Group([ y3, y*y3 ]), Group([ y*y3 ]) ]
 ##  gap> g:=SymmetricGroup(4);;
-##  gap> PCentralSeries(g,2);
-##  [ Sym( [ 1 .. 4 ] ), Group([ (1,2,3), (2,3,4) ]) ]
+##  gap> List(PCentralSeries(g,2), StructureDescription);
+##  [ "S4", "A4" ]
 ##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -1938,9 +1938,9 @@ DeclareAttribute( "MinimalNormalSubgroups", IsGroup );
 ##  <Description>
 ##  returns a list of all normal subgroups of <A>G</A>.
 ##  <Example><![CDATA[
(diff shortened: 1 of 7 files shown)


lib/grp.gd
```diff
@@ -729,6 +729,43 @@ InstallTrueMethod( IsGroup and IsNonTrivial, IsAlmostSimpleGroup );
...
+##
+DeclareProperty( "IsQuasisimpleGroup", IsGroup );
+InstallTrueMethod( IsGroup and IsNonTrivial, IsQuasisimpleGroup );
+
+# Nonabelian simple groups are quasisimple, quasisimple groups are perfect.
+InstallTrueMethod( IsQuasisimpleGroup, IsNonabelianSimpleGroup );
+InstallTrueMethod( IsPerfectGroup, IsQuasisimpleGroup );
+
+# We can expect that people will try the name with capital s.
+DeclareSynonymAttr( "IsQuasiSimpleGroup", IsQuasisimpleGroup );
```
fingolfin, line 767: Hmmm, indeed...  In particular since it is `IsAlmostSimpleGroup` and `IsNonabelianSimpleGroup`. What's the logic here? Is it because "quasisimple" is one word but "almost simple" is two?

So I guess for `IsNonabelianSimpleGroup ` someone had `nonabelian` in mind, while for `NonAbelianTensorSquare` (in packages) someone else uses `non-abelian`... *sigh*

We don't seem to provide such "up to case synonyms" for other things, though. Should we really start this here? If we do, I'd add at least a comment explaining why we do this here but not in other cases, e.g. the above reasoning (assuming it is what you had in mind).

fingolfin (commented):
There is a still a diff in an example (very annoying, maybe we should turn off the output of that example):
```
gap> ########> Diff in /home/runner/work/gap/gap/tst/testmanuals/ref-chapter03\
9.tst/home/runner/work/gap/gap/doc/ref/../../lib/grp.gd:3304:615
# Input is:
PCentralSeries(g,2);
# Expected output:
[ Sym( [ 1 .. 4 ] ), Group([ (1,2,3), (2,3,4) ]) ]
# But found:
[ Sym( [ 1 .. 4 ] ), Group([ (1,2,3), (1,3,4) ]) ]
########
```

lib/grp.gd
ThomasBreuer: I thought that "camel case" means that the first letter of each word is capitalized. In the current situation, both quasisimple (one word) and quasi-simple (two words?) are common. The easiest way to deal with this situation is to support both names. In my eyes, this is analogous to `Centre` vs. `Center` and `Centralizer` vs. `Centraliser`.

fingolfin (approved):

PR #4654, opened 2021-09-12
adjusted a statement about `InfoObsolete`

ThomasBreuer:
Variables created by `DeclareObsoleteSynonym` and `DeclareObsoleteSynonymAttr`
print a message if the `InfoObsolete` level is high enough.
The current default threshold is 2 for `DeclareObsoleteSynonym`
and 1 for `DeclareObsoleteSynonymAttr`,
but the documentation claimed 1 for both.

Perhaps it would be better to change the code to use default 1 (or 2?)
in both cases.

### DIFF
--- a/lib/obsolete.gd
+++ b/lib/obsolete.gd
@@ -57,15 +57,17 @@
 ##  they may be removed when they don't seem to be used any more.
 ##  <P/>
 ##  Information about obsolete names is printed by <Ref Func="Info"/> using the
-##  InfoObsolete Info class. By default InfoObsolete is set to 1. Newly
+##  <Ref InfoClass="InfoObsolete"/> Info class.
+##  By default <Ref InfoClass="InfoObsolete"/> is set to 1. Newly
 ##  obsoleted identifiers should at first be outputted at info level 2. Once they
 ##  have been removed from all packages, they should then be moved to info level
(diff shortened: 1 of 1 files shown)


hulpke (approved):

fingolfin:
I believe the use of two levels was introduced by @ChrisJefferson in PR #2923. So perhaps he'd like to chime in here. Also @alex-konovalov might.

Generally said, I am not sure whether we *really* want to remove obsolete stuff, ever, unless we have strong reasons for doing so (e.g.: a function requires a certain internal way to represent things in order to work; we want to change those internals, but don't want to put efforts into rewriting an obsolete function for use with the new setup). After all, if there is no immediate gain, why inconvenience our users? We also may want to warn about some obsoletes (without removing them) if their user suggests potentially inefficient or unsafe code, though.

ThomasBreuer:
O.k., this topic is not relevant enough for more comments. The solution is perhaps not optimal, but it makes the world a little bit better.

ChrisJefferson:
Just to say, I am happy this was merged, and believe it fixed an earlier mistake of mine. I missed the PR in the pile of emails I received while on holiday. Sorry.

alex-konovalov:
> Perhaps it would be better to change the code to use default 1 (or 2?) in both cases.

I think so (if yes, then to 1).

PR #4655, opened 2021-09-12
renamed `RadicalGroup` to `SolvableRadical`

ThomasBreuer:
and added the treatment of `RadicalGroup` as an obsolescent variable

Note that there is the notion of radical (p-)subgroups of a group,
which means something completely different,
and that solvable radical seems to be the term used in the litarature.

(I guess that the name `RadicalGroup` had been chosen --a long time ago--
because `Radical` was too unspecific, and thus adding the suffix `Group`
provided at least some context.)

## Text for release notes

## Further details

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -2039,25 +2039,26 @@ DeclareAttribute( "ComputedAgemos", IsGroup, "mutable" );
 
 #############################################################################
 ##
-#A  RadicalGroup( <G> )
+#A  SolvableRadical( <G> )
 ##
-##  <#GAPDoc Label="RadicalGroup">
+##  <#GAPDoc Label="SolvableRadical">
 ##  <ManSection>
-##  <Attr Name="RadicalGroup" Arg='G'/>
+##  <Attr Name="SolvableRadical" Arg='G'/>
 ##
 ##  <Description>
-##  is the radical of <A>G</A>, i.e., the largest solvable normal subgroup of <A>G</A>.
+##  is the solvable radical of the group <A>G</A>,
+##  i.e., the largest solvable normal subgroup of <A>G</A>.
 ##  <Example><![CDATA[
-##  gap> RadicalGroup(SL(2,5));
+##  gap> rad:= SolvableRadical( SL(2,5) );
 ##  <group of 2x2 matrices of size 2 over GF(5)>
-##  gap> Size(last);
+##  gap> Size( rad );
 ##  2
 ##  ]]></Example>
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-DeclareAttribute( "RadicalGroup", IsGroup );
+DeclareAttribute( "SolvableRadical", IsGroup );
 
 
 #############################################################################
--- a/lib/obsolete.gd
+++ b/lib/obsolete.gd
@@ -756,3 +756,26 @@ DeclareObsoleteSynonym( "IsLexicographicallyLess", "<" );
 ##
 ##  Still used in fr (06/2021)
 DeclareSynonym( "FirstOp", First );
+
+
+#############################################################################
+##
+#A  RadicalGroup( <G> )
+##
+##  <ManSection>
+##  <Attr Name="RadicalGroup" Arg='G'/>
+##
+##  <Description>
+##  is the radical of <A>G</A>, i.e., the largest solvable normal subgroup of <A>G</A>.
+##  <Example><![CDATA[
+##  gap> RadicalGroup(SL(2,5));
+##  <group of 2x2 matrices of size 2 over GF(5)>
+##  gap> Size(last);
+##  2
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##
+##  'RadicalGroup' was documented until version 4.11.1.
+##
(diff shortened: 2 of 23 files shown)


hulpke (changes requested):
Is there a need to tag the synonym `RadicalGroup` immediately as obsolete? I have packages and private code that install methods, and there are publications out that explicitly refer to `RadicalGroup`. We can slowly change the princip[al use over, but I don't see a benefit in marking `RadicalGroup` for deletion.

ThomasBreuer:
@hulpke Would it be sufficient to set the minimal level of `InfoObsolete` that triggers the message about the recommended function name to 2? (There seems to be confusion about the default level, see pull request #4654)

Actually, I see no reason to remove obsolete synonyms for names that are regarded as better, except if they are wrong. The situation is different for obsolete code, where one can argue that it would be a burden to maintain it further.

hulpke (approved):

hulpke:
As long as `RadicalGroup` continues to work and does not print warning or error messages I have no objection.

lib/grp.gd
```diff
@@ -2039,25 +2039,26 @@ DeclareAttribute( "ComputedAgemos", IsGroup, "mutable" );
...
 ##
-##  <#GAPDoc Label="RadicalGroup">
+##  <#GAPDoc Label="SolvableRadical">
 ##  <ManSection>
-##  <Attr Name="RadicalGroup" Arg='G'/>
+##  <Attr Name="SolvableRadical" Arg='G'/>
 ##
 ##  <Description>
-##  is the radical of <A>G</A>, i.e., the largest solvable normal subgroup of <A>G</A>.
+##  is the solvable radical of the group <A>G</A>,
```
wilfwilson, line 2049: Do you think it would be a good idea to add an index entry `RadicalGroup` that takes the user directly to the `SolvableRadical` manual entry? I'm not sure, but thought I thought I'd suggest it.
```suggestion
##  <Index Key="RadicalGroup"><C>RadicalGroup</C></Index>
##  is the solvable radical of the group <A>G</A>,
```

wilfwilson (changes requested):
When compiling the reference manual with this PR, there is a warning about an unresolved reference to `RadicalGroup`, which therefore needs to be updated to `SolvableRadical`. I'm not sure where the broken reference is.
```
#W WARNING: non resolved reference: rec(
  Attr := "RadicalGroup" )
```

lib/obsolete.gd
```diff
@@ -756,3 +756,26 @@ DeclareObsoleteSynonym( "IsLexicographicallyLess", "<" );
...
+##  <Example><![CDATA[
+##  gap> RadicalGroup(SL(2,5));
+##  <group of 2x2 matrices of size 2 over GF(5)>
+##  gap> Size(last);
+##  2
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##
+##  'RadicalGroup' was documented until version 4.11.1.
```
fingolfin, line 779: It should still be documented! Normally when we rename things and make the old name obsolete, then we add an entry for it (including an index entry) to `doc/ref/obsolete.xml`, and I think that should be done here.

So I'd rather say:
```suggestion
##  'RadicalGroup' was renamed in GAP 4.12
```

lib/obsolete.gd
```diff
@@ -756,3 +756,26 @@ DeclareObsoleteSynonym( "IsLexicographicallyLess", "<" );
...
+##  is the radical of <A>G</A>, i.e., the largest solvable normal subgroup of <A>G</A>.
+##  <Example><![CDATA[
+##  gap> RadicalGroup(SL(2,5));
+##  <group of 2x2 matrices of size 2 over GF(5)>
+##  gap> Size(last);
+##  2
+##  ]]></Example>
+##  </Description>
+##  </ManSection>
+##
```
fingolfin, line 778:
```suggestion
```
I don't see a need to keep this?

ThomasBreuer:
@wilfwilson Thanks for the hint. This is very strange, in fact many instances of `RadicalGroup` were not replaced by the pull request; I do not understand what happened, but I will try to update the pull request in order to fix this problem.

lib/obsolete.gd
ThomasBreuer: Yes.

lib/obsolete.gd
ThomasBreuer: I can make this change.
But my understanding of the word "documented" is a positive one in the sense that one can rely on documented variables, and using undocumented variables can be risky. This was also the idea behind the (undocumented) function `IsDocumentedWord`; having index entries for deprecated variable names causes that the current implementation of this function does not fit to its intended meaning.

Concerning the sentence "Normally ...":
Currently we do not claim to have a complete list of name changes in the chapter called "Replaced and Removed Command Names" in the Reference Manual. The section about group actions lists "some examples of such name changes", and the section "Miscellaneous Name Changes or Removed Names" states that it lists "some further name changes". I think it would be better to have a complete list, and to state this --a topic for a new issue.

lib/grp.gd
ThomasBreuer: See the remark about `doc/ref/obsolete.xml`: Now there is an index entry that points to a section that lists name changes.

wilfwilson (approved):

fingolfin (approved):

Issue #4659, opened 2021-09-17
Stabilizer and OrbitStabilizer sometimes produce incorrect results


daniel-dpk:
Apologies if I am using GAP incorrectly, but I could not figure out what exactly is not right with my input. So, here is my report.

When computing the stabilizer using either `Stabilizer()` or `OrbitStabilizer()`, the result is in some cases incorrect. In particular, it seems to depend on what other operations have been performed with the group before the call (e.g. `Size()`).

I have tested this with GAP 4.11.1 and the current master branch, both on Ubuntu 20.04 (gcc 9.3.0).

### Observed behaviour

The following reproduces the issue:

```
gap> p_gens := [PermutationMat((1,2), 6), PermutationMat((1,2,3,4,5,6), 6)];;
gap> p_group := Group(p_gens);;
gap> s1 := DiagonalMat([-1, -1, 1, 1, 1, 1]);;
gap> s_gens := Orbit(p_group, s1);;
gap> ps_gens := Concatenation(p_gens, s_gens);;
gap> ps_group := Group(ps_gens);;
gap> v := [1,2,1,1,0,0];;
gap> orbit := Orbit(ps_group, v);;
gap> Size(ps_group);;    # <-- IMPORTANT: Removing this line fixes it!
gap> stab := Stabilizer(ps_group, v);;
gap> Size(orbit) * Size(stab) = Size(ps_group);
false
```

If the `Size(ps_group)` call is removed, the last line results in `true`. However, if the `Stabilizer(...)` line is run again after this, it produces the wrong result again. My current workaround is to create a new group object for each stabilizer.

### Expected behaviour

The stabilizer calculation should not depend on other operations performed using the group.

### Copy and paste GAP banner (to tell us about your setup)

```
 ┌───────┐   GAP 4.11.1 of 2021-03-02
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv7
 Configuration:  gmp 6.2.0, GASMAN, readline
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2020.08.11, AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.23,
[...]

PR #4660, opened 2021-09-17
Fix (rare) wrong result for stabilizer computations (and a few other things)

hulpke:
A fix for #4659 and a number of small improvements (which are added so I don't need to wait for another approving review)

## Text for release notes

- A rare error in calculating stabilizers with the generic method was fixed.

### DIFF
--- a/lib/ctblgrp.gi
+++ b/lib/ctblgrp.gi
@@ -2056,11 +2056,16 @@ end );
 ##
 #F  DixonSplit(<D>) . .  calculate matrix,split spaces and obtain characters
 ##
-InstallGlobalFunction( DixonSplit, function(D)
-local r,i,j,ch,ra,bsm,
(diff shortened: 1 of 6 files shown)


lib/ctblgrp.gi
```diff
@@ -2056,11 +2056,16 @@ end );
 ##
 #F  DixonSplit(<D>) . .  calculate matrix,split spaces and obtain characters
 ##
-InstallGlobalFunction( DixonSplit, function(D)
-local r,i,j,ch,ra,bsm,
+InstallGlobalFunction( DixonSplit, function(arg)
```
fingolfin, line 2059: This is never called with the new argument, though, right? So is this for future work, or just for manual invocations by "super expert users"?

hulpke: The latter.

fingolfin (approved):

PR #4667, opened 2021-10-01
Enable tab completion for non-internal records

fingolfin:
If one implements a non-internal record type, then with this patch tab
completion for instances of this type becomes possible.

Note that this means we can't call `NamesOfComponents` to get the record
names, in case the record type is implemented as a component object:
`NamesOfComponents` will then return the internal names of that
component object, not the names of the record it implements.

For the same reason, we must split the code, which so far uniformly used
`r!.(name)` to access components of record and component objects.

There is other code which arguably should work with non-internal record
objects but currently calls `NamesOfComponents`; I've adjusted a few
more of these to call `RecNames` instead, but this is likely not
comprehensive.

Note that I deliberately decided against changing the behaviour of
`NamesOfComponents` as it is explicitly documented, has been so for some
time, and a bunch of packages use it.

This is for use in the Julia integration, but it is useful on its own.


fingolfin:
Upon re-reading the `NamesOfComponents` documentation, it is perhaps not as clear cut as I thought earlier: it just states what happens for component objects, then what happens for records. But it never discusses what happens with records that are component objects... So two possible ways:
1. Codify the current behaviour, and change the `NamesOfComponents` documentation to talk about "internal records"
2. Change the current behaviour (check for `IsRecord` before checking for `IsComponentObjectRep`) and then adjust the documentation accordingly (i.e., move the paragraph about records first, then talk about "component objects which are not record..." (and perhaps make it clear that a user could implement custom record types).

Before we do 2 (if at all), I would suggest to check at least how packages use `NamesOfComponents` to avoid unexpected surprises. But since non-internal record types are rare, I am hopeful.

ThomasBreuer:
I think that the current documentation of `NamesOfComponents`

> For a component object comobj, NamesOfComponents returns a list of strings,
> which are the names of components currently bound in comobj.
> 
> For  a  record  comobj,  NamesOfComponents  returns  the result of RecNames (29.1-2).

expresses that nobody thought about `IsRecord` objects implemented via `IsComponentObjectRep`.
The only consistent way to read this documentation is that for such objects, both `NamesOfComponents` and `RecNames` return the list of names of the components, which means that there is no function that computes the list of *record* components.
However, this contradicts the documentation of `RecNames`.

Thus my suggestion is to change the documentation of `NamesOfComponents` such that the second statement is claimed only for internal records (or for records that are not component objects), and to regard this change as a bugfix.

Perhaps also the documentation of records (currently chapter 29 of the GAP Reference Manual) should be changed such that it talks about *record components* not about *components*, and a cross-reference to "Component Objects" could emphasize that the notation is subtle.

fingolfin:
I adjusted the `NamesOfComponents` documentation.

frankluebeck:
I agree with the clarification in the documentation of `NamesOfComponents`.

So far, in the TAB completion of component/record names we do not distinguish between
```
myrec.[TAB]
myrec!.[TAB]
```
But in case `myrec` is in `IsRecord and IsComponentObject` these should be distinguished (using `RecNames` in the first case and `NamesOfComponents` in the second).

fingolfin:
@frankluebeck ... and that's what this PR does... so do you agree with the changes?

frankluebeck:
@fingolfin No, this is not what this PR does. Example:
```
gap> fam := NewFamily("MockFamily");;
gap> cat := NewCategory("IsMockRecord", IsRecord);;
gap> type2 := NewType(fam, cat and IsComponentObjectRep);;
gap> InstallMethod(RecNames, [cat], r-> List(NamesOfComponents(r), a-> 
>      Concatenation(a, "_asrecname")));
gap> InstallMethod(\., [cat,IsInt], function(r, rnam)
>   local nam;
>   nam := NameRNam(rnam);
>   return r!.(nam{[1..Length(nam)-10]}); end);
gap> 
gap> rr := Objectify(type2, rec(a := 1, b := 2));;
gap> NamesOfComponents(rr);
[ "b", "a" ]
gap> RecNames(rr);
[ "b_asrecname", "a_asrecname" ]
gap> rr!.a;
1
gap> rr.a_asrecname;
1
```
Now this is ok:
```
gap> rr.[TAB]
a_asrecname  b_asrecname  
```
But we also have:
```
gap> rr!.[TAB]
a_asrecname  b_asrecname  
```
and this should  yield 
```
gap> rr!.[TAB]
a b
```

PR #4668, opened 2021-10-03
Add SpinorNorm.

danielrademacher:
For the maximal subgroup project we need the SpinorNorm. In the package "matrix" of GAP3 I found the WallForm and SpinorNorm functions of this PR. I modified the SpinorNorm output slightly in order to have a more intuitive feedback. For this, the function "IsSquare" has been added.

## Text for release notes

see title

### DIFF
--- a/grp/classic.gi
+++ b/grp/classic.gi
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
 InstallMethod( Omega,
     [ IsFunction, IsInt, IsPosInt, IsField and IsFinite ],
     { filt, e, d, R } -> OmegaCons( filt, e, d, Size( R ) ) );
+    
+    
+#############################################################################
+##
+#F  WallForm( <form>, <m> ) . . . . . . . . . . . . . compute the wall of <m>
+##
+# Computes the Wall Form of a matrix. The defintition can be found in
+# [Taylor, page 163].
+BindGlobal( "WallForm", function( form, m )
+    local id,  w,  b,  p,  i,  x,  j;
+
+    # first argument should really be something useful
+    id := One( m );
+
+    # compute a base for Image(id-m), use the most stupid algorithm
+    w := id - m;
+    b := [];
+    p := [];
+    for i  in [ 1 .. Length(w) ]  do
+        if Length(b) = 0  then
+            if w[i] <> 0*w[i]  then
+                Add( b, w[i] );
+                Add( p, i );
+            fi;
+        elif RankMat(b) <> RankMat(Concatenation(b,[w[i]]))  then
+            Add( b, w[i] );
+            Add( p, i );
+        fi;
+    od;
+
+    # compute the form
+    x := List( b, x -> [] );
+    for i  in [ 1 .. Length(b) ]  do
+        for j  in [ 1 .. Length(b) ]  do
+            x[i][j] := id[p[i]] * form * b[j];
+        od;
+    od;
+
+    # and return
+    return rec( base := b, pos := p, form := x );
+
+end );
+
+
+#############################################################################
+##
+#F  IsSquare( fld, e) . . . . . . . . . . . Tests whether <e> is a square element
+##   in <fld>
+##
+# Input: Field fld, e element of fld
+# Output: true if e is a square element in fld. Otherwise false.
+BindGlobal( "IsSquare", function( fld, e )
+    local char, q;
+    
+    char := Characteristic(fld);
+    # If the characteristic of fld is equal to 2, we know that every element is a
+    # sqare. Hence, we can return true.
+    if char = 2 then
+        return true;
+    fi;
+    q := Size(fld);
+    
+    # If the characteristic of fld is not 2, we know that there are exactly
+    # (q+1)/2 elements which are a square (Huppert LA, Theorem 2.5.4). Now observe
+    # that for a square element e we have that e^((q-1)/2) = 1. And, thus, the
+    # polynomial X^((q-1)/2) - 1 has already (q-1)/2 different roots (every square
+    # except 0). Hence, for a non-square element e' we have that (e')^((q-1)/2) <> 1
+    # which proves the lines below.
+    if e^((q-1)/2) = One(fld) then
+        return true;
+    else
+        return false;
+    fi;
+    
+end );
+
+
+#############################################################################
+##
+#F  SpinorNorm( <form>, <m> ) . . . . . . . .  compute the spinor norm of <m>
+##
+# Output: 1 if the discriminant of the Wall form of <m> is (F^*)^2.
+#          Otherwise -1.
+# The defintition can be found in [Taylor, page 163].
+BindGlobal( "SpinorNorm", function( form, m )
+    local one;
+    one := One(m[1][1]);
+    if IsOne(m) then return one; fi;
+    
+    if IsSquare(FieldOfMatrixList([m]), DeterminantMat( WallForm(form,m).form )) then
+        return one;
+    else
+        return -1 * one;
+    fi;
+end );
+
 
 
 #############################################################################
--- a/tst/testinstall/grp/classic-G.tst
+++ b/tst/testinstall/grp/classic-G.tst
@@ -1,7 +1,7 @@
 #
 # Tests for the "general" group constructors: GL, GO, GU, GammaL
 #
-#@local G, H, d, q, S, grps
+#@local G, H, d, q, S, grps, gens, w, form, g
 
 gap> START_TEST("classic-G.tst");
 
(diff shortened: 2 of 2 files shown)


grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+#############################################################################
+##
+#F  WallForm( <form>, <m> ) . . . . . . . . . . . . . compute the wall of <m>
+##
+# Computes the Wall Form of a matrix. The defintition can be found in
+# [Taylor, page 163].
+BindGlobal( "WallForm", function( form, m )
+    local id,  w,  b,  p,  i,  x,  j;
+
+    # first argument should really be something useful
```
fingolfin, line 1929: I don't understand this comment

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+    id := One( m );
+
+    # compute a base for Image(id-m), use the most stupid algorithm
```
fingolfin, line 1932: But... why? Why not just write `b := BaseMat( id - m );` ? Probably because you want a special such base, namely a subset of the rows of `id - m` with a corresponding `p`... then state so?

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+#F  IsSquare( fld, e) . . . . . . . . . . . Tests whether <e> is a square element
+##   in <fld>
+##
+# Input: Field fld, e element of fld
+# Output: true if e is a square element in fld. Otherwise false.
+BindGlobal( "IsSquare", function( fld, e )
+    local char, q;
+    
+    char := Characteristic(fld);
+    # If the characteristic of fld is equal to 2, we know that every element is a
```
fingolfin, line 1973: Aaand if fld is finite resp perfect!

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+    fi;
+    q := Size(fld);
+    
+    # If the characteristic of fld is not 2, we know that there are exactly
+    # (q+1)/2 elements which are a square (Huppert LA, Theorem 2.5.4). Now observe
+    # that for a square element e we have that e^((q-1)/2) = 1. And, thus, the
+    # polynomial X^((q-1)/2) - 1 has already (q-1)/2 different roots (every square
+    # except 0). Hence, for a non-square element e' we have that (e')^((q-1)/2) <> 1
+    # which proves the lines below.
+    if e^((q-1)/2) = One(fld) then
```
fingolfin, line 1986:
```suggestion
    if IsOne(e^((q-1)/2)) then
```

fingolfin (changes requested):
Some quick comments from my phone

fingolfin:
(Also: thanks for submitting this, I appreciate it despite my nitpicks :-)

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
 InstallMethod( Omega,
     [ IsFunction, IsInt, IsPosInt, IsField and IsFinite ],
     { filt, e, d, R } -> OmegaCons( filt, e, d, Size( R ) ) );
+    
+
```
fingolfin, line 1919:
```suggestion


```

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+            Add( b, w[i] );
+            Add( p, i );
+        fi;
+    od;
+
+    # compute the form
+    x := List( b, x -> [] );
+    for i  in [ 1 .. Length(b) ]  do
+        for j  in [ 1 .. Length(b) ]  do
+            x[i][j] := id[p[i]] * form * b[j];
```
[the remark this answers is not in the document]
fingolfin: Actually:
```suggestion
            x[i,j] := form[p[i]] * b[j];
```

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+#############################################################################
+##
+#F  IsSquare( fld, e) . . . . . . . . . . . Tests whether <e> is a square element
+##   in <fld>
+##
+# Input: Field fld, e element of fld
+# Output: true if e is a square element in fld. Otherwise false.
+BindGlobal( "IsSquare", function( fld, e )
+    local char, q;
+
```
fingolfin, line 1971:
```suggestion

    if IsZero(e) then return true; fi;

```

grp/classic.gi
```diff
@@ -1915,6 +1915,102 @@ InstallMethod( Omega,
...
+    # (q+1)/2 elements which are a square (Huppert LA, Theorem 2.5.4). Now observe
+    # that for a square element e we have that e^((q-1)/2) = 1. And, thus, the
+    # polynomial X^((q-1)/2) - 1 has already (q-1)/2 different roots (every square
+    # except 0). Hence, for a non-square element e' we have that (e')^((q-1)/2) <> 1
+    # which proves the lines below.
+    if e^((q-1)/2) = One(fld) then
+        return true;
+    else
+        return false;
+    fi;
```
fingolfin, line 1990:
```suggestion
    return IsOne(e^((q-1)/2));
```

grp/classic.gi
```diff
@@ -1917,6 +1917,122 @@ InstallMethod( Omega,
     { filt, e, d, R } -> OmegaCons( filt, e, d, Size( R ) ) );
 
 
+#############################################################################
+##
+#F  WallForm( <form>, <m>, <fld> ) . . . . . . . . . . . compute the wall of <m>
```
fingolfin, line 1922: Does it compute "the wall"? Or the "Wall form of <m> with respect to <form>".
```suggestion
#F  WallForm( <form>, <m>, <fld> ) . . .  compute Wall form of <m> wrt <form>
```

Or maybe simplify your life and just drop this here
```suggestion
#F  WallForm( <form>, <m>, <fld> )
```

grp/classic.gi
```diff
@@ -1917,6 +1917,122 @@ InstallMethod( Omega,
...
+##
+# Input: <m> is an element of the orthogonal group which is given by the form <form>,
+#        <fld> is the finite field of the orthogonal group (in which <m> is defined)
+# Output: The Wall Form of <m> correpsponding to <form>
+# Computes the Wall Form of a matrix. The definition can be found in
+# [Taylor, page 163].
```
fingolfin, line 1928:
```suggestion
##  Return the Wall form of <m>, where <m> is a matrix over <fld> which is orthogonal
##  with respect to the bilinear form <form>, also given as a matrix.
##  For the definition of Wall forms, see [Taylor, page 163].
```
The reference should be clearer, though. In fact I have no idea what this refers to, could you please clarify?

grp/classic.gi
```diff
@@ -1917,6 +1917,122 @@ InstallMethod( Omega,
...
+        if Length(b) = 0  then
+            if w[i] <> 0*w[i]  then
+                Add( b, w[i] );
+                Add( p, i );
+            fi;
+        elif RankMat(b) <> RankMat(Concatenation(b,[w[i]]))  then
+            Add( b, w[i] );
+            Add( p, i );
+        fi;
+    od;
```
fingolfin, line 1949: This computes a lot of ranks. We can improve this slightly by adding a local variable `rank`, which is zero at the start; and then do this:
```suggestion
    p := [];
    rank := 0;
    for i in [ 1 .. Length(w) ]  do
        # add a new row and see if that increases the rank
        Add( b, w[i] );
        if RankMat(b) > rank then
            Add( p, i );
            rank := rank + 1;
        else
            Remove( b ); # rank was not increased, so remove the added row again
        fi;
    od;
```
Of course one can do even better, by also keeping a semi-echelonized version of the matrix `b` around, then sifting the "new" vectors through that; if the result is zero we skip the vector; otherwise we add it to `b` and can add the sifted vector to the semi-echelonized version. Unfortunately GAP seems to lack good primitives for this (something that should be changed); the closes I know of are `CloseMutableBasis` and `IsContainedInSpan`, but they end up reducing the same vector twice. The `cvec` package has a better API for this: `CleanRow( b, v, true, fail )` tests if `v` is in the span of the matrix `b` which must be in semi-echelon form, and return a corresponding boolean, but it also grows `b` if `v` is not yet in there.

grp/classic.gi
```diff
@@ -1917,6 +1917,122 @@ InstallMethod( Omega,
...
+
+    # and return
+    return rec( base := b, pos := p, form := x );
+
+end );
+
+
+#############################################################################
+##
+#F  IsSquareWithoutZeroFFE( fld, e) . . . . . . . . . . . Tests whether <e> (not zero) is a
```
fingolfin, line 1968: How about `IsNonZeroSquareFFE` ?

But why do you need this anyway? Your determinants are always non-zero, aren't they?

fingolfin:
@danielrademacher thank you for the great effort and keeping at it. I am hopeful that future PRs will go through easier

PR #4673, opened 2021-10-12
Make GroupWithGenerators compatible with MatrixObj

ssiccha:
If GroupWithGenerators is called with a list of MatrixObj matrices which are
not compressed matrices, then `ImmutableMatrix` can't be called.

TODO: Add tests.

@fingolfin Do you have an idea on how to avoid the `StructuralCopy` call?


fingolfin (changes requested):

fingolfin:
We'll have to do this eventually but I don't think this old PR will help

PR #4674, opened 2021-10-18
kernel: move GAP_KERNEL_{MINOR,MAJOR}_VERSION to version.h resp. gap_version.h

fingolfin:
The goal is to eventually move `#include "config.h"` from `common.h` into all the `.c` and `.cc` files. This greatly simplifies the installation of headers as part of `make install`. It also means that the same set of headers can be shared by multiple binary builds of GAP; i.e. one could then install plain GAP, HPC-GAP and Julia-GAP in parallel.

This is just a step towards this goal (not the last one, nor the first one), but we are getting pretty close now.

This is a subset of PR #4492.

UPDATE: most of this PR has now been merged via PR #4678. What remains is the code which moves `GAP_KERNEL_{MINOR,MAJOR}_VERSION` to `version.h` resp. `gap_version.h`. The reasons why I give two names is because I've switched around between the two: I prefer the former one as simpler, but I thought for a while that this was perhaps the reason behind the strange failures in the CI tests for the release process, which break down when building HPC-GAP, because `version.h` resp. `gap_version.h` are not found (i.e., they are not generated). But apparently the name is *not* the cause for the breakage after all... very weird

Closes #4679.


fingolfin:
Weird failure in the "Wrap release" CI scripts when building HPC-GAP out of tree. E.g. in [this build log](https://github.com/gap-system/gap/runs/3944108287?check_suite_focus=true), we see:
```
In file included from ../src/ariths.c:17:
../src/modules.h:19:10: fatal error: version.h: No such file or directory
   19 | #include "version.h"
      |          ^~~~~~~~~~~
compilation terminated.
make: *** [../Makefile.rules:357: build/obj/src/ariths.c.lo] Error 1
```

However, I cannot reproduce this on locally (yet).

PR #4683, opened 2021-10-27
Faster Wreath Product `Embedding`

FriedrichRober:
I want to try to add faster implementations for `Embedding` for wreath products and
also fix errors that might happen in `ListWreathProductElement` for top groups not being the symmetric group.
Those issues might be related to the current `Embedding` implementations.

The motivation for this pull request is to fix annoying errors for the `\in` implementation for wreath product provided in my package [WPE](https://github.com/FriedrichRober/WreathProductElements)

### DIFF
--- a/lib/gprdperm.gi
+++ b/lib/gprdperm.gi
@@ -863,7 +863,7 @@ end);
 InstallMethod( ListWreathProductElementNC, "perm wreath product", true,
  [ IsPermGroup and HasWreathProductInfo, IsObject, IsBool ], 0,
 function(G, x, testDecomposition)
-  local info, list, h, f, degK, i, j, constPoints, imageComponents, comp, restPerm;
+  local info, list, h, hIm, f, degK, i, j, constPoints, imageComponents, comp, restPerm;
 
   info := WreathProductInfo(G);
   # The top group element
@@ -871,8 +871,12 @@ function(G, x, testDecomposition)
   if h = fail then
     return fail;
   fi;
+  hIm := ImageElm(Embedding(G, info.degI + 1), h);
+  if hIm = fail then
+    return fail;
+  fi;
   # The product of the base group elements
-  f := x * Image(Embedding(G, info.degI + 1), h) ^ (-1);
+  f := x * hIm ^ (-1);
   list := EmptyPlist(info!.degI + 1);
   list[info.degI + 1] := h;
   # Imprimitive Action, tuple (i, j) corresponds
--- a/lib/gprdperm.gi
+++ b/lib/gprdperm.gi
@@ -610,6 +610,7 @@ local	G,H,	    # factors
     if IsEmpty( domI )  then
         domI := [ 1 ];
     fi;
+    domI := MakeImmutable(Set(domI));
     degI := Length( domI );
 
     # make the generators of the direct product of <deg> copies of <G>
@@ -674,6 +675,7 @@ local	G,H,	    # factors
       basegens:=basegens,
       I      := I,
       degI   := degI,
+      domI   := domI,
       hgens  := hgens,
       components := components,
       embeddingType := NewType(
@@ -773,11 +775,7 @@ InstallMethod( ImagesRepresentative,
     fi;
     # Embedding into top group
     x := g ^ info.alpha;
-    domI := MovedPoints(Range(info.alpha));
-    # force trivial group to act on 1 point
-    if IsEmpty( domI )  then
-      domI := [ 1 ];
-    fi;
+    domI := info.domI;
     degG := NrMovedPoints(info.groups[1]);
     # force trivial group to act on 1 point
     if degG = 0 then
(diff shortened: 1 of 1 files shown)


FriedrichRober:
The pull request is basically finished, but we now have duplicate code in `WreathProduct` and `Embedding`, since computing generators of our wreath product is basically just embedding generators of the components.

I would like to move the duplicate code into `private` functions, so that we can call it from both places. What are the naming conventions for methods which should not be public?

The code is tested by the `WreathProductElementList` calls in `tst/testinstall/opers/ListWreathProductElement.tst`. (and probably a lot of other places)

fingolfin:
Naming convention: unfortunately, we don't have good conventions in place. Sometimes ALL_UPPERCASE is used. You could also use a leading underscore, or even combine those.

lib/gprdperm.gi
```diff
@@ -750,7 +762,35 @@ InstallMethod( ImagesRepresentative,
         [ IsEmbeddingImprimitiveWreathProductPermGroup,
           IsMultiplicativeElementWithInverse ], 0,
     function( emb, g )
-    return g ^ WreathProductInfo( Range( emb ) ).perms[ emb!.component ];
+    local info, degI, x, shift, domI, degG, i, k, l;
+    info := WreathProductInfo(Range(emb));
+    degI := info.degI;
+    if emb!.component = degI + 1 then
```
fingolfin, line 768: I would suggest to reverse the logic: instead of having a loooong case followed by a one line case, how about writing it more like this:
```suggestion
    if emb!.component <> degI + 1 then
      return g ^ info.perms[emb!.component];
    fi;
```
followed by the long case, which now can be indented one level less.

lib/gprdperm.gi
```diff
@@ -750,7 +762,35 @@ InstallMethod( ImagesRepresentative,
...
+      if IsEmpty( domI )  then
+        domI := [ 1 ];
+      fi;
+      degG := Length(MovedPoints(info.groups[1]));
+      # force trivial group to act on 1 point
+      if degG = 0 then
+        degG := 1;
+      fi;
+      shift := [];
+      for i  in [1 .. degI ]  do
```
fingolfin, line 781:
```suggestion
      for i  in [1 .. degI]  do
```

lib/gprdperm.gi
```diff
@@ -750,7 +762,35 @@ InstallMethod( ImagesRepresentative,
...
+    info := WreathProductInfo(Range(emb));
+    degI := info.degI;
+    if emb!.component = degI + 1 then
+      x := g ^ info.alpha;
+      domI := MovedPoints(Range(info.alpha));
+      # force trivial group to act on 1 point
+      if IsEmpty( domI )  then
+        domI := [ 1 ];
+      fi;
+      degG := Length(MovedPoints(info.groups[1]));
```
fingolfin, line 775:
```suggestion
      degG := NrMovedPoints(info.groups[1]);
```

lib/gprdperm.gi
```diff
@@ -750,7 +762,35 @@ InstallMethod( ImagesRepresentative,
...
+    local info, degI, x, shift, domI, degG, i, k, l;
+    info := WreathProductInfo(Range(emb));
+    degI := info.degI;
+    if emb!.component = degI + 1 then
+      x := g ^ info.alpha;
+      domI := MovedPoints(Range(info.alpha));
+      # force trivial group to act on 1 point
+      if IsEmpty( domI )  then
+        domI := [ 1 ];
+      fi;
```
fingolfin, line 774: How about just storing this as `info.domI` when creating the morphism? Perhaps instead of `info.degI` which could be replaced by `Length(info.domI)`.

Motivation: simplifies the code a bit, ensures consistency, and is a bit better performance wise (fewer allocations and less code).

lib/gprdperm.gi
fingolfin: Also, perhaps store this as `info.degG`?

lib/gprdperm.gi
fingolfin: Oh and when you create and store `domI` this way, make sure to pass it through `MakeImmutable(Set(...))` first, so that GAP stores that it is a sorted list of integers, and can later perform faster searches on it, and have reduced method lookup costs.

lib/gprdperm.gi
```diff
@@ -764,11 +804,15 @@ InstallMethod( PreImagesRepresentative,
     function( emb, g )
     local info;
     info := WreathProductInfo( Range( emb ) );
-    if not g in info.base then
-      return fail;
+    if emb!.component = info.degI + 1 then
+      return g ^ Projection(Range(emb));
+    else
```
fingolfin, line 809: How about
```suggestion
    fi;
```
and then shift everything afterwards down one indent level?

hulpke (approved):
While one can argue about minutiae of coding, I think this is overall a nice contribution and could be merged as is.

lib/gprdperm.gi
FriedrichRober: Yes, I will change this.

lib/gprdperm.gi
FriedrichRober: This is a good idea.

fingolfin (approved):

Issue #4695, opened 2021-11-11
`PROGRAM_CLEAN_UP()`  (and thus exit handlers) not always run when quitting GAP


fingolfin:
When GAP exits, then normally exit handlers installed via `InstallAtExit` should be run; this is done by the function `PROGRAM_CLEAN_UP()`. (The exception being when `ForceQuitGap()` is used, but that's not relevant here).

Alas, this is not always the case. Consider this test program, stored as `foo.g`:
```gap
Print("hello\n");
InstallAtExit(function() Print("bye\n"); end);
QUIT;
```
When we interactively `Read` this program, all is fine:
```console
$ gap -A -b
gap> Read("foo.g");
hello
bye
$
```
But if we instead pass the file name to GAP as an argument so that it is read as an init file by `ProcessInitFiles`, the exit handlers are *not* invoked:
```console
$ gap -A -b foo.g
hello
$
```

What goes wrong is this: entering `QUIT` leads to the C function `ReadEvalCommand()` returning the execution status `STATUS_QQUIT`. The various read/eval loops in the kernel then set the kernel variable `STATE(UserHasQUIT)` to 1 to indicate the user requested to 

Now, in an interactive session, what happens is that the REPL (emphasis on *L* = *loop* here) in the kernel function `SHELL()` exits and does this (code edited for clarity, I'll submit the edits in a future PR)
```C
  if (STATE(UserHasQUIT)) {
      // If we are in a break loop, throw so that the next higher up read&eval
      // loop can process the QUIT
      if (breakLoop)
          GAP_THROW();
      // If we are the topmost REPL, then indicating we are QUITing to the GAP
      // language level, and simply end the loop. This implicitly assumes that
      // the only places using SHELL() are the primary REPL and break loops.
      STATE(UserHasQuit) = 0;
      STATE(UserHasQUIT) = 0;
      AssGVarWithoutReadOnlyCheck(QUITTINGGVar, True);
      return Fail;
  }
```

But init files are processed *before* the main REPL is started (via the GAP command `SESSION()` that ultimately invokes the kernel function `SHELL()`, a wrapper around the C function `Shell()`). Thus what happens is that as soon as the `READ` function processes the `QUIT` statement, `UserHasQUIT` gets set to 1, and stays this way. 

The read/eval loop processing `init.g` (implemented in the C function `READ_INNER`) then aborts processing, and GAP exits. The instructions invoking `PROGRAM_CLEAN_UP()` are never reached. In fact, the file

[...]

PR #4696, opened 2021-11-11
kernel: refactor SHELL

fingolfin:
Just some refactoring, motivated in parts by issue #4695. For reviewing it is probably best to view each commit separately.

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -168,6 +168,9 @@ static Obj FuncSHELL(Obj self,
                      Obj prompt,
                      Obj preCommandHook)
 {
+    //
+    // validate all arguments
+    //
     if (!IS_LVARS_OR_HVARS(context))
         RequireArgument(SELF_NAME, context, "must be a local variables bag");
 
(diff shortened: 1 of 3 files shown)


src/gap.c
```diff
@@ -160,235 +160,226 @@ void ViewObjHandler ( Obj obj )
...
+        }
+        STATE(ErrorLVars) = errorLVars;
+
+        // read and evaluate one command (statement or expression)
+        UInt dualSemicolon;
+        status =
+            ReadEvalCommand(errorLVars, &input, &evalResult, &dualSemicolon);
+
+        // if the user called some code which QUIT then bail out (the case
+        // were the statement we just processed was `QUIT` is handled below,
```
wilfwilson, line 285:
```suggestion
        // where the statement we just processed was `QUIT` is handled below,
```

wilfwilson (approved):
I noticed some very very minor things, but apart from that, this looks very nice

In the commit message `kernel: remove one assert (only active in debug builds) by Panic` I think you mean 'replace' not 'remove'

PR #4697, opened 2021-11-12
Disabled `NaturalHomomorphism` for `FactorGroup`, as it has side effects.

### LINKED ISSUE #4656: Confusing behaviour in repeated calls of the same code
The following happens in the current master branch as well as in GAP 4.11.1.

I execute some code (essentially from a manual example) in a new GAP session.
```
gap> G:= Group( (1,2,3,4), (1,2) );;                                           
gap> N:= DerivedSubgroup( G );;
gap> F:= FactorGroupNC( G, DerivedSubgroup( G ) );
Group([ f1 ])
gap> HasParent( F );
true
gap> IsIdenticalObj( F, Parent( F ) );
true
```
When I enter the same lines again in the same session, the result looks different.
```
gap> G:= Group( (1,2,3,4), (1,2) );;
gap> N:= DerivedSubgroup( G );;
gap> F:= FactorGroupNC( G, DerivedSubgroup( G ) );
<pc group with 1 generators>
gap> HasParent( F );
false
```
(I observed this difference because it appears as a difference in manual tests after the changes from pull request #4653.)

Does it make sense at all to set `F` as its own `Parent` value in the first case? And if yes then why is the parent not set in the second case (and in all subsequent calls during the session)?

hulpke:
(Moved into obsoletes).
This resolves #4656

Also included the (trivial) fix for #4698

## Text for release notes

The use of `FactorGroup` succeeded by `NaturalHomomorphism` has been made obsolete, as the implementation relied on side-effects.

### DIFF
--- a/lib/grp.gi
+++ b/lib/grp.gi
@@ -2669,6 +2669,7 @@ function( G, N )
 local hom,F,new;
   hom:=NaturalHomomorphismByNormalSubgroupNC( G, N );
   F:=ImagesSource(hom);
+  #TODO: Remove the !.nathom component
   if not IsBound(F!.nathom) then
     F!.nathom:=hom;
   else
@@ -2681,16 +2682,6 @@ local hom,F,new;
   fi;
   return F;
 end );
-
-InstallMethod( NaturalHomomorphism, "for a group with natural homomorphism stored",
-    [ IsGroup ],
-function(G)
-  if IsBound(G!.nathom) then
-    return G!.nathom;
-  else
-    Error("no natural homomorphism stored");
-  fi;
-end);
 
 InstallOtherMethod( \/,
     "generic method for two groups",
--- a/lib/obsolete.gi
+++ b/lib/obsolete.gi
@@ -1029,6 +1029,21 @@ end );
 BindGlobal("R_N", fail);
 BindGlobal("R_X", fail);
 
+# Moved to obsolete in Nov. 2021 for 4.12
+InstallMethod( NaturalHomomorphism, "for a group with natural homomorphism stored",
+    [ IsGroup ],
+function(G)
+  Info(InfoWarning,0,"The use of `NaturalHomomorphism` for a `FactorGroup`\n",
+    "has been deprecated, as it caused side-effects.\n",
+    "Proceed at risk!");
+
+  if IsBound(G!.nathom) then
+    return G!.nathom;
+  else
+    Error("no natural homomorphism stored");
+  fi;
+end);
+
 
 
 #############################################################################
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -2848,8 +2848,10 @@ DeclareOperation( "CosetTableNormalClosure", [ IsGroup, IsGroup ] );
 ##
 ##  <Description>
 ##  returns the image of the <C>NaturalHomomorphismByNormalSubgroup(<A>G</A>,<A>N</A>)</C>.
-##  The homomorphism will be returned by calling the function
-##  <C>NaturalHomomorphism</C> on the result.
+##  This function is provided for compatibility with older code, but if a
+##  connection between group and factor is desired, users need to start by
+##  obtaining the <C>NaturalHomomorphismByNormalSubgroup</C> in the first
+##   place.
 ##  The <C>NC</C> version does not test whether <A>N</A> is normal in <A>G</A>.
 ##  <Example><![CDATA[
 ##  gap> g:=Group((1,2,3,4),(1,2));;n:=Subgroup(g,[(1,2)(3,4),(1,3)(2,4)]);;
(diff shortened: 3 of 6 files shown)


lib/grp.gi
```diff
@@ -2685,6 +2685,9 @@ end );
 InstallMethod( NaturalHomomorphism, "for a group with natural homomorphism stored",
     [ IsGroup ],
 function(G)
+  Error("The use of `NaturalHomomorphism` for a `FactorGroup`\n",
+    "has been removed, as it caused side-effects.");
```
fingolfin, line 2689: But won't that break existing user code that relies on this?

hulpke: Yes, should code rely onthis functionality. We could change it to add text "enter `return;` to continue, or change the error to a warning message.

fingolfin: Also we should perhaps quickly check if any distributed packages rely on this?

hulpke: I have checked through the packages in 4.11. Only XMod,  laguna, primgrp, rds, wedderga and xgap call `FactorGroup(`. None of them uses `NaturalHomomorphism(`. I will change the error to a warning.

lib/obsolete.gi
```diff
@@ -1029,6 +1029,21 @@ end );
 BindGlobal("R_N", fail);
 BindGlobal("R_X", fail);
 
+# Moved to obsolete in Nov. 2021 for 4.12
+InstallMethod( NaturalHomomorphism, "for a group with natural homomorphism stored",
+    [ IsGroup ],
+function(G)
+  Info(InfoWarning,0,"The use of `NaturalHomomorphism` for a `FactorGroup`\n",
+    "has been removed, as it caused side-effects.\n",
```
fingolfin, line 1037: It has not (yet) been removed, though, has it? Perhaps say "deprecated" or so instead?
```suggestion
    "has been deprecated, as it caused side-effects.\n",
```

fingolfin (approved):
Thank you

lib/grp.gi
cdwensley: XMod does use NaturalHomomorphism

hulpke: @cdwensley Where does it use `NaturalHomomorphism`? I found it uses `NaturalHomomorphismByNormalSubgroup` (which is safe and well) only in version 2.82.

cdwensley: I'm happy to change it if that is the sensible thing to do. 
Christopher-Wensleys-iMac:lib christopherwensley$ grep -n NaturalHomomorphism *
catone.gi:575:                nat:=NaturalHomomorphismByNormalSubgroup(G,N);
gp2map.gi:1764:    nat1 := NaturalHomomorphism( KR1 ); 
gp2map.gi:1795:                          nat2 := NaturalHomomorphism( KQ ); 
gp2map.gi:1862:    nat1 := NaturalHomomorphism( KR1 ); 
gp2map.gi:1891:                        natKQ := NaturalHomomorphism( KQ );
gp2map.gi:1897:                        natKR1 := NaturalHomomorphism( KR1 );
gp2map.gi:1902:                        natKQ := NaturalHomomorphism( KQ );

cdwensley: Yes, this does appear to be a recent change.

hulpke: Since this is a recent change (and thus possibly not even in a distributed version) I would urge to replace code of the form
```
F:=FactorGroup(G,N);
hom:=NaturalHomomorphism(F);
```
by
```
hom:=NaturalHomomorphismByNormalSubgroup(G,N);
F:=Image(hom,G);
```
The cost will be the same, but it does not rely on the side-effect of aching the homomorphism in the object.

cdwensley: It all happened when quasi-isomorphisms were introduced: PR #107 (17/2/21). 
I'll try using NaturalHomomorphismByNormalSubgroup instead.

cdwensley:
I recall now that this advice was given and followed some years ago.
As the operation was still available, it was used again this year.
The 5 occurrences of NaturalHomomorphism in XMod have now been replaced.
Since this is a recent change (and thus possibly not even in a 
distributed version) I would urge to replace code of the form
    F:=FactorGroup(G,N);   hom:=NaturalHomomorphism(F);
by
    hom:=NaturalHomomorphismByNormalSubgroup(G,N);
    F:=Image(hom,G);
The cost will be the same, but it does not rely on the side-effect of 
aching the homomorphism in the object.

fingolfin:
Unfortunately something in here broke the test suite, specifically `tst/teststandard/grpprmcs.tst:1813`, which I didn't notice as there is this 32bit float CI error... *sigh*. 
```
testing: /home/runner/work/gap/gap/tst/teststandard/grpprmcs.tst
########> Diff in /home/runner/work/gap/gap/tst/teststandard/grpprmcs.tst:1813
# Input is:
ImagesRepresentative(iso,Product(GeneratorsOfGroup(g)));
# Expected output:
F3*F4*F5*F2*F1^-1*F2^2*(F1*F2^-1)^2
# But found:
F3*F4*F5*(F1^-1*F2^-1)^2
########
```

hulpke:
Yes. This is a side-effect of a small improvement to `SmallGeneratingSet' which I had added since the PR was hanging anyhow (did not realize you would then override the merge).
The test is simply for evaluating the homomorphism on an element that is not a generator, but since the generating set changed, the image (entirely harmlessly) also changed. Do you want me to submit a PR for changing the test example?

PR #4702, opened 2021-11-16
CI: cancel redundant PR builds

fingolfin:
When updating a PR, automatically cancel any previously running instances
of the build.

Borrowed from https://github.com/JuliaCI/rootfs-images/blob/192cf1edb49f128f3eb9d4ffda2f6675a8d0a05d/.github/workflows/linux.yml#L14-L18

See also https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

I think we'll need to merge it in order to test it...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,6 +9,12 @@ on:
     # Every day at 2:30 AM UTC
     - cron: '30 2 * * *'
 
+concurrency:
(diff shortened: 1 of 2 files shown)


wilfwilson (approved):

wilfwilson:
Happy to give it a go.

PR #4703, opened 2021-11-16
fixed `Coefficients` for large finite fields created via a polynomial

### LINKED ISSUE #4694: problem with bases of finite fields that are constructed using prescribed polynomials
The following happens in GAP 4.11 and in the master branch.

```
gap> p:= 1000003;;  Fp:= GF(p);;  # a prime field
gap> pol:= UnivariatePolynomial(Fp, [ 912543, 810, 1 ] * One(Fp) );
x_1^2+ZmodpZObj(810,1000003)*x_1+ZmodpZObj(912543,1000003)
gap> F:= GF(Fp, pol);
<field of size 1000006000009>
```
The field `F` is both a finite field and an algebraic extension of the prime field.
There are two applicable methods for `CanonicalBasis`, one for each of these two aspects;
the method `"for a finite field"` has higher rank.
```
gap> B:= CanonicalBasis( F );;
gap> RepresentationsOfObject( B );
[ "IsComponentObjectRep", "IsAttributeStoringRep", "IsBasisFiniteFieldRep" ]
```
The problem is that the `Coefficients` method for this basis expects that the `LeftActingDomain` of the field is a subset of the field,
which is not the case for the given field. As a consequence, the coefficients are elements of the field, not of its `LeftActingDomain`.
```
gap> coeffs:= Coefficients( B, One( F ) );
[ !ZmodpZObj(1,1000003), !ZmodpZObj(0,1000003) ]
gap> IsSubset( F, coeffs );
true
gap> IsSubset( LeftActingDomain( F ), coeffs );
false
```
This problem disappears when we restrict the `CanonicalBasis` method `"for a finite field"` to situations where the `LeftActingDomain` of the field is a subset of the field, by calling `TryNextMethod()` if this conditions is not satisfied.

The `Basis` method `"for a finite field, and a hom. list"` is in fact already restricted to the situation that the field consists of `FFE` elements, this is safe.
The `BasisNC` method `"for a finite field, and a hom. list"` is installed more generally, thus we run into the same problem as with `CanonicalBasis`.

ThomasBreuer:
resolves #4694

### DIFF
--- a/lib/fieldfin.gi
+++ b/lib/fieldfin.gi
@@ -475,7 +475,7 @@ InstallMethod( Basis,
 InstallMethod( BasisNC,
     "for a finite field, and a hom. list",
     IsIdenticalObj,
-    [ IsField and IsFinite, IsHomogeneousList ], 10,
+    [ IsField and IsFinite, IsFFECollection and IsList ], 10,
     function( F, gens )
 
     local B,     # the basis, result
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Looks good to me

PR #4705, opened 2021-11-22
buildsys: add ABI_CFLAGS back into {C,CXX,LD}FLAGS

fingolfin:
We retain them in CC and CXX, so now when building 32bit GAP on a 64 bit machine, we may  pass -m32 twice to the compiler; but that doesn't hurt.

But on the upside, this way packages which obtain compiler flags from flags.

This is meant to help packages like `float` (which currently fails to build in our 32bit CI tests), but also should simplify `bin/BuildPackages.sh`, which currently pushes the `-m32` bit flag via CFLAGS/CXXFLAGS/LDFLAGS -- with this PR, this shouldn't be necessary anymore for most (all?) packages

This PR is still work in progress, I did not yet have time to properly test and debug it, but I wanted to get it out there so I can reference it.

PR #4707, opened 2021-11-25
Fix return marker

ChrisJefferson:
This is a subtle issue -- we were putting pointers to a C object into plists, which is fine until we do SaveWorkspace. This was (I think) breaking 32-bit tests (the bug was shown when float was loaded, but I think it's not float's fault).

The reason this was so hard to track down is (also, I think) usually bags with this object in them are GCed before we SaveWorkspace, but if things are lined up just right the bag can end up staying alive due to "false sharing".

I also add a GAP_ASSERT, as I hadn't noticed this failure was occurring. I am tempted to suggest we make SaveWorkspace always fail if the test for "bad bags" fails, but I thought to start I'd put in a GAP_ASSERT and see if we notice it triggering.

### DIFF
--- a/src/intrprtr.c
+++ b/src/intrprtr.c
@@ -126,7 +126,7 @@ static Obj VoidReturnMarker;
 
 static void PushFunctionVoidReturn(IntrState * intr)
 {
-    PushPlist(intr->StackObj, (Obj)&VoidReturnMarker);
+    PushPlist(intr->StackObj, VoidReturnMarker);
 }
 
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
I have rebased this on top of #4705 to see if it fixes the crash

fingolfin:
Seems sensible, even though it doesn't fix that crash. If you drop the extra commit, we can merge it anyway?

fingolfin (approved):

PR #4710, opened 2021-11-29
kernel: improve position tracking after ReadEvalCommand

fingolfin:
When evaluating inputs, ReadEvalCommand and the GAP scanner greedily read full
lines of input, even if only a part of the line ends up being evaluated; e.g.
because there are multiple commands on a line.

Unfortunately, this rendered the (undocumented) command `READ_COMMAND_REAL`
far less useful: when invoked on a generic input stream, it evaluates exactly
one command; but afterwards the end of that command can not be inferred from
the position of the input stream, as it always reads full lines.

This patch changes things so that when closing an input *stream*, backed by a
GAP stream object, we try adjust its seek position to where the actual command
ended.

This means that now one can repeatedly call `READ_COMMAND_REAL` on a string
stream to iteratively evaluated the statements in the underlying string.
Combined with `CALL_WITH_STREAM` this mostly allows to reimplement
`READ_ALL_COMMANDS` in GAP code, but with far more flexibility. This is
potentially useful for e.g. the Jupyter interface.

(The one thing that is not yet available is detection of dual semicolons,
which `READ_ALL_COMMANDS` supports but `READ_COMMAND_REAL` does not. But this
could be tacked on to `READ_ALL_COMMANDS`, or we could implement a new
variant of this command providing that information.)

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -21,6 +21,7 @@
 
 #include "io.h"
 
+#include "ariths.h"
 #include "bool.h"
 #include "calls.h"
 #include "error.h"
@@ -53,6 +54,9 @@ static Obj ReadLineFunc;
 static Obj WriteAllFunc;
 static Obj IsInputStringStream;
 static Obj IsOutputStringStream;
+static Obj PositionStream;
+static Obj SeekPositionStream;
+
 static Obj PrintPromptHook = 0;
 Obj EndLineHook = 0;
 static Obj PrintFormattingStatus;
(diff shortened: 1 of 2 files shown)


fingolfin:
One more thing missing in `READ_COMMAND_REAL` is that it doesn't allow detecting syntax errors, as it treats the `ReadEvalCommand` return values `STATUS_END` and `STATUS_ERROR` both the same way.

For that matter, also `return`, `quit` and `QUIT` are not distinguished.

As such, perhaps there should be a new function wrapping `ReadEvalCommand` more tightly, and which returns the full return status -- either as a numerical constant (and then we export the values `STATUS_ERROR` etc. to the GAP level), or as a string (e.g. `"end"`, `"return"`, `"eof"`, ...). That should then make it possible to emulate a REPL in pure GAP code much more accurately, e.g. the `Demonstration` function in `lib/demo.g` would benefit from such a thing.

ChrisJefferson:
I agree that might be better, but this is better than what we had before.

ChrisJefferson (approved):

PR #4713, opened 2021-12-07
fixed `ListOp` for length zero `IsVectorObj`

ThomasBreuer:
Such vectors can arise for example as coefficient lists of polynomials,
for example of `2 * One( Indeterminate( GF(2) ) )`.

There are still issues with length zero vectors.
If one wants to convert  a length zero vector in `IsGF2VectorRep` to `Is8BitVectorRep` then one runs into an error.
```
gap> v:= [ Z(2) ];;  ConvertToVectorRep( v, 2 );;  Unbind( v[1] );  v;
<a GF2 vector of length 0>
gap> ConvertToVectorRep( v, 4 );
Error, List Element: <list>[1] must have an assigned value in
[...]
```
(The conversion in the other direction works.)

I am not sure whether we should fix this, and how.

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -567,10 +567,12 @@ InstallMethod( ListOp,
   local result, i, len;
   len := Length(vec);
   result := [];
-  result[len] := vec[len];
-  for i in [ 1 .. len - 1 ] do
-    result[i] := vec[i];
-  od;
+  if 0 < len then
+    result[len] := vec[len];
+    for i in [ 1 .. len - 1 ] do
+      result[i] := vec[i];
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
As far as I understand, one of the tests failed because the `float` package cannot be built.

fingolfin (approved):
Looks good to me.

The issue with `float` failing is known. There is an update for float that ought to resolve this. Unfortunately @alex-konovalov did not yet have time to update the package distribution with this fix (the upcoming GAP Days seek to resolve the bottleneck of only person being able to do this kind of update)

PR #4719, opened 2021-12-11
Improve `NiceMonomorphism` docs

fingolfin:
... and point to `NiceObject`. But note that the two are right next to each others, so when looking at them in the help system or on the website, one normally sees both at once. But this way, one is explicitly told to look at that example.

### DIFF
--- a/lib/grpnice.gd
+++ b/lib/grpnice.gd
@@ -26,6 +26,10 @@
 ##  and whose restriction to <A>obj</A> is injective.
 ##  The concrete morphism (and also the image group) will depend on the
 ##  representation of <A>obj</A>.
+##  <P/>
+##  WARNING: The domain of the homomorphism may be larger than <A>obj</A>.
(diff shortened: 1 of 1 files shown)


laurentbartholdi:
This looks good, thanks! However:

Is it guaranteed that `IsSubgroup(Image(NiceMonomorphism(g)),NiceObject(g))`? In the current implementation with attributes, this is fine since the nice monomorphism is computed once and then cached; but e.g. in `LatticeSubgroups` the code needs both the nice object and the nice monomorphism, and the paranoid in me fears that after some implementation change they could stop being linked.

fingolfin:
> This looks good, thanks! However:
> 
> Is it guaranteed that `IsSubgroup(Image(NiceMonomorphism(g)),NiceObject(g))`? 

Yes. That's one of the fundamental principals of the whole nice monomorphism system.

hulpke (approved):

wilfwilson (approved):
Nice improvement.

PR #4720, opened 2021-12-14
Fix a few test regressions

fingolfin:
... introduced in PR #4697.

Incidentally, the regression was apparently caused by a commit in that PR which was added in after my review or the PR and which was (as far as I can tell) unrelated to the PR topic? This is *not* meant to assign blame (if at all, that belongs with me: I merged, so it's my responsibility!) Rather the main problem was/is is that one repeatedly failing CI job goads slow humans like myself into ignoring the CI results overall for jobs that "passed just fine yesterday, and how should this cause regressions anyway" *sigh*. Lesson learned: double check in these situations before hitting "merge". Also, next time we should be quicker to resolve the regressions, if needs by a "brutal" approach, like "don't test float in `testpackages` until it is fixed" (which I do in this PR). Luckily, the 32bit CI failure will be fixed once float 1.0.2 has been picked up for the package distribution by @alex-konovalov.

### DIFF
--- a/dev/ci.sh
+++ b/dev/ci.sh
@@ -47,6 +47,9 @@ do
     # skip itc because it requires xgap
     rm -rf itc*
 
+    # HACK: skip float until 1.0.2 is released, as it crashes in the 32bit builds
+    rm -rf float-1.0.1
+
     # HACK to work out timestamp issues with anupq
     touch anupq*/configure* anupq*/Makefile* anupq*/aclocal.m4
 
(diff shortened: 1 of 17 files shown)


hulpke (approved):
These are all fine (and basically just needed because the testing setup is too rigid).

wilfwilson (approved):
There's still one in a test file.

wilfwilson (approved):

alex-konovalov:
Anyhow, Float 1.0.2 is now in https://github.com/gap-system/gap-distribution/releases/tag/package-archives

PR #4721, opened 2021-12-29
kernel: minor tweaks to read-eval code

fingolfin:
- kernel: turn dualSemicolon into a BOOL
- kernel: document ReturnObjStat in gapstate.h
- kernel: merge STATUS_RETURN_VAL and STATUS_RETURN_VOID
- kernel: ReadEvalCommand returns an ExecStatus

### DIFF
--- a/src/gap.h
+++ b/src/gap.h
@@ -56,15 +56,14 @@ void ViewObjHandler(Obj obj);
 typedef UInt ExecStatus;
 
 enum {
-    STATUS_END         =  0,    // ran off the end of the code
-    STATUS_RETURN_VAL  =  1,    // value returned
-    STATUS_RETURN_VOID =  2,    // void returned
-    STATUS_BREAK       =  4,    // 'break' statement
-    STATUS_QUIT        =  8,    // quit command
-    STATUS_CONTINUE    =  8,    // 'continue' statement
-    STATUS_EOF         = 16,    // End of file
-    STATUS_ERROR       = 32,    // error
-    STATUS_QQUIT       = 64,    // QUIT command
+    STATUS_END      = 0,    // ran off the end of the code
+    STATUS_RETURN   = 1<<0, // 'return' statement
+    STATUS_BREAK    = 1<<1, // 'break' statement
+    STATUS_CONTINUE = 1<<2, // 'continue' statement
+    STATUS_QUIT     = 1<<3, // 'quit' statement
+    STATUS_QQUIT    = 1<<4, // 'QUIT' statement
+    STATUS_EOF      = 1<<5, // end of file while parsing
+    STATUS_ERROR    = 1<<6, // syntax error while parsing
 };
 
 
--- a/src/stats.c
+++ b/src/stats.c
@@ -64,13 +64,21 @@ extern inline Obj EXEC_CURR_FUNC(void)
     return result;
 }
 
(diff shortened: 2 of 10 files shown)


src/gap.h
```diff
@@ -56,15 +56,14 @@ void ViewObjHandler(Obj obj);
...
-    STATUS_ERROR       = 32,    // error
-    STATUS_QQUIT       = 64,    // QUIT command
+    STATUS_END         = 0,     // ran off the end of the code
+    STATUS_RETURN      = 1<<0,  // 'return' statement
+    STATUS_QUIT        = 1<<1,  // 'quit' statement
+    STATUS_QQUIT       = 1<<2,  // 'QUIT' statement
+    STATUS_EOF         = 1<<3,  // End of file
+    STATUS_ERROR       = 1<<4,  // error
+    STATUS_BREAK       = 1<<5,  // 'break' statement
+    STATUS_CONTINUE    = 1<<6,  // 'continue' statement
```
fingolfin, line 66: Perhaps I should add a comment that indicates that these enum values are used as "bit masks", but only as a minor optimization. In general, any status value should have precisely one bit set.

src/gap.h
```diff
@@ -56,15 +56,14 @@ void ViewObjHandler(Obj obj);
...
-    STATUS_ERROR       = 32,    // error
-    STATUS_QQUIT       = 64,    // QUIT command
+    STATUS_END      = 0,    // ran off the end of the code
+    STATUS_RETURN   = 1<<0, // 'return' statement
+    STATUS_BREAK    = 1<<1, // 'break' statement
+    STATUS_CONTINUE = 1<<2, // 'continue' statement
+    STATUS_QUIT     = 1<<3, // 'quit' statement
+    STATUS_QQUIT    = 1<<4, // 'QUIT' statement
+    STATUS_EOF      = 1<<5, // end of file while parsing
+    STATUS_ERROR    = 1<<6, // syntax error while parsing
```
fingolfin, line 66: I've reordered these values now so that the first four are precisely the four that can occur when executing a loop or a function; and the final two indicate parsing failures; and the middle ones are all about statement types; and QUIT/QQUIT are next to each other.

src/stats.c
```diff
@@ -64,13 +64,21 @@ extern inline Obj EXEC_CURR_FUNC(void)
...
+// which is not break/continue/return) is handled first.
 #define EXEC_STAT_IN_LOOP(stat)                                              \
     {                                                                        \
         UInt status = EXEC_STAT(stat);                                       \
-        if (status != 0) {                                                   \
+        if (status != STATUS_END) {                                          \
             if (status == STATUS_CONTINUE)                                   \
                 continue;                                                    \
-            return (status & (STATUS_RETURN_VAL | STATUS_RETURN_VOID));      \
+            return (status == STATUS_RETURN) ? STATUS_RETURN : STATUS_END;   \
```
fingolfin, line 81: This part of the change is actually a bit subtle and I initially got it wrong -- and it still worked *by pure chance*. What I had done was this:
```suggestion
            return status == STATUS_RETURN
```
Which actually does the same as long as `STATUS_RETURN` is 1 and `STATUS_END` is 0, as is the case right now. But when I later experimented with changing these values, I got weird errors that I didn't understand at first.

I've now added a comment explaining what's going on here...

wilfwilson (approved):
All looks sensible to me.

PR #4729, opened 2022-01-07
kernel: turn ExecStatus into a proper enum

fingolfin:
... and don't use it as a bitmask. It was a bitmask before as a (minor)
performance optimization. Alas, in most call sites, this did not matter: the
read-eval-loops in gap.c and streams.c are not affected by a few extra
instructions, especially if one is careful to first check the common case
(STATUS_END) before the exceptional cases (STATUS_QUIT, STATUS_QQUIT,
STATUS_ERROR, STATUS_EOF)

The only potentially performance critical check was in `EXEC_STAT_IN_LOOP`,
which was already switched away from using the bitmasks in a previous commit.
But even there, really only the case of a `break` or `return` statement
inside of a loop is affected, which once again is an "exceptional" case.

(All that said, if we want better performance for executing GAP code, we'd have to
implement a bytecode VM. If anybody wants to tackle that or has a student they'd like to assign to tackling that, I'd be happy to (co-)mentor this, I've thought about it plenty and could give some pointers as to how and where to start.)

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2506,10 +2506,10 @@ static void RecreateStackNams(ReaderState * rs, Obj context)
 **  will be set to 1 if the command was followed by a double semicolon, else
 **  it is set to 0. If 'dualSemicolon' is zero then it is ignored.
 */
-ExecStatus ReadEvalCommand(Obj            context,
-                           TypInputFile * input,
-                           Obj *          evalResult,
-                           BOOL *         dualSemicolon)
+UInt ReadEvalCommand(Obj            context,
+                     TypInputFile * input,
+                     Obj *          evalResult,
+                     BOOL *         dualSemicolon)
 {
     volatile ExecStatus status;
     volatile Obj        tilde;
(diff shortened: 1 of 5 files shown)


src/read.c
```diff
@@ -2506,10 +2506,10 @@ static void RecreateStackNams(ReaderState * rs, Obj context)
 **  will be set to 1 if the command was followed by a double semicolon, else
 **  it is set to 0. If 'dualSemicolon' is zero then it is ignored.
 */
-ExecStatus ReadEvalCommand(Obj            context,
-                           TypInputFile * input,
-                           Obj *          evalResult,
-                           BOOL *         dualSemicolon)
+UInt ReadEvalCommand(Obj            context,
```
fingolfin, line 2509: This change was necessary because in `read.h` this function is declared as returning `UInt`, which is now a different type than `ExecStatus`. I have another patch ready which builds on this one and changes it to use `ExecStatus`, but that requires moving the `typedef ExecStatus` to `common.h` and I'd rather not mix these changes unnecessarily.

PR #4730, opened 2022-01-07
kernel: use ExecStatus instead of UInt where possible

fingolfin:
- kernel: turn ExecStatus into a proper enum (PR #4729)
- kernel: use ExecStatus instead of UInt where possible

This contains #4729. So either merge this PR to get both; or if PR #4729 is merged first, I can rebase this one afterwards.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -246,6 +246,25 @@ typedef struct TypOutputFile TypOutputFile;
 
 /****************************************************************************
 **
+*T  ExecStatus . . . .  type of status values returned by read, eval and exec
+**                      subroutines, explaining why evaluation, or execution
+**                      has terminated.
+*/
+
+typedef enum {
+    STATUS_END,         // ran off the end of the code
+    STATUS_RETURN,      // 'return' statement
+    STATUS_BREAK,       // 'break' statement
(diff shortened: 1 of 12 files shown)


src/common.h
```diff
@@ -244,6 +244,28 @@ typedef struct TypInputFile TypInputFile;
...
 
 
+/****************************************************************************
+**
+*T  ExecStatus . . . .  type of status values returned by read, eval and exec
+**                      subroutines, explaining why evaluation, or execution
+**                      has terminated.
+**
+**  Values are powers of two; this is used to test a given status for
+**  multiple possible values simultaneously.
```
ChrisJefferson, line 254: this statement isn't true?

ChrisJefferson: on reading, I believe the plan is to remove this 'power of two' and not do bit-oring tricks (which I agree are better to get rid of), so it's just this comment that needs removing?

fingolfin: those two lines are removed in the first patch and then accidentally re-introduced in the second. I'll get rid of them.

fingolfin: Fixed now, thanks for pointing this out!

ChrisJefferson (approved):

Issue #4682, opened 2021-10-27
buildsys: should we drop libtool and simply use our own list of build rules?


fingolfin:
The main reason to use libtool for us is that it provides a uniform way to generate shared libraries and loadable modules (for kernel extensions). That used to be quite complicated because every vendor had their own way of doing it, and they differed, sometimes by a lot.

But these days, things are much simpler: essentially, you have to cover Linux, macOS, Windows, *BSD.

And using libtool has its costs, too:
- create `.la` and `.lo` files and `.libs` dirs
- involves a huge complicated shell script, intransparent
- relatively slow (doesn't matter *much* for us)
- complicates installation of GAP, as `gac` currently calls into `libtool`, so we'd have to install a copy of it (and some distros frown upon that, or at least it used to be that way)

I've been using this as a replacement in GAP.jl for some time for building loadable modules: 
```julia
        if Sys.islinux() || Sys.isfreebsd()
            c_compiler = "$(CC) -fPIC -DPIC"
            cxx_compiler = "$(CXX) -fPIC -DPIC"
            c_dyn_linker = "$(CC) -shared -fPIC -DPIC"
        elseif Sys.isapple()
            c_compiler = "$(CC) -fno-common -DPIC"
            cxx_compiler = "$(CXX) -fno-common -DPIC"
            c_dyn_linker = "$(CC) -bundle"
        else
            error("OS not supported")
        end
```
Note that it lacks code for Windows, but I am sure we could overcome that. It also doesn't cover building the `libgap.so` shared library, but this requires rather similar code.

The main *downside* to dropping libtool is that we may loose support for some operating systems which someone (who??) out there might still care about. But I don't expect many (if any) of these, and adding back support for such a system generally is not much harder than the above.

Note that this would essentially return us to how we did things before the new build system was introduced in GAP 4.9, so it wouldn't be really something new either. Specifically, we had this (I've removed a bunch of useless parts dealing with NeXTStep, Irix etc.):
```sh
AC_DEFUN(GP_PROG_CC_DYNFLAGS,
[AC_CACHE_CHECK(dynamic module compile options, gp_cv_prog_cc_cdynoptions,
 [ case "$host-$CC" in
    *cygwin* )
        gp_cv_prog_cc_cdynoptions="${ABI_CFLAGS}";;
    *-apple-darwin*gcc* )
        gp_cv_prog_cc_cdynoptions="-fPIC -Wall ${ABI_CFLAGS}";;
    *-gcc* | *-egcs )
[...]

PR #4731, opened 2022-01-09
buildsys: progress towards `make install`

fingolfin:
- `sysinfo.gap`: adjust comments; remove `GAP_LIBTOOL_CURRENT` and `GAP_LIBTOOL_AGE`
- `Makefile.rules`: add `install-doc` target, tweak tweak `install-bin`

This contains more of PR #4492


Makefile.rules
```diff
@@ -155,46 +162,49 @@ SOURCES_ALL += src/compstat_empty.c
...
-# Finally add user provided flags
-GAP_CPPFLAGS += $(CPPFLAGS)
-GAP_PKG_CPPFLAGS += $(CPPFLAGS)
+# Add user provided flags to common flags
+GAP_COMMON_CPPFLAGS += $(CPPFLAGS)
+
+# Add common flags
+GAP_CPPFLAGS += $(GAP_COMMON_CPPFLAGS)
+GAP_PKG_CPPFLAGS += $(GAP_COMMON_CPPFLAGS)
+GAP_INSTALL_CPPFLAGS += $(GAP_COMMON_CPPFLAGS)
```
fingolfin, line 207: Just to explain the motivation: we split the CPPFLAGS (and other flags) into three:
- one for use of compiling GAP "in place", i.e. used by `make`
- one (`GAP_PKG_CPPFLAGS`) for putting in `sysinfo.gap` right away, used by packages (this may include e.g. fewer `-I` directives, as packages don't need to be exposed to e.g. the flags for accessing the GNU readline header or GMP etc.)
- one (`GAP_INSTALL_CPPFLAGS`) for the installed version of `sysinfo.gap`, mostly a variant of `GAP_PKG_CPPFLAGS` but with paths adjusted to the installed location.

Also, the same must be done for `GAP_LDFLAGS` and perhaps more

I should probably explain this in a comment.

fingolfin:
The main part left to do is installation of `gac` and ensuring that after installation of GAP one can still build kernel extensions. This is complicated by our use of GNU libtool (see also issue #4682). I am currently working on all that, but can't make any promises (this week I managed to work maybe 1-2 hours at it, at night...)

orlitzky:
> The main part left to do is installation of `gac` and ensuring that after installation of GAP one can still build kernel extensions. This is complicated by our use of GNU libtool (see also issue #4682). I am currently working on all that, but can't make any promises (this week I managed to work maybe 1-2 hours at it, at night...)

No problem, there's no rush. My main intent was to ping two people who might know more about this since the problem is already partially solved in SageMath and in the sage-on-gentoo ebuild.

dimpase:
> the problem is already partially solved in SageMath and in the sage-on-gentoo ebuild.

In Debian 11+, too, I think.

fingolfin:
Sure, and I've helped multiple maintainers of GAP distro packages to "solve" it. But they either can solve it because they already have a `libtool` package that reliably install a libtool binary in a fixed location (something I can't rely on in general), or else they have a broken `gac` tool, or a combination.

kiwifb:
Fixing `gac` was so much fun - not - but you are right about libtool. I, at least, still rely on installing some part of gap libtool and making sure it is findable. https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/gap-core/files/gap-core-4.11.1-gac.patch
But I also have quite the fun with the packages that use `Makefile.gappkg` and some of my poking in gac is also to simplify when I use it. Because of the way gap is used in sage we usually need the module to link to libgap rather than relying on the loader to provide it. The patch above includes provision for that.

PR #4733, opened 2022-01-11
Fix minor quirks and inconveniences

### LINKED ISSUE #4649: Naming of groups is inconsistent
Consider and compare these:
```
gap> G:=CyclicGroup(2);;
gap> StructureDescription(G);
"C2"
gap> DisplayCompositionSeries(G);
G (1 gens, size 2)
 | Z(2)
1 (0 gens, size 1)
gap> Z(2);  # actually the user might think `Z(2)` is a valid command... which it is, but not in this way:
Z(2)^0
```
Of course there are good reasons for all this, but a user (esp. one who is not a group theorist) may be quite confused by this.

[The manual entry for`DisplayCompositionSeries`](https://www.gap-system.org/Manuals/doc/ref/chap39.html#X82C0D0217ACB2042) does not specify which names it prints. Looking at it, it uses `IsomorphismTypeInfoFiniteSimpleGroup`, whose [manual entry](https://www.gap-system.org/Manuals/doc/ref/chap39.html#X7C6AA6897C4409AC) explains the notation -- perhaps the manual should make this explicit, by adding a reference?

Furthermore, perhaps [`StructureDescription`](https://www.gap-system.org/Manuals/doc/ref/chap39.html#X8199B74B84446971) could reference `IsomorphismTypeInfoFiniteSimpleGroup` and explain that it uses different naming? (and vice versa)

CC @fieker

hulpke:
Fixed some of the issues reported by B. Sambale in an email on Jan-11-22.
Also resolve  #4649

## Text for release notes

None

## Further details

### DIFF
--- a/lib/fitfree.gi
+++ b/lib/fitfree.gi
@@ -1398,6 +1398,36 @@ local l;
   fi;
 end);
 
+
+#############################################################################
+##
+#M  SylowSubgroupOp( <G>, <pi> )
+##
+## Fitting free approach
+##
+InstallMethod( SylowSubgroupOp, "fitting free",true,
+  [ IsGroup and CanComputeFittingFree,IsPosInt ],
+  -1, # deliberate lower ranking to amke sure this method only runs in cases
+  # in which no more specialized method is installed. Once the method has
+  # been used more broadly, and performance is better understood, this can
+  # be changed to 0
+function(G,pi)
+local l;
+  if Set(Factors(Size(G)))=[pi] then 
+    SetIsPGroup(G,true);
+    SetPrimePGroup(G, pi);
+    return G;
+  fi;
+  l:=HallViaRadical(G,[pi]);
+  if Length(l)=1 then
+    SetIsPGroup(l[1],true);
+    SetPrimePGroup(l[1],pi);
+    return l[1];
+  else
+    Error("There can be only one class");
+  fi;
+end);
+
 InstallMethod(ChiefSeriesTF,"fitting free",true,
   [IsGroup and CanComputeFittingFree ],0,
 function(G)
--- a/lib/ghomfp.gi
+++ b/lib/ghomfp.gi
@@ -245,6 +245,11 @@ InstallMethod( ImagesRepresentative, "map from (sub)fp group, rewrite",
     IsMultiplicativeElementWithInverse ], 0,
 function( hom, word )
 local aug,si,r,i,j,tt,ct,cft,c,f,g,ind,e,eval;
+  # catch trivial group
+  if HasMappingGeneratorsImages(hom) 
+    and Length(MappingGeneratorsImages(hom)[1])=0 then
+    return One(Range(hom));
+  fi;
   # get a coset table
   aug:=CosetTableFpHom(hom);
   r:=One(Range(hom));
--- a/lib/norad.gi
+++ b/lib/norad.gi
@@ -1050,3 +1050,15 @@ local sus,ser,len,factorhom,uf,n,d,up,mran,nran,mpcgs,pcgs,pcisom,nf,ng,np,sub,
   
 end);
 
+InstallMethod( NormalizerOp,"solvable radical", IsIdenticalObj,
+  [ IsGroup and CanComputeFittingFree, IsGroup],
+  -1, # deliberate lower ranking to amke sure this method only runs in cases
+  # in which no more specialized method is installed. Once the method has
+  # been used more broadly, and performance is better understood, this can
+  # be changed to 0
+function(G,U)
+  # small pc groups fall back on generic -- don't trigger here!
+  if IsPcGroup(G) then TryNextMethod();fi;
+  return NormalizerViaRadical(G,U);
+end);
+
--- a/lib/clashom.gi
+++ b/lib/clashom.gi
@@ -2998,6 +2998,7 @@ InstallMethod( CentralizerOp, "TF method:elm",IsCollsElms,
   [ IsGroup and IsFinite and HasFittingFreeLiftSetup,
   IsMultiplicativeElementWithInverse ], OVERRIDENICE,
 function( G, e )
+local ffs;
   if IsPermGroup(G) or IsPcGroup(G) then TryNextMethod();fi;
   if not e in G then 
     # TODO: form larger group containing e.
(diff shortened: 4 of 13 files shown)


lib/fitfree.gi
```diff
@@ -1398,6 +1398,31 @@ local l;
...
+##
+## Fitting free approach
+##
+InstallMethod( SylowSubgroupOp, "fitting free",true,
+    [ IsGroup and CanComputeFittingFree,IsPosInt ],
+    -1, # rank lower than anything that would imply CanComputeFittingFree
+function(G,pi)
+local l;
+  if Set(Factors(Size(G)))=[pi] then 
+    SetIsPGroup(G,true);
```
fingolfin, line 1414: Ideally the prime should also be set (some code gets confused if it isn't, or at least it used to be that way)
```suggestion
    SetIsPGroup(G,true);
    SetPrimePGroup(G, pi);
```

lib/fitfree.gi
```diff
@@ -1398,6 +1398,31 @@ local l;
...
+    return G;
+  fi;
+  l:=HallViaRadical(G,[pi]);
+  SetIsPGroup(l,true);
```
fingolfin, line 1418:
```suggestion
  SetIsPGroup(l,true);
  SetPrimePGroup(G, pi);
```

lib/fitfree.gi
```diff
@@ -1398,6 +1398,31 @@ local l;
...
+#M  SylowSubgroupOp( <G>, <pi> )
+##
+## Fitting free approach
+##
+InstallMethod( SylowSubgroupOp, "fitting free",true,
+    [ IsGroup and CanComputeFittingFree,IsPosInt ],
+    -1, # rank lower than anything that would imply CanComputeFittingFree
+function(G,pi)
+local l;
+  if Set(Factors(Size(G)))=[pi] then
```
fingolfin, line 1413: Optional suggestion: could also do this:
```suggestion
  if PrimeDivisors(Size(G))=[pi] then 
```

lib/fitfree.gi
```diff
@@ -1398,6 +1398,31 @@ local l;
...
+
+#############################################################################
+##
+#M  SylowSubgroupOp( <G>, <pi> )
+##
+## Fitting free approach
+##
+InstallMethod( SylowSubgroupOp, "fitting free",true,
+    [ IsGroup and CanComputeFittingFree,IsPosInt ],
+    -1, # rank lower than anything that would imply CanComputeFittingFree
```
fingolfin, line 1410: Why -1 here? And why -5 for the Normalizer method? Are you trying to rank below specific method? Or just undoing the upranking caused by CanComputeFittingFree? In my GAP install, `RankFilter(IsGroup)` is 35 and `RankFilter(CanComputeFittingFree)` equals `RankFilter(IsGroup and CanComputeFittingFree)` equals 52.

hulpke: -1 to guarantee that it will be lower than the permutation group method (probably is anyhow)

With `Normalizer`, it turns out that the pc group method can dispatch to the generic case. I had thought initially it would just be a ranking issue (thus the -5), but in the end this made it necessary to explicitly exclude the pc group case.

lib/fitfree.gi
hulpke: Yes, one could. But neither runtime cost, nor size of code makes a big difference.

lib/fitfree.gi
fingolfin: OK, but then the comment should reflect that, otherwise when we run into problems with this in the future, it will be hard to  understand why this downranking happens.

Though it is indeed unnecessary right now:
```
gap> RankFilter(CanComputeFittingFree);
52
gap> RankFilter(IsPermGroup);
65
```
But of course strictly speaking it also isn't sufficient. as this ranking order could in principle reverse. That said, this seems highly unlikely, as it doesn't seem plausible that `CanComputeFittingFree` shouldn't obtain implications (there isn't really anything you can deduce from this filter, is there?).

So perhaps just drop that line?

lib/ghomfp.gi
```diff
@@ -245,6 +245,11 @@ InstallMethod( ImagesRepresentative, "map from (sub)fp group, rewrite",
     IsMultiplicativeElementWithInverse ], 0,
 function( hom, word )
 local aug,si,r,i,j,tt,ct,cft,c,f,g,ind,e,eval;
+  # catch trivial group
+  if HasMappingGeneratorsImages(hom)
```
ThomasBreuer, line 249: In the situation of #4737, the `MappingGeneratorsImages` value is indeed known, and we do not run into the problem anymore after the proposed changed.
My question is: Are we sure that `MappingGeneratorsImages` is set whenever the group in question is trivial?
(Another formulation would be: Are we sure that `CosetTableFpHom` is never called in the case of trivial groups, since its code assumes this?)

fingolfin: Also, about the first thing `CosetTableFpHom` does is to invoke `MappingGeneratorsImages`, so perhaps we don't need to check `HasMappingGeneratorsImages` here?

lib/norad.gi
```diff
@@ -1050,3 +1050,12 @@ local sus,ser,len,factorhom,uf,n,d,up,mran,nran,mpcgs,pcgs,pcisom,nf,ng,np,sub,
   
 end);
 
+InstallMethod( NormalizerOp,"solvable radical", IsIdenticalObj,
+  [ IsGroup and CanComputeFittingFree, IsGroup],
+  -5, # rank lower than methods for reprersentations that imply CCFF
```
fingolfin, line 1055: So can we remove this?
```suggestion
```

lib/ghomfp.gi
hulpke: The method is installed for `IsFromFpGroupGeneralMappingByImages and IsGroupGeneralMappingByImages`. So `MappingGeneratorsImages` is known. And the error came from accessing the first generator in the `MappingGeneratorsImages` list. So I think this is a perfectly plausible workaround for the concrete error reported.

If your question is more generally: Does every documented operation work OK with a trivial finitely presented group, the answer is probably "no" (and could be another boring issue), and I make no claim to have fixed every case of that.

lib/norad.gi
hulpke: If it makes you happier we can rank at -1. But giv en the prominence of `Normalizer` I want to make sure the method currently ends up only as fallback if no better method is known.

Another option would be to rank one higher than the generic finite group routine, but that seemed harder.

lib/fitfree.gi
hulpke: > Though it is indeed unnecessary right now:

> But of course strictly speaking it also isn't sufficient. as this ranking order could in principle reverse. That said, this seems highly unlikely, as it doesn't seem plausible that `CanComputeFittingFree` shouldn't obtain implications (there isn't really anything you can deduce from this filter, is there?).
> 
> So perhaps just drop that line?

In the long run.

`Centralizer` (as well as `Normalizer` below) are prominent operations, and I want to make sure this ``cross representation`` installation does not suddenly break things by overriding existing methods due to surprising implications.
I have not tried every representation of groups that currently exist.

Ranking lower seems safer.

fingolfin (approved):

PR #4736, opened 2022-01-15
Adds primitivity check for SocleTypePrimitiveGroup

### LINKED ISSUE #4735: SocleTypePrimitiveGroup returns wrong result for non-primitive groups
### Observed behaviour
```
G := DirectProduct( AlternatingGroup(5), AlternatingGroup(6));; SocleTypePrimitiveGroup(G);
rec( name := "A(6) ~ A(1,9) = L(2,9) ~ B(1,9) = O(3,9) ~ C(1,9) = S(2,9) ~ 2A(1,9) = U(2,9)", parameter := 6, series := "A", width := 2 )
```
### Expected behaviour
Clearly the socle of `G` is a direct product of `A_5` and `A_6`. The socle of a non-primitive group is generally not isomorphic to `T^k` for a non-abelian finite simple group `T` so this function should return an error as `G` does not primitively.

DominikBernhardt:
Fixes #4735 .

This function could also return `fail`, I'm not quite sure. Maybe somebody has suggestions? 
Also, if we return an error, should I write a test and if so, what would be the best way?

### DIFF
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -2291,6 +2291,9 @@ end);
 InstallMethod(SocleTypePrimitiveGroup,"primitive permgroups",true,
   [IsPermGroup],0,function(G)
 local s,cs,t,id,r;
+  if not IsPrimitive(G) then 
+    ErrorNoReturn("<G> must be primitive on its moved points");
+  fi;
   s:=Socle(G);
   cs:=CompositionSeries(s);
   t:=cs[Length(cs)-1];


hulpke (approved):
Obviously fine -- the test adds little cost.

Generically I have found that I often wished that functions would return `fail` instead of an error, but I'm not sure whther this is needed here.

DominikBernhardt:
I'm also tempted to have this function return fail since asking for the socle type of a non primitive group doesn't make sense, but I'm not sure what the GAP policy on this is. Maybe @fingolfin could help. 😊

hulpke:
>but I'm not sure what the GAP policy on this is.

There is no official policy. Initially almost all functions gave errors. Over time we have changed it in some cases to `fail`, as algorithms otherwise would have to do the test a second time before calling the routine.

In the end the question is how the function is called -- within algorithms would get `fail`, while mostly top level user would be error messages. E.g. returning `fail` when multiplying elements from two different pc groups is likely to be more confusing than an error message, and there is no situation in an algorithm where one would like to get `fail`.

DominikBernhardt:
From the packages distributed with GAP, a search using `grep` yields that only the _PrimGrp_ package and _JupyterKernel_ use `SocleTypePrimitiveGroup`. It is also used occasionally in files in `lib/`, so I'm still torn on this. Probably and error message is better then?

lib/grpperm.gi
```diff
@@ -2291,6 +2291,9 @@ end);
 InstallMethod(SocleTypePrimitiveGroup,"primitive permgroups",true,
   [IsPermGroup],0,function(G)
 local s,cs,t,id,r;
+  if not IsPrimitive(G) then 
+    ErrorNoReturn("<G> must be primitive on its moved points.");
```
fingolfin, line 2295:
```suggestion
    ErrorNoReturn("<G> must be primitive on its moved points");
```

fingolfin:
As @ThomasBreuer pointed out on the issue, the documentation for this function states:

> If G does not have a faithful primitive action, the result is undefined

So "legally" we can do whatever we want, and in fact even the current code is "correct". But I do indeed think we should catch it, and if checking for primitivity indeed adds negligible cost, then here I'd just do exactly what this patch does, i.e. call `Error`. There is no point in returning `fail` because code which wants to handle the situation itself can easily do so by invoking `IsPrimitive` before `SocleTypePrimitiveGroup` (at no extra cost, as we cache this property). And existing code will at worst stop working with an "obvious" way to fix it (unless it was broken to start with). On the other hand, code which expects only `true` or `false` can sometimes accidentally keep running if it gets `fail` instead, and lead to wrong results later on.

BTW, `JupyterKernel` almost definitely does not use this, it only has a long list of function names defined in the GAP library to facilitate code highlighting.

fingolfin (approved):

PR #4746, opened 2022-02-02
WIP: change gac to not use libtool

fingolfin:
This is incomplete work in progress, but I wanted to post it here
just for reference. I'll hopefully be able to resume work on this
in the not so distant future.

For now, things to be done include adding a full test of MockPkg with
its new kernel extension (this should be completed and merged *before* this PR,
so that we can verify things keep working).


fingolfin:
Compiling NormalizInterface, we get:
```
gcc -m32 -m32 --coverage -L/home/runner/work/gap/gap -shared -m32 -lstdc++ -R/home/runner/work/gap/gap/pkg/NormalizInterface-1.3.0/NormalizInstallDir/lib -L/home/runner/work/gap/gap/pkg/NormalizInterface-1.3.0/NormalizInstallDir/lib -lnormaliz -o bin/x86_64-pc-linux-gnu-default32-kv8/NormalizInterface.so gen/src/normaliz.lo
gcc: error: unrecognized command line option ‘-R’
```
That's something to be fixed in there, I guess... *sigh*

fingolfin:
Failed to load profiling; perhaps there should also be a test kernel extension using C++...
```
Error, failed to load dynamic module /home/runner/work/gap/gap/pkg/profiling-2.4.1/bi\
n/x86_64-pc-linux-gnu-default64-kv8/profiling.so, /home/runner/work/gap/gap/pk\
g/profiling-2.4.1/bin/x86_64-pc-linux-gnu-default64-kv8/profiling.so: undefine\
d symbol: _ZTVN10__cxxabiv120__si_class_type_infoE in
  LOAD_DYN( filename ) at /home/runner/work/gap/gap/lib/files.gd:590 called from 
<function "LoadDynamicModule">( <arguments> )
 called from read-eval loop at /home/runner/work/gap/gap/pkg/profiling-2.4.1/init.g:15
```

Makefile.rules
```diff
@@ -226,9 +236,51 @@ GAP_CXXFLAGS += $(CXXFLAGS)
...
+  # that person to help test it)
+
+  # also add this, from GAP's LDFLAGS????
+  # -Wl,--stack -Wl,16777216 -Wl,--allow-multiple-definition
+
+# gcc  -shared  src/*.o
+#    -Wl,/home/appveyor/gap/bin/x86_64-pc-cygwin-default64-kv8/gap.dll
+#    -o bin/x86_64-pc-cygwin-default64-kv8/.libs/datastructures.dll
+#    -Wl,--enable-auto-image-base
+#    -Wl,--out-implib,bin/x86_64-pc-cygwin-default64-kv8/.libs/datastructures.dll
```
fingolfin, line 264: I wonder if we need to / ought to add any of these flags; perhaps @ChrisJefferson has some insights?

In either case, we'll have to test the resulting kernel extensions carefully on cygwin. Just because the trivial kext of `mockpkg` works doesn't mean more complex ones will, too...

ChrisJefferson: From GAP's LDFLAGS, the --stack shouldn't be necessary (only the top file needs that). --allow-multiple-definition is often necessary for things compilers get wrong -- I haven't tried without it for a few years so it might be improved, but it also doesn't hurt (it just lets you define the same function multiple times).

ChrisJefferson: Just double checked, --stack only does something when building an executable, it doesn't do anything to DLLs (it doesn't hurt either).

fingolfin: Any idea about the other options, `--enable-auto-image-base` and `--out-implib`? I looked a the manpage explanations for them, but really couldn't tell... I guess the fact that the Windows CI test pass suggest not using them is fine, but just in case there is a hidden pittrap you might by chance be aware of (don't waste time researching it deeply, though, we'll just gotta try it

fingolfin:
Two failures left, both in building packages:
1. `NormalizInterface` dies with the following error; this is already fixed in the latest release of the package (for unrelated reasons)
```
gcc -o bin/x86_64-pc-linux-gnu-default64-kv8/NormalizInterface.so gen/src/normaliz.lo --coverage -shared -lstdc++ -R/home/runner/work/gap/gap/pkg/NormalizInterface-1.3.0/NormalizInstallDir/lib -L/home/runner/work/gap/gap/pkg/NormalizInterface-1.3.0/NormalizInstallDir/lib -lnormaliz
  gcc: error: unrecognized command line option ‘-R’
  Makefile.gappkg:131: recipe for target 'bin/x86_64-pc-linux-gnu-default64-kv8/NormalizInterface.so' failed
```
2. `nq` fails because it relies on getting the compiler/linker flags for accessing GMP from GAP. I'll think about how to best address this...

fingolfin:
`nq` 2.5.8 resolved those issues! Sadly `testpackages` tests still fail because `NormalizInterface` crashes. Hmm. Maybe related to https://github.com/gap-system/PackageDistro/issues/285 ? Anyway, I've next tried to build it using `libnormaliz-dev` instead, but that doesn't work in ubuntu-18.04. So I switched to ubuntu-latest -- which is fine, except that they've dropped 32bit support there *sigh*

fingolfin:
Note to self: `make install` should also install `etc` and `hpcgap/lib`. Some further candidates:
- benchmarks
- hpcgap/demo
- tst

fingolfin:
I'll try and see if I can get this ready for GAP 4.12.0.

Perhaps @ChrisJefferson has an idea why NormalizInterface might crash?

fingolfin:
@ChrisJefferson this is finally ready for review and merging -- with the latest NormalizInterface update, that last blocker has been removed! Also, the patch is now fairly small, as I already got most of the changes merged separately.

ChrisJefferson (approved):

PR #4750, opened 2022-02-04
CI: use `windows-2019` in Cygwin test job

wilfwilson:
In #4745, @ChrisJefferson updated the 'Wrap releases' workflow to use `windows-2019` rather than `windows-latest` (which is currently `windows-2019`, but will still be `windows-2022`).

I approved and merged, but forgot that there was another Windows job, which runs the tests.

@ChrisJefferson should we change this one be `windows-2019` too?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -141,7 +141,7 @@ jobs:
             test-suites: "testinstall"
             extra: "JULIA=yes CONFIGFLAGS=\"--enable-debug --disable-Werror\""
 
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Clearly @ChrisJefferson should comment, but in any case, I don't see harm in doing this?

ChrisJefferson:
Yes, this shouldn't change anything, and avoids surprises in future with things changing behind our back (we can worry about upgrading explicitly, if and when we need to down the line)

PR #4753, opened 2022-02-04
CI: Upload code coverage for Windows build

wilfwilson:
I noticed the presence of a line stopping code coverage from being reported on the Windows build.

Let's remove this line to see if it works again.

This may let us close issue #4640, if it no longer persists.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -251,7 +251,6 @@ jobs:
         run: ${{ matrix.extra }} dev/ci-gather-coverage.sh
       - name: "Upload coverage data to Codecov"
         uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


fingolfin:
The messages are gone from the log, but of course codecov upload was kinda broken. I've restarted the job here now that codecov upload is fixed.

fingolfin:
In any case, https://app.codecov.io/gh/gap-system/gap/compare/4753 shows a substantial increase in coverage of the kernel code (2.74%) -- though it seems to be mostly caused by it reporting lines such as `Obj     f, copy_img, copy_ker;` being reported as "covered", probably due to the compiler being used..

fingolfin (approved):

Issue #4751, opened 2022-02-04
GAP tab-completion of record component names


wilfwilson:
The following occurs for me in the GAP master branch (with readline):

When typing the name of a record, GAP tab-completes the partially-completed variable. And, once the variable name is written and followed by `.`, then GAP tab-completion gives the names of components of the record:
```gap
 ┌───────┐   GAP 4.12dev-1155-g7d8e747 built on 2022-02-04 10:42:35+0000
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin21.3.0-default64-kv8
 Configuration:  gmp 6.2.1, GASMAN, readline
 Loading the library and packages ...
 Packages:   GAPDoc 1.dev, PrimGrp 3.4.1, SmallGrp 1.4.2, TransGrp 3.3
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
gap> var := rec(member := true, data1 := 1, data2 := 2);;
gap> va                 # <DOUBLE TAB>
var
gap> var                # <DOUBLE TAB>
var
gap> var.               # <DOUBLE TAB>
data1   data2   member
```
This is all fine.

However, what GAP does not do, but which I wish it would do, is then tab-complete the partially-written component names. Instead, it draws possible completions from a much wider list of variable names:
```gap
gap> var.m              # <DOUBLE TAB>
memory_allocated  mod
gap> var.d              # <DOUBLE TAB> 
d                               denombasis
daleth                          denominator
dashleftarrow                   denominfo
dashrightarrow                  denomstring
dashv                           denpardepths
data                            depth
data1                           depthMap
data2                           depthMapFromParent
datastructures                  depths
ddots                           depthsInParent
dead                            desc
deadbags                        description
deadkb                          design
dec                             detect
decinv                          diag
decmat                          diam
decomp                          diamond
decompinfo                      diamondsuit
decompose                       dictionary
deductions                      diff
deepthought                     difsets
```
I'm not sure where this list comes from, and note that this list does not necessarily include the names of the components of the record.

Ideally, tab-completion would complete `var.m` to `var.member`, and we would have:
```gap
gap> var.d              # <DOUBLE TAB>
data1   data2
```

(This issue was first raised in https://github.com/peal/vole/issues/34).

fingolfin:
[...]

PR #4754, opened 2022-02-05
Fix tab completion of partial names of record members

ChrisJefferson:
(the following is all in the commit message, as I wanted to describe what happened here)

Given a variable like "abc := rec(grape := 2);", we expect if we
type "abc.gr<tab>", then the single completion "grape" will be
suggested. Instead, before this commit we are suggested all global
variables which begin 'gr'.

Note that "abc.<tab>" did correctly suggest "grape", so only partial
completing was broken.

This was caused by an off-by-one error, where when steping through
the record members, we try finding the last member (which doesn't
exist).

This can't (currently) be sensibly tested.

Fixes github issue #4751
(more text follows)

This can't be tested (as far as I know) -- we do have some tests that test input/output, but I tried putting tabs into them and found that eeadline is too clever, and realises we are redirecting a file in so doesn't do tab completion.

I have tested, as much as I can, I haven't broken anything involving tab-completing records, but I'd be happy if at least one other people tried this PR and made sure everything seemed good (@wilfwilson ?)

### DIFF
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -964,7 +964,7 @@ GAPInfo.CommandLineEditFunctions.Functions.Completion := function(l)
       r := fail;
       if Length(cmps) > 0 and cmps[1] in idbnd then
         r := ValueGlobal(cmps[1]);
-        for j in [2..Length(cmps)] do
+        for j in [2..Length(cmps)-1] do
           if not hasbang[j-1] and IsBound(r.(cmps[j])) then
             r := r.(cmps[j]);
           elif hasbang[j-1] and IsBound(r!.(cmps[j])) then


wilfwilson:
> Instead, before this commit we are suggested all global variables which begin 'gr'

I think it was actually 'worse' than this (unless I misunderstand what global variables are) – see e.g. my example in issue #4751, where the tab-autocompletion of `var.d` suggested:
```gap
gap> var.d              # <DOUBLE TAB> 
d                               denombasis
daleth                          denominator
dashleftarrow                   denominfo
dashrightarrow                  denomstring
dashv                           denpardepths
data                            depth
data1                           depthMap
data2                           depthMapFromParent
datastructures                  depths
ddots                           depthsInParent
dead                            desc
deadbags                        description
deadkb                          design
dec                             detect
decinv                          diag
decmat                          diam
decomp                          diamond
decompinfo                      diamondsuit
decompose                       dictionary
deductions                      diff
deepthought                     difsets
[...more stuff omitted...]
```
I think that most of these are the names of other record component names.

wilfwilson:
Things are better than they were, but not perfect.
```gap
gap> var := rec(member1 := true, data1 := 1, data2 := 2);;
gap> new := rec(member2 := false, data3 := 1, data4 := 2);;
gap> var.  # PRESS <DOUBLE TAB>: it suggests the members of var
data1    data2    member1
gap> var.d  # PRESS <TAB> once: it autocompletes to var.data
gap> var.data  # PRESS <TAB> again: it suggests the completions that are members of var
data1  data2
```
So far, so good, but from this position:
```gap
gap> var.data  # PRESS <DOUBLE TAB> *again*:
data            data2           data4
data1           data3           datastructures
```
And we're again getting suggestions that (in my opinion) are inappropriate.

fingolfin:
Note that while we have no setup in place to test this right now, we really should have some. There are of course tools to deal with this, the most prominent being `expect` (but there is also `pexpect` in Python and many more).

fingolfin:
I don't think the analysis is quite right; the line in question has existed like this for years, and it worked well before my commit 5a852ca7f, which introduced this regression.

fingolfin:
OK, I should qualify this: perhaps this *is* the correct fix, but perhaps really the fix needs to be applied to how cmps is computed, which I changed in that commit.

@wilfwilson I believe what you describe in terms on how this PR works for you matches what GAP 4.11.1 and before did... but perhaps you can verify in case I am misunderstanding

fingolfin:
There are definitely more issues introduced by my PR back then. E.g. to continue your example, if you enter `func(var.d` and the press TAB in the old version it autocompleted to `func(var.data` and now it just does generic completion. This is because it interprets the `func(var` together as an identifier. It is in principle clear to me how to fix that; but the core message here to me is that it really REALLY would be better to have a way to regression test this code....

wilfwilson:
> @wilfwilson I believe what you describe in terms on how this PR works for you matches what GAP 4.11.1 and before did... but perhaps you can verify in case I am misunderstanding

You are correct @fingolfin.

PR #4757, opened 2022-02-05
Fix regression in the tab completion code

fingolfin:
This restores tab completion of record components to its
previous glory:

    gap> var := rec(member := true, data1 := 1, data2 := 2);;
    gap> var.d              # <TAB> completes to the next line:
    gap> var.data           # <TAB> again offers these completions:
    data1  data2
    gap> var.data

Also fixes this:

    gap> func(var.d         # <TAB>
    gap> func(var.data      # <TAB>
    data1  data2
    gap> func(var.data

Resolves #4751

Alternative to PR #4754 and hence closes #4754

PR #4758, opened 2022-02-05
CI: use `expect` to test TAB completion

fingolfin:
This contains PR #4757. It is a minimal proof of concept: we certainly should test more things in the tab completion.


fingolfin:
For some reason CI is not running for this PR and I don't understand why?!

wilfwilson (approved):

PR #4759, opened 2022-02-08
Knuth-Bendix, Earns and some code cleanup

hulpke:
Recent improvements:

- The Knuth-Bendix uses a DAG for identifying which rule could apply at a given position. This cleans up other ad-hoc improvements, resulting in shorter and cleaner code
- The requested fix for `Earns` to agree with its definition.
- `SubgroupConditionAbove` works for arbitrary groups, not just permutation groups
- `LowLayerSubgroups` uses orbit in favor of pairwise conjugacy, if this is a plausible approach.
- Info statement in Dixon-Schneider prints more compact information.
-  Clean up code duplication in compuattion of rewriting system for simple groups

## Text for release notes

As promised in the manual, Earns now always returns a list.

### DIFF
--- a/doc/ref/ctbl.xml
+++ b/doc/ref/ctbl.xml
@@ -492,7 +492,7 @@ yielding smaller spaces and several irreducible characters.
 <Example><![CDATA[
 gap> DixonSplit( d );
 #I  Matrix 2,Representative of Order 3,Centralizer: 5832
-#I  Dimensions: [ 1, 2, 1, 4, 12, 1, 1, 2, 1, 2, 1 ]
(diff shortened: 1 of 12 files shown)


fingolfin (approved):

wilfwilson (commented):
The fix for `Earns` is not complete. An empty list is now given instead of `fail` (which is good), but the other case still returns an actual group, rather than a list containing one group (which is what the documentation suggests).
```
gap> Earns(PrimitiveGroup(8, 3));
Group([ (1,7)(2,8)(3,5)(4,6), (1,3)(2,4)(5,7)(6,8), (1,2)(3,4)(5,6)(7,8) ])
```

hulpke:
> Earns(PrimitiveGroup(8, 3));

Thank you @wilfwilson -- did not realize that was part of the issue. Will fix.

PR #4760, opened 2022-02-09
changed a generic `OneSameMutability` method

ThomasBreuer:
Up to now, the lowest rank method for `OneSameMutability` called `OneMutable` and then `MakeImmutable` for an immutable argument.
This approach runs into a "no method found" error if the argument is a domain.
```
gap> OneSameMutability( GF(2) );
Error, no 1st choice method found for `OneMutable' on 1 arguments
[...]
```
Note that the analogous situation for `ZeroImmutable` was different:
Here the lowest rank method calls `ZeroAttr` for an immutable argument.
(Calling `MakeImmutable` on the result seems to be superfluous.)
```
gap> ZeroSameMutability( GF(2) );
0*Z(2)
```
This asymmetry was observed in the context of the GAP-Oscar integration:
We had installed a method for the Oscar function `one` that delegates to GAP's `OneSameMutability` in the case that the argument is a GAP object, and analogously a `zero` method that delegates to `ZeroSameMutability`.
The latter works also for GAP domains, the former does not.

### DIFF
--- a/lib/arith.gi
+++ b/lib/arith.gi
@@ -88,17 +88,21 @@ InstallOtherMethod( Zero,
 #T which takes less than 1 microsecond on my system.
 #T         Steve
 
+
+#############################################################################
+##
+#M  ZeroSameMutability( <obj> ) . . . . . . . . . . for an (immutable) object
+##
+##  This method is applicable for example to domains.
+##
 InstallOtherMethod( ZeroSameMutability,
     "for an (immutable) object",
     [ IsObject ],
-    function(elm)
-    local z;
-    if IsMutable( elm ) then
+    function( obj )
+    if IsMutable( obj ) then
       TryNextMethod();
     fi;
-    z:= ZeroAttr( elm );
-    MakeImmutable( z );
-    return z;
+    return ZeroAttr( obj );
     end );
 
 
@@ -295,17 +299,21 @@ InstallMethod( OneOp,
     return one;
     end );
 
+
+#############################################################################
+##
+#M  OneSameMutability( <obj> )  . . . . . . . . . . for an (immutable) object
+##
+##  This method is applicable for example to domains.
+##
 InstallOtherMethod( OneSameMutability,
     "for an (immutable) object",
     [ IsObject ],
-    function( elm )
-    local o;
-    if IsMutable( elm ) then
+    function( obj )
+    if IsMutable( obj ) then
       TryNextMethod();
     fi;
-    o:= OneMutable( elm );
-    MakeImmutable( o );
-    return o;
+    return OneImmutable( obj );
     end );
 
 
--- a/tst/teststandard/arith.tst
+++ b/tst/teststandard/arith.tst
@@ -1,0 +1,33 @@
+#@local x, D
+gap> START_TEST( "arith.tst" );
+
+# 'OneSameMutability' for mult. elements
+gap> for x in [ Z(2), [ [ 1 ] ], Immutable( [ [ 1 ] ] ) ] do
+>      if IsMutable( OneSameMutability( x ) ) <> IsMutable( x ) then
+>        Error( "result must have same mutability value" );
+>      fi;
+>    od;
+
+# 'OneSameMutability' for domains with one (result shall be immutable)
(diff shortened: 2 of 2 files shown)


fingolfin:
Could we get a test for this? Like, right now, we get:
```
gap> ZeroSM(GF(2));
0*Z(2)
gap> OneSM(GF(2));
```
So perhaps this could be turned into a test?

Perhaps we could even test more systematically... One/Zero/Inverse/AdditiveInverse, each in their three variants, on a bunch of things (were applicable): domains (various fields, rings, groups) and elements... But of course that doesn't have to be in this PR.

ThomasBreuer:
O.k., I will add some tests.

lib/arith.gi
```diff
@@ -295,17 +299,21 @@ InstallMethod( OneOp,
...
 InstallOtherMethod( OneSameMutability,
     "for an (immutable) object",
     [ IsObject ],
-    function( elm )
-    local o;
-    if IsMutable( elm ) then
+    function( obj )
+    if IsMutable( obj ) then
       TryNextMethod();
     fi;
```
fingolfin, line 315: Actually, I also wonder: why do we use `TryNextMethod()` here? The method is already has rank 0. Of course in principle one can have methods with lower rank, but we don't have any AFAIK:
```
gap> for m in MethodsOperation(ZeroSameMutability, 1) do Print(m.rank, ": ", m.info, "\n"); od;
45: ZeroSameMutability: for GF2 Matrix
44: ZeroSameMutability: 8 bit matrix
35: ZeroSameMutability: for null map matrix
25: ZeroSameMutability: for GF2 vector, mutable
24: ZeroSameMutability: for an 8 bit vector
24: ZeroSameMutability: for GF2 vector, immutable
15: ZeroSameMutability: for a plist matrix
14: ZeroSameMutability
11: ZeroSameMutability: for a plist vector
0: ZeroSameMutability: for an (immutable) object
```

And what would such an even lower rank method do, anyway?

So can't we just do this:
```suggestion
    if IsMutable( obj ) then
      ZeroMutable( obj );
    fi;
```

This would be in line with the manual which implies that `ZeroMutable` is the fundamental operation... Well, it says:

> The default method of Zero for additive elements calls ZeroMutable (note that methods for ZeroMutable must not delegate to Zero); so other methods to compute zero elements need to be installed only for ZeroMutable and (in the case of copyable objects) ZeroSameMutability.

I am not quite sure I understand the last bit: why is it necessary/useful "in the case of copyable objects" to also install a method for `ZeroSameMutability` ? Can't we just have the rule as:
- `ZeroSameMutability(x)` by default calls `ZeroMutable` or `ZeroImmutable` (as needed)
- `ZeroImmutable(x)` by default calls `MakeImmutable(ZeroMutable(x))`

Then any other methods would only be installed "as an optimization".

The main problem with that is reality: Plenty of package (and possibly also library?) are installing methods for `Zero` (i.e. `ZeroImmutable`) and for `AINV` (i.e, `AdditiveInverseSameMutability`)... but then again, these are already problematic right now, aren't they?

tst/teststandard/arith.tst
```diff
@@ -0,0 +1,33 @@
+#@local x, D
+gap> START_TEST( "arith.tst" );
```
fingolfin, line 2: Why in `teststandard`, though? aren't these pretty fast tests?

lib/arith.gi
ThomasBreuer: Concerning the `TryNextMethod` call: This just expresses that the method claims to deal only with the case of immutable objects; the method need not care about whether it is the last one, since the "no method found" error will catch this case automatically.

Concerning the documentation: I agree that the sentence in question is confusing; not copyable objects do not need a `ZeroMutable` method.

Concerning the question about general delegation rules: Currently we do not have these rules. The first step would be to check which methods exist, in order to find out whether they cause problems with an intended general rule.

(By the way, one could regard the methods which are changed in this pull request as problematic, since they make no assumption on their arguments. There are generic methods for `IsMultiplicativeElementWithOne`and `IsAdditiveElementWithZero`, respectively. We have (mis)used the methods for `IsObject` for delegations from Julia calls to GAP, but here it would have been sufficient to have methods for `IsDomain`.)

PR #4761, opened 2022-02-09
kernel: rename several arithmetic functions for clarity

fingolfin:
This rectifies a problem I've first run into several years ago: the kernel functions for basic arithmetic function (one, zero, additive & multiplicative inverse) had inconsistent names. This PR works towards rectifying this: it renames all functions which preserve mutability to follow a consistent naming scheme. Specifically:

- lib: avoid AINV(_MUT), ZERO(_MUT), instead use high-level names
- kernel: {INV,ONE}_MUT -> {INV,ONE}_SAMEMUT
- kernel: ZERO -> ZERO_SAMEMUT
- kernel: AINV -> AINV_SAMEMUT

I've grepped through all packages (both in the latest distribution, and also many git clones I have around) to see if this breaks any packages, but to the best of my ability, this is not the case, with one exception: the `Modules` package from homalg installs a method for `ZERO`, see https://github.com/homalg-project/homalg_project/pull/462.

Thus we may want to retain `ZERO` as synonym for `ZERO_SAMEMUT`, at least for a while ?

Ultimately, I'd also like to rename `ONE -> ONE_MUT` and `INV -> INV_MUT`. But note that `ONE_MUT` and `INV_MUT` were already in use before this PR. Moreover, several packages use `INV` or `ONE`... So I'd rather do this in a separate PR, after the dust on this one has settled...

### DIFF
--- a/hpcgap/lib/vec8bit.gi
+++ b/hpcgap/lib/vec8bit.gi
@@ -441,7 +441,7 @@ InstallMethod( ZeroOp, "for an 8 bit vector",
 
 #############################################################################
 ##
-#M  ZEROOp( <vec> )
+#M  ZeroSameMutability( <vec> )
 ##
 ##  A  zero vector of the same field and length and mutability
 ##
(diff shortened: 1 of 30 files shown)


ThomasBreuer (approved):
Looks good. Thanks.

fingolfin:
I really had to add those backwards compatibility mappings, far too many packages rely on those :-(.

Issue #4763, opened 2022-02-11
Regression of 6208432d336e6806ef733e2b6539ae1a416690df regarding group characters


zickgraf:
Execute
```
ForAny( Irr( DihedralGroup( 8 ) ), IsOne );
```
with 6208432d336e6806ef733e2b6539ae1a416690df.

### Observed behaviour

`false`

### Expected behaviour

`true`

### Copy and paste GAP banner (to tell us about your setup)

```
 ┌───────┐   GAP 4.12dev-1165-g6208432 built on 2022-02-11 08:48:25+0100
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv8
 Configuration:  gmp 6.2.1, GASMAN, readline
 Loading the library and packages ...
 Packages:   GAPDoc 1.dev, IO 4.7.0dev, PrimGrp 3.4.0, SmallGrp 1.4.1, TransGrp 2.0.5
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
```

fingolfin:
@ThomasBreuer any ideas on this?

ThomasBreuer:
@fingolfin, in the master branch we get
```
gap> x:= TrivialCharacter( SmallGroup( 1, 1 ) );
Character( CharacterTable( <pc group of size 1 with 0 generators> ), [ 1 ] )
gap> x^0;
1
```
but in GAP 4.11 we have
```
gap> x:= TrivialCharacter( SmallGroup( 1, 1 ) );
Character( CharacterTable( <pc group of size 1 with 0 generators> ), [ 1 ] )
gap> x^0;
Character( CharacterTable( <pc group of size 1 with 0 generators> ), [ 1 ] )
```
However, already in GAP 4.11 we get
```
gap> OneImmutable( x );
1
```
Thus there was already a problem before the recent change in #4760 that caused the regression.
I have to think about a solution.

PR #4766, opened 2022-02-14
Don't use One on collections that are not domains

fingolfin:
Motivated by the discussion in issue #4763. This change is orthogonal to other changes being worked on by @ThomasBreuer

### DIFF
--- a/lib/monoid.gi
+++ b/lib/monoid.gi
@@ -69,7 +69,7 @@ InstallMethod( ViewString,
 InstallOtherMethod(MonoidByGenerators, "for a collection",
 [IsCollection],
 function(gens)
-  local M, pos;
+  local M, pos, one;
 
   M := Objectify(NewType(FamilyObj(gens), IsMonoid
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
@fingolfin I think this fails because the changes in `SemigroupByGenerators` in this PR are not equivalent to the original code. The method for `One` of the generators of a partial perm semigroup or collection is given in `lib/pperm.gi:699`

~~~
InstallMethod(OneImmutable, "for a partial perm coll",
[IsPartialPermCollection],
function(x)
  return JoinOfIdempotentPartialPermsNC(List(x, OneImmutable));
end);
~~~
which is fundamentally different to the method that seems to be causing the problems.

I've commented on @ThomasBreuer's PR #4767, that I think the code that this touches should probably be removed in any case, it was a bad idea (of mine) to have ever introduced it.

fingolfin:
Closing in favor of James' PR #4768

PR #4767, opened 2022-02-14
workaround for a `One` method

### LINKED ISSUE #4763: Regression of 6208432d336e6806ef733e2b6539ae1a416690df regarding group characters

ThomasBreuer:
The GAP Reference Manual defines `One` only for an argument in `IsMultiplicativeElementWithOne` or `IsDomain`.
A method in `lib/arith.gi` tries to extend this definition to collections of `IsMultiplicativeElementWithOne` objects that aren`t domains, by delegating to a representative of the collection.

This causes a logical problem if the collection is itself in `IsMultiplicativeElementWithOne`, because of a different meaning of `One`.
Such a situation occurs in the case of a group character, for which `One` shall return not `1` but the trivial character.

It would be good if we could eventually get rid of this method.
For the moment, the method gets changed in such a way that it delegates to the next applicable method if the argument is in `IsMultiplicativeElementWithOne`.
This change fixes #4763.

### DIFF
--- a/lib/arith.gi
+++ b/lib/arith.gi
@@ -13,11 +13,32 @@
 ##
 
 
+#############################################################################
+##
+##  The GAP Reference Manual defines 'One' only for an argument in
+##  'IsMultiplicativeElementWithOne' or 'IsDomain'.
+##  The following method tries to extend this definition to collections of
+##  'IsMultiplicativeElementWithOne' objects that aren't domains,
+##  by delegating to a representative of the collection.
+##
+##  This causes a logical problem if the collection is itself in
+##  'IsMultiplicativeElementWithOne', because of a different meaning.
+##  Such a situation occurs in the case of a group character,
+##  for which 'One' shall return not '1' but the trivial character.
+##
+##  It would be good if we could eventually get rid of this method.
+##
 InstallOtherMethod(One, "for a multiplicative element with one collection", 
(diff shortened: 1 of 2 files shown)


fingolfin:
@james-d-mitchell That method was added by you and I think it is important for Semigroups. Perhaps you could have a quick look at this PR and let us know if you spot any potential troubles for the semigroups code in GAP resp. in the Semigroups package?

Perhaps also @wilfwilson has any insights?

james-d-mitchell:
@fingolfin I just tried this with the version of Semigroups that I'm working on and it doesn't break anything, I'm also not against the change in principle even if it did break something. I confess I have no memory of writing this method, and I can't think what it would be useful for, and would also be in favour of removing it completely. 

Perhaps alongside removing the code that detects whether or not the `One` is a generator of a semigroup (and in that case returning a `Monoid`), this is the code touch in #4766. In retrospect this was a terrible idea (of mine).

PR #4768, opened 2022-02-14
Fix `One` issue

james-d-mitchell:
This PR resolves Issue #4763 and changes the behaviour of `SemigroupByGenerators` to stop it from returning a monoid if the `One` of the generators is one of the generators. This PR builds on #4766 and supersedes #4767.

### DIFF
--- a/doc/ref/pperm.xml
+++ b/doc/ref/pperm.xml
@@ -1551,11 +1551,11 @@ multiplicative neutral element (i.e. an identity element) but not to satisfy
 <Example><![CDATA[
 gap> f := PartialPerm( [ 1, 2, 3, 6, 8, 10 ], [ 2, 6, 7, 9, 1, 5 ] );;
 gap> S := Semigroup(f, One(f));
-<commutative partial perm monoid of rank 9 with 1 generator>
+<partial perm semigroup of rank 9 with 2 generators>
 gap> IsMonoid(S);
-true
+false
 gap> IsPartialPermMonoid(S);
-true]]></Example>
+false]]></Example>
 
 Note that unlike transformation semigroups, the <Ref Attr="One"/> of a partial permutation semigroup must coincide with the multiplicative neutral element, if either exists.<P/>
 
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,4 @@ doc/gapmacrodoc.idx
 *~
 \#*\#
 .\#*
+gen
--- a/doc/ref/reesmat.xml
+++ b/doc/ref/reesmat.xml
@@ -337,7 +337,7 @@ gap> V:=Semigroup(GeneratorsOfSemigroup(U));
 gap> IsReesZeroMatrixSemigroup(V);
 true
 gap> S:=Semigroup(Transformation([1,1]), Transformation([1,2]));
-<commutative transformation monoid of degree 2 with 1 generator>
+<transformation semigroup of degree 2 with 2 generators>
 gap> IsSimpleSemigroup(S);
 false
 gap> mat:=[[0, One(S), 0, One(S)], [One(S), 0, One(S), 0], 
--- a/doc/ref/trans.xml
+++ b/doc/ref/trans.xml
@@ -970,7 +970,7 @@ gap> S := FullTransformationSemigroup( 5 );
 gap> SmallestMovedPoint( S );              
 1
 gap> S := Semigroup( IdentityTransformation );
-<trivial transformation group of degree 0 with 1 generator>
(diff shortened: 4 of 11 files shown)


ThomasBreuer:
Are we sure that removing the method in question from `lib/arith.gi` will not break some users' code?
(I agree that this method is likely to do something that is not intended, but perhaps it is too brave to just remove it.)

fingolfin:
Hi @james-d-mitchell thanks for the PR, wasn't expecting you to spend time on this :-).

But I agree with @ThomasBreuer, I think we need to be a bit careful. E.g. the Semigroups package definitely relies on this method (I just checked), and others might as well. Before removing this, I'd rather first have the ability again to run tests of all distributed packages (to be tackled during GAP Days next week); then we can update packages; and then *maybe* we can think about it for GAP 4.13 -- although that doesn't solve the issue of people possibly having unpublished code relying on this...

james-d-mitchell:
Thanks for your comments, @fingolfin and @ThomasBreuer. I agree this is probably too big a change to make all at once, perhaps the right thing to do is to merge #4767, and to leave this PR open to be merged at some point in the future?

ThomasBreuer:
@james-d-mitchell Thanks. Yes, eventually this PR should be merged, but for the moment merging just #4767 is safer.

doc/ref/pperm.xml
```diff
@@ -1551,11 +1551,11 @@ multiplicative neutral element (i.e. an identity element) but not to satisfy
...
 gap> f := PartialPerm( [ 1, 2, 3, 6, 8, 10 ], [ 2, 6, 7, 9, 1, 5 ] );;
 gap> S := Semigroup(f, One(f));
-<commutative partial perm monoid of rank 9 with 1 generator>
+<partial perm semigroup of rank 9 with 2 generators>
 gap> IsMonoid(S);
-true
+false
 gap> IsPartialPermMonoid(S);
-true]]></Example>
+false]]></Example>
```
wilfwilson, line 1558: The paragraph preceding this example show that there has been a problem for a long time:

> ```Note that it is possible for a partial perm semigroup to have a multiplicative neutral element (i.e. an identity element) but not to satisfy <C>IsPartialPermMonoid</C>. For example,```

So it's good to get this sorted.

I like the general change of `Semigroup` not producing `IsMonoid` objects any more. I'll be happy for this PR to be merged.

I've long been aware that semigroups of partial permutation are weird with respect to `One`: the `One` of a partial perm does not necessarily match the `One` of any inverse monoid to which it belongs. But I've never found a good way of resolving this problem.

fingolfin:
Should this PR also remove these methods?
```gap
InstallMethod(OneImmutable, "for a partial perm coll",
[IsPartialPermCollection],
function(x)
  return JoinOfIdempotentPartialPermsNC(List(x, OneImmutable));
end);

InstallMethod(OneMutable, "for a partial perm coll",
[IsPartialPermCollection], OneImmutable);
```

doc/ref/pperm.xml
wilfwilson: I can no longer make sense of my comment here (at least the first half) so I've hidden it.

fingolfin:
@james-d-mitchell @ThomasBreuer I wonder, is this still something worth pursuing?

ThomasBreuer:
> @james-d-mitchell @ThomasBreuer I wonder, is this still something worth pursuing?

The comment from the workaround that was added in #4767 says:
"It would be good if we could eventually get rid of this method."

I am not aware of a concrete problem because of the method in question.
Perhaps we should deal with this issue after the upcoming GAP release, not immediately before it.

james-d-mitchell:
I think this is ultimately a good idea to pursue, the issue that arises is that you cannot depend on the output of `Semigroup` being a semigroup, sometimes it is a semigroup other times a monoid (which is also a semigroup of course). This makes reasoning about the output more complicated, which is not ideal. The idea was that the output of `Semigroup` should always belong to the category `IsSemigroup` and not something else. 

Having said all of that I agree with @ThomasBreuer this shouldn't be a priority for v4.16

fingolfin:
So 4.16.0 is out now, so perhaps we can think about this some more.

I just realized there is also PR #5155 which seems to be a rehash of the abandoned #4766 -- I may  have opened #5155 because I forgot the backstory, or perhaps deemed it relevant again... ugh *sigh*

james-d-mitchell:
Good idea, would a reasonable start be to rebase this PR, and see what breaks in the CI?

fingolfin:
@james-d-mitchell sounds good to me

Issue #4771, opened 2022-02-16
Performance regression


fingolfin:
In PR #2585 I attempted to rewrite `EvalString` in a "cleaner" fashion. Back then this was rejected because it was reported to slow down things by a factor of three. Today I discussed this approach with @laurentbartholdi (he suggested the same refactoring of the code independently), which lead me to reexamine the performance issue... And I think we have a performance regression here.

So here is my test code, comparing "old" and "new" `EvalString` implementation:
```gap
myeval := function( s )
  local a, f;
  # Prepend 'return' and append ';' to the given string; also add newlines to
  # ensure that if there are syntax errors in the evaluated string, GAP
  # prints only the evaluated string, but not the 'return' and ';' we prepend
  # resp. append.
  a := "return\n";
  Append(a, s);
  Append(a, ";\n");
  f := ReadAsFunction(InputTextString(a));
  return f();
end;;
for i in [1..100000] do EvalString("12+13"); od; time;
for i in [1..100000] do myeval("12+13"); od; time;
```

Some timings done on my MacBok Pro 14" with M1 CPU:
```
# GAP master
gap> for i in [1..100000] do EvalString("12+13"); od; time;
572
gap> for i in [1..100000] do myeval("12+13"); od; time;
552

# GAP 4.11.1
gap> for i in [1..100000] do EvalString("12+13"); od; time;
360
gap> for i in [1..100000] do myeval("12+13"); od; time;
338

# GAP 4.10.2
gap> for i in [1..100000] do EvalString("12+13"); od; time;
128
gap> for i in [1..100000] do myeval("12+13"); od; time;
448

# GAP 4.9.3
gap> for i in [1..100000] do EvalString("12+13"); od; time;
125
gap> for i in [1..100000] do myeval("12+13"); od; time;
327

# GAP 4.8
for i in [1..100000] do EvalString("12+13"); od; time;
83
for i in [1..100000] do myeval("12+13"); od; time;
96
```
So when looking at GAP 4.9 or 4.10, one really sees a factor of 2.5 - 3.5. But in GAP master and in GAP 4.8 both methods are about the same speed -- except GAP 4.8 is faster by a factor of ~5.5...

After some more experimenting, I noticed the following: if I run the second loop multiple times in GAP 4.8, it gets slower, until it stabilizes at about ~350 (which is about the same as in GAP 4.11); even if I force a garbage collection. This is really strange:
```
[...]

PR #4772, opened 2022-02-16
kernel: avoid memset'ing TypInputFile / TypOutputFile

fingolfin:
Before this patch:

    gap> for i in [1..100000] do EvalString("12+13"); od; time;
    380

After this patch:

    gap> for i in [1..100000] do EvalString("12+13"); od; time;
    314

See issue #4771

### DIFF
--- a/src/gap.c
+++ b/src/gap.c
@@ -209,11 +209,11 @@ static Obj FuncSHELL(Obj self,
         outFile = "*stdout*";
     }
 
-    TypOutputFile output = { 0 };
+    TypOutputFile output;
     if (!OpenOutput(&output, outFile, FALSE))
         ErrorQuit("SHELL: can't open outfile %s", (Int)outFile, 0);
 
-    TypInputFile input = { 0 };
+    TypInputFile input;
     if (!OpenInput(&input, inFile)) {
         CloseOutput(&output);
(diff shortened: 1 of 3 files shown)


src/gap.c
```diff
@@ -209,11 +209,11 @@ static Obj FuncSHELL(Obj self,
         outFile = "*stdout*";
     }
 
-    TypOutputFile output = { 0 };
```
ChrisJefferson, line 212: Being super-picky, I'm a tiny bit nervous we might not be setting some important value to 0 (or might not in the future).

Could we (hidden in a macro/function?) do a memset which sets everything except the 'line' buffer (which I see you moved to the end)

fingolfin: I indeed moved the `line` buffer in an earlier version of this PR, for the very reason you mention, but then decided against it for now after I verified that we really init all fields as necessary.

Of course it is possible that we might add a new field in the future and then "forget" to initialize it, but I think there are bigger dangers in there?

Anyway, I'll think about it... What Id' also really like to *hide* that struct definition again, i.e. move it once more from the header to the C file...

fingolfin: Maybe a better approach would be to insert a `memset(output, 0xFF, sizeof(output);` insider an `#ifdef` that enables it only in debug builds?

ChrisJefferson (approved):

PR #4773, opened 2022-02-16
Matrix Objects for matrices of ZmodnzObj, Use in Dixon-Schneider

hulpke:
Added code for matrix objects that represent matrices over (Integers mod m) for m nonprime or >2^16.
By keeping entries as integers and reducing in the arithmetic we gain a significant amount over lists of lists of ZmodnZObj, by avoiding packing/unpacking at each arithmetic step.
Use this in the Dixon-Schneider Algorithm. In examples (e.g. a group of order ~10^9 with ~550 classes) this leads to an overall speedup of 50% or more.

## Text for release notes

The linear algebra part of character table calculations for groups has been sped up significantly.

### DIFF
--- a/lib/ctblgrp.gi
+++ b/lib/ctblgrp.gi
@@ -91,6 +91,7 @@ local D,C,cl,pl;
   SortParallel(cl,pl,ClassComparison);
   D.perm:=PermList(pl);
   D.permlist:=pl;
+  D.invpermlist:=ListPerm(D.perm^-1,Length(D.permlist));
   D.currentInverseClassNo:=0;
 
   D.characterTable:=C;
@@ -103,32 +104,6 @@ local D,C,cl,pl;
 
   return D;
 end);
-
-# #############################################################################
-# ##
-# #F  DxPowerClass(<D>,<cl>,<pow>)  . . . . . . . . . . . . . evaluate powermap
-# ##
-# DxPowerClass := function(D,nu,power)
-#   local p,primes,cl;
-#   cl:=nu;
-#   power:=power mod D.characterTable.orders[cl];
-#   if power=0 then
-#     return 1;
-#   elif power=1 then
-#     return cl;
-#   else
-#     primes:=Factors(power);
-#     for p in primes do
-#       if not IsBound(D.characterTable.powermap[p]) then
-#         return D.ClassElement(D,
-#           Representative(D.classes[nu])^power);
-#       else
-#         cl:=D.characterTable.powermap[p][cl];
-#       fi;
-#     od;
-#     return cl;
-#   fi;
-# end;
 
 #############################################################################
 ##
@@ -270,13 +245,11 @@ end;
 
 #############################################################################
 ##
-#F  DxNiceBasis(v,space)
+#F  DxIsInSpace(v,space)
 ##
 DxIsInSpace := function(v,r)
-  if not IsBound(r.space) then
-    r.space:=VectorSpace(Field(r.base[1]),r.base);
-  fi;
-  return v in r.space;
+local nr,nc,m,i,j;
+  return SolutionMat(Matrix(BaseDomain(r.base[1]),r.base),v)<>fail;
 end;
 
 #############################################################################
@@ -286,9 +259,9 @@ end;
 DxNiceBasis := function(d,r)
 local b;
   if not IsBound(r.niceBasis) then
-    b:=List(r.base,i->i{d.permlist}); # copied and permuted according to order
+    b:=Matrix(BaseDomain(r.base[1]),List(r.base,i->i{d.permlist})); # copied and permuted according to order
     TriangulizeMat(b);
-    b:=List(b,i->Permuted(i,d.perm)); # permuted back
+    b:=List(b,i->i{d.invpermlist}); # permuted back
     r.niceBasis:=Immutable(b);
   fi;
   Assert(1,Length(r.niceBasis)=Length(r.base));
@@ -336,6 +309,7 @@ DxRegisterModularChar := function(D,c)
 local d,p;
   # it may happen,that an irreducible character will be registered twice:
   # 2-dim space,1 Orbit,combinatoric split. Avoid this!
+  if IsPlistRep(c) then c:=Vector(c); fi;
   if not(c in D.modulars) then
     Add(D.modulars,c);
     d:=Int(c[1]);
@@ -412,25 +386,28 @@ local i,pcomp,m,r,D,neue,tm,news,opt;
   od;
 
   if opt<>true then
-    pcomp:=NullspaceMat(D.projectionMat*TransposedMat(D.modulars));
-    for i in [1..Length(D.raeume)] do
-      r:=D.raeume[i];
-      if r.dim = Length(r.base[1]) then
-	# trivial case: Intersection with full space in the beginning
-	r:=rec(base:=pcomp);
-      else
-	r:=rec(base:=SumIntersectionMat(pcomp,r.base)[2]);
-      fi;
-      r.dim:=Length(r.base);
-      # note stabilizer
-      if IsBound(D.raeume[i].stabilizer) then
-	r.stabilizer:=D.raeume[i].stabilizer;
-      fi;
-      if r.dim>0 then
-	DxActiveCols(D,r);
-      fi;
-      D.raeume[i]:=r;
-    od;
+    pcomp:=NullspaceMat(D.projectionMat*TransposedMatImmutable(Matrix(D.field,D.modulars)));
+    if Length(pcomp)>0 then
+      for i in [1..Length(D.raeume)] do
+        r:=D.raeume[i];
+        if r.dim = Length(r.base[1]) then
+          # trivial case: Intersection with full space in the beginning
+          r:=rec(base:=pcomp);
+        else
+          r:=rec(base:=SumIntersectionMat(pcomp,
+            Matrix(BaseDomain(r.base[1]),r.base))[2]);
+        fi;
+        r.dim:=Length(r.base);
+        # note stabilizer
+        if IsBound(D.raeume[i].stabilizer) then
+          r.stabilizer:=D.raeume[i].stabilizer;
+        fi;
+        if r.dim>0 then
+          DxActiveCols(D,r);
+        fi;
+        D.raeume[i]:=r;
+      od;
+    fi;
   fi;
   D.raeume:=Filtered(D.raeume,i->i.dim>0);
 end );
@@ -627,7 +604,7 @@ local ml,nu,ret,r,p,v,alo,ofs,orb,i,j,inv,b;
       alo:=Length(b);
     od;
   od;
-  inv:=b^(-1);
+  inv:=Matrix(BaseDomain(b[1]),b)^(-1);
   for i in ml do
     v:=i*inv;
     for r in [1..Length(D.raeume)] do
@@ -636,7 +613,7 @@ local ml,nu,ret,r,p,v,alo,ofs,orb,i,j,inv,b;
         p[j]:=v[j];
       od;
       p:=p*b;
-      if p<>nu then
+      if not IsZero(p) then
         AddSet(ret,DxLiftCharacter(D,p));
       fi;
     od;
@@ -653,7 +630,7 @@ DxEigenbase := function(M,f)
   local dim,i,k,eigenvalues,base,minpol,bases;
   k:=Length(M);
 
-  minpol:=MinimalPolynomial(f,M);
+  minpol:=MinimalPolynomial(BaseDomain(M),M);
   
   Assert(2,IsDuplicateFree(RootsOfUPol(minpol)));
   eigenvalues:=Set(RootsOfUPol(minpol));
@@ -664,7 +641,6 @@ DxEigenbase := function(M,f)
     if base=[] then
       Error("This can`t happen: Wrong Eigenvalue ???");
     else
-      #TriangulizeMat(base);
       dim:=dim+Length(base);
       Add(bases,base);
     fi;
@@ -705,7 +681,7 @@ InstallGlobalFunction(SplitStep,function(D,bestMat)
   if ForAny(raeume,i->i.dim>1) then
     bestMatCol:=D.requiredCols[bestMat];
     bestMatSplit:=D.splitBases[bestMat];
-    M:= NullMat( k, k, 0 );
+    M:= ZeroMatrix(Integers,k,k);
     Info(InfoCharacterTable,1,"Matrix ",bestMat,",Representative of Order ",
        Order(D.classreps[bestMat]),
        ",Centralizer: ",D.centralizers[bestMat]);
@@ -729,15 +705,21 @@ InstallGlobalFunction(SplitStep,function(D,bestMat)
       dim:=raum.dim;
       # cut out the 'active part' for computation of an eigenbase
       activeCols:=DxActiveCols(D,raum);
-      N:= NullMat( dim, dim, o );
+      N:= ZeroMatrix(f,dim,dim);
       for row in [1..dim] do
-        Row:=base[row]*M;
+        #Row:=base[row]*M;
+        Row:=CompatibleVector(M);
+        for col in [1..Length(Row)] do
+          Row[col]:=base[row,col];
+        od;
+        Row:=Row*M;
         for col in [1..dim] do
-          N[row][col]:=Row[activeCols[col]];
+          N[row,col]:=Row[activeCols[col]];
         od;
       od;
       eigen:=DxEigenbase(N,f);
       # Base umrechnen
+      base:=Matrix(BaseDomain(base[1]),base);
       eigenbase:=List(eigen.base,i->List(i,j->j*base));
     #eigenvalues:=List(eigen.values,i->i/D.classiz[bestMat]);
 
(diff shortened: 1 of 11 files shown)


lib/matobj.gi
```diff
@@ -293,6 +297,10 @@ InstallMethod( Matrix,
 
 InstallMethod( Matrix,
     [ IsSemiring, IsList ],
+# rank higher than a method in the semigroups package, which otherwise jumps
+# in and causes an error when testing
+# line 318 of semigroups-3.4.0/gap/elements/semiringmat.gi
+20,
```
fingolfin, line 303: I think we'll have to sort it out together with the Semigroups folks, though, don't want to break that package (paging @james-d-mitchell).

lib/matrix.gi
```diff
@@ -2440,9 +2456,16 @@ InstallMethod( SemiEchelonMatTransformation,
...
+function( mat )
+  local copymat;
+  copymat:=MutableCopyMatrix(mat);
+  return SemiEchelonMatTransformationDestructive( copymat );
+end);
+
+InstallOtherMethod( SemiEchelonMatTransformationDestructive,
     "generic method for matrices",
-    [ IsMatrix and IsMutable],
+    [ IsMatrixOrMatrixObj and IsMutable],
```
fingolfin, line 2468: This is not appropriate: this method requires `mat` to be in filter `IsRowListMatrix`, i.e., to be  row-based and allow access to matrix objects via `mat[i]`. It also uses `DefaultFieldOfMatrix`, and indeed, it expects non-zero entries to be invertible.

fingolfin (commented):
Thanks for working on this! I left some remarks.

lib/matobj.gi
```diff
@@ -1345,6 +1353,8 @@ InstallMethod( Characteristic,
     [ IsMatrixOrMatrixObj ],
     M -> Characteristic( BaseDomain( M ) ) );
 
+InstallOtherMethod(DefaultFieldOfMatrix,[IsMatrixOrMatrixObj],BaseDomain);
```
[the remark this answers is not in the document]
hulpke: I have been using `DefaultFiledOfMatrix` in my code to get the definition field of an (ordinary) matrix. How would we get it in a uniform way, regardless whether an object is `IsMatrix' or `IsMatrixObj`. Will `BaseDomain` work for the "classical" matrices?

lib/matobjnz.gi
```diff
@@ -0,0 +1,1388 @@
...
+InstallMethod( PostMakeImmutable, "for a zmodnz vector", [ IsZmodnZVectorRep ],
+  function( v )
+    MakeImmutable( v![ELSPOS] );
+  end );
+
+############################################################################
+# Representation preserving constructors:
+############################################################################
+
+InstallMethod( ZeroVector, "for an integer and a zmodnz vector",
```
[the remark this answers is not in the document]
hulpke: With these installations I have been going parallel to `matobjplist.gi` -- indeed some code is almost verbatim copied. I'll see whether it works without the `IsZero` methods.

lib/matobjnz.gi
```diff
@@ -0,0 +1,1388 @@
...
+  [ IsZmodnZVectorRep, IsZmodnZVectorRep ],
+function( a, b )
+local ty,i,m;
+  if not IsMutable(a) and IsMutable(b) then
+      ty := TypeObj(b);
+  else
+      ty := TypeObj(a);
+  fi;
+  m:=Size(a![BDPOS]);
+  b:=DIFF_LIST_LIST_DEFAULT(a![ELSPOS],b![ELSPOS]);
```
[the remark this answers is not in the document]
hulpke: Again a carryover from `matobjplist.gi`. Will change,

lib/matobjnz.gi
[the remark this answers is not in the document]
fingolfin: Sorry about that: Unfortunately that file is  not up-to-date. My PR #2973 to update it hasn't been finished but should be closer to how things ought to be.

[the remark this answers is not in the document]
hulpke: So probably one should look over it again, once #2973 is included. In any case it is worth having the curretn version of the file already now as it really speeds up arithmetic.

lib/matobjnz.gi
```diff
@@ -0,0 +1,1388 @@
...
+    return Length(m![ROWSPOS]);
+  end );
+
+InstallMethod( NumberColumns, "for a zmodnz matrix",
+  [ IsZmodnZMatrixRep ],
+  function( m )
+    return m![RLPOS];
+  end );
+
+InstallMethod( DimensionsMat, "for a zmodnz matrix",
```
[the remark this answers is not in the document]
hulpke: Ditto

lib/matobjnz.gi
```diff
@@ -0,0 +1,1388 @@
...
+
+InstallMethod( ZeroImmutable, "for a zmodnz vector", [ IsZmodnZVectorRep ],
+  function( v )
+    local res;
+    res := Objectify(TypeObj(v),[v![BDPOS],ZeroImmutable(v![ELSPOS])]);
+    MakeImmutable(res);
+    return res;
+  end );
+
+InstallMethod( ZeroMutable, "for a zmodnz vector", [ IsZmodnZVectorRep ],
```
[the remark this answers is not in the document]
hulpke: OK.

lib/matrix.gi
fingolfin: In general `mat[i]` won't work on general MatrixObj instances. We debated about providing a default implementation which uses proxy objects to "fake it", but there are all kinds of pitfalls with this, and results in super inefficient code. So in the end, we agreed that it's better to get an error and be able to fix the affected code.

Ideally, it shouldn't be necessary to have two versions of the code, though: if `AddMatrixRows` etc. are used then the resulting code should work both for new and old style matrices, i.e. for everything in the filter `IsMatrixOrMatrixObj`. The main question is performance: using `AddMatrixRows` might be slower initially when used on "old style matrices". However, there is still quite some optimization potential there. E.g. one source of slow down is method dispatch is slow for lists, and thus for old-style matrices; this could be dealt with using `InstallEarlyMethod` and/or some kernel methods; I'd be happy to help work on tuning this.

fingolfin:
This looks pretty good now; @ThomasBreuer could you please also have a look?

One concern is that a lot of the new code is not being tested at all. But this is also something that could (and should) be addressed on a generic level: We need to write a (set of) generic test routine(s) that take a matrix object (or a function construction some, or whatever) and performs a bunch of generic tests with them, to verify that the APIs work as intended for that implementation. That could probably borrow from (and replace) the limited tests in `tst/testinstall/MatrixObj/` and elsewhere. Then authors of new implementations could simply invoke those existing generic tests, and already get substantial coverage of their new implementation (and confidence that things work as they should)

lib/ctblgrp.gi
```diff
@@ -270,13 +245,11 @@ end;
...
-#F  DxNiceBasis(v,space)
+#F  DxIsInSpace(v,space)
 ##
 DxIsInSpace := function(v,r)
-  if not IsBound(r.space) then
-    r.space:=VectorSpace(Field(r.base[1]),r.base);
-  fi;
-  return v in r.space;
+local nr,nc,m,i,j;
+  return SolutionMat(Matrix(BaseDomain(r.base[1]),r.base),v)<>fail;
```
ThomasBreuer, line 252: `DxIsInSpace` seems to be the only function that up to now has set the component `space`. After the changes proposed here, `CombinatoricSplit` still tries to access `space` components. Did I  misunderstand something?

lib/ctblgrp.gi
```diff
@@ -412,25 +386,28 @@ local i,pcomp,m,r,D,neue,tm,news,opt;
...
+    pcomp:=NullspaceMat(D.projectionMat*TransposedMatImmutable(Matrix(D.field,D.modulars)));
+    if Length(pcomp)>0 then
+      for i in [1..Length(D.raeume)] do
+        r:=D.raeume[i];
+        if r.dim = Length(r.base[1]) then
+          # trivial case: Intersection with full space in the beginning
+          r:=rec(base:=pcomp);
+        else
+          r:=rec(base:=SumIntersectionMat(pcomp,
+            Matrix(BaseDomain(r.base[1]),r.base))[2]);
```
ThomasBreuer, line 398: Would it perhaps make sense to make sure that the `base` component is a `MatrixObj` from the beginning, then calling `Matrix` here would not be necessary?

lib/ctblgrp.gi
```diff
@@ -1756,9 +1738,9 @@ StandardClassMatrixColumn := function(D,M,r,t)
       p:=RepresentativeAction(Stabilizer(orb.group,r),c,t);
       if p<>fail then
 	# was the first column of the galois class active?
-	if ForAny(M,i->i[c]>0) then
+	if ForAny([1..NrRows(M)],i->M[i,c]>2) then
```
ThomasBreuer, line 1741: Just for curiosity: The check for `>0` gets replaced by a check for `>2`; correct?

lib/matrix.gi
```diff
@@ -2153,6 +2153,22 @@ InstallMethod( NullspaceMat,
...
+  nr:=Length(r);
+  nc:=Length(r[1]);
+  n:=ZeroMatrix(BaseDomain(mat),nr,nc);
+  for i in [1..nr] do
+    for j in [1..nc] do
+      n[i,j]:=r[i][j];
+    od;
+  od;
+  return n;
+end);
```
ThomasBreuer, line 2170: I think we have to define what `SemiEchelonMatTransformation` returns if the argument is a MatrixObj.
The above function assumes that `relations` contains a list of lists; is this reasonable?
(And in case that the nullspace is trivial, the result should be a MatrixObj.)

lib/matrix.gi
```diff
@@ -2348,9 +2364,9 @@ InstallOtherMethod( RankMat,
 ##
 #M  SemiEchelonMat( <mat> )
 ##
-InstallMethod( SemiEchelonMatDestructive,
+InstallOtherMethod( SemiEchelonMatDestructive,
     "generic method for matrices",
-    [ IsMatrix and IsMutable ],
+    [ IsMatrixOrMatrixObj and IsMutable ],
```
ThomasBreuer, line 2369: I think rather the declaration of `SemiEchelonMatDestructive` should be changed to expect `IsMatrixOrMatrixObj and IsMutable`.

lib/matobjnz.gi
```diff
@@ -0,0 +1,1422 @@
...
+  m:=Size(a![BDPOS]);
+  b:=a![ELSPOS] - b![ELSPOS];
+  for i in [1..Length(b)] do if b[i]<0 then b[i]:=b[i] mod m;fi;od;
+  return Objectify(ty,[a![BDPOS],b]);
+end );
+
+InstallMethod( \=, "for two zmodnz vectors",
+  [ IsZmodnZVectorRep, IsZmodnZVectorRep ],
+  function( a, b )
+    return EQ_LIST_LIST_DEFAULT(a![ELSPOS],b![ELSPOS]);
```
ThomasBreuer, line 295: Is it intended that two VectorObj over different rings are regarded as equal (and can be added, etc.)?

ThomasBreuer:
@fingolfin I had a look and left some comments.
@hulpke Thanks a lot. Concerning my more general comments on the setup of matrix objects (and the documentation), perhaps I can deal with these issues after your pull request got merged.

lib/ctblgrp.gi
hulpke: Thank you for spotting. No, this is an editing error (even though its effects are harmless and only introduce tiny inefficiencies). Wil fix.

lib/ctblgrp.gi
hulpke: It could be. In the old setup a matrix is just a list of vectors, and there was no need to distinguish the concepts. Here, one has to, and it seemed naively that a basis should be a list of vectors rather than a matrix.

lib/ctblgrp.gi
hulpke: I think the line in `CombinatoricSplit` is a relic of gap3 code that now never runs. In any case, it should have been a call instead to DxIsInSpace.

lib/matrix.gi
hulpke: I would be happy with that.

lib/matobjnz.gi
hulpke: AH -- did not spot that a test for the families is missing -- will add.

lib/matrix.gi
hulpke: Yes, this needs cleanup, and I felt I did not yet have sufficient experience with matrix objects to make this definition.

hulpke:
> One concern is that a lot of the new code is not being tested at all. 

I have further code that makes `ImmutableMatrix` over ZmodnZ rings use the new matrices and (for fields) enable the MeatAxe. This makes it easy to get lots of testing through calling higher level routines.

lib/ctblgrp.gi
```diff
@@ -336,6 +309,7 @@ DxRegisterModularChar := function(D,c)
 local d,p;
   # it may happen,that an irreducible character will be registered twice:
   # 2-dim space,1 Orbit,combinatoric split. Avoid this!
+  if IsPlistRep(c) then c:=Vector(c); fi;
```
fingolfin, line 312: Are the entries of `D.modulars` modified later on, or will they be immutable? In the latter case, you could also try
```suggestion
  if IsPlistRep(c) then c:=ImmutableVector(SOME_RING, c); fi;
```
provided `SOME_RING` can be derived from `D` or can be passed in some other way to this function.

Even if it is mutable, it might be a good idea to pass the basering and/or filter: The problem with `Vector(c)` is that if one passes a list-of-GF(4) elements to it, then sometimes one will get a GF(4) vector yet sometimes a GF(2) vector, which usually is not desirable

lib/ctblgrp.gi
```diff
@@ -729,15 +705,21 @@ InstallGlobalFunction(SplitStep,function(D,bestMat)
...
-      N:= NullMat( dim, dim, o );
+      N:= ZeroMatrix(f,dim,dim);
       for row in [1..dim] do
-        Row:=base[row]*M;
+        #Row:=base[row]*M;
+        Row:=CompatibleVector(M);
+        for col in [1..Length(Row)] do
+          Row[col]:=base[row,col];
+        od;
+        Row:=Row*M;
```
fingolfin, line 715: Note to self: I think we need a function doing this; i.e. which takes a matrix `M`, and a list of co/a vector `v`, and returns a new vector in "compatible" representation. 

Though in this case, perhaps `base` could be turned into a list of vectors of the right type from the start, i.e. as part of its construction?

lib/matobj.gi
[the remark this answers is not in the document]
fingolfin: Yes `BaseDomain` works for "classical" matrices (but there was a bug: it only looked .

But since it has different semantics, it is not so clear what the "correct" answer should be for e.g. an identity matrix with integer entries: should it be `Integers` or `Rationals`? This is of course clear with `DefaultFiledOfMatrix`. But `BaseDomain` currently returns `Integers`. We could of course change that, but in the end, it'll always be a guessing game; which is after all one of the primary reasons why we want MatrixObj: so that we can stop guessing and start specifying what we want.

fingolfin (commented):
How shall we proceed with this? There are some issues, but it still is an improvement, and the issues listed are not more severe than other issues that still exist in the MatrixObj code, sooo... we could merge it and improve later, together with other things; perhaps even organize another MatrixObj coding sprint.

The only "danger" I see in merging is that someone might start to rely on some specific (but undocumented) code behavior and then if we change it, things break for them. But that's already the case with the rest of MatrixObj...

Thoughts?

hulpke:
> How shall we proceed with this? There are some issues, but it still is an improvement, and the issues listed are not more severe than other issues that still exist in the MatrixObj code, sooo... we could merge it and improve later, together with other things; perhaps even organize another MatrixObj coding sprint.

I am very much in favor of merging. As far as I can see , while not perfect, it does not break (or slow down) existing functionality, but gives a substantial speedup to character table calculations.

If there are changes that *must* be done before merging, please let me know and I will be happy to make them, but lets not hold these improvements back just because some concepts for matrix objects are not yet perfect -- only if we use them we will find out what the right objects are.

ThomasBreuer:
@hulpke
> I am very much in favor of merging.

I would also propose to merge the current state.
My comment from last week was intended to propose this, and to open a new pull request in order to deal with the general changes that are motivated by the current pull request.

lib/ctblgrp.gi
ThomasBreuer: @fingolfin What about the syntax `Vector( <M>, <v> )`?

hulpke:
So, @fingolfin @ThomasBreuer , is one of you wiling to approve, so we can merge?

ThomasBreuer (approved):

fingolfin (approved):

james-d-mitchell:
It's a shame that this was merged before the impact on packages was fully realised, [e689844](https://github.com/gap-system/gap/pull/4773/commits/e6898446bc6a89e48c9ed37c3be18cb637ac005e) completely breaks the Semigroups package. Please see Issue #4814. @hulpke @fingolfin

fingolfin:
As far as I know, use of this enhancement has been effectively disabled by PR #4983 (hopefully only temporarily, see #4988), so I am not going to add this to the release notes. Should I be mistaken in this assessment, we can still add it retroactively in GAP 4.12.1.

PR #4774, opened 2022-02-17
kernel: add SymbolTable type, change GetInputFilenameID to use it

### LINKED ISSUE #4771: Performance regression

fingolfin:
- kernel: get rid of RNamNameWithLen
- kernel: add a dedicated SymbolTable type
- kernel: switch GetInputFilenameID to use a SymbolTable

Resolves #4771 (mostly, at least...?)

Needs some cleanup, but overall seems to work. Using the test function from #4771, and together with PR #4772, I now get the following:
```
gap> for i in [1..100000] do EvalString("12+13"); od; time;
125
gap> for i in [1..100000] do myeval("12+13"); od; time;
119
```

Granted, this is still slower than GAP 4.8, but only by a factor of about 1.5, not 5.5.

### DIFF
--- a/src/gvars.c
+++ b/src/gvars.c
@@ -1567,7 +1567,7 @@ static Int PostSave (
 static Int InitLibrary (
     StructInitInfo *    module )
 {
-    InitSymbolTableLibrary(&GVarSymbolTable);
+    InitSymbolTableLibrary(&GVarSymbolTable, 28069);
 
     /* make the error functions for 'AssGVar'                              */
     ErrorMustEvalToFuncFunc = NewFunctionC(
--- a/src/gvars.c
+++ b/src/gvars.c
@@ -39,6 +39,7 @@
 #include "modules.h"
 #include "plist.h"
 #include "stringobj.h"
+#include "symbols.h"
 #include "sysstr.h"
 
 #ifdef HPCGAP
@@ -97,6 +98,7 @@ static Obj   ValGVars;
 static Obj * PtrGVars;
 #endif
 
+static SymbolTable GVarSymbolTable;
 
 #ifdef HPCGAP
 
@@ -106,46 +108,6 @@ static Obj * PtrGVars;
 */
 
 static Obj TLVars;
-
-/****************************************************************************
-**
-*V  GVarLock  . . . . . . . . . . . . . . . . . .  lock for global variables
-**
-**  This lock is only needed for accessing global variables by name rather
-**  than index and to initialize copy/fopy information.
-*/
-
-static pthread_rwlock_t GVarLock;
-static void *GVarLockOwner;
-static UInt GVarLockDepth;
-
-static void LockGVars(int write) {
-  if (PreThreadCreation)
-    return;
-  if (GVarLockOwner == GetTLS()) {
-    GVarLockDepth++;
-    return;
-  }
-  if (write) {
-    pthread_rwlock_wrlock(&GVarLock);
-    GVarLockOwner = GetTLS();
-    GVarLockDepth = 1;
-  }
-  else
-    pthread_rwlock_rdlock(&GVarLock);
-}
-
-static void UnlockGVars(void) {
-  if (PreThreadCreation)
-    return;
-  if (GVarLockOwner == GetTLS()) {
-    GVarLockDepth--;
-    if (GVarLockDepth != 0)
-      return;
-    GVarLockOwner = NULL;
-  }
-  pthread_rwlock_unlock(&GVarLock);
-}
 
 #endif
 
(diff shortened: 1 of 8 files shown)


fingolfin:
Regarding performance impact on `GVarName` and `RNamName`: they both now use the new `LookupSymbol` helper. To allow this to be generic, it uses a callback `symtab->nameFunc` to map symbol ids back to symbol strings. So this requires a jump to a function pointer, which usually just amounts to an `ELM_PLIST` invocation. Whereas in the old code, we directly used `ELM_PLIST`, inlined and all. So I was a bit vary about a performance impact.

But it turns out there isn't really one, at least not according to this incredibly sophisticated test:
```
gap> str:="dfsdfdsf";;GASMAN("collect"); for i in [1..1000000] do RNamObj(str); od; time;
31
gap> str:="dfsdfdsf";;GASMAN("collect"); for i in [1..1000000] do RNamObj(str); od; time;
34
gap> str:="dfsdfdsf";;GASMAN("collect"); for i in [1..1000000] do RNamObj(str); od; time;
33

# the first iteration below generates tons of hash table entries, so its slower than
# the repetitions
gap> GASMAN("collect"); for i in [1..1000000] do RNamObj(i); od; time;
199
gap> GASMAN("collect"); for i in [1..1000000] do RNamObj(i); od; time;
59
gap> GASMAN("collect"); for i in [1..1000000] do RNamObj(i); od; time;
60
```

I am only going to paste one, because the results for master and this PR are basically indistinguishable, they both look like that (with the usual fluctuations).

src/gvars.c
```diff
@@ -107,46 +109,6 @@ static Obj * PtrGVars;
...
-static pthread_rwlock_t GVarLock;
-static void *GVarLockOwner;
-static UInt GVarLockDepth;
-
-static void LockGVars(int write) {
-  if (PreThreadCreation)
-    return;
-  if (GVarLockOwner == GetTLS()) {
-    GVarLockDepth++;
-    return;
```
fingolfin, line 127: Note that this old locking code allowed for recursive locking when called from the same thread.

I thought this was irrelevant, as I saw no way this could ever be needed.

Alas, it is, as `UpdateCopyFopyInfo` also takes that lock, and it then calls `DeclareAllGVars` while holding the lock, and *that* calls `GVarName`.... *sigh*. I think at this point it's best to just excise the complete `GVarDescriptor` code; almost nothing really uses it anyway. I'll look into that.

fingolfin: *sigh* it's not quite as simple. And I don't want to accidentally break HPC-GAP even more (though perhaps I should just give up on that pipe dream *double sigh*).

So, I guess I'll have to replicate the recursive locking ability

ChrisJefferson (approved):

PR #4776, opened 2022-02-18
CI: add a test kernel extension to mockpkg

fingolfin:
This is preparation for PR #4746 but also useful on its own.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -55,7 +55,7 @@ jobs:
             "testmanuals",
 
             # test error reporting and compiling as well as libgap
-            "testexpect testspecial test-compile testlibgap testkernel",
+            "testexpect testmockpkg testspecial test-compile testlibgap testkernel",
 
             # compile packages and run GAP tests
             # don't use --enable-debug to prevent the tests from taking too long
@@ -98,7 +98,7 @@ jobs:
           # this job also tests GAP without readline
           - os: macos-latest
             shell: bash
-            test-suites: "testinstall"
+            test-suites: "testmockpkg testinstall"
             extra: "BOOTSTRAP_MINIMAL=yes"
 
           # run bugfix regression tests
@@ -123,7 +123,7 @@ jobs:
           # tests using valgrind, as it is too slow.
           - os: ubuntu-18.04
             shell: bash
-            test-suites: "testbuildsys testinstall"
+            test-suites: "testbuildsys testmockpkg testinstall"
             extra: "NO_COVERAGE=1 ABI=64 BUILDDIR=out-of-tree
                     CONFIGFLAGS=\"--enable-valgrind\""
             packages: "valgrind"
@@ -132,7 +132,7 @@ jobs:
           # elsewhere in this file for an explanation).
           - os: ubuntu-18.04
             shell: bash
-            test-suites: "testbuildsys testinstall"
+            test-suites: "testbuildsys testmockpkg testinstall"
             extra: "NO_COVERAGE=1 ABI=32 BUILDDIR=out-of-tree CONFIGFLAGS=\"\""
 
           # test Julia integration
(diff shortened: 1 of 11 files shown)


wilfwilson:
The [CI failure](https://github.com/gap-system/gap/runs/5245121335?check_suite_focus=true) is unrelated and something that I have noticed a few times:
```
+ echo 'Running test suite testexpect'
+ case $TEST_SUITE in
+ actual=/tmp/gaptest.expect.actual
+ INPUTRC=/tmp/inputrc
+ expect -c 'spawn bin/gap.sh --quitonbreak -A -b  --cover coverage/testexpect.coverage' /home/runner/work/gap/gap/dev/gaptest.expect
Running test suite testexpect
spawn bin/gap.sh --quitonbreak -A -b --cover coverage/testexpect.coverage
 TIMEOUT 
```

fingolfin:
Yeah, probably need to sprinkle a few `sleep` statements into the expect script

fingolfin:
I've restarted this twice now: the first time it failed the `CI / Send Slack notification on status change (push)`, the second time on `testexpect`, now again on `CI / Send Slack notification on status change (push), with message:
```
workflow id: 3569081
Invalid $last_status: skipped
Error: Process completed with exit code 1.
```
Any idea what that's about?

wilfwilson:
I think I understand:

Since the branch of this PR is actually on this repo, then every time you push, two builds are triggered: the one for pushing a branch, and the one for updating a PR.

For some reason, the PR build is the newest one (but it is immediately skipped because it's a duplicate).

When the push CI finishes on the pushed branch, it runs the `CI / Send Slack notification on status change (push)` job, which looks for the most recent CI build on the `mh/tst-mock-pkg-kext` branch.

It finds the build on the branch trigger by the _PR update_. This build was skipped.

`gap-actions/should-i-notify-action` does not regard `skipped` as a valid status: https://github.com/gap-actions/should-i-notify-action/blob/main/action.yml and therefore is fails.


The correct approach would be for `gap-actions/should-i-notify-action` to ignore builds that were triggered by pull requests. It should also gracefully accept `skipped` as a valid status (although this should probably never occur in practice).

wilfwilson:
I have made a change to `gap-actions/should-i-notify-action` that might fix things...

fingolfin:
@wilfwilson seems your change did fix it, yay :-)

wilfwilson (changes requested):
Actually, it doesn't necessarily work properly... https://github.com/gap-system/gap/runs/5254890953?check_suite_focus=true
```
 ┌───────┐   GAP 4.12dev built on 2022-02-18 23:39:17+0000
 └───────┘   Architecture: x86_64-pc-cygwin-default64-kv8
 Packages:   GAPDoc 1.6.5, PrimGrp 3.4.1, SmallGrp 1.4.2, TransGrp 3.3
Could not read file "/cygdrive/d/a/gap/gap/tst/mockpkg;".
Could not read file "coverage/testmockpkg.coverage".
########> Diff in /cygdrive/d/a/gap/gap/tst/mockpkg/tst/kext.tst:4
# Input is:
LoadPackage("mockpkg");
# Expected output:
true
# But found:
#I  mockpkg package is not available. Check that the name is correct
fail
########
########> Diff in /cygdrive/d/a/gap/gap/tst/mockpkg/tst/kext.tst:6
# Input is:
IsKernelExtensionAvailable("mockpkg");
# Expected output:
true
# But found:
false
########
########> Diff in /cygdrive/d/a/gap/gap/tst/mockpkg/tst/kext.tst:8
# Input is:
LoadKernelExtension("mockpkg");
# Expected output:
true
# But found:
false
########
########> Diff in /cygdrive/d/a/gap/gap/tst/mockpkg/tst/kext.tst:10
# Input is:
TestCommand();
# Expected output:
true
# But found:
Error, Variable: 'TestCommand' must have a value
########
```

wilfwilson:
I guess the test file needs to be run in such a way that GAP returns a non-zero exit code when it fails.

fingolfin:
Argh, I wanted to run `tst/testall.g` instead of the .tst but forgot. Thanks for paying close attention!

wilfwilson (approved):

fingolfin:
Seems to be really working now, even on Cygwin (as in: I checked the logs)

The build failure earlier was because I stupidly messed up the order of some command line arguments for gap 🙄

PR #4777, opened 2022-02-18
CI: tweak expect test to make it a bit more robust

fingolfin:
I think it sometimes takes too long to start GAP on the CI, hence it dies. So let's increase the timeout


wilfwilson (approved):
Happy to see if this helps

PR #4778, opened 2022-02-18
gac: remove support for static linking

fingolfin:
This has virtually no use case anymore since dynamic linking is supported
everywhere. At the same time it is an impediment for `make install` support.
So it just doesn't seem worth the effort to preserve it.


ChrisJefferson (approved):

ChrisJefferson:
I agree that static linking was useful in the past, but (hopefully) isn't going to be required in the future.

frankluebeck (approved):
Static linking was very useful for some time to produce portable binaries (e.g., for the old version of the GAP rsync-distribution). But this doesn't work reasonably well for some years.

So, I agree with removing the support for static linking.

The code changes look ok to me.

PR #4789, opened 2022-02-22
Allow Cygwin binaries to be stored in /usr/bin or /bin

ChrisJefferson:
Some cygwin installs have binaries in /bin instead of /usr/bin.
We cannot use Filename(DirectoriesSystemPrograms(), X), because
system.g is loaded before most of the standard library.


fingolfin (approved):
Sure, why not

PR #4795, opened 2022-02-24
Better errors when loading packages fails

ChrisJefferson:
This PR tries to improve the messages we show users when a package fails to load. It means when a package fails to load users are shown a (hopefully brief) explanation of what went wrong, without having to rerun, turning up the info level.

Some general comments:

* Messages are only shown if the package banner isn't hidden (under the assumption people who want the banner hiding don't want lots of output).
* Messages are only shown for failures of required dependencies (unlike if you turn up the info level when you see information about both required and optional dependencies).
* This PR also adds version numbers to errors, which is useful if users have multiple versions installed (this could be separated)

As an example, the package `Vole` has a number of requirements, one of which is `BacktrackKit`, which requires `Digraphs`. I have two digraphs installed, neither is compiled. The new output is:

```
#I  Digraphs version 1.5.0: PackageAvailabilityInfo: the AvailabilityTest function returned fail
#I  Digraphs version 1.3.1: PackageAvailabilityInfo: the AvailabilityTest function returned fail
#I  Digraphs: PackageAvailabilityInfo: no installed version fits
#I  BacktrackKit version 0.6.2: PackageAvailabilityInfo: dependency 'digraphs' is not satisfied
#I  BacktrackKit: PackageAvailabilityInfo: no installed version fits
#I  Vole version 0.5.3: PackageAvailabilityInfo: dependency 'backtrackkit' is not satisfied
#I  Vole: PackageAvailabilityInfo: no installed version fits
#I  Vole package is not available. To see further details, enter
#I  SetInfoLevel(InfoPackageLoading,4); and try to load the package again.
```

There is one possible annoyance -- errors are printed for packages if a newer version does not work, but an older one does, for example, if I compile digraphs 1.3.1 from above, now the only error I see is:

```
#I  Digraphs version 1.5.0: PackageAvailabilityInfo: the AvailabilityTest function returned fail
```

I personally feel this is useful -- it tells the user they have a more recent version that wasn't loaded. However, it's also hard to hide this message without significant overhaul.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -599,7 +599,7 @@ InstallGlobalFunction( LogPackageLoadingMessage, function( arg )
       currpkg:= arg[3];
     elif IsBound( GAPInfo.PackageCurrent ) then
       # This happens inside availability tests.
-      currpkg:= GAPInfo.PackageCurrent.PackageName;
+      currpkg:= Concatenation(GAPInfo.PackageCurrent.PackageName, " version ", GAPInfo.PackageCurrent.Version);
     else
       currpkg:= "(unknown package)";
     fi;
@@ -666,7 +666,7 @@ InstallGlobalFunction( PackageAvailabilityInfo,
     function( name, version, record, suggested, checkall )
     local InvalidStrongDependencies, Name, equal, comp, pair, currversion,
           inforec, skip, msg, dep, record_local, wght, pos, needed, test,
-          name2, testpair;
+          name2, testpair, fullname;
 
     InvalidStrongDependencies:= function( dependencies, weights,
                                           strong_dependencies )
@@ -761,6 +761,7 @@ InstallGlobalFunction( PackageAvailabilityInfo,
     for inforec in PackageInfo( name ) do
 
       Name:= inforec.PackageName;
+      fullname := StringFormatted("{} version {}", inforec.PackageName, inforec.Version);
       skip:= false;
       currversion[2]:= inforec.Version;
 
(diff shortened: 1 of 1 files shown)


hulpke (approved):
Very useful functionality

wilfwilson (approved):
I really like this too.

ThomasBreuer:
My interpretation of this proposal is as follows:

One part is to extend those info messages about package loading that belong to a specific package version by this version number. I think this is clearly an improvement.

The second part is to show by default those messages that belong to the failure when trying to load some version of a package. 
This feature is restricted to the case when one loads a package such that printing the package banner is enabled. This means that this feature strikes when the user calls `LoadPackage`; it is not intended for autoloaded packages.
The effect is achieved by showing the relevant messages at info level 1 of `InfoPackageLoading` instead of level 3.
Up to now, a level 1 (=`PACKAGE_ERROR`) was shown only in the situation that a needed package of GAP was not loadable, and then also `Error` was called. Thus this level was not really needed.

I think that if we regard the messages that will now get ranked up from level 3 to 1 as so important that users should see them then also the current level 2 messages should be shown. They usually arise in the individual `AvailabilityTest` functions, and they belong to situations where only parts of the package functionality can be provided, due to some missing functionality. I think that from the viewpoint of getting information about "something not being installed properly", they are of the same interest.

It seems that the current symbolic names for the info levels (`PACKAGE_ERROR`, `PACKAGE_WARNING`, `PACKAGE_INFO`, `PACKAGE_DEBUG`) are not really suitable.

ChrisJefferson:
I agree getting the messages from the AvailabilityTests is useful, I will think about the best way to do that (without having to go change all the packages, as that would be annoying). Probably change the `InfoLevel` (I did not want to do that in general, as I do not want to show the user INFO messages which just say the package was successfully loaded).

PR #4796, opened 2022-02-24
DOC: Note on `IsFpGroup(FreeGroup(infinity))`

### LINKED ISSUE #4793: `IsFpGroup(FreeGroup(infinity))` should be `false`
In the current GAP master branch:

### Observed behaviour
```
gap> IsFpGroup(FreeGroup(infinity));
true
```
### Expected behaviour
```
gap> IsFpGroup(FreeGroup(infinity));
false
```
since a free group of infinite rank is not finitely generated, and therefore not finitely presented.

---
Note also that the following behaviour, although not mathematically weird, is inconsistent with the behaviour for free semigroups, as detailed in https://github.com/gap-system/gap/issues/4786#issuecomment-1049726750:
```
gap> IsFpGroup(FreeGroup(2));
true
gap> IsFpSemigroup(FreeSemigroup(2));
false
```

hulpke:
This resolves #4793

### DIFF
--- a/lib/grpfp.gd
+++ b/lib/grpfp.gd
@@ -211,8 +211,8 @@ DeclareCategory( "IsSubgroupFgGroup", IsGroup );
 ##  <Filt Name="IsSubgroupFpGroup" Arg='H' Type='Category'/>
 ##
 ##  <Description>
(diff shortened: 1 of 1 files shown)


wilfwilson (approved):
Thanks @hulpke

PR #4797, opened 2022-02-25
ImmutableMatrix creates MatrixObj for Z/nZ

hulpke:
and subsequent changes in group, meataxe, and polynomial code to make it work. This also provides further  tests for #4773

The result is a small speedup when working with matrix groups/modules over ZmodnZ rings.

## Text for release notes

N/A

### DIFF
--- a/lib/grpmat.gi
+++ b/lib/grpmat.gi
@@ -240,6 +240,7 @@ local field, dict, acts, start, j, zerov, zero, dim, base, partbas, heads,
   else
     Error("illegal action");
   fi;
+  start:=List(start,x->ImmutableVector(field,x));
 
   zerov:=Zero(start[1]);
   zero:=zerov[1];
@@ -355,7 +356,13 @@ local field, dict, acts, start, j, zerov, zero, dim, base, partbas, heads,
 	if v<>zerov then
 	  Add(base,orb[i]);
 	  Add(partbas,ShallowCopy(orb[i]));
+          if ForAny(partbas,IsVectorObj) then
+            partbas:=Matrix(BaseDomain(partbas[1]),partbas);
+          fi;
 	  TriangulizeMat(partbas);
+          if IsMatrixObj(partbas) then
+            partbas:=ShallowCopy(RowsOfMatrix(partbas));
+          fi;
 	  heads:=List(partbas,PositionNonZero);
 	  if Length(partbas)>=dim then
 	    # full dimension reached
--- a/lib/grpmat.gi
+++ b/lib/grpmat.gi
@@ -356,7 +356,8 @@ local field, dict, acts, start, j, zerov, zero, dim, base, partbas, heads,
 	if v<>zerov then
 	  Add(base,orb[i]);
 	  Add(partbas,ShallowCopy(orb[i]));
-          if ForAny(partbas,IsVectorObj) then
+          # filter for vector objects, not compressed FF vectors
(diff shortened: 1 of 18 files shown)


tst/teststandard/opers/Matobjnz.tst
```diff
@@ -0,0 +1,17 @@
...
+gap> g:=AlternatingGroup(5);;
+gap> mo:=IrreducibleModules(g,GF(p));;
+gap> Set(List(mo[2],x->x.dimension));
+[ 1, 4, 5, 6 ]
+gap> h:=Group(mo[2][2].generators);
+<matrix group with 2 generators>
+gap> Size(h);
+60
+gap> Length(ConjugacyClassesSubgroups(h));
+9
```
fingolfin, line 14: These two tests fail with errors

lib/vecmat.gi
```diff
@@ -1586,10 +1595,16 @@ local sf, rep, ind, ind2, row, i,big,l;
     rep:=IsPlistRep;
   fi;
 
+  if IsPlistRep(matrix) and not ForAll(matrix,IsPlistRep) then
+    nr:=Length(matrix);
+  else
+    nr:=NrRows(matrix);
+  fi;
+
```
fingolfin, line 1603: Why is this necessary? `NrRows` should work for "old-style" matrices, too

lib/grpmat.gi
```diff
@@ -972,6 +979,81 @@ local G,fam,typ,f;
...
+local G,typ,f;
+
+  if not IsFinite(gens) then TryNextMethod(); fi;
+  typ:=MakeGroupyType(FamilyObj(gens), IsGroup and IsAttributeStoringRep 
+          and HasGeneratorsOfMagmaWithInverses and IsFinitelyGeneratedGroup
+          and HasIsEmpty and IsFinite and HasOne,
+          gens,id,true);
+
+  f:=DefaultScalarDomainOfMatrixList(gens);
+  gens:=List(Immutable(gens),i->ImmutableMatrix(f,i));
```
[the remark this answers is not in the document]
hulpke: I suspect this is the result of a global insertion of `ImmutableMatrix` in many places without considering in each place whether the local code could be improved subsequently.

lib/vecmat.gi
hulpke: Did not realize that -- chaned.

tst/teststandard/opers/Matobjnz.tst
hulpke: y worked initially, but subsequent changed had made them fail. Fixed again.

lib/vecmat.gi
hulpke: Actually...
It is needed, because the input could be a mixed format list, for which `NrRows` will not work.

fingolfin: Could you elaborate on that a bit? What is a "mixed format list" and where does it occur?

fingolfin (approved):
Let's just merge this

lib/vecmat.gi
hulpke: While merged already, a brief explanation for later:

The function is `ImmutableMatrix`. It takes somethig that could be construed as a list of vectors, and converts it into a compact format matrix (with the implicit claim that this representation will be efficient to work with).
This could be a list of vectors of different storage/mutability/base ring status (and indeed, this can happen when calling from the MeatAxe).

PR #4799, opened 2022-02-27
kernel: change `SyExecuteProcess` to use `posix_spawn`

fingolfin:
@ChrisJefferson does this work under cygwin? If so, does it fix the bug where `Process` is changing the working dir for the parent process?

The patch is not ready for merging in any case:
- handling of SIGCHLD not integreated
- moving `SyExecuteProcess` to `iostream.c` is kind of a hack, but for the sake of quickly experimenting I didn't want to bother with trying to come up with a "clean" solution.

### DIFF
--- a/src/iostream.c
+++ b/src/iostream.c
@@ -66,10 +66,6 @@
 #include <spawn.h>
 #endif
 
-#ifdef SYS_IS_CYGWIN32
-#include <process.h>
-#endif
-
 #ifdef HAVE_SELECT
 #include <sys/time.h>
 #endif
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
This works on cygwin, and stops the directory changing behaviour

ChrisJefferson (approved):

fingolfin:
I decided to just make the move of `SyExecuteProcess` to `iostream.c` permanent, there seems little point in trying to abstract this away and it makes code sharing much easier.

src/iostream.c
```diff
@@ -548,6 +555,211 @@ StartChildProcess(const Char * dir, const Char * prg, Char * args[])
...
+
+    // spawn subprocess
+    pid_t pid;
+    if (posix_spawn_with_dir(&pid, prg, &file_actions, 0, args, environ,
+                             dir)) {
+        PErr("SyExecuteProcess: posix_spawn_with_dir failed");
+        return -1;
+    }
+
+    // wait for some action
```
fingolfin, line 627: Note: the "classic" code also has some logic to "ignore a CTRL-C". And it disables our SIGCHLD handler temporarily. We may need to add one or both back here.

PR #4801, opened 2022-03-01
BuildPackages.sh: don't use GAP_ABI anymore

fingolfin:
Don't try to inject 32bit build flags into package build systems. Any necessary flags should already be contained in GAP_{C,CXX,LD}FLAGS, which package build systems should honor. If they don't, they should be adjusted.

This change might break tests in 32bit mode, requiring fixes to individual packages. But overall it should be fine. And 32bit mode realistically only is a "2nd class citizen" these days, and we may eventually want to make this official, too...

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -124,26 +124,6 @@ else
 fi
 
 
-
-# detect whether GAP was built in 32bit mode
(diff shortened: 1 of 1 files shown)


fingolfin:
Building `nq` fails...

PR #4807, opened 2022-03-02
improve `GroupHomomorphismByFunction`

### LINKED ISSUE #4806: missing `PreImagesRepresentative` method?
The following happens in GAP 4.11 as well as in the master branch.
```
gap> G:= SL(2,2);;
gap> hom:= GroupHomomorphismByFunction( G, G, x -> x );;
gap> PreImagesRepresentative( hom, G.1 );  
Error, no default method for s.p. general mapping at /.../lib/mapping.gi:... called from
<function "PreImagesRepresentative for s.p. general mapping, and element">( <arguments> )
you can 'return;' to continue
brk> 
```
For permutation groups and PC groups, GAP uses the defining function (`x -> x`) for creating a group homomorphism by images, and then delegates the task to this mapping.
(This is done by setting the filter `IsPreimagesByAsGroupGeneralMappingByImages` during the construction of the homomorphism, for these kinds of groups.)
What is the reason to forbid this for other kinds of groups?

ThomasBreuer:
resolves #4806

Set `IsPreimagesByAsGroupGeneralMappingByImages` for any
homomorphism constructed with `GroupHomomorphismByFunction`
that does not know a function or computing preimages.

### DIFF
--- a/lib/ghom.gi
+++ b/lib/ghom.gi
@@ -1552,6 +1552,17 @@ RedispatchOnCondition(IsomorphismPcGroup,true,[IsGroup],[IsFinite],0);
 ##
 #F  GroupHomomorphismByFunction( <D>, <E>, <fun> )
 #F  GroupHomomorphismByFunction( <D>, <E>, <fun>, <invfun> )
+#F  GroupHomomorphismByFunction( <D>, <E>, <fun>, false, <prefun> )
+##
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Looks good to me

hulpke (approved):

PR #4810, opened 2022-03-02
Let `SimplifiedFpGroup` and `IsomorphismSimplifiedFpGroup` transfer attributes (such as size, being abelian etc.) to the simplified group

fingolfin:


hulpke (commented):
It it clear that `UseIsomorphismRelation` does not do anything (such as comparing elements) which could cause problems for fp groups?

fingolfin:
It should: It purely transfers select filters/properties/attributes installed via `InstallIsomorphismMaintenance` or `InstallFactorMaintenance`. I.e. if one of these is set for the one group, it is set identical for the other.

The only hypothetical problem I can think of would be if this leads to some immediate method triggering that does something problematic. But I'd say this then would be a problem with that immediate method, which should be resolved with that.

hulpke (approved):
Thank you --  this should be all safe.

PR #4813, opened 2022-03-04
BuildPackage.sh: when run in GH Action, use grouping feature

fingolfin:
- HACK: temporarily disable some CI workflows
- BuildPackage.sh: when run in GH Action, use grouping feature

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -216,6 +216,7 @@ GAPInput
 build_one_package() {
   # requires one argument which is the package directory
   PKG="$1"
+  [[ $GITHUB_ACTIONS = true ]] && echo "::group::$PKG"
   echo ""
   date
   echo ""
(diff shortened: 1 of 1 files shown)


ChrisJefferson:
I agree with this PR in principle, assuming it does what it is supposed to.

fingolfin:
Seems to work well. Here is an example where I deliberately inserted a failure: https://github.com/gap-system/gap/runs/5420553265?check_suite_focus=true 

I've also changed BuildPackage.sh so that if you give it no package names, and it determines a list of packages to build by itself, it will sort those alphabetically.

Anyway, I've dropped the hacks, this should be ready to merge if/once tests pass.

fingolfin:
See also https://github.com/gap-system/gap/runs/5421339772?check_suite_focus=true for a nice example.

@ChrisJefferson if you like it please approve/merge :-)

ChrisJefferson (approved):

PR #4823, opened 2022-03-10
BugFix: Fix crash when QuitGap is called while reading from an InputTextString

### LINKED ISSUE #4820: Segfault doing `./gap -A  -c 'FORCE_QUIT_GAP(TestPackage("aclib"));'`
This "works" for any package, so this is a bug in GAP, not the fault of `aclib`, which just happened to be the example I last tried this with.

ChrisJefferson:
Fixes #4820 

I don't really like this, but I also don't have a better option.


ChrisJefferson:
The title isn't quite right, but is a simplified explanation for release notes

fingolfin (approved):

PR #4824, opened 2022-03-10
CI: install fewer texlive packages

fingolfin:
Before: installs 79 packages, downloads 542 MB, uses 1482 MB disk space

After: installs 26 packages, downloads 77.8 MB, uses 274 MB disk space

By far the biggest chunk was `texlive-fonts-extra` which needed to
download ~350 MB and used ~900 MB disk space.

---

Let's see how it works in practice. I may have to add back some packages.

See also https://github.com/gap-actions/build-pkg-docs/pull/12

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -199,9 +199,7 @@ jobs:
                          texlive-latex-base
                          texlive-latex-recommended
                          texlive-latex-extra
-                         texlive-extra-utils
                          texlive-fonts-recommended
-                         texlive-fonts-extra
                        )
                    fi
                    if [[ $ABI == 32 ]] ; then
@@ -212,7 +210,7 @@ jobs:
                        packages+=(gcc-multilib g++-multilib)
                    fi
                    sudo apt-get update
-                   sudo apt-get install "${packages[@]}"
+                   sudo apt-get install --no-install-recommends "${packages[@]}"
                elif [ "$RUNNER_OS" == "macOS" ]; then
                    brew install gmp zlib
                fi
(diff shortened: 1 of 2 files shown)


wilfwilson (commented):
Something's not working:
[makemanuals - - ubuntu-18.04](https://github.com/gap-system/gap/runs/5503806870?check_suite_focus=true)
```
No files were found with the provided path: doc/dev/manual.pdf doc/hpc/manual.pdf doc/ref/manual.pdf doc/tut/manual.pdf.
No artifacts will be uploaded.
```
Presumably because:
```
#W There were LaTeX errors:
! Font T1/cmr/m/n/10.95=ecrm1095 at 10.95pt not loadable: Metric (TFM) file no\
t
 found.
<to be read again> 
____________________
! LaTeX Error: File `enumitem.sty' not found.

Type X to quit or <RETURN> to proceed,
____________________
! Emergency stop.
<read *> 
         
____________________
!  ==> Fatal error occurred, no output PDF file produced!
____________________

#I ERROR: no .pdf file produced (writing incomplete .six file)
```

fingolfin:
Yet the CI tests report as "passing", which is to me almost more worrying :-(

wilfwilson:
@sTertooy suggests in https://github.com/gap-actions/build-pkg-docs/issues/9#issuecomment-1065960972 that we could add the `if-no-files-found: error` option to our `actions/upload-artifact` steps.

I have done this in #4825, so perhaps you could rebase this PR on top of that so that we at least get correct failures.

(Of course, it would be better if the test _itself_ generated the failure, but I don't see anything wrong with adding such a check to the `actions/upload-artifact` steps too.)

wilfwilson:
It seems that that worked; the job now fails.

fingolfin:
Updated, should work now. Savings are still pretty good

ChrisJefferson (approved):

PR #4828, opened 2022-03-20
buildsys: switch to new package distribution

fingolfin:
The new package distribution we started during the recent GAP Days is basically ready for use. I am not saying everything is done in terms of the automation. But I think it is sufficiently good for us to use. And it already contains dozens of package updates not in the current / "old" distribution.

Also please feel free to have a look at https://github.com/gap-system/PackageDistro for information on how things work. Feedback, suggestions and help are welcome.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -1233,9 +1233,9 @@ testkernel: ${KERNELTESTS}
 
 # change PKG_BRANCH to stable-X.Y in the stable branch
 PKG_BRANCH = master
-PKG_BOOTSTRAP_URL = https://github.com/gap-system/gap-distribution/releases/download/package-archives/
-PKG_MINIMAL = packages-required-$(PKG_BRANCH).tar.gz
-PKG_FULL = packages-$(PKG_BRANCH).tar.gz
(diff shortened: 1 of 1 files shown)


ChrisJefferson (approved):

alex-konovalov (commented):
Great! As I can see, from now on we don't have separate archives for stable branches? How will we deal with situations when a new package added, or a new release of a package published, which requires features only in master? Is the plan to introduce more merged package archives later, or to require package authors to be more careful about compatibility?

fingolfin:
So far the plan was that the package distro is always compatible with both the latest GAP release, and master, and so we only use a single distro.

But perhaps this was naive; e.g. it doesn't deal with the possibility that some package authors may wish to have new versions of their packages included with 4.12.0 that require 4.12. Indeed, the latest `matgrp` release requires GAP 4.12 and hence [has not yet been merged into the distro](https://github.com/gap-system/PackageDistro/pull/227).

alex-konovalov:
I see... That could probably still be addressed with on the packages level - e.g. loading some "bleeding edge" code conditionally?

PR #4830, opened 2022-03-20
Fix infinite recursion in `AlternatingDegree` and `SymmetricDegree` when the given 

### LINKED ISSUE #4826: Infinite recursion in AlternatingDegree of a non-natural alternating group constructed with AlternatingGroup
### Observed behaviour

```
gap> A := AlternatingGroup( IsPcGroup, 4 );;
gap> AlternatingDegree( A );
Error, recursion depth trap (5000) in
  AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
AlternatingDegree( G ) at /usr/lib/gap/lib/grpnames.gi:1205 called from
...  at *stdin*:16
```
The problem seems to lie here:
```
gap> HasIsAlternatingGroup( A ) and IsAlternatingGroup( A );
true
gap> IsNaturalAlternatingGroup( A );
false
gap> Print( ApplicableMethod( AlternatingDegree, [A] ) );
function ( G )
    if not IsFinite( G ) then
        TryNextMethod();
    fi;
    if IsNaturalAlternatingGroup( G ) then
        return DegreeAction( G );
    fi;
    if IsAlternatingGroup( G ) then
        return AlternatingDegree( G );
    else
        return fail;
    fi;
    return;
end
```
So, if a group has `IsAlternatingGroup` set to `true` but does not have the  `AlternatingDegree` attribute set, we get an infinite recursion.
However, if the group doesn't have `IsAlternatingGroup` set yet, calling it will also set the `AlternatingDegree` attribute, avoiding this bug:
```
gap> B := AlternatingGroup( IsPermGroup, 4 );;
gap> C := Image( IsomorphismPcGroup( B ) );;
gap> HasIsAlternatingGroup( C ) or HasAlternatingDegree( C );
false
gap> IsAlternatingGroup( C ) and HasAlternatingDegree( C );
true
gap> AlternatingDegree( C );
4
```

### Expected behaviour
Ideally, we would get the alternating degree.
```
gap> A := AlternatingGroup( IsPcGroup, 4 );;
gap> AlternatingDegree( A );
4
```

```
 ┌───────┐   GAP 4.12dev-1220-g516e42e built on 2022-03-09 15:58:57+0100
 Packages:   GAPDoc 1.6.5, PrimGrp 3.4.1, SmallGrp 1.4.2, TransGrp 3.3
```
The same error is already present in GAP 4.9.3.

fingolfin:
Fixes #4826

Not for the release notes, as these functions are undocumebted

### DIFF
--- a/tst/testbugfix/2022-03-20-AlternatingDegree.tst
+++ b/tst/testbugfix/2022-03-20-AlternatingDegree.tst
@@ -1,0 +1,15 @@
+# Very fix for infinite recursion in AlternatingDegree and SymmetricDegree,
+# see https://github.com/gap-system/gap/issues/4826
+gap> TweakGroup:=function(G)
+>   local H;
+>   H := Action(G,Orbit(G,[1,2],OnTuples),OnTuples);
+>   SetIsAlternatingGroup(H,IsAlternatingGroup(G));
+>   SetIsSymmetricGroup(H,IsSymmetricGroup(G));
+>   return H;
+> end;;
+
+#
+gap> List([1..6], n -> AlternatingDegree(TweakGroup(AlternatingGroup(n))));
+[ 0, 0, 3, 4, 5, 6 ]
+gap> List([1..6], n -> SymmetricDegree(TweakGroup(SymmetricGroup(n))));
+[ 0, 2, 3, 4, 5, 6 ]
--- a/lib/grpnames.gi
+++ b/lib/grpnames.gi
@@ -1202,8 +1202,14 @@ InstallMethod( AlternatingDegree,
   function ( G )
     if not IsFinite(G) then TryNextMethod(); fi;
     if IsNaturalAlternatingGroup(G) then return DegreeAction(G); fi;
-    if IsAlternatingGroup(G) then return AlternatingDegree(G);
-                             else return fail; fi;
+    if not IsAlternatingGroup(G) then return fail; fi;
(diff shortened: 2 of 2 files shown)


tst/testbugfix/2022-03-20-AlternatingDegree.tst
```diff
@@ -0,0 +1,15 @@
+# Very fix for infinite recursion in AlternatingDegree and SymmetricDegree,
+# see https://github.com/gap-system/gap/issues/4826
+gap> TweakGroup:=function(G)
+>   local H;
+>   H := Action(G,Orbit(G,[1,2],OnTuples),OnTuples);
+>   SetIsAlternatingGroup(H,IsAlternatingGroup(G));
+>   SetIsSymmetricGroup(H,IsSymmetricGroup(G));
+>   return H;
+> end;
```
fingolfin, line 9:
```suggestion
> end;;
```

ChrisJefferson (approved):

PR #4831, opened 2022-03-20
CI: use ccache to speed up compilation, esp. of packages

fingolfin:
For the PackageDistro, this reduced the time to compile GAP package binaries by several minutes in case of a cache hit (i.e. all the time unless the package is updated in the package distribution).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -217,6 +217,13 @@ jobs:
                fi
                python -m pip install gcovr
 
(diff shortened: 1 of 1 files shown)


ChrisJefferson (approved):

fingolfin:
I am still not 100% sure this works as intended here, but my attempts to test it failed (this is due to how the GH caches work). But it shouldn't hurt either... So I'll just merge this, and we can then watch how it affects build time, if at all.

PR #4835, opened 2022-03-22
OrbitStabilizerAlgorithm: use GroupWithGenerators

fingolfin:
... instead of Group. This allows code which uses "special" elements as
`acts` to keep working, i.e., objects which de facto technically form a
group under multiplication but are for some reason or another not in the
filter `IsGeneratorsOfMagmaWithInverses`.

### DIFF
--- a/lib/oprt.gi
+++ b/lib/oprt.gi
@@ -1408,7 +1408,7 @@ local pr,hom,pgens,pacts,pdopr,erg,cs,i,dict,orb,stb,rep,getrep,q,img,gen,
  
     pr:=PerfectResiduum(G);
     if IndexNC(G,pr)>3 then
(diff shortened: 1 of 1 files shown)


hulpke (approved):
This is of course fine, but there might be other places in similar situations that wil cause errors.

Is the only issue to skip the `IsGeneratorsOfMagmaWithInverses` test? If so, would it possibly be cleaner to add an `:`option that will cause the call from `Group` to `IsGeneratorsOfMagmaWithInverses` to be skipped?

fingolfin:
Actually I realized that `IsGeneratorsOfMagmaWithInverses` is a property, not a filter, and so I was able to just install a new method `IsGeneratorsOfMagmaWithInverses` as an alternate workaround.

PR #4838, opened 2022-03-24
CI: tweak ccache integration further

fingolfin:


ChrisJefferson:
Can you explain "override_cache_key_fallback" ? Google returns no results(!)

.github/workflows/CI.yml
```diff
@@ -225,7 +225,9 @@ jobs:
         uses: Chocobo1/setup-ccache-action@v1
         with:
           update_packager_index: false
-          override_cache_key: setup-ccache_${{ runner.os }}_${{ hashFiles('extra.flags') }}_${{ github.ref }}
+          override_cache_key: ${{ runner.os }}-${{ hashFiles('extra.flags') }}-${{ github.ref }}
+          override_cache_key_fallback:
```
fingolfin, line 229: @ChrisJefferson to learn more about a GitHub action, usually the best way is to visit its repository, i.e., https://github.com/Chocobo1/setup-ccache-action where you can search for the more info

fingolfin: Roughly speaking, this action uses another action https://github.com/actions/cache and this fallback key controls the "fallback" key of the actions/cache. This in turn is crucial to make these cache useful, as caches are actually immutable (once created, they can't be modified, although a GC will eventually discard them). The fallback then does this: if there is no cache yet with the given key (which of course is the the case the first time a workflow runs for a new PR), it will try to use a different cache; namely any cache with an idea that starts with the fallback. So here the idea is to fallback to something which has the same OS and flags, but using a different ref.

fingolfin: The first time around, that other cache is used to for retrieval; then we update it (because ccache might store new cache data), and at the end of a successful run of the workflow this new modified data is stored again into the cache. If we re-run the workflow for the same PR, it will now find a cache with the exact primary key, and use that (but it won't be modified afterawrds, as it is immutable).

ChrisJefferson (approved):

fingolfin:
But the CI tests should run. I'll try again...

.github/workflows/CI.yml
fingolfin: Apparently line 230 has a syntax error. Need to figure that one out *sigh*

PR #4845, opened 2022-03-30
Use new package distro in stable-4.11

fingolfin:
I think this should work.

However, it made me realize we have one thing to take care of: how to ensure old stable-4.X branches keep working. Let me explain: the plan right now is this: the package distro always aims to support two GAP versions:
1. the latest GAP release (and by extension, the latest commit on the corresponding `stable-X.Y` branch)
2. the latest GAP master branch

So right now that means supporting GAP master, stable-4.11 and v4.11.1

Once we release v4.12.0, though we will start merging packages that may require GAP 4.12. Indeed, we already have such an updated lined up: the new `matgrp` package already requires GAP 4.12 (see https://github.com/gap-system/PackageDistro/pull/227).

Now if all GAP branches (master, stable-*) use https://github.com/gap-system/PackageDistro/releases/download/latest then as soon as we start to merge stuff that requires 4.12, we start to cause trouble for the stable-4.11 branch.

So really at that point, we want stable-4.11 to stay at whatever package distro last was compatible with it. To achieve that, we could switch to using a separate tag, say `latest-4.11`. That way, we could even keep modifying the distro beyond that date.

Some thoughts for dealing with this:
1. we create that tag now and use it here. But then we also need to update that tag (either manually, or by adding to the scripts -- the latter is not hard but yet more work I'd rather avoid)
2. we postpone the creation until we need it. Very easy, but also easy to forget (but not too bad -- someone will complain at some point and then we can create that tag and update the repository here)
3. we add a level of indirection: instead of hardcoding the URL resp. tag name here, we could fetch it from a server (that doesn't have to be hard: basically a file on a website somewhere with the tag name; I could expand on the finer details, but maybe that's not necessary right now). Anyway, very flexible, but maybe overengineered...

All in all, I am leaning towards option 2, with the obvious note that of course we can revisit this at any time.


alex-konovalov:
@fingolfin CI fails:
```
+ rm -rf 'profiling-*'
+ git clone https://github.com/gap-packages/profiling
fatal: destination path 'profiling' already exists and is not an empty directory.
```

fingolfin:
Fixed the `profiling` problem. Also fixed the macOS build failure (use of `vfork` triggers a compiler warning there these days, which we treat as an error; so I backported 90ba7f47e66ed8e828d4b471b4d28a02335ad805 which removed it from `master` in 2019...

Alas, `testpackages` still fails because a long list of packages fails to even load. E.g. the `Gauss` package -- which is weird, because its only dependency is `GAPDoc`, and even though it has a compiled part, it can be loaded without it. Huh?

fingolfin:
OK, the packages fail to load because of this:
```
#I  4ti2Interface: replace prefix '4ti2int' by the unique completion '4ti2Interface'
#I  4ti2Interface: entering LoadPackage
#I  4ti2Interface: PackageAvailabilityInfo for version 2022.03-01
#I  (unknown package): PackageAvailabilityInfo: required GAP version (>= 4.11.1) does not fit4ti2Interface
```

Indeed, the version on this branch was still set to 4.11.0. I've now added a commit to fix this.

ChrisJefferson (approved):

PR #4853, opened 2022-04-06
Remove a test for data in the small groups library

fingolfin:
This test started to fail with the release of SmallGrp 1.5, as that corrects
the number of groups of order 1024 which was previously misrecorded.

I could have adjusted the test here, but instead I removed it, as SmallGrp
has its own test suite which already contains similar tests, so there really
is no point in repeating these checks here.


FriedrichRober (approved):
As this is very similar to this [test file in `smallgrp`](https://github.com/gap-packages/smallgrp/blob/master/tst/small_groups.tst), I see no problem with removing this test from here.

wucas (approved):

danielrademacher (approved):

AnnaKDS (approved):

fingolfin:
Thank you all :-)

PR #4855, opened 2022-04-07
Update trivial case for SubgroupsSolvableGroup with retnorm

RamonEstebanRomero:
The function `SubgroupsSolvableGroup` with a trivial group with an argument does not return the list of normalizers when called with the option `retnorm`. I suggest considering this flag for trivial groups. I propose moving first the parsing of the `retnorm` flag and immediately checking the trivial case. Please check I have not destroyed the rest of the function!

## Text for release notes

See title (or leave it up to you)

### DIFF
--- a/lib/grppclat.gi
+++ b/lib/grppclat.gi
@@ -570,9 +570,6 @@ local g,	# group
       xo;	# xternal orbits
 
   g:=arg[1];
-  if Size(g)=1 then
-    return [g];
-  fi;
   if Length(arg)>1 and IsRecord(arg[Length(arg)]) then
     opt:=arg[Length(arg)];
   else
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Hi!

Thanks for the PR. Would you consider writing some tests for this function? It looks to me like there are currently no tests at all :(

Obviously within reason the more tests the better! But it's also fine to just put in some basic tests, and also some tests which cover the case you are changing. If you want some help writing tests please do ask, but as a quick suggestion, I'd look at some files in `tst/opers`, to get a feeling for what the tests should look like.

hulpke (approved):
Yes, I never considered the case of the trivial group in these cases, this should fix it.

fingolfin:
I tweaked the formatting and commit message and added a test case.

PR #4856, opened 2022-04-10
Improve `Intersection` for permutation groups

MathieuDutSik:
The code does make a small improvement to the Intersection of permutation groups.
The code in the master right now is replacing G with its stabilizer for the difference
Moved(G) \ Moved(H) and the same for H. But the thing is once you do this, the sets
Moved(G) and Moved(H) change which gives new opportunities for reducing the
groups.

### DIFF
--- a/lib/stbcbckt.gi
+++ b/lib/stbcbckt.gi
@@ -2974,20 +2974,15 @@ local   Omega, P, rbase, L, mg, mh, IsFinished, mg_minus_mh, mh_minus_mg;
         return TrivialGroup(IsPermGroup);
       fi;
 
-      IsFinished:=true;
       mg_minus_mh:=Difference(mg,mh);
       if Length(mg_minus_mh) > 0 then
         G:=Stabilizer(G,mg_minus_mh,OnTuples);
-        IsFinished:=false;
+        continue;
       fi;
       mh_minus_mg:=Difference(mh,mg);
       if Length(mh_minus_mg) > 0 then
         H:=Stabilizer(H,mh_minus_mg,OnTuples);
-        IsFinished:=false;
-      fi;
-
-      if IsFinished then
-        break;
+        continue;
       fi;
     od;
 
--- a/lib/stbcbckt.gi
+++ b/lib/stbcbckt.gi
@@ -2955,7 +2955,7 @@ end );
 InstallMethod( Intersection2, "perm groups", IsIdenticalObj,
   [ IsPermGroup, IsPermGroup ], 0,
 function( G, H )
-local   Omega, P, rbase, L, mg, mh, IsFinished, mg_minus_mh, mh_minus_mg;
+local   Omega, P, rbase, L, mg, mh, mg_minus_mh, mh_minus_mg;
 
     if IsIdenticalObj( G, H )  then
       return G;
--- a/lib/stbcbckt.gi
+++ b/lib/stbcbckt.gi
@@ -2955,24 +2955,41 @@ end );
 InstallMethod( Intersection2, "perm groups", IsIdenticalObj,
   [ IsPermGroup, IsPermGroup ], 0,
 function( G, H )
-local   Omega,  P,  rbase,  L,mg,mh,i;
-    
+local   Omega, P, rbase, L, mg, mh, IsFinished, mg_minus_mh, mh_minus_mg;
+
     if IsIdenticalObj( G, H )  then
       return G;
     fi;
-    
-    # align the acting domains
-    mg:=MovedPoints(G);
-    mh:=MovedPoints(H);
-    Omega := Intersection(mg,mh);
 
-    # no two points moved in common?
-    if Length(Omega)<=1 then
-      return TrivialSubgroup(Parent(G));
-    fi;
+    # iterate taking stabilizer until G and H have the same set of moved points
+    while(true)
+    do
+      # align the acting domains
+      mg:=MovedPoints(G);
+      mh:=MovedPoints(H);
+      Omega := Intersection(mg,mh);
 
-    G:=Stabilizer(G,Difference(mg,mh),OnTuples);
-    H:=Stabilizer(H,Difference(mh,mg),OnTuples);
+      # no two points moved in common?
+      if Length(Omega)<=1 then
+        return TrivialSubgroup(Parent(G));
+      fi;
+
+      IsFinished:=true;
+      mg_minus_mh:=Difference(mg,mh);
+      if Length(mg_minus_mh) > 0 then
+        G:=Stabilizer(G,mg_minus_mh,OnTuples);
+        IsFinished:=false;
+      fi;
+      mh_minus_mg:=Difference(mh,mg);
+      if Length(mh_minus_mg) > 0 then
+        H:=Stabilizer(H,mh_minus_mg,OnTuples);
+        IsFinished:=false;
+      fi;
+
+      if IsFinished then
+        break;
+      fi;
+    od;
 
     if IsSubset(G,H) then
       return H;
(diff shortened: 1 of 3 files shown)


lib/stbcbckt.gi
```diff
@@ -2955,24 +2955,41 @@ end );
...
+    while(true)
+    do
+      # align the acting domains
+      mg:=MovedPoints(G);
+      mh:=MovedPoints(H);
+      Omega := Intersection(mg,mh);
+
+      # no two points moved in common?
+      if Length(Omega)<=1 then
+        return Group(());
```
ChrisJefferson, line 2974: I agree the "Parent" didn't make a lot of sense here, but I would instead write `TrivialSubgroup(G)`. The reason to do `TrivialSubgroup`, instead of `Group(())`, is that GAP knows the result of `TrivialSubgroup` has size 1 (as the method sets it at construction time), while `Group(())` doesn't set that (as GAP doesn't special-case making an identity group)

lib/stbcbckt.gi
```diff
@@ -2955,24 +2955,41 @@ end );
 InstallMethod( Intersection2, "perm groups", IsIdenticalObj,
   [ IsPermGroup, IsPermGroup ], 0,
 function( G, H )
-local   Omega,  P,  rbase,  L,mg,mh,i;
-    
+local   Omega, P, rbase, L, mg, mh, IsFinished, mg_minus_mh, mh_minus_mg;
```
ChrisJefferson, line 2958: This is minor, but I'd write is_finished instead of IsFinished, just because readers of GAP code will probably think `IsFinished` is an attribute.

ChrisJefferson:
Hi, this looks good, I made a couple of small comments. It would also be nice to add a test or two, just to make sure this new case is being handled correctly (I played around with a couple of examples and believe it does work correctly, but tests never hurt!)

lib/stbcbckt.gi
fingolfin: Actually, it does:
```
gap> HasIsTrivial(Group(()));
true
```
The more salient difference is the set of generators:
```
gap> G:=Group((1,2));; H:=Group((1,2,3));;
gap> GeneratorsOfGroup(Intersection(G,H));
[  ]
gap> GeneratorsOfGroup(TrivialSubgroup(G));
[  ]
gap> GeneratorsOfGroup(Group(()));
[ () ]
```

But note that using `G` as parent is still arbitrary (though harmless). All in all, I would recommend this:
```suggestion
        return TrivialGroup(IsPermGroup);
```
But using `TrivialSubgroup(G)` would also be fine.

lib/stbcbckt.gi
fingolfin: I agree that in general local variables definitely should not start with upper case letters, I only make exceptions for single-letter variables.

Then again, this function uses `Omega` as a local variable name (which I think is a bad idea, esp. as there is a global function `Omega`). So... 🤷 

(BTW, I'd use `isFinished` or `done` or `isdone`... but this is splitting hairs, and I am pretty sure that @ChrisJefferson would agree with me when I see we won't block a PR over such a minuscule "issue" :-) .

lib/stbcbckt.gi
```diff
@@ -2955,24 +2955,41 @@ end );
...
-    mh:=MovedPoints(H);
-    Omega := Intersection(mg,mh);
 
-    # no two points moved in common?
-    if Length(Omega)<=1 then
-      return TrivialSubgroup(Parent(G));
-    fi;
+    # iterate taking stabilizer until G and H have the same set of moved points
+    while(true)
+    do
```
fingolfin, line 2966:
```suggestion
    while true do
```

lib/stbcbckt.gi
```diff
@@ -2955,24 +2955,41 @@ end );
...
+      fi;
+      mh_minus_mg:=Difference(mh,mg);
+      if Length(mh_minus_mg) > 0 then
+        H:=Stabilizer(H,mh_minus_mg,OnTuples);
+        IsFinished:=false;
+      fi;
+
+      if IsFinished then
+        break;
+      fi;
```
fingolfin, line 2991: How about this instead
```suggestion
      mg_minus_mh:=Difference(mg,mh);
      if Length(mg_minus_mh) > 0 then
        G:=Stabilizer(G,mg_minus_mh,OnTuples);
        continue;
      fi;
      mh_minus_mg:=Difference(mh,mg);
      if Length(mh_minus_mg) > 0 then
        H:=Stabilizer(H,mh_minus_mg,OnTuples);
        continue;
      fi;
```
This way, we compute one stabilizer, then immediately recheck -- we might already be done, because that one stabilizer could already be trivial, or intersect the moved points of the other group trivially. If not, then the next iteration will compute the other stabilizer.

Plus, you avoid the whole discussion about how to spell `IsFinished` 😂

fingolfin (commented):
Thank you for your contribution, much appreciated! I left some more comments.

One general thing: do you have any concrete examples where this makes a difference? If so, perhaps they could be added as a test case (if not too big).

MathieuDutSik:
> Thank you for your contribution, much appreciated! I left some more comments.
> 
> One general thing: do you have any concrete examples where this makes a difference? If so, perhaps they could be added as a test case (if not too big).

Actually no.
The reason I came to this is that I reimplemented in https://github.com/MathieuDutSik/permutalib
many algorithm from GAP as a C++ library which I need for various purposes.
By doing so, in the process of debugging, I came to see many small issues of the GAP code.

lib/stbcbckt.gi
```diff
@@ -2955,24 +2955,35 @@ end );
 InstallMethod( Intersection2, "perm groups", IsIdenticalObj,
   [ IsPermGroup, IsPermGroup ], 0,
 function( G, H )
-local   Omega,  P,  rbase,  L,mg,mh,i;
-    
+local   Omega, P, rbase, L, mg, mh, IsFinished, mg_minus_mh, mh_minus_mg;
```
fingolfin, line 2958:
```suggestion
local   Omega, P, rbase, L, mg, mh, mg_minus_mh, mh_minus_mg;
```

fingolfin:
I think an example demonstrating the improvement is this:
```
gap> G:=Group( (1,2,3), (4,5,6), (11,12,13), (11,12,14) );;
gap> H:=Group( (1,7,8), (4,9,11), (10,12), (13,14) );;
```
Then we get after "one round" of stabilizers (as the old code did):
```
gap> mg:=MovedPoints(G);
[ 1, 2, 3, 4, 5, 6, 11, 12, 13, 14 ]
gap> mh:=MovedPoints(H);
[ 1, 4, 7, 8, 9, 10, 11, 12, 13, 14 ]
gap> G:=Stabilizer(G,Difference(mg,mh),OnTuples);
Group([ (11,12,13), (11,12,14) ])
gap> H:=Stabilizer(H,Difference(mh,mg),OnTuples);
Group([ (13,14) ])
```
Neither of these is contained in the other, so it went to the partition backtrack code (which of course made short work of this!). 

But if we compute more stabilizers:
```
gap> mg:=MovedPoints(G);
[ 11, 12, 13, 14 ]
gap> mh:=MovedPoints(H);
[ 13, 14 ]
gap> G:=Stabilizer(G,Difference(mg,mh),OnTuples);
Group(())
gap> H:=Stabilizer(H,Difference(mh,mg),OnTuples);
Group([ (13,14) ])
```
And now we see the intersection is trivial.

Whether this is actually faster than what happened before or not: I don't know... This here is just a toy example anyway.

MathieuDutSik:
Thank you. I added this test to **testextra/grpperm.tst**

fingolfin (approved):

fingolfin:
Tests still fail due to a difference in an example, in `doc/ref/debug.xml`. I've pushed a change that *should* address it, we'll see.

To anyone who merges this once the tests pass: please "squash" the changes, no point in having 9 commits for this :-) (alternatively @MathieuDutSik could squash this branch, but this is not required)

PR #4863, opened 2022-04-14
kernel: don't waitpid(-1)

fingolfin:
This causes problems when using GAP as a library. It also
shouldn't be necessary as we record a list of all known child
processes.

### DIFF
--- a/src/iostream.c
+++ b/src/iostream.c
@@ -280,13 +280,6 @@ static void ChildStatusChanged(int whichsig)
     HashUnlock(PtyIOStreams);
 
 #if !defined(HPCGAP)
-    /* Collect up any other zombie children */
-    do {
-        retcode = waitpid(-1, &status, WNOHANG);
-        if (retcode == -1 && errno != ECHILD)
-            Pr("#E Unexpected waitpid error %d\n", errno, 0);
-    } while (retcode != 0 && retcode != -1);
-
     signal(SIGCHLD, ChildStatusChanged);
 #endif
 }


fingolfin:
This *might* help with some issues we are seeing in Julia, CC @benlorenz.

And it might also be totally bogus, because I am forgettinng something important :-). I really help @ChrisJefferson knows / remembers why we are doing this. Was it perhaps done in the past to workaround the fact that `io` did its own child process handling? But then our SIGCHLD handle simply wasn't called anymore, was it?!?

If there *is* a good reason why we need this, then at the very least this should be spelled out in a code comment.

fingolfin:
I traced the origin of this code through the old repository history and arrived at this commit by @stevelinton:
```
Author:     Steve Linton
AuthorDate: 2000-03-15 10:34:46 +0000

    Fixed "Exec on solaris" bug. It turns out that some old UNIX versions
    like to send you another SIGCHLD when you leave a SIGCHLD handler and
    still have zombie children (clue found on Dejanews). Added a loop to
    make sure that we have collected up all zombie children before leaving
    the handler.
                    Steve
```

Based on this I think we should be able to safely remove this workaround for an ancient, long dead Unix variant.

ChrisJefferson:
Just an extra comment, I imagine you know this, but you will have similar problems if the "IO" package's alternative IO handler is installed (as it once again does a `waitpid(-1)` loop. Fixing that might require a better understanding of how Julia handles signal handlers -- I'm happy to help if someone knows generally how Julia handles signals, and how we can hook into that.

ChrisJefferson (approved):

fingolfin:
Yes I am aware of the problem with `io`. Julia does not use a SIGCHLD signal handler itself but rather relies on `libuv` (resp. a heavily modified version of it) to handle child processes...

For now my plan is to first worry about GAP itself: I've already written a pure-Julia replacement for the GAP kernel function `ExecuteProcess`. And I can write similar such functions for `iostream`, at which point the GAP signal handler won't be needed anymore, at least for "pure" GAP code.

But that leaves the `io` package, and that is sadly much harder to deal with, because it is offers these low-level wrappers around POSIX APIs; I mean, there is even `IO_fork()` *and packages use it*, e.g.:
- `profiling` for its test suite (could perhaps be rewritten using some other approach?)
- `JupyterKernel` in a crucial way; but that package is broken anyway and should simply should rewritten from scratch
- `IO_ForHomalg` -> could probably use `IO_Popen`/`IO_Popen2`/`IO_Popen3`
- `vole` -> could probably use `IO_Popen`/`IO_Popen2`/`IO_Popen3`; or ZeroMQ; or rewrite to use a kernel extension; or ...
- ...

So my plan long-term for dealing with `io` for now is to simply disable certain low-level things such as `IO_fork`, `IO_execv*`; and adjust the higher-level code to use replacements written in Julia where possible, e.g. `IO_Popen`/`IO_Popen2`/`IO_Popen3`.

Issue #4865, opened 2022-04-17
Intersection of cosets


MathieuDutSik:
Let us take a group <img src="https://render.githubusercontent.com/render/math?math=G">, two subgroups <img src="https://render.githubusercontent.com/render/math?math=H_1"> and <img src="https://render.githubusercontent.com/render/math?math=H_2"> and <img src="https://render.githubusercontent.com/render/math?math=x_1,x_2\in G"> then the intersection <img src="https://render.githubusercontent.com/render/math?math=x_1H_1\cap x_2H_2"> is either empty of a coset <img src="https://render.githubusercontent.com/render/math?math=z H_1\cap H_2">.

The functionality for cosets in GAP appears deficient on several points.

The first problem is that the intersection is not recognized to be a coset.
```GAP
gap> cos1:=RightCoset(SymmetricGroup([1..6]), (6,7,8));
gap> cos2:=RightCoset(SymmetricGroup([3..8]), (1,2,3));
gap> eInt:=Intersection(cos1, cos2);
[ (1,2,3)(6,7,8), (1,2,3)(5,7,8,6), (1,2,3)(4,5)(6,7,8), (1,2,3)(4,5,7,8,6), (1,2,3)(4,7,8,6,5), (1,2,3)(4,7,8,6), (1,2,3,4)(6,7,8), (1,2,3,4)(5,7,8,6),
  (1,2,3,4,5)(6,7,8), (1,2,3,4,5,7,8,6), (1,2,3,4,7,8,6,5), (1,2,3,4,7,8,6), (1,2,3,5,4)(6,7,8), (1,2,3,5,7,8,6,4), (1,2,3,5)(6,7,8), (1,2,3,5,7,8,6),
  (1,2,3,5)(4,7,8,6), (1,2,3,5,4,7,8,6), (1,2,3,7,8,6,5,4), (1,2,3,7,8,6,4), (1,2,3,7,8,6,5), (1,2,3,7,8,6), (1,2,3,7,8,6,4,5), (1,2,3,7,8,6)(4,5) ]
````

And one can check that indeed this is a coset even if GAP does not recognize it as such.

The second problem related to the first is that the code is horribly slow. For the special case of a permutation group, one would hope to use the same backtracking scheme as the one used for computing the intersection of subgroups that was recently merged, but it is unclear to me how to do this.

MathieuDutSik:
[...]

PR #4874, opened 2022-04-20
Speed up computation of the intersection of group cosets in the generic cases and even more so for permutation groups

MathieuDutSik:
The intersection of cosets in the gap code is right now inadequate. The cosets are considered as sets which means that we do not have the key information that it is either empty or a closet in the output.

The PR addresses that issue. For general groups we write the intersection as  and we simply iterate over the right cosets of  with respect to .

For permutation groups, we do a number of preprocessing tricks in order to simplify and/or resolve the problem. But in the end, we still have to iterate over the cosets. There are possibilities of further improvements, but I think this is a good first step.

#4865


hulpke (approved):
Thank you. Looks very good to me!

lib/csetgrp.gi
```diff
@@ -788,7 +788,34 @@ local a,r;
...
+        else
+            return [];
+        fi;
+    fi;
+    shift:=x1;
+    sigma:=x2 * Inverse(x1);
+    TheInt:=Intersection(H1, H2);
+    CosTest:=RightCoset(H2, sigma);
+    for eCos in RightCosets(H1, TheInt)
+    do
```
fingolfin, line 811:
```suggestion
    for eCos in RightCosets(H1, TheInt) do
```

lib/csetgrp.gi
```diff
@@ -788,7 +788,34 @@ local a,r;
   TryNextMethod();
 end);
 
-
+InstallMethod(Intersection2, "general groups", IsIdenticalObj,
+              [IsRightCoset,IsRightCoset],
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, TheInt, CosTest, eCos, rho;
```
fingolfin, line 794: local variables usually should start with a lower case latter (we make exceptions for things like `G` or `H1`, but that's usually about it). So `theInt` or `the_int` instead of `TheInt` etc.

lib/csetgrp.gi
```diff
@@ -788,7 +788,34 @@ local a,r;
...
+            return cos1;
+        else
+            return [];
+        fi;
+    fi;
+    shift:=x1;
+    sigma:=x2 * Inverse(x1);
+    TheInt:=Intersection(H1, H2);
+    CosTest:=RightCoset(H2, sigma);
+    for eCos in RightCosets(H1, TheInt)
```
fingolfin, line 810: Why `eCos`? (Not asking it to be changed, just wondering)

lib/csetperm.gi
```diff
@@ -548,3 +548,151 @@ function(a,b)
...
+  [IsRightCoset and IsPermCollection,IsRightCoset and IsPermCollection],0,
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, LMoved_H1, LMoved_H2, LMoved_H12, LMoved_sigma, TheInt, set2, set1, eRepr, set2_img, set1_img, H1_sigma, H2_sigma, test, GRPtot, swap, eCos, rho, CosTest, diff12, diff21, fset1, fset2;
+    # We set int = cos1 cap cos2 = H1 x1 cap H2 x2
+    H1:=ActingDomain(cos1);
+    H2:=ActingDomain(cos2);
+    x1:=Representative(cos1);
+    x2:=Representative(cos2);
+    shift:=x1;
+    sigma:=x2 * Inverse(x1);
```
fingolfin, line 564:
```suggestion
    sigma:=x2 / x1;
```

lib/csetperm.gi
```diff
@@ -548,3 +548,151 @@ function(a,b)
...
+            #     = (H1 sigma^{-1} \cap H2) sigma shift
+            #     = (stab(H1) repr sigma^{-1} \cap H2) sigma shift
+            #     = (stab(H1) \cap H2 sigma repr^{-1} )    repr shift
+            set1_img:=OnTuples(set1, sigma);
+            eRepr:=RepresentativeAction(H1, set1, set1_img, OnTuples);
+            if eRepr=fail then
+                return [];
+            fi;
+            H1:=Stabilizer(H1, set1, OnTuples);
+            sigma:=sigma * Inverse(eRepr);
```
fingolfin, line 611:
```suggestion
            sigma:=sigma / eRepr;
```

lib/csetperm.gi
```diff
@@ -548,3 +548,151 @@ function(a,b)
...
+            H1:=Stabilizer(H1, fset1, OnTuples);
+            continue;
+        fi;
+        fset2:=Difference(LMoved_H2, Union(LMoved_H1, LMoved_sigma));
+        if Length(fset2) > 0 then
+            H2:=Stabilizer(H2, fset2, OnTuples);
+            continue;
+        fi;
+        # More general but more expensive than previous check
+        H1_sigma:=Group(Concatenation(GeneratorsOfGroup(H1), [sigma]));
```
fingolfin, line 636:
```suggestion
        H1_sigma:=ClosureGroup(H1, sigma);
```
Similarly elsewhere

lib/csetperm.gi
```diff
@@ -548,3 +548,151 @@ function(a,b)
...
+        H2_sigma:=Group(Concatenation(GeneratorsOfGroup(H2), [sigma]));
+        if IsSubgroup(H2_sigma, H1) = false then
+            H1:=Intersection(H2_sigma, H1);
+            continue;
+        fi;
+        # No more reduction tricks available
+        break;
+    od;
+    # A final termination criterion
+    GRPtot:=Group(Concatenation(GeneratorsOfGroup(H1), GeneratorsOfGroup(H2)));
```
fingolfin, line 650:
```suggestion
    GRPtot:=ClosureGroup(H1, H2);
```

lib/csetperm.gi
```diff
@@ -548,3 +548,151 @@ function(a,b)
...
+    test:=sigma in GRPtot;
+    if test = false then
```
fingolfin, line 652:
```suggestion
    if not sigma in GRPtot then
```

fingolfin (changes requested):
Thank you, much appreciated! I did not yet have time to check this out in detail, but it looks promising

Unfortunately most of the new code is not yet covered by our code coverage, which reduces confidence in it a bit. But on the upside, I see you already added tests, which simply are not yet run. So if you can move them as indicated by my comments elsewhere, we can see if that rectifies the issue :-)

lib/csetgrp.gi
MathieuDutSik: I have a tendency to use names like **eCos**, **fCos** or **cos1**.
That is my style. I can change if that is a problem.

tst/testextra/grpperm.tst
```diff
@@ -112,6 +112,15 @@ gap> Intersection(Group( (1,2,3), (4,5,6), (11,12,13), (11,12,14) ),
 Group(())
 gap> Intersection(Group((1,2), (3,4)), Group((3,4),(5,6)));
 Group( (3,4) )
+gap> Intersection(RightCoset(Group([ (1,2,3,4,5,6,7), (5,6,7) ]),(3,6)(4,7)),
+                  RightCoset(Group([ (1,2,3,4,5,6,8), (1,3,2,6,4,5), (1,6)(2,3)(4,5)(7,8) ]),(1,7,6,8,3,5)));
```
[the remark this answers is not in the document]
MathieuDutSik: Thanks. I did not realize this problem. So yes I should move those tests. But I realize there are several problems:
* I added the tests for Intersection of permutation groups (merged a few days ago) in testextra. So, I should move them to another file in testinstall. Because the same issue applies here.
* The RightCoset has almost no tests related to it. I think I should create a cset.tst in testinstall. 

Shall this be part of the same PR? Let me know what you think.

lib/csetgrp.gi
MathieuDutSik: Ok, I renamed the variables according to the standard.

MathieuDutSik:
Regarding checks, I implemented the function as a separate user function before putting it into the cset.gi and csetperm.gi files. I tested it on 10000 pairs of groups chosen at random with about 100 intersections being non-trivial. The obtained cosets considered as sets were compared with the list of permutations being created in the master code.

I put the two cases that showed up when debugging it in testextra (see my other remark) but yes I can put them elsewhere and add more tests as requested.

tst/testextra/grpperm.tst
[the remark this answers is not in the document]
fingolfin: Ideally, yes, this should be part of this PR, so that we can see from the Codecov reports which parts of the code are exercised.

Feel free to add a new file `tst/testinstall/coset.tst` or so and put your tests in there. There should ideally be also tests for (cosets of) groups which are not permutation groups, so that the code for those is also tested

fingolfin (commented):
Some more remarks. Still not done reading all, but I have to leave now and figured this is already something

Thank you once more for doing this.

lib/csetgrp.gi
```diff
@@ -788,7 +788,33 @@ local a,r;
   TryNextMethod();
 end);
 
-
+InstallMethod(Intersection2, "general groups", IsIdenticalObj,
+              [IsRightCoset,IsRightCoset],
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, theInt, cosTest, eCos, rho;
```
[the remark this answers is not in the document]
MathieuDutSik: I am concerned about the cases where this algorithm applies. I think there can be cases where H1 and H2 are infinite and the index is finite and that algo still works.

Does your corrections works in that case? The comparison with index also makes sense for infinite groups. But it is not sure to me how far we can go for the case of general groups as not everything is available for them.

lib/csetperm.gi
```diff
@@ -548,3 +548,150 @@ function(a,b)
...
+    # --- The naive algorithm for computing H1 \cap H2 sigma is to iterate
+    # over elements of H1 and testing if one belongs to H2 sigma. If we find
+    # one such z then the result is the coset RightCoset(theInt, z). If not
+    # then it is empty.
+    # --- Since the result depends is independent of the cosets theInt, what
+    # we can do is iterate over the RightCosets(H1, theInt). The algorithm is
+    # the one of Proposition 3.2
+    # for r in RightCosets(H1, theInt)
+    # do
+    #   if r in H1 sigma then
```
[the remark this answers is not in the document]
MathieuDutSik: Yes, sorry for that.

lib/csetperm.gi
```diff
@@ -548,3 +548,150 @@ function(a,b)
...
+    # We are now inspired by the algorithm from
+    # Lazlo Babai, Coset Intersection in Moderately Exponential Time
+    #
+    # We use the algorithm from Page 10 of coset analysis and we reformulate
+    # it here in order to avoid errors:
+    # --- The naive algorithm for computing H1 \cap H2 sigma is to iterate
+    # over elements of H1 and testing if one belongs to H2 sigma. If we find
+    # one such z then the result is the coset RightCoset(theInt, z). If not
+    # then it is empty.
+    # --- Since the result depends is independent of the cosets theInt, what
```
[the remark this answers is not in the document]
MathieuDutSik: Corrected.

lib/csetperm.gi
```diff
@@ -548,3 +548,150 @@ function(a,b)
...
+    # we can do is iterate over the RightCosets(H1, theInt). The algorithm is
+    # the one of Proposition 3.2
+    # for r in RightCosets(H1, theInt)
+    # do
+    #   if r in H1 sigma then
+    #     return RightCoset(theInt, r shift)
+    #   fi;
+    # od;
+    # --- The question is how to use AscendingChains between theInt and H1.
+    # The iteration over the conjugacy classes. We need to understand
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, the comments has been reformulated into something clearer.

lib/csetgrp.gi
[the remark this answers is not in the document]
fingolfin: All bets are off when the cosets are infinite. Whether anything will work in that case will depend largely on what is implemented and how for those groups.

That said, the above check should still work fine: if both groups are infinite, it won't ever swap, and if one is finite and one is infinite, it will ensure that the finite one comes first. Sure, one could extend this to use the index of the intersection to handle things "even better", but just computing the intersection of `H1` and `H2` may already be infeasible in that case -- so it shouldn't be done before at least handling basic cases like checking whether `cos1 = cos2` etc.

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,43 @@
...
+gap> Intersection(SymmetricGroup([1..5]),SymmetricGroup([3..8]));
+Sym( [ 3 .. 5 ] )
+gap> Intersection(SymmetricGroup([1..5]),AlternatingGroup([3..8]));
+Alt( [ 3 .. 5 ] )
+gap> Intersection(AlternatingGroup([1..5]),AlternatingGroup([3..8]));
+Alt( [ 3 .. 5 ] )
+gap> Intersection(AlternatingGroup([1..5]),SymmetricGroup([3..8]));
+Alt( [ 3 .. 5 ] )
+gap> Intersection(Group( (1,2,3), (4,5,6), (11,12,13), (11,12,14) ),
+                  Group( (1,7,8), (4,9,11), (10,12), (13,14) ));
```
fingolfin, line 15: That's not valid syntax for a `.tst` file, and hence they fail in CI.
``suggestion
>                 Group( (1,7,8), (4,9,11), (10,12), (13,14) ));
```

Have you tested them locally? Assuming you started GAP from the root of the git repo, you can do so via `Test("tst/testinstall/cset.tst");`. As an advanced feature, you can use `Test("tst/testinstall/cset.tst", rec(rewriteToFile:=true));` to update the .tst file (this will update the "outputs" in the .tst file to match whatever GAP actually outputs when reading the "inputs". So this is not good to deal with issues in the inputs, as here, but it can be helpful when the outputs need tweaking)

File: tst/testinstall/cset.tst [hunk 15]
```diff
@@ -0,0 +1,43 @@
...
+gap> IsRightCoset(RightCoset(MathieuGroup(12), (1,2,3)));
+true
+gap> Intersection(RightCoset(Group([ (1,2,3,4,5,6,7), (5,6,7) ]),(3,6)(4,7)),
+                  RightCoset(Group([ (1,2,3,4,5,6,8), (1,3,2,6,4,5), (1,6)(2,3)(4,5)(7,8) ]),(1,7,6,8,3,5)));
+RightCoset(Group([ (2,6,7)(3,5,4), (1,2,4)(3,6,5) ]),(1,3,7,5)(4,6))
+gap> Intersection(RightCoset(Group([ (1,4)(2,5), (1,3,5)(2,4,6), (1,5)(2,4)(3,6) ]),(1,7,6,5)(3,4,8)),
+                  RightCoset(Group([ (3,4), (5,6,7,8), (5,6) ]),(1,8,6,2)(3,7)));
+RightCoset(Group(()),(1,8,3,4,7,6,5,2))
+gap> Intersection(RightCoset(Group([ [ [ -1, 0 ], [ 0, 1 ] ], [ [ 0, 1 ], [ 1, 0 ] ] ]), IdentityMat(2)),
+                  RightCoset(Group([ [ [ -1, 0 ], [ 0, -1 ] ] ]),[[0,1],[1,0]]));
```
fingolfin, line 42:
```suggestion
>                 RightCoset(Group([ (1,2,3,4,5,6,8), (1,3,2,6,4,5), (1,6)(2,3)(4,5)(7,8) ]),(1,7,6,8,3,5)));
RightCoset(Group([ (2,6,7)(3,5,4), (1,2,4)(3,6,5) ]),(1,3,7,5)(4,6))
gap> Intersection(RightCoset(Group([ (1,4)(2,5), (1,3,5)(2,4,6), (1,5)(2,4)(3,6) ]),(1,7,6,5)(3,4,8)),
>                 RightCoset(Group([ (3,4), (5,6,7,8), (5,6) ]),(1,8,6,2)(3,7)));
RightCoset(Group(()),(1,8,3,4,7,6,5,2))
gap> Intersection(RightCoset(Group([ [ [ -1, 0 ], [ 0, 1 ] ], [ [ 0, 1 ], [ 1, 0 ] ] ]), IdentityMat(2)),
>                 RightCoset(Group([ [ [ -1, 0 ], [ 0, -1 ] ] ]),[[0,1],[1,0]]));
```

ChrisJefferson:
I had a quick test comparing this to Vole, and once I fixed some typos (U -> theInt, shift -> x1, LMoved -> listMoved), they seem to produce the same answers.

I tested the generic method by temporarily forcing GAP to use it for rightcosets of permutations by editing the code, I'm not sure of the best way of testing it in general.

fingolfin:
I took the liberty of rebasing, squashing and fixing things. Let's see how CI fares (modulo the failures in the `testpackages` jobs, see issue #4878

tst/testinstall/cset.tst
```diff
@@ -42,7 +42,7 @@ true
 # test intersecting cosets
 gap> Intersection(RightCoset(Group([ (1,2,3,4,5,6,7), (5,6,7) ]),(3,6)(4,7)),
 >                 RightCoset(Group([ (1,2,3,4,5,6,8), (1,3,2,6,4,5), (1,6)(2,3)(4,5)(7,8) ]),(1,7,6,8,3,5)));
-RightCoset(Group([ (2,6,7)(3,5,4), (1,2,4)(3,6,5) ]),(1,3,7,5)(\
```
ChrisJefferson: While this isn't how GAP tests are often written, I'd suggest changing this to:

```
gap> Intersection(...) = RightCoset(..);
true
```

As that is more resiliant to these annoying types of changes.

ChrisJefferson: you don't have to do this, only if it makes it easier for you to write / update tests.

MathieuDutSik: ok, will do that.
Maybe we could parse the result of the runs and check them. I had that issue when running C++ vs GAP, because of gap formatting, I had to postprocess the output in order for the comparison to match.

MathieuDutSik:
> I had a quick test comparing this to Vole, and once I fixed some typos (U -> theInt, shift -> x1, LMoved -> listMoved), they seem to produce the same answers.
> 
> I tested the generic method by temporarily forcing GAP to use it for rightcosets of permutations by editing the code, I'm not sure of the best way of testing it in general.

Thank you for correcting the typos! Those stylistic correction create a lot of mess.
Besides I did check on 10000 random cases by comparing the intersection of sets and
no error was found.

MathieuDutSik:
> I took the liberty of rebasing, squashing and fixing things. Let's see how CI fares (modulo the failures in the `testpackages` jobs, see issue #4878

Now, the CI pass. Let me know if something else prevents merging the PR.

fingolfin (changes requested):
I have two main concerns with this PR (all the other comments really are optional suggestion):
1. all these repeated computations of closure groups and intersections aren't cheap. I wonder esp. for smaller examples whether they aren't slower than what your "naive" method above does, or even a naive intersection? Perhaps if one of the cosets has less than, say, 50 elements, it would be faster to fall back to a brute force search? As such: did you perform any performance timings to check that this really is an improvement in all cases, or at least does not slow down things too much?
2. Why is all this correct? There is one longish comment explaining some stuff and referencing a paper by Babai. But a ton of "criterion" come before it, but without any explanations as to why they are correct. Granted, it may not be hard to figure this out with some thinking for each. But if we ever need to a fix a bug in this case we'd have to slowly walk through all of this, to understand what's going on. Some comments would do wonders. Or at least some tests that trigger each case -- but right now most (all?) of them are not covered by a test at all, as Codecov shows.

Now, don't get me wrong, I am really grateful you took the effort to prepare this. Also, a lot of existing GAP code does not match this quality criterion (which I think is really bad). But to merge this, I really need to be convinced this (a) an improvement and (b) correct.

Oh, and it also feels a bit to me as if this code is redoing a lot of what Partition backtrack does. As such, I wonder if our partition backtrack code doesn't have a way to find one element of a coset intersection? @ChrisJefferson do you know? Doesn't ferret have such a thing? What about vole?

If I find the time, I'll try to read through more of the code and suggest e.g. possible code comments that I feel would explain things.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+    x1:=Representative(cos1);
+    x2:=Representative(cos2);
+    shift:=x1;
+    sigma:=x2 / x1;
+    # Initial very easy check
+    listMoved_H1:=MovedPoints(H1);
+    listMoved_H2:=MovedPoints(H2);
+    listMoved_H12:=Union(listMoved_H1, listMoved_H2);
+    listMoved_sigma:=MovedPoints(sigma);
+    if not IsSubset(listMoved_H12, listMoved_sigma) then
```
fingolfin, line 570: Hm. I guess there are more necessary conditions; e.g. I think also the union of the points moved by H2 and by sigma must contain the points moved by H1. And the same but with H1 and H2 swapped. Not sure whether any of these imply the other?

But then I think you can do better by analyzing orbits: If `x` is an element of the intersection, then for all positive integers `n`, we have `n^x` is an element of  `n^H1` but also of `(n^H2)^\sigma`. I think this condition strengthens all of the above. We would have:
```suggestion
    listMoved_all := Union(listMoved_H12, listMoved_sigma);
    if ForAny(listMoved_all, n -> IsEmpty(Intersection(Orbit(H1,n), OnSets(Orbit(H2,n),sigma)))) then
```
Of course that's a lazy implementation in that it wastes efforts by recomputing all orbits. But I'll leave optimizations to later.

---

For all this, I wonder: aren't we essentially replicating work that partition backtrack also does, just in a more adhoc fashion?

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
          <CanonicalRepresentativeOfExternalSet(b);
 end);
+
+
+
+InstallMethod(Intersection2, "perm groups", IsIdenticalObj,
+  [IsRightCoset and IsPermCollection,IsRightCoset and IsPermCollection],0,
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, listMoved_H1, listMoved_H2, listMoved_H12, listMoved_sigma, theInt, set2, set1, eRepr, set2_img, set1_img, H1_sigma, H2_sigma, test, H12, swap, eCos, rho, cosTest, diff12, diff21, fset1, fset2;
+    # We set int = cos1 cap cos2 = H1 x1 cap H2 x2
```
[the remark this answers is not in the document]
MathieuDutSik: Thank you for your comments.
I will try to do better and cover the issues you mentioned. As I said before, none of the RightCoset was tested and the PR is actually an improvement over this. So, I saw the PR as a step in the right direction though of course it is incomplete, heuristics are difficult to choose.

I also found some way to use ascending chains for making the iteration over the cosets more manageable. 

I also said that this is related to my work on permutalib where I want to achieve the best speed in permutation group algorithm. I want to contribute back to GAP as I gained a lot from it.

But yes, the goal is to get the best code, so I will work on implementing your correction that I find reasonable and discuss the one I think are not. Thank you for your review.

ChrisJefferson:
There is probably any interesting study to be done on the best way to intersect cosets in general.

A slight diversion: One major reason "vole" exists is after written "ferret" (my C++ implementation of partition backtracking), I realised I'd hard-wired various assumptions that made solving coset, and canonical image problems, very hard to implement. It's easy to do cosets if you think about it from the start, but given the lack of documentation of GAP's partition backtracker, I don't really want to dig through it and try to make it do cosets now, if it never did before.

I'm also not sure, in general, the best way of doing cosets. I think it might be an interesting research project -- coset intersection is in practice the problem vole finds hardest, and in some cases this PR is beating vole (in other cases vole is winning, so it's not a clear win one way or the other).

Based on past experience, I would encourage Mathieu, if you want to, to drop more comments in explaining how things work - in general most people never read library code, and it is useful for those who come along 10 years later to know what is going on :)

fingolfin (commented):
A few more comments.

I also want to explicitly say: @MathieuDutSik I can totally understand if you are fed up by constant "complaints" from me; I hope it is at least clear that they are well-intentioned, even if perhaps annoying...

lib/csetgrp.gi
```diff
@@ -789,6 +789,43 @@ local a,r;
 end);
 
 
+InstallMethod(Intersection2, "general groups", IsIdenticalObj,
```
ChrisJefferson, line 792: minor comment, change this to "general cosets"

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
   return CanonicalRepresentativeOfExternalSet(a)
          <CanonicalRepresentativeOfExternalSet(b);
 end);
+
+
+
+InstallMethod(Intersection2, "perm groups", IsIdenticalObj,
```
ChrisJefferson, line 554: similarly change this to "perm cosets"

ChrisJefferson:
While testing the code I wrote some more tests (I didn't find any bugs) -- in particular the matrix group tests provide complete coverage for the non-perm code. I paste them here, feel free to use them as is, or alter if you want.

```
# test trivial cases
gap> Intersection(RightCoset(Group([],()), ()), RightCoset(Group([],()), (1,2))) = [];
true
gap> Intersection(RightCoset(Group((1,2,3)), ()), RightCoset(Group((1,2,3)), (1,2))) = [];
true
gap> Intersection(RightCoset(AlternatingGroup(6), ()), RightCoset(AlternatingGroup(6),(1,2))) = [];
true
gap> Intersection(RightCoset(AlternatingGroup([1..5]), (1,2)), RightCoset(AlternatingGroup([6..10]), (1,2))) = RightCoset(Group(()), (1,2));
true

# test intersection non-permutation cosets
gap> matcyc := CyclicGroup(IsMatrixGroup, GF(3), 2);;
gap> [] = Intersection(RightCoset(matcyc, [[0*Z(3), Z(3)], [Z(3), Z(3)^0]]),
>                      RightCoset(matcyc, [[Z(3), Z(3)], [Z(3), 0*Z(3)]] ) );
true
gap> RightCoset(matcyc, [[0*Z(3), Z(3)], [Z(3), Z(3)^0]]) =
>      Intersection(RightCoset(matcyc, [[0*Z(3), Z(3)], [Z(3), Z(3)^0]]),
>                   RightCoset(matcyc, [[Z(3), Z(3)^0], [0*Z(3), Z(3)]] ) );
true
gap> rc1 := RightCoset(Group( [[0*Z(3), Z(3)], [Z(3), 0*Z(3)]], [[Z(3), Z(3)], [Z(3)^0, 0*Z(3)]],
>   [[Z(3), 0*Z(3)], [0*Z(3), Z(3)]] ), [[Z(3), Z(3)^0], [Z(3)^0, Z(3)^0]]);;
gap> rc2 := RightCoset(Group( [[Z(3), Z(3)], [0*Z(3), Z(3)^0]], [[0*Z(3), Z(3)], [Z(3)^0, 0*Z(3)]],
>   [[Z(3), Z(3)], [Z(3), Z(3)^0]], [[Z(3), 0*Z(3)], [0*Z(3), Z(3)]]), [[0*Z(3), Z(3)], [Z(3)^0, Z(3)]]);;
gap> RightCoset(Group( [[Z(3), Z(3)], [0*Z(3), Z(3)^0]], [[Z(3)^0, Z(3)^0], [0*Z(3), Z(3)]] ),[[0*Z(3), Z(3)^0], [Z(3), Z(3)^0]]) =
>      Intersection(rc1, rc2);
true
gap> [] = Intersection(RightCoset(matcyc, One(matcyc)), rc1);
true
gap> RightCoset(Group( [ [ Z(3)^0, 0*Z(3) ], [ 0*Z(3), Z(3)^0 ] ] ),
> [ [ 0*Z(3), Z(3)^0 ], [ Z(3)^0, 0*Z(3) ] ]) = Intersection(RightCoset(matcyc, One(matcyc)), rc2);
true
```

MathieuDutSik:
> I also want to explicitly say: @MathieuDutSik I can totally understand if you are fed up by constant "complaints" from me; I hope it is at least clear that they are well-intentioned, even if perhaps annoying...

My only issue is about the kind of double standard. A lot of code in GAP would require some upgrade and cleanup.

But I cannot really complain as I wrote something like 20-40 papers using GAP, so really no issue here.

fingolfin:
Well, I would not accept a lot of the code that is in there right now in its current form, and I wish more professors taught their students about the important of systematic tests, documentation algorithms and describing why they are correct, etc. People rely on systems like GAP to produce correct results and they'd be dismayed if they knew how often they produce garbage. See also https://github.com/gap-system/gap/labels/kind%3A%20bug%3A%20wrong%20result

lib/csetgrp.gi
MathieuDutSik: Thanks, done.

lib/csetperm.gi
MathieuDutSik: Thanks, done.

MathieuDutSik:
Thank you, I inserted those tests.

lib/csetperm.gi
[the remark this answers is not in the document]
MathieuDutSik: > There is no variable `int`, and I am not sure what this line is about. I'd just remove it

I choose to keep the line and rename it as **cosInt** which is clearer and is used later
in the comments.
I also renamed **theInt** as **grpInt**, again for clarity.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, listMoved_H1, listMoved_H2, listMoved_H12, listMoved_sigma, theInt, set2, set1, eRepr, set2_img, set1_img, H1_sigma, H2_sigma, test, H12, swap, eCos, rho, cosTest, diff12, diff21, fset1, fset2;
+    # We set int = cos1 cap cos2 = H1 x1 cap H2 x2
+    H1:=ActingDomain(cos1);
+    H2:=ActingDomain(cos2);
+    x1:=Representative(cos1);
+    x2:=Representative(cos2);
+    shift:=x1;
+    sigma:=x2 / x1;
+    # Initial very easy check
```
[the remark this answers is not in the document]
MathieuDutSik: You are correct, I added that check.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
   return CanonicalRepresentativeOfExternalSet(a)
          <CanonicalRepresentativeOfExternalSet(b);
 end);
+
+
+
+InstallMethod(Intersection2, "perm groups", IsIdenticalObj,
+  [IsRightCoset and IsPermCollection,IsRightCoset and IsPermCollection],0,
+function(cos1,cos2)
+    local H1, H2, x1, x2, shift, sigma, listMoved_H1, listMoved_H2, listMoved_H12, listMoved_sigma, theInt, set2, set1, eRepr, set2_img, set1_img, H1_sigma, H2_sigma, test, H12, swap, eCos, rho, cosTest, diff12, diff21, fset1, fset2;
```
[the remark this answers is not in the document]
MathieuDutSik: Ok, done.

lib/csetperm.gi
MathieuDutSik: It could be related to the Partition Backtrack, but if so I am unsure how this can be done.
It is certainly true that the intersection of subgroups is using partition backtrack. I do not
know how this could be implemented for the intersection of cosets.
More specific help would be appreciated.

lib/csetgrp.gi
```diff
@@ -789,6 +789,43 @@ local a,r;
...
+        cos2 := swap;
+    fi;
+    H1:=ActingDomain(cos1);
+    H2:=ActingDomain(cos2);
+    x1:=Representative(cos1);
+    x2:=Representative(cos2);
+    sigma := x1 / x2;
+    if H1 = H2 then
+        if sigma in H1 then
+            return cos1;
```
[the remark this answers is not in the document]
MathieuDutSik: This has been covered in a test contributed to by Christopher Jefferson.

lib/csetgrp.gi
```diff
@@ -789,6 +789,43 @@ local a,r;
...
+        else
+            return [];
```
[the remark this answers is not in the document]
MathieuDutSik: That has been added.

lib/csetgrp.gi
```diff
@@ -789,6 +789,43 @@ local a,r;
...
+    # To find U*rho, we iterate over all cosets of U in H1 and for each test
+    # if it is contained in H2*x2/x1, which is the case if and only if rho is
+    # in H2*x2/x1, if and only if rho/(x2/x1) = rho*x1/x2 is in H2
+    U:=Intersection(H1, H2);
+    for rho in RightTransversal(H1, U) do
+        if rho * sigma in H2 then
+            return RightCoset(U, rho * x1);
+        fi;
+    od;
+    return [];
```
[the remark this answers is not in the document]
MathieuDutSik: This has been added thanks to a contribution by Christopher Jefferson, so that is covered.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+    x2:=Representative(cos2);
+    shift:=x1;
+    sigma:=x2 / x1;
+    # Initial very easy check
+    listMoved_H1:=MovedPoints(H1);
+    listMoved_H2:=MovedPoints(H2);
+    listMoved_H12:=Union(listMoved_H1, listMoved_H2);
+    listMoved_sigma:=MovedPoints(sigma);
+    if not IsSubset(listMoved_H12, listMoved_sigma) then
+        return [];
```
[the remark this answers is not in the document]
MathieuDutSik: A test has been added.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+    theInt:=Intersection(H1, H2);
+    # Reducing as much as possible in advance
+    while true do
+        listMoved_H1:=MovedPoints(H1);
+        listMoved_H2:=MovedPoints(H2);
+        listMoved_H12:=Union(listMoved_H1, listMoved_H2);
+        listMoved_sigma:=MovedPoints(sigma);
+        # First exclusion case
+        if not IsSubset(listMoved_H12, listMoved_sigma) then
+            return [];
```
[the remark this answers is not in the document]
MathieuDutSik: A test has been added.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        fi;
+        # Easy reductions: points that are moved by sigma outside of one group allow us to reduce the problem
+        diff12:=Difference(listMoved_H1, listMoved_H2);
+        diff21:=Difference(listMoved_H2, listMoved_H1);
+        set2:=Intersection(diff21, listMoved_sigma);
+        if Length(set2) > 0 then
+            set2_img:=OnTuples(set2, Inverse(sigma));
+            eRepr:=RepresentativeAction(H2, set2, set2_img, OnTuples);
+            if eRepr=fail then
+                return [];
```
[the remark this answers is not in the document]
MathieuDutSik: Added.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+            continue;
+        fi;
+        # easy termination criterion
+        if sigma in H2 then
+            return RightCoset(theInt, shift);
+        fi;
+        if sigma in H1 then
+            # int = (H1 \cap H2 sigma) shift
+            #     = (H1 sigma^{-1} \cap H2) sigma shift
+            return RightCoset(theInt, sigma * shift);
```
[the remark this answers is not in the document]
MathieuDutSik: I added a test that should cover it.

lib/csetperm.gi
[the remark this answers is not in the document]
MathieuDutSik: Added a test. Hopefully covers this.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        fi;
+        # reduction on sets which is easy
+        fset1:=Difference(listMoved_H1, Union(listMoved_H2, listMoved_sigma));
+        if Length(fset1) > 0 then
+            H1:=Stabilizer(H1, fset1, OnTuples);
+            continue;
```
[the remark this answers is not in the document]
MathieuDutSik: I added a test for this.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+    # Initial very easy check
+    listMoved_H1:=MovedPoints(H1);
+    listMoved_H2:=MovedPoints(H2);
+    listMoved_H12:=Union(listMoved_H1, listMoved_H2);
+    listMoved_sigma:=MovedPoints(sigma);
+    if not IsSubset(listMoved_H12, listMoved_sigma) then
+        return [];
+    fi;
+    # We pass it, now getting into the hard computation
+    theInt:=Intersection(H1, H2);
```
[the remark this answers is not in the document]
MathieuDutSik: ok, done and corrected.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+                return [];
+            fi;
+            H1:=Stabilizer(H1, set1, OnTuples);
+            sigma:=sigma / eRepr;
+            shift:=eRepr * shift;
+            continue;
+        fi;
+        # easy termination criterion
+        if sigma in H2 then
+            return RightCoset(theInt, shift);
```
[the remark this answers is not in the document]
MathieuDutSik: Corrected.

lib/csetperm.gi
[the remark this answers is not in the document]
MathieuDutSik: Done and corrected.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        #     = (H1 sigma^{-1} \cap H2) sigma shift
+        swap:=H1;
+        H1:=H2;
+        H2:=swap;
+        shift:=sigma * shift;
+        sigma:=Inverse(sigma);
+    fi;
+    # So now Order(H1) <= Order(H2)
+    cosTest:=RightCoset(H2, sigma);
+    for rho in RightTransversal(H1, theInt) do
```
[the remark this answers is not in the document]
MathieuDutSik: Done and corrected.

lib/csetperm.gi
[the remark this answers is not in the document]
MathieuDutSik: Correct.

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,111 @@
...
+true
+gap> RightCoset(Group(()),(1,8,3,4,7,6,5,2)) =
+>    Intersection(RightCoset(Group([ (1,4)(2,5), (1,3,5)(2,4,6), (1,5)(2,4)(3,6) ]),(1,7,6,5)(3,4,8)),
+>                 RightCoset(Group([ (3,4), (5,6,7,8), (5,6) ]),(1,8,6,2)(3,7)));
+true
+gap> [] = Intersection(RightCoset(SymmetricGroup(4), ()), RightCoset(SymmetricGroup([3..6]), (4,7)));
+true
+gap> [] = Intersection(RightCoset(Group([(1,2,3,4,5)]),(4,5)), RightCoset(AlternatingGroup(4),()));
+true
+gap> RightCoset(SymmetricGroup(5), (7,9)) =
```
fingolfin, line 55:
```suggestion
gap> RightCoset(SymmetricGroup([3..5]), (7,9)) =
```

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,111 @@
...
+>    Intersection(RightCoset(SymmetricGroup(5), (1,2)(7,9)),
+>                 RightCoset(SymmetricGroup([3..7]), (7,9)));
+true
```
fingolfin, line 58: Tests fails as this actually returns `false`.

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,111 @@
...
+gap> [] = Intersection(RightCoset(Group([(1,2,3,4,5)]),(1,4)(3,5)), RightCoset(SymmetricGroup(3),()));
+true
+gap> RightCoset(Group([(5,6)]),(4,5)) =
+>    Intersection(RightCoset(SymmetricGroup(6), ()),
+>                 RightCoset(SymmetricGroup([5..8]), (4,5)));
+true
```
fingolfin, line 64: Also fails

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,111 @@
...
+
+# test intersection non-permutation cosets
+gap> RightCoset(Group([ [ [ -1, 0 ], [ 0, -1 ] ] ]),[[0,1],[1,0]]) =
+>    Intersection(RightCoset(Group([ [ [ -1, 0 ], [ 0, 1 ] ], [ [ 0, 1 ], [ 1, 0 ] ] ]), IdentityMat(2)),
+>                 RightCoset(Group([ [ [ -1, 0 ], [ 0, -1 ] ] ]),[[0,1],[1,0]]));
+true
+gap> RightCoset(Group([-IdentityMat(2)]),[[0,1],[1,0]]) =
+>    Intersection(RightCoset(Group([-IdentityMat(2)]),[[0,-1],[-1,0]]) =
+>                 RightCoset(Group([-IdentityMat(2)]),[[0,1],[1,0]]));
+true
```
fingolfin, line 85: This tests fails with a "method not found" error.

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,111 @@
...
+gap> [] = Intersection(RightCoset(Group([-IdentityMat(2)]),[[0,1],[1,0]]) =
+>                      RightCoset(Group([-IdentityMat(2)]),[[2,1],[1,2]]));
+true
```
fingolfin, line 88: Also `Error, no method found`

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        fset2:=Difference(listMoved_H2, Union(listMoved_H1, listMoved_sigma));
+        if Length(fset2) > 0 then
+            H2:=Stabilizer(H2, fset2, OnTuples);
+            continue;
+        fi;
+        # More general but more expensive than previous check
+        H1_sigma:=ClosureGroup(H1, sigma);
+        if not IsSubgroup(H1_sigma, H2) then
+            H2:=Intersection(H1_sigma, H2);
+            continue;
```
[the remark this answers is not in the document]
MathieuDutSik: I did a lot of random searches and could find cases covering it.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        fi;
+        H2_sigma:=ClosureGroup(H2, sigma);
+        if not IsSubgroup(H2_sigma, H1) then
+            H1:=Intersection(H2_sigma, H1);
+            continue;
```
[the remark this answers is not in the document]
MathieuDutSik: I did random searches and could find cases covering it.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+        sigma:=Inverse(sigma);
+    fi;
+    # So now Order(H1) <= Order(H2)
+    cosTest:=RightCoset(H2, sigma);
+    for rho in RightTransversal(H1, theInt) do
+        if rho in cosTest then
+            return RightCoset(theInt, rho * shift);
+        fi;
+    od;
+    return [];
```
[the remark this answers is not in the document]
MathieuDutSik: I did find test.

lib/csetperm.gi
MathieuDutSik: Ok, so I added your test. I think now that everything is covered.

lib/csetperm.gi
```diff
@@ -548,3 +548,148 @@ function(a,b)
...
+            H1:=Intersection(H2_sigma, H1);
+            continue;
+        fi;
+        # No more reduction tricks available
+        break;
+    od;
+    # A final termination criterion
+    H12:=ClosureGroup(H1, H2);
+    if not sigma in H12 then
+        return [];
```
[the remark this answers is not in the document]
MathieuDutSik: Test coverage has been added.

tst/testinstall/cset.tst
MathieuDutSik: I cleaned all that.

tst/testinstall/cset.tst
MathieuDutSik: I cleaned all that.

tst/testinstall/cset.tst
MathieuDutSik: I cleaned all that.

tst/testinstall/cset.tst
MathieuDutSik: I cleaned that.

tst/testinstall/cset.tst
MathieuDutSik: Also cleaned.

tst/testinstall/cset.tst
```diff
@@ -0,0 +1,140 @@
+#############################################################################
+##
+##  test of group intersection and RightCoset
+##
+gap> START_TEST("cset.tst");
+
```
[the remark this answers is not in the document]
fingolfin: I've added such a tester now

fingolfin:
I tweaked the tests a bit, and added another commit which adds some comments, and restores one quick criterion that was previously removed, and also improved "my" orbit criterion.

Alas, I also did a quick non-scientifically performance test, and compared how long it takes to process `tst/testinstall/cset.tst`.
- Before this patch: 59 msec
- After this patch: 117 msec

So these simple examples actually got slower by a factor 2! Of course that's a bit unfair because things might actually be a lot faster for bigger examples. And to a degree, it's unavoidable, as extra checks and conditions of course always add overhead which is mostly noticeable when the wins they bring are small.

I think the slow down is mostly due to the "generic" coset intersection. Indeed, computing `Intersection(H1, H2)` is expensive, as is working with the transversals. I added a heuristic in my final commit which falls back to the previous code if H1 or H2 contains less than 10 elements. That restores the timings for `cset.tst` to the old value. Alas, I think what this shows is that one should perhaps take some time to investigate which of the many heuristics added in the new methods are worth it, and perhaps have some beefier examples.

Of course it is trivially to produce some. E.g. `Intersection(RightCoset(SymmetricGroup(15), (1,2)(7,9)), RightCoset(SymmetricGroup([3..17]), (7,9)));` takes 1 millisecond with this PR to compute the intersection which has size 6227020800, while the previous code has no chance to work here. But I note that I get this performance already with the generic method (w/o any permutation group specific tricks).

@MathieuDutSik I wonder, do you have any examples where some of the additional permutation group heuristics really pay off?

MathieuDutSik:
No, I did not do such an analysis.
I took care that all cases are indeed occurring.
I don't think that a few Milliseconds really matters all that much.

On the other hand, I think reducing the number of cosets that need to be considered is really important as a reduction by 10 gets you a speed up by 10. So all those checks really make sense in my opinion.

To state my opinion most clearly, the problem is not with groups of size 10 but with the groups will size 10^10 or so. At least that is my opinion after 20 years of GAP.

Regarding the heuristic, yes some improvement are possible. But I think this is beyond the scope of the PR. At least I have other things to do right now.

ChrisJefferson:
I would expect this new code to be better on almost any large problem, as GAP currently has no code for coset intersection -- it just calls `Intersection(Set(coset1), Set(coset2))` (literally, it's coll.gi lines 2379, 2380 that do the work).

Similar to your results, you can beat GAP at group intersection with `Intersection(Set(group1), Set(group2));` for small groups as well.

fingolfin:
Yeah, I am confident that this PR is a great improvement over what we had before!

What is somewhat less clear to me is how much the new method for permutation group cosets is an improvement over the new one for generic groups. As usual, there is a balance between heuristics to reduce to "simpler" cases and actually doing something. 

It is of course true that the benefit in huge cases tends to infinity, while the slowdown for small cases is comparatively tiny. The reason why I don't think this should be completely ignored, though, is that code which for one reason or another deals with gazillions of small cases still is affected by such a speed regression. This is always a delicate matter.

Anyway, if that happens, I guess then at least there's a workaround (just call `Intersection(Set(coset1), Set(coset2))` directly), and we can add heuristics later on which improve things further. For example: add a new heuristic which says: don't bother with clever heuristics anymore as soon as the size of the smaller group is lower than some threshold.

MathieuDutSik:
It may be true that there is room to improve on the heuristics, when isn't there room for that?
But I consider this outside of the scope of this PR, at least from my viewpoint.

fingolfin:
Oh I fully agree.

fingolfin:
I'll make a few more tweaks (mostly adding some comments) and then will merge.

Thank you very much @MathieuDutSik for your great contribution and your patience -- and sorry for pestering you so much. But I really think the result is even better than what it started out with.

fingolfin (approved):
Finally found the time to go through this and add the explanatory comments. Also refactored the code a bit more.

I am very happy with this now. Once test pass, we can merge it.

Thanks again!

PR #4877, opened 2022-04-22
Add `MakeGroupyObj` helper

fingolfin:
Also set more filters automatically in `MakeGroupyType`. Only one caller to `MakeGroupyType` is left, namely in `MakeGroupyObj`.

### DIFF
--- a/grp/classic.gd
+++ b/grp/classic.gd
@@ -1413,8 +1413,8 @@ BindGlobal( "DECLARE_PROJECTIVE_SEMILINEAR_GROUPS_OPERATION",
       d:= Length( facts );
       p:= facts[1];
 
-      if d = 1 then
-        return lin;
+      if n = 1 then
+        return CyclicGroup( filt, d );
       fi;
 
       F:= GF( q );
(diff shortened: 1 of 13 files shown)


ChrisJefferson:
I wonder if when adding `IsTrivial` is a group, we should also add `Size = 1` (as that isn't currently auto-deduced, and various group algorithms check if the size is known).

fingolfin:
Good idea, will do

fingolfin:
Actually this revealed a genuine bug: `PGammaL(1,9)` is trivial (as in: `IsTrivial` is set and the single generator is the identity permutation), while its `Size` attribute is set to 2. Reading the documentation for `PGammaL`, I'd argue that `Size` is right and the returned group is wrong. I'll adjust this PR accordingly.

But I found a second issue: 
```
gap> G:=PGammaL(2,3); Size(G); Length(Elements(G));
GL(2,3)
48
48
```
But that doesn't conform to what the manual says; I would expect `PSL(2,3)` as result, which has only 24 elements.

CC @ThomasBreuer

fingolfin:
I've now fixed all the issues I mentioned in PGammaL and PSigmaL, and adjusted the tests accordingly. It would be good if @ThomasBreuer would double check this (the first commit in this PR)

fingolfin:
(Since PGammaL and PSigmaL are new in GAP 4.12, there is no need for release notes)

grp/classic.gd
```diff
@@ -1413,8 +1413,8 @@ BindGlobal( "DECLARE_PROJECTIVE_SEMILINEAR_GROUPS_OPERATION",
       d:= Length( facts );
       p:= facts[1];
 
-      if d = 1 then
-        return lin;
+      if n = 1 then
+        return CyclicGroup( filt, d );
```
ThomasBreuer, line 1417: Yes, the case n = 1 needs a special treatment because we need a faithful action of the field automorphisms.
I would still make a distinction for d = 1, in the sense that one does not need an additional generator in this case.

ThomasBreuer (approved):
Thanks for spotting and fixing these bugs.
(For d = 1, one did not even get a permutation group, so this was clearly wrong. The problem with n = 1 was more subtle.)

grp/classic.gd
fingolfin: done

PR #4882, opened 2022-04-27
kernel: tweak Julia integration

fingolfin:
Avoid all access to struct members defined by Julia, as these structs tend to
change across releases, requiring recompilation of `julia_gc.c`; this makes it
annoying to develop against ever-changing Julia nightly builds.

Also removed the last parts of the ability to load Julia from GAP; we've not
supported this code path for quite some time now.

### DIFF
--- a/src/julia_gc.c
+++ b/src/julia_gc.c
@@ -40,6 +40,7 @@
 #include <unistd.h>
 
 #include <julia.h>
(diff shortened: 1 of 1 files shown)


fingolfin:
Urgh... regarding "load Julia from GAP" of course I forgot that we *do* use this in one case, namely when testing. So I'll have to roll that back for now.

Issue #4881, opened 2022-04-26
Recent change (?) in coll.gi appears to break XMod test 


cdwensley:
The test in question is at xmod/tst/manual/icoclinic.tst, but the code below reproduced the problem. 
This test has recently started to fail in GAPdev, so I assume something has changed in coll.gi ? 
X8 is a crossed module and its size is a list of length 2: [ Size(Source(X8)), Size(Range(X8)) ]. 
The following code constructs crossed modules Z8, D8 and K8, but there is only a problem with the size [2,1] of K8. 
The line in question in coll.gi is:   Assert(0, not HasIsFinite(obj) or (IsFinite(obj) = (sz<infinity))); 
but I do not understand why this line should only be called with K8. 
### Observed behaviour
gap> X8;  Size(X8);
[Group( [ f1*f2*f3, f3, f4 ] )->Group( [ f2, f2 ] )]
[ 8, 2 ]
gap> Z8 := CentreXMod( X8 );;  Size(Z8); 
[ 2, 1 ]
gap> D8 := DerivedSubXMod(X8);;  Size( D8 );
[ 4, 1 ]
gap> K8 := IntersectionSubXMods( X8, Z8, D8 );;  Size(K8);
#I  X.action trivial => not checking a homomorphism!
Error, Assertion failure in
  IsFinite( obj ) = (sz < infinity) at /Applications/gap/gapdev/lib/coll.gi:3109 called from 
<function "SetSize method">( <arguments> )
 called from read-eval loop at *stdin*:23
you may 'return;'
brk> sz;
[ 2, 1 ]

### Copy and paste GAP banner (to tell us about your setup)
Christopher-Wensleys-iMac:xmod christopherwensley$ ../../dev.sh 
 ┌───────┐   GAP 4.12dev-1285-gdeede6c built on 2022-04-26 15:53:43+0100
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin17-default64-kv8
 Configuration:  gmp 6.2.1, GASMAN
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2020.08.11, 
             AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.24, CrystCat 1.1.9, 
             CTblLib 1.3.2, FactInt 1.6.3, FGA 1.4.0, GAPDoc 1.6.4, 
[...]

PR #4883, opened 2022-04-27
Make SetSize compatible again with non-standard Size

fingolfin:
The XMod package sets Size to values like `[ 8, 2 ]`. While this is
agains the specification for `Size` on collections, let's still try
not to break this needlessly. Luckily the fix is simple enough.


FriedrichRober (approved):

danielrademacher (approved):

cdwensley:
Yes, this does solve the problem reported in issue #4881.

PR #4892, opened 2022-05-05
extend the documentation of `AllBlocks`

ThomasBreuer:
State explicitly that smallest representatives are chosen.
(The current implementation has this property.)

If one wants to compare block systems of a group and a subgroup (acting transitively on the same set of points), it is useful to know that compatible representatives are chosen when this is possible --there is no need to compute the full block systems and then to compare them.


hulpke (approved):
Yes, this is true and makes sense to document.

PR #4893, opened 2022-05-19
Add `HexSHA256` function for computing SHA256 checksums

### LINKED ISSUE #4844: Add SHA256 function(s) to GAP itself?
For various purposes it would be useful if GAP had a built-in way to compute standard checksums. While the [crypting package](https://github.com/gap-packages/crypting) does provide this functionality, fact is that this package may not always be available, so it is problematic to rely on it.

So I propose that we add this to the GAP kernel -- e.g. by copying the `crypting` kernel sources. Then update crypting to use the built-in versions if available.

Thoughts? @ThomasBreuer? @ChrisJefferson?

fingolfin:
This uses C kernel code directly taken from the crypting packages, with the kernel functions renamed to avoid conflicting with crypting.

Resolves #4844.

Some concerns and ideas:
1. I picked `HexSHA256String` based on the name of the `SHA256String` in crypting, which does the same except the return value is not a hex string but rather a list of integers. So "String" seems to encode the argument type. But is this useful here? Can't we just call it `HexSHA256`? Well, I guess we might want a variant which takes a filename as argument, and then calling that e.g. `HexSHA256File` would be natural -- but this could be implemented as `HexSHA256String(StringFile(filename))`, so do we really need this? Hmm, thinking about it, perhaps yes: to deal with checksumming huge files, which we don't want to read into memory all at once?
2. It would probably be a useful to add a variant `HexSHA256Stream` which takes an input stream as argument.  And again: this could alternatively be also called `HexSHA256`, and then we can decide what to do based on the type of the argument. (If we handle streams, that also alleviates the issue of checksumming huge files without reading them into memory all at once).
3. How to ensure a smooth transition for crypting? Well, for now this PR tries hard to be compatible with `crypting`, and I just checked, it is. So we could leave things at that. But we could also tweak `crypting` to not load its kernel extension when it detects that the GAP kernel provides the required functionality. But there is no pressing need for that right now, I think? As far as I know, only `JupyterKernel` needs crypting, anyway.
4. Maybe we want other APIs beyond `HexSHA256String` (or whatever else to call it)? Suggestions, wishes?

### DIFF
--- a/src/sha256.c
+++ b/src/sha256.c
@@ -1,0 +1,406 @@
+/****************************************************************************
+**
+**  This file is part of GAP, a system for computational discrete algebra.
+**
+**  Copyright of GAP belongs to its developers, whose names are too numerous
+**  to list here. Please refer to the COPYRIGHT file for details.
+**
+**  SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#include "sha256.h"
+
+#include "bool.h"
+#include "error.h"
+#include "integer.h"
+#include "modules.h"
+#include "objects.h"
+#include "plist.h"
+#include "stringobj.h"
+
+#include "config.h"
+
+#include <string.h>
+
+static Obj GAP_SHA256_State_Type;
+
+// Implements the SHA256 hash function as per the description in
+// https://web.archive.org/web/20130526224224/http://csrc.nist.gov/groups/STM/cavp/documents/shs/sha256-384-512.pdf
+//
+
+
+// For the moment we assume the input is a string, we should probably have a
+// list of bytes, or words or something
+
+static inline UInt4 RotateRight(UInt4 x, const UInt4 n)
+{
+    return (x >> n) | (x << (32 - n));
+}
+
+static inline UInt4 Ch(UInt4 x, UInt4 y, UInt4 z)
+{
+    return (x & y) ^ (~x & z);
+}
+
+static inline UInt4 Maj(UInt4 x, UInt4 y, UInt4 z)
+{
+    return (x & y) ^ (x & z) ^ (y & z);
+}
+
+static inline UInt4 Sigma0(UInt4 x)
+{
+    return RotateRight(x, 2) ^ RotateRight(x, 13) ^ RotateRight(x, 22);
+}
+
+static inline UInt4 Sigma1(UInt4 x)
+{
+    return RotateRight(x, 6) ^ RotateRight(x, 11) ^ RotateRight(x, 25);
+}
+
+static inline UInt4 sigma0(UInt4 x)
+{
+    return RotateRight(x, 7) ^ RotateRight(x, 18) ^ (x >> 3);
+}
+
+static inline UInt4 sigma1(UInt4 x)
+{
+    return RotateRight(x, 17) ^ RotateRight(x, 19) ^ (x >> 10);
+}
+
+static const UInt4 k[] = {
+    0x428a2f98, 0x71374491, 0xb5c0fbcf, 0xe9b5dba5, 0x3956c25b, 0x59f111f1,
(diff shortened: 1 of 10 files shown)


src/sha256.c
```diff
@@ -0,0 +1,406 @@
...
+}
+
+/****************************************************************************
+**
+*F  InitSHA256() . . . . . . . . . . . . . . . . . .  table of init functions
+*/
+static StructInitInfo module = {
+    // init struct using C99 designated initializers; for a full list of
+    // fields, please refer to the definition of StructInitInfo
+    .type = MODULE_DYNAMIC,
```
ChrisJefferson, line 397: this needs updating now it's internal code

ThomasBreuer (commented):
Thanks for this addition.
Concerning the function name(s) and possible arguments, one function `HexSha256` with argument either a string or a stream would be sufficient for all use cases I have in mind.

src/sha256.c
fingolfin: indeed

tst/testinstall/sha256.tst
```diff
@@ -0,0 +1,28 @@
...
+gap> GAP_SHA256_HMAC(fail, fail);
+Error, GAP_SHA256_HMAC: <key> must be a string (not the value 'fail')
+gap> GAP_SHA256_HMAC("", fail);
+Error, GAP_SHA256_HMAC: <text> must be a string (not the value 'fail')
+
+#
+gap> HexSHA256String("abcd");
+"88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589"
+gap> HexSHA256String("abcd\n");
+"fc4b5fd6816f75a7c81fc8eaa9499d6a299bd803397166e8c4cf9280b801d62c"
```
ChrisJefferson, line 25: This is very low level, but maybe add:

`HexSHA256String(['a', 'b', 'c', 'd']);` (same as string).
`HexSHA256String("");` (should be 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 according to wikipedia?)
`HexSHA256String("abcd\r\n"); (just to check nothing stupid happens on windows, can't imagine how it would).

fingolfin: Good ideas. Also, the documentation needs to be hooked up so it actually appears in the manual ;-), and a few other things. I'll work on it later

fingolfin:
Hrrrmmmm... it turns out that our input streams don't offer a way to, say "read the next 4096 bytes or until EOF". You can either read one byte, one line or all data. And reading one byte at a time of course is prohibitively slow. Reading a line also is not fast. This seems like a major hole in the stream API. And I guess this might be part of why people consider GAP's built-in streams as too slow ?

Anyway, I pushed an update which supports streams via `ReadAll` -- this is not great for huge files, but least it works, and we can improve it later if someone needs it.

I also hooked up the documentation into the ref manual, and added more tests etc.

ThomasBreuer (approved):
The pull request provides SHA256 checksums in GAP, I like this, and the changes are fine.

Concerning the documentation, `HexSHA256` will appear in the "Strings" chapter (number 27) not in the "Streams" chapter (number 10), thus it might be difficult to find the function for streams.
If one searches for `??sha` then one gets a hit, perhaps it would be useful to add explicit index entries for "hash function" and "checksum" that point to `HexSHA256` (and also to `CrcFile`, `CrcString`?).

(I have added the label that causes the title to appear in the release notes, I hope this will be enough for the release notes.)

fingolfin:
Good ideas. We could also add "see also..." links from CrcString/CrcFile to the new function. (CrcString already points this way to CrcFile -- but not the other way around. So perhaps that, too, should be added)

ThomasBreuer:
@fingolfin Yes, links from `CrcFile`/`CrcString` to `HexSHA256` (and from `CrcFile` to `CrcString`) would be very useful.

Issue #4898, opened 2022-06-18
The incorrect displayed symbols in GAP online reference manual 39.7-6 CosetDecomposition.


hongyi-zhao:
I noticed the following incorrect displayed symbol [here](https://www.gap-system.org/Manuals/doc/ref/chap39.html#X82F6ABE378B928D1):

![image](https://user-images.githubusercontent.com/11155854/174422422-d5614308-4528-4850-a31c-9d8265b68bf1.png)

It's obvious that the correct form should be **$S \le G$**.

Regards,
HZ

fingolfin:
Go to the top of the page and click "Mathjax on", and you'll get to https://www.gap-system.org/Manuals/doc/ref/chap39_mj.html#X82F6ABE378B928D1 which should have all math formulas rendered nicely.

We probably should link to the mathjax version by default.

however, normally GAPDoc should (I think?!) convert such simple math formulas to Unicode even in the non-MathJax version, so perhaps more is going on? @frankluebeck will know.

PR #4900, opened 2022-06-18
Update teaching.g

frankluebeck:
Makes a formula in text and HTML (no MathJax) documentation look nicer. (Issue #4898)


hulpke (approved):
Completely uncontroversial. Thanks

frankluebeck:
I was not aware that `\le` is an existing synonym for `\leq` (only the latter is in the original LaTeX book).
But I have added `\le` and `\ge` to the `TEXTMTRANSLATIONS` a while ago. The main point of my proposed change was the spacing.

PR #4905, opened 2022-06-23
Speed up `IrrBaumClausen` (used for computing character tables of supersolvable groups, including $p$-groups)

frankluebeck:
For the non-linear characters fewer multiplications of monomial matrices
are needed.
Also improved further details: multiplication of monomial matrices,
avoiding arithmetic with cyclotomics.

Comments: 

This should never be slower than the previous version.
On my computer in the following example `IrrBaumClausen` (used to compute the irreducible characters of the group) becomes about 6 times faster:

```
gap> s:=SylowSubgroup(Sp(4,25),5);
ps<group of 4x4 matrices of size 390625 over GF(5^2)>
gap> ps:=Image(IsomorphismPcGroup(s));
Group([ f1, f2, f3, f4, f5, f6, f7, f8 ])
gap> ConjugacyClasses(ps);;time;
310
gap> BaumClausenInfo(ps);;time;
41
gap> ir:=IrrBaumClausen(ps);;time;
3608
```
## Text for release notes

Speedup of `IrrBaumClausen` which is used by GAP to compute the character table of supersolvable (in particular p-) groups.

### DIFF
--- a/lib/ctblsolv.gi
+++ b/lib/ctblsolv.gi
@@ -2038,30 +2038,33 @@ InstallMethod( IrrBaumClausen,
           info,           # result of `BaumClausenInfo'
           pcgs,           # value of `info.pcgs'
           lg,             # composition length
-          evl,            # list encoding exponents of class representatives
-          i, j, k,        # loop variables
-          exps,           # exponent vector of a group element
+          exps,           # exponents of class representatives
+          cr,             # exps sorted
+          dobylevel,      # generate matrices and traces of class reps
+          i, j, jj, k,    # loop variables
           t,              # intermediate representation value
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
This looks good.
- Avoiding arithmetic operations with cyclotomics is always a good idea; here this is done by creating the character values with `CycList`.
- Using `{}` instead of a `for` loop yields a speedup (in spite of the additional intermediate list).
- Memorizing partial products of the evaluation of a class representative and using the common initial part for the next representative is a good idea. (Since the lengths of the exponent vectors are in general much larger than the number of conjugacy classes, there are enough cases where common initial parts occur.)

PR #4906, opened 2022-06-25
Website: link to MathJax version of docs if available

fingolfin:
... at least when wrapping the next release. I see no downside to this?

### DIFF
--- a/dev/releases/HelpLinks-to-JSON.g
+++ b/dev/releases/HelpLinks-to-JSON.g
@@ -74,7 +74,6 @@ for x in NamesOfComponents(HELP_BOOKS_INFO) do
       path := "FAIL";
     else
       path := ReplacedString(path, fulldir, "");
-      path := ReplacedString(path, "_mj.html", ".html");
     fi;
 
     r.(bname).reference.(name) := path;


fingolfin:
@alex-konovalov this change is simply about not artificially removing the `_mj`. So, if a package has `HTMLStart := "doc/chap0.html"` we should keep using it. But if it has `HTMLStart := "doc/chap0_mj.html"` then we should not strip the `_mj` (which arguably may even be incorrect, as a package could also only ship the MathJax version in principle).

That said, I just checked and this patch does not have the desired effect anyway... So I'll have to take a closer look

Note that most packages using GAPDoc hardcode `chap0.html` right now (in the current distribution, 118 use `chap0.html`, just two use `chap0_mj.html`), but I've already committed patches to a couple dozen packages to change that, including the `example` package.

fingolfin:
All that said, I think @wilfwilson wrote the line removing the `_mj` and he may have had a reason for doing it that I am not are of!

fingolfin:
Actually, I think I did something wrong in my earlier testing... I've now redone it, and the patch *does* same to work: in fact "better" (?) than I expected, it gives links to the `_mj` files *whenever they are present*. So e.g. for `float`, which does not have `_mj` files, it still links to `chap0.html`.

So from that perspective: all is well :-)

alex-konovalov (approved):
Still a bit puzzled why it works and what it does, but happy with @fingolfin's testing.

alex-konovalov:
> Now the GAPDoc help handler (which is ultimately called here, of course) by default provides the nice MathJax enabled version, if available 

but this doesn't work in GAP 4.11.1 installation. If that's change later and I didn't know that, then this explains it.

fingolfin:
> but this doesn't work in GAP 4.11.1 installation. If that's change later and I didn't know that, then this explains it.

I don't follow. We are talking here about a script that is used while producing a *new* release. What does this have to do with GAP 4.11.1?

(That said, I don't see why re-running the modified script in GAP 4.11.1 shouldn't work as well?!?)

wilfwilson:
> All that said, I think @wilfwilson wrote the line removing the `_mj` and he may have had a reason for doing it that I am not are of!

No, I didn't write this line unfortunately, so I can't offer any insight.

You can see from this commit https://github.com/gap-system/gap/commit/e2d357a62c243e4ede96928db222537dc37c4e00 that the script [dev/releases/HelpLinks-to-JSON.g](https://github.com/gap-system/gap/pull/4906#diff-cac05791223f957cdfedbc4dd55b404fdbdfa2043df1e6d6532726dbf3b4e1e2) was adapted from the `LinksOfAllHelpSections.g` script of `gap-system/GapWWW`, see [this link](https://github.com/gap-system/GapWWW/blob/67e706efa3dde66fe67eb30fe95c7eac90f7bd48/etc/LinksOfAllHelpSections.g) for its final version. 

I don't remember what process I went through to create `HelpLinks-to-JSON.g`, But in any case, the predecessor script already contains the line `path := ReplacedString( path, "_mj.html", ".html" );`.

alex-konovalov:
@fingolfin what I mean by "his doesn't work in GAP 4.11.1 installation" referred to 

> Now the GAPDoc help handler (which is ultimately called here, of course) by default provides the nice MathJax enabled version, if available

In GAP 4.11.1 MathJax version is available but not opened by default.

PR #4918, opened 2022-06-30
kernel: make Bag pointer to an opaque struct

fingolfin:
This helps to further enforce that kernel code only accesses bags
through the APIs provided for that (PTR_BAG/ADDR_OBJ, BAG_HEADER etc.),
though of course, this being C, we cannot completely prevent abuse.

This is motivated by a desire to make further changes to GC internals (in another upcoming PR). The change was meant to expose potentially problematic code in the kernel (I found none!). There is a still chance that CI will reveal issues in a package kernel extension...

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -144,7 +144,7 @@ enum { FALSE = 0, TRUE = 1 };
 **
 **  (The documentation of 'Bag' is contained in 'gasman.h'.)
 */
-typedef UInt * *        Bag;
+typedef struct OpaqueBag * Bag;
 
 
 /****************************************************************************
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -133,6 +133,14 @@
 #include <sys/mman.h>
 #endif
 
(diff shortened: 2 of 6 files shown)


fingolfin:
Fails compiling ferret; fix submitted to @ChrisJefferson via https://github.com/gap-packages/ferret/pull/46

fingolfin:
Excellent, it seems ferret is the only package with such a problem.

ChrisJefferson:
I'm suprised how easy that was!

I plan on making a new ferret release shortly, so this PR shouldn't get held up too long.

ChrisJefferson:
I did some quick performance tests on mac and linux, and everything is the same.

At first it looks like this PR makes the executable slightly larger but that's just debugging information, after stripping they are the same size to the byte.

fingolfin:
@ChrisJefferson tests pass now thanks to the new ferret release

ChrisJefferson (approved):

PR #4920, opened 2022-06-30
kernel: make SET_PTR_BAG internal for GASMAN and Julia GC

fingolfin:
For Boehm / HPC-GAP, it is still needed in some places

### DIFF
--- a/src/weakptr.c
+++ b/src/weakptr.c
@@ -194,16 +194,14 @@ static inline void GROW_WPOBJ(Obj wp, UInt need)
     if (plen > INT_INTOBJ_MAX)
         plen = INT_INTOBJ_MAX;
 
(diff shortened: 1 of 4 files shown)


src/weakptr.c
```diff
@@ -194,16 +194,14 @@ static inline void GROW_WPOBJ(Obj wp, UInt need)
     if (plen > INT_INTOBJ_MAX)
         plen = INT_INTOBJ_MAX;
 
-#ifdef USE_BOEHM_GC
```
ChrisJefferson, line 197: Tthis bit of the PR seems unrelated, did it get merged in accidentally? I also wasn't sure if it was reasonable (I've not used Boehm in a long time, to know if this is correct).

ChrisJefferson (approved):

PR #4921, opened 2022-07-01
kernel: make three global variables static

fingolfin:

## Text for release notes

## Further details

### DIFF
--- a/src/saveload.c
+++ b/src/saveload.c
@@ -292,17 +292,14 @@ void SaveSubObj( Obj subobj )
     SaveUInt((UInt) subobj);
   else if (IS_FFE(subobj))
     SaveUInt((UInt) subobj);
-  else if ((((UInt)subobj & 3) != 0) || 
-           subobj < (Bag)MptrBags || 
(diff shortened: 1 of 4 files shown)


src/saveload.c
```diff
@@ -292,28 +292,25 @@ void SaveSubObj( Obj subobj )
     SaveUInt((UInt) subobj);
   else if (IS_FFE(subobj))
     SaveUInt((UInt) subobj);
-  else if ((((UInt)subobj & 3) != 0) || 
-           subobj < (Bag)MptrBags || 
-           subobj > (Bag)MptrEndBags ||
-           (Bag *)PTR_BAG(subobj) < MptrEndBags)
+  else if (IS_BAG_ID(subobj))
```
ChrisJefferson, line 295: Is this the same? It seems to have dropped the `PTR_BAG(subobj) < MptrEndBags` part? Which I think I've seen occasionally trigger when memory corruption bugs occur.

fingolfin: I've now adjusted the test to do this once again, too

ChrisJefferson (approved):

PR #4922, opened 2022-07-01
Added GPL 2 as second license

### LINKED ISSUE #4542: License issue: perfect groups library now partially under non-GPL license
PR #4530 just was merged; unfortunately I did not have the opportunity to review it before, otherwise I would have pointed it out there earlier: that PR added files which are under the artistic license 2.0. And there is nothing wrong with that at all (I was one of the people suggesting this as a suitable license for the various group libraries).

However, GAP right now claims that all its parts (other than packages) is under the GPL, which now is not accurate anymore. We'll have to adjust license statements in various places to address this.

This is a blocker for 4.12.

hulpke:
This resolves #4542


fingolfin (approved):
Thank you!

PR #4923, opened 2022-07-04
More GASMAN refactoring

fingolfin:
- exploit the new `struct OpaqueBag` by replacing `*bag` with `bag->body` in a bunch of places
- change `MptrBags` and `MptrEndBags` to be of type `Bag` (not `Bag *`), as they mark the beginning and end of the array of master pointers, so *really* a range of "bags". In contrast, `OldBags`, `YoungBags`, `AllocBags`, `EndBags` mark parts of the workspaces that are used to store the bodies of bags (their current type, `Bag *`, does not reflect this very well, but so be it...)
- fix `IS_VALID_BAG_ID` to compare the body pointer to `OldBags` instead of `MptrEndBags`
- change `SwapMasterPoint` to use the `SWAP` macro
- make some local variables in `CollectBags_Mark` even more local by moving there declarations directly to where they are used (this makes it a bit easier to see that the various parts of this function are not interacting directly)

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -214,8 +214,6 @@ GAP_STATIC_ASSERT(sizeof(void *) == sizeof(struct OpaqueBag),
 */
 
 enum {
-    SIZE_MPTR_BAGS = 1,
-
     T_DUMMY = NUM_TYPES - 1,
 };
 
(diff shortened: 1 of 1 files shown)


src/gasman.c
```diff
@@ -554,7 +552,7 @@ static inline BOOL IS_BAG_ID(void * ptr)
 
 BOOL IS_VALID_BAG_ID(Bag bag)
 {
-    return IS_BAG_ID(bag) && (PTR_BAG(bag) >= MptrEndBags);
+    return IS_BAG_ID(bag) && (PTR_BAG(bag) >= OldBags);
```
fingolfin, line 555: @ChrisJefferson FYI this should not cause problems in regular usage, but it should catch a few more potential memory corruptions (note in particular that for unused master pointers, `PTR_BAG` points to the next unused master pointer, in which case there is a chance the old `IS_VALID_BAG_ID` might have succeeded; while the modified version should catch it.

ChrisJefferson: I had a check and it didn't seem to find anything (currently), but also didn't break anything

ChrisJefferson (approved):

PR #4924, opened 2022-07-05
Revise SmallGeneratingSet

fingolfin:
A [recent change](https://github.com/gap-system/gap/commit/9d380c4091451fe8f6b8f5335c8bfc8b76a51556) made `SmallGeneratingSet` give up very quickly in certain cases. This broke the test suites of a bunch of packages (at least `groupoids` and `xmod`), and in general provided output which users may regard as suboptimal, e.g. including trivial permutations and duplicate generators.

This patch tries to go for a compromise: we still abort earlier than we used to, but not *quite* as early, and we try to at least remove "obvious" redundancies.

### DIFF
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1983,46 +1983,41 @@ end);
 InstallMethod(SmallGeneratingSet,"random and generators subset, randsims",true,
   [IsPermGroup],0,
 function (G)
-local  i, j, U, gens,o,v,a,sel,mintry,orb,orp,isok;
+local  i, j, U, gens,a,mintry,orb,orp,isok;
 
   gens := ShallowCopy(Set(GeneratorsOfGroup(G)));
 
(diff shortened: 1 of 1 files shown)


hulpke (approved):
Thank you -- this indeed was a a use case I did not think of, but makes complete sense.

cdwensley:
I've come to this rather late.  Spent some time this morning fixing tests in Groupoids which were failing under dev.  This now appears to have been unnecessary, but I will have a good look tomorrow afternoon.

fingolfin:
@cdwensley sorry for wasting your time :-/. I should have notified you about this patch.

hulpke:
These messages (which were prompted by users misunderstanding how the functions work) are issued by `InfoPerformance`. Wouldn't it make sense to set the InfoLevel of this to zero during testing?

PR #4925, opened 2022-07-06
Document `IsSquareInt` (it has been around for a long time)

fingolfin:
This is a useful helper and not going anywhere

### DIFF
--- a/doc/ref/integers.xml
+++ b/doc/ref/integers.xml
@@ -71,6 +71,7 @@ in chapter&nbsp;<Ref Chap="Combinatorics"/>.
 <#Include Label="LogInt">
 <#Include Label="RootInt">
 <#Include Label="SmallestRootInt">
+<#Include Label="IsSquareInt">
 <#Include Label="ListOfDigits">
 
 <ManSection>
--- a/lib/primality.gd
+++ b/lib/primality.gd
@@ -76,16 +76,19 @@
 ##
 #F  IsSquareInt(<n>)
 ##
+##  <#GAPDoc Label="IsSquareInt">
 ##  <ManSection>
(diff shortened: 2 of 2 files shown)


ChrisJefferson:
Do we care about the following? (I can't find any input where `IsSquareInt` returns `true` for a wrong input, just sometimes false and sometimes an error). Comments / tests / fixes welcome :)

```
gap> IsSquareInt(2/3);
false
gap> IsSquareInt(3/2);
Error, ModRat: for <r>/<s> mod <n>, <s>/gcd(<r>,<s>) and <n> must be coprime in
  n mod 64 at /home/caj/files/reps/gap/gap/lib/primality.gi:147 called from
<function "IsSquareInt">( <arguments> )
 called from read-eval loop at *stdin*:20
brk>
```

fingolfin:
The function performs some initial tests which reject things that are not squares, and those tests can indeed work for inputs that are not integers, as long as `mod n` does something meaningful for them. However, it only ever returns `true` after computing `RootInt`, so this function will never return `true` for anything that is not truly a square integer.

We could of course add a check at the start which produces an error message for any non-integers inputs. But let me point out that tons of functions in the GAP library has similar issues.  E.g. spending 5 minutes going through `lib/numtheory.gi`:

```
gap> OrderMod(0, -1/2);  # note the error which does not mention that it expects an integers
Error, <m> must be positive

gap> OrderMod(0, 1/2);  # still only gives an indirect error
Error, GCD_INT: <a> must be an integer (not a rational)

gap> IsPrimitiveRootMod(-1/2, 5);
true

gap> Legendre(-1/2, 5);
-1

gap> RootMod(-1/2, 3, 5);
3
```

wucas (approved):

PR #4926, opened 2022-07-06
Remove last regular uses of InstallValue

fingolfin:
... from the GAP library -- package still use it. Also, it is still used in
InstallFlushableValue and InstallFlushableValueFromFunction

@ThomasBreuer I am curious about the use of `InstallFlushableValue ` for `LARGEST_IDENTIFIER_NUMBER` in regular GAP -- the documentation for `Identifier` to me strongly suggests that the values should be unique "in the current GAP session", but they are not if `FlushCaches()` is ever called. Is that really desirable?

### DIFF
--- a/lib/ctbl.gd
+++ b/lib/ctbl.gd
@@ -1647,6 +1647,8 @@ DeclareAttributeSuppCT( "Identifier", IsNearlyCharacterTable, [] );
 ##  <Var Name="LARGEST_IDENTIFIER_NUMBER"/>
 ##
 ##  <Description>
+##  list containing the largest identifier of an ordinary character table in
+##  the current session.
 ##  <!--  We have to use a list in order to admit
 ##    <C>DeclareGlobalVariable</C> and
 ##    <C>InstallFlushableValue</C>.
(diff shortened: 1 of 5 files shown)


ThomasBreuer:
@fingolfin The question is what "cache" means. If a "cache" is just something that helps to improve performance then flushing the cache does not change the behaviour of objects in the current GAP session, thus flushing the cache does not cause problems.

Concerning the caches that are controlled by `InstallFlushableValue` and `FlushCaches`, this is in general not the case. `LARGEST_IDENTIFIER_NUMBER` is one example, and also `Z_MOD_NZ` can be regarded as problematic in this sense:
```
gap> R:= Integers mod 6;
(Integers mod 6)
gap> ZmodnZObj( 1, 6 ) in R;
true
gap> FlushCaches();
gap> ZmodnZObj( 1, 6 ) in R;
false
```
(The point is that `Z_MOD_NZ` stores the families of the `Integers mod <n>` objects in the current session.)

fingolfin:
@ThomasBreuer I think we've long established that there are many different kinds of "caches", and one has to qualify what one means. So I really was only concerned about `LARGEST_IDENTIFIER_NUMBER` specifically, as I know nothing about that. Yes, one can also question whether `Z_MOD_NZ` does "the right thing" (I tend to agree with you there), but that's separate, and there I at least have a fairly good idea what is is good for -- but at the same time I have no idea what the point of `LARGEST_IDENTIFIER_NUMBER` is... And that's fine, I don't need to know. It's just that I am wondering why it claims to provide an id which is unique across sessions but then it can be reset after all. But I don't know what that is used for, and what ramifications the "flushing" have ... Since I am touching this now, I could easily change it to be not flushed. Or leave it.

wucas (approved):
Lets merge this so #4929 can be rebased.

PR #4928, opened 2022-07-06
Fix off-by-one error in CopyListEntries

### LINKED ISSUE #4927: CopyListEntries can create lists when the last element is unbound
In GAP, lists should always have a bound last element, but `CopyListEntries` can be used to create a completely unbound list, for example:

```
gap> l := [1,2,3,4,5,6];
[ 1, 2, 3, 4, 5, 6 ]
gap> CopyListEntries([], 1, 1, l, 1, 1, 6);
gap> l;
[ ,,,,, ]
```

This list is very "fragile", for example:

```
gap> IsBound(l[6]);
false
gap> l;
[ ,,,,, ]
gap> Length(l);
6
gap> Unbind(l[6]);
gap> l;
[  ]
```

ChrisJefferson:
The length of the list may be changed if we add new members to the
list to the end of list, or unbind the final element.

Fixes #4927

### DIFF
--- a/src/listfunc.c
+++ b/src/listfunc.c
@@ -1388,7 +1388,7 @@ static Obj FuncCOPY_LIST_ENTRIES(Obj self, Obj args)
         }
     }
 
-  if (dstmax > LEN_PLIST(dstlist))
+  if (dstmax >= LEN_PLIST(dstlist))
     {
       sptr = CONST_ADDR_OBJ(dstlist)+dstmax;
       ct = dstmax;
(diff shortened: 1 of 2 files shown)


fingolfin:
In GAP 4.4, when this wasn't a kernel function yet, the problematic code triggered an error. I wonder: do we really want to support copying "unbound entries" via `CopyListEntries`?

ChrisJefferson:
Banning copying unbound members would mean we couldn't use SyMemmove for the most efficient case (although this function seems to do lots of things, I wonder how often it's even used with a step value other than 1?)

PR #4929, opened 2022-07-06
Don't use InstallFlushableValue, and more "caching changes"

fingolfin:
This builds upon PR #4926 which perhaps should be merged first, then I can rebase this PR and reviewing gets easier...

- add a helper `NEW_SORTED_CACHE` for use in tandem with the existing `GET_FROM_SORTED_CACHE`
- change a bunch of code to use that new helper
- add a comment explaining how to use `GET_FROM_SORTED_CACHE`
- merge the HPC-GAP and plain GAP implementations of `GET_FROM_SORTED_CACHE` (they now were identical other than for some `atomic` guards anyway)
- changed `ZEV_DATA` to use nested "SORTED_CACHE" instances as well

I agree with @ThomasBreuer that in principle, `Z_MOD_NZ` should be changed to not flush anymore. But then it essentially becomes a leak, because the user can fill it up but not clear anything from it (at least officially)... The proper fix would be to use a weak pointer object for the families here, and then letting `FlushCaches` just compact the thing to clean out unused entries...

### DIFF
--- a/lib/cache.gd
+++ b/lib/cache.gd
@@ -113,11 +113,52 @@ InstallMethod( FlushCaches, "return method", [], function() end );
 ##  </ManSection>
 ##  <#/GAPDoc>
 ##
-
 DeclareGlobalFunction("MemoizePosIntFunction");
+
+
+#############################################################################
+##
+#F  NEW_SORTED_CACHE(<flushable>)
+##
+##  Set up a cache object suitable for use with `GET_FROM_SORTED_CACHE`.
+##  If `flushable` is `true` (which is also the default), then the cache will
+##  be flushed whenever `FlushCaches` is called.
+##
+BIND_GLOBAL("NEW_SORTED_CACHE",
+function(flushable)
+    local cache;
+    cache := [ [], [] ];
+    if IsHPCGAP then
+        ShareSpecialObj(cache);
+    fi;
+    if flushable then
+        InstallMethod( FlushCaches, [],
+          function()
+              atomic readwrite cache do
+                cache[1] := MigrateObj([], cache);
+                cache[2] := MigrateObj([], cache);
+              od;
+              TryNextMethod();
+          end );
+    fi;
+    return cache;
+end);
+
 
 #############################################################################
 ##
 #F  GET_FROM_SORTED_CACHE(<cache>, <key>, <maker>)
+##
+##  Lookup the the given `key` inside `cache`, and return it. If the key is
+##  not yet in the cache, call the 0-argument function `maker`, store its
+##  return value under `key`, and return that value.
+##
+##  Internally, the cache is represented by a list of two lists. The first
+##  of the two lists contains the sorted keys; the second list contains the
+##  corresponding values. So if `cache[1][i]` equals `key`, then the
+##  associated value is stored in `cache[2][i]`.
+##
+##  The advantage of using this helper function is reduced code duplication,
+##  and thread safety when using HPC-GAP.
 ##
 DeclareGlobalFunction("GET_FROM_SORTED_CACHE");
--- a/lib/cache.gd
+++ b/lib/cache.gd
@@ -13,6 +13,37 @@
 
 #############################################################################
 ##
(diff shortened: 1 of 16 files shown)


lib/cache.gd
```diff
@@ -82,11 +113,52 @@
...
+function(flushable)
+    local cache;
+    cache := [ [], [] ];
+    if IsHPCGAP then
+        ShareSpecialObj(cache);
+    fi;
+    if flushable then
+        InstallMethod( FlushCaches, [],
+          function()
+              atomic readwrite cache do
```
wucas, line 137: What does `atomic readwrite` do?

lib/ctblfuns.gi
```diff
@@ -4749,50 +4742,21 @@ InstallGlobalFunction( ZevDataValue, function( q, n )
...
-    pos2:= Position( ZEV_DATA[2][ pos ][1], n );
-    if pos2 = fail then
-      Add( ZEV_DATA[2][ pos ][1], n );
-      pos2:= Length( ZEV_DATA[2][ pos ][1] );
-    fi;
+InstallGlobalFunction( ZevData, function( q, n, listofpairs... )
+    local aux, result;
 
-    if Length( arg ) = 3 then
+    aux := GET_FROM_SORTED_CACHE(ZEV_DATA, q, {} -> NEW_SORTED_CACHE(false));
```
wucas, line 4748: Am I understanding this right, that `ZEV_DATA` itself caches other cashes? And if so, why is `ZEV_DATA` flushable but the caches contained in it are not?

wucas (commented):
I don't see anything wrong with this with the caveat that I don't quite understand what `ZEV_DATA` is for. 
I also never came across `atomic readwrite` and according to the [manual](https://www.gap-system.org/Manuals/doc/ref/chap4.html#X87506BDC7D5F789E) these keywords are unused. Can someone explain what they do?

lib/cache.gd
fingolfin: This is from HPC-GAP, it does nothing in plain GAP, but it takes an exclusive lock on the object `cache` in HPC-GAP

lib/ctblfuns.gi
fingolfin: I've never noticed `ZEV_DATA` before I made this PR either, I only found it because I looked for places using `InstallFlushableValue`.

It is really a cache whose keys are pairs `(q,n)`, but it is implemented in this nested fashion, which I mapped to a "cache of caches". Depending on actual usage, it might be better to use a single cache with the pairs `[q,n]` as key, but I did not want to stray too far from what the code does right now...

The contained caches are not flushable because they are automatically flushed (i.e., discarded): when `ZEV_DATA` is flushed, its content is discarded, i.e. the caches it contained are removed; flushing them separately would just incur extra work with no benefit

lib/cache.gd
wucas: Ok thanks for the explanation. So in normal GAP the line `atomic readwrite cache do` and the corresponding `od;` are simply ignored. There is no hidden loop or anything in there?

lib/ctblfuns.gi
wucas: Thanks! That sounds sensible, I wanted to be sure that no other code assumes that the contents of `ZEV_DATA` is still there after flushing.

wucas (approved):

lib/ffe.gd
```diff
@@ -322,10 +322,8 @@ DeclareGlobalFunction( "FFEFamily" );
 ##  </Description>
 ##  </ManSection>
 ##
-BIND_GLOBAL( "FAMS_FFE_LARGE", [ [], [] ] );
-if IsHPCGAP then
-    ShareSpecialObj( FAMS_FFE_LARGE );
-fi;
+BIND_GLOBAL( "FAMS_FFE_LARGE", NEW_SORTED_CACHE(false) );
```
fingolfin, line 325: I've noticed and added a few more places which could use `NEW_SORTED_CACHE`, like this one.

PR #4930, opened 2022-07-06
Prepare etc/Makefile.gappkg for future gac changes

fingolfin:
In the future I hope to replace all use of GNU libtool in gac,
to make life easier for downstream packagers (e.g. Debian, Fedora, ...)
One of the changes will be that object files created by gac will go
back to use .o as extension instead of .lo. This patch prepares
`Makefile.gappkg` for this.

The idea here is that by including this update now, we can start
changing packages to use it *now*, rather than when the other
gac changes are ready.

(BTW, this is *not* strictly necessary, as I am trying hard in
the changed gac to still support the old `Makefile.gappkg`. That
said, any change we can merge now makes the gac PR smaller and
easier to review)

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -850,7 +850,10 @@ GAP="$(SYSINFO_GAP)"
 # GAC: path to the `gac` executable (may be a shell script)
 GAC="$(SYSINFO_GAC)"
 
-# Build flags for use by `gac` resp. build systems: these correspond to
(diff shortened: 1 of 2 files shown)


wucas (approved):
I am not really experienced with Makefiles but the changes look sensible to me and everything compiles fine with the changes on my machine.

PR #4934, opened 2022-07-08
kernel: more GC related refactoring

fingolfin:
- make SET_PTR_BAG implementation private for Boehm GC, too
- use `struct OpaqueBag` for Bohem GC and Julia GC, too
- use `SWAP` in all `SwapMasterPoint` implementations

### DIFF
--- a/src/boehm_gc.c
+++ b/src/boehm_gc.c
@@ -16,6 +16,7 @@
 #include "boehm_gc.h"
 
 #include "gapstate.h"
+#include "gaputils.h"
 #include "gasman.h"
 #include "objects.h"
 
@@ -39,6 +40,13 @@
 
 Int SyStorKill;
 
+// fill in the data "behind" bags
+struct OpaqueBag {
+    void * body;
+};
+GAP_STATIC_ASSERT(sizeof(void *) == sizeof(struct OpaqueBag),
+                  "sizeof(OpaqueBag) is wrong");
+
 #ifndef WARD_ENABLED
 
 static inline Bag * DATA(BagHeader * bag)
@@ -46,6 +54,11 @@ static inline Bag * DATA(BagHeader * bag)
     return (Bag *)(((char *)bag) + sizeof(BagHeader));
 }
 
+void SET_PTR_BAG(Bag bag, Bag *val)
+{
+    GAP_ASSERT(bag != 0);
+    bag->body = val;
+}
 
 /****************************************************************************
 **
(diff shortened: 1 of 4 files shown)


src/boehm_gc.c
```diff
@@ -39,13 +40,25 @@
 
 Int SyStorKill;
 
+// fill in the data "behind" bags
+struct OpaqueBag {
+    void * body;
+};
+GAP_STATIC_ASSERT(sizeof(void *) == sizeof(struct OpaqueBag),
```
wucas, line 47: Just to understand, this Assert checks if the Byte size of a pointer and the byte size of the struct is the same. How can it not be?

src/boehm_gc.c
```diff
@@ -356,7 +369,7 @@ void RetypeBagIntern(Bag bag, UInt new_type)
             old_mem = PTR_BAG(bag);
             old_mem = ((char *)old_mem) - sizeof(BagHeader);
             memcpy(new_mem, old_mem, size);
-            SET_PTR_BAG(bag, (void *)(((char *)new_mem) + sizeof(BagHeader)));
+            SET_PTR_BAG(bag, DATA(new_mem));
```
wucas, line 372: I don't quite understand why `DATA(new_mem)` and `(void *)(((char *)new_mem) + sizeof(BagHeader))` to the same thing.

wucas (approved):

src/boehm_gc.c
ChrisJefferson: Technically, structs in C can be bigger than just their members.

This is more common with multiple members, for example {int x; char c;} is size 8 on my computer, not 4+1=5 as you may expect.

I would be very suprised is there was a computer where this assert was false, but the C standard technically allows it.

src/boehm_gc.c
fingolfin: The definition of `DATA` is
```C
static inline Bag * DATA(BagHeader * bag)
{
    return (Bag *)(((char *)bag) + sizeof(BagHeader));
}
```
hence `DATA(new_mem)` evaluates to almost the exact same expression, the only difference is that the final result is case to `Bag *` instead of `void *`, but since `SET_PTR_BAG` expects a `Bag *` as its second argument, the `void *` was implicitly converted to a `Bag *` anyway.

Does that clear it up?

wucas: Yes thanks! I was erroneously looking at 
```C
static inline Bag *DATA(BagHeader *bag)
{
    return (Bag *)(bag + 1);
}
```
in `gasman.c`.

fingolfin: That's actually equivalent, too ;-). In C, if you have a pointer at a type (such as `BagHeader`), then adding 1 to that pointer increases the pointer (viewed as a "byte address") by `sizeof(ThatType)`

Issue #4901, opened 2022-06-19
Algebraic extensions in large characteristic broken


frankluebeck:
### Observed behaviour
After updating the master branch (to commit cd0a8a743) some of my currently developed code was broken. This could be traced back to:
```
gap> F := GF(2305843009213693967);
GF(2305843009213693967)
gap> xx := Indeterminate(F, "xx");; pol := xx^2 + One(F);;
gap> F2 := AlgebraicExtension(F, pol);
<field of size 5316911983139663560790518517532197089>
gap> a := Random(F2);
ZmodpZObj(38949595212977430,2305843009213693967)*a+ZmodpZObj(20384525485352021\
80,2305843009213693967)
gap> a![1];
[ ZmodpZObj( 2038452548535202180, 2305843009213693967 ), 
  ZmodpZObj( 38949595212977430, 2305843009213693967 ) ]
gap> b := a*a;
ZmodpZObj(1878660435820231628,2305843009213693967)*a+ZmodpZObj(718878163927484\
220,2305843009213693967)
gap> b![1];
<vector mod 2305843009213693967: [ 718878163927484220, 1878660435820231628 ]>
gap> c := b*b;
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `ProductCoeffs' on 2 arguments at /export/home/luebeck/gap4/lib/methsel2.g:249 called from
ProductCoeffs( x![1], y![1] ) at /export/home/luebeck/gap4/lib/algfld.gi:490 called from
<function "* AlgElm*AlgElm">( <arguments> )
 called from read-eval loop at *stdin*:8
type 'quit;' to quit to outer loop
brk> 
```

### Expected behaviour
Obvious...

As a temporary work around I'm now using:

```
gap> F := GF(2305843009213693967);
GF(2305843009213693967)
gap> xx := Indeterminate(F, "xx");; pol := xx^2 + One(F);;
gap> F2 := AlgebraicExtension(F, pol);
<field of size 5316911983139663560790518517532197089>
gap> fam := ElementsFamily(FamilyObj(F2));;
gap> fam!.reductionMat := List(fam!.reductionMat,List);              # Change back to list of lists!
[ [ ZmodpZObj( 2305843009213693966, 2305843009213693967 ), 
      ZmodpZObj( 0, 2305843009213693967 ) ] ]
gap> a := Random(F2);
[...]

PR #4935, opened 2022-07-09
Some fixes concerning matrices and polynomials over large prime fields (related to issue #4901)

frankluebeck:
I stumbled over more problems with matrices and polynomials over Integers mod p for large p. Some problems were already reported in "Algebraic extensions in large characteristic broken #4901".

I could now trace back some of the problems to calls of `ImmutableMatrix` in the library, which internally calls `Matrix` and `DefaultMatrixRepForBaseDomain`. This lead to some places where objects (algebraic extensions, groups, ...) suddenly have stored attributes in `IsMatrixObj` which do not interact properly (e.g., with "classical" vectors and matrices, that is
lists and lists of lists). 

This pull request disables for now the use of `IsZmodnZMatrixObj` as a default representation (see issue #4901 for some discussion).
After that the default representation became `IsPlistMatrixRep` and running the tests in `tst/testinstall.g` also showed various problems with this representation. We add a few missing methods and some other changes which fix the problems shown in the current test suite. 

It is very likely that there are more bugs and missing methods in this area. It is not the aim of this pull request to resolve these problems more thoroughly. But at least is resolves the concrete bugs which I found while working on the `StandardFF` package.

The commit messages contain further details.

 ## Text for release notes
none
(Probably not a topic for CHANGES because this does not concern code that was in 4.11.)

\

### DIFF
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -174,6 +174,8 @@ InstallMethod( Display, "for a plist vector", [ IsPlistVectorRep ],
     Print(v![ELSPOS],"\n>\n");
   end );
 
+InstallMethod( CompatibleVectorFilter, ["IsPlistMatrixRep"],
+  M -> IsPlistVectorRep );
 
 ############################################################################
 ############################################################################
@@ -1291,4 +1293,10 @@ InstallMethod( NewCompanionMatrix,
     Add(ll,l);
     return ll;
   end );
+
+# tell method selection that objects know their BaseDomain
+# (this should be in the first section above, but that causes
+# various "method matches more than one declaration" messages)
+InstallTrueMethod(HasBaseDomain, IsPlistVectorRep);
+InstallTrueMethod(HasBaseDomain, IsPlistMatrixRep);
 
--- a/grp/glzmodmz.gi
+++ b/grp/glzmodmz.gi
@@ -171,7 +171,7 @@ local geni,m,slmats,gens,f,rels,i,j,k,l,mat,mat1,mats,id,nh,g;
 
   g := Group(slmats);
   mat := Concatenation(id{[nh+1..n]},-id{[1..nh]});
-  SetInvariantBilinearForm(g,rec(matrix:=mat));
+  SetInvariantBilinearForm(g,rec(matrix:=Matrix(ring,mat)));
   return g;
 end);
 
@@ -436,7 +436,7 @@ local oper,n,R,o,nrit,
     g:=Group(List(GeneratorsOfGroup(g),TransposedMat));
     SetSize(g,e);
   fi;
-  SetInvariantBilinearForm(g,rec(matrix:=f*oner));
+  SetInvariantBilinearForm(g,rec(matrix:=Matrix(R, f*oner)));
   
   return g;
 end);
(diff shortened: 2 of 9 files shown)


frankluebeck:
Ok, there seems to be another test file  `tst/teststandard/opers/Matobjnz.tst` with errors.
I have just traced one of these errors and it leads to the problem I have mentioned (and not solved) in my commit messages:
It is due to a call of `ImmutableVector` inside the MeatAxe code which yields a vector in `IsPlistVectorRep` in an environment where all other vectors and matrices are not in `IsVectorObj/IsMatrixObj` .

```
gap> p := NextPrimeInt(10^10);;
gap> F := GF(p);;
gap> mat := One(F)*[[1,2],[3,4]];
[ [ ZmodpZObj( 1, 10000000019 ), ZmodpZObj( 2, 10000000019 ) ], 
  [ ZmodpZObj( 3, 10000000019 ), ZmodpZObj( 4, 10000000019 ) ] ]
gap> v := mat[1];
[ ZmodpZObj( 1, 10000000019 ), ZmodpZObj( 2, 10000000019 ) ]
gap> vp := ImmutableVector(F, v);
<immutable plist vector over GF(10000000019) of length 2>
gap> v * mat;
[ ZmodpZObj( 7, 10000000019 ), ZmodpZObj( 10, 10000000019 ) ]
gap> w := vp * mat;
<immutable plist vector over GF(10000000019) of length 2>
gap> w[1];
[ [ ZmodpZObj( 1, 10000000019 ), ZmodpZObj( 2, 10000000019 ) ], 
  [ ZmodpZObj( 3, 10000000019 ), ZmodpZObj( 4, 10000000019 ) ] ]
gap> w[2];
[ [ ZmodpZObj( 2, 10000000019 ), ZmodpZObj( 4, 10000000019 ) ], 
  [ ZmodpZObj( 6, 10000000019 ), ZmodpZObj( 8, 10000000019 ) ] ]
```
I'm not sure about a good solution. 
(A) Installing more methods for mixed arithmetic??? That probably leads to more new problems than it solves. For example, it is not clear what the type of the result should be. (This is the approach that is partially taken in `lib/matobjnz.gi` and which lead to problems I try to address with this pull request.)
(B) Maybe `ImmutableMatrix` and `ImmutableVector` should always return a matrix or vector in the same representation as the input? In particular when called on list of lists matrices it should not return an object in `IsMatrixObj`. Such objects are never useful when created within the existing code for groups, MeatAxe and so on. (Of course, these latter codes should work nicely when their input consists of object in `IsVectorObj/IsMatrixObj`.)

hulpke:
The main reason for overloading `ImmutavleMatrix` has been so the compact matrices get used for matrix groups over residue class rings -- this is a significant speed issue.

lib/matobjplist.gi
```diff
@@ -1292,3 +1294,9 @@ InstallMethod( NewCompanionMatrix,
     return ll;
   end );
 
+# tell method selection that objects know their BaseDomain
+# (this should be in the first section above, but that causes
+# various "method matches more than one declaration" messages)
+InstallTrueMethod(HasBaseDomain, IsPlistVectorRep);
+InstallTrueMethod(HasBaseDomain, IsPlistMatrixRep);
```
fingolfin, line 1301: The "method matches more than one declaration" messages are caused by "hidden implications", i.e. the same as issue https://github.com/gap-system/gap/issues/1649. :-(

That's quite annoying...

frankluebeck:
Yes.

PR #4937, opened 2022-07-12
kernel: change how ours GCs represent master pointers

### LINKED ISSUE #817: Investigate replacing the foreign type `GAP_jll.GapObj` by `struct GapObj  x::Any end`
This may be able to avoid https://github.com/oscar-system/Oscar.jl/issues/1400 and https://github.com/JuliaLang/PackageCompiler.jl/issues/700

But I am not sure whether it (a) is possible at all, and (b) will resolve the issue. Of course if it was possible, that'd be nice in any case....

To test this theory, I need to
1. create a patch for the GAP kernel that allows to pass in the GapObj type (bonus points if it fallbacks to work as it does right now).
   - idea for this: check if the "parent module" we pass to `GAP_InitJuliaMemoryInterface` already contains a global `GapObj` which is a struct type, and if so, use that
2. adapt GAP.jl and GAP_jll to support that; this may even involve moving the early GAP initialization back from GAP_jll into GAP.jl (no need to do it in the JLL anymore, as the one reason for doing that was to have the type GapObj available when precompiling GAP.jl)
3. ...

fingolfin:
So far, our master pointers are pointers to a pointer to the data portion of a bag. This patch changes them to instead point at the start of the bag.

Pointing inside the bag, as was done before, was an optimization for the common case where one wants to access the content of a bag. However, at least on x86 and ARM architectures, there is no difference in practice, because adding an offset of 16 (GASMAN) resp. 8 (Boehm, Julia) is essentially free for most machine code opcodes used to access the data.

So this should not introduce a drawback, but at the same time there is no real advantage to doing this.

BUT of course there is still a reason why I want to do this: I need this for experimental work on the Julia GC integration (in an attempt to resolve https://github.com/oscar-system/GAP.jl/issues/817). But right now the result is crashing, in the Julia GC -- hence I adapted the other GCs to also use this mode, so I can test that everything is fine in GASMAN with this change. It is locally for me, but having it here as a PR gives a chance for more extensive CI tests to check it; also, the change is ABI breaking, so testing it on a "clean slate" is better than on my system.

tl;dr: this PR is not for merging but to allow me to experiment

### DIFF
--- a/src/boehm_gc.c
+++ b/src/boehm_gc.c
@@ -55,15 +55,10 @@ struct OpaqueBag {
 
 #ifndef WARD_ENABLED
 
-static inline Bag * DATA(BagHeader * bag)
-{
-    return (Bag *)(((char *)bag) + sizeof(BagHeader));
-}
-
 void SET_PTR_BAG(Bag bag, Bag *val)
 {
     GAP_ASSERT(bag != 0);
-    bag->body = val;
+    bag->body = (Bag *)((BagHeader *)val - 1);
 }
 
 /****************************************************************************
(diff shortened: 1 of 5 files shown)


ChrisJefferson:
Just to say, I'm generally in support of this -- I don't think it will make things any worse for GAP, and I can understand how it will improve integration with other systems. Good luck!

ChrisJefferson (approved):

Issue #2864, opened 2018-09-25
document functions in libgap-api.h, and related functions


dimpase:
as this is meant for the outside world, it need more info. E.g. the signature of 
`Obj GAP_EvalString(const char * cmd)` does not say what `Obj` will be. Here are few examples I tried (calling `ViewObj()` on each result:
```
GAP_EvalString("a := SymmetricGroup(4);") # ->
[ [ true, Sym( [ 1 .. 4 ] ), false,, "Sym( [ 1 .. 4 ] )" ] ]

GAP_EvalString("[1,2,3,4];9*[[42,7]];")          # ->
[ [ true, [ 1, 2, 3, 4 ], false,, "[ 1, 2, 3, 4 ]" ], 
  [ true, [ [ 378, 63 ] ], false,, "[ [ 378, 63 ] ]" ] ]

GAP_EvalString("[1,2,3,4];;")                          # ->
[ [ true, [ 1, 2, 3, 4 ], true,, "" ]]
```
One thing seems to be clear: it outputs the list of certain evaluations of the input,
one item per statement. 
I know (again, not from docs, but from experiments) that `ViewObj()` outputs `Obj` more or less as they are, without adding extra info, and thus each item is what `GAP_EvalString()` produces.
Each item is a list; its members are

1) a boolean, meaning unknown
2) something that looks like the string rep. of the result, regardless of the presense of `;;`
3) boolean, seems to indicate whether `;;` was present or not
4) ???
5) screen representation of the object (`;;` suppresses it)

Can this be clarified? I'd make a pull request with this in comments to the code...

markuspf:
It executes `READ_ALL_COMMANDS` as "documented" here: https://github.com/gap-system/gap/blob/d7c32f314cd512fdd879dba22007626ea14ed8d8/src/streams.c#L85

note that `GAP_EvalString` is also just an example of what one *could* do; I am flexible as to what it should do/return/etc.

dimpase:
Would it be correct to say that `GAP_EvalString` calls `READ_ALL_COMMANDS` on the input stream "provided" (how exactly?) by its argument?

dimpase:
[...]

PR #4949, opened 2022-07-15
kernel: add docs for GAP_EvalString

fingolfin:
I wrote this a long time ago as part of a different branch that I never finished, but I figure this one is useful on its own. It addresses issue #2864 at least partially. Perhaps @dimpase is interested in reviewing it?

### DIFF
--- a/src/libgap-api.h
+++ b/src/libgap-api.h
@@ -189,11 +189,36 @@ void GAP_CollectBags(BOOL full);
 //// program evaluation and execution
 ////
 
-// Evaluate a string of GAP commands.
+// GAP_EvalString attempts to execute all statements read from the string
+// <cmd> as if they occurred in a GAP source file that is read by the GAP
+// command `Read`. However, no break loops are triggered. If evaluation a
+// statement leads to an error, the next statement is executed, and so on.
+//
+// It returns a list of lists, each entry of which reflects the result of the
+// execution of one statement. Specifically, each entry is another list, each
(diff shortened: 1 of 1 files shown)


dimpase (changes requested):
It's not clear from the description whether side effects of the statements are ignored, or the statements are executed sequentially.

fingolfin:
What you mean with "ignoring side effects of statements "? The whole purpose of statements are their side effects. The only way to suppress them is to not execute them

dimpase (commented):
Say, if statement number N wasn't successfully executed, does this mean that statements number N+1, N+2, etc will fail (or even left unevaluated) ?

dimpase:
This differs from the usual interactive use, where a failure to execute a statement brings one into a `brk` loop, right?

fingolfin:
@dimpase edited the comment a bit.

dimpase (approved):
Thanks. One may wonder why bother to proceed executing commands after an error (but that's for another PR/issue).

fingolfin:
I agree the interface is a bit strange and we could do better these days, but in its defense, back when Markus wrote it, things in the GAP kernel were a lot more tangled up so it wasn't clear how. Also, this interface was originally created for use in Jupyter notebooks, and apparently (??) it was deemed "sensible" to do it that in that situation... though I don't know how much was intention and how much was "we need to wrangle this into something sensible and this it the best we can achieve for now"...

Anyway, for now, I only want to document the status quo :-).

PR #4951, opened 2022-07-15
Make `SwapMatrixColumns` and `SwapMatrixRows` faster for classic matrices (lists-of-lists)

fingolfin:
Yet another commit from an unfinished branch/PR that might be considered useful on its own: faster "early methods" for SwapRows/SwapCols using kernel methods. It is not tested thoroughly and I am not even sure we want to go in this direction, but we could e.g. use this as basis for discussions during the GAP Days

---

This speeds up swap operations for plain list matrices a lot, and
even for "real" matrix objs, as the overhead of a GAP early method
is much higher than the overhead of one written in the kernel.

Now whether this gain is worth the extra complexity is another question;
though of course when doing heavy linear algebra, these row and column
can add up.

Before:

    gap> mat:=IdentityMat(10,Integers);;
    gap> for i in [1..1000000] do SwapMatrixColumns(mat,1,2); od; time;
    326
    gap> for i in [1..1000000] do SwapMatrixRows(mat,1,2); od; time;
    90
    gap> for i in [1..1000000] do tmp:=mat[1];mat[1]:=mat[2];mat[2]:=tmp; od; time;
    37

    gap> mat:=IdentityMat(10,GF(2));; ConvertToMatrixRep(mat);; mat;
    <a 10x10 matrix over GF2>
    gap> for i in [1..1000000] do SwapMatrixRows(mat,1,2); od; time;
    146
    gap> for i in [1..1000000] do tmp:=mat[1];mat[1]:=mat[2];mat[2]:=tmp; od; time;
    84

After:

    gap> mat:=IdentityMat(10,Integers);;
    gap> for i in [1..1000000] do SwapMatrixColumns(mat,1,2); od; time;
    153
    gap> for i in [1..1000000] do SwapMatrixRows(mat,1,2); od; time;
    19
    gap> for i in [1..1000000] do tmp:=mat[1];mat[1]:=mat[2];mat[2]:=tmp; od; time;
    45

    gap> mat:=IdentityMat(10,GF(2));; ConvertToMatrixRep(mat);; mat;
    <a 10x10 matrix over GF2>
    gap> for i in [1..1000000] do SwapMatrixRows(mat,1,2); od; time;
    116
    gap> for i in [1..1000000] do tmp:=mat[1];mat[1]:=mat[2];mat[2]:=tmp; od; time;
    83

### DIFF
--- a/src/lists.c
+++ b/src/lists.c
@@ -1799,6 +1799,61 @@ void AsssListLevelCheck (
 
 /****************************************************************************
 **
+*F  FuncSWAP_MAT_ROWS( <self>, <mat>, <row1>, <row2> )
+*/
+static Obj SwapMatRows;
+
+static Obj FuncSWAP_MAT_ROWS(Obj self, Obj mat, Obj row1, Obj row2)
+{
+    if (IS_POS_INTOBJ(row1) && IS_POS_INTOBJ(row2) && IS_PLIST(mat)) {
+        Int r1 = INT_INTOBJ(row1);
+        Int r2 = INT_INTOBJ(row2);
+        Obj row1list = r1 <= LEN_PLIST(mat) ? ELM_PLIST(mat, r1) : 0;
+        Obj row2list = r1 <= LEN_PLIST(mat) ? ELM_PLIST(mat, r2) : 0;
+        if (!row1list)
+            ErrorMayQuit("Matrix Element: <mat>[%d] must have an assigned value",
+                         (Int)r1, 0);
+        if (!row2list)
+            ErrorMayQuit("Matrix Element: <mat>[%d] must have an assigned value",
+                         (Int)r2, 0);
+
+        SET_ELM_PLIST(mat, r1, row2list);
+        SET_ELM_PLIST(mat, r2, row1list);
+        return 0;
+    }
+
+    return DoOperation3Args(SwapMatRows, mat, row1, row2);
+}
+
+
+/****************************************************************************
+**
+*F  FuncSWAP_MAT_COLS( <self>, <mat>, <col1>, <col2> )
+*/
+static Obj SwapMatCols;
+
+static Obj FuncSWAP_MAT_COLS(Obj self, Obj mat, Obj col1, Obj col2)
+{
+    if (IS_POS_INTOBJ(col1) && IS_POS_INTOBJ(col2) && IS_PLIST(mat)) {
+        Int c1 = INT_INTOBJ(col1);
+        Int c2 = INT_INTOBJ(col2);
+        Int nrows = LEN_PLIST(mat);
+        for (int r = 1; r <= nrows; r++) {
+            Obj row = ELM_LIST(mat, r);
+            Obj elm1 = ELM_LIST(row, c1);
+            Obj elm2 = ELM_LIST(row, c2);
+            ASS_LIST(row, c1, elm2);
+            ASS_LIST(row, c2, elm1);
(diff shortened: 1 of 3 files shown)


fingolfin:
The work is done and I see no harm. The idea of course is that ideally code switching to `SwapMatrixRows` should be penalized as little as possible for it, or even better, should benefit from it. This is clearly the case for the plist matrix. For the GF2 rep it still is a little bit slower but that could be mitigated further in the future if desired.

So I see no harm in merging this. Agreed @ThomasBreuer ?

ThomasBreuer (approved):
No problem with the idea and with the use of the kernel operation for matrices without holes.
I do not understand the explicit error cases in `FuncSWAP_MAT_ROWS`. Apparently the code deals also with holes inside the lists:
```
gap> l:= [ 1, 2,, 4 ];; SWAP_MAT_ROWS( l, 2, 3 ); l;
[ 1,, 2, 4 ]
```
(If the hole is beyond the length of the list, the result is different.)

src/lists.c
```diff
@@ -1797,6 +1797,63 @@ void AsssListLevelCheck (
...
+*F  FuncSWAP_MAT_ROWS( <self>, <mat>, <row1>, <row2> )
+*/
+static Obj SwapMatRows;
+
+static Obj FuncSWAP_MAT_ROWS(Obj self, Obj mat, Obj row1, Obj row2)
+{
+    if (IS_POS_INTOBJ(row1) && IS_POS_INTOBJ(row2) && IS_PLIST(mat)) {
+        Int r1 = INT_INTOBJ(row1);
+        Int r2 = INT_INTOBJ(row2);
+        if (r1 <= LEN_PLIST(mat) && r2 <= LEN_PLIST(mat)) {
```
fingolfin, line 1811: This is wrong: if one is in range and the other is not, we now do nothing. But either we should swap them (possibly requiring some resizing), or report an error.

src/lists.c
```diff
@@ -1797,6 +1797,63 @@ void AsssListLevelCheck (
...
+static Obj SwapMatCols;
+
+static Obj FuncSWAP_MAT_COLS(Obj self, Obj mat, Obj col1, Obj col2)
+{
+    if (IS_POS_INTOBJ(col1) && IS_POS_INTOBJ(col2) && IS_PLIST(mat)) {
+        Int c1 = INT_INTOBJ(col1);
+        Int c2 = INT_INTOBJ(col2);
+        Int nrows = LEN_PLIST(mat);
+        for (int r = 1; r <= nrows; r++) {
+            Obj row = ELM_LIST(mat, r);
```
fingolfin, line 1844: We could skip unbound rows...
```suggestion
            Obj row = ELM0_LIST(mat, r);
            if (!row) continue;
```

src/lists.c
```diff
@@ -1797,6 +1797,63 @@ void AsssListLevelCheck (
...
+            Obj elm1 = ELM_LIST(row, c1);
```
fingolfin, line 1845: We could also allow handling unbound entries by using `ELM0_LIST` here. But then we can't just call `ASS_LIST` as that requires the element to be non-NULL.

fingolfin:
Revised. Now we get
```
gap> l:= [ 1, 2,, 4 ];; SWAP_MAT_ROWS( l, 2, 3 ); l;
Error, Matrix Element: <mat>[3] must have an assigned value
brk>
```
and the error handling should be consistent overall.

PR #4957, opened 2022-07-18
Temporary workaround for Cygwin CI failure

### LINKED ISSUE #4955: CI: timeout in `testmockpkg testinstall - - windows-2019`
Maybe @ChrisJefferson or @wilfwilson have an idea about it: since recently, GAP CI job `testmockpkg testinstall - - windows-2019` is failing due to a timeout. Looking at <https://github.com/gap-system/gap/actions/workflows/CI.yml?query=branch%3Amaster> this started two days ago after PR #4946 was merged -- however, I don't think anything in there could explain this timeout.

So I am wondering if instead something changed in the Cygwin setup GH action?

fingolfin:
Based on discussion with @wilfwilson and @ChrisJefferson on Slack, this might resolves #4955.

If it does, then we can either revert https://github.com/gap-actions/setup-cygwin/commit/7f953df8576fbc43264718b21a602e72605f0859 there, or else debug further why it hangs now

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -164,6 +164,12 @@ jobs:
 
       - uses: gap-actions/setup-cygwin@v1
         if: ${{ runner.os == 'Windows' }}
+        with:
+          # HACK/FIXME/TODO: workaround for https://github.com/gap-system/gap/issues/4955
+          # which essentially removes xdg-utils ; its presence leads GAP to think the
+          # Cygwin environment is Unix, not Windows, which enables some tests which then
+          # hang.
+          PKGS_TO_INSTALL: 'wget,git,gcc-g++,gcc-core,m4,libgmp-devel,make,automake,libtool,autoconf,autoconf2.5,zlib-devel,libreadline-devel,libmpc-devel,libmpfr-devel'
 
       - name: "Set up compiler and linker flags"
         run: |


fingolfin:
This indeed seems to help.

So here is what I think is going on: the hang is in `tst/testinstall/testunix/streams.tst` which only really does something when `ARCH_IS_UNIX()` returns `true`. Which is defined as:
```gap
BIND_GLOBAL("ARCH_IS_UNIX",function()
  return not ARCH_IS_WINDOWS();
end);
```
and `ARCH_IS_WINDOWS` in turn is defined as:
```gap
BIND_GLOBAL("ARCH_IS_WINDOWS",function()
  # Exit early is we are not in Cygwin
  if POSITION_SUBSTRING (GAPInfo.Architecture, "cygwin", 0) = fail then
    return false;
  fi;
  # Check if programs we need exist in their standard Cygwin locations
  return not (IsExistingFile("/usr/bin/sh") or IsExistingFile("/bin/sh")) or
         not (IsExistingFile("/usr/bin/xdg-open") or IsExistingFile("/bin/xdg-open"));
end);
```

Thanks to the CI changes, `xdg-open` is now available -- leading GAP to decide "this is not Windows", i.e. "this is unix", and hence `tst/testinstall/testunix/streams.tst ` actually is executed, and... hangs. Dunno why it hangs, but now it is clear how that `setup-cygwin` change triggers it.

But to me the real culprit here is this crazy way to define `ARCH_IS_WINDOWS` and `ARCH_IS_UNIX`. Ugh.

wilfwilson:
Since `gap-actins/setup-cygwin` seems to be doing 'the right thing', and we have a possible workaround here, I suppose reverting that PR over there would not be the right course of action.

I wouldn't object to merging this PR temporarily (possibly amended to not re-add `gcc`, and possibly just removing `xdg-utils` rather than replacing it with `xdg-open`) so that the CI isn't affected in the meantime. But it seems like a fairly important thing to resolve (before 4.12, surely?) and doing so might mean we lose track of it.

PR #4962, opened 2022-07-27
Initializing `SyNumProcessors` in runtime with the number of available CPUs for Linux, Windows and macOS

### LINKED ISSUE #1219: Compute number of CPUs (and set `SyNumProcessors` accordingly) at runtime
Right now, HPC-GAP initializes `SyNumProcessors` to a value given at compile time (defaulting to 4). But this number really should be computed at runtime. Of course this requires non-portable code, but that's no excuse to do the wrong thing: After all, the machine we compile on may not be the machine we run on; for distros like Debian, that's indeed virtually always the case.

So, we need to add a new function, say `SyCountProcessors()` and provide implementations for at least Windows, Linux, OS X; and a generic fallback function which takes e.g. 4 as default).

nskeip:
`SyCountProcessors()` retrieves the number of active logical processors. In case of Linux, calls `get_nprocs()`. In case of Windows 7 and higher, calls `GetActiveProcessorCount(ALL_PROCESSOR_GROUPS)`. In case of older versions of Windows calls `GetSystemInfo` with proper arguments. In case of macOS `sysctl` is called with `HW_AVAILCPU` flag. In case of other operating systems the value of `4` is returned.

Resolves #1219.

## Text for release notes

see title

### DIFF
--- a/src/hpc/cpu.c
+++ b/src/hpc/cpu.c
@@ -15,8 +15,6 @@
 #include <sys/unistd.h>
 #elif _WIN32_WINNT >= _WIN32_WINNT_WIN7
 #include <windows.h>
-#elif _WIN32
-#include <sysinfoapi.h>
 #endif
 #endif
 
@@ -51,13 +49,6 @@ UInt SyCountProcessors(void)
     return result;
 #elif _WIN32_WINNT >= _WIN32_WINNT_WIN7
     return GetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
-#elif _WIN32
-    SYSTEM_INFO info;
-    GetSystemInfo(&info);
-    if (info.dwNumberOfProcessors < 1) {
-        return fallback_cpus_number;
-    }
-    return info.dwNumberOfProcessors;
 #else
     return fallback_cpus_number;
 #endif
--- a/src/hpc/cpu.c
+++ b/src/hpc/cpu.c
@@ -26,14 +26,6 @@ UInt SyNumProcessors = 4;
 *F  SyCountProcessors() . . . . . . . . . . . . compute the number of CPUs.
 **
 **  SyCountProcessors() retrieves the number of active logical processors.
-**  In case of Linux, calls get_nprocs().
-**  In case of Windows 7 and higher, calls
-**  GetActiveProcessorCount(ALL_PROCESSOR_GROUPS).
-**  In case of older versions of Windows, tries to load the number of
-**  processors in current processor group via GetSystemInfo with proper
-**  arguments and falls back to the value of 4 in case of an error.
-**  In case of macOS, sysctl is called with HW_AVAILCPU flag.
-**  In case of other operating systems the value of 4 is returned.
 */
 
 UInt SyCountProcessors(void) {
--- a/src/hpc/cpu.c
+++ b/src/hpc/cpu.c
@@ -10,12 +10,10 @@
 
 #include "hpc/cpu.h"
 
-#ifndef NUM_CPUS
 #ifdef _POSIX_C_SOURCE
 #include <sys/unistd.h>
 #elif _WIN32_WINNT >= _WIN32_WINNT_WIN7
 #include <windows.h>
-#endif
 #endif
 
 /****************************************************************************
@@ -23,11 +21,7 @@
 *V  SyNumProcessors  . . . . . . . . . . . . . . . . . number of logical CPUs
 **
 */
-#ifdef NUM_CPUS
-UInt SyNumProcessors = NUM_CPUS;
-#else
 UInt SyNumProcessors = 4;
-#endif
 
 /****************************************************************************
 **
@@ -37,12 +31,9 @@ UInt SyNumProcessors = 4;
 */
 UInt SyCountProcessors(void)
 {
-#ifdef NUM_CPUS
-    return NUM_CPUS;
-#else
     const UInt fallback_cpus_number = 4;
 #if _POSIX_C_SOURCE
-    const int  result = sysconf(_SC_NPROCESSORS_ONLN);
+    const int result = sysconf(_SC_NPROCESSORS_ONLN);
     if (result < 1) {
         return fallback_cpus_number;
     }
@@ -51,6 +42,5 @@ UInt SyCountProcessors(void)
     return GetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
 #else
     return fallback_cpus_number;
-#endif
 #endif
 }
--- a/src/hpc/cpu.c
+++ b/src/hpc/cpu.c
@@ -16,7 +16,7 @@
 #include <windows.h>
 #endif
 
-static const UInt FALLBACK_CPUS_NUMBER = 4;
+#define FALLBACK_CPUS_NUMBER 4
 
 /****************************************************************************
 **
--- a/src/hpc/cpu.c
+++ b/src/hpc/cpu.c
@@ -10,6 +10,20 @@
 
 #include "hpc/cpu.h"
 
+#ifndef NUM_CPUS
+#ifdef __linux__
+#include <sys/sysinfo.h>
+#elif _WIN32
+#if _WIN32_WINNT >= _WIN32_WINNT_WIN7
+#include <windows.h>
+#else
+#include <sysinfoapi.h>
+#endif
+#elif __APPLE__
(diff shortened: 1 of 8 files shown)


src/hpc/cpu.c
```diff
@@ -0,0 +1,74 @@
...
+**  In case of macOS, sysctl is called with HW_AVAILCPU flag.
+**  In case of other operating systems the value of 4 is returned.
+*/
+
+UInt SyCountProcessors(void) {
+#ifdef NUM_CPUS
+    return NUM_CPUS;
+#else
+    #if __linux__
+        return get_nprocs();
```
fingolfin, line 44: Actually, this file is not generally available on Linux; rather, it is a GNU extension and as such provided by glibc, but not necessarily all libc implementations available on Linux. I just checked, luckily Musl libc *does* provide it (as a legacy wrapper).

To fix this, I see the following alternatives:
1. use a macro check for GNU libc and use that as a guard (least work needed)
2. actually check for the availability of `get_nprocs` (and also `sys/sysinfo.h` !) in `configure.ac` (better, but more work)
3. avoid `get_nprocs` and instead call `sched_getaffinity(getpid(), ...)` and then count the CPUs in the returned CPU set (this is how musl libc implements `get_nprocs`, and for GNU libc it is the final fallback -- for some reason the first thing it tries is to parse `/sys/devices/system/cpu/online`, if that fails it falls back to parsing `/proc/stat`; if that fails it calls `sched_getaffinity`).

My personal preference would be number 2. I do not mean to burden you with this as a requirement for merging your check; so, if you feel inclined to work on this: great! If not, no worries, then I can take care of it later.

src/hpc/cpu.c
```diff
@@ -0,0 +1,74 @@
...
+        #if _WIN32
+            #if _WIN32_WINNT >= _WIN32_WINNT_WIN7
+                return GetActiveProcessorCount(ALL_PROCESSOR_GROUPS);
+            #else
+                SYSTEM_INFO info;
+                GetSystemInfo(&info);
+                if(info.dwNumberOfProcessors < 1) {
+                    return fallback_cpus_number;
+                }
+                return info.dwNumberOfProcessors;
```
fingolfin, line 56: So this is for Windows versions before Windows 7? @ChrisJefferson do we even support that anymore?

Actually, going one step further: does this even matter since we use Cygwin on Windows, which has `get_nprocs()` as far as I could tell? If so, it might be sufficient if the `get_nprocs()` code is modified to use an `autoconf` check

src/hpc/cpu.c
fingolfin: Actually, to be precise, on musl libc `get_nprocs` is implemented by a simple call to `sysconf(_SC_NPROCESSORS_ONLN)`, and *that* is handled by checking `sched_getaffinity`. 

According to https://stackoverflow.com/a/150971/928031 it works on Linux, Solaris, AIX and Mac OS X >=10.4 ; and I just checked, current FreeBSD and OpenBSD have it. So perhaps that's the easiest way to go about all this?

src/hpc/cpu.c
```diff
@@ -0,0 +1,74 @@
...
+            int nm[2] = {CTL_HW, HW_AVAILCPU};
+            size_t len = 4;
+            uint32_t count;
+
+            sysctl(nm, 2, &count, &len, NULL, 0);
+
+            if (count < 1) {
+                return fallback_cpus_number;
+            }
+            return count;
```
fingolfin, line 68: Supposedly this works for FreeBSD, MacOS X, NetBSD, OpenBSD (see https://stackoverflow.com/a/150971/928031); but then again we also have a working `sysconf` call there (but it may be that the `sysconf` call is not available on older versions of these systems; then again, I don't think we really care about those anyway -- or rather, we'll accept patches by people who actually use such old OS versions, but I'd not bother trying to support them proactively.

src/hpc/cpu.c
```diff
@@ -0,0 +1,74 @@
...
+
+#include "hpc/cpu.h"
+
+/****************************************************************************
+**
+*V  SyNumProcessors  . . . . . . . . . . . . . . . . . number of logical CPUs
+**
+*/
+#ifdef NUM_CPUS
+UInt SyNumProcessors = NUM_CPUS;
```
fingolfin, line 19: I wonder: should we even bother to keep `NUM_CPUS`? Your code kinda renders it obsolete, no? the user can still override it at runtime using a command line option anyway. What do you think?
 
@ChrisJefferson any thoughts on that?

fingolfin (changes requested):
Thank you very much for your contribution! I left some remarks as feedback.

Out of curiosity, do you have a specific need this addresses (i.e. are you using HPC-GAP for something?), or did you just do it "for fun" (or whatever)?

nskeip:
Thank you very much for quick reponse!

> Out of curiosity, do you have a specific need this addresses (i.e. are you using HPC-GAP for something?), or did you just do it "for fun" (or whatever)?

For fun, I guess. I just looked through the list of issues in this rep, and felt like this was the one I could try to solve :)

src/hpc/cpu.c
```diff
@@ -0,0 +1,74 @@
...
+**  In case of Windows 7 and higher, calls
+**  GetActiveProcessorCount(ALL_PROCESSOR_GROUPS).
+**  In case of older versions of Windows, tries to load the number of
+**  processors in current processor group via GetSystemInfo with proper
+**  arguments and falls back to the value of 4 in case of an error.
+**  In case of macOS, sysctl is called with HW_AVAILCPU flag.
+**  In case of other operating systems the value of 4 is returned.
+*/
+
+UInt SyCountProcessors(void) {
```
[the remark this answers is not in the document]
nskeip: > If it is not too much trouble, perhaps you could run `clang-format` on this file to apply our code formatting preferences?

Yes, sure. I'll run it.

src/hpc/cpu.h
```diff
@@ -0,0 +1,58 @@
...
+    #elif _WIN32
+        #if _WIN32_WINNT >= _WIN32_WINNT_WIN7
+            #include <windows.h>
+        #else
+            #include <sysinfoapi.h>
+        #endif
+    #elif __APPLE__
+        #include <sys/sysctl.h>
+    #endif
+#endif
```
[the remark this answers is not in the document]
nskeip: Okay, moved them.

src/hpc/cpu.c
nskeip: Yes, calling `sysconf(_SC_NPROCESSORS_ONLN)` makes things really easy. Check out the new code, please.

src/hpc/cpu.c
nskeip: Yes, I agree. Removed this block.

src/hpc/cpu.c
nskeip: Seems like there are not many supercomputers on Windows Vista, so probably it should be skipped.
Removed the code in 46e785ea73678be7efb6d3ad30f6852ff11f3c7f. If it is still needed, I can revert it.

src/hpc/cpu.c
nskeip: > the user can still override it at runtime using a command line option anyway

Didn't know about this option. Seems like `NUM_CPUS` is not needed then.
Removed it in 7d43aa767c9a30b85c9424d2487171f327b1cc13. If it is needed, I can revert it.

nskeip:
Hi, @fingolfin!

I made the changes you requested. Plus introduced a `static const` inside `cpu.c` (see d82b1e7fdefde81985cadc5f1733a331abaaef6f) to get rid of 4 as repeating 'magic number'.

Let me know if anyting should be corrected.

src/hpc/cpu.c
ChrisJefferson: I generally plan on only supporting Windows 10 in the future.

All other versions, except 8.1, already aren't getting security updates, and 8.1 will only get security updates until January 10, 2023.

src/hpc/cpu.c
ChrisJefferson: Yes, I don't think we need this NUM_CPUS (just agreeing)

fingolfin (approved):
Thanks!

fingolfin:
Just FYI I made a follow-up PR #4963, as compilation of HPC-GAP on macOS was broken.

Issue #1043, opened 2016-12-28
IsPositionalObjectRep and IsComponentObjectRep are not consistently set


fingolfin:
(This is taken from issue #897)

Many representation for positional respectively component representations do now have the `IsPositionalObjectRep` respectively `IsComponentObjectRep` filters set. This can lead to surprising behavior and bugs, see e.g. issue #897. According to @markuspf this affects at least these representations in the library (and more in packages):

- `IsSyllableAssocWordRep`
- `IsLetterAssocWordRep`
- `IsBLetterAssocWordRep`
- `IsWLetterAssocWordRep` 
- `IsPermGroupGeneralMapping`
- `IsSubgroupOfWholeGroupByQuotientRep`
- `IsGenericCharacterTableRep`
- `IsNullMapMatrix` 
- `IsDefaultGeneralMappingRep`
- `IsInverseGeneralMappingRep`
- `IsMemberPcSeriesPermGroup`
- `IsSlicedPerm`
- `IsSlicedPermInv`
- `IsFromTheLeftCollectorRep`
- `IsGroupClassByListRep`
- `IsNilpotentLieAutomorphismRep`

(Found by filtering `AllRepresentations()` by not having any of the four Reps above in their IMPL_FLAGS, `AllRepresentations()` being a function @markuspf  added to the library).

I was bitten by this yesterday, as polycyclic's `IsFromTheLeftCollectorRep` suffers from the same problem (and thus a collector was shown to use 300 bytes of RAM when it really are 300,000 bytes...)

So all in all, it seems that `IsPositionalObjectRep` serves no useful purpose, other than perhaps to increase some ranks a little bit. It is not well-documented either: entering `?IsPositionalObjectRep` gives nothing, though it is mentioned in section 79.11 "Positional Objects".

So, what to do? Some random ideas (without consideration how easy and/or efficient it would be to implement them)

1. Change GAP to automatically add set `IsInternalRep`, `IsPositionalObjectRep`, `IsComponentObjectRep`, `IsDataObjectRep` for objects based on their TNUMs.
2. Change `NewRepresentation` to enforce that every newly declared representation sets on of the four representation types
3. Phase out those four types *as filters*, instead use `IS_COMOBJ`, `IS_POSOBJ`, `IS_DATOBJ` in code that really has to act based on the representation kind.
4. A combination of the above.
5. ... other ideas?

[...]

PR #4964, opened 2022-07-29
Objectify: warn if Is{Component,Positional}ObjectRep absent

fingolfin:
This works towards resolving #1043: now `Objectify` warns if `IsPositionalObjectRep` resp. `IsComponentObjectRep` is not set for an object that should have one of them; and it also gives an error if both are set / "the wrong one" is set (i.e. `IsPositionalObjectRep` for a `T_COMOBJ` or `IsComponentObjectRep` for a `T_POSOBJ`). 

I've eliminated all the warnings triggered by loading GAP and running `tst/testinstall.g`, and submitted https://github.com/gap-packages/fr/pull/50 to fix one violation in a package (I've already fixed a bunch of others in packages in the past couple years). I am guessing there may still be a couple more, so I am reluctant to turn the warnings into an error just now... at the very least, before we attempt such a thing, the test suites of all packages should be run against a GAP with that error activated.

Also, before merging this PR we should check how this extra test affects performance: `Objectify` is a bit of a bottleneck, and we don't want to make it worse than it already is...

### DIFF
--- a/lib/alglie.gi
+++ b/lib/alglie.gi
@@ -3975,7 +3975,8 @@ InstallMethod( NormalizedElementOfMagmaRingModuloRelations,
            Append( tlist, todo[i] );
          od;
 
-         return ObjByExtRep( Fam, [ zero, tlist ] );
+         tlist := ObjByExtRep( Fam, [ zero, tlist ] );
+         return [ zero, tlist![2] ];
 
      end );
 
--- a/lib/type1.g
+++ b/lib/type1.g
@@ -569,13 +569,6 @@ BIND_GLOBAL( "FlagsObj", obj -> FlagsType( TypeObj( obj ) ) );
 BIND_GLOBAL( "DataObj", obj -> DataType( TypeObj( obj ) ) );
 
 
-BIND_GLOBAL( "IsNonAtomicComponentObjectRepFlags",
-        FLAGS_FILTER(IsNonAtomicComponentObjectRep));
-BIND_GLOBAL( "IsAtomicPositionalObjectRepFlags",
-        FLAGS_FILTER(IsAtomicPositionalObjectRep));
-BIND_GLOBAL( "IsReadOnlyPositionalObjectRepFlags",
-        FLAGS_FILTER(IsReadOnlyPositionalObjectRep));
-
 #############################################################################
 ##
 #F  Objectify( <type>, <obj> )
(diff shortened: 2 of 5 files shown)


fingolfin:
Hmm, thinking about it, I am not even sure why to warn -- instead we could just add that flag? Perhaps for performance reasons: a missing flag that has to be added in requires us to create a new sub-type...

lib/type1.g
```diff
@@ -592,26 +585,40 @@ BIND_GLOBAL( "Objectify", function ( type, obj )
...
                 fi;
-            elif not IS_SUBSET_FLAGS(flags, IsNonAtomicComponentObjectRepFlags) then
+            elif not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsNonAtomicComponentObjectRep)) then
                 FORCE_SWITCH_OBJ( obj, AtomicRecord(obj) );
             fi;
         fi;
     fi;
     if IS_LIST( obj )  then
+        #Assert(0, IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsPositionalObjectRep)));
+        Assert(0, not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsComponentObjectRep)));
```
fingolfin, line 606: Maybe disable this by default, but have it enabled e.g. in most test suites (as long as the use `START_TEST`)?
```suggestion
        #Assert(1, IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsPositionalObjectRep)));
        Assert(1, not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsComponentObjectRep)));
```

lib/type1.g
```diff
@@ -592,26 +585,40 @@ BIND_GLOBAL( "Objectify", function ( type, obj )
...
+        if not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsPositionalObjectRep)) then
+            INFO_DEBUG(1, "Objectify: type does not imply IsPositionalObjectRep ",
+                 INPUT_FILENAME(), ":", STRING_INT(INPUT_LINENUMBER()));
+        fi;
         SET_TYPE_POSOBJ( obj, type );
     elif IS_REC( obj )  then
+        #Assert(0, IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsComponentObjectRep)));
+        Assert(0, not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsPositionalObjectRep)));
```
fingolfin, line 614:
```suggestion
        #Assert(1, IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsComponentObjectRep)));
        Assert(1, not IS_SUBSET_FLAGS(flags, FLAGS_FILTER(IsPositionalObjectRep)));
```

fingolfin:
I'd like to get this out of my system one way or another:
1. we could just change this PR to add the missing filter (any reason against that)
2. we could just give up and close this
3. we could try to get all packages that are distributed and violated this fixed (but of course there are also many not distributed packages)


Perhaps @ThomasBreuer has some insights on this?

lib/alglie.gi
```diff
@@ -3975,7 +3975,8 @@ InstallMethod( NormalizedElementOfMagmaRingModuloRelations,
            Append( tlist, todo[i] );
          od;
 
-         return ObjByExtRep( Fam, [ zero, tlist ] );
+         tlist := ObjByExtRep( Fam, [ zero, tlist ] );
+         return [ zero, tlist![2] ];
```
ThomasBreuer, line 3979: As far as I see, this change is a bugfix.
(And `NormalizedElementOfMagmaRingModuloRelations` is a misleading name.)

ThomasBreuer (commented):
When `Objectify( type, r )` is called with a record `r` then `SET_TYPE_COMOBJ` is called, and the result admits component access.
When `Objectify( type, l )` is called with a plain list `l` then `SET_TYPE_POSOBJ` is called, and the result admits access to positions.
This is independent of the type `type`, and it fits to the documentation of `Objectify`.
(In other words, apparently `IsComponentObjectRep` is not needed to guarantee component access, and `IsPositionalObjectRep` is not needed to guarantee position access.)

In this sense, it is safe to set automatically `IsComponentObjectRep` in the first case and `IsPositionalObjectRep` in the second case, and to signal an error if this yields an object in `IsComponentObjectRep and IsPositionalObjectRep`.

Perhaps `Objectify` should check that `IsInternalRep` or `IsDataObjectRep` is not set in the given type.

lib/alglie.gi
fingolfin: I wish I had made a note what the error was / why this change is the correct fix *sigh*

ThomasBreuer: `NormalizedElementOfMagmaRingModuloRelations` is called in the GAP library only in `lib/mgmring.gi`. In all cases, `Objectify` is called immediately afterwards with the result. For the old code, this means that the object had already the desired type before the `Objectify` call, and gets objectified again with the same type.

In this sense, I think the library code does not produce wrong results. Nevertheless, the old code contradicts the definition of `NormalizedElementOfMagmaRingModuloRelations`. And there may be code outside `lib/mgmring.gi` which wants to use `NormalizedElementOfMagmaRingModuloRelations`. If the package tests do not show an example where the proposed change yields a different result then we should just apply the proposed fix and add a test, perhaps even as a manual example.

```
gap> R:= GroupRing( Rationals, SymmetricGroup( 3 ) );;
gap> Fam:= FamilyObj( Zero( R ) );;
gap> x:= Sum( GeneratorsOfAlgebra( R ) );;
gap> descr:= [ 0, CoefficientsAndMagmaElements( x ) ];;
gap> NormalizedElementOfMagmaRingModuloRelations( Fam, descr ) = descr;
true
gap> R:= FreeLieAlgebra( Rationals, 2 );;
gap> Fam:= FamilyObj( Zero( R ) );;
gap> x:= Sum( GeneratorsOfAlgebra( R ) );;
gap> descr:= [ 0, CoefficientsAndMagmaElements( x ) ];;
gap> NormalizedElementOfMagmaRingModuloRelations( Fam, descr ) = descr;
true
```
(The first `true` comes from a correct `NormalizedElementOfMagmaRingModuloRelations` method, the second `true` will occur only after the fix.)

ThomasBreuer:
> I'd like to get this out of my system one way or another:
> 1. we could just change this PR to add the missing filter (any reason against that)
> 2. we could just give up and close this
> 3. we could try to get all packages that are distributed and violated this fixed (but of course there are also many not distributed packages)

I suggest variant 1.:
In `Objectify`, set `IsComponentObjectRep` if the argument in question is a record, and set `IsPositionalObjectRep` if the argument in question is a plain list.

And add an example for the `NormalizedElementOfMagmaRingModuloRelations` fix, for example the one shown above.

PR #4965, opened 2022-08-01
CI: build manual & deploy to gh-pages

### LINKED ISSUE #4362: Use GitHub Actions to  upload manual to `gh-pages`
Recently I was bitten by the fact that the manuals on our website, such as https://www.gap-system.org/Manuals/doc/ref/chap0.html, are only available for the last release, but not for the current development version of GAP.

It would be great if we also offered manuals for development snapshots of GAP. We are building the full GAP manual as part of our CI jobs. So all that's left to do is to deploy the result somewhere -- e.g. into the `gh-pages` branch of this repository, making it available on http://gap-system.github.io.

If we do it really cool, we could do it similar to what is usual in the Julia ecosystem, i.e., have...
- http://gap-system.github.io/docs/dev/{ref,tut,hpc} for the version corresponding to master,
- http://gap-system.github.io/docs/v4.11.1/{ref,tut,hpc} for the version corresponding to the v4.11.1 tag (and similar for future tags)
- http://gap-system.github.io/docs/latest/{ref,tut,hpc} for the latest stable release (implemented via a symlink).

Deploying generated HTML into a gh-pages thing is a routine operation with GitHub Actions, i.e, there exist plenty of example actions to do it.

fingolfin:
Resolves #4362 

Tested on my GAP fork, see
- https://github.com/fingolfin/gap/actions/workflows/update-gh-pages.yml
- https://github.com/fingolfin/gap/actions/runs/2773544122
- https://fingolfin.github.io/gap/

To make this work, I've now changed https://github.com/gap-system/gap/settings/pages to deploy from a GH action (see <https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta>). Of course this is a beta feature, so it might break or go away at some point, but then we are not worse off than we are right now.

The built files will appear on http://gap-system.github.io/gap/ but only after this PR has been merged, as it is configured to only work from the `master` branch.

A future improvement would be to tweak the generated file `doc/versiondata` to replace `today` by the actual current date and similar for `this year`

### DIFF
--- a/.github/workflows/update-gh-pages.yml
+++ b/.github/workflows/update-gh-pages.yml
@@ -1,0 +1,84 @@
+# For information about the parts of this workflow,
+# see <https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/>
+# and <https://github.com/actions/starter-workflows/tree/main/pages>
+name: Deploy GAP manual to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: [$default-branch]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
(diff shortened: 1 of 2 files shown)


fingolfin:
Note that there are no code changes in here, and really nothing that CI can test. I mainly post this as a PR to make sure people are aware of it, and for documentation purposes.

PR #4975, opened 2022-08-06
Speed up some stabilizer computations in matrix groups

fingolfin:
Ensure there is a nice monomorphism on the group in which we compute the stabilizers. This speeds up the following example command from the Sophus package:

    AreIsomorphicNilpotentLieAlgebras( L7[100], L7[101] );

Before this patch this takes 13 to 18 seconds on my laptop (timings vary substantially, presumably due to randomization). After this patch, this is down to 0.8 to 0.9 seconds.

### DIFF
--- a/lib/oprt.gi
+++ b/lib/oprt.gi
@@ -1146,6 +1146,11 @@ local   orb,  stb,  rep,  p,  q,  img,  sch,  i,d,act,r,
   fi;
   dict:=NewDictionary(d,true,D);
 
+  if HasIsHandledByNiceMonomorphism(G) and IsHandledByNiceMonomorphism(G) then
+    # ensure a nice monomorphism for the parent has been computed
+    NiceMonomorphism(G);
+  fi;
+
   if IsBound(dopr.stabsub) then
     stabsub:=AsSubgroup(Parent(G),dopr.stabsub);
   else


hulpke (changes requested):
This is weird. I think this hides an issue, not really fix it.

I don't think there is anything about nice monomorphisms that would require the homomorphism to be computed -- the functions call the attribute, and compute it on the fly if needed.
Is it in fact that the test
`HasIsHandledByNiceMonomorphism(G) and IsHandledByNiceMonomorphism(G)`
trigger a nice monomorphism flag that was not set before? (Say, if the matrix group is in characteristic zero, a call to `IsFinite` would be appropriate here.)

Is there a way of reproducing it without the packge? L7 gives an error message for me.

fingolfin:
The patch works because it enables this method, which requires that the parent already has a computed nice monomorphism:
```gap
InstallMethod(NiceMonomorphism,
    "for subgroups that get the nice monomorphism by their parent", true,
    [ IsGroup and IsHandledByNiceMonomorphism and HasParent],
    # to rank higher than matrix group methods.
    {} -> RankFilter(IsFinite and IsMatrixGroup),

function(G)
    local P;

    P :=Parent(G);
    if not (HasIsHandledByNiceMonomorphism(P) 
        and IsHandledByNiceMonomorphism(P) and HasNiceMonomorphism(P)) then
      TryNextMethod();
    fi;
    return NiceMonomorphism(P);
end );
```
As you can see, it checks for `HasNiceMonomorphism(P)` on the parent.

fingolfin:
Here is the full reproducer code.
```gap
LoadPackage("sophus");
L1 := [ AbelianLieAlgebra( GF(2), 1 ) ];;
L2 := [ AbelianLieAlgebra( GF(2), 2 ) ];;
L3 := [ AbelianLieAlgebra( GF(2), 3 ) ];;
Append( L3, Descendants( L2[1], 1 ));
L4 := [ AbelianLieAlgebra( GF(2), 4 ) ];;
for i in L3 do Append( L4, Descendants( i, 1 )); od;
L5 := [ AbelianLieAlgebra( GF(2), 5 ) ];;
for i in L3 do Append( L5, Descendants( i, 2 )); od;
for i in L4 do Append( L5, Descendants( i, 1 )); od;
L6 := [ AbelianLieAlgebra( GF(2), 6 ) ];;
for i in L3 do Append( L6, Descendants( i, 3 )); od;
for i in L4 do Append( L6, Descendants( i, 2 )); od;
for i in L5 do Append( L6, Descendants( i, 1 )); od;
L7 := [ AbelianLieAlgebra( GF(2), 6 ) ];;
for i in L4 do Append( L7, Descendants( i, 3 )); od;
for i in L5 do Append( L7, Descendants( i, 2 )); od;
for i in L6 do Append( L7, Descendants( i, 1 )); od;

AreIsomorphicNilpotentLieAlgebras( L7[100], L7[100] ); time;
#AreIsomorphicNilpotentLieAlgebras( L7[100], L7[101] ); time;
```

hulpke:
Thank you.
So the issue is: If we calculate in a group, we want to inherit a nice monomorphism from a parent, only if it already exists. (This is basically to avoid calculating a nice monomorphism for the parent, if it is not needed.)
Now a stabilizer calculation will do nothing to the *group* G (that is handled by a nice monomorphism, and in which we calculate the stabilizer), but only do membership in the subgroups (the approximate stabilizers) of G calculated on the way.

The reason I'm reluctant to force `NiceMonomorphism` calculations in general code is that once we have composition tree (such as in the `matgrp` package) we actually want to avoid using nice monomorphism, but use the tree.

In this example, I believe that just calling `Size` on G. I'm trying to think what would do so if the size is pre-set -- maybe a membership test of a non-generator.
In fact this same issue then would apply to other stabilizer methods.

fingolfin:
I've now modified the `sophus` package to invoke `Size` on the group before passing it to the stabilizer computation. That still gives the desired speed-up.

PR #4977, opened 2022-08-08
Add IsVecOrMatObj, add filters to several MatrixObj implementations

fingolfin:
- Add IsVecOrMatObj category
- Move some declaration from matobj2.gd to matobj1.gd
- Set various filters for MatrixObj implementations

Since I moved code *and* edited it, you may wish to look at the individual commits to see what exactly I changed.

This is motivated by https://github.com/gap-system/gap/pull/4935/files#r940054140


fingolfin:
@hulpke @frankluebeck this PR started out small and motivated by a change in Frank's PR, but it lead me to discover that `FieldOfMatrixList` was broken, and then some more changes ensued... Please review (I recommend looking at each commit separately, their commit message contain extra thoughts).

Maybe @ThomasBreuer also is interested in taking a look

lib/matobj1.gd
```diff
@@ -264,3 +264,111 @@ DeclareCategory( "IsMatrixObj", IsMatrixOrMatrixObj );
 ##
 DeclareCategory( "IsRowListMatrix", IsMatrixObj );
 
+
```
wucas, line 267: Why do two files exist and what is supposed to be in one or the other?

lib/matobjnz.gi
```diff
@@ -1640,3 +1640,20 @@ local i, n, ords, base, imat, vec, one,cp,op,zero,fam;
...
     return cp;
 end );
+
+
+##
+
+InstallOtherMethod( DegreeFFE, "for IsZmodnZVectorRep",
+    [ IsZmodnZVectorRep ],
+function(vec)
+    # TODO: check that modulus is a prime
```
wucas, line 1650: That implementation is missing.

wucas (changes requested):
In `lib/matobjnz.gi` there's a TODO left.

hulpke (approved):

frankluebeck (approved):
So, `IsVecOrMatObj` is meant as a temporary filter such that in various places two declarations can be substituted by a single one?
This commit can be reverted when we manage to get rid of hidden implications?

frankluebeck (commented):
I cannot comment on this commit because I'm not aware of the rationale behind having two files `matobj1.gd` and `matobj2.gd`.

frankluebeck (approved):
It looks sensible to me to mention these filters in the declaration of the representation instead of adding immedite methods later.

frankluebeck (commented):
I cannot comment on this since `FieldOfMatrixList`  is not documented.

frankluebeck (approved):

lib/matobjnz.gi
```diff
@@ -1640,3 +1640,20 @@ local i, n, ords, base, imat, vec, one,cp,op,zero,fam;
...
+    return 1;
+end);
+
+InstallOtherMethod( DegreeFFE, "for IsZmodnZVectorRep",
+    [ IsZmodnZMatrixRep ],
```
frankluebeck, line 1655: Second argument is wrong. Maybe also a bit simpler:
```
InstallOtherMethod( DegreeFFE, ["IsZmodnZMatrixRep"], m-> 1);
```

lib/matobjnz.gi
frankluebeck: Maybe a bit simpler:
```
InstallOtherMethod( DegreeFFE, ["IsZmodnZVectorRep"], v-> 1);
```

frankluebeck (changes requested):

lib/grpffmat.gi
```diff
@@ -75,9 +73,7 @@ local   deg,  mat,  row,  char,  B;
     elif Is8BitMatrixRep(mat) then
       deg:=LcmInt( deg, Length(FactorsInt(Q_VEC8BIT(mat![2]))));
```
frankluebeck, line 74: This case seems unnecessary. The code after `else` calls an efficient method.

frankluebeck (changes requested):

fingolfin:
- for now `IsVecOrMatObj` is intended to be a temporary workaround, removing it shouldn't be hard.
- `matobj1.gd` and `matobj2.gd` are split for technical reasons: the first file provides definition that many other files need, and so have to be provided early; the other uses/references things that are provided by other .gd files that are loaded much later. So the split is needed unless one wants to severely reorder the order the other .gd files are loaded
- `FieldOfMatrixList` is documented in `.gd` file with a GAPDoc documentation comment, which however is indeed not included in the reference manual -- not sure if that's intentional or not. Anyway, the comment outlines quite specifically what the intent of the function is; and indeed we use it with the expectation that it returns a minimal field containing all elements in various places.

lib/grpffmat.gi
fingolfin: The `else` method calls a default method which iterates over all rows; this check only checks a single row, under the assumption that in a `Is8BitMatrixRep` all rows are defined over the same field... I have not checked if this is even correct to do; and of course it could also be achieved by installing a dedicated `DegreeFFE ` method for `Is8BitMatrixRep`. But for now I wanted to keep the changes to a minimum to reduce the risk of breaking more than I am fixing...

lib/matobj1.gd
fingolfin: This is explained at the top of these two files: one is read earlier in the GAP loading process than the other, to deal with various dependencies between declarations in multiple files.

lib/matobj1.gd
```diff
@@ -19,6 +19,16 @@
 ##
 
 
+# This is an internal filter mostly meant for declarations of methods that
+# are meant to apply to bother VectorObj and MatrixObj implementations;
```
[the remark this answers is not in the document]
fingolfin: yes, thanks

lib/matobjnz.gi
fingolfin: Not sure what you mean? The fact there is an unresolved TODO here?

Well, it's a TODO for future work, the current code also is not performing this check, and in principle one may wish to do this not in that function but e.g. when constructing the matrix, and then carrying around the info in an `isprime` value... perhaps the modulus and the `isprime` bool could also be stored in the type or family... I did not want to tackle these decisions here, and considering that `DegreeFFE` does not specify what it does for an invalid argument, I figure returning `1` here while not great is also not the worst...

lib/grpffmat.gi
fingolfin: That said, I'll just add DegreeFFE methods for `IsGF2MatrixRep` and `Is8BitMatrixRep` in the next update for this PR

fingolfin: Aha, but actually a dedicated `DegreeFFE` for `Is8BitMatrixRep` still would be de-optimization: because it would then end up checking all entries of the matrix to determine the smallest field; while the existing code here just looks at a single row to determine the desired field / degree (as in this function, we are not looking for the smallest field containing all elements, just *some* field)

fingolfin:
@frankluebeck @wucas @ThomasBreuer does this look OK now to you? further questions? I'd like to merge this ASAP, then we can possibly update PR #4983 or move forward in some other way so that we can get 4.12.0 out...

lib/matobjnz.gi
wucas: Thank for the explanation! I was simply not sure if that was an oversight and wanted to make sure.

wucas (approved):

lib/matobjnz.gi
fingolfin: Sure! I appreciated it as such, just wasn't sure if I understood the question right :-)

frankluebeck (approved):

PR #4980, opened 2022-08-10
Re-enable all cygwin features, remove non-functional tests

### LINKED ISSUE #4955: CI: timeout in `testmockpkg testinstall - - windows-2019`

ChrisJefferson:
After trying quite a bit, I can't make the tests which spawn external processes and do IO with them work on github actions -- it seems any program which GAP starts (including another GAP) cannot communicate with GAP.

This lets us re-enable xdg-utils, and therefore fix #4955 properly

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -166,11 +166,16 @@ jobs:
       - uses: gap-actions/setup-cygwin@v1
         if: ${{ runner.os == 'Windows' }}
         with:
-          # HACK/FIXME/TODO: workaround for https://github.com/gap-system/gap/issues/4955
-          # which essentially removes xdg-utils ; its presence leads GAP to think the
-          # Cygwin environment is Unix, not Windows, which enables some tests which then
-          # hang.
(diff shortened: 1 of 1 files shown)


fingolfin:
What advantage does this workaround have over the previous workaround? You write that you "re-enable xdg-utils, and therefore fix https://github.com/gap-system/gap/issues/4955 properly" but what benefit does it then have to have xdg-utils available?

ChrisJefferson:
In the long term, my preference (which I'm happy to discuss in more depth) is that our cygwin install always has xdg-utils (and in general, a full cygwin install) provided, and `ARCH_IS_UNIX` returns true (because Cygwin is a lot more like a UNIX than windows).

fingolfin (approved):

PR #4982, opened 2022-08-10
GAPInfo: add NonBareDependencies for use of package manager

mtorpey:
If GAP is loaded with `--bare`, `GAPInfo.Dependencies` is overwritten with an empty list.  One obvious use-case of `--bare` is loading GAP without packages for just long enough to use PackageManager to install required packages.  PackageManager has the `InstallRequiredPackages` function for this purpose, but since it requires `GAPInfo.Dependencies`, it ironically becomes useless in the one circumstance where it's really needed!

This commit would allow the normal required packages to be found in `GAPInfo.NonBareDependencies` even if `--bare` is used, which will allow PackageManager to install required packages without the need to hard-code anything.

**Release notes**: use title.

### DIFF
--- a/lib/system.g
+++ b/lib/system.g
@@ -33,6 +33,10 @@ BIND_GLOBAL( "GAPInfo", rec(
 # There is no SuggestedOtherPackages here because the default value of
 # the user preference PackagesToLoad does the job      
 
+    # If run with --bare, this means that the normal dependencies
+    # can still be found by the package manager
+    NonBareDependencies := ~.Dependencies,
+
     HasReadGAPRC:= false,
 
     # list of all reserved keywords


fingolfin (changes requested):
I am not a fan of putting ever more stuff into GAPInfo, it is already overcrowded here.

Instead, I'd suggest to change how `--bare` is handled: Change this
```gap
    if CommandLineOptions.bare then
      CommandLineOptions.A := true;
      GAPInfo.Dependencies := MakeImmutable(rec( NeededOtherPackages := [] ));
    fi;
```
by removing the line handling `GAPInfo.Dependencies`. Then instead modify `AutoloadPackages` to honor `CommandLineOptions.bare`

mtorpey:
Thanks for the tip!

Replaced by #4985.

PR #4984, opened 2022-08-10
Try to keep in StringRep when appending strings

ChrisJefferson:
I was getting some strange behaviour in the 'JSON' package, and I realised it was caused by code like:

```
gap> s := "";;
gap> o := OutputTextString(s, false);;
gap> WriteAll(o, ['a', 'b', 'c']);
true
gap> s;
"abc"
gap> IsStringRep(s);
false
```

What is happening is that the `WriteAll` uses `Append`, and `Append(stringrep,plist)` results in a `plist`, even if the result is a string. This doesn't result in bugs (well, it did in my code, but I've fixed that separately), but does still slow things down and create memory bloat.

This PR tries harder to keep the string as a stringrep -- it could possibly be written more efficiently, but I couldn't find a better way reusing existing pieces of code.

### DIFF
--- a/src/listfunc.c
+++ b/src/listfunc.c
@@ -225,7 +225,10 @@ static Obj FuncAPPEND_LIST_INTR(Obj self, Obj list1, Obj list2)
     RequireSmallList(SELF_NAME, list2);
 
     /* handle the case of strings now */
-    if (IS_STRING_REP(list1) && IS_STRING_REP(list2)) {
+    if (IS_STRING_REP(list1) && IS_STRING(list2)) {
+        if (!IS_STRING_REP(list2)) {
+            list2 = ImmutableString(list2);
+        }
         AppendString(list1, list2);
         return 0;
     }
--- a/tst/testinstall/stringobj.tst
+++ b/tst/testinstall/stringobj.tst
@@ -5209,5 +5209,25 @@ true
 gap> "IsHomogeneousList" in CategoriesOfObject("");
 true
 
+# Append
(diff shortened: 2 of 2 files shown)


src/listfunc.c
```diff
@@ -225,9 +225,16 @@ static Obj FuncAPPEND_LIST_INTR(Obj self, Obj list1, Obj list2)
...
+      if(IS_STRING(list2)) {
+        if(IS_STRING_REP(list2)) {
+          AppendString(list1, list2);
+        }
+        else {
+          AppendString(list1, ImmutableString(list2));
+        }
         return 0;
+      }
     }
```
fingolfin, line 238: Code formatting!

How about simplifying to this:
```suggestion
    if (IS_STRING_REP(list1) && IS_STRING(list2)) {
        if (!IS_STRING_REP(list2)) {
            list2 = ImmutableString(list2);
        }
        AppendString(list1, list2);
        return 0;
    }
```

Of course even simpler would be this, but yeah, we kinda wanted to get away from automatic in-place conversions, so yeah...
```suggestion
    if (IS_STRING_REP(list1) && IsStringConv(list2)) {
        AppendString(list1, list2);
        return 0;
    }
```

ChrisJefferson:
I took your simpler change (thanks!) I agree `ConvString` might be faster, but it would also be nice to reduce the number of those! (in particular on arguments which we aren't changing).

fingolfin (approved):

PR #4985, opened 2022-08-11
Handle --bare in AutoloadPackages instead of modifying GAPInfo.

mtorpey:
This means that `GAPInfo.Dependencies` still contains the list of required packages even if GAP is loaded with `--bare`, which allows the package manager to look up the required packages correctly and install them when `InstallRequiredPackages` is called.

This replaces #4982 after a suggestion from @fingolfin.

Release notes: none.


fingolfin (approved):
Thanks!

PR #4990, opened 2022-08-16
Various updates to the documentation

fingolfin:
- remove references to -win.zip format
- clarify that the `gap` executable can now be started directly
- remove references to obsolete line break handling
- remove helper code for converting line breaks
- remove references to non-existent service for converting package
  source code archives (we used to do that, but no more)
- mention `make install`
- update some FAQs
- mention build prerequisite also in INSTALL.md
- many more tweaks

## Text for release notes

## Further details


frankluebeck (approved):

fingolfin:
@ThomasBreuer the functions in `lib/lbutil.g` certainly could be useful for someone, but then again they were not loaded by default and as such hard to find if one didn't know they exist; plus lots of command line tools (e.g. `dos2unix`) exist also for this purpose. And in the worst case it is easy to recover this code from the git history

frankluebeck:
I had introduced the functions in `lbutil.g` quite a while ago to have such conversion functions available under all operating systems.

But nowadays the different line endings are not a big practical problem anymore (since most editors on all systems can deal with them) . And, of course, at least on Unix/Linux there are standard utilities (which existed already when these functions were introduced).

So, I'm happy with removing `lbutil.g` from the distribution.

PR #4991, opened 2022-08-16
Document the `-c` command line option

### LINKED ISSUE #4196: Reference manual should document the `-c EXPR` option
It is not currently described under https://www.gap-system.org/Manuals/doc/ref/chap3.html#X782751D5858A6EAF

fingolfin:
Resolves #4196

### DIFF
--- a/doc/ref/run.xml
+++ b/doc/ref/run.xml
@@ -86,6 +86,14 @@ prints the prompt. This is useful when, after a while, you get  tired  of
 the banner. This option can be repeated to enable the banner; each <C>-b</C>
 toggles the state of banner display.
 </Item>
+<Mark><Index Key="-c"><C>-c</C></Index>
+<C>-c </C><A>gapcode</A></Mark>
+<Item>
+tells &GAP; to execute the given &GAP; code as if it was entered into
+a temporary file which then is processed together with the other files
+given to GAP (see the explanation at the end of this list for further
+details on how filenames are processed). 
+</Item>
 <Mark><Index Key="-D"><C>-D</C></Index>
 <C>-D</C></Mark>
 <Item>
@@ -391,9 +399,10 @@ redirect the output to a printer, you may want to increase this value.
 <Item>
 Further arguments are taken as filenames of files that are read by &GAP;
 during startup, after the system  and private init   files are read,  but
-before the first prompt is  printed.
-The files are read in the order in which they appear on the command line.
-&GAP; only accepts up to 14 filenames on the command line.
+before the first prompt is  printed. Filenames ending with <C>.tst</C>
+are processed by <Ref Func="Test" />, all other files by <Ref Oper="Read" />.
+These files and also commands specified via the <C>-c</C> option are processed
+in the order in which they appear on the command line.
 If a file cannot be opened &GAP; will print an error message and will abort.
 </Item>
 </List>


doc/ref/run.xml
```diff
@@ -391,9 +399,10 @@ redirect the output to a printer, you may want to increase this value.
...
 Further arguments are taken as filenames of files that are read by &GAP;
 during startup, after the system  and private init   files are read,  but
-before the first prompt is  printed.
-The files are read in the order in which they appear on the command line.
-&GAP; only accepts up to 14 filenames on the command line.
+before the first prompt is  printed. Filenames ending with <C>.tst</C>
+are processed by <Ref Func="Test" />, all other files by <Ref Oper="Read" />.
+These files and also commands specified via the <C>-c</C> option are processed
+in the order in which they appear on the command line.
 If a file cannot be opened &GAP; will print an error message and will abort.
```
fingolfin, line 406: Off-topic, but: this final sentence (which I did not touch, it was here before) is factually wrong: if the file cannot be opened then GAP prints `Could not read file "FILANEM".` and happily chugs on. Should I update the documentation to match the code, or the other way around? Personally I'd prefer to change the code, as it almost always means I made a typo, and sometimes I don't notice; if GAP would instead exit with a non-zero exit code, that'd help me find bugs earlier. But on the downside, who knows, maybe this change would break someones workflow somewhere?

frankluebeck (changes requested):
The description of `-c` looks good.

doc/ref/run.xml
frankluebeck, line 406: Since this last sentence is wrong, it should be corrected in this pull request.

I would prefer that the behaviour is not changed and only the documentation is adjusted. Currently, one can do some debugging in the running GAP.

How about:
```
If a file cannot be opened by &GAP; or an argument of  the <C>-c</C> option cannot be evaluated then &GAP; will print an error message and continue to start up.
```

fingolfin: Of course the situation is even more complex, because one also should explain what happens with additional filenames resp. `-c` options after the one which failed: if the "failure" was a syntax error, we continue processing the others; if it was an error then we abort processing the others. That's IMHO not very intuitive, so it should then also be explained.

fingolfin: I've tried to write something and thought some more, and I really don't see how this final behavior is sensible or useful: if at all, shouldn't it be reverse? I.e. if the user decides to proceed after an `Error`, after being prompted what to do, we may as well continue to process, as the user is well-aware there was a problem. On the other hand, a syntax error is already just a message; it can easily be overlooked esp. if subsequent code prints a lot. So by effectively ignoring it, we make debugging *harder* in my eyes; I'd thus understand if we aborted processing after a syntax error.

On the other hand, if the idea really is that syntax errors are "fine" (why, though??), then why do we still abort processing of the file the syntax error occurred in (by using `READ_NOCOVERY` instead of `READ`) yet happily proceed to process the next file?

This really don't make sense to me....


That all said: I'd rather merge this PR first, which contains a genuine improvement, and *then* worry about the broken documentation text (which was around for years and years and nobody noticed), instead of holding the PR hostage...

ThomasBreuer (approved):
The proposed changes are fine.
Concerning the question how to fix the wrong statement about the handling of errors that occur when files and code specified in the command line are processed, I agree with @fingolfin that the current behaviour is not consistent. Perhaps we can merge the pull request and keep a comment that reminds us about the open question.

fingolfin:
I've merged this as-is and filed issue #4994 to remind us to deal with the discrepancy. Feel free to comment there, we can then perhaps resolve it for 4.12.1

PR #4992, opened 2022-08-16
Fix Gcd to return StandardAssociate

frankluebeck:
Method for Gcd of multivariate polynomials did not always return the
standard associate of the result.

Added an example as test.

## Text for release notes

Fixed `Gcd` of multivariate polynomials to always return a standard associate element.


frankluebeck:
The test failure of "CI / testpackages testinstall-loadall - ABI=32 NO_COVERAGE=1 - ubuntu-18.04 (pull_request) " seems unrelated to this PR.

fingolfin (approved):

PR #4997, opened 2022-08-17
CI: NormalizInterface doesn't need libboost anymore

fingolfin:


frankluebeck (approved):
I could successfully install the current release 1.3.4 of `NormalizInterface` on a system without `libboost`.

PR #4998, opened 2022-08-17
INSTALL.md: list some GAP package prerequisites

fingolfin:


INSTALL.md
```diff
@@ -157,8 +157,33 @@ On other operating systems, you will need to figure out equivalent commands
...
+(excluding GMP, which many packages require):
+
+- 4ti2Interface
+  - Debian/Ubuntu: 4ti2
+  - Homebrew: (not currently available)
+- alnuth
+  - Debian/Ubuntu: pari-gp
+  - Homebrew: pari
+- browse
+  - Debian/Ubuntu: libncurses5-dev
```
ThomasBreuer, line 170: better just `libncurses-dev`? (in fact `libncurses6-dev` seems to be used on my machine)

ThomasBreuer (approved):
Good idea. I have added one minor comment.

INSTALL.md
fingolfin: Thanks for the pointer. I copied what I had added a long time ago to the CI setup, I simply was not aware of this. I've adjusted this PR and also the CI workflow to use `libncurses-dev`

PR #5000, opened 2022-08-17
Adjust tests for GAPDoc 1.6.6

fingolfin:
This is rather ugly, but it was by far the easiest and quickest workaround to implement on short notice. After the 4.12 release we can remove the `@else` part and only keep the GAPDoc 1.6.6 output around. For now I just want to get on with the release

### DIFF
--- a/tst/testinstall/package.tst
+++ b/tst/testinstall/package.tst
@@ -495,7 +495,63 @@ new methods:
   mockpkg_Property( ... )*
 
 
-# Test the Cite() command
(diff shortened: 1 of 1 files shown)


ChrisJefferson (approved):

wilfwilson:
Congratulations on bagging PR 5000 😉

frankluebeck (approved):
It would be good to find an easy way to specify several outputs as acceptable in the CI tests, such that a hack like this one is not needed.

PR #5005, opened 2022-08-17
Release notes for GAP 4.12, various updates to the tooling

fingolfin:
For your convenience, a preview of the release notes is also available at https://hackmd.io/0NfdKaGqTGaaYjVTsUbb3Q -- this makes it very easy to open certain PRs (to edit their title or labels, or inspect the changes they contain). However, please don't edit there, as those changes would be lost when I next update it).

My plan is to merge this tomorrow. We can then still tweak it if needed, but the release must happen tomorrow.


pedritomelenas:
I saw several entries devoted to `IteratorOfPartitionsSet`, and wondered if they could be merged into a single entry

fingolfin:
@pedritomelenas good catch! In fact the second mention should be dropped, as it fixes a bug in this function; but that function is new, so nobody only uses release versions of GAP was ever exposed to it

pedritomelenas:
@fingolfin, there are three entries mentioning  `IteratorOfPartitionSet`

dev/releases/release_notes.readme.md
```diff
@@ -52,36 +67,10 @@ Currently the following labels are regarded as relevant for the creation of rele
...
-| 7 | Improvements in the experimental way to allow 3rd party code to link GAP as a library (libgap) | `topic: libgap` | PRs that are related to libgap |
-| 8 | Improvements in for the **Julia** integration | `topic: julia` | PRs that are related to the Julia integration |
-| 9 | Changed documentation | `topic: documentation` | PRs improving the documentation |
-| 10 | Packages | `topic: packages` | PRs related to package handling, or specific to a package (for packages w/o issue tracker) |
-| 11 | Other changes | `release notes: needed` OR `release notes: added` | PRs relevant for the release notes but without a label that fits --in each case, think about assigning such a label or introducing a new one |
+  The script `generate_release_notes.py` groups the pull request based on labels in a list called  `prioritylist`. Each pull request gets assigned to the **first** group/subheading that corresponds to one of its labels. This list can be adjusted over time as needed. Note that any such ordering is always a compromise, there is no best way to do it.
```
ThomasBreuer, line 70: Apparently it makes no sense to give an explicit list of the relevant labels, since this list changes.
Perhaps replace `This list can be adjusted over time as needed.` by an explicit sentence that before processing the relevant pull request with `generate_release_notes.py`, one should fetch the list of labels in question, and adjust the script accordingly.

ThomasBreuer (approved):
The list of release notes looks o.k.
(Some longer text form the "Text for the release notes" section of pull requests did not make it into the release notes, as is discussed in issue #5007, but this is actually not a problem.)

I have added one comment about the documentation.

fingolfin:
@pedritomelenas ah, I only saw two because the third has a spelling mistake *sigh* thanks again for keeping a close watch

dev/releases/release_notes.readme.md
fingolfin: Thanks for the feedback. My apologies, but I don't understand the suggestion: from where should be one "fetch the list of labels in question" ? I am not sure what I am supposed to write here, a concrete suggestion would perhaps clarify this...

That said, the view I see it, both the set of labels, how we apply them, *and* the choice which of them are used for grouping, and how, are all in flux. So one may argue that instead of fetching labels from elsewhere, one might as well add/change/remove labels elsewhere... So what we really need is some clear concise guidelines how to apply labels, and which labels are used (to avoid the list of de-syncing, this could simply point at the relevant portions of the script)

fingolfin: As I need to move forward with the release process, I am going to merge this PR now; that's not intended to shut down this discussion, though; we can still adjust this later on, after all

ThomasBreuer: @fingolfin Sorry for the confusion.
I wanted to say that creating release notes with `generate_release_notes.py` relies on the labels that are actually used in the relevant pull requests; some of them may be new, some labels used in the script may have been renamed, etc.
Thus the first step should be to fetch all labels from the pull requests. Perhaps `generate_release_notes.py` should be extended to collect these labels, compare this list with the labels which it knows already from its `prioritylist`, and then output the newly found labels. Before the next run of `generate_release_notes.py`, `prioritylist gets adjusted.

PR #5022, opened 2022-09-02
Fix claims about emacs support in the reference manual

fingolfin:
We stopped shipping this some time ago, as it was (a) outdated
and (b) the FSF asked us to. So don't claim we ship them, instead
point to a better alternative.


danielrademacher (approved):

fingolfin:
Backported to stable-4.12 via f33860a84

PR #5024, opened 2022-09-02
Fix performance regression (caused by changes in `SmallGeneratingSet`)

fingolfin:
The early abort condition `Length(gens)<=Length(AbelianInvariants(G))+2`
was inserted in 9d380c4091451fe8f6b8f5335c8bfc8b76a51556 to speed up
calculations in certain large cases. However this caused regressions
in various package test suites, so bdf94cb1f963f83e6931243e00096fb0b0a1d1bd
attempted to address this by reordering some computations. Alas, that
was not enough: the YangBaxter test suite got slowed down by an order of
magnitude (it went from about 30 seconds to several minutes on my laptop).

This commit attempts a middle ground: we still abort immediately if we
provably reached an optimum. But the problematic abort condition is now
used a bit later, and only for non-solvable groups or solvable groups
with up to 3 generators.

### DIFF
--- a/lib/grpperm.gi
+++ b/lib/grpperm.gi
@@ -1987,6 +1987,11 @@ local  i, j, U, gens,a,mintry,orb,orp,isok;
 
   gens := ShallowCopy(Set(GeneratorsOfGroup(G)));
 
+  # we can't do better than rank(G/G') generators...
+  if Length(gens)<=Length(AbelianInvariants(G)) then
+    return gens;
+  fi;
+
   # remove obvious redundancies...
   # sort elements by descending order; trivial permutations
   # at the end
@@ -2006,7 +2011,8 @@ local  i, j, U, gens,a,mintry,orb,orp,isok;
     i := i + 1;
   od;
 
-  if Length(gens)<=Length(AbelianInvariants(G))+2 then
+  # we can't do better than rank(G/G') generators...
+  if Length(gens)<=Length(AbelianInvariants(G)) then
     return gens;
   fi;
 
@@ -2017,6 +2023,11 @@ local  i, j, U, gens,a,mintry,orb,orp,isok;
       IsSolvableGroup(G)
       and Length(gens)>3 then
     return MinimalGeneratingSet(G);
+  fi;
+
+  # stop here if close to rank(G/G')
+  if Length(gens)<=Length(AbelianInvariants(G))+2 then
+    return gens;
   fi;
 
   # store orbit data


hulpke (commented):
I don't think this is really a solution. Basically it forces again a solvability test (and subsequent `MinimalGeneratingSet`) in many cases which was what the initial change was trying to avoid, as it can be costly in larger degree groups. It might avoid the problem in this particular package, but other calculations will slow down instead.

I wonder whether the underlying issue in something else, related to the YangBaxter package, or routines called by it. Is it a few examples, or an overall slowdown?
Presumbly it is that a `SmallGeneratingSet` returned is not as optimal as it could be.
But what calls `SmallGeneratingSet`? Should it maybe rather use `MinimalGeneratingSet` if the group is solvable?

fingolfin:
The problem occurs in `AllHomomorphismClasses` invoked with relatively small inputs.
Here is a concrete example:
```gap
G:=Group([ (1,5,2,6)(3,7,4,8)(9,10,11), (1,7,2,8)(3,5,4,6)(9,10,11), (1,3)(2,4)(5,7)(6,8)(9,10,11),
  (1,4)(2,3)(5,8)(6,7)(9,10,11) ]);;
H:=Group([ (1,15,6,14,3,17,2,16,5,13,4,18)(7,21,12,20,9,23,8,22,11,19,10,24), (1,21,6,20,3,23,2,22,5,19,4,24)(7,15,12,14,9,
    17,8,16,11,13,10,18), (1,9,5,7,3,11)(2,10,6,8,4,12)(13,21,17,19,15,23)(14,22,18,20,16,24),
  (1,10,5,8,3,12)(2,9,6,7,4,11)(13,22,17,20,15,24)(14,21,18,19,16,23) ]);;
all:=AllHomomorphismClasses(G,H);;
time;
```

When I run this in GAP 4.12.0, sometimes it takes 70 milliseconds, but sometimes it takes several seconds; the worst I saw was 12 seconds, so about 170 times slower than the "best case". In GAP 4.11.1, it consistently takes ~45 milliseconds, which I was never able to match in 4.12.

Indeed, with this test loop:
```
while true do
G:=Group([ (1,5,2,6)(3,7,4,8)(9,10,11), (1,7,2,8)(3,5,4,6)(9,10,11), (1,3)(2,4)(5,7)(6,8)(9,10,11),
  (1,4)(2,3)(5,8)(6,7)(9,10,11) ]);;
H:=Group([ (1,15,6,14,3,17,2,16,5,13,4,18)(7,21,12,20,9,23,8,22,11,19,10,24), (1,21,6,20,3,23,2,22,5,19,4,24)(7,15,12,14,9,
    17,8,16,11,13,10,18), (1,9,5,7,3,11)(2,10,6,8,4,12)(13,21,17,19,15,23)(14,22,18,20,16,24),
  (1,10,5,8,3,12)(2,9,6,7,4,11)(13,22,17,20,15,24)(14,21,18,19,16,23) ]);;
t:=Runtime();
all:=AllHomomorphismClasses(G,H);;
Print(Runtime()-t,"\n");
od;
```
I just got this result in GAP 4.12.0:
```
119
88
120
57
55
119
57
#I  The group tested requires many generators. `AllHomomorphismClasses' often
#I  does not perform well for such groups -- see the documentation.
2337
85
1365
55
91
#I  The group tested requires many generators. `AllHomomorphismClasses' often
#I  does not perform well for such groups -- see the documentation.
7100
87
118
89
117
122
55
53
57
56
56
118
121
54
119
53
120
86
54
61
119
58
58
76
54
85
52
54
55
55
226
55
117
55
54
52
56
56
53
60
120
57
55
53
117
55
116
85
119
57
91
57
87
56
117
350
86
120
57
91
55
57
85
89
56
56
54
1354
57
55
59
124
#I  The group tested requires many generators. `AllHomomorphismClasses' often
#I  does not perform well for such groups -- see the documentation.
4465
115
118
53
481
227
123
115
114
56
56
54
#I  The group tested requires many generators. `AllHomomorphismClasses' often
#I  does not perform well for such groups -- see the documentation.
12622
57
121
55
56
...
```

hulpke:
Thank you -- this is easily fixed by having `AllHomomorphismClasses` force a `MinimalGeneratingSet` for solvable groups.
(Surprised that someone uses this toy operation in a production setup.)

fingolfin:
@hulpke great, and I see you already have a patch for that. do you mind if I submit just that commit from your work branch as a PR?

hulpke:
@fingolfin 
I have submitted a PR that contains this change amongst others

PR #5026, opened 2022-09-03
CI: switch from ubuntu-18.04 to ubuntu-latest

### LINKED ISSUE #5009: CI: ubuntu-18.04 environment is deprecated
CI runs now report this:

> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

Shouldn't be too hard. Volunteers to write the PR?

fingolfin:
The former will be removed in early 2023

Resolves #5009

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-latest]
         shell: [bash]
         test-suites:
           [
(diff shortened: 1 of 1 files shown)


fingolfin:
Unfortunately the 32bit support (resp. availability of binaries) for several packages was removed in the latest Ubuntu version, which breaks our 32bit package tests. Of course we could just build the relevant libraries ourselves from scratch, or just not test those packages. Or just stop bothering to test packages in 32bit mode..

ChrisJefferson:
My personal temptation is to drop support -- I find it increasingly difficult to test 32-bit GAP (I've not got it to work in Ubuntu 22.04 on WSL for example, I think it's technically possible but I got weird linker messages).

If we are going to withdraw support officially, I'd be tempted to add some kind of message when a 32-bit GAP is run, to say it's not supported anymore and might have bugs and asking for an email to gap-support for anyone using it, so we can find out if there is a large enough group of people who need 32-bits we are willing to go to the effort of testing it?

fingolfin:
> If we are going to withdraw support officially, I'd be tempted to add some kind of message when a 32-bit GAP is run, to say it's not supported anymore and might have bugs and asking for an email to gap-support for anyone using it, so we can find out if there is a large enough group of people who need 32-bits we are willing to go to the effort of testing it?

Fine by me. Open an issue for that, or just do it?

fingolfin:
I've changed this now to *retain* the 32bit package tests using ubuntu-18.04 for now. Of course that means they'll break in early 2023. But it gives us the benefit for the other tests *now*, without debating on what to do with the 32bit tests, which we can discuss elsewhere.

ChrisJefferson (approved):

PR #5027, opened 2022-09-03
Do not define unused functions with SYS_DEFAULT_PATHS

jamesjer:
GCC complains about functions that are defined but not used if SYS_DEFAULT_PATHS is defined.  Wrap them with `#ifndef SYS_DEFAULT_PATHS` ... `#endif`, and move `SyDotGapPath` down so there is a single block of such definitions.

## Text for release notes

None

## Further details

None


fingolfin (approved):
Thanks

fingolfin:
Backported to `stable-4.12` in 3a266fb6c113ffbcfa3693153501484fb3b06e92

Issue #5020, opened 2022-09-01
problem with "nice monomorphisms"


ThomasBreuer:
(The following holds for released GAP versions as well as for the master branch.)

The problem mentioned in the first comment of oscar-system/Oscar.jl/pull/1520 is essentially the following.

We start with a matrix group.
```
gap> G:= Group( [ [ [ 0, 1 ], [ 1, 0 ] ] ] );;
gap> IsHandledByNiceMonomorphism( G );
true
```
Apparently it would be good for this group to store a nice monomorphism. We know such a monomorphism, so we try to help GAP by setting it.
```
gap> vecs:= Orbit( G, [ 1, 0 ], OnRight );
[ [ 1, 0 ], [ 0, 1 ] ]
gap> acthom:= ActionHomomorphism( G, vecs, OnRight );
<action homomorphism>
gap> SetNiceMonomorphism( G, acthom );
```
Now the following happens.
```
gap> Kernel( acthom );
Error, recursion depth trap (5000) in
  IsEvenInt( Length( attr ) ) at /.../gap/lib/grp.gi:4596 called from
...
```
The reason for this problem is that the `Kernel` method (more precisely, the `KernelOfMultiplicativeGeneralMapping` method) for the action homomorphism `acthom` wants to delegate to a `GroupHomomorphismByImagesNC` object `hom`, say, which is created from the given data, and that the source of `hom` is a group that knows the `NiceMonomorphism` object `acthom`.

In this situation, `GroupHomomorphismByImagesNC` creates `hom` as a composition mapping of `acthom` with another mapping. But then calling `Kernel` with `hom` will first try to compute the kernel of `acthom`, which yields the infinite recursion.

Setting nice monomorphisms by hand is apparently not safe. What would be needed to avoid this kind of error?

If one sets the `IsInjective` flag for `acthom` in the above example before calling `Kernel` then the error disappears. Is this the only requirement for safely setting nice monomorphisms?

And if yes, would it be a solution (and not only a workaround) to extend the dedicated `SetNiceMonomorphism` method in `lib/grpnice.gi` by an `IsInjective` check? Note that the automatically produced `NiceMonomorphism` objects get this flag during their construction, thus the additional check would be cheap for them.

fingolfin:
[...]

PR #5029, opened 2022-09-05
Document and enforce that "nice monomorphisms" must have `IsInjective` set (otherwise an infinite recursion can happen)

ThomasBreuer:
This addresses issue #5020.
We check whether the map that shall be set as nice monomorphism has already the `IsInjective` flag, and signal an error if not.

The documentation mentions/recommends in the introduction of the section on nice monomorphisms that one can set a known monomorphism with stored `IsInjective` flag as a nice monomorphism by hand.
(The Tutorial had already an example for that.)

### DIFF
--- a/doc/tut/group.xml
+++ b/doc/tut/group.xml
@@ -1216,6 +1216,11 @@ only if this map knows that it is injective.
 In the above example, this is the case because the map is created as the
 composition of two maps (a nice monomorphism and an isomorphism)
 which know that they are injective.
+In general, the injectivity flag of a map <C>map</C> gets set if one tests
+the property by calling <C>IsInjective( map )</C>;
+if one is sure that <C>map</C> is injective and wants to avoid the
+potentially expensive test then one can set the flag by explicitly
+calling <C>SetIsInjective( map, true )</C>.
 <P/>
 <E>Summary.</E>  In this section we have  seen  how calculations in groups
 can be carried  out in isomorphic  images in  nicer groups. We  have seen
--- a/doc/tut/group.xml
+++ b/doc/tut/group.xml
@@ -1211,6 +1211,12 @@ Reconstructing the
 automorphism group may of course result in the  loss of other information
 &GAP; had already gathered, besides the (not-so-)nice monomorphism.
 <P/>
+Prescribing a known map as the nice monomorphism of a group is possible
+only if this map knows that it is injective.
+In the above example, this is the case because the map is created as the
+composition of two maps (a nice monomorphism and an isomorphism)
+which know that they are injective.
+<P/>
 <E>Summary.</E>  In this section we have  seen  how calculations in groups
 can be carried  out in isomorphic  images in  nicer groups. We  have seen
 that &GAP;  pursues this technique  automatically for certain classes of
--- a/lib/grpnice.gi
+++ b/lib/grpnice.gi
@@ -1057,12 +1057,14 @@ InstallMethod( NiceMonomorphism, "SeedFaithfulAction supersedes", true,
         [ IsGroup and IsHandledByNiceMonomorphism and
 	  HasSeedFaithfulAction], 1000,
 function(G)
-local b;
+  local b, hom;
   b:=SeedFaithfulAction(G);
   if b=fail then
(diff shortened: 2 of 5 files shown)


hulpke (approved):
Thank you

doc/tut/group.xml
```diff
@@ -1211,6 +1211,12 @@ Reconstructing the
 automorphism group may of course result in the  loss of other information
 &GAP; had already gathered, besides the (not-so-)nice monomorphism.
 <P/>
+Prescribing a known map as the nice monomorphism of a group is possible
+only if this map knows that it is injective.
+In the above example, this is the case because the map is created as the
+composition of two maps (a nice monomorphism and an isomorphism)
+which know that they are injective.
```
fingolfin, line 1218: Perhaps this and/or the other new text could mention that one can ensure this is the case by either invoking `IsInjective(map)` (safe but potentially slow) or `SetIsInjective(map, true)` (fast but the caller has to make sure this is correct) ?

lib/grpnice.gi
```diff
@@ -16,11 +16,22 @@
 ##
 #M  SetNiceMonomorphism(<G>,<hom>)
 ##
-##  This setter method is special because we want to tell every nice
-##  monomorphism that it is one.
+##  We install a special setter method because we have to make sure that only
+##  injective maps get stored as nice monomorphisms.
+##  More precisely, we the stored map must know that it is injective,
```
fingolfin, line 21:
```suggestion
##  More precisely, the stored map must know that it is injective,
```

lib/grpnice.gi
```diff
@@ -16,11 +16,22 @@
...
+##  otherwise computing its kernel may run into an infinite recursion.
+##  (The maps computed by 'NiceMonomorphism' have the 'IsInjective' flag,
+##  the test affects only those maps that shall be set by hand as
+##  nice monomorphisms.)
```
fingolfin, line 25: Strictly speaking this is just the case for all the *current* such methods. Anyone adding more such methods in the future should also be aware of it, I think? So perhaps the sentence here should either be changed to acknowledge that -- or just remove it? I don't think it is all that useful: we want to enforce an invariant, period. That some code is (supposedly) already correctly enforcing the invariant is immaterial, isn't it?

lib/grpnice.gi
```diff
@@ -103,6 +114,9 @@ InstallMethod( GroupByNiceMonomorphism,
 function( nice, grp )
     local   fam,  pre;
 
+    if not IsInjective( nice ) then
+      Error( "<nice> is not injective" );
```
fingolfin, line 118:
```suggestion
      Error( "<nice> is not known to be injective" );
```

fingolfin (approved):
I approve in general but perhaps you may wish to address some or all of my comments.

BTW I don't really think this fits in the category "bug fix", but rather it improves the documentation and explicitly enforces an invariant, so it stops the user from shooting themselves into the foot. OK, so maybe it is a bug fix then, but a bug fix for the documentation / specification of nice monomorphisms. Not that it really matters, except for where to put this in the release notes 😂

fingolfin:
Do we need / want this to be backported to 4.12 ? I have no issue with that, though strictly speaking it is not necessary, and any backported change has a risk of causing a regression somewhere...

ThomasBreuer:
@fingolfin Thanks for your suggestions, I am going to update the pull request.

Concerning the question whether the changes are a bugfix, I would argue that the starting point was an infinite recursion which is now avoided. Of course one gets an error in this situation also after the changes, but we have replaced an "unexpected error" (which is one of our bug classes) by an "expected error" (not a bug); 

Concerning the question whether we should backport the changes to 4.12, I have no strong opinion. On the one hand, the Oscar code that uncovered the problem is safe now; on the other hand, the same problem may be hit in other situations, and then the new error handling makes it easy to solve this problem.

lib/grpnice.gi
ThomasBreuer: Well, the current code actually checks injectivity, thus the message is correct.
The question is whether we should instead check `HasIsInjective`, and signal an error if not.

PR #5031, opened 2022-09-05
Small fixes and a first step towards using `ClassicalMaximals`

hulpke:
This PR provides fixes for minor inconveniences that did not make 4.12.
It also adds code that will use `ClassicalMaximals` for the purpose of finding maximal subgroups of L2(q) (supplementing SL(2,5)'s if neccessary).

### DIFF
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -4052,7 +4052,7 @@ InstallGlobalFunction( DirectSumMat, function (arg)
           if IsSubset(c,F) then
             F:=c;
           else
-            Error("build field/ring");
+            F:=Field(Concatenation(GeneratorsOfField(F),GeneratorsOfField(c)));
           fi;
         fi;
       od;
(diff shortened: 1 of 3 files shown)


lib/matrix.gi
```diff
@@ -4052,7 +4052,7 @@ InstallGlobalFunction( DirectSumMat, function (arg)
           if IsSubset(c,F) then
             F:=c;
           else
-            Error("build field/ring");
+            F:=Field(Concatenation(GeneratorsOfField(F),GeneratorsOfField(c)));
```
fingolfin, line 4055: I noticed there is an (undocumented, but used) operation `ClosureField`, an alias for `ClosureDivisionRing` (which has a GAPDoc comment, but it is not linked to by the reference manual). I don't think it does any better right now, but in theory a more clever implementation could be added there one day? So perhaps:
```suggestion
            F:=ClosureField(F,c);
```

fingolfin (approved):

PR #5032, opened 2022-09-05
Fix performance regression in `AllHomomorphismClasses` and possibly other code searching isomorphisms

fingolfin:
Since this is costly, it is worth to force a minimal generating set for
solvable groups. This resolves problems the yangbaxter package has had.

Close #5024

This patch is taken from @hulpke's PR #5031, but having different, unrelated improvements and bug fixes on separate PRs reduces the work needed to make GAP releases quite a bit (as it allows us to deal with those PRs in automation, instead of adding them to the list of PRs that need manual work), and is also more convenient for reviewers, so I am splitting it off here.

For reference, here is the motivation for this patch from the description of PR #5024:

> The early abort condition `Length(gens)<=Length(AbelianInvariants(G))+2`
was inserted in 9d380c4091451fe8f6b8f5335c8bfc8b76a51556 to speed up
calculations in certain large cases. However this caused regressions
in various package test suites, so bdf94cb1f963f83e6931243e00096fb0b0a1d1bd
attempted to address this by reordering some computations. Alas, that
was not enough: the YangBaxter test suite got slowed down by an order of
magnitude (it went from about 30 seconds to several minutes on my laptop).

This PR addresses this.

### DIFF
--- a/lib/morpheus.gi
+++ b/lib/morpheus.gi
@@ -1746,8 +1746,11 @@ InstallGlobalFunction(Morphium,function(G,H,DoAuto)
 local len,combi,Gr,Gcl,Ggc,Hr,Hcl,bg,bpri,x,dat,
       gens,i,c,hom,free,elms,price,result,rels,inns,bcl,vsu;
 
-  IsSolvableGroup(G); # force knowledge
-  gens:=SmallGeneratingSet(G);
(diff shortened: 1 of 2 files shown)


danielrademacher (approved):

fingolfin:
Backported to stable-4.12 via fa8d50a11

PR #5034, opened 2022-09-07
INSTALL.md: remove "Expert Windows Installation"

fingolfin:
This section was partially outdated (reference to spaces in paths is hopefully obsolete?) and in general seems of little use these days, seeing as it mostly just recommends to use the .exe installer, which we recommend anyway...

Also removing some trailing bits at the end of `INSTALL.md` which may or may not be supposed to be part of the remove section, but which seem out of place in any case.


fingolfin:
Backported to stable-4.12 via a45db7446

PR #5036, opened 2022-09-07
Fix `Centre` for pc groups giving wrong results, e.g. `Size(Center(SmallGroup(2^9,261648)))` gave 4 instead of 8

### LINKED ISSUE #3940: Bug in Centre for PcGroups
The following calculation

    G:=SmallGroup(2^9,261648);;
    Size(Center(G));

returns a center of order 4, which is wrong (One can easily check that the center has order 8).
The error persists when converting ->FP->PC, but does not arise when converting ->Perm->Fp->PC, which indicates that it is related to the particular PC presentation.

This was reported by Benjamin Sambale

fingolfin:
Fixes #3940 with a change provided by @beick.

### DIFF
--- a/lib/grppcatr.gi
+++ b/lib/grppcatr.gi
@@ -717,37 +717,15 @@ end);
 #F  NextStepCentralizer( <gens>, <cent>, <pcgsF>, <field> )
 ##
 NextStepCentralizer := function( gens, cent, pcgsF, field )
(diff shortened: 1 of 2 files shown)


fingolfin:
Normally I'd not do this, but I'll merge this anyway: the code is not actually by me, but by @beick. However, I reviewed it, and it seems correct to me. So I'll just count this as approval, and merge this now.

fingolfin:
Backported to stable-4.12 via 7a0499209

PR #5041, opened 2022-09-09
Fix unexpected error in `MinimalGeneratingSet` for solvable non-pc groups

fingolfin:
It could end up returning nothing and not invoking TryNextMethod().
This caused a failure in the hap test suite.

This is a recent regression in 4.12.0, introduced in commit 97feedb5c078eef4a24ecf4e535295e31e3e0218


ThomasBreuer (approved):
Looks good. Thanks.

fingolfin:
Backported to stable-4.12 via d11ab9608

PR #5044, opened 2022-09-09
doc: add heading to `Centralizer` entry

fingolfin:
This fixes a reference to `Centralizer` in the xgap manual, and also matches what we do for `Normalizer`


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 via b576deedb

PR #5047, opened 2022-09-13
`Display` for a character table optionally shows the degrees of the character fields

ThomasBreuer:
Show the degrees of the field extensions of the shown characters if the `characterField` component of the options record given to `Display` is set to `true`; the default is `false`.


fingolfin (approved):

fingolfin:
Backported to stable-4.12 via commit ab9ba171e

PR #5048, opened 2022-09-14
Makefile: signal error via exit code

fingolfin:
... not just via a message printed to the user.

I noticed this because an automated build silently *passed* when I forgot to run `./configure`. Ouch.

While this hurt me quite a bit, overall I think this is rather obscure and does not warrant an entry in the release notes.


fingolfin:
Backported to stable-4.12 via commit 4b06cbb83

PR #5049, opened 2022-09-14
Execute #@exec earlier to allow #@if to depend on it

zickgraf:
If one wants to use code from packages in #@if, one can now load those packages via #@exec.

The first commit fixes some typos.

## Text for release notes

see title


lib/test.gi
```diff
@@ -346,7 +349,11 @@ end);
 ##  <C>#@if</C> are used until either a <C>#@else</C> or <C>#@fi</C> is
 ##  reached. If a <C>#@else</C> is present then the code after the <C>#@else</C>
 ##  is used if and only if <C>EXPR</C> evaluated to <K>false</K>. Finally,
-##  once <C>#endif</C> is reached, evaluation continues normally.
+##  once <C>#fi</C> is reached, evaluation continues normally.
+##  <P/>
+##  Note that <C>EXPR</C> is evaluated after the lines marked with <C>#@exec</C> but before
```
fingolfin, line 354:
```suggestion
##  Note that <C>EXPR</C> is evaluated after all <C>#@exec</C> lines have been executed but before
```

fingolfin (approved):
Seem plausible to me. Clarifying the context in which `EXPR` is evaluated is a very good idea in any case!

Would be good to know what @ChrisJefferson thinks.

ChrisJefferson:
This looks good to me. It would be nice to have a test (it's always nice to have a test), but I've checked it works so I'd accept without.

fingolfin:
@zickgraf could you please squash the final commit into its parent (so two commits remain) ? I'd normally do this myself, but it seems you turned off the "Allow edits and access to secrets by maintainers" checkbox, so I can't.

(Might also wrap the new documentation lines to < 80 characters)

fingolfin:
Thanks!

PR #5059, opened 2022-09-16
Fix `AllHomomorphisms` regression

fingolfin:
The problem is that we only checked `IsSolvableGroup` to decide whether to call `MinimalGeneratingSet` on a group. However, in reality this is only implemented for groups in the filter `CanEasilyComputePcgs` -- thus for perm and pc groups but not for fp groups.

This broke the HAP test suite.

This regression was introduced in PR #5032 and is not in GAP 4.12.0, so it also does not need a release notes entry.


fingolfin:
@hulpke does this seem sensible to you?

ThomasBreuer (approved):
Yes, I agree with the analysis of the problem, and with the solution.

fingolfin:
Backported to stable-4.12 via commit 2cd2903eb

Issue #5062, opened 2022-09-20
one more vector * matrix problem


ThomasBreuer:
The following happens in GAP 4.11, GAP 4.12, and in the master branch.
(It was observed by Bernhard Böhmler.)
```
gap> z:= Z(3,11);;
gap> m:= [ [ Z(3)^0 ] ];;
gap> ConvertToMatrixRep( m, 3 );;
gap> z * m;
[ [ z ] ]
gap> Zero( z ) * m;
Error, cannot set filter for internal object called from
SetFilterObj( pv, IsLockedRepresentationVector );
[...]
```
The problem is that the matrix `m` is written over `GF(3)`, that `Zero( z )` is an element of this field but not in `IsInternalRep`, and that the first choice `\*` method (for scalar and 8 bit matrix) assumes that in this situation the product of the scalar and a row of the matrix is in `Is8BitVectorRep`.
A possible workaround (solution?) would be to call `TryNextMethod` if the scalar is not in `IsInternalRep`.
Perhaps there are better solutions. If not then I will open a pull request with this idea.

fingolfin:
This was fixed by PR #5066

PR #5066, opened 2022-09-25
fix multiplication of `IsFFE` and `Is8BitMatrixRep`

ThomasBreuer:
(and the other way round)
by installing the current methods only for mutiplying with internal FFEs and adding new methods for multiplying with non-internal FFEs.

(This addresses issue #5062.)


fingolfin (approved):

fingolfin:
Backported to stable-4.12 via commit ee8a61df8

PR #5067, opened 2022-09-25
fixed `AbsolutelyIrreducibleModules` for a pc group and a finite prime field

### LINKED ISSUE #5061: problem with `AbsolutelyIrreducibleModules`
The following happens in GAP 4.12 and in the master branch.
It was observed by Bernhard Böhmler.

The documentation of `AbsolutelyIrreducibleModules` says

>  71.15-2 AbsolutelyIrreducibleModules
>  
>  ‣ AbsolutelyIrreducibleModules( G, F, dim ) ────────────────────── operation
>  
>  returns a list of length 2. The first entry is a generating system of G. The
>  second  entry  is  a  list of all absolute irreducible modules of G over the
>  field  F  in  dimension  dim,  given  as  MeatAxe modules (see GModuleByMats
>  (69.1-1)).

For the dihedral group of order 10, we get

```
gap> g:= DihedralGroup( 10 );;
gap> t:= CharacterTable( g ) mod 2;;
gap> List( Irr( t ), x -> x[1] );
[ 1, 2, 2 ]
gap> List( Irr( t ), x -> SizeOfFieldOfDefinition( x, 2 ) );
[ 2, 4, 4 ]
gap> reps2:= AbsolutelyIrreducibleModules( g, GF(2), 10 );;
gap> Length( reps2[2] );
2
gap> List( reps2[2], r -> [ r.dimension, r.field ] );
[ [ 1, GF(2) ], [ 2, GF(2^2) ] ]
gap> reps4:= AbsolutelyIrreducibleModules( g, GF(4), 10 );;
gap> Length( reps4[2] );
3
gap> List( reps4[2], r -> [ r.dimension, r.field ] );
[ [ 1, GF(2^2) ], [ 2, GF(2^2) ], [ 2, GF(2^2) ] ]
```
First of all, the documentation should be corrected, the argument `dim` is some upper bound for the dimensions of the modules in question.

As for the returned list of modules, I think that the result of `AbsolutelyIrreducibleModules( g, GF(2), 10 )` should be the list that contains just the trivial representation.
There is in fact a generic method for `AbsolutelyIrreducibleModules` that does this:
```
gap> meth:= ApplicableMethod( AbsolutelyIrreducibleModules, [ g, GF(2), 10 ], 0, "all" );
[ function( G, F, dim ) ... end, function( G, F, dim ) ... end ]
gap> List( meth[2]( g, GF(2), 10 )[2], r -> [ r.dimension, r.field ] );
[ [ 1, GF(2) ] ]
```
Moreover, I do not understand why `AbsolutelyIrreducibleModules` requires three arguments. The dimension bound should be optional.
And it would look more natural to enter the characteristic as the second argument, not a prescribed field over which the modules live; in this case, the modules could be written over the minimal field, similar to the above (wrong) result of `AbsolutelyIrreducibleModules( g, GF(2), 10 )`.

ThomasBreuer:
The two methods now give equivalent results when both are applicable (for a pc group and a finite prime field).

This resolves #5061.

### DIFF
--- a/lib/grppcrep.gi
+++ b/lib/grppcrep.gi
@@ -587,10 +587,8 @@ end );
 ##
 #M AbsolutIrreducibleModules( <G>, <F>, <dim> ). . . . . . .up to equivalence
 ##
-## <dim> is the limit of Dim( F ) * Dim( M ) for the modules M
-##
 InstallMethod( AbsolutIrreducibleModules,
-    "generic method for groups with pcgs",
+    "method for group with pcgs and finite prime field",
     true, 
     [ IsGroup and CanEasilyComputePcgs, IsField and IsFinite and IsPrimeField, IsInt ],
     0,
@@ -598,7 +596,9 @@ InstallMethod( AbsolutIrreducibleModules,
 function( G, F, dim )
     local modus;
     modus := AbsAndIrredModules( G, F, dim );
-    return [modus[1],List( modus[2], x -> x.absirr )];
+    return [ modus[1],
+             Filtered( List( modus[2], x -> x.absirr ),
+                       x -> IsPrimeField( x.field ) ) ];
 end );
 
 #############################################################################
--- a/lib/grpreps.gd
+++ b/lib/grpreps.gd
@@ -22,10 +22,13 @@
 ##  <Oper Name="AbsolutIrreducibleModules" Arg='G, F, dim'/>
 ##
 ##  <Description>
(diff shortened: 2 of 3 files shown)


hulpke (approved):

fingolfin (approved):

fingolfin:
Backported to stable-4.12 via commit dce0c1add

PR #5070, opened 2022-09-25
Makefile.gappkg: move pkgconfig.h.in from gen to src

fingolfin:
The `gen` directory is for things that are created by `make` and it is deleted by `make clean`. But `pkgconfig.h.in` is *not* generated by `make`. So it does not belong into `src` but rather into `gen`.

Unfortunately GAP's `BuildPackages.sh` script runs `make clean` before `make`, which triggered this problematic behavior, causing build failures in `io` and `curlInterface`.


danielrademacher (approved):

fingolfin:
Backported to stable-4.12 via commit 859d0b2ae

Issue #4809, opened 2022-03-02
problem with `PreImagesRepresentative`


ThomasBreuer:
The following happens in GAP 4.11 as well as in the current master branch.
```
gap> G:= SymmetricGroup( 4 );;  H:= SylowSubgroup( G, 2 );;
gap> gens:= GeneratorsOfGroup( H );;
gap> f:= GroupHomomorphismByImages( H, G, gens, gens );;
gap> x:= (1,2,3,4);;
gap> x in Range( f );
true
gap> x in Image( f );
false
gap> PreImagesRepresentative( f, x );
(1,2)(3,4)
gap> y:= (1,2,3);;
gap> y in Range( f );
true
gap> y in Image( f );
false
gap> PreImagesRepresentative( f, y );
Error, List Element: <list>[2] must have an assigned value in
  img := opr( img, S.transimages[bimg[1]] )
[...]
```
The documentation of `PreImagesRepresentative` says:

>  ‣ PreImagesRepresentative( map, elm ) ──────────────────────────── operation
>   
>  If  elm  is  an  element  of  the  range  of  the  general  mapping map then
>  PreImagesRepresentative  returns  either  a  representative  of  the  set of
>  preimages  of  elm  under  map or fail, the latter if and only if elm has no
>  preimages under map.
>  
>  Anything may happen if elm is not an element of the range of map.

Thus I want to get `fail` for each of the inputs `PreImagesRepresentative( f, x )`, `PreImagesRepresentative( f, y )`.

fingolfin:
See also #4088

fingolfin:
Also perhaps vaguely relevant, at least on a long-term scaler: #574

hulpke:
I understand where the request comes from, but simply enforcing a membership test is going to slow a lot of calculations down. Can we

1. Create a NC version first
2. Go through the library and packages and see that they use the NC version whenever possible
3. Only then change the current version to enforce tests.

The same issuemight hold in some cases for `ImagesRepresentative`

ThomasBreuer:
@hulpke Thanks.

I understand your step 1 as follows: Copy the existing methods for `PreImagesRepresentative`, install them for a new operation `PreImagesRepresentativeNC`, and document this new operation such that the second argument is assumed to be an element in the image of the first argument.

[...]

PR #28, opened 2015-03-11
List extensions: allow arbitrary objects as list "indices"

fingolfin:
The changes in this pull request make it possible to implement methods for [], []:=, etc. which take arbitrary objects as "index", not just positive integers. At the same time, it unifies some code and fixes a few places which should have only accepted positive integers, but actually accepted arbitrary integers.

The original motivation for this patch was to make it possible to implement matrix methods which take a pair of integers, like so:
```gap
   a[[i,j]] := 2;
```
to se the entry in row i, column j of the matrix a to 2. 
Internally, this would invokes the []:= method with the parameters  "a" and "[i,j]", and the method can then set the matrix entry. E.g. for IsMatrixObj objects, it could invoke SetMatElm.

There are other applications, however, e.g. for nicer interfaces to dictionaries: 
```gap
  dict["key"]  := value;
  if IsBound(dict["key"]) then
    ...
```

The biggest missing thing in this PR is a documentation update, and some review by somebody other than Steve and me. If the changes look sane, we could merge them now, and coument the changes later.

BTW, for the matrix application, it would be even nicer if we had a language extensions that would allow directly writing
```gap
   a[i,j] := 2;
```
i.e. without the double brackets. But this is apparently harder to achieve (?), and could still be done later, and I think the changes presented here are useful for other applications (such as dictionaries / maps).

Issue #26, opened 2015-03-10
HPC-GAP grpperm.tst Diff


markuspf:
Running the grpperm.tst in HPC-GAP on my machine does not only take an exorbitant amount of time (4 minutes in GAP, more than 10 in HPC-GAP) it also Diffs:

```
########> Diff in tst/grpperm.tst, line 76:
# Input is:
NrMovedPoints(g/h);
# Expected output:
168
# But found:
112
########
```

I can reproduce this with the current hpcgap-default branch on my laptop and lovelace, one of our Linux test machines. Hence, it is not just a fluke caused by my obscure operating system (DragonFly 4.1-DEVELOPMENT).

alex-konovalov:
Thanks. This diff is known and safe to ignore for now (though the test probably should be reworked to run smoothly). It comes from a different degree of a permutation representation. 

The slowdown should be investigated indeed.

alex-konovalov:
I've tried to have a look at slowdown, but at the moment I can't run `grpperm.tst` in HPC-GAP. The following example is broken:
```
gap> g:=WreathProduct(MathieuGroup(11),Group((1,2)));
<permutation group of size 125452800 with 5 generators>
gap> Length(ConjugacyClassesSubgroups(g));
Error, Attempt to write object 4413342128 of type object (weakptr) without having write access in
  while i <= Length( f )  do
    a := ElmWPObj( f, i );
...
...
od; on line 1117 of file /Users/alexk/GITREPS/hpcgap-fork/lib/pcgsperm.gi
```
and the next as well, with the backtrace pointing to the same location at `pcgsperm.gi`:
```
gap> g:=SemidirectProduct(GL(3,5),GF(5)^3);
<matrix group of size 186000000 with 3 generators>
gap> g:=Image(IsomorphismPermGroup(g));
```
Looks like this is a side-effect of PR #51.

markuspf:
I am having a hard time reproducing this. It seems to mainly happen in travis tests. I have only seen this fail once on my laptop and now have run testbugfix.g and tesstravis.g on lovelace multiple times and these examples just work.

Even though this is the mapping code, the error happens in a region that (as far as I can see) does not want to write to the weakptr object. I have tried (in my own fork) to use atomic readwrite, but something is odd here.

I will open a new issue as soon as I have more information.

ChrisJefferson:
This could be an issue with weakptrs -- I don't know if hpc-gap actually does anything with weak ptrs yet, probably worth having a look.

[...]

PR #5073, opened 2022-09-29
NC versions of PreImages, PreImagesSet, PreImagesElm and PreImagesRepresentative

cdwensley:
This aims to start work on the process outlined in issue #4809.  The operation PreImagesRepresentative has been renamed PreImagesRepresentativeNC throughout the library, many doc files, and some tests.  A new PreImagesRepresentative has been introduced which just, for now, calls PreImagesRepresentativeNC.  If this change is included in 4.12.1 then package authors can be given the option of converting their calls of this operation.  After that a test can be introduced to check that the element is in the range of the map.  Further changes may be recommended, e.g. a new PreImageElmNC. 

## Text for release notes
Too early to write this down.


fingolfin:
Thanks for working on this!

So the new `PreImagesRepresentative` method is in `lib/mapping.gi`.

Once major concern I have is how we plan to transition packages which install methods for `PreImagesRepresentative`. Those will be broken if we start calling `PreImagesRepresentativeNC` everywhere now without giving them a chance to provide such methods themselves!

So I think we need to go the other way around: we *first* must work with package authors to give them the chance to adapt and release their packages, *then* we can move forward. Here is one way this could be done:

- perform the renaming here, but for now leave `PreImagesRepresentative` as a synonym for `PreImagesRepresentativeNC` (so that methods installed for the one are also installed for the other)
- identify all packages which install methods for `PreImagesRepresentative`; on a quick scan these seem to be
    - [ ] cryst
    - [ ] fga
    - [ ] fr
    - [ ] orb
    - [ ] polycyclic
    - [ ] qpa
    - [ ] rcwa
    - [ ] semigroups
    - [ ] utils
    - [ ] wedderga
- either tell the author which changes to make, or directly submit a PR with the required changes
   - I suggest the following: they should install their methods for `PreImagesRepresentativeNC` instead of `PreImagesRepresentative`; but to stay compatible with prior GAP versions, they can add code to their `init.g` of the kind `if not IsBound(PreImagesRepresentativeNC) then BindGlobal("PreImagesRepresentativeNC", PreImagesRepresentative); fi;`
   - wait until they all made releases

That's how I've done similar migrations in the past. It's a slow and tiresome process, Ia m afraid... Here is a pull request from 2019 by me for FGA which still has not been merged, much less released: https://github.com/chsievers/fga/pull/7. On the upside, all the other packages have a fairly good track record for releases (well perhaps except for `polycyclic`, of which I am maintainer; there the background is that we have a major bug that must be fixed and I can't seem to get around to it... ARGH).

Anyway: because this is so tedious, it's best to not do this too often. So here one may wonder if there aren't other methods that need similar treatment and that should be dealt with in one go. E.g. I wonder whether we also need `ImagesRepresentativeNC` for similar reasons?

cdwensley:
There were 3 files where it appeared that after a call to PreImagesRepresentativeNC a test for the return of fail followed immediately: alghom.gi, ghomperm.gi, mapphomo.gi.

cdwensley:
I am very perplexed by some methods in mapprep.gi which are full of tests for "im = fail".  Should these be methods for the non-NC PreImagesRepresentative?

cdwensley:
@fingolfin  Sorry if I'm slow on the uptake, but I do not understand what will be broken.  As it stands at present PreImagesRepresentative just calls PreImagesRepresentativeNC (without any test).  So, it seems to me, everything should work as before, and changes to packages can come later. 
The Utils package installs a method for PreImagesRepresentative and tests this, and the test works fine in my preimrep branch.

lib/field.gi
```diff
@@ -1329,7 +1329,7 @@ InstallMethod( PreImagesElm,
     [ IsFieldHomomorphism, IsObject ],
     function ( hom, elm )
     if IsInjective( hom ) = 1 then
-      return [ PreImagesRepresentative( hom, elm ) ];
+      return [ PreImagesRepresentativeNC( hom, elm ) ];
```
fingolfin, line 1332: @cdwensley To give an example of what might break if packages are not adapted first: if this function is called on a homomorphism provided by a package, it will, after the changes in this PR, suddenly produce an error, because it invokes `PreImagesRepresentativeNC` yet the package (not having been adapted yet to changes) only provides a method for `PreImagesRepresentativeNC`.

cdwensley: Without understanding why it happens, trying out the Semigroups tests has convinced me that errors do occur.  So, following your "other way around", in what way does "perform the renaming here, but for now leave PreImagesRepresentative as a synonym for PreImagesRepresentativeNC (so that methods installed for the one are also installed for the other)" differ from what has been done so far?  Using the alternative 
    DeclareSynonym( "PreImagesRepresentative", PreImagesRepresentativeNC );
the Semigroups tests now pass.  Any further suggestions?

lib/mapping.gd
```diff
@@ -1032,6 +1033,17 @@ DeclareOperation( "PreImageElm",
 ##  <P/>
 ##  Anything may happen if <A>elm</A> is not an element of the range of
 ##  <A>map</A>.
+##  <P/>
+##  In order to improve this state of affairs, 
+##  following the discussion in issue number 4809,
```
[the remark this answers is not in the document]
cdwensley: Changes have been made as you suggest and will be submitted shortly.

lib/field.gi
cdwensley: Utils now uses PreImagesRepresentativeNC as suggested above.

cdwensley:
PreImages produces the same error as PreImagesRepresentative, so the same procedure applies.

cdwensley:
Further details of the use by packages of one or more of PreImages, PreImagesElm, PreImagesSet and PreImagesRepresentative. 
In the following list 'I' indicates that an InstallMethod for one of these functions has been declared; while 'L', 'T' and 'M' indicate an occurrence in the library; tests or examples, and the manual. 
ace               T 
alnuth            T M 
atlasrep        L 
automgrp        L T M 
autpgrp         L 
congruence      L 
crisp           L T 
cryst         I L 
ctbllib           T M 
cubefree        L 
difsets         L
fga           I   T M 
fining        I L
fr            I L T M 
grpconst        L 
guarana         L T 
hap             L   M 
irredsol        L  
JupyterKernel   L 
kan             L 
laguna          L 
liealgdb        L 
liering         L T M 
matgrp        I L 
modison         L
orb           I 
permut          L
polenta             M 
polycyclic    I L T M 
qpa           I L T M 
radiroot        L T M 
rcwa          I L T M 
recog           L
repsn           L 
rewriting       L
semigroups    I L T M 
smallsemi       L 
sonata          L
sophus          L 
transgrp        L 
utils         I L T M 
walrus        I L T M 
xmod            L
xmodalg         L 
YangBaxter      L

cdwensley:
When nothing happens for weeks one can forget exactly what has been done. 
But who would want the hassle of reviewing a PR with 77 files changed?

doc/tut/group.xml
```diff
@@ -168,9 +168,9 @@ action domain of <C>norm</C>. (It only happens that both are subsets of the
 natural numbers.) We can now form  images and preimages under the natural
 homomorphism. The set of  preimages of an element under  <C>hom</C> is a coset
 modulo <C>elab</C>.
-We use the function <Ref Func="PreImages" BookName="ref"/> here because
```
ChrisJefferson, line 171: I think this can just be left as `PreImages`, in general in documentation we don't encourage people to use `NC` methods of functions. Of course at the moment it doesn't make any difference.

lib/alghom.gi
```diff
@@ -906,7 +906,7 @@ PreImagesRepresentativeOperationAlgebraHomomorphism := function( ophom, mat )
     return mat;
 end;
 
-InstallMethod( PreImagesRepresentativeNC,
+InstallMethod( PreImagesRepresentative,
```
ChrisJefferson: Is the idea here that these methods already do all appropriate checking, so we don't need to attach "NC" to them?

ChrisJefferson: The reason I ask is that there are two possible issues this raises:

1) We should probably have tests which make sure that make sure this is a "non-NC" method.

2) Installing a method for PreImagesRepresentative, but not for PreImagesRepresentativeNC, means there could be cases where PreImageRep works, while PreImageRepNC doesn't.

I realise that these two things are currently issues, as PreImageRep and PreImageRepNC are the same operation with different names, but it's probably worth having a long term plan for handling this before installing methods for both PreImagesRep and PreImagesRepNC in different places.

ChrisJefferson:
I started writing this in a comment, but it got too long, so I'll pop it here.

We could consider adding something like: `InstallMethods([PreImagesRepresentativeNC, PreImagesRepresentative], ...`, which is like `InstallMethod` but lets you install methods for multiple things at once -- except `InstallMethods` could only install once when the objects are identical. This would give a compact way of handling the case of having one method for both `PreImagesRepresentativeNC` and `PreImagesRepresentative`.

This could, in principle, then be used in other places, for just having a quick search for `NC` I notice in ctbl.gi:3345, we install the same method for `Index`, `IndexOp` and `IndexNC`.

doc/tut/group.xml
ThomasBreuer: @ChrisJefferson Is this rule formulated somewhere in the documentation?

(If we do not want to encourage using `NC` variants in the documentation then this pull request should not change examples from the manual to use `PreImagesRepresentativeNC` instead of `PreImagesRepresentative`.)

I would agree that the `NC` variants are mainly intended to be used inside functions where it is clear from the context that the tests are not necessary. On the other hand, one can argue that the same holds for an interactive GAP session, for example, if one wants to compute a preimage of an element that was just constructed as an element of the range of the mapping in question.

ChrisJefferson: @ThomasBreuer Looking around when defining function we often define the NC and non-NC version at the same time, but when we reference we often only reference one.

I do think if we are mentioning `PreImagesNC` we should say what the difference is -- I know someone could follow the links to PreImages and PreImagesNC, but it is a bit strange here to see both mentioned.

In general it is probably worth thinking about if we want to use `PreImagesRepresentative` or `PreImagesRepresentativeNC` in most examples. I am not sure in general how much slower `PreImagesRepresentative` is going to be, so which we should be encouraging as a default -- but I would hope `PreImagesRepresentative`, because it is safer for users who are beginners.

ThomasBreuer: @ChrisJefferson Thanks for your thoughts.

Currently I find one statement about `NC` variants in the Tutorial Manual, which talks about `Sub<something>` vs. `Sub<something>NC`, and one statement about `Sub<struct>NC` in the Reference Manual Section "Constructing Subdomains"
I think there should be a *general* statement about `Func` vs. `FuncNC` (in the Tutorial as well as in the Reference Manual), saying that the latter omits some argument checks and that the documentation of the individual functions will explain which checks are omitted.
Some example that illustrates the difference should be shown in the Tutorial: When one creates a subgroup of a group then calling `SubgroupNC` instead of `Subgroup` will avoid membership tests for the subgroup generators; in the case of a permutation group, this may mean that the computation of a stabilizer chain for the big group can be avoided, which would hopefully be not expensive; for a matrix group such as the Baby Monster, calling `SubgroupNC` will be the only reasonable way to create a subgroup, since asking for membership will be in general hopeless.

In the Reference Manual, `Func` and `FuncNC` should be documented in the same ManSection, and then cross-references need to mention only the non-`NC` variant.

Concerning the use of `NC` variants in manual examples, I see two possible strategies. Either we use only the non-`NC` variants, or we use the `NC` variants in all those cases where the context admits this (perhaps with a textual comment why the `NC` variant may be called in examples in the Tutorial and in those cases where this is not obvious).
I would expect that the differences in runtime are negligible in manual examples, and if this is not the case for some example then this example deserves a comment about this fact.

lib/alghom.gi
cdwensley: Surely all methods, such as in alghom.gi, should be for PreImagesRepresentativeNC, PreImagesSetNC and PreImagesElmNC.  The real question is whether calls to such methods should be to the NC or non-NC versions.  Since, up until now, there has been no test that element elm in in the image, it is reasonable to assume that the authors of such calls were confident that this was the case.  That is why all the calls have been changed to the NC versions.  In due course, once packages have had the chance to make changes, non-NC methods, making a check, will be added.  Once that is done the status of these calls could be reviewed.

doc/tut/group.xml
cdwensley: I agree that it is not a good idea to mention NC variants in the tutorial manual, and will make the necessary changes.  The more general comments are very interesting but are beyond the scope of this PR.

cdwensley: Removed NC references in doc/tut/group.xml and doc/tut/algvspc.xml.  Had problems trying to push these - no idea why - so used --force.  Hope that does not cause any problems.

cdwensley:
@fingolfin  Sorry to be bothering you again.  With my system more or less back to where it was, I would like to restart work on this PR.  However, the branch of gapdev I was using has been lost.  I have used "git fetch upstream refs/pull/5073/head:preimrep" to restore it; rebased on yesterday's master; and dealt with all the resulting diffs.  How to push this to 5073?  If I "git push origin preimrep" I am asked to create a new PR.  If I "git push upstream preimrep" the request is rejected (probably not a fast-forward).  Is there a simple solution, or is a new PR a good idea, with a comment here that the PR has been updated?  There are a large number of suggestions here that should not be lost. 
And while I'm asking: it's reasonable to extend from PreImagesRepresentative to PreImages, PreimagesElm and PreImagesSet, but is adding Images, etc. really a good idea?

cdwensley:
No progress on this since last February.  Restarting activity today with PR#98 for Wedderga.

cdwensley:
Progress report: 
(25/10/23)  RCWA: agreed to leave files unchanged as any tests would not be costly. 
(15/11/23)  SEMIGROUPS: PR#965 merged.

cdwensley:
In order to fix conflicts in 22 lib/*.gi and lib/grpnice.gd I have made copies from master; changed all the PreImages* to nPreImages*.NC; and then committed these edited copies. The result is that that checks fail because of changes in some of the corresponding lib/*.gd files. These can be fixed in the same way by making copies from master and, so far, grplatt.gd and gpr.gd have been processed in this way.

cdwensley:
And now fitfreee.gd.  It's a pity tests fail after finding the first undefined variable, rather than looking for more.

cdwensley:
There were so many conflicts between files edited in this PR and changes made in master that it seemed sensible to rebase the PR on the current master and make sure all the previous changes to ...NC were made. This has now been done, which will hopefully allow most of the checks to pass.

cdwensley:
PR #58 for FR merged 11/01/2024and included in release 2.4.13.

fingolfin:
@cdwensley there was a merge conflict which I took the liberty of resolving. It would be great to get this finally done. Thank you for not dropping the torch on this!

fingolfin:
Perhaps to keep track of progress, we could take the list from https://github.com/gap-system/gap/pull/5073/#issuecomment-1271306150 and copy it in slightly edited form to the issue description; I would turn it into a task list, i.e.
```
- [ ] task 1
- [x] task 2
- [ ] task 3
```
which renders as 
- [ ] task 1
- [x] task 2
- [ ] task 3

and one can "tick the boxes". Then for each package, if there is a relevant issue or PR, I'd add a link to that.

This way one can quickly see what's done already, what is in progress, and what is incomplete.

cdwensley:
The plan was to get those packages with methods installed for one of PreImages, PreImagesElm, PreImagesSet and PreImagesRepresentative to prepare for this merge. Here is my list of such packages:
cryst         merged PR#31 13/2/23 in release 4.1.25 
fga            PR#10 not yet merged (reminder for @fingolfin) 
fining        merged PR#28 26/7/23 
fr               merged PR#58 11/1/24 in release 2.4.13 
matgrp     not on GitHub, small package, just one method for PreImagesRepresentative
orb            merged PR#67 16/1/24 
polycyclic recently created PR#34
qpa           nothing to do 
rcwa          issue #26 - agreed to make no changes 
semigroups merged PR#965 14/11/23 in release 5.3.3 
utils           merged PR#52 4/10/22 in release 0.81 
walrus       nothing to do 
wedderga  PR #98 opened 12/9/23 but no response 

A number of the other distributed packages call one of more of these functions a handful of times in their library, but it seems unnecessary for them to make any adjustments at this stage. 
The exception is HAP, which has 157 calls, and has recently defined the four NC versions in its init.g in PR#121 17/1/24. 
Package sonata has 17 calls - there's nothing in between. 

If @fingolfin is able to turn this into a task list, that can only be helpful.

fingolfin:
@cdwensley turned it into a list, and the PR refs into links... please verify I set the checkboxes right (and otherwise just adjust them). Some additional remarks:
- matgrp is on GitHub, just not in the gap-packages org: https://github.com/hulpke/matgrp
- I'll try to look into the FGA and polycyclic PRs during or before GAP Days
- I don't understand the "conclusion" for rcwa: wouldn't new code that uses, say, `PreImagesNC` on a group homomorphism potentially run into an error when used with an RCWA homomorphism due there not being a `PreImagesNC` method, just one for `PreImages`? I think we'll have `PreImages` delegate to `PreImagesNC` but we can't also delegate in the other direction otherwise there'll be infinite recursion?

cdwensley:
@fingolfin: how you changed the PR references in the list above into links I cannot imagine, but they all appear to be correct. 
matgrp: there is now PR#17. 
rcwa: do not understand your comment, but have created a new PR #28 which just defines the 4 PreImages...NC in init.g. 
delegation: How does "I think we'll have PreImages delegate to PreImagesNC" not conflict with all the "if not IsBound( PreImagesNC ) then BindGlobal( "PreImagesNC", PreImages );" added to numerous init.g's?

fingolfin:
What's the state of this now, I wonder, @cdwensley ? Tests here fail

fingolfin:
(and there is a conflict now, too)

cdwensley:
@wilfwilson : Max has asked me what is the state of this PR so I have looked at it again and found your comment of 16/9/25.  Recently, when starting on something else, I mistakenly thought that this PR was going nowhere, so deleted it on my system.  So my question is - what is the GAP command to download the version you made to my system?

cdwensley:
Have successfully rebuilt my version of this branch.
It remains to see whether the checks work.
Had done nothing with this PR for some time because I mistakenly thought that there were still packages which had not made the required changes.  It does seem that these have now all been done, but I will make further checks.

cdwensley:
Lots more rebasing going on.

cdwensley:
Assuming all necessary packages have added "NC" to PreImagesRepresentative, this moves to the next stage with edits in lib/mapping.g* by adding the test to PreImagesRepresentative that the element _is_ in the image.  Have also edited out the "NC"s in tst/testinstall/*.tst.  These all appear to pass.

cdwensley:
Have now added methods for PreImages, PreImagesElm, PreImagesSet and PreImagesRepresentative, and removed the "NC" from their calls in tst/testinstall/*.tst.  @fingolfin - was this what I was supposed to do?

cdwensley:
Review of what has happened so far:
The packages which implement methods for one or more of the four operations, have merged a PR, and made a release after that are: cryst, fr, orb, polycyclic, matgrp, qpa, rcwa, semigroups, utils and wedderga.
fga has merged a PR but not made a release.
fining (incorrect info above), for reasons I forget, has closed PR#28 without merging it, and installs an other method for PreImagesSet.
Should we have included PreImagesRange in this process?  It is an attribute, not an operation.

cdwensley:
Attempted to do another "git rebase master" but gave up after fixing about 30 conflicts because there seemed to be dozens more.  Have no idea how to proceed with this PR.  Is the best plan to start a new PR, making the same changes in the main library, and then close the old one?

cdwensley:
This PR has been superceded by a second version: PR #6409.

fingolfin:
Superseded by PR #6409.

PR #5074, opened 2022-09-29
Warn users on 32-bit CPUs and big-endian CPUs

ChrisJefferson:
This PR adds a message which is displayed whenever GAP is started on a 32-bit CPU, or a big-endian CPU, pointing out that GAP don't support this CPU, and we aren't doing full testing.

There are I feel two questions here:

1) Is adding such a message reasonable? I think it is, because in general people want to trust the answers they get from GAP.

2) Is this message right? Here's what the output looks like:

```
 ┌───────┐   GAP 4.13dev-42-g7cf6fb7 built on 2022-09-29 20:35:04+0100

 **** Warning, this copy of GAP was built for: 32-bit CPUs
 **** This configuration is not officially supported or tested

```

### DIFF
--- a/lib/init.g
+++ b/lib/init.g
@@ -421,7 +421,7 @@ end );
 
 BindGlobal( "ShowKernelInformation", function()
   local sysdate, linelen, indent, btop, vert, bbot, print_info,
-        config, str, gap;
+        config, str, gap, warnings;
 
   linelen:= SizeScreen()[1] - 2;
   print_info:= function( prefix, values, suffix )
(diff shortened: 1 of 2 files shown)


pedritomelenas (approved):

fingolfin:
I am fine with this in principle, but we should not do this silently. Instead I'd suggest to bring this PR up on the GAP mailing list, so that people have a chance to react and comment. Do you want to do that, @ChrisJefferson ?

This also ties in with https://github.com/gap-system/gap/pull/5026 ...

fingolfin:
If we do this, we might as well also drop the (soon broken) 32bit package tests from CI; see also discussion in PR #5026.

PR #5082, opened 2022-10-05
Fix broken shebang

fingolfin:
Oops!

Motivated by https://src.fedoraproject.org/rpms/gap/blob/rawhide/f/gap.spec#_233 by @jamesjer

### DIFF
--- a/tst/testspecial/run_gap.sh
+++ b/tst/testspecial/run_gap.sh
@@ -1,4 +1,4 @@
-#/usr/bin/env bash
+#!/usr/bin/env bash
 
 set -e
 


ChrisJefferson (approved):
Woops indeed!

ChrisJefferson:
Out of interest, were you just having a look at the fedora packaging, or was this pointed out to us?

fingolfin:
I got this from looking at the Fedora package myself in this case, but in general @jamesjer is very good about upstreaming his patches; after we released 4.12.0 he made a ton of PRs for various packages to fix issue here and there; and also PR #5027 here.

fingolfin:
Backported to stable-4.12 via commit 03ddf4d33

PR #5083, opened 2022-10-05
Makefile.rules: use LTINSTALL to install gap executable

fingolfin:
Suggested by Bill Allombert


fingolfin:
Backported to stable-4.12 via commit 7157d0b42

PR #5085, opened 2022-10-06
buildsys: fix broken `__builtin_mul_overflow` check

fingolfin:
Without this, the check failed with at least some compilers; here is an error from Apple Clang:

    conftest.c:34:28: error: result argument to overflow builtin must be
    a pointer to a non-const integer ('int' invalid)
    __builtin_mul_overflow(0,0,0);
                               ^

And here with GCC 9.4:

    conftest.c:34:28: error: argument 3 in call to function
    '__builtin_mul_overflow' does not have pointer to integral type
       34 | __builtin_mul_overflow(0,0,0);
          |                            ^

Luckily the effects are not so bad because `intobj.h` has fallbacks, so for GCC >= 5 and clang >= 3.8 we are good anyway. Hence I don't think this needs release notes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -173,7 +173,7 @@ AC_DEFUN([CHECK_COMPILER_BUILTIN],
     AC_LINK_IFELSE(
         [AC_LANG_PROGRAM(
             [[]],
-            [$1[($2)];
+            [int x; $1[($2)];
             ]
         )],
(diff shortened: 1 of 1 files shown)


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 via commit d577b9dd5

PR #5088, opened 2022-10-06
sysinfo.gap: set GAP to PREFIX/bin/gap

fingolfin:
Suggested by Bill Allombert on the GAP support list: instead of pointing package or anything which parses `sysinfo.gap` to the "real but hidden away" GAP executable, point to `PREFIX/bin/gap` -- which for now by default is a shell script starting the "real" GAP, which can be tweaked further, and arguably is what should be used anyway


fingolfin:
Backported to stable-4.12 via commit 2fa979c4c

PR #5089, opened 2022-10-06
kernel: don't include `stdio.h` in `compiled.h`

fingolfin:
This was only there for GAP packages that erroneously relied on it. We will instead identify and fix all these packages.

For now let's keep this as a draft PR. Then I can run the package distro tests against it to see what, if anything, breaks.

### DIFF
--- a/src/compiled.h
+++ b/src/compiled.h
@@ -14,11 +14,6 @@
 #ifndef GAP_COMPILED_H
 #define GAP_COMPILED_H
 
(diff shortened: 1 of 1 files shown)


fingolfin:
Test run at https://github.com/gap-system/PackageDistro/actions/runs/3201323272

fingolfin:
All packages seem to have compiled, so this should be safe to go (in the worst case we can still revert it)

ChrisJefferson (approved):

PR #5090, opened 2022-10-06
Fix permissions for convert.pl when installing

fingolfin:


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 via commit 11d5ad179

PR #5091, opened 2022-10-06
Don't attempt to install the 'dev' manual...

fingolfin:
... as it is not contained in release tarballs!

An alternative would of course be to include the dev manual in release tarballs; but given its really bad state, I don't think that's a good idea (this may of course change in the future)

Inspired by https://github.com/NixOS/nixpkgs/pull/192548


fingolfin:
Backported to stable-4.12 via commit 459f4d78f

PR #5095, opened 2022-10-07
Tweak ConvertToMatrixRepNC behavior, add tests

### LINKED ISSUE #5081: More questions about `ConvertToMatrixRep`, `ConvertToVectorRep` and their `NC` variants
I just had to debug code involving these functions, and while I found a solution, I was left with a bunch of questions. The documentation for them is a bit sparse perhaps...

First off, the return values. The manual says this for `ConvertToVectorRep` and it is similar for matrices:
> The return value is the size of the field over which the vector ends up written, if it is written in a compressed representation

But what is supposed to happen if a conversion is not possible?
- for `ConvertToMatrixRep` and its NC variant, it seems we get `fail`:

      gap> ConvertToMatrixRep([[1,2,3]]);
      fail
      gap> ConvertToMatrixRepNC([[1,2,3]]);
      fail

- for `ConvertToVectorRep` and its NC variant, it seems we get `true` (which I find quite surprising):

      gap> ConvertToVectorRep([1,2,3]);
      true
      gap> ConvertToVectorRepNC([1,2,3]);
      true

- with `ConvertToMatrixRepNC` we also can get an error -- I kinda think this is a bug? This is the initial issue I run into; note that the non-NC variant "works". This error happens because `ConvertToVectorRepNC` "fails"

      gap> ConvertToMatrixRep([[1,2,3]]*Z(17), GF(17^2));
      fail
      gap> ConvertToMatrixRepNC([[1,2,3]]*Z(17), GF(17^2));
      Error, ConvertToMatrixRep: Failed to convert a row

My suggestion would be to return `fail` in all these cases, and document it as such.

CC @ThomasBreuer

fingolfin:
Before this, a command like

    ConvertToMatrixRepNC( [[1,0,1,1],[0,1,1,1]]*Z(2), 512 );

would result in an error. Now it returns `fail`, just like

    ConvertToMatrixRep( [[1,0,1,1],[0,1,1,1]]*Z(2), 512 );

does. The rationale here is that the request is reasonable in principle; it only fails because we do not currently offer compressed vectors over GF(512).

Contrast this with, say

    ConvertToMatrixRepNC( [[1,0,1,1],[0,1,1,1]]*Z(2), 3 );

where the inputs really do not make sense, so raising an error here is justified (and we still do).

Also add some tests for the current behavior of `ConvertToMatrixRep` and `ConvertToMatrixRepNC`; to accommodate them, I modified two `Assert` statements in `ConvertToVectorRepNC`; one was redundant and removed, the other was moved to a later point, to get a better error message in more cases (when the error is raised before the assertion check is reached).

---

Also change `ConvertToVectorRep` and its NC variant to return `fail` instead of `true` in case conversion fails. This matches `ConvertToMatrixRep` and seems more logical, too.

---

This resolves #5081 from my point of view. I don't think this needs a release notes entry, but I can improve the title if people think it should be documented.


danielrademacher (approved):

ThomasBreuer (approved):
Looks good. Thanks.

fingolfin:
I need to adjust the HPC-GAP version if the code, too

Issue #5096, opened 2022-10-07
Having `gap` use `libgap` (at least the binary installed by `make install`); and set a good `SYS_DEFAULT_PATHS` default in the installed GAP


fingolfin:
@antonio-rojas (packages GAP for Arch Linux) pointed out that kernel extension in package like `io` can not be loaded from Sage or other programs which use `libgap`, at least not without additional work. One fix is to link those kernel extensions against `libgap`, i.e., adding `-lgap` to the linker flags.

I actually was aware of this (but never logged it as an issue before; I should have...) because I have to deal with the exact same situation in [GAP.jl](https://github.com/oscar-system/GAP.jl), the Julia wrapper for GAP.

However, we cannot just add `-lgap` to the linker flags. First off, GAP itself does not use `libgap`, so it wouldn't work. Secondly, in a "classic" GAP set up, GAP is *not* installed but used from its source folder where it was built.

One possible solution for this would be to leave things as they are in the "classic" GAP builds.

But for `make install`, we already install a different `sysinfo.gap`; it would be trivial to add `-lgap` to the linker flags for use by kernel extensions. Then, we could build a second `gap` binary just for the installation; that GAP binary would link against `libgap`, and basically just contain a `main` function calling into `libgap`. (This is by the way precisely how the Windows binary works, so it's not exactly new).

That way, we'd have the best of both worlds: for "classic" builds, nothing changes. For installed builds, things start working better.

---

And if we build a separate GAP binary anyway, we could also use that to `#define SYS_DEFAULT_PATHS` sensibly

PR #5100, opened 2022-10-07
kernel: rename gap95w.c to main.c and always use it

fingolfin:
So now there is just one definition for function `main`. Also, libgap now does not contain `main` anymore (oops)

This is a small step towards realizing #5096 but I think it's useful by itself.

There is a chance I've inadvertently broken Cygwin with the changes, but I hope the CI tests will reveal that if it happened (and I am sure we can resolve it).


ChrisJefferson (approved):

fingolfin:
Backported to `stable-4.12` in 71a235d556a35b42c2ed98a12e40909c243bab7d

PR #5102, opened 2022-10-11
Basic WASM / web-building support

ChrisJefferson:
This PR provides basic support for building a WASM/web version of GAP. I mainly want to distribute it so the various flags / fixes aren't lost. While the PR isn't big, it has taken me quite a while to figure out everything that has to be done.

While the resulting GAP is (as far as I can tell) generally function, it isn't pleasant to use -- it requires downloading a 462MB file (103MB is your web server will send files compressed), and the interface requires entering into an "alert" box.

However, this does deal with correctly building, in particular dealing with issues with GASMAN.

If anyone is an expert in javascript / async / interfaces, I would be happy to help improving the interface, but this is not an area where I have expertise.

### DIFF
--- a/etc/emscripten/README.md
+++ b/etc/emscripten/README.md
@@ -1,0 +1,27 @@
+Code to allow building gap to WASM using Emscripten.
+
+Files:
+
+- `build.sh`: Run as `etc/emscripten/build.sh` from a fresh copy of GAP.
+
+Note that this built copy has one major weakness -- the 'gap.data' (which includes all packages files) is huge.
+
+
+- `web-template`: Uses 'xterm-pty' to create a "nice" interface to the Wasm GAP.
+
+See 'run-web-demo.sh' as an example on how to set up a working website.
+
+Note that this demo uses xterm-pty, a library which provides a terminal interface
+for emscripten-compiled programs. This uses a javascript feature called 
+"SharedArrayBuffer", which requires some headers are returned by the server:
+
+```
+Cross-Origin-Opener-Policy: same-origin
+Cross-Origin-Embedder-Policy: require-corp
+```
+
+For more details, see for [this article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer).
+
+The file "coi-serviceworker.js" works around this problem on Github pages. This won't
+work locally, so "server.rb" is a simple ruby script, which just starts a web-server
+which returns the required headers.
--- a/etc/emscripten/build.sh
+++ b/etc/emscripten/build.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+set -eux
+
+BASEDIR="$(pwd)"
+
+if ! command -v emmake &> /dev/null; then
+    echo Please install, and source, emscripten
+    echo This script was tested with version 3.1.23
+    echo See https://emscripten.org/docs/getting_started/downloads.html for install instructions
+    exit 1
+fi;
+
+# Build the configure script if this is a fresh git checkout
+if [[ ! -f ./configure ]]; then
+    ./autogen.sh
+fi
+
+# First build a standard GAP install, for some files
+# we will need during building
+(
+    mkdir -p native-build
+    cd native-build
+    if [[ ! -f config.status ]]; then
+        ../configure
+    fi
+    make -j8
+)
+
+AUX_BUILD=$PWD/extern/emscripten/build
+AUX_PREFIX=$PWD/extern/emscripten/install
+
+mkdir -p "$AUX_BUILD"
+mkdir -p "$AUX_PREFIX"
+
+(
+    mkdir -p "$AUX_BUILD/gmp"
+    cd "$AUX_BUILD/gmp" &&
+    if [[ ! -f config.status ]]; then
+        CC_FOR_BUILD=/usr/bin/gcc ABI=standard \
+        emconfigure $BASEDIR/extern/gmp/configure \
+        --build i686-pc-linux-gnu --host none \
+        --disable-assembly --enable-cxx \
+        --prefix=$AUX_PREFIX
+    fi &&
+    emmake make -j8 &&
+    emmake make install
+)
+
+(
+    mkdir -p "$AUX_BUILD/zlib"
+    cd "$AUX_BUILD/zlib" &&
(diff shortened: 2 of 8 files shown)


etc/emscripten-build.sh
```diff
@@ -0,0 +1,59 @@
...
+emconfigure ./configure ABI=32 --with-gmp=$(pwd)/extern/install --with-zlib=$(pwd)/extern/install LDFLAGS="-s ASYNCIFY=1 -O2"
+
+# Get basic required packages
+emmake make -j8 bootstrap-pkg-minimal
+
+# Copy in files from 'GOODGAP'
+cp ${GOODGAP}/build/c_*.c ${GOODGAP}/build/ffdata.* src/
+cp ${GOODGAP}/ffgen .
+
+# The EXEEXT is usually for windows, but here it lets us set GAP's extension, which lets us produce a html page to run GAP in
```
fingolfin, line 58:
```suggestion
# The EXEEXT is usually for windows, but here it lets us set GAP's extension,
# which lets us produce a html page to run GAP in
```

fingolfin (commented):
Cool :-)

fingolfin:
> the interface requires entering into an "alert" box.

you mean instead of a REPL or what?

fingolfin:
I wonder how hard it would be to couple it with e.g. https://xtermjs.org ... aha apparently that's exactly wha https://xterm-pty.netlify.app does...

ChrisJefferson:
I thought it didn't work with xterm.js. Turned out (I just made a new push), the problem was just to do with `read` returning length 0 if there was nothing to do.

So, here is a demo. I used xterm.js, and stripped various data files out of transgrp, primgrp and smallgrp to fit inside a reasonable data size: https://chrisjefferson.github.io/gap-web-demo/

ChrisJefferson:
> > the interface requires entering into an "alert" box.
> 
> you mean instead of a REPL or what?

Yes, the default `gap.html` you get out is functional, but uses a "javascript alert" box for input, which is just annoying. However, as discussed below, I managed to get it linked to xterm-pty, after figuring out why it wasn't working (because `read` kept returning 0 when there was no current user input, which upset GAP).

src/sysfiles.c
```diff
@@ -1352,7 +1352,18 @@ Int SyRead(Int fid, void * ptr, size_t len)
         return gzread(syBuf[fid].gzfp, ptr, len);
     }
     else {
+#ifdef EMSCRIPTEN
+        // BUG?: The EMSCRIPTEN library will return "EOF" when
```
fingolfin, line 1356: Does it really return `EOF` (so `-1`) ? From the code it looks more like it returns 0, and that would be valid, I think

fingolfin: Aha, I should have first re-read the `read(2)` man page. From https://man7.org/linux/man-pages/man2/read.2.html
> On success, the number of bytes read is returned (zero indicates end of file)

Huh

fingolfin: Is the returning of `0` here necessarily an issue in emscripten? Could it also be a problem with xterm.js, or with the code that connects the two?

ChrisJefferson: You are right, at first I thought it was returning -1, but it was infact 0.

I'm not positive what piece of code is to blame, I did submit a bug report to xterm-pty.

ChrisJefferson: fixed in xterm-pty, so removed this code

ChrisJefferson:
I cleaned up the code based on @fingolfin's suggestions. I also made a directory, and put some other helper files in there, to help build a fully functioning GAP.

The big missing thing here is an work on more cleverly choosing, or compressing, which files are used, but that isn't needed for this PR.

etc/emscripten/README.md
```diff
@@ -0,0 +1,19 @@
+Code to allow building gap to WASM using Emscripten.
+
+Files:
+
+- build.sh : Run as 'etc/emscripten/build.sh' from a fresh copy of GAP.
```
fingolfin, line 5:
```suggestion
- `build.sh`: Run as `etc/emscripten/build.sh` from a fresh copy of GAP.
```

etc/emscripten/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Note that this built copy has one major weakness -- the 'gap.data' (which includes all packages files) is huge.
+
+
+- web-template : Uses 'xterm-pty' to create a "nice" interface to the Wasm GAP.
```
fingolfin, line 10:
```suggestion
- `web-template`: Uses 'xterm-pty' to create a "nice" interface to the Wasm GAP.
```

etc/emscripten/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+See 'run-web-demo.sh' as an example on how to set up a working website.
+
+Note that this these scripts require a feature called "SharedArrayBuffer", which requires
+some webserver report.
```
fingolfin, line 15: do you mean "support"?

Could you be a bit more specific what is needed? I googled a bit and I think it might be "Cross-origin isolation" ?

etc/emscripten/build.sh
```diff
@@ -0,0 +1,93 @@
...
+    cd "$AUX_BUILD/zlib" &&
+    if [[ ! -f Makefile ]]; then
+        emconfigure $BASEDIR/extern/zlib/configure --prefix=$AUX_PREFIX
+    fi;
+    emmake make -j8 &&
+    emmake make install
+)
+
+# There are two problems with building GAP
+# 1) GAP Builds some executables (ffgen and gap-nocomp), which it wants to
```
fingolfin, line 61:
```suggestion
# 1) GAP builds some executables (ffgen and gap-nocomp), which it wants to
```

etc/emscripten/run-web-demo.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -eux
+
+echo This script assumes you have already run 'build.sh'
+
+mkdir -p web-example
+cp etc/emscripten/web-template/* web-example
+cp gap.js gap.wasm gap.data web-example
+cd web-example
+../etc/emscripten/server.rb
```
fingolfin, line 11:
```suggestion
../etc/emscripten/server.rb

```
Add trailing newline

fingolfin (approved):

PR #5103, opened 2022-10-11
Add `Exec2`, the "better version of Exec"

fingolfin:
(This is code I wrote some time ago, but I never found time to clean it up enough for inclusion. Due to a recent question in the GAP forum, I thought it might be a good idea to at least upload it here as a draft; perhaps someone else is interested in helping to finish it up? E.g. by coming up with a better name for the new function ;-)

Add a new function `Exec2` which is as easy to use as `Exec` but
much more powerful and less easy to misuse:

- don't pass everything to a shell, thus avoiding compatibility issues due to different shells on different systems (this also has one downside: you can't use redirects like ">/dev/null")
- pass arguments as separate strings, thus avoiding any issues with quoting quotes, quoting spaces, etc.
- return the exit code of the executed command so that one can determine its success or failure; the return value is actually a record to allow returning other data, such as the programs output
- make it very easy to override the input and output streams
- by default, pass no input to the program (instead of passing anything the user might type, as `Exec` does); this can be changed by adding an input stream to the argument list
- by default, capture any output of the program into a string and return it, instead of just printing the output to the console; this can be changed by adding an output stream to the argument list

### DIFF
--- a/lib/helpview.gi
+++ b/lib/helpview.gi
@@ -80,7 +80,7 @@ if ARCH_IS_WINDOWS() then
       winfilename:=MakeExternalFilename( SplitString( filename, "#" )[1] );
     fi;
     Print( "Opening help page ", winfilename, " in default windows browser ... \c" );
-    Exec( Concatenation("start ", winfilename ) );
+    Exec2( "start", winfilename );
     Print( "done! \n" );
   end
   );
@@ -138,7 +138,7 @@ elif ARCH_IS_MAC_OS_X() then
             fi;
             file := file.file;
           fi;
-          Exec(Concatenation("open -a Preview ", file));
+          Exec2("open", "-a", "Preview", file);
           Print("#  see page ", page, " in the Preview window.\n");
         end
     );
@@ -155,7 +155,7 @@ elif ARCH_IS_MAC_OS_X() then
             fi;
             file := file.file;
           fi;
-          Exec(Concatenation("open -a \"Adobe Reader\" ", file));
+          Exec2("open", "-a", "Adobe Reader", file);
           Print("#  see page ", page, " in the Adobe Reader window.\n");
         end
     );
@@ -172,7 +172,7 @@ elif ARCH_IS_MAC_OS_X() then
             fi;
             file := file.file;
           fi;
-          Exec(Concatenation("open ", file));
+          Exec2("open ", file);
           Print("#  see page ", page, " in the pdf viewer window.\n");
         end
     );
@@ -187,16 +187,15 @@ elif ARCH_IS_MAC_OS_X() then
                 fi;
                 file := file.file;
             fi;
-            Exec( Concatenation(
-                "osascript <<ENDSCRIPT\n",
-                    "tell application \"Skim\"\n",
-                    "activate\n",
-                    "open \"", file, "\"\n",
-                    "set theDoc to document of front window\n",
-                    "go theDoc to page ",String(page)," of theDoc\n",
-                    "end tell\n",
-                "ENDSCRIPT\n" ) );
-            return;
+            Exec2("osascript",
+                InputTextString(Concatenation("""
+                    tell application "Skim"
+                    activate
+                    open """, ViewString(file), """
+                    set theDoc to document of front window
+                    go theDoc to page """,String(page),""" of theDoc
+                    end tell
+                """)));
         end
     );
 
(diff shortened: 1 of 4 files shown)


lib/process.gi
```diff
@@ -261,3 +261,77 @@ InstallGlobalFunction( Exec, function( arg )
...
+    result := rec();
+
+    # parse the inputs
+    for a in arg do
+        if IsDirectory(a) then
+            if IsBound(dir) then
+                Error("must specify at most one working directory");
+            fi;
+            dir := a;
+        elif IsInputStream(a) then
```
fingolfin, line 279: Right now input and output stream can come at any position and in any order. Perhaps I should make this code stricter, and require that they come in a specific place (canonical places: 1. as the first two arguments; 2. after the command name; 3. as the last two arguments)

lib/process.gi
```diff
@@ -261,3 +261,77 @@ InstallGlobalFunction( Exec, function( arg )
...
+    fi;
+
+    # determine full executable path if it is not already a path
+    if not '/' in cmd then
+        a := Filename( DirectoriesSystemPrograms(), cmd );
+        if a = fail and ARCH_IS_WINDOWS() then
+            a := Filename( DirectoriesSystemPrograms(), Concatenation( cmd, ".exe" ) );
+        fi;
+        if a = fail then
+            Error("could not locate executable for '", cmd, "'");
```
laurentbartholdi, line 312: Perhaps, here, just fallback to  the previous`Exec`? Then you can rename `Exec2` into `Exec` since it extends the previous command

fingolfin: No, that would not be backwards compatible at all. E.g. this would change its behavior: `Exec("ls", "-l", filename, ">tmpfile");`

lib/process.gi
ChrisJefferson: While it would be a bigger change, I wonder if the input should be a record, with keys like `input :=`, `output := `.. this would mean in future it would be easier to do things like add stderr support, and make it clear which output stream is stdout and which is stderr, things like that.

fingolfin: I like that idea. It could then also be made backwards compatible: if there is a single argument which is a record, then use the new code; otherwise call the old code.

fingolfin: What I don't like about it is that for simple uses, it is much more verbose. Compare: e.g.:

old code:
```gap
Exec( Concatenation("start ", winfilename ) );
```
this PR:
```gap
Exec2( "start", winfilename );
```
with a record (record entries subject to change, of course):
```gap
Exec(rec(cmd:= "start", args := [ winfilename ] );
```
A compromise would be to only use a `rec` for the "extra" arguments like input, output, etc., like so:

```gap
Exec2( "start", winfilename, rec( input := BLAH, output := BLEH ) );
```
but then of course I can't use this to differentiate old and new callers.

fingolfin: So here's a possible way to go:
- allow input and output redirects *only* via an option record
- the option record must come last
- otherwise it stays as it is in this PR, that is: first the command name, then the list of arguments
  - arguments must be strings
  - for convenience it might be nice to also allow integers (and turn them into strings automatically)
  - anything else would be an error (for now, at least; we can add more things in the future this way; what I don't want is to `String` everything automatically, that just paints us into another corner)

Does that sound reasonable @ChrisJefferson ?

lib/helpview.gi
```diff
@@ -226,7 +228,10 @@ else # UNIX but not macOS
             fi;
             file := file.file;
           fi;
-          Exec(Concatenation("explorer.exe \"$(wslpath -a -w \"",file, "\")\""));
+
+          str := "";
+          Exec2("wslpath", "-a", "-w", url, OutputTextString(str, false));
```
wilfwilson, line 233: Here you're using the variable `url` as the `-w` argument, but the replaced code was using the variable `file`. Is this a mistake or a deliberate change?

lib/process.gi
```diff
@@ -261,3 +261,77 @@ InstallGlobalFunction( Exec, function( arg )
     Process( dir, shell, InputTextUser(), OutputTextUser(), [ cs, cmd ] );
 
 end );
+
+# TODO: document this, come up with a better name, write some tests...
```
wilfwilson, line 265: Still a big TODO 🙂 I don't currently have a better name suggestion but I'll have a think.

lib/process.gi
wilfwilson: It sounds reasonable to me at least @fingolfin.

fingolfin:
Closing this in favor of #6536.

PR #5112, opened 2022-10-14
buildsys: remove a few warnings

fingolfin:
These targets *are* complete from my point of view. Meaning that while I am sure there is room for improvements, and suggestions are welcome, right now there is nothing I am aware to be added to these.


fingolfin:
Backported to stable-4.12 in commit 1e5e64e0c

PR #5122, opened 2022-10-17
Restrict `\/` for free groups to full free groups, instead of returning wrong results for other cases

### LINKED ISSUE #4897: quotient groups problems
It seems to be allowed to form a factor group of a subgroup of a free group ...
```
gap> f:= FreeGroup( 2 );;
gap> s:= Subgroup( f, [ f.1^2 ]);;
gap> q:= s / [ s.1^5 ];
<fp group on the generators [ f1^2 ]>
```
... but this factor does not behave well.
```
gap> Size( q );
Error, no 1st choice method found for `*' on 2 arguments
[...]
```
Trying to form a quotient of a subgroup of a f.p. group immediately yields an error because `/` is defined for `IsFpGroup` but not for `IsSubgroupFpGroup`.

Should perhaps also the construction of a factor group of a subgroup of a free group be forbidden?
And if yes, what would be the criterion for the decision whether a group is a full free group?
Do we expect that the group is identical to the value of the `!.freeGroup` component of the elements family, or is it enough if the generators of the full free group are among the generators?
`IsWholeFamily` gives the right answer, but this is a property, and it can in general not be computed if it is not set a priori.

ThomasBreuer:
resolves #4897

### DIFF
--- a/grp/perf.grp
+++ b/grp/perf.grp
@@ -313,7 +313,7 @@ end);
 ##
 InstallMethod(PerfGrpConst,"perm grp",true,[IsPermGroup,IsList],0,
 function(filter,l)
-local G,i,j,g1,g2,gl,e1,e2,gens,rels,subs,pgens,deg,ind,ct,fp,extend,num;
+local G,i,j,g1,g2,gl,e1,e2,free,gens,rels,subs,pgens,deg,ind,ct,fp,extend,num;
   if Length(l)=0 then
     # special treatment for the trivial group
     G:= GroupByGenerators( [], () );
@@ -332,7 +332,8 @@ local G,i,j,g1,g2,gl,e1,e2,gens,rels,subs,pgens,deg,ind,ct,fp,extend,num;
     else
       extend:=false;
     fi;
-    gens:=GeneratorsOfGroup(FreeGroup(gl));
+    free:=FreeGroup(gl);
+    gens:= GeneratorsOfGroup(free);
     num:=Length(l[1][2]);
     rels:=CallFuncList(l[1][3],gens{[1..num]});
     subs:=rels[2];
@@ -356,7 +357,7 @@ local G,i,j,g1,g2,gl,e1,e2,gens,rels,subs,pgens,deg,ind,ct,fp,extend,num;
       od;
 
       # Tietze
-      extend:=PresentationFpGroup(Group(gens)/rels);
+      extend:=PresentationFpGroup(free/rels);
       TzOptions(extend).generatorsLimit:=Length(gens);
       TzOptions(extend).printLevel:=0;
       TzGo(extend);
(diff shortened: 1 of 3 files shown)


fingolfin:
Tests fail after `LoadAllPackages`:
```
gap> G:=PrimitiveGroup(8,3);
ASL(3, 2)
gap> MinimalNormalSubgroups(G);
Error, currently quotients of a free group are supported only if the group knows to contain all generators of its parent group at GAPROOT/lib/grpfp.gi:1618 called from
F / Rels at GAPROOT/pkg/classicpres/lib/sl2q.gi:226 called from
Modified_CR@classicpres( p ) at GAPROOT/pkg/classicpres/lib/sl2q.gi:237 called from
PresentationForSL2@classicpres( p, e ) at GAPROOT/pkg/classicpres/lib/sl.gi:301 called from
Internal_StandardPresentationForSL@classicpres( d, F ) at GAPROOT/pkg/classicpres/lib/standard.gi:120 called from
ClassicalStandardPresentation( type, d, q ) at GAPROOT/pkg/classicpres/lib/isofp.gi:39 called from
IsomorphismFpGroup( G, "F" ) at GAPROOT/lib/gpfpiso.gi:21 called from
IsomorphismFpGroup( sf ) at GAPROOT/lib/gpfpiso.gi:425 called from
IsomorphismFpGroupByChiefSeriesFactor( M, "x", N ) at GAPROOT/lib/grplatt.gi:2155 called from
NormalSubgroups( G ) at GAPROOT/lib/grp.gi:5070 called from
<function "MinimalNormalSubgroups generic search in NormalSubgroups">( <arguments> )
 called from read-eval loop at *errin*:1
```
So this seems to be due to an issue in the `classicpres` package by @hulpke and colleagues. Specifically, it does something like this in a bunch of places:
```
  F:=FreeGroup("a","b");
  F:=Group(StraightLineProgGens(GeneratorsOfGroup(F)));
```
and then processed to compute quotients of this new `F`.

I am not sure how useful the above really is (?) but it definitely breaks with this PR

ThomasBreuer:
We could allow the `F / Rels` construction if the list of generators of `F` is equal to the list of generators of the original full free group. This hack would be logically acceptable, and it would help in the situations which now fail in the tests.

hulpke:
The issue with the failing test is purely in the way the test is done. The group `F` does not have `IsWholeFamily` set, but is the whole family. The fallback method (before triggering the error) could check whether this is the case and if so set the flag and redispatch.

> I am not sure how useful the above really is (?)

The reason is to allow tracing back ways how elements were formed as product.

ThomasBreuer:
@ahulpke Well, finding out the value for `IsWholeFamily` works well only if the list of the group's generators contains all generators of the original full free group; if not then we (hopefully) get an error that the answer cannot be computed.
As I had proposed, I can modify the pull request accordingly.
(I prefer that the error message is related to `\/` not to families.)

fingolfin (approved):
Thank you!

PR #5126, opened 2022-10-18
Fix crashes in `ConvertToVectorRep` and `CopyToVectorRep`

### LINKED ISSUE #5123: problem with compressed vectors
The following happens in the GAP master branch as well as in GAP 4.11.1 and GAP 4.12.0.
(Thanks to Bernhard Böhmler for reporting this problem.)
```
gap> v:= [ 0 * Z(2) ];;
gap> ConvertToVectorRep( v, 4 );
4
gap> ConvertToVectorRep( v, 8 );
gap: src/vec8bit.c:544: RewriteVec8Bit: Assertion `!(D_FIELDINFO_8BIT(info) % D_FIELDINFO_8BIT(info1))' failed.
Abort (core dumped)
```
In our situation, the entries of the vector lie in the prime field but are represented w.r.t. the field with four elements, which is not a subfield of the desired field with eight elements.
The GAP function `CONV_VEC8BIT` calls the kernel function `ConvVec8Bit`, which calls the kernel function `RewriteVec8Bit`, and the latter function expects that the field of the result vector contains the field of the given vector.

fingolfin:
Fixes #5123

### DIFF
--- a/src/vec8bit.c
+++ b/src/vec8bit.c
@@ -528,10 +528,17 @@ static void RewriteVec8Bit(Obj vec, UInt q)
 
     if (q1 == q)
         return;
-    assert(q > q1);
+
+    if (q < q1) {
+        ErrorMayQuit("Cannot convert a vector compressed over GF(%i) to small field GF(%i)", q1, q);
+    }
+
+    if (((q - 1) % (q1 - 1)) != 0) {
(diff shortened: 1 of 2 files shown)


ThomasBreuer (approved):
Yes, this fixes the crashes.
Additionally, the check for the containment of the finite fields makes the example from issue  #5123 work, thus it is in fact not needed to adjust the GAP function `ConvertToVectorRep`.
Very nice.

PR #5127, opened 2022-10-18
Make sure owner and group in tarballs are set uniformly to 0

### LINKED ISSUE #5016: User, group and permissions in release tarballs
Currently, the release tarballs (e.g. https://github.com/gap-system/gap/releases/download/v4.12.0/gap-4.12.0.tar.gz) contain user and group information:
```
drwx------ horn/110          0 2022-08-18 16:57 gap-4.12.0/pkg/
```
If one blindly extracts the tarball, this is unexpected on a single user system and a security risks on a multi user system, I think. Additionally, the pkg directory has permissions 700, which makes the installation unusable when starting GAP as a different user.

Suggestion: change user and group to numeric ID 0, and change the permissions of the pkg directory to 755.

fingolfin:
Fixes #5016

This is not yet complete: it only ensures the owner and group are set to a sane default. But it does nothing to prevent files with 0700 or 0600 access flags. Gotta think whether and how to ensure this. Perhaps a matter of calling `umask`? or iterating over all files recurively and fixing up their flags, somehow?

CC @zickgraf

### DIFF
--- a/dev/releases/make_archives.py
+++ b/dev/releases/make_archives.py
@@ -16,13 +16,15 @@
 from utils import *
 
 import glob
+import grp
 import gzip
+import json
+import pwd
 import re
 import shutil
 import subprocess
 import sys
 import tarfile
-import json
 
 # Insist on Python >= 3.6 for f-strings and other goodies
 if sys.version_info < (3,6):
@@ -223,7 +225,9 @@ def make_and_record_archive(name, compression, root_dir, base_dir):
 
     filename = f"{name}{ext}"
     notice(f"Creating {filename}")
-    shutil.make_archive(name, compression, root_dir, base_dir)
+    owner = pwd.getpwuid(0).pw_name
+    group = grp.getgrgid(0).gr_name
+    shutil.make_archive(name, compression, root_dir, base_dir, owner = owner, group = group)
     manifest_list.append(filename)
 
 # Create the remaining archives
--- a/dev/releases/make_archives.py
+++ b/dev/releases/make_archives.py
@@ -156,6 +156,12 @@ with working_directory(tmpdir + "/" + basename):
     notice("Extracting package tarballs")
     with tarfile.open(tmpdir+"/"+all_packages_tarball) as tar:
         tar.extractall(path="pkg")
+    # for some reason pkg sometimes ends up with permission 0700 so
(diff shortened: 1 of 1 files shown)


zickgraf:
> This is not yet complete: it only ensures the owner and group are set to a sane default. But it does nothing to prevent files with 0700 or 0600 access flags. Gotta think whether and how to ensure this. Perhaps a matter of calling `umask`? or iterating over all files recurively and fixing up their flags, somehow?

I think most files should have correct permissions due to coming form a git checkout, a build, or extraction of another tarball. In any case, I think I only had to change the permissions of `pkg`. So maybe a `chmod 755` after creation of `pkg` here https://github.com/gap-system/gap/blob/2ecb09e0eb92c9e6bc7dbb2ab9fd79e63e931e57/dev/releases/make_archives.py#L156 would suffice?

fingolfin:
I downloaded the latest nightly build (from [here](https://github.com/gap-system/gap/actions/runs/3278772708)) to analyze permissions. It seems multiple generated files have unexpected (?) permissions, e.g:
```
-rw-r--r--  0 runner docker   2197 Oct 19 06:22 gap-4.13dev-81-g114824d/CITATION
-rw-rw-r--  0 runner docker   2220 Oct 18 17:07 gap-4.13dev-81-g114824d/CITATION.in
```
Though arguably in this case the generated file is right (it has 0644 permissions), and all the files we get from git are "wrong", with their 0664 permissions... *sigh*

One idea would be to iterate over all the files and directories, and do this:
- all directories get 0755
- anything for which the user executable bit is set also gets 0755
- everything else is set to 0644

That said, I've now looked through a ton of tar.gz files from popular open source projects, and as far as I can see, almost none of them gets any of this right either:
- semi-random owner and group names
- in fact multiple different owner and/or group names appearing
- random files with e.g. 444 or 664 or 775 permission (read-only for everyone)


I guess if projects like GNU readline, curl or Boost (amongst others) can't get this right, we don't need to feel too bad about it either ;-)

fingolfin:
I guess we should make sure that the `umask` is set to `022`.

ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 in commits 33d96073b and 2baa3b3b3

PR #5128, opened 2022-10-19
etc/Makefile.gappkg: fix 'distclean' target

fingolfin:


fingolfin:
Backported to stable-4.12 in commit c5b1874ab

PR #5129, opened 2022-10-19
CI: test whether `make install` works with release tarballs

### LINKED ISSUE #5094: CI: add a `make install` test that uses the release tarball
There was a stupid bug that totally broke `make install` in release tarballs, see #5091 for a fix and details.

This should have been caught before. But while we do test `make install`, we only test it from a git snapshot. But we really should also test it for the "release" tarballs we create as part of CI.

Basically, after creating those tarballs, we should run the `testmakeinstall` test suite of `dev/ci.sh` (this may of course involve tweaking it a bit).

fingolfin:
Resolves #5094

Working on it revealed an actual bug in `make install` for macOS (a regression compared to GAP 4.12.0). yay!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,6 +85,23 @@ jobs:
       - name: "Make archives"
         run: python -u ./dev/releases/make_archives.py
 
+      - name: "Test building GAP from the primary release archive"
+        run: |
+          mkdir -p test-gap-tarball
+          pushd test-gap-tarball
+          tar xvf ../tmp/gap-${{ steps.get-build.outputs.name }}.tar.gz
+          cd gap*
+          # test building GAP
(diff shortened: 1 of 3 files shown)


ruthhoffmann:
Looks good to me. Do you need an independent (and idiot-proof) `make install` check?

fingolfin:
@ruthhoffmann a bunch of downstream packagers for various distros have been testing it, but of course they are a "special crowd". 

So if you are interested to try it, why not, could be helpful! As a starting point, search for `make install` in `INSTALL.md` (hint: you'll want to specify a custom prefix dir, as described there, e.g. I use `/tmp/gapprefix` in this PR)

ruthhoffmann (approved):
Skipping the dummy test of the install as I believe it is stable without it.

PR #5130, opened 2022-10-19
Matrix object type: upper triangular matrices

danielrademacher:
This adds a new type of matrix object for upper triangular matrices.

## Text for release notes

see title

### DIFF
--- a/lib/matobjtriangularmatrices.gd
+++ b/lib/matobjtriangularmatrices.gd
@@ -67,6 +67,6 @@
  # Position in the positional object of the base domain
  BindGlobal( "UPPERTRIANGULARMATREP_BDPOS", 1 );
  # Position in the positional object of the number of rows
- BindGlobal( "UPPERTRIANGULARMATTREP_NRPOS", 2 );
+ BindGlobal( "UPPERTRIANGULARMATREP_NRPOS", 2 );
  # Position in the positional object of the list of entries
  BindGlobal( "UPPERTRIANGULARMATREP_ELSPOS", 3 );
--- a/lib/matobjtriangularmatrices.gd
+++ b/lib/matobjtriangularmatrices.gd
@@ -56,7 +56,7 @@
      and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
      [] );
 
- # If we implement our object a a positional object we often have to access its
+ # If we implement our object as a positional object we often have to access its
  # properties in the code. To make that more readable we declare global
  # variables. If you do this too make sure you use variables that are unique and
  # unlikely to be used someplace else, even though that might mean using longer
@@ -65,8 +65,8 @@
 
  # Some constants for matrix access:
  # Position in the positional object of the base domain
- BindGlobal( "UPPERTRIANGULARMATREP_BDPOS", 1 );
- # Position in the positional object of the number of rows
- BindGlobal( "UPPERTRIANGULARMATREP_NRPOS", 2 );
+ BindConstant( "UPPERTRIANGULARMATREP_BDPOS", 1 );
+ # Position in the positional object of the number of rows and columns
+ BindConstant( "UPPERTRIANGULARMATREP_NRPOS", 2 );
  # Position in the positional object of the list of entries
- BindGlobal( "UPPERTRIANGULARMATREP_ELSPOS", 3 );
+ BindConstant( "UPPERTRIANGULARMATREP_ELSPOS", 3 );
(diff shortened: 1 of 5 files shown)


danielrademacher:
Documentation, tests and arithmetical operations are WIP.

lib/matobjtriangularmatrices.gd
```diff
@@ -0,0 +1,72 @@
...
+ ##  <Item>
+ ##    a plain list (see <Ref Filt="IsPlistRep"/> of its entries.
+ ##  </Item>
+ ##  </Enum>
+ ##  </Description>
+ ##  </ManSection>
+ ##  <#/GAPDoc>
+ ##
+ # Here we declare the new representation and tell GAP which properties it
+ # implies. FListMatrices e.g. are positional objects and so on.
```
fingolfin, line 51:
```suggestion
 # implies. FListMatrices e.g. are positional objects and so on.
```
FListMatrices?

lib/matobjtriangularmatrices.gi
```diff
@@ -0,0 +1,599 @@
...
+   function( mat, row )
+   # could this cause problems if some entries in the vector are not bound?
+     local index_start, index_end, vec, i;
+     index_start := (row-1)*mat![UPPERTRIANGULARMATREP_NRPOS] + 1;
+     index_end := index_start + mat![UPPERTRIANGULARMATREP_NRPOS];
+     vec := NewZeroVector(IsPlistVectorRep,mat![UPPERTRIANGULARMATREP_BDPOS],mat![UPPERTRIANGULARMATREP_NRPOS]);
+     for i in [row..mat![UPPERTRIANGULARMATREP_NRPOS]] do 
+        vec[i] := mat![UPPERTRIANGULARMATREP_ELSPOS][(-row*row+row)/2+mat![UPPERTRIANGULARMATREP_NRPOS]*(row-1) + i];
+     od;
+     return NewVector(IsPlistVectorRep,mat![UPPERTRIANGULARMATREP_BDPOS],List(vec));
```
fingolfin, line 171: `mat[i][j] := 5;` is not the same as `mat[i,j]:=5` now -- the former does not modify `mat`, the latter does... but historic code expects both to be the same...

My suggestion: return an *immutable* vector here, then the user at least gets an error.

(More generally: maybe we should not provide this at all, to prevent users from writing in-efficient code, *or* a generic function for this should be provided... dunnno)

fingolfin, line 171: why not just
```suggestion
     return vec;
```

lib/matobjtriangularmatrices.gi
```diff
@@ -0,0 +1,599 @@
...
+ #T we can use 'InstallMethod'.
+   [ IsUpperTriangularMatrixRep, IsPosInt ],
+   function( mat, row )
+   # could this cause problems if some entries in the vector are not bound?
+     local index_start, index_end, vec, i;
+     index_start := (row-1)*mat![UPPERTRIANGULARMATREP_NRPOS] + 1;
+     index_end := index_start + mat![UPPERTRIANGULARMATREP_NRPOS];
+     vec := NewZeroVector(IsPlistVectorRep,mat![UPPERTRIANGULARMATREP_BDPOS],mat![UPPERTRIANGULARMATREP_NRPOS]);
+     for i in [row..mat![UPPERTRIANGULARMATREP_NRPOS]] do 
+        vec[i] := mat![UPPERTRIANGULARMATREP_ELSPOS][(-row*row+row)/2+mat![UPPERTRIANGULARMATREP_NRPOS]*(row-1) + i];
```
fingolfin, line 169:
```suggestion
     for i in [row..NrCols(mat)] do 
        vec[i] := mat[row,i];
```

lib/matobjtriangularmatrices.gi
```diff
@@ -0,0 +1,599 @@
...
+        return Zero(mat![UPPERTRIANGULARMATREP_BDPOS]);
+    else
+        return mat![UPPERTRIANGULARMATREP_ELSPOS][(-row*row+row)/2+mat![UPPERTRIANGULARMATREP_NRPOS]*(row-1) + col];
+    fi;
+ end );
+
+ InstallMethod( SetMatElm, "for an IsUpperTriangularMatrixRep matrix, two positions, and an object",
+   [ IsUpperTriangularMatrixRep and IsMutable, IsPosInt, IsPosInt, IsObject ],
+   function( mat, row, col, obj )
+    if (col < row) then
```
fingolfin, line 331:
```suggestion
    if col < row then
```

PR #5132, opened 2022-10-19
Unifying the references for matrix and vector

mihalova:
In the documentation, there were two reference names for the matrix: "mat" and "M". I unified it to "M" in every function, which was used from the beginning of the documentation. I did the same also for the vector which was referred to as "v" and "vec". I unified it to "v".

### DIFF
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -490,23 +490,23 @@ DeclareOperation( "AddVector",
 
 #############################################################################
 ##
-#O  MultVector( <vec>, <mul>[, <from>, <to>] )
-#O  MultVectorLeft( <vec>, <mul>[, <from>, <to>] )
-#O  MultVectorRight( <vec>, <mul>[, <from>, <to>] )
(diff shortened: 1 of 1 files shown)


lib/matobj2.gd
```diff
@@ -1433,17 +1433,17 @@ DeclareOperation( "IsBound[]", [ IsRowListMatrix, IsPosInt ] );
...
+#O  Unbind\[\]( <M>, <pos> )
 ##
 ##  <#GAPDoc Label="RowListMatObj_Unbind">
 ##  <ManSection>
-##  <Oper Name="Unbind\[\]" Arg='mat, pos' Label="for a row list matrix"/>
+##  <Oper Name="Unbind\[\]" Arg='M, pos' Label="for a row list matrix"/>
 ##
 ##  <Returns>nothing</Returns>
-##
+##a
```
ChrisJefferson, line 1443: This extra 'a' is confusing GAPDoc, once this is cleaned up I think everything will work!

ChrisJefferson, line 1443:
```suggestion
##
```

PR #5138, opened 2022-10-19
Add generic Unpack for MatrixObjects

wucas:
Quite a few generic MatObj methods use `Unpack` which, given a matrix object `M` say, returns a classical GAP matrix with the same entries as `M`. This method has so far been implemented by the various MatrixObject implementations. However, we do not require that a MatrixObject implements `Unpack`. In order for as many as possible generic methods to work for a MatrixObject that adheres to the specification this PR implements a generic `Unpack`.

This method should be tested of course. Since the existing MatrixObjects implement `Unpack` calling `Unpack` with any existing MatrixObject would not actually call the generic method. A minimal MatrixObject is in the works as part of a new test suite for MatrixObjects and as soon as that is merged I will add tests for this method. We should wait with merging until then.


lib/matobj.gi
```diff
@@ -84,6 +84,13 @@ InstallMethod( \=,
                                 j -> M1[i,j] = M2[i,j] ) );
     end );
 
+InstallMethod( Unpack,
+    "generic method for matrix objects",
+    [ IsMatrixObj ],
+    function(M)
+      local res;
```
fingolfin, line 91:
```suggestion
```

ChrisJefferson (approved):

PR #5139, opened 2022-10-19
Enhance `Display` for functions to print where the function was defined, if possible

fingolfin:
<s>This is not yet done: while I was updating the tests, I discovered a crash 😂 </s>


fingolfin:
OK the crash seems to have been spurious (it went away after `make`, seems I forgot to do that after the last pull).

ChrisJefferson (approved):

ruthhoffmann (approved):

PR #5140, opened 2022-10-19
Extend StreamFormatting to seperate linewrap and indenting

ChrisJefferson:
This is another attempt at an idea previous discussed in #4496 and #4511

This allows passing a record to SetPrintFormattingStatus with separate `linewrap` and `indent` options. `PrintFormattingStatus` now always returns a record.

This code needs obvious documentation (help welcome!) but I thought I'd see if it merges, and the direction seems basically sensible.

After this we may want to make further changes, for example disabling `linewrap` by default for things like `OutputTextFile`, but that can be a separate PR.


fingolfin:
I think you need to run `TestDirectory("tst/testinstall", rec(rewriteToFile:=true));`

lib/streams.gi
```diff
@@ -1676,14 +1668,29 @@ InstallMethod( SetPrintFormattingStatus, "for non-text output stream",
...
+        if IsBool(fs) then
+            if fs = fail then
+                Error("Formatting status cannot be 'fail'");
+            fi;
+        elif IsRecord(fs) then
+            if Set(RecNames(fs)) <> ["indent", "linewrap"] then
+                Error("Formatting status records must contain only 'indent' and 'linewrap'");
+            fi;
+        else
+            Error("Formatting status must be a Boolean or a Record");
```
fingolfin, line 1690:
```suggestion
            Error("Formatting status must be a boolean or a record");
```

src/io.c
```diff
@@ -849,7 +851,7 @@ UInt OpenOutput(TypOutputFile * output, const Char * filename, BOOL append)
     else if (streq(filename, "*errout*"))
         output->format = IO()->PrintFormattingForErrout;
     else
-        output->format = TRUE;
+        output->format = MakeStreamFormat(1, 1);
```
fingolfin, line 854: How about
```suggestion
        output->format = MakeStreamFormat(TRUE, TRUE);
```

src/io.h
```diff
@@ -92,6 +92,26 @@ enum {
     MAXLENOUTPUTLINE = 4096,
 };
 
+typedef struct {
+    unsigned linewrap : 1;
+    unsigned indent : 1;
+} StreamFormat;
+
+EXPORT_INLINE StreamFormat MakeStreamFormat(unsigned linewrap,
+                                            unsigned indent)
```
fingolfin, line 101:
```suggestion
EXPORT_INLINE StreamFormat MakeStreamFormat(BOOL linewrap,
                                            BOOL indent)
```

src/io.c
```diff
@@ -2029,8 +2069,8 @@ static Int InitKernel (
     IO()->Output = 0;
     IO()->InputLog = 0;
     IO()->OutputLog = 0;
-    IO()->PrintFormattingForStdout = TRUE;
-    IO()->PrintFormattingForErrout = TRUE;
+    IO()->PrintFormattingForStdout = MakeStreamFormat(1, 1);
+    IO()->PrintFormattingForErrout = MakeStreamFormat(1, 1);
```
fingolfin, line 2073:
```suggestion
    IO()->PrintFormattingForStdout = MakeStreamFormat(TRUE, TRUE);
    IO()->PrintFormattingForErrout = MakeStreamFormat(TRUE, TRUE);
```

src/io.c
```diff
@@ -1881,6 +1885,42 @@ void SPrTo(Char *buffer, UInt maxlen, const Char *format, Int arg1, Int arg2)
...
+
+    BOOL i = (indent == True);
+    BOOL l = (linewrap == True);
+
+    return MakeStreamFormat(l, i);
+}
+
+Obj ObjFromStreamFormat(StreamFormat sf)
+{
+    Obj o = NEW_PREC(0);
```
fingolfin, line 1915:
```suggestion
    Obj o = NEW_PREC(2);
```

lib/streams.gi
```diff
@@ -1676,14 +1668,35 @@ InstallMethod( SetPrintFormattingStatus, "for non-text output stream",
...
+            fi;
+            for r in ["indent","linewrap"] do
+                if not fs.(r) in [false, true] then
+                    Error(Concatenation(r, " must be a Boolean in Formatting status"));
+                fi;
+            od;
+        else
+            Error("Formatting status must be a boolean or a record");
+        fi;
+end);
```
fingolfin, line 1698:
```suggestion
function(fs)
    local r;
    if IsBool(fs) then
        if fs = fail then
            Error("Formatting status cannot be 'fail'");
        fi;
    elif IsRecord(fs) then
        if Set(RecNames(fs)) <> ["indent", "linewrap"] then
            Error("Formatting status records must contain only 'indent' and 'linewrap'");
        fi;
        for r in ["indent","linewrap"] do
            if not fs.(r) in [false, true] then
                Error(Concatenation(r, " must be a Boolean in Formatting status"));
            fi;
        od;
    else
        Error("Formatting status must be a boolean or a record");
    fi;
end);
```

ChrisJefferson:
I've rebased this PR, after a bunch of changes / reformattings. It still lacks documentation, which I'm happy to write, but would like to know if anyone is fundamentally opposed to the idea first.

lib/streams.gi
```diff
@@ -1676,14 +1668,35 @@ InstallMethod( SetPrintFormattingStatus, "for non-text output stream",
         TryNextMethod();
     fi;
 
-    if stat = true then
-        Error("non-text streams support onlyPrint formatting status false");
-    elif stat = fail then
-        Error("Print formatting status must be true or false");
+    CheckValidPrintFormattingStatus(stat);
+    if (( IsBool(stat) and stat = true ) or
```
fingolfin, line 1672: I think you can just write this:
```suggestion
    if stat = true or
```

lib/streams.gi
```diff
@@ -1676,14 +1668,35 @@ InstallMethod( SetPrintFormattingStatus, "for non-text output stream",
...
+        if fs = fail then
+            Error("Formatting status cannot be 'fail'");
+        fi;
+    elif IsRecord(fs) then
+        if Set(RecNames(fs)) <> ["indent", "linewrap"] then
+            Error("Formatting status records must contain only 'indent' and 'linewrap'");
+        fi;
+        for r in ["indent","linewrap"] do
+            if not fs.(r) in [false, true] then
+                Error(Concatenation(r, " must be a Boolean in Formatting status"));
```
fingolfin, line 1692:
```suggestion
                Error(r, " must be 'true' or 'false in formatting status record");
```

lib/streams.gi
```diff
@@ -1676,14 +1668,35 @@ InstallMethod( SetPrintFormattingStatus, "for non-text output stream",
...
+InstallGlobalFunction( "CheckValidPrintFormattingStatus",
+function(fs)
+    local r;
+    if IsBool(fs) then
+        if fs = fail then
+            Error("Formatting status cannot be 'fail'");
+        fi;
+    elif IsRecord(fs) then
+        if Set(RecNames(fs)) <> ["indent", "linewrap"] then
+            Error("Formatting status records must contain only 'indent' and 'linewrap'");
```
fingolfin, line 1688:
```suggestion
            Error("Formatting status records must contain exactly two components named 'indent' and 'linewrap'");
```

src/io.c
```diff
@@ -1318,55 +1320,57 @@ static void PutChrTo(TypOutputFile * stream, Char ch)
...
-        // print line up to the best split position
-        stream->line[ spos++ ] = '\n';
-        stream->line[ spos   ] = '\0';
-        PutLineTo( stream, spos );
-        spos--;
+      else if (stream->format.linewrap && spos != 0) {
+
+          // add character to the line, terminate it
+          stream->line[stream->pos++] = ch;
+          stream->line[stream->pos++] = '\0';
```
fingolfin, line 1332: Let's maybe not mix reformatting with code changes?

src/io.h
```diff
@@ -92,6 +92,25 @@ enum {
...
 };
 
+typedef struct {
+    BOOL linewrap;
+    BOOL indent;
+} StreamFormat;
+
+EXPORT_INLINE StreamFormat MakeStreamFormat(BOOL linewrap, BOOL indent)
+{
+    StreamFormat s = { .linewrap = linewrap, .indent = indent };
```
fingolfin, line 102: This won't work in C++, I think?

fingolfin (commented):
I (still) like this. Of course docs are needed. Some minor comments.

One more ping to @frankluebeck and @ThomasBreuer in case they would like to comment

ChrisJefferson:
I've cleaned up this PR so it applies to master, and added some basic documentation. I removed (most) of the whitespace-only changes, which should make the PR easier to read, hopefully.

This is quite a significant change to a fairly fundamental part of GAP (how indenting and line wrapping works), so there might be some fallout I haven't noticed.

zickgraf:
I'm not sure if the following is the expected behaviour:
```
gap> PrintFormattingStatus(OutputTextString("",false));
true
```
I would expect the return value to be a record.

fingolfin:
Started a package distro CI run against this PR, logs at https://github.com/gap-system/PackageDistro/actions/runs/7589478131

james-d-mitchell:
Semigroups package tests fail but I'm happy to make a release and update things to accommodate this Pr

ChrisJefferson:
Some of that output looks really strange, with very strange wrapping -- should probably try to figure out what's going on, as the only change I would expect from this would be if you actually tried to print out the value of `PrintFormattingStatus`, and saw it's now a record instead of a boolean.

wilfwilson:
It only seems to be the Semigroups package that is affected, as @james-d-mitchell reported. This seems to be to do with the viewing of "nested" objects, e.g.:
```
######## > Diff in:
/home/runner/gap/pkg/semigroups-5.5.4/tst/standard/congruences/congwordgraph.t\
st:322
# Input is:
C2 := NextIterator(it);
# Expected output:
<left congruence by word graph over <monoid of size 11, 3x3 boolean ma\
trices 
 with 3 generators>>
# But found:
<left congruence by word graph over <monoid of size 11, 3x3 boolean ma\
trices w\
ith 3 generators>>
########
```
and
```
######## > Diff in:
/home/runner/gap/pkg/semigroups-5.5.4/tst/standard/congruences/conglatt.tst:
350
# Input is:
poset := LatticeOfCongruences(S);
# Expected output:
<lattice of 4 two-sided congruences over 
 <symmetric inverse monoid of degree 2>>
# But found:
<lattice of 4 two-sided congruences over <symmetric inverse monoid of \
degree 2\
>>
########
```
and
```
######## > Diff in:
/home/runner/gap/pkg/semigroups-5.5.4/tst/standard/semigroups/semifp.tst:2587
# Input is:
map := AntiIsomorphismDualFpSemigroup(S);
# Expected output:
MappingByFunction( <fp semigroup with 2 generators and 
  3 relations of length 14>, <fp semigroup with 2 generators and 
  3 relations of length 14>, function( x ) ... end, function( x ) ... end )[0\
m
# But found:
MappingByFunction( <fp semigroup with 2 generators and 3 relations of \
length 1\
4>, <fp semigroup with 2 generators and 3 relations of length 14>, function( x\
\
 ) ... end, function( x ) ... end )
########
```
I'd say the new output is worse in this case/doesn't look great.

PR #5147, opened 2022-10-20
CI: replace obsolete set-output

### LINKED ISSUE #5107: CI: replace `set-output` by new system (by appending to `$GITHUB_OUTPUT`)
GitHub is deprecating `set-output` (which we e.g. use to generate job matrices) for security reasons, see [this blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). For new it will keep working with no timeline set to when it might be removed; but we could pro-actively try to move to the new system, [described here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter).

fingolfin:
Resolves #5107

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -296,7 +296,7 @@ jobs:
     steps:
       - name: Get branch name
         id: get-branch
-        run: echo ::set-output name=branch::${GITHUB_REF#refs/*/}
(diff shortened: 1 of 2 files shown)


fingolfin:
Backported to stable-4.12 in commit 32c27faff

PR #5150, opened 2022-10-20
call `FreeGeneratorsOfWholeGroup` not `FreeGeneratorsOfFpGroup`

### LINKED ISSUE #5145: Failure in `classicpres` testsuite cause by `\/` method changes for fp groups
Unfortunately PR #5122 broke the `classicpres` testsuite. To wit:
```
$ ./gap pkg/classicpres/tst/testall.g
 ┌───────┐   GAP 4.13dev-82-gf4d370a-dirty built on 2022-10-19 18:43:38+0200
 └───────┘   Architecture: aarch64-apple-darwin21-default64-kv8
 Packages:   AClib 1.3.2, Alnuth 3.2.1, AtlasRep 2.1.5, AutoDoc 2022.07.10, AutPGrp 1.11,
             Browse 1.8.18, CaratInterface 2.3.4, Cryst 4.1.25, CrystCat 1.1.10,
             CTblLib 1.3.4, curlInterface 2.3.1, FactInt 1.6.3, FGA 1.4.0, Forms 1.2.8,
             GAPDoc 1.6.6, genss 1.6.8, IO 4.7.3, IRREDSOL 1.4.3, LAGUNA 3.9.5,
             orb 4.9.0, Polenta 1.3.10, Polycyclic 2.16, PrimGrp 3.4.2, RadiRoot 2.9,
             recog 1.4.2, ResClasses 4.7.3, SmallGrp 1.5, Sophus 1.27, TomLib 1.2.9,
             TransGrp 3.6.3, utils 0.77
Classical Group Presentations, v.1.22
Architecture: aarch64-apple-darwin21-default64-kv8

testing: /Users/mhorn/Projekte/GAP/gap/pkg/classicpres/tst/calltest.tst
########> Diff in /Users/mhorn/Projekte/GAP/gap/pkg/classicpres/tst/calltest.tst:
4
# Input is:
LittleTest();
# Expected output:
# But found:
Error, no 1st choice method found for `FreeGeneratorsOfFpGroup' on 1 arguments
########
      24 ms (23 ms GC) and 245KB allocated for calltest.tst
-----------------------------------
total        24 ms (23 ms GC) and 245KB allocated
              1 failures in 1 of 1 files

#I  Errors detected while testing
```

ThomasBreuer:
This is intended to resolve #5145.
There are groups in `IsSubgroupFpGroup and IsWholeFamily` for which `FreeGeneratorsOfFpGroup` has no method.
(Should such a method perhaps be added?)
The test suite of the `classicpres` package contains examples which run into errors inside the GAP library function DoGGMBINC;
the problems disappear when one calls `FreeGeneratorsOfWholeGroup` instead of `FreeGeneratorsOfFpGroup`.

### DIFF
--- a/lib/ghom.gi
+++ b/lib/ghom.gi
@@ -345,12 +345,12 @@ local   filter,  hom,pcgs,imgso,mapi,l,obj_args,p;
   if IsSubgroupFpGroup(G) then
     if HasIsWholeFamily(G) and IsWholeFamily(G) 
       # total on free generators
-      and Set(FreeGeneratorsOfFpGroup(G))=Set(gens,UnderlyingElement)
+      and Set(FreeGeneratorsOfWholeGroup(G))=Set(gens,UnderlyingElement)
       then
         l:=List(gens,UnderlyingElement);
-        p:=List(l,i->Position(FreeGeneratorsOfFpGroup(G),i));
+        p:=List(l,i->Position(FreeGeneratorsOfWholeGroup(G),i));
         # test for duplicate generators, same images
-        if Length(gens)=Length(FreeGeneratorsOfFpGroup(G)) or
+        if Length(gens)=Length(FreeGeneratorsOfWholeGroup(G)) or
           ForAll([1..Length(gens)],x->imgs[x]=imgs[Position(l,l[x])]) then
           filter := filter and IsFromFpGroupStdGensGeneralMappingByImages;
           hom.genpositions:=p;
--- a/tst/testinstall/grpfree.tst
+++ b/tst/testinstall/grpfree.tst
@@ -266,4 +266,15 @@ gap> Group(F.1 ^ 100);
 Group(<1 generator>)
 
 #
+gap> F:= FreeGroup( 2 );;
+gap> F:= Group( GeneratorsOfGroup( F ) );;
+gap> gens:= GeneratorsOfGroup( F );;
+gap> HasIsWholeFamily( F );
+false
+gap> H:= SymmetricGroup( 5 );;
+gap> GroupHomomorphismByImagesNC( F, H, gens, GeneratorsOfGroup( H ) );;
+gap> SetIsWholeFamily( F, true );
+gap> GroupHomomorphismByImagesNC( F, H, gens, GeneratorsOfGroup( H ) );;
+
+#
 gap> STOP_TEST( "grpfree.tst", 1);


hulpke (approved):
The change is fine, of course. I wonder however about the distinction with `OfWholeGroup`. What is `FreeGeneratorsOfFpGroup` then supposed to do. Do we need such a function?
Further, there are brother/sister functions for monoids and semigroups, as well as `RelatorsOfFpGroup` that now will be named inconsistently. I think the easier fix is to have `classicpres` add an attribute assignment.

fingolfin:
Some more thoughts: The manual says this:

>   FreeGeneratorsOfFpGroup  returns  the  underlying  free generators corresponding to the
>  generators of the finitely presented group G which must be a full FpGroup.
>
>  FreeGeneratorsOfWholeGroup  also works for subgroups of an FpGroup and returns the free
>  generators of the full group that defines the family.

(What is a "full group" ? Is it the same as assuming that `IsWholeFamily` returns `true` for the group?)

Maybe the problem then is that there should be a method for `FreeGeneratorsOfFpGroup` which is triggered at least for `IsWholeFamily` fp groups?!? 

But then there is also `IsGroupOfFamily` which seems to be a strong condition than `IsWholeFamily`. While it is not officially documented, there is still an (unused) GAPDoc comment for it which says:
```
##  This filter indicates that the group <A>G</A> is the group
##  which is stored in the family <A>fam</A> of its elements
##  as <C><A>fam</A>!.wholeGroup</C>.
```

ThomasBreuer:
Concerning the relation between `IsGroupOfFamily` and `IsWholeFamily`, I think that at least an implication from `IsGroupOfFamily` to `IsWholeFamily` would be appropriate.

It would be tempting to get rid of `IsGroupOfFamily`, by replacing `IsGroupOfFamily` by `IsWholeFamily`. In most of the cases this is in fact no problem.
Currently one obstacle is the `FreeGeneratorsOfFpGroup` method with requirements `IsSubgroupFpGroup and IsGroupOfFamily and IsFreeGroup`: Here `IsWholeFamily` is not enough because the method just returns the `GeneratorsOfGroup` value of the argument; if we would change the code to take the generators of the full group, we could get rid of `IsGroupOfFamily` in this method installation.

fingolfin:
or we could store "the" generators of the family in the family when creating the family, and then return those?

ThomasBreuer:
The family is the right place from where one fetches the information. Currently one can access `GeneratorsOfGroup( Fam!.freeGroup )`.
Storing "the free generators" in another component would be possible, but I would in fact prefer to avoid relying on `.!` component access.
(One argument against another component would be that a free group has also "the free generators" as a monoid or semigroup, which can be easily produced from the group on demand.)

fingolfin:
OK, also fine by me.

I will merge this for now, as it fixes the immediate issue. If someone wants to follow up, I suggest to open an issue with the details as a reminder.

Issue #5142, opened 2022-10-20
MatrixObj: add "minimal" MatrixObj and VectorObj implementations


fingolfin:
... which implement *exactly* the required operations and nothing more. The idea being that we can then use these together with the MatrixObj / VectorObj test suites to test all our generic implementations.

To prevent users from accidentally using them we maybe shouldn't make them part of the library but instead load them from the test suite only?

(@wucas wanted to work on this, and perhaps already did, but I am filing an issue so this isn't forgotten)

PR #5152, opened 2022-10-20
Add a minimal matrix object implementation

wucas:
This PR adds a matrix object `IsMinimalMatrixRep`. Internally it represents the matrix as a flat list in row-major form. It only the required methods. 

This is intended to be used as a minimal example and in order to test code, in particular the generic methods.

Regarding tests, if nothing obvious is wrong I'd like to merge this asap and then use it with the new test suite. Any errors could be fixed then.


lib/matobjminimal.gd
```diff
@@ -0,0 +1,77 @@
...
+# Here we declare the new representation and tell GAP which properties it
+# implies. FListMatrices e.g. are positional objects and so on.
+DeclareRepresentation( "IsMinimalMatrixRep",
+        IsMatrixObj and IsMatrixOrMatrixObj and IsPositionalObjectRep
+    and IsNoImmediateMethodsObject
+    and HasNumberRows and HasNumberColumns
+    and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
+    [] );
+
+# If we implement our object a a positional object we often have to access its
```
ChrisJefferson, line 62:
```suggestion
# If we implement our object as a positional object we often have to access its
```

lib/matobjminimal.gd
```diff
@@ -0,0 +1,77 @@
...
+
+# Some constants for matrix access:
+# Position in the positional object of the base domain
+BindGlobal( "MINREP_BDPOS", 1 );
+# Position in the positional object of the number of rows
+BindGlobal( "MINREP_NRPOS", 2 );
+# Position in the positional object of the number of columns
+BindGlobal( "MINREP_NCPOS", 3 );
+# Position in the positional object of the list of entries
+BindGlobal( "MINREP_ELSPOS", 4 );
```
ChrisJefferson, line 77:
```suggestion
BindGlobal( "MINREP_ELSPOS", 4 );

```

ChrisJefferson:
This code seems in a basically good state. Obviously tests for it, and using it, are important but can be part of future work.

Only one thought I have now, before this is merged -- given that most people don't read manuals, I worry that people might see `IsMinimalMatrixRep` and think it is small, or minimal, and start using it. At the moment, that wouldn't be a completely stupid choice, as it does compress the matrix quite well.

As it might get slower (and less useful) as tests are added to it, I might give it a more explicit name, like `IsExampleMinimalMatrixRep` (other names OK of course, just something to make it clear people shouldn't be using it).

wucas:
@ChrisJefferson Thanks for the suggestion. I didn't think of that way to understand the name but it is sensible to make it absolutely clear in the name that it is not intended for usage in computations. I decided for the name `IsMinimalExampleMatrixRep`.

fingolfin:
Hint: if you replace `This takes care of Issue #5142.` by `This resolves #5142.` then it will automatically close that issue when this PR is closed (we can also activate this  by manually "linking" the issue and PR, but I think it's good to know about the possibility to activate this with keywords, it's IMHO quite useful.

lib/matobjminimal.gd
```diff
@@ -0,0 +1,80 @@
...
+##  </Enum>
+##  It implements the MatrixObj specification in a minimal way, that is it 
+##  implements exactly the required methods. This is intended for testing and
+##  development and should not be used for actual calculations!
+##  </Description>
+##  </ManSection>
+##  <#/GAPDoc>
+##
+# Here we declare the new representation and tell GAP which properties it
+# implies. FListMatrices e.g. are positional objects and so on.
```
fingolfin, line 57: "FListMatrices" ? -> lots of copy & paste in various PR ;-)

lib/matobjminimal.gi
```diff
@@ -0,0 +1,96 @@
...
+end );
+
+InstallMethod( ConstructingFilter, "for a minimal example matrix rep",
+  [ IsMinimalExampleMatrixRep ],
+  function( mat )
+    return IsMinimalExampleMatrixRep;
+end );
+
+InstallMethod( CompatibleVectorFilter, [ IsMinimalExampleMatrixRep ],
+  M -> IsPlistVectorRep );
```
fingolfin, line 96:
```suggestion
  M -> IsPlistVectorRep );

```

ChrisJefferson (approved):

Issue #5135, opened 2022-10-19
`MatrixObj `doesn't check the `ncols` - while creating


mihalova:
### Observed behaviour
I created the MatrixObj with the command `Matrix( R, list, ncols )`. As `list` I put an 1x3 matrix, but for the `ncols` parameter I put 4. However, the MatrixObj was created. I tried to access the element (1,4) and GAP gave me an error.

### Expected behaviour
I would expect either change in the `ncols` parameter or adding an empty elements to the matrix.

### GAP example
```
gap> mat:=Matrix(Integers, [[1,2,3]], 4); 
<1x4-matrix over Integers>
gap> Print(mat);
NewMatrix(IsPlistMatrixRep,Integers,4,[ [ 1, 2, 3 ] ])
gap> mat[1,4];
Error, List Element: <list>[4] must have an assigned value in
  return m![ROWSPOS][row]![ELSPOS][col]
 ; at /Users/dominika/gap-4.12.0/lib/matobjplist.gi:824 called from 
<function "MatElm for a plist matrix and two positions">( <arguments> )
 called from read-eval loop at *stdin*:129
type 'quit;' to quit to outer loop
```

fingolfin:
This has been fixed in PR #5153

Issue #5133, opened 2022-10-19
Create `MatrixObj` with wrong `BaseDomain`


mihalova:
### Observed behaviour
MatrixObj allows creating a matrix with BaseDomain which is different from the domain of values inside the list.
I used the command `Matrix( R, list )`.
I tried to create the MatrixObj with the parameter `Integers` as `R` and I put a matrix with `Rationals` for the second parameter `list`, e.g. 1/4, 1.5. The GAP created the MatrixObj without checking the domains of the parameters.

### GAP example
```
gap> mat:=Matrix(Integers, [[1.2,1/3,1.4]]);  
<1x3-matrix over Integers>
gap> Print(Matrix(Integers, [[1.2,1/3,1.4]]));
NewMatrix(IsPlistMatrixRep,Integers,3,[ [ 1.2, 1/3, 1.4 ] ])
```

ThomasBreuer:
This had been observed already in #4884.

fingolfin:
Thanks for the report! We are closing this as a duplicate of #4884.

Issue #4884, opened 2022-04-28
Inverse of a non-invertible matrix over the integers is a (non-integer) integer matrix


james-d-mitchell:
In the current GAP master branch I get the following, inverting a non-invertible integer `MatrixObj` results in a matrix that believes it is an integer matrix but whose entries are not integers. 

### Observed behaviour
~~~
gap> mat := Matrix(Integers, [[0, -3, 0, -2], [-1, 1, -1, 0],
> [0, 1, 0, 1], [0, 0, 2, 0]]);;
gap> InverseOp(mat);
<4x4-matrix over Integers>
gap> Display(last);
<4x4-matrix over Integers:
[[ -1, -1, -2, -1/2 ]
 [ -1, 0, -2, 0 ]
 [ 0, 0, 0, 1/2 ]
 [ 1, 0, 3, 0 ]
]>
~~~
### Expected behaviour
~~~
gap> mat := Matrix(Integers, [[0, -3, 0, -2], [-1, 1, -1, 0],
> [0, 1, 0, 1], [0, 0, 2, 0]]);;
gap> InverseOp(mat);
fail
~~~
or perhaps at a stretch:
~~~
gap> mat := Matrix(Integers, [[0, -3, 0, -2], [-1, 1, -1, 0],
> [0, 1, 0, 1], [0, 0, 2, 0]]);;
gap> InverseOp(mat);
<4x4-matrix over Rationals>
gap> Display(last);
<4x4-matrix over Rationals:
[[ -1, -1, -2, -1/2 ]
 [ -1, 0, -2, 0 ]
 [ 0, 0, 0, 1/2 ]
 [ 1, 0, 3, 0 ]
]>
~~~

hulpke:
This is an interesting question: What should `Inverse` do for a matrix declared over a ring, which is invertible over the fraction field? And should `Matrix` check whether the given entries actually lie over the ring? @ThomasBreuer might know.

ThomasBreuer:
Concerning the example `Matrix(Integers, [[1, 1], [2, E(8)]])`, this input must yield either an error or `fail`, and the documentation of `Matrix` must state what happens in such cases.
Currently the documentation of `Matrix` just states that the returned matrix is defined over the given base domain, and the documentation of `BaseDomain` refers to an introductory section that states that the matrix entries must either lie in the base domain or naturally embed into the base domain.
Thus `Matrix` must check this property, and perhaps something like `MatrixNC` should be provided, which omits these checks.

[...]

PR #5153, opened 2022-10-20
support consistency checks for `Vector`, `Matrix`

ThomasBreuer:
The idea is to admit/force consistency checks in the constructions of vector and matrix objects, via `Vector`, `NewVector`, `Matrix`, `NewMatrix`:
One either may want to make sure that the given arguments are consistent (e.g., the entries are really in the given ring, the intended shape of a matrix fits to the given number of columns, and the entries also fit to the intended representation) or one wants to omit such checks if possible.

Some problems have been described in #4884, #5133, #5135.
(Consistency checks will also be needed in operations that change existing vector and matrix objects, such as assignments via `:=`.)

I propose to use a global option `Check` that is supported by `Vector`, `NewVector`, `Matrix`, and `NewMatrix`, as follows:
If `Check` is `false` then the code for methods of these operations is allowed to omit consistency checks, otherwise consistency checks are recommended for methods of `NewVector` and `NewMatrix`.

(Before that, I had tried the idea to introduce new operations `VectorNC` and `NewVectorNC` (likewise for matrices) such that
- `Vector` computes the necessary data for calling `NewVector`,
- `VectorNC` computes the necessary data for calling `NewVectorNC`,
- `NewVector` first checks for consistency and then calls `NewVectorNC`.
Up to now, each implementation of a new type of vector objects must provide a `NewVector` method. After the change, a `NewVectorNC` method would be needed for each type of vector objects.
Due to the fact that `NewVector` and `NewVectorNC` must be constructors (in the sense of GAP, i. e., they get created with `DeclareConstructor`), we cannot have one default method for `NewVector` with the above behaviour.
Thus each implementation of a new type of vector objects would have to provide also a `NewVector` method.
This idea does not look sensible.)

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -180,6 +180,17 @@ end);
 
 # methods to create vector objects
 
+#############################################################################
+##
+#O  Vector( <filt>, <R>, <list> )
+#O  Vector( <filt>, <R>, <vec> )
+#O  Vector( <R>, <list> )
+#O  Vector( <R>, <vec> )
+#O  Vector( <list>, <vec> )
+#O  Vector( <vec1>, <vec2> )
+##
+##  Compute the missing arguments for 'NewVector' and then call it.
+##
 InstallMethod( Vector,
     [ IsOperation, IsSemiring, IsList ],
     NewVector );
@@ -282,6 +293,8 @@ InstallMethod( ZeroVector,
 #M  Matrix( <M1>, <M2> )
 #M  Matrix( <list>, <ncols> )
 #M  Matrix( <list> )
+##
+##  Compute the missing arguments for 'NewMatrix' and then call it.
 ##
 InstallMethod( Matrix,
   [ IsOperation, IsSemiring, IsList, IsInt ],
@@ -389,6 +402,7 @@ InstallMethod( Matrix,
     # perhaps adjust NewMatrix to take an IsMatrixOrMatrixObj?
     return NewMatrix( ConstructingFilter(example), BaseDomain(example), NrCols(mat), Unpack(mat) );
   end );
+
 
 #
 #
--- a/lib/matobj2.gd
+++ b/lib/matobj2.gd
@@ -646,6 +646,9 @@ DeclareOperation( "ZeroVector", [ IsInt, IsVecOrMatObj ] );
 ##  this list.
 ##  <P/>
 ##  It is <E>not</E> guaranteed that the given list of entries is copied.
+##  <P/>
+##  If the global option <C>Check</C> is set to <K>false</K> then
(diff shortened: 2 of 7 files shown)


lib/matobj2.gd
```diff
@@ -1222,6 +1231,9 @@ DeclareOperation( "CompanionMatrix",
 ##  of <Ref Oper="ShallowCopy"/>.
 ##  If <A>list</A> is a nested list then it is <E>not</E> guaranteed
 ##  that also the entries of <A>list</A> are copied.
+##  <P/>
+##  If the global option <C>Check</C> is set to <K>false</K> then
```
fingolfin, line 1235: I note that the majority of our existing global options are all-lowercase. I wonder if we have established any rules on this anywhere?

lib/matobj.gi
```diff
@@ -277,6 +288,8 @@ InstallMethod( ZeroVector,
 #M  Matrix( <list>, <ncols> )
 #M  Matrix( <list> )
 ##
+##  Compute the missing arguments for 'NewMatrix' and then call it.
+##
 InstallMethod( Matrix,
   [ IsOperation, IsSemiring, IsList, IsInt ],
   { filt, R, list, nrCols } -> NewMatrix( filt, R, nrCols, list ) );
```
fingolfin, line 295: Wasn't the idea to insert (optional) checks into these `Matrix` functions, to validate things like that `nrCols` and `Length(list)` are compatible, that the given lists don't have holes, that entries are contained in the given basedomain etc ?

ThomasBreuer: My point of view is as follows:

The documentation says that implementing new types of matrix objects requires the installation of a `NewMatrix` method,
whereas there are default `Matrix` methods which compute the missing arguments for `NewMatrix`.
The question whether the arguments of `Matrix` (and hence the arguments that are then passed to `NewMatrix`) are consistent depends on the type of the matrix objects in question; it may be required that all matrix entries lie in the base domain, or more general situations are possible. The documentation is (deliberately?) vague in this respect, and adding the concrete consistency conditions to the `NewMatrix` method in question will make precise what is allowed for the given type of matrix objects.

I think that consistency checks in `Matrix` methods can be either only partial checks (which need more checks in the `NewMatrix` methods) or default checks (which need individual `Matrix` methods with modified checks if the default does not fit). Wouldn't that be more complicated in the end?

fingolfin:
@ThomasBreuer so should I merge this, or do you plan to make changes based on our discussion from last week? (I am not sure if / which changes are needed right now, if any; but e.g. perhaps `Check` renamed to `check`?)

fingolfin (approved):

PR #5155, opened 2022-10-20
Don't use` One` on collections that are not domains

fingolfin:
This patch is a few months old, I think meant to workaround some problem with semigroups. I think this was resolved differently but I still think the change is generally a good idea...


tst/testinstall/semipperm.tst
```diff
@@ -289,7 +289,9 @@ true
 # Test IsomorphismPartialPermMonoid for a partial perm semigroup
 gap> S := Semigroup(PartialPerm([2], [2]), PartialPerm([1], [1]));;
 gap> IsomorphismPartialPermMonoid(S);
-Error, the argument must be a semigroup with a multiplicative neutral element
+MappingByFunction( <partial perm monoid of rank 2 with 2 generators>,
```
james-d-mitchell, line 292: This makes no sense I’m afraid, the semigroup generated by the two partial perms given is not a monoid, and I think that with the changes in this pr it wouldn’t be the case that the return value of One was a multiplicative neutral element. I’m not sure what the motivation behind this is, and I’m pretty sure that making this change will badly break lots of the code for Semigroups

jackschmidt: In case it helps others reviewing, the identity returned by One is not the identity for the entire family of partial perms, but only for those partial perms with the same support. This is very different from the permutation case where all permutations have the same support (all positive integers). This means that One would need to be called on the list or the domain generated by the list, but this is the function that creates that domain, so seems hard to get right without calling it on a list.

```
gap> gens := [ PartialPerm([2], [2]), PartialPerm([1], [1]) ];
[ <identity partial perm on [ 2 ]>, <identity partial perm on [ 1 ]> ]
gap> One(gens);
<identity partial perm on [ 1, 2 ]>
gap> One(Representative(gens));
<identity partial perm on [ 2 ]>
gap> One(gens[1]);
<identity partial perm on [ 2 ]>
```

ThomasBreuer:
I understand this pull request as a step towards removing the `One` method in `lib/arith.gi` that takes an argument `coll` in the filter `IsMultiplicativeElementWithOneCollection` and returns `One(Representative(coll))`, provided that `coll` is not itself in the filter `IsMultiplicativeElementWithOne`.
I think it was a bad idea to have this method, and a comment in the code states that it causes logical problems.

The remark by @james-d-mitchell points to a second problem. Namely, `lib/pperm.gi` introduces another `One` method, which takes an argument `coll` in `IsPartialPermCollection`.
My interpretation is that the abovementioned `One` method for `IsMultiplicativeElementWithOneCollection` is not suitable for the situation of an `IsPartialPermCollection` (thanks to @jackschmidt for the explanation), but instead of avoiding the call of `One` in this situation, a different method of higher rank was installed that returns something reasonable.

I think the right solution would be to eventually get rid of both `One` methods, and for the moment to change the library code such that these methods are not used, in each case according to the `One` method in question.

PR #5156, opened 2022-10-21
Switch (Set)MatElm with `[,]` resp. `[,]:=`

fingolfin:
This should only affect printing, and is for consistency with `[]` and `[]:=`.

Not sure whether this is a good or a bad idea...


wucas:
I'd say this is a good idea (although I might not be aware of all the consequences this change has). Certainly from a users perspective who didn't read the MatObj documentation or didn't look at any MatObj code (which we can't and shouldn't expect of course) it might be confusing to get an error message with a function the user didn't call (explicitly).

wucas (approved):

PR #5157, opened 2022-10-21
Various fixes for package loading

zickgraf:
See commit messages for details.

## Text for release notes

none (as its hard to describe actual situations which get fixed by this)

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -355,7 +355,7 @@ InstallGlobalFunction( InitializePackagesInfoRecords, function( arg )
     # if the two arguments are equal.)
     StableSortParallel( List( GAPInfo.PackagesInfo, r -> r.Version ),
                   GAPInfo.PackagesInfo,
-                  { a, b } -> a <> b and CompareVersionNumbers( a, b ) );
+                  { a, b } -> not CompareVersionNumbers( a, b, "equal" ) and CompareVersionNumbers( a, b ) );
 
     # Turn the lists into records.
     record:= rec();
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -379,7 +379,7 @@ InstallGlobalFunction( InitializePackagesInfoRecords, function( arg )
     GAPInfo.PackagesInfoInitialized:= true;
     LogPackageLoadingMessage( PACKAGE_DEBUG,
         "return from InitializePackagesInfoRecords", "GAP" );
-    end );
+end );
 
 
 #############################################################################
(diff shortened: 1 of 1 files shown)


lib/package.gi
```diff
@@ -355,7 +355,7 @@ InstallGlobalFunction( InitializePackagesInfoRecords, function( arg )
     # if the two arguments are equal.)
     StableSortParallel( List( GAPInfo.PackagesInfo, r -> r.Version ),
                   GAPInfo.PackagesInfo,
-                  { a, b } -> a <> b and CompareVersionNumbers( a, b ) );
+                  { a, b } -> not CompareVersionNumbers( a, b, "equal" ) and CompareVersionNumbers( a, b ) );
```
fingolfin, line 358: Regarding release notes you write
> none (as its hard to describe actual situations which get fixed by this)
which is fine by me, but I still wonder, what *are* the conditions that this meant to fix here? :-)

lib/package.gi
```diff
@@ -1898,9 +1901,6 @@ InstallGlobalFunction( AutoloadPackages, function()
     fi;
     LogPackageLoadingMessage( PACKAGE_DEBUG, msg, "GAP" );
```
fingolfin, line 1902: I assume you move the `SetInfoLevel` so that the above `LogPackageLoadingMessage` is shown? That makes sense to me

zickgraf: Have you looked at the commit message bodies? If yes I'm of course happy to elaborate, but from your questions I have the feeling that they contain the details you are looking for, that's why I'm asking,

fingolfin: Sorry I missed the bit about the commit messages

lib/package.gi
fingolfin: OK, this is explained in the comments.

Anyway, I think a simpler implementation would be this (well... shorter... not sure if it is simpler), by using that `!(b <= a) == (a < b)` for a total ordering
```suggestion
                  { a, b } -> not CompareVersionNumbers( b, a ) );
```

fingolfin: (Perhaps it would be nicer if `CompareVersionNumbers` also supported an argument `"strict"`, or perhaps `"<"` and `"="` (and then also `"<="` for completeness?)

fingolfin (approved):

lib/package.gi
zickgraf: > Perhaps it would be nicer if `CompareVersionNumbers` also supported an argument `"strict"`, or perhaps `"<"` and `"="` (and then also `"<="` for completeness?)

I think something along this line would be the cleanest approach. But also the most time-consuming, that's why I did not want to touch `CompareVersionNumbers`.

Issue #4512, opened 2021-05-27
MatrixObj: Add type-parametrized tests


ssiccha:
Add test functions which take as argument something like `IsGF2MatrixRep` (to
indicate a specific MatrixObj implementation) plus some other arguments (e.g. a
base domain) and then run a bunch of tests to verify that certain APIs are
implemented right. E.g.  a ZeroMatrix test could ensure that matrices of the
right dimension, domain and type are produced.

The file `tst/teststandard/arithlst.g` might give an idea how such generic
tests can look.

fingolfin:
Also take a look at `tst/testinstall/MatrixObj/testmatobj.g`

fingolfin:
@AnnaKDS and @wucas are working on this as part of GAP Days Summer 2022

fingolfin:
The old attempt in PR #5162 unfortunately is not easily recoverable in my eyes, although we can certainly use it as a quarry and see if we can extract individual tests.

I think we need a different approach. The first thing to do is to ditch any concerns for "special" matrix types that e.g. only support matrices of a specific shape (upper triangular, monomial, etc.): that decision immediately simplifies things. For example, it becomes trivial to turn something like this (from `tst/testinstall/MatrixObj/ElementaryMatrices.tst`, but see also `tst/testinstall/matrix.tst`) into a test function that one can call with a matrix obj filter:
```
gap> mat := Matrix([ [ 1, 0, 0 ], [ 0, 2, 0 ] ]);;
gap> IsDiagonalMatrix(mat);
true
gap> IsUpperTriangularMatrix(mat);
true
gap> IsLowerTriangularMatrix(mat);
true

#
gap> mat := Matrix([ [ 1, 2, 0 ], [ 0, 3, 0 ] ]);;
gap> IsDiagonalMatrix(mat);
false
gap> IsUpperTriangularMatrix(mat);
true
gap> IsLowerTriangularMatrix(mat);
false

#
gap> mat := Matrix([ [ 1, 0, 0 ], [ 4, 3, 0 ] ]);;
gap> IsDiagonalMatrix(mat);
false
gap> IsUpperTriangularMatrix(mat);
false
gap> IsLowerTriangularMatrix(mat);
true
```

[...]

PR #5162, opened 2022-10-21
Provide a generic test suite for MatrixObjects

wucas:
This PR resolves Issue #4512. 

Still missing:

- [ ] Comments
- [ ] More tests
- [ ] ability to test wrong inputs, i.e. situations when it is expected and correct that an error is thrown
- [ ] Documentation

### DIFF
--- a/tst/testinstall/MatrixObj/testmatobj_new.g
+++ b/tst/testinstall/MatrixObj/testmatobj_new.g
@@ -137,11 +137,11 @@ end;
 MatObjTest_TestBaseDomain := function(ex, opt, errors)
     local domain;
 
-    domain := MatObjTest_CallFunc(BaseDomain,[ex.matobj], opt.breakOnError, errors);
+    domain := MatObjTest_CallFunc(BaseDomain,[ex.matObj], opt.breakOnError, errors);
 
     if domain <> fail then
         if domain <> ex.baseDomain then
-            MatObjTest_HandleErrorWrongResult(ex.filter, "BaseDomain", [ex.matobj], opt.breakOnError, errors);
+            MatObjTest_HandleErrorWrongResult("BaseDomain", [ex.matObj], opt.breakOnError, ex, errors);
         fi;
     fi;
 end;
@@ -149,11 +149,11 @@ end;
 MatObjTest_TestNrRows := function(ex, opt, errors)
     local nrRows;
 
-    nrRows := MatObjTest_CallFunc(NrRows, [ex.matobj], opt.breakOnError, errors);
+    nrRows := MatObjTest_CallFunc(NrRows, [ex.matObj], opt.breakOnError, errors);
 
     if nrRows <> fail then
         if nrRows <> ex.nrRows then
-            MatObjTest_HandleErrorWrongResult(ex.filter, "NrRows", [ex.nrRows], opt.breakOnError, errors);
+            MatObjTest_HandleErrorWrongResult("NrRows", [ex.matObj], opt.breakOnError, ex, errors);
         fi;
     fi; 
 end;
@@ -161,11 +161,11 @@ end;
 MatObjTest_TestNrCols := function(ex, opt, errors)
     local nrCols;
 
-    nrCols := MatObjTest_CallFunc(NrCols, [ex.matobj], opt.breakOnError, errors);
-
+    nrCols := MatObjTest_CallFunc(NrCols, [ex.matObj], opt.breakOnError, errors);
+    
     if nrCols <> fail then
         if nrCols <> ex.nrCols then
-            MatObjTest_HandleErrorWrongResult(ex.filter, "NrCols", [ex.nrCols], opt.breakOnError, errors);
+            MatObjTest_HandleErrorWrongResult("NrCols", [ex.matObj], opt.breakOnError, ex, errors);
         fi;
     fi; 
 end;
@@ -186,6 +186,10 @@ TestMatrixObj := function(filter, opt)
 
     if not IsBound(opt.forbidInv) then
         opt.forbidInv := false;
+    fi;
+
+    if not IsBound(opt.breakOnError) then
+        opt.breakOnError := false;
     fi;
 
     examples := MatObjTest_GenerateExample(filter, opt);
--- a/tst/testinstall/MatrixObj/testmatobj_new.g
+++ b/tst/testinstall/MatrixObj/testmatobj_new.g
@@ -170,6 +170,47 @@ MatObjTest_TestNrCols := function(ex, opt, errors)
     fi; 
 end;
 
+MatObjTest_TestMatElm := function(ex, opt, errors)
+    local col, row, elm;
+
+    col := NrCols(ex.matObj);
+    row := NrRows(ex.matObj);
+    elm := MatObjTest_CallFunc(MatElm, [ex.matObj, row, col], opt.breakOnError, errors);
+
+    if elm <> fail then
+        if elm <> ex.sourceOfTruth[col, row] then
+            MatObjTest_HandleErrorWrongResult("MatElm", [ex.matObj, row, col], opt.breakOnError, errors);
+        fi;
+    fi;
+end;
+
+MatObjTest_TestSetMatElm := function(ex, opt, errors)
+    local col, row, elm;
+
+    col := NrCols(ex.matObj);
+    row := NrRows(ex.matObj);
+    elm := Zero(ex.baseDomain);
+    ex.sourceOfTruth[row, col] := elm;
+    ex.mat[row, col] := elm;
+    MatObjTest_CallFunc(SetMatElm, [ex.matObj, row, col, elm], opt.breakOnError, errors);
+
+    if ex.matObj[row, col] <> ex.sourceOfTruth[row, col] then
+        MatObjTest_HandleErrorWrongResult("SetMatElm", [ex.matObj, row, col, elm], opt.breakOnError, errors);
+    fi;
+end;
+
+MatObjTesT_TestConstructingFilter := function(ex, opt, errors)
+    local filter;
+
+    filter := MatObjTest_CallFunc("ConstructingFilter", [ex.matObj], opt.breakOnError, errors);
+
+    if filter <> fail then
+        if filter <> ex.filter then
+            MatObjTest_HandleErrorWrongResult("ConstructingFilter", [ex.matObj], opt.breakOnError, errors);
+        fi;
+    fi;
+end;
+
 
 TestMatrixObj := function(filter, opt)
     local errors, examples, ex;
@@ -204,6 +245,14 @@ TestMatrixObj := function(filter, opt)
 
     for ex in examples do
         MatObjTest_TestNrCols(ex, opt, errors);
+    od; 
+
+    for ex in examples do
+        MatObjTest_TestMatElm(ex, opt, errors);
+    od; 
+
+    for ex in examples do
+        MatObjTest_TestSetMatElm(ex, opt, errors);
     od; 
 
     #TODO other tests
--- a/tst/testinstall/MatrixObj/testmatobj_new.g
+++ b/tst/testinstall/MatrixObj/testmatobj_new.g
@@ -1,3 +1,61 @@
+MatObjTest_CallAndCatchError := function(f, args)
+    local out, res, oldBreakOnError, oldErrorOutput, str;
+
+    str := "";
+    out := OutputTextString(str, true);
+
+    oldBreakOnError := BreakOnError;
+    oldErrorOutput := ERROR_OUTPUT;
+    BreakOnError := false;
+    MakeReadWriteGlobal("ERROR_OUTPUT");
+    ERROR_OUTPUT := out;
+    res := CALL_WITH_CATCH(f, args);
+    BreakOnError := oldBreakOnError;
+    ERROR_OUTPUT := oldErrorOutput;
+    CloseStream(out);
+    MakeReadOnlyGlobal("ERROR_OUTPUT");
+    return [res,str];
+end;
+
+MatObjTest_AppendErrorFail := function(f, args, catch, errors)
+    local errorMsg, arg;
+
+    errorMsg := Concatenation("The function ", NameFunction(f), " failed with Error Message:\n", catch[2], "\n Called with arguments:\n");
+    for arg in args do 
+        Append(errorMsg, Concatenation(String(arg),"\n\n"));
+    od;
+    Add(errors, errorMsg);
+end;
+
+MatObjTest_HandleErrorWrongResult := function(msg, args, breakOnError, ex, errors)
+    local errorMsg, arg;
+
+    if breakOnError then 
+        Error(msg);
+    else 
+        errorMsg := Concatenation(msg, "\n Called with arguments:\n");
+        for arg in args do 
+            Append(errorMsg, Concatenation(String(arg),"\n\n"));
+        od;
+        Add(errors, errorMsg);
+    fi;
+end;
+
+MatObjTest_CallFunc := function(f, args, breakOnError, errors)
+    local catch;
+    if breakOnError then 
+        return CallFuncList(f, args);
+    else
+        catch := MatObjTest_CallAndCatchError(f, args);
+        if catch[1][1] then 
+            return catch[1][2];
+        else 
+            MatObjTest_AppendErrorFail(f, args, catch, errors);
+            return fail;
+        fi;
+    fi;
+end;
+
 MatObjTest_GenerateExample := function(filter, opt)
     local inv, dims, dim, doms, dom, customGen, examples;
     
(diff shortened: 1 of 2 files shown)


tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+
+MatObjTest_GenerateExample := function(filter, opt)
+    local inv, dims, dim, doms, dom, customGen, examples;
+    
+    examples := [];
+
+    if IsBound(opt.dimensions) then 
+        dims := opt.dimensions;
+    else 
+        dims := [[2,2],[3,3],[2,5],[5,5],[1,3],[4,1],[6,2]];
```
fingolfin, line 90: How about also `[1,1]` ?

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+    for ex in examples do
+        MatObjTest_TestNrCols(ex, opt, errors);
+    od; 
+
+    for ex in examples do
+        MatObjTest_TestMatElm(ex, opt, errors);
+    od; 
+
+    for ex in examples do
+        MatObjTest_TestSetMatElm(ex, opt, errors);
```
fingolfin, line 278: This tests only works for MatrixObj types that allow modification. If the matrices are immutable, we should not try this resp. should not let this trigger an error

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+    od; 
+
+    for ex in examples do
+        MatObjTest_TestConstructingFilter(ex, opt, errors);
+    od;
+
+    #TODO other tests
+
+    if not opt.forbidInv then
+        opt.invertible := true;
```
fingolfin, line 288: This doesn't fit with the code using `opt.invertible` where `opt.invertible` is -- at least in some cases -- assumed to be a generator function

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+
+    if IsBound(opt.sourceOfTruth) then 
+        return List(examples, x-> 
+            rec(matObj := NewMatrix(filter, x[4], x[3], x[1]),
+            filter := filter,
+            baseDomain := x[4],
+            nrCols := x[3],
+            nrRows := x[2],
+            mat := x[1],
+            sourceOfTruth := NewMatrix(opt.sourceOfTruth, x[4], x[3], x[1]))
```
fingolfin, line 143: As a reminder from our video chat: Note how `sourceOfTruth` has two different meanings here (once it is a matrix, once it is a filter)

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+        #return List(examples, x -> [NewMatrix(filter, x[3], x[2], x[1]), NewMatrix(opt.sourceOfTruth, x[3], x[2], x[1])]);
+    else 
+        return List(examples, x-> 
+            rec(matObj := NewMatrix(filter, x[4], x[3], x[1]),
+            filter := filter,
+            baseDomain := x[4],
+            nrCols := x[3],
+            nrRows := x[2],
+            mat := x[1],
+            sourceOfTruth := x[1])
```
fingolfin, line 154: Code duplication could be reduced here by doing something like
```suggestion
            sourceOfTruth := f(x))
```
and then do something like this:
```gap
    if IsBound(opt.sourceOfTruth) then
        f := x -> NewMatrix(opt.sourceOfTruth, x[4], x[3], x[1]));
    else
        f := x -> x[1];
    fi;
```

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+            MatObjTest_HandleErrorWrongResult("ConstructingFilter", [ex.matObj], opt.breakOnError, ex, errors);
+        fi;
+    fi;
+end;
+
+
+TestMatrixObj := function(filter, opt)
+    local errors, examples, ex;
+
+    errors := [];
```
fingolfin, line 241: Maybe copy `opt` before modifying it:
```suggestion
    errors := [];
    opt := ShallowCopy(opt);
```

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+    od;
+
+    #TODO other tests
+
+    if not opt.forbidInv then
+        opt.invertible := true;
+        examples := MatObjTest_GenerateExample(filter, opt);
+
+        for ex in examples do
+            #TestInverse(ex);
```
fingolfin, line 292: We may also wish to do fewer tests (and with smaller dimensions) for a slower operation like `Inverse` or `MinimalPolynomial` compared to somethign very fast with `NrRows`

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+            sourceOfTruth := NewMatrix(opt.sourceOfTruth, x[4], x[3], x[1]))
+            );
+        #return List(examples, x -> [NewMatrix(filter, x[3], x[2], x[1]), NewMatrix(opt.sourceOfTruth, x[3], x[2], x[1])]);
+    else 
+        return List(examples, x-> 
+            rec(matObj := NewMatrix(filter, x[4], x[3], x[1]),
+            filter := filter,
+            baseDomain := x[4],
+            nrCols := x[3],
+            nrRows := x[2],
```
fingolfin, line 152: Instead of storing these here, how about you immediately perform a test here that `NrCols(matObj) = x[3]` etc. holds?

Note that it is *not* necessary here to catch an error! If there is an error in `NrCols`, then we enter a break loop, and the developer can and must deal with it immediately.

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,299 @@
...
+    col := ex.nrCols;
+    row := ex.nrRows;
+    elm := MatObjTest_CallFunc(MatElm, [ex.matObj, row, col], opt.breakOnError, errors, ex);
+
+    if elm <> fail then
+        if elm <> ex.sourceOfTruth[row,col] then
+            MatObjTest_HandleErrorWrongResult("MatElm", [ex.matObj, row, col], opt.breakOnError, ex, errors);
+        fi;
+    fi;
+end;
```
fingolfin, line 208: As discussed: perhaps `MatObjTest_CallFunc` and  `MatObjTest_HandleErrorWrongResult` could be called by the code invoking `MatObjTest_TestMatElm` etc., and the code here would become something like this:
```suggestion
MatObjTest_TestMatElm := function(ex)
    local col, row, elm;
    for row in [1..NrRows(ex.matobj)] do
        for col in [1..NrCols(ex.matobj)] do
            elm := ex.matobj[row, col];
            if elm <> ex.sourceOfTruth[row,col] then
                return false;
            fi; 
        od;
    od;
    return true;
```
We'd loose some level of information in the error report, but we'd make it a lot easier to write and debug these test cases, which is not irrelevant given that we need to write a lot, and then people implementing matrixObj need to interact with them.

My argument in the discussion we just had was that the lost information is not important: to debug this, as a developer I would turn on `breakOnError` and re-run the test to debug it and find out the position the error occurred. Then I might want to write a stand-alone regression test case; for this it is of course very helpful that you print a `NewMatrix(...)` call, but that's enough for me. (If we really want more info, we can always add back a way to do that: e.g. return a string instead of a bool, or a pair `[result_bool, error_desc]`; or provide a callback function `MatObjTest_AddErrorDetails` (or whatever). But for now I'd just leave this out, and wait what our experience is with using the test suite.

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: The filter specifying the sourceOfTruth is now called sourceOfTruthFilter. It is only used in the record of options passed to the function. The examples generated and used in the actual tests are records as well. There sourceOfTruth is always a matrix (either some MatricObj or a classical GAP matrix).

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: There now is a specific option to specify a different set of dimensions and domains to use for expensive tests. There is even an option to provide a specific list of examples to use in these cases.

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+    fi;
+ 
+    result := [];
+
+    for ex in examples do
+        Add(result, rec(matObj := NewMatrix(filter, ex[4], ex[3], ex[1]),
+            mat := ex[1],
+            sourceOfTruth := getSourceOfTruth(ex)));
+        if BaseDomain(Last(result).matObj) <> ex[4] or NumberRows(Last(result).matObj) <> ex[2] or NumberColumns(Last(result).matObj) <> ex[3] or ConstructingFilter(Last(result).matObj) <> filter then 
+            ErrorNoReturn(Concatenation("Error while generating examples. In a case with NrRows = ", String(ex[2], " NrCols = ", String(ex[3]), " domain = ", String(ex[4]), "and filter := ", String(filter), " NumberRows, NumberColumns, BaseDomain or ConstructingFilter does not work.")));
```
fingolfin, line 225: It should not be necessary to concatenate the arguments for `Error` or `ErrorNoReturn`; and then one also does not need to call `String` on the non-string arugments

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+        getSourceOfTruth := x -> x[1];
+    fi;
+ 
+    result := [];
+
+    for ex in examples do
+        Add(result, rec(matObj := NewMatrix(filter, ex[4], ex[3], ex[1]),
+            mat := ex[1],
+            sourceOfTruth := getSourceOfTruth(ex)));
+        if BaseDomain(Last(result).matObj) <> ex[4] or NumberRows(Last(result).matObj) <> ex[2] or NumberColumns(Last(result).matObj) <> ex[3] or ConstructingFilter(Last(result).matObj) <> filter then
```
fingolfin, line 224: IMHO it'd be more readable if your store the generated matrix in its own local variable (which would have to be added to the `local` statement at the top, of course). E.g. like this
```suggestion
        matObj := NewMatrix(filter, ex[4], ex[3], ex[1]);
        Add(result, rec(matObj := matObj,
            mat := ex[1],
            sourceOfTruth := getSourceOfTruth(ex)));
        if BaseDomain(matObj) <> ex[4] or NumberRows(matObj) <> ex[2] or NumberColumns(matObj) <> ex[3] or ConstructingFilter(matObj) <> filter then 
```

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+        fi;
+    fi;
+end;
+
+# This function requires a filter defining a matrix object and a record of
+# options and generates a list of examples for tests according to the provided 
+# options. 
+# Examples are records with three entries:
+# matObj: The generated matrix object that is to be tested
+# mat: The corresponding calssical GAP matrix (i.e. matrix as list of rows as
```
fingolfin, line 90:
```suggestion
# mat: The corresponding classical GAP matrix (i.e. matrix as list of rows as
```

lib/matobjminimal_BROKEN.gi
```diff
@@ -0,0 +1,97 @@
...
+InstallMethod( \<, "for two minimal example matrices",
+  [ IsMinimalBROKENMatrixRep, IsMinimalBROKENMatrixRep ],
+  function( a, b )
+    return LT_LIST_LIST_DEFAULT(a![MINREP_ELSPOS],b![MINREP_ELSPOS]);
+end );
+
+InstallMethod( ConstructingFilter, "for a minimal example matrix rep",
+  [ IsMinimalBROKENMatrixRep ],
+  function( mat )
+    return IsMinimalExampleMatrixRep;
```
fingolfin, line 92: `Syntax warning: Unbound global variable`

lib/matobjminimal_BROKEN.gi
```diff
@@ -0,0 +1,97 @@
...
+        for rowindex in [1..Length(list_in)] do 
+            if Length(list_in[rowindex]) <> nrcols then
+                Error( "NewMatrix: Each row must have nrcols entries." );
+            fi;
+            for colindex in [1..nrcols] do 
+                list[(rowindex-1)*nrcols + colindex] := list_in[rowindex][colindex];
+            od;
+        od;
+    else
+        if Length(list_in) mod nrcols <> 0 then
```
fingolfin, line 36: You need to remove all trailing whitespace in this file, like on this line
```suggestion
        if Length(list_in) mod nrcols <> 0 then
```
Ideally, configure your editor to honor our `.editorconfig`, see https://editorconfig.org

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+#   matrices.
+# breakOnError:
+#   Boolean set to false by default. If set to true Errors or Wrong results
+#   cause a break loop. Otherwise problems are returned as a list of messages.
+# isImmutable:
+#    Boolean set to false by default. If set to true tests of functions which 
+#    change an object and therefore require it to be mutable are skipped. This
+#    option is intended for objects which are always immutable. 
+
+TestMatrixObj := function(filter, optIn)
```
fingolfin, line 342: What calls `TestMatrixObj` and where?

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+
+    errors := [];
+
+    opt := ShallowCopy(optIn);
+
+    if not IsBound(opt.randomSource) then
+        opt.randomSource := GlobalMersenneTwister;
+    fi;
+
+    # usage of state makes no sense here. State should be set if bound!
```
fingolfin, line 353: I don't understand this. "should be set if bound"??!? Huh?

Also you then deal with it after all?! 

Is this perhaps a TODO or FIXME comment?

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+# options. 
+# This function runs tests to check whether the specified MatrixObject
+# implementation implements required methods and if these and other methods are 
+# correct (in the sense that they give correct results for the cases tested 
+# here).
+# The following options are supported:
+# randomSource:
+#   See TestMatrixObj_GenerateExample
+# state:
+#   See TestMatrixObj_GenerateExample
```
fingolfin, line 299: So is `state` supported or not? The comment in the body of the function makes this unclear

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+# dimensions: 
+#   See TestMatrixObj_GenerateExample
+# dimensions_red: A list of tuples defining dimensions for matrices. If not set
```
fingolfin, line 302:
```suggestion
# dimensions_red:
#   A list of tuples defining dimensions for matrices. If not set
```
Instead of "If not set" here and "By default this is set" elsewhere and so on, I'd suggest to settle on a uniform way to describe defaults. E.g. at the end of a description, add, in a line of its own:
```
#   Default value: XYZ
```

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+#   this is set to [[1,1],[2,2],[3,3],[5,5]]. These dimensions are used to
+#   generate examples for more expensive tests. 
+# domains:
+#   See TestMatrixObj_GenerateExample
+# domains_red:
+#   A list of domains in which matrices should be generated. By default
+#   this is set to [Integers, Rationals, GF(2), GF(5)]. These domains are used 
+#   to generate examples for more expensive tests.
```
fingolfin, line 310: Based on what you write elsewhere, I assume "red" stands for "reduced" and not the color "red"? But why "reduced" ?

It might also be good to explain how this relates to `domains`. E.g. is this right? "If domains_red is given, then it is used instead of domains for certain expensive tests.

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+# reduced_cases: 
+#   A list of triples defining the number of rows, the number of columns and the
+#   domain over which examples are generated. By default this option is unbound.
+#   If bound the options dimensions_red and domains_red are ignored and instead
+#   the cases specified in reduced_cases are used to generate examples for
+#   expensive tests. The reason to use this is that otherwise when specifying
+#   domains and dimensions via domains_red and dimensions_red all combinations 
+#   of the specified dimensions and domains are used. If tests are very 
+#   expensive or just certain combinations should be tested one can use 
+#   reduced_cases instead.
```
fingolfin, line 320: This option is documented but does not seem to actually exist?

tst/testinstall/MatrixObj/testmatobj_new.g
```diff
@@ -0,0 +1,422 @@
...
+
+# This function requires a filter defining a MatrixObject and a record of 
+# options. 
+# This function runs tests to check whether the specified MatrixObject
+# implementation implements required methods and if these and other methods are 
+# correct (in the sense that they give correct results for the cases tested 
+# here).
+# The following options are supported:
+# randomSource:
+#   See TestMatrixObj_GenerateExample
```
fingolfin, line 297: Do you mean `MatObjTest_GenerateExample`?

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: That is the function one should call to execute the tests.

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: Yes it is now. TestMatrixObj sets the randomSource to the default if none is provided and resets it to the state if it's bound. If a randomSource other than the default one is used one should set its state manually before calling the test suite. We could, however, also reset other random sources to the state.

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: I chose 'reduced' since its an option to choose a smaller set of cases to test. There might be a better name though.

tst/testinstall/MatrixObj/testmatobj_new.g
wucas: Yeah that should have been a TODO comment. I fixed it now.

lib/matobjminimal_BROKEN.gi
wucas: I still had the older .editorconfig where that was disabled in this branch. I
removed them now.

PR #5163, opened 2022-10-21
Add generic DisplayString function

### LINKED ISSUE #5144: MatrixObj: add generic ViewObj/PrintObj/Display/ViewString/PrintString/DisplayString/String methods for MatrixObj
We want them anyway, because they are not part of the mandatory interface. And this way we could implement some nice extra features available to all MatrixObj, in a consistent way.

Some thoughts:

- `Print` is meant to output code which can be read in back; I am not sure whether we can do that in general, but if we want to try it, it could start with something like `NewMatrix(FILTER, BASEDOMAIN, MAT)` where `MAT` is the result of `Unpack` for the original matrix
- `ViewObj` and/or `Display` could replace zeros by `.` and generally produce something more for humans, e.g.
```
gap> some_matobj;
A 2x3 matrix over GF(5):
. 1 3
3 . 2
```

danielrademacher:
Fixes #5144

## Text for release notes

none

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -1419,8 +1419,32 @@ InstallMethod( ViewString,
     ViewStringForMatrixObj );
 
 InstallMethod( DisplayString,
-    [ IsMatrixOrMatrixObj ],
-    ViewStringForMatrixObj );
+[ IsMatrixOrMatrixObj ],
+function(M)
+local i,j,m,numberCols,numberRows,baseDom,zeroEle,outputstring;
+    numberRows := NrRows(M);
+    numberCols := NrCols(M);
+    baseDom := BaseDomain(M);
+    zeroEle := Zero(baseDom);
+    outputstring := "<";
+    if not IsMutable(M) then Append(outputstring, "immutable "); fi;
+    outputstring := Concatenation(outputstring,String(numberRows),"x",String(numberCols),"-matrix over ",String(baseDom),":\n");
+    Append(outputstring,"[");
+    for i in [1..numberRows] do
+        for j in [1..numberCols] do
+            if j = 1 then
+                Append(outputstring,"[");
+            else
+                Append(outputstring," ");
+            fi;
+            Append(outputstring,String(M[i,j]));
+            od;
+        Append(outputstring,"]\n");
+    od;
+    Append(outputstring,"]");
+    Append(outputstring, ">\n");
+    return outputstring;
+end);
 
 InstallMethod( String,
     [ IsMatrixObj ],


fingolfin:
You write that this "fixes" (should perhaps be: "resolves" ?) issue #5144, but as far as I can tell, it only addresses part of it?

lib/matobj.gi
```diff
@@ -1419,8 +1419,32 @@ InstallMethod( ViewString,
...
+[ IsMatrixOrMatrixObj ],
+function(M)
+local i,j,m,numberCols,numberRows,baseDom,zeroEle,outputstring;
+    numberRows := NrRows(M);
+    numberCols := NrCols(M);
+    baseDom := BaseDomain(M);
+    zeroEle := Zero(baseDom);
+    outputstring := "<";
+    if not IsMutable(M) then Append(outputstring, "immutable "); fi;
+    outputstring := Concatenation(outputstring,String(numberRows),"x",String(numberCols),"-matrix over ",String(baseDom),":\n");
```
fingolfin, line 1431: Perhaps this should use `Append`?

lib/matobj.gi
```diff
@@ -1419,8 +1419,32 @@ InstallMethod( ViewString,
...
+    Append(outputstring,"[");
```
fingolfin, line 1432: This causes the first line to be indented differently. I would either add a corresponding whitespace before each other line; or just make the `[` belong the end of the previous line (probably nicer). So instead of
```
<a 2x2 matrix over Rationals
[[1 2 ]
[3 4 ]
]>
```
print
```
<a 2x2 matrix over Rationals [
[ 1 2 ]
[3  4 ]
]>
```

The `ViewString` could be similar but perhaps without the entries being shown? I.e.
```
```

lib/matobj.gi
```diff
@@ -1419,8 +1419,32 @@ InstallMethod( ViewString,
...
+                Append(outputstring,"[");
+            else
+                Append(outputstring," ");
+            fi;
+            Append(outputstring,String(M[i,j]));
+            od;
+        Append(outputstring,"]\n");
+    od;
+    Append(outputstring,"]");
+    Append(outputstring, ">\n");
```
fingolfin, line 1445:
```suggestion
    Append(outputstring,"]>\n");
```

danielrademacher:
How should we deal with numbers of different length, i.e.
'''
gap> m:=NewMatrix(IsPlistMatrixRep,Rationals,2,[-1,-2,0,-4]);;
gap> Print(DisplayString(m));
<a mutable 2x2-matrix over Rationals [
[-1 -2]
[0 -4]
]>
'''

lib/matobj.gi
```diff
@@ -1419,8 +1419,32 @@ InstallMethod( ViewString,
...
+    return outputstring;
```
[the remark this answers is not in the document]
danielrademacher: Will be done. Which file would you prefer?

ChrisJefferson:
While I realise this would require significantly changing how your code is structured, one way to fix the "numbers of different length" would be to first make a 2d array of strings representing what you are going to print, then finding the maximum length of a string in that 2d array, and finally printing. Nicely formatting is going to require going over the matrix twice whatever you do (first time to get length, second time to output).

fingolfin:
You could also take a look `PrintArray` and either just call that, or take inspiration from that code.

danielrademacher:
> You could also take a look `PrintArray` and either just call that, or take inspiration from that code.

Thanks a lot! I could inspired from the PrintArray code.

ChrisJefferson:
Just checking, what's the state of this PR, does it want reviewing?

I don't mind if you are still working on it, just didn't want it to get ignored.

PR #5165, opened 2022-10-21
Fix a number of bugs in `SyntaxTree` and `SyntaxTreeCode`

### LINKED ISSUE #5117: Segfault when getting syntax tree of function with nested sublist extraction of level 3 (or more)
### Observed behaviour
```gap
gap> func := function ( x ) return [ [ [ 1 ] ] ]{[ 1 ]}{[ 1 ]}{[ 1 ]}; end;
function( x ) ... end
gap> func( 1 );
[ [ [ 1 ] ] ]
gap> SYNTAX_TREE( func );
zsh: segmentation fault (core dumped)  gap
```

### Expected behaviour

No segfault.

```
 ┌───────┐   GAP 4.13dev-80-g4a1b294 built on 2022-10-14 17:39:56+0200
```

ChrisJefferson:
I believe this fixes #5117, but it would be good if @zickgraf could have a look.

Note that this changes some previously generated trees (see the diff), but I believe they were always wrong, it's just that the integer `1` was being interpreted as an Expr (where it is `0`), but the integer `2` interpreted as an Expr caused a crash.


src/syntaxtree.c
```diff
@@ -159,6 +159,8 @@ static inline Obj NewSyntaxTreeNode(UInt1 tnum)
     return result;
 }
 
+#include <stdio.h>
+
```
fingolfin, line 163:
```suggestion
```
Presumably debugging leftovers? :-)

src/syntaxtree.c
```diff
@@ -864,7 +866,7 @@ static const CompilerT Compilers[] = {
     COMPILER_(EXPR_ELMS_LIST_LEV,
               ARG_EXPR_("lists"),
               ARG_EXPR_("poss"),
-              ARG_EXPR_("level")),
+              ARG_EXPR("level", ObjInt_UInt, UInt_ObjInt)),
```
fingolfin, line 869: Ahhh, that makes sense. Nice catch!

zickgraf:
Thanks a lot, works perfectly! And it confirms my expectation that a stack of `EXPR_ELMS_LIST_LEV` always terminates with an `EXPR_ELMS_LIST`, which means my solution in #5116 is actually correct :D

While reading the documentation again, I noticed that this level thing also exists for the usual list access (`EXPR_ELM_LIST_LEV`). Trying to use this gives a segfault which I guess has an analogous fix:
```
SYNTAX_TREE( function ( x ) return [ [ [ [ 1 ] ] ] ]{[ 1 ]}[1]{[1]}[1]; end )
```
Note: This is in no way important for me, I'm just trying to find further ways to produce crashes :D

fingolfin:
Thanks @zickgraf , good catch. Perhaps @ChrisJefferson will also fix that, too :-)

fingolfin (dismissed):

ChrisJefferson:
This PR grew slightly out of control while fixing some other SYNTAX_TREE issues, I'm happy to pull it apart.

Now it contains:

(outside of SYNTAX_TREE) while tracking down a crash I added asserts to SIZE_EXPR. This revealed for chars we make bags of size uchar(1), which mean we shouldn't the read and write Exprs (which are size 8) from them. As we already round allocations up to word size this wasn't actually doing any harm in practice, but let's fix it anyway.

I also (this is the slightly horrible code) generalised the SYNTAX_TREE code to support variadics in the middle of an object.

src/code.h
```diff
@@ -561,7 +561,13 @@ EXPORT_INLINE Int TNUM_EXPR(Expr expr)
 */
 #define CONST_ADDR_EXPR(expr) CONST_ADDR_STAT(expr)
 
-#define READ_EXPR(expr, idx) (CONST_ADDR_EXPR(expr)[idx])
+EXPORT_INLINE Stat READ_EXPR(Expr expr, Int idx) {
```
fingolfin, line 564: @ChrisJefferson need I say it? 😂

ChrisJefferson:
Just to point out (in particular if @zickgraf is using this), this PR changes how matrix indexing is done, to make it (correctly) handle 2d matrices. Also (last push, as I'm sat in the airport), I decided to add an assert which makes sure all of an expr is used in SyntaxTree. This showed up a bug in handling EXPR_REC_TILDE. On deep investigation, it turns out this doesn't need to be handled differently (as we don't need to worry here where the tildes bind to, just preserve their presence).

zickgraf (approved):
Thanks for the effort! I have run all of my tests using this PR and there are no regressions :-)

src/syntaxtree.c
```diff
@@ -194,29 +196,36 @@ static Obj SyntaxTreeDefaultCompiler(Obj result, Expr expr)
...
-            // special case: the last argument may have zero as decompiler,
-            // meaning that all remaining slots of the statement should be
-            // decompiled into a single list
-            const UInt offset = comp.arity - 1;
-            GAP_ASSERT(i == offset);
-
-            // compile the complete rest into one statement
+            // special case: one argument may have zero as decompiler,
+            // which expresses that this slot is "variadic", and
+            // is decompiled into a single list
```
fingolfin, line 205: Why is this part of the change necessary?

ChrisJefferson: We use `ARGS_EXPR` to denote a variadic section of a bag. Previously, the ARGS_EXPR section was always at the end of the bag, but in EXPR_ELM_LIST_LEV, the variadic bit (which represents the dimensions indexed is in the middle of the bag, so I need to support ARGS_EXPR in any position.

While I've already written this code, an alternative option would be to change how we store EXPR_ELM_LIST_LEV expressions (which after investigation seem to be the only type which don't put the variadic bit at the end), to put the variadic bit at the end -- that would obviously break any saved workspaces, but I can't think of any other serious problems it would cause.

fingolfin: Nah, it's fine, I just wanted to understand what's going on. Thank you!

fingolfin (approved):

PR #5167, opened 2022-10-24
Fix indentation when printing EXPR_ELM_MAT or EXPR_ELM_LIST_LEVEL

zickgraf:
See testfile for examples.

## Text for release notes

none

## Further details

I have no deep understanding of the indentation modifiers (`%>`, `%<`), but I guess they should be balanced. Maybe someone with more knowledge about this could add an easy check that after each `Print...` the indentation level is the same as before?

### DIFF
--- a/src/vars.c
+++ b/src/vars.c
@@ -1055,7 +1055,7 @@ static void PrintElmMat(Expr expr)
     PrintExpr(READ_EXPR(expr, 0));
     Pr("%<[", 0, 0);
     PrintExpr(READ_EXPR(expr, 1));
-    Pr("%<, %<", 0, 0);
+    Pr("%<, %>", 0, 0);
     PrintExpr(READ_EXPR(expr, 2));
     Pr("%<]", 0, 0);
 }
@@ -1069,7 +1069,7 @@ static void PrintElmListLevel(Expr expr)
     Pr("%<[", 0, 0);
     PrintExpr(READ_EXPR(expr, 1));
     for (i = 2; i <= narg; i++) {
-        Pr("%<, %<", 0, 0);
+        Pr("%<, %>", 0, 0);
         PrintExpr(READ_EXPR(expr, i));
     }
     Pr("%<]", 0, 0);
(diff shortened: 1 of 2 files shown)


ChrisJefferson:
Yes, they should balance, good catch!

Just in case anyone (like me) wanted to check GAP master, it currently prints like this (which is definitely wrong):

```
gap> Print(foo);
function ( x )
    local a, b, c;
    a := x[1, 1];
  b := x[1, 1];
c := x[1, 1];
return 1;
end
gap>
```

ChrisJefferson (approved):

zickgraf:
I just noticed that this will break my tests because of differences between GAP 4.12 and GAP master. Is another point release of GAP 4.12 planned in the very near future? If not, I will probably use the function call notation `\[\,\]( ... )` to work around this.

ChrisJefferson:
My initial temptation is to say this should be part of 4.13, rather than 4.12.X -- as you say changing this could break some tests that rely on the old behaviour, which would be annoying for a point release.

Do you know if this bug was newly introduced in 4.12, or has it been there for a long time?

zickgraf:
> My initial temptation is to say this should be part of 4.13, rather than 4.12.X -- as you say changing this could break some tests that rely on the old behaviour, which would be annoying for a point release.

Makes sense. On the other hand, having a bug with such a simple fix for possibly some more years is also annoying :D But of course this is a policy decision.

> Do you know if this bug was newly introduced in 4.12, or has it been there for a long time?

It's been there since list access with two indices was introduced seven years ago in 91bac31f11bab0c407dc09beafdd0d4a28cfbbac.

fingolfin:
With the new release process and new website infrastructure, I don't think we'll have to wait "some more years" again. My plan is to release 4.12.2 in mid-December, and 4.13 could be out in April (see https://github.com/gap-system/gap/milestones for the current plan -- but these are not set in stone and also depend a bit on available time)

fingolfin:
And yeah, this is not a critical fix, and at the same time potentially breaking, so I don't think it's really appropriate for backporting to 4.12...

fingolfin (approved):
Great catch, thanks!

PR #5169, opened 2022-10-26
buildsys: better solution for src/compiled.h

fingolfin:
So far `make install` created a symlink from `$(includedir)/gap/src` to `$(includedir)/gap` to ensure compatibility with GAP packages that still `#include "src/compiled.h"`.

This is something various downstream packagers for Linux distros don't like, as it results in `$(includedir)/gap/src`, `$(includedir)/gap/src/src`, `$(includedir)/gap/src/src/src`, ... all being valid paths.

This patch implements a different solution, which we actually already use for our "compatibility mode" when doing out of tree builds: in `configure.ac` we have this code:
```
            AC_CONFIG_COMMANDS([src/compiled.h],
                [
                echo "#include \"$ac_abs_top_srcdir/src/compiled.h\"" > src/compiled.h
                ])
```


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 in commit 0dcb1f90d

PR #5170, opened 2022-10-26
kernel: fix compiler warning on some systems

fingolfin:
When using `clock_gettime` one must `#include <time.h>`, otherwise one gets compiler warnings if not errors. Observed while building GAP for Termux.


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 in commit 5b0a2386a

PR #5171, opened 2022-10-26
buildsys: fix out-of-tree cross compiling

fingolfin:
Cross compiling GAP releases works, as do out-of-tree builds, but the combination unfortunately was broken so far due to some insufficient checks in the code.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -120,17 +120,17 @@ ifeq ($(HPCGAP),yes)
 endif
 
 # allow overriding ffdata.c with a fixed version for cross compilation
-ifneq ("$(wildcard src/ffdata.c)","")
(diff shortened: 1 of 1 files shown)


ChrisJefferson (approved):

fingolfin:
In principle it would be good; in practice there are a bunch of obstacles: we'd have to test this as part of the `release` workflow (but this could done, we already have that test); but then we should also do a cross compilation I guess, which seems like a lot more work... Then again perhaps it would suffice to do a normal build and insert a check that certain files are *not* created, e.g. `build/ffdata.c`

fingolfin:
Backported to stable-4.12 in commit 43096b651

PR #5177, opened 2022-10-30
Emscripten

ChrisJefferson:
This PR switches `SyTime()` to use `NanosecondsSinceEpoch` in EMSCRIPTEN.

`SyTime` was previously broken, for two reasons:

1) Emscripten has a stack corruption bug when you call `getrusage` (but I don't plan on working around this, as it will get fixed soon).

2) Even after that is fixed, it just returns a fixed value at the moment.

Various bits of GAP expect `SyTime` to progress -- some standard library functions enter an infinite loop for example. We do have an option to avoid this problem (`ReproducableBehaviour`), but I thought making `SyTime` return something useful (even if it is wall time rather than CPU time) was the simplest and most useful fix.

This PR also adds the `tst` directory to the standard distributed files -- this makes it much bigger but also makes it easy to run the tests.

After this PR, GAP under emscripten passes all tests in testinstall except for those which involve temporary directory, or spawning processes.

### DIFF
--- a/src/gaptime.c
+++ b/src/gaptime.c
@@ -54,6 +54,15 @@
 */
 UInt SyTime(void)
 {
+#ifdef EMSCRIPTEN
+    // Emscripten's standard library always returns the same values
+    // for getrusage, which makes some GAP functions enter an
+    // infinite loop, as they use this function to try running
+    // for some period of time. Use NanosecondsSinceEpoch() as
+    // a substitute (it is not perfect, as NanosecondsSinceEpoch()
+    // is walltime, while RUSAGE_SELF is CPU time).
+    return SyNanosecondsSinceEpoch()/1000000000;
+#else
     struct rusage buf;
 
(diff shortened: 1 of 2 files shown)


src/gaptime.c
```diff
@@ -131,6 +110,37 @@ Int8 SyNanosecondsSinceEpoch(void)
 }
 
 
+/****************************************************************************
+**
+*F  SyTime()  . . . . . . . . . . . . . . . return time spent in milliseconds
```
fingolfin, line 115: Any reason to move this code? Note that `SyNanosecondsSinceEpoch` is declared in a header so it ought to be accessible without reordering the code (unless I miss something?)

ChrisJefferson: I just didn't think to check if it was in the header and moved it on instinct. I'll move it back.

fingolfin (approved):

PR #5180, opened 2022-11-01
Force `START_TEST` and `STOP_TEST` to be always used in tandem

zickgraf:
See commit messages for details. The third commit ("Force START_TEST and STOP_TEST to appear as a pair with the same name") changes the meaning of the first argument of `STOP_TEST`. If you do not like this (or it causes to many test failures), I will remove this commit.

## Text for release notes

TBD

### DIFF
--- a/lib/profile.g
+++ b/lib/profile.g
@@ -1086,7 +1086,7 @@ START_TEST := function( name )
     fi;
 end;
 
-STOP_TEST := function( name, args... )
+STOP_TEST_QUIET := function( name, args... )
     local time;
 
     if not IsBound( GAPInfo.TestData.START_NAME ) then
@@ -1101,12 +1101,22 @@ STOP_TEST := function( name, args... )
     fi;
     
     time:= Runtime() - GAPInfo.TestData.START_TIME;
-    Print( GAPInfo.TestData.START_NAME, "\n" );
-    Print( "msecs: ", time, "\n" );
     SetAssertionLevel( GAPInfo.TestData.AssertionLevel );
     SetInfoLevel( InfoPerformance, GAPInfo.TestData.InfoPerformanceLevel );
     Unbind( GAPInfo.TestData.AssertionLevel );
     Unbind( GAPInfo.TestData.START_TIME );
     Unbind( GAPInfo.TestData.START_NAME );
     Unbind( GAPInfo.TestData.InfoPerformanceLevel );
+    
+    return time;
+end;
+
+STOP_TEST := function( name, args... )
+    local time;
+    
+    time := STOP_TEST_QUIET( name );
+    
+    Print( name, "\n" );
+    Print( "msecs: ", time, "\n" );
+    
 end;
--- a/lib/profile.g
+++ b/lib/profile.g
@@ -1026,14 +1026,14 @@ end);
 
 #############################################################################
 ##
-#F  START_TEST( <id> )  . . . . . . . . . . . . . . . . . . . start test file
-#F  STOP_TEST( <file> )  . . . . . . . . . . . . . . . . . . . stop test file
+#F  START_TEST( <name> )  . . . . . . . . . . . . . . . . . . start test file
+#F  STOP_TEST( <name> )  . . . . . . . . . . . . . . . . . . . stop test file
 ##
 ##  <#GAPDoc Label="StartStopTest">
 ##  <ManSection>
 ##  <Heading>Starting and stopping test</Heading>
-##  <Func Name="START_TEST" Arg='id'/>
-##  <Func Name="STOP_TEST" Arg='file'/>
+##  <Func Name="START_TEST" Arg='name'/>
+##  <Func Name="STOP_TEST" Arg='name'/>
 ##
 ##  <Description>
 ##  <Ref Func="START_TEST"/> and <Ref Func="STOP_TEST"/> may be optionally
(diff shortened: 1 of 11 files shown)


zickgraf:
As expected, the CI is failing, e.g. due to the `STOP_TEST` in `tst/testinstall/grp/classic-PS.tst` which is called with "classic.tst-PS" instead of "classic-PS.tst". I can fix those problems but want to make sure you like the change before investing time.

fingolfin:
Even if you fix all those failures here, it is very likely that many similar failures will exist in packages. So it only makes sense to work on this aspect if you *also* plan to get all affected packages updated. In that case, the algorithm would be this:
1. fix the immediate issues on this PR (to make CI pass here)
2. afterwards I'll start a run of the `PackageDistro` CI using this PR, and we can collect the failing packages.
3. Then for each package a PR must be made to update it ...
4. ... and the maintainer(s) must be coaxed into releasing updates for their packages
5. We can re-do 2-4 as needed, until all passes again
6. finally merge this

lib/profile.g
```diff
@@ -1084,19 +1086,36 @@ START_TEST := function( name )
...
     Unbind( GAPInfo.TestData.AssertionLevel );
     Unbind( GAPInfo.TestData.START_TIME );
     Unbind( GAPInfo.TestData.START_NAME );
+    Unbind( GAPInfo.TestData.InfoPerformanceLevel );
+    
+    return time;
+end;
+
+STOP_TEST := function( name, args... )
+  local time;
```
fingolfin, line 1114:
```suggestion
    local time;
```

lib/test.gi
```diff
@@ -867,7 +867,7 @@ InstallGlobalFunction( "TestDirectory", function(arg)
   totalGcTime := 0;
   
   STOP_TEST_CPY := STOP_TEST;
-  STOP_TEST := function(arg) end;
+  STOP_TEST := function(arg) CallFuncList( STOP_TEST_QUIET, arg ); end;
```
fingolfin, line 870:
```suggestion
  STOP_TEST := STOP_TEST_QUIET;
```

ChrisJefferson:
Have you considered switching to an info message rather than `Error`? We already have a few places where we use `InfoWarning` at level 1 to print out problems with test files.

lib/test.gi
zickgraf: This is on purpose: the return value has to be dropped, otherwise it gets printed during the tests.

zickgraf:
> Have you considered switching to an info message rather than `Error`? We already have a few places where we use `InfoWarning` at level 1 to print out problems with test files.

Thanks for the suggestion, that sounds like a good idea. I would make it print at level 2 to not have to adapt all packages, would that be okay?

fingolfin:
@zickgraf sure

ChrisJefferson:
Easiest to start at level 2, then move to level 1 after packages with issues have been giving a reasonable chance to fix things.

 Personally (other reviewers may disagree!) I wouldn't mind if some packages printed an info message about inconsistent END_TEST message, as it wouldn't stop those package's tests from passing (unlike an Error, which leads to the tests appearing to fail).

zickgraf:
I have turned the errors into warnings and also fixed various test files calling STOP_TEST with a wrong identifier (or not at all).

> Personally (other reviewers may disagree!) I wouldn't mind if some packages printed an info message about inconsistent END_TEST message, as it wouldn't stop those package's tests from passing (unlike an Error, which leads to the tests appearing to fail).

The warning in START_TEST could make tests fail, so I chose info level 2 for now.

fingolfin (approved):

tst/testbugfix/2018-12-06-GroupWithGenerators.tst
```diff
@@ -1,5 +1,7 @@
+gap> START_TEST("2018-12-06-GroupWithGenerators.tst");
```
fingolfin, line 1: I don't understand: why is this needed?

zickgraf: Otherwise, it prints a warning due to InfoPerformance not being zero.
Previously, TestDirectory did not properly call `STOP_TEST` which is why
InfoPerformance was always 0 after the first `START_TEST`.

zickgraf:
Unexpectedly, in one test the actual output changed :o I will have to look into the details when I find the time again.

fingolfin:
Not a surprise: IIRC `START_TEST` resets the RNG state, and so my theory would be that by inserting one `START_TEST` you modified the RNG state for all subsequent tests. And it just so happens that there is one test file without its own `START_TEST` but which is affected by the RNG state... fix should be to add `START_TEST` to that file (to prevent similar problems in the future) and then adjust the output

zickgraf:
> Not a surprise: IIRC `START_TEST` resets the RNG state, and so my theory would be that by inserting one `START_TEST` you modified the RNG state for all subsequent tests. And it just so happens that there is one test file without its own `START_TEST` but which is affected by the RNG state... fix should be to add `START_TEST` to that file (to prevent similar problems in the future) and then adjust the output

Thanks for the hint, that indeed was the culprit! Now finally everything is green :-)

ChrisJefferson:
Just checking, are you now happy with this PR? I am, but just wanted to check before merging.

lib/test.gi
fingolfin: Then perhaps add a comment stating that?

fingolfin (approved):
Thank you! One minor suggestion, but fine anyway

PR #5183, opened 2022-11-02
sysinfo.gap: remove GAP_LIBS

fingolfin:
This was empty, and is not actually needed by any package build system. In fact only the Browse package even uses it, but it is not actually needed there as it uses `gac` anyway.

(Ping @ThomasBreuer)


ThomasBreuer (approved):
Concerning the Browse package, yes, this works also without `GAP_LIBS`.

PR #5187, opened 2022-11-04
Allow kernel extensions to use GAP header files via `#include <gap/HEADER.h>` instead of `#include "src/HEADER.h"` or `#include "HEADER.h"` (improves compatibility with "install" GAP versions shipped by Linux distributions)

fingolfin:
Right now, some GAP packages still use `#include "src/HEADER.h"` to
access GAP headers. Typically one sees `#include "src/compiled.h"` to
get "all" (well, most) GAP headers in one go.

That's not great for various reasons. For one, it is not obvious that
this is a GAP header (this is often dealt with by adding a comment). It
also doesn't mesh well with GAP installations made via `make install`,
which install headers into `PREFIX/include/gap`.

Indeed, we have a workaround in place specifically to deal with that,
which installs a file `PREFIX/include/gap/src/compiled.h` that includes
the real `compiled.h` to allow such packages to compile anyway. But
that's not quite enough as some packages directly include some other
headers (e.g. because `compiled.h` does not include them). To deal with
that better, this patch now installs a second set of the headers into a
`src` directory so that all those packages work again.

Moreover, `compiled.h` (or rather `gap_all.h`) is extended to cover two
more headers used by packages.

The longterm plan is to migrate packages away from this. Indeed, I
already changed many packages to at least use `#include "compiled.h"`
with the goal to eventually dispose of the `src` hack altogether.

Going one step further, the plan is that one day packages will be able
to use `#include <gap/HEADER.h>`. This PR enables this: for classic GAP
builds, it does so by creating a symlink `build/gap` pointing to `src`.
And for `make install` we modify the `GAP_CPPFLAGS` in the installed
`sysinfo.gap` suitably.

In addition, already today, packages can do `#include <gap_all.h>`,
provided one is willing to require at least GAP 4.11.1.


ChrisJefferson (approved):

fingolfin:
Backported to stable-4.12 in commit 183d2e50b

Issue #5181, opened 2022-11-02
problem with plist vectors


ThomasBreuer:
The following happens since GAP 4.12.0.

```
gap> G:= Group([ [ [ Z(5^6)^6944, 0*Z(5) ], [ 0*Z(5), Z(5^2)^20 ] ] ]);
Group([ [ [ Z(5^6)^6944, 0*Z(5) ], [ 0*Z(5), Z(5^2)^20 ] ] ])
gap> Size( G );
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `Enumerator' on 1 arguments at /.../gap/lib/methsel2.g:249 called from
Enumerator( C ) at /.../gap/lib/coll.gi:831 called from
...
```
The problem seems to be that `Enumerator` is called for the plist vector `C`, and that there is no method for it.

fingolfin:
According to `git bisect`, the regression was introduced by 1b134ae8df9ff165873efce21a42629aea252524 in PR #4983

ThomasBreuer:
I think that the idea of #4983 was correct: The comments say explicitly that `ImmutableMatrix` is not allowed to return a `IsMatrixObj` result if the input is a list of lists, and that `ImmutableVector` is not allowed to return a `IsVectorObj` result if the input is a list.
This works for the `IsMatrixObj` vs. `IsMatrix` case. These filters are intended to be disjoint.
However, `Is8BitVectorRep` and `IsGF2VectorRep` imply `IsVectorObj`:
```
gap> v:= [ 0, 1 ] * Z(5)^0;; ConvertToVectorRep( v );;
gap> Is8BitVectorRep( v );  IsList( v );  IsVectorObj( v );
true
true
true
```
Because of the `IsVectorObj` flag, the current `ImmutableVector` is allowed to return a `IsPlistVectorRep` result, which is not a list; and then we are in trouble because the result does not support `Enumerator` (in order to run over its entries), `IsBound\[\]`, and so on.

As a workaround, we could change the criterion for `ImmutableVector` such that `Vector` must not be called for an `IsList` input.
However, in the long run it might be unavoidable to take the same approach for vector objects as for matrix objects, that is, to define the filters `IsVectorObj` and `IsRowVector` such that they are disjoint, and to introduce `IsRowVectorOrVectorObj`.

fingolfin:
This was fixed by #5188

PR #5188, opened 2022-11-04
fixes for `ImmutableVector`

ThomasBreuer:
This addresses #5181.

- fixed `ImmutableVector` for `IsVectorObj`, it may be necessary to adjust the base domain
- workaround in `ImmutableVector` for `IsRowVector`: We are not allowed to return a non-list, thus we are not allowed to call `Vector` (This change is just intended to make the tests pass. In the long run, we need a better solution.)
- added tests for the two situations (plus an implicit test of the whole setup; this is the one that was broken in the Oscar test suite, which uncovered the problems)
- changed the documentation of `ImmutableMatrix` and `ImmutableVector`, in order to explain the differences w.r.t. to list-of-list matrices vs. matrix objects, and w.r.t. vectors that are lists vs. vector objects, respectively.

### DIFF
--- a/hpcgap/lib/vecmat.gi
+++ b/hpcgap/lib/vecmat.gi
@@ -1739,7 +1739,7 @@ end);
 
 #############################################################################
 ##
-#F  ImmutableVector( <field>, <vector>
+#F  ImmutableVector( <field>, <vector> )
 ##
 InstallMethod( ImmutableVector,"general,2",[IsObject,IsRowVector],0,
 function(f,v)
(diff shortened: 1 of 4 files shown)


fingolfin (approved):
Looks good to me. I suggest we merge this soon, unless @frankluebeck wants to have a look first ? (But even if it is gets merged now, we can of course still edit later...)

fingolfin:
Backported to stable-4.12 in commit 46e24a9bc

PR #5193, opened 2022-11-05
buildsys: link installed gap executable against libgap and set SYS_DEFAULT_PATHS

### LINKED ISSUE #5096: Having `gap` use `libgap` (at least the binary installed by `make install`); and set a good `SYS_DEFAULT_PATHS` default in the installed GAP

fingolfin:
When doing `make install`, install a version of GAP that links against libgap. That gap has `SYS_DEFAULT_PATHS` set to point to the GAP roots `$prefix/lib/gap` and `$prefix/share/gap`. Also adjust the installed sysinfo.gap accordingly.

Resolves #5096


fingolfin:
Backported to `stable-4.12` in afa8fbd9d2e6d2084463d63b69eaece9bf4dc98f plus 97b8a3f7ca11e41821c9c6e54d9b5ce950ab5137 plus 96d8bf0147793bb16dfeae116952efded2f859c9

dimpase:
> I don't think anyone is likely to review this, so I'll just merge it.

well, why,  we're always happy to review stuff related to using GAP in Sage...

PR #5194, opened 2022-11-07
Support building packages which just have a Makefile

ChrisJefferson:
I had not noticed that `BuildPackages.sh` doesn't support packages which do not have a `./configure` file, only a `Makefile`. `Vole` is such a package.

This does remove the message `No building required for $PKG`, as it's less obvious where to put it now (I can't just put an `else` after the `Makefile` check, because then the message would be printed whenever the configure script failed to generate a Makefile).

If anyone really wants that message I'm happy to work on reintroducing it, but I thought it might be more useful to be a little quieter.

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -207,9 +207,11 @@ GAPInput
       echo_run "$MAKE" clean
       echo_run ./configure "$GAPROOT"
     fi
+  fi
+
+  if [[ -f makefile || -f Makefile || -f GNUmakefile ]]
+  then
     echo_run "$MAKE"
-  else
-    notice "No building required for $PKG"
   fi
 }
 


bin/BuildPackages.sh
```diff
@@ -207,9 +207,11 @@ GAPInput
       echo_run "$MAKE" clean
       echo_run ./configure "$GAPROOT"
     fi
+  fi
+
+  if [[ -f makefile || -f Makefile || -f GNUmakefile ]]
```
fingolfin, line 212:
```suggestion
  if [[ -f makefile -o -f Makefile -o -f GNUmakefile ]]
```

ChrisJefferson:
This is failing because it turns out there are already packages with a makefile, but no configure (`io_fromhomalg` is the first one we hit), which don't work in CI (`io_fromhomalg` assumes `gap` is in the path)

So this can't be merged straight away, but we should probably decide if we should be running all Makefiles (in which case `io_fromhomalg` needs fixing), or we shouldn't be (in which case I can add an empty `./configure` to vole).

ChrisJefferson:
The following packages (on my machine) have a makefile, but no configure script:

```
4ti2interface
OrbitalGraphs
Vole
autodoc
cap
examplesforhomalg
gaussforhomalg
generalizedmorphismsforcap
gradedmodules
gradedringforhomalg
homalg
homalgtocas
io_forhomalg
linearalgebraforcap
localizeringforhomalg
matricesforhomalg
modulepresentationsforcap
modules
monoidalcategories
nconvex
newss
ringsforhomalg
sco
toolsforhomalg
toricvarieties
```

ChrisJefferson:
Many of those have a makefile which on default target does two things -- run tests (which we probably don't want), and build docs (which we do want).

fingolfin:
Yeah, I don't think we can do this in general. For now, I recommend adding an empty `configure` script to your package.

What we really want is to finally tackle something we have talked about for a long time: properly specify and document how package build systems must be set up so that GAP can correctly use them automatically. One way would be to just document the status quo ("must have a configure script taking these and those arguments" etc.). 

But we could also try for something a bit more flexible and e.g. specify some new fields in `PackageInfo.g` which specify how a package can be built (as metadata, or even as code -- the latter is quite flexible but has the drawback of allowing no introspection which on the long run could be painful on its own if not done right...)

This is also relevant when it comes to eventually adding `make install` support to packages: for any without compiled code, this could be as simple as "copy all files to a certain target destination". But for packages with compiled code (hence architecture dependant), this might be a bit more complicated. E.g. we don't want to install `.o` files or `.libs` directories etc.; and we may wish to install all the non-arch dependant parts (GAP code, documentation, ...) into `PREFIX/share/gap/pkg/MYPACKAGE` while binaries (which are arch dependant) perhaps should go into `PREFIX/lib/gap/pkg/MYPACKAGE`. Of course we currently have no precedent for a package being split across two different directories like that... but being able to support that is one of several ulterior motives I had for adding `LoadKernelExtension` ....  

All of this is of course also relevant for package distros and e.g. @dimpase recently asked questions about it.

This PR is not the place to discuss this further, someone should open one or multiple issues for this. I just wanted to get it out of my system for now.

fingolfin:
So what is your plan with this PR? I currently see no short-term way forward with it, and long-term we should go for something better anyway. Close it?

Oh and BTW, actually the `configure` script in `vole` could do something useful: right now you hardcode the path to GAP in your `Makefile`, in the `doc` target... with a proper `configure` script taking a GAP path, that could be fixed

PR #5201, opened 2022-11-14
Don't strip binaries when installing them

fingolfin:
The stripping causes issues in various situations. Some distros
also disable it for that reason. It is also easy to strip the
libraries manually later on if so desired.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -582,10 +582,10 @@ install: install-bin install-doc install-gaproot install-sysinfo install-headers
 install-bin: build/gap-install install-libgap
 	# install a special build of gap with SYS_DEFAULT_PATHS set suitably
 	$(INSTALL) -d -m 0755 $(DESTDIR)$(bindir)
-	$(LTINSTALL) -s build/gap-install $(DESTDIR)$(bindir)/gap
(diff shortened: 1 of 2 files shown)


ChrisJefferson (approved):

fingolfin:
This fails right now because our tests for `make install` verify that the build path do not show up in any installed files, which is broken if we don't strip (due to debug data referencing it).

I could adjust the test, but I am really torn on what the "right" thing here is. Perhaps downstream packagers have feedback on this? I.e., what do you prefer `make install` should do:
1. strip installed `gap` and `libgap`
2. don't strip it
3. don't care
4. something else

CC @jamesjer @isuruf @dimpase @collares @antonio-rojas @yurivict

antonio-rojas:
Please don't strip on make install, as it breaks debug package generation. Distros usually have their own tooling to strip binaries at packaging time (after debug symbols have been extracted).

yurivict:
No.

Both GNU tools and cmake are able to install either stripped or unstripped binaries based on the value of the ```STRIP``` environment variable.

fingolfin:
Backported to `stable-4.12` in bbfbd2e4bd07f099f2d1966d36a45b4cb00fcc32 and 69c13035e9920728c0815c4599f85511beeb51cd

PR #5206, opened 2022-11-15
buildsys: install-bin depends on install-libgap

fingolfin:


fingolfin:
Backported to `stable-4.12` in 19b2cb0d55c2a932a7fc2c843364827c44c61df3

Issue #5209, opened 2022-11-19
Show diffs in `Test` immediately, not at the end


fingolfin:
We recently wasted some time tracing down the problem in a `.tst` file; the problem could be condensed to this example
```
gap> a:=SymmetricGroup(1000);;
gap> # ... do some tests
gap> a:=SymmetricGroup(1 0);;
gap> Length(Elements(a)) = Factorial(10);
true
```
Of course this has a typo. But we keep running past it, and end up in an essentially infinite loop. Reading this `.tst` file just looks like this
```
gap> Test("hang.tst");
# line 3 of 3 (100%)
```
and then it stays there forever / until you press ctrl-c.

In contrast, if evaluated "manually" in the REPL, we see:
```
gap> a:=SymmetricGroup(1000);;
gap> # ... do some tests
gap> a:=SymmetricGroup(1 0);;
Syntax error: ) expected
a:=SymmetricGroup(1 0);;
                    ^
gap> Length(Elements(a)) = Factorial(10);
```
If I remove the final two lines and use `Test`:
```
gap> Test("hang.tst");
########> Diff in hang.tst:3
# Input is:
a:=SymmetricGroup(1 0);
# Expected output:
# But found:
Syntax error: ) expected in stream:1
a:=SymmetricGroup(1 0);
                    ^
########
false
```

So, can't we show that diff *immediately*? That would have helped a lot in tracking this down.

fingolfin:
This was resolved by @ChrisJefferson in PR #5210

PR #5210, opened 2022-11-19
Change 'Test' so it immediately outputs failed test results

ChrisJefferson:
Decided to quickly look to see how hard it would be to implement #5209 . Turns out not too hard!

This should probably have some more tests (which would be added to testspecial I think, like the single existing one).

### DIFF
--- a/tst/testspecial/test-1.tst
+++ b/tst/testspecial/test-1.tst
@@ -1,0 +1,5 @@
+gap> a:=SymmetricGroup(5);;
+gap> # ... do some tests
+gap> a:=SymmetricGroup(1 0);;
+gap> Length(Elements(a)) = Factorial(10);
+true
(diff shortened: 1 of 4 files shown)


tst/testspecial/test-1.tst
```diff
@@ -0,0 +1,5 @@
+gap> a:=SymmetricGroup(5);;
+gap> # ... do some tests
+gap> a:=SymmetricGroup(1 0);;
+gap> Length(Elements(a)) = Factorial(10);
+true
```
fingolfin, line 5:
```suggestion
true

```

tst/testspecial/testing-test.g.out
```diff
@@ -0,0 +1,20 @@
...
+########> Diff in test-1.tst:4
+# Input is:
+Length(Elements(a)) = Factorial(10);
+# Expected output:
+true
+# But found:
+false
+########
+false
+gap> QUIT;
```
fingolfin, line 20: So how does the output of this test differ from what one gets without your `Test` patch?

ChrisJefferson: My hope is the output is identical, it just comes out faster

fingolfin (approved):
Works great for me!

PR #5211, opened 2022-11-22
fix the help system

ThomasBreuer:
If the Browse package is not loaded and if the user preference `"Pager"` of GAP has the value `"builtin"` and if a help request yields more than one match then the entries of the list shown by the help request are not cut at `SizeScreen()[1]` columns up to now.

For example, `?Random` yields several entries that need more than 80 columns to be shown.
Since the pager prints `SizeScreen()[2]` lines, the first entries are not visible.

The current proposal fixes this.
Note that the GAPDoc function `InitialSubstringUTF8String` does not take care about escape sequences.

Perhaps it would be more appropriate to add the new auxiliary function `InitialSubstringUTF8Text` to the GAPDoc package.

(Besides this fix, the `\n` character at the end of the first line in the pager has been removed, such that this line becomes visible.)

### DIFF
--- a/lib/helpbase.gi
+++ b/lib/helpbase.gi
@@ -1013,6 +1013,27 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
   return [exact, match];
 end);
 
+
+#############################################################################
+##
+#F  InitialSubstringUTF8Text( <str>, <cols> )
+##
+##  This is a utility that extends the GAPDoc function
+##  <C>InitialSubstringUTF8String</C>, which deals with strings containing
+##  unicode characters but does not deal with escape sequences, i.e.,
+##  sequences starting with ESC and stopping with the first letter
+##  afterwards).
+##  Note that the text version of GAPDoc manuals contains both
+##  unicde characters and escape sequences.
+##
+##  <Ref Func="InitialSubstringUTF8Text"/> returns a string that is the
+##  longest prefix of the string <A>str</A> that has visible/printed length
+##  at most <A>cols</A> and contains all escape sequences from <C>str</C>.
+##
+
+# The following global variables will be defined via the GAPDoc package.
+# We assign them here (and unbind them later on) in order to avoid syntax
+# warnings.
 if not IsBound( InitialSubstringUTF8String ) then
   InitialSubstringUTF8String:= "dummy";
 fi;
@@ -1026,10 +1047,6 @@ fi;
 BindGlobal( "InitialSubstringUTF8Text", function( str, cols )
     local esc, parts, len, res, j, pos, word, w;
 
-    # Collect printable parts up to visible length 'cols',
-    # and all "escape sequences" in 'str'
-    # (sequences starting with ESC and stopping with the first letter
-    # afterwards).
     esc:= CHAR_INT(27);
     parts:= [];
     len:= Length( str );
@@ -1052,11 +1069,10 @@ BindGlobal( "InitialSubstringUTF8Text", function( str, cols )
       if len < pos then
         break;
       fi;
-      j:= pos + 1;
-      while j <= len and not str[j] in LETTERS do
-        j:= j+1;
-      od;
-      if j > len then
+      # Now pos points at an ESC character; all escape sequences we
+      # support are terminated by a letter, so search for one.
+      j:= PositionProperty( str, c -> c in LETTERS, pos );
+      if j = fail then
         Error( "string end inside escape sequence" );
       fi;
       Append( res, str{ [ pos .. j ] } );
--- a/lib/helpbase.gi
+++ b/lib/helpbase.gi
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
   return [exact, match];
 end);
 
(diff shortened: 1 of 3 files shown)


lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
+  InitialSubstringUTF8String:= "dummy";
+fi;
+if not IsBound( LETTERS ) then
+  LETTERS:= "dummy";
+fi;
+if not IsBound( WidthUTF8String ) then
+  WidthUTF8String:= "dummy";
+fi;
+
+BindGlobal( "InitialSubstringUTF8Text", function( str, cols )
```
fingolfin, line 1026: This is a bit confusing: there is function `InitialSubstringUTF8String` (from GAPDoc?) with an extremely similar name. What's the difference? Perhaps a code comment could explain that?

lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
 
+if not IsBound( InitialSubstringUTF8String ) then
+  InitialSubstringUTF8String:= "dummy";
+fi;
+if not IsBound( LETTERS ) then
+  LETTERS:= "dummy";
+fi;
+if not IsBound( WidthUTF8String ) then
+  WidthUTF8String:= "dummy";
+fi;
```
fingolfin, line 1024: I assume all these globals come from GAPDoc, and this code is here to ensure this file loads (without syntax warnings) regardless of whether GAPDoc is loaded first or not? Perhaps a code comment could be added to clarify this?

lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
+
+BindGlobal( "InitialSubstringUTF8Text", function( str, cols )
+    local esc, parts, len, res, j, pos, word, w;
+
+    # Collect printable parts up to visible length 'cols',
+    # and all "escape sequences" in 'str'
+    # (sequences starting with ESC and stopping with the first letter
+    # afterwards).
```
fingolfin, line 1032: Is this a full description of what this function does? Then perhaps it could be moved to be just before the function, then this would be clearer, IMHO

lib/helpbase.gi
fingolfin: Looking at the code, my guess is that `InitialSubstringUTF8String` takes an initial segment of `str` which fits into `cols` column in a terminal (taking multi byte encoded characters into account, perhaps "wide" Unicode characters or invisible ones like accent decorators) ?

And this function here "just" adds in handling of escape sequences? If so, perhaps the function name could reflect this better? InitialSubstringUTF8StringIgnoringEscapeSequences` rolls right off my tongue 😂 😝

lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
+      if w <= cols then
+        Append( res, word );
+        cols:= cols - w;
+      elif cols > 0 then
+        Append( res, InitialSubstringUTF8String( word, cols ) );
+        cols:= 0;
+      fi;
+      if len < pos then
+        break;
+      fi;
```
fingolfin, line 1054:
```suggestion
      fi;
      # now pos points at an ESC character; all escape sequences we
      # support are terminated by a letter, so search for one
```

lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
+      j:= pos + 1;
+      while j <= len and not str[j] in LETTERS do
+        j:= j+1;
+      od;
+      if j > len then
```
fingolfin, line 1059: This is very subjective, but personally I'd find this easier to quickly grasp if it used `PositionProperty`, to match the first search
```suggestion
      j:= PositionProperty(str, c -> c in LETTERS, pos);
      if j = fail then
```

lib/helpbase.gi
```diff
@@ -1013,13 +1013,74 @@ InstallGlobalFunction(HELP_GET_MATCHES, function( books, topic, frombegin )
...
+        Error( "string end inside escape sequence" );
+      fi;
+      Append( res, str{ [ pos .. j ] } );
+    od;
+    return res;
```
fingolfin, line 1064: Perhaps a few simple test cases could be added?

lib/helpbase.gi
```diff
@@ -1050,16 +1111,18 @@ InstallGlobalFunction(HELP_SHOW_MATCHES, function( books, topic, frombegin )
   # more than one topic found, show overview in pager
   else
     lines := 
-        ["Help: several entries match this topic - type ?2 to get match [2]\n"];
+        ["Help: several entries match this topic - type ?2 to get match [2]"];
```
fingolfin, line 1114: Good catch

fingolfin (commented):
Thank you, makes sense, and a great catch

fingolfin (approved):

PR #5213, opened 2022-11-22
kernel: workaround GCC 12 warning

fingolfin:
With GCC 12, we get a spurious warning indicating that `buf` in `GetStr` may be used uninitialized. Weirdly, it does not complain about the almost identical code in e.g. `GetPragma`.


ChrisJefferson (approved):

ChrisJefferson:
It's better this way anyway, I wish C just default initalised all stack variables by default

fingolfin:
Backported to `stable-4.12` in 246384957e222184610a5d573770340207296dfb

PR #5196, opened 2022-11-07
Always enable pthread flags

fingolfin:
Resolves #5192

PR #5214, opened 2022-11-22
Automatically find GMP, zlib, GNU readline when installed by Homebrew, MacPorts or Fink

### LINKED ISSUE #1945: Automatically search for GMP and GNU Readline in a few additional places, if we cannot find them elsewhere

fingolfin:
This way mac users with Homebrew, Fink or MacPorts will automatically benefit from GMP/zlib/GNU readline installed via these package managers.

Resolves #1945.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -235,7 +235,7 @@ description of each is also available via
 GMP
 ---
 
-GAP 4 uses the external library GMP (see <https://www.gmplib.org>) for large
+GAP 4 uses the external library GMP (see <https://gmplib.org>) for large
 integer arithmetic, replacing the built-in code used in previous versions
 and achieving a significant speed-up in related computations. There is a
 version of GMP included with the GAP archive you downloaded and this will
@@ -673,31 +673,12 @@ to enter an administrator password).
 
      xcode-select --install
 
-You might want to consider using GNU readline by installing it via:
+If you are using a macOS package manager, we recommend installing a few for
+faster compilation and a better user experience:
 
  * using Homebrew: `brew install gmp readline`
  * using Fink: `fink install gmp readline7`
  * using MacPorts: `port install gmp readline`
- 
-After that you have to compile GAP and tell it where to find your installation of GNU readline
-using the following commands.
-
-For Homebrew, use these commands:
-
-    ./configure --with-gmp=$(brew --prefix) --with-readline=$(brew --prefix)/opt/readline
-    make
-
-For Fink, use these commands:
-
-    ./configure CPPFLAGS=-I/sw/include LDFLAGS=-L/sw/lib
-    make
-
-For MacPorts, use these commands:
-
-    ./configure CPPFLAGS=-I/opt/local/include LDFLAGS=-L/opt/local/lib
-    make
-
-For further information on how to use GNU readline, refer to section 5 above.
 
 Now simply follow the Unix installation instructions to compile and start
 GAP and then it will run in this Terminal window.
--- a/configure.ac
+++ b/configure.ac
@@ -521,6 +521,13 @@ dnl
 dnl External dependencies
 dnl
 
+# We support finding dependencies in certain non-standard locations:
+#  Homebrew on macOS: $(brew --prefix) (usually /opt/homebrew or /usr/local)
+#  Fink on macOS: /sw
+#  MacPorts: /opt/local
+HOMEBREW_PREFIX="$(brew --prefix 2>/dev/null || :)"
+DEFAULT_SEARCH_PATH="$HOMEBREW_PREFIX /opt/local /sw"
+
 dnl Find GMP
 AC_ARG_WITH([gmp],
   [AS_HELP_STRING([--with-gmp@<:@=builtin|PREFIX@:>@],
@@ -528,6 +535,7 @@ AC_ARG_WITH([gmp],
     [],[with_gmp=yes])
 
 BUILD_GMP=no
+GMP_SEARCH_PATH="DEFAULTS ${DEFAULT_SEARCH_PATH}"
 GMP_CPPFLAGS=
 GMP_LDFLAGS=
 GMP_LIBS="-lgmp"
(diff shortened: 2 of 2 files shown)


fingolfin:
I've tested this on macOS with Homebrew and on Linux. @dimpase you provided the original OpenBSD compatibility patch for the readline check. If you have a way to test the new code on OpenBSD (i.e. whether it still detected `libreadline` (or rather `libereadline` there), that'd be great.

dimpase:
@fingolfin - I will test on OpenBSD and report here.

fingolfin:
@dimpase thank you!

@ChrisJefferson it keeps failing the `testexpect testmockpkg testspecial test-compile testlibgap testkernel - - ubuntu-latest` test, due to a timeout?!? This is totally baffling to me. Any idea what might cause this?

dimpase:
> @fingolfin - I will test on OpenBSD and report here.

OK, it works just fine. I didn't try installing packages needing building GAP kernel extensions, should I?

fingolfin:
@dimpase I don't see a reason why kernel extensions should be affected by this. But of course it never hurts to check, except for the effort it requires... so if it is "easy" for you, please do; but if it is inconvenient, don't worry about it (GAP doesn't seem to be in the OpenBSD ports system either, so I guess there are not many GAP users there to start with)

dimpase:
@fingolfin I just tried `semigroups` package, and got 
```
...
make[2]: Entering directory '/home/dima/tmp/gap/pkg/semigroups'
  CXX      src/semigroups_la-bipart.lo
  CXX      src/semigroups_la-cong.lo
LLVM ERROR: out of memory
Allocation failed
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: c++ -DHAVE_CONFIG_H -I/home/dima/tmp/gap/build -I/home/dima/tmp/gap/src -I/home/dima/tmp/gap -DUSE_GASMAN=1 -I/usr/local/include -I./bin/include -I./bin/include/libsemigroups -I./gapbind14/include -DHPCOMBI_CONSTEXPR_FUN_ARGS -I./libsemigroups/extern -I./libsemigroups/extern/HPCombi/include -I./libsemigroups/extern/HPCombi/include/fallback -I./libsemigroups/extern/fmt-8.0.1/include -I./libsemigroups/include -DFMT_HEADER_ONLY -DNDEBUG -I/usr/local/include -I/usr/local/include -I./bin/include -I./bin/include/libsemigroups -std=gnu++14 -O3 -flax-vector-conversions -g -O2 -MT src/semigroups_la-cong.lo -MD -MP -MF src/.deps/semigroups_la-cong.Tpo -c src/cong.cpp -fPIC -DPIC -o src/.libs/semigroups_la-cong.o
1.      <eof> parser at end of file
2.      Code generation
c++: error: clang frontend command failed with exit code 134 (use -v to see invocation)
OpenBSD clang version 13.0.0
Target: amd64-unknown-openbsd7.1
Thread model: posix
InstalledDir: /usr/bin
c++: note: diagnostic msg:
********************
...
```

dimpase:
cf. https://github.com/llvm/llvm-project/issues/59194

fingolfin:
@dimpase thank you for testing! I am not concerned about that semigroups failure; as you pointed out, it's a compiler bug; but even if not, it'd be a problem specific to that package, and to be dealt with by the package authors/maintainers, not us

dimpase:
> as you pointed out, it's a compiler bug; but even if not, it'd be a problem specific to that package

it was merely lack of my knowledge of OpenBSD `datalimit` feature - this compilation needs many gigs of RAM to finish successfully, and it was limited too much.

dimpase:
OK, at least some packages with kernel extenstions work on OpenBSD, e.g. `io` works. `Semigroups` and `digraphs` build, but don't work, as `.so` file of `digraphs` is corrupt. I'll open an issue there.

dimpase (approved):
LGTM

dimpase:
> @ChrisJefferson it keeps failing the `testexpect testmockpkg testspecial test-compile testlibgap testkernel - - ubuntu-latest` test, due to a timeout?!? This is totally baffling to me. Any idea what might cause this?

is this the same problem as in  https://github.com/gap-system/gap/pull/5196#issuecomment-1305740328 ?

fingolfin:
@dimpase aha, interesting observation! I have no idea, but they sure look similar, and in both cases `configure.ac` was modified... I still don't understand how that can lead to a timeout in that test, but it might give a clue...?

dimpase:
> @dimpase aha, interesting observation! I have no idea, but they sure look similar, and in both cases `configure.ac` was modified... I still don't understand how that can lead to a timeout in that test, but it might give a clue...?

We use a lot of `expect` and `pexpect` in Sage, and it's in general dodgy. Various OS distros tend to break them once in a while. :-(

dimpase:
> GAP doesn't seem to be in the OpenBSD ports system either

Creating an OpenBSD port is a nontrivial effort, especially if it involves things like extra packages, extra dynamic libraries, etc. Once @kaashif made a draft port of GAP, but didn't complete it. (The `ereadline` PR I did was meant to help it).

james-d-mitchell:
> @dimpase thank you for testing! I am not concerned about that semigroups failure; as you pointed out, it's a compiler bug; but even if not, it'd be a problem specific to that package, and to be dealt with by the package authors/maintainers, not us

I don't really agree with this I'm afraid @fingolfin, if this PR breaks the Semigroups package, or any other package, then shouldn't that be investigated and resolved **before** this PR is merged? Just thinking about recent similar issues with `MatrixObj` that caused a huge amount of work...

dimpase:
> > @dimpase thank you for testing! I am not concerned about that semigroups failure; as you pointed out, it's a compiler bug; but even if not, it'd be a problem specific to that package, and to be dealt with by the package authors/maintainers, not us
> 
> I don't really agree with this I'm afraid @fingolfin, if this PR breaks the Semigroups package, or any other package, then shouldn't that be investigated and resolved **before** this PR is merged? Just thinking about recent similar issues with `MatrixObj` that caused a huge amount of work...

For the record, this PR does not break Semigroups (same error on OpenBSD with GAP 4.12.1 and Semigroups there)

james-d-mitchell:
Thanks @dimpase, my earlier comment is now obsolete!

fingolfin:
@james-d-mitchell yeah I think we actually agree in the end then: my comment was based on the assumption that this PR is *not* the cause for the semigroups failure on OpenBSD. Of course if this PR *did* cause it, I would want to resolve it.

fingolfin:
Found the problem: I had forgotten to copy the code defining `HAVE_LIBREADLINE`. As a result, GAP was not using GNU readline, which broke the expect test.

dimpase:
Would it work in a Conda environment ? There locations of /usr, etc, are not fixed in advance.

We have in Sage a bunch of autoconf macros (for the libraries you deal with here, too) which allow such a scenario.

fingolfin:
I have no idea, I've never used Conda beyond a few short experiments several years ago

dimpase:
One does not have to go through a list of prefixes, such as `/sw, /opt/foo`, etc, as all the sane toolchains know where to look for headers and libraries.
This is how one can support more or less arbitrary locations.

fingolfin:
@dimpase I don't really understand what you are trying to say with your last comment?

In the meantime this PR addresses a real world paper cut issue that I have seen multiple people struggling with, often without them even knowing this (eg waiting longer for GMP to compile, not benefiting from GNU readline)

dimpase:
I was misled by the PR description, and thought (without looking at the code) that you loop over a fixed list of prefixes in search for libraries. But you actually don't do this, right?

fingolfin:
We do this, among other things there is no alternative that works w/o expert user knowledge. 

What it does now is this: if the user specified an explicit prefix for, say, GMP, use that. Otherwise, try to see if it "just works" (e.g. because it is in /usr, or because someone/something set CPPFLAGS etc. appropriately). Otherwise, fall back to checking if Homebrew, MacPorts or Fink provide it.

PR #5217, opened 2022-11-24
WIP: positional matrix obj rep

fingolfin:
The plan for this PR is to provide two incomplete representations for vector obj and matrix obj, on which many other implementations (such as the existing plist matrix obj) can be based; this is meant to reduce work for new implementation, reduce code duplication, and provide some uniformity.

Part of the work is to take constants like `BDPOS` and give them a better name. While doing this, I tried to reorder them, but discovered this isn't easy because their values are *implicitly* used in a `Objectify` calls (at least 70 are relevant for this). Since they also represent further code duplication, I plan to put them into a few helper functions, to reduce the number to 4 or so (one each for matrix vs. vector, and for plist rep vs. zmodnz rep). But then @ThomasBreuer pointed out the extra overhead this indirection causes... While I am not overly worried about this (in the worst case I can rewrite these as kernel functions), that made us think about the general impact of objection creation overhead for matrix obj vs. old-style matrices... So my work plan for this PR right now is this:

1. benchmark matrix and vector creation times; add the code for this into the repo somewhere; then act on the result. The idea is to compare at least these:
   - old style
   - plistrep
   - compressed
   - other??

2. factor out `Objectify` calls

3. really switch to the new constants (possibly not reordered for now)

### DIFF
--- a/lib/matobj/positional.gd
+++ b/lib/matobj/positional.gd
@@ -1,0 +1,49 @@
+#############################################################################
+##
+##  This file is part of GAP, a system for computational discrete algebra.
+##
+##  SPDX-License-Identifier: GPL-2.0-or-later
+##
+##  Copyright of GAP belongs to its developers, whose names are too numerous
+##  to list here. Please refer to the COPYRIGHT file for details.
+##
+
+# TODO: document this
+DeclareRepresentation( "IsPositionalVectorRep",
+        IsVectorObj and IsPositionalObjectRep
+    and IsNoImmediateMethodsObject
+    and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
+    [] );
+
+# TODO: document this
+DeclareRepresentation( "IsPositionalMatrixRep",
+        IsMatrixObj and IsPositionalObjectRep
+    and IsNoImmediateMethodsObject
+    and HasNumberRows and HasNumberColumns
+    and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
+    [] );
+
+
+#
+# Some constants for matrix resp. vector access
+#
+# TODO: For now the order follows the order of the predecessors:
+# BDPOS = 1, RLPOS = 3, ROWSPOS = 4; the goal is to
+# eventually change this. But this needs us to carefully revisit
+# all Objectify calls
+
+# Position of the base domain
+BindConstant( "MAT_BD_POS", 1 );
+# Position of the number of rows
+BindConstant( "MAT_NROWS_POS", 5 );  # FIXME: in many cases superfluous (can be computed from NCOLS and DATA)
+# Position of the number of columns
+BindConstant( "MAT_NCOLS_POS", 3 );
+# Position of the data
+BindConstant( "MAT_DATA_POS", 4 );
+
+# Position of the base domain
+BindConstant( "VEC_BD_POS", 1 );
+# Position of the data
+BindConstant( "VEC_DATA_POS", 2 );
+# Position of the length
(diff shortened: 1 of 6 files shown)


lib/matobj/positional.gi
```diff
@@ -0,0 +1,88 @@
...
+
+
+InstallMethod( ShallowCopy, [ IsPositionalVectorRep ],
+  function( v )
+    local i, res;
+    res := List([1..LEN_POSOBJ(v)], i -> v![i]);
+    res![VEC_DATA_POS] := ShallowCopy(v![VEC_DATA_POS]);
+    res := Objectify(TypeObj(v), res);
+
+    # 'ShallowCopy' MUST return a mutable object if such an object exists at all!
```
ThomasBreuer, line 35: My interpretation of `ShallowCopy` is that it returns its argument if this is not copyable.
We could restrict the method installation to `IsPositionalVectorRep and IsCopyable`.
And perhaps we should add an explicit reference to `IsCopyable` to the documentation of `ShallowCopy`.

lib/matobj/positional.gi
```diff
@@ -0,0 +1,88 @@
...
+    if not IsMutable(v) then
+      SetFilterObj(res, IsMutable);
+    fi;
+    return res;
+  end );
+
+# StructuralCopy works automatically
```
ThomasBreuer, line 42: `StructuralCopy` is not an operation, so we cannot do anything for it here.

lib/matobj/positional.gi
```diff
@@ -0,0 +1,88 @@
...
+  end );
+
+InstallMethod( ShallowCopy, [ IsPositionalMatrixRep ],
+  function( m )
+    local res;
+    res := List([1..LEN_POSOBJ(m)], i -> m![i]);
+    res![MAT_DATA_POS] := ShallowCopy(m![MAT_DATA_POS]);
+    res := Objectify(TypeObj(m), res);
+
+    # 'ShallowCopy' MUST return a mutable object if such an object exists at all!
```
ThomasBreuer, line 78: see the comment about `ShallowCopy` for `IsPositionalVectorRep`

lib/matobj/positional.gd
```diff
@@ -0,0 +1,49 @@
...
+##
+##  SPDX-License-Identifier: GPL-2.0-or-later
+##
+##  Copyright of GAP belongs to its developers, whose names are too numerous
+##  to list here. Please refer to the COPYRIGHT file for details.
+##
+
+# TODO: document this
+DeclareRepresentation( "IsPositionalVectorRep",
+        IsVectorObj and IsPositionalObjectRep
```
ThomasBreuer, line 13: `and IsCopyable` if we adopt the idea to make this explicit

lib/matobj/positional.gd
```diff
@@ -0,0 +1,49 @@
...
+    and IsNoImmediateMethodsObject
+    and HasBaseDomain and HasOneOfBaseDomain and HasZeroOfBaseDomain,
+    [] );
+
+# TODO: document this
+DeclareRepresentation( "IsPositionalMatrixRep",
+        IsMatrixObj and IsPositionalObjectRep
```
ThomasBreuer, line 20: and `IsCopyable` if we adopt the idea to make this explicit

ThomasBreuer (approved):
The idea to provide a generic implementation of vector and matrix objects via positional objects is nice.
I have inserted a few comments.

Eventually the new representations should become documented, and then the files `lib/matobjnz.g*` can be changed in the same way as `lib/matobjplist.g*`.
(Doing this now would create conflicts with pull request #5254.)

PR #5224, opened 2022-11-25
Use `jl_reinit_foreign_type` if available

fingolfin:
This is part of the plan outlined in https://github.com/oscar-system/GAP.jl/issues/846


ThomasBreuer (approved):
As far as I see, this looks good. (@ChrisJefferson will be able to give detailed comments.)

fingolfin:
Backported to `stable-4.12` in fccdb8aee1106889a04d65d66b8c95cc8c6066dd

PR #5225, opened 2022-11-25
Enable `IsomorphismFpGroupByChiefSeries` for all groups

fingolfin:
This allows using `ConfluentMonoidPresentationForGroup` on e.g. fp groups


hulpke (approved):
I would add at least a filter "IsFinite" in the installation.

hulpke:
> This allows using `ConfluentMonoidPresentationForGroup` on e.g. fp groups

Just the generic warning that, if it works for fp groups, it is of course fine, but I would not recommend to users to represent known finite groups as an fp group.

Issue #5232, opened 2022-12-01
Cannot load kernel extensions built against non-install GAP into an executable built against `libgap`


dimpase:
`libgap` is broken in the current master and in 4.12.1 (but still OK on 4.11.1) on macOS 13.0.1 if `io` is built, as can be seen by running
`tst/testlibgap/basic`: (which is built by `make testlibgap`, for instance)
```
% ./tst/testlibgap/basic -l .
# Initializing GAP...
 ┌───────┐   GAP 4.13dev-169-g28d5faa built on 2022-11-30 18:44:58+0000
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-apple-darwin22-default64-kv8
 Configuration:  gmp 6.2.1, GASMAN, readline
 Loading the library and packages ...
Error, LOAD_DYN: failed to load kernel module ./pkg/io//bin/x86_64-apple-darwin2\
2-default64-kv8/io.so, dlopen(./pkg/io//bin/x86_64-apple-darwin22-default\
64-kv8/io.so, 0x0005): Symbol not found: _AssGVar
  Referenced from: <691ED98B-03DB-3A1C-98A6-D5307CB4FCF6> /Users/dima/sof\
tware/gap/pkg/io/bin/x86_64-apple-darwin22-default64-kv8/io.so
  Expected in:     <6EA6D1DA-9BAB-322E-AEE4-887EFCA0E1EF> /Users/dima/sof\
tware/gap/tst/testlibgap/.libs/basic in
  LOAD_DYN( filename ) at ./lib/files.gd:595 called from 
<function "LoadDynamicModule">( <arguments> )
 called from read-eval loop at ./pkg/io//init.g:23
Error, was not in any namespace at ./lib/variable.g:269 called from
LEAVE_NAMESPACE(  ); at ./lib/package.gi:1327 called from
ReadPackage( pkgname, "init.g" ); at ./lib/package.gi:1688 called from
LoadPackage( pair[1], pair[2], false ) at ./lib/package.gi:1933 called from
func(  ); at ./lib/system.g:228 called from
<function "CallAndInstallPostRestore">( <arguments> )
 called from read-eval loop at ./lib/init.g:700
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> 
```

On linux this works fine:
```
$ ./tst/testlibgap/basic -l .
# Initializing GAP...
[...]

PR #5234, opened 2022-12-01
tests that libgap works correctly with kernel modules - draft

dimpase:
as a test, we load IO package - assuming it was built so that its kernel module is created.

That was missing, and caused a regression reported in https://github.com/gap-system/gap/issues/5232

When done, this PR will allow testing for the latter issue - which needs to be fixed elsewhere.

### DIFF
--- a/tst/testlibgap/basic.c
+++ b/tst/testlibgap/basic.c
@@ -6,6 +6,7 @@ int main(int argc, char ** argv)
 {
     printf("# Initializing GAP...\n");
     GAP_Initialize(argc, argv, 0, 0, 1);
+    test_eval("LoadPackage(\"io\");");
     test_eval("1+2+3;");
     test_eval("g:=FreeGroup(2);");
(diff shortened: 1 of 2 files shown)


dimpase:
this most probably will need a tweaking of `Makefile.rules`

dimpase:
@fingolfin - how can we make sure in `Makefile.rules` (?) that `io` is installed for this test?

fingolfin:
Thanks, it is a good idea to test that kernel extensions work with `libgap`.

Indeed, we already test that, as part of the `testmakeinstall` CI test, which runs `make install`, then builds `mockpkg` against that and tries loading it.

This PR however tests for something else: it builds a package against a non-installed `gap` (which is not using `libgap`) and then tries to load it in a separate binary which does use `libgap`. There is no promise whatsoever that this works, and so this is not supported, even though it currently may work in some setups (but this may change in the future).

PR #5236, opened 2022-12-01
Remove obsolete code from `BuildPackages.sh`, add comments

fingolfin:
- NormalizInterface now uses `prerequisites.sh` instead of `build-normaliz.sh`
- atlasrep uses `prerequisites.sh` to do `chmod 1777 datagens dataword`
- pargap is no longer distributed and supported
- if copying xgap.sh around is necessary for xgap, then the xgap build system should do that (I hope it's not necessary, but I can't test it)
- simpcomp also doesn't need the workaround anymore

Also add comments explaining why we do `make clean` (and why sometimes `configure` is run twice)

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -201,10 +201,16 @@ GAPInput
 )
       local CONFIG_ARGS_FLAG_NAME="PACKAGE_CONFIG_ARGS_${PKG_NAME}"
       echo_run ./configure --with-gaproot="$GAPROOT" $CONFIGFLAGS ${!CONFIG_ARGS_FLAG_NAME}
+      # hack: run `make clean` in case the package was built before with different settings
       echo_run "$MAKE" clean
     else
       echo_run ./configure "$GAPROOT"
+      # hack: run `make clean` in case the package was built before with different settings
       echo_run "$MAKE" clean
+      # hack: in browse and edim, `make clean` removes `Makefile` so run configure
+      # again to ensure we can actually build them (we could restrict this hack to
+      # the two offending packages, but since non-autoconf configure is super cheap,
+      # there seems little reason to bother)
       echo_run ./configure "$GAPROOT"
     fi
     echo_run "$MAKE"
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
Concerning XGAP, [Section 3.5 of the XGAP manual](https://docs.gap-system.org/pkg/xgap/htm/CHAP003.htm#SECT005) says that the startup script `xgap.sh` can be found in the `xgap` directory, which is still true after the change.
The additional instance in `$GAPROOT/bin`, which will not be created after the changes, is not mentioned in the XGAP. manual.

bin/BuildPackages.sh
```diff
@@ -201,10 +201,16 @@ GAPInput
 )
       local CONFIG_ARGS_FLAG_NAME="PACKAGE_CONFIG_ARGS_${PKG_NAME}"
       echo_run ./configure --with-gaproot="$GAPROOT" $CONFIGFLAGS ${!CONFIG_ARGS_FLAG_NAME}
+      # hack: run `make clean` in case the package was built before with different settings
       echo_run "$MAKE" clean
     else
       echo_run ./configure "$GAPROOT"
+      # hack: run `make clean` in case the package was built before with different settings
```
frankluebeck, line 208: I don't understand this code. At least for Browse and EDIM it would be enough to just call `./configure "$GAPROOT"` (a `"$MAKE" clean` is not needed). This generates a new `Makefile` and this causes that `"$MAKE"` below will recompile the object file.

fingolfin: yes but there are other package whose build system is not as clever; if one does `./configure` for a second time but with different arguments (or with a GAP that has changed to a different configuration), they don't rebuild their kernel extension at all. I've been working on getting those packages to fix / improve their build system, but until all are complete, we inserted as a workaround a `make clean` call to ensure they really rebuild. Which works great with all packages except edim and browse; hence we call configure again as a workaround for the new problem introduced by the first workaround.

ChrisJefferson (commented):
These are all good cleanups. I can't even remember why the "xgap copying" was done -- I suspect it came from some earlier script for building packages, when BuildPackages.sh was originally created. It is good we are moving towards everything been done in a standard way.

Issue #5233, opened 2022-12-01
add index entries for supported user preferences to the Reference Manual


ThomasBreuer:
The idea came up in the discussion of issue #5230.

Since there is no special GAPDoc magic for user preferences that would automatically generate index entries to their documentation, we can insert `<Index>` elements by hand in those places where the effects of the preferences are described.

For example, entering `??autoload` will then find the manual section(s) where the user preference `ExcludeFromAutoload` is mentioned.

fingolfin:
I think an actual "table" with all the user preferences defined by GAP (not by packages) would be useful to have.

We could write a small script to do that, which iterates over all the declared user preferences, and which generates an XML file from their names and descriptions, which we then just include into the manual. I imagine the code for `ShowUserPreferences` has all the required bits to do this.

I.e., the script could generate a file `doc/ref/_userpref.xml` or so. It could either be generated as part of `doc/make_doc`, or perhaps easier: it could be added to git, and we just check that it is kept up-to-date by adding a CI test regenerating it (and reporting a test failure if there is a diff).

That script then can of course also insert the appropriate `<Index>` elements (if it is necessary; depends a bit on the exact XML it generates).

fingolfin:
Bonus points if those entries can be referenced, so that when 76.2 talks about `ExcludeFromAutoload`, that could a `<Ref>` to its manual entry

fingolfin:
@ThomasBreuer yes that sums it up nicely.

ThomasBreuer:
O.k., I will provide a pull request for that.
(If we add GAPDoc markup to the text of the declarations then a release of the Browse will be needed to keep `BrowseUserPreferences` working.)

fingolfin:
This was resolved by PR #5248

PR #5248, opened 2022-12-06
better documentation for user preferences

ThomasBreuer:
This addresses issue #5233:

- added a new subsection "User Preferences Defined by GAP" to the section of the GAP Reference Manual that deals with user preferences,
- added the file `doc/ref/user_pref_list.xml` that contains the current list of descriptions of user preferences that are declared for GAP (not for GAP packages); this includes an index entry for each user preference,
- added the function `XMLForUserPreferences` that creates the contents of `doc/ref/user_pref_list.xml` from the declarations of the user preferences,
- added GAPDoc markup to the description strings in the declarations of GAP's user preferences,
- changed `ShowStringUserPreference` such that this markup disappears when the description is shown in the GAP session,
- added a remark about `BrowseUserPreferences` in the documentation of `ShowUserPreferences`,
- added a test that complains when `doc/ref/user_pref_list.xml` is  not up-to-date,
- moved some declarations of user preferences outside an `if` branch, in order to declare these preferences also if `readline` is not available,
- added some HPCGAP specific code to `lib/userpref.g`, in order to make `hpcgap/lib/userpref.g` superfluous.

*Note*:
The change concerning HPCGAP may cause trouble because the records stored in `GAPInfo.DeclarationsOfUserPreferences` are now *immutable*. Up to now it was in principle possible to declare a user preference with a *mutable* value, for example by defining a mutable list as its default, which can be modified later on. The changes proposed in this pull request will forbid this, and I am not sure whether we really want to do this --and then document it-- or whether we have to change the implementation in order to admit mutable values.

## Text for release notes
The new subsection "User Preferences Defined by GAP" lists GAP's user preferences,
and index entries for the preferences make it now easier to find their documentation.


doc/ref/user_pref_list.xml
```diff
@@ -0,0 +1,293 @@
...
+
+The defaults are computed at runtime.
+</Item>
+<Mark>
+<Index Key='ReadObsolete'><C>ReadObsolete</C></Index>
+<C>ReadObsolete</C>
+</Mark>
+<Item>
+May be useful to say <K>false</K> here to check if you are using commands
+which may vanish in a future version of &GAP;
```
fingolfin, line 212: Ugh, this text really should be improved. But of course that's out of scope for this PR!

fingolfin (approved):
Looks good to me but I think @ThomasBreuer wanted to make a few more adjustments, so I'll wait with merging until he declares it complete

doc/ref/user_pref_list.xml
ThomasBreuer: Yes, some of the "description" texts have not been written under the assumption that they will appear in the manual.

PR #5250, opened 2022-12-06
Fix an unexpected error during tab completion for input like `a.b.` when `a` is not a record

### LINKED ISSUE #5241: Error in autocompletion when trying to access a component object via record access by accident
Steps to reproduce:
```gap
gap> G := SymmetricGroup( 3 );;
gap> G.test.
```
and now press Tab to autocomplete.

### Observed behaviour

```gap
Error, illegal access to record component `IsBound(obj.test)'
of the object <obj>. (Objects by default do not have record components.
The error might be a relic from translated GAP3 code.) at /usr/share/gap/lib/record.gi:171 called from
if not hasbang[j - 1] and IsBound( r.(cmps[j]) ) then
    r := r.(cmps[j]);
elif hasbang[j - 1] and IsBound( r!.(cmps[j]) ) then
    r := r!.(cmps[j]);
else
    r := fail;
    break;
fi; at /usr/share/gap/lib/cmdledit.g:979 called from
GAPInfo.CommandLineEditFunctions.Functions.(l[2])( l ) at /usr/share/gap/lib/cmdledit.g:291 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:2
you can 'return;' to continue
brk>
```

### Expected behaviour

Of course the real problem is the user typing a wrong expression, but I think it would be nice to fail in a more graceful way here. If you don't think so, or if fixing this is not easy (there might actually be a method for `\.` installed and I'm not sure if one can easily detect this), feel free to close this issue.

```
 ┌───────┐   GAP 4.12.1 of 2022-10-20
 Packages:   GAPDoc 1.6.6.dev, IO 4.7.0dev, PrimGrp 3.4.2, SmallGrp 1.5, TransGrp 3.6.3
```

fingolfin:
Resolves #5241 (cc @zickgraf)

### DIFF
--- a/dev/gaptest.expect
+++ b/dev/gaptest.expect
@@ -50,6 +50,23 @@ expect "gap> "
 send -- "g!.S\t"
 expect "ize"
 send -- ";;\r"
+
+# issue 5241
+expect "gap> "
+# this used to trigger an error
+send -- "g.x.\t"
+# ctrl-u to clear the input given so far
+send -- "\025"
+# just start a new line to verify we are not in a break loop
+send -- "\r"
+expect "gap> "
+# this also used to trigger an error
+send -- "g!.x.\t"
+# ctrl-u to clear the input given so far
+send -- "\025"
+# just start a new line to verify we are not in a break loop
+send -- "\r"
 expect "gap> "
 
 exit
+
--- a/lib/cmdledit.g
+++ b/lib/cmdledit.g
@@ -969,9 +969,9 @@ GAPInfo.CommandLineEditFunctions.Functions.Completion := function(l)
       if Length(cmps) > 0 and cmps[1] in idbnd then
         r := ValueGlobal(cmps[1]);
         for j in [2..Length(cmps)] do
-          if not hasbang[j-1] and IsBound(r.(cmps[j])) then
+          if not hasbang[j-1] and IsRecord(r) and IsBound(r.(cmps[j])) then
             r := r.(cmps[j]);
-          elif hasbang[j-1] and IsBound(r!.(cmps[j])) then
+          elif hasbang[j-1] and (IsRecord(r) or IsComponentObjectRep(r)) and IsBound(r!.(cmps[j])) then
             r := r!.(cmps[j]);
           else
             r := fail;


FriedrichRober:
So I checked your branch on my laptop right now, it doesn't throw an error anymore.

Just want to make sure, that the following behaviour is what you wanted to achieve.
I run the following code on my laptop (MacBook Pro, macOS)
```gap
gap> G := SymmetricGroup( 3 );;
gap> G.test.
```
Pressing tab yields a list of all available commands:
```gap
*
+
-
.
.:=
/
16Bits_AssocWord
16Bits_CombiCollector
16Bits_DepthOfPcElement
16Bits_Equal
16Bits_ExponentOfPcElement
16Bits_ExponentSums1
16Bits_ExponentSums3
16Bits_ExponentSyllable
16Bits_ExponentsOfPcElement
16Bits_ExtRepOfObj
16Bits_GeneratorSyllable
16Bits_HeadByNumber
16Bits_LeadingExponentOfPcElement
16Bits_LengthWord
16Bits_Less
16Bits_ObjByVector
16Bits_Power
16Bits_Product
16Bits_Quotient
16Bits_SingleCollector
32Bits_AssocWord
32Bits_CombiCollector
32Bits_DepthOfPcElement
32Bits_Equal
32Bits_ExponentOfPcElement
32Bits_ExponentSums1
32Bits_ExponentSums3
32Bits_ExponentSyllable
32Bits_ExponentsOfPcElement
32Bits_ExtRepOfObj
32Bits_GeneratorSyllable
32Bits_HeadByNumber
32Bits_LeadingExponentOfPcElement
32Bits_LengthWord
32Bits_Less
32Bits_ObjByVector
32Bits_Power
32Bits_Product
32Bits_Quotient
32Bits_SingleCollector
8Bits_AssocWord
8Bits_CombiCollector
8Bits_DepthOfPcElement
--More--
```

FriedrichRober (approved):

fingolfin:
@FriedrichRober thank you for the testing and feedback! Yes, this is intended behavior for now -- it just matches how GAP behaved in similar cases for years: when GAP has no matches to offer, it just suggest any identifier used as a record entry name. Of course that's not very helpful if you start with just say `a.b.` and it prints thousands of potential completions; but if you start with `a.limitw` and then it autocompletes to `a.limitworkspace` that could be helpful. That said, if somebody has a better idea how to deal with this situation, I'd be open to a PR to change it -- but this is way beyond the scope of this PR.

fingolfin:
Backported to stable-4.12 via commit dbce4445e88910a7b3e12ea8084741868abba648

Issue #5231, opened 2022-11-30
change the way how `ImmutableVector` and `ImmutableMatrix` are used


ThomasBreuer:
Currently `ImmutableVector` is used in situations where one wants to replace a given vector by a better one, from the viewpoint of the internal representation.
In the context of vector objects, this is not exactly what one wants to achieve. One of the ideas behind vector objects is that one wants to create new vectors in a representation that fits to the one of a given vector.
For that, the syntax `Vector( v, w )` is intended, where `w` is a sample vector of the nice type, and `v` is the input vector for which the conversion to the type of `w` is returned.

The `ImmutableVector` calls in GAP library code belong to two different situations:
Those (few!) cases where initial vectors get created can remain as they are, for the moment.
And those where one wants to get a vector that fits to a known one should be replaced by `Vector( v, w )` calls.
For that, it would be useful to admit the `Vector( v, w )` syntax also for the case that `w` is a plain list.

For example, the code that deals with the coefficient vectors stored in elements of fields created with `AlgebraicExtension` needs `ImmutableVector` only in one place, where the first element gets constructed and stored in the elements family of the field; all coefficient vectors created later on can/should be constructed w.r.t. this first vector.

If nobody objects then I will create a pull request that implements these changes.

An open question:
Currently it is not specified whether the return value of `Vector( v, w )` is mutable or not, perhaps depending on the mutability of `v` and/or `w`.
In many places in the library, one wants immutable results, `w` is immutable, and `v` has just been created and is usually mutable (but can be made immutable in place without much overhead).
What shall we define?

(`ImmutableMatrix` should be handled analogously.)
(See also issues #5030 and #5181)

ThomasBreuer:
Concerning the mutability of the results of `Vector( v, w )` and `Matrix( v, w )`,
currently my favourite definition is that the result is immutable if and only if `v` is immutable.
That is, the (mutability of the) first argument determines the mutability of the result, and the second argument determines the `BaseDomain` and internal representation of the result.

This way, `Vector` is not forced to make a mutable or immutable copy of `v` in those cases where `v` gets wrapped in some vector object; note that the documentation of `Vector` states already that it is not guaranteed that a given list of entries will be copied.
For `Matrix`, the situation is a bit different, the documentation says

> It  is  guaranteed  that  the  given  list  list  is  copied in the sense of ShallowCopy  (12.7-1).  If  list  is a nested list then it is not guaranteed that also the entries of list are copied.

[...]

PR #5254, opened 2022-12-07
matrixobj: deal with mutability questions

ThomasBreuer:
(This addresses issue #5231.)

I hope this proposal makes sense. Comments are welcome.

- changed the mutability definition for vector and matrix objects:
  - The `ConstructingFilter` value indicates whether mutable variants are possible at all, as follows: If the filter implies `IsCopyable` then one can create mutable variants, otherwise not.
  - In the definitions of `NewVector`, `ZeroVector`, `NewMatrix`, `ZeroMatrix`, `IdentityMatrix`, promise (fully) mutable results if and only if the filter in question implies `IsCopyable`.
  - In the definitions of `Vector` and `Matrix`, promise mutable results if and only if the filter in question implies `IsCopyable` and the vector/matrix that defines the entries is mutable.

- changed the currently available `ConstructingFilter` values such that they imply `IsCopyable`. (I assume that these filters can be found as follows:  ``` flags:= List( [ IsVectorObj, IsMatrixObj ], FLAG1_FILTER );; Filtered( FILTERS, x -> IsRepresentation( x ) and Intersection( flags, TRUES_FLAGS( WITH_IMPS_FLAGS( FLAGS_FILTER( x ) ) ) ) <> [] ); ``` )

- extended the definition of `Vector( v, w )` such that also the case of `w` in `IsPlistRep` is supported, and added a method for both arguments in `IsList and IsCyclotomicCollection` (to be used for example for objects in `IsAlgebraicElement`).


wucas (approved):

lib/matobjnz.gd
```diff
@@ -12,14 +12,78 @@
...
+##  <#GAPDoc Label="IsZmodnZMatrixRep">
+##  <ManSection>
+##  <Filt Name="IsZmodnZMatrixRep" Arg='obj' Type='Representation'/>
+##
+##  <Description>
+##  An object <A>obj</A> in <Ref Filt="IsZmodnZMatrixRep"/> describes
+##  a matrix object (see <Ref Filt="IsMatrixObj"/>) that behaves like the
+##  list of its rows (see <Ref Filt="IsRowListMatrix"/>).
+##  <A>obj</A> is internally represented as a positional object
+##  (see <Ref Filt="IsPositionalObjectRep"/>) with <M>4</M> entries.
```
fingolfin, line 59: As we just discussed on gather.town, I'd prefer to either keep these implementation specifics out of the user-visible documentation, or else to at least qualify them to indicate that these are subject to change and one should not write code exploiting them but rather use official interfaces.

fingolfin (changes requested):
As discussed I am overall very happy with this except for the low-level implementation details (see comment). Thomas is working on this; once resolved we can merge!

ThomasBreuer:
@fingolfin I have addressed your request.
(Somehow your comment did not get marked as "outdated": The lines in question are still there, but they were moved away from the block that is used for the documentation.)

fingolfin (approved):

PR #5256, opened 2022-12-08
kernel: refactor code.c and related code

fingolfin:
I've been working on refactoring the coder and related code. Right now the same infrastructure is used to read function bodies for executions, and to write them. That prevents us from declaring more pointers `const`, (and thus might miss some optimizations -- but I don't expect much in this regard, so that's not really a factor for me). More importantly, it untangles the coder mostly from `STATE(PtrBody)` and related machinery, which is somewhat "magical" and difficult to understand. Indeed, I discovered and removed one place where we violated the layering between reader -> interpreter -> coder (see commit `kernel: replace hack in the reader for setting function locks`).

Future planned steps: removing the `visited` flag and the line numbers from stats, moving them to separate data structures. And getting rid of the final uses of `PtrBody` in there, to wit, by replacing uses of `SWITCH_TO_NEW_LVARS` and `SWITCH_TO_OLD_LVARS` with new alternatives specific to the coder.

### DIFF
--- a/src/code.c
+++ b/src/code.c
@@ -81,9 +81,9 @@ Obj NewFunctionBody(void)
 **  Note  that  it is *fatal*  to apply  'ADDR_EXPR'   to expressions of type
 **  'EXPR_REF_LVAR' or 'EXPR_INT'.
 */
-static Expr * ADDR_EXPR(Expr expr)
+static Expr * ADDR_EXPR(CodeState * cs, Expr expr)
 {
-    return (Expr *)STATE(PtrBody) + expr / sizeof(Expr);
+    return (Expr *)PTR_BAG(cs->currBody) + expr / sizeof(Expr);
 }
 
 /****************************************************************************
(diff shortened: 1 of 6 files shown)


fingolfin:
Assertion fixed in new commit `kernel: initialize CodeState in SYNTAX_TREE_CODE`

fingolfin:
Actually I should adjust CodeBegin to properly init the `CodeState`. But I first want to wait if CI passes now

fingolfin:
@ChrisJefferson this is good to go now from my perspective

src/code.h
```diff
@@ -46,9 +46,14 @@ struct CodeState {
     // the offset in the current body. It is only valid while coding
     Stat OffsBody;
 
+    // a linked list of further OffsBody values, used when coding
+    // nested function expressions
+    Obj OffsBodyList;
```
ChrisJefferson, line 64: I think this needs InitGlobalBag adding to code.c:InitKernel?

src/code.h
```diff
@@ -55,6 +55,12 @@ struct CodeState {
 
     // place to store the previous value of STATE(CurrLVars)
     Bag CodeLVars;
+
+    //
+    Obj currFunc;
```
ChrisJefferson, line 73: This should probably be InitGlobalBad in code.c as well (I'm going on the fact that CodeLVars is, which is mentioned just above)

src/code.h
fingolfin: Ah, good one... For a moment I was sweating 😅 However, all instances of `struct CodeState` are allocated on the stack, so I think `InitGlobalBag` is neither possible nor necessary. Or am I missing something?

ChrisJefferson: I'll be honest, my reasoning was CodeResult, which is stored just below, is `InitGlobalBag`ed on code.c:3237

fingolfin: That InitGlobalBag call gets removed by this PR

src/code.c
```diff
@@ -3233,9 +3168,6 @@ static Int InitKernel (
     MakeBagTypePublic(T_BODY);
 #endif
 
-    /* make the result variable known to Gasman                            */
-    InitGlobalBag(&CS(CodeResult), "CodeResult");
```
fingolfin, line 3237: @ChrisJefferson see, `CodeResult` is not handled by `InitGlobalBag` anymore.

Any other concerns? I am happy to answer. Of course this PR is rather too big, though I think reviewing it commit by commit is just barely manageable (the bigger commits are mostly a result of search and replace)

ChrisJefferson (approved):

Issue #4914, opened 2022-06-29
MatrixObj: support `Is8BitMatrixRep` with zero rows (so `0 x N`)


fingolfin:
Reported by @james-d-mitchell : the above has a the problem that while you can create such an object, it does not know its basedomain, because it relies on the rows to store the basedomain (and also the number of columns)

To fix this we "just" need to store info about the base domain (coefficient field) and the number of columns.

PR #5257, opened 2022-12-08
Add assertions and other argument checks for 8 bit matrix rep

james-d-mitchell:
This PR contains some minor changes to `src/vec8bit.c` mostly formatting (just running clang-format on the file), change `assert` to `GAP_ASSERT` everywhere for consistency, and I've added some assertions and `Requires` to the functions for matrices in `Is8BitMatrixRep`. 

This is just some preparation for tackling #4914, to make the changes required for that easier. 

## Text for release notes

None.


james-d-mitchell:
I'm not sure if the failing CI is a result of any changes in this PR or not, any ideas anyone?

ChrisJefferson (approved):

james-d-mitchell:
Thanks @ChrisJefferson

fingolfin:
I've restarted the failing check earlier today, and it has passed now :-)

lib/mat8bit.gi
```diff
@@ -1004,7 +1004,7 @@ InstallMethod(SemiEchelonMatTransformationDestructive,
 ##
 
 InstallMethod(TriangulizeMat,
-        "kernel method for plain list of GF2 vectors",
+        "kernel method for plain list of GF2 vectors", # TODO not sure this is accurate
```
fingolfin, line 1007: This is definitely wrong, likely copy & past. From the C source comment:
`Method for TriangulizeMat for plain lists of 8 bit vectors`
So perhaps use this:
```suggestion
        "kernel method for plain list of 8 bit vectors",
```
The same problem then affects other methods, e.g. `DeterminantMatDestructive` below. Also, the comment for this method here in line 1003 refers to GF2, and so on.

fingolfin (approved):
Looks good to me. You may want to either resolve or drop that `TODO` comment at the top (fixing this "fully" likely would require touching quite a few other functions, not sure you'd want to (or should) tackle this here...?)

james-d-mitchell:
Thanks, I've removed the TODO and will tackle that in another PR.

fingolfin:
Thanks!

PR #5261, opened 2022-12-12
Remove VISITED_STAT

ChrisJefferson:
This remove VISITED_STAT and SET_VISITED_STAT, and the 'visited' bit from Stat, and keeps track of these things in profiling.c, and keep track of covered lines in profiling.c

This remove `ActivateProfileColour`, which allowed colouring functions with covered lines, hitting the exact parts of lines executed, as that was easy to do when we tracked what was covered per Stat, but I don't do that now (only by line), as I believe no-one was ever using this functionality!

If this is fine, please don't merge it yet, as I'll have to release an updating to the profiling package, as to sync up with this change (in particular, remove tests which use `ActivateProfileColour`)

### DIFF
--- a/src/code.h
+++ b/src/code.h
@@ -26,10 +26,6 @@
 **  Header for any statement or expression encoded in a function body.
 */
 typedef struct {
-    // `visited` starts out as 0 and is set to 1 if the statement or
-    // expression has ever been executed while profiling is turned on
-    unsigned int visited : 1;
-
     // `line` records the line number in the source file in which the
     // statement or expression started
     unsigned int line : 31;
@@ -374,30 +370,6 @@ EXPORT_INLINE Int LINE_STAT(Stat stat)
 {
     return CONST_STAT_HEADER(stat)->line;
 }
-
-
-/****************************************************************************
-**
-*F  VISITED_STAT(<stat>) . . . . . . . . . . . if statement has even been run
-**
-**  'VISITED_STAT' returns true if the statement has ever been executed
-**  while profiling is turned on.
-*/
-EXPORT_INLINE Int VISITED_STAT(Stat stat)
-{
-    return CONST_STAT_HEADER(stat)->visited;
-}
-
-
-/****************************************************************************
-**
-*F  SET_VISITED_STAT(<stat>) . . . . . . . . . . mark statement as having run
-**
-**  'SET_VISITED_STAT' marks the statement as having been executed while
-**  profiling was turned on.
-*/
-void SET_VISITED_STAT(Stat stat);
-
 
 /****************************************************************************
 **
--- a/src/profile.c
+++ b/src/profile.c
@@ -149,12 +149,6 @@ static struct ProfileState
   int profiledThread;
 #endif
 
-  // Have we previously profiled this execution of GAP? We need this because
-  // code coverage doesn't work more than once, as we use a bit in each Stat
-  // to mark if we previously executed this statement, which we can't
-  // clear
-  UInt profiledPreviously;
-
   Int LongJmpOccurred;
 
   // We store the value of RecursionDepth each time we enter a function.
@@ -723,7 +717,6 @@ enableAtStartup(char * filename, Int repeats, TickMethod tickMethod)
 
     profileState.status = Profile_Active;
     RegisterThrowObserver(ProfileRegisterLongJmpOccurred);
-    profileState.profiledPreviously = 1;
 #ifdef HPCGAP
     profileState.profiledThread = TLS(threadID);
 #endif
@@ -775,12 +768,6 @@ static Obj FuncACTIVATE_PROFILING(Obj self,
 {
     if (profileState.status != Profile_Disabled) {
       return Fail;
-    }
-
-    if(profileState.profiledPreviously &&
-       coverage == True) {
-        ErrorMayQuit("Code coverage can only be started once per"
-                     " GAP session. Please exit GAP and restart. Sorry.",0,0);
     }
 
     memset(&profileState, 0, sizeof(profileState));
@@ -851,7 +838,6 @@ static Obj FuncACTIVATE_PROFILING(Obj self,
 
     profileState.status = Profile_Active;
     RegisterThrowObserver(ProfileRegisterLongJmpOccurred);
-    profileState.profiledPreviously = 1;
 #ifdef HPCGAP
     profileState.profiledThread = TLS(threadID);
 #endif
--- a/src/profile.c
+++ b/src/profile.c
@@ -532,7 +532,7 @@ static inline void printOutput(int fileid, int line, BOOL exec, BOOL visited)
 
 // Mark line as visited, and return true if the line has been previously
 // visited (executed)
-BOOL markVisited(int fileid, UInt line)
+static BOOL markVisited(int fileid, UInt line)
 {
     // Some STATs end up without a file or line -- do not output these
     // as they would just confuse the profile generation later.
@@ -553,31 +553,6 @@ BOOL markVisited(int fileid, UInt line)
         return FALSE;
     }
     return TRUE;
-}
-
-// Return TRUE is Stat has been visited (executed) before
-BOOL visitedStat(Stat stat)
-{
-    int fileid = getFilenameIdOfCurrentFunction();
-    int line = LINE_STAT(stat);
-
-    if (fileid == 0 || line == 0) {
-        return TRUE;
-    }
-
-    if (LEN_PLIST(profileState.visitedStatements) < fileid ||
-        !ELM_PLIST(profileState.visitedStatements, fileid)) {
-        return FALSE;
-    }
-
-    Obj linelist = ELM_PLIST(profileState.visitedStatements, fileid);
-
-    if (LEN_PLIST(linelist) < line || !ELM_PLIST(linelist, line)) {
-        return 0;
-    }
-    else {
-        return 1;
-    }
 }
 
 // type : the type of the statement
--- a/src/profile.c
+++ b/src/profile.c
@@ -275,7 +275,8 @@ static Obj JsonEscapeString(Obj param)
     return copy;
 }
 
-
+// Check if the filename of the file 'id' has even been outputted.
+// We only output this once per file to reduce file size.
 static inline void outputFilenameIdIfRequired(UInt id)
 {
     if (id == 0) {
@@ -343,6 +344,7 @@ static void HookedLineOutput(Obj func, char type)
   HashUnlock(&profileState);
 }
 
+// Called whenever a function is entered
 static void enterFunction(Obj func)
 {
 #ifdef HPCGAP
@@ -354,6 +356,7 @@ static void enterFunction(Obj func)
     HookedLineOutput(func, 'I');
 }
 
+// Called whenever a function exits
 static void leaveFunction(Obj func)
 {
 #ifdef HPCGAP
@@ -377,6 +380,10 @@ static void leaveFunction(Obj func)
 ** If we could rely on the existence of the IO package, we would use that here.
 ** however, we want to be able to start compressing files right at the start
 ** of GAP's execution, before anything else is done.
+**
+** This has not switched to using GAP's internal gzip support because by
+** using an external gzip we get free parallelisation, and GAP code is not
+** 'charged' for the time taken to compress the profile output.
 */
 
 static BOOL endsWithgz(const char * s)
@@ -558,6 +565,7 @@ static BOOL markVisited(int fileid, UInt line)
 // type : the type of the statement
 // exec : are we executing this statement
 // visit: Was this statement previously visited (that is, executed)
+// This deals with code which has been compiled into bytecode.
 static inline void
 outputStat(Int fileid, int line, int type, BOOL exec, BOOL visited)
 {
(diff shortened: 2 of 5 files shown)


src/profile.c
```diff
@@ -170,6 +163,9 @@ static struct ProfileState
   // We need to store the actual values, as RecursionDepth can increase
   // by more than one when a GAP function is called
   Obj visitedDepths;
+
+  // Mark which statements in which files have been visited
```
fingolfin, line 167: Perhaps state what kind of object it is and how it is structured?

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
     }
 }
 
+// Mark line as visited, and return true if the line has been previously
+// visited (executed)
+BOOL markVisited(int nameid, UInt line)
+{
+    if (nameid == 0 || line == 0) {
```
fingolfin, line 490: Can this ever happen, and when?

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+        return 1;
```
fingolfin, line 491:
```suggestion
        return TRUE;
```

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+        !ELM_PLIST(profileState.visitedStatements, nameid)) {
+        AssPlist(profileState.visitedStatements, nameid,
+                 NEW_PLIST(T_PLIST, 0));
+    }
+
+    Obj linelist = ELM_PLIST(profileState.visitedStatements, nameid);
+
+    if (LEN_PLIST(linelist) < line || !ELM_PLIST(linelist, line)) {
+        AssPlist(linelist, line, True);
+        return 0;
```
fingolfin, line 504:
```suggestion
        return FALSE;
```

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+    if (LEN_PLIST(profileState.visitedStatements) < nameid ||
+        !ELM_PLIST(profileState.visitedStatements, nameid)) {
+        AssPlist(profileState.visitedStatements, nameid,
+                 NEW_PLIST(T_PLIST, 0));
+    }
+
+    Obj linelist = ELM_PLIST(profileState.visitedStatements, nameid);
+
+    if (LEN_PLIST(linelist) < line || !ELM_PLIST(linelist, line)) {
+        AssPlist(linelist, line, True);
```
fingolfin, line 503: Perhaps we can use a blist instead of a plist to reduce memory usage by a factor of 64 or so? I think the main missing ingredient would be a function which takes a blist and efficiently increases its size by adding zero blocks.

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+        return 0;
+    }
+    else {
+        return 1;
+    }
```
fingolfin, line 508:
```suggestion
    return TRUE;
```

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+}
+
+// Return TRUE is Stat has been visited (executed) before
+BOOL visitedStat(Stat stat)
+{
+    int nameid = getFilenameIdOfCurrentFunction();
+    int line = LINE_STAT(stat);
+
+    if (nameid == 0 || line == 0) {
+        return 1;
```
fingolfin, line 518:
```suggestion
        return TRUE;
```

src/profile.c
```diff
@@ -487,13 +483,61 @@ static inline void printOutput(UInt line, int nameid, int exec, int visited)
...
+    }
+
+    if (LEN_PLIST(profileState.visitedStatements) < nameid ||
+        !ELM_PLIST(profileState.visitedStatements, nameid)) {
+        return 1;
```
fingolfin, line 523: Shouldn't this return *false*?
```suggestion
        return FALSE;
```

src/profile.c
fingolfin: Oh and perhaps another helper which takes a blist and an index and returns `TRUE` or `FALSE` (not `True` or `False`) ... this helper then could also just return 0 for anything beyond the bounds. That would simplify the code here a bit, no?

fingolfin:
I've converted this to "draft" status so that it is clear that it should not be merged. You can turn mark it as "ready for review" as soon as `profiling` is updated (and despite the name, I'll be happy to review it regardless of the "draft" status ;-)

src/profile.c
ChrisJefferson: As you say, I started trying to use blists then discovered there is no way to extend a blist while having it remain a blist.

It probably wouldn't be a bad idea to add such extensions to blists (maybe even accessible from the GAP level), but that was a rabbit-hole I didn't want to fall down, as it could be done later / more generally (should we just add a generic "extend list, and use this member to fill the new values, and special case that for blists?)

src/profile.c
ChrisJefferson: I think I've seen this with SyntaxTree (that doesn't bother setting up this information), and also something the encoder does something weird when special casing some value, for example from memory FALSE and TRUE statements are handled specially. These could get filtered out earlier.

fingolfin: Fair enough. Perhaps we should add a brief comment mentioning this?

src/profile.c
ChrisJefferson: I'll have a quick look. I worry about not testing such functions (you didn't do SET_LEN_BLIST in SAFE_SET_BIT_BLIST, which I think is required), I will take a look at doing this with BLists, however.

fingolfin: Oh sure. I'm not even sure that code compiles, I just copy&pasted it quickly. Yeah, it should have something like `if (pos > LEN_BLIST(list)) SET_LEN_BLIST(list, pos);` added

fingolfin:
So this now needs a rebase, but hopefully it won't be too hard.

Also you wanted to update `profiling` to match before merging this PR -- so I assume you'll do that next (if this is just about dropping the `ActivateProfileColour` code there should be no problem to release the `profiling` update first, and put it into the package distro, before merging this PR. Right?)

ChrisJefferson:
I believe this PR is now complete. I haven't switched to using blists as the memory usage isn't that big, and I would prefer some general and tested code for extending blists, rather than a poorly tested bit of code straight in the profiling.

This does make GAP as a whole a bit slower, but not massively, and that's probably unavoidable as we used to be able to just dig into the STAT, while now we have to go and look up through a couple of plists. After merging, and ensuring everything is working correctly, I might have a go at optimising.

There is, however one issue -- this pushes the time taken to test HPC-GAP very close to 1 hour, which is (a) too long, and (b) close to where the test will fail if it hits the one-hour limit and times out. I'm going to investigate if this slowness is related to this PR, or a more general issue.

fingolfin:
Note that the 1 hour limit is self-imposed by us, we could increase it -- but I'd rather now. HPC-GAP has been exceptionally slow for quite some time now, much more so in the CI than on my laptop, and I don't know why

src/profile.c
```diff
@@ -83,9 +83,8 @@
 **     we provide -P (profiling) and -c (code coverage) options to the GAP
 **     executable so the user can start before code loading starts
 **
-**  3) Operating at just a line basis can sometimes be too course. We can
-**     see this when we use ActivateProfileColour. However, without some
-**     serious additional overheads, I can't see how to provide this
+**  3) Operating at just a line basis can sometimes be too course. However,
```
fingolfin, line 86:
```suggestion
**  3) Operating at just a line basis can sometimes be too coarse. However,
```

src/profile.c
```diff
@@ -83,9 +83,8 @@
...
+**     without some serious additional overheads, I can't see how to provide this
 **     functionality in output (basically we would have to store
 **     line and character positions for the start and end of every expression).
```
fingolfin, line 89: Actually I do think we could provide that without a problem nor overhead, by replacing the line number in the stat by a byte offset into the file. Converting this into line number + offset into the line can then be deferred to a post-processing step.

ChrisJefferson: There is a bit more problem here ( I could describe it better).

We tend to set up statements we tend to (as you would expect) parse, then make a `Stat`, so if we just grabbed the position when making the `Stat` we would point to the end of the statement / loop, / whatever.

With just covering lines this (usually) isn't too bad, but if you want to capture expressions usefully you probably want to catch their start and end (as expressions are often nested), and then we'd need to catch the start of expressions.

ChrisJefferson:
You are right, I looked on my machine and profiling is a little slower, unfortunately for CI generating coverage is particularly hit, because once a line is covered it was very cheap to check if the stat was covered and exit. However, the slowdown isn't unreasonable.

On my machine, HPC-GAP is about 50% slower than master, and coverage is about 50% slower than no coverage.

I could make a PR where we do a much smaller set of HPC-GAP testing? We could consider stopping testing it, but I don't know if the code is being partially used for the Julia integration?

fingolfin:
Could you perhaps move the removal of `ActivateProfileColour` into its own PR? Then we could merge that earlier, which would in turn simplify things for me in PR #5292

fingolfin:
I have rebased this now

fingolfin:
Rebased again.

fingolfin (commented):
Overall this looks fine to me, except for some HPC-GAP related locking concerns. Also, do you understand why this makes HPC-GAP so much slower? Or does it also make regular GAP with profiling a lot slower?

src/profile.c
```diff
@@ -480,11 +478,61 @@ static inline void printOutput(int fileid, int line, BOOL exec, BOOL visited)
...
+        return TRUE;
+    }
+
+    if (LEN_PLIST(profileState.visitedStatements) < fileid ||
+        !ELM_PLIST(profileState.visitedStatements, fileid)) {
+        AssPlist(profileState.visitedStatements, fileid,
+                 NEW_PLIST(T_PLIST, 0));
+    }
+
+    Obj linelist = ELM_PLIST(profileState.visitedStatements, fileid);
```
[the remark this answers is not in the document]
fingolfin: (I have no idea of either of these would be worthwhile, but then again, this is executed *for every single expression or statement*, so we should measure and make it as fast as possible?

src/profile.c
```diff
@@ -532,11 +530,61 @@ static inline void printOutput(int fileid, int line, BOOL exec, BOOL visited)
...
+
+    if (LEN_PLIST(linelist) < line || !ELM_PLIST(linelist, line)) {
+        AssPlist(linelist, line, True);
+        return FALSE;
+    }
+    return TRUE;
+}
+
+// Return TRUE is Stat has been visited (executed) before
+BOOL visitedStat(Stat stat)
```
fingolfin, line 559: This function is never called, maybe it can be removed? Or is it supposed to be called from somewhere after all?

fingolfin:
Rebased once again

src/profile.c
```diff
@@ -532,11 +530,36 @@ static inline void printOutput(int fileid, int line, BOOL exec, BOOL visited)
     }
 }
 
+// Mark line as visited, and return true if the line has been previously
+// visited (executed)
+static BOOL markVisited(int fileid, UInt line)
```
fingolfin, line 535: This function is not thread safe, yet is not called from a critical section either. Can't that cause problems in HPC-GAP?

So perhaps it needs a lock? Or perhaps we can make `profileState.visitedStatements` thread local -- it suffices to merge this into a single plist at the time we write out the profiles, no? (Though I guess that would break if a thread was killed before we do that...)

ChrisJefferson: (old reply!)

We only call this from one thread -- we lock profiling to a single thread (you can see this from the caller).

There could be issues to do with ownership of the datastructures here in HPC-GAP, if they are created in one thread and used in another. I'm not opposed to a fix for that, but it wouldn't shock me if various things don't work in HPC-GAP with profiling.

src/profile.c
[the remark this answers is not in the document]
ChrisJefferson: markedVisited is only about 2% of the time taken in profiling (that's less than I expected, but it's what cachegrind reports). Almost all the time is spent in printing.

ChrisJefferson:
Almost all the time taken in profiling is from the fprintf printing out JSON -- the main way to speed this up would be to switch to a binary format, or do something where we just print less output. Both of these are reasonable projects, but I'm not planning on working on either of them soon.

fingolfin:
@ChrisJefferson what are your intentions with this PR? It is still marked as draft. What is needed to complete it then?

src/code.h
```diff
@@ -26,10 +26,6 @@
 **  Header for any statement or expression encoded in a function body.
 */
 typedef struct {
-    // `visited` starts out as 0 and is set to 1 if the statement or
-    // expression has ever been executed while profiling is turned on
-    unsigned int visited : 1;
-
     // `line` records the line number in the source file in which the
     // statement or expression started
     unsigned int line : 31;
```
fingolfin, line 31: Might as well put that bit to use again (and possibly allow generation of slightly nicer code?)
```suggestion
    unsigned int line : 32;
```

fingolfin:
@ChrisJefferson ping?

PR #5262, opened 2022-12-13
kernel: move some doc comments for GAPState members, and some more comment cleanup

fingolfin:


fingolfin:
@ChrisJefferson this has a minor conflict with your PR #5261 but should be trivial to resolve either direction.

wilfwilson (approved):

PR #5263, opened 2022-12-13
Mark as obsolete: `ZeroSM`/`ZeroAttr`, `AdditiveInverseSM`/`AdditiveInverseAttr` `OneSM`/`OneAttr`, `InverseSM`/`InverseAttr`, `TransposedMatAttr` 

fingolfin:
Currently we have these three variants of `One`, together with a bunch of synonyms for each:

- `OneImmutable` == `One` == `OneAttr` == `Identity`
- `OneMutable` == `OneOp`
- `OneSameMutability` == `OneSM`

This is then repeated for `Zero`, `Inverse`, `AdditiveInverse`.

The delegation between these is already complicated enough, but each time I have to look into this, my head starts spinning due to the many synonyms, and I can't quite remember which is which. I propose we narrow this down a bit:

- the variants with `SM` seems to be purely to avoid some typing, so I've removed all uses of them. Turns out no packages uses it, either. So we can mark them as obsolete.

- the `Attr` variants are also almost unused, except for a single use in the FinInG package (removed in FinInG 1.5.4). Still, we can safely mark it as obsolete, and remove all users of it beyond the documentation. (As a bonus, also make `TransposedMatAttr` obsolete.)

- the `Op` synonym to me also seems redundant. Alas, a ton of packages use them, so I mostly left those alone.

- I think it is fine to keep `One` (and `Identity`) as synonyms for `OneImmutable`; at least for me personally it seems I can deal with that one synonym sufficiently well in my head.

## Text for release notes

## Further details

### DIFF
--- a/doc/ref/lists.xml
+++ b/doc/ref/lists.xml
@@ -1368,8 +1368,8 @@ an operation that returns a result that is mutable
 (<Ref Oper="ZeroOp"/>, <Ref Oper="AdditiveInverseOp"/>,
 <Ref Oper="OneOp"/>, <Ref Oper="InverseOp"/>),
 and an operation whose result has the same immutability level as the argument
-(<Ref Oper="ZeroSM"/>, <Ref Oper="AdditiveInverseSM"/>,
-<Ref Oper="OneSM"/>, <Ref Oper="InverseSM"/>).
+(<Ref Oper="ZeroSameMutability"/>, <Ref Oper="AdditiveInverseSameMutability"/>,
+<Ref Oper="OneSameMutability"/>, <Ref Oper="InverseSameMutability"/>).
 The last kind of operations is equivalent to the corresponding infix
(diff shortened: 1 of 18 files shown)


ThomasBreuer (approved):
This looks good.
(Apparently there is still one instance of `AdditiveInverseAttr` in `tst/testspecial/bad-minus.g.out`.)

Concerning the history of the names in question, I think that `Zero` (for the mathematical concept) is the oldest name, and `ZeroAttr` and `ZeroOp` had been introduced at the same time in order to specify an immutable and a mutable zero.
If I remember well then the names `ZeroImmutable` and `ZeroMutable` have been introduced later.
Names of the form `SomethingAttr` and `SomethingOp` occur also in other situations (for example `BlocksAttr` and `BlocksOp`), but they are usually undocumented.

fingolfin:
@ThomasBreuer thanks for the explanation regarding the background, that's indeed illuminating, and makes a lot of sense.

PR #5268, opened 2022-12-14
Remove trailing whitespace from all files in `lib`

james-d-mitchell:
This PR removes all tabs and trailing whitespace in all files in the `lib` directory, as mentioned in #5260. I could also add a CI job that would fail if tabs or trailing whitespace are introduced again in the future, but since I expect this will be controversial enough without that too, I haven't done it.

## Text for release notes

None


james-d-mitchell:
Thanks @wilfwilson I think I've added the appropriate file also.

wilfwilson:
@james-d-mitchell I don't think that `.git-blame-ignore-revs` will work as-is. Firstly, it references an old commit, but most importantly, I think it needs to be done after the PR is merged: the hashes of the commits in this PR will definitely change when it is rebase-and-merged onto master.

In addition, there are two relevant commits to ignore (the tab one and the trailing whitespace one), not just one, so ultimately two commits will need to be listed.

james-d-mitchell:
Aha, thanks @wilfwilson, that makes sense, I'll drop that commit.

lib/algfld.gi
```diff
@@ -1279,9 +1279,9 @@ local opt,irrfacs, coeffring, i, factors, ind, coeffs, val,
       Add( factors, factor );
       q:= Quotient( R, g, factor );
       while q <> fail do
-	Add( factors, factor );
-	g:= q;
-	q:= Quotient( R, g, factor );
+    Add( factors, factor );
+    g:= q;
+    q:= Quotient( R, g, factor );
```
fingolfin, line 1284: Indentation is broken here. I would advise against trying to remove tabs with a fully automated script: in many places our tabs are 8 wide, but I think there are exceptions with 4 wide tabs (and who knows what else)

fingolfin (changes requested):
Thank you!

In principle I am all for this. The main objection I am aware of are:
- this could cause conflicts for in-flight PRs: OK, we have ~65 open PRs; but the majority of those are dormant and old and either are trivial to rebase or need major rebasing work anyway, and I don't intend to be held hostage by old unfinished work (which includes many of my own PRs -- I will gladly rebase them through whitespace changes, and also am happy to assist others with theirs)
- there may also be unsubmitted big changes -- well, if you don't show your work via a PR, we can't take it into account
- it "break git blame etc." -- but that's what we have `.git-blame-ignore-revs` for (to be added in a future PR)

So all in all I don't mind. In fact I have ulterior motives, I want to make some code formatting tweaks to the kernel code ;-).

But the tab changes are delicate. I would recommend to separate the "remove trailing whitespaces" changes and the tab changes into separate PRs. Then we can merge the "remove trailing tabs" PR rather quickly, but the tabs require careful vetting (and perhaps need to be done in multiple steps, i.e., convert just a couple files carefully and submit that, and deal with others later). Also, trailing spaces could also be removed in `src` and in `doc/*/*.xml` and possibly in `grp`.

Oh, and actually: please make sure to at least also apply all these changes to `hpcgap/lib`, to avoid divergence between those files and their counterparts in `lib`

james-d-mitchell:
I've dropped the remove tabs from this PR, I can see the rationale for replacing tabs more carefully (although the idea that anything could make the formatting of the library **worse** is amusing).

james-d-mitchell:
About HPC gap, I just did the same transformation on `hpcgap/lib` and there were only changes to 24 files, whereas in this PR there were changes to 421, so I think the idea that `lib` and `hpcgap/lib` are somehow in sync might need to be reexamined.

james-d-mitchell:
@fingolfin aha, I see, ok, I'll add that to the PR now.

fingolfin (approved):

fingolfin:
I will remove a couple more trailing whitespace with a direct push momentarily, and add `.git-blame-ignore-revs`

james-d-mitchell:
Thanks @fingolfin !

fingolfin:
Done. And of course I idiotically broken the tests based on manual examples, so I'll fix that ASAP.

PR #5270, opened 2022-12-14
Fix whitespace in a few documentation examples

fingolfin:
We recently removed a ton of trailing whitespace; but in some cases, these have meaning, such as in .tst files or inside `<Example>` elements after "empty" prompts such as `gap> ` or `> ` or `brk> `


mtorpey (approved):
Looks good to me!

PR #5271, opened 2022-12-14
kernel: convert many C comments to C++ comments

fingolfin:
Why? For the inline comments deal with in the first commit (made using a perl invocation) the reasoning is this: Adjusting comments like
```C
    /* catch infix operations                                          */
```
or
```C
    UInt                t1;             /* type of left  operand           */
    UInt                t2;             /* type of right operand           */
    ArithMethod2        func;           /* mod function                    */
```
to always fill *exactly* the correct number of columns is a major PITA. Failing to shorten such a comment also interacts badly with `clang-format`..

For multi-line comments, I usually also find it easier to come up with a uniform styling when using C++ comments; for multiline C comments, we have a plethora of ways to format them. Which of course isn't a serious problem on its own, except it adds mental load ("how should I format this comment??")

Note that this PR deliberately does not touch the big "block" comments of this kind (other than "fixing" some with "obvious" deviations):
```C
**
*F  SyRmdir( <name> )  . . . . . . . . . . . . . . .  remove directory <name>
*/
```
I actually also have a script converting those, but this may require more discussion and so I wanted to keep it seperate.


src/compiled.h
```diff
@@ -27,14 +27,14 @@ extern "C" {
...
 
-/* types, should go into 'gvars.c' and 'records.c' * * * * * * * * * * * * */
+// types, should go into 'gvars.c' and 'records.c' * * * * * * * * * * * *
 
 typedef UInt    GVar;
 
 typedef UInt    RNam;
 
 
-/* checks, should go into 'gap.c'  * * * * * * * * * * * * * * * * * * * * */
```
ChrisJefferson, line 37: If doing a cleanup, would it be possible to look for cases like this, where comments end in a line of stars, and clean those up?

src/costab.c
```diff
@@ -26,40 +26,40 @@
...
-static Obj      objExponent;            /* handle of subgroup order        */
-static Obj      objWordValue;           /* handle of word value            */
+static Obj      objExponent;            // handle of subgroup order
+static Obj      objWordValue;           // handle of word value
 
-static Int      treeType;               /* tree type                       */
-static Int      treeWordLength;         /* maximal tree word length        */
+static Int      treeType;               // tree type
+static Int      treeWordLength;         // maximal tree word length
 static Int      firstDef;               /*                                 */
```
ChrisJefferson, line 46: Maybe spot entirely empty comments and just remove them?

ChrisJefferson:
i'm happy for this to be merged as-is, and my comments to be dealt with later, just wanted to raise them as I checked to make sure nothing looked broken.

src/costab.c
```diff
@@ -372,15 +372,15 @@ static Obj FuncMakeConsequences(Obj self, Obj list)
...
+    Obj *               ptNums;         // pointer to this list
+    Int                 lp;             // left pointer into relator
+    Int                 lc;             // left coset to apply to
+    Int                 rp;             // right pointer into relator
+    Int                 rc;             // right coset to apply to
+    Int                 tc;             // temporary coset
+    Int                 i;              // loop variable
+    Obj                 hdTmp;          // temporary variable
 
     /*T 1996/12/03 fceller this should be replaced by 'PlistConv'          */
```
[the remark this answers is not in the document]
fingolfin: Aha, indeed. This is because of the `T`. We have to decide how to treat such comments. Presumably could be converted to `// TODO: `. Does that seems sensible?

Just to explain, the main part of this (in the first commit) was created using a very simple perl invocation. The idea is to take care of the 95% of "easy" cases in a fully automated way, to make it easy to rebase the PR if needed. Then the remaining cases may need some more manual work. Of course I can add more "automated" changes, but I'd like to keep each separate transformation as simple as possible.

src/costab.c
fingolfin: Indeed, my script didn't look for those. Will take care of them

fingolfin:
Ideally I'd like to get PR #5256 merged *first*, as it is easier to deal with resulting conflicts in this PR here (as most of the changes can be easily re-created by re-running a script.

src/compiled.h
fingolfin: Found just a few, all in this file. Fixed now.

src/costab.c
fingolfin: Fixed all but one were in `costab.c`, one more in `integer.c`

fingolfin:
This is now good to go from my POV

PR #5274, opened 2022-12-15
Remove unused lvars a files

james-d-mitchell:
This PR removes all declared but unused local variables in the files in the library starting with the letter "a", and removes tabs from one of these files again. 

## Text for release notes

None

### DIFF
--- a/lib/algebra.gi
+++ b/lib/algebra.gi
@@ -2056,7 +2056,6 @@ InstallMethod( ViewObj,
     "for an algebra-with-one with known generators",
     [ IsAlgebraWithOne and HasGeneratorsOfAlgebraWithOne ],
     function( A )
-    local nrgens;
     Print( "<algebra-with-one over ", LeftActingDomain( A ), ", with ",
            Pluralize( Length( GeneratorsOfAlgebraWithOne( A ) ), "generator" ),
            ">" );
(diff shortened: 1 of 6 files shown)


fingolfin (approved):

ChrisJefferson:
Out of interest, was this done by some kind of script, or hand-checking?

james-d-mitchell:
Done using `gaplint` with a config file that disables all rules except the one that analyses local variables:

https://github.com/james-d-mitchell/gaplint

fingolfin:
Off-topic, but: we could easily add code to the GAP kernel which issues a warning about unused locals. If someone is interested in that, just open an issue to remind us about it; I might work on it (or can leave some hints how to do it).

james-d-mitchell:
@fingolfin good idea, although it would likely generate a very large number of such warnings. This is probably a better approach than the one using gaplint.

fingolfin:
@james-d-mitchell yeah I definitely would want there to be an option to turn off that warning. Then people can work on fixing the warning in the GAP library, and eventually in packages.

fingolfin:
I don't understand what's going on. I've closed and reopened this PR; I've even rebased & force pushed its content. Each time the same CI job ends up as "cancelled". Why?!?

fingolfin:
OK, this time it seems I managed to get it started properly. We'll see

PR #5278, opened 2022-12-15
dev: change make_archives.py to use more PackageDistro data

### LINKED ISSUE #5172: CI for releases: take packages tarball and metadata from PackageDistro repository?
The GitHub workflow in `.github/workflows/release.yml` blindly uses `make bootstrap-pkg-full` to get whatever set of packages is current in that moment, by accessing https://github.com/gap-system/PackageDistro/releases/download/latest/

I propose to make a few changes to this process:

1. if we build an *actual* release, i.e., from a tag `v4.X.Y`, we should not download from https://github.com/gap-system/PackageDistro/releases/download/latest/ but rather from https://github.com/gap-system/PackageDistro/releases/download/v4.X.Y/
   - this way, we get the exact same tarball as was tagged there
   - this improves reproducibility
2. also fetch the `package-infos.json.gz` file from there, and re-use that, instead of generating our own
   - also improves reproducibility
   - ensures consistent data is present; e.g. the PackageDistro version of that JSON file adds SHA256 checksums of some files, it would be nice to have this present in the version of the file "shipped" in our GitHub release of GAP, too
3. the same could be done for `help-links.json` if we just teach the PackageDistro to produce this file, too (shouldn't be a problem)
4. bonus: with the previous two in place, `dev/releases/make_archives.py` does not longer need to build GAP and can thus be simplified

A downside of this could be that preparing a release *manually* requires slightly more work after these changes. To minimize the impact of this, the scripts in the PackageDistro should make sure it is still easy and convenient to create `package-infos.json` and `help-links.json`, say by ensuring it suffices to invoke a single script

fingolfin:
Change make_archives.py for actual releases `v4.x.y` to use the `v4.x.y` tag from https://github.com/gap-system/PackageDistro to download the package distribution (instead of the `latest` tag).

Also download package-infos.json.gz from there instead of recreating it.

<s>This resolves most of #5172 except for the handling of `help-links.json`.</s> Actually on second thought, this fully resolves #5172 to my content (the `help-links.json` bit is mostly irrelevant / a matter of taste)


ruthhoffmann (approved):
All looks good to me!

Issue #5285, opened 2022-12-18
Release Ci once again failed to build windows executables


fingolfin:
... and this time I only noticed after sending the release announcement *sigh*.

This is essentially a variant of #5011

The failed job: https://github.com/gap-system/gap/actions/runs/3722840754/jobs/6314090227

Immediate cause reported at https://github.com/gap-packages/agt/issues/15

This is really annoying, and hints at multiple problems:

1. of course I should have noticed the missing windows binaries
2. but since humans miss such things,  *our automation* should have noticed -- e.g. the scripts updating the website could have bailed out and refused the output based on the absence of those files
3. the CI tests for the PackageDistro perhaps could have also picked this one up, by checking for broken symlinks (or perhaps really *any* kind of symlinks, as those seem to be a repeated source of issues on Windows)
4. we lack a good mechanism to "heal" such issues: after all, it is trivial to work around the issue (just delete the offending symlink). But in practice I don't see any good way to achieve this without release 4.12.3: I can't just insert this `rm PATH` invocation into the workflow. And I also can't just re-tag, as it quite likely would produce new tarballs with differing shasums.

How to resolve this now? I see these options:
1. release GAP 4.12.3 with *identical* source code, just a `rm -f pkg/agt/doc/mathjax` added (urgh, doesn't seem appealing)
2. find a way to inject that `rm` into a re-run of the CI job (I see no way to do that, though I guess if we had `tmate` integration set up for that job, perhaps that would allow for it...)
3. I could re-tag 4.12.2 after all, after first downloading all relevant tarballs; then after all CI run, replace any of the new tarballs that changed SHA256
4. someone (@ChrisJefferson perhaps) could perhaps build the GAP .exe "manually" and upload it to the release

fingolfin:
Ok one more idea: I can try to hack the release CI script in my fork of this repository and create a 4.12.2 tag there to produce the desired .exe files.

fingolfin:
Also, another minor action to take against this kind of issue: let's add a step to dev/releases/README.md that suggest to check that all files are present.

another could be to have a "check release" script or GH workflow that checks this (and also verifies shasums while at it)

ChrisJefferson:
I have added windows 4.12.2 to the windows release, obviously, that doesn't fix the underlying problem, but it solves the short term issue!

fingolfin:
[...]

PR #5289, opened 2022-12-19
CI: revise the 'cygwin' job of the "Wrap releases" workflow

### LINKED ISSUE #5011: CI for creating Windows .exe for release tags is broken
I tagged v4.12.0 and our CI workflow for creating .exe files failed [(log)](https://github.com/gap-system/gap/runs/7907771045?check_suite_focus=true) with this error:
```
tar: gap-4.12.0/pkg/ace/htm/CHAP007.htm: Cannot create symlink to 'CHAP00A.htm': No such file or directory
tar: gap-4.12.0/pkg/ace/htm/CHAP008.htm: Cannot create symlink to 'CHAP00B.htm': No such file or directory
tar: gap-4.12.0/pkg/ace/htm/CHAP009.htm: Cannot create symlink to 'CHAP00C.htm': No such file or directory
tar: Exiting with failure status due to previous errors
```
Indeed, the latest ACE release has some symlinks here, to resolve issue #4430. When I made that change, I did not have in mind that Windows doesn't have true symlinks and might not like it.

We need to fix the immediate problem; but I also think there is a clear case that we have a long-term issue here, too: while we test the "release CI job" daily, the test was insufficient to catch this issue; because it occurred in one of the parts of the CI job that is really only executed for releases. Bad!

We need to resolve both.

*Short term*: I will cook up a workaround so the 4.12.0 release can be completed. I may just be evil and modify that gap-4.12.0 tarball by replacing (?) those symlinks

*Mid term*: A proper fix for the above issue needs to be found. Perhaps ACE should just not create a symlink there, but instead just copy the HTML file? Then again... why does it only fail when it extracts `https://github.com/gap-system/gap/releases/download/v4.12.0/gap-4.12.0.tar.gz` but does not fail when it extracts https://github.com/gap-system/PackageDistro/releases/download/latest/packages.tar.gz ?? My only theory is that the order in which the files occurs in the tarballs is different; in the "good" one the symlinks come after the real files, while in the "bad" one the symlinks come before, resulting in the observed error. However, this theory is backed by nothing other than my free interpretation of the error message, so it might be utter nonsense.

*Long term*: We need to enhance the CI test to find this earlier! To this end, I'd like to minimize the difference between when it runs on a release tag vs. the "usual" runs. One way would be to modify the offending step `Download the appropriate GAP release tarball` as follows: instead of downloading the release tarball, we download an artifact created by the previous job in the workflow. However, to conserve storage, we only upload those artifacts for scheduled runs; well, let's upload them also for release tags; and then modify the `Download the appropriate GAP release tarball`` to use these artifacts, but also run it on the scheduled runs...   Of course I am also open for other ideas!

CC @wilfwilson @ChrisJefferson @FriedrichRober

fingolfin:
First, we change the 'unix' job to always upload the .tar.gz it produces as an artifact. This puts a little extra burden on GitHub; but it lets us modify the 'cygwin' job to always use this artifact.

This removes a major point in which our simulated releases differed from actual releases. This subtle difference already caused trouble for two releases. With this change, I hope it won't cause trouble a third time.

As an added benefit, it also allows us to remove a bunch of custom code from the 'cygwin' job, and let's us avoid cloning GAP.

Motivated by issue #5285

Resolves #5011


fingolfin:
Excellent, the failure of https://github.com/gap-system/gap/issues/5285 is perfectly reproduced.

fingolfin:
So I've added a HACK commit which works around the issue currently troubling the package distro (i.e. it removes the broken symlink in the AGT package before creating the `gap-VERSION.tar.gz`) and with that the CI seems to pass; without it, it breaks as expected.

So in principle this seems ready to me, though I'd appreciate if it got some review; and of course that HACK commit should be removed, and the package distro be fixed instead (ideally by getting an update for the AGT package in).

.github/workflows/release.yml
```diff
@@ -158,98 +152,33 @@ jobs:
...
-      # In all cases, we currently need to clone GAP for its release scripts.
-      #
-      # If the GitHub event is a pushed release tag, then we compile/wrap a GAP
-      # tarball that we download from the GitHub release that was just created
-      # and uploaded by the preceding 'unix' job.
-      #
-      # Otherwise, we compile and wrap the appropriate GAP development clone
-      # into the installer.
-      #
-      # Alternatively, we could upload the desired GAP tarball via the
```
fingolfin, line 202: So basically this PR implements the alternative sketched in this comment

.github/workflows/release.yml
```diff
@@ -158,98 +152,33 @@ jobs:
...
-      # 'actions/upload-artifact' action in the preceding 'unix' job, and then
-      # use this in all cases in this 'cygwin' job. This would save some
-      # repeated work and hence time, and unify the code here slightly.
-      #
-      # Furthermore, this would remove the need for the
-      # download_release_archive.py script, and if we would also remove the need
-      # for the upload_file_to_github_release.py script, as described below,
```
fingolfin, line 209: argh, ok, so we still need to make sure `upload_files_to_github_release.py` is available, or replace it (which I'd prefer anyway).

But also: this mistake of mine was not revealed because this code path is only tested for a release. Bad!

Perhaps we can always perform the upload -- but to where? Not sure this is possible. But perhaps we can do it at least for the scheduled daily jobs: we could let that (re)create a release `nightly` (i.e. delete it if it exists, then tag the commit from which the scheduled workflow is run, and upload there). Then we still notice failures relatively quickly

fingolfin: Maybe also should get rid of `make_github_release.py`... I'll give it a spin

ruthhoffmann (approved):
Looks good to me. Was worried about `{{ matrix.arch }}` being replaced with a hardcoded `x86_64` (with more ARM arch being out there) but I don't see how this is an issue atm.

fingolfin:
Is there even aa cygwin version for ARM Windows? Anyway, if GitHub adds runners for ARM Windows, we can of course add that. The real reason for the matrix of course was that building these installers is soooo slooooooow. I've opened an issue about that

PR #5292, opened 2022-12-20
kernel: more coder refactoring

fingolfin:
- kernel: replace OffsBodyList by a stack again
- kernel: further untangle coder from executor

### DIFF
--- a/tst/testspecial/debug-var.g.out
+++ b/tst/testspecial/debug-var.g.out
@@ -118,32 +118,28 @@ brk_2>
 brk_2> # check coding of dvars
 brk_2> function() return unbound_higher; end;
 Error, Variable: <debug-variable-1-4> cannot be used here in
-  <corrupted statement>  called from 
-Error( "foobar" ); at *stdin*:20 called from
+  Error( "foobar" ); at *stdin*:20 called from 
 Error( "breakpoint" ); at *stdin*:9 called from
 g( 10 ) at *stdin*:12 called from
 <function "f">( <arguments> )
  called from read-eval loop at *errin*:9
 brk_2> function() return IsBound(unbound_higher); end;
 Error, Variable: <debug-variable-1-4> cannot be used here in
-  <corrupted statement>  called from 
-Error( "foobar" ); at *stdin*:20 called from
+  Error( "foobar" ); at *stdin*:20 called from 
 Error( "breakpoint" ); at *stdin*:9 called from
 g( 10 ) at *stdin*:12 called from
 <function "f">( <arguments> )
  called from read-eval loop at *errin*:10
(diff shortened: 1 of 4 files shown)


fingolfin:
Seems this breaks `profiling` and also some of the test in `testspecial` related to debug-vars. Will investigate at some point, but not now :-)

ChrisJefferson:
I haven't done a full explore, but i thought I'd drop some initial investigation, in case you can give me the easy answer.

The problem seems to be the following function which `profile.c` uses for finding the filename often (but not always) returns 0 (the returning 0 is reliable. For example in `init.g`, while parsing the body of `Error` on lines 57 to 66, no filename is returned some these lines are not marked as a line which was read during parsing, because there is no filename. I'm guessing this is now not a good way of finding the current filename.

In principle we could actually stuff the filename in `StatHeader` nowadays (now it's always 64-bits), at the moment we use 31 bits for the line (just because there are that many bits spare), could go 16 bits for line, 15 bits for filename (obviously making sure nothing crashes if there are too long files or too many opened files) -- not saying we have to, just mentioning it as an option.


```
// This function checks gets the filenameId of the current function.
static inline UInt getFilenameIdOfCurrentFunction(void)
{
    Obj func = CURR_FUNC();
    Obj body = BODY_FUNC(func);
    return GET_GAPNAMEID_BODY(body);
}
```

fingolfin:
We used to stuff the file id into the StatHeader but I *removed* it from there because it is totally superfluous -- it suffices to store it in the BodyHeader, which is what we do. I'll investigate why this patch breaks it, thanks

fingolfin:
OK, so I didn't realize that there is a call to `RegisterStatWithHook` in `NewStatOrExpr`. So it's just a matter of adjusting that. No problem.

fingolfin:
Hmmm, in one test it dies while saving a workspace. Unfortunately I was not yet able to reproduce this anywhere.

tst/testspecial/debug-var.g.out
```diff
@@ -118,32 +118,28 @@ brk_2>
 brk_2> # check coding of dvars
 brk_2> function() return unbound_higher; end;
 Error, Variable: <debug-variable-1-4> cannot be used here in
-  <corrupted statement>  called from
```
fingolfin, line 121: I think this is actually a bug fix :-). For a very fringe and remote bug, but still... I've always wondered where those "corrupted statement" came from, now it's clear: from the fact that we throw an error while in the middle of coding.

fingolfin:
Huh, back to `SaveSubObj: Assertion `0' failed`, I thought I had fixed that...

fingolfin:
@ChrisJefferson does this look OK to you now?

ChrisJefferson (approved):

PR #5293, opened 2022-12-20
BuildPackages.sh: tweak how GAP is invoked

### LINKED ISSUE #5286: problem with reading`.gap/gaprc`  from `bin/BuildPackages.sh`
The function `run_configure_and_make` in the script `bin/BuildPackages.sh` determines the name of the current package by starting GAP, reading the `PackageInfo.g` file, and then printing the value of `GAPInfo.PackageInfoCurrent.PackageName`.

Since #5182 got merged and thus in GAP 4.12.2, this GAP is started without the `-r` option, which means that the file `.gap/gaprc` gets read.
If this file contains a statement of the form `Print( "something\n" );` then `something` is taken as the package name.
A a consequence, several packages (`curlinterface`, `ferret`, `io`, ...) cannot be compiled in this situation.
(My `.gap/gaprc` contains such a `Print` statement.)

Do we perhaps need a way to enable the user's root directory but without reading `.gap/gaprc`?

fingolfin:
... to determine package names. Namely, add back the -r command line option for GAP we recently removed, and instead add --bare to achieve a similar effect: namely to allow starting GAP with the required packages smallgrp etc. not present in the pkg directory of the primary GAP root

Should resolve #5286

### DIFF
--- a/bin/BuildPackages.sh
+++ b/bin/BuildPackages.sh
@@ -194,7 +194,7 @@ run_configure_and_make() {
   then
     if grep Autoconf ./configure > /dev/null
     then
-      local PKG_NAME=$($GAP -q -T -A -M <<GAPInput
+      local PKG_NAME=$($GAP -q -T -A -r -M --bare <<GAPInput
 Read("PackageInfo.g");
 Print(GAPInfo.PackageInfoCurrent.PackageName);
 GAPInput


ThomasBreuer (approved):
Looks good.

PR #5297, opened 2022-12-23
Correctly encode function and filenames in JSON while profiling

### LINKED ISSUE #5259: Profiling outputs malformed JSON if function names contain special characters
`NameFunction` might contain special characters like `"` or `\`:
```gap
gap> func := x -> x;
function( x ) ... end
gap> SetNameFunction( func, "\"\\" );
gap> NameFunction( func );
"\"\\"
```

Real life application: In CAP we give functions telling names to simplify debugging, see https://github.com/homalg-project/CAP_project/blob/a7f6b2311a38ad3e941fa6a5aa6493b0415085fb/CAP/gap/InstallAdds.gi#L606 In these names we use the name of a CAP category, which sometimes contains `"`.

When profiling, GAP puts function names into JSON strings without escaping them: https://github.com/gap-system/gap/blob/5fa271513ff2fdd68be23724243ca354bdc4f92d/src/profile.c#L294-L297 Thus, a function name containing `"` produces a malformed output.

Possible implementation: According to https://stackoverflow.com/questions/3020094/how-should-i-escape-strings-in-json/3020108#3020108, only `"`, and `\` and control characters must be escaped, so this should not be too hard (and I guess one could also simply ignore control characters) but my GAP + C skills are not sufficient for implementing this cleanly :(

Another "solution" would of course be to simply forbid using `"` and `\` in function names.

ChrisJefferson:
Fixes #5259, correctly handing both function names and filenames into JSON correctly.

Technically this isn't complete, I don't encode non-printable characters, but I don't expect those to arise in practice.

This is tested in a branch of the profiling package, but I don't want to release a new version with those tests until this is in master (I'll make sure those tests don't run in 4.12, so the package won't break in stable GAP).


src/profile.c
```diff
@@ -239,6 +239,49 @@ static void CheckLeaveFunctionsAfterLongjmp(void)
     }
 }
 
+static Obj JsonEscapeString(Obj param)
```
fingolfin, line 242: Maybe add a comment saying what this is about?
```suggestion
// Escape a string for serialization in a JSON file, following the
// escaping rules laid out at json.org.
static Obj JsonEscapeString(Obj param)
```

src/profile.c
```diff
@@ -239,6 +239,49 @@ static void CheckLeaveFunctionsAfterLongjmp(void)
...
+        out[0] = '\\';                                                       \
+        out[1] = y;                                                          \
+        out += 2;                                                            \
+        break;
+            ESCAPE_CASE('\b', 'b');
+            ESCAPE_CASE('\t', 't');
+            ESCAPE_CASE('\n', 'n');
+            ESCAPE_CASE('\f', 'f');
+            ESCAPE_CASE('\r', 'r');
+#undef ESCAPE_CASE
```
fingolfin, line 273: Alternative solution, without macros:
```suggestion
        case '\b':
        case '\f':
        case '\n':
        case '\r':
        case '\t':
            out[0] = '\\';
            out[1] = u + ('a' - '\a');
            out += 2;
            break;
```

src/profile.c
```diff
@@ -239,6 +239,49 @@ static void CheckLeaveFunctionsAfterLongjmp(void)
     }
 }
 
+static Obj JsonEscapeString(Obj param)
+{
+    Int lenString = LEN_LIST(param);
+
+    // Massively over-long string
```
fingolfin, line 246: Maybe more helpful to actually spell out what's going on:
```suggestion
    // Allocate an output string with twice the size of the input
    // string, as in the worst case every single has to be escaped.
```

src/profile.c
```diff
@@ -239,6 +239,49 @@ static void CheckLeaveFunctionsAfterLongjmp(void)
...
+    Obj     copy = NEW_STRING(lenString * 2 + 2);
```
fingolfin, line 247: Why add 2, it seems to serve no point?
```suggestion
    Obj     copy = NEW_STRING(lenString * 2);
```

src/profile.c
```diff
@@ -239,6 +239,49 @@ static void CheckLeaveFunctionsAfterLongjmp(void)
...
+    UChar * in = CHARS_STRING(param);
+    UChar * base = CHARS_STRING(copy);
+    UChar * out = base;
+
+    for (Int i = 0; i <= lenString; ++i) {
```
fingolfin, line 252: That's the wrong bounds. Perhaps you did it to "cleverly" set the zero terminator, but I'd argue that'd be too clever for its own good. Plus it is not necessary as the newly allocated string is guaranteed to be zeroed.
```suggestion
    for (Int i = 0; i < lenString; ++i) {
```

fingolfin (changes requested):

src/profile.c
ChrisJefferson: wait, does this work? is this how they decided which character to call `\n`, `\t`, etc??

ChrisJefferson: No, I don't think it does, as `\n` is 10 and `\t` is 9 -- I don't think there is any way to map between `\n` and `n` without a hard-wired list somewhere (doesn't have to be in a macro of course)

fingolfin: Argh, indeed, that was utter nonsense on my part, sorry. I got confused by the mapping of Ctrl-A, Ctrl-B, ... etc. to "ascii" codes 1, 2, ....

ChrisJefferson:
I rebased and retested this, and I think it's fine for merging.

fingolfin (approved):

PR #5275, opened 2022-12-15
Remove tabs and unused local variables (DONE)

james-d-mitchell:
This PR removes all tabs and unused local variables from files starting "c" in the library. There were no tabs and no unused local variables in files starting "b" in the library, hence the lack of a PR for these.

## Text for release notes

None

PR #5298, opened 2022-12-24
Fix a rare  error message in isomorphism test

hulpke:
This fixes a bug reported by B. Sambale in the forum on 12/24/2022: When adding to the series, the relevant radical automorphisms were only computed in one case,  not both. This will lead to an error message if wrong.

It also contains updates to fitting free setup so that it will work for hybrid groups.


fingolfin:
The `testexpect` error is most likely unrelated, and appears elsewhere. I'll open an issue to deal with it.

Otherwise this PR looks good to me. However, I would prefer to first merge PR #5275; afterwards this and a a few other PRs may need to be adjusted, which I will take care of myself (i.e. no need for @hulpke to update this or the other PR), and then I hope we can merge both of your PRs soon.

Sorry for the delay.

fingolfin (approved):

fingolfin:
This PR apparently contains simultaneously a bug fix, a new feature and enhancements, thus would need multiple entries in the release notes and can't just be handled via the title. 

But since also the "release notes: use title" was set, I assume only the bugfix is meant to be in the release notes. I've thus adjusted the labels to only indicate the bug fix, so as to not mislead the script generating the release notes.

hulpke:
@fingolfin 
Yes, that is correct.
I'm sorry that the clustering of items causes a problem. If there wasn't the required approval, I would be most happy to avoid putting items together that way.

PR #5300, opened 2022-12-29
Make `Remove` with two arguments faster for internal lists

### LINKED ISSUE #767: Reduce performance penalty in `Remove` caused by method dispatch overhead for lists?
When fixing a bug in a kernel method for `Remove` I noticed the following:

```
gap> l := List([1..100000], i->[i]);;
gap> while Length(l)>0 do Remove(l); od; time;
8
gap> l := List([1..100000], i->[i]);;
gap> while Length(l)>0 do Remove(l,Length(l)); od; time;
134912
gap> l := List([1..100000], i->[i]);;
gap> while Length(l)>0 do Unbind(l[Length(l)]); od; time;
12
gap> l := List([1..100000], i->[i]);;
gap> met := ApplicableMethod(Remove, [l,1]);
function( l, p ) ... end
gap> while Length(l)>0 do met(l, Length(l)); od; time;
152
```

This shows that
- the fast kernel function (which had a bug for plists with holes) is not really needed, just use `Unbind`
- we have still not addressed the general problem that it is sometimes far too expensive to compute the type of a list (almost all the time in the second loop of the example is spent on this, although all the information which is expensive to gather is completely irrelevant here)

I don't see a quick and compatible way to improve the second loop. If we proceeded as in other cases (the last I remember was `Reversed`) we could change `Remove` to a function which catches the main cases of internal lists efficiently and calls a newly introduced operation `RemoveOp` for other cases. The disadvantage of this approach is that current method installations for `Remove` must be adjusted.

But maybe this is not worth the effort and the time could be spent better on addressing the problem in general?

Further remarks:
- if I understand the (complicated) mechanism with the current kernel operation correctly, then the method `met` in the fourth loop of the example is never called and could be removed
- the current kernel method for `Remove` with one argument shrinks the memory allocated for a list when its length becomes less than 3/4 of the available entries. Maybe this should be done by `UnbPlist` in general?

fingolfin:
Before:

    gap> l := List([1..100000], i->[i]);;
    gap> while Length(l)>0 do Remove(l,Length(l)); od; time;
    53937

After:

    gap> l := List([1..100000], i->[i]);;
    gap> while Length(l)>0 do Remove(l,Length(l)); od; time;
    17

Resolves #767

### DIFF
--- a/lib/list.gi
+++ b/lib/list.gi
@@ -1984,9 +1984,12 @@ InstallMethod(Remove, "one argument", [IsList and IsMutable],
     return x;
 end);
 
-InstallMethod(Remove, "two arguments, fast", [IsList and IsPlistRep and IsMutable, IsPosInt],
+InstallEarlyMethod(Remove,
         function(l,p)
     local ret,x,len;
+    if not (IsPlistRep(l) and IsMutable(l) and IsPosInt(p)) then
+        TryNextMethod();
+    fi;
     len := Length(l);
     ret := IsBound(l[p]);
     if ret then


ChrisJefferson:
The method used to reject immutable lists and now accepts them. This mostly just means the error occurs in a different place, but it does make some changes, for example `p := Immutable([1,2,3]); Remove(p,100);` now works, when it used to break.

This doesn't greatly make a difference, but probably best to explicitly reject immutable lists to `TryNextMethod`, just to avoid changing behaviour unnessarily?

fingolfin:
Thanks for the feedback! I've changed it to explicitly test for `IsMutable`. No point in bogging this patch down with changes to which methods are applied -- if we want to change that, it should be in a separate PR.

frankluebeck (approved):
Yes, `InstallEarlyMethod` looks like a good solution for this kind of performance problem.
The one argument version of `Remove` is already a kernel operation and again quite a bit faster than this early method.

PR #5301, opened 2022-12-29
Remove `ActivateProfileColour` (it was rarely used or even known, and blocked work on the profiling engine)

ChrisJefferson:
Part of #5261 , which removes a function which was previously tested in the `profiling` package.

### DIFF
--- a/src/profile.c
+++ b/src/profile.c
@@ -60,11 +60,7 @@
 **    We also use 1 bit to mark that we have executed this statement, so
 **    we can (if we want to) only output each executed statement once.
 **
-**    We use this information in a few ways:
-**
-**  1) Output a straight list of every executed statement
-**  2) Provide coloured printing (by wrapping PrintStatFuncs and PrintExprFuncs)
-**     which show which parts of an expression have been executed
+**  We use this information to output a straight list of every executed statement.
 **
 **  There are not that many tricky cases here. We have to be careful that
 **  some Expr are integers or local variables, and not touch those.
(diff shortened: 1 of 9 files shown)


src/profile.c
```diff
@@ -798,128 +797,6 @@ static Obj FuncIsLineByLineProfileActive(Obj self)
...
-static Obj activate_colored_output_from_profile(void)
-{
-    HashLock(&profileState);
-
-    if(profileState.ColouringOutput) {
-      HashUnlock(&profileState);
-      return Fail;
-    }
-
-    ActivatePrintHooks(&profilePrintHooks);
```
fingolfin, line 875: Is it worth it to keep the whole print hook machinery in place?

src/profile.c
```diff
@@ -83,9 +83,8 @@
 **     we provide -P (profiling) and -c (code coverage) options to the GAP
 **     executable so the user can start before code loading starts
 **
-**  3) Operating at just a line basis can sometimes be too course. We can
-**     see this when we use ActivateProfileColour. However, without some
-**     serious additional overheads, I can't see how to provide this
+**  3) Operating at just a line basis can sometimes be too course. However,
+**     without some serious additional overheads, I can't see how to provide this
```
fingolfin, line 87: There is another reference to colo(u)r in line 66. Also, the variable `ColouringOutput` is not actually removed. 

I am going to update this PR to fix this and my other remarks.

fingolfin (approved):

PR #5302, opened 2023-01-02
kernel: cleanup hookintrprtr and profile code

fingolfin:
- unify nameid, file, fileID, fileid -> fileid
- unify argument order: first fileid, then line
- remove unused return value from `ActivateHooks` and `DeactivateHooks`
- rename `HookCount` to `MAX_HOOK_COUNT`
- convert some `UInt` / `Int` values to just `int`
- use `BOOL`, `TRUE`, `FALSE` in more places
- change `outputStat` to take `fileid`, `line` and `type` as argument instead of `stat` (to enable future refactoring)

### DIFF
--- a/src/hookintrprtr.c
+++ b/src/hookintrprtr.c
@@ -25,7 +25,7 @@
 
 
 // List of active hooks
-struct InterpreterHooks * activeHooks[HookCount];
+struct InterpreterHooks * activeHooks[MAX_HOOK_COUNT];
 
 // Number of active hooks
 static Int HookActiveCount;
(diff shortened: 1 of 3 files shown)


fingolfin:
I wonder if there is anyone besides @ChrisJefferson willing and able to review changes to our C kernel. Perhaps @james-d-mitchell or @zickgraf would be willing to help out (I totally understand if you are more than busy enough with your actual work, though, so I have no expectations at all, but I figured it would be worth a shot... feel free to just completely ignore me )

james-d-mitchell:
I'm happy to do this in principal @fingolfin, I'm not completely sure that I'll have much to contribute to this particular PR, but I can try.

james-d-mitchell (commented):
I've had a quick look at this PR and it looks ok to me, although I don't exactly know what the code does, it seems most of the changes are harmless.  It looks like `Int` and `int` are used inconsistently, I'm not sure if that's really on purpose or not (for example, the loop variable `i` indexing into `activeHooks` is an `int` in some places, and an `Int` in others)

PR #5306, opened 2023-01-06
Stop using GNU libtool for linking GAP and libgap

### LINKED ISSUE #4682: buildsys: should we drop libtool and simply use our own list of build rules?

### LINKED ISSUE #5383: Libtool errors with recent libtool
When building with libtool 2.4.7 we saw the following error at configure time:

```
config.status: executing libtool commands
+ make
    * new C compiler flags
    * new preprocessor flags
libtool: Version mismatch error.  This is libtool 2.4.7, but the
libtool: definition of this LT_INIT comes from libtool 2.4.6.
libtool: You should recreate aclocal.m4 with macros from libtool 2.4.7
libtool: and run autoconf again.
```

For the conda-forge distribution, we fixed this with [this patch](https://github.com/conda-forge/gap-feedstock/pull/66/files#diff-c2aea2d513c0db971da5166d41ba7b6e1219265f0c54a4bcf7d8efa05fa6f058), i.e., `rm aclocal.m4` and use `AC_CONFIG_MAROS_DIR` to tell autoconf where it's macros can be found instead.

What is the reason for GAP's non-standard setup here?

fingolfin:
Resolves #4682
Resolves #5383

TODOs include at least:
- [x] Windows support
- [x] `make install` tweaks
- [ ] testing on some *BSD variants
- ...


james-d-mitchell:
The title of this PR is missing a "t", `libool` -> `libtool`.

dimpase:
just a note that there are packages which don't build on OpenBSD due to a libtool problem.

fingolfin:
This seems to be working now. What's not yet tested is anything other than Linux, macOS, Windows...

dimpase:
I'll test on OpenBSD and report.

dimpase (approved):
All is good on OpenBSD 7.3. GAP builds, semigroups package builds, most tests in `TestPackage("semigroups");`  pass.

fingolfin:
Thanks @dimpase !

ChrisJefferson (approved):

fingolfin:
I think this should be mentioned in the release notes as it is quite relevant for downstream packagers

fingolfin:
(well ideally it is transparent for them, but I think it's good to make sure they are aware as it could cut down debugging time if anything breaks, and they might be able to get rid of some libtool specific adjustments etc.)

fingolfin:
Started a run of the package distro test suite at https://github.com/gap-system/PackageDistro/actions/runs/7589464817

fingolfin:
That package distro CI run failed due to the scripts there not being ready to deal with branches containing a slash... So I made https://github.com/gap-system/PackageDistro/pull/877 and based on that started this test run which shows that all packages still pass with this PR here, yay!

https://github.com/gap-system/PackageDistro/actions/runs/7604428542/job/20707309882

PR #5307, opened 2023-01-06
get rid of `ImmutableVector` calls (WIP)

ThomasBreuer:
The idea is to use "example vectors" with the desired internal representation whenever this is possible.
Thus eventually only those (few) `ImmutableVector` calls will remain where GAP is expected to choose an appropriate internal representation. (And we will then see how this situation can be improved further.)


fingolfin:
This seems plausible to me

fingolfin:
@ThomasBreuer what to do with this PR? You opened it, but marked as WIP, so I never merged it, and then it seems to have been forgotten?

PR #5310, opened 2023-01-10
Address 'TODO is this line required'

fingolfin:


ThomasBreuer (approved):
My understanding of this pull request is as follows:
In Pull request #5275, assignments to local variables had been removed which were not used later on in the function. The function calls whose results had been assigned to these variables were kept, and marked with `# TODO is this line required`. Now the point is to remove these lines where appropriate, or to remove the comment where the line must be kept.

I do not see an immediate reason why the lines to be removed should be kept.
However, it is still possible that the code behaves differently after the removals, due to some attributes which will not be set from now on.

fingolfin:
> However, it is still possible that the code behaves differently after the removals, due to some attributes which will not be set from now on.

Indeed, there is always a chance of that. I had a look in each case and it *seemed* to me that this shouldn't happen, but of course it's almost impossible to say for certain.

My view is: only one way to find out ;-). If any of these *were* necessary, we can add them back, but this time with a comment explaining the need. And possibly rethinking the relevant part of the code to make it more robust against that kind of problem in the future...

PR #5324, opened 2023-01-16
kernel: merge GMP_REDUCE into GMP_NORMALIZE

fingolfin:
Old patch I wrote a couple years ago as part of another bigger change (which is still incomplete, but I figured this bit could be merged now...).

### DIFF
--- a/src/integer.c
+++ b/src/integer.c
@@ -62,7 +62,7 @@
 **  small integer). Internally, a large integer may temporarily not be
 **  normalized or not be reduced, but all kernel functions must make sure
 **  that they eventually return normalized and reduced values. The function
-**  GMP_NORMALIZE and GMP_REDUCE can be used to ensure this.
+**  GMP_NORMALIZE can be used to ensure this.
 */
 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
Can you say a few words about the rationale for this change?

fingolfin:
We always calls `GMP_NORMALIZE` then `GMP_REDUCE`. There are zero places left that only call one of them. And they must be called in the right order.

james-d-mitchell:
Ah ok, I see, I somehow failed to read the word "merge" in the title and commit message. This looks like a good change to me!

ChrisJefferson (approved):
Looks good to me, and neither function is used by any package, so it can't break anything else.

fingolfin:
> Looks good to me, and neither function is used by any package, so it can't break anything else.

Just to say, it was used by the `float` package for a long time, even for the one distributed with GAP 4.11 -- since then a patch of mine to remove its usage there was merged, which is why we can now finally apply this patch :-).

Issue #5322, opened 2023-01-13
Regression in `DeclareHandlingByNiceBasis`/`InstallHandlingByNiceBasis`


zickgraf:
The documentation of `DeclareHandlingByNiceBasis`/`InstallHandlingByNiceBasis` claims:

> name must be a string, a filter f with this name is created, and a logical implication from f to IsHandledByNiceBasis (61.11-6) is installed.

This is not true anymore since commit 2f4a518d36912392d7102c940873e77b2c076c23 (see https://github.com/gap-system/gap/commit/2f4a518d36912392d7102c940873e77b2c076c23#diff-14aba9b3712179af58d64fc6572ca6107a7c746245c8c795a98b79ec746676d4R690-L694).

This leads to an error when loading [QPA2](https://github.com/sunnyquiver/QPA2):
```
Error, required filters [
    "IsSubspaceOfQPAVectorSpace" 
]
for 1st argument do not match a declaration of NiceFreeLeftModule called from
<<compiled GAP code>> from GAPROOT/lib/oper1.g:336 in function InstallMethod called from
<function "InstallMethod">( <arguments> )
 called from read-eval loop at QPA2/lib/vecspace.gi:766
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
```

I'm not sure if the documentation or the regressing commit should be updated.

ThomasBreuer:
@zickgraf Yes, the new filter `f` is now just a flag such that `f and IsFreeLeftModule and IsAttributeStoringRep` implies `IsHandledByNiceBasis`. I think this is an improvement over the situation before #3006 got merged, and I am going to update the documentation accordingly.

The mentioned problem in the QPA2 package arises from the fact that the `NiceFreeLeftModule` method wants to be applicable as soon as the argument is in `IsSubspaceOfQPAVectorSpace`, which does no longer imply `IsFreeLeftModule`.
(Most of the vector spaces types in the GAP library use the generic `NiceFreeLeftModule` method, that's why I had overlooked this problem.)
If it is acceptable to adjust the QPA2 code then I would suggest to change the `InstallMethod` call such that `IsFreeLeftModule and IsSubspaceOfQPAVectorSpace` is required.

zickgraf:
Fixed by https://github.com/gap-system/gap/pull/5336.

PR #5325, opened 2023-01-16
fix the documentation of `DeclareHandlingByNiceBasis`

ThomasBreuer:
Since pull request #3006 got merged, the statement about the implication installed by `InstallHandlingByNiceBasis` was not correct.

This change addresses issue #5322.

### DIFF
--- a/lib/basis.gd
+++ b/lib/basis.gd
@@ -787,8 +787,9 @@ DeclareOperation( "RelativeBasisNC", [ IsBasis, IsHomogeneousList ] );
 ##  <P/>
 ##  <A>name</A> must be a string,
 ##  a filter <M>f</M> with this name is created, and
-##  a logical implication from <M>f</M> to <Ref Filt="IsHandledByNiceBasis"/>
-##  is installed.
+##  a logical implication from the join of <M>f</M> with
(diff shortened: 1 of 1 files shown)


fingolfin (commented):
Technically that means our changes to `DeclareHandlingByNiceBasis` were breaking.

To me that indicates we should...
- carefully consider if we *really* want to introduce this breakage, or perhaps instead revert it (it may be "right", but is it worth the trouble? )
- if we do, then we should document it explicitly as a breaking change in the release notes
- ... and also check for affected packages. A quick grep show that it is used in `fr`, `guava`, `quagroup`, `qpa`.

ThomasBreuer:
@fingolfin The problem described in issue #5322 arises because the QPA2 package wants to install a `NiceFreeLeftModule` method with requirement only the filter that has been created by `DeclareHandlingByNiceBasis`. The mentioned packages `fr`, `guava`, `quagroup`, `qpa` do not install `NiceFreeLeftModule` methods.

I could modify the code such that the filters created by `DeclareHandlingByNiceBasis` again imply `IsFreeLeftModule`, then the problem from issue #5322 would disappear. The problem to be fixed was the implication to `IsAttributeStoringRep`.
(In general, I think it is not a good idea to "waste" a new simple filter by automatically restricting its context via implications.)

fingolfin (approved):

PR #5332, opened 2023-01-19
`IsVecOrMatObj` does no longer imply `IsCopyable`

ThomasBreuer:
The filter `IsVecOrMatObj` is not documented.
The change means that `IsVectorObj` and `IsMatrixObj` objects are not automatically in `IsCopyable`.

The idea is that each representation of such objects (such as `IsPlistVectorRep`) decides whether its objects admit mutable variants or not,
by installing an implication to `IsCopyable` in the former case.

The currently available "constructing filters" of vector and matrix objects in the GAP library behave like this, and their documentation states that the objects are in fact copyable.

### DIFF
--- a/lib/matobj1.gd
+++ b/lib/matobj1.gd
@@ -26,7 +26,7 @@
 # IsMatrixObj, then the first definition incurs a "hidden implication"
 # which then later leads to "method matches more than one declaration" messages.
 # The proper fix is to remove hidden implication
-DeclareCategory( "IsVecOrMatObj", IsCopyable );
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
We discussed this and it logically makes sense, and is needed to support things like our block matrices which are read-only.

I am very slightly worried about the fallout, as this changes ranks, applicability of certain methods etc. -- but that's not a reason against it, in the end, we just have to hope that our test suite plus that of our packages will catch any regressions, and then deal with them. (We did have a preliminary look at packages and it *seems* as if all should be fine, but you never can be sure 100%).

PR #5333, opened 2023-01-19
Introduce MakeIsPlistVectorRep helper

### LINKED ISSUE #5330: Problem with `IsPlistVectorRep` and `IsPlistMatrixRep`
The current implementation of `IsPlistVectorRep` and `IsPlistMatrixRep` must be revised.
The idea to use the type of an argument as the type of the result (and perhaps setting `IsMutable` or calling `MakeImmutable`) cannot work:
```
gap> z:= Immutable( ZeroVector( Integers, 1 ) );;
gap> IsZero( z );  HasIsZero( z );  # the type of z is correct
true
true
gap> v:= Vector( [ 1 ], z );;
gap> HasIsZero( v );  IsZero( v );  # the type of z is used as the type of v
true
true
gap> Print( v );
NewVector(IsPlistVectorRep,Integers,[ 1 ])
```
The `Vector( v1, v2 )` method creates a new object from the values of `v1` and the type of `v2`.
However, the type of `v2` may store flags that are not true for the intended result.
I think that we have to take a designated type for the result. In order not to go through `NewType` for each call (cf. `MakePlistVectorType`), we can cache the types in question: one type for immutable objects and one for mutable objects (if mutable variants of the given type are supported at all).
The question is where these types shall be cached, in the base domain or in its family; fetching the right type should be not much slower than the current `TypeObj` call plus a subsequent `SetFilterObj` call.

fingolfin:
This centralizes the creation of `IsPlistVectorRep` instances, reducing code duplication, and enabling future refactoring.

It also fixes bugs caused by incorrect reusing of the TypeObj of a vector to create another vector; this introduces bugs because the type can store things like the value of `IsZero` which must not be carelessly copied over to a new vector with different content.

Resolves #5330

TODO:
- [ ] set up benchmarks to determine the performance impact this has
- [ ] ... and then possibly work on reducing that, e.g. by caching types
- [ ] do something similar for `IsPlistMatrixRep`
- [ ] do something similar for `IsZmodnZVectorRep`
- [ ] do something similar for `IsZmodnZMatrixRep`

Ideally for each of the four types, only a single place calling `Objectify` for them would be left. This would then make further refactoring of their internal layout easy (for PR #5217)

### DIFF
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -18,48 +18,77 @@
 # Constructors:
 ############################################################################
 
-InstallGlobalFunction( MakePlistVectorType,
-  function( basedomain, filter )
-    local T, filter2;
-    filter2 := filter and IsMutable;
-    if HasCanEasilyCompareElements(Representative(basedomain)) and
-       CanEasilyCompareElements(Representative(basedomain)) then
-        filter2 := filter2 and CanEasilyCompareElements;
-    fi;
-    if IsIdenticalObj(basedomain,Integers) then
-        T := NewType(FamilyObj(basedomain),
-                       filter2 and IsIntVector);
-    elif IsFinite(basedomain) and IsField(basedomain) then
-        T := NewType(FamilyObj(basedomain),
-                       filter2 and IsFFEVector);
+BindGlobal( "MakeIsPlistVectorRep",
+  function( basedomain, list )
+    local fam, types, typ;
+    fam := FamilyObj(basedomain);
+    #types := _PlistVectorRepTypeCache(basedomain);
+
+    # special case: integers
+    if IsIntegers(basedomain) then
+        if not IsBound(basedomain!.PlistVectorRepTypes) then
+            # initialize type cache
+            # TODO: make this thread safe for HPC-GAP
+            basedomain!.PlistVectorRepTypes := [
+                NewType(fam, IsPlistVectorRep and IsIntVector and CanEasilyCompareElements),
+                NewType(fam, IsPlistVectorRep and IsIntVector and CanEasilyCompareElements and IsMutable),
+            ];
+        fi;
+        types := basedomain!.PlistVectorRepTypes;
+    elif IsFFECollection(basedomain) then
+        if not IsBound(basedomain!.PlistVectorRepTypes) then
+            # initialize type cache
+            # TODO: make this thread safe for HPC-GAP
+            basedomain!.PlistVectorRepTypes := [
+                NewType(fam, IsPlistVectorRep and IsFFEVector and CanEasilyCompareElements),
+                NewType(fam, IsPlistVectorRep and IsFFEVector and CanEasilyCompareElements and IsMutable),
+            ];
+        fi;
+        types := basedomain!.PlistVectorRepTypes;
     else
-        T := NewType(FamilyObj(basedomain),
-                       filter2);
+        if not IsBound(fam!.PlistVectorRepTypes) then
+            # initialize type cache
+            # TODO: make this thread safe for HPC-GAP
+            fam!.PlistVectorRepTypes := [
+                NewType(fam, IsPlistVectorRep),
+                NewType(fam, IsPlistVectorRep and IsMutable),
+            ];
+            fam!.PlistVectorRepTypesEasyCompare := [
+                NewType(fam, IsPlistVectorRep and CanEasilyCompareElements),
+                NewType(fam, IsPlistVectorRep and CanEasilyCompareElements and IsMutable),
+            ];
+        fi;
+        if HasCanEasilyCompareElements(Representative(basedomain)) and
+           CanEasilyCompareElements(Representative(basedomain)) then
+            types := fam!.PlistVectorRepTypesEasyCompare;
+        else
+            types := fam!.PlistVectorRepTypes;
+        fi;
     fi;
-    return T;
-  end);
+    if IsMutable(list) then
+        typ := types[2];
+    else
+        typ := types[1];
+    fi;
+
+    return Objectify(typ, [ basedomain, list ]);
+  end );
 
 InstallMethod( NewVector, "for IsPlistVectorRep, a ring, and a list",
   [ IsPlistVectorRep, IsRing, IsList ],
   function( filter, basedomain, l )
-    local typ, v;
     if ValueOption( "check" ) <> false and not IsSubset( basedomain, l ) then
       Error( "the elements in <l> must lie in <basedomain>" );
     fi;
-    typ := MakePlistVectorType(basedomain,IsPlistVectorRep);
-    v := [basedomain,ShallowCopy(l)];
-    Objectify(typ,v);
-    return v;
+    return MakeIsPlistVectorRep(basedomain, ShallowCopy(l));
   end );
 
 InstallMethod( NewZeroVector, "for IsPlistVectorRep, a ring, and an int",
   [ IsPlistVectorRep, IsRing, IsInt ],
-  function( filter, basedomain, l )
-    local typ, v;
-    typ := MakePlistVectorType(basedomain,IsPlistVectorRep);
-    v := [basedomain,Zero(basedomain)*[1..l]];
-    Objectify(typ,v);
-    return v;
+  function( filter, basedomain, len )
+    local list;
+    list := ListWithIdenticalEntries(len, Zero(basedomain));
+    return MakeIsPlistVectorRep(basedomain, list);
   end );
 
 InstallMethod( NewMatrix,
@@ -213,46 +242,34 @@ InstallMethod( Length, "for a plist vector", [ IsPlistVectorRep ],
 
 InstallMethod( ZeroVector, "for an integer and a plist vector",
   [ IsInt, IsPlistVectorRep ],
-  function( l, t )
-    local v;
-    v := Objectify(TypeObj(t),
-                   [t![BDPOS],ListWithIdenticalEntries(l,Zero(t![BDPOS]))]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+  function( l, v )
+    return NewZeroVector(IsPlistVectorRep, v![BDPOS], l);
   end );
 
 InstallMethod( ZeroVector, "for an integer and a plist matrix",
   [ IsInt, IsPlistMatrixRep ],
   function( l, m )
-    local v;
-    v := Objectify(TypeObj(m![EMPOS]),
-                   [m![BDPOS],ListWithIdenticalEntries(l,Zero(m![BDPOS]))]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+    return NewZeroVector(IsPlistVectorRep, m![BDPOS], l);
   end );
 
 InstallMethod( Vector, "for a plain list and a plist vector",
   [ IsList and IsPlistRep, IsPlistVectorRep ],
-  function( l, t )
-    local v;
-    v := Objectify(TypeObj(t),[t![BDPOS],l]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+  function( l, v )
+    # wrap the given list without copying it (this is documented behavior)
+    return MakeIsPlistVectorRep(v![BDPOS], l);
+    # TODO: optionally check that l is consistent with t
   end );
 
 InstallMethod( Vector, "for a list and a plist vector",
   [ IsList, IsPlistVectorRep ],
-  function( l, t )
-    local v;
-    v := ShallowCopy(l);
-    if IsGF2VectorRep(l) then
-        PLAIN_GF2VEC(v);
-    elif Is8BitVectorRep(l) then
-        PLAIN_VEC8BIT(v);
+  function( l, v )
+    local m;
+    m := IsMutable(l);
+    l := PlainListCopy(l);
+    if not m then
+        MakeImmutable(l);
     fi;
-    v := Objectify(TypeObj(t),[t![BDPOS],v]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+    return MakeIsPlistVectorRep(v![BDPOS], l);
   end );
 
 # compatibility method for older representations as list of elements
@@ -282,7 +299,7 @@ InstallMethod( \[\]\:\=, "for a plist vector, a positive integer, and an obj",
 InstallMethod( \{\}, "for a plist vector and a list",
   [ IsPlistVectorRep, IsList ],
   function( v, l )
-    return Objectify(TypeObj(v),[v![BDPOS],v![ELSPOS]{l}]);
+    return MakeIsPlistVectorRep(v![BDPOS], v![ELSPOS]{l});
   end );
 
 InstallMethod( PositionNonZero, "for a plist vector", [ IsPlistVectorRep ],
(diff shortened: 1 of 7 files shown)


lib/matobjplist.gi
```diff
@@ -464,71 +432,61 @@ InstallOtherMethod( MultVectorLeft, "for an integer vector, and a small integer"
...
 
 InstallMethod( ZeroMutable, "for a plist vector", [ IsPlistVectorRep ],
   function( v )
-    local res;
-    res := Objectify(TypeObj(v),
-                     [v![BDPOS],ZeroMutable(v![ELSPOS])]);
-    if not IsMutable(v) then SetFilterObj(res,IsMutable); fi;
-    return res;
+    return MakeIsPlistVectorRep(v![BDPOS], true,
+            ZeroSameMutability(v![ELSPOS]));
```
ThomasBreuer, line 489: `ZeroMutable`?

ThomasBreuer (approved):
Thanks, this fixes the problem which type is taken for the results, and simplifies further improvements.
I have added a few comments.
I think this pull request should be merged soon, afterwards we can address the open questions about caching, the mutability of results, and necessary consistency checks.

james-d-mitchell:
Wouldn't MakePlistVectorRep be a possibly better name?

lib/matobjplist.gi
```diff
@@ -18,48 +18,45 @@
...
-  function( basedomain, filter )
-    local T, filter2;
-    filter2 := filter and IsMutable;
+BindGlobal( "MakeIsPlistVectorRep",
+  function(basedomain, mutable, list )
+    local filter, typ;
+    # TODO: cache filter and type in the family obj;
+    # we need variants for mutable vs. immutable, and also for
+    # CanEasilyCompareElements true/false; we might reduce this
+    # by using CanEasilyCompareElementsFamily
```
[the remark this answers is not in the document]
fingolfin: The base domain may not be a component object, so it is not clear how to store something in it in general.

Also, the type itself does not depend on the basedomain at all, so it seems sensibel to share it between different basedomains in the same family?

lib/matobjplist.gi
```diff
@@ -213,46 +210,28 @@ InstallMethod( Length, "for a plist vector", [ IsPlistVectorRep ],
...
 InstallMethod( Vector, "for a plain list and a plist vector",
   [ IsList and IsPlistRep, IsPlistVectorRep ],
-  function( l, t )
-    local v;
-    v := Objectify(TypeObj(t),[t![BDPOS],l]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+  function( l, v )
+    # wrap the given list without copying it (this is documented behavior)
+    return MakeIsPlistVectorRep(v![BDPOS], true, l);
```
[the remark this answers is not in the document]
fingolfin: Ah, OK, so this then?
```suggestion
    return MakeIsPlistVectorRep(v![BDPOS], IsMutable(l), l);
```

lib/matobjplist.gi
```diff
@@ -464,71 +432,61 @@ InstallOtherMethod( MultVectorLeft, "for an integer vector, and a small integer"
...
 InstallMethod( AdditiveInverseImmutable, "for a plist vector",
   [ IsPlistVectorRep ],
   function( v )
-    local res;
-    res := Objectify( TypeObj(v),
-       [v![BDPOS],AdditiveInverseSameMutability(v![ELSPOS])] );
-    MakeImmutable(res);
-    return res;
+    return MakeIsPlistVectorRep(v![BDPOS], false,
+            AdditiveInverseSameMutability(v![ELSPOS]));
```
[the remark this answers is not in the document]
fingolfin: I was careless and did not notice that it called `AdditiveInverseSameMutability`. Thank you for checking closely! So how about this:
```suggestion
            AdditiveInverseImmutable(v![ELSPOS]));
```

lib/matobjplist.gi
```diff
@@ -464,71 +432,61 @@ InstallOtherMethod( MultVectorLeft, "for an integer vector, and a small integer"
...
 InstallMethod( AdditiveInverseMutable, "for a plist vector",
   [ IsPlistVectorRep ],
   function( v )
-    local res;
-    res := Objectify(TypeObj(v),
-                     [v![BDPOS],AdditiveInverseMutable(v![ELSPOS])]);
-    if not IsMutable(v) then SetFilterObj(res,IsMutable); fi;
-    return res;
+    return MakeIsPlistVectorRep(v![BDPOS], true,
+            AdditiveInverseSameMutability(v![ELSPOS]));
```
fingolfin, line 471:
```suggestion
            AdditiveInverseMutable(v![ELSPOS]));
```

lib/matobjplist.gi
```diff
@@ -464,71 +432,61 @@ InstallOtherMethod( MultVectorLeft, "for an integer vector, and a small integer"
...
   end );
 
 InstallMethod( ZeroImmutable, "for a plist vector", [ IsPlistVectorRep ],
   function( v )
-    local res;
-    res := Objectify(TypeObj(v),[v![BDPOS],ZeroImmutable(v![ELSPOS])]);
-    MakeImmutable(res);
-    return res;
+    return MakeIsPlistVectorRep(v![BDPOS], false,
+            ZeroSameMutability(v![ELSPOS]));
```
fingolfin, line 483:
```suggestion
            ZeroImmutable(v![ELSPOS]));
```

lib/matobjplist.gi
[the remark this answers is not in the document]
ThomasBreuer: Yes.

lib/matobjplist.gi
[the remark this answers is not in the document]
ThomasBreuer: Yes.

lib/matobjplist.gi
ThomasBreuer: Right.

lib/matobjplist.gi
ThomasBreuer: Right.

ThomasBreuer:
@james-d-mitchell concerning `MakePlistVectorRep` vs. `MakeIsPlistVectorRep`:
I am not sure what is preferable in the documentation: I feel that talking about "matrix objects" is not precise enough, "objects in the filter `IsMatrixObj`" is precise but long, "`IsMatrixObj` objects" is shorter but looks strange (like `MakeIsPlistVectorRep`), "`MatrixObj` objects" looks less strange but is not correct.

lib/matobjplist.gi
```diff
@@ -282,7 +261,7 @@ InstallMethod( \[\]\:\=, "for a plist vector, a positive integer, and an obj",
 InstallMethod( \{\}, "for a plist vector and a list",
   [ IsPlistVectorRep, IsList ],
   function( v, l )
-    return Objectify(TypeObj(v),[v![BDPOS],v![ELSPOS]{l}]);
+    return MakeIsPlistVectorRep(v![BDPOS], true, v![ELSPOS]{l});
```
james-d-mitchell, line 264: And this one...

lib/matobjplist.gi
```diff
@@ -329,10 +308,7 @@ InstallMethod( Unpack, "for a plist vector",
 
 InstallMethod( ShallowCopy, "for a plist vector", [ IsPlistVectorRep ],
   function( v )
-    local res;
-    res := Objectify(TypeObj(v),[v![BDPOS],ShallowCopy(v![ELSPOS])]);
-    if not IsMutable(v) then SetFilterObj(res,IsMutable); fi;
-    return res;
+    return MakeIsPlistVectorRep(v![BDPOS], true, ShallowCopy(v![ELSPOS]));
```
james-d-mitchell, line 311: Doesn't `ShallowCopy` return `v` when `v` is immutable, and only return a copy when `v` is mutable?

ThomasBreuer: If `v` is in the filter `IsCopyable` then `ShallowCopy( v )` returns always a new object which is mutable.
If `v` is not in the filter `IsCopyable` then `ShallowCopy( v )` returns `v`.

james-d-mitchell: Thanks @ThomasBreuer I didn't know this!

ThomasBreuer: see https://docs.gap-system.org/doc/ref/chap12_mj.html#X846BC7107C352031

(`IsCopyable` seems to be a not well-known concept in GAP. Perhaps the documentation should better explicitly mention `IsCopyable` instead of talking about "copyable objects" or "objects that support a mutable form".)

lib/matobjplist.gi
james-d-mitchell: Sorry please ignore

lib/matobjplist.gi
```diff
@@ -213,46 +210,28 @@ InstallMethod( Length, "for a plist vector", [ IsPlistVectorRep ],
...
-    if IsGF2VectorRep(l) then
-        PLAIN_GF2VEC(v);
-    elif Is8BitVectorRep(l) then
-        PLAIN_VEC8BIT(v);
-    fi;
-    v := Objectify(TypeObj(t),[t![BDPOS],v]);
-    if not IsMutable(v) then SetFilterObj(v,IsMutable); fi;
-    return v;
+  function( l, v )
+    return MakeIsPlistVectorRep(v![BDPOS], true, PlainListCopy(l));
```
james-d-mitchell, line 234: Shouldn't this `true` also be `IsMutable(l)`?

lib/matobjplist.gi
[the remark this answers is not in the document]
ThomasBreuer: In the current example, the type depends on whether the base domain is `Integers` or not, and one can think about similar situations where special types for vectors/matrices over prime fields are of interest.
We can decide that we do not support this feature.

If we want to support this feature then the "default vector/matrix object types" that should be cached for objects in a given family depend on the constructing filter, the base domain, and the mutability.
Caching these types in the family is possible, but if we can cache them in the base domain then this is simpler. For example, a cache in the family should be "weak" in the sense that entries can be deleted whose base domain parts are not referenced in other places.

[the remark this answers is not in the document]
fingolfin: But the filter and the type do not reference the basedomain at all. So I don't see why we need them to be "weak". Yes, *in theory* we might need something more general, but I'd rather solve that problem then, instead of trying to anticipate the general case now and implementing a solution for that (currently) hypothetical scenario -- in my experience doing that just leads to complexity that is not actually helpful when the real thing happens.

So instead, the caching I now pushed handles integers and finite fields separately (storying the types for those in the domains). We can easily do this *differently* later when the need arises, but for now it seems to work well. I'll provide some benchmark results below.

lib/matobjplist.gi
fingolfin: Indeed, thanks. I've pushed a fixup addressing this (will clean up the history later)

lib/matobjplist.gi
ThomasBreuer: Yes.
In fact, the documentation of `\{\}` for vector objects does currently not state that the result shall be mutable if possible, but I think this should be added to the documentation.
(For matrix objects, there is already such a statement.)

fingolfin:
Addressed comments, added benchmarks, added caching.

Benchmark results *before* the changes in this PR (beware, these timings can fluctuate quite a bit):
```
+---------------+
| Testing GF(2) |
+---------------+
Testing ListWithIdenticalEntries:
  8 µs per iteration; 123954 iterations per second; (200 iterations)

+-------------------+
| Testing Rationals |
+-------------------+
Testing ListWithIdenticalEntries:
  6 µs per iteration; 158724 iterations per second; (200 iterations)

+--------------------------------+
| Testing GF(2) IsPlistVectorRep |
+--------------------------------+
Testing NewZeroVector( filt, R, n ):
  24 µs per iteration; 41748 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  25 µs per iteration; 40334 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  8 µs per iteration; 117861 iterations per second; (200 iterations)

+----------------------------------+
| Testing integer IsPlistVectorRep |
+----------------------------------+
Testing NewZeroVector( filt, R, n ):
  16 µs per iteration; 62270 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  17 µs per iteration; 57292 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  8 µs per iteration; 120867 iterations per second; (200 iterations)

+-----------------------------------+
| Testing rational IsPlistVectorRep |
+-----------------------------------+
Testing NewZeroVector( filt, R, n ):
  16 µs per iteration; 64371 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  16 µs per iteration; 62485 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  8 µs per iteration; 122187 iterations per second; (200 iterations)
```


After the changes in this PR (so with caching):
```
+---------------+
| Testing GF(2) |
+---------------+
Testing ListWithIdenticalEntries:
  6 µs per iteration; 164389 iterations per second; (200 iterations)

+-------------------+
| Testing Rationals |
+-------------------+
Testing ListWithIdenticalEntries:
  6 µs per iteration; 173843 iterations per second; (200 iterations)

+--------------------------------+
| Testing GF(2) IsPlistVectorRep |
+--------------------------------+
Testing NewZeroVector( filt, R, n ):
  10 µs per iteration; 99948 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  12 µs per iteration; 85477 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  11 µs per iteration; 93609 iterations per second; (200 iterations)

+----------------------------------+
| Testing integer IsPlistVectorRep |
+----------------------------------+
Testing NewZeroVector( filt, R, n ):
  10 µs per iteration; 103500 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  11 µs per iteration; 95228 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  11 µs per iteration; 95001 iterations per second; (200 iterations)

+-----------------------------------+
| Testing rational IsPlistVectorRep |
+-----------------------------------+
Testing NewZeroVector( filt, R, n ):
  11 µs per iteration; 87841 iterations per second; (200 iterations)
Testing ZeroVector( filt, R, n ):
  12 µs per iteration; 81375 iterations per second; (200 iterations)
Testing ZeroVector( n, M ):
  12 µs per iteration; 81043 iterations per second; (200 iterations)
```

So the case were an "example" vector was given still is a bit slower. On the upside, the case were no example vector was given got faster.


I am going to experiment a bit with using an operation to determine where to store the cache and for initializing it, so that it becomes easier to extend this for other cases, but this might be too expensive, we'll see. But first I have to teach now :-)

fingolfin:
BTW, I used `ListWithIdenticalEntries` for comparing to "old-style" vector, but perhaps more realistically would be to follow that with a call to `ConvertToVectorRepNC`. In that case it gets quite a bit slower:
```
+---------------+
| Testing GF(2) |
+---------------+
Testing ListWithIdenticalEntries:
  24 µs per iteration; 42544 iterations per second; (200 iterations)

+-------------------+
| Testing Rationals |
+-------------------+
Testing ListWithIdenticalEntries:
  12 µs per iteration; 82883 iterations per second; (200 iterations)
```

lib/matobjplist.gi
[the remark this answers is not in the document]
ThomasBreuer: What I meant was that if the family caches some type that depends on the base domain then the *key* for the entry involves the base domain.

Thanks for updates.

lib/matobjplist.gi
fingolfin: Yeah, I've been working on GAP for years but only realized in 2022 (when Thomas told me) what it really does; I somehow never had read its documentation entry. It really means `IsAbleToMakeMutableInstances` or `CanHaveMutableInstances` so (I am not proposing a renaming, just sayin')

fingolfin:
Re `MakePlistVectorRep` vs. `MakeIsPlistVectorRep` -- I really don't care either way, this is a purely internal helper, nobody should ever call it...

lib/matobjplist.gi
fingolfin: I've removed the `mutable` argument for `MakeIsPlistVectorRep` which should take care of this and other issues.

fingolfin:
I've cleaned this up, rebased it, and marked it as "ready for review": while we still should do a similar changes for the `IsPlistMatrixRep`, `IsZmodnZVectorRep`, `IsZmodnZMatrixRep`, merging this ASAP would allow @ThomasBreuer to continue working on that code

PR #5338, opened 2023-01-23
changed `ExtendRootDirectories` such that each added path ends with a slash `/`

ThomasBreuer:
The function `ExtendRootDirectories` takes a list of strings as its argument, and stores them in `GAPInfo.RootPaths`.
Up to now, these paths had to end with a slash `/`; if not then `DirectoriesLibrary` misunderstood them, in the sense that `DirectoriesLibrary` just concatenates the root path with the given directory name, and if the result is not a valid path then this root path gets ignored.

Now `ExtendRootDirectories` adds a slash to the end of the paths if it is missing.

(Once we are thinking about it: `ExtendRootDirectories` could require that the given paths exist, and it could apply `UserHomeExpand` to them. However, one can think of situations where such features are unwanted.)

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -1793,8 +1793,16 @@ InstallGlobalFunction( SetPackagePath, function( pkgname, pkgpath )
 #F  ExtendRootDirectories( <paths> )
 ##
 InstallGlobalFunction( ExtendRootDirectories, function( rootpaths )
+    local i;
+
     rootpaths:= Filtered( rootpaths, path -> not path in GAPInfo.RootPaths );
     if not IsEmpty( rootpaths ) then
+      # 'DirectoriesLibrary' concatenates root paths with directory names.
+      for i in [ 1 .. Length( rootpaths ) ] do
+        if not EndsWith( rootpaths[i], "/" ) then
+          rootpaths[i]:= Concatenation( rootpaths[i], "/" );
+        fi;
+      od;
       # Append the new root paths.
       GAPInfo.RootPaths:= Immutable( Concatenation( GAPInfo.RootPaths,
           rootpaths ) );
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

fingolfin:
I would argue this is a bug fix, not an enhancement: nothing in the documentation for `ExtendRootDirectories` suggests that trailing slashes are necessary, and I find it quite surprising, as it is e.g. not necessary when using the `-l PATHLIST` command line option.

cdwensley:
Maybe I should add here what I have already said to Thomas.  Where should this command be used, and should that be stated in the documentation?  I have used it in my gap.ini, and it worked.  Thomas advised that it should better be in gaprc.  I tried that, and it did not work - i.e. LoadPackage reported that the package requested was not available.

ThomasBreuer:
Concerning the question where `ExtendRootDirectories` should be used:
- The [documentation for the user's `gap.ini` file](https://docs.gap-system.org/doc/ref/chap3_mj.html#X87DF11C885E73583) states that this file should only set user preferences.
- `ExtendRootDirectories` was written for the situation that one calls it during a GAP session, and then loads some packages in the additional directories by hand. Calling the  function already in `.gap/gap.ini` is perhaps dangerous because the function calls `InitializePackagesInfoRecords` before this is expected to happen. (I have to check whether this call is safe.)
- For me, adding `ExtendRootDirectories` and corresponding `LoadPackage` calls to my `.gap/gaprc` file works.
  However, it is *not* possible to use an `ExtendRootDirectories` call in the `.gap/gaprc` file in order to achieve that packages in the additional directories are autoloaded. This is because the autoloading of packages happens before `.gap/gaprc` is read.

cdwensley:
Many apologies - sorry to waste everyone's time. 
The file was in the correct place, but I had not adjusted the settings of TextEdit.

PR #5339, opened 2023-01-24
Ensure `IsomorphismPermGroup(G)` always returns an isomorphism for `G` (before if `G` was a finite matrix group it sometimes returned an isomorphism for a general linear group containing `G`)

### LINKED ISSUE #3601: problem with IsomorphismPermGroup
Graham Ellis has sent a bug report to GAP support.
```
>  ┌───────┐   GAP 4.10.2 of 19-Jun-2019
>  │  GAP  │   https://www.gap-system.org
>  └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv3
>  Configuration:  gmp 6.1.2, readline
>  Loading the library and packages ...
>  Packages:   GAPDoc 1.6.2, PrimGrp 3.3.2, SmallGrp 1.3, TransGrp 2.0.4
>  Try '??help' for help. See also '?copyright', '?cite' and '?authors'
> gap> G:=SL(2,3);;Order(G);
> 24
> gap> P:=Image(IsomorphismPermGroup(G));;Order(P);
> 48
```
In the current master branch, this bug may show up or not, depending on the loaded packages;
without packages (via the `-A` option) the bug is present.

What happens when one calls `IsomorphismPermGroup( SL(2,3) )` is the following.
- In GAP 4.9 and 4.10.2, the method for matrix groups (from `grpmat.gi`) has the highest rank.
  The code of this method differs in the two versions.
  The GAP 4.10.2 variant replaces the given group by the full GL overgroup, which results in the wrong result.
  The GAP 4.9 variant does not (yet) contain such a replacement, this GAP version returns a correct result.
- In the master branch, the method for niceomorphisms (from `grpnice.gi`) can have the highest rank, provided that some packages are loaded, and this method restricts the mapping to the given group, as required; in this case, the bug is not present.

In both cases, one gets the following in the current master branch. 
```
gap> meth:= ApplicableMethod( IsomorphismPermGroup, [ SL(2,3) ], 0, "all" );;
gap> Set( List( meth{ [ 1, 2 ] }, f -> Size( Image( f( SL(2,3) ) ) ) ) );
[ 24, 48 ]
```

fingolfin:
Fixes #3601

### DIFF
--- a/lib/grpmat.gi
+++ b/lib/grpmat.gi
@@ -556,10 +556,6 @@ function(G)
 local map;
   if HasNiceMonomorphism(G) and IsPermGroup(Range(NiceMonomorphism(G))) then
     map:=NiceMonomorphism(G);
-    if IsIdenticalObj(Source(map),G) then
-      return map;
-    fi;
-    return GeneralRestrictedMapping(map,G,NiceObject(G));
   else
     if not HasIsFinite(G) then
       Info(InfoWarning,1,
@@ -567,8 +563,11 @@ local map;
     fi;
     map:=NicomorphismOfGeneralMatrixGroup(G,false,false);
     SetNiceMonomorphism(G,map);
+  fi;
+  if IsIdenticalObj(Source(map),G) then
     return map;
   fi;
+  return GeneralRestrictedMapping(map,G,NiceObject(G));
 end);
 
 InstallMethod( IsomorphismPermGroup,
--- a/lib/semitran.gi
+++ b/lib/semitran.gi
@@ -119,12 +119,12 @@ InstallMethod(IsomorphismPermGroup,
 function( h )
   local enum, permgroup, i, perm, j, elts;
 
-  if not(IsFinite(h)) then
+  if not IsFinite(h) then
     # What is an example where this can happen?
     TryNextMethod();
   fi;
 
-  if not( IsGroupHClass(h) ) then
+  if not IsGroupHClass(h) then
     # What is an example where this can happen?
     ErrorNoReturn("can only create isomorphisms of group H-classes");
   fi;
@@ -145,7 +145,8 @@ function( h )
     i := i+1;
   od;
 
-  return MappingByFunction( h, permgroup, a -> elts[Position( enum, a )]);
+  return MappingByFunction( h, permgroup, a -> elts[Position( enum, a )],
+                a -> enum[Position( elts, a )]);
 end);
 
 # TODO can this be removed? It doesn't seem to work
--- a/tst/testbugfix/2023-01-24-IsomorphismPermGroup.tst
+++ b/tst/testbugfix/2023-01-24-IsomorphismPermGroup.tst
@@ -1,0 +1,5 @@
+# see <https://github.com/gap-system/gap/issues/3601>
+gap> G:=SL(2,3);;Order(G);
+24
+gap> P:=Image(IsomorphismPermGroup(G));;Order(P);
+24
--- a/tst/testinstall/semitran.tst
+++ b/tst/testinstall/semitran.tst
@@ -1,5 +1,5 @@
 #@local BruteForceAntiIsoCheck,BruteForceInverseCheck,BruteForceIsoCheck
-#@local G,H,I,S,T,enum,inv,map
+#@local G,H,I,S,T,enum,inv,map,x,y
 gap> START_TEST("semitran.tst");
 
 # Test that the inverse of an isomorphism from a partial perm monoid to a
@@ -145,10 +145,11 @@ gap> S := Semigroup(IdentityTransformation);
 # Test IsomorphismPermGroup for an H-class
 gap> S := FullTransformationMonoid(4);;
 gap> H := GreensHClassOfElement(S, One(S));;
-gap> IsomorphismPermGroup(H);;
-gap> H := GreensHClassOfElement(S, Transformation([1, 1, 2, 3]));;
+gap> map := IsomorphismPermGroup(H);;
+gap> for x in H do y:=Image(map, x); Assert(0, PreImagesRepresentative(map, y) = x); od;
 
 # The Semigroups package produces different output so this test is suppressed
+#gap> H := GreensHClassOfElement(S, Transformation([1, 1, 2, 3]));;
 #gap> IsomorphismPermGroup(H);
 #Error, can only create isomorphisms of group H-classes
 


ThomasBreuer (approved):
Looking at the code, my interpretation of the proposed change is that it fixes first of all an inconsistency. Up to now, the result of `IsomorphismPermGroup` depends on whether or not the `NiceMonomorphism` value of the group is already available.
```
gap> g:= SL(2,5);
SL(2,5)
gap> iso:= IsomorphismPermGroup( g );;  Source( iso );
GL(2,5)
gap> g:= SL(2,5);  NiceMonomorphism(g);;
SL(2,5)
gap> iso:= IsomorphismPermGroup( g );;  Source( iso );
SL(2,5)
```
Thus I approve the change.

If this is the only place in the code where `IsomorphismPermGroup` has returned a map whose image is not necessarily isomorphic with the given group then I support the idea that `IsomorphismPermGroup` should always return an isomorphism defined on the given group, as is stated in the Reference Manual.
(Perhaps @hulpke wants to object.)

fingolfin:
(Just to be clear, this PR should not be merged before @hulpke had a chance to comment)

hulpke (commented):
This is patching in one place to satisfy up a complaining user, but is not fundamentally solving the problem:

- There are other places (nice monomorphism, automorphism groups) where a (nice) isomorphism to perm is through an action and is inherited by subgroups. One should go through all of them.
- It is unclear to me whether `GeneralRestrictedMapping` in all cases works effective (and will not accidentally redo an action, change from action to ByImages or have to recompute a range. We should check that this indeed works as expected and without notable overhead or base cost.

fingolfin:
> This is patching in one place to satisfy up a complaining user, but is not fundamentally solving the problem:

I am not sure I follow. The "complaining user*s*" (we had at least two reports, and I'd happily add myself to that list, and basically all the people I spoke to about this seem to agree, too) complained about one specific place, and this fixes that place. What is the fundamental problem beyond "ensure IsomorphismPermGroup methods conform to the documentation?

> 
> * There are other places (nice monomorphism, automorphism groups) where a (nice) isomorphism to perm is through an action and is inherited by subgroups. One should go through all of them.

I don't understand: in how far do these relate to `IsomorphismPermGroup`? NiceMonomorphism and other technical operations of course are not bound by the same rules as `IsomorphismPermGroup`.

Perhaps a concrete example of what you have in mind would help to illuminate this point?

> * It is unclear to me whether `GeneralRestrictedMapping` in all cases works effective (and will not accidentally redo an action, change from action to ByImages or have to recompute a range. We should check that this indeed works as expected and without notable overhead or base cost.

`GeneralRestrictedMapping` is a single global function and pretty clear, in contrast to `RestrictedMapping` which I find odd (e.g. it dispatches to `GeneralRestrictedMapping` for  group homomorphisms which are *known* to be injective but which are not GHBIs, but for a non-GHBI hom which is not (known to be) injective, it forces it to be rebuilt as a GHBI -- that strikes me as problematic. Luckily it does not affect this PR, though :-)

hulpke:
> What is the fundamental problem beyond "ensure IsomorphismPermGroup methods

The problem is that it only fixes one such situation. There are other cases of permutation mappings and nice monomorphisms (automorphism groups being one of them) where exactly the same might be happening. If you want to fix it you need to go through all the `IsomorphismPermGroup` methods and assignments in the library and check.

> GeneralRestrictedMapping is a single global function 

So this *is* checked. Good.

> in contrast to RestrictedMapping which I find odd 

I'm not claiming that the current setup is perfect, it will merit reviewing. But ability to do things, and performance can be a serious difficulty:

- A permutation group acting on all but points or subsets likely will perform faster with GHBI
- Other groups acting might need the action homomorphism, since we cannot decompose
- When restricting to a much smaller group it might merit to rebuild as new GHBI, since its stabilizer chain will not be as deep/wide, so images wil be faster.

ThomasBreuer:
@hulpke @fingolfin
Concerning the statement

>  If you want to fix it you need to go through all the IsomorphismPermGroup methods and assignments in the library and check.

I have checked the currently 18 `IsomorphismPermGroup` methods in the master branch of the GAP library, including those that belong to the `grp` subdirectory, that is, the functions obtained by `MethodsOperation( IsomorphismPermGroup, 1 )`.

The following methods may return a map `map` that is not an isomorphism on the given group.

- file `lib/grpfp.gi`,
  the method for `IsGroup and IsSubgroupFpGroup and IsGroupOfFamily`,
  which calls `IsomorphismPermGroupOrFailFpGroup`:
  This method has the correct `Source` value but might not be surjective; thus calling `Image( map )` will yield an isomorphic permutation group as requested.
  If we keep this method then we will have to state in the documentation of `IsomorphismPermGroup` that the returned map may be not surjective.

- file `lib/grpfp.gi`,
  the method for `IsGroup and IsSubgroupFpGroup`:
  This method may return a map that is bijective but which is defined on the full f.p. group (this happens in the `else` branch of the code).
  Thus this method will need a fix.

- file `lib/grpmat.gi`,
  the two methods for `IsMatrixGroup` and `IsMatrixGroup and IsFinite and IsHandledByNiceMonomorphism`, respectively, for which the current pull request proposes a fix

- file `lib/grpnice.gi`,
  the method for `IsGroup and IsFinite and IsHandledByNiceMonomorphism`:
  According to the code, this method can return `fail`.
  I do not see why `fail` is a reasonable result for a group that is known to be finite.
  (By the way, the documentation of `IsomorphismPermGroup` does not say what happens if the given group is infinite.)

- file `lib/grpramat.gi`,
  the method for `IsCyclotomicMatrixGroup`:
  This method calls either `RestrictedMapping` for a `NiceMonomorphism` map
  (thus the source of `map` is correct, but it may happen that `map` is not surjective)
  or `NicomorphismOfGeneralMatrixGroup`, whose result may have a larger source, as far as I understand.

Besides that, I noticed that
- the method in `lib/grppc.gi` calls `DoMinimalFaithfulPermutationDegree`, which should set `IsBijective` for the map which gets returned, and
- the method in `lib/semitran.gi` creates a mapping by function but without a function for the inverse.

How shall we proceed with the above list?
Shall we put it into a new issue and then solve the problem with a series of pull requests?

hulpke:
Thank you, @ThomasBreuer 

Yes, I think they all need to be sorted out (not something I want to do) before we can claim this to have been changed.

fingolfin:
I don't follow why we can't fix one instance of a bug (which has been reported several times in the past years) just because there are other similar but different bugs? I agree we should fix all instances, but should this be all or nothing?

Anyway, I've had a first look at other methods: Some remarks:
- the one in `lib/grpnice.gi` seems fine to me: even if were to returns `fail`, no serious harm is done (i.e., no chance to get "bad" data via `Image(IsomorphismPermGroup(G)))` in this case, just a plain error, which is fine).
- I improved the one in `lib/semitran.gi`
- There are also a few calls to `SetIsomorphismPermGroup` which need to be investigated
- And then I guess we could also look `IsomorphismPcGroup` and `IsomorphismFpGroup` methods...

hulpke:
> I don't follow why we can't fix one instance of a bug (which has been reported several times in the past years) just because there are other similar but different bugs? I agree we should fix all instances, but should this be all or nothing?

Of course we can change one instance. We just cannot claim that `IsomorphismPermGroup` is always total, until this is done.

PR #5340, opened 2023-01-25
next step for `matobjplist`

ThomasBreuer:
- added consistency checks to `MakeIsPlistVectorRep` (which can be switched off via the third argument)
- added `MakeIsPlistMatrixRep` analogous to `MakeIsPlistVectorRep`
- changed the installed methods such that they call `MakeIsPlistMatrixRep` instead of `Objectify`, and such that hey perform the necessary consistency checks if needed (this fixes a problem from issue #4884)
- state explicitly which consistency checks are needed when creating objects in the filters `IsPlistVectorRep`, `IsPlistMatrixRep` (not in the visible documentation but near to the declarations of the filters)
- moved some code from `matobjplist.gi` to `matobj.gi`
- changed the `InstallMethod` syntax: use strings describing the filters in the requirements instead of an additional info string where possible
- document the mutability status of the resullts of `ChangedBaseDomain` and `CompatibleVector`

The situation with consistency checks (when creating vector and matrix objects) looks as follows:

- The constructors (`Vector`, `NewVector`, `Matrix` ...) can take user input, therefore checks are necessary in their methods; do the entries fit to the base domain, are the base domains of the matrix and its rows equal (currently: identical), are the lengths of the matrix rows correct.
- Operations that take some vector/matrix objects and create a new one have to check whether the arguments are compatible, but then often it is clear that the result will be consistent; for example, the sum of two vector objects with the same base domain is safe.
- There are cases where this is not clear, `Inverse` for a matrix object must check that the result fits to the base domain.
- For in-place operations such as `AddRowVector` and `MultVectorLeft`, we have the problem that the consistency can be checked only after the operation, and if not then we have already created a corrupted object. (Note that the aim of the in-place operation is to avoid making a copy.) We could avoid this problem by requesting that scalar multiplications are allowed only with elements of the base domain (or also with integers?).

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -1177,7 +1177,7 @@ InstallMethod( CopySubMatrix,
 ##
 InstallMethod( TransposedMatImmutable,
     [ IsMatrixOrMatrixObj ],
-    TransposedMatMutable );
+    M -> MakeImmutable( TransposedMatMutable( M ) ) );
 
 
 #############################################################################
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -287,6 +287,15 @@ InstallMethod( ZeroVector,
     { len, M } -> NewZeroVector( CompatibleVectorFilter( M ),
                                  BaseDomain( M ), len ) );
 
+# compatibility method for representations as list of elements
+# (covers the cases of the second argument in `IsRowVector` or `IsMatrix`)
+InstallOtherMethod( ZeroVector, "for an integer and a plain list",
+  [ IsInt, IsPlistRep ],
+  -1, # rank lower than default as only fallback
+  function( l, t )
+    return ListWithIdenticalEntries( l, ZeroOfBaseDomain( t ) );
+  end);
+
 
 #############################################################################
 ##
@@ -1160,6 +1169,15 @@ InstallMethod( CopySubMatrix,
       od;
     od;
     end );
+
+
+#############################################################################
+##
+#M  TransposedMatImmutable( <M> )
+##
+InstallMethod( TransposedMatImmutable,
+    [ IsMatrixOrMatrixObj ],
+    TransposedMatMutable );
 
 
 #############################################################################
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -1127,20 +1127,19 @@ InstallMethod( IsZero,
 InstallMethod( IsOne,
   [ "IsPlistMatrixRep" ],
   function( m )
-    local i,j,n;
+    local n, i, row;
     if Length(m![ROWSPOS]) <> m![RLPOS] then
         #Error("IsOne: Matrix must be square");
         return false;
     fi;
     n := m![RLPOS];
     for i in [1..n] do
-        if not IsOne(m![ROWSPOS][i]![ELSPOS][i]) then return false; fi;
-        for j in [1..i-1] do
-            if not IsZero(m![ROWSPOS][i]![ELSPOS][j]) then return false; fi;
-        od;
-        for j in [i+1..n] do
-            if not IsZero(m![ROWSPOS][i]![ELSPOS][j]) then return false; fi;
-        od;
+      row:= m![ROWSPOS][i];
+      if PositionNonZero( row ) <> i or
+         not IsOne( row![ELSPOS][i] ) or
+         PositionNonZero( row, i ) <= n then
+        return false;
+      fi;
     od;
     return true;
   end );
--- a/lib/matobjplist.gi
+++ b/lib/matobjplist.gi
@@ -18,6 +18,20 @@
 # Constructors:
 ############################################################################
 
+
+############################################################################
+##
+#F  MakeIsPlistVectorRep( <basedomain>, <list>, <check> )
+##
+##  Construct a new vector in the filter 'IsPlistVectorRep' with base domain
+##  <basedomain> and entries in the list <list> (without copying).
+##
+##  If <check> is set to 'true' *and* 'ValueOption( "check" )' is 'true',
+##  then it is checked that the entries of <list> are in <basedomain>.
+##  So whenever you know that the input is guaranteed to be in <basedomain>,
+##  pass 'false' for <check> to omit these (potentially costly) consistency
+##  checks.
+##
 BindGlobal( "MakeIsPlistVectorRep",
   function( basedomain, list, check )
     local fam, types, typ;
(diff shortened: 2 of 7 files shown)


lib/matobjplist.gi
```diff
@@ -19,7 +19,7 @@
 ############################################################################
 
 BindGlobal( "MakeIsPlistVectorRep",
```
fingolfin, line 21: Maybe add a comment which explains `check`?

Something along these lines (feel free to rewrite this, or to ignore this suggestion)
```suggestion
# Construct a new vector in the filter IsPlistVectorRep with the given
# basedomain and entries copied form the given list.
# If `check` is set to true *and* ValueOption( "check" ) is true, then
# it is checked that the list entries are in the basedomain.
# So whenever it is known that the input is guaranteed to be in the basedomain,
# pass false for check to avoid these (potentially costly) consistency checksl
BindGlobal( "MakeIsPlistVectorRep",
```

lib/matobjplist.gi
```diff
@@ -71,33 +71,81 @@ BindGlobal( "MakeIsPlistVectorRep",
...
+    if IsMutable( list ) then
+      typ:= types[2];
+    else
+      typ:= types[1];
+    fi;
+
+    if check and ValueOption( "check" ) <> false then
+      for row in list do
+        if not IsPlistVectorRep( row ) then
+          Error( "the entries of <list> must be in 'IsPlistVectorRep'" );
```
fingolfin, line 117: Should we have a test case for this? And likewise for the next three errors cases, which according to Codecov are currently not getting exercised?

lib/matobjplist.gi
```diff
@@ -71,33 +71,81 @@ BindGlobal( "MakeIsPlistVectorRep",
...
     list := ListWithIdenticalEntries(len, Zero(basedomain));
-    return MakeIsPlistVectorRep(basedomain, list);
+    return MakeIsPlistVectorRep(basedomain, list, false);
   end );
 
 InstallMethod( NewMatrix,
-  "for IsPlistMatrixRep, a ring, an int, and a list",
-  [ IsPlistMatrixRep, IsRing, IsInt, IsList ],
+  [ "IsPlistMatrixRep", "IsRing", "IsInt", "IsList" ],
   function( filter, basedomain, rl, l )
```
fingolfin, line 147: this is out of scope for this PR, but I would suggest we consistently replace `rl` by `ncols` or `numcols` or something like that (because `row length` is mostly gone now and so it is hard to guess what `rl` stands for). Also, perhaps `l` should be turned into `list`

lib/matobjplist.gi
```diff
@@ -579,50 +657,55 @@ InstallMethod( DimensionsMat, "for a plist matrix",
...
-  [ IsInt, IsPlistMatrixRep ],
+InstallMethod( IdentityMatrix,
+  [ "IsInt", "IsPlistMatrixRep" ],
   function( rows,m )
-    local i,l,o,t,res;
+    local t,l,o,i;
     t := m![EMPOS];
     l := List([1..rows],i->ZeroVector(rows,t));
     o := One(m![BDPOS]);
+#T can 'o = fail' happen?
```
fingolfin, line 676: Perhaps explain the motivation for this question? I assume it is because we discussed the potential use of matrices over semirings where there might be no one element (CC @james-d-mitchell ) ? 

I mean, we can't exclude the chance of this returning `fail` or triggering an error as long as we permit arbitrary BasedDomain... but then the same could be asked about zero, too? Like, the basedomain could be `PositiveIntegers`, too?

So I'd not worry too much about it, but if we feel this warrants a TODO comments, let's at least add some context to it?

lib/matobjplist.gi
```diff
@@ -1173,7 +1251,7 @@ InstallMethod( RankMat, "for a plist matrix",
 
 
 InstallMethodWithRandomSource( Randomize,
-  "for a random source and a mutable plist matrix",
+  "for a random source and a mutable plist vector",
```
fingolfin, line 1253: Why this change? The second argument *is* a plist matrix, not a vector, isn't it?

lib/vspcrow.gi
```diff
@@ -161,16 +161,24 @@ InstallOtherMethod( LeftModuleByGenerators,
...
       TryNextMethod();
     fi;
     typ:=IsAttributeStoringRep and HasIsEmpty and IsFiniteDimensional;
     if ForAll( mat, row -> IsSubset( F, BaseDomain(row) ) ) then
+      # Replace 'mat' by vector objects with base domain 'F'.
+      mat:= List( mat, v -> ChangedBaseDomain( v, F ) );
       typ:=typ and IsGaussianRowSpace;
     else
+      # Replace 'mat' by vector objects with base domain containing 'F'.
+      K:= F;
```
fingolfin, line 177: Can we add a test case to exercise this code path? Codecov says it currently isn't covered

lib/matobj.gi
```diff
@@ -1162,6 +1171,15 @@ InstallMethod( CopySubMatrix,
     end );
 
 
+#############################################################################
+##
+#M  TransposedMatImmutable( <M> )
+##
+InstallMethod( TransposedMatImmutable,
+    [ IsMatrixOrMatrixObj ],
+    TransposedMatMutable );
```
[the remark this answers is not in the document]
ThomasBreuer: Yes.
The method call yields a mutable result, and then it is a feature of attributes that the `Immutable` value of the result is returned.
(The documentation of `NewAttribute` is not really precise in this respect.)

lib/vspcrow.gi
ThomasBreuer: No.
The point is the `IsNonGaussianRowsSpace` filter which gets set without the necessary code for the `IsVectorObj` case.
The `else` branch did not work up to now because of this problem, and still does not work,
the aim of the proposed change is to provide at least suitable vector objects.
I will mention this in the `FIXME` comment below.

lib/matobjplist.gi
ThomasBreuer: O.k.

lib/matobjplist.gi
```diff
@@ -71,33 +71,81 @@ BindGlobal( "MakeIsPlistVectorRep",
...
     return Objectify(typ, [ basedomain, list ]);
   end );
 
-InstallMethod( NewVector, "for IsPlistVectorRep, a ring, and a list",
-  [ IsPlistVectorRep, IsRing, IsList ],
-  function( filter, basedomain, l )
-    if ValueOption( "check" ) <> false and not IsSubset( basedomain, l ) then
-      Error( "the elements in <l> must lie in <basedomain>" );
+BindGlobal( "MakeIsPlistMatrixRep",
+  function( basedomain, emptyvector, rowlength, list, check )
```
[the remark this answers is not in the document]
ThomasBreuer: O.k.

lib/matobjplist.gi
ThomasBreuer: O.k., I remove this comment.
First I wanted to move it to a place that defines what an admissible base domain is, but apparently this is currently left open.

lib/matobjplist.gi
ThomasBreuer: Thanks for spotting this. This change had happened by chance.

fingolfin (approved):
Thank you, looks great!

PR #5342, opened 2023-01-26
Add faster IsOne method for matrices

### LINKED ISSUE #5341: Performance regression for matrices.
What I believe is the current development branch. Big compact matrices are highly inefficient in some operations:
```
map:=RegularActionHomomorphism(MathieuGroup(12));;
p:=Image(map);
mats:=List(GeneratorsOfGroup(p),x->PermutationMat(x,95040,GF(2))); # list of vectors. Really?
IsOne(mats[1]);
mats:=List(mats,x->ImmutableMatrix(GF(2),x)); #proper compact matrices
IsOne(mats[1]);
```

Both `IsOne` take lots of time and memory, since they call `Unpack`. Clearly not needed.

Same for `Order`, or if I change the field.

fingolfin:
Also add similar IsOne and IsZero methods for rowlist matobj

Resolves #5341 by @hulpke 

Example (before these took veery long... I didn't time it, but let's say closer to minutes than to milliseconds):
```
gap> map:=RegularActionHomomorphism(MathieuGroup(12));;
gap> p:=Image(map);
<permutation group of size 95040 with 3 generators>
gap>
gap> m:=PermutationMat(p.1,95040,GF(2));;
gap> IsOne(m); time;
false
13
gap> ConvertToMatrixRep(m);
2
gap> IsOne(m); time;
false
0
gap>
gap> m2:=PermutationMat(One(p),95040,GF(2));;
gap> IsOne(m2); time;
true
81
gap> ConvertToMatrixRep(m2);
2
gap> IsOne(m2); time;
true
70
```

### DIFF
--- a/lib/matobj.gi
+++ b/lib/matobj.gi
@@ -1381,6 +1381,38 @@ InstallMethod( InverseSameMutability,
     end );
 
 InstallMethod( IsZero,
+    [ IsRowListMatrix and IsMatrixObj ],
+    function( mat )
+    local ncols, row;
+
+    ncols:= NrCols( mat );
+    for row in mat do
+      if PositionNonZero( row ) <= ncols then
+        return false;
+      fi;
+    od;
+    return true;
+    end );
+
+InstallMethod( IsOne,
+    [ IsRowListMatrix and IsMatrixObj ],
+    function( mat )
+    local ncols, i, row;
+
+    ncols:= NrCols( mat );
+    for i in [1 .. NrRows( mat )] do
+      row := mat[i];
+      if PositionNonZero( row ) <> i or not IsOne( row[i] ) then
+        return false;
+      fi;
+      if PositionNonZero( row, i ) <= ncols then
+        return false;
+      fi;
+    od;
+    return true;
+    end );
+
+InstallMethod( IsZero,
     [ IsMatrixObj ],
     M -> IsZero( Unpack( M ) ) );
 
--- a/lib/matrix.gi
+++ b/lib/matrix.gi
@@ -1174,6 +1174,32 @@ InstallMethod( IsZero,
     return true;
     end );
 
+
+#############################################################################
+##
+#M  IsOne( <mat> )
+##
+InstallMethod( IsOne,
+    "method for a matrix",
+    [ IsMatrix ],
+    function( mat )
+    local ncols,  # number of columns
+          i,
+          row;    # loop over rows in 'obj'
+
+    ncols:= NrCols( mat );
+    for i in [1 .. NrRows( mat )] do
+      row := mat[i];
+      if PositionNonZero( row ) <> i or not IsOne( row[i] ) then
+        return false;
+      fi;
+      if PositionNonZero( row, i ) <= ncols then
+        return false;
+      fi;
+    od;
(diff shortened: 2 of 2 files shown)


lib/matobj.gi
```diff
@@ -1380,6 +1380,38 @@ InstallMethod( InverseSameMutability,
...
     end );
 
+InstallMethod( IsZero,
+    [ IsRowListMatrix and IsMatrixObj ],
+    function( mat )
+    local ncols, row;
+
+    ncols:= NrCols( mat );
+    for row in mat do
+      if PositionNonZero( row ) <= ncols then
```
james-d-mitchell, line 1390: I was going to suggest that this might be clearer if it was:
```suggestion
      if PositionNonZero( row ) <> fail then
```
only to discover that `PositionNonZero` returns `Length(row) + 1` if no non-zero is located. Is this intentional, it seems to be at odds with `Position`.

lib/matobj.gi
```diff
@@ -1380,6 +1380,38 @@ InstallMethod( InverseSameMutability,
...
+    function( mat )
+    local ncols, i, row;
+
+    ncols:= NrCols( mat );
+    for i in [1 .. NrRows( mat )] do
+      row := mat[i];
+      if PositionNonZero( row ) <> i or not IsOne( row[i] ) then
+        return false;
+      fi;
+      if PositionNonZero( row, i ) <= ncols then
```
james-d-mitchell, line 1408: Think you might have copy pasted lines 1408 to 1410 from `IsZero` here, this can't be right.

lib/matrix.gi
```diff
@@ -1174,6 +1174,32 @@ InstallMethod( IsZero,
...
+
+    ncols:= NrCols( mat );
+    for i in [1 .. NrRows( mat )] do
+      row := mat[i];
+      if PositionNonZero( row ) <> i or not IsOne( row[i] ) then
+        return false;
+      fi;
+      if PositionNonZero( row, i ) <= ncols then
+        return false;
+      fi;
```
james-d-mitchell, line 1198:
```suggestion
```

james-d-mitchell: Same copy paste error as above.

james-d-mitchell (changes requested):
Looks good apart from the copy paste errors indicated.

lib/matobj.gi
james-d-mitchell: Sorry my mistake.

lib/matrix.gi
james-d-mitchell: Sorry I'm mistaken, this is correct, thanks to @ThomasBreuer for pointing this out.

james-d-mitchell (approved):

lib/matobj.gi
ThomasBreuer: The documentation says that `PositionNonZero` is a special case of `PositionNot`, and `PositionNot` treats unbound entries as "not the element which shall be avoided".

ThomasBreuer (approved):
Good.

(I will change the available `IsOne` method in `matobjplist.gi` accordingly, in an update of my pull request #5340.)

PR #5343, opened 2023-01-26
mat8bit: fix bug in PostMakeImmutable

james-d-mitchell:
This PR fixes an off-by-one error in `PostMakeImmutable` for `Is8BitMatrixRep` objects. As I understand from discussions with @ThomasBreuer, all of the rows of an immutable matrix must be immutable, but previously for `Is8BitMatrixRep` the last row did not have `MakeImmutable` called on it due to an off-by-one error. 

## Text for release notes

I'm not sure if this should go in the release notes or not.


ThomasBreuer (approved):
Good catch!

PR #5345, opened 2023-01-26
Make `ZeroSameMutability` for 8-bit matrices consistent with `AdditiveInverseSameMutability`

james-d-mitchell:
The previous behaviour of the methods installed for `ZeroSameMutability` was inconsistent with the behaviour of `AdditiveInverseSameMutability`, this commit makes them have the same behaviour, i.e. that if `ZeroSameMutability` is called for a mutable matrix `m`, then the `m` can have mutable or immutable rows (in any combination), but the mutability of the rows in the output was determined by the mutability of the first row only. This commit gives the result of `ZeroSameMutability` the same pattern of mutable/immutable rows as the input.

I believe that this was the desired behaviour after a discussion with @ThomasBreuer and @fingolfin  some weeks ago.

This PR builds on #5343, can separate this dependency if necessary.

## Text for release notes

None?

### DIFF
--- a/lib/mat8bit.gi
+++ b/lib/mat8bit.gi
@@ -1068,7 +1068,7 @@ InstallMethod(NestingDepthA, [Is8BitVectorRep], m->1);
 InstallMethod(PostMakeImmutable, [Is8BitMatrixRep],
         function(m)
     local i;
-    for i in [2..m![1]] do
+    for i in [2..m![1]+1] do
         MakeImmutable(m![i]);
     od;
 end);
--- a/lib/mat8bit.gi
+++ b/lib/mat8bit.gi
@@ -646,33 +646,25 @@ InstallMethod( ZeroImmutable, "8 bit matrix", true,
     return z;
 end);
 
-InstallMethod( ZeroSameMutability, "8 bit matrix", true,
-        [Is8BitMatrixRep and IsMatrix and IsAdditiveElementWithZero
-         and IsSmallList ],
-        0,
-        function(mat)
-    local z, i,zv;
-    z := [mat![1]];
-    if not IsMutable(mat![2]) then
-        zv := ZERO_VEC8BIT(mat![2]);
-        SetFilterObj(zv, IsLockedRepresentationVector);
-        MakeImmutable(zv);
-        for i in [2..mat![1]+1] do
-            z[i] := zv;
-        od;
+InstallMethod(ZeroSameMutability, "8 bit matrix",
+[Is8BitMatrixRep and IsMatrix and IsAdditiveElementWithZero
+and IsSmallList],
+function(mat)
+  local z, i, zv;
+  z := [mat![1]];
+  zv := ZERO_VEC8BIT(mat![2]);
+  SetFilterObj(zv, IsLockedRepresentationVector);
+  MakeImmutable(zv);
+  for i in [2 .. mat![1] + 1] do
+    if IsMutable(mat![i]) then
+      z[i] := ZERO_VEC8BIT(mat![i]);
+      SetFilterObj(z[i], IsLockedRepresentationVector);
     else
-        for i in [2..mat![1]+1] do
-            zv := ZERO_VEC8BIT(mat![i]);
-            SetFilterObj(zv,IsLockedRepresentationVector);
-            z[i] := zv;
-        od;
+      z[i] := zv;
     fi;
-    if IsMutable(mat) then
-       Objectify(TYPE_MAT8BIT(Q_VEC8BIT(mat![2]),true), z);
-    else
-        Objectify(TYPE_MAT8BIT(Q_VEC8BIT(mat![2]),false), z);
-    fi;
-    return z;
+  od;
+  Objectify(TYPE_MAT8BIT(Q_VEC8BIT(mat![2]), IsMutable(mat)), z);
+  return z;
 end);
 
 #############################################################################
--- a/tst/testinstall/mat8bit.tst
+++ b/tst/testinstall/mat8bit.tst
@@ -205,5 +205,22 @@ false
 gap> Zero( mz ) = m * Zero( z );
 true
 
+# PostMakeImmutable bug
+gap> m := [[1, 0, 1, 1], [0, 1, 1, 1]] * Z(4);
+[ [ Z(2^2), 0*Z(2), Z(2^2), Z(2^2) ], [ 0*Z(2), Z(2^2), Z(2^2), Z(2^2) ] ]
+gap> ConvertToMatrixRepNC(m);;
+gap> m;
+[ [ Z(2^2), 0*Z(2), Z(2^2), Z(2^2) ], [ 0*Z(2), Z(2^2), Z(2^2), Z(2^2) ] ]
+gap> IsMutable(m);
+true
+gap> MakeImmutable(m);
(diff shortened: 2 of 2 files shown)


james-d-mitchell:
BTW: `OneSameMutability` for `Is8BitMatrixRep` suffers the same problem, if this PR is acceptable, then I'll update that too (either in this PR or another one, whatever's preferable).

ThomasBreuer (approved):
Thanks.

The question is whether the now consistent behaviour shall be documented as a feature of 8-bit-matrices (and for `IsRowListMatrix` objects in general).

ThomasBreuer:
@james-d-mitchell

> BTW: OneSameMutability for Is8BitMatrixRep suffers the same problem, if this PR is acceptable, then I'll update that too (either in this PR or another one, whatever's preferable).

I think it makes sense to add the analogous change for `OneSameMutability` to this pull request, thus I do not merge it now.

james-d-mitchell:
Will do, thanks @ThomasBreuer !

james-d-mitchell:
I start to wonder if this is really the correct solution. Are there any use cases where having the same pattern of mutable/immutable rows in a mutable zero matrix as in the original matrix is actually useful? Isn't this just creating more complexity? Wouldn't it make more sense for `ZeroSameMutability` to return a mutable matrix with mutable rows if the input is a mutable matrix (with any pattern of mutable/immutable rows), and an immutable matrix if the input is immutable? 

Before this pull request `AdditiveInverseSameMutability` seems to be the odd one out, `ZeroSameMutability`, `OneSameMutability`, and `InverseSameMutability` all have the following behaviour:

* if the input is mutable and the **first** row is immutable, then the output is mutable **every** row is immutable;
* if the input is mutable and the **first** row is mutable, then the output is mutable **every** row is mutable;
* if the input it immutable, then the output is immutable and every row is immutable.

Also as an aside there is the kernel function `FuncINV_MAT8BIT_IMMUTABLE` which isn't used in the library at all, and so currently calling `InverseImmutable` on a matrix in `Is8BitMatrixRep` just uses the default method for "any object"!!

ThomasBreuer:
As far as I see, there is no documentation yet which promises a particular mutability of the rows of a mutable object returned by `SomethingSameMutability`. Thus we are in principle free to define *some* behaviour, and to adjust the code accordingly. (Of course there is a danger that such changes break existing code.)

One possibility would be to state that if the input is fully mutable then also the output is fully mutable, and that nothing is guaranteed in the case of mixed mutability or in the case that all rows of the input are immutable.

I think the idea that mutable rows in the input should correspond to mutable rows in the result comes from the fact that the natural algorithm to compute an additive inverse, a zero matrix, or an identity matrix is to proceed row by row (which does of course not work for inverses).
In fact, I had checked that the code for `IsPlistMatrixRep` yields results whose mutability fits to the changes proposed in the current pull request.

fingolfin:
The documentation for `AdditiveInverseSameMutability` says this:

> [...] it returns a result of the same immutability level as the argument. For  instance, if the argument is a mutable matrix with immutable rows, it returns a similar object.

This lead me to wonder what "immutability level" means, so I grepped for it, and discovered that [the GAP manual defines it](https://docs.gap-system.org/doc/ref/chap21_mj.html#X8676EFE67972FD06). Among other things, it says:

> Moreover, the results have a "homogeneous" mutability status, that is, if the first bound entry at nesting depth $i$ is immutable (mutable) then all entries at nesting depth $i$ are immutable (mutable, provided that a mutable version of this entry exists in GAP).
>
> Thus the sum of two mutable matrices whose first rows are mutable is a matrix all of whose rows are mutable, and the product of two matrices whose first rows are immutable is a matrix all of whose rows are immutable, independent of the mutability status of the other rows of the arguments.

There is more there, and overall it seems sensible enough to me, at least for row list matrices.

ThomasBreuer:
@fingolfin Thanks for digging this out, I was not aware of the manual section in question.

This means that the situation for 8bit matrices is defined, and the inconsistency should be fixed according to the manual.

Row list matrices are not affected by the definition because their rows are not in `IsGeneralizedRowVector`,
but we can adopt the definition for row list matrices (and adjust the implementation for `IsPlistMatrixRep` accordingly.

(The quoted definition from the manual makes sense because is does not force one to distribute a "mixed mutability" status of objects via arithmetic operations to new objects. On the other hand, it means that "the obvious code" for computing the sum etc. of matrices and row list matrix objects is not correct w.r.t. the mutability of the rows.)

james-d-mitchell:
Thanks @ThomasBreuer and @fingolfin  for your comments. Do I understand correctly that the consensus is that the correct behaviour for matrices in `Is8BitMatrixRep` is the following:

* if the input is mutable and the **first** row is immutable, then the output is mutable **every** row is immutable;
* if the input is mutable and the **first** row is mutable, then the output is mutable **every** row is mutable;
* if the input it immutable, then the output is immutable and every row is immutable.

with respect to (at least) the operations `AdditiveInverseSameMutability`, `ZeroSameMutability`, `InverseSameMutability`, 
`OneSameMutability`?  At present I believe that `ZeroSameMutability`, `InverseSameMutability`, 
`OneSameMutability` have the behaviour described above, but that `AdditiveInverseSameMutability` does not. I'm happy to update the PR to make `AdditiveInverseSameMutability` have the same behaviour as the others (although I don't exactly follow why the above behaviour is desirable).

james-d-mitchell:
Thanks @ThomasBreuer, I've made the changes discussed here in a new PR #5356, and I'm going to close this PR.

ThomasBreuer:
@james-d-mitchell @fingolfin 
I am still trying to understand the meaning of the section about Mutability Status and List Arithmetic in the GAP Reference Manual which is quoted above.

- As far as I understand, GAP does not behave consistently w.r.t. the rules stated there, already for `IsPlistRep` matrices.
```
gap> m1:= [ [ 1, 2 ], [ 3, 4 ] ];;  m2:= m1 + 1;;
gap> MakeImmutable( m1[1] );;  MakeImmutable( m2[1] );;
gap> List( m1 + m2, IsMutable );
[ false, true ]
gap> List( m1 - m2, IsMutable );
[ false, true ]
gap> List( m1 * m2, IsMutable );
[ false, true ]
gap> List( AdditiveInverseSameMutability( m1 ), IsMutable );
[ false, true ]
gap> List( ZeroSameMutability( m1 ), IsMutable );
[ false, true ]
gap> List( OneSameMutability( m1 ), IsMutable );
[ true, true ]
gap> List( InverseSameMutability( m1 ), IsMutable );
[ false, false ]
gap> List( SUM_LIST_LIST_DEFAULT( m1, m2 ), IsMutable );
[ false, true ]
gap> List( DIFF_LIST_LIST_DEFAULT( m1, m2 ), IsMutable );
[ false, true ]
```
  Do we have to fix most of these results according to the rules from the manual section? Or is this too dangerous because code that relies on the current behaviour would be broken?

- If we take the abovementioned manual section for granted, would it make sense to adopt its rules also for defining the mutability of `IsRowListMatrix` objects (or perhaps only for `IsPlistMatrixRep` objects), or is this asking for complications?
If yes, does this imply that `ChangedBaseDomain` for such matrices behaves according to these rules, and `Matrix( list, len, M )` for a row list matrix `M` will create a row list matrix whose rows are mutable or immutable depending on the mutability of the first entry in `list`? (If yes then all this must be documented.)

PR #5351, opened 2023-01-28
Convert more tabs to spaces

fingolfin:
Also remove a few trailing newlines

### DIFF
--- a/dev/releases/README.md
+++ b/dev/releases/README.md
@@ -1,10 +1,10 @@
 # How to release GAP version X.Y.Z
- 
+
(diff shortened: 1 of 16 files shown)


ChrisJefferson:
I'm happy for this PR to be merged, once the problem with the failing expect test is fixed (which I think will just require making sure the test and it's output are in sync?)

fingolfin:
Fixed. I accidentally removed tabs in a test which used tabs to test error handling in a specific way

ChrisJefferson (approved):

PR #5352, opened 2023-01-28
Remove obsolete (?) MutableCopyMat method

fingolfin:


lib/matrix.gi
```diff
@@ -2031,15 +2031,6 @@ InstallOtherMethod( MutableCopyMatrix, "for empty lists", [ IsList and IsEmpty ]
   mat -> [ ] );
 
 
-# FIXME: remove the following method again sometime soon (as of February
-# 2020); it is only here to work around an issue in the polycyclic package
-# which calls MutableCopyMat on a vector; once a new polycyclic release is
-# out, we should revert this again.
```
fingolfin, line 2037: The current `polycyclic` release 2.16 from 25/07/2020 took care of this.

ThomasBreuer (approved):
Let us see whether other GAP packages "misuse" `MutableCopyMatrix`/`MutableCopyMat` for lists that are not matrices, and run into a "no method found" error after the change.

My interpretation is that `MutableCopyMat` had been introduced and documented in the `cvec` package, for its cmat objects.
The GAP library had introduced (in `lib/matrix.gd`) the synonym `MutableCopyMatrix` for the more general situation of arbitrary lists, but this name was not documented.
And `lib/matobj2.gd` introduces and documents `MutableCopyMatrix` for `IsMatrixOrMatrixObj` objects.

I think that eventually `MutableCopyMat` can be regarded as an obsolete synonym; there are only few occurrences of `MutableCopyMat` in the GAP library, which can be replaced by `MutableCopyMatrix`.

james-d-mitchell (approved):
This looks good, although it might also be a good idea to remove `MutableCopyMat` except as a synonym for `MutableCopyMatrix` if they are indeed synonymous (i.e. don't install methods for or use `MutableCopyMat` in the library but retain the synonym declaration).

ThomasBreuer:
I agree with @james-d-mitchell:
The library should use only `MutableCopyMatrix`, and `MutableCopyMat` should be turned into a synonym of it, defined in `lib/obsolete.gd`, with the remark that the cvec package uses and documents it.
I will propose a pull request for this change.

fingolfin:
It seems this broke the hap and fining test suites, see https://github.com/gap-system/PackageDistro/actions/runs/4125512088/jobs/7126824210 and https://github.com/gap-system/PackageDistro/actions/runs/4125512088/jobs/7126819955 -- I have not yet investigated closer. Of course we can always just revert this PR, but I don't think it is super urgent, so perhaps we can first investigate -- and possibly submit fixes (?) to those packages, too

Issue #5348, opened 2023-01-26
documentation of Omega


BrauerSuzuki:
The documentation states that Omega(G,p,n) is the **set** of elements g in a p-group G such that p^(p^n)=1. However, it is the subgroup generated by this set. The documentation of Agemo is however correct.

### Copy and paste GAP banner (to tell us about your setup)
 ┌───────┐   GAP 4.12.0 of 2022-08-18
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv8

fingolfin:
Fixed by PR #5353

PR #5353, opened 2023-01-28
Fix docs for Omega subgroup

jackschmidt:
Fixes issue #5348 -- previously the documentation used set braces, but now it uses subgroup generated langle and rangle braces.

## Text for release notes

none


fingolfin (approved):
Thanks!

PR #5356, opened 2023-01-30
Add tests for matrices in `Is8BitMatrixRep` and fix some inconsistent behaviour

james-d-mitchell:
This PR adds a bunch of tests to `test/testinstall/mat8bit.tst` so that most of `lib/mat8bit.gi` is tested, and fixes some anomalies along the way. See #5345  for a discussion. 

## Text for release notes

?

### DIFF
--- a/lib/mat8bit.gi
+++ b/lib/mat8bit.gi
@@ -706,7 +706,7 @@ InstallMethod( InverseSameMutability, "8 bit matrix", true,
 
 #############################################################################
 ##
-#M <mat>^0
+#M One
 ##
 
 InstallMethod( OneSameMutability, "8 bit matrix", true,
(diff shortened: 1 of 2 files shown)


lib/mat8bit.gi
```diff
@@ -1011,30 +996,26 @@ InstallMethod(TriangulizeMat,
...
+  for i in [1 .. NrRows(m)] do
+    if not IsMutable(m[i]) then
+      m[i] := ShallowCopy(m[i]);
+    fi;
+  od;
+  TRIANGULIZE_LIST_VEC8BITS(m);
+
+  CONV_MAT8BIT(m,q);
+  if not mut then
+    PostMakeImmutable(m);
```
fingolfin, line 1017: I don't think we are supposed to call `PostMakeImmutable` directly... but then again it does the desired thing here, so.... *shrug*

fingolfin (approved):

james-d-mitchell:
Thanks for your comments @fingolfin i will update accordingly

fingolfin (approved):

PR #5357, opened 2023-01-30
Remove undocumented & unused AgemoAbove

fingolfin:
Last caller was removed 2000-05-23

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -2470,20 +2470,6 @@ DeclareOperation( "EulerianFunction", [ IsGroup, IsPosInt ] );
 
 #############################################################################
 ##
-#F  AgemoAbove( <G>, <C>, <p> )
-##
-##  <ManSection>
-##  <Func Name="AgemoAbove" Arg='G, C, p'/>
-##
-##  <Description>
-##  </Description>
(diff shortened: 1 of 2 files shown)


hulpke (commented):
It is clearly not used. But in general code for `Omega` and `Agemo` was (at least 10 years ago) somewhat haphazard. So this might have been a relic of trying to do it better.

ThomasBreuer (commented):
Perhaps we are almost sure that nobody uses the function `AgemoAbove`, but why don't we just move it to `lib/obsolete.gi`?

fingolfin:
@hulpke it definitely was used before the date I listed, but that was when during a rewrite the last caller to it was removed.

@ThomasBreuer well it is a completely undocumented function (not even a code comment hinting at what it does) with trivial code, and no package or GAP library code uses it (and if anything external were to use it, (a) its their fault, and (b) it'd be trivial for them to repair it). As such I see little benefit in putting this into `obsolete.gi`, where it would just keep sitting indefinitely?

ThomasBreuer (approved):

james-d-mitchell (approved):
Seems safe to remove this based on what @fingolfin has written

PR #5358, opened 2023-02-01
Document that attribute methods better make their results immutable

ThomasBreuer:


fingolfin (approved):

doc/ref/types.xml
```diff
@@ -435,10 +435,18 @@ computed, and claim that it knows these values.  Note that <Q>store</Q>
...
 (see&nbsp;<Ref Sect="Mutability and Copyability"/>),
 except if the attribute had been specially constructed as
-<Q>mutable attribute</Q>.
+<Q>mutable attribute</Q>, see <Ref Func="NewAttribute"/>.
+<P/>
+Note that functions installed as methods for attributes may return mutable
+objects,
+the attribute returns the value of <Ref Func="Immutable"/> for the result.
+Thus it is recommended that these methods make their return values immutable
+by calling <Ref Func="MakeImmutable"/> if this is possible,
```
fingolfin, line 447: Maybe slightly more defensive (there are other ways to get an immutable object):
```suggestion
Thus it is recommended that these methods ensure their return values are immutable
e.g. by calling <Ref Func="MakeImmutable"/> if this is possible,
```

ThomasBreuer: O.k.

doc/ref/types.xml
```diff
@@ -435,10 +435,18 @@ computed, and claim that it knows these values.  Note that <Q>store</Q>
 and <Q>know</Q> have to be understood in the sense that it is very cheap
 to get such a value when the attribute is called again.
 <P/>
-The stored value of an attribute is in general immutable
+The value returned by the call of an attribute is in general immutable
```
james-d-mitchell, line 438:
```suggestion
The value returned by an attribute is in general immutable
```

ThomasBreuer: O.k., thanks.

james-d-mitchell: Aha, ok, then better with just "returned by an attribute" then.

fingolfin (approved):

PR #5359, opened 2023-02-01
Add new property `IsRegularPGroup`

fingolfin:
Unfortunately I am not aware of an efficient algorithm for this in general. But the provided one at least performs reasonably for e.g. the groups of order 3^7 or 5^7.

### DIFF
--- a/lib/grp.gd
+++ b/lib/grp.gd
@@ -439,11 +439,8 @@ DeclareOperation( "KnowsHowToDecompose", [ IsGroup, IsList ] );
 DeclareProperty( "IsPGroup", IsGroup );
 InstallTrueMethod( IsGroup, IsPGroup );
 
-InstallSubsetMaintenance( IsPGroup,
-    IsGroup and IsPGroup, IsGroup );
-
-InstallFactorMaintenance( IsPGroup,
-    IsGroup and IsPGroup, IsObject, IsGroup );
+InstallSubsetMaintenance( IsPGroup, IsPGroup, IsGroup );
+InstallFactorMaintenance( IsPGroup, IsPGroup, IsObject, IsGroup );
 
 InstallTrueMethod( IsPGroup, IsGroup and IsTrivial );
 InstallTrueMethod( IsPGroup, IsGroup and IsElementaryAbelian );
@@ -458,7 +455,7 @@ InstallTrueMethod( IsPGroup, IsGroup and IsElementaryAbelian );
 ##
 ##  <Description>
 ##  <Index Key="PowerfulPGroup">Powerful <M>p</M>-group</Index>
-##  A finite p-group <A>G</A> is said to be a <E>powerful <M>p</M>-group</E>
+##  A finite <M>p</M>-group <A>G</A> is said to be a <E>powerful <M>p</M>-group</E>
 ##  if the commutator subgroup <M>[<A>G</A>,<A>G</A>]</M> is contained in
 ##  <M><A>G</A>^{p}</M> if the prime <M>p</M> is odd, or if
 ##  <M>[<A>G</A>,<A>G</A>]</M> is contained in <M><A>G</A>^{4}</M>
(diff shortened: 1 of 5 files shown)


tst/testinstall/pgroups.tst
```diff
@@ -168,37 +168,59 @@ true
...
 > Agemo(g,5);
 > od;
-gap> Number(newList,g->IsPowerfulPGroup(g));
+gap> Number(newList,IsPowerfulPGroup);
 9
 gap> myList:=AllSmallGroups(2^4);;
-gap> Number(myList,g->IsPowerfulPGroup(g));
+gap> Number(myList,IsPowerfulPGroup);
+6
+gap> Number(myList,IsPowerfulPGroup);
```
ThomasBreuer, line 221: `IsRegularPGroup`?

lib/grp.gd
```diff
@@ -475,12 +472,42 @@ InstallTrueMethod( IsPGroup, IsGroup and IsElementaryAbelian );
...
+#P  IsRegularPGroup( <G> ) . . . . . . . . . . is a group a regular p-group ?
+##
+##  <#GAPDoc Label="IsRegularPGroup">
+##  <ManSection>
+##  <Prop Name="IsRegularPGroup" Arg='G'/>
+##
+##  <Description>
+##  <Index Key="RegularPGroup">Regular <M>p</M>-group</Index>
+##  A finite <M>p</M>-group <A>G</A> is said to be a <E>regular <M>p</M>-group</E>
+##  if for all <M>a,b</M> in <A>G</A>, one has <M>a^p b^p = (ab^p) c^p</M>
```
ThomasBreuer, line 492: `<M>a^p b^p = (ab)^p c^p</M>`?

lib/grp.gi
```diff
@@ -428,6 +428,93 @@ InstallMethod( IsPowerfulPGroup,
...
+      if a*b = b*a then continue; fi;
+
+      # regularity is also automatic if a^p * b^p = (a*b)^p
+      ap_bp := a^p * b^p;
+      ab_p := (a*b)^p;
+      if ap_bp = ab_p then continue; fi;
+
+      # if the subgroup generated H by a and b is itself regular, we are also
+      # done. However we don't use recursion, here, as H may be equal to G;
+      # and also we have to be careful to not be use too expensive code here.
```
ThomasBreuer, line 501: omit "be"?

lib/grp.gi
```diff
@@ -428,6 +428,93 @@ InstallMethod( IsPowerfulPGroup,
...
+    # see [Hal36, Theorem 2.3], [Hup67, Satz 10.13]
+    return true;
+  fi;
+
+  # Fallback to actually check the defining criterion, i.e.:
+  # for all a,b in G, we must have that a^p*b^p/(a*b)^p in (<a,b>')^p
+
+  # It suffices to pick 'a' among conjugacy class representatives.
+  # Moreover, if 'a' is central then the criterion automatically holds.
+  # For z,z'\in Z(G), if the criterion holds for (a,b) iff it holds for (az,bz').
```
ThomasBreuer, line 480: omit "if"?

ThomasBreuer (commented):
Very nice.
I have left a few minor comments.

Just for curiosity:
Why is `Agemo` not an attribute?
(`Agemo` is defined only for p-groups, why is it necessary to enter p as an argument?)

fingolfin:
`Agemo` is not an attribute because `Agemo` takes 2 or 3 arguments, so it can't be an attribute. Instead, the computed values are cached in mutable attribute `ComputedAgemos`.

I am also puzzled why `Agemo` takes the prime as argument, given that `p = PrimePGroup(G)` must always hold (and this is enforced in `Agemo`). This also conflicts with the notation in the literature... But I guess we are stuck with it now.

fingolfin:
Thanks for the careful checking and excellent feedback!

ThomasBreuer (approved):

jackschmidt:
I wrote a few more tests to improve test coverage. I didn't find any group where the last special case code applied, but the other cases all have "smallest" examples. The cases are numbered (I added debug prints to each if statement), and all but case 11 (use the definition, and get true as the answer) are very quick.

Add the following (except maybe the last two lines, case 11) to `tst/testinstall/pgroups.tst` right near the end

```
gap> IsRegularPGroup(SmallGroup(8, 3)); # Case 1
false
gap> IsRegularPGroup(SmallGroup(2187, 224)); # Case 2
false
gap> IsRegularPGroup(SmallGroup(27,3)); # Case 3
true
gap> IsRegularPGroup(SmallGroup(81, 1)); # Case 4
true
gap> IsRegularPGroup(SmallGroup(243,22)); # Case 5
true
gap> IsRegularPGroup(SmallGroup(2187, 4487)); # Case 6
true
gap> IsRegularPGroup(SmallGroup(81, 7)); # Case 7
false
gap> IsRegularPGroup(SmallGroup(78125, 684)); # Case 8
true
gap> IsRegularPGroup(SmallGroup(243, 51)); # Case 10 (no case 9 found)
false
gap> IsRegularPGroup(SmallGroup(2187, 663)); # Case 11, takes 3 seconds
true
```

The following would be a nice case to record, but it takes over a minute to run, so I'm not sure it is worth including.

```
gap> G := SmallGroup(243,22);;
gap> IsRegularPGroup(G); # Case 5
true 
gap> IsRegularPGroup(DirectProduct(G,G)); # Hup 10.3.c, Wielandt's example
false
```

I've run out of time myself. I wanted to find a case 9, and submit a pull request, or improve the speed on Wielandt's example, but perhaps this is enough.

jackschmidt:
Oh, and thank you for adding such a nice check for regular p-groups :-)  I'm also happy about the nice powerful p-group check above as well!

PR #5362, opened 2023-02-03
Make most global functions read-only to prevent accidents

fingolfin:
The first couple commits clean up various minor issues. The final commit inserts many `BindGlobal` calls, and also sometimes `DeclareGlobalName` to accommodate for recursive functions.

One case of "library function getting overwritten" remains, namely `MappedVector`; I've now at least added a FIXME comment next to it explaining the situation.


lib/grppcrep.gi
```diff
@@ -14,6 +14,14 @@
 ##
 #F MappedVector( <exp>, <list> ). . . . . . . . . . . . . . . . . . . . local
 ##
+# FIXME: the polycyclic package overwrite polycyclic with a slightly different
```
jackschmidt, line 17: Should the second "polycyclic" be "MappedVector"?

hulpke (commented):
The reason for the SMTX_ aliases was (if I recall correctly) to enable profiling of these functions.

ThomasBreuer (approved):
Most of the changes are clearly improvements.
(I have left a few minor comments.)

I am not sure about the removal of the `SMTX_Something` functions.
Probably this is harmless, and if not then we could easily reconstruct them from the `SMTX.Something` functions via a single command.

The only surprising change happens iin `lib/schursym.gi`. Apparently a lot of method installations happen inside a `Perform` call, in order not to bind certain global variables (such as `SpinDimSym` or `BasicSpinRepSym`). I am not sure that this is really a good idea,
and I am also not sure that adjusting the indentation in the file to this structure is a good idea.

(The most surprising fact in the current changes is that the function `EnumeratorOfIdeal` had been defined twice --according to the comment "bugfixes in the general code" apparently on purpose?)

ThomasBreuer:
Concerning `MappedVector`, as far as I understand the two functions from the GAP library and from the polycyclic package are equal since pull request #4973 has been merged.
Isn't is sufficient now to just remove the function from the polycyclic package?

jackschmidt:
> The only surprising change happens iin `lib/schursym.gi`. Apparently a lot of method installations happen inside a `Perform` call, in order not to bind certain global variables (such as `SpinDimSym` or `BasicSpinRepSym`). I am not sure that this is really a good idea, and I am also not sure that adjusting the indentation in the file to this structure is a good idea.

Probably the Perform thing is my fault. The code had lots of helper functions; some might be useful on their own, but some are not. Rather than decide, I used a function wrapper to get a local scope. This was used in javascript web development at the time, before there were proper "modules". I don't think it was often used in GAP; certainly not to this extent.

But `bmat`, `S`, `S3` would be very rude to make global read only variables, so some of these need to be renamed or be private/local.

I agree indenting the whole file is probably not wise (in web development this is also not done; the function wrapper is added by a program when the javascript is "compiled" by webpack or similar tools).

It is exam time in the US, so I won't have time for a week or two to consider better options, but hopefully my description of why it was done is helpful to undo it :-)

fingolfin:
Thanks for the feedback. I addressed all code comments. I've also dropped the SMTX changes as well as the indentation changes in `lib/schursym.gi` as those seems to be more controversial. I'll reintroduce them as separate PRs for discussion, so that the bulk of this PR is not blocked by them.

Regarding "indenting the whole file is probably not wise" -- I take the opposite view: to me, having intentional wrong indentation that obscures that certain code is not running at the top level is not wise.

PR #5370, opened 2023-02-07
Remove SMTX_* aliases for SMTX.* functions

fingolfin:
This used to be part of PR #5362 but @hulpke and @ThomasBreuer expressed some concern about it (see discussion there).

One remark was:

> The reason for the SMTX_ aliases was (if I recall correctly) to enable profiling of these functions.

Unfortunately I don't understand this: what's the problem with profiling e.g. `SMTX.SpinnedBasis` instead of `SMTX_SpinnedBasis` ?

Anyway, I am mostly concerned about avoiding accidental overriding of methods -- so  if there is a problem with this PR in its current form I could just change this PR to retain the `SMTX_` aliases, just making them read-only...

Well, except for `SMTX_InvariantBilinearForm` which is redefined in `recog`... Oops. But it seems to be essentially equivalent to the one in the library. I've already removed this in the recog git repository, so the next recog version will be fine in this regard.


hulpke:
Exactly. Thank you.

On Feb 7, 2023, at 06:39, Thomas Breuer ***@***.***> wrote:

﻿

** Caution: EXTERNAL Sender **

@fingolfin<https://nam10.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Ffingolfin&data=05%7C01%7CAlexander.Hulpke%40colostate.edu%7C7d0b016f64f74a076d6f08db0910aaa5%7Cafb58802ff7a4bb1ab21367ff2ecfc8b%7C0%7C0%7C638113739420570196%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=ifD69Tc%2FuZA7P5bUX3l0takkXYe46HsvgXaurZTeAQE%3D&reserved=0> My interpretation of

The reason for the SMTX_ aliases was (if I recall correctly) to enable profiling of these functions.

is that ProfileGlobalFunctions( true ) will cover also SMTX_Something but not SMTX.Something (if there is no declaration a of global variable pointing to it).

fingolfin:
Indeed, the `SMTX_` aliases were introduced in 1999-04-20 by @hulpke with commit message
> Assigned the main functions to variable names as well to permit profiling

However, @ThomasBreuer and me just checked and noticed that `ProfileGlobalFunctions` is documented to only consider functions declared via `DeclareGlobalFunction` (on a technical level: it looks at functions listed in `GLOBAL_FUNCTION_NAMES`). As far as I can tell, it was always like that.

But looking at that code led me to discover the undocumented function `ProfileFunctionsInGlobalVariables`, which indeed would honor `SMTX_*` but not `SMTX.*`. So I guess that's what this is about then.

Since it is undocumented anyway, I guess I could just extend it to add all functions in the record `SMTX`...

ThomasBreuer (approved):
Since #5378 got merged, the profiling problem for the `SMTX.Something` functions is solved, and I think this pull request should be merged.

PR #5371, opened 2023-02-07
Stop calling CHANGED_BAG during workspace restore, as it can cause crashes

ChrisJefferson:
While we are loading a workspace, gasman is kept in a slightly-inconsistent state, with various pointers only set at the end. This means we mustn't call functions like NewBag, ResizeBag and (the problem we were hitting, subtly), CHANGED_BAG.

This stops objset/objmap from calling CHANGED_BAG, and adds asserts to make sure this doesn't happen again in future (and verifies nothing else is calling any of these methods).

### DIFF
--- a/src/gasman.c
+++ b/src/gasman.c
@@ -958,6 +958,13 @@ Bag * GlobalByCookie(
 
 static Bag NextMptrRestoring;
 
+#ifdef GAP_KERNEL_DEBUG
+BOOL InWorkspaceRestore(void)
+{
+    return NextMptrRestoring != 0;
+}
+#endif
+
 void StartRestoringBags( UInt nBags, UInt maxSize)
 {
   UInt target;
@@ -1036,6 +1043,7 @@ void FinishedRestoringBags( void )
   SizeDeadBags = 0;
   NrHalfDeadBags = 0;
   ChangedBags = 0;
+  NextMptrRestoring = 0;
 }
 
 
@@ -1360,6 +1368,8 @@ Bag NewBag (
     CollectBags(0,0);
 #endif
 
+    GAP_ASSERT(!InWorkspaceRestore());
+
     CANARY_DISABLE_VALGRIND();
 
     // check that a masterpointer and enough storage are available
@@ -1537,6 +1547,8 @@ UInt ResizeBag (
 #ifdef TREMBLE_HEAP
     CollectBags(0,0);
 #endif
+
+    GAP_ASSERT(!InWorkspaceRestore());
 
     CANARY_DISABLE_VALGRIND();
 
--- a/src/gasman.h
+++ b/src/gasman.h
@@ -388,8 +388,14 @@ void CHANGED_BAG(Bag b);
 
 extern Bag * YoungBags;
 extern Bag   ChangedBags;
+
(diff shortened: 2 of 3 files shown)


src/gasman.c
```diff
@@ -958,6 +958,14 @@ Bag * GlobalByCookie(
 
 static Bag NextMptrRestoring;
 
+#ifdef GAP_KERNEL_DEBUG
+BOOL InWorkspaceRestore(void)
+{
+    // Double negation turns pointer into a boolean
+    return !!NextMptrRestoring;
```
fingolfin, line 965: or instead of being clever, we could just do an old-fashioned null pointer test? :-)
```suggestion
    return NextMptrRestoring != 0;
```

src/gasman.h
```diff
@@ -388,8 +388,14 @@ void CHANGED_BAG(Bag b);
 
 extern Bag * YoungBags;
 extern Bag   ChangedBags;
+
+#ifdef GAP_KERNEL_DEBUG
+BOOL         InWorkspaceRestore(void);
```
fingolfin, line 393:
```suggestion
BOOL InWorkspaceRestore(void);
```

src/objset.c
```diff
@@ -244,9 +244,12 @@ Int FindObjSet(Obj set, Obj obj) {
  *  Add `obj` to `set`.
  *
  *  Precondition: `set` must not already contain `obj`.
+ *
+ * inLoadWorkspace is TRUE if we are currently loading a workspace,
+ * so must not call CHANGED_BAG.
```
fingolfin, line 249: How about we don't add that argument, and instead just remove the `CHANGED_BAG` calls from `AddObj*New`, and leave it to callers to invoke them afterwards? Then we need to add calls to it in `AddObjSet` and `AddObjMap`. But not in `ResizeObjSet/Map` (in fact we can remove the `CHANGED_BAG` there and replace it with a comment saying something like `SwapMasterPoint ensures both bags are marked as changed / dirty`

src/gasman.h
ChrisJefferson: The place where I've put it is where clang-format wants to put it (I guess to line us with the variables above?), so I'd prefer to leave it there, else it might just move in future if we ever clang-format everything.

ChrisJefferson:
Now (hopefully) all fixed up

src/objset.c
```diff
@@ -721,8 +727,8 @@ static void ResizeObjMap(Obj map, UInt bits)
   for (i = 0; i < size; i++) {
     Obj obj = READ_SLOT(map, i*2);
     if (obj && obj != Undefined) {
-      AddObjMapNew(new, obj,
-        READ_SLOT(map, i*2+1));
+        AddObjMapNew(new, obj, READ_SLOT(map, i * 2 + 1));
+        CHANGED_BAG(new);
```
fingolfin, line 731: This is not needed as no GC can occur in this loop, and afterwards SwapMasterPointer ensures both bags are marked aa changed (so the two CHANGED_BAG calls after it could also go)

src/gasman.h
fingolfin: Makes sense. I thought it would be the other way around

fingolfin (approved):

PR #5373, opened 2023-02-07
Change indentation to clarify code is not toplevel

fingolfin:
This was part of PR #5362 but met some opposition, so I separated this out into its own PR so we can focus on it.

@ThomasBreuer and @jackschmidt were concerned that indenting this code is not a good idea.

I am not sure I understand why that is, though; as far as I can tell, no arguments were given ("not sure that adjusting the indentation [...] is a good idea" and "indenting the whole file is probably not wise" are not very informative :-). And the rules for JavaScript programs hardly apply to GAP code).

Here is why I think it's actually quite reasonable to do this: the modified code is *inside a function*, so indenting it as if it was on the top level is misleading, and can cause to problems when editing the code unaware of its structure (that is *precisely* what happened and led me to make this change).

Maybe one objection is that calling `InstallMethod` inside another function is unusual; but we already do that in a bunch of places (though usually with computed operation names).

An alternative to this PR would be to move most of these function definitions to the top level, possibly renaming some/all, at least the problematic ones (`S`, `S1`, `coeffS2`, `S2`, `coeffS3`, `S3`, `bmat`, `spinsteps`).

### DIFF
--- a/lib/schursym.gi
+++ b/lib/schursym.gi
@@ -23,384 +23,383 @@
 ##  of (private) global variables.
 ##
 
-Perform( [1], function(x)
+BindGlobal("BasicSpinRepSym", CallFuncList(function()
   local S, S1, coeffS2, S2, coeffS3, S3, bmat, spinsteps, SpinDimSym,
-    BasicSpinRepSymPos, BasicSpinRepSymNeg, BasicSpinRepSym,
+    BasicSpinRepSymPos, BasicSpinRepSymNeg,
     SanityCheckPos, SanityCheckNeg;
 
+  ##  let 2S+(n) = < t_1, ..., t_(n-1) > subject to the relations
+  ##    (t_i)^2 = z for 1 <= i <= n-1,
+  ##    z^2 = 1,
+  ##    ( t_i*t_(i+1) )^3 = z for 1 <= i <= n-2,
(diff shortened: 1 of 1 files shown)


jackschmidt:
I looked at this a little more and I think I found an idea that looks better to me (and hopefully Max and Thomas too).

The top block of code's purpose is to define a helper function that computes all the info needed for the representation. The implementation is supposed to match the paper, so uses names from the paper.

Instead of hiding even the public interface in the Perform block, have that block return the single function `BasicSpinRepSym`.

Then the install methods can use just this one (low level, undocumented) function.

I would suggest changing the name of BasicSpinRepSym so that it doesn't appear in tab-completion before the documented function.

I haven't enough time to make the pull request, but I was thinking `Perform([1],function(ignored) ... end)` is replaced with `BindGlobal("PrivateBasicSpinRepSym", CallFuncList( function() .... return BasicSpinRepSym; end, []) )`

and the install methods are moved outside the private scope.  The private scope would then only have the functions whose names closely match the paper and handle the recursive construction. Outside that scope is the more public-friendly interface.

However, I don't have strong feelings about white space and indenting is fine :-)

I hope my suggested change makes the division of the file more logical: indented code is the recursive definition with names from the paper; un-indented top-level code is mostly method installations or short simple methods.

fingolfin:
@jackschmidt I like your suggestion and have implemented it now

jackschmidt:
Thanks! Looks good

PR #5374, opened 2023-02-08
use `MutableCopyMatrix` not `MutableCopyMat`

ThomasBreuer:
As suggested in the discussion of #5352,
`MutableCopyMat` is just an (obsolete) synonym of `MutableCopyMatrix` and therefore `MutableCopyMat` should not be used in GAP library code.


fingolfin:
Perhaps also document this in `doc/ref/obsolete.xml`?

fingolfin (approved):

Issue #5372, opened 2023-02-07
Improve "extension features" of GAP packages by turning them into explicit dependencies


ThomasBreuer:
(This idea arose from a discussion with @fingolfin. Note that the Julia package manager now provides a similar mechanism.)

Currently several packages install additional features provided that other packages are known to be available later in the GAP session.
A typical example is CTblLib, which reads certain package files that depend on the availability of the Browse package:
Since one wants to be able to use CTblLib without Browse, CTblLib lists Browse only as a "suggested" package but not as a "needed" package in the dependencies in `PackageInfo.g`. The `ReadPackage` calls for the additional files are executed only if `IsPackageMarkedForLoading` returns `true` for Browse (in a sufficiently new version).

This works well if loading suggested packages is enabled and if the suggested packages are available at the time when the `IsPackageMarkedForLoading` calls are executed.
It does not work well if one starts a GAP session, loads certain packages, and then installs more packages.
It does not work well if one starts a GAP session without packages and then loads some packages by hand.
With the current setup, if CTblLib has been loaded in a situation when Browse was not available then the Browse dependent code will never become available in the GAP session, also if one (installs and) loads Browse later during the session.

We could improve this situation by extending the dependencies in `PackageInfo.g` by a new list `Extensions`; in a simple form, each entry could have the format `[ <packagename>, <version>, <filename> ]`, meaning that the file <filename> (relative to the package directory) shall read as soon as the package `<packagename>` has been loaded in version (at least) `<version>`.

In the above example, the file <filename> of CTblLib would consist of `ReadPackage` calls for the Browse dependent code.
The function `LoadPackage` would get modified as follows:
When `LoadPackage` has loaded CTblLib, the `Extensions` list of CTblLib's `PackageInfo.g` (together with the information to which package the extension belongs) gets appended to a global list of pending extensions of packages, and then this whole list gets evaluated --whenever the condition of an entry is satisfied, the file in question gets read, and the entry gets removed from the list.
If Browse gets loaded via `LoadPackage` after CTblLib has been loaded, this `LoadPackage` call will find the list entry that has been added when CTblLib was loaded, and trigger that the extension becomes available.
(Thinking about pending extensions in saved workspaces, one does not want to replace the given relative filenames <filename> by absolute ones before actually reading these files.)

Perhaps this approach is too simpleminded.
- One can think about more involved conditions: Some feature may depend on more than one other package, then one could store lists of package specifications.
[...]

PR #5375, opened 2023-02-08
Support package extensions (see the manual section "Extensions Provided by a Package")

ThomasBreuer:
The feature is  proposed in issue #5372.

~~Eventually the feature must be documented.
(And the tile must be extended by a link to the relevant manual section, in order to serve as an entry for the release notes.)~~
Concerning the documentatiion, as far as I understand, it is recommended to add the new component in `PackageInfo.g` files in the `PackageInfo.g` file of the Example package, and to support the component in PackageMaker,
but there is no exhaustive list of supported components in `PackageInfo.g` files in the Reference Manual.

### DIFF
--- a/doc/ref/gappkg.xml
+++ b/doc/ref/gappkg.xml
@@ -1184,6 +1184,57 @@ is now a needed package for &GAP; itself.
 </Section>
 
 
+<!-- %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% -->
+<Section Label="Extensions Provided by a Package">
+<Heading>Extensions Provided by a Package</Heading>
+
+Sometimes a package <C>A</C> can provide additional functionality,
+such as better methods or additional data,
+if some other packages <C>B</C>, <C>C</C>, etc. are loaded.
+However, one wants that package <C>A</C> can be used also without these
+packages,
+and therefore <C>B</C>, <C>C</C>, etc. shall not be regarded as needed
+packages (see Section <Ref Sect="Package dependencies"/>) of <C>A</C>.
+
+<P/>
+
+One way to deal with this situation is to put those parts of code of <C>A</C>
+that depend on <C>B</C>, <C>C</C>, etc., into files that get read only
+in the situation that the packages in question have actually been loaded
+into the current &GAP; session,
+and to notify this conditional <E>extension</E> of <C>A</C> via a record in
+the <C>Extensions</C> component of the <F>PackageInfo.g</F> file of <C>A</C>;
+this record must have the components <C>needed</C> (a list of the form
+<C>[ [ pkgname1, version1 ], [ pkgname2, version2 ], </C><M>\ldots</M><C> ]</C>,
+meaning that the extension shall be loaded as soon as all packages
+<C>pkgname1</C>, <C>pkgname2</C>, <M>\ldots</M>, with versions (at least)
+<C>version1</C>, <C>version2</C>, <M>\ldots</M>, have been loaded)
+and <C>filename</C> (the path, relative to the package directory of <C>A</C>,
+of a file such that reading this file will load the code of the extension).
+Calls of <Ref Func="LoadPackage"/> trigger that package extensions will
+be loaded as soon as their conditions are satisfied.
+
+<P/>
+
+For example, the package <C>A</C> can be loaded early in a &GAP; session,
+and announce an extension that requires the package <C>B</C>.
+If <C>B</C> has not yet been loaded then this extension will not be loaded
+together with <C>A</C>.
+However, as soon as <C>B</C> gets (installed and) loaded later in the
+session, also the extension of <C>A</C> will automatically get loaded.
+
+<P/>
+
+The contents of <C>Extensions</C> in a <F>PackageInfo.g</F> file
+does not affect the lists of needed or suggested packages.
+If an extension of <C>A</C> is beneficial for the functions of <C>A</C>
+then it makes sense to list the packages needed for the extension among the
+suggested packages of <C>A</C>,
+but this may not be the case if the extension is beneficial only for the
+functions of its needed packages.
+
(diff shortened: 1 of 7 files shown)


fingolfin:
Adding an exhaustive list of supported PackageInfo.g entries somewhere is one of our open issues (It could be in the reference manual, or in a separate "package author guides", or whatever; once it exists, moving it wouldn't be hard either).

lib/package.gi
```diff
@@ -1677,6 +1677,17 @@ InstallGlobalFunction( LoadPackage, function( arg )
         # Notify the documentation (for the available version).
         LoadPackageDocumentation( info );
 
+        # Notify extensions provided by the package.
+        if IsBound( info.Dependencies ) and
+           IsBound( info.Dependencies.Extensions ) then
+          for entry in info.Dependencies.Extensions do
```
fingolfin, line 1683: Extensions are not dependencies, so why `info.Dependencies.Extensions` instead of `info.Extensions`?

lib/package.gi
```diff
@@ -1706,7 +1717,20 @@ InstallGlobalFunction( LoadPackage, function( arg )
 
     od;
 
-    if not IsBound( GAPInfo.LibraryLoaded ) then
+    if IsBound( GAPInfo.LibraryLoaded ) then
+      # Load those package extensions whose condition is satisfied.
+      for i in [ 1 .. Length( GAPInfo.PackageExtensionsPending ) ] do
+        entry:= GAPInfo.PackageExtensionsPending[i];
+        if IsPackageLoaded( entry[1], entry[2] ) then
```
fingolfin, line 1724: What if an extension needs more than one package? Perhaps it should be possible to provide a list of dependencies, not just a single one?

doc/ref/gappkg.xml
```diff
@@ -997,6 +997,57 @@ is now a needed package for &GAP; itself.
...
+packages,
+and therefore <C>B</C>, <C>C</C>, etc. shall not be regarded as needed
+packages (see Section <Ref Sect="Package dependencies"/>) of <C>A</C>.
+
+<P/>
+
+One way to deal with this situation is to put those parts of code of <C>A</C>
+that depend on <C>B</C>, <C>C</C>, etc., into files that get read only
+in the situation that the packages in question have actually been loaded
+into the current &GAP; session,
```
fingolfin, line 1017: Maybe stop the sentence here, and insert a brief excursion, something in this vein (exact text can of course be completely different)
```suggestion
into the current &GAP; session.
<P/>

However, this leaves the question when to load these files. In the past, the only
option for <C>A</C> was to check for the presence of <C>B</C>, <C>C</C>, etc., while
it itself was being loaded. The result of this is that which features are available or not depends on the order in which packages get loaded: If <C>B</C> is loaded before <C>A</C>, the extension might be loaded as well; if <C>B</C> is loaded only after <C>A</C>, then the extension is not loaded.
<P/>

To deal with this, &GAP; offers a dedicated mechanism for packages to deal
with extensions:
```

doc/ref/gappkg.xml
```diff
@@ -997,6 +997,57 @@ is now a needed package for &GAP; itself.
...
+and to notify this conditional <E>extension</E> of <C>A</C> via a record in
```
fingolfin, line 1018: The extension is not really "notified" of anything, is it? It get loaded.

Or maybe you meant to say that GAP is informed about the existence of such an extension? So maybe:
```suggestion
and to signal the existence of this conditional <E>extension</E> of <C>A</C> via a record in
```

doc/ref/gappkg.xml
```diff
@@ -997,6 +997,57 @@ is now a needed package for &GAP; itself.
...
+the <C>Extensions</C> component of the <F>PackageInfo.g</F> file of <C>A</C>;
+this record must have the components <C>needed</C> (a list of the form
```
fingolfin, line 1020: I think I'd find it more readable if the fields were described as entries in a list

doc/ref/gappkg.xml
```diff
@@ -997,6 +997,57 @@ is now a needed package for &GAP; itself.
...
+<C>[ [ pkgname1, version1 ], [ pkgname2, version2 ], </C><M>\ldots</M><C> ]</C>,
+meaning that the extension shall be loaded as soon as all packages
+<C>pkgname1</C>, <C>pkgname2</C>, <M>\ldots</M>, with versions (at least)
+<C>version1</C>, <C>version2</C>, <M>\ldots</M>, have been loaded)
+and <C>filename</C> (the path, relative to the package directory of <C>A</C>,
+of a file such that reading this file will load the code of the extension).
+Calls of <Ref Func="LoadPackage"/> trigger that package extensions will
+be loaded as soon as their conditions are satisfied.
```
fingolfin, line 1028:
```suggestion
Whenever <Ref Func="LoadPackage"/> is called, &GAP; checks for package
extensions whose conditions now are satisfied, and loads them.
```

doc/ref/gappkg.xml
```diff
@@ -997,6 +997,57 @@ is now a needed package for &GAP; itself.
...
+
+<P/>
+
+For example, the package <C>A</C> can be loaded early in a &GAP; session,
+and announce an extension that requires the package <C>B</C>.
```
fingolfin, line 1033: The "announce" bit makes it sound as if the package has to actively do something (as in: execute some code), but in reality this is passive (the metadata is static). So perhaps
```suggestion
and declare in its <F>PackageInfo.g</F> the availability of an extension
that requires package <C>B</C>.
```

fingolfin (approved):
Great!

I've made some suggestions for the docs. Happy to discuss them, perhaps someone has better suggestions

ThomasBreuer:
The failure of "Wrap releases ..." is due to an inconsistency of https://github.com/gap-system/PackageDistro/releases/download/latest/package-infos.json.gz.sha256 and https://github.com/gap-system/PackageDistro/releases/download/latest/package-infos.json.gz

fingolfin (approved):

PR #5376, opened 2023-02-10
document the supported components of `PackageInfo.g`

ThomasBreuer:
Having an exhaustive list of supported `PackageInfo.g` entries as an official reference is desirable.

The list proposed here was derived from the function `ValidatePackageInfo` and the contents of the `PackageInfo.g` file of the `Example` package.
After discussing this proposal and fixing the contents which we really want to have, these sources should then be adjusted to the list in the Reference Manual.

(Should also the `AutoDoc` component be mentioned?
When #5375 will be merged, also the `Extensions` component must be mentioned.)


fingolfin (approved):
Great!

PR #5380, opened 2023-02-14
Remove `SmallGroupsInformation(512)` test

fingolfin:
All this tests is the output of SmallGroupsInformation(512),
which changed in the most recent smallgrp update.

The smallgrp package now lives in its own repository and has
its own test suite and CI setup, and tests this function and
much more. So there is no point in GAP doing this as well.

Ironically, the test here was originally added to verify
another change to the output of that command...


ThomasBreuer (approved):
Good solution.

PR #5389, opened 2023-03-05
Added testfiles ...

ThomasBreuer:
... which contain the input and output of the GAP related examples in the book [Representations of Groups - A Computational Approach](https://www.cambridge.org/catalogue/catalogue.asp?isbn=9780521768078).


fingolfin:
I am not sure what the intent of this is, though? In the current form, I don't think these tests will ever be run? This would be different if they were inside the `tst/testinstall/` directory.

ThomasBreuer:
@fingolfin Yes, once it is clear that we want to use these examples as tests, the next question is when they shall be run. The `tst/testinstall` tests are perhaps not the right place because a few of the examples do not run quickly.

ChrisJefferson:
The obvious places would be teststandard or testextra, depending on how long they take -- while testextra isn't run very often, I do try to run it every so often.

ThomasBreuer:
@ChrisJefferson Thanks. I will move the files into `tst/teststandard`.

PR #5344, opened 2023-01-26
Support method dispatch based on tags (see the manual section "Tag Based Operations")

ThomasBreuer:
In calls to operations such as `DiagonalMatrix`, `CompanionMatrix`, etc., one wants to specify the internal representation of the result via a filter (such as `IsPlistMatrixRep`) as the first argument. In order to be able to install individual methods depending on this filter, one does not really want to declare the operations as [constructors](https://docs.gap-system.org/doc/ref/chap78_mj.html#X86EC0F0A78ECBC10) because the result shall have *exactly* the given filter (and not perhaps something better), and also because one wants to be able to install a default method. (See the discussion of #4398 for details.)

The changes proposed here provide an implementation (see commit 949cbc4) and documentation (see commit 9aaf40f) for this feature. Note that the operations in question are still GAP operations, the tag based mechanism is called only in the situation that the first argument is a filter.

The rest of the pull request applies the new mechanism to the operation `DiagonalMatrix`:

- added a new operation `DiagonalMatrix`, and added its documentation to the Reference Manual
- added a new filter `IsRowVectorOrVectorObj`, analogous to `IsMatrixOrMatrixObj`, in order to simplify some method installations
- moved the declarations of `DefaultVectorRepForBaseDomain` and `DefaultMatrixRepForBaseDomain` from `lib/matobj.gi` to `lib/matobj2.gd`
- added a `CompatibleVectorFilter` method for `IsMatrix` objects
- added `NewZeroMatrix` and `NewIdentityMatrix` methods for `IsPlistRep`
- added preliminary tests for `DiagonalMatrix`
- removed some comment lines

As soon as we agree on the details, the same changes can be applied to `PermutationMatrix`, `ReflectionMatrix`, `RandomMatrix`, see issue #4367.

Here are the perhaps controversial pieces of the changes:

- We could decide that `DiagonalMatrix( list )` behaves like the global function `DiagonalMat( list )`, and then `DiagonalMat` could become deprecated.
- It is not clear to me to which extent we want `IsPlistRep` matrices to be supported w.r.t. functionality for matrix objects. Thus perhaps the "auxiliary" `NewZeroMatrix` and `NewIdentityMatrix` methods for `IsPlistRep` and the `CompatibleVectorFilter` method for `IsMatrix` objects are not what we want.

Issue #4398, opened 2021-04-14
MatrixObj: concerns regarding constructors like NewZeroMatrix, NewVector, etc. 


fingolfin:
Currently, to construct MatrixObj objects (I'll include VectorObj in that for simplicity), it always boils down to a call of a constructor like `NewZeroMatrix`, `NewVector`, etc.

After working a bit on tests for MatrixObj and trying to use MatrixObj, I have a few concerns about doing it this way:

1. constructor lookup is linear: when you invoke `NewZeroMatrix(myrep, GF(2), 5)` then GAP scans the list of methods installed for `NewZeroMatrix` linearly, each time. So depending on where in that list `myrep` appears, it can be faster or slower. And if we ever have a lot of matrixobj implementations, the difference may grow quite a bit.
2. constructor lookup is still method dispatch, so it has to determine the type of all inputs; for `NewVector` and `NewMatrix`, these are lists (possibly lists of lists), and determining the type of such a list has overhead; at the same time, it is usually (always?) unnecessary here
3. there is no way (as far as I could tell) to install a "generic" constructor: e.g. I wanted to provide a generic version of `NewIdentityMatrix`, which invokes `NewZeroMatrix` and then sets the diagonal entries to one. But I found no way to install such a generic constructor. That makes implementing the matrixobj API more work, with no apparent good reason: most (all???) MatrixObj implementations won't have any need or desire to install custom methods for `NewIdentityMatrix`, so it's annoying they are forced to provide one.

Regarding 1.: I may worry unnecessarily, the time taken to perform the construction may outweigh the method dispatch overhead by several orders of magnitude. It would be instructive to at least perform some benchmarks, though.

Regarding 2.: Again, some benchmarks would be usful.

Regarding 3.: One way out of this would be to just drop `NewIdentityMatrix` (instead put the generic code into the generic `IdentityMatrix`) and likewise `NewCompanionMatrix`, and simply change `IdentityMatrix` and `CompanionMatrix` to first create a zero matrix, then fill that. Another option would be to have a global function `GenericNewIdentityMatrix` and then MatrixObj implementations can use it (still annoying, but less than the current status).

Another approach altogether would be to do away with the various constructors, and instead add a global `OBJ_MAP` which maps filters like `IsGF2MatrixRep` to a record; that record contains entries for the various constructors. Say: `rec(newVector:=..., newZeroVector:=..., ...`. Then we could have code like this:

```gap
MAP_OF_MATRIX_REP := OBJ_MAP()
[...]

PR #5396, opened 2023-03-08
use early methods for tag based operations

ThomasBreuer:
This is a follow-up of #5344, according to a comment in #4398.

### DIFF
--- a/lib/methsel.g
+++ b/lib/methsel.g
@@ -135,9 +135,10 @@ end );
 ##  <P/>
 ##  Installing methods with <Ref Func="InstallMethod"/> for a tag based
 ##  operation is possible.
-##  (Installing such methods with the same requirements as the ones for the
-##  tag based methods is not recommended, because this may lead to unwanted
-##  effects.)
+##  However, installing such methods with the same requirements as the ones
+##  for the tag based methods will have no effect because the handling of
+##  tag based methods gets installed with <Ref Func="InstallEarlyMethod"/>
+##  and thus has higher priority.
 ##  </Description>
 ##  </ManSection>
 ##  <#/GAPDoc>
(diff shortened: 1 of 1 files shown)


fingolfin:
The documentation test fails in the line
```gap
d1:= DiagonalMatrix( GF(9), [ 1, 2 ] * Z(3)^0 );
```
Both @ThomasBreuer and me had trouble reproducing this, so I went and did exactly what the CI does for testing the manuals (i.e. following the steps in `dev/ci.sh` for the `testmanuals` test suite...

... and the problem seems to be caused by saving and restoring workspaces:
```
$ ./gap -b -A
gap> SaveWorkspace("testmanuals.wsp");
true
gap> QUIT;
$ ./gap -b -L testmanuals.wsp
gap> d1:= DiagonalMatrix( GF(9), [ 1, 2 ] * Z(3)^0 );
Error, no 1st choice method found for `DiagonalMatrix' on 3 arguments at /Users/mhorn/Projekte/GAP/gap/lib/methsel2.g:250 called from
DiagonalMatrix( DefaultMatrixRepForBaseDomain( R ), R, vector
```
I've not yet dug deeper from here, I first need some sleep.

ChrisJefferson:
The workspace issue is fixed (hopefully!) in #5400

ThomasBreuer:
@ChrisJefferson Thanks a lot for the fix.
As far as I understand, the manual tests are currently the only place where GAP workspaces are exercised.
Would it make sense to add a test that first creates a workspace and then runs the tests from `testinstall` and `teststandard` with this workspace?

fingolfin:
@ThomasBreuer Yes that would make sense.

fingolfin (approved):

PR #5398, opened 2023-03-08
Remove obsolete placeholders for group library features

### LINKED ISSUE #5200: Remove obsolete uses of `AllGroups` and `OneGroup` from packages and library
... and then eventually also remove them from `smallgrp`.

fingolfin:
Resolves #5200


ThomasBreuer (approved):
Yes, these variables are not used in the GAP library.
Those which are overwritten by the needed packages of GAP can be used by GAP packages also after these changes.

PR #5400, opened 2023-03-09
Fix workspaces to save and restore "early methods" of operations

ChrisJefferson:
When earlyMethods were added, they were not added to the save / load workspace functions, so they were lost when a workspace is loaded.

Allows #5396 to work when a workspace is saved then loaded.

### DIFF
--- a/src/opers.cc
+++ b/src/opers.cc
@@ -2962,6 +2962,8 @@ void SaveOperationExtras (
         SaveSubObj(header->methods[i]);
     for (UInt i = 0; i <= MAX_OPER_ARGS; i++)
         SaveSubObj(header->cache[i]);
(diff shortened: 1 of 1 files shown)


fingolfin:
Ouch! Good catch. Early methods are in GAP 4.12.x, and nobody seems to have noticed so far. Probably because it is only use for matrix obj methods nobody really uses (yet), and for `First`, which has fallback methods that can do the job -- so it'd be slower in a restored workspace, but not unusable.

I'll mark this as "backport", just in case we make a 4.12.3 release before 4.13.0 (right now I have no plans to do that, but who knows...)

fingolfin (approved):

PR #5407, opened 2023-03-13
Add user preference "ShortBanners"

zickgraf:
If this option is set to <K>true</K>, package banners printed during
loading will only show the name, version and description of a package.

Rationale: The package banners contain lots of information which one usually does not need during development (authors, homepage, link to the issue tracker) and which clutters the terminal, sometimes hiding information which one would actually need (like syntax warnings/errors).

The first commit simply removes a duplicate whitespace. If this whitespace is deliberate, I can of course add it back, although I would suggest to remove it at least for the short banners because I think it looks a bit out of place there.

## Text for release notes

Add user preference "ShortBanners"

If this option is set to `true`, package banners printed during
loading will only show the name, version and description of a package.

### DIFF
--- a/lib/package.gi
+++ b/lib/package.gi
@@ -1099,8 +1099,29 @@ InstallGlobalFunction( IsPackageMarkedForLoading, function( name, version )
 ##
 #F  DefaultPackageBannerString( <inforec> )
 ##
-InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
+
+DeclareUserPreference( rec(
+  name:= "ShortBanners",
+  description:= [
+    "If this option is set to <K>true</K>, package banners printed during \
+loading will only show the name, version and description of a package."
+    ],
+  default:= false,
+  values:= [ true, false ],
+  multi:= false,
+  ) );
+
+InstallGlobalFunction( DefaultPackageBannerString,
+    function( inforec, useShortBanner... )
     local len, sep, i, str, authors, maintainers, contributors, printPersons;
+
+    if Length( useShortBanner ) = 0 then
+      useShortBanner := false;
+    elif Length( useShortBanner ) = 1 then
+      useShortBanner := useShortBanner[1];
+    else
+      Error( "DefaultPackageBannerString must be called with at most two arguments" );
+    fi;
 
     # Start with a row of `-' signs.
     len:= SizeScreen()[1] - 3;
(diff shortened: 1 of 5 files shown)


zickgraf:
The CI is failing due to the removed duplicate space, but I will wait for feedback before updating the tests.

lib/package.gi
```diff
@@ -1099,6 +1099,19 @@ InstallGlobalFunction( IsPackageMarkedForLoading, function( name, version )
 ##
 #F  DefaultPackageBannerString( <inforec> )
 ##
+
+DeclareUserPreference( rec(
+  name:= "ShortBanners",
+  description:= [
+    "If this option is set to <K>true</K>, only the first line of each \
+package banner (including the name, version and description of the package) \
+is displayed."
```
fingolfin, line 1108: That is only true for packages that use the default banner, is it? Or do you also do something for packages with a custom banner? But then this text also ought to be adjusted.

lib/package.gi
```diff
@@ -1124,7 +1137,7 @@ InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
     # Add package name and version number.
     if IsBound( inforec.PackageName ) and IsBound( inforec.Version ) then
       Append( str, Concatenation(
-              "Loading  ", inforec.PackageName, " ", inforec.Version ) );
+              "Loading ", inforec.PackageName, " ", inforec.Version ) );
```
fingolfin, line 1140: I am somewhat wary that this might break package tests. Although a quick look suggests that this is not the case, so perhaps all is good in that regard.

I can't think of a good reason for the extra space otherwise.

lib/package.gi
```diff
@@ -1099,6 +1099,19 @@ InstallGlobalFunction( IsPackageMarkedForLoading, function( name, version )
...
+    "If this option is set to <K>true</K>, only the first line of each \
+package banner (including the name, version and description of the package) \
+is displayed."
+    ],
+  default:= false,
+  values:= [ true, false ],
+  multi:= false,
+  ) );
+
 InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
```
fingolfin, line 1115: A couple packages "patch" the output of `DefaultPackageBannerString`. Changing its content globally would likely break those.

Instead of having `DefaultPackageBannerString` depend on a user preference, we could add a second optional argument `useShortBanner` (or whatever); if missing it'd default to `false`.

Then you can adjust the one call to `DefaultPackageBannerString ` in `lib/package.gi` to pass in `UserPreference( "ShortBanners" )` as that second argument.

lib/package.gi
zickgraf: Ah, I wasn't aware of custom package banners. Fixed.

lib/package.gi
zickgraf: Done :-)

lib/package.gi
zickgraf: I have now (hopefully) fixed the tests in GAP. Maybe we can merge and see how many packages are affected by this, if any at all? If the number is small I could open PRs, otherwise we could revert.

lib/package.gi
```diff
@@ -1099,9 +1099,31 @@ InstallGlobalFunction( IsPackageMarkedForLoading, function( name, version )
...
 #F  DefaultPackageBannerString( <inforec> )
 ##
-InstallGlobalFunction( DefaultPackageBannerString, function( inforec )
+
+DeclareUserPreference( rec(
+  name:= "ShortBanners",
+  description:= [
+    "If this option is set to <K>true</K>, only the first line of each \
+package banner (including the name, version and description of the package) \
+is printed for packages using the default banner."
```
fingolfin, line 1108: OK, I've thought some more about this, and I wonder if it wouldn't be more useful if this option affected *all* packages. I.e., if it is on, then all packages just get a shortened banner printed when loaded. To implement this, instead of modifying `DefaultPackageBannerString` we could add a separate function, say `ShortPackageBannerString`. And then the code which prints the package banner makes a high-level decision what to do

lib/package.gi
```diff
@@ -1429,7 +1455,8 @@ BindGlobal( "LoadPackage_ReadImplementationParts",
         elif IsBound( info.BannerString ) then
```
fingolfin, line 1455: So in this code, the first `if` could be changed to 
```gap
if UserPreference( "ShortBanners" ) then
  bannerstring := ShortPackageBannerString(info);
elif IsBound( info.BannerFunction ) then
  ...
```

Thoughts?

zickgraf: So custom banners would be complete ignored when `ShortBanners` is set to `true`? I just looked at some packages and, for example, `curlInterface` adds the versions of various libraries to the banner. I guess that this information is printed for debugging and in particular to have this information available in bug reports? If so, I think ignoring the custom banner would be suboptimal.

zickgraf: A clean interface would maybe be that packages can specify additional information which gets included in the banner instead of letting `BannerFunction` patch `DefaultPackageBannerString`. And of course this additional information could then come in two variants, a long and a short version. But this would be beyond what I had planned for this PR :D

fingolfin: The curl version can also be obtained in other ways.

And I'd try not to overthink this problem. Honestly, the whole quest for "short banners" is already extremely fringe. Most people will either not mind the long banner, or just turn banners off. Having a middle ground is not something I can see a lot of people having much interest in (I might be totally wrong, of course)

fingolfin:
Initiated PackageDistro CI run against this branch here: https://github.com/gap-system/PackageDistro/actions/runs/4494333014

ChrisJefferson (approved):
Just to say I don't need this, but I also know some packages do print out an awful lot when they, and all their dependancies, are loaded!

zickgraf:
Thanks @ChrisJefferson for the review!

Maybe the PR could still make it into GAP 4.13? Or if there is not enough of an agreement that the option is useful, I would also be fine with just closing the PR.

fingolfin (commented):
It doesn't seem to cause any package tests to fail. As long as this is just an option, I see no harm either.

zickgraf:
Thanks @fingolfin for the merge!

PR #5408, opened 2023-03-14
add a fallback method for `NiceMonomorphism`

ThomasBreuer:
`NiceMonomorphism` shall work for finite rational matrix groups, also if they do not know yet that they are finite.

(The problem had be mentioned in a message by Hongyi Zhao in a message to the GAP Forum.)

### DIFF
--- a/lib/grpramat.gi
+++ b/lib/grpramat.gi
@@ -496,3 +496,7 @@ RedispatchOnCondition( IsomorphismPcGroup, true,
 RedispatchOnCondition( CompositionSeries, true,
     [ IsCyclotomicMatrixGroup ],
     [ IsFinite ], 0 );
+
+RedispatchOnCondition( NiceMonomorphism, true,
+    [ IsCyclotomicMatrixGroup ],
+    [ IsFinite ], 0 );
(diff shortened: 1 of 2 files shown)


hongyi-zhao:
If I remember correctly, it seems that there are many similar functions and methods in GAP, which all rely on the results of finiteness test to work. So, is there a better solution to solve these problems, whether now or in the future, once and for all?

ThomasBreuer:
@hongyi-zhao The file `lib/grpramat.gi` contains already `RedispatchOnCondition` calls that trigger `IsFinite` tests for the usual operations; `NiceMonomorphism` had been missing up to now.

hongyi-zhao:
@ThomasBreuer 

> @hongyi-zhao The file `lib/grpramat.gi` contains already `RedispatchOnCondition` calls that trigger `IsFinite` tests for the usual operations; `NiceMonomorphism` had been missing up to now.

Thank you for your explanation again. I've already read your PR code, and it's just that I'm not familiar with this command, which has caused my doubts.

fingolfin (approved):

PR #5427, opened 2023-03-29
Add type checks to Concatenation

zickgraf:
In particular, this lets the undefined expression `Concatenation( [ 1 ] )` throw an error instead of returning `1`.

The first commit fixes some inconsistencies I discovered while grepping for functions similar to `Concatenation`.

## Text for release notes

none

### DIFF
--- a/lib/coll.gd
+++ b/lib/coll.gd
@@ -2979,14 +2979,14 @@ DeclareOperation( "IsSubset", [ IsListOrCollection, IsListOrCollection ] );
 
 #############################################################################
 ##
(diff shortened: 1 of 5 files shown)


fingolfin:
Can you add a testcase (perhaps in a new file in `tst/testbugfix` ?), e.g. `Concatenation( [ 1 ] )`?

zickgraf:
> Can you add a testcase (perhaps in a new file in `tst/testbugfix` ?), e.g. `Concatenation( [ 1 ] )`?

Done. I don't think this is a bugfix in the strict sense, so I have added a file in `tst/testinstall/opers` (this directory seems to also cover functions which are not operations).

PR #5432, opened 2023-04-04
fix `AbelianGroup` w.r.t. generators of order 1

### LINKED ISSUE #5430: generators of groups created with `AbelianGroup`
The following looks rather confusing.
The generators of the groups generated by `AbelianGroup` do not always correspond to the prescribed exponents.
```
gap> AbelianGroup( [ 2, 3, 4 ] );
<pc group of size 24 with 3 generators>
gap> AbelianGroup( [ 1, 2, 3 ]);
<pc group of size 6 with 3 generators>
gap> AbelianGroup( [ 1, 2 ] );
<pc group of size 2 with 2 generators>
gap> AbelianGroup( [ 1 ] );
<pc group of size 1 with 0 generators>
gap> AbelianGroup( [ 1, 1 ] );
<pc group of size 1 with 0 generators>
gap> AbelianGroup( [ 1, 1, 2 ] );
<pc group of size 2 with 3 generators>
```
If one prescribes a result in `IsPermGroup`, the situation is different.
```
gap> AbelianGroup(IsPermGroup, [ 1, 1 ] );
Group(())
gap> AbelianGroup(IsPermGroup, [ 1, 2 ] );
Group([ (), (1,2) ])
gap> AbelianGroup(IsPermGroup, [ 1 ] );
Group(())
gap> AbelianGroup(IsPermGroup, [ 1, 1, 1 ] );
Group(())
```
And an empty list of exponents works for `IsPcGroup` (which is the default) but is not allowed for `IsPermGroup` and `IsFpGroup`.
```
gap> AbelianGroup( [] );
<pc group of size 1 with 0 generators>
gap> AbelianGroup( IsPcGroup, [] );
<pc group of size 1 with 0 generators>
gap> AbelianGroup( IsPermGroup, [] );
Error, <arg> must be nonempty
[...]
gap> AbelianGroup( IsFpGroup, [] );
Error, <ints> must be a list of integers
[...]
```
It would be helpful if we could rely on a correspondence of the generators with the given exponents. The statement in the documentation

> The generators of the group returned are the elements corresponding to the factors `Cints[i]` and hence the integers in `ints`.

can be understood as the promise that this should hold.

ThomasBreuer:
- support `AbelianGroup( IsPermGroup, [] )` and `AbelianGroup( IsFpGroup, [] )`
- changed the output of `AbelianGroup( IsPcGroup, ints )` such that also generators of order 1 appear

resolves #5430

### DIFF
--- a/grp/basicfp.gi
+++ b/grp/basicfp.gi
@@ -32,7 +32,7 @@ InstallMethod( AbelianGroupCons, "fp group", true,
 function( filter, ints )
 local   f,g,i,j,rels,gfam,fam;
 
-  if Length(ints)=0 or not ForAll( ints, x -> IsInfinity(x) or (IsInt(x) and x >= 0) )  then
+  if not ForAll( ints, x -> IsInfinity(x) or (IsInt(x) and x >= 0) )  then
       Error( "<ints> must be a list of integers" );
   fi;
 
--- a/grp/basicpcg.gi
+++ b/grp/basicpcg.gi
@@ -44,7 +44,12 @@ local   pis,  f,  g,  r,  k,  pi,  i,  geni,  j,  name,  ps;
     fi;
     if ForAll(ints,i->i=1) then
       # the stupid trivial group case
-      return CyclicGroup( IsPcGroup, 1 );
+      g:= CyclicGroup( IsPcGroup, 1 );
+      if Length( ints ) > 0 then
+        g:= GroupWithGenerators(
+                ListWithIdenticalEntries( Length( ints ), One( g ) ) );
+      fi;
+      return g;
     fi;
 
     pis := List( ints, Factors );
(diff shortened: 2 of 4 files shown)


grp/basicfp.gi
```diff
@@ -32,7 +32,7 @@ InstallMethod( AbelianGroupCons, "fp group", true,
 function( filter, ints )
 local   f,g,i,j,rels,gfam,fam;
 
-  if Length(ints)=0 or not ForAll( ints, x -> IsInfinity(x) or (IsInt(x) and x >= 0) )  then
+  if not ForAll( ints, x -> IsInfinity(x) or (IsInt(x) and x >= 0) )  then
       Error( "<ints> must be a list of integers" );
```
fingolfin, line 36: Off-topic (not a change request, but perhaps we can some day address it): I guess strictly speaking this error is not accurate (as we allow infinity these days).

fingolfin (approved):

PR #5433, opened 2023-04-05
Multiclass identification in Dixon and minor cleanup (by Alexander Hulpke)

fingolfin:
This contains those changes from PR #5394 which I think are uncontroversial, and so once the CI for this PR passes, I am happy to merge right away.

### DIFF
--- a/lib/ctblperm.gi
+++ b/lib/ctblperm.gi
@@ -38,7 +38,7 @@ end );
 ##  The class invariant consists of the cycle structure and - if computation
 ##  might improve results - of the Fingerprint of the permutation
 ##
-BindGlobal( "IdentificationPermGroup", function(D,el)
+BindGlobal("CheapIdentificationPermGroup",function(D,el)
   local s,t,i,l; # guter Programmier s t i l !
   s:=CycleStructurePerm(el);
   s:=ShallowCopy(s);
@@ -58,13 +58,19 @@ BindGlobal( "IdentificationPermGroup", function(D,el)
     Add(s,-FingerprintPerm(D,el,D.p1,D.p2,D.fingerprintOrbitStabilizer,
                                     D.fingerprintRepresentatives));
   fi;
+  return s;
+end);
+
+BindGlobal("IdentificationPermGroup",function(D,el)
+local s,l;
+  s:=CheapIdentificationPermGroup(D,el);
   if IsBound(D.usefitfree) and not s in D.nocanonize then
     l:=First(D.faclaimg,x->x[1]=s);
     l:=TFCanonicalClassRepresentative(D.group,[el]:candidatenums:=l[2]);
     Add(s,l[1][2]);
   fi;
   return s;
-end );
+end);
 
 
 #############################################################################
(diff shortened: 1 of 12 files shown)


lib/ringsc.gi
```diff
@@ -917,11 +917,25 @@ end);
...
 function(R,e)
-  if not IsFinite(R) then
+local o,pow,a;
+  if not IsFinite(R) or One(R)=fail then
     TryNextMethod();
   fi;
-  return One(R)<>fail
-         and ForAny(Enumerator(R),x->x*e=One(R) and e*x=One(R));
+
+  # compute powers of e until we reach one or repeat
```
ThomasBreuer, line 925: The previous code checked whether there is an element that is both a left and a right inverse.
The proposed code checks for a power that is the identity element, which is sufficient if the multiplication is associative, but why is it correct in nonassociative situations? The code establishes a left inverse, but why is it a right inverse?

lib/ghomperm.gi
```diff
@@ -1074,6 +1074,8 @@ InstallMethod( CompositionMapping2, "group hom. with perm group hom.",
     function( hom1, hom2 )
     local   prd,  stb,  levs,  S,t,i,oli;
 
+    Size(Source(hom2));
+    Size(Range(hom2));
     stb := StructuralCopy( StabChainMutable( hom2 ) );
```
ThomasBreuer, line 1079: A comment why the `Size` computations are useful would be helpful.
I would have guessed that `StabChainMutable` will compute the orders of source and range as soon as the values are interesting.

lib/ctblperm.gi
```diff
@@ -213,16 +220,22 @@ local k,structures,ambiguousStructures,i,j,p,cem,ces,z,t,cen,a,
...
   D.ids:=[];
+  D.chids:=[];
   D.rids:=[];
+  D.canreps:=[];
   for i in [1..D.klanz] do
     D.ids[i]:=D.identification(D,D.classreps[i]);
+    D.chids[i]:=D.cheapIdentification(D,D.classreps[i]);
     D.rids[i]:=
-     D.rationalidentification(D,D.classreps[i]);
+    D.rationalidentification(D,D.classreps[i]);
```
ThomasBreuer, line 230: Shouldn't the line be indented by one more space than before?

ThomasBreuer (commented):
I have added some comments.

lib/ringsc.gi
hulpke: Ah. My world is associative. Will fix.

hulpke (commented):
With Thomas remark this requires changes. I will do so in my *original* PR (because that is the one I can change)

lib/ctblperm.gi
hulpke: Will indent (in original PR)

lib/ringsc.gi
hulpke: Fixed in original

fingolfin:
I have backported the fixes here plus two more commits from #5394

hulpke (approved):

PR #5436, opened 2023-04-05
CI: ubuntu-18.04 are not supported anymore

fingolfin:
... so I am trying to adapt it to use Ubuntu 20.04.

But from what I recall, this may be difficult, as some packages may be
missing. In which case I intend to just drop this test entirely.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -86,14 +86,6 @@ jobs:
             test-suites: "testinstall"
             extra: "HPCGAP=yes ABI=64"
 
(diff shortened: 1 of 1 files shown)


fingolfin:
Aha:
```
E: Unable to locate package 4ti2:i386
E: Unable to locate package libcdd-dev:i386
E: Unable to locate package libfplll-dev:i386
E: Unable to locate package libmpfi-dev:i386
E: Unable to locate package pari-gp:i386
E: Unable to locate package singular:i386
```

OK, then: bye-bye 32bit package tests

